PR #385, opened 2019-05-23
Move builds entirely into images

dparalen:
The binary build process now happens entirely in images

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ CSI_IMAGE_VERSION?=canary
 $(info rbd    image settings: $(RBD_IMAGE_NAME) version $(RBD_IMAGE_VERSION))
 $(info cephfs image settings: $(CEPHFS_IMAGE_NAME) version $(CEPHFS_IMAGE_VERSION))
 
+BUILD_IN_DOCKER?=false
+
 all: cephcsi
 
 test: go-test static-check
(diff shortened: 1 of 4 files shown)


dparalen:
Hi,

I'd like to introduce the possibility to build ceph-csi inside a docker image esp. to be able to build on different systems like for instance on a stable Debian...
Feedback appreciated!

Thanks!
milan

Madhu-1:
@dparalen below are my concerns

one has to do a docker build (building the docker context, installing dep and copying source code and vendor directory. which is time-consuming) for code compilation.

even if a developer using different golang version other than `golang:1.11` which is mentioned in dockerfile. finding and fixing errors due to different golang versions might be a problem.

this will become a kind of problem for developers to do local testing

@humblec @ShyamsundarR  any thoughts?

JohnStrunk:
By setting up the Dockerfile efficiently, the docker build cache eliminates a lot of the build latency from repeated operations. Having a containerized build definitely goes a long way to creating a repeatable build process that works everywhere.

That said, if the main concern is Debian having an old version of golang, you may want to consider using [gvm](https://github.com/moovweb/gvm)

dparalen:
@Madhu-1 @JohnStrunk yeah, the latencies of repeated builds are amortised with Docker caching (some) intermediate image layers. The possibility to reproduce a build from a particular (historical) commit, wrapped in a container, with vendored deps and the "guest OS image" is also attractive to me. Another reason is being able to build for a particular host system on that same system --- in my case an arm64 machine. ...which can be worked around with cross-compiling but this way I hoped might be more useful...

Local-built unit testing might be an issue indeed; how about making the Docker build somehow conditional in the Makefile?

Thanks for the input!

humblec:
>how about making the Docker build somehow conditional in the Makefile?

Considering the use case, its a worth try. Would you be able to give a try on this ?

dparalen:
> > how about making the Docker build somehow conditional in the Makefile?
> 
> Considering the use case, its a worth try. Would you be able to give a try on this ?

yeah, I'll try to come up with some conditional build logic inside the Makefile

> 
> Or what about keeping different build mechanism or reference on build from different arch directories, say one for x86 and another one for ARM?

do you mean to make some sort of an arch-dependent import into the Makefile?

...another thing is the different tagging that might be necessary for the containers built on e.g arm64 if they're supposed to be hosted on Quay. IIRC Quay doesn't support manifest lists. I've been wondering how to tackle that too...

Thanks!

dparalen:
> > > how about making the Docker build somehow conditional in the Makefile?
> > 
> > 
> > Considering the use case, its a worth try. Would you be able to give a try on this ?
> 
> yeah, I'll try to come up with some conditional build logic inside the Makefile
> 
> > Or what about keeping different build mechanism or reference on build from different arch directories, say one for x86 and another one for ARM?
> 
> do you mean to make some sort of an arch-dependent import into the Makefile?
> 
> ...another thing is the different tagging that might be necessary for the containers built on e.g arm64 if they're supposed to be hosted on Quay. IIRC Quay doesn't support manifest lists. I've been wondering how to tackle that too...
> 
> Thanks!

@humblec I didn't manage to figure out an easy way how to make the Dockerfile--Makefile build seamlessly conditional while as well leveraging the multi-stage Dockerfile feature. To my best knowledge I'd have to use some text pre-processor (m4) to be able to avoid Dockerfile content from repeating or use some ugly shell/sed magic from inside the Makefile.

I'm pushing a "verbose" conditional build, using distinct `Dockerfile.builder`, for container-based builds.

Another option would be to allow build of the binaries to happen in a container without modifying the resulting container build similarly to what e.g [kube-router](https://github.com/cloudnativelabs/kube-router/blob/master/Makefile#L42#L49) does...

Feedback welcome!

humblec:
@dparalen may be we could start with simple approach first.  Lets have a seperate make/build directory for `ARM` arch in this repo. We can tag the docker image...etc with `arm` string in it. What do you think ?

Madhu-1:
@dparalen any update on this one? would you still like to work on this one?

PR #224, opened 2019-02-22
[Proposal] Maintain a stateless CSI implementation

ShyamsundarR:
Introduce CSI plugin configuration, per Ceph cluster that is
used for provisioning storage, in lieu of the existing config
map based volume store.

Signed-off-by: ShyamsundarR <srangana@redhat.com>

Issue #328, opened 2019-04-17
[csi-cephfs] use ceph mgr volumes module to provision/manage csi volumes


ajarr:
Currently, cephfs CSI drivers fuse/kernel mounts a ceph filesytem to create/purge csi volumes, which are cephfs subdirectories with a specific quota and layout. There are already ceph CLI commands (nautilus version of Ceph onwards) that can create/delete csi volumes. The cephfs CSI driver can issue the following commands to ceph-mgr daemon to do this,
```
$ ceph fs subvolume create <filesystem-name> <csi-vol-id>  --size <size in GB>
$ ceph fs subvolume rm <filesystem-name> <csi-vol-id>`
```

humblec:
@ajarr as per yesterday's discussion we are tracking this for `v1.1` release. 
@Madhu-1 @ShyamsundarR any further thoughts? 

As discussed in CSI call, one outstanding item is the migration story. The implementation and code availability looks feasible by v1.1 release deadline.

humblec:
Current plan is to get the volume provisioning and deletion based on ceph mgr in v1.1. Once that is in , we will enhance further functionalities.

gman0:
Does this mean dropping support for pre-Nautilus clusters? That's not exactly ideal.

humblec:
>Does this mean dropping support for pre-Nautilus clusters? That's not exactly ideal.
@gman0 not really. the `v1.0` images will stay and will be supporting previous ceph clusters. However `v1.1` will be only supported with Nautilus clusters due to this dependency.

Madhu-1:
@ajarr anything pending on this one?

humblec:
This has been taken care by https://github.com/ceph/ceph-csi/releases v1.1.0.  @ajarr can you confirm and close the same?

ajarr:
Completed in V1.1.0

PR #390, opened 2019-05-30
Make CephFS plugin stateless reusing RADOS based journal scheme

### LINKED ISSUE #359: Provide ability to select CephFS fsname to provision from in the StorageClass
Currently the CephFS plugin storage class does not provide a means to select a file system name to provision from. As multiple FS support is present in CephFS, the storage class is better equipped to handle the same in the future, if the fsname is explicitly mentioned in the same.

This was based on a review conversation around volume ID encoding changes in the rbd plugin and can be found [here](https://github.com/ceph/ceph-csi/pull/288/commits/885da4d0b4d917064f45ec707098e6cfa6db1ed7#r270070908).

### LINKED ISSUE #382: Use static PV based provisioning for pre-provisioned CephFS volume support
OR, **Title: Use CephFS CSI storage class for dynamic provisioning only**

Current implementation of CephFS CSI plugin, has the option to use pre-provisioned CephFS subvolumes. This is done by crafting a StorageClass that contains `provisionVolume` as `false` and giving the `rootPath` a value of the actual path on the CephFS volume to mount and use. Further, when mentioning such pre-provisioned storage classes the secrets should contain a `userID` and `userKey` combination for use with the mount operation. (see [1])

The implementation for the above, makes the `CreateVolume` and `DeleteVolume` close to NO-OPs, just crafting and returning a `VolumeID` for use to identify the volume during mount operations.

An alternative to using pre-provisioned volumes is to create static PVs instead, and use the CSI nodeplugin to mount these PVs. The PV can have a `claimRef`, with required `name` and `namespace` keys and values, that can help match a PVC claim, and the PVC can have a `volumeName` specification to match the PV. Thus, retaining a PVC to PV 1:1 relationship.

I tried the above alternative, and can see that it functions as required, thus eliminating the need for the controller service to support the notion of detecting and faking Create/Delete for pre-provisioned CephFS volumes. A sample pod spec, all the way from defining the `userID` and key secret, to the pod using a PVC is provided here [2].
 
The CSI spec does not preclude using a nodeplugin exclusively, or rather does not mandate that only dynamically provisioned volumes must be honored by the nodeplugin. Thus, this is in line with the CSI specification.

This issue hence, is to remove the notion of static provisioning using a storage class, and instead using static PV provisioning, thus removing support for static provisioning or pre-provisioned CephFS volumes using a storage class from the 1.1.0 version of the plugin.

NOTE: 1.1.0 will retain backward compatibility with pre-provisioned volumes created prior to the 1.1.0 version 

[1] StorageClass provisionVolume details: https://github.com/ceph/ceph-csi/blob/d02e50aa9b5af752ccdb53c79a1a14bf001af81c/examples/cephfs/storageclass.yaml#L12-L19

[2] YAML manifest for static PV with PVC and pod using the same: https://gist.github.com/ShyamsundarR/1009b79966992470f588a277429ae2e1

ShyamsundarR:
This is a part of the stateless set of commits for CephCSI.

This commit removes the dependency on config maps to store cephFS provisioned
volumes, and instead relies on RADOS based objects and keys, and required
CSI VolumeID encoding to detect the provisioned volumes.

## Changes ##
- Provide backward compatibility to provisioned volumes by older plugin versions (1.0.0 or older)
- Remove Create/Delete support for statically provisioned volumes (fixes #382)
- Added namespace support to RADOS OMaps and used the same to store RADOS CSI objects and keys in the CephFS metadata pool
- Added support to mention fsname for CephFS provisioning (fixes #359)
- Changed field name in CSI Identifier to 'location', to denote a pool or fscid
- Updated mounter cache to use new scheme
- Required Helm manifests are updated
- Required documentation and other manifests are updated
- Made driver option 'metadatastorage' as optional, as fresh installs do not need to specify the same

## Testing done ##
- Create/Mount/Delete PVC
- Create/Delete 5 PVCs
- Mount version 1.0.0 PVC
- Delete version 1.0.0 PV
- Mount Statically defined PV/PVC/Pod
- Mount Statically defined version 1.0.0 PV/PVC/Pod
- Delete Statically defined version 1.0.0 PV/PVC/Pod
- Node restart when mounted to test mountcache
- Use InstanceID other than 'default'
- RBD basic round of tests, as namespace is added to OMaps
- csitest against ceph-fs plugin
  - NOTE: CephFS plugin still does not detect and address already created
  volumes but of a different size
- Test not providing any value to the metadata storage parameter

## Related issues ##

Fixes: #382, #359
Implements proposal as stated in #224 

## Future concerns ##

- volumeIDMutex lock can be deleted as it is not providing any further value for retaining idempotent nature of requests, both in RBD and CephFS plugin implementations. Some more study is required before taking this action though.
- Like missing subvolumes are treated as a success during delete, missing per subvolume credentials should also be treated as a success during deletes and further deletion steps should be followed
- CephFS plugin should detect subvolume feature compatibility when there are request name conflicts and respond with a success on matches or an error AlreadyExists on mismatches.
- Need to take a hard re-look at the mount cache, as it is storing secrets in config maps and that can be a violation of trust and have security implications.


ShyamsundarR:
Attention: @ajarr @poornimag 

For moving forward with #328 and any reviews that can be provided, thanks.

gman0:
Please keep in mind that having the ability to mount pre-provisioned volumes with  having access only to that particular volume is absolutely crucial for things like shares managed by OpenStack Manila to work. There are cases where the operators may have limited direct access to the ceph cluster.

`NodeStageVolume` should be able to get by with {mons, path to the directory} in `VolumeContext` and ceph user secret in `Secret`. Could be done as a fallback option perhaps

ShyamsundarR:
> Please keep in mind that having the ability to mount pre-provisioned volumes with having access only to that particular volume is absolutely crucial for things like shares managed by OpenStack Manila to work. There are cases where the operators may have limited direct access to the ceph cluster.
> 
> `NodeStageVolume` should be able to get by with {mons, path to the directory} in `VolumeContext` and ceph user secret in `Secret`. Could be done as a fallback option perhaps

I have already provided the option for static PVs in the `NodeStageVolume` code with this PR and this can be traced from [here](https://github.com/ceph/ceph-csi/pull/390/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml&owned-by%5B%5D=#diff-2c3d26c7b241d30c75b15c67e15fa7a9R102).

I still need to add documentation and possibly example `yaml` for the same, to help with detailing how this would look.

For my testing I used [this](https://gist.github.com/ShyamsundarR/1009b79966992470f588a277429ae2e1) manifest and should serve as an example to understand if this satisfies your request. For example it avoids monitors in the VolumeContext and uses the ceph-csi-config config map, and should help with changing MONs more easily even for static PVs. @gman0  Thoughts?

Issue #378, opened 2019-05-17
Backward compatibility and migration to 1.1 volumes from existing 1.0 versions


ShyamsundarR:
As the plugin moves from the current config maps to instead use a more descriptive VolumeID and RADOS objects (once #312 is merged), to store Kubernetes VolumeName and its backing image details, it needs to ensure the following for existing users:

1) Backward compatibility with older Volumes provisioned
2) Migration possibilities to the new scheme

#### 1) Backward compatibility:
**NOTE:** This is as mentioned in this comment, https://github.com/ceph/ceph-csi/issues/296#issuecomment-486466182

For volumes created using existing 1.0 versions of the Ceph-CSI plugins the following actions would be supported by 1.1 version of the plugin,
    DeleteVolume
    DeleteSnapshot
    NodePublishVolume (IOW, mounting and using the volume for required IO operations)

And, the following would be unsupported:
    CreateVolume from snapshot source which is from an older version
    CreateSnapshot from volume source which is from an older version

NOTE: Support for 0.3 created PVs requires feasibility analysis to ensure the above compatibility can be guaranteed for the same.

Method for doing this, would be to continue using the **existing** config maps, and on detecting older style VolumeID in the mentioned RPC requests, read the required data from the config maps and process the request.

#### 2) Migration
These were discussed in issue #296

Option A: https://github.com/ceph/ceph-csi/issues/296#issuecomment-487183805
In short, create a new PVC using the new version of the plugin, and using Ceph CLIs clone the image backing the older PVC to the image backing the newer PVC, further updating pods that use the older PVC to the new PVC.

Option B: https://github.com/ceph/ceph-csi/issues/296#issuecomment-487647575
Deals with PV.ClaimRef juggling as in the link above.

Option C: Migrate data using a PV to PV data copy

Option A is what I have tested and am far more comfortable recommending, as that would update the driver name as well as the required VolumeID and other fields. Hence makes it future proof from a Ceph-CSI implementation.

Further with Option-A and the backward compatibility included as above, the migration can be staged, and need not happen in one go. Hopefully alleviating some concerns around down time for the pods using the PVs.

Documentation and instructions for achieving the same would be provided.

**Other options:**
- convert config maps to RADOS objects as required by the 1.1 version of the plugin
  - This will not address the VolumeID that is already exchanged with the Kubernetes and is immutable and hence cannot be modified
- csi-translation-lib
[...]

PR #391, opened 2019-05-31
Support mounting and deleting version 1.0.0 RBD volumes

ShyamsundarR:
This commit adds support to mount and delete volumes provisioned by older
plugin versions (1.0.0) in order to support backward compatibility to 1.0.0
created volumes.

It adds back the ability to specify where older meta data was specified, using
the metadatastorage option to the plugin. Further, using the provided meta data
to mount and delete the older volumes.

It also supports a variety of ways in which monitor information may have been
specified (in the storage class, or in the secret), to keep the monitor
information current.

Testing done:
- Mount/Delete 1.0.0 plugin created volume with monitors in the StorageClass
- Mount/Delete 1.0.0 plugin created volume with monitors in the secret with
  a key "monitors"
- Mount/Delete 1.0.0 plugin created volume with monitors in the secret with
  a user specified key
- PVC creation and deletion with the current version (to ensure at the minimum
  no broken functionality)
- Tested some negative cases, where monitor information is missing in secrets
  or present with a different key name, to understand if failure scenarios work
  as expected

Updates #378

Follow-up work:
- Documentation on how to upgrade to 1.1 plugin and retain above functionality
  for older volumes

### DIFF
--- a/pkg/rbd/controllerserver.go
+++ b/pkg/rbd/controllerserver.go
@@ -38,6 +38,7 @@ const (
 // controller server spec.
 type ControllerServer struct {
 	*csicommon.DefaultControllerServer
+	MetadataStore util.CachePersister
 }
 
 func (cs *ControllerServer) validateVolumeReq(req *csi.CreateVolumeRequest) error {
@@ -83,7 +84,7 @@ func (cs *ControllerServer) parseVolCreateRequest(req *csi.CreateVolumeRequest) 
 	}
 
 	// if it's NOT SINGLE_NODE_WRITER and it's BLOCK we'll set the parameter to ignore the in-use checks
-	rbdVol, err := genVolFromVolumeOptions(req.GetParameters(), (isMultiNode && isBlock))
+	rbdVol, err := genVolFromVolumeOptions(req.GetParameters(), nil, (isMultiNode && isBlock), false)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
@@ -226,6 +227,50 @@ func (cs *ControllerServer) checkSnapshot(req *csi.CreateVolumeRequest, rbdVol *
 	return nil
 }
 
+// DeleteLegacyVolume deletes a volume provisioned using version 1.0.0 of the plugin
+func (cs *ControllerServer) DeleteLegacyVolume(req *csi.DeleteVolumeRequest, cr *util.Credentials) (*csi.DeleteVolumeResponse, error) {
+	volumeID := req.GetVolumeId()
+
+	if cs.MetadataStore == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "missing metadata store configuration to"+
+			" proceed with deleting legacy volume ID (%s)", volumeID)
+	}
+
+	idLk := legacyVolumeIDLocker.Lock(volumeID)
+	defer legacyVolumeIDLocker.Unlock(idLk, volumeID)
+
+	rbdVol := &rbdVolume{}
+	if err := cs.MetadataStore.Get(volumeID, rbdVol); err != nil {
+		if err, ok := err.(*util.CacheEntryNotFound); ok {
+			klog.V(3).Infof("metadata for legacy volume %s not found, assuming the volume to be already deleted (%v)", volumeID, err)
+			return &csi.DeleteVolumeResponse{}, nil
+		}
+
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	// Fill up Monitors
+	if err := updateMons(rbdVol, nil, req.GetSecrets()); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	// Update rbdImageName as the VolName when dealing with version 1 volumes
+	rbdVol.RbdImageName = rbdVol.VolName
+
+	klog.V(4).Infof("deleting legacy volume %s", rbdVol.VolName)
+	if err := deleteImage(rbdVol, cr); err != nil {
+		// TODO: can we detect "already deleted" situations here and proceed?
+		klog.V(3).Infof("failed to delete legacy rbd image: %s/%s with error: %v", rbdVol.Pool, rbdVol.VolName, err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if err := cs.MetadataStore.Delete(volumeID); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	return &csi.DeleteVolumeResponse{}, nil
+}
+
 // DeleteVolume deletes the volume in backend and removes the volume metadata
 // from store
 func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
@@ -247,6 +292,18 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 
 	rbdVol := &rbdVolume{}
 	if err := genVolFromVolID(rbdVol, volumeID, cr); err != nil {
+		// If error is ErrInvalidVolID it could be a version 1.0.0 or lower volume, attempt
+		// to process it as such
+		if _, ok := err.(ErrInvalidVolID); ok {
+			if isLegacyVolumeID(volumeID) {
+				klog.V(2).Infof("attempting deletion of potential legacy volume (%s)", volumeID)
+				return cs.DeleteLegacyVolume(req, cr)
+			}
+
+			// Consider unknown volumeID as a successfully deleted volume
+			return &csi.DeleteVolumeResponse{}, nil
+		}
+
 		// if error is ErrKeyNotFound, then a previous attempt at deletion was complete
 		// or partially complete (image and imageOMap are garbage collected already), hence return
 		// success as deletion is complete
@@ -281,6 +338,12 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	if err := deleteImage(rbdVol, cr); err != nil {
 		klog.Errorf("failed to delete rbd image: %s/%s with error: %v",
 			rbdVol.Pool, rbdVol.RbdImageName, err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if err := undoVolReservation(rbdVol, cr); err != nil {
+		klog.Errorf("failed to remove reservation for volume (%s) with backing image (%s) (%s)",
+			rbdVol.RequestName, rbdVol.RbdImageName, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -67,6 +67,22 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 	disableInUseChecks := false
 
+	isLegacyVolume := false
+	volName, err := getVolumeName(req)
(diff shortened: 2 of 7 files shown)


ShyamsundarR:
@kfox1111  this commit adds back using the existing config maps to enable backward compatibility. Tagging you for your attention/perusal.

ShyamsundarR:
@humblec Added the 1.1.0 release label as the commitment was to provide backward compatibility with existing config maps in 1.0 release, and hence this PR is also required for the 1.1.0 release.

ShyamsundarR:
> Sorry I missed reviewing this before. I'm not very familiar with the code base so can't review too closely. but in general, it looks ok to me.

No worries, I was more looking at usage and such, than a code review. Thanks for taking a peek.

ShyamsundarR:
@humblec @Madhu-1 PTAL (rebase will be sometime later today)

pkg/rbd/controllerserver.go
```diff
@@ -235,6 +271,18 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 
 	rbdVol := &rbdVolume{}
 	if err := genVolFromVolID(rbdVol, volumeID, req.GetSecrets()); err != nil {
+		// If error is ErrInvalidVolID it could be a version 1.0.0 or lower volume, attempt
+		// to process it as such
+		if _, ok := err.(ErrInvalidVolID); ok {
+			if cs.MetadataStore != nil && isV1VolumeID(volumeID) {
```
dillaman: Nit: if you can detect that it's a V1 format volume ID, why not just move this block to before you attempt to lookup the volume from the omap?

pkg/rbd/nodeserver.go
```diff
@@ -65,9 +65,20 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	volName, err := getVolumeName(req)
 	if err != nil {
-		return nil, err
+		// error ErrInvalidVolID may mean this is an 1.0.0 version volume, check for name
+		// pattern match in addition to error to ensure this is a likely v1.0.0 volume
+		if _, ok := err.(ErrInvalidVolID); !ok || !isV1VolumeID(req.GetVolumeId()) {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+
+		volName, err = getVolumeNameVersion1(req)
```
dillaman: Nit: same comment here about testing for V1 and avoid the omap test.

pkg/rbd/nodeserver.go
```diff
@@ -111,18 +122,40 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-		klog.Errorf("error decoding volume ID (%s) (%s)", err, req.GetVolumeId())
-		return "", status.Error(codes.InvalidArgument, err.Error())
+		err = fmt.Errorf("error decoding volume ID (%s) (%s)", err, req.GetVolumeId())
+		return "", ErrInvalidVolID{err}
 	}
 
 	return volJournal.NamingPrefix() + vi.ObjectUUID, nil
 }
 
+func getVolumeNameVersion1(req *csi.NodePublishVolumeRequest) (string, error) {
```
dillaman: Nit: name ```getVolumeNameV1``` to keep V1 naming consistent

pkg/rbd/rbd_util.go
```diff
@@ -418,7 +416,66 @@ func getIDs(options map[string]string) (adminID, userID string) {
...
+	}
+
+	if rbdVol.Monitors == "" {
+		return errors.New("either monitors or monValueFromSecret must be set")
+	}
+
+	return nil
+}
+
+func genVolFromVolumeOptions(volOptions, credentials map[string]string, disableInUseChecks, v1Volume bool) (*rbdVolume, error) {
```
dillaman: Nit: ```v1Volume``` -> ```isV1Volume``` (thoughout)?

pkg/rbd/nodeserver.go
```diff
@@ -65,9 +65,20 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}()
 	disableInUseChecks := false
 
-	volName, err := ns.getVolumeName(req)
+	v1Volume := false
```
dillaman: Nit: ```isV1Volume```?

ShyamsundarR:
> The "V1" volume might be a misnomer since I thought the new-style volume IDs have a version 1 prefix. Perhaps instead of "V1", it should be "legacy" throughout?

Moved required naming to `legacy` instead of V1, makes better sense to do so.

pkg/rbd/nodeserver.go
ShyamsundarR: Updated to use `legacy` naming instead.

pkg/rbd/rbd_util.go
ShyamsundarR: Done.

pkg/rbd/nodeserver.go
ShyamsundarR: Done.

pkg/rbd/controllerserver.go
ShyamsundarR: The call to `genVolFromVolID` first unpacks the `VolID` and will fail there with the error `ErrInvalidVolID`. It hence does not make any ceph/rados calls to determine the failure to decode the `VolID`. (same with the call to `getVolumeName` in the nodeserver.go file)

The further choice was to fall back to legacy rather than check legacy first. If this hampers readability (as the early failure is not obvious) let me know if it needs a reversal still.

pkg/rbd/controllerserver.go
```diff
@@ -247,6 +286,18 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
...
+		// If error is ErrInvalidVolID it could be a version 1.0.0 or lower volume, attempt
+		// to process it as such
+		if _, ok := err.(ErrInvalidVolID); ok {
+			if cs.MetadataStore != nil && isLegacyVolumeID(volumeID) {
+				klog.V(2).Infof("attempting deletion of potential legacy volume (%s)", volumeID)
+				return cs.DeleteLegacyVolume(req, cr)
+			}
+
+			// Consider unknown volumeID as a successfully deleted volume
+			return &csi.DeleteVolumeResponse{}, nil
```
dillaman: Seems like if the volume id is an invalid format, it should through an error and if it's a known legacy format and you don't have the metadata store configured, it should be yet another.

pkg/rbd/controllerserver.go
dillaman: I didn't look at `genVolFromVolID` to see that it bails out early if it cannot decode the volume id. In that case, I think it's fine as-is.

pkg/rbd/controllerserver.go
ShyamsundarR: Ah ok, this is a weirdness of the CSI spec tester that expects that an invalid/unknown VolID is a success (see the test for the same [here](https://github.com/kubernetes-csi/csi-test/blob/65a7181bfd6ab2ed4753816774014bd8b50615ae/pkg/sanity/controller.go#L855-L864)).

Now, we have enough data to ensure the VolID is valid, and can return errors, but the effort to pass the spec test may require us to treat this as a success. Thoughts?

dillaman: Given that test, how about it throws an error if it sees that it's a legacy volid but you haven't configured the metadata store?

ShyamsundarR: Oh yes, my bad. Got distracted with the csi-test reference... fixing.

ShyamsundarR: fixed (moved the store check to the legacy function to reduce nested if's and really just make the linters happy).

pkg/rbd/controllerserver.go
```diff
@@ -226,6 +227,50 @@ func (cs *ControllerServer) checkSnapshot(req *csi.CreateVolumeRequest, rbdVol *
 	return nil
 }
 
+// DeleteVolumeV1 deletes a volume provisioned using version 1.0.0 of the plugin
```
humblec: DeleteLegacyVolume

ShyamsundarR: @humblec updated.

humblec:
@Madhu-1 PTAL .

pkg/rbd/rbd_util.go
```diff
@@ -336,7 +338,66 @@ func getMonsAndClusterID(options map[string]string) (monitors, clusterID string,
 	return
 }
 
-func genVolFromVolumeOptions(volOptions map[string]string, disableInUseChecks bool) (*rbdVolume, error) {
+// isLegacyVolumeID checks if passed in volume ID string conforms to volume ID naming scheme used
+// by the version 1.0.0 (legacy) of the plugin, and returns true if found to be conforming
+func isLegacyVolumeID(volumeID string) bool {
```
Madhu-1: can this be reused for cephfs?

pkg/rbd/rbd_util.go
```diff
@@ -336,7 +338,66 @@ func getMonsAndClusterID(options map[string]string) (monitors, clusterID string,
...
+	if uuid.Parse(volumeID[12:]) == nil {
+		return false
+	}
+
+	return true
+}
+
+// upadateMons function is used to update the rbdVolume.Monitors for volumes that were provisioned
+// using the 1.0.0 version (legacy) of the plugin.
+func updateMons(rbdVol *rbdVolume, options, credentials map[string]string) error {
```
Madhu-1: same here.

pkg/rbd/controllerserver.go
```diff
@@ -247,6 +292,18 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
...
+		// If error is ErrInvalidVolID it could be a version 1.0.0 or lower volume, attempt
+		// to process it as such
+		if _, ok := err.(ErrInvalidVolID); ok {
+			if isLegacyVolumeID(volumeID) {
+				klog.V(2).Infof("attempting deletion of potential legacy volume (%s)", volumeID)
+				return cs.DeleteLegacyVolume(req, cr)
+			}
+
+			// Consider unknown volumeID as a successfully deleted volume
+			return &csi.DeleteVolumeResponse{}, nil
```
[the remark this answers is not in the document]
ShyamsundarR: See discussion [here](https://github.com/ceph/ceph-csi/pull/391#discussion_r299469504)

pkg/rbd/rbd_util.go
ShyamsundarR: Prefix would be different for CephFS legacy volume, so direct reuse is not possible.

Madhu-1: then I would like to move this function to utils package and pass prefix  as input

pkg/rbd/rbd_util.go
ShyamsundarR: CephFS code is slightly different in the way it is written to extract code from options and from secrets. Either can be made to align with the other.

With the changes to CephFS using ceph Manager commands, should we do this now? As that commit is leaving out backward compat to be added again later afaict.

pkg/rbd/rbd_util.go
ShyamsundarR: CephFS is not checking ID format as well, so it will not be called from CephFS even if moved, should I still move it?

pkg/rbd/rbd_util.go
Madhu-1: we can do  it in while addressing backward compatibility in ceph manager commands

pkg/rbd/rbd_util.go
Madhu-1: then no need to move it to utils.

PR #392, opened 2019-05-31
Remove redundant pool parameter from snapshot class

### LINKED ISSUE #379: Additional parameter `pool` in SnapshotClass for RBD seems redundant
The SnapshotClass for RBD requires a `pool` parameter [1]. This is redundant as a snapshot is not created on a different pool than the source image of the snapshot (refer rbd man page [2]).

Further, when a snapshot needs to be created its source CSI `VolumeID` is passed to the creation call, and hence the source volumes pool needs to be reused to create the snapshot.

Similarly to clone a snapshot, the create request would come in with a `SnapshotID` to help identify the snapshot pool, and the same create request parameters would contain the storage class based `pool` parameter to create the clone into (as clones can be in different pools as compared to their parent snapshots).

Thus, the parameter `pool` seems redundant in the snapshot class and should be removed to improve ease of use.

[1] pool reference in snapshotclass: https://github.com/ceph/ceph-csi/blob/f60a07ae8268136f40de3d62537f4f8b020f0ec4/examples/rbd/snapshotclass.yaml#L8

[2] rbd man page: http://docs.ceph.com/docs/nautilus/rbd/rbd-snapshot/

ShyamsundarR:
The SnapshotClass for RBD requires a pool parameter. This is redundant
as a snapshot is not created on a different pool than the source image
of the snapshot (refer rbd man page).

Further, when a snapshot needs to be created its source CSI VolumeID
is passed to the creation call, and hence the source volumes pool
needs to be reused to create the snapshot.

Similarly to clone a snapshot, the create request would come in with a
SnapshotID to help identify the snapshot pool, and the same create
request parameters would contain the storage class based pool parameter
to create the clone into (as clones can be in different pools as
compared to their parent snapshots).

Thus, the parameter pool seems redundant in the snapshot class and
should be removed to improve ease of use.

Fixes #379


humblec:
LGTM. Thanks !

Issue #350, opened 2019-05-03
Could not create the CephFsvolume 


kannanvr:
Hi,
   I could not create the CephFs volume on rook-ceph
I have deployed rook-ceph. Also, I have deployed filesystem Crd on rook-ceph
Then I have deployed the Ceph-CSI to allocate the PVC dynamically

Following is my secret
```
---
apiVersion: v1
kind: Secret
metadata:
  name: csi-cephfs-secret
  namespace: default
data:
  # Required if provisionVolume is set to false
  userID: YWRtaW4K
  userKey: QVFBMHdNcGNWYkZ0RUJBQTU4Y3FUZmdhY0tET1lHbTRNZjdxQkE9PQo=

  # Required if provisionVolume is set to true
  adminID: YWRtaW4K
  adminKey: QVFBMHdNcGNWYkZ0RUJBQTU4Y3FUZmdhY0tET1lHbTRNZjdxQkE9PQo=
```

Also, I have mentioned the rook-ceph Mon IP and port as a monitor IP's on storageclass.
Then I have created the PVC as mentioned on the example of this project
But PVC is not created by provisoner.
Following are the logs, i have observed from csi-cephfsplugin-provisioner-0 on the container csi-provisioner
```
[root@uklab9 cephfs]# kubectl logs -f csi-cephfsplugin-provisioner-0 -c csi-provisioner
I0503 00:55:19.542766       1 csi-provisioner.go:82] Version: v1.0.1-0-g9a258a0a
I0503 00:55:19.542896       1 csi-provisioner.go:96] Building kube configs for running in cluster...
I0503 00:55:19.564932       1 controller.go:197] Connecting to unix:///csi/csi-provisioner.sock
I0503 00:55:19.565226       1 controller.go:224] Still trying, connection is CONNECTING
I0503 00:55:19.565500       1 controller.go:224] Still trying, connection is TRANSIENT_FAILURE
I0503 00:55:20.565627       1 controller.go:224] Still trying, connection is CONNECTING
I0503 00:55:20.565787       1 controller.go:221] Connected
I0503 00:55:20.565821       1 controller.go:188] GRPC call: /csi.v1.Identity/GetPluginInfo
I0503 00:55:20.565835       1 controller.go:189] GRPC request: {}
I0503 00:55:20.571245       1 controller.go:191] GRPC response: {"name":"cephfs.csi.ceph.com","vendor_version":"1.0.0"}
[...]

Issue #318, opened 2019-04-12
The CephX key should never be passed via a command-line argument


dillaman:
When using Ceph CLI tools, never use the ```--key``` parameter. Instead, write the key to a temporary file and point the ```--keyfile``` optional to the file.

Issue #98, opened 2018-11-06
Can't get example working: GRPC error: <nil>


CompileNix:
I was following the readme from `examples/README.md`.
The problem happens after `kubectl create -f pvc.yaml`
The Persistent Volume Claim `rbd-pvc` hangs in the `pending` state.

I noticed that the csi-rbdplugin-provisioner and csi-rbdplugin-attacher pod are not able to reach anything in the cluster, local network or internet but the csi-rbdplugin pod can reach the k8s dns server, all ceph mon's and hosts on the internet. other pods on the same k8s nodes have no network problems, too.

Here are the logs:
[logs-from-csi-provisioner-in-csi-rbdplugin-provisioner-0.txt](https://github.com/ceph/ceph-csi/files/2555403/logs-from-csi-provisioner-in-csi-rbdplugin-provisioner-0.txt)
[logs-from-csi-rbdplugin-attacher-in-csi-rbdplugin-attacher-0.txt](https://github.com/ceph/ceph-csi/files/2555404/logs-from-csi-rbdplugin-attacher-in-csi-rbdplugin-attacher-0.txt)
[logs-from-driver-registrar-in-csi-rbdplugin-8zkw6.txt](https://github.com/ceph/ceph-csi/files/2555405/logs-from-driver-registrar-in-csi-rbdplugin-8zkw6.txt)

compilenix:
the network problem was on my end and had nothing to do with ceph-csi.

it's still not working and the logs are looking the same as uploaded.

rootfs:
at least the ceph keyring doesn't look right to me in the provisioner log, there was an extra trailing `\n`
`AQDi4QVa4oKkKRAAjNMW7IzhwfqSDFWyW2jaxQ==\n`

Please recreate the secret with 
`echo -n AQDi4QVa4oKkKRAAjNMW7IzhwfqSDFWyW2jaxQ==|base64`

rootfs:
please also attach the driver log from container csi-rbdplugin here https://github.com/ceph/ceph-csi/blob/master/deploy/rbd/kubernetes/csi-rbdplugin.yaml#L41

compilenix:
Thanks, i've recreated the secret, this part looks fine now.
A pvc can't be created, yet.

Here are the current logs:
[...]

Issue #80, opened 2018-10-02
doc: adminid and userid not documented


pohly:
`examples/rbd/storageclass.yaml` contains these settings:
```
    # Ceph users for operating RBD
    adminid: admin
    userid: kubernetes
```

Those are not mentioned under https://github.com/ceph/ceph-csi/blob/master/docs/deploy-rbd.md#configuration. Are both users needed?

I found the entire "Required secrets"  section a bit confusing:
- `the value is its password` - is that really called `password` in Ceph? I think this refers to the base64 encoded `key` that is stored in ceph keyrings, right? Linking to http://docs.ceph.com/docs/mimic/rbd/rados-rbd-cmds/#create-a-block-device-user might be useful here, with a few words how to use the result.
- `CSI RBD expects admin keyring and Ceph config file in /etc/ceph` - how can that be achieved when deplying in Kubernetes? Is it really necessary?

I have set up a test cluster, but haven't actually tried provisioning. I guess I'll find out soon whether I interpreted the instructions correctly ;-}

pohly:
Nope, looks like I didn't populate my secret correctly. The external-provisioner is sent binary data for the secrets and then complains (these are disposable secrets, so I can log them as they are):
```
[...]

PR #395, opened 2019-06-01
Modify RBD plugin to use a single ID and move the id and key into the…

### LINKED ISSUE #270: Suggested changes to ID and key requirements for rbd CSI plugin
Currently the CSI rbd plugin requires 2 IDs and its respective keys in secrets. Each ID seems to serve the following purpose,
- adminID: Used for create/delete/snapshot operations (IOW, controller service operations)
- userID: Used for mount/umount of the rbd image on nodes using the same (IOW, node service operations)

There are a few issues with the same, some of which is related to lack of documentation as stated in #80 but some others need consideration as follows,

1) Do we need separate ID for the Controller Service and the Node Service? Or, can these be folded into a single ID/key instead of requiring 2?

I can see the following hierarchy of secrets (with some further granularity possible) that maybe used, and the question is, if we need to separate the admin from the mounter as in the following list,
- Cluster-wide admin
- Per-pool admin
- Per-pool user/mounter
- Per-image user/mounter

2) Either-way, the ID and key can move into the secret, rather than mentioning the ID in the StorageClass and the key as a secret in the secret. This makes it easier to configure for a user, than having to do the same in 2 places. IOW, make this similar to a secret in CephFS plugin, where the secret carries both the ID and the key.

3) Further make the secret a single ID/key pair, such that if required the same can be used as the provisioner and the publisher secret, rather than repeating the pair of user.key within the secret and also repeating the same in the StorageClass

4) Ideally, if we needed to separate the provisioner secret from the node secret, these should not be placed in the same secret file, as otherwise the nodes get both secrets and that sort of defeats the purpose? I would suggest that we pass the secrets in different files as examples, instead of a single file (and if we choose to do (3) this would be handled automatically)

ShyamsundarR:
… secret

RBD plugin needs only a single ID to manage images and operations against a
pool, mentioned in the storage class. The current scheme of 2 IDs is hence not
needed and removed in this commit.

Further, unlike CephFS plugin, the RBD plugin splits the user id and the key
into the storage class and the secret respectively. Also the parameter name
for the key in the secret is noted in the storageclass making it a variant and
hampers usability/comprehension. This is also fixed by moving the id and the key
to the secret and not retaining the same in the storage class, like CephFS.

Fixes #270

Testing done:
- Basic PVC creation and mounting


examples/rbd/secret.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
...
   namespace: default
 data:
-  # Key value corresponds to a user name defined in ceph cluster
-  admin: BASE64-ENCODED-PASSWORD
-  # Key value corresponds to a user name defined in ceph cluster
-  kubernetes: BASE64-ENCODED-PASSWORD
+  # Key values correspond to a user name and its key, base 64 encoded, as
+  # defined in the ceph cluster. User ID should have required access to the
+  # 'pool' specified in the storage class
+  userID: BASE64-ENCODED-VALUE
```
dillaman: The ID shouldn't need to be base64 encoded

pkg/rbd/controllerserver.go
```diff
@@ -122,7 +128,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, err
 	}
 
-	found, err := checkVolExists(rbdVol, req.GetSecrets())
+	found, err := checkVolExists(rbdVol, cr.ID, cr.Key)
```
Madhu-1: pass `cr` struct to `checkVolExists`

Madhu-1: same applies wherever we are passing `ID and Key`

pkg/rbd/controllerserver.go
```diff
@@ -176,7 +182,12 @@ func (cs *ControllerServer) createBackingImage(rbdVol *rbdVolume, req *csi.Creat
 			return err
 		}
 	} else {
-		err = createImage(rbdVol, volSizeMiB, rbdVol.AdminID, req.GetSecrets())
+		cr, err := util.GetUserCredentials(req.GetSecrets())
```
Madhu-1: can we reuse the `cr` got in line 114

Madhu-1:
@ShyamsundarR please fix merge conflicts

examples/rbd/secret.yaml
ShyamsundarR: If the key `userID` is mentioned in a section named `stringData` it need not be encoded in base64 and can be mentioned in plain text. This is how a secret is specified in Kubernetes [1].

If required I can change the documentation to reflect the added `stringData` and mentioning the `userID` key within the same.

[1] Kubernetes manual secrets and `data` `stringData` options for base64 and plain text secrets: https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually

dillaman: I'd vote for ```stringData``` --- honestly, since the key is already base64 encoded, I always thought it was a little silly (usability-wise) that it needed to be encoded again, so perhaps both could be moved? or perhaps just move the ```userId``` back to the ```StorageClass```?

ShyamsundarR: Then `stringData` it is. I will move both, as the key otherwise is double base64 encoded, as you point out.

Fractured ID and key in 2 files, to me, is a far worse usability proposition, so no I would not go that way.

pkg/rbd/controllerserver.go
ShyamsundarR: Done.

Did this more across the code, so that it is cleaner, than just certain instances.

pkg/rbd/controllerserver.go
ShyamsundarR: I did not want to pass around `cr` in the `ControllerServer` methods and further the `req` is passed as is to this method, hence left it as is.

ShyamsundarR:
> @ShyamsundarR please fix merge conflicts

Done, also changed some code in e2e, to adapt to the changes storage class and secrets for RBD and CephFS.

ShyamsundarR:
@humblec PTAL

examples/cephfs/secret.yaml
```diff
@@ -4,11 +4,11 @@ kind: Secret
 metadata:
   name: csi-cephfs-secret
   namespace: default
-data:
+stringData:
   # Required for statically provisioned volumes
-  userID: BASE64-ENCODED-VALUE
-  userKey: BASE64-ENCODED-VALUE
+  userID: <plaintext ID>
+  userKey: <Ceph auth key corresponding to ID above>
```
humblec: why are we moving away from BASE64 to plain text here?

ShyamsundarR: See usability discussion here, https://github.com/ceph/ceph-csi/pull/395#discussion_r289988124

Basically, moving away to make it easy for an user t ospecify plain text IDs and key is already base64 encoded by Ceph.

ShyamsundarR:
@humblec PTAL (rebased post recent PR merges to master)

humblec:
@ShyamsundarR apart from PLAIN TEXT password, changes looks good to me. Please use base64 encoded values in secret.

dillaman:
> @ShyamsundarR apart from PLAIN TEXT password, changes looks good to me. Please use base64 encoded values in secret.

The CephX secrets are already base64 encoded, so this just eliminates the unnecessary double base64 encoding. Plus, it's not like an extra layer of base64 encoding offers any protection for the key.

ShyamsundarR:
> > @ShyamsundarR apart from PLAIN TEXT password, changes looks good to me. Please use base64 encoded values in secret.
> 
> The CephX secrets are already base64 encoded, so this just eliminates the unnecessary double base64 encoding. Plus, it's not like an extra layer of base64 encoding offers any protection for the key.

@humblec This would be my reasoning as well, and also on reading the difference between `data` and `stringData` as in the kubernetes docs [1], it is just a convenience.

[1] https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually

humblec:
@ShyamsundarR @dillaman the best practice of the secret key creation and use is `base 64` encoded values. Thats been the practice in the field and also the way people use secrets . I dont see a reason why we have to move away from it. IMO,  regardless of how the backend use it, it is better to store it in base 64 for kubernetes artifacts. @Madhu-1 what do you think ?

dillaman:
Please provide a reference for "the best practice of the secret key creation and use is base 64 encoded values".

Again, it's already base64 encoded and double-encoding it adds exactly zero value. Plus, it's confusing to users (look at the issues where folks have incorrectly re-encoded the CephX base64 key by either incorrectly appending a carriage return or not re-encoding it).

examples/cephfs/secret.yaml
```diff
@@ -4,11 +4,11 @@ kind: Secret
...
 
   # Required for dynamically provisioned volumes
-  adminID: BASE64-ENCODED-VALUE
-  adminKey: BASE64-ENCODED-VALUE
+  adminID: <plaintext ID>
```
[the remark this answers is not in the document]
ShyamsundarR: Please specify what the security issue is? The plain text `stringData` is just a user convenience, it is not stored and displayed in plain text if the secret is queried, just like passing it in using `data` instead.

The documentation here [1] states this: "stringData allows specifying non-binary secret data in string form. It is provided as a write-only convenience method. All keys and values are merged into the data field on write, overwriting any existing values. It is never output when reading from the API."

[1] Read `stringData` section: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#secret-v1-core

ShyamsundarR:
> > Please provide a reference for "the best practice of the secret key creation and use is base 64 encoded values".
> > Again, it's already base64 encoded and double-encoding it adds exactly zero value. Plus, it's confusing to users (look at the issues where folks have incorrectly re-encoded the CephX base64 key by either incorrectly appending a carriage return or not re-encoding it).
> 
> @dillaman @shyamsundar I dont know there exist a document called best practice. However I can point out resources like, in OCS we use `base64` encoded secret as shown here https://github.com/kubernetes/examples/blob/master/staging/persistent-volume-provisioning/glusterfs/glusterfs-secret.yaml . The examples in kubernetes repo also list `many` instances ( RBD ..etc) with `base 64` encoded secrets. Also think, in first place we put `base64` encoded values in secret here (` ceph-csi`) too. These examples lists from kube and the first samples/existing method of ceph-csi itself pointing out that thats the common or best practice people follow or used to. I dont want to mention it explictly, however, existing users also have a practice of using it in that way. Thats why I was wondering why to change it ? if thats not going to give us any advantages? If there are user issues or complaints about , using `base 64` is a pain or need to be reconsidered, please point , may be I missed those references. Please note, I am trying to share my view based on the experience, but it doesnot mean that, we should not change, but atleast I have to point out.

Repeating comment on security issue here again:
The plain text stringData is just a user convenience, it is not stored and displayed in plain text if the secret is queried, just like passing it in using data instead.

The documentation here [1] states this: "stringData allows specifying non-binary secret data in string form. It is provided as a write-only convenience method. All keys and values are merged into the data field on write, overwriting any existing values. It is never output when reading from the API."

I believe @humblec you are missing the usability point, and as already pointed out by @dillaman there are users who have incorrectly encoded the key or ID (with trailing newlines or missing the encoding). Here are some that I could find as issues #350 #98 #80 and there has been at least one or two more that I addressed in slack channels.

Finally: This PR blocks for me a couple of other PRs in flight (#391, and partially at least #318 ), so a decision here sooner the better.

[1] Read stringData section: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#secret-v1-core

humblec:
@ShyamsundarR please rebase.

shyamsundar:
Hi ,

I think you are talking to a wrong shyam Sundar. I don't know anything this
project.

Regards.

humblec:
> Hi , I think you are talking to a wrong shyam Sundar. I don't know anything this project. 
@shyamsundar  ah.. my bad .. sorry for the wrong notification, please discard. :(

ShyamsundarR:
> @ShyamsundarR please rebase.

@humblec Done PTAL.

humblec:
Thanks @ShyamsundarR . As discussed we may need a mention in the `release-note` or `changelog` of `v1.1` that, we are flexible on `plain text` or `base 64` input . We also need a followup PR to `Rook` to accommodate these changes in manifests shipped in Rook.

LGTM. thanks !

nehaberry:
@Madhu-1 @shyamsundar I have a question:

1. Does that mean for rbd, we only need to specify a single user and its key and not 2 users like cephfs (admin and kubernetes) ?.

Also,  that user in rbd/secret.yaml can either be an "admin" user, or a normal user , e.g. ceph_user1 ?  What if we want to give both , is that not supported now ? 

```
ceph_user1 = ceph auth get-or-create-key client.kubernetes mon  "allow profile rbd" osd "profile rbd pool=rbd"
```

userID: ${admin_id}
  userKey: ${admin_key}

OR

userID: ${ceph_user1_id}
  userKey: ${ceph_user1_key}

nehaberry:
@Madhu-1 @shyamsundar I tried this and it worked.. But I am a little confused if 2 users in secret are supported or not:


```

oc get pvc

rbd-pvc2     Bound     pvc-6f89e5ba-9896-11e9-9734-0a13693d844a   1Gi        RWO            csi-rbd2       4s


(new-ocs-ci) [nberry@localhost rbd]$ oc get secret/csi-rbd-secret2 -n default -o yaml
apiVersion: v1
data:
  adminID: YWRtaW4=
  adminKey: QVFCdVBSTmRxSFJTTmhBQVNGMmxlVkJkVGV3T3NBWFJHZnFxVXc9PQ==
  userID: a3ViZXJuZXRlcw==
  userKey: QVFDM1FCTmR4NzRmR0JBQVI1d1NVbkg3SjAvWFYyZUxLN1U4SWc9PQ==
kind: Secret
metadata:
  creationTimestamp: "2019-06-27T04:44:12Z"
  name: csi-rbd-secret2
  namespace: default
  resourceVersion: "373315"
  selfLink: /api/v1/namespaces/default/secrets/csi-rbd-secret2
  uid: 35db2587-9896-11e9-b24d-02b94c70af94
type: Opaque


(new-ocs-ci) [nberry@localhost rbd]$ oc get sc/csi-rbd2 -n default -o yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2019-06-27T04:44:43Z"
  name: csi-rbd2
  resourceVersion: "373465"
  selfLink: /apis/storage.k8s.io/v1/storageclasses/csi-rbd2
  uid: 47f4e969-9896-11e9-b24d-02b94c70af94
parameters:
  adminid: admin
  clusterID: openshift-storage
  csi.storage.k8s.io/node-publish-secret-name: csi-rbd-secret2
  csi.storage.k8s.io/node-publish-secret-namespace: default
  csi.storage.k8s.io/provisioner-secret-name: csi-rbd-secret2
  csi.storage.k8s.io/provisioner-secret-namespace: default
  imageFeatures: layering
  imageFormat: "2"
  pool: rbd
  userid: kubernetes
provisioner: rbd.csi.ceph.com
reclaimPolicy: Delete
volumeBindingMode: Immediate

```

Madhu-1:
@nehaberry `adminID` and `adminKey` are supported for cephfs not for rbd.
rbd requires only one user, even if you specify any keys other than userID and userKey we are not validating the extra keys in secret

Madhu-1:
> @Madhu-1 @shyamsundar I have a question:
> 
> 1. Does that mean for rbd, we only need to specify a single user and its key and not 2 users like cephfs (admin and kubernetes) ?.
> 
> Also, that user in rbd/secret.yaml can either be an "admin" user, or a normal user , e.g. ceph_user1 ? What if we want to give both , is that not supported now ?
> 
> ```
> ceph_user1 = ceph auth get-or-create-key client.kubernetes mon  "allow profile rbd" osd "profile rbd pool=rbd"
> ```
> 
> userID: ${admin_id}
> userKey: ${admin_key}
> 
> OR
> 
> userID: ${ceph_user1_id}
> userKey: ${ceph_user1_key}

you can  have both users in secret, either you  can use admin user  or ceph_user1 but he should have the permission for the pool mentioned in storage class

nehaberry:
@Madhu-1 thanks for the clarification. So, we actually don't need a separate adminID anymore since we can specify admin user in "userID" irtself. Makes sense.

Thanks again

Issue #380, opened 2019-05-18
Parameter `pool` in CephFS storageclass should be made optional


ShyamsundarR:
The CephFS storage class parameters mention that `pool` is a required parameter when `provisionVolume` parameter is `true` [1].

When a pool name is provided, its value is being used to set the following extended attributes on the subdirectory created for the CSI volume request in order to govern its layout [2],
- ceph.dir.layout.pool
and,
- ceph.dir.layout.pool_namespace

This parameter should be made optional, such that if not provided, the data goes to the default layouts (hence pools) as configured for the parent FS where the subvolume is created.

[1] storageclass reference: https://github.com/ceph/ceph-csi/blob/f60a07ae8268136f40de3d62537f4f8b020f0ec4/examples/cephfs/storageclass.yaml#L21

[2] CephFS layouts: http://docs.ceph.com/docs/nautilus/cephfs/file-layouts/

humblec:
@poornimag Is it something u can address with Ceph Mgr PR #400 ?

Issue #269, opened 2019-03-20
Provide an option to use a common Ceph credential across CephFS subvolumes


ShyamsundarR:
Currently, each CephFS volume gets a sub-directory under the CephFS root and further gets a user created with rw access to the directory (and a few other rights as required to access the directory). For each stage and publish request of such a sub-directory to a node, the CephFS root volume is mounted, and then the sub-directory is bind mounted into the requested directory, using the user credentials created for this volume.

In certain situations where the volume counts are high it is desirable to not have the additional per directory Ceph user, and instead leverage a single user that can mount the sub-directory, leaving kuberetes (or other container orchestration systems) to deal with access restrictions on bind mounted paths.

The above scheme also lends itself to staging the CephFS root once during staging the volume the first time on a node, and bind mounting sub-directories for each volume that recieves further publish requests on the same node. Again at scale, such a scheme helps, as it reduces the number of CephFS mounts on a node.

Thoughts on implementation:
- The option itself can be provided as a run-time option to the CSI plugin, thus reducing overhead on one more parameter to specify in the StorageClass and can also optionally be controlled by the StorageClass.

- The volume_options in CreateVolume response can carry the boolean regarding the nature of secret to use to mount the volume, and hence node level plugin can make a dynamic decision regarding the same.

gman0:
I had a similar idea in mind, but wanted to wait till we have cluster IDs in place - those could be used to simplify things down further.

ShyamsundarR:
:) totally agree, this can wait. I am just filing away stuff that is in my notes for consideration. Thanks!

ajarr:
@ShyamsundarR can we close this issue now?

ShyamsundarR:
Fixed, closing.

PR #400, opened 2019-06-04
Modify CephFs provisioner to use the ceph mgr commands

poornimag:
Currently CephFs provisioner mounts the ceph filesystem
and creates a subdirectory as a part of provisioning the
volume. Ceph now supports commands to provision fs subvolumes,
hance modify the provisioner to use ceph mgr commands to
(de)provision fs subvolumes.

### DIFF
--- a/pkg/cephfs/controllerserver.go
+++ b/pkg/cephfs/controllerserver.go
@@ -183,7 +183,7 @@ func (cs *ControllerServer) deleteVolumeDeprecated(req *csi.DeleteVolumeRequest)
 	idLk := volumeIDLocker.Lock(string(volID))
 	defer volumeIDLocker.Unlock(idLk, string(volID))
 
-	if err = purgeVolume(volID, cr, &ce.VolOptions); err != nil {
+	if err = purgeVolumeDeprecated(volID, cr, &ce.VolOptions); err != nil {
 		klog.Errorf("failed to delete volume %s: %v", volID, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
--- a/pkg/cephfs/volume.go
+++ b/pkg/cephfs/volume.go
@@ -17,6 +17,9 @@ limitations under the License.
 package cephfs
 
 import (
+	"fmt"
+	"os"
+	"path"
 	"strconv"
 	"strings"
 
@@ -36,6 +39,18 @@ var (
 	// its unnecessary
 	cephfsInit = false
 )
+
+func getCephRootVolumePathLocalDeprecated(volID volumeID) string {
+	return path.Join(getCephRootPathLocalDeprecated(volID), "csi-volumes", string(volID))
+}
+
+func getVolumeRootPathCephDeprecated(volID volumeID) string {
+	return path.Join("/", "csi-volumes", string(volID))
+}
+
+func getCephRootPathLocalDeprecated(volID volumeID) string {
+	return fmt.Sprintf("%s/controller/volumes/root-%s", PluginFolder, string(volID))
+}
 
 func getVolumeRootPathCeph(volOptions *volumeOptions, cr *util.Credentials, volID volumeID) (string, error) {
 	stdout, _, err := util.ExecCommand(
@@ -105,6 +120,70 @@ func createVolume(volOptions *volumeOptions, cr *util.Credentials, volID volumeI
 	if err != nil {
 		klog.Errorf("failed to create subvolume %s(%s) in fs %s", string(volID), err, volOptions.FsName)
 		return err
+	}
+
+	return nil
+}
+
+func mountCephRoot(volID volumeID, volOptions *volumeOptions, adminCr *util.Credentials) error {
+	cephRoot := getCephRootPathLocalDeprecated(volID)
+
+	// Root path is not set for dynamically provisioned volumes
+	// Access to cephfs's / is required
+	volOptions.RootPath = "/"
+
+	if err := createMountPoint(cephRoot); err != nil {
+		return err
+	}
+
+	m, err := newMounter(volOptions)
+	if err != nil {
+		return fmt.Errorf("failed to create mounter: %v", err)
+	}
+
+	if err = m.mount(cephRoot, adminCr, volOptions); err != nil {
+		return fmt.Errorf("error mounting ceph root: %v", err)
+	}
+
+	return nil
+}
+
+func unmountCephRoot(volID volumeID) {
+	cephRoot := getCephRootPathLocalDeprecated(volID)
+
+	if err := unmountVolume(cephRoot); err != nil {
+		klog.Errorf("failed to unmount %s with error %s", cephRoot, err)
+	} else {
+		if err := os.Remove(cephRoot); err != nil {
+			klog.Errorf("failed to remove %s with error %s", cephRoot, err)
+		}
+	}
+}
+
+func purgeVolumeDeprecated(volID volumeID, adminCr *util.Credentials, volOptions *volumeOptions) error {
+	if err := mountCephRoot(volID, volOptions, adminCr); err != nil {
+		return err
+	}
+	defer unmountCephRoot(volID)
+
+	var (
+		volRoot         = getCephRootVolumePathLocalDeprecated(volID)
+		volRootDeleting = volRoot + "-deleting"
+	)
+
+	if pathExists(volRoot) {
+		if err := os.Rename(volRoot, volRootDeleting); err != nil {
+			return fmt.Errorf("couldn't mark volume %s for deletion: %v", volID, err)
+		}
+	} else {
+		if !pathExists(volRootDeleting) {
+			klog.V(4).Infof("cephfs: volume %s not found, assuming it to be already deleted", volID)
+			return nil
+		}
+	}
+
+	if err := os.RemoveAll(volRootDeleting); err != nil {
+		return fmt.Errorf("failed to delete volume %s: %v", volID, err)
 	}
 
 	return nil
--- a/pkg/cephfs/volume.go
+++ b/pkg/cephfs/volume.go
@@ -17,9 +17,8 @@ limitations under the License.
 package cephfs
 
 import (
-	"fmt"
-	"os"
-	"path"
+	"strconv"
+	"strings"
 
 	"github.com/ceph/ceph-csi/pkg/util"
 
@@ -27,132 +26,109 @@ import (
 )
 
 const (
-	cephVolumesRoot = "csi-volumes"
-
-	namespacePrefix = "ns-"
+	namespacePrefix   = "fsvolumens_"
+	csiSubvolumeGroup = "csi"
 )
 
-func getCephRootPathLocal(volID volumeID) string {
-	return fmt.Sprintf("%s/controller/volumes/root-%s", PluginFolder, string(volID))
-}
+var (
+	// cephfsInit is used to create "csi" subvolume group for the first time the csi plugin loads.
+	// Subvolume group create gets called every time the plugin loads, though it doesn't result in error
+	// its unnecessary
+	cephfsInit = false
+)
 
-func getCephRootVolumePathLocal(volID volumeID) string {
-	return path.Join(getCephRootPathLocal(volID), cephVolumesRoot, string(volID))
-}
+func getVolumeRootPathCeph(volOptions *volumeOptions, cr *util.Credentials, volID volumeID) (string, error) {
+	stdout, _, err := util.ExecCommand(
+		"ceph",
+		"fs",
+		"subvolume",
+		"getpath",
+		volOptions.FsName,
+		string(volID),
+		"--group_name",
+		csiSubvolumeGroup,
+		"-m", volOptions.Monitors,
+		"-c", util.CephConfigPath,
+		"-n", cephEntityClientPrefix+cr.ID,
+		"--key="+cr.Key)
 
-func getVolumeRootPathCeph(volID volumeID) string {
-	return path.Join("/", cephVolumesRoot, string(volID))
+	if err != nil {
+		klog.Errorf("failed to get the rootpath for the vol %s(%s)", string(volID), err)
+		return "", err
+	}
+	return strings.TrimSuffix(string(stdout), "\n"), nil
 }
 
 func getVolumeNamespace(volID volumeID) string {
 	return namespacePrefix + string(volID)
 }
 
-func setVolumeAttribute(root, attrName, attrValue string) error {
-	return execCommandErr("setfattr", "-n", attrName, "-v", attrValue, root)
-}
-
-func createVolume(volOptions *volumeOptions, adminCr *util.Credentials, volID volumeID, bytesQuota int64) error {
-	if err := mountCephRoot(volID, volOptions, adminCr); err != nil {
-		return err
-	}
-	defer unmountCephRoot(volID)
-
-	var (
-		volRoot         = getCephRootVolumePathLocal(volID)
-		volRootCreating = volRoot + "-creating"
-	)
-
-	if pathExists(volRoot) {
-		klog.V(4).Infof("cephfs: volume %s already exists, skipping creation", volID)
-		return nil
-	}
-
-	if err := createMountPoint(volRootCreating); err != nil {
-		return err
-	}
-
-	if bytesQuota > 0 {
-		if err := setVolumeAttribute(volRootCreating, "ceph.quota.max_bytes", fmt.Sprintf("%d", bytesQuota)); err != nil {
+func createVolume(volOptions *volumeOptions, cr *util.Credentials, volID volumeID, bytesQuota int64) error {
+	//TODO: When we support multiple fs, need to hande subvolume group create for all fs's
+	if !cephfsInit {
+		err := execCommandErr(
+			"ceph",
+			"fs",
+			"subvolumegroup",
+			"create",
+			volOptions.FsName,
+			csiSubvolumeGroup,
+			"--mode",
+			"777",
+			"--pool_layout",
+			volOptions.Pool,
+			"-m", volOptions.Monitors,
+			"-c", util.CephConfigPath,
+			"-n", cephEntityClientPrefix+cr.ID,
+			"--key="+cr.Key)
+		if err != nil {
+			klog.Errorf("failed to create subvolume group csi, for the vol %s(%s)", string(volID), err)
 			return err
 		}
+		klog.V(4).Infof("cephfs: created subvolume group csi")
+		cephfsInit = true
 	}
-
-	if err := setVolumeAttribute(volRootCreating, "ceph.dir.layout.pool", volOptions.Pool); err != nil {
-		return fmt.Errorf("%v\ncephfs: Does pool '%s' exist?", err, volOptions.Pool)
-	}
-
-	if err := setVolumeAttribute(volRootCreating, "ceph.dir.layout.pool_namespace", getVolumeNamespace(volID)); err != nil {
-		return err
-	}
-
-	if err := os.Rename(volRootCreating, volRoot); err != nil {
-		return fmt.Errorf("couldn't mark volume %s as created: %v", volID, err)
-	}
-
-	return nil
-}
-
-func purgeVolume(volID volumeID, adminCr *util.Credentials, volOptions *volumeOptions) error {
-	if err := mountCephRoot(volID, volOptions, adminCr); err != nil {
-		return err
-	}
-	defer unmountCephRoot(volID)
-
-	var (
-		volRoot         = getCephRootVolumePathLocal(volID)
-		volRootDeleting = volRoot + "-deleting"
-	)
-
-	if pathExists(volRoot) {
-		if err := os.Rename(volRoot, volRootDeleting); err != nil {
-			return fmt.Errorf("couldn't mark volume %s for deletion: %v", volID, err)
-		}
-	} else {
-		if !pathExists(volRootDeleting) {
-			klog.V(4).Infof("cephfs: volume %s not found, assuming it to be already deleted", volID)
-			return nil
-		}
-	}
-
-	if err := os.RemoveAll(volRootDeleting); err != nil {
-		return fmt.Errorf("failed to delete volume %s: %v", volID, err)
-	}
-
-	return nil
-}
-
-func mountCephRoot(volID volumeID, volOptions *volumeOptions, adminCr *util.Credentials) error {
-	cephRoot := getCephRootPathLocal(volID)
-
-	// Root path is not set for dynamically provisioned volumes
-	// Access to cephfs's / is required
-	volOptions.RootPath = "/"
-
-	if err := createMountPoint(cephRoot); err != nil {
-		return err
-	}
-
-	m, err := newMounter(volOptions)
+	err := execCommandErr(
+		"ceph",
+		"fs",
+		"subvolume",
+		"create",
+		volOptions.FsName,
+		string(volID),
+		strconv.FormatInt(bytesQuota, 10),
+		"--group_name",
+		csiSubvolumeGroup,
+		"-m", volOptions.Monitors,
+		"-c", util.CephConfigPath,
+		"-n", cephEntityClientPrefix+cr.ID,
+		"--key="+cr.Key)
 	if err != nil {
-		return fmt.Errorf("failed to create mounter: %v", err)
-	}
-
-	if err = m.mount(cephRoot, adminCr, volOptions); err != nil {
-		return fmt.Errorf("error mounting ceph root: %v", err)
+		klog.Errorf("failed to create subvolume %s(%s) in fs %s", string(volID), err, volOptions.FsName)
+		return err
 	}
 
 	return nil
 }
 
-func unmountCephRoot(volID volumeID) {
-	cephRoot := getCephRootPathLocal(volID)
-
-	if err := unmountVolume(cephRoot); err != nil {
-		klog.Errorf("failed to unmount %s with error %s", cephRoot, err)
-	} else {
-		if err := os.Remove(cephRoot); err != nil {
-			klog.Errorf("failed to remove %s with error %s", cephRoot, err)
-		}
+func purgeVolume(volID volumeID, cr *util.Credentials, volOptions *volumeOptions) error {
+	err := execCommandErr(
+		"ceph",
+		"fs",
+		"subvolume",
+		"rm",
+		volOptions.FsName,
+		string(volID),
+		"--group_name",
+		csiSubvolumeGroup,
+		"--force",
+		"-m", volOptions.Monitors,
+		"-c", util.CephConfigPath,
+		"-n", cephEntityClientPrefix+cr.ID,
+		"--key="+cr.Key)
+	if err != nil {
+		klog.Errorf("failed to purge subvolume %s(%s) in fs %s", string(volID), err, volOptions.FsName)
+		return err
 	}
+
+	return nil
 }
--- a/pkg/cephfs/volumeoptions.go
+++ b/pkg/cephfs/volumeoptions.go
@@ -18,7 +18,6 @@ package cephfs
 
 import (
 	"fmt"
-	"path"
 	"strconv"
 
 	"github.com/pkg/errors"
@@ -272,7 +271,7 @@ func newVolumeOptionsFromVersion1Context(volID string, options, secrets map[stri
 			return nil, nil, err
 		}
 
-		opts.RootPath = path.Join("/csi-volumes", string(volumeID(volID)))
+		opts.RootPath = getVolumeRootPathCephDeprecated(volumeID(volID))
 	} else {
 		if err = extractOption(&opts.RootPath, "rootPath", options); err != nil {
 			return nil, nil, err
--- a/pkg/cephfs/volumeoptions.go
+++ b/pkg/cephfs/volumeoptions.go
@@ -18,6 +18,7 @@ package cephfs
 
 import (
 	"fmt"
+	"path"
 	"strconv"
 
 	"github.com/pkg/errors"
@@ -225,7 +226,11 @@ func newVolumeOptionsFromVolID(volID string, volOpt, secrets map[string]string) 
 		}
 	}
 
-	volOptions.RootPath = getVolumeRootPathCeph(volumeID(vid.FsSubvolName))
+	volOptions.RootPath, err = getVolumeRootPathCeph(&volOptions, cr, volumeID(vid.FsSubvolName))
+	if err != nil {
+		return nil, nil, err
+	}
+
 	volOptions.ProvisionVolume = true
 
 	return &volOptions, &vid, nil
@@ -267,7 +272,7 @@ func newVolumeOptionsFromVersion1Context(volID string, options, secrets map[stri
 			return nil, nil, err
 		}
 
-		opts.RootPath = getVolumeRootPathCeph(volumeID(volID))
+		opts.RootPath = path.Join("/csi-volumes", string(volumeID(volID)))
 	} else {
 		if err = extractOption(&opts.RootPath, "rootPath", options); err != nil {
 			return nil, nil, err
--- a/pkg/util/cephcmds.go
+++ b/pkg/util/cephcmds.go
@@ -31,9 +31,10 @@ import (
 // ExecCommand executes passed in program with args and returns separate stdout and stderr streams
 func ExecCommand(program string, args ...string) (stdout, stderr []byte, err error) {
 	var (
-		cmd       = exec.Command(program, args...) // nolint: gosec
-		stdoutBuf bytes.Buffer
-		stderrBuf bytes.Buffer
+		cmd           = exec.Command(program, args...) // nolint: gosec
+		sanitizedArgs = StripSecretInArgs(args)
+		stdoutBuf     bytes.Buffer
+		stderrBuf     bytes.Buffer
 	)
 
 	cmd.Stdout = &stdoutBuf
@@ -41,7 +42,7 @@ func ExecCommand(program string, args ...string) (stdout, stderr []byte, err err
 
 	if err := cmd.Run(); err != nil {
 		return stdoutBuf.Bytes(), stderrBuf.Bytes(), fmt.Errorf("an error (%v)"+
-			" occurred while running %s", err, program)
+			" occurred while running %s args: %v", err, program, sanitizedArgs)
 	}
 
 	return stdoutBuf.Bytes(), nil, nil
--- a/e2e/deploy-rook.go
+++ b/e2e/deploy-rook.go
@@ -69,8 +69,6 @@ func deployOperator(c kubernetes.Interface) {
 
 	_, err := framework.RunKubectl("create", "-f", opPath)
 	Expect(err).Should(BeNil())
-	err = waitForDaemonSets("rook-ceph-agent", rookNS, c, deployTimeout)
-	Expect(err).Should(BeNil())
 	err = waitForDaemonSets("rook-discover", rookNS, c, deployTimeout)
 	Expect(err).Should(BeNil())
 	err = waitForDeploymentComplete("rook-ceph-operator", rookNS, c, deployTimeout)
@@ -80,10 +78,12 @@ func deployOperator(c kubernetes.Interface) {
 func deployCluster(c kubernetes.Interface) {
 	opPath := fmt.Sprintf("%s/%s", rookURL, "cluster-test.yaml")
 	framework.RunKubectlOrDie("create", "-f", opPath)
+	err := waitForDaemonSets("rook-ceph-agent", rookNS, c, deployTimeout)
+	Expect(err).Should(BeNil())
 	opt := &metav1.ListOptions{
 		LabelSelector: "app=rook-ceph-mon",
 	}
-	err := checkCephPods(rookNS, c, 1, deployTimeout, opt)
+	err = checkCephPods(rookNS, c, 1, deployTimeout, opt)
 	Expect(err).Should(BeNil())
 }
 
--- a/pkg/cephfs/cephuser.go
+++ b/pkg/cephfs/cephuser.go
@@ -82,6 +82,10 @@ func getCephUser(volOptions *volumeOptions, adminCr *util.Credentials, volID vol
 
 func createCephUser(volOptions *volumeOptions, adminCr *util.Credentials, volID volumeID) (*cephEntity, error) {
 	adminID, userID := genUserIDs(adminCr, volID)
+	volRootPath, err := getVolumeRootPathCeph(volOptions, adminCr, volID)
+	if err != nil {
+		return nil, err
+	}
 
 	return getSingleCephEntity(
 		"-m", volOptions.Monitors,
@@ -91,7 +95,7 @@ func createCephUser(volOptions *volumeOptions, adminCr *util.Credentials, volID 
 		"-f", "json",
 		"auth", "get-or-create", userID,
 		// User capabilities
-		"mds", fmt.Sprintf("allow rw path=%s", getVolumeRootPathCeph(volID)),
+		"mds", fmt.Sprintf("allow rw path=%s", volRootPath),
 		"mon", "allow r",
 		"osd", fmt.Sprintf("allow rw pool=%s namespace=%s", volOptions.Pool, getVolumeNamespace(volID)),
 	)
--- a/pkg/cephfs/mountcache.go
+++ b/pkg/cephfs/mountcache.go
@@ -95,11 +95,16 @@ func mountOneCacheEntry(volOptions *volumeOptions, vid *volumeIdentifier, me *vo
 	volID := vid.VolumeID
 
 	if volOptions.ProvisionVolume {
-		volOptions.RootPath = getVolumeRootPathCeph(volumeID(vid.FsSubvolName))
 		cr, err = util.GetAdminCredentials(decodeCredentials(me.Secrets))
 		if err != nil {
 			return err
 		}
+
+		volOptions.RootPath, err = getVolumeRootPathCeph(volOptions, cr, volumeID(vid.FsSubvolName))
+		if err != nil {
+			return err
+		}
+
 		var entity *cephEntity
 		entity, err = getCephUser(volOptions, cr, volumeID(vid.FsSubvolName))
 		if err != nil {
--- a/pkg/cephfs/util.go
+++ b/pkg/cephfs/util.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"os"
 	"os/exec"
 
 	"google.golang.org/grpc/codes"
@@ -85,11 +84,6 @@ func isMountPoint(p string) (bool, error) {
 	}
 
 	return !notMnt, nil
-}
-
-func pathExists(p string) bool {
-	_, err := os.Stat(p)
-	return err == nil
 }
 
 // Controller service request validation
--- a/pkg/cephfs/util.go
+++ b/pkg/cephfs/util.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"os/exec"
 
 	"google.golang.org/grpc/codes"
@@ -84,6 +85,11 @@ func isMountPoint(p string) (bool, error) {
 	}
 
 	return !notMnt, nil
+}
+
+func pathExists(p string) bool {
+	_, err := os.Stat(p)
+	return err == nil
 }
 
 // Controller service request validation
--- a/scripts/travis-functest.sh
+++ b/scripts/travis-functest.sh
@@ -11,6 +11,6 @@ sudo scripts/minikube.sh k8s-sidecar
 sudo chown -R travis: "$HOME"/.minikube /usr/local/bin/kubectl
 # functional tests
 
-go test github.com/ceph/ceph-csi/e2e --rook-version=v1.0.1 --deploy-rook=true --deploy-timeout=10 -timeout=30m -v
+go test github.com/ceph/ceph-csi/e2e --rook-version=master --deploy-rook=true --deploy-timeout=10 -timeout=30m -v
 
 sudo scripts/minikube.sh clean


humblec:
Thanks @poornimag . We ( + @Madhu-1 ) are reviewing this change.

humblec:
@poornimag please check at the build failure.

humblec:
@ajarr PTAL.

examples/cephfs/storageclass.yaml
```diff
@@ -22,6 +22,9 @@ parameters:
   # Required for provisionVolume: "true"
   pool: cephfs_data
 
+  # Name of the ceph filesystem with in which the volumes will be created
```
humblec: with in/within ?

pkg/cephfs/volume.go
```diff
@@ -46,111 +35,27 @@ func getVolumeNamespace(volID volumeID) string {
...
-	m, err := newMounter(volOptions)
-	if err != nil {
-		return fmt.Errorf("failed to create mounter: %v", err)
-	}
-
-	if err = m.mount(cephRoot, adminCr, volOptions); err != nil {
-		return fmt.Errorf("error mounting ceph root: %v", err)
-	}
+func createVolume(volOptions *volumeOptions, volID volumeID, bytesQuota int64) error {
+	_, _, err := util.ExecCommand(
```
Madhu-1: @poornimag do we need any changes in  Node plugin for this?

Madhu-1:
Fixes: #328

Madhu-1:
@gman0 PTAL

humblec:
@poornimag PTAL

pkg/cephfs/volume.go
poornimag: Nope, we do not need

humblec:
@poornimag do we need any changes in the readme/documentation ?

poornimag:
Succesfully tested the PVC creation and mounting:

[root@Centos1 ceph-csi]# docker images
REPOSITORY                                 TAG                                   IMAGE ID            CREATED             SIZE
quay.io/cephcsi/cephcsi                    canary                                fc60f8587fb3        2 hours ago         926MB
rook/ceph                                  master                                c42f5954684b        11 hours ago        921MB
ajarr/ceph                                 v14.2.1-20190430-and-mgr-subvolumes   fdd956f8d15d        40 hours ago        850MB
k8s.gcr.io/kube-proxy                      v1.14.3                               004666307c5b        2 days ago          82.1MB
k8s.gcr.io/kube-controller-manager         v1.14.3                               ac2ce44462bc        2 days ago          158MB
k8s.gcr.io/kube-apiserver                  v1.14.3                               9946f563237c        2 days ago          210MB
k8s.gcr.io/kube-scheduler                  v1.14.3                               953364a3ae7a        2 days ago          81.6MB
nginx                                      latest                                62c261073ecf        3 days ago          109MB
quay.io/cephcsi/cephfsplugin               v1.0.0                                06ae0938b3a2        11 days ago         924MB
ceph/ceph                                  v14.2                                 c3c0d6aa89f0        5 weeks ago         850MB
k8s.gcr.io/kube-addon-manager              v9.0                                  119701e77cbc        4 months ago        83.1MB
k8s.gcr.io/coredns                         1.3.1                                 eb516548c180        4 months ago        40.3MB
quay.io/k8scsi/csi-node-driver-registrar   v1.0.2                                8fed7d1b1015        5 months ago        20.1MB
quay.io/k8scsi/csi-provisioner             v1.0.1                                2004b031bce2        6 months ago        48MB
quay.io/k8scsi/csi-attacher                v1.0.1                                2ef5329b7139        6 months ago        50.2MB
k8s.gcr.io/etcd                            3.3.10                                2c4adeb21b4f        6 months ago        258MB
k8s.gcr.io/pause                           3.1                                   da86e6ba6ca1        17 months ago       742kB
gcr.io/k8s-minikube/storage-provisioner    v1.8.1                                4689081edb10        19 months ago       80.8MB
[root@Centos1 ceph-csi]# 
[root@Centos1 ceph-csi]# 
[root@Centos1 ceph-csi]# kubectl get pvc
NAME         STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
cephfs-pvc   Bound    pvc-66d0584c-89b8-11e9-800b-52540091bbf9   96Mi       RWO            csi-cephfs     57m
[root@Centos1 ceph-csi]# kubectl get pods
NAME                 READY   STATUS    RESTARTS   AGE
csicephfs-demo-pod   1/1     Running   0          53m
[root@Centos1 ceph-csi]# 
[root@Centos1 ceph-csi]# kubectl describe pod csicephfs-demo-pod
Name:               csicephfs-demo-pod
Namespace:          default
Priority:           0
PriorityClassName:  <none>
Node:               minikube/192.168.121.48
Start Time:         Sat, 08 Jun 2019 06:45:26 +0000
Labels:             <none>
Annotations:        kubectl.kubernetes.io/last-applied-configuration:
                      {"apiVersion":"v1","kind":"Pod","metadata":{"annotations":{},"name":"csicephfs-demo-pod","namespace":"default"},"spec":{"containers":[{"im...
Status:             Running
IP:                 172.17.0.12
Containers:
  web-server:
    Container ID:   docker://7d51f01e6203c98890ab721d5584945794ef59882afd7f434279882c37f944f3
    Image:          nginx
    Image ID:       docker-pullable://nginx@sha256:12db363acf5b2d2f9f5fed240e228a04692bdac68402430fbd2f720c3a967d01
    Port:           <none>
    Host Port:      <none>
    State:          Running
      Started:      Sat, 08 Jun 2019 06:45:53 +0000
    Ready:          True
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /var/lib/www/html from mypvc (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-s4wp6 (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             True 
  ContainersReady   True 
  PodScheduled      True 
Volumes:
  mypvc:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  cephfs-pvc
    ReadOnly:   false
  default-token-s4wp6:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-s4wp6
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  <none>
Tolerations:     node.kubernetes.io/not-ready:NoExecute for 300s
                 node.kubernetes.io/unreachable:NoExecute for 300s
Events:
  Type    Reason                  Age   From                     Message
  ----    ------                  ----  ----                     -------
  Normal  Scheduled               53m   default-scheduler        Successfully assigned default/csicephfs-demo-pod to minikube
  Normal  SuccessfulAttachVolume  53m   attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-66d0584c-89b8-11e9-800b-52540091bbf9"
  Normal  Pulling                 53m   kubelet, minikube        Pulling image "nginx"
  Normal  Pulled                  52m   kubelet, minikube        Successfully pulled image "nginx"
  Normal  Created                 52m   kubelet, minikube        Created container web-server
  Normal  Started                 52m   kubelet, minikube        Started container web-server
[root@Centos1 ceph-csi]# 




[root@Centos1 ceph-csi]# kubectl exec -it rook-ceph-tools-b8c679f95-56mbl -nrook-ceph /bin/bash
[root@Centos1 /]# ll /mnt/myfs/csi-volumes/                
total 1
drwxr-x---. 2 root root 0 Jun  8 06:33 csi-cephfs-pvc-6386ec62-89b7-11e9-800b-52540091bbf9-creating
drwxr-x---. 2 root root 0 Jun  8 06:46 csi-cephfs-pvc-66d0584c-89b8-11e9-800b-52540091bbf9
[root@Centos1 /]# ll /mnt/myfs/csi-volumes/csi-cephfs-pvc-66d0584c-89b8-11e9-800b-52540091bbf9
total 0
-rw-r--r--. 1 root root 0 Jun  8 06:46 abc
[root@Centos1 /]# 
[root@Centos1 ceph]# kubectl exec -it csicephfs-demo-pod /bin/bash
root@csicephfs-demo-pod:/# 
root@csicephfs-demo-pod:/# ls /var/lib/www/html
abc

poornimag:
> @poornimag do we need any changes in the readme/documentation ?

No, the fsName parameter is already added in a different path, so this patch has not added any new parameters

humblec:
@poornimag Awesome! Can you also please test or write some data in this volume from the pod and check its writable ?

poornimag:
> @poornimag Awesome! Can you also please test or write some data in this volume from the pod and check its writable ?

Yes, have created a file abc in the ngnix test pod, and i see that its created from a different mount point (in the tool box). The log i have pasted also contains this...

poornimag:
However, during the testing came across an issue:
- There was a stale subvolume that was created csi-cephfs-pvc-6386ec62-89b7-11e9-800b-52540091bbf9-creating. There were some attempts of pv creation that failed, may be one of the failure code paths didn't cleanup the temp pvc
Will retest and raise an issue.

humblec:
>There was a stale subvolume that was created csi-cephfs-pvc-6386ec62-89b7-11e9-800b-52540091bbf9-creating. There were some attempts of pv creation that failed, may be one of the failure code paths didn't cleanup the temp pvc
Will retest and raise an issue.
>
Do we know what the error or cause for the failure of PV creation?  that would become handy while raising the issue too. Yeah, lets log it if there is such condition.

pkg/cephfs/volume.go
```diff
@@ -17,27 +17,17 @@ limitations under the License.
...
 
-	"k8s.io/klog"
+	"github.com/ceph/ceph-csi/pkg/util"
 )
 
 const (
 	cephVolumesRoot = "csi-volumes"
 
-	namespacePrefix = "ns-"
+	namespacePrefix = "fsvolumens-"
```
humblec: @poornimag any specific reason for namespace prefix change here ?

humblec:
The CI is failed and I believe its expected to fail as the ceph cluster in travis is not the version of nautilus we used for testing or rook templates dont have these latest images. The error what I can see there is:
>Jun  8 08:05:39.755: INFO: At 2019-06-08 07:55:49 +0000 UTC - event for csi-cephfs-pvc: {cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-0_cd9cb849-89c2-11e9-864c-0242ac11000b } ProvisioningFailed: failed to provision volume with StorageClass "csi-cephfs-sc": rpc error: code = Internal desc = an error (exit status 1) occurred while running ceph
>

I believe we could list better error to identify the exact reason though.

pkg/cephfs/volume.go
poornimag: Its changed to the default value specified in the ceph subvol create code

humblec: Thanks for clarifying

poornimag:
> The CI is failed and I believe its expected to fail as the ceph cluster in travis is not the version of nautilus we used for testing or rook templates dont have these latest images. The error what I can see there is:
> 
> > Jun  8 08:05:39.755: INFO: At 2019-06-08 07:55:49 +0000 UTC - event for csi-cephfs-pvc: {cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-0_cd9cb849-89c2-11e9-864c-0242ac11000b } ProvisioningFailed: failed to provision volume with StorageClass "csi-cephfs-sc": rpc error: code = Internal desc = an error (exit status 1) occurred while running ceph
> 
> I believe we could list better error to identify the exact reason though.

Yes, this is the exact error i got without Ramana's fix for quota in subvolumes. Then we wait for his patch to be merged and csi ci to succeed?
The error message is what is returned by the ceph subvolume command execution. Will see how to change

poornimag:
> Do we know what is the error logged at time of failed PV creation? that would become handy while raising the issue too. Yeah, lets log it if there is such condition.

The error was due to, not providing the monitor ip in the storage class.

humblec:
@poornimag do we need createBackingVolume() func now ?  or in short, do we need getAdminCredentials() and createCephUser() funcs from it?

pkg/cephfs/volume.go
humblec: @poornimag but Isnt it `fsvolumens_`

poornimag:
> @poornimag do we need createBackingVolume() func now ? or in short, do we need getAdminCredentials() and createCephUser() funcs from it?

I did try to remove that, but createCephUser wasn't necessary even without this patch. Its the same as issue https://github.com/ceph/ceph-csi/issues/269 . Will send it out as a separate patch.

pkg/cephfs/volume.go
```diff
@@ -46,111 +36,27 @@ func getVolumeNamespace(volID volumeID) string {
...
-	}
+func createVolume(volOptions *volumeOptions, volID volumeID, bytesQuota int64) error {
+	_, _, err := util.ExecCommand(
+		"ceph",
+		"fs",
+		"subvolume",
+		"create",
+		volOptions.FsName,
+		string(volID),
+		string(bytesQuota))
```
humblec: @poornimag To allow non root write to the directory we could make use of `mode` arg in this call and set mode 777 and root:root instead of 755

pkg/cephfs/volume.go
ajarr: Yes. This needs to 'fsvolumens_'

pkg/cephfs/volume.go
ajarr: @humblec, there is no 'mode' args in the `ceph fs subvolume create` command. If that's what you meant.

pkg/cephfs/volume.go
```diff
@@ -46,111 +36,27 @@ func getVolumeNamespace(volID volumeID) string {
...
-func unmountCephRoot(volID volumeID) {
-	cephRoot := getCephRootPathLocal(volID)
+func purgeVolume(volID volumeID, volOptions *volumeOptions) error {
+	_, _, err := util.ExecCommand(
+		"ceph",
+		"fs",
+		"subvolume",
+		"rm",
+		volOptions.FsName,
+		string(volID))
```
ajarr: How do you handle the case when trying to purge_subvolume that is already removed as was done earlier?  This command will raise ENOENT if the subvolume doesn't exist. To suppress ENOENT error, you can add do,
`ceph fs subvolume rm --force`

pkg/cephfs/volume.go
humblec: @ajarr Oh.. I noticed this in one of the PR:
>
def create_subvolume(self, subvolume_path, size=None, namespace_isolated=True, mode=0o755):
>

which list `mode` arg.

ajarr:
The `ceph fs subvolume create` command creates a cephfs subdir with the path /volumes/_nogroup/<csi-vol-id> that is served as the csi volume. I prefer that that the subvolume be created within the subvolume group 'csi' to differentiate between OpenStack manila and CSI use cases, which use the same `fs subvolume` interface.
I suggest,
`ceph fs subvolumegroup create <vol name> csi`
and then,
`ceph fs subvolume create <vol name> <subvolname> --group_name csi`

You don't have to create the subvolume group for every volume create, just some time before you start creating volumes.

@batrick, what do you think?

pkg/cephfs/volume.go
ajarr: That's the method that's called by the command `fs subvolume create`,
https://github.com/ceph/ceph/blob/master/src/pybind/mgr/volumes/module.py#L353
The 'mode' arg is not exposed to the command line user. Is this required for CSI?

humblec:
> The `ceph fs subvolume create` command creates a cephfs subdir with the path,`/volumes/_nogroup/<csi-vol-id>`, which is served as the csi volume. I prefer that that the subvolume be created within the subvolume group 'csi' to differentiate between OpenStack manila and CSI use cases, which use the same `fs subvolume` interface.
>
That basically restrict the volume group of `all csi subvolumes` to `csi` group , Isnt it ? or are you saying we can create different groups with `prefix` called `csi` in a fsnamespace ? I may be missing the internals, if that the case please feel free to correct :)

pkg/cephfs/volume.go
humblec: @ajarr thats required. Because the app pod can be run as any user and this user should have access to this volume by exposing the mode as mentioned here https://github.com/ceph/ceph-csi/pull/400#discussion_r291903118. If not, the pod user will get `Permission denied` and the volume is not usable.

ajarr:
> > The `ceph fs subvolume create` command creates a cephfs subdir with the path,`/volumes/_nogroup/<csi-vol-id>`, which is served as the csi volume. I prefer that that the subvolume be created within the subvolume group 'csi' to differentiate between OpenStack manila and CSI use cases, which use the same `fs subvolume` interface.
> 
> That basically restrict the volume group of `all csi subvolumes` to `csi` group , Isnt it ? or are you saying we can create different groups with `prefix` called `csi` in a fsnamespace ? I may be missing the internals, if that the case please feel free to correct :)

If you create subvolumes within the `csi` subvolume group, then the subvolumes will have the path,
`/volumes/csi/<csi-vol-id>`. We can then create auth IDs that have restricted access to the 'csi` subvolume group and all its subvolumes, which would be useful for  issue #269.

pkg/cephfs/volume.go
ajarr: How was  @poornimag able to write files to the pod? https://github.com/ceph/ceph-csi/pull/400#issuecomment-500104579  

@poornimag, ?

OK. We can allow the perms of the directory to be changed. You wanted the owner:group to be changed to root:root. What's it right now?

pkg/cephfs/volume.go
```diff
@@ -17,27 +17,17 @@ limitations under the License.
...
-	"fmt"
-	"os"
 	"path"
 
-	"k8s.io/klog"
+	"github.com/ceph/ceph-csi/pkg/util"
 )
 
 const (
 	cephVolumesRoot = "csi-volumes"
```
ajarr: Doesn't this need to be changed?

batrick:
> The `ceph fs subvolume create` command creates a cephfs subdir with the path,`/volumes/_nogroup/<csi-vol-id>`, which is served as the csi volume. I prefer that that the subvolume be created within the subvolume group 'csi' to differentiate between OpenStack manila and CSI use cases, which use the same `fs subvolume` interface.
> I suggest,
> `ceph fs subvolumegroup create <vol name> csi`
> and then,
> `ceph fs subvolume create <vol name> <subvolname> --group_name csi`
> This would create subvolumes with paths, `/volume/csi/<csi-vol-id>`
> 
> You don't have to create the subvolume group for every volume create, just some time before you start creating volumes.
> 
> @batrick, what do you think?

I like it!

pkg/cephfs/volume.go
humblec: @ajarr In above example, the write happened as 'root' user, thats why she was able to write to it. However as mentioned in earlier comment, pod can be run as any user, so we should allow the write to the volume. We could work around it with the `mode` or setting `777`. iic, `owner:group` is (already) `root:root` now. We only need change in `mode`. @poornimag can confirm.

Madhu-1:
@poornimag can you test  this with normal user use this [pod](https://github.com/gluster/gluster-csi-driver/pull/142#issuecomment-454652309) template

pkg/cephfs/volume.go
Madhu-1: https://github.com/ceph/ceph-csi/pull/400#issuecomment-500677369

humblec:
> @poornimag can you test this with normal user use this [pod](https://github.com/gluster/gluster-csi-driver/pull/142#issuecomment-454652309) template

@Madhu-1 iic, its not going to work  as mentioned https://github.com/ceph/ceph-csi/pull/400#discussion_r292248444

pkg/cephfs/volume.go
Madhu-1: @humblec setting `0777` on `targetPath` will help
@poornimag  can you try once?

create targetPath with `0777` and once the mounting is done and also change the permission of targetPath to `0777` at last

poornimag: I tested with the pod template you mentioned, but it didn't run. Hence built my own. The volume is not accessible by a non root user:
```
/ $ ls /target/
ls: can't open '/target/': Permission denied
/ $ 
```
If we set the targetpath chmod to 777 during mount in the nodeserver, it will work, but thats doing it everytime we mount, or are you suggesting during the create volume? @Madhu-1 

```[root@Centos1 alpine-nonroot]# cat Dockerfile 
FROM alpine:latest

# Create a group and user
RUN addgroup -g 1015 testgroup5 && adduser -S testuser5 -u 1015 -g 1015
#
# # Tell docker that all future commands should run as the appuser user
USER 1015


[root@Centos1 ceph]# cat test-non-root-user.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: pod-run-as-non-root
spec:
  containers:
  - name: main
    image: alpine-non-root:latest
    command: ["/bin/sleep", "999999"]
    securityContext:
       runAsNonRoot: true
    imagePullPolicy: "IfNotPresent"
    volumeMounts:
     - name: target
       mountPath: /target
  volumes:
   - name: target
     persistentVolumeClaim:
       claimName: fs-perf-target

```

humblec: @poornimag @Madhu-1 as mentioned earlier (https://github.com/ceph/ceph-csi/pull/400#discussion_r291903118) its better to do it at create time and it should be sufficient. Its just that , the `mode` has to be exposed in the createVolume() func as mentioned earlier and iic, @ajarr is working on this or will be on it. We will use it once available. For now, as a workaround we can set it on targetPath.

Madhu-1: @poornimag changing permission after the mount is good.

pkg/cephfs/volume.go
ShyamsundarR: +1

But, some further comments,
- @poornimag For 1.0 volumes the volume root would need to be retained as "csi-volumes", as these are not created by the `ceph fs subvolume ...` commands
- For post 1.0 volumes, the volumes root **maybe** `/volume/<subvolume-group>/<csi-vol-id>`
  - can this path be assumed as the volume root path? @ajarr 
  - There was a prior discussion that we would be able to fetch this root path from one of the subcommands, to aid in keeping the interface abstraction in place (during node plugin actions). IOW, do not leak the internals of exactly where the subvolumes are created, but pass it back when asked for. I would prefer that, to hard coding it in the code here, so that any changes can be easily accommodated. @ajarr ?

pkg/cephfs/controllerserver.go
```diff
@@ -184,7 +184,7 @@ func (cs *ControllerServer) deleteVolumeDeprecated(req *csi.DeleteVolumeRequest)
 	mtxControllerVolumeID.LockKey(string(volID))
 	defer mustUnlock(mtxControllerVolumeID, string(volID))
 
-	if err = purgeVolume(volID, cr, &ce.VolOptions); err != nil {
+	if err = purgeVolume(volID, &ce.VolOptions); err != nil {
```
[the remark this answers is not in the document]
ajarr: > * Where will the `subvolume rm` command look for the passed in <CSI volID/name>? @ajarr
>   "Where" as in which path, I would assume based on [this](https://github.com/ceph/ceph-csi/pull/400#issuecomment-500358206) comment this would be, `/volumes/_nogroup/<csi-vol-id>` (and possibly in the future `/volume/csi/<csi-vol-id>`)

Correct.

> * 1.0 created volumes would not be in the said location, unless there is more smartness built into the `ceph fs subvolume ...` commands implementation (which I do not expect). @ajarr
> 

No, the `ceph fs subvolume` doesn't remove the earlier version of subvolume that had a different path.

pkg/cephfs/volume.go
```diff
@@ -46,111 +36,27 @@ func getVolumeNamespace(volID volumeID) string {
...
-	m, err := newMounter(volOptions)
-	if err != nil {
-		return fmt.Errorf("failed to create mounter: %v", err)
-	}
-
-	if err = m.mount(cephRoot, adminCr, volOptions); err != nil {
-		return fmt.Errorf("error mounting ceph root: %v", err)
-	}
+func createVolume(volOptions *volumeOptions, volID volumeID, bytesQuota int64) error {
+	_, _, err := util.ExecCommand(
```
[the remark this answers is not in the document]
ajarr: The provisioner pod/container needs to have the monitor list, and auth ID (with key) with enough privileges to talk to monitor.

pkg/cephfs/volume.go
```diff
@@ -46,111 +36,27 @@ func getVolumeNamespace(volID volumeID) string {
...
-		return err
-	}
-
-	if bytesQuota > 0 {
-		if err := setVolumeAttribute(volRootCreating, "ceph.quota.max_bytes", fmt.Sprintf("%d", bytesQuota)); err != nil {
-			return err
-		}
-	}
-
-	if err := setVolumeAttribute(volRootCreating, "ceph.dir.layout.pool", volOptions.Pool); err != nil {
```
ShyamsundarR: Did we just loose the option to set the data pool for this subvolume? I do not see the `Pool` parameter being passed to create, and neither do we have code here to set the volume attributes, so is this something that is missing and will be handled by the `subvolume` command, or should be done in the CSI code?

pkg/cephfs/volume.go
ShyamsundarR: Ideally, once user creation moves into the `ceph fs subvolume create ...` command I assume CSI code will not deal with this namespace name.

@ajarr Supporting user creation (based on some flag) is part of the implementation plan in the future?

pkg/cephfs/volume.go
ajarr: So you can fetch the absolute path of the subvolume,
`ceph fs subvolume getpath <vol name/fs name> <subvol name/csi-vol-id> --group_name csi` 
The command would return, `/volumes/csi/<csi-vol-id>`.

@ShyamsundarR , does that answer your question?

pkg/cephfs/volume.go
ajarr: Yes, subvolume create command doesn't give the option to set the data pool layout. The default data pool of the volume/filesystem is picked. Do we really need the 'data pool' to be configurable?

pkg/cephfs/volume.go
ShyamsundarR: Yes, thanks. This helps in the nodeplugin to not have to assume the path, and change/use something like `cephVolumesRoot` variable.

@poornimag this needs to be incorporated in the publish sequence to fetch the path rather than construct it.

pkg/cephfs/controllerserver.go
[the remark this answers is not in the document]
ShyamsundarR: > > * 1.0 created volumes would not be in the said location, unless there is more smartness built into the `ceph fs subvolume ...` commands implementation (which I do not expect). @ajarr
> 
> No, the `ceph fs subvolume` doesn't remove the earlier version of subvolume that had a different path.

@poornimag we would need to invoke the older purge routines to delete and purge volumes created using v1.0 and earlier. Current code would break the backward compatibility otherwise.

pkg/cephfs/volume.go
ShyamsundarR: So this is an interesting question, tagging @gman0 as well for his thoughts/requirements.

Currently, we want to make the `pool` specification **optional** (as noted in #380), but not entirely remove it.

For use cases where data needs to reside in a different pool or different type of pool, I would assume this is a useful option to retain. Would this be a correct assumption?

Possibly when multiple FS volumes are supported in the future, we can remove this option, in lieu of creating FS volumes that have differing data pool properties?

pkg/cephfs/volume.go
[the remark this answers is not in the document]
ShyamsundarR: The places where CSI requires credential information is passed along from the container orchestration systems (kunerbetes say) as required and the config maps carry the monitor information. So, we have all the information to make the call.

The current pattern to invoke any CEPH CLI in the CSI code base is to pass the, monitors, id, and key, as part of the CLI invocation. I would prefer to see the same applied here, than use other schemes at the moment.

There is also the issue #318 where we do not want the credentials as part of the CLI, and when we arrive at a common solution for the same, we would address it for these invocations as well.

@poornimag assuming there are CLI options available to pass the id, key and monitor list here, request that the command invocation be changed to use the same.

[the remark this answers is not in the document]
ajarr: @ShyamsundarR  We (@poornimag and I) passed the monitor address, cephx auth ID and key, as part of the `ceph fs subvolume command` and it worked. We just have to keep in mind that cephx auth ID needs ceph-mgr 'rw'  capabilities.

e.g,
```
$ ./bin/ceph auth get client.test
exported keyring for client.test
[client.test]
	key = AQDM1Ahdx1XNDBAA2ZJ9/90kEWoke1WOQlw8HA==
	caps mgr = "allow rw"
	caps mon = "allow r"

$ ./bin/ceph fs subvolume create a ram -m 10.76.98.58:40905 --name client.test --key=AQDM1Ahdx1XNDBAA2ZJ9/90kEWoke1WOQlw8HA==
```

pkg/cephfs/volume.go
ajarr: @batrick ^^

batrick: Retaining the option to set a data pool seems useful. It also seems fairly trivial to support. Is there a reason we didn't port that over @ajarr?

ajarr: I missed implementing this,
http://tracker.ceph.com/issues/40431

I'm on it.

pkg/cephfs/volume.go
poornimag: Done

pkg/cephfs/volume.go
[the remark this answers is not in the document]
poornimag: Done

pkg/cephfs/controllerserver.go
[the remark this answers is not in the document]
poornimag: @Madhu-1 @humblec Suggested to not have backward compatibility in the code. IMO, It will be easier to have backward compatibility than migrating old volumes to new paths, as i m not sure what happens to the attributes set, user permission also needs to be modified in that case. Thoughts?
Also Is it alright if its a different PR?

pkg/cephfs/volume.go
poornimag: Will add it once the fs command supports the pool arg.

Madhu-1:
@poornimag let's have a precheck for ceph cluster version if the version is less than 1.14.2 will return precondition error

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
-		return nil
-	}
-
-	if err := createMountPoint(volRootCreating); err != nil {
-		return err
-	}
-
-	if bytesQuota > 0 {
-		if err := setVolumeAttribute(volRootCreating, "ceph.quota.max_bytes", fmt.Sprintf("%d", bytesQuota)); err != nil {
+func createVolume(volOptions *volumeOptions, cr *credentials, volID volumeID, bytesQuota int64) error {
```
ajarr: where do you set quota size?

pkg/cephfs/controllerserver.go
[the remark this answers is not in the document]
humblec: @poornimag considering this is a breaking change which is inevitable, the backward compatibility can be taken care later.  We could think about a documentation or different PR if at all we have a  mechanism for backward compatibility.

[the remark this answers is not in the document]
ShyamsundarR: As long as we ensure we release v1.1.0 with the backward compatibility I am fine it being a different PR (this is the same I did with stateless+RBD and the compat PR is being tracked for v1.1.0 #391 )

humblec:
@poornimag can you please list whats pending in this PR ? so that we can track it on followup PRs ?

poornimag:
> @poornimag can you please list whats pending in this PR ? so that we can track it on followup PRs ?

The pending thing is to have backward compatibility. Can it be in different pr?

Couple of other things:
Add pool parameter, only after that gets merged in ceph-mgr. This will be a separate pr.
Add chmod for the create volume, only after that gets merged in ceph-mgr. This will be a separate pr as well.

ShyamsundarR:
> > @poornimag can you please list whats pending in this PR ? so that we can track it on followup PRs ?
> 
> The pending thing is to have backward compatibility. Can it be in different pr?

As commented [here](https://github.com/ceph/ceph-csi/pull/400#discussion_r295800369) I am fine with it being a separate PR, if handled as part of the 1.1.0 release.

> 
> Couple of other things:
> Add pool parameter, only after that gets merged in ceph-mgr. This will be a separate pr.

This should be fine, as volumes created prior to the change and post the change will have the same functionality if this is used.

> Add chmod for the create volume, only after that gets merged in ceph-mgr. This will be a separate pr as well.

Again if done before 1.1.0 there is no compatibility problem I foresee.

If done after, does this remove the existing chmod change in the node plugin, and if so, how will we know when the volume was created (1.1.0 or 1.1.1) and hence conditionally execute the chmod code?

humblec:
@poornimag oornimag it looks like we have natulius upstream container to support this here https://hub.docker.com/r/ceph/daemon-base/tags/ , ie `latest-nautlius-devel` tagged container .

poornimag:
> @poornimag oornimag it looks like we have natulius upstream container to support this here https://hub.docker.com/r/ceph/daemon-base/tags/ , ie `latest-nautlius-devel` tagged container .

Will send the PR for rook, if Madhu has already not sent it. Once that is merged, e2e should succeed for the patch. And then, if no more review can merge it?

humblec:
>Will send the PR for rook, if Madhu has already not sent it. Once that is merged, e2e should succeed for the patch. And then, if no more review can merge it?
>

@poornimag please feel free to send a PR to Rook, there is no PR for above. We need to rebase !!!

Madhu-1:
> has already not sent it. Once that is merged, e2e should succeed for the patch. And then, if no more review can merge it?

I haven't sent one, please send it

poornimag:
> > has already not sent it. Once that is merged, e2e should succeed for the patch. And then, if no more review can merge it?
> 
> I haven't sent one, please send it

Done. https://github.com/rook/rook/pull/3377

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
-	namespacePrefix = "ns-"
+var (
+	cephfsInit = false
 )
 
-func getCephRootPathLocal(volID volumeID) string {
-	return fmt.Sprintf("%s/controller/volumes/root-%s", PluginFolder, string(volID))
-}
+func getVolumeRootPathCeph(volOptions *volumeOptions, cr *credentials, volID volumeID) string {
+	stdout, _, err := util.ExecCommand(
```
Madhu-1: use `execCommandErr` function if you want to discard stdErr function

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
 )
 
 const (
-	cephVolumesRoot = "csi-volumes"
+	namespacePrefix = "fsvolumens_"
+)
 
-	namespacePrefix = "ns-"
+var (
+	cephfsInit = false
```
Madhu-1: please  add a comment for this one

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
-
-	if err := createMountPoint(volRootCreating); err != nil {
-		return err
-	}
-
-	if bytesQuota > 0 {
-		if err := setVolumeAttribute(volRootCreating, "ceph.quota.max_bytes", fmt.Sprintf("%d", bytesQuota)); err != nil {
+func createVolume(volOptions *volumeOptions, cr *credentials, volID volumeID, bytesQuota int64) error {
+	if !cephfsInit {
+		_, _, err := util.ExecCommand(
```
Madhu-1: if the subvolume group already exists will it pass or return an error?

pkg/cephfs/volume.go
Madhu-1: use this wherever its applicable

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
+		string(volID),
+		strconv.FormatInt(bytesQuota, 10),
+		"--group_name",
+		"csi",
+		"-m", volOptions.Monitors,
+		"-c", util.CephConfigPath,
+		"-n", cephEntityClientPrefix+cr.id,
+		"--key="+cr.key)
+	if err != nil {
+		klog.Errorf("failed to create subvolume %s(%s) in fs %s", string(volID), err, volOptions.FsName)
```
Madhu-1: add space  after volume  and error message in log

pkg/cephfs/volumeoptions.go
```diff
@@ -268,7 +269,7 @@ func newVolumeOptionsFromVersion1Context(volID string, options, secrets map[stri
 			return nil, nil, err
 		}
 
-		opts.RootPath = getVolumeRootPathCeph(volumeID(volID))
+		opts.RootPath = path.Join("/csi-volumes", string(volumeID(volID)))
```
Madhu-1: is this path configurable or hardcoded?

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
-	return nil
-}
-
-func purgeVolume(volID volumeID, adminCr *credentials, volOptions *volumeOptions) error {
-	if err := mountCephRoot(volID, volOptions, adminCr); err != nil {
+	_, _, err := util.ExecCommand(
+		"ceph",
+		"fs",
+		"subvolume",
+		"create",
```
ShyamsundarR: `createCephUser` allows the user created for this subvolume RW to the namespace. In this command we are not passing the namespace, so is create setting the `ceph.dir.layout.pool_namespace` yet internally or not? @ajarr ?

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
 
-func getCephRootPathLocal(volID volumeID) string {
-	return fmt.Sprintf("%s/controller/volumes/root-%s", PluginFolder, string(volID))
-}
+func getVolumeRootPathCeph(volOptions *volumeOptions, cr *credentials, volID volumeID) string {
+	stdout, _, err := util.ExecCommand(
+		"ceph",
+		"fs",
+		"subvolume",
+		"getpath",
```
ShyamsundarR: I was expecting a `getbasepath` command, rather than a `getpath` command, such that we need to call it only once and can construct the subvolume name as `basepath` + '/' + `volID`. It would optimize a call to Ceph to retrieve this data for each Create/Delete/Mount as needed. @ajarr something to think about for the future, or not a viable alternative?

pkg/cephfs/volume.go
ajarr: It passes

pkg/cephfs/volume.go
ajarr: The pool namespace is set internally during subvolume create.
https://github.com/ceph/ceph/blob/master/src/pybind/mgr/volumes/fs/subvolume.py#L91

pkg/cephfs/volume.go
ajarr: Yes, we can do this, `ceph fs subvolumegroup getpath`. Added a tracker ticket,
http://tracker.ceph.com/issues/40617

ajarr: I don't like the idea of Ceph CSI constructing the `subvolume path` based on `subvolumegroup path`.  Aren't we breaking the abstraction if we let CSI construct subvolume path based on the subvolumegroup path?
@ShyamsundarR

ShyamsundarR: The way I look at it the `VolID` that we pass is the `subvolume` name, hence `subvolbasepath` + `subvolname` (that we set) is the subvolume path. As a result I do not see this as an abstraction break, but maybe I am reading it this way because I want to :)

My intention for the base path is to avoid repeated calls to Ceph to fetch the same, as a result improving per operation performance by that much.

pkg/cephfs/volume.go
ajarr: @poornimag , with latest nautilus you can set the pool layout, by add the args, `--pool_layoyut <pool_name>``.  can you add this now?

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
+	_, _, err := util.ExecCommand(
+		"ceph",
+		"fs",
+		"subvolume",
+		"create",
+		volOptions.FsName,
+		string(volID),
+		strconv.FormatInt(bytesQuota, 10),
+		"--group_name",
+		"csi",
```
ajarr: @poornimag , you can now set the mode and pool layout with latest nautilus. Maybe you can do this in this PR?
Add args,`` "--mode", "777", "--pool_layout", <pool_name>".

humblec: Thats great @ajarr! yeah, lets add it with this PR itself!

pkg/cephfs/volumeoptions.go
poornimag: This is the old volume type, its hardcoded and not configurable. In backward compatibility patch have added getVolumeRootPathCephDepricated

pkg/util/cephcmds.go
```diff
@@ -41,7 +41,7 @@ func ExecCommand(program string, args ...string) (stdout, stderr []byte, err err
 
 	if err := cmd.Run(); err != nil {
 		return stdoutBuf.Bytes(), stderrBuf.Bytes(), fmt.Errorf("an error (%v)"+
-			" occurred while running %s", err, program)
+			" occurred while running %s cmd: %v", err, program, cmd)
```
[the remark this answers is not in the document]
poornimag: Done

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
+			"-m", volOptions.Monitors,
+			"-c", util.CephConfigPath,
+			"-n", cephEntityClientPrefix+cr.id,
+			"--key="+cr.key)
+		if err != nil {
+			klog.Errorf("failed to create subvolume group csi, for the vol %s(%s)", string(volID), err)
 			return err
 		}
+		klog.V(4).Infof("cephfs: created subvolume group csi")
+		cephfsInit = true
```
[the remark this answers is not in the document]
poornimag: newVolumeOptions itself assumes that there is only one FsName present in the storageclass, FsName is a string and not an array. Hence had left it to be handled when we support multiple fs, in any case will add a TODO.

pkg/cephfs/volume.go
poornimag: Done

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
+func getVolumeRootPathCeph(volOptions *volumeOptions, cr *credentials, volID volumeID) string {
+	stdout, _, err := util.ExecCommand(
+		"ceph",
+		"fs",
+		"subvolume",
+		"getpath",
+		volOptions.FsName,
+		string(volID),
+		"--group_name",
+		"csi",
```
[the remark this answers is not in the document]
poornimag: Done

pkg/cephfs/volume.go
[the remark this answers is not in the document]
ShyamsundarR: The plugin can handle different fsnames as passed in from different storage classes. The comment is not about the storage class, which can only have one fsname, but about the pluigns ability to not assume it is always dealing with a single fsname. Hope this clarifies the comment, and helps when the TODO is addressed.

pkg/cephfs/volume.go
ShyamsundarR: I assume that once user creation (if needed) moves into ceph manager CLI, the plugin does not have to use the same namespace prefix as the manager to set the credential correctly (I think I have said this earlier in another comment as well). IOW, this will be transparently managed by the ceph manager CLI command, as currently we are leaking an abstraction (the assumed prefix and format of the namespace)

humblec:
@poornimag can you please update the PR description mainly focused on whats pending ?

humblec:
@ajarr @Madhu-1 can you please do final review on this PR ?

humblec:
@poornimag can you also please list the testing which has been performed on this patch ?

pkg/cephfs/volume.go
```diff
@@ -17,140 +17,109 @@ limitations under the License.
...
 
-	namespacePrefix = "ns-"
+var (
+	cephfsInit = false
 )
 
-func getCephRootPathLocal(volID volumeID) string {
-	return fmt.Sprintf("%s/controller/volumes/root-%s", PluginFolder, string(volID))
-}
+func getVolumeRootPathCeph(volOptions *volumeOptions, cr *credentials, volID volumeID) string {
```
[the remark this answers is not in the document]
poornimag: Done

Madhu-1:
let's wait for CI to pass and rook pr https://github.com/rook/rook/pull/3377 to merge

poornimag:
> @poornimag can you please update the PR description mainly focused on whats pending ?

Backward compatibility is the only thing pending. Will update it.

poornimag:
> @poornimag can you also please list the testing which has been performed on this patch ?

Have modified the e2e to use latest-nautilus-devel, hence the e2e should succeed. Its failing to execute some command in tool-box, debugging it. The create and mount of cephfs pvc is tested and succeeds in e2e.

poornimag:
> @poornimag can you also please list the testing which has been performed on this patch ?

Have modified the e2e to use latest-nautilus-devel, hence the e2e should succeed. Its failing to execute some command in tool-box, debugging it. The create and mount of cephfs pvc is tested and succeeds in e2e.

humblec:
@poornimag The rook PR (https://github.com/rook/rook/pull/3377) is merged now and we have the ground to play :)..Thanks ..

pkg/cephfs/volume.go
ajarr: +1

pkg/cephfs/volume.go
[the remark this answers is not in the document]
ajarr: where has this been fixed?

pkg/cephfs/volume.go
ajarr: where has this been fixed? If not, why?

ajarr:
LGTM, from the POV of ceph-mgr subvolume commands

humblec:
Thanks @poornimag !

humblec:
@poornimag please update the PR description :)

Madhu-1:
@poornimag E2E is passing locally?

humblec:
@Madhu-1 Can you revisit this PR ? urgency due to release .

humblec:
> please update readme for min ceph version, can be done in next PR

Yeah , we will take care in a follow up PR!

pkg/cephfs/volume.go
```diff
@@ -122,37 +189,25 @@ func purgeVolume(volID volumeID, adminCr *util.Credentials, volOptions *volumeOp
...
+		"-c", util.CephConfigPath,
+		"-n", cephEntityClientPrefix+cr.ID,
+		"--key="+cr.Key)
 	if err != nil {
-		return fmt.Errorf("failed to create mounter: %v", err)
-	}
-
-	if err = m.mount(cephRoot, adminCr, volOptions); err != nil {
-		return fmt.Errorf("error mounting ceph root: %v", err)
+		klog.Errorf("failed to purge subvolume %s(%s) in fs %s", string(volID), err, volOptions.FsName)
```
Madhu-1: avoid `volID` to string conversion use  `%v` for printing

humblec:
Thanks all!! I am removing `DNM` as per our agreement.

Issue #353, opened 2019-05-03
Release v1.1.0 of Ceph-CSI driver


humblec:
- [x] Define release process
  - Release doc would be a good start
- [x] Define release deadlines
  - Code Freeze : 25-May-2019
  - RC : 04-May -2019
  - GA : 14- Jun-2019 
- [x] Define/list roadmap for v1.1.0
  - Proposed contents for v1.1.0 release
  -- Provision cephfs volumes using Ceph manager
  -- E2E CI
  -- Kubernetes 1.14 base.
  -- CSI spec 1.1.0 
  -- Stateless implementation ( RBD, CephFS)
  -- Update sidecar containers to latest available
  - The release issues/PRs are labelled with `release-1.1.0`.

List of PRs labelled with release-1.1.0
https://github.com/ceph/ceph-csi/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3Arelease-1.1.0+

- [x] Document features and announce the release.
- [x] Release team:
  -- @humblec 
  -- @Madhu-1

humblec:
@rootfs @gman0 @j-griffith @leseb @Madhu-1 @ShyamsundarR Can you share your thoughts or add input ?

Madhu-1:
we need to make master as the default branch for this one.
@humblec assigning this to you as I see you already working on this one.

humblec:
> we need to make master as the default branch for this one.
> @humblec assigning this to you as I see you are already working on this one.

Yep, lets make `master` as the default branch. 

Before we make `master` as the default branch, I think we have to do or make `csiv1.0` and `master` in parity.

humblec:
> > we need to make master as the default branch for this one.
> > @humblec assigning this to you as I see you are already working on this one.
> 
> Yep, lets make `master` as the default branch.
> 
> Before we make `master` as the default branch, we have to do or make `csiv1.0` and `master` in parity.

https://github.com/ceph/ceph-csi/pull/363 is to bring `master` and `csi v1.0` in parity.

humblec:
Once above PR is merged, @Madhu-1 would you be changing the default branch to `master`?

Madhu-1:
yeah, am planning to make master as the default branch. I will wait till EOD if anyone has any suggestions.

humblec:
What has been achieved so far:
[x] Master is the default branch now.
[x] Master branch has been filled with missed PRs from CSI v1.0 branch

humblec:
Progress:

[...]

PR #405, opened 2019-06-04
Driver version updated to 1.1.0

humblec:
Fixes: #353 

### DIFF
--- a/pkg/cephfs/driver.go
+++ b/pkg/cephfs/driver.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	// version of ceph driver
-	version = "1.0.0"
+	version = "1.1.0"
 
 	// volIDVersion is the version number of volume ID encoding scheme
 	volIDVersion uint16 = 1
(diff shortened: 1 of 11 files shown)


humblec:
@Madhu-1 what are your thoughts on this ? will we do this as the last item for the release or its fine to do it even before it?

Madhu-1:
will do it as the last item

pkg/cephfs/driver.go
```diff
@@ -28,7 +28,7 @@ import (
 const (
 
 	// version of ceph driver
-	version = "1.0.0"
+	version = "1.1.0"
```
Madhu-1: need to update driver version for rbd  also
Note: we may need to do cleanup to have a single variable for the version

humblec:
This PR should be against `v1.1.0` branch. Adding `DNM`.

pkg/cephfs/driver.go
BlaineEXE: You could consider using a method similar to what Rook uses to set the version automatically using `go` linker flags at build time. That would prevent any version updates from being missed due to human error. 

You can look for `LDFLAGS` in Rook's `Makefile` and see how it propagates into `build/makelib/golang.mk`

humblec:
Getting this PR in as part of release process.

PR #409, opened 2019-06-06
Add mount options support in CephFS CSI driver

humblec:

This enables mount option support in cephfs driver.


humblec:
@Madhu-1 PTAL.

pkg/cephfs/nodeserver.go
```diff
@@ -177,6 +179,31 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	volCap := req.GetVolumeCapability()
```
Madhu-1: this check is already done here https://github.com/ceph/ceph-csi/blob/master/pkg/cephfs/util.go#L168

humblec: will remove that. Thanks.

humblec: @Madhu-1 updated the PR.

humblec:
@Madhu-1 it looks like the previous merge of stateless cephfs broke the system and causing CI to fail.

humblec:
>@humblec can you update readme for different supported mount options for cephfs?
>
Whatever ceph support should be supported here too! Thanks

PR #412, opened 2019-06-08
    Resize CephFS Volumes

humblec:

    Resize CephFS Volumes
    
    This feature enables CephFS Volume expansion on demand
    based on the CO resizer request. The solution is based on Ceph Mgr.
    Fix # https://github.com/ceph/ceph-csi/issues/660
    Follow up PR # README update

### DIFF
--- a/pkg/cephfs/controllerserver.go
+++ b/pkg/cephfs/controllerserver.go
@@ -324,13 +324,15 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	if err = createVolume(ctx, volOptions, cr, volumeID(volIdentifier.FsSubvolName), req.GetCapacityRange().GetRequiredBytes()); err != nil {
+	RoundOffSize := util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
+
+	if err = createVolume(ctx, volOptions, cr, volumeID(volIdentifier.FsSubvolName), RoundOffSize); err != nil {
 		klog.Errorf("failed to expand volume %s: %v", volumeID(volIdentifier.FsSubvolName), err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
 	return &csi.ControllerExpandVolumeResponse{
-		CapacityBytes:         req.GetCapacityRange().GetRequiredBytes(),
+		CapacityBytes:         RoundOffSize,
 		NodeExpansionRequired: false,
 	}, nil
 
--- a/pkg/cephfs/controllerserver.go
+++ b/pkg/cephfs/controllerserver.go
@@ -292,3 +292,39 @@ func (cs *ControllerServer) ValidateVolumeCapabilities(
 		},
 	}, nil
 }
+
+// ExpandVolume expand CephFS Volumes on demand based on resizer request
+func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
+
+	if err := cs.validateExpandVolumeRequest(req); err != nil {
+		klog.Errorf("ControllerExpandVolumeRequest validation failed: %v", err)
+		return nil, err
+	}
+
+	volID := req.GetVolumeId()
+	secret := req.GetSecrets()
+
+	cr, err := util.NewAdminCredentials(secret)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	defer cr.DeleteCredentials()
+
+	volOptions, volIdentifier, err := newVolumeOptionsFromVolID(ctx, volID, nil, secret)
+
+	if err != nil {
+		klog.Errorf("validation and extraction of volume options failed: %v", err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	if err = createVolume(ctx, volOptions, cr, volumeID(volIdentifier.FsSubvolName), req.GetCapacityRange().GetRequiredBytes()); err != nil {
+		klog.Errorf("failed to expand volume %s: %v", volumeID(volIdentifier.FsSubvolName), err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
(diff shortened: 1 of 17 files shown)


humblec:
This PR depends on #400

yuchunyun:
May I ask when the PR can be passed. I am eager for it

D1abloRUS:
> May I ask when the PR can be passed. I am eager for it

yep we need this! @humblec

yuchunyun:
> yep we need this

I copied the project and modified the code by myself. This function has been implemented

D1abloRUS:
> > yep we need this
> 
> I copied the project and modified the code by myself. This function has been implemented

You tested on k8s? this is worked?

yuchunyun:
> > > yep we need this
> > 
> > 
> > I copied the project and modified the code by myself. This function has been implemented
> 
> You tested on k8s? this is worked?

yes

humblec:
@D1abloRUS @yuchunyun Sure, thanks for the confirmation, I will rebase this and push it soon!

humblec:
@yuchunyun reabsed to master, this should be it :), please confirm. 

@Madhu-1 @poornimag PTAL .

pkg/cephfs/controllerserver.go
```diff
@@ -285,3 +285,39 @@ func (cs *ControllerServer) ValidateVolumeCapabilities(
...
+	defer cr.DeleteCredentials()
+
+	volOptions, volIdentifier, err := newVolumeOptionsFromVolID(ctx, volID, nil, secret)
+
+	if err != nil {
+		klog.Errorf("validation and extraction of volume options failed: %v", err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	if err = createVolume(ctx, volOptions, cr, volumeID(volIdentifier.FsSubvolName), req.GetCapacityRange().GetRequiredBytes()); err != nil {
```
ajarr: Ideally, we don't want to use the `create subvolume` to resize. how urgent is this feature? We should use `fs subvolume resize` interface to resize CephFS subvolumes, https://github.com/ceph/ceph/pull/30054/

Madhu-1: why create volume is used here?

pkg/cephfs/controllerserver.go
```diff
@@ -285,3 +285,39 @@ func (cs *ControllerServer) ValidateVolumeCapabilities(
 		},
 	}, nil
 }
+
+// ExpandVolume expand CephFS Volumes on demand based on resizer request
+func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
```
Madhu-1: volume ID locking is missing

pkg/cephfs/controllerserver.go
Madhu-1: we need to round off volume size and return the same

joscollin: @humblec We will have `fs subvolume resize` available in master by this weekend. But not sure about the next nautilus release date for you to use it. I'd suggest you to check with the Ceph Community Mailinglist about the date, so that you could take a decision.

[create_subvolume](https://github.com/ceph/ceph/blob/master/src/pybind/mgr/volumes/fs/subvolume.py#L60) should allow you to resize the subvolume for now. Then it's better you have [this bugfix](https://tracker.ceph.com/issues/41884) in the version of nautilus you're using. This bug hits the exact code you're using now.

ajarr: > @ajarr this PR Is important, so I would rather stick to createVolume for now and we can accomodate fs subvolume resize whenever its available to consume.

Sounds OK

ajarr:
@humblec , does the CSI driver check whether the requested size is
* a positive integer
* greater than the existing size, as  only ExpandVolume is implemented here?

pkg/cephfs/controllerserver.go
humblec: Added

pkg/cephfs/controllerserver.go
humblec: done.

humblec:
> please add an E2E

Done.

pkg/cephfs/controllerserver.go
humblec: :+1:

humblec:
The tests are passing except `1.13`, this failure looks to be because of some CI issue where minikube is failing to deploy 1.13 version. In short, this PR is complete and E2E is also passing!!. :+1:  

Can I get final review on this PR ? @Madhu-1 @nixpanic @ajarr  ?

humblec:
@joscollin PTAL..

pkg/cephfs/util.go
```diff
@@ -114,3 +114,21 @@ func (cs *ControllerServer) validateDeleteVolumeRequest() error {
...
+		return fmt.Errorf("invalid ExpandVolumeRequest: %v", err)
+	}
+
+	if req.GetVolumeId() == "" {
+		return status.Error(codes.InvalidArgument, "Volume ID cannot be empty")
+	}
+
+	capRange := req.GetCapacityRange()
+	if capRange == nil {
+		return status.Error(codes.InvalidArgument, "CapacityRange cannot be empty")
```
joscollin: How do you unset the quota for no size limit? See the `create_subvolume` definition below.

```
    def create_subvolume(self, spec, size=None, namespace_isolated=True, mode=0o755, pool=None):                                                                                              
        """                                                                                                                                                                                   
        Set up metadata, pools and auth for a subvolume.                                                                                                                                      
                                                                                                                                                                                              
        This function is idempotent.  It is safe to call this again                                                                                                                           
        for an already-created subvolume, even if it is in use.                                                                                                                               
                                                                                                                                                                                              
        :param spec: subvolume path specification                                                                                                                                             
        :param size: In bytes, or None for no size limit       
```

humblec: @joscollin you will never get a request with size NONE from CO, so we are good .

joscollin: What I meant was, You need to allow size `NONE` (or what ever) to allow the user to unset the quota.

humblec:
The tests are passed ! can I get final review here ?

PR #423, opened 2019-06-11
change permission of targetpath

Madhu-1:
setting the permission of  targetpath to 777 will allow non-root user to write to pv

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -103,13 +103,10 @@ func waitForDeploymentComplete(name, ns string, c clientset.Interface, t int) er
 	return nil
 }
 
-func execCommandInToolBox(f *framework.Framework, c string) string {
+func execCommandInPod(f *framework.Framework, c, ns string, opt *metav1.ListOptions) string {
 
 	cmd := []string{"/bin/sh", "-c", c}
-	opt := metav1.ListOptions{
-		LabelSelector: "app=rook-ceph-tools",
-	}
-	podList, err := f.PodClientNS(rookNS).List(opt)
+	podList, err := f.PodClientNS(ns).List(*opt)
 	framework.ExpectNoError(err)
 	Expect(podList.Items).NotTo(BeNil())
 	Expect(err).Should(BeNil())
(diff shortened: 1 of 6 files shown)


e2e/utils.go
```diff
@@ -486,3 +496,76 @@ func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
...
+
+	err = createApp(f.ClientSet, app, deployTimeout)
+	if err != nil {
+		Fail(err.Error())
+	}
+
+	opt := metav1.ListOptions{
+		LabelSelector: "app=pod-run-as-non-root",
+	}
+	execCommandInPod(f, "echo testing > /target/testing", app.Namespace, &opt)
```
humblec: Dont we need an error check here ? @Madhu-1

Madhu-1:
@humblec @ShyamsundarR PTAL

e2e/utils.go
Madhu-1: @humblec error check already done here https://github.com/ceph/ceph-csi/blob/9d27b6b425c0a63d8b664047dfd1a9172cbbfe2f/e2e/utils.go#L128-L129

humblec:
LGTM. Approved..

humblec:
Any comments here @poornimag @ShyamsundarR

Madhu-1:
> The solution to set the mode bits to 777 **every** publish seems like risky behavior. I understand that this was used for similar issues in other CSI implementations, and that a more generic Kubernetes alternative to pass in the `fsgroup` from the security context of the pod is in progress.
> 
> It would be really useful it to have an issue to track the same in this repository and get the final version of this fix in place once the Kubernetes end is fixed. @Madhu-1 If we have such an issue I can add my approval to this and we can move forward.
> 

@humblec @JohnStrunk is there any issue created for this in kubernetes 

> Otherwise changes here look fine, and I have no other comments.

ShyamsundarR:
> > The solution to set the mode bits to 777 **every** publish seems like risky behavior. I understand that this was used for similar issues in other CSI implementations, and that a more generic Kubernetes alternative to pass in the `fsgroup` from the security context of the pod is in progress.
> > It would be really useful it to have an issue to track the same in this repository and get the final version of this fix in place once the Kubernetes end is fixed. @Madhu-1 If we have such an issue I can add my approval to this and we can move forward.
> 
> @humblec @JohnStrunk is there any issue created for this in kubernetes

Understood (offline) that this is being handled in possibly a few different ways in Kubernetes. For now, let's go ahead with this option.

> 
> > Otherwise changes here look fine, and I have no other comments.

Issue #424, opened 2019-06-12
ceph versions supported by ceph-csi


Madhu-1:
we need to add a matrix related to the minimum and maximum version of ceph supported by the ceph-csi driver.

humblec:
Yep . I will update it along with the release update. Thanks

humblec:
Similar to #421

Madhu-1:
Fixed in https://github.com/ceph/ceph-csi/pull/468

PR #425, opened 2019-06-12
Remove child clone volume reference from the parent volume

### LINKED ISSUE #70: [csi-rbd] detach cloned images from snapshots
If rbd snapshots have child cloned images, we can't remove the snapshots because of protecting the snapshots. We can use `rbd flatten` for detaching cloned images from snapshots or we can consider using RBD clone v2 feature with later versions of Mimic as well.
The short-term plan would be using `rbd flatten` for detaching cloned images from snapshots.

### LINKED ISSUE #227: Unremoved RBD volume after removing PVC with snapshot
**Describe the bug**
Unremoved RBD volume after removing PVC with snapshot

**Kubernetes and Ceph CSI Versions**
Kubernetes: 1.13
Ceph CSI: 1.0.0

**Ceph CSI Driver logs**
Post logs from rbdplugin/cephfs plugin, provisioner, etc.

**To Reproduce**
From examples/rbd dir: (with Block mode for all PVCs)

1. Create PVC
2. Create VolumeSnapshot from PVC
3. Create PVC from VolumeSnapshot
4. Delete PVC from step 1 ( Volume still present in Ceph Server after deleting from k8s)
5. Delete PVC from step 2
6. Delete VolumeSnapshot
7. RBD volume of PVC from step 1 is still present in the Ceph Server

**Expected behavior**
The RBD volume from step 1 should be deleted.

Madhu-1:

when a Cloned images retain a reference to the parent snapshot then
we cannot delete the volume. flattening will remove the reference
so that volume can be deleted even if it is cloned from a snapshot

Added a check in delete volume, if the volume contains any snapshot
we are not allowing the volume deletion, prior to volume delete
all its associated snapshots should be deleted first.

## Related issues ##
closes #70, closes #227

### DIFF
--- a/pkg/rbd/controllerserver.go
+++ b/pkg/rbd/controllerserver.go
@@ -18,6 +18,7 @@ package rbd
 
 import (
 	"fmt"
+	"strings"
 
 	csicommon "github.com/ceph/ceph-csi/pkg/csi-common"
 	"github.com/ceph/ceph-csi/pkg/util"
@@ -210,6 +211,12 @@ func (cs *ControllerServer) checkSnapshot(req *csi.CreateVolumeRequest, rbdVol *
 	if err != nil {
 		return status.Error(codes.Internal, err.Error())
 	}
+
+	err = flattenImage(rbdVol, req.GetSecrets())
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+
 	klog.V(4).Infof("create volume %s from snapshot %s", req.GetName(), rbdSnap.RbdSnapName)
 	return nil
 }
@@ -272,6 +279,30 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		}
 	}()
 
+	// check if rbd image has snapshot
+	key, err := getKey(rbdVol.AdminID, req.GetSecrets())
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	snaps, err := listSnapshot(rbdVol.Monitors, rbdVol.AdminID, key, rbdVol.Pool,
+		rbdVol.RbdImageName)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	// reject volume deletion if it has snapshot
+	if len(snaps) != 0 {
+		return nil, status.Errorf(codes.FailedPrecondition, "%s is having snapshot,please delete snapshot before deleting volume", req.GetVolumeId())
+	}
+
+	// remove the reference from volume and snapshot
+	// this will occur if flatten image failed during volume clone
+	err = flattenImage(rbdVol, req.GetSecrets())
+	// if the image does not contain reference to snapshot, flattern will return
+	// image has no parent error
+
+	if err != nil && !strings.Contains(err.Error(), "image has no parent") {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
 	// Deleting rbd image
 	klog.V(4).Infof("deleting image %s", rbdVol.RbdImageName)
 	if err := deleteImage(rbdVol, rbdVol.AdminID, req.GetSecrets()); err != nil {
--- a/pkg/rbd/rbd_util.go
+++ b/pkg/rbd/rbd_util.go
@@ -550,6 +550,26 @@ func createSnapshot(pOpts *rbdSnapshot, adminID string, credentials map[string]s
 	return nil
 }
 
+// flattenImage remove the reference from the child clone to the parent
+// snapshot (refer http://docs.ceph.com/docs/giant/rbd/rbd-snapshot/#flattening-a-cloned-image)
+func flattenImage(vol *rbdVolume, credentials map[string]string) error {
+	var output []byte
+
+	key, err := getKey(vol.AdminID, credentials)
+	if err != nil {
+		return err
+	}
+	klog.V(4).Infof("rbd: flatten image %s using mon %s, pool %s", vol.RbdImageName, vol.Monitors, vol.Pool)
+	args := []string{"flatten", "--pool", vol.Pool, "--image", vol.RbdImageName, "--id", vol.AdminID, "-m", vol.Monitors, "--key=" + key}
+
+	output, err = execCommand("rbd", args)
+
+	if err != nil {
+		return errors.Wrapf(err, "failed to flatten image, command output: %s", string(output))
+	}
+
+	return nil
+}
 func unprotectSnapshot(pOpts *rbdSnapshot, adminID string, credentials map[string]string) error {
 	var output []byte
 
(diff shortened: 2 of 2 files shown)


Madhu-1:
don't merge this PR till I add snapshot e2e for the current code

pkg/rbd/controllerserver.go
```diff
@@ -272,6 +279,30 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
...
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	// reject volume deletion if it has snapshot
+	if len(snaps) != 0 {
+		return nil, status.Errorf(codes.FailedPrecondition, "%s is having snapshot,please delete snapshot before deleting volume", req.GetVolumeId())
+	}
+
+	// remove the reference from volume and snapshot
+	// this will occuer if flatten image failed during volume clone
```
humblec: typo ..

pkg/rbd/rbd_util.go
```diff
@@ -550,6 +550,25 @@ func createSnapshot(pOpts *rbdSnapshot, adminID string, credentials map[string]s
 	return nil
 }
 
+// flattenImage remove the reference from the child clone to the parent snapshot
```
humblec: @Madhu-1 its also good to refer the documentation here in the source code comment.

pkg/rbd/controllerserver.go
```diff
@@ -210,6 +211,12 @@ func (cs *ControllerServer) checkSnapshot(req *csi.CreateVolumeRequest, rbdVol *
 	if err != nil {
 		return status.Error(codes.Internal, err.Error())
 	}
+
+	err = flattenImage(rbdSnap.Pool, rbdSnap.RbdImageName, rbdVol.Monitors, rbdVol.AdminID, req.GetSecrets())
```
humblec: @Madhu-1 flattenImage can take rbdSnap struct to make better argument handling. Isnt it ?

humblec:
@Madhu-1 quick review comments from me, PTAL.

Madhu-1:
@humblec Thanks for the quick review. addressed review comments PTAL

pkg/rbd/rbd_util.go
```diff
@@ -550,6 +550,26 @@ func createSnapshot(pOpts *rbdSnapshot, adminID string, credentials map[string]s
...
+	key, err := getKey(vol.AdminID, credentials)
+	if err != nil {
+		return err
+	}
+	klog.V(4).Infof("rbd: flatten image %s using mon %s, pool %s", vol.RbdImageName, vol.Monitors, vol.Pool)
+	args := []string{"flatten", "--pool", vol.Pool, "--image", vol.RbdImageName, "--id", vol.AdminID, "-m", vol.Monitors, "--key=" + key}
+
+	output, err = execCommand("rbd", args)
+
+	if err != nil {
```
dillaman: This should be expected to fail if the image doesn't have a parent. Additionally, if the image has snapshots and does not have the 'deep-flatten' feature enabled, the snapshots themselves will still be linked to the parent image.

Madhu-1: @dillaman Thanks for the review,
do I need to explicitly enable `deep-flatten` feature during `clone` and `create` or is  it enabled by default

dillaman: You cannot enable it on existing images, it needs to have been enabled when the image was created. Support for this in krbd wasn't added until v5.1.

Madhu-1: @dillaman don't we need to enable this feature during `clone` (cloned image can be used later to snapshot and clone a PVC from it)

dillaman: @Madhu-1 negative -- ```deep-flatten``` isn't required for cloning, ```layering``` is the required feature.

Thinking holistically, we added the "rbd trash mv" tool to support cases where you cannot delete a parent image due to linked clones. In the Nautilus release, this is even automated for you via the ```rbd_move_parent_to_trash_on_remove``` config option (requires Mimic-or-later cluster). We are also adding automated background operation handling [1]. 

From an operator's point-of-view, we should never really expose the RBD implementation details for how snapshots and volume from snapshots (RBD clones) work on the backend. If you attempt to delete an image that has linked clones, move it to the trash for later cleanup. If you attempt to delete a snapshot with a linked clone, w/ clone v2 it just works. To support clone v1 (and v2), perhaps k8s snapshots could create a new RBD cloned image (and then it follows the previous rule about 'move-to-trash' upon delete). 

[1] https://trello.com/c/JO5BPkRG

Madhu-1: https://github.com/ceph/ceph-csi/issues/424

Madhu-1:
@dillaman can you confirm below-listed steps are correct to clone an image with a different version of ceph
# steps to create rbd clone
```
ceph versions less than Mimic

[1]create an image with the `deep-flattern` option
[2]create a snapshot
[3] protect snapshot
[4]clone a new image from snapshot
[5]flattern newly  created image
[6]delete the newly created image[5] (this will be independent step can be one at last  also)
[7]unprotect snapshot
[8]delete snapshot 
[8]delete image  create in  step[1](this image cannot be deleted if it has snapshots)

ceph version Mimic onwards

[1]create an image with the `deep-flattern` option
[2]create a snapshot
[3]clone a new image from snapshot
[4]flattern newly cloned image
[5]delete the cloned image[3]  (this will be independent step can be one at last  also)
[6]delete image created in [1] with `rbd rm --rbd_move_to_trash_on_remove=true`
[7]delete snapshot ([6] and [7] may interchange)
The image moved to trash will be automatically deleted once the snapshot is deleted?
```
@humblec @j-griffith @ShyamsundarR let me know your thoughts on below query
* what versions of ceph we should support with csi?
* if version less than Mimic do we need to handle cloning differently 
   - use clone v1 for ceph version less than Mimic
   - use clone v2 for Mimic later

based on the answer I will update this patch (this patch will help to finish smart cloning work)

Madhu-1:
@dillaman @humblec @ShyamsundarR @j-griffith  can I get an answer to the above questions?

dillaman:
@Madhu-1 The issue w/ ```deep-flatten``` + ```k8s``` is that by default and design, the ```krbd``` kernel driver is used. This driver is not tied to a Ceph release and doesn't support ```deep-flatten``` until v5.1.

Since ```deep-flatten``` is only required once a snapshot is created on a cloned image, perhaps you can just defer the ```flatten``` operation until the first snapshot is created on the cloned image (since it may never occur for most images and therefore saves disk space) and/or run ```flatten``` immediately after cloning if ```deep-flatten``` isn't enabled.

One important note is that if the ```exclusive-lock``` feature is enabled, you cannot run the ```flatten``` operation when ```krbd``` is using the image, so it would need to be run immediately after the clone operation.

j-griffith:
@Madhu-1 
>     what versions of ceph we should support with csi?
I don't have enough visibility in to the field to know the best answer here.  My thought was that we would not have a limitation on ceph version, but provide different capabilities based on version.
>     if version less than Mimic do we need to handle cloning differently
>         use clone v1 for ceph version less than Mimic
Or don't support cloning in the capabilities when the version is < Mimic?  I don't know the implications of this in terms of customers in production with Kube; but given that cloning is alpha in 1.15 it's pretty "new" so setting a cut off doesn't seem heavy handed to me.
>         use clone v2 for Mimic later
Same note as above, it's up to you though if you want to have two implementations and select based on ceph version that's great; so long as the implementations are clean and easily separated.  I would hate to see too much overhead or maintenance cost by trying to support both.

humblec:
>   what versions of ceph we should support with csi?
    if version less than Mimic do we need to handle cloning differently
        use clone v1 for ceph version less than Mimic
        use clone v2 for Mimic later
>

@Madhu-1 imo, from `>v1.0` onwards we will be supporting `nautilus`, especially for `clone` its good enough to declare we support this feature  ONLY for nautlius.

Madhu-1:
closing this one, as flatten will be taken care in volume cloning PR https://github.com/ceph/ceph-csi/pull/693

PR #429, opened 2019-06-13
update travis to run functional test against different kube version

Madhu-1:

updated Travis to run functional tests against different kubernetes versions
1.13.7 and 1.14.3

This will ensure csi will be working against 2 kube major version
if required we can add all major version to test against it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,11 @@ go: 1.11.x
 
 env:
   global:
-    - GOLANGCI_VERSION="v1.17.0"
+    - GOLANGCI_VERSION=v1.17.0
     - TEST_COVERAGE=stdout
     - GO_METALINTER_THREADS=1
     - GO_COVER_DIR=_output
     - VM_DRIVER=none
-    - KUBE_VERSION=v1.14.2
     - MINIKUBE_VERSION=v1.1.1
     - CHANGE_MINIKUBE_NONE_USER=true
     - KUBECONFIG=$HOME/.kube/config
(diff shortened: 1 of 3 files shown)


Madhu-1:
@humblec @ShyamsundarR PTAL

.travis.yml
```diff
@@ -39,26 +38,26 @@ jobs:
         - pip install --user --upgrade pip
         - pip install --user yamllint
         # install golangci-lint
+        # workaround for https://github.com/golangci/golangci-lint/issues/575
         - curl -sf
-          "https://install.goreleaser.com/github.com/golangci/golangci-lint.sh"
+          "https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh"
```
ShyamsundarR: I always cringe when using `master` branch versions of dependencies... letting this one pass through, and hopefully in the future we would stick to a version.

.travis.yml
```diff
@@ -39,26 +38,26 @@ jobs:
...
           | bash -s -- -b $GOPATH/bin "${GOLANGCI_VERSION}"
       script:
         - scripts/lint-text.sh --require-all
         - scripts/lint-go.sh
         - scripts/test-go.sh
 
-    - name: cephcsi
+    - name: cephcsi with kube 1.13.7
       script:
         - scripts/skip-doc-change.sh || travis_terminate 0;
```
ShyamsundarR: We can possibly move this line up to the `static-check-make` section (last line) and hence skip the check in future sections like in line 58 in the new file.

Can be taken care the next time we edit this file.

scripts/travis-functest.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
ShyamsundarR: Add -e to the end?

In the older case when each of the following script lines fail, travis would exit with an error, now as these are moved to a script, travis would check final exit state of the script as I understand. Hence to catch failures of each command/script invocation and fail immediately we should add the -x. I have not tested this though.

.travis.yml
Madhu-1: will go back to the older version of code once the above-mentioned issue is fixed

.travis.yml
Madhu-1: yes nice idea will update if it works

scripts/travis-functest.sh
Madhu-1: done

Madhu-1:
@humblec can you review this. This will Fix  CI issue also

PR #432, opened 2019-06-14
Revert "Use Deployment with leader election instead of StatefulSet"

Madhu-1:

This reverts commit a151bec94b93f0f8ddbe50b572bcbe94e4c344ae.
we need to revert back to use statefulset as we will face issue in using deployment

Do you have any questions?

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##


Madhu-1:
@poornimag PTAL

Issue #360, opened 2019-05-07
Document exact capabilities a id requires against a Ceph cluster for RBD and CephFS plugins


ShyamsundarR:
The [various](https://github.com/ceph/ceph-csi/blob/3a308eaeeec724ed2bf4163a9e58d5c50bbde2b2/examples/rbd/secret.yaml#L9) [secrets](https://github.com/ceph/ceph-csi/blob/3a308eaeeec724ed2bf4163a9e58d5c50bbde2b2/examples/cephfs/secret.yaml#L13) that need to be provided to the plugins for provisioning and mounting needs documentation regarding exact capabilities that these identities require.

This helps create identities that have enough access to perform the plugin actions and also ensure that the capabilities are not too lenient to cause a security concern.

Madhu-1:
do we need to document these capabilities or we can just point to the ceph user management for information?

ShyamsundarR:
We need to document what capabilities CSI needs. Ceph documentation links can be pointed to for how to create it etc. but exact capabilities that are needed by the ID needs documentation from the CSI end I would think.

ajarr:
For CephFS CSI plugin, please see,
https://github.com/ceph/ceph-csi/pull/437#issuecomment-511460712

ajarr:
@ShyamsundarR would you know the capabilities required by the RBD plugin? We can then close this issue

ShyamsundarR:
> @ShyamsundarR would you know the capabilities required by the RBD plugin? We can then close this issue

I actually do not, the best reference I have/had is/was [here](https://github.com/rook/rook/blob/7fb43d72ebfaba5afbb8faf017962a38364b2901/Documentation/ceph-csi-drivers.md#create-rbd-secret)

The credentials needed would be (from above),

`ceph auth get-or-create-key client.kubernetes mon \"allow profile rbd\" osd \"profile rbd pool=<poolname>\""`

If someone could validate this, and add it (and the CephFS requirement) to the documentation we can close this issue.

ajarr:
This is a critical issue. Currently, the CSI volumes are mounted with cephx admin user with full privileges. This allows Ceph CSI volume users to misuse them, e.g, RWX CephFS volume users might take lot CephFS snapshots of the CSI volume without the k8s csi interface that is currently unavailable.
We need to restrict the privileges of the cephx user used to mount the CSI volumes.

dillaman:
[...]

PR #437, opened 2019-06-20
Cephfs: Remove user creation for every volume

poornimag:
Currently, provisioner creates user for every volume and nodeplugin
uses this user to mount that volume. But nodeplugin and provisioner
already have admin credentials, hence using the admin credentials
to mount the volume and getting rid of user creation for each volume.

### DIFF
--- a/pkg/cephfs/controllerserver.go
+++ b/pkg/cephfs/controllerserver.go
@@ -44,8 +44,7 @@ var (
 	volumeNameLocker = util.NewIDLocker()
 )
 
-// createBackingVolume creates the backing subvolume and user/key for the given volOptions and vID,
-// and on any error cleans up any created entities
+// createBackingVolume creates the backing subvolume and on any error cleans up any created entities
 func (cs *ControllerServer) createBackingVolume(volOptions *volumeOptions, vID *volumeIdentifier, secret map[string]string) error {
 	cr, err := util.GetAdminCredentials(secret)
 	if err != nil {
@@ -63,11 +62,6 @@ func (cs *ControllerServer) createBackingVolume(volOptions *volumeOptions, vID *
 			}
 		}
 	}()
-
-	if _, err = createCephUser(volOptions, cr, volumeID(vID.FsSubvolName)); err != nil {
-		klog.Errorf("failed to create ceph user for volume %s: %v", volOptions.RequestName, err)
-		return status.Error(codes.Internal, err.Error())
-	}
 
 	return nil
 }
@@ -188,7 +182,7 @@ func (cs *ControllerServer) deleteVolumeDeprecated(req *csi.DeleteVolumeRequest)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if err = deleteCephUser(&ce.VolOptions, cr, volID); err != nil {
+	if err = deleteCephUserDeprecated(&ce.VolOptions, cr, volID); err != nil {
 		klog.Errorf("failed to delete ceph user for volume %s: %v", volID, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
@@ -244,11 +238,6 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 
 	if err = purgeVolume(volumeID(vID.FsSubvolName), cr, volOptions); err != nil {
 		klog.Errorf("failed to delete volume %s: %v", volID, err)
-		return nil, status.Error(codes.Internal, err.Error())
-	}
-
-	if err = deleteCephUser(volOptions, cr, volumeID(vID.FsSubvolName)); err != nil {
-		klog.Errorf("failed to delete ceph user for volume %s: %v", volID, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
--- a/pkg/cephfs/cephuser.go
+++ b/pkg/cephfs/cephuser.go
@@ -17,8 +17,6 @@ limitations under the License.
 package cephfs
 
 import (
-	"fmt"
-
 	"github.com/ceph/ceph-csi/pkg/util"
 )
 
@@ -27,81 +25,15 @@ const (
 	cephEntityClientPrefix = "client."
 )
 
-type cephEntityCaps struct {
-	Mds string `json:"mds"`
-	Mon string `json:"mon"`
-	Osd string `json:"osd"`
-}
-
-type cephEntity struct {
-	Entity string         `json:"entity"`
-	Key    string         `json:"key"`
-	Caps   cephEntityCaps `json:"caps"`
-}
-
-func (ent *cephEntity) toCredentials() *util.Credentials {
-	return &util.Credentials{
-		ID:  ent.Entity[len(cephEntityClientPrefix):],
-		Key: ent.Key,
-	}
+func genUserIDs(adminCr *util.Credentials, volID volumeID) (adminID, userID string) {
+	return cephEntityClientPrefix + adminCr.ID, cephEntityClientPrefix + getCephUserName(volID)
 }
 
 func getCephUserName(volID volumeID) string {
 	return cephUserPrefix + string(volID)
 }
 
-func getSingleCephEntity(args ...string) (*cephEntity, error) {
-	var ents []cephEntity
-	if err := execCommandJSON(&ents, "ceph", args...); err != nil {
-		return nil, err
-	}
-
-	if len(ents) != 1 {
-		return nil, fmt.Errorf("got unexpected number of entities: expected 1, got %d", len(ents))
-	}
-
-	return &ents[0], nil
-}
-
-func genUserIDs(adminCr *util.Credentials, volID volumeID) (adminID, userID string) {
-	return cephEntityClientPrefix + adminCr.ID, cephEntityClientPrefix + getCephUserName(volID)
-}
-
-func getCephUser(volOptions *volumeOptions, adminCr *util.Credentials, volID volumeID) (*cephEntity, error) {
-	adminID, userID := genUserIDs(adminCr, volID)
-
-	return getSingleCephEntity(
-		"-m", volOptions.Monitors,
-		"-n", adminID,
-		"--key="+adminCr.Key,
-		"-c", util.CephConfigPath,
-		"-f", "json",
-		"auth", "get", userID,
-	)
-}
-
-func createCephUser(volOptions *volumeOptions, adminCr *util.Credentials, volID volumeID) (*cephEntity, error) {
-	adminID, userID := genUserIDs(adminCr, volID)
-	volRootPath, err := getVolumeRootPathCeph(volOptions, adminCr, volID)
-	if err != nil {
-		return nil, err
-	}
-
-	return getSingleCephEntity(
-		"-m", volOptions.Monitors,
-		"-n", adminID,
-		"--key="+adminCr.Key,
-		"-c", util.CephConfigPath,
-		"-f", "json",
-		"auth", "get-or-create", userID,
-		// User capabilities
-		"mds", fmt.Sprintf("allow rw path=%s", volRootPath),
-		"mon", "allow r",
-		"osd", fmt.Sprintf("allow rw pool=%s namespace=%s", volOptions.Pool, getVolumeNamespace(volID)),
-	)
-}
-
-func deleteCephUser(volOptions *volumeOptions, adminCr *util.Credentials, volID volumeID) error {
+func deleteCephUserDeprecated(volOptions *volumeOptions, adminCr *util.Credentials, volID volumeID) error {
 	adminID, userID := genUserIDs(adminCr, volID)
 
 	// TODO: Need to return success if userID is not found
--- a/pkg/cephfs/mountcache.go
+++ b/pkg/cephfs/mountcache.go
@@ -104,13 +104,6 @@ func mountOneCacheEntry(volOptions *volumeOptions, vid *volumeIdentifier, me *vo
 		if err != nil {
 			return err
 		}
-
-		var entity *cephEntity
-		entity, err = getCephUser(volOptions, cr, volumeID(vid.FsSubvolName))
-		if err != nil {
-			return err
-		}
-		cr = entity.toCredentials()
 	} else {
 		cr, err = util.GetUserCredentials(decodeCredentials(me.Secrets))
 		if err != nil {
--- a/pkg/cephfs/nodeserver.go
+++ b/pkg/cephfs/nodeserver.go
@@ -40,7 +40,7 @@ var (
 	nodeVolumeIDLocker = util.NewIDLocker()
 )
 
-func getCredentialsForVolume(volOptions *volumeOptions, volID volumeID, req *csi.NodeStageVolumeRequest) (*util.Credentials, error) {
+func getCredentialsForVolume(volOptions *volumeOptions, req *csi.NodeStageVolumeRequest) (*util.Credentials, error) {
 	var (
 		cr      *util.Credentials
 		secrets = req.GetSecrets()
@@ -56,14 +56,7 @@ func getCredentialsForVolume(volOptions *volumeOptions, volID volumeID, req *csi
 			return nil, fmt.Errorf("failed to get admin credentials from node stage secrets: %v", err)
 		}
 
-		// Then get the ceph user
-
-		entity, err := getCephUser(volOptions, adminCr, volID)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get ceph user: %v", err)
-		}
-
-		cr = entity.toCredentials()
+		cr = adminCr
 	} else {
 		// The volume is pre-made, credentials are in node stage secrets
 
@@ -82,7 +75,6 @@ func getCredentialsForVolume(volOptions *volumeOptions, volID volumeID, req *csi
 func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
 	var (
 		volOptions *volumeOptions
-		vid        *volumeIdentifier
 	)
 	if err := validateNodeStageVolumeRequest(req); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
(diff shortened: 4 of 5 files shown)


ShyamsundarR:
@poornimag the intention is to provide an option to choose between, per PV ID/key generation versus using the admin credentials for the FS to mount the volume, as detailed in issue #269 

This PR removes the entire per PV user creation which is not what was desired. Is there a different reasoning at the moment that necessitates this removal?

poornimag:
> @poornimag the intention is to provide an option to choose between, per PV ID/key generation versus using the admin credentials for the FS to mount the volume, as detailed in issue #269
> 
> This PR removes the entire per PV user creation which is not what was desired. Is there a different reasoning at the moment that necessitates this removal?

Apologies, i wasn't aware of the approach of making user id optional. When nodeserver already has admin credentials, what is the need of creating user per volume? We create the user and use that user, so i m not sure what security issue its fixing. Also, Ramana mentioned that he would be adding subvolume group user create which would have access to a given group. Eventually we would need to use that instead of admin?

humblec:
@ShyamsundarR I do believe per volume user creation may not be something we have to consider atm for the reasons mentioned in @poornimag's comment.

ShyamsundarR:
> > @poornimag the intention is to provide an option to choose between, per PV ID/key generation versus using the admin credentials for the FS to mount the volume, as detailed in issue #269
> > This PR removes the entire per PV user creation which is not what was desired. Is there a different reasoning at the moment that necessitates this removal?
> 
> Apologies, i wasn't aware of the approach of making user id optional. When nodeserver already has admin credentials, what is the need of creating user per volume? We create the user and use that user, so i m not sure what security issue its fixing. Also, Ramana mentioned that he would be adding subvolume group user create which would have access to a given group. Eventually we would need to use that instead of admin?

As per my understanding when opening the issue, the intention was to make it optional to improve the performance of VolumeCreate (and Delete), it was not to do away with the per PV secret for users who needed it.

I am not aware of why there is a per PV user required or its security implications. I am tagging @liewegas @batrick and @gman0  for comments and direction.

Even if this functionality is removed for 1.1, one thing I would like to point out is that for backward compatibility reasons, 1.0 CephFS created volumes would need to use the user created for the same during mount, and so we cannot entirely delete the code.

poornimag:
> > > @poornimag the intention is to provide an option to choose between, per PV ID/key generation versus using the admin credentials for the FS to mount the volume, as detailed in issue #269
> > > This PR removes the entire per PV user creation which is not what was desired. Is there a different reasoning at the moment that necessitates this removal?
> > 
> > 
> > Apologies, i wasn't aware of the approach of making user id optional. When nodeserver already has admin credentials, what is the need of creating user per volume? We create the user and use that user, so i m not sure what security issue its fixing. Also, Ramana mentioned that he would be adding subvolume group user create which would have access to a given group. Eventually we would need to use that instead of admin?
> 
> As per my understanding when opening the issue, the intention was to make it optional to improve the performance of VolumeCreate (and Delete), it was not to do away with the per PV secret for users who needed it.
> 
> I am not aware of why there is a per PV user required or its security implications. I am tagging @liewegas @batrick and @gman0 for comments and direction.
> 
> Even if this functionality is removed for 1.1, one thing I would like to point out is that for backward compatibility reasons, 1.0 CephFS created volumes would need to use the user created for the same during mount, and so we cannot entirely delete the code.
WRT backward compatibility, i think the admin user will be able to mount the volumes irrespective of the users that were created for the subvolume. So we may not need backward compatibility code, need to confirm.

humblec:
Once we have subvol group user in ceph manager , we can consider adding the functionality for cephfs volumes. For now, i believe this per pv user creation or option/choice does not add much value, rather confusing.

poornimag:
> > > > @poornimag the intention is to provide an option to choose between, per PV ID/key generation versus using the admin credentials for the FS to mount the volume, as detailed in issue #269
> > > > This PR removes the entire per PV user creation which is not what was desired. Is there a different reasoning at the moment that necessitates this removal?
> > > 
> > > 
> > > Apologies, i wasn't aware of the approach of making user id optional. When nodeserver already has admin credentials, what is the need of creating user per volume? We create the user and use that user, so i m not sure what security issue its fixing. Also, Ramana mentioned that he would be adding subvolume group user create which would have access to a given group. Eventually we would need to use that instead of admin?
> > 
> > 
> > As per my understanding when opening the issue, the intention was to make it optional to improve the performance of VolumeCreate (and Delete), it was not to do away with the per PV secret for users who needed it.
> > I am not aware of why there is a per PV user required or its security implications. I am tagging @liewegas @batrick and @gman0 for comments and direction.
> > Even if this functionality is removed for 1.1, one thing I would like to point out is that for backward compatibility reasons, 1.0 CephFS created volumes would need to use the user created for the same during mount, and so we cannot entirely delete the code.
> 
> WRT backward compatibility, i think the admin user will be able to mount the volumes irrespective of the users that were created for the subvolume. So we may not need backward compatibility code, need to confirm.

Although there is one thing to handle in backward compatibility is to delete user when we delete the old type PVC.

humblec:
@poornimag looks like we need a rebase :)

batrick:
> > > @poornimag the intention is to provide an option to choose between, per PV ID/key generation versus using the admin credentials for the FS to mount the volume, as detailed in issue #269
> > > This PR removes the entire per PV user creation which is not what was desired. Is there a different reasoning at the moment that necessitates this removal?
> > 
> > 
> > Apologies, i wasn't aware of the approach of making user id optional. When nodeserver already has admin credentials, what is the need of creating user per volume? We create the user and use that user, so i m not sure what security issue its fixing. Also, Ramana mentioned that he would be adding subvolume group user create which would have access to a given group. Eventually we would need to use that instead of admin?
>>
> As per my understanding when opening the issue, the intention was to make it optional to improve the performance of VolumeCreate (and Delete), it was not to do away with the per PV secret for users who needed it.

In general, CSI shouldn't be using the **admin** key anyway. That is the veritable keys-to-the-kingdom for Ceph. A CephFS credential with full access to the file system is a starting point:

http://docs.ceph.com/docs/nautilus/cephfs/client-auth/

And that's just for the nodes doing the mounts.

For the CSI driver, it could be using the admin key (because it needs to do `ceph fs volume` calls) but even ideally it should be using a watered down key. In fact, the CSI driver itself doesn't really need access to the file system at all.

@ajarr @liewegas

ajarr:
> 
> In general, CSI shouldn't be using the **admin** key anyway. That is the veritable keys-to-the-kingdom for Ceph. A CephFS credential with full access to the file system is a starting point:
> 
> http://docs.ceph.com/docs/nautilus/cephfs/client-auth/
> 
> And that's just for the nodes doing the mounts.
> 
> For the CSI driver, it could be using the admin key (because it needs to do `ceph fs volume` calls) but even ideally it should be using a watered down key. In fact, the CSI driver itself doesn't really need access to the file system at all.
>

For `fs subvolume/volume` calls, the node would need an auth ID with 
following capabilities,
```
caps mon = "allow r"
caps mgr = "allow rw"
```

For mounting Ceph FS subvolumes within the CSI subvolume group, the node would need an auth ID with following capabilities,
```
caps mon = "allow r"
caps mds = "allow rw path=/volumes/csi"
caps osd  = "allow rw pool=cephfs.data.a"
```
were '/volumes/csi' is the backend path of the subvolume group 'csi', and 'cephfs.data.a' is the data pool of the volume/filesystem within which the subvolumes are created.


The former auth ID should be created beforehand and passed to the CephFS CSI plugin. This will allow the plugin to talk to the ceph-mgr volumes module for provisioning/managing the FS subvolumes. The CephFS CSI plugin issues the following `fs subvolumegroup` command to the ceph-mgr,
```
ceph fs subvolumegroup authorize <auth ID> <volname> <subvolname> --group_name csi 
```
to create the latter auth ID and get it's key, which will be used to mount the FS subvolumes.

ShyamsundarR:
> > > > @poornimag the intention is to provide an option to choose between, per PV ID/key generation versus using the admin credentials for the FS to mount the volume, as detailed in issue #269
> > > > This PR removes the entire per PV user creation which is not what was desired. Is there a different reasoning at the moment that necessitates this removal?
> > > 
> > > 
> > > Apologies, i wasn't aware of the approach of making user id optional. When nodeserver already has admin credentials, what is the need of creating user per volume? We create the user and use that user, so i m not sure what security issue its fixing. Also, Ramana mentioned that he would be adding subvolume group user create which would have access to a given group. Eventually we would need to use that instead of admin?
> > 
> > 
> > As per my understanding when opening the issue, the intention was to make it optional to improve the performance of VolumeCreate (and Delete), it was not to do away with the per PV secret for users who needed it.
> 
> In general, CSI shouldn't be using the **admin** key anyway. That is the veritable keys-to-the-kingdom for Ceph. A CephFS credential with full access to the file system is a starting point:
> 
> http://docs.ceph.com/docs/nautilus/cephfs/client-auth/
> 
> And that's just for the nodes doing the mounts.
> 
> For the CSI driver, it could be using the admin key (because it needs to do `ceph fs volume` calls) but even ideally it should be using a watered down key. In fact, the CSI driver itself doesn't really need access to the file system at all.
> 
> @ajarr @liewegas

Just to be clear the final statement is, we are not supporting per subvolume/csi-volume ID creation (or rather dropping support for the same).

liewegas:
This seems like a fine choice for CSI.  @batrick do you concur?

Of course we still need per-subvolume creds for manila, but that's a different type of environment where you don't have a [semi-]trusted intermediary doing the mounts for you.

batrick:
> were /volumes/csi is the backend path of the subvolume group csi, and cephfs.data.a is the data pool of the volume/filesystem within which the subvolumes are created.

We can be even

> > In general, CSI shouldn't be using the **admin** key anyway. That is the veritable keys-to-the-kingdom for Ceph. A CephFS credential with full access to the file system is a starting point:
> > http://docs.ceph.com/docs/nautilus/cephfs/client-auth/
> > And that's just for the nodes doing the mounts.
> > For the CSI driver, it could be using the admin key (because it needs to do `ceph fs volume` calls) but even ideally it should be using a watered down key. In fact, the CSI driver itself doesn't really need access to the file system at all.
> 
> For `fs subvolume/volume` calls, the node would need an auth ID with
> following capabilities,
> 
> ```
> caps mon = "allow r"
> caps mgr = "allow rw"
> ```
> 
> For mounting Ceph FS subvolumes within the CSI subvolume group, the node would need an auth ID with following capabilities,
> 
> ```
> caps mon = "allow r"
> caps mds = "allow rw path=/volumes/csi"
> caps osd = "allow rw pool=cephfs.data.a"
> ```
> 
> were `/volumes/csi` is the backend path of the subvolume group `csi`, and `cephfs.data.a` is the data pool of the volume/filesystem within which the subvolumes are created.

Keep in mind that the CSI provisioner may be creating volumes in different pools to achieve different QoS/redundancy. I think this should just be:

`caps osd = "allow rw tag cephfs data=cephfs"`

So that the nodes can access any data pool used by the file system using the single credential.

> The former auth ID should be created beforehand and passed to the CephFS CSI plugin. This will allow the plugin to talk to the ceph-mgr volumes module for provisioning/managing the FS subvolumes. The CephFS CSI plugin issues the following `fs subvolumegroup` command to the ceph-mgr,
> 
> ```
> ceph fs subvolumegroup authorize <auth ID> <volname> <subvolname> --group_name csi 
> ```
> 
> to create the latter auth ID and get its key, which will be used to mount the FS subvolumes.

As long as the subvolume module is issuing an `fs authorize` in the backend, that should setup things properly.

batrick:
> This seems like a fine choice for CSI. @batrick do you concur?
> 
> Of course we still need per-subvolume creds for manila, but that's a different type of environment where you don't have a [semi-]trusted intermediary doing the mounts for you.

Right. CSI doesn't need multiple credentials. It just adds complexity. 

Manila definitely will though but only for kernel/ceph-fuse mounts. Eventually when we complete work on the nfs clustering automagic setup, the ganesha daemons will have a pre-configured credential setup by the ceph-mgr. Manila can be blissfully uninvolved.

@tombarron

ajarr:
> This will allow the plugin to talk to the ceph-mgr volumes module for provisioning/managing the FS subvolumes. The CephFS CSI plugin issues the following `fs subvolumegroup` command to the ceph-mgr,
> 
> ```
> ceph fs subvolumegroup authorize <auth ID> <volname> <subvolname> --group_name csi 
> ```
> 
> to create the latter auth ID and get its key, which will be used to mount the FS subvolumes.

@ShyamsundarR , I've implemented this command. The [PR](https://github.com/ceph/ceph/pull/28973) is in review. Would CSI use case require evicting clients that have mounted using the common auth ID?

ShyamsundarR:
> > This will allow the plugin to talk to the ceph-mgr volumes module for provisioning/managing the FS subvolumes. The CephFS CSI plugin issues the following `fs subvolumegroup` command to the ceph-mgr,
> > ```
> > ceph fs subvolumegroup authorize <auth ID> <volname> <subvolname> --group_name csi 
> > ```
> > 
> > 
> > to create the latter auth ID and get its key, which will be used to mount the FS subvolumes.
> 
> @ShyamsundarR , I've implemented this command. The [PR](https://github.com/ceph/ceph/pull/28973) is in review. Would CSI use case require evicting clients that have mounted using the common auth ID?

Do we need this? As the code would **not** be invoking this directly, as we are not creating per subvolume IDs and keys.

Also, the user credentials that are needed to invoke CephFS operations by the CSI plugin, are passed in te required secrets. I would assume generating these credentials is an admin operation, prior to passing the id/key into the secrets and that as long as we document how to create an ID with required permissions (as in #360 ), we should be fine.

What am I missing here?

ajarr:
> > > This will allow the plugin to talk to the ceph-mgr volumes module for provisioning/managing the FS subvolumes. The CephFS CSI plugin issues the following `fs subvolumegroup` command to the ceph-mgr,
> > > ```
> > > ceph fs subvolumegroup authorize <auth ID> <volname> <subvolname> --group_name csi 
> > > ```
> > > 
> > > 
> > > to create the latter auth ID and get its key, which will be used to mount the FS subvolumes.
> > 
> > 
> > @ShyamsundarR , I've implemented this command. The [PR](https://github.com/ceph/ceph/pull/28973) is in review. Would CSI use case require evicting clients that have mounted using the common auth ID?
> 
> Do we need this? As the code would **not** be invoking this directly, as we are not creating per subvolume IDs and keys.
> 
> Also, the user credentials that are needed to invoke CephFS operations by the CSI plugin, are passed in te required secrets. I would assume generating these credentials is an admin operation, prior to passing the id/key into the secrets and that as long as we document how to create an ID with required permissions (as in #360 ), we should be fine.
> 
> What am I missing here?

Agree. The admin (or operator) can create one auth ID with capabilities to issue ceph-mgr fs subvolume commands, and to mount the fs subvolumes. The auth IDs capabilites would be,
```
caps mon = "allow r"
caps mgr = "allow rw"
caps mds = "allow rw path=/volumes/csi"
caps osd = "allow rw tag cephfs data=cephfs "
```
assuming the ceph volume/filesystem name is cephfs.

@poornimag , you've no dependencies on ceph-mgr development. You should be able to move this PR forward.

@ShyamsundarR , do we expect that users would create snapshots of files/directories within the PVs? If so then, the OSD caps needs to be "allow rwp" instead of "allow rw". Then again if users create snapshots within the PV, the PV deletion would fail if snapshots exist within it.

ShyamsundarR:
> @poornimag , you've no dependencies on ceph-mgr development. You should be able to move this PR forward.

@poornimag are we handling backward compat in this patch or is that going to be a separate patch? As discussed earlier, use the created IDs to mount or use the provided credentials to mount, but on deletes the legacy volume IDs need to be cleaned up.

> 
> @ShyamsundarR , do we expect that users would create snapshots of files/directories within the PVs? If so then, the OSD caps needs to be "allow rwp" instead of "allow rw". Then again if users create snapshots within the PV, the PV deletion would fail if snapshots exist within it.

No, the users just get to see an FS that they can operate on, they are not meant to create snapshots of objects etc. within the same. If at all in the future, if the need to do so arises, modifying the caps should be enough to get this working (I never see this happening, but just noting it nevertheless).

ajarr:
> > @ShyamsundarR , do we expect that users would create snapshots of files/directories within the PVs? If so then, the OSD caps needs to be "allow rwp" instead of "allow rw". Then again if users create snapshots within the PV, the PV deletion would fail if snapshots exist within it.
> 
> No, the users just get to see an FS that they can operate on, they are not meant to create snapshots of objects etc. within the same. If at all in the future, if the need to do so arises, modifying the caps should be enough to get this working (I never see this happening, but just noting it nevertheless).

@ShyamsundarR , OK. To clarify, it's the MDS caps that need an additional 'p' flag to set layout and quota, and 's' flag to create snapshots.
http://docs.ceph.com/docs/master/cephfs/client-auth/#layout-and-quota-restriction-the-p-flag

humblec:
> @poornimag , you've no dependencies on ceph-mgr development. You should be able to move this PR forward.

Based on above comment, @poornimag can you please revisit this PR ?

poornimag:
@ShyamsundarR  Backward compatibility is taken care of.

As per the discussion, we would use admin credentials for create/mount/delete of pvc instead of per subvolumegroup user provided by ceph-mgr.

ajarr:
> @ShyamsundarR Backward compatibility is taken care of.
> 
> As per the discussion, we would use admin credentials for create/mount/delete of pvc instead of per subvolumegroup user provided by ceph-mgr.

@poornimag , what discussion? can you please elaborate the reason here to stick to admin credentials?

ajarr:
@batrick I was referring to this comment and its replies,
https://github.com/ceph/ceph-csi/pull/437#issuecomment-511202488

ajarr:
> > @ShyamsundarR Backward compatibility is taken care of.
> > As per the discussion, we would use admin credentials for create/mount/delete of pvc instead of per subvolumegroup user provided by ceph-mgr.
> 
> @poornimag , what discussion? can you please elaborate here on the reason to stick to admin credentials?

OK. Shyam clarified. The admin credentials refers to credentials passed to Ceph-CSI, and not ceph's admin key credentials.

pkg/cephfs/controllerserver.go
```diff
@@ -188,12 +182,12 @@ func (cs *ControllerServer) deleteVolumeDeprecated(req *csi.DeleteVolumeRequest)
...
 	}
 
-	if err = deleteCephUser(&ce.VolOptions, cr, volID); err != nil {
-		klog.Errorf("failed to delete ceph user for volume %s: %v", volID, err)
+	if err = cs.MetadataStore.Delete(string(volID)); err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if err = cs.MetadataStore.Delete(string(volID)); err != nil {
+	if err = deleteCephUserDeprecated(&ce.VolOptions, cr, volID); err != nil {
```
Madhu-1: @poornimag do `cs.MetadataStore.Delete` after userdelete

ajarr:
Great! Thanks @poornimag 

And now @ShyamsundarR we'll get to see how good CephFS CSI is ;-)

Issue #356, opened 2019-05-03
Ceph-csi should expose a prometheus metric for CSI service status


anmolsachan:
Ceph-csi should expose a prometheus metric for its own service status. This metric is required for detecting if the CSI is working and PV and PVC's can be created. If possible PV and PVC creation ability metrics should also be available. Otherwise on a minimum the Ceph-csi service running status metric should be available.

These metrics are required to generate prometheus alerts, in cases where storage provisioning is not possible.

Madhu-1:
there is no way to check PV and PVC can be created.(But I think we can check the ceph backend status, but this also required the storage class and secret to be created prior to the probing)

if we want to check the ceph-csi is running you can just run a sidecar container to send ProbeRequest and check the response.

humblec:
>Otherwise on a minimum the Ceph-csi service running status metric should be available.
>
@anmolsachan I am not sure checking ceph-csi service status is something of interest or in plate of CSI. The reason being, as these are pods, it will be taken care by the kubelet, but definitely there is a possibility for service not available, however its possible to get it from kubelet metrics about the pod status `up` or not. isnt it ?

anmolsachan:
@Madhu-1 The CSI service metric is reqired as a prometheus metric to create alerts. So creating a Sidecar container will a big overhead. Also, in kubernetes each service can give its own metrics, so it will be much better to have from Ceph-csi. 
Regarding PV and PVC if its not possible, its not a major requirement. Would have been a add on to create better alerts to the user.

Daniel-Pivonka:
There are 2 different implementation ideas to add a liveness prometheus metric. 

1.sidecar implementation
2.built into the plugin


There is a similar liveness probe [1] that is part of kubernetes-csi but it just supplies an http endpoint but we want a prometheus endpoint. This implementation creates a sidecar container inside the csi-plugin pod that when an http request happens a grpc request is made to the csi-plugin containers probe function [2] and the liveness metric is returned to the sidecar and then returned to the http request. 

[1] (https://github.com/kubernetes-csi/livenessprobe)
[2] (https://github.com/ceph/ceph-csi/blob/4b33c4a9df0c234d54700ab4e87264d01dcbe547/pkg/csi-common/identityserver-default.go#L51)






[...]

PR #438, opened 2019-06-20
Prometheus liveness probe sidecar

Daniel-Pivonka:
This provides a sidecar container that will expose a Prometheus endpoint containing the liveness probe value as a Prometheus gauge 

## Related issues ##
This addresses the liveness issue in #356 based on approach number 1 described in my comment

## Future concerns ##
This will require an image pushed to quay

### DIFF
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -22,24 +22,28 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/ceph/ceph-csi/pkg/cephfs"
+	"github.com/ceph/ceph-csi/pkg/liveness"
 	"github.com/ceph/ceph-csi/pkg/rbd"
 	"github.com/ceph/ceph-csi/pkg/util"
 	"k8s.io/klog"
 )
 
 const (
-	rbdType    = "rbd"
-	cephfsType = "cephfs"
+	rbdType      = "rbd"
+	cephfsType   = "cephfs"
+	livenessType = "liveness"
 
-	rbdDefaultName    = "rbd.csi.ceph.com"
-	cephfsDefaultName = "cephfs.csi.ceph.com"
+	rbdDefaultName      = "rbd.csi.ceph.com"
+	cephfsDefaultName   = "cephfs.csi.ceph.com"
+	livenessDefaultName = "liveness.csi.ceph.com"
 )
 
 var (
 	// common flags
-	vtype      = flag.String("type", "", "driver type [rbd|cephfs]")
+	vtype      = flag.String("type", "", "driver type [rbd|cephfs|liveness]")
 	endpoint   = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
 	driverName = flag.String("drivername", "", "name of the driver")
 	nodeID     = flag.String("nodeid", "", "node id")
@@ -55,6 +59,12 @@ var (
 	// cephfs related flags
 	volumeMounter = flag.String("volumemounter", "", "default volume mounter (possible options are 'kernel', 'fuse')")
 	mountCacheDir = flag.String("mountcachedir", "", "mount info cache save dir")
+
+	// livenes related flags
+	livenessport = flag.Int("livenessport", 8080, "TCP port for liveness requests")
+	livenesspath = flag.String("livenesspath", "/metrics", "path of prometheus endpoint where metrics will be available")
+	pollTime     = flag.Duration("polltime", time.Second*60, "time interval in seconds between each poll")
+	timeout      = flag.Duration("timeout", time.Second*3, "probe timeout in seconds")
 )
 
 func init() {
@@ -90,6 +100,8 @@ func getDriverName() string {
 		return rbdDefaultName
 	case cephfsType:
 		return cephfsDefaultName
+	case livenessType:
+		return livenessDefaultName
 	default:
 		return ""
 	}
@@ -147,6 +159,9 @@ func main() {
 	case cephfsType:
 		driver := cephfs.NewDriver()
 		driver.Run(dname, *nodeID, *endpoint, *volumeMounter, *mountCacheDir, *instanceID, csipluginPath, cp, driverType)
+
+	case livenessType:
+		liveness.Run(*endpoint, *livenesspath, *livenessport, *pollTime, *timeout)
 
 	default:
 		klog.Fatalln("invalid volume type", vtype) // calls exit
--- a/deploy/cephfs/kubernetes/v1.13/csi-cephfsplugin-provisioner.yaml
+++ b/deploy/cephfs/kubernetes/v1.13/csi-cephfsplugin-provisioner.yaml
@@ -4,13 +4,15 @@ apiVersion: v1
 metadata:
   name: csi-cephfsplugin-provisioner
   labels:
-    app: csi-cephfsplugin-provisioner
+    app: csi-liveness
 spec:
   selector:
     app: csi-cephfsplugin-provisioner
   ports:
-    - name: dummy
-      port: 12345
+    - name: http-metrics
+      port: 8080
+      protocol: TCP
+      targetPort: 8081
 
 ---
 kind: StatefulSet
@@ -97,6 +99,26 @@ spec:
               mountPath: /etc/ceph-csi-config/
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
+        - name: liveness-prometheus
+          image: quay.io/cephcsi/cephcsi:canary
+          args:
+            - "--type=liveness"
+            - "--endpoint=$(CSI_ENDPOINT)"
+            - "--livenessport=8081"
+            - "--livenesspath=/metrics"
+            - "--polltime=60s"
+            - "--timeout=3s"
+          env:
+            - name: CSI_ENDPOINT
+              value: unix:///csi/csi.sock
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+          imagePullPolicy: "IfNotPresent"
       volumes:
         - name: socket-dir
           hostPath:
--- a/deploy/rbd/kubernetes/v1.13/csi-rbdplugin-provisioner.yaml
+++ b/deploy/rbd/kubernetes/v1.13/csi-rbdplugin-provisioner.yaml
@@ -4,13 +4,15 @@ apiVersion: v1
 metadata:
   name: csi-rbdplugin-provisioner
   labels:
-    app: csi-rbdplugin-provisioner
+    app: csi-liveness
 spec:
   selector:
     app: csi-rbdplugin-provisioner
   ports:
-    - name: dummy
-      port: 12345
+    - name: http-metrics
+      port: 8080
+      protocol: TCP
+      targetPort: 8080
 
 ---
 kind: StatefulSet
@@ -110,6 +112,26 @@ spec:
               mountPath: /etc/ceph-csi-config/
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
+        - name: liveness-prometheus
+          image: quay.io/cephcsi/cephcsi:canary
+          args:
+            - "--type=liveness"
+            - "--endpoint=$(CSI_ENDPOINT)"
+            - "--livenessport=8080"
+            - "--livenesspath=/metrics"
+            - "--polltime=60s"
+            - "--timeout=3s"
+          env:
+            - name: CSI_ENDPOINT
+              value: unix:///csi/csi.sock
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+          imagePullPolicy: "IfNotPresent"
       volumes:
         - name: host-dev
           hostPath:
--- a/docs/deploy-rbd.md
+++ b/docs/deploy-rbd.md
@@ -1,3 +1,4 @@
+
 # CSI RBD Plugin
 
 The RBD CSI plugin is able to provision new RBD images and
@@ -36,6 +37,10 @@ make image-cephcsi
 | `--instanceid`      | "default"             | Unique ID distinguishing this instance of Ceph CSI among other instances, when sharing Ceph clusters across CSI instances for provisioning                                   |
 | `--metadatastorage` | _empty_               | Points to where legacy (1.0.0 or older plugin versions) metadata about provisioned volumes are kept, as file or in as k8s configmap (`node` or `k8s_configmap` respectively) |
 | `--pidlimit`        | _0_                   | Configure the PID limit in cgroups. The container runtime can restrict the number of processes/tasks which can cause problems while provisioning (or deleting) a large number of volumes. A value of `-1` configures the limit to the maximum, `0` does not configure limits at all.   |
+| `--livenessport`    | `8080`                | TCP port for liveness requests                                                                                                                                               |
+| `--livenesspath`    | `"/metrics"`          | Path of prometheus endpoint where metrics will be available                                                                                                                  |
+| `--polltime`        | `"60s"`               | Time interval in between each poll                                                                                                                                           |
+| `--timeout`         | `"3s"`                | Probe timeout in seconds                                                                                                                                                     |
 
 **Available volume parameters:**
 
@@ -126,11 +131,11 @@ After successfully completing the steps above, you should see output similar to 
 ```bash
 $ kubectl get all
 NAME                              READY     STATUS    RESTARTS   AGE
-pod/csi-rbdplugin-fptqr           2/2       Running   0          21s
-pod/csi-rbdplugin-provisioner-0   4/4       Running   0          22s
+pod/csi-rbdplugin-fptqr           3/3       Running   0          21s
+pod/csi-rbdplugin-provisioner-0   5/5       Running   0          22s
 
 NAME                                TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)     AGE
-service/csi-rbdplugin-provisioner   ClusterIP   10.104.2.130   <none>        12345/TCP   23s
+service/csi-rbdplugin-provisioner   ClusterIP   10.104.2.130   <none>        8080/TCP   23s
 ...
 ```
 
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,32 @@
+# Metrics
+
+CSI deploys a sidecar container that is responsible for collecting metrics.
+
+## Liveness
+
+Liveness metrics are intended to be collected by prometheus but can be accesesed
+through a GET request to a specific pod ip.
+
+for example
+`curl -X get http://[pod ip]:[liveness-port][liveness-path] 2>/dev/null | grep csi`
+
+the expected output should be
+
+```bash
+[root@worker2 /]# curl -X GET http://10.109.65.142:8080/metrics 2>/dev/null | grep csi
+# HELP csi_liveness Liveness Probe
+# TYPE csi_liveness gauge
+csi_liveness 1
+```
+
+Promethues can be deployed through the promethues operator described [here](https://coreos.com/operators/prometheus/docs/latest/user-guides/getting-started.html).
+The [service-monitor](../examples/service-monitor.yaml) will tell promethues how
+to pull metrics out of CSI.
+
+Each CSI pod has a service to expose the end point to prometheus. By default rbd
+pods run on port 8080 and cephfs 8081.
+These can be changed if desired or if multiple ceph clusters are deployed more
+ports will be used for additional CSI pods.
+
+You may need to open the ports used in your firewall depending on how you
+cluster is setup.
--- a/examples/service-monitor.yaml
+++ b/examples/service-monitor.yaml
@@ -1,0 +1,22 @@
+---
+# This is a servicemonitor that would be used by a prometheus collector to pick
+# up liveness metrics. The label your prometheus instance is looking for will
+# need to be supplied and the namespace may need to changed
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: csi-liveness
+  namespace: rook-ceph
+  labels:
+    team: rook
+spec:
+  namespaceSelector:
+    matchNames:
+      - default
+  selector:
+    matchLabels:
+      app: csi-liveness
+  endpoints:
+    - port: http-metrics
+      path: /metrics
+      interval: 5s
--- a/pkg/liveness/liveness.go
+++ b/pkg/liveness/liveness.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2019 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package liveness
+
+import (
+	"context"
+	"net"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	connlib "github.com/kubernetes-csi/csi-lib-utils/connection"
+	"github.com/kubernetes-csi/csi-lib-utils/rpc"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"k8s.io/klog"
+)
+
+var (
+	liveness = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "csi",
+		Name:      "liveness",
+		Help:      "Liveness Probe",
+	})
+)
+
+func getLiveness(endpoint string, timeout time.Duration) {
+
+	csiConn, err := connlib.Connect(endpoint)
+	if err != nil {
+		// connlib should retry forever so a returned error should mean
+		// the grpc client is misconfigured rather than an error on the network
+		klog.Fatalf("failed to establish connection to CSI driver: %v", err)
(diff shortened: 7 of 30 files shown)


Daniel-Pivonka:
@anmolsachan

travisn:
One thought is that the code for the liveness probe should be included in the same ceph-csi image so that we can avoid adding another image to the mix.

ShyamsundarR:
> One thought is that the code for the liveness probe should be included in the same ceph-csi image so that we can avoid adding another image to the mix.

I agree, makes better sense to keep the code within the ceph-csi repository at least, and as an incarnation of the same image.

Daniel-Pivonka:
liveness side car is now built into the ceph csi image and can be deploy by setting the appropriate flags.

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,85 @@
+package liveness
```
leseb: You need to add the csi author header.

Madhu-1:
@Daniel-Pivonka please let us know when this is ready for review

Daniel-Pivonka:
@Madhu-1 @humblec im facing some issues while running the e2e test locally 

my test run has failed because a csi-rbd-demo-pod is stuck in a terminating state and when the pvc  attached is deleted adn error is throw because the pod still exists. here is the log of the test https://pastebin.com/LTqGCZ3N

I cloned ceph-csi master and ran the e2e test again and got the same result so it seems there is a bug unrelated to my work. here is the log from that https://pastebin.com/LTqGCZ3N

@ShyamsundarR  pointed me to a bug bz-1728848 that seems to be this issue.

Madhu-1:
@Daniel-Pivonka the current E2E doesn't print the plugin logs in case of failure, can you please get  the logs from the rbd daemon set to figure out whats the issue

Daniel-Pivonka:
@Madhu-1 This is ready for review some of the e2e tests are failing because of this issue though #466

Madhu-1:
@Daniel-Pivonka will review it, Thanks

Madhu-1:
the vendor is out of sync

```
# Gopkg.lock is out of sync:
github.com/kubernetes-csi/csi-lib-utils/connection: imported or required, but missing from Gopkg.lock's input-imports
github.com/kubernetes-csi/csi-lib-utils/rpc: imported or required, but missing from Gopkg.lock's input-imports
github.com/prometheus/client_golang/prometheus: imported or required, but missing from Gopkg.lock's input-imports
github.com/prometheus/client_golang/prometheus/promhttp: imported or required, but missing from Gopkg.lock's input-imports
k8s.io/utils/keymutex: in Gopkg.lock's input-imports, but neither imported nor required

# vendor is out of sync:
github.com/kubernetes-csi/csi-lib-utils: hash of vendored tree not equal to digest in Gopkg.lock
github.com/prometheus/client_golang: hash of vendored tree not equal to digest in Gopkg.lock
github.com/prometheus/client_model: hash of vendored tree not equal to digest in Gopkg.lock
github.com/prometheus/common: hash of vendored tree not equal to digest in Gopkg.lock
github.com/prometheus/procfs: hash of vendored tree not equal to digest in Gopkg.lock
```

cmd/cephcsi.go
```diff
@@ -52,6 +55,11 @@ var (
 	// cephfs related flags
 	volumeMounter = flag.String("volumemounter", "", "default volume mounter (possible options are 'kernel', 'fuse')")
 	mountCacheDir = flag.String("mountcachedir", "", "mount info cache save dir")
+
+	// livenes related flags
+	port             = flag.String("port", "8080", "TCP port for listening requests")
+	livenessendpoint = flag.String("livenessendpoint", "/metrics", "endpoint for requests")
+	pollTime         = flag.Int("poll-time", 60, "seconds for between liveness polls")
```
Madhu-1: can you use `flag.Duration`

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
```diff
@@ -13,7 +13,6 @@ spec:
         app: csi-cephfsplugin
     spec:
       serviceAccount: cephfs-csi-nodeplugin
-      hostNetwork: true
```
Madhu-1: why this is removed? removing this cause any issue for rbd or cephfs plugins?

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
```diff
@@ -90,6 +89,21 @@ spec:
               mountPath: /dev
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+        - name: liveness-prometheus
+          image: quay.io/cephcsi/cephcsi:canary
```
Madhu-1: can you follow the order like other containers

* image
* args
* env
* volumeMounts

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
```diff
@@ -90,6 +89,21 @@ spec:
...
+          volumeMounts:
```
Madhu-1: need to add this to helm chats also

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
```diff
@@ -90,6 +89,21 @@ spec:
               mountPath: /dev
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+        - name: liveness-prometheus
```
Madhu-1: liveness will be running as part of ds, not with provisioner stateful set?

cmd/cephcsi.go
```diff
@@ -52,6 +55,11 @@ var (
 	// cephfs related flags
 	volumeMounter = flag.String("volumemounter", "", "default volume mounter (possible options are 'kernel', 'fuse')")
 	mountCacheDir = flag.String("mountcachedir", "", "mount info cache save dir")
+
+	// livenes related flags
+	port             = flag.String("port", "8080", "TCP port for listening requests")
```
Madhu-1: need to update port and livenessendpoint description, can it be more descriptive?

can we change `seconds for between liveness polls `  to   `time interval in seconds between each poll` ?

cmd/cephcsi.go
Madhu-1: need to update the document for these new parameters

cmd/cephcsi.go
Madhu-1: the port can be `flag.Int32`

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"context"
+	"log"
+	"net"
+	"net/http"
+	"time"
+
+	"k8s.io/klog"
+
```
Madhu-1: a new line is not required

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,107 @@
...
+	})
+)
+
+func getLiveness(endpoint string) {
+
+	csiConn, err := connlib.Connect(endpoint)
+	if err != nil {
+		// connlib should retry forever so a returned error should mean
+		// the grpc client is misconfigured rather than an error on the network
+		klog.Infof("failed to establish connection to CSI driver: %v", err)
```
Madhu-1: don't we need to exit if the connection  fails? 
use `klog.Errorf()`

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	log.Printf("calling CSI driver to discover driver name")
```
Madhu-1: use klog.Infof() instead of `log.Printf` in other places  also

humblec:
>@Madhu-1 This is ready for review some of the e2e tests are failing because of this issue though #466
>

As discussed in yesterday's call, I am not sure why its failing only for this PR , especially when this code does not touch that area of the code. Something fishy here. 

Regarding the liveness probe addition in the main container, my understanding was that we were suggesting to keep it in seperate sidecar as it can grow over time. 

@Madhu-1 @ShyamsundarR  @poornimag Are you in agreement to keep it inside the main container ? or do you suggest to keep this in a seperate sidecar container ?  I am fine with any approach but wanted to make sure we have majority on the approach we take here.

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,107 @@
...
+	csiDriverName, err := rpc.GetDriverName(context.Background(), csiConn)
+	if err != nil {
+		klog.Infof("failed to get CSI driver name: %v", err)
+	}
+	log.Printf("CSI driver name: %q", csiDriverName)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
```
Madhu-1: can this time be configurable? will it help in case of poor network connection?

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,107 @@
...
+
+func recordLiveness(endpoint string, pollTime int) {
+	// register prometheus metrics
+	err := prometheus.Register(liveness)
+	if err != nil {
+		klog.Fatalln(err)
+	}
+
+	// get liveness periodically
+	go func() {
```
Madhu-1: use time.Ticker with for loop instead of starting a goroutine

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,107 @@
...
+			// wait to poll metrics again
+			time.Sleep(time.Duration(pollTime) * time.Second)
+		}
+	}()
+}
+
+func Run(endpoint, port, livenessendpoint string, pollTime int) {
+	klog.Infof("Liveness Running")
+	// start liveness collection
+	recordLiveness(endpoint, pollTime)
```
Madhu-1: start this with goroutine

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// start up prometheus endpoint
+	addr := net.JoinHostPort("0.0.0.0", port)
```
Madhu-1: why this is running on `0.0.0.0` ? can we create a [service](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#communicating-with-daemon-pods) and make use of it?

ShyamsundarR:
> > @Madhu-1 This is ready for review some of the e2e tests are failing because of this issue though #466
> Regarding the liveness probe addition in the main container, my understanding was that we were suggesting to keep it in seperate sidecar as it can grow over time.

@humblec are you referring to the container "image" as in, addition of code to the cephcsi image? Or, are you referring to the "container" retaining the Prometheus endpoint advertiser as a sidecar?

The latter is what is done in this PR, hence required the clarification on the comment.

> 
> @Madhu-1 @ShyamsundarR @poornimag Are you in agreement to keep it inside the main container ? or do you suggest to keep this in a seperate sidecar container ? I am fine with any approach but wanted to make sure we have majority on the approach we take here.

Retaining the same image is what I would like to see, as this can reduce the memory footprint on the host as I understand it, and is also easier to maintain versions and dependencies with a single image. This was discussed in this PR as [above](https://github.com/ceph/ceph-csi/pull/438#issuecomment-507661649).

humblec:
>@humblec are you referring to the container "image" as in, addition of code to the cephcsi image?
>

I am referring above, the container image which is `cephcsi image`.

>Retaining the same image is what I would like to see, as this can reduce the memory footprint on the host as I understand it, and is also easier to maintain versions and dependencies with a single image. This was discussed in this PR as above.
>

Thanks for letting me know the choice . In this discussion, I thought we were/are in agreement to have seperate image https://github.com/ceph/ceph-csi/issues/356#issuecomment-505759784 and felt @madhu-1 was also referring the same way. Again if everyone think we can have it in the same image, I am fine. @Madhu-1 can you also share your thought ?

Madhu-1:
> > @humblec are you referring to the container "image" as in, addition of code to the cephcsi image?
> 
> I am referring above, the container image which is `cephcsi image`.
> 
> > Retaining the same image is what I would like to see, as this can reduce the memory footprint on the host as I understand it, and is also easier to maintain versions and dependencies with a single image. This was discussed in this PR as above.
> 
> Thanks for letting me know the choice . In this discussion, I thought we were/are in agreement to have seperate image [#356 (comment)](https://github.com/ceph/ceph-csi/issues/356#issuecomment-505759784) and felt @Madhu-1 was also referring the same way. Again if everyone think we can have it in the same image, I am fine. @Madhu-1 can you also share your thought ?

@humblec including liveness within the image will reduce the image space consumption wherever it gets deployed (as this image will be pulled by daemonset plugin) it will speed up the deployment also.

one more advantage is less maintenance and will go with the normal release process (it always carries the same tag as the cephcsi image even if there is no code changes)

humblec:
It looks like majority is on using the same image, so Am I ? :)
@Daniel-Pivonka please address the review comments, we are good on the approach :)

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
Daniel-Pivonka: This flag causes the plugin pods to share the ip of the node they run on so when both the rbd plugin and the cephfs plugin are running on the same node the liveness sidecars try to open an endpoint but one will fail because the other is already using that port on the ip they share. Removing this causes all pods to get there own ip resolving this problem. Removing it caused no issues in my testing which makes me question why it was there in the first place?

pkg/liveness/liveness.go
Daniel-Pivonka: i have created a Service and ServiceMonitor that rook will deploy to plumb this all into Prometheus.

Daniel-Pivonka:
ive addressed all the issues besides the vendor issue as im not sure how to fix that at the moment.

Madhu-1:
@Daniel-Pivonka can you paste sample output for /metrics endpoint?

Madhu-1:
> ive addressed all the issues besides the vendor issue as im not sure how to fix that at the moment.

one pointer to fix this is deleting `vendor` directory and `Gopkg.lock` and try `dep ensure -v`  

`dep ensure -v` should not give any warning or error

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
Madhu-1: @rootfs @gman0 can you pleaes let us know why `hostNetwork` was added. same for `hostPID` for `rbd` also

Daniel-Pivonka:
vendor should be good and I added a doc about metrics

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+func getLiveness(endpoint string, probeTimeout time.Duration) {
+
+	csiConn, err := connlib.Connect(endpoint)
+	if err != nil {
+		// connlib should retry forever so a returned error should mean
+		// the grpc client is misconfigured rather than an error on the network
+		klog.Errorf("failed to establish connection to CSI driver: %v", err)
+	}
```
Madhu-1: as we are continuing in case of connecting error, will we get the driver name once we get the driver back to running the state?

cmd/cephcsi.go
```diff
@@ -52,6 +56,12 @@ var (
 	// cephfs related flags
 	volumeMounter = flag.String("volumemounter", "", "default volume mounter (possible options are 'kernel', 'fuse')")
 	mountCacheDir = flag.String("mountcachedir", "", "mount info cache save dir")
+
+	// livenes related flags
+	livenessport = flag.Int("liveness-port", 8080, "TCP port for liveness requests")
+	livenesspath = flag.String("liveness-path", "/metrics", "path of prometheus endpoint where metrics will be available")
+	pollTime     = flag.Duration("poll-time", time.Second*60, "time interval in seconds between each poll")
+	probeTimeout = flag.Duration("probe-timeout", time.Second*3, "probe timeout in seconds")
```
Madhu-1: can we call this as just `timeout` as other sidecar containers are doing?

Madhu-1: currently, we don't use `-` in the flags. do we need to deprecate the current flags and add `-` to it (some mechanism to support old flags) (not in this PR)

docs/metrics.md
```diff
@@ -0,0 +1,25 @@
...
+
+Liveness metrics are intended to collected by prometheus but can be acceses through a GET request to a specific node ip.
+
+for example `curl -X get http://[node ip]:[liveness-port][liveness-path] 2>/dev/null | grep csi`
+
+the expected out put should
+
+```bash
+[root@worker2 /]# curl -X GET http://10.109.65.142:8080/metrics 2>/dev/null | grep csi
+# HELP csi_liveness Liveness Probe
```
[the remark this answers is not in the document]
Daniel-Pivonka: The pod IP you make the request to would indicate which which container you are making the request too. This is not the indented use though just showing that it could be done . The intention is for Prometheus to collect the metric and when collected by Prometheus it keeps a record of what pod the metrics came from. I can add an image of how it looks in Prometheus if you'd like.

examples/service-monitor.yaml
```diff
@@ -0,0 +1,40 @@
+---
+# This is a servicemonitor that would be used by a prometheus collector to pick
+# up liveness metrics. The label your prometheus instance is looking for will
+# need to be supplied and the namespace may need to changed
+apiVersion: monitoring.coreos.com/v1
```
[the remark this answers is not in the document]
Daniel-Pivonka: this is intended to be deploy with [ceph monitoring](https://github.com/rook/rook/blob/master/Documentation/ceph-monitoring.md). I've deployed this in kube with no problem following the guide in the doc mentioned.

humblec:
@Daniel-Pivonka Ping ? can you please revisit this PR ? Would like to get this asap.

Daniel-Pivonka:
@humblec @Madhu-1 everything should be all set

Madhu-1:
@Daniel-Pivonka can you provide some output, how this looks in Prometheus UI.

Daniel-Pivonka:
image

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func Run(endpoint, livenessendpoint string, port int, pollTime, timeout time.Duration) {
+	klog.Infof("Liveness Running")
+	// start liveness collection
+	go recordLiveness(endpoint, pollTime, timeout)
+
+	// start up prometheus endpoint
+	addr := net.JoinHostPort("0.0.0.0", strconv.Itoa(port))
```
Madhu-1: instead of `0.0.0.0` can you use below in your deployment template

```
- name: POD_IP
      valueFrom:
        fieldRef:
          fieldPath: status.podIP
```
 to get the pod IP

humblec:
@Daniel-Pivonka please revisit :)

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
```diff
@@ -11,9 +11,9 @@ spec:
     metadata:
       labels:
         app: csi-cephfsplugin
+        contains: liveness
     spec:
       serviceAccount: cephfs-csi-nodeplugin
-      hostNetwork: true
```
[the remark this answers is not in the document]
Daniel-Pivonka: this line originates from one of the very first [commits](https://github.com/ceph/ceph-csi/commit/27334ac60d2782ade8064e07fc384cb3fbde9e8d) from @sbezverk

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,104 @@
...
+
+	csiConn, err := connlib.Connect(endpoint)
+	if err != nil {
+		// connlib should retry forever so a returned error should mean
+		// the grpc client is misconfigured rather than an error on the network
+		klog.Fatalf("failed to establish connection to CSI driver: %v", err)
+	}
+
+	klog.Infof("calling CSI driver to discover driver name")
+	csiDriverName, err := rpc.GetDriverName(context.Background(), csiConn)
```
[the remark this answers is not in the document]
Daniel-Pivonka: This is pulled directly from [kubernetes-csi/livenessprobe](https://github.com/kubernetes-csi/livenessprobe/blob/master/cmd/livenessprobe/main.go).  why are you recommending this?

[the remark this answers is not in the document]
Madhu-1: question is do you want to block the call till you get driver name. 
 note: 
```
context.Background()  returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.
``` 
if yes it looks good

if no see with a timeout.

https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L148
https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/csi-snapshotter/main.go#L151


onside note, as you are not displaying the driver name in Prometheus metrics, you can discard `GetDriverName` itself. just use Probe to check it's alive or not.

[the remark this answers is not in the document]
Daniel-Pivonka: Good point. This information is not needed. Ill remove it.

Daniel-Pivonka:
@Madhu-1  ive put the hostnetworking back in and changed how i am dealing with the ip/port conflict. rbd runs on one port, cephfs runs on a different port. they each get there own service to connect them the prometheus service monitor.

deploy/rbd/kubernetes/v1.13/csi-rbdplugin-provisioner.yaml
```diff
@@ -111,6 +111,26 @@ spec:
               mountPath: /etc/ceph-csi-config/
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
+        - name: liveness-prometheus
+          image: quay.io/cephcsi/cephcsi:canary
+          args:
+            - "--type=liveness"
+            - "--endpoint=$(CSI_ENDPOINT)"
+            - "--livenessport=8080"
```
Madhu-1: just to clarify you will be using host port 8080 for rbd and 8081 for cephfs right?

do we need to enable firewall rules to listen on these ports?

deploy/cephfs/kubernetes/v1.13/csi-cephfsplugin-provisioner.yaml
```diff
@@ -117,3 +137,19 @@ spec:
           emptyDir: {
             medium: "Memory"
           }
+---
+# This is a service to expose the liveness side car
+apiVersion: v1
```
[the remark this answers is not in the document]
Madhu-1: this  applies for both cephfs and rbd

examples/service-monitor.yaml
```diff
@@ -0,0 +1,22 @@
+---
```
Madhu-1: we need to update the document with full setup details on  how users can set up  Prometheus and use it with ceph-csi

note: if not possible in this PR, we can do it in the next PR also. if you want to do it in other PR create an issue so that we won't miss it.

some people are using ceph-csi without rook.

pkg/liveness/liveness.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+func Run(endpoint, livenessendpoint string, port int, pollTime, timeout time.Duration) {
+	klog.Infof("Liveness Running")
+
+	ip := os.Getenv("POD_IP")
+
+	if ip == "" {
+		klog.Error("missing POD_IP env var defaulting to 0.0.0.0")
```
Madhu-1: use `Warning` instead of  error

deploy/cephfs/kubernetes/v1.13/csi-cephfsplugin-provisioner.yaml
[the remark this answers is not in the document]
Daniel-Pivonka: i can do that but will have to make changes to the existing service. it will end up looking like this

> kind: Service
apiVersion: v1
metadata:
  name: csi-cephfsplugin-provisioner
  labels:
    app: csi-liveness
spec:
  selector:
    app: csi-cephfsplugin-provisioner
  ports:
    - name: dummy
      port: 12345
    - name: http-metrics
      port: 8080
      protocol: TCP
      targetPort: 8081

is that ok?

examples/service-monitor.yaml
Daniel-Pivonka: I think a doc on how to setup prometheus is out of the scope of this repo. I will point to the promethues operator documentation on how to use that. https://coreos.com/operators/prometheus/docs/latest/user-guides/getting-started.html

deploy/rbd/kubernetes/v1.13/csi-rbdplugin-provisioner.yaml
Daniel-Pivonka: yes the defualt is 8080 for rbd, 8081 for cephfs. 

If multiple ceph clusters where deployed with multiple instances of csi this would change for the additional sets of rbd and cephfs pods (ex 8082, 8083...)

ill put this in the metrics doc as well as a mention of opening the firewall if needed.

deploy/cephfs/kubernetes/v1.13/csi-cephfsplugin-provisioner.yaml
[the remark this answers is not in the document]
Madhu-1: just remove the dummy port . its of no use.

examples/service-monitor.yaml
Madhu-1: sound good.

Daniel-Pivonka:
@Madhu-1 everything should be good

Madhu-1:
@Daniel-Pivonka you need to update service templates in helm chats

docs/deploy-rbd.md
```diff
@@ -129,8 +134,8 @@ After successfully completing the steps above, you should see output similar to
 ```bash
 $ kubectl get all
 NAME                              READY     STATUS    RESTARTS   AGE
-pod/csi-rbdplugin-fptqr           2/2       Running   0          21s
-pod/csi-rbdplugin-provisioner-0   4/4       Running   0          22s
+pod/csi-rbdplugin-fptqr           3/3       Running   0          21s
+pod/csi-rbdplugin-provisioner-0   5/5       Running   0          22s
 
 NAME                                TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)     AGE
 service/csi-rbdplugin-provisioner   ClusterIP   10.104.2.130   <none>        12345/TCP   23s
```
Madhu-1: service port need to be updated for cephfs and rbd

Daniel-Pivonka:
@Madhu-1  anything else needed?

humblec:
LGTM. Thanks @Daniel-Pivonka !

humblec:
@Daniel-Pivonka can you please rebase ?

Issue #279, opened 2019-03-23
Need to change from using 0 as the initializer value for keymutex based hash locks


ShyamsundarR:
Currently the code uses ["k8s.io/kubernetes/pkg/util/keymutex"](https://github.com/kubernetes/utils/blob/master/keymutex/hashed.go#L30) based hash mutex lock scheme.

All lock initialization pass in a value of 0 for the count of hash buckets (or # mutexes to use). 0 is a special value that keymutex computes as NumCPU and allocates as many hash buckets.

These locks are used for VolName and VolID derived hashes, thus when using only #CPU number of locks, there is a lot of contention or potential contention in create/delete and other requests that acquire the same bucket numbered locks.

The work done within these locks are not CPU bound, as we make one or more CLI calls to create images or directories, or mount before performing certain activities, etc. As a result, using only #CPU number of locks, only increases contention for these locks.

Suggested way forward is to instead decide on a saner default  number of locks, and change all non-CPU bound locks to this value instead.

I suggest we start this out at 128, which is 28 more than [worker threads](https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L57) of the provisioner at present, OR closely set this to the same value as worker threads in our side car deploy yaml.

NOTE: When testing delete volumes in bulk ('bulk' was just 5 deletes simultaneously), one of the deletes hung for more than 9 seconds on a lock. This caused the RPC to time out and send the delete request again, but as elaborated above the delete need not have waited for a lock slot for so long.

gman0:
Recently I've implemented a bit different synchronization scheme for csi-manila which is imho better than either of those approaches...i can post a PR once things settle down (esp. the branches and your PRs)

PR #443, opened 2019-06-24
Move locks to more granular locking than CPU count based

ShyamsundarR:
As detailed in issue #279, current lock scheme has hash
buckets that are count of CPUs. This causes a lot of contention
when parallel requests are made to the CSI plugin. To reduce
lock contention, this commit introduces granular locks per
identifier.

The commit also changes the timeout for gRPC requests to Create
and Delete volumes, as the current timeout is 10s (kubernetes
documentation says 15s but code defaults are 10s). A virtual
setup takes about 12-15s to complete a request at times, that leads
to unwanted retries of the same request, hence the increased
timeout to enable operation completion with minimal retries.

Tests to create PVCs before and after these changes look like so,

Before:
Default master code + sidecar provisioner --timeout option set
to 30 seconds

20 PVCs
Creation: 3 runs, 396/391/400 seconds
Deletion: 3 runs, 218/271/118 seconds
  - Once was stalled for more than 8 minutes and cancelled the run

After:
Current commit + sidecar provisioner --timeout option set to 30 sec
20 PVCs
Creation: 3 runs, 42/59/65 seconds
Deletion: 3 runs, 32/32/31 seconds

Fixes: #279

### DIFF
--- a/deploy/rbd/helm/templates/provisioner-statefulset.yaml
+++ b/deploy/rbd/helm/templates/provisioner-statefulset.yaml
@@ -32,6 +32,8 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--v=5"
+            - "--timeout=60s"
+            - "--retry-interval-start=500ms"
           env:
             - name: ADDRESS
               value: "{{ .Values.socketDir }}/{{ .Values.socketFile }}"
@@ -46,8 +48,8 @@ spec:
           imagePullPolicy: {{ .Values.nodeplugin.plugin.image.pullPolicy }}
           args:
             - "--csi-address=$(ADDRESS)"
-            - "--connection-timeout=15s"
             - "--v=5"
+            - "--timeout=60s"
           env:
             - name: ADDRESS
               value: "{{ .Values.socketDir }}/{{ .Values.socketFile }}"
--- a/deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
+++ b/deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
@@ -35,6 +35,8 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--v=5"
+            - "--timeout=60s"
+            - "--retry-interval-start=500ms"
           env:
             - name: ADDRESS
               value: unix:///csi/csi-provisioner.sock
@@ -46,8 +48,8 @@ spec:
           image: quay.io/k8scsi/csi-snapshotter:v1.1.0
           args:
             - "--csi-address=$(ADDRESS)"
-            - "--connection-timeout=15s"
             - "--v=5"
+            - "--timeout=60s"
           env:
             - name: ADDRESS
               value: unix:///csi/csi-provisioner.sock
--- a/pkg/rbd/controllerserver.go
+++ b/pkg/rbd/controllerserver.go
@@ -116,17 +116,13 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	volumeNameMutex.LockKey(req.GetName())
-	defer func() {
-		if err = volumeNameMutex.UnlockKey(req.GetName()); err != nil {
-			klog.Warningf("failed to unlock mutex volume:%s %v", req.GetName(), err)
-		}
-	}()
-
 	rbdVol, err := cs.parseVolCreateRequest(req)
 	if err != nil {
 		return nil, err
 	}
+
+	idLk := volumeNameLocker.Lock(req.GetName())
+	defer volumeNameLocker.Unlock(idLk, req.GetName())
 
 	found, err := checkVolExists(rbdVol, cr)
 	if err != nil {
@@ -248,12 +244,6 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	if volumeID == "" {
 		return nil, status.Error(codes.InvalidArgument, "empty volume ID in request")
 	}
-	volumeIDMutex.LockKey(volumeID)
-	defer func() {
-		if err := volumeIDMutex.UnlockKey(volumeID); err != nil {
-			klog.Warningf("failed to unlock mutex volume:%s %v", volumeID, err)
-		}
-	}()
 
 	rbdVol := &rbdVolume{}
 	if err := genVolFromVolID(rbdVol, volumeID, cr); err != nil {
@@ -272,12 +262,8 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		// If error is ErrImageNotFound then we failed to find the image, but found the imageOMap
 		// to lead us to the image, hence the imageOMap needs to be garbage collected, by calling
 		// unreserve for the same
-		volumeNameMutex.LockKey(rbdVol.RequestName)
-		defer func() {
-			if err := volumeNameMutex.UnlockKey(rbdVol.RequestName); err != nil {
-				klog.Warningf("failed to unlock mutex volume:%s %v", rbdVol.RequestName, err)
-			}
-		}()
+		idLk := volumeNameLocker.Lock(rbdVol.RequestName)
+		defer volumeNameLocker.Unlock(idLk, rbdVol.RequestName)
 
 		if err := undoVolReservation(rbdVol, cr); err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
@@ -287,12 +273,8 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 
 	// lock out parallel create requests against the same volume name as we
 	// cleanup the image and associated omaps for the same
-	volumeNameMutex.LockKey(rbdVol.RequestName)
-	defer func() {
-		if err := volumeNameMutex.UnlockKey(rbdVol.RequestName); err != nil {
-			klog.Warningf("failed to unlock mutex volume:%s %v", rbdVol.RequestName, err)
-		}
-	}()
+	idLk := volumeNameLocker.Lock(rbdVol.RequestName)
+	defer volumeNameLocker.Unlock(idLk, rbdVol.RequestName)
 
 	// Deleting rbd image
 	klog.V(4).Infof("deleting image %s", rbdVol.RbdImageName)
@@ -341,13 +323,6 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	snapshotNameMutex.LockKey(req.GetName())
-	defer func() {
-		if err = snapshotNameMutex.UnlockKey(req.GetName()); err != nil {
-			klog.Warningf("failed to unlock mutex snapshot:%s %v", req.GetName(), err)
-		}
-	}()
-
 	// Fetch source volume information
 	rbdVol := new(rbdVolume)
 	err = genVolFromVolID(rbdVol, req.GetSourceVolumeId(), cr)
@@ -369,6 +344,9 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	rbdSnap.SizeBytes = rbdVol.VolSize
 	rbdSnap.SourceVolumeID = req.GetSourceVolumeId()
 	rbdSnap.RequestName = req.GetName()
+
+	idLk := snapshotNameLocker.Lock(req.GetName())
+	defer snapshotNameLocker.Unlock(idLk, req.GetName())
 
 	// Need to check for already existing snapshot name, and if found
 	// check for the requested source volume id and already allocated source volume id
@@ -502,13 +480,6 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 		return nil, status.Error(codes.InvalidArgument, "snapshot ID cannot be empty")
 	}
 
-	snapshotIDMutex.LockKey(snapshotID)
-	defer func() {
-		if err = snapshotIDMutex.UnlockKey(snapshotID); err != nil {
-			klog.Warningf("failed to unlock mutex snapshot:%s %v", snapshotID, err)
-		}
-	}()
-
 	rbdSnap := &rbdSnapshot{}
 	if err = genSnapFromSnapID(rbdSnap, snapshotID, cr); err != nil {
 		// if error is ErrKeyNotFound, then a previous attempt at deletion was complete
@@ -518,24 +489,25 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 			return &csi.DeleteSnapshotResponse{}, nil
 		}
 
-		// Consider missing snap as already deleted, and proceed to remove the omap values
+		// All errors other than ErrSnapNotFound should return an error back to the caller
 		if _, ok := err.(ErrSnapNotFound); !ok {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
+
+		// Consider missing snap as already deleted, and proceed to remove the omap values,
+		// safeguarding against parallel create or delete requests against the same name.
+		idLk := snapshotNameLocker.Lock(rbdSnap.RequestName)
+		defer snapshotNameLocker.Unlock(idLk, rbdSnap.RequestName)
+
 		if err = undoSnapReservation(rbdSnap, cr); err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 		return &csi.DeleteSnapshotResponse{}, nil
 	}
 
-	// lock out parallel create requests against the same snap name as we
-	// cleanup the image and associated omaps for the same
-	snapshotNameMutex.LockKey(rbdSnap.RequestName)
-	defer func() {
-		if err = snapshotNameMutex.UnlockKey(rbdSnap.RequestName); err != nil {
-			klog.Warningf("failed to unlock mutex snapshot:%s %v", rbdSnap.RequestName, err)
-		}
-	}()
+	// safeguard against parallel create or delete requests against the same name
+	idLk := snapshotNameLocker.Lock(rbdSnap.RequestName)
+	defer snapshotNameLocker.Unlock(idLk, rbdSnap.RequestName)
 
 	// Unprotect snapshot
 	err = unprotectSnapshot(rbdSnap, cr)
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -62,18 +62,10 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	targetPathMutex.LockKey(targetPath)
-	defer func() {
-		if err = targetPathMutex.UnlockKey(targetPath); err != nil {
-			klog.Warningf("failed to unlock mutex targetpath:%s %v", targetPath, err)
-		}
-	}()
-	disableInUseChecks := false
+	idLk := targetPathLocker.Lock(targetPath)
+	defer targetPathLocker.Unlock(idLk, targetPath)
 
-	volName, err := ns.getVolumeName(req)
-	if err != nil {
-		return nil, err
-	}
+	disableInUseChecks := false
 
 	isBlock := req.GetVolumeCapability().GetBlock() != nil
 	// Check if that target path exists properly
@@ -100,7 +92,13 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	if err != nil {
 		return nil, err
 	}
+
+	volName, err := ns.getVolumeName(req)
+	if err != nil {
+		return nil, err
+	}
 	volOptions.RbdImageName = volName
+
 	// Mapping RBD image
 	devicePath, err := attachRBDImage(volOptions, cr)
 	if err != nil {
@@ -206,13 +204,8 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 		return nil, status.Error(codes.InvalidArgument, "empty volume ID in request")
 	}
 
-	targetPathMutex.LockKey(targetPath)
-
-	defer func() {
-		if err := targetPathMutex.UnlockKey(targetPath); err != nil {
-			klog.Warningf("failed to unlock mutex targetpath:%s %v", targetPath, err)
-		}
-	}()
+	idLk := targetPathLocker.Lock(targetPath)
+	defer targetPathLocker.Unlock(idLk, targetPath)
 
 	notMnt, err := mount.IsNotMountPoint(ns.mounter, targetPath)
 	if err != nil {
--- a/pkg/rbd/rbd_attach.go
+++ b/pkg/rbd/rbd_attach.go
@@ -242,13 +242,8 @@ func attachRBDImage(volOptions *rbdVolume, cr *util.Credentials) (string, error)
 
 	devicePath, found := waitForPath(volOptions.Pool, image, 1, useNBD)
 	if !found {
-		attachdetachMutex.LockKey(imagePath)
-
-		defer func() {
-			if err = attachdetachMutex.UnlockKey(imagePath); err != nil {
-				klog.Warningf("failed to unlock mutex imagepath:%s %v", imagePath, err)
-			}
-		}()
+		idLk := attachdetachLocker.Lock(imagePath)
+		defer attachdetachLocker.Unlock(idLk, imagePath)
 
 		_, err = execCommand("modprobe", []string{moduleName})
 		if err != nil {
--- a/pkg/util/idlocker.go
+++ b/pkg/util/idlocker.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2019 ceph-csi authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"sync"
+)
+
+/*
+IDLock is a per identifier lock with a use counter that retains a number of users of the lock.
+IDLocker is a map of IDLocks holding the IDLocks based on a passed in identifier.
+Typical usage (post creating an IDLocker) is to Lock/Unlock based on identifiers as per the API.
+*/
+type (
+	IDLock struct {
+		mtx      sync.Mutex
+		useCount int
+	}
+
+	IDLocker struct {
+		lMutex sync.Mutex
+		lMap   map[string]*IDLock
+	}
+)
+
+func NewIDLocker() *IDLocker {
+	return &IDLocker{
+		lMap: make(map[string]*IDLock),
+	}
+}
+
+func (lkr *IDLocker) Lock(identifier string) *IDLock {
+	var (
+		lk *IDLock
+		ok bool
+	)
+
+	newlk := new(IDLock)
+
+	lkr.lMutex.Lock()
+
+	if lk, ok = lkr.lMap[identifier]; !ok {
+		lk = newlk
+		lkr.lMap[identifier] = lk
+	}
+	lk.useCount++
+	lkr.lMutex.Unlock()
+
+	lk.mtx.Lock()
+
+	return lk
+}
+
+func (lkr *IDLocker) Unlock(lk *IDLock, identifier string) {
+	lk.mtx.Unlock()
+
+	lkr.lMutex.Lock()
+	lk.useCount--
+	if lk.useCount == 0 {
+		delete(lkr.lMap, identifier)
+	}
+	lkr.lMutex.Unlock()
+}
--- a/deploy/cephfs/helm/templates/provisioner-statefulset.yaml
+++ b/deploy/cephfs/helm/templates/provisioner-statefulset.yaml
@@ -32,6 +32,8 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--v=5"
+            - "--timeout=60s"
+            - "--retry-interval-start=500ms"
           env:
             - name: ADDRESS
               value: "{{ .Values.socketDir }}/{{ .Values.socketFile }}"
--- a/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
+++ b/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
@@ -35,6 +35,8 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--v=5"
+            - "--timeout=60s"
+            - "--retry-interval-start=500ms"
           env:
             - name: ADDRESS
               value: unix:///csi/csi-provisioner.sock
--- a/pkg/cephfs/controllerserver.go
+++ b/pkg/cephfs/controllerserver.go
@@ -25,7 +25,6 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/klog"
-	"k8s.io/utils/keymutex"
 )
 
 // ControllerServer struct of CEPH CSI driver with supported methods of CSI
@@ -41,8 +40,8 @@ type controllerCacheEntry struct {
 }
 
 var (
-	mtxControllerVolumeID   = keymutex.NewHashed(0)
-	mtxControllerVolumeName = keymutex.NewHashed(0)
+	volumeIDLocker   = util.NewIDLocker()
+	volumeNameLocker = util.NewIDLocker()
 )
 
 // createBackingVolume creates the backing subvolume and user/key for the given volOptions and vID,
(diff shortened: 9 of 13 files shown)


ShyamsundarR:
### Setup where the following tests were run:
- Kubernetes 1.14.2 with, 1 master + 3 workers using [k8s-cluster](https://github.com/JohnStrunk/k8s-cluster) based setup.
- Machine had 10 cores with 2 processors each, dedicated to running just this test.
- Vagrant file from k8s-cluster repository was changed to increase the cpu count per vm instance to 4 and memory to 4GiB.

### Performance numbers of master branch without the fix looks like below,
**Table Legend:**
- `PVC Count` is the number of PVC requests create in parallel
- `Runs` is the number of times the benchmark was executed
- `Time to Create (in seconds)/run` is the total time taken to create `PVC Count` per run
- `Time to Delete (in seconds)/run` is the total time taken to delete `PVC Count` per run

| PVC Count | Runs | Time to Create (in seconds)/run | Time to Delete (in seconds)/run |
|---|---|---|---|
| 20 | 3 | 396/391/400 | 218/271/118 |

**NOTE:** Runs for 100 and 200 PVC counts were not performed, as it was taking a long time to complete

### Performance numbers with the provided fix looks like so,

| PVC Count | Runs | Time to Create (in seconds)/run | Time to Delete (in seconds)/run |
|---|---|---|---|
| 20 | 3 | 42/59/65 | 32/32/31 |
| 100 | 3 | 101/112/93 | 85/97/90 |
| 200 | 1 | 148 | 149 |

### Reasons for performance loss with the current locking scheme:
- As the number of locks were very low, and intended for threads consuming CPU resources, each lock was held for a longer duration causing other requests contending for the same lock to queue up.
- Default timeout for gRPC requests (Create/Delete) was set at 10 seconds by the provisioner sidecar. Once a request is queued behind a lock, it accrues the delay of the prior PVC Create/Delete operation in terms of time to respond. This causes the provisioner to timeout and try the operation again after some interval (even though the first call would be processed by the CSI driver).
- The retry duration between calls starts at 1s by default and is increased by a factor of 2 each time, hence when reaching retry counts of 5-6 the interval of wait by the provisioner itself is 16s-32s and so on. This, combined with the queuing and response delays causes larger time to completion of the request.
- Finally, on the setup where the tests were executed, a single Create could take 8-15s to complete (similarly a delete could take about as much time). (This was one of the key reasons to increase the timeout in the sidecars, to accommodate slower setups)

### Reasons for choosing current locking scheme:
- Provided granular locking at the cost of a some additional CPU cycles, and CPU was not a bottleneck in the problem. Thus, addressed corner cases where lock hashes may even collide to give a lower system performance.
- Cost to check a Created or Deleted volume is lesser than creating one or deleting one from scratch, hence retries were not much of a concern (yet)
  - Other optimization possibilities exist in the future to further reduce checking Created or Deleted volumes (pool/fs name and ID caching)
- Increasing the provisioner timeout helps reducing retires, thus not (yet) needing a response broker like approach

### What were the options to improve the locking scheme:

1) Just move the number of locks in the initialization of keymutex.NewHashed() from 0 to 128

This was attempted and the performance numbers were no different than the current scheme of granular locks

Numbers generated using this scheme (including setting provisioner timeout to 30s) is as follows,

| PVC Count | Runs | Creates/run/sec | Deletes/run/sec |
|---|---|---|---|
| 20 | 3 | 50/72/33 | 32/99/37 |
| 100 | 3 | 116/94/112 | 95/71/93 |
| 200 | 1 | 162 | 145 |

This scheme still can potentially have RPC requests queuing behind the same lock on hash collision, and hence, although simpler, is not used as the solution.

2) Use [responsebroker](https://github.com/kubernetes/cloud-provider-openstack/tree/master/pkg/csi/manila/responsebroker) as in Manila-CSI

Cost to check a volume is provisioned or already deleted, is lower than the initial create/delete operation and has other optimizations in the future (as noted earlier). Also, tweaking the provisioner timeouts, reduces the duplicate requests in the pipeline, avoiding the leverage a response cache provides at present.

Further response broker currently does not check exact request equivalence, and depends on request name equivalence only (just a note for the future)


3) Use golang [sync.Map](https://golang.org/pkg/sync/#Map) and possibly sync.Pool to optimize CPU utilization

These are along the lines of solutions like,
- https://github.com/enfipy/locker/blob/master/locker.go
- https://github.com/vburenin/nsync/blob/master/namedmutex.go

But are not chosen as these seem more premature optimizations based on current results using regular maps and locked access to the same.

dillaman:
@ShyamsundarR Are those create/delete times the total time to create/delete that many PVs or is that the average time per PV? I ask because creating 50 PVs in under a second is fine, but creating 20 PVs in 50 seconds is terrible.

ShyamsundarR:
> @ShyamsundarR Are those create/delete times the total time to create/delete that many PVs or is that the average time per PV? I ask because creating 50 PVs in under a second is fine, but creating 20 PVs in 50 seconds is terrible.

Total time to create as many PVCs. IOW, it took 396 seconds to create 20 PVCs using existing master code. Does this explain the numbers?

Is there a better way to word this part of the table legend? "`Time to Create (in seconds)/run` is the total time taken to create `PVC Count` per run"

I do not want to state per PVC create time as 396/20, as the creation is/was not serial, but parallel in nature.

dillaman:
Yes, that explains the numbers. I'm just concerned about the speed since it seems really terrible (albeit way better thanks to this PR).

For comparison, here is me creating 20 1GiB images on the command-line in roughly 1 second total time:

> $ time bash -c 'for x in `seq 1 20`; do echo $x ; ./bin/rbd create --image-feature layering image$x --size 1G; done'
> 1
> 2
> 3
> 4
> 5
> 6
> 7
> 8
> 9
> 10
> 11
> 12
> 13
> 14
> 15
> 16
> 17
> 18
> 19
> 20
> 
> real	0m1.016s
> user	0m0.834s
> sys	0m0.184s
>

ShyamsundarR:
> Yes, that explains the numbers. I'm just concerned about the speed since it seems really terrible (albeit way better thanks to this PR).
> 
> For comparison, here is me creating 20 1GiB images on the command-line in roughly 1 second total time:
> 
> > $ time bash -c 'for x in `seq 1 20`; do echo $x ; ./bin/rbd create --image-feature layering image$x --size 1G; done'
> > 1
<snip>
> > 20
> > real	0m1.016s
> > user	0m0.834s
> > sys	0m0.184s

In the setup where I ran the tests the above takes,
```

File: pkg/util/idlocker.go [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+
+func NewIDLocker() *IDLocker {
+	return &IDLocker{
+		lMap: make(map[string]*IDLock),
+	}
+}
+
+func (lkr *IDLocker) Lock(identifier string) *IDLock {
+	var lk *IDLock
+	newlk := new(IDLock)
```
dillaman: Nit: perhaps this is just a golang limiation, but why not only create the new lock if the lock doesn't exist (i.e. move to line 53)?

ShyamsundarR: :) this is my premature optimization of reducing work within the lock, either works and is not because of a golang limitation.

dillaman: Fair enough. :+1:

humblec:
@ShyamsundarR first of all thanks for this PR! 

I have some general comments on this : That said, this PR mix the "timeout" of request via `sidecar timeout argument and lock improvement`. afacit, the performance numbers are captured with `both` are in place. If possible, I would like to split this and validate where it stand. The reason being, more granular level of analysis or fix is required especially dealing with performance numbers as the result can vary depending on the request type/setup..etc. 
This also helps us to understand or validate the response broker type implementation which @gman0 did in manilla helps us to tackle the issues we could encounter.  

Also, a validation with ONLY one type of request ( for ex: either create or delete) while measuring performance may not help much as shown in PR description/comment. We have to inject say "creation" , "deletion", "snapshot" ..etc in parallel to really see how we behave on lock holding and releasing...etc. Because thats one of the very possible scenario in a cloud or kubernetes kind of setup.  


I can help to split the experiments or code, but IMO, we have to understand above scenarios before we make some changes in this area as its a critical path for serving requests.

mykaul:
How many PVs did we end up having in the system? It's unclear to me. If we are doing 200 in parallel, and it took us 162 seconds overall to complete the initial 200, I assume there were many more created in parallel, right? To me it'd be easier to understand something like:
"To create 1000 PVs in total with 200 PVC requests in parallel took X seconds to complete" and "To create 1000 PVs in total with 10 PVC requests in parallel took X seconds to complete" kind of measurements.

ShyamsundarR:
> @ShyamsundarR first of all thanks for this PR!
> 
> I have some general comments on this : That said, this PR mix the `sidecar timeout argument and lock improvement`. afacit, the performance numbers are captured with `both` are in place. If possible, I would like to split this and validate where it stand invidually. The reason being, more granular level of analysis or fix is required especially dealing with performance numbers as the result can vary depending on the request type/setup..etc.
> This also helps us to understand or validate the response broker type implementation which @gman0 did in manilla to tackle the issues we could encounter.
> 
> Also, a validation with ONLY one type of request ( for ex: either create or delete) while measuring performance may not help much . We have to inject say "creation" , "deletion", "snapshot" ..etc in parallel to really see how we behave on lock holding and releasing...etc. Because thats one of the very possible scenario in a cloud or kubernetes kind of setup.
> 
> I can help to split the experiments or code, but IMO, we have to understand above scenarios before we make some changes in this area as its a critical path for serving requests.

@humblec My intention was to focus on the single lock that we are contending on and fix the same.

Also, during the fix the issue of response time per request was causing a problem, which is why I moved the side car timeouts as well.

The lock is granular, so combining operations will not impact each other with the fix.

Snapshots have a whole different problem as noted in this [comment](https://github.com/ceph/ceph-csi/issues/446#issuecomment-506096894)

ShyamsundarR:
> How many PVs did we end up having in the system? It's unclear to me. If we are doing 200 in parallel, and it took us 162 seconds overall to complete the initial 200, I assume there were many more created in parallel, right?

I created the `PVC Count` number of PVCs (as in the tables provided) in parallel and measured the time it takes to complete the same.

scripts to create and delete PVCs are as below:
[create-perf.sh.txt](https://github.com/ceph/ceph-csi/files/3334431/create-perf.sh.txt)
[delete-perf.sh.txt](https://github.com/ceph/ceph-csi/files/3334432/delete-perf.sh.txt)

> To me it'd be easier to understand something like:
> "To create 1000 PVs in total with 200 PVC requests in parallel took X seconds to complete" and "To create 1000 PVs in total with 10 PVC requests in parallel took X seconds to complete" kind of measurements.

The above was not the test that I performed, hence cannot summarize the results as such. What can be summarized from the provided numbers is "To create `PVC Count` PVCs in parallel it took `Time to create(in seconds)` time"

humblec:
>Also, during the fix the issue of response time per request was causing a problem, which is why I moved the side car timeouts as well.
>
@ShyamsundarR my point is, lets split or track the timeout fix separately and may be do a validation whether that helps in performance or to avoid some contention.

ShyamsundarR:
> > Also, during the fix the issue of response time per request was causing a problem, which is why I moved the side car timeouts as well.
> 
> @ShyamsundarR my point is, lets split or track the timeout fix separately and may be do a validation whether that helps in performance or to avoid some contention.

I have captured the reasons in text in this [comment]() under the titles,
- "Reasons for performance loss with the current locking scheme"
- "Reasons for choosing current locking scheme"

But I guess you need numbers instead of text. Let me see what I can generate in my day and post it here.

humblec:
>The commit also changes the timeout for gRPC requests to Create
and Delete volumes, as the current timeout is 10s (kubernetes
documentation says 15s but code defaults are 10s). A virtual
setup takes about 12-15s to complete a request at times, that leads
to unwanted retries of the same request, hence the increased
timeout to enable operation completion with minimal retries.
>

@ShyamsundarR I am trying to figure out "only increasing" the timeout helps to improve the performance or not ? so I was asking can we see how the performance looks `only with the "timeout" change.`

ShyamsundarR:
@humblec Here are the requested numbers.

Scripts used:
[create-perf.sh.txt](https://github.com/ceph/ceph-csi/files/3335502/create-perf.sh.txt)
[delete-perf.sh.txt](https://github.com/ceph/ceph-csi/files/3335503/delete-perf.sh.txt)

### Only timeout changes against master code (IOW, just adding the timeout options to code `--timeout=30s` `--retry-interval-start=500ms`)

NOTE: Only a single run as time taken is far longer
20 PVC created in parallel took 544 seconds
20 PVC deleted in parallel took 194 seconds

Notes:
- Time for create got stuck in the last 5 PVCs as I could see then contending for the same lock, from the logs, and with exponential backoff in Kube.

Creation rate looked as below

./create-perf.sh 20
**Creation Start time:
Thu Jun 27 11:15:28 EDT 2019**
persistentvolumeclaim/rbd-pvc-iter-20 created
persistentvolumeclaim/rbd-pvc-iter-19 created
persistentvolumeclaim/rbd-pvc-iter-18 created
persistentvolumeclaim/rbd-pvc-iter-17 created
persistentvolumeclaim/rbd-pvc-iter-16 created
persistentvolumeclaim/rbd-pvc-iter-15 created
persistentvolumeclaim/rbd-pvc-iter-14 created
persistentvolumeclaim/rbd-pvc-iter-13 created
persistentvolumeclaim/rbd-pvc-iter-12 created
persistentvolumeclaim/rbd-pvc-iter-11 created
persistentvolumeclaim/rbd-pvc-iter-10 created
persistentvolumeclaim/rbd-pvc-iter-9 created
persistentvolumeclaim/rbd-pvc-iter-8 created
persistentvolumeclaim/rbd-pvc-iter-7 created
persistentvolumeclaim/rbd-pvc-iter-6 created
persistentvolumeclaim/rbd-pvc-iter-5 created
persistentvolumeclaim/rbd-pvc-iter-4 created
persistentvolumeclaim/rbd-pvc-iter-3 created
persistentvolumeclaim/rbd-pvc-iter-2 created
persistentvolumeclaim/rbd-pvc-iter-1 created
Done moved up 1
Thu Jun 27 11:15:33 EDT 2019
Done moved up 2
Thu Jun 27 11:15:45 EDT 2019
Done moved up 4
Thu Jun 27 11:15:46 EDT 2019
Done moved up 5
Thu Jun 27 11:15:48 EDT 2019
Done moved up 6
Thu Jun 27 11:16:00 EDT 2019
Done moved up 7
Thu Jun 27 11:16:01 EDT 2019
Done moved up 8
Thu Jun 27 11:16:06 EDT 2019
Done moved up 9
Thu Jun 27 11:16:07 EDT 2019
Done moved up 10
Thu Jun 27 11:16:12 EDT 2019
Done moved up 11
Thu Jun 27 11:16:40 EDT 2019
Done moved up 13
Thu Jun 27 11:16:41 EDT 2019
Done moved up 14
Thu Jun 27 11:16:51 EDT 2019
**Done moved up 15
Thu Jun 27 11:16:53 EDT 2019**

**NOTE:** First 15 got created in 85 seconds

Done moved up 16
Thu Jun 27 11:22:01 EDT 2019
Done moved up 17
Thu Jun 27 11:22:25 EDT 2019
Done moved up 19
Thu Jun 27 11:24:31 EDT 2019
Done moved up 20
Thu Jun 27 11:24:32 EDT 2019
Creation End time:
Thu Jun 27 11:24:32 EDT 2019

**NOTE:** last 5 took the rest of the time as they were contending on the same lock.

ShyamsundarR:
Further data on only using locks and no changes to timeouts.

### Only granular locks and notimeout changes with the current PR (although this was not requested)
20 PVC created in parallel took 139/102/137 seconds in each create run (3 in total)
200 PVC created in parallel took 224 seconds (single run)

20 PVC deleted in parallel took 38/60/98 seconds in each delete run (3 in total)
200 PVC deleted in parallel took188 seconds (single run)

deploy/rbd/helm/templates/provisioner-statefulset.yaml
```diff
@@ -46,8 +48,8 @@ spec:
           imagePullPolicy: {{ .Values.nodeplugin.plugin.image.pullPolicy }}
           args:
             - "--csi-address=$(ADDRESS)"
-            - "--connection-timeout=15s"
             - "--v=5"
+            - "--timeout=60s"
```
Madhu-1: let's keep the timeout constant in all sidecars.

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
```diff
@@ -32,6 +32,8 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--v=5"
+            - "--timeout=30s"
```
Madhu-1: check attacher sidecar also  need timeout option

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
```diff
@@ -43,8 +45,8 @@ spec:
           image: quay.io/k8scsi/csi-snapshotter:v1.1.0
           args:
             - "--csi-address=$(ADDRESS)"
-            - "--connection-timeout=15s"
             - "--v=5"
+            - "--timeout=60s"
```
Madhu-1: note: snapshot default timeout  is 1 minute

https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/csi-snapshotter/main.go#L54

pkg/rbd/controllerserver.go
```diff
@@ -272,12 +262,8 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		// If error is ErrImageNotFound then we failed to find the image, but found the imageOMap
 		// to lead us to the image, hence the imageOMap needs to be garbage collected, by calling
 		// unreserve for the same
-		volumeNameMutex.LockKey(rbdVol.RequestName)
-		defer func() {
-			if err := volumeNameMutex.UnlockKey(rbdVol.RequestName); err != nil {
-				klog.Warningf("failed to unlock mutex volume:%s %v", rbdVol.RequestName, err)
-			}
-		}()
+		idLk := volumeNameLocker.Lock(rbdVol.RequestName)
```
Madhu-1: can we lock the resource before fetching volume info from rados omap?
as a corner case if there is a volume  request with the same name created at the same we fetch the same info may clash, better to take a lock and fetch  omap

pkg/rbd/controllerserver.go
```diff
@@ -502,33 +480,27 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
...
 	if err = genSnapFromSnapID(rbdSnap, snapshotID, cr); err != nil {
-		// Consider missing snap as already deleted, and proceed to remove the omap values
+		// All errors other than ErrSnapNotFound should return an error back to the caller
 		if _, ok := err.(ErrSnapNotFound); !ok {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
+
+		// Consider missing snap as already deleted, and proceed to remove the omap values,
+		// safeguarding against parallel create or delete requests against the same name.
+		idLk := snapshotNameLocker.Lock(rbdSnap.RequestName)
```
Madhu-1: same comment as delete volume

pkg/util/idlocker.go
```diff
@@ -0,0 +1,73 @@
...
+		useCount int
+	}
+
+	IDLocker struct {
+		lMutex sync.Mutex
+		lMap   map[string]*IDLock
+	}
+)
+
+func NewIDLocker() *IDLocker {
```
Madhu-1: can this a function be renamed as New? IMO the call should not be `idlocker.NewIDLocker()` it should be `idlocker.New()`

pkg/util/idlocker.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+func (lkr *IDLocker) Lock(identifier string) *IDLock {
+	var lk *IDLock
+	newlk := new(IDLock)
+
+	lkr.lMutex.Lock()
+	lk = lkr.lMap[identifier]
+	if lk == nil {
```
Madhu-1: use  ok to check key present in map
```
if lk, ok = lkr[identifier]; !ok {
          lk = newlk
          lkr.lMap[identifier] = lk
}
```

pkg/util/idlocker.go
Madhu-1: IMO a map with key as identifier and value as mutex will be enough? the consumer will get the WriteLock during operation and unlock it once it's done.

pkg/util/idlocker.go
```diff
@@ -0,0 +1,73 @@
...
+		lk = newlk
+		lkr.lMap[identifier] = lk
+	}
+	lk.useCount++
```
Madhu-1: am failing to understand what's the use of `useCount`?

pkg/util/idlocker.go
Madhu-1: @ShyamsundarR can we use sync.Map inbuilt function?

pkg/rbd/controllerserver.go
```diff
@@ -370,6 +345,9 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	rbdSnap.SourceVolumeID = req.GetSourceVolumeId()
 	rbdSnap.RequestName = req.GetName()
 
+	idLk := snapshotNameLocker.Lock(req.GetName())
```
Madhu-1: don't we need to take a lock on parent volume name? we may end up in race condition (volume delete and snap create)

humblec: Yeah, as discussed in one of the previous PRs (iic stateless), there is a race scenario we could encounter.

pkg/util/idlocker.go
Madhu-1: now got it. with the use of `useCount` which helps in avoiding the key deleting from the map (any performance improvement)?

humblec:
@ShyamsundarR thanks for fetching the details separately. What I see here is, with just `timeout` change the operations are faster for first 15 requests, from there on, it crawl. Not sure how that happens , ie if the lock is good enough for "15" parallel requests, I was expecting it should work for next set too :). Btw, the `create` or `delete script` introduce a `sleep` in each iteration, so I am not sure we can call these samples as 'parallel` requests.

deploy/rbd/helm/templates/provisioner-statefulset.yaml
ShyamsundarR: The reason is pretty simple, the snapshotter side car default is already 60s, but the option line states this as 10s (see [this](https://github.com/kubernetes-csi/external-snapshotter/blob/ac33959738598e55a0a0a9d5a2d08d7445f27f36/cmd/csi-snapshotter/main.go#L69)). I wanted to retain it at 60s for now even if the snapshotter side car corrects the same. Also, the snapshot is attempted only once, and not retried on timeouts, hence higher timeouts for the same.

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
ShyamsundarR: Attacher does not call into CSI yet, as we do not have ControllerPublishVolume support. Hence adding a timeout for the same is not required.

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
ShyamsundarR: Yes, please see above comment on default disagreeing with the text for the option: "The timeout for any RPCs to the CSI driver. Default is 10s." and hence ensuring that this stays even if the snapshotter gets fixed one way or the other.

deploy/rbd/helm/templates/provisioner-statefulset.yaml
Madhu-1: the comment looks wrong. the default value is 1 minute, yes I agree, I would  suggest keeping all  sidecar container  timeout  to  60sec

Madhu-1: comment already fixed in master https://github.com/kubernetes-csi/external-snapshotter/blob/master/cmd/csi-snapshotter/main.go#L69

pkg/util/idlocker.go
ShyamsundarR: Right, UseCount is to delete when there are no more users.

Performance improvement as in due to UseCount or key deletion? IMO none.

pkg/rbd/rbd_attach.go
```diff
@@ -241,13 +241,8 @@ func attachRBDImage(volOptions *rbdVolume, cr *util.Credentials) (string, error)
...
 	devicePath, found := waitForPath(volOptions.Pool, image, 1, useNBD)
 	if !found {
-		attachdetachMutex.LockKey(imagePath)
-
-		defer func() {
-			if err = attachdetachMutex.UnlockKey(imagePath); err != nil {
-				klog.Warningf("failed to unlock mutex imagepath:%s %v", imagePath, err)
-			}
-		}()
+		idLk := attachdetachLocker.Lock(imagePath)
```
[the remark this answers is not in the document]
ShyamsundarR: (same comment here) I did an exercise to understand the need for this mutex, and was not comfortable removing it or retaining it as I was unsure. If there is sufficient reasoning to not need it please send a PR to remove the same.

pkg/rbd/nodeserver.go
```diff
@@ -206,13 +204,8 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
 	}
 
-	targetPathMutex.LockKey(targetPath)
-
-	defer func() {
-		if err := targetPathMutex.UnlockKey(targetPath); err != nil {
-			klog.Warningf("failed to unlock mutex targetpath:%s %v", targetPath, err)
-		}
-	}()
+	idLk := targetPathLocker.Lock(targetPath)
```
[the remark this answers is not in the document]
ShyamsundarR: I did an exercise to understand the need for this mutex, and was not comfortable removing it or retaining it as I was unsure. If there is sufficient reasoning to not need it please send a PR to remove the same.

ShyamsundarR:
> @ShyamsundarR thanks for fetching the details separately. What I see here is, with just `timeout` change the operations are faster for first 15 requests, from there on, it crawl. Not sure how that happens , ie if the lock is good enough for "15" parallel requests, I was expecting it should work for next set too :). Btw, the `create` or `delete script` introduce a `sleep` in each iteration, so I am not sure we can call these samples as 'parallel` requests.

If you are referring to the scripts in [this](https://github.com/ceph/ceph-csi/pull/443#issuecomment-506400488) comment, then the sleep (of 1 second) is to check how many PVCs are in the bound state (post create issued) and how many PVs are deleted (post delete issued). This happens post issuing the creates or deletes in a tight loop above and does not impede the parallelism.

pkg/util/idlocker.go
ShyamsundarR: For sync.Map usage, please see point (3) in section "What were the options to improve the locking scheme" in [this](https://github.com/ceph/ceph-csi/pull/443#issuecomment-505129896) comment. I consider this a premature optimization as we are not a CPU bound task in between the locks, hence this should not be required.

ShyamsundarR: > can this a function be renamed as New? IMO the call should not be `idlocker.NewIDLocker()` it should be `idlocker.New()`

The call is `NewIDLocker()` and not `idlocker.NewIDLocker()`, IOW we do not have an `idlocker` to start with and are asking for a new instance of the same.

The pattern is inspired by this section in effective go: https://golang.org/doc/effective_go.html#composite_literals

I do not see the problem, am I looking at it wrong?

deploy/rbd/helm/templates/provisioner-statefulset.yaml
ShyamsundarR: What is required finally? 60s everywhere? That should be fine as well.

Madhu-1: yes please make it 60s in other places too

pkg/util/idlocker.go
ShyamsundarR: I will have to declare the `ok` variable ahead as well, as otherwise there would be a shadow declaration of `lk` if I use `if lk, ok := ...`, so the reasoning was to not create one. Is this mandatory to use `ok` to check?

pkg/rbd/controllerserver.go
ShyamsundarR: (relevant to DeleteSnapshot as well, hence not reposting against the same)

We need `requestName` to lock, that is fetched from the RADOS OMaps, hence we cannot lock before fetching the information.

We hence need to analyze, is the post lookup locking of OMap safe.

1: Failing to fetch the UUID OMap during lookup
 - This means that the UUID is deleted, so we respond with an ack, not a problem

2: Fetching the incorrect UUID Omap
- We fetch the UUID OMap information, based on the UUID encoded on the `VolumeID`, the CO can only have this `VolumeID` if there was a prior successful `CreateVolume` call
- In the `CreateVolume` the UUID is generated to ensure there are no clashes with existing UUIDs on the pool, IOW we go beyond the usual "UUIDs are unique enough statistically" statements during its generation, to ensure it's uniqueness.
- So we will not fetch an incorrect UUID, and we further check if the UUID has the same `requestName` as the request name directory

3: Fetching an in flight UUID OMap
This case is slightly different from (2) and is a **problem**.
- Assume a `VolID` was just deleted by the CO, but the `DeleteVolume` response timed out at the CO, hence the CO would replay he same `VolID` for deletion
- Later a `CreateVolume` (for a different `RequestName`) used the just deleted UUID as it's UUID (this call can complete or be in-flight for the rest of the discussion)

Now, the next `DeleteVolume` for the older `VolID` would garbage collect the new UUID based RADOS entries and images.

The `DeleteVolume` could get into a situation where it finds the OMaps but the image still missing (as `CreateVolume` that is racing with this, and has not created the image yet) and only garbage collect the RADOS entries, this needs to be protected using an **UUID lock**, so that we lookup/create without conflicts.

BUT (and this is a big one), `DeleteVolume` will still garbage collect the new UUID, as once the Create is complete all entries align (`RequestName` OMAP and UUID OMap and image etc.). What this essentially means is that we cannot reuse a UUID as we cannot be sure `DeleteVolume` **caller** ever processed a success return from the plugin.

The ONLY resolution for the same would be to retain the RADOS UUID OMaps so that these are never reused (space consumption issues?), hence VolIDs are always unique? Or, never garbage collect the `RequestName` in the CSI directories OMap to ensure the the same request name gets the same UUID, slightly better but assumes request names are reused (which is not true in kubernetes as it uses a UUID for the request names anyway).

Also, it is interesting how other CSI vendors handle the same, as the problem is outside the scope of Ceph-CSI in reality, and should impact other implementations as well.

pkg/util/idlocker.go
Madhu-1: oops sorry got confused with filename and package name. the package name is `util` this should be fine

pkg/rbd/controllerserver.go
ShyamsundarR: > don't we need to take a lock on parent volume name? we may end up in race condition (volume delete and snap create)

`genVolFromVolID` may pass or fail, if it fails (due to parent image not found or under deletion) the snapshot call fails immediately. If it passes the parent information is used to create the snapshot in `createSnapshot` call, and that passes the parent name to rbd for snapshot creation.

If the parent does not exist at this point in time, the snapshot will fail again, and required cleanup would be done.

If the delete on the parent is issued when a snap create is in flight to `rbd` that is something RBD would ensure to order and allow or disallow.

So there is no real reason for the plugin to lock and protect the parent from parallel calls, `rbd` will error out if the parent is missing, or handle a delete when a snap is in flight.

and tagging @dillaman to keep me honest around rbd statements made.

pkg/rbd/controllerserver.go
ShyamsundarR: Leaving my thought on the UUID collision problem I alluded to in the previous comment:
- We need to depend on [UUID collision](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions) properties to satisfy this requirement, as otherwise we need to remember and not use any older UUID that was ever used
- We run a single instance of the provisioner plugin, hence we are at least not generating UUIDs across different systems for time stamp based or MAC address based collision probabilities to increase
- If we intend to solve this, it also has to be understood why others do not have this issue or avoid it (i.e never reuse an older `VolID`)

**IOW, I think this problem can be left unsolved.**

If there are alternative thoughts on resolving this do share.

mykaul:
How many PVs did we end up having in the system? It's unclear to me. If we are doing 200 in parallel, and it took us 162 seconds overall to complete the initial 200, I assume there were many more created in parallel, right? To me it'd be easier to understand something like:
"To create 1000 PVs in total with 200 PVC requests in parallel took X seconds to complete" and "To create 1000 PVs in total with 10 PVC requests in parallel took X seconds to complete" kind of measurements.

ShyamsundarR:
> How many PVs did we end up having in the system? It's unclear to me. If we are doing 200 in parallel, and it took us 162 seconds overall to complete the initial 200, I assume there were many more created in parallel, right? To me it'd be easier to understand something like:
> "To create 1000 PVs in total with 200 PVC requests in parallel took X seconds to complete" and "To create 1000 PVs in total with 10 PVC requests in parallel took X seconds to complete" kind of measurements.

@mykaul looks like you reposted your comment, my original answer to this is [here](https://github.com/ceph/ceph-csi/pull/443#issuecomment-506309372)

deploy/rbd/helm/templates/provisioner-statefulset.yaml
ShyamsundarR: Updated in the latest push.

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
ShyamsundarR: Assuming the reasoning provided is good, marking this resolved.

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
ShyamsundarR: Resolved, as per discussion on the timeout comment.

pkg/util/idlocker.go
Madhu-1: this is the way we check key is present in  map

pkg/util/idlocker.go
Madhu-1: we have to do performance measurement, it can be done in other PR
i want to use sync.Map and check which one performs better

Madhu-1:
added DNM label jus to avoid merging

ShyamsundarR:
> nit: https://github.com/ceph/ceph-csi/pull/443/files#r298987001
> other than that LGTM

Addressed the nit. @Madhu-1 PTAL.

Madhu-1:
@humblec PTAL

humblec:
@ShyamsundarR @Madhu-1 I am doing final review , will get back soon.

humblec:
I will remove the targetpath and attachlocks in other PRs. Rest of the discussions here can be taken up in a follow up PR. 

LGTM.

Issue #354, opened 2019-05-03
Find ways to avoid ceph CLI keyring error strings in logs


ShyamsundarR:
As we do not have ceph keyring files on the CSI containers, CLIs that fail and are reported in the logs have additional errors calling out the same. This is a little noisy when debugging issues and is at times misleading. Need to find ways to determine how to let Ceph CLI commands ignore the absence of a keyring.

Sample output:
```
 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin
.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
```

leseb:
Does that mean you're executing a `ceph` command without a key? Can you describe in under which circumstances you are getting this error?

ShyamsundarR:
For example the following command from within the rbdplugin provisioner container would throw such warnings(?) but proceed. These get captured by the plugin when executing these commands from the [code](https://github.com/ceph/ceph-csi/blob/3f35bfd4d770adbb31480bec7575787d56eaab50/pkg/rbd/rbd_util.go#L573) and pollute the logs.

```
$ kubectl exec -it csi-rbdplugin-provisioner-0 -c csi-rbdplugin bash
[root@csi-rbdplugin-provisioner-0 /]# rbd snap rm --pool replicapool --snap id1 image1 --id admin -m 10.100.117.85:6790,10.108.182.120:6790,10.111.75.172:6790 --key=AQC50NBcusHYGBAAJmSLm9Aw98mBRT5FyeS1/w==
2019-05-13 11:12:31.544093 7f4532f68d40 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
rbd: error opening image image1: (2) No such file or directory
```
We do have a minimal ceph.conf in the [code](https://github.com/ceph/ceph-csi/blob/3f35bfd4d770adbb31480bec7575787d56eaab50/pkg/cephfs/cephconf.go#L24), but providing the same does not make the logs go away, is that a way forward? Does it need any further options in it?

dillaman:
@ShyamsundarR Until the CSI gets to the point where it no longer passes the key on the command-line, you can just create an empty "/etc/ceph/ceph.conf" to remove the log message.

PR #445, opened 2019-06-25
Use --keyfile option to pass keys to all Ceph CLIs

ShyamsundarR:
Every Ceph CLI that is invoked at present passes the key via the
--key option, and hence is exposed to key being displayed on
the host using a ps command or such means.

This commit addresses this issue by stashing the key in a tmp
file, which is again created on a tmpfs (or empty dir backed by
memory). Further using such tmp files as arguments to the --keyfile
option for every CLI that is invoked.

This prevents the key from being visible as part of the argument list
of the invoked program on the system.

Fixes: #318

### DIFF
--- a/deploy/cephfs/helm/templates/nodeplugin-daemonset.yaml
+++ b/deploy/cephfs/helm/templates/nodeplugin-daemonset.yaml
@@ -106,6 +106,8 @@ spec:
               readOnly: true
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+            - name: keys-tmp-dir
+              mountPath: /tmp/csi/keys
           resources:
 {{ toYaml .Values.nodeplugin.plugin.resources | indent 12 }}
       volumes:
@@ -142,6 +144,10 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: {{ .Values.configMapName | quote }}
+        - name: keys-tmp-dir
+          emptyDir: {
+            medium: "Memory"
+          }
     {{- if .Values.nodeplugin.affinity -}}
       affinity:
 {{ toYaml .Values.nodeplugin.affinity . | indent 8 }}
--- a/pkg/util/credentials.go
+++ b/pkg/util/credentials.go
@@ -18,22 +18,54 @@ package util
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 )
 
 const (
-	credUserID   = "userID"
-	credUserKey  = "userKey"
-	credAdminID  = "adminID"
-	credAdminKey = "adminKey"
-	credMonitors = "monitors"
+	credUserID           = "userID"
+	credUserKey          = "userKey"
+	credAdminID          = "adminID"
+	credAdminKey         = "adminKey"
+	credMonitors         = "monitors"
+	tmpKeyFileLocation   = "/tmp/csi/keys"
+	tmpKeyFileNamePrefix = "keyfile-"
 )
 
 type Credentials struct {
-	ID  string
-	Key string
+	ID      string
+	KeyFile string
 }
 
-func getCredentials(idField, keyField string, secrets map[string]string) (*Credentials, error) {
+func storeKey(key string) (string, error) {
+	tmpfile, err := ioutil.TempFile(tmpKeyFileLocation, tmpKeyFileNamePrefix)
+	if err != nil {
+		return "", fmt.Errorf("error creating a temporary keyfile (%s)", err)
+	}
+	defer func() {
+		if err != nil {
+			os.Remove(tmpfile.Name())
+		}
+	}()
+
+	if _, err = tmpfile.Write([]byte(key)); err != nil {
+		return "", fmt.Errorf("error writing key to temporary keyfile (%s)", err)
+	}
+
+	keyFile := tmpfile.Name()
+	if keyFile == "" {
+		err = fmt.Errorf("error reading temporary filename for key (%s)", err)
+		return "", err
+	}
+
+	if err = tmpfile.Close(); err != nil {
+		return "", fmt.Errorf("error closing temporary filename (%s)", err)
+	}
+
+	return keyFile, nil
+}
+
+func newCredentialsFromSecret(idField, keyField string, secrets map[string]string) (*Credentials, error) {
 	var (
 		c  = &Credentials{}
 		ok bool
@@ -43,19 +75,41 @@ func getCredentials(idField, keyField string, secrets map[string]string) (*Crede
 		return nil, fmt.Errorf("missing ID field '%s' in secrets", idField)
 	}
 
-	if c.Key, ok = secrets[keyField]; !ok {
+	key := secrets[keyField]
+	if key == "" {
 		return nil, fmt.Errorf("missing key field '%s' in secrets", keyField)
 	}
 
-	return c, nil
+	keyFile, err := storeKey(key)
+	if err == nil {
+		c.KeyFile = keyFile
+	}
+
+	return c, err
 }
 
-func GetUserCredentials(secrets map[string]string) (*Credentials, error) {
-	return getCredentials(credUserID, credUserKey, secrets)
+func (cr *Credentials) DeleteCredentials() {
+	os.Remove(cr.KeyFile)
 }
 
-func GetAdminCredentials(secrets map[string]string) (*Credentials, error) {
-	return getCredentials(credAdminID, credAdminKey, secrets)
+func NewUserCredentials(secrets map[string]string) (*Credentials, error) {
+	return newCredentialsFromSecret(credUserID, credUserKey, secrets)
+}
+
+func NewAdminCredentials(secrets map[string]string) (*Credentials, error) {
+	return newCredentialsFromSecret(credAdminID, credAdminKey, secrets)
+}
+
+func NewCredentials(id, key string) (*Credentials, error) {
+	var c = &Credentials{}
+
+	c.ID = id
+	keyFile, err := storeKey(key)
+	if err == nil {
+		c.KeyFile = keyFile
+	}
+
+	return c, err
 }
 
 func GetMonValFromSecret(secrets map[string]string) (string, error) {
--- a/deploy/cephfs/helm/templates/provisioner-statefulset.yaml
+++ b/deploy/cephfs/helm/templates/provisioner-statefulset.yaml
@@ -90,6 +90,8 @@ spec:
               mountPath: "/rootfs"
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+            - name: keys-tmp-dir
+              mountPath: /tmp/csi/keys
           resources:
 {{ toYaml .Values.nodeplugin.plugin.resources | indent 12 }}
       volumes:
@@ -102,6 +104,10 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: {{ .Values.configMapName | quote }}
+        - name: keys-tmp-dir
+          emptyDir: {
+            medium: "Memory"
+          }
     {{- if .Values.provisioner.affinity -}}
       affinity:
 {{ toYaml .Values.provisioner.affinity . | indent 8 }}
--- a/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
+++ b/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
@@ -94,6 +94,8 @@ spec:
               mountPath: /dev
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+            - name: keys-tmp-dir
+              mountPath: /tmp/csi/keys
       volumes:
         - name: socket-dir
           hostPath:
@@ -111,3 +113,7 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: ceph-csi-config
+        - name: keys-tmp-dir
+          emptyDir: {
+            medium: "Memory"
+          }
--- a/deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
+++ b/deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
@@ -90,6 +90,8 @@ spec:
               mountPath: /dev
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+            - name: keys-tmp-dir
+              mountPath: /tmp/csi/keys
       volumes:
         - name: mount-cache-dir
           emptyDir: {}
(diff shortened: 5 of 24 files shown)


ShyamsundarR:
This is WIP, as the change to use `--keyfile` is not made to all CLIs.

Wanted a review of the approach and if this looks fine, post which I will change all locations and required pod manifests to mount the named tmp directory as an emptyDir stored in memory.

Also, e2e would be enhanced to add a test to ensure that the tmp directory is empty at the end of tests before the pods are deleted (if such things are possible using the e2e framework).

The code is otherwise complete, and ready for a review.

@phlogistonjohn tagging you for a review/look as well.

pkg/util/credentials.go
```diff
@@ -58,6 +100,19 @@ func GetAdminCredentials(secrets map[string]string) (*Credentials, error) {
 	return getCredentials(credAdminID, credAdminKey, secrets)
 }
 
+func GetCredentials(id, key string) (*Credentials, error) {
```
dillaman: Nit: personally don't like very similar named methods doing different things (for clarity purposes). Perhaps rename ```getCredentials``` internal method to something like ```getCredentialsFromSecrets```

ShyamsundarR: Ack! will address in the next PR.

Along the same lines should we rename `GetXXXCredentials` to `NewXXXCredentials` so that `DeleteXXXCredentials` is its destructor?

dillaman: > Along the same lines should we rename `GetXXXCredentials` to `NewXXXCredentials` so that `DeleteXXXCredentials` is its destructor?

Makes sense to me since it helps to imply it needs to be cleaned up after use via the `DeleteXXXCredentials`

ShyamsundarR: Modified as per provided comments.

ShyamsundarR:
NOTE: Thanks to @phlogistonjohn for help with approach and implementation details.

ShyamsundarR:
@Madhu-1 @humblec PTAL

pkg/util/credentials.go
```diff
@@ -18,22 +18,54 @@ package util
...
 
 type Credentials struct {
-	ID  string
-	Key string
+	ID      string
+	KeyFile string
 }
 
-func getCredentials(idField, keyField string, secrets map[string]string) (*Credentials, error) {
+func storeKey(key string) (string, error) {
```
Madhu-1: is there any performance impact for creating and delete a file on each request? 

for the same userID and password are we creating a new file each time? if yes can we avoid it?

deploy/cephfs/helm/templates/nodeplugin-daemonset.yaml
```diff
@@ -142,6 +144,10 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: {{ .Values.configMapName | quote }}
+        - name: keys-tmp-dir
+          emptyDir: {
```
Madhu-1: as this increases the file operation performance, will this cause any bottleneck for serving more number for requests? as this affects the container memory.

ShyamsundarR: Here is my **theoretical** reasoning on the footprint we would use,
- Number of requests in flight, depends on number of threads by the various side cars, seem to be configured at 100 (and I do not see a need to increase these yet)
- Hence we would use up, 100*(say 3)files*(say 2) sidecars in all, which is about 600, round it off to 1KiB (1024 files)
- 1KiB files with content as the base64 encoded Ceph key of 40 bytes each, should take up an inode each. Assuming an inode is 256 bytes this should take 256KiB in memory

I would hence assume this is fine and not a concern from a memory footprint standpoint.

pkg/util/credentials.go
ShyamsundarR: > is there any performance impact for creating and delete a file on each request?

I have not measured the performance impact due to this, do you want me to?

> 
> for the same userID and password are we creating a new file each time? if yes can we avoid it?

I would say, we should avoid credential caching in any form, unless we have no other way around it. Thus, my answer is we cannot avoid it.

Madhu-1: @humblec any comments?

deploy/cephfs/helm/templates/nodeplugin-daemonset.yaml
Madhu-1: cc @humblec

ShyamsundarR:
@Madhu-1 @humblec PTAL or let me know what is required to move this forward.

humblec:
@ShyamsundarR I am trying to understand why we need this change:

>Every Ceph CLI that is invoked at present passes the key via the
--key option, and hence is exposed to key being displayed on
the host using a ps command or such means.
>

I am not sure about the `security` implication if the `ps` list the keyfile. That said, first of all, who got the access to the system and can issue `ps` command ? iic, its going to be an admin. Obviously its not an end user, Isnt it ? If admin got access to the system , why we need to hide it from admin or what security issue we are trying to solve here ? 

Also I have a question on, how this differs from "non container" environment?  if we consider this as a security issue, its applicable for that kind of environment too , Isnt it ?


Coming back to the approach itself,  storing key file in `tmp` location for each and every request looks to be a burden or unwanted to me .

ShyamsundarR:
> @ShyamsundarR I am trying to understand why we need this change:
> 
> > Every Ceph CLI that is invoked at present passes the key via the
> > --key option, and hence is exposed to key being displayed on
> > the host using a ps command or such means.
> 
> I am not sure about the `security` implication if the `ps` list the keyfile. That said, first of all, who got the access to the system and can issue `ps` command ? iic, its going to be an admin. Obviously its not an end user, Isnt it ? If admin got access to the system , why we need to hide it from admin or what security issue we are trying to solve here ?

The key is visible to any admin on the host of the system and is in general a bad practice that we need to rectify.

> 
> Also I have a question on, how this differs from "non container" environment? if we consider this as a security issue, its applicable for that kind of environment too , Isnt it ?

If you read Ceph documentation, the options provided there are similar (and more elaborate), to enable not showing the key for various CLIs.

For CephFS mounts, the process is even long running and hence the key exposed to ps far longer.

> 
> Coming back to the approach itself, storing key file in `tmp` location for each and every request looks to be a burden or unwanted to me .

The tmp file is in a in-memory tmpfs, performance and burdens seem marginal to me.

Tagging @dillaman and @phlogistonjohn who had the initial concerns (including me as well) to chime in on this comment.

dillaman:
:+1: to not seeing the CephX keys in the process command line.

humblec:
>The key is visible to any admin on the host of the system and is in general a bad practice that we need to rectify.
>
Well, admin created the key mostly :).. 

@ShyamsundarR Do we need the change from `GetAdminCredentials` to `NewAdminCredentials` ? I doubt,  this change add any value  or IOW, we are used to `GetAdminCredentials` and now we have to keep the new func name . May be tomorrow, we may feel it can be changed to "createAdminCredentials` ..etc :). 

One other thought is that, as the key is stored in `tmpfs` , its a complete wipe when the pod restarted/re spawned. 

Also, with this change, we are actually disallowing the use of custom `keyfile` location,  Isnt it ?

ShyamsundarR:
> > The key is visible to any admin on the host of the system and is in general a bad practice that we need to rectify.
> 
> Well, admin created the key mostly :)..

Ceph admin/user passed in via secrets are different than Kube/node admin, so not the same "admin" in this context. But more generically anybody who can see ps (like) output on the node.

> 
> @ShyamsundarR Do we need the change from `GetAdminCredentials` to `NewAdminCredentials` ? I doubt, this change add any value or IOW, we are used to `GetAdminCredentials` and now we have to keep the new func name . May be tomorrow, we may feel it can be changed to "createAdminCredentials` ..etc :).

We return a new interface/object here, and the canonical form is NewXXX (afaict, as in go and current code examples `NewCSIDriver` `NewNonBlockingGRPCServer` `NewIDLocker` `NewCachePersister`) hence modified as such. It was not a choice of the day decision that we should need to change it in the future unless its nature changes.

> 
> One other thought is that, as the key is stored in `tmpfs` , its a complete wipe when the pod restarted/re spawned.

Yes it is a complete wipe, and hence the choice to use `tmpfs` in the first place.

> 
> Also, with this change, we are actually disallowing the use of custom `keyfile` location, Isnt it ?

"custom" keyfile as in? I mean where are we using or where do we think of using custom keyfile?

Madhu-1:
@ShyamsundarR please provide some numbers regarding parallel PVC creation on this PR, so that we can take this PR in. this would help us/others that this PR is not impacting the parallel PVC creation time.

@dillaman suggested in [slack](https://cephcsi.slack.com/archives/CJV6G2DFT/p1563465019063900) on how to do some performance test

ShyamsundarR:
> @ShyamsundarR please provide some numbers regarding parallel PVC creation on this PR, so that we can take this PR in. this would help us/others that this PR is not impacting the parallel PVC creation time.
> 
> @dillaman suggested in [slack](https://cephcsi.slack.com/archives/CJV6G2DFT/p1563465019063900) on how to do some performance test

Using the following script from within one of the containers on my vagrant based setup, looked at execution time for couple of iteration counts (as below). The script does a create/write/read/unlink pattern as in the code.

**tmpfs-perf.sh**:
```
#!/bin/bash

if [ $# -ne 1 ]; then
  echo "Usage: $0 <count>"
  exit 1
fi

count=$1
basepath="/tmp/csi/keys"

echo "Running test for $count iterations against $basepath"

while [ "$count" -ne "0" ]; do
  echo "20chars890012345678920chars890012345678920chars890012345678920chars890012345678920chars8900123456789" > "${basepath}"/a.txt
  cat "${basepath}"/a.txt > /dev/null 2>&1
  unlink "${basepath}"/a.txt

  count=$((count-1))
done
```

**Results from the above are as follows:**
```
# time ./tmpfs-perf.sh 100
Running test for 100 iterations against /tmp/csi/keys

real	0m0.252s
user	0m0.105s
sys	0m0.156s

# time ./tmpfs-perf.sh 100
Running test for 100 iterations against /tmp/csi/keys

real	0m0.288s
user	0m0.132s
sys	0m0.136s

# time ./tmpfs-perf.sh 100
Running test for 100 iterations against /tmp/csi/keys

real	0m0.234s
user	0m0.115s
sys	0m0.128s

# time ./tmpfs-perf.sh 1000
Running test for 1000 iterations against /tmp/csi/keys

real	0m4.759s
user	0m1.685s
sys	0m1.843s

# time ./tmpfs-perf.sh 1000
Running test for 1000 iterations against /tmp/csi/keys

real	0m3.155s
user	0m1.443s
sys	0m1.476s

# time ./tmpfs-perf.sh 1000
Running test for 1000 iterations against /tmp/csi/keys

real	0m2.968s
user	0m1.295s
sys	0m1.411s

# time ./tmpfs-perf.sh 1000
Running test for 1000 iterations against /tmp/csi/keys

real	0m4.164s
user	0m1.600s
sys	0m1.808s
```

### Summary of test would be as below:
Average for run of 100 iterations:
  - Real time per run: 0.258 seconds
  - Per iteration (or key file store) within run: 0.0026 seconds
  
Average for run of 1000 iterations:
  - Real time per run: 3.7615 seconds
  - Per iteration (or key file store) within run: 0.0038 seconds

Madhu-1:
@ShyamsundarR Thanks for the numbers

humblec:
LGTM.  Approved.

Madhu-1:
will this Fixes: #354

Madhu-1:
> will this Fixes: #354

I think NO, from CI logs

```
Jul 23 04:16:16.286: INFO: At 2019-07-23 04:10:56 +0000 UTC - event for csi-cephfs-pvc: {cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-0_378dbb85-acff-11e9-923d-0242ac11000b } ProvisioningFailed: (combined from similar events): failed to provision volume with StorageClass "csi-cephfs-sc": rpc error: code = Internal desc = an error occurred while running (2321) ceph [fs subvolumegroup create myfs csi --mode 777 --pool_layout myfs-data0 -m rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789 -c /etc/ceph/ceph.conf -n client.admin --keyfile=***stripped***]: exit status 22: 2019-07-23 04:15:14.307 7fa4290cc700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-23 04:15:14.307 7fa4290cc700 -1 AuthRegistry(0x7fa424063b68) no keyring found at /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,, disabling cephx
2019-07-23 04:15:14.359 7fa4290cc700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-23 04:15:14.359 7fa4290cc700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-23 04:15:14.359 7fa4290cc700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-23 04:15:14.391 7fa4290cc700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-23 04:15:14.391 7fa4290cc700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-23 04:15:14.395 7fa4290cc700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-23 04:15:14.395 7fa4290cc700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
```

humblec:
@ShyamsundarR  Can you rebase ?

ShyamsundarR:
Reserving a DNM label to check a merge conflict and do not want this inadvertently merged.

ShyamsundarR:
> Reserving a DNM label to check a merge conflict and do not want this inadvertently merged.

Removing the DNM.

Issue #420, opened 2019-06-10
RBD's PV claim can require the volume to be encrypted


hairesis:
# Describe the feature you'd like to have #

PV claims with RBD can require the volume to be encrypted. 
Note to avoid confusion: I do know ceph encrypts at rest with LUKS already. What I am actually trying to achieve is to encrypt the single volume, similarly at what an EBS request would do.

# What is the value to the end user? (why is it a priority?) #
- This feature would put Ceph almost in par with what is offered with Amazon EBS volume claims.
- In-Transit encryption might not be necessary for many Ceph use cases (messenger v2 might not be the answer for everybody)
- Companies using ceph as their back-end for their public or private offerings, might offer that as a feature also.
- Encryption at rest and in transit, even within the perimeter of a company, is paramount these days.
- ...

# How will we know we have a good solution? (acceptance criteria) #
I'm looking for guidance here, it would not be a problem to develop it. I am not even sure this is the right repo to do this but I would really appreciate on how to implement this solution.

# Additional context #
I've worked on a little [gist here](https://gist.github.com/hairesis/1ff36e3c61d45d4e2e1b2ec74f46ef00) to prototype the process using LUKS. As I said above, I would like to get some guidance on where this process can be plugged in and / or there is an equivalent solution already worked on I am not aware of.

dillaman:
@hairesis Is the goal to have the RBD CSI interface w/ LUKS behind the scenes? For example, you could store a k8s secret and either instruct the storage class to encrypt new volumes using the secret or provide the secret for individual PVCs?

I'd imagine snapshots would remain encrypted with the original key but any attempt to create a new volume from the snapshot w/ a new secret would require a deep-copy of the snapshot to the new volume.

hairesis:
@dillaman thanks.

> Is the goal to have the RBD CSI interface w/ LUKS behind the scenes? For example, you could store a k8s secret and either instruct the storage class to encrypt new volumes using the secret or provide the secret for individual PVCs?

yes, that is the idea. We would like to encrypt new volumes for individual persistent volume claims.

> I'd imagine snapshots would remain encrypted with the original key but any attempt... 

At the moment we are not concerned about snapshots but it's certainly something we would like to think about  going forward.

hairesis:
I've created a proof of concept change according to what described above.
Code changes can be found here: 
https://github.com/hairesis/ceph-csi/blob/csi-v0.3-encrypt/pkg/rbd/rbd_attach.go
[...]

PR #448, opened 2019-06-26
[Proposal] Persistent Volume Claims Encryption

hairesis:
Introduces encryption and kms configuration in StorageClass parameter
Implements rbd volume encryption relying on dm-crypt and cryptsetup
using LUKS extension.

Relates to Feature request #420 
Requesting review comments on approach

Signed-off-by: Andrea Baglioni <0x41ndrea@gmail.com>

### DIFF
--- a/docs/design/proposals/encrypted-pvc.md
+++ b/docs/design/proposals/encrypted-pvc.md
@@ -1,0 +1,99 @@
+# Encrypted Persistent Volume Claims
+
+## Proposal
+Subject of this proposal is to add an encrypt feature to ceph-csi in order to allow encryption at rbd volumes level.
+
+Some but not all the benefits of this approach:
+- guarantee encryption in transit to rbd without using messenger v2
+- extra security layer to application with special regulatory needs
+- at rest encryption can be disabled to selectively allow encryption only where required
+
+## Document Terminology
+- volume encryption: encryption of a volume attached by rbd
+- encryption at rest: encryption of physical disk done by ceph
+- LUKS: Linux Unified Key Setup: stores all of the needed setup information for dm-crypt on the disk
+- dm-crypt: linux kernel device-mapper crypto target
+- cryptsetup: the command line tool to interface with dm-crypt
+
+## Proposed Solution
+
+The proposed solution in this document, is to address the volume encryption requirement by using dm-crypt module through cryptsetup cli interface.
+
+### Implementation Summary
+
+- Encryption is implemented using cryptsetup with LUKS extension.
+  A good introduction to LUKS and dm-crypt in general can be found [here](https://wiki.archlinux.org/index.php/Dm-crypt/Device_encryption#Encrypting_devices_with_cryptsetup)
+  Functions to implement necesary interaction are implemented in a separate dmcrypt.go file.
+  1. applyLUKS
+  2. openLUKS
+  3. getPassphrase
+  4. closeLUKS
+
+- `createRBDImage`: refactored to call `applyLUKS`, before returning, if encrypted volume option is set.
+- `applyLUKS`: attaches newly generated image to the worker node where provisioner is running and sets up LUKS using `cryptsetup` command.
+  The function:
+  - resolves the passphrase from a `getPassphrase` function according to kms volume options.
+  - A new passphrase, if not found, is stored in the kms using target volume name (i.e. pvc-8a710f4c934811e9) for lookup
+  - applies LUKS to the attached rbd volume
+  - detaches the volume from the worker node where the provisioner is running.
+- `attachRBDImage`: refactored into a strategy function to call `openLUKS` when an encrypted volume attach is requested.
+- `openLUKS`: looks up for a passphrase matching the volume name, return the new device path in the form: `/dev/mapper/luks-pvc-8a710f4c934811e9`.
+  On the woker node where the attach is scheduled:
+  ```
+  [user@worker-01 ~]$ lsblk
+  NAME                            MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINT
+  sda                               8:0    0   10G  0 disk
+  └─sda1                            8:1    0   10G  0 part  /
+  sdb                               8:16   0   20G  0 disk
+  rbd0                            253:0    0    1G  0 disk
+  └─luks-pvc-8a710f4c934811e9 252:0    0 1020M  0 crypt /var/lib/kubelet/pods/9eaceaef-936c-11e9-b396-005056af3de0/volumes/kubernetes.io~csi/pvc-8a710f4c934811e9/mount
+  ```
+- `detachRBDImage`: calls `closeLUKS` function to remove the LUKS mapping before detaching the volume.
+
+
+- StorageClass extended with following parameters:
+  1. `encrypted: bool`
+  2. `kms: map`
+  cephcsi plugin will support different kms vendors with different type of authentication
+
+#### Annotated YAML for RBD StorageClass
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+   name: csi-rbd
+provisioner: rbd.csi.ceph.com
+parameters:
+
+    monitors: <monitor-1>, <monitor-2>, <monitor-3>
+
+    # Ceph pool into which the RBD image shall be created
+    pool: <pool-name>
+
+    # RBD image format. Defaults to "2".
+    imageFormat: "2"
+
+    imageFeatures: layering
+
+    # Encryption instructions
+    encrypted: "true"
+    kms:
+        # The type of kms we want to connect to: Barbican, aws kms or others can be supported
+        type: "vault"
+
+        # Auth methods offered by the kms.
+        # In this case one of https://www.vaultproject.io/docs/auth/index.html
+        auth:
+            method: "kubernetes"
+
+    csiProvisionerSecretName: csi-rbd-secret
+    csiProvisionerSecretNamespace: default
+    csiNodePublishSecretName: csi-rbd-secret
+    csiNodePublishSecretNamespace: default
+
+    # Ceph users for operating RBD
+    adminid: admin
+    userid: username
+
+reclaimPolicy: Delete
+```


docs/design/proposals/encrypted-pvc.md
```diff
@@ -0,0 +1,99 @@
...
+    # RBD image format. Defaults to "2".
+    imageFormat: "2"
+
+    imageFeatures: layering
+
+    # Encryption instructions
+    encrypted: "true"
+    kms:
+        # The type of kms we want to connect to: Barbican, aws kms or others can be supported
+        type: "vault"
```
dillaman: I'd imagine you will also have to store secrets for how to communicate w/ Vault like endpoint addresses, keys/certificates, etc (could be stuffed into the existing ```csi-rbd-secret``` or I suppose or injected into the container).

docs/design/proposals/encrypted-pvc.md
```diff
@@ -0,0 +1,99 @@
...
+
+        # Auth methods offered by the kms.
+        # In this case one of https://www.vaultproject.io/docs/auth/index.html
+        auth:
+            method: "kubernetes"
```
dillaman: Just to clarify, I suspect this means the Vault auth is tied to the RBD controller/node service accounts and not the end-user requesting the PVC?

docs/design/proposals/encrypted-pvc.md
```diff
@@ -0,0 +1,99 @@
...
+  1. applyLUKS
+  2. openLUKS
+  3. getPassphrase
+  4. closeLUKS
+
+- `createRBDImage`: refactored to call `applyLUKS`, before returning, if encrypted volume option is set.
+- `applyLUKS`: attaches newly generated image to the worker node where provisioner is running and sets up LUKS using `cryptsetup` command.
+  The function:
+  - resolves the passphrase from a `getPassphrase` function according to kms volume options.
+  - A new passphrase, if not found, is stored in the kms using target volume name (i.e. pvc-8a710f4c934811e9) for lookup
```
dillaman: Will this always be an auto-generated passphrase when requesting a new PVC (each PV w/ its own unique key)? In the case of Vault, that might be fine. For other secret sources, might need a way to store a `StorageClass`-supplied key and/or a `PVC`-supplied key.

docs/design/proposals/encrypted-pvc.md
```diff
@@ -0,0 +1,99 @@
...
+## Proposed Solution
+
+The proposed solution in this document, is to address the volume encryption requirement by using dm-crypt module through cryptsetup cli interface.
+
+### Implementation Summary
+
+- Encryption is implemented using cryptsetup with LUKS extension.
+  A good introduction to LUKS and dm-crypt in general can be found [here](https://wiki.archlinux.org/index.php/Dm-crypt/Device_encryption#Encrypting_devices_with_cryptsetup)
+  Functions to implement necesary interaction are implemented in a separate dmcrypt.go file.
+  1. applyLUKS
```
dillaman: Nit: `createLUKS` or `initLUKS` instead of `applyLUKS`?

docs/design/proposals/encrypted-pvc.md
hairesis: I was thinking to piggyback on `csi-rbd-secret`.

docs/design/proposals/encrypted-pvc.md
hairesis: In this proposal yes. Although I am open for a more sophisticated configuration. Do you have something specific in mind?

docs/design/proposals/encrypted-pvc.md
hairesis: That's a good question. Would definitely be useful to allow users specify their key id, or refer to it regardless of the source; however, unless there are strong objections, I would probably limit for an initial implementation, to auto-generate passphrases.

docs/design/proposals/encrypted-pvc.md
hairesis: 👍

docs/design/proposals/encrypted-pvc.md
dillaman: I'm mostly just concerned about designing a feature that can be cleanly expanded in the future. I think it's fine if the initial implementation only auto-generates passphrases. However, if a `StorageClass` global passphrase is eventually supported, there wouldn't be a way to have a PVC auto-generate its own passphrase unless perhaps a new attribute was added to indicate the passphrase source is "generate" vs "static".

docs/design/proposals/encrypted-pvc.md
dillaman: Cool. Since I am not familiar w/ the details of what is required to talk to Vault, it might be good to also document what Vault requires in this proposal as well.

docs/design/proposals/encrypted-pvc.md
dillaman: No, just wanted to ensure that was the intention (i.e. the service account owns the keys to access the passphrase, not the user requesting the PV).

docs/design/proposals/encrypted-pvc.md
Madhu-1: initLUKS looks good

docs/design/proposals/encrypted-pvc.md
Madhu-1: are we going to support secret+storageclass based encryption? let's say the user specifies the secret name in the storage class and we are going to fetch the secret details in ceph-csi rbd plugin and use it for encryption.

docs/design/proposals/encrypted-pvc.md
hairesis: sure thing, working on it.

docs/design/proposals/encrypted-pvc.md
hairesis: The idea is to have the storageclass define the connection to the kms. Each PVC provisioned with this storage class will then generate a new passphrase and store it / read it from a pre-configured key path. In a second iteration we could allow PVC to specify a key path that overrides the storageclass conguration.
I'm open for any suggestion to improve it anyway.

docs/design/proposals/encrypted-pvc.md
hairesis: [Kubernetes authentication method](https://www.vaultproject.io/docs/auth/kubernetes.html) does not require much configuration from K8S side; it uses a k8s serviceAccount to authenticate and my plan is to leverage the serviceAccount setup for the csiplugin pod. Conversely there is a bit more configuration on the Vault side but I think it's beyond the scope of this document.

However I realized some more information is required to correctly configure Vault for it to be accessible.
```
# Encryption instructions
    encrypted: "true"
    kms:
        # The type of kms we want to connect to: Barbican, aws kms or others can be supported
        backend: "vault"
        address: "https://vault.example.com:8080"
        ca: vault-rbd-ca-secret
        keyPath: "kv/rbd-luks-passphrases"
        # Auth methods offered by the kms.
        # In this case one of https://www.vaultproject.io/docs/auth/index.html
        auth:
            method: "kubernetes"
```
@dillaman  How do you think configuration block above can be improved? I don't have extensive experience at this.

docs/design/proposals/encrypted-pvc.md
mykaul: So the user asked to encrypt, but someone else holds the keys?

humblec:
@hairesis Cant we close this PR in favor of #697 ?

Madhu-1:
@humblec as this is proposal Doc I think we can merge this one  (once CI  fixed)

ShyamsundarR:
> @humblec as this is proposal Doc I think we can merge this one (once CI fixed)

I agree, retaining proposal docs is useful for the project.

hairesis:
@ShyamsundarR nice one, let me review the proposal one more time @vasyl-purchel to make sure it's consistent to what we have implemented.

vasyl-purchel:
@humblec, @ShyamsundarR this proposal is not implemented completely yet - I'm soon to create one more PR adding vault as KMS for per-volume encryption (each volume will receive new generated passphrase that will be stored in vault with possibility to extend this to any other KMS in future)

I think I can add this doc to the same PR and then we can close this one in favor of a new one with implementation.

jimmyjones2:
@vasyl-purchel Can you also support Kubernetes secrets to store the volume keys please?

humblec:
> @humblec, @ShyamsundarR this proposal is not implemented completely yet - I'm soon to create one more PR adding vault as KMS for per-volume encryption (each volume will receive new generated passphrase that will be stored in vault with possibility to extend this to any other KMS in future)

Sure, lets have a documentation/proposal on the same .

> 
> I think I can add this doc to the same PR and then we can close this one in favor of a new one with implementation.

Sure.

vasyl-purchel:
added current proposal here: https://github.com/ceph/ceph-csi/pull/762
@jimmyjones2 it is already possible with latest master to have a passphrase per storageclass in kubernetes secrets to encrypt volumes, current focus is to add vault as a kms and after adding passphrase for each volume using kubernetes secrets should be easy enough but I would avoid putting it all into same PR as it will get too big.

jimmyjones2:
@vasyl-purchel Is there an issue or PR to have per-volume encryption using kubernetes secrets?

vasyl-purchel:
> @vasyl-purchel Is there an issue or PR to have per-volume encryption using kubernetes secrets?

@jimmyjones2 not that I know of, since it won't be more secure than a single passphrase per storageclass we did not implement it, but feel free to send a PR for that - code is done in a way that adding other per-volume KMS should be easy

jimmyjones2:
@vasyl-purchel I respectfully disagree.

Can you explain why per volume Kubernetes secrets would be no more secure than a single passphrase per storageclass?

In my opinion the main threat is a container escape on a node after compromise of an app running in the cluster. Container escape vulnerabilities seems to come up every year (CVE-2016-5195, CVE-2017-5123, CVE-2019-5736) or zero days are sold for about [~$50k](https://zerodium.com/program.html). The other avenue is a bug in the CSI driver itself to gain execution, eg. command injection, again after compromise of an app running in the cluster.

As it stands, even with Vault encryption, if you have control of *any* instance of the ceph-csi daemonset, you can fetch any volume secret and mount any encrypted volume you like. For Kubernetes secrets, this has been thought about - the API server will only release secrets to a kubelet if it's been scheduled a pod that needs it. I don't know if this restriction has been implemented for [per volume CSI secrets](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#per-volume-secrets), but if it hasn't, I see no reason why it couldn't be.

PR #450, opened 2019-06-28
Fix returning success from DeleteSnapshot for stale requests

ShyamsundarR:
Also reduced code duplication in fetching pool list from Ceph.

DeleteSnapshot like DeleteVolume, should return a success when it
detects that the snapshot keys are missing from the RADOS OMaps that
store the snapshot UUID to request name mapping.

This was missing in the code, and is now added.

### DIFF
--- a/pkg/util/cephcmds.go
+++ b/pkg/util/cephcmds.go
@@ -53,9 +53,8 @@ type cephStoragePoolSummary struct {
 	Number int64  `json:"poolnum"`
 }
 
-// GetPoolID searches a list of pools in a cluster and returns the ID of the pool that matches
-// the passed in poolName parameter
-func GetPoolID(monitors string, cr *Credentials, poolName string) (int64, error) {
+// GetPools fetches a list of pools from a cluster
+func getPools(monitors string, cr *Credentials) ([]cephStoragePoolSummary, error) {
 	// ceph <options> -f json osd lspools
 	// JSON out: [{"poolnum":<int64>,"poolname":<string>}]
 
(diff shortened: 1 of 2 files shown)


pkg/util/cephcmds.go
```diff
@@ -47,15 +47,14 @@ func ExecCommand(program string, args ...string) (stdout, stderr []byte, err err
 	return stdoutBuf.Bytes(), nil, nil
 }
 
-// cephStoragePoolSummary strongly typed JSON spec for osd ls pools output
-type cephStoragePoolSummary struct {
+// CephStoragePoolSummary strongly typed JSON spec for osd ls pools output
+type CephStoragePoolSummary struct {
```
humblec: @ShyamsundarR why we need above change? afaict, its used internally here .

ShyamsundarR: static linters complained that this was bad style, hence fixed it up.

humblec: We dont need to export a type unwantedly..

Madhu-1: @humblec if you don't fix it CI will  fail

Madhu-1: this function may get  called by other functions  in  different package  hence it's required

pkg/util/cephcmds.go
```diff
@@ -47,15 +47,14 @@ func ExecCommand(program string, args ...string) (stdout, stderr []byte, err err
...
 	Name   string `json:"poolname"`
 	Number int64  `json:"poolnum"`
 }
 
-// GetPoolID searches a list of pools in a cluster and returns the ID of the pool that matches
-// the passed in poolName parameter
-func GetPoolID(monitors string, cr *Credentials, poolName string) (int64, error) {
+// GetPools fetches a list of pools from a cluster
+func GetPools(monitors string, cr *Credentials) ([]CephStoragePoolSummary, error) {
```
Madhu-1: why this function is exported? is this expected to be used in other packages?

ShyamsundarR: It is currently not called form other packages, one way to possibly fix the other comment on the linter problem is to not expose this. Is that what you want?

Madhu-1: yes, unexporting this function should fix it.

pkg/util/cephcmds.go
ShyamsundarR: Fixed.

pkg/util/cephcmds.go
ShyamsundarR: Fixed.

pkg/util/cephcmds.go
humblec: @Madhu-1 if it was used in other packages, this wanted to be "exported" before this PR itself, Isnt it ? :).

humblec: Thanks for fixing it @ShyamsundarR

pkg/util/cephcmds.go
```diff
@@ -69,14 +68,25 @@ func GetPoolID(monitors string, cr *Credentials, poolName string) (int64, error)
...
-		return 0, err
+		return nil, err
 	}
 
 	var pools []cephStoragePoolSummary
 	err = json.Unmarshal(stdout, &pools)
 	if err != nil {
 		klog.Errorf("failed to parse JSON output of pool list from cluster (%s)", err)
-		return 0, fmt.Errorf("unmarshal failed: %+v.  raw buffer response: %s", err, string(stdout))
+		return nil, fmt.Errorf("unmarshal failed: %+v.  raw buffer response: %s", err, string(stdout))
```
humblec: @ShyamsundarR may be "unmarshal of pool list failed..." looks more informative.

ShyamsundarR: Done.

Issue #153, opened 2019-01-31
Why not implement NodeStageVolume/NodeUnstageVolume?


kindule:
Hi, all

I've read the code of [ceph-csi/pkg/rbd/nodeserver.go](https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/nodeserver.go). And wonder why NodeStageVolume/NodeUnstageVolume not implemented. In my opinion, NodeStageVolume is responsible for mkfs/mount rbd to global path of the node and NodeUnstageVolume is responsible for umount rbd from global path of the node, which currently is the part of the NodePublishVolume/NodeUnpublishVolume. And that's make some complex flow in the NodePublishVolume/NodeUnpublishVolume.

If that need some contributor to fix it, I can make my best effort.

```go
// NodeStageVolume returns unimplemented response
func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
	return nil, status.Error(codes.Unimplemented, "")
}

// NodeUnstageVolume returns unimplemented response
func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolumeRequest) (*csi.NodeUnstageVolumeResponse, error) {
	return nil, status.Error(codes.Unimplemented, "")
}
```

rootfs:
sure, PR is always welcome 👍

Madhu-1:
@kindule are you working on this issue?

Madhu-1:
during the `NodeStageVolume` the stagingPath is a directory, not file (this directory is created by kube not by csi driver). can we bind a /dev/rbdx` to the directory and again bind that directory to a file in nodePublish volume?

@j-griffith @dillaman @humblec @nixpanic  any idea how to solve this.

obnoxxx:
In the block type case, can't we create a device file inside the stagingPath and map that into the container at publish?

PR #457, opened 2019-07-03
Implement NodeStage and NodeUnstage for rbd

Madhu-1:

In NodeStage RPC call we have to map the device to the node plugin and make sure the device will be mounted to the global path.

in  nodeUnstage request unmount the device from the global path and unmap the device if the volume mode is block  we will be creating a file inside a stageTargetPath  and it will be
considered as the global path.

## Related issues ##

Fixes: #153

### DIFF
--- a/deploy/rbd/kubernetes/csi-rbdplugin.yaml
+++ b/deploy/rbd/kubernetes/csi-rbdplugin.yaml
@@ -7,6 +7,8 @@ spec:
   selector:
     matchLabels:
       app: csi-rbdplugin
+  updateStrategy:
+    type: OnDelete
   template:
     metadata:
       labels:
@@ -39,7 +41,7 @@ spec:
                 fieldRef:
                   fieldPath: spec.nodeName
           volumeMounts:
-            - name: plugin-dir
+            - name: socket-dir
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
@@ -69,14 +71,8 @@ spec:
               value: unix:///csi/csi.sock
           imagePullPolicy: "IfNotPresent"
           volumeMounts:
-            - name: plugin-dir
+            - name: socket-dir
               mountPath: /csi
-            - name: pods-mount-dir
-              mountPath: /var/lib/kubelet/pods
-              mountPropagation: "Bidirectional"
-            - name: plugin-mount-dir
-              mountPath: /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/
-              mountPropagation: "Bidirectional"
             - mountPath: /dev
               name: host-dev
             - mountPath: /rootfs
@@ -88,22 +84,28 @@ spec:
               readOnly: true
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+            - name: plugin-dir
+              mountPath: /var/lib/kubelet/plugins
+              mountPropagation: "Bidirectional"
+            - name: mountpoint-dir
+              mountPath: /var/lib/kubelet/pods
+              mountPropagation: "Bidirectional"
       volumes:
-        - name: plugin-dir
+        - name: socket-dir
           hostPath:
             path: /var/lib/kubelet/plugins/rbd.csi.ceph.com
             type: DirectoryOrCreate
-        - name: plugin-mount-dir
+        - name: plugin-dir
           hostPath:
-            path: /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/
+            path: /var/lib/kubelet/plugins
+            type: Directory
+        - name: mountpoint-dir
+          hostPath:
+            path: /var/lib/kubelet/pods
             type: DirectoryOrCreate
         - name: registration-dir
           hostPath:
             path: /var/lib/kubelet/plugins_registry/
-            type: Directory
-        - name: pods-mount-dir
-          hostPath:
-            path: /var/lib/kubelet/pods
             type: Directory
         - name: host-dev
           hostPath:
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -41,59 +41,15 @@ type NodeServer struct {
 	mounter mount.Interface
 }
 
-// NodePublishVolume mounts the volume mounted to the device path to the target
-// path
-func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
-	targetPath := req.GetTargetPath()
-	if targetPath == "" {
-		return nil, status.Error(codes.InvalidArgument, "empty target path in request")
-	}
-
-	if req.GetVolumeCapability() == nil {
-		return nil, status.Error(codes.InvalidArgument, "empty volume capability in request")
-	}
-
-	if req.GetVolumeId() == "" {
-		return nil, status.Error(codes.InvalidArgument, "empty volume ID in request")
-	}
-
-	cr, err := util.GetUserCredentials(req.GetSecrets())
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
-	}
-
-	idLk := targetPathLocker.Lock(targetPath)
-	defer targetPathLocker.Unlock(idLk, targetPath)
-
-	disableInUseChecks := false
-
-	isLegacyVolume := false
-	volName, err := getVolumeName(req)
-	if err != nil {
-		// error ErrInvalidVolID may mean this is an 1.0.0 version volume, check for name
-		// pattern match in addition to error to ensure this is a likely v1.0.0 volume
-		if _, ok := err.(ErrInvalidVolID); !ok || !isLegacyVolumeID(req.GetVolumeId()) {
-			return nil, status.Error(codes.InvalidArgument, err.Error())
-		}
-
-		volName, err = getLegacyVolumeName(req)
-		if err != nil {
-			return nil, status.Error(codes.InvalidArgument, err.Error())
-		}
-		isLegacyVolume = true
-	}
-
-	isBlock := req.GetVolumeCapability().GetBlock() != nil
-	// Check if that target path exists properly
-	notMnt, err := ns.createTargetPath(targetPath, isBlock)
-	if err != nil {
+// NodeStageVolume mounts the volume to a staging path on the node.
+func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
+	if err := util.ValidateNodeStageVolumeRequest(req); err != nil {
 		return nil, err
 	}
 
-	if !notMnt {
-		return &csi.NodePublishVolumeResponse{}, nil
-	}
-
+	stagingTargetPath := req.GetStagingTargetPath()
+	isBlock := req.GetVolumeCapability().GetBlock() != nil
+	disableInUseChecks := false
 	// MULTI_NODE_MULTI_WRITER is supported by default for Block access type volumes
 	if req.VolumeCapability.AccessMode.Mode == csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER {
 		if isBlock {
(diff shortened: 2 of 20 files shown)


Madhu-1:
@j-griffith @humblec @ShyamsundarR @obnoxxx @nixpanic PTAL

Madhu-1:
@j-griffith @humblec @ShyamsundarR @obnoxxx @nixpanic PTAL

pkg/rbd/rbd_util.go
```diff
@@ -96,10 +96,10 @@ var (
 	volumeNameLocker = util.NewIDLocker()
 	// serializes operations based on "snapshot name" as key
 	snapshotNameLocker = util.NewIDLocker()
```
ShyamsundarR: `attachdetachLocker` above can be removed, as this locks the imageName appended with the Pool name, and will be handled by the `nodeVolumeIDLocker` lock.

Madhu-1:
@ShyamsundarR @humblec PTAL

pkg/rbd/nodeserver.go
```diff
@@ -227,17 +281,11 @@ func (ns *NodeServer) createTargetPath(targetPath string, isBlock bool) (bool, e
 // NodeUnpublishVolume unmounts the volume from the target path
 func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
 	targetPath := req.GetTargetPath()
```
ShyamsundarR: Move this line below the validate call below, as validation is ensuring target path is non-empty.

pkg/rbd/nodeserver.go
```diff
@@ -248,21 +296,118 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
-		// TODO should consider deleting path instead of returning error,
-		// once all codes become ready for csi 1.0.
-		return nil, status.Error(codes.NotFound, "volume not mounted")
+		if err = os.RemoveAll(targetPath); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		return &csi.NodeUnpublishVolumeResponse{}, nil
+	}
+
+	// TODO remove this once  we stop supporting legacy volume delete/mount
```
ShyamsundarR: Entire legacy volume handling in Unpublish should not be required. We are not dealing with the RBD image at all in this call.

pkg/rbd/nodeserver.go
```diff
@@ -248,21 +296,118 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+		}
+		if err = ns.unmount(targetPath, devicePath, cnt); err != nil {
+			return nil, err
+		}
+		klog.Infof("rbd: successfully unbinded volume %s from %s", req.GetVolumeId(), targetPath)
+
+		return &csi.NodeUnpublishVolumeResponse{}, nil
+	}
+
+	if _, err = execCommand("umount", []string{targetPath}); err != nil {
```
ShyamsundarR: Shouldn't this use `ns.mounter.Unmount(targetPath)` instead of calling umount directly?

pkg/rbd/nodeserver.go
```diff
@@ -117,101 +116,156 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+func (ns *NodeServer) mountVolumeToStagePath(req *csi.NodeStageVolumeRequest, stagingPath, devicePath string) error {
+	// Publish Path
+	fsType := req.GetVolumeCapability().GetMount().GetFsType()
+	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: mount.NewOsExec()}
+	opt := []string{}
+	isBlock := req.GetVolumeCapability().GetBlock() != nil
+	if isBlock {
+		opt = append(opt, "bind")
+	}
+	err := diskMounter.FormatAndMount(devicePath, stagingPath, fsType, opt)
```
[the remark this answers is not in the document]
Madhu-1: This is not added in this code. please open an issue

pkg/rbd/nodeserver.go
```diff
@@ -117,101 +116,156 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-	// Publish Path
-	err = ns.mountVolume(req, devicePath)
+	// nodeStage Path
+	err = ns.mountVolumeToStagePath(req, stagingTargetPath, devicePath)
 	if err != nil {
 		return nil, err
 	}
-	err = os.Chmod(targetPath, 0777)
+
+	err = os.Chmod(stagingTargetPath, 0777)
```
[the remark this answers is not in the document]
Madhu-1: not added in this PR. please open a separate  issue

pkg/rbd/nodeserver.go
Madhu-1: why it's not required, what happens for the already mapped device in publish call. don't we need to unmap the rbd device?

pkg/rbd/nodeserver.go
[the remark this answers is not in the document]
ShyamsundarR: In the older code there was no mount to the staging target path, and to the target path the mount was conditionally a format and mount only for non-block volumes: https://github.com/ceph/ceph-csi/blob/0f1bee84ff7b91aea5f5234a21764bf26bde830d/pkg/rbd/nodeserver.go#L178-L191

This is introduced by the current changes for the staging path, and for this path we would need to retain the same logic as the older code did for the target path.

pkg/rbd/nodeserver.go
ShyamsundarR: Node publish creates target path and bind mounts the staging path to it, hence on Unpublish we need to revert the same. This is true for legacy volumes as well.

We need not unmap the rbd device or even have to detect if it is a legacy volume in the Unpublish (I think the path is independent of the naming, but there could be a corner case).

The unmap should happen in the Unstage, so the detection and calling of ns.umount in line 339 need not be done in the Unpubish call.

Madhu-1: > Node publish creates target path and bind mounts the staging path to it, hence on Unpublish we need to revert the same. This is true for legacy volumes as well.

how this is true for volumes mounted using v1.0.0  cephcsi? in v1.0.0 targetPath was directly mounted to the device not for the stagingPath right?

> We need not unmap the rbd device or even have to detect if it is a legacy volume in the Unpublish (I think the path is independent of the naming, but there could be a corner case).
> 
> The unmap should happen in the Unstage, so the detection and calling of ns.umount in line 339 need not be done in the Unpubish call.

if the volume and created and mounted using the v1.0.0 (targetpath -> rbddevice)  how to unmount stagingpath from a device is possible?

pkg/rbd/nodeserver.go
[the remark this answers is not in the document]
ShyamsundarR: Opened https://github.com/ceph/ceph-csi/issues/483

pkg/rbd/nodeserver.go
ShyamsundarR: > > Node publish creates target path and bind mounts the staging path to it, hence on Unpublish we need to revert the same. This is true for legacy volumes as well.
> 
> how this is true for volumes mounted using v1.0.0 cephcsi? in v1.0.0 targetPath was directly mounted to the device not for the stagingPath right?
> 
> > We need not unmap the rbd device or even have to detect if it is a legacy volume in the Unpublish (I think the path is independent of the naming, but there could be a corner case).
> > The unmap should happen in the Unstage, so the detection and calling of ns.umount in line 339 need not be done in the Unpubish call.
> 
> if the volume and created and mounted using the v1.0.0 (targetpath -> rbddevice) how to unmount stagingpath from a device is possible?

The above questions are applicable to v1.1.0 volumes as well, if we are dealing with plugin pod upgrades post this change and there are calls to Unstage or Unpublish already mapped/mounted devices by the older plugin version. IOW, upgrade scenarios. My call on upgrade scenarios is that for this change nodes need to be drained and updated (at least for now, till there are other reasons provided).

Once this plugin is rolled out, even for Legacy volumes we would map in Stage and hence would/should unmap in Unstage. Similarly for legacy volumes Publish and Unpublish would not be any different than for non-legacy volumes.

Madhu-1: have we documented steps for upgrade scenarios?

why the drain is required?  this code handles both legacy and other new volumes without a drain.

Madhu-1: node drain with the daemonset upgrade  (With OnDelete update strategy (default one) seems correct idea. updating the code

deploy/rbd/kubernetes/csi-rbdplugin.yaml
```diff
@@ -71,12 +71,6 @@ spec:
           volumeMounts:
             - name: plugin-dir
               mountPath: /csi
-            - name: pods-mount-dir
-              mountPath: /var/lib/kubelet/pods
-              mountPropagation: "Bidirectional"
-            - name: plugin-mount-dir
-              mountPath: /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/
```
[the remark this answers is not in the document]
ShyamsundarR: Any reason to not restrict `plugin-dir` to further include `kubernetes.io/csi`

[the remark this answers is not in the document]
Madhu-1: this is taken from the csi example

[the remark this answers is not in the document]
ShyamsundarR: Ack thanks referred to [this](https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/kubernetes-1.15/hostpath/csi-hostpath-plugin.yaml) through the CSI [examples](https://kubernetes-csi.github.io/docs/deploying.html).

pkg/rbd/nodeserver.go
```diff
@@ -248,21 +299,75 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+			stagingTargetPath = blockStagingPath
+			goto unmount
+		}
+		if err = os.RemoveAll(stagingTargetPath); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		return &csi.NodeUnstageVolumeResponse{}, nil
+	}
+	// Unmount the volume
+	devicePath, cnt, err := mount.GetDeviceNameFromMount(ns.mounter, stagingTargetPath)
```
ShyamsundarR: Code from here on seems to do a few things that may need a revisit with the above hack. For example should we bother about the `cnt` as CO should stage once and unstage it, should it bother us that there are more mounts using the device? Anyway, possibly an issue filed to take a relook at this with the above is required. What are your thoughts?

pkg/rbd/nodeserver.go
ShyamsundarR: Apologies I think this should be `mount.Unmount`. I was looking at inverse of what we use in the Stage/Publish in Unstage/Unpublish, not sure if Unstage earlier was not using ns.mounter.Unmount (as Stage is using ns.mounter.Mount/FormatAndMount), but here the inverse is `mount.Unmount`

pkg/rbd/nodeserver.go
```diff
@@ -41,67 +41,65 @@ type NodeServer struct {
...
-		} else {
-			klog.Warningf("MULTI_NODE_MULTI_WRITER currently only supported with volumes of access type `block`, invalid AccessMode for volume: %v", req.GetVolumeId())
-			return nil, status.Error(codes.InvalidArgument, "rbd: RWX access mode request is only valid for volumes with access type `block`")
-		}
+	if !isNotMnt {
+		klog.Infof("rbd: volume %s is already mounted to %s, skipping", req.GetVolumeId(), stagingTargetPath)
+		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
 	volOptions, err := genVolFromVolumeOptions(req.GetVolumeContext(), req.GetSecrets(), disableInUseChecks, isLegacyVolume)
```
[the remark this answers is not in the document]
Madhu-1: Fixed

pkg/util/validate.go
```diff
@@ -0,0 +1,75 @@
...
+func ValidateNodeStageVolumeRequest(req *csi.NodeStageVolumeRequest) error {
+	if req.GetVolumeCapability() == nil {
+		return status.Error(codes.InvalidArgument, "volume capability missing in request")
+	}
+
+	if req.GetVolumeId() == "" {
+		return status.Error(codes.InvalidArgument, "volume ID missing in request")
+	}
+
+	if req.GetStagingTargetPath() == "" {
```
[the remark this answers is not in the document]
Madhu-1: agree. https://github.com/ceph/ceph-csi/issues/485

pkg/rbd/nodeserver.go
Madhu-1: staging should be done only once per PVC on a perticular node as the staging path is derived from PVC name.


> Code from here on seems to do a few things that may need a revisit with the above hack. For example should we bother about the `cnt` as CO should stage once and unstage it, should it bother us that there are more mounts using the device? Anyway, possibly an issue filed to take a relook at this with the above is required. What are your thoughts?

can we end up in such a situation? if yes let's file an issue and look at it

ShyamsundarR: I suggest we open an issue to keep track of the hack and this issue once we merge this code, so that we do not lose sight of the same.

Madhu-1:
@poornimag  @ajarr

any idea why cephfs is failing?
```
Error EINVAL: Traceback (most recent call last):
  File "/usr/share/ceph/mgr/mgr_module.py", line 909, in _handle_command
    return self.handle_command(inbuf, cmd)
  File "/usr/share/ceph/mgr/volumes/module.py", line 182, in handle_command
    return handler(inbuf, cmd)
  File "/usr/share/ceph/mgr/volumes/module.py", line 208, in _cmd_fs_subvolumegroup_create
    return self.vc.create_subvolume_group(vol_name, group_name, mode=mode, pool=pool_layout)
  File "/usr/share/ceph/mgr/volumes/fs/volume.py", line 305, in create_subvolume_group
    sv.create_group(spec, pool=pool, mode=self.octal_str_to_decimal_int(mode))
  File "/usr/share/ceph/mgr/volumes/fs/subvolume.py", line 178, in create_group
    self._mkdir_p(path, mode)
  File "/usr/share/ceph/mgr/volumes/fs/subvolume.py", line 58, in _mkdir_p
    self.fs.mkdir(subpath, mode)
  File "cephfs.pyx", line 801, in cephfs.LibCephFS.mkdir (/home/jenkins-build/build/workspace/ceph-dev-build/ARCH/x86_64/AVAILABLE_ARCH/x86_64/AVAILABLE_DIST/centos7/DIST/centos7/MACHINE_SIZE/huge/release/14.2.1-636-g64f035c/rpm/el7/BUILD/ceph-14.2.1-636-g64f035c/build/src/pybind/cephfs/pyrex/cephfs.c:8613)
Error: (13, "error in mkdir 'volumes': error code 13")
```

PR #464, opened 2019-07-10
Mark master driver version to `canary` instead of release versions.

humblec:

### DIFF
--- a/pkg/cephfs/driver.go
+++ b/pkg/cephfs/driver.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	// version of ceph driver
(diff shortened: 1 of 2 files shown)


humblec:
@Madhu-1 @poornimag @ShyamsundarR PTAL.

Madhu-1:
@humblec lets update driver version using `ldflags` so that we don't need to make changes in multiple places. even it's good to add git version also.

using `ldflags` will helps in the release also.

Madhu-1:
suppressed by https://github.com/ceph/ceph-csi/pull/469

PR #467, opened 2019-07-12
Add support for standalone snapshot creation

Madhu-1:

with the current implementation in ceph-csi, it's not possible to
delete the cloned volume if the snapshot is present
due to the child linking, to remove this dependency
we had a discussion and come up with an idea to separate
out the clone and snapshot, so that we can delete the
snapshot and cloned image in any order.

The steps followed to create an independent snapshot as follows
## Create a snapshot
* Create  a temporary snapshot from the parent volume
* Clone a new image from a temporary snapshot with options
`--rbd-default-clone-format 2 --image-feature layering,deep-flatten`
* Deletetemprary snapshot created
* Create a snapshot with requested Name

## Create new PVC from a snapshot
* Check the depth of the image as the maximum number of nested volume
clones can be (max 16 can be changed based on the configuration)
if the depth is reached flatten the newly cloned image

## Delete cloned PVC
* Delete the cloned image (earlier we were removing the image with `rbd rm`
command with the  new design we will be moving the images to the trash)
same applies for normal volume deletion also

## Delete snapshot
* Delete the temporary cloned image which was created for a snapshot
* Delete the snapshot

**example commands:-**
```
layering,deep-flatten <RBD image for src  k8s volume>@<random snap <RBD image for temporary snap image>
4) rbd snap create <RBD image for temporary snap image>@<random snap name> <k8s snap name>
5) rbd clone --rbd-default-clone-format 2 --image-feature <k8s dst vol config>
<RBD image for temporary snap image>@<random snap name> <RBD image for k8s dst vol>
```

Do you have any questions?

No backward compatibility with previously created snapshots

Fixes: #issue_number

## Future concerns ##

@dillaman @ShyamsundarR @humblec PTAL

### DIFF
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -47,7 +47,8 @@ var (
 	metadataStorage = flag.String("metadatastorage", "", "metadata persistence method [node|k8s_configmap]")
 
 	// rbd related flags
-	containerized = flag.Bool("containerized", true, "whether run as containerized")
+	containerized    = flag.Bool("containerized", true, "whether run as containerized")
+	rbdMaxCloneDepth = flag.Uint("rbdmaximumclonedepth", 16, "Maximum number of nested volume clones that are taken before a flatten occurs")
 
 	// cephfs related flags
 	volumeMounter = flag.String("volumemounter", "", "default volume mounter (possible options are 'kernel', 'fuse')")
@@ -117,7 +118,7 @@ func main() {
 			}
 		}
 		driver := rbd.NewDriver()
-		driver.Run(dname, *nodeID, *endpoint, *instanceID, *containerized, cp)
+		driver.Run(dname, *nodeID, *endpoint, *instanceID, *containerized, *rbdMaxCloneDepth, cp)
 
 	case cephfsType:
 		cephfs.PluginFolder += dname
--- a/deploy/rbd/helm/templates/provisioner-statefulset.yaml
+++ b/deploy/rbd/helm/templates/provisioner-statefulset.yaml
@@ -88,6 +88,7 @@ spec:
             - "--v=5"
             - "--drivername=$(DRIVER_NAME)"
             - "--containerized=true"
+            - "--rbdmaximumclonedepth=14"
           env:
             - name: HOST_ROOTFS
               value: "/rootfs"
--- a/deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
+++ b/deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
@@ -85,6 +85,7 @@ spec:
             - "--v=5"
             - "--drivername=rbd.csi.ceph.com"
             - "--containerized=true"
+            - "--rbdmaximumclonedepth=14"
           env:
             - name: HOST_ROOTFS
               value: "/rootfs"
--- a/pkg/rbd/controllerserver.go
+++ b/pkg/rbd/controllerserver.go
@@ -17,10 +17,9 @@ limitations under the License.
 package rbd
 
 import (
-	"fmt"
-
 	csicommon "github.com/ceph/ceph-csi/pkg/csi-common"
 	"github.com/ceph/ceph-csi/pkg/util"
+	"github.com/pkg/errors"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
@@ -224,6 +223,12 @@ func (cs *ControllerServer) checkSnapshot(req *csi.CreateVolumeRequest, rbdVol *
 		return status.Error(codes.Internal, err.Error())
 	}
 	klog.V(4).Infof("create volume %s from snapshot %s", req.GetName(), rbdSnap.RbdSnapName)
+
+	// do we need to call this in go routine to make it background job
+	err = flattenRbdImage(rbdVol, rbdMaxCloneDepth, cr)
+	if err != nil {
+		klog.Errorf("failed to  flatten image %v", err)
+	}
 	return nil
 }
 
@@ -385,22 +390,11 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-
 	// Fetch source volume information
-	rbdVol := new(rbdVolume)
-	err = genVolFromVolID(rbdVol, req.GetSourceVolumeId(), cr)
+	rbdVol, err := validateVolHasSnapFeature(req.GetSourceVolumeId(), cr)
 	if err != nil {
-		if _, ok := err.(ErrImageNotFound); ok {
-			return nil, status.Errorf(codes.NotFound, "source Volume ID %s not found", req.GetSourceVolumeId())
-		}
-		return nil, status.Errorf(codes.Internal, err.Error())
+		return nil, err
 	}
-
-	// Check if source volume was created with required image features for snaps
-	if !hasSnapshotFeature(rbdVol.ImageFeatures) {
-		return nil, status.Errorf(codes.InvalidArgument, "volume(%s) has not snapshot feature(layering)", req.GetSourceVolumeId())
-	}
-
 	// Create snap volume
 	rbdSnap := genSnapFromOptions(rbdVol, req.GetParameters())
 	rbdSnap.RbdImageName = rbdVol.RbdImageName
@@ -411,8 +405,17 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	idLk := snapshotNameLocker.Lock(req.GetName())
 	defer snapshotNameLocker.Unlock(idLk, req.GetName())
 
-	// Need to check for already existing snapshot name, and if found
-	// check for the requested source volume id and already allocated source volume id
+	// updating the rbdImage name to point to temparory cloned image name, this
+	// is to check if the snapshot is already present or not
+	img, err := getCloneImageName(rbdSnap, cr)
+	if err != nil {
+		if _, ok := err.(ErrImageNotFound); !ok {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	} else {
+		rbdSnap.RbdImageName = img
+	}
+
 	found, err := checkSnapExists(rbdSnap, cr)
 	if err != nil {
 		if _, ok := err.(util.ErrSnapNameConflict); ok {
@@ -433,12 +436,20 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		}, nil
 	}
 
+	// this will be used to unreserve snapshot if we failed to get snapshot
+	// information from the ceph backend
+	updatedRbdImage := ""
+	// update the rbd image name to clone from parent volume
+	rbdSnap.RbdImageName = rbdVol.RbdImageName
 	err = reserveSnap(rbdSnap, cr)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	defer func() {
 		if err != nil {
+			if updatedRbdImage != "" {
+				rbdSnap.RbdImageName = updatedRbdImage
+			}
 			errDefer := undoSnapReservation(rbdSnap, cr)
 			if errDefer != nil {
 				klog.Warningf("failed undoing reservation of snapshot: %s %v", req.GetName(), errDefer)
@@ -446,9 +457,44 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		}
 	}()
 
-	err = cs.doSnapshot(rbdSnap, cr)
+	// generate temp snap struct and vol struct to create a temprory snapshot
+	// and to create a new volume which will be used to take new snapshot
+	tmpSnap := &rbdSnapshot{}
+	copySnapStruct(tmpSnap, rbdSnap)
+
+	// using request name as the temp snapshot request name. this will help us to
+	// identify if the temp snap is already created.
+	// this will look like `csi-tmp-snap-1-PVC-XXXX-XXX-XXXX`
+	tmpSnap.RequestName = snapJournal.GetTmpNamePrefix(req.GetName(), true, true)
+	// generate tempVolume struct
+	tmpVol := generateVolFromSnap(rbdSnap)
+	tmpVol.ImageFeatures = "layering,deep-flatten"
+	// using request name as the volume request name. this will help us to
+	// identify if the temp snap is already created.
+	// this will look like `csi-tmp-clone-PVC-XXXX-XXX-XXXX`
+	tmpVol.RequestName = volJournal.GetTmpNamePrefix(req.GetName(), true, false)
+	err = cs.createSnapFromClone(tmpVol, tmpSnap, cr)
 	if err != nil {
 		return nil, err
+	}
+	// update the parent image to create snapshot
+	rbdSnap.RbdImageName = tmpVol.RbdImageName
+	err = createSnapshot(rbdSnap, cr)
+	if err != nil {
+		return nil, err
+	}
+	err = updateReservedSnap(rbdSnap, cr)
+	if err != nil {
+		klog.Errorf("failed to update snapshot parent name: %v", err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	updatedRbdImage = rbdSnap.RbdImageName
+
+	err = getSnapshotMetadata(rbdSnap, cr)
+	if err != nil {
+		klog.Errorf("failed to fetch snapshot metadata: %v", err)
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 
 	return &csi.CreateSnapshotResponse{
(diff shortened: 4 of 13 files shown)


Madhu-1:
@dillaman am facing one issue while mounting the cloned PVC
logs 
```
E0712 05:14:52.497014    7792 utils.go:109] GRPC error: rbd: map failed exit status 6, rbd output: 2019-07-12 05:14:52.390 7f463497cb00 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
rbd: sysfs write failed
2019-07-12 05:14:52.432 7f463497cb00 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-12 05:14:52.433 7f463497cb00 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-12 05:14:52.433 7f463497cb00 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-12 05:14:52.439 7f463497cb00 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-12 05:14:52.439 7f463497cb00 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2019-07-12 05:14:52.439 7f463497cb00 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
RBD image feature set mismatch. You can disable features unsupported by the kernel with "rbd feature disable replicapool/csi-vol-d8f76c11-a463-11e9-8a0c-c2953b9dc10a".
In some cases useful info is found in syslog - try "dmesg | tail".
```

rbd volume Info
```
sh-4.2# rbd ls --pool=replicapool
csi-vol-c2af1e5a-a463-11e9-8a0c-c2953b9dc10a (Parent volume)
csi-vol-cba8d8ce-a463-11e9-8a0c-c2953b9dc10a (Temp clone volume)
csi-vol-d8f76c11-a463-11e9-8a0c-c2953b9dc10a ( K8s dest volume)
sh-4.2# rbd info csi-vol-c2af1e5a-a463-11e9-8a0c-c2953b9dc10a --pool=replicapool
rbd image 'csi-vol-c2af1e5a-a463-11e9-8a0c-c2953b9dc10a':
	size 1 GiB in 256 objects
	order 22 (4 MiB objects)
	snapshot_count: 1
	id: 39855c9906c
	block_name_prefix: rbd_data.39855c9906c
	format: 2
	features: layering, operations
	op_features: clone-parent, snap-trash
	flags: 
	create_timestamp: Fri Jul 12 05:13:22 2019
	access_timestamp: Fri Jul 12 05:13:22 2019
	modify_timestamp: Fri Jul 12 05:13:22 2019


sh-4.2# rbd info csi-vol-cba8d8ce-a463-11e9-8a0c-c2953b9dc10a --pool=replicapool
rbd image 'csi-vol-cba8d8ce-a463-11e9-8a0c-c2953b9dc10a':
	size 1 GiB in 256 objects
	order 22 (4 MiB objects)
	snapshot_count: 1
	id: 39d056ad63b2
	block_name_prefix: rbd_data.39d056ad63b2
	format: 2
	features: layering, deep-flatten, operations
	op_features: clone-parent, clone-child
	flags: 
	create_timestamp: Fri Jul 12 05:13:34 2019
	access_timestamp: Fri Jul 12 05:13:34 2019
	modify_timestamp: Fri Jul 12 05:13:34 2019
	parent: replicapool/csi-vol-c2af1e5a-a463-11e9-8a0c-c2953b9dc10a@219fc232-148d-43ff-bdf5-d970a6d464bf
	overlap: 1 GiB
sh-4.2# rbd info csi-vol-d8f76c11-a463-11e9-8a0c-c2953b9dc10a --pool=replicapool
rbd image 'csi-vol-d8f76c11-a463-11e9-8a0c-c2953b9dc10a':
	size 1 GiB in 256 objects
	order 22 (4 MiB objects)
	snapshot_count: 0
	id: 3a36eb833854
	block_name_prefix: rbd_data.3a36eb833854
	format: 2
	features: layering, operations
	op_features: clone-child
	flags: 
	create_timestamp: Fri Jul 12 05:13:57 2019
	access_timestamp: Fri Jul 12 05:13:57 2019
	modify_timestamp: Fri Jul 12 05:13:57 2019
	parent: replicapool/csi-vol-cba8d8ce-a463-11e9-8a0c-c2953b9dc10a@csi-snap-ca6c7cf3-a463-11e9-8a0c-c2953b9dc10a
	overlap: 1 GiB

```

dillaman:
@Madhu-1 sorry for the delay, I didn't see the github notification email.

> am facing one issue while mounting the cloned PVC logs

What kernel are you using? v4.16 (and later) should ignore the "operations" feature flag since it's a no-op for krbd. RHEL 8 should be based on at least v4.18 (I believe).

Madhu-1:
@dillaman am using `CentOS Linux release 7.6.1810 (Core)` kernel version is `3.10.0-957.21.3.el7.x86_64`

also, need to check what Travis CI  is  using

dillaman:
@Madhu-1 OK, so in addition to the Ceph release restriction for PV cloning (and standalone snapshot creation as the intermediary step), we need to have a semi-recent kernel (like RHEL 8 or CentOS 7 w/ kernel-ml)

dillaman:
@Madhu-1 Looks like travis would be using Xenial [1]. Bionic should be on the v4.15 kernel and that feature was backported to v4.15 [2], so it might be worth a test.

[1] https://github.com/ceph/ceph-csi/blob/master/.travis.yml#L4
[2] https://lwn.net/Articles/747916/

Madhu-1:
@dillaman Thanks for the hint, I have tested it locally it's working with 5.2.1 kernel centos (i will test with Travis with bionic )

Madhu-1:
up for initial review

cmd/cephcsi.go
```diff
@@ -47,7 +47,8 @@ var (
 	metadataStorage = flag.String("metadatastorage", "", "metadata persistence method [node|k8s_configmap]")
 
 	// rbd related flags
-	containerized = flag.Bool("containerized", true, "whether run as containerized")
+	containerized    = flag.Bool("containerized", true, "whether run as containerized")
+	rbdMaxCloneDepth = flag.Uint("rbdmaximumclonedepth", 16, "Maximum number of nested volume clones that are taken before a flatten occurs")
```
dillaman: Nit: while 16 is technically the maximum, I think it should be set much lower before a flatten is required since a deeper depth will impact IO speeds. For example, Cinder uses a max depth of 5 [1] so perhaps copy that as a starting point. 

Additionally, in the future it might be nice to have a soft vs hard depth limit (i.e. when you hit the soft-depth, it doesn't block PV/snapshot creation but it kicks off a background flatten whereas the hard limit does block until the flatten is complete). In that case, can we name this something like `rbdhardmaxclonedepth`)?

[1] https://github.com/openstack/cinder/blob/master/cinder/volume/drivers/rbd.py#L83

deploy/rbd/helm/templates/provisioner-statefulset.yaml
```diff
@@ -88,6 +88,7 @@ spec:
             - "--v=5"
             - "--drivername=$(DRIVER_NAME)"
             - "--containerized=true"
+            - "--rbdmaximumclonedepth=14"
```
dillaman: Nit: leave it at its default and don't override?

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
```diff
@@ -85,6 +85,7 @@ spec:
             - "--v=5"
             - "--drivername=rbd.csi.ceph.com"
             - "--containerized=true"
+            - "--rbdmaximumclonedepth=14"
```
dillaman: Nit: same comment here

pkg/util/voljournal.go
```diff
@@ -160,6 +170,36 @@ func (cj *CSIJournal) SetNamespace(ns string) {
...
+
+// GetTmpNamePrefix returns the temporary name prefix
+/* if first is set
+if the tmpName is csi-tmp-snap-rbd.ceph.csi.com then it will return
+csi-tmp-snap1-rbd.ceph.csi.com
+if  first is not set returns csi-tmp-snap2-rbd.ceph.csi.com
+this will help us storing the snap info on rados and also helps in cleaning the
+stale snaps
+*/
+func (cj *CSIJournal) GetTmpNamePrefix(sufix string, first, isSnap bool) string {
```
dillaman: Nit: `suffix` instead of `sufix`

dillaman: Might be better to split into two functions -- one for volumes and the other for snaps. Having to pass the `isSnap` param just to pick an entirely different logic path seems odd to me.

pkg/util/voljournal.go
```diff
@@ -160,6 +170,36 @@ func (cj *CSIJournal) SetNamespace(ns string) {
 	cj.namespace = ns
 }
 
+// SetTmpNamePrefix updates the temp name in which all temoprary cloned volume
```
dillaman:
```suggestion
// SetTmpNamePrefix updates the temp name in which all temporary cloned volume
```

pkg/util/voljournal.go
```diff
@@ -160,6 +170,36 @@ func (cj *CSIJournal) SetNamespace(ns string) {
...
+*/
+func (cj *CSIJournal) GetTmpNamePrefix(sufix string, first, isSnap bool) string {
+	if !isSnap {
+		return cj.tmpNamePrefix + sufix
+	}
+	t := strings.Split(cj.tmpNamePrefix, tmpSnapName)
+	if len(t) != 2 {
+		return ""
+	}
+	if first {
```
dillaman: What is the purpose of `first`? It seems like it's always true and I don't understand the need to inject "1" and "2" into the temporary names.

pkg/rbd/controllerserver.go
```diff
@@ -462,66 +508,128 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	tmpSnap.RbdImageName = rbdSnap.RbdImageName
+	tmpSnap.RequestName = rbdSnap.RequestName
+	tmpSnap.SizeBytes = rbdSnap.SizeBytes
+	tmpSnap.SourceVolumeID = rbdSnap.SourceVolumeID
+}
 
-	// Check sanity of request Snapshot Name, Source Volume Id
-	if req.Name == "" {
-		return status.Error(codes.InvalidArgument, "snapshot Name cannot be empty")
+func getCloneImageName(rbdSnap *rbdSnapshot, cr *util.Credentials) (string, error) {
```
dillaman: `getCloneImageName` -> `getSnapCloneImageName`?

pkg/rbd/controllerserver.go
```diff
@@ -411,8 +405,17 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	idLk := snapshotNameLocker.Lock(req.GetName())
 	defer snapshotNameLocker.Unlock(idLk, req.GetName())
 
-	// Need to check for already existing snapshot name, and if found
-	// check for the requested source volume id and already allocated source volume id
+	// updating the rbdImage name to point to temparory cloned image name, this
```
dillaman:
```suggestion
	// updating the rbdImage name to point to temparory cloned image name, this
```

pkg/rbd/controllerserver.go
```diff
@@ -411,8 +405,17 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	img, err := getCloneImageName(rbdSnap, cr)
+	if err != nil {
+		if _, ok := err.(ErrImageNotFound); !ok {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	} else {
+		rbdSnap.RbdImageName = img
+	}
+
 	found, err := checkSnapExists(rbdSnap, cr)
```
dillaman: If the "snapshot" RBD image wasn't found above, why do we need to check if the snapshot exists in the metadata? It just seems odd how the `rbdSnap.RbdImageName` field is being manipulated. Maybe if it wasn't set on line 400 and reset on line 443 it would be clearer. It seems like it should always only be set to the snapshot image name.

pkg/rbd/controllerserver.go
```diff
@@ -433,23 +436,66 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+			}
 			errDefer := undoSnapReservation(rbdSnap, cr)
 			if errDefer != nil {
 				klog.Warningf("failed undoing reservation of snapshot: %s %v", req.GetName(), errDefer)
 			}
 		}
 	}()
 
-	err = cs.doSnapshot(rbdSnap, cr)
+	// generate temp snap struct and vol struct to create a temprory snapshot
```
dillaman:
```suggestion
	// generate temp snap struct and vol struct to create a temporary snapshot
```

pkg/rbd/controllerserver.go
```diff
@@ -462,66 +508,128 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		}
+		return snapErr
+	}
+	err = reserveVol(rbdVol, cr)
 	if err != nil {
-		klog.Errorf("failed to fetch snapshot metadata: %v", err)
 		return status.Error(codes.Internal, err.Error())
 	}
 
+	volErr = restoreSnapshot(rbdVol, rbdSnap, cr)
```
dillaman: Nit: can we rename `restoreSnapshot` to clarify that it's actually just performing an "rbd clone"? i.e. `cloneSnapshot`?

pkg/rbd/controllerserver.go
```diff
@@ -433,23 +436,66 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
 			if errDefer != nil {
 				klog.Warningf("failed undoing reservation of snapshot: %s %v", req.GetName(), errDefer)
 			}
 		}
 	}()
 
-	err = cs.doSnapshot(rbdSnap, cr)
+	// generate temp snap struct and vol struct to create a temprory snapshot
+	// and to create a new volume which will be used to take new snapshot
+	tmpSnap := &rbdSnapshot{}
```
dillaman: Perhaps just my opinion, but it's confusing that `rbdSnapshot` and `rbdVolume` are being re-used to describe internal RBD images / snapshots that don't have associated CSI snapshots and volumes. It might be nice to re-org these structures to a parent struct that is only the RBD data and a derived class that adds CSI request related fields.

pkg/rbd/controllerserver.go
```diff
@@ -572,21 +680,42 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
...
 	klog.V(4).Infof("deleting Snaphot %s", rbdSnap.RbdSnapName)
 	if err := deleteSnapshot(rbdSnap, cr); err != nil {
-		return nil, status.Errorf(codes.FailedPrecondition,
+		return nil, status.Errorf(codes.Internal,
 			"failed to delete snapshot: %s/%s with error: %v",
 			rbdSnap.Pool, rbdSnap.RbdSnapName, err)
 	}
 
+	if found {
+		// TODO need to delete stale volumes
```
dillaman: What are stale volumes?

pkg/rbd/rbd_util.go
```diff
@@ -177,8 +177,8 @@ func deleteImage(pOpts *rbdVolume, cr *util.Credentials) error {
 		return fmt.Errorf("rbd %s is still being used", image)
 	}
 
-	klog.V(4).Infof("rbd: rm %s using mon %s, pool %s", image, pOpts.Monitors, pOpts.Pool)
-	args := []string{"rm", image, "--pool", pOpts.Pool, "--id", cr.ID, "-m", pOpts.Monitors,
+	klog.V(4).Infof("rbd: trash mv %s using mon %s, pool %s", image, pOpts.Monitors, pOpts.Pool)
+	args := []string{"trash", "mv", image, "--pool", pOpts.Pool, "--id", cr.ID, "-m", pOpts.Monitors,
```
dillaman: This will need to be combined w/ the "ceph rbd trash remove" call otherwise we will just be leaking images to the trash.

pkg/rbd/controllerserver.go
```diff
@@ -433,23 +436,66 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
 	if err != nil {
 		return nil, err
 	}
+	// update the parent image to create snapshot
+	rbdSnap.RbdImageName = tmpVol.RbdImageName
+	err = createSnapshot(rbdSnap, cr)
+	if err != nil {
+		return nil, err
+	}
+	err = updateReservedSnap(rbdSnap, cr)
```
dillaman: Why do we need to update the metadata? Shouldn't we know all the finalized names before we start the process?

pkg/rbd/controllerserver.go
Madhu-1: at line 400 am setting the rbd image name to parent Volume from which the snapshot has to be created.

as the snapshot will be created from a cloned image, not the parent image (i need to set the snaps rbd name to temp cloned image and check if the snapshot is already present )

so am checking if the temp image at line 410, if its present am updating the snapshot parent image name as temp cloned image name at line 416 and checking if snapshot already created

as snapshot parent image name is updated to temp image at line 416. at line 443 resetting back to the parent vol.

pkg/rbd/rbd_util.go
Madhu-1: @dillaman as you know currently we are doing `rbd task add remove` to delete the image from backend

as with this implementation do I need to replace  `rbd task add remove` with `rbd trash mv` and `rbd task add trash remove` to delete image from backend?

if the user does not have the manager permission is `rbd trash mv` enough?

dillaman: I think in a perfect world if `rbd task add trash remove` fails the CSI should rollback and fail the CO request. Otherwise, you are just leaking images into the trash and if folks don't know to go looking in the trash, they will just cause a hidden problem down the road.

dillaman: You would need to delete the clone and run `rbd trash rm`

humblec:
@Madhu-1 can you revisit this PR ?

Issue #449, opened 2019-06-27
Can the CSI RBD plugin catch up in terms of performance to native CLI calls?


ShyamsundarR:
As a part of the discussion in PR https://github.com/ceph/ceph-csi/pull/443 it was noted by @dillaman [that native CLI calls](https://github.com/ceph/ceph-csi/pull/443#issuecomment-505135380) outperform the plugin by a large factor.

As part of the discussion I [noted](https://github.com/ceph/ceph-csi/pull/443#issuecomment-505172783) that the plugin does more work than just an image create, and that there would be kubernetes factors to consider.

As a result this issue is opened to help with the analysis of native calls versus calls made by the plugin to understand and maybe improve the performance of the plugin.

Various experiments may need to be conducted to arrive at an answer, and this issue can hopefully help track the progress.

ShyamsundarR:
First test, to compare numbers generated using PR https://github.com/ceph/ceph-csi/pull/443 and the script below, that generate exact nature and order of calls as the plugin, is provided in this comment.

Script to mimic what the plugin does as of commit hash `59d3365d3b13f37a9b914d59d931c71269a21eec` on master: [raw-rbd-perf.sh.txt](https://github.com/ceph/ceph-csi/files/3336340/raw-rbd-perf.sh.txt)

Note: Script maybe a bit rough but works, execute a create like so `# time ./raw-rbd-perf.sh create 20` and follow with a delete like so `# time ./raw-rbd-perf.sh delete 20`, repeat for multiple runs.

Note: Following numbers were generated on the same setup as numbers generated in [this](https://github.com/ceph/ceph-csi/pull/443#issuecomment-505129896) comment. Further the script was executed from within the container csi-rbdplugin in the pod csi-rbdplugin-provisioner-0, to keep things as close to the plugin.

Script times:
Executed 20 creates using the script, Run1 `real	0m32.030s` Run2 `real	0m38.861s`
Executed 20 deletes using the script, Run1 `real	0m29.651s` Run2 `real	0m28.819s`

Further, modified the script to not run the create/delete in parallel and instead do a serial create/delete (just for a baseline), and following are the results from the same,
Executed 20 **serial** creates using the script, Run1 `real	2m0.101s` Run2 `real	2m48.814s`
Executed 20 **serial** deletes using the script, Run1 `real	4m25.355s` Run2 `real	3m56.247s`

**Observations:**
[...]

PR #473, opened 2019-07-13
Use correct file descriptor to parse errors

ShyamsundarR:
File descriptors in use to parse errors from a few command
invocations were incorrect. This led to inability to detect
certain errors cases and act accordingly.

One of the easiest noticeable issues was when an image is deleted
but its RADOS keys and maps are still intact. In such cases
the DeleteVolume call always errored out unable to find the
image rather than, proceed with cleaning up the RADOS objects
and returning a success.

The original method of using stdout was incorrect, as the command
was tested from within a shell script and the scripts STDIN/OUT/ERR
was redirected to understand behavior. This is now tested using just
the CLI in question, and also examining Ceph code, and further
testing a couple of edge conditions by deleting backing images
for PVs

### DIFF
--- a/pkg/rbd/rbd_util.go
+++ b/pkg/rbd/rbd_util.go
@@ -620,7 +620,7 @@ func getImageInfo(monitors string, cr *util.Credentials, poolName, imageName str
 
 	var imgInfo imageInfo
 
-	stdout, _, err := util.ExecCommand(
+	stdout, stderr, err := util.ExecCommand(
 		"rbd",
 		"-m", monitors,
 		"--id", cr.ID,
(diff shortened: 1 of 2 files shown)


mykaul:
The real solution is not to parse CLI output, but rely on some stable API, no?

ShyamsundarR:
> The real solution is not to parse CLI output, but rely on some stable API, no?

The final solution is to not use CLI, even from a per-PV operations performance standpoint, as stated and tracked [here](https://github.com/ceph/ceph-csi/issues/449#issuecomment-506485209).

Currently the code relies on CLI for **all** purposes, and the error stream reading and parsing needs a fix.

humblec:
@Madhu-1 PTAL.

PR #492, opened 2019-07-24
Temperory workaround for the regression to pass

poornimag:
Currently the cephfs PVC create/delete and all other operations
related to cephfs are failing. One of the recent commits in rook
900abbc967e108ad622648b740a7c57f1268209f has modified ceph-mgr
to run as ceph user rather than root user. The ceph user currently
has no permission to write to the root of the cephfs filesystem.
The fix will be external to CSI itself, but until that lands, sending
a workaround patch so the CSI CI is unblocked.

In this patch, we are setting the permission 777 on root of the cephfs
filesystem. Thus ceph user will be able to modify the cephfs filesystem.

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -745,3 +745,17 @@ func listSnapshots(f *framework.Framework, pool, imageName string) ([]snapInfo, 
 	err := json.Unmarshal([]byte(stdout), &snapInfos)
 	return snapInfos, err
 }
+
+func GivePermToCephfsRoot(f *framework.Framework) {
+	opt := metav1.ListOptions{
+		LabelSelector: "app=rook-ceph-tools",
+	}
+	out := execCommandInPod(f, "mkdir /mnt/cephfs", rookNS, &opt)
+	e2elog.Logf("Creating temperory mount point in tools pod %s", out)
+
+	out = execCommandInPod(f, "ceph-fuse /mnt/cephfs >/dev/null 2>&1", rookNS, &opt)
+	e2elog.Logf("Mounting ceph file system at /mnt/cephfs %s", out)
+
+	out = execCommandInPod(f, "chmod 777 /mnt/cephfs/", rookNS, &opt)
+	e2elog.Logf("Setting chmod 777 on the cepfs root %s", out)
+}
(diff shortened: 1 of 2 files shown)


e2e/utils.go
```diff
@@ -745,3 +745,24 @@ func listSnapshots(f *framework.Framework, pool, imageName string) ([]snapInfo,
 	err := json.Unmarshal([]byte(stdout), &snapInfos)
 	return snapInfos, err
 }
+
+func GivePermToCephfsRoot(f *framework.Framework) {
+	opt := metav1.ListOptions{
+		LabelSelector: "app=rook-ceph-tools",
+	}
+	err := execCommandInPod(f, "mkdir /mnt/cephfs", rookNS, &opt)
```
Madhu-1: @poornimag can you make `/mntcephfs` as const 
Thanks for the Fix :)

Madhu-1: @poornimag one question do we need to create `/mnt/cephfs` ? or this will be created by rook operator ? and when this will be created

poornimag: > @poornimag can you make `/mnt/cephfs` as const
> Thanks for the Fix :)

Np:) Sure will do in the next update

poornimag: > @poornimag one question do we need to create `/mnt/cephfs` ? or this will be created by rook operator ? and when this will be created

This is created in tool pod for temperory mount of the cephfs filesystem. This happens before each in cephfs. Rook will not do this, this patch will be reverted once the actual fix is in place, so we need not do it in rook IMO.

humblec:
Thanks @poornimag !! Much appreciated for a quick action here.

e2e/utils.go
```diff
@@ -136,6 +136,31 @@ func execCommandInPod(f *framework.Framework, c, ns string, opt *metav1.ListOpti
 	return stdOut
 }
 
+func execCommandInToolPod(f *framework.Framework, c, ns string, opt *metav1.ListOptions) string {
```
Madhu-1: @poornimag use `execCommandInPod` function.

poornimag: Using that fails the command "ceph-fuse" i think though it is successful the stderr is still present. Hence copied it to another function and got rid of check for err and stderr empty. Since it is short lived didn't combine the functions.

humblec:
@poornimag Thanks ! @Madhu-1 any further comments from your side ?

Madhu-1:
@humblec please approve and unblock other PR's

humblec:
> @humblec please approve and unblock other PR's

Sure. .Done.

humblec:
> @humblec please approve and unblock other PR's

Sure. .Done.

PR #497, opened 2019-07-25
Enable leader election in v1.14+

Madhu-1:
Use Deployment with leader election instead of StatefulSet

Deployment behaves better when a node gets disconnected
from the rest of the cluster - new provisioner leader
is elected in ~15 seconds, while it may take up to
5 minutes for StatefulSet to start a new replica.

Refer: kubernetes-csi/external-provisioner@52d1fbc


Madhu-1:
@humblec why DNM label is added here?

humblec:
@Madhu-1 was added by keeping something in mind. I have removed it . :)

Issue #499, opened 2019-07-25
cephfs: do not set subvolumegroup's pool layout


ajarr:
Currently, while creating CSI volumes, the subvolume group `csi` the pool layout is set based on the pool setting in the storageclass.
https://github.com/ceph/ceph-csi/commit/32ea550e3#diff-33a13f3231aad3a3f5ec45f308007699R78
The storage class's pool setting is meant to change the pool layout of CSI volumes/CephFS subvolumes and not that of the `csi` subvolume group, which may contain subvolumes of different storage class types. The `csi` subvolume group's pool layout should not be set. It should just be the default. Instead, set the pool layout of the FS subvolume within the subvolume group.

Initially Reported by @ShyamsundarR

humblec:
@poornimag

ajarr:
@humblec , can I get the permissions to just set the labels, and assignee?

humblec:
> @humblec , can I get the permissions to just set the labels, and assignee?

You are in!! Please feel free to send PRs too!

ajarr:
> Please feel free to send PRs too! :)

ah! ;-)  I'll try.

FYI, I'm creating issues so that upstream development effort can be better tracked in the agile model.

PR #502, opened 2019-07-25
cephfs: set pool layout and the mode when creating fs subvolumes and not when creating subvolume group `csi`

ajarr:
Fixes: #499


humblec:
@ajarr why are we not using `mode` in the arg ?

ajarr:
@poornimag PTAL

ajarr:
> @ajarr why are we not using `mode` in the arg ?

I'm not sure why the subvolume group's mode needs to be `777`.  I get that the subvolumes' mode need to be `777`.

Daniel-Pivonka:
doesn't the pool layout flag still need to be set when creating the subvolume after creating the subvolumegroup?

ajarr:
> doesn't the pool layout flag still need to be set when creating the subvolume after creating the subvolumegroup?

True. That's being done here, 

> doesn't the pool layout flag still need to be set when creating the subvolume after creating the subvolumegroup?

ah! good catch. I overlooked it being set in the fs subvolume create

ajarr:
@poornimag can you merge this once you approve?

Issue #487, opened 2019-07-22
Allow custom volume prefix for Ceph {FS && RBD }CSI volumes


humblec:
# Describe the feature you'd like to have #

What new functionality do you want?

As an admin I would like to create custom volume name prefixed volumes via Ceph CSI. There is an option for `volumenameprefix` per driver, but that is not getting tracked here, rather by this issue, we are looking for an option to specify volumename prefixes via SC for CephFS and RBD volumes.

# What is the value to the end user? (why is it a priority?) #

In a setup where you have many volumes, by this functionality admin can filter volumes based on this prefix.

# How will we know we have a good solution? (acceptance criteria) #

Create volumes with different prefixes and list/filter the volumes on a particular filter.

Madhu-1:
same applies for a snapshot also

humblec:
> same applies for a snapshot also

Sure, we could start with volumes first.

ShyamsundarR:
Is it not feasible to just filter PVs provisioned using a named storage class, and then using the `volumeID` to list the image names backing these PVs?

This way the same scripts/applications that help trace a PVC all the way to a image can be reused with this and other such use-cases. Also, removes the need to add extra parameters to the storage class to track names or other such information.

Tagging @phlogistonjohn and @raghavendra-talur for thoughts on this.

raghavendra-talur:
@ShyamsundarR @Madhu-1 @humblec IMO, overloading the name with a filtering requirement would be less flexible thing to do. Based on what we have learnt from gluster and heketi we know:

1. Getting a list of PVs and PVCs backed by a storageclass is very easy both using the API and by using go templates in cli.
2. It is much more flexible if such attributes which may be required for filtering are placed as key/value pairs(labels) on the resources than prefixing/suffixing them in the name. 

If the labels are stored in both the kubernetes side(in PVs spec) and *also* passed down to the ceph as some user provided attributes it would tie things up very well.

skoef:
I would like to see such a feature as well. I'm not interested in filtering the PVs whatsoever, I need such a feature to integrate with an already existing ceph management plane we have running. I think there are two things we could understand by custom volume prefixes:

1. a `--volume-name-prefix` flag, just like the CSI provisioner has. This would only allow you to customise the default `csi-vol-` prefix, applied to all volumes.
[...]

PR #508, opened 2019-07-26
Add support for custom volume/snapshot name prefix

Madhu-1:
if the user provides a volume/snapshot name prefix via storage/snapshot class.we will be creating
the backend volume/snapshot with the prefix name, this will help admin to easily check which image
belongs to which storage class.

Fixes: #487 

### DIFF
--- a/examples/rbd/storageclass.yaml
+++ b/examples/rbd/storageclass.yaml
@@ -24,6 +24,10 @@ parameters:
    # CSI RBD currently supports only `layering` feature.
    imageFeatures: layering
 
+   # With volumenameprefix option an admin can now prefix the desired volume
+   # name from storageclass.
+   # volumenameprefix: custom-vol-name
+
    # The secrets have to contain Ceph credentials with required access
    # to the 'pool'.
    csi.storage.k8s.io/provisioner-secret-name: csi-rbd-secret
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -19,6 +19,7 @@ package util
 import (
 	"os"
 	"path"
+	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -27,6 +28,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/util/mount"
+)
+
+var (
+	volumeNameRE = regexp.MustCompile("^[a-zA-Z0-9_-]+$")
 )
 
 // remove this once kubernetes v1.14.0 release is done
@@ -99,27 +104,6 @@ func ValidateDriverName(driverName string) error {
 	return err
 }
 
-// GenerateVolID generates a volume ID based on passed in parameters and version, to be returned
-// to the CO system
-func GenerateVolID(monitors string, cr *Credentials, pool, clusterID, objUUID string, volIDVersion uint16) (string, error) {
-	poolID, err := GetPoolID(monitors, cr, pool)
-	if err != nil {
-		return "", err
-	}
-
-	// generate the volume ID to return to the CO system
-	vi := CSIIdentifier{
-		LocationID:      poolID,
-		EncodingVersion: volIDVersion,
-		ClusterID:       clusterID,
-		ObjectUUID:      objUUID,
-	}
-
-	volID, err := vi.ComposeCSIID()
-
-	return volID, err
-}
-
 // CreateMountPoint creates the directory with given path
 func CreateMountPoint(mountPath string) error {
 	return os.MkdirAll(mountPath, 0750)
@@ -140,4 +124,9 @@ func IsMountPoint(p string) (bool, error) {
 func Mount(source, target, fstype string, options []string) error {
 	dummyMount := mount.New("")
 	return dummyMount.Mount(source, target, fstype, options)
+}
+
+// isValidName validates Volume name
+func isValidName(name string) bool {
+	return volumeNameRE.MatchString(name)
 }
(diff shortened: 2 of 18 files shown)


humblec:
Thanks @Madhu-1 . Please feel free to update the example output.

examples/rbd/storageclass.yaml
```diff
@@ -24,6 +24,10 @@ parameters:
    # CSI RBD currently supports only `layering` feature.
    imageFeatures: layering
 
+   # With volumenameprefix option an admin can now prefix the desired volume
+   # name from storageclass.
+   # volumenameprefix: custom-vol-name
```
dillaman: Nit: seems this should be camelCase?

pkg/util/util.go
```diff
@@ -141,3 +125,8 @@ func Mount(source, target, fstype string, options []string) error {
 	dummyMount := mount.New("")
 	return dummyMount.Mount(source, target, fstype, options)
 }
+
+// isValidName validates Volume name
+func isValidName(name string) bool {
```
dillaman: Nit: perhaps a more descriptive name is `isValidNamingPrefix`?

pkg/util/volid.go
```diff
@@ -60,20 +71,39 @@ Version 1 of the encoding scheme is as follows,
...
+	snapNamePrefix   = "snapshotnameprefix"
 )
 
+// ComposeCSIID composes a CSI ID from passed in parameters
 func (ci CSIIdentifier) ComposeCSIID() (string, error) {
 	buf16 := make([]byte, 2)
 	buf64 := make([]byte, 8)
 
-	if (knownFieldSize + len(ci.ClusterID)) > maxVolIDLen {
+	ci.EncodingVersion = EncodingV2
```
dillaman: Nit: should you default to V1 unless a custom naming prefix is used? I am thinking about rolling updates when now it would be possible for new V2 names to be created and attachers might not be able to decode them.

pkg/util/util.go
```diff
@@ -29,6 +30,10 @@ import (
 	"k8s.io/kubernetes/pkg/util/mount"
 )
 
+var (
+	volumeNameRE = regexp.MustCompile("^[a-zA-Z0-9_-]+$")
```
dillaman: Should you put a max length on this? The V2 schema won't work if the prefix is longer than X characters.

examples/rbd/storageclass.yaml
Madhu-1: @dillaman this is to support feature parity with OCS 3.x. if it's accepted I can  change it

pkg/util/util.go
Madhu-1: the max length of volID will be calculated on 2 things. one is clusterID length and custom prefix length.
we need to decide what should be the max value for both

pkg/util/volid.go
Madhu-1: ComposeCSIID will be called only for the new volumes. I added code to support both V1 and V2 volumes decoding for mounting and deleting.

pkg/util/util.go
Madhu-1: sure will fix it

examples/rbd/storageclass.yaml
dillaman: It's a new storage class so there is zero need to be binary compatible w/ old naming.

dillaman: ... i.e. feature parity does not mean exactly duplicated implementation (for obvious reasons)

pkg/util/volid.go
dillaman: Understood -- but you need to support rolling upgrades. Like I said, if you deploy a new provisioner before you fully deploy your node attacher, you are broken.

pkg/util/util.go
dillaman: Cluster name is known a priori based on the storage class.

pkg/util/volid.go
Madhu-1: my assumption is that there should not be any ongoing PVC creation or Binding during the upgrade procedure.

@humblec is having more idea I think he can answer it better.

dillaman: That doesn't make sense to me ... these are going to be potentially large deployments and we should *never* require the end user to shut down k8s usage while you upgrade things. That is just not how distributed systems are built and maintained.

ShyamsundarR:
After a **very brief** look at the code I am dropping 2 comments here,

1) I think this "feature parity" is not required as I see alternate ways of determining which images back which PVs that belong to a storage class as commented [here](https://github.com/ceph/ceph-csi/issues/487#issuecomment-515799825)

2) If we want the custom name prefix nevertheless, then I would store the prefix in the image OMap as a key, as the image OMap is just named with the UUID of the image, and we extract the UUID and prefix it with the constant image prefix to get the image name, this can now come from another key in this OMap. This helps, with a few things,
  - VolID remains as before, and we do not waste more bytes to store the prefix data within the same (it has a 128 byte maximum limitation)
  - Presence of the prefix key implies, custom naming and hence all backward compatibility is relatively easier to handle

Current approach seems to be storing the custom prefix in the VolID itself, I see that as not required considering the design we have at hand of the OMaps.

**NOTE:** I may not be able to engage on this for the next 2 weeks than sporadic comments, as there are other activities that my time is stretched on at the moment, apologies for any future delays here.

ShyamsundarR:
One other alternative, is to not even prefix or change the names, but just leave the tag from the storage class as a key:value in the image OMap, and use tools to list the OMaps and filter out images with specific tags. This would ensure we do not need to refer to the container orchestrator for any data, but still not change the image names, and VolumeID versions or even have to deal with different image names during execution (for example in Delete calls that only comes with the `VolumeID`.

Madhu-1:
closing this one, as this is already fixed

Issue #466, opened 2019-07-11
Failed to unmount PVC after nodeplugin  restart


Madhu-1:
# Describe the bug #
after updating to use latest  canary  cephcsi images, failed to  unmount legacy volume created  with v1.0.0 plugin from the pod
# Environment details #

- Image/version of Ceph CSI driver: canary
- helm chart version
- Kubernetes cluster version :1.15.0
- Logs

# Steps to reproduce #
* Create PVC with v1.0.0 driver
* Mount PVC to app pod
* Delete old plugin  deployment
* Deploy plugin with canary  images (from master branch)
* Try to delete application pod.

# Actual results #
Failed to delete the application pod.
Describe what happened

# Expected behavior #
legacy volumes need to be unmounted and deleted from canary images.

Logs:
```
I0711 06:21:32.874855   10863 utils.go:105] GRPC call: /csi.v1.Node/NodeUnpublishVolume
I0711 06:21:32.874911   10863 utils.go:106] GRPC request: {"target_path":"/var/lib/kubelet/pods/4498eadf-092b-4c23-a054-4ddc114a2a88/volumes/kubernetes.io~csi/pvc-e925f233-cc14-43d7-9812-1e2fc5d01a0c/mount","volume_id":"csi-rbd-vol-8310dcae-a3a2-11e9-ad6c-3e95656eb29f"}
I0711 06:21:32.878896   10863 nsenter.go:132] Running nsenter command: nsenter [--mount=/rootfs/proc/1/ns/mnt -- /bin/realpath -e /var/lib/kubelet/pods/4498eadf-092b-4c23-a054-4ddc114a2a88/volumes/kubernetes.io~csi/pvc-e925f233-cc14-43d7-9812-1e2fc5d01a0c/mount]
I0711 06:21:32.903738   10863 nsenter_mount.go:178] nsenter findmnt args: [-o target,fstype --noheadings --first-only --target /var/lib/kubelet/pods/4498eadf-092b-4c23-a054-4ddc114a2a88/volumes/kubernetes.io~csi/pvc-e925f233-cc14-43d7-9812-1e2fc5d01a0c/mount]
[...]

PR #510, opened 2019-07-29
Proceed to unmap RBD device on failing to find staging mount point

ShyamsundarR:
Currently, we return a success if we do not find a mount at the staging
mount point. This is incorrect, as there are cases when a prior attempt
at unstage, umounts the staging path, but errors out due to RBD image
being busy (which maybe transient). Thus, on retries instead of attempting
to unmap the RBD image, the image is left mapped on the node.

The fix provided here is to attempt to find the RBD device mapped and
unmap the same if found, even when the staging mount point/path is missing.

The fix has a few parts to it,
- Move away from using staging path as the mount, when mounting the image as a filesystem, such that the staging path directory can be used to stash other metadata about the image
- Update how nbd based mapping it done to use the rbd command instead
  - Like "rbd device map --device-type [krbd|nbd]"
- Fix implementation of waitForPath to use rbd CLI to detect device and image names
  - Like, "rbd device list --format json --device-type [krbd|nbd]"
- Fix the current problem of using `rbd device list` and stashed JSON image matadata in the absence of a mount point to detect if an rbd image is mapped on the node

Future, as we get the fix from rbd CLI to add fsid to its device list output
  - stash the image fsid into a JSON file on the staging path before mapping the image
  - during unmap, when mount point does not exist (or even otherwise) consult the fsid and compare it to detect and unmap the correct image in case of image UUID conflicts

Fixes: #466 

### DIFF
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -19,8 +19,6 @@ package rbd
 import (
 	"fmt"
 	"os"
-	"os/exec"
-	"regexp"
 	"strings"
 
 	csicommon "github.com/ceph/ceph-csi/pkg/csi-common"
@@ -45,9 +43,15 @@ type NodeServer struct {
 // Implementation notes:
 // - stagingTargetPath is the directory passed in the request where the volume needs to be staged
 //   - We stage the volume into a directory, named after the VolumeID inside stagingTargetPath if
-// it is a file system
+//    it is a file system
 //   - We stage the volume into a file, named after the VolumeID inside stagingTargetPath if it is
-// a block volume
+//    a block volume
+// - Order of operation execution: (useful for defer stacking and when Unstaging to ensure steps
+//	are done in reverse, this is done in undoStagingTransaction)
+//   - Stash image metadata under staging path
+//   - Map the image (creates a device)
+//   - Create the staging file/directory under staging path
+//   - Stage the device (mount the device mapped for image)
 func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
 	if err := util.ValidateNodeStageVolumeRequest(req); err != nil {
 		return nil, err
@@ -89,8 +93,8 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		isLegacyVolume = true
 	}
 
-	stagingTargetPath := req.GetStagingTargetPath()
-	stagingTargetPath += "/" + volID
+	stagingParentPath := req.GetStagingTargetPath()
+	stagingTargetPath := stagingParentPath + "/" + req.GetVolumeId()
 
 	idLk := nodeVolumeIDLocker.Lock(volID)
 	defer nodeVolumeIDLocker.Unlock(idLk, volID)
@@ -113,22 +117,29 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	}
 	volOptions.RbdImageName = volName
 
+	isMounted := false
+	isStagePathCreated := false
+	devicePath := ""
+
+	// Stash image details prior to mapping the image (useful during Unstage as it has no
+	// voloptions passed to the RPC as per the CSI spec)
+	err = stashRBDImageMetadata(volOptions, stagingParentPath)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	defer func() {
+		if err != nil {
+			ns.undoStagingTransaction(stagingParentPath, devicePath, volID, isStagePathCreated, isMounted)
+		}
+	}()
+
 	// Mapping RBD image
-	devicePath, err := attachRBDImage(volOptions, cr)
+	devicePath, err = attachRBDImage(volOptions, cr)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	klog.V(4).Infof("rbd image: %s/%s was successfully mapped at %s\n", req.GetVolumeId(), volOptions.Pool, devicePath)
 
-	isMounted := false
-	isStagePathCreated := false
-	// if mounting to stagingpath fails unmap the rbd device. this wont leave any
-	// stale rbd device if unstage is not called
-	defer func() {
-		if err != nil {
-			ns.cleanupStagingPath(stagingTargetPath, devicePath, volID, isStagePathCreated, isMounted)
-		}
-	}()
 	err = ns.createStageMountPoint(stagingTargetPath, isBlock)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
@@ -152,24 +163,40 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	return &csi.NodeStageVolumeResponse{}, nil
 }
 
-func (ns *NodeServer) cleanupStagingPath(stagingTargetPath, devicePath, volID string, isStagePathCreated, isMounted bool) {
+func (ns *NodeServer) undoStagingTransaction(stagingParentPath, devicePath, volID string, isStagePathCreated, isMounted bool) {
 	var err error
+
+	stagingTargetPath := stagingParentPath + "/" + volID
 	if isMounted {
 		err = ns.mounter.Unmount(stagingTargetPath)
 		if err != nil {
 			klog.Errorf("failed to unmount stagingtargetPath: %s with error: %v", stagingTargetPath, err)
+			return
 		}
 	}
-	// remove the block file created on staging path
+
+	// remove the file/directory created on staging path
 	if isStagePathCreated {
 		err = os.Remove(stagingTargetPath)
 		if err != nil {
 			klog.Errorf("failed to remove stagingtargetPath: %s with error: %v", stagingTargetPath, err)
+			// continue on failure to unmap the image, as leaving stale images causes more issues than a stale file/directory
 		}
 	}
+
 	// Unmapping rbd device
-	if err = detachRBDDevice(devicePath); err != nil {
-		klog.Errorf("failed to unmap rbd device: %s for volume %s with error: %v", devicePath, volID, err)
+	if devicePath != "" {
+		err = detachRBDDevice(devicePath)
+		if err != nil {
+			klog.Errorf("failed to unmap rbd device: %s for volume %s with error: %v", devicePath, volID, err)
+			// continue on failure to delete the stash file, as kubernetes will fail to delete the staging path otherwise
+		}
+	}
+
+	// Cleanup the stashed image metadata
+	if err = cleanupRBDImageMetadataStash(stagingParentPath); err != nil {
+		klog.Errorf("failed to cleanup image metadata stash (%v)", err)
+		return
 	}
 }
 
@@ -190,8 +217,10 @@ func (ns *NodeServer) createStageMountPoint(mountPath string, isBlock bool) erro
 
 	err := os.Mkdir(mountPath, 0750)
 	if err != nil {
-		klog.Errorf("failed to create mountPath:%s with error: %v", mountPath, err)
-		return status.Error(codes.Internal, err.Error())
+		if !os.IsExist(err) {
+			klog.Errorf("failed to create mountPath:%s with error: %v", mountPath, err)
+			return status.Error(codes.Internal, err.Error())
+		}
 	}
 
 	return nil
@@ -372,7 +401,7 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	klog.Infof("rbd: successfully unbinded volume %s from %s", req.GetVolumeId(), targetPath)
+	klog.Infof("rbd: successfully unbound volume %s from %s", req.GetVolumeId(), targetPath)
 
 	return &csi.NodeUnpublishVolumeResponse{}, nil
 }
@@ -384,111 +413,72 @@ func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 		return nil, err
 	}
 
-	stagingTargetPath := req.GetStagingTargetPath()
-	stagingTargetPath += "/" + req.GetVolumeId()
+	stagingParentPath := req.GetStagingTargetPath()
+	stagingTargetPath := stagingParentPath + "/" + req.GetVolumeId()
 
 	notMnt, err := mount.IsNotMountPoint(ns.mounter, stagingTargetPath)
 	if err != nil {
-		if os.IsNotExist(err) {
-			// staging targetPath has already been deleted
-			klog.V(4).Infof("stagingTargetPath: %s has already been deleted", stagingTargetPath)
-			return &csi.NodeUnstageVolumeResponse{}, nil
+		if !os.IsNotExist(err) {
+			return nil, status.Error(codes.NotFound, err.Error())
 		}
-		return nil, status.Error(codes.NotFound, err.Error())
+		// Continue on ENOENT errors as we may still have the image mapped
+		notMnt = true
 	}
-	if notMnt {
-		// TODO: IsNotExist error should have been caught above
-		if err = os.Remove(stagingTargetPath); err != nil {
+	if !notMnt {
+		// Unmounting the image
+		err = ns.mounter.Unmount(stagingTargetPath)
+		if err != nil {
+			klog.V(3).Infof("failed to unmount targetPath: %s with error: %v", stagingTargetPath, err)
 			return nil, status.Error(codes.Internal, err.Error())
 		}
-		return &csi.NodeUnstageVolumeResponse{}, nil
-	}
-
-	// Unmount the volume
-	devicePath, cnt, err := mount.GetDeviceNameFromMount(ns.mounter, stagingTargetPath)
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
-	}
-
-	if err = ns.unmount(stagingTargetPath, devicePath, cnt); err != nil {
-		return nil, err
 	}
 
 	if err = os.Remove(stagingTargetPath); err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
-	}
-
-	klog.Infof("rbd: successfully unmounted volume %s from %s", req.GetVolumeId(), stagingTargetPath)
-
-	return &csi.NodeUnstageVolumeResponse{}, nil
-}
-
-func (ns *NodeServer) unmount(targetPath, devicePath string, cnt int) error {
-	var err error
-	// Bind mounted device needs to be resolved by using resolveBindMountedBlockDevice
-	if devicePath == "devtmpfs" {
-		devicePath, err = resolveBindMountedBlockDevice(targetPath)
-		if err != nil {
-			return status.Error(codes.Internal, err.Error())
+		// Any error is critical as Staging path is expected to be empty by Kubernetes, it otherwise
+		// keeps invoking Unstage. Hence any errors removing files within this path is a critical
+		// error
+		if !os.IsNotExist(err) {
+			klog.Errorf("failed to remove staging target path (%s): (%v)", stagingTargetPath, err)
+			return nil, status.Error(codes.Internal, err.Error())
 		}
-		klog.V(4).Infof("NodeUnpublishVolume: devicePath: %s, (original)cnt: %d\n", devicePath, cnt)
-		// cnt for GetDeviceNameFromMount is broken for bind mouted device,
-		// it counts total number of mounted "devtmpfs", instead of counting this device.
-		// So, forcibly setting cnt to 1 here.
-		// TODO : fix this properly
-		cnt = 1
 	}
 
-	klog.V(4).Infof("NodeUnpublishVolume: targetPath: %s, devicePath: %s\n", targetPath, devicePath)
-
-	// Unmounting the image
-	err = ns.mounter.Unmount(targetPath)
+	imgInfo, err := lookupRBDImageMetadataStash(stagingParentPath)
 	if err != nil {
-		klog.V(3).Infof("failed to unmount targetPath: %s with error: %v", targetPath, err)
-		return status.Error(codes.Internal, err.Error())
-	}
+		klog.V(2).Infof("failed to find image metadata: %v", err)
+		// It is an error if it was mounted, as we should have found the image metadata file with
+		// no errors
+		if !notMnt {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
 
-	cnt--
-	if cnt != 0 {
-		// TODO should this be fixed not to success, so that driver can retry unmounting?
-		return nil
+		// If not mounted, and error is anything other than metadata file missing, it is an error
+		if _, ok := err.(ErrMissingStash); !ok {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		// It was not mounted and image metadata is also missing, we are done as the last step in
+		// in the staging transaction is complete
+		return &csi.NodeUnstageVolumeResponse{}, nil
 	}
 
 	// Unmapping rbd device
-	if err = detachRBDDevice(devicePath); err != nil {
-		klog.V(3).Infof("failed to unmap rbd device: %s with error: %v", devicePath, err)
-		return status.Error(codes.Internal, err.Error())
+	imageSpec := imgInfo.Pool + "/" + imgInfo.ImageName
+	if err = detachRBDImageOrDeviceSpec(imageSpec, true, imgInfo.NbdAccess); err != nil {
+		klog.Errorf("error unmapping volume (%s) from staging path (%s): (%v)",
+			req.GetVolumeId(), stagingTargetPath, err)
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	return nil
-}
-func resolveBindMountedBlockDevice(mountPath string) (string, error) {
-	// #nosec
-	cmd := exec.Command("findmnt", "-n", "-o", "SOURCE", "--first-only", "--target", mountPath)
-	out, err := cmd.CombinedOutput()
-	if err != nil {
-		klog.V(2).Infof("Failed findmnt command for path %s: %s %v", mountPath, out, err)
-		return "", err
-	}
-	return parseFindMntResolveSource(string(out))
-}
+	klog.Infof("successfully unmounted volume (%s) from staging path (%s)",
+		req.GetVolumeId(), stagingTargetPath)
 
-// parse output of "findmnt -o SOURCE --first-only --target" and return just the SOURCE
-func parseFindMntResolveSource(out string) (string, error) {
-	// cut trailing newline
-	out = strings.TrimSuffix(out, "\n")
-	// Check if out is a mounted device
-	reMnt := regexp.MustCompile("^(/[^/]+(?:/[^/]*)*)$")
-	if match := reMnt.FindStringSubmatch(out); match != nil {
-		return match[1], nil
+	if err = cleanupRBDImageMetadataStash(stagingParentPath); err != nil {
+		klog.Errorf("failed to cleanup image metadata stash (%v)", err)
+		return nil, status.Error(codes.Internal, err.Error())
 	}
-	// Check if out is a block device
-	// nolint
-	reBlk := regexp.MustCompile("^devtmpfs\\[(/[^/]+(?:/[^/]*)*)\\]$")
-	if match := reBlk.FindStringSubmatch(out); match != nil {
-		return fmt.Sprintf("/dev%s", match[1]), nil
-	}
-	return "", fmt.Errorf("parseFindMntResolveSource: %s doesn't match to any expected findMnt output", out)
+
+	return &csi.NodeUnstageVolumeResponse{}, nil
 }
 
 // NodeGetCapabilities returns the supported capabilities of the node server
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -42,12 +42,17 @@ type NodeServer struct {
 }
 
 // NodeStageVolume mounts the volume to a staging path on the node.
+// Implementation notes:
+// - stagingTargetPath is the directory passed in the request where the volume needs to be staged
+//   - We stage the volume into a directory, named after the VolumeID inside stagingTargetPath if
+// it is a file system
+//   - We stage the volume into a file, named after the VolumeID inside stagingTargetPath if it is
+// a block volume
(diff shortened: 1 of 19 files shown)


ShyamsundarR:
@dillaman Current code uses the staging path where the device is expected to be mounted, to determine the device that is mounted in that path. Hence uses this information to unmap the device.

The code in this fix (which is WIP, and put up here to get the approach clarified), can only get this information from crawling through the devices already mounted. We lose some information that we need to hence extrapolate as follows,

1. Thoughts on using Pool-id+image-name to find a device that is mapped to a node?
  - My theory would be, the 'pool-id' may conflict if there is another Ceph cluster in play, but so can the `pool-name` that the current code uses. So either combination is as safe as the other.
  - Is there a way to make it safer given we do not have any credentials to use and validate extra information from the Ceph cluster itself?
  - Reference function that does this: https://github.com/ceph/ceph-csi/blob/dfbdec4b6ab2667df2512140270b110115d954b2/pkg/rbd/rbd_attach.go#L55-L95

- Thoughts on using just the image-name (IOW, its UUID) in the case of rbd-nbd?
  - Current code uses the `imagespec` and hence considers the poolname in addition to the image name
  - Code reference: https://github.com/ceph/ceph-csi/blob/dfbdec4b6ab2667df2512140270b110115d954b2/pkg/rbd/rbd_attach.go#L151-L194

@Madhu-1 When you [attempted the experiment](https://github.com/ceph/ceph-csi/issues/466#issuecomment-510400005) you noted that the rbd device was being reported as busy, did you attempt to get the lsof and other output as requested and make progress? I hit this situation once and did not find any useful output from lsof hence asking.

@humblec this is where we are w.r.t Unstage issue for busy rbd images at the moment. Any help is appreciated.

dillaman:
@ShyamsundarR Just my opinion, but `ceph-csi` shouldn't be poking around like it currently does. Instead, use `rbd device list --format json --device-type [krbd|nbd]`:

> $ sudo ./bin/rbd device list --device-type krbd --format json
> [{"id":"0","pool":"rbd","namespace":"","name":"foo","snap":"-","device":"/dev/rbd0"}]
> $ sudo ./bin/rbd device list --device-type nbd --format json
> [{"id":21979,"pool":"rbd","namespace":"","image":"foo","snap":"","device":"/dev/nbd0"}]

We could also include the cluster fsid and/or pool id if that helps anything.

dillaman:
BTW -- that suggestion requires updating `ceph-csi` to use `rbd device map --device-type [krbd|nbd]` instead of directly invoking `rbd-nbd` as it currently does.

ShyamsundarR:
@dillaman Interesting and agree we should not be poking around device lists like we do now.

It does look much better, even if we still may have 2 devices sharing a name. I think parsing the output as above would be a way forward, but still leaves the limitations as before?

dillaman:
> still leaves the limitations as before

Sorry, I might be missing it, but what are the limitations that need to be addressed?

ShyamsundarR:
> > still leaves the limitations as before
> 
> @ShyamsundarR Sorry, I might be missing it, but what are the limitations that need to be addressed?

:) no worries... I don't think I did a great job explaining it initially as well. So here goes,

Consider a system where we have an rbd image mapped. Assuming we just know the `imagename` and `poolid` of the image, how to determine which devicename is associated with the image? So that we can run the command `rbd unmap <devicepath>` (this is now answered, and was not really part of the initial question).

The problem/limitation is that we could have 2 images with similar properties as above (`poolid` and `imagename`), but from different Ceph clusters mapped on the node. Hence **without** the cluster `fsid` is this even possible to distinguish them?

One other manner of solving the problem would be to fetch the image info (inclusive of fsid) before mapping it, and storing it on the node (under the path where it was to be mapped) and consulted later when unmapping the same. (this path is the staging path in CSI and not the device path, so still feasible)

FWIW, the image name has an UUID in our case, and maybe it is enough for now.

dillaman:
> The problem/limitation is that we could have 2 images with similar properties as above (`poolid` and `imagename`), but from different Ceph clusters mapped on the node. Hence **without** the cluster `fsid` is this even possible to distinguish them?

That's why I offered to add the `fsid` (cluster id) to the output. krbd already stores this information and it could be added to nbd as well.

ShyamsundarR:
> > The problem/limitation is that we could have 2 images with similar properties as above (`poolid` and `imagename`), but from different Ceph clusters mapped on the node. Hence **without** the cluster `fsid` is this even possible to distinguish them?
> 
> That's why I offered to add the `fsid` (cluster id) to the output. krbd already stores this information and it could be added to nbd as well.

That is surely needed, but additionally CSI also has to "stash" the fsid somewhere before mapping the device to solve this.

So I think overall, way forward is to do both to be sure we are unmapping the right device. I will get back after a little thought on the requirement for `fsid` I think this maybe useful for the `NodeGetVolumeStats` call as well.

dillaman:
> One other manner of solving the problem would be to fetch the image info (inclusive of fsid) before mapping it, and storing it on the node (under the path where it was to be mapped) and consulted later when unmapping the same

For this case, where would you stash this information? Just in a tmpfs in the node container?  I like this approach since the publish command could take care of storing anything required for the unpublish.

dillaman:
> That is surely needed, but additionally CSI also has to "stash" the fsid somewhere before mapping the device to solve this.

Sorry, I thought that was already stashed in the volume id.

ShyamsundarR:
> > That is surely needed, but additionally CSI also has to "stash" the fsid somewhere before mapping the device to solve this.
> 
> Sorry, I thought that was already stashed in the volume id.

Nope, it is a choice for the user of CSI and not mandatory, although we say "it is good to use the fsid" we do not enforce it :/

dillaman:
> this was to support user friendly cluster IDs

The user would never see these, correct? Just odd that we force them to populate it instead of just retrieving it upon startup.

ShyamsundarR:
> > this was to support user friendly cluster IDs
> 
> The user would never see these, correct? Just odd that we force them to populate it instead of just retrieving it upon startup.

So the `StorageClass` contains the `clusterID` that a user has to mention, and further the same has to be used within the CSI configuration. This is where the container orchestrator user comes across the ID and hence we did not enforce it to be the Ceph cluster `fsid`.

ShyamsundarR:
Updated design/approach (based on the discussion above):

- Fix (reimplement) waitForPath to use `rbd device list ...`
- Fix mounting the device into subdir/file inside passed in staging path in NodeStageVolume
  - Should also handle the hack within Stage for Block versus FS mode of accessing RBD device
- Fetch and store cluster `fsid`, into a file under staging path, before mapping the image into the staging path in NodeStageVolume
  - maybe even store the pool name for building image-spec for use with other commands
- Fix current problem, looking for rbd device in reimplemented waitForPath call, and unmapping the same as needed
- Further fix possible device detection using cluster `fsid` and stashed cluster `fsid` in the staging path, once the rbd command starts adding that to the output as well
  - Depends on rbd implementation of the same
  - @dillaman do I need to raise a ticket somewhere for this?

@Madhu-1 @poornimag @humblec if there are any questions or thoughts please share.

dillaman:
> * Fix mounting the device into subdir/file inside passed in staging path in NodeStageVolume
> * Fetch and store cluster `fsid`, into a file under staging path, before mapping the image into the staging path in NodeStageVolume

Is this permitted or would this be a hack? Perhaps just store it as part of a "mount -o x-ceph-fsid=XYZ" optional on the (bind) mount?

> * Further fix possible device detection using cluster `fsid` and stashed cluster `fsid` in the staging path, once the rbd command starts adding that to the output as well
>   * @dillaman do I need to raise a ticket somewhere for this?

If you still need that data from "rbd device list" and cannot just stash the vol id or similar as a (bind) mount optional, I can create a ticket for it. Let me know.

ShyamsundarR:
> > * Fix mounting the device into subdir/file inside passed in staging path in NodeStageVolume
> > * Fetch and store cluster `fsid`, into a file under staging path, before mapping the image into the staging path in NodeStageVolume
> 
> Is this permitted or would this be a hack? Perhaps just store it as part of a "mount -o x-ceph-fsid=XYZ" optional on the (bind) mount?

This is not necessarily a hack, we are given a staging directory to use, and we already create a file within it for mounting a block only volume.

BUT, having said that, I would like lesser work to do, and want to understand how we would extract the mount option if passed as above? Also, if the mount disappears, how will we find it to ensure we are unmapping the right rbd device (I think the latter is still a concern if we use a mount option, if it was an rbd map option that is a 1:1 correspondence, no?)

dillaman:
> Also, if the mount disappears, how will we find it to ensure we are unmapping the right rbd device

D'oh, yes of course. You could store it as an xattr on the staging path directory prior to mapping, though, right?

ShyamsundarR:
> > Also, if the mount disappears, how will we find it to ensure we are unmapping the right rbd device
> 
> D'oh, yes of course. You could store it as an xattr on the staging path directory prior to mapping, though, right?

yes, but assumes xattr support in the underlying FS/system (maybe a given at this point, but i'd rather trust a local .json file, if it is all the same :) )

Madhu-1:
@ShyamsundarR please go ahead with .json file creation approach

ShyamsundarR:
Added code to to fix the following till now (2 commits),
- Move away from using staging path as the mount, when mounting the image as a filesystem, such that the staging path directory can be used to stash other metadata about the image
- Update how nbd based mapping it done to use the rbd command instead
    - Like "rbd device map --device-type [krbd|nbd]"
- Fix implementation of waitForPath to use rbd CLI to detect device and image names
    - Like, "rbd device list --format json --device-type [krbd|nbd]"

ShyamsundarR:
@dillaman I lost the comment thread on the `/proc` availability as I force pushed a commit (thanks github! :/ )

So here is a summary:
- detecting mapped nbd devices needs access to `/proc/<pid>/cmdline`
- this is available for processes started within the pod/container namespace

So we are good to go.

While testing the changes I noticed a [discrepancy](https://github.com/ceph/ceph-csi/pull/510/files#diff-7939c144f71f04cae39a4901fb54e4d3R57-R59) in `rbd device list --format=json ..` output when listing nbd versus krbd devices. I have worked around it, but tagging you to check and see if this is a bug or the expectation.

ShyamsundarR:
@Madhu-1 @humblec PTAL

@dillaman one further question,
- Why call waitForImage post mapping the image? We seem to be getting the devicePath as the output from `rbd map ...` and so should we really wait and check the device is available? Code reference here: https://github.com/ceph/ceph-csi/pull/510/files#diff-7939c144f71f04cae39a4901fb54e4d3R212-R226

deploy/cephfs/helm/templates/nodeplugin-daemonset.yaml
```diff
@@ -73,8 +73,6 @@ spec:
             - "--metadatastorage=k8s_configmap"
             - "--mountcachedir=/mount-cache-dir"
           env:
-            - name: HOST_ROOTFS
```
Madhu-1: why this is removed? please add a comment it will help in review

pkg/rbd/nodeserver.go
```diff
@@ -148,24 +164,39 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
 		}
 	}
+
 	// remove the block file created on staging path
-	if isBlock && isStagePathCreated {
+	if isStagePathCreated {
 		err = os.Remove(stagingTargetPath)
 		if err != nil {
 			klog.Errorf("failed to remove stagingtargetPath: %s with error: %v", stagingTargetPath, err)
+			return
```
Madhu-1: this failure can leave stable rbd  devices, as there won't be any node unstage call if node staging fails

pkg/rbd/nodeserver.go
```diff
@@ -42,12 +40,23 @@ type NodeServer struct {
 }
 
 // NodeStageVolume mounts the volume to a staging path on the node.
+// Implementation notes:
+// - stagingTargetPath is the directory passed in the request where the volume needs to be staged
```
Madhu-1: why do we need to create directory and files based on the block/file?

can we just create a file inside the stagingPath and use it?

pkg/rbd/nodeserver.go
```diff
@@ -378,126 +411,73 @@ func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
...
+		return &csi.NodeUnstageVolumeResponse{}, nil
 	}
 
-	cnt--
-	if cnt != 0 {
-		// TODO should this be fixed not to success, so that driver can retry unmounting?
-		return nil
-	}
+	// Find device mapped to image
+	devicePath, found := findDeviceMappingImage(imgInfo.Pool, imgInfo.ImageName, imgInfo.NbdAccess)
```
Madhu-1: as we know the devicePath after rbd map we can add the devicePath to the json file?

deploy/cephfs/helm/templates/nodeplugin-daemonset.yaml
ShyamsundarR: Code prior to this was using this env variable to set hostRootFs here https://github.com/ceph/ceph-csi/blob/02bcb5f16ad062fd2b5814d866daad44c7ace738/pkg/rbd/rbd_attach.go#L34-L50 (in rbd)

The above variable is used to inspect `<hostRootFs>/proc` which is no longer needed as that function is removed.

CephFS is not using this variable at all, and was present superfluously, hence removed.

pkg/rbd/nodeserver.go
ShyamsundarR: Hmmm... ok we can remove this return. I am assuming you are fine with the other returns in this function, the reasoning for the same was, an umount failure will result in an unmap failure as the device would be busy anyway.

pkg/rbd/nodeserver.go
ShyamsundarR: Target of a flesystem mount needs to be a directory, and we need the target as a file when bind mounting a devoce for the block use-case. Hence the difference.

pkg/rbd/nodeserver.go
ShyamsundarR: True, it can be done, BUT I did not want to break the transaction order, and hence add or assume we will find this later in the image metadata stash. IOW, if partway through the transaction errors or faults occur, we will not have that information to stash, and hence will need to fall back to other means.

pkg/rbd/nodeserver.go
Madhu-1: cant we just create a `.config.json` inside a stagingTargetPath directory and use it?  why create directory and file?

pkg/rbd/nodeserver.go
Madhu-1: what transaction order?

what are the transaction errors can happen here?

pkg/rbd/nodeserver.go
ShyamsundarR: The comment reads as below:

// - stagingTargetPath is the directory passed in the request where the volume needs to be staged
//   - We stage the volume into a directory, named after the VolumeID inside stagingTargetPath if
//    it is a file system
//   - We stage the volume into a file, named after the VolumeID inside stagingTargetPath if it is
//    a block volume

So, we mount the device on a directory inside the `stagingTargetPath`, the directory is named based on the `VolumeID` OR we bind mount the device to a file named similarly.

We create and maintain the `config.json` inside the 'stagingTargetPath` which is outside either of these 2 mounts and hence outside the mount point and accessible even post the umount etc.

So to answer your question, `config.json` is created just within the passed in `stagingTargetPath`, the file/directory is created to stage the image as required.

I do not see how not creating a sub-entity where we stage can be avoided if that is your direction of thought.

pkg/rbd/nodeserver.go
ShyamsundarR: Read this comment: https://github.com/ceph/ceph-csi/pull/510/files#diff-a2c76a9049f52b0f018e9dac09c35f0cR49-R54

IOW, we create the `config.json` first before mapping the device, so we are sure to find the file to help with unmap and other operations as required. Hence, updating it (with the device name) post the fact, brings in an assumption in the code that this may or may not have happened, which I want to avoid.

pkg/rbd/nodeserver.go
Madhu-1: if `stagingTargetPath` looks like `/var/lib/kubelet/kubernetes.io/pv/pvc-xxxx-xxxx-xxx/globalmount` can we just create `/var/lib/kubelet/kubernetes.io/pv/pvc-xxxx-xxxx-xxx/.config.json` and use it?

this will be available for unmapping also

ShyamsundarR: I would not attempt to use a path outside the scope of what is passed to the plugin, IOW something akin to `cd ..`  and then create files in there. As this assumes we have rights to manage files and directories outside the passed in `stagingTargetPath` which I believe is breaking the abstraction provided.

So even though we know what the path is, I would surely not want to do this.

pkg/rbd/nodeserver.go
ShyamsundarR: Another consideration here would be to use the `imagespec` as we know the `pool` and `imagename` from the stashed data, instead of determining the devicePath at all, as pointed out by @Madhu-1 [here](https://github.com/ceph/ceph-csi/issues/520#issuecomment-518109000)

I am looking at this (to ensure we get errors that help when the image is not mapped, to detect and ignore the same), to refrain from attempting to get the `devicePath` at all in the first place

ShyamsundarR: > I am looking at this (to ensure we get errors that help when the image is not mapped, to detect and ignore the same), to refrain from attempting to get the `devicePath` at all in the first place

The CLI output when the images are not mapped like as below,
```
# rbd unmap --device-type krbd replicapool/csi-vol-7101dd25-b7c3-11e9-8b61-d252735dc0f9
rbd: replicapool/csi-vol-7101dd25-b7c3-11e9-8b61-d252735dc0f9: not a mapped image or snapshot
rbd: unmap failed: (22) Invalid argument

# rbd unmap --device-type nbd replicapool/csi-vol-7101dd25-b7c3-11e9-8b61-d252735dc0f9
rbd-nbd: replicapool/csi-vol-7101dd25-b7c3-11e9-8b61-d252735dc0f9 is not mapped
rbd: rbd-nbd failed with error: /usr/bin/rbd-nbd: exit status: 1
```

We could use `imagespec` and look for these errors as benign and proceed, or leave the code as is for now, fixing the unmap problem first. @Madhu-1 thoughts?

ShyamsundarR:
CI failed in kube 1.13.7 test with the following logs,
```
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
apt-get.diagnostics

apt-get install failed

The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install realpath ruby" failed and exited with 100 during .
```

Which is unrelated to the patch. The tests did not get started at all, just noting it here for reviewers.

pkg/rbd/nodeserver.go
ShyamsundarR: fixed.

Madhu-1:
I see 3 approaches below to fix this  issue

1) create `.rbd.json` file and use it for unmapping

we if go it this approach I would suggest adding image name and pool name to the config.json
use this info to unmap the image

2) as we have the volume ID and pool id in unstage request we can use volume name and pool name for unmapping the image
this requires us to  find out the pool_name from pool_id (this  will increase the computation time also as we need to get the  devices list and compare pool id  and image name)

even  if the method  1 fails,  we should not undo the node stage request as use method 2 as back up for unmapping during unstage request

3)create a directory (dir2) with the image name and pool name (even with pool name is also fine as we will have volumeId in node unstage request) and use it for mounting. ( with this naming convention we can avoid `.rbd.json` file creation also.  we can extract the image name and pool name during unstage.

I would like to go with option 3

@humblec @ShyamsundarR  any thoughts?

humblec:
> I see 3 approaches below to fix this issue
> 
>     1. create `.rbd.json` file and use it for unmapping
> 
> 
> we if go it this approach I would suggest adding image name and pool name to the config.json
> use this info to unmap the image
> 
>     1. as we have the volume ID and pool id in unstage request we can use volume name and pool name for unmapping the image
>        this requires us to  find out the pool_name from pool_id (this  will increase the computation time also as we need to get the  devices list and compare pool id  and image name)
> 
> 
> even if the method 1 fails, we should not undo the node stage request as use method 2 as back up for unmapping during unstage request
> 
> 3)create a directory (dir2) with the image name and pool name (even with pool name is also fine as we will have volumeId in node unstage request) and use it for mounting. ( with this naming convention we can avoid `.rbd.json` file creation also. we can extract the image name and pool name during unstage.
> 
> I would like to go with option 3
> 
> @humblec @ShyamsundarR any thoughts?

@madhu-1 I would also prefer to go ahead with option 3. The main reason being, it avoid lots of complexities in the code and also avoid computational overhead  with other 2 options. The third option also gives the flexibility of smooth identiifcation of the volume where it belongs.

Madhu-1:
> @dillaman Current code uses the staging path where the device is expected to be mounted, to determine the device that is mounted in that path. Hence uses this information to unmap the device.
> 
> The code in this fix (which is WIP, and put up here to get the approach clarified), can only get this information from crawling through the devices already mounted. We lose some information that we need to hence extrapolate as follows,
> 
> 1. Thoughts on using Pool-id+image-name to find a device that is mapped to a node?
> 
> * My theory would be, the 'pool-id' may conflict if there is another Ceph cluster in play, but so can the `pool-name` that the current code uses. So either combination is as safe as the other.
> * Is there a way to make it safer given we do not have any credentials to use and validate extra information from the Ceph cluster itself?
> * Reference function that does this: https://github.com/ceph/ceph-csi/blob/dfbdec4b6ab2667df2512140270b110115d954b2/pkg/rbd/rbd_attach.go#L55-L95
> * Thoughts on using just the image-name (IOW, its UUID) in the case of rbd-nbd?
>   
>   * Current code uses the `imagespec` and hence considers the poolname in addition to the image name
>   * Code reference: https://github.com/ceph/ceph-csi/blob/dfbdec4b6ab2667df2512140270b110115d954b2/pkg/rbd/rbd_attach.go#L151-L194
> 
> @Madhu-1 When you [attempted the experiment](https://github.com/ceph/ceph-csi/issues/466#issuecomment-510400005) you noted that the rbd device was being reported as busy, did you attempt to get the lsof and other output as requested and make progress? I hit this situation once and did not find any useful output from lsof hence asking.
> 
nope, I didn't look into deeper for resource busy issue. does this PR fix resource busy issue?

> @humblec this is where we are w.r.t Unstage issue for busy rbd images at the moment. Any help is appreciated.

ShyamsundarR:
> I see 3 approaches below to fix this issue

What issue are we talking about here? Where/how to store image details for Unstage to pick image/pool from?

> 
>     1. create `.rbd.json` file and use it for unmapping
> 
> 
> we if go it this approach I would suggest adding image name and pool name to the config.json
> use this info to unmap the image

The current code does just this. I.e adds the .json and stores the pool name and image name into the same for use later, see https://github.com/ceph/ceph-csi/pull/510/files#diff-8e27516879efe971bf4755fadf585e06R712-R717

> 
>     1. as we have the volume ID and pool id in unstage request we can use volume name and pool name for unmapping the image
>        this requires us to  find out the pool_name from pool_id (this  will increase the computation time also as we need to get the  devices list and compare pool id  and image name)

More than computation time, we do not have credentials to use during Unstage to fetch the pool name for a given pool ID, hence this is not feasible.

> 
> 
> even if the method 1 fails, we should not undo the node stage request as use method 2 as back up for unmapping during unstage reque st
> 
> 3)create a directory (dir2) with the image name and pool name (even with pool name is also fine as we will have volumeId in node unstage request) and use it for mounting. ( with this naming convention we can avoid `.rbd.json` file creation also. we can extract the image name and pool name during unstage.

Creating a directory and decoding the same to fetch the image and pool names are tedious and not extensible for other needs in the future. It is much more elegant to store this in an encoded file that allows for easy extraction of the same.

I am not sure why we are attempting to avoid the .json file here. What am I missing?

> 
> I would like to go with option 3
> 
> @humblec @ShyamsundarR any thoughts?

Madhu-1:
> > I see 3 approaches below to fix this issue
> 
> What issue are we talking about here? Where/how to store image details for Unstage to pick image/pool from?

am talking about the approaches to solve this issue
> > ```
> > 1. create `.rbd.json` file and use it for unmapping
> > ```
> > 
> > 
> > we if go it this approach I would suggest adding image name and pool name to the config.json
> > use this info to unmap the image
> 
> The current code does just this. I.e adds the .json and stores the pool name and image name into the same for use later, see https://github.com/ceph/ceph-csi/pull/510/files#diff-8e27516879efe971bf4755fadf585e06R712-R717
> 
> > ```
> > 1. as we have the volume ID and pool id in unstage request we can use volume name and pool name for unmapping the image
> >    this requires us to  find out the pool_name from pool_id (this  will increase the computation time also as we need to get the  devices list and compare pool id  and image name)
> > ```
> 
> More than computation time, we do not have credentials to use during Unstage to fetch the pool name for a given pool ID, hence this is not feasible.

why it's not feasible you can get the pool name from the device in `/sys/device/X/pool`
> 
> > even if the method 1 fails, we should not undo the node stage request as use method 2 as back up for unmapping during unstage reque st
> > 3)create a directory (dir2) with the image name and pool name (even with pool name is also fine as we will have volumeId in node unstage request) and use it for mounting. ( with this naming convention we can avoid `.rbd.json` file creation also. we can extract the image name and pool name during unstage.
> 
> Creating a directory and decoding the same to fetch the image and pool names are tedious and not extensible for other needs in the future. It is much more elegant to store this in an encoded file that allows for easy extraction of the same.
> 
just trying to understand what are the other needs in future? 
> I am not sure why we are attempting to avoid the .json file here. What am I missing?
> 

if the issue can be solved without the creation of the json file, why to create json file?

> > I would like to go with option 3
> > @humblec @ShyamsundarR any thoughts?

ShyamsundarR:
> > Creating a directory and decoding the same to fetch the image and pool names are tedious and not extensible for other needs in the future. It is much more elegant to store this in an encoded file that allows for easy extraction of the same.
> 
> just trying to understand what are the other needs in future?

Say, stashing the fsid of the image, so that we can avoid image/pool conflicts. The point is, it is extensible, whereas encoding this information in directory names is not as easily extensible.

> 
> > I am not sure why we are attempting to avoid the .json file here. What am I missing?
> 
> if the issue can be solved without the creation of the json file, why to create json file?

The alternate solution seems to be to encode the pool and image names etc. in the directory name, which is encoding it somewhere anyway. In the case of the JSON file, it stores the same information in a file. So I am not sure either approach is avoiding anything, it is using different means to store the information.

If we are storing information anyway, I prefer (and strongly support) using easier forms of parsing and extensible approaches, like encoded files, than storing these encoded in directory names.

ShyamsundarR:
> > > 1. as we have the volume ID and pool id in unstage request we can use volume name and pool name for unmapping the image
> > >    this requires us to  find out the pool_name from pool_id (this  will increase the computation time also as we need to get the  devices list and compare pool id  and image name)
> > > ```
> > 
> > 
> > More than computation time, we do not have credentials to use during Unstage to fetch the pool name for a given pool ID, hence this is not feasible.
> 
> why it's not feasible you can get the pool name from the device in `/sys/device/X/pool`

If I am not wrong we went through this in chat when we found the pool file under devices, and initial patch was to change the way we look at device directories to find this pool mapping.

The change in approach as discussed in this PR above, was when we found cleaner ways to do this via `rbd` cli itself, and also to handle the problem of nbd mappings, instead of poking around the device directories.

ShyamsundarR:
> @dillaman one further question,
> 
>     * Why call waitForImage post mapping the image? We seem to be getting the devicePath as the output from `rbd map ...` and so should we really wait and check the device is available? Code reference here: https://github.com/ceph/ceph-csi/pull/510/files#diff-7939c144f71f04cae39a4901fb54e4d3R212-R226

(dropping a note here) This is fixed as part of the PR https://github.com/ceph/ceph-csi/pull/525

pkg/rbd/nodeserver.go
Madhu-1: if you want devicePath during unmapping. why can't we just store the devicePath in metadata and use it for unmapping? 
the above error can be discarded as the image will not be mapped at all

ShyamsundarR: As answered [here](https://github.com/ceph/ceph-csi/pull/510#discussion_r310554247) and [here](https://github.com/ceph/ceph-csi/pull/510#discussion_r310560683), I do not want to touch the json file again post creating it.

Further we can also avoid determining which device file it is as noted [here](https://github.com/ceph/ceph-csi/pull/510#discussion_r310624209), if we use `imagespec` from the saved information. Which is an open option for now.

I want to follow a strict order, such that we do not update the file post creating it. In case of plugin restarts in between a Stage request, recovery cannot rely on device name in the file, as that may not have been updated post mapping the image. I hence do not want to add add the device path to the file, but would rather determine it as in the code before or use `imagespec`.

ShyamsundarR: Discussed this off-line with @Madhu-1 and decided to optimize out the call to get device lists and detect the `devicePath` to unmap, and instead use the `imagespec` version of `rbd unmap` command to achieve the same.

ShyamsundarR: Fixed as per last comment.

pkg/rbd/nodeserver.go
```diff
@@ -148,24 +163,40 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
 			klog.Errorf("failed to unmount stagingtargetPath: %s with error: %v", stagingTargetPath, err)
+			return
 		}
 	}
-	// remove the block file created on staging path
-	if isBlock && isStagePathCreated {
+
+	// remove the file/directory created on staging path
+	if isStagePathCreated {
 		err = os.Remove(stagingTargetPath)
```
Madhu-1: need to use `os.Removeall` incase of block volume we need to delete both file and directory

pkg/rbd/nodeserver.go
```diff
@@ -148,24 +163,40 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
 	}
+
 	// Unmapping rbd device
-	if err = detachRBDDevice(devicePath); err != nil {
-		klog.Errorf("failed to unmap rbd device: %s for volume %s with error: %v", devicePath, volID, err)
+	if devicePath != "" {
+		err = detachRBDDevice(devicePath)
+		if err != nil {
+			klog.Errorf("failed to unmap rbd device: %s for volume %s with error: %v", devicePath, volID, err)
+			return
```
Madhu-1: if detach fails do we need to leave the image metadata file?

ShyamsundarR: Strictly we do not expect a NodeUnstage call (which is where the file is used) as Stage failed or has not returned in time. This is retained to maintain the order of operations and not partially cleanup the state created.

Madhu-1: I would like to get it fixed, this will leave a stale file and directory, as Kube will fail to delete staging path

ShyamsundarR: fixed.

ShyamsundarR:
> LGTM Thanks

Oh! we need another vote to get this into the code base... @humblec or @poornimag PTAL

humblec:
LGTM. Approved!

PR #512, opened 2019-07-30
Implement metrics for RBD plugin

humblec:


humblec:
@Madhu-1 @ShyamsundarR  Can you please review ?

pkg/csi-common/nodeserver-default.go
```diff
@@ -71,6 +79,96 @@ func (ns *DefaultNodeServer) NodeGetCapabilities(ctx context.Context, req *csi.N
...
+			return nil, status.Errorf(codes.InvalidArgument, "targetpath %s doesnot exist", targetPath)
+		}
+		return nil, err
+	}
+	if !isMnt {
+		return nil, status.Errorf(codes.InvalidArgument, "targetpath %s is not mounted", targetPath)
+	}
+
+	cephfsProvider := volume.NewMetricsStatFS(targetPath)
+	volMetrics, volMetErr := cephfsProvider.GetMetrics()
```
ShyamsundarR: If it is a block volume what happens when we do a statfs on a file? Which stats are returned?

I would assume we need to check if this is a block volume and report total size.

pkg/csi-common/nodeserver-default.go
```diff
@@ -17,16 +17,24 @@ limitations under the License.
...
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/klog"
+	"k8s.io/kubernetes/pkg/volume"
 )
 
 // DefaultNodeServer stores driver object
 type DefaultNodeServer struct {
 	Driver *CSIDriver
+	Type   string
```
ShyamsundarR: I did not follow why we need this passed to the NodeServer, is it used anywhere? Or, is it some form of future proofing?

pkg/csi-common/nodeserver-default.go
```diff
@@ -71,6 +79,96 @@ func (ns *DefaultNodeServer) NodeGetCapabilities(ctx context.Context, req *csi.N
...
+		if os.IsNotExist(err) {
+			return nil, status.Errorf(codes.InvalidArgument, "targetpath %s doesnot exist", targetPath)
+		}
+		return nil, err
+	}
+	if !isMnt {
+		return nil, status.Errorf(codes.InvalidArgument, "targetpath %s is not mounted", targetPath)
+	}
+
+	cephfsProvider := volume.NewMetricsStatFS(targetPath)
```
ShyamsundarR: (nit) change variable name from cephfs to something more generic. Maybe volMetrics or something?

pkg/csi-common/nodeserver-default.go
humblec: Its future proofing to treat different scenarios in this common func.

pkg/csi-common/nodeserver-default.go
humblec: Block volume invocation is still under discussion from CO side, so its fine.

pkg/csi-common/nodeserver-default.go
humblec: Changed to `CephMetricsProvider`

pkg/csi-common/nodeserver-default.go
ShyamsundarR: Should we add a check and ignore the request if it is against a block volume, as we cannot depend on what the CO may do, and it would be prudent to stick to the specification?

humblec: @ShyamsundarR As mentioned in earlier comment, its still under discussion at CO side or I dont see a risk here. I can take it as a follow up item or its well in my queue :)

ShyamsundarR: Filed https://github.com/ceph/ceph-csi/issues/519 as a tracker for the same, thanks.

Issue #516, opened 2019-08-01
cephfs-csi not able to handle massive PVC create requests


ekuric:
# Describe the bug #

cephfs-csi not able to handle massive PVC create requests 

In this test we created 
-> one project 
-> inside project 900 pods with 1 GB PVC from csi-cephfs.

Pods fails to mount PVC

# Environment details #
OCP 4.1 
```
 oc get clusterversion 
NAME      VERSION                             AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.1.0-0.nightly-2019-07-31-005945   True        False         15h     Cluster version is 4.1.0-0.nightly-2019-07-31-005945
``` 
- Image/version of Ceph CSI driver
```
# grep image operator-openshift-with-csi.yaml 
        image: rook/ceph:master
root@ip-172-31-81-56: ~/rook-v4.2-aws-test/cluster/examples/kubernetes/ceph # grep image cluster.yaml 
    # The container image used to launch the Ceph daemon pods (mon, mgr, osd, mds, rgw).
    image: ceph/ceph:v14.2.2-20190722
root@ip-172-31-81-56: ~/rook-v4.2-aws-test/cluster/examples/kubernetes/ceph # git describe 
v1.0.0-312-ge5779f7
root@ip-172-31-81-56: ~/rook-v4.2-aws-test/cluster/examples/kubernetes/ceph # git branch 
* master
root@ip-172-31-81-56: ~/rook-v4.2-aws-test/cluster/examples/kubernetes/ceph # git log --pretty=oneline | head -5
e5779f79d59b4ce58d1879b100859a2e073ea736 Merge pull request #3533 from ethan-daocloud/patch-1
0acdf9a4f1d8d6cc1834a58a246b16a7d22b2028 Merge pull request #3532 from zgfh/patch-1
aeb47eaf521d7d2fcb47b71777a6b829c42883a4 fix misspelling in bug_report.md
d50c9ef3bf0114f6eabdce1ac1d2c104722a6bf5 Merge pull request #3530 from renekalff/fix-k8s-ownerreferences
25780878ebee5b1326c550b15fe06390c2e7f181 Make the ownerreference name match the cluster name referenced
``` 
- Kubernetes cluster version
```
# kubectl version
[...]

PR #517, opened 2019-08-01
provisioners: add reconfiguring of PID limit

nixpanic:

The container runtime CRI-O limits the number of PIDs to 1024 by
default. When many PVCs are requested at the same time, it is possible
for the provisioner to start too many threads (or go routines) and
executing 'rbd' commands can start to fail. In case a go routine can not
get started, the process panics.

The PID limit can be changed by passing an argument to kubelet, but this
will affect all pids running on a host. Changing the parameters to
kubelet is also not a very elegant solution.

Instead, the provisioner pod can change the configuration itself. The
pod is running in privileged mode and can write to /sys/fs/cgroup where
the limit is configured.

With this change, the limit is configured to 'max', just as if there is
no limit at all. The logs of the csi-rbdplugin in the provisioner pod
will reflect the change it makes when starting the service:

    $ oc -n rook-ceph logs -c csi-rbdplugin csi-rbdplugin-provisioner-0
    ..
    I0726 13:59:19.737678       1 cephcsi.go:127] Initial PID limit is set to 1024
    I0726 13:59:19.737746       1 cephcsi.go:136] Reconfigured PID limit to -1 (max)
    ..

It is possible to pass a different limit on the commandline of the
cephcsi executable. The following flag has been added:

    --pidlimit=<int>       the PID limit to configure through cgroups

This accepts special values -1 (max) and 0 (default, do not
reconfigure). Other integers will be the limit that gets configured in
cgroups.

Fixes: #516

### DIFF
--- a/pkg/util/pidlimit.go
+++ b/pkg/util/pidlimit.go
@@ -1,0 +1,121 @@
+/*
+Copyright 2019 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+)
+
+const (
+	procCgroup    = "/proc/self/cgroup"
+	sysPidsMaxFmt = "/sys/fs/cgroup/pids%s/pids.max"
+)
+
+// return the cgouprs "pids.max" file of the current process
+//
+// find the line containing the pids group from the /proc/self/cgroup file
+// $ grep 'pids' /proc/self/cgroup
+// 7:pids:/kubepods.slice/kubepods-besteffort.slice/....scope
+// $ cat /sys/fs/cgroup/pids + *.scope + /pids.max
+func getCgroupPidsFile() (string, error) {
+	cgroup, err := os.Open(procCgroup)
+	if err != nil {
+		return "", err
+	}
+	defer cgroup.Close()
+
+	scanner := bufio.NewScanner(cgroup)
+	var slice string
+	for scanner.Scan() {
+		parts := strings.Split(scanner.Text(), ":")
+		if parts == nil || len(parts) < 3 {
+			continue
+		}
+		if parts[1] == "pids" {
+			slice = parts[2]
+			break
+		}
+	}
+	if slice == "" {
+		return "", fmt.Errorf("could not find a cgroup for 'pids'")
+	}
+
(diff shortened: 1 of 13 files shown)


humblec:
@nixpanic PLease check the travis failure. Thanks for the PR

nixpanic:
Tests have passed after minor modifications.

Please review @humblec @Madhu-1. Thanks!

humblec:
@nixpanic will be on this soon :) .

pkg/util/pidlimit.go
```diff
@@ -0,0 +1,121 @@
...
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+)
+
+const (
+	procCgroup    = "/proc/self/cgroup"
```
Madhu-1: @nixpanic question, `proc` and `sys` will be shared btw host and container for both cephfs and rbd?

nixpanic: Not sure what you mean by "shared"... They are available in the container, but the contents can be slightly different from what you would see on the host. I think the provisioners do not use `hostPIDs: true`, so PID 1 will be the provisioner, not systemd/init.

Madhu-1: nevermind I was aksing shared/ bind-mounted from host to the container?

rbd plugin uses `hostPIDs: true` will this causes any issue https://github.com/ceph/ceph-csi/blob/master/deploy/rbd/kubernetes/v1.14%2B/csi-rbdplugin.yaml#L19

nixpanic: No, there is no problem with the rbd plugin and `hostPIDs: true`. The PID limit that gets configured is for the container, and those limits are configured through cgroups. It is only needed to fetch the cgroup/pids name from `/proc` and then it PID limit can be reconfigured through `/sys/fs/cgroup/pids/..`. In case the executable is not running inside a container, it could adjust the PID limit for more processes on the host. In that case, just don't pass a `--pidlimit=...` option :-)

nixpanic:
rebased on master and resolved merge conflicts.

humblec:
@ekuric can you please share the test result ?

humblec:
We have test result from @ekuric. Here it is :

>
Issue not visible with image that has adapted --pidlimit @nixpanic @humblec @Madhu-1
>

With this proof merging this PR!! Thanks all.

Madhu-1:
@nixpanic Thanks for the fix, this helps a lot :+1:

Madhu-1:
@nixpanic can you make the changes in helm  chat and templates in 1.14+ for pid limit

Madhu-1:
adding DNM label to fix helm chat and 1.14 templates

nixpanic:
On Thu, Aug 08, 2019 at 05:45:30AM -0700, Madhu Rajanna wrote:
> @nixpanic can you make the changes in helm  chat and templates in 1.14+ for pid limit

Done!

obnoxxx:
It seems some travis runs have timed out. Can the tests be re-triggered?

Madhu-1:
@obnoxxx this PR need rebase sent msg to @nixpanic on slack for rebasing

we recently fixed CI issue

humblec:
Yeah, CI issue is temporarily fixed and we are asking PR rebase on PRs which are in merge queue. @obnoxxx

obnoxxx:
There seem to be more issues here. This one didn't even run, as it seems.

humblec:
Restarted the CI . Lets see what happens now :)

PR #522, opened 2019-08-05
Make parameter `pool` optional in CephFS storageclass

Daniel-Pivonka:
This makes the parameter `pool` optional in CephFS storageclass. When the 'pool' is provided the value is used and when it is not provided the default layout is used.

Fixes: #380 


humblec:
@Madhu-1

examples/cephfs/storageclass.yaml
```diff
@@ -17,7 +17,7 @@ parameters:
   # CephFS filesystem name into which the volume shall be created
   fsName: myfs
 
-  # Ceph pool into which the volume shall be created
+  # Ceph pool into which the volume shall be created (this is an optional parameter)
```
ShyamsundarR: Reword: "(optional) Ceph pool into which volume data shall be stored (also change the string in deploy-cephfs.md)

Also comment the line below (line 21)

pkg/cephfs/volume.go
```diff
@@ -74,6 +74,7 @@ func getVolumeRootPathCeph(volOptions *volumeOptions, cr *util.Credentials, volI
 }
 
 func createVolume(volOptions *volumeOptions, cr *util.Credentials, volID volumeID, bytesQuota int64) error {
+
```
ShyamsundarR: (nit) empty line

pkg/cephfs/volume.go
```diff
@@ -104,12 +105,22 @@ func createVolume(volOptions *volumeOptions, cr *util.Credentials, volID volumeI
...
 		csiSubvolumeGroup,
-		"--pool_layout", volOptions.Pool,
-		"--mode", "777",
-		"-m", volOptions.Monitors,
-		"-c", util.CephConfigPath,
-		"-n", cephEntityClientPrefix+cr.ID,
-		"--keyfile="+cr.KeyFile)
+	}
+
+	if volOptions.Pool != "" {
```
ShyamsundarR: This option can be added to the end of the CLI args, so no need to split the command at exactly this point to append the same conditionally.

ShyamsundarR:
@poornimag @ajarr PTAL as well, thanks.

humblec:
@Daniel-Pivonka there are example yamls, may be update those too ?

e2e/utils.go
```diff
@@ -203,7 +203,6 @@ func getSnapshot(path string) v1alpha1.VolumeSnapshot {
 func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework) {
 	scPath := fmt.Sprintf("%s/%s", cephfsExamplePath, "storageclass.yaml")
 	sc := getStorageClass(scPath)
-	sc.Parameters["pool"] = "myfs-data0"
```
Madhu-1: as this is optional, please add a test case with and without pool parameter.

@poornimag is there a way we can identify in the backend (via toolbox pod)this image is having pool parameter enabled?

Daniel-Pivonka:
> @Daniel-Pivonka there are example yamls, may be update those too ?

@humblec ive updated examples/cephfs/storageclass.yaml. Are there any others that need to be updated?

humblec:
> > @Daniel-Pivonka there are example yamls, may be update those too ?
> 
> @humblec ive updated examples/cephfs/storageclass.yaml. Are there any others that need to be updated?

Thanks ! This need a follow up PR to rook as well, Isnt it ?  @Madhu-1

e2e/utils.go
Daniel-Pivonka: added a test for with and without the pool

Daniel-Pivonka:
> > > @Daniel-Pivonka there are example yamls, may be update those too ?
> > 
> > 
> > @humblec ive updated examples/cephfs/storageclass.yaml. Are there any others that need to be updated?
> 
> Thanks ! This need a follow up PR to rook as well, Isnt it ? @Madhu-1

@humblec yes some changes in rook could be made to the example storageclass and the csi doc

e2e/utils.go
```diff
@@ -200,11 +200,13 @@ func getSnapshot(path string) v1alpha1.VolumeSnapshot {
 	return sc
 }
 
-func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework) {
+func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, enablePool bool) {
```
Madhu-1: how about passing a map instead of a bool value. this function can be reused if someone wants to add new parameters
to storageclass

ajarr:
> @poornimag @ajarr PTAL as well, thanks.

@ShyamsundarR , sorry missed this.

ajarr:
LGTM

e2e/utils.go
ShyamsundarR: @Madhu-1 I would assume we can merge this in, without this change. When the next parameter needs to be added we can make this a map. Not scoring an approve on this as I do not want to merge without giving you an opportunity to get back on this.

humblec:
LGTM. Approved.

Issue #527, opened 2019-08-06
Config.json invalid error is obfuscated


wilmardo:
# Describe the bug #

Unhelpful error message when the config.json isn't valid.

# Environment details #

Helm charts of #513 

# Steps to reproduce #

Use a configmap with for example a trailing comma in the monitors list
```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: ceph-csi-config
  namespace: storing
data:
    [
      {
        "clusterID": "9d9cc8f6-0843-46c3-8ca3-0309dff9978d",
        "monitors": [
          "192.168.2.100:6789",
        ]
      }
    ]
```

# Actual results #

The error message failed to fetch monitor list:

```
E0806 08:35:21.899841       1 controller.go:908] error syncing claim "4e2e414f-53be-41c2-93b7-8939d17784fc": failed to provision volume with StorageClass "csi-rbd-sc": rpc error: code = InvalidArgument desc = failed to fetch monitor list using clusterID (9d9cc8f6-0843-46c3-8ca3-0309dff9978d)
I0806 08:35:21.899893       1 event.go:209] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"default", Name:"test-ceph-claim", UID:"4e2e414f-53be-41c2-93b7-8939d17784fc", APIVersion:"v1", ResourceVersion:"105709", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "csi-rbd-sc": rpc error: code = InvalidArgument desc = failed to fetch monitor list using clusterID (9d9cc8f6-0843-46c3-8ca3-0309dff9978d)
```

# Expected behavior #

An error about an invalid config.json instead of failed to fetch monitor list.

PR #529, opened 2019-08-06
Wrap error if failed to fetch mon

Madhu-1:
This will help the user to check what's the actual error. if the config file is having an issue or the cluster id is not valid.
Fixes: #527 


humblec:
@ShyamsundarR @poornimag PTAL .

Issue #523, opened 2019-08-05
Use latest Ceph Manager extensions for RBD to speed up responses to DeleteVolume requests


ShyamsundarR:
Ceph manager has been enhanced with a few rbd commands to enable moving image deletion (and other activities) as a Ceph manager managed rbd task. This helps speed up the responses to DeleteVolume requests in the CSI driver, as image deletion takes time proportional to image size otherwise.

CSI driver for ceph can consume these rbd extensions to move the responsibility of image deletion to the Ceph Manager, and thus speed up the response for a DeleteVolume operation.

The ceph enhancement can be found here: https://github.com/ceph/ceph/pull/29054

ShyamsundarR:
@dillaman I tried the following experiment, as I understood the deletion process as a 2 step process, like so,
- Invoke `ceph rbd task add remove <image_spec>` to move the image to trash
- Followed by, `ceph rbd task add trash remove <image_id_spec>` to empty the image from trash

When I installed a cluster based of the Ceph master daemon image, I find that the first command itself deletes the image, and there is nothing in trash or otherwise to further invoke the second command.

What am I missing? As I was under the impression we needed to stash the `<image-id>` and then invoke the removal from trash of the image, looks like this step is not required.

Here are the outputs of the various commands that I attempted: (**note:** tbox is a script enabling a convoluted kubectl command execution within one of Rooks toolbox pods)
```
// create an image
$ ./tbox rbd create -s 10G replicapool/trashtestimage

// Get its info
$ ./tbox rbd ls -p replicapool
trashtestimage
$ ./tbox rbd info replicapool/trashtestimage
rbd image 'trashtestimage':
	size 10 GiB in 2560 objects
	order 22 (4 MiB objects)
	snapshot_count: 0
	id: 3d4ecb97f770
	block_name_prefix: rbd_data.3d4ecb97f770
	format: 2
	features: layering, exclusive-lock, object-map, fast-diff, deep-flatten
	op_features: 
	flags: 
	create_timestamp: Mon Aug  5 17:24:30 2019
	access_timestamp: Mon Aug  5 17:24:30 2019
	modify_timestamp: Mon Aug  5 17:24:30 2019

// Invoke adding a task to remove the image
$ ./tbox ceph rbd task add remove replicapool/trashtestimage
[...]

PR #531, opened 2019-08-06
Add support to use ceph manager rbd command to delete an image

ShyamsundarR:
Image deletion takes time proportional to the size of the
image. Hence, ceph manager is enhanced to support async
deletion of an image, or rather passing the task of
deleting an image to the ceph manager.

This commit leverages the ceph manager enhancement in the CSI code.

NOTE: This is tested against a ceph cluster that is running
Ceph master version of the code. Once other releases
catch up in terms of the feature, the optimization would be
available to the CSI driver as well.

Fixes: #523

### DIFF
--- a/pkg/rbd/rbd_util.go
+++ b/pkg/rbd/rbd_util.go
@@ -42,6 +42,12 @@ const (
 	rbdImageWatcherFactor    = 1.4
 	rbdImageWatcherSteps     = 10
 	rbdDefaultMounter        = "rbd"
+
+	// Output strings returned during invocation of "ceph rbd task add remove <imagespec>" when
+	// command is not supported by ceph manager. Used to check errors and recover when the command
+	// is unsupported.
+	rbdTaskRemoveCmdInvalidString1 = "no valid command found"
+	rbdTaskRemoveCmdInvalidString2 = "Error EINVAL: invalid command"
 )
 
 // rbdVolume represents a CSI volume and its RBD image specifics
@@ -163,6 +169,31 @@ func rbdStatus(pOpts *rbdVolume, cr *util.Credentials) (bool, string, error) {
 	return false, output, nil
 }
 
+// rbdManagerTaskDelete adds a ceph manager task to delete an rbd image, thus deleting
+// it asynchronously. If command is not found returns a bool set to false
+func rbdManagerTaskDeleteImage(pOpts *rbdVolume, cr *util.Credentials) (bool, error) {
+	var output []byte
+
+	args := []string{"rbd", "task", "add", "remove",
+		pOpts.Pool + "/" + pOpts.RbdImageName,
+		"--id", cr.ID,
+		"--keyfile=" + cr.KeyFile,
+		"-m", pOpts.Monitors,
+	}
+
+	output, err := execCommand("ceph", args)
+	if err != nil {
+		if strings.Contains(string(output), rbdTaskRemoveCmdInvalidString1) &&
+			strings.Contains(string(output), rbdTaskRemoveCmdInvalidString2) {
+			klog.Infof("cluster with cluster ID (%s) does not support Ceph manager based rbd image"+
+				" deletion (minimum ceph version required is v14.2.3)", pOpts.ClusterID)
+			return false, err
+		}
+	}
+
+	return true, err
+}
+
 // deleteImage deletes a ceph image with provision and volume options.
 func deleteImage(pOpts *rbdVolume, cr *util.Credentials) error {
 	var output []byte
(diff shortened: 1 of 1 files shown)


ShyamsundarR:
Initial static check build failed with the following  error, re-triggered the same.

```
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
apt-get.diagnostics

apt-get install failed

The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install realpath ruby" failed and exited with 100 during .
```

pkg/rbd/rbd_util.go
```diff
@@ -163,6 +163,30 @@ func rbdStatus(pOpts *rbdVolume, cr *util.Credentials) (bool, string, error) {
...
+		"--id", cr.ID,
+		"--keyfile=" + cr.KeyFile,
+		"-m", pOpts.Monitors,
+	}
+
+	output, err := execCommand("ceph", args)
+	if err != nil {
+		if strings.Contains(string(output), "no valid command found") &&
+			strings.Contains(string(output), "Error EINVAL: invalid command") {
+			klog.Infof("cluster with cluster ID (%s) does not support Ceph manager based rbd image deletion", pOpts.ClusterID)
```
Madhu-1: log this as a warning

Madhu-1: I think the first check for "no valid command found" will be sufficient for this isn't it?

Madhu-1: what is the minimum supported ceph version for this command?
please add a comment

ShyamsundarR: I am going to do the following, let me know if it works for you.
- Retain info, as this is just informational and not a warning message (using info here was by purpose)
- Retain 2 checks as I prefer to be sure, and use more than one message output available to detect the same
- Add the minimum supported version as 14.3, but it may so happen that the feature maybe backported to mimic as well, at which point we may need to revisit this, and so will add a comment in the code around this

dillaman: >     * Add the minimum supported version as 14.3, but it may so happen that the feature maybe backported to mimic as well, at which point we may need to revisit this, and so will add a comment in the code around this

It would be v14.2.3 -- and you are correct that it will be backported to mimic as well.

pkg/rbd/rbd_util.go
```diff
@@ -163,6 +163,30 @@ func rbdStatus(pOpts *rbdVolume, cr *util.Credentials) (bool, string, error) {
...
+	args := []string{"rbd", "task", "add", "remove",
+		pOpts.Pool + "/" + pOpts.RbdImageName,
+		"--id", cr.ID,
+		"--keyfile=" + cr.KeyFile,
+		"-m", pOpts.Monitors,
+	}
+
+	output, err := execCommand("ceph", args)
+	if err != nil {
+		if strings.Contains(string(output), "no valid command found") &&
```
humblec: @ShyamsundarR if possible define these ( `no valid command found` ) as Error var/const.

ShyamsundarR: As a practice in other places in the code I/we have not done so, do we still want to change it? (searching for `strings.Contains` in the code shows up 7 instances)

ShyamsundarR:
> lgtm - but at some point we will need to wrap the concept of deleting an RBD images with attached clones w/o returning an error. Maybe we need another librbd concept for auto-deleting parent images from the trash when the last clone is deleted. Right now you can schedule a deletion of an image with linked clones but it will just keep failing and restarting. Perhaps if I tweak the logic to only start the background task once all children are deleted it would be cleaner. I’ll take that as a task for myself when I get back.

Agreed to most of the above, but here is what I observed, deletion of an image (or invoking `ceph rbd task add remove <imagespec>`) fails for images with snapshots or clones, IOW fails to trigger the task I would say. So the part of your statement about the deletion "failing and restarting" I did not understand if it related to task above. Sample output below,
```
#ceph rbd task add remove replicapool/testimg001
Error EBUSY: [errno 16] RBD image is busy (Image replicapool/testimg001 has snapshots)
command terminated with exit code 16
```

dillaman:
> IOW fails to trigger the task I would say

As part of that effort, I'll have it filter out snapshots in the trash namespace. Given all the previous discussions about how to create snapshots, the goal would be that you would never have a non-trashed snapshot on an RBD image.

pkg/rbd/rbd_util.go
humblec: @ShyamsundarR the problem is that, if there changes or some manual errors in the code ( say a space or similar) can cause these code path to fail. Instead if we have well defined constant/vars it is less error prone. So , I would recommend to have it addressed.

ShyamsundarR: Fixed to use constant strings.

humblec: Thanks!

Issue #528, opened 2019-08-06
make csi deployment optional in E2E


Madhu-1:
# Describe the bug #

currently, we deploy csi in E2E. so we cannot run the E2E against the already deployed cluster. if we can separate out the E2E and csi deployment, it will help us/others to run E2E tests and make sure things are working fine.

Madhu-1:
This is already fixed, closing this one

PR #536, opened 2019-08-08
adds flags to make csi deployment optional

umangachapagain:
Signed-off-by: Umanga Chapagain <chapagainumanga@gmail.com>

--ceph-csi flag to enable/disable ceph-csi (default: true)
--cephfs-csi flag to enable/disable cephfs-csi only (default: false)
--rbd-csi flag to enable/disable rbd-csi only (default: false)

any combination of flags can be used to skip respective tests

Do you have any questions?

For example:

## Related issues ##

Fixes: #528 

## Future concerns ##

### DIFF
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -40,6 +40,9 @@ var _ = Describe("cephfs", func() {
 	f := framework.NewDefaultFramework("cephfs")
 	// deploy cephfs CSI
 	BeforeEach(func() {
+		if !csiRequired && !cephFsRequired {
+			Skip("Skipping Test cephfs CSI")
+		}
 		updateCephfsDirPath(f.ClientSet)
 		createFileSystem(f.ClientSet)
 		createConfigMap(cephfsDirPath, f.ClientSet, f)
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -12,9 +12,12 @@ import (
 )
 
 var (
-	RookVersion   string
-	rookRequired  bool
-	deployTimeout int
+	RookVersion    string
+	rookRequired   bool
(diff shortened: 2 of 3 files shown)


humblec:
@umangachapagain Welcome on board!! Thanks for the PR :100:

ShyamsundarR:
@umangachapagain the change looks fine based on flags.

I wanted to understand if we could have leveraged `--focus` and `--skip` options based on [focused specs](https://onsi.github.io/ginkgo/#focused-specs) section of ginkgo instead.

I am not sure if passing a focus of `cephfs` or `RBD` as the expressions would work for all tests in that `Describe` section, but wanted to understand if this was possible.

NOTE: I am new to ginkgo so this is more a suggestion than a comment at the moment.

umangachapagain:
> NOTE: I am new to ginkgo so this is more a suggestion than a comment at the moment.
>
Even I just got started with it. Will read more and get back to you on this.

e2e/cephfs.go
```diff
@@ -40,6 +40,9 @@ var _ = Describe("cephfs", func() {
 	f := framework.NewDefaultFramework("cephfs")
 	// deploy cephfs CSI
 	BeforeEach(func() {
+		if !csiRequired && !cephFsRequired {
```
Madhu-1: @umangachapagain skipping in beforeEach will skip filesystem creation, secret creation. you need to skip the deletion part also.

umangachapagain: You mean skip the afterEach ? After skipping in beforeEach, I did not see afterEach getting executed so did not add skips there.

Madhu-1:
closing this as we have flags to skip cephfs and rbd deployment in E2E

PR #538, opened 2019-08-08
Removing appendix from license.

kshithijiyer:
Removing the appendix part from the license as it is just a sample/example boiler plate notice which is to be added to all the files in any project which is licensed under the Apache v2 license. Ideally it should be removed from LICENSE file when added to the project as you would only want the terms and conditions under which a project is open source in the LICENSE file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -175,28 +175,3 @@
       of your accepting any such warranty or additional liability.
 
    END OF TERMS AND CONDITIONS
-
-   APPENDIX: How to apply the Apache License to your work.
-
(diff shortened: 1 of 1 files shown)


humblec:
@kshithijiyer Welcome on board :), btw, any specific reason to remove this section  ? If yes, can you update the PR description ?

kshithijiyer:
> @kshithijiyer Welcome on board :), btw, any specific reason to remove this section ? If yes, can you update the PR description ?

Issue #471, opened 2019-07-12
Context based logging is required in the CSI code base


ShyamsundarR:
### Describe the feature you'd like to have

All logs should carry a request or unique context. This is to help correlate logs form a single request when there are multiple requests in flight.

### What is the value to the end user? (why is it a priority?)

Currently the logs are intertwined when there are multiple requests. With context added to logs, we can troubleshoot and identify logs for a single event much faster and with certainty.

### How will we know we have a good solution? (acceptance criteria)

Close to all messages for a request have a unique ID, would ensure we have a good solution for the problem. Further, if we could provide tools/scripts that can extract and dump logs for a request would be a useful win.

### Additional context

Currently klog (which is the infrastructure in use for logging) does not seem to have contextual logging support.

Golang "context" module seems to provide what we need, but requires changes to all functions in the code (which is feasible as the code base is not too huge). This seems to be the course to take, unless there are other choices.

Madhu-1:
@ShyamsundarR unique ID should be the PVC/snapshot name or volume/snapshot ID?

ShyamsundarR:
> @ShyamsundarR unique ID should be the PVC/snapshot name or volume/snapshot ID?

ID that all messages for a request can log in addition to the message. We can even choose different IDs for different requests, as long as they are used for all the messages.

So I would say we can even create/craft one that can help identify a request, as soon as we recieve on one of the gRPC calls. It does not need to be from the RPC content (i.e the VolID/Name).

ShyamsundarR:
> > @ShyamsundarR unique ID should be the PVC/snapshot name or volume/snapshot ID?
> 
> ID that all messages for a request can log in addition to the message. We can even choose different IDs for different requests, as long as they are used for all the messages.
> 
> So I would say we can even create/craft one that can help identify a request, as soon as we recieve on one of the gRPC calls. It does not need to be from the RPC content (i.e the VolID/Name).

One further clarification, as we have RPCs being re-triggered due to timeouts, we should refrain from using the VolID/ReqName as the ID to distinguish between repeated RPC calls for the same ID or name.

Madhu-1:
Fixed by #540

PR #540, opened 2019-08-09
Implement context based logging

Daniel-Pivonka:
Begin exploring context based logging.

For now I've added a basic way to implement context based logging. 

Each request is intercepted and an ID is added to the requests context. The context is passed to all functions used in the request. The ID embedded in the context is then printed in all log messages related to the request to make debugging easier. 

This will allow logs to be sorted by ID. For example I can just get request 6's logs 

```
[dpivonka@dhcp-41-48 k8s-cluster]$ ./k logs pod/csi-rbdplugin-khj69 -c csi-rbdplugin | grep "ID: 6"
I0809 15:21:13.732979   18034 utils.go:118] ID: 6 GRPC call: /csi.v1.Node/NodePublishVolume
I0809 15:21:13.732992   18034 utils.go:119] ID: 6 GRPC request: {"staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-43875a1c-bab9-11e9-bd5d-525400ccac75/globalmount","target_path":"/var/lib/kubelet/pods/49fcfca6-bab9-11e9-bd5d-525400ccac75/volumes/kubernetes.io~csi/pvc-43875a1c-bab9-11e9-bd5d-525400ccac75/mount","volume_capability":{"AccessType":{"Mount":{"fs_type":"ext4","mount_flags":["discard"]}},"access_mode":{"mode":1}},"volume_context":{"clusterID":"c4d308e2-22fe-416d-80e6-c4dbeba23ce3","imageFeatures":"layering","imageFormat":"2","pool":"rbd","storage.kubernetes.io/csiProvisionerIdentity":"1565363218262-8081-rbd.csi.ceph.com"},"volume_id":"0001-0024-c4d308e2-22fe-416d-80e6-c4dbeba23ce3-0000000000000002-43a2b5b5-bab9-11e9-97ba-baf9e1247865"}
I0809 15:21:13.744812   18034 nodeserver.go:288] ID: 6 target /var/lib/kubelet/pods/49fcfca6-bab9-11e9-bd5d-525400ccac75/volumes/kubernetes.io~csi/pvc-43875a1c-bab9-11e9-bd5d-525400ccac75/mount
I0809 15:21:13.757023   18034 nodeserver.go:221] ID: 6 rbd: successfully mounted stagingPath /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-43875a1c-bab9-11e9-bd5d-525400ccac75/globalmount to targetPath /var/lib/kubelet/pods/49fcfca6-bab9-11e9-bd5d-525400ccac75/volumes/kubernetes.io~csi/pvc-43875a1c-bab9-11e9-bd5d-525400ccac75/mount
I0809 15:21:13.757051   18034 utils.go:124] ID: 6 GRPC response: {}
```

I've only added this to RBD NodePublishVolume requests for now to begin a discussion on how this can be implemented across the entire project.

My initial thought is to create a set of wrapper function for klog: Info, Error, Warning, Fatal. 

For example they would look something like this when used

csiLog.Infof(ctx, "rbd: successfully mounted stagingPath %s to targetPath %s", stagingPath, targetPath)

These functions will take the context, format string, and arguments variables then inject the id at the begging of the string and print with klog. 

resulting in a log looking like this 
```
ID: 6 rbd: successfully mounted stagingPath /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-43875a1c-bab9-11e9-bd5d-525400ccac75/globalmount to targetPath /var/lib/kubelet/pods/49fcfca6-bab9-11e9-bd5d-525400ccac75/volumes/kubernetes.io~csi/pvc-43875a1c-bab9-11e9-bd5d-525400ccac75/mount
```

Do this will make changing all log messages over easier than manually editing them like I have done so far. 

Is this a good approach? are there any other thoughts on how this could be done? 

#### Related issue ####
#471


Daniel-Pivonka:
Basic example of klog wrapper

```
package main

import (
	"context"

	"k8s.io/klog"
)

func Infof(ctx context.Context, format string, args ...interface{}) {
	id := "ID: %d "

	format = id + format 

	args = append([]interface{}{ctx.Value("ID")}, args...)

	klog.Infof(format, args...)
}

func main() {
	klog.InitFlags(nil)

	ctx := context.WithValue(context.Background(), "ID", 1)

	Infof(ctx, "test %d %s", 6, "dan")
}
```
 Output: 
`I0809 15:39:55.573081    4068 test.go:16] ID: 1 test 6 dan`

Madhu-1:
@Daniel-Pivonka am good with writing wrap for logging, one thing from my side is
check if the request  contains the name or ID and log it.

for createRequest, we will get the `request.GetName()` and for volume Delete and other node plugin operation we will get `request.GetVolumeId()` this will help us to identify the log belongs to which request,
 
we can do the improvement for snapshot which should log source volume ID  and for  volume Create Request we need to log source volume ID and source snapshot ID

you can handle the snapshot request logging by checking the request type

ShyamsundarR:
> @Daniel-Pivonka am good with writing wrap for logging, one thing from my side is
> check if the request contains the name or ID and log it.

@Madhu-1 Is this in addition to the generate per message ID?

As stated [here](https://github.com/ceph/ceph-csi/issues/471#issuecomment-513855796) due to timeouts and repeated invocations of the RPCs we need to use an `generatedID` outside of the passed in `requestName` or `VolumeID` to recognize messages from the same idempotent RPC when they are resent.

Madhu-1:
> o recognize messages from the same idempotent RPC when they are resent.

yes, this is addition, and it will help to filter out the logs easily.

Daniel-Pivonka:
@Madhu-1 @ShyamsundarR Ive implemented the klog wrapper and used it in the NodePublishVolume request. let me know what you think.

pkg/csi-common/utils.go
```diff
@@ -105,14 +107,22 @@ func RunControllerandNodePublishServer(endpoint string, d *CSIDriver, cs csi.Con
 	s.Wait()
 }
 
+var id uint64
```
Madhu-1: how this behaves if you have parallel requests?

Madhu-1: never mind didn't notice `atomic.Add` below

Madhu-1:
how about moving to logrus `github.com/sirupsen/logrus`  from klog?
@humblec ?

humblec:
> how about moving to logrus `github.com/sirupsen/logrus` from klog?
> @humblec ?

@Madhu-1 lets stick with `klog` as its the one kube use. :+1:

Daniel-Pivonka:
@Madhu-1 @humblec Is this wrapper ok then? if so can we merge it and I will open another pr changing more requests. I will try to do as many as I can but if we could spread out some of the load of changing all the request that would be helpful.

pkg/util/log.go
```diff
@@ -0,0 +1,183 @@
+package util
```
Madhu-1: license?

pkg/util/log.go
```diff
@@ -0,0 +1,183 @@
...
+	format = id + format
+
+	args = append([]interface{}{ctx.Value(Key)}, args...)
+
+	klog.V(klog.Level(v.level)).Infof(format, args...)
+}
+
+func Infof(ctx context.Context, format string, args ...interface{}) {
+	format = id + format
+
```
Madhu-1: avoid newlines inside functions  in this file

pkg/util/log.go
```diff
@@ -0,0 +1,183 @@
...
+	level int
+}
+
+func V(i int) *Verbosity {
+	return &Verbosity{level: i}
+}
+
+type contextKey string
+
+var Key = contextKey("ID")
```
Madhu-1: do we need to expose this var?

Daniel-Pivonka: yes is is used [here](https://github.com/ceph/ceph-csi/blob/7dedcda5fdb3a67218db1f66245da786a397589d/pkg/csi-common/utils.go#L114)

Madhu-1:
@humblec @ShyamsundarR PTAL

Issue #507, opened 2019-07-26
Rbd map takes more than 3 minutes to map a volume on node


Madhu-1:
# Describe the bug #
during node staging for some requests rbd map takes more than 3 minutes to map the rbd volume on the node (note volume size was 1 Gb) this could be a bottleneck to support more parallel request from node plugin, this will affect the pod creation time.

sometimes the map fails without and we can see rbd volume mapped to the node, this will affect 
the volume mounting to other node and also volume deletion as this volume is marked as being used.

@dillaman can you  please let me  know what logs you need in this case

# Environment details #

- Image/version of Ceph CSI driver: canary
- helm chart version
- Kubernetes cluster version:1.15.0
- Logs

# Steps to reproduce #
* Create 50/100 parallel pod which uses `RWO` PVC

Madhu-1:
@rootfs @travisn how this is solved in rook and Kube rbd provisioner?

dillaman:
There is really no reason why "rbd map" should ever take 3 minutes. Has anyone attempted to 'strace' or gcore the 'rbd' process when this occurs? I still suspect it's related to udev but I haven't been provided w/ my requested data yet.

Madhu-1:
> There is really no reason why "rbd map" should ever take 3 minutes. Has anyone attempted to 'strace' or gcore the 'rbd' process when this occurs? 
 i havent tried it. let me get the strace for rbd process

I still suspect it's related to udev but I haven't been provided w/ my requested data yet.

@dillaman strace is enough or you need any other data?

dillaman:
@Madhu-1 It's it's udev I would suspect that strace would show it stuck in a `poll` syscall:

> [pid 19545] poll([{fd=7, events=POLLIN}], 1, 120000) = 1 ([{fd=7, revents=POLLIN}])
[...]

PR #537, opened 2019-08-08
Testing E2E  [dont merge this one]

Madhu-1:
# Describe what this PR does #

we have seen better performance in device format and mounting by setting the fstype to xfs
from default ext4.

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

## Related issues ##

Mention any github issues relevant to this PR. Adding below line
will help to auto close the issue once the PR is merged.

Fixes: #507

PR #545, opened 2019-08-13
Add xfs fstype as default type in storageclass

Madhu-1:
we have seen better performance in device format and mounting by setting the fstype to xfs from default ext4.

Fixes: #507

Note: #537 will be used for testing CI

### DIFF
--- a/examples/rbd/storageclass.yaml
+++ b/examples/rbd/storageclass.yaml
@@ -30,7 +30,9 @@ parameters:
    csi.storage.k8s.io/provisioner-secret-namespace: default
    csi.storage.k8s.io/node-stage-secret-name: csi-rbd-secret
    csi.storage.k8s.io/node-stage-secret-namespace: default
-
+   # Specify the filesystem type of the volume. If not specified,
+   # csi-provisioner will set default as `ext4`.
+   csi.storage.k8s.io/fstype: xfs
    # uncomment the following to use rbd-nbd as mounter on supported nodes
    # mounter: rbd-nbd
 reclaimPolicy: Delete


Madhu-1:
not a good idea to default it in code, we should give the flexibility to use to select ext4 or xfs or other formats, as in storage class will ask the user to use xfs for better performance.

humblec:
LGTM. Approved.

ShyamsundarR:
> not a good idea to default it in code, we should give the flexibility to use to select ext4 or xfs or other formats, as in storage class will ask the user to use xfs for better performance.

I meant change the current default in code from `ext4` to `xfs`, we will still retain the ability to mention either or totally another file system in its place.

It avoids the need to mention this in the storageclass as we are overriding the default.

Madhu-1:
> > not a good idea to default it in code, we should give the flexibility to use to select ext4 or xfs or other formats, as in storage class will ask the user to use xfs for better performance.
> 
> I meant change the current default in code from `ext4` to `xfs`, we will still retain the ability to mention either or totally another file system in its place.
> 
> It avoids the need to mention this in the storageclass as we are overriding the default.

in code where? rbd nodeplugin?

ShyamsundarR:
> > > not a good idea to default it in code, we should give the flexibility to use to select ext4 or xfs or other formats, as in storage class will ask the user to use xfs for better performance.
> > 
> > 
> > I meant change the current default in code from `ext4` to `xfs`, we will still retain the ability to mention either or totally another file system in its place.
> > It avoids the need to mention this in the storageclass as we are overriding the default.
> 
> in code where? rbd nodeplugin?

Yes, currently we do not check if passed in fsType is empty and set a default, leaving the mount package to set it for us (which defaults to ext4). Instead if we wanted **our** default to be xfs, and not miss it in case users failed to specify the same in the storage class, we could add it to our code.

In any case, please note that I am not nack'ing this PR, it is a suggestion to set favorable defaults for rbd instead of pushing the onus to the user.

humblec:
@ShyamsundarR it was actually discussed and agree that, if we can make  `xfs` from our code as default, lets handle it.

Madhu-1:
fstype is set to ext4 if not passed either by Kube or  sidecar container, I can't remember Kube code check  sidecar code https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L437

Madhu-1:
@humblec @ShyamsundarR i sure that we will get ext4 as default during nodestage volume.  if you want I can test and confirm that

ShyamsundarR:
> @humblec @ShyamsundarR i sure that we will get ext4 as default during nodestage volume. if you want I can test and confirm that

Looks like there is no option to control this in the sidecar, maybe this is a case to request addition of on in the kubernetes csi repository for use?

humblec:
> @humblec @ShyamsundarR i sure that we will get ext4 as default during nodestage volume. if you want I can test and confirm that>

Sure. I think I was not clear in my last comment. The external provisioner definitely set it to `ext4` as we discussed and verified from code. What I was trying to convey here was, afacit, setting the SC parameter with `xfs` as the default option is the only option we could use, which this PR does.

humblec:
@Madhu-1 can you open an issue in external provisioner with a request that, is there any way SP can set this regardless or by neglecting SC option ? I see less possibility to get it acked though.

ShyamsundarR:
> @Madhu-1 can you open an issue in external provisioner with a request that, is there any way SP can set this regardless or by neglecting SC option ? I see less possibility to get it acked though.

I am not stating to **neglect** the SC option, that should be honored and not ignored. I am stating we ask for an option to set the default fsType and not default to xfs in the sidecar.

Madhu-1:
> @Madhu-1 can you open an issue in external provisioner with a request that, is there any way SP can set this regardless or by neglecting SC option ? I see less possibility to get it acked though.

same here let's see if Kube-csi accepts.

humblec:
> > @Madhu-1 can you open an issue in external provisioner with a request that, is there any way SP can set this regardless or by neglecting SC option ? I see less possibility to get it acked though.
> 
> same here let's see if Kube-csi accepts.

Yeah.. may be an arg to sidecar as we have `--timeout`, but I doubt it will be accepted as there will be many options which can be configured this way, but we can not have cli args for everything.  But. lets give a try :)

Issue #271, opened 2019-03-20
Provide hints to the plugin binaries on their roles


ShyamsundarR:
Currently each plugin (CephFS or rbd in their Run routines), create and start all 3 services (Identity, Node and Controller). This is incorrect, but does not cause issues, as the consumer of these services are the k8s sidecars, which are selectively installed with the plugin. Hence, due to the sidecars the right services get used.

Suggest that we make this configurable, such that only the required services are created and started. Such a configuration also provides the feasibility in the future to do certain controller level tasks in the startup (for now load the config maps etc.) and avoid them on node services, and vice versa.

gman0:
Could you please elaborate how is this incorrect? The CSI spec allows either of those configurations. Anyway, as you pointed out, there could be some advantages.

tombarron:
> The CSI spec allows either of those configurations.

Yeah, it explicitly a variety of [deployment architectures](https://github.com/container-storage-interface/spec/blob/master/spec.md#architecture) including ones where "unified Plugin component supplies both the Controller Service and Node Service."  Still worth considering the advantages w.r.t. doing some specialized tasks in startup, I'm not disagreeing with that.

ShyamsundarR:
This is how I see it.
- At present we [create](https://github.com/ceph/ceph-csi/blob/csi-v1.0/pkg/rbd/rbd.go#L117-L123) and [start](https://github.com/ceph/ceph-csi/blob/csi-v1.0/pkg/rbd/rbd.go#L130) CS, NS and IS on all instances of the plugin
- IS running on all instances of the plugin is required
- CS in our case should run only as a singleton in the cluster, hence creating and starting a listener for the CS in each running instance of the plugin looks incorrect
- NS should run on all nodes, and we deploy the same using a DaemonSet, but the StatefulSet that deploys the provisioner would also create/listen on an instance of the NS, which looks incorrect

Why it still works is because, the client end of the listeners that we create are the kubernetes sidecars, and each deployment only packages either the provisioner or the node driver as required, and hence we do not see requests for the other in our running instances.

Side note: The IS as per the spec should claim all capabilities in GetPluginCapabilities hence that would not be specialized based on the actual components the plugin is running.

The unified plugin example in the spec to my understanding, is applicable when we can run CS on all nodes of the cluster and not as a singleton as in our case. IOW, I believe the unified case is not about code layout, but capabilities of each instance that is running, hence we can have the same binary, but we need to ensure we start listening on services that we would serve, and not all.

[...]

PR #550, opened 2019-08-14
Make cephcsi controller server and node server configurable

Madhu-1:

start controller or node server based on configuration, if both controller and node server flags are set/unset cephcsi will start both servers, if only one flag is set, it will start relevant service.

move flag configuration variable to util remove unwanted checks remove getting driver type from the binary name

rename rbd.go to driver.go  to keep code struct similar with cephfs  renamed rbd.go to driver.go

Do you have any questions?

Fixes: #271 

## Future concerns ##


Madhu-1:
@humblec @ShyamsundarR @poornimag PTAL

pkg/cephfs/driver.go
```diff
@@ -147,20 +147,30 @@ func (fs *Driver) Run(conf *util.Config, cachePersister util.CachePersister) {
...
 
-	fs.cd.AddControllerServiceCapabilities([]csi.ControllerServiceCapability_RPC_Type{
-		csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
-	})
-
-	fs.cd.AddVolumeCapabilityAccessModes([]csi.VolumeCapability_AccessMode_Mode{
-		csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
-	})
-
+	if conf.IsControllerServer || !(conf.IsControllerServer && conf.IsNodeServer) {
```
ShyamsundarR: If isNS is `true`and isCS is `false` we would still go ahead and add the capability information as the `!(isCS && isNS)` evaluates to true, similarly in line 170, if only one of the services are true, then both services would be started.

I think the code intends to do the following in line 150,
```
if isCS || !isNS
```

and in line 170 this,
```
if !isCS && !isNS
```

Madhu-1: yes, updated code now

humblec:
@Madhu-1 please rebase. We will get this in `1.2.0`.

Madhu-1:
@humblec done

Issue #305, opened 2019-04-04
Fix image vulnerabilities found in quay.io


Madhu-1:
currently there are 2 vulnerabilities issues  found in both [rbdplugin](https://quay.io/repository/cephcsi/rbdplugin/manifest/sha256:893f9d27d3abfb3a4a263566f638af6b98c0660ad684f94ecd4e0586a90c05ec?tab=vulnerabilities&fixable=true)  and [cephfsplugin](https://quay.io/repository/cephcsi/cephfsplugin/manifest/sha256:60a139192580f4d06aabe075842cd987d8147e569b6d04e7a9b849c64d35a1d3?tab=vulnerabilities).

Madhu-1:
CC @humblec @leseb

leseb:
This basically means:

* build more often
* add a timestamp to each build

Madhu-1:
@humblec looks like this is not fixed. can you take a look at it https://quay.io/repository/cephcsi/cephcsi/manifest/sha256:d34fa7a188662e385437eee2dd2ce72481e0f4b678a8c548396ad0169fdf7b9a?tab=vulnerabilities&fixable=true

humblec:
@Madhu-1 I was looking at this today morning. There are still good number of vulnerabilities. Enabling epel repo is one of the solution mentioned and removal of iscsi repo.  Will experiment more , however this should get some or good amount of support from Base Image I believe. @leseb can you please take a look at these vulnaribilites and any thoughts on how to fix it in base image ?

humblec:
OK.. so most of the vulnarabilities are addressed now and it an ongoing process. Atm, it report 1 vulnariability which would be addressed with the base images we use and its common for different dockerfiles in the ceph org. There is no point in keeping an issue opened here for long. Closing it for now.

PR #551, opened 2019-08-14
Fix the vulnarabilities in the image.

humblec:
Fix#https://github.com/ceph/ceph-csi/issues/305

### DIFF
--- a/deploy/cephcsi/image/Dockerfile
+++ b/deploy/cephcsi/image/Dockerfile
@@ -2,6 +2,8 @@ FROM ceph/ceph:v14.2
 LABEL maintainers="Ceph-CSI Authors"
 LABEL description="Ceph-CSI Plugin"
 
+# Removing ceph-iscsi repo to workaround the repo issue while upgrading
+RUN rm -f /etc/yum.repos.d/ceph-iscsi.repo && yum -y update && yum clean all
 ENV CSIBIN=/usr/local/bin/cephcsi
 
 COPY cephcsi $CSIBIN


Madhu-1:
@humblec whats the docker image size after this ?  do we need to do this or ask ceph base image to rebuild frequently as these vulnerabilities introduced in ceph base image?

humblec:
> @humblec whats the docker image size after this ? do we need to do this or ask ceph base image to rebuild frequently as these vulnerabilities introduced in ceph base image?

The image size should not increase actually as we do `yum update` and `yum clean` , but I will list it down.  Even if they do, this helps us to update the image with latest packages in our image whenever we build as there can be image updates between their updates and ours especially when we use their released base images.

humblec:
@Madhu-1 thanks for the help on disabling the problematic repo :(.  The extra download size was `13M` as discussed.  Please review.

Madhu-1:
@leseb can something be done on ceph container side, these issues are found in ceph base image. 
am sure  ceph base image will have the same issue

humblec:
@ShyamsundarR  @poornimag

humblec:
Considering this is going to avoid around 35 vulnerabilities in the image which is critical, lets get this in.

deploy/cephcsi/image/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM ceph/ceph:v14.2
 LABEL maintainers="Ceph-CSI Authors"
 LABEL description="Ceph-CSI Plugin"
 
+# Removing ceph-iscsi repo to workaround the repo issue while upgrading
+RUN rm -f /etc/yum.repos.d/ceph-iscsi.repo && yum -y update && yum clean all
```
nixpanic: deleting the `.repo` file is not appropriate. without it, it will be difficult to inspect the container and see where the installed packages come from.

It is better to `--disablerepo=<name>` where `<name>` is the name listed in the `.repo` file. The option can be used multiple times in case needed.

humblec: @nixpanic before removing I tried to disable the repo by the options it provided it was not working for me, so the removal. Is it working for you if you disable the repo file ?

ShyamsundarR:
I would still prefer the base image fix the update of packages, and we just consume the same. Current scheme is a workaround for the base ceph image not being updated IMO. Tagging @leseb once more for feedback.

deploy/cephcsi/image/Dockerfile
nixpanic: It should, we do that with other containers and installation scripts too. I have not tried it with this container yet, and I do not know what the contents of `ceph-iscsi.repo` so I can not give an immediate example of how to structure the yum command.

humblec:
> I would still prefer the base image fix the update of packages, and we just consume the same. Current scheme is a workaround for the base ceph image not being updated IMO. Tagging @leseb once more for feedback.

@ShyamsundarR Having `yum update` in  container is a common practice. We do that for almost all containers. Base image also need update , but thats actually different or separately tracked.

deploy/cephcsi/image/Dockerfile
humblec: >
  yum --disablerepo=ceph-iscsi-noarch ...

     4. Disable the repository permanently, so yum won't use it by default. Yum
        will then just ignore the repository until you permanently enable it
        again or use --enablerepo for temporary usage:

            yum-config-manager --disable ceph-iscsi-noarch
        or
            subscription-manager repos --disable=ceph-iscsi-noarch

>

These are the main options, tried these and its failing.

nixpanic: Do you have the exact snippet of the `Dockerfile` and the error messages of how/when it fails?

humblec:
Merging this one as the preparation for `1.2.0` release. Will sort out the repo issues in a follow up PR.

deploy/cephcsi/image/Dockerfile
nixpanic:
```
sh-4.2# grep -e '^\[' -e ^enabled /etc/yum.repos.d/ceph-iscsi.repo 
[ceph-iscsi]
enabled=1
[ceph-iscsi-noarch]
enabled=1
[ceph-iscsi-source]
enabled=1
```
So the command to disable the repository would be:
```
yum -y --disablerepo='ceph-iscsi*' update
```

PR #555, opened 2019-08-15
fix: Adds liveness sidecar to v1.14+ helm charts

wilmardo:

Follow-up of #438 

Helm charts where split in the time the PR was open so I hasn't been added to the new helm charts.

### DIFF
--- a/deploy/rbd/kubernetes/v1.14+/helm/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
+++ b/deploy/rbd/kubernetes/v1.14+/helm/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
@@ -108,6 +108,28 @@ spec:
               mountPath: /tmp/csi/keys
           resources:
 {{ toYaml .Values.nodeplugin.plugin.resources | indent 12 }}
+        - name: liveness-prometheus
+          image: "{{ .Values.nodeplugin.plugin.image.repository }}:{{ .Values.nodeplugin.plugin.image.tag }}"
+          args:
+            - "--type=liveness"
+            - "--endpoint=$(CSI_ENDPOINT)"
+            - "--livenessport=8081"
+            - "--livenesspath=/metrics"
+            - "--polltime=60s"
+            - "--timeout=3s"
+          imagePullPolicy: {{ .Values.nodeplugin.plugin.image.pullPolicy }}
+          env:
+            - name: CSI_ENDPOINT
+              value: "unix:/{{ .Values.socketDir }}/{{ .Values.socketFile }}"
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+          volumeMounts:
+            - name: plugin-dir
+              mountPath: {{ .Values.socketDir }}
+          resources:
+{{ toYaml .Values.nodeplugin.plugin.resources | indent 12 }}
       volumes:
         - name: socket-dir
           hostPath:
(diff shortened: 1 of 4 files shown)


deploy/cephfs/kubernetes/v1.14+/helm/templates/nodeplugin-daemonset.yaml
```diff
@@ -108,6 +108,28 @@ spec:
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
           resources:
+{{ toYaml .Values.nodeplugin.plugin.resources | indent 12 }}
```
Madhu-1: can you update provisioner helm chat for the same? for both rbd and cephfs

wilmardo: Good catch! Missed it being added there :)

humblec:
@wilmardo please revisit :)

wilmardo:
@humblec Rebase I guess? Will do in a couple minutes :)

wilmardo:
@humblec Rebased against latest master

deploy/rbd/kubernetes/v1.14+/helm/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
```diff
@@ -107,6 +107,28 @@ spec:
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
           resources:
+{{ toYaml .Values.nodeplugin.plugin.resources | indent 12 }}
+        - name: liveness-prometheus
+          image: "{{ .Values.nodeplugin.plugin.image.repository }}:{{ .Values.nodeplugin.plugin.image.tag }}"
+          args:
+            - "--type=liveness"
+            - "--endpoint=$(CSI_ENDPOINT)"
+            - "--livenessport=8081"
```
humblec: @Madhu-1  What about making this `port` a configurable value . May be in another PR?

Madhu-1: yes, this can be added to Value.yaml @wilmardo isn't it?

 also, we can add all ceph-csi and sidecar configuration parameters to value.yaml  @wilmardo sounds like a good idea?

wilmardo: Yeah had this as an improvement but decided to keep this PR simple, just to sync 1.14+ with the 1.13 charts.

Will follow up this PR with those values split into the values.yaml of all the charts :)

Madhu-1: @wilmardo Thanks :)

kfox1111:
Question. whats the point of 1.14 vs 1.13 charts? helm can template out things differently if need be for different versions.

kfox1111:
The changes look minimal but related to having snapshotting enabled. You can use .Capabilities.KubeVersion.Minor in the helm chart to know what version you are running in. So could wrap all the snapshotting in:
{{- if ge .Capabilities.KubeVersion.Minor 14 }}
Snapshot specific things.
{{- end }}
And then get back to only needing one chart.

wilmardo:
@kfox1111 Thanks! I will look into this and if it is indeed so simple I might join the helm charts!
Thanks for all the pointers 👍

Issue #520, opened 2019-08-02
Detect `rbd map` timeout errors and detect if image is mapped, to undo the mapping


ShyamsundarR:
It is noted that there are cases when,
- `rbd map` invocation times out #507 
- It leaves behind a mapped image to a device (because mapping actually completed, but state/success of the same was missed out)

This causes future attempts to map the image on other nodes, or attempts to delete the image, to error out stating image is in use.

This issue is being opened to add some safety checks and workarounds in the code, that will detect such timeout errors, and check to see if the image is mapped to a device, and undo the same.

NOTE: as we do not know the device path when rbd map returns a timeout failure, it may need the current PR #510 work to complete to leverage the detection of device using `rbd device list` instead of current schemes.

Madhu-1:
I think we can call unmap with image name also, why to wait for #510?

idryomov:
Yes, "rbd unmap" supports unmapping by image name.  No need to list mapped images just for that.

ShyamsundarR:
Ack! was not aware of that. We should use imagespec instead of hunting for the devicePath. Also, I assume this is true when `--device-type` is nbd?

dillaman:
> Ack! was not aware of that. We should use imagespec instead of hunting for the devicePath. Also, I assume this is true when `--device-type` is nbd?

Yes, with the caveat for both where it doesn't handle different clusters.

idryomov:
@ShyamsundarR Yes, it should be true.

I'm not sure I understand the need to unmap the device in this case.  If "rbd map" times out because udev decided to take a break but the device is there (i.e. you listed mapped images and got the device node), would pretending that nothing happened not work?

If we keep encountering issues with udev, we could probably adopt this course of action for the general case as well -- make "rbd map" wait for udev for a much shorter period of time and then instead of immediately bailing with a timeout error, try to get the device node by listing the devices.

idryomov:
> Yes, with the caveat for both where it doesn't handle different clusters.

This is another thing that I don't fully understand.  I thought we controlled image names in CSI?  In https://github.com/ceph/ceph-csi/pull/510#issuecomment-516074646 @ShyamsundarR mentioned that we indeed do use UUIDs there, but the discussion continued with various workarounds...

ShyamsundarR:
> @ShyamsundarR Yes, it should be true.
> 
[...]

PR #561, opened 2019-08-19
unmap rbd image if connection timeout.

Madhu-1:
Sometime rbd images are mapped even if the connection timeout error occurs, this will
try to unmap if the received error message is connection timeout. This will fix stale maps
and rbd image deletion issue

### DIFF
--- a/pkg/rbd/rbd_attach.go
+++ b/pkg/rbd/rbd_attach.go
@@ -44,6 +44,7 @@ const (
 	// NOTE: When using devicePath instead of imageSpec, the error strings are different
 	rbdUnmapCmdkRbdMissingMap = "rbd: %s: not a mapped image or snapshot"
 	rbdUnmapCmdNbdMissingMap  = "rbd-nbd: %s is not mapped"
+	rbdMapConnectionTimeout   = "Connection timed out"
 )
 
 var hasNBD = false
@@ -193,6 +194,7 @@ func attachRBDImage(volOptions *rbdVolume, cr *util.Credentials) (string, error)
 }
 
 func createPath(volOpt *rbdVolume, cr *util.Credentials) (string, error) {
+	isNbd := false
 	image := volOpt.RbdImageName
 	imagePath := fmt.Sprintf("%s/%s", volOpt.Pool, image)
 
@@ -209,6 +211,7 @@ func createPath(volOpt *rbdVolume, cr *util.Credentials) (string, error) {
 	// Choose access protocol
 	accessType := accessTypeKRbd
 	if volOpt.Mounter == rbdTonbd && hasNBD {
+		isNbd = true
 		accessType = accessTypeNbd
 	}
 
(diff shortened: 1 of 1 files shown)


humblec:
@ShyamsundarR @poornimag can you please review this in priority ?

Madhu-1:
Fixes: https://github.com/ceph/ceph-csi/issues/520

pkg/rbd/rbd_attach.go
```diff
@@ -193,6 +194,7 @@ func attachRBDImage(volOptions *rbdVolume, cr *util.Credentials) (string, error)
 }
 
 func createPath(volOpt *rbdVolume, cr *util.Credentials) (string, error) {
+	var isNbd bool
```
nixpanic: I prefer explicit assigning of variables, like `isNbd := false` in this case

pkg/rbd/rbd_attach.go
```diff
@@ -218,6 +221,13 @@ func createPath(volOpt *rbdVolume, cr *util.Credentials) (string, error) {
 	// Execute map
 	output, err := execCommand(rbd, mapOptions)
 	if err != nil {
+		// unmap rbd image if connection timeout
+		if strings.Contains(err.Error(), rbdMapConnectionTimeout) {
+			err = detachRBDImageOrDeviceSpec(imagePath, true, isNbd)
```
nixpanic: You're overwriting the previous return value of `err` here. The `klog.Warningf()` on line 231 does not print the right error anymore. We need to log the original error too.

humblec: @nixpanic if its a timeout, thats the error message or error, Isnt it ?

Madhu-1: yes, nice catch!

pkg/rbd/rbd_attach.go
Madhu-1: done.

Madhu-1:
@nixpanic  thanks for the review. PTAL

pkg/rbd/rbd_attach.go
humblec: @nixpanic The default of `bool` is `false`.  iic, the explicit assigning is not the recommended practice if its on default value..

pkg/rbd/rbd_attach.go
humblec: ah.. we are using the `err` later, so its good to be on different var. thanks.

pkg/rbd/rbd_attach.go
nixpanic: Sure, a bool defaults to false, but having it stated makes the code more readable (at least to me). The main objection was not for this bool, but for the not logging of the original error.

pkg/rbd/rbd_attach.go
nixpanic: Using a different variable is ok, or you could have moved the log from line 231 further up. In case mapping fails, and unmapping too, there will be a log about unmapping first, and the error about mapping 2nd. That is just cosmetics though, so not much of an issue

Issue #371, opened 2019-05-09
ceph-csi metrics 


Daniel-Pivonka:
Ceph-csi should collect metrics about grpc response times. These metrics could be useful for setting timeout values that could prevent creating endless loops (#284). We will also be able to see how the system response time is affected by a large amount of requests.

In addition to the response time this will allow us to build in other metrics in the future.

Here are a few examples of other metrics we could collect:
type of requests coming in 
counts of certain types of request
size of volumes requests
amount of snapshots

Through the use of a UnaryInterceptor already in place grpcs requests can be intercepted and collected to made available to a metrics collection system such as Prometheus or dump to logs.

humblec:
@Madhu-1 @ShyamsundarR @phlogistonjohn PTAL

Madhu-1:
i will work on this
/assign @Madhu-1

Madhu-1:
@Daniel-Pivonka are you working on this PR?

Daniel-Pivonka:
@Madhu-1  yes ive started some work on this already

Madhu-1:
@Daniel-Pivonka Thanks, assigning this to you as you are already working on this

ShyamsundarR:
@Daniel-Pivonka as investigated we should possibly use https://github.com/grpc-ecosystem/go-grpc-prometheus instead of writing our own interceptor, as the other one seems richer in what it provides. Other intentions of how to integrate this with ceph-csi remains the same.

humblec:
@Daniel-Pivonka I believe this is taken care now. Closing for the same reason. :+1:

Madhu-1:
@humblec this is not done yet reopening

Madhu-1:
having  https://github.com/grpc-ecosystem/go-grpc-prometheus interceptor seems heavy and also provides a lot of information which is not relevant,  it also doesn't provide info like total time spent on a particular request.  I would like to see below information in metrics

Request type, request unique ID, time spent on request, the total number of request received, the total number of success, failed requests, etc 

if we implement our own interceptor we can add if we need extra information later also. 

@ShyamsundarR @humblec I hope with this information we can check the time consumed per request in larger eco-system.

Madhu-1:
am still thinking do we need to implement metrics at all, I would suggest using some shell script to fetch the logs, and grep for the request ID. one thing we need to do is to log the time taken for the request.
this will avoid the running metrics server in cephcsi

ShyamsundarR:
[...]

PR #566, opened 2019-08-21
Implement grpc metrics for ceph-csi

Madhu-1:

Add grpc metrics for cephfs and rbd plugin

## Related issues ##

Fixes: #371

## Future concerns ##


Madhu-1:
@humblec @ShyamsundarR @Daniel-Pivonka PTAL

Madhu-1:
added DNM just to avoid auto merging, will remove DNM once I prepare the same PR for rook also. as am renaming few flags this can break the ceph-csi deployment in rook.

pkg/csi-common/server.go
```diff
@@ -109,9 +112,11 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, c
 	if ns != nil {
 		csi.RegisterNodeServer(server, ns)
 	}
-
 	klog.Infof("Listening for connections on address: %#v", listener.Addr())
-
+	if metrics {
+		grpc_prometheus.EnableHandlingTimeHistogram()
```
ShyamsundarR: I would like to see if we can enable a different histogram than the default. The default puts all requests that are beyond 10s into one bucket, I think we can be better served if we shift the buckets upto 30 or 60 seconds.

Also, if we can do the above, we can have lower count of buckets than the default of 12 (if memory serves right). Thus reducing number of metric points that are collected and stored across the plugin.

Buckets like, 0.5, 1.0, 2.5, 5.0, 10.0, 25.0, 60.0

Of course at some point this may need to be configurable, but I would think we can achieve that in a later PR when the need arises.

Reference:
[1] Exponential buckets: https://godoc.org/github.com/prometheus/client_golang/prometheus#ExponentialBuckets
[2] Ability to configure buckets in gRPC prom integration (as far as I understand it): https://github.com/grpc-ecosystem/go-grpc-prometheus/blob/ae0d8660c5f2108ca70a3776dbe0fb53cf79f1da/server_metrics.go#L60-L64

deploy/cephfs/kubernetes/v1.13/csi-cephfsplugin-provisioner.yaml
```diff
@@ -74,7 +78,14 @@ spec:
             - "--drivername=cephfs.csi.ceph.com"
             - "--metadatastorage=k8s_configmap"
             - "--pidlimit=-1"
+            - "--metricsport=8091"
+            - "--metricspath=/metrics"
+            - "--enablegrpcmetrics=false"
```
humblec: I like the flag for enabling :+1:

deploy/cephfs/kubernetes/v1.13/helm/ceph-csi-cephfs/templates/nodeplugin-service.yaml
```diff
@@ -0,0 +1,25 @@
...
+    release: {{ .Release.Name }}
+  ports:
+    - name: http-metrics
+      port: 8080
+      protocol: TCP
+      targetPort: 8081
+    - name: grpc-metrics
+      port: 8090
+      protocol: TCP
+      targetPort: 8091
```
humblec: Just a suggestion, we have these ports hardcoded, may be we can get it from `.Values`

pkg/util/util.go
```diff
@@ -53,30 +53,25 @@ var (
...
-	// cephfs related flags
-	VolumeMounter string // default volume mounter (possible options are 'kernel', 'fuse')
-	MountCacheDir string // mount info cache save dir
-
-	// livenes related flags
-	LivenessPort int           // TCP port for liveness requests"
-	LivenessPath string        // path of prometheus endpoint where metrics will be available
-	PollTime     time.Duration // time interval in seconds between each poll
-	PoolTimeout  time.Duration // probe timeout in seconds
+	Vtype              string        // driver type [rbd|cephfs|liveness]
```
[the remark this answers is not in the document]
Madhu-1: rearranging the struct fields is necessary as it reduces memory consumption.

deploy/cephfs/kubernetes/v1.13/helm/ceph-csi-cephfs/templates/nodeplugin-service.yaml
Madhu-1: yes, @humblec having plan to move all configuration to Value.yaml as a follow-up PR.

pkg/csi-common/server.go
Madhu-1: done

pkg/util/util.go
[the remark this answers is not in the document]
ShyamsundarR: I understand (or guessed the reason). My request is not to go back to the old order, but add to the comments if an configuration variable is only meant for a particular mode of the plugin (as described above).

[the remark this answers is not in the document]
Madhu-1: done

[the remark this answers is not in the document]
ShyamsundarR: Thanks.

pkg/csi-common/server.go
ShyamsundarR: I think this may come back for a different default, but let's see. The current 0.5,2,6 will make buckets like, 0.5 - 2 - 8- 32... the workload sensitivity may demand a more granular set of initial buckets than the 2 - 8 - 32 jump.

I would have preferred a non-exponential set in this lower order, say 0.5, 2.0, 5.0, 10.0, 30.0, 60.0 (even though I was walling for an exponential initially).

Leaving this thought here for the future.

Madhu-1: this is configurable, I will check how to expose this flag in rook.  having non-exponential values need code changes

humblec:
Approving this PR , however keeping `DNM` in place for holding till the rook PR is ready.

wilmardo:
Just a question (no knownledgde of prometheus just yet), how does this relate to the already present liveness (#438)? Can this metrics endpoint be used for liveness? If so what is the use of the current liveness sidecar?

deploy/cephfs/kubernetes/v1.13/helm/ceph-csi-cephfs/templates/nodeplugin-service.yaml
wilmardo: I have moved the ports for the current liveness sidecar to values.yaml in #570 
Wouldn't be to hard to rebase and implement this for you :)

deploy/cephfs/kubernetes/v1.13/csi-cephfsplugin-provisioner.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: v1
 metadata:
   name: csi-cephfsplugin-provisioner
   labels:
-    app: csi-liveness
+    app: csi-metrics
```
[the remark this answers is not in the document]
Madhu-1: @Daniel-Pivonka Thanks for the review, updated example service monitor can you please check

Madhu-1:
removed DNM label for merging this as rook PR is approved

PR #571, opened 2019-08-26
Cephfs: Use ceph kernel client if kernel version >= 4.17

poornimag:
Ceph kernel client is more performant than ceph fuse client.
The kernel client has Quota support only in the kernel version >=4.17.
Hence use ceph kernel client when the kernel version is >=4.17

### DIFF
--- a/pkg/cephfs/driver.go
+++ b/pkg/cephfs/driver.go
@@ -50,9 +50,6 @@ type Driver struct {
 }
 
 var (
-	// DefaultVolumeMounter for mounting volumes
-	DefaultVolumeMounter string
-
 	// CSIInstanceID is the instance ID that is unique to an instance of CSI, used when sharing
 	// ceph clusters across CSI instances, to differentiate omap names per CSI instance
 	CSIInstanceID = "default"
@@ -103,17 +100,8 @@ func (fs *Driver) Run(conf *util.Config, cachePersister util.CachePersister) {
 	if conf.VolumeMounter != "" {
 		if err := validateMounter(conf.VolumeMounter); err != nil {
 			klog.Fatalln(err)
-		} else {
-			DefaultVolumeMounter = conf.VolumeMounter
 		}
-	} else {
-		// Pick the first available mounter as the default one.
-		// The choice is biased towards "fuse" in case both
-		// ceph fuse and kernel mounters are available.
-		DefaultVolumeMounter = availableMounters[0]
 	}
-
-	klog.Infof("cephfs: setting default volume mounter to %s", DefaultVolumeMounter)
 
 	if err := util.WriteCephConfig(); err != nil {
 		klog.Fatalf("failed to write ceph configuration file: %v", err)
--- a/pkg/cephfs/volumemounter.go
+++ b/pkg/cephfs/volumemounter.go
@@ -55,12 +55,32 @@ func loadAvailableMounters() error {
 	// #nosec
 	kernelMounterProbe := exec.Command("mount.ceph")
 
-	if fuseMounterProbe.Run() == nil {
-		availableMounters = append(availableMounters, volumeMounterFuse)
+	err := kernelMounterProbe.Run()
+	if err == nil {
+		kernelVersion, _, err := execCommand(context.TODO(), "uname", "-r")
+		if err != nil {
+			return err
+		}
+		vers := strings.Split(string(kernelVersion), ".")
+		majorVers, err := strconv.Atoi(vers[0])
+		if err != nil {
+			return err
+		}
+		minorVers, err := strconv.Atoi(vers[1])
+		if err != nil {
+			return err
+		}
+		if majorVers >= 4 && minorVers >= 17 {
+			klog.Infof("loaded mounter: %s", volumeMounterKernel)
+			availableMounters = append(availableMounters, volumeMounterKernel)
+		} else {
+			klog.Infof("kernel version < 4.17 might not support quota feature, hence not loading kernel client")
+		}
 	}
 
-	if kernelMounterProbe.Run() == nil {
-		availableMounters = append(availableMounters, volumeMounterKernel)
+	if fuseMounterProbe.Run() == nil {
+		klog.Infof("loaded mounter: %s", volumeMounterFuse)
+		availableMounters = append(availableMounters, volumeMounterFuse)
 	}
 
 	if len(availableMounters) == 0 {
@@ -80,25 +100,21 @@ func newMounter(volOptions *volumeOptions) (volumeMounter, error) {
 
 	wantMounter := volOptions.Mounter
 
-	if wantMounter == "" {
-		wantMounter = DefaultVolumeMounter
-	}
-
 	// Verify that it's available
 
 	var chosenMounter string
 
 	for _, availMounter := range availableMounters {
-		if chosenMounter == "" {
-			if availMounter == wantMounter {
-				chosenMounter = wantMounter
-			}
+		if availMounter == wantMounter {
+			chosenMounter = wantMounter
+			break
 		}
 	}
 
 	if chosenMounter == "" {
 		// Otherwise pick whatever is left
 		chosenMounter = availableMounters[0]
+		klog.Infof("requested mounter: %s, chosen mounter: %s", wantMounter, chosenMounter)
 	}
 
 	// Create the mounter


pkg/cephfs/driver.go
```diff
@@ -108,9 +108,9 @@ func (fs *Driver) Run(conf *util.Config, cachePersister util.CachePersister) {
 		}
 	} else {
 		// Pick the first available mounter as the default one.
```
ajarr: remove this sentence

ajarr: Can this be simplifed as, 'pick kernel mounter and kernel version >= 4.17 if available, else pick fuse mounter'?

pkg/cephfs/volumemounter.go
```diff
@@ -74,6 +74,31 @@ type volumeMounter interface {
 	name() string
 }
 
+func getDefaultMounter() string {
+	mounter := volumeMounterFuse
```
ajarr: So what if FuseMounter wasn't available? Instead may be you can check kernel version in loadAvailableMounters, and make add kernel mounter as available only when kernel version >=4.17?

humblec:
@poornimag is this PR still in WIP ?  or you can remove it from the title.

pkg/cephfs/driver.go
poornimag: Done

pkg/cephfs/volumemounter.go
poornimag: The newMounter takes care of checking, if the defauls/the requested mounter is available, if we check in load available, it will be duplicate check in newMounter too..

poornimag: Done

joscollin: A quick look. Wondering why `mounter` is needed here, if it is not changing.

ajarr: > The newMounter takes care of checking, if the defauls/the requested mounter is available, if we check in load available, it will be duplicate check in newMounter too..

So you're OK with kernel mounter being in the slice `availableMounters` even though the kernel version is less than 4.17?

I was thinking that you can get rid of this function `getDefaultMounter()`.

In `loadAvailableMounter()`, if kernel version >=4.17, let the kernel mounter be the first element in the splice  availableMounters`, else don't have kernel mounter in the splice `availableMounters` and log a warning that kernel is not an availableMounter since version kernel version < 4.17.

In `driver.go`'s `Run()`, you can leave `DefaultVolumeMounter = availableMounter[0]` as was before.

And when storage admin chooses kernel mounter via the storage class option and the setup has a kernel less than 4.17,  raise an error that the kernel version >= 4.17 is required for Ceph CSI.

pkg/cephfs/volumemounter.go
```diff
@@ -88,10 +115,9 @@ func newMounter(volOptions *volumeOptions) (volumeMounter, error) {
 	var chosenMounter string
 
 	for _, availMounter := range availableMounters {
-		if chosenMounter == "" {
-			if availMounter == wantMounter {
-				chosenMounter = wantMounter
-			}
+		if availMounter == wantMounter {
```
ajarr: if kernel version < 4.17 and volOptions.Mounter == kernelMounter, the chosen mounter would be set as kernelMounter, since kernelmounter is an available mounter, right? We don't want this.

poornimag: I thought you said that its required? If user doesn't need Quota feature, they could still use kernel mounter?

poornimag: But then Quota is always set on the directories when PVC gets created. So we rather not use kernel mount irrespective of what user specifies if the kernel version is <= 4.17?

ajarr: "Quotas are cooperative and non-adversarial. CephFS quotas rely on the cooperation of the client who is mounting the file system to stop writers when a limit is reached. A modified or adversarial client cannot be prevented from writing as much data as it needs. Quotas should not be relied on to prevent filling the system in environments where the clients are fully untrusted."

"Quotas are implemented in the kernel client 4.17 and higher. Quotas are supported by the userspace client (libcephfs, ceph-fuse). Linux kernel clients >= 4.17 support CephFS quotas but only on mimic+ clusters. Kernel clients (even recent versions) will fail to handle quotas on older clusters, even if they may be able to set the quotas extended attributes."

Source, https://docs.ceph.com/docs/master/cephfs/quota/#limitations

ajarr: > I thought you said that its required? If user doesn't need Quota feature, they could still use kernel mounter?

Sorry, I don't follow. What did I say was required? 

In CSI is there a concept of volumes with infinite capacity? In that case kernel client don't need to be version >= 4.17.

ajarr: > But then Quota is always set on the directories when PVC gets created. So we rather not use kernel mount irrespective of what user specifies if the kernel version is <= 4.17?

If kernel client version less than 4.17, and quota xattr is set on the subvolume. the kernel client wouldn't respect the quota limit set on the subvolume.

pkg/cephfs/volumemounter.go
```diff
@@ -54,12 +54,27 @@ func loadAvailableMounters() error {
...
+		vers := strings.Split(string(kernelVersion), ".")
+		majorVers, err := strconv.Atoi(vers[0])
+		if err != nil {
+			return err
+		}
+		minorVers, err := strconv.Atoi(vers[1])
+		if err != nil {
+			return err
+		}
+		if majorVers >= 4 && minorVers >= 17 {
```
ajarr: else, log a warning that kernel client < 4.17 can't be used as a mounter as it doesn't support quotas?

jtlayton:
Keying off of kernel versions may be problematic. We often backport new features to older kernels (RHEL being a case in point here). That said, I'm not sure what we can do in lieu of that.

You may want to consider provide a mechanism for the admin to force usage of the kernel mounts even when the version seems to low, in order to handle these cases. Is that already possible?

ajarr:
> Keying off of kernel versions may be problematic. We often backport new features to older kernels (RHEL being a case in point here). That said, I'm not sure what we can do in lieu of that.
> 

Good point.

> You may want to consider providing a mechanism for the admin to force usage of the kernel mounts even when the version seems too low, in order to handle these cases. Is that already possible?

It's not possible with this change. 

We'd need to introduce an extra option to override the kernel version check. @poornimag @ShyamsundarR what do you think? Or, should we just stick to @poornimag's  earlier patch set, where if someone explicitly chooses the kernel mounter via the 'mounter' setting, we don't do a kernel version check?

ShyamsundarR:
> > Keying off of kernel versions may be problematic. We often backport new features to older kernels (RHEL being a case in point here). That said, I'm not sure what we can do in lieu of that.
> 
> Good point.
> 
> > You may want to consider providing a mechanism for the admin to force usage of the kernel mounts even when the version seems too low, in order to handle these cases. Is that already possible?
> 
> It's not possible with this change.
> 
> We'd need to introduce an extra option to override the kernel version check. @poornimag @ShyamsundarR what do you think? Or, should we just stick to @poornimag's earlier patch set, where if someone explicitly chooses the kernel mounter via the 'mounter' setting, we don't do a kernel version check?

No, that would break directory quota and allow users on older/unsupported kernel deployments to abuse the limits if it defaults to the kernel mounter.

Better to aim for correctness rather than providing the convenience that an older kernel in the near/far future would support a required feature. IMO, We can change the supported kernel versions in a subsequent patch to the image, instead.

poornimag:
Even if we introduce another option, users can still abuse the limits, we rather stick to a known version where it is definitely available, or a command that can tell if the required features are available.

poornimag:
Even if we introduce another option, users can still abuse the limits, we rather stick to a known version where it is definitely available, or a command that can tell if the required features are available.

humblec:
>Prior to this commit the plugin args could specify a default, which is no longer the case as we are not using DefaultVolumeMounter as noted in the review. We would need to cleanup the option from the possible set of options and update the doc etc. to reflect the same. This could be a separate PR if required.
>


Yeah, this can be a followup PR . @poornimag please keep track of it .. 
LGTM. Approved.

humblec:
@Madhu-1 PTAL.

ShyamsundarR:
> > Prior to this commit the plugin args could specify a default, which is no longer the case as we are not using DefaultVolumeMounter as noted in the review. We would need to cleanup the option from the possible set of options and update the doc etc. to reflect the same. This could be a separate PR if required.
> 
> Yeah, this can be a followup PR . @poornimag please keep track of it ..
> LGTM. Approved.

Added an issue to track the same, https://github.com/ceph/ceph-csi/issues/584

pkg/cephfs/volumemounter.go
Madhu-1: @poornimag please add a log message if kernel client is less than expected version

Madhu-1:
@poornimag  please rebase with latest changes

Madhu-1:
@ShyamsundarR @ajarr PTAL

ShyamsundarR:
I'm good with the changes, @ajarr has a comment open, need to understand if the current state of the code is in alignment with his concerns. @ajarr?

ShyamsundarR:
> @poornimag @Madhu-1 I'm not sure why https://github.com/ceph/ceph-csi/pull/571/files#r319136430 couldn't be addressed. I still think we need this. I'm OK with other code changes. I'll leave it up to your discretion to merge this.

Agreed, @Madhu-1 also had the same concern, let's get the logs in as it would help address any user queries or concerns when put to use. @poornimag other changes look good, request you to add the log message as well. Thanks.

humblec:
@poornimag can you please address  https://github.com/ceph/ceph-csi/pull/571/files#r319136430, we are good to go once its in. thanks !

poornimag:
> @poornimag @Madhu-1 I'm not sure why https://github.com/ceph/ceph-csi/pull/571/files#r319136430 couldn't be addressed. I still think we need this. I'm OK with other code changes. I'll leave it up to your discretion to merge this.

@ajarr  IMO there is a log message in the if condition, log message in the else is redundant.

Issue #493, opened 2019-07-24
Use ceph image with a more specific tag that includes the date or commit ID


ajarr:
Currently, ceph csi image is built using ceph image with tag, v14.2. We can use a ceph image with tag that is more specific, e.g., with the date the image was created, v14.2.2-20190722, or with the HEAD commit ID. Rook does this currently.

Madhu-1:
@ajarr how this is helpful for the end user?

ajarr:
I'm not sure who you're referring to as end user?

IMO, this is good practice for the person who builds the CSI image that he uses a image with a more detailed tag. Unless otherwise you want the base Ceph image to always be the latest 14.2.x ? What if there was a bug in latest 14.2.x stable that affects Ceph CSI?

Madhu-1:
the commit tag added to the image in this PR https://github.com/ceph/ceph-csi/pull/469. from this, we can check from which tag the image is built.

ajarr:
PR #469 helps check the tag of Ceph-CSI release. I'm talking about the base Ceph image.

Madhu-1:
@ajarr if you do docket inspect on the image you will see ceph version 

```
docker inspect quay.io/cephcsi/cephcsi:canary
[
    {
        "Id": "sha256:afb0079dd922597a926b8692af128455e1d33bf0c41244def8ce59f3636be7f0",
        "RepoTags": [
            "quay.io/cephcsi/cephcsi:canary"
        ],
        "RepoDigests": [
            "madhupr001/cephcsi@sha256:c16e6807a91102645840afbcf2311126610bb72683c8f79460d9fe8019a79e1a"
        ],
        "Parent": "sha256:91bd0b5601734afff652662f400e7176f0a106c6c29a22d727994abb4341bf0e",
        "Comment": "",
        "Created": "2019-07-24T11:35:45.79180354Z",
        "Container": "5b5c2db9e9b7de7aecb03da8a300b7aa02f20578123e0c032249c78919d40f03",
        "ContainerConfig": {
            "Hostname": "5b5c2db9e9b7",
            "Domainname": "",
            "User": "",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": [
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "CEPH_VERSION=nautilus",
[...]

PR #573, opened 2019-08-26
Use latest base image for CSI provisioner image

humblec:
Fixes: #493

### DIFF
--- a/deploy/cephcsi/image/Dockerfile
+++ b/deploy/cephcsi/image/Dockerfile
@@ -1,11 +1,10 @@
-FROM ceph/ceph:v14.2
+FROM ceph/ceph:v14.2.2-20190826
 LABEL maintainers="Ceph-CSI Authors"
 LABEL description="Ceph-CSI Plugin"
 
-# Removing ceph-iscsi repo to workaround the repo issue while upgrading
-RUN rm -f /etc/yum.repos.d/ceph-iscsi.repo && yum -y update && yum clean all
 ENV CSIBIN=/usr/local/bin/cephcsi
 
+RUN yum -y update && yum clean all
 COPY cephcsi $CSIBIN
 
 RUN chmod +x $CSIBIN


humblec:
The build is passing locally, but failing in CI , specifically the base  image pull @Madhu-1 any thoughts?

ajarr:
Fixes:  https://github.com/ceph/ceph-csi/issues/493
in the commit message

humblec:
> Fixes: #493
> in the commit message

Done.. thanks!

humblec:
> FYI -- v14.2.3 is coming very shortly (this week?)

Yeah. heard so. Will update once its available. !

deploy/cephcsi/image/Dockerfile
```diff
@@ -1,9 +1,7 @@
-FROM ceph/ceph:v14.2
+FROM ceph/ceph:v14.2.2-20190826
 LABEL maintainers="Ceph-CSI Authors"
 LABEL description="Ceph-CSI Plugin"
 
-# Removing ceph-iscsi repo to workaround the repo issue while upgrading
-RUN rm -f /etc/yum.repos.d/ceph-iscsi.repo && yum -y update && yum clean all
```
Madhu-1: @humblec any reason why ` yum update` is removed?

humblec: @Madhu-1 I think we can keep it, updated the patch set.

nixpanic:
This is better than using an old 14.2 image, but not by much. Is there no way to consume the latest 14.2 image? Something like 14.2.latest or so? Maybe we can request such a tag if it does not exist yet.

We should assume the images that Ceph releases are more stable and more bug free than the previous release. I think it is better to consume the most recent releases, than knowing we consume an old build which has more known bugs (why would the image get updated otherwise?).

humblec:
> This is better than using an old 14.2 image, but not by much. Is there no way to consume the latest 14.2 image? Something like 14.2.latest or so? Maybe we can request such a tag if it does not exist yet.>

@nixpanic  Agree. This is the latest change made in rook https://github.com/rook/rook/pull/3694 , so I believe this is the most latest image or IOW, this is the counterpart of Rook fix.

humblec:
@ShyamsundarR @poornimag PTAL

nixpanic:
On Tue, Aug 27, 2019 at 03:42:43AM -0700, Humble Devassy Chirammal wrote:
> > This is better than using an old 14.2 image, but not by much. Is there no way to consume the latest 14.2 image? Something like 14.2.latest or so? Maybe we can request such a tag if it does not exist yet.>
> 
> @nixpanic  Agree. This is the latest change made in rook https://github.com/rook/rook/pull/3694 , so I believe this is the most latest image or IOW, this is the counterpart of Rook fix.

For deployments (like done by Rook) it is recommended to stick to a
specific image version. However, these images can consume the latest
v14.2 releases without problems. The deployment tool (Rook) should be
configured to deploy Ceph-CSI with specific images instead.

At the moment Rook uses cephcsi:canary as the reference, it would be
better if Rook can select an image by SHA/digest instead.

The problem (which I know of) that needs to be prevented, is that an
image gets an update and while restarting a pod on a server a new
updated image gets pulled. The new image may have an updated Ceph
version, which could be incompatible with the image used for OSDs, MONs
etc.

Maybe there is more to this, and I am missing something. Further details
to make me understand why this PR is a good thing would be welcome.

humblec:
@nixpanic I fail to understand what you are conveying in previous comment. :) Whatever specific to `Rook` in the comment can be discussed in Rook repo for sure.  On a specific change to this Dockerfile, the ceph cluster or the version or ..etc does not make any difference, the reason being the image is treated as just the BASE image to create a container , nothing else.

humblec:
>Further, as discussed in #493 the intention was to be very prescriptive in our image to avoid any problems due to breaking changes in the base image (which was v14.2). If the intention of this commit is to address the same, then I am fine with it.
>

Correct, the intention is to use more specific image version as mentioned in #493 .

nixpanic:
On Wed, Aug 28, 2019 at 04:25:08AM -0700, ShyamsundarR wrote:
> @nixpanic this is not better than using an older v14.2 image, as the
> v14.2 image is always updated to the latest in the v14.2 line. So when
> v14.2.X is built, v14.2 is also updated. IOW, v14.2 is the "latest"
> that you are looking for.

In that case, I really prefer to use the base image referenced by the
v14.2 tag. In order to update the ceohcsi:canary image with the latest
bug fixes from the Ceph base image, only a rebuild is needed.

I do not see an advantage to pin the base image to a very specific
release in this repository. Any productized Ceph-CSI image would likely
have a base image pinned to a specific product Ceph image. This
repository however, is upstream, and we should rely on the quality that
the Ceph team deems sufficient for tagging an updated v14.2 image.

mykaul:
Why do we need to rely on Ceph base image and not minimal UBI8 for example and install additional packages that we might need? I would assume it'll have smaller footprint and less potential security issues to worry about.

nixpanic:
At the moment UBI8 can not be used yet, for all I know neither Ceph or the CentOS Storage SIG provide el8 builds of the packages.

There is a UBI7 available from `registry.access.redhat.com/ubi7:latest` too, so maybe that could be used (which Ceph build?). Or, use `centos:7` and install the Ceph packages from the CentOS Storage SIG.

I tried the last option, but there seem to be some missing dependencies (just reported in #centos-devel IRC to @gfidente). So for now using `ceph:v14.2` as base image would be best.

gfidente:
thanks for pinging, I just tagged the missing deps into the repos (which should be available in the testing repos already) but for the -release repos this might take a bit longer, especially given [1]

asked on the thread if people can help @alphacc

1. https://lists.centos.org/pipermail/centos-devel/2019-September/017650.html

nixpanic:
Using `centos:7` as base and installing

1. `ceph-base` the resulting image becomes 461 MB
1. `ceph` the resulting image becomes 570 MB

This is without including the `cephcsi` executable.

Not sure what is used as a base for the `ceph:v14.2` image, but that is only 271 MB.

```
FROM centos:7
LABEL maintainers="Ceph-CSI Authors"
LABEL description="Ceph-CSI Plugin"

# need to enable centos-ceph-nautilus-test as the release repo misses some RPMs
RUN true \
    && yum -y install centos-release-ceph \
    && yum -y --enablerepo=centos-ceph-nautilus-test install ceph-base \
    && yum -y update \
    && yum clean all \
    && true

#ENV CSIBIN=/usr/local/bin/cephcsi
#COPY cephcsi $CSIBIN
#RUN chmod +x $CSIBIN

ENTRYPOINT ["/usr/local/bin/cephcsi"]
```

mykaul:
@nixpanic - what is https://github.com/ceph/ceph-container/tree/master/ceph-releases/ALL/ubi8 for? (came with https://github.com/ceph/ceph-container/commit/ae8dde4f25d61b934f3bbbdc2f00cb328e7b2c24 )

nixpanic:
On Tue, Sep 03, 2019 at 12:26:57AM -0700, Yaniv Kaul wrote:
> @nixpanic - what is https://github.com/ceph/ceph-container/tree/master/ceph-releases/ALL/ubi8 for? (came with https://github.com/ceph/ceph-container/commit/ae8dde4f25d61b934f3bbbdc2f00cb328e7b2c24 )

I do not think it is used yet. I was not able to find el8 RPMs on
download.ceph.com. UBI8 images are also not published on
https://hub.docker.com/r/ceph/ceph for all I can see.

mykaul:
Ping?

deploy/cephcsi/image/Dockerfile
```diff
@@ -1,11 +1,10 @@
-FROM ceph/ceph:v14.2
+FROM ceph/ceph:v14.2.2-20190826
```
nixpanic: `ceph/ceph:v14.2` is being updated as latest 14.2.x version, but might also see new builds for OS package updates. It is important to have the latest (security/bugfix) updates included in new builds. With that, I do not approve of requiring a fixed (outdated) base image.

deploy/cephcsi/image/Dockerfile
```diff
@@ -1,11 +1,10 @@
...
 LABEL maintainers="Ceph-CSI Authors"
 LABEL description="Ceph-CSI Plugin"
 
-# Removing ceph-iscsi repo to workaround the repo issue while upgrading
-RUN rm -f /etc/yum.repos.d/ceph-iscsi.repo && yum -y update && yum clean all
 ENV CSIBIN=/usr/local/bin/cephcsi
 
+RUN yum -y update && yum clean all
```
nixpanic: This should work again (fixed Ceph repositories), so it is good to have.

nixpanic:
#493 has been closed as `ceph/ceph:v14.2` should have stable updates only and we want to take advantage of those.

deploy/cephcsi/image/Dockerfile
ShyamsundarR: (clarification required) So when we build the CSI image, stating the FROM tag, does a download of the CSI image to a running cluster, pick the latest base image (IOW the FROM tag), or is the base FROM image baked into the CSI image and needs a rebuild of CSI image to catch up to the latest?

If the former (auto updated), I can see how latest 14.2 can always be picked up by installations running the CSI image, assuming updated to base image is somehow tracked and understood by the system.

If the latter (CSI image needs to be rebuilt), then we need to rebuild CSI image each time the base image is updated for (almost) any reason.

So, depending on 14.2 image version is fine, but I suspect we need to rebuild CSI images whenever 14.2 changes to keep our image contents updated.

This is a request for clarification as I am not well versed in how this is processed when we build our images. Based on the answer, we either need a rebuild "policy" or depend on a specific version and move it higher when we rebuild.

nixpanic:
On Tue, Nov 19, 2019 at 07:53:07AM -0800, ShyamsundarR wrote:
> ShyamsundarR commented on this pull request.
> 
> 
> 
> > @@ -1,11 +1,10 @@
> -FROM ceph/ceph:v14.2
> +FROM ceph/ceph:v14.2.2-20190826
> 
> (clarification required) So when we build the CSI image, stating the FROM tag, does a download of the CSI image to a running cluster, pick the latest base image (IOW the FROM tag), or is the base FROM image baked into the CSI image and needs a rebuild of CSI image to catch up to the latest?

The image that is built refers to a hash for the version of the base
image. It is also possible to flatten images and not have the different
layers of images.

> If the former (auto updated), I can see how latest 14.2 can always be picked up by installations running the CSI image, assuming updated to base image is somehow tracked and understood by the system.

This is not the case.

> If the latter (CSI image needs to be rebuilt), then we need to rebuild CSI image each time the base image is updated for (almost) any reason.

That is correct. Inspection tools (Quay has them) can identify updates
to base images and inform maintainers to do a rebuild.

> So, depending on 14.2 image version is fine, but I suspect we need to rebuild CSI images whenever 14.2 changes to keep our image contents updated.

Right, that is needed.

> This is a request for clarification as I am not well versed in how this is processed when we build our images. Based on the answer, we either need a rebuild "policy" or depend on a specific version and move it higher when we rebuild.

Images that have been built and published will stick to the base Ceph
image that was used at build-time. I think Travis CI is used to build
the images and push them into Quay (see deploy.sh). Maybe we can have a
CI job (not sure if Travis allows cron-like jobs) that checks for the
layers in the published image, and rebuilds it in case there is an
update to the base image.

Some scripting should be able to do this, inspecting example:
  skopeo inspect docker://quay.io/cephcsi/cephcsi:v1.2-canary

PR #585, opened 2019-09-03
Build versioned images in autobuild

kshlm:

This make the image autobuilder, build images with unique tags and push them to the registry. The built images will also be tagged 'canary' to allow workflows using the 'canary' images to continue working.

Do you have any questions?

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,9 +56,11 @@ cephcsi:
 image-cephcsi: cephcsi
 	cp _output/cephcsi deploy/cephcsi/image/cephcsi
 	$(CONTAINER_CMD) build -t $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) deploy/cephcsi/image
+	if [ "canary" != "$(CSI_IMAGE_VERSION)" ]; then $(CONTAINER_CMD) tag $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) $(CSI_IMAGE_NAME):canary; fi
 
 push-image-cephcsi: image-cephcsi
 	$(CONTAINER_CMD) push $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION)
+	if [ "canary" != "$(CSI_IMAGE_VERSION)" ]; then $(CONTAINER_CMD) push $(CSI_IMAGE_NAME):canary; fi
 
 
 clean:
--- a/deploy.sh
+++ b/deploy.sh
@@ -28,7 +28,8 @@ if [ "${TRAVIS_BRANCH}" == 'csi-v0.3' ]; then
 	export ENV_RBD_IMAGE_VERSION='v0.3-canary'
 	export ENV_CEPHFS_IMAGE_VERSION='v0.3-canary'
 elif [ "${TRAVIS_BRANCH}" == 'master' ]; then
-	export ENV_CSI_IMAGE_VERSION='canary'
+	ENV_CSI_IMAGE_VERSION=$(git describe --tags --dirty)
+	export ENV_CSI_IMAGE_VERSION
 else
 	echo "!!! Branch ${TRAVIS_BRANCH} is not a deployable branch; exiting"
 	exit 0 # Exiting 0 so that this isn't marked as failing


Makefile
```diff
@@ -56,9 +56,11 @@ cephcsi:
 image-cephcsi: cephcsi
 	cp _output/cephcsi deploy/cephcsi/image/cephcsi
 	$(CONTAINER_CMD) build -t $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) deploy/cephcsi/image
+	if [ "canary" != "$(CSI_IMAGE_VERSION)" ]; then $(CONTAINER_CMD) tag $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) $(CSI_IMAGE_NAME):canary; fi
 
 push-image-cephcsi: image-cephcsi
 	$(CONTAINER_CMD) push $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION)
+	if [ "canary" != "$(CSI_IMAGE_VERSION)" ]; then $(CONTAINER_CMD) $(CSI_IMAGE_NAME):canary; fi
```
nixpanic: missing `push` after `CONTAINER_CMD` ?

deploy.sh
```diff
@@ -28,7 +28,7 @@ if [ "${TRAVIS_BRANCH}" == 'csi-v0.3' ]; then
 	export ENV_RBD_IMAGE_VERSION='v0.3-canary'
 	export ENV_CEPHFS_IMAGE_VERSION='v0.3-canary'
 elif [ "${TRAVIS_BRANCH}" == 'master' ]; then
-	export ENV_CSI_IMAGE_VERSION='canary'
+  export ENV_CSI_IMAGE_VERSION=$(git describe --tags --dirty)
```
wilmardo: Nitpick: Indenting is off

wilmardo:
FYI: You can run `make static-check`  to test errors like this locally:

```
In ./deploy.sh line 31:

  export ENV_CSI_IMAGE_VERSION=$(git describe --tags --dirty)

         ^-------------------^ SC2155: Declare and assign separately to avoid masking return values.
```

Makefile
```diff
@@ -56,9 +56,11 @@ cephcsi:
 image-cephcsi: cephcsi
 	cp _output/cephcsi deploy/cephcsi/image/cephcsi
 	$(CONTAINER_CMD) build -t $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) deploy/cephcsi/image
+	if [ "canary" != "$(CSI_IMAGE_VERSION)" ]; then $(CONTAINER_CMD) tag $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) $(CSI_IMAGE_NAME):canary; fi
 
 push-image-cephcsi: image-cephcsi
 	$(CONTAINER_CMD) push $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION)
+	if [ "canary" != "$(CSI_IMAGE_VERSION)" ]; then $(CONTAINER_CMD) push $(CSI_IMAGE_NAME):canary; fi
```
ShyamsundarR: Pardon my docker (or other) CLI ignorance, we will be pushing 2 images here if I am not wrong? IOW, in quay (when a PR in master is merged) we will have 2 images? This is not desiable.

The images in quay are meant to be like so,
- master
  - tag: `canary`
    - Tracks latest tip of master, it would be nice to have a docker image label like so `LABEL git_commit=$GIT_COMMIT` which is missing, and can help users reporting issues to provide the label in addition to the image version in use (which is `canary` for this branch by default)
    - For users who like living on the edge... or for development and testing as master progresses.
- Release (say 1.1.x)
  - tag: `1.1-canary`
    - Tracks tip of 1.1 release, any auto built PR is part of this
    - For users willing to stay at tip of a released branch that has not yet been officially released
  - tag: 1.1.X
    - When a specific release is made for the 1.1 branch (to include bug fixes and other changes), this is created
    - Updated only when a release is made
    - For users who want to stick to a specific image
  - tag: 1.1
    - Updated whenever a 1.1.X release is made
    - For users to stick to latest releases in 1.1 branch

Now all of the above lack the `git commit ` tag and would be helpful to add the same.

Expanding on images published from `master` for each PR is IMO undesirable at present if I understood the 2 image push changes in this commit.

ShyamsundarR: Had a discussion with @kshlm and the need is for users developing against `master` and wanting to stay at a pinned image version. Hence the need to maintain images per PR built upstream.

IMO, this causes noise in image tags maintained in quay, alternatives would be a fork and maintain separate image tags by the project depending on this one for development needs (and rebasing its forked master and updating the image, from time to time).

If essential then providing an per PR `master` image can be done.

Tagging @humblec and @Madhu-1 who own the releases for comments.

kshlm: > If essential then providing an per PR `master` image can be done.

Just curious, where would this be hosted? I you push it quay.io/ceph/ceph-csi, you just end up with this.

ShyamsundarR: I would assume quay.io/cephsi, and it is with this commit that it would be achieved.

Madhu-1: I thought of doing it as per @obnoxxx suggestion, but could not get a better approach to build (git describe) tagged images on timley bases

@kshlm  as @ShyamsundarR explained this pushes2 images, I would suggest  to build and host it in other `quay.io` account, whenever it's needed you can update the image, no need to depend on cephcsi

@ShyamsundarR @humblec thoughts?

Madhu-1:
closing this PR, please feel free to reopen if required

PR #591, opened 2019-09-05
Fix liveness sidecar from opening endless connections to the csi socket

Daniel-Pivonka:
Change the liveness side car to reuse one connection to the csi socket.

Previously the liveness sidecar repeatedly opened a new connection to to the csi socket every time a health check happened. 


Daniel-Pivonka:
@ShyamsundarR @raghavendra-talur

PR #598, opened 2019-09-06
Consider Kernel >=5.x as sufficient for using the Kernel mounter

KingJ:

#571 allows the mounter to use the kernel Ceph FS client if the kernel is version 4.17 or higher. However, it requires the kernel's major version to be greater than or equal to 4 _and_ the minor version to be greater than or equal to 17. As a result, kernel version 5.x is considered invalid - at least until 5.17 is released!

This PR adds a logical OR to the version check so that if the major version is 5 or higher, it is considered safe to use the kernel mounter regardless of the minor version.

As this is the first PR i've submitted to this project and I am still quite inexperienced with Go, a little extra scrutiny would be welcome even though the change is quite minor!

## Related issues ##

#571

## Future concerns ##

None

### DIFF
--- a/pkg/cephfs/volumemounter.go
+++ b/pkg/cephfs/volumemounter.go
@@ -70,7 +70,7 @@ func loadAvailableMounters() error {
 		if err != nil {
 			return err
 		}
-		if (majorVers >= 4 && minorVers >= 17) || majorVers >= 5 {
+		if majorVers > 4 || (majorVers >= 4 && minorVers >= 17) {
 			klog.Infof("loaded mounter: %s", volumeMounterKernel)
 			availableMounters = append(availableMounters, volumeMounterKernel)
 		} else {
--- a/pkg/cephfs/volumemounter.go
+++ b/pkg/cephfs/volumemounter.go
@@ -70,7 +70,7 @@ func loadAvailableMounters() error {
 		if err != nil {
 			return err
 		}
-		if majorVers >= 4 && minorVers >= 17 {
+		if (majorVers >= 4 && minorVers >= 17) || majorVers >= 5 {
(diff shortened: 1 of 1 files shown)


Madhu-1:
@KingJ thanks for the fix, @humblec this need to be back ported to 1.2.0

pkg/cephfs/volumemounter.go
```diff
@@ -70,7 +70,7 @@ func loadAvailableMounters() error {
 		if err != nil {
 			return err
 		}
-		if majorVers >= 4 && minorVers >= 17 {
+		if (majorVers >= 4 && minorVers >= 17) || majorVers >= 5 {
```
Madhu-1: Can this be changed to major>4 || (major >=4 & & minor >=17)

KingJ: Done, thanks for the feedback!

humblec:
> @KingJ thanks for the fix, @humblec this need to be back ported to 1.2.0

@Madhu-1 sure.. are u filing a backport PR?

Madhu-1:
> > @KingJ thanks for the fix, @humblec this need to be back ported to 1.2.0
> 
> @Madhu-1 sure.. are u filing a backport PR?

https://github.com/ceph/ceph-csi/pull/601

lpic10:
Is there any reason #601 won't get merged to 1.2.0 branch? 

Ubuntu LTS + HWE is running kernel 5.0 but there is no way I can force ceph-csi to use the kernel mounter as it detects incorrectly the kernel version as < 4.17

Madhu-1:
@lpic10 reopened #601, not use why I closed that backport

PR #570, opened 2019-08-24
Refactor helm charts

wilmardo:
# Describe what this PR does #

* Make the helm chart compatible with both Kubernetes versions.
* Bonus is that the generated templates by Helm are now linted by yamllint :)
* All the ports to variables (http-metrics and grpc metrics), chose to provide a service for each since Prometheus annotations do not support multiple ports (see [prometheus/prometheus#3756](https://github.com/prometheus/prometheus/issues/3756))

Fixes: #565
Implements changes discussed in: #555 

* ~Generate deployment files to keep the maintaince of changes in one place (the helm chart) to avoid errors and changed not being implement in one of the versions as we saw with #555~  
* ~Generated deployments deploy to the ceph-csi namespace which is created on deployment.~
Both split to #604

PR #604, opened 2019-09-10
WIP: Generate deployment files fom Helm charts

### LINKED ISSUE #143: static manifests from helm chart
Currently we have two different implementations of the driver plumbing. The helm chart and the static manifests. They can and do get out of sync. We could generate the static manifests as part of the release process by running 'helm template' on the chart. Then there would only be one set of templates to update for both use cases.

wilmardo:
# TODO:

- [x] Rebase after #570 is merged
- [ ] Rewrite E2E tests
- [ ] Rewrite documentation
- [ ] Rewrite examples
- [ ] Adds readme for deployment files ([example](https://github.com/keel-hq/keel/tree/master/deployment))
- [ ] Look into rook dependency on deployment files

Followup of #570 and fixes #143

Generate deployment files to keep the maintenance of changes in one place (the Helm chart) to avoid errors and changed not being implement in one of the versions as we saw with #555 
A single source of truth :)

Also generated deployments deploy to the ceph-csi namespace which is created on deployment. Bonus is that the generated templates by Helm are now linted by yamllint :)

### TLDR
* Deployment files in `/deploy` now generated
* Generated deployments deploy to ceph-csi namespace instead of default
* Generated deployment files are linted by yamllint which also brings up errors in the Helm template indenting and styling.

* The deploy files are now splitted by the template names in the helm chart which is different from the previous split by function(?).
* The topoly.enabled in the Helm chart now defaults to true to generate the same deployments files as commited in #816. In the PR the static manifests enable the topology feature gate without domains. But the Helm chart defaults to a disabled topology and the feature gate isn't enabled.
Now it is consistent, by default the feature gate gets enabled without domains.
* Podsecurity is default enabled in the Helm charts now, simplifies the generation and E2E tests, no need to override the values each time. I see no problem with this since in 1.14 psp was already beta and default enabled:
https://v1-14.docs.kubernetes.io/docs/concepts/policy/pod-security-policy/
* Default all the tests to use the Helm chart since this is the new source of truth. For the sake of completeness the static manifests are tested against the latest k8s version only.

# How to test this PR / New workflow

* Make a change to one of the templates in the `charts/` directory
* Run `./scripts/gen-deploy.sh` to generate the deployment files.
* Run `./scripts/lint-text.sh` to lint all the newly generated files and the changes in the Helm chart


nixpanic:
I would love to see this in release-3.0, @Madhu-1 what do you think?

Madhu-1:
@nixpanic this is one of the long pending improvement. @wilmardo thanks again for taking time to work on this

wilmardo:
Will rewrite the OP of this PR once I am up to speed with the changes again. Not everything is relevant anymore (1.13 is removed for example).

Will pick this up soon, today I rebased again master and setup the generation with `--show-only`. 
Next on the agenda are the E2E tests :)

.travis.yml
```diff
@@ -27,7 +27,11 @@ env:
     - TEST_COVERAGE=stdout
     - GO_METALINTER_THREADS=1
     - GO_COVER_DIR=_output
+<<<<<<< HEAD
     - HELM_VERSION=v3.1.2
+=======
+    - HELM_VERSION=v3.2.0-rc.1
+>>>>>>> fix: set to Helm3 in CI
```
nixpanic, line 34: this will likely break things, weird that Travis CI passed!?

nixpanic: Ah, it did not even run! Probably because of this YAML error.

Madhu-1:
@wilmardo hey any update on this one?

Madhu-1:
moved this one out of release-v3.0.0 milestone

wilmardo:
@Madhu-1 Sorry for this to linger so long, my time for opensource has been sparse lately.

@nixpanic @Madhu-1 Is there a convenient way to test the E2E test like the CI does locally? I tried running `scripts/travis-helmtest.sh v1.18.5 --test-cephfs=false --test-rbd=true` and even with some adaption to use the docker driver for minikube it wouldn't work. Before I dig into this further, how do you guys dev locally?

Madhu-1:
> @Madhu-1 Sorry for this to linger so long, my time for opensource has been sparse lately.
> 
> @nixpanic @Madhu-1 Is there a convenient way to test the E2E test like the CI does locally? I tried running `scripts/travis-helmtest.sh v1.18.5 --test-cephfs=false --test-rbd=true` and even with some adaption to use the docker driver for minikube it wouldn't work. Before I dig into this further, how do you guys dev locally?
> 
I use VM_DRIVER=kvm2 as extra i dont use docker driver, nothing much 

> Also made it a draft since that is now possible with existing PR's :)

Issue #199, opened 2019-02-14
[csi-rbd] Support erasure coded pools


lbogdan:
As suggested by @travisn, I'm opening this to track a feature gap between the rook RBD volume provisioner and the ceph CSI RBD plugin:

From what I could gather from the [source](https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/rbd_util.go#L138), when defining a `StorageClass`, the plugin only supports specifying a single pool, which will be sent as `--pool` argument to `rbd create`.
But, in order to be able to use erasure coded pools with RBD, you also need to send `--data-pool`, as described [here](http://docs.ceph.com/docs/mimic/rados/operations/erasure-code/#erasure-coding-with-overwrites).
Rook allows you to specify both the replicated metadata and the erasure coded data pool, as documented [here](https://github.com/rook/rook/blob/master/Documentation/ceph-block.md#advanced-example-erasure-coded-block-storage):
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
...
parameters:
  blockPool: replicated-metadata-pool
  dataBlockPool: ec-data-pool
```

rootfs:
this will be addressed in rook https://github.com/rook/rook/issues/2650

humblec:
@poornimag can you please take a look at this ?

dillaman:
> which will be sent as `--data` argument

Just to clarify, it is sent as the ```--pool``` argument

lbogdan:
@dillaman Updated, thanks for the heads up!

lbogdan:
Ping!
@rootfs From what I can see, there's still no way of creating a `StorageClass` backed by different metadata and  data pools. Any idea if this was considered / is on the roadmap?

Madhu-1:
@lbogdan #605 adds support for erasure-coded pools

PR #605, opened 2019-09-10
Add support for erasure pool in rbd

Madhu-1:
Allow specifying different metadata and data pools in a CSI RBD StorageClass

Fixes: #199
Fixes: https://github.com/rook/rook/issues/2650
Fixes: https://github.com/rook/rook/issues/3763


pkg/rbd/rbd_util.go
```diff
@@ -67,6 +67,7 @@ type rbdVolume struct {
 	VolID              string `json:"volID"`
 	Monitors           string `json:"monitors"`
 	Pool               string `json:"pool"`
+	DataPool           string `json:"dataPool"`
```
ShyamsundarR: (nit) I removed json annotations from elements of the structure that were not stores in the config maps in v1.0 (and earlier) of the code base. Going by the same yardstick, remove the trailing json annotation on this field.

pkg/rbd/rbd_util.go
```diff
@@ -127,10 +128,16 @@ func createImage(ctx context.Context, pOpts *rbdVolume, volSz int64, cr *util.Cr
 	} else {
 		klog.V(4).Infof(util.Log(ctx, "rbd: create %s size %s format %s using mon %s, pool %s"), image, volSzMiB, pOpts.ImageFormat, pOpts.Monitors, pOpts.Pool)
 	}
+	if pOpts.DataPool != "" {
```
ShyamsundarR: (nit) instead of 2 messages when an image is created against an EC pool, can this be folded into one of the messages above?

pkg/rbd/rbd_util.go
Madhu-1: forgot about 1.0.0 configmaps

pkg/rbd/rbd_util.go
Madhu-1: done

Madhu-1:
updated code

docs/deploy-rbd.md
```diff
@@ -50,6 +50,7 @@ make image-cephcsi
...
 | `clusterID`                                                                                         | yes                  | String representing a Ceph cluster, must be unique across all Ceph clusters in use for provisioning, cannot be greater than 36 bytes in length, and should remain immutable for the lifetime of the Ceph cluster in use |
 | `pool`                                                                                              | yes                  | Ceph pool into which the RBD image shall be created                                                                                                                                                                     |
+| `dataPool`                                                                                          | no                   | Erasure coded pool used for the data of the RBD images.                                                                                                                                                                 |
```
dillaman: Nit: it's not necessarily an EC pool.

pkg/rbd/rbd_util.go
```diff
@@ -125,12 +126,20 @@ func createImage(ctx context.Context, pOpts *rbdVolume, volSz int64, cr *util.Cr
 		klog.V(4).Infof(util.Log(ctx, "rbd: create %s size %s format %s (features: %s) using mon %s, pool %s"),
 			image, volSzMiB, pOpts.ImageFormat, pOpts.ImageFeatures, pOpts.Monitors, pOpts.Pool)
 	} else {
-		klog.V(4).Infof(util.Log(ctx, "rbd: create %s size %s format %s using mon %s, pool %s"), image, volSzMiB, pOpts.ImageFormat, pOpts.Monitors, pOpts.Pool)
+		logMsg := "rbd: create %s size %s format %s using mon %s, pool %s "
+		if pOpts.DataPool != "" {
+			logMsg += fmt.Sprintf("erasure coded pool %s", pOpts.DataPool)
```
dillaman:
```suggestion
			logMsg += fmt.Sprintf("data pool %s", pOpts.DataPool)
```

Madhu-1: done

docs/deploy-rbd.md
Madhu-1: update to ceph pool

Madhu-1:
@dillaman @ShyamsundarR @humblec PTAL

dillaman:
lgtm

pkg/rbd/rbd_util.go
```diff
@@ -125,12 +126,20 @@ func createImage(ctx context.Context, pOpts *rbdVolume, volSz int64, cr *util.Cr
 		klog.V(4).Infof(util.Log(ctx, "rbd: create %s size %s format %s (features: %s) using mon %s, pool %s"),
```
ShyamsundarR: (nit) Most often our image format is 2, as a matter of fact there was a discussion not to support v1 formats anyway with @dillaman . Considering this, the log message will miss out the data pool argument in most cases, as it gets appended to the message in the else condition. Do we want this corrected? (I am fine either way as the RPC payload message for CreateVolume will contain the arguments).

dillaman: @ShyamsundarR in a perfect world (as a separate PR), the `imageFormat` option should be removed since V1 images are not supported and cannot be created (and V1 doesn't support data pools). Otherwise agree that the log message should move.

Madhu-1:
@humblec PTAL

humblec:
One question before approving. Do we have any dependency to ceph version .etc to support EC pools @dillaman . If yes, we have to document it , so the question.

dillaman:
@humblec EC pools are supported starting w/ Luminous+BlueStore OSDs (i.e. cannot use Luminous+FileStore).

Madhu-1:
as per readme https://github.com/ceph/ceph-csi#ceph-csi-features-and-available-versions , rbd requires  >=Mimic this should be fine

Issue #615, opened 2019-09-11
 Handle concurrent operation of NodeUnstage with NodeStage and NodeUnpublish with NodePublish 


gnufied:
/kind bug

While reviewing the CEPH driver code, I found that we don't correctly handle concurrent processing of NodeUnstage call if a previous NodeStage was in-progress and same thing applies to NodeUnpublish call too.

The specs says, if an operation is in-progress for the volume then SP should return error code "10 ABORTED" - https://github.com/container-storage-interface/spec/blob/master/spec.md#timeouts

humblec:
Thanks @gnufied for reporting this ! much needed fix. The PR #616  address this issue. Thanks @Madhu-1 for quick action on this. Lets keep this in `1.2.1`. I am tagging the PR with the same.

PR #616, opened 2019-09-12
Change the logic of locking

Madhu-1:

if any ongoing operation is seen, we have to return an Abort error message

Fixes: #615

### DIFF
--- a/pkg/cephfs/controllerserver.go
+++ b/pkg/cephfs/controllerserver.go
@@ -33,17 +33,15 @@ import (
 type ControllerServer struct {
 	*csicommon.DefaultControllerServer
 	MetadataStore util.CachePersister
+	// A map storing all volumes with ongoing operations so that additional operations
+	// for that same volume (as defined by VolumeID/volume name) return an Aborted error
+	VolumeLocks *util.VolumeLocks
 }
 
 type controllerCacheEntry struct {
 	VolOptions volumeOptions
 	VolumeID   volumeID
 }
-
-var (
-	volumeIDLocker   = util.NewIDLocker()
-	volumeNameLocker = util.NewIDLocker()
-)
 
 // createBackingVolume creates the backing subvolume and on any error cleans up any created entities
 func (cs *ControllerServer) createBackingVolume(ctx context.Context, volOptions *volumeOptions, vID *volumeIdentifier, secret map[string]string) error {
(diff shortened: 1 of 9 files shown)


humblec:
@dillaman @ShyamsundarR PTAL.

humblec:
@dillaman @ShyamsundarR @poornimag can you please take a look at this ?

pkg/rbd/nodeserver.go
```diff
@@ -94,10 +97,13 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	}
 
 	stagingParentPath := req.GetStagingTargetPath()
-	stagingTargetPath := stagingParentPath + "/" + req.GetVolumeId()
+	stagingTargetPath := stagingParentPath + "/" + volID
 
-	idLk := nodeVolumeIDLocker.Lock(volID)
-	defer nodeVolumeIDLocker.Unlock(idLk, volID)
+	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
```
ShyamsundarR: Move this below line 75 above, fail fast. Previously we did work till we had no conflict with the prior request, now that we are aborting, there is little reason to do any work and abort.

pkg/cephfs/controllerserver.go
```diff
@@ -86,8 +84,11 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 
 	// Existence and conflict checks
-	idLk := volumeNameLocker.Lock(requestName)
-	defer volumeNameLocker.Unlock(idLk, requestName)
+	if acquired := cs.VolumeLocks.TryAcquire(requestName); !acquired {
```
ShyamsundarR: Move this below line 80 (fail fast)

pkg/cephfs/controllerserver.go
```diff
@@ -235,10 +239,19 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	}
 	defer cr.DeleteCredentials()
 
+	// lock out parallel delete operations
```
ShyamsundarR: Locking on `VolID` is not needed, if you are locking on `RequestName` just below. If you want to fail fast, before determining the 'RequestName`, then this lock is useful, and needs to move to line 214

pkg/cephfs/nodeserver.go
```diff
@@ -102,8 +101,11 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		}
 	}
 
-	idLk := nodeVolumeIDLocker.Lock(string(volID))
-	defer nodeVolumeIDLocker.Unlock(idLk, string(volID))
+	if acquired := ns.VolumeLocks.TryAcquire(req.GetVolumeId()); !acquired {
```
ShyamsundarR: (fail fast) move up to line 81

pkg/cephfs/nodeserver.go
```diff
@@ -167,11 +169,15 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, err
 	}
 
-	// Configuration
-
 	targetPath := req.GetTargetPath()
 	volID := req.GetVolumeId()
 
+	if acquired := ns.VolumeLocks.TryAcquire(targetPath); !acquired {
```
ShyamsundarR: Should this conflict with Staging operation? I would say it is safer it it does, as the assumed need for this is due to the CO sending multiple calls due to state loss, hence lock on the `VolID` instead.
It would serialize parallel publish requests of the same volume to different targets, but metrics show this is a fast operation on the local node, and hence should not suffer much due to the serialization slowness. (Unpublish would also need to change to the `VolID`)

pkg/rbd/nodeserver.go
```diff
@@ -238,10 +244,14 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	targetPath := req.GetTargetPath()
 	isBlock := req.GetVolumeCapability().GetBlock() != nil
 	stagingPath := req.GetStagingTargetPath()
-	stagingPath += "/" + req.GetVolumeId()
+	volID := req.GetVolumeId()
+	stagingPath += "/" + volID
 
-	idLk := targetPathLocker.Lock(targetPath)
-	defer targetPathLocker.Unlock(idLk, targetPath)
+	if acquired := ns.VolumeLocks.TryAcquire(targetPath); !acquired {
```
ShyamsundarR: Same comment as the Publish/Unpublish locks in CephFS, change this to lock on the `VolID` instead, and safeguard against parallel Staging and Publish requests for the same volume.

Madhu-1:
@ShyamsundarR comments are addressed PTAL

Madhu-1:
@humblec PTAL

humblec:
LGTM. Approved. Lets backport this to `1.2.0` branch

Issue #554, opened 2019-08-14
cephcsi-cephfs nodeplugin  pod got OOM killing when using FUSE mounter


yydzhou:
# Describe the bug #
When using cephfs FUSE mounter in cephfs csi, the cephcsi-cephfs nodeplugin  pod is very easy to get killed due to OOM. I have tried with 256M and then 1G resource setting, but the issue still happens. 

# Environment details #

- Image/version of Ceph CSI driver 
image: quay.io/cephcsi/cephfsplugin:v1.0.0
- helm chart version
- Kubernetes cluster version
1.13.4
- Logs

# Steps to reproduce #

deploying ceph and then cephfs csi driver + storageclass (mounter: fuse). 
Then create multiple cephfs pv/pvc and consume them with pods.
1. Setup details: '...'
1. Deployment to trigger the issue '....'
1. See error

# Actual results #

Describe what happened
`[167475.834265] Hardware name: RDO OpenStack Compute, BIOS 1.11.0-2.el7 04/01/2014
[167475.837017] Call Trace:
[167475.838397]  [<ffffffffb390e78e>] dump_stack+0x19/0x1b
[167475.840546]  [<ffffffffb390a110>] dump_header+0x90/0x229
[167475.842691]  [<ffffffffb34d805b>] ? cred_has_capability+0x6b/0x120
[167475.845070]  [<ffffffffb3397c44>] oom_kill_process+0x254/0x3d0
[167475.847349]  [<ffffffffb34d813e>] ? selinux_capable+0x2e/0x40
[167475.849595]  [<ffffffffb340f326>] mem_cgroup_oom_synchronize+0x546/0x570
[167475.852109]  [<ffffffffb340e7a0>] ? mem_cgroup_charge_common+0xc0/0xc0
[167475.854595]  [<ffffffffb33984d4>] pagefault_out_of_memory+0x14/0x90
[167475.856997]  [<ffffffffb3908232>] mm_fault_error+0x6a/0x157
[167475.859178]  [<ffffffffb391b8c6>] __do_page_fault+0x496/0x4f0
[167475.861395]  [<ffffffffb391ba06>] trace_do_page_fault+0x56/0x150
[167475.863692]  [<ffffffffb391af92>] do_async_page_fault+0x22/0xf0
[167475.865966]  [<ffffffffb39177b8>] async_page_fault+0x28/0x30
[...]

PR #623, opened 2019-09-18
cephfs: Choose kernel mounter when mounter specified is kernel

### LINKED ISSUE #617: allowing kernel mount for versions of kernel version < 4.17
Kernel mount for cephfs on kernel version < 4.17
# What new functionality do you want?
 When the `mounter: kernel` is specified in the storage class, the csi should do their best to use the specified mounter even through this may result in lose some features.  

There are a couple reasons. 
1. The kernel has better performance on latency and throughput, this is important for some latency sensitive applications.
2. The kernel mounting is much more stable as the FUSE has the issue https://github.com/ceph/ceph-csi/issues/554
3. This request is actually to match the default behavior of previous csi version as well as the old external provisioners. There is no quota support for a long time, and existing users are good with it. However, the new version of cephfs-csi changed this behavior (forces use FUSE mount when kernel < 4.17 https://github.com/ceph/ceph-csi/blob/master/pkg/cephfs/volumemounter.go#L73), which results a lot more effort needed for an upgrade.  

In addition, kernel 4.17 is toooo new for production requirement as the current Long time support kernel is 4.14. https://www.kernel.org/category/releases.html. There is 4.19 claimed to be long time support, but its support will end in next year. I think this enforcement should be delayed until there is a real long term support kernel version >= 4.17. 

Add a list of criteria that should be met for this feature to be useful

# Additional context #

poornimag:
Currently, the kernel mounter is chosen only when the kernel version is >= 4.17,
as the quota feature required for PV creation, is not available in older kernels.
With this patch we are changing the behaviour to choose kernel mounter if the
mounter specified in SC is kernel, irrespective of the kernel version.

Fixes #617

### DIFF
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -99,12 +99,12 @@ var _ = Describe("cephfs", func() {
 			}
 
 			By("create a storage class with pool and a PVC then Bind it to an app", func() {
-				createCephfsStorageClass(f.ClientSet, f, true)
+				createCephfsStorageClass(f.ClientSet, f, true, make(map[string]string))
 				validatePVCAndAppBinding(pvcPath, appPath, f)
 				deleteResource(cephfsExamplePath + "storageclass.yaml")
 			})
 
-			createCephfsStorageClass(f.ClientSet, f, false)
+			createCephfsStorageClass(f.ClientSet, f, false, make(map[string]string))
 
 			By("create and delete a PVC", func() {
 				By("create a PVC and Bind it to an app", func() {
@@ -114,6 +114,14 @@ var _ = Describe("cephfs", func() {
 
 				By("create a PVC and Bind it to an app with normal user", func() {
 					validateNormalUserPVCAccess(pvcPath, f)
+				})
+
+				By("create a PVC and Bind it to an app with kernel mounter", func() {
+					deleteResource(cephfsExamplePath + "storageclass.yaml")
+					createCephfsStorageClass(f.ClientSet, f, false, map[string]string{"mounter": "kernel"})
+					validatePVCAndAppBindingKernelMounter(pvcPath, appPath, f)
+					deleteResource(cephfsExamplePath + "storageclass.yaml")
+					createCephfsStorageClass(f.ClientSet, f, false, make(map[string]string))
 				})
 
 				By("create/delete multiple PVCs and Apps", func() {
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -193,7 +193,7 @@ func getStorageClass(path string) scv1.StorageClass {
 // 	return sc
 // }
 
-func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, enablePool bool) {
+func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, enablePool bool, parameters map[string]string) {
 	scPath := fmt.Sprintf("%s/%s", cephfsExamplePath, "storageclass.yaml")
 	sc := getStorageClass(scPath)
 	sc.Parameters["fsName"] = "myfs"
@@ -209,6 +209,9 @@ func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, en
 	fsID = strings.Trim(fsID, "\n")
 
 	sc.Parameters["clusterID"] = fsID
+	for k, v := range parameters {
+		sc.Parameters[k] = v
+	}
 	_, err := c.StorageV1().StorageClasses().Create(&sc)
 	Expect(err).Should(BeNil())
 }
@@ -591,6 +594,31 @@ func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
 	}
 	pvc.Namespace = f.UniqueName
 	e2elog.Logf("The PVC  template %+v", pvc)
+
+	app, err := loadApp(appPath)
+	if err != nil {
+		Fail(err.Error())
+	}
+	app.Namespace = f.UniqueName
+
+	err = createPVCAndApp("", f, pvc, app)
+	if err != nil {
+		Fail(err.Error())
+	}
+
+	err = deletePVCAndApp("", f, pvc, app)
+	if err != nil {
+		Fail(err.Error())
+	}
+}
+
+func validatePVCAndAppBindingKernelMounter(pvcPath, appPath string, f *framework.Framework) {
+	pvc, err := loadPVC(pvcPath)
+	if pvc == nil {
+		Fail(err.Error())
+	}
+	pvc.Namespace = f.UniqueName
+	e2elog.Logf("The PVC template %+v", pvc)
 
 	app, err := loadApp(appPath)
 	if err != nil {
--- a/examples/cephfs/storageclass.yaml
+++ b/examples/cephfs/storageclass.yaml
@@ -35,7 +35,9 @@ parameters:
   csi.storage.k8s.io/node-stage-secret-namespace: default
 
   # (optional) The driver can use either ceph-fuse (fuse) or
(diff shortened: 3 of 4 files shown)


Madhu-1:
@poornimag changes look good, can you add an E2E for kernel mounter? (PVC+APP)

humblec:
@poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?

poornimag:
> @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?

Nope, we will not be honouring the PVC size. The issue requested for this behaviour.

humblec:
> > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> 
> Nope, we will not be honouring the PVC size. The issue requested for this behaviour.

hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?

poornimag:
> hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?

Exactly. This sounds wrong to do so, but then the users requested for this behaviour. The SC has a comment, so the user would know what they are loosing when they choose kernel mounter on older kernels. @Madhu-1 ^

yydzhou:
> > > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> > 
> > 
> > Nope, we will not be honouring the PVC size. The issue requested for this behaviour.
> 
> hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> @poornimag



> > > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> > 
> > 
> > Nope, we will not be honouring the PVC size. The issue requested for this behaviour.
> 
> hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> @poornimag



> > > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> > 
> > 
> > Nope, we will not be honouring the PVC size. The issue requested for this behaviour.
> 
> hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> @poornimag



> > > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> > 
> > 
> > Nope, we will not be honouring the PVC size. The issue requested for this behaviour.
> 
> hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> @poornimag
This is to match the default behavior of previous version, and let the user to decide whether they want new feature (with upgrading kernel). Otherwise, the upgrade for existing users are very hard, giving the kernel 4.17 is too new and the FUSE mount has known issue https://github.com/ceph/ceph-csi/issues/554

e2e/cephfs.go
```diff
@@ -116,6 +116,14 @@ var _ = Describe("cephfs", func() {
 					validateNormalUserPVCAccess(pvcPath, f)
 				})
 
+				By("create a PVC and Bind it to an app with kernel mounter", func() {
+					deleteResource(cephfsExamplePath + "storageclass.yaml")
+					createCephfsStorageClass(f.ClientSet, f, false, map[string]string{"mounter": "kernel"})
+					validatePVCAndAppBindingKernelMounter(pvcPath, appPath, f)
```
Madhu-1: @poornimag you can  reuse `validatePVCAndAppBinding` function here

Madhu-1:
> > > > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> > > 
> > > 
> > > Nope, we will not be honouring the PVC size. The issue requested for this behaviour.
> > 
> > 
> > hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> > @poornimag
> 
> > > > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> > > 
> > > 
> > > Nope, we will not be honouring the PVC size. The issue requested for this behaviour.
> > 
> > 
> > hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> > @poornimag
> 
> > > > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> > > 
> > > 
> > > Nope, we will not be honouring the PVC size. The issue requested for this behaviour.
> > 
> > 
> > hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> > @poornimag
> 
> > > > @poornimag one question, if quota is not applied, are we respecting the size of a volume or PV?
> > > 
> > > 
> > > Nope, we will not be honouring the PVC size. The issue requested for this behaviour.
> > 
> > 
> > hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> > @poornimag
> 
> @humblec This is to match the default behavior of previous version, and let the user to decide whether they want new feature (with upgrading kernel). Otherwise, the upgrade for existing users are very hard, giving the kernel 4.17 is too new and the FUSE mount has known issue #554

@yydzhou we cant accept the PR if any subvol user can write to it regardless of the volume size they requested

humblec:
> > hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> 
> Exactly. This sounds wrong to do so, but then the users requested for this behaviour. The SC has a comment, so the user would know what they are loosing when they choose kernel mounter on older kernels. @Madhu-1 ^

@poornimag correct, as per kube, we have to make sure user get requested space at least. If we enable kernel mounter without quota its not guaranteed and its a breakage of Kube spec. So, I would like to NACK this PR for the same reason. @yydzhou

yydzhou:
> > > hmmm.. so as long as we have space in the parent volume, any subvol user can write to it regardless of the volume size they requested ?
> > 
> > 
> > Exactly. This sounds wrong to do so, but then the users requested for this behaviour. The SC has a comment, so the user would know what they are loosing when they choose kernel mounter on older kernels. @Madhu-1 ^
> 
> @poornimag correct, as per kube, we have to make sure user get requested space at least. If we enable kernel mounter without quota, its not guaranteed and its a breakage of Kube spec. Also, this is going to have a big effect on 'resize' , So, I would like to NACK this PR for the same reason. @yydzhou

@Madhu-1 @poornimag @humblec 
I don't think it's wrong. There is nothing to prevent users who want the quota support/sizing from using quota feature. We are just asking to provide an option to uses who do not want the feature to have to kernel mount option, which is much more important for existing users. Quota support is new feature and should be option to users, instead of breaking back compatibility, right?  the EFS, cephfs external provisioner, and even the csi v1.0 all support the kernel mount without support the quota. And lots of user are using them without issues. However, the CSI v1.2 forcely use FUSE mount, which is not reliable at all, to just enforce user to adopt the new feature. This sounds like wrong to me.  This enforcement code is checked in like a month ago. It prevents any users with kernel elder than 4.17 from upgrade in production environment. We have tried with the FUSE mount. It shows reliability issues even in our CICD pipeline, not to mention the production environment, which has much heavier workload and longer running time. To my understanding, forbidden using of kernel mount is almost equal to forbidden of upgrade in production environment, as the kernel upgrade has much wider impact. 

Again, I didn't see any drawback to those who want to quota feature, as the default behavior already ensure that. We  just want an option to overwrite this behavior and we know what we are doing.  If you are not comfort with it, you can add another switch in storage class like "skipQuota" to by pass the kernel restriction so that any user who setup this would definitely know what they are doing.

ajarr:
@poornimag , can we close this PR in favor of https://github.com/ceph/ceph-csi/pull/664 ?

Madhu-1:
@poornimag closing this in favor of https://github.com/ceph/ceph-csi/pull/664, Thanks for looking into this :+1:

Issue #621, opened 2019-09-17
Disable mkfs.ext4/xfs discard for RBD RXO volumes


dillaman:
# Describe the feature you'd like to have #620 

It's expensive (both in time and IO) to have ext4 or XFS discard a newly created RBD-backed block device when laying down the file system. Either get a patch merged into upstream k8s' `FormatAndMount` API that allows the RBD CSI to specify the no-discard flags to mkfs, or wrap the k8s `FormatAndMount` API call w/ a custom call that (1) calls `GetDiskFormat` to determine if there is a FS on the block device, (2) if not, directly invokes `mkfs`, and then (3) falls back to the original `FormatAndMount` call.   

# What is the value to the end user? (why is it a priority?) #

This will speed up volume creating on RBD-backed block devices since there is no need to zero the already zeroed RBD image.

# How will we know we have a good solution? (acceptance criteria) #

The `mkfs.ext4` call passes `-K` and `mkfs.xfs` calls passes `-E nodiscard` options.

# Additional context #

dillaman:
Also revert https://github.com/ceph/ceph-csi/pull/545 and https://github.com/rook/rook/pull/3601

humblec:
@dillaman I am already working on wrapping the formatAndMount in our driver. so will take that route.

PR #624, opened 2019-09-18
Use a wrapper for formatAndMount with rbd specific actions.

humblec:
Use a wrpper for formatAndMount with rbd specific actions.
    
    Currently rbd CSI plugin uses formatAndMount of
    mount.SafeFormatAndMount. This does not allow to pass or use
    specific formatting arguments with it. This patch introduce
    RBD specific formatting options with both `xfs` and `ext4`,
    for example: `-E no-discard` with ext4 and `-k` option with
    XFS to boost formatting performance of RBD device.

Fixes: https://github.com/ceph/ceph-csi/issues/621

## Future concerns ##

Follow up PR required to take `fmtArgs` from SC.

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -65,7 +65,7 @@ var _ = Describe("RBD", func() {
 		createRBDPool()
 		createConfigMap(rbdDirPath, f.ClientSet, f)
 		deployRBDPlugin()
-		createRBDStorageClass(f.ClientSet, f)
+		createRBDStorageClass(f.ClientSet, f, make(map[string]string))
 		createRBDSecret(f.ClientSet, f)
 
 	})
@@ -115,6 +115,17 @@ var _ = Describe("RBD", func() {
 			By("create a PVC and Bind it to an app with normal user", func() {
 				validateNormalUserPVCAccess(pvcPath, f)
 			})
+			// Skipping ext4 FS testing
+			/*
+				By("create a PVC and Bind it to an app with ext4 as the FS ", func() {
+					deleteResource(rbdExamplePath + "storageclass.yaml")
+					createRBDStorageClass(f.ClientSet, f, map[string]string{"csi.storage.k8s.io/fstype": "ext4"})
+					validatePVCAndAppBinding(pvcPath, appPath, f)
+					deleteResource(rbdExamplePath + "storageclass.yaml")
+					createRBDStorageClass(f.ClientSet, f, make(map[string]string))
+				})
+			*/
+
 			// skipping snapshot testing
 
 			// By("create a PVC clone and Bind it to an app", func() {
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -302,7 +302,7 @@ func getLegacyVolumeName(mountPath string) (string, error) {
 func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeStageVolumeRequest, stagingPath, devicePath string) error {
 	// Publish Path
 	fsType := req.GetVolumeCapability().GetMount().GetFsType()
-	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: mount.NewOsExec()}
+	diskMounter := &RFormatAndMount{mount.SafeFormatAndMount{Interface: ns.mounter, Exec: mount.NewOsExec()}}
 	opt := []string{}
 	isBlock := req.GetVolumeCapability().GetBlock() != nil
 	var err error
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -302,6 +302,29 @@ func getLegacyVolumeName(mountPath string) (string, error) {
 func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeStageVolumeRequest, stagingPath, devicePath string) error {
 	// Publish Path
 	fsType := req.GetVolumeCapability().GetMount().GetFsType()
+
+	// Use 'ext4' as the default
+	if fsType == "" {
+		fsType = "ext4"
+	}
+	// Disk is unformatted so format it.
+	fmtArgs := []string{devicePath}
+
+	if fsType == "ext4" {
+		fmtArgs = []string{
+			"-E nodiscard", // Do not attempt to discard blocks at mkfs time - Mainly used for RBD performance.
+			"-m0",          // Zero blocks reserved for super-user
+			"-F",           // Force flag
+			devicePath,
+		}
+
+	} else if fsType == "xfs" {
+		fmtArgs = []string{
+			"-K", // Do not attempt to discard blocks at mkfs time - Mainly used for RBD performance.
+			"-f",
+			devicePath,
+		}
+	}
 	diskMounter := &RFormatAndMount{mount.SafeFormatAndMount{Interface: ns.mounter, Exec: mount.NewOsExec()}}
 	opt := []string{}
 	isBlock := req.GetVolumeCapability().GetBlock() != nil
@@ -311,7 +334,7 @@ func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeS
 		opt = append(opt, "bind")
 		err = diskMounter.Mount(devicePath, stagingPath, fsType, opt)
 	} else {
-		err = diskMounter.FormatAndMount(devicePath, stagingPath, fsType, opt)
+		err = diskMounter.FormatAndMount(ctx, req, devicePath, stagingPath, fsType, opt, fmtArgs)
 	}
 	if err != nil {
 		klog.Errorf(util.Log(ctx, "failed to mount device path (%s) to staging path (%s) for volume (%s) error %s"), devicePath, stagingPath, req.GetVolumeId(), err)
--- a/pkg/rbd/rbd_mount.go
+++ b/pkg/rbd/rbd_mount.go
@@ -17,10 +17,13 @@ limitations under the License.
 package rbd
 
 import (
+	"context"
 	"fmt"
 
-	"github.com/pkg/errors"
+	"github.com/ceph/ceph-csi/pkg/util"
 
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/pkg/errors"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/util/mount"
 	utilexec "k8s.io/utils/exec"
@@ -33,7 +36,7 @@ const (
 	fsckErrorsUncorrected = 4
 )
 
-// rFormatAndMount probes a device to see if it is formatted.
+// RFormatAndMount probes a device to see if it is formatted.
 // Namely it checks to see if a file system is present. If so it
 // mounts it otherwise the device is formatted first then mounted.
 type RFormatAndMount struct {
@@ -42,7 +45,7 @@ type RFormatAndMount struct {
 
 // formatAndMount uses unix utils to format and mount the given disk
 // nolint: gocyclo
-func (mounter *RFormatAndMount) FormatAndMount(source, target, fstype string, options []string) error {
+func (mounter *RFormatAndMount) FormatAndMount(ctx context.Context, req *csi.NodeStageVolumeRequest, source, target, fstype string, options, fmtArgs []string) error {
 	readOnly := false
 	for _, option := range options {
 		if option == "ro" {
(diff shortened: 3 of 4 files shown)


humblec:
@Madhu-1 @dillaman @ShyamsundarR PTAL ..

dillaman:
@humblec Nit: your commit message has ext4 and XFS reversed for the options.

pkg/rbd/rbd_mount.go
```diff
@@ -0,0 +1,133 @@
...
+			readOnly = true
+			break
+		}
+	}
+
+	options = append(options, "defaults")
+
+	if !readOnly {
+		// Run fsck on the disk to fix repairable issues, only do this for volumes requested as rw.
+		klog.V(4).Infof("Checking for issues with fsck on disk: %s", source)
```
Madhu-1: log with req-ID here like https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/nodeserver.go#L110

Madhu-1: this applies for other places also

pkg/rbd/rbd_mount.go
```diff
@@ -0,0 +1,133 @@
...
+	// 'fsck' found errors and corrected them
+	fsckErrorsCorrected = 1
+	// 'fsck' found errors but exited without correcting them
+	fsckErrorsUncorrected = 4
+)
+
+// rFormatAndMount probes a device to see if it is formatted.
+// Namely it checks to see if a file system is present. If so it
+// mounts it otherwise the device is formatted first then mounted.
+type RFormatAndMount struct {
```
Madhu-1: RBDFormatAndMounter , `rFormatAndMount` is used in comment please update the comment also

humblec:
Sure, will add E2E in this PR itself.

Madhu-1:
@idryomov PTAL

pkg/rbd/rbd_mount.go
```diff
@@ -0,0 +1,133 @@
...
+	mount.SafeFormatAndMount
+}
+
+// formatAndMount uses unix utils to format and mount the given disk
+// nolint: gocyclo
+func (mounter *RFormatAndMount) FormatAndMount(source, target, fstype string, options []string) error {
```
Madhu-1: Can you pass the format options from the node stage request and make it as an argument for this function

pkg/rbd/rbd_mount.go
```diff
@@ -0,0 +1,133 @@
...
+				args = []string{
+					"-F",           // Force flag
+					"-m0",          // Zero blocks reserved for super-user
+					"-E nodiscard", // Do not attempt to discard blocks at mkfs time - Mainly used for RBD performance.
+					source,
+				}
+
+			} else if fstype == "xfs" {
+				args = []string{
+					"-K", // Do not attempt to discard blocks at mkfs time - Mainly used for RBD performance.
```
idryomov: Need -f here for consistency with with ext3/4.

e2e/rbd.go
```diff
@@ -115,6 +115,15 @@ var _ = Describe("RBD", func() {
 			By("create a PVC and Bind it to an app with normal user", func() {
 				validateNormalUserPVCAccess(pvcPath, f)
 			})
+
+			By("create a PVC and Bind it to an app with ext4 as the FS ", func() {
+				deleteResource(rbdExamplePath + "storageclass.yaml")
+				createRBDStorageClass(f.ClientSet, f, map[string]string{"csi.storage.k8s.io/fstype": "ext4"})
+				validatePVCAndAppBindingExt4Mount(pvcPath, appPath, f)
```
Madhu-1: validatePVCAndAppBindingExt4Mount is not needed please reuse `validatePVCAndAppBinding`

Madhu-1:
> This looks like a cut-and-paste of the core function. I would do something
> along the lines of:
> 
> ```
> existingFormat, err := mounter.GetDiskFormat(source)
> if err != nil {
> 	return err
> }
> if existingFormat == "" && !readOnly {
> 	if fstype == "" {
> 		fstype = "ext4"
> 	}
> 	args := []string{source}
> 	if fstype == "ext4" {
> 		args = []string{"-F", "-m0", "-E nodiscard", source}
> 	} else if fstype == "xfs" {
> 		args = []string{"-f", "-K", source}
>         }
>         _, err := mounter.Exec.Run("mkfs."+fstype, args...)
>         if err != nil {
>         	return err
>         }
> }
> ```
> 
> and then call the core function for mounting.

@idryomov what about other formats `ext3` etc

idryomov:
> @idryomov what about other formats `ext3` etc

My snippet runs mkfs with no options for them, just like the core function would.  But thinking about it, I would call mkfs only for ext4 and xfs here and leave everything else for the core function.

humblec:
>This looks like a cut-and-paste of the core function. I would do something
along the lines of:
>
@idryomov Not really, there were small changes. However I thought of keeping the workflow, error returns  as it is or almost untouched. The main reason being, its a well tested `common` code considering 'linux' and possible FSs in mind and also while considering all the drivers which work with kubernetes and its FS format and mount workflow. We havent seen any issues with this code in kube or its pretty solid. I would like to keep to the same as much as we can for the same reason.  We can think about simplifying or improving it just for RBD later though.

pkg/rbd/rbd_mount.go
humblec: Done.

humblec:
@Madhu-1 @dillaman @idryomov I have disabled the E2E test for "ext4" now due to some CI issues. However it was tested manually and it worked, @Madhu-1 can also confirm. I will have a followup PR for getting this configured from SC parameter list and other improvements can be done on the same if any. Please review this PR . Thanks.

pkg/rbd/nodeserver.go
```diff
@@ -302,7 +302,31 @@ func getLegacyVolumeName(mountPath string) (string, error) {
...
+	}
+	// Disk is unformatted so format it.
+	fmtArgs := []string{source}
+
+	if fsType == "ext4" {
+		fmtArgs = []string{
+			"-E nodiscard", // Do not attempt to discard blocks at mkfs time - Mainly used for RBD performance.
+			"-m0",          // Zero blocks reserved for super-user
+			"-F",           // Force flag
+			source,
```
Madhu-1: can we use the devicePath instead of source here, let's not create a new variable

pkg/rbd/rbd_mount.go
```diff
@@ -0,0 +1,116 @@
...
+	// 'fsck' found errors and corrected them
+	fsckErrorsCorrected = 1
+	// 'fsck' found errors but exited without correcting them
+	fsckErrorsUncorrected = 4
+)
+
+// RFormatAndMount probes a device to see if it is formatted.
+// Namely it checks to see if a file system is present. If so it
+// mounts it otherwise the device is formatted first then mounted.
+type RFormatAndMount struct {
```
Madhu-1: can we change this to `RBBFormatAndMount` or `FormatAndMount`

pkg/rbd/rbd_mount.go
```diff
@@ -0,0 +1,116 @@
...
+*/
+
+package rbd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ceph/ceph-csi/pkg/util"
+	"github.com/container-storage-interface/spec/lib/go/csi"
```
Madhu-1: new line required here

pkg/rbd/rbd_mount.go
```diff
@@ -0,0 +1,116 @@
...
+			readOnly = true
+			break
+		}
+	}
+
+	options = append(options, "defaults")
+
+	if !readOnly {
+		// Run fsck on the disk to fix repairable issues, only do this for volumes requested as rw.
+		klog.V(4).Infof("Checking for issues with fsck on disk: %s", source)
```
Madhu-1: extra logging, this can be removed

humblec: Done.

pkg/rbd/rbd_mount.go
humblec: Done.

pkg/rbd/rbd_mount.go
humblec: it will land on rbd.RBDFormatAndMount, so shortformed.

pkg/rbd/nodeserver.go
humblec: Done.

pkg/rbd/nodeserver.go
```diff
@@ -302,7 +302,30 @@ func getLegacyVolumeName(mountPath string) (string, error) {
 func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeStageVolumeRequest, stagingPath, devicePath string) error {
 	// Publish Path
 	fsType := req.GetVolumeCapability().GetMount().GetFsType()
-	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: mount.NewOsExec()}
+
+	// Use 'ext4' as the default
+	if fsType == "" {
```
Madhu-1: note: `ext4` is defaulted in external-provisioner if no fstype is specified in SC

idryomov:
> > This looks like a cut-and-paste of the core function. I would do something
> > along the lines of:
> 
> @idryomov Not really, there were small changes. However I thought of keeping the workflow, error returns as it is or almost untouched. The main reason being, its a well tested `common` code considering 'linux' and possible FSs in mind and also while considering all the drivers which work with kubernetes and its FS format and mount workflow. We havent seen any issues with this code in kube or its pretty solid. I would like to keep to the same as much as we can for the same reason. We can think about simplifying or improving it just for RBD later though.

This is exactly my point.  Given a well tested common function, why duplicate it in the plugin and make changes to it?  Do just the part we care about (i.e. mkfs) and call the common function, unchanged.

humblec:
@idryomov the `fsck` part has to be there and some code from `FormatAndMount ` moved to caller. Whatever the workflow you have in mind, can we do it in a follow up PR ? I would let you to send the PR for the same :).

dillaman:
> the `fsck` part has to be there and some code from `FormatAndMount ` moved to caller

@humblec Isn't that only true for pre-mount? If there is no file system (as determined by `mounter.GetDiskFormat`), you would `mkfs` and then call the original `FormatAndMount` logic (which would run `fsck` and mount).

idryomov:
> @idryomov the `fsck` part has to be there and some code from `FormatAndMount ` moved to caller. 

I'm confused.  Are you saying we need the fsck part in the plugin?

> Whatever the workflow you have in mind, can we do it in a follow up PR ? I would let you to send the PR for the same :).

What I'm proposing is:

```
do mkfs in the plugin
call the common FormatAndMount to do fsck and mount
```

instead of duplicating the entire FormatAndMount in the plugin and supplementing it with fmtArgs.

I don't see a point in adding some code just to remove it in a follow-up.  In the end of the day your approach works, my comment was just a suggestion.

humblec:
> > the `fsck` part has to be there and some code from `FormatAndMount ` moved to caller
> 
> @humblec Isn't that only true for pre-mount? If there is no file system (as determined by `mounter.GetDiskFormat`), you would `mkfs` and then call the original `FormatAndMount` logic (which would run `fsck` and mount).

@idryomov I doubt thats only the case for premount. Try checking the FS first - assuming a FS exist already..etc make use of this `fsck` code path. I have also seen some discussions saying that, this ( assuming a FS already exist) workflow  could help in various scenarios .., Also, there are various combinations like `, different volumeModes ( file, block..etc) , PV access modes,  network attached FSs ..etc and this code path  kind of take care all of these . One thing we are sure here is, this wont break, thats why I am sticking to it and its an important code path. 

Again, we can always come up with some changes on this  later if needed based on our testing and after making sure nothing breaks with those changes...etc..

dillaman:
>  doubt thats only the case for premount.

@humblec I think you were responding to the wrong person. The `GetDiskFormat` just runs `blkid` so it doesn't matter if the file system has errors. If there is no FS id, you can make a new FS and than mount it via `FormatAndMount`.

humblec:
@dillaman @idryomov at times, its difficult to make sure we are understanding each other from PR comments. any way, were you suggesting something like this https://github.com/ceph/ceph-csi/pull/629 ? if thats the case, the main caveat I see with that is: *) We do repeat the 'formatting' in mount failed conditions in the core function and it wont have `discard` or whatever option we want.  When we land in that code path is  a question though. *) Secondly `readOnly` flag and check in `mountVolumeToStagePath` does not make any sense . however please let me know your thoughts. we can confirm the path accordingly.

humblec:
Also, with latest approach ( PR 629) we close the possibilities of expanding our own mounter if needed.

dillaman:
> @dillaman @idryomov at times, its difficult to make sure we are understanding each other from PR comments :). any way, were you suggesting something like this #629 ?

Yes, that is what I was expecting.

> if thats the case, the main caveat I see with that is: *) We do repeat the 'formatting' in mount failed conditions in the core function and it wont have `discard` or whatever option we want. 

I am not really concerned about that case since it should never hit. If the `mkfs` succeeds, now `mounter.GetDiskFormat` will return the FS type and therefore `FormatAndMount` will not attempt to format it. 

>  When we land in that code path is a question though. *) Secondly `readOnly` flag and check in `mountVolumeToStagePath` does not make any sense . however please let me know your thoughts. we can confirm the path accordingly.

I guess I don't understand. What doesn't make sense? Just the fact that you would need to check `options` for `ro`?

humblec:
> Yes, that is what I was expecting.

@dillaman Thanks for the confirmation.

> 
> > if thats the case, the main caveat I see with that is: *) We do repeat the 'formatting' in mount failed conditions in the core function and it wont have `discard` or whatever option we want.
> 
> I am not really concerned about that case since it should never hit. If the `mkfs` succeeds, now `mounter.GetDiskFormat` will return the FS type and therefore `FormatAndMount` will not attempt to format it.
>

Ideally yes, but not sure whether there are any conditions which can cause we land into that path. The source code comment in the core function says:

```	// Mount failed. This indicates either that the disk is unformatted or
		// it contains an unexpected filesystem.
```
so, do you expect a condition where FS formatting was initiated ( may be superblock is written) and broken or disconnected can cause such scenario? in short, next time  blkid say there is a FS on top of it , but mount fails? I was thinking these kind of scenarios in first place or with the approach here in the PR. 


> 
> I guess I don't understand. What doesn't make sense? Just the fact that you would need to check `options` for `ro`?
>

We added `readonly` check in caller, I dont think we have reference to readOnly in `mountVolumeToStagePath`. Also, the `options` we pass to core function is just filled with `bind` in it. Thats what I was referring in my comment.

dillaman:
> so, do you expect a condition where FS formatting was initiated ( may be superblock is written) and broken or disconnected can cause such scenario? in short, next time blkid say there is a FS on top of it , but mount fails? I was thinking these kind of scenarios in first place or with the approach here in the PR.

That's no different from the current code. If the mount fails and `existingFormat != ""`, it just fails. It doesn't attempt to re-format the device nor should it since it implies some higher level issue. 

> We added `readonly` check in caller, I dont think we have reference to readOnly in `mountVolumeToStagePath`. Also, the `options` we pass to core function is just filled with `bind` in it. Thats what I was referring in my comment.

If `opt` was always an empty array in the past, it implies that we don't support read-only RBD-backed file system volumes? Again, this wouldn't change the existing behaviour of `FormatAndMount` since if we never pass any valid options, it will default to R/W mode and format the empty block device.

idryomov:
> We added `readonly` check in caller, I dont think we have reference to readOnly in `mountVolumeToStagePath`. Also, the `options` we pass to core function is just filled with `bind` in it. Thats what I was referring in my comment.

I didn't notice that we control the options and "ro" is impossible.  Just the drop readOnly.

humblec:
>That's no different from the current code. If the mount fails and existingFormat != "", it just fails. It doesn't attempt to re-format the device nor should it since it implies some higher level issue.
>

@dillaman  @idryomov  So, the expectation here is that, we will never into the situation of `mount fails` + `existingFormat == ""` ?  Isnt it ? If thats the case, I will pick (629) as the final PR for now. 

Ragarding , `RO` check in caller and `core function`, lets simply dont worry about `RO` volumes at all :)..

dillaman:
> @dillaman @idryomov So, the expectation here is that, we will never into the situation of `mount fails` + `existingFormat == ""` ? Isnt it ? If thats the case, I will pick (629) as the final PR for now.

Yes, I can sleep soundly at night with that expectation. It should never happen since we would have formatted the disk and returned a failure if it failed.  But in the worse-case, it just misses the reformats w/o the discard options and life goes on.

humblec:
Thanks for the confirmation. Closing this in favor of https://github.com/ceph/ceph-csi/pull/629 :+1:

PR #89, opened 2018-10-15
support nsmounter when running in containerized mode

rootfs:
this is to address the two bugs discussed in the issue with @pohly 
fix #86

PR #628, opened 2019-09-19
Fix unmount if rbd nodeplugin restarts

Madhu-1:

unmount bind-mounted directory in rbd
if the node plugin restarts there  will be extra mounts in the `/rootfs/proc/mounts` we need to unmount both the directories to completely unmount the PVC from application

once the node plugin restarts there will 2X entry for target path in `/proc/mounts`

ex:=  if 20 pods are running on  a node after node restart the entry in  `/proc/mounts` will be 20x2 +10
= 30 entries  
20 entries for `/var/lib/kubelet/xxx` --- targetpath
20 entries for /rootfs/var/lib/xxx` --- targetpath
10 entries for stagingpath

leaving above comment for reference

with new changes, we have deprecated the `containerized` flag and removed mount  `/` of the host to  `rootfs` of container
## Related issues ##

Fixes: #466


Madhu-1:
@humblec @nixpanic @ShyamsundarR PTAL

e2e/utils.go
```diff
@@ -608,6 +608,13 @@ func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
 		Fail(err.Error())
 	}
 }
+func deletePodWithLabel(label string) error {
+	_, err := framework.RunKubectl("delete", "-nrook-ceph", "po", "-l", label)
+	if err != nil {
+		e2elog.Logf("failed to delete rbd daemonset %v", err)
```
Madhu-1: corrected log here, waiting for  CI to finish, will update the PR

Madhu-1:
CI Passed

ShyamsundarR:
Looking at this I am trying to understand why there are the additional 2 mounts? What is different in cephFS that this does not happen (or maybe is untested)?

It looks like CephFS does not do bind mounts in Stage, and also does not use nsenter to do the mounts. Further, attempting to understand where `/rootfs` is actually coming in, and showing up the 2 mounts that were performed by the older version of the plugin.

Attempting to connect the dots and then review the fix. At present this seems like a workaround, as **I** do not understand the reason or the root cause.

NOTE: `/rootfs` is required when using nsenter based mounts: https://github.com/ceph/ceph-csi/blob/18151e9ca8d288fcf80c5ee91a22b683c5178c41/vendor/k8s.io/utils/nsenter/nsenter.go#L56

Madhu-1:
> Looking at this I am trying to understand why there are the additional 2 mounts? What is different in cephFS that this does not happen (or maybe is untested)?
> 

as you said below cephfs doesn't uses the nsenter for  mounting

> It looks like CephFS does not do bind mounts in Stage, and also does not use nsenter to do the mounts. Further, attempting to understand where `/rootfs` is actually coming in, and showing up the 2 mounts that were performed by the older version of the plugin.
> 

we don't need to have an older version of the plugin.  this  will occur if the node plugin pod contains app mount and it is restarted

> Attempting to connect the dots and then review the fix. At present this seems like a workaround, as **I** do not understand the reason or the root cause.
> 
@nixpanic can help with this, I have Niels has done the analysis, he said double entry are expected as `/` is bind mounted in to `/rootfs` 

> NOTE: `/rootfs` is required when using nsenter based mounts:
> 
> https://github.com/ceph/ceph-csi/blob/18151e9ca8d288fcf80c5ee91a22b683c5178c41/vendor/k8s.io/utils/nsenter/nsenter.go#L56

ShyamsundarR:
> > It looks like CephFS does not do bind mounts in Stage, and also does not use nsenter to do the mounts. Further, attempting to understand where `/rootfs` is actually coming in, and showing up the 2 mounts that were performed by the older version of the plugin.
> 
> we don't need to have an older version of the plugin. this will occur if the node plugin pod contains app mount and it is restarted

Ack, I was not referring to any older version of the plugin here, but we are on the same page.

> 
> > Attempting to connect the dots and then review the fix. At present this seems like a workaround, as **I** do not understand the reason or the root cause.
> 
> @nixpanic can help with this, I have Niels has done the analysis, he said double entry are expected as `/` is bind mounted in to `/rootfs`

FWIW, I did not restart the nodeplugin, but started a pod that mounts `/` as `/rootfs` and see that within this pod, the rbd mounts show up (under `/rootfs`).

I think we need to understand why we need nsenter and hence `/rootfs` mounted on the node plugin pod (I am thinking this should be removed as of now). Further, why a host path mount propagates a mount from within it, and hence shows this up as a double mount.

NOTE: Attempting to understand the behavior further during my day

ShyamsundarR:
Here is an alternate fix that removes the need to use `nsenter`.

We already have `/var/lib/kubelet/[plugin|pod]` mounted to the nodeplugin with Bi-directional mount propagation semantics. Thus, any mount we do within the nodeplugin container/pods will be visible to the host, and there is no reason to use `nsenter` to mount this using the host namespace.

Tested the scenario of nodeplugin being restarted, and this works as expected as there is no longer a `/rootfs` that points to `/`. (also tested block volumeMode)

I was not able to narrow down, why a hostPath mount of `/` to `/rootfs` causes these additional mounts to appear within the nodeplugin container, but not outside of it, but as posted earlier, a simple pod that does the same (i.e mounts `/` as a hostPath) also gets these mounts. It would have been nice if i had an answer for the same, but for now, removing the nsenter based mounting should be good to resolve this.

```
diff --git a/deploy/rbd/kubernetes/v1.14+/csi-rbdplugin.yaml b/deploy/rbd/kubernetes/v1.14+/csi-rbdplugin.yaml
index 2897bdae..0af50310 100644
--- a/deploy/rbd/kubernetes/v1.14+/csi-rbdplugin.yaml
+++ b/deploy/rbd/kubernetes/v1.14+/csi-rbdplugin.yaml
@@ -81,8 +81,6 @@ spec:
               mountPath: /csi
             - mountPath: /dev
               name: host-dev
-            - mountPath: /rootfs
-              name: host-rootfs
             - mountPath: /sys
               name: host-sys
             - mountPath: /lib/modules
@@ -138,9 +136,6 @@ spec:
         - name: host-dev
           hostPath:
             path: /dev
-        - name: host-rootfs
-          hostPath:
-            path: /
         - name: host-sys
           hostPath:
             path: /sys
diff --git a/pkg/rbd/driver.go b/pkg/rbd/driver.go
index a1bdbc91..634956fe 100644
--- a/pkg/rbd/driver.go
+++ b/pkg/rbd/driver.go
@@ -23,9 +23,6 @@ import (
        "k8s.io/klog"
        "k8s.io/kubernetes/pkg/util/mount"
-       nsutil "k8s.io/kubernetes/pkg/volume/util/nsenter"
-       "k8s.io/utils/exec"
-       "k8s.io/utils/nsenter"
 )
 
 const (
@@ -80,13 +77,6 @@ func NewControllerServer(d *csicommon.CSIDriver, cachePersister util.CachePersis
 // NewNodeServer initialize a node server for rbd CSI driver.
 func NewNodeServer(d *csicommon.CSIDriver, containerized bool, t string) (*NodeServer, error) {
        mounter := mount.New("")
-       if containerized {
-               ne, err := nsenter.NewNsenter(nsenter.DefaultHostRootFsPath, exec.New())
-               if err != nil {
-                       return nil, err
-               }
-               mounter = nsutil.NewMounter("", ne)
-       }
        return &NodeServer{
                DefaultNodeServer: csicommon.NewDefaultNodeServer(d, t),
                mounter:           mounter,
```

Madhu-1:
@ShyamsundarR before trying to find out I tried this one removing `/`  and `rootfs` from the nodeplugin and setting `containerized=false` it works but this is not going to fix the issue for the people who are already using CSI  now right?

now the question is why `nsenter` was added earlier ( there must be a reason for it) this was added by @rootfs in https://github.com/ceph/ceph-csi/pull/89 

without getting an answer why it was added  simply I cant go and remove it

even if you want to remove the `nsenter` i would consider this as a breaking change which is good for `v2.0.0` not for `v1.2.1` @humblec can confirm.

with  this fix, we can fix the issues for users who are using csi `v1.2.0` without any deployment changes just be updating the container image

note:- Even kubelet  as deprecated the use of `nsenter` so so many reasons, we  need  to find out do we need `nsenter` in ceph-csi, if not remove the use of `nsenter` and do a major release as we are  removing it

@nixpanic ^^

ShyamsundarR:
> @ShyamsundarR before trying to find out what was the issue with `nsenter` I also tried removing `/` and `rootfs` from the nodeplugin and setting `containerized=false` it works but this is not going to fix the issue for the people who are already using CSI now right?

It will, as once nodeplugin is restarted with the new image, and manifest, it will not mount `/` as `/rootfs` and hence not have to deal with the double mounts.

> 
> now the question is why `nsenter` was added earlier ( there must be a reason for it) this was added by @rootfs in #89
> 
> without getting an answer why it was added simply I cant go and remove it

I do not see a need for it. So removing it seems logical to solve this problem.

> 
> even if you want to remove the `nsenter` i would consider this as a breaking change which is good for `v2.0.0` not for `v1.2.1` @humblec can confirm.

Please elaborate why this is a breaking change?

> 
> with this fix, we can fix the issues for users who are using csi `v1.2.0` without any deployment changes just be updating the container image

The current fix is a workaround at best, we also do not know why we have those extra mounts in the first place.

> 
> note:- Even kubelet as deprecated the use of `nsenter` so so many reasons, we need to find out do we need `nsenter` in ceph-csi, if not remove the use of `nsenter` and do a major release as we are removing it
> 
> @nixpanic ^^

nixpanic:
If there are no processes that need to be executed on the host through `nsenter` (like LVM commands), dropping `/rootfs` is definitely the best option. It would be required to have the mount namespace from the host shared with the container, otherwise the mounts will disappear when the container restarts.

I am not sure what is needed to have the mount namespace of the host inside the container. Possibly just running the container as privileged is sufficient?

Madhu-1:
deprecated `containerized` flag and removed `nsenter`

@ShyamsundarR @nixpanic @humblec PTAL

ShyamsundarR:
> If there are no processes that need to be executed on the host through `nsenter` (like LVM commands), dropping `/rootfs` is definitely the best option. It would be required to have the mount namespace from the host shared with the container, otherwise the mounts will disappear when the container restarts.
> 
> I am not sure what is needed to have the mount namespace of the host inside the container. Possibly just running the container as privileged is sufficient?

The `/var/lib/kubelet/[pods|plugins]` mounts are set with [`mountPropagation: "Bidirectional"`](https://github.com/ceph/ceph-csi/blob/e2890a27ffe7c530295c62de3bf159dbdbc8537e/deploy/rbd/kubernetes/v1.14%2B/csi-rbdplugin.yaml#L93-L98), which ensures the mounts propagate outside the nodeplugin for the required orchestration by Kube to attach them to the pods.

I also further think the `hostPath` directive that mounts `/` to `/rootfs` is using the `--rbind` option, but not able to find out where in code that is being used/attempted. This would cause the double mount as seen, but possibly is a defined/understood behavior of using hostPath mounts? (anyway mostly academic now, for the purposes of this patch)

humblec:
One thing which I am trying to understand here is this:

Why the "/rootfs" mount [1] *not* available in the plugin pod, before plugin pod restart if the same code path of staging/publishing is executed before or after restart ?  

[1]
For ex:

/dev/rbd0 on /rootfs/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-6ce678a2-67fc-40c9-b349-9bc50dae7410/globalmount/0001-0009-rook-ceph-0000000000000001-bbe13109-dabe-11e9-80a3-eaf51db2f8da type ext4 (rw,relatime,seclabel,stripe=1024,data=ordered)

Madhu-1:
> @Madhu-1 @nixpanic @ShyamsundarR One thing which I am trying to understand here is this:
> 
> Why the "/rootfs" mount [1] _**not**_ available in the plugin pod, before plugin pod restart if the same code path of staging/publishing is executed before or after restart ?
> 
> [1]
> For ex:
> 
> /dev/rbd0 on /rootfs/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-6ce678a2-67fc-40c9-b349-9bc50dae7410/globalmount/0001-0009-rook-ceph-0000000000000001-bbe13109-dabe-11e9-80a3-eaf51db2f8da type ext4 (rw,relatime,seclabel,stripe=1024,data=ordered)
am not sure why it's happening. I don't have answer for this one

as per my understanding earlier `nsenter` was added to fix some issue related to mapping/unmapping was done at the node publish call not at node stage call, as we implemented the mapping in node stage call do we need to debug this further? if yes @humblec what are your suggestion? please let me know

Madhu-1:
use of `nsenter` can fail if the OS doesn't ship few binaries see https://github.com/rook/rook/issues/3935#issuecomment-533921396 https://github.com/rook/rook/issues/3923

Madhu-1:
@humblec ^^

humblec:
Lets get this in and continue our analysis in parallel. LGTM. approved

PR #635, opened 2019-09-24
Set default FS to "ext4" on RBD device

humblec:
We were defaulting to ext4 at first and then moved to xfs.
However further testing shows that, ext4 should be
the default or preferred fs for RBD devices.

This patch bring that change


humblec:
@Madhu-1 @ShyamsundarR

humblec:
PR subject is addressed.

PR #636, opened 2019-09-24
Format RBD volumes with `nodiscard` formatting options.

humblec:
Currently rbd CSI plugin uses formatAndMount of
mount.SafeFormatAndMount. This does not allow to pass or use
specific formatting arguments with it. This patch introduce
RBD specific formatting options with both xfs and ext4,
for example: -E no-discard with ext4 and -k option with
XFS to boost formatting performance of RBD device.

(cherry picked from commit 0e6617e1ffff4dc93e7606bfdf0dcaf60f3c327d)


humblec:
@Madhu-1 @ShyamsundarR PTAL.

PR #641, opened 2019-09-25
We were defaulting to ext4 at first and then moved toxfs.

humblec:
We were defaulting to ext4 at first and then moved to xfs.
However further testing shows that there are some issues with XFS locking while we use XFS on top of RBD device, while xfs issues are getting addressed, we are changing the default FS type to ext4.
This patch bring that change

(cherry picked from commit b44a81bdbc7abf4f4f565300cc82a1d4f36d7d00)


humblec:
@Madhu-1 PTAL.

PR #644, opened 2019-09-25
Fix volsize for cephfs and rbd

Madhu-1:
once the PVC is required and backend is created kubernetes sidecar will round off the values to a human-readable format, we have to create images with the same size to avoid volume size mismatch between PVC response and backend image size.

### DIFF
--- a/pkg/cephfs/controllerserver.go
+++ b/pkg/cephfs/controllerserver.go
@@ -84,17 +84,24 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	defer cs.VolumeLocks.Release(requestName)
 
-	volOptions, err := newVolumeOptions(ctx, requestName, req.GetCapacityRange().GetRequiredBytes(),
-		req.GetParameters(), secret)
+	volOptions, err := newVolumeOptions(ctx, requestName, req.GetParameters(), secret)
 	if err != nil {
 		klog.Errorf(util.Log(ctx, "validation and extraction of volume options failed: %v"), err)
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	if req.GetCapacityRange() != nil {
+		volOptions.Size = util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
+	}
+	// TODO need to add check for 0 volume size
+
 	vID, err := checkVolExists(ctx, volOptions, secret)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+
+	// TODO return error message if requested vol size greater than found volume return error
+
 	if vID != nil {
 		return &csi.CreateVolumeResponse{
 			Volume: &csi.Volume{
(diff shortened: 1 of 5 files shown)


pkg/cephfs/controllerserver.go
```diff
@@ -84,17 +84,25 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+	volOptions, err := newVolumeOptions(ctx, requestName, req.GetParameters(), secret)
 	if err != nil {
 		klog.Errorf(util.Log(ctx, "validation and extraction of volume options failed: %v"), err)
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	if req.GetCapacityRange() != nil {
+		volOptions.Size = util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
+	}
+	// TODO need to add check for 0 volume size
```
humblec: I doubt we get a response at `0` size. It should have been blocked by CO.  have to double confirm though.

pkg/cephfs/controllerserver.go
```diff
@@ -84,17 +84,25 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+
 	vID, err := checkVolExists(ctx, volOptions, secret)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+
+	// TODO comapare request volume size and found volume sizem if requested is
```
humblec: typo on sizem? or better to reword this todo.

humblec:
@Madhu-1 please check the review comments.

pkg/cephfs/controllerserver.go
Madhu-1: not sure other CO will not process the vol size of `0`, but as per CSI spec it says volume size must not be negative but it can be zero. even earlier in cephfs, we are able to  create backend images with `0` size

pkg/cephfs/controllerserver.go
Madhu-1: updated

humblec:
@ShyamsundarR @poornimag PTAL.

humblec:
@ShyamsundarR @poornimag can you review this in priority?

pkg/util/util.go
```diff
@@ -36,12 +37,43 @@ import (
...
+	roundedUp := volumeSizeBytes / allocationUnitBytes
+	if volumeSizeBytes%allocationUnitBytes > 0 {
+		roundedUp++
+	}
+	return roundedUp
+}
+
+// RoundOffBytes converts roundoff the size
+// 1.1Mib will be round off to 2Mib same for GiB
+// size less than will be round off to 1MiB
```
poornimag: nit: size less than 1MiB will be round off to 1MiB.

pkg/util/util.go
```diff
@@ -36,12 +37,43 @@ import (
 const (
 	// MiB - MebiByte size
 	MiB = 1024 * 1024
+	GiB = MiB * 1024
 )
 
-// RoundUpToMiB rounds up given quantity upto chunks of MiB
-func RoundUpToMiB(size int64) int64 {
+// RoundOffVolSize rounds up given quantity upto chunks of MiB/GiB
+func RoundOffVolSize(size int64) int64 {
```
poornimag: RoundOffVolSize could call RoundOffBytes and convert it into number of MiBs and return?

Madhu-1: yes  this can be done, updated now

pkg/util/util.go
Madhu-1: done

PR #649, opened 2019-09-26
Backport PR when label is set

Madhu-1:
backport PR to release-v1.2.0 branch from master branch  when a label is set on PR

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -19,3 +19,11 @@ pull_request_rules:
         strict: smart
       dismiss_reviews: {}
       delete_head_branch: {}
+  - name: backport patches to release branch
+    conditions:
+      - base=master
+      - label=backport-to-release-v1.2.0
+    actions:
+      backport:
+        branches:
+          - release-v1.2.0
--- a/docs/development-guide.md
+++ b/docs/development-guide.md
@@ -89,3 +89,10 @@ need to be met before it will be merged:
 
 When the criteria are met, a project maintainer can merge your changes into
 the project's master branch.
+
+### Backporting a PR to release branch
+
+If the PR needs to be backported to a release branch, a project maintainer adds
+`backport-to-release-vX.Y.Z` label to the PR. Mergify bot will take care of
+sending the backport PR to release branch once the PR is merged in the master
+branch.


Madhu-1:
This PR needs to be merged manually as there  is a change in mergify rules

.mergify.yml
```diff
@@ -10,12 +10,20 @@ pull_request_rules:
   - name: automatic merge
     conditions:
       - label!=DNM
-      - '#approved-reviews-by>=2'
-      - 'status-success=continuous-integration/travis-ci/pr'
+      - "#approved-reviews-by>=2"
+      - "status-success=continuous-integration/travis-ci/pr"
```
nixpanic: Why change the quoting?

.mergify.yml
```diff
@@ -10,12 +10,20 @@ pull_request_rules:
...
       merge:
         method: rebase
         rebase_fallback: merge
         strict: smart
       dismiss_reviews: {}
       delete_head_branch: {}
+  - name: backport patches to release branch
+    conditions:
+      - base=master
+      - label=backport-to-release-v1.2.0
```
nixpanic: There should be a document with the meanings of the labels. This change should add the automatic backport feature to that doc. But I can not find the document... Maybe `./docs/development-guide.md` would be the right place?

Madhu-1: updated doc

.mergify.yml
Madhu-1: done by the editor, mergify rules are verified its correct

nixpanic: Ok.

**Nit:** It is better to make these changes as a separate commit, or mention is in the commit message as not being related to the functional change. What should be done when someone else uses a different editor/settings and the quoting gets replaced again?

.mergify.yml
nixpanic: Thanks!

docs/development-guide.md
```diff
@@ -89,3 +89,10 @@ need to be met before it will be merged:
 
 When the criteria are met, a project maintainer can merge your changes into
 the project's master branch.
+
+### Backporting a PR to release branch
+
+If the PR needs to be backport to a release branch, add
+`backport-to-release-vX.Y.Z` label to the PR. Mergify bot will take care of
```
nixpanic: How can contributors add the label? I can not do so...

Madhu-1: updated doc, once maintainers can add  labels

.mergify.yml
Madhu-1: good  question removed the change

PR #629, opened 2019-09-19
 Format RBD volumes with `nodiscard` option.

humblec:
Currently rbd CSI plugin uses formatAndMount of
mount.SafeFormatAndMount. This does not allow to pass or use
specific formatting arguments with it. This patch introduce
RBD specific formatting options with both `xfs` and `ext4`,
for example: `-E no-discard` with ext4 and `-k` option with
XFS to boost formatting performance of RBD device.

Signed-off-by: Humble Chirammal hchiramm@redhat.com

Fixes: #621

## Future concerns ##
Follow up PR required to take fmtArgs from SC.

PR #666, opened 2019-10-04
Format ext4 with lazy_journal_init

idryomov:
Skip zeroing of the journal on freshly created images.  As only dynamic
PVs are supported, it is fine to assume that existingFormat == "" image
has never been mapped and written to before.

lazy_itable_init is enabled by default, passing it for consistency with
lazy_journal_init.

### DIFF
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -310,25 +310,38 @@ func getLegacyVolumeName(mountPath string) (string, error) {
 }
 
 func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeStageVolumeRequest, stagingPath, devicePath string) error {
-	// Publish Path
 	fsType := req.GetVolumeCapability().GetMount().GetFsType()
 	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: mount.NewOsExec()}
+
+	// rbd images are thin-provisioned and return zeros for unwritten areas.  A freshly created
+	// image will not benefit from discard and we also want to avoid as much unnecessary zeroing
+	// as possible.  Open-code mkfs here because FormatAndMount() doesn't accept custom mkfs
+	// options.
+	//
+	// Note that "freshly" is very important here.  While discard is more of a nice to have,
+	// lazy_journal_init=1 is plain unsafe if the image has been written to before and hasn't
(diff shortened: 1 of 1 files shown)


humblec:
@idryomov can you please update the PR and commit description with the details. I see source code is well commented though :+1:

idryomov:
I've updated the PR with the text from the commit message.

humblec:
@idryomov ooc, are we seeing any performance gain with this change ? or did you get a chance to validate it?

pkg/rbd/nodeserver.go
```diff
@@ -310,25 +310,38 @@ func getLegacyVolumeName(mountPath string) (string, error) {
...
+	// uninitialized and carries a risk until the journal is overwritten and wraps around for
+	// the first time).
 	existingFormat, err := diskMounter.GetDiskFormat(devicePath)
 	if err != nil {
 		klog.Errorf(util.Log(ctx, "failed to get disk format for path %s, error: %v"), devicePath, err)
 		return err
 	}
-	if existingFormat == "" && (fsType == "ext4" || fsType == "xfs") {
+	// TODO: update this when adding support for static (pre-provisioned) PVs
+	if existingFormat == "" /* && !staticVol */ {
```
ShyamsundarR: Why did we remove the `fsType` check here? The discussion in the prior PR (https://github.com/ceph/ceph-csi/pull/629) was to retain the same, hence wondering why.

We do send empty `options` to `FormatAndMount`,  which in turn uses mkfs in case the format comes back as an empty string for all but `ext4` and `ext3`, so functionally we are doing the same operation.

idryomov:
This takes my base test (a stripped down replica of the original 50-100 1G PVs test that started PV creation/deletion performance conversations in https://github.com/ceph/ceph-csi/issues/507 -- concurrent rbd map + mkfs.ext4 + mount for 50 1G images on a vstart cluster) from 90-100 seconds to 12-15 seconds.

humblec:
Thats great. Thanks for sharing @idryomov

pkg/rbd/nodeserver.go
idryomov: For clarity.  It is effectively replaced with `len(args) > 0` check below.  We still intercept only ext4 and xfs, and call `FormatAndMount` with no prior action for everything else.

ShyamsundarR: :man_facepalming: thanks!

Issue #676, opened 2019-10-09
Manage RBD image mirroring via StorageClass parameter


dillaman:
# Describe the feature you'd like to have #

Add a new SC 'mirroring' parameter that can be used to annotate that PVs should be mirrored via RBD mirroring. The current values for this parameter can be "off" and "journal" (or similar). When set to "journal", upon PVC creation the exclusive-lock and journaling RBD image features should be enabled. Additionally, the CSI will need to enable mirroring on the newly created RBD image via `rbd mirror image enable <image-spec>`. This will also imply that NBD will need to be used instead of krbd.

# What is the value to the end user? (why is it a priority?) #

How would the end user gain value from having this feature?

# How will we know we have a good solution? (acceptance criteria) #

Add a list of criteria that should be met for this feature to be useful

# Additional context #

techhanx:
Hi. is this issue still on-going? or developing by someone?
If not, I would like to work on it. @dillaman @humblec @Madhu-1 

and I suggest that we can divide this issue to two issues.
1. add rbd image feature: journaling
2. add rbd mirror image enable

humblec:
@WoohyungHan can you please revisit the PR or patch which solve this issue?  we would like to get this in `2.0.0`.

WoohyungHan:
> @WoohyungHan can you please revisit the PR or patch which solve this issue? we would like to get this in `2.0.0`.

sure, I will. @humblec

dillaman:
This requires k8s CSI external-provisioner PR https://github.com/kubernetes-csi/external-provisioner/pull/399

dillaman:
We want to support the new Ceph Octopus snapshot-based mirroring, so the storage class's "mirroring" attribute should really be an enum of "disabled" (default), "journal", or "snapshot". The snapshot-based mirroring should be functional with the existing krbd driver. The journal-based mirroring mode requires rbd-nbd, but that is not yet production ready since we have no upgrade path for it.

Madhu-1:
Moved this out of release-v3.0.0 milestone

humblec:
Considering we have mirroring support available with release v3.3.0. I am closing this one .Please feel free to reopen if required.

Madhu-1:
We may still need this one If someone wants to enable mirroring when creating the PVC. But they need new volume replication after some time to perform failover and failback. Let's keep it open for some time.

Rakshith-R:
[...]

PR #684, opened 2019-10-15
rbd: Add exclusive-lock and journaling image features for RBD image

woohhan:
Add exclusive-lock and journaling image features for rbd image
    
    Current rbd plugin only supports the layering feature
    for rbd image. Add exclusive-lock and journaling image
    features for the rbd plugin.

If we enable the journaling feature, krbd doesn't work for the mount. We need to force mapping by rbd-nbd.

## Related issues ##
closes: #676 

## Future concerns ##
NONE


techhanx:
Hi, I'm glad to join ceph-csi. If I have wrong code or concept, please let me know. I will fix up immediately. Thanks!

Madhu-1:
@dillaman PTAL

techhanx:
@Madhu-1 
Thanks for your reply. I will add E2E test to cover this.
but i didn't understand other two suggestion. could you please explain more?

- passing exlusive-lock and journaling from storageclass
I'm testing by creating storage class parameter with 'imageFeatures: layering,journaling,exclusive-lock' and check the image has these features. is there any other to do? (except e2e tests)

- Testing with krbd and rbd-nbd
If we enable image journaling option, we can't mount with krbd. because It is not supported that mounting the journaling volume in krbd. we need to use rbd-nbd.

Madhu-1:
> @Madhu-1
> Thanks for your reply. I will add E2E test to cover this.
> but i didn't understand other two suggestion. could you please explain more?
> 
> * passing exlusive-lock and journaling from storageclass

okay I missed adding more information we need to update documentation and storageclass  examples for these features https://github.com/ceph/ceph-csi/blob/master/examples/rbd/storageclass.yaml#L24-L26 and  https://github.com/ceph/ceph-csi/blob/master/docs/deploy-rbd.md

>   I'm testing by creating storage class parameter with 'imageFeatures: layering,journaling,exclusive-lock' and check the image has these features. is there any other to do? (except e2e tests)
> * Testing with krbd and rbd-nbd
>   If we enable image journaling option, we can't mount with krbd. because It is not supported that mounting the journaling volume in krbd. we need to use rbd-nbd.

i that case we need to error out  in the node stage request if the mounter is `krbd`

pkg/rbd/rbd_util.go
```diff
@@ -468,13 +468,27 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
...
+			if f == "journaling" {
+				journaling = true
+			} else if f == "exclusive-lock" {
+				exclusivelock = true
+			}
 		}
+
+		// if journaling is enabled, exclusive-lock must be enabled too.
+		if journaling && !exclusivelock {
+			return nil, fmt.Errorf("cannot use journaling without exclusive lock")
```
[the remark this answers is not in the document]
WoohyungHan: Done

WoohyungHan:
@Madhu-1 I apply your suggestion. Could you please check the commit? Thanks.

dimm0:
How about object-map feature? Any reason to not enable it?

woohhan:
@dimm0 Yes, there's no reason to not enable the object-map feature. But the pr stale for now, If pr is processed, I will add the feature. Thanks!! :)

dimm0:
Is this still pending in the new version?

cr1cr1:
@Madhu-1 did you get a chance to review this PR? Thank you.

pkg/rbd/rbd_util.go
```diff
@@ -468,13 +470,26 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
...
+			if f == "journaling" {
+				journaling = true
+			} else if f == "exclusive-lock" {
+				exclusivelock = true
+			}
+		}
+
+		// if journaling is enabled, exclusive-lock must be enabled too.
+		if journaling && !exclusivelock {
+			return nil, errors.Errorf("cannot use journaling without exclusive lock")
```
Madhu-1, line 490: this should be errors.New

woohhan: Done!

woohhan:
@Madhu-1 I made some code changes to make it easy to add image features in the future. For example, if we want to add a fast-diff image feature, we need to specify it needs an object map in the future. Could you please review the code again?

pkg/rbd/rbd_util.go
```diff
@@ -113,8 +114,32 @@ type rbdSnapshot struct {
...
+	needRbdNbd bool
+	// required is the image features required for this ImageFeature
+	required []string
+}
+
 var (
-	supportedFeatures = sets.NewString("layering")
+	supportedFeatures = map[string]ImageFeature{
+		"layering": {
+			value:      librbd.RbdFeatureLayering,
```
nixpanic, line 130: These names will be deprecated with go-ceph v0.3.0 (to be released today) use `FeatureLayering` and similat (without the `Rbd` prefix) instead.

Functions to help mapping the integer values to names, and the other way around have been added too. Have a look at https://github.com/ceph/go-ceph/blob/master/rbd/features.go to see what you can use later today (and update go-ceph in `go.mod`).

woohhan: Thanks for your suggestion. I've seen the implementation and I think It would be great If this PR used these. @Madhu-1 Is it okay to change the go-ceph version in this PR? or do we need another PR to change the go-ceph version?

Madhu-1: sure , we can do it as part of this PR as a separate commit, but we need to wait till go-ceph makes a new release

woohhan: @nixpanic It would be nice to have a function that checks if a `FeatureSet` has a specific feature. and I think we can confirm that the requirement of the feature is met or not. Is there anything in progress in go-ceph?

```
func (fs *FeatureSet) hasFeature(name string) bool
```

nixpanic: No, that is not planned to be included. I would convert the `name` to the bit(s) and check if `fs` has those bits set.

Untested (idea from [the unit tests](https://github.com/ceph/go-ceph/blob/master/rbd/features_test.go#L47)):
```go
fs && FeatureSetFromNames([]string{name}) == FeatureSetFromNames([]string{name})
```

I am not sure how useful such a feature in go-ceph would be. You can file an Issue or PR for the project, and we'll consider/review it. We try to keep go-ceph as simple as possible, and do not try to add additional coating/sugar/utility functions, so we may not want to include this kind of functionality.

woohhan: ok, I agree that go-ceph should be as simple as possible. After go-ceph v0.3.0 is released, I'll fix this PR with the changes. Thanks.

nixpanic: Thanks!

woohhan:
@Madhu-1 I am looking at how to change this PR and trying to change the `ImageFeature` field from string to `librbd.FeatureSet` in `rbdVolume` struct. Can this makes another impact? Why is there a JSON tag in the rbdVolume type? The implementation of FeatureSet is here: https://github.com/ceph/go-ceph/blob/master/rbd/features.go#L106

current rbdVolume
```
type rbdVolume struct {
	RbdImageName       string
// ...
	ImageFeatures      string `json:"imageFeatures"`
// ...
}
```

changed rbdVolume
```
type rbdVolume struct {
	RbdImageName       string
// ...
	ImageFeatures      librbd.FeatureSet
// ...
}
```

Madhu-1:
> @Madhu-1 I am looking at how to change this PR and trying to change the `ImageFeature` field from string to `librbd.FeatureSet` in `rbdVolume` struct. Can this makes another impact? Why is there a JSON tag in the rbdVolume type? The implementation of FeatureSet is here: https://github.com/ceph/go-ceph/blob/master/rbd/features.go#L106
> 
> current rbdVolume
> 
> ```
> type rbdVolume struct {
> 	RbdImageName       string
> // ...
> 	ImageFeatures      string `json:"imageFeatures"`
> // ...
> }
> ```
> 
> changed rbdVolume
> 
> ```
> type rbdVolume struct {
> 	RbdImageName       string
> // ...
> 	ImageFeatures      librbd.FeatureSet
> // ...
> }
> ```

we have json tag to support backward compatibility for v1.0.0 volumes, removing/changing the fields may result in breakage

woohhan:
> we have json tag to support backward compatibility for v1.0.0 volumes, removing/changing the fields may result in breakage

Ok, I will proceed without changing the field.

woohhan:
@Madhu-1 I added a new commit, and it seems to work fine. Can you please review again?

go.mod
```diff
@@ -28,7 +28,7 @@ require (
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
 	google.golang.org/appengine v1.6.5 // indirect
 	google.golang.org/genproto v0.0.0-20200413115906-b5235f65be36 // indirect
-	google.golang.org/grpc v1.26.0
+	google.golang.org/grpc v1.27.0
```
nixpanic, line 31: this grpc update seems unrelated?

pkg/rbd/rbd_util_test.go
```diff
@@ -54,3 +54,68 @@ func TestHasSnapshotFeature(t *testing.T) {
...
+			"",
+			"",
+			"feature exclusive-lock requires rbd-nbd for mounter",
+		},
+		{
+			map[string]string{"imageFeatures": "layering,exclusive-lock,journaling", "mounter": rbdDefaultMounter},
+			"",
+			"",
+			"feature exclusive-lock requires rbd-nbd for mounter",
+		},
```
nixpanic, line 100: Can you add a test for an image feature that does not exist? Imagine a typo in the `imageFeatures` value.

go.mod
woohhan: Oh, It seems that I skipped `go mod verify` in the go-ceph version update. Thanks!

pkg/rbd/rbd_util_test.go
woohhan: I added a unit test for when the image feature does not exist.

go.mod
woohhan: I changed this to the go-ceph's version change.

nixpanic: I am still confused by this... If there is a grpc update, those files should get included in `vendor/` too, but I do not see them. go-ceph does not depend on grpc, so I do not know where it comes from.

Ideally you would not include a grpc update in here at all, or maybe as a 3rd commit (1st in the series)?

Madhu-1:
@woohhan did you make `go mod vendor`? as grpc is having a minor update, the go modules will get auto-updated, we cannot control it how we controller it in dep

woohhan:
@Madhu-1 @nixpanic 

I think I found the cause. Ran the `go mod verify` command on the master branch, the go.mod file was changed in the GRPC version from 1.26.0 to 1.27.0. I think we need to update the master branch with `go mod verify` command. 

We're checking `go mod verify` in the CI pipeline: [link](https://github.com/ceph/ceph-csi/blob/master/Makefile#L52) however, this doesn't seem enough. (It always return 0)

I think the reason there is no modification in `/vendor` is because we have the line below in `go.mod`. When I deleted that line and ran `go mod vendor`, the modification was applied to the vendor directory.

```
replace (
    google.golang.org/grpc => google.golang.org/grpc v1.26.0
// ...
)
```

woohhan:
I think it would be ok if I add the third commit for updating the GRPC version.

woohhan:
The commit was updated.

woohhan:
> Looks good to me, but maybe you want to correct the subject of the 3rd commit, it has a typo in "journlaing"

Oh, Thanks. Fixed typo!

woohhan:
@Madhu-1 Hi, Could you please check the PR? Thanks!

go.mod
nixpanic, line 31: It seems grpc is expected to be in version 1.26. I am working in #961 to get that corrected. Can you please drop this change from your PR? Once #961 is merged, the `go.mod` should not get out-of-sync any longer.

woohhan:
> updating grpc does not seem to be expected, it needs to stay at v1.26
Ok, I Removed unnecessary commits. @nixpanic @Madhu-1

pkg/rbd/rbd_util.go
```diff
@@ -126,8 +127,27 @@ type rbdSnapshot struct {
 	RequestName    string
 }
 
+// ImageFeature represents required image features and value
+type ImageFeature struct {
+	// needRbdNbd indicates whether this image feature requires an rbd-nbd mounter
+	needRbdNbd bool
+	// required is the image features required for this ImageFeature
+	required []string
```
phlogistonjohn, line 135: This attribute name is a bit confusing IMO. Perhaps calling it `dependencies` or `dependsOn` would clarify what it means.

pkg/rbd/rbd_util.go
```diff
@@ -126,8 +127,27 @@ type rbdSnapshot struct {
 	RequestName    string
 }
 
+// ImageFeature represents required image features and value
+type ImageFeature struct {
```
phlogistonjohn, line 131: This probably does not need to be a public type, right?

pkg/rbd/rbd_util.go
```diff
@@ -613,6 +619,40 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
...
+	if !found {
+		mounter = rbdDefaultMounter
+	}
+
+	imageFeatures, found := volOptions["imageFeatures"]
+	if !found {
+		return "", mounter, nil
+	}
+
+	arr := strings.Split(imageFeatures, ",")
```
phlogistonjohn, line 634: This is probably outside the scope of the PR, but why split strings on a delimiter. Isn't the data coming in thru yaml, as structured data? Can't this be a proper list/slice before it gets here?

pkg/rbd/rbd_util_test.go
```diff
@@ -54,3 +54,80 @@ func TestHasSnapshotFeature(t *testing.T) {
...
+		imageFeatures, mounter, err := validateAndGetImageFeaturesWithMounter(test.volOptions)
+
+		if !isCorrectError(err, test.isErrMsg) || (imageFeatures != test.isImageFeature) || (mounter != test.isMounter) {
+			t.Errorf("validateAndGetImageFeaturesWithMounter(%#v) = %s/%s/%v, want %s/%s/%s",
+				test.volOptions, imageFeatures, mounter, err, test.isImageFeature, test.isMounter, test.isErrMsg)
+		}
+	}
+}
+
+func isCorrectError(err error, errorMsg string) bool {
```
phlogistonjohn, line 125: Instead of hand-rolling functions like this perhaps it would be better to start using a packge like testify?

Even if not, treating nil errors as equal to an empty string seems like a trap for future devs. If you keep this function I highly recommend rolling a nil-check and a same-content check as two different functions.

woohhan:
hi, @phlogistonjohn. Thanks for your review! I will check the review and may apply the suggestion soon.

pkg/rbd/rbd_util.go
woohhan: We have imageFeatures as String in storageclass before, here see [link](https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/csi/rbd/storageclass.yaml#L39). It would be nice to change to yaml's structured data, but it would be difficult because it is not compatible with the previous version.

pkg/rbd/rbd_util_test.go
woohhan: yes, that's correct. @Madhu-1 Do we have a package to check for errors? If not, I can add an appropriate package, or split the function as suggested by @phlogistonjohn

pkg/rbd/rbd_util.go
woohhan: yes, I will fix this. Thanks!

pkg/rbd/rbd_util.go
woohhan: yes, I will fix this. Thanks!

pkg/rbd/rbd_util.go
Madhu-1: yeah, the imagefeature in storage-class is a comma-separated values

pkg/rbd/rbd_util_test.go
Madhu-1: currently, we don't have any package, but you can check https://godoc.org/github.com/stretchr/testify

nixpanic: Hi @woohhan, are you planning to update this PR with the suggestion?

woohhan: just now I updated the PR with the suggestion!

nixpanic:
I wonder how testing succeeds... we should probably add a check that nothing makes uncommitted changes to `vendor/`.

Madhu-1:
> I wonder how testing succeeds... we should probably add a check that nothing makes uncommitted changes to `vendor/`.

testify is already part for vendor directory as this is required by some dependency. we need to set it
explicitly to some specific version in  go.mod

woohhan:
> > I wonder how testing succeeds... we should probably add a check that nothing makes uncommitted changes to `vendor/`.
> 
> testify is already part for vendor directory as this is required by some dependency. we need to set it
> explicitly to some specific version in go.mod

@nixpanic When I ran the `go mod vendor` command, there's nothing changed, so I checked the vendor directory and found that it already exists as @Madhu-1 said.

Do we need to set the version of the testify? If we specify a verison that is the same as in the vendor directory, there will be no changes.

humblec:
> > > I wonder how testing succeeds... we should probably add a check that nothing makes uncommitted changes to `vendor/`.
> > 
> > 
> > testify is already part for vendor directory as this is required by some dependency. we need to set it
> > explicitly to some specific version in go.mod
> 
> @nixpanic When I ran the `go mod vendor` command, there's nothing changed, so I checked the vendor directory and found that it already exists as @Madhu-1 said.
> 
> Do we need to set the version of the testify? If we specify a version that is the same as in the vendor directory, there will be no changes. (except go.mod and go.sum)

@woohhan can you `make all` ?

internal/rbd/rbd_util.go
```diff
@@ -614,6 +620,40 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
...
+	if !found {
+		return "", mounter, nil
+	}
+
+	arr := strings.Split(imageFeatures, ",")
+	featureSet := sets.NewString(arr...)
+	for _, f := range arr {
+		sf, found := supportedFeatures[f]
+		if !found {
+			return "", "", errors.Errorf("invalid feature %s for volume csi-rbdplugin", f)
```
humblec, line 640: @woohhan just wondering will this cause some backward compatibiliy issues depend on the backend ceph cluster.  That said, suppose a feature called `X` is enabled and in ceph cluster version "Y" and if its working for the customer atm. If this is getting connected to a cluster which is `Y-1` and if `X` is not a supported feature, we will return saying "invalid feature" , Isnt it ?

woohhan: Thanks for the comments! I understood what your saying, but I wonder what scenario does that happen. Are there any situations where a running ceph-csi connects to a lower version Ceph? (for example, Ceph downgrade??)

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
   name: rook-ceph-block
provisioner: rook-ceph.rbd.csi.ceph.com
parameters:
    clusterID: rook-ceph
    // If the version goes down, we must delete this `Y` and re-create the storage class.
    imageFeatures: Y
```

woohhan:
> > > > I wonder how testing succeeds... we should probably add a check that nothing makes uncommitted changes to `vendor/`.
> > > 
> > > 
> > > testify is already part for vendor directory as this is required by some dependency. we need to set it
> > > explicitly to some specific version in go.mod
> > 
> > 
> > @nixpanic When I ran the `go mod vendor` command, there's nothing changed, so I checked the vendor directory and found that it already exists as @Madhu-1 said.
> > Do we need to set the version of the testify? If we specify a version that is the same as in the vendor directory, there will be no changes. (except go.mod and go.sum)
> 
> @woohhan can you `make all` ?

I did `make all` but nothing changed!

internal/rbd/rbd_util.go
humblec: @woohhan the point here is, as long as customer is free to connect to any cluster version which is listed in the matrix ( README) at any point in time. So, if SC has a param which is not compatible, I am wondering `returning` error or `just logging` and moving on is the better way. also there is a fact that, may be soon multicluster support..etc may get added .

woohhan: @humblec I didn't think about that. I changed the code with just log when error cases. It will help the user to track the error in case of invalid image features(or mounter)

woohhan:
Semantic Pull Request not restart?

Madhu-1:
@woohhan can you force push once again?

nixpanic:
I think the CI failed because of this problem:

```
May 11 12:21:12.848: INFO: At 2020-05-11 12:11:12 +0000 UTC - event for csi-rbd-demo-pod: {attachdetach-controller } SuccessfulAttachVolume: AttachVolume.Attach succeeded for volume "pvc-806abbbd-9380-11ea-9312-42010a140134" 
May 11 12:21:12.848: INFO: At 2020-05-11 12:11:29 +0000 UTC - event for csi-rbd-demo-pod: {kubelet minikube} FailedMount: MountVolume.MountDevice failed for volume "pvc-806abbbd-9380-11ea-9312-42010a140134" : rpc error: code = Internal desc = rbd: map failed exit status 6, rbd output: rbd: sysfs write failed
RBD image feature set mismatch. You can disable features unsupported by the kernel with "rbd feature disable replicapool/csi-vol-80a56ad0-9380-11ea-9c69-0242ac110010 journaling".
```

In case the attacher/rbd-plugin needs to write to `/sys/...`, it may need additional configuration in the deployment spec.

(note that the *Semantic Pull Request* bot is not enforced, failing is accepted - and we don't understand yet why it fails)

woohhan:
> I think the CI failed because of this problem:
> 
> ```
> May 11 12:21:12.848: INFO: At 2020-05-11 12:11:12 +0000 UTC - event for csi-rbd-demo-pod: {attachdetach-controller } SuccessfulAttachVolume: AttachVolume.Attach succeeded for volume "pvc-806abbbd-9380-11ea-9312-42010a140134" 
> May 11 12:21:12.848: INFO: At 2020-05-11 12:11:29 +0000 UTC - event for csi-rbd-demo-pod: {kubelet minikube} FailedMount: MountVolume.MountDevice failed for volume "pvc-806abbbd-9380-11ea-9312-42010a140134" : rpc error: code = Internal desc = rbd: map failed exit status 6, rbd output: rbd: sysfs write failed
> RBD image feature set mismatch. You can disable features unsupported by the kernel with "rbd feature disable replicapool/csi-vol-80a56ad0-9380-11ea-9c69-0242ac110010 journaling".
> ```
> 
> In case the attacher/rbd-plugin needs to write to `/sys/...`, it may need additional configuration in the deployment spec.
> 
> (note that the _Semantic Pull Request_ bot is not enforced, failing is accepted - and we don't understand yet why it fails)

looks deploy mounts /sys [here](https://github.com/ceph/ceph-csi/blob/master/deploy/rbd/kubernetes/csi-rbdplugin.yaml#L155)? But it's strange, I haven't changed the code. I think I need to build a test environment again and fix it...

internal/rbd/rbd_util.go
```diff
@@ -575,29 +595,19 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
...
-			}
-		}
-		rbdVol.ImageFeatures = imageFeatures
+	rbdVol.ImageFeatures = volOptions["imageFeatures"]
+	if rbdVol.Mounter, ok = volOptions["mounter"]; ok {
+		rbdVol.Mounter = rbdDefaultMounter
 	}
+	if err = validateImageFeatures(rbdVol.ImageFeatures, rbdVol.Mounter); err != nil {
+		// for backward compatibility, just log and pass
+		klog.Errorf(err.Error())
```
Madhu-1, line 604: logging and passing is not a good idea.

internal/rbd/rbd_util.go
```diff
@@ -621,6 +631,28 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
 	return rbdVol, nil
 }
 
+func validateImageFeatures(imageFeatures, mounter string) error {
+	arr := strings.Split(imageFeatures, ",")
+	featureSet := sets.NewString(arr...)
+	for _, f := range arr {
+		sf, found := supportedFeatures[f]
+		if !found {
+			return errors.Errorf("invalid feature %s for volume csi-rbdplugin", f)
```
Madhu-1, line 640:
```suggestion
return errors.Errorf("invalid feature %s for %s", f,reqName)
```
please pass  request name for more  clarity

Madhu-1:
>I0511 12:16:37.634510    2911 rbd_util.go:606] ID: 72 Req-ID: 0001-0024-b7b15817-390a-4571-b21f-ca6e7f7a502b-0000000000000001-1758ca5c-9380-11ea-90d0-0242ac110011 setting rbd image features: layering,journaling,exclusive-lock, mounter: rbd 

the issue here is its  using rbd mounted for image features `layering,journaling,exclusive-lock` is it something supported?

Madhu-1:
moved this out of release-v3.0.0 milestone

dimm0:
> moved this out of release-v3.0.0 milestone

Why!?

Madhu-1:
> > moved this out of release-v3.0.0 milestone
> 
> Why!?

As this PR is stale for at least last 2 months moved it to next release -v3.1.0

Madhu-1:
@woohhan do you have free time to work on this one? or shall I take it over and mention your name as a co-author?

woohhan:
@Madhu-1 Could you please take it over? unfortunately i have no time to proceed... thanks

Madhu-1:
> @Madhu-1 Could you please take it over? unfortunately i have no time to proceed... thanks

@woohhan Thanks for the confirmation :+1:

Madhu-1:
closing this one in favor of https://github.com/ceph/ceph-csi/pull/1325

PR #689, opened 2019-10-16
 Delete IDE releated files in .gitignore

WoohyungHan:
    The IDE is a user preference, and they should configure in
    individual envrionment. Delete IDE releated files in the
    gitignore.

NO

## Related issues ##
NO

## Future concerns ##
NO

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,18 +4,6 @@
 # docker build
 /deploy/cephcsi/image/cephcsi
 
-# Emacs save files
-*~
-\#*\#
-.\#*
-
-# Vim-related files
-[._]*.s[a-w][a-z]
-[._]s[a-w][a-z]
-*.un~
-Session.vim
-.netrwhist
-
 # git merge files
 *.orig
 *.patch


.gitignore
```diff
@@ -20,3 +20,6 @@ Session.vim
 *.orig
 *.patch
 *.rej
+
+# Idea(goland) files
+.idea/
```
Madhu-1: can you add `.vscode/` also?

WoohyungHan: Done!

.gitignore
```diff
@@ -20,3 +20,7 @@ Session.vim
 *.orig
 *.patch
 *.rej
+
+# goland, vscode
```
Madhu-1: can you please update the comment like `IDE related files (golang,vscode)`

WoohyungHan: oh, sorry about detail... done!

Madhu-1:
> I would prefer users to add these files to their `$HOME/git/ignore` or similar personal configuration. See https://git-scm.com/docs/gitignore for details.

in that case, we need to remove `emac and vim related files from gitignore`

WoohyungHan:
> > I would prefer users to add these files to their `$HOME/git/ignore` or similar personal configuration. See https://git-scm.com/docs/gitignore for details.
> 
> in that case, we need to remove `emac and vim related files from gitignore`

It looks better to me.

nixpanic:
On Wed, Oct 16, 2019 at 03:46:04AM -0700, Madhu Rajanna wrote:
> > I would prefer users to add these files to their `$HOME/git/ignore` or similar personal configuration. See https://git-scm.com/docs/gitignore for details.
> 
> in that case, we need to remove `emac and vim related files from gitignore`

Yes, I agree. I do not think it makes sense to add directories and files
for all possible IDEs. The IDE is a user preference, and they should
configure their git environment according to their preferences.

WoohyungHan:
> > I would prefer users to add these files to their `$HOME/git/ignore` or similar personal configuration. See https://git-scm.com/docs/gitignore for details.
> 
> in that case, we need to remove `emac and vim related files from gitignore`

Could I delete it or not?

Madhu-1:
> > > I would prefer users to add these files to their `$HOME/git/ignore` or similar personal configuration. See https://git-scm.com/docs/gitignore for details.
> > 
> > 
> > in that case, we need to remove `emac and vim related files from gitignore`
> 
> Could I delete it or not?

please delete and update the PR and PR description and commit message

Issue #688, opened 2019-10-16
Running E2E with rook-version default value is not working


WoohyungHan:
# Describe the bug #
Running E2E with rook-version 'master' is not working.
Default setup of rook-version is 'master'. If we do not setup specific rook-version(like v1.1.0) the test try to execute with master version. 

# Environment details #
- ceph-csi/master
- rook/master

# Steps to reproduce #
Run 'go test ./e2e/'

# Actual results #
The test is waiting for 'rook-ceph-agent' and timeout.

# Expected behavior #

# Additional context #

Madhu-1:
yes, flex driver is disabled in rook in master, need to add skip wait for agent pod if the rook master is used

WoohyungHan:
@Madhu-1 
I'm not sure that using 'latest' or 'master' is good choice in the code. The meaning of 'latest' can be change over time, but code(commit) is not. how about explicitly set default version which is working in current code?

Madhu-1:
> @Madhu-1
> I'm not sure that using 'latest' or 'master' is good choice in the code. The meaning of 'latest' can be change over time, but code(commit) is not. how about explicitly set default version which is working in current code? and if code is updated, we can update default verison too.

we can change the default value and also if the rook version is less than `v1.1.0` we can skip the check for agent pods

WoohyungHan:
That's so true!

Madhu-1:
> > we can change the default value and also if the rook version is less than v1.1.0 we can skip the check for agent pods
> 
> That's so true!

@WoohyungHan would you like to send a patch?

WoohyungHan:
Sure, I will.

PR #691, opened 2019-10-17
Add skip wait for rook-ceph-agent when version > 1.1.0

WoohyungHan:
after version 1.1.0, flex driver is disabled and doesn't need to wait for rook-ceph-agent pod anymore. Add skip wait for rook-ceph-agent.

NOTHING

## Related issues ##
Fixes: #688 

## Future concerns ##
NOTHING

### DIFF
--- a/e2e/deploy-rook.go
+++ b/e2e/deploy-rook.go
@@ -2,6 +2,8 @@ package e2e
 
 import (
 	"fmt"
+	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -79,13 +81,50 @@ func deployOperator(c kubernetes.Interface) {
 func deployCluster(c kubernetes.Interface) {
 	opPath := fmt.Sprintf("%s/%s", rookURL, "cluster-test.yaml")
 	framework.RunKubectlOrDie("create", "-f", opPath)
-	err := waitForDaemonSets("rook-ceph-agent", rookNS, c, deployTimeout)
-	Expect(err).Should(BeNil())
+
+	// After rook-ceph v1.1.0, flex driver is disabled. We don't need to wait for rook-ceph-agent.
+	if !strings.EqualFold(RookVersion, "master") && isOlderRookVersionThan(RookVersion, "v1.1.1") {
+		err := waitForDaemonSets("rook-ceph-agent", rookNS, c, deployTimeout)
+		Expect(err).Should(BeNil())
+	}
+
 	opt := &metav1.ListOptions{
 		LabelSelector: "app=rook-ceph-mon",
 	}
-	err = checkCephPods(rookNS, c, 1, deployTimeout, opt)
+	err := checkCephPods(rookNS, c, 1, deployTimeout, opt)
 	Expect(err).Should(BeNil())
+}
+
+func isOlderRookVersionThan(targetVersion, compareToVersion string) bool {
+	rv := extractRookVersion(targetVersion)
+	cv := extractRookVersion(compareToVersion)
+
+	for i := 0; i < 3; i++ {
+		if rv[i] < cv[i] {
+			return true
+		} else if rv[i] > cv[i] {
+			return false
+		}
+	}
+
+	return false
+}
+
+// extract rook version that form is v1.3.2-beta
+func extractRookVersion(versionString string) []int {
+	reg := regexp.MustCompile(`^v(\d+).(\d+).(\d+)`)
+	parsedVersionString := reg.FindStringSubmatch(versionString)
+	Expect(len(parsedVersionString)).Should(BeNumerically(">=", 4))
+
+	var version []int
+	for i := 1; i < 4; i++ {
+		j, err := strconv.Atoi(parsedVersionString[i])
+		Expect(err).Should(BeNil())
+
+		version = append(version, j)
+	}
+
+	return version
 }
 
 func deployToolBox(c kubernetes.Interface) {


e2e/deploy-rook.go
```diff
@@ -79,12 +79,16 @@ func deployOperator(c kubernetes.Interface) {
 func deployCluster(c kubernetes.Interface) {
 	opPath := fmt.Sprintf("%s/%s", rookURL, "cluster-test.yaml")
 	framework.RunKubectlOrDie("create", "-f", opPath)
-	err := waitForDaemonSets("rook-ceph-agent", rookNS, c, deployTimeout)
-	Expect(err).Should(BeNil())
+
+	// After rook-ceph v1.1.0, flex driver is disabled. We don't need to wait for rook-ceph-agent.
+	if RookVersion <= "v1.1.0" {
```
Madhu-1: this does not hold true if rook version is master

WoohyungHan: wow that's right. I will fix this. thanks!

e2e/deploy-rook.go
```diff
@@ -79,12 +79,16 @@ func deployOperator(c kubernetes.Interface) {
 func deployCluster(c kubernetes.Interface) {
 	opPath := fmt.Sprintf("%s/%s", rookURL, "cluster-test.yaml")
 	framework.RunKubectlOrDie("create", "-f", opPath)
-	err := waitForDaemonSets("rook-ceph-agent", rookNS, c, deployTimeout)
-	Expect(err).Should(BeNil())
+
+	// After rook-ceph v1.1.0, flex driver is disabled. We don't need to wait for rook-ceph-agent.
+	if RookVersion != "master" && RookVersion <= "v1.1.0" {
```
Madhu-1: sorry I missed one more thing, this will not  hold good for beta version also like `v1.1.0-beta.1` we need to do extract Semantic version and check it

WoohyungHan: hey, I apologize I didn't notice these... I'll extract v1.1.0 part and check it. Thanks again!

WoohyungHan: > sorry I missed one more thing, this will not hold good for beta version also like `v1.1.0-beta.1` we need to do extract Semantic version and check it

Done! but i wonder that 
I wrote version comparing code in util.go, however how do you think about using go-version libary? (https://github.com/hashicorp/go-version) I'm now sure i could import new third-party library. If you doesn't care I will fix up to use that. (in that case need add to /vendor directory?)

WoohyungHan:
Hey, Thanks for your review and I appreciate your quick and detailed suggestion. 

> can you add a test case of `0.8.0` and `0.9.0-beta`?
Done!

> as we know the target version how about declaring it as variable instead of extracting?
As I understand, you mean no need to target version because of global var of RookVersion?
If I correct, Done!

> and return just bool value instead of int
Done!

> move this function to E2E folder as we don't need in actual code
Done!

Madhu-1:
@WoohyungHan CI failure PTAL

e2e/deploy-rook_test.go
```diff
@@ -0,0 +1,59 @@
+package e2e
```
Madhu-1: this need to be removed we are not doing unit testing for packages under E2E folder

WoohyungHan: done!

e2e/deploy-rook.go
```diff
@@ -79,15 +81,50 @@ func deployOperator(c kubernetes.Interface) {
...
 	}
-	err = checkCephPods(rookNS, c, 1, deployTimeout, opt)
+	err := checkCephPods(rookNS, c, 1, deployTimeout, opt)
 	Expect(err).Should(BeNil())
 }
 
+func isOrderRookVersionThan(targetVersion string, compareToVersion string) bool {
+	rv := extractRookVersion(targetVersion)
+	cv := extractRookVersion(compareToVersion)
+
```
Madhu-1: add validation for len of `rv` and `cv` without it we may get `index out of range`

e2e/deploy-rook.go
```diff
@@ -79,15 +81,50 @@ func deployOperator(c kubernetes.Interface) {
...
+// extract rook version that form is v1.3.2-beta
+func extractRookVersion(versionString string) []int {
+	reg := regexp.MustCompile(`^v(\d+).(\d+).(\d+)`)
+	parsedVersionString := reg.FindStringSubmatch(versionString)
+
+	var version []int
+	for i := 1; i < 4; i++ {
+		j, _ := strconv.Atoi(parsedVersionString[i])
+
```
Madhu-1: in case of error return  `[]int`  as  we are validating length above

e2e/deploy-rook.go
WoohyungHan: I added length check in extractRookVersion, and this function will return always same length.

e2e/deploy-rook.go
WoohyungHan: Thanks, Done!

WoohyungHan:
our ci looks broken because of golang version issue...

Madhu-1:
@WoohyungHan any update on this?

e2e/deploy-rook.go
```diff
@@ -79,15 +81,52 @@ func deployOperator(c kubernetes.Interface) {
...
 	opt := &metav1.ListOptions{
 		LabelSelector: "app=rook-ceph-mon",
 	}
-	err = checkCephPods(rookNS, c, 1, deployTimeout, opt)
+	err := checkCephPods(rookNS, c, 1, deployTimeout, opt)
 	Expect(err).Should(BeNil())
 }
 
+func isOrderRookVersionThan(targetVersion, compareToVersion string) bool {
```
nixpanic: Did you mean to call this function `isOlderRookVersionThan`?

nixpanic:
Re-running the CI test that failed, caused it to succeed now :+1:

e2e/deploy-rook.go
WoohyungHan: @nixpanic Oh, Thanks. It's my mistake. Change Done!

WoohyungHan:
> Re-running the CI test that failed, caused it to succeed now
I appreciate your reply!

WoohyungHan:
> @WoohyungHan any update on this?

@Madhu-1 Hi, Could you please check If I change some more? I think I already updated your suggestion. Thanks!

Issue #675, opened 2019-10-09
Add support for smart cloning in rbd


Madhu-1:
below are the sub-tasks to  achieve the smart cloning

* Implement stand-alone snapshot creation more info here (https://github.com/ceph/ceph-csi/pull/467)
* Implement rbd image flattening
*  Implement rbd cloning from a snapshot
* Implement cloning image from another image (combined with above 3 implementations)
*  Enable E2E testing for snapshot
* Add E2E testing for smart cloning

zhucan:
@Madhu-1 I don't understand it that restore from snapshot not use “rbd rollback” but use "rbd clone" and the restore from snapshot not support flatten， why？

Madhu-1:
we will support flatten with the new implementation

dillaman:
> @Madhu-1 I don't understand it that restore from snapshot not use “rbd rollback” but use "rbd clone" and the restore from snapshot not support flatten， why？

@zhucan From the point-of-view of k8s, the "restore" operation is to just create a new PV from the snapshot. Therefore, you *cannot* just rollback (plus it's very inefficient).

zhucan:
@dillaman Thanks， There is plan to do volume clone for ceph-csi?

dillaman:
@zhucan Indeed -- I believe that's what this issue is tracking.

zhucan:
@dillaman Thanks

Madhu-1:
I had a discussion with @dillaman since we are facing [issue](https://github.com/ceph/ceph-csi/pull/693#issuecomment-544950196) with the current approach. we are moving to a new implementation as detailed below
```
# CO create snapshot process
* create temporary RBD snapshot on source volume's RBD image
* RBD clone CO source volume's RBD image's snapshot to new RBD image
* remove temporary RBD snapshot on CO source volume's RBD image

# CO create volume from snapshot process
* serialize work on RBD snapshot image clone
* test RBD snapshot image clone's depth
* if depth over the hard limit (no soft-limit anymore), flatten clone
* create RBD snapshot on RBD snapshot image clone
* create RBD volume image clone from RBD snapshot image clone

# CO create volume from the volume process
* create RBD snapshot image as if running "CO create snapshot" process
* create RBD volume image as if running "CO create volume from snapshot" process
```

Madhu-1:
update snapshot from alpha to beta once we change the design for snapshot and cloning

Madhu-1:
This is fixed now closing this one

Issue #403, opened 2019-06-04
dump cephcsi container logs in E2E


Madhu-1:
dumping cephcsi  container logs in E2E will help us to analyze the failure

Madhu-1:
This is already fixed

PR #693, opened 2019-10-18
Add support for standalone snapshot creation and cloning support

Madhu-1:
Add support for standalone snapshot creation and creating new PVC from snapshot

with the current implementation in ceph-csi, it's not possible to delete the cloned volume if the snapshot is present due to the child linking, to remove this dependency we had a discussion (Thanks for @dillaman for suggesting this idea) and come up with an idea to separate out the clone and snapshot, so that we can delete the snapshot and cloned image in any order.

the steps followed to create an independent snapshot as follows

### Create a snapshot
* Create  a temporary snapshot from the parent volume
* Clone a new image from a temporary snapshot with options
`--rbd-default-clone-format 2 --image-feature layering,deep-flatten`
* Delete temporary snapshot created
* Create a snapshot with requested Name from cloned volume

### Create PVC from a snapshot
* Check the depth of the image as the maximum number of nested volume
clones can be (max 10 can be changed based on the configuration)
if the depth is reached flatten the newly cloned image

### Delete a snapshot
* Delete the temporary cloned image which was created for a snapshot
* Delete the snapshot

example commands:-
```
layering,deep-flatten <RBD image for src  k8s volume>@<random snap <RBD image for temporary snap image>
4) rbd snap create <RBD image for temporary snap image>@<random snap name> <k8s snap name>
5) rbd clone --rbd-default-clone-format 2 --image-feature <k8s dst vol config>
<RBD image for temporary snap image>@<random snap name> <RBD image for k8s dst vol>
```

Partial Fixes: #675 
Fixes: https://github.com/ceph/ceph-csi/issues/403
https://github.com/ceph/ceph-csi/pull/467 will be abandoned with this PR

Note: @dillaman I have addressed the review comment in https://github.com/ceph/ceph-csi/pull/467 PR PTAL


pkg/rbd/snapshot.go
```diff
@@ -0,0 +1,105 @@
...
+		delErr = fmt.Errorf("clone created but failed to delete snapshot due to"+" other failures: %v", delErr)
+		delErr = status.Error(codes.Internal, delErr.Error())
+		return delErr
+	}
+
+	return nil
+}
+
+func cleanUpSnapshot(ctx context.Context, rbdSnap *rbdSnapshot, rbdVol *rbdVolume, cr *util.Credentials) error {
+	err := deleteSnapshot(ctx, rbdSnap, cr)
```
dillaman: Is this the snapshot on the source volume RBD image? Shouldn't that have been deleted in the create step?

pkg/rbd/rbd_util.go
```diff
@@ -210,19 +235,51 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
-	klog.V(4).Infof(util.Log(ctx, "rbd: rm %s using mon %s, pool %s"), image, pOpts.Monitors, pOpts.Pool)
+	err = moveImageToTrash(ctx, pOpts, cr)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to move image %s to trash: %v"), image, err)
+		return err
+	}
 
-	// attempt to use Ceph manager based deletion support if available
-	rbdCephMgrSupported, err := rbdManagerTaskDeleteImage(ctx, pOpts, cr)
+	id, err = getImageIDfromTrash(ctx, pOpts, cr)
```
dillaman: This seams potentially dangerous. It seems like it would be safer to record the image id in the omap tables at CO snapshot / RBD image creation time or prior to moving to the trash.

pkg/rbd/rbd_util.go
```diff
@@ -552,32 +592,13 @@ func createSnapshot(ctx context.Context, pOpts *rbdSnapshot, cr *util.Credential
...
-}
-
 func deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot, cr *util.Credentials) error {
 	var output []byte
 
 	image := pOpts.RbdImageName
 	snapName := pOpts.RbdSnapName
 
-	klog.V(4).Infof(util.Log(ctx, "rbd: snap rm %s using mon %s, pool %s"), image, pOpts.Monitors, pOpts.Pool)
+	klog.V(4).Infof(util.Log(ctx, "rbd: snap rm %s@%s using mon %s, pool %s"), image, snapName, pOpts.Monitors, pOpts.Pool)
```
dillaman: Nit: use image-spec style

pkg/rbd/rbd_util.go
```diff
@@ -601,12 +616,11 @@ func restoreSnapshot(ctx context.Context, pVolOpts *rbdVolume, pSnapOpts *rbdSna
 	image := pVolOpts.RbdImageName
 	snapName := pSnapOpts.RbdSnapName
 
-	klog.V(4).Infof(util.Log(ctx, "rbd: clone %s using mon %s, pool %s"), image, pVolOpts.Monitors, pVolOpts.Pool)
+	klog.V(4).Infof(util.Log(ctx, "rbd: clone %s@%s %s using mon %s, pool %s"), pSnapOpts.RbdImageName, snapName, image, pVolOpts.Monitors, pVolOpts.Pool)
```
dillaman: Nit: use image-spec style

pkg/rbd/rbd_util.go
```diff
@@ -813,3 +840,81 @@ func cleanupRBDImageMetadataStash(path string) error {
...
+// asynchronously in the background. If command is not found returns a bool set to false
+func rbdManagerTaskAddFlatten(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) (bool, error) {
+	var output []byte
+
+	klog.V(4).Infof(util.Log(ctx, "rbd: task add flatten for image (%s) using mon %s, pool %s"), pOpts.RbdImageName, pOpts.Monitors, pOpts.Pool)
+	args := []string{"rbd", "task", "add", "flatten", pOpts.Pool + "/" + pOpts.RbdImageName, "--id", cr.ID, "--keyfile=" + cr.KeyFile, "-m", pOpts.Monitors}
+
+	output, err := execCommand("ceph", args)
+	if err != nil {
+		if strings.Contains(string(output), rbdTaskRemoveCmdInvalidString1) &&
```
dillaman: Nit: can we move these to a common helper?

pkg/util/voljournal.go
```diff
@@ -261,6 +261,7 @@ func (cj *CSIJournal) UndoReservation(ctx context.Context, monitors string, cr *
 		cj.csiNameKeyPrefix+reqName)
 	if err != nil {
 		klog.Errorf(Log(ctx, "failed removing oMap key %s (%s)"), cj.csiNameKeyPrefix+reqName, err)
+		// TODO return nil if OMap key not found
```
dillaman: TODO

Madhu-1:
> Looks like a good start. One concern that I couldn't easily comment on inline since it wasn't a changed line was the structure of `rbdSnapshot` and how `RbdImageName` sometimes points to the original volume and sometimes to the clone. I think that it should always refer to the snapshot clone image name -- and therefore `RbdSnapName` would always be the same as `RbdImageName` in the case of this new v2 snapshot implementation. For the legacy CSI snapshots, they could be different.

rbdSnapshot refers to original image only during the first time when we  need to reserve the snapshot in Omap once its done it  will be point to the cloned image name

will  add support to remove older snapshot

pkg/util/voljournal.go
Madhu-1: this is not part of this PR, would like to take it up later

dillaman:
> rbdSnapshot refers to original image only during the first time when we need to reserve the snapshot in Omap once its done it will be point to the cloned image name
> 
> will add support to remove older snapshot

Except you are also passing the `rbdVol` for the source at the same time. Therefore, there is no need to have the confusion w/ the name switching back-and-forth.

Madhu-1:
> > rbdSnapshot refers to original image only during the first time when we need to reserve the snapshot in Omap once its done it will be point to the cloned image name
> > will add support to remove older snapshot
> 
> Except you are also passing the `rbdVol` for the source at the same time. Therefore, there is no need to have the confusion w/ the name switching back-and-forth.

setting RBDImageName from source volume [here](https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/controllerserver.go#L441) because we are generating  the snapshotID and reserving snapshot details in Omap [here](https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/controllerserver.go#L474) prior to this we don't have the snapshotID to set the RbdImageName in snapshot struct, once we do the reservation we are setting back  to clone volume  [here](https://github.com/ceph/ceph-csi/pull/693/files#diff-35c77cc181d49675cd67a6095520b41fR542) to create a  final snapshot from cloned volume

if am  missing anything or doing it wrong let  me know

pkg/rbd/snapshot.go
```diff
@@ -0,0 +1,105 @@
...
+			if err = deleteSnapshot(ctx, snap, cr); err != nil {
+				klog.Errorf(util.Log(ctx, "failed to cleanup snapshot: %v"), err)
+			}
+		}
+	}()
+RESTORE:
+	// create clone image and delete snapshot
+	err = restoreSnapshot(ctx, cloneRbdVol, snap, cr)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
```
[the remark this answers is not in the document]
Madhu-1: :+1:  Nice idea this can be done

pkg/rbd/snapshot.go
Madhu-1: This is snapshot from cloned volume image

dillaman: ... so it should have already been deleted above?

dillaman:
> if am missing anything or doing it wrong let me know

If you need it for reserving the snapshot, pass the source volume struct instead. This is for clarity to avoid the back-and-forth.

pkg/rbd/rbd_util.go
Madhu-1: do I need to record the image ID? or can I fetch the image ID from image prior to moving it to trash?

pkg/rbd/rbd_util.go
```diff
@@ -813,3 +840,81 @@ func cleanupRBDImageMetadataStash(path string) error {
...
+
+	depth, err := getCloneDepth(ctx, rbdVol.Monitors, rbdVol.Pool, rbdVol.RbdImageName, 0, cr)
+	if err != nil {
+		return err
+	}
+
+	klog.Infof(util.Log(ctx, "image depth is %v and maximum configured clone depth hard limit is %v and soft limit is %v"), depth, hardmaxDepth, softmaxDepth)
+	if depth >= int(hardmaxDepth) {
+		klog.Infof(util.Log(ctx, "maximum clone depth (%d) has been reached, flattening %v volume "), hardmaxDepth, rbdVol.RbdImageName)
+		args := []string{"-m", rbdVol.Monitors,
```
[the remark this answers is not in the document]
Madhu-1: CSI spec does not mention anything related to  in-progress response

pkg/rbd/rbd_util.go
dillaman: I guess it depends if volume/snapshot IDs are 100% guaranteed to be unique for the lifetime of the CO. If, for example, we had a volume/snapshot named "X" that the CO deleted and we moved to the trash but cannot delete, can the CO ever create a new volume named "X"? If there is no guarantee that these CO names cannot be reused, then for safety we would need to record the image id prior to removing the image.

... but thinking about that some more, I guess we would just immediately fail in the create step since we would see existing omap entries for the volume/snapshot the CO thought it had deleted and then attempted to re-use?

pkg/rbd/rbd_util.go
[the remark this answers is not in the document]
dillaman: I thought CephFS was (or was planning) to send some type of "in-progress" when it was cloning (i.e. slowly copying)?

pkg/rbd/rbd_util.go
Madhu-1: we are generating some unique ID for each PVC name and we create a mapping for unique ID and PVC name if the image deletion fails we don't remove the mapping from the Omap,  if we remove it from Omap means the image is already deleted.
even if the unique ID and PVC name mapping is not present in omap  we generate a new ID and use it as image name, I think it should be fine

pkg/rbd/rbd_util.go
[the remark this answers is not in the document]
Madhu-1: I don't think that is possible. need a second pair of Eyes  can you please confirm  it once https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume

pkg/rbd/rbd_util.go
dillaman: Yeah, but we have an ask out to the external-provisioner sidecar to provide a deterministic name to the CSI based on the namespace and PVC name. If you then delete a PVC (and the CSI keeps it behind the scenes) and then re-create it, you are saying the CSI will generate a unique internal volume id? If so, I think the current implementation is fine. It could be more efficient to just look up the image id before moving it to the trash, but then you would have unique logic on the recovery path where it needs to scan the full trash which might have hundreds or thousands of images. The middle ground would be to record the id before moving it to the trash to avoid the full trash list and search.

Madhu-1:
@dillaman addressed review comments PTAL

pkg/rbd/controllerserver.go
```diff
@@ -625,21 +631,21 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 	}
 	defer cs.SnapshotLocks.Release(rbdSnap.RequestName)
 
-	// Unprotect snapshot
-	err = unprotectSnapshot(ctx, rbdSnap, cr)
```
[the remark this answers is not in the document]
Madhu-1: done

pkg/rbd/snapshot.go
Madhu-1: we are calling this function if we fail to get snapshot information after creating the snapshot, I think this should be fine

pkg/rbd/rbd_util.go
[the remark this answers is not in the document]
dillaman: Looking at the external-provisioner sidecar [1], it looks like you can return numerous errors to tell it it's pending [2]


[1] https://github.com/openshift/csi-external-provisioner/blob/master/pkg/controller/controller.go#L542
[2] https://github.com/openshift/csi-external-provisioner/blob/master/pkg/controller/controller.go#L1139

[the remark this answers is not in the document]
dillaman: ... specifically the `ABORTED` error code looks like it should be used

pkg/rbd/snapshot.go
dillaman: ... but the actual "RBD" snapshot on the source volume RBD image was already deleted, so what does this do? There isn't an "RBD" snapshot on the CO snapshot RBD image.

pkg/rbd/rbd_util.go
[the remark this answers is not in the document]
Madhu-1: yeah, this can be used, working late night forgot about ABORT at all :D

once we reach the hard limit we need to check the flatten status right (if it's not done need to  return Abort error message)? can you provide me the command to check task status?

pkg/rbd/snapshot.go
Madhu-1: this is called here https://github.com/ceph/ceph-csi/pull/693/files#diff-35c77cc181d49675cd67a6095520b41fR559 if we fail to get snapshot information (we have deleted the temp snapshot created from source RBD volume but  not the snapshot created from cloned rbd image)

pkg/rbd/rbd_util.go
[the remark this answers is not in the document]
Madhu-1: I think I can use `ceph rbd task list`  if the rbd image name is  present on this list means the task is not completed and have to return  `ABORT` error message
if the task is completed  means return  success
@dillaman let me know if my assumption is correct will implement this.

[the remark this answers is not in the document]
dillaman: Once flatten is complete, you wouldn't even enter this check, right? Since you would have gotten the clone depth and it should return 0. Otherwise, feel free to keep running "ceph rbd task add flatten" since it will just keep returning the original task id back to you if it already existed (i.e. no need to parse the list I don't think).

pkg/rbd/snapshot.go
dillaman: Ah, I missed the second "rbd snap create" (`createSnapshot` call) against the new RBD clone image that represents the CO snapshot. Apologies.

Madhu-1:
@dillaman still am facing the mounting issue for cloned PVC below are the details

```
Linux version 4.15.0-1040-gcp (buildd@lgw01-amd64-044) (gcc version 7.4.0 (Ubuntu 7.4.0-1ubuntu1~18.04.1)) #42-Ubuntu SMP Wed Aug 7 15:17:54 UTC 2019
```

```
W1019 05:16:16.699187   24874 rbd_attach.go:225] ID: 101 Req-ID: 0001-0024-3df72d6e-396a-4789-a181-a5aa0b38e2c8-0000000000000001-8c2e5b6d-f22f-11e9-9ac1-0242ac11000c rbd: map error exit status 6, rbd output: rbd: sysfs write failed
RBD image feature set mismatch. You can disable features unsupported by the kernel with "rbd feature disable replicapool/csi-vol-8c2e5b6d-f22f-11e9-9ac1-0242ac11000c".
In some cases useful info is found in syslog - try "dmesg | tail".
E1019 05:16:16.699657   24874 utils.go:161] ID: 101 Req-ID: 0001-0024-3df72d6e-396a-4789-a181-a5aa0b38e2c8-0000000000000001-8c2e5b6d-f22f-11e9-9ac1-0242ac11000c GRPC error: rpc error: code = Internal desc = rbd: map failed exit status 6, rbd output: rbd: sysfs write failed
RBD image feature set mismatch. You can disable features unsupported by the kernel with "rbd feature disable replicapool/csi-vol-8c2e5b6d-f22f-11e9-9ac1-0242ac11000c".
In some cases useful info is found in syslog - try "dmesg | tail".
```

but in my local testing on kernel versions `5.3.7-1.el7.elrepo.x86_64` tests passed

Madhu-1:
@dillaman PTAL let me know if anything is pending here

@nixpanic what would be kernel version in centos CI?

nixpanic:
Currently the kernel on the CentOS-7 has version 3.10.0-1062.4.1.el7 (see the [updates here](http://mirror.centos.org/centos/7/updates/x86_64/Packages/)). I do not know what parts related to Ceph have been backported. You would need to check the changelog of the package or check the [online version from RHEL](https://access.redhat.com/downloads/content/kernel/3.10.0-1062.4.1.el7/x86_64/fd431d51/package-changelog).

Madhu-1:
> Currently the kernel on the CentOS-7 has version 3.10.0-1062.4.1.el7 (see the [updates here](http://mirror.centos.org/centos/7/updates/x86_64/Packages/)). I do not know what parts related to Ceph have been backported. You would need to check the changelog of the package or check the [online version from RHEL](https://access.redhat.com/downloads/content/kernel/3.10.0-1062.4.1.el7/x86_64/fd431d51/package-changelog).

Thanks for the info @nixpanic, anyway we can create a vagrant based kube cluster, any idea when centos 8 libvirt vagrant box will be available? i tried to find some info https://bugs.centos.org/view.php?id=16455

Madhu-1:
Even we can upgrade the kernel version in Vm created using vagrant.  @nixpanic any idea when can we have E2E setup on centos CI?

dillaman:
@Madhu-1  Ouch -- I thought krbd passed the optional read-only flag [1] to the `get_features` call since that would have let the older kernels use deep-flatten on the snapshots since those are only opened in read-only mode. I think w/o that support, we are going to need to force flattening of snapshots prior to the creation of any RBD snapshots on the CO snapshot backing RBD image.

@idryomov It's too late to help us now, but can we get that flag added to krbd so that in the future read-write image features don't affect read-only images?

[1] https://github.com/ceph/ceph/blob/master/src/cls/rbd/cls_rbd.cc#L862

dillaman:
... kernel v5.1 supports the deep-flatten feature and perhaps we can get the read-only feature check integrated into a future RHEL 8.x release, so in the meantime we can add a parameter to opt-in to fast (zero-copy) snapshot cloning.

Madhu-1:
> ... kernel v5.1 supports the deep-flatten feature and perhaps we can get the read-only feature check integrated into a future RHEL 8.x release, so in the meantime we can add a parameter to opt-in to fast (zero-copy) snapshot cloning.

@dillaman can you give some info whats needs to be done now?

idryomov:
> @idryomov It's too late to help us now, but can we get that flag added to krbd so that in the future read-write image features don't affect read-only images?

Sure.  I even filed a ticket for that in 2018: https://tracker.ceph.com/issues/23688.

Madhu-1:
@dillaman PTAL

pkg/rbd/rbd_util.go
```diff
@@ -813,3 +812,59 @@ func cleanupRBDImageMetadataStash(path string) error {
...
+	if depth >= int(maxDepth) {
+		klog.Infof(util.Log(ctx, "maximum clone depth (%d) has been reached, flattening %v volume "), maxDepth, rbdVol.RbdImageName)
+		args := []string{"-m", rbdVol.Monitors,
+			"--id", cr.ID,
+			"--pool", rbdVol.Pool,
+			"--no-progress",
+			"--image", rbdVol.RbdImageName,
+			"--keyfile=" + cr.KeyFile,
+			"-c", util.CephConfigPath,
+			"flatten"}
```
dillaman: We should still be using `ceph rbd task add flatten` if possible

pkg/rbd/rbd_util.go
```diff
@@ -586,67 +609,43 @@ func deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot, cr *util.Credential
...
 	snapName := pSnapOpts.RbdSnapName
 
-	klog.V(4).Infof(util.Log(ctx, "rbd: clone %s using mon %s, pool %s"), image, pVolOpts.Monitors, pVolOpts.Pool)
+	klog.V(4).Infof(util.Log(ctx, "rbd: clone %s/%s@%s %s using mon %s"), pVolOpts.Pool, pSnapOpts.RbdImageName, snapName, image, pVolOpts.Monitors)
 	args := []string{"clone", pSnapOpts.Pool + "/" + pSnapOpts.RbdImageName + "@" + snapName,
-		pVolOpts.Pool + "/" + image, "--id", cr.ID, "-m", pVolOpts.Monitors, "--keyfile=" + cr.KeyFile}
+		pVolOpts.Pool + "/" + image, "--rbd-default-clone-format=2", "--id", cr.ID, "-m", pVolOpts.Monitors, "--keyfile=" + cr.KeyFile, "--image-feature", pVolOpts.ImageFeatures}
 
 	output, err := execCommand("rbd", args)
-
```
dillaman: Nit: perhaps should add flatten depth check here and start a background flatten if required and return `ABORT` error if flatten is needed. Also add in the `CreateSnapshot` path prior to creating the snapshot on the RBD volume image (due to the lack of deep-flatten). Basically, attempt to flatten (if required by depth) both prior to creating any RBD snapshots on an image and after cloning an RBD image.

pkg/rbd/controllerserver.go
```diff
@@ -228,20 +227,48 @@ func (cs *ControllerServer) checkSnapshot(ctx context.Context, req *csi.CreateVo
 		return status.Error(codes.Internal, err.Error())
 	}
 	defer cr.DeleteCredentials()
-
+	SnapVol := &rbdVolume{}
 	rbdSnap := &rbdSnapshot{}
-	if err = genSnapFromSnapID(ctx, rbdSnap, snapshotID, cr); err != nil {
-		if _, ok := err.(ErrSnapNotFound); !ok {
+	if err = genSnapFromSnapID(ctx, rbdSnap, snapshotID, false, cr); err != nil {
```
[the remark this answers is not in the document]
Madhu-1: added locking for snapshotID

pkg/rbd/controllerserver.go
```diff
@@ -228,20 +227,48 @@ func (cs *ControllerServer) checkSnapshot(ctx context.Context, req *csi.CreateVo
...
+
+		}
+		if err != nil {
 			return status.Error(codes.Internal, err.Error())
 		}
-		return status.Error(codes.InvalidArgument, "missing requested Snapshot ID")
 	}
 
-	err = restoreSnapshot(ctx, rbdVol, rbdSnap, cr)
+	err = flattenRbdImage(ctx, SnapVol, rbdMaxCloneDepth, cr)
```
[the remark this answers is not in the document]
Madhu-1: do we need to support it? as we have backward  compatibility for snapshot deletion and also snapshot is an alpha feature

pkg/rbd/controllerserver.go
```diff
@@ -625,21 +639,62 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
...
+		klog.Errorf(util.Log(ctx, "failed to remove reservation for snapname (%s) with backing snap (%s) on image (%s) (%s)"),
+			rbdSnap.RequestName, rbdSnap.RbdSnapName, rbdSnap.RbdImageName, err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	return &csi.DeleteSnapshotResponse{}, nil
+}
+
+func isLegacySnapshot(ctx context.Context, rbdSnap *rbdSnapshot, cr *util.Credentials) (bool, error) {
+	err := unprotectSnapshot(ctx, rbdSnap, cr)
+	if err != nil {
```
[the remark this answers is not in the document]
Madhu-1: is there a way to identify it?

pkg/rbd/rbd_util.go
Madhu-1: can we add this support once we add a soft limit?

pkg/rbd/controllerserver.go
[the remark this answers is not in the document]
Madhu-1: added code to populate `snapVol`

pkg/rbd/rbd_util.go
Madhu-1: little confused here, are you suggesting to  flatten the source  volume before creating a snapshot and flatten the cloned  image after creating a clone from snapshot during CreateSnapshot

is there a way to find out if the flattening in progress for an image?

dillaman: It's for the case where you have VOL1 (img1) -> SNAP1 (img2) ->VOL2 (img3) -> SNAP2 (img4). We can hit the hard limit at either the snapshot creation or the volume creation.

 If the depth is >= the hard limit, it implies it needs to be flattened and you can schedule it via the background task and `ABORT` so that it backs off. If you fail trying to add the background flatten and need to fall-back to the blocking `rbd flatten`, you could try-lock on affected image and if it is in-progress (lock busy) you can again return `ABORT`

pkg/rbd/rbd_util.go
dillaman: We really don't want to be blocking the CSI -- that's why we added the background task stuff to begin with.

pkg/rbd/controllerserver.go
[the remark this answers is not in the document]
dillaman: EBUSY return code?

pkg/rbd/controllerserver.go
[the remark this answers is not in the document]
dillaman: Fair enough -- I'll let other weigh in. If we effectively prevent the recovery from legacy k8s snapshots, however, it should be well documented.

pkg/rbd/rbd_util.go
Madhu-1: @dillaman one concern I have here is, as the volume creation failed with `Abort` error message if the flatten is already in progress as sidecar doesn't send the delete request for the operation in progress(if the PVC delete request is sent), this can lead us to stale backend images

dillaman: What image are you concerned about? The one about to be created (that hasn't been created yet if you need to flatten its parent)?

Madhu-1: @dillaman parent in the sense cloned volume during snapshot creation, right? if yes how can we handle creating PVC from PVC ( adding background task can leave stale snapshot and clone image) and also just to add here we are already returning `ABORT` error message if  volume operation in progress here https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/controllerserver.go#L138

dillaman: If the CO is free to treat `CreateVolume` `ABORT` errors as if there is nothing to clean up, we have a bigger issue since we could be leaking images all over the place. The CO needs to treat errors from the CSI as an unknown state (maybe the backing PV was created, maybe it wasn't) and force the `DeleteVolume` call to clean up. If this is not happening, I think it's a bug in the CSI spec.

Madhu-1: yes As  there is no intermediate step between CreateVolume and CreateVolume success, if we return ABORT error message during CreateVolumeRequest as there is no VolumeID is returned if the User sends the PVC delete  request in between as sidecar doesn't have the volumeID to send DeleteRequest the backend image will go stage

dillaman: Right now if you return `ABORT` it will just keep trying [1]. It seems odd that it would allow you to `DeleteVolume` against a PVC that is in-progress. Have you tested it to see if it allows you to do that -- or perhaps it just hangs waiting for `CreateVolume` to succeed (or non-retryable fail) before permitting the `DeleteVolume`.

[1] https://github.com/openshift/csi-external-provisioner/blob/master/pkg/controller/controller.go#L545

Madhu-1:
@dillaman updated PR PTAL

pkg/rbd/rbd_util.go
dillaman: Did you ever get a chance to clarify if this is a potential for image leaking?

pkg/rbd/rbd_util.go
dillaman: ... I think this comment is still valid.

pkg/rbd/rbd_util.go
```diff
@@ -813,3 +827,148 @@ func cleanupRBDImageMetadataStash(path string) error {
...
+			"--no-progress",
+			"--image", rbdVol.RbdImageName,
+			"--keyfile=" + cr.KeyFile,
+			"-c", util.CephConfigPath,
+			"flatten"}
+
+		_, err := execCommand("rbd", args)
+		return err
+	}
+	if softmaxDepth > 0 && depth >= int(softmaxDepth) {
```
dillaman: Nit: 
```suggestion
	if depth >= int(softmaxDepth) {
```

pkg/rbd/rbd_journal.go
```diff
@@ -123,40 +122,55 @@ func checkSnapExists(ctx context.Context, rbdSnap *rbdSnapshot, cr *util.Credent
...
-		if _, ok := err.(ErrSnapNotFound); ok {
+		if _, ok := err.(ErrImageNotFound); ok {
 			err = snapJournal.UndoReservation(ctx, rbdSnap.Monitors, cr, rbdSnap.Pool,
 				rbdSnap.RbdSnapName, rbdSnap.RequestName)
 			return false, err
 		}
 		return false, err
 	}
 
+	ok, err := checkIsFlattenInProgress(ctx, vol, cr)
```
dillaman: This seems like you should just be calling `flattenRbdImage`. If the image is at the soft-snap limit it will just (re)schedule the background task to get it flattened. If it's at the hard-limit, we should `ABORT` while the flatten is running in the background (or block only if you cannot schedule the background flatten).

pkg/rbd/rbd_journal.go
```diff
@@ -183,9 +197,19 @@ func checkVolExists(ctx context.Context, rbdVol *rbdVolume, cr *util.Credentials
 				rbdVol.RbdImageName, rbdVol.RequestName)
 			return false, err
 		}
+
 		return false, err
 	}
-
+	if clone {
+		ok, flatErr := checkIsFlattenInProgress(ctx, rbdVol, cr)
```
dillaman: Same comment here -- if we are restarting a create volume request, we just re-use the existing flatten logic.

pkg/rbd/controllerserver.go
```diff
@@ -230,18 +244,43 @@ func (cs *ControllerServer) checkSnapshot(ctx context.Context, req *csi.CreateVo
...
 
 	rbdSnap := &rbdSnapshot{}
-	if err = genSnapFromSnapID(ctx, rbdSnap, snapshotID, cr); err != nil {
-		if _, ok := err.(ErrSnapNotFound); !ok {
+	if acquired := cs.SnapshotLocks.TryAcquire(snapshotID); !acquired {
+		klog.Infof(util.Log(ctx, util.SnapshotOperationAlreadyExistsFmt), snapshotID)
+		return status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, snapshotID)
+	}
+	defer cs.SnapshotLocks.Release(snapshotID)
+
```
dillaman: Need to enforce flattening limits prior to the creation of the RBD snapshot below since you cannot flatten after the snapshot it created (i.e. it might already be running the background=).

ShyamsundarR:
Still working through the PR review, but here are some points to consider:

1) I would suggest that we use the ready_to_use [1] flag/parameter in the response for a CreateSnapshot to denote a flatten in progress, and that the snapshot is not yet usable.

IOW, the flow is as below (for now considering only success cases),
- create rbd snap for image
- create rbd clone of snapshot
- check if flatten is required:
  - yes: start a flatten and return ready_to_use as false (the CO will retry to finally determine when snap is ready to use)
  - no: return ready_to_use as true

A snapshot is not ready to use, so we will/should not see a case where a PVC provides such a snapshot as the volume content source. This helps the CreateVolume case, as we will not deal with snapshot volumes (IOW, rbd clones) that are in the process of being flattened. Also, the CO cannot use such snapshots for other operations as it is not ready.

A subsequent snapshot of the same image may be requested, at which point we need to determine that a flatten is in progress and bail out OR, if we can we again snap->clone->flatten-new-clone->return ready_to_use as false.

For the case when the volume content source is another volume, we snap the image and then clone it, post which we decide if the depth is larger than tolerance, at which point a flatten is started. For such cases, if a flatten is required, we should check and start a flatten first, and return ABORTED for the CreateVolume request. IOW, allowing a snap->clone only when it will not violate the depth. (This is what @dillaman seems to suggesting as well)

2) The image leaking discussion in comments [2] as it relates to [1]

Considering we implement (1) above, the 2 cases for image leaks are as follows,

a) CreateSnapshot:
- In this case the snap->clone is already created and the CO has the VolumeID returned back for any future operation (like DeleteVolume). IOW, we have passed a success back and the CO now can take future actions as it sees fit. There should not be a case where this is leaked.

- If we choose to bail out of CreateSnapshot as a flatten is in progress (although we should be flattening the clone, so base image should not return said status if I understand right), we can return an error from a class of non-retry errors (IOW, **not** ABORT/EAGAIN sets), to state that a snapshot cannot be taken at this time.

b) CreateVolume from VolumeContentSource that is a PVC (IOW, **not** a snapshot source)
- If we determine the snap->clone is going to violate the depth, we start a flatten, as stated in (1) and return ABORT
- There is nothing to clean up here as the image is being flattened, and that needs to be done irrespective for future operations to succeed
- I assume that the image chain while being flattened can still be use to perform IO

So, in both cases there are no image/snap/clone leaks that I can foresee.

NOTE: CreateSnapshot may not be retried, if it fails the first time, as the applications may be frozen to create said snapshot and halting applications for longer duration's may not be a viable manner of progress. The CO does have the responsibility, as per the spec, to retry the operation till it gets a "final" error/success, and take appropriate action. This is being fixed/addressed here [3].

[1] ready_to_use: https://github.com/container-storage-interface/spec/blob/master/spec.md#the-ready_to_use-parameter
[2] image leak and when should an image be created: https://github.com/ceph/ceph-csi/pull/693#discussion_r338205980
[3] Snapshot cleanup on timeouts/retry errors: `https://github.com/kubernetes-csi/external-snapshotter/issues/134`

ShyamsundarR:
@Madhu-1 I have an overarching comment on the transaction of creating a csi-snapshot, using the snap->clone method as follows.

The transaction flows like so,

1) Regular programming:
  a) Reserve
  b) Create snap
  c) Create clone
  d) Delete snap
  e) Flatten clone
    - return ready_to_use:false OR ABORT (under discussion); if flatten **task** is started
    - return success; if no flatten

2) Retry for ABORT/ready_to_use and or other RETRY errors due to intermittent failures:
  - Check reservation (this is done in the code in checkSnapCloneExists)
    - Checks till it finds the clone, ensuring prior transaction was run to completion
      - If the clone is not found, indicates a failed transaction, and we need to check and roll back all steps d-to-a
        - Currently the code (checkSnapCloneExists) is only undoing the transaction, and that can leave behind a snapshot due to step (d) not completed
      - If clone is found, checks if flatten is required, and starts one or ensures one is running, returning required state
        - The above covers the case where a transaction was complete till (d) and a irrecoverable failure occurred post the same

Once we reach a state where it is not a retry as above, we have a clean slate, and we also take the csi-snapname lock to ensure non-racy transactions, and so can degrade/rely on the "regular programming" to do its job correctly.

What the above means, is that in the function createRBDClone we do not need to check each step, but rather just perform the required steps. Failure of createRBDClone would result in a similar backtracking of undoing all steps d-to-a (like, in the retry case when the image is not found).

Thus, if we either had a single undo of the steps leading to an cloned image, we can invoke the same based when we UndoReservation to ensure there are no leaked snaps and keys.

NOTE: Also, switching the delete of snap to before flattening the clone is desirable, as it makes the transaction a little easier to handle

3) DeleteSnapshot on ready_to_use being false (as CO will have the SnapID to issue the operation)

This will follow a similar undo like above, except that the undo also needs to address any in-flight flatten that was invoked (cancelling the same).

If we could organize the code and functions like above, it would be easier to comprehend and reason, as it stands.

pkg/rbd/rbd_util.go
Madhu-1: verified this once ( will test it a few more times). if the driver creates the rbd image and returns the ABORT error message as flattening in progress. if the PVC is deleted before flattening completes, CreateVolume is not retried till it gets success response nor DeleteVolume will not be called by CSI provisioner as it doesn't have the volume ID (we have an issue on image leaking for smart volume cloning)

Madhu-1:
@dillaman @ShyamsundarR PTAL

Madhu-1:
@ShyamsundarR @dillaman PTAL

dillaman:
What's the status of snapshot / clone support for the v2.1 release?

humblec:
@Madhu-1 can you revisit this PR ?

Madhu-1:
@humblec am already working on this one, already updated the PR (need to do more work)

humblec:
[Status update]
Last week we had a meeting around snapshot/clone design/implementation we introduced recently with this PR, to support further use cases like Bakcup/DR , Data replication..etc. The main reason behind this meeting was to avoid a revamp again with snapshot/clone implementation while we start supporting these use cases too in near future. Atm, we are listing down the requirements to support it along with possible changes/enhancements we need with current implementation/design. Further technical notes will be followed/pointed here to those discussion. Considering we are at code freeze of `v2.1.0` and the state we are with this features, the decision is to defer it to next month release (`v3.0.0` https://github.com/ceph/ceph-csi/issues/865 )  and release `v2.1.0` as planned with other bug fixes and enhancements.

ShyamsundarR:
> [Status update]
> Last week we had a meeting around snapshot/clone design/implementation we introduced recently with this PR, to support further use cases like Bakcup/DR , Data replication..etc. The main reason behind this meeting was to avoid a revamp again with snapshot/clone implementation while we start supporting these use cases too in near future. Atm, we are listing down the requirements to support it along with possible changes/enhancements we need with current implementation/design. Further technical notes will be followed/pointed here to those discussion. Considering we are at code freeze of `v2.1.0` and the state we are with this features, the decision is to defer it to next month release (`v3.0.0` #865 ) and release `v2.1.0` as planned with other bug fixes and enhancements.

Details of other use-cases is present [here](https://gist.github.com/ShyamsundarR/6650152aa20ccac952bf1de9dd661e00).

The decision is to continue with the current design and implementation as it does not impede the future uses as understood at present.

ShyamsundarR:
@dillaman an image whose ancestor is undergoing a flatten is still mountable by the clients, right?

e.g, if the image chain is, image1 -> image2 (CSI-snap) -> image3 (CSI-clone), when cloning from a CSI-Snapshot. image2 could be undergoing a flatten (either due to soft or hard flatten limits), whereas image3 could be mounted and used in the interim.

An image that is undergoing flattening is not mountable by the clients, right? As a result our base rule is always flatten the images that serve as the intermediate CSI-snapshot

If the above is true, when cloning from an CSI-Volume rather than from a CSI-Snapshot, where the current image chain would be image1 (source) -> image2 (RBD clone serving as hidden CSI-snapshot) -> image3 (RBD clone serving as the CSI-clone), could we flatten `image2` assuming `image3` is at the hard limit of clone depth, but still return a success for the operation?

dillaman:
> e.g, if the image chain is, image1 -> image2 (CSI-snap) -> image3 (CSI-clone), when cloning from a CSI-Snapshot. image2 could be undergoing a flatten (either due to soft or hard flatten limits), whereas image3 could be mounted and used in the interim.

Yes, but only if at the "soft" limit. If the total chain length is >15, krbd cannot properly handle opening that image with 14 ancestors. If your entire chain is >=15, we need to pause the creation of the new PVC that violates the limit.

Now, we could always just choose to restrict the chain depth upon k8s snapshot creation by just taking the k8s snapshot chain length + 1 (i.e. assume it will get a parent eventually). That would allow you to use the "snapshot creation pending / uploading" status return while it's being flattened and therefore prevent its use via a new PVC from snapshot operation.

dillaman:
@ShyamsundarR I went down the rabbit hole on this one this afternoon. In order to support thin-provisioned clones and PVCs from snapshots, we require kernel v5.1 or later (RHEL 8.2 or later). Any platform that attempts to use this functionality on older kernels (I won't name names), will need to fully flatten the clone/PVC from snapshot prior to giving access to the end-user. That means we will need to support flatten checks upon PVC creation (i.e. max image chain length is 1 -- no parents are supported).

ShyamsundarR:
> @ShyamsundarR I went down the rabbit hole on this one this afternoon. In order to support thin-provisioned clones and PVCs from snapshots, we require kernel v5.1 or later (RHEL 8.2 or later). Any platform that attempts to use this functionality on older kernels (I won't name names because it's very silly that they are still using such an old OS), will need to fully flatten the clone/PVC from snapshot prior to giving access to the end-user. That means we will need to support flatten checks upon PVC creation (i.e. max image chain length is 1 -- no parents are supported).

Here is what I understand what CSI must do,
- Add image feature deep-flatten to RBD images representing CSI-snapshots (includes intermediate snapshot image created for a CSI-clone from CSI-volume)
  - These intermediate images are never mapped, and hence provide enough exclusivity during flatten, to ensure RBD-snapshots of these images are also flattened
- Set (or provide documentation) hard limit for depth as 1, for setups using kernel versions older than v5.1
  - This should prevent any change in the implementation, and help with setups having older kernel versions
- For kernel versions at or newer than v5.1 hard/soft limits can be set higher

Thoughts/Questions:
- It is technically feasible to use rbd-nbd, where kernel versions are older than v5.1 to leverage a higher hard/soft limit
  - Barring the pod eviction requirements, required at present, when updating the nodeplugin pods

ShyamsundarR:
> > @ShyamsundarR I went down the rabbit hole on this one this afternoon. In order to support thin-provisioned clones and PVCs from snapshots, we require kernel v5.1 or later (RHEL 8.2 or later). Any platform that attempts to use this functionality on older kernels (I won't name names because it's very silly that they are still using such an old OS), will need to fully flatten the clone/PVC from snapshot prior to giving access to the end-user. That means we will need to support flatten checks upon PVC creation (i.e. max image chain length is 1 -- no parents are supported).
> 
> Here is what I understand what CSI must do,
> 
>     * Add image feature deep-flatten to RBD images representing CSI-snapshots (includes intermediate snapshot image created for a CSI-clone from CSI-volume)
>       
>       * These intermediate images are never mapped, and hence provide enough exclusivity during flatten, to ensure RBD-snapshots of these images are also flattened
> 
>     * Set (or provide documentation) hard limit for depth as 1, for setups using kernel versions older than v5.1
* When hard limit is 1, do not flatten intermediate images that represent CSI-Snapshots, this reduces one round of flattening (reduces space consumption as well) and keeps CSI-snapshot operation relatively quick. **NOTE:** Intermediate images will be flattened when per image hard/soft snapshot limits are reached though #1098 .
* This is an optimization for older kernels, as discussed with @Madhu-1 and as pointed by @dillaman above ("will need to fully flatten the clone/PVC from snapshot prior to giving access to the end-user"), but stating it more explicitly here
>       
>       * This should prevent any change in the implementation, and help with setups having older kernel versions
> 
>     * For kernel versions at or newer than v5.1 hard/soft limits can be set higher
> 
> 
> Thoughts/Questions:
> 
>     * It is technically feasible to use rbd-nbd, where kernel versions are older than v5.1 to leverage a higher hard/soft limit
>       
>       * Barring the pod eviction requirements, required at present, when updating the nodeplugin pods

dillaman:
> When hard limit is 1, do not flatten intermediate images that represent CSI-Snapshots, this reduces one round of flattening (reduces space consumption as well) and keeps CSI-snapshot operation relatively quick. **NOTE:** Intermediate images will be flattened when per image hard/soft snapshot limits are reached though #1098 .

... or just have a node kernel capability flag in the RBD controller saying deep-flatten is supported or not. If it's not supported, always flatten the PVC image.

Madhu-1:
This will be closed by https://github.com/ceph/ceph-csi/pull/1160. keeping this open till #1160 is merged

PR #697, opened 2019-10-23
Adds persistent volume claims encryption with LUKS

vasyl-purchel:
Adds encryption in StorageClass as a parameter. Encryption passphrase is stored in kubernetes secrets per StorageClass. Implements rbd volume encryption relying on dm-crypt and cryptsetup
using LUKS extension

The change is related to [proposal](https://github.com/ceph/ceph-csi/pull/448) made earlier. This is a first part of the full feature that adds encryption with passphrase stored in secrets.

Signed-off-by: Vasyl Purchel <vasyl.purchel@workday.com>
Signed-off-by: Andrea Baglioni <andrea.baglioni@workday.com>
Signed-off-by: Ioannis Papaioannou <ioannis.papaioannou@workday.com>
Signed-off-by: Paul Mc Auley <paul.mcauley@workday.com>
Signed-off-by: Sergio de Carvalho <sergio.carvalho@workday.com>

### DIFF
--- a/docs/deploy-rbd.md
+++ b/docs/deploy-rbd.md
@@ -54,6 +54,7 @@ make image-cephcsi
 | `csi.storage.k8s.io/provisioner-secret-name`, `csi.storage.k8s.io/node-stage-secret-name`           | yes (for Kubernetes) | name of the Kubernetes Secret object containing Ceph client credentials. Both parameters should have the same value                                                                                                     |
 | `csi.storage.k8s.io/provisioner-secret-namespace`, `csi.storage.k8s.io/node-stage-secret-namespace` | yes (for Kubernetes) | namespaces of the above Secret objects                                                                                                                                                                                  |
 | `mounter`                                                                                           | no                   | if set to `rbd-nbd`, use `rbd-nbd` on nodes that have `rbd-nbd` and `nbd` kernel modules to map rbd images                                                                                                              |
+| `encrypted`                                                                                         | no                   | disabled by default, use `"true"` to enable LUKS encryption on pvc and `"false"` to disable it. **Do not change for existing storageclasses**                                                                           |
 
 **NOTE:** An accompanying CSI configuration file, needs to be provided to the
 running pods. Refer to [Creating CSI configuration](../examples/README.md#creating-csi-configuration)
@@ -155,3 +156,59 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
 **Deploy Helm Chart:**
 
 [See the Helm chart readme for installation instructions.](../charts/ceph-csi-rbd/README.md)
+
+## Encryption for RBD volumes
+
+> Enabling encryption on volumes created without encryption is **not supported**
+>
+> Enabling encryption for storage class that has PVs created without encryption
+> is **not supported**
+
+Volumes provisioned with Ceph RBD do not have encryption by default. It is
+possible to encrypt them with ceph-csi by using LUKS encryption.
+
+To enable encryption set `encrypted` option in storage class to `"true"` and
+set encryption passphrase in kubernetes secrets under `encryptionPassphrase` key.
+
+To use different passphrase you need to have different storage classes and point
+to a different K8s secrets (different `csi.storage.k8s.io/node-stage-secret-name`
+and `csi.storage.k8s.io/node-stage-secret-namespace`).
+
+### Life-cycle for encrypted volumes
+
+**Create volume**:
+
+* create volume request received
+* volume requested to be created in Ceph
+* encrypted state "requiresEncryption" is saved in image-meta in Ceph
+
+**Attach volume**:
+
+* attach volume request received
+* volume is attached to provisioner container
+* on first time attachment
+  (no file system on the attached device, checked with blkid)
+  * device is encrypted with LUKS using a passphrase from K8s secrets
+  * image-meta updated to "encrypted" in Ceph
+* device is open and device path is changed to use a mapper device
+* mapper device is used instead of original one with usual workflow
+
+**Detach volume**:
+
+* mapper device closed and device path changed to original volume path
+* volume is detached as usual
+
+### Encryption configuration
+
+To encrypt rbd volumes with LUKS you need to set encryption passphrase in
+secrets under `encryptionPassphrase` key and switch `encrypted` option in
+StorageClass to `"true"`. This is not supported for storage classes that already
+have PVs provisioned.
+
+### Encryption prerequisites
+
+In order for encryption to work you need to make sure that `dm-crypt` kernel
+module is enabled on the nodes running ceph-csi attachers.
+
+If custom image is built for the rbd-plugin instance, make sure that it contains
+`cryptsetup` tool installed to be able to use encryption.
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"regexp"
 	"strings"
 	"time"
 
@@ -20,6 +21,7 @@ import (
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/wait"
 	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
@@ -553,7 +555,7 @@ func deletePVCAndApp(name string, f *framework.Framework, pvc *v1.PersistentVolu
 	return err
 }
 
-func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
+func createPVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) (*v1.PersistentVolumeClaim, *v1.Pod) {
 	pvc, err := loadPVC(pvcPath)
 	if pvc == nil {
 		Fail(err.Error())
@@ -572,11 +574,84 @@ func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
 		Fail(err.Error())
 	}
 
+	return pvc, app
+}
+
+func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
+	pvc, app := createPVCAndAppBinding(pvcPath, appPath, f)
+	err := deletePVCAndApp("", f, pvc, app)
+	if err != nil {
+		Fail(err.Error())
+	}
+}
+
+func getRBDImageSpec(pvcNamespace, pvcName string, f *framework.Framework) (string, error) {
+	c := f.ClientSet.CoreV1()
+	pvc, err := c.PersistentVolumeClaims(pvcNamespace).Get(pvcName, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	pv, err := c.PersistentVolumes().Get(pvc.Spec.VolumeName, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+
+	imageIDRegex := regexp.MustCompile(`(\w+\-?){5}$`)
+	imageID := imageIDRegex.FindString(pv.Spec.CSI.VolumeHandle)
+	return fmt.Sprintf("replicapool/csi-vol-%s", imageID), nil
+}
+
+func getImageMeta(rbdImageSpec, metaKey string, f *framework.Framework) (string, error) {
+	cmd := fmt.Sprintf("rbd image-meta get %s %s", rbdImageSpec, metaKey)
+	opt := metav1.ListOptions{
+		LabelSelector: "app=rook-ceph-tools",
+	}
+	stdOut, stdErr := execCommandInPod(f, cmd, rookNS, &opt)
+	if stdErr != "" {
+		return strings.TrimSpace(stdOut), fmt.Errorf(stdErr)
+	}
+	return strings.TrimSpace(stdOut), nil
+}
+
+func getMountType(appName, appNamespace, mountPath string, f *framework.Framework) (string, error) {
+	opt := metav1.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector("metadata.name", appName).String(),
+	}
+	cmd := fmt.Sprintf("lsblk -o TYPE,MOUNTPOINT | grep '%s' | awk '{print $1}'", mountPath)
+	stdOut, stdErr := execCommandInPod(f, cmd, appNamespace, &opt)
+	if stdErr != "" {
+		return strings.TrimSpace(stdOut), fmt.Errorf(stdErr)
+	}
+	return strings.TrimSpace(stdOut), nil
+}
+
+func validateEncryptedPVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
+	pvc, app := createPVCAndAppBinding(pvcPath, appPath, f)
+
+	rbdImageSpec, err := getRBDImageSpec(pvc.Namespace, pvc.Name, f)
+	if err != nil {
+		Fail(err.Error())
+	}
+	encryptedState, err := getImageMeta(rbdImageSpec, ".rbd.csi.ceph.com/encrypted", f)
+	if err != nil {
+		Fail(err.Error())
+	}
+	Expect(encryptedState).To(Equal("encrypted"))
+
+	volumeMountPath := app.Spec.Containers[0].VolumeMounts[0].MountPath
+	mountType, err := getMountType(app.Name, app.Namespace, volumeMountPath, f)
+	if err != nil {
+		Fail(err.Error())
+	}
+	Expect(mountType).To(Equal("crypt"))
+
 	err = deletePVCAndApp("", f, pvc, app)
 	if err != nil {
 		Fail(err.Error())
 	}
 }
+
 func deletePodWithLabel(label string) error {
 	_, err := framework.RunKubectl("delete", "po", "-l", label)
 	if err != nil {
--- a/examples/rbd/secret.yaml
+++ b/examples/rbd/secret.yaml
@@ -10,3 +10,6 @@ stringData:
   # specified in the storage class
   userID: <plaintext ID>
   userKey: <Ceph auth key corresponding to ID above>
+
+  # Encryption passphrase
+  encryptionPassphrase: test_passphrase
--- a/examples/rbd/storageclass.yaml
+++ b/examples/rbd/storageclass.yaml
@@ -38,6 +38,11 @@ parameters:
    csi.storage.k8s.io/fstype: ext4
    # uncomment the following to use rbd-nbd as mounter on supported nodes
    # mounter: rbd-nbd
+
+   # Instruct the plugin it has to encrypt the volume
+   # By default it is disabled. Valid values are “true” or “false”.
+   # A string is expected here, i.e. “true”, not true.
+   # encrypted: "true"
 reclaimPolicy: Delete
 allowVolumeExpansion: true
 mountOptions:
--- a/pkg/rbd/controllerserver.go
+++ b/pkg/rbd/controllerserver.go
@@ -149,6 +149,14 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if found {
+		if rbdVol.Encrypted {
+			err = ensureEncryptionMetadataSet(ctx, cr, rbdVol)
+			if err != nil {
+				klog.Errorf(util.Log(ctx, err.Error()))
+				return nil, err
+			}
+		}
+
 		return &csi.CreateVolumeResponse{
 			Volume: &csi.Volume{
 				VolumeId:      rbdVol.VolID,
@@ -174,6 +182,20 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	err = cs.createBackingImage(ctx, rbdVol, req, util.RoundOffVolSize(rbdVol.VolSize))
 	if err != nil {
 		return nil, err
+	}
+
+	if rbdVol.Encrypted {
+		err = ensureEncryptionMetadataSet(ctx, cr, rbdVol)
+		if err != nil {
+			klog.Errorf(util.Log(ctx, "failed to save encryption status, deleting image %s"),
+				rbdVol.RbdImageName)
+			if deleteErr := deleteImage(ctx, rbdVol, cr); err != nil {
+				klog.Errorf(util.Log(ctx, "failed to delete rbd image: %s/%s with error: %v"),
+					rbdVol.Pool, rbdVol.RbdImageName, deleteErr)
+				return nil, deleteErr
+			}
+			return nil, err
+		}
 	}
 
 	return &csi.CreateVolumeResponse{
@@ -211,6 +233,7 @@ func (cs *ControllerServer) createBackingImage(ctx context.Context, rbdVol *rbdV
 
 	return nil
 }
+
 func (cs *ControllerServer) checkSnapshot(ctx context.Context, req *csi.CreateVolumeRequest, rbdVol *rbdVolume) error {
 	snapshot := req.VolumeContentSource.GetSnapshot()
 	if snapshot == nil {
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -87,24 +87,13 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	}
 	defer ns.VolumeLocks.Release(volID)
 
-	isLegacyVolume := false
-	volName, err := getVolumeName(volID)
-	if err != nil {
-		// error ErrInvalidVolID may mean this is an 1.0.0 version volume, check for name
-		// pattern match in addition to error to ensure this is a likely v1.0.0 volume
-		if _, ok := err.(ErrInvalidVolID); !ok || !isLegacyVolumeID(volID) {
-			return nil, status.Error(codes.InvalidArgument, err.Error())
-		}
-
-		volName, err = getLegacyVolumeName(req.GetStagingTargetPath())
-		if err != nil {
-			return nil, status.Error(codes.InvalidArgument, err.Error())
-		}
-		isLegacyVolume = true
-	}
-
 	stagingParentPath := req.GetStagingTargetPath()
 	stagingTargetPath := stagingParentPath + "/" + volID
+
+	isLegacyVolume, volName, err := getVolumeNameByID(volID, stagingParentPath)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
 
 	var isNotMnt bool
 	// check if stagingPath is already mounted
@@ -123,6 +112,7 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	volOptions.RbdImageName = volName
+	volOptions.VolID = req.GetVolumeId()
 
 	isMounted := false
 	isStagePathCreated := false
@@ -145,7 +135,15 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-	klog.V(4).Infof(util.Log(ctx, "rbd image: %s/%s was successfully mapped at %s\n"), req.GetVolumeId(), volOptions.Pool, devicePath)
+	klog.V(4).Infof(util.Log(ctx, "rbd image: %s/%s was successfully mapped at %s\n"),
+		req.GetVolumeId(), volOptions.Pool, devicePath)
+
+	if volOptions.Encrypted {
+		devicePath, err = ns.processEncryptedDevice(ctx, volOptions, devicePath, cr, req.GetSecrets())
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	}
 
 	err = ns.createStageMountPoint(ctx, stagingTargetPath, isBlock)
 	if err != nil {
@@ -194,7 +192,7 @@ func (ns *NodeServer) undoStagingTransaction(ctx context.Context, stagingParentP
 
 	// Unmapping rbd device
 	if devicePath != "" {
-		err = detachRBDDevice(ctx, devicePath)
+		err = detachRBDDevice(ctx, devicePath, volID)
 		if err != nil {
 			klog.Errorf(util.Log(ctx, "failed to unmap rbd device: %s for volume %s with error: %v"), devicePath, volID, err)
 			// continue on failure to delete the stash file, as kubernetes will fail to delete the staging path otherwise
@@ -271,18 +269,6 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 	klog.Infof(util.Log(ctx, "rbd: successfully mounted stagingPath %s to targetPath %s"), stagingPath, targetPath)
 	return &csi.NodePublishVolumeResponse{}, nil
-}
-
-func getVolumeName(volID string) (string, error) {
-	var vi util.CSIIdentifier
-
-	err := vi.DecomposeCSIID(volID)
-	if err != nil {
-		err = fmt.Errorf("error decoding volume ID (%s) (%s)", err, volID)
-		return "", ErrInvalidVolID{err}
-	}
-
-	return volJournal.NamingPrefix() + vi.ObjectUUID, nil
 }
 
 func getLegacyVolumeName(mountPath string) (string, error) {
@@ -521,7 +507,7 @@ func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 
 	// Unmapping rbd device
 	imageSpec := imgInfo.Pool + "/" + imgInfo.ImageName
-	if err = detachRBDImageOrDeviceSpec(ctx, imageSpec, true, imgInfo.NbdAccess); err != nil {
+	if err = detachRBDImageOrDeviceSpec(ctx, imageSpec, true, imgInfo.NbdAccess, req.GetVolumeId()); err != nil {
 		klog.Errorf(util.Log(ctx, "error unmapping volume (%s) from staging path (%s): (%v)"), req.GetVolumeId(), stagingTargetPath, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
@@ -612,4 +598,111 @@ func (ns *NodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetC
 			},
 		},
 	}, nil
+}
+
+func (ns *NodeServer) processEncryptedDevice(ctx context.Context, volOptions *rbdVolume, devicePath string, cr *util.Credentials, secrets map[string]string) (string, error) {
+	imageSpec := volOptions.Pool + "/" + volOptions.RbdImageName
+	encrypted, err := util.CheckRbdImageEncrypted(ctx, cr, volOptions.Monitors, imageSpec)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to get encryption status for rbd image %s: %v"),
+			imageSpec, err)
+		return "", err
+	}
+
+	if encrypted == rbdImageRequiresEncryption {
+		diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: mount.NewOsExec()}
+		// TODO: update this when adding support for static (pre-provisioned) PVs
+		var existingFormat string
+		existingFormat, err = diskMounter.GetDiskFormat(devicePath)
+		if err != nil {
+			return "", fmt.Errorf("failed to get disk format for path %s, error: %v", devicePath, err)
+		}
+		if existingFormat != "" {
+			return "", fmt.Errorf("can not encrypt rbdImage %s that already has file system: %s",
+				imageSpec, existingFormat)
+		}
+		err = encryptDevice(ctx, volOptions, secrets, cr, devicePath)
+		if err != nil {
+			return "", fmt.Errorf("failed to encrypt rbd image %s: %v", imageSpec, err)
+		}
+	} else if encrypted != rbdImageEncrypted {
+		return "", fmt.Errorf("rbd image %s found mounted with unexpected encryption status %s",
+			imageSpec, encrypted)
+	}
+
+	devicePath, err = openEncryptedDevice(ctx, volOptions, devicePath, secrets)
+	if err != nil {
+		return "", err
+	}
+
+	return devicePath, nil
+}
+
+func encryptDevice(ctx context.Context, rbdVol *rbdVolume, secret map[string]string, cr *util.Credentials, devicePath string) error {
+	passphrase, err := util.GetCryptoPassphrase(secret)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to get crypto passphrase for %s/%s: %v"),
+			rbdVol.Pool, rbdVol.RbdImageName, err)
+		return err
+	}
+
+	if err = util.EncryptVolume(ctx, devicePath, passphrase); err != nil {
+		err = fmt.Errorf("failed to encrypt volume %s/%s: %v", rbdVol.Pool, rbdVol.RbdImageName, err)
+		klog.Errorf(util.Log(ctx, err.Error()))
+		return err
+	}
+
+	imageSpec := rbdVol.Pool + "/" + rbdVol.RbdImageName
+	err = util.SaveRbdImageEncryptionStatus(ctx, cr, rbdVol.Monitors, imageSpec, rbdImageEncrypted)
+
+	return err
+}
+
+func openEncryptedDevice(ctx context.Context, volOptions *rbdVolume, devicePath string, secrets map[string]string) (string, error) {
+	passphrase, err := util.GetCryptoPassphrase(secrets)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to get passphrase for encrypted device %s/%s: %v"),
+			volOptions.Pool, volOptions.RbdImageName, err)
+		return "", status.Error(codes.Internal, err.Error())
+	}
+
+	mapperFile, mapperFilePath := util.VolumeMapper(volOptions.VolID)
+
+	isOpen, err := util.IsDeviceOpen(ctx, mapperFilePath)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to check device %s encryption status: %s"), devicePath, err)
+		return devicePath, err
+	}
+	if isOpen {
+		klog.V(4).Infof(util.Log(ctx, "encrypted device is already open at %s"), mapperFilePath)
+	} else {
+		err = util.OpenEncryptedVolume(ctx, devicePath, mapperFile, passphrase)
+		if err != nil {
+			klog.Errorf(util.Log(ctx, "failed to open device %s/%s: %v"),
+				volOptions.Pool, volOptions.RbdImageName, err)
+			return devicePath, err
+		}
+	}
+
+	return mapperFilePath, nil
+}
+
+func getVolumeNameByID(volID, stagingTargetPath string) (bool, string, error) {
+	volName, err := getVolumeName(volID)
+	if err != nil {
+		// error ErrInvalidVolID may mean this is an 1.0.0 version volume, check for name
+		// pattern match in addition to error to ensure this is a likely v1.0.0 volume
+		if _, ok := err.(ErrInvalidVolID); !ok || !isLegacyVolumeID(volID) {
+			return false, "", status.Error(codes.InvalidArgument, err.Error())
+		}
+
+		volName, err = getLegacyVolumeName(stagingTargetPath)
+		if err != nil {
+			return false, "", status.Error(codes.InvalidArgument, err.Error())
+		}
+
+		return true, volName, nil
+	}
+
+	return false, volName, nil
 }
--- a/pkg/rbd/rbd_attach.go
+++ b/pkg/rbd/rbd_attach.go
@@ -225,7 +225,7 @@ func createPath(ctx context.Context, volOpt *rbdVolume, cr *util.Credentials) (s
 		klog.Warningf(util.Log(ctx, "rbd: map error %v, rbd output: %s"), err, string(output))
 		// unmap rbd image if connection timeout
 		if strings.Contains(err.Error(), rbdMapConnectionTimeout) {
-			detErr := detachRBDImageOrDeviceSpec(ctx, imagePath, true, isNbd)
+			detErr := detachRBDImageOrDeviceSpec(ctx, imagePath, true, isNbd, volOpt.VolID)
 			if detErr != nil {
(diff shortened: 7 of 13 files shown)


pkg/rbd/rbd_attach.go
```diff
@@ -260,21 +260,38 @@ func waitForrbdImage(ctx context.Context, backoff wait.Backoff, volOptions *rbdV
...
 // when imageSpec is used to decide if image is already unmapped
-func detachRBDImageOrDeviceSpec(ctx context.Context, imageOrDeviceSpec string, isImageSpec, ndbType bool) error {
-	var err error
+func detachRBDImageOrDeviceSpec(ctx context.Context, imageOrDeviceSpec string, isImageSpec, ndbType bool, volumeID string) error {
 	var output []byte
 
+	mapperFile, mapperPath := util.VolumeMapper(volumeID)
+	mappedDevice, mapper, err := util.DeviceEncryptionStatus(mapperPath)
+	if err != nil {
+		klog.Warningf(util.Log(ctx, "error determining LUKS device on %s, %s: %s"),
```
Madhu-1: instead of warning log as an error here and also where ever you are returning an error  message

pkg/rbd/rbd_util.go
```diff
@@ -64,22 +66,24 @@ type rbdVolume struct {
...
+	ImageFeatures        string `json:"imageFeatures"`
+	AdminID              string `json:"adminId"`
+	UserID               string `json:"userId"`
+	Mounter              string `json:"mounter"`
+	ClusterID            string `json:"clusterId"`
+	RequestName          string
+	VolName              string `json:"volName"`
+	MonValueFromSecret   string `json:"monValueFromSecret"`
+	PassphraseFromSecret string `json:"passphraseFromSecret"`
+	VolSize              int64  `json:"volSize"`
```
dillaman: Nit: why did you move these two fields (this line and below) around?

pkg/rbd/rbd_util.go
```diff
@@ -486,7 +490,23 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
 		rbdVol.Mounter = rbdDefaultMounter
 	}
 
-	return rbdVol, nil
+	rbdVol.Encrypted = false
+	encrypted, ok := volOptions["encrypted"]
+	if ok {
+		rbdVol.Encrypted, err = strconv.ParseBool(encrypted)
+		if err != nil {
+			rbdVol.Encrypted = false
```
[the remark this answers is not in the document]
dillaman: Nit: if this parse fails due to a typo, it seems like it would be nicer to fail instead of silently disabling encryption.

pkg/rbd/controllerserver.go
```diff
@@ -177,6 +177,14 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, err
 	}
 
+	if rbdVol.Encrypted {
+		// format the image with cryptsetup
+		if err = encryptDevice(ctx, rbdVol, req.GetSecrets(), cr); err != nil {
```
dillaman: I think it would be a good idea to persistently record that the image is encrypted. This will be needed for properly handling (or disabling) snapshots and cloning operations.

examples/encrypted_rbd/secret.yaml
```diff
@@ -0,0 +1,16 @@
...
+stringData:
+  # Key values correspond to a user name and its key, as defined in the
+  # ceph cluster. User ID should have required access to the 'pool'
+  # specified in the storage class
+  userID: <plaintext ID>
+  userKey: <Ceph auth key corresponding to ID above>
+
+  # Key value is a base64 encoded encryption passphrase. Key name is defined
+  # in storageclass
+  ceph-csi-encrypt-passphrase: dGVzdA==
```
dillaman: Should this be in a `data` section instead of `stringData`?

pkg/rbd/controllerserver.go
vasyl-purchel: what do I need to add to persist this information? I thought the volOptions structure is saved in configmaps

examples/encrypted_rbd/csi-rbdplugin.yaml
```diff
@@ -0,0 +1,171 @@
+---
```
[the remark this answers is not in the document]
vasyl-purchel: the only addition is to make provisioner to run on a host network (which requires changing the ports to not overlap with attacher)
but I have not tested the upgrade procedure as I'm not sure what is the usual upgrade process is used

examples/encrypted_rbd/storageclass.yaml
```diff
@@ -0,0 +1,47 @@
...
+   csi.storage.k8s.io/node-stage-secret-name: csi-rbd-secret
+   csi.storage.k8s.io/node-stage-secret-namespace: default
+   # Specify the filesystem type of the volume. If not specified,
+   # csi-provisioner will set default as `ext4`.
+   csi.storage.k8s.io/fstype: ext4
+   # uncomment the following to use rbd-nbd as mounter on supported nodes
+   # mounter: rbd-nbd
+
+   # Instruct the plugin it has to encrypt the volume
+   encrypted: "true"
```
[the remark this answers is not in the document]
vasyl-purchel: I need to double check but adding should be allowed if "encrypted" field is persisted for the volume. but existing volumes are not going to get encrypted, this is only for a newly created ones

pkg/rbd/controllerserver.go
```diff
@@ -643,3 +652,71 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
...
+		err        error
+		devicePath string
+	)
+
+	if !volOptions.Encrypted {
+		klog.V(4).Infof(util.Log(ctx, "Device %s is not encrypted, skipping encrypt map"), volOptions.VolID)
+		return nil
+	}
+
+	if len(volOptions.RbdImageName) == 0 {
```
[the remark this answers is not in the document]
vasyl-purchel: I can remove check and leave the next line only as from testing it on actual cluster I found that name doesn't appear to be set at this point but it is required for the attach process

pkg/rbd/rbd_util.go
vasyl-purchel: this was due to a failure in travis - it mentions that the size of the structure can be reduced if rearrange the fields

pkg/rbd/controllerserver.go
dillaman: > what do I need to add to persist this information? 

You can store it in the RBD image-meta fields via `rbd image-meta set <key> <value>`, where perhaps `key` is something like `.rbd.csi.ceph.com/encrypted`. Then snapshot/clone operations can check for this and disallow for now.

> I thought the volOptions structure is saved in configmaps

Not to my knowledge -- that might have been the legacy behaviour.

pkg/rbd/nodeserver.go
```diff
@@ -146,6 +147,26 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	}
 	klog.V(4).Infof(util.Log(ctx, "rbd image: %s/%s was successfully mapped at %s\n"), req.GetVolumeId(), volOptions.Pool, devicePath)
 
+	if volOptions.Encrypted {
```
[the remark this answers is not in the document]
vasyl-purchel: could you recommend where this documentation should be added? from what I can see existing deploy-rbd.md only goes through simple case without any branching or specifics of some options. Should I add a new document for this?

pkg/rbd/controllerserver.go
vasyl-purchel: would another field in CSIJournal be the right place for this? https://github.com/ceph/ceph-csi/blob/8e117db80d197c46e4205352caada92b8e490085/pkg/util/voljournal.go#L97

I don't see `image-meta` to be used anywhere else, but if journal is not a right place for such kind of information I can save it in image-meta

dillaman: The `CSIJournal` is a directory used for mapping RBD <-> CO names so I don't think it would be a good place to store other attributes.

vasyl-purchel: ok, in that case I'll use image-meta for this as you've suggested

ShyamsundarR:
I would like to propose the following change to the design/implementation here:

Assumptions:
- Initial encryption can be performed by the nodeserver rather than the controller server, IOW it is acceptable the the node that is going to mount the encrypted device also performs the initial encryption setup.

Revision:
1) Perform initial encryption post attachRBDImage called from NodeStageVolume
  - This needs to be conditional to the rbd image not having been formatted prior to this
  - Leverage suggestion by @dillaman to stash a `rbd image-meta set <key> <value>` that denotes initial encryption is done or not

2) Perform OpenEncryptedVolume subsequently in NodeStageVolume
  - This is as done in the current version of the code

Why:
- Avoid the controller plugin from having to map and encrypt the image
- Keep all map related requirements within the scope of the nodeplugin, and leverage initial map that needs to be performed anyway, to further encrypt the image

vasyl-purchel:
@ShyamsundarR, we have thought about such approach, but it is not safe enough to encrypt volumes when attaching.

If for any reason volume was provisioned as not encrypted and it got some data on it, the last thing you want is to loose that data. If we encrypt with LUKS a volume on a first attach, it is possible that we will format the disk that was not encrypted previously and loose all the data.

Encrypting when creating an image is much safer approach as you know that you just created this volume (hence in a change I'm about to push to address previous comments we fail if volume was created but not encrypted) and it is safe to assume that there is no data on it yet.

dillaman:
> If for any reason volume was provisioned as not encrypted and it got some data on it, the last thing you want is to loose that data. If we encrypt with LUKS a volume on a first attach, it is possible that we will format the disk that was not encrypted previously and loose all the data.

How do you get a volume w/ data on it when it's dynamically provisioned? I don't think the CSI should ever fallback to unencrypted mounting if something fails (when encryption is enabled).

vasyl-purchel:
attach process currently configured in such a way that if you requested encrypted volume to be attached, but it is not encrypted - the process fails

as a possible scenario I can think of when volume have data and not encrypted is:

* someone upgrades to a new version and sets encrypted: true for existing volume - on next attach process will see volume as not encrypted and try to encrypt it - hence delete all data

even though in documentation I've added that it is not supported, we still do not want to delete customer data for a mistake in configuration

dillaman:
> even though in documentation I've added that it is not supported, we still do not want to delete customer data for a mistake in configuration

That's where that metadata property on the image can be used to prevent such an occurrence. If encryption is enabled, the controller could record the state that encryption should be used. Then on attach, if encryption is enabled, you can test the block via `blkid`. If it shows nothing exists, you can create the LUKS layer. If it shows something exists and it's not LUKS, error out.

vasyl-purchel:
ok, I'll check such approach with my team and see if we can implement it

ShyamsundarR:
> > even though in documentation I've added that it is not supported, we still do not want to delete customer data for a mistake in configuration
> 
> That's where that metadata property on the image can be used to prevent such an occurrence. If encryption is enabled, the controller could record the state that encryption should be used. Then on attach, if encryption is enabled, you can test the block via `blkid`. If it shows nothing exists, you can create the LUKS layer. If it shows something exists and it's not LUKS, error out.


Let me combine what I am saying with what @dillaman is here,
- Let the provisioning stage add an image meta key `csi.encrypted` with an initial value of `false`
  - This helps distinguish encrypted volumes from non-encrypted volumes, only encrypted volumes will have this key
  - The VolumeContext will also carry the provisioning parameters, so that can also be used to distinguish between encrypted volumes and others (although the CO may request back the volume context, at which time the CSI driver should be able to regenerate it, the same as before, where the image meta key again can help)

- The Staging call can then check both the VolumeContext and the presence of `csi.encrypted`, to decide to first encrypt, and then flip the `csi.encrypted` to `true` and then map and proceed.
  - This safeguards older volumes from being encrypted (even when the CO admin forces an recreate of the StorageClass (in kube) to add the encryption parameter)
  - Also provides a means to encrypt it only once before it is used the first time

When a volume is dynamically provisioned, we do have the leverage of taking action during the first time it is Staged (mapped and mounted). We even create a filesystem on rbd images, the first time we map and  mount them, I see this as no different at present.

docs/deploy-rbd.md
```diff
@@ -156,3 +157,50 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
 **Deploy Helm Chart:**
 
 [See the Helm chart readme for installation instructions.](../charts/ceph-csi-rbd/README.md)
+
+## Encryption for RBD volumes
+
+* Enabling encryption on volumes created without encryption is **not supported**
+* Enabling encryption for storage class that has PVs created without encryption
+  is **not supported**
```
Madhu-1: can you add it as Note?

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+		return "", err
+	}
+
+	encrypted := strings.TrimSpace(value)
+	klog.V(4).Infof(Log(ctx, "image %s encrypted state metadata reports '%s'"), imageSpec, encrypted)
+	return encrypted, nil
+}
+
+// SaveRbdImageEncryptionStatus sets image metadata for encryption status
+func SaveRbdImageEncryptionStatus(ctx context.Context, cr *Credentials, monitors, imageSpec, status string) error {
```
Madhu-1: I was thinking can  we just save the image metadata with rbd image itself, not with rados omap

example:
```
sh-4.2# rbd image-meta set replicapool/test encrypted true
sh-4.2# rbd image-meta list replicapool/test
There is 1 metadatum on this image:

Key       Value 
encrypted true 
sh-4.2# rbd image-meta get replicapool/test encrypted
true
sh-4.2# rbd image-meta remove replicapool/test encrypted
```

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+func IsDeviceOpen(device string) (bool, error) {
+	_, mappedFile, err := DeviceEncryptionStatus(device)
+	return (mappedFile != ""), err
+}
+
+// DeviceEncryptionStatus looks to identify if the passed device is a LUKS mapping
+// and if so what the device is and the mapper name as used by LUKS.
+// If not, just returns the original device and an empty string.
+func DeviceEncryptionStatus(devicePath string) (mappedDevice, mapper string, err error) {
+	if !strings.HasPrefix(devicePath, "/dev/mapper") {
```
Madhu-1: can you use mapperFilePathPrefix here instead of `"/dev/mapper"` ?

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+		return devicePath, "", nil
+	}
+	mapPath := strings.TrimPrefix(devicePath, "/dev/mapper/")
+	stdout, _, err := LuksStatus(mapPath)
+	if err != nil {
+		// Non-zero return implies not an active LUKS device
```
Madhu-1: can you just log the error?

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+		return devicePath, "", nil
+	}
+	for i := 1; i < len(lines); i++ {
+		kv := strings.SplitN(strings.TrimSpace(lines[i]), ":", 2)
+		if strings.Compare(kv[0], "device") == 0 {
+			return strings.TrimSpace(kv[1]), mapPath, nil
+		}
+	}
+	// Identified as LUKS, but failed to identify a mapped device
+	return "", "", errors.New("mapped device not found")
```
Madhu-1: fmt.Errorf(" mapped device not found in path  %v",devicePath)

pkg/util/crypto.go
Madhu-1: @dillaman @ShyamsundarR ^^

vasyl-purchel: that's what SetImageMeta is doing, it is using `rbd image-meta set`

examples/rbd/secret.yaml
```diff
@@ -10,3 +10,6 @@ stringData:
   # specified in the storage class
   userID: <plaintext ID>
   userKey: <Ceph auth key corresponding to ID above>
+data:
+  # Key value is a base64 encoded encryption passphrase
+  encryptionPassphrase: dGVzdA==
```
Madhu-1: as the value is already encoded can we use stringData instead of data?

e2e/utils.go
```diff
@@ -606,11 +607,90 @@ func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
...
+	if err != nil {
+		Fail(err.Error())
+	}
+	encryptedState, err := getImageMeta(rbdImageSpec, ".rbd.csi.ceph.com/encrypted", f)
+	if err != nil {
+		Fail(err.Error())
+	}
+	Expect(encryptedState).To(Equal("encrypted"))
+
+	mountType, err := getMountType(app.Name, app.Namespace, "/var/lib/www/html", f)
```
Madhu-1: instead of `/var/lib/www/html` get it from app spec

pkg/rbd/controllerserver.go
```diff
@@ -177,6 +185,14 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, err
 	}
 
+	if rbdVol.Encrypted {
+		err = ensureEncryptionMetadataSet(ctx, cr, rbdVol)
+		if err != nil {
+			klog.Errorf(util.Log(ctx, err.Error()))
```
Madhu-1: we need to delete the rbd image if medataset fails, as  we  should not leave stale rbd image

pkg/rbd/nodeserver.go
```diff
@@ -559,3 +545,111 @@ func (ns *NodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetC
 		},
 	}, nil
 }
+
+func (ns *NodeServer) processEncryptedDevice(ctx context.Context, volOptions *rbdVolume, devicePath string, cr *util.Credentials, secrets map[string]string) (string, error) {
+	imageSpec := volOptions.Pool + "/" + volOptions.RbdImageName
+	encrypted, err := util.CheckRbdImageEncrypted(ctx, cr, volOptions.Monitors, imageSpec)
+
```
Madhu-1: extra new line

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// DeviceEncryptionStatus looks to identify if the passed device is a LUKS mapping
+// and if so what the device is and the mapper name as used by LUKS.
+// If not, just returns the original device and an empty string.
+func DeviceEncryptionStatus(devicePath string) (mappedDevice, mapper string, err error) {
+	if !strings.HasPrefix(devicePath, "/dev/mapper") {
+		return devicePath, "", nil
+	}
+	mapPath := strings.TrimPrefix(devicePath, "/dev/mapper/")
```
Madhu-1: can you use mapperFilePathPrefix here instead of "/dev/mapper" ?

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+		return devicePath, "", nil
+	}
+	lines := strings.Split(string(stdout), "\n")
+	if !strings.HasSuffix(lines[0], " is active.") {
+		// Implies this is not a LUKS device
+		return devicePath, "", nil
+	}
+	for i := 1; i < len(lines); i++ {
+		kv := strings.SplitN(strings.TrimSpace(lines[i]), ":", 2)
+		if strings.Compare(kv[0], "device") == 0 {
```
Madhu-1: validate the length of `kv` before accessing it

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+		return devicePath, "", nil
+	}
+	mapPath := strings.TrimPrefix(devicePath, "/dev/mapper/")
+	stdout, _, err := LuksStatus(mapPath)
+	if err != nil {
+		// Non-zero return implies not an active LUKS device
+		return devicePath, "", nil
+	}
+	lines := strings.Split(string(stdout), "\n")
+	if !strings.HasSuffix(lines[0], " is active.") {
```
Madhu-1: validate the length of `lines` before using it

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+// CheckRbdImageEncrypted verifies if rbd image was encrypted when created
+func CheckRbdImageEncrypted(ctx context.Context, cr *Credentials, monitors, imageSpec string) (string, error) {
+	value, err := GetImageMeta(ctx, cr, monitors, imageSpec, encryptionMetaKey)
+	if err != nil {
+		klog.Errorf(Log(ctx, "checking image %s encrypted state metadata failed: %s"), imageSpec, err)
+		return "", err
+	}
+
+	encrypted := strings.TrimSpace(value)
+	klog.V(4).Infof(Log(ctx, "image %s encrypted state metadata reports '%s'"), imageSpec, encrypted)
```
Madhu-1: klog.V(4).Infof(Log(ctx, "image %s encrypted state metadata reports %q"), imageSpec, encrypted)

pkg/util/crypto.go
Madhu-1: sorry I missed  it

pkg/util/crypto.go
```diff
@@ -0,0 +1,136 @@
...
+
+// VolumeMapper returns file name and it's path to where encrypted device should be open
+func VolumeMapper(volumeID string) (mapperFile, mapperFilePath string) {
+	mapperFile = mapperFilePrefix + volumeID
+	mapperFilePath = path.Join(mapperFilePathPrefix, mapperFile)
+	return mapperFile, mapperFilePath
+}
+
+// GetCryptoPassphrase Retrieves passphrase to encrypt volume
+func GetCryptoPassphrase(volumeID string, secrets map[string]string) (string, error) {
```
Madhu-1: `volumeID` not used

examples/rbd/secret.yaml
vasyl-purchel: using stringData will have plaintext passphrase instead of base64 encoded version

pkg/util/crypto.go
vasyl-purchel: done

pkg/util/crypto.go
vasyl-purchel: done

pkg/util/crypto.go
vasyl-purchel: as discussed offline: the confusion came from testing part where I was looking for rados omap value to map pvc to VolumeID in ceph, but as it appears there is a way to do so and I've changed that now.

pkg/util/crypto.go
vasyl-purchel: removed

pkg/util/crypto.go
vasyl-purchel: done

pkg/rbd/nodeserver.go
vasyl-purchel: removed

pkg/rbd/controllerserver.go
vasyl-purchel: done

e2e/utils.go
vasyl-purchel: done

pkg/util/crypto.go
vasyl-purchel: done

pkg/util/crypto.go
vasyl-purchel: added logging, though it's in info format as device not active in LUKS doesn't mean it is a failure

pkg/util/crypto.go
vasyl-purchel: done

docs/deploy-rbd.md
vasyl-purchel: done

vasyl-purchel:
@ShyamsundarR and @dillaman the proposition to change original workflow that you have asked is implemented, could you please check if it is what you were thinking about?

Madhu-1:
@humblec @nixpanic @ShyamsundarR @dillaman PTAL

pkg/rbd/nodeserver.go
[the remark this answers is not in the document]
ShyamsundarR: Note: `volume_options` in the `NodeStageVolume` request is what is returned as the volume context from CreateVolume request, and it is not directly from the StorageClass (or other such constructs in other COs). Thus the limitation to not change existing StorageClasses to make then dynamically provision encrypted volumes in the future would be incorrect. IOW, all volumes generated prior to adding the `encrypted` parameter to the StorageClass will remain unencrypted and not carry the `encrypted` parameter in their respective `volume_options` and all future volumes would. We can remove riders around the same in the documentation.

pkg/rbd/nodeserver.go
```diff
@@ -559,3 +545,110 @@ func (ns *NodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetC
...
+	}
+
+	if err = util.EncryptVolume(ctx, devicePath, passphrase); err != nil {
+		err = fmt.Errorf("failed to encrypt volume %s/%s: %v", rbdVol.Pool, rbdVol.RbdImageName, err)
+		klog.Errorf(util.Log(ctx, err.Error()))
+		return err
+	}
+
+	imageSpec := rbdVol.Pool + "/" + rbdVol.RbdImageName
+	err = util.SaveRbdImageEncryptionStatus(ctx, cr, rbdVol.Monitors, imageSpec, rbdImageEncrypted)
```
ShyamsundarR: If we succeed in line 595, but the plugin fails before this step is executed, we would on an retry (post the plugin status is restored to running) encrypt the device again. Is there a way to safeguard against such double encryption? Maybe options to check if device is already encrypted and to then proceed with the required encryption?

pkg/rbd/rbd_attach.go
```diff
@@ -260,21 +260,38 @@ func waitForrbdImage(ctx context.Context, backoff wait.Backoff, volOptions *rbdV
...
 }
 
 // detachRBDImageOrDeviceSpec detaches an rbd imageSpec or devicePath, with additional checking
 // when imageSpec is used to decide if image is already unmapped
-func detachRBDImageOrDeviceSpec(ctx context.Context, imageOrDeviceSpec string, isImageSpec, ndbType bool) error {
-	var err error
+func detachRBDImageOrDeviceSpec(ctx context.Context, imageOrDeviceSpec string, isImageSpec, ndbType bool, volumeID string) error {
 	var output []byte
 
+	mapperFile, mapperPath := util.VolumeMapper(volumeID)
```
ShyamsundarR: (optimization) Instead of checking this for each detach request, that will introduce unwanted luks checks for non-encrypted volumes as well, pass in a flag indicating this volume is encrypted or not, and act accordingly.

In Stage phase we can stash the encrypted state of the device into the stash file and use it appropriately to pass in the flag when Unstage calls this function. (see `stashRBDImageMetadata`)

pkg/rbd/controllerserver.go
```diff
@@ -150,6 +150,14 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if found {
+		if rbdVol.Encrypted {
+			err = ensureEncryptionMetadataSet(ctx, cr, rbdVol)
```
ShyamsundarR: If this fails, the CO has no way to garbage collect the volume created thus far (unlike calling `deleteImage` in line 193 below). The reason being the vol reservation is cancelled only if we fail to find the image, and that will not happen as the image is never deleted, so the next create volume request will attempt to use the same image and reset the key.

The solution I think falls into moving the encryption key to the RADOS OMap based journal, rather than retaining it in the image itself (I understand this is contradictory to earlier review, but I think this is better than amending image meta attributes post creating the image).

If we retain the encryption status on the CSI image RADOS OMap, we need not bother with attempting to set it post the image is created as well, which is better when returning found status for an existing image, than modifying it post it being found.

As this is data that is stored on disk (IOW, the image meta or the RADOS key), we need to evaluate this choice before we merge this change.

examples/rbd/storageclass.yaml
```diff
@@ -36,6 +36,11 @@ parameters:
    csi.storage.k8s.io/fstype: ext4
    # uncomment the following to use rbd-nbd as mounter on supported nodes
    # mounter: rbd-nbd
+
+   # Instruct the plugin it has to encrypt the volume
+   # By default it is disabled. Valid values are “true” or “false”.
+   # A string is expected here, i.e. “true”, not true.
+   # encrypted: "true"
```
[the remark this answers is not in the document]
vasyl-purchel: if you check existing StorageClass for [AWS EBS](https://kubernetes.io/docs/concepts/storage/storage-classes/#aws-ebs) you can see that it is using also `encrypted` field rather than image features, so I think it's more natural to expose it in a similar way.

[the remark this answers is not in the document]
ShyamsundarR: Yes, I agree and I have seen that. They also lack (or do not need) an `imageFeatures` parameter, which we could leverage.

This is just a thought, as I said, I am more for reducing the StorageClass parameters, but let's hear from others and see if this makes sense in this particular case.

pkg/rbd/nodeserver.go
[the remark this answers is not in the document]
vasyl-purchel: I do believe this is true, but I think it's better to mention this in documentation still that it's not supported as it will be very confusing for users after changing existing storage class to `encrypted: true` and later seeing some volumes encrypted and some not

or else we could explain the behaviour in the documentation (this won't reduce confusion I think, but at least there will be a place in documentation where we can point users if they have questions around the topic)

pkg/rbd/rbd_attach.go
vasyl-purchel: agree that an optimization can be done, but I would prefer to merge existing PR first as it's going on for a long time already with a follow up optimization PR

pkg/rbd/nodeserver.go
vasyl-purchel: encryptDevice method is called when we need an encryption and in case of encryption success but saving status fails (or smth fails in between) we'll have a device treated as requires encryption but have a file system on it (check line 566) and on a second attempt we would fail - this is a safeguard from making sure that we do not try to format a device to encrypt it while it have an existing data on it...

as an improvement I can see we could change line 566 to verify if the existing format is `crypt` which means it's already encrypted and then just change the metadata instead of re-encrypting

pkg/util/cryptsetup.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// LuksOpen opens LUKS encrypted partition and sets up a mapping
+func LuksOpen(devicePath, mapperFile, passphrase string) (stdout, stderr []byte, err error) {
+	return execCryptsetupCommand(&passphrase, "luksOpen", devicePath, mapperFile, "-d", "/dev/stdin")
+}
+
+// LuksClose removes existing mapping
+func LuksClose(mapperFile string) (stdout, stderr []byte, err error) {
+	return execCryptsetupCommand(nil, "luksClose", mapperFile)
```
humblec: @vasyl-purchel suppose the luksClose fails for some reason ( errors like 'still mapped' or 'resource busy'..etc) , do we need to do forceful close ?

humblec: @vasyl-purchel also what happens to the mapperfile after close. Is the "/dev" entry for that remains in the system as stale?

pkg/util/cryptsetup.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// LuksStatus returns encryption status of a provided device
+func LuksStatus(mapperFile string) (stdout, stderr []byte, err error) {
+	return execCryptsetupCommand(nil, "status", mapperFile)
+}
+
+func execCryptsetupCommand(stdin *string, args ...string) (stdout, stderr []byte, err error) {
+	var (
+		program       = "cryptsetup"
```
humblec: can we do a 'check' for this binary presence in the node and fail before we proceed with command execution @vasyl-purchel

docs/deploy-rbd.md
```diff
@@ -156,3 +157,51 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
 **Deploy Helm Chart:**
 
 [See the Helm chart readme for installation instructions.](../charts/ceph-csi-rbd/README.md)
+
+## Encryption for RBD volumes
```
humblec: @vasyl-purchel we should list the `prereq` to be met in the node for supporting encryption , example : binaries to be present, the kernel module to be loaded..etc

humblec:
@vasyl-purchel.et al, Thanks for the PR ! overall PR looks good to me. Have few review comments. Also I would like to get this in for `release-2.0.0`, so flagging this PR accordingly.

pkg/rbd/nodeserver.go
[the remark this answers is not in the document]
ShyamsundarR: I agree to the above, I just did not want documentation to sound like, "if this is changed older volumes will no longer work" or some such. To reduce ambiguity it is better never to edit an StorageClass anyway, if possible some rewording in the deploy-rbd.md file can be made.

Bottom line I wanted ensure we are in sync on why a new StorageClass is needed, and what happens if we edit older ones.

pkg/rbd/nodeserver.go
ShyamsundarR: > as an improvement I can see we could change line 566 to verify if the existing format is `crypt` which means it's already encrypted and then just change the metadata instead of re-encrypting

I was not aware the above could be done, and that was my doubt/question. We can track that as a TODO (preferably via an issue) and deal with it subsequently.

pkg/rbd/rbd_attach.go
ShyamsundarR: Ack! Let's track this as well in an issue, so that it gets done. I am not going to block the patch on an optimization (or nits for that matter).

docs/deploy-rbd.md
vasyl-purchel: so regarding binaries - there is no requirements, encryption is done inside rbd-plugin container that have all binaries included, the only requirement probably needed is dm-crypt module in kernel (that was missing previously from minikube image), will add the note on kernel module

pkg/util/cryptsetup.go
vasyl-purchel: this binary is not required on the node but rather on rbd-plugin container and it is present there

pkg/util/cryptsetup.go
vasyl-purchel: as far as I know if luksClose failed because of resource is busy there is no real "force" close available. I believe it might be possible to automate some steps but this is not an easy task and usually requires human decision - you need to find what keeps the resource busy, free the resource and after you should be able to do luksClose with no issues

regarding mapperfiles - from what I've seen /dev/mapper/... devices are removed, and then we handle /dev/rbd... device as if there was no such thing as mapper device, so it's detached as if there was no encryption.

e2e/utils.go
```diff
@@ -606,11 +608,84 @@ func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
...
+
+	imageIDRegex := regexp.MustCompile(`(\w+\-?){5}$`)
+	imageID := imageIDRegex.FindString(pv.Spec.CSI.VolumeHandle)
+	return fmt.Sprintf("replicapool/csi-vol-%s", imageID), nil
+}
+
+func getImageMeta(rbdImageSpec, metaKey string, f *framework.Framework) (string, error) {
+	cmd := fmt.Sprintf("rbd image-meta get %s %s", rbdImageSpec, metaKey)
+	opt := metav1.ListOptions{
+		LabelSelector: "app=rook-ceph-tools",
```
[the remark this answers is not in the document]
vasyl-purchel: yes, to verify that encryption process have set encryption status on the image correctly I'm retrieving image metadata stored in Ceph through toolbox provided by rook.

pkg/rbd/nodeserver.go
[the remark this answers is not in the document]
vasyl-purchel: if you could suggest better wording I could change that, for now I think about "not supported" as "you're on your own if you do this". Explaining each case on how it will work will be documenting too deep into internals and will only add more confusion around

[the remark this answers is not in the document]
ShyamsundarR: Ack! Let's leave the doc as is "not supported" is good enough.

pkg/rbd/controllerserver.go
vasyl-purchel: my takeaway on this would be that using image-meta is better since this is closer to the image itself which means safer solution as we don't have a separate object somewhere holding crucial information...

if I'm getting it right, the next request will try to use the same image, so it will reset the encryption key on this image and will attempt second time image-meta. And this becomes an issue only if this call fails persistently and not a once-off failure due to network issues for example.

so personally I would prefer to be on a safer side to have information about the image on the image itself, but happy to hear more suggestions

vasyl-purchel:
@ShyamsundarR regarding:
> How are volumes created from Snapshots and other volumes retaining the encryption status? Or is this a future concern? Can we just add a safeguard in the code, if it is a future concern, to return an error if source volume is encrypted?
I don't really know how snapshots work in ceph-csi and how to check this. For now I think it was discussed that this will be a future concern for snapshots. We could add a safeguard in the code though I don't know how I can check that and verify it works (from what I can see snapshot e2e tests are commented out)

docs/deploy-rbd.md
humblec: Sure, `dm-crypt` is a must for encryption to work here, so we should definitely mention it. Also there are different flavors of OS which is kind of stripped for container host platform. I am not sure everyone carry this module in it though.

pkg/util/cryptsetup.go
humblec: @vasyl-purchel yes in plugin, but wondering do we have a check regardless our container image has it or not. The reason why I am bringing this here is, I have spoken to users/customers who containerize the plugins with their base image..etc.

pkg/util/cryptsetup.go
humblec: Ok., as long as you confirm, there is no stale entries remains in the system in a close/failed operation, I am fine.

e2e/utils.go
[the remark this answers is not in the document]
humblec: @vasyl-purchal , In other words, cant we enable `encryption` for volumes if the ceph cluster is not deployed by Rook in the backend ?

pkg/util/cryptsetup.go
vasyl-purchel: though having a check if tool exists is extra overhead for every command call to encrypt volume... I'd say maybe better to just mention in the docs that it must be present on the container?

e2e/utils.go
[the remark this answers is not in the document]
vasyl-purchel: um... yes it can be enabled for non rook clusters. for e2e tests we have here rook cluster deployed and that's why I have it here... the actual plugin code does not use any rook labels, it calls what ever cluster is used in configuration

pkg/util/cryptsetup.go
vasyl-purchel: if close operation fails, the mapper device will remain and the /dev/rbd-... device won't be able to be removed as it's still busy too. it will remain as an issue until manual intervention to successfully close it...

docs/deploy-rbd.md
```diff
@@ -55,6 +55,7 @@ make image-cephcsi
...
 | `csi.storage.k8s.io/provisioner-secret-namespace`, `csi.storage.k8s.io/node-stage-secret-namespace` | yes (for Kubernetes) | namespaces of the above Secret objects                                                                                                                                                                                  |
 | `mounter`                                                                                           | no                   | if set to `rbd-nbd`, use `rbd-nbd` on nodes that have `rbd-nbd` and `nbd` kernel modules to map rbd images                                                                                                              |
+| `encrypted`                                                                                         | no                   | disabled by default, use `"true"` to enable LUKS encryption on pvc and `"false"` to disable it. **Do not change for existing storage classes**                                                                          |
```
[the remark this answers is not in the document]
vasyl-purchel: done

pkg/rbd/controllerserver.go
ShyamsundarR: The workflow I would like to see for `CreateVolume` when returning an existing volume, is one where it does not modify any volume attributes when it finds a volume matching the request.

This is important if and when we support `ListVolumes`, as in that case we cannot change (set/reset) volume attributes for all encrypted volumes that we list, as some may be already in use by then.

The [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume-deletevolume-listvolumes) calls out that, the CO can send repeated `CreateVolume` or `ListVolumes` to rebuild its state, or to retry and get a `volume_id` for a timed out request. Which makes the above (`ListVolumes`) possible in the future.

For the above reasons, and also as an implementation workflow, I want to ensure there is no modification of any attribute of a volume, when we determine a pre-existing volume matching the criteria in a `CreateVolume` request.

Coming to where we maintain the encrypted flag,

1. if we choose to maintain it on the image (for reasons stated in [this](https://github.com/ceph/ceph-csi/pull/697#discussion_r355543759) comment), then I would state we should delete the image and undo the reservation in the function [`checkVolExists`](https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/rbd_journal.go#L159). This will recreate the requested volume and ensure we finish all steps (including setting up the image key) before returning a success as a part of a single running transaction.
Even with the above, when we start supporting `ListVolumes`, we would return a volume that has a `volume_context` which would state it is not encrypted if the image key was not set (due to interrupted transaction on a prior `CreateVolume` request), but the reservation and the image were found. Also, we now need to handle image delete as part of the `undoVolReservation`, or just before the same.

1. If we maintain the encrypted flag, as part of the CSI OMap for the image, we would have created the image post setting up the required attributes (in this case in the CSI OMap), and the workflow in `checkVolExists` can flow as it is currently, with an additional check for the added encrypted key. There would be no image deletion that would need to be done and `ListVolumes` can respond with concrete data as the reservation, its attributes and the image is found on disk.

I hence prefer that we store the encrypted (and other future keys) in the CSI OMap for the image, rather than the image itself.

The further workflow ensures that `NodeStageVolume` gets the `volume_context` that enables working with encrypted volumes without checking the actual OMap or the image meta. `NodeStageVolume` though will have to update either the OMap to reflect that the volume is already encrypted, or we can add a image key, when we actually encrypt the volume, to state it is encrypted.

IOW, `NodeStageVolume` can use image keys, and `CreateVolume` can use the CSI OMaps to store the encryption status. (This is assuming we can tackle not re-encrypting an image as discussed [here](https://github.com/ceph/ceph-csi/pull/697#discussion_r355185194))

pkg/util/cryptsetup.go
humblec: @vasyl-purchel it shouldnt be an overhead for example , in kube it check for binary existence in various places.  But we can survive with documentation too. :)

vasyl-purchel: the documentation now have the part mentioning to have cryptsetup installed in the container for custom builds

docs/deploy-rbd.md
vasyl-purchel: note about kernel module is added

examples/rbd/storageclass.yaml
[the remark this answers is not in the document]
dillaman: I'd prefer to avoid overloading image features. In a perfect world, honestly, the CSI should just manage them and a user should never touch them. We want to get to the point where users of the CSI really don't need to understand how RBD/CephFS work internally.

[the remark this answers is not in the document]
WoohyungHan: +1 @dillaman The k8s user does not need to know about the Ceph. But I agree too @ShyamsundarR , we could reduce the StorageClass parameters as much possible. I think the CSI interface has to accept more arguments like 'encryption'. (and kubernetes itself maybe?)

But for now, we still need a new variable in the StorageClass, and i'd prefer the variable name is not 'imageFeatures'. The word is using in the Ceph, and it could be confused. (a user doesn't want to know about the Ceph word...)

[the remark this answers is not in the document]
ShyamsundarR: Ack! thanks for the feedback. We will proceed without overloading the Ceph `imageFeatures` term, and use separate parameters for now (till other systemic improvements provide a way to optimize parameters in the StorageClass).

pkg/rbd/controllerserver.go
ShyamsundarR: Based on a discussion with @dillaman it is better to maintain the the encrypted key with the image, IOW as `image-meta` key:value. This helps copying the key across image clones, and hence preserving encryption status for clones as well.

Further, to prevent incomplete `CreateVolume` we will maintain an "in-progress" key in the CSI OMap, such that repeat `CreateVolumes` can choose to complete the transaction or can garbage collect the image and the reservation and start afresh. This is not a necessary concern for this patch, IOW can be added subsequently as well.

This issue will hence be marked resolved once we create an issue to track the above. Also, this comment no longer blocks this PR.

ShyamsundarR:
> @ShyamsundarR regarding:
> 
> > How are volumes created from Snapshots and other volumes retaining the encryption status? Or is this a future concern? Can we just add a safeguard in the code, if it is a future concern, to return an error if source volume is encrypted?
> 
> I don't really know how snapshots work in ceph-csi and how to check this. For now I think it was discussed that this will be a future concern for snapshots. We could add a safeguard in the code though I don't know how I can check that and verify it works (from what I can see snapshot e2e tests are commented out)

Cloned images should carry the encrypted keys, so that these are staged by LUKS. As `image-meta` maintain copying the image keys across clones, as long as we store the encryption keys on the image (as concluded [here](https://github.com/ceph/ceph-csi/pull/697#discussion_r356896828)) clones will not be a concern.

dillaman:
> Cloned images should carry the encrypted keys, so that these are staged by LUKS. As `image-meta` maintain copying the image keys across clones, as long as we store the encryption keys on the image (as concluded [here](https://github.com/ceph/ceph-csi/pull/697#discussion_r356896828)) clones will not be a concern.

I think for short-term we need to prevent k8s snapshots and clones -- at least if they are not using the same encryption settings + key. We should open a follow-up issue ticket since I don't think that snapshot/clone support has merged yet(?) so that it doesn't get lost. 

Longer term, we will also want a design that can support re-encrypting upon snapshot/clone operations with new keys since w/o re-encryption you can start to attack the encryption [1][2].

[1] https://cloud.google.com/compute/docs/disks/customer-supplied-encryption#create_snapshot
[2] https://mbroz.fedorapeople.org/talks/DevConf2016/devconf2016-luks2.pdf

Madhu-1:
@vasyl-purchel CI is failing can you please fix it.

@ShyamsundarR can you please open a new issue to track the things which need to be taken care of later?

pkg/rbd/controllerserver.go
ShyamsundarR: Suggested improvements tracked in #744

pkg/rbd/nodeserver.go
ShyamsundarR: Suggested improvements tracked in #744

pkg/rbd/rbd_attach.go
ShyamsundarR: Suggested improvements tracked in #744

ShyamsundarR:
> > Cloned images should carry the encrypted keys, so that these are staged by LUKS. As `image-meta` maintain copying the image keys across clones, as long as we store the encryption keys on the image (as concluded [here](https://github.com/ceph/ceph-csi/pull/697#discussion_r356896828)) clones will not be a concern.
> 
> I think for short-term we need to prevent k8s snapshots and clones -- at least if they are not using the same encryption settings + passphrase key. We should open a follow-up issue ticket since I don't think that snapshot/clone support has merged yet(?) so that it doesn't get lost.

My understanding is that the encryption parameter added as image-meta on the image will be retained by clones (as a part of rbd copying the keys across when a clone operation is performed). Hence CSI will also treat such images as encrypted (as the image-meta parameter is present), and Stage them appropriately.

The gap would be, if a different storage class is used to create a volume from a snapshot or an existing volume, as the volume content source, at which point the volume_options that gets carried over to the Stage request **may** not have the encryption parameter.

I opened issue #745 for this purpose, to check and ensure we do not end up in any unsupported or unidentifiable situation.

humblec:
Thanks all !!

PR #698, opened 2019-10-24
Implement CSIDriver object creation in cephcsi

Madhu-1:
This will be helpful to list the CSI drivers deployed in the kubernetes cluster. if we have `N`
number of CSI drivers deployed in the cluster, the user can  query `kubectl get  csidriver` and get
the list of registered CSI drivers and  it requires kube 1.14+ and also the functionality is disabled by default


pkg/util/k8s_register.go
```diff
@@ -0,0 +1,120 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
nixpanic: we live in 2019 :-)

pkg/util/k8s_register.go
```diff
@@ -0,0 +1,120 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Spec: k8scsi.CSIDriverSpec{
+			AttachRequired: &attach,
+			PodInfoOnMount: &mountInfo,
+		},
+	}
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		csidrivers := csiClientset.StorageV1beta1().CSIDrivers()
```
nixpanic: I know there is a check in `charts/...` for the kubernetes version, but would this not give an error in case uses deploy manually on an older kubernetes version, or even on a different container platform?

charts/ceph-csi-cephfs/templates/provisioner-rules-clusterrole.yaml
```diff
@@ -37,4 +37,11 @@ rules:
     resources: ["volumeattachments"]
     verbs: ["get", "list", "watch", "update", "patch"]
 {{- end -}}
+{{- if semverCompare ">=1.14" .Capabilities.KubeVersion.GitVersion }}
+{{- if .Values.registerDriver }}
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["csidrivers"]
+    verbs: ["create", "delete"]
```
nixpanic: I guess `update` is not needed yet, but maybe a future version of the CSIDriver object needs to adjust its registration?

pkg/util/k8s_register.go
```diff
@@ -0,0 +1,120 @@
...
+		klog.Errorf("failed to delete CSIDriver object: %v", err)
+		return err
+	})
+	return retryErr
+}
+
+// RegisterCSIDriver create CSIDriver object during the controller plugin
+// starts and delete the CSIDriver object when controller plugin terminates
+func RegisterCSIDriver(c chan os.Signal, name string) {
+	k8sClient := NewK8sClient()
```
nixpanic: this would fail on non Kubernetes platforms?

pkg/util/k8s_register.go
Madhu-1: This is taken from  https://github.com/kubernetes-csi/cluster-driver-registrar/blob/master/cmd/csi-cluster-driver-registrar/k8s_register.go#L2 do i still need to change year?

pkg/util/k8s_register.go
Madhu-1: this is specific to kubernetes and not enabled by default

pkg/util/k8s_register.go
Madhu-1: yes this is specific to kubernetes and OCP

Madhu-1:
> Some more error checking would be good.
> 
> What is the plan for adding a test-case?

This will be disabled by default, we don't have enough flexibility in E2E to twick templates.

nixpanic:
On Thu, Oct 24, 2019 at 03:25:37AM -0700, Madhu Rajanna wrote:
> Madhu-1 commented on this pull request.
> 
> 
> 
> > @@ -0,0 +1,120 @@
> +/*
> +Copyright 2018 The Kubernetes Authors.
> 
> This is taken from  https://github.com/kubernetes-csi/cluster-driver-registrar/blob/master/cmd/csi-cluster-driver-registrar/k8s_register.go#L2 do i still need to change year?

If you modified it, yes. And it is good to mention the source in the
commit message.

nixpanic:
On Thu, Oct 24, 2019 at 03:26:01AM -0700, Madhu Rajanna wrote:
> Madhu-1 commented on this pull request.
> 
> 
> 
> > +func createOrUpdateCSIDriverInfo(csiClientset *kubernetes.Clientset,
> +	name string) error {
> +	attach := false
> +	mountInfo := true
> +	// Create CSIDriver object
> +	csiDriver := &k8scsi.CSIDriver{
> +		ObjectMeta: metav1.ObjectMeta{
> +			Name: name,
> +		},
> +		Spec: k8scsi.CSIDriverSpec{
> +			AttachRequired: &attach,
> +			PodInfoOnMount: &mountInfo,
> +		},
> +	}
> +	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
> +		csidrivers := csiClientset.StorageV1beta1().CSIDrivers()
> 
> this is specific to kubernetes and not enabled by default

Sure, but users can enabled it, right? The description of the option
does not mention it is limited to k8s.

nixpanic:
On Thu, Oct 24, 2019 at 03:26:54AM -0700, Madhu Rajanna wrote:
> Madhu-1 commented on this pull request.
> 
> 
> 
> > @@ -37,4 +37,11 @@ rules:
>      resources: ["volumeattachments"]
>      verbs: ["get", "list", "watch", "update", "patch"]
>  {{- end -}}
> +{{- if semverCompare ">=1.14" .Capabilities.KubeVersion.GitVersion }}
> +{{- if .Values.registerDriver }}
> +  - apiGroups: ["storage.k8s.io"]
> +    resources: ["csidrivers"]
> +    verbs: ["create", "delete"]
> 
> the update requires code changes to update the driver object, I don't think it's required as of now

Ok

Madhu-1:
@nixpanic updated readme to use this flag if CO is built around Kubernetes and updated the year

ShyamsundarR:
This feature should be implemented using a deployment manifest, and not in code. IOW the details of [CSIDriver object](https://kubernetes-csi.github.io/docs/csi-driver-object.html) documentation covers it in required detail.

Further from the same documentation, the lifetime of the object is supposed to be,
`
CSIDriver instance should exist for whole lifetime of all pods that use volumes provided by corresponding CSI driver, so Skip Attach and Pod Info on Mount features work correctly.
`, hence cleaning up on provisioner exits or restarts, is not required.

Lastly the object has a few properties that I think we can set as follows,
- `attachRequired` false (this is already the case in this PR. Helps speeding up attach rates by avoiding calling the provisioner sidecar for a fake attach)
  - Sub point here is to understand if we can hence remove the attacher sidecar `quay.io/k8scsi/csi-attacher:v1.2.0`?
- `podInfoOnMount` false (this is currently true in this PR, as we do not need pod details from kubernetes yet(?))

Madhu-1:
> This feature should be implemented using a deployment manifest, and not in code. IOW the details of [CSIDriver object](https://kubernetes-csi.github.io/docs/csi-driver-object.html) documentation covers it in required detail.
> 

I went through the document before making this PR. I have added this in the code to avoid more step on creating and deleting the CSIDriver object, I feel no harm in creating the CSI Driver Object from code, do you see any disadvantage in creating this object from code?

> Further from the same documentation, the lifetime of the object is supposed to be,
> `CSIDriver instance should exist for whole lifetime of all pods that use volumes provided by corresponding CSI driver, so Skip Attach and Pod Info on Mount features work correctly.`, hence cleaning up on provisioner exits or restarts, is not required.
> 

even if you delete the CSI Driver object things will work as it is, no harm in deleting the driver object. as we are deleting the CSI driver object when provisioner is deleted, ( if some consumer example UI is making use of this to list the deployed CSI driver names to create storage class can easily drop the provisioner name from the UI if provisioner is no longer running.

> Lastly the object has a few properties that we can set as follows,
> 
> * `attachRequired` false (this is already the case in this PR. Helps speeding up attach rates by avoiding calling the provisioner sidecar for a fake attach)
>   
>   * Sub point here is to understand if we can hence remove the attacher sidecar `quay.io/k8scsi/csi-attacher:v1.2.0`?

will remove it once the approach seems Ok

> * `podInfoOnMount` false (this is currently true in this PR, as we do not need pod details from kubernetes yet(?))

no harm in passing the pod info ( if required I can set it to false).

ShyamsundarR:
> > This feature should be implemented using a deployment manifest, and not in code. IOW the details of [CSIDriver object](https://kubernetes-csi.github.io/docs/csi-driver-object.html) documentation covers it in required detail.
> 
> I went through the document before making this PR. I have added this in the code to avoid more step on creating and deleting the CSIDriver object, I feel no harm in creating the CSI Driver Object from code, do you see any disadvantage in creating this object from code?

For one, this makes the code CO specific, where as this can easily be handled via a deployment manifest. I would do this in code when it is required, else it is better left as a part of the manifests. In this case code seems unwarranted. I would hence request that this move to a yaml as part of the deployment and remove the code.

> 
> > Further from the same documentation, the lifetime of the object is supposed to be,
> > `CSIDriver instance should exist for whole lifetime of all pods that use volumes provided by corresponding CSI driver, so Skip Attach and Pod Info on Mount features work correctly.`, hence cleaning up on provisioner exits or restarts, is not required.
> 
> even if you delete the CSI Driver object things will work as it is, no harm in deleting the driver object. as we are deleting the CSI driver object when provisioner is deleted, ( if some consumer example UI is making use of this to list the deployed CSI driver names to create storage class can easily drop the provisioner name from the UI if provisioner is no longer running.

True, but the properties and defaults will not be carried over to the node plugins when staging the PV. Also, it is not meant to be deleted, and is hence incorrect, as a result I would like this to be corrected.

> 
> > Lastly the object has a few properties that we can set as follows,
> > 
> > * `attachRequired` false (this is already the case in this PR. Helps speeding up attach rates by avoiding calling the provisioner sidecar for a fake attach)
> >   
> >   * Sub point here is to understand if we can hence remove the attacher sidecar `quay.io/k8scsi/csi-attacher:v1.2.0`?
> 
> will remove it once the approach seems Ok
> 
> > * `podInfoOnMount` false (this is currently true in this PR, as we do not need pod details from kubernetes yet(?))
> 
> no harm in passing the pod info ( if required I can set it to false).

If pod info needs to be passed, it needs to be fetched from some source and passed. Such actions can delay attach operations due to extra calls to fetch said information. Thus, there is a potential harm or a performance impact here. Correctness, before assuming it as harmless IMO.

Madhu-1:
> > > This feature should be implemented using a deployment manifest, and not in code. IOW the details of [CSIDriver object](https://kubernetes-csi.github.io/docs/csi-driver-object.html) documentation covers it in required detail.
> > 
> > 
> > I went through the document before making this PR. I have added this in the code to avoid more step on creating and deleting the CSIDriver object, I feel no harm in creating the CSI Driver Object from code, do you see any disadvantage in creating this object from code?
> 
> For one, this makes the code CO specific, where as this can easily be handled via a deployment manifest. I would do this in code when it is required, else it is better left as a part of the manifests. In this case code seems unwarranted. I would hence request that this move to a yaml as part of the deployment and remove the code.
> 

this can be done in CO (rook) also as (rook) needs to know the driver name and create and delete the csi driver object based on the deploy/teardown of the CSI driver, I thought other CO who are based on the kubernetes can  easily make use of this feature without many changes,
same mechanism was  followed in cluster-driver-registrar https://github.com/kubernetes-csi/cluster-driver-registrar for creation and deletion of CSI driver object 

and also work required in the kubernetes based CO will be minimal with these changes


> > > Further from the same documentation, the lifetime of the object is supposed to be,
> > > `CSIDriver instance should exist for whole lifetime of all pods that use volumes provided by corresponding CSI driver, so Skip Attach and Pod Info on Mount features work correctly.`, hence cleaning up on provisioner exits or restarts, is not required.
> > 
> > 
> > even if you delete the CSI Driver object things will work as it is, no harm in deleting the driver object. as we are deleting the CSI driver object when provisioner is deleted, ( if some consumer example UI is making use of this to list the deployed CSI driver names to create storage class can easily drop the provisioner name from the UI if provisioner is no longer running.
> 
> True, but the properties and defaults will not be carried over to the node plugins when staging the PV. Also, it is not meant to be deleted, and is hence incorrect, as a result I would like this to be corrected.

I didn't get why it is not meant to be deleted and incorrect?
> 
> > > Lastly the object has a few properties that we can set as follows,
> > > 
> > > * `attachRequired` false (this is already the case in this PR. Helps speeding up attach rates by avoiding calling the provisioner sidecar for a fake attach)
> > >   
> > >   * Sub point here is to understand if we can hence remove the attacher sidecar `quay.io/k8scsi/csi-attacher:v1.2.0`?
> > 
> > 
> > will remove it once the approach seems Ok
> > > * `podInfoOnMount` false (this is currently true in this PR, as we do not need pod details from kubernetes yet(?))
> > 
> > 
> > no harm in passing the pod info ( if required I can set it to false).
> 
> If pod info needs to be passed, it needs to be fetched from some source and passed. Such actions can delay attach operations due to extra calls to fetch said information. Thus, there is a potential harm or a performance impact here. Correctness, before assuming it as harmless IMO.

I can remove this but I don't think it's true related to potential harm and performance impact here. CO will have all the required pod information  before calling the publish request

ShyamsundarR:
> > > > Further from the same documentation, the lifetime of the object is supposed to be,
> > > > `CSIDriver instance should exist for whole lifetime of all pods that use volumes provided by corresponding CSI driver, so Skip Attach and Pod Info on Mount features work correctly.`, hence cleaning up on provisioner exits or restarts, is not required.
> > > 
> > > 
> > > even if you delete the CSI Driver object things will work as it is, no harm in deleting the driver object. as we are deleting the CSI driver object when provisioner is deleted, ( if some consumer example UI is making use of this to list the deployed CSI driver names to create storage class can easily drop the provisioner name from the UI if provisioner is no longer running.
> > 
> > 
> > True, but the properties and defaults will not be carried over to the node plugins when staging the PV. Also, it is not meant to be deleted, and is hence incorrect, as a result I would like this to be corrected.
> 
> I didn't get why it is not meant to be deleted and incorrect?

kubelet has an AttachVerify operation that is performed, which sends the request to the controller (or caches it, still tracing the code). This is sent based on the `attachRequired` option. If the object is deleted when the provisioner is inactive, then the node plugin will not be able to refer to the action to be taken and default to a determined behavior. Which in this case is to check the controller for an attach. There are 2 problems with this,
- Performance of attach is impacted, it moves from 9-12 seconds to 9-22 seconds for a pod to come up if a dummy AttachVerify needs to be performed. ( jFYI, I have benchmarked this)
- If we remove the attacher side car, the attach will never go through, as the dummy response will not be sent back

> 
> > > > Lastly the object has a few properties that we can set as follows,
> > > > 
> > > > * `attachRequired` false (this is already the case in this PR. Helps speeding up attach rates by avoiding calling the provisioner sidecar for a fake attach)
> > > >   
> > > >   * Sub point here is to understand if we can hence remove the attacher sidecar `quay.io/k8scsi/csi-attacher:v1.2.0`?
> > > 
> > > 
> > > will remove it once the approach seems Ok
> > > > * `podInfoOnMount` false (this is currently true in this PR, as we do not need pod details from kubernetes yet(?))
> > > 
> > > 
> > > no harm in passing the pod info ( if required I can set it to false).
> > 
> > 
> > If pod info needs to be passed, it needs to be fetched from some source and passed. Such actions can delay attach operations due to extra calls to fetch said information. Thus, there is a potential harm or a performance impact here. Correctness, before assuming it as harmless IMO.
> 
> I can remove this but I don't think it's true related to potential harm and performance impact here. CO will have all the required pod information before calling the publish request

We need to remove this (i.e `podInfoOnMount`), what the kubelet has, and what it needs to fetch may change, assuming behavior and requesting additional data, that is not yet used by the CSI driver, is not required.

PR #706, opened 2019-11-07
cephfs: add kernel version detection for mounting with client

nixpanic:

Linux kernel 4.17.0 adds support for quota with CephFS. Without quota,
it is not possible to fullfill the requirements of the CSI Spec and
guarantee sufficient space on the filesystem for a volume. With this in
mind, usage of the kernel client is only allowed with kernel 4.17.0 or
newer.

However, some Linux vendors backport features and patches to their
Enterprise products. These kernels may have an older version, but do
support quota. One of these is the kernel that comes with RHEL-7.7.

By comparing the current running version of the Linux kernel against
known versions that support quota, we can now automatically decide to
use the kernel client, or not.

Note that this does not change the 'forcekernelclient' parameter. The
parameter is still available and can be used for kernels that are not in
the 'known to support quota list'. Or users can pass the parameter to
use a CephFS kernel client that does not support quota.

### DIFF
--- a/pkg/cephfs/volumemounter.go
+++ b/pkg/cephfs/volumemounter.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/ceph/ceph-csi/pkg/util"
 
+	"golang.org/x/sys/unix"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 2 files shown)


nixpanic:
@shaas this might be of interest for you too

Madhu-1:
@nixpanic this is supressed by https://github.com/ceph/ceph-csi/pull/708

Issue #671, opened 2019-10-07
Plugins are not compatible with ARM64 when running on rook


Glorf:
Originally filled as rook issue: https://github.com/rook/rook/issues/4051

Madhu-1:
got some idea from @travisn here https://github.com/rook/rook/issues/4051#issuecomment-539852395

@nixpanic is it possible to take it up?

cyb70289:
Hi @nixpanic , @Madhu-1,  are you working on this issue? We met with same problem and fixed it with multi arch image, would like to upstream if you are not on it.

billimek:
Related slack thread for this from the rook slack space: https://rook-io.slack.com/archives/CG3HUV94J/p1569616214102700

cyb70289:
Thanks @billimek

Madhu-1:
@cyb70289 waiting for @nixpanic response, if he has not started this one, am more happy to assign this to you

nixpanic:
Thanks @billimek , I have re-assigned this issue to you. Let me know if I should put it in my backlog again.

nixpanic:
Sorry, I should have paid attention while reading, now assigned to @cyb70289 :+1:

cyb70289:
Per my test, quay.io doesn't support multi arch images. Though quay v3.0 release notes say it supports.[1]

To build multi arch image is easy(take my personal dockerhub account as an example):
- on x86 host: build cephcsi image, push to cyb70289/cephcsi-amd64:v1.2.1
- on arm host: build cephcsi image, push to cyb70289/cephcsi-arm64:v1.2.1
- on x86 host: download manifest-tool[2], run "./manifest-tool-linux-amd64 push from-args --platforms linux/amd64,linux/arm64 --template cyb70289/cephcsi-ARCH:v1.2.1 --target cyb70289/cephcsi:v1.2.1"
Now "docker pull cyb70289/cephcsi:v1.2.1" will download x86 or arm image matching host architecture.
On quay.io, this command failed with error "FATA[0022] Couldn't push manifests referenced in our manifest list: Referenced manifest push unsuccessful: response 400: 400 BAD REQUEST".

I plan to add a travis job to build arm64 image(travis supports arm64 now[3]) and automate above steps. But it looks no-go if quay.io doesn't support multi arch images.

[1] https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/red_hat_quay_release_notes/index#multi_architecture_container_image_support
[...]

Issue #138, opened 2019-01-21
Switch driver-registrar canary image to other releases


rootfs:
currently csi-v1.0 uses `quay.io/k8scsi/driver-registrar:canary`, since there is no 1.0 image yet.

Create an issue to track upstream release and make a switch once 1.0 tag is there.

Madhu-1:
@rootfs any idea when we will be switching to node-driver-registrar and cluster-driver-registrar as driver-registrar is deprecated?

rootfs:
@Madhu-1 make a PR :)

Madhu-1:
@rootfs can we close this one?

PR #707, opened 2019-11-08
Support Arm64 image

cyb70289:
### **This change log is obsolete. Please see https://github.com/ceph/ceph-csi/pull/707#issuecomment-558032838 for latest change log.**

Support multi-arch images

    Add jobs to build arm64 image and support multi-arch docker image.
    By multi-arch it means "docker pull cephcsi/cephcsi:canary" will
    download amd64 or arm64 image per host archtecture automatically.
    Architecture dependent images are named as cephcsi/cephcsi-amd64
    and cephcsi/cephcsi-arm64.
    
    Major changes:
    - Add travis job "build arm64 image" which runs on travis arm64 nodes
      to build cephcsi image for arm64.
    - Add travis job "push image manifest" which pushes multi-arch image
      manifest to docker registry.
    - Update Makefile to tag and push architecture dependent image with
      architecture appended in name (e.g, cephcsi/cephcsi-amd64:canary)
    - Add Makefile target "push-manifest-cephcsi" to push image manifest
      and support multi arch.
    - Add ENV_DOCKER_REGISTRY variable to configure docker registry
      for hosting built image, rather than hard coded quay.io.
    
    NOTES:
    - CI test jobs are disabled on Arm now build due to minikube issue.
    - Job "push-manifest-cephcsi" runs for every PR. It should only run
      when patch merges. Though it does nothing for normal PR (deploy.sh
      checks ${TRAVIS_PULL_REQUEST}), it still needs booting, installing
      packages, etc, which costs about 1 minute.
    - In current jobs, docker image is pushed if one test succeeds and
      runs deployment, even if other tests fails. After applying this
      patch, architecure image is handled same, pushed when any test
      succeeds. But multi-arch manifest is pushed only after all tests
      succeed, any broken test will block pushing multi-arch manifest.

CI test log
https://travis-ci.org/cyb70289/ceph-csi/builds/608636375

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ env:
     - DEP_VERSION=v0.5.4
 
 before_install:
+  - mkdir -p $GOPATH/bin
   - curl
     https://raw.githubusercontent.com/golang/dep/${DEP_VERSION}/install.sh|sh
 
@@ -75,6 +76,17 @@ jobs:
         - scripts/skip-doc-change.sh || travis_terminate 0;
         - make image-cephcsi || travis_terminate 1;
         - scripts/travis-functest.sh v1.17.0 || travis_terminate 1;
+
+    - name: cephcsi on Arm64
+      arch: arm64
+      script:
+        - scripts/skip-doc-change.sh || travis_terminate 0;
+        - make image-cephcsi || travis_terminate 1;
+        # No CI test job is availabe for Arm64 now due to below issues
+        # - k8s-csi sidecar images for Arm64 are not available
+        # - Travis Arm64 CI job runs inside unprivileged LXD which blocks
+        #   launching minikube test environment
+        - travis_terminate 0    # deploy only on x86
 
 deploy:
   - provider: script
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ CONTAINER_CMD?=docker
 
 CSI_IMAGE_NAME=$(if $(ENV_CSI_IMAGE_NAME),$(ENV_CSI_IMAGE_NAME),quay.io/cephcsi/cephcsi)
 CSI_IMAGE_VERSION=$(if $(ENV_CSI_IMAGE_VERSION),$(ENV_CSI_IMAGE_VERSION),canary)
+CSI_IMAGE=$(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION)
 
 $(info cephcsi image settings: $(CSI_IMAGE_NAME) version $(CSI_IMAGE_VERSION))
 
@@ -30,6 +31,11 @@ LDFLAGS ?=
 LDFLAGS += -X $(GO_PROJECT)/pkg/util.GitCommit=$(GIT_COMMIT)
 # CSI_IMAGE_VERSION will be considered as the driver version
 LDFLAGS += -X $(GO_PROJECT)/pkg/util.DriverVersion=$(CSI_IMAGE_VERSION)
+
+# set GOARCH explicitly for cross building, default to native architecture
+ifeq ($(origin GOARCH), undefined)
+GOARCH := $(shell go env GOARCH)
+endif
 
 all: cephcsi
 
@@ -52,15 +58,18 @@ func-test:
 .PHONY: cephcsi
 cephcsi:
 	if [ ! -d ./vendor ]; then dep ensure -vendor-only; fi
-	CGO_ENABLED=0 GOOS=linux go build -a -ldflags '$(LDFLAGS) -extldflags "-static"' -o  _output/cephcsi ./cmd/
+	CGO_ENABLED=0 GOOS=linux GOARCH=$(GOARCH) go build -a -ldflags '$(LDFLAGS) -extldflags "-static"' -o  _output/cephcsi ./cmd/
 
 image-cephcsi: cephcsi
 	cp _output/cephcsi deploy/cephcsi/image/cephcsi
-	$(CONTAINER_CMD) build -t $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) deploy/cephcsi/image
+	chmod +x deploy/cephcsi/image/cephcsi
+	$(CONTAINER_CMD) build -t $(CSI_IMAGE) deploy/cephcsi/image
 
 push-image-cephcsi: image-cephcsi
-	$(CONTAINER_CMD) push $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION)
-
+	$(CONTAINER_CMD) tag $(CSI_IMAGE) $(CSI_IMAGE)-$(GOARCH)
+	$(CONTAINER_CMD) push $(CSI_IMAGE)-$(GOARCH)
+	# push amd64 image as default one
+	if [ $(GOARCH) = amd64 ]; then $(CONTAINER_CMD) push $(CSI_IMAGE); fi
 
 clean:
 	go clean -r -x
--- a/deploy.sh
+++ b/deploy.sh
@@ -33,6 +33,42 @@ push_helm_charts() {
 
 }
 
+# Build and push images. Steps as below:
+# 1. get base image from original Dockerfile (FROM ceph/ceph:v14.2)
+# 2. parse manifest to get image digest per arch (sha256:XXX, sha256:YYY)
+# 3. patch Dockerfile with amd64 base image (FROM ceph/ceph:v14.2@sha256:XXX)
+# 4. build and push amd64 image
+# 5. patch Dockerfile with arm64 base image (FROM ceph/ceph:v14.2@sha256:YYY)
+# 6. build and push arm64 image
+build_push_images() {
+	# "docker manifest" requires experimental feature enabled
+	export DOCKER_CLI_EXPERIMENTAL=enabled
+
+	# get baseimg (ceph/ceph:tag)
+	dockerfile="deploy/cephcsi/image/Dockerfile"
+	baseimg=$(awk '/^FROM/ {print $NF}' "${dockerfile}")
+
+	# get image digest per architecture
+	# {
+	#   "arch": "amd64",
+	#   "digest": "sha256:XXX"
+	# }
+	# {
+	#   "arch": "arm64",
+	#   "digest": "sha256:YYY"
+	# }
+	manifests=$("${CONTAINER_CMD:-docker}" manifest inspect "${baseimg}" | jq '.manifests[] | {arch: .platform.architecture, digest: .digest}')
+
+	# build and push per arch images
+	for ARCH in amd64 arm64; do
+		ifs=$IFS; IFS=
+		digest=$(awk -v ARCH=${ARCH} '{if (archfound) {print $NF; exit 0}}; {archfound=($0 ~ "arch.*"ARCH)}' <<< "${manifests}")
+		IFS=$ifs
+		sed -i "s|\(^FROM.*\)${baseimg}.*$|\1${baseimg}@${digest}|" "${dockerfile}"
+		GOARCH=${ARCH} make push-image-cephcsi
+	done
+}
+
 if [ "${TRAVIS_BRANCH}" == 'master' ]; then
 	export ENV_CSI_IMAGE_VERSION='canary'
 else
@@ -42,9 +78,10 @@ fi
 
 if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
 	"${CONTAINER_CMD:-docker}" login -u "${QUAY_IO_USERNAME}" -p "${QUAY_IO_PASSWORD}" quay.io
-	make push-image-cephcsi
 
 	set -xe
+
+	build_push_images
 
 	mkdir -p tmp
 	pushd tmp >/dev/null
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ Independent CSI plugins are provided to support RBD and CephFS backed volumes,
   refer [cephfs doc](https://github.com/ceph/ceph-csi/blob/master/docs/deploy-cephfs.md).
 - For example usage of RBD and CephFS CSI plugins, see examples in `examples/`.
 
+NOTE:
+
+- Ceph CSI **`Arm64`** support is experimental.
+
 ## Project status
 
 Status: **GA**
--- a/deploy/cephcsi/image/Dockerfile
+++ b/deploy/cephcsi/image/Dockerfile
@@ -2,12 +2,14 @@ FROM ceph/ceph:v14.2
 LABEL maintainers="Ceph-CSI Authors"
 LABEL description="Ceph-CSI Plugin"
 
+# To support cross building, do NOT RUN native binaries here.
+# If we have to run native binaries, qemu-user-static must be installed on
+# build host and mounted to container.
+
 # Removing ceph-iscsi repo to workaround the repo issue while upgrading
 #RUN rm -f /etc/yum.repos.d/ceph-iscsi.repo && yum -y update && yum clean all
 ENV CSIBIN=/usr/local/bin/cephcsi
 
 COPY cephcsi $CSIBIN
-
-RUN chmod +x $CSIBIN
 
 ENTRYPOINT ["/usr/local/bin/cephcsi"]
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -44,16 +44,17 @@ function install_minikube() {
     fi
 
     echo "Installing minikube. Version: ${MINIKUBE_VERSION}"
-    curl -Lo minikube https://storage.googleapis.com/minikube/releases/"${MINIKUBE_VERSION}"/minikube-linux-amd64 && chmod +x minikube && mv minikube /usr/local/bin/
+    curl -Lo minikube https://storage.googleapis.com/minikube/releases/"${MINIKUBE_VERSION}"/minikube-linux-"${MINIKUBE_ARCH}" && chmod +x minikube && mv minikube /usr/local/bin/
 }
 
 function install_kubectl() {
     # Download kubectl, which is a requirement for using minikube.
     echo "Installing kubectl. Version: ${KUBE_VERSION}"
-    curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/"${KUBE_VERSION}"/bin/linux/amd64/kubectl && chmod +x kubectl && mv kubectl /usr/local/bin/
+    curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/"${KUBE_VERSION}"/bin/linux/"${MINIKUBE_ARCH}"/kubectl && chmod +x kubectl && mv kubectl /usr/local/bin/
 }
 
 # configure minikube
+MINIKUBE_ARCH=${MINIKUBE_ARCH:-"amd64"}
 MINIKUBE_VERSION=${MINIKUBE_VERSION:-"latest"}
 KUBE_VERSION=${KUBE_VERSION:-"v1.14.2"}
 MEMORY=${MEMORY:-"3000"}


Makefile
```diff
@@ -59,8 +69,17 @@ image-cephcsi: cephcsi
 	$(CONTAINER_CMD) build -t $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) deploy/cephcsi/image
 
 push-image-cephcsi: image-cephcsi
-	$(CONTAINER_CMD) push $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION)
-
+	$(CONTAINER_CMD) tag $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) $(CSI_IMAGE_NAME)-$(ARCH):$(CSI_IMAGE_VERSION)
+	$(CONTAINER_CMD) push $(CSI_IMAGE_NAME)-$(ARCH):$(CSI_IMAGE_VERSION)
+
+# "docker manifest" requires experimental feature enabled
+push-manifest-cephcsi: export DOCKER_CLI_EXPERIMENTAL=enabled
```
Madhu-1:
```
Experimental features provide early access to future product functionality. These features are intended only for testing and feedback as they may change between releases without warning or can be removed entirely from a future release.

Experimental features must not be used in production environments.
```
I would rather go with approach mentioned  in https://github.com/ceph/ceph-csi/issues/671#issuecomment-539912465 which  is currently used by rook

Madhu-1: https://github.com/estesp/manifest-tool says
`
UPDATE (Feb 2018): The Docker client PR #138 is merged, providing a docker manifest command that can replace the use of manifest-tool in most scenarios. Other follow-on PRs are in process to add functionality to docker manifest to make it completely functional for all multi-platform image use cases.
`
@travisn we can enable the experimental feature and use docker cli, any comments?

cyb70289: My first attempt was using manifest-tool. It introduced extra efforts like downloading a third party tool and configuring executable path. Using docker cli directly is more concise.
But I'm not sure if this "docker manifest" command, as an experimental feature, may change in future.

cyb70289: @Madhu-1 @travisn, shall we update this patch and turn to manifest-tool or keep using docker cli experimental feature?

.travis.yml
```diff
@@ -32,12 +32,14 @@ env:
     - DEP_VERSION=v0.5.4
 
 before_install:
+  - mkdir -p $GOPATH/bin
```
Madhu-1: why this is required?

.travis.yml
```diff
@@ -32,12 +32,14 @@ env:
...
   - curl
     https://raw.githubusercontent.com/golang/dep/${DEP_VERSION}/install.sh|sh
 
 jobs:
   include:
     - name: static-check-make
+      arch: amd64
```
Madhu-1: why this is required?

Makefile
```diff
@@ -59,8 +69,17 @@ image-cephcsi: cephcsi
...
+push-manifest-cephcsi:
+	$(CONTAINER_CMD) manifest create \
```
Madhu-1: Can this be part of `push-image-cephcsi` which will be called when the PR is merged?

.travis.yml
cyb70289: Arm build failed without this directory. There may be some minor gaps between arm and x86 golang env.

.travis.yml
cyb70289: This is not required actually, as amd64 is the default arch. I'm putting it here just to make clear arm64 build doesn't run these tests. They can be removed.

Makefile
cyb70289: The problem is that push-manifest-cephcsi should run after both x86 and arm successfully run push-image-cephcsi.

push-image-cephcsi are launched in parallel on x86 and arm nodes to build arch dependent image separately. We don't know which one finishes early.
After they all succeed, push-manifest-cephcsi job is started on one x86 node to update manifest to refer to new built images.

Madhu-1: i have done some testing below are for the reference

https://github.com/Madhu-1/test-travis
https://hub.docker.com/r/madhupr001/multiarch/tags

```
docker manifest inspect madhupr001/multiarch

{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1374,
         "digest": "sha256:fea065be9e2157d01ebd157130fad27cd977e50044c77ad8c2be456733de109a",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1374,
         "digest": "sha256:fea065be9e2157d01ebd157130fad27cd977e50044c77ad8c2be456733de109a",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "v7"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1374,
         "digest": "sha256:fea065be9e2157d01ebd157130fad27cd977e50044c77ad8c2be456733de109a",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
```

cyb70289: Cool! I don't know we can easily cross build go program with GOARCH. It can simplify the code much. No new job is necessary on Arm node.
Thanks @Madhu-1, will update per your reference.

cyb70289: Tested this approach and found some issues.

GOARCH helps to cross build Arm cephcsi image on x86 easily. But [Dockerfile](https://github.com/ceph/ceph-csi/blob/master/deploy/cephcsi/image/Dockerfile) is not easy to handle.
```
FROM ceph/ceph:v14.2

ENV CSIBIN=/usr/local/bin/cephcsi

COPY cephcsi $CSIBIN

RUN chmod +x $CSIBIN

ENTRYPOINT ["/usr/local/bin/cephcsi"]
```
When running on x86, FROM always pulls x86 base image. So the final Arm image contains x86 ceph/ceph base image and Arm cephcsi binary, which is problematic. Image seems okay as entrypoint is Arm binary. Try starting the image on Arm host with "--entrypoint /bin/bash" leads to "exec format error".

To fix this issues, we can replace base image on the fly before building Arm image. Rook uses similar trick. See [1](https://github.com/rook/rook/blob/master/images/ceph/Dockerfile#L16), [2](https://github.com/rook/rook/blob/master/images/ceph/Makefile#L52).

After fixing it, I met with another issue at "RUN chmod +x $CSIBIN". The "chown" binary is from ceph/ceph base image, so it's now an Arm binary and cannot run x86 host. Though it's possible to workaround, things may break if we add new commands later.

@Madhu-1, comments?

Madhu-1: we can follow the similar approach what rook follows,
we need to add a Test  in Travis to use new arm image in ARM architecture

cyb70289: To test cephcsi Arm image is not easy now.
Current test script uses minikube to launch a k8s storage cluster for testing purpose.
It requires some K8s csi sidecar images which are missing on Arm, such as quay.io/k8scsi/csi-provisioner, quay.io/k8scsi/csi-snapshotter, quay.io/k8scsi/csi-attacher. Without them, cephcsi cannot be validated.

Is it acceptable to only build cephcsi Arm image without testing for now(or add a simpler validation for Arm)? If not, I think we have to enable Arm [k8scsi](https://github.com/kubernetes-csi) images first.

Madhu-1: okay, I think we can add the code for testing it and comment it as of now and enable it later when we have the container for ARM architecture?

cyb70289: Yes, will try.

cyb70289:
Major changes:
- Add travis job "cephcsi on arm64" which runs on travis arm64 node to build and test cephcsi image for arm64.
- Update deploy.sh and Makefile to build and push multi-arch images.

NOTE: arm64 test job is disabled by default due to two reasons
- Kubernetes-csi sidecar images for arm64 are missing
- Instead of virtual machine, travis arm64 job runs in LXD, and it denies privileged filesystem access. It causes trouble to launch minikube cluster.

deploy.sh is not run for normal PR. To see full test logs, please check my personal CI job:
- Test log with deploy.sh https://travis-ci.org/cyb70289/ceph-csi/builds/616541161
- Test image pushed to https://hub.docker.com/r/cyb70289/cephcsi/tags

cyb70289:
@Madhu-1 I've updated this patch. Would you review it? Thanks.

Madhu-1:
> Major changes:
> 
> * Add travis job "cephcsi on arm64" which runs on travis arm64 node to build and test cephcsi image for arm64.
> * Update deploy.sh and Makefile to build and push multi-arch images.
> 

> NOTE: arm64 test job is disabled by default due to two reasons
> 
> * Kubernetes-csi sidecar images for arm64 are missing
> * Instead of virtual machine, travis arm64 job runs in LXD, and it denies privileged filesystem access. It causes trouble to launch minikube cluster.
> 
how can we test amd64 images now?

> deploy.sh is not run for normal PR. To see full test logs, please check my personal CI job:
> 
> * Test log with deploy.sh https://travis-ci.org/cyb70289/ceph-csi/builds/616541161
> * Test image pushed to https://hub.docker.com/r/cyb70289/cephcsi/tags

.travis.yml
```diff
@@ -75,6 +76,17 @@ jobs:
         - make image-cephcsi || travis_terminate 1;
         - scripts/travis-functest.sh v1.15.0 || travis_terminate 1;
 
+    - name: cephcsi on Arm64
```
Madhu-1: default machines provided by Travis will be amd64. i believe this should be amd64 not arm64.

deploy.sh
```diff
@@ -41,11 +83,10 @@ else
 fi
 
 if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
-	"${CONTAINER_CMD:-docker}" login -u "${QUAY_IO_USERNAME}" -p "${QUAY_IO_PASSWORD}" quay.io
```
Madhu-1: let the docker login be here only and call `build_push_images` here only

cyb70289:
> > Major changes:
> > 
> > * Add travis job "cephcsi on arm64" which runs on travis arm64 node to build and test cephcsi image for arm64.
> > * Update deploy.sh and Makefile to build and push multi-arch images.
> 
> > NOTE: arm64 test job is disabled by default due to two reasons
> > 
> > * Kubernetes-csi sidecar images for arm64 are missing
> > * Instead of virtual machine, travis arm64 job runs in LXD, and it denies privileged filesystem access. It causes trouble to launch minikube cluster.
> 
> how can we test amd64 images now?
> 

I tested ceph-csi arm64 image in a local k8s cluster running inside VM, and with all necessary sidecar images manually built.

> > deploy.sh is not run for normal PR. To see full test logs, please check my personal CI job:
> > 
> > * Test log with deploy.sh https://travis-ci.org/cyb70289/ceph-csi/builds/616541161
> > * Test image pushed to https://hub.docker.com/r/cyb70289/cephcsi/tags

Madhu-1:
> > > Major changes:
> > > 
> > > * Add travis job "cephcsi on arm64" which runs on travis arm64 node to build and test cephcsi image for arm64.
> > > * Update deploy.sh and Makefile to build and push multi-arch images.
> > 
> > 
> > > NOTE: arm64 test job is disabled by default due to two reasons
> > > 
> > > * Kubernetes-csi sidecar images for arm64 are missing
> > > * Instead of virtual machine, travis arm64 job runs in LXD, and it denies privileged filesystem access. It causes trouble to launch minikube cluster.
> > 
> > 
> > how can we test amd64 images now?
> 
> I tested ceph-csi arm64 image in a local k8s cluster running inside VM, and with all necessary sidecar images manually built.

how can we test  it on CI?
> 
> > > deploy.sh is not run for normal PR. To see full test logs, please check my personal CI job:
> > > 
> > > * Test log with deploy.sh https://travis-ci.org/cyb70289/ceph-csi/builds/616541161
> > > * Test image pushed to https://hub.docker.com/r/cyb70289/cephcsi/tags

cyb70289:
> > > > Major changes:
> > > > 
> > > > * Add travis job "cephcsi on arm64" which runs on travis arm64 node to build and test cephcsi image for arm64.
> > > > * Update deploy.sh and Makefile to build and push multi-arch images.
> > > 
> > > 
> > > > NOTE: arm64 test job is disabled by default due to two reasons
> > > > 
> > > > * Kubernetes-csi sidecar images for arm64 are missing
> > > > * Instead of virtual machine, travis arm64 job runs in LXD, and it denies privileged filesystem access. It causes trouble to launch minikube cluster.
> > > 
> > > 
> > > how can we test amd64 images now?
> > 
> > 
> > I tested ceph-csi arm64 image in a local k8s cluster running inside VM, and with all necessary sidecar images manually built.
> 
> how can we test it on CI?
> 

I'm afraid there's no easy way to test ceph-csi fully in arm64 CI now. The major barrier is travis arm64 node runs in non-priviledged LXD which prohibits launching minikube cluster. Without actual k8s environment, I think ceph-csi cannot be tested thoroughly.

> > > > deploy.sh is not run for normal PR. To see full test logs, please check my personal CI job:
> > > > 
> > > > * Test log with deploy.sh https://travis-ci.org/cyb70289/ceph-csi/builds/616541161
> > > > * Test image pushed to https://hub.docker.com/r/cyb70289/cephcsi/tags

mykaul:
Can't we use https://aws.amazon.com/blogs/aws/new-ec2-instances-a1-powered-by-arm-based-aws-graviton-processors/ ?

mykaul:
Or CentOS CI?
Or Ceph CI has some ARM support?

Madhu-1:
CC @nixpanic

cyb70289:
> Or CentOS CI?
> Or Ceph CI has some ARM support?

Ceph-csi is using travis CI. To leverage machines from amazon or other CI, we have to add a new set of CI code to run in parallel with travis. Looks no good.

Madhu-1:
we have a plan to use centos CI for ceph-csi, CC'ed @nixpanic to provide more information

cyb70289:
Posted a travis issue to try replacing LXD with VM:
https://travis-ci.community/t/arm64-unprivileged-lxd-blocks-kubernetes-tests/6272

nixpanic:
On Tue, Dec 03, 2019 at 02:05:18AM -0800, Yaniv Kaul wrote:
> Or CentOS CI?
> Or Ceph CI has some ARM support?

The CentOS CI does not have ARM hardware available at this time:
- https://wiki.centos.org/QaWiki/PubHardware

I do not know about the Ceph infrastructure.

Travis does seem to have some support for ARM.

.travis.yml
cyb70289: Next line "arch: arm64" tells travis to schedule this job to arm64 node.
We can validate it by checking the CI log of this patch: https://travis-ci.org/ceph/ceph-csi/builds/616549765

deploy.sh
cyb70289: will change

humblec:
Isnt it good to target this for v2.0.0 ?

Madhu-1:
@humblec we cannot test ARM images on Travis CI, we have two options 1) we need a CI to test arm images 2) declare ARM images as experimental(not tested)

travisn:
It's important to make this feature available, even if it is marked experimental initially. If the only reason is that the CI doesn't cover it, the risk with golang is fairly low at least from what we've seen in rook.

Madhu-1:
> It's important to make this feature available, even if it is marked experimental initially. If the only reason is that the CI doesn't cover it, the risk with golang is fairly low at least from what we've seen in rook.

in that case, will take it up for 2.0.0 release  @humblec @nixpanic @ShyamsundarR PTAL

humblec:
Sure.  I am fine to take this in for `release-2.0.0`. The only thing which I could suggest here is, updating the documentation that its experimental. @cyb70289 can you please update the README with this ?

humblec:
@cyb70289 can you also mention the subjected minikube issue `- CI test jobs are disabled on Arm now build due to minikube issue.` referred here in the code or source comments ?

cyb70289:
@humblec Cool! Will update soon.

cyb70289:
diffs to last commit:
- add note "ceph csi Arm64 support is experimental" to README.md
- add reasons why disabling Arm64 test job in .travis.yml

humblec:
Thanks @cyb70289

humblec:
@Madhu-1 @nixpanic ptal.

Makefile
```diff
@@ -52,15 +58,23 @@ func-test:
...
 	cp _output/cephcsi deploy/cephcsi/image/cephcsi
-	$(CONTAINER_CMD) build -t $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION) deploy/cephcsi/image
+	chmod +x deploy/cephcsi/image/cephcsi
+	$(CONTAINER_CMD) build -t $(CSI_IMAGE) deploy/cephcsi/image
 
 push-image-cephcsi: image-cephcsi
-	$(CONTAINER_CMD) push $(CSI_IMAGE_NAME):$(CSI_IMAGE_VERSION)
-
+	$(CONTAINER_CMD) tag $(CSI_IMAGE) $(CSI_IMAGE)-$(GOARCH)
+	$(CONTAINER_CMD) push $(CSI_IMAGE)-$(GOARCH)
```
Madhu-1: @cyb70289 want final confirmation before merging this PR, can we push MultiArch images to quay.io? have you tested it

cyb70289: @Madhu-1 my free quay.io account does NOT support multiarch image, as we've talked in [this thread](https://github.com/ceph/ceph-csi/issues/671#issuecomment-548271431). I validated multiarch pushing code with my dockerhub account.
Quay.io v3.0 [announces multiarch support](https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/red_hat_quay_release_notes/index#multi_architecture_container_image_support). I don't know if ceph-csi registry is hosted by quay.io v3.0. It's better to try it offline first.

cyb70289: Below commands are for a quick test. It pushes multiarch image of busybox to a repo. You may adjust per your requirements.

# set docker registry
export REPO=quay.io/cephcsi

# login registry, input user, pass
docker login quay.io

# push busybox amd64 and arm64 images
docker pull busybox
docker pull arm64v8/busybox

docker tag busybox:latest $REPO/busybox-amd64:latest
docker tag arm64v8/busybox:latest $REPO/busybox-arm64:latest

docker push $REPO/busybox-amd64:latest
docker push $REPO/busybox-arm64:latest

# push multi arch image
export DOCKER_CLI_EXPERIMENTAL=enabled

docker manifest create --amend $REPO/busybox:latest $REPO/busybox-amd64:latest $REPO/busybox-arm64:latest

docker manifest annotate $REPO/busybox:latest $REPO/busybox-amd64:latest --os linux --arch amd64
docker manifest annotate $REPO/busybox:latest $REPO/busybox-arm64:latest --os linux --arch arm64

docker manifest push $REPO/busybox:latest

Madhu-1: @humblec can you please test above one?

humblec:
@Madhu-1 sure.. I will revert once I verify this.

hh:
Note that we can ask CNCF / Packet for an arm node to help with the CI.
Let me know if I can help coordinate that.

https://github.com/rook/rook/issues/4051#issuecomment-569386390

humblec:
@cyb70289 I was testing this in my system and facing below issue:
```
[root@localhost ceph-csi]# docker manifest create --amend $REPO/busybox:latest $REPO/busybox-amd64:latest $REPO/busybox-arm64:latest
unsupported manifest media type and no default available: application/vnd.docker.distribution.manifest.v1+prettyjws
[root@localhost ceph-csi]# docker version
Client: Docker Engine - Community
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea838
 Built:             Wed Nov 13 07:26:10 2019
 OS/Arch:           linux/amd64
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       633a0ea838
  Built:            Wed Nov 13 07:24:42 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
[root@localhost ceph-csi]# 

```

Any idea whats causing this ? I came across this [1] https://github.com/docker/cli/issues/1724

Btw, below commands were successful:

```

docker pull busybox
docker pull arm64v8/busybox

docker tag busybox:latest $REPO/busybox-amd64:latest
docker tag arm64v8/busybox:latest $REPO/busybox-arm64:latest

docker push $REPO/busybox-amd64:latest
docker push $REPO/busybox-arm64:latest
```

If this is good enough, please let me know we can go ahead .

cyb70289:
@humblec @Madhu-1 From the error message, looks quay.io/cephcsi does not support multiarch manifest. Same commands run correctly on dockerhub. Maybe I'm missing something but current patch will break cephcsi image pushing process. We should NOT go ahead.

If we cannot switch to a docker registry with multiarch support(quay.io v3 or dockerhub) recently, I would recommend a temporary solution(similar approach used by [flannel](https://quay.io/repository/coreos/flannel?tab=tags)):

1. build and push x86 image quay.io/cephcsi/cephcsi:v2.0.0-amd64
2. build and push arm image quay.io/cephcsi/cephcsi:v2.0.0-arm64
3. re-tag and push x86 image as quay.io/cephcsi/cephcsi:v2.0.0

So the default image is still x86, user can pull arm image by appending "-arm64" after the tag.

humblec:
>
    build and push x86 image quay.io/cephcsi/cephcsi:v2.0.0-amd64
    build and push arm image quay.io/cephcsi/cephcsi:v2.0.0-arm64
    re-tag and push x86 image as quay.io/cephcsi/cephcsi:v2.0.0

So the default image is still x86, user can pull arm image by appending "-arm64" after the tag.
>

@cyb70289 ACK. lets do that !

billimek:
So we won't be able to have multi-arch images, but instead specifically-named images for certain types?

Madhu-1:
> So we won't be able to have multi-arch images, but instead specifically-named images for certain types?

@cyb70289 yes that a temp solution for now till quay supports it

billimek:
> > So we won't be able to have multi-arch images, but instead specifically-named images for certain types?
> 
> @cyb70289 yes that a temp solution for now till quay supports it

Thanks for the clarification.  I would suggest re-naming this PR to indicate that what is being delivered are multiple _different_ arch images and not multi-arch images.  These are very much different things and it would be important to note the distinction for those who are expecting to be able to run these images in a multiple-arch cluster.

Also, it looks like quay 3.0 started supporting multi-arch images based on [these release notes](https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/red_hat_quay_release_notes/index#multi_architecture_container_image_support) but not sure what are the mechanics of using this new version for cephcsi.

Madhu-1:
> > > So we won't be able to have multi-arch images, but instead specifically-named images for certain types?
> > 
> > 
> > @cyb70289 yes that a temp solution for now till quay supports it
> 
> Thanks for the clarification. I would suggest re-naming this PR to indicate that what is being delivered are multiple _different_ arch images and not multi-arch images. These are very much different things and it would be important to note the distinction for those who are expecting to be able to run these images in a multiple-arch cluster.
> 
Yes PR title needs to be changed,  we  will update doc to use this image if you are running in  different architecture 

> Also, it looks like quay 3.0 started supporting multi-arch images based on [these release notes](https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/red_hat_quay_release_notes/index#multi_architecture_container_image_support) but not sure what are the mechanics of using this new version for cephcsi.

AFAIK ceph-csi uses free service provided by the quay, so we  cant use multi-arch support for now

nixpanic:
> AFAIK ceph-csi uses free service provided by the quay, so we cant use multi-arch support for now

This is not relevant. Quay.io does not support multi-arch images for now, it is a feature that will be added later. Reply from Quay Support on the topic:

> Mutliarch images are supported through the Docker v2 schema 2, which is yet not generally enabled. This feature is still under
> testing and will available for usage after next general roll out.

Madhu-1:
@humblec PTAL

cyb70289:
Thanks @Madhu-1 @humblec 
I see cephcsi arm64 image is pushed, great !
https://quay.io/repository/cephcsi/cephcsi?tab=tags

Madhu-1:
@cyb70289 Thanks for the work!

PR #712, opened 2019-11-13
cephfs: add kernel version detection for mounting with client

Madhu-1:
Linux kernel 4.17.0 adds support for quota with CephFS. Without quota,
it is not possible to fullfill the requirements of the CSI Spec and
guarantee sufficient space on the filesystem for a volume. With this in
mind, usage of the kernel client is only allowed with kernel 4.17.0 or
newer.

However, some Linux vendors backport features and patches to their
Enterprise products. These kernels may have an older version, but do
support quota. One of these is the kernel that comes with RHEL-7.7.

By comparing the current running version of the Linux kernel against
known versions that support quota, we can now automatically decide to
use the kernel client, or not.

Note that this does not change the 'forcekernelclient' parameter. The
parameter is still available and can be used for kernels that are not in
the 'known to support quota list'. Or users can pass the parameter to
use a CephFS kernel client that does not support quota.

Signed-off-by: Niels de Vos <ndevos@redhat.com>
(cherry picked from commit 290beb4dda802beaab32a20177e58ca0ac0d1121)

Do you have any questions?

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1366,6 +1366,7 @@
     "github.com/pkg/errors",
     "github.com/prometheus/client_golang/prometheus",
     "github.com/prometheus/client_golang/prometheus/promhttp",
+    "golang.org/x/sys/unix",
     "google.golang.org/grpc",
     "google.golang.org/grpc/codes",
     "google.golang.org/grpc/status",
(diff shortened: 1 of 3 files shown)


Madhu-1:
will merge this once CI passes

PR #713, opened 2019-11-13
setup Travis CI for  release-4.2 branch

Madhu-1:
setup Travis CI for  release-4.2 branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ language: go
 branches:
   only:
     - master
+    - release-4.2
 
 go: 1.12.x
 


Madhu-1:
@ShyamsundarR PTAL

ShyamsundarR:
@Madhu-1 Sorry, I must have been living under a rock, what is this `release-4.2` branch all about?

Madhu-1:
@ShyamsundarR This is the branch used for the downstream build

ShyamsundarR:
> @ShyamsundarR This is the branch used for the downstream build

Thanks, but since when? Also, why is a downstream (DS) requirement making it to the upstream repositories? Is the plan to have multiple such DS branches in the future?

In short, this seems incorrect to use the upstream repository to maintain DS branches and needs. So if there is an issue tracking the same with better explanation on why we need this and what happens in the future it would be helpful.

Not attempting to stonewall this patch, but want to know some of the details before this gets out of control.

nixpanic:
Hmm, yeah, I agree with Shyam, downstream/product branches should not be part of upstream projects. This branch should be removed once a special (internal?) downstream repository is available.

Issue #721, opened 2019-12-01
Not working e2e framework with no option kubeconfig


WoohyungHan:
# Describe the bug #
If we try to run e2e test without kubeconfig option, It does not found kubeconfig file. Its because k8s e2e framework try to find kubeconfig file using environment variable: KUBECONFIG. 

# Environment details #
- all version

# Steps to reproduce #
1. run e2e test without --kubeconfig option

# Actual results #
Not set kubeconfig correctly

# Expected behavior #
running e2e test

# Additional context #
NONE

PR #722, opened 2019-12-01
Fix default kubeconfig working correctly in e2e test

WoohyungHan:
## Describe what this PR does #
Kuberentes e2e framework get kubeconfig information from
KUBECONFIG environment variable. set the variable by default.

NONE

## Related issues ##
Fixes: #721

## Future concerns ##
NONE

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -4,6 +4,8 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"os"
+	"path/filepath"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


e2e/e2e_test.go
```diff
@@ -32,6 +37,14 @@ func init() {
 	fmt.Println("timeout for deploytimeout ", deployTimeout)
 }
 
+func setDefaultKubeconfig() {
+	_, exists := os.LookupEnv("KUBECONFIG")
+	if !exists {
+		defaultKubeconfig := filepath.Join(os.Getenv("HOME"), ".kube", "config")
```
Madhu-1: This will only work if you are running E2E on the same machine where kubernetes is deployed

it will fail if you want to run E2E against the remote cluster, I think setting `KUBECONFIG` is the best option

WoohyungHan: Hi, @Madhu-1 :) I'm sorry for lack of explanation. I think this code is same what you're saying. It is not related with real kubernetes cluster but it related with kubeconfig file. In our e2e README.md, default kubeconfig is $HOME/.kube/config. This code just set KUBECONFIG envorinment variable as $HOME/.kube/config. Of course, this code works with remote kubernetes cluster.

```
// in our e2e docs...
kubeconfig | Path to kubeconfig containing embedded authinfo (default: $HOME/.kube/config)
```

e2e/e2e_test.go
```diff
@@ -32,6 +37,14 @@ func init() {
...
+		os.Setenv("KUBECONFIG", defaultKubeconfig)
```
Madhu-1: as we are setting this ENV now do we need to unset this ENV after finishing testing?

Madhu-1: instead of making code changes can we update the readme to set the KUBECONFIG env before running e2e?

WoohyungHan: I don't think so. The ENV variables are only set for the current process and it's child processes. So if e2e tests is ended, the variable is gone.

WoohyungHan: > instead of making code changes can we update the readme to set the KUBECONFIG env before running e2e?

I think we can just remove the kubeconfig default option. but all other flags are having default option and most of user do not want to modify kubectl config path.

Madhu-1: as the env will be set in the current process and also this code honors the e2e readme documentation, am fine with the changes

Issue #724, opened 2019-12-02
Move rook-deploy code from e2e to ./scripts/minikube.sh


WoohyungHan:
# Describe the feature you'd like to have #
We have the e2e test with --deploy-rook=true that makes all test environment. It works fine, but It does not seem to be the role of e2e test. In addition, when developing the code we need to run full test scenario with deploying rook every time, or we need to build rook environment by hand. I think we can move this to ./scripts/minikube.sh.

If we make new developing environment, we just do
```shell
./scripts/minkube.sh up
./scripts/minikube.sh cephcsi
./scripts/minikube.sh k8s-sidecar
./scripts/minikube.sh deploy-rook
```

# What is the value to the end user? (why is it a priority?) #
No value for end user.

# How will we know we have a good solution? (acceptance criteria) #
- Easy to build development environment

# Additional context #
NONE

PR #727, opened 2019-12-03
Move rook-deploy code from e2e to ./scripts/minikube.sh

WoohyungHan:
We have the e2e test with --deploy-rook=true that makes all test
environment. It works fine, but It does not seem to be the role of
e2e test. In addition, when developing the code we need to run full
test scenario with deploying rook every time, or we need to build
rook environment by hand. Move rook-deploy code to minikube.sh

NONE

## Related issues ##
Closes: #724

## Future concerns ##
NONE


scripts/minikube.sh
```diff
@@ -53,6 +53,33 @@ function install_kubectl() {
...
+    ROOK_URL=https://raw.githubusercontent.com/rook/rook/$ROOK_VERSION/cluster/examples/kubernetes/ceph
+
+    kubectl create -f $ROOK_URL/common.yaml
+    kubectl create -f $ROOK_URL/operator.yaml
+    kubectl create -f $ROOK_URL/cluster-test.yaml
+    kubectl create -f $ROOK_URL/toolbox.yaml
+
+    for ((retry=0; retry<=${ROOK_DEPLOY_TIMEOUT}; retry++)); do
+        echo "Wait for rook deploy... ${retry}s"
+        sleep 1
```
Madhu-1: move this to a new file, so people who want to deploy rook for testing can use it  easily

Madhu-1: increase sleep to 5Sec?

WoohyungHan: That looks better, I will fix it.

WoohyungHan: If we move to a new file (maybe scripts/deploy_rook.sh), testing will be like this.

```shell
./scripts/minikube.sh up
./scripts/depoy_rook.sh
(...)
```

how about just don't move and use minikube.sh?
```shell
./scripts/minikube.sh up
./scripts/minikube.sh deploy-rook
(...)
```

Madhu-1: let it be, it will be more useful for the developers who will bring up there own Kubernetes cluster other than minikube

WoohyungHan: > let it be, it will be more useful for the developers who will bring up there own Kubernetes cluster other than minikube

hmm, that make sense. I will fix it.

WoohyungHan: I moved the code like your suggestion. but I keep ./scripts/minikube.sh deploy-rook. now, user who want to deploy own cluster can run ./scripts/rook.sh file but others can run just ./scripts/minikube.sh.

scripts/rook.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash -e
```
Madhu-1: why this is required?

scripts/minikube.sh
```diff
@@ -99,6 +99,18 @@ ssh)
...
+deploy-rook)
+    echo "deploy rook"
+    ./scripts/rook.sh deploy
+    ;;
+clean-rook)
+    echo "clean rook"
+    ./scripts/rook.sh clean
+
+    # delete rook data for minikube
+    minikube ssh "sudo rm -rf /mnt/${DISK}/var/lib/rook"
```
Madhu-1: can you please check is this directory correct?

scripts/rook.sh
WoohyungHan: I just copied from minikube.sh and I think it means bash script. or i missed something?

scripts/minikube.sh
WoohyungHan: in the minkube.sh, we have following code to create a link so the default dataDirHostPath will work.
```
minikube ssh "sudo mkdir -p /mnt/${DISK}/var/lib/rook; sudo ln -s /mnt/${DISK}/var/lib/rook /var/lib/rook"
```

I want to delete both link(ln -s) and real directory (mkdir -p), and recreate it.

scripts/rook.sh
Madhu-1: I mean we have rook.sh and deploy-rook.sh in the script folder, do we need both?

WoohyungHan: oh, It was my mistake. I deleted it. Thanks!

scripts/rook.sh
```diff
@@ -0,0 +1,51 @@
...
+
+ROOK_VERSION=${ROOK_VERSION:-"v1.1.7"}
+ROOK_DEPLOY_TIMEOUT=${ROOK_DEPLOY_TIMEOUT:-300}
+
+function deploy_rook() {
+    ROOK_URL=https://raw.githubusercontent.com/rook/rook/${ROOK_VERSION}/cluster/examples/kubernetes/ceph
```
Madhu-1: define `ROOK_URL` above and use it in  both deploy_rook and celan_rook ( i prefer to call it as  teadown_rook)

WoohyungHan: > define ROOK_URL above and use it in both deploy_rook and celan_rook ( i prefer to call it as teadown_rook)

nice catch, I fix that ROOK_URL is used by both. and yes, teardown is look clear than clean. I will fix it.

humblec:
@WoohyungHan Thanks for the PR, can you please rebase?

WoohyungHan:
> @WoohyungHan Thanks for the PR, can you please rebase?

Sure :) Done

e2e/e2e_test.go
```diff
@@ -45,74 +39,12 @@ func setDefaultKubeconfig() {
 	}
 }
 
-// removeCephCSIResource is a temporary fix for CI to remove the ceph-csi resources deployed by rook
-func removeCephCSIResource() {
```
ShyamsundarR: (clarification) why is this no longer needed? Rook still rolls out the CSI artifacts, afaict.

scripts/rook.sh
```diff
@@ -0,0 +1,48 @@
...
+    kubectl create -f "${ROOK_URL}/cluster-test.yaml"
+    kubectl create -f "${ROOK_URL}/toolbox.yaml"
+    kubectl create -f "${ROOK_URL}/filesystem-test.yaml"
+    kubectl create -f "${ROOK_URL}/pool-test.yaml"
+
+    for ((retry=0; retry<=ROOK_DEPLOY_TIMEOUT; retry=retry+5)); do
+        echo "Wait for rook deploy... ${retry}s"
+        sleep 5
+
+        if kubectl get cephclusters -n rook-ceph | grep HEALTH_OK &> /dev/null; then
```
ShyamsundarR: (nit) is this enough or should we check and wait till the pool and fs are created, just so the automated testing does not rush into testing CSI operations against the same?

WoohyungHan: I had very considered which one is best to know the end of rook-ceph deploy. and I didn't think about that. however i tried with that (rbdpool and fs) it just created after yaml file is run, not ceph deploy. there is no field like status, so I think we can't get result by this.

```shell
whhan91@whhan91-pc:~/go/src/github.com/ceph/ceph-csi$ kubectl get cephfilesystems.ceph.rook.io -A -o wide
NAMESPACE   NAME   ACTIVEMDS   AGE
rook-ceph   myfs   1           7s

whhan91@whhan91-pc:~/go/src/github.com/ceph/ceph-csi$ kubectl get cephblockpools.ceph.rook.io -A -o wide
NAMESPACE   NAME          AGE
rook-ceph   replicapool   15s
```

e2e/e2e_test.go
WoohyungHan: Hi @ShyamsundarR Thanks for your review :) I recheck the file, and found the code delete following objects.

```
csi-cephfsplugin
csi-rbdplugin
csi-rbdplugin-provisioner
csi-cephfsplugin-provisioner
csi-rbdplugin-provisioner
csi-cephfsplugin-provisioner

rbd provisioner rbac
rbd nodeplugin rbac
cephfs provisioner rbac
cephfs nodeplugin rbac
```

but our e2e code looks like deploy this and delete after e2e.

```golang
func deployRBDPlugin() {
	// deploy provisioner
	framework.RunKubectlOrDie("create", "-f", rbdDirPath+rbdProvisioner)
	framework.RunKubectlOrDie("apply", "-f", rbdDirPath+rbdProvisionerRBAC)
	// deploy nodeplugin
	framework.RunKubectlOrDie("create", "-f", rbdDirPath+rbdNodePlugin)
	framework.RunKubectlOrDie("apply", "-f", rbdDirPath+rbdNodePluginRBAC)
}

func deleteRBDPlugin() {
	_, err := framework.RunKubectl("delete", "-f", rbdDirPath+rbdProvisioner)
	if err != nil {
		e2elog.Logf("failed to delete rbd provisioner %v", err)
	}
	_, err = framework.RunKubectl("delete", "-f", rbdDirPath+rbdProvisionerRBAC)
	if err != nil {
		e2elog.Logf("failed to delete provisioner rbac %v", err)
	}
	_, err = framework.RunKubectl("delete", "-f", rbdDirPath+rbdNodePlugin)
	if err != nil {
		e2elog.Logf("failed to delete nodeplugin %v", err)
	}
	_, err = framework.RunKubectl("delete", "-f", rbdDirPath+rbdNodePluginRBAC)
	if err != nil {
		e2elog.Logf("failed to delete nodeplugin rbac %v", err)
	}
}
```

Madhu-1: @WoohyungHan all CSI related resources are created in rook-ceph namespace. only thing is you need to delete the cluster role and cluster role binding for both cephfs and rbd created by the rook.

as we are using `kubectl  apply`  while creating RBAC for csi this will be updated with new changes

https://travis-ci.org/ceph/ceph-csi/jobs/620627227#L471-L477
https://travis-ci.org/ceph/ceph-csi/jobs/620627227#L749-L757

I think a separate patch should be good to remove RBAC created by rook and replace `apply`  with `create`

scripts/rook.sh
Madhu-1: you can check for MDS pod once you  create the filesystem (make sure at least one pod is created) https://github.com/ceph/ceph-csi/blob/master/e2e/deploy-rook.go#L43

in current e2e we are not checking whether the block pool is created or not, with the help of toolbox pod we can check block pool is created or not) this also can be a separate PR

WoohyungHan: I post a new issue to handle this. #730

e2e/e2e_test.go
WoohyungHan: I added this as issue #731

PR #729, opened 2019-12-04
[RFC] rewrite RBD provisioning to use go-ceph instead of rbd command

nixpanic:

This is a PR that should not be merged as is, but is used for Request For Comments.

There is a need to extend go-ceph with certain functions so that the current RBD provisioner can be rewritten to use the librbd.so library directly. This should increase performance.

Initial changes, and planned work, in no particular order:
* implement `createVolume()`
* implement `deleteVolume()`, foreground only, not as a background task
* re-use connection for multiple RBD images/operations with the same parameters (StorageClass)
* more object oriented programming, there is already a `rbdVolume` object that can have functions
* implement cloning
* implement snapshots
* ...

The changes in the branch used to create this PR depend on a work-in-progress version of go-ceph. Hopefully it will be possible to merge parts of this [RFC] PR when go-ceph functionality gets merged,

## Related issues ##

See-also: #449


humblec:
@nixpanic thanks ! one question, instead of referring to your fork of ceph-go, cant we refer directly to go-ceph master branch? is it the case that, there are unmerged PRs in master against your fork ?

nixpanic:
There are currently changes under review in the go-ceph project that this PR takes advantage of. Hence the `Godep.toml` points to the `wip/for-ceph-csi` branch in my repository instead of the master branch upstream.

pkg/rbd/rbd_util.go
```diff
@@ -107,64 +108,86 @@ var (
...
+		return nil, err
 	}
 
-	// If command never succeed, returns its last error.
+	ioctx, err := conn.OpenIOContext(rv.Pool)
 	if err != nil {
-		return false, output, err
+		return nil, err
 	}
+	defer ioctx.Destroy()
```
ShyamsundarR: This defer would be executed during the function return, thus destroying the created ioctx?

pkg/rbd/rbd_util.go
```diff
@@ -107,64 +108,86 @@ var (
...
+	return conn, nil
+}
 
-	if err, ok := err.(*exec.Error); ok {
-		if err.Err == exec.ErrNotFound {
-			klog.Errorf(util.Log(ctx, "rbd cmd not found"))
-			// fail fast if command not found
-			return false, output, err
-		}
+func (rv *rbdVolume) getIoctx(cr *util.Credentials) (*rados.IOContext, error) {
```
ShyamsundarR: Is the intention to request for an IOCtx each time it is required? Because, @Madhu-1 worked with an prototype like this and it did not yield any benefits to performance, as we establish the connection and io context per CLI/API call.

Instead I would like to see the credentials function 'NewUserCredentials` enhanced to store a connection and an io context, for use across any API that we call for the said request. The credentials are currently setup to use a defer based destroy where these can be garbage collected.

I had gone with the above [approach](https://github.com/ShyamsundarR/ceph-csi/commit/381dca46c37af18d9f2ade0549cc056bcb95b3ab) and it yielded much better results as the said operations are done once across the various calls being made by the driver for a particular RPC request.

dillaman: The librados/librbd C APIs are thread-safe so in theory you really only need to share a single connection. If you implement a connection pool in the CSI somewhere for a given set of cluster credentials, it could return a cached connection. We don't want to keep opening a new connection for each CSI request since for large clusters that would still take up the majority of CSI-time to bootstrap up the connection.

nixpanic: I am planning to keep the ioctx around, and have it re-used. Basically one ioctx per StorageClass (options, credentials, ..) and have it garbage collected after a timeout.

pkg/rbd/rbd_util.go
nixpanic: Uhm, yes, I think so.

pkg/rbd/rbd_util.go
```diff
@@ -107,64 +108,95 @@ var (
...
 
+	image, err := librbd.CreateWithOptions(ioctx, pOpts.RbdImageName, uint64(volSz), options)
 	if err != nil {
-		return errors.Wrapf(err, "failed to create rbd image, command output: %s", string(output))
+		return errors.Wrapf(err, "failed to create rbd image")
 	}
 
+	// image is not returned to the caller
+	// no need to Close(), the image is not open after Create()
+	_ = image
```
WoohyungHan: just curios, what is the meaning of this line?

humblec:
@nixpanic as we are tracking this for `2.0.0` , can you please refresh the subjected issue/PR if you have any update ?

nixpanic:
A cleaned-up version, using the upstream available go-ceph API is available at #777. That PR only implements CreateVolume, this PR will get updated with more functionality while go-ceph gets extended.

pkg/rbd/rbd_util.go
nixpanic: This prevents Go from complaining that the `image` variable is unused.

dillaman: Why is an "image" being returned?

nixpanic: It is not needed, I'll drop it and move the comments to just before the `librbd.Create4()` call.

nixpanic:
Something is weird here, compiling on arm64 fails as `rados_read_op_omap_get_vals2` can not be found. The required packages seen to have been installed :confused:

nixpanic:
Okay, so there are no official up-to-date Ceph Ubuntu packages for arm64... The `.travis.yml` now contains a fall-back repository for Ceph Nautilus 14.2.5, which gets added only on aarch64.

nixpanic:
The majority of calls has been replaced by a go-ceph implementation. Follow the [Use go-ceph for volume management operations](https://github.com/ceph/ceph-csi/projects/3) project for an updated status.

PR #754, opened 2019-12-18
Fix block resize issue in RBD

Madhu-1:
For the filesystem resize the staging path will be passed to NodeExpandVolume and for the block
resize the target path will be passed. Added code to handle resize for both block and file system resize
    
### DIFF
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -543,6 +543,20 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	// volumePath is targetPath for block PVC and stagingPath for filesystem.
+	// check the path is mountpoint or not, if it is
+	// mountpoint treat this as block PVC or else it is filesystem PVC
+	// TODO remove this once ceph-csi supports CSI v1.2.0 spec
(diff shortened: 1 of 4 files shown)


humblec:
@Madhu-1 just waiting for the e2e to be corrected.

pkg/rbd/nodeserver.go
```diff
@@ -543,6 +543,19 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	// volumePath is targetPath for block PVC and stagingPath for filesystem.
+	// check the path is mountpoint or not, if it is
+	// mountpoint treat this as block PVC or else it is filesystem PVC
+	// TODO remove this once ceph-csi supports CSI v1.2.0 spec
+	notMnt, err := mount.IsNotMountPoint(ns.mounter, volumePath)
+	if err != nil {
+		if !os.IsNotExist(err) {
```
ShyamsundarR: I am unable to reason out why we treat `IsNotExist` error as not an error and proceed. What I get is that, this will return `notMnt` as true, and then the call at line 559 will error out as the parent path of the stash is absent. Should we be treating this error specially here?

Madhu-1: updated to handle IsNotExist

Madhu-1:
@humblec @ShyamsundarR PTAL CI should pass now

note: Block Resize is supported in kubernetes 1.16+

PR #758, opened 2019-12-19
added csiImageKey that allows us to keep track of the image name in omap

skoef:

This PR prepares ceph-csi to be able to use different volume name prefixes. Currently, all images are prefixed with `csi-vol-` and this is assumed every time the image name becomes relevant. Instead of assuming this name, we store it in the named omap under the `csi.imagename` key. This way, it shouldn't matter if at some point a different naming prefix is used, because the name will be stored in the omap upon creation of the image and referred to this name further on.

This PR applies to both volumes and snapshots.

At this point, this only works on pool that wasn't used for Ceph-CSI before. The reason for that is the `csi.imagename` is not set for volumes created before and there is currently no mechanism of synchronising the `csi.imagename` for existing volumes.

I could create something that on startup of the controllerserver would fetch all volumes, looks for volumes without the new key and then tries to find the images based on the UID suffix.

## Related issues ##

Fixes: #487

## Future concerns ##

Follow-up PR could add a `--volume-name-prefix` and `--snapshot-name-prefix` flag, so we can choose a prefix freely, perhaps even allowing templating the prefix, based on the volumeRequest!


skoef:
Little proof of concept: after I've created a PVC through kubernetes, this is what I end up with:

```bash
$ sudo rados -p data  listomapvals csi.volumes.default
csi.volume.pvc-157f9a6c-3123-4745-a0a5-c59a696cfe3f
value (36 bytes) :
00000000  38 61 63 62 37 64 62 65  2d 32 32 34 62 2d 31 31  |8acb7dbe-224b-11|
00000010  65 61 2d 39 39 64 33 2d  39 32 64 61 37 34 61 32  |ea-99d3-92da74a2|
00000020  36 38 66 34                                       |68f4|
00000024

$ sudo rados -p data  listomapvals csi.volume.8acb7dbe-224b-11ea-99d3-92da74a268f4
csi.imagename
value (44 bytes) :
00000000  63 73 69 2d 76 6f 6c 2d  38 61 63 62 37 64 62 65  |csi-vol-8acb7dbe|
00000010  2d 32 32 34 62 2d 31 31  65 61 2d 39 39 64 33 2d  |-224b-11ea-99d3-|
00000020  39 32 64 61 37 34 61 32  36 38 66 34              |92da74a268f4|
0000002c

csi.volname
value (40 bytes) :
00000000  70 76 63 2d 31 35 37 66  39 61 36 63 2d 33 31 32  |pvc-157f9a6c-312|
00000010  33 2d 34 37 34 35 2d 61  30 61 35 2d 63 35 39 61  |3-4745-a0a5-c59a|
00000020  36 39 36 63 66 65 33 66                           |696cfe3f|
00000028
```

humblec:
This looks to be a nice addition. However let me seek feedback from others.. @dillaman

skoef:
I've added a function called `SyncImages` for a CSIJournal. This will go over all volumes in a dictionary *once* and checks if the csiImageKey is present. If not, it will look for an image which names ends with the same UUID. It then updates the key.

Currently, the mechanism is called when `CreateVolume` is called, but that is obviously not the best place. Perhaps @humblec has a suggestion where to hook into this? It is pool/cluster specific so I can only jump hook into it upon some sort of CSI request.

dillaman:
Why wouldn't the CSI just fallback to assuming the RBD image name is `csi.volume.` if the `csi.imagename` omap key is missing? How will this handle snapshots (currently prefixed w/ `csi.snap.`)?

skoef:
Because if we would allow custom prefixes (what this is all about), we couldn’t assume the image was created with `csi-vol-`. Snapshots are handled similarly, what is now prefixed by `csi-snap-` will then be stored in the same omap key.

dillaman:
> Because if we would allow custom prefixes (what this is all about), we couldn’t assume the image was created with `csi-vol-`. Snapshots are handled similarly, what is now prefixed by `csi-snap-` will then be stored in the same omap key.

For legacy images you most certainly can assume the lack of the key implies the image was created w/ the default prefix. If you are using a custom prefix, it implies you are using an updated CSI that will add the key. What am I missing?

skoef:
Ah right, yes that would be a fair assumption. I could wrap fetching the image name and default to the static prefix if the key is not found. That would make the syncing of image names obsolete!

skoef:
I think the PR is lot simpler now, thanks for the input @dillaman! I’ll throw in the nee flags for overriding the prefixes and I think it’s ready to be merged, if you all agree!

skoef:
in order to be even more flexible with prefix management, I've refactored some stuff so we're not referring to `CSIJournal.namingPrefix` from the outside and all name generation is now don in `CSIJournal.GetNameForUUID`.

skoef:
@dillaman @humblec I think I'm about done with this PR, what do you guys think?

skoef:
Any comments? I think this is a nice and clean approach solving the initial issue while being completely backwards compatible!

cmd/cephcsi.go
```diff
@@ -55,6 +55,9 @@ func init() {
 		" instances, when sharing Ceph clusters across CSI instances for provisioning")
 	flag.StringVar(&conf.MetadataStorage, "metadatastorage", "", "metadata persistence method [node|k8s_configmap]")
 	flag.StringVar(&conf.PluginPath, "pluginpath", "/var/lib/kubelet/plugins/", "the location of cephcsi plugin")
+	flag.StringVar(&conf.VolumeNamingPrefix, "volume-name-prefix", "", "override the default volume name prefix")
```
Madhu-1: the custom volume name prefix or snapshot name prefix should come from `storage-class` or `snapshot class`

humblec: Yeah, agree to @Madhu-1's comment.. So that its kind of a grouping which admin can apply to filter the volumes from SC .etc

skoef: I think I almost have this figured out, except for the 2 cases in `nodeserver.go` where I don't have enough context for this:
there is a code-path from `NodeStageVolume` and `NodeExpandVolume` (via `getVolumeName`) to `GetNameForUUID`. It makes sense to pass the configured name prefix through that path and this works in the case of ControllerServer's `CreateVolume` etc. These types of request contain the parameters that are accessible through `GetParameters()` but `NodeStageVolumeRequest` and `NodeExpandVolumeRequest` don't have these parameters accessible. 

How do you suggest this should be handled?

skoef: Perhaps, at that point, I can just get the volume name from the omap key. I’ll see if I can make that work!

skoef:
For now I have to revert f281eba4751a82ffd91d6a8459d6f76bdd6f227c, since we can't assume an volume name anymore and we don't have enough information (like monitors and pool) at that point to check this.

skoef:
Setting the `namePrefix` from the storageClass parameters works now! Proof:

First I created a PVC without setting the `namePrefix` parameters:
```
I0114 08:45:52.429248       1 utils.go:157] ID: 35 Req-ID: pvc-295dc502-eeb3-43ae-829f-8e65756863af GRPC call: /csi.v1.Controller/CreateVolume
I0114 08:45:52.432063       1 utils.go:158] ID: 35 Req-ID: pvc-295dc502-eeb3-43ae-829f-8e65756863af GRPC request: {"capacity_range":{"required_bytes":8589934592},"name":"pvc-295dc502-eeb3-43ae-829f-8e65756863af","parameters":{"clusterID":"a4334c89-41a5-4a27-9b87-e600421c7430","dataPool":"data","pool":"pool"},"secrets":"***stripped***","volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4","mount_flags":["discard"]}},"access_mode":{"mode":1}}]}
I0114 08:45:52.434510       1 rbd_util.go:492] ID: 35 Req-ID: pvc-295dc502-eeb3-43ae-829f-8e65756863af setting disableInUseChecks on rbd volume to: false
I0114 08:46:34.559050       1 rbd_journal.go:270] ID: 35 Req-ID: pvc-295dc502-eeb3-43ae-829f-8e65756863af generated Volume ID (0001-0024-a4334c89-41a5-4a27-9b87-e600421c7430-0000000000000009-4599fcb4-36aa-11ea-9c32-d63125295626) and image name (csi-vol-4599fcb4-36aa-11ea-9c32-d63125295626) for request name (pvc-295dc502-eeb3-43ae-829f-8e65756863af)
```
As you can see, the namePrefix is not part of the `parameters` passed to the csi-rbdplugin and the volume name is prefixed with `csi-vol-` as expected.

Now I adjust the storageClass by adding the namePrefix `foo-bar-` and create another PVC:
```
I0114 08:51:34.160914       1 utils.go:157] ID: 42 Req-ID: pvc-0a36c682-3a13-40dc-a4d5-594c3a77324b GRPC call: /csi.v1.Controller/CreateVolume
I0114 08:51:34.161002       1 utils.go:158] ID: 42 Req-ID: pvc-0a36c682-3a13-40dc-a4d5-594c3a77324b GRPC request: {"capacity_range":{"required_bytes":8589934592},"name":"pvc-0a36c682-3a13-40dc-a4d5-594c3a77324b","parameters":{"clusterID":"a4334c89-41a5-4a27-9b87-e600421c7430","dataPool":"data","namePrefix":"foo-bar-","pool":"pool"},"secrets":"***stripped***","volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4","mount_flags":["discard"]}},"access_mode":{"mode":1}}]}
I0114 08:51:34.164794       1 rbd_util.go:492] ID: 42 Req-ID: pvc-0a36c682-3a13-40dc-a4d5-594c3a77324b setting disableInUseChecks on rbd volume to: false
I0114 08:51:45.536843       1 rbd_journal.go:270] ID: 42 Req-ID: pvc-0a36c682-3a13-40dc-a4d5-594c3a77324b generated Volume ID (0001-0024-a4334c89-41a5-4a27-9b87-e600421c7430-0000000000000009-11284e6c-36ab-11ea-9c32-d63125295626) and image name (foo-bar-11284e6c-36ab-11ea-9c32-d63125295626) for request name (pvc-0a36c682-3a13-40dc-a4d5-594c3a77324b)
```
Now, the namePrefix parameter *is* set in the `parameters` and the resulting volume name is prefixed with `foo-bar-`. Omap values check out as well:

```
$ sudo rados -p pool  listomapvals csi.volume.11284e6c-36ab-11ea-9c32-d63125295626
csi.imagename
value (43 bytes) :
00000000  66 6f 6f 2d 62 61 72 31  31 32 38 34 65 36 63 2d  |foo-bar-11284e6c|
00000010  33 36 61 62 2d 31 31 65  61 2d 39 63 33 32 2d 64  |36ab-11ea-9c32-d|
00000020  36 33 31 32 35 32 39 35  36 32 36                 |63125295626|
0000002b

csi.volname
value (40 bytes) :
00000000  70 76 63 2d 30 61 33 36  63 36 38 32 2d 33 61 31  |pvc-0a36c682-3a1|
00000010  33 2d 34 30 64 63 2d 61  34 64 35 2d 35 39 34 63  |3-40dc-a4d5-594c|
00000020  33 61 37 37 33 32 34 62                           |3a77324b|
00000028
```

@Madhu-1 thanks for the excellent suggestion!

skoef:
Snapshots and Volumes obviously had different prefixes before, so I changed the parameter names to be able to specify both independently. Resulting output would be:
```
I0115 08:25:02.758592       1 utils.go:157] ID: 17 Req-ID: pvc-7457ca83-d775-466d-a838-d3a4c77051fd GRPC call: /csi.v1.Controller/CreateVolume
I0115 08:25:02.758618       1 utils.go:158] ID: 17 Req-ID: pvc-7457ca83-d775-466d-a838-d3a4c77051fd GRPC request: {"capacity_range":{"required_bytes":8589934592},"name":"pvc-7457ca83-d775-466d-a838-d3a4c77051fd","parameters":{"clusterID":"a4334c89-41a5-4a27-9b87-e600421c7430","dataPool":"-data","pool":"pool","volumeNamePrefix":"foo-bar-"},"secrets":"***stripped***","volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4","mount_flags":["discard"]}},"access_mode":{"mode":1}}]}
I0115 08:25:02.761962       1 rbd_util.go:491] ID: 17 Req-ID: pvc-7457ca83-d775-466d-a838-d3a4c77051fd setting disableInUseChecks on rbd volume to: false
I0115 08:25:04.140618       1 rbd_journal.go:270] ID: 17 Req-ID: pvc-7457ca83-d775-466d-a838-d3a4c77051fd generated Volume ID (0001-0024-a4334c89-41a5-4a27-9b87-e600421c7430-0000000000000009-8704b2d0-3770-11ea-b981-5e6715679e62) and image name (foo-bar-8704b2d0-3770-11ea-b981-5e6715679e62) for request name (pvc-7457ca83-d775-466d-a838-d3a4c77051fd)
```

@Madhu-1  @humblec @dillaman what do you guys think?

Madhu-1:
@ShyamsundarR @dillaman PTAL

nixpanic:
This looks reasonable and is a great feature to have!

You'll need to squash some commits together so that each commit becomes a single standalone functionality addition/improvement. A testcase under `e2e/` would be very good too.

Madhu-1:
@nixpanic pointed  out please squash commits

skoef:
I'll see if I can make an e2e testcase, but about the squashing: can't we just push *squash & merge* when we feel like merging? Wouldn't that have the same result?

Madhu-1:
> I'll see if I can make an e2e testcase, but about the squashing: can't we just push _squash & merge_ when we feel like merging? Wouldn't that have the same result?

we don't manually merge PR, Mergify bot will merge the PR upon receiving 2 approvals, it won't do squash and merge

nixpanic:
On Thu, Jan 16, 2020 at 06:31:10AM -0800, Madhu Rajanna wrote:
> > I'll see if I can make an e2e testcase, but about the squashing: can't we just push _squash & merge_ when we feel like merging? Wouldn't that have the same result?
> 
> we don't manually merge PR, Mergify bot will merge the PR upon receiving 2 approvals, it won't do squash and merge 

In addition to this, it is better to have descriptive commit messages
per functional change. Please use `git rebase` (keep a backup copy of
your changes in a 2nd branch) and see if you can make this PR a bit
cleaner with that. Let us know if you need assistance or hints for
rebasing + force pushing.

Thanks!

skoef:
@nixpanic I squashed it using `git rebase`, hope this is ok to merge now!

pkg/rbd/nodeserver.go
```diff
@@ -103,16 +97,29 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
-	volOptions.VolID = req.GetVolumeId()
+
+	var vi util.CSIIdentifier
+	err = vi.DecomposeCSIID(volID)
+	if err != nil {
+		err = fmt.Errorf("error decoding volume ID (%s) (%s)", err, volID)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	_, volOptions.RbdImageName, _, err = volJournal.GetObjectUUIDData(ctx, volOptions.Monitors, cr, volOptions.Pool, vi.ObjectUUID, false)
```
ShyamsundarR: For legacy volumes the image name was the PV name [for v1.0 version](https://github.com/ceph/ceph-csi/blob/2623bce76156f6de6cebb06b57ddfcd6b4c4e807/pkg/rbd/controllerserver.go#L119-L124).

The code here would fail to return the right legacy image name, as the volume name would be returned as "csi-vol-<UUID>", rather than the PV name that was passed initially.

As we do not store the PV name the function `getLegacyVolumeName` was the workaround to determine the image name from the mount path, as this carries the PV name in the path.

The code above needs to perform the fetch of image name based on the OMap if this is not a legacy volume, else fall back to the said function to get the same.

ShyamsundarR: ack!

skoef:
@ShyamsundarR I've done some cleaning up in `nodeserver.go` to make static-check pass. I think it's ready to be merged now:

- I squashed the commits again
- the commit message mentions RBD specifically
- the whole MR started off of an issue that was already there (#487). I can comment to it that work is done for RBD only?

skoef:
@nixpanic I noticed you put this issue on the board under **In Progress**. Since it has been somewhat quiet around this issue, does that mean you guys are still waiting for me to do something?

docs/deploy-rbd.md
```diff
@@ -50,6 +50,8 @@ make image-cephcsi
...
 | `pool`                                                                                              | yes                  | Ceph pool into which the RBD image shall be created                                                                                                                                                                     |
 | `dataPool`                                                                                          | no                   | Ceph pool used for the data of the RBD images.                                                                                                                                                                          |
+| `volumeNamePrefix`                                                                                  | no                   | Prefix to use for naming RBD volume images (defaults to `csi-vol-`).                                                                                                                                                    |
```
Madhu-1: this applies for cephfs also right not only for rbd? same question for snapshot also

pkg/cephfs/fsjournal.go
```diff
@@ -105,8 +112,10 @@ func undoVolReservation(ctx context.Context, volOptions *volumeOptions, vid volu
 // to generate the volume identifier for the reserved UUID
 func reserveVol(ctx context.Context, volOptions *volumeOptions, secret map[string]string) (*volumeIdentifier, error) {
 	var (
-		vi  util.CSIIdentifier
-		vid volumeIdentifier
+		vi        util.CSIIdentifier
+		vid       volumeIdentifier
+		imageUUID string
+		err       error
```
Madhu-1: let's not group different types of variables.

note: not related to this PR current code does this, we need to clean up that one as well.

pkg/rbd/nodeserver.go
```diff
@@ -66,12 +66,12 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	disableInUseChecks := false
 	// MULTI_NODE_MULTI_WRITER is supported by default for Block access type volumes
 	if req.VolumeCapability.AccessMode.Mode == csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER {
-		if isBlock {
-			disableInUseChecks = true
-		} else {
+		if !isBlock {
```
Madhu-1: nice catch, thanks for making this change

pkg/rbd/nodeserver.go
```diff
@@ -441,6 +481,20 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	return &csi.NodeUnpublishVolumeResponse{}, nil
 }
 
+// getStagingTargetPath concats NodeUnstageVolumeRequest's target path with
+// with the volumeID
+func getStagingTargetPath(req interface{}) string {
+	if nsvr, ok := req.(*csi.NodeStageVolumeRequest); ok {
```
Madhu-1: how about using switch case?

pkg/rbd/nodeserver.go
```diff
@@ -441,6 +481,20 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+		return nsvr.GetStagingTargetPath() + "/" + nsvr.GetVolumeId()
+	}
+
+	if nuvr, ok := req.(*csi.NodeUnstageVolumeRequest); ok {
+		return nuvr.GetStagingTargetPath() + "/" + nuvr.GetVolumeId()
+	}
+
+	return ""
```
Madhu-1: This may never occur, but still, we need to validate this empty string

pkg/rbd/nodeserver.go
```diff
@@ -441,6 +481,20 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	return &csi.NodeUnpublishVolumeResponse{}, nil
 }
 
+// getStagingTargetPath concats NodeUnstageVolumeRequest's target path with
+// with the volumeID
+func getStagingTargetPath(req interface{}) string {
```
Madhu-1: need to update function comment, its says NodeUnstageRequest, this function is doing for both stage and unstage

pkg/rbd/nodeserver.go
```diff
@@ -104,16 +99,36 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
 	volOptions, err := genVolFromVolumeOptions(ctx, req.GetVolumeContext(), req.GetSecrets(), disableInUseChecks, isLegacyVolume)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-	volOptions.RbdImageName = volName
-	volOptions.VolID = req.GetVolumeId()
+
+	// get rbd image name from the volume journal
+	// for legacy volumes (v1.0.0), the image name can be found in the staging path
+	if isLegacyVolume {
```
Madhu-1: can we move this to a separate function? which returns the rbd image name/volID and error?

pkg/rbd/nodeserver.go
```diff
@@ -557,14 +608,12 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
 		return &csi.NodeExpandVolumeResponse{}, nil
 	}
-	imgInfo, devicePath, err := getDevicePathAndImageInfo(ctx, volumePath)
+
+	devicePath, err := getDevicePath(ctx, volumePath)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if volName != imgInfo.ImageName {
```
Madhu-1: why this check is removed?

pkg/rbd/nodeserver_test.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2018 The Ceph-CSI Authors.
```
Madhu-1: 2020!

pkg/rbd/rbd_util.go
```diff
@@ -848,23 +856,16 @@ func resizeRBDImage(rbdVol *rbdVolume, newSize int64, cr *util.Credentials) erro
...
-func getVolumeName(volID string) (string, error) {
+func ensureEncryptionMetadataSet(ctx context.Context, cr *util.Credentials, rbdVol *rbdVolume) error {
 	var vi util.CSIIdentifier
 
-	err := vi.DecomposeCSIID(volID)
+	err := vi.DecomposeCSIID(rbdVol.VolID)
 	if err != nil {
-		err = fmt.Errorf("error decoding volume ID (%s) (%s)", err, volID)
-		return "", ErrInvalidVolID{err}
+		err = fmt.Errorf("error decoding volume ID (%s) (%s)", err, rbdVol.VolID)
```
Madhu-1: the error should be at last, as you touched this line can you please fix it here?

pkg/rbd/rbd_util_test.go
```diff
@@ -0,0 +1,56 @@
+/*
+Copyright 2018 The Ceph-CSI Authors.
```
Madhu-1: 2020

pkg/util/voljournal.go
```diff
@@ -109,46 +118,54 @@ type CSIJournal struct {
...
-// CSISnapshotSnapshot returns an instance of snapshot keys
+// NewCSISnapshotJournal returns an instance of CSIJournal for snapshots
 func NewCSISnapshotJournal() *CSIJournal {
 	return &CSIJournal{
 		csiDirectory:            "csi.snaps",
 		csiNameKeyPrefix:        "csi.snap.",
 		cephUUIDDirectoryPrefix: "csi.snap.",
 		csiNameKey:              "csi.snapname",
-		namingPrefix:            "csi-snap-",
+		csiImageKey:             "csi.imagename",
```
Madhu-1: csi.imagename or snapname?

pkg/rbd/nodeserver.go
```diff
@@ -539,10 +594,6 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	}
 	defer ns.VolumeLocks.Release(volumeID)
 
-	volName, err := getVolumeName(volumeID)
```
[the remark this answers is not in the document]
Madhu-1: as the we don't support ControllerExpandVolume for v1.0.0 images, we are good here

pkg/rbd/nodeserver.go
skoef: As explained in https://github.com/ceph/ceph-csi/pull/758#issuecomment-573666885, the imageName is fetched from the omap now and in this context we don't have enough information (like pool and monitors) to fetch that information.

pkg/rbd/nodeserver_test.go
skoef: The fact that I could've copied that entire header from any other files in this repo indicates that that deserves a PR on its own :)

pkg/util/voljournal.go
skoef: Doesn't really matter, it's an arbitrary name. I'm fine changing it to `csi.snapname`

skoef: While it is an arbitrary key name, I think it shouldn't conflict with either `csi.volname` or `csi.snapname` which are already the `csiNameKey`'s. This key specifically points to the Ceph image involved, so I think in both cases `csi.imagename` applies.

pkg/util/voljournal.go
```diff
@@ -327,75 +354,101 @@ func (cj *CSIJournal) ReserveName(ctx context.Context, monitors string, cr *Cred
...
 	requestName, err := GetOMapValue(ctx, monitors, cr, pool, cj.namespace,
 		cj.cephUUIDDirectoryPrefix+objectUUID, cj.csiNameKey)
 	if err != nil {
-		return "", "", err
+		return "", "", "", err
+	}
+
+	// image key was added at some point, so not all volumes will have this key set
+	// when ceph-csi was upgraded
+	imageName, err := GetOMapValue(ctx, monitors, cr, pool, cj.namespace,
```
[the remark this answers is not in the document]
skoef: I thought about implementing it right away, but I couldn't figure out how to handle the cases of `ErrKeyNotFound`s if you requested multiple keys at once.

pkg/rbd/nodeserver.go
skoef: The only reason I refactored this function at all was to reduce cyclomatic complexity. That was already quite high, so any changes of mine triggered it instantly. Apart from that, the changes don't really have something to do with this PR imho. While I applaud cleaning it up further, perhaps dedicated another PR to that.

But if I understand correctly, you'd like to see a function that returns rbdImageName and volID for a given request? I think I can do that.

skoef: I created  a switch/case for it now, any further improvements can be done outside of this PR imho

skoef:
I think I ticked all the boxes now and rebased it again. Imho ready to be merged!

PR #761, opened 2019-12-19
Add missing env for namespace.

sbskas:
Add a missing env for the rbdplugin to get the object from the correct namespace.
The env is correctly set in the csi-cephfsplugin.yaml but missing from the rbd.

### DIFF
--- a/deploy/rbd/kubernetes/v1.14+/csi-rbdplugin.yaml
+++ b/deploy/rbd/kubernetes/v1.14+/csi-rbdplugin.yaml
@@ -72,6 +72,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
             - name: CSI_ENDPOINT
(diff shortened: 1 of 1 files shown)


deploy/rbd/kubernetes/v1.14+/csi-rbdplugin.yaml
```diff
@@ -72,6 +72,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: POD_NAMESPACE
```
Madhu-1: @sbskas if I remember correctly this is required if  you are doing some operation with  PVC created with  1.0.0 ceph-csi

sbskas: I think this is required if you deploy in another namespace also.

Madhu-1: what you said is correct and as per https://github.com/ceph/ceph-csi/blob/46b19a93f5485ad5bb248fd638cfe8c8da220efa/pkg/util/k8scmcache.go#L54 required to support operate on PVC created using ceph-csi v1.0.0

PR #762, opened 2019-12-20
Adds per volume encryption with Vault integration

### LINKED ISSUE #420: RBD's PV claim can require the volume to be encrypted

### LINKED ISSUE #744: RBD Encryption improvements that need addressing
1) Ensure volume creation is complete before responding with a volume match for a `CreateVolume` request

Comment: https://github.com/ceph/ceph-csi/pull/697/files?file-filters%5B%5D=.go#r356896828

The intention is to note the status of creation and later, either complete the creation if certain steps are missed (like adding the encryption attribute to the image) or garbage collect the image and the reservation, and create afresh.

This provides a defense against any duplicate responses that the CO may have processed, to ensure the controller only responded when an entire `CreateVolume` call is complete.

2) `processEncryptedDevice` should check if device is already encrypted, prior to invoking `encryptDevice` and update encryption parameter on the image.

Comment: https://github.com/ceph/ceph-csi/pull/697/files?file-filters%5B%5D=.go#r355185194

This is to ensure if the plugin fails between calling `EncryptVolume` and `SaveRbdImageEncryptionStatus` in `encryptDevice`, we do not encrypt it again (which will not happen in the current code, as the device format will not be an empty string in `processEncryptedDevice`), but we need to update the `SaveRbdImageEncryptionStatus` on detecting such cases.

3) Optimize checks in `detachRBDImageOrDeviceSpec` to avoid checking for encrypted devices based on the image stash

Comment: https://github.com/ceph/ceph-csi/pull/697/files?file-filters%5B%5D=.go#r355185378

vasyl-purchel:
- adds proposal document for PVC encryption from PR448
- adds per-volume encription by generating encryption passphrase for each volume and storing it in a KMS
- adds HashiCorp Vault integration as a KMS for encryption passphrases (#420)
- avoids encrypting volume second time if it was already encrypted but no file system created (#744)
- avoids unnecessary checks if volume is a mapped device when encryption was not requested (#744)
- prevents resizing encrypted volumes (it is not currently supported)
- prevents creating snapshots from encrypted volumes to prevent attack on encryption key (security guard until re-encryption of volumes implemented)

Signed-off-by: Vasyl Purchel vasyl.purchel@workday.com
Signed-off-by: Andrea Baglioni andrea.baglioni@workday.com

Fixes: #420 
Fixes: #744

### DIFF
--- a/docs/deploy-rbd.md
+++ b/docs/deploy-rbd.md
@@ -1,4 +1,3 @@
-
 # CSI RBD Plugin
 
 The RBD CSI plugin is able to provision new RBD images and
@@ -55,6 +54,8 @@ make image-cephcsi
 | `csi.storage.k8s.io/provisioner-secret-namespace`, `csi.storage.k8s.io/node-stage-secret-namespace` | yes (for Kubernetes) | namespaces of the above Secret objects                                                                                                                                                                                  |
 | `mounter`                                                                                           | no                   | if set to `rbd-nbd`, use `rbd-nbd` on nodes that have `rbd-nbd` and `nbd` kernel modules to map rbd images                                                                                                              |
 | `encrypted`                                                                                         | no                   | disabled by default, use `"true"` to enable LUKS encryption on pvc and `"false"` to disable it. **Do not change for existing storageclasses**                                                                           |
+| `encryptionKMS`                                                                                     | no                   | specifies key management system for encrypytion. Currently supports `vault`                                                                                                                                             |
+| `encryptionKMSID`                                                                                   | no                   | required if `encryptionKMS` is set to `vault` to specify a unique identifier for vault configuration                                                                                                                    |
 
 **NOTE:** An accompanying CSI configuration file, needs to be provided to the
 running pods. Refer to [Creating CSI configuration](../examples/README.md#creating-csi-configuration)
@@ -196,8 +197,10 @@ and `csi.storage.k8s.io/node-stage-secret-namespace`).
 * volume is attached to provisioner container
 * on first time attachment
   (no file system on the attached device, checked with blkid)
+  * new passphrase is generated and stored in selected KMS if KMS is in use
   * device is encrypted with LUKS using a passphrase from K8s secrets
   * image-meta updated to "encrypted" in Ceph
+* passphrase is retrieved from selected KMS if KMS is in use
 * device is open and device path is changed to use a mapper device
 * mapper device is used instead of original one with usual workflow
 
@@ -205,6 +208,7 @@ and `csi.storage.k8s.io/node-stage-secret-namespace`).
 
 * mapper device closed and device path changed to original volume path
 * volume is detached as usual
+* passphrase removed from KMS if needed (with failures ignored)
 
 ### Encryption configuration
 
@@ -212,6 +216,38 @@ To encrypt rbd volumes with LUKS you need to set encryption passphrase in
 secrets under `encryptionPassphrase` key and switch `encrypted` option in
 StorageClass to `"true"`. This is not supported for storage classes that already
 have PVs provisioned.
+
+### Encryption KMS configuration
+
+To further improve security robustness it is possible to use unique passphrases
+generated for each volume and stored in a Key Management System (KMS). Currently
+HashiCorp Vault is the only KMS supported.
+
+To use Vault as KMS set `encryptionKMS` to `vault` and `encryptionKMSID` to a
+unique identifier for Vault configuration. You will also need to create vault
+configuration similar to the [example](../examples/rbd/kms-config.yaml)
+and use same `encryptionKMSID`. In order for ceph-csi to be able to access the
+configuration you will need to have it mounted to csi-rbdplugin containers in
+both daemonset (so kms client can be instantiated to encrypt/decrypt volumes)
+and deployment pods (so kms client can be instantiated to delete passphrase on
+volume delete) `ceph-csi-encryption-kms-config` config map.
+
+#### Configuring HashiCorp Vault
+
+Using Vault as KMS you need to configure Kubernetes authentication method as
+described in [official
+documentation](https://www.vaultproject.io/docs/auth/kubernetes.html).
+
+If token reviewer is used, you will need to configure service account for
+that also like in [example](../examples/rbd/csi-vaulttokenreview-rbac.yaml) to
+be able to review jwt tokens.
+
+Configure a role(s) for service accounts used for ceph-csi:
+
+* provisioner service account (`rbd-csi-provisioner`) requires only **delete**
+  permissions to delete passphrases on pvc delete
+* nodeplugin service account (`rbd-csi-nodeplugin`) requires **create** and
+  **read** permissions to save new keys and retrieve existing
 
 ### Encryption prerequisites
 
--- a/docs/design/proposals/encrypted-pvc.md
+++ b/docs/design/proposals/encrypted-pvc.md
@@ -1,0 +1,131 @@
+# Encrypted Persistent Volume Claims
+
+## Proposal
+
+Subject of this proposal is to add support for encryption of RBD volumes in
+Ceph-CSI.
+
+Some but not all the benefits of this approach:
+
+* guarantee encryption in transit to rbd without using messenger v2
+* extra security layer to application with special regulatory needs
+* at rest encryption can be disabled to selectively allow encryption only where
+  required
+
+## Document Terminology
+
+* volume encryption: encryption of a volume attached by rbd
+* encryption at rest: encryption of physical disk done by ceph
+* LUKS: Linux Unified Key Setup: stores all of the needed setup information for
+  dm-crypt on the disk
+* dm-crypt: linux kernel device-mapper crypto target
+* cryptsetup: the command line tool to interface with dm-crypt
+
+## Proposed Solution
+
+The proposed solution in this document, is to address the volume encryption
+requirement by using dm-crypt module through cryptsetup cli interface.
+
+### Implementation Summary
+
+* Encryption is implemented using cryptsetup with LUKS extension.
+  A good introduction to LUKS and dm-crypt in general can be found
+  [here](https://wiki.archlinux.org/index.php/Dm-crypt/Device_encryption#Encrypting_devices_with_cryptsetup)
+  Functions to implement necessary interaction are implemented in a separate
+  `cryptsetup.go` file.
+  * LuksFormat
+  * LuksOpen
+  * LuksClose
+  * LuksStatus
+
+* `CreateVolume`: refactored to prepare for encryption (tag image that it
+  requires encryption later), before returning, if encrypted volume option is
+  set.
+* `NodeStageVolume`: refactored to call `encryptDevice` method on the very first
+  volume attach request
+* `NodeStageVolume`: refactored to open encrypted device (`openEncryptedDevice`)
+* `openEncryptedDevice`: looks up for a passphrase matching the volume id,
+  returns the new device path in the form: `/dev/mapper/luks-<volume_id>`.
+  On the woker node where the attach is scheduled:
+
+  ```shell
+  $ lsblk
+  NAME                            MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINT
+  sda                               8:0    0   10G  0 disk
+  └─sda1                            8:1    0   10G  0 part  /
+  sdb                               8:16   0   20G  0 disk
+  rbd0                            253:0    0    1G  0 disk
+  └─luks-pvc-8a710f4c934811e9 252:0    0 1020M  0 crypt /var/lib/kubelet/pods/9eaceaef-936c-11e9-b396-005056af3de0/volumes/kubernetes.io~csi/pvc-8a710f4c934811e9/mount
+  ```
+
+* `detachRBDDevice`: calls `LuksClose` function to remove the LUKS mapping
+  before detaching the volume.
+
+* StorageClass extended with following parameters:
+  1. `encrypted` ("true" or "false")
+  1. `encryptionKMS` (string representing kms of choice)
+  ceph-csi plugin may support different kms vendors with different type of
+  authentication
+  1. `encryptionKMSID` (string representing kms configuration)
+
+* New KMS Configuration created.
+
+#### Annotated YAML for RBD StorageClass
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+   name: csi-rbd
+provisioner: rbd.csi.ceph.com
+parameters:
+   # String representing Ceph cluster configuration
+   clusterID: <cluster-id>
+   # ceph pool
+   pool: rbd
+
+   # RBD image features, CSI creates image with image-format 2
+   # CSI RBD currently supports only `layering` feature.
+   imageFeatures: layering
+
+   # The secrets have to contain Ceph credentials with required access
+   # to the 'pool'.
+   csi.storage.k8s.io/provisioner-secret-name: csi-rbd-secret
+   csi.storage.k8s.io/provisioner-secret-namespace: default
+   csi.storage.k8s.io/controller-expand-secret-name: csi-rbd-secret
+   csi.storage.k8s.io/controller-expand-secret-namespace: default
+   csi.storage.k8s.io/node-stage-secret-name: csi-rbd-secret
+   csi.storage.k8s.io/node-stage-secret-namespace: default
+   # Specify the filesystem type of the volume. If not specified,
+   # csi-provisioner will set default as `ext4`.
+   csi.storage.k8s.io/fstype: ext4
+
+   # Encrypt volumes
+   encrypted: "true"
+
+   # The type of kms we want to connect to: Barbican, aws kms or others can be
+   # supported
+   encryptionKMS: vault
+   # String representing a KMS configuration
+   encryptionKMSID: <kms-id>
+
+reclaimPolicy: Delete
+```
+
+And kms configuration:
+
+```yaml
+---
+apiVersion: v1
+kind: ConfigMap
+data:
+  config.json: |-
+    [
+      {
+        "kmsID": "<kms-id>",
+        kms specific config...
+      }
+    ]
+metadata:
+  name: ceph-csi-encryption-kms-config
+```
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -76,7 +76,7 @@ var _ = Describe("RBD", func() {
 		deployRBDPlugin()
 		createRBDStorageClass(f.ClientSet, f, make(map[string]string))
 		createRBDSecret(f.ClientSet, f)
-
+		deployVault(f.ClientSet, deployTimeout)
 	})
 
 	AfterEach(func() {
@@ -91,6 +91,7 @@ var _ = Describe("RBD", func() {
 		deleteResource(rbdExamplePath + "secret.yaml")
 		deleteResource(rbdExamplePath + "storageclass.yaml")
 		// deleteResource(rbdExamplePath + "snapshotclass.yaml")
+		deleteVault()
 	})
 
 	Context("Test RBD CSI", func() {
@@ -135,7 +136,20 @@ var _ = Describe("RBD", func() {
 			By("create a PVC and Bind it to an app with encrypted RBD volume", func() {
 				deleteResource(rbdExamplePath + "storageclass.yaml")
 				createRBDStorageClass(f.ClientSet, f, map[string]string{"encrypted": "true"})
-				validateEncryptedPVCAndAppBinding(pvcPath, appPath, f)
+				validateEncryptedPVCAndAppBinding(pvcPath, appPath, "", f)
+				deleteResource(rbdExamplePath + "storageclass.yaml")
+				createRBDStorageClass(f.ClientSet, f, make(map[string]string))
+			})
+
+			By("create a PVC and Bind it to an app with encrypted RBD volume with Vault KMS", func() {
+				deleteResource(rbdExamplePath + "storageclass.yaml")
+				scOpts := map[string]string{
+					"encrypted":       "true",
+					"encryptionKMS":   "vault",
+					"encryptionKMSID": "vault-test",
+				}
+				createRBDStorageClass(f.ClientSet, f, scOpts)
+				validateEncryptedPVCAndAppBinding(pvcPath, appPath, "vault", f)
 				deleteResource(rbdExamplePath + "storageclass.yaml")
 				createRBDStorageClass(f.ClientSet, f, make(map[string]string))
 			})
--- a/examples/rbd/storageclass.yaml
+++ b/examples/rbd/storageclass.yaml
@@ -43,6 +43,13 @@ parameters:
    # By default it is disabled. Valid values are “true” or “false”.
    # A string is expected here, i.e. “true”, not true.
    # encrypted: "true"
+
+   # Use external key management system for encryption passphrases
+   # encryptionKMS: vault
+
+   # String representing KMS configuration. Should be unique and match ID in
+   # KMS ConfigMap. The ID is only used for correlation to config map entry.
+   # encryptionKMSID: <kms-config-id>
 reclaimPolicy: Delete
 allowVolumeExpansion: true
 mountOptions:
--- a/pkg/rbd/controllerserver.go
+++ b/pkg/rbd/controllerserver.go
@@ -95,7 +95,7 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
 	}
 
 	// if it's NOT SINGLE_NODE_WRITER and it's BLOCK we'll set the parameter to ignore the in-use checks
-	rbdVol, err := genVolFromVolumeOptions(ctx, req.GetParameters(), nil, (isMultiNode && isBlock), false)
+	rbdVol, err := genVolFromVolumeOptions(ctx, req.GetParameters(), req.GetSecrets(), (isMultiNode && isBlock), false)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
@@ -343,7 +343,7 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	defer cs.VolumeLocks.Release(volumeID)
 
 	rbdVol := &rbdVolume{}
-	if err := genVolFromVolID(ctx, rbdVol, volumeID, cr); err != nil {
+	if err = genVolFromVolID(ctx, rbdVol, volumeID, cr, req.GetSecrets()); err != nil {
 		// If error is ErrInvalidVolID it could be a version 1.0.0 or lower volume, attempt
 		// to process it as such
 		if _, ok := err.(ErrInvalidVolID); ok {
@@ -377,7 +377,7 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		}
 		defer cs.VolumeLocks.Release(rbdVol.RequestName)
 
-		if err := undoVolReservation(ctx, rbdVol, cr); err != nil {
+		if err = undoVolReservation(ctx, rbdVol, cr); err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 		return &csi.DeleteVolumeResponse{}, nil
@@ -393,16 +393,22 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 
 	// Deleting rbd image
 	klog.V(4).Infof(util.Log(ctx, "deleting image %s"), rbdVol.RbdImageName)
-	if err := deleteImage(ctx, rbdVol, cr); err != nil {
+	if err = deleteImage(ctx, rbdVol, cr); err != nil {
 		klog.Errorf(util.Log(ctx, "failed to delete rbd image: %s/%s with error: %v"),
 			rbdVol.Pool, rbdVol.RbdImageName, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if err := undoVolReservation(ctx, rbdVol, cr); err != nil {
+	if err = undoVolReservation(ctx, rbdVol, cr); err != nil {
 		klog.Errorf(util.Log(ctx, "failed to remove reservation for volume (%s) with backing image (%s) (%s)"),
 			rbdVol.RequestName, rbdVol.RbdImageName, err)
 		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if rbdVol.Encrypted {
+		if err = rbdVol.KMS.DeletePassphrase(rbdVol.VolID); err != nil {
+			klog.V(3).Infof(util.Log(ctx, "failed to clean the passphrase for volume %s: %s"), rbdVol.VolID, err)
+		}
 	}
 
 	return &csi.DeleteVolumeResponse{}, nil
@@ -447,12 +453,18 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 
 	// Fetch source volume information
 	rbdVol := new(rbdVolume)
-	err = genVolFromVolID(ctx, rbdVol, req.GetSourceVolumeId(), cr)
+	err = genVolFromVolID(ctx, rbdVol, req.GetSourceVolumeId(), cr, req.GetSecrets())
 	if err != nil {
 		if _, ok := err.(ErrImageNotFound); ok {
 			return nil, status.Errorf(codes.NotFound, "source Volume ID %s not found", req.GetSourceVolumeId())
 		}
 		return nil, status.Errorf(codes.Internal, err.Error())
+	}
+
+	// TODO: re-encrypt snapshot with a new passphrase
+	if rbdVol.Encrypted {
+		return nil, status.Errorf(codes.Unimplemented, "source Volume %s is encrypted, "+
+			"snapshotting is not supported currently", rbdVol.VolID)
 	}
 
 	// Check if source volume was created with required image features for snaps
@@ -698,12 +710,17 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	defer cr.DeleteCredentials()
 
 	rbdVol := &rbdVolume{}
-	err = genVolFromVolID(ctx, rbdVol, volID, cr)
+	err = genVolFromVolID(ctx, rbdVol, volID, cr, req.GetSecrets())
 	if err != nil {
 		if _, ok := err.(ErrImageNotFound); ok {
 			return nil, status.Errorf(codes.NotFound, "volume ID %s not found", volID)
 		}
 		return nil, status.Errorf(codes.Internal, err.Error())
+	}
+
+	if rbdVol.Encrypted {
+		return nil, status.Errorf(codes.InvalidArgument, "encrypted volumes do not support resize (%s/%s)",
+			rbdVol.Pool, rbdVol.RbdImageName)
 	}
 
 	// always round up the request size in bytes to the nearest MiB/GiB
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -116,6 +116,7 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	volOptions.VolID = req.GetVolumeId()
 
 	isMounted := false
+	isEncrypted := false
 	isStagePathCreated := false
 	devicePath := ""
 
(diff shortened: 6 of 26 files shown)


Madhu-1:
@vasyl-purchel CI failure PTAL

deploy/rbd/kubernetes/v1.14+/csi-rbdplugin-provisioner.yaml
```diff
@@ -180,6 +182,9 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: ceph-csi-config
+        - name: ceph-csi-encryption-kms-config
```
Madhu-1: as this is configmap mounting to the plugin, I would like to discuss below the point


This is not directly related to this PR

* Do we need to support multiple CO  in ceph-csi (other than CO built on kubernetes)?
if not remove the configmap attaching as a volume in deployment and use kubernetes API to fetch the configmap
* Even updating  of configmap needed driver restart I believe
CC @humblec @ShyamsundarR @nixpanic @travisn

docs/design/proposals/encrypted-pvc.md
```diff
@@ -0,0 +1,131 @@
...
+  set.
+* `NodeStageVolume`: refactored to call `encryptDevice` method on the very first
+  volume attach request
+* `NodeStageVolume`: refactored to open encrypted device (`openEncryptedDevice`)
+* `openEncryptedDevice`: looks up for a passphrase matching the volume id,
+  returns the new device path in the form: `/dev/mapper/luks-<volume_id>`.
+  On the woker node where the attach is scheduled:
+
+  ```shell
+  [user@worker-01 ~]$ lsblk
```
Madhu-1: remove `user@worker-01`

e2e/rbd.go
```diff
@@ -141,6 +142,19 @@ var _ = Describe("RBD", func() {
 				createRBDStorageClass(f.ClientSet, f, make(map[string]string))
 			})
 
+			By("create a PVC and Bind it to an app with encrypted RBD volume with Vault KMS", func() {
```
Madhu-1: tests will fail as there are no deployment file changes for 1.13

examples/rbd/csi-vaulttokenreview-rbac.yaml
```diff
@@ -0,0 +1,28 @@
+---
+apiVersion: v1
```
Madhu-1: can we create a new folder called `kms/valut` and move these deployment files related to vault to it?

examples/rbd/storageclass.yaml
```diff
@@ -43,6 +43,13 @@ parameters:
    # By default it is disabled. Valid values are “true” or “false”.
    # A string is expected here, i.e. “true”, not true.
    # encrypted: "true"
+
+   # Use external key management system for encryption passphrases
+   # encryptionKMS: vault
+
+   # String representing KMS configuration. Should be unique and match ID in
+   # KMS ConfigMap.
```
Madhu-1: `Should be unique and match ID in KMS ConfigMap` what is ID here?

docs/deploy-rbd.md
```diff
@@ -205,6 +210,34 @@ secrets under `encryptionPassphrase` key and switch `encrypted` option in
...
+documentation](https://www.vaultproject.io/docs/auth/kubernetes.html).
+
+If token reviewer is used, you will need to configure service account for
+that also like in [example](../examples/rbd/csi-vaulttokenreview-rbac.yaml) to
+be able to review jwt tokens.
+
+Configure a role(s) for service accounts used for ceph-csi:
+
+* provisioner requires only delete permissions
+* attacher requires create and read permissions
```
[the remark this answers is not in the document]
Madhu-1: create should happen at CreateVolume right? if you want to support encryption per PVC?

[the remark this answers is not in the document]
vasyl-purchel: we do not encrypt volumes on creation but rather on first time attaching it (same way as we create filesystem there), so that's why we need "create" permission in vault for service account used in attachment process

examples/rbd/storageclass.yaml
vasyl-purchel: this ID is just a correlation to the KMS config object in configMap (implemented this just by following pattern how ceph connection is configured)

docs/deploy-rbd.md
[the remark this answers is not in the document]
Madhu-1: below are the 2 scenarios am trying to understand
 as the creation is happening in the NodeStageVolume request

1) if you are creating RWX PVC so you will be creating multiple keys if it is scheduled in multiple nodes
2) if the pod is moved from one node to another node again you will be creating a new key to encrypt it?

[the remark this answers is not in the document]
vasyl-purchel: it should not create a new key as we store it by volumeID
and because process is:
- try to get key
- if key not found:
  - generate new key
  - save it in kms
if it fails before it saved and get rescheduled - the new key will be generated, but once it's saved it should be reused

docs/design/proposals/encrypted-pvc.md
vasyl-purchel: done

e2e/rbd.go
vasyl-purchel: done

examples/rbd/csi-vaulttokenreview-rbac.yaml
vasyl-purchel: done

deploy/rbd/kubernetes/v1.14+/csi-rbdplugin-provisioner.yaml
```diff
@@ -141,6 +141,8 @@ spec:
               readOnly: true
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+            - name: ceph-csi-encryption-kms-config
+              mountPath: /etc/ceph-csi-encryption-kms-config/
```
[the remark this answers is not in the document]
vasyl-purchel: done

deploy/rbd/kubernetes/v1.14+/csi-rbdplugin-provisioner.yaml
```diff
@@ -141,6 +141,8 @@ spec:
               readOnly: true
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
+            - name: ceph-csi-encryption-kms-config
```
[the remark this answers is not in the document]
vasyl-purchel: done

deploy/rbd/kubernetes/v1.14+/csi-rbdplugin-provisioner.yaml
ShyamsundarR: > as this is configmap mounting to the plugin, I would like to discuss below the point
> 
> This is not directly related to this PR
> 
>     * Do we need to support multiple CO  in ceph-csi (other than CO built on kubernetes)?

We should keep the implementation as close to the ideal as possible, so yes to the above.

>       if not remove the configmap attaching as a volume in deployment and use kubernetes API to fetch the configmap
> 
>     * Even updating  of configmap needed driver restart I believe

No, [updates](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically) to the config map would be reflected within the pod in time.

>       CC @humblec @ShyamsundarR @nixpanic @travisn

pkg/rbd/nodeserver.go
```diff
@@ -565,6 +568,9 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if volName != imgInfo.ImageName {
 		return nil, status.Errorf(codes.InvalidArgument, "volume name missmatch between request (%s) and stored metadata (%s)", volName, imgInfo.ImageName)
 	}
+	if imgInfo.Encrypted {
```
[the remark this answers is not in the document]
vasyl-purchel: I don't believe we should allow resize for encrypted volumes at all since it's not supported atm. It might be supported later (https://unix.stackexchange.com/questions/320957/extend-a-luks-encrypted-partition-to-fill-disk) but it is not an easy task to implement hence we should prevent any attempts to do that

I thought this place will capture all the cases, are there any more or only ControllerExpandVolume misssed?

pkg/rbd/controllerserver.go
```diff
@@ -95,7 +95,7 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
 	}
 
 	// if it's NOT SINGLE_NODE_WRITER and it's BLOCK we'll set the parameter to ignore the in-use checks
-	rbdVol, err := genVolFromVolumeOptions(ctx, req.GetParameters(), nil, (isMultiNode && isBlock), false)
+	rbdVol, err := genVolFromVolumeOptions(ctx, req.GetParameters(), req.GetSecrets(), (isMultiNode && isBlock), false)
```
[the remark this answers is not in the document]
vasyl-purchel: I think we should init KMS - it's not much overhead on the process, but this way we'll verify that KMS is configured correctly (fail fast).

If we skip KMS initialisation for provisioner that means we'll create an image in ceph, but when we try to attach it - it will fail due to misconfiguration... I believe failing fast is better user experience.

pkg/rbd/nodeserver.go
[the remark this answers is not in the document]
ShyamsundarR: ControllerExpandVolume is the first call that expands the rbd image. Later, NodeExpand is called to expand an in-use volume. So if we return said errors in ControllerExpand we should be fine. (NodeExpand will never be called if we error out in ControllerExpand)

pkg/rbd/controllerserver.go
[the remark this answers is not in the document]
ShyamsundarR: fail-fast is fine. I would possibly on CSI init check if the KMS config is correct, rather than at each CreateVolume call though to fail fast. There is the outlier that an new KMS ID was added later, during runtime, but barring those we should catch config issues once during init if possible.

There are various scenarios where speed of volume creation is crucial, so optimizing those would be helpful.

pkg/rbd/controllerserver.go
```diff
@@ -391,18 +391,24 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
...
 
-	if err := undoVolReservation(ctx, rbdVol, cr); err != nil {
+	if err = undoVolReservation(ctx, rbdVol, cr); err != nil {
 		klog.Errorf(util.Log(ctx, "failed to remove reservation for volume (%s) with backing image (%s) (%s)"),
 			rbdVol.RequestName, rbdVol.RbdImageName, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if rbdVol.Encrypted {
+		if err = rbdVol.KMS.DeletePassphrase(rbdVol.VolID); err != nil {
```
ShyamsundarR: @Madhu-1 Note for PR https://github.com/ceph/ceph-csi/pull/693 : 
- Currently the base image can only be deleted once all the snapshots and clones are deleted, afaik
- Hence deleting this KMS key when the base image is deleted should be fine, as all dependent snaps and clones using this key would already be deleted
- Once the said PR is implemented though, we should be able to delete base image indenpendent of other relations (although this just means moving it to trash), so the key would also be deleted. This may need some handling, as we do not want to lose a key that has data in other related images encrypted.

(although in the final solution, we do not want snapshots and clones to share the same key, but till then we need the snapshot/clone story with encryption clear)

pkg/rbd/controllerserver.go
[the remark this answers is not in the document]
vasyl-purchel: I don't think it is possible to validate the KMS config on CSI init:

the KMS type is defined in storageclass, so we can't just assume that all entries from kms config are vault configurations (currently this is the only supported type but the interface is generic enough that other KMS would be easy to add - i.e. k8s secrets to store passphrase per volume in secrets).

In order to optimise scenario you've mentioned - we can introduce another flag `validateKMSOnCreate: true` that can be used to bypass the KMS validation and instantiation

pkg/rbd/controllerserver.go
vasyl-purchel: I think more secure approach would be to prevent creating snapshots from encrypted volumes. As mentioned before - creating many copies that use same encryption key opens it to attacks. Also I believe this will depend heavily on implementing re-encryption for volumes.

pkg/rbd/controllerserver.go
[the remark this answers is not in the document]
ShyamsundarR: > In order to optimise scenario you've mentioned - we can introduce another flag `validateKMSOnCreate: true` that can be used to bypass the KMS validation and instantiation

No new flags please :) (I am config options blow up averse) let's hold onto this for now, as KMS initialization happens only when encryption is turned on, so we can evaluate and later optimize if required.

pkg/rbd/controllerserver.go
ShyamsundarR: > I think more secure approach would be to prevent creating snapshots from encrypted volumes. As mentioned before - creating many copies that use same encryption key opens it to attacks. Also I believe this will depend heavily on implementing re-encryption for volumes.

Ack! so no intermediate solutions, we will turn off snapshots/clones for encrypted volumes, till we add offline re-encrypt support. (@Madhu-1 something for PR #693 I would assume)

pkg/rbd/controllerserver.go
```diff
@@ -391,18 +391,24 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
...
 	}
 
-	if err := undoVolReservation(ctx, rbdVol, cr); err != nil {
+	if err = undoVolReservation(ctx, rbdVol, cr); err != nil {
 		klog.Errorf(util.Log(ctx, "failed to remove reservation for volume (%s) with backing image (%s) (%s)"),
 			rbdVol.RequestName, rbdVol.RbdImageName, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if rbdVol.Encrypted {
```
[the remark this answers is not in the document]
vasyl-purchel: just a re-cap of offline discussion:
yes, delete passphrase does not work, so we need to figure out where to put data about what kms was used and if image was encrypted

* option one is to use extra meta-data field for kms type - it will allow to keep track of where passphrase is with the image itself, the downside is it will require extra request to ceph to get each meta-data field (so in proposed scenario - 1st call to check if encrypted + 2nd to find kms type)
* option two - extend omapval object with new fields - this will require a way to handle old version of the object and new... I have not yet investigated what this will involve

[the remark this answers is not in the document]
ShyamsundarR: @vasyl-purchel I suggest we go ahead with option two. (like, the image name extension in the PR https://github.com/ceph/ceph-csi/pull/758 ).
- One consideration for storing extra attributes in the image itself, was it extends well to mirrored rbd images, but for such cases we need to anyway extend mirroring the vault keys as well. So this does not seem to yield a large benefit yet.
- The OMap for the image has only 3 keys (if we include the image naming PR and snapshot source), so we are not yet stressing the OMap key limits IMO and can add some more (for example I am thinking of adding the pool [parameter](https://github.com/ceph/ceph-csi/pull/760#discussion_r368085409) for topology support as well).

So keeping various moving parts with the same design idea, seems a better option at present.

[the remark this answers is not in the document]
vasyl-purchel: Ack, will update review once it's implemented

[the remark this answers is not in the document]
vasyl-purchel: this is implemented - new key in rados object `csi.volume.encryptKMS` with `<kms-type>|<kms-id>`. If not found or empty - treated as not encrypted with KMS hence no need for passphrase clean up

pkg/rbd/rbd_util.go
```diff
@@ -775,6 +785,7 @@ func stashRBDImageMetadata(volOptions *rbdVolume, path string) error {
 		Version:   1, // Stash a v1 for now, in case of changes later, there are no checks for this at present
```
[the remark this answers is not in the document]
vasyl-purchel: done

pkg/rbd/nodeserver.go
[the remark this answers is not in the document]
vasyl-purchel: done, also tested manually and you were right - original check was allowing to expand rbd image but not file system getting stuck in the middle... now it fails out earlier in the process

pkg/rbd/controllerserver.go
vasyl-purchel: add guard for this [here](https://github.com/ceph/ceph-csi/pull/762/files#diff-35c77cc181d49675cd67a6095520b41fR465-R468)

docs/deploy-rbd.md
```diff
@@ -213,6 +218,36 @@ secrets under `encryptionPassphrase` key and switch `encrypted` option in
...
+described in [official
+documentation](https://www.vaultproject.io/docs/auth/kubernetes.html).
+
+If token reviewer is used, you will need to configure service account for
+that also like in [example](../examples/rbd/csi-vaulttokenreview-rbac.yaml) to
+be able to review jwt tokens.
+
+Configure a role(s) for service accounts used for ceph-csi:
+
+* provisioner service account (`rbd-csi-provisioner`) requires only **delete**
```
Madhu-1: why delete permission is given to provisioner here, in that case, both create and delete should be performed by provisioner or both create and delete should be performed by node plugin

@ShyamsundarR any comments here?

docs/deploy-rbd.md
```diff
@@ -213,6 +218,36 @@ secrets under `encryptionPassphrase` key and switch `encrypted` option in
...
+  permissions to delete passphrases on pvc delete
+* attacher service account (`rbd-csi-nodeplugin`) requires **create** and
```
Madhu-1: change `attacher service account` to `nodeplugin service account` attach will be done by external-attacher, not node-plugin

docs/design/proposals/encrypted-pvc.md
```diff
@@ -0,0 +1,131 @@
+# Encrypted Persistent Volume Claims
+
+## Proposal
+
+Subject of this proposal is to add support for encryption of RBD volumes in
+Ceph-CSI.
+
+Some but not all the benefits of this approach:
+
+* guarantee encryption in transit to rbd without using messenger v2
```
Madhu-1: can you please explain what is this?

docs/design/proposals/encrypted-pvc.md
```diff
@@ -0,0 +1,131 @@
...
+   # String representing a KMS configuration
+   encryptionKMSID: <kms-id>
+
+reclaimPolicy: Delete
+```
+
+And kms configuration:
+
+```yaml
+---
```
Madhu-1: can you please remove this line as it is not required

pkg/rbd/controllerserver.go
```diff
@@ -393,18 +393,24 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
...
+		if err = rbdVol.KMS.DeletePassphrase(rbdVol.VolID); err != nil {
+			klog.V(3).Infof(util.Log(ctx, "failed to clean the passphrase for volume %s: %s"), rbdVol.VolID, err)
```
Madhu-1: why the error is not handled here?

pkg/rbd/controllerserver.go
```diff
@@ -447,14 +453,20 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	err = genVolFromVolID(ctx, rbdVol, req.GetSourceVolumeId(), cr, req.GetSecrets())
 	if err != nil {
 		if _, ok := err.(ErrImageNotFound); ok {
 			return nil, status.Errorf(codes.NotFound, "source Volume ID %s not found", req.GetSourceVolumeId())
 		}
 		return nil, status.Errorf(codes.Internal, err.Error())
 	}
 
+	// TODO: re-encrypt snapshot with a new passphrase
+	if rbdVol.Encrypted {
```
Madhu-1: if not documented, we need to document this.

pkg/rbd/controllerserver.go
```diff
@@ -698,14 +710,19 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
-	err = genVolFromVolID(ctx, rbdVol, volID, cr)
+	err = genVolFromVolID(ctx, rbdVol, volID, cr, req.GetSecrets())
 	if err != nil {
 		if _, ok := err.(ErrImageNotFound); ok {
 			return nil, status.Errorf(codes.NotFound, "volume ID %s not found", volID)
 		}
 		return nil, status.Errorf(codes.Internal, err.Error())
 	}
 
+	if rbdVol.Encrypted {
```
Madhu-1: same here also, we need to document this as well.

pkg/util/vault.go
```diff
@@ -0,0 +1,347 @@
...
+func InitVaultKMS(opts, secrets map[string]string) (EncryptionKMS, error) {
+	var config []VaultKMS
+
+	vaultID, ok := opts["encryptionKMSID"]
+	if !ok {
+		return nil, fmt.Errorf("missing encryptionKMSID for vault as encryption KMS")
+	}
+
+	// #nosec
+	content, err := ioutil.ReadFile(kmsConfigPath)
```
Madhu-1: as we are reading from the same file always, we should read this during initialization and keep it in memory to make operation fast.

docs/deploy-rbd.md
Madhu-1: the question because am not sure you will get network access to  reach vault server from both provisioner and nodeplugin

pkg/util/crypto.go
```diff
@@ -36,24 +39,130 @@ const (
...
+	return nil, fmt.Errorf("unknown encryption KMS type %s", kmsType)
+}
+
+// GetKMSConfig returns required keys for KMS to instantiate from it's config
+// - map with kms type and ID keys
+// - error if format is invalid
+func GetKMSConfig(config string) (map[string]string, error) {
+	kmsConfigParts := strings.Split(config, "|")
+	if len(kmsConfigParts) != 2 {
+		return make(map[string]string), fmt.Errorf("failed to parse encryption KMS "+
```
Madhu-1: return nil,fmt.Errorf("") here?

pkg/util/vault.go
```diff
@@ -0,0 +1,347 @@
...
+			key, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == 404 {
+		return "", MissingPassphrase{fmt.Errorf("passphrase for %s not found", key)}
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		body, bodyParseErr := ioutil.ReadAll(resp.Body)
+		if bodyParseErr != nil {
```
Madhu-1: log `bodyParseErr` which helps for debugging

pkg/util/vault.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+
+	resp, err := kms.request("POST", kms.getKeyDataURI(key), data)
+	if err != nil {
+		return fmt.Errorf("failed to POST passphrase for %s to vault: %s", key, err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
```
Madhu-1: log err here

pkg/util/vault.go
Madhu-1: can  we rename `bodyParseErr` to `err`?

pkg/util/vault.go
```diff
@@ -0,0 +1,347 @@
...
+
+	resp, err := kms.send("DELETE", kms.getKeyMetadataURI(key), &vaultToken, nil)
+	if err != nil {
+		return fmt.Errorf("delete passphrase at %s request to vault failed: %s", key, err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode < 200 || (resp.StatusCode >= 300 && resp.StatusCode != 404) {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("failed to delete passphrase for %s with %v, <body parsing failed>",
```
Madhu-1: log err

pkg/util/vault.go
```diff
@@ -0,0 +1,347 @@
...
+
+/*
+getVaultAccessToken retrieves vault token using kubernetes authentication:
+ 1. read jwt service account token from well known location
+ 2. request token from vault using service account jwt token
+Vault will verify service account jwt token with Kubernetes and return token
+if the requester is allowed
+*/
+func (kms *VaultKMS) getAccessToken() (string, error) {
+	saToken, err := ioutil.ReadFile(serviceAccountTokenPath)
```
Madhu-1: can we read and store it in memory instead of reading all-time to speed up operation?

pkg/util/vault.go
```diff
@@ -0,0 +1,347 @@
...
+	if err != nil {
+		return "", fmt.Errorf("vault token request data is broken: %s", err)
+	}
+	resp, err := kms.send("POST", kms.VaultAuthPath, nil, data)
+	if err != nil {
+		return "", fmt.Errorf("failed to retrieve vault token: %s", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
Madhu-1: can we move this code to a common function? this  code is used in multiple places

pkg/util/vault.go
```diff
@@ -0,0 +1,347 @@
...
+		body, bodyParseErr := ioutil.ReadAll(resp.Body)
+		if bodyParseErr != nil {
+			return "", fmt.Errorf("failed to retrieve vault token (%v), <body parsing failed>",
```
Madhu-1: log error

docs/deploy-rbd.md
vasyl-purchel: so regarding this:
- volumes are deleted by provisioner, hence delete permissions in vault is required here
- since we're not encrypting volume on create but rather on first attach - we'll need to have create permissions and also read since each attach call will need to get passphrase

considering that in vault you can have quite flexible permissions and have different roles for different actions - it's easy to restrict only to operations that are required

pkg/util/vault.go
vasyl-purchel: we init kms only once for a request... so even if we make few requests to vault - we do not reload this file

pkg/util/vault.go
vasyl-purchel: can't rename as then static check complains that we're shadowing `err` from line 151

docs/design/proposals/encrypted-pvc.md
vasyl-purchel: so when you're using ceph to create volumes, you can have osd encryption that will ensure data is encrypted before it's stored on the disk, and with messenger v2 you can get encryption in transit to make sure data is encrypted while it's passed over the network (look [here](https://docs.ceph.com/docs/master/rados/configuration/msgr2/))

but encrypting network traffic is not required if you have encrypted data stored on rbd image itself, so it doesn't matter if you use msg v2 or not but you have guarantee that data is safe while it's transfered

pkg/util/vault.go
vasyl-purchel: so taking into account following:
* in most cases only single operation will be used with kms, so it will be read once
* when we create volume we verify kms configuration, so we do instantiate kms class and if we read the file at that point - it will slow down creates
* the only time when we will read this file multiple times is when we first time attach encrypted volume since we will need to (1) save the passphrase and then (2) read it from vault to ensure we're using correct passphrase - and this can be improved in the future. (with re-encryption we'll need to improve it, but not for now)

I'd prefer to leave it as is and improve it if it impacts performance a lot or when re-encryption is added.

charts/ceph-csi-cephfs/templates/provisioner-deployment.yaml
```diff
@@ -178,6 +180,9 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: {{ .Values.configMapName | quote }}
+        - name: ceph-csi-encryption-kms-config
```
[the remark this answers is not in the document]
vasyl-purchel: removed

docs/deploy-rbd.md
```diff
@@ -213,6 +218,36 @@ secrets under `encryptionPassphrase` key and switch `encrypted` option in
...
+
+To further improve security robustness it is possible to use unique passphrases
+generated for each volume and stored in a Key Management System (KMS). Currently
+HashiCorp Vault is the only KMS supported.
+
+To use Vault as KMS set `encryptionKMS` to `vault` and `encryptionKMSID` to a
+unique identifier for Vault configuration. You will also need to create vault
+configuration similar to the [example](../examples/rbd/kms-config.yaml)
+and use same `encryptionKMSID`. In order for ceph-csi to be able to access the
+configuration you will need to have it mounted to csi-rbdplugin containers as
```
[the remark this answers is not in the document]
vasyl-purchel: done

docs/deploy-rbd.md
vasyl-purchel: done

pkg/util/vault.go
vasyl-purchel: moved into common function and added error into the final error output

pkg/util/vault.go
vasyl-purchel: moved into common function and added error into the final error output

pkg/util/vault.go
vasyl-purchel: done

pkg/util/vault.go
vasyl-purchel: done

docs/deploy-rbd.md
ShyamsundarR: I am fine with the reasoning (IOW, Create is by the nodeplugin, and Delete is by the Provisioner).


@Madhu-1 what is the case where vault access from nodeplugin would be prevented? As the provisioner can (like the nodeplugin) run on any node, and so if there are node specific access restrictions to vault, then even the provisioner can face the same, when running on any such worker node, afaict.

ShyamsundarR:
Added DNM for any inadvertent tool based auto-merge.

dillaman:
> As I understand expanding on the number of keys is not a scalable idea, rather a single key:value is a better construct to use with RADOS. @dillaman thoughts?

I have to be honest, I haven't been keeping up w/ this since I've been heads-down on the new RBD mirroring for Octopus. If you are saying that it is now storing even more image properties in the CSI omap directory, I really don't like that idea since it's not scalable. I'll try to take an in-depth look early next week.

vasyl-purchel:
The reason I've decided to go with a single key containing data in format `<kms-type>|<kms-id>`:
1. it represents all required data to re-instantiate kms client (same as set in storage class)
2. json format will take more space for no reason - we do not need flexibility, all flexibility comes with this data being only a reference to config map that contains all the necessary data

kms endpoint, creds, etc. should not be stored with the image - because then if you move vault instance to a new endpoint you'll have to go and update that in all previously created images.
also versioning is easy to control too - instead of "update kms config" to new version - you create a new kms config with it's own id and use new id in storage class for new images.

ShyamsundarR:
> The reason I've decided to go with a single key containing data in format `<kms-type>|<kms-id>`:
> 
>     1. it represents all required data to re-instantiate kms client (same as set in storage class)
> 
>     2. json format will take more space for no reason - we do not need flexibility, all flexibility comes with this data being only a reference to config map that contains all the necessary data
> 
> 
> kms endpoint, creds, etc. should not be stored with the image - because then if you move vault instance to a new endpoint you'll have to go and update that in all previously created images.
> also versioning is easy to control too - instead of "update kms config" to new version - you create a new kms config with it's own id and use new id in storage class for new images.

@vasyl-purchel The key:value comment is not just for this KMS key, it is for all keys that are being added, and instead serialize it to JSON. This is to control key proliferation.

If we decide to use a single key for "all" such extensions, then the value that is stored can be a JSON blob is my opinion. If we come to this decision, one of us (liberally volunteering others) will take care of ensuring this is done before 2.1.0.

vasyl-purchel:
ok, I agree that if the whole structure is migrated then json make sense (or any other serialisation)

dillaman:
I reviewed this and my only concern is still the added omap key on the CSI directory object. I really don't want to see a proliferation of extraneous keys on the CSI directory. Why can't this state be stored in the image-meta?

ShyamsundarR:
> I reviewed this and my only concern is still the added omap key on the CSI directory object. I really don't want to see a proliferation of extraneous keys on the CSI directory. Why can't this state be stored in the image-meta?

@dillaman I wanted to expand on this conversation. There are 2 other PRs in flight that add keys to the CSI directory, in both cases the keys help identify the [image name](https://github.com/ceph/ceph-csi/pull/758/files?file-filters%5B%5D=.go#diff-dbad340382ae55fa4ae6cc5429ca80dbR386-R389) or [which pool](https://github.com/ceph/ceph-csi/pull/816/files#diff-dbad340382ae55fa4ae6cc5429ca80dbR387-R393) the image is located in.

As a result storing the above on the image as an image-meta is not feasible.

Taking into consideration this PR and the other 2 in flight PRs, my thought for controlling proliferation of CSI directory keys is as stated in point (2) [here](https://github.com/ceph/ceph-csi/pull/762#pullrequestreview-351585306).

Leaving this comment here, to discuss possibilities if we decide against image meta for the KMS key and retain it as an CSI directory key instead.

dillaman:
> I reviewed this and my only concern is still the added omap key on the CSI directory object. I really don't want to see a proliferation of extraneous keys on the CSI directory. Why can't this state be stored in the image-meta?

Disregard -- I didn't realize there was a per-image CSI object and instead I thought this was going into the potentially large CSI directory object. Given that, I would say use individual key/value pairs and not try to create one giant aggregated key/value to contain all possible features of the CSI.

ShyamsundarR:
> > I reviewed this and my only concern is still the added omap key on the CSI directory object. I really don't want to see a proliferation of extraneous keys on the CSI directory. Why can't this state be stored in the image-meta?
> 
> Disregard -- I didn't realize there was a per-image CSI object and instead I thought this was going into the potentially large CSI directory object. Given that, I would say use individual key/value pairs and not try to create one giant aggregated key/value to contain all possible features of the CSI.

Thanks @dillaman, noting down my understanding (based on the conversation with you) the scale factors are close to 100k-200k keys per object, and image object has a very low key count (even considering the new PRs) and does not require key reduction optimization.

ShyamsundarR:
> I have the following comments, which need to be closed (addressed or decided to leave as is) by the 2.1.0 release (IOW, the next release). This is to ensure we do not get into any backward compatibility issue once we make a release with the feature.
> 
>     1. Can we reduce 2 keys to one, and encode "encryptionKMS" value in the JSON for the corresponding "encryptionKMSID". As it stands the KMS JSON config can in the future carry different KMS solutions, so each section in there possibly will call out the KMS type, for the code to function right. Hence 2 keys in the StorageClass seems superfluous at present.

@Madhu-1 @vasyl-purchel thoughts on the above reduction in StorageClass parameters?

vasyl-purchel:
@ShyamsundarR , it can be done I believe but this will complicate the parsing:
right now we know exactly what kind of configuration we're trying to parse so we can call specific type parsing method... if we will have only the id - we'll need to walk through configuration till we find required id and get type from there... and then we won't be able to parse it into a type but do that manually...

ShyamsundarR:
> @ShyamsundarR , it can be done I believe but this will complicate the parsing:
> right now we know exactly what kind of configuration we're trying to parse so we can call specific type parsing method... if we will have only the id - we'll need to walk through configuration till we find required id and get type from there... and then we won't be able to parse it into a type but do that manually...

I agree that parsing or auto detection is more complex when we do not know the type, but from a usability point helps the user to not have to reflect that information in 2 places.

So if it is a question of code complexity, I would prefer we attempt reducing this to a single value, possibly post this PR is merged and before 2.1.0 release.

vasyl-purchel:
sure, I'm ok with changing it but will prefer to merge the PR and change the config in follow up PR since it's already quite complex and often gets hard to rebase without breaking things

ShyamsundarR:
@Madhu-1 if your concerns are addressed, we can move this along as we need one more approval to get this in. Removing the DNM added to discuss the points raised.

PR #777, opened 2020-01-07
Implement CreateVolume with go-ceph API

nixpanic:

This is the initial step for improving performance during provisioning
of CSI volumes backed by RBD.

While creating a volume, an existing connection to the Ceph cluster is
used from the ConnPool. This should speed up the creation of a batch of
volumes significantly.

go-ceph depends on librbd and other Ceph shared libraries. The development headers need to be available in order to be able to build ceph-csi. This might need a reference in the developers documentation.

## Related issues ##

See-also: #729 #449 

## Future concerns ##

Deletion of volumes, and anything related to snapshots is something that will need to be done at a later time. go-ceph does not offer all APIs yet, and needs to get extended first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,19 @@ before_install:
   - curl
     https://raw.githubusercontent.com/golang/dep/${DEP_VERSION}/install.sh|sh
 
+before_script:
+  - curl https://download.ceph.com/keys/release.asc | sudo apt-key add -
+  - sudo apt-add-repository
+    "deb https://download.ceph.com/debian-nautilus xenial main"
+  # Ceph does not guarantee arm64 builds, fallback to ceph 14.2.5
+  # yamllint disable rule:line-length
+  - test "$(arch)" != "aarch64" || sudo apt-add-repository
+    "deb https://chacra.ceph.com/r/ceph/nautilus/af06652dc9b2da8c6aadbbecdfafdc7e235abe7d/ubuntu/xenial/flavors/default/ xenial main"
+  # yamllint enable rule:line-length
+  - sudo apt-get -qq update
+  # only the arm64 fallback repo is unsigned and needs --allow-unauthenticated
+  - sudo apt-get -y --allow-unauthenticated install librados-dev librbd-dev
+
 jobs:
   include:
     - name: static-check-make
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -75,6 +75,10 @@
   branch = "master"
   name = "k8s.io/utils"
 
+[[constraint]]
+  name = "github.com/ceph/go-ceph"
+  version = "0.2"
+
 [prune]
   go-tests = true
   non-go = true
@@ -84,4 +88,3 @@
 [[prune.project]]
   name = "github.com/kubernetes-csi/external-snapshotter"
   non-go = false
-
--- a/pkg/rbd/controllerserver.go
+++ b/pkg/rbd/controllerserver.go
@@ -125,6 +125,8 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, err
 	}
 
+	// TODO: create/get a connection from the the ConnPool, and do not pass
+	// the credentials to any of the utility functions.
 	cr, err := util.NewUserCredentials(req.GetSecrets())
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
@@ -135,6 +137,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	if err != nil {
 		return nil, err
 	}
+	defer rbdVol.Destroy()
 
 	// Existence and conflict checks
 	if acquired := cs.VolumeLocks.TryAcquire(req.GetName()); !acquired {
@@ -294,6 +297,7 @@ func (cs *ControllerServer) DeleteLegacyVolume(ctx context.Context, req *csi.Del
 	defer cs.VolumeLocks.Release(volumeID)
 
 	rbdVol := &rbdVolume{}
+	defer rbdVol.Destroy()
 	if err := cs.MetadataStore.Get(volumeID, rbdVol); err != nil {
 		if err, ok := err.(*util.CacheEntryNotFound); ok {
 			klog.V(3).Infof(util.Log(ctx, "metadata for legacy volume %s not found, assuming the volume to be already deleted (%v)"), volumeID, err)
@@ -352,6 +356,7 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	defer cs.VolumeLocks.Release(volumeID)
 
 	rbdVol := &rbdVolume{}
+	defer rbdVol.Destroy()
 	if err = genVolFromVolID(ctx, rbdVol, volumeID, cr, req.GetSecrets()); err != nil {
 		if _, ok := err.(util.ErrPoolNotFound); ok {
 			klog.Warningf(util.Log(ctx, "failed to get backend volume for %s: %v"), volumeID, err)
@@ -742,6 +747,7 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	defer cr.DeleteCredentials()
 
 	rbdVol := &rbdVolume{}
+	defer rbdVol.Destroy()
 	err = genVolFromVolID(ctx, rbdVol, volID, cr, req.GetSecrets())
 	if err != nil {
 		if _, ok := err.(ErrImageNotFound); ok {
--- a/pkg/rbd/rbd_util.go
+++ b/pkg/rbd/rbd_util.go
@@ -42,7 +42,6 @@ import (
 
 const (
 	imageWatcherStr = "watcher="
-	rbdImageFormat2 = 2
 	// The following three values are used for 30 seconds timeout
 	// while waiting for RBD Watcher to expire.
 	rbdImageWatcherInitDelay = 1 * time.Second
@@ -77,7 +76,6 @@ type rbdVolume struct {
 	Monitors           string `json:"monitors"`
 	Pool               string `json:"pool"`
 	DataPool           string
-	ImageFormat        string `json:"imageFormat"`
 	ImageFeatures      string `json:"imageFeatures"`
 	AdminID            string `json:"adminId"`
 	UserID             string `json:"userId"`
@@ -131,30 +129,16 @@ func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 	volSzMiB := fmt.Sprintf("%dM", util.RoundOffVolSize(pOpts.VolSize))
 	options := librbd.NewRbdImageOptions()
 
-	var err error
-	imageFormat := rbdImageFormat2
-	if pOpts.ImageFormat != "" {
-		imageFormat, err = strconv.Atoi(pOpts.ImageFormat)
-		if err != nil {
-			return errors.Wrapf(err, "failed to convert ImageFormat (%v) to integer", pOpts.ImageFormat)
-		}
-	}
-
-	err = options.SetUint64(librbd.RbdImageOptionFormat, uint64(imageFormat))
-	if err != nil {
-		return errors.Wrapf(err, "failed to set ImageFormat to %v", imageFormat)
-	}
-
-	logMsg := "rbd: create %s size %s format %d (features: %s) using mon %s, pool %s "
+	logMsg := "rbd: create %s size %s (features: %s) using mon %s, pool %s "
 	if pOpts.DataPool != "" {
 		logMsg += fmt.Sprintf("data pool %s", pOpts.DataPool)
-		err = options.SetString(librbd.RbdImageOptionDataPool, pOpts.DataPool)
+		err := options.SetString(librbd.RbdImageOptionDataPool, pOpts.DataPool)
 		if err != nil {
 			return errors.Wrapf(err, "failed to set data pool")
 		}
 	}
 	klog.V(4).Infof(util.Log(ctx, logMsg),
-		pOpts.RbdImageName, volSzMiB, imageFormat, pOpts.ImageFeatures, pOpts.Monitors, pOpts.Pool)
+		pOpts.RbdImageName, volSzMiB, pOpts.ImageFeatures, pOpts.Monitors, pOpts.Pool)
 
 	if pOpts.ImageFeatures != "" {
 		features := imageFeaturesToUint64(ctx, pOpts.ImageFeatures)
@@ -327,12 +311,6 @@ func updateVolWithImageInfo(ctx context.Context, rbdVol *rbdVolume, cr *util.Cre
 	if err != nil {
 		return err
 	}
-
-	if imageInfo.Format != 2 {
-		return fmt.Errorf("unknown or unsupported image format (%d) returned for image (%s)",
-			imageInfo.Format, rbdVol.RbdImageName)
-	}
-	rbdVol.ImageFormat = strconv.Itoa(rbdImageFormat2)
 
 	rbdVol.VolSize = imageInfo.Size
 	rbdVol.ImageFeatures = strings.Join(imageInfo.Features, ",")
@@ -546,7 +524,7 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
 	}
 
 	// if no image features is provided, it results in empty string
-	// which disable all RBD image format 2 features as we expected
+	// which disable all RBD image features as we expected
 
 	imageFeatures, found := volOptions["imageFeatures"]
 	if found {
@@ -765,7 +743,6 @@ func getSnapshotMetadata(ctx context.Context, pSnapOpts *rbdSnapshot, cr *util.C
 type imageInfo struct {
 	ObjectUUID string   `json:"name"`
 	Size       int64    `json:"size"`
-	Format     int64    `json:"format"`
 	Features   []string `json:"features"`
 	CreatedAt  string   `json:"create_timestamp"`
 }
--- a/pkg/rbd/rbd_util.go
+++ b/pkg/rbd/rbd_util.go
@@ -30,6 +30,8 @@ import (
 
 	"github.com/ceph/ceph-csi/pkg/util"
 
+	"github.com/ceph/go-ceph/rados"
+	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/pborman/uuid"
@@ -40,7 +42,7 @@ import (
 
 const (
 	imageWatcherStr = "watcher="
-	rbdImageFormat2 = "2"
+	rbdImageFormat2 = 2
 	// The following three values are used for 30 seconds timeout
 	// while waiting for RBD Watcher to expire.
 	rbdImageWatcherInitDelay = 1 * time.Second
@@ -88,6 +90,9 @@ type rbdVolume struct {
 	DisableInUseChecks bool   `json:"disableInUseChecks"`
 	Encrypted          bool
 	KMS                util.EncryptionKMS
+
+	// connection
+	conn *rados.Conn
 }
 
 // rbdSnapshot represents a CSI snapshot and its RBD snapshot specifics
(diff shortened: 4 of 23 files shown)


humblec:
Thanks @nixpanic ! 
@Madhu-1 @ShyamsundarR @dillaman PTAL.

Gopkg.toml
```diff
@@ -76,3 +76,7 @@
 [[prune.project]]
   name = "github.com/kubernetes-csi/external-snapshotter"
   non-go = false
+
+[[constraint]]
```
humblec: @nixpanic not sure whats the plan in go ceph project about the releases. It would be really better if we can stick onto the release versions than the `master`.

Madhu-1: yes, it's better to use the release version

pkg/rbd/rbd_util.go
```diff
@@ -109,36 +116,98 @@ type rbdSnapshot struct {
...
 
 // createImage creates a new ceph image with provision and volume options.
 func createImage(ctx context.Context, pOpts *rbdVolume, volSz int64, cr *util.Credentials) error {
-	var output []byte
-
-	image := pOpts.RbdImageName
 	volSzMiB := fmt.Sprintf("%dM", volSz)
+	options := librbd.NewRbdImageOptions()
+
+	imageFormat, err := strconv.Atoi(rbdImageFormat2)
```
Madhu-1: can we avoid this conversion and define `rbdImageFormat2` as int?

pkg/rbd/rbd_util.go
```diff
@@ -109,36 +116,98 @@ type rbdSnapshot struct {
...
+	if err != nil {
+		return errors.Wrapf(err, "[BUG] rbdImageFormat2 should be convertable to integer")
+	}
+	if pOpts.ImageFormat != "" {
+		imageFormat, err = strconv.Atoi(pOpts.ImageFormat)
```
Madhu-1: this should be done at input validation isn't it? means before reserving rados keys

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,194 @@
+/*
+Copyright 2020 The Ceph-CSI Authors.
```
Madhu-1: @nixpanic can we open a connect in https://github.com/ceph/ceph-csi/blob/72e37675945d67d3f88943eedb3fed4d146db42d/pkg/rbd/controllerserver.go#L125 and delete it in https://github.com/ceph/ceph-csi/blob/72e37675945d67d3f88943eedb3fed4d146db42d/pkg/rbd/controllerserver.go#L129

Gopkg.toml
nixpanic: There is no timeline for a release yet. Some APIs might still change (unlikely for functions that are used now).

pkg/rbd/rbd_util.go
nixpanic: Yeah, I'll probably do this. Will need to see if it is used in other places.

pkg/rbd/rbd_util.go
nixpanic: I am not sure what you mean with this. Can you explain a little more?

pkg/util/conn_pool.go
nixpanic: We could, and that is probably a good idea. However there are still a few functions that need the credentials so that the `rbd` commandline can be built. Not all places can use the go-ceph API yet.

I'll mark it as a TODO for now, and will work towards removing the passing of the credentials to functions in `rbd_utils.go`.

Gopkg.toml
humblec: If the apis which we are consuming here are respected in future releases, its fine to stick on to `master` for now.

pkg/util/conn_pool.go
nixpanic: A TODO comment has been added now. @Madhu-1 are you ok with that?

Madhu-1: @nixpanic ack

pkg/rbd/rbd_util.go
dillaman: Image format 1 is not supported by librbd anymore, so it will always be image format 2.

nixpanic: @Madhu-1 @humblec are you fine with removing ImageFormat references? I do not see it configurable or used elsewhere. If there is any historic need, please let me know.

nixpanic: References to ImageFormat have been removed now with an extra commit.

.travis.yml
```diff
@@ -36,6 +36,19 @@ before_install:
...
     https://raw.githubusercontent.com/golang/dep/${DEP_VERSION}/install.sh|sh
 
+before_script:
+  - curl https://download.ceph.com/keys/release.asc | sudo apt-key add -
+  - sudo apt-add-repository
+    "deb https://download.ceph.com/debian-nautilus xenial main"
+  # Ceph does not guarantee arm64 builds, fallback to ceph 14.2.5
+  # yamllint disable rule:line-length
+  - test "$(arch)" != "aarch64" || sudo apt-add-repository
+    "deb https://chacra.ceph.com/r/ceph/nautilus/af06652dc9b2da8c6aadbbecdfafdc7e235abe7d/ubuntu/xenial/flavors/default/ xenial main"
```
Madhu-1: going forward is this going to be an issue from building with nautilus?

pkg/rbd/rbd_util.go
```diff
@@ -74,7 +75,6 @@ type rbdVolume struct {
 	Monitors           string `json:"monitors"`
 	Pool               string `json:"pool"`
 	DataPool           string
-	ImageFormat        string `json:"imageFormat"`
```
Madhu-1: I think this is needed for volume created by CSI v1.0.0 driver.

pkg/rbd/rbd_util.go
```diff
@@ -109,36 +114,80 @@ type rbdSnapshot struct {
...
+	ioctx, err := pOpts.getIoctx(cr)
+	if err != nil {
+		return err
 	}
-	output, err := execCommand("rbd", args)
+	defer ioctx.Destroy()
 
+	// image is not returned to the caller
+	// no need to Close(), the image is not open after Create()
+	_, err = librbd.Create4(ioctx, pOpts.RbdImageName, uint64(volSz*util.MiB), options)
```
Madhu-1: `volSzMiB` is used in this function anywhere?

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,194 @@
...
+		// there was a race, oldCe already exists
+		ce.(*connEntry).destroy()
+		// TODO: increase refcount on oldCe
+		return oldCe.(*connEntry).conn, unique, nil
+	}
+
+	return conn, unique, nil
+}
+
+func (cp *ConnPool) Put(unique string) {
```
Madhu-1: need to add function explanation (what it does)

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,194 @@
...
+	ce.lastUsed = time.Now()
+	ce.users++
+}
+
+// Reduce number of references. If this returns true, there are no more users.
+// /!\ Only call this while holding the ConnPool.lock.
+func (ce *connEntry) put() bool {
+	ce.users--
+	return ce.users == 0
+	// do not call ce.destroy(), let ConnPool.gc() do that
```
Madhu-1: remove comment?

.travis.yml
nixpanic: ARM builds are currently best effort, and with some releases they are not available. The additional repository contains a build that can be used to fall-back on (got it from the Ceph release maintainers). As ARM systems become more common, I expect to see builds for it improving.
The Ceph team also provides the ceph:ceph base image we consume, that has similar limitations as the builds of the packages.

pkg/rbd/rbd_util.go
nixpanic: Yes, for logging on line 137/138.

pkg/util/conn_pool.go
nixpanic: This comment is intentional. Most reference counting approaches release the objects on `put()` or `release()` calls. It is an exception to not do that in this case where the connection is cached.

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,194 @@
...
+	cp.lock.Lock()
+	defer cp.lock.Unlock()
+
+	ce, ok := cp.conns.Load(unique)
+	if !ok {
+		return
+	}
+
+	if ce.(*connEntry).put() {
+		cp.conns.Delete(unique)
```
nixpanic: This is actually not right. The key should not be removed from the `cp.conns` map, it would not be possible to retrieve it afterwards.

pkg/util/conn_pool.go
nixpanic: Explanation has been added, and updated the `ConnPool.Get()` method to refer to this too.

humblec:
@nixpanic with createVolume of current implementation this is in parity on features, Isnt it  ?

nixpanic:
On Mon, Jan 13, 2020 at 12:06:19AM -0800, Humble Devassy Chirammal wrote:
> @nixpanic with createVolume of current implementation this is in parity on features, Isnt it  ?

Yes, it should be. go-ceph got Create4() added just for the purpose to
select a different data-pool.

The only question that currently remains is about the ImageFormat.
librbd does not support anything else than version 2. Was there ever a
Cehp-CSI release where ImageFormat 1 was a functional option?

pkg/rbd/controllerserver.go
```diff
@@ -698,6 +703,7 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	defer cr.DeleteCredentials()
 
 	rbdVol := &rbdVolume{}
+	defer rbdVol.Destroy()
```
Madhu-1: instead of calling Destroy on Each function, we can call in in one function (maybe `DeleteCredentials` function) this   need to be taken care in next patch

pkg/rbd/rbd_util.go
Madhu-1: Niels, can you please test that we should be able to delete PVC created by csi v1.0.0 driver

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,196 @@
+/*
+Copyright 2020 The Ceph-CSI Authors.
+
```
Madhu-1: can we add unit testing for this file?

pkg/rbd/controllerserver.go
nixpanic: This can easily be done after https://github.com/nixpanic/ceph-csi/commit/c304e9220b0665c234037a37ed94e7381db66158 as then the credentials are only passed to `rbVol.Connect(cr)`. `createImage()` can be refactored quite a bit then too.

pkg/util/conn_pool.go
nixpanic: Yes, I'll do something by not creating a real connection to a Ceph cluster. That way `go test` does not rely on a Ceph environment.

nixpanic:
> @nixpanic looks good, have few comments, can you also create some 100+ parallel PVC and test it? just want to make sure nothing breaks.

Things do not break when creating many volume. But on the other hand, the current replacement of the few rbd commands do not bring much performance improvement either. Running ceph-csi v.1.2 and this current implementation (almost v2.0.0) both take between 60-65 seconds for 100 PVCs on a OCP-4.4 cluster based on libvirt VMs

This has been moved out or v2.0.0 planning, until more rbd commands have been replaced (done by follow-up PRs).

phlogistonjohn:
When you next update this PR I'd like to suggest that you define an interface type for the (minimal functions of the) connection pool. Use the interface outside of the setup code and implementation of the pool. Then you can also define a naive/fake pool type for testing & benchmarking without having to replace the implementation you have.
You also don't want to use a lock and a sync.Map. Use one or the other.

humblec:
@nixpanic may be good time to revisit this PR ? :)

pkg/rbd/rbd_util.go
```diff
@@ -109,36 +114,80 @@ type rbdSnapshot struct {
...
+	if pOpts.ImageFeatures != "" {
+		// TODO: parse imagefeatures and set RbdImageOptionFeatures
+		// err = options.SetUint64(librbd.RbdImageOptionFeatures, 0)
+		klog.Infof(util.Log(ctx, "ImageFeatures (set to \"%v\" are currently "+
+			"ignored"), pOpts.ImageFeatures)
+	}
 
-	if pOpts.DataPool != "" {
-		args = append(args, "--data-pool", pOpts.DataPool)
+	ioctx, err := pOpts.getIoctx(cr)
```
ShyamsundarR: I prefer not requesting a connection per call and returning it, instead would be easier to get it once in credentials (or otherwise). Saves adding code to each call, adding the get and put wrappers.

Also, is there a plan to retain the IO context and reuse it across calls that require the same?

nixpanic: I thought about putting everything in the credentials. It  might be cleaner to replace the passing of the credentials with a connection, or IOContext. That is something I want to do at a later time though.

pkg/util/conn_pool.go
nixpanic: Tests for the ConnPool have been added now

nixpanic:
@humblec @Madhu-1 care to review this again?

pkg/rbd/rbd_util.go
```diff
@@ -87,6 +87,11 @@ type rbdVolume struct {
 	DisableInUseChecks bool   `json:"disableInUseChecks"`
 	Encrypted          bool
 	KMS                util.EncryptionKMS
+
+	// connection
```
Madhu-1: if we keep this is Credentials structure, can we reuse the connection for rados and ceph fs also?

pkg/rbd/rbd_util.go
Madhu-1: if we do that changes we are removing this part of code right?

pkg/rbd/rbd_util.go
```diff
@@ -550,6 +592,21 @@ func hasSnapshotFeature(imageFeatures string) bool {
...
+// imageFeaturesToUint takes the comma separated image features and converts
+// them to a RbdImageOptionFeatures value.
+func imageFeaturesToUint(ctx context.Context, imageFeatures string) uint64 {
+	features := uint64(0)
+
+	for _, f := range strings.Split(imageFeatures, ",") {
+		if f == "layering" {
+			features |= librbd.RbdFeatureLayering
+		} else {
+			klog.V(4).Infof(util.Log(ctx, "rbd: image feature %s not recognized, skipping"), f)
```
Madhu-1: warning log?

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,198 @@
...
+		delete(cp.conns, key)
+	}
+}
+
+func (cp *ConnPool) generateUnique(pool, monitors, keyfile string) string {
+	return fmt.Sprintf("%s|%s|%s", pool, monitors, keyfile)
+}
+
+// getExisting returns the existing rados.Conn associated with the unique key.
+//
```
Madhu-1: remove this line

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,198 @@
...
+	return nil
+}
+
+// Return a rados.Conn for the given arguments. Creates a new rados.Conn in
+// case there is none in the pool. Use the returned unique string to reduce the
+// reference count with ConnPool.Put(unique).
+func (cp *ConnPool) Get(pool, monitors, keyfile string) (*rados.Conn, string, error) {
+	unique := cp.generateUnique(pool, monitors, keyfile)
+
+	cp.lock.RLock()
```
Madhu-1: how about moving this Read lock to getConn function. as the consumer dont need to bother about taking lock

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,198 @@
...
+	ce := &connEntry{
+		conn:     conn,
+		lastUsed: time.Now(),
+		unique:   unique,
+		users:    1,
+	}
+
+	cp.lock.Lock()
+	defer cp.lock.Unlock()
+	oldConn := cp.getConn(unique)
```
Madhu-1: can this be checked before establishing the connection to the ceph cluster? if the old connect exists reuse it.

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	// this really is a new connection, add it to the map
+	cp.conns[unique] = ce
+
+	return conn, unique, nil
+}
+
+// Reduce the reference count of the rados.Conn object identified by unique
+// that was returned with ConnPool.Get().
+func (cp *ConnPool) Put(unique string) {
```
Madhu-1: can we rename this to something else, IMHO `Put` seems like we are storing/adding some data in general terms

pkg/util/conn_pool_test.go
```diff
@@ -0,0 +1,166 @@
...
+package util
+
+import (
+	"testing"
+	"time"
+
+	"github.com/ceph/go-ceph/rados"
+)
+
+var (
```
Madhu-1: can this be const instead of var

pkg/util/conn_pool_test.go
```diff
@@ -0,0 +1,166 @@
...
+			t.Errorf("there is more than a single conn in cp.conns: %v", len(cp.conns))
+		}
+
+		// the ce should have a single user
+		ce, exists := cp.conns[unique]
+		if !exists {
+			t.Errorf("getting the conn from cp.conns failed")
+		}
+		if ce.users != 1 {
+			t.Errorf("There should only be one user: %v", ce.users)
```
Madhu-1: replace `There` with `there` here and also below error messages

pkg/util/conn_pool.go
nixpanic: It is checked before, but there can be a race as multiple operations may request a new connection. The connection that ends up in the map will be used in the end. I do not think it is worth it to check more often, the chances that this happens is really small.

pkg/util/conn_pool.go
nixpanic: It is the whole idea that the private `getConn()` does not take the lock. This is used for the testing, where locks would prevent certain tests.

pkg/util/conn_pool.go
nixpanic: With reference counting Get()/Put() are commonly used. I can rename it to Ref()/Unref() if you prefer, but the Get() makes that feel a little awkward.

pkg/rbd/rbd_util.go
nixpanic: Yes, it can be moved to the Credentials structure. I want to make changes like that in follow-up patches, so that there is no need to pass credentials around anymore.

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,198 @@
...
+func (cp *ConnPool) Destroy() {
+	cp.timer.Stop()
+	// wait until gc() has finished, in case it is running
+	cp.lock.Lock()
+	defer cp.lock.Unlock()
+
+	for key, ce := range cp.conns {
+		if ce.users != 0 {
+			// should never happen
+			// TODO: return error, or panic()?
```
phlogistonjohn: If the only way for this to happen is a logic bug in this code, I'd suggest to panic with a very clear message.

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,198 @@
...
+			return
+		}
+
+		ce.destroy()
+		delete(cp.conns, key)
+	}
+}
+
+func (cp *ConnPool) generateUnique(pool, monitors, keyfile string) string {
+	return fmt.Sprintf("%s|%s|%s", pool, monitors, keyfile)
```
phlogistonjohn: Is this string exposed to the users anywhere? The pipe chars as delimiters looks a little odd to my eyes, and I suspect a less informed user's too.

pkg/util/conn_pool.go
phlogistonjohn: Acquire/Release?

pkg/rbd/rbd_util.go
```diff
@@ -112,36 +119,93 @@ type rbdSnapshot struct {
...
+	err = librbd.CreateImage(ioctx, pOpts.RbdImageName, uint64(volSz*util.MiB), options)
 	if err != nil {
-		return errors.Wrapf(err, "failed to create rbd image, command output: %s", string(output))
+		return errors.Wrapf(err, "failed to create rbd image")
 	}
 
 	return nil
 }
 
+func (rv *rbdVolume) getIoctx(cr *util.Credentials) (*rados.IOContext, error) {
```
phlogistonjohn: Functions that don't use its receiver proably should just be top-level helper functions without being glued to a particular type.

pkg/util/conn_pool.go
nixpanic: No, it is internal to the ConnPool only. Well, the key does get returned by `Get()` so that `Put()` can be used with it. I could probably hide it completely by making ConnEntry contain the rados.Conn and add the Put() or Release() as a method of the type.

pkg/rbd/rbd_util.go
nixpanic: I think this does use its receiver? The functions I modified got their 'receiver' as an argument. The partial object-oriented, partial procedural approach is rather ugly. Hopefully we can move to a more object-oriented approach with follow-up changes.

pkg/util/conn_pool.go
nixpanic: I've removed the unique string usage, but kept Get()/Put() for now. Release() sounds like it is the last user that needs to call it (like filedescriptors can have multiple close() calls, but only the last one also calls release() in the kernel/fuse).

pkg/rbd/rbd_util.go
nixpanic: Deleting an PVC created with an older version works fine.

Madhu-1:
@nixpanic can you fix CI issue.

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// Reduce number of references. If this returns true, there are no more users.
+// /!\ Only call this while holding the ConnPool.lock.
+func (ce *connEntry) put() {
+	ce.users--
+	// do not call ce.destroy(), let ConnPool.gc() do that
+}
+
+// Destroy a connEmtry object, close the connection to the Ceph cluster.
```
humblec: s/connEmtry/connEntry

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,205 @@
...
+			panic("this connEntry still has users, operations" +
+				"might still be in-flight")
+		}
+
+		ce.destroy()
+		delete(cp.conns, key)
+	}
+}
+
+func (cp *ConnPool) generateUnique(pool, monitors, keyfile string) (string, error) {
```
humblec: May be `generateUniqueKey` is better than `generateUnique`

pkg/util/conn_pool.go
```diff
@@ -0,0 +1,205 @@
...
+	// the keyfile can be uique for operations, contents will be the same
```
humblec: s/uique/unique

pkg/rbd/rbd_util.go
```diff
@@ -558,6 +595,21 @@ func hasSnapshotFeature(imageFeatures string) bool {
 	return false
 }
 
+// imageFeaturesToUint takes the comma separated image features and converts
+// them to a RbdImageOptionFeatures value.
+func imageFeaturesToUint(ctx context.Context, imageFeatures string) uint64 {
+	features := uint64(0)
+
+	for _, f := range strings.Split(imageFeatures, ",") {
```
humblec: This logic basically means, we only account `layering` feature.  Cant we just check for `,layering,` in the string (.Contains) and return `uint64` value ? why because, as this func returns unique/single `uint64` value, I doubt this looping and condition check going to make sense even in future for more options.

pkg/rbd/rbd_util.go
```diff
@@ -558,6 +595,21 @@ func hasSnapshotFeature(imageFeatures string) bool {
 	return false
 }
 
+// imageFeaturesToUint takes the comma separated image features and converts
+// them to a RbdImageOptionFeatures value.
+func imageFeaturesToUint(ctx context.Context, imageFeatures string) uint64 {
```
humblec: it should be `...Uint64`

humblec:
@nixpanic apart from correcting CI , I also have some review comments.

pkg/rbd/rbd_util.go
```diff
@@ -112,36 +115,76 @@ type rbdSnapshot struct {
 
 var (
 	supportedFeatures = sets.NewString("layering")
+
+	// large interval and timeout, it should be longer than the maximum
+	// time an operation can take (until refcounting of the connections is
```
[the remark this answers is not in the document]
nixpanic: the timeout is only used for triggering the garbage collector, as long as there is an active user of the connection, it is not disconnected. Using a larger timeout might result in more connections to be cached (in the case there are more StorageClasses with different options).

pkg/rbd/rbd_util.go
```diff
@@ -112,36 +115,76 @@ type rbdSnapshot struct {
...
+	// available)
+	connPool = util.NewConnPool(15*time.Minute, 10*time.Minute)
```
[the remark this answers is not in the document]
nixpanic: sure, I'll do that

pkg/rbd/rbd_util.go
nixpanic: The next release of go-ceph will have a way to map feature strings to uint64's in a better way. When that is available, this function will be rewritten so that new features automatically will be supported.

See https://github.com/ceph/go-ceph/blob/master/rbd/features.go#L110 for more details.

nixpanic:
@humblec thanks for the review, please check again

humblec:
@nixpanic thanks for addressing the comments. Reviewing again ..

humblec:
As we are planning for more improvements in a follow up PR, this looks good for first version. Any comments here @Madhu-1 and @ShyamsundarR

nixpanic:
Conflicts introduced with a recent merge have been resolved. Tests should pass again and ready for merging once reviewed.

pkg/rbd/rbd_util.go
```diff
@@ -112,36 +115,79 @@ type rbdSnapshot struct {
...
+	cpExpiry   = 10 * time.Minute
+	connPool   = util.NewConnPool(cpInterval, cpExpiry)
 )
 
 // createImage creates a new ceph image with provision and volume options.
 func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
-	var output []byte
-
-	image := pOpts.RbdImageName
 	volSzMiB := fmt.Sprintf("%dM", util.RoundOffVolSize(pOpts.VolSize))
```
Madhu-1: I think we can remove this `volSzMiB` now. this is only used for logging now

Madhu-1:
@nixpanic do we need to backport this to releasev2.0 branch now? do we need to backport it later once we have all the pieces to support native-go?

Issue #730, opened 2019-12-05
Add checking for rook rbdpool and cephfilesystem


woohhan:
# Describe the feature you'd like to have #
Currently in the rook deploy script, we only check ceph health. We can add checking rbd pool and cephfilesystem for safety. Thanks for suggestion @Madhu-1 

# What is the value to the end user? (why is it a priority?) #
NONE

# How will we know we have a good solution? (acceptance criteria) #
We can ensure the rook is ready to rbd/cephfs test.

# Additional context #
NONE

Madhu-1:
@WoohyungHan would you like to work on this one? if yes i can assign this to you

WoohyungHan:
@Madhu-1 sure, i will

Madhu-1:
done, Thanks for picking this up

PR #787, opened 2020-01-14
Rook deploy script, Adding rbd_pool check and cephfilesystem check.

sophalHong:
- **script/minikube.sh** : adding current directory for executing `rook.sh deploy|teardown`
- **script/rook.sh** : Adding more code to check whether `cephfilesystem` and `rbdpool` is successfully created.

Do you have any questions?
No

No

No

Nothing...

## Related issues ##
Fixes: #730

## Future concerns ##

### DIFF
--- a/scripts/rook.sh
+++ b/scripts/rook.sh
@@ -12,14 +12,20 @@ function deploy_rook() {
     kubectl create -f "${ROOK_URL}/filesystem-test.yaml"
     kubectl create -f "${ROOK_URL}/pool-test.yaml"
 
-    for ((retry=0; retry<=ROOK_DEPLOY_TIMEOUT; retry=retry+5)); do
-        echo "Wait for rook deploy... ${retry}s"
-        sleep 5
+	# Check if CephCluster is empty
+	if ! kubectl -n rook-ceph get cephclusters -oyaml | grep 'items: \[\]' &> /dev/null; then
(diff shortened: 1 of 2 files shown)


scripts/rook.sh
```diff
@@ -12,14 +12,31 @@ function deploy_rook() {
...
 
-    for ((retry=0; retry<=ROOK_DEPLOY_TIMEOUT; retry=retry+5)); do
-        echo "Wait for rook deploy... ${retry}s"
-        sleep 5
-
-        if kubectl get cephclusters -n rook-ceph | grep HEALTH_OK &> /dev/null; then
-            break
-        fi
-    done
+#    for ((retry=0; retry<=ROOK_DEPLOY_TIMEOUT; retry=retry+5)); do
```
Madhu-1: need to remove this  commented code

sophalHong: yes, I did.

scripts/rook.sh
```diff
@@ -31,6 +37,61 @@ function teardown_rook() {
     kubectl delete -f "${ROOK_URL}/common.yaml"
 }
 
+function check_ceph_cluster_health(){
+	while true; do
```
Madhu-1: the check should not be infinite, have some constant value for t loop

PR #794, opened 2020-01-16
rbd: only load nbd module if not available yet

nixpanic:

In case the attacher is not allowed to load kernel modules, there is not need to try to `modprobe nbd`. There is also a remote possibility that the `nbd` kernel module is compiled into the kernel (no loadable modules at all for security is a thing).

## Future concerns ##

By default only 16 nbd devices are available. On systems where more than 16 devices are needed, the kernel module needs to be loaded with additional options (see `modinfo nbd`). Modifying the number of devices after the module is loaded, does not seem to be possible atm.


humblec:
Getting this in `2.0.0`.

PR #797, opened 2020-01-17
util: use MiB and GiB constants from k8s helpers

wilmardo:

Removes a const definition now available in the upstream:
https://github.com/kubernetes/cloud-provider/blob/master/volume/helpers/rounding.go#L25

This should be backward compatible and can safely be removed now :)

### DIFF
--- a/pkg/rbd/rbd_util.go
+++ b/pkg/rbd/rbd_util.go
@@ -38,6 +38,7 @@ import (
 	"github.com/pborman/uuid"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/cloud-provider/volume/helpers"
 	"k8s.io/klog"
 )
 
@@ -168,7 +169,7 @@ func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 	defer ioctx.Destroy()
 
 	err = librbd.CreateImage(ioctx, pOpts.RbdImageName,
-		uint64(util.RoundOffVolSize(pOpts.VolSize)*util.MiB), options)
+		uint64(util.RoundOffVolSize(pOpts.VolSize)*helpers.MiB), options)
 	if err != nil {
 		return errors.Wrapf(err, "failed to create rbd image")
 	}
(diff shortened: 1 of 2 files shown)


Madhu-1:
CI failure

nixpanic:
This fails to build with errors like 
```
--- testing: github.com/ceph/ceph-csi/cmd ---
# github.com/ceph/ceph-csi/pkg/util
pkg/util/util.go:39:16: undefined: MiB
pkg/util/util.go:49:18: undefined: GiB
pkg/util/util.go:50:38: undefined: MiB
pkg/util/util.go:51:10: undefined: MiB
pkg/util/util.go:53:38: undefined: GiB
pkg/util/util.go:54:10: undefined: GiB
```

This should be a trivial correction, please let us know if you want someone else to take this over from you, @wilmardo

wilmardo:
@nixpanic Fixed it :)

It think some of the logic can be improved by using the `RoundUpTo*` functions from upstream but I have no time to look into it at the moment. Golang is still very new to me so everything is a challenge ;)

I had something like this working but the logic was off, leaving it here for future reference.
```golang
q := resource.NewQuantity(size, resource.BinarySI)
return helpers.RoundUpToMiB(*q)
```

pkg/rbd/rbd_util.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog"
+	"k8s.io/cloud-provider/volume/helpers"
```
nixpanic, line 42: This new import will fail testing. Go prefers grouped imports to be sorted, `klog` and `helpers` should be switched.

These things can now be easier caught on developers system with commands like:

    make containerized-test TARGET=static-check

The container for testing will be built once, and re-used for new testing, Drop the `TARGET=static-check` to have all tests run (well, after #964).

nixpanic:
Hi @wilmardo, I have updated your PR so that CI should pass. The commit message now mentions the purpose of the change a little better too. Your authorship has been kept ,FWIW.

I re-considered asking you going through the modify/test/push process, hope you don't mind!

Thanks for the contribution!
Niels

pkg/rbd/rbd_util.go
wilmardo: Thanks for the pointer, containerized tests are awesome!
I was struggling locally with an plain Alpine container and installing the deps myself.

Might be nice to add something about testing locally to the [development-guide.md](https://github.com/ceph/ceph-csi/blob/master/docs/development-guide.md)?

Issue #407, opened 2019-06-06
How to create a pre-provisioned PV for an existing rbd image


itmuckel:
In https://kubernetes.io/blog/2019/01/15/container-storage-interface-ga/ under the section "Pre-Provisioned Volumes" is a template of how to create a PV from a prepopulated CSI volume. What values do I need to create a PV that references an rbd image? Once the PV is created, I'm sure just creating a PVC will bind automatically to the PV.

I've read a lot of the official kubernetes docs, but how do I get the csi driver, volume handle, fsType? What volumeAttributes do I need? What secrets do I need to reference.

I've set up ceph-csi using rook following the following documentation: https://rook.io/docs/rook/v1.0/ceph-csi-drivers.html

ShyamsundarR:
I did a round of tests with pre-provisioned CephFS volumes (or directories) and exposing them as static PVs or pre-provisioned volumes, the requires YAML for that test/example can be found [here](https://gist.github.com/ShyamsundarR/1009b79966992470f588a277429ae2e1). This maybe a little dated, but can serve as an example for RBD as well.

Let me know if this helps, else I can create one for RBD as well and post it.

itmuckel:
@ShyamsundarR Yes, please show me how a file for rbd restoration should look.

My current file looks like this:
```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: verdaccio-restored-from-rbd
spec:
  accessModes:
    - ReadWriteOnce
  capacity:
    storage: 8Gi
  csi:
    driver: rbd.csi.ceph.com
    volumeHandle: verdaccio-pvc-backup
    readOnly: false
    fsType: ext4
    nodeStageSecretRef:
      name: csi-rbd-secret
      namespace: csi
    controllerPublishSecretRef:
      name: csi-rbd-secret
      namespace: csi
    nodePublishSecretRef:
      name: csi-rbd-secret
      namespace: csi
  persistentVolumeReclaimPolicy: Retain
  claimRef:
    namespace: verdaccio-test-backup
    name: verdaccio-restored-from-rbd

---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  namespace: verdaccio-test-backup
  name: verdaccio-restored-from-rbd
spec:
  storageClassName: csi-rbd
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 8Gi
```

[...]

PR #799, opened 2020-01-20
Add support for rbd static PVC

Madhu-1:
* Add support for mounting and unmounting static PVC
* Add E2E for rbd static PVC
* Add doc for rbd static PVC

Fixes: https://github.com/ceph/ceph-csi/issues/407

### DIFF
--- a/docs/static-pvc.md
+++ b/docs/static-pvc.md
@@ -1,0 +1,119 @@
+# Static PVC with ceph-csi
+
+- [Static PVC with ceph-csi](#static-pvc-with-ceph-csi)
+  - [RBD static PVC](#rbd-static-pvc)
+    - [Create RBD image](#create-rbd-image)
+    - [Create RBD static PV](#create-rbd-static-pv)
+    - [RBD Volume Attributes in PV](#rbd-volume-attributes-in-pv)
+    - [Create RBD static PVC](#create-rbd-static-pvc)
+
+This document outlines how to create static PV and static PVC from existing rbd image
+
+**warning** static PVC can be created, deleted, mounted and unmounted but
+currently ceph-csi doesn't support other operations like snapshot,clone,
+resize, etc for static PVC
+
+## RBD static PVC
+
+RBD images created manually can be mounted and unmounted to an app, below step
+shows how to create a rbd image, static PV, static PVC
+
+### Create RBD image
+
+If you already have a rbd image created and contains some data which you want
+to access by the application pod you can skip this step.
+
+Lets create a new rbd image in ceph cluster which we are going to use for
+static PVC
+
+```bash
+[$]rbd create static-image --size=1024 --pool=replicapool
+```
+
+### Create RBD static PV
+
+To create the rbd PV you need to know the `rbd image name`,`clusterID` and
+`pool` name in which the rbd image is created
+
+```yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: fs-static-pv
+spec:
+  accessModes:
+  - ReadWriteOnce
+  capacity:
+    storage: 1Gi
+  csi:
+    driver: rbd.csi.ceph.com
+    fsType: ext4
+    nodeStageSecretRef:
+      # node stage secret name
+      name: csi-rbd-secret
+      # node stage secret namespace where above secret is created
+      namespace: default
+    volumeAttributes:
+      # Required options from storageclass parameters need to be added in volumeAttributes
+      "clusterID": "ba68226a-672f-4ba5-97bc-22840318b2ec"
+      "pool": "replicapool"
+      "staticVolume": "true"
+      #mounter: rbd-nbd
+    # volumeHandle should be same as rbd image name
+    volumeHandle: static-image
+  persistentVolumeReclaimPolicy: Retain
+  # The volumeMode can be either `Filesystem` or `Block` if you are creating Filesystem PVC it should be `Filesystem`, if you are creating Block PV you need to change it to `Block`
+  volumeMode: Filesystem
+```
+
+### RBD Volume Attributes in PV
+
+Below table explains the list of volume attributes can be set when creating a
+static RBD PV
+
+|  Attributes  |                                                                Description                                                                 | Required |
+| :----------: | :----------------------------------------------------------------------------------------------------------------------------------------: | :------: |
+|  clusterID   | The is used by the CSI plugin to uniquely identify and use a Ceph cluster (this is the key in configmap created duing ceph-csi deployment) |   Yes    |
+|     pool     |                                                The pool name in which rbd image is created                                                 |   Yes    |
+| staticVolume |                                      Value must be set to `true` to mount and unmount static rbd PVC                                       |   yes    |
+|   mounter    |                 If set to `rbd-nbd`, use `rbd-nbd` on nodes that have `rbd-nbd` and `nbd` kernel modules to map rbd images                 |    No    |
+
+**Note** ceph-csi does not supports rbd image deletion for static PV.
+`persistentVolumeReclaimPolicy` in PV spec must be set to `Retain` to avoid PV
+delete attempt in csi-provisioner.
+
+```bash
+[$] kubectl create -f fs-static-pv.yaml
+persistentvolume/fs-static-pv created
+```
+
+### Create RBD static PVC
+
+To create the rbd PVC you need to know the PV name which is created above
+
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: fs-static-pvc
+  namespace: default
+spec:
+  accessModes:
+  # ReadWriteMany is only supported for Block PVC
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+  # The volumeMode can be either `Filesystem` or `Block` if you are creating Filesystem PVC it should be `Filesystem`, if you are creating Block PV you need to change it to `Block`
+  volumeMode: Filesystem
+  # volumeName should be same as PV name
+  volumeName: fs-static-pv
+```
+
+```bash
+[$] kubectl create -f fs-static-pvc.yaml
+persistentvolumeclaim/fs-static-pvc created
+```
+
+**Note** deleting PV and PVC doesnot deleted the backend rbd image, user need to
+manually delete the rbd image if required
--- a/pkg/rbd/nodeserver.go
+++ b/pkg/rbd/nodeserver.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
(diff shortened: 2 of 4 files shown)


docs/static-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod-with-raw-block-volume
+spec:
+  containers:
+    - name: fc-container
+      image: fedora:26
+      command: ["/bin/sh", "-c"]
+      args: ["tail -f /dev/null"]
```
nixpanic: What does this do? Can you not just use `sleep infinity` instead?

docs/static-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+      name: rook-csi-rbd-node
+      # node stage secret namespace where above secret is created
+      namespace: rook-ceph
+    volumeAttributes:
+      # cluster id specified in configmap during csi deployment
+      clusterID: abcd
+      imageFeatures: layering
+      imageFormat: "2"
+      pool: replicapool
+      staticVolume: "true"
```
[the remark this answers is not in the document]
Madhu-1: all keys and values should be in string form

docs/static-pvc.md
Madhu-1: This example is copied from https://github.com/ceph/ceph-csi/blob/master/examples/rbd/raw-block-pod.yaml

nixpanic:
On Tue, Jan 21, 2020 at 01:08:13AM -0800, Madhu Rajanna wrote:
> Madhu-1 commented on this pull request.
> 
> 
> 
> > +    storage: 4Gi
> +  csi:
> +    driver: rbd.csi.ceph.com
> +    fsType: ext4
> +    nodeStageSecretRef:
> +      # node stage secret name
> +      name: rook-csi-rbd-node
> +      # node stage secret namespace where above secret is created
> +      namespace: rook-ceph
> +    volumeAttributes:
> +      # cluster id specified in configmap during csi deployment
> +      clusterID: abcd
> +      imageFeatures: layering
> +      imageFormat: "2"
> +      pool: replicapool
> +      staticVolume: "true"
> 
> all keys and values should be in string form

In YAML everything is a string, except for integers...

docs/static-pvc.md
nixpanic: We should improve it there too.

And not use a **very** old fedora image in the references either.

nixpanic: See #805 for a better solution :wink:

docs/static-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+doesn't support other operations like (snapshot, resize, etc)
+
+## RBD static volume
+
+rbd images created manually can be mounted and unmounted to an app, below step
+shows how to create a static PV, static PVC and bind PVC to an application
+
+### Create static RBD Filesystem PV
+
+to create the rbd PV you need to know the `rbd image name`,`clusterID` and
```
ShyamsundarR: (nit) capitalize every section start?

docs/static-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+    volumeAttributes:
+      # cluster id specified in configmap during csi deployment
+      clusterID: abcd
+      imageFeatures: layering
+      imageFormat: "2"
+      pool: replicapool
+      staticVolume: "true"
+    # name of the rbd image created
+    volumeHandle: rbd-test
+  persistentVolumeReclaimPolicy: Retain
```
ShyamsundarR: Need to explicitly mention setting `persistentVolumeReclaimPolicy: Retain` else on PVC delete a delete will be attempted with the provisioner, which will not pass.

docs/static-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+[$] kubectl delete pvc/fs-static-pvc
+[$] kubectl delete pc/fs-static-pv
+```
+
+*Note* deleting PV and PVC doesnot deleted the backend rbd image, user need to
+manually delete the rbd image if required
+
+### Create static RBD Block PV
+
+to create the rbd PV you need to know the `rbd image name`,`clusterID` and
```
ShyamsundarR: This section is mostly a repeat of the filesystem PV other than ` volumeMode: Block`, I would suggest just calling that part out and not repeating the entire process. Helps keeps the reading shorter and also future maintainability easier.

docs/static-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+    storage: 4Gi
+  csi:
+    driver: rbd.csi.ceph.com
+    fsType: ext4
+    nodeStageSecretRef:
+      # node stage secret name
+      name: rook-csi-rbd-node
+      # node stage secret namespace where above secret is created
+      namespace: rook-ceph
+    volumeAttributes:
```
ShyamsundarR: I think we need to leave a note about `volumeAttributes` and that all parameters in the storage class can be added here.

docs/static-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+**warning** static PVC can be created, deleted, mounted and unmounted but
+doesn't support other operations like (snapshot, resize, etc)
+
+## RBD static volume
+
+rbd images created manually can be mounted and unmounted to an app, below step
+shows how to create a static PV, static PVC and bind PVC to an application
+
+### Create static RBD Filesystem PV
+
```
ShyamsundarR: We need to mention that `staticVolume` parameter should be mentioned in the PV, as that detail seems to be buried in the YAML otherwise.

docs/static-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+metadata:
+  name: fs-static-pv
+spec:
+  accessModes:
+  - ReadWriteOnce
+  capacity:
+    storage: 4Gi
+  csi:
+    driver: rbd.csi.ceph.com
+    fsType: ext4
```
idryomov: How is the filesystem created?  If `mountVolumeToStagePath()` is involved, it needs an update:
https://github.com/ceph/ceph-csi/blob/d0eabca1f2dac9346dccad05ea0f6e7a0032ceeb/pkg/rbd/nodeserver.go#L318-L319

Madhu-1: @idryomov nice catch will fix this one also

pkg/rbd/nodeserver.go
```diff
@@ -315,7 +326,7 @@ func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeS
 		return err
 	}
 	// TODO: update this when adding support for static (pre-provisioned) PVs
```
idryomov: Can drop this TODO now.

docs/static-pvc.md
```diff
@@ -0,0 +1,117 @@
...
+    nodeStageSecretRef:
+      # node stage secret name
+      name: csi-rbd-secret
+      # node stage secret namespace where above secret is created
+      namespace: default
+    volumeAttributes:
+      # Required options from storageclass parameters need to be added in volumeAttributes
+      clusterID: ba68226a-672f-4ba5-97bc-22840318b2ec
+      imageFeatures: layering
```
idryomov: Just curious, why does one need to specify image features for an image that is to be created and removed manually?

docs/static-pvc.md
```diff
@@ -0,0 +1,117 @@
...
+| :-----------: | :----------------------------------------------------------------------------------------------------------------------------------------: | :------: |
+|   clusterID   | The is used by the CSI plugin to uniquely identify and use a Ceph cluster (this is the key in configmap created duing ceph-csi deployment) |   Yes    |
+| imageFeatures |                                      CSI RBD currently supports images with only `layering` feature.                                       |    No    |
+|     pool      |                                                The pool name in which rbd image is created                                                 |   Yes    |
+| staticVolume  |                                      Value must be set to `true` to mount and unmount static rbd PVC                                       |   yes    |
+|    mounter    |                 If set to `rbd-nbd`, use `rbd-nbd` on nodes that have `rbd-nbd` and `nbd` kernel modules to map rbd images                 |    No    |
+
+**Note** ceph-csi does not supports rbd image deletion for static PV.
+`persistentVolumeReclaimPolicy` in PV spec must be set to `Delete` to avoid PV
```
idryomov: I think this should be `Retain`.

docs/static-pvc.md
Madhu-1: This is not mandatory, i have added a table below about required flags, anyhow will remove it

Madhu-1:
E2E passed for me  locally, let's see in CI

Madhu-1:
@ShyamsundarR @idryomov @humblec PTAL

docs/static-pvc.md
```diff
@@ -0,0 +1,115 @@
...
+- [Static PVC with ceph-csi](#static-pvc-with-ceph-csi)
+  - [RBD static PVC](#rbd-static-pvc)
+    - [Create RBD image](#create-rbd-image)
+    - [Create RBD static PV](#create-rbd-static-pv)
+    - [RBD Volume Attributes in PV](#rbd-volume-attributes-in-pv)
+    - [Create RBD static PVC](#create-rbd-static-pvc)
+
+This document outlines how to create static PV and static PVC from existing rbd image
+
+**warning** static PVC can be created, deleted, mounted and unmounted but
```
dillaman: Nit: while not an immediate concern, I can see the use for being able to clone a static PV into a dynamic PV for the purposes of recovery. i.e. you are using static PVs because you had some type of failure and you need access to your pre-existing PVs again. If we don't offer a path to get out of this state, they will forever be stuck as static PVs.

docs/static-pvc.md
```diff
@@ -0,0 +1,115 @@
...
+doesn't support other operations like (snapshot, resize, etc)
+
+## RBD static PVC
+
+RBD images created manually can be mounted and unmounted to an app, below step
+shows how to create a rbd image, static PV, static PVC
+
+### Create RBD image
```
dillaman: Do we really want to encourage manual image creation to plug into a static PV or should this really be focused on recovery. i.e. how to find an existing RBD image for a given PV?

Madhu-1: yes, we can add it also, the main idea for this doc is to use the rbd image (which may not be created by CSI driver itself) as an enhancement we can do this also.

currently reworded the statement, we can add recovery steps later.

docs/static-pvc.md
Madhu-1: yes, Jason, I agree, once we add support for other operation like a clone, etc. we can remove this warning just added here so that users should be aware of it for now

Madhu-1:
resolved merge conflict. @ShyamsundarR @humblec PTAL

humblec:
@Madhu-1 may be a quick touch on this pr to make CI happy. :)

humblec:
LGTM. Thanks.

PR #808, opened 2020-01-23
Check for ENOENT errors when deleting CephFS volumes

### LINKED ISSUE #474: getVolumeRootPathCeph should detect and return ENOENT errors
When a subvolume is deleted but before its relevant RADOS objects and keys are deleted a service fault occurs, then the next delete request is expected to garbage collect the same. This is done on errors where the image is not found in the background, but RADOS information leads us to the image (like in `newVolumeOptionsFromVolID`).

To satisfy this requirement and edge case, ENOENT errors from `getVolumeRootPathCeph` need to be detected and appropriately handled in the DeleteVolume request to only garbage collect the RADOS information stored for the same.

This is akin to returning `ErrImageNotFound` from `getImageInfo` and handling the same in DeleteVolume request in the rbd plugin code.

ShyamsundarR:
In cases where a prior DeleteVolume deleted the backing subvolume, and was interrupted post the same, would leave behind an in-tact CSI OMap entries, but no backing image.

In such cases, detecting that the backing image is deleted and proceeding with garbage collecting the OMaps should be the way forward for future DeleteVolume calls. Current code does not honor the same, as detailed in issue #474.

This PR attempts to rectify the same, adding the required missing volume checks to the code.

Fixes #474

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -557,7 +557,7 @@ func validatePVCAndAppBinding(pvcPath, appPath string, f *framework.Framework) {
 	}
 }
 
-func getRBDImageSpec(pvcNamespace, pvcName string, f *framework.Framework) (string, error) {
+func getImageIDFromPVC(pvcNamespace, pvcName string, f *framework.Framework) (string, error) {
 	c := f.ClientSet.CoreV1()
 	pvc, err := c.PersistentVolumeClaims(pvcNamespace).Get(pvcName, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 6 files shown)


ShyamsundarR:
Failed run with just the test case added: https://travis-ci.org/ceph/ceph-csi/builds/640730478 (for reference)

e2e/utils.go
```diff
@@ -574,6 +574,23 @@ func getRBDImageSpec(pvcNamespace, pvcName string, f *framework.Framework) (stri
 	return fmt.Sprintf("replicapool/csi-vol-%s", imageID), nil
 }
 
+func getCephFSVolumeName(pvcNamespace, pvcName string, f *framework.Framework) (string, error) {
```
ShyamsundarR: Function should be rewritten to re-use common parts of the above function (getRBDImageSpec). Will do in a subsequent revision.

pkg/cephfs/volumeoptions.go
```diff
@@ -245,13 +245,13 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 		}
 	}
 
+	volOptions.ProvisionVolume = true
+
 	volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
 	if err != nil {
-		return nil, nil, err
+		return &volOptions, &vid, err
```
nixpanic: What is the reason this is needed? I can not see it being used in this PR.

ShyamsundarR: If image is not found, we still need the options and vid returned to cleanup the journal. For reference the rbd return is similar, but we pass it the structure and refer to the same.

This is needed in DeleteVolume, when on receiving volume not found errors, we still need the contents of the passed returns to delete the journal and OMaps that are maintained.

humblec:
@nixpanic any further comments ? this patch looks good to me.

pkg/cephfs/volumeoptions.go
nixpanic: Ok, got it. Thanks!

PR #809, opened 2020-01-23
Add _netdev as default mount options for both cepfs and rbd

### LINKED ISSUE #756: Pass "_netdev" mount option by default for RBD/CephFS mounts
As discussed in kubernetes [1], for network devices, addition of the `_netdev` ensures that the network is not stopped prior to unmounting the network devices.

This issue is opened to discuss and fix, if we should automatically add `_netdev` mount time option, and not require users to add these to the mount options in the storage class.

NOTE: This assumes passing the `_netdev` option from the nodeplugin container to the host, is fixed or works

[1] Kubernetes references:
  - Patch adding `_netdev` handling during bind mounts: `https://github.com/kubernetes/kubernetes/pull/68626`
  - Current upstream issue on `_netdev` not appearing on the host mount options: `https://github.com/kubernetes/kubernetes/issues/86194`

Madhu-1:
Add _netdev as default mount options in plugins. This value will be added at both node stage and node publish for RBD, NBD and ceph kernel client.
  As cephfs fuse doesn't support this value, this is added only during the node publish.

Add run host path to daemonset pods. `/run/mount` need to be shared between the host and csi-plugin containers for `/run/mount/utab` this is required to ensure that the network is not stopped prior to unmounting the network devices.

Resolves #756

### DIFF
--- a/pkg/cephfs/volumemounter.go
+++ b/pkg/cephfs/volumemounter.go
@@ -36,6 +36,7 @@ import (
 const (
 	volumeMounterFuse   = "fuse"
 	volumeMounterKernel = "kernel"
+	netDev              = "_netdev"
 )
 
 var (
@@ -229,6 +230,7 @@ func mountFuse(ctx context.Context, mountPoint string, cr *util.Credentials, vol
 	if volOptions.FuseMountOptions != "" {
 		args = append(args, ","+volOptions.FuseMountOptions)
 	}
+
 	if volOptions.FsName != "" {
 		args = append(args, "--client_mds_namespace="+volOptions.FsName)
 	}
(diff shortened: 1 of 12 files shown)


Madhu-1:
E2E is failing for LUKS encrypted PVC. will check  can we pass `_netdev` for mounting encrypted device of not

Madhu-1:
@humblec @ShyamsundarR @nixpanic @dillaman @joscollin PTAL

pkg/cephfs/volumemounter.go
```diff
@@ -288,6 +290,11 @@ func mountKernel(ctx context.Context, mountPoint string, cr *util.Credentials, v
 	if volOptions.KernelMountOptions != "" {
 		optionsStr += fmt.Sprintf(",%s", volOptions.KernelMountOptions)
 	}
+
+	if !strings.Contains(volOptions.KernelMountOptions, netDev) {
+		optionsStr += fmt.Sprintf(",%s", "_netdev")
```
dillaman: Nit: why not re-use your constant here since you tested against it in the previous line?

Madhu-1: done

Madhu-1:
> Wanted to check why `/run` was mapped yesterday, nice that it is already changed to `/run/mount` to reduce what we need mapped to the nodeplugin.

Mounting '/run' cause more issues related to udev, '/run/mount'  what we need for netdev

humblec:
LGTM. Approved.. Thanks.

PR #813, opened 2020-01-29
[Fix rbd chart]Repair typo in host-mount and host-dev definition. It was swithced be…

kuja53:

Repair typo in host-mount and host-dev definition. It was swithced between each other and actually there is not possible to deploy nodeplugin-daemonset via helm chart. I was test it actually in my own deployment of ceph-csi-rbd.

Do you have any questions?
No

Yes, this will be backward compatible.

No

There is no context.


kuja53:
No problem. I should help with charts in the future.

PR #815, opened 2020-01-31
Handle Delete operation if pool not found

### LINKED ISSUE #781: Failed to Delete PVC if rbd pool is already deleted
# Describe the bug #

currently, we are checking the pool details and also removing the keys reserved in the rados omap, we need to make the PVC removal successful even if the pool is deleted
# Steps to reproduce #

* create PVC
* Delete rbd pool in the backend
* Delete PVC

# Expected behavior: #

PVC removal should be successful even if the pool is deleted in the backend

Madhu-1:
If the backend rbd or cephfs pool is already deleted we need to return success to the  DeleteVolume RPC call to make it idempotent.

Fixes #781


humblec:
@nixpanic @ShyamsundarR ptal.

humblec:
rebase required.

pkg/cephfs/controllerserver.go
```diff
@@ -235,6 +235,7 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		// or partially complete (subvolume and imageOMap are garbage collected already), hence
 		// return success as deletion is complete
 		if _, ok := err.(util.ErrKeyNotFound); ok {
+			klog.Warningf(util.Log(ctx, "Failed to volume options for %s: %v"), string(volID), err)
```
ShyamsundarR: (nit) error message needs some massaging "failed to find backing volume %s for volume ID %v" (same in the rbd controllerserver file as well)

(nit) here and in some other places, remove capitalization of leading message string

pkg/util/errors.go
```diff
@@ -16,14 +16,14 @@ limitations under the License.
 
 package util
 
-// ErrKeyNotFound is returned when requested key in omap is not found
+// ErrKeyNotFound is returned when requested key is not found
```
ShyamsundarR: I am not comfortable in overloading this error for missing pools. I would prefer a separate error code, although dealing with an additional error code in the callers is more code, it keeps the error codes cleaner.

pkg/cephfs/controllerserver.go
Madhu-1: This is a log, why to remove capitalization?

Madhu-1: I checked the current codebase, in most of the cases we are logging as small case but it's not required always see https://github.com/golang/go/wiki/CodeReviewComments#error-strings

ShyamsundarR: Technically the log message is also prepended with a "Req-ID: XXX : <message>" so capitalization rules as per the error message link applies (but can be argued otherwise as well). It also makes sense to follow a single style across the code, which I assumed was not capitalization. Anyway, I am not spending more time on this nit...

Madhu-1:
CI failure will go into it.

pkg/cephfs/controllerserver.go
Madhu-1: Yes we  need to follow single code side in the codebase, will revisit the codebase to check it and will send a PR

Madhu-1:
@ShyamsundarR CI passed PTAL

humblec:
lgtm. Thanks !

PR #816, opened 2020-02-04
Add topology support to ceph-csi

ShyamsundarR:
This PR adds Topology based volume provisioning.

The details of the design as implemented is detailed in https://github.com/ceph/ceph-csi/pull/760


ShyamsundarR:
Items to close beyond current code are as follows,
- Should topology be left ON by default in the provisioner sidecar?
- Should we remove topology from `volume_context` in CreateVolumeResponse
- Address 2 TODOs as follows,
  - rbd-journal.go // TODO check if need any undo operation here, or ErrVolNameConflict
  - rbd-util.go // TODO: If pool is not found we may leak the image (as DeleteVolume will return success)
- CephFS major TODO, as there is no subvolume command to fetch subvolume attributes
- Testing:
    - Topology.go needs unit test file
    - Update/add e2e as appropriate
- Manifests:
    - Helm charts update
    - Update cephfs StorageClass (later)
- Update tracevol.py
- Write documentation

humblec:
@ShyamsundarR are you planning to rebase it after initial review ?

Madhu-1:
@ShyamsundarR will review this in the next two days

ShyamsundarR:
> @ShyamsundarR are you planning to rebase it after initial review ?

Yes, or sooner if I start on the TODOs before the reviews.

I am held up on some RBD mirroring testing for now, so will pick this one up a post the same.

cmd/cephcsi.go
```diff
@@ -166,6 +168,13 @@ func main() {
 		}
 	}
 
+	// read and formulate conf.Domains from passed in domain labels
+	// TODO: decide if this will be from an init container and move as a separate image or instance
```
humblec: Whats the advantage we are getting it if its part of init Container ?

cmd/cephcsi.go
```diff
@@ -166,6 +168,13 @@ func main() {
...
+	conf.Domains, err = util.GetDomainsFromLabels(conf.DomainLabels, conf.NodeID)
```
humblec: Also how  are we going to deal with dynamically changed domain lables?

pkg/util/topology.go
```diff
@@ -0,0 +1,240 @@
...
+			return nil, fmt.Errorf("duplicate label (%s) found in domain labels", label)
+		}
+
+		labelsIn[label] = true
+		labelCount++
+	}
+
+	// Create a k8s client to read node labels
+	client = NewK8sClient()
+	node, err := client.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
```
humblec: @ShyamsundarR this looks to be problematic to me. ooc, have we tested this PR in a setup?  reading `node` object need special permissions , Isnt it ?

pkg/util/voljournal.go
```diff
@@ -323,8 +386,15 @@ Return values:
 	- string: Contains the UUID that was reserved for the passed in reqName
 	- error: non-nil in case of any errors
 */
-func (cj *CSIJournal) ReserveName(ctx context.Context, monitors string, cr *Credentials, pool, reqName, parentName, encryptionKmsConfig string) (string, error) {
-	var snapSource bool
+func (cj *CSIJournal) ReserveName(ctx context.Context, monitors string, cr *Credentials,
```
humblec: in one way, the `pool` parameter now has more seggregation like `imagePool`, `journalPool`, `csiJounalPool` , imo, its better to document these concepts in source code comments for further use.

pkg/util/voljournal.go
```diff
@@ -26,6 +28,8 @@ import (
 	"k8s.io/klog"
 )
 
+const uuidEncodedLength = 36
```
humblec: @ShyamsundarR anything special about the length of `36`. Just trying to understand..

humblec:
@ShyamsundarR few review comments from my side.  One of the concern part is, how are we going to deal with dynamically changing node lables and how we can adopt the change. Have we thought about it ?

humblec:
`Should we remove topology from volume_context in CreateVolumeResponse`

IMO, we should not remove `tpology` from the context field in `CreateVolumeResponse` . It may be required later and no harm in keeping it.

humblec:
@ShyamsundarR  as we have decided to not support `allowedTopologies`, do we have to do anything here if we get that from CO ?
Additional question I have here is , how do we seperate or use the `domainLables` as specific to ceph-csi. If we read it from node objects , we are reading global domain labels and not something specific to ceph-csi and using it. Am I missing something here ?

cmd/cephcsi.go
ShyamsundarR: Using an init container keeps the ceph-csi code independent of k8s calls, and retains the CO agnostic nature of the same.

cmd/cephcsi.go
ShyamsundarR: domain labels are set for a node, which indicate its relative location within the defined cluster domains (say rack/room/DC etc.). These are not going to change dynamically, unless the node instance is moved physically (or virtually) to another domain, at which point, I would assume, the node is restarted with the new labels, allowing us to read the new values and advertise the same.

pkg/util/topology.go
ShyamsundarR: There are a few ways to deal with this, for example to pass in the domain label values to the nodeplugins or create subset deployments per lowest domain with static labels. This is discussed [here](https://github.com/ceph/ceph-csi/pull/760/files#r360098011)

The manner of handling reading node labels seems the best when using daemon sets, as we need a nodeplugin running on all nodes. This does require the required roles and permissions.

Cloud providers reach into their API endpoints ([example](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/7c1a7afee2c79bdfe3f934ae8c4e839addc55720/pkg/cloud/metadata.go#L81-L106)) to get the node information, as these are used to both set the node labels (k8s) and also return CSI node domains. Further, they have access to the required credentials to make the said API calls. We lack such abilities, as we do not decide on the node labels. as these are setup by the cluster admin or the cloud provider.

In the face of the above limitations requiring additional permissions is required.

pkg/util/voljournal.go
ShyamsundarR: Agree, will deal with it when amending the patch for further changes.

pkg/util/voljournal.go
ShyamsundarR: This is the string representation of a UUID which is a constant length of 36 bytes ([reference](https://github.com/ceph/ceph-csi/blob/ca8dd2d8f28b6d7747330dc11b77c622ab76070f/vendor/github.com/pborman/uuid/uuid.go#L24-L28))

Unfortunately there is no constant in the uuid package to import and use, hence defined here. This needs a comment though to state it is the UUID string representation length. Will add it in the next revision to the patch set.

ShyamsundarR:
> > Should we remove topology from volume_context in CreateVolumeResponse>
> 
> IMO, we should not remove `topology` from the context field in `CreateVolumeResponse` . It may be required later and no harm in keeping it.

Agree in general, I was worried about the volume_context bloat at present, but it is not reached a critical enough size to start reducing parameters yet from it. But, something to keep in mind.

ShyamsundarR:
> @ShyamsundarR as we have decided to not support `allowedTopologies`, do we have to do anything here if we get that from CO ?

The proposal [states](https://github.com/ceph/ceph-csi/pull/760/files#diff-133c3c07046f779e3502b6b5f3f1236eR74-R80) we do not support this with the current effort. There are no checks from the past or in this patch to disallow it, it will just not be parsed and used.

However, as CSI specification for CreateVolume passes only parameters for consideration, we do not get the `allowedTopologies` parameter passed in the raw to the csi-plugin. Instead the external-provisioner side car folds this into the request [like so](https://github.com/kubernetes-csi/external-provisioner/blob/efcaee79e47446e38910a3c1a024824387fcf235/pkg/controller/controller.go#L534-L547).

Essentially if we test this out, there is no further support needed from the ceph-csi implementation to support `allowedTopologies`.

> Additional question I have here is , how do we seperate or use the `domainLables` as specific to ceph-csi. If we read it from node objects , we are reading global domain labels and not something specific to ceph-csi and using it. Am I missing something here ?

`domainLabels` for ceph-csi are generated in the function [fillNodesTopology](https://github.com/ceph/ceph-csi/pull/816/files#diff-68db63fd9a7297eacb6129328d3a6ce5R57-R81). The domain labels hence are of the form `topology.<driver-name>/<domain-label>`. They carry the "value" that is the same as the the one for the node (as the node belongs to the specified domain).

The function [GetDomainsFromLabels](https://github.com/ceph/ceph-csi/pull/816/files#diff-10487ffe93db35abe03d787f0126993dR36-R40) does the required node specific prefix stripping, to provide "values" and corresponding `<domain-label>` values that is used in fillNodesTopology.

As a result, we are not mirroring the node labels, we are creating our own. Further, we need to use the nodes domain label values as we do not determine the same. Which domain labels of the node, need to be reflected by the ceph-CSI driver is controlled by the CLI args passed in with `domainlabels`

humblec:
@ShyamsundarR Can you revisit this PR ?

pkg/util/voljournal.go
ShyamsundarR: Addressed in the latest update.

pkg/util/voljournal.go
ShyamsundarR: Addressed in the latest update.

cmd/cephcsi.go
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/util/topology.go
```diff
@@ -0,0 +1,240 @@
...
+		if _, ok := labelsIn[label]; ok {
+			return nil, fmt.Errorf("duplicate label (%s) found in domain labels", label)
+		}
+
+		labelsIn[label] = true
+		labelCount++
+	}
+
+	// Create a k8s client to read node labels
+	client = NewK8sClient()
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
```diff
@@ -66,6 +66,12 @@ spec:
             - "--metricsport=8090"
             - "--metricspath=/metrics"
             - "--enablegrpcmetrics=false"
+            # If topology based provisioning is desired, configure required
+            # kubernetes node labels representing the nodes topology domain
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/cephfs/volumeoptions.go
```diff
@@ -175,6 +182,18 @@ func newVolumeOptions(ctx context.Context, requestName string, volOptions, secre
...
+	// store topology information from the request
+	opts.TopologyPools, opts.TopologyRequirement, err = util.GetTopologyFromRequest(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: we need an API to fetch subvolume attributes (size/datapool and others), based
+	// on which we can evaluate which topology this belongs to.
+	if opts.TopologyPools != nil {
+		return nil, fmt.Errorf("topology based provisioning is not supported for CephFS backed volumes")
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/util/topology.go
```diff
@@ -0,0 +1,240 @@
...
+	// check if there are any accessibility requirements in the request
+	accessibilityRequirements := req.GetAccessibilityRequirements()
+	if accessibilityRequirements == nil {
+		return nil, nil, nil
+	}
+
+	// extract topology based pools configuration
+	err := json.Unmarshal([]byte(strings.Replace(topologyPoolsStr, "\n", " ", -1)), &topologyPools)
+	if err != nil {
+		klog.Infof(Log(ctx, "failed to parse JSON encoded topology constrained pools parameter (%s): %v"), topologyPoolsStr, err)
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/csi-common/nodeserver-default.go
```diff
@@ -53,12 +54,47 @@ func (ns *DefaultNodeServer) NodeExpandVolume(ctx context.Context, req *csi.Node
...
+	}
+
+	return topology, nil
+}
+
 // NodeGetInfo returns node ID
 func (ns *DefaultNodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (*csi.NodeGetInfoResponse, error) {
 	klog.V(5).Infof(util.Log(ctx, "Using default NodeGetInfo"))
 
+	topology, err := ns.fillNodesTopology()
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/csi-common/nodeserver-default.go
```diff
@@ -53,12 +54,47 @@ func (ns *DefaultNodeServer) NodeExpandVolume(ctx context.Context, req *csi.Node
...
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to update nodes topology, error : %v", err)
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

Function to derive topology from domain labels is merged, and called early on to detect any errors and reused in the  NodeGetInfo function.

cmd/cephcsi.go
```diff
@@ -58,6 +58,8 @@ func init() {
 	flag.IntVar(&conf.PidLimit, "pidlimit", 0, "the PID limit to configure through cgroups")
 	flag.BoolVar(&conf.IsControllerServer, "controllerserver", false, "start cephcsi controller server")
 	flag.BoolVar(&conf.IsNodeServer, "nodeserver", false, "start cephcsi node server")
+	flag.StringVar(&conf.DomainLabels, "domainlabels", "", "list of kubernetes node labels, that determines the topology"+
```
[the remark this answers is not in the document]
ShyamsundarR: Should we push this till we actually get another orchestrator that we work with?

pkg/util/topology.go
```diff
@@ -0,0 +1,240 @@
+/*
```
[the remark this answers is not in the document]
ShyamsundarR: I do not have an opinion either way, but why do we need a separate folder?

deploy/rbd/kubernetes/csi-nodeplugin-rbac.yaml
```diff
@@ -3,3 +3,37 @@ apiVersion: v1
...
+    verbs: ["get"]
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: rbd-csi-nodeplugin
+subjects:
+  - kind: ServiceAccount
+    name: rbd-csi-nodeplugin
+    namespace: rook-ceph
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

deploy/cephfs/kubernetes/csi-nodeplugin-rbac.yaml
```diff
@@ -3,3 +3,37 @@ apiVersion: v1
...
+    verbs: ["get"]
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cephfs-csi-nodeplugin
+subjects:
+  - kind: ServiceAccount
+    name: cephfs-csi-nodeplugin
+    namespace: rook-ceph
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

deploy/rbd/kubernetes/csi-rbdplugin.yaml
```diff
@@ -66,6 +66,12 @@ spec:
             - "--metricsport=8090"
             - "--metricspath=/metrics"
             - "--enablegrpcmetrics=false"
+            # If topology based provisioning is desired, configure required
+            # kubernetes node labels representing the nodes topology domain
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/util/topology.go
```diff
@@ -0,0 +1,240 @@
...
+	)
+
+	// check if parameters have pool configuration pertaining to topology
+	topologyPoolsStr := req.GetParameters()["topologyConstrainedPools"]
+	if topologyPoolsStr == "" {
+		return nil, nil, nil
+	}
+
+	// check if there are any accessibility requirements in the request
+	accessibilityRequirements := req.GetAccessibilityRequirements()
```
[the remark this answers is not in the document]
ShyamsundarR: The AccessibilityRequirements is filled in by the CO, based on where the pod can be scheduled and by collecting the CSI node topology that is advertised by NodeGetInfo. IOW, user/app need not mention anything more than the topology constrained pools in the StorageClass.

examples/rbd/storageclass.yaml
```diff
@@ -48,6 +52,25 @@ parameters:
    # a unique ID matching KMS ConfigMap. The ID is only used for correlation to
    # config map entry.
    # encryptionKMSID: <kms-config-id>
+
+   # Add topology constrained pools configuration, if topology based pools
+   # are setup, and topology constrained provisioning is required.
+   # For further information read TODO<doc>
+   # topologyConstrainedPools: |
+   #   [{"poolName":"pool0",
```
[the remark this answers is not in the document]
ShyamsundarR: This will start to get ugly, because in the topology constrained pools section we will now add one more parameter.

I forget what the limitation is to directly create the image OMap and the image on an EC pool, or is it just not supported? Tagging @dillaman as well for details.

If it is an RBD behavior, then we have no go but to add it to this section as well. But, it gets trickier than just that, we now need 2 pools per topology domain, one replicated for the OMap and image, and one EC pool for the data.

ShyamsundarR:
> we may need to add below
> 
>     * A Readme doc topology-based provisioning in CSI (examples)
> 
>     * An E2E just to cover basic scenario (once we Fix #857 we can extend it to multi-node tests)

Tracked in the TODO list present in this [comment](https://github.com/ceph/ceph-csi/pull/816#issuecomment-586775091)

examples/rbd/storageclass.yaml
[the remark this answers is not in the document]
Madhu-1: Need clarification on this one.

pkg/cephfs/fsjournal.go
```diff
@@ -46,52 +46,44 @@ request name lock, and hence any stale omaps are leftovers from incomplete trans
...
+				volOptions.MetadataPool, vid.FsSubvolName, volOptions.RequestName)
 			return nil, err
 		}
 		return nil, err
 	}
+
+	// check if topology constraints match what is found
+	// TODO: we need an API to fetch subvolume attributes (size/datapool and others), based
+	// on which we can evaluate which topology this belongs to.
+	// TODO: CephFS topology support is postponed till we get the same
```
Madhu-1, line 81: is this true? because I see template  changes for cephfs also

pkg/rbd/driver.go
```diff
@@ -133,8 +133,16 @@ func (r *Driver) Run(conf *util.Config, cachePersister util.CachePersister) {
...
+	topology := map[string]string{}
 	if conf.IsNodeServer {
-		r.ns, err = NewNodeServer(r.cd, conf.Vtype)
+		topology, err = util.GetTopologyFromDomainLabels(conf.DomainLabels, conf.NodeID, conf.DriverName)
+		if err != nil {
+			klog.Fatalln(err)
+		}
+	}
+
+	if conf.IsNodeServer {
```
Madhu-1, line 144: we can remove this change and move below code to line 142

pkg/rbd/driver.go
```diff
@@ -144,7 +152,7 @@ func (r *Driver) Run(conf *util.Config, cachePersister util.CachePersister) {
 		r.cs = NewControllerServer(r.cd, cachePersister)
 	}
 	if !conf.IsControllerServer && !conf.IsNodeServer {
-		r.ns, err = NewNodeServer(r.cd, conf.Vtype)
+		r.ns, err = NewNodeServer(r.cd, conf.Vtype, topology)
```
Madhu-1, line 155: the topology will be empty here isn't it?

scripts/rook-rbd-pool.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash -e
+
+ROOK_VERSION=${ROOK_VERSION:-"v1.1.7"}
+ROOK_DEPLOY_TIMEOUT=${ROOK_DEPLOY_TIMEOUT:-300}
+ROOK_URL="https://raw.githubusercontent.com/rook/rook/${ROOK_VERSION}/cluster/examples/kubernetes/ceph"
+ROOK_RBD_POOL=${ROOK_RBD_POOL:-"newrbdpool"}
+
+function create_pool() {
```
Madhu-1, line 8: move this code changes to rook.sh file

scripts/rook-rbd-pool.sh
```diff
@@ -0,0 +1,54 @@
...
+		echo "[Timeout] Failed to get RBD pool $ROOK_RBD_POOL stats"
+		exit 1
+	fi
+	echo ""
+}
+
+function delete_pool() {
+	curl -o newpool.yaml "${ROOK_URL}/pool-test.yaml"
+    sed -i "s/replicapool/$ROOK_RBD_POOL/g" newpool.yaml
+	kubectl delete -f "./newpool.yaml"
```
Madhu-1, line 37: we also need to delete newpool.yaml file better to save it in tmp folder

charts/ceph-csi-cephfs/templates/provisioner-deployment.yaml
```diff
@@ -37,6 +37,9 @@ spec:
             - "--enable-leader-election=true"
             - "--leader-election-type=leases"
             - "--retry-interval-start=500ms"
+{{- if .Values.topology.enabled }}
+            - "--feature-gates=Topology=true"
```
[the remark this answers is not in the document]
ShyamsundarR: TODO: Documentation should cover versions and support information, will keep that in mind.

The support status is as [here](https://kubernetes-csi.github.io/docs/topology.html#status), IOW this needs kubelet to be v1.14 or greater, and the external-provisioner side car to be 1.1 or greater. Further, there is the kubernetes feature gate for versions lower than GA as covered [here](https://kubernetes-csi.github.io/docs/topology.html#beta)

charts/ceph-csi-rbd/templates/nodeplugin-clusterrole.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.rbac.create -}}
+{{- if .Values.topology.enabled }}
```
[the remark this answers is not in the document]
ShyamsundarR: True, when we come to that point we need to change the if conditions appropriately, at present as you/we cleaned up the RBAC, the entire file is either needed or not. So, I used an overarching if block here.

scripts/rook-rbd-pool.sh
ShyamsundarR: `tmp` folder save did not work for me, due to (possible) minikube settings, hence I decided to just dump it in place. If this is fine, we can look at this later.

pkg/cephfs/fsjournal.go
ShyamsundarR: Everything else is in place, **except** the ability to inspect an subvolumes data pool. Hence, leaving it in place and will cover support/feature situation for CephFS in documentation.

charts/ceph-csi-cephfs/templates/nodeplugin-clusterrole.yaml
```diff
@@ -0,0 +1,19 @@
...
+    component: {{ .Values.nodeplugin.name }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+aggregationRule:
+  clusterRoleSelectors:
+    - matchLabels:
+        rbac.cephfs.csi.ceph.com/aggregate-to-{{ include "ceph-csi-cephfs.nodeplugin.fullname" . }}: "true"
+rules: []
+{{- end }}
+{{- end -}}
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

e2e/utils.go
```diff
@@ -1119,3 +1145,120 @@ func checkMountOptions(pvcPath, appPath string, f *framework.Framework, mountFla
...
+
+	return nil
+}
+
+func checkPVSelectorValuesForPVC(f *framework.Framework, pvc *v1.PersistentVolumeClaim) {
+	pv, err := e2epv.GetBoundPV(f.ClientSet, pvc)
+	if err != nil {
+		Fail(err.Error())
+	}
+	for _, expression := range pv.Spec.NodeAffinity.Required.NodeSelectorTerms[0].MatchExpressions {
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/rbd/driver.go
ShyamsundarR: Addressed in the latest update.

pkg/rbd/driver.go
ShyamsundarR: Addressed in the latest update. Good catch!

scripts/rook-rbd-pool.sh
ShyamsundarR: Addressed in the latest update.

scripts/rook.sh
```diff
@@ -115,11 +150,19 @@ deploy)
...
+	;;
 *)
 	echo " $0 [command]
 Available Commands:
-  deploy         Deploy a rook
-  teardown       Teardown a rook
+  deploy             Deploy a rook
+  teardown           Teardown a rook
+  create-block-pool  Create a rook block pool
+  delete-block-pool  Delete a rook block pool
```
Madhu-1, line 165: are we using delete-block-pool any where?

scripts/rook.sh
```diff
@@ -37,6 +38,40 @@ function teardown_rook() {
...
+		exit 1
+	fi
+	echo ""
+}
+
+function delete_block_pool() {
+	curl -o newpool.yaml "${ROOK_URL}/pool-test.yaml"
+    sed -i "s/replicapool/$ROOK_BLOCK_POOL_NAME/g" newpool.yaml
+	kubectl delete -f "./newpool.yaml"
+	rm -f "./newpool.yaml"
```
Madhu-1, line 72: do we need to check in toolbox that pool is deleted?

scripts/rook.sh
```diff
@@ -37,6 +38,40 @@ function teardown_rook() {
...
+			break
+		fi
+	done
+
+	if [ "$retry" -gt "$ROOK_DEPLOY_TIMEOUT" ]; then
+		echo "[Timeout] Failed to get RBD pool $ROOK_BLOCK_POOL_NAME stats"
+		exit 1
+	fi
+	echo ""
+}
```
Madhu-1, line 66: can we avoid the code duplication here currently we have function to check the pool stats. i think we can have one function which accepts input name as pool name and check in toolbox pod https://github.com/ceph/ceph-csi/blob/master/scripts/rook.sh#L88-L109

Madhu-1:
@humblec @nixpanic PTAL

scripts/rook.sh
ShyamsundarR: no, provided for completeness.

scripts/rook.sh
ShyamsundarR: This is provided for completeness, so not adding any checks as such (like create pool)

scripts/rook.sh
ShyamsundarR: Addressed in the latest update.

cmd/cephcsi.go
```diff
@@ -58,6 +58,8 @@ func init() {
...
+		" domain the node belongs to, separated by ';'")
```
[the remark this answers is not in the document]
phlogistonjohn: +1 to that and pflag already has support for doing that kind of thing baked in:
https://godoc.org/github.com/spf13/pflag#StringSliceVar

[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

I did not use `StringSliceVar` from spf13/pflag as suggested by @phlogistonjohn  as the current flag package in use, seems to be the one that is default with go (although we do have the mentioned package in the vendor directory). Reason being, I did not want to add a new dependency with this commit.

charts/ceph-csi-cephfs/values.yaml
```diff
@@ -165,6 +165,17 @@ provisioner:
...
+topology:
+  # Specifies whether topology based provisioning support should
+  # be exposed by CSI
+  enabled: false
+  # domainLabels define which node labels to use as domains
+  # for CSI nodeplugins to advertise their domains
+  # Each node label is ; separated in the value below
+  # NOTE: the value here serves as an example and needs to be
+  # updated with node labels that define domains of interest
+  domainLabels: failure-domain/region;failure-domain/zone
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

scripts/rook.sh
```diff
@@ -37,6 +38,22 @@ function teardown_rook() {
...
+    sed -i "s/replicapool/$ROOK_BLOCK_POOL_NAME/g" newpool.yaml
+	kubectl create -f "./newpool.yaml"
+	rm -f "./newpool.yaml"
+
+	check_rbd_stat "$ROOK_BLOCK_POOL_NAME"
+}
+
+function delete_block_pool() {
+	curl -o newpool.yaml "${ROOK_URL}/pool-test.yaml"
+    sed -i "s/replicapool/$ROOK_BLOCK_POOL_NAME/g" newpool.yaml
```
nixpanic, line 52: indention?

scripts/rook.sh
```diff
@@ -37,6 +38,22 @@ function teardown_rook() {
 	kubectl delete -f "${ROOK_URL}/common.yaml"
 }
 
+function create_block_pool() {
+    curl -o newpool.yaml "${ROOK_URL}/pool-test.yaml"
+    sed -i "s/replicapool/$ROOK_BLOCK_POOL_NAME/g" newpool.yaml
```
nixpanic, line 43: indention?

cmd/cephcsi.go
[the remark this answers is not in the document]
phlogistonjohn: Oh, I was unaware. Sorry for the noise. I just assumed everyone was using pflag now. :-)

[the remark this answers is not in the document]
ShyamsundarR: @phlogistonjohn Suggestion was useful, I almost did 3 lines of change, before hunting for the API in the **current** flag package, so no worries. We should move to pflag is that is more appropriate in the future.

pkg/rbd/rbd_util.go
```diff
@@ -394,17 +413,29 @@ func genVolFromVolID(ctx context.Context, rbdVol *rbdVolume, volumeID string, cr
 	if err != nil {
 		return err
 	}
+	rbdVol.JournalPool = rbdVol.Pool
```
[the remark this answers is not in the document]
ShyamsundarR: Added a comment on what `DataPool` is, and where it is used (only during creates) in the `rbdVolume` struct.

When generating the volume from the VolID, we do not need to populate the `DataPool` as was the case prior to this change.

pkg/util/voljournal.go
ShyamsundarR: Tried the same in go playground, it is disallowed. I could have replaced the single instance of the use with len() as above, but I **do** want to remember the length as this 36, as that was what was used in the past. So if we dynamically use len of the UUID and that changes, assumptions in code will change. Further added a note to this effect.

pkg/util/voljournal.go
```diff
@@ -198,71 +219,114 @@ Return values:
...
+		var buf64 []byte
+		components := strings.Split(objUUIDAndPool, "/")
+		objUUID = components[1]
+		savedImagePoolIDStr := components[0]
+
+		buf64, err = hex.DecodeString(savedImagePoolIDStr)
+		if err != nil {
+			return nil, err
+		}
+		savedImagePoolID = int64(binary.BigEndian.Uint64(buf64))
```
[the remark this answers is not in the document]
ShyamsundarR: We store this as BigEndian at line 516 in the same file

pkg/util/voljournal.go
```diff
@@ -366,54 +455,78 @@ func (cj *CSIJournal) ReserveName(ctx context.Context, monitors string, cr *Cred
...
 			if errDefer != nil {
 				klog.Warningf(Log(ctx, "failed undoing reservation of volume: %s (%v)"), reqName, errDefer)
 			}
 		}
 	}()
 
-	// Update UUID directory to store CSI request name and image name
-	err = SetOMapKeyValue(ctx, monitors, cr, pool, cj.namespace, cj.cephUUIDDirectoryPrefix+volUUID,
+	// Update UUID directory to store CSI request name
+	err = SetOMapKeyValue(ctx, monitors, cr, imagePool, cj.namespace, cj.cephUUIDDirectoryPrefix+volUUID,
```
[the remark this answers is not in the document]
ShyamsundarR: The image journal, or OMap directory is stored with the image, which goes to the `imagePool`.

This was due to the reasons [here](https://github.com/ceph/ceph-csi/pull/760/files#r368085409). Added a comment in the code regarding the same.

scripts/minikube.sh
```diff
@@ -118,6 +121,16 @@ deploy-rook)
     DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
     "$DIR"/rook.sh deploy
     ;;
+create-block-pool)
```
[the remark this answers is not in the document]
ShyamsundarR: This is a glue between `rook.sh` and `minikube.sh`, just so that the travis scripts only invoke `minikube.sh`. I would assume their presence in rook.sh helps running them on non-minikube environments.

pkg/util/topology.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+		return nil, fmt.Errorf("missing domain labels (%s) on node (%s)", missingLabels, nodeName)
+	}
+
+	klog.Infof("list of domains processed : %+v", domainMap)
+
+	topology := make(map[string]string)
+	for domain, value := range domainMap {
+		topology[topologyPrefix+"/"+domain] = value
+		// TODO: when implementing domain takeover/giveback, enable a domain value that can remain pinned to the node
```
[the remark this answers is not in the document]
ShyamsundarR: Update the TODO list in the tracking [comment](https://github.com/ceph/ceph-csi/pull/816#issuecomment-586775091). I will open issues for these once the PR merges.

pkg/util/topology.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	_, _, topology, err := FindPoolAndTopology(&topologyPool, accessibilityRequirements)
+
+	return topology, err
+}
+
+// FindPoolAndTopology loops through passed in "topologyPools" and also related
+// accessibility requirements, to determine which pool matches the requirement.
+// The return variables are, poolname, topology map of matched requirement
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/cephfs/volumeoptions.go
```diff
@@ -176,6 +183,18 @@ func newVolumeOptions(ctx context.Context, requestName string, volOptions, secre
...
 	}
 
+	// store topology information from the request
+	opts.TopologyPools, opts.TopologyRequirement, err = util.GetTopologyFromRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: we need an API to fetch subvolume attributes (size/datapool and others), based
+	// on which we can evaluate which topology this belongs to.
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/rbd/rbd_util.go
```diff
@@ -70,24 +71,28 @@ type rbdVolume struct {
...
+	TopologyPools       *[]util.TopologyConstrainedPool
+	TopologyRequirement *csi.TopologyRequirement
+	Topology            map[string]string
+	RbdImageName        string
+	NamePrefix          string
+	VolID               string `json:"volID"`
+	Monitors            string `json:"monitors"`
+	JournalPool         string
+	Pool                string `json:"pool"`
+	DataPool            string
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/util/topology.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get node (%s) information : %v", nodeName, err)
+	}
+
+	return node.GetLabels(), nil
+}
+
+// GetTopologyFromDomainLabels returns the CSI topology map, determined from
+// the domain labels and their values from the CO system
+// Expects domainLabels in arg to be in the format "[prefix/]<name>,[prefix/]<name>;...",
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/util/topology.go
```diff
@@ -0,0 +1,257 @@
...
+		nameIdx := strings.IndexRune(key, keySeparator)
+		domain := key[nameIdx+1:]
+		domainMap[domain] = value
+		labelsIn[key] = false
+		found++
+	}
+
+	// Ensure all labels are found
+	if found != labelCount {
+		missingLabels := ""
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

pkg/util/voljournal.go
```diff
@@ -423,70 +536,89 @@ func (cj *CSIJournal) ReserveName(ctx context.Context, monitors string, cr *Cred
...
-*/
-func (cj *CSIJournal) GetObjectUUIDData(ctx context.Context, monitors string, cr *Credentials, pool, objectUUID string, snapSource bool) (string, string, string, string, error) {
-	var sourceName string
+// ImageAttributes contains all CSI stored image attributes, typically as OMap keys
+type ImageAttributes struct {
+	RequestName   string // Contains the request name for the passed in UUID
+	SourceName    string // Contains the parent image name for the passed in UUID, if it is a snapshot
+	ImageName     string // Contains the image or subvolume name for the passed in UUID
+	KmsID         string // Contains encryption KMS, if it is an encrypted image
+	JournalPoolID int64
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

scripts/install-helm.sh
```diff
@@ -130,6 +136,13 @@ install_cephcsi_helm_charts() {
     if [ -z "$NAMESPACE" ]; then
         NAMESPACE="default"
     fi
+
+    # label the nodes uniformly for domain information
+    for node in $(kubectl get node --no-headers | cut -f 1 -d ' '); do
```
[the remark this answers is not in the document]
ShyamsundarR: Addressed in the latest update.

scripts/rook.sh
ShyamsundarR: Addressed in the latest update.

scripts/rook.sh
ShyamsundarR: Addressed in the latest update.

e2e/utils.go
```diff
@@ -319,6 +319,12 @@ func createRBDStorageClass(c kubernetes.Interface, f *framework.Framework, param
 		sc.Parameters[k] = v
 	}
 	sc.Namespace = cephCSINamespace
+	for k, v := range scOptions {
+		if k == "volumeBindingMode" && v == "WaitForFirstConsumer" {
```
nixpanic, line 323: Are these strings not available as constants?

pkg/util/voljournal.go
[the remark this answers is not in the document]
nixpanic: Yes, I noticed that later too. Is there a reason for BigEndian, and not just a string?

nixpanic:
Looks pretty good and complete to me. Just a few more minor things and I'd give it a :+1:

e2e/utils.go
ShyamsundarR: I cannot find a reusable constant for `volumeBindingMode`.

For `WaitForFirstConsumer` there is a typed string constant (which is used just below this as `scv1.VolumeBindingWaitForFirstConsumer`), but this makes them non type compliant causing linter failures.

Also, this means we need to map for `scOptions` that this function takes, be a map with different keys and values?

If you have an alternative I can code that up, else I am out of options on this.

pkg/util/voljournal.go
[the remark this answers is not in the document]
ShyamsundarR: This is an pool ID that is an int64 type, so prefer to retain this encoded. The same encoding is used in the `volumeID` as well, when I wrote that up.

Madhu-1:
@nixpanic can you please approve this if all your comments are resolved?

e2e/utils.go
```diff
@@ -319,6 +319,12 @@ func createRBDStorageClass(c kubernetes.Interface, f *framework.Framework, param
...
+			value := scv1.VolumeBindingWaitForFirstConsumer
+			sc.VolumeBindingMode = &value
+		}
+	}
```
Madhu-1, line 327:
```suggestion
if scOptions["volumeBindingMode"]=="WaitForFirstConsumer"{
value := scv1.VolumeBindingWaitForFirstConsumer
sc.VolumeBindingMode = &value
}
```

deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
```diff
@@ -40,6 +40,7 @@ spec:
             - "--enable-leader-election=true"
             - "--leader-election-type=leases"
             - "--retry-interval-start=500ms"
+            - "--feature-gates=Topology=true"
```
humblec, line 43: @ShyamsundarR do we need to explicitly have a parameter here wih `true` value ?

docs/deploy-cephfs.md
```diff
@@ -44,22 +44,23 @@ that should be resolved in v14.2.3.
...
 | `--pidlimit`              | _0_                         | Configure the PID limit in cgroups. The container runtime can restrict the number of processes/tasks which can cause problems while provisioning (or deleting) a large number of volumes. A value of `-1` configures the limit to the maximum, `0` does not configure limits at all. |
-| `--metricsport`           | `8080`                      | TCP port for liveness metrics requests                                                                                                                                                                                                                                                  |
+| `--metricsport`           | `8080`                      | TCP port for liveness metrics requests                                                                                                                                                                                                                                               |
```
humblec, line 56: @ShyamsundarR whats the change here for `metricsport` and `enable...` ?  not able to figure out.

e2e/utils.go
humblec: declared `type VolumeBindingMode string` constants can be used here @ShyamsundarR instead of hard-coding these.

pkg/util/topology_test.go
```diff
@@ -0,0 +1,379 @@
...
+	// check for existing pool
+	topoSegment, err = MatchTopologyForPool(&validMultipleTopoPools, &validAccReq, pool1)
+	err = checkOutput(err, pool1, topoSegment)
+	if err != nil {
+		t.Errorf("expected success got: (%v)", err)
+	}
+}
+
+/*
+// TODO: To test GetTopologyFromDomainLabels we need it to accept a k8s client interface, to mock k8sGetNdeLabels output
```
humblec, line 355: @ShyamsundarR is this TODO planned for follow up PR ?

humblec:
Overal looks good to me. some minor comments from end

deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
ShyamsundarR: yes, see [this](https://kubernetes-csi.github.io/docs/topology.html#sidecar-deployment) document for the same.

docs/deploy-cephfs.md
ShyamsundarR: Merge issue, these need to be removed. Working on it now.

pkg/util/topology_test.go
ShyamsundarR: It is left in here if decided to take this up as an InitContainer.

e2e/utils.go
ShyamsundarR: As stated, I am using these constants just below, but due to type difference not able to cleanly use them here. I would assume this does not stop the PR from being merged, though. If it does, please let me know.

docs/deploy-cephfs.md
ShyamsundarR: Ah! ok, this was not a merge issue after all (which is a relief).

I have seen that whenever we modify YAML tables we are aligning these to the same width. The same was done here post addition of the new option. Although the new option did not change the table width, other added options that did not adjust for the table width (in their commits) were now auto-corrected. IOW, the trailing `|` is now auto-corrected to the right location, which is the change here.

Also, when reviewing this, noticed that the block `[rbd | cephfs]` in option --type was incorrectly aligned due to the table width correction, which is now reverted.

e2e/utils.go
ShyamsundarR: Updated in the latest rebase.

pkg/util/topology_test.go
ShyamsundarR: IOW, no followup PR planned for this.

humblec:
@nixpanic any final comments from your side ?

humblec:
LGTM.. Thanks ! @ShyamsundarR can you please do  rebase ?

pkg/util/topology.go
```diff
@@ -0,0 +1,255 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
```
nixpanic, line 19: Add `"context"`

pkg/util/topology.go
```diff
@@ -0,0 +1,255 @@
...
+)
+
+const (
+	keySeparator   rune   = '/'
+	labelSeparator string = ","
+)
+
+func k8sGetNodeLabels(nodeName string) (map[string]string, error) {
+	client := NewK8sClient()
+	node, err := client.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
```
nixpanic, line 36: With the upgrade to k8s 1.18.0 this needs an extra parameter, probably `context.TODO()`

ShyamsundarR:
Rebased to pull in the latest k8s 1.18 changes. Had to do the following,
- Added function getBoundPV as that is no longer available from the k8s framework package see [this](https://github.com/ceph/ceph-csi/pull/816/files#diff-0781e6445e81ef5ec3953f597be8dc1aR1311-R1316)
- Updated k8s calls to add the `context.TODO()` as appropriate (in e2e/utils.go and in topology.go)

Ran only build and static checks on my development machine. Will await e2e results for a final check.

humblec:
> Rebased to pull in the latest k8s 1.18 changes. Had to do the following,
> 
>     * Added function getBoundPV as that is no longer available from the k8s framework package see [this](https://github.com/ceph/ceph-csi/pull/816/files#diff-0781e6445e81ef5ec3953f597be8dc1aR1311-R1316)
> 
>     * Updated k8s calls to add the `context.TODO()` as appropriate (in e2e/utils.go and in topology.go)
> 
> 
> Ran only build and static checks on my development machine. Will await e2e results for a final check.

Thanks @ShyamsundarR . Just waiting for the e2e to pass !

humblec:
All the tests are green. LGTM .. Approved.!

PR #819, opened 2020-02-05
Add CEPH objectstore test on script/minikube.sh deploy-rook

sophalHong:
Signed-off-by: sophal <sophalHong@github.com>

When we execute ./scripts/minikube.sh deploy-rook, only CEPH cluster, FileSystem, and RBD BlockPool are created.  Why not CEPH object store?  Ceph object store is not implemented.  
This PR does:
1. Create cephObjectStore object into cluster
2. Check if cephObjectStore is successfully created
3. Delete cephObjectStore when execute ./scripts/minikube.sh teardown-rook
4. Add printing error message when timeout is over

Tested on environment:
MINIKUBE_VERSION="v1.5.2"
KUBE_VERSION="v1.15.0"
VM_DRIVER="virtualbox"

Do you have any questions?
No

No

No

No

## Future concerns ##

### DIFF
--- a/scripts/rook.sh
+++ b/scripts/rook.sh
@@ -11,6 +11,7 @@ function deploy_rook() {
 	kubectl create -f "${ROOK_URL}/toolbox.yaml"
 	kubectl create -f "${ROOK_URL}/filesystem-test.yaml"
 	kubectl create -f "${ROOK_URL}/pool-test.yaml"
+	kubectl create -f "${ROOK_URL}/object-test.yaml"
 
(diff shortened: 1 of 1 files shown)


Madhu-1:
@sophalHong sorry to say that ceph-csi doesn't support object store, I don't think this is useful

sophalHong:
> @sophalHong sorry to say that ceph-csi doesn't support object store, I don't think this is useful

Ah, Okay

PR #821, opened 2020-02-07
Add printing failed message when timeout

sophalHong:
Signed-off-by: sophal <sophalHong@github.com>

Base on the previous particular script implementation, when something does not work well, it will run again and again in for-loop until timeout is over. Once timeout is over, it will break the loop, and continue to the step without notifying that it is failed.
so this PR just added few lines of script code to check whether timeout is over and then printing failed message.

Do you have any questions?
No

No

No

No

## Future concerns ##

### DIFF
--- a/scripts/rook.sh
+++ b/scripts/rook.sh
@@ -51,6 +51,11 @@ function check_ceph_cluster_health() {
 			fi
 		fi
 	done
+
+	if [ "$retry" -gt "$ROOK_DEPLOY_TIMEOUT" ]; then
+		echo "[Timeout] CEPH cluster not in a healthy state (timeout)"
+		exit 1
+	fi
 	echo ""
 }
 
@@ -72,6 +77,11 @@ function check_mds_stat() {
 			fi
 		fi
 	done
+
+	if [ "$retry" -gt "$ROOK_DEPLOY_TIMEOUT" ]; then
+		echo "[Timeout] Failed to get ceph filesystem pods"
+		exit 1
+	fi
 	echo ""
 }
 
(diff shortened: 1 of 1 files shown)


scripts/rook.sh
```diff
@@ -51,6 +51,11 @@ function check_ceph_cluster_health() {
 			fi
 		fi
 	done
+
+	if [ "$retry" -gt "$ROOK_DEPLOY_TIMEOUT" ]; then
+		echo "[ERROR] Check CEPH Cluster state failed! (timeout)"
+		exit
```
Madhu-1: echo "[Timeout] CEPH cluster not  in a healthy state (timeout)"
exit 1

scripts/rook.sh
```diff
@@ -72,6 +77,10 @@ function check_mds_stat() {
 			fi
 		fi
 	done
+
+	if [ "$retry" -gt "$ROOK_DEPLOY_TIMEOUT" ]; then
+		echo "[ERROR] Check CEPH Filesystem state failed! (timeout)"
```
Madhu-1: echo "[Timeout]  Failed to get ceph filesystem pods"
exit 1

scripts/rook.sh
```diff
@@ -82,14 +91,18 @@ function check_rbd_stat() {
...
+		[[ "$TOOLBOX_POD_STATUS" != "Running" ]] && \
+			{ echo "Toolbox POD ($TOOLBOX_POD) status: [$TOOLBOX_POD_STATUS]"; continue; }
 
 		if kubectl exec -n rook-ceph "$TOOLBOX_POD" -it -- rbd pool stats "$RBD_POOL_NAME" &>/dev/null; then
 			echo "RBD ($RBD_POOL_NAME) is successfully created..."
 			break
 		fi
 	done
+
+	if [ "$retry" -gt "$ROOK_DEPLOY_TIMEOUT" ]; then
```
Madhu-1: echo "[Timeout] Failed to get RBD pool  stats"
exit 1

humblec:
LGTM. Thanks !

PR #822, opened 2020-02-07
Reduce encryption KMS configuration SC parameters

vasyl-purchel:
* moves KMS type from StorageClass into KMS configuration itself
 * updates omapval used to identify KMS to only it's ID without the type

why?

1. when using multiple KMS configurations (not currently supported)
automated parsing of kms configuration will be failing because some
entries in configs won't comply with the requested type
2. less options are needed in the StorageClass and less data used to
identify the KMS

Signed-off-by: Vasyl Purchel vasyl.purchel@workday.com
Signed-off-by: Andrea Baglioni andrea.baglioni@workday.com

### DIFF
--- a/charts/ceph-csi-rbd/values.yaml
+++ b/charts/ceph-csi-rbd/values.yaml
@@ -31,9 +31,15 @@ csiConfig: []
 # Ref: https://github.com/ceph/ceph-csi/blob/master/docs/deploy-rbd.md
 # Example:
 # encryptionKMSConfig:
-#   - encryptionKMSID: "<kms-id>"
-#     <kms-specific-configs>
-encryptionKMSConfig: []
+#   vault-unique-id-1:
+#     encryptionKMSType: vault
+#     vaultAddress: https://vault.example.com
+#     vaultAuthPath: /v1/auth/kubernetes/login
+#     vaultRole: csi-kubernetes
+#     vaultPassphraseRoot: /v1/secret
+#     vaultPassphrasePath: ceph-csi/
+#     vaultCAVerify: "false"
+encryptionKMSConfig: {}
 
 nodeplugin:
   name: nodeplugin
--- a/pkg/util/crypto.go
+++ b/pkg/util/crypto.go
@@ -19,7 +19,9 @@ package util
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"path"
 	"strings"
 
@@ -39,6 +41,10 @@ const (
 
 	// Encryption passphrase location in K8s secrets
 	encryptionPassphraseKey = "encryptionPassphrase"
+	kmsTypeKey              = "encryptionKMSType"
+
+	// Default KMS type
+	defaultKMSType = "default"
 
 	// kmsConfigPath is the location of the vault config file
 	kmsConfigPath = "/etc/ceph-csi-encryption-kms-config/config.json"
@@ -54,7 +60,7 @@ type EncryptionKMS interface {
 	GetPassphrase(key string) (string, error)
 	SavePassphrase(key, value string) error
 	DeletePassphrase(key string) error
-	KmsConfig() string
+	GetID() string
 }
 
 // MissingPassphrase is an error instructing to generate new passphrase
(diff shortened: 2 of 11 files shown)


charts/ceph-csi-rbd/values.yaml
```diff
@@ -31,9 +31,15 @@ csiConfig: []
 # Ref: https://github.com/ceph/ceph-csi/blob/master/docs/deploy-rbd.md
 # Example:
 # encryptionKMSConfig:
-#   - encryptionKMSID: "<kms-id>"
-#     <kms-specific-configs>
-encryptionKMSConfig: []
+#   vault-unique-id-1:
```
humblec: @vasyl-purchel can we add some dcoumentation around this "unique" id too ?

vasyl-purchel: it's already documented in deploy-rbd.md:
> To use Vault as KMS set `encryptionKMSID` to a unique identifier for Vault
configuration. You will also need to create vault configuration similar to the
[example](../examples/rbd/kms-config.yaml) and use same `encryptionKMSID`.

pkg/util/crypto.go
```diff
@@ -96,31 +94,53 @@ func (kms SecretsKMS) DeletePassphrase(key string) error {
 	return nil
 }
 
+// GetID is returning empty string
+func (kms SecretsKMS) GetID() string {
```
ShyamsundarR: For secrets based KMS, if we return an empty string the following will occur,
- We will store an empty string in the OMap key for KMS [here](https://github.com/ceph/ceph-csi/pull/822/files?file-filters%5B%5D=.go#diff-a6cef7e66fbdf22f7065bf167606d504R244-R245)
- When using VolID based context building, we would retrieve it as an empty string, and not set encrypted property for the volune [here](https://github.com/ceph/ceph-csi/pull/822/files?file-filters%5B%5D=.go#diff-8e27516879efe971bf4755fadf585e06R355-R362)

From an testing standpoint, expanding this volume would have succeeded even though this is encrypted (also, creating a snapshot from this encrypted image will succeed). This is because these are the places where the `genVolFromVolID` is used.

We could leverage presence of the key, but absence of value (or empty value), to denote secrets based encryption.
OR,
As before, we choose a reserved name for the `kmsID` when it is defaulting to secrets, and note it in the doc. as a reserved name. (maybe, "csi-secrets", as this is from the CSI secrets field that we get the required data).

Reserving a name is always tricky, but I am borderline fine with it if others have no objections.

vasyl-purchel: good point actually, I'll update to actually store the value for default kms also

pkg/util/vault.go
```diff
@@ -61,88 +62,83 @@ Example JSON structure in the KMS config is,
...
+	}
+	kms.VaultPassphraseRoot, ok = config["vaultPassphraseRoot"]
+	if !ok || kms.VaultPassphraseRoot == "" {
+		kms.VaultPassphraseRoot = vaultDefaultPassphraseRoot
+	}
+	kms.VaultPassphrasePath, ok = config["vaultPassphrasePath"]
+	if !ok || kms.VaultPassphrasePath == "" {
+		kms.VaultPassphrasePath = vaultDefaultPassphrasePath
+	}
+	kms.VaultCAVerify = true
```
Madhu-1: do we need to enable CAVerify by default? if yes, can you add a comment here that we are enabling CAVerify by default

pkg/util/voljournal.go
```diff
@@ -416,8 +416,8 @@ func (cj *CSIJournal) GetObjectUUIDData(ctx context.Context, monitors string, cr
 		cj.cephUUIDDirectoryPrefix+objectUUID, cj.encryptKMSKey)
 	if err != nil {
 		if _, ok := err.(ErrKeyNotFound); !ok {
-			klog.Errorf(Log(ctx, "=> GetObjectUUIDData encryptedKMS failed: %s (%s)"), cj.cephUUIDDirectoryPrefix+objectUUID, err)
-			return "", "", "", err
+			return "", "", "", fmt.Errorf("OMapVal for %s/%s failed to get encryption KMS value: %s",
```
Madhu-1: don't we need to log here?

vasyl-purchel: the original logging was committed by mistake while I was just working on how it should be done and since we're returning error - no reason to log it multiple times

pkg/util/vault.go
vasyl-purchel: in general all tools have SSL verification on by default, this is an option to be able to disable it if needed, for the code itself it remains as it was before, just a bit more explicit

Issue #824, opened 2020-02-09
Revert code added to maintain cephfs mount cache on nodes


ShyamsundarR:
As functionality added by PR #282 for CephFS fuse mounts does not work as intended, from a code maintainability standpoint we should revert this commit or remove the added code.

jianglingxia:
if the #282  PR reverted ,the ceph csi driver daemonset plugin existed,caused the pod mount path transport endpoint is not connected error  problem can resolved? @ShyamsundarR 
thanks

Madhu-1:
it won't resolve any issue. this code is not working as expected am working on removing the code

jianglingxia:
@Madhu-1  really thanks, i want to know that the transport endpoint is not connected error ,the problem is cephfs-fuse client reason? not csi driver reason is right?

Madhu-1:
@jianglingxia see https://github.com/ceph/ceph-csi/issues/703#issuecomment-557520551 for the more detailed explanation

Madhu-1:
closing this one as #825 is merged

huaizong:
@ShyamsundarR  @jianglingxia  As PR #282 describe,  mount cache only work if Pod can restart when csi container exit, it may trigger  by health probe check or other similar technology. 

If Pod not check mount path, mount cache can not help in thhe scenarios. 

** NoGoal **

***Pod should exit and restart*** when csi plugin pod exit and mount point lost. if pod not exit will get error of transport endpoint is not connected.

PR #282, opened 2019-03-25
remount old mount point when csi plugin unexpect exit 

huaizong:
issue #217 #91

Goal

we try to solve when csi exit unexpect, the pod use cephfs pv can not auto recovery because lost mount relation until pod be killed and reschedule to other node. i think this is may be a problem. may be csi plugin can do more thing to remount the old path so when pod may be auto recovery when pod exit and restart, the old mount path can use.

NoGoal

Pod should exit and restart when csi plugin pod exit and mount point lost. if pod not exit will get error of **transport endpoint is not connected**.

implment logic

csi-plugin start:

	1. load all MountCachEntry  from node local dir
	2. check if volID exist in cluster, if no we ignore this entry, if yes continue
	3. check if stagingPath exist, if yes we mount the path
	4. check if all targetPath exist, if yes we binmount to staging path

NodeServer:

1. NodeStageVolume: add MountCachEntry on local dir include readonly attr and ceph secret
2. NodeStagePublishVolume: add pod bind mount path to MountCachEntry  and persist local dir
3. NodeStageunPublishVolume: remove pod bind mount path From MountCachEntry  and persist local dir
4. NodeStageunStageVolume: remove MountCachEntry  from local dir

PR #825, opened 2020-02-10
Remove mount cache for cephfs

Madhu-1:
PR #282 introduces the mount cache to solve cephfs fuse mount issue when cephfs plugin pod
restarts. This is not working as intended. This PR removes the code for maintainability.

## Related issues ##

Updates  #824


humblec:
> As the commit has "Fixes" keyword for the issue, it will close it automatically, we may want to track the issue to remove the deprecated option in the next release. I would suggest "Updates" as the keyword referencing the issue hence.

@madhu-1 can you address this ?  rest looks good to me. Thanks !

Madhu-1:
@humblec done

jianglingxia:
can somebody see the issue https://github.com/ceph/ceph-csi/issues/836
that v1.2.0 csi version used the cephfs and the mount-cache-dir function why only remounted one volume when csiplugin existed? but in the minion has many mountpath,thanks all!

jianglingxia:
@Madhu-1  
can you help me ? now i found that when plugin restart the remountCachedVolumes function only remount one volume,cause the monion mount path transport endpoint is not connected

the code bug that follow 
func (nc *NodeCache) ForAll(pattern string, destObj interface{}, f ForAllFunc) error {
	err := nc.EnsureCacheDirectory(nc.CacheDir)
	if err != nil {
		return errors.Wrap(err, "node-cache: couldn't ensure cache directory exists")
	}
	files, err := ioutil.ReadDir(path.Join(nc.BasePath, nc.CacheDir))
	if err != nil {
		return errors.Wrapf(err, "node-cache: failed to read %s folder", nc.BasePath)
	}
	cachePath := path.Join(nc.BasePath, nc.CacheDir)
	for _, file := range files {
		err = decodeObj(cachePath, pattern, file, destObj)
		if err == errDec {
			continue
		} else if err == nil {
			if err = f(strings.TrimSuffix(file.Name(), filepath.Ext(file.Name()))); err != nil {
				return err
			}
		}
		return err   ----?this line?

	}
	return nil
}

Issue #823, opened 2020-02-09
Revisit and reduce, requested RBAC rules for kubernetes deployments


ShyamsundarR:
While debugging a particular setup, it was interesting to note that nodeplugins are given an RBAC that includes ability to "watch", "update" VolumeAttachments. I believe this is not required, and hence on taking a look at the RBAC, I find a few other permissions granted that are not required, for example [these](https://github.com/ceph/ceph-csi/blob/1695c6965dbae56fd04eebbe3efefa9eb227cd64/deploy/rbd/kubernetes/csi-nodeplugin-rbac.yaml#L28-L39).

At a point when config maps were used to store the PV and volume related data by ceph-csi, there were some RBAC rules added, but a lot of these are not required unless volumes with older versions are in use. (possibly not required at all for the nodeplugin as it never consulted the stored config maps)

Ideally the nodeplugin should need no RBAC to work, as kubelet invokes the required CSI calls and updates the required kubernetes objects.

As we have not reviewed the RBAC set, or possibly not kept them updated as the code has moved, opening this issue to,
- Update/cleanup RBAC
- Make parts of RBAC optional (at least in helm), under the case where ceph-csi 1.0 volumes are still in use
- Or, at least add some notes on why certain RBACs are required

Madhu-1:
this is on my pending task list for a long time, let me try to fix it up

ShyamsundarR:
Once RBAC is updated as per PR #827, prior to the next release we would need to,
- Rename (or merge the contents of) the dangling rbac files
- Update e2e/scripts/docs where the above files are referred to
- Add required upgrade documentation/steps to cover the RBAC change if any

Madhu-1:
@ShyamsundarR am closing this one, as this is already covered.

PR #827, opened 2020-02-11
Remove unwanted RBAC rules from ceph-csi

Madhu-1:
There are currently unwanted RBAC permission is given for ceph-csi, This PR reduces removes such unwanted RBAC resources.

currently notes.txt container old branch name. This PR updated the branch name with the release
   
updates: #823 


deploy/cephfs/kubernetes/csi-nodeplugin-rbac.yaml
```diff
@@ -3,51 +3,3 @@ apiVersion: v1
 kind: ServiceAccount
```
ShyamsundarR: Should we rename this file from `csi-nodeplugin-rbac.yaml` to ` csi-nodeplugin-sa.yaml` or even better yet just merge this content into `csi-cephfsplugin.yaml`? (comment/thought applies to the rbd counterpart files as well)

Madhu-1: I thought of doing it but the reason for not doing it is if we change the filename same changes need to be reflected in E2E and install script in example folder also. in feature, if we add new roles to node plugin we need to revert back the same changes. to avoid this keeping the filename as it it for now.

ShyamsundarR: I understand that more files need to be touched, but IMHO as the deployment manifests are one of the first user facing files, it is better to be clear in this regard.

Should we track it for renaming before the next release, such that if there are no RBAC changes till then, we will rename (or remove) this file and make required e2e, script and doc changes?

Also, does this removal require upgrade related documentation?

Madhu-1: Yes, I will let the issue be open, and also upgrade documentation will be updated during release time.

PR #840, opened 2020-02-24
check subvolume present in backend

### LINKED ISSUE #839: Interrupted VolumeCreate gRPC to CephFS backed volumes, may cause missing backing subvolumes
If a CreateVolume call is interrupted, post creating the required CSI journal entries, but prior to creating the backing CephFS subvolume, then a subsequent CreateVolume call will return a valid response with a VolumeID that has it's backing image missing.

This is because the code does not check for subvolume existence (like in the case for RBD) prior to responding that a volume creation was successful.

Madhu-1:
If a CreateVolume call is interrupted, post creating the required CSI journal entries,
but prior to creating the backing CephFS subvolume, then a subsequent CreateVolume call will return a valid response with a VolumeID that has it's backing image missing. This PR adds a check
for the backend image. If the image is not found it deletes the reserved keys in omap.

fixes #839


ShyamsundarR:
> I think The caller of `checkVolExists` could create the volume if the return value is `ErrVolumeNotFound`

In the RBD and in the CephFS (hence), when a half complete VolumeCreate is found, we undo the entire reservation and proceed to create it fresh. I would hence retain this as is.

Also, the number of times this should happen would be relatively rare, on active provisioner restarts, hence the cost penalty is not high.

Madhu-1:
@Mergifyio backports release-v2.0

Madhu-1:
@Mergifyio refresh

PR #842, opened 2020-02-25
created struct for keeping the state of a staging transaction

skoef:

As a follow-up on #758 I did some cleaning up on the way `NodeStageVolume` transactions can be rolled back if they fail halfway staging. By using a struct keeping state of the transaction, the code is a little cleaner and can be extended in the future more easily.

## Future concerns ##

I think a lot more cleaning up can be done to normalise code paths in the future.

### DIFF
--- a/pkg/cephfs/controllerserver.go
+++ b/pkg/cephfs/controllerserver.go
@@ -55,13 +55,6 @@ func (cs *ControllerServer) createBackingVolume(ctx context.Context, volOptions 
 		klog.Errorf(util.Log(ctx, "failed to create volume %s: %v"), volOptions.RequestName, err)
 		return status.Error(codes.Internal, err.Error())
 	}
-	defer func() {
-		if err != nil {
-			if errDefer := purgeVolume(ctx, volumeID(vID.FsSubvolName), cr, volOptions); errDefer != nil {
-				klog.Warningf(util.Log(ctx, "failed purging volume: %s (%s)"), volOptions.RequestName, errDefer)
-			}
-		}
-	}()
 
 	return nil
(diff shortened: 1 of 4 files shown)


skoef:
@Madhu-1 I also noticed a piece of [unreachable code](https://github.com/ceph/ceph-csi/blob/master/pkg/cephfs/controllerserver.go#L58), shall I removed that too?

Madhu-1:
Yes please

skoef:
Some other thing that I noticed, in  `CreateVolume`, the volume's size is [passed](https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/controllerserver.go#L186) to `createBackingImage` as well as the `rbdVolume` itself. `createBackingImage` doesn't do anything with the size except passing it to `createImage`, again alongside the `rbdVolume` itself. 

I suggest we remove the `volSize` parameter from both `createBackingImage` and `createImage` and just use `rbdVolume.VolSize` (called `pOpts.VolSize` in `createImage`) to round and format the size to pass to the RBD command.

skoef:
`resizeRBDImage` had the same redundant parameter as `createBackingImage` and `createImage`, fixed that as well.

Madhu-1:
@skoef Thanks for the cleanup

Madhu-1:
@ShyamsundarR PTAL

humblec:
@skoef thanks for the PR ! nice cleanup. The PR looks good to me! lgtm.  However I will wait may be some more for others to chime in.

humblec:
There is no point to wait any more as we are nearing code freeze. Approving this.

PR #860, opened 2020-03-16
Fix logging if the rbd manager command is supported

Madhu-1:
if there is an error when adding the rbd task we are logging the output which is empty. This PR logs the error if the rbd task is supported and there is an error.

### DIFF
--- a/pkg/rbd/rbd_util.go
+++ b/pkg/rbd/rbd_util.go
@@ -269,15 +269,19 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 
 	// attempt to use Ceph manager based deletion support if available
 	rbdCephMgrSupported, err := rbdManagerTaskDeleteImage(ctx, pOpts, cr)
+	if rbdCephMgrSupported && err != nil {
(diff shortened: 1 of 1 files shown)


pkg/rbd/rbd_util.go
```diff
@@ -269,15 +269,18 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
-		// attempt older style deletion
-		args := []string{"rm", image, "--pool", pOpts.Pool, "--id", cr.ID, "-m", pOpts.Monitors,
-			"--keyfile=" + cr.KeyFile}
-		output, err = execCommand("rbd", args)
+	if rbdCephMgrSupported && err != nil {
+		klog.Errorf(util.Log(ctx, "failed to add task to delete rbd image: %s, %v"), image, err)
+		return err
 	}
 
+	// attempt older style deletion
```
nixpanic: This old-style delete should only be done when `rbdCephMgrSupported` is false. Otherwise a 2nd delete is attempted, and that causes the CI to fail.

Madhu-1: @nixpanic done PTAL

PR #862, opened 2020-03-16
rbd: implement deleteImage() with go-ceph

nixpanic:

Use go-ceph librdb bindings to delete an RBD image in case CephMgr does not support background deletion.

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/ceph/ceph-csi/internal/util"
 
+	"github.com/ceph/go-ceph/rados"
 	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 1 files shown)


humblec:
@Madhu-1 @ShyamsundarR ptal.

nixpanic:
Marking Do-Not-Merge for now. Reworking the connection stuff so that it can be shared between rbd/cephfs/rados requests.

nixpanic:
This should be ready for review now, @Madhu-1

internal/rbd/rbd_util.go
```diff
@@ -288,12 +287,17 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
-			"--keyfile=" + cr.KeyFile}
-		output, err = execCommand("rbd", args)
+		var ioctx *rados.IOContext
+		ioctx, err = pOpts.conn.GetIoctx(pOpts.Pool)
+		if err != nil {
+			return err
+		}
+		defer ioctx.Destroy()
+
+		image := librbd.GetImage(ioctx, pOpts.RbdImageName)
```
Madhu-1, line 297: we already have a variable called image here https://github.com/ceph/ceph-csi/blob/1f0f2c04e12bb931b8d634aace760d73d2a75a0d/internal/rbd/rbd_util.go#L270. rename this to something else

nixpanic: Interesting, the linters are so strict, and mostly complain about these kind of things. Somehow this sneaked through!

PR #864, opened 2020-03-16
Replace rbd-image metadata commands with go-ceph implementation

nixpanic:

Improve performance of volume creation by replacing command execution (and reconnecting to the cluster) by using the go-ceph bindings and re-using the connection to the cluster.

This also removes passing credentials to the metadata helper functions. Later updates can remove them even more, and reduce them to a single `RbdVolume.Connect(credentials)` call.

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -156,7 +156,13 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	defer cs.VolumeLocks.Release(req.GetName())
 
-	found, err := checkVolExists(ctx, rbdVol, cr)
+	err = rbdVol.Connect(cr)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to connect to volume %v: %v"), rbdVol.RbdImageName, err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	found, err := rbdVol.Exists(ctx)
 	if err != nil {
 		if _, ok := err.(ErrVolNameConflict); ok {
 			return nil, status.Error(codes.AlreadyExists, err.Error())
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -129,6 +129,36 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
 	return rbdVol, nil
 }
 
+func buildCreateVolumeResponse(ctx context.Context, req *csi.CreateVolumeRequest, rbdVol *rbdVolume) (*csi.CreateVolumeResponse, error) {
+	if rbdVol.Encrypted {
+		err := rbdVol.ensureEncryptionMetadataSet(rbdImageRequiresEncryption)
+		if err != nil {
+			klog.Error(util.Log(ctx, err.Error()))
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	}
+
+	volumeContext := req.GetParameters()
+	volumeContext["pool"] = rbdVol.Pool
+	volumeContext["journalPool"] = rbdVol.JournalPool
+	volumeContext["imageName"] = rbdVol.RbdImageName
+	volume := &csi.Volume{
+		VolumeId:      rbdVol.VolID,
+		CapacityBytes: rbdVol.VolSize,
+		VolumeContext: volumeContext,
+		ContentSource: req.GetVolumeContentSource(),
+	}
+	if rbdVol.Topology != nil {
+		volume.AccessibleTopology =
+			[]*csi.Topology{
+				{
+					Segments: rbdVol.Topology,
+				},
+			}
+	}
+	return &csi.CreateVolumeResponse{Volume: volume}, nil
+}
+
 // CreateVolume creates the volume in backend
 func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	if err := cs.validateVolumeReq(ctx, req); err != nil {
@@ -171,33 +201,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if found {
-		if rbdVol.Encrypted {
-			err = rbdVol.ensureEncryptionMetadataSet(rbdImageRequiresEncryption)
-			if err != nil {
-				klog.Errorf(util.Log(ctx, err.Error()))
-				return nil, err
-			}
-		}
-
-		volumeContext := req.GetParameters()
-		volumeContext["pool"] = rbdVol.Pool
-		volumeContext["journalPool"] = rbdVol.JournalPool
-		volumeContext["imageName"] = rbdVol.RbdImageName
-		volume := &csi.Volume{
-			VolumeId:      rbdVol.VolID,
-			CapacityBytes: rbdVol.VolSize,
-			VolumeContext: volumeContext,
-			ContentSource: req.GetVolumeContentSource(),
-		}
-		if rbdVol.Topology != nil {
-			volume.AccessibleTopology =
-				[]*csi.Topology{
-					{
-						Segments: rbdVol.Topology,
-					},
-				}
-		}
-		return &csi.CreateVolumeResponse{Volume: volume}, nil
+		return buildCreateVolumeResponse(ctx, req, rbdVol)
 	}
 
 	rbdSnap, err := cs.checkSnapshotSource(ctx, req, cr)
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -389,9 +389,8 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	}
 	defer cs.VolumeLocks.Release(volumeID)
 
-	rbdVol := &rbdVolume{}
-	defer rbdVol.Destroy()
-	if err = genVolFromVolID(ctx, rbdVol, volumeID, cr, req.GetSecrets()); err != nil {
+	rbdVol, err := genVolFromVolID(ctx, volumeID, cr, req.GetSecrets())
+	if err != nil {
 		if _, ok := err.(util.ErrPoolNotFound); ok {
 			klog.Warningf(util.Log(ctx, "failed to get backend volume for %s: %v"), volumeID, err)
 			return &csi.DeleteVolumeResponse{}, nil
@@ -436,6 +435,7 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		}
 		return &csi.DeleteVolumeResponse{}, nil
 	}
+	defer rbdVol.Destroy()
 
 	// lock out parallel create requests against the same volume name as we
 	// cleanup the image and associated omaps for the same
@@ -506,8 +506,7 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	defer cr.DeleteCredentials()
 
 	// Fetch source volume information
-	rbdVol := new(rbdVolume)
-	err = genVolFromVolID(ctx, rbdVol, req.GetSourceVolumeId(), cr, req.GetSecrets())
+	rbdVol, err := genVolFromVolID(ctx, req.GetSourceVolumeId(), cr, req.GetSecrets())
 	if err != nil {
 		if _, ok := err.(ErrImageNotFound); ok {
 			return nil, status.Errorf(codes.NotFound, "source Volume ID %s not found", req.GetSourceVolumeId())
@@ -519,6 +518,7 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		}
 		return nil, status.Errorf(codes.Internal, err.Error())
 	}
+	defer rbdVol.Destroy()
 
 	// TODO: re-encrypt snapshot with a new passphrase
 	if rbdVol.Encrypted {
@@ -780,9 +780,7 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	}
 	defer cr.DeleteCredentials()
 
-	rbdVol := &rbdVolume{}
-	defer rbdVol.Destroy()
-	err = genVolFromVolID(ctx, rbdVol, volID, cr, req.GetSecrets())
+	rbdVol, err := genVolFromVolID(ctx, volID, cr, req.GetSecrets())
 	if err != nil {
 		if _, ok := err.(ErrImageNotFound); ok {
 			return nil, status.Errorf(codes.NotFound, "volume ID %s not found", volID)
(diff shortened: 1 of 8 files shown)


Madhu-1:
@nixpanic how about reworking on the framework to reuse context for all api calls before updating other commands with go-ceph.

nixpanic:
The static checks forced me to clean up more than I intended, but all CI jobs succeed now. Please have a look at this again. Thanks!

pkg/rbd/rbd_util.go
```diff
@@ -63,6 +62,8 @@ const (
 
 // rbdVolume represents a CSI volume and its RBD image specifics
 type rbdVolume struct {
+	util.ClusterConnection
```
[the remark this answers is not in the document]
nixpanic: My logic was that an `rbdVolume` is a connected object, it makes rewriting the current usage a little easier (less things need replacing). What if I renamed `ClusterConnection` to `ClusterObject` or `ConnectedObject`?

nixpanic:
This has been updated, some cleanups added to make the tests (mainly golint) pass again.

Please review, as this is the base for further go-ceph improvements. @humblec @Madhu-1

pkg/rbd/controllerserver.go
```diff
@@ -129,6 +129,36 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
 	return rbdVol, nil
 }
 
+func (cs *ControllerServer) getExistingVolume(ctx context.Context, req *csi.CreateVolumeRequest, rbdVol *rbdVolume) (*csi.CreateVolumeResponse, error) {
```
Madhu-1, line 132: can we update the function name to some more meaningful, as it's not getting the existingVolume as function name says, this can be a function, not a method as it won't use `cs`

pkg/rbd/controllerserver.go
```diff
@@ -129,6 +129,36 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
...
+	if rbdVol.Encrypted {
+		err := rbdVol.ensureEncryptionMetadataSet(rbdImageRequiresEncryption)
+		if err != nil {
+			klog.Errorf(util.Log(ctx, err.Error()))
+			return nil, err
```
Madhu-1, line 137: ensureEncryptedMetadataSet returns fmt.Errorf. need to return. GRPC errorcode and message to consumer

pkg/rbd/controllerserver.go
```diff
@@ -156,7 +186,13 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	defer cs.VolumeLocks.Release(req.GetName())
 
-	found, err := checkVolExists(ctx, rbdVol, cr)
+	err = rbdVol.Connect(cr)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to connect to volume %v: %v"), rbdVol.RbdImageName, err)
+		return nil, err
```
Madhu-1, line 192: Connects just return Errorf. need to return grpc error code and message

pkg/rbd/controllerserver.go
```diff
@@ -383,9 +393,11 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	}
 	defer cs.VolumeLocks.Release(volumeID)
 
-	rbdVol := &rbdVolume{}
-	defer rbdVol.Destroy()
-	if err = genVolFromVolID(ctx, rbdVol, volumeID, cr, req.GetSecrets()); err != nil {
+	rbdVol, err := genVolFromVolID(ctx, volumeID, cr, req.GetSecrets())
+	if rbdVol != nil {
+		defer rbdVol.Destroy()
```
Madhu-1, line 398: do we need to Destroy if the err!=nil?

pkg/rbd/nodeserver.go
```diff
@@ -766,10 +772,13 @@ func encryptDevice(ctx context.Context, rbdVol *rbdVolume, cr *util.Credentials,
 		return err
 	}
 
-	imageSpec := rbdVol.Pool + "/" + rbdVol.RbdImageName
-	err = util.SaveRbdImageEncryptionStatus(ctx, cr, rbdVol.Monitors, imageSpec, rbdImageEncrypted)
+	err = rbdVol.ensureEncryptionMetadataSet(rbdImageEncrypted)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, err.Error()))
```
Madhu-1, line 777: use klog.Error as formatting is not required. need to ensure this in other places too

pkg/rbd/rbd_util.go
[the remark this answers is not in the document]
Madhu-1: @phlogistonjohn any suggestion here?

pkg/rbd/rbd_util.go
```diff
@@ -977,21 +955,53 @@ func resizeRBDImage(rbdVol *rbdVolume, cr *util.Credentials) error {
 	return nil
 }
 
-func ensureEncryptionMetadataSet(ctx context.Context, cr *util.Credentials, rbdVol *rbdVolume) error {
-	var vi util.CSIIdentifier
+func (rv *rbdVolume) GetMetadata(key string) (string, error) {
+	ioctx, err := rv.GetIoctx(rv.Pool)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to get ioctx for \"%s\"", rv.RbdImageName)
```
Madhu-1, line 961:
```suggestion
return "", errors.Wrapf(err, "failed to get ioctx for %q", rv.RbdImageName)
```

pkg/rbd/rbd_util.go
```diff
@@ -977,21 +955,53 @@ func resizeRBDImage(rbdVol *rbdVolume, cr *util.Credentials) error {
...
+	}
 
-	err := vi.DecomposeCSIID(rbdVol.VolID)
+	image, err := librbd.OpenImage(ioctx, rv.RbdImageName, librbd.NoSnapshot)
+	if err != nil {
+		return "", errors.Wrapf(err, "could not open image \"%s\"", rv.RbdImageName)
```
Madhu-1, line 966: same here also use %q

Madhu-1: handle this in other places also

pkg/rbd/rbd_util.go
nixpanic: done in the next update

pkg/rbd/rbd_util.go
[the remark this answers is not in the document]
phlogistonjohn: I still think using composition is better than tweaking the name. Give the variable a short name like conn, and I don't see that much additional pain in using it that way.

pkg/rbd/controllerserver.go
nixpanic: the `if` is about `rbdVol != nil`, not `err`

pkg/rbd/controllerserver.go
nixpanic: done with the next update

pkg/rbd/nodeserver.go
nixpanic: Done with the next update

pkg/rbd/controllerserver.go
nixpanic: Done with the next update.

pkg/rbd/controllerserver.go
nixpanic: Ok, do you like `buildCreateVolumeResponse()` as name of the function? That is what will be included in the next update.

pkg/rbd/rbd_util.go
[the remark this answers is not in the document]
nixpanic: Right, I can do that now. Before the `Pool` was part of the `ClusterConnection` and that would have caused me updating every occurrence spread throughout the code. With the merge of the topology support, more Ceph pools are used, and the single `Pool` in the `ClusterConnection` would not be sufficient, so the `Pool` attribute moved back to `rbdVolume` and CephFS `volOptions`.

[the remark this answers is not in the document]
nixpanic: This is still a TODO at the moment, all other comments should have been addressed.

pkg/rbd/controllerserver.go
phlogistonjohn: If it is needed then it is probably also needed at around line 527 below.

pkg/rbd/rbd_journal.go
```diff
@@ -163,19 +163,19 @@ volume names as requested by the CSI drivers. Hence, these need to be invoked on
 respective CSI snapshot or volume name based locks are held, as otherwise racy access to these
 omaps may end up leaving the omaps in an inconsistent state.
 */
-func checkVolExists(ctx context.Context, rbdVol *rbdVolume, cr *util.Credentials) (bool, error) {
-	err := validateRbdVol(rbdVol)
+func (rv *rbdVolume) Exists(ctx context.Context) (bool, error) {
```
phlogistonjohn, line 166: it does not look like any of the other functions in this file are methods. Also, does it have to be a public function? It wasn't one before.

nixpanic: It is slowly moving to be more object oriented, more functions are rewritten into methods.
I can rename it to `exists()` and make it private, but there are no common guidelines in the project to make a function public/private. Everything related to RDB is currently in the rbd package, including the provisioner/attacher. At one point these may get split a little more to divide the CSI application and internal details.

phlogistonjohn: OK, I just am concerned that we get in the habit of uppercasing things rather than starting private (lowercase) and then only making what's truly needed public. I don't have strong feelings about it, I just noticed the change.

nixpanic: I prefer to keep it like this, and separate the attacherl, provisioner and `rbdVolume` into their own packages at one point.

pkg/rbd/controllerserver.go
nixpanic: I think this has been addressed already?

nixpanic:
@Madhu-1 @phlogistonjohn I think all comments have been addressed. In case I forgot something, please let me know and I'll try to get it addressed as soon as possible. Thanks!

pkg/rbd/rbd_journal.go
phlogistonjohn: OK, that sounds fine. This was more of a have-you-considered than this-is-an-issue anyway. :-)

internal/rbd/rbd_util.go
```diff
@@ -129,15 +130,31 @@ type rbdSnapshot struct {
...
+		return err
+	}
+
+	rv.conn = conn
+	return nil
+}
+
+// Destroy cleans up the rbdVolume and closes the connection to the Ceph
+// cluster in case one was setup.
+func (rv *rbdVolume) Destroy() {
```
phlogistonjohn, line 152: Disconnect rather than Destroy? Destroy sounds like it might apply to the rbd volume rather than just the cluster connection. To me this is no biggie and could probably be done in a follow up.

internal/rbd/rbd_util.go
```diff
@@ -129,15 +130,31 @@ type rbdSnapshot struct {
...
-	// time an operation can take (until refcounting of the connections is
-	// available)
-	cpInterval = 15 * time.Minute
-	cpExpiry   = 10 * time.Minute
-	connPool   = util.NewConnPool(cpInterval, cpExpiry)
 )
 
+// Connect an rbdVolume to the Ceph cluster
+func (rv *rbdVolume) Connect(cr *util.Credentials) error {
+	if rv.conn != nil {
```
phlogistonjohn, line 137: This logic seems like it could be useful outside of just rbd volume but we can tackle those changes in a later pr.

internal/rbd/rbd_util.go
nixpanic: `Disconnect()` is also not really correct, as the connection is kept open in the util.ConnPool (ok, maybe the comment is not correct either, but implementation details...). I do not like `Destroy()` much either, but it feels a little more suitable.

nixpanic:
@humblec PTAL, thanks!

internal/rbd/controllerserver.go
```diff
@@ -156,7 +186,13 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	defer cs.VolumeLocks.Release(req.GetName())
 
-	found, err := checkVolExists(ctx, rbdVol, cr)
+	err = rbdVol.Connect(cr)
```
humblec, line 189: At this time,`rbdvol.Connect` and cluster connection  do same thing. But I believe the idea is to have the rbbvol.Connect to have more functionlities in the future. So one question is, what exactly going to be  part of rbdvol.Connect()? @nixpanic

nixpanic: The cluster connection is a private component of the rbdVolume. In order to be able to clean up the contents of the packages, we need to make a clearer distinction what API's are private and which are public. Ideally the nodeserver and controllerserver do not need to access any internals from the rbd package, but only the rbd package's public API. The Connect() method is part of cleaning this up.

humblec: Sure. but my query was mainly on, does rbdvol is a "connect" to a 'specific' volume or its going to a handle to all RBD volumes in the Pool ? @nixpanic

nixpanic: `Connect()` will build the (shared) connection to the Ceph cluster. Depending on the functions that are executed, there is an `IOContext` needed, which is specific to a pool. Everything will go over the shared connection in the end (not I/O, only the image create/delete/... operations).

humblec: yeah, that matches my expectation. If that's the case, do you think a better name would be `rbdpool.Connect` or `rbdhandle.Connect` which can be embedded in rbdVol struct field? just a thought though. @nixpanic

nixpanic: It was suggested to not embed the connection in the rbdVolume, so I removed the embedding. The rbdVolume still needs to connect to the cluster, so I called the method `Connect()`. If that is confusing, I could rename the method to `ConnectToCluster()` or such.

humblec: IMO, one thing which we have to narrow down here is :

 Do we need rbdvolume to connect to cluster or not ? if yes, embedding a connector in the rbdvol looks a decent solution to me. If we do that, what we actually need? cluster connection or a connection-specific to list or operate only `rbd` commands or a connection which can operate only this specific Volume?

If its about cluster connection, we can embed cluster connector itself. If its about specific volume, I dont think we need check in later stage of this code path like `rbdVol.Exist()` ..etc. if its about the permission to operate on `any` RBD volume in the cluster, may be this connection has to be established by a specific credential that has limited permissions..etc, not sure its a valid use case though. @nixpanic

nixpanic: Currently an rbdVolume has a single connection. In the future we may want to split the connection, and have shared+cached IOContext objects instead (or in addition). Depending on what connections an rbdVolume needs (different IOContexts for different pools, separate journalling, topology-aware, multi-site, ...), embedding a connection object is not sufficient.

humblec: > Currently an rbdVolume has a single connection. In the future we may want to split the connection, and have shared+cached IOContext objects instead (or in addition). Depending on what connections an rbdVolume needs (different IOContexts for different pools, separate journalling, topology-aware, multi-site, ...), embedding a connection object is not sufficient.

but if this is about supporting above mentioned features, this connection object is basically a pool connection. Need some thoughts on this. Lets see.

Meanwhile, lets get this in and do the improvements accordingly. 

Approved.  Thanks @nixpanic

humblec:
LGTM.. Approved.

PR #875, opened 2020-03-19
util/cephcmds: have GetOMapValue() return ErrPoolNotFound

nixpanic:
On occasion the e2e tests fail as there is an unexpected error while
deleting an RBD image. The particular tests forcefully removes the pool
where the RBD image is stored. Deleting a volume that has been removed
already (or when its parent pool has been wiped), should succeed.

By catching the error that a pool does not exist (anymore), the
provisioner responds to the DeleteVolume request with succes.

### DIFF
--- a/pkg/util/cephcmds.go
+++ b/pkg/util/cephcmds.go
@@ -182,6 +182,11 @@ func GetOMapValue(ctx context.Context, monitors string, cr *Credentials, poolNam
 			return "", ErrKeyNotFound{poolName + "/" + oMapName + "/" + oMapKey, err}
 		}
 
+		if strings.Contains(stdoutanderr, "error opening pool "+
+			poolName+": (2) No such file or directory") {
+			return "", ErrPoolNotFound{poolName, err}
+		}
+
 		// log other errors for troubleshooting assistance
 		klog.Errorf(Log(ctx, "failed getting omap value for key (%s) from omap (%s) in pool (%s): (%v)"),
 			oMapKey, oMapName, poolName, err)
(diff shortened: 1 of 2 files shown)


pkg/util/cephcmds.go
```diff
@@ -182,6 +182,11 @@ func GetOMapValue(ctx context.Context, monitors string, cr *Credentials, poolNam
 			return "", ErrKeyNotFound{poolName + "/" + oMapName + "/" + oMapKey, err}
 		}
 
+		if strings.Contains(stdoutanderr, "error opening pool "+
```
Madhu-1: you need to return invalid request during the volume/snapshot Create operation. basically, you need to handle this error and return an appropriate error message

with current code, the Delete request won't fail because after decoding the pool id we first check the pool is present or not if it's not present we will return success.

nixpanic: I have added the handling of ErrPoolNotFound in the calling function now. Is that what you mean?

nixpanic: @Madhu-1 did I address your concern, or did I misunderstand?

Madhu-1: @nixpanic we also need to handle `GetObjectUUIDData`  https://github.com/ceph/ceph-csi/blob/master/pkg/util/voljournal.go#L434 . whoever is calling `GetObjectUUIDData` may also need to return proper error message also

humblec: @nixpanic can you please address above ^^^^

nixpanic: done!

nixpanic:
@humblec could you review this please?

Issue #872, opened 2020-03-19
make test failed


cxhuawei:
# Describe the bug #

 I pull the code of master branch, when I run 'make test', it failed with err logs

# Environment details #

system: macos

# Steps to reproduce #

1、go get -d github.com/ceph/ceph-csi
2、make test 

# Actual results #

# github.com/ceph/go-ceph/rados
../../../../pkg/mod/github.com/ceph/go-ceph@v0.2.0/rados/conn.go:5:11: fatal error: 'rados/librados.h' file not found
 #include <rados/librados.h>
          ^~~~~~~~~~~~~~~~~~
1 error generated.

Madhu-1:
@nixpanic can you please update development documentation related to this?

nixpanic:
@cxhuawei can you make sure to install the Ceph development headers? Packages are called like `librbd-devel`, `libcephfs-devel` on Fedora. Other Linux distributions may have different names.

chenxu1990:
> @cxhuawei can you make sure to install the Ceph development headers? Packages are called like `librbd-devel`, `libcephfs-devel` on Fedora. Other Linux distributions may have different names.

Packages are called `librados2-devel` and `librbd1-devel` on CentOS 7 , but I have not found packages on macos : )

nixpanic:
On Sun, Mar 22, 2020 at 07:52:58PM -0700, chenxu wrote:
> > @cxhuawei can you make sure to install the Ceph development headers? Packages are called like `librbd-devel`, `libcephfs-devel` on Fedora. Other Linux distributions may have different names.
> 
> Packages are called `librados2-devel` and `librbd1-devel` on CentOS 7 , but I have not found packages on macos : )

It will only be possible to build on MacOS if the Ceph client libraries
are available for the OS. I do not have any experience with MacOS, but I
could not find any references that they exist...

If it is possible to run Linux containers on MacOS, it should be
possible to have the build done inside a container. That is planned to
get done through PR #385 (or a replacement of it).

PR #876, opened 2020-03-19
dev-guide: add reference to required go-ceph dependencies

nixpanic:

Add documentation about the required dependencies for go-ceph.

## Related issues ##

Closes: #872

### DIFF
--- a/docs/development-guide.md
+++ b/docs/development-guide.md
@@ -17,6 +17,13 @@ it is **highly** encouraged to:
    [install](https://golang.org/doc/install) it on your system.
 * Setup the [GOPATH](http://www.g33knotes.org/2014/07/60-second-count-down-to-go.html)
    environment.
+* Ceph-CSI uses the native Ceph libaries through the [go-ceph
+   package](https://github.com/ceph/go-ceph). It is required to install the
+   Ceph C headers in order to compile Ceph-CSI. The packages are called
+   `libcephfs-devel`, `librados-devel` and `librbd-devel` on many Linux
+   distributions. See the [go-ceph installaton
+   instructions](https://github.com/ceph/go-ceph#installation) for more
+   details.
 * Run `$ go get -d github.com/ceph/ceph-csi`
    This will just download the source and not build it. The downloaded source
    will be at `$GOPATH/src/github.com/ceph/ceph-csi`


docs/development-guide.md
```diff
@@ -17,6 +17,13 @@ it is **highly** encouraged to:
    [install](https://golang.org/doc/install) it on your system.
 * Setup the [GOPATH](http://www.g33knotes.org/2014/07/60-second-count-down-to-go.html)
    environment.
+* Ceph-CSI uses the native Ceph libaries through the [go-ceph
+   package](https://github.com/ceph/go-ceph). It is required to install the
+   Ceph C headers in order to compile Ceph-CSI. The packages are called
+   `libcephfs-devel`, `librados-devel` and `librbd-devel` on many Linux
```
Madhu-1: I think we don't need to point to go-ceph project here as it pointer to do `go get` `build` `tests` and all, we can just mention the packages required as told in go-ceph
ex
```text
The native RADOS library and development headers are expected to be installed.

On debian systems (apt):

libcephfs-dev librbd-dev librados-dev
On rpm based systems (dnf, yum, etc):

libcephfs-devel librbd-devel librados-devel
```

docs/development-guide.md
```diff
@@ -17,6 +17,13 @@ it is **highly** encouraged to:
    [install](https://golang.org/doc/install) it on your system.
 * Setup the [GOPATH](http://www.g33knotes.org/2014/07/60-second-count-down-to-go.html)
    environment.
+* Ceph-CSI uses the native Ceph libaries through the [go-ceph
+   package](https://github.com/ceph/go-ceph). It is required to install the
+   Ceph C headers in order to compile Ceph-CSI. The packages are called
```
Madhu-1: `Ceph-CSI` to `ceph-csi` as  this is not header

nixpanic: Really? Ceph is a name, and CSI is an abbreviation. The main README also uses "Ceph CSI" (sometimes with the "-", sometimes without).

docs/development-guide.md
nixpanic: I think the reference is helpful, just as the pointer to install Go is. In case there are additional dependencies for go-ceph later on (or for other OS'es), there is no need to update this text.

docs/development-guide.md
Madhu-1: yeah, in lot of places we are using ceph-csi in few places we are calling Ceph-CSI

Issue #890, opened 2020-03-24
Add a 'make containerized-test' target to run tests in a container


nixpanic:
# Describe the feature you'd like to have #

Make is possible to run `make containerized-test` on developers systems, so that there is no need to install all the test tools. The tools would be provided in a container that is built the 1st time the tests are run (and updated when needed).

# What is the value to the end user? (why is it a priority?) #

It will be easier for developers to run the tests locally. This should speed up sending PRs, as developers will not be surprised by failures anymore.

# How will we know we have a good solution? (acceptance criteria) #

`make containerized-tests` work the same as the `make test` target.

# Additional context #

Testing and building in containerized environments makes it possible for developers to use different Linux distributions (and maybe MacOS?) for development, without the need to install many different tools.

PR #892, opened 2020-03-24
Use go modules from ./vendor in test scripts

nixpanic:

The test script do not use `-mod=vendor` and will download all dependencies even if they are available in the `./vendor` directory.

This is annoying for developers that want to run the tests, as downloading all dependencies takes time and costs bandwidth.

In addition to the above, `go get` expects a recent version of git to be available, but the version in the Ceph container (CentOS-7) is too old. `go get` fails to download any dependency that is not available yet.

## Related issues ##

Updates: #890

### DIFF
--- a/scripts/gosec.sh
+++ b/scripts/gosec.sh
@@ -3,6 +3,9 @@
 set -o pipefail
 
 if [[ -x "$(command -v gosec)" ]]; then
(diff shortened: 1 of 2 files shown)


humblec:
LGTM. Approved.

PR #894, opened 2020-03-24
tests: in case 'go test' is run in a container, skip TestGetPIDLimit()

nixpanic:

In (standard, non-privileged) container environments the /sys/fs/cgroup
mountpoint is not available. This would cause the tests to fail, as
TestGetPIDLimit() tries to write to the cgroup configuration.

The test will work when run as root on a privileged container or
directly on a host (as Travis CI does).

## Related issues ##

Updates: #890

### DIFF
--- a/docs/development-guide.md
+++ b/docs/development-guide.md
@@ -66,6 +66,10 @@ tests:
 
 `$ make test`
 
+Certain unit tests may require extended permissions or other external resources
+that are not available by default. To run these tests as well, export the
+environment variable `CEPH_CSI_RUN_ALL_TESTS=1` before running the tests.
+
 You will need to provide unit tests and functional tests for your changes
 wherever applicable.
 
--- a/pkg/util/pidlimit_test.go
+++ b/pkg/util/pidlimit_test.go
@@ -17,12 +17,18 @@ limitations under the License.
 package util
 
 import (
+	"os"
 	"testing"
 )
 
 // minimal test to check if GetPIDLimit() returns an int
 // changing the limit require root permissions, not tested
-func TestGetPIDLimix(t *testing.T) {
+func TestGetPIDLimit(t *testing.T) {
+	runTest := os.Getenv("CEPH_CSI_RUN_ALL_TESTS")
+	if runTest == "" {
+		t.Skip("not running test that requires root permissions and cgroup support")
+	}
+
 	limit, err := GetPIDLimit()
 
 	if err != nil {
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ env:
     - MINIKUBE_VERSION=v1.6.0
     - CHANGE_MINIKUBE_NONE_USER=true
     - KUBECONFIG=$HOME/.kube/config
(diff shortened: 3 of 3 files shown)


phlogistonjohn:
You might want to consider a flag/environment variable to make this un-skippable during CI runs as opposed to developer desktop machines, etc. So, continue to skip by default but that will ensure that if something in the ci environment changes the test fails rather than starts to quietly skip.

nixpanic:
On Tue, Mar 24, 2020 at 08:01:27AM -0700, John Mulligan wrote:
> You might want to consider a flag/environment variable to make this un-skippable during CI runs as opposed to developer desktop machines, etc. So, continue to skip by default but that will ensure that if something in the ci environment changes the test fails rather than starts to quietly skip.

Sure, that can be done. I am not a huge fan of magic environment
variables, but forcing to run the test in Travis CI would be good to
detect any failures.

humblec:
@ShyamsundarR @Madhu-1 ptal .

humblec:
@ShyamsundarR can you please review ?

pkg/util/pidlimit_test.go
```diff
@@ -17,12 +17,18 @@ limitations under the License.
...
 	"testing"
 )
 
 // minimal test to check if GetPIDLimit() returns an int
 // changing the limit require root permissions, not tested
-func TestGetPIDLimix(t *testing.T) {
+func TestGetPIDLimit(t *testing.T) {
+	runTest := os.Getenv("CEPH_CSI_RUN_ALL_TESTS")
+	if runTest == "" {
+		t.Skipf("not running test that requires root permissions and cgroup support")
```
Madhu-1: use `Skip` instead of `Skipf` as formatting is not required here

pkg/util/pidlimit_test.go
```diff
@@ -17,12 +17,18 @@ limitations under the License.
...
 import (
+	"os"
 	"testing"
 )
 
 // minimal test to check if GetPIDLimit() returns an int
 // changing the limit require root permissions, not tested
-func TestGetPIDLimix(t *testing.T) {
+func TestGetPIDLimit(t *testing.T) {
+	runTest := os.Getenv("CEPH_CSI_RUN_ALL_TESTS")
```
Madhu-1: @nixpanic sorry I forgot about it in the last review, can we update the doc for this ENV?

nixpanic: Added a note about it in the developers guide.

docs/development-guide.md
```diff
@@ -66,6 +66,10 @@ tests:
 
 `$ make test`
 
+Certain unit tests may require extended permissions or other external resources
+that are not available by default. To run these tests as well, export the
+environment variable `CEPH_CSI_RUN_ALL_TESTS=1` befor running the tests.
```
Madhu-1: befor to before

humblec:
@nixpanic can you revisit this PR ?

docs/development-guide.md
nixpanic: done, thanks

Issue #798, opened 2020-01-17
Support for Ceph's namespaces (not kubernetes namespaces)


simonpie:
# Describe the feature you'd like to have #

Support for Ceph's namespace as [described here](https://docs.ceph.com/docs/master/rados/operations/user-management/#namespace). 

# What is the value to the end user? (why is it a priority?) #

This would allow multiple kubernetes clusters to share one unique ceph cluster without creating a pool  per kubernetes cluster.  Pools in ceph *can be a computationally expensive* as stated in documentation and namespaces are the recommended way to segregate tenants/users.
 

# How will we know we have a good solution? (acceptance criteria) #

The kubernetes admin can specify a ceph namespace to use in the helm chart and all ceph's operations done in the backend are automatically using `--namespace=$mynamespace`.  Hence, all rbd operations are going to the specified namespace.  

A different kubernetes cluster using a different ceph client, could be assigned to a different ceph's namespace and would not be able to see the image done by the first cluster.

simonpie:
So, are there any plans to implement this ?

mehdy:
If no one's going to implement it. I'd be glad to do it.

@Madhu-1 Should I start working on it?

clwluvw:
@mehdy I can help you with it, too if you want :)))

simonpie:
We can help in testing the solution.

Madhu-1:
@mehdy yes please go ahead

Madhu-1:
@mehdy assigned this issue to you.

mehdy:
@Madhu-1 Good. Thanks.

mehdy:
@Madhu-1 I've been working on this lately and I found out that it's necessary to detect the `namespace` inside [`genVolFromID`](https://github.com/ceph/ceph-csi/blob/master/pkg/rbd/rbd_util.go#L368). So it makes sense to add the `namespace` to [`CSIIdentifier`](https://github.com/ceph/ceph-csi/blob/master/pkg/util/volid.go#L45) but ceph allows namespace names to be 64 chars long. 
I'm not sure it's a good idea to do it this way because of the [`maxVolIDLen`](https://github.com/ceph/ceph-csi/blob/master/pkg/util/volid.go#L53). And this is possible only if we put the namespace name as a string and not its hex. And I'm not sure if changing the `CSIIdentifier` is a breaking change or not?

[...]

PR #931, opened 2020-04-09
Make CephFS SubvolumeGroup configurable

madddi:

The name of the CephFS SubvolumeGroup for the CSI volumes is currently hardcoded to `csi`. To make permission management in multi tenancy environments easier, we'd like to place volumes of different Kubernetes clusters in different SubvolumeGroups.

The parameter is optional and defaults to `csi`, so it should be backward compatible.

### DIFF
--- a/internal/cephfs/volumeoptions.go
+++ b/internal/cephfs/volumeoptions.go
@@ -45,6 +45,7 @@ type volumeOptions struct {
 	ProvisionVolume     bool   `json:"provisionVolume"`
 	KernelMountOptions  string `json:"kernelMountOptions"`
 	FuseMountOptions    string `json:"fuseMountOptions"`
+	SubvolumeGroup      string
 }
 
 func validateNonEmptyField(field, fieldName string) error {
@@ -109,24 +110,30 @@ func extractMounter(dest *string, options map[string]string) error {
 	return nil
 }
 
-func getMonsAndClusterID(options map[string]string) (string, string, error) {
+func getClusterInformation(options map[string]string) (string, string, string, error) {
 	clusterID, ok := options["clusterID"]
 	if !ok {
 		err := fmt.Errorf("clusterID must be set")
-		return "", "", err
+		return "", "", "", err
 	}
 
 	if err := validateNonEmptyField(clusterID, "clusterID"); err != nil {
-		return "", "", err
+		return "", "", "", err
 	}
 
 	monitors, err := util.Mons(csiConfigFile, clusterID)
 	if err != nil {
 		err = errors.Wrapf(err, "failed to fetch monitor list using clusterID (%s)", clusterID)
-		return "", "", err
+		return "", "", "", err
 	}
 
-	return monitors, clusterID, err
+	subvolumeGroup, err := util.CephFSSubvolumeGroup(csiConfigFile, clusterID)
+	if err != nil {
+		err = errors.Wrapf(err, "failed to fetch subvolumegroup using clusterID (%s)", clusterID)
+		return "", "", "", err
+	}
+
+	return clusterID, monitors, subvolumeGroup, err
 }
 
 // newVolumeOptions generates a new instance of volumeOptions from the provided
@@ -140,7 +147,7 @@ func newVolumeOptions(ctx context.Context, requestName string, req *csi.CreateVo
 
 	volOptions := req.GetParameters()
 
-	opts.Monitors, opts.ClusterID, err = getMonsAndClusterID(volOptions)
+	opts.ClusterID, opts.Monitors, opts.SubvolumeGroup, err = getClusterInformation(volOptions)
 	if err != nil {
 		return nil, err
 	}
@@ -225,6 +232,10 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 		return nil, nil, errors.Wrapf(err, "failed to fetch monitor list using clusterID (%s)", vi.ClusterID)
 	}
 
+	if volOptions.SubvolumeGroup, err = util.CephFSSubvolumeGroup(csiConfigFile, vi.ClusterID); err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to fetch subvolumegroup list using clusterID (%s)", vi.ClusterID)
+	}
+
 	cr, err := util.NewAdminCredentials(secrets)
 	if err != nil {
 		return nil, nil, err
@@ -259,6 +270,10 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 		}
 
 		if err = extractOptionalOption(&volOptions.FuseMountOptions, "fuseMountOptions", volOpt); err != nil {
+			return nil, nil, err
+		}
+
+		if err = extractOptionalOption(&volOptions.SubvolumeGroup, "subvolumeGroup", volOpt); err != nil {
 			return nil, nil, err
 		}
 
@@ -358,7 +373,7 @@ func newVolumeOptionsFromStaticVolume(volID string, options map[string]string) (
 	// store NOT of static boolean
 	opts.ProvisionVolume = !staticVol
 
-	opts.Monitors, opts.ClusterID, err = getMonsAndClusterID(options)
+	opts.ClusterID, opts.Monitors, opts.SubvolumeGroup, err = getClusterInformation(options)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/internal/util/csiconfig.go
+++ b/internal/util/csiconfig.go
@@ -23,53 +23,88 @@ import (
 	"strings"
 )
 
-// clusterInfo strongly typed JSON spec for the below JSON structure
+const (
+	// defaultCsiSubvolumeGroup defines the default name for the CephFS CSI subvolumegroup.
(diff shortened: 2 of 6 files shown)


ShyamsundarR:
If there are multiple Ceph clusters in use by the same CSI instance/cluster, then the namespace value remains the same across all clusters if passed in via the CLI. If different namespaces are required across clusters, the it forces the CO cluster to run multiple instances of ceph-csi, each with its own namespace value.

It would hence be better to move this to the csi per-cluster [config](https://github.com/ceph/ceph-csi/blob/2cc59ca41155b21cfcf30999a17e4b029fc2f26c/examples/csi-config-map-sample.yaml#L22-L28) instead.

A similar requirement is being discussed in #798 .

madddi:
I'm not quite sure I understand you both correctly... First, this is not about RADOS namespaces, it's just about the name of the subvolumegroup. This should make it possible to restrict client permissions to their specific subvolumegroup.

Second, this parameter is intended to and can be used in StorageClasses, I tested it.

nixpanic:
Sorry, I was wrong in assuming that the parameter was passed on the cmdline of the provisioner. I have no idea how I came to that idea. So, my only comment related to the missing e2e tests still stands.

pkg/cephfs/volumeoptions.go
```diff
@@ -240,6 +257,10 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 			return nil, nil, err
 		}
 
+		if err = extractOptionalOption(&volOptions.SubvolumeGroup, "subvolumeGroup", volOpt); err != nil {
```
ShyamsundarR, line 260: `volOpt` is nil, when called from DeleteVolume, hence we will assume `SubvolumeGroup` as the default (as it is missing). As a result Delete would not work if the group name is stored as part of volume options, which in turn, partly comes from the StorageClass.

ShyamsundarR:
> I'm not quite sure I understand you both correctly... First, this is not about RADOS namespaces, it's just about the name of the subvolumegroup. This should make it possible to restrict client permissions to their specific subvolumegroup.
> 
> Second, this parameter is intended to and can be used in StorageClasses, I tested it.

Ack, I got messed up with the CLI as well for some reason. But, passing this via the StorageClass (as noted in the review) will not work, as the volume options are not passed in for DeleteVolume and related calls.

The CSI specification states that the storage provider should be able to identify the volume given just the volume handle/ID (which forms most of the discussion in #798 as well). As a result leveraging this just via the StorageClass is not feasible, unless this is decipherable given only the volume ID.

madddi:
I see... Makes total sense. I'd suggest to wait for a decision in #798 before continuing here. It would make sense to take the same approach, I think.

madddi:
As discussed in #798 I'll start implementing this as an option in the JSON config file. I'd propose the following structure:

```json
    [
      {
        "clusterID": "<cluster-id>",
        "monitors": [
          "<MONValue1>",
          "<MONValue2>",
          "<MONValueN>"
        ],
        "cephFS": {
          "subvolumeGroup": "<subvolumegroup for cephfs volumes>"
        }
      }
    ]
```

`subvolumeGroup` is a special option for CephFS, so I think we should place it in a dedicated "subtree".

Madhu-1:
> As discussed in #798 I'll start implementing this as an option in the JSON config file. I'd propose the following structure:
> 
> ```json
>     [
>       {
>         "clusterID": "<cluster-id>",
>         "monitors": [
>           "<MONValue1>",
>           "<MONValue2>",
>           "<MONValueN>"
>         ],
>         "cephFS": {
>           "subvolumeGroup": "<subvolumegroup for cephfs volumes>"
>         }
>       }
>     ]
> ```
> 
> `subvolumeGroup` is a special option for CephFS, so I think we should place it in a dedicated "subtree".
looks good , but take care of supporting backward compatibility for already created PVC

madddi:
I think this should do it. Rebased on `master` to resolve conflicts. Maybe someone can already take a look while I figure out how to extend the e2e tests.

@nixpanic If you've got a hint regarding e2e tests for me, I'd be very thankful :slightly_smiling_face:

docs/deploy-cephfs.md
```diff
@@ -83,8 +83,8 @@ is used to define in which namespace you want the configmaps to be stored
...
-| `volumeNamePrefix`                                                                                  | no             | Prefix to use for naming volumes (defaults to `csi-vol-`).                                                                                                                                                    |
-| `snapshotNamePrefix`                                                                                | no             | Prefix to use for naming snapshots (defaults to `csi-snap-`)
+| `volumeNamePrefix`                                                                                  | no             | Prefix to use for naming volumes (defaults to `csi-vol-`).                                                                                                                                                              |
+| `snapshotNamePrefix`                                                                                | no             | Prefix to use for naming snapshots (defaults to `csi-snap-`)                                                                                                                                                            |
```
Madhu-1, line 87: can you revert back this change?

internal/cephfs/volumeoptions.go
```diff
@@ -138,9 +145,13 @@ func newVolumeOptions(ctx context.Context, requestName string, req *csi.CreateVo
 		err  error
 	)
 
+<<<<<<< HEAD:internal/cephfs/volumeoptions.go
```
Madhu-1, line 148: need to remove this

Madhu-1:
> I think this should do it. Rebased on `master` to resolve conflicts. Maybe someone can already take a look while I figure out how to extend the e2e tests.
> 
> @nixpanic If you've got a hint regarding e2e tests for me, I'd be very thankful

you need to update E2E here https://github.com/ceph/ceph-csi/blob/master/e2e/utils.go#L385

docs/deploy-cephfs.md
madddi: Sorry, sure. This was my markdown-autoformatter...

madddi:
E2E tests seem to be failing because of
```
Apr 27 09:24:17.036: INFO: command vault login -address=http://vault.cephcsi-e2e-22450.svc.cluster.local:8200 sample_root_token_id > /dev/null && vault kv get -address=http://vault.cephcsi-e2e-22450.svc.cluster.local:8200 /secret/ceph-csi/0001-0024-bfc88ddf-8939-48a4-8145-3f157489838a-0000000000000001-be918f5d-8868-11ea-97a7-0242ac110010 failed: command terminated with exit code 2
```

Is this something I broke? It looks like a different issue...

Madhu-1:
> command vault login

that is not the actual issue, the actual issue is 

```text
Apr 27 09:31:57.791: INFO: >>> kubeConfig: /home/travis/.kube/config
Apr 27 09:31:58.366: INFO: stdErr occurred: Error ENOENT: Subvolume group 'csi' not found, cannot remove subvolume 'csi-vol-efd5b5a6-8869-11ea-8e70-0242ac110010'
```

madddi:
E2E tests passed :tada: @Madhu-1 @nixpanic could you review this again?

internal/cephfs/volumeoptions.go
```diff
@@ -109,24 +110,30 @@ func extractMounter(dest *string, options map[string]string) error {
 	return nil
 }
 
-func getMonsAndClusterID(options map[string]string) (string, string, error) {
+func getClusterInformation(options map[string]string) (string, string, string, error) {
```
nixpanic, line 113: Wow, returning three strings now, that becomes quite a lot. Maybe the `util.clusterInfo` struct should become usable. This can be done at a later time, in a separate PR (see #998).

internal/util/csiconfig.go
```diff
@@ -23,52 +23,78 @@ import (
...
+
+// CephFSSubvolumeGroup returns the subvolumeGroup for CephFS volumes. If not set, it returns the default value "csi"
+func CephFSSubvolumeGroup(pathToConfig, clusterID string) (string, error) {
+	cluster, err := readClusterInfo(pathToConfig, clusterID)
+	if err != nil {
+		return "", err
+	}
+
+	if cluster.CephFS.SubvolumeGroup == "" {
+		return "csi", nil
```
nixpanic, line 97: Please use a const for `"csi"` and define it somewhere in the top of this file with a  clear comment.

internal/cephfs/volumeoptions.go
madddi: Yes, I had the same thought but didn't want to blow up the PR. I see you've already assigned someone but I could also have a go if you want me to.

internal/util/csiconfig.go
madddi: Done :-)

madddi:
Perfect, will do that :slightly_smiling_face: Was already planning on squashing them but wanted to wait for your review to finish.

nixpanic:
That is unfortunate, #995 got merged and added comments to the `clusterInfo` struct. You''ll have to manually rebase and fix that up.

madddi:
@nixpanic Done and added the respective comments for the new `clusterInfo` properties.

internal/cephfs/volumeoptions.go
nixpanic: It is not urgent, so it really does not matter to me who works on it. Yug is a new colleague here at Red Hat and is expected to contribute more to Ceph-CSI in the future, so these tasks are nice for him to learn more about the project.

Issue #858, opened 2020-03-13
Removed PVs are not deleted from RBD


nextsux:
RBD images are not deleted from CEPH after PV is deleted. 

# Environment details #

- image used: quay.io/cephcsi/cephcsi:v2.0.0
- kubernetes: v1.17.3
- csi-rbdplugin logs:
```
I0313 14:38:12.145964       1 utils.go:158] ID: 26 GRPC request: {}
I0313 14:38:12.146421       1 utils.go:163] ID: 26 GRPC response: {}
I0313 14:38:56.482056       1 utils.go:157] ID: 27 Req-ID: pvc-50373129-5d28-4ab2-9911-2c3861ad28d3 GRPC call: /csi.v1.Controller/CreateVolume
I0313 14:38:56.482073       1 utils.go:158] ID: 27 Req-ID: pvc-50373129-5d28-4ab2-9911-2c3861ad28d3 GRPC request: {"capacity_range":{"required_bytes":1073741824000},"name":"pvc-50373129-5d28-4ab2-9911-2c3861ad28d3","parameters":{"clusterID":"015836b8-4c51-4825-94a6-776a67593b9a","pool":"kubernetes"},"secrets":"***stripped***","volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4","mount_flags":["discard"]}},"access_mode":{"mode":1}}]}
I0313 14:38:56.483702       1 rbd_util.go:487] ID: 27 Req-ID: pvc-50373129-5d28-4ab2-9911-2c3861ad28d3 setting disableInUseChecks on rbd volume to: false
I0313 14:38:57.425137       1 rbd_journal.go:250] ID: 27 Req-ID: pvc-50373129-5d28-4ab2-9911-2c3861ad28d3 generated Volume ID (0001-0024-015836b8-4c51-4825-94a6-776a67593b9a-0000000000000001-5e7a0462-6538-11ea-96df-d20f7c6000a1) and image name (csi-vol-5e7a0462-6538-11ea-96df-d20f7c6000a1) for request name (pvc-50373129-5d28-4ab2-9911-2c3861ad28d3)
I0313 14:38:57.425211       1 rbd_util.go:125] ID: 27 Req-ID: pvc-50373129-5d28-4ab2-9911-2c3861ad28d3 rbd: create csi-vol-5e7a0462-6538-11ea-96df-d20f7c6000a1 size 1024000M format 2 (features: ) using mon 100.125.15.200, pool kubernetes 
[...]

PR #932, opened 2020-04-09
rbd: fallback to inline image deletion if adding it as a task fails

nixpanic:

In case adding an image deletion task to Ceph Manager fails, fallback to deleting the image inline.

## Related issues ##

Fixes: #858


Madhu-1:
@nixpanic ci failure

nixpanic:
go linting failed with this:
```
pkg/rbd/rbd_util.go:239:3: ifElseChain: rewrite if-else to switch statement (gocritic)
		if strings.Contains(string(output), rbdTaskRemoveCmdInvalidString1) &&
		^
```
So, replaced the if/else with a switch statement.

Issue #725, opened 2019-12-02
Add support ReadOnlyMany for rbd


D1abloRUS:
# Describe the feature you'd like to have #

Add support ReadOnlyMany for rbd

# What is the value to the end user? (why is it a priority?) #

Pods on different nodes can mount ceph rbd volume on read

Madhu-1:
@D1abloRUS can you more information about the use-case?

D1abloRUS:
@Madhu-1 I have data, I need to run several pods on different nodes, and to read this data, without ReadOnlyMany it is impossible

troeger:
@Madhu-1 We operate a Kubernetes cluster for machine learning workload. Some of the training jobs operate on millions of single text files, which currently must be stored on a Ceph file system for concurrent read access. The amount of files puts heavy stress on the MDS daemons, especially when multiple training pods access this data in parallel all the time.

A ReadOnlyMany support for volumes would allow us to store the training data directly on a shared volume. This would remove the MDS daemons from the performance equation.

dillaman:
> A ReadOnlyMany support for volumes would allow us to store the training data directly on a shared volume. This would remove the MDS daemons from the performance equation.

How would you get the data into the ROX RBD-backed filesystem? Would you plan to create a new ROX volume from a snapshot?

WoohyungHan:
If you app use filesystem multiple active MDS daemon would be better. but if the app use block device itself, ROX rbd can be considered.

troeger:
@dillaman We didn't thought about this. The obvious first choice would be a job that fills the volume first. The data then usually persists forever in an unmodified state after this initial copy step.

@WoohyungHan We have that, 4 active MDS daemons and their backups. They still run under full load. Maybe this is only a case for MDS tuning, but without ROX volumes, we simply have no alternative to compare with.

PavloZabudskyi:
When deployment pod is scheduled at node that it lost, it can't be reschedulated at other node without ReadOnlyMany option.

```
wordpress-6bd7cdd99f-7rpxr   0/1     ContainerCreating   0          26m
wordpress-6bd7cdd99f-rpgx2   1/1     Terminating         0          62m
```
```
  Normal   Scheduled           <unknown>             default-scheduler        Successfully assigned wordpress/wordpress-6bd7cdd99f-7rpxr to server1
[...]

PR #949, opened 2020-04-16
Add support for rbd ROX PVC mounting

Madhu-1:
if the PVC access mode is ReadOnlyMany or single node read-only, mounting the rbd
device path to the staging path as read-only to avoid the write operation.

Fixes https://github.com/ceph/ceph-csi/issues/725


Madhu-1:
will rebase this after  https://github.com/ceph/ceph-csi/pull/951

pkg/rbd/nodeserver.go
```diff
@@ -386,7 +394,7 @@ func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeS
 	existingFormat, err := diskMounter.GetDiskFormat(devicePath)
 	if err != nil {
 		klog.Errorf(util.Log(ctx, "failed to get disk format for path %s, error: %v"), devicePath, err)
-		return err
+		return readOnly, err
 	}
 
 	if existingFormat == "" && !staticVol {
```
dillaman, line 400: Do not attempt to create a filesystem on a ROX volume

Madhu-1: Thanks for the pointers @dillaman, will make necessary changes

Madhu-1: @dillaman if I don't format it am hitting `cannot mount unformatted disk %s as we are manipulating it in read-only mode` https://github.com/kubernetes/utils/blob/master/mount/mount_linux.go#L353

dillaman: Correct. ROX would need to be a pre-existing disk image (a la static PV) -- or you would need to move all the formatting logic to the `ControllerCreateVolume` call to avoid the potential for races between nodes attempting to format the disk (ROX must always be read-only).

Madhu-1: one use case where the ROX need not be static PV is, cloning a PVC and mounting it as readOnly(ROX).

in `ControllerCreateVolume` we are supposed to create rbd image and map the rbd image and format it ext4 or xfs based on the format type and unmap it one the formatting is completed isn't it?

dillaman: > one use case where the ROX need not be static PV is, cloning a PVC and mounting it as readOnly(ROX).

Cloning/create from snapshot wouldn't be an issue since you would already have a file system on the RBD image.

> in `ControllerCreateVolume` we are supposed to create rbd image and map the rbd image and format it ext4 or xfs based on the format type and unmap it one the formatting is completed isn't it?

`ControllerCreateVolume` just creates the image, correct? I think all mapping and formatting is deferred to the node server plug-in.

Madhu-1: yes, in the current code base mapping and formatting are deferred to the node server.

dillaman: I'm not really suggesting to move it, I am just saying that would be the only thing you could do to safely allow the creation of file systems on ROX volumes. However, I think the ROX use-case should be for static PVs or PVs from existing PVs/snapshots, so no need to re-engineer the code.

Madhu-1: okay let me do some more testing and get back to you :)

Madhu-1:
@dillaman I did some more testing on this one. yes we can mount the rbd image without formatting as the cloned rbd image will have the filesystem on it as you told earlier. however, with current code, the mounting of the rbd image fails
 
```bash
# map rbd images as read-only
rbd map csi-vol-ef7c2743-98f3-11ea-a84a-0242ac110010 --pool=replicapool --read-only

# mount rbd image no the node
sh-4.2# mount -t ext4 -o _netdev,ro,defaults /dev/rbd2 /tmp/test
mount: cannot mount /dev/rbd2 read-only

# dmesg
[11004.179875] EXT4-fs (rbd2): INFO: recovery required on readonly filesystem
[11004.179882] EXT4-fs (rbd2): write access unavailable, cannot proceed (try mounting with noload)
[11059.642337] EXT4-fs (rbd2): INFO: recovery required on readonly filesystem
[11059.642344] EXT4-fs (rbd2): write access unavailable, cannot proceed (try mounting with noload)
[11070.063567] EXT4-fs (rbd2): INFO: recovery required on readonly filesystem
[11070.063568] EXT4-fs (rbd2): write access unavailable, cannot proceed (try mounting with noload)
```
mount rbd image with `noload` or `norecovery` fixes  the issue of mounting. do  we need to make `noload` as default option when mounting  the rbd image?

dillaman:
> mount rbd image with `noload` or `norecovery` fixes the issue of mounting. do we need to make `noload` as default option when mounting the rbd image?

That just sounds like you cloned an in-use/open file system.

Madhu-1:
> > mount rbd image with `noload` or `norecovery` fixes the issue of mounting. do we need to make `noload` as default option when mounting the rbd image?
> 
> That just sounds like you cloned an in-use/open file system.

yes, you are correct, I have cloned PVC which is already attached to the pod. now cloned a PVC which is not attached, working now @dillaman Thanks for the pointer

internal/csi-common/nodeserver-default.go
```diff
@@ -195,3 +195,13 @@ func ConstructMountOptions(mountOptions []string, volCap *csi.VolumeCapability)
 	}
 	return mountOptions
 }
+
+// Contains checks the opt is present in mountOptions
+func Contains(mountOptions []string, opt string) bool {
```
humblec, line 200: This would be better named to MountOptionContains for clarity.

Madhu-1: done

Madhu-1:
@humblec @nixpanic PTAL

internal/rbd/nodeserver.go
```diff
@@ -403,10 +412,29 @@ func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeS
...
+		req.VolumeCapability.AccessMode.Mode == csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY {
+		if !csicommon.MountOptionContains(opt, rOnly) {
+			opt = append(opt, rOnly)
+		}
+	}
+	if csicommon.MountOptionContains(opt, rOnly) {
+		readOnly = true
+	}
+
+	if fsType == "xfs" {
```
nixpanic, line 433: This needs a comment. Maybe it is even suitable to be in its own commit (related to volumes with the same uuid).

Madhu-1: This is just moved from line number 425 to here to keep options in one place

Madhu-1:
> Does the CephFS driver already support ROX as well? If not, support for the CephFS side should be added as well to maintain parity. Also, should the `README.md` table be updated to indicate this support is available?

Added cephfs ROX support in https://github.com/ceph/ceph-csi/pull/1169

Madhu-1:
@humblec @nixpanic resolved merge conflict and updated the Readme

humblec:
> @humblec @nixpanic resolved merge conflict and updated the Readme

Sure.. Thanks.!  Will get on this review soon!

e2e/rbd.go
```diff
@@ -602,6 +604,166 @@ var _ = Describe("RBD", func() {
...
+					err = createPVCAndvalidatePV(f.ClientSet, pvcClone, deployTimeout)
+					if err != nil {
+						Fail(err.Error())
+					}
+					appClone, err := loadApp(appClonePath)
+					if err != nil {
+						Fail(err.Error())
+					}
+
+					totalCount := 4
```
humblec, line 664: Its better to have a comment over here why this is `4` if there is a specific reason for this number.

Madhu-1: There is no reason for it and it is done the same way what we have done in other places also https://github.com/ceph/ceph-csi/blob/master/e2e/rbd.go#L299 https://github.com/ceph/ceph-csi/blob/master/e2e/cephfs.go#L197

humblec:
@Madhu-1 mostly looks good. A minor comment.

e2e/rbd.go
humblec: Then lets leave it as it is. Thanks.

humblec:
LGTM. Approved.

Madhu-1:
@nixpanic PTAL

Madhu-1:
@nixpanic can you please reivew it.

Madhu-1:
@MergifyIO refresh

Issue #903, opened 2020-03-31
export ceph rbd driver field


yuquan-wang:
# Describe the feature you'd like to have #
I'm developing my own ceph csi plugin based on this project, I want to access those unexported fields defined in driver.go like following

```
type Driver struct {
	cd *csicommon.CSIDriver // can we make it public or add get method to export it?

	ids *IdentityServer // same as above
	ns  *NodeServer // same as above
	cs  *ControllerServer // same as above
}
```
By the way, i'm also using ebs-driver, which i can use `driver.publishVolume()` for my own logic

If you don't mind, I can contribute a PR for this issue.

# What new functionality do you want? #
access to unexported fields so that we can make our own plugin based on this project

# What is the value to the end user? (why is it a priority?) #
enable user to develop their own plugin

# How would the end user gain value from having this feature? #
obviously user can make their own logic for csi

nixpanic:
Could you explain why you want your own variation of ceph-csi? We're happy to get more features and other enhancements included in this project. Have you considered contributing your changes?

yuquan-wang:
Sure, so we have the special usage to let all applications in one node to share a pv, that's the reason we may need variation of ceph-csi and create our own storage class for use.
Not sure if it is common usage or not, if it is, we'd like to contribute.

Madhu-1:
> Sure, so we have the special usage to let all applications in one node to share a pv, that's the reason we may need variation of ceph-csi and create our own storage class for use.
> Not sure if it is common usage or not, if it is, we'd like to contribute.

If I understood the use-case correctly I think it should be possible with current driver itself, as you can attach a PVC to all applications, whichever comes on the same node will share the PVC. 

am I missing something?

yuquan-wang:
@Madhu-1  thanks for reply, "attach a PVC to all applications" didn't meet our requirement, reason as follow:
1.  PVC has namespace, we can not let all application under different namespace share one PVC
2.  Application may have several instances, then all of them may refer to same pv, that's not what we want in our case, what we want is application share pv in same node, but pv between nodes should be different.

Madhu-1:
@humblec @nixpanic @ShyamsundarR do you have any comments?

humblec:
@yuquan-wang let me try to understand this scenario some more better. Some comments inline. 

> 
>     1. PVC has namespace, we can not let all application under different namespace share one PVC
> 
[...]

PR #955, opened 2020-04-17
cleanup: move pkg/ to internal/

nixpanic:

The `internal/` directory in Go has a special meaning, and indicates that
those packages are not meant for external consumption. Ceph-CSI does
provide public APIs for other projects to consume. There is no plan to
keep the API of the internally used packages stable.

By moving all the files from `pkg/` to `internal/` all open PRs will need to
get rebased. It seems that git is smart enough to handle most of the
changes without manual intervention (unless the PR changes import
statements near the `pkg -> internal` modification, of course).

## Related issues ##

Closes: #903


nixpanic:
> we are following the folder structure followed by other kubernetes sidecar projects. this is a breaking change.

This is not a side-car project, and our APIs are not expected to be re-used. It is quite different from core kubernetes projects in that respect.

> Note: with this automatic backporting will fail.

I do not think so. My local test show that cherry-picking changes work without issues, so does rebasing existing PRs on top of this.

nixpanic:
@humblec @Madhu-1 could you review this? By placing all internal APIs in the `internal/` directory, other external projects can not use them. This makes it clear the API is not stable and can change without breaking other consumers (as there are none).

humblec:
> @humblec @Madhu-1 could you review this? By placing all internal APIs in the `internal/` directory, other external projects can not use them. This makes it clear the API is not stable and can change without breaking other consumers (as there are none).

I am not sure currently anyone using it or not. iic, in some of the issues, it sounded like they are using. But, this change could help avoiding breakage if happens.  Considering we are moving towards `3.0.0` a major change or release, I belive this is fine. Approved.

PR #956, opened 2020-04-17
travis: split static checks and continue with tests in case one fails

nixpanic:
Run all `make static-checks`, even in case one fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,9 +69,11 @@ jobs:
         # install helm for helm lint
         - curl -L https://git.io/get_helm.sh | bash
       script:
-        - make static-check
+        - make go-lint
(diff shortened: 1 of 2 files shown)


nixpanic:
Well, that contradicts with the `make static-check` target. It has a list of checks, and it will abort execution in case one test fails. Shall we split that up in multiple executions then?

Madhu-1:
> Well, that contradicts with the `make static-check` target. It has a list of checks, and it will abort execution in case one test fails. Shall we split that up in multiple executions then?

Yes we can split it

nixpanic:
Updated to split the `make static-checks`, and run them all in case one fails.

humblec:
LGTM..thanks

PR #958, opened 2020-04-18
add make target for containerized tests

nixpanic:

`make containerized-test` has been added as a make target. This runs the
'make test' target in a container. All test dependencies are installed
in the container once, and the container is kept around for running
`make containerized-test` subsequently.

The test container is based on Fedora:latest so that all test tools get
easily installed and are available in a recent version. The production
container is based on the Ceph container which has CentOS as Operating
System and therefor a more stable (too old) toolset.

## Future concerns ##

Ideally the e2e tests will also run inside a container. However deploying k8s in a container, and rook for the e2e tests is a huge hurdle to take.

Closes: #890

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,8 +73,9 @@ jobs:
         - make go-test
         - make mod-check || travis_terminate 1;
 
-    - name: containerized build
+    - name: containerized test (Fedora) and build (CentOS)
       script:
+        - make containerized-test || travis_terminate 1;
         - make containerized-build || travis_terminate 1;
 
     - name: cephcsi with kube 1.14.10
--- a/Makefile
+++ b/Makefile
@@ -65,15 +65,24 @@ cephcsi:
 	if [ ! -d ./vendor ]; then (go mod tidy && go mod vendor); fi
 	GOOS=linux GOARCH=$(GOARCH) go build -mod vendor -a -ldflags '$(LDFLAGS)' -o _output/cephcsi ./cmd/
 
-.PHONY: containerized-build
+.PHONY: containerized-build containerized-test
 containerized-build: .devel-container-id
 	$(CONTAINER_CMD) run --rm -v $(PWD):/go/src/github.com/ceph/ceph-csi$(SELINUX_VOL_FLAG) $(CSI_IMAGE_NAME):devel make -C /go/src/github.com/ceph/ceph-csi cephcsi
+
+containerized-test: .test-container-id
+	$(CONTAINER_CMD) run --rm -v $(PWD):/go/src/github.com/ceph/ceph-csi$(SELINUX_VOL_FLAG) $(CSI_IMAGE_NAME):test make test
 
 # create a (cached) container image with dependencied for building cephcsi
 .devel-container-id: scripts/Dockerfile.devel
 	[ ! -f .devel-container-id ] || $(CONTAINER_CMD) rmi $(CSI_IMAGE_NAME):devel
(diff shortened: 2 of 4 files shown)


scripts/Dockerfile.test
```diff
@@ -0,0 +1,43 @@
+# Container image for running the Ceph-CSI tests
```
Madhu-1, line 1: how about having a single dockerfile with multistage build support for container build and test. i see the way going forward to have everything inside one container with a single command

.travis.yml
```diff
@@ -73,6 +73,10 @@ jobs:
         - make go-test
         - make mod-check || travis_terminate 1;
 
+    - name: containerized test
+      script:
+        - make containerized-test || travis_terminate 1;
```
Madhu-1, line 78: can we have this with the containerized build itself? spanning more jobs takes lot of time when we have multiple builds in flight

Makefile
```diff
@@ -65,16 +65,25 @@ cephcsi:
 	if [ ! -d ./vendor ]; then (go mod tidy && go mod vendor); fi
 	GOOS=linux GOARCH=$(GOARCH) go build -mod vendor -a -ldflags '$(LDFLAGS)' -o _output/cephcsi ./cmd/
 
-.PHONY: containerized-build
+.PHONY: containerized-build containerized-check
```
Madhu-1, line 68: containerized-check where this is defined?

go.mod
```diff
@@ -28,7 +28,7 @@ require (
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
 	google.golang.org/appengine v1.6.5 // indirect
 	google.golang.org/genproto v0.0.0-20200413115906-b5235f65be36 // indirect
-	google.golang.org/grpc v1.26.0
+	google.golang.org/grpc v1.27.0
```
Madhu-1, line 31: is this change expected?

nixpanic: Ah, no, it seems to be an issue in the master branch. The `go mod verify` 'check' that exists does not detect an error, it modifies the file...

Makefile
nixpanic: Ah, should be `containerized-test`

.travis.yml
nixpanic: sure, I can combine them in a single job

scripts/Dockerfile.test
nixpanic: That is quite difficult. I tried to use the `ceph/cecp:latest` image as base, but the tools we use for testing are not readily available on CentOS, and that is the target OS for the final images. Developers also often have a more modern distribution than the OS used for production, so I think there is some value in running tests with two distributions.

Madhu-1: but if we use fedora as the first base image and we can switch to ceph/ceph:latest later in the multistage docker build, its possible isn't it?

go.mod
nixpanic: Removed, will send an other PR to correct the `make mod-check` test and have it report an error.

Makefile
nixpanic: done

.travis.yml
nixpanic: done!

scripts/Dockerfile.test
nixpanic: It is possible, but we may have different versions of dependencies (`librados`, `librbd` and `libcephfs`). If the version difference is large enough, there could be problems during runtime. Compiling & linking against the versions that are available in the image will prevent potential issues.

go.mod
nixpanic: See #961

PR #962, opened 2020-04-20
Cleanup the containerized-build Dockerfile

nixpanic:
This cleans up a few minor things in the `scripts/Dockerfile.devel`:

- use `dnf` instead of `yum` as the Ceph base images now uses CentOS 8
- hardcode the GOROOT `/usr/local/go` path only once
- set `WORKINGDIR` so that `make` commands do not need to pass the path

### DIFF
--- a/scripts/Dockerfile.devel
+++ b/scripts/Dockerfile.devel
@@ -1,13 +1,14 @@
 FROM ceph/ceph:v15
 
+ARG GOROOT=/usr/local/go
+
 ENV GOPATH=/go \
- GOROOT=/usr/local/go \
- GO111MODULE=on
+ GOROOT=${GOROOT} \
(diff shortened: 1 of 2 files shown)


scripts/Dockerfile.devel
```diff
@@ -1,19 +1,22 @@
 FROM ceph/ceph:v15
 
+ARG GOROOT=/usr/local/go
```
Madhu-1, line 3: why GOROOT need to be an argument inside the container?

nixpanic: `ARG` does not export the variable in the environment, but is a notation to provide a variable in the Dockerfile. This way `GOROOT` only needs to be set once, and can be used on the single `ENV` line.

PR #967, opened 2020-04-21
doc: add description for containerized-test to development-guide

nixpanic:
As suggested by @wilmardo in #797 add a description of how to run the tests locally.

### DIFF
--- a/docs/development-guide.md
+++ b/docs/development-guide.md
@@ -44,6 +44,26 @@ To build ceph-csi in a container:
 
 The built binary will be present under `_output/` directory.
 
+### Running Ceph-CSI tests in a container
+
+Once the changes to the sources compile, it is good practise to run the tests
+that validate the style and other basics of the source code. Execute the unit
+tests (in the `*_test.go` files) and check the formatting of YAML files,
+MarkDown documents and shell scripts:
+
+`$ make containerized-test`
+
+It is also possible to run only selected tests, these are the targets in the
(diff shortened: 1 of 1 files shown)


docs/development-guide.md
```diff
@@ -44,6 +44,26 @@ To build ceph-csi in a container:
 
 The built binary will be present under `_output/` directory.
 
+### Running Ceph-CSI tests
+
+Once the changes to the sources compile, it is good practise to run the tests
+that validate the style and other basics of the source code. Execute the unit
+tests (in the `*_test.go` files) and check the formatting of YAML files and
```
Madhu-1, line 51: it also validates the .md files

docs/development-guide.md
```diff
@@ -44,6 +44,26 @@ To build ceph-csi in a container:
...
+shell scripts:
+
+`$ make containerized-test`
```
Madhu-1, line 54: this is more specific containerized tests. need to update the title or put a subtitle as `running containerized test`

docs/development-guide.md
nixpanic: yes, and more. But I added "MarkDown documents" too now.

docs/development-guide.md
nixpanic: Yeah, that makes sense. I was not going to explain how to run the tests outside of a container as it requires installing quite some dependencies. Title has been updated.

Issue #975, opened 2020-04-23
Improve code quality and adhere better to Golang standards


nixpanic:
# Describe the feature you'd like to have #

[Go Report Card](https://goreportcard.com/report/github.com/ceph/ceph-csi) shows nice statistics about the code quality of the project. There are a few sections that do not reach 100%. This can easily be addressed, after which the configuration for the tests configuration in `scripts/golangci.yml` and similar can be updated.

# What is the value to the end user? (why is it a priority?) #

Higher code quality, more Go standardization and easier to understand code (less complex).

# How will we know we have a good solution? (acceptance criteria) #

All sections on the Go Report Card should list 100% and the tests in the repository should enforce this as much as possible.

Yuggupta27:
Thanks. On it!

nixpanic:
> /assign @Yuggupta27

We do not have a bot that handles these requests. Done manually now.

I have just sent a PR for one of the parts, see #977. Reviews welcome :love_you_gesture:

PR #977, opened 2020-04-23
e2e: fix IneffAssign warnings in checkPVSelectorValuesForPVC()

nixpanic:
IneffAssign warns about the two following statements:

  Line 1342: warning: ineffectual assignment to rFound (ineffassign)
  Line 1350: warning: ineffectual assignment to zFound (ineffassign)

rFound and zFound should be set before entering the loop, otherwise the
initial value will overwrite the updated value on each iteration.

Reported-by: https://goreportcard.com/report/github.com/ceph/ceph-csi
Updates: #975

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -1330,10 +1330,9 @@ func checkPVSelectorValuesForPVC(f *framework.Framework, pvc *v1.PersistentVolum
 		Fail("Found empty NodeSelectorTerms in PV")
 	}
 
+	rFound := false
+	zFound := false
 	for _, expression := range pv.Spec.NodeAffinity.Required.NodeSelectorTerms[0].MatchExpressions {
-		rFound := false
-		zFound := false
-
 		switch expression.Key {
 		case nodeCSIRegionLabel:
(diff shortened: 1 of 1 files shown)


nixpanic:
@phlogistonjohn wrote:
> I do worry a bit that this means this logic was not exercised in any tests?

Indeed, I was expecting that the test would now fail, but it does not. Maybe @ShyamsundarR wants to have a look at it and send a follow-up?

humblec:
LGTM.

ShyamsundarR:
> @phlogistonjohn wrote:
> 
> > I do worry a bit that this means this logic was not exercised in any tests?
> 
> Indeed, I was expecting that the test would now fail, but it does not. Maybe @ShyamsundarR wants to have a look at it and send a follow-up?

Yes, the logic is not exercised in any tests, as we need to find duplicate selector labels for it to fail (IOW, xFound to be True). It is added in there to check if there are duplicates, but nothing generates a duplicate. I also do not know how to generate duplicates to write a follow-up.

Issue #968, opened 2020-04-22
Update Prepare.sh in centos CI branch


Madhu-1:
Update the `prepare.sh` script in [ci/centos](https://github.com/ceph/ceph-csi/blob/ci/centos/prepare.sh) to pull the latest pull request (PR)  which is required for centos CI integration with cephcsi project.

Madhu-1:
cc @Yuggupta27

nixpanic:
The `prepare.sh` script is executed in the [Jenkins Pipeline `make-containerized-build.groovy`](https://github.com/ceph/ceph-csi/blob/ci/centos/make-containerized-build.groovy) and receives several parameters. The script does currently not interpret or handle any parameters at all.

Ideally the `prepare.sh` script uses a parameter `--ref=pulls/123/head` that points to the PR or branch that needs to get cloned/checkedout. The [Samba nightly build script](https://github.com/gluster/centosci/blob/master/jobs/scripts/nightly-samba-builds/nightly-samba-builds.sh#L32) does something similar, see the usage of `ghprbPullId` (as an environment variable there, not paramater as we want in `prepare.sh`).

Yuggupta27:
Okay, I'll keep that in mind! Thanks

Madhu-1:
closing this as #997 is merged

PR #986, opened 2020-04-24
Prepare centos/ci

Yuggupta27:
[WIP] Update **Prepare.sh** in centos ci branch

The script accepts three command-line arguments:
- reference to the latest pr
- working directory
- git repository 

**Sample usage:** 
./prepare.sh  --gitrepo=https://github.com/example --workdir=opt/build --ref=pull/123/head

Fixes: #968

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -1,7 +1,65 @@
 #!/bin/sh
 
+ARGUMENT_LIST=(
+    "ref"
+    "workdir"
+    "gitrepo"
+)
+
+opts=$(getopt \
+    --longoptions "$(printf "%s:," "${ARGUMENT_LIST[@]}") $(printf "help::")" \
+    --name "$(basename "${0}")" \
+    --options "" \
+    -- "$@"
+)
+
+eval set -- ${opts}
+
+while true; do
+    case "${1}" in
+    --help)  
+        shift
+        echo "Options:"
+        echo "--help|-h                 specify the flags"
+        echo "--ref                     specify the reference of pr"
+        echo "--workdir                 specify the working directory"
+        echo "--gitrepo                 specify the git repository" 
+        echo " "
+        echo "Sample Usage:"
+        echo "./prepare.sh --gitrepo=https://github.com/example --workdir=/opt/build --ref=pull/123/head"
+        exit 0
+        ;;
+    --gitrepo)  
+        shift
+        gitrepo=${1}
+        ;;
+    --workdir)  
+        shift
+        workdir=${1}
+        ;;
+    --ref)  
+        shift
+        ref=${1} 
+        #extracting Pull Id from passed reference
+        ghprbPullId="$(echo ${ref} | cut -d '/' -f2)"         
+        ;;
+    --)
+        shift
+        break
+        ;;
+    esac
+    shift
+done
+
 set -x
 
 yum -y install git podman
 
-git clone --single-branch --branch=master https://github.com/ceph/ceph-csi /opt/build
+#validates to true if ghprbPullId is not empty 
+if [ -n "${ghprbPullId}" ]
+then
+    git clone --depth=1 ${gitrepo} ${workdir}
+    cd ${workdir}
+    git fetch origin "${ref}:pr_${ghprbPullId}"
+    git checkout "pr_${ghprbPullId}"
+fi


prepare.sh
```diff
@@ -4,4 +4,44 @@ set -x
...
+      echo "options:"
+      echo "--help|-h                 specify the flags"
+      echo "--ref                     specify the reference of pr"
+      echo "--workdir                 specify the working directory"
+      exit 0
+      ;;
+
+#pr request will be passed through --ref flag
+    --ref*)  
+      export ref=`echo $1 | sed -e 's/^[^=]*=//g'`
```
nixpanic, line 20: there is not need to `export` the variables, you can just use them without that insided this script.

prepare.sh
```diff
@@ -4,4 +4,44 @@ set -x
 
 yum -y install git podman
 
-git clone --single-branch --branch=master https://github.com/ceph/ceph-csi /opt/build
+while test $# -gt 0; do
```
nixpanic, line 7: consider using `getopt` for parsing arguments, use `--longoptions help,ref:,workfdir:` for long names

prepare.sh
nixpanic, line 20: Instead of using backticks \`command\`, use `$(command)` for executing commands in a sub-shell.

prepare.sh
```diff
@@ -4,4 +4,44 @@ set -x
...
+      echo "--help|-h                 specify the flags"
+      echo "--ref                     specify the reference of pr"
+      echo "--workdir                 specify the working directory"
+      exit 0
+      ;;
+
+#pr request will be passed through --ref flag
+    --ref*)  
+      export ref=`echo $1 | sed -e 's/^[^=]*=//g'`
+      export ghprbPullId=`echo $ref | cut -d '/' -f2`
```
nixpanic, line 21: when you use `getopt`, you do not need to split commandline switch and parameter, that makes things much simpler

prepare.sh
```diff
@@ -4,4 +4,44 @@ set -x
...
+      export workdir=`echo $1 | sed -e 's/^[^=]*=//g'`
+      shift
+      ;;
+    *)
+      break
+      ;;
+  esac
+done
+
+git clone --depth=1 https://github.com/ceph/ceph-csi $workdir
```
nixpanic, line 38: pass the repository as a parameter to the script too, that makes it possible to run tests from other repositories if that is needed at one point. And it makes it easier to share this script with other CentOS CI projects in the future.

prepare.sh
```diff
@@ -4,4 +4,44 @@ set -x
...
+cd $workdir
+
+#validates to true if ghprbPullId is not empty 
+if [ -n "${ghprbPullId}" ]
+then
+	git fetch origin "$ref:pr_${ghprbPullId}"
```
nixpanic, line 44: try to use the notation `${variable}` for variables everywhere, it is clearer than `$variable`

prepare.sh
```diff
@@ -1,7 +1,64 @@
...
+        shift
+        echo "Options:"
+        echo "--help|-h                 specify the flags"
+        echo "--ref                     specify the reference of pr"
+        echo "--workdir                 specify the working directory"
+        echo "--gitrepo                 specify the git repository" 
+        echo " "
+        echo "Sample Usage:"
+        echo "./prepare.sh --gitrepo=https://github.com/example --workdir=/opt/build --ref=pull/123/head"
+        ;;
```
phlogistonjohn, line 34: You probably want to exit 0 (or equivalent) after printing the help.

prepare.sh
```diff
@@ -1,7 +1,64 @@
-#!/bin/sh
+#!/bin/bash
 
 set -x
 
 yum -y install git podman
 
-git clone --single-branch --branch=master https://github.com/ceph/ceph-csi /opt/build
+ARGUMENT_LIST=(
```
phlogistonjohn, line 7: Argument handling should happen before other "real" actions. Even if you pretty much always need to install git and podman it should happen after your parse the arguments.

prepare.sh
```diff
@@ -1,7 +1,64 @@
...
+)
+
+opts=$(getopt \
+    --longoptions "$(printf "%s:," "${ARGUMENT_LIST[@]}") $(printf "help::")" \
+    --name "$(basename "${0}")" \
+    --options "" \
+    -- "$@"
+)
+
+eval set --$opts
```
phlogistonjohn, line 20: IIRC there should be a space between the "--" and "$opts". Plus, it should probably be "${opts}". I highly encourage all vars to use quotes ("") and `${...}` in all locations (unless there's a good reason not to).

prepare.sh
```diff
@@ -1,7 +1,64 @@
...
+        ;;
+    --workdir)  
+        shift
+        workdir=${1}
+        ;;
+    --ref)  
+        shift
+        ref=${1} 
+        #extracting Pull Id from passed reference
+        ghprbPullId="$(echo ${ref} | cut -d '/' -f2)"
```
phlogistonjohn, line 47: Why discard a standard ref and force it into a PR style?

prepare.sh
```diff
@@ -1,7 +1,64 @@
...
+    shift
+done
+
+#validates to true if ghprbPullId is not empty 
+if [ -n "${ghprbPullId}" ]
+then
+    git clone --depth=1 ${gitrepo} ${workdir}
+    cd ${workdir}
+    git fetch origin "${ref}:pr_${ghprbPullId}"
+	  git checkout "pr_${ghprbPullId}"
```
phlogistonjohn, line 63: Odd indent.

Yuggupta27: Thanks for your valuable feedback! I'll make the required changes :)

prepare.sh
Yuggupta27: My apologies, I'm new with scripting and couldn't completely get your point. Could you please elaborate?

phlogistonjohn: Nothing too specific to scripting here. :-)
The command line option is --ref, which implies that any ref that is valid to git could be passed in and used. However, this line of code discards everything but the PR number (assuming  the github style of pr refs). Instead of silently discarding information I recommend either renaming the option, to say "--prnum" and explicitly making it depend on github style PRs or making the code general enough to check out any generic ref. Since the code for pr handling is already done the former is certainly going to be easier than the latter. I see that @nixpanic asked for an option of this naming scheme but I want to make sure what's (implicitly) advertised by the option's name matches what is implemented by the code.

phlogistonjohn: OK, now I went back an re-read everything.... and its sort of going to work but in an odd fashion.

Anything that is passed as a ref goes to the "git fetch origin ..." command but the actual local  name that is created has tried to get a pr number out of it and then names it "pr_BLAH".
If my cli is called with --ref=master you end up executing `git fetch origin master:pr_master`. If my cli was --ref=foo/bar `git fetch origin foo/bar:pr_bar. I just find this confusing.

Yuggupta27: I might be wrong here, but this script is meant to be triggered only during creation of a new Pull Request on the ceph-csi repo by Jenkins. Given the use case, I expected the references to be only of type "pull/ghprbPullId/head". Should we expect a different format of reference in our use case? Let me know what you think, if you recommend, I'll make the changes :)

nixpanic: The `--ref=...` will mostly be used for passing the reference to the Pull Request. However in future we will likely have nightly (or periodic) jobs that run from the master branch or others. These would be jobs that combine ceph-csi with other projects, think of updated versions of side-cars, an unreleased version of Kubernetes or very time-consuming tests that do not make sense to be part of the requirements for every PR (certain workload/functionality tests as [CDI](https://github.com/kubevirt/containerized-data-importer)).

So, try not to use `ghprbPullId` in this script, but only use `ref` or similar. In case `--ref` is not passed, you can set it to `master` as default.

prepare.sh
```diff
@@ -1,7 +1,65 @@
...
 yum -y install git podman
 
-git clone --single-branch --branch=master https://github.com/ceph/ceph-csi /opt/build
+#validates to true if ghprbPullId is not empty 
+if [ -n "${ghprbPullId}" ]
+then
+    git clone --depth=1 ${gitrepo} ${workdir}
+    cd ${workdir}
+    git fetch origin "${ref}:pr_${ghprbPullId}"
+    git checkout "pr_${ghprbPullId}"
```
nixpanic, line 64: As this script can checkout anything and not only PRs, rename the `pr_` part of the new branch. I like "Testing In Progress" (`tip/`) as prefix, other projects use that as well.

prepare.sh
```diff
@@ -1,7 +1,65 @@
...
+    shift
+done
+
 set -x
 
 yum -y install git podman
 
-git clone --single-branch --branch=master https://github.com/ceph/ceph-csi /opt/build
+#validates to true if ghprbPullId is not empty 
+if [ -n "${ghprbPullId}" ]
```
nixpanic, line 59: even if `ghprbPullId` or `ref` is not set, you want to clone the repo (the master branch would do). If you make sure `${ref}` is always set, you can remove this if-statement.

Yuggupta27: Thanks! I'll make the changes and revert back.

Issue #987, opened 2020-04-26
Support building via Podman


mykaul:
# Describe the feature you'd like to have #
It'd be nice to be able to build and run containerized tests using podman, and not Docker (which requires root privs).

What new functionality do you want?
See above.
# What is the value to the end user? (why is it a priority?) #
No root privs.
How would the end user gain value from having this feature?
See above.
# How will we know we have a good solution? (acceptance criteria) #
Be able to build without root, on a machine which has Podman, and not Docker.

Add a list of criteria that should be met for this feature to be useful

# Additional context #

PR #988, opened 2020-04-26
Makefile: allow building with Podman

mykaul:
If Docker does not exist, use Podman.
Use all CPUs for the build, which hopefully will make it faster.

Signed-off-by: Yaniv Kaul <ykaul@redhat.com>
Fixes: #987

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,12 @@
 
 .PHONY: all cephcsi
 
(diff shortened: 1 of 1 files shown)


humblec:
LGTM. 
jfyi, I have removed some template strings from the PR description.

PR #990, opened 2020-04-27
rbd: default nouuid if the formattype is xfs

### LINKED ISSUE #966: XFS: Superblock has unknown read-only compatible features (0x4) enabled
# Describe the bug #

# Environment details #

- Image/version of Ceph CSI driver v2.1.0
- helm chart version 1.2.5
- Kubernetes cluster version 1.16.7
- Logs

ern.alert <1> Apr 21 09:28:18 xxxx-2-05-worker-0 kernel: XFS (rbd0): Superblock has unknown read-only compatible features (0x4) enabled.
kern.warning <4> Apr 21 09:28:18 xxxx-2-05-worker-0 kernel: XFS (rbd0): Attempted to mount read-only compatible filesystem read-write.
kern.warning <4> Apr 21 09:28:18 xxxx-2-05-worker-0 kernel: XFS (rbd0): Filesystem can only be safely mounted read only.
kern.warning <4> Apr 21 09:28:18 xxxx-2-05-worker-0 kernel: XFS (rbd0): SB validate failed with error -22.
daemon.info <30> Apr 21 09:28:18 xxxx-2-05-worker-0 kubelet: E0421 09:28:18.051150    5822 csi_attacher.go:329] kubernetes.io/csi: attacher.MountDevice failed: rpc error: code = Internal desc = mount failed: exit status 32
daemon.info <30> Apr 21 09:28:18 xxxx-2-05-worker-0 kubelet: Mounting command: mount
daemon.info <30> Apr 21 09:28:18 xxxx-2-05-worker-0 kubelet: Mounting arguments: -t xfs -o _netdev,defaults /dev/rbd0 /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-a886c5d8-c6f7-4c5f-b9e9-e00acdb2b940/globalmount/0001-0009-rook-ceph-0000000000000001-60ac60d1-83b2-11ea-824a-8ae73a759cf9
daemon.info <30> Apr 21 09:28:18 xxxx-2-05-worker-0 kubelet: Output: mount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-a886c5d8-c6f7-4c5f-b9e9-e00acdb2b940/globalmount/0001-0009-rook-ceph-0000000000000001-60ac60d1-83b2-11ea-824a-8ae73a759cf9: wrong fs type, bad option, bad superblock on /dev/rbd0, missing codepage or helper program, or other error.
daemon.info <30> Apr 21 09:28:18 xxxx-2-05-worker-0 kubelet: I0421 09:28:18.051177    5822 controlbuf.go:508] transport: loopyWriter.run returning. connection error: desc = "transport is closing"

# Steps to reproduce #

1. Setup details: '...'
1. Deployment to trigger the issue '....'
1. See error

# Actual results #

Describe what happened

# Expected behavior #

# Additional context #

For example:

Any existing bug report which describe about the similar issue/behavior

Madhu-1:
make `nouuid` as the default mount option if the format type is `xfs` to avoid mounting issues.

fixes #966


Madhu-1:
> The problem happens when multiple PVCs with the same UUID are attached/mounted on a node. This can happen after creating a PVC from a snapshot, or cloning a PVC.

done. PTAL

dillaman:
lgtm

chenxu1990:
After this patch is merged , if the image of v2.1.0 will update automatically ？

Madhu-1:
> After this patch is merged , if the image of v2.1.0 will update automatically ？

no, it won't be, but the canary image will be updated

Madhu-1:
> Adding `-m reflink=0` by default is not good, as reflinks are a great feature for filesystems to have. Users on current OSs should have support for reflink on XFS. Please make this a conditional, maybe an option in the StorageClass and commandline option for the provisioner.

will see what to do. to have it as a optional parameter in sc or not

nixpanic:
> will see what to do. to have it as a optional parameter in sc or not

Definitely needs to be a commandline parameter for the provisioner, otherwise users on older OSs will get the new CentOS-8 based containers with the updated `mkfs.xfs`, and their existing StorageClass(es) can not be used anymore.

For new users, adding a compat option in the StorageClass might be good. But I would prefer to add a `csi.storage.k8s.io/mkfs-options` to the external-provisioner instead (like `csi.storage.k8s.io/fstype`).

humblec:
I see this is same as we check the `kernel` support and enable `quota`. We could follow the same route , so based on the kernel we will knob the reflink. 
https://github.com/ceph/ceph-csi/blob/master/internal/cephfs/volumemounter.go#L81

@Madhu-1 @nixpanic

Madhu-1:
> I see this is same as we check the `kernel` support and enable `quota`. We could follow the same route , so based on the kernel we will knob the reflink.
> https://github.com/ceph/ceph-csi/blob/master/internal/cephfs/volumemounter.go#L81
> 
> @Madhu-1 @nixpanic

looks good, only one thing we need here is exact kernel version from where on we need to disable reflink

nixpanic:
> looks good, only one thing we need here is exact kernel version from where on we need to disable reflink

Not only the kernel version. But e2e testing fails because the `mkfs.xfs` userspace tool does not have the `-m reflink=0` option. The build might be done with an older Ceph base container that is not running CentOS-8?

Madhu-1:
> > looks good, only one thing we need here is exact kernel version from where on we need to disable reflink
> 
> Not only the kernel version. But e2e testing fails because the `mkfs.xfs` userspace tool does not have the `-m reflink=0` option. The build might be done with an older Ceph base container that is not running CentOS-8?

something went wrong  i will check, the E2E should use the new build image not the old Ceph base image

Madhu-1:
> > looks good, only one thing we need here is exact kernel version from where on we need to disable reflink
> 
> Not only the kernel version. But e2e testing fails because the `mkfs.xfs` userspace tool does not have the `-m reflink=0` option. The build might be done with an older Ceph base container that is not running CentOS-8?

@nixpanic CI passed, 

@humblec only thing we need now is kernel version to disable it

Madhu-1:
@dillaman any idea on which kernel version we need to disable  reflink?

nixpanic:
On Wed, Apr 29, 2020 at 03:55:11AM -0700, Madhu Rajanna wrote:
> @humblec only thing we need now is kernel version to disable it

Not in an environment where different nodes can have the PVC attached.
In case there is a mix of recent OS's (like CentOS-8) and older versions
(like CentOS-7), the `-m reflink=0` option should be passed as well.

Users definitely need to be able to instruct the provisioner to not
format XFS filesystems with reflink support.

Detection depending on the kernel version is nice, but I see that as
optional. Linux 4.16 has XFS with reflink declared stable: https://kernelnewbies.org/Linux_4.16#XFS_reverse_mapping_and_reflink_features_are_now_stable

dillaman:
> @dillaman any idea on which kernel version we need to disable reflink?

What happens if the provisioner and node plug-in have different kernel versions? Unlike quota, this isn't a mount-time option and once enabled, it's enabled.

dillaman:
> we don't need it in the provisioner pod. we need to check only on node plugin pod as it is responsible to mount PVC

Indeed, sorry. Still have an issue if it ever migrates.

e2e/rbd.go
```diff
@@ -359,7 +359,10 @@ var _ = Describe("RBD", func() {
 					}
 
 					deleteResource(rbdExamplePath + "storageclass.yaml")
-					createRBDStorageClass(f.ClientSet, f, nil, map[string]string{"csi.storage.k8s.io/fstype": "xfs"})
+					parameters := map[string]string{"mkfsOptions": "-m reflink=0",
```
phlogistonjohn, line 362: I'd like to recommend not "snuggling" the value on the first line of the multi-line map when possible. I find it very easy to miss important content when some lines are snuggled together but others are not.

nixpanic:
I still think the provisioner should be able to get instructed by the commandline parameter to add `mkfsOptions` in case these are not set in the StorageClass. That makes it possible for users to keep the existing StorageClass, and update to a Ceph-CSI version based on CentOS-8 while the nodes run on an older Linux kernel.

pkalever:
@Madhu-1 may be we should split this PR to two:
1. addressing `nouuid`
2. addressing `-m reflink=0`

As the discussion is pending on 2 let it wait if needed, but 1 can be merged.
Let me know if you need any help in sending 1 as a separate PR.

Thanks!

Madhu-1:
Closing this as its not required anymore

nixpanic:
#1257 addressed this partially (reflink), an other PR added the `nouuid` option.

See #1256 for enhancements that are still needed.

pkalever:
> #1257 addressed this partially (reflink), an other PR added the `nouuid` option.
> 
> See #1256 for enhancements that are still needed.

Many thanks for the details @nixpanic , will check them out.

PR #993, opened 2020-04-28
travis: run deploy as final stage

nixpanic:

The `deploy:` section is executed as part of each successful job. That
means the container images are built and pushed many times during
testing after a push/merge event on the master branch.

By creating a deploy stage and listing it after the tests, it is
guaranteed to run only once after all tests have succeeded. By adding a
condition to only run the stage on the master branch for push events,
the images should only get built+pushed after a PR has been merged into
the master branch.

See-also: https://docs.travis-ci.com/user/conditional-builds-stages-jobs#Specifying-conditions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,12 +51,17 @@ before_script:
   # only the arm64 fallback repo is unsigned and needs --allow-unauthenticated
   - sudo apt-get -y --allow-unauthenticated install librados-dev librbd-dev
 
+# Two stages for testing, each stage runs its jobs in parallel, but stages are
+# run after each other, unless the last stage fails.
+# Only run the deploy stage on push (not pull_request) events.
+stages:
+  - build testing
+  - e2e testing
+  - name: deploy
+    if: type = push
+
 jobs:
   include:
-    # two stages for testing, each stage runs its jobs in parallel, but stages
-    # are run after each other, unless the last stage fails
-    # - build testing
-    # - e2e testing
     - stage: build testing
       name: static-check-make
       install:
@@ -132,8 +137,6 @@ jobs:
         - make image-cephcsi || travis_terminate 1;
         - scripts/travis-helmtest.sh v1.17.0 || travis_terminate 1;
 
-deploy:
-  - provider: script
-    on:  # yamllint disable-line rule:truthy
-      all_branches: true
-    script: ./deploy.sh
+    - stage: deploy
+      name: push artifacts to repositories
+      script: ./deploy.sh


.travis.yml
```diff
@@ -51,12 +51,18 @@ before_script:
...
 
+# Two stages for testing, each stage runs its jobs in parallel, but stages are
+# run after each other, unless the last stage fails.
+# Only run the deploy stage for the master branch on push (not pull_request)
+# events.
+stages:
+  - build testing
+  - e2e testing
+  - name: deploy
+    if: branch = master AND type = push
```
Madhu-1, line 62: can we remove branch Comparision just check its a PR or push?

.travis.yml
```diff
@@ -132,8 +138,6 @@ jobs:
         - make image-cephcsi || travis_terminate 1;
         - scripts/travis-helmtest.sh v1.17.0 || travis_terminate 1;
 
-deploy:
-  - provider: script
-    on:  # yamllint disable-line rule:truthy
-      all_branches: true
-    script: ./deploy.sh
+    - stage: deploy
+      name: push container images to quay.io
```
Madhu-1, line 142: need to change the name. it also pushes the helm charts

.travis.yml
nixpanic: Removed now. I added it as it was part of the original `deploy:` section.

.travis.yml
nixpanic: Sure, renamed to 'push artifacts to repositories'

.travis.yml
```diff
@@ -51,12 +51,18 @@ before_script:
   # only the arm64 fallback repo is unsigned and needs --allow-unauthenticated
   - sudo apt-get -y --allow-unauthenticated install librados-dev librbd-dev
 
+# Two stages for testing, each stage runs its jobs in parallel, but stages are
+# run after each other, unless the last stage fails.
+# Only run the deploy stage for the master branch on push (not pull_request)
```
Madhu-1, line 56: need to remove master branch reference

.travis.yml
```diff
@@ -51,12 +51,18 @@ before_script:
...
+# events.
+stages:
+  - build testing
+  - e2e testing
+  - name: deploy
```
Madhu-1, line 61: do we need `-name: deploy` here or just `deploy` is fine here to maintain consistency with above stages

Madhu-1: even am not sure the stages will be retriggered when the PR is merged.

nixpanic: > do we need `-name: deploy` here or just `deploy` is fine here to maintain consistency with above stages

I agree this looks weird, but this is the correct notation from what I understand from the docs.

nixpanic: > even am not sure the stages will be retriggered when the PR is merged.

That happens when the merge is pushed to the master branch. At least, I expect that to be the case :crossed_fingers:

.travis.yml
nixpanic: done!

.travis.yml
Madhu-1: then the changes look good. just we need to monitor the build logs once the PR is merged

Madhu-1:
@Mergifyio backport release-v2.1

nixpanic:
This can be backported once #999 is merged.

PR #1000, opened 2020-04-30
set CGO_ENABLED=1 in makefile

Madhu-1:
if we set the CGO_ENABLED=1 in makefile the user dont need to set the CGO_ENABLED=1
before running any commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,28 +65,28 @@ test: go-test static-check mod-check
 static-check: go-lint go-lint-text gosec
 
 go-test:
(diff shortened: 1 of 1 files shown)


Madhu-1:
> CGO is enabled by default, but some users may have `CGO_ENABLED=0` set. There should probably be a test that checks for dependencies/configuration, instead of setting the environment variable everywhere.

sounds good we can add a check for both CGO_ENABLED and go version also we need to go1.13 as minimum version

PR #1004, opened 2020-05-04
Revert back the ceph base image

Madhu-1:
as we are facing the xfs issue as the xfsprogs version has changed in ceph:v15 base image.
due to this cephcsi is not working in older kernel versions

updates: #966 

### DIFF
--- a/deploy/cephcsi/image/Dockerfile
+++ b/deploy/cephcsi/image/Dockerfile
@@ -1,6 +1,6 @@
 ARG SRC_DIR="/go/src/github.com/ceph/ceph-csi/"
 ARG GO_ARCH
-ARG BASE_IMAGE=ceph/ceph:v15
+ARG BASE_IMAGE=ceph/ceph:v14.2
 
 FROM ${BASE_IMAGE} as builder
 
@@ -17,7 +17,7 @@ ARG GOROOT=/usr/local/go
 RUN mkdir -p ${GOROOT} && \
     curl https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-${GO_ARCH}.tar.gz | tar xzf - -C ${GOROOT} --strip-components=1
 
-RUN dnf install libcephfs-devel librados-devel librbd-devel /usr/bin/cc make -y
+RUN yum install libcephfs-devel librados-devel librbd-devel /usr/bin/cc make -y
 
 ENV GOROOT=${GOROOT} \
     GOPATH=/go \


deploy/cephcsi/image/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG SRC_DIR="/go/src/github.com/ceph/ceph-csi/"
 ARG GO_ARCH
-ARG BASE_IMAGE=ceph/ceph:v15
+ARG BASE_IMAGE=ceph/ceph:v14.2
```
nixpanic, line 3: `dnf` is CentOS-8, you need to use `yum` for installing packages.

Madhu-1:
@humblec @nixpanic PTAL

Madhu-1:
this needs manual merging.

nixpanic:
The actual problem is `XFS (rbd0): Superblock has unknown read-only compatible features (0x4) enabled.` (in the kernel logs, `dmesg`) as that points to the "filesystem formatted with new xfsprogs, mounting on older kernel".

Madhu-1:
> The actual problem is `XFS (rbd0): Superblock has unknown read-only compatible features (0x4) enabled.` (in the kernel logs, `dmesg`) as that points to the "filesystem formatted with new xfsprogs, mounting on older kernel".

IMO the user will check the nodeplugin container logs not the dmesg at first

do you want me to add the dmesg logs in the commit message?

nixpanic:
> IMO the user will check the nodeplugin container logs not the dmesg at first
> 
> do you want me to add the dmesg logs in the commit message?

Yes, I expect that too. But the error is the generic "failed to mount" problem, it is required to look in `dmesg` for reporting. No need to add the message, lets get this out of the door quickly.

Madhu-1:
Merging this one as the CI is passed and we have two approvals

mykaul:
It's unclear to me - at some point, we'll have to upgrade - what's our long term solution to this?

humblec:
> It's unclear to me - at some point, we'll have to upgrade - what's our long term solution to this?

we could make a matrix based on `xfsprogs` and `kernel` in our code which allows to enable/disable `reflink` .
Additional reference # https://github.com/ceph/ceph-csi/issues/966

Madhu-1:
> It's unclear to me - at some point, we'll have to upgrade - what's our long term solution to this?

@mykaul this is a temp fix to unblock the upstream users who are hitting this issue, we are going to fix the issue in long term(we need more discussion to understand how things should work in mix kernel version cluster)

nixpanic:
> we could make a matrix based on `xfsprogs` and `kernel` in our code which allows to enable/disable `reflink` .

This is not sufficient. The pod where the provisioner or 1st attacher is running may have a different version of the kernel than other worker nodes. If that is the case, even a detection during provisioning, or during the 1st attaching, is not reliable.

There needs to be a commandline or StorageClass option that makes it possible to enable new feature (like reflink). Be conservative and enable only on request, as that prevents issues like this from happening again. #1009 is the PR for the master branch where this is also discussed.

PR #1005, opened 2020-05-04
build: add check for functional environment

nixpanic:
In case building fails, users may not always understand what is wrong. By adding a `check-env.sh` script, most common issues should be caught and explained.

### DIFF
--- a/scripts/check-env.sh
+++ b/scripts/check-env.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+#
+# Check the environment for dependencies and configuration
+#
+
+# count errors, run script to the end before exiting
+ERRORS=0
+fail() {
+	echo "${*}" > /dev/stderr
+	ERRORS=$((ERRORS+1))
+}
+
+# check if 'go' is available
+[ -n "$(command -v go)" ] \
+	|| fail "could not find 'go' executable"
+
+# parse the Golang version, return the digit passed as argument
+# 1.13.9 -> go_version 1 -> 1
+# 1.13.9 -> go_version 2 -> 13
+# 1.13.9 -> go_version 3 -> 9
+go_version() {
+	go version | cut -d' ' -f3 | sed 's/^go//' | cut -d'.' -f"${1}"
+}
+
+# Golang needs to be > 1.13
+GO_MAJOR=$(go_version 1)
+GO_MINOR=$(go_version 2)
+if ! [ "${GO_MAJOR}" -gt 1 ]
+then
+	if ! { [ "${GO_MAJOR}" -eq 1 ] && [ "${GO_MINOR}" -ge 13 ]; }
+	then
+		fail "go version needs to be >= 1.13"
+	fi
+fi
+
+# we're building with modules, so GO111MODULE needs to be 'on' or 'auto'
+# some versions of Go return nothing when GO111MODULE is not explicitly
+# configured
+[ "$(go env GO111MODULE 2>&1)" = '' ] || [ "$(go env GO111MODULE)" = 'on' ] || [ "$(go env GO111MODULE)" = 'auto' ] \
+	|| fail "GO111MODULE should be set to 'on' or 'auto'"
+
+# CGO needs to be enabled, we build with go-ceph language bindings
+[ "$(go env CGO_ENABLED)" = '1' ] \
+	|| fail "CGO_ENABLED should be set to '1'"
+
+exit ${ERRORS}
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: all cephcsi
+.PHONY: all cephcsi check-env
 
 CONTAINER_CMD?=$(shell docker version >/dev/null 2>&1 && echo docker)
 ifeq ($(CONTAINER_CMD),)
@@ -62,12 +62,12 @@ all: cephcsi
 
 .PHONY: go-test static-check mod-check go-lint go-lint-text gosec
 test: go-test static-check mod-check
-static-check: go-lint go-lint-text gosec
+static-check: check-env go-lint go-lint-text gosec
 
-go-test:
+go-test: check-env
 	./scripts/test-go.sh
 
-mod-check:
+mod-check: check-env
 	@echo 'running: go mod verify'
 	@go mod verify && [ "$(shell sha512sum go.mod)" = "`sha512sum go.mod`" ] || ( echo "ERROR: go.mod was modified by 'go mod verify'" && false )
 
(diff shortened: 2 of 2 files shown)


scripts/check-env.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
Madhu-1, line 1: we need to check `go` is installed or not in the first place

scripts/check-env.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+# Golang needs to be > 1.13
+GO_MAJOR=$(go_version 1)
+GO_MINOR=$(go_version 2)
+if ! [ ${GO_MAJOR} -gt 1 ]
+then
+	if ! ( [ ${GO_MAJOR} -eq 1 ] && [ ${GO_MINOR} -ge 13 ] )
+	then
+		fail "go version needs to be >= 1.13"
```
Madhu-1, line 29: can we just exit here, no need to perform other checks as the go version itself is not compatible

scripts/check-env.sh
```diff
@@ -0,0 +1,39 @@
...
+	fi
+fi
+
+# we're building with modules, so GO11MODULE needs to be 'on'
+[ "$(go env GO111MODULE)" = 'on' ] || fail "GO111MODULE should be set to 'on'"
```
Madhu-1, line 34: check go111MODULES is off here as it can be auto also

scripts/check-env.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# CGO needs to be enabled, we build with go-ceph language bindings
+[ "$(go env CGO_ENABLED)" = '1' ] || fail "CGO_ENABLED should be set to '1'"
```
Madhu-1, line 37: the comparison seems wrong here

scripts/check-env.sh
nixpanic: It is useful to list all failures at the end. Users can then address multiple issues before trying again.

scripts/check-env.sh
nixpanic: good point!

scripts/check-env.sh
nixpanic: Will add that

scripts/check-env.sh
nixpanic: What is wrong about it? It needs to be `1` OR needs to fail.

Madhu-1: it should be `==` or `=` here?

nixpanic: No, `=` is correct for shell scripts.

deploy/cephcsi/image/Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN dnf install libcephfs-devel librados-devel librbd-devel /usr/bin/cc make -y
 
 ENV GOROOT=${GOROOT} \
     GOPATH=/go \
+    GO111MODULE=on \
```
Madhu-1, line 24: GOMODULE is on by default in go 1.13.x we don't need to set this explicitly

nixpanic: What is the reason then that building fails? `go env` does not return `on` or `auto` it seems? Other variables that are enabled by default return a non-empty value with `go env`.

Maybe I am missing something?

nixpanic: ok, I see it is `on` by default, but `go env GO111MODULE` returns `on` in that case too. By setting the variable empty on the cmdline `make GO111MODULE=""` the script detects it as `on`. By setting it to `off` or an other value, the script reports the error.

Madhu-1: `""` is also a valid value. we need to consider it

nixpanic: Depending on the Golang version, `go env GO111MODULE` may return an empty string... Which means `on` or `auto`, so an empty-non-error-return should be valid too.

scripts/check-env.sh
```diff
@@ -0,0 +1,46 @@
...
+	if ! { [ "${GO_MAJOR}" -eq 1 ] && [ "${GO_MINOR}" -ge 13 ]; }
+	then
+		fail "go version needs to be >= 1.13"
+	fi
+fi
+
+# we're building with modules, so GO111MODULE needs to be 'on' or 'auto'
+# some versions of Go return nothing when GO111MODULE is not explicitly
+# configured
+[ "$(go env GO111MODULE 2>&1)" = '' ] || [ "$(go env GO111MODULE)" = 'on' ] || [ "$(go env GO111MODULE)" = 'auto' ] \
```
Madhu-1, line 39: can this be simplified by just checking the GO111MODULE  is off?

nixpanic: maybe, but we would also need to check if it is set to an invalid value in that case, I think? that would not simplify things much more

Madhu-1:
@humblec PTAL

humblec:
@nixpanic can you also document when this makefile has to be updated in the developer guide ?

nixpanic:
> @nixpanic can you also document when this makefile has to be updated in the developer guide ?

I'm not sure what you mean. Why would the makefile need to be updated?

humblec:
> > @nixpanic can you also document when this makefile has to be updated in the developer guide ?
> 
> I'm not sure what you mean. Why would the makefile need to be updated?

@nixpanic we check the version and fail in makefile target https://github.com/ceph/ceph-csi/pull/1005/files#diff-212af17ce32e5c0911abd264ca05d9fdR30 . What I was asking here is, when can we(developer) change the `go` version compatibility check to a different version and make a change in the make `target` ?

Approving this pr for now though.

nixpanic:
On Mon, May 11, 2020 at 11:35:25AM -0700, Humble Devassy Chirammal wrote:
> > > @nixpanic can you also document when this makefile has to be updated in the developer guide ?
> > 
> > I'm not sure what you mean. Why would the makefile need to be updated?
> 
> @nixpanic we check the version and fail in makefile target https://github.com/ceph/ceph-csi/pull/1005/files#diff-212af17ce32e5c0911abd264ca05d9fdR30 . What I was asking here is, when can we(developer) change the `go` version compatibility check to a different version and make a change in the make `target` ?

In case a newer version of Golang is needed, the scripts/check-env.sh
can be updated with the new requirement(s). This is not different for
Golang versions, or any other settings or tools. Everything that is
needed to build Ceph-CSI is now in the scripts/check-env.sh script.

We probably should add a check for libcephfs, librados and librbd header
files too. But that is a little more tricky to get right.

nixpanic:
@mergifyio refresh

nixpanic:
@mergifyio rebase

PR #1008, opened 2020-05-04
Doc: update development documentation

Madhu-1:
Updated golang version to 1.13.x and also updated user to set GO111MODULE=on and CGO_ENABLED=1 when doing development in cephcsi

### DIFF
--- a/docs/development-guide.md
+++ b/docs/development-guide.md
@@ -13,10 +13,14 @@ it is **highly** encouraged to:
 
 ### Workspace and repository setup
 
-* [Download](https://golang.org/dl/) Go (>=1.11.x) and
+* [Download](https://golang.org/dl/) Go (>=1.13.x) and
    [install](https://golang.org/doc/install) it on your system.
 * Setup the [GOPATH](http://www.g33knotes.org/2014/07/60-second-count-down-to-go.html)
    environment.
+* `CGO_ENABLED` is enabled by default, if `CGO_ENABLED` is set to `0` we need
+  to set it to `1` as we need to build with go-ceph bindings.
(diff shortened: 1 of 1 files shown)


docs/development-guide.md
```diff
@@ -13,10 +13,12 @@ it is **highly** encouraged to:
 
 ### Workspace and repository setup
 
-* [Download](https://golang.org/dl/) Go (>=1.11.x) and
+* [Download](https://golang.org/dl/) Go (>=1.13.x) and
    [install](https://golang.org/doc/install) it on your system.
 * Setup the [GOPATH](http://www.g33knotes.org/2014/07/60-second-count-down-to-go.html)
    environment.
+* Set `CGO_ENABLED=1` as we need to build with go-ceph bindings.
```
nixpanic, line 20: it should probably be mentioned that this is the default. Disabling with `CGO_ENABLED=0` should __not__ be done.

Madhu-1: as the user might have set it to false for other projects he is working on. explicitly asking the user to set it to CGO_ENABLED=1 seems correct, even overriding is not a problem if it is already set

nixpanic: I don't know, but it would be unusual to disable CGO. Have we ever encountered users with such an environment? Obviously using an old Makefile with newer sources can cause this problem, but that should be very uncommon.

Madhu-1: i have seen people building with CGO_ENABLED=0 in pure golang projects. let me update more details

Madhu-1: @nixpanic updated PTAL

Madhu-1:
@humblec PTAL

PR #1009, opened 2020-05-04
rbd: default nouuid if the formattype is xfs

Madhu-1:
The problem happens when multiple PVCs with the same UUID are attached/mounted on a node. This can happen after creating a PVC from a snapshot or cloning a PVC.

make nouuid as the default mount option if
the format type is xfs to avoid mounting
issues.

updates: #966


Madhu-1:
@humblec PTAL

Madhu-1:
@humblec PTAL

PR #1020, opened 2020-05-06
prepare: exit when passed incorrect parameters

Yuggupta27:
Now script does not continue execution
when an invalid argument is passed.

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -18,6 +18,12 @@ opts=$(getopt \
     -- "$@"
 )
 
+if [ $? -ne 0 ]
+then
+    echo "Try '--help' for more information." 
+    exit 1   
+fi
+
 eval set -- ${opts}
 
 while true; do


prepare.sh
```diff
@@ -49,6 +49,7 @@ while true; do
         ;;
     --)
         shift
+        exit 0
```
nixpanic, line 52: is this hit when parameters are passed that are not recognized? If so, please display an error and `exit 1`

make-containerized-build.groovy
```diff
@@ -28,7 +28,7 @@ node('cico-workspace') {
 	try {
 		stage('prepare bare-metal machine') {
 			sh 'scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ./prepare.sh root@${CICO_NODE}:'
-			sh 'ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./prepare.sh --workdir=/opt/build --gitrepo=${GIT_REPO} --branch=${GIT_BRANCH}'
+			sh 'ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./prepare.sh --workdir=tip/ --gitrepo=${GIT_REPO} --ref=pull/${ghprbPullId}/head'
```
nixpanic, line 31: use `/opt/build` as directory (or better, `/opt/build/go/src/github.com/ceph/ceph-csi`), otherwise it lands in some local place (probably `/root/tip`). If we want to use standard go tools, the directory needs to be in the `GOPATH` format.

nixpanic: There needs to be a check that `ghprbPullId` is available as environment  variable, if it is not, fall back to `--ref=master`.
Time to learn some Groovy :disappointed:

prepare.sh
```diff
@@ -31,7 +31,7 @@ while true; do
         echo "--gitrepo                 specify the git repository" 
         echo " "
         echo "Sample Usage:"
-        echo "./prepare.sh --gitrepo=https://github.com/example --workdir=/opt/build --ref=pull/123/head"
+        echo "./prepare.sh --gitrepo=https://github.com/example --workdir=tip/ --ref=pull/123/head"
```
nixpanic, line 34: this can be a separate commit. One for the corrections to the prepare.sh script, one for the update to the groovy job

Yuggupta27: Thanks, I'll make the required changes

Yuggupta27:
Done!

Issue #1023, opened 2020-05-07
Remove unmaintained kubernetes versions from testing


nixpanic:
Kubernetes maintains the latest minor version, and the two before it. That means with the release of Kubernetes 1.18, version 1.17 and 1.16 are still maintained. Deployments should not use unmaintained versions, so there is no need to test on older versions either.

See [supported versions in the k8s documentation](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions).

humblec:
@nixpanic are you working on the PR ?

PR #1024, opened 2020-05-07
Remove testing of unmaintained k8s versions

nixpanic:

Version 1.18.x is currently out, and the k8s community does not maintain
versions 1.14 and 1.15 anymore. No need to test on versions that users
should not deploy.

See-also: [Kubernetes Patch Releases](https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#timelines)

Fixes: #1023

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,20 +110,6 @@ jobs:
         - travis_terminate 0    # deploy only on x86
 
     - stage: e2e testing
-      name: cephcsi with kube 1.14.10
-      script:
-        - scripts/skip-doc-change.sh || travis_terminate 0;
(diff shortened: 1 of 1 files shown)


nixpanic:
Adding the do-not-merge label as Travis CI does not seem to display the jobs correctly. The feedback from the [Tarvis build](https://travis-ci.org/github/ceph/ceph-csi/builds/684221394) is not available in this PR either?

nixpanic:
Removing the [DNM] label as the adding of Kubernetes 1.18.2 did not work. [Humble also sent a PR](./1025) for that, so maybe we should go the known-to-work approach instead.

PR #1025, opened 2020-05-07
dependency: update kube version dependency to v1.18.2

humblec:


nixpanic:
I have included this in the update for CI jobs with PR #1024

humblec:
> I have included this in the update for CI jobs with PR #1024

You mean the travis/CI part , Isnt it?

humblec:
@nixpanic any idea why travis didnt run on this PR ?

nixpanic:
> I have included this in the update for CI jobs with PR #1024

I've dropped the adding of 1.16.9 and 1.18.2 versions from that PR. Lets get them added through this one (as #1024 also addresses the removal).

nixpanic:
Can you add the `go mod ...` commands that you used to update this in the commit message please?

nixpanic:
> I've dropped the adding of 1.16.9 and 1.18.2 versions from that PR. Lets get them added through this one (as #1024 also addresses the removal).

Ah, sorry, I thought you found issue #1023 and send this PR to test against kubernetes 1.18.2. Now I see that this is something else :+1:

nixpanic:
> if there is any bug fix that is required we can update it to a minor version.

I agree, if there is no need for us to update, why all the churn in the repo? The commit message (and PR) should explain what bugs are fixed that affect the cephcsi executable.

humblec:
> is there any bug fix in this version that is required for us? or else it's better to wait for 1.19 and than update it.
> can we update the vendor only for kube major version? if there is any bug fix that is required we can update it to a minor version.

@Madhu-1 @nixpanic I dont really get the logic. :(. that said, we are not in parity or in the release schedule of Kubernetes, so updating only when the major version is available does not look correct to me.  The minor or patch release are in place with some bug fixes, enhancements..etc. so updating and catching up with what's available is actually a good idea. More or less, Kube ensures the backward compatibility, and also it's not going to break the things in a minor version for sure. We do have e2e ..etc in place to double confirm. can you let me know whats the real concern here ? is it about breakage of the code? or something else?

nixpanic:
The main question that needs answering is:

> Why update `vendor/` to a new Kubernetes version.

If the new version fixes bugs in functions/features that we build into the cephcsi executable, it is of course a good thing. If there are no bugs fixed in the API that we consumer, there is no need to spend time on rebasing, testing and reviewing.

I agree that in general it is good to update to a new version, but I also hope it is done with a good reason. That reason should be explained in the PR and commit message(s).

Madhu-1:
> > is there any bug fix in this version that is required for us? or else it's better to wait for 1.19 and than update it.
> > can we update the vendor only for kube major version? if there is any bug fix that is required we can update it to a minor version.
> 
> @Madhu-1 @nixpanic I dont really get the logic. :(. that said, we are not in parity or in the release schedule of Kubernetes, so updating only when the major version is available does not look correct to me. The minor or patch release are in place with some bug fixes, enhancements..etc. so updating and catching up with what's available is actually a good idea. More or less, Kube ensures the backward compatibility, and also it's not going to break the things in a minor version for sure. We do have e2e ..etc in place to double confirm. can you let me know whats the real concern here ? is it about breakage of the code? or something else?

IMO we don't want to keep updating the vendor directory for each patch release at least. we can update the vendor directory if there is a major release or if there is a bug fix which we need

Madhu-1:
suppressed by https://github.com/ceph/ceph-csi/pull/1380

PR #1026, opened 2020-05-07
ci: add configuration for "Semantic Pull Request" bot

nixpanic:

The "Semantic Pull Request" bot has been enabled as replacement for [https://github.com/z0al/commitlint-bot](commitlint-bot) that is not being maintained anymore. It requires some configuration before it can be practically used with Ceph-CSI.

This configuration enforces the format of commits that are part of a
Pull Request. The title of the PR can be checked as well, that will not
be required for now.

The format of a subject of all commit messages needs to be:

    <component>: <description>

For example:

    ci: add configuration for "Semantic Pull Request" bot

The valid components in the commit message are listed in the
.github/semantic.yaml file as 'types'.

See-also: https://github.com/zeke/semantic-pull-requests#semantic-pull-requests
See-also: https://www.conventionalcommits.org

### DIFF
--- a/.github/semantic.yaml
+++ b/.github/semantic.yaml
@@ -1,0 +1,44 @@
+---
+# Always validate the PR title, and ignore the commits
+# titleOnly: true
+
+# Always validate all commits, and ignore the PR title
+commitsOnly: true
+
+# Always validate the PR title AND all the commits
+# titleAndCommits: true
+
+# Require at least one commit to be valid
+# this is only relevant when using commitsOnly: true or titleAndCommits: true,
+# which validate all commits by default
+# anyCommit: true
+
+# Allow use of Merge commits (eg on github: "Merge branch 'master' into
+# feature/ride-unicorns")
+# this is only relevant when using commitsOnly: true (or titleAndCommits: true)
+allowMergeCommits: false
+
+##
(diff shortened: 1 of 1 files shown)


.github/semantic.yaml
```diff
@@ -0,0 +1,44 @@
+---
+# Always validate the PR title, and ignore the commits
```
Madhu-1, line 2: can you add a source for this file that you referred to write this configuration?

.github/semantic.yaml
```diff
@@ -0,0 +1,44 @@
...
+# You can define a list of valid scopes
+# scopes:
+#  - scope1
+#  - scope2
+
+# By default types specified in commitizen/conventional-commit-types is used.
+##
+## We override the types by components we have.
+##
+types:
```
Madhu-1, line 33: need to update the development guide for this one.

nixpanic: yes, but I prefer to get this in before, so that all contributors have some standard to follow.

.github/semantic.yaml
nixpanic: It is listed in the commit message. There is no full example (or I did not find it), only snippets on the site.

PR #1028, opened 2020-05-08
travis: test functionality of travis with github

### LINKED ISSUE #1011: GitHub should not report all tests as passed when parsing .travis.yaml fails
# Describe the bug #

When there is an error in the `.travis.yaml` file, Travis CI does not run. The checks in the PR are green, as everything (mergify only) passed. There should be a check in the mergify configuration that highlights the not-passing of Travis CI.

agarwal-mudit:
Github considers all the checks as passed even in the case when there is an error
in travis.yml and ci doesn't run. This results in auto merging of the code without a valid
CI run, trying to repro the same by introducing a parsing error in travis.yml.
Please not, that this is a just a repro attempt, it doesn't fix #1011

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
 services:
   - docker
 
-language: go
+language:go
 branches:
   only:
     - master


.travis.yml
```diff
@@ -12,7 +12,7 @@ addons:
 services:
   - docker
 
-language: go
+language:go
```
nixpanic, line 15: This will break Travis CI, it will not be able to parse this YAML file :+1:

Madhu-1:
@nixpanic any comment on this one? can this be closed i don't think we can do something in travis.yaml file to  fix this issue. if its a configuration in  GitHub UI only @humblec can do it, he is the only person who is having permission to do it

nixpanic:
> if its a configuration in GitHub UI only @humblec can do it, he is the only person who is having permission to do it

We need to know if this is the case.
If there is no such option, it could be reported (if there is no such issue yet) to the Travis CI community.

Madhu-1:
> > if its a configuration in GitHub UI only @humblec can do it, he is the only person who is having permission to do it
> 
> We need to know if this is the case.
> If there is no such option, it could be reported (if there is no such issue yet) to the Travis CI community.

agree! do we need to keep open this PR or we can discuss more on the issue? @nixpanic @agarwal-mudit how about opening an issue in https://travis-ci.community/ and take it forward?

nixpanic:
> agree! do we need to keep open this PR or we can discuss more on the issue? @nixpanic @agarwal-mudit how about opening an issue in https://travis-ci.community/ and take it forward?

Yes, and lets track the state in #1011

Madhu-1:
@agarwal-mudit closing this one, to track the status in #1011. Thanks for testing this one.

PR #1032, opened 2020-05-11
cleanup: use switch/case for genVolFromVolID() error checking

nixpanic:

By using switch/case it is easier to follow the error checking of the
genVolFromVolID() function. In case a new error is added as a return of
the function, it will be simpler to add checking for it.

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -395,14 +395,14 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 
 	rbdVol, err := genVolFromVolID(ctx, volumeID, cr, req.GetSecrets())
 	if err != nil {
-		if _, ok := err.(util.ErrPoolNotFound); ok {
+		switch err.(type) {
+		case util.ErrPoolNotFound:
 			klog.Warningf(util.Log(ctx, "failed to get backend volume for %s: %v"), volumeID, err)
 			return &csi.DeleteVolumeResponse{}, nil
-		}
 
 		// If error is ErrInvalidVolID it could be a version 1.0.0 or lower volume, attempt
 		// to process it as such
-		if _, ok := err.(ErrInvalidVolID); ok {
+		case ErrInvalidVolID:
 			if isLegacyVolumeID(volumeID) {
 				klog.V(2).Infof(util.Log(ctx, "attempting deletion of potential legacy volume (%s)"), volumeID)
 				return cs.DeleteLegacyVolume(ctx, req, cr)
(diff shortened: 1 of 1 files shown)


internal/rbd/controllerserver.go
```diff
@@ -786,16 +784,15 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
 			return nil, status.Errorf(codes.NotFound, "volume ID %s not found", volID)
-		}
-
-		if _, ok := err.(util.ErrPoolNotFound); ok {
+		case *util.ErrPoolNotFound:
 			klog.Errorf(util.Log(ctx, "failed to get backend volume for %s: %v"), volID, err)
 			return nil, status.Errorf(codes.InvalidArgument, err.Error())
+		default:
+			return nil, status.Errorf(codes.Internal, err.Error())
 		}
```
Madhu-1, line 795:
```suggestion
switch err.(type) {
		case *ErrImageNotFound:
			err=status.Errorf(codes.NotFound, "source Volume ID %s not found", req.GetSourceVolumeId())
		case *util.ErrPoolNotFound:
			klog.Errorf(util.Log(ctx, "failed to get backend volume for %s: %v"), req.GetSourceVolumeId(), err)
			fallthrough
		default:
			err=status.Errorf(codes.Internal, err.Error())
		}
		return nil,err
```

Madhu-1: this can be applied to https://github.com/ceph/ceph-csi/pull/1032/files#diff-f010f466063f62f823361f972d08b554R513 to avoid multiple returns

internal/rbd/controllerserver.go
```diff
@@ -395,33 +395,31 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
...
 		// success as deletion is complete
-		if _, ok := err.(util.ErrKeyNotFound); ok {
+		case *util.ErrKeyNotFound:
 			klog.Warningf(util.Log(ctx, "Failed to volume options for %s: %v"), volumeID, err)
 			return &csi.DeleteVolumeResponse{}, nil
-		}
 
 		// All errors other than ErrImageNotFound should return an error back to the caller
-		if _, ok := err.(ErrImageNotFound); !ok {
+		default:
```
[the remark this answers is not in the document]
nixpanic: hmm, I did see the exception for `ErrImageNotFound` but somehow did not save that change... I'll make sure to add it!

internal/rbd/controllerserver.go
nixpanic: Sure, I do not like all those return statements either :+1:

nixpanic: Except:
```
internal/rbd/controllerserver.go:520:4: cannot fallthrough in type switch
```

So, no `fallthrough` but multiple `err` assignments.

Madhu-1: Yeah as you are comparing with type the fall through won't work, we can remove it

Madhu-1:
@nixpanic  CI failure

nixpanic:
> @nixpanic CI failure

Yeah, I am not sure where that error comes from... Trying to track it down :eyes:

nixpanic:
@Madhu-1 this is ready for review, the `switch err.(type)` has been corrected to not check for a pointer to a type :man_facepalming:

internal/rbd/controllerserver.go
```diff
@@ -786,16 +787,15 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 
 	rbdVol, err := genVolFromVolID(ctx, volID, cr, req.GetSecrets())
 	if err != nil {
-		if _, ok := err.(ErrImageNotFound); ok {
+		switch err.(type) {
+		case ErrImageNotFound:
 			return nil, status.Errorf(codes.NotFound, "volume ID %s not found", volID)
```
Madhu-1, line 792: @nixpanic can you avoid return inside each case here? same as you did above

nixpanic: sure!

PR #1033, opened 2020-05-11
reduce default worker-threads for csi-provisioner

levindecaro:
Have various testing on small and medium size cluster,  default worker-threads=100 will cause multiple grpc errors from API during PVC dynamic provisioning, as the result PVC creation randomly spent 15 to 60 mins. Reducing worker-threads helps eliminate the API errors.

Issue #921, opened 2020-04-07
resource usage with forks


kfox1111:
In the csi-rbdplugin container in the provisioner pod, we see large numbers of processes forking off (rados & ceph) during provisioning. This makes it difficult to set resource limits on the pods to grantee availability. We need some mechanism to guarantee only a certain number of children will be running at a time so we can control the resource usage.

nixpanic:
When we move to go-ceph for provisioning, the number of execs to external commands (like `rados` and `rbd`) will be minimized if not completely removed. This reduces not only the exec to the specific command, but also the spawning of threads that a Ceph client does when connecting to a Ceph cluster. With go-ceph we will re-use existing connection (for volume management, not I/O), so resource consumption will change drastically when running many operations on PV(C)s.

kfox1111:
We're still running hard into this issue. https://github.com/ceph/ceph-csi/projects/3 makes it seem like the go based fix is still a long time out. How do we fix this in the mean time?

Can we limit the number of grpc calls it takes a time? That may help.

kfox1111:
Looks like the provisioner supports a flag:
--worker-threads <num>: Number of simultaneously running ControllerCreateVolume and ControllerDeleteVolume operations. Default value is 100.

This value is WAY too high with forking. But setting it lower should fix the problem.

Currently the helm chart does not allow this to be set. Please can we add support for that?

Thanks!

kfox1111:
I can confirm on my cluster once I patched in --worker-threads=4 the cluster became reliable.

kfox1111:
There may be other --worker-threads options in other sidecars. We should set them all for better reliability.

nixpanic:
There are also PRs #1033 and #1034 that add `--worker-threads`. In one of the PRs there is a discussion/question about the ideal value for this, and how testing was done.

PR #1034, opened 2020-05-11
reduce default worker-threads for csi-provisioner

levindecaro:
Have various testing on small and medium size cluster,  default worker-threads=100 will cause multiple grpc errors from API during PVC dynamic provisioning, as the result PVC creation randomly spent 15 to 60 mins. Reducing worker-threads helps eliminate the API errors.

### DIFF
--- a/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
+++ b/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
@@ -41,6 +41,7 @@ spec:
             - "--leader-election-type=leases"
             - "--retry-interval-start=500ms"
             - "--feature-gates=Topology=true"
+            - "--worker-threads=5"
           env:
             - name: ADDRESS
               value: unix:///csi/csi-provisioner.sock


humblec:
@levindecaro I am trying to understand whats the error you are getting and how the `size` of the cluster have an effect on number of workers here. Can you please share more details on the issue?

levindecaro:
> @levindecaro I am trying to understand whats the error you are getting and how the `size` of the cluster have an effect on number of workers here. Can you please share more details on the issue?

@humblec I was tested the ceph-csi on OCP4.3 and 4.4, the size of cluster described as below

OCP 4.3
- 3 x Master RHCOS (6 CPU 32GB Memory per node)
- 10 x Worker RHEL ( 8 CPU 16GB Memory per node)
about 300 Pods, 100 deployment, 30 active PV

OCP 4.4
- 3 x Master RHCOS (4 CPU 16GB Memory per node)
- 2 x Worker RHCOS ( 8 CPU 16GB Memory per node)
clean install


The dynamic provisioning works out of the box without any tuning on both cluster. However when application deployment comes in, and cluster starting to growth with adding hundred deployment objects,  then the ceph-csi will have trouble to finish the PVC bounding randomly with the message as below.

```
Generated from cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-5ff845849c-fmqvq_41da825a-fa8f-467b-8396-7c9d9b895025
14 times in the last 30 minutes
failed to provision volume with StorageClass "csi-cephfs-sc": rpc error: code = Aborted desc = an operation with the given Volume ID pvc-c8a1e41b-0b3e-4653-9d5b-872b5fc92b6f already exists
```

When i received this message and checked the underlying cephfs, the subvolume actually was created already at the first attempt (matching the provisioning time). So I thought there are somehow race issues from the csi provisioiner i can't tell, even with debug logs, they are just flooding with same "already exists" errors. If u have patient enough, it will finish the PVC bound after mins or a hour or two.

On the other hands, the etcd on OCP are often come up "took too long to execute > 100ms" message after CSI comes in, it was indication that relate to performance problem where APIs are not fast enough to catch up the operation.

In this connection, I start to stress the csi pv provisioning to create thousand of PVC sequentially with 1 PV per second rate, however both cluster appear stuck creation at about 300 ~450 PVC. When it stuck at this point, it will never resume until I flush all pending PVC. Quickly look at the csi-provisioner  usage are climb up to 8GB memory usage and they all flooding with grpc errors.

Finally, I tried to reduce the worker threads down to somehow at 5, thousand PVC can be created without any errors, and ran so far stable after 1 week observation.

Madhu-1:
@levindecaro let's track the changes in one PR either here on in #1033

Madhu-1:
one thing to consider here is with current implementation cephcsi it will open a new connection to connect to the backend ceph cluster and this will be solved by the go-ceph binding and I believe your issues should also be solved by it, @nixpanic what do you say?

nixpanic:
> one thing to consider here is with current implementation cephcsi it will open a new connection to connect to the backend ceph cluster and this will be solved by the go-ceph binding and I believe your issues should also be solved by it, @nixpanic what do you say?

Yes, most likely. that is why I referenced it from #921

humblec:
@levindecaro first of all thanks a lot for the detailed comment. I really appreciate it!
 Some details/thoughts inline.



> @humblec I was tested the ceph-csi on OCP4.3 and 4.4, the size of cluster described as below
> 
> OCP 4.3
> 
>     * 3 x Master RHCOS (6 CPU 32GB Memory per node)
> 
>     * 10 x Worker RHEL ( 8 CPU 16GB Memory per node)
>       about 300 Pods, 100 deployment, 30 active PV
> 
> 
> OCP 4.4
> 
>     * 3 x Master RHCOS (4 CPU 16GB Memory per node)
> 
>     * 2 x Worker RHCOS ( 8 CPU 16GB Memory per node)
>       clean install
> 
> 
> The dynamic provisioning works out of the box without any tuning on both cluster. However when application deployment comes in, and cluster starting to growth with adding hundred deployment objects, then the ceph-csi will have trouble to finish the PVC bounding randomly with the message as below.

Am I correct in the assumption that this is the same behaviour in both of the above clusters ?  like OCP 4.3 and OCP 4.4 ?

> 
> ```
> Generated from cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-5ff845849c-fmqvq_41da825a-fa8f-467b-8396-7c9d9b895025
> 14 times in the last 30 minutes
> failed to provision volume with StorageClass "csi-cephfs-sc": rpc error: code = Aborted desc = an operation with the given Volume ID pvc-c8a1e41b-0b3e-4653-9d5b-872b5fc92b6f already exists
> ```
> 
> When i received this message and checked the underlying cephfs, the subvolume actually was created already at the first attempt (matching the provisioning time). So I thought there are somehow race issues from the csi provisioiner i can't tell, even with debug logs, they are just flooding with same "already exists" errors. If u have patient enough, it will finish the PVC bound after mins or a hour or two.
> 

I am trying to figure out mainly 2 things here:

1) Is it specific to cephfs or is it the same behaviour for both RBD and CephFS? because it may be some 'lock' contention in the cephfs layer too which cause the delay in processing further requests. If there is a lock contention or some improvement required, we have to isolate that.
` it will finish the PVC bound after mins or a hour or two.` -> this is not good , taking an hour or two to finish the request is too much.

@Batrick  @dillaman @kotresh @vshankar any thoughts here on what could be causing such behaviour? 

2) If the error is "already exist", I am wondering why the next retry goes again.  It should be the case or is it that, its just a random logging .. not sure atm. 
Btw, are you confirming this is the behaviour when there are around 1000 requests issued in one shot? 



> On the other hands, the etcd on OCP are often come up "took too long to execute > 100ms" message after CSI comes in, it was indication that relate to performance problem where APIs are not fast enough to catch up the operation.
> 

There is already an issue in sidecar shipped by kube where it goes on toss if there are more than 100 requests in a shot.  There were some improvements, but maybe not up to the mark.

> In this connection, I start to stress the csi pv provisioning to create thousand of PVC sequentially with 1 PV per second rate, however both cluster appear stuck creation at about 300 ~450 PVC. When it stuck at this point, it will never resume until I flush all pending PVC. Quickly look at the csi-provisioner usage are climb up to 8GB memory usage and they all flooding with grpc errors.
> 
> Finally, I tried to reduce the worker threads down to somehow at 5, thousand PVC can be created without any errors, and ran so far stable after 1 week observation.

Reducing worker threads will obviously help the backend layer, but there is also an expectation  that, 
we should be capable of provisioning or serving good number of parellel requests. So isolating the issue to the layer could help. 

Btw, in this scenario increasing the `csiTimeout` to a higher value could also help I believe. 
say for ex: 300 , atleast it should reduce the load to a good extend.
https://github.com/ceph/ceph-csi/blob/master/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml#L56

levindecaro:
> @levindecaro first of all thanks a lot for the detailed comment. I really appreciate it!
> Some details/thoughts inline.
> 
> > @humblec I was tested the ceph-csi on OCP4.3 and 4.4, the size of cluster described as below
> > OCP 4.3
> > ```
> > * 3 x Master RHCOS (6 CPU 32GB Memory per node)
> > 
> > * 10 x Worker RHEL ( 8 CPU 16GB Memory per node)
> >   about 300 Pods, 100 deployment, 30 active PV
> > ```
> > 
> > 
> > OCP 4.4
> > ```
> > * 3 x Master RHCOS (4 CPU 16GB Memory per node)
> > 
> > * 2 x Worker RHCOS ( 8 CPU 16GB Memory per node)
> >   clean install
> > ```
> > 
> > 
> > The dynamic provisioning works out of the box without any tuning on both cluster. However when application deployment comes in, and cluster starting to growth with adding hundred deployment objects, then the ceph-csi will have trouble to finish the PVC bounding randomly with the message as below.
> 
> Am I correct in the assumption that this is the same behaviour in both of the above clusters ? like OCP 4.3 and OCP 4.4 ?

Yes, same behaviour, same error.

> 
> > ```
> > Generated from cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-5ff845849c-fmqvq_41da825a-fa8f-467b-8396-7c9d9b895025
> > 14 times in the last 30 minutes
> > failed to provision volume with StorageClass "csi-cephfs-sc": rpc error: code = Aborted desc = an operation with the given Volume ID pvc-c8a1e41b-0b3e-4653-9d5b-872b5fc92b6f already exists
> > ```
> > 
> > 
> > When i received this message and checked the underlying cephfs, the subvolume actually was created already at the first attempt (matching the provisioning time). So I thought there are somehow race issues from the csi provisioiner i can't tell, even with debug logs, they are just flooding with same "already exists" errors. If u have patient enough, it will finish the PVC bound after mins or a hour or two.
> 
> I am trying to figure out mainly 2 things here:
> 
> 1. Is it specific to cephfs or is it the same behaviour for both RBD and CephFS? because it may be some 'lock' contention in the cephfs layer too which cause the delay in processing further requests. If there is a lock contention or some improvement required, we have to isolate that.
>    ` it will finish the PVC bound after mins or a hour or two.` -> this is not good , taking an hour or two to finish the request is too much.

Yes, I have experience same "already exist" issue for rbd as well.

> 
> @batrick @dillaman @kotresh @vshankar any thoughts here on what could be causing such behaviour?
> 
> 1. If the error is "already exist", I am wondering why the next retry goes again.  It should be the case or is it that, its just a random logging .. not sure atm.
>    Btw, are you confirming this is the behaviour when there are around 1000 requests issued in one shot?

This behaviour can be reproduced for large volume PV request, if just test for few pv, it won't always hit that.

> 
> > On the other hands, the etcd on OCP are often come up "took too long to execute > 100ms" message after CSI comes in, it was indication that relate to performance problem where APIs are not fast enough to catch up the operation.
> 
> There is already an issue in sidecar shipped by kube where it goes on toss if there are more than 100 requests in a shot. There were some improvements, but maybe not up to the mark.
> 
> > In this connection, I start to stress the csi pv provisioning to create thousand of PVC sequentially with 1 PV per second rate, however both cluster appear stuck creation at about 300 ~450 PVC. When it stuck at this point, it will never resume until I flush all pending PVC. Quickly look at the csi-provisioner usage are climb up to 8GB memory usage and they all flooding with grpc errors.
> > Finally, I tried to reduce the worker threads down to somehow at 5, thousand PVC can be created without any errors, and ran so far stable after 1 week observation.
> 
> Reducing worker threads will obviously help the backend layer, but there is also an expectation that,
> we should be capable of provisioning or serving good number of parellel requests. So isolating the issue to the layer could help.
> 
> Btw, in this scenario increasing the `csiTimeout` to a higher value could also help I believe.
> say for ex: 300 , atleast it should reduce the load to a good extend.
> https://github.com/ceph/ceph-csi/blob/master/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml#L56

csiTimeout do nothing with this behaviour, I had tired with different values before get into worker-threads.

kfox1111:
I ran it on a heavily loaded cluster and needed to set resource requests/limits in order to schedule the workload and guarantee resources for the driver. Even with significant amounts of dedicated resource to the driver, it was failing when a user created a cluster with an operator that spawned 32 pvc's at the same time. I tried lots of things and when I found the worker-threads flag, set it to 4, and all the problems went away. Its been solid since.

I don't know if it is the right default value. It probably depends on what you set resources/limits to. But 100 is DEFINITELY too much in a heavy load environment. There may be some things we could do with setting timeouts on the sidecar grpc calls as well. When it was under heavy load, ceph calls were taking a while to resolve, causing grpc timeouts and then executing more grpc calls.

It sounds like the csi sidecar may start defaulting to worker-threads 2 in the near future as well as most drivers don't know what to set it to and smaller is safer then bigger.

Defaulting to 2 or 4 should be fine. It may be slow to create, but should always work. Then the user can tune up the number if they choose.

humblec:
@levindecaro and @kfox1111 thanks for the detailed testing. These are also a proof that, we have better reliability and response when we have small worker threads. I am **Ack**ing for reducing the `worker-threads to 4` for now.  @levindecaro can you revisit  https://github.com/ceph/ceph-csi/pull/1033/ ? please let me know if you need any help!

kfox1111:
Did another fix get in for this?

kfox1111:
I think this has been resolved by a golang based library implementation rather then forking?

Madhu-1:
@kfox1111 @levindecaro This should be fixed in cephcsi master now, instead of using CLI we are using the go-ceph library to talk to ceph cluster. please let us know if we still need this fix?

kfox1111:
I think it being in master is good enough to close this. Thanks! :)

kfox1111:
do you know off the top of your head if there is a version released from master after the go-ceph stuff was added? something like, if your running after xxxx then you have it? Would help me not have to hunt through change logs. if you don't know, thats ok.

Madhu-1:
Closing this one as there is no much work. Please feel free to reopen and rebase.

PR #1029, opened 2020-05-08
ci: check for empty ghprbPullId 

Yuggupta27:
The script checks for ghprbPullId env variable
and if found unset, runs the script for master
to support regular builds.

Signed-off-by: Yug Gupta <ygupta@redhat.com>

PR #1037, opened 2020-05-11
ci: pass ghprPullId in make-containerized-build job

nixpanic:

In case the environment variable is not set at all (as opposed to
empty), the following error is reported while running the job:

    groovy.lang.MissingPropertyException: No such property: ghprbPullId for class: groovy.lang.Binding

Also, add a cron schedule to check every 5 minutes for a new PR to run the
make-containerized-build job.

## Related issues ##

Updates: #1029


nixpanic:
@Yuggupta27 could you please review this change?

PR #1036, opened 2020-05-11
doc: add list of valid components to the development guide

nixpanic:
# Describe what this PR does #

The commitlint CI job uses the configuration from .commitlintrc.yaml
which contains the different components that Ceph-CSI uses. A short
description of each component has been added, so that contributors
understand what component to mention in the prefix of the subject in
commit messages.

## Related issues ##

See-also: #1058

PR #1039, opened 2020-05-11
journal: remove SetCSIDirectorySuffix to simplify journal creation

phlogistonjohn:

The function SetCSIDirectorySuffix was used only one per (long-lived,
gloabl) journal object. It is simpler to construct the journal objects
with this needed parameter:
1. As it is required to function and non-optional AFAICT
2. Removes the temptation to mutate global object
3. Reduces LOC with exact same functionality
4. SetCSIDirectorySuffix would not behave correctly if called a 2nd time
   anyway.

Point 4. means that if you called the function twice to change the
suffix when you previously had "csi.volumes.alice", you'd get
"csi.volumes.alice.bob" instead of "csi.volumes.bob" what one would
expect.

This is partly done in preparation to make the journal get/set/or-otherwise-make-use-of go-ceph for omap.

Is the change backward compatible? yes - refactoring only

Are there concerns around backward compatibility? n/a

### DIFF
--- a/internal/cephfs/driver.go
+++ b/internal/cephfs/driver.go
@@ -109,11 +109,8 @@ func (fs *Driver) Run(conf *util.Config, cachePersister util.CachePersister) {
 	if conf.InstanceID != "" {
 		CSIInstanceID = conf.InstanceID
 	}
-	// Get an instance of the volume journal
-	volJournal = journal.NewCSIVolumeJournal()
-
-	// Update keys with CSI instance suffix
-	volJournal.SetCSIDirectorySuffix(CSIInstanceID)
+	// Create an instance of the volume journal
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
I can't tell what this "Semantic Pull Request" thingie doesn't like about my commit. I checked the updated guide in PR #1036  and I still don't see the issue. It's very unfortunate that it doesn't produce anything useful when you use the "Details" link.

Madhu-1:
@nixpanic @humblec why Semantic Bot is failing on this PR any idea? 

Sematic Bot is not mandatory for PR to  be merged at least for now in cephcsi

nixpanic:
> @nixpanic @humblec why Semantic Bot is failing on this PR any idea?

I don't know, I have the same problem with some of the other PRs. No idea why the bot does not accept the commit messages :worried:

PR #1045, opened 2020-05-12
Revert "cephfs: use new 'ceph fs resize' command when available"

humblec:
Reverts ceph/ceph-csi#1044


humblec:
@Madhu-1 PTAL.

humblec:
@nixpanic #1044 accidently got merged which could update the 2.1 image with the change. Ideally it would have been on `2.1.1-canary`. But the `release-2.1` havent fall back to `canary` yet. so reverting the merge and getting back to the state which was previously with this PR. Please see #1044

PR #1031, opened 2020-05-08
rbd: use go-ceph to implement getImageInfo()

nixpanic:
Use the go-ceph language bindings for librbd to get the information about an RBD image.

This will remove an other execution of the `rbd` command and use the pooled cluster connection.

go-ceph is updated to v0.3.0 as it adds support for parsing feature bits of an image to user understandable strings.

PR #1052, opened 2020-05-12
rbd: use rbdVolume.open() for metadata operations

nixpanic:

Simple cleanup to remove redundant implementation of opening an RBD image with go-ceph.

## Related issues ##

The `rbdVolume.open()` method also added with PR #1031


internal/rbd/rbd_util.go
```diff
@@ -200,6 +200,20 @@ func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 	return nil
 }
 
+// Open the rbdVolume after it has been connected.
+func (rv *rbdVolume) open() (*librbd.Image, error) {
+	ioctx, err := rv.conn.GetIoctx(rv.Pool)
+	if err != nil {
+		return nil, ErrImageNotFound{rv.Pool, err}
```
Madhu-1, line 207: this seems not correct right?, due to some internal error, we are returning ImageNotFound.

internal/rbd/rbd_util.go
```diff
@@ -200,6 +200,20 @@ func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 	return nil
 }
 
+// Open the rbdVolume after it has been connected.
```
Madhu-1, line 203:
```suggestion
open the rbdVolume after it has been connected.
```

internal/rbd/rbd_util.go
nixpanic: Maybe, it should probably make a distinction between PoolNotFound and some PermissionError. The connection is already available, and that requires some authentication too. I'll update it to return different errors.

nixpanic:
@Madhu-1 please review this again, the changes you requested/suggested have been made.

Madhu-1:
need to remove https://github.com/ceph/ceph-csi/blob/master/e2e/utils.go#L369 and make use of exported clusterinfo in e2e as well

nixpanic:
> need to remove https://github.com/ceph/ceph-csi/blob/master/e2e/utils.go#L369 and make use of exported clusterinfo in e2e as well

That was probably for a different PR?

Madhu-1:
> > need to remove https://github.com/ceph/ceph-csi/blob/master/e2e/utils.go#L369 and make use of exported clusterinfo in e2e as well
> 
> That was probably for a different PR?

:facepalm: yes this comment is for  different PR

nixpanic:
> /retest

Yay, this works!

internal/util/connection.go
```diff
@@ -71,7 +71,13 @@ func (cc *ClusterConnection) GetIoctx(pool string) (*rados.IOContext, error) {
 
 	ioctx, err := cc.conn.OpenIOContext(pool)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to open IOContext for pool %s", pool)
+		// ErrNotFound indicates the Pool was not found
+		if err == rados.ErrNotFound {
```
humblec, line 75: @nixpanic are there any references available to the `rados` error codes ?

humblec: If yes, it would be good to mention here.

nixpanic: Not much, they are just `errno` values converted to Golang errors. The go-ceph functions do not specify what errors can be returned, as that depends directly on the Ceph libraries (and these do not document it either).

humblec: @nixpanic sure, I was confused by this ` 		// ErrNotFound indicates the Pool was not found ` . Is it true that, `ErrNotFound` is the error code  just for `Pool was not found`, If that's the case , I am wondering why it was not used as `ErrPoolNotFound` in first place?

nixpanic: `ErrNotFound` is a translation of the errno `ENOENT` and can indicate any missing resource. When opening a pool, a  return of `ENOENT` is most likely referring to the non-existence of the pool. There is no `ErrPoolNotFound` in go-ceph/rados/errors.go.

humblec: @nixpanic oh.. my understanding was that it could be any failure like `osds` are down or  `PG related` or `node down`..etc. Isn't it the case?

nixpanic: in that case there would be some connection error

humblec: @nixpanic so you are confirming we wont get into the openContext at all or we would have failed early, Isn't it?

PR #1053, opened 2020-05-13
Disable semantic bot from the repo

humblec:


humblec:
@nixpanic @Madhu-1 ptal.

Madhu-1:
@humblec how about just disabling the bot from this now for now instead of removing the configuration? we have discussed in yesterday's meeting that we will temporarily disable the bot till we understand how it works.

humblec:
> @humblec how about just disabling the bot from this now for now instead of removing the configuration? we have discussed in yesterday's meeting that we will temporarily disable the bot till we understand how it works.

This is a way to disable the bot. More or less, we don't need any extra configuration in the repo if we are not using it.

Madhu-1:
> we don't need any extra configuration in the repo if we are not using it.

if we remove this configuration will bot be removed? i was thinking it need  to be removed from github.com/apps

nixpanic:
> @nixpanic @Madhu-1 ptal.

Please add me as a reviewer, so GitHub can sort/list the PRs that explicitly have my review request. I'll review other PRs too, but with lower priority.

humblec:
One thing which I don't understand here is this:

There are 2 options:

1) Remove the bot from repo 
2) Remove the configuration file from repo to disable the bot and keep bot integrated till we have some checkpoint. 

if we go with option 1, what are the choices ?  keep configuration file for a removed bot ? 
experiment with configuration file without bot ? then bring it back once we know the right configuration for removed bot ? how long are we going to experiment ?  

if we go with option 2, bot is disabled or not acting at all . If we confirmed to use some other bot, its just matter of removing this bot and bringing the next one. If its about enabling it again with correct configuration, we can still experiment with bot being here.  

That's why this  PR. 
Does it make sense?

@Madhu-1 @nixpanic

nixpanic:
> Remove the configuration file from repo to disable the bot and keep bot integrated till we have some checkpoint.

This does not disable the bot, but have it fallback to its default configuration. This is definitely not what we want.

We currently have the `commitlint` bot as well, and that should be replaced in the near future. It seems the Semantic Pull Request bot is not a suitable replacement for us (so that one can be removed).

humblec:
> > Remove the configuration file from repo to disable the bot and keep bot integrated till we have some checkpoint.
> 
> This does not disable the bot, but have it fallback to its default configuration. This is definitely not what we want.

> 
> We currently have the `commitlint` bot as well, and that should be replaced in the near future. It seems the Semantic Pull Request bot is not a suitable replacement for us (so that one can be removed).

Sure, even in that case, Its 2 step process:

1) Removing its configuration from this repo ( this is what the PR does - so this is good to go? )
2) Removing it from the repo ( I am on it)

Madhu-1:
> > > Remove the configuration file from repo to disable the bot and keep bot integrated till we have some checkpoint.
> > 
> > 
> > This does not disable the bot, but have it fallback to its default configuration. This is definitely not what we want.
> 
> > We currently have the `commitlint` bot as well, and that should be replaced in the near future. It seems the Semantic Pull Request bot is not a suitable replacement for us (so that one can be removed).
> 
> Sure, even in that case, Its 2 step process:
> 
> 1. Removing its configuration from this repo ( this is what the PR does - so this is good to go? )
> 2. Removing it from the repo ( I am on it)

are we removing the configuration? if yes no need to do experiment and add it back.

Madhu-1:
will merge this one once we remove the Sematic bot from repo

humblec:
Yeah removing the bot. We will find alternative later and go with that.

nixpanic:
https://commitlint.js.org/#/guides-ci-setup seems like an easy replacement, using a travis job and custom configuration file.

humblec:
> https://commitlint.js.org/#/guides-ci-setup seems like an easy replacement, using a travis job and custom configuration file.

sure.. thanks for the pointer.. let me go through the same! :+1:

nixpanic:
> > https://commitlint.js.org/#/guides-ci-setup seems like an easy replacement, using a travis job and custom configuration file.
> 
> sure.. thanks for the pointer.. let me go through the same! +1

I am testing this as well, adding it to the containerized-test so contributors can check it locally. Not sure how things will work out though (or what format the configuration is).

You can already remove the Semantic Pull Request bot from the repository, it erroneously gives errors.

humblec:
@nixpanic @Madhu-1 `semantic bot` is removed, so this PR is good to go.

nixpanic:
It seems the Semantic Pull Request bot/app is still there? PR #1058 was created after your comment, and still the status gets reported?

humblec:
> It seems the Semantic Pull Request bot/app is still there? PR #1058 was created after your comment, and still the status gets reported?

Its surely gone for now :). commitlint was removed mistakenly instead of semanticpull. Now back to the state we want @nixpanic

Issue #998, opened 2020-04-29
make util.clusterInfo usable outside of the util package


nixpanic:
Functions like `getClusterInformation()` should probably return a `util.ClusterInfo` struct, instead of the single members.

This requires the current `util.clusterInfo` from to be renamed to `util.ClusterInfo` so that `internal/cephfs/volumeoptions.go` can use it. Also the copies in `e2e/utils.go` can then be removed.

Madhu-1:
@Yuggupta27 let me know if you take it up.

nixpanic:
> /assign @Yuggupta27

This does not work until we have a bot that handles these requests. After @Yuggupta27 responds in this issue, we can manually assign him.

Yuggupta27:
Thanks! I'm on it👍

PR #1056, opened 2020-05-13
util: make util.ClusterInfo usable outside util package

Yuggupta27:
functions like getClusterInfo() returns ClusterInfo
struct instead of a set of strings.

Fixes: #998 


internal/cephfs/volumeoptions.go
```diff
@@ -110,30 +111,35 @@ func extractMounter(dest *string, options map[string]string) error {
...
 }
 
-func getClusterInformation(options map[string]string) (string, string, string, error) {
+func getClusterInformation(options map[string]string) (util.ClusterInfo, error) {
+	var clusterData util.ClusterInfo
 	clusterID, ok := options["clusterID"]
 	if !ok {
 		err := fmt.Errorf("clusterID must be set")
-		return "", "", "", err
+		return util.ClusterInfo{}, err
```
nixpanic, line 119: it is probably better to return `nil` in the error cases

internal/cephfs/volumeoptions.go
```diff
@@ -110,30 +111,35 @@ func extractMounter(dest *string, options map[string]string) error {
...
 
 	subvolumeGroup, err := util.CephFSSubvolumeGroup(csiConfigFile, clusterID)
 	if err != nil {
 		err = errors.Wrapf(err, "failed to fetch subvolumegroup using clusterID (%s)", clusterID)
-		return "", "", "", err
+		return util.ClusterInfo{}, err
 	}
-
-	return clusterID, monitors, subvolumeGroup, err
+	clusterData = util.ClusterInfo{
```
nixpanic, line 137: You are creating a new `util.ClusterInfo` here, so it is not needed to define it as a `var` in the beginning of the function (use `clusterData := util.ClusterInfo{` ,with the `:=` here).

internal/cephfs/volumeoptions.go
```diff
@@ -147,7 +153,11 @@ func newVolumeOptions(ctx context.Context, requestName string, req *csi.CreateVo
 
 	volOptions := req.GetParameters()
 
-	opts.ClusterID, opts.Monitors, opts.SubvolumeGroup, err = getClusterInformation(volOptions)
+	clusterData, err := getClusterInformation(volOptions)
+	opts.ClusterID = clusterData.ClusterID
```
nixpanic, line 157: check `err` before using `clusterData` (it should panic/segfault when `clusterData` is `nil`)

internal/cephfs/volumeoptions.go
```diff
@@ -373,7 +383,11 @@ func newVolumeOptionsFromStaticVolume(volID string, options map[string]string) (
 	// store NOT of static boolean
 	opts.ProvisionVolume = !staticVol
 
-	opts.ClusterID, opts.Monitors, opts.SubvolumeGroup, err = getClusterInformation(options)
+	clusterData, err := getClusterInformation(options)
```
nixpanic, line 386: check `err` before using `clusterData`

internal/util/csiconfig.go
```diff
@@ -58,19 +58,19 @@ type clusterInfo struct {
...
-	var config []clusterInfo
+func readClusterInfo(pathToConfig, clusterID string) (ClusterInfo, error) {
+	var config []ClusterInfo
 
 	// #nosec
 	content, err := ioutil.ReadFile(pathToConfig)
 	if err != nil {
 		err = fmt.Errorf("error fetching configuration for cluster ID (%s). (%s)", clusterID, err)
-		return clusterInfo{}, err
+		return ClusterInfo{}, err
```
nixpanic, line 68: I prefer `nil` returns when an error happens. It was an empty structure before, but that is a little more sensitive to problems as it can be used without detecting an error.

Yuggupta27: Thanks, I'll make the required changes.

Madhu-1:
need to remove https://github.com/ceph/ceph-csi/blob/master/e2e/utils.go#L369 and make use of exported clusterinfo in e2e as well

Yuggupta27:
Okay, will do that.

internal/cephfs/volumeoptions.go
```diff
@@ -373,10 +385,18 @@ func newVolumeOptionsFromStaticVolume(volID string, options map[string]string) (
...
+
 	if err != nil {
 		return nil, nil, err
 	}
+	if clusterData == nil {
```
nixpanic, line 393: I think it is safe to assume that `clusterData` is not `nil` if there was no error. No need to have this if-statement for a 2nd error check.

Yuggupta27: Okay.

nixpanic:
CI reports the following error:
```
e2e/utils.go:15: File is not `goimports`-ed with -local github.com/ceph/ceph-csi (goimports)
```
This means the imports are not sorted correctly.

You can test this locally by running `make containerized-tests`.

e2e/utils.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	// _ "github.com/kubernetes-csi/external-snapshotter/pkg/apis/volumesnapshot/v1alpha1"                             // nolint
 	// _ "github.com/kubernetes-csi/external-snapshotter/pkg/client/clientset/versioned/typed/volumesnapshot/v1alpha1" // nolint
+	"github.com/ceph/ceph-csi/internal/util"
```
nixpanic, line 15: Add an empty line between 'sections' of imports, that should address the CI complaint.

Yuggupta27: > Add an empty line between 'sections' of imports, that should address the CI complaint.

Yes, This was the issue... Thanks!

Issue #979, opened 2020-04-23
makefile target 'go-lint-text' is poorly named


phlogistonjohn:
# Describe the bug #

Running `make containerized-test` is failing for me. While attempting to debug the failure I focused a bit on the name reported by the make failure. 
```
make: *** [Makefile:67: go-lint-text] Error 123
make: *** [Makefile:85: containerized-test] Error 2
```

However, it turns out the failure had nothing to do with changes I made to go code and rather was a failure of shellcheck. Since the script `lint-text.sh` does not check Go language code at all I'd like to suggest renaming the makefile target to something a bit more accurate.

While I'm complaining about the naming-of-things, "lint-text" is not quite accurate either, I'd hesitate to call shell scripts "text".

Suggestions:
"lint-extras" or "lint-non-go"

Or perhaps it would be even clearer if the linting of shell scripts was split from the liniting of yaml and markdown files?

# Steps to reproduce #

Run `make containerized-test CONTAINER_CMD=podman` and it fails every time for me, both on master and my working branch.

nixpanic:
I would prefer to have different targets for each test, `make containerized-test TARGET=...` can then e used to run a single target to make developing/fixing/debugging faster.

PR #1060, opened 2020-05-14
ci: updated shell script to run separate test

Madhu-1:
Earlier we were running all the linter for non-go files in one short, this won't be helpful for the users who want to run particular tests.

Fixes: #979


Madhu-1:
> Typo in the subjects 'indivisual'.
> 
> The commit message is not very clear... Is this not just renaming the `make go-lint-text` target to `make lint-extras`?

its doing more than that, I have separated each test into separate commands.

scripts/lint-extras.sh
```diff
@@ -0,0 +1,93 @@
+#! /bin/bash
```
nixpanic, line 1: No space allowed between `#! /executable`

scripts/lint-extras.sh
```diff
@@ -0,0 +1,93 @@
...
+# vim: set ts=4 sw=4 et :
+
```
nixpanic, line 3: Add a description of the script in the header. Also include some example of how the script can be executed.

scripts/lint-extras.sh
```diff
@@ -0,0 +1,93 @@
...
+set -e
+
+# Run checks from root of the repo
+scriptdir="$(dirname "$(realpath "$0")")"
+cd "$scriptdir/.."
+
+# run_check <file_regex> <checker_exe> [optional args to checker...]
+# Pass empty regex when no regex is needed
+function run_check() {
+    regex="$1"
```
nixpanic, line 13: this is a bash script, use `local` for variables used only in the function

scripts/lint-extras.sh
```diff
@@ -0,0 +1,93 @@
...
+    shift
+    exe="$1"
+    shift
+
+    if [ -x "$(command -v "$exe")" ]; then
+        if [ -z "$regex" ]; then
```
nixpanic, line 19: use `${regex}` notations instead `$regex`, everywhere

scripts/lint-extras.sh
```diff
@@ -0,0 +1,93 @@
...
+            "$exe" "$@"
+        else
+            find . -path ./vendor -prune -o -regextype egrep -iregex "$regex" -print0 |
```
nixpanic, line 22: maybe use `git ls-files --exclude='vendor/*' "${regex}"` instead of `find`

Madhu-1: this is not working, i will leave it as it is for now as this is copy-pasted from old script file

PR #1063, opened 2020-05-14
util: use go-ceph instead of "rados" command for GetPoolID()

nixpanic:
## Describe what this PR does #

Remove an other command execution and replace it by using go-ceph.

GetPoolID() did not return ErrPoolNotFound in case the pool could not be
found. This has been addressed as well, so that looking for an existing
pool behaves the same for checking by Name or ID.

Also remove unused context.Context parameter from GetPoolID().

### DIFF
--- a/internal/util/cephcmds.go
+++ b/internal/util/cephcmds.go
@@ -89,7 +89,7 @@ func getPools(ctx context.Context, monitors string, cr *Credentials) ([]cephStor
 
 // GetPoolID fetches the ID of the pool that matches the passed in poolName
 // parameter
-func GetPoolID(ctx context.Context, monitors string, cr *Credentials, poolName string) (int64, error) {
+func GetPoolID(monitors string, cr *Credentials, poolName string) (int64, error) {
 	conn, err := connPool.Get(monitors, cr.ID, cr.KeyFile)
 	if err != nil {
 		return 0, err
@@ -127,14 +127,14 @@ func GetPoolName(ctx context.Context, monitors string, cr *Credentials, poolID i
 // the passed in pools
 // TODO this should take in a list and return a map[string(poolname)]int64(poolID)
 func GetPoolIDs(ctx context.Context, monitors, journalPool, imagePool string, cr *Credentials) (int64, int64, error) {
-	journalPoolID, err := GetPoolID(ctx, monitors, cr, journalPool)
+	journalPoolID, err := GetPoolID(monitors, cr, journalPool)
 	if err != nil {
 		return InvalidPoolID, InvalidPoolID, err
 	}
 
 	imagePoolID := journalPoolID
 	if imagePool != journalPool {
-		imagePoolID, err = GetPoolID(ctx, monitors, cr, imagePool)
+		imagePoolID, err = GetPoolID(monitors, cr, imagePool)
 		if err != nil {
 			return InvalidPoolID, InvalidPoolID, err
 		}
--- a/internal/util/cephcmds.go
+++ b/internal/util/cephcmds.go
@@ -106,21 +106,20 @@ func GetPoolID(monitors string, cr *Credentials, poolName string) (int64, error)
 	return id, nil
 }
 
-// GetPoolName lists all pools in a ceph cluster, and matches the pool whose pool ID is equal to
-// the requested poolID parameter
+// GetPoolName fetches the pool whose pool ID is equal to the requested poolID
+// parameter
 func GetPoolName(ctx context.Context, monitors string, cr *Credentials, poolID int64) (string, error) {
-	pools, err := getPools(ctx, monitors, cr)
+	conn, err := connPool.Get(monitors, cr.ID, cr.KeyFile)
 	if err != nil {
 		return "", err
 	}
+	defer connPool.Put(conn)
 
-	for _, p := range pools {
-		if poolID == p.Number {
-			return p.Name, nil
-		}
+	name, err := conn.GetPoolByID(poolID)
+	if err != nil {
+		return "", ErrPoolNotFound{string(poolID), fmt.Errorf("pool ID (%d) not found in Ceph cluster", poolID)}
 	}
-
-	return "", ErrPoolNotFound{string(poolID), fmt.Errorf("pool ID (%d) not found in Ceph cluster", poolID)}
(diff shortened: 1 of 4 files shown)


internal/util/cephcmds.go
```diff
@@ -86,21 +87,24 @@ func getPools(ctx context.Context, monitors string, cr *Credentials) ([]cephStor
...
 }
 
-// GetPoolID searches a list of pools in a cluster and returns the ID of the pool that matches
-// the passed in poolName parameter
-func GetPoolID(ctx context.Context, monitors string, cr *Credentials, poolName string) (int64, error) {
-	pools, err := getPools(ctx, monitors, cr)
+// GetPoolID fetches the ID of the pool that matches the passed in poolName
+// parameter
+func GetPoolID(monitors string, cr *Credentials, poolName string) (int64, error) {
+	conn, err := connPool.Get(poolName, monitors, cr.KeyFile)
```
Madhu-1, line 93: seems like arguments are not passed in the proper order can you please double-check? https://github.com/ceph/ceph-csi/blob/master/internal/util/conn_pool.go#L123

nixpanic: right, good catch! the pool is not needed for `Get()`, but the userId.

phlogistonjohn:
Apologies in advance but I'm going to hijack this PR for a moment to try and discuss some go-ceph integration "meta issues".

I was recently working on making the "journal" package go-ceph based and hit some issues that sent me off on a refactoring quest. Compared to these changes I see that I was probably trying to do too much at once but it gets me thinking...
When do you feel one should be taking the approach of using `connPool.Get` directly versus using an approach like in rbd_util.go with `*util.ClusterConnection` and `rbdVolume.Connect`? Or to put it another way: how do *you* decide what approach to take? I figure right now this is not written down yet and mostly lives in your head. :-)

There's also the issue that go-ceph calls aren't trivally cancel-able the way an external process is. That's why context is no longer needed, but it gives me pause and I wonder if we shouldn't at least keep the context and add a helper call that monitors the call and at least logs if it takes longer than a certain deadline (provided by the context).
I had advocated a similar logging approach before a couple of times, but mainly to @ShyamsundarR  in person. :-) What I'm not clear on is at what layer the logging would occur.

nixpanic:
> I was recently working on making the "journal" package go-ceph based and hit some issues that sent me off on a refactoring quest. Compared to these changes I see that I was probably trying to do too much at once but it gets me thinking...
> When do you feel one should be taking the approach of using `connPool.Get` directly versus using an approach like in rbd_util.go with `*util.ClusterConnection` and `rbdVolume.Connect`? Or to put it another way: how do _you_ decide what approach to take? I figure right now this is not written down yet and mostly lives in your head. :-)

My (sortof) goal is to have a rbdVolume (or probably RbdVolume) API that is pretty much self contained. This can then easily be used by the nodeserver and controllerserver, without caring about the implantation details. It is a long term thing, and the steps towards it are pretty small.

In the end, the structure of a large shared `rbd` package would change, and smaller packages with a clear API should be easier to understand and maintain.

> There's also the issue that go-ceph calls aren't trivally cancel-able the way an external process is. That's why context is no longer needed (after your changes), but it gives me pause and I wonder if we shouldn't at least keep the context and add a helper call that monitors the call and at least logs if it takes longer than a certain deadline (provided by the context).

Currently the `Context` is only used for logging. Where functions do not do any logging at all, or anymore, I am dropping the unused arguments.

When there is a use for passing `Context` and have it abort, we can re-add the arguments to whatever function needs it.

> I had advocated a similar logging approach before a couple of times, but mainly to @ShyamsundarR in person. :-) What I'm not clear on is at what layer the logging would occur.

That is something that I have been thinking a little about too. Maybe by passing the context to `connPool.Get()` so that the connection can be marked faulty and destroyed. But that is only an idea, and no prototyping exists for this yet.

phlogistonjohn:
Thanks! I've thought about this some more, looking at the changes you're making here and those I have been attempting for the omap stuff, and I'm thinking that a new, simple interface may be useful here. The interface's job is to intermediate the job of fetching a connection for functions that make use of go-ceph. Pseudo-go proposal:

```go

  
// General interface:
// (Based on your earlier comment this could also take a ctx argument)

type ConnectionMaker interface {
    Get() (*rados.Conn, err)
    Put(*rados.Conn)
}


// Helper function to partially apply monitors + credentials:
// For call sites that currently look like util.Foo(ctx, monitors, cr, bar, baz)
// (this could also be a public type, but I'd argue a function is shorter/clearer)

func FromConnPool(monitors string, cr *util.Credentials) ConnectionMaker {
    // ...
}

// Functions that "do go-ceph stuff" become:
  
func GetPoolName(cm ConnectionMaker, poolID int64) (string, error) {
    conn, err := cm.Get()
    if err != nil {
        return "", err
    }
    defer cm.Put(conn)
    name, err := conn.GetPoolByID(poolID)
    if err != nil {
        return "", ErrPoolNotFound{string(poolID), fmt.Errorf("pool ID (%d) not found in Ceph cluster", poolID)}
    }
    return name, nil
}

// Now code that needs to establish a totally new connection and has no other helper types
// from a "parent" object becomes:

poolName, err := GetPoolName(FromConnPool(monitors, cr), myPoolID)

// Something like util.ClusterConnection then grows new Get/Put methods, to meet the interface:

func (cc *ClusterConnection) Get() (*rados.Conn, err) {
    return cc.conn
}

func (cc *ClusterConnection) Put(c *rados.Conn) {}
    
    
// Then other code which has already established a connection would call GetPoolName like so:

poolName, err := GetPoolName(conn, myPoolID)

```

I did something similar in Heketi with database connections so that you can make use call db related functions by a db connection from a top level function OR just creating a "connection" from an already created db transaction. It's worked pretty well at avoiding having to create many various functions that basically do the same thing IMO.

Thoughts?

nixpanic:
> I did something similar in Heketi with database connections so that you can make use of db related functions with a full blown db connection from a top level function OR just creating a "connection" from an already created db transaction in a nested situation. It's worked pretty well at avoiding having to create many various functions that basically do the same thing IMO.
> 
> Thoughts?

I do not have any objections to this approach. But I do wonder how much the benefit will be...

PR #1066, opened 2020-05-14
journal: split journal types creating a new Connection type

phlogistonjohn:

Before, the one CSIJournal type was handling both configuration and
providing methods to make changes to the journal. This created the
temptation to modify the state of the global configuration object to
enact changes through the method calls.

This change creates a new type `journal.Connection` that takes the
monitors and credentials to create a short(er)-lived object to actually
read and make changes on the journal. This also avoids mixing the
arguments needed to connect to the cluster with the arguments needed
for the various journal read & update calls.

Lastly, rename CSIJournal to Config to better reflect what it actually does now.

This is, hopefully, the last pure refactoring work done before adding go-ceph support to the journal package.

Are there concerns around backward compatibility? no


internal/journal/voljournal.go
```diff
@@ -204,6 +204,25 @@ type ImageData struct {
...
+// that can make changes to the journal(s).
+type Connection struct {
+	config *Config
+	// connection metadata
+	monitors string
+	cr       *util.Credentials
+}
+
+// Connect establishes a new connection to a ceph cluster for journal metadata.
+func (cj *Config) Connect(monitors string, cr *util.Credentials) (*Connection, error) {
```
Madhu-1, line 217: do we need to return an error here? it always returns nil

phlogistonjohn: It's future proofing as I have upcoming code changes that will be able to return a real error. If I don't add error-handling to all the call sites now I end up doing some of the work twice.

PR #1078, opened 2020-05-18
rbd: unmap rbd image if the mounting fails

Madhu-1:
There is a bug in the current code where the device path is always empty and the rbd image unmap never happens if node plugin fails to mount the rbd image to the staging path.

This is a fix to unmap the rbd image if some issue occurs after the rbd image is mapped.


Madhu-1:
> The change looks good, but please add an e2e test for this.
> 
> Thinking about the test-case, I can come up with these issues:
> 
> * forcefully corrupt the image to prevent mounting
> * temporary remove/rename the `/sbin/mount.xfs` utility
> * pass mount options that will cause mount to fail

@nixpanic added an E2E for invalid mount options

humblec:
@Madhu-1 I thought this we fixed some time back :(.

Madhu-1:
> @Madhu-1 I thought this we fixed some time back :(.

This regression introduced in https://github.com/ceph/ceph-csi/pull/842

e2e/rbd.go
```diff
@@ -568,6 +568,40 @@ var _ = Describe("RBD", func() {
...
+					Fail(err.Error())
+				}
+				pvc.Namespace = f.UniqueName
+
+				app, err := loadApp(appPath)
+				if err != nil {
+					Fail(err.Error())
+				}
+				app.Namespace = f.UniqueName
+				err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
```
humblec, line 587: Its a naming convention, but `createPVCAndValidatePV` looks better.

e2e/rbd.go
```diff
@@ -568,6 +568,40 @@ var _ = Describe("RBD", func() {
...
+				if err != nil {
+					Fail(err.Error())
+				}
+				// create an app and wait for 2 min for it to go to running state
+				err = createApp(f.ClientSet, app, 2)
```
humblec, line 592: just wondering how we arrive on this timeout of `2` mins and is this enough? or will it cause the e2e goes wrong if the creation takes more time ?

e2e/rbd.go
```diff
@@ -568,6 +568,40 @@ var _ = Describe("RBD", func() {
...
+				if err == nil {
+					Fail("application should not go to running state")
```
humblec, line 594: May be error can be `application should not go to running state due to invalid mount options` or similar message will be more descriptive.

e2e/rbd.go
Madhu-1: never seen an issue with this number, waiting for more time increases the time taken to execute E2E. if we face any issue this can be changed later also

e2e/rbd.go
Madhu-1: the function already exists. I don't want to change it in this PR

Madhu-1:
@humblec changes done PTAL

Madhu-1:
@humblec This is not merged yet

PR #1081, opened 2020-05-18
ci: retry if no machine is immediately available

Yuggupta27:
To prevent the failure of a job due to unavailability of 
a machine immediately, a retry mechanism is used.

If unable to reserve a machine, it will retry every
5 mins for 30 times to avoid job failure.

### DIFF
--- a/containerized-tests.groovy
+++ b/containerized-tests.groovy
@@ -13,12 +13,19 @@ node('cico-workspace') {
 	}
 
 	stage('reserve bare-metal machine') {
-		cico = sh(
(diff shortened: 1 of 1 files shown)


Yuggupta27:
> What happens when the number of retries is up, and no machine has been found?

The build will fail if that happens

nixpanic:
> The build will fail if that happens

I guess at the stage step?

Yuggupta27:
It will be unable to reserve machine and will fail at that stage

PR #1093, opened 2020-05-21
deploy: Enable `extraCreateMetadata` for provisioner

humblec:
Pass extra metadata when creating volumes makes it possible to detect the Kubernetes Namespace where the PVC is created. This can then be used for fetching the Vault Token (or other configuration) from the namespace where the PVC lives.

The extra metadata is available in the parameters of the CreateVolume request, for example:
```
GRPC call: /csi.v1.Controller/CreateVolume
GRPC request: {"accessibility_requirements":{"preferred":[{"segments":{"topology.rbd.csi.ceph.com/region":"testregion","topology.rbd.csi.ceph.com/zone":"testzone"}}],"requisite":[{"segments":{"topology.rbd.csi.ceph.com/region":"testregion","topology.rbd.csi.ceph.com/zone":"testzone"}}]},"capacity_range":{"required_bytes":1073741824},"name":"pvc-a18380b9-d9b7-46df-be26-693e0530c715","parameters":{"clusterID":"fa4b2e2b-ceb9-4db2-a96b-162a815a31a9","csi.storage.k8s.io/pv/name":"pvc-a18380b9-d9b7-46df-be26-693e0530c715","csi.storage.k8s.io/pvc/name":"rbd-pvc","csi.storage.k8s.io/pvc/namespace":"rbd-7887","encrypted":"true","imageFeatures":"layering","pool":"replicapool"},"secrets":"***stripped***","volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4","mount_flags":["discard"]}},"access_mode":{"mode":1}}]}
```

See-also: #1305

### DIFF
--- a/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
+++ b/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
@@ -50,6 +50,7 @@ spec:
             - "--leader-election=true"
             - "--retry-interval-start=500ms"
             - "--feature-gates=Topology=false"
+            - "--extra-create-metadata=true"
           env:
             - name: ADDRESS
               value: unix:///csi/csi-provisioner.sock
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -260,6 +260,15 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("create a PVC and validate owner", func() {
+				err := validateImageOwner(pvcPath, f)
+				if err != nil {
+					e2elog.Failf("failed to validate owner of pvc with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0)
+			})
+
 			By("create a PVC and bind it to an app", func() {
 				err := validatePVCAndAppBinding(pvcPath, appPath, f)
 				if err != nil {
@@ -1109,8 +1118,15 @@ var _ = Describe("RBD", func() {
 
 				updateConfigMap("e2e-ns")
 
+				err := validateImageOwner(pvcPath, f)
+				if err != nil {
+					e2elog.Failf("failed to validate owner of pvc with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0)
+
 				// Create a PVC and bind it to an app within the namesapce
-				err := validatePVCAndAppBinding(pvcPath, appPath, f)
+				err = validatePVCAndAppBinding(pvcPath, appPath, f)
 				if err != nil {
 					e2elog.Failf("failed to validate pvc and application binding with error %v", err)
 				}
--- a/e2e/rbd_helper.go
+++ b/e2e/rbd_helper.go
@@ -190,6 +190,50 @@ func getImageMeta(rbdImageSpec, metaKey string, f *framework.Framework) (string,
 	return strings.TrimSpace(stdOut), nil
 }
 
+// validateImageOwner checks the "csi.volume.owner" key on the image journal
+// and verifies that the owner is set to the namespace where the PVC is
+// created.
+func validateImageOwner(pvcPath string, f *framework.Framework) error {
+	const ownerKey = "csi.volume.owner"
+
+	pvc, err := loadPVC(pvcPath)
+	if err != nil {
+		return err
+	}
+	pvc.Namespace = f.UniqueName
+	pvc.Name = f.UniqueName
+	err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+	if err != nil {
+		return err
+	}
+
+	imageData, err := getImageInfoFromPVC(pvc.Namespace, pvc.Name, f)
+	if err != nil {
+		return err
+	}
+
+	stdOut, stdErr, err := execCommandInToolBoxPod(f,
+		fmt.Sprintf("rados %s getomapval csi.volume.%s %s", rbdOptions(defaultRBDPool), imageData.imageID, ownerKey), rookNamespace)
+	if err != nil {
+		return err
+	}
+	if stdErr != "" {
+		return fmt.Errorf("failed to getomapval %v", stdErr)
+	}
+
+	if radosNamespace != "" {
+		e2elog.Logf("found image journal %s in pool %s namespace %s", "csi.volume."+imageData.imageID, defaultRBDPool, radosNamespace)
+	} else {
+		e2elog.Logf("found image journal %s in pool %s", "csi.volume."+imageData.imageID, defaultRBDPool)
+	}
+
+	if !strings.Contains(stdOut, pvc.Namespace) {
+		return fmt.Errorf("%q does not contain %q: %s", ownerKey, pvc.Namespace, stdOut)
+	}
+
+	return deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+}
+
 func validateEncryptedPVCAndAppBinding(pvcPath, appPath, kms string, f *framework.Framework) error {
 	pvc, app, err := createPVCAndAppBinding(pvcPath, appPath, f, deployTimeout)
 	if err != nil {
--- a/internal/cephfs/fsjournal.go
+++ b/internal/cephfs/fsjournal.go
@@ -232,7 +232,7 @@ func reserveVol(ctx context.Context, volOptions *volumeOptions, secret map[strin
 	imageUUID, vid.FsSubvolName, err = j.ReserveName(
 		ctx, volOptions.MetadataPool, util.InvalidPoolID,
 		volOptions.MetadataPool, util.InvalidPoolID, volOptions.RequestName,
-		volOptions.NamePrefix, "", "", volOptions.ReservedID)
+		volOptions.NamePrefix, "", "", volOptions.ReservedID, "")
 	if err != nil {
 		return nil, err
 	}
@@ -269,7 +269,7 @@ func reserveSnap(ctx context.Context, volOptions *volumeOptions, parentSubVolNam
 	imageUUID, vid.FsSnapshotName, err = j.ReserveName(
 		ctx, volOptions.MetadataPool, util.InvalidPoolID,
 		volOptions.MetadataPool, util.InvalidPoolID, snap.RequestName,
-		snap.NamePrefix, parentSubVolName, "", snap.ReservedID)
+		snap.NamePrefix, parentSubVolName, "", snap.ReservedID, "")
 	if err != nil {
 		return nil, err
 	}
--- a/internal/journal/voljournal.go
+++ b/internal/journal/voljournal.go
@@ -148,6 +148,9 @@ type Config struct {
 	// encryptKMS in which encryption passphrase was saved, default is no encryption
 	encryptKMSKey string
 
+	// ownerKey is used to identify the owner of the volume, can be used with some KMS configurations
+	ownerKey string
+
 	// commonPrefix is the prefix common to all omap keys for this Config
 	commonPrefix string
 }
@@ -165,6 +168,7 @@ func NewCSIVolumeJournal(suffix string) *Config {
 		namespace:               "",
 		csiImageIDKey:           "csi.imageid",
 		encryptKMSKey:           "csi.volume.encryptKMS",
+		ownerKey:                "csi.volume.owner",
 		commonPrefix:            "csi.",
 	}
 }
@@ -182,6 +186,7 @@ func NewCSISnapshotJournal(suffix string) *Config {
 		namespace:               "",
 		csiImageIDKey:           "csi.imageid",
 		encryptKMSKey:           "csi.volume.encryptKMS",
+		ownerKey:                "csi.volume.owner",
 		commonPrefix:            "csi.",
 	}
 }
@@ -494,6 +499,7 @@ Input arguments:
 	- kmsConf: Name of the key management service used to encrypt the image (optional)
         - volUUID: UUID need to be reserved instead of auto-generating one (this is
           useful for mirroring and metro-DR)
+	- owner: the owner of the volume (optional)
 
 Return values:
 	- string: Contains the UUID that was reserved for the passed in reqName
@@ -503,7 +509,7 @@ Return values:
 func (conn *Connection) ReserveName(ctx context.Context,
 	journalPool string, journalPoolID int64,
 	imagePool string, imagePoolID int64,
-	reqName, namePrefix, parentName, kmsConf, volUUID string) (string, string, error) {
+	reqName, namePrefix, parentName, kmsConf, volUUID, owner string) (string, string, error) {
 	// TODO: Take in-arg as ImageAttributes?
 	var (
 		snapSource bool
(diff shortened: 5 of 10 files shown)


deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
```diff
@@ -41,6 +41,7 @@ spec:
             - "--leader-election-type=leases"
             - "--retry-interval-start=500ms"
             - "--feature-gates=Topology=true"
+            - "--extra-create-metadata=true
```
nixpanic, line 44: Missing closing `"`.

nixpanic (dismissed):
Please correct the subject of the commit message, no trailing `:`

deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
humblec: added.

Madhu-1 (approved):

humblec:
Adding `DNM` , just to revisit the doc updates if required any.

humblec:
> Adding `DNM` , just to revisit the doc updates if required any.

Looks like we can leave without doc update. Removing `DNM` .

humblec:
@nixpanic ptal .

nixpanic (commented):
Does there need to be an e2e tests to verify that the expected metadata is indeed available?
It would help to link to the documentation of the option in the commit message.

Madhu-1 (dismissed):
Holding this one as we have not decided yet to go with pvc-name+namespace name for DR usecase

humblec:
> Holding this one as we have not decided yet to go with pvc-name+namespace name for DR usecase

@Madhu-1 ideally this is not related to DR use case or naming schema. This is independent to one use case. I dont see a point in having it blocked because of that.

Madhu-1:
> > Holding this one as we have not decided yet to go with pvc-name+namespace name for DR usecase
> 
> @Madhu-1 ideally this is not related to DR use case or naming schema. This is independent to one use case. I dont see a point in having it blocked because of that.

can you please update the same details in the PR description and also provide detail on how this can be used? I don't see a value in just  passing the metadata to CSI driver without any valid use-case

humblec:
> > > Holding this one as we have not decided yet to go with pvc-name+namespace name for DR usecase
> > 
> > 
> > @Madhu-1 ideally this is not related to DR use case or naming schema. This is independent to one use case. I dont see a point in having it blocked because of that.
> 
> can you please update the same details in the PR description and also provide detail on how this can be used? I don't see a value in just passing the metadata to CSI driver without any valid use-case

@Madhu-1 updated the commit message and PR description with some more details.

humblec:
@Madhu-1 @nixpanic ptal :+1:

nixpanic (dismissed):
If this is an enablement as part of the the design discussed in #1305, we should probably wait with merging until the design has been approved and merged.

humblec:
> If this is an enablement as part of the the design discussed in #1305, we should probably wait with merging until the design has been approved and merged.

@nixpanic this is not tied to those features in full context. It could be a useful input for CSI driver to consume as mentioned in the PR description , be it vault enhancement or rbd mirroring or more.. But this is an independent one which dont harm anything and we could consume accordingly. so IMO, we could get this in without any concerns.

nixpanic (changes requested):
> @nixpanic this is not tied to those features in full context. It could be a useful input for CSI driver to consume as mentioned in the PR description , be it vault enhancement or rbd mirroring or more.. But this is an independent one which dont harm anything and we could consume accordingly. so IMO, we could get this in without any concerns.

In that case, this is just a stand-alone feature. Please add something to the documentation so that users (and developers) know it exists. It also needs e2e testing so that we can detect unintentional breakage.

humblec:
> > @nixpanic this is not tied to those features in full context. It could be a useful input for CSI driver to consume as mentioned in the PR description , be it vault enhancement or rbd mirroring or more.. But this is an independent one which dont harm anything and we could consume accordingly. so IMO, we could get this in without any concerns.
> 
> In that case, this is just a stand-alone feature. Please add something to the documentation so that users (and developers) know it exists. It also needs e2e testing so that we can detect unintentional breakage.

@nixpanic I failed to get it , this is not a feature which is user facing! its just that, CSI driver get the additional metadata, thats it and its internal in all the sense. 
I believe the documentation or other user facing things will come in ONLY when CSI driver started to consume it for features like RBD mirroring  or Vault ..etc.. Happy to chat and explain :)

humblec:
@Madhu-1 please take a look at this version. I have updated the PR description with the details.

Madhu-1:
Am not going with the approach mentioned in https://github.com/ceph/ceph-csi/pull/1305 for rbd mirroring.

nixpanic:
> @nixpanic I failed to get it , this is not a feature which is user facing! its just that, CSI driver get the additional metadata, thats it and its internal in all the sense.

Right, but there are no requirements to have this metadata at this time? And there are no consumers within Ceph-CSI that use this metadata. No e2e tests that verify the metadata is passed, is stored or retrieved anywhere. This currently looks like bloat that is not necessary.

humblec:
> > @nixpanic I failed to get it , this is not a feature which is user facing! its just that, CSI driver get the additional metadata, thats it and its internal in all the sense.
> 
> Right, but there are no requirements to have this metadata at this time? And there are no consumers within Ceph-CSI that use this metadata.
@nixpanic it is true that, there is NO fixed or confirmed consumer at this time from our CSI driver. I agree to that.!

> No e2e tests that verify the metadata is passed, is stored or retrieved anywhere. This currently looks like bloat that is not necessary.

The E2E tests are part of the sidecar ( here external-provisioner) and its well qualified against the CSI driver. 
Unfortunately I dont understand `stored` and `retrieved` -> this is on the consuming part. Again, it depends on the feature it consume.
 
Regardless, one of the thought I have here is `recording` namespace of a PVC  in volumeContext of a PV for better filtering of the request in later cycle.

Madhu-1:
> > > @nixpanic I failed to get it , this is not a feature which is user facing! its just that, CSI driver get the additional metadata, thats it and its internal in all the sense.
> > 
> > 
> > Right, but there are no requirements to have this metadata at this time? And there are no consumers within Ceph-CSI that use this metadata.
> 
> @nixpanic it is true that, there is NO fixed or confirmed consumer at this time from our CSI driver. I agree to that.!
> 
> > No e2e tests that verify the metadata is passed, is stored or retrieved anywhere. This currently looks like bloat that is not necessary.
> 
> The E2E tests are part of the sidecar ( here external-provisioner) and its well qualified against the CSI driver.
> Unfortunately I dont understand `stored` and `retrieved` -> this is on the consuming part. Again, it depends on the feature it consume.
> 
> Regardless, one of the thought I have here is `recording` namespace of a PVC in volumeContext of a PV for better filtering of the request in later cycle.

How feasible it is to record the namespace in PV? PV is not bound to any namespace or the PVC. a user can statically bind the PV to any pvc on any namespace

humblec:
>How feasible it is to record the namespace in PV? PV is not bound to any namespace or the PVC. a user can statically bind the PV to any pvc on any namespace

I added some more data to https://github.com/ceph/ceph-csi/pull/1093/#issuecomment-715206813 in between. Any way to the above,  PV is a global object , but when its bound it always have a 1:1 mapping to a PVC, and PVC is namespace scoped.  Does that answer ?

Madhu-1:
> > How feasible it is to record the namespace in PV? PV is not bound to any namespace or the PVC. a user can statically bind the PV to any pvc on any namespace
> 
> I added some more data to #1093 in between. Any way to the above, PV is a global object , but when its bound it always have a 1:1 mapping to a PVC, and PVC is namespace scoped. Does that answer ?

what if the user deletes the PVC and retain the PV and bind it to another PVC in a different namespace?

humblec:
> > > How feasible it is to record the namespace in PV? PV is not bound to any namespace or the PVC. a user can statically bind the PV to any pvc on any namespace
> > 
> > 
> > I added some more data to #1093 in between. Any way to the above, PV is a global object , but when its bound it always have a 1:1 mapping to a PVC, and PVC is namespace scoped. Does that answer ?
> 
> what if the user deletes the PVC and retain the PV and bind it to another PVC in a different namespace?

Unfortunately I dont understand this :(.. In this case, namespace is different and NOT what we have recorded, Isnt it ?

Madhu-1:
> > > > How feasible it is to record the namespace in PV? PV is not bound to any namespace or the PVC. a user can statically bind the PV to any pvc on any namespace
> > > 
> > > 
> > > I added some more data to #1093 in between. Any way to the above, PV is a global object , but when its bound it always have a 1:1 mapping to a PVC, and PVC is namespace scoped. Does that answer ?
> > 
> > 
> > what if the user deletes the PVC and retain the PV and bind it to another PVC in a different namespace?
> 
> Unfortunately I dont understand this :(.. In this case, namespace is different and NOT what we have recorded, Isnt it ?

Yes.

humblec:
> > > > > How feasible it is to record the namespace in PV? PV is not bound to any namespace or the PVC. a user can statically bind the PV to any pvc on any namespace
> > > > 
> > > > 
> > > > I added some more data to #1093 in between. Any way to the above, PV is a global object , but when its bound it always have a 1:1 mapping to a PVC, and PVC is namespace scoped. Does that answer ?
> > > 
> > > 
> > > what if the user deletes the PVC and retain the PV and bind it to another PVC in a different namespace?
> > 
> > 
> > Unfortunately I dont understand this :(.. In this case, namespace is different and NOT what we have recorded, Isnt it ?
> 
> Yes.

So, its a static PV workflow and not part of  dynamic provisioning workflow. Also, this gives a `validation checkpoint` if needed later while we are operating on such a conflicting PV and if at all the requested operation depend on the `namespace` specific decisions. If its not specific or not depend on the namespace specific operation, we are free to discard it. !

humblec:
@nixpanic @Madhu-1 are we converging to the point that, we dont have to consider this unless there is a consumer for this ? please confirm

nixpanic:
> @nixpanic @Madhu-1 are we converging to the point that, we dont have to consider this unless there is a consumer for this ? please confirm

Yes, I would say so. We should not add features that have no clear plan yet of how they benefit the project. Once there is a requirement and way to verify the functioning of this feature, it can be added of course.

nixpanic:
Rebased on top of the master branch and added a 1st part of a consumer. The RBD image now has an `Owner` that is set by the extra `csi.storage.k8s.io/pvc/namespace` metadata.

The [project to add support for a KMS with Vault Tokens](https://github.com/ceph/ceph-csi/projects/8) has more information on for this requirement.

nixpanic (dismissed):

nixpanic (approved):

nixpanic (dismissed):

nixpanic:
**ci/centos/mini-e2e-helm/k8s-1.18** and **ci/centos/mini-e2e-helm/k8s-1.19** failed. It seems the Helm chart was not updated with the additional parameter for the sidecar :man_facepalming:

nixpanic (approved):
@Madhu-1 this is ready for review. In case the `rbdVolume.Owner` needs to be stored as RBD Image metadata (to be confirmed at a later time in #1743), I'll do that in a follow-up PR.

e2e/rbd_helper.go
```diff
@@ -190,6 +190,46 @@ func getImageMeta(rbdImageSpec, metaKey string, f *framework.Framework) (string,
...
+	}
+
+	if radosNamespace != "" {
+		e2elog.Logf("found image journal %s in pool %s namespace %s", "csi.volume."+imageData.imageID, defaultRBDPool, radosNamespace)
+	} else {
+		e2elog.Logf("found image journal %s in pool %s", "csi.volume."+imageData.imageID, defaultRBDPool)
+	}
+
+	ownerKey := "csi.volume.owner"
+	if !strings.Contains(stdOut, ownerKey) {
```
Madhu-1, line 226: can we validate the owner's name? instead of checking the key, best would be to validate the value we store

internal/rbd/rbd_util.go
```diff
@@ -819,6 +822,15 @@ func genVolFromVolumeOptions(ctx context.Context, volOptions, credentials map[st
 		rbdVol.Mounter = rbdDefaultMounter
 	}
 
+	// if the KMS is of type VaultToken, additional metadata is needed
+	// depending on the tenant, the KMS can be configured with other
+	// options
+	// FIXME: this works only on Kubernetes, how do other CO supply metadata?
+	rbdVol.Owner, ok = volOptions["csi.storage.k8s.io/pvc/namespace"]
```
Madhu-1, line 829: Is this needed only when encryption is enabled? if this is not passed we are going to get Debug logs for all volume create requests.

e2e/rbd.go
```diff
@@ -259,6 +259,15 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("create a PVC and validate owner", func() {
+				err := validateImageOwner(pvcPath, f)
```
Madhu-1, line 263: same tests need to be added for the rados namespace test also

Madhu-1 (commented):
small nit, changes looks good

internal/rbd/rbd_util.go
nixpanic: It is not required (at the moment) for non-encrypted volumes. The option to pass extra metadata is passed with all deployments, as we can not  know in advance if encryption will be used, or not. So the owner will be available everywhere.

In production environments, debug logging should not be enabled. I'd expect only errors (and higher) to be logged in production environments by default.

e2e/rbd_helper.go
nixpanic: Sure, will adapt it.

e2e/rbd.go
nixpanic: ok, will do!

nixpanic (approved):

nixpanic (approved):

nixpanic (approved):

internal/cephfs/fsjournal.go
```diff
@@ -232,7 +232,7 @@ func reserveVol(ctx context.Context, volOptions *volumeOptions, secret map[strin
 	imageUUID, vid.FsSubvolName, err = j.ReserveName(
 		ctx, volOptions.MetadataPool, util.InvalidPoolID,
 		volOptions.MetadataPool, util.InvalidPoolID, volOptions.RequestName,
-		volOptions.NamePrefix, "", "", volOptions.ReservedID)
+		volOptions.NamePrefix, "", "", volOptions.ReservedID, "")
```
Madhu-1, line 235: the flag `extra-create-metadata` is set for both cephfs and rbd deployment above but we are not setting the owner for cephfs is there any reason for that one?

internal/journal/voljournal.go
```diff
@@ -182,6 +186,7 @@ func NewCSISnapshotJournal(suffix string) *Config {
 		namespace:               "",
 		csiImageIDKey:           "csi.imageid",
 		encryptKMSKey:           "csi.volume.encryptKMS",
+		ownerKey:                "csi.volume.owner",
```
Madhu-1, line 189: is this needed for a snapshot also?

internal/rbd/rbd_util.go
Madhu-1: in storageclass "encrypted" will be set to true right based on that we can do this operation?

internal/cephfs/fsjournal.go
nixpanic: It has been added to have the options in sync as much as possible. CephFS does not set the KMS-ID either yet, at that point the Owner can be added too.

internal/journal/voljournal.go
nixpanic: I'd say it is needed where the `csi.volume.encryptKMS` is used, otherwise the options to would not be complete.

internal/rbd/rbd_util.go
nixpanic: We could, but in the end it is just an other attribute of the PVC and RBD-image. It does not hurt to have it (and makes testing of the rados key/value simpler).

Madhu-1 (approved):

PR #1083, opened 2020-05-19
ci: [WIP] run e2e tests on multi-node k8s cluster

Yuggupta27:
Add e2e testing in a multi-node 
k8s cluster for Centos CI. 

Signed-off-by: Yug Gupta <ygupta@redhat.com>

PR #1099, opened 2020-05-25
build: verify that all dynamically linked libraries are available

nixpanic:

When building with go-ceph, there are several dynamically linked
libraries used directly (libcephfs, librados and librbd), and many more
indirectly.

By adding an additional RUN statement to check if all libraries are
available in the final image, problems related to missing libraries
should be caught before publishing/consuming the image.


nixpanic:
Adding an `ldd` check so that dynamically linked libraries get verified a little. Compiling a static binary with go-ceph is not realistic in the foreseeable future.

@phlogistonjohn @obnoxxx @agarwal-mudit @ansiwen

nixpanic:
/test ci/centos/e2e

Testing the new e2e CI job from #1083

nixpanic:
/retest ci/centos/containerized-tests

Retesting for #1100 (manually changed the job in Jenkins)

nixpanic:
/test ci/centos/containerized-tests

It seems `/retest` did not get picked up?

nixpanic:
it is

/retest all

or

/retest ci/centos/containerized-tests
(for a specific job)

deploy/cephcsi/image/Dockerfile
```diff
@@ -48,4 +48,7 @@ LABEL maintainers="Ceph-CSI Authors" \
 
 COPY --from=builder ${SRC_DIR}/_output/cephcsi /usr/local/bin/cephcsi
 
+# verify that all dynamically linked libraries are available
+RUN [ $(ldd /usr/local/bin/cephcsi | grep -c '=> not found') = '0' ]
```
Madhu-1, line 52: @nixpanic do we need this check as we are building the go binary with the same base image. won't the go build will fail if the libraries are not available?

nixpanic: The same base image, but we do install additional tools/libraries in the builder container. It is unlikely that this check will fail, but in case it does, we need to get informed.

PR #1100, opened 2020-05-25
ci: use parameters set by github-pull-request-builder

nixpanic:

The parameters are set in the environment. It seems the best way to
check if the parameters are available, is with `params.<arg> == null`.

See-also: https://www.jenkins.io/doc/book/pipeline/getting-started/#global-variable-reference#

And for the `map.key == null` at http://groovy-lang.org/syntax.html#_maps

### DIFF
--- a/containerized-tests.groovy
+++ b/containerized-tests.groovy
@@ -30,7 +30,7 @@ node('cico-workspace') {
 
 	try {
 		stage('prepare bare-metal machine') {
-			if ("${ghprbPullId}".length() != 0) {
+			if (params.ghprbPullId != null) {
 				ref = "pull/${ghprbPullId}/head"
 			}
 			sh 'scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ./prepare.sh root@${CICO_NODE}:'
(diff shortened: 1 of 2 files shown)


nixpanic:
@Yuggupta27 please have a look, this makes it a little cleaner and there is no need to add the parameters to the `.yaml` configuration.

containerized-tests.groovy
```diff
@@ -30,7 +30,7 @@ node('cico-workspace') {
 
 	try {
 		stage('prepare bare-metal machine') {
-			if ("${ghprbPullId}".length() != 0) {
+			if (params.ghprbPullId != null) {
```
Yuggupta27, line 33: Will it work even if ghprbPullId is not set?

nixpanic: I do not think it will be possible to be empty anymore, once remove from the parameters in the `.yaml` file. The github-pull-request builder plugin will set it to a number of the PR, or it will not be set in case the job is started an other way.

Yuggupta27: Great!

nixpanic: yes, it should, that is the whole idea :)

PR #1035, opened 2020-05-11
feat: Add rados namespace support to rbd plugin

mehdy:
# Describe what this PR does #

Make sure to operate within the namespace if any given when dealing
with rbd images and snapshots and volume and snapshot journals.

Add some basic e2e tests using a namespace.

## Is there anything that requires special attention ##

Do you have any questions?
No.

Is the change backward compatible?
Yes.

Are there concerns around backward compatibility?
No.

Provide any external context for the change, if any.

* [rbd namespace](https://docs.ceph.com/docs/master/man/8/rbd/?highlight=namespace)

## Related issues ##

Mention any github issues relevant to this PR. Adding below line
will help to auto close the issue once the PR is merged.

Fixes: #798 

## Future concerns ##

List items that are not part of the PR and do not impact it's
functionality, but are work items that can be taken up subsequently.

PR #1116, opened 2020-05-28
rbd: Fix error checking mistake in CreateVolume

mehdy:
deleteErr should be checked if it's nil instead of err

No.

Do you have any questions?
No.

Yes.

No.

## Related issues ##
None.

## Future concerns ##
None.


mehdy:
@nixpanic a small change from #1035 :))

humblec:
@Madhu-1 ptal

PR #1118, opened 2020-05-29
ci: Use bionic dist for test environments

mehdy:
As bionic uses a newer version of linux we'd also be using a newer version of krbd.

No.

Do you have any questions?
No.

Yes.

No.

## Related issues ##

This is required for using the rados namespace with krbd
#798 

## Future concerns ##
I'm not sure if it's okay to let the arm64 still be using xenial.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 ---
 # need for docker build
 sudo: true
-dist: xenial
+dist: bionic
 
 addons:
   apt:
     packages:
-      - realpath
+      - coreutils
       - ruby
       - socat
 services:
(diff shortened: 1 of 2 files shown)


mehdy:
a small change from #1035

nixpanic:
The error in the CI is weird, I do not see a change to `go.mod` in this PR:
```
running: go mod verify
go: github.com/golang/protobuf@v1.3.5: reading github.com/golang/protobuf/go.mod at revision v1.3.5: unknown revision v1.3.5
ERROR: go.mod was modified by 'go mod verify'
```

Will restart the test, maybe there was some intermediate failure while verifying the protobuf dependency.

/test ci/centos/containerized-test

nixpanic:
The command is `/test ci/centos/containerized-tests` (with the **s**)

PR #1122, opened 2020-05-29
build: use CURDIR instead of PWD

nixpanic:

`make -C ${GOPATH/src/github.com/ceph/ceph-csi container...` commands
fail as the PWD variable gets substituted with the path where `make` is
called, not with the path of the Makefile that is passed with the -C
option.

The CURDIR variable is expended to the correct path, so `make -C ...`
works for all make targets now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,10 +112,10 @@ cephcsi: check-env
 
 .PHONY: containerized-build containerized-test
 containerized-build: .devel-container-id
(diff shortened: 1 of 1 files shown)


Yuggupta27:
Since the file/dir owned outside the user namespace were not allowed to be relabelled due to incorrect path; this will rectify that issue, right?

nixpanic:
> Since the file/dir owned outside the user namespace were not allowed to be relabelled due to incorrect path; this will rectify that issue, right?

Yes. The volume that gets mapped into the container should have the sources of ceph-csi and not be `/root` or any other path from where `make -C ...` is executed.

nixpanic:
@Yuggupta27 if you review this, @Madhu-1 can add the `ready-to-merge` label :wink:

PR #1124, opened 2020-06-02
ci: Add a job to run e2e tests on a single node cluster

Yuggupta27:
Job runs e2e tests on a single node minikube cluster on centos ci.

Fixes: #1276

### DIFF
--- a/mini-e2e.groovy
+++ b/mini-e2e.groovy
@@ -1,0 +1,73 @@
+def cico_retries = 16
+def cico_retry_interval = 60
+// temporary git repository for testing purpose
+def ci_git_repo = 'https://github.com/nixpanic/ceph-csi'
+def ci_git_branch = 'mini-e2e'
+def git_repo = 'https://github.com/ceph/ceph-csi'
+def ref = "master"
+
+def ssh(cmd) {
+	sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} '${cmd}'"
+}
+
+node('cico-workspace') {
+	stage('checkout ci repository') {
+		git url: "${ci_git_repo}",
+			branch: "${ci_git_branch}",
+			changelog: false
+	}
+
+	stage('reserve bare-metal machine') {
+		def firstAttempt = true
+		retry(30) {
+			if (!firstAttempt) {
+				sleep(time: 5, unit: "MINUTES")
+			}
+			firstAttempt = false
+			cico = sh(
+				script: "cico node get -f value -c hostname -c comment --release=8 --retry-count=${cico_retries} --retry-interval=${cico_retry_interval}",
+				returnStdout: true
+			).trim().tokenize(' ')
+			env.CICO_NODE = "${cico[0]}.ci.centos.org"
+			env.CICO_SSID = "${cico[1]}"
+		}
+	}
+
+	try {
+		stage('prepare bare-metal machine') {
+			if (params.ghprbPullId != null) {
+				ref = "pull/${ghprbPullId}/head"
+			}
+			sh 'scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ./prepare.sh ./single-node-k8s.sh root@${CICO_NODE}:'
+			ssh "./prepare.sh --workdir=/opt/build/go/src/github.com/ceph/ceph-csi --gitrepo=${git_repo} --ref=${ref}"
+		}
+		stage('build artifacts') {
+			// build container image
+			ssh 'cd /opt/build/go/src/github.com/ceph/ceph-csi && make image-cephcsi GOARCH=amd64 CONTAINER_CMD=podman'
+			// build e2e.test executable
+			ssh 'cd /opt/build/go/src/github.com/ceph/ceph-csi && make containerized-build CONTAINER_CMD=podman TARGET=e2e.test'
+		}
+		stage('deploy k8s v1.18.3 and rook') {
+			timeout(time: 30, unit: 'MINUTES') {
+				ssh './single-node-k8s.sh --k8s-version=v1.18.3'
+			}
+		}
+		stage('run e2e') {
+			timeout(time: 60, unit: 'MINUTES') {
+				ssh 'cd /opt/build/go/src/github.com/ceph/ceph-csi && make run-e2e'
+			}
+		}
+	}
+
+	catch(exc) {
+		stage('debug time!') {
+			ssh 'sleep 8h'
+		}
+	}
+
+	finally {
+		stage('return bare-metal machine') {
+			sh 'cico node done ${CICO_SSID}'
+		}
+	}
+}
--- a/single-node-k8s.sh
+++ b/single-node-k8s.sh
@@ -1,0 +1,147 @@
+#!/bin/bash
+#
+# Start minukube with the selected k8s version and deploy rook.
+#
+# This script uses helper scripts from the main ceph-csi branch.
+#
+
+# fail when a command returns an error
+set -e -o pipefail
+
+ARGUMENT_LIST=(
+    "k8s-version"
+)
+
+opts=$(getopt \
+    --longoptions "$(printf "%s:," "${ARGUMENT_LIST[@]}")help" \
+    --name "$(basename "${0}")" \
+    --options "" \
+    -- "$@"
+)
+ret=$?
+
+if [ ${ret} -ne 0 ]
+then
+    echo "Try '--help' for more information."
+    exit 1
+fi
+
+eval set -- "${opts}"
+
+while true; do
+    case "${1}" in
+    --help)
+        shift
+        echo "Options:"
+        echo "--help|-h                 specify the flags"
+        echo "--k8s-version             specify the kubernetes version"
+
+        echo " "
+        echo "Sample Usage:"
+        echo "./single-node-k8s.sh --k8s-version=v1.18.3"
+        exit 0
+        ;;
+    --k8s-version)
+        shift
+        k8s_version=${1}
+        ;;
+    --)
+        shift
+        break
+        ;;
+    esac
+    shift
+done
+
+test -n "${k8s_version}" || { echo "k8s_version is not set"; exit 1;}
+
+function set_env() {
+    export GOPATH="/opt/build/go"
+    # build.env is not part of the ci/centos branch, so shellcheck can not find
+    # it and will cause a failure if not annotated with /dev/null.
+    # shellcheck source=/dev/null
+    source "${GOPATH}"/src/github.com/ceph/ceph-csi/build.env
+    export GO111MODULE="on"
+    export TEST_COVERAGE="stdout"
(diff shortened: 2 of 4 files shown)


mini-e2e.groovy
```diff
@@ -0,0 +1,66 @@
+def cico_retries = 16
+def cico_retry_interval = 60
+def ci_git_repo = 'https://github.com/Yuggupta27/ceph-csi'
```
nixpanic, line 3: Ideally you place a comment here, so that it is clear the URL/branch needs to change

mini-e2e.groovy
```diff
@@ -0,0 +1,66 @@
...
+def ci_git_branch = 'mini-e2e'
+def ref = "master"
+
+node('cico-workspace') {
+
```
nixpanic, line 8: unneeded empty line after opening `{`

mini-e2e.groovy
```diff
@@ -0,0 +1,66 @@
...
+			if (params.ghprbPullId != null) {
+				ref = "master"
+				//ref = "pull/${ghprbPullId}/head"
+			}
+			sh 'scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ./prepare.sh ./single-node-k8s.sh root@${CICO_NODE}:'
+			sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./prepare.sh --workdir=/opt/build/go/src/github.com/ceph/ceph-csi --gitrepo=${ci_git_repo} --ref=${ref}"
+		}
+		stage('e2e') {
+			timeout(time: 25, unit: 'MINUTES') {
+				parallel kubeversion1: {
```
nixpanic, line 42: a name `kubeversion1` does not tell a lot, is it possible to use a string here?

mini-e2e.groovy
```diff
@@ -0,0 +1,66 @@
...
+					node ('cico-workspace') {
+						sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./single-node-k8s.sh --ref=${ref} --k8s_version=v1.17.5 --test_type=functest"
+					}
+				},
+				kubeversion2: {
+					node('cico-workspace') {
+						sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./single-node-k8s.sh --ref=${ref} --k8s_version=v1.16.9 --test_type=functest"
```
nixpanic, line 49: Now you are running multiple instances of the script on the same bare-metal system. Is it expected to work, and will the scripts not step on each others toes?

single-node-k8s.sh
```diff
@@ -0,0 +1,130 @@
...
+  fi
+}
+
+# Set environment variables
+set_env
+
+mkdir -p "$GOPATH"/bin
+# sudo rpm --import 'https://download.ceph.com/keys/release.asc'
+curl https://download.ceph.com/keys/release.asc | sudo apt-key add -
+sudo apt-add-repository "deb https://download.ceph.com/debian-nautilus $(lsb_release -sc) main"
```
nixpanic, line 95: make sure to install packages with yum/dnf, apt commands are for Debian based distributions

single-node-k8s.sh
```diff
@@ -0,0 +1,130 @@
...
+# yamllint disable rule:line-length
+test "$(arch)" != "aarch64" || sudo apt-add-repository "deb https://chacra.ceph.com/r/ceph/nautilus/af06652dc9b2da8c6aadbbecdfafdc7e235abe7d/ubuntu/xenial/flavors/default/ xenial main"
+
+# yamllint enable rule:line-length
+sudo apt-get -qq update
+# only the arm64 fallback repo is unsigned and needs --allow-unauthenticated
+sudo apt-get -y --allow-unauthenticated install librados-dev librbd-dev
+
+if [ "$test_type" = "functest" ]; then
+    sh scripts/skip-doc-change.sh "${ref}"
```
nixpanic, line 107: no need for the `sh` prefix, scripts should be marked executable before they are checked into the repository (`chmod +x <file>`)

single-node-k8s.sh
```diff
@@ -0,0 +1,130 @@
...
+    check_errs $? "Error occured while checking document changes"
+
+    sh make image-cephcsi
```
nixpanic, line 110: maybe run the build in a container? that should prevent the need to install additional development tools and libraries

single-node-k8s.sh
```diff
@@ -0,0 +1,130 @@
...
+    check_errs $? "Error occured during image creation"
+
+    sh scripts/travis-functest.sh "${k8s_version}"
```
nixpanic, line 113: `travis-functest.sh` calls `sudo` for some things. When running these tests as root (as you do now), the script should not add `sudo` for commands as that cleans exported variables from the environment.

mini-e2e.groovy
```diff
@@ -0,0 +1,69 @@
...
+			sh 'scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ./prepare.sh ./single-node-k8s.sh root@${CICO_NODE}:'
+			sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./prepare.sh --workdir=/opt/build/go/src/github.com/ceph/ceph-csi --gitrepo=${ci_git_repo} --ref=${ref}"
+		}
+		stage('e2e kube v1.17.5') {
+			// timeout is kept minimal as of now to catch errors quickly. It will be extended when pipeline has expected execution
+			timeout(time: 25, unit: 'MINUTES') {
+					node ('cico-workspace') {
+						sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./single-node-k8s.sh --ref=${ref} --k8s_version=v1.17.5 --test_type=functest"
+					}
+				}
```
nixpanic, line 46: indention seem to be a bit off

mini-e2e.groovy
```diff
@@ -0,0 +1,69 @@
...
+	try {
+		stage('prepare bare-metal machine') {
+			if (params.ghprbPullId != null) {
+				ref = "master"
+				//ref = "pull/${ghprbPullId}/head"
+			}
+			sh 'scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ./prepare.sh ./single-node-k8s.sh root@${CICO_NODE}:'
+			sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./prepare.sh --workdir=/opt/build/go/src/github.com/ceph/ceph-csi --gitrepo=${ci_git_repo} --ref=${ref}"
+		}
+		stage('e2e kube v1.17.5') {
```
nixpanic, line 40: This will run the stages in sequence. It is fine to only run one e2e in this job, and we can configure multiple jobs with this groovy file with different parameters. Assume that the Kubernetes version gets passed as a parameter (like `params.ghprbPullId`), default to a version of your choice if not set, and only run the e2e for this version.

We can add the parameters when adding the `.yaml` file for the job (under the `jobs/` directory)

single-node-k8s.sh
```diff
@@ -0,0 +1,132 @@
...
+
+if [ "$test_type" = "functest" ]; then
+    cd /opt/build/go/src/github.com/ceph/ceph-csi
+    # sh scripts/skip-doc-change.sh "${ref}"
+    # check_errs $? "Error occured while checking document changes"
+
+    make containerized-build
+    check_errs $? "Error occured during image creation"
+
+    yum install https://download.docker.com/linux/centos/7/x86_64/stable/Packages/docker-ce-18.03.1.ce-1.el7.centos.x86_64.rpm -y
```
nixpanic, line 101: Is there a `docker-release` RPM of some kind? Installing that should make the repository available, and you can `yum -y install docker-ce` then. No need to list versions or architectures that way.

Yuggupta27: will ```yum -y install docker-ce``` work fine even if repo is not configured?

mini-e2e.groovy
Yuggupta27: Thanks, I've made the changes. I'll also add a parameter so that we can reuse the script for different k8s versions.

single-node-k8s.sh
nixpanic: `yum -y install docker-ce` will only work if the repository is available. But there is an other way too:

```
# TODO: use podman, but minikube currently depends on Docker
yum -y install docker-latest
# minikube/kubelet is configured with cgroupfs
sed 's/native.cgroupdriver=systemd/native.cgroupdriver=cgroupfs/' < /usr/lib/systemd/system/docker-latest.service > /etc/systemd/system/docker.service
# start the docker service (actually docker-latest, but minikube checks for 'docker.service')
systemctl daemon-reload
systemctl enable --now docker
```

This does not depend on any additional repository, and the docker version will get updated automatically.

single-node-k8s.sh
```diff
@@ -0,0 +1,130 @@
...
+
+    yum install -y systemd-resolved
+    systemctl start systemd-resolved
+    ls -l /run/systemd/resolve/resolv.conf
```
nixpanic, line 105: You do not really need to check for the file, and with #1128 installing syste,d-resolved should not be needed anymore

Yuggupta27: Okay. Thanks.

single-node-k8s.sh
```diff
@@ -0,0 +1,121 @@
...
+    exit "${1}"
+  fi
+}
+
+# Set environment variables
+set_env
+mkdir -p "$GOPATH"/bin
+
+if [ "$test_type" = "functest" ]; then
+    cd /opt/build/go/src/github.com/ceph/ceph-csi || exit
```
nixpanic, line 94: It is possible a different directory is used, use `${GOPATH}/src/github.com/ceph/ceph-csi` instead

single-node-k8s.sh
```diff
@@ -0,0 +1,129 @@
...
+# Disable swap
+swapoff -a
+sed -i '/ swap / s/^\(.*\)$/#\1/g' /etc/fstab
+
+if [ "$test_type" = "functest" ]; then
+    cd ${GOPATH}/src/github.com/ceph/ceph-csi || exit
+    # sh scripts/skip-doc-change.sh "${ref}"
+    # check_errs $? "Error occured while checking document changes"
+
+    make containerized-build
```
nixpanic, line 106: This should be `make image-cephcsi`

nixpanic:
I've just pushed the current version that was used for testing #1140. Will clean things up, and make CI happy about this PR.

Yuggupta27:
> I've just pushed the current version that was used for testing #1140. Will clean things up, and make CI happy about this PR.

Thanks Niels!

nixpanic:
The CI failure with ci/centos/job-validation is addressed with #1299

nixpanic:
@mergifyio rebase

PR #1125, opened 2020-06-02
util: remove unused GetImageMeta() and SetImageMeta()

nixpanic:

The GetImageMeta() and SetImageMeta() functions are never called. There is no reason to keep them around.

This removes the need to replace the command execution by go-ceph functions.


nixpanic:
> LGTM , why it was not caught in CI :(

There functions are in the `util` package and are exported, so anything external from the package could use them. If they were lowercase and in the `rbd` package, CI would catch it.

nixpanic:
@Yuggupta27 approved, adding `ready-to-merge` label

humblec:
@nixpanic @Madhu-1 my suggestion would be placing `ready-to-merge` by the first approver may be. Because when first approver review, he may not be knowing how this is going to be merged.

PR #1126, opened 2020-06-02
util: implement CreateObject() and RemoveObject with go-ceph

nixpanic:

Implement CreateObject() and RemoveObject with go-ceph to remove further command executions in the journalling path.


nixpanic:
@phlogistonjohn more go-ceph stuff :wink:

internal/util/cephcmds.go
```diff
@@ -249,27 +249,27 @@ func CreateObject(ctx context.Context, monitors string, cr *Credentials, poolNam
...
-		"-p", poolName,
-		"rm", oMapName,
+	conn, err := connPool.Get(monitors, cr.ID, cr.KeyFile)
+	if err != nil {
+		return err
+	}
+	defer connPool.Put(conn)
+
+	ioctx, err := conn.OpenIOContext(poolName)
+	if err != nil {
```
Madhu-1, line 259: if the pool is not present, we may need to return the `ErrObjectNotFound`

internal/util/cephcmds.go
```diff
@@ -219,27 +219,27 @@ func RemoveOMapKey(ctx context.Context, monitors string, cr *Credentials, poolNa
...
 func CreateObject(ctx context.Context, monitors string, cr *Credentials, poolName, namespace, objectName string) error {
-	// Command: "rados <options> create objectName"
-	args := []string{
-		"-m", monitors,
-		"--id", cr.ID,
-		"--keyfile=" + cr.KeyFile,
-		"-c", CephConfigPath,
-		"-p", poolName,
-		"create", objectName,
+	conn, err := connPool.Get(monitors, cr.ID, cr.KeyFile)
```
Madhu-1, line 222: how about making use of funcitons already defined in https://github.com/ceph/ceph-csi/blob/master/internal/util/connection.go file?

internal/util/cephcmds.go
nixpanic: In case `ErrPoolNotFound` is returned, it will not get translated to `ErrObjectNotFound`.

internal/util/cephcmds.go
```diff
@@ -219,27 +219,31 @@ func RemoveOMapKey(ctx context.Context, monitors string, cr *Credentials, poolNa
...
-		"-p", poolName,
-		"create", objectName,
+	conn := ClusterConnection{}
+	err := conn.Connect(monitors, cr)
+	if err != nil {
+		return err
+	}
+	defer conn.Destroy()
+
+	ioctx, err := conn.GetIoctx(poolName)
```
Madhu-1, line 229: same comment as https://github.com/ceph/ceph-csi/pull/1129#discussion_r434992638

nixpanic: I am not sure there is a benefit in caching the IOContext in addition to the ClusterConnection. If you want, open an issue for it so we can investigate and improve that at a later time.

Issue #1121, opened 2020-05-29
Travis CI fails with minikube errors, unable to launch Kubernetes


nixpanic:
# Describe the bug #

Travis CI fails to launch Kubernetes from within minikube. This might be caused with #1118 merged?

# Environment details #

- Kernel version : whatever comes with Ubuntu Bionic?
- Kubernetes cluster version : 1.16, 1.17

# Steps to reproduce #

1. post a PR, see the Travis CI job fail

# Actual results #

```
$ scripts/travis-functest.sh v1.16.9 || travis_terminate 1;
Installing minikube. Version: v1.6.0
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 43.4M  100 43.4M    0     0  80.5M      0 --:--:-- --:--:-- --:--:-- 80.6M
Installing kubectl. Version: v1.16.9
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 40.9M  100 40.9M    0     0   151M      0 --:--:-- --:--:-- --:--:--  151M
prepare minikube to support pod security policies
starting minikube with kubeadm bootstrapper
* minikube 1.10.1 is available! Download it: https://github.com/kubernetes/minikube/releases/tag/v1.10.1
* To disable this notice, run: 'minikube config set WantUpdateNotification false'

! minikube v1.6.0 on Ubuntu 16.04
  - KUBECONFIG=/home/travis/.kube/config
  - MINIKUBE_VERSION=v1.6.0
* Selecting 'none' driver from user configuration (alternates: [])
! The 'none' driver does not respect the --memory flag
* Running on localhost (CPUs=2, Memory=7976MB, Disk=69422MB) ...
* OS release is Ubuntu 16.04.6 LTS
* Preparing Kubernetes v1.16.9 on Docker '18.06.0-ce' ...
  - apiserver.enable-admission-plugins=PodSecurityPolicy
  - kubelet.resolv-conf=/run/systemd/resolve/resolv.conf
* Downloading kubelet v1.16.9
* Downloading kubeadm v1.16.9
* Pulling images ...
* Launching Kubernetes ... 
* 
[...]

PR #1128, opened 2020-06-02
ci: pass target filename of /etc/resolv.conf in case it is a symlink

nixpanic:

In some Linux distributions the /etc/resolv.conf file is a symlink. This
file gets included in the Kubernetes containers and will be used for
resolving hostnames. By including the symlink, it is possible that that
target file is not available in the container(s). This will cause
problems when resolving hostnames, and Kubernetes will not get deployed.

The default minikube VM provides /run/systemd/resolve/resolv.conf, with
/etc/resolv.conf being a symlink. Therefor, it is needed to pass the
`--extra-config=kubelet.resolv-conf=..` parameter to `kubeadm`.

In case minikube is started with `--vm-driver=none`, the local
/etc/resolv.conf will be used for inclusion in the Kubelet container. If
this is a symlink, the final destination should get passed with
`--extra-config=kubelet.resolv-conf=..` so that a working hostname
resolution configuration is available in the container.

## Related issues ##

Updates: #1121

### DIFF
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -78,8 +78,19 @@ fi
 K8S_FEATURE_GATES=${K8S_FEATURE_GATES:-"BlockVolume=true,CSIBlockVolume=true,VolumeSnapshotDataSource=true,ExpandCSIVolumes=true"}
 
 #extra-config for kube https://minikube.sigs.k8s.io/docs/reference/configuration/kubernetes/
-EXTRA_CONFIG=${EXTRA_CONFIG:-"--extra-config=apiserver.enable-admission-plugins=PodSecurityPolicy \
-  --extra-config=kubelet.resolv-conf=/run/systemd/resolve/resolv.conf"}
+EXTRA_CONFIG=${EXTRA_CONFIG:-"--extra-config=apiserver.enable-admission-plugins=PodSecurityPolicy"}
+
+# kubelet.resolv-conf needs to point to a file, not a symlink
+# the default minikube VM has /etc/resolv.conf -> /run/systemd/resolve/resolv.conf
+RESOLV_CONF='/run/systemd/resolve/resolv.conf'
+if [[ "${VM_DRIVER}" == "none" ]] && [[ ! -e "${RESOLV_CONF}" ]]; then
+	# in case /run/systemd/resolve/resolv.conf does not exist, use the
+	# standard /etc/resolv.conf (with symlink resolved)
+	RESOLV_CONF="$(readlink -f /etc/resolv.conf)"
+fi
+# TODO: this might overload --extra-config=kubelet.resolv-conf in case the
+# caller did set EXTRA_CONFIG in the environment
+EXTRA_CONFIG="${EXTRA_CONFIG} --extra-config=kubelet.resolv-conf=${RESOLV_CONF}"
 
 #extra Rook configuration
 ROOK_BLOCK_POOL_NAME=${ROOK_BLOCK_POOL_NAME:-"newrbdpool"}


scripts/minikube.sh
```diff
@@ -95,6 +95,9 @@ up)
 
     enable_psp
 
+    # /run/systemd/resolve/resolv.conf is required
```
Madhu-1, line 98: I have not seen any issues in atleast in my local deployment, everything works even without this.

nixpanic: Really? And you have the /run/systemd/resolve/resolv.conf file on the host where you run the containers? Was this on a cleanly installed system, or were there possibly some Kubernetes containers still running? Some distributions provide the syetemd-resolved service by default, others do not.

The `EXTRA_CONFIG` points to `--extra-config=kubelet.resolv-conf=/run/systemd/resolve/resolv.conf` and without that file, kubelet should not be able to resolve hostnames to download further container images or similar.

nixpanic: Alternatively we could check if the file exists, and only then append the parameter to `EXTRA_CONFIG`. Is that something you would prefer?

Madhu-1: Note this is the vm created from minikube where the specify the vmdriver, currently i use kvm2.

Madhu-1: > Alternatively we could check if the file exists, and only then append the parameter to `EXTRA_CONFIG`. Is that something you would prefer?

yes that would be good check this only the extra config is present

Madhu-1: how you will come to know that this file is missing inside a VM? currently, this is checked only on the host machine if the VM driver is none

nixpanic: We hit the problem with `--vm-driver=none`, so in that case it is easy to verify. If the default VM that minikube starts has the file, it should cover the common scenarios. (You can pass `--iso-url=centos`, I think, to install a CentOS VM. But we do not use that in our CI.)

Madhu-1: okay, in that case, we will run this check if the VM driver is none and extra config is provided isn't it?

nixpanic: Updated this PR with a different approach. Please review again @Madhu-1 @Yuggupta27

scripts/minikube.sh
```diff
@@ -78,8 +78,15 @@ fi
...
-  --extra-config=kubelet.resolv-conf=/run/systemd/resolve/resolv.conf"}
+EXTRA_CONFIG=${EXTRA_CONFIG:-"--extra-config=apiserver.enable-admission-plugins=PodSecurityPolicy"}
+
+# kubelet.resolv-conf needs to point to a file, not a symlink
+# the default minikube VM has /etc/resolv.conf -> /run/systemd/resolve/resolv.conf
+RESOLV_CONF='/run/systemd/resolve/resolv.conf'
+if [[ "${VM_DRIVER}" == "none" ]]; then
+	RESOLV_CONF="$(readlink -f /etc/resolv.conf)"
+fi
+EXTRA_CONFIG="${EXTRA_CONFIG} --extra-config=kubelet.resolv-conf=${RESOLV_CONF}"
```
Madhu-1, line 89: if the EXTRA_CONFIG which is user-defined already has this param what happens now?

nixpanic: I am not sure which would get used by kubeadm. Do you only want to add `--extra-config=kubelet.resolv-conf` in case `EXTRA_CONFIG` is completely empty?

Madhu-1: we can add it to EXTRA_CONFIG if this flag is not present in it?

nixpanic: We can, but it will get more complicated than needed. Who would pass a different `-extra-config=kubelet.resolv-conf=..` parameter?!

Madhu-1: no idea, but we can pass this as per the code, in that case, this will break isn't it? if we want to fix it later, put a TODO.

nixpanic: I'll add a `TODO` note. In case someone reports a problem, we'll fix it at that time.

nixpanic:
@Yuggupta27 can you have a look at this again?

Yuggupta27:
> @Yuggupta27 can you have a look at this again?

Sure.

nixpanic:
It seems Ubuntu is a little exceptional:
```console
vagrant@ubuntu1804:~$ ls -l /etc/resolv.conf 
lrwxrwxrwx 1 root root 39 May 21 00:51 /etc/resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
vagrant@ubuntu1804:~$ ls -l /run/systemd/resolve/
total 8
-rw-r--r-- 1 systemd-resolve systemd-resolve 727 Jun  3 10:58 resolv.conf
-rw-r--r-- 1 systemd-resolve systemd-resolve 715 Jun  3 10:58 stub-resolv.conf
vagrant@ubuntu1804:~$ cat /run/systemd/resolve/resolv.conf
# This file is managed by man:systemd-resolved(8). Do not edit.
#
# This is a dynamic resolv.conf file for connecting local clients directly to
# all known uplink DNS servers. This file lists all configured search domains.
#
# Third party programs must not access this file directly, but only through the
# symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a different way,
# replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.
...
```

Reading the symlink is not correct, as it points to a different DNS server on localhost (which may not work from within the container). The upstream DNS servers need to be used in the running containers.

Will check if /run/systemd/resolve/resolv.conf exists, and prefer that instead of using `readlink`.

nixpanic:
@Yuggupta27 could you check the last update too? This works for me on CentOS and Ubuntu, but I would like your confirmation that it helps your e2e tests as well.

Yuggupta27:
> @Yuggupta27 could you check the last update too? This works for me on CentOS and Ubuntu, but I would like your confirmation that it helps your e2e tests as well.

Thanks @nixpanic, it works fine for installing minikube on Centos CI too.

nixpanic:
Adding `ready-to-merge` for Yugs review and testing.

@humblec @Madhu-1 could you review this so the CentOS CI jobs can use it?

PR #1129, opened 2020-06-03
rbd: keep rados.IOContext for re-use

nixpanic:

This prevents the need to open the IOContext for additional operations
on the image.

It also addresses a leak of the IOContext in case `rbdVolume.open()` was
called. The method only returned the `rbd.Image` without the possibility
to close the related IOContext.

This only affected the GetMetadata() and SetMetadata() calls (f814bd72e5), which have never been part of a release.

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -104,6 +104,8 @@ type rbdVolume struct {
 
 	// conn is a connection to the Ceph cluster obtained from a ConnPool
 	conn *util.ClusterConnection
+	// an opened IOContext, call .openIoctx() before using
+	ioctx *rados.IOContext
 }
 
 // rbdSnapshot represents a CSI snapshot and its RBD snapshot specifics
@@ -151,6 +153,9 @@ func (rv *rbdVolume) Connect(cr *util.Credentials) error {
 // Destroy cleans up the rbdVolume and closes the connection to the Ceph
 // cluster in case one was setup.
 func (rv *rbdVolume) Destroy() {
+	if rv.ioctx != nil {
+		rv.ioctx.Destroy()
+	}
 	if rv.conn != nil {
 		rv.conn.Destroy()
 	}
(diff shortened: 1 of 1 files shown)


internal/rbd/rbd_util.go
```diff
@@ -215,18 +219,33 @@ func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 	return nil
 }
 
+func (rv *rbdVolume) openIoctx() error {
```
Madhu-1, line 222: we need this IOctx for rados connection also isn't it? I and @ShyamsundarR plan was to openIoctx in the Main functions ie. CreateVolume, DeleteVolume, CreateSnapshot, DeleteSnapshot, etc. and pass it do calling functions. so that we don't need to fetch the context map in each sub-functions and use it. 

few comments regarding the same in older PR's are

https://github.com/ceph/ceph-csi/pull/777#discussion_r364067477
https://github.com/ceph/ceph-csi/pull/777#discussion_r377137785
https://github.com/ceph/ceph-csi/pull/729#discussion_r353780435

nixpanic: Getting the IOContext re-uses the connection that is already established. I do not think getting a new IOContext is an expensive operation. I'd prefer to keep the librbd/librados calls out of the main CreateVolume and other operations, instead have them be self-contained in methods of the rbdVolume type.

If it shows later on that the IOContext can benefit from a connection pool as well, we can always add that.

Currently, the `.open()` function gets an IOContext without it being able to get destroyed. This is a resource leak and by placing the IOContext in the rbdVolume, it can be destroyed (or reference minus-counted) when the rbdVolume is destroyed. A benefit of that is, that per rbdVolume object, the IOContext is created only once. Moving the IOContext to an even more shared structure would be an other enhancement (although I am unsure of the benefits).

Madhu-1: will open a separate issue to discuss more about it.

humblec:
LGTM.

PR #1134, opened 2020-06-05
build: fix kubelet issue#60987

chengyu-l:
When a Pod which has used PVC is deleted from kubuernetes cluster
we can get a error message repeatedly(Orphaned pod "1f415e-...-4b38"
found, but volume paths are still present on disk) from kubelet log.
this error message is caused by kubelet because it cannot clean the
parent path of targetPath when Pod is deleted from cluster
(targetPath like:
/var/lib/kubelet/pods/9cda187a-7fb3-11ea-80b3-246e968d4b38/volumes/kubernetes.io~csi/pvc-9ae9405c-7fb3-11ea-80b3-246e968d4b38/mount
). Ceph CSI can remove the parent path of targetPath first instead of
waiting kubelet to fix this issue.

Signed-off-by: chengyu-l <chengyu_l@126.com>

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -242,9 +242,8 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	err = os.Remove(targetPath)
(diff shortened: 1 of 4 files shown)


humblec:
There are few things which I would like to discuss here, so adding DNM for now.

chengyu-l:
is this method work?  @humblec

PR #1135, opened 2020-06-05
rbd: unmap rbd image if the mounting fails

Madhu-1:
There is a bug in the current code where the device path is always empty and the rbd image unmap never happens if node plugin fails to mount the rbd image to the staging path.

This is a fix to unmap the rbd image if some issue occurs after the rbd image is mapped.

Signed-off-by: Madhu Rajanna madhupr007@gmail.com

Backport of https://github.com/ceph/ceph-csi/pull/1078


Madhu-1:
Merging this one as CI is passed

PR #1140, opened 2020-06-05
Enable running e2e test for Travis CI on other environments

nixpanic:

With these changes, it becomes possible to run `minikube` with the `kvm2` driver and attach additional disks to the VM. This allows using Rook 1.3 and newer to be used for e2e testing.

PR #1124 has been used in combination with these changes to test the functionality. Once this PR is merged #1124 can get merged as well.

Most of these changes are required for deploying minikube/rook on a CentOS host.


nixpanic:
@Yuggupta27 the mini-e2e tests will be much easier to run with these changes

Yuggupta27:
> /test ci/centos/mini-e2e

I'll make some changes and re-run the same.

Yuggupta27:
@nixpanic , currently the ci git repo is set as https://github.com/Yuggupta27/ceph-csi ; which has a modified version of some scripts. 
Either we can change the target or verify the files that we are running.

nixpanic:
This is currently rebased on top of #1212 which is expected to get merged first.

scripts/minikube.sh
```diff
@@ -5,7 +5,7 @@
 function wait_for_ssh() {
     local tries=100
     while ((tries > 0)); do
-        if minikube ssh echo connected &>/dev/null; then
+        if /usr/local/bin/minikube ssh echo connected &>/dev/null; then
```
Madhu-1, line 8: minikube won't be installed always in this path isn't it?

scripts/rook.sh
```diff
@@ -26,7 +26,15 @@ function deploy_rook() {
 		kubectl create -f "${TEMP_DIR}/cluster-test.yaml"
 		rm -rf "${TEMP_DIR}"
 	else
-		kubectl create -f "${ROOK_URL}/cluster-test.yaml"
+		# add "mon_warn_on_pool_no_redundancy = false" to ceph.conf if missing
+		# see https://github.com/rook/rook/pull/5925 for upstream status
+		TEMP_DIR="$(mktemp -d)"
```
Madhu-1, line 31: we can move TEMP_DIR and curl before if check and after else we can do kubectl create and rm -rf as we can avoid duplication inside if and else

scripts/minikube.sh
nixpanic: That is where this script installs it. It is just that `/usr/local/bin` is not in the `PATH` for root on CentOS, so it will not be found, even after installing it there.

scripts/rook.sh
nixpanic: That can be done, but once rook/rook#5925 is merged (and backported) this should not be needed anymore. Do you still want me to change it?

scripts/minikube.sh
Madhu-1: will it cause any issue if the minikube is already present and set in PATH for a different location?

nixpanic: no, I do not think so

humblec: @nixpanic if we run from our system the path may be different. Hard ocding the path makes it difficult Isnt it ?

Madhu-1: This seems problematic if the minikube is already present in  `/usr/bin/minikube` with current changes it tries to install at `/usr/local/bin` even if the minikube is already present.

scripts/rook.sh
```diff
@@ -26,7 +26,15 @@ function deploy_rook() {
...
+		curl -o "${TEMP_DIR}"/cluster-test.yaml "${ROOK_URL}/cluster-test.yaml"
```
Yuggupta27, line 32: Can we curl directly in "/tmp" so that we won't need to delete it to? (As it's automatically cleaned up)

nixpanic: I'm not a big fan of creating the directory, but that is the pattern used here and in other scripts.
Also, cleaning up temporary files after use is recommended, you dont want to clutter /tmp with lots of unneeded stuff (it is also a RAM backed filesystem, so unneeded space has a cost).

Yuggupta27: Okay :+1:

scripts/minikube.sh
nixpanic: Well, the script already installed minikube in `/usr/local/bin`, I did not want to change that. I could add detection to see if a minikube executable is available, and use that instead. And then have it only install under `/usr/local/bin` if there is no other minikube executable.

Is that something you prefer?

nixpanic: This has been done now, all minikube calls use `${minikube}` which gets detected on start of the script.

nixpanic:
@Yuggupta27 @humblec @Madhu-1 this is now ready for re-review. All comments (AFAIK) have been addressed. The ci/centos/mini-e2e job depends on this PR. So further consumption of the CentOS CI (and new version of Rook) is blocked until this is merged.

Yuggupta27:
Thanks @nixpanic ,This looks much cleaner now :)

Yuggupta27:
Looks Good... Can't approve as I don't have approval permissions :)

scripts/rook.sh
```diff
@@ -26,7 +26,15 @@ function deploy_rook() {
...
+		if ! grep -q mon_warn_on_pool_no_redundancy "${TEMP_DIR}"/cluster-test.yaml; then
```
humblec, line 33: @nixpanic `mon_warn_on_pool_no_redundancy` can be declared as a const?

scripts/rook.sh
```diff
@@ -26,7 +26,15 @@ function deploy_rook() {
...
+			sed -i '/osd_pool_default_size =/a \    mon_warn_on_pool_no_redundancy = false' "${TEMP_DIR}"/cluster-test.yaml
```
humblec, line 34: also `osd_pool_default_size`

scripts/minikube.sh
```diff
@@ -125,38 +154,40 @@ up)
...
-        minikube start --memory="${MEMORY}" -b kubeadm --kubernetes-version="${KUBE_VERSION}" --vm-driver="${VM_DRIVER}" --feature-gates="${K8S_FEATURE_GATES}"
+        ${minikube} start --force --memory="${MEMORY}" -b kubeadm --kubernetes-version="${KUBE_VERSION}" --vm-driver="${VM_DRIVER}" --feature-gates="${K8S_FEATURE_GATES}"
         DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
-        kubectl apply -f "$DIR"/psp.yaml
-        minikube stop
+        ${minikube} kubectl -- apply -f "$DIR"/psp.yaml
+        ${minikube} stop
         # shellcheck disable=SC2086
-        minikube start --memory="${MEMORY}" -b kubeadm --kubernetes-version="${KUBE_VERSION}" --vm-driver="${VM_DRIVER}" --feature-gates="${K8S_FEATURE_GATES}" ${EXTRA_CONFIG}
+        ${minikube} start --force --memory="${MEMORY}" -b kubeadm --kubernetes-version="${KUBE_VERSION}" --vm-driver="${VM_DRIVER}" --feature-gates="${K8S_FEATURE_GATES}" ${EXTRA_CONFIG}
```
humblec, line 172: why we need to `force` start?

scripts/minikube.sh
```diff
@@ -125,38 +154,40 @@ up)
         install_kubectl
     fi
 
+    disable_storage_addons
```
humblec, line 157: any specific reason for above ?

humblec:
@nixpanic almost there. but have few comments..

scripts/minikube.sh
nixpanic: yes, see the commit message about it. minikube with kvm2 does not start when running as root, it needs `--force`

scripts/rook.sh
nixpanic: how do you mean? this is an option for ceph.conf.

scripts/rook.sh
nixpanic: Sorry, I don't get what you mean. `osd_pool_default_size` is present in the ceph.conf for in the `cluster-test.yaml`, the sed command adds the `mon_warn_on_pool_no_redundancy` option on the line after it (with added indention).

scripts/minikube.sh
nixpanic: It does not need to deploy any storage components that it deploys by default. Just saving resources by not having them. Not sure if there are potential conflicts, but they did clutter logs about failing to start or something.

nixpanic:
> @nixpanic almost there. but have few comments..

Thanks for the review. I think I answered your questions. If there are changes needed, please let me know.

scripts/rook.sh
humblec: I understand its an option.. What I meant was using something like `ROOK_CEPH_CLUSTER_VERSION_IMAGE_PATH` and using it in `sed` command.

scripts/rook.sh
humblec: same as above.

scripts/minikube.sh
humblec: its better to document it ?

scripts/rook.sh
nixpanic: I do not think there is any benefit from that. The value is not configurable or constructed dynamically. Replacing it with a variable makes the code more difficult to understand, imho. The value is used only twice, and they are close together, a typo is unlikely.

scripts/rook.sh
nixpanic: In that case, it would make sense to replace `cluster-test.yaml` with a variable. We do not do that because it will become unreadable. `osd_pool_default_size` is used only once, and there is no intention to make it configurable. It would not make the code any better.

scripts/minikube.sh
nixpanic: Will add some documentation for the function and update its commit message.

scripts/rook.sh
humblec: The other advantage is that, it (top level declared var) helps some one to understand there is some where the changes happening or its getting referred  and will be mindful about it. If these strings hide in between you may get miss those as well. I am fine either way.

scripts/minikube.sh
humblec: Thanks!

Issue #1042, opened 2020-05-12
Check for available Ceph development headers


nixpanic:
# Describe the feature you'd like to have #

The `scripts/check-env.sh` script should verify that the Ceph development headers are available and fail with a useful error when they are missing.

# What is the value to the end user? (why is it a priority?) #

New developers that run `make` will get useful error messages, and a hint to install additional tools, libraries or make other changes to their environment.

# How will we know we have a good solution? (acceptance criteria) #

When the development headers are not available, an error message is printed.

# Additional context #

Maybe also add a suggestion in case of failure, to run `make containerized-build` instead.

See-also: #1005

Madhu-1:
@Yuggupta27 would you like to take this one?

Yuggupta27:
Sure. Please assign.

PR #1141, opened 2020-06-08
build: add check for ceph development headers

Yuggupta27:
The script checks for the ceph development headers.
In case required packages are not found, script
suggests to run containerized build.

Fixes: #1042 

### DIFF
--- a/scripts/check-env.sh
+++ b/scripts/check-env.sh
@@ -3,6 +3,10 @@
 # Check the environment for dependencies and configuration
 #
 
+# check appropriate package installer to recommend corresponding packages
+RPM_CMD=$(command -v rpm)
+DPKG_CMD=$(command -v dpkg)
+
 # count errors, run script to the end before exiting
 ERRORS=0
 fail() {
@@ -10,9 +14,50 @@ fail() {
 	ERRORS=$((ERRORS+1))
 }
 
+# create a temp go file
+LIBCHECK=$(mktemp)
+mv "${LIBCHECK}" "${LIBCHECK}".go
+LIBCHECK=${LIBCHECK}.go
+
+# check for packages using a compile test
+cat << EOF > "${LIBCHECK}"
+package main
+
+/*
+#include <rados/librados.h>
+#include <rbd/librbd.h>
+*/
+import "C"
+
+func main() {
+	_ = C.LIBRADOS_VERSION_CODE
+	_ = C.LIBRBD_VER_MAJOR
+	_ = C.RBD_MAX_IMAGE_NAME_SIZE
+}
+EOF
+
 # check if 'go' is available
-[ -n "$(command -v go)" ] \
-	|| fail "could not find 'go' executable"
+if [ -n "$(command -v go)" ]; then
+	# in case of a failed execution, the user will be informed about
+	# the missing packages based on whether they are on rpm or debian
+	# based systems.
+	if ! go run "${LIBCHECK}" > /dev/null; then
+		if [ -n "${RPM_CMD}" ]; then
+			echo "Packages libcephfs-devel librbd-devel librados-devel need to be installed"
+		elif [ -n "${DPKG_CMD}" ]; then
+			echo "Packages libcephfs-dev librbd-dev librados-dev need to be installed"
+		else
+			fail "error can't verify Ceph development headers"
+		fi
+		echo "To build ceph-csi in a container: $ make containerized-build"
+	fi
+
+	# remove the temp file
+	rm -f "${LIBCHECK}"
+else
+	fail "could not find 'go' executable"
+	echo "To build ceph-csi in a container: $ make containerized-build"
+fi
 
 # parse the Golang version, return the digit passed as argument
 # 1.13.9 -> go_version 1 -> 1


scripts/check-env.sh
```diff
@@ -3,13 +3,37 @@
 # Check the environment for dependencies and configuration
 #
 
+# check appropriate package installer to verify the packages
+YUM_CMD=$(command -v yum)
```
nixpanic, line 7: On newer systems `yum` is replaced by `dnf`. So maybe you should check for `rpm` instead. You're executing `rpm` commands below, so it would be good to verify that the command exists in any case.

scripts/check-env.sh
```diff
@@ -3,13 +3,37 @@
...
+APT_GET_CMD=$(command -v apt-get)
```
nixpanic, line 8: You're not executing any `apt-get` command, so check for `dpkg` instead

scripts/check-env.sh
```diff
@@ -43,4 +67,17 @@ fi
 [ "$(go env CGO_ENABLED)" = '1' ] \
 	|| fail "CGO_ENABLED should be set to '1'"
 
+if [ -n "${YUM_CMD}" ]; then
+    check_rpm_pkg libcephfs-devel librbd-devel librados-devel
+elif [ -n "${APT_GET_CMD}" ]; then
+    check_deb_pkg libcephfs-dev librbd-dev librados-dev
+else
+    fail "error can't verify Ceph development headers"
```
nixpanic, line 75: You could also consider building a simple program like this:

```go
package main

/*
#include <rados/librados.h>
*/
import "C"

func main() {
    _ = C.LIBRADOS_VERSION_CODE
}
```

This only runs  (with `go run <filename>`) when librados.h is available. Use a wrong filename in the `#include` statement to test.

Yuggupta27: Thanks. I'll make the changes.

scripts/check-env.sh
```diff
@@ -43,4 +47,41 @@ fi
...
+mv "${LIBCHECK}" "${LIBCHECK}".go
+LIBCHECK=${LIBCHECK}.go
+
+# check for packages using a compile test
+cat << EOF > "${LIBCHECK}"
+package main
+
+/*
+#include <rados/librados.h>
+#include <rbd/lirbd.h>
```
nixpanic, line 61: typo?

scripts/check-env.sh
```diff
@@ -43,4 +47,41 @@ fi
...
+*/
+import "C"
+
+func main() {
+	_ = C.LIBRADOS_VERSION_CODE
```
nixpanic, line 66: You can check `/usr/include/rbd/librbd.h` and use one of the `#define`'s from it too. Building most likely fails if the file is missing, so it is not strictly required.

scripts/check-env.sh
```diff
@@ -43,4 +47,41 @@ fi
...
+}
+EOF
+
+# in case of a failed execution, the user will be informed about
+# the missing packages based on whether they are on rpm or debian
+# based systems.
+if ! go run "${LIBCHECK}" &> /dev/null; then
+	if [ -n "${RPM_CMD}" ]; then
```
nixpanic, line 74: indention is a little off

scripts/check-env.sh
```diff
@@ -43,4 +47,44 @@ fi
...
+	_ = C.LIBRADOS_VERSION_CODE
+	_ = C.LIBRBD_VER_MAJOR
+	_ = C.RBD_MAX_IMAGE_NAME_SIZE
+}
+EOF
+
+# in case of a failed execution, the user will be informed about
+# the missing packages based on whether they are on rpm or debian
+# based systems.
+if ! go run "${LIBCHECK}" > /dev/null; then
```
nixpanic, line 76: In case `go` is not available, there is no need to try and run this

Yuggupta27:
@Madhu-1 PTAL.

scripts/check-env.sh
```diff
@@ -3,16 +3,85 @@
...
+
 # count errors, run script to the end before exiting
 ERRORS=0
 fail() {
 	echo "${*}" > /dev/stderr
 	ERRORS=$((ERRORS+1))
 }
 
+ceph_version() {
+	TOOLBOX_POD=$(kubectl -n rook-ceph get pods -l app=rook-ceph-tools -o jsonpath='{.items[0].metadata.name}')
```
Madhu-1, line 18: This is not possible, we don't have/need kubernetes cluster or ceph cluster during development

Madhu-1: What I meant in yesterday discuss was , as you are checking the packages are installed, it's also good to check the version of packages installed

Yuggupta27: So do we have to skip ceph version check then?

Madhu-1: yeah we can skip it now, we can have an enhancement in follow up patch, maybe need to see how to do it in golang

Yuggupta27: Okay. Thanks to clear the misunderstanding. I'll make the changes.

scripts/check-env.sh
```diff
@@ -3,16 +3,61 @@
...
+	_ = C.LIBRBD_VER_MAJOR
+	_ = C.RBD_MAX_IMAGE_NAME_SIZE
+}
+EOF
+
 # check if 'go' is available
-[ -n "$(command -v go)" ] \
-	|| fail "could not find 'go' executable"
+if [ -n "$(command -v go)" ]; then
+
```
nixpanic, line 41: no need for this empty line

scripts/check-env.sh
```diff
@@ -3,16 +3,61 @@
...
+		else
+			fail "error can't verify Ceph development headers"
+		fi
+		echo "To build ceph-csi in a container: $ make containerized-build"
+	fi
+
+	# remove the temp file
+	rm -f "${LIBCHECK}"
+else
+	fail "could not find 'go' executable"
```
nixpanic, line 59: In this case, running the containerized build/test should still work, so recommend that,

PR #1145, opened 2020-06-08
Update the golang from 1.13.9 to 1.14.2

humblec:

### DIFF
--- a/deploy/cephcsi/image/Dockerfile
+++ b/deploy/cephcsi/image/Dockerfile
@@ -6,7 +6,7 @@ FROM ${BASE_IMAGE} as builder
 
 LABEL stage="build"
 
-ARG GOLANG_VERSION=1.13.9
+ARG GOLANG_VERSION=1.14.4
 ARG CSI_IMAGE_NAME=quay.io/cephcsi/cephcsi
 ARG CSI_IMAGE_VERSION=canary
 ARG GO_ARCH


humblec:
@Madhu-1 ptal.

humblec:
@nixpanic ptal.

humblec:
> Why not use 1.14.4 as that is the latest release?

thought of staying bit away from very latest, but I can use it too.
> 
> Instead of the `deploy` prefix, maybe use `build`.

Could use build too.

humblec:
@nixpanic addressed. PTAL.

humblec:
> This could be kind of risky I think, are we making use of go-1.14.x. if the new features added in go 1.14 (which we might use in cephcsi)might not be available in 1.13. the build may fail( i would like to keep 1.13 for now)

@madhu considering the CI passed, that gives the guarantee, Isnt it ? also are you aware of any specific changes which could break the system?

Madhu-1:
> > This could be kind of risky I think, are we making use of go-1.14.x. if the new features added in go 1.14 (which we might use in cephcsi)might not be available in 1.13. the build may fail( i would like to keep 1.13 for now)
> 
> @Madhu considering the CI passed, that gives the guarantee, Isnt it ? 
we need to add CI for both golang versions. and make sure everything works. this could be risky for production builds ( as mostly cephcsi will be built with go 1.13 as 1.14 is new)

>also are you aware of any specific changes which could break the system?
Currently nothing, we might miss something if we don't test cephcsi with go 1.13

deploy/cephcsi/image/Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM ${BASE_IMAGE} as builder
 
 LABEL stage="build"
 
-ARG GOLANG_VERSION=1.13.9
+ARG GOLANG_VERSION=1.14.4
```
Madhu-1, line 9: This will be overridden by https://github.com/ceph/ceph-csi/blob/master/Makefile#L146 ie 1.13.8

nixpanic:
Madhu has a point on the 1.13.x support. If we only test with 1.14.x, we may introduce new things that do not work with 1.13.x anymore. It might be a good idea to check what other projects (Kubernetes) have as guidelines for updating the Golang version.

humblec:
Closing this for avoiding a wrong push .

PR #1146, opened 2020-06-08
doc: correct mdlinter errors on doc

humblec:
mdlinter errors are fixed with this patch.


humblec:
@Madhu-1 @nixpanic ptal.

nixpanic:
Which errors are these? Do you know why the static-check does not report them?

humblec:
> @humblec can you please check why CI didn't catch this one?

@Madhu-1 @nixpanic I can check why it was not caught on CI. However as long as this change looks good we could let this get in.

nixpanic:
It would help if you can at least include the error messages from the linter. I am not sure the changes are correct/required.

Issue #446, opened 2019-06-26
create and delete a large number of snapshots  leaving  stale snapshots  in the backend


Madhu-1:
# Describe the bug #

create and delete a large number of snapshots  leaving  stale snapshots  in the backend
# Environment details #

- Image/version of Ceph CSI driver: canary
- helm chart version
- Kubernetes cluster version:1.14.2
- Logs

rbd plugin logs
[rbd.log](https://github.com/ceph/ceph-csi/files/3329777/rbd.log)

snapshotter logs

[snap.log](https://github.com/ceph/ceph-csi/files/3329776/snap.log)

# Steps to reproduce #

1. Setup details: '...'
1. Deployment to trigger the issue '....'
1. See error

# Actual results #

created 50 snapshots and deleted 50 snapshots, there are around 14 stale snapshots in the backend
# Expected behavior #

once we delete Kube snapshots there should not be any stale snapshots in the backend

Madhu-1:
@ShyamsundarR @humblec PTAL

ShyamsundarR:
There are few things happening here, primarily because the requests are timing out.

The snapshotter sidecar does NOT retry taking snapshots, instead it tries once and if the call times out, the `volumesnapshot` object gets an error. In the background the snapshot is actually created by the first RPC to request the same, but kubernetes does not have a record of the same (i.e the `SnapID`). Thus, when the snapshot is deleted, kubernetes only deletes the `volumesnapshot` object and does not invoke DeleteSnapshot RPC against the plugin. This hence leaks snapshots.

The lock fix patch (https://github.com/ceph/ceph-csi/pull/443) alleviates the problem, as it speeds up the operations (including snapshots) but there are still corner cases that may occur leaking snapshots.

BTW, Similarly there can be corner cases in PVC->PV creates and deletes, that can leak images as kubernetes never recorded a success from the plugin. I have seen this happen when working on the performance improvements with the locks. It needs some more analysis, but there are cases when on large RPC response times we may leak an image.

There is the `ReadyToUse` flag with snapshots that can help, as the CO is supposed to keep retrying once a response is sent with this set to false, to ensure this becomes ready at some point in time. We can only do this post creating the RADOS maps for the snapshot and possibly taking the snapshot, so if these calls take time we will again get back to leaking snapshots. Also, the sidecar code needs to be checked if this is retried as expected by the specification.

[...]

Issue #263, opened 2019-03-14
RBD maximum parent image chain for clones


dillaman:
When cloning an RBD image, it might be required to immediately flatten the image if it has more than 16 parents (ancestors). This is a limit of the kernel driver [1], but for IO performance you still wouldn’t want a potentially large chain of parent images to search through for read requests.

Note that Cinder will perform a flatten at a much more manageable 5 parents [2].

[1] 
https://github.com/torvalds/linux/blob/master/drivers/block/rbd.c#L92
[2] 
https://github.com/openstack/cinder/blob/9660f7010e3f54ec4c873cc3154cf275e046ca24/cinder/volume/drivers/rbd.py#L82

Madhu-1:
https://github.com/ceph/ceph-csi/pull/467 will take care of flattening the image

PR #1160, opened 2020-06-10
rbd: Implement snapshot and clone from snapshot

Madhu-1:
This Adds support for create,delete snapshot and creating a new rbd image from the snapshot.
    
 ##  Create a snapshot
    
* Create a temporary snapshot from the parent volume
* Clone a new image from a temporary snapshot with options --rbd-default-clone-format 2 --image-feature layering,deep-flatten
* Delete temporary snapshot created
* Create snapshot from clone image
* Check the image chain depth, if the Softlimit is reached Add a task Flatten the cloned image and return success. if the depth is reached hard limit Add a task Flatten the cloned image and return snapshot status ready as false
    
    ```bash
    layering,deep-flatten <RBD image for src k8s volume>@<random snap>
     <RBD image for temporary snap image>
    4) rbd snap create <RBD image for temporary snap image>@<RBD image for temporary snap image>
    5) check the depth, if the depth is greater than configured hard
    limit add a task to flatten the cloned image return snapshot status
    ready as false if the depth is greater than soft limit add a task
    to flatten the image and return success
    
    ```
    
## Create a clone from a snapshot
    
* Check the depth(n) of the cloned image if n>=(hard limit) Add a task to flatten the image and return ABORT (to avoid image leak)
    
    ```bash
    <k8s dst vol config> <RBD image for temporary snap image>@<random snap name>
     <RBD image for k8s dst vol>
    4) check the depth, if the depth is greater than the configured hard limit
    add a task to flatten the cloned image return ABORT error if the depth is
    greater than soft limit add a task to flatten the image and return success
    ```
    
## Delete snapshot or PVC
    
* Move the temporary cloned image to the trash
* Add task to remove the image from the trash
    
    ```bash
    1) rbd trash mv <cloned image>
    2) ceph rbd task trash remove <cloned image>
    ```
    
With earlier implementation to delete the image, we used to add a task to remove the image with new changes this cannot be done as the image may contain snapshots or links.
   
    
Note: Created a new PR to make review easy and addressed review comments in https://github.com/ceph/ceph-csi/pull/693 and https://github.com/ceph/ceph-csi/pull/467

closes #693 
closes #467 
closes https://github.com/ceph/ceph-csi/issues/263
closes https://github.com/ceph/ceph-csi/issues/446

The snapshot, creating a PVC from Restore and creating a PVC from PVC will be done in different PR's

TODO

- [x] Implement snapshot create, delete and create PVC from a snapshot
- [ ] Create PVC from PVC
- [ ] Fix issues with snapshot/clone (512) limits https://github.com/ceph/ceph-csi/issues/1098 
- [ ] Allow Parallel cloning and snapshots creation on one PVC/clone (Need to work on in-memory locking) 
- [ ] Improve/Add Documentation for snapshot and cloning
- [ ] Check Anything Needed to support datapool
- [ ] Add doc file for the commands
- [ ] If required some code cleanup
- [ ] Replace GetStoredImageID with GetImageAttributes in all places


Madhu-1:
Converted to Draft as scale testing is pending and also need to add TODO for remaining tasks

Madhu-1:
still need to do some smaller changes, marked as ready of review.

Madhu-1:
@dillaman PTAL

internal/rbd/rbd_util.go
```diff
@@ -335,38 +330,62 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
+	if pOpts.ImageID == "" {
+		pOpts.ImageID, err = pOpts.getImageID()
+		if err != nil {
+			return err
+		}
+	}
+	klog.V(4).Infof(util.Log(ctx, "rbd: delete %s using mon %s, pool %s"), image, pOpts.Monitors, pOpts.Pool)
 
-	klog.V(4).Infof(util.Log(ctx, "rbd: rm %s using mon %s"), pOpts, pOpts.Monitors)
+	var ioctx *rados.IOContext
```
nixpanic, line 355: just use `pOpts.open()` to get the imahe, and use `pOpts.ioctx` after that for the IOContext. No need to destroy the IOContext, as it will be cleaned up by `pOpts.Destroy()` at a later time.

You need to defer closing the image though.

internal/rbd/controllerserver.go
```diff
@@ -262,7 +294,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	return &csi.CreateVolumeResponse{Volume: volume}, nil
 }
 
-func createBackingImage(ctx context.Context, cr *util.Credentials, rbdVol *rbdVolume, rbdSnap *rbdSnapshot) error {
+func (cs *ControllerServer) createBackingImage(ctx context.Context, cr *util.Credentials, rbdVol *rbdVolume, rbdSnap *rbdSnapshot) error {
```
nixpanic, line 315: why make this a method for ControllerServer? it is not used in the function...

internal/rbd/rbd_util.go
```diff
@@ -552,6 +554,16 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
 		}
 	}
 
+	rbdVol.ImageID, err = j.GetReservedImageID(ctx, rbdVol.JournalPool, rbdVol.ReservedID, cr)
+	if err != nil {
+		// skip Not found error for older RBD images, as we have not saved
```
nixpanic, line 545: should we insert the image id here? might make it easier for upgrades? file an issue and mark as TODO?

internal/rbd/rbd_util.go
```diff
@@ -773,19 +773,17 @@ func imageFeaturesToUint64(ctx context.Context, imageFeatures string) uint64 {
...
 
-func createSnapshot(ctx context.Context, pOpts *rbdSnapshot, cr *util.Credentials) error {
-	var output []byte
-
-	klog.V(4).Infof(util.Log(ctx, "rbd: snap create %s using mon %s"), pOpts, pOpts.Monitors)
-	args := []string{"snap", "create", pOpts.String(), "--id", cr.ID, "-m", pOpts.Monitors, "--keyfile=" + cr.KeyFile}
-
-	output, err := execCommand("rbd", args)
+func createSnapshot(ctx context.Context, vol *rbdVolume, pOpts *rbdSnapshot, cr *util.Credentials) error {
+	klog.V(4).Infof(util.Log(ctx, "rbd: snap create %s/%s@%s using mon %s"), pOpts.Pool, pOpts.RbdImageName, pOpts.RbdSnapName, pOpts.Monitors)
```
nixpanic: just print `%s` <- `pOpts` here instead of building an image name

internal/rbd/rbd_util.go
```diff
@@ -786,24 +786,18 @@ func createSnapshot(ctx context.Context, vol *rbdVolume, pOpts *rbdSnapshot, cr
...
-
-	if err := undoSnapReservation(ctx, pOpts, cr); err != nil {
-		klog.Errorf(util.Log(ctx, "failed to remove reservation for snapname (%s) with backing snap (%s): %s"),
-			pOpts.RequestName, pOpts, err)
+		return err
 	}
+	defer image.Close()
 
-	return nil
+	snap := image.GetSnapshot(pOpts.RbdSnapName)
```
nixpanic, line 848: please check if `snap` is `nil`

internal/rbd/rbd_util.go
```diff
@@ -786,24 +786,18 @@ func createSnapshot(ctx context.Context, vol *rbdVolume, pOpts *rbdSnapshot, cr
...
 
-func deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot, cr *util.Credentials) error {
-	var output []byte
-
-	klog.V(4).Infof(util.Log(ctx, "rbd: snap rm %s using mon %s"), pOpts, pOpts.Monitors)
-	args := []string{"snap", "rm", pOpts.String(), "--id", cr.ID, "-m", pOpts.Monitors, "--keyfile=" + cr.KeyFile}
-
-	output, err := execCommand("rbd", args)
+func deleteSnapshot(ctx context.Context, vol *rbdVolume, pOpts *rbdSnapshot, cr *util.Credentials) error {
+	klog.V(4).Infof(util.Log(ctx, "rbd: snap rm %s/%s@%s using mon %s"), pOpts.Pool, pOpts.RbdImageName, pOpts.RbdSnapName, pOpts.Monitors)
```
nixpanic, line 840: same here, just use `pOpts`

internal/rbd/rbd_util.go
```diff
@@ -800,18 +803,36 @@ func deleteSnapshot(ctx context.Context, vol *rbdVolume, pOpts *rbdSnapshot, cr
 	return err
 }
 
-// nolint: interfacer
-func restoreSnapshot(ctx context.Context, pVolOpts *rbdVolume, pSnapOpts *rbdSnapshot, cr *util.Credentials) error {
-	var output []byte
+func cloneRbdImageFromSnapshot(ctx context.Context, pVolOpts *rbdVolume, pSnapOpts *rbdSnapshot, cr *util.Credentials) error {
```
nixpanic: This is a nice chance on moving `cloneRbdImageFromSnapshot()` to a method of `rbdSnapshot`. Can be done as a cleanup later.

internal/rbd/rbd_util.go
```diff
@@ -800,18 +803,36 @@ func deleteSnapshot(ctx context.Context, vol *rbdVolume, pOpts *rbdSnapshot, cr
...
+			return errors.Wrapf(err, "failed to set image features")
+		}
+	}
+
+	err := options.SetUint64(imageOptionCloneFormat, 2)
+	if err != nil {
+		return errors.Wrapf(err, "failed to set image features")
+	}
+
+	ioctx, err := pVolOpts.conn.GetIoctx(pVolOpts.Pool)
```
nixpanic: use `pVolOpts.openIoctx()` and `pVolOpts.ioctx` after that

internal/rbd/controllerserver.go
```diff
@@ -161,6 +160,7 @@ func buildCreateVolumeResponse(ctx context.Context, req *csi.CreateVolumeRequest
 }
 
 // CreateVolume creates the volume in backend
+// nolint: gocyclo
```
nixpanic, line 163: add a comment why `gocyclo` should not be checked here

internal/rbd/controllerserver.go
```diff
@@ -294,27 +259,123 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	return &csi.CreateVolumeResponse{Volume: volume}, nil
 }
 
+func (cs *ControllerServer) checkFlatten(ctx context.Context, rbdVol *rbdVolume, cr *util.Credentials) error {
```
nixpanic, line 262: should this not be a method of `rbdVolume` instead? `cs` is not used in this function, only `rbdVolume`.

internal/rbd/controllerserver.go
```diff
@@ -294,27 +259,123 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+			klog.Warningf(util.Log(ctx, "failed undoing reservation of volume: %s (%s)"), rbdVol.RequestName, errDefer)
+		}
+		return status.Error(codes.Internal, err.Error())
+	}
+	return nil
+}
+
+func (cs *ControllerServer) createVolumeFromSnapshot(ctx context.Context, cr *util.Credentials, rbdVol *rbdVolume, snapshotID string) error {
+	rbdSnap := &rbdSnapshot{}
+	if acquired := cs.SnapshotLocks.TryAcquire(rbdSnap.SnapID); !acquired {
```
nixpanic, line 283: `rbdSnap.SnapID` is not set? The object just got created with empty values. Maybe you meant to use `snapshotID`?

internal/rbd/controllerserver.go
```diff
@@ -294,27 +259,123 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+	// update parent name(rbd image name in snapshot)
+	rbdSnap.RbdImageName = rbdSnap.RbdSnapName
+	// create clone image and delete snapshot
+	err = cloneRbdImageFromSnapshot(ctx, rbdVol, rbdSnap)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to clone rbd image %s from snapshot %s: %v"), rbdSnap, err)
+		return err
+	}
+
+	err = flattenRbdImage(ctx, rbdVol, cr)
```
nixpanic, line 307: `flattenRbdImage` should be a method of `rbdVolume`, can be done at a later point too

internal/rbd/controllerserver.go
```diff
@@ -660,33 +734,81 @@ func (cs *ControllerServer) validateSnapshotReq(ctx context.Context, req *csi.Cr
 	return nil
 }
 
-func (cs *ControllerServer) doSnapshot(ctx context.Context, rbdSnap *rbdSnapshot, cr *util.Credentials) (err error) {
-	err = createSnapshot(ctx, rbdSnap, cr)
-	// If snap creation fails, even due to snapname already used, fail, next attempt will get a new
-	// uuid for use as the snap name
+func (cs *ControllerServer) doSnapshot(ctx context.Context, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cr *util.Credentials) (bool, *rbdVolume, error) {
+	// generate cloned volume details from snapshot
+	cloneRbd := generateVolFromSnap(rbdSnap)
```
nixpanic, line 739: can this return `nil`?

internal/rbd/rbd_util.go
```diff
@@ -355,13 +353,12 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 	klog.V(4).Infof(util.Log(ctx, "rbd: delete %s using mon %s, pool %s"), image, pOpts.Monitors, pOpts.Pool)
 
 	var ioctx *rados.IOContext
-	ioctx, err = pOpts.conn.GetIoctx(pOpts.Pool)
+	err = pOpts.openIoctx()
```
nixpanic, line 356: hmm, in a follow-up change you call `openIoctx()`. Still easier to call `open()`

internal/rbd/rbd_util.go
```diff
@@ -480,15 +466,12 @@ func genSnapFromSnapID(ctx context.Context, rbdSnap *rbdSnapshot, snapshotID str
...
-	err = updateSnapWithImageInfo(ctx, rbdSnap, cr)
-
-	return err
+	return nil
 }
 
 // genVolFromVolID generates a rbdVolume structure from the provided identifier, updating
 // the structure with elements from on-disk image metadata as well
-func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials, secrets map[string]string) (*rbdVolume, error) {
+func genVolFromVolID(ctx context.Context, volumeID string, cleanupVol bool, cr *util.Credentials, secrets map[string]string) (*rbdVolume, error) {
```
nixpanic, line 474: I was wondering about the new boolean that was passed to this function. However, `cleanupVol` does not help me at all. Maybe document its intention?

internal/rbd/rbd_util.go
```diff
@@ -568,10 +551,71 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
...
+				return nil, ErrImageNotFound{rbdVol.RbdImageName, errors.New("image not present in backend")}
+			}
+		}
+		return nil, err
+	}
 
 	return rbdVol, err
 }
 
+func removeImageFromTrash(rbdVol *rbdVolume) error {
```
nixpanic, line 572: just make this a method of `rbdVolume`

internal/rbd/rbd_util.go
```diff
@@ -568,10 +551,71 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
...
+	ioctx, err := rbdVol.conn.GetIoctx(rbdVol.Pool)
```
nixpanic, line 573: call `rbdVol.openIoctx()` and use `rbdVol.ioctx` (no need to destroy the ioctx)

internal/rbd/rbd_util.go
```diff
@@ -568,10 +551,71 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
...
+	}
+
+	for _, snap := range snaps {
+		if snap.Name == rbdSnap.RbdSnapName {
+			return snap, nil
+		}
+	}
+
+	return info, ErrSnapNotFound{rbdSnap.RbdSnapName, fmt.Errorf("snap (%s) for image (%s) not found",
+		rbdSnap.RbdSnapName, rbdSnap.Pool+"/"+rbdSnap.RbdImageName)}
```
nixpanic, line 616: use `rbdSnap.String()` from #1117 (or have it automatically done)

internal/rbd/rbd_util.go
```diff
@@ -902,58 +933,99 @@ func (rv *rbdVolume) getImageInfo(ctx context.Context, monitors string, cr *util
 	return imgInfo, nil
 }
 
-// snapInfo strongly typed JSON spec for snap ls rbd output
-type snapInfo struct {
-	ID        int64  `json:"id"`
-	Name      string `json:"name"`
-	Size      int64  `json:"size"`
-	Timestamp string `json:"timestamp"`
+func getCloneDepth(ctx context.Context, monitors, poolName, imageName string, depth int, cr *util.Credentials) (int, error) {
```
nixpanic, line 936: can this be a method of rbdVolume instead? you're passing all important members of the type, and are constructing a new object from those

no need to pass `depth`?

internal/rbd/rbd_util.go
```diff
@@ -902,58 +933,99 @@ func (rv *rbdVolume) getImageInfo(ctx context.Context, monitors string, cr *util
...
 
-	stdout, stderr, err := util.ExecCommand(
-		"rbd",
-		"-m", monitors,
-		"--id", cr.ID,
-		"--keyfile="+cr.KeyFile,
-		"-c", util.CephConfigPath,
-		"--format="+"json",
-		"snap", "ls", rs.imageSpec())
+	depth, err = getCloneDepth(ctx, rbdVol.Monitors, rbdVol.Pool, rbdVol.RbdImageName, 0, cr)
```
nixpanic, line 963: why not use `rbdVol` directly in `getCloneDepth()`?

internal/rbd/rbd_util.go
```diff
@@ -902,58 +933,99 @@ func (rv *rbdVolume) getImageInfo(ctx context.Context, monitors string, cr *util
...
-		klog.Errorf(util.Log(ctx, "failed getting snap (%s) information from image (%s): (%s)"),
-			rs.RbdSnapName, rs.imageSpec(), err)
-		if strings.Contains(string(stderr), "rbd: error opening image "+rs.RbdImageName+
-			": (2) No such file or directory") {
-			return snpInfo, ErrImageNotFound{rs.imageSpec(), err}
+		return err
+	}
+	klog.Infof(util.Log(ctx, "clone depth is (%d), configured softlimit (%v) and hardlimit (%v) for %s"), depth, rbdSoftMaxCloneDepth, rbdHardMaxCloneDepth, rbdVol)
+
+	if (depth != 0 && flattenImage) || (depth >= int(rbdHardMaxCloneDepth) || depth >= int(rbdSoftMaxCloneDepth)) {
```
nixpanic, line 969: conversion of `rbdHardMaxCloneDepth` seems clutter, can these configuration options not be an int?

internal/rbd/rbd_util.go
```diff
@@ -902,58 +933,99 @@ func (rv *rbdVolume) getImageInfo(ctx context.Context, monitors string, cr *util
...
+			}
 		}
-		return snpInfo, err
 	}
+	return nil
+}
+
+// rbdManagerTaskAddFlatten adds a ceph manager task to flatten image
+// asynchronously in the background. If command is not found returns a bool set to false
+func rbdManagerTaskAddFlatten(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) (bool, error) {
```
nixpanic, line 1004: probably should be a method of `rbdVolume`

internal/rbd/rbd_util.go
```diff
@@ -902,58 +933,99 @@ func (rv *rbdVolume) getImageInfo(ctx context.Context, monitors string, cr *util
...
-			rs.RbdSnapName, rs.imageSpec(), err)
-		if strings.Contains(string(stderr), "rbd: error opening image "+rs.RbdImageName+
-			": (2) No such file or directory") {
-			return snpInfo, ErrImageNotFound{rs.imageSpec(), err}
+		return err
+	}
+	klog.Infof(util.Log(ctx, "clone depth is (%d), configured softlimit (%v) and hardlimit (%v) for %s"), depth, rbdSoftMaxCloneDepth, rbdHardMaxCloneDepth, rbdVol)
+
+	if (depth != 0 && flattenImage) || (depth >= int(rbdHardMaxCloneDepth) || depth >= int(rbdSoftMaxCloneDepth)) {
+		ns, err := rbdManagerTaskAddFlatten(ctx, rbdVol, cr)
```
nixpanic, line 970: I guess `ns` stands for `NotSupported`? Better to return an error type that matches that instead of a bool

internal/rbd/rbd_util.go
```diff
@@ -902,58 +933,99 @@ func (rv *rbdVolume) getImageInfo(ctx context.Context, monitors string, cr *util
...
+			if flattenImage || depth >= int(rbdHardMaxCloneDepth) {
+				err = rbdVol.Connect(cr)
+				if err != nil {
+					return err
+				}
+				err := rbdVol.openIoctx()
+				if err != nil {
+					return err
+				}
+				rbdImage := librbd.GetImage(rbdVol.ioctx, rbdVol.RbdImageName)
```
nixpanic, line 991: use `rbdVol.open()` to get the image

internal/rbd/snapshot.go
```diff
@@ -0,0 +1,124 @@
...
+
+func createRBDClone(ctx context.Context, parentVol, cloneRbdVol *rbdVolume, snap *rbdSnapshot, cr *util.Credentials) error {
+	// create snapshot
+	err := createSnapshot(ctx, parentVol, snap)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to create snapshot %s: %v"), snap, err)
+		return err
+	}
+
+	cloneFailed := false
```
nixpanic, line 36: default to `true`, always assume it fails so that the cleanup path is preferred

internal/rbd/snapshot.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to clone rbd image %s from snapshot %s: %v"), cloneRbdVol.RbdImageName, snap.RbdSnapName, err)
+		err = errors.Errorf("failed to clone rbd image %s from snapshot %s: %v", cloneRbdVol.RbdImageName, snap.RbdSnapName, err)
+	}
+	errSnap := deleteSnapshot(ctx, parentVol, snap)
+	if errSnap != nil {
+		klog.Errorf(util.Log(ctx, "failed to delete snapshot: %v"), errSnap)
+	}
+
+	if cloneFailed {
```
nixpanic, line 48: not getting cleaned up now, `cloneFailed` is not changed  from `false`

internal/rbd/controllerserver.go
```diff
@@ -191,6 +190,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		klog.Errorf(util.Log(ctx, "failed to connect to volume %v: %v"), rbdVol.RbdImageName, err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	defer rbdVol.Destroy()
```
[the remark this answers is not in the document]
Madhu-1: no, I dint get what is the issue. Destroy is called right after the Connect.

internal/rbd/rbd_util.go
```diff
@@ -299,17 +300,12 @@ func rbdStatus(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) (boo
...
 
-	args := []string{"rbd", "task", "add", "remove",
-		pOpts.String(),
-		"--id", cr.ID,
-		"--keyfile=" + cr.KeyFile,
-		"-m", pOpts.Monitors,
-	}
+	klog.V(4).Infof(util.Log(ctx, "executing %v for image (%s) using mon %s, pool %s"), args, pOpts.RbdImageName, pOpts.Monitors, pOpts.Pool)
 
 	output, err := execCommand("ceph", args)
```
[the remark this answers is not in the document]
Madhu-1: task manager commands can only be added with "ceph"

internal/rbd/rbd_util.go
Madhu-1: would like to avoid opening an image as its not required it.

internal/rbd/controllerserver.go
Madhu-1: This is required for locking

internal/rbd/rbd_util.go
Madhu-1: it's already handled in deleteVolume, if required later we can move it here also

internal/rbd/rbd_util.go
Madhu-1: no idea in which case it can return `nil` as per https://github.com/ceph/go-ceph/blob/v0.2.0/rbd/rbd.go#L958. adding nil check for safer side.

internal/rbd/rbd_util.go
Madhu-1: sorry already done, not pushed the latest change :(

internal/rbd/controllerserver.go
Madhu-1: earlier i stored the hard and soft depth flags in cs. after that i removed. fixed now

internal/rbd/controllerserver.go
Madhu-1: this should be fine, this function will be called after validating the snapshotID. this cannot be unset.

internal/rbd/controllerserver.go
Madhu-1: not possible. it will generate the volume struct from snapshot struct(snap struct wont be empty)

internal/rbd/rbd_util.go
Madhu-1: This is not added as part of this PR, if the command is not supported we will kind of fallback.

internal/rbd/snapshot.go
Madhu-1: removed this variable for cleaner approach

internal/rbd/snapshot.go
Madhu-1: removed this variable for cleaner approach

Madhu-1:
@nixpanic addressed review comments. PTAL. as i need to do follow up PR's to address TODO if any minor changes are required i can do it. 

@dillaman @ShyamsundarR @humblec can you please review this one?

internal/rbd/controllerserver.go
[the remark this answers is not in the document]
nixpanic: yes, but it was called after `parseVolCreateRequest()`, why change that?
See https://github.com/ceph/ceph-csi/pull/1160/commits/33238dccf0831e96374e174840cca60c6c5d194e#diff-f010f466063f62f823361f972d08b554L176-L193

internal/rbd/controllerserver.go
```diff
@@ -345,6 +344,13 @@ func (cs *ControllerServer) DeleteLegacyVolume(ctx context.Context, req *csi.Del
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	err := rbdVol.Connect(cr)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	defer rbdVol.Destroy()
```
nixpanic: Same here, the defer for `Destroy()` can be done early

internal/rbd/rbd_util.go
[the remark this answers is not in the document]
nixpanic: is it not **always** `ceph rbd task add ...`, having that as a first constant for args would be better

internal/rbd/rbd_util.go
```diff
@@ -335,38 +331,62 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
 		return err
 	}
 	if found {
-		klog.Errorf(util.Log(ctx, "rbd %s is still being used"), pOpts)
-		return fmt.Errorf("rbd %s is still being used", pOpts)
+		klog.Errorf(util.Log(ctx, "rbd is still being used "), image)
+		return fmt.Errorf("rbd %s is still being used", image)
 	}
+	// Support deleting the older rbd images whose imageID is not stored in omap
+	if pOpts.ImageID == "" {
```
nixpanic, line 348: the `pOpts.ImageID == ""` check should be inside `pOpts.getImageID()`, so that you can always call the function.

internal/rbd/rbd_util.go
nixpanic: ah, right... Calling librbd from outside a `rbdVolume` method is not very nice, but there is currently no cleaner alternative

internal/rbd/controllerserver.go
nixpanic: but it is not used in this function? or maybe a follow up commit does that? It looks unneeded in https://github.com/ceph/ceph-csi/pull/1160/commits/b5209f8f74bde7b2a84f287c9b5c58e09e02a751

internal/rbd/rbd_util.go
nixpanic: no idea if it is needed, maybe others can recommend (against) it?

internal/rbd/controllerserver.go
nixpanic: No, I mean you create it on line 283 with `rbdSnap := &rbdSnapshot{}` and then use `rbdSnap.SnapID` on line 284. `rbdSnap.SnapID` will always be an empty string?

internal/rbd/controllerserver.go
```diff
@@ -758,11 +863,61 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
...
+	// Deleting snapshot and cloned volume
+	klog.V(4).Infof(util.Log(ctx, "deleting cloned rbd volume %s"), rbdSnap.RbdSnapName)
+
+	rbdVol := generateVolFromSnap(rbdSnap)
+
+	err = rbdVol.Connect(cr)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	defer rbdVol.Destroy()
```
nixpanic, line 875: ```defer rbdVol.Destroy()``` can be called immediately after `generateVolFromSnap()`

internal/rbd/rbd_util.go
```diff
@@ -481,15 +467,14 @@ func genSnapFromSnapID(ctx context.Context, rbdSnap *rbdSnapshot, snapshotID str
...
-	err = updateSnapWithImageInfo(ctx, rbdSnap, cr)
-
-	return err
+	return nil
 }
 
 // genVolFromVolID generates a rbdVolume structure from the provided identifier, updating
 // the structure with elements from on-disk image metadata as well
-func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials, secrets map[string]string) (*rbdVolume, error) {
+// cleanupVol will be set during the DeleteVolume operation to delete the image,
```
nixpanic, line 475: this adds more complexity to the function. is it not easier to understand if the DeleteVolume operation takes care of it? A better maintainable solution has my preference.

internal/rbd/rbd_util.go
```diff
@@ -768,16 +812,19 @@ func imageFeaturesToUint64(ctx context.Context, imageFeatures string) uint64 {
 	features := uint64(0)
 
 	for _, f := range strings.Split(imageFeatures, ",") {
```
nixpanic, line 814: You can use `librbd.FeatureSetFromNames(strings.Split(imageFeatures, ",")` and get the int value directly. Use `Names()` to get a `[]string` with the features that are recognized and see if it misses any of the original requested features.

Also, the strings for features are available like `FeatureNameDeepFlatten`.

Doing it this way, even features that are not known to ceph-csi can get enabled (by advanced users).

internal/rbd/rbd_util.go
nixpanic: this really should return an error `ErrCmdNotFound` or similar instead of an additional `bool`.

internal/rbd/rbd_util.go
Madhu-1: no other rbdvolume related information is passed from genVolFromVolID to delete the image, in case of failure we return nil and error

internal/rbd/rbd_util.go
Madhu-1: maybe in followup PR, as I just added one more case  to it.

Madhu-1:
@nixpanic GitHub is not allowing me to reply to a few comments, let me know if we need to discuss any particular comment.

nixpanic:
> @nixpanic GitHub is not allowing me to reply to a few comments, let me know if we need to discuss any particular comment.

You might be able to reply if you click the `Commits` tab, the first commit, and then `Next >` until you find a comment you want to respond to? That is also how I left the comments.

internal/rbd/controllerserver.go
```diff
@@ -262,27 +257,121 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+		return err
+	}
+	klog.V(4).Infof(util.Log(ctx, "create volume %s from snapshot %s"), rbdVol.RequestName, rbdSnap.RbdSnapName)
+	return nil
+}
+
+func (cs *ControllerServer) createBackingImage(ctx context.Context, cr *util.Credentials, rbdVol *rbdVolume, rbdSnap *rbdSnapshot) error {
 	var err error
 
+	var j = new(journal.Connection)
```
phlogistonjohn, line 316: Why new? It's exceedingly rare to see a call to new used in most go code IMO.

Madhu-1: >internal/rbd/controllerserver.go:783:2: SA4006: this value of `j` is never used (staticcheck)
	j := &journal.Connection{}

the short variable declaration was giving a static check error for this now, now change back to address of struct.

ShyamsundarR:
>     * we need to have some flag to indicate that all RBD images for PVCs need to be flattened upon creation. This flag would be set to `true` if they have at least one compute / worker node with an old OS that doesn't support the deep-flatten feature and the operations feature that is set when you use clone v2 (setting of the flag can be an exercise for the user / Rook).

From a standalone csph-csi perspective, the NodeStageVolume call can possibly detect the local OS version and also the image depth that is required to be mapped, and decide to allow/disallow the operation.

dillaman:
> > ```
> > * we need to have some flag to indicate that all RBD images for PVCs need to be flattened upon creation. This flag would be set to `true` if they have at least one compute / worker node with an old OS that doesn't support the deep-flatten feature and the operations feature that is set when you use clone v2 (setting of the flag can be an exercise for the user / Rook).
> > ```
> 
> From a standalone csph-csi perspective, the NodeStageVolume call can possibly detect the local OS version and also the image depth that is required to be mapped, and decide to allow/disallow the operation.

If we just wanted to randomly fail at that step, you could just let it attempt to `rbd block map` and that will fail as well.

Madhu-1:
> > ```
> > * we need to have some flag to indicate that all RBD images for PVCs need to be flattened upon creation. This flag would be set to `true` if they have at least one compute / worker node with an old OS that doesn't support the deep-flatten feature and the operations feature that is set when you use clone v2 (setting of the flag can be an exercise for the user / Rook).
> > ```
> 
> From a standalone csph-csi perspective, the NodeStageVolume call can possibly detect the local OS version and also the image depth that is required to be mapped, and decide to allow/disallow the operation.

This could give a bad user experience- Any cloned or restored PVC cannot be mapped on the nodes where the kernel version is <8.2+(the user has the manually reschedule the app to the nodes where the kernel is 8.2+ with some affinity). instead of this, we can flatten the image. The only issue with this is Creation operation can be slow. eventually when users move to new kernel versions we can disable the flag by default

charts/ceph-csi-rbd/values.yaml
```diff
@@ -107,6 +107,15 @@ provisioner:
...
   timeout: 60s
+  # Hard limit for maximum number of nested volume clones that are taken before
+  # a flatten occurs
+  hardMaxCloneDepth: 8
+  # Soft limit for maximum number of nested volume clones that are taken before
+  # a flatten occurs
+  softMaxCloneDepth: 4
+  # Flatten all cloned rbd images upon creation, if kernel version doesnot
+  # support deep-flatten feature
+  flattenOnCreate: true
```
humblec, line 118: may `enabledeepflatten` would be a better naming I believe.

Madhu-1: The name was suggested by @dillaman in one of the discussion, This seems correct as we need to flatten the image upon creation, I don't have hard feelings I can change it if required.

humblec: No, thats fine. lets keep it as `flattenOnCreate`

dillaman: Alternative suggestion could be `enableThinProvisionedClones` (or similar). I don't think anything referencing the RBD `deep-flatten` feature bit would mean much to end-users.

ShyamsundarR:
> > > ```
> > > * we need to have some flag to indicate that all RBD images for PVCs need to be flattened upon creation. This flag would be set to `true` if they have at least one compute / worker node with an old OS that doesn't support the deep-flatten feature and the operations feature that is set when you use clone v2 (setting of the flag can be an exercise for the user / Rook).
> > > ```
> > 
> > 
> > From a standalone csph-csi perspective, the NodeStageVolume call can possibly detect the local OS version and also the image depth that is required to be mapped, and decide to allow/disallow the operation.
> 
> This could give a bad user experience- Any cloned or restored PVC cannot be mapped on the nodes where the kernel version is <8.2+(the user has the manually reschedule the app to the nodes where the kernel is 8.2+ with some affinity). instead of this, we can flatten the image. The only issue with this is Creation operation can be slow. eventually when users move to new kernel versions we can disable the flag by default

I am not suggesting we do away with a flag, I am suggesting that we also add a check on the nodeplugin to detect and prevent the Staging.

charts/ceph-csi-rbd/values.yaml
humblec: May be what we do internally - `flatten at create` or `do it in a different queue` also may not add value to the user. in that sense something like `enableThinProvisionedClone` looks better to me.

internal/rbd/rbd_util.go
```diff
@@ -707,115 +842,76 @@ func imageFeaturesToUint64(ctx context.Context, imageFeatures string) uint64 {
...
-	args := []string{"clone", pSnapOpts.String(), pVolOpts.String(),
-		"--id", cr.ID, "-m", pVolOpts.Monitors, "--keyfile=" + cr.KeyFile}
+	klog.V(4).Infof(util.Log(ctx, "rbd: clone %s %s using mon %s"), pSnapOpts, image, pVolOpts.Monitors)
 
-	output, err := execCommand("rbd", args)
+	options := librbd.NewRbdImageOptions()
 
-	if err != nil {
-		return errors.Wrapf(err, "failed to restore snapshot, command output: %s", string(output))
+	if pVolOpts.ImageFeatures != "" {
```
nixpanic, line 894: Note, `ImageFeatures` was removed, use `imageFeatureSet` instead.

internal/rbd/rbd_util.go
```diff
@@ -843,62 +939,79 @@ func (rv *rbdVolume) getImageInfo() error {
...
+	poolName := rv.Pool
+	for {
+		if imageName == "" {
+			return depth, nil
+		}
+		vol := rbdVolume{
+			Pool:         poolName,
+			RbdImageName: imageName,
+		}
+		err := vol.getImageInfo()
```
nixpanic, line 959: This most likely connects the rbdVolume, so it needs to be destroyed afterwards. Easiest is to always add a `defer vol.Destroy()` immediately after creating the rbdVolume object.

internal/rbd/rbd_util.go
```diff
@@ -843,62 +939,79 @@ func (rv *rbdVolume) getImageInfo() error {
...
+			if err != nil {
+				klog.Errorf(util.Log(ctx, "failed to add task flatten for %s : %v"), rv, err)
+				return err
+			}
+			if flattenOnCreate || depth >= rbdHardMaxCloneDepth {
+				return ErrFlattenInProgress{err: fmt.Errorf("flatten is in progress for image %s", rv.RbdImageName)}
+			}
+		}
+		if !ns {
+			klog.Errorf(util.Log(ctx, "rbd task add flatten not supported will flatten image once hardlimit is reached: %v"), err)
```
nixpanic, line 996: maybe quote the command, or describe it like "task manager does not support flatten, image will be .."

internal/rbd/snapshot.go
```diff
@@ -0,0 +1,122 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package rbd
+
+import (
+	"context"
+	"strings"
+
+	"github.com/ceph/ceph-csi/internal/util"
```
nixpanic, line 22: I think it is recommended to place imports of the local project after all others

internal/rbd/snapshot.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/pkg/errors"
+	"k8s.io/klog"
+)
+
+func createRBDClone(ctx context.Context, parentVol, cloneRbdVol *rbdVolume, snap *rbdSnapshot, cr *util.Credentials) error {
```
nixpanic, line 28: Please consider making all/most of the functions in this file methods of the rbdSnapshot type

internal/rbd/rbd_util.go
Madhu-1: will make changes to reuse the ioctx and conn of parent volume, no need to destroy it, as parent will destroy it.

internal/rbd/snapshot.go
Madhu-1: Nope i sent a PR to change order as per the standards.

internal/rbd/snapshot.go
Madhu-1: will take care of it when doing fixing the TODO.

internal/rbd/rbd_util.go
```diff
@@ -707,115 +842,76 @@ func imageFeaturesToUint64(ctx context.Context, imageFeatures string) uint64 {
 	features := uint64(0)
 
 	for _, f := range strings.Split(imageFeatures, ",") {
-		if f == "layering" {
+		switch f {
```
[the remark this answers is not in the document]
Madhu-1: waiting for #1178 to be merged

internal/rbd/snapshot.go
nixpanic: oh, ok, my mistake

internal/rbd/rbd_util.go
nixpanic: try not to over complicate things, if the connection exists. Optimizations like this can be done later when ioctx's are in a shared pool.

Madhu-1: we cannot use getImageInfo as it doesn't return the parent Image name, filed an issue https://github.com/ceph/go-ceph/issues/312. in followup PR i will change back to getImageInfo.

internal/rbd/rbd_util.go
```diff
@@ -821,62 +921,79 @@ func (rv *rbdVolume) getImageInfo() error {
...
+		RbdImageName: rv.RbdImageName,
+		conn:         rv.conn,
+	}
+	for {
+		if vol.RbdImageName == "" {
+			return depth, nil
+		}
+		err := vol.updateVolWithImageInfo(ctx, rv.conn.Creds)
+		if err != nil {
+			klog.Errorf(util.Log(ctx, "failed to check depth on image:%s %s"), vol, err)
```
nixpanic, line 943: nit: misplaced colon: `failed to ... depth on %s: %s`

internal/rbd/rbd_util.go
```diff
@@ -821,62 +921,79 @@ func (rv *rbdVolume) getImageInfo() error {
...
-		return snpInfo, err
-	}
-
-	err = json.Unmarshal(stdout, &snaps)
-	if err != nil {
-		klog.Errorf(util.Log(ctx, "failed to parse JSON output of image snap list (%s): (%s)"), rs.imageSpec(), err)
-		return snpInfo, fmt.Errorf("unmarshal failed: %+v. raw buffer response: %s", err, string(stdout))
+		return err
 	}
+	klog.Infof(util.Log(ctx, "clone depth is (%d), configured softlimit (%v) and hardlimit (%v) for %s"), depth, rbdSoftMaxCloneDepth, rbdHardMaxCloneDepth, rv)
```
nixpanic, line 963: nit: mixing `%d` and `%v` in the formatting string?

internal/rbd/controllerserver.go
```diff
@@ -622,49 +731,81 @@ func (cs *ControllerServer) validateSnapshotReq(ctx context.Context, req *csi.Cr
 	return nil
 }
 
-func (cs *ControllerServer) doSnapshot(ctx context.Context, rbdSnap *rbdSnapshot, cr *util.Credentials) (err error) {
-	err = createSnapshot(ctx, rbdSnap, cr)
-	// If snap creation fails, even due to snapname already used, fail, next attempt will get a new
-	// uuid for use as the snap name
+func (cs *ControllerServer) doSnapshot(ctx context.Context, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cr *util.Credentials) (bool, *rbdVolume, error) {
```
chenxu1990, line 734: Is it better to call 'doSnapClone' ?

internal/rbd/rbd_journal.go
```diff
@@ -137,26 +140,42 @@ func checkSnapExists(ctx context.Context, rbdSnap *rbdSnapshot, cr *util.Credent
...
 
-	klog.V(4).Infof(util.Log(ctx, "found existing snap (%s) with snap name (%s) for request (%s)"),
+	klog.V(4).Infof(util.Log(ctx, "found existing image (%s) with name (%s) for request (%s)"),
 		rbdSnap.SnapID, rbdSnap.RbdSnapName, rbdSnap.RequestName)
+	// Madhu TODO check flatten in progress
+	err = vol.Connect(cr)
+	if err != nil {
+		return false, err
+	}
+	vol.Destroy()
```
chenxu1990, line 170: defer ?

internal/rbd/snapshot.go
```diff
@@ -0,0 +1,122 @@
...
+	// create clone image and delete snapshot
+	err = cloneRbdImageFromSnapshot(ctx, cloneRbdVol, snap)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to clone rbd image %s from snapshot %s: %v"), cloneRbdVol.RbdImageName, snap.RbdSnapName, err)
+		err = errors.Errorf("failed to clone rbd image %s from snapshot %s: %v", cloneRbdVol.RbdImageName, snap.RbdSnapName, err)
+	}
+	errSnap := deleteSnapshot(ctx, parentVol, snap)
+	if errSnap != nil {
+		klog.Errorf(util.Log(ctx, "failed to delete snapshot: %v"), errSnap)
+		delErr := deleteImage(ctx, cloneRbdVol, cr)
```
ShyamsundarR, line 45: If there was an `err` in line 37 above, then deleteImage need not be invoked.

Possibly cleaner to defer `deleteSnapshot` post `createSnapshot` and let the rest of the code play out. Similarly wrap an error routine for defering `deleteImage` (latter is just a thought, former seems more useful)

internal/rbd/controllerserver.go
```diff
@@ -622,49 +731,81 @@ func (cs *ControllerServer) validateSnapshotReq(ctx context.Context, req *csi.Cr
...
+	var j = &journal.Connection{}
+	// save image ID
+	j, err = snapJournal.Connect(rbdSnap.Monitors, cr)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to connect to cluster: %v"), err)
+		return ready, cloneRbd, err
+	}
+	defer j.Destroy()
+
+	err = j.ReserveImageID(ctx, rbdSnap.JournalPool, rbdSnap.ReservedID, cloneRbd.ImageID, cr)
```
ShyamsundarR, line 794: If we fail just prior to this call (as in provisioner was restarted, say), we would fail to save the image ID. The snapshot although is complete, so on a `CreateSnapshot` retry we would succeed in `checkSnapCloneExists` and not save the image ID for `DeleteSnapshot` call.

Either `checkSnapCloneExists` should ensure the image ID is saved, or undo the entire work, or the image ID should be saved sooner in the transaction to prevent such cases.

internal/rbd/controllerserver.go
```diff
@@ -622,49 +731,81 @@ func (cs *ControllerServer) validateSnapshotReq(ctx context.Context, req *csi.Cr
...
+				if errCleanUp != nil {
+					klog.Errorf(util.Log(ctx, "failed to cleanup snapshot and clone: %v"), errCleanUp)
+				}
 			}
-			err = status.Error(codes.Internal, err.Error())
 		}
 	}()
-	err = protectSnapshot(ctx, rbdSnap, cr)
+
+	err = createSnapshot(ctx, cloneRbd, rbdSnap)
```
ShyamsundarR, line 766: I am slightly confused by this additional snapshot creation here. Is this the single RBD snapshot from which we will create future clones? As commented [here](https://github.com/ceph/ceph-csi/pull/1160#issuecomment-642069930)?

If so, we again do not seem to be checking if this snapshot is created when we call `checkSnapCloneExists`, as a result we would fail in the future to use this snapshot once it is marked ready?

internal/rbd/rbd_journal.go
```diff
@@ -137,26 +140,42 @@ func checkSnapExists(ctx context.Context, rbdSnap *rbdSnapshot, cr *util.Credent
...
+
+	err = vol.flattenRbdImage(ctx, cr)
```
ShyamsundarR, line 172: (nit) Flatten does not seem to belong to this function as such, which checks if the journal and hence the image is intact. I would see if we can move the flatten check outside of this code to the caller.

internal/rbd/controllerserver.go
```diff
@@ -262,27 +258,121 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+	}()
+	err = rbdVol.getImageID()
 	if err != nil {
-		klog.Errorf(util.Log(ctx, "failed to create volume: %v"), err)
+		klog.Errorf(util.Log(ctx, "failed to get volume id %s: %v"), rbdVol, err)
 		return status.Error(codes.Internal, err.Error())
 	}
 
-	klog.V(4).Infof(util.Log(ctx, "created volume %s backed by image %s"), rbdVol.RequestName, rbdVol.RbdImageName)
+	err = j.ReserveImageID(ctx, rbdVol.JournalPool, rbdVol.ReservedID, rbdVol.ImageID, cr)
```
ShyamsundarR, line 355: We may need to move this further up, like for snapshots, and ensure when we find an existing image, this data is already stored.

internal/journal/voljournal.go
```diff
@@ -649,6 +655,24 @@ func (conn *Connection) GetImageAttributes(ctx context.Context, pool, objectUUID
...
+func (conn *Connection) ReserveImageID(ctx context.Context, pool, reservedUUID, imageID string, cr *util.Credentials) error {
+	err := util.SetOMapKeyValue(ctx, conn.monitors, cr, pool, conn.config.namespace, conn.config.cephUUIDDirectoryPrefix+reservedUUID, conn.config.csiImageIDKey, imageID)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetReservedImageID retrives the reserved image ID from the omap
+func (conn *Connection) GetReservedImageID(ctx context.Context, pool, reservedUUID string, cr *util.Credentials) (string, error) {
```
ShyamsundarR, line 668: Ideally we should be able to get this from `GetImageAttributes` (maybe once we have the fetch all keys in a single call).

internal/rbd/snapshot.go
```diff
@@ -0,0 +1,122 @@
...
+	vol.ClusterID = rbdSnap.ClusterID
+	vol.VolID = rbdSnap.SnapID
+	vol.Monitors = rbdSnap.Monitors
+	vol.Pool = rbdSnap.Pool
+	vol.RbdImageName = rbdSnap.RbdSnapName
+	return vol
+}
+
+func undoSnapshotCloning(ctx context.Context, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cloneVol *rbdVolume, cr *util.Credentials, deleteVol bool) error {
+	err := deleteSnapshot(ctx, parentVol, rbdSnap)
```
ShyamsundarR, line 106: This is the snapshot of the cloned image, that would be used for future CSI-cloning, right? Attempting to get my bearing straight on this one :)

Madhu-1:
As per the latest discussion to handle mounting of rbd images or clone images in the chain which has the deep-flatten feature on the nodes who is having lower kernel version will flatten the image before use,
There are two problems we need to address

* We don't have a specific error message to return in NodeStageRequest(intermediate error) currently we are return InternalError to retry with backoff( maybe we need to file an issue in CSI spec to extend node stage error message).as the kubelet retry's when an internal server error occurs we are good here.
* The secret we are using for node stage in the rook doesn't have the permission to talk to the manager to issue rbd flatten images. we need to update rook to extend user caps

Madhu-1:
@ShyamsundarR PTAL.

Madhu-1:
some cleanup required in commits will do it once changes look good.

cmd/cephcsi.go
```diff
@@ -77,6 +77,12 @@ func init() {
 	flag.StringVar(&conf.HistogramOption, "histogramoption", "0.5,2,6",
 		"[DEPRECATED] Histogram option for grpc metrics, should be comma separated value, ex:= 0.5,2,6 where start=0.5 factor=2, count=6")
 
+	flag.UintVar(&conf.RbdHardMaxCloneDepth, "rbdhardmaxclonedepth", 8, "Hard limit for maximum number of nested volume clones that are taken before a flatten occurs")
+	flag.UintVar(&conf.RbdSoftMaxCloneDepth, "rbdsoftmaxclonedepth", 4, "Soft limit for maximum number of nested volume clones that are taken before a flatten occurs")
+
+	flag.BoolVar(&conf.ForceFlattenImage, "forceflatten", true,
```
Madhu-1, line 83: Need to flip this value and change the logic to flatten in nodestage

Madhu-1:
@ShyamsundarR @nixpanic @humblec PTAL

internal/rbd/nodeserver.go
```diff
@@ -60,6 +61,22 @@ type stageTransaction struct {
 	devicePath string
 }
 
+var (
+	release = ""
```
nixpanic, line 65: call this `kernelRelease` or `kernelVersion` please. Just `release` is confusing (was it attached?).

internal/util/util.go
```diff
@@ -94,9 +95,18 @@ type Config struct {
 	IsNodeServer       bool // if set to true start node server
 	Version            bool // cephcsi version
 
+	// SkipForceFlatten is set to false if the kernel supports mounting of
+	// rbd image/image chain has the deep-flatten feature.
```
nixpanic, line 99: instead of `image/image`, maybe write `image or the image` to be clearer

internal/util/util.go
```diff
@@ -151,6 +161,90 @@ func KernelVersion() (string, error) {
 	return strings.TrimRight(string(utsname.Release[:]), "\x00"), nil
 }
 
+// KernelVersion holds kernel related informations
```
nixpanic, line 164: This could be it's own PR that can get reviewed+merged quicker

Madhu-1: @nixpanic as no other PR is dependent on this one, as it's in a separate commit lets get it done here only.

nixpanic: sure, thats fine

internal/util/util.go
```diff
@@ -143,7 +143,7 @@ func ValidateDriverName(driverName string) error {
 
 // KernelVersion returns the version of the running Unix (like) system from the
```
nixpanic, line 153: Need to update the doc with the new function name too

internal/rbd/rbd_util.go
```diff
@@ -306,74 +308,109 @@ func rbdStatus(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) (boo
 	return false, output, nil
 }
 
-// rbdManagerTaskDelete adds a ceph manager task to delete an rbd image, thus deleting
-// it asynchronously. If command is not found returns a bool set to false
-func rbdManagerTaskDeleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) (bool, error) {
+// addRbdManagerTask adds a ceph manager task to execute command
+// asynchronously. If command is not found returns a bool set to false
+// example arg ["trash", "remove","pool/image"]
+func addRbdManagerTask(ctx context.Context, pOpts *rbdVolume, arg []string) (bool, error) {
```
nixpanic, line 315: I would really like to see a single `error` returned, not `(bool, error)`. It can be `ErrNotSupported` instead of the potential `false`.

internal/rbd/rbd_util.go
```diff
@@ -175,11 +175,6 @@ func (rs *rbdSnapshot) String() string {
 	return fmt.Sprintf("%s/%s@%s", rs.Pool, rs.RbdImageName, rs.RbdSnapName)
 }
 
-// imageSpec returns the image-spec (pool/image) format of the snapshot
```
nixpanic, line 175: Why does commit 62aadd543269f07ab8d934274fc2d62b64259eb3 remove `imageSpec()`? It was not used in the rest of the change of the commit.

internal/rbd/rbd_util.go
```diff
@@ -757,18 +757,33 @@ func (rv *rbdVolume) deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot) err
...
-	var output []byte
+func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *rbdSnapshot) error {
+	image := rv.RbdImageName
+	var err error
+	klog.V(4).Infof(util.Log(ctx, "rbd: clone %s %s using mon %s"), pSnapOpts, image, rv.Monitors)
 
-	klog.V(4).Infof(util.Log(ctx, "rbd: clone %s using mon %s"), pVolOpts, pVolOpts.Monitors)
-	args := []string{"clone", pSnapOpts.String(), pVolOpts.String(),
-		"--id", cr.ID, "-m", pVolOpts.Monitors, "--keyfile=" + cr.KeyFile}
+	options := librbd.NewRbdImageOptions()
```
nixpanic, line 900: please add `defer options.Destroy()` to prevent memory leaks

internal/rbd/controllerserver.go
```diff
@@ -192,53 +192,49 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
 	found, err := rbdVol.Exists(ctx)
 	if err != nil {
 		if _, ok := err.(ErrVolNameConflict); ok {
 			return nil, status.Error(codes.AlreadyExists, err.Error())
 		}
-
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if found {
+		if rbdSnap != nil {
```
nixpanic, line 208: Add some comments why `checkFlatten()` is done? I assume the function may also trigger flattening in case it is needed. Writing it explicitly is very helpful.

internal/rbd/controllerserver.go
```diff
@@ -192,53 +192,49 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
-		if err != nil {
-			klog.Errorf(util.Log(ctx, "failed to save encryption status, deleting image %s: %s"),
-				rbdVol, err)
-			if deleteErr := deleteImage(ctx, rbdVol, cr); deleteErr != nil {
-				klog.Errorf(util.Log(ctx, "failed to delete rbd image: %s with error: %v"),
-					rbdVol, deleteErr)
-				return nil, deleteErr
-			}
-			return nil, err
+		if _, ok := err.(ErrFlattenInProgress); ok {
```
nixpanic, line 234: use `error.Is(err, ErrFlattenInProgress)` as described at https://github.com/golang/go/wiki/ErrorValueFAQ#how-should-i-change-my-error-handling-code-to-work-with-the-new-features

internal/rbd/controllerserver.go
```diff
@@ -262,27 +258,117 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	return &csi.CreateVolumeResponse{Volume: volume}, nil
 }
 
-func createBackingImage(ctx context.Context, cr *util.Credentials, rbdVol *rbdVolume, rbdSnap *rbdSnapshot) error {
+func checkFlatten(ctx context.Context, rbdVol *rbdVolume, cr *util.Credentials) error {
+	err := rbdVol.flattenRbdImage(ctx, cr, false)
+	if err != nil {
+		if _, ok := err.(ErrFlattenInProgress); ok {
```
nixpanic, line 264: Same here, use the new `errors.Is()`

internal/rbd/controllerserver.go
```diff
@@ -262,27 +258,117 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+	rbdSnap := &rbdSnapshot{}
+	if acquired := cs.SnapshotLocks.TryAcquire(snapshotID); !acquired {
+		klog.Infof(util.Log(ctx, util.SnapshotOperationAlreadyExistsFmt), snapshotID)
+		return status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, snapshotID)
+	}
+	defer cs.SnapshotLocks.Release(snapshotID)
+
+	err := genSnapFromSnapID(ctx, rbdSnap, snapshotID, cr)
+	if err != nil {
+		if _, ok := err.(util.ErrPoolNotFound); ok {
```
nixpanic, line 290: Same here, `errors.Is()`

Madhu-1: Yes I already tried it wont work, would like to keep it as it is

internal/rbd/rbd_util.go
Madhu-1: Imagespec is not used anymore, The relevant changes are removed in final commit you want this change to be part of same commit?

internal/rbd/rbd_util.go
Madhu-1: This is not new implementation the code already , if any changes are required we can do it as an enhancement

Madhu-1:
> The commit message of [db5d142](https://github.com/ceph/ceph-csi/commit/db5d142aa35d9919de1d352cee0430ebf0913c17) (rbd: Implement snapshot and clone from snapshot) contains lots of details. Is this also available in some .md file in the repo? If not, adding it as a design would be great (and mention in the comments in the implementation that the doc exists, so it gets updated)

I can do it as a follow up pr as I still need to address few todo. Will add a md file for both snapshot , restore and clone

internal/rbd/rbd_util.go
Madhu-1: corrected

Madhu-1:
@nixpanic comments are addressed PTAL

internal/rbd/controllerserver.go
nixpanic: Ok, I'm considering to clean the other occasions up. Since Golang 1.13 these new ways are recommended.

Not a blocker, it can be done later.

nixpanic:
Please keep the PR text updated with TODOs. Add them to the list, and mark them done (don't remove them, keep them listed).

Madhu-1:
> Please keep the PR text updated with TODOs. Add them to the list, and mark them done (don't remove them, keep them listed).

@nixpanic done.

internal/rbd/rbd_util.go
nixpanic: ok!

internal/rbd/rbd_util.go
nixpanic: hmm, I don't like it. It should probably get cleaned up at one point.

Madhu-1: This is fixed now.

Madhu-1:
created a Tracker issue for TODO https://github.com/ceph/ceph-csi/issues/1188. once this is merged the next plan is to fix the TODO which is present in tracker

Madhu-1:
> I'm ok with getting this merged. There are many TODO's left. What is the plan to address them? If you want to sent them as follow-up PRs that is fine, but each TODO needs an issue in that case, and add it to the open TODO list in this PR.
> 
> Make sure all of the TODO-issues get added to the "Backup & Restore" project, and/or the v3.0.0 milestone. Without that, tracking the status of the feature will be difficult.

opened a single issue to track all the TODO https://github.com/ceph/ceph-csi/issues/1188

Madhu-1:
@ShyamsundarR @nixpanic PTAL.

internal/rbd/nodeserver.go
```diff
@@ -226,6 +245,27 @@ func (ns *NodeServer) stageTransaction(ctx context.Context, req *csi.NodeStageVo
...
 
+	if kernelRelease == "" {
+		// fetch the current running kernel info
+		kernelRelease, err = util.GetKernelVersion()
+		if err != nil {
+			return transaction, err
+		}
+	}
+	if !util.CheckKernelSupport(kernelRelease, deepFlattenSupport) {
+		if !skipForceFlatten {
```
nixpanic, line 256: you might want to combine these two if-statements so that the code block below is not indented 2x

Madhu-1: Any suggestion?

Madhu-1:
> Was there a request for `skipForceFlatten`? It would need an e2e test-case, I think.

This cannot be tested in CI as kernel in CI doesnt support deep-flatten
 
> Not sure the commit message `high performance` is suitable. Volume creation might be faster when flattening is skipped, but that also means there is more indirection when reading/writing.

Updated the commit message

docs/deploy-rbd.md
```diff
@@ -26,22 +26,25 @@ make image-cephcsi
...
+| `--rbdhardmaxclonedepth` | `8`                   | Hard limit for maximum number of nested volume clones that are taken before a flatten occurs                                                                                                                                                                                         |
+| `--rbdsoftmaxclonedepth` | `4`                   | Soft limit for maximum number of nested volume clones that are taken before a flatten occurs                                                                                                                                                                                         |
+| `--skipforceflatten`     | `false`               | skip image flattening on kernel < 5.2 which support mapping of rbd images which has the deep-flatten feature                                                                                                                                                                         |
```
dillaman, line 47: I thought we were killing this in favor of just force-flattening an image upon first use on an incompatible node?

internal/journal/voljournal.go
ShyamsundarR: Tracked in the TODO list.

docs/deploy-rbd.md
Madhu-1: @dillaman This is just an option provided to the user if the kernel version which supports deep-flatten is not in the list that cephcsi maintains, the user can still skip flattening of the rbd images before use in NodeStageRequest, the user doesn't need to wait for a new cephcsi release which adds the kernel to the list of the kernel it maintains that supports the deep-flatten feature.

internal/rbd/controllerserver.go
```diff
@@ -519,6 +519,11 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 
 		// All errors other than ErrImageNotFound should return an error back to the caller
 		case ErrImageNotFound:
+			err = rbdVol.removeImageFromTrash()
+			if err != nil {
+				klog.Errorf(util.Log(ctx, "failed to delete rbd image: %s from trash with error: %v"), rbdVol, err)
+				return nil, status.Error(codes.Internal, err.Error())
```
nixpanic, line 525: Is this a critical error? Is the trash not cleaned up at a later time automatically?
Deleting an image should probably succeed when the image is gone (which could be in the trash).
I think logging the error (and mentioning the trash is cleared later) is sufficient?

docs/deploy-rbd.md
```diff
@@ -26,22 +26,25 @@ make image-cephcsi
...
+| `--histogramoption`      | `0.5,2,6`             | [Deprecated] Histogram option for grpc metrics, should be comma separated value (ex:= "0.5,2,6" where start=0.5 factor=2, count=6)                                                                                                                                                   |
+| `--domainlabels`         | _empty_               | Kubernetes node labels to use as CSI domain labels for topology aware provisioning, should be a comma separated value (ex:= "failure-domain/region,failure-domain/zone")                                                                                                             |
+| `--rbdhardmaxclonedepth` | `8`                   | Hard limit for maximum number of nested volume clones that are taken before a flatten occurs                                                                                                                                                                                         |
```
humblec, line 45: @Madhu-1 can we restrict the readme changes only for newly added parameters ?

internal/rbd/controllerserver.go
```diff
@@ -894,6 +899,11 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 			klog.Errorf(util.Log(ctx, "failed to delete rbd image: %s/%s with error: %v"), rbdVol.Pool, rbdVol.VolName, err)
 			return nil, status.Error(codes.Internal, err.Error())
 		}
+		err = rbdVol.removeImageFromTrash()
```
nixpanic, line 902: Same here, removing from trash is probably not a real failure (from the CSI specs point of view)

docs/deploy-rbd.md
Madhu-1: The table alignment need to be changes as the number of char in line increases

e2e/rbd.go
```diff
@@ -263,33 +263,90 @@ var _ = Describe("RBD", func() {
...
+						err = createSnapshot(&snap, deployTimeout)
+						if err != nil {
+							Fail(err.Error())
+						}
+					}
+
+					imageList := listRBDImages(f)
+					// total images in cluster is 1 parent rbd image+ total snaps
+					if len(imageList) != totalCount+1 {
+						e2elog.Logf("backend images not matching kubernetes pvc,snap count,image count %d kubernetes resource count %d", len(imageList), totalCount+1)
```
humblec, line 282: this error message has to be revisited or can be better formatted.

Madhu-1: Sure, can do a enhancement

e2e/rbd.go
```diff
@@ -263,33 +263,90 @@ var _ = Describe("RBD", func() {
...
 					}
-					if len(snapList) != 1 {
-						e2elog.Logf("backend snapshot not matching kubernetes snap count,snap count = % kubernetes snap count %d", len(snapList), 1)
-						Fail("validate backend snapshot failed")
+					// create clone and bind it to an app
+					for i := 0; i < totalCount; i++ {
+						name := fmt.Sprintf("%s%d", f.UniqueName, i)
+						pvcClone.Spec.DataSource.Name = name
+						pvcClone.Namespace = f.UniqueName
+						appClone.Namespace = f.UniqueName
```
humblec, line 299: these  2 assignments can be clubbed to single.

internal/rbd/rbd_journal.go
```diff
@@ -280,6 +280,24 @@ func (rv *rbdVolume) Exists(ctx context.Context) (bool, error) {
 		return false, err
 	}
 
+	rv.ImageID, err = j.GetStoredImageID(ctx, rv.JournalPool, rv.ReservedID, rv.conn.Creds)
+	if _, ok := err.(util.ErrKeyNotFound); ok {
```
nixpanic, line 284: is this safe if `err == nil`?

e2e/rbd.go
```diff
@@ -263,33 +263,90 @@ var _ = Describe("RBD", func() {
...
 
-					validatePVCAndAppBinding(pvcClonePath, appClonePath, f)
-
-					err = deleteSnapshot(&snap, deployTimeout)
-					if err != nil {
-						Fail(err.Error())
+					imageList = listRBDImages(f)
+					// total images in cluster is 1 parent rbd image+ total
+					// snaps+ total clones
+					if len(imageList) != totalCount+totalCount+1 {
```
humblec, line 309: For clarity , we could make use of `totalSnapCount` or `totalCloneCount` variables instead of relying on `totalCount`

internal/rbd/rbd_journal.go
Madhu-1: Yes its safe

e2e/rbd.go
```diff
@@ -263,33 +263,90 @@ var _ = Describe("RBD", func() {
...
+
+					// delete parent pvc
 					err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
 					if err != nil {
 						Fail(err.Error())
 					}
+
+					imageList = listRBDImages(f)
+					// total images in cluster is total snaps+ total clones
+					if len(imageList) != totalCount+totalCount {
```
humblec, line 322: same here , totalSnapCount for snapshot is better.

internal/rbd/controllerserver.go
Madhu-1: Removed now, lets ceph handle it, if required this can be added later also

Madhu-1:
@humblec @nixpanic can you please review it?

e2e/rbd.go
humblec: @Madhu-1 just to confirm these changes are tracked for next set of PRs ?

Madhu-1: @humblec changes are already done

PR #1162, opened 2020-06-11
ci: Enable stale bot to auto close the old issues

humblec:
atm, the daysUntilStale is 60 which is really huge and it
will take some time to trigger the bot itself and come into
action. This PR reduce or make stale bot active on cases (7days)
to reduce the huge backlog and set daysUntilClose to 25 days for now.
It gives enough window for issues to be revisited.

### DIFF
--- a/.github/stale.yaml
+++ b/.github/stale.yaml
@@ -23,8 +23,8 @@ limitPerRun: 5
 
 # Specify configuration settings that are specific to issues
 issues:
-  daysUntilStale: 60
-  daysUntilClose: 7
+  daysUntilStale: 7
+  daysUntilClose: 25
   markComment: >
     This issue has been automatically marked as stale because it has not had
     recent activity. It will be closed in a week if no further activity occurs.


.github/stale.yaml
```diff
@@ -23,8 +23,8 @@ limitPerRun: 5
 
 # Specify configuration settings that are specific to issues
 issues:
-  daysUntilStale: 60
-  daysUntilClose: 7
+  daysUntilStale: 7
```
Madhu-1, line 26: it would be better to keep it in reverse order, marking it as stale in only 7 days doesn't seem correct. we can mark it as stale for 25/30 days and close it after 7 days if there is no activity

humblec: @Madhu-1 exactly that was the condition or previous to this PR. With this change the aim is to see the stale bot come in action after 7 days or not. If it happens, we could exchange the params. Otherwise we have to wait for another 25/30 days to see or check stale bot behaviiour. I dont see a mechanism to run a single trigger forcefully.

nixpanic: I do not see the need for testing this. Can you not place an event in your calendar to check on an issue that is meant to become stale in 60 days?

nixpanic:
@humblec, I think this was fixed with #1474?

humblec:
Yeah.. Closing it

Issue #1168, opened 2020-06-16
cephfs: unmount targetpath if chmod fails 


Madhu-1:
currently, we do not unmount the target path if the chmod fails at https://github.com/ceph/ceph-csi/blob/2793b79ad8e04628ade7bce4d16c4954363d8947/internal/cephfs/nodeserver.go#L215-L218. This can cause issues with the subsequent request comes in as the mount is already present we return success https://github.com/ceph/ceph-csi/blob/2793b79ad8e04628ade7bce4d16c4954363d8947/internal/cephfs/nodeserver.go#L200-L203  even if the permission is not set on the target path.

PR #55, opened 2018-07-31
Add nonempty option to ceph-fuse to support ReadWriteMany

nak3:
fuse mount does not allow to mount directory if it contains some
files. Due to this, currently scaled pod with cephfs failed to mount
by ceph-fuse.

This patch adds nonempty option to ceph-fuse command to support
ReadWriteMany with ceph-fuse.

Fixes https://github.com/ceph/ceph-csi/issues/54

PR #1169, opened 2020-06-16
Add support for cephfs ROX PVC

Madhu-1:
This PR adds the support for the mounting for ReadOnlyMany cephfs PVC. and it also fixes the bug https://github.com/ceph/ceph-csi/issues/1168

closes https://github.com/ceph/ceph-csi/issues/1168


internal/cephfs/nodeserver.go
```diff
@@ -181,6 +184,18 @@ func (*NodeServer) mount(ctx context.Context, volOptions *volumeOptions, req *cs
 			err)
 		return status.Error(codes.Internal, err.Error())
 	}
+	if !csicommon.MountOptionContains(kernelMountOptions, readOnly) && !csicommon.MountOptionContains(fuseMountOptions, readOnly) {
+		// #nosec - allow anyone to write inside the stagingtarget path
+		err = os.Chmod(stagingTargetPath, 0777)
```
nixpanic, line 189: Should this be done recursive for all the contents? I guess it was not the case, but for creating a volume from a snapshot or other pvc, it may be needed.

internal/cephfs/volumemounter.go
```diff
@@ -229,12 +229,13 @@ func mountFuse(ctx context.Context, mountPoint string, cr *util.Credentials, vol
 		"-c", util.CephConfigPath,
 		"-n", cephEntityClientPrefix + cr.ID, "--keyfile=" + cr.KeyFile,
 		"-r", volOptions.RootPath,
-		"-o", "nonempty",
 	}
 
+	mntOpt := "nonempty"
```
nixpanic, line 234: Is `nonempty` a valid mount option? This looks like a very ugly hack.

internal/cephfs/nodeserver.go
Madhu-1: This is needed to provide write access to normal users

internal/cephfs/nodeserver.go
```diff
@@ -150,6 +151,29 @@ func (*NodeServer) mount(ctx context.Context, volOptions *volumeOptions, req *cs
 
 	klog.V(4).Infof(util.Log(ctx, "cephfs: mounting volume %s with %s"), volID, m.name())
 
+	if req.VolumeCapability.AccessMode.Mode == csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY ||
+		req.VolumeCapability.AccessMode.Mode == csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY {
+		readOnly := "ro"
+		switch m.(type) {
```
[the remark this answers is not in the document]
Madhu-1: Yeah once the feature works we can do enhancement or do simplification

internal/cephfs/volumemounter.go
Madhu-1: Not added now, it was added in https://github.com/ceph/ceph-csi/pull/55

internal/cephfs/nodeserver.go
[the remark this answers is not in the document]
nixpanic: Adjusting mount options in string format (as opposed to string-slices) is not as common as I thought. Anyway, PR #1174 has been posted to make this part of the code a little cleaner.

internal/cephfs/nodeserver.go
nixpanic: Yes, I understand that, but the current command assumes there is no content on the volume that needs changes in permissions. In case contents is there (cloned volume), the `chmod` should probably change the permissions of everything on it.

internal/cephfs/volumemounter.go
nixpanic: hmm, ok

internal/cephfs/nodeserver.go
Madhu-1: i hope same is going to happen without this change also, We need to have a separate issue to handle it can you please open one?

nixpanic: #1183 has been opened for this

internal/cephfs/volumemounter.go
```diff
@@ -229,12 +229,13 @@ func mountFuse(ctx context.Context, mountPoint string, cr *util.Credentials, vol
...
 	if volOptions.FuseMountOptions != "" {
-		args = append(args, ","+volOptions.FuseMountOptions)
+		mntOpt += "," + volOptions.FuseMountOptions
```
nixpanic, line 236: You can use `util.MountOptionsAdd()` and drop the if statement

humblec:
Anything pending here ?

Madhu-1:
> Anything pending here ?

Nothing, Review is pending

PR #1170, opened 2020-06-16
Make use of latest go ceph library

humblec:
The go-ceph version 0.4.0 is available now which got
some important library changes required for ceph csi
project. This PR update the library to latest released
version.


humblec:
```
Golang errors in file: [/home/travis/gopath/src/github.com/ceph/ceph-csi/internal/rbd/rbd_util.go]:

  > [line 187 : column 28] - cannot use librbd.RbdImageOptionDataPool (constant unknown with invalid type) as rbd.ImageOption value in argument to options.SetString
```

phlogistonjohn:
Looks like ceph-csi was still on go-ceph v0.2.0. We renamed those constants in [v0.3.0](https://github.com/ceph/go-ceph/releases/tag/v0.3.0). However, I don't remember fully removing them, which is odd.

phlogistonjohn:
OK, so that error is coming from one of the linters, not the compiler. The compiler builds everything OK *if* -tags nautilus is specified. Right now, I can't figure out which linter this is (I'm not a fan of gometalinter) and why it is having trouble with this. It doesn't seem to be the aliasing... changing the line to ImageOptionDataPool instead of RbdImageOptionDataPool doesn't fix the problem. The error message it emits is practically the same.

humblec:
> please use `rebase` as prefix for the commit.
> 
> As go-ceph v0.4 supports Octopus, but ceph-csi uses Nautilus, the `go build`, `go test` and possibly other commands need to pass `-tags nautilus` so that features available in Octopus do not get build (and building will fail).

@nixpanic tried `-tags nautliius` , waiting for the travis run.

nixpanic:
You also need to pass `-tags nautilus` to `gosec`:
```patch
diff --git a/scripts/gosec.sh b/scripts/gosec.sh
index 83bd49974..83090b01e 100755
--- a/scripts/gosec.sh
+++ b/scripts/gosec.sh
@@ -6,7 +6,7 @@ if [[ -x "$(command -v gosec)" ]]; then
   # gosec does not support -mod=vendor, so fallback to non-module support and
   # assume all dependencies are available in ./vendor already
   export GO111MODULE=off
-  find cmd internal -type d -print0 | xargs --null gosec
+  find cmd internal -type d -print0 | xargs --null gosec -tags nautilus
 else
   echo "WARNING: gosec not found, skipping security tests" >&2
 fi
```

Makefile
```diff
@@ -105,7 +105,7 @@ gosec:
 	./scripts/gosec.sh
 
 func-test:
-	go test -mod=vendor github.com/ceph/ceph-csi/e2e $(TESTOPTIONS)
+	go test -tags nautilus -mod=vendor github.com/ceph/ceph-csi/e2e $(TESTOPTIONS)
```
Madhu-1, line 108: instead of specifying tag in multiple places define a variable and use it

humblec: @Madhu-1 sure, can that be a followup PR ?

Madhu-1: the change will be minimal as below, IMO as this PR is introducing this change we can do it here. 
```
ifndef CEPH_TAG
CEPH_TAG=nautilus
endif

go test -tags $(CEPH_TAG) -mod=vendor github.com/ceph/ceph-csi/e2e $(TESTOPTIONS)
```
the above tag need to be passed to other scripts also when its invoked from Makfile

humblec: ok.. if that's the case, let me address in this PR itself. On it.

humblec: @Madhu-1 ptal.. The changes required are in. For `travis-*.sh` its hardcoded. Let me know if any further changes are required.

scripts/gosec.sh
```diff
@@ -2,11 +2,32 @@
 
 set -o pipefail
 
+case "${1:-}" in
```
nixpanic, line 5: if you want to parse it, you can better use real options with `getopt`. In case `CEPH_TAG` is not set, just do not pass `-tags` at all.

Requiring changing scripts on a Ceph update is not ideal.

scripts/test-go.sh
```diff
@@ -3,10 +3,32 @@
 GOPACKAGES="$(go list -mod=vendor ./... | grep -v -e vendor -e e2e)"
 COVERFILE="${GO_COVER_DIR}/profile.cov"
 
+
+case "${1:-}" in
```
nixpanic, line 7: same here

scripts/gosec.sh
humblec: @nixpanic it looks like we do similar parsing in other scripts too. Considering that, may be atleast time time we will stick to this parsing. We can change all together later if required.

scripts/test-go.sh
humblec: same as above.

humblec:
>I would prefer to have slightly adjusted subjects of the changes:

    util: Make use of latest go ceph library -> rebase: ...
    rebase: Add tags nautilustogo testandgo build`` -> build: ...

It would be better to check for CEPH_TAG in the environment, and only add -tags .. then. Depending on changing the scripts when a new Ceph version is released is not a nice solution for this.
>
One of the thought here is, this tag switching can be used for local dev environment against go test, go sec..etc.  Also, I was in preference to handle this tag passing ..etc in a different PR altogether. But, it seemed that we have to use variables and pass it from makefile to other scripts. So, achieving the same with little more flexibility.

scripts/gosec.sh
nixpanic: Sorry, where are we parsing the Ceph version? I can not find it in other scripts.

nixpanic:
>  But, it seemed that we have to use variables and pass it from makefile to other scripts. So, achieving the same with little more flexibility.

Or by using environment variables that us used by most/many of the scripts in Travis CI. Passing parameters to a script is IMHO better, but only if done correctly and in a way it can easily be extended. Positional parameters to a script do not address that. Just use environment variables for the scripts that do not use them yet?

nixpanic:
I'm including `CEPH_VERSION` in #1182 too. Shall I update this PR with a rebase on top of that?

nixpanic:
This has now been rebased on top of #1182. That one should get merged before this. Adding `DNM` to prevent confusion.

humblec:
> Obviously I approve this, as I have updated the PR to the current state grin

Thats great.. approval from my side too!

nixpanic:
@mergifyio rebase

nixpanic:
#1182 got merged, so this is ready now too (dropped the `DNM` label)

Makefile
nixpanic: @Madhu-1 this has been addressed by building on top of #1182

humblec:
@Madhu-1 can you approve this ?

PR #1174, opened 2020-06-17
util: add MountOptionsAdd() to add mount options

nixpanic:

There are a few occasions where the mount-options string is modified. This is not always done cleanly and could cause weird issues (fuse cephfs not allowing empty options).

The new util function MountOptionsAdd() takes a string, and adds the option in case it does not exist yet. It makes sure to return a correctly formatted (`,` separated, none at the start, nor end) string.

Idea while reviewing #1169 that does some more mount option string manipulations and corrections.

### DIFF
--- a/internal/cephfs/volumemounter.go
+++ b/internal/cephfs/volumemounter.go
@@ -285,17 +285,13 @@ func mountKernel(ctx context.Context, mountPoint string, cr *util.Credentials, v
 		fmt.Sprintf("%s:%s", volOptions.Monitors, volOptions.RootPath),
 		mountPoint,
 	}
-	optionsStr := fmt.Sprintf("name=%s,secretfile=%s", cr.ID, cr.KeyFile)
-	if volOptions.FsName != "" {
-		optionsStr += fmt.Sprintf(",mds_namespace=%s", volOptions.FsName)
-	}
-	if volOptions.KernelMountOptions != "" {
-		optionsStr += fmt.Sprintf(",%s", volOptions.KernelMountOptions)
-	}
 
-	if !strings.Contains(volOptions.KernelMountOptions, netDev) {
-		optionsStr += fmt.Sprintf(",%s", netDev)
+	optionsStr := fmt.Sprintf("name=%s,secretfile=%s", cr.ID, cr.KeyFile)
+	mdsNamespace := ""
+	if volOptions.FsName != "" {
+		mdsNamespace = fmt.Sprintf("mds_namespace=%s", volOptions.FsName)
 	}
+	optionsStr = util.MountOptionsAdd(optionsStr, mdsNamespace, volOptions.KernelMountOptions, netDev)
 
 	args = append(args, "-o", optionsStr)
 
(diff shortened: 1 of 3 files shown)


internal/util/util.go
```diff
@@ -193,3 +193,25 @@ func Mount(source, target, fstype string, options []string) error {
 	dummyMount := mount.New("")
 	return dummyMount.Mount(source, target, fstype, options)
 }
+
+// MountOptionsAdd adds the `add` mount option to the `options` and returns a
+// new string. In case `add` is already present in the `options`, `add` is not
+// added again.
+func MountOptionsAdd(options, add string) string {
```
Madhu-1, line 200: add can be variadic

nixpanic: sure, I can do that

nixpanic: Done, please review again. Thanks!

internal/cephfs/volumemounter.go
```diff
@@ -289,16 +289,7 @@ func mountKernel(ctx context.Context, mountPoint string, cr *util.Credentials, v
...
-		optionsStr += fmt.Sprintf(",mds_namespace=%s", volOptions.FsName)
-	}
-	if volOptions.KernelMountOptions != "" {
-		optionsStr += fmt.Sprintf(",%s", volOptions.KernelMountOptions)
-	}
-
-	if !strings.Contains(volOptions.KernelMountOptions, netDev) {
-		optionsStr += fmt.Sprintf(",%s", netDev)
-	}
+	optionsStr = util.MountOptionsAdd(optionsStr, volOptions.FsName, volOptions.KernelMountOptions, netDev)
```
Madhu-1, line 292: `mds_namespace=%s"` is missing here.

nixpanic: no, that is already set in `optionsStr` a line higher?

Madhu-1: nope, can you please check again
https://github.com/ceph/ceph-csi/blob/1aa106ff1d6bfc6f1b921e885d22b7aff9fb276f/internal/cephfs/volumemounter.go#L281-L297

nixpanic: Ah, now I see the `mds_namespace=` piece is missing, the option is not correct. Thanks!

nixpanic: this should have been addressed now too :facepalm:

internal/util/util.go
```diff
@@ -193,3 +193,35 @@ func Mount(source, target, fstype string, options []string) error {
 	dummyMount := mount.New("")
 	return dummyMount.Mount(source, target, fstype, options)
 }
+
+// MountOptionsAdd adds the `add` mount options to the `options` and returns a
+// new string. In case `add` is already present in the `options`, `add` is not
+// added again.
+func MountOptionsAdd(options string, add ...string) string {
+	opts := strings.Split(options, ",")
```
phlogistonjohn, line 201: Idle speculation: perhaps instead of pulling a string apart over and over, mount options could be a new type over a slice and then only converted to a string as the last step before use?

nixpanic: Yes, but there is a weird mix of strings formatted mount options and string slices. Reworking this would be a great cleanup, but also needs a bit more time.

nixpanic:
@mergifyio rebase

nixpanic:
> This pull request now has conflicts with the target branch. Could you please resolve conflicts and force push the corrected changes? pray

These have been resolved, please review again.

Madhu-1:
@humblec PTAL

nixpanic:
@humblec, this one is still waiting for your review too

Issue #1123, opened 2020-06-02
Add support to create multiple subvolume groups 


Madhu-1:
in https://github.com/ceph/ceph-csi/pull/931 we have added support to have subvolumegroup configurable(which means we can have more than one subvolume group in cephcsi). with the current code base the subvolumegroup will be created once ie https://github.com/ceph/ceph-csi/blob/44da7ffb4e4d8861265f3ecd421cd06dfce9f34a/internal/cephfs/volume.go#L90 we need to make changes to support the creation of multiple subvolumegroups.

Madhu-1:
@Yuggupta27 would you like to give a  try for this one, I can help you out if you need anything

Yuggupta27:
> @Yuggupta27 would you like to give a try for this one, I can help you out if you need anything

Sure. Please Assign.

ShyamsundarR:
Something to think about would be, when a subvolumegroup and the cephx credentials for the same are pre-created, to restrict access just to that subvolume group, would the called out function still work?

IOW, we started out with creating the subvolume group in CSI, but it may so happen that for security reasons, these are pre-created and the credentials for CSI are restricted to this subvolumegroup. In such cases, it would help to understand if the `create` command above still works, or would error out.

Madhu-1:
we need to check what if the subvolume is already created, and the csi user doesn't have the permission to create it, how we can check this if the create subvolume group fails in this case and on what basis we need to continue with creating subvolumes etc.
 we need to create a user who is having access to create subvolume, not the subvolume group and test this out,  the caps can be found in this comment https://github.com/ceph/ceph-csi/pull/437#issuecomment-511460712 need to modify the path according to our requirement. 

can we have a user who is not having any read and write permission or only read permission to subvolume groups. if someone can provide the caps it would be useful to test it out.
cc @ShyamsundarR @kotreshhr @joscollin

joscollin:
> can we have a user who is not having any read and write permission or only read permission to subvolume groups.

There is a `--mode` parameter for subvolumegroup create command. 

From your description, I don't quiet understand what you are trying to achieve.

ShyamsundarR:
> > can we have a user who is not having any read and write permission or only read permission to subvolume groups.
> 
> There is a `--mode` parameter for subvolumegroup create command.
> 
> From your description, I don't quiet understand what you are trying to achieve.

Pasting from #437 on caps required to work with a subvolume group,
```
The auth IDs capabilities would be,

[...]

PR #1175, opened 2020-06-17
cephfs: Add support for multiple subvolumegroups

Yuggupta27:
With the current code base, the subvolumegroup will
be created once, and even for a different cluster,
subvolumegroup creation is not allowed again.

Added support multiple subvolumegroups creation by 
validating one subvolumegroup creation per cluster.

Fixes: #1123

### DIFF
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -176,12 +176,12 @@ var _ = Describe("cephfs", func() {
 			})
 
 			By("create a storage class with pool and a PVC then Bind it to an app", func() {
-				createCephfsStorageClass(f.ClientSet, f, true)
+				createCephfsStorageClass(f.ClientSet, f, true, "")
 				validatePVCAndAppBinding(pvcPath, appPath, f)
 				deleteResource(cephfsExamplePath + "storageclass.yaml")
 			})
 
-			createCephfsStorageClass(f.ClientSet, f, false)
+			createCephfsStorageClass(f.ClientSet, f, false, "")
 
 			By("create and delete a PVC", func() {
 				By("create a PVC and Bind it to an app", func() {
@@ -257,6 +257,38 @@ var _ = Describe("cephfs", func() {
 						Fail(err.Error())
 					}
 				})
+
+				By("validate multiple subvolumegroup creation", func() {
+					deleteResource(cephfsExamplePath + "storageclass.yaml")
+					// re-define configmap with information of multiple clusters.
+					subvolgrpInfo := map[string]string{
+						"clusterID-1": "subvolgrp1",
+						"clusterID-2": "subvolgrp2",
+					}
+					createCustomConfigMap(f.ClientSet, cephfsDirPath, subvolgrpInfo)
+					createCephfsStorageClass(f.ClientSet, f, false, "clusterID-1")
+					validatePVCAndAppBinding(pvcPath, appPath, f)
+					deleteResource(cephfsExamplePath + "storageclass.yaml")
+					// verify subvolumegroup creation.
+					err := validateSubvolumegroup(f, "subvolgrp1")
+					if err != nil {
+						Fail(err.Error())
+					}
+
+					// create resources and verify subvolume group creation
+					// for the second cluster.
+					createCephfsStorageClass(f.ClientSet, f, false, "clusterID-2")
+					validatePVCAndAppBinding(pvcPath, appPath, f)
+					deleteResource(cephfsExamplePath + "storageclass.yaml")
+					err = validateSubvolumegroup(f, "subvolgrp2")
+					if err != nil {
+						Fail(err.Error())
+					}
+					deleteConfigMap(cephfsDirPath)
+				})
+
+				createConfigMap(cephfsDirPath, f.ClientSet, f)
+				createCephfsStorageClass(f.ClientSet, f, false, "")
 
 				By("Resize PVC and check application directory size", func() {
 					v, err := f.ClientSet.Discovery().ServerVersion()
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -258,7 +258,7 @@ func getStorageClass(path string) scv1.StorageClass {
 	return sc
 }
 
-func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, enablePool bool) {
+func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, enablePool bool, clusterID string) {
 	scPath := fmt.Sprintf("%s/%s", cephfsExamplePath, "storageclass.yaml")
 	sc := getStorageClass(scPath)
 	sc.Parameters["fsName"] = "myfs"
@@ -278,6 +278,10 @@ func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, en
 	Expect(stdErr).Should(BeEmpty())
 	// remove new line present in fsID
 	fsID = strings.Trim(fsID, "\n")
+
+	if clusterID != "" {
+		fsID = clusterID
+	}
 	sc.Namespace = cephCSINamespace
 	sc.Parameters["clusterID"] = fsID
 	_, err := c.StorageV1().StorageClasses().Create(context.TODO(), &sc, metav1.CreateOptions{})
@@ -1206,4 +1210,55 @@ func checkPVSelectorValuesForPVC(f *framework.Framework, pvc *v1.PersistentVolum
 func addTopologyDomainsToDSYaml(template, labels string) string {
 	return strings.ReplaceAll(template, "# - \"--domainlabels=failure-domain/region,failure-domain/zone\"",
 		"- \"--domainlabels="+labels+"\"")
+}
+
+// createCustomConfigMap provides multiple clusters information.
+func createCustomConfigMap(c kubernetes.Interface, pluginPath string, subvolgrpInfo map[string]string) {
+	path := pluginPath + configMap
+	cm := v1.ConfigMap{}
+	err := unmarshal(path, &cm)
+	Expect(err).Should(BeNil())
+
+	// get mon list
+	mons := getMons(rookNamespace, c)
+	// get clusterIDs
+	var clusterID []string
+	for key := range subvolgrpInfo {
+		clusterID = append(clusterID, key)
+	}
+	conmap := []util.ClusterInfo{
+		{
+			ClusterID: clusterID[0],
+			Monitors:  mons,
+		},
+		{
+			ClusterID: clusterID[1],
+			Monitors:  mons,
+		}}
+	for i := 0; i < len(subvolgrpInfo); i++ {
+		conmap[i].CephFS.SubvolumeGroup = subvolgrpInfo[clusterID[i]]
+	}
+	data, err := json.Marshal(conmap)
+	Expect(err).Should(BeNil())
+	cm.Data["config.json"] = string(data)
+	cm.Namespace = cephCSINamespace
+	// since a configmap is already created, update the existing configmap
+	_, updateErr := c.CoreV1().ConfigMaps(cephCSINamespace).Update(context.TODO(), &cm, metav1.UpdateOptions{})
+	Expect(updateErr).Should(BeNil())
+}
+
+// validateSubvolumegroup validates whether subvolumegroup is present.
+func validateSubvolumegroup(f *framework.Framework, subvolgrp string) error {
+	cmd := fmt.Sprintf("ceph fs subvolumegroup getpath myfs %s", subvolgrp)
+	stdOut, err := execCommandInToolBoxPod(f, cmd, rookNamespace)
+	Expect(err).Should(BeEmpty())
+	if err != "" {
+		return fmt.Errorf("error subvolumegroup %s doesn't exist", subvolgrp)
+	}
+	expectedGrpPath := "/volumes/" + subvolgrp
+	stdOut = strings.TrimSpace(stdOut)
+	if stdOut != expectedGrpPath {
+		return fmt.Errorf("error unexpected group path. Found: %s", stdOut)
+	}
+	return nil
 }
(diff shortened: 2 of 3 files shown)


Madhu-1:
Please add E2E testing to support multiple subvolume groups

Yuggupta27:
> Please add E2E testing to support multiple subvolume groups

On it!

internal/cephfs/volume.go
```diff
@@ -144,7 +150,13 @@ func createVolume(ctx context.Context, volOptions *volumeOptions, cr *util.Crede
...
 // CreateVolume to resize the subvolume.
 func resizeVolume(ctx context.Context, volOptions *volumeOptions, cr *util.Credentials, volID volumeID, bytesQuota int64) error {
-	if supported, ok := resizeSupportedList[volOptions.ClusterID]; supported || !ok {
+	// verify if corresponding ClusterID entry is present in the map,
+	// and if not, initialize with default values(false).
+	if _, ok := clusterAdditionalInfo[volOptions.ClusterID]; !ok {
+		clusterAdditionalInfo[volOptions.ClusterID] = &additionalClusterInfo{}
+	}
+
+	if supported := clusterAdditionalInfo[volOptions.ClusterID].resizeSupported; supported {
```
Madhu-1, line 159: This seems incorrect.this will be subvolume resize is called only if the resizeSupported is set to true. we  also need to try to resize for the first time to check whether resize is supported or not

Yuggupta27: Thanks, I'll make the changes.

Yuggupta27: I had a slight doubt regarding this, were we checking whether resize is supported beforehand in https://github.com/ceph/ceph-csi/blob/master/internal/cephfs/volume.go#L147 ?

Yuggupta27: I had a doubt regarding this, is resizevolume skipped for the first run since resize supported is set to false? https://github.com/ceph/ceph-csi/blob/master/internal/cephfs/volume.go#L147

Madhu-1: we have double check here, `if supported, ok := resizeSupportedList[volOptions.ClusterID]; supported || !ok {` either its supported or key not present

internal/cephfs/volume.go
```diff
@@ -30,14 +30,13 @@ import (
 )
 
 var (
-	// cephfsInit is used to create "csi" subvolume group for the first time the csi plugin loads.
-	// Subvolume group create gets called every time the plugin loads, though it doesn't result in error
-	// its unnecessary
-	cephfsInit = false
+	// checkAdditionalInfo stores the mapping for clusterID and resize is
+	// supported in the particular cluster and whether subvolumegroup is
+	// created or not.
```
humblec, line 35: may be the requirement of knowing about these additional information can be put in the source comment? why or where its useful?

Yuggupta27: > may be the requirement of knowing about these additional information can be put in the source comment? why or where its useful?

Thanks. I'll add it.

Yuggupta27:
@Madhu-1 Please take a look.

e2e/utils.go
```diff
@@ -1207,3 +1211,43 @@ func addTopologyDomainsToDSYaml(template, labels string) string {
 	return strings.ReplaceAll(template, "# - \"--domainlabels=failure-domain/region,failure-domain/zone\"",
 		"- \"--domainlabels="+labels+"\"")
 }
+
+// createCustomConfigMap provides multiple clusters information.
+func createCustomConfigMap(pluginPath string, c kubernetes.Interface, clusterID1, clusterID2 string) {
```
Madhu-1, line 1216: instead of sting pass a map with key cluster and subvolumegroupname as the value.

e2e/utils.go
```diff
@@ -1207,3 +1211,43 @@ func addTopologyDomainsToDSYaml(template, labels string) string {
...
+	cm.Namespace = cephCSINamespace
+	// since a configmap is already created, update the existing configmap
+	_, updateErr := c.CoreV1().ConfigMaps(cephCSINamespace).Update(context.TODO(), &cm, metav1.UpdateOptions{})
+	Expect(updateErr).Should(BeNil())
+}
+
+// validateSubvolumegroup validates whether subvolumegroup is present.
+func validateSubvolumegroup(f *framework.Framework, subvolgrp string) error {
+	cmd := fmt.Sprintf("ceph fs subvolumegroup getpath myfs %s", subvolgrp)
+	_, err := execCommandInToolBoxPod(f, cmd, rookNamespace)
```
Madhu-1, line 1247: need to validate the stdout, to make sure the subvolume getpath returns the group path

internal/cephfs/volume.go
```diff
@@ -85,9 +83,21 @@ func getVolumeRootPathCeph(ctx context.Context, volOptions *volumeOptions, cr *u
...
+type additionalClusterInfo struct {
+	resizeSupported       bool
+	subvolumegroupCreated bool
+}
+
 func createVolume(ctx context.Context, volOptions *volumeOptions, cr *util.Credentials, volID volumeID, bytesQuota int64) error {
-	//TODO: When we support multiple fs, need to hande subvolume group create for all fs's
-	if !cephfsInit {
+	// keyPresent checks whether corresponding clusterID key is present in clusterAdditionalInfo
+	var keyPresent bool
```
Madhu-1, line 93: you don't need keyPresent as a separate variable.

internal/cephfs/volume.go
```diff
@@ -85,9 +83,21 @@ func getVolumeRootPathCeph(ctx context.Context, volOptions *volumeOptions, cr *u
 	return strings.TrimSuffix(string(stdout), "\n"), nil
 }
 
+type additionalClusterInfo struct {
+	resizeSupported       bool
+	subvolumegroupCreated bool
```
Madhu-1, line 88: subVolumeGroupCreated

internal/cephfs/volume.go
```diff
@@ -30,14 +30,12 @@ import (
...
 var (
-	// cephfsInit is used to create "csi" subvolume group for the first time the csi plugin loads.
-	// Subvolume group create gets called every time the plugin loads, though it doesn't result in error
-	// its unnecessary
-	cephfsInit = false
-
-	// resizeSupportedList stores the mapping for clusterID and resize is
-	// supported in the particular cluster
-	resizeSupportedList = make(map[string]bool)
+	// checkAdditionalInfo contains information regarding if resize is
```
Madhu-1, line 33: clusterAdditionalInfo contains

e2e/cephfs.go
```diff
@@ -258,6 +258,34 @@ var _ = Describe("cephfs", func() {
...
+					createCephfsStorageClass(f.ClientSet, f, false, "clusterID-1")
+					validatePVCAndAppBinding(pvcPath, appPath, f)
+					deleteResource(cephfsExamplePath + "storageclass.yaml")
+
+					// create resources for the second cluster.
+					createCephfsStorageClass(f.ClientSet, f, false, "clusterID-2")
+					validatePVCAndAppBinding(pvcPath, appPath, f)
+					deleteResource(cephfsExamplePath + "storageclass.yaml")
+
+					// verify multiple subvolumegroup creation.
```
Madhu-1, line 274: move 1st validation to line 267 to keep tests in order

Yuggupta27: Thanks, I'll make the changes.

e2e/utils.go
Yuggupta27: @Madhu-1 Since we will require keys of the map too, will passing only a map and retrieving the keys afterwards increase the complexity? Also, you are right, Passing the map along with strings will remove the hardcoded nature of setting subvolume group name in utils. What should be a better approach?

internal/cephfs/volume.go
Yuggupta27: Currently keypresent variable mainly helps in initializing the struct with default values which might otherwise throw an error. Since keypresent is also used outside it's scope, the variable was created. Should I find a workaround?

e2e/utils.go
Madhu-1: better to pass the map as of now.

internal/cephfs/volume.go
Madhu-1: where it is used outside of loop in createVolume?

Yuggupta27: Here https://github.com/ceph/ceph-csi/pull/1175/files#diff-fa3ab9f6b6a9e75fdc1e69a6ee650d9eR167 ; to try resize for the first run instead of create.

internal/cephfs/volume.go
```diff
@@ -144,7 +152,17 @@ func createVolume(ctx context.Context, volOptions *volumeOptions, cr *util.Crede
...
+	var keyPresent bool
+	// verify if corresponding ClusterID key is present in the map,
+	// and if not, initialize with default values(false).
+	if _, keyPresent = clusterAdditionalInfo[volOptions.ClusterID]; !keyPresent {
+		clusterAdditionalInfo[volOptions.ClusterID] = &additionalClusterInfo{}
+	}
+
+	// resize subvolume when either it's supported, or when corresponding
+	// clusterID key is not present.
+	if supported := clusterAdditionalInfo[volOptions.ClusterID].resizeSupported; supported || !keyPresent {
```
nixpanic, line 165: you can just use `clusterAdditionalInfo[volOptions.ClusterID].resizeSupported`, no need  to assigning it to `supported`

internal/cephfs/volume.go
```diff
@@ -144,7 +152,17 @@ func createVolume(ctx context.Context, volOptions *volumeOptions, cr *util.Crede
...
+	// keyPresent checks whether corresponding clusterID key is present in clusterAdditionalInfo
+	var keyPresent bool
+	// verify if corresponding ClusterID key is present in the map,
+	// and if not, initialize with default values(false).
+	if _, keyPresent = clusterAdditionalInfo[volOptions.ClusterID]; !keyPresent {
+		clusterAdditionalInfo[volOptions.ClusterID] = &additionalClusterInfo{}
+	}
+
+	// resize subvolume when either it's supported, or when corresponding
+	// clusterID key is not present.
```
nixpanic, line 164: `key is not present` -> `key was not present`, you added the key above

internal/cephfs/volume.go
```diff
@@ -85,9 +83,19 @@ func getVolumeRootPathCeph(ctx context.Context, volOptions *volumeOptions, cr *u
...
 func createVolume(ctx context.Context, volOptions *volumeOptions, cr *util.Credentials, volID volumeID, bytesQuota int64) error {
-	//TODO: When we support multiple fs, need to hande subvolume group create for all fs's
-	if !cephfsInit {
+	// verify if corresponding ClusterID key is present in the map,
+	// and if not, initialize with default values(false).
+	if _, keyPresent := clusterAdditionalInfo[volOptions.ClusterID]; !keyPresent {
+		clusterAdditionalInfo[volOptions.ClusterID] = &additionalClusterInfo{}
+	}
+
+	if created := clusterAdditionalInfo[volOptions.ClusterID].subVolumeGroupCreated; !created {
```
nixpanic, line 98: no need to assign to `created`, just use `!clusterAdditionalInfo[volOptions.ClusterID].subVolumeGroupCreated` directly

Yuggupta27: Thanks for the review, I'll make the required changes.

internal/cephfs/volume.go
```diff
@@ -85,9 +83,19 @@ func getVolumeRootPathCeph(ctx context.Context, volOptions *volumeOptions, cr *u
 	return strings.TrimSuffix(string(stdout), "\n"), nil
 }
 
+type additionalClusterInfo struct {
```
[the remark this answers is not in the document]
Yuggupta27: localClusterState seems the most apt. I'll change it.

e2e/utils.go
```diff
@@ -1207,3 +1211,53 @@ func addTopologyDomainsToDSYaml(template, labels string) string {
...
+	conmap := []util.ClusterInfo{
+		{
+			ClusterID: clusterID[0],
+			Monitors:  mons,
+		},
+		{
+			ClusterID: clusterID[1],
+			Monitors:  mons,
+		}}
+	conmap[0].CephFS.SubvolumeGroup = getSubVolGrp["clusterID-1"]
```
Madhu-1, line 1238: is it possible to do like below

```
for i:=0;i<len(getSubVolGrp);i++{
conmap[i].CephFS.SubvolumeGroup = clusterInfo[clusterID[i]]
}
```

e2e/utils.go
```diff
@@ -1207,3 +1211,53 @@ func addTopologyDomainsToDSYaml(template, labels string) string {
 	return strings.ReplaceAll(template, "# - \"--domainlabels=failure-domain/region,failure-domain/zone\"",
 		"- \"--domainlabels="+labels+"\"")
 }
+
+// createCustomConfigMap provides multiple clusters information.
+func createCustomConfigMap(pluginPath string, c kubernetes.Interface, getSubVolGrp map[string]string) {
```
Madhu-1, line 1216: change getSubVolGrp to clusterInfo?

Madhu-1, line 1216:
```suggestion
func createCustomConfigMap(c kubernetes.Interface, pluginPath string,clusterInfo map[string]string) {
```

e2e/utils.go
Yuggupta27: Definitely a better approach... Thanks!

e2e/utils.go
Yuggupta27: clusterInfo exists as a struct and is used in the same program which might add confusion https://github.com/ceph/ceph-csi/blob/master/e2e/utils.go#L345 ... Maybe `subvolgrpInfo` or any other name suggestion?

Madhu-1: Yeah anything will be better with meaningful name

Yuggupta27: Okay, Thanks :)

Issue #1167, opened 2020-06-16
Failed to load cephfs kernel mounter even if kernel>4.17


Madhu-1:
The cephfs plugin container is failed to load the kernel mounter even if the kernel is higher than the 4.17, this issue can be reproducible if we created kubernetes cluster with minikube VM(vm_driver=kvm2)

Logs from cephfsplugin
```text
E0616 05:21:30.439584       1 volumemounter.go:101] failed to parse sublevel from 4.19.107: strconv.Atoi: parsing "107\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00": invalid syntax
I0616 05:21:30.439640       1 volumemounter.go:165] kernel version < 4.17 might not support quota feature, hence not loading kernel client
```

the kernel version on minikube node
```console
$ uname -a
Linux minikube 4.19.107 #1 SMP Mon May 11 14:51:04 PDT 2020 x86_64 GNU/Linux
$ uname -r
4.19.107
```

Note:  we need to discard `0` in bytes  at https://github.com/ceph/ceph-csi/blob/master/internal/cephfs/volumemounter.go#L159

nixpanic:
That looks weird. What version of Golang are you using? I would expect that a `\x00` byte is a marker for the end of a string. Maybe there is a cleaner way to convert a byte-array to a string?

Madhu-1:
> That looks weird. What version of Golang are you using? I would expect that a `\x00` byte is a marker for the end of a string. Maybe there is a cleaner way to convert a byte-array to a string?

go version go1.13.10 linux/amd64. 

```bash
-               release := string(utsname.Release[:64])
 
+               // Remove the \x00 from the release for Atoi to parse correctly
+               release := string(utsname.Release[:bytes.IndexByte(utsname.Release[:], 0)])
+               fmt.Printf("the final release is %v\n", release)
```
the above change should fix the issue, we need to discard the 0's  in the byte array, just need testing and e2e

nixpanic:
[...]

PR #1176, opened 2020-06-18
Make fetching the kernel version a utility function

nixpanic:

It is useful to have the kernel version logged while starting binaries.
Some functionality depends on the version of the kernel, debugging
issues related to this will be easier.

For this, moving the version checking from CephFS into its own function.

It seems that convering the release component from the unix.Utsrelease
type leaves some trailing "\x00" characters.

While splitting the string to compare kernel versions, these additional
characters might prevent converting the string to an int. Strip the
additional characters before returning the string.

Note:
  "\x00" characters are not visible when printing to a file or screen.
  They can be seen in hex-editors, or sending the output through 'xxd'.

## Related issues ##

Fixes: #1167

### DIFF
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -111,6 +111,9 @@ func main() {
 		fmt.Println("Go Version:", runtime.Version())
 		fmt.Println("Compiler:", runtime.Compiler)
 		fmt.Printf("Platform: %s/%s\n", runtime.GOOS, runtime.GOARCH)
+		if kv, err := util.KernelVersion(); err == nil {
+			fmt.Println("Kernel:", kv)
+		}
 		os.Exit(0)
 	}
 
--- a/internal/cephfs/volumemounter.go
+++ b/internal/cephfs/volumemounter.go
@@ -29,7 +29,6 @@ import (
 
 	"github.com/ceph/ceph-csi/internal/util"
 
-	"golang.org/x/sys/unix"
 	"k8s.io/klog"
 )
 
@@ -151,12 +150,10 @@ func loadAvailableMounters(conf *util.Config) error {
 		klog.Errorf("failed to run mount.ceph %v", err)
 	} else {
 		// fetch the current running kernel info
-		utsname := unix.Utsname{}
-		err = unix.Uname(&utsname)
-		if err != nil {
-			return err
+		release, kvErr := util.KernelVersion()
+		if kvErr != nil {
+			return kvErr
 		}
-		release := string(utsname.Release[:64])
 
 		if conf.ForceKernelCephFS || kernelSupportsQuota(release) {
 			klog.V(1).Infof("loaded mounter: %s", volumeMounterKernel)
--- a/internal/util/util_test.go
+++ b/internal/util/util_test.go
@@ -142,3 +142,13 @@ func TestRoundOffVolSize(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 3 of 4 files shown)


cmd/cephcsi.go
```diff
@@ -111,6 +111,8 @@ func main() {
 		fmt.Println("Go Version:", runtime.Version())
 		fmt.Println("Compiler:", runtime.Compiler)
 		fmt.Printf("Platform: %s/%s\n", runtime.GOOS, runtime.GOARCH)
+		kv, _ := util.KernelVersion()
```
Madhu-1, line 114: Error is discarded here.

internal/cephfs/volumemounter.go
```diff
@@ -151,12 +150,10 @@ func loadAvailableMounters(conf *util.Config) error {
 		klog.Errorf("failed to run mount.ceph %v", err)
 	} else {
 		// fetch the current running kernel info
-		utsname := unix.Utsname{}
-		err = unix.Uname(&utsname)
-		if err != nil {
+		release, err := util.KernelVersion()
+		if err == nil {
```
Madhu-1, line 154: it should be err!=nil isnt it?

internal/util/util_test.go
```diff
@@ -142,3 +143,16 @@ func TestRoundOffVolSize(t *testing.T) {
...
 	}
 }
+
+func TestKernelVersion(t *testing.T) {
+	version, err := KernelVersion()
+	if err != nil {
+		t.Errorf("failed to get kernel version: %s", err)
+	}
+	if version == "" {
+		t.Errorf("version is empty, this is unexpected?!")
```
Madhu-1, line 153: use Error() as formatting is not required

internal/util/util_test.go
```diff
@@ -142,3 +143,16 @@ func TestRoundOffVolSize(t *testing.T) {
...
+	}
+	if strings.HasSuffix(version, "\x00") {
+		t.Errorf("version ends with \\x00 byte(s)")
```
Madhu-1, line 156: same here also

cmd/cephcsi.go
nixpanic: yes, would you prefer to log the error in case the kernel version can not be obtained?
I am considering to have the function never fail, and just return an empty string in case the kernel function can not be found (which should be extremely unlikely)

Madhu-1: i would prefer to log kernel version if err==nil instead of logging empty value

internal/cephfs/volumemounter.go
nixpanic: well, yes, of course!?

cmd/cephcsi.go
nixpanic: ok, only logging the kernel version if it could get fetched

nixpanic:
@mergifyio rebase

Madhu-1:
>internal/cephfs/volumemounter.go:153:12: shadow: declaration of "err" shadows declaration at line 148 (govet)
release, err := util.KernelVersion()

@nixpanic CI failure can you please fix it.

nixpanic:
> > internal/cephfs/volumemounter.go:153:12: shadow: declaration of "err" shadows declaration at line 148 (govet)
> > release, err := util.KernelVersion()
> 
> @nixpanic CI failure can you please fix it.

Done. Please review again.

PR #1178, opened 2020-06-18
rbd: use librbd.FeatureSet for features

nixpanic:

go-ceph v0.3 adds constants for ImageFeature values and their names.
Instead of hardcoding "layering" in several places, use the constant
given by librbd.

The rbdVolume.ImageFeatures does not seem to be used anywhere after the
conversion. Stashing the image metadata does include the ImageFeatures
as these are retrieved when getting the image information. It is safe to
drop ImageFeatures altogether and only use the imageFeatureSet instead.

### DIFF
--- a/internal/rbd/rbd_util_test.go
+++ b/internal/rbd/rbd_util_test.go
@@ -17,7 +17,10 @@ limitations under the License.
 package rbd
 
 import (
+	"strings"
 	"testing"
+
+	librbd "github.com/ceph/go-ceph/rbd"
 )
 
 func TestIsLegacyVolumeID(t *testing.T) {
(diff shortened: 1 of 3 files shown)


nixpanic:
@phlogistonjohn some go-ceph changes for you to review :grin:

internal/rbd/rbd_util_test.go
```diff
@@ -18,6 +18,9 @@ package rbd
 
 import (
 	"testing"
+	"strings"
```
phlogistonjohn, line 21: The linter is upset by this ordering, I think.

nixpanic: Indeed, I would expect that too

humblec:
> Nice reduction in code. :-)

Indeed, we need more :)

Madhu-1:
@humblec PTAL

Issue #1180, opened 2020-06-19
Use a single file with versions and parameters for building, CI and deployment


nixpanic:
Currently versions of different components and tools are listed in several places. Including `Makefile`, `.travis.yml` and several of the `Dockerfile`s. When updating a version of a tool or component, the change needs to get done everywhere.

Instead, all versions should be stored in `build.env` (or other filename) that can be consumed everywhere.

humblec:
This is much required one! the current way of env handling is a mess.

PR #1182, opened 2020-06-22
Introduce build.env for single-point of configuration of environment variables

nixpanic:

The new `build.env` file contains environment variables that are used in several locations. By placing these in a single file, it will be easier to test with newer versions of tools and upgrade everything accordingly.

The following versions/settings have been added to `build.env`:

- [x] Golang version
- [x] Golang CI version
- [x] Gosec version
- [x] Helm version
- [x] minikube variables
- [x] `BASE_IMAGE` for the `FROM` in container images
- [x] `CEPH_VERSION` like "nautilus"
- [x] e2e configuration
- [x] build check for arm64/amd64 container images

## Related issues ##

Fixes: #1180

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,10 @@ before_install:
   - mkdir -p $GOPATH/bin
 
 before_script:
+  - export CV=$(source build.env ; echo ${CEPH_VERSION})
   - curl https://download.ceph.com/keys/release.asc | sudo apt-key add -
   - sudo apt-add-repository
-    "deb https://download.ceph.com/debian-nautilus $(lsb_release -sc) main"
+    "deb https://download.ceph.com/debian-${CV} $(lsb_release -sc) main"
   # yamllint enable rule:line-length
   - sudo apt-get -qq update
   # only the arm64 fallback repo is unsigned and needs --allow-unauthenticated
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ env:
     - TEST_COVERAGE=stdout
     - GO_METALINTER_THREADS=1
     - GO_COVER_DIR=_output
-    - HELM_VERSION=v3.1.2
     - VM_DRIVER=none
     - MINIKUBE_VERSION=v1.6.0
     - CHANGE_MINIKUBE_NONE_USER=true
@@ -74,6 +73,7 @@ jobs:
           "https://raw.githubusercontent.com/securego/gosec/master/install.sh"
           | sh -s -- -b $GOPATH/bin "${GOSEC_VERSION}"
         # install helm for helm lint
+        - export HELM_VERSION=$(source build.env ; echo ${HELM_VERSION})
         - curl -L https://git.io/get_helm.sh
           | bash -s -- -v "${HELM_VERSION}"
         # yamllint disable rule:line-length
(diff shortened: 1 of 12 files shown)


nixpanic:
@Madhu-1 @humblec PTAL, this is now ready for review.

build.env
```diff
@@ -0,0 +1,40 @@
...
+# helm chart generation, testing and publishing
+HELM_VERSION=v3.1.2
+
+# minikube settings
+MINIKUBE_VERSION=v1.6.0
+VM_DRIVER=none
+CHANGE_MINIKUBE_NONE_USER=true
+
+# e2e settings
+# - enable CEPH_CSI_RUN_ALL_TESTS when running tests with if it has root
```
Madhu-1, line 38:
```suggestion
#  enable CEPH_CSI_RUN_ALL_TESTS when running tests with if it has the root
#   permissions on the host
```

deploy/cephcsi/image/Dockerfile
```diff
@@ -1,22 +1,27 @@
...
 ARG GOROOT=/usr/local/go
 
-RUN mkdir -p ${GOROOT} && \
+COPY build.env /
+
+RUN source /build.env && \
+    mkdir -p ${GOROOT} && \
     curl https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-${GO_ARCH}.tar.gz | tar xzf - -C ${GOROOT} --strip-components=1
 
+# test if the downloaded version of Golang works (different arch?)
```
Madhu-1, line 22: is it required?

nixpanic: It is just an additional check. While working on this PR, I hit an issue where the amd64 version of Golang was installed in the arm64 image. Errors were pretty unclear, this check causes it to abort early.

build.env
nixpanic: I'll phrase it like
```
# - enable CEPH_CSI_RUN_ALL_TESTS when running tests has root
#   permissions on the host (needed for cgroup modifications)
```
The prefix of ` - ` is intentional, there could be other `CEPH_CSI_` variables that affect the test suite (I would like to be able to only test rbd or cephfs). The list format will make it easier for others to add more variables.

Is that ok?

humblec:
@nixpanic typo here `condifguration` in the title of the PR.

.travis.yml
```diff
@@ -65,14 +58,17 @@ jobs:
         - pip install --user --upgrade pip
         - pip install --user yamllint
         # install golangci-lint
+        - export GOLANGCI_VERSION=$(source build.env ; echo ${GOLANGCI_VERSION})
```
humblec, line 61: @nixpanic cant we source a file in a single place here instead of differently doing it in each iterations?

humblec:
overall it looks good to me. :+1:

.travis.yml
nixpanic: I was hoping so, but I could not find how to do that nicely in travis. After sourcing, the variables need to be exported so that child processes of the shell can consume them.

Listing them explicitly also makes it easier to find problems.

nixpanic:
> @nixpanic typo here `condifguration` in the title of the PR.

Fixed, please let me know if there is something else that needs changing.

Issue #1132, opened 2020-06-04
RUN pylint as part of CI process


Madhu-1:
We have one python file in the project to trace the pvc-backend volumes. we need to run static check on the file to catch the static issues as part of CI

Yuggupta27:
Please assign @Madhu-1

PR #1186, opened 2020-06-25
ci: add pylint as part of CI

Yuggupta27:
add pylint to catch static issues of python
files in the repo.

User can now run `make lint-py` for pylint
check on python files.

Fixes: #1132 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,8 +74,12 @@ jobs:
         # yamllint disable rule:line-length
         - mkdir -p /opt/commitlint && pushd /opt/commitlint
           && npm init -y && npm install --save-dev @commitlint/travis-cli && popd
-        # yamllint enable rule:line-length
         - export PATH=/opt/commitlint/node_modules/.bin:$PATH
+        # install pylint and prettytable.
+        # target python script requires python3 dependencies.
+        - sudo apt install python3-pip python3-prettytable python3-setuptools python3-wheel -y
+        - pip3 install pylint
+        # yamllint enable rule:line-length
       script:
         - commitlint-travis
         - make go-lint
--- a/scripts/Dockerfile.test
+++ b/scripts/Dockerfile.test
@@ -31,6 +31,8 @@ RUN source /build.env \
 	ShellCheck \
 	yamllint \
 	npm \
+      python3-prettytable \
+      pylint \
     && dnf -y update \
     && dnf -y clean all \
     && gem install mdl \
--- a/scripts/lint-extras.sh
+++ b/scripts/lint-extras.sh
@@ -59,6 +59,11 @@ function lint_helm() {
     run_check '' helm lint --namespace=test charts/*
 }
 
+function lint_py() {
+    # Install via: sudo apt-get install python3-pylint
+    run_check '.*\.py' pylint --score n --output-format=colorized
+}
+
 function lint_all() {
     # runs all checks
     all_required=1
@@ -66,6 +71,7 @@ function lint_all() {
     lint_yaml
     lint_markdown
     lint_helm
+    lint_py
 }
 case "${1:-}" in
 lint-shell)
@@ -80,6 +86,9 @@ lint-markdown)
 lint-helm)
     lint_helm
     ;;
+lint-py)
+    lint_py
+    ;;
 lint-all)
     lint_all
     ;;
@@ -90,6 +99,7 @@ Available Commands:
   lint-yaml              Lint yaml files
   lint-markdown          Lint markdown files
   lint-helm              Lint helm charts
+  lint-py                Lint python files
   lint-all               Run lint on all non-go files
 " >&2
     ;;
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,9 @@ lint-yaml:
 lint-helm:
 	./scripts/lint-extras.sh lint-helm
 
+lint-py:
+	./scripts/lint-extras.sh lint-py
+
 gosec:
 	./scripts/gosec.sh
 


.travis.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         - gem install mdl
         - pip install --user --upgrade pip
         - pip install --user yamllint
+        - pip3 install --user pylint
```
nixpanic, line 67: you probably should not mix `pip` and `pip3` commands, just always use `pip`

scripts/Dockerfile.test
```diff
@@ -30,6 +30,7 @@ RUN dnf -y install \
 	ShellCheck \
 	yamllint \
 	npm \
+   pylint \
```
nixpanic, line 33: make sure to indent correctly

scripts/lint-extras.sh
```diff
@@ -59,13 +59,19 @@ function lint_helm() {
     run_check '' helm lint --namespace=test charts/*
 }
 
+function lint_py() {
+    # Install via: pip install pylint
+    run_check '.*\.py' pylint -s n -f colorized
```
nixpanic, line 64: Ideally run commands with long options, like `--score=y` and so on. People reading the short options will not know what they stand for, long options are a little more user-friendly.

Yuggupta27: I'll keep that in mind. Thanks!

nixpanic:
in Travis, `pylint` worked and returned the following issues:
```
pylint --score n --output-format=colorized ./troubleshooting/tools/tracevol.py 
No config file found, using default configuration
************* Module tracevol
C:125, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
C:126, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
C:234, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
C:506, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
C:509, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
R:175, 0: Too many local variables (16/15) (too-many-locals)
R:238, 0: Too many local variables (16/15) (too-many-locals)
```
In the Fedora container-image, pretty-table is missing. That should normally get pulled in by RPM dependencies. Check if there is no bug against pylint in Fedora yet, and report it otherwise.

It should not be needed to install pretty-table separately in Travis CI either...

Madhu-1:
> in Travis, `pylint` worked and returned the following issues:
> 
> ```
> pylint --score n --output-format=colorized ./troubleshooting/tools/tracevol.py 
> No config file found, using default configuration
> ************* Module tracevol
> C:125, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
> C:126, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
> C:234, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
> C:506, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
> C:509, 0: Unnecessary parens after 'print' keyword (superfluous-parens)
> R:175, 0: Too many local variables (16/15) (too-many-locals)
> R:238, 0: Too many local variables (16/15) (too-many-locals)
> ```
> 
> In the Fedora container-image, pretty-table is missing. That should normally get pulled in by RPM dependencies. Check if there is no bug against pylint in Fedora yet, and report it otherwise.
> 
> It should not be needed to install pretty-table separately in Travis CI either...

I don't see any errors locally, make sure the pylint is compatible with python3

.travis.yml
```diff
@@ -64,6 +64,8 @@ jobs:
         - gem install mdl
         - pip install --user --upgrade pip
         - pip install --user yamllint
+        - pip install --user pylint
```
nixpanic, line 67: it seems that on Ubuntu you can install this with `apt-get` instead, that is cleaner in most cases:
```
apt-get install python3-pylint
```
and that would also get you the python3 version (no more warnings?)

nixpanic:
Ah, and the missing prettytable is a dependency of the python script that is tested, not of `pylint` itself. In that case, you need to install `python3-prettytable` too, and explain why it is needed.

scripts/Dockerfile.test
```diff
@@ -41,6 +42,7 @@ RUN dnf -y install \
     && mkdir /opt/commitlint && pushd /opt/commitlint \
     && npm init -y \
     && npm install @commitlint/cli \
+    && pip install prettytable \
```
nixpanic, line 45: please do not use `pip` but `dnf install python3-prettytable`

Yuggupta27:
> Ah, and the missing prettytable is a dependency of the python script that is tested, not of `pylint` itself. In that case, you need to install `python3-prettytable` too, and explain why it is needed.

Yes, Locally also I had to install prettytable to resolve the issue. I need to install it too.

scripts/Dockerfile.test
```diff
@@ -41,6 +42,7 @@ RUN dnf -y install \
     && mkdir /opt/commitlint && pushd /opt/commitlint \
     && npm init -y \
     && npm install @commitlint/cli \
+    && dnf install python3-prettytable \
```
nixpanic, line 45: better add this to the list where pylint is installed too. This will cause the package metadata to be downloaded again (because it is after `dnf clean all` above)

Yuggupta27: Oh sorry, I missed that.

.travis.yml
```diff
@@ -80,6 +80,8 @@ jobs:
           && npm init -y && npm install --save-dev @commitlint/travis-cli && popd
         # yamllint enable rule:line-length
         - export PATH=/opt/commitlint/node_modules/.bin:$PATH
+        # install pylint for python files
+        - apt-get install python3-pylint
```
nixpanic, line 84: probably need to add python3-prettytable here as well

nixpanic: It also seems this needs `sudo`...

nixpanic:
The `Dockerfile.test` looks good to me, just need to correct the `.travis.yml`

.travis.yml
Yuggupta27: Yes, there was a permission issue; apt doesn't directly support non-root user.

scripts/lint-extras.sh
```diff
@@ -59,13 +59,19 @@ function lint_helm() {
     run_check '' helm lint --namespace=test charts/*
 }
 
+function lint_py() {
+    # Install via: pip install pylint
```
Madhu-1, line 63: the comment need to be removed.

.travis.yml
```diff
@@ -80,6 +80,9 @@ jobs:
           && npm init -y && npm install --save-dev @commitlint/travis-cli && popd
         # yamllint enable rule:line-length
         - export PATH=/opt/commitlint/node_modules/.bin:$PATH
+        # install pylint for python files and prettytable dependency.
```
Madhu-1, line 83:
```suggestion
        - sudo apt-get install python3-pylint python3-prettytable -y
```

.travis.yml
```diff
@@ -80,6 +80,9 @@ jobs:
           && npm init -y && npm install --save-dev @commitlint/travis-cli && popd
         # yamllint enable rule:line-length
         - export PATH=/opt/commitlint/node_modules/.bin:$PATH
+        # install pylint and prettytable.
+        - sudo apt-get install python3-prettytable -y
+        - sudo apt install pylint3
```
nixpanic, line 85: installing the packages with the same command is normally a little faster. If that does not work for some reason, there needs to be a comment about it

scripts/lint-extras.sh
```diff
@@ -59,13 +59,19 @@ function lint_helm() {
     run_check '' helm lint --namespace=test charts/*
 }
 
+function lint_py() {
+    # Install via: sudo apt-get install python3-pylint
+    run_check '.*\.py' pylint --score n --output-format=colorized
```
nixpanic, line 64: This may need to be `pylint3` on Ubuntu?

Issue #1184, opened 2020-06-25
Update golangci to v1.27.0 (or current latest version)


nixpanic:
We are testing with golangci v1.21.0, while v1.27.0 is [currently the latest version](https://github.com/golangci/golangci-lint/tags).

Several new linters have been added since. Our codebase is not following best practices, so the new golangci linters report errors.

This issue is a tracker for how to run with the new linter version, and which PRs address code cleanups (or linter configuration tuning) to move towards making the new version the default for the project.

## Usage

1. edit `build.env` (or `scripts/Dockerfile.test` in case #1182 is not merged yet)
1. replace `GOLANGCI_VERSION=v1.21.0` with `GOLANGCI_VERSION=v1.27.0`
1. run `make containerized-test TARGET=go-lint`
1. pick a series of warnings that you want to address, copy/paste the warning somewhere (you'll need it later)
1. address the warning in the source file
1. run `make containerized-test TARGET=go-lint` again to verify that the warning is fixed an no new ones introduced
1. commit the source changes only (not `build.env` or `scripts/Dockerfile.test`), use a commit message like
    ```
    cleanup: address <linter/warning> in <function/file>
    
    kubernetes/go-client was imported twice in the <file>. Make sure it is only imported once.
    The DupImport linter reported this issue as:
    
    <paste the copied original warning, don't add additional linebreaks, lines can be long>

    Signed-off-by: Your Real Name <email@example.net>
    ```
1. push the commit to your GitHub fork of the project
1. create a new PR and mention `Updates: #1184` so that this issue gets linked

ansiwen:
I have to run the test with `CPUSET= make containerized-test TARGET=go-lint`, because I use root-less podman. Is this a known issue?

PR #1185, opened 2020-06-25
ci: disable 'testpackage' linter in golangci

nixpanic:
# Describe what this PR does #

The 'testpackage' linter recommends "black box" testing, which prevents
testing internal/non-exported functions from being tested. We have tests
that *do* test non-exported functions and types.

Disabling the linter allows us to test non-exported types and functions.

## Related issues ##

Updates: #1184

PR #1187, opened 2020-06-26
cleanup: [WIP] address golangci warnings

Yuggupta27:
Package name should be util_test instead of
util. util package needs to be imported
seperately.

Issue Reported:
internal/util/csiconfig_test.go:17:9: package should be `util_test` instead of `util` (testpackage)
package util

Updates: #1184 

### DIFF
--- a/build.env
+++ b/build.env
@@ -18,7 +18,7 @@ GOLANG_VERSION=1.13.9
 GO111MODULE=on
 
 # static checks and linters
-GOLANGCI_VERSION=v1.21.0
+GOLANGCI_VERSION=v1.27.0
 GOSEC_VERSION=v2.3.0
 
 # "go test" configuration
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -40,6 +40,9 @@ const (
 	rbdDefaultName      = "rbd.csi.ceph.com"
 	cephfsDefaultName   = "cephfs.csi.ceph.com"
 	livenessDefaultName = "liveness.csi.ceph.com"
+
+	pollTime     = 60 // seconds
+	probeTimeout = 3  // seconds
 )
 
 var (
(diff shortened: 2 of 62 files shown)


nixpanic:
Note that we can not change the package name for all packages. See #1185 for a description. As an alternative to that PR, it is also possible to mark a `package util` file with `// nolint:testpackage // we're testing internals of the package`. More tests are valuable, and dropping them or reworking them in awkward ways would not be beneficial.

Yuggupta27:
> Note that we can not change the package name for all packages. See #1185 for a description. As an alternative to that PR, it is also possible to mark a `package util` file with `// nolint:testpackage // we're testing internals of the package`. More tests are valuable, and dropping them or reworking them in awkward ways would not be beneficial.
Oh my bad, I tried to directly resolve the warnings. Should we then no lint on all test packages?

nixpanic:
> Should we then no lint on all test packages?

If there are `*_test.go` files where it works without too much re-working, then you can rename the package. For tests that use internals of the package, you can add the `// nolint:....` comment. I opted for disabling the `testpackage` linter completely, but that might not be the best approach (I did not try anything else).

Yuggupta27:
> > Should we then no lint on all test packages?
> 
> If there are `*_test.go` files where it works without too much re-working, then you can rename the package. For tests that use internals of the package, you can add the `// nolint:....` comment. I opted for disabling the `testpackage` linter completely, but that might not be the best approach (I did not try anything else).

Thanks, I'll try to rectify the issues wherever feasible, and nolint otherwise.

Yuggupta27:
@humblec , these commits can be used now to reduce some of the warnings.

humblec:
@Yuggupta27 this can be closed right ?

Yuggupta27:
> @Yuggupta27 this can be closed right ?

Yes Humble. I'll close it.

Yuggupta27:
Closing as changes are included in #1157 .

PR #1189, opened 2020-06-26
Use errors.As() to compare errors

nixpanic:
Golang 1.13 improves the way errors can be checked against certain types.

- `fmt.Errorf()` adds a `%w` formatting string to create wrapped errors
- `errors.As()` can (try to) convert errors of one type, to an other

With these adaptions, there is no need to use the `"github.com/pkg/errors"` package anymore. The standard `"errors"` package is now sufficient.

See-also: https://github.com/golang/go/wiki/ErrorValueFAQ#how-should-i-change-my-error-handling-code-to-work-with-the-new-features


nixpanic:
@Yuggupta27 you may want to review this. There are some `fmt.Errorf()` places that need the `%w` formatting to create wrapped errors. You've probably noticed the warnings while working on #1187

internal/rbd/controllerserver.go
```diff
@@ -304,11 +306,13 @@ func (cs *ControllerServer) checkSnapshotSource(ctx context.Context, req *csi.Cr
 
 	rbdSnap := &rbdSnapshot{}
 	if err := genSnapFromSnapID(ctx, rbdSnap, snapshotID, cr); err != nil {
-		if _, ok := err.(ErrSnapNotFound); !ok {
+		var esnf *ErrSnapNotFound
+		if errors.As(err, &esnf) {
```
nixpanic, line 310: ah, inverted the check here... CI might fail because of that.

internal/rbd/controllerserver.go
```diff
@@ -696,20 +707,23 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
...
-		if _, ok := err.(util.ErrKeyNotFound); ok {
+		var eknf *util.ErrKeyNotFound
+		if errors.As(err, &eknf) {
 			return &csi.DeleteSnapshotResponse{}, nil
 		}
 
 		// All errors other than ErrSnapNotFound should return an error back to the caller
-		if _, ok := err.(ErrSnapNotFound); !ok {
+		var esnf *ErrSnapNotFound
+		if errors.As(err, &esnf) {
```
nixpanic, line 726: Same here, inverted the check :disappointed:

nixpanic:
I do not spot why this fails by reviewing the change. Starting to run `git bisect` and pushing changes to this PR. Expect to see some updates with fewer commits :)

Yuggupta27:
> I do not spot why this fails by reviewing the change. Starting to run `git bisect` and pushing changes to this PR. Expect to see some updates with fewer commits :)

Nothing specific seems to catch my eye either, I'll continue to look at it and see if I can find the issue.

nixpanic:
@mergifyio rebase

nixpanic:
Due to the GitHub outages this morning, it seems that the PR is out of sync with the `cleanup/errors.As` branch :roll_eyes:

humblec:
> Due to the GitHub outages this morning, it seems that the PR is out of sync with the `cleanup/errors.As` branch roll_eyes

Thats interesting :).. Is this uptodate now ?

nixpanic:
> Nothing specific seems to catch my eye either, I'll continue to look at it and see if I can find the issue.

I think I found the problem. Our own errors need to implement the `Unwrap()` method too. This has been added now, so hopefully e2e passes again :crossed_fingers:

internal/cephfs/nodeserver.go
```diff
@@ -87,14 +88,16 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 
 	volOptions, _, err := newVolumeOptionsFromVolID(ctx, string(volID), req.GetVolumeContext(), req.GetSecrets())
 	if err != nil {
```
nixpanic, line 90: CI fails here:
``
rpc error: code = Internal desc = error decoding volume ID (failed to decode CSI identifier, string underflow) (pv-name)
```
`err` should be a `ErrInvalidVolID` error, it is not clear to me why `errors.As()` fails

File: internal/rbd/controllerserver.go [hunk 4]
```diff
@@ -786,13 +799,14 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 
 	rbdVol, err := genVolFromVolID(ctx, volID, cr, req.GetSecrets())
 	if err != nil {
-		switch err.(type) {
-		case ErrImageNotFound:
+		var einf *ErrImageNotFound
```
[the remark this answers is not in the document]
nixpanic: I agree that a switch is nicer, but [the docs](https://github.com/golang/go/wiki/ErrorValueFAQ#how-should-i-change-my-error-handling-code-to-work-with-the-new-features) mention:

> Rewrite a type switch as a sequence of if-elses.

internal/rbd/rbd_journal.go
```diff
@@ -216,7 +217,8 @@ func (rv *rbdVolume) Exists(ctx context.Context) (bool, error) {
 	// Fetch on-disk image attributes and compare against request
 	err = rv.getImageInfo()
 	if err != nil {
-		if _, ok := err.(ErrImageNotFound); ok {
+		var einf *ErrImageNotFound
```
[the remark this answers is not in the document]
nixpanic: ErrImageNotFound is currently a type, do you suggest to make it a variable instead?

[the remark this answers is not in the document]
Madhu-1: is it possible to have a new function returns this struct pointer?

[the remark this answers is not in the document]
nixpanic: That is what `errors.As()` does? Because errors can have several wrapped errors, just type-casting to the 'upper' error may not be sufficient. Internally `errors.As()` uses the `(err *error) Unwrap() error` method to go  through the list of wrapped errors.

internal/journal/voljournal.go
```diff
@@ -281,13 +281,14 @@ func (conn *Connection) CheckReservation(ctx context.Context,
...
-	case nil:
-	case util.ErrKeyNotFound, util.ErrPoolNotFound:
-		// pool or omap (oid) was not present
-		// stop processing but without an error for no reservation exists
-		return nil, nil
-	default:
+	if err != nil {
+		var eknf *util.ErrKeyNotFound
+		var epnf *util.ErrPoolNotFound
+		if errors.As(err, &eknf) || errors.As(err, &epnf) {
```
nixpanic, line 287: A failure while creating an RBD volume:
```
omap.go:66] ID: 17 Req-ID: pvc-bd7ebc0d-3d40-4a63-8094-09e30d28466a omap not found (pool="replicapool", namespace="", name="csi.volumes.default"): rados: ret=2, No such file or directory
```
This suggests the return of `util.NewErrKeyNotFound()` from `getOMapValues()` does not work with `errors.As()`.

I must be missing something, https://github.com/golang/go/wiki/ErrorValueFAQ is pretty clear :thinking:

vendor/modules.txt
```diff
@@ -124,7 +124,7 @@ github.com/onsi/gomega/types
 github.com/opencontainers/go-digest
 # github.com/pborman/uuid v1.2.0
 github.com/pborman/uuid
-# github.com/pkg/errors v0.9.1
+# github.com/pkg/errors v0.8.1
```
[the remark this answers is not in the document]
nixpanic: not sure, there seems to be an indirect dependency that prefers v0.8.1. Ceph-CSI does not import this package directly anymore.

internal/journal/voljournal.go
nixpanic: found it :partying_face: The example in the [wiki](https://github.com/golang/go/wiki/ErrorValueFAQ#how-should-i-change-my-error-handling-code-to-work-with-the-new-features) shows a `var e *ErrSomeFailure`, but in most cases it should be `var e ErrSomeFailure` instead.

nixpanic:
I think all failures have been addressed now. Please review again, @Madhu-1 @humblec

vendor/modules.txt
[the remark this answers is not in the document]
Madhu-1: we do use it directly in cephcsi 
```bash
grep -ri "github.com/pkg/errors" . |grep -v vendor
./internal/journal/voljournal.go:	"github.com/pkg/errors"
./internal/rbd/rbd_util.go:	"github.com/pkg/errors"
./internal/rbd/snapshot.go:	"github.com/pkg/errors"
./internal/rbd/rbd_journal.go:	"github.com/pkg/errors"
./internal/util/util.go:	"github.com/pkg/errors"
./internal/util/k8scmcache.go:	"github.com/pkg/errors"
./internal/util/nodecache.go:	"github.com/pkg/errors"
./internal/util/crypto.go:	"github.com/pkg/errors"
./internal/util/connection.go:	"github.com/pkg/errors"
./internal/util/conn_pool.go:	"github.com/pkg/errors"
./internal/cephfs/volumeoptions.go:	"github.com/pkg/errors"
./go.mod:	github.com/pkg/errors v0.9.1
```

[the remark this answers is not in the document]
Madhu-1: okay this PR removes this import.

internal/cephfs/errors.go
```diff
@@ -44,3 +52,7 @@ type ErrVolumeNotFound struct {
 func (e ErrVolumeNotFound) Error() string {
 	return e.err.Error()
 }
+
+func (e ErrVolumeNotFound) Unwrap() error {
```
Madhu-1, line 56: Need to add a comment for exported functions

internal/journal/voljournal.go
```diff
@@ -608,12 +613,14 @@ func (conn *Connection) GetImageAttributes(ctx context.Context, pool, objectUUID
...
-	switch err.(type) {
-	case nil:
-	case util.ErrPoolNotFound, util.ErrKeyNotFound:
-		klog.Warningf(util.Log(ctx, "unable to read omap keys: pool or key missing: %v"), err)
-	default:
-		return nil, err
+	if err != nil {
+		var eknf util.ErrKeyNotFound
+		var epnf util.ErrPoolNotFound
+		if errors.As(err, &eknf) || errors.As(err, &epnf) {
```
Madhu-1, line 619: can we remove else block if we change this to
```go
if !errors.As(err, &eknf) && !errors.As(err, &epnf) {
return nil error
}
 klog.Warningf(util.Log(ctx, "unable to read omap keys: pool or key missing: %v"), err)
```

internal/cephfs/errors.go
nixpanic: yeah, but it was not done for others, so I skipped it here. Will add comments for the other ones as well.

internal/journal/voljournal.go
nixpanic: sure

nixpanic:
@Madhu-1, @humblec, @Yuggupta27 this is ready for review, please take a look

PR #1191, opened 2020-06-30
rebase: use gosec-2.3.0

nixpanic:
gosec 2.3.0 has been released a while ago and adds more checks. With a few modifications ceph-csi passes the new additions. With the newer version of gosec, the code will be checked for more security issues than with the previous 2.0.0 version.

See-also: https://github.com/securego/gosec/releases/tag/v2.3.0

### DIFF
--- a/internal/util/cephcmds.go
+++ b/internal/util/cephcmds.go
@@ -142,7 +142,12 @@ func GetOMapValue(ctx context.Context, monitors string, cr *Credentials, poolNam
 		klog.Errorf(Log(ctx, "failed creating a temporary file for key contents"))
 		return "", err
 	}
-	defer tmpFile.Close()
+	defer func() {
+		ce := tmpFile.Close()
+		if ce != nil {
+			klog.Warningf(Log(ctx, "failed closing temporary file: %s"), ce)
+		}
+	}()
 	defer os.Remove(tmpFile.Name())
 
 	args := []string{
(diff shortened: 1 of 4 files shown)


internal/util/cephcmds.go
```diff
@@ -142,7 +142,11 @@ func GetOMapValue(ctx context.Context, monitors string, cr *Credentials, poolNam
 		klog.Errorf(Log(ctx, "failed creating a temporary file for key contents"))
 		return "", err
 	}
-	defer tmpFile.Close()
+	defer func() {
+		if tmpFile.Close() != nil {
+			klog.Warningf(Log(ctx, "failed closing temporary file"))
```
Madhu-1, line 147: log the error

internal/util/pidlimit.go
```diff
@@ -41,7 +41,7 @@ func getCgroupPidsFile() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer cgroup.Close()
+	defer cgroup.Close() // #nosec: error on close is not critical here
```
Madhu-1, line 44: how about handle log error in defer function here and in other places  as we dont need to handle nosec

nixpanic: I did not do so, because we do not care about the error here. The function worked with closing the file returning an error.

internal/util/cephcmds.go
nixpanic: sure, I can do that

PR #1192, opened 2020-06-30
Cleanup messages related to Travis CI configuration

nixpanic:
When browsing to the Travis CI build, each build has a "view config" tab. This tab lists several messages that can be addressed. Instead of the warning orange dot next to the tab, things should now become green.


nixpanic:
@mergifyio refresh

Issue #1188, opened 2020-06-26
Tracker to fix todo in #1160 snapshot PR


Madhu-1:
This issue is for tracking the TODO which needs to be fixed after merging #1160, this will be closed after fixing all the todo's
for easier tracking am adding the TODO list here from the PR

- [ ] Create PVC from PVC
- [ ]  Fix issues with snapshot/clone (512) limits #1098
- [ ]  Allow Parallel cloning and snapshots creation on one PVC/clone (Need to work on in-memory locking)
- [ ]  Improve/Add Documentation for snapshot and cloning
- [ ]  Check Anything Needed to support datapool
- [ ]  Add doc file for the commands
- [ ]  If required some code cleanup
- [ ]  Replace GetStoredImageID with GetImageAttributes in all places

Madhu-1:
> Add a list of kernel versions that supports the deep-flatten feature

@dillaman @idryomov can you please help us to know the kernel version which supports the deep-flatten features.

idryomov:
deep-flatten is supported starting with upstream 5.1 and RHEL 8.2.

Madhu-1:
> deep-flatten is supported starting with upstream 5.1 and RHEL 8.2.

any exact version of RHEL 8.2? like `4.18.0-193` because we need to flatten the image if the kernel version is less than the required version.

idryomov:
Yes, use `4.18.0-193`.  Technically it is in `4.18.0-154`, but going by released versions is better.

Madhu-1:
@idryomov So we can flatten the rbd images before use if the kernel version is less than 4.18.0-154 ?

idryomov:
I think it's better to use `4.18.0-193` as the cutoff version.

Madhu-1:
> I think it's better to use `4.18.0-193` as the cutoff version.

Okay will consider it, Thanks @idryomov

Madhu-1:
> This issue is for tracking the TODO which needs to be fixed after merging #1160, this will be closed after fixing all the todo's
> for easier tracking am adding the TODO list here from the PR
> 
> * [x]  Create PVC from PVC
> * [x]   Fix issues with snapshot/clone (510) limits #1098
> * [x]  Add list of kernel versions that supports deep-flatten feature
> * [ ]   Allow Parallel cloning and snapshots creation on one PVC/clone (Need to work on in-memory locking)
> * [x]   Check Anything Needed to support datapool
> * [x]   Add doc file for the commands
[...]

PR #1195, opened 2020-07-02
rbd: Handle maximum snapshots on a single rbd image

### LINKED ISSUE #1098: Limits for RBD create PVC from snapshot
# Describe the bug #
Prior to the GA release of snapshot support, we need to ensure that the ceph-csi driver enforces some sane limits on snapshot creation and creating PVCs from snapshots.

**RBD snapshot limits** 
* no more than 510 RBD snapshots per RBD image

**RBD clone limits**
* no more than 15 images in clone chain
* no more than TBD total clones from a parent image

The ceph-csi driver can attempt to hide this internal limits by flattening child images as necessary to provide more "space" for future snapshots / cloned images.

Madhu-1:
We can have only a maximum of 510 snapshots on rbd images, due to the limitations mentioned in https://github.com/ceph/ceph-csi/issues/1098. In CephCSI we are setting a limit of maximum 450 snapshots on an rbd image (keeping a buffer to avoid issues) During the SnapshotCreate RPC call we are creating a temporary snapshot and we create a clone from the snapshot and we delete the snapshot, Even if we delete the snapshot it will be in the trash as it's still linked to the cloned images, when we list all the snapshots `snap ls --all` on an rbd image, it will show all the snapshots which are in the trash also. to remove the snapshots from the trash we need to remove the linking from cloned and the snapshots. 

As the cloned image is having the deep-flatten feature and once we flatten the cloned images which are created from the parent image, the temporary snapshots which are in the trash will be garbage collected, This will allow as to create more number of snapshots on a single rbd image. 

closes #1098 
updates: #1188


charts/ceph-csi-rbd/values.yaml
```diff
@@ -113,6 +113,8 @@ provisioner:
   # Soft limit for maximum number of nested volume clones that are taken before
   # a flatten occurs
   softMaxCloneDepth: 4
+  # Maximum number of snapshots allowed on rbd image without flattening
+  maxSnapshotsOnImage: 450
```
nixpanic, line 117: maybe mention that Ceph limits this to 510 - setting it higher will cause issues?

cmd/cephcsi.go
```diff
@@ -212,3 +213,11 @@ func validateCloneDepthFlag(conf *util.Config) {
 		klog.Fatalln("rbdsoftmaxclonedepth flag value should not be greater than rbdhardmaxclonedepth")
 	}
 }
+
+func validateMaxSnaphostFlag(conf *util.Config) {
+	// maximum number of snapshots on an image are 510,keeping limit to 500 to
+	// avoid issues.
```
nixpanic, line 219: Is this documented as a limitation in Ceph RBD? If so, add a link to theirs docs.

/cc @dillaman

internal/rbd/rbd_util.go
```diff
@@ -1149,3 +1178,44 @@ func (rv *rbdVolume) ensureEncryptionMetadataSet(status string) error {
...
+	Size      int64  `json:"size"`
+	Protected string `json:"protected"`
+	Timestamp string `json:"timestamp"`
+	Namespace struct {
+		Type         string `json:"type"`
+		OriginalName string `json:"original_name"`
+	} `json:"namespace"`
+}
+
+func (rv *rbdVolume) listSnapshots(ctx context.Context, cr *util.Credentials) ([]snapshotInfo, error) {
```
nixpanic, line 1194: please document a TODO like:
```
// TODO: use go-ceph once https://github.com/ceph/go-ceph/issues/300 is available in a release
```
We can then find it in [tickgit](https://www.tickgit.com/browse?repo=github.com/ceph/ceph-csi) or other tools

scripts/install-helm.sh
```diff
@@ -122,7 +122,7 @@ install_cephcsi_helm_charts() {
     # deleting configmap as a workaround to avoid configmap already present
     # issue when installing ceph-csi-rbd
     kubectl delete cm ceph-csi-config --namespace ${NAMESPACE}
-    "${HELM}" install --namespace ${NAMESPACE} --set provisioner.fullnameOverride=csi-rbdplugin-provisioner --set nodeplugin.fullnameOverride=csi-rbdplugin --set configMapName=ceph-csi-config --set provisioner.podSecurityPolicy.enabled=true --set nodeplugin.podSecurityPolicy.enabled=true ${RBD_CHART_NAME} "${SCRIPT_DIR}"/../charts/ceph-csi-rbd --set topology.enabled=true --set topology.domainLabels="{${NODE_LABEL_REGION},${NODE_LABEL_ZONE}}"
+    "${HELM}" install --namespace ${NAMESPACE} --set provisioner.fullnameOverride=csi-rbdplugin-provisioner --set nodeplugin.fullnameOverride=csi-rbdplugin --set configMapName=ceph-csi-config --set provisioner.podSecurityPolicy.enabled=true --set nodeplugin.podSecurityPolicy.enabled=true ${RBD_CHART_NAME} "${SCRIPT_DIR}"/../charts/ceph-csi-rbd --set topology.enabled=true --set topology.domainLabels="{${NODE_LABEL_REGION},${NODE_LABEL_ZONE}}" --set provisioner.maxSnapshotsOnImage=10
```
nixpanic, line 125: Why `10` here?

charts/ceph-csi-rbd/values.yaml
Madhu-1: Yes this will be documented when we add helm documentation, setting higher is not allowed as krbd will have issues.

scripts/install-helm.sh
Madhu-1: added the reason in commit message. as we cannot create 450 snapshots to test this feature. lowering the limit to 10 so that this can be tested in CI

internal/rbd/rbd_util.go
```diff
@@ -1149,3 +1179,45 @@ func (rv *rbdVolume) ensureEncryptionMetadataSet(status string) error {
...
+	Size      int64  `json:"size"`
+	Protected string `json:"protected"`
+	Timestamp string `json:"timestamp"`
+	Namespace struct {
+		Type         string `json:"type"`
+		OriginalName string `json:"original_name"`
+	} `json:"namespace"`
+}
+
+// TODO: use go-ceph once https://github.com/ceph/go-ceph/issues/300 is available in a release
```
Yuggupta27, line 1195:
```suggestion
// TODO: use go-ceph once https://github.com/ceph/go-ceph/issues/300 is available in a release.
```

Yuggupta27: Minor Nit.

cmd/cephcsi.go
dillaman: This is the closest documentation for upstream [1][2] for 510 snapshots and 16 images in a parent/child chain. 

[1] https://github.com/torvalds/linux/blob/master/drivers/block/rbd.c#L98
[2] https://github.com/torvalds/linux/blob/master/drivers/block/rbd.c#L92

nixpanic: @Madhu-1 could you  please add the links to these notes somewhere?

internal/rbd/controllerserver.go
```diff
@@ -699,7 +699,23 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 			},
 		}, nil
 	}
+	var snaps []snapshotInfo
+	// check the number of snapshots on image
+	snaps, err = rbdVol.listSnapshots(ctx, cr)
+	if err != nil {
+		if _, ok := err.(ErrImageNotFound); ok {
```
nixpanic, line 713: You should be able to use the new recommended notation:
```
var einf ErrImageNotFound
if errors.As(err, &einf) {
```

e2e/rbd.go
```diff
@@ -263,7 +263,9 @@ var _ = Describe("RBD", func() {
 						e2elog.Logf("backend image count %d expected image count %d", len(images), 1)
 						Fail("validate backend image failed")
 					}
-					totalCount := 5
+					// Increased the count to 8 as we can test the
+					// maxsnapshotsonimage in CI
+					totalCount := 8
```
nixpanic, line 268: I am confused by this, why pass `provisioner.maxSnapshotsOnImage=5` in the previous commit?

scripts/travis-functest.sh
```diff
@@ -39,7 +39,7 @@ if [[ "${KUBE_MAJOR}" -ge 1 ]] && [[ "${KUBE_MINOR}" -ge 17 ]]; then
 fi
 
 # functional tests
-go test "${GO_TAGS}" github.com/ceph/ceph-csi/e2e --deploy-timeout=10 -timeout=40m --cephcsi-namespace=cephcsi-e2e-$RANDOM -v -mod=vendor
+go test "${GO_TAGS}" github.com/ceph/ceph-csi/e2e --deploy-timeout=10 -timeout=60m --cephcsi-namespace=cephcsi-e2e-$RANDOM -v -mod=vendor
```
nixpanic, line 42: You could place this timeout in `build.env`, have you considered that?

e2e/rbd.go
Madhu-1: as we are setting the `provisioner.maxSnapshotsOnImage=5` the csi driver will start flattening the rbd images once the parent image has 5 snapshots (including trash). as we need to test this one at least we need to create 5+ snapshots. hope its clear now

Madhu-1:
@humblec @nixpanic E2E tests are getting timeout as the default timeout for a Travis CI instance is 50Min and our E2E tests are taking more time. we need to split out cephfs and rbd into separate tests or move to centos CI. 
for now, decreased the snapshot count to 5 to allow CI to pass. PTAL

nixpanic:
> split out cephfs and rbd into separate tests

That would be my preference. We can then run tests in parallel and hopefully speed things up.

I am still working to get the tests running in CentOS CI. Not sure when it is stable enough to replace some of the Travis CI jobs.

PR #1196, opened 2020-07-02
rbd: check rbd image status only duing DeleteVolume RPC

Madhu-1:
currently, various calls to delete image does not need the rbdStatus check. That is only required when calling from DeleteVolume. This PR optimizes the rbd image deletion by removing the status check.

Updates https://github.com/ceph/ceph-csi/issues/1188

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -547,6 +547,16 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	}
 	defer cs.VolumeLocks.Release(rbdVol.RequestName)
 
+	found, _, err := rbdStatus(ctx, rbdVol, cr)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed getting information for image (%s): (%s)"), rbdVol, err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if found {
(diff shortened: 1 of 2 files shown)


internal/rbd/controllerserver.go
```diff
@@ -547,6 +547,16 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	}
 	defer cs.VolumeLocks.Release(rbdVol.RequestName)
 
+	found, _, err := rbdStatus(ctx, rbdVol, cr)
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed getting information for image (%s): (%s)"), rbdVol, err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if found {
+		klog.Errorf(util.Log(ctx, "rbd %s is still being used"), rbdVol)
```
humblec, line 556: do we need to log it something like `rbd image %s is still ...`?

humblec: @Madhu-1

humblec:
LGTM..

internal/rbd/controllerserver.go
Madhu-1: Yes @humblec it helps for debugging. and also the code was already present, This PR is just moving the check outside

PR #1198, opened 2020-07-02
ci: generate golangci.yml with correct CEPH_VERSION

nixpanic:

When building against go-ceph, the most recent version of Ceph is
assumed to be available (currently Octopus). In case an older version of
the development packages is installed, building go-ceph will fail.

Golangci-lint does not accept the `-tags nautilus` parameter like other
Golang tools. Instead, the build-constraints need to be configured in a
confguration file.

This change takes care of the following:
- move the current scripts/golangci.yml to a template
- add the @@CEPH_VERSION@@ substitute
- generate the configuration file when needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,10 @@ mod-check: check-env
 	@echo 'running: go mod verify'
 	@go mod verify && [ "$(shell sha512sum go.mod)" = "`sha512sum go.mod`" ] || ( echo "ERROR: go.mod was modified by 'go mod verify'" && false )
 
-go-lint:
+scripts/golangci.yml: scripts/golangci.yml.in
+	sed "s/@@CEPH_VERSION@@/$(CEPH_VERSION)/g" < scripts/golangci.yml.in > scripts/golangci.yml
+
+go-lint: scripts/golangci.yml
 	./scripts/lint-go.sh
 
 lint-extras:
@@ -163,6 +166,7 @@ clean:
 	go clean -mod=vendor -r -x
 	rm -f deploy/cephcsi/image/cephcsi
 	rm -f _output/cephcsi
+	$(RM) scripts/golangci.yml
 	$(RM) e2e.test
 	[ ! -f .devel-container-id ] || $(CONTAINER_CMD) rmi $(CSI_IMAGE_NAME):devel
 	$(RM) .devel-container-id
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ deploy/cephcsi/image/cephcsi
 *.orig
 *.patch
 *.rej
+
+# generated golangci-lint configuration
+scripts/golangci.yml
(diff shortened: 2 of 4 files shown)


nixpanic:
Note that the test now shows the required build tag in the output:
```
INFO [loader] Using build tags: [nautilus]
```

build.env
```diff
@@ -18,7 +18,7 @@ GOLANG_VERSION=1.13.9
 GO111MODULE=on
 
 # static checks and linters
-GOLANGCI_VERSION=v1.21.0
+GOLANGCI_VERSION=v1.27.0
```
nixpanic, line 21: ah! this was not supposed to be part of the PR :-1:

nixpanic:
Interesting, CI fails with this error:
```
internal/rbd/controllerserver.go:468:1: cyclomatic complexity 21 of func `(*ControllerServer).DeleteVolume` is high (> 20) (gocyclo)

func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {

^
```

This PR does not touch that :thinking:

Yuggupta27:
I think It is reporting golangci-lint issues.

nixpanic:
> I think It is reporting golangci-lint issues.

yes, but why did previous golangci-lint runs succeed if there were problems!?

Yuggupta27:
RBD status PR got merged which might have triggered this...

Madhu-1:
> Interesting, CI fails with this error:
> 
> ```
> internal/rbd/controllerserver.go:468:1: cyclomatic complexity 21 of func `(*ControllerServer).DeleteVolume` is high (> 20) (gocyclo)
> 
> func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
> 
> ^
> ```
> 
> This PR does not touch that

Why CI didnt catch this one earlier? :(

nixpanic:
> Why CI didnt catch this one earlier? :(

Yeah... I don't understand that either. Now added a PR with a TODO note to look at it later.

Makefile
```diff
@@ -86,7 +86,10 @@ mod-check: check-env
 	@echo 'running: go mod verify'
 	@go mod verify && [ "$(shell sha512sum go.mod)" = "`sha512sum go.mod`" ] || ( echo "ERROR: go.mod was modified by 'go mod verify'" && false )
 
-go-lint:
+scripts/golangci.yml: scripts/golangci.yml.in
+	sed "s/@@CEPH_VERSION@@/$(CEPH_VERSION)/g" < scripts/golangci.yml.in > scripts/golangci.yml
```
Madhu-1, line 90: can we remove this generated `scripts/golangci.yml` once the testing is done?

nixpanic: We could, but there is no need to do so. If there are changes, to the template, it should get generated again.

nixpanic: Well, maybe it should get removed with `make clean` just like the container images. I'll add that.

PR #1202, opened 2020-07-03
ci: display a warning when GO_ARCH is not set for image-cephcsi

nixpanic:

`make image-cephcsi` will fail when Golang is not installed. There is no
strict requirement for Golang to be available, it is only used to gather
the architecture of the OS where the image is built. It is possible to
build the image successfully with `make image-cephcsi GOARCH=amd64`.

In case Golang is not installed, GOARCH can not be detected
automatically. This will cause a failure while installing Golang in the
container image. Because the failure is not very clear, display a
warning in the case the GO_ARCH (from ${GOARCH} in the Makefile) is not
set.

This problem hit me several times while working on CI improvements. It
is very annoying to see weird Golang installation failures that are caused
by a missing/empty variable.


Madhu-1:
>Missing GO_ARCH argument for building image https://travis-ci.org/github/ceph/ceph-csi/jobs/704574591#L488

CI failure

deploy/cephcsi/image/Dockerfile
```diff
@@ -8,14 +8,14 @@ LABEL stage="build"
 
 ARG CSI_IMAGE_NAME=quay.io/cephcsi/cephcsi
 ARG CSI_IMAGE_VERSION=canary
-ARG GO_ARCH
```
Madhu-1, line 11: This should not be removed as its a docker build argument

deploy/cephcsi/image/Dockerfile
```diff
@@ -8,14 +8,14 @@ LABEL stage="build"
...
 ARG SRC_DIR
 ARG GIT_COMMIT
 ARG GOROOT=/usr/local/go
 
 COPY build.env /
 
 RUN source /build.env && \
+    ( test -n "${GO_ARCH}" && exit 0; echo -e "\n\nMissing GO_ARCH argument for building image\n\n"; exit 1 ) && \
```
Madhu-1, line 18: can you update the error message on how to run make image-cephcsi with GO_ARCH flag?

deploy/cephcsi/image/Dockerfile
nixpanic: It is there 2x, this is a duplicate from the above

deploy/cephcsi/image/Dockerfile
nixpanic: sure

nixpanic:
> > Missing GO_ARCH argument for building image https://travis-ci.org/github/ceph/ceph-csi/jobs/704574591#L488
> 
> CI failure

This is weird, it works locally :thinking: Maybe the 2nd `ARCH GO_ARCH` is needed for Docker builds?

PR #1205, opened 2020-07-06
rbd: Add RHEL 8.2 kernel to the list to map images with deep-flatten feature.

Madhu-1:
As per @idryomov comment https://github.com/ceph/ceph-csi/issues/1188#issuecomment-653396519, the upstream kernel version v5.1 and RHEL 8.2 supports the mapping of rbd images with the deep-flatten feature. 

updates: #1188


Madhu-1:
@idryomov Thanks for providing the kernel versions, can you please review this one?

internal/rbd/nodeserver.go
```diff
@@ -69,12 +69,20 @@ var (
 	deepFlattenSupport = []util.KernelVersion{
 		{
 			Version:      5,
-			PatchLevel:   2,
+			PatchLevel:   1,
```
idryomov, line 72: Yeah, 5.2 is incorrect, not sure where it came from.

You may want to update `internal/util/util_test.go` too -- it's just a test (which looks redundant to me), but it also references 5.2.

Madhu-1: Yeah, it's just a test as 5.2 is greater than 5.1 it will pass. will correct it later as it should not block this one.

PR #1208, opened 2020-07-06
build: error out when Podman or Docker is not available

nixpanic:

In case Podman or Docker is not installed, running containerized tests
or builds will fail with unclear errors.

Introduce a 'need-container-cmd' make target that checks if the
CONTAINER_CMD is available (and non-empty). When running without Podman
and Docker, the following error is displayed:

    $ make containerized-build
    Missing container support, install Podman or Docker
    make: *** [Makefile:135: need-container-cmd] Error 1

## Related issues ##

I just hit this problem while testing CI jobs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: all cephcsi check-env
+.PHONY: all cephcsi check-env need-container-cmd
 
 CONTAINER_CMD?=$(shell docker version >/dev/null 2>&1 && echo docker)
 ifeq ($(CONTAINER_CMD),)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -130,33 +130,37 @@ cephcsi: check-env
 e2e.test: check-env
 	go test $(GO_TAGS) -mod=vendor -c ./e2e
 
+.PHONY: need-container-cmd
```
Madhu-1, line 133: Can this be added to current PHONY at line 137

nixpanic: Well, there are many `.PHONY` statements throughout the file. I'll add it where `check-env` is called.

nixpanic:
Fixed the CI issue. Ubuntu does not use a default shell that has the `command` command, using `which` instead.

PR #1212, opened 2020-07-08
Update Minikube and kubernetes to latest versions

Madhu-1:
This PR updates the things listed below
* Update Minikube to the latest version
* Fix issue with PSP in minikube >1.6.2 and <1.11.1 
* Run E2E with kubernetes 1.18.5
* Update kubernetes v1.17.5 to latest patch release in CI


Madhu-1:
The E2E is failing as it exceeds 50 Min in Travis CI

scripts/minikube.sh
```diff
@@ -104,11 +112,25 @@ up)
...
 
-    enable_psp
+    MINIKUBE_MAJOR=$(minikube_version 1)
+    MINIKUBE_MINOR=$(minikube_version 2)
+    MINIKUBE_PATCH=$(minikube_version 3)
 
     echo "starting minikube with kubeadm bootstrapper"
-    # shellcheck disable=SC2086
-    minikube start --memory="${MEMORY}" -b kubeadm --kubernetes-version="${KUBE_VERSION}" --vm-driver="${VM_DRIVER}" --feature-gates="${K8S_FEATURE_GATES}" ${EXTRA_CONFIG}
+    if [[ "${MINIKUBE_MAJOR}" -ge 1 ]] && [[ "${MINIKUBE_MINOR}" -ge 11 ]] && [[ "${MINIKUBE_PATCH}" -ge 1 ]] || [[ "${MINIKUBE_MAJOR}" -ge 1 ]] && [[ "${MINIKUBE_MINOR}" -ge 12 ]]; then
```
nixpanic, line 120: Please handle this in the `enable_psp` function, that is much cleaner.

scripts/minikube.sh
```diff
@@ -104,11 +112,25 @@ up)
...
+        enable_psp
+        # shellcheck disable=SC2086
```
nixpanic, line 122: What is being reported here? Please leave a comment when disabling a shellcheck.

.travis.yml
```diff
@@ -40,7 +40,7 @@ before_script:
   - sudo apt-get -qq update
   # only the arm64 fallback repo is unsigned and needs --allow-unauthenticated
   - sudo apt-get -y --allow-unauthenticated install librados-dev librbd-dev
-
+  - sudo apt-get -y install conntrack
```
nixpanic, line 43: please squash this change in the commit where you add the job for 1.18.5

scripts/minikube.sh
Madhu-1: not added in this one, it just got moved from line 110

scripts/minikube.sh
Madhu-1: If an additional check in enable_psp again I need to check here to handle minikube start, as it will be different for v16.2 to v1.11.0 and >v1.11.1, Handling it in enable_psp might not make sense, this is mainly related to starting a minikube

nixpanic: I do not like the current situation. This huge if-statement is very hard to read. Can you at least place that in a function, like
```
# minikube v16.2 to v1.11.0 needs to be restarted to enable PSP
function enabling_psp_needs_restart() {
    local MINIKUBE_MAJOR=...
    ...
    if ...; then
        return 1
    fi
    return 0
}
```

Madhu-1: With the suggested approach still I need to do the same if check,  not only restart we need to skip passing the psp flag to the minikube start, if you look at the minikube start it's different in both if and else case.

Madhu-1: Let me know if you prefer to move same check to a new function.

nixpanic: Yes, please move it to a function to make it more maintainable.

Feel free to name the function to something more suitable, maybe `need_psp_workaround` or `minikube_supports_psp`?

Madhu-1: Moved it to a new function

Madhu-1:
@nixpanic @humblec PTAL

Madhu-1:
CI was failing due to minor check, fixed now PTAL

Madhu-1:
@humblec PTAL

PR #695, opened 2019-10-22
Add support for smart PVC cloning

Madhu-1:
This PR adds the support for rbd to create smart cloning from a PVC, to create a new image from an existing image we are following below steps

### Create a PVC from PVC
* Create a temporary snapshot from the parent volume
* Clone a new image from a temporary snapshot with options  --rbd-default-clone-format 2 --image-feature layering,deep-flatten
* Delete temporary snapshot created
* Create a snapshot with requested Name from cloned volume
* Create a clone from a snapshot  with user-provided image features

###  Delete a PVC
* Move the image to the trash
*  Add a task to remove the image from the trash

example commands:-
```
1) rbd snap create <RBD image for src k8s volume>@<random snap name>
2) rbd clone --rbd-default-clone-format 2 --image-feature
layering,deep-flatten <RBD image for src  k8s volume>@<random snap <RBD image for temporary snap image>
3) rbd snap rm <RBD image for src k8s volume>@<random snap name>
4) rbd snap create <RBD image for temporary snap image>@<random snap name> <k8s snap name>
5) rbd clone --rbd-default-clone-format 2 --image-feature <k8s dst vol config>
<RBD image for temporary snap image>@<random snap name> <RBD image for k8s dst vol>
6) rbd snap rm <RBD image for temporary snap image>@<random snap name> 
```

Fixes: #675

Note: This PR is built on top of https://github.com/ceph/ceph-csi/pull/693, will squash/remove commits once #693 is merged

PR #1213, opened 2020-07-08
rbd: Add support for PVC-PVC cloning

Madhu-1:
Added support for RBD PVC to PVC cloning, below commands are executed to create a PVC-PVC clone from RBD side.

### steps

* Check the depth(n) of the cloned image if n>=(hard limit -2) or ((soft limit-2) Add a task to flatten the image and return about (to avoid image leak) **Note** will try to flatten the temp clone image in the chain if available
* Reserve the key and values in omap (this will help us to avoid the leak as it's not reserved earlier as we have returned ABORT (the request may not come back))
* Create a snapshot of rbd image
* Clone the snapshot (temp clone)
* Delete the snapshot
* Snapshot the temp clone
* Clone the snapshot (final clone)
* Delete the snapshot

### commands
```bash
1) check the image depth of the parent image if flatten required add a task to flatten image and return ABORT to avoid leak( hardlimit-2 and softlimit-2  check will be done)
2) Reserve omap keys
2) rbd snap create <RBD image for src k8s volume>@<random snap name>
3) rbd clone --rbd-default-clone-format 2 --image-feature
layering,deep-flatten <RBD image for src k8s volume>@<random snap> <RBD image for temporary snap image>
4) rbd snap rm <RBD image for src k8s volume>@<random snap name>
5) rbd snap create <cloned RBD image created in snapshot process>@<random snap name>
6) rbd clone --rbd-default-clone-format 2 --image-feature <k8s dst vol config> <RBD image for temporary snap image>@<random snap name> <RBD image for k8s dst vol>
7)rbd snap rm <RBD image for src k8s volume>@<random snap name>
```

### steps
* Delete temporary clone image created as part of clone(delete if present)
* Delete rbd image 

updates https://github.com/ceph/ceph-csi/issues/1188
closes https://github.com/ceph/ceph-csi/pull/695


dillaman:
Why can't this be very generic? The existing PVC snapshot and PVC from snapshot logic should already handle all the restriction logic. It would be great if you could just create a generic helper library that could be re-used by both RBD and CephFS provisioners.

Madhu-1:
> Why can't this be very generic? The existing PVC snapshot and PVC from snapshot logic should already handle all the restriction logic. It would be great if you could just create a generic helper library that could be re-used by both RBD and CephFS provisioners.

@dillaman Thanks for the suggestion that's a lot of code cleanup has to be done for that one(Need to think about it as the inputs and structures used for cephfs and rbd are quite different) Maybe once we get the feature support will work on refactoring the code.

internal/rbd/clone.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/ceph/ceph-csi/internal/journal"
+	"github.com/ceph/ceph-csi/internal/util"
+
+	librbd "github.com/ceph/go-ceph/rbd"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"k8s.io/klog"
+)
+
+// checkCloneImage check the cloned image exists, if the cloned image is not
```
nixpanic, line 33: This comment really needs some better structure. Please add line-endings and maybe even some empty lines.

The text is extremely long (very good!!), but that suggests the function could have been split in different pieces.

internal/rbd/clone.go
```diff
@@ -0,0 +1,236 @@
...
+func (rv *rbdVolume) createCloneFromImage(ctx context.Context, parentVol *rbdVolume) error {
+	// generate temp cloned volume
+	tempClone := rv.generateTempClone()
+	snap := &rbdSnapshot{
+		RbdSnapName: rv.RbdImageName,
+		Pool:        rv.Pool,
+	}
+
+	var (
+		iErr error
```
nixpanic, line 150: Please comment these errors or give them proper names, I am wondering what `i` and `f` stand for. While reading code, nobody should need to think twice about these minor things.

`iErr` might better be called `errFlatten`? (this took me surely 2 minutes to figure out, that could have been saved when proper naming/comments were used)

internal/rbd/controllerserver.go
```diff
@@ -161,6 +161,18 @@ func buildCreateVolumeResponse(ctx context.Context, req *csi.CreateVolumeRequest
 	return &csi.CreateVolumeResponse{Volume: volume}, nil
 }
 
+func getGRPCError(err error) error {
```
nixpanic, line 164: This function should probably not be used to convert any error to a GRPCError? Please clearly comment the intention so that when a new error gets added to the conversion, the comment gets updated too.

internal/rbd/clone.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		klog.Errorf(util.Log(ctx, "failed to store volume %s: %v"), rv, err)
+		return err
+	}
+	return nil
+}
+
+func (rv *rbdVolume) flattenCloneImage(ctx context.Context) error {
+	tempClone := rv.generateTempClone()
+	// reducing the limit for cloned images to make sure the limit is in range
```
nixpanic, line 216: Why!? I think I understand, but explain the need please. The comment describes what the code does, but not why it is needed.

internal/rbd/controllerserver.go
```diff
@@ -384,38 +409,53 @@ func (cs *ControllerServer) createBackingImage(ctx context.Context, cr *util.Cre
 	return nil
 }
 
-func (cs *ControllerServer) checkSnapshotSource(ctx context.Context, req *csi.CreateVolumeRequest,
-	cr *util.Credentials) (*rbdSnapshot, error) {
+func (cs *ControllerServer) checkContentSource(ctx context.Context, req *csi.CreateVolumeRequest, cr *util.Credentials) (*rbdVolume, *rbdSnapshot, error) {
```
nixpanic, line 412: This function does not use `cs *ControllerServer`, so it should most likely not be a method of that type.
(It was already like this before, so no need to change, but should be a cleanup later)

internal/rbd/rbd_journal.go
```diff
@@ -228,7 +228,7 @@ volume names as requested by the CSI drivers. Hence, these need to be invoked on
 respective CSI snapshot or volume name based locks are held, as otherwise racy access to these
 omaps may end up leaving the omaps in an inconsistent state.
 */
-func (rv *rbdVolume) Exists(ctx context.Context) (bool, error) {
+func (rv *rbdVolume) Exists(ctx context.Context, parentVol *rbdVolume) (bool, error) {
```
nixpanic, line 231: please add a description of the parameter to the comment

Madhu-1:
> just need some clarifications/comments in the code
> 
> ideally the text of the 1st commit would be copied to a file under `docs/design`.

That is one for the pending task, will add both snapshot and clone commands to a doc
> 
> No major concerns, just worried a little about the maintainability (but better comments would help with that)

@nixpanic requested changes are done PTAL.

examples/rbd/pvc-clone.yaml
```diff
@@ -0,0 +1,16 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: rbd-pvc-clone
+spec:
+  storageClassName: csi-rbd-sc
+  dataSource:
+    name: rbd-pvc
+    kind: PersistentVolumeClaim
+    apiGroup: ""
```
nixpanic, line 11: why is `apiGroup` needed?

internal/rbd/clone.go
```diff
@@ -0,0 +1,245 @@
...
+	"google.golang.org/grpc/status"
+	"k8s.io/klog"
+)
+
+// checkCloneImage check the cloned image exists, if the cloned image is not
+// found it will check the temporary cloned snapshot exists, and again it will
+// check the snapshot exists on the temporary cloned image, if yes it will
+// create a new cloned and delete the temporary snapshot and adds a task to
+// flatten the temp cloned image and return success.
+
```
nixpanic, line 38: not sure if this gets rendered correctly, might have to comment the empty lines too?
```
// flatten the temp cloned image and return success.
//
// if the temporary snapshot does not exists it creates a temporary snapshot on
```

internal/rbd/clone.go
```diff
@@ -0,0 +1,245 @@
...
+	errFlatten = tempClone.flattenRbdImage(ctx, rv.conn.Creds, false, rbdHardMaxCloneDepth, rbdSoftMaxCloneDepth)
+	if errFlatten != nil {
+		return err
+	}
+	// create snap of temp clone from temporary cloned image
+	// create final clone
+	// delete snap of temp clone
+	errClone = createRBDClone(ctx, tempClone, rv, snap, rv.conn.Creds)
+	if errClone != nil {
+		// set iErr error to cleanup temporary snapshot and temporary clone
```
nixpanic, line 200: reference to the now renamed `iErr`

**Many thanks for all the additional comments and better naming of variables!**

internal/rbd/clone.go
```diff
@@ -0,0 +1,245 @@
...
+		klog.Errorf(util.Log(ctx, "failed to store volume %s: %v"), rv, err)
+		return err
+	}
+	return nil
+}
+
+func (rv *rbdVolume) flattenCloneImage(ctx context.Context) error {
+	tempClone := rv.generateTempClone()
+	// reducing the limit for cloned images to make sure the limit is in range,
+	// If the internmediate clone reaches the depth we may need to return ABORT
```
nixpanic, line 221: Great, thanks for adding!

Note, `internmediate` should be `intermediate`

internal/rbd/controllerserver.go
```diff
@@ -161,6 +161,21 @@ func buildCreateVolumeResponse(ctx context.Context, req *csi.CreateVolumeRequest
 	return &csi.CreateVolumeResponse{Volume: volume}, nil
 }
 
+// getGRPCErrorForCreateVolume converts the returns the GRPC errors based on
```
nixpanic, line 164: lol, great name :grin:

examples/rbd/pvc-clone.yaml
Madhu-1: yeah default will be `""` removed now

Madhu-1:
@nixpanic @humblec PTAL.

Issue #1203, opened 2020-07-04
Error types carry unused data and are overly complicated


ansiwen:
The error types in internal/util/errors.go contain metadata, that is constructed and managed in many places in the code, but actually never used, for example `keyName` field in the `ErrKeyNotFound` type or `objectName` in `ErrObjectExists` [here](https://github.com/ceph/ceph-csi/blob/826f7126cdf18a698b2ac637e0821567c90131d2/internal/util/errors.go#L21). Also, inspecting errors by their type is not without issues, as you can read [here](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully).

I would suggest to remove all unnecessary code in the errors and implement them in a way that doesn't expose their types/implementation, by only using the new `Is()`/`As()`/`Wrap()` methods and maybe further interfaces. Then we minimize coupling and still can later add features as required.

PR #1225, opened 2020-07-09
simplify error handling

ansiwen:
The sentinel error code had additional fields in the errors, that are used
nowhere. This lead to unnecessarily complicated code. This change
replaces the sentinel errors in utils with standard errors created
with errors.New() and adds a simple JoinErr() function to be able to merge
sentinel errors from different code tiers.

Related: #1203

Signed-off-by: Sven Anderson <sven@redhat.com>


ansiwen:
> Note that there are more `errors.go` files in other packages.
@nixpanic This change already addresses util, cephfs and rbd. I don't think there are others. Are there?

internal/util/errors.go
```diff
@@ -16,98 +16,46 @@ limitations under the License.
...
-	Pool string
-	Err  error
+func (e errPair) Error() string {
+	return fmt.Sprintf("%v: %v", e.first, e.second)
 }
 
-// Error returns the error string for ErrPoolNotFound.
-func (e ErrPoolNotFound) Error() string {
-	return e.Err.Error()
+// Is checks if target error is wrapped in the first error
```
Yuggupta27, line 47:
```suggestion
// Is checks if target error is wrapped in the first error.
```

internal/util/errors.go
```diff
@@ -16,98 +16,46 @@ limitations under the License.
...
+func (e errPair) Is(target error) bool {
+	return errors.Is(e.first, target)
 }
 
-// Unwrap returns the encapsulated error.
-func (e ErrPoolNotFound) Unwrap() error {
-	return e.Err
+// Unwrap returns the second error
```
Yuggupta27, line 52:
```suggestion
// Unwrap returns the second error.
```

internal/util/errors.go
ansiwen: done

internal/util/errors.go
ansiwen: done

internal/util/errors.go
```diff
@@ -16,98 +16,46 @@ limitations under the License.
...
+	ErrSnapNameConflict = errors.New("snapshot name conflict")
+	// ErrPoolNotFound is returned when pool is not found
+	ErrPoolNotFound = errors.New("pool not found")
+)
 
-// NewErrSnapNameConflict returns a ErrSnapNameConflict error when CSI snap name already exists.
-func NewErrSnapNameConflict(name string, err error) ErrSnapNameConflict {
-	return ErrSnapNameConflict{name, err}
+type errPair struct {
+	first  error
```
Madhu-1, line 39: can you add a comment what is the first error and second error? what we get more errors returns from multiple internal functions?

internal/util/errors.go
```diff
@@ -16,98 +16,46 @@ limitations under the License.
...
+func (e errPair) Unwrap() error {
+	return e.second
 }
 
-// NewErrPoolNotFound returns a new ErrPoolNotFound error.
-func NewErrPoolNotFound(pool string, err error) ErrPoolNotFound {
-	return ErrPoolNotFound{pool, err}
+// JoinErr combines two errors. Of the returned error, Is() follows the first
+// branch, Unwrap() folllows the second branch.
+func JoinErr(e1, e2 error) error {
```
Madhu-1, line 59: this function can be renamed to JoinError

ansiwen: But shouldn't it be `JoinErrors` then? And I would prefer the shorter version, because all errors also have to start with `Err`. Do you folks have a style guide regarding that, or is it your personal preference? What do others think?

Madhu-1: Yeah JoinErrors makes more sense , it was my personal preference,
Cc @humblec @nixpanic

internal/util/errors.go
ansiwen: Can you rephrase the second sentence? I don't understand it. The `first` and `second` don't have deeper semantics, it's how pair members are named in [C++](http://www.cplusplus.com/reference/utility/pair/), and I just used that. I could also call it `left` and `right` like in binary tree nodes. I have comments later with the exported `Is()` and `Unwrap()` methods, how they treat `first` and `second` differently. But a `errors.Is(myJoinedError, MySentinelError)` would always automatically look into both `first` and `second` errors, if it can find `MySentinelError` there, if this is your question.

nixpanic: I would probably go for `inner` and `outer` error, so that `Unwrap()` returns the inner one. Thoughts?

internal/util/errors.go
nixpanic: For most functions, `..Error..()` and `..Errors..()` seem to be used. `Err..` as constants/variables is common, but would not apply here.

internal/util/errors.go
```diff
@@ -16,98 +16,46 @@ limitations under the License.
...
+	// different properties, and hence is in conflict with the passed in CSI volume name
+	ErrSnapNameConflict = errors.New("snapshot name conflict")
+	// ErrPoolNotFound is returned when pool is not found
+	ErrPoolNotFound = errors.New("pool not found")
+)
 
-// NewErrSnapNameConflict returns a ErrSnapNameConflict error when CSI snap name already exists.
-func NewErrSnapNameConflict(name string, err error) ErrSnapNameConflict {
-	return ErrSnapNameConflict{name, err}
+type errPair struct {
```
nixpanic, line 38: As with `JoinErrors()` below, maybe the type should be called `errorPair`?

internal/cephfs/volumeoptions.go
```diff
@@ -230,8 +230,8 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 	// before other errors
 	err := vi.DecomposeCSIID(volID)
 	if err != nil {
-		err = fmt.Errorf("error decoding volume ID (%s) (%s)", err, volID)
-		return nil, nil, ErrInvalidVolID{err}
+		return nil, nil, fmt.Errorf("%w: error decoding volume ID (%s) (%s)",
+			ErrInvalidVolID, err, volID)
```
nixpanic, line 234: This drops the wrapping of `err`. It was probably never used, but doing a `JoinErrors(ErrInvalidVolID, err)` might be better?

ansiwen: It wasn't wrapped before, but it's a good idea to do so, so I did. :-)

internal/util/errors.go
ansiwen: For me inner and outer would suggest that there is just a linear wrapping, but in fact it's a bifurcation, so you could unwrap the inner _and_ outer errors. (In my first version I also had a `UnwrapFirst()` function, to make it accessible, but then I changed my mind and wanted to wait until we really need that before adding it.)

ansiwen:
I think I addressed all points and rebased. PTAL.

internal/util/errors.go
nixpanic: Yes, both errors are equal in the ErrorPair. Let's put it differently... In many use-cases I expect that one of the errors has a higher priority, and that one should get returned with `Unwrap()`. Users of `JoinErrors()` would normally not call `Unwrap()` themselves, but use `errors.As()` instead. No need for additional functions that will likely not get used.

It would be good to use terminology that gives some understanding which error (the higher priority one) gets returned in `Unwrap()`.

I don't really care how they are called, but `left` and `right` does not give a hint what `Unwrap()` would return.

internal/util/errors.go
```diff
@@ -16,98 +16,46 @@ limitations under the License.
...
+func (e errorPair) Error() string {
+	return fmt.Sprintf("%v: %v", e.first, e.second)
 }
 
-// Error returns the error string for ErrPoolNotFound.
-func (e ErrPoolNotFound) Error() string {
-	return e.Err.Error()
+// Is checks if target error is wrapped in the first error.
+func (e errorPair) Is(target error) bool {
+	return errors.Is(e.first, target)
```
nixpanic, line 49: I think this makes more sense if this is

    return errors.Is(e.first, target) || errors.Is(e.second, target)

internal/util/errors.go
```diff
@@ -16,98 +16,46 @@ limitations under the License.
...
 }
 
-// Unwrap returns the encapsulated error.
-func (e ErrPoolNotFound) Unwrap() error {
-	return e.Err
+// Unwrap returns the second error.
+func (e errorPair) Unwrap() error {
+	return e.second
```
nixpanic, line 54: I see `Unwrap()` as a function to convert one error into an other. That makes it confusing to get the 2nd error back with `Unwrap()` (my thinking is something along primary, secondary error).

What is your motivation to use `first` and `second`? If I can follow your line of thought, it might make more sense to me.

ansiwen: An `errorPair` in fact wraps two errors, with no further meaning. So the `Unwrap()` method must return one of them. It doesn't matter which one. And the `Is()` must check for the other one. So that both branches are searched for the given Error.

As I said before, I just took the naming from C++ pairs, where they are also called first and second, without any further meaning. The only "meaning" in this case would be, that when converted to a string, it concatenates them like `first.Error() + ": " + second.Error()`. Also, when doing an `errors.Is()`, the `first` error would be checked first (by using `Is()`), after that it would check `second` (by using `Unwrap()`). That's just how `errors.Is()` is implemented, that it first uses `Is()` and then `Unwrap()`. That's all. In my first patch-set I called it `left` and `right`. Would that work better for you?

internal/util/errors.go
ansiwen: Then second would be unnecessarily checked twice, because `errors.Is()` would try unwrapping, if `Is()` returns false. We _could_ of course let `Unwrap()` always return `nil`, and let `Is()` check both branches. But we would come to exactly the same behavior with more complicated code. (Because the `Is()` and `Unwrap()` methods are not meant to be called directly, but only from the `errors` package.)

internal/util/errors.go
nixpanic: Ok, thanks for the explanation. I did not check how `errors.Is()` and `errors.As()` are implemented, so with these details it should be good to go :+1:

Issue #1230, opened 2020-07-12
Parallelize the Snapshot Create and Restore and PVC-PVC cloning


Madhu-1:
With the current implementation, the locking is taken based on the ID. if the lock already exists we return operation already exists Error Message, This works fine for Volume create/Delete/Resize. The current locking makes the operation serialized When its come for snapshot create and delete we need to take a lock on the parent volume when we are creating and deleting to Ensure that the parent volume should not be deleted or resized when we are taking a snapshot, This Holds true when we are restoring a snapshot into a new RBD image also. 

During the cloning operation
1. Snapshot Restore:- When we get a request to Restore a snapshot to a new RBD Imae, we need to take a lock on the snapshot and also we need to ensure that the snapshot should not be deleted when restore operation is going on

2. Image to Image Clone:- When we get a request to Create an RBD Image-Image cloning we need to ensure that the request parent image should not be deleted or Resized when we are creating an RBD Image-Image cloning.

If we take an Operation based locking this helps in allowing parallel operation, Which means we can create parallel snapshots on an RBD image in a given time. Even We can Restore snapshots and Create Clones from an RBD image.

Madhu-1:
This is tracked in  https://github.com/ceph/ceph-csi/issues/1188

PR #447, opened 2019-06-26
E2E for parallel create/delete RBD pvc and snapshot

Madhu-1:
changing  PVC size will help us to create a large number of PVC  and create a large number of snapshots and validate the same in backend

updated E2E to create/delete  parallel PVC and check images in the backend

updated E2E to create/delete the parallel snapshot and check snapshots in the backend

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

PR #1234, opened 2020-07-13
Take operation lock to allow parallel snapshot Create, PVC Restore and PVC Clone

Madhu-1:
As we are adding new functionalities like Create/Delete snapshot, Clone from Snapshot, and Clone from Volume. with the current implementation, there are only serial operations allowed for these functionalities, for some function we can allow parallel operations like Clone from snapshot and Clone from Volume and Create `N` snapshots on a single volume.

Delete Volume: Need to ensure that there is no clone, Snapshot create and  Expand volume in progress. if the operation exists return error to the caller

Expand Volume: Need to ensure that there is no clone, snapshot create and cloning in progress, if the operation exists return error to the caller

Delete Snapshot: Need to ensure that there is no cloning in progress if the operation exists return error to the caller

Restore Volume/Snapshot: Need to ensure that there is no Expand or delete operation in progress. if the operation exists return error to the caller

closes: #1230 
updates #1188

### DIFF
--- a/internal/util/idlocker.go
+++ b/internal/util/idlocker.go
@@ -14,9 +14,11 @@ limitations under the License.
 package util
 
 import (
+	"fmt"
 	"sync"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	klog "k8s.io/klog/v2"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


internal/util/idlocker.go
```diff
@@ -59,3 +61,201 @@ func (vl *VolumeLocks) Release(volumeID string) {
...
+	switch op {
+	case createOpt:
+		// check any delete operation is going on for given volume ID
+		if _, ok := ol.locks[deleteOpt][volumeID]; ok {
+			return fmt.Errorf("a Delete operation with given id %s already exists", volumeID)
+		}
+		// increment the counter for snapshot create operation
+		val := ol.locks[createOpt][volumeID]
+		ol.locks[createOpt][volumeID] = val + 1
+	case cloneOpt:
```
chenxu1990, line 108: Is clone or restore operation safe when Expand volume in progress ?

Madhu-1: Good catch. Yeah need to add check for expand also, restore is safe as its from snapshot

Madhu-1: Added check for expand also

internal/util/idlocker.go
```diff
@@ -59,3 +61,206 @@ func (vl *VolumeLocks) Release(volumeID string) {
...
+		if _, ok := ol.locks[cloneOpt][volumeID]; ok {
+			return fmt.Errorf("a Clone operation with given id %s already exists", volumeID)
+		}
+		// check any delete operation is going on for given volume ID
+		if _, ok := ol.locks[createOpt][volumeID]; ok {
+			return fmt.Errorf("a Create operation with given id %s already exists", volumeID)
+		}
+
+		// check any expand operation is going on for given volume ID
+		if _, ok := ol.locks[expandOpt][volumeID]; ok {
```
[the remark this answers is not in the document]
Madhu-1: got added when rebasing :facepalm: removed double-check,

Madhu-1:
> Should there be a GetSnapshotDeleteLock also?

GetDeleteLock will handle it.

Madhu-1:
@MergifyIO rebase

Madhu-1:
@humblec @ShyamsundarR PTAL

Madhu-1:
@ShyamsundarR @humblec PTAL

internal/util/idlocker.go
```diff
@@ -59,3 +61,202 @@ func (vl *VolumeLocks) Release(volumeID string) {
...
+	deleteOpt  operation = "delete"
+	cloneOpt   operation = "clone"
+	restoreOpt operation = "restore"
+	expandOpt  operation = "expand"
+)
+
+// OperationLock implements a map with atomic operations.
+type OperationLock struct {
+	locks map[operation]map[string]int
+	mux   sync.Mutex
```
nixpanic, line 78: What does this mutex protect, when to acquire/release it?

internal/util/idlocker.go
```diff
@@ -59,3 +61,202 @@ func (vl *VolumeLocks) Release(volumeID string) {
...
+}
+
+// NewOperationLock returns new OperationLock.
+func NewOperationLock() *OperationLock {
+	lock := make(map[operation]map[string]int)
+	lock[createOpt] = make(map[string]int)
```
nixpanic, line 84: `...Opt`? Ah, `operation`! See above

internal/util/idlocker.go
```diff
@@ -59,3 +61,202 @@ func (vl *VolumeLocks) Release(volumeID string) {
 	defer vl.mux.Unlock()
 	vl.locks.Delete(volumeID)
 }
+
+type operation string
+
+var (
```
nixpanic, line 67: these should probably be const

internal/util/idlocker.go
```diff
@@ -59,3 +61,202 @@ func (vl *VolumeLocks) Release(volumeID string) {
...
+	defer ol.mux.Unlock()
+	switch op {
+	case createOpt:
+		// check any delete operation is going on for given volume ID
+		if _, ok := ol.locks[deleteOpt][volumeID]; ok {
+			return fmt.Errorf("a Delete operation with given id %s already exists", volumeID)
+		}
+		// increment the counter for snapshot create operation
+		val := ol.locks[createOpt][volumeID]
+		ol.locks[createOpt][volumeID] = val + 1
```
nixpanic, line 107: Increasing the count is not mentioned in the comment for the function. It is not really clear why that is needed (but maybe the comment for the locks structure is just missing)

internal/util/idlocker.go
```diff
@@ -59,3 +61,202 @@ func (vl *VolumeLocks) Release(volumeID string) {
...
+}
+
+// TryAcquire tries to acquire the lock for operating on volumeID and returns true if successful.
+// If another operation is already using volumeID, returns false.
+func (ol *OperationLock) TryAcquire(op operation, volumeID string) error {
+	ol.mux.Lock()
+	defer ol.mux.Unlock()
+	switch op {
+	case createOpt:
+		// check any delete operation is going on for given volume ID
```
nixpanic, line 101: Instead of commenting what the next check obviously does, explain why this operation is not allowed to be queued. Here, and for each switch/case.

internal/util/idlocker.go
```diff
@@ -59,3 +61,202 @@ func (vl *VolumeLocks) Release(volumeID string) {
...
+		}
+		ol.locks[expandOpt][volumeID] = 1
+	default:
+		return fmt.Errorf("%v operation not supported", op)
+	}
+	return nil
+}
+
+// GetSnapshotCreateLock gets the snapshot lock on given volumeID.
+func (ol *OperationLock) GetSnapshotCreateLock(volumeID string) error {
```
nixpanic, line 169: My initial feeling is that these functions are not very useful. Make the operations like `createOpt` available as `CREATE_OPERATION` and pass it to `TryAcquire()` directly is just as clear (if not clearer).

Or, make `tyrAcquire()` and `release()` internal, and force using these function.

internal/util/idlocker_test.go
```diff
@@ -50,3 +50,80 @@ func TestIDLocker(t *testing.T) {
...
+		t.Errorf("failed to acquire clone lock for %s %s", volumeID, err)
+	}
+	err = lock.GetCloneLock(volumeID)
+	if err != nil {
+		t.Errorf("failed to acquire clone lock for %s %s", volumeID, err)
+	}
+	// release all clone locks
+	lock.ReleaseCloneLock(volumeID)
+	lock.ReleaseCloneLock(volumeID)
+	lock.ReleaseCloneLock(volumeID)
```
nixpanic, line 89: what happens if there are more releases than acquires?

internal/util/idlocker.go
```diff
@@ -14,9 +14,11 @@ limitations under the License.
 package util
 
 import (
+	"fmt"
 	"sync"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	klog "k8s.io/klog/v2"
```
nixpanic, line 21: use the new log functions from `util` instead please

Madhu-1: This is an error log, not defined in util

internal/util/idlocker_test.go
Madhu-1: its safer no op nothing will happen as the key won't be present to delete it

Madhu-1:
@nixpanic added more comments to provide more clarity PTAL

internal/util/idlocker.go
nixpanic: @Yuggupta27 it seems logging errors need to get added

internal/util/idlocker.go
```diff
@@ -59,3 +61,224 @@ func (vl *VolumeLocks) Release(volumeID string) {
...
+			} else {
+				delete(ol.locks[restoreOp], volumeID)
+			}
+		}
+	case expandOp:
+		// remove volumeID from map
+		delete(ol.locks[expandOp], volumeID)
+	default:
+		klog.Errorf("%v operation not supported", op)
+	}
```
nixpanic, line 283: It would be shorter to just do the `val - 1` for everything, and then a single `if val == 0 { delete(..) }` before returning from this function. No need to have the `val > 1` check for each case.

internal/util/idlocker_test.go
nixpanic: ok

internal/util/idlocker.go
Madhu-1: The error doesn't have any depth as const value, we are using the Errorf from klog directly.

Madhu-1:
@nixpanic PTAL

Madhu-1:
applied https://github.com/ceph/ceph-csi/pull/447 on top of this PR to verify the fcunctionality

internal/util/idlocker.go
nixpanic: There is a `klog.ErrorDepth()` function, just like `klog.WarningDepth()`. These should get added to our `util/log.go` so that we can use a common logging interface (and use klog internally there).

Not a blocker for this PR, but definitely something we should have as an improvement.

internal/util/idlocker.go
nixpanic: I meant that `delete()` is not needed in every case. Just return in the `default:` 

Remove these from the cases
```
			if ol.locks[cloneOpt][volumeID] == 0 {
				delete(ol.locks[cloneOpt], volumeID)
			}
```
Only have the subtraction there (for each case?).

And before the end of the function you can do
```
	if ol.locks[op][volumeID] == 0 {
		delete(ol.locks[op], volumeID)
	}
```

This reduces a number of if-statements, and no special handling is needed for `expandOp` either. You could even use `fallthrough` and have a single handler for all operations that way.

internal/util/idlocker.go
Madhu-1: Yeah, just now noticed, Thanks

Yuggupta27: Sure, more logging functions are needed. I'll create an issue for this one and address this at the rest of the code too.

humblec:
lgtm. Will wait for the e2e to pass.

Madhu-1:
@nixpanic @humblec fixed small issue in e2e . CI passed PTAL

humblec:
@nixpanic can you give final approval on this ?

PR #1244, opened 2020-07-16
rbd: DisAllow CreateVoulume for missmatch volume size

Madhu-1:
If the requested volume size and the snapshot or the parent volume from which the clone is to be created is not equal cephcsi returns an error message.

updates #1188

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -175,6 +175,33 @@ func getGRPCErrorForCreateVolume(err error) error {
 	return status.Error(codes.Internal, err.Error())
 }
 
+// validateRequestedVolumeSize validates the request volume size with the
+// source snapshot or volume size, if there is a size missmatch it returns an error.
+func validateRequestedVolumeSize(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cr *util.Credentials) error {
+	if rbdSnap != nil {
+		vol := generateVolFromSnap(rbdSnap)
+		err := vol.Connect(cr)
+		if err != nil {
+			return status.Error(codes.Internal, err.Error())
+		}
+		defer vol.Destroy()
+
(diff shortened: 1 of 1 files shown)


internal/rbd/controllerserver.go
```diff
@@ -175,6 +175,33 @@ func getGRPCErrorForCreateVolume(err error) error {
 	return status.Error(codes.Internal, err.Error())
 }
 
+// validateRequestedVolumeSize validates the request volume size with the
+// source snapshot or volume size, if there is a size missmatch it returns an error.
+func validateRequestedVolumeSize(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cr *util.Credentials) error {
+	if rbdSnap != nil {
+		vol := generateVolFromSnap(rbdSnap)
+		err := vol.Connect(cr)
+		if err != nil {
```
Yuggupta27, line 184:
```suggestion
		if err := vol.Connect(cr); err != nil {
```

internal/rbd/controllerserver.go
```diff
@@ -175,6 +175,33 @@ func getGRPCErrorForCreateVolume(err error) error {
...
+			return status.Error(codes.Internal, err.Error())
+		}
+		defer vol.Destroy()
+
+		err = vol.getImageInfo()
+		if err != nil {
```
Yuggupta27, line 190:
```suggestion
		if err = vol.getImageInfo(); err != nil {
```

internal/rbd/controllerserver.go
Madhu-1: Will it fail in correct model? Any ci check will fail?

Madhu-1: If I do short variable declared with if I need to se for other places also, I prefer this way as its used  in most of the places

Issue #1256, opened 2020-07-21
RBD: Enable support for reflink on XFS


nixpanic:
# Describe the feature you'd like to have #

XFS supports reflinking that makes it possible to do copy-on-write with the `copy_file_range()` syscall. Not all Linux kernels support this feature, and when enabled on an RBD image, kernels without support will not be able to mount the filesystem.

Being conservative, and preventing breakage of users that migrate from Ceph Nautilus to Ceph Octopus based container images, reflink for XFS has been disabled by default (#966).

Users may want to enable reflink support on Kubernetes clusters that have current Linux kernels. The following proposed option in a StorageClass would make this possible:

```yaml
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
   name: csi-rbd-sc
provisioner: rbd.csi.ceph.com
parameters:
   pool: rbd
   reflink: disabled|enabled|auto
```
Values for the new `reflink` option:
- **disabled**: detect support in `mkfs.xfs`, and pass `-m reflink=0` if supported (default)
- **enabled**: force enabling, always pass `-m reflink=1`, will fail to mount on old Linux kernels
- **auto**: do nothing, do not pass `-m reflink=` and make this default in some next Ceph-CSI version

# Additional context #

See #966 for related problems.

Madhu-1:
Thanks for opening this one. Let's track it for the next release.

humblec:
Closing this issue as its already fixed.

Madhu-1:
@humblec this is not fixed, this is a feature request to enable reflink, reopening it for the se reason

Madhu-1:
@nixpanic can this be done in 3.1.0? or do we need to move it out to the next release?

humblec:
@nixpanic please adjust the milestone accordingly.

humblec:
Moving this from 3.1.0 considering the release timeline.

Madhu-1:
@nixpanic should we keep it alive or close it now and reopen if required?

nixpanic:
I don't think we care much about reflink on RBD volumes. There have been no user requests about it, so closing it should be fine. We can always revisit if someone needs reflink on RBD volumes.

saolof:
The latest version of GNU coreutils `cp` that has made it into most mainstream distros will reflink by default, and scripts are starting to assume that copying files is fast. Postgres pg_upgrade scripts use it for non-destructive in-place version upgrades with low downtime and without doubling the size of the storage used for example.

[...]

PR #1257, opened 2020-07-21
rbd: disable reflink while creating XFS filesystems

nixpanic:

Current versions of the mkfs.xfs binary enable reflink support by
default. This causes problems on systems where the kernel does not
support this feature. When the kernel the feature does not support, but
the filesystem has it enabled, the following error is logged in `dmesg`:

    XFS: Superblock has unknown read-only compatible features (0x4) enabled

Introduce a check to see if mkfs.xfs supports the `-m reflink=` option.
In case it does, pass `-m reflink=0` while creating the filesystem.

The check is executed once during the first XFS filesystem creation. The
result of the check is cached until the nodeserver restarts.

By disabling reflink, PVCs with RBD+XFS will be usable on systems that have a kernel that does not support reflink (like CentOS-7). Mixing systems with new XFS (CentOS-8) and old kernels will not be a problem.

## Related issues ##

Fixes: #966

## Future concerns ##

Reflink is a feature that could be useful for some workloads. There is a plan to enable the feature through #1256.


internal/rbd/nodeserver.go
```diff
@@ -460,6 +472,11 @@ func (ns *NodeServer) mountVolumeToStagePath(ctx context.Context, req *csi.NodeS
 			args = []string{"-m0", "-Enodiscard,lazy_itable_init=1,lazy_journal_init=1", devicePath}
 		} else if fsType == "xfs" {
 			args = []string{"-K", devicePath}
+			// always disable reflink
+			// TODO: make enabling an option, see ceph/ceph-csi#1256
+			if ns.xfsSupportsReflink() {
+				args = append(args, "-m", "reflink=0")
```
Madhu-1, line 478: Define a new variable to store 0 or 1 , it will help when we want to re-enable it

internal/rbd/nodeserver.go
```diff
@@ -870,3 +887,27 @@ func openEncryptedDevice(ctx context.Context, volOptions *rbdVolume, devicePath
 
 	return mapperFilePath, nil
 }
+
+// xfsSupportsReflink checks if mkfs.xfs supports the "-m reflink=-|1"
```
Madhu-1, line 891: reflink=0|1

internal/rbd/nodeserver.go
nixpanic: There is no logic to enable it yes, so I do not see why this PR should make it configurable.

internal/rbd/nodeserver.go
nixpanic: ah, yeah...

Madhu-1:
@humblec PTAL

internal/rbd/nodeserver.go
```diff
@@ -870,3 +887,27 @@ func openEncryptedDevice(ctx context.Context, volOptions *rbdVolume, devicePath
...
 }
+
+// xfsSupportsReflink checks if mkfs.xfs supports the "-m reflink=0|1"
+// argument. In case it is supported, return true.
+func (ns *NodeServer) xfsSupportsReflink() bool {
+	// return cached value, if set
+	if xfsHasReflink != xfsReflinkUnset {
+		return xfsHasReflink == xfsReflinkSupport
+	}
+
```
humblec, line 898: @nixpanic do we need this whitespace here?

internal/rbd/nodeserver.go
```diff
@@ -870,3 +887,27 @@ func openEncryptedDevice(ctx context.Context, volOptions *rbdVolume, devicePath
...
+	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: utilexec.New()}
+	out, err := diskMounter.Exec.Command("mkfs.xfs").CombinedOutput()
+	if err != nil {
+		// mkfs.xfs should fail with an error message (and help text)
+		if strings.Contains(string(out), "reflink=0|1") {
+			xfsHasReflink = xfsReflinkSupport
+			return true
+		}
+	}
+
```
humblec, line 910: @nixpanic do we need this whitespace here?

internal/rbd/nodeserver.go
```diff
@@ -84,6 +92,10 @@ var (
 			Backport:     true,
 		}, // RHEL 8.2
 	}
+
```
humblec, line 95: @nixpanic do we need this whitespace here?

nixpanic: not sure if it is required, but splitting a multi-line variable with an empty line between the next multi-line comment looks clean to me

internal/rbd/nodeserver.go
nixpanic: not required, but it splits two parts of unrelated code:
1. check the cached value
2. no cache, so test for it

internal/rbd/nodeserver.go
nixpanic: Just clean coding, I think? This is not an else statement that would follow the if

humblec:
@nixpanic my small concern here is, the pattern we follow for especially for whitespace or lines in between . Some reviewers say remove it ,  we should have some guidelines . Otherwise its going to be messy to read through the code in different sources..etc. @Madhu-1 @ShyamsundarR thoughts?

Madhu-1:
> @nixpanic my small concern here is, the pattern we follow for especially for whitespace or lines in between . Some reviewers say remove it , we should have some guidelines . Otherwise its going to be messy to read through the code in different sources..etc. @Madhu-1 @ShyamsundarR thoughts?

Sure , please open an issue to add guidelines doc.

nixpanic:
> my small concern here is, the pattern we follow for especially for whitespace or lines in between . Some reviewers say remove it , we should have some guidelines . Otherwise its going to be messy to read through the code in different sources..etc.

My preference strongly goes to having an empty line between sections of code; basically the place where a comment could be placed for sections that might be non-trivial to understand. Dense code is much more exhausting to read. Splitting code sections by empty lines also makes it a little easier to identify if a section could be split into its own function, making the work for maintainers a little easier.

internal/rbd/nodeserver.go
humblec: so in many places we dont leave a newline before multiline comment and in some places like this we do. Thats what I am trying to understand.

ShyamsundarR:
> > my small concern here is, the pattern we follow for especially for whitespace or lines in between . Some reviewers say remove it , we should have some guidelines . Otherwise its going to be messy to read through the code in different sources..etc.
> 
> My preference strongly goes to having an empty line between sections of code; basically the place where a comment could be placed for sections that might be non-trivial to understand. Dense code is much more exhausting to read. Splitting code sections by empty lines also makes it a little easier to identify if a section could be split into its own function, making the work for maintainers a little easier.

I agree with Niels, sections of code should be separated by a line space, allows for better redability of code.

PR #1259, opened 2020-07-22
Remove duplicated execCommand() functions and use util.ExecCommand()

nixpanic:

Both cephFS and RBD components have their own `execCommand()` with slightly different semantics. There is also a `util.ExecCommand()` that is used a few times too.

In order to clean things up, remove the `cephfs.execCommand()` and `rbd.execCommand()` functions, and replace the calls with `util.ExecCommand()`.

### DIFF
--- a/internal/cephfs/util.go
+++ b/internal/cephfs/util.go
@@ -31,12 +31,12 @@ import (
 type volumeID string
 
 func execCommandErr(ctx context.Context, program string, args ...string) error {
-	_, _, err := util.ExecCommand(program, args...)
+	_, _, err := util.ExecCommand(ctx, program, args...)
 	return err
 }
 
 func execCommandJSON(ctx context.Context, v interface{}, program string, args ...string) error {
-	stdout, _, err := util.ExecCommand(program, args...)
+	stdout, _, err := util.ExecCommand(ctx, program, args...)
 	if err != nil {
 		return err
 	}
--- a/internal/cephfs/util.go
+++ b/internal/cephfs/util.go
@@ -41,7 +41,7 @@ func execCommandJSON(ctx context.Context, v interface{}, program string, args ..
 		return err
 	}
 
-	if err = json.Unmarshal(stdout, v); err != nil {
+	if err = json.Unmarshal([]byte(stdout), v); err != nil {
 		return fmt.Errorf("failed to unmarshal JSON for %s %v: %s: %w", program, util.StripSecretInArgs(args), stdout, err)
 	}
 
--- a/internal/cephfs/util.go
+++ b/internal/cephfs/util.go
@@ -17,11 +17,9 @@ limitations under the License.
 package cephfs
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
-	"os/exec"
 
 	"github.com/ceph/ceph-csi/internal/util"
 
@@ -32,38 +30,13 @@ import (
 
 type volumeID string
 
-func execCommand(ctx context.Context, program string, args ...string) (stdout, stderr []byte, err error) {
-	var (
-		cmd           = exec.Command(program, args...) // #nosec:G204, not called with user specified parameters.
-		sanitizedArgs = util.StripSecretInArgs(args)
-		stdoutBuf     bytes.Buffer
-		stderrBuf     bytes.Buffer
-	)
-
-	cmd.Stdout = &stdoutBuf
-	cmd.Stderr = &stderrBuf
-
-	util.DebugLog(ctx, "cephfs: EXEC %s %s", program, sanitizedArgs)
-
-	if err := cmd.Run(); err != nil {
-		if cmd.Process == nil {
-			return nil, nil, fmt.Errorf("cannot get process pid while running %s %v: %w: %s",
-				program, sanitizedArgs, err, stderrBuf.Bytes())
-		}
-		return nil, nil, fmt.Errorf("an error occurred while running (%d) %s %v: %w: %s",
-			cmd.Process.Pid, program, sanitizedArgs, err, stderrBuf.Bytes())
-	}
-
-	return stdoutBuf.Bytes(), stderrBuf.Bytes(), nil
-}
-
 func execCommandErr(ctx context.Context, program string, args ...string) error {
-	_, _, err := execCommand(ctx, program, args...)
+	_, _, err := util.ExecCommand(program, args...)
 	return err
 }
 
 func execCommandJSON(ctx context.Context, v interface{}, program string, args ...string) error {
-	stdout, _, err := execCommand(ctx, program, args...)
+	stdout, _, err := util.ExecCommand(program, args...)
 	if err != nil {
 		return err
 	}
--- a/internal/rbd/rbd_attach.go
+++ b/internal/rbd/rbd_attach.go
@@ -159,13 +159,13 @@ func checkRbdNbdTools() bool {
 	_, err := os.Stat(fmt.Sprintf("/sys/module/%s", moduleNbd))
 	if os.IsNotExist(err) {
 		// try to load the module
-		_, err = execCommand("modprobe", []string{moduleNbd})
+		_, _, err = util.ExecCommand("modprobe", moduleNbd)
 		if err != nil {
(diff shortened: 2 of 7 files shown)


internal/cephfs/util.go
```diff
@@ -32,38 +30,13 @@ import (
...
-		return nil, nil, fmt.Errorf("an error occurred while running (%d) %s %v: %w: %s",
-			cmd.Process.Pid, program, sanitizedArgs, err, stderrBuf.Bytes())
-	}
-
-	return stdoutBuf.Bytes(), stderrBuf.Bytes(), nil
-}
-
 func execCommandErr(ctx context.Context, program string, args ...string) error {
-	_, _, err := execCommand(ctx, program, args...)
+	_, _, err := util.ExecCommand(program, args...)
```
Madhu-1, line 34: Make ctx a parameter of execcommand it helps for debugging, as we log request if and failure

nixpanic: I think logging should be part of the function that tries to execute the command. It should only be needed to log the command in case an error occurs, if it is not expected/handled.

Would you be fine with additional logging in cases where it is not done sufficiently?

Madhu-1: When debugging in larger system logging with request id always helps. 

Yeah we also need to log the commands what we are executing

Madhu-1: Ctx is unused in this function

nixpanic: So, do you want me to drop context from all command execution calls, and add logging in the function that calls `util.ExecCommand()`, or do you prefer logging in `util.ExecCommand()` itself, regardless of the success/failure of the command?

Madhu-1: I prefer to log the command with ctx regardless of success/failure in ExecCommand function

nixpanic: This should be done now, passing the context.Context when suitable (only 2 occasions where it isn't needed).

internal/rbd/rbd_util.go
```diff
@@ -893,7 +882,9 @@ func (rv *rbdVolume) updateVolWithImageInfo(cr *util.Credentials) error {
 	// rbd --format=json info [image-spec | snap-spec]
 	var imgInfo imageInfo
 
-	stdout, stderr, err := util.ExecCommand("rbd",
+	stdout, stderr, err := util.ExecCommand(
+		context.TODO(),
```
Madhu-1, line 886: pass the ctx to `updateVolWithImageInfo` which helps to log request ID

internal/rbd/rbd_util.go
```diff
@@ -1063,16 +1054,14 @@ func cleanupRBDImageMetadataStash(path string) error {
...
 // resizeRBDImage resizes the given volume to new size.
 func resizeRBDImage(rbdVol *rbdVolume, cr *util.Credentials) error {
-	var output []byte
-
 	mon := rbdVol.Monitors
 	volSzMiB := fmt.Sprintf("%dM", util.RoundOffVolSize(rbdVol.VolSize))
 
 	args := []string{"resize", rbdVol.String(), "--size", volSzMiB, "--id", cr.ID, "-m", mon, "--keyfile=" + cr.KeyFile}
-	output, err := execCommand("rbd", args)
+	_, stderr, err := util.ExecCommand(context.TODO(), "rbd", args...)
```
Madhu-1, line 1061: same here also

internal/rbd/rbd_util.go
nixpanic: this is to be re-written with the go-ceph API, and in that case no command will get executed anymore. The context.Context() will then get removed. The only caller (`getImageInfo()`) would need to get the context passed as well. I do not think this is worth the effort.

internal/rbd/rbd_util.go
nixpanic: this one is straight forward, I'll add that

internal/rbd/rbd_util.go
Madhu-1: Sounds good

nixpanic:
@mergifyio rebase

nixpanic:
Rebased, @Madhu-1 please review again. just fixed one minor conflict in `internal/cephfs/volume.go` where `JoinErrors()` is used.

internal/rbd/rbd_attach.go
```diff
@@ -229,19 +229,19 @@ func createPath(ctx context.Context, volOpt *rbdVolume, cr *util.Credentials) (s
 		mapOptions = append(mapOptions, "--read-only")
 	}
 	// Execute map
-	output, err := execCommand(rbd, mapOptions)
+	stdout, stderr, err := util.ExecCommand(ctx, rbd, mapOptions...)
 	if err != nil {
-		klog.Warningf(util.Log(ctx, "rbd: map error %v, rbd output: %s"), err, string(output))
+		klog.Warningf(util.Log(ctx, "rbd: map error %v, rbd output: %s"), err, stderr)
```
humblec, line 234: it should have been stdout , Isnt it ?

internal/rbd/rbd_attach.go
```diff
@@ -229,19 +229,19 @@ func createPath(ctx context.Context, volOpt *rbdVolume, cr *util.Credentials) (s
...
 		// unmap rbd image if connection timeout
 		if strings.Contains(err.Error(), rbdMapConnectionTimeout) {
 			detErr := detachRBDImageOrDeviceSpec(ctx, imagePath, true, isNbd, volOpt.Encrypted, volOpt.VolID)
 			if detErr != nil {
 				klog.Warningf(util.Log(ctx, "rbd: %s unmap error %v"), imagePath, detErr)
 			}
 		}
-		return "", fmt.Errorf("rbd: map failed %v, rbd output: %s", err, string(output))
+		return "", fmt.Errorf("rbd: map failed %v, rbd output: %s", err, stderr)
```
humblec, line 242: same here?

internal/rbd/rbd_attach.go
humblec: interchably using stdout and stderr is confusing..

nixpanic: Before, stdout+stderr were combined into a single output. Now there is a split between stdout and stderr. Error messages would be placed in stderr, so that contains the valuable information for troubleshooting.

internal/rbd/rbd_attach.go
nixpanic: stdout would not contain the error message, so that would not help troubleshooting

internal/rbd/rbd_attach.go
humblec: sure. but in above warning we say here is "err:" and here is "output:" , then we use "err" and "stderr" variables, why? . Either using the` stderr `variable in output  is not correct or its a bad naming or the log message is not mentioning it correct.

internal/rbd/rbd_attach.go
humblec: If thats the case why we have "rbd output: %s" ?

internal/rbd/rbd_attach.go
nixpanic: stderr is output, just as stdout? There is no reference to stdout, so I am not sure where the confusion is. There is no output variable anymore.

The message contains `error: %err, output: %stderr`. Do you mean you want to have the variable names instead of error/output?

internal/rbd/rbd_attach.go
nixpanic: the rbd command wrote an error message on stderr, which I consider output? Do you suggest to have the log message `"rbd: map failed with error %v, rbd errror output: %s"` or something? Suggestions welcome.

humblec: `"rbd: map failed with error %v, rbd errror output: %s"` looks good to me .

nixpanic: Ok, updated and rebased. Please check again.

Madhu-1:
@nixpanic need rebase

nixpanic:
> This pull request now has conflicts with the target branch. Could you please resolve conflicts and force push the corrected changes? pray

Rebased, so reviews are needed again :roll_eyes:

Issue #1261, opened 2020-07-22
cephfs: pvc mount fails after csi upgrade


Yuggupta27:
# Describe the bug #

A PVC created in CSI v2.0.0 is unable to get mounted after upgrading to the latest release.

# Steps to reproduce #

1. Create a pvc in v2.0.0
1. upgrade to master
1. Bind the existing pvc to a pod.

# Actual results #
```
FailedMount: Unable to attach or mount volumes: unmounted volumes=[mypvc], unattached volumes=[mypvc default-token-2f2g4]: timed out waiting for the condition
```
# Expected behavior #

The pvc gets bound seamlessly.

# Logs #

cephfs plugin logs: https://termbin.com/z8wm
cephfs attacher logs: https://termbin.com/wtt0

# Additional context #

Also see: https://github.com/ceph/ceph-csi/pull/1243

Yuggupta27:
The Default Subvolumegroup name is `csi` in v2.0.0 and `e2e` now for the end to end testing.

Madhu-1:
This is really good Catch Thanks for catching this one, or else we would have introduced a regression for subvolumes groups

Madhu-1:
This issue exists for Expand or Deleting the older PVC,

Madhu-1:
CC @madddi The subvolumegroup is not backward compatible after upgrading the ceph-csi to lastest version if the user updates the current configmap with subvolume group name

Madhu-1:
This won't be a regression, This is working properly and we need to document that the user should not update the information for the current clusterID. for the clusterID which are used by the ceph-csi driver(<3.0.0 which is master) the subvolume should remain as it is, **we just need to document it in upgrade procedure and also we need to extend the documentation on how to configure a subvolume**, we should mention that the user **shouldnot/never** update the clusterID configuration(for subvolumegoup name) he should always create a new one.

Madhu-1:
@Yuggupta27 can you please take care of it in your E2E upgrade testing. and also please send a documentation update for this one.

Yuggupta27:
> @Yuggupta27 can you please take care of it in your E2E upgrade testing. and also please send a documentation update for this one.

Sure. I'll do that. Thanks for looking into it.

humblec:
@Yuggupta27 isnt this solved already ?

Madhu-1:
This is not fixed yet, we need to update the documentation

PR #1263, opened 2020-07-22
cephfs: fix regression in subvolumegroups

Madhu-1:
If the PVC is created by ceph-csi <v3.0.0 (master as of now) the subvolume will be default ie "csi" if the user updates the cephcsi
to v3.0.0 and have the subvolume group in configmap, remounting of older PVC will start failing as the ceph-csi will read the subvolumegroup from configmap and use it for mounting the PVC created cephcsi <3.0.0.

This commit fixes it by adding subvolume group name in the volume context and use it to get the root path of the subvolume.

closes https://github.com/ceph/ceph-csi/issues/1261

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -38,6 +38,11 @@ type ControllerServer struct {
 	VolumeLocks *util.VolumeLocks
 }
 
+var (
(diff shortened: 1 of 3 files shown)


Madhu-1:
This won't work for expanding older PVC also.

PR #1237, opened 2020-07-15
ci: update rook to 1.2.7 to make use of v14.2.10 ceph cluster

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

# Describe what this PR does #
    Rook version is currently 1.1.7 in our e2e deployment which brings 14.2.4 version
    of ceph cluster. To support cephfs snapshot e2e, we need latest version of Ceph Cluster
    in E2E. Rook 1.2.7 is good enough which on patching bring up ceph 14.2.10 cluster.

PR #1270, opened 2020-07-23
ci: fix rook cluster version fetching

humblec:
As part of https://github.com/ceph/ceph-csi/pull/1237/ there was
a patching enabled for the ceph cluster deployed, however due to
an error in the version fetching logic, the patching was not applied


scripts/rook.sh
```diff
@@ -7,7 +7,7 @@ ROOK_BLOCK_POOL_NAME=${ROOK_BLOCK_POOL_NAME:-"newrbdpool"}
 ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
 
 rook_version() {
-	echo "${ROOK_VERSION#?v}" | cut -d'.' -f"${1}"
+	echo "${ROOK_VERSION#?}" | cut -d'.' -f"${1}"
```
nixpanic, line 10: why leave the `#$` in here?

humblec: @nixpanic in case of first iteration on the string `v14.2.10` it was giving back `v1`.

nixpanic: Ah, yes. But in this case you are cutting the 1st character from the ROOK_VERSION, which may not be the `v` for all (future) release tags. It is a little safer to use `echo "${ROOK_VERSION#v}"` that will only drop an initial `v`.

humblec: Sure. done.. ptal

humblec:
@nixpanic @Madhu-1 ptal

PR #1271, opened 2020-07-24
cleanup: make godot warnings fixed in our source repo

humblec:

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -27,7 +27,7 @@ var (
 	rbdDeploymentName  = "csi-rbdplugin-provisioner"
 	rbdDaemonsetName   = "csi-rbdplugin"
 	defaultRBDPool     = "replicapool"
-	// Topology related variables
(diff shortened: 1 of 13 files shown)


humblec:
@Madhu-1  @nixpanic  @Yuggupta27 ptal

Yuggupta27:
Were these godot warnings reported by golangci? Afaik, only `Top level comments should end in a period`. Comments inside a block don't need to satisfy this condition.

humblec:
> Were these godot warnings reported by golangci? Afaik, only `Top level comments should end in a period`. Comments inside a block don't need to satisfy this condition.
> For reference see: [b7e4794](https://github.com/ceph/ceph-csi/commit/b7e4794713cd6ad7f6b11b81d7eba0d3a750eaa7)
True.. let me correct it .. Thanks :+1:

Madhu-1:
we don't need this one, just verified it with `godot` tool.

PR #1273, opened 2020-07-24
Migrate ci/centos branch to new OCP4 Jenkins deployment

nixpanic:
The CentOS team requested all the projects that use their CI to move to the new OCP4 cluster. At the moment, Ceph-CSI has a Jenkins deployment hosted on the old CentOS OCP-3.6 cluster.

This PR updates the links to Jenkins and the OCP console for administration of the Ceph-CSI namespace.

The new Jenkins deployment is not 100% ready yet, the changes in this PR will make it easier to start using it.

PR #1274, opened 2020-07-24
ci: require jjb-validate job to succeed for ci/centos branch

nixpanic:
The jjb-validate job can not be run in parallel, so it may fail when
multiple PRs for the ci/centos branch are sent. However, the ci/centos
branch is not very active, so problems should be minimal.

In case of problems, leave the following comment in the PR and the job
should restart:

    /retest ci/centos/jjb-validate

See-also: #1273


Yuggupta27:
This would be great for CentOS CI to have. Thanks.

PR #1295, opened 2020-07-30
rbd: return error if fetching cluster id fails

Madhu-1:
if we are not able to fetch the cluster-ID from the createSnapshot request and also if we are not able to get the monitor information 
 from the cluster-ID return error instead of using the parent image information.


internal/rbd/controllerserver.go
```diff
@@ -726,7 +726,10 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		return nil, status.Errorf(codes.InvalidArgument, "volume(%s) has not snapshot feature(layering)", req.GetSourceVolumeId())
 	}
 
-	rbdSnap := genSnapFromOptions(ctx, rbdVol, req.GetParameters())
+	rbdSnap, err := genSnapFromOptions(ctx, rbdVol, req.GetParameters())
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
```
nixpanic, line 731: Should there be different errors, depending on what goes wrong?

Madhu-1: Yeah, you are correct. for this case, the Mon check is already done earlier. we are safe here we don't need other errors. but we differently need to fix it in other places. it will be a good one for @yati1998 or @Yuggupta27 . let me open a separate issue for it.

Madhu-1: created issue for error check https://github.com/ceph/ceph-csi/issues/1301

PR #1305, opened 2020-08-03
doc: design document for rbd-mirroring

Madhu-1:
add a design doc for rbd based mirroring in ceph-csi.


docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,112 @@
...
+imply that NBD will need to be used instead of krbd.
+
+### Promotion of the rbd image
+
+During the **NodeStage request** from the CSI side, ceph-csi will promote the
+image and use it.
+
+If its a ROX or RWX PVC we need to handle image already mirrored case.
+
+Currently, **ceph-csi will not take care of demoting the rbd images, it will be
```
dillaman, line 81: I am still advocating for `ceph-csi` to automatically promote on `ControllerPublishVolume` and automatically demote on `ControllerUnpublishVolume`.

dillaman, line 81: If the current primary cluster is "dead" and never demotes, what's the timeout or operator intervention required to force `ceph-csi` to issue the equivalent of `rbd mirror image promote --force`?

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,112 @@
...
+someone else responsibility to demote the image first before promoting it on
+the remote cluster**.
+
+### For snapshot mirroring
+
+This will remain same as the rbd image mirroring in CreateVolume request, here
```
dillaman, line 87: Is this the end-user expectation? Note that RBD mirroring snapshots are not the same as user-snapshots and will be auto-deleted by the system when no longer required.

Madhu-1:
@MergifyIO rebase

Madhu-1:
@MergifyIO rebase

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,112 @@
...
+things like (information need to store and omap for image identification,
+finding rbd image based on PVC name and namespace name in the remote cluster)
+
+The value for the mirroring will be  either one of these "disabled" (default),
+"journal", or "snapshot"
+
+* "disabled" - (default) no need to mirror the rbd image
+* "journal"  - journal based mirroring. The mounter should be rnd-nbd as the
+  krbd doesnot support it yet.
+* "snapshot" - snapshot based mirroring. It should work with existing krbd
```
Madhu-1, line 45: Note to add here is that older PVC's cannot be mirrored

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+  to change the logic of storing/generating the unique ID based on the PVC name
+  and namespace name.
+
+* ceph-csi generated journal(omap data) is not mirrored to the remote cluster.
+
+* Promotion and demotion of RBD image.
+
+## Below provides the outline of how this can be implemented in ceph-csi
+
+* A storage-class Need to have a parameter to tell that the image needs to be
```
humblec, line 23: s/Need/need

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+  mirrored.
+* The PVC name and namespace need to be passed to the CSI driver as part of
```
humblec, line 25: This would be taken care by https://github.com/ceph/ceph-csi/pull/1093

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+  volume create request (rbd image and omap keys/value will be derived from
+  these names if its mirrored image.
+* The RBD mirror daemon required setup has to be done prior enabling the image
```
humblec, line 28: may be good to add a doc.

Madhu-1: This is outside of csi I don't think adding a doc is required for this one

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+required omap data and store its information and create the rbd image, the CSI
+will need to enable mirroring on the newly created RBD image via `rbd mirror
+image enable image-spec`.
+
+If it is a remote cluster the image will be already created and mirrored by the
+local cluster, we need to check for the image first(if it is present we need to
+generate the omap data and return success to the create volume (as its a remote
+cluster we don't need to enable mirroring here.
+
+Incase of journal based mirroring the image need to be craeted with
```
humblec, line 72: s/craeted/created

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+exclusive-lock and journaling image features. Additionally, This will also
```
humblec, line 73: s/This/this

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+imply that NBD will need to be used instead of krbd.
```
humblec, line 74: rbd-nbd mounter

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+During the **NodePublish request** from the CSI side, ceph-csi will
+promote the image and use it. and during **NodeUnpublish request** the ceph-csi
+will  demote the image and use it.
+
+as the NodePublish is per node so cephcsi  need to handle if the image is
+already promoted.
+
+But during **NodeUnpublish request** the ceph-csi should make sure that the rbd
+image is not attached on any node  before demoting the image as it can cause
+data lose. May be cephcsi can make use of volumeattachment kubernetes object to
```
humblec, line 87: my understanding was that, its not possible to `demote` if its already attached to a node. Isnt it @dillaman

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+check the pvc is attached to any nodes or not.
+
+If its a ROX or RWX PVC we need to handle image already mirrored case.
```
humblec, line 90: iic, ROX case is different. If image is ROX, ideally we may not have to worry about it, Isnt it ? because there is no data corruption can occur. I have one side question, is it possible or restriction to `promote` or `demote` a `READONLY` image?  @dillaman

docs/design/proposals/rbd-mirror.md
Madhu-1: Yes, that why we are making use of kubernetes volume attachment to make sure volume is not attached to any nodes.

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Incase of dead primary/remote cluster
+
+In case the ceph cluster is dead ceph-csi cannot demote the image to over come
+this issue we will provide a new flag (configuration) in cephcsi which will be
+set by the high level operator who is responsible for data migration will set
+the flag to `true` so that CSI can forcefull promote the image.
```
humblec, line 97: I understand this was the discussion we had. However we have to make sure the higher level operator should make sure there is a complete fence of the access from the other end to avoid data corruption and other issues due to this force enablement.

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+This will remain same as the rbd image mirroring in CreateVolume request, here
+instead of storageclass parameter we will check snapshotclass parameter.
+
+same as PVC name and namespace name to generate the rbd image name. we will use
+snapshot name and snapshot namespace  name to generate the rbd snapshot name
+and mirror the (intermediate rbd image we create  to the remote cluster),
+created an issue [2] in external snapshotter to pass extra snapshot metadata in
+create snapshot request.
+
+### Changes requried in cephcsi related to watchers
```
humblec, line 110: required.

Madhu-1:
@ShyamsundarR  PTAL

Madhu-1:
@travisn @leseb @agarwal-mudit FYI

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+we need to add a new parameter to the storage-class to indicate the image need
+to be mirrored upon creation, based on this parameter ceph-csi will decide few
+things like (information need to store and omap for image identification,
+finding rbd image based on PVC name and namespace name in the remote cluster)
+
+The value for the mirroring will be  either one of these "disabled" (default),
+"journal", or "snapshot"
+
+* "disabled" - (default) no need to mirror the rbd image
+* "journal"  - journal based mirroring. The mounter should be rnd-nbd as the
```
leseb, line 43:
```suggestion
* "journal"  - journal-based mirroring. The mounter should be rbd-nbd as the
```

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,137 @@
...
+* The PVC name and namespace need to be passed to the CSI driver as part of
+  volume create request (rbd image and omap keys/value will be derived from
+  these names if its mirrored image.
+* The RBD mirror daemon required setup has to be done prior enabling the image
+  mirroring from ceph-csi.
+* Promotion and Demotion of RBD image.
+
+### Storage-class parameter
+
+we need to add a new parameter to the storage-class to indicate the image need
```
leseb, line 34: It'd be nice to explain that we can have multiple SC serving the same pool but bringing different mirroring capabilities.

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+If the app and the PVC created in the local cluster as part of recovery the app
+and the PVC will be recreated in the remote cluster, the data present in the
+PVC in remote  cluster should be the same in the local cluster.
+
+As the PVC name and the namespace name should be sent in volume create request,
+cephcsi will check that the rbd image is already present cluster, we cannot
+check the omap data first as omap data maintained by ceph-csi is not mirrored
+to the remote cluster. This is the challenge ceph-csi need to handle)If the
+image is not present(means its a local cluster) we need to generate the
+required omap data and store its information and create the rbd image, the CSI
```
nixpanic, line 66: 'the CSI' - CSI is only an API specification, it can not do anything. What component should "enable mirroring"?

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+### Pass PVC name and namespace name in CreateVolume
+
+If the app and the PVC created in the local cluster as part of recovery the app
+and the PVC will be recreated in the remote cluster, the data present in the
+PVC in remote  cluster should be the same in the local cluster.
+
+As the PVC name and the namespace name should be sent in volume create request,
+cephcsi will check that the rbd image is already present cluster, we cannot
+check the omap data first as omap data maintained by ceph-csi is not mirrored
+to the remote cluster. This is the challenge ceph-csi need to handle)If the
```
nixpanic, line 64: 'handle)' -> 'handle.'

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+image is not present(means its a local cluster) we need to generate the
+required omap data and store its information and create the rbd image, the CSI
+will need to enable mirroring on the newly created RBD image via `rbd mirror
+image enable image-spec`.
+
+If it is a remote cluster the image will be already created and mirrored by the
+local cluster, we need to check for the image first(if it is present we need to
+generate the omap data and return success to the create volume (as its a remote
```
nixpanic, line 72: 2x opening (, this reads a little weird

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+cluster we don't need to enable mirroring here.
+
+Incase of journal based mirroring the image need to be created with
+exclusive-lock and journaling image features. Additionally, this will also
+imply that `rbd-nbd` mounter will need to be used instead of krbd.
```
nixpanic, line 77: Maybe list `requirements` or `restrictions` in the document?

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Promotion/Demotion of the rbd image
+
+During the **ControllerPublish request** from the CSI side, ceph-csi will
+promote the image and use it. and during **ControllerUnpublish request** the
+ceph-csi will demote the image and use it.
```
nixpanic, line 83: `will demote the image and use it`, after `ControllerUnpublish`? I do not understand how the image is used afterwards.

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+cause data lose. May be cephcsi can make use of volumeattachment kubernetes
+object to check the pvc is attached to any nodes or not.
+
+If its a ROX or RWX PVC we need to handle image already mirrored case.
+
+#### Incase of dead primary/remote cluster
+
+In case the ceph cluster is dead ceph-csi cannot demote the image to over come
+this issue we will provide a new flag (configuration) in cephcsi which will be
+set by the high level operator who is responsible for data migration will set
```
nixpanic, line 99: Which 'high level operator'? I think it is the first time you mention it in this doc. It should probably be explained a little more, and added to the diagram.

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
+# RBD MIRRORING
+
+![mirror](https://i.imgur.com/YF5WeCV.png)
+
+RBD images can be asynchronously mirrored between two Ceph clusters.
+
+## Challenges in ceph-csi
+
+* Currently, ceph-csi generates its unique ID for each rbd image and stores the
+  corresponding PV name and the unique ID mapping, and creates the rbd image
```
obnoxxx, line 10:
```suggestion
  corresponding PV name and the unique ID mapping. It creates the rbd image
```

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+  with the unique ID and return the encoded value which contains all the
```
obnoxxx, line 11:
```suggestion
  with the unique ID and returns the encoded value which contains all the
```

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+  required information for other operations. As the same rbd image needs to be
```
obnoxxx, line 12:
```suggestion
  required information for other operations. For mirroring, the same rbd image needs to be
```

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+  used in remote OCP cluster and the remote ceph cluster. In ceph-csi, we need
```
obnoxxx, line 13:
```suggestion
  used in remote OCP cluster and the remote ceph cluster. Therefore, we need
```

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+  to change the logic of storing/generating the unique ID based on the PVC name
```
obnoxxx, line 14:
```suggestion
  to change the ceph-csi logic of generating/storing the unique ID based on the PVC name
```

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
+# RBD MIRRORING
+
+![mirror](https://i.imgur.com/YF5WeCV.png)
```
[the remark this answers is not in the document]
Madhu-1: let me see i can do it

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,140 @@
...
+
+RBD images can be asynchronously mirrored between two Ceph clusters.
```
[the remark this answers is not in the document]
Madhu-1: not talking about the use-cases just mentioning about the core functionality. will see if can add a section for usecase

docs/design/proposals/rbd-mirror.md
Madhu-1: sure let me add that to the image I added. This is the operator sits on top of the rook

docs/design/proposals/rbd-mirror.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Changes required in cephcsi related to watchers
+
+currently, we are making use of watchers in node stage request we may need to
+change the watchers logic in the node stage request as when we enable the rbd
+mirroring on an image a watcher will be added on an rbd image.
+
+## Work in rook
+
+* Rook may need to provide an option to users (by exposing the CR) to configure
```
leseb, line 126: I think we need clearly identify the items Rook will take care of as well as the one from ceph-csi.

leseb: I think Rook will be able to provide the Ceph cluster details, I'm just not sure how much configuration it's is going to do. Especially if the replication is at the image level.
They are things like peer additions, I believe ceph-csi should take care of since it knows on which pools it is operating on.

humblec:
>  Therefore, we need to change the ceph-csi logic of generating/storing the
  unique ID based on the PVC name and namespace name.
>

On a general comment , I am not feeling comfortable to have this naming schema for CSI volumes with PVC name and Namespace . The main concern is that, the PVC can be "recreated"  ( with same name) in the same namespace with totally different set of parameters ! Could be different SC altogether..etc. 

For ex:
Suppose the first PVC was created with reclaimPolicy `retain`, and PVC is recreated with same name, we failed completely by referring to the 'OLD' RBD image, Isnt it ?

ShyamsundarR:
> > Therefore, we need to change the ceph-csi logic of generating/storing the
> > unique ID based on the PVC name and namespace name.
> 
> On a general comment , I am not feeling comfortable to have this naming schema for CSI volumes with PVC name and Namespace . The main concern is that, the PVC can be "recreated" ( with same name) in the same namespace with totally different set of parameters ! Could be different SC altogether..etc.
> 
> For ex:
> Suppose the first PVC was created with reclaimPolicy `retain`, and PVC is recreated with same name, we failed completely by referring to the 'OLD' RBD image, Isnt it ?

Recreating the same PVC within the same <name>-<namespace>, when the PV is left as "Retained" would have one of the following consequences,

1) A CreateVolume will be invoked with a request name that is unique. The steps executed would be,
  - The CSI journal will not find the reservation
  - CSI will start listing images with the `<name>-<namespace>` **prefix**, with the actual image being named as `<name>-<namespace>-<uuid>`
  - If an image with the prefix is found, its journal will be inspected and we will end up with a request name, which would be different than the new request
  - CSI would hence not link the image to the request as the image is already linked to a different request
  - CSI would further create a new image for the request, with the `<name>-<namespace>-<new-uuid>` image name

2) On the secondary cluster, if none of the PVs were created and a create request is invoked, it would need to determine that multiple images with the same prefix match the request that do not have a request name in the journal, and hence would have to error out

So in addition to (1) above, if we handle (2) we should not have issues with the presented use-case.

Going further out to another use-case pointed out by @johnstrunk where a PV namespace transfer has been initiated, and there is no PV that is "Retained" but image with the prefix exists as before, the loop would behave as above.

There maybe more corner cases in the PV namespace ownership transfer that may need attention though, and invalidate the name@namespace prefix based naming scheme.

Barring all these, the reason/per-requisite for `<name>-<namespace>` prefix is to ensure that the PVC request is the only one made across the clusters, no data from a provisioned PV is extracted and used in the secondary cluster for provisioning or static-PV creation needs. If that restriction is not mandated or necessary, then changing the scheme would possibly make it more deterministic.

NOTE: Also, as this design document stands, there are thoughts in flight that introduce a replication controller and possibly a replication CR, that also can be leveraged if the above per-requsite does not hold. More on this as it evolves though.

humblec:
> > > Therefore, we need to change the ceph-csi logic of generating/storing the
> > > unique ID based on the PVC name and namespace name.
> > 
> > 
> > On a general comment , I am not feeling comfortable to have this naming schema for CSI volumes with PVC name and Namespace . The main concern is that, the PVC can be "recreated" ( with same name) in the same namespace with totally different set of parameters ! Could be different SC altogether..etc.
> > For ex:
> > Suppose the first PVC was created with reclaimPolicy `retain`, and PVC is recreated with same name, we failed completely by referring to the 'OLD' RBD image, Isnt it ?
> 
> Recreating the same PVC within the same -, when the PV is left as "Retained" would have one of the following consequences,
> 
>     1. A CreateVolume will be invoked with a request name that is unique. The steps executed would be,
> 
> 
>     * The CSI journal will not find the reservation
> 
>     * CSI will start listing images with the `<name>-<namespace>` **prefix**, with the actual image being named as `<name>-<namespace>-<uuid>`
> 

Any way, It was *not* mentioned that, PVC name and namespace combination is just **prefix**, rather it was saying the unique ID based on the PVC name and Namespace. Thats why I was mentioning the mapping to the OLD image could happen in case of a `retain` scenario.

Madhu-1:
I am parking this design as of now and going with static PVC's. as of now, it's not suitable to go with pvc-name and namespace name. as we have new features like pvc transfer and data populators, etc.

nixpanic:
@Madhu-1 is this PR still relevant?

Madhu-1:
closing this one, as we are taking a different approach. wil send separate PR for it.

PR #1306, opened 2020-08-03
ci: use ceph-csi:ci/centos for fetching job scripts

nixpanic:
Commit f5cba3aaa830f added the mini-e2e job, but still referred to the
temporary location that was used for testing the job. As everything is
available in the ceph-csi:ci/centos repository:branch, there is no need
to refer to the temporary location.

### DIFF
--- a/mini-e2e.groovy
+++ b/mini-e2e.groovy
@@ -1,8 +1,7 @@
 def cico_retries = 16
 def cico_retry_interval = 60
-// temporary git repository for testing purpose
-def ci_git_repo = 'https://github.com/nixpanic/ceph-csi'
-def ci_git_branch = 'mini-e2e'
+def ci_git_repo = 'https://github.com/ceph/ceph-csi'
+def ci_git_branch = 'ci/centos'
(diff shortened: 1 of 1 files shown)


Yuggupta27:
Looks Great! Thanks @nixpanic

Yuggupta27:
Please review @Madhu-1

mini-e2e.groovy
```diff
@@ -1,8 +1,7 @@
 def cico_retries = 16
 def cico_retry_interval = 60
-// temporary git repository for testing purpose
-def ci_git_repo = 'https://github.com/nixpanic/ceph-csi'
-def ci_git_branch = 'mini-e2e'
+def ci_git_repo = 'https://github.com/ceph-csi/ceph-csi'
```
Madhu-1, line 3: it should be `https://github.com/ceph/ceph-csi` if yes can we reuse  the git_repo?

nixpanic: yes, of course!

Yuggupta27: In case the script is used for a different repo, using the same variable may make it fail.

PR #1312, opened 2020-08-04
ci: make number of CPUs for minikube VM configurable

nixpanic:
By default minikube uses 2 CPUs, which might be too little for some of
the tests. When not passing a CPUS environment variable, use all CPUs
available on the system (detected with 'nproc').


Yuggupta27:
Looks Great. Thanks!

PR #1313, opened 2020-08-04
ci: give minikube 8GB RAM on CentOS bare-metal systems

nixpanic:
All bare-metal systems in the CentOS CI have 16GB or more RAM. To
improve performance of the minikube VM, give is 8GB RAM.

See-also: https://wiki.centos.org/QaWiki/PubHardware

### DIFF
--- a/single-node-k8s.sh
+++ b/single-node-k8s.sh
@@ -64,6 +64,7 @@ function set_env() {
     export GO111MODULE="on"
     export TEST_COVERAGE="stdout"
     export VM_DRIVER="kvm2"
+    export MEMORY="8192"
     export CEPH_CSI_RUN_ALL_TESTS=true
     # downloading rook images is sometimes slow, extend timeout to 15 minutes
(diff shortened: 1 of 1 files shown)


single-node-k8s.sh
```diff
@@ -64,6 +64,7 @@ function set_env() {
     export GO111MODULE="on"
     export TEST_COVERAGE="stdout"
     export VM_DRIVER="kvm2"
+    export MEMORY="8192"
```
Madhu-1, line 67: @nixpanic can you also increase the cpu for minikube vm

nixpanic: yes, increased by default through #1312

Yuggupta27:
Can we also use RAM like we are currently with CPU? (Automatically detecting and assigning)

Yuggupta27:
We can then give half the RAM to the minikube depending on which instance gets assigned(As the majority of instances have 32GB ram)

PR #1317, opened 2020-08-05
ci: include build.env variables when installing Helm

nixpanic:
By default the install-helm.sh script uses "latest" as version for Helm.
Unfortunately this version does not exist. The HELM_VERSION variable is
already set in build.env, so source the configuration file as one of the
first actions in install-helm.sh.

### DIFF
--- a/scripts/install-helm.sh
+++ b/scripts/install-helm.sh
@@ -4,12 +4,15 @@
 
 TEMP="/tmp/cephcsi-helm-test"
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+# shellcheck source=build.env
+[ ! -e "${SCRIPT_DIR}"/../build.env ] || source "${SCRIPT_DIR}"/../build.env
+
 HELM="helm"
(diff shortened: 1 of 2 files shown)


scripts/install-helm.sh
```diff
@@ -4,12 +4,15 @@
 
 TEMP="/tmp/cephcsi-helm-test"
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+# shellcheck source=build.env
+[ ! -e "${SCRIPT_DIR}"/build.env ] || source "${SCRIPT_DIR}"/build.env
```
nixpanic, line 9: this is wrong, the build.env file is in `SCRIPT_DIR/../`

Madhu-1: removed ready-to-merge label

Madhu-1: This is correct isnt it. the build.env is in main directory

Madhu-1: sorry i missed this one, now its correct as build.env is outside of script dir

nixpanic:
once #1318 is merged, tests can be restarted and should :crossed_fingers: succeed

PR #1335, opened 2020-08-07
ci: add commitlint fallback job

nixpanic:
Each PR gets tested by the commitlint app (which is deprecated). On
occasion the app does not pickup the change in a PR, and the only way to
start a retest is to re-push the PR. All tests will then get executed
again, even if only commitlint was needed.

The new commitlint CI job introduces '/retest commitlint' and runs in
the CentOS CI like other Jenkins jobs. However, the job does not start
automatically on each (updated) PR, but needs manual starting with the
'/test commitlint' or '/retest commitlint' comments. The normal way of
running commitlint is with the app (at least for now).

### DIFF
--- a/commitlint.groovy
+++ b/commitlint.groovy
@@ -1,0 +1,53 @@
+def cico_retries = 16
+def cico_retry_interval = 60
+def ci_git_repo = 'https://github.com/ceph/ceph-csi'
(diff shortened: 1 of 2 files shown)


humblec:
Thanks

Issue #1242, opened 2020-07-16
 Tracker to fix todo in #394 CephFS  snapshot/Clone PR 


humblec:
This issue is for tracking the TODO which needs to be fixed after merging #394 , this will be closed after fixing all the todo's
for easier tracking am adding the TODO list here from the PR

- [x]  Update Documentation
<del> Describing e2e documentation on snapshot create,delete, restore and PVC clone
<del>Update support matrix to describe  which ceph versions are supported 
- [x] Plugin restart collecting cluster information (subvolume info support.etc) by  code **( Deferred to v3.2.0)**
https://github.com/ceph/ceph-csi/pull/394#discussion_r457558680
- [x] <del>Address total snapshots current limit (400) (**via documentation)

- [x]  <del>Add backend validation for CephFS subvolumes in E2E
*) Also address https://github.com/ceph/ceph-csi/pull/394#discussion_r464251501 ( both cephfs and rbd)

- [x] <del>Move `getMonsAndClusterID` to util as both RBD and CephFS use it
- [x] Revalidate the error strings and CSI error returns. ( **Deferred to 3.2.0)**

Madhu-1:
@humblec Adding E2E for TODO doesn't seem correct. Without E2E in #394 how to ensure that everything works fine?

humblec:
> @humblec Adding E2E for TODO doesn't seem correct. Without E2E in #394 how to ensure that everything works fine?

Sure, the rook cluster upgrade made me to put it here in first place, but removing it from the list as we need to sort this out any way. thanks ! :+1:

Madhu-1:
- [ ] Add test in upgrade e2e to create snapshot and clone from older PVC

Madhu-1:
@humblec can you please add documentation for 400 snapshot limit in release notes

Madhu-1:
> @humblec can you please add documentation for 400 snapshot limit in release notes

@humblec FYI this also needs to be updated  in current cephfs snapshot doc

Madhu-1:
> Revalidate the error strings and CSI error returns. ( Deferred to 3.2.0)

as this is deferred to 3.2.0 moving this to next milestone

humblec:
Closing this one for now as these are already addressed.

PR #1344, opened 2020-08-10
cephfs: add maximum snapshot check on subvolume

### LINKED ISSUE #1133: Limits for CephFS snapshots
The limit is in krbd (and kernel CephFS) since they only allocate 1 4KiB page to handle all the snapshot ids for an image / file. 

The snapshot limit only counts for the image where the snapshot actually exists -- it does not apply to the total number of snapshots in the entire grandparent-parent-child hierarchy.

_Originally posted by @dillaman in https://github.com/ceph/ceph-csi/issues/1098#issuecomment-638969666_

Teeing off from the above comment, we would need to handle this for CephFS as well.

As we do not have any flatten or related operations to reduce the snapshots for a given subvolume, the determined maximum number of snapshots for CephFS would be a hard limit, post which we would need to return RESOURCE_EXHAUSTED errors till some older snapshots are deleted for CreateSnapshot calls.

Cloning volumes from snapshots should not be a concern, as that is a full copy of the source volume. Thus, limits to the same will not apply.

There maybe a limit on how many clone operations are in flight though, for resource consumption reasons.

Madhu-1:
we cannot have more than 400 active snapshots on a single subvolume due to the kernel limitation.  we need to restrict the users creating more snapshots on a single subvolume during CreateSnapshot and PVC to PVC cloning.
    
fixes #1133
updates #1242

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -48,6 +48,14 @@ type ControllerServer struct {
 	OperationLocks *util.OperationLock
 }
 
+const (
+	// maximumSnapshotsOnSubVolume represents the maximum number of snapshots
+	// allowed on a single subvolume.
+	// Kernel client only supports up to 400 snapshots per filesystem
+	// see https://tracker.ceph.com/issues/21420.
+	maximumSnapshotsOnSubVolume = 400
+)
+
 // createBackingVolume creates the backing subvolume and on any error cleans up any created entities.
 func (cs *ControllerServer) createBackingVolume(
 	ctx context.Context,
@@ -543,6 +551,14 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		}, nil
 	}
 
+	snaps, err := listSnapshots(ctx, parentVolOptions, cr, volumeID(vid.FsSubvolName))
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if len(snaps) > maximumSnapshotsOnSubVolume {
+		return nil, status.Errorf(codes.ResourceExhausted, "for subvolume %s snaphot count %d reached the maximum %d", vid.FsSubvolName, len(snaps), maximumSnapshotsOnSubVolume)
+	}
 	// Reservation
 	sID, err := reserveSnap(ctx, parentVolOptions, vid.FsSubvolName, cephfsSnap, cr)
 	if err != nil {
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -245,6 +245,18 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}()
 
+	// check for number of snapshots for volume-volume cloning
+	if pvID != nil && parentVol != nil {
+		snaps, err := listSnapshots(ctx, parentVol, cr, volumeID(pvID.FsSubvolName))
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		if len(snaps) > maximumSnapshotsOnSubVolume {
+			return nil, status.Errorf(codes.ResourceExhausted, "for subvolume %s snaphot count %d reached the maximum %d", pvID.FsSubvolName, len(snaps), maximumSnapshotsOnSubVolume)
+		}
+	}
+
 	// Create a volume
 	err = cs.createBackingVolume(ctx, volOptions, parentVol, vID, pvID, sID, cr)
 	if err != nil {
--- a/internal/cephfs/snapshot.go
+++ b/internal/cephfs/snapshot.go
@@ -238,3 +238,40 @@ func cloneSnapshot(ctx context.Context, parentVolOptions *volumeOptions, cr *uti
 	}
 	return nil
 }
+
+// Snapshots holds the snapshot name.
+type Snapshots struct {
+	Name string `json:"name"`
+}
+
+func listSnapshots(ctx context.Context, volOptions *volumeOptions, cr *util.Credentials, volID volumeID) ([]Snapshots, error) {
+	snaps := []Snapshots{}
+	args := []string{
+		"fs",
+		"subvolume",
+		"snapshot",
+		"ls",
+		volOptions.FsName,
+		string(volID),
+		"--group_name",
+		volOptions.SubvolumeGroup,
+		"-m", volOptions.Monitors,
+		"-c", util.CephConfigPath,
+		"-n", cephEntityClientPrefix + cr.ID,
+		"--keyfile=" + cr.KeyFile,
+		"--format=json",
+	}
+	err := execCommandJSON(
+		ctx,
+		&snaps,
+		"ceph",
+		args[:]...)
+	if err != nil {
+		if strings.Contains(err.Error(), ErrSnapNotFound.Error()) {
+			return []Snapshots{}, err
+		}
+		klog.Errorf(util.Log(ctx, "failed to list subvolume snapshots %s(%s) in fs %s"), string(volID), err, volOptions.FsName)
+		return []Snapshots{}, err
+	}
+	return snaps, nil
+}


humblec:
@Madhu-1 one general comment here is, this is getting addressed in upstream (https://tracker.ceph.com/issues/46074), either we can wait and adopt when its available which is soon.  Additionally fetching the entire snapshot list and validating based on the count looks to be really heavy in the transaction. My thought here is, if it reaches a limit and no longer allowed to create snapshot we get an error code ( I am still fetching whats that error code though :) may be @kotreshhr can help us ) , ideally we could make use of it and that would also give some atomicity on the transaction itself and avoid some false positive scenarios with listSnapshots 

@ShyamsundarR thoughts?

Madhu-1:
> @Madhu-1 one general comment here is, this is getting addressed in upstream (https://tracker.ceph.com/issues/46074), either we can wait and adopt when its available which is soon.

waiting is not a good idea people will face the issues if they create more snapshots

> But, having a check till then is also good  ,Additionally fetching the entire snapshot list and validating based on the count looks to be really heavy in the transaction. My thought here is, if it reaches a limit and no longer allowed to create snapshot we get an error code ( I am still fetching whats that error code though :) may be @kotreshhr can help us ) , 

Any other suggestions from your side instead of fetching the snapshot lists?

AFAIK cephfs does not return any error if you create more than 400 snapshots (the issue will come when you are mounting the subvolume as the limit is in cephfs kernel client)

>ideally we could make use of it and that would also give some atomicity on the transaction itself and avoid some false positive scenarios with listSnapshots
> 
> @ShyamsundarR thoughts?

humblec:
> > @Madhu-1 one general comment here is, this is getting addressed in upstream (https://tracker.ceph.com/issues/46074), either we can wait and adopt when its available which is soon.
> 
> waiting is not a good idea people will face the issues if they create more snapshots

yeah.. thats also true.

> 
> > But, having a check till then is also good  ,Additionally fetching the entire snapshot list and validating based on the count looks to be really heavy in the transaction. My thought here is, if it reaches a limit and no longer allowed to create snapshot we get an error code ( I am still fetching whats that error code though :) may be @kotreshhr can help us ) ,
> 
> Any other suggestions from your side instead of fetching the snapshot lists?

Yeah, thats what I was trying to figure out. In the tracker issue (#1133) @ShyamsundarR has proposed RESOURCE_EXHAUSTED error code when we reach the limit. Not sure this is already available/addressed in CephFS.  Have to figure out whats the return or error from CephFS when we currently reach this limit. We could make use of it as a solution here too if its already addressed. Otherwise , we are lmiited to below options I believe:

1) Fetch snapshot list and validate the count ( as in the PR, but if we have many snapshots, getting the entire list would be a time consuming call I believe, thus some performance issues..  )
2) Having a snapshot count updated for subvolume ( volumeIdentifier)  with create and delete of snapshot on it, making use of it while we create a snapshot  ? ( have to think through it though )

Madhu-1:
> > > @Madhu-1 one general comment here is, this is getting addressed in upstream (https://tracker.ceph.com/issues/46074), either we can wait and adopt when its available which is soon.
> > 
> > 
> > waiting is not a good idea people will face the issues if they create more snapshots
> 
> yeah.. thats also true.
> 
> > > But, having a check till then is also good  ,Additionally fetching the entire snapshot list and validating based on the count looks to be really heavy in the transaction. My thought here is, if it reaches a limit and no longer allowed to create snapshot we get an error code ( I am still fetching whats that error code though :) may be @kotreshhr can help us ) ,
> > 
> > 
> > Any other suggestions from your side instead of fetching the snapshot lists?
> 
> Yeah, thats what I was trying to figure out. In the tracker issue (#1133) @ShyamsundarR has proposed RESOURCE_EXHAUSTED error code when we reach the limit. Not sure this is already available/addressed in CephFS. Have to figure out whats the return or error from CephFS when we currently reach this limit. We could make use of it as a solution here too if its already addressed. Otherwise , we are lmiited to below options I believe:

There is nothing that currently exists in the code.  This PR returns the error code what @ShyamsundarR is suggesting

> 
> 1. Fetch snapshot list and validate the count ( as in the PR, but if we have many snapshots, getting the entire list would be a time consuming call I believe, thus some performance issues..  )
> 2. Having a snapshot count updated for subvolume ( volumeIdentifier)  with create and delete of snapshot on it, making use of it while we create a snapshot  ? ( have to think through it though )

am not sure that the proposing code could cause issues as the complete list is sent in one network call (which contains only names).

humblec:
> > > > @Madhu-1 one general comment here is, this is getting addressed in upstream (https://tracker.ceph.com/issues/46074), either we can wait and adopt when its available which is soon.
> > > 
> > > 
> > > waiting is not a good idea people will face the issues if they create more snapshots
> > 
> > 
> > yeah.. thats also true.
> > > > But, having a check till then is also good  ,Additionally fetching the entire snapshot list and validating based on the count looks to be really heavy in the transaction. My thought here is, if it reaches a limit and no longer allowed to create snapshot we get an error code ( I am still fetching whats that error code though :) may be @kotreshhr can help us ) ,
> > > 
> > > 
> > > Any other suggestions from your side instead of fetching the snapshot lists?
> > 
> > 
> > Yeah, thats what I was trying to figure out. In the tracker issue (#1133) @ShyamsundarR has proposed RESOURCE_EXHAUSTED error code when we reach the limit. Not sure this is already available/addressed in CephFS. Have to figure out whats the return or error from CephFS when we currently reach this limit. We could make use of it as a solution here too if its already addressed. Otherwise , we are lmiited to below options I believe:
> 
> There is nothing that currently exists in the code. This PR returns the error code what @ShyamsundarR is suggesting

> 
> > 1. Fetch snapshot list and validate the count ( as in the PR, but if we have many snapshots, getting the entire list would be a time consuming call I believe, thus some performance issues..  )
> > 2. Having a snapshot count updated for subvolume ( volumeIdentifier)  with create and delete of snapshot on it, making use of it while we create a snapshot  ? ( have to think through it though )
> 
> am not sure that the proposing code could cause issues as the complete list is sent in one network call (which contains only names).

sure.. just wanted to bring up and figure out, is there a better method here to tackle this. 

Approving it, but better to  wait for @ShyamsundarR's thought here too.

kotreshhr:
> @Madhu-1 one general comment here is, this is getting addressed in upstream (https://tracker.ceph.com/issues/46074), either we can wait and adopt when its available which is soon. But, having a check till then is also good +1 ,Additionally fetching the entire snapshot list and validating based on the count looks to be really heavy in the transaction. My thought here is, if it reaches a limit and no longer allowed to create snapshot we get an error code ( I am still fetching whats that error code though :) may be @kotreshhr can help us ) , ideally we could make use of it and that would also give some atomicity on the transaction itself and avoid some false positive scenarios with listSnapshots

I have just validated that if per directory snapshot limit is exceeded, cephfs would error out with EMLINK (too many links).
I have not tried the 400 snapshot limit across subvolumes yet.
> 
> @ShyamsundarR thoughts?

humblec:
> EMLINK

Thats great! Thanks for the input @kotreshhr ! Considering we have the directory with us, it would be a good idea to check for this Error Code I believe.

Madhu-1:
> > EMLINK
> 
> Thats great! Thanks for the input @kotreshhr ! Considering we have the directory with us, it would be a good idea to check for this Error Code I believe.

as per the issue https://github.com/ceph/ceph-csi/issues/1133 if the snapshot count goes above 400 we are not able to mount the subvolumes due to the ceph kernel issue. @ShyamsundarR  am i missing anything here. 

ceph fs returns `EMLINK`  when we are not able to create more snapshots @kotreshhr am I correct. is there any limit what is the maximum number of snapshots I can create on a single subvolume?
as per @ShyamsundarR https://github.com/ceph/ceph-csi/issues/1133#issuecomment-639842149 he was able to create 550 snapshots per single subvolume. in this PR am trying to address the issue related to kernel mounting (which will occur if we have more than 400 snapshots per single  subvolume)

kotreshhr:
> > > EMLINK
> > 
> > 
> > Thats great! Thanks for the input @kotreshhr ! Considering we have the directory with us, it would be a good idea to check for this Error Code I believe.
> 
> as per the issue #1133 if the snapshot count goes above 400 we are not able to mount the subvolumes due to the ceph kernel issue. @ShyamsundarR am i missing anything here.
> 
> ceph fs returns `EMLINK` when we are not able to create more snapshots @kotreshhr am I correct. is there any limit what is the maximum number of snapshots I can create on a single subvolume?

I validated max snapshots per directory limit (default 100) i.e., max snapshots per subvolume, where EMLINK is returned. I have not validated max snapshots across subvolume i.e., 400.

> as per @ShyamsundarR [#1133 (comment)](https://github.com/ceph/ceph-csi/issues/1133#issuecomment-639842149) he was able to create 550 snapshots per single subvolume. in this PR am trying to address the issue related to kernel mounting (which will occur if we have more than 400 snapshots per single subvolume)

humblec:
> > > > EMLINK
> > > 
> > > 
> > > Thats great! Thanks for the input @kotreshhr ! Considering we have the directory with us, it would be a good idea to check for this Error Code I believe.
> > 
> > 
> > as per the issue #1133 if the snapshot count goes above 400 we are not able to mount the subvolumes due to the ceph kernel issue. @ShyamsundarR am i missing anything here.
> > ceph fs returns `EMLINK` when we are not able to create more snapshots @kotreshhr am I correct. is there any limit what is the maximum number of snapshots I can create on a single subvolume?
> 
> I validated max snapshots per directory limit (default 100) i.e., max snapshots per subvolume, where EMLINK is returned. I have not validated max snapshots across subvolume i.e., 400.
> 
> > as per @ShyamsundarR [#1133 (comment)](https://github.com/ceph/ceph-csi/issues/1133#issuecomment-639842149) he was able to create 550 snapshots per single subvolume. in this PR am trying to address the issue related to kernel mounting (which will occur if we have more than 400 snapshots per single subvolume)

IN short, we could make use of `EMLINK` I believe.

kotreshhr:
> > > > > EMLINK
> > > > 
> > > > 
> > > > Thats great! Thanks for the input @kotreshhr ! Considering we have the directory with us, it would be a good idea to check for this Error Code I believe.
> > > 
> > > 
> > > as per the issue #1133 if the snapshot count goes above 400 we are not able to mount the subvolumes due to the ceph kernel issue. @ShyamsundarR am i missing anything here.
> > > ceph fs returns `EMLINK` when we are not able to create more snapshots @kotreshhr am I correct. is there any limit what is the maximum number of snapshots I can create on a single subvolume?
> > 
> > 
> > I validated max snapshots per directory limit (default 100) i.e., max snapshots per subvolume, where EMLINK is returned. I have not validated max snapshots across subvolume i.e., 400.
> > > as per @ShyamsundarR [#1133 (comment)](https://github.com/ceph/ceph-csi/issues/1133#issuecomment-639842149) he was able to create 550 snapshots per single subvolume. in this PR am trying to address the issue related to kernel mounting (which will occur if we have more than 400 snapshots per single subvolume)
> 
> IN short, we could make use of `EMLINK` I believe.

I think no. This is not that simple. There are two limits here.

1. Snapshots per directory: This returns EMLINK while creating snapshot on exceeding this limit.
2. Max kernel snapshot limit: This is 400 as of now. I think, this doesn't return error during snapshot creation and problem appears only in later stages like during mount...

So if the per directory limit is set to say 500, we will have to work within the max kernel snapshot limit i.e., 400. And since exceeding that won't error out as per [1] and [2],  I don't think we can rely on errno here. @ShyamsundarR please add if I have missed anything.

[1] https://github.com/ceph/ceph-csi/issues/1133
[2] http://tracker.ceph.com/issues/21420

humblec:
>
So if the per directory limit is set to say 500, we will have to work within the max kernel snapshot limit i.e., 400. 
>
One question here, if there is a limit of 400 across subvolume snapshots per FS , is it possible to configure a value higher than 400 in ` mds_max_snaps_per_dir` ? if yes, does that configuration make sense ? or its completely depend on the mounter, so we are still allowing it ?

humblec:
@kotreshhr @Madhu-1 out of all discussion, I feel relying on `EMLINK` is not possible and there is NO proper error code if the limit exceeded from CephFS side. So, we are pretty much limited to do listSnapshot and validate the count. So approval stay here. However will wait for @ShyamsundarR too for his thoughts :+1:

internal/cephfs/snapshot.go
```diff
@@ -238,3 +238,40 @@ func cloneSnapshot(ctx context.Context, parentVolOptions *volumeOptions, cr *uti
 	}
```
nixpanic, line 238: commenting on the commit message:

    subvolume firts. This commits adds

firts -> first
commits -> commit

internal/cephfs/controllerserver.go
```diff
@@ -48,6 +48,12 @@ type ControllerServer struct {
 	OperationLocks *util.OperationLock
 }
 
+var (
+	// maximumSnapshotsOnSubVolume represents the maximum number of snapshots
+	// allowed on a single subvolume.
+	maximumSnapshotsOnSubVolume = 400
```
nixpanic, line 54: should this not be a `const`?

internal/cephfs/controllerserver.go
```diff
@@ -243,6 +243,20 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		}
 	}()
 
+	// check for number of snapshots for volume-volume cloning
+	if pvID != nil && parentVol != nil {
+		// Kernel client only supports up to 400 snapshots per filesystem
+		// see https://tracker.ceph.com/issues/21420
```
nixpanic, line 249: you explain this 2x now, maybe place the link at the `maximumSnapshotsOnSubVolume` definition?

humblec:
@nixpanic ptal.

ShyamsundarR:
afaict, the 400 limitation is for total snapshots on the FS instance and not per subvolume/directory. Hence counting snapshots per subvolume and limiting it to 400 is not going to prevent the issue from occurring.

If we need to ensure we stay below this limit till this is fixed in CephFS, we need to count all snapshots across all subvolumes in the CephFS instance.

Madhu-1:
> afaict, the 400 limitation is for total snapshots on the FS instance and not per subvolume/directory. Hence counting snapshots per subvolume and limiting it to 400 is not going to prevent the issue from occurring.
> 
> If we need to ensure we stay below this limit till this is fixed in CephFS, we need to count all snapshots across all subvolumes in the CephFS instance.

This seems to be a heavy operation, it's better to declare it as a known BUG in cephfs snapshot and clone functionality in that case? @ShyamsundarR any thoughts?

humblec:
> afaict, the 400 limitation is for total snapshots on the FS instance and not per subvolume/directory. Hence counting snapshots per subvolume and limiting it to 400 is not going to prevent the issue from occurring.
> 
> If we need to ensure we stay below this limit till this is fixed in CephFS, we need to count all snapshots across all subvolumes in the CephFS instance.

Thanks @ShyamsundarR for the input.  Yeah! so counting 400 per subvolume may not be enough. Have to check is there a way to list the entire snapshot list from an "FS" instead of Subvolume ( there is also FS snapshot , so I am not sure we could get a single command to fetch snapshots of all subvolumes on a FS) . If yes, we could make use of it.

humblec:
> > afaict, the 400 limitation is for total snapshots on the FS instance and not per subvolume/directory. Hence counting snapshots per subvolume and limiting it to 400 is not going to prevent the issue from occurring.
> > If we need to ensure we stay below this limit till this is fixed in CephFS, we need to count all snapshots across all subvolumes in the CephFS instance.
> 
> This seems to be a heavy operation, it's better to declare it as a known BUG in cephfs snapshot and clone functionality in that case? @ShyamsundarR any thoughts?

yeah, its heavy.  IMO, documentation is good enough for now as we may get solved this soon in CephFS with the mentioned tracker issue. Till then we can survive I believe. :+1:

Madhu-1:
Am closing this PR as it cannot be handled in CSI

PR #1330, opened 2020-08-07
cleanup: use common logging functions in the util package

nixpanic:
# Describe what this PR does #

Introduce additional log functions in the util package, so that `klog.Errorf()`, `klog.Fatalf()` and `klog.Warningf()` can be replaced by our own common logging functions. In the end, this makes us less dependent on the klog package, so upgrading becomes easier.

## Future concerns ##

Other parts of the code need replacing the klog package too. This can be done is multiple follow-up PRs.

PR #1345, opened 2020-08-10
ci: always fetch GIT_SINCE for commitlint

nixpanic:
When running the commitlint CI job, the branch that the PR is based on
may not be available. That makes it impossible for commitlint to detect
the changes between the HEAD of the branch, and the commits in the PR.

By fetching GIT_SINCE unconditionally, commitlint should be able to
detect the changes and only run the tests against the commits that were
added through the PR.


nixpanic:
once merged, this can be tried out on #1330

PR #1346, opened 2020-08-10
ci: make members of the GitHub "ceph" organization admins

nixpanic:
Not everyone can restart CI jobs, only members of the Ceph organization
in GitHub. All jibs (except "commitlint") are started automatically, so
there is no functional change.


nixpanic:
Without this, nobody can run the `commitlint` CI job :disappointed:

Issue #1357, opened 2020-08-11
Use scripts/skip-doc-change.sh to not run certain CI jobs


nixpanic:
# Describe the feature you'd like to have #

All jobs in the CentOS CI run on all PRs. This is not needed, as some PRs only make changes to documentation or some files that are not used for testins, deployment or runtime.

# What is the value to the end user? (why is it a priority?) #

By not running all CI jobs (or return success without completely running), affected PRs will be able to get merged much faster.

# Additional context #

Jobs in Travis CI (see `.travis.yaml`) check for a document-only change, and exist the test with success. This is something that could be done in the CentOS CI jobs too.

nixpanic:
This has been addressed with #1369

PR #1369, opened 2020-08-12
ci: skip CI jobs when a PR contains a doc-only change

nixpanic:
Adapt the Jenkins Pipelines to check for a doc-only change. If that is the case, there is no need to reserve a bare-metal system and run the tests.

The CI jobs will run, but get aborted (with `SUCCESS`) in an early stage. That means the PR status context will get a green :heavy_check_mark: for the jobs.

Closes: #1357

### DIFF
--- a/containerized-tests.groovy
+++ b/containerized-tests.groovy
@@ -2,7 +2,11 @@ def cico_retries = 16
 def cico_retry_interval = 60
 def ci_git_repo = 'https://github.com/ceph/ceph-csi'
 def ci_git_branch = 'ci/centos'
+def git_repo = 'https://github.com/ceph/ceph-csi'
 def ref = "master"
+def git_since = 'master'
+def workdir = '/opt/build/go/src/github.com/ceph/ceph-csi'
+def doc_change = 0
 
 node('cico-workspace') {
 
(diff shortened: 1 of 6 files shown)


containerized-tests.groovy
```diff
@@ -12,6 +16,29 @@ node('cico-workspace') {
...
+		sh "cd ~/build/ceph-csi && git fetch origin ${ref} && git checkout -b ${ref} FETCH_HEAD"
+	}
+
+	stage('check doc-only change') {
+		doc_change = sh(
+			script: "cd ~/build/ceph-csi && \${OLDPWD}/scripts/skip-doc-change.sh origin/${git_since}",
+			returnStatus: true)
+	}
+	// if doc_change (return value of skip-doc-change.sh is 1, do not run the other stages
+	if (doc_change == 1) {
```
Madhu-1, line 37: Is doc_change should be called as "{doc_change}"

mini-e2e-helm.groovy
```diff
@@ -4,6 +4,7 @@ def ci_git_repo = 'https://github.com/ceph/ceph-csi'
 def ci_git_branch = 'ci/centos'
 def git_repo = 'https://github.com/ceph/ceph-csi'
 def ref = "master"
+def doc_change = 0
```
nixpanic, line 7: this is wrong, it seems to be missing `git_since`

containerized-tests.groovy
nixpanic: no, only when embedding in strings

scripts/skip-doc-change.sh
```diff
@@ -0,0 +1,38 @@
...
+[[ -z $CHANGED_FILES ]] && exit 1
+
+skip=0
+#files to be skipped
+declare -a FILES=(^docs/ .md$ ^scripts/ LICENSE .mergify.yml .github .gitignore)
+
+function check_file_present() {
+    local file=$1
+    for FILE in "${FILES[@]}"; do
+        if [[ $file =~ $FILE ]]; then
```
Madhu-1, line 20: There are few others files are present in script folder will this check be enough?

nixpanic: This script is a copy from the original/master `scripts/skip-doc-change.sh` with minor changes. If it needs to be adapted, please file an issue with your concerns.

nixpanic:
The CI job is stuck posting the `SUCCESS` status :disappointed:

PR #1374, opened 2020-08-13
ci: use build.env:CSI_IMAGE_VERSION for mini-e2e jobs

nixpanic:
By placing CSI_IMAGE_VERSION=v3.1.0 in build.env in the release-v3.1
branch, the CI jobs using single-node-k8s.sh can pickup the version
dynamically.

Other release branches can do the same, and the CI jobs will not need
modifications for each release.

### DIFF
--- a/single-node-k8s.sh
+++ b/single-node-k8s.sh
@@ -61,6 +61,7 @@ function set_env() {
     # it and will cause a failure if not annotated with /dev/null.
     # shellcheck source=/dev/null
     source "${GOPATH}"/src/github.com/ceph/ceph-csi/build.env
+    CSI_IMAGE_VERSION=${CSI_IMAGE_VERSION:-"canary"}
     export GO111MODULE="on"
     export TEST_COVERAGE="stdout"
     export VM_DRIVER="kvm2"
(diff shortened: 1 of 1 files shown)


single-node-k8s.sh
```diff
@@ -61,6 +61,7 @@ function set_env() {
     # it and will cause a failure if not annotated with /dev/null.
     # shellcheck source=/dev/null
     source "${GOPATH}"/src/github.com/ceph/ceph-csi/build.env
+    CSI_IMAGE_VERSION=${CSI_IMAGE_VERSION:-"canary"}
```
Yuggupta27, line 64:
```suggestion
    CSI_IMAGE_VERSION="${CSI_IMAGE_VERSION:-"canary"}"
```

Madhu-1: @nixpanic PTAL

Madhu-1:
@MergifyIO rebase

single-node-k8s.sh
nixpanic: I do not think this is needed, `shellcheck` does not complain about it

Issue #1379, opened 2020-08-13
Make use of Volumecapability in Expand request to avoid NodeExpansion when its of type block volumemode.


humblec:
# Describe the bug #

```
	// volumePath is targetPath for block PVC and stagingPath for filesystem.
	// check the path is mountpoint or not, if it is
	// mountpoint treat this as block PVC or else it is filesystem PVC
	// TODO remove this once ceph-csi supports CSI v1.2.0 spec
	notMnt, err := mount.IsNotMountPoint(ns.mounter, volumePath)
	if err != nil {
		if os.IsNotExist(err) {
			return nil, status.Error(codes.NotFound, err.Error())
		}
		return nil, status.Error(codes.Internal, err.Error())
	}
	if !notMnt {
		return &csi.NodeExpandVolumeResponse{}, nil
	}
```

As we are already on SPEC 1.2 we could get rid of this TODO.

humblec:
Targeting this for next release, so adding milestone. Self assigning too.

PR #1386, opened 2020-08-17
rbd: Bail out from nodeexpansion if its block mode pvc

humblec:
At CSI spec < 1.2.0, there was no volumecapability in the
expand request. However its available from v1.2+ which allows
us to declare the node operations based on the volume mode.

Fixes: https://github.com/ceph/ceph-csi/issues/1379

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -1050,6 +1050,12 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, status.Error(codes.InvalidArgument, "capacityRange cannot be empty")
 	}
 
+	nodeExpansion := false
+	// Get the nodeexpansion flag set based on the volume mode
+	if req.GetVolumeCapability().GetBlock() == nil {
+		nodeExpansion = true
+	}
+
 	// lock out parallel requests against the same volume ID
 	if acquired := cs.VolumeLocks.TryAcquire(volID); !acquired {
 		klog.Errorf(util.Log(ctx, util.VolumeOperationAlreadyExistsFmt), volID)
@@ -1097,7 +1103,6 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	volSize := util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
 
 	// resize volume if required
-	nodeExpansion := false
 	if rbdVol.VolSize < volSize {
 		util.DebugLog(ctx, "rbd volume %s size is %v,resizing to %v", rbdVol, rbdVol.VolSize, volSize)
 		err = rbdVol.resize(volSize)
@@ -1105,7 +1110,6 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 			klog.Errorf(util.Log(ctx, "failed to resize rbd image: %s with error: %v"), rbdVol, err)
 			return nil, status.Error(codes.Internal, err.Error())
 		}
-		nodeExpansion = true
 	}
 
 	return &csi.ControllerExpandVolumeResponse{
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -715,21 +715,6 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	}
 	defer ns.VolumeLocks.Release(volumeID)
 
-	// volumePath is targetPath for block PVC and stagingPath for filesystem.
-	// check the path is mountpoint or not, if it is
-	// mountpoint treat this as block PVC or else it is filesystem PVC
-	// TODO remove this once ceph-csi supports CSI v1.2.0 spec
-	notMnt, err := mount.IsNotMountPoint(ns.mounter, volumePath)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return nil, status.Error(codes.NotFound, err.Error())
-		}
-		return nil, status.Error(codes.Internal, err.Error())
-	}
-	if !notMnt {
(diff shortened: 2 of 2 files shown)


internal/rbd/controllerserver.go
```diff
@@ -20,16 +20,16 @@ import (
...
-	"github.com/ceph/ceph-csi/internal/util"
-
-	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	klog "k8s.io/klog/v2"
+
+	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
```
Madhu-1, line 29: this is not the standard we follow see https://github.com/ceph/ceph-csi/blob/master/docs/coding.md#imports

internal/rbd/controllerserver.go
```diff
@@ -1050,6 +1050,9 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, status.Error(codes.InvalidArgument, "capacityRange cannot be empty")
 	}
 
+	// Get the access Block type
+	isBlock := req.GetVolumeCapability().GetBlock() != nil
```
Madhu-1, line 1054: nodeExpansion can be set here no need to have a separate variable `isBlock`

internal/rbd/nodeserver.go
```diff
@@ -24,18 +24,18 @@ import (
...
-	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	klog "k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/util/resizefs"
 	utilexec "k8s.io/utils/exec"
 	"k8s.io/utils/mount"
+
+	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
```
Madhu-1, line 35: This is not the import order we follow

nixpanic:
@mergifyio refresh

internal/rbd/controllerserver.go
humblec: done.

internal/rbd/controllerserver.go
humblec: Some auto plugin stuff rearranging it.. will revert that..

humblec:
@Madhu-1 ptal.

internal/rbd/controllerserver.go
```diff
@@ -1050,6 +1050,13 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, status.Error(codes.InvalidArgument, "capacityRange cannot be empty")
 	}
 
+	// Get the nodeexpansion flag set based on the volume mode
```
nixpanic, line 1053: Is there really no need to do expansion on the node when using a block PV? At the very least, the block-device needs re-scanning so that the new size is known on the node. I wonder if that needs `nodeExpansion` or some flag like that.

humblec: True.. it need confirmation. That said, the mapped device should reflect, considering we declare ONLINE, I believe this will go through.. but good to validate it too.

humblec:
hmmm.. why commitlint fails now.? :(

Yuggupta27:
> hmmm.. why commitlint fails now.? :(

```
header must not be longer than 72 characters, current length is 88 [header-max-length]
```

humblec:
> > hmmm.. why commitlint fails now.? :(
> 
> ```
> header must not be longer than 72 characters, current length is 88 [header-max-length]
> ```

@Yuggupta27 below is the header which is of 54 char. I am confused.

```rbd: Bail out from nodeexpansion if its block mode pvc```

**Comment by humblec** [comment]
> > > hmmm.. why commitlint fails now.? :(
> > 
> > 
> > ```
> > header must not be longer than 72 characters, current length is 88 [header-max-length]
> > ```
> 
> @Yuggupta27 below is the header which is of 54 char. I am confused.
> 
> `rbd: Bail out from nodeexpansion if its block mode pvc`

@nixpanic any thoughts here?

File: internal/rbd/controllerserver.go (hunk 4 above)
Comment (humblec) [reply internal/rbd/controllerserver.go:1054]: @nixpanic all the tests pass, so we are safe. :+1:

**Comment by nixpanic** [comment]
> > > > hmmm.. why commitlint fails now.? :(
> > > 
> > > 
> > > ```
> > > header must not be longer than 72 characters, current length is 88 [header-max-length]
> > > ```
> > 
> > 
> > @Yuggupta27 below is the header which is of 54 char. I am confused.
> > `rbd: Bail out from nodeexpansion if its block mode pvc`
> 
> @nixpanic any thoughts here?

See #1394 that improves the commitlint CI job with a rebase so the history of the commits can be tracked better.

**Comment by nixpanic** [comment]
> /retest commitlint

Unsure if this works without #1394 merged, we'll see.

**Comment by humblec** [comment]
@Madhu-1 @nixpanic ptal

File: internal/rbd/controllerserver.go [hunk 5]
```diff
@@ -1050,6 +1050,12 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, status.Error(codes.InvalidArgument, "capacityRange cannot be empty")
 	}
 
+	nodeExpansion := false
+	// Get the nodeexpansion flag set based on the volume mode
```
Madhu-1, line 1054: Does this comment needs to be updated? `nodeexpansion` to `nodeExpansion` ?

nixpanic (approved):

Madhu-1 (approved):

Issue #1387, opened 2020-08-17
rbd: e2e test fails while validating the backend RBD images.


humblec:
# Describe the bug #

Noticed that  CentOS CI jobs fail the RBD cloning e2e tests.  The error is caused by the rbd image backend validation to the clone count. 

```

Aug 11 15:46:29.707: INFO: rbd-40590 app  is in Pending phase expected to be in Running  state (16 seconds elapsed)
Aug 11 15:46:29.727: INFO: rbd-40598 app  is in Pending phase expected to be in Running  state (16 seconds elapsed)
Aug 11 15:46:29.727: INFO: rbd-40595 app  is in Pending phase expected to be in Running  state (16 seconds elapsed)
Aug 11 15:46:29.727: INFO: rbd-40591 app  is in Pending phase expected to be in Running  state (24 seconds elapsed)
Aug 11 15:46:29.790: INFO: rbd-40596 app  is in Pending phase expected to be in Running  state (26 seconds elapsed)
Aug 11 15:46:29.816: INFO: rbd-40594 app  is in Pending phase expected to be in Running  state (26 seconds elapsed)
Aug 11 15:46:31.707: INFO: rbd-40590 app  is in Pending phase expected to be in Running  state (18 seconds elapsed)
Aug 11 15:46:31.731: INFO: rbd-40591 app  is in Pending phase expected to be in Running  state (26 seconds elapsed)
Aug 11 15:46:31.732: INFO: rbd-40598 app  is in Pending phase expected to be in Running  state (18 seconds elapsed)
Aug 11 15:46:31.789: INFO: rbd-40596 app  is in Pending phase expected to be in Running  state (28 seconds elapsed)
Aug 11 15:46:31.817: INFO: rbd-40594 app  is in Pending phase expected to be in Running  state (28 seconds elapsed)
Aug 11 15:46:33.708: INFO: rbd-40590 app  is in Pending phase expected to be in Running  state (20 seconds elapsed)
Aug 11 15:46:33.724: INFO: rbd-40598 app  is in Pending phase expected to be in Running  state (20 seconds elapsed)
Aug 11 15:46:33.789: INFO: rbd-40596 app  is in Pending phase expected to be in Running  state (30 seconds elapsed)
Aug 11 15:46:33.821: INFO: rbd-40594 app  is in Pending phase expected to be in Running  state (30 seconds elapsed)
[...]

PR #1389, opened 2020-08-17
rbd: fix cleanup in case  of clone failure

Madhu-1:
Rearrange the cleanup order incase clone fails, in case of clone failure, we need to first delete the clone and the snapshot from which we created the clone, then as part of the cleanup we need to remove the temporary cloned image and the temporary snapshot created on the parent image.

validate backend rbd images count in each E2E test cases. This helps a lot to catch the issues in each test case.

fixes https://github.com/ceph/ceph-csi/issues/1387

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -213,16 +213,34 @@ var _ = Describe("RBD", func() {
 
 			By("create a PVC and Bind it to an app", func() {
 				validatePVCAndAppBinding(pvcPath, appPath, f)
+				// validate created backend rbd images
+				images := listRBDImages(f)
+				if len(images) != 0 {
+					e2elog.Logf("backend image count %d expected image count %d", len(images), 0)
+					Fail("validate backend image failed")
+				}
 			})
 
 			By("create a PVC and Bind it to an app with normal user", func() {
 				validateNormalUserPVCAccess(pvcPath, f)
+				// validate created backend rbd images
+				images := listRBDImages(f)
+				if len(images) != 0 {
+					e2elog.Logf("backend image count %d expected image count %d", len(images), 0)
+					Fail("validate backend image failed")
+				}
 			})
 
 			By("create a PVC and Bind it to an app with ext4 as the FS ", func() {
 				deleteResource(rbdExamplePath + "storageclass.yaml")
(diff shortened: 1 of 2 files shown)


Madhu-1:
I would like to trigger this test at least for  3 days to catch if we have any issues in rbd E2E.

internal/rbd/clone.go
```diff
@@ -177,6 +177,13 @@ func (rv *rbdVolume) createCloneFromImage(ctx context.Context, parentVol *rbdVol
 	}
 
 	defer func() {
+		if err != nil || errClone != nil {
+			cErr := cleanUpSnapshot(ctx, tempClone, cloneSnap, rv, rv.conn.Creds)
+			if cErr != nil {
+				klog.Errorf(util.Log(ctx, "failed to cleanup image %s or snapshot %s: %v"), cloneSnap, tempClone, cErr)
```
nixpanic, line 183: you're touching this code, so please use the `util.*Log()` function too.

nixpanic (approved):

internal/rbd/clone.go
Madhu-1: sent PR for using util logger here https://github.com/ceph/ceph-csi/pull/1409

Madhu-1:
retesting many times to see I can see the issue again, I tried to  reproduce the issue for around 10 runs with 100 PVC

Madhu-1:
CI passed triggering CI  again

/test all

Madhu-1:
@mergifyio rebase

e2e/rbd.go
```diff
@@ -213,16 +213,34 @@ var _ = Describe("RBD", func() {
 
 			By("create a PVC and Bind it to an app", func() {
 				validatePVCAndAppBinding(pvcPath, appPath, f)
+				// validate created backend rbd images
```
nixpanic, line 216: Consider making this its own function? It is repeated really often now.

Madhu-1: will do it in next PR as am working on refactoring the E2E tests

Madhu-1:
@nixpanic @humblec PTAL. i did many CI run seems like the issue is fixed now

nixpanic (approved):

Madhu-1:
@humblec PTAL

humblec (approved):

PR #1394, opened 2020-08-18
ci: pass REBASE=1 to have commitlint rebase the branch

nixpanic:
When Mergify adds a merge commit to the branch that is being tested with
commitlint, the tool tries to detect the most recent changes based on
the newly merged commit. This is for most PRs the master branch, and
that contains incorrect commit messages in the history. Because of this,
commitlint will fail.

By adding an option (REBASE=1) to the commitlint make target, CI jobs
can request a rebase so that the history of the PR becomes linear again
and commitlint should be able to detect only the new commits.

The 'need-container-cmd' make target is marked as .PHONY which will
cause it to be run every time. That triggers a rebuild of the container
images, even when that is not required.

By removing the 'need-container-cmd' target from .PHONY, and storing the
contents of CONTAINER_CMD in .container-cmd, unneeded rebuilds are
prevented.

PR #1395, opened 2020-08-18
Rebase PRs on top of the target branch when running commitlint

nixpanic:
By passing REBASE=1 to the commitlint make target, the current checked
out PR will be rebased on top of the parent branch. This should result
in a linear history that commitlint can parse correctly.

When Mergify adds a merge commit to the branch that is being tested with
commitlint, the tool tries to detect the most recent changes based on
the newly merged commit. This is for most PRs the master branch, and
that contains incorrect commit messages in the history. Because of this,
commitlint will fail.

By adding an option (REBASE=1) to the commitlint make target, CI jobs
can request a rebase so that the history of the PR becomes linear again
and commitlint should be able to detect only the new commits.

See-also: #1394

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,13 +53,21 @@ lint-markdown:
 lint-yaml:
 	./scripts/lint-extras.sh lint-yaml
 
+#
+# commitlint will do a rebase on top of GIT_SINCE when REBASE=1 is passed.
+#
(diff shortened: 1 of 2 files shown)


nixpanic:
Once this is merged, #1394 can be used to test the updated commitlint CI job (`/retest commitlint`).

PR #1399, opened 2020-08-18
ci: prepare.sh should accept --history without argument

nixpanic:
The --history does not take an argument, similar to --help. Move it out
of ARGUMENT_LIST so 'getopt' does not complain about it.

Currently the commitlint CI job fails with:
```
prepare.sh: option '--history' requires an argument
```

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -14,11 +14,10 @@ ARGUMENT_LIST=(
     "workdir"
     "gitrepo"
     "base"
-    "history"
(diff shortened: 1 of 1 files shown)


nixpanic:
The **commitlint** CI job is expected to fail, it uses the latest merged `prepare.sh` script that contains the error.

Madhu-1:
Merging this one Manually

PR #1401, opened 2020-08-18
ci: fix the minikube version check failure

pkalever:
## Describe what this PR does ##

 Problem:                                                                        
 ------------                                                                         
Fix the minikube version check failure
                                                                                 
 $ minikube version                                                              
 minikube version: v1.12.2                                                       
 commit: be7c19d391302656d27f1f213657d925c4e1cfc2-dirty                          
                                                                                 
 $ ./scripts/minikube.sh up                                                      
 installed minikube version v1.12.2 is not matching requested version latest     
                                                                                 
 Here v1.12.2 is the latest version of minikube, but the script simply bails out.

### DIFF
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -47,7 +47,6 @@ function detect_minikube() {
     # default if minikube is not available
     echo '/usr/local/bin/minikube'
 }
-minikube="$(detect_minikube)"
 
 # install minikube
 function install_minikube() {
@@ -99,6 +98,19 @@ function minikube_losetup() {
     ${minikube} ssh 'sudo sh -c "rm -f /sbin/losetup && cp ~docker/losetup /sbin"'
 }
 
+function minikube_supports_psp() {
+    local MINIKUBE_MAJOR
+    local MINIKUBE_MINOR
+    local MINIKUBE_PATCH
+    MINIKUBE_MAJOR=$(minikube_version 1)
+    MINIKUBE_MINOR=$(minikube_version 2)
+    MINIKUBE_PATCH=$(minikube_version 3)
+    if [[ "${MINIKUBE_MAJOR}" -ge 1 ]] && [[ "${MINIKUBE_MINOR}" -ge 11 ]] && [[ "${MINIKUBE_PATCH}" -ge 1 ]] || [[ "${MINIKUBE_MAJOR}" -ge 1 ]] && [[ "${MINIKUBE_MINOR}" -ge 12 ]]; then
+        return 1
+    fi
+    return 0
+}
+
 # configure minikube
 MINIKUBE_ARCH=${MINIKUBE_ARCH:-"amd64"}
 MINIKUBE_VERSION=${MINIKUBE_VERSION:-"latest"}
@@ -136,18 +148,7 @@ EXTRA_CONFIG="${EXTRA_CONFIG} --extra-config=kubelet.resolv-conf=${RESOLV_CONF}"
 #extra Rook configuration
 ROOK_BLOCK_POOL_NAME=${ROOK_BLOCK_POOL_NAME:-"newrbdpool"}
 
-function minikube_supports_psp() {
-    local MINIKUBE_MAJOR
-    local MINIKUBE_MINOR
-    local MINIKUBE_PATCH
-    MINIKUBE_MAJOR=$(minikube_version 1)
-    MINIKUBE_MINOR=$(minikube_version 2)
-    MINIKUBE_PATCH=$(minikube_version 3)
-    if [[ "${MINIKUBE_MAJOR}" -ge 1 ]] && [[ "${MINIKUBE_MINOR}" -ge 11 ]] && [[ "${MINIKUBE_PATCH}" -ge 1 ]] || [[ "${MINIKUBE_MAJOR}" -ge 1 ]] && [[ "${MINIKUBE_MINOR}" -ge 12 ]]; then
-        return 1
-    fi
-    return 0
-}
+minikube="$(detect_minikube)"
 
 case "${1:-}" in
 up)
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -76,7 +76,7 @@ function enable_psp() {
     echo "prepare minikube to support pod security policies"
     mkdir -p "$HOME"/.minikube/files/etc/kubernetes/addons
     DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-	  cp "$DIR"/psp.yaml "$HOME"/.minikube/files/etc/kubernetes/addons/psp.yaml
+    cp "$DIR"/psp.yaml "$HOME"/.minikube/files/etc/kubernetes/addons/psp.yaml
 }
 
 # Storage providers and the default storage class is not needed for Ceph-CSI
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -54,10 +54,6 @@ function install_minikube() {
         local mk_version version
         read -ra mk_version <<<"$(${minikube} version)"
         version=${mk_version[2]}
-        if [[ "${version}" != "${MINIKUBE_VERSION}" ]]; then
-            echo "installed minikube version ${version} is not matching requested version ${MINIKUBE_VERSION}"
-            exit 1
-        fi
         echo "minikube already installed with ${version}"
         return
     fi
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -50,16 +50,26 @@ function detect_minikube() {
 
 # install minikube
 function install_minikube() {
+    local mku_version latest_version
+    mku_version=$(${minikube} update-check 2> /dev/null | grep "LatestVersion" || true)
+    latest_version=$(echo "${mku_version}" | cut -d' ' -f2)
+    if [[ -z "${latest_version}" ]]; then
+        # skip: update-check failed for some reason, lets continue with what we have
+        latest_version=${MINIKUBE_VERSION}
+    fi
     if type "${minikube}" >/dev/null 2>&1; then
         local mk_version version
         read -ra mk_version <<<"$(${minikube} version)"
         version=${mk_version[2]}
         echo "minikube already installed with ${version}"
-        return
+        if [[ "${version}" == "${latest_version}" ]]; then
+            echo "minikube is already the latest version"
+            return
+        fi
     fi
 
-    echo "Installing minikube. Version: ${MINIKUBE_VERSION}"
-    curl -Lo minikube https://storage.googleapis.com/minikube/releases/"${MINIKUBE_VERSION}"/minikube-linux-"${MINIKUBE_ARCH}" && chmod +x minikube && mv minikube /usr/local/bin/
+    echo "Installing minikube. Version: ${latest_version}"
+    curl -Lo minikube https://storage.googleapis.com/minikube/releases/"${latest_version}"/minikube-linux-"${MINIKUBE_ARCH}" && chmod +x minikube && mv minikube /usr/local/bin/
 }
 
 function install_kubectl() {


scripts/minikube.sh
```diff
@@ -113,7 +113,7 @@ function minikube_supports_psp() {
 
 # configure minikube
 MINIKUBE_ARCH=${MINIKUBE_ARCH:-"amd64"}
-MINIKUBE_VERSION=${MINIKUBE_VERSION:-"latest"}
+MINIKUBE_VERSION=${MINIKUBE_VERSION:-"v1.12.2"}
```
Madhu-1, line 116: This is not always true. what if we get a new minikube release.

scripts/minikube.sh
```diff
@@ -99,6 +98,19 @@ function minikube_losetup() {
     ${minikube} ssh 'sudo sh -c "rm -f /sbin/losetup && cp ~docker/losetup /sbin"'
 }
 
+function minikube_supports_psp() {
```
Madhu-1, line 101: please update the commit message why the reordering is required

scripts/minikube.sh
```diff
@@ -77,7 +76,7 @@ function enable_psp() {
     echo "prepare minikube to support pod security policies"
     mkdir -p "$HOME"/.minikube/files/etc/kubernetes/addons
     DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-	  cp "$DIR"/psp.yaml "$HOME"/.minikube/files/etc/kubernetes/addons/psp.yaml
+    cp "$DIR"/psp.yaml "$HOME"/.minikube/files/etc/kubernetes/addons/psp.yaml
```
Madhu-1, line 79: This can be a separate commit this should not be part of this commit

scripts/minikube.sh
```diff
@@ -99,9 +98,22 @@ function minikube_losetup() {
...
+    if [[ "${MINIKUBE_MAJOR}" -ge 1 ]] && [[ "${MINIKUBE_MINOR}" -ge 11 ]] && [[ "${MINIKUBE_PATCH}" -ge 1 ]] || [[ "${MINIKUBE_MAJOR}" -ge 1 ]] && [[ "${MINIKUBE_MINOR}" -ge 12 ]]; then
+        return 1
+    fi
+    return 0
+}
+
 # configure minikube
 MINIKUBE_ARCH=${MINIKUBE_ARCH:-"amd64"}
-MINIKUBE_VERSION=${MINIKUBE_VERSION:-"latest"}
+MINIKUBE_VERSION=${MINIKUBE_VERSION:-"v1.12.2"}
```
nixpanic, line 116: Fixed versions should only be in the `build.env` file. When trying to install minikube with `latest`, I think it should not check the currently installed minikube version, but just install it.

There might be a minikube command to check if there is an update available. That could be used in addition to the above.

scripts/minikube.sh
pkalever: We should do the same thing what we do for KUBE_VERSION variable in this script :-)

or maybe we should not care for which version of minikube is in-use, like get rid of below:
`if [[ "${version}" != "${MINIKUBE_VERSION}" ]]; then`

scripts/minikube.sh
pkalever: +1 @nixpanic

scripts/minikube.sh
pkalever: Sure Madhu, will update.

scripts/minikube.sh
pkalever: Yep. Until then:

Generally, we define all the functions at the top and start the runnable commands (entry points to script start).
Only this function is odd in the script unlike the rest of them taking away the readability.

pkalever:
> we don't have minikube component for the commit please go through https://github.com/ceph/ceph-csi/blob/master/docs/development-guide.md#code-contribution-workflow to know more about it

Request not to take the commit msgs as direct indication of the component. Instead, they just refer to some context.
Like it could be a directory name, file name, or the component itself.

One close example:
```
commit 58c2a3f15fdd176dc3165082c49f5b7ef5fe7970                          
Author:     Niels de Vos <ndevos@redhat.com>                             
AuthorDate: Sat Apr 18 16:47:56 2020 +0200                               
Commit:     mergify[bot] <37929162+mergify[bot]@users.noreply.github.com>
CommitDate: Tue Apr 21 05:26:32 2020 +0000                               

    minikube: fix ShellCheck errors

    Signed-off-by: Niels de Vos <ndevos@redhat.com>                      
```

I think having `minikube:` doesn't hurt as it is the one which we are playing around in this change, but if you insist, I can change it for you :-)

Thanks!

Madhu-1:
> > we don't have minikube component for the commit please go through https://github.com/ceph/ceph-csi/blob/master/docs/development-guide.md#code-contribution-workflow to know more about it
> 
> Request not to take the commit msgs as direct indication of the component. Instead, they just refer to some context.
> Like it could be a directory name, file name, or the component itself.
> 
> One close example:
> 
> ```
> commit 58c2a3f15fdd176dc3165082c49f5b7ef5fe7970                          
> Author:     Niels de Vos <ndevos@redhat.com>                             
> AuthorDate: Sat Apr 18 16:47:56 2020 +0200                               
> Commit:     mergify[bot] <37929162+mergify[bot]@users.noreply.github.com>
> CommitDate: Tue Apr 21 05:26:32 2020 +0000                               
> 
>     minikube: fix ShellCheck errors
> 
>     Signed-off-by: Niels de Vos <ndevos@redhat.com>                      
> ```
> 
> I think having `minikube:` doesn't hurt as it is the one which we are playing around in this change, but if you insist, I can change it for you :-)
> 
> Thanks!

For the above example was before we made some standards for commit messages. we do request contributors to go through https://github.com/ceph/ceph-csi/blob/master/docs/development-guide.md#commit-messages and understand the current standard we follow in ceph-csi repo.

nixpanic:
> I think having `minikube:` doesn't hurt as it is the one which we are playing around in this change, but if you insist, I can change it for you :-)

The valid components are now listed in `.commitlintrc.yml`. Instead of `minikube:`, `ci:` would be more appropriate.

scripts/minikube.sh
```diff
@@ -54,10 +54,6 @@ function install_minikube() {
         local mk_version version
         read -ra mk_version <<<"$(${minikube} version)"
         version=${mk_version[2]}
```
nixpanic, line 56: If the requested version is `latest`, this check can be skipped. In other cases, it should be kept.

nixpanic (changes requested):

scripts/minikube.sh
pkalever: @nixpanic have you looked at the next new commit, where the commit tries to update the minikube if the latest version exists.

Thanks!

scripts/minikube.sh
```diff
@@ -55,7 +55,17 @@ function install_minikube() {
...
         echo "minikube already installed with ${version}"
-        return
+        mku_version=$(minikube update-check 2> /dev/null || ret=${?})
```
nixpanic, line 58: you need to execute `${minikube}` like done in other places.

`ret` is set in a subshell (started with `$(...)`), I would be surprised if the value of `ret` is stored in `ret` after the subshell exits. Please check.

scripts/minikube.sh
nixpanic: I somehow missed that, thanks!

scripts/minikube.sh
```diff
@@ -55,7 +55,17 @@ function install_minikube() {
...
+        latest_version=$(echo ${mku_version} | cut -d' ' -f4)
+        if [[ ${ret} -ne 0 ]] || [[ -z ${latest_version} ]]; then
```
nixpanic, line 60: Place variables in `"` quotes please. I guess `shellcheck` will complain about that too.

nixpanic (changes requested):

scripts/minikube.sh
pkalever: > you need to execute `${minikube}` like done in other places.

Sure.

> 
> `ret` is set in a subshell (started with `$(...)`), I would be surprised if the value of `ret` is stored in `ret` after the subshell exits. Please check.

Yes, it will. The trick was played here because simply using:
mku_version=$(minikube update-check)
will exit the whole script if the `minikube update-check` command fails, like in case the network is not available etc..

In the case of failure the above trick will 1. help not exit the script 2. collect the exit code of the command too

Thanks!

scripts/minikube.sh
pkalever: sure will take care of this.

scripts/minikube.sh
```diff
@@ -51,11 +51,21 @@ function detect_minikube() {
 # install minikube
 function install_minikube() {
     if type "${minikube}" >/dev/null 2>&1; then
-        local mk_version version
+        local mk_version version mku_version latest_version ret
         read -ra mk_version <<<"$(${minikube} version)"
         version=${mk_version[2]}
         echo "minikube already installed with ${version}"
-        return
+        mku_version=$(${minikube} update-check 2> /dev/null || ret=${?})
```
nixpanic, line 58: shellcheck (part of `make-containerized-test`) agrees that changing `ret` in a sub-shell is not working as you intend.

nixpanic (changes requested):

scripts/minikube.sh
pkalever: I had tested it enabling `set -x` with various conditions and it worked for me and logically it looks fine too :-)
Any have, now that it bothers shellcheck, I had made some changes and updated the PR. Please take a look.

Thanks @nixpanic

scripts/minikube.sh
```diff
@@ -50,16 +50,28 @@ function detect_minikube() {
 
 # install minikube
 function install_minikube() {
+     local mku_version latest_version
+     mku_version=$(${minikube} update-check 2> /dev/null || true)
+     latest_version=$(echo ${mku_version} | cut -d' ' -f4)
```
nixpanic, line 55: I have an old version of minikube, that outputs:
```
$ minikube update-check
CurrentVersion: v0.27.0
LatestVersion: v1.12.3
```
Better set `latest_version` to `awk '/LatestVersion {print $NF}' <<< "${mku_version}"` so that the order of the lines, or additions do not influence the result.

This should also probably be inside the `if type ${minikube}` statement in case minikube is not installed yet.

nixpanic: Indention looks a little off too, compared to the original lines below.

scripts/minikube.sh
```diff
@@ -50,16 +50,28 @@ function detect_minikube() {
...
+     fi
     if type "${minikube}" >/dev/null 2>&1; then
         local mk_version version
         read -ra mk_version <<<"$(${minikube} version)"
         version=${mk_version[2]}
         echo "minikube already installed with ${version}"
-        return
+        MINIKUBE_VERSION=${version}
+        if [[ "${version}" == "${latest_version}" ]]; then
+          echo "minikube is already the latest version"
```
nixpanic, line 67: weird indention, make it 4 spaces please.

nixpanic (changes requested):

scripts/minikube.sh
pkalever: > I have an old version of minikube, that outputs:
> 
> ```
> $ minikube update-check
> CurrentVersion: v0.27.0
> LatestVersion: v1.12.3
> ```
> 
> Better set `latest_version` to `awk '/LatestVersion {print $NF}' <<< "${mku_version}"` so that the order of the lines, or additions do not influence the result.

This won't work Niels, as storing the output of the command will already convert the above multiline output to single line.

✨ mku_version=$(minikube update-check 2> /dev/null || true)
✨ echo $mku_version
CurrentVersion: v1.12.3 LatestVersion: v1.12.3

But I get your point, will defend on it.

> 
> This should also probably be inside the `if type ${minikube}` statement in case minikube is not installed yet.

I'm trying to capture the latest_version which will be used after the if block, and absence of minkube command will be taken care, as it will result in empty string ${mku_version}

pkalever: > Indention looks a little off too, compared to the original lines below.

Hmm, tweaked my vimrc, thanks!

scripts/minikube.sh
pkalever: tweaked my vimrc now, thanks!

scripts/minikube.sh
```diff
@@ -50,16 +50,28 @@ function detect_minikube() {
...
+        # skip: update-check failed for some reason, lets continue with what we have
+        latest_version=${MINIKUBE_VERSION}
+    fi
     if type "${minikube}" >/dev/null 2>&1; then
         local mk_version version
         read -ra mk_version <<<"$(${minikube} version)"
         version=${mk_version[2]}
         echo "minikube already installed with ${version}"
-        return
+        MINIKUBE_VERSION=${version}
```
nixpanic, line 65: Is it needed to change the `MINIKUBE_VERSION` in this function? If it contains `latest`, that should be fine, I think?

pkalever: Just feeding it with the real version number for future use `within the script`.
Here it `might` be a lower version, and at the end, once the update is done then we assign with the latest.

@nixpanic do you see any issue with this?

pkalever: This way MINIKUBE_VERSION is always set to the currently installed version rather than faking as `lastest` always.

nixpanic: Because `MINIKUBE_VERSION` gets passed into the script through an environment variable, changing its value is ugly. If there is a need for using the actual version later on, it is cleaner to add a new (lowercase) variable for that.

pkalever: @nixpanic ok, updated addressing the review comments. Thanks!

pkalever:
@Madhu-1 @nixpanic FYI the CI is failing with:
```
error: unable to read URL "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml", server reported 404 Not Found, status code=404
```

Madhu-1:
> @Madhu-1 @nixpanic FYI the CI is failing with:
> 
> ```
> error: unable to read URL "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml", server reported 404 Not Found, status code=404
> ```

rebase with the latest master.

nixpanic (approved):

Madhu-1 (approved):

PR #1413, opened 2020-08-19
ci: keep kubectl install in sync with minikube install

pkalever:
## Describe what this PR does ##
* Update minikube version in the env file
* keep kubectl install in sync with minikube install

You will appreciate the changes when rebased on top of https://github.com/ceph/ceph-csi/pull/1401 (might require manual rebase)


pkalever:
@nixpanic should we `yum install VirtualBox` in the ci?

Thanks!

nixpanic:
> @nixpanic should we `yum install VirtualBox` in the ci?

No, we don't want to use VirtualBox, but stick with KVM. Some scripts use vagrant, see the `ci/centos` branch for details.

build.env
```diff
@@ -31,6 +31,7 @@ HELM_VERSION=v3.1.2
 
 # minikube settings
 MINIKUBE_VERSION=v1.12.1
+KUBE_VERSION=v1.18.8
```
nixpanic, line 34: Adding a variable is fine, but the commit message should describe why it is used. You can also add the actual usage in the same commit to make things clearer. No need to split everything in very small commits.

build.env
```diff
@@ -30,7 +30,7 @@ GOSEC_VERSION=v2.3.0
 HELM_VERSION=v3.1.2
 
 # minikube settings
-MINIKUBE_VERSION=v1.12.1
+MINIKUBE_VERSION=v1.12.3
```
nixpanic, line 33: Use `rebase:` as prefix for version changes please.

build.env
```diff
@@ -32,7 +32,7 @@ HELM_VERSION=v3.1.2
 # minikube settings
 MINIKUBE_VERSION=v1.12.3
 KUBE_VERSION=v1.18.8
-VM_DRIVER=none
+VM_DRIVER=virtualbox
```
nixpanic, line 35: Travis CI does not allow VMs to get started, so we default to `none` for that.

scripts/minikube.sh
```diff
@@ -99,6 +99,10 @@ function minikube_losetup() {
     ${minikube} ssh 'sudo sh -c "rm -f /sbin/losetup && cp ~docker/losetup /sbin"'
 }
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+# shellcheck source=build.env
+[ ! -e "${SCRIPT_DIR}"/../build.env ] || source "${SCRIPT_DIR}"/../build.env
```
nixpanic, line 131: I do not think this is helpful. Some CI jobs pass/change the environment to their needs before running the minikube.sh script. This will undo that again.

scripts/minikube.sh
```diff
@@ -67,10 +67,37 @@ function install_minikube() {
...
+    # default if kubectl is not available
+    echo '/usr/local/bin/kubectl'
+}
+
 function install_kubectl() {
+    local latest_version
+    latest_version=$(curl -L https://storage.googleapis.com/kubernetes-release/release/stable.txt 2> /dev/null)
+    if [[ -z "${latest_version}" ]]; then
+        # skip: getting latest version failed, continue with what we have
+        latest_version=${KUBE_VERSION}
```
nixpanic, line 84: `KUBE_VERSION` should probably default to `latest` if you are adding the check for `release/stable.txt`

scripts/minikube.sh
```diff
@@ -67,10 +67,37 @@ function install_minikube() {
...
+    fi
+    if type "${kubectl}" >/dev/null 2>&1; then
+        local kctl_version version
```
nixpanic, line 87: instead of `kctl_version`, just use `kubectl_version`

nixpanic (changes requested):

scripts/minikube.sh
Madhu-1: won't it cause any issue as we need to install different kubernetes version for e2e and helm testing?

nixpanic: yes, that is one example of what I am referring to

build.env
pkalever: sure

build.env
pkalever: Ah! got it.

scripts/minikube.sh
pkalever: Hmm, I see.
It's like the top-level script source build.env and then calls minikube.sh
I thought it will be good to make minikube.sh self-sufficient by loading the env variables within itself.

scripts/minikube.sh
pkalever: @nixpanic, unfortunately, there is no tag like `latest` for kubectl like minikube.
But I will consider not updating the KUBE_VERSION variable.

scripts/minikube.sh
pkalever: sure

build.env
pkalever: If we do not want to source the build.env in the minikube.sh, then we don't need this change.
Thanks!

pkalever:
@Madhu-1 @nixpanic FYI the CI is failing with:
```
error: unable to read URL "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/master/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml", server reported 404 Not Found, status code=404
```

scripts/minikube.sh
nixpanic: There is no `latest` tag, but you can have `KUBE_VERSION=latest` and then fetch the last stable version as you did. It would be helpful to be able to specify `latest` as default (so we do not run with old versions at one point).

pkalever: I hope you remember our yesterday's discussion about changing MINIKUBE_VERSION inside the script.

@nixpanic as part of install_kubectl() we will always fetch the latest kube version and store in a global variable (here latest_version, I'm changing it to KUBE_LATEST_VERSION in my next version of patches) which we will use through out this script.

And IMHO we should set KUBE_VERSION to something valid like tag `vx.y.z` so that we have something to fall back to, just in case if `curl -L https://storage.googleapis.com/kubernetes-release/release/stable.txt` fail for some practical reasons.

Thanks!

nixpanic: Sorry, what I meant is, that I would love to see the _default_ `KUBE_VERSION` to be set to `latest` instead of `v1.14.10`.

scripts/minikube.sh
```diff
@@ -72,10 +72,33 @@ function install_minikube() {
...
+    # default if kubectl is not available
+    echo '/usr/local/bin/kubectl'
+}
+
 function install_kubectl() {
+    KUBE_LATEST_VERSION=$(curl -L https://storage.googleapis.com/kubernetes-release/release/stable.txt 2> /dev/null)
+    if [[ -z "${KUBE_LATEST_VERSION}" ]]; then
+        # skip: getting latest version failed, continue with what we have
+        KUBE_LATEST_VERSION=${KUBE_VERSION}
+    fi
```
pkalever, line 89: @nixpanic 
I think it started making sense to me, let me get some more clarification before I start tweaking this again.

1. Change s/KUBE_VERSION=${KUBE_VERSION:-"v1.14.10"}/KUBE_VERSION=${KUBE_VERSION:-"v1.18.3"}/g
2. May be also move the above block out of function install_kubectl() so that KUBE_LATEST_VERSION is available all the time.

Thoughts?

nixpanic: In the top of the script, use `KUBE_VERSION=${KUBE_VERSION:-"latest"}`, similar to `MINIKUBE_VERSION`. Then, when installing kubectl, fetch the `latest` version that is in `stable.txt`.

No need to have `KUBE_LATEST_VERSION`, and the installation of a selected minikube version would be the same as selecting a version for kubectl.

scripts/minikube.sh
```diff
@@ -154,7 +172,20 @@ EXTRA_CONFIG="${EXTRA_CONFIG} --extra-config=kubelet.resolv-conf=${RESOLV_CONF}"
...
+kube_latest_version=$(curl -L https://storage.googleapis.com/kubernetes-release/release/stable.txt 2> /dev/null)
+if [[ -z "${kube_latest_version}" ]]; then
+    # skip: getting latest version failed, continue with what we have
+    if [[ "${KUBE_VERSION}" == "latest" ]]; then
+        echo "ERROR: failed getting kube version"
+        exit 1
+    fi
+else
+    # update the version string from latest with the real version
+    KUBE_VERSION=${kube_latest_version}
```
nixpanic, line 184: Why? This should not be needed.

nixpanic:
The https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/mini-e2e-helm_k8s-1.17.8/412/console job failed. It is a job that installs minikube (any version), and needs to use kubernetes 1.17.8. `KUBE_VERSION` is passed to the script, and it should not try to use/install `kubectl` v1.18.8:
```
* minikube v1.12.3 on Centos 8.2.2004
  - MINIKUBE_VERSION=v1.12.3
...
* Preparing Kubernetes v1.18.8 on Docker 19.03.12 ...
...
cp: cannot stat '/root/.minikube/cache/linux/v1.17.8/kubectl': No such file or directory
```

scripts/minikube.sh
```diff
@@ -154,7 +172,20 @@ EXTRA_CONFIG="${EXTRA_CONFIG} --extra-config=kubelet.resolv-conf=${RESOLV_CONF}"
 #extra Rook configuration
 ROOK_BLOCK_POOL_NAME=${ROOK_BLOCK_POOL_NAME:-"newrbdpool"}
 
+kube_latest_version=$(curl -L https://storage.googleapis.com/kubernetes-release/release/stable.txt 2> /dev/null)
```
nixpanic, line 175: you only need to check `stable.txt` in case `KUBE_VERSION == latest`, otherwise just install `kubectl` with version `KUBE_VERSION`.

scripts/minikube.sh
pkalever: There are other places where KUBE_VERSION is used, if KUBE_VERSION="latest" then it is just a string indicating us that we need latest kube and is not useful.

scripts/minikube.sh
pkalever: sure

pkalever:
> The https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/mini-e2e-helm_k8s-1.17.8/412/console job failed. It is a job that installs minikube (any version), and needs to use kubernetes 1.17.8. `KUBE_VERSION` is passed to the script, and it should not try to use/install `kubectl` v1.18.8:
> 
> ```
> * minikube v1.12.3 on Centos 8.2.2004
>   - MINIKUBE_VERSION=v1.12.3
> ...
> * Preparing Kubernetes v1.18.8 on Docker 19.03.12 ...
> ...
> cp: cannot stat '/root/.minikube/cache/linux/v1.17.8/kubectl': No such file or directory
> ```

Thanks @nixpanic for the details, I was really unsure of the root cause. Will fix it now.

pkalever:
@Madhu-1 @nixpanic 

Job `ci/centos/mini-e2e-helm/k8s-1.18.5 ` failed with:

```
 Problem: package docker-ce-3:19.03.12-3.el7.x86_64 requires containerd.io >= 1.2.2-3, but none of the providers can be installed
  - cannot install the best candidate for the job
```
can we retrigger once again and see if the issue persist ?

Thanks!

nixpanic:
> can we retrigger once again and see if the issue persist ?

You should be able to leave a comment (only this line) to run a CI job again:
```
/retest ci/centos/mini-e2e-helm/k8s-1.18.5
```

nixpanic (approved):

Madhu-1 (approved):

PR #1425, opened 2020-08-24
cephfs: Add SINGLE_NODE_WRITER capability to CephFS

humblec:
This mode specify a volume can only be published once as read/write
on a single node, at any given time.

### DIFF
--- a/internal/cephfs/driver.go
+++ b/internal/cephfs/driver.go
@@ -123,7 +123,7 @@ func (fs *Driver) Run(conf *util.Config) {
 			csi.ControllerServiceCapability_RPC_CLONE_VOLUME,
 		})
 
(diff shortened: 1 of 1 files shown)


Madhu-1 (commented):
Isn't this feature already supported in cephfs now?

humblec:
> Isn't this feature already supported in cephfs now?

Looks like we dont need this for CephFS. Considering the fact that, we can have more than one RW Publish on single node..

nixpanic (changes requested):
> > Isn't this feature already supported in cephfs now?
> 
> Looks like we dont need this for CephFS. Considering the fact that, we can have more than one RW Publish on single node..

Does that mean this PR can be closed without merging?

Also, adding a capability without tests seems wrong. You'll need to add an e2e test, or explain why that is not needed.

Issue #1408, opened 2020-08-19
Make sure the FS resize is attempted on proper path in NodeExpand call


humblec:
# Describe the bug #
At present, we go ahead and issue FS resize regardless of whether the mapped RBD device already reached to the new requested size or not. we should check the size of the mount before attempting the resize on the volumepath given at the node expansion request which could avoid issues on various scenarios or help us to make sure we are doing `really required resize` instead of doing it blindly and thus avoid false positive returns

```
	// volumePath is targetPath for block PVC and stagingPath for filesystem.
	// check the path is mountpoint or not, if it is
	// mountpoint treat this as block PVC or else it is filesystem PVC
	// TODO remove this once ceph-csi supports CSI v1.2.0 spec
	notMnt, err := mount.IsNotMountPoint(ns.mounter, volumePath)
	if err != nil {
		if os.IsNotExist(err) {
			return nil, status.Error(codes.NotFound, err.Error())
		}
		return nil, status.Error(codes.Internal, err.Error())
	}
	if !notMnt {
		return &csi.NodeExpandVolumeResponse{}, nil
	}

	devicePath, err := getDevicePath(ctx, volumePath)
	if err != nil {
		return nil, status.Error(codes.Internal, err.Error())
	}

	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: utilexec.New()}
	 --------> [1]
       // TODO check size and return success or error
	volumePath += "/" + volumeID
	resizer := resizefs.NewResizeFs(diskMounter)
	ok, err := resizer.Resize(devicePath, volumePath)
	if !ok {
		return nil, fmt.Errorf("rbd: resize failed on path %s, error: %v", req.GetVolumePath(), err)
	}
	return &csi.NodeExpandVolumeResponse{}, nil
}
```
[1] So the size validation of the mount path has to be happen here and the FS resize has to be attempted accordingly.

PR #1380, opened 2020-08-14
build: update kube dependency to v1.19.0

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

Note for reviewer: We could either update to the latest version of 1.18 or I can adopt 1.19 as soon as the release is out.  Let me know the choice accordingly I will update the PR.

PR #1426, opened 2020-08-24
rbd: use staging path for resize when available

agarwal-mudit:
During the resize operation,  use the staging path if its available or else use the volume path for FS resize.

Fixes: #1408

### DIFF
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -701,7 +701,16 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if volumeID == "" {
 		return nil, status.Error(codes.InvalidArgument, "volume ID must be provided")
 	}
-	volumePath := req.GetVolumePath()
+
+	// Get volume path
+	// With Kubernetes version>=v1.19.0, expand request carries volume_path and
+	// staging_target_path, what csi requires is staging_target_path.
+	volumePath := req.GetStagingTargetPath()
+	if volumePath == "" {
+		// If Kubernetes version < v1.19.0 the volume_path would be
+		// having the staging_target_path information
+		volumePath = req.GetVolumePath()
+	}
 	if volumePath == "" {
 		return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
 	}


internal/rbd/nodeserver.go
```diff
@@ -718,8 +719,20 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
 
 	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: utilexec.New()}
-	// TODO check size and return success or error
 	volumePath += "/" + volumeID
+
+	// check size and return success or error
+	var vStat syscall.Statfs_t
+	err = syscall.Statfs(volumePath, &vStat)
+	if err != nil {
+		return nil, status.Errorf(codes.Aborted, util.VolumePathStatfsFailedFmt, req.GetVolumePath())
```
Madhu-1, line 728: This should be an internal error

internal/rbd/nodeserver.go
```diff
@@ -718,8 +719,20 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
+	}
+	volSize := util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
+	vPathSize := util.RoundOffBytes(vStat.Bsize * int64(vStat.Blocks))
+	if vPathSize == volSize {
```
Madhu-1, line 732: we should return success if its already resized to new size

humblec, line 732: @agarwal-mudit  If both are equal , we dont have to error out, rather say the nodeExpandVolume was successful . Isnt it ?

internal/rbd/nodeserver.go
```diff
@@ -718,8 +719,20 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
+		return nil, status.Errorf(codes.Aborted, util.VolumeAlreadyResizedFmt, req.GetVolumePath())
```
humblec, line 733: @agarwal-mudit if the mount size is less than requested,  it has to be retried.

internal/util/errors.go
```diff
@@ -21,6 +21,13 @@ import (
 	"fmt"
 )
 
+const (
+	// VolumeAlreadyResizedFmt string format to return when FS size is already done.
+	VolumeAlreadyResizedFmt = "FS at the path %s is already resized"
```
Madhu-1, line 26: error messages we return to user should not be capitalized. as these errors are used only in one place better to return error directly from the function

internal/rbd/nodeserver.go
agarwal-mudit: What if  the CSI driver prematurely issued the FS resize before krbd noticed that the volume was resized like what Jason said in https://bugzilla.redhat.com/show_bug.cgi?id=1868703#c10
In that case also both sizes would be same, shouldn't we fail and retry in such case or I am missing something.

agarwal-mudit: replied in Humble's comment as both are same.

Madhu-1: This is against the CSI spec. we return success as the CSI call should be idempotent.

agarwal-mudit:
Lets keep this on hold, there is complete change in the way NodeExpandVolume Request goes in  D/S and U/S.
I will update the BZ for more details.

internal/rbd/nodeserver.go
```diff
@@ -718,8 +725,24 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	}
 
 	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: utilexec.New()}
-	// TODO check size and return success or error
 	volumePath += "/" + volumeID
+
+	// check size of the mount path
+	var vStat syscall.Statfs_t
```
Madhu-1, line 731: @agarwal-mudit do we still need this change?

for me looks like above volumePath check should be enough to fix resize issue.

agarwal-mudit: Yes, this is a precautionary change. Not related to that issue.

Madhu-1: can we remove this change?

internal/rbd/nodeserver.go
```diff
@@ -701,7 +702,13 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if volumeID == "" {
 		return nil, status.Error(codes.InvalidArgument, "volume ID must be provided")
 	}
-	volumePath := req.GetVolumePath()
+
+	// Get volume path
+	volumePath := req.GetStagingTargetPath()
+	if volumePath == "" {
+		// If Kubernetes version < v19.0
```
Madhu-1, line 709: it's better to add more details here what is changed in v1.19 and why this check is required

internal/rbd/nodeserver.go
agarwal-mudit: I will let @humblec comment on that

agarwal-mudit:
There is a complete change in the way NodeExpandVolume request is being called in Kube v18.0 and v19.0:

**### v18.0:**
**Expand request carries only volume_path**

I0824 07:29:10.172338   22867 utils.go:159] ID: 15 Req-ID: 0001-0009-rook-ceph-0000000000000001-168e6382-e5db-11ea-9c7e-0242ac110013 GRPC call: /csi.v1.Node/NodeExpandVolume

I0824 07:29:10.173992   22867 utils.go:160] ID: 15 Req-ID: 0001-0009-rook-ceph-0000000000000001-168e6382-e5db-11ea-9c7e-0242ac110013 GRPC request:
 {"capacity_range":{"required_bytes":2147483648},"volume_id":"0001-0009-rook-ceph-0000000000000001-168e6382-e5db-11ea-9c7e-0242ac110013","volume_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-53226d06-3de6-4998-bb10-ba6e6ea479f5/globalmount"}

**### v19.0**
**Expand request carries volume_path  and staging_target_path, what csi requires is staging_target_path**

I0824 09:24:45.735515  507082 utils.go:159] ID: 66 Req-ID: 0001-0011-openshift-storage-0000000000000001-cce6ebe3-e5ea-11ea-a920-0a580a810229 GRPC call: /csi.v1.Node/NodeExpandVolume
I0824 09:24:45.737321  507082 utils.go:160] ID: 66 Req-ID: 0001-0011-openshift-storage-0000000000000001-cce6ebe3-e5ea-11ea-a920-0a580a810229 GRPC request: {"capacity_range":{"required_bytes":2147483648},"staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-3c6426fa-7476-478f-b685-f8b7831ce70b/globalmount","volume_capability":{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":1}},"volume_id":"0001-0011-openshift-storage-0000000000000001-cce6ebe3-e5ea-11ea-a920-0a580a810229","volume_path":"/var/lib/kubelet/pods/43de2997-5f4e-4fb3-a0e4-7e0d303c6c37/volumes/kubernetes.io~csi/pvc-3c6426fa-7476-478f-b685-f8b7831ce70b/mount"}

Because of this, expansion is a no-op in Kube v19.0 as getDevidePath() is not able to find the device path based on volume_path.

Suggested fix:
```
// Get volume path
	volumePath := req.GetStagingTargetPath()
	if volumePath == "" {
		// If Kubernetes version < v19.0
		volumePath = req.GetVolumePath()
	}
```

Kube change: https://github.com/kubernetes/kubernetes/commit/69613da0ae9743397ba7b2b3daebdc18d68e795e

internal/rbd/nodeserver.go
agarwal-mudit: I have added a comment in the PR, can I give the PR reference in the commit because otherwise it is difficult to explain that in the comments here.

internal/rbd/nodeserver.go
```diff
@@ -718,8 +719,24 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	}
 
 	diskMounter := &mount.SafeFormatAndMount{Interface: ns.mounter, Exec: utilexec.New()}
-	// TODO check size and return success or error
 	volumePath += "/" + volumeID
+
+	// check size of the mount path
+	var vStat syscall.Statfs_t
+	err = syscall.Statfs(volumePath, &vStat)
```
nixpanic, line 732: This asks for the size that is available in the filesystem, but you'll need to check the size of the block device instead. Filesystems have some overhead, so the available size on the filesystem is less than was requested for the PVC. Depending on the size of the block device, and type of filesystem used, the different may run into many MBs.

Not sure how you can check that a filesystem occupies (near to) the full block ...

internal/rbd/nodeserver.go
nixpanic: It would be better to explain the different here. Having it in the PR is good, but everyone reading this comment would then need to open the PR, and that seems unnecessary.

nixpanic (changes requested):

internal/rbd/nodeserver.go
agarwal-mudit: I think RoundOffBytes will take care of that. Also, block device would be already resized by this time and always be equal to the requested size.

So, IMO, this is the best we can do.

nixpanic: I do not think `RoundOffBytes` will do it sufficiently. You can try to  create a very small PVC (1 MB, and resize it to 2 MB). Or create a very large PVC (multiple TBs and double the size). Both with XFS and ext4... There will be a large difference in the filesystem overhead, and the `vPathSize == volSize` check may never get to `true`.

agarwal-mudit: Hmm.. we don't have previous size also here otherwise we could have checked if latest size is greater than that.

Then AFAICT, there is no good way of doing this and we should remove this check.

@humblec any comments?

humblec: Actually we could check the size and compare between `volume_path and staging_target_path `and decide we already got it covered. Isnt it ? I am fine to remove this check too.   @agarwal-mudit

agarwal-mudit: 1. Those both would have same size i.e. the previous size
2. We don't have both in kube <v19.0

humblec: Lets remove this check.. Its fine if we tried resize again.. it will be NO OP.

internal/rbd/nodeserver.go
agarwal-mudit: Removed

internal/rbd/nodeserver.go
agarwal-mudit: fixed

agarwal-mudit:
Please add DNM label, I need to verify this once in DS after the change. Will let you know once that is done.

humblec:
This should be validated with 1.19 update too . #1380

agarwal-mudit:
> Please add DNM label, I need to verify this once in DS after the change. Will let you know once that is done.

Tested on OCS4.6, working as expected. Ready to go from my side.

PTAL

Madhu-1 (approved):

nixpanic (approved):

internal/rbd/nodeserver.go
```diff
@@ -701,7 +701,16 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	if volumeID == "" {
 		return nil, status.Error(codes.InvalidArgument, "volume ID must be provided")
 	}
-	volumePath := req.GetVolumePath()
+
+	// Get volume path
+	// With Kubernetes version>=v19.0, expand request carries volume_path and
```
humblec, line 706: version >= v1.19.0 

@agarwal-mudit ^^^

Madhu-1 (changes requested):
requesting change for  https://github.com/ceph/ceph-csi/pull/1426/files#r476363068

internal/rbd/nodeserver.go
agarwal-mudit: Thanks for catching this

Madhu-1 (approved):

humblec:
@agarwal-mudit the PR title need a change.

Madhu-1:
> @agarwal-mudit the PR title need a change.

@humblec i have modified it PTAL

humblec (approved):

PR #1429, opened 2020-08-24
deploy: remove preStop hook from daemonset templates

Madhu-1:
The lifecycle preStop hook fails on container stop/exit because /bin/sh is not present in the driver registrar container
image. the driver-registrar (v2.0.0)  will remove the socket file before stopping. we don't need to have any preStop hook to remove the socket as it was not working as expected

### DIFF
--- a/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
+++ b/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
@@ -44,18 +44,6 @@ spec:
             - "--v=5"
             - "--csi-address=/csi/{{ .Values.pluginSocketFile }}"
             - "--kubelet-registration-path={{ .Values.socketDir }}/{{ .Values.pluginSocketFile }}"
-          lifecycle:
-            preStop:
-              exec:
-                {{- /*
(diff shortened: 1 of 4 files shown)


nixpanic (approved):
Indeed, if there is no shell, we also do not need these commands.

@wilmardo might want to review this too.

humblec:
lgtm. not approving now and will wait for @wilmardo's review with a time out :+1:

pkalever (approved):
LGTM, no sh/bash command in the driver registrar

```
# kubectl exec -it csi-rbdplugin-79k2j -c driver-registrar -- /bin/sh
OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"/bin/sh\": stat /bin/sh: no such file or directory": unknown
command terminated with exit code 126
```

wilmardo:
This can be removed, if I remember correctly this was for some major migration of the socket path.

When I updated the helm chart to v3 I just improved the logic a little and left the hooks to avoid accidentally creating breaking changes. The NOTE is just to clarify the reason behind the logic :+1: 

Nice cleanup @Madhu-1 LGTM!

Issue #1260, opened 2020-07-22
RBD: implement rbd.updateVolWithImageInfo() with go-ceph


nixpanic:
# Describe the feature you'd like to have #

`updateVolWithImageInfo()` is currently executing an `rbd` command to get details about an RBD image.

Fetching the needed timestamps it not possible with a released version of go-ceph yet, PR ceph/go-ceph#310 is only available in the master branch.

Once go-ceph v0.5 has been released, this issue can get addressed.

agarwal-mudit:
Please assign to me if no one has planned yet to work on it.

PR #1432, opened 2020-08-24
rbd: implement rbd.updateVolWithImageInfo() with go-ceph

agarwal-mudit:
updateVolWithImageInfo() is currently executing an rbd command
to get details about an RBD image. Replaced it with the
required go-ceph functions.

Fixes: #1260

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -900,64 +900,6 @@ func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *r
 	return nil
 }
 
-// imageInfo strongly typed JSON spec for image info.
-type imageInfo struct {
-	ObjectUUID string     `json:"name"`
-	Size       int64      `json:"size"`
-	Features   []string   `json:"features"`
-	CreatedAt  string     `json:"create_timestamp"`
-	Parent     parentInfo `json:"parent"`
-}
-
-// parentInfo  spec for parent volume  info.
-type parentInfo struct {
-	Image    string `json:"image"`
-	Pool     string `json:"pool"`
-	Snapshot string `json:"snapshost"`
-}
-
-// updateVolWithImageInfo updates provided rbdVolume with information from on-disk data
-// regarding the same.
-func (rv *rbdVolume) updateVolWithImageInfo(cr *util.Credentials) error {
-	// rbd --format=json info [image-spec | snap-spec]
-	var imgInfo imageInfo
-
-	stdout, stderr, err := util.ExecCommand(
-		context.TODO(),
-		"rbd",
-		"-m", rv.Monitors,
-		"--id", cr.ID,
-		"--keyfile="+cr.KeyFile,
-		"-c", util.CephConfigPath,
-		"--format="+"json",
-		"info", rv.String())
-	if err != nil {
-		util.ErrorLogMsg("failed getting information for image (%s): (%s)", rv, err)
-		if strings.Contains(stderr, "rbd: error opening image "+rv.RbdImageName+
-			": (2) No such file or directory") {
-			return util.JoinErrors(ErrImageNotFound, err)
-		}
-		return err
-	}
-
-	err = json.Unmarshal([]byte(stdout), &imgInfo)
-	if err != nil {
-		util.ErrorLogMsg("failed to parse JSON output of image info (%s): (%s)", rv, err)
-		return fmt.Errorf("unmarshal failed: %+v.  raw buffer response: %s", err, stdout)
-	}
-
-	rv.VolSize = imgInfo.Size
-	rv.ParentName = imgInfo.Parent.Image
-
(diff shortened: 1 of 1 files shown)


internal/rbd/rbd_util.go
```diff
@@ -979,11 +921,36 @@ func (rv *rbdVolume) getImageInfo() error {
...
+	parentSnapname := make([]byte, 128)
+	err = image.GetParentInfo(parentPool, parentName, parentSnapname)
+	if err != nil {
+		// Caller should decide whether not finding
+		// the parent is an error or not.
+		if strings.Contains(err.Error(), "No such file or directory") {
+			rv.ParentName = ""
+		} else {
+			return err
+		}
```
Madhu-1, line 939: why go-ceph is returning an error if the parent is not present. isn't it's expected to return success with an empty parent name?

agarwal-mudit: ceph is returning ENOENT: https://github.com/ceph/ceph/blob/103fd4af216820fefa12a2ec9576f9eb305fc05f/src/librbd/api/Image.cc#L244

internal/rbd/rbd_util.go
```diff
@@ -979,11 +921,36 @@ func (rv *rbdVolume) getImageInfo() error {
 		return err
 	}
 	rv.imageFeatureSet = librbd.FeatureSet(features)
-	err = rv.updateVolWithImageInfo(rv.conn.Creds)
+
+	// Get parent information.
+	// TODO: Replace GetParentInfo() after
+	// https://github.com/ceph/go-ceph/issues/347 is fixed.
```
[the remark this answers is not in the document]
agarwal-mudit: So, the current implementation of GetParentinfo() is not good because of the reaons mentioned in that go-ceph issue.
Plan is to add a new function, once we have that (in go-ceph v0.6) we can replace it with that.

internal/rbd/rbd_util.go
```diff
@@ -979,11 +921,36 @@ func (rv *rbdVolume) getImageInfo() error {
...
+	parentPool := make([]byte, 128)
+	parentName := make([]byte, 128)
+	parentSnapname := make([]byte, 128)
+	err = image.GetParentInfo(parentPool, parentName, parentSnapname)
+	if err != nil {
+		// Caller should decide whether not finding
+		// the parent is an error or not.
+		if strings.Contains(err.Error(), "No such file or directory") {
```
Madhu-1, line 935: go-ceph should return a proper error message which we can use for error comparison instead of string comparison

example
```
if errors.Is(err, librbd.ErrNotFound) {
```

Madhu-1 (commented):
changes look good, need some clarification with error comparison

internal/rbd/rbd_util.go
agarwal-mudit: We don't have that in current implementation of go-ceph, probably we can do that in https://github.com/ceph/go-ceph/issues/347 and keep this PR on hold till then.

@nixpanic @phlogistonjohn

nixpanic: The error string would currently by `rbd: ret=2, No such file or directory`.

`GetParentInfo()` does not call `getError()`, so it will not return `librbd.ErrNotFound`. The errors in `rbd/errors.go` from go-ceph should probably be converted to instances of `rbdError` so that they can match the errors returned by `rbdError()` or `getError()`.

nixpanic (approved):
The change looks like an improvement to me, it does not make things worse than they were. I am fine with merging this now, and updating the error handling later.

@agarwal-mudit if you want to wait with merging this, make it a Draft and update it when go-ceph v0.6 is released.

internal/rbd/rbd_util.go
Madhu-1: yes as it won't affect any functionality. in that case we can take this in and make the error handling changes when new go-ceph release is available

Madhu-1 (approved):

internal/rbd/rbd_util.go
phlogistonjohn: I read over this thread a few times and I'm not entirely sure what the issue is still. ErrNotFound is not meant to be equivalent to ENOENT (although maybe it should be called ErrImageNotFound). Regardless CSI can still match on errno value rather than string. We recently discussed this in go-ceph on [issue 343](github.com/ceph/go-ceph/issues/343)

Making the existing errors in rbd/errors.go based on rbdError but with extra context is a good idea but I'm not sure it would help here. Let's discuss that in an issue over at go-ceph.

agarwal-mudit: Yes, CSI can match on error value rather than string.

The issue here is, because GetParentInfo() currently doesn't call getError() it won't be returning librbd.ErrNotFound else it would have returned that in case of ENOENT and it would have been easier/cleaner to compare with that.

Right now, comparing with actual error value or string are both same.

PR #1456, opened 2020-09-04
ci: update minikube and kubernetes version

Madhu-1:
Updated minikube version in build configuration and kubernetes version in E2E upgrade testing.


pkalever (approved):
LGTM

.travis.yml
```diff
@@ -102,7 +102,7 @@ jobs:
       script:
         - scripts/skip-doc-change.sh || travis_terminate 0;
         - make image-cephcsi || travis_terminate 1;
-        - scripts/travis-functest.sh v1.18.5 --test-cephfs=true
+        - scripts/travis-functest.sh v1.19.0 --test-cephfs=true
```
Yuggupta27, line 105: I just have a minor question, Do we always want to use the latest release version in this job? If yes, should we directly fetch it to avoid manual updates in the future?

pkalever:
@Madhu-1 while MINIKUBE, GOLANG, and KUBE are fast developing projects can we add some intelligence to make build.env prepare dynamically as part of Makefile Job?

The advantage here is, we can avoid this additional maintenance of updating version values every time some new release of dependencies happens.

While some people still argue consuming the latest versions of GOLANG, MINIKUBE and K8s might introduce regressions in CSI. Yes, if there is a regression in any of these dependency new releases, the CI will catch them and alert us.

In the practical implementation, say if we have a regression with the new GOLANG release we can implement to pick the static version of GOLANG that we provide rather than picking the on-the-fly.

Thoughts?

.travis.yml
Madhu-1: not always it's better to test with the latest version. and these jobs will  be moved to centos ci and we don't need it here 
PR's are always welcome to make work simplier

Madhu-1:
> @Madhu-1 while MINIKUBE, GOLANG, and KUBE are fast developing projects can we add some intelligence to make build.env prepare dynamically as part of Makefile Job?
>

We can make it for MINIKUBE and KUBE not for GOLANG as it's not easy to fix regression in GOLANG versions.
 
> The advantage here is, we can avoid this additional maintenance of updating version values every time some new release of dependencies happens.
> 

> While some people still argue consuming the latest versions of GOLANG, MINIKUBE and K8s might introduce regressions in CSI. Yes, if there is a regression in any of these dependency new releases, the CI will catch them and alert us.
> 

Yes that correct, the problem is that CI will start and this can cause an issue for all the PR's which are in the queue. once the problem with this is it can cause super delays in merging PR's and doing releases as it won't allow PR to be merged manually.


> In the practical implementation, say if we have a regression with the new GOLANG release we can implement to pick the static version of GOLANG that we provide rather than picking the on-the-fly.
> 
> Thoughts?

Better to open an issue and get opinions from People.

.travis.yml
Yuggupta27: Okay, Thanks.

pkalever:
> > @Madhu-1 while MINIKUBE, GOLANG, and KUBE are fast developing projects can we add some intelligence to make build.env prepare dynamically as part of Makefile Job?
> 
> We can make it for MINIKUBE and KUBE not for GOLANG as it's not easy to fix regression in GOLANG versions.

Yeah, we can think on what can be static and what can be fetched on-the-fly, and improve it incrementally.

> 
> > The advantage here is, we can avoid this additional maintenance of updating version values every time some new release of dependencies happens.
> 
> > While some people still argue consuming the latest versions of GOLANG, MINIKUBE and K8s might introduce regressions in CSI. Yes, if there is a regression in any of these dependency new releases, the CI will catch them and alert us.
> 
> Yes that correct, the problem is that CI will start and this can cause an issue for all the PR's which are in the queue. once the problem with this is it can cause super delays in merging PR's and doing releases as it won't allow PR to be merged manually.

I agree on this, but we can retrigger the effected PRs as always.

> 
> > In the practical implementation, say if we have a regression with the new GOLANG release we can implement to pick the static version of GOLANG that we provide rather than picking the on-the-fly.
> > Thoughts?
> 
> Better to open an issue and get opinions from People.

Sure I will open one, Thanks for sharing your opinion @Madhu-1

humblec (approved):

Issue #1338, opened 2020-08-08
Accomodate `--retain-snaps` and `auto-protect` functionalities to CephFS CSI snapshot operations.


humblec:
# Describe the bug #

There were some good enhancements in CephFS ( :+1: ) recently to support snapshot and clone workflow.

Decoupling of snapshot from parent subvol: https://github.com/ceph/ceph/pull/35647

Protect and unprotect deprecation : https://github.com/ceph/ceph/pull/35755

As part of the decoupling work of snapshot and parent sub volume, and deprecation of protect and unprotect, couple of new flags has been added to the subvolume management. 

For ex: `--retain-snaps` flag and an additional feature `autoproect`...etc.

We have to accommodate these changes here in this repo so that we are in parity with upstream cephfs and moving along. The current code wont break but we should consume this for smooth operation snapshot managment. 

Cc @ShyamsundarR

humblec:
https://docs.ceph.com/docs/master/cephfs/fs-volumes/ 

So we have to make use of `features` field in subvol info and operate accordingly.

humblec:
As discussed last week , I am assigning to you @agarwal-mudit

agarwal-mudit:
Hi @ShyamsundarR,

Just a couple of questions for clarification:

-  --retain-snaps is a flag while auto-protect is a subvolume feature?
- Both should be used while creating a subvolume not later or before?
- What about the existing subvolumes, pardon me if that is already discussed 

If we already have a pvc based on subvolume which was not created using these feature/flag then we have a problem.
And, because of such cases we can't remove protect/unprotect completely but have to add check and call protect/unprotect
only for such vols which were not created using these feature/flag.
 
And I guess retain-snaps shouldn't be a big problem in such case because if there is a volume without the --retain-snaps flag, worst case we won't be able to delete it without deleting the snapshots. And we can document the same.

Please correct me if I am wrong or if I have missed anything relevant here.

Madhu-1:
What is the minimum required ceph version for this one?

the changes we do should be backward compatible for older volumes (we should be able to do all operation on older subvolumes and snapshots)

agarwal-mudit:
@ShyamsundarR 
I just realized that --retain-snaps needs to be provided while deleting a subvolume not while creating it, so in that case will it work with a subvolume which was created in a release where this feature was not present?

Yuggupta27:
[...]

PR #1458, opened 2020-09-07
cephfs: Add support for snapshot retention

Yuggupta27:
Snapshots can be retained even after subvolume deletion in
Ceph v14.2.12.
Add `--retain-snapshots` flag to purge volume operations
for retaining snapshots.

Updates: https://github.com/ceph/ceph-csi/issues/1338

### DIFF
--- a/internal/cephfs/volume.go
+++ b/internal/cephfs/volume.go
@@ -269,3 +269,16 @@ func purgeVolume(ctx context.Context, volID volumeID, cr *util.Credentials, volO
 
 	return nil
 }
+
+// checkSubvolumeHasFeature verifies if the referred subvolume has
+// the required feature.
+func checkSubvolumeHasFeature(feature string, subVolFeatures []string) bool {
+	// The subvolume "features" are based on the internal version of the subvolume.
+	// Verify if subvolume supports the required feature.
+	for _, subvolFeature := range subVolFeatures {
+		if subvolFeature == feature {
+			return true
+		}
+	}
+	return false
+}
--- a/internal/cephfs/volume.go
+++ b/internal/cephfs/volume.go
@@ -42,13 +42,19 @@ const (
 type Subvolume struct {
 	BytesQuota    int      `json:"bytes_quota"`
 	DataPool      string   `json:"data_pool"`
+	Features      []string `json:"features"`
 	GID           int      `json:"gid"`
 	Mode          int      `json:"mode"`
 	MonAddrs      []string `json:"mon_addrs"`
 	Path          string   `json:"path"`
 	PoolNamespace string   `json:"pool_namespace"`
-	Type          string   `json:"type"`
-	UID           int      `json:"uid"`
+	// The subvolume "state" is based on the current state of the subvolume.
+	// It contains one of the following values:
+	// * "complete": subvolume is ready for all operations.
+	// * "snapshot-retained": subvolume is removed but its snapshots are retained.
+	State string `json:"state"`
+	Type  string `json:"type"`
+	UID   int    `json:"uid"`
 }
 
 func getVolumeRootPathCephDeprecated(volID volumeID) string {
@@ -247,6 +253,9 @@ func purgeVolume(ctx context.Context, volID volumeID, cr *util.Credentials, volO
 	}
 	if force {
 		arg = append(arg, "--force")
+	}
+	if checkSubvolumeHasFeature("snapshot-retention", volOptions.Features) {
+		arg = append(arg, "--retain-snapshots")
 	}
 
 	err := execCommandErr(ctx, "ceph", arg...)
--- a/internal/cephfs/volumeoptions.go
+++ b/internal/cephfs/volumeoptions.go
@@ -47,6 +47,7 @@ type volumeOptions struct {
 	KernelMountOptions  string `json:"kernelMountOptions"`
 	FuseMountOptions    string `json:"fuseMountOptions"`
 	SubvolumeGroup      string
+	Features            []string
 }
 
 func validateNonEmptyField(field, fieldName string) error {
@@ -224,6 +225,7 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 		vi         util.CSIIdentifier
 		volOptions volumeOptions
 		vid        volumeIdentifier
+		info       Subvolume
 	)
 
 	// Decode the VolID first, to detect older volumes or pre-provisioned volumes
@@ -299,11 +301,18 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 	}
 
 	volOptions.ProvisionVolume = true
-	volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
-	if err != nil {
-		return &volOptions, &vid, err
+
+	info, err = getSubVolumeInfo(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+	if err == nil {
+		volOptions.RootPath = info.Path
+		volOptions.Features = info.Features
 	}
-	return &volOptions, &vid, nil
+
+	if errors.Is(err, ErrInvalidCommand) {
+		volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+	}
+
+	return &volOptions, &vid, err
 }
 
 // newVolumeOptionsFromMonitorList generates a new instance of volumeOptions and


internal/cephfs/controllerserver.go
```diff
@@ -712,3 +712,26 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 
 	return &csi.DeleteSnapshotResponse{}, nil
 }
+
+// validateSnapshotRetention verifies if the referred subvolume has
+// 'snapshot-retention' feature.
+func validateSnapshotRetention(ctx context.Context, volID volumeID, cr *util.Credentials, volOptions *volumeOptions) (bool, error) {
```
Madhu-1, line 718: make this function more generalized. pass `snapshot-retention` as an argument and check the subvolume feature is present

internal/cephfs/volume.go
```diff
@@ -42,11 +42,13 @@ const (
 type Subvolume struct {
 	BytesQuota    int      `json:"bytes_quota"`
 	DataPool      string   `json:"data_pool"`
+	Features      []string `json:"features"`
 	GID           int      `json:"gid"`
 	Mode          int      `json:"mode"`
 	MonAddrs      []string `json:"mon_addrs"`
 	Path          string   `json:"path"`
 	PoolNamespace string   `json:"pool_namespace"`
+	State         string   `json:"state"`
```
Madhu-1, line 51: what is this state and where it's getting used in this PR?

internal/cephfs/controllerserver.go
```diff
@@ -712,3 +712,26 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
...
+		// support the intended features.
+		// The Purge operation should then continue without snapshot retention.
+		if errors.Is(err, ErrInvalidCommand) {
+			return false, nil
+		}
+		return false, status.Error(codes.Internal, err.Error())
+	}
+	// check if subvolume supports snapshot retention feature.
+	for _, feature := range info.Features {
+		if feature == "snapshot-retention" {
```
Madhu-1, line 732: why this feature will be set on the subvolume? during create or it will be auto-added by ceph when you create subvolume? please add a comment here

Madhu-1 (changes requested):
Alter the E2E tests to delete PVC if the snapshots are present and then delete the snapshot and make sure the snapshot and subvolumes are deleted at last

internal/cephfs/controllerserver.go
Yuggupta27: Thanks. Will do it.

internal/cephfs/volume.go
Yuggupta27: The subvolume "state" is based on the current state of the subvolume and is also returned now with a subvolume info call. It contains one of the following values:
* "complete": subvolume is ready for all operations
* "snapshot-retained": subvolume is removed but its snapshots are retained

`State` got added as one of the parameters added in the [subvolume](https://github.com/ceph/ceph/blob/master/doc/cephfs/fs-volumes.rst#fs-subvolumes). As soon as PurgeVolume operation is performed retaining the snapshots, the state will change from `complete` to `snapshot-retained`.

The `Subvolume Info` calls also return the same to inform whether certain operations on a subvolume can be done.
For example, If a subvolume is in `snapshot-retained` state, `rm` operation will be disallowed on it.

internal/cephfs/controllerserver.go
Yuggupta27: Thanks. I will add a comment here.

internal/cephfs/volume.go
Madhu-1: Are you doing any operation on this state in this PR?

Yuggupta27: No Madhu, Just added it to keep csi subvolume parameters sync with ceph. Should I remove it for now and add later when used?

Madhu-1: Let it be there. just asked to make sure that am not missing anything when reviewing

Yuggupta27: Thanks :)

internal/cephfs/controllerserver.go
```diff
@@ -712,3 +712,26 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
...
+		// or `subvolume info` command is not supported,it is assumed that these do not
+		// support the intended features.
+		// The Purge operation should then continue without snapshot retention.
+		if errors.Is(err, ErrInvalidCommand) {
+			return false, nil
+		}
+		return false, status.Error(codes.Internal, err.Error())
+	}
+	// check if subvolume supports snapshot retention feature.
+	for _, feature := range info.Features {
```
humblec, line 731: I would like to keep this as a seperate function where the input string is "feature name", so that it can be reused later. @Yuggupta27

internal/cephfs/volume.go
humblec, line 51: @Yuggupta27 please feel free to add a comment about this field and its optional values. so that it will be useful later.

Yuggupta27: Sure Humble, will add it.

internal/cephfs/controllerserver.go
Yuggupta27: Thanks. Will make it reusable.

humblec:
It looks good, also we have to make sure we have a test case which take care of this scenario and no backward compatibility is broken

Madhu-1:
currently, we are calling getRootPath at a common function https://github.com/ceph/ceph-csi/blob/edca3188288c0aa9fe9732fb0dd4eed5a19f7543/internal/cephfs/volumeoptions.go#L302 before deleting the subvolume as this PR also adds a new network call for subvolume info this can add a little more performance overhead( if we delete 100 subvolumes parallel we will make extra 100 CLI calls for this one. I would like to see the optimization of the above code. do `subvolume info` and if the command is not supported then only do getRootPath with this one we will make no extra network calls. 

@Yuggupta27 JFYI `subvolume info` also returns the `Path` make use of it

humblec (dismissed):

internal/cephfs/volume.go
```diff
@@ -248,8 +254,15 @@ func purgeVolume(ctx context.Context, volID volumeID, cr *util.Credentials, volO
 	if force {
 		arg = append(arg, "--force")
 	}
+	retainSnap, err := validateSubvolumeFeature(ctx, "snapshot-retention", volID, cr, volOptions)
```
Madhu-1, line 257: can we rename `validateSubvolumeFeature` to `IsSubVolumeHasFeature` to better readability?

internal/cephfs/volumeoptions.go
```diff
@@ -24,6 +24,8 @@ import (
 	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
Madhu-1, line 28: don't return GRPC error from this file

Yuggupta27:
@Madhu-1 , As currently, we use ceph v14.2.10 for our e2e tests(which does not support the snapshot retention feature), An attempt to delete a subvolume with snapshots throws an error(as not supported yet). Thus we need to delete the snapshots before deleting the PVC(for now).

Although, I tested it locally with ceph master and the snapshot deletion, in the end, does not throw any error and e2e passes. 
Once we upgrade to the latest Ceph release(Having snapshot retention feature), we can defer snapshot deletion in the end for our e2e tests too.

internal/cephfs/volumeoptions.go
```diff
@@ -299,10 +302,18 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
...
+			volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+			if err != nil {
+				return &volOptions, &vid, err
+			}
+			return &volOptions, &vid, nil
+		}
+		return &volOptions, &vid, status.Error(codes.Internal, err.Error())
 	}
+	volOptions.RootPath = info.Path
 	return &volOptions, &vid, nil
```
Madhu-1, line 317: Avoid multiple reaturns

```suggestion
info, err = getSubVolumeInfo(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
	// This will be overridden if the err is ErrInvalidCommand
	volOptions.RootPath = info.Path
	if err != nil {
		if errors.Is(err, ErrInvalidCommand) {
			volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
		}
	}

	if err != nil {
		return &volOptions, &vid, err
	}
	return &volOptions, &vid, nil
```

Madhu-1 (changes requested):
please change the E2E changes to delete parent PVC when snapshots are present and restore snapshots to new PVC

internal/cephfs/volumeoptions.go
Yuggupta27: Thanks, Will make the change.

internal/cephfs/volume.go
Yuggupta27: maybe `checkSubvolumeHasFeature` ?

Madhu-1: sounds good.

Yuggupta27:
> please change the E2E changes to delete parent PVC when snapshots are present and restore snapshots to new PVC

In that case, the e2e test will fail now due to the reason as snapshot retention is not available in ceph v14.2.10 :( 
See https://github.com/ceph/ceph-csi/pull/1458#issuecomment-689339093

internal/cephfs/controllerserver.go
```diff
@@ -712,3 +712,27 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 
 	return &csi.DeleteSnapshotResponse{}, nil
 }
+
+// checkSubvolumeHasFeature verifies if the referred subvolume has
+// the required feature.
+func checkSubvolumeHasFeature(ctx context.Context, feature string, volID volumeID, cr *util.Credentials, volOptions *volumeOptions) (bool, error) {
```
Madhu-1, line 718: Sorry I missed this one, This function should be defined in volume.go not in controllerserver.go

Yuggupta27:
Addressed all the comments. Please take a look @Madhu-1 @humblec

Madhu-1:
@Yuggupta27 IMO we should not merge a PR when the functionality is not available on any ceph release or based on some manual testing (as things can be tested in CI). if others agree we can merge it.

humblec:
@Madhu-1 I am fine with merging with the manual testing effort on this too. Atleast we know the backward compatibility is not broken.. thats good enough..

humblec (dismissed):

internal/cephfs/volume.go
```diff
@@ -260,3 +277,27 @@ func purgeVolume(ctx context.Context, volID volumeID, cr *util.Credentials, volO
...
+	info, err := getSubVolumeInfo(ctx, volOptions, cr, volID)
+	if err != nil {
+		// For older Ceph versions where 'features' was not returned in the info output,
+		// or `subvolume info` command is not supported,it is assumed that these do not
+		// support the intended features.
+		// The Purge operation should then continue without snapshot retention.
+		if errors.Is(err, ErrInvalidCommand) {
+			return false, nil
+		}
+		return false, status.Error(codes.Internal, err.Error())
```
Madhu-1, line 293: we should not return the GRPC error code from volume.go file

Yuggupta27:
> @Yuggupta27 please provide the test output and the scenarios you have tested
> 
> @ShyamsundarR please review this one.

Thanks for the concern Madhu. As per your suggestion, I tested the deletion of subvolume containing snapshots which seems to work fine for the ceph master. If we want to remove the doubts, I can add a temporary commit where we can test with `ceph-master` on the centos ci and verify this. Will that be fine?

internal/cephfs/volume.go
Yuggupta27: Oh my bad. Will get rid of it.

Madhu-1:
> > @Yuggupta27 please provide the test output and the scenarios you have tested
> > @ShyamsundarR please review this one.
> 
> Thanks for the concern Madhu. As per your suggestion, I tested the deletion of subvolume containing snapshots which seems to work fine for the ceph master. If we want to remove the doubts, I can add a temporary commit where we can test with `ceph-master` on the centos ci and verify this. Will that be fine?

Yes please that gives clarity, please update E2E to create subvolume, create a snapshot, delete subvolume, restore a snapshot to a new PVC and delete the snapshot and delete newly created PVC.

Yuggupta27:
> > > @Yuggupta27 please provide the test output and the scenarios you have tested
> > > @ShyamsundarR please review this one.
> > 
> > 
> > Thanks for the concern Madhu. As per your suggestion, I tested the deletion of subvolume containing snapshots which seems to work fine for the ceph master. If we want to remove the doubts, I can add a temporary commit where we can test with `ceph-master` on the centos ci and verify this. Will that be fine?
> 
> Yes please that gives clarity, please update E2E to create subvolume, create a snapshot, delete subvolume, restore a snapshot to a new PVC and delete the snapshot and delete newly created PVC.

No issue. I'll add a temporary commit to verify it :)

internal/cephfs/volumeoptions.go
```diff
@@ -299,7 +300,13 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 	}
 
 	volOptions.ProvisionVolume = true
-	volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+	info, err = getSubVolumeInfo(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+	volOptions.RootPath = info.Path
```
ShyamsundarR, line 304: Processing `info.Path` in case there are errors is not useful. I would put this into an else of the err check.

internal/cephfs/volumeoptions.go
```diff
@@ -299,7 +300,13 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
...
+	if err != nil {
+		if errors.Is(err, ErrInvalidCommand) {
+			volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+		}
+	}
```
ShyamsundarR, line 309: Store `info->Features` into `volOptions->Features`, such that in `checkSubvolumeHasFeature` function, you can invoke `getSubVolumeInfo` in case `Features` is not populated. Saves a call to ceph in the workflow.

internal/cephfs/volume.go
ShyamsundarR: > The `Subvolume Info` calls also return the same to inform whether certain operations on a subvolume can be done.
> For example, If a subvolume is in `snapshot-retained` state, `rm` operation will be disallowed on it

Subvolume rm will not fail (it will be a No-OP), if removed again and state is "snapshot-retained" (assuming of course rm is with `--retain-snapshots` option). Other than that, storing the state is fine, can be leveraged as needed later.

internal/cephfs/volumeoptions.go
Madhu-1: I suggested the same somehow I missed this one https://github.com/ceph/ceph-csi/pull/1458#issuecomment-688729196 :facepalm:

internal/cephfs/volumeoptions.go
Yuggupta27: Sure, will change it.

internal/cephfs/volumeoptions.go
Yuggupta27: Thanks Shyam, I'll make the changes.

Yuggupta27:
To Validate snapshot retention on ceph nautilus devel, the Following steps are added to e2e:
1. Create PVC and app
2. Create Snapshot
3. Delete the PVC and app
4. Create a PVC from snapshot
5. Delete the PVC and app
6. Delete the snapshot

e2e/cephfs.go
```diff
@@ -440,6 +441,100 @@ var _ = Describe("cephfs", func() {
...
+					e2elog.Logf("failed to get server version with error %v", err)
+					Fail(err.Error())
+				}
+				// pvc clone is only supported from v1.16+
+				if v.Major > "1" || (v.Major == "1" && v.Minor >= "17") {
+					pvcClone, err := loadPVC(pvcClonePath)
+					if err != nil {
+						Fail(err.Error())
+					}
+					pvcClone.Namespace = f.UniqueName
```
Madhu-1, line 504: where you are setting the snapshot name on pvcClone object?

internal/cephfs/volume.go
```diff
@@ -260,3 +274,31 @@ func purgeVolume(ctx context.Context, volID volumeID, cr *util.Credentials, volO
...
+			// or `subvolume info` command is not supported,it is assumed that these do not
+			// support the intended features.
+			// The Purge operation should then continue without snapshot retention.
+			if errors.Is(err, ErrInvalidCommand) {
+				return false, nil
+			}
+			return false, err
+		}
+		subVolFeatures = info.Features
+	}
```
Madhu-1, line 295: This part of the code is not required. let the user fetch the subvolume features and call this function

internal/cephfs/volumeoptions.go
```diff
@@ -299,7 +301,15 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
 	}
 
 	volOptions.ProvisionVolume = true
-	volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+	info, err = getSubVolumeInfo(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+	if err != nil {
```
Madhu-1, line 305: multiple if else if can be converted as below?

```go
info, err = getSubVolumeInfo(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
if err == nil {
 volOptions.RootPath = info.Path
 volOptions.Features = info.Features
}

if err != nil && errors.Is(err, ErrInvalidCommand) {
volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
}

return &volOptions, &vid, err
```

Madhu-1 (changes requested):

internal/cephfs/volume.go
Yuggupta27: Added it as suggested by Shayam in https://github.com/ceph/ceph-csi/pull/1458#discussion_r485671411 . This will make the function fail safe even if voloptions.Features is not populated. Should I get rid of it?

Madhu-1: Yes please, as this would be a helper function we should check only passed feature is present in `volOptions.Features` array

e2e/cephfs.go
Yuggupta27: Chaning name was not required as the snapshot name set [here](https://github.com/ceph/ceph-csi/pull/1458/files/a07b47b5fe5ecc4e887d0703860a1a8bfca51403#diff-4c5b6240068e5c89f188890303acf08fR478)  it is the same mentioned in it's YAML.

internal/cephfs/volume.go
Yuggupta27: Okay, will do it.

e2e/cephfs.go
Madhu-1: ah okay you are making use of the same name :+1:

Madhu-1:
@Yuggupta27 as E2E is passing please remove the extra commit. let's get this in

internal/cephfs/volumeoptions.go
Yuggupta27: we also want to return VolumeOptions for successful case, which brings us to this :
```
info, err = getSubVolumeInfo(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
if err == nil {
 volOptions.RootPath = info.Path
 volOptions.Features = info.Features
}

if err != nil && errors.Is(err, ErrInvalidCommand) {
volOptions.RootPath, err = getVolumeRootPathCeph(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
}
if err != nil {
	return &volOptions, &vid, err
}
	return &volOptions, &vid, nil
```
This will result in nearly similar situation we have now. Should I change it?

Madhu-1:
```go
// If error other than ErrInvalidCommand
if err != nil {
	return &volOptions, &vid, err
}
```
volumeOptions are returned in all the cases why this check is required?

Yuggupta27: Sorry, I am a bit confused. In https://github.com/ceph/ceph-csi/pull/1458#discussion_r486218737 , How are we returning the success case i.e `return &volOptions, &vid, nil` . Please correct me if I am missing something

Madhu-1: err will be nil if `getSubVolumeInfo` succeeded or if it's an ErrInvalidCommand it will be overridden by `getVolumeRootPathCeph` so `err` will be set to `nil` in both success cases.

Yuggupta27: Oh Got it. Thanks for explaining :)

internal/cephfs/volumeoptions.go
```diff
@@ -299,11 +301,18 @@ func newVolumeOptionsFromVolID(ctx context.Context, volID string, volOpt, secret
...
-		return &volOptions, &vid, err
+
+	info, err = getSubVolumeInfo(ctx, &volOptions, cr, volumeID(vid.FsSubvolName))
+	if err == nil {
+		volOptions.RootPath = info.Path
+		volOptions.Features = info.Features
 	}
-	return &volOptions, &vid, nil
+
+	if err != nil && errors.Is(err, ErrInvalidCommand) {
```
Madhu-1, line 311: just a nit this can be simplified as `if  errors.Is(err, ErrInvalidCommand) {`

Madhu-1 (commented):
small nit, LGTM

Madhu-1 (approved):

humblec:
LGTM. However will wait for @ShyamsundarR to have a quick final check as he aleady reviewed the PR.

ShyamsundarR (approved):

Issue #291, opened 2019-03-27
Can't pass any map options to rbd map


Nefelim4ag:
https://github.com/ceph/ceph-csi/blob/csi-v1.0/pkg/rbd/rbd_attach.go#L295-L300

As example we use that options to pass: `osd_request_timeout=300,queue_depth=1024`

Madhu-1:
@Nefelim4ag currently we don't support passing extra options to rbd commands. 

@rootfs @gman0 @ShyamsundarR do we need to consider this use case?

rootfs:
yes, csi mount option supported in kubernetes https://github.com/kubernetes/kubernetes/pull/67898

ShyamsundarR:
The options mentioned (osd_request_timeout, queue_depth) seem to be rbd options that have to be passed to the `rbd map` command when using the krbd client. Possibly other clients also support these options in different ways, but my investigation has been till [krbd](http://docs.ceph.com/docs/mimic/man/8/rbd/#kernel-rbd-krbd-options) client.

The code that you link to also points to `rbd map` so we are in sync till here as to who the options should be passed to.

Now, when I trace the [mount options](https://kubernetes.io/docs/concepts/storage/storage-classes/#mount-options) from the StorageClass to the in tree attacher for rbd (i.e what exists outside the scope of CSI till now), I see that these mount options are passed to the [FS mounter](https://github.com/kubernetes/kubernetes/blob/98e41b7038a3333f72e1ab19b6c6346143f18aa4/pkg/volume/rbd/attacher.go#L176) rather than the [rbd map](https://github.com/kubernetes/kubernetes/blob/98e41b7038a3333f72e1ab19b6c6346143f18aa4/pkg/volume/rbd/rbd_util.go#L219) command.

Thus, I do not understand if your request is to support RBD map options, or that these were supported in the past and we need to continue supporting them in the future. As I do not see them as supported in the past.

@Nefelim4ag Request clarification on the same.

Further, from the use-case of mount options, we should rightly be passing this to the FS mounter, and we [are](https://github.com/ceph/ceph-csi/blob/ac9b656d8117b9f08ae8114a5b4c5bee2b7beb4a/pkg/rbd/nodeserver.go#L148) in the CSI plugin as well. This cannot be overloaded to add RBD mount options IMO.

If the request is that the RBD CSI plugin should support RBD options in addition to existing options, we may need to consider this as a new option to be added to the StorageClass and evaluate the same.

[...]

PR #1450, opened 2020-09-01
rbd: add support for network namespaces (Multus CNI)

idryomov:
Depends on https://github.com/ceph/ceph/pull/36927, see discussion in https://github.com/ceph/ceph-csi/issues/1323.

Only compile tested at this point.

PR #1460, opened 2020-09-08
Add support for rbd map options

Madhu-1:
Added a new parameter in rbd storage-class to pass the rbd map options.

fixes https://github.com/ceph/ceph-csi/issues/291


humblec:
@Madhu-1 the rbd map options can also be depend on `krbd` or `nbd` mounters , Isnt it ? if thats the case its better to have a seggregation either in the input param from "SC" or while applying it.   @dillaman @idryomov thoughts ? 
May be we also have to consider the common map options and default map options too.

examples/rbd/storageclass.yaml
```diff
@@ -29,6 +29,11 @@ parameters:
    # CSI RBD currently supports only `layering` feature.
    imageFeatures: layering
 
+   # mapOptions is a comma-separated list of map options.
+   # for krbd options refer
```
pkalever, line 33: Minor, cosmetic
s/# for krbd/# For krbd/

pkalever: Also, the CI complains there is a trailing-spaces at the end of this line.

pkalever (commented):
Rest all looks good to me.

docs/deploy-rbd.md
```diff
@@ -55,7 +55,8 @@ make image-cephcsi
...
 | `snapshotNamePrefix`                                                                                | no                   | Prefix to use for naming RBD snapshot images (defaults to `csi-snap-`).                                                                                                                                                 |
-| `imageFeatures`                                                                                     | no                   | RBD image features. CSI RBD currently supports only `layering` feature. See [man pages](http://docs.ceph.com/docs/nautilus/man/8/rbd/#cmdoption-rbd-image-feature)                                                         |
+| `imageFeatures`                                                                                     | no                   | RBD image features. CSI RBD currently supports only `layering` feature. See [man pages](http://docs.ceph.com/docs/nautilus/man/8/rbd/#cmdoption-rbd-image-feature)                                                      |
```
pkalever, line 58: IMHO the above line 58 has nothing to do with this change.

Madhu-1:
> @Madhu-1 the rbd map options can also be depend on `krbd` or `nbd` mounters , Isnt it ? if thats the case its better to have a seggregation either in the input param from "SC" or while applying it. @dillaman @idryomov thoughts ?
> May be we also have to consider the common map options and default map options too.

As currently no map options are passed i would like to keep as it is for now, when we are making the default options we can have segregated options.

docs/deploy-rbd.md
Madhu-1: as the line length changes the table gets auto adjusted are you seeing any issue when viewing .md file?

pkalever: Ah! got it, no the md file view looks good. thanks for the clarification.

pkalever (approved):
LGTM

docs/deploy-rbd.md
```diff
@@ -55,7 +55,8 @@ make image-cephcsi
...
+| `mapOptions`                                                                                        | no                   | Map options to use when mapping rbd image. See [man pages](https://docs.ceph.com/docs/octopus/man/8/rbd/#kernel-rbd-krbd-options).                                                                                      |
```
idryomov, line 59: Nit: link to the master version (i.e. https://docs.ceph.com/docs/master/man/8/rbd/#kernel-rbd-krbd-options), here and on the previous line.

examples/rbd/storageclass.yaml
```diff
@@ -29,6 +29,11 @@ parameters:
    # CSI RBD currently supports only `layering` feature.
    imageFeatures: layering
 
+   # mapOptions is a comma-separated list of map options.
+   # For krbd options refer
+   # https://docs.ceph.com/docs/octopus/man/8/rbd/#kernel-rbd-krbd-options
```
idryomov, line 34: Ditto.

internal/rbd/rbd_attach.go
```diff
@@ -235,6 +235,10 @@ func createPath(ctx context.Context, volOpt *rbdVolume, cr *util.Credentials) (s
 	if volOpt.readOnly {
 		mapOptions = append(mapOptions, "--read-only")
 	}
+
+	if volOpt.MapOptions != "" {
+		mapOptions = append(mapOptions, "--options", volOpt.MapOptions)
+	}
```
idryomov, line 241: If the intent is to supersede https://github.com/ceph/ceph-csi/pull/1450, a separate `UnmapOptions` is needed.

Madhu-1: @idryomov this option is to provide more user options, we still need to have defaulted in our code.

>a separate UnmapOptions is needed.

 I thought of adding it, do you have examples I can refer for supported unmap options?

idryomov: It's in the end of that same section, https://docs.ceph.com/docs/master/man/8/rbd/#kernel-rbd-krbd-options.  Currently it's just `force` (which shouldn't be used), but https://github.com/ceph/ceph/pull/36927 adds `noudev`.

Madhu-1: Thanks, @idryomov I have included unmap options also

internal/rbd/rbd_util.go
```diff
@@ -1011,6 +1014,7 @@ func stashRBDImageMetadata(volOptions *rbdVolume, path string) error {
 		RadosNamespace: volOptions.RadosNamespace,
 		ImageName:      volOptions.RbdImageName,
 		Encrypted:      volOptions.Encrypted,
```
ShyamsundarR, line 1016: Technically the version should be bumped above, from 2 to 3.

ShyamsundarR (commented):
Other than the version LGTM!

ShyamsundarR (dismissed):

e2e/rbd.go
```diff
@@ -1240,6 +1240,14 @@ var _ = Describe("RBD", func() {
 				validateRBDImageCount(f, 0)
 			})
 
+			By("create a PVC and Bind it to an app for mapped rbd image with options", func() {
+				deleteResource(rbdExamplePath + "storageclass.yaml")
+				createRBDStorageClass(f.ClientSet, f, nil, map[string]string{"mapOptions": "lock_on_read,queue_depth=1024"})
```
idryomov, line 1245: I'd add an unmap option here to cover `unmapOptions` stashing code as well.

Madhu-1: as you said earlier it's not good idea to use `force` so I have not added unmap options. let me update the storageclass for force options

idryomov: It's fine to use in a E2E test.  It's a supported option after all -- just one that is rarely useful (and particularly in the CSI case, where the details of mapping and unmapping are hidden from the end user).

Madhu-1: added now.

Madhu-1:
@idryomov added E2E for unmapOptions also

humblec (dismissed):

Madhu-1:
@ShyamsundarR PTAL

idryomov (approved):
LGTM.  My only nit is that I'd squash everything but "doc: replace nautilus with master in ceph doc link" into a single commit.

docs/deploy-rbd.md
```diff
@@ -55,7 +55,9 @@ make image-cephcsi
...
-| `imageFeatures`                                                                                     | no                   | RBD image features. CSI RBD currently supports only `layering` feature. See [man pages](http://docs.ceph.com/docs/nautilus/man/8/rbd/#cmdoption-rbd-image-feature)                                                         |
+| `imageFeatures`                                                                                     | no                   | RBD image features. CSI RBD currently supports only `layering` feature. See [man pages](http://docs.ceph.com/docs/nautilus/man/8/rbd/#cmdoption-rbd-image-feature)                                                      |
+| `mapOptions`                                                                                        | no                   | Map options to use when mapping rbd image. See [man pages](https://docs.ceph.com/docs/master/man/8/rbd/#kernel-rbd-krbd-options).                                                                                       |
```
nixpanic, line 59: Is https://docs.ceph.com/docs/master/man/8/rbd-nbd/#options not possible too? Best to mention that the options depend on the `mounter`

e2e/rbd.go
```diff
@@ -1240,6 +1240,31 @@ var _ = Describe("RBD", func() {
 				validateRBDImageCount(f, 0)
 			})
 
+			By("create a PVC and Bind it to an app for mapped rbd image with options", func() {
+				err := deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				err = createRBDStorageClass(f.ClientSet, f, nil, map[string]string{
```
nixpanic, line 1248: In this case you need to use krbd mounter (set, or check that it is configured). Maybe add a test for using rbd-ndb too?

internal/rbd/rbd_attach.go
```diff
@@ -235,13 +235,17 @@ func createPath(ctx context.Context, volOpt *rbdVolume, cr *util.Credentials) (s
 	if volOpt.readOnly {
 		mapOptions = append(mapOptions, "--read-only")
 	}
+
+	if volOpt.MapOptions != "" {
+		mapOptions = append(mapOptions, "--options", volOpt.MapOptions)
```
nixpanic, line 240: if NBD is used, this will not work. It needs to be checked.

nixpanic (changes requested):

docs/deploy-rbd.md
Madhu-1: User has to understand that he should use nbd options when he is using nbd driver and use krbd options when  he is using krbd driver

e2e/rbd.go
Madhu-1: there are no tests for nbd driver as of now, when we start supporting nbd we can add/modify tests

internal/rbd/rbd_attach.go
Madhu-1: Any idea why it won't work?

nixpanic: `rbd-nbd map ...` does not seem to have a `--options` parameter. The arguments for kRBD `rbd map` are just different.

e2e/rbd.go
nixpanic: In that case, maybe just force kRBD to be used in this test. Once rbd-nbd gets added, this test should not suddenly start to fail?

docs/deploy-rbd.md
nixpanic: yes, providing links to both sets of options would be good for that reason

internal/rbd/rbd_attach.go
idryomov: > if NBD is used, this will not work. It needs to be checked.

I don't think so.  `rbd device map -t nbd` accepts nbd-specific options (and even if it didn't, we are checking for `""`).  `mapOptions` and `unmapOptions` are low-level, if anything is specified we want them passed through unconditionally.

e2e/rbd.go
Madhu-1: by default cephcsi uses rbd, once we start supporting nbd in CI we need to make code changes in E2E to pass krbd and nbd mounter I think this should be fine.

internal/rbd/rbd_attach.go
idryomov: > `rbd-nbd map ...` does not seem to have a `--options` parameter.

Correct, but `rbd device map -t nbd --options foo,bar ...` would be translated into `rbd-nbd map --foo --bar ...`.

nixpanic: Ah, ok! Thank for clarifying. `--device-type` is used with the `rbd` command, and `rbd-nbd` is executed as a child process.

nixpanic (approved):

humblec (approved):

PR #1446, opened 2020-08-27
rbd: replace cli with go-ceph for snapshot namespace

agarwal-mudit:
cephcsi uses cli for fetching snap list as well as to check the
snapshot namespace, replaced that with go-ceph calls.

Signed-off-by: Mudit Agarwal <muagarwa@redhat.com>

PR #1477, opened 2020-09-15
test: replace cli with go-ceph for snapshot namespace PR

Madhu-1:
This PR is to test changes of https://github.com/ceph/ceph-csi/pull/1446 and to avoid panic in E2E tests

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -415,7 +415,7 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
 	}
 }
 
-func flattenClonedRbdImages(ctx context.Context, snaps []snapshotInfo, pool, monitors string, cr *util.Credentials) error {
+func flattenClonedRbdImages(ctx context.Context, snaps []librbd.SnapInfo, pool, monitors string, cr *util.Credentials) error {
 	rv := &rbdVolume{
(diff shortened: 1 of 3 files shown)


Madhu-1:
This is just a testing PR
cc @agarwal-mudit

agarwal-mudit:
Thanks @Madhu-1

internal/rbd/rbd_util.go
```diff
@@ -426,9 +426,24 @@ func flattenClonedRbdImages(ctx context.Context, snaps []snapshotInfo, pool, mon
 		util.ErrorLog(ctx, "failed to open connection %s; err %v", rv, err)
 		return err
 	}
-	for _, s := range snaps {
-		if s.Namespace.Type == "trash" {
-			rv.RbdImageName = s.Namespace.OriginalName
+
+	for _, snapInfo := range snaps {
+		// check if the snapshot belongs to trash namespace.
+		isTrash, err := rv.isTrashSnap(snapInfo.Id)
```
Madhu-1, line 432: @agarwal-mudit I think this won't be valid as the rbd image name will be updated at line 446 isn't it?

agarwal-mudit: Nice catch @Madhu-1, Yeah that should be an issue here. Let me think of a solution. Thanks for the help.

Issue #1485, opened 2020-09-16
ceph-fuse fails when FuseMountOptions is set


flarno11:
# Describe the bug #

When setting `fuseMountOptions` (e.g. to "debug") in the storageClass, ceph-fuse fails with 
```
MountVolume.MountDevice failed for volume "pvc-xxx" : rpc error: code = Internal desc = an error (exit status 22) occurred while running ceph-fuse args: [/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-xxx/globalmount -m v2:xxx:3300,v2:yyy:3300,v2:1zzz:3300 -c /etc/ceph/ceph.conf -n client.csi-cephfs-node --keyfile=***stripped*** -r /volumes/csi/csi-vol-yyy/xxx -o nonempty ,debug --client_mds_namespace=sbdfs]
```

# Environment details #

- Image/version of Ceph CSI driver : 3.1.0
- Helm chart version : no helm in use
- Kernel version : 3.10.0
- Mounter used for mounting PVC : fuse
- Kubernetes cluster version : 1.17.5
- Ceph cluster version : 15.2.4

# Steps to reproduce #

1. Setup details: create storageClass with `fuseMountOptions: debug`
```
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: cephfs
provisioner: cephfs.csi.ceph.com
parameters:
  clusterID: {{ ceph_csi__cluster_id }}
  fsName: {{ ceph_csi__ceph_fs_name }}
  pool: {{ ceph_cluster__osd_data_pool_name }}
  fuseMountOptions: debug
  ...
```

2. Deployment to trigger the issue: create a PVC and bind it to a POD
3. See error

# Actual results #

See error.

I suspect the error comes from the additional space between `-o nonempty` and ` ,debug` because `FuseMountOptions` is appended to `args` and not to the string "nonempty" in:
https://github.com/ceph/ceph-csi/blob/master/internal/cephfs/volumemounter.go#L161

# Expected behavior #

ceph-fuse command to success with mount options `-o nonempty,debug` (and not `-o nonempty, debug`).

# Logs #

If the issue is in PVC creation, deletion, cloning please attach complete logs
of below containers.

Logs of csi-cephfsplugin:
```
[...]

PR #1499, opened 2020-09-21
cephfs: trim spaces around fuseMountOptions

nixpanic:
When passing

    fuseMountOptions: debug

in the StorageClass, the mount options passed on the ceph-fuse
commandline result in "-o nonempty ,debug". The additional space before
the ",debug" causes the mount command to fail.

Fixes: #1485

### DIFF
--- a/e2e/cephfs_helper.go
+++ b/e2e/cephfs_helper.go
@@ -34,7 +34,7 @@ func validateSubvolumegroup(f *framework.Framework, subvolgrp string) error {
 	return nil
 }
 
-func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, enablePool bool, clusterID string) error {
+func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, enablePool bool, params map[string]string) error {
 	scPath := fmt.Sprintf("%s/%s", cephfsExamplePath, "storageclass.yaml")
 	sc, err := getStorageClass(scPath)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/cephfs/volumemounter.go
```diff
@@ -154,12 +154,13 @@ func mountFuse(ctx context.Context, mountPoint string, cr *util.Credentials, vol
 		"-c", util.CephConfigPath,
 		"-n", cephEntityClientPrefix + cr.ID, "--keyfile=" + cr.KeyFile,
 		"-r", volOptions.RootPath,
-		"-o", "nonempty",
 	}
 
+	fmo := "nonempty"
 	if volOptions.FuseMountOptions != "" {
-		args = append(args, ","+volOptions.FuseMountOptions)
+		fmo += "," + strings.TrimSpace(volOptions.FuseMountOptions)
```
Madhu-1, line 161: @nixpanic is it possible to cover an E2E for this one?

nixpanic: Sure, we could create a StorageClass with the `debug` parameter and mount the PVC.

nixpanic: e2e tests with ceph-fuse have been added, lets see if those work :crossed_fingers:

e2e/cephfs_helper.go
```diff
@@ -53,20 +53,30 @@ func createCephfsStorageClass(c kubernetes.Interface, f *framework.Framework, en
...
+
+	// overload any parameters that were passed
+	if params == nil {
+		// create an empty params, so that params["clusterID"] below
+		// does not panic
+		params = map[string]string{}
+	} else {
+		for param, value := range params {
+			sc.Parameters[param] = value
+		}
```
Madhu-1, line 65: Else can be removed as we are intializing to empty map the for loop wont get executed.
```suggestion
if params == nil {
// create an empty params, so that params["clusterID"] below
// does not panic
   params = map[string]string{}
}
for param, value := range params {
     sc.Parameters[param] = value
}

nixpanic: Sure, if that makes it better to read. Updated in the next version of this PR.

Madhu-1 (approved):

humblec (approved):

PR #1505, opened 2020-09-22
ci: update upgrade tests job

Yuggupta27:
Instead of accessing the upgrade version from the environment
variable, set the variable directly in the configuration file.
Also, use 'make run-e2e' to run the functional tests.

### DIFF
--- a/jobs/upgrade-tests.yaml
+++ b/jobs/upgrade-tests.yaml
@@ -5,6 +5,7 @@
     test_type:
       - 'cephfs'
       - 'rbd'
+    csi_upgrade_version: 'v3.1.0'
     jobs:
       - 'upgrade-tests-{test_type}'
 
@@ -21,6 +22,7 @@
     # default variables
     k8s_version: '<unset>'
     test_type: '<unset>'
+    csi_upgrade_version: '<unset>'
     # generated parameters for the job (used in the groovy script)
     parameters:
       - string:
@@ -31,6 +33,10 @@
           name: test_type
           default: '{test_type}'
           description: Mentions whether upgrade tests run for rbd/cephfs.
+      - string:
+          name: csi_upgrade_version
+          default: '{csi_upgrade_version}'
+          description: Ceph CSI base version to be used before upgrade.
     pipeline-scm:
       scm:
         - git:
--- a/upgrade-tests.groovy
+++ b/upgrade-tests.groovy
@@ -106,7 +106,7 @@ node('cico-workspace') {
 				if ("${test_type}" == "cephfs"){
 					upgrade_args = "--test-cephfs=true --test-rbd=false --upgrade-testing=true"
 				}
-				ssh "cd /opt/build/go/src/github.com/ceph/ceph-csi && source build.env && scripts/travis-functest.sh ${k8s_release} --upgrade-version=\${UPGRADE_VERSION} ${upgrade_args}"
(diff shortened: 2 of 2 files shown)


nixpanic (approved):

Yuggupta27:
Hmm..It seems to show line length too long even though the commit reduces it..... Wondering why it didn't get caught earlier

jobs/upgrade-tests.yaml
```diff
@@ -5,6 +5,7 @@
     test_type:
       - 'cephfs'
       - 'rbd'
+    csi_upgrade_version: 'v3.1.0'
```
nixpanic, line 8: Why v3.1.0? `UPGRADE_VERSION` in `build.env` has v3.0.0.

Yuggupta27: Snapshots functionality for cephfs is available from csi v3.1.0... Due to that, it needs to change.

Madhu-1: why can't we use from build.env itself? or else we need to maintain configuration in 2 places?

Yuggupta27: The variable will then be removed from `build.env` so there won't be a need to maintain it there. Also, it might be better to keep job related variables in it's configuration file. What do you guys think?

nixpanic: > Snapshots functionality for cephfs is available from csi v3.1.0... Due to that, it needs to change.

- Is this the `upgrade-from-version`?
- Is this version different for backports, or do backports not get tested with this job?
- In general, how configurable should this version be, when does the value change?

I guess the answer where to place the version, depends on the answers of the above questions.

Yuggupta27: > > Snapshots functionality for cephfs is available from csi v3.1.0... Due to that, it needs to change.
> 
> * Is this the `upgrade-from-version`?

Yes, So basically we create objects in this version and then upgrade csi to the validate their proper working.

> * Is this version different for backports, or do backports not get tested with this job?

afaik, it runs for backport PRs too.

> * In general, how configurable should this version be, when does the value change?

Not frequently. The changes are being made now as new test cases are added for testing cephfs snapshot clone functionality.
> 
> I guess the answer where to place the version, depends on the answers of the above questions.

nixpanic: I think that having it in the CI job is sufficient. But it should cover the lowest version that can be used for upgrading, or multiple (templated) jobs similar to the Kubernetes version in the mini-e2e ones. Additional tests should have a check for a version, and if the old version does not have the functionality, the test(s) should be skipped. We should make sure that specially upgrade tests can run against almost all versions that are supported for upgrades (possibly without additional newly introduced functionality).

Yuggupta27: Great Idea! Once the upgarde tests with v3.1.0 run fine, We can add 2 more(rbd+cephfs) templated jobs later which will verify the csi upgrade from say(v2.1.0) to latest (running the tests only which the version supports).

Yuggupta27: What do you think @Madhu-1 ?

Madhu-1: sounds good, just make sure the few operations in upgrade E2E need to be done based on the version? like snapshot supported only after 3.0.0 for rbd and 3.1.0 for cephfs etc?

Yuggupta27: Sure, Once the upgrade tests pass, I'll configure 2 more jobs to run tests according to the version that supports them.

nixpanic (approved):

Madhu-1 (approved):

PR #1515, opened 2020-09-24
ci: do not install Docker on the minikube host

nixpanic:
Installing Docker CE on CentOS-8 currently conflicts with Podman. Both
tools can not be installed at the same time.

Instead of using the Docker command to push the image to to minikube VM,
read the image from stdin over ssh and load it with the Docker command
that is available inside the VM.

See-also: #1516

### DIFF
--- a/single-node-k8s.sh
+++ b/single-node-k8s.sh
@@ -75,14 +75,6 @@ function set_env() {
     export PATH=$PATH:/usr/local/bin
 }
 
-# FIXME: unfortunately minikube does not work with podman (yet)
-function install_docker()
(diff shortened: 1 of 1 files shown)


Madhu-1 (approved):

Yuggupta27 (approved):

pkalever:
@nixpanic as I see the doesn't have conditional code, is this change backward compatible with Centos 7?

Madhu-1:
This is only for centos/ci we don't have plan to go back to centos 7 I believe.

PR #1517, opened 2020-09-24
ci: run minikube.sh with podman

nixpanic:
The mini-e2e jobs fail to run `minikube.sh` because if defaults to use Docker. However, on recent CentOS-8 deployments Docker and Podman can not be installed at the same time. Everything should use podman on the host, but the minikube VM can still use Docker.

### DIFF
--- a/single-node-k8s.sh
+++ b/single-node-k8s.sh
@@ -70,6 +70,8 @@ function set_env() {
     # downloading rook images is sometimes slow, extend timeout to 15 minutes
     export ROOK_VERSION='v1.3.9'
     export ROOK_DEPLOY_TIMEOUT=900
+    # use podman for minikube.sh, Docker is not installed on the host
+    export CONTAINER_CMD='podman'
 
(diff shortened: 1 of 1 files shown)


Madhu-1 (approved):

single-node-k8s.sh
```diff
@@ -70,6 +70,8 @@ function set_env() {
     # downloading rook images is sometimes slow, extend timeout to 15 minutes
     export ROOK_VERSION='v1.3.9'
     export ROOK_DEPLOY_TIMEOUT=900
+    # use podman for minikube.sh, Docker is not installed on the host
+    export CONTAINER_CMD='podman'
```
pkalever, line 74: @nixpanic is it a good idea to replace the hardcoded use of `podman` in this file with CONTAINER_CMD variable?

nixpanic: No, we do not really need to have it configurable. And ideally we never use `docker` anywhere.

Issue #1291, opened 2020-07-29
Update Snapshot restore E2E to check  data consistency


Madhu-1:
Currently, in rbd snapshot restore and volume clone E2E we are not checking any data consistency after doing snapshot restore or volume clone.

### For a snapshot 
* create  PVC and App
* create snapshots from PVC
* create new PVC from the snapshot

### For clone
* Create PVC and APP
* Create new PVC from  above PVC

Additional to the above steps, we need  to do a few  extra steps like below

### For a snapshot 
* create  PVC and App
* Write data to the PVC with `dd command` if PVC is `1GB write 1GB data`
* get the md5sum of data
* create snapshots from PVC
* create new PVC from the snapshot
* check the size of a directory  in new POD where new PVC is mounted
* Verify the md5sum of the data is matching

### For clone
* Create PVC and APP
* Write data to the PVC with `dd command` if PVC is `1GB write 1GB data`
* get the md5sum of data
* Create new PVC from  above PVC
* check the size of a directory  in new POD where new PVC is mounted
* Verify the md5sum of the data is matching

PR #1464, opened 2020-09-09
ci: Add Snapshot and PVC cloning test cases to upgrade tests

Yuggupta27:
For rbd and cephfs upgrade tests, add Snapshot cloning and PVC to PVC 
cloning test cases. This might enable us to transition to Centos CI completely.
.

PR #1523, opened 2020-09-25
update snapshot restore e2e to check data consistency

yati1998:
# Updates snapshot restore e2e to check data consistency
Currently, in rbd snapshot restore and volume clone E2E we are not checking any data consistency after doing snapshot restore or volume clone. Hence, this PR writes the data in the PVC and checks the md5sum of the file and verify it with the snapshot or cloned PVC.

Fixes: #1291


e2e/utils.go
```diff
@@ -351,6 +351,28 @@ func writeDataInPod(app *v1.Pod, f *framework.Framework) error {
 	return err
 }
 
+//calculate md5sum
+func calMd5sum(app *v1.Pod, f *framework.Framework) (string, error) {
```
nixpanic, line 355: @Yuggupta27 introduces `calculateSHA512sum()` for similar tests, please check with him and use a single checksum function

Yuggupta27: Yes please, You can find it here https://github.com/ceph/ceph-csi/pull/1464/files#diff-c62628027491c2b96afe9b96aa706363R221

Yuggupta27 (changes requested):
Can you please update commit message as per format mentioned here https://github.com/ceph/ceph-csi/blob/master/docs/development-guide.md#commit-messages

yati1998:
Updated with new PR:  https://github.com/ceph/ceph-csi/pull/1557

Issue #1543, opened 2020-09-29
rbd node service: flattern image when it has references to parent


pkalever:
# Describe the bug #

Currently, as part of node service, we add rbd flatten task for new PVC creates. Ideally, we should add a flatten task only for snapshots/cloned PVCs as required.

```

[0] pkalever 😎 rbd✨ kubectl describe pods csi-rbd-demo-pod
[...]
Events:
  Type     Reason                  Age                  From                     Message
  ----     ------                  ----                 ----                     -------
  Normal   Scheduled               10m                  default-scheduler        Successfully assigned default/csi-rbd-demo-pod to minikube
  Normal   SuccessfulAttachVolume  10m                  attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-cec90cb9-8273-4eb1-b2fa-32d3206a1f7d"
  Warning  FailedMount             113s (x12 over 10m)  kubelet, minikube        MountVolume.MountDevice failed for volume "pvc-cec90cb9-8273-4eb1-b2fa-32d3206a1f7d" : rpc error: code = Internal desc = an error (exit status 2) occurred while running ceph args: [rbd task add flatten rbd-pool/csi-vol-1dae0d96-0238-11eb-93fe-0242ac110004 --id admin --keyfile=***stripped*** -m 192.168.121.136]
  Warning  FailedMount             93s (x4 over 8m21s)  kubelet, minikube        Unable to attach or mount volumes: unmounted volumes=[mypvc], unattached volumes=[mypvc default-token-2xr4n]: timed out waiting for the condition

```

# Environment details #

```
[root@minikube /]# cephcsi --version
Cephcsi Version: canary
Git Commit: fd4328cd5333f4275be52c604c30801fc612fa75
Go Version: go1.15
Compiler: gc
Platform: linux/amd64
Kernel: 4.19.114
[root@minikube /]# 

[root@ceph-node1 ~]# rados lspools
rbd-pool
cephfs-datapool
cephfs-metapool
[root@ceph-node1 ~]# rbd ls -l rbd-pool
NAME                                         SIZE  PARENT FMT PROT LOCK 
csi-vol-e15f6413-023f-11eb-93fe-0242ac110004 1 GiB          2           
image1                                       1 GiB          2           
[root@ceph-node1 ~]# rbd info rbd-pool/csi-vol-e15f6413-023f-11eb-93fe-0242ac110004
[...]

PR #1544, opened 2020-09-29
rbd: flattern image when it has references to parent

pkalever:
Currently, as part of node service we add rbd flatten task for new PVC
creates. Ideally, we should add flatten task only for snapshots/cloned
PVCs as required.

Fixes: #1543
Signed-off-by: Prasanna Kumar Kalever <prasanna.kalever@redhat.com>

### DIFF
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -238,6 +238,7 @@ func (ns *NodeServer) stageTransaction(ctx context.Context, req *csi.NodeStageVo
 	var err error
 	var readOnly bool
 	var feature bool
+	var depth uint
 
 	var cr *util.Credentials
 	cr, err = util.NewUserCredentials(req.GetSecrets())
@@ -273,7 +274,13 @@ func (ns *NodeServer) stageTransaction(ctx context.Context, req *csi.NodeStageVo
 			if err != nil {
 				return transaction, err
 			}
-			if feature {
+
+			depth, err = volOptions.getCloneDepth(ctx)
+			if err != nil {
+				return transaction, err
+			}
+
+			if feature && (depth != 0) {
 				err = volOptions.flattenRbdImage(ctx, cr, true, rbdHardMaxCloneDepth, rbdSoftMaxCloneDepth)
 				if err != nil {
 					return transaction, err


nixpanic:
Not sure what causes the problem, but there seems to be a component missing in the PV-created-from-snapshot:
```
Sep 30 05:26:36.941: INFO: At 2020-09-30 05:16:53 +0100 BST - event for csi-rbd-restore-demo-pod: {kubelet minikube} FailedMount: MountVolume.MountDevice failed for volume "pvc-8b836dd9-8b2d-49ea-91e8-a2fb02a8df83" : rpc error: code = Internal desc = rbd: map failed with error an error (exit status 6) occurred while running rbd args: [--id csi-rbd-node -m rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789 --keyfile=***stripped*** map replicapool/e2e-ns/csi-vol-ba2559b8-02d3-11eb-994a-0242ac110015 --device-type krbd], rbd error output: rbd: sysfs write failed
```

pkalever:
> Not sure what causes the problem, but there seems to be a component missing in the PV-created-from-snapshot:

Yes Neils, and for me, that looked irrelevant to this change. (Or maybe I missing something)
I'm yet to look at the other logs though.

nixpanic:
> for me, that looked irrelevant to this change

For me too, but I doubt it is coincidence that the rbd e2e tests fail, whereas cephfs works.

nixpanic:
@mergifyio rebase

nixpanic (approved):

internal/rbd/nodeserver.go
```diff
@@ -273,7 +274,13 @@ func (ns *NodeServer) stageTransaction(ctx context.Context, req *csi.NodeStageVo
...
 			}
-			if feature {
+
+			depth, err = volOptions.getCloneDepth(ctx)
+			if err != nil {
+				return transaction, err
+			}
+
+			if feature && (depth != 0) {
 				err = volOptions.flattenRbdImage(ctx, cr, true, rbdHardMaxCloneDepth, rbdSoftMaxCloneDepth)
```
Madhu-1, line 284: @pkalever `flattenRbdImage` already handles this error can you please check why its not working?

Madhu-1 (changes requested):
marking for request changes to get analysis

internal/rbd/nodeserver.go
pkalever: Yes Madhu, I remember this is on me, will get back.

pkalever: @Madhu-1  looks like it is well handled in `flattenRbdImage` but as for my understanding it is skipping `getCloneDepth` check because we are calling  `flattenRbdImage` with `forceFlatten=true`.

Is there any specific reason why `stageTransaction` should call `flattenRbdImage` with `forceFlatten=true` ?

If not we should rather call `flattenRbdImage` with `forceFlatten=false`, which should fix it all.

pkalever: @Madhu-1 here is the change:
```
diff --git a/internal/rbd/nodeserver.go b/internal/rbd/nodeserver.go
index a9499df95..ed5664aa0 100644
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -274,7 +274,7 @@ func (ns *NodeServer) stageTransaction(ctx context.Context, req *csi.NodeStageVo
                                return transaction, err
                        }
                        if feature {
-                               err = volOptions.flattenRbdImage(ctx, cr, true, rbdHardMaxCloneDepth, rbdSoftMaxCloneDepth)
+                               err = volOptions.flattenRbdImage(ctx, cr, false, rbdHardMaxCloneDepth, rbdSoftMaxCloneDepth)
                                if err != nil {
                                        return transaction, err
                                }
(END)
```

which solves the issue:

```
[0] pkalever 😎 ceph-cluster✨ kubectl describe pod csi-rbd-demo-pod
[...]
Events:
  Type     Reason                  Age                From                     Message
  ----     ------                  ----               ----                     -------
  Normal   Scheduled               67s                default-scheduler        Successfully assigned default/csi-rbd-demo-pod to minikube
  Normal   SuccessfulAttachVolume  67s                attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-4d0d467d-f98c-4a77-ae34-b6720a45add3"
  Warning  FailedMount             15s (x7 over 55s)  kubelet, minikube        MountVolume.MountDevice failed for volume "pvc-4d0d467d-f98c-4a77-ae34-b6720a45add3" : rpc error: code = Internal desc = rbd: map failed with error an error (exit status 6) occurred while running rbd args: [--id admin -m 192.168.121.120,192.168.121.229 --keyfile=***stripped*** map rbd-pool/csi-vol-a24be4f3-1476-11eb-86e1-0242ac110004 --device-type krbd], rbd error output: rbd: sysfs write failed
```

But I'm unsure if this change causes any regressions, thoughts?

nixpanic: I think just calling `volOptions.flatten()` in the nodeserver is sufficient. The image can be flattened inline, and not through the TaskManager which `flattenRbdImage()` does.

pkalever: I think the idea of adding it to the task is to get this done asynchronously and not to get blocked there for too long, because the time it takes to flatten a clone increases with the size of the snapshot.

nixpanic: Yes, but ideally the task is scheduled when this image got provisioned (in the controllerserver). If it was not done yet for some reason, doing it inline before mounting might result in more predictable performance during runtime (no flattening while the image is mounted and actively used).

Madhu-1: This is only to support the kernels which do not support the flattening feature. we don't want to block the RPC call for flattening the image( as we can face other issues if we block the RPC calls). as @pkalever mentioned the time taken for flattening depends on the size of the data. I will check why the current code is not working and will update.

pkalever: @nixpanic hmm, yes it makes sense up to some extent, especially when the PVC create and attach happens one after the other immediately. But, In case if provisioning of the volume is done much ahead of time before attaching it to the application pod, I still see that adding it as a task is helpful

pkalever: Yes, as mentioned, +1 for the task considering the RPC timeouts too.

Madhu-1: > @Madhu-1 looks like it is well handled in `flattenRbdImage` but as for my understanding it is skipping `getCloneDepth` check because we are calling `flattenRbdImage` with `forceFlatten=true`.
> 
> Is there any specific reason why `stageTransaction` should call `flattenRbdImage` with `forceFlatten=true` ?
>
Yes, we don't want to check any image depth in the node stage Request. we want to flatten all image in node stage request (a special check as been added to discard no parent error). This check is added only to support the kernel's which don't support the deep-flatten image feature.

> If not we should rather call `flattenRbdImage` with `forceFlatten=false`, which should fix it all.

PR #1558, opened 2020-10-05
doc: add design doc for metro DR

Madhu-1:
added a design document for naming the csi volumes to get identified on the secondary clusters in case of metro DR.


docs/design/proposals/metro-dr.md
```diff
@@ -0,0 +1,70 @@
...
+`csi.volume.<uniqueID>.PVName` : `volume UUID` (this is required to identify
+the RBD image/cephFS subvolume belongs to the PV) and creates an RBD image with
+the name `csi-vol-<pvcname>-<namespacename>-<UUID>`.  When doing the failover,
+cephcsi in the secondary cluster will check the volume is already present, If
+the image is present cephcsi will add its request name(PV Name) mapping to the
+UUID (`csi.volume.<uniqueID>.PVName` : `volume UUID`). As ID passed in
+configmap is unique across all the cephcsi instances accessing the ceph cluster
+will help cephcsi to identify the volume.
+
+Currently, we are making use of watchers in node stage request to make sure RWO
```
ShyamsundarR, line 49: The concern of watchers can be moved to the RBD mirroring case, rather than here, as we do not have a mirroring solution in place for this one.

docs/design/proposals/metro-dr.md
```diff
@@ -0,0 +1,70 @@
...
+### For snapshot
+
+This will be the same as the CreateVolume request, here instead of the
+storageclass parameter, cephcsi will check the snapshotclass parameter.
+
+Same as PVC name and namespace name to generate the volume name. We will use
+snapshot name and snapshot namespace name to generate the snapshot name, created
+an [issue](https://github.com/kubernetes-csi/external-snapshotter/issues/341)
+in external snapshotter to pass extra snapshot metadata to create the snapshot
+request.
```
ShyamsundarR, line 70: I would like to add a scenarios section here, with how this would be used. Something along the following lines,

```
**Setup:**
**Scenario 1: Single shared pool across peer kubernetes clusters:**
  - Kubernetes cluster-a
    - ceph-csi config map
      - clusterID: ceph-cluster-foo
        - uniqueID: bar
  - Kubernetes cluster-b
    - ceph-csi config map
      - clusterID: ceph-cluster-foo
        - uniqueID: baz
  - Kubernetes cluster-c
    - ceph-csi config map
      - clusterID: ceph-cluster-foo
        - uniqueID: qux
  - StorageClass-ceph:
    - clusterID: ceph-cluster-foo
    - PVCNaming: true
  PVC:
    - StorageClaassName: StorageClass-ceph

**Scenario 2: Affined pool (pool per peer kubernetes cluster):**
  - Kubernetes cluster-a
    - ceph-csi config map
      - clusterID: ceph-cluster-foo
        - uniqueID: bar
  - Kubernetes cluster-b
    - ceph-csi config map
      - clusterID: ceph-cluster-foo
        - uniqueID: baz
  - Kubernetes cluster-c
    - ceph-csi config map
      - clusterID: ceph-cluster-foo
        - uniqueID: qux
  - SC-a:
    - clusterID: ceph-cluster-foo
    - pool: pool-affined-a
    - PVCNaming: true
  - SC-b:
    - clusterID: ceph-cluster-foo
    - pool: pool-affined-b
    - PVCNaming: true
  - SC-c:
    - clusterID: ceph-cluster-foo
    - pool: pool-affined-c
    - PVCNaming: true
  PVC:
    - SC: SC-a/SC-b/SC-c
      - Possibly derived using kustomize during initial deployment to a chosen affined StorageClass for the cluster

**Scenario 3: Adding a peer cluster to an existing kubernetes cluster with existing provisioned PVCs**
  - NOTE: As existing ceph-csi would already have used the empty `uniqueID`, only one such cluster can be added as a peer to reuse existing provisioned PVCs. Adding more than one such peer would result in more than one of them using the empty `uniqueID` and will result in deployments not getting their storage bound when migrated to peer clusters.
  - Kubernetes cluster-a
    - ceph-csi config map
      - clusterID: ceph-cluster-foo
      - NOTE: no `uniqueID` as cluster was setup prior to DR needs
  - Kubernetes cluster-b
    - ceph-csi config map
      - clusterID: ceph-cluster-foo
        - uniqueID: baz
  - SC-a:
    - clusterID: ceph-cluster-foo
    - PVCNaming: true
      - NOTE: The above parameter should be set prior to provisioning volumes that may in the future be migrated to peer kubernetes clusters, as it is used as the identifying signature to find backend images for the request
  PVC:
    - SC: SC-a
```

docs/design/proposals/metro-dr.md
```diff
@@ -0,0 +1,70 @@
...
+watchers logic in the node stage request as when we enable the RBD mirroring on
+an image a watcher will be added on an RBD image.
+
+In case of the PVC is Retained and recreated with the same name in the same
+namespace as the request name is unique cephcsi will not return the underlying
+storage. But if the PVC is Retained on one cluster and PVC gets created with
+the same name in the same namespace in another cluster will get underlying
+storage (it won't create a new volume (the support for the need to be added
+later by marking RBD/CephFS storage for not to be reused))
+
```
ShyamsundarR, line 60: Can we also add in some pseudo-code as we think this will behave, say:

```
Pseudo code:
  - CreateVolume
    - CheckVolExists
    - if PVCNaming:
      - List and check if image exists, and is still unbound to a request in this cluster, if so bind it and update common image OMap
    - Else:
      - Create a new volume
  - DeleteVolume
    - Determine image OMap, and store it
    - Delete image
    - Delete image OMap
    - Recursively delete CSI OMap keys from RequestName in image OMap
```

docs/design/proposals/metro-dr.md
```diff
@@ -0,0 +1,70 @@
...
+and creates its request name (PV name) and unique ID mapping in omap
+`csi.volume.<uniqueID>.PVName` : `volume UUID` (this is required to identify
+the RBD image/cephFS subvolume belongs to the PV) and creates an RBD image with
+the name `csi-vol-<pvcname>-<namespacename>-<UUID>`.  When doing the failover,
+cephcsi in the secondary cluster will check the volume is already present, If
+the image is present cephcsi will add its request name(PV Name) mapping to the
+UUID (`csi.volume.<uniqueID>.PVName` : `volume UUID`). As ID passed in
+configmap is unique across all the cephcsi instances accessing the ceph cluster
+will help cephcsi to identify the volume.
+
```
ShyamsundarR, line 48: As discussed, add that retaining the same image OMap is useful to preserve and reuse the volumeContext, and hence better, than creating a separate image OMap per peer kubernetes cluster that is sharing this Ceph pool.

docs/design/proposals/metro-dr.md
```diff
@@ -0,0 +1,70 @@
...
+PVC is mounted on a single node at a given point of time. We need to change the
+watchers logic in the node stage request as when we enable the RBD mirroring on
+an image a watcher will be added on an RBD image.
+
+In case of the PVC is Retained and recreated with the same name in the same
+namespace as the request name is unique cephcsi will not return the underlying
+storage. But if the PVC is Retained on one cluster and PVC gets created with
+the same name in the same namespace in another cluster will get underlying
+storage (it won't create a new volume (the support for the need to be added
+later by marking RBD/CephFS storage for not to be reused))
```
ShyamsundarR, line 59: As discussed, add a note about delete volume deleting all OMap entries, and thus clearing up the state for all peer clusters.

ShyamsundarR (commented):
Mostly good, I am just requesting more context in this document. If moving onto code, then we can just handle the context there, than writing up the document as such (possibly the added comments are sufficient for discussion in case anything is off the mark).

docs/design/proposals/metro-dr.md
Madhu-1: i update this one we still need to think about the stale watchers if the unmapping is not done on the cluster before failover)

docs/design/proposals/metro-dr.md
```diff
@@ -0,0 +1,216 @@
...
+When cephcsi receives a request for Delete operation it will delete the backend
+storage and all OMap entries as well and thus clean up the state for all peer
+kubernetes clusters.
+
+#### Pseudo code
+
+* CreateVolume
+  * CheckVolExists
+    * if PVCNaming:
+      * List and check if the image exists, and is still unbound to a request
```
dillaman, line 73: Is this really going to list all PVs?

docs/design/proposals/metro-dr.md
```diff
@@ -0,0 +1,216 @@
...
+external snapshotter to pass extra snapshot metadata to create the snapshot
+request.
+
+### For Encrypted PVC
+
+Currently, volumeID is used as the key to store the encryption key in KMS, the
+encryption will not work in case the workload is moved to another cluster (the
+voluemID will be completely different than the volumeID on the previous
+cluster). The common parameter across the cluster is the volume name, So
+cephcsi need to store the volume name instead of volumeID in KMS.
```
dillaman, line 216: ... or store the `key` (to the encryption key) in the image's metadata.

Madhu-1: sounds good for rbd. can this be done for cephfs also @ShyamsundarR wihtout mounting the cephfs subvolumes?

docs/design/proposals/metro-dr.md
Madhu-1: Yes it has to list all the rbd images and check the image belongs to different PV

Madhu-1: @dillaman reason for listing the images/subvolumes is that the pvc can be retained so a user can delete the pvc and the PV and rbd image will be left, if the user creates a new pvc we should not give back the same rbd image, we need to create a new image and return it. 

The reason for not naming the rbd image with the only pvc_name+namespace_name is that retained pvc can also be done by the user just by editing the pvc spec and change the Delete policy. The admin or an operator has to rename the rbd/cephfs subvolume but the admin/operator cannot update the PV spec(immutable object) to point to the newly renamed rbd/cephfs volume and currently, we don't have any mechanism to rename the cephfs subvolume.

dillaman: We should really try to avoid such operations that are not constant time. You can achieve the same effect by naming the image under some constant naming scheme and then adding metadata to the image to tie in its unique ID.

docs/design/proposals/metro-dr.md
dillaman: I'd imagine you can do something similar using a flat file co-located with the subvolumes.

docs/design/proposals/metro-dr.md
dillaman: ... this is similar to what we discussed when handling async DR cases.

Madhu-1:
Closing this one, as we don't need it in cephcsi for now.

Issue #1500, opened 2020-09-21
Enable token based authentication in vault 


humblec:
# Describe the feature you'd like to have #

At the moment vault token authentication is not enabled in CSI code and it is good to enable this method too so that we have more coverage on different type of vault authentication and can make use of it based on the user input. This issue tracks the same.

humblec:
Solution: The vault token auth can use token based auth in a
    setup. If we want to use vault token based authentication in
    a setup, the ceph-csi-encryption-kms-config should have below
    values set:
    
    `
           "vaultTokenAuth": "true",
           "vaultTokenFromSecret": "vault-token",
    `
    
    Here the first param `vaultTokenAuth` say whether token based
    auth is enabled or not. If set, the vault auth token will be fetched
    from node stage/provisioner secret of the key `vault-token`.
    ie the secret key is the value/variable set for `vaultTokenFromSecret`
    parameter in the configmap.


Reference # https://github.com/ceph/ceph-csi/pull/1559#issuecomment-706915448

PR #1559, opened 2020-10-06
rbd: add vault token authentication support

humblec:
Fix: https://github.com/ceph/ceph-csi/issues/1500
    Solution: The vault token auth can use token based auth in a
    setup. If we want to use vault token based authentication in
    a setup, the ceph-csi-encryption-kms-config should have below
    values set:
    
    `
           "vaultTokenAuth": "true",
           "vaultTokenFromSecret": "vault-token",
    `
    
    Here the first param `vaultTokenAuth` say whether token based
    auth is enabled or not. If set, the vault auth token will be fetched
    from node stage/provisioner secret of the key `vault-token`.
    ie the secret key is the value/variable set for `vaultTokenFromSecret`
    parameter in the configmap.

*NOTE for the reviewer:*

The E2E is validated for vault token authentication with this PR. 

### DIFF
--- a/docs/design/proposals/encrypted-pvc.md
+++ b/docs/design/proposals/encrypted-pvc.md
@@ -127,3 +127,10 @@ data:
 metadata:
   name: ceph-csi-encryption-kms-config
 ```
+
+NOTE:
+
+KMS vault can be authenticated using token auth method or kubernetes auth
+method. Ceph CSI driver is capable of authenticating with both methods,
+however kubernetes auth based authentication method is recommended for
+production use.
--- a/e2e/rbd_helper.go
+++ b/e2e/rbd_helper.go
@@ -139,6 +139,10 @@ func createRBDSecret(c kubernetes.Interface, f *framework.Framework) error {
 	}
 
 	err = updateSecretForEncryption(c)
+	if err != nil {
+		return err
+	}
+	err = updateVaultTokenForEncryption(c)
 	return err
 }
 
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -37,6 +37,10 @@ const (
 
 	rookTolBoxPodLabel = "app=rook-ceph-tools"
 	rbdmountOptions    = "mountOptions"
+
+	// vault specific constants
+	defaultVaultToken    = "sample_root_token_id"
+	defaultVaultTokenKey = "vault-token"
 )
 
 var (
@@ -125,6 +129,32 @@ func updateSecretForEncryption(c kubernetes.Interface) error {
 		return err
 	}
 	return nil
+}
+
+// updateVaultTokenForEncryption is an hack to update the secrets created by rook to
+// include the vault token key.
+func updateVaultTokenForEncryption(c kubernetes.Interface) error {
+	secrets, err := c.CoreV1().Secrets(rookNamespace).Get(context.TODO(), rbdProvisionerSecretName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	secrets.Data[defaultVaultTokenKey] = []byte(defaultVaultToken)
+
+	_, err = c.CoreV1().Secrets(rookNamespace).Update(context.TODO(), secrets, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+
+	secrets, err = c.CoreV1().Secrets(rookNamespace).Get(context.TODO(), rbdNodePluginSecretName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	secrets.Data[defaultVaultTokenKey] = []byte(defaultVaultToken)
+
+	_, err = c.CoreV1().Secrets(rookNamespace).Update(context.TODO(), secrets, metav1.UpdateOptions{})
+	return err
 }
 
 func deleteResource(scPath string) error {
--- a/examples/kms/vault/kms-config.yaml
+++ b/examples/kms/vault/kms-config.yaml
@@ -9,6 +9,8 @@ data:
         "vaultAddress": "http://vault.default.svc.cluster.local:8200",
         "vaultAuthPath": "/v1/auth/kubernetes/login",
         "vaultRole": "csi-kubernetes",
+        "vaultTokenAuth": "true",
+        "vaultTokenFromSecret": "vault-token",
         "vaultPassphraseRoot": "/v1/secret",
         "vaultPassphrasePath": "ceph-csi/",
         "vaultCAVerify": "false"
--- a/examples/rbd/secret.yaml
+++ b/examples/rbd/secret.yaml
@@ -13,3 +13,6 @@ stringData:
 
   # Encryption passphrase
   encryptionPassphrase: test_passphrase
+
+  # Vault token
+  vault-token: sample_root_token_id
--- a/internal/util/vault.go
+++ b/internal/util/vault.go
@@ -58,6 +58,8 @@ Example JSON structure in the KMS config is,
 		"vaultNamespace": "",
 		"vaultPassphraseRoot": "/v1/secret",
 		"vaultPassphrasePath": "",
+                "vaultTokenAuth": true,
+		"vaultTokenFromSecret": "vault-token",
 		"vaultCAVerify": true,
 		"vaultCAFromSecret": "vault-ca"
 	},
@@ -72,6 +74,8 @@ type VaultKMS struct {
 	VaultNamespace      string
 	VaultPassphraseRoot string
 	VaultPassphrasePath string
+	VaultToken          string
+	VaultTokenAuth      bool
 	VaultCAVerify       bool
 	vaultCA             *x509.CertPool
 }
@@ -89,6 +93,14 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	if !ok || kms.VaultAddress == "" {
 		return nil, fmt.Errorf("missing 'vaultAddress' for vault KMS %s", kmsID)
 	}
+	kms.VaultTokenAuth = false
+	err = parseVaultToken(kms, config, secrets)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse 'vaultTokenAuth' for vault <%s> kms config: %s",
+			kmsID, err)
+	}
+
+	// we are falling back to kubernetes auth method
 	kms.VaultAuthPath, ok = config["vaultAuthPath"]
 	if !ok || kms.VaultAuthPath == "" {
 		kms.VaultAuthPath = vaultDefaultAuthPath
@@ -97,6 +109,7 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	if !ok || kms.VaultRole == "" {
 		kms.VaultRole = vaultDefaultRole
 	}
+
 	kms.VaultNamespace, ok = config["vaultNamespace"]
 	if !ok || kms.VaultNamespace == "" {
 		kms.VaultNamespace = vaultDefaultNamespace
@@ -134,6 +147,35 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	}
 
 	return kms, nil
+}
+
+func parseVaultToken(kms *VaultKMS, config, secrets map[string]string) error {
+	var ok bool
+	var err error
+	var token, authToken, vaultTokenFromSecret string
+	authToken, ok = config["vaultTokenAuth"]
+	if ok {
+		kms.VaultTokenAuth, err = strconv.ParseBool(authToken)
+		if err != nil {
+			return fmt.Errorf("failed to parse 'vaultTokenAuth' for vault  kms config: %v", err)
+		}
+	}
+	// Check if vaultTokenAuth is enabled and if yes, use token based auth
+	if kms.VaultTokenAuth {
+		// Parse VaultToken from secret
+		vaultTokenFromSecret, ok = config["vaultTokenFromSecret"]
+		if ok && vaultTokenFromSecret != "" {
+			token, ok = secrets[vaultTokenFromSecret]
+			if !ok {
+				return fmt.Errorf("missing vault Token in secret %s", vaultTokenFromSecret)
+			}
+			if token == "" {
+				return fmt.Errorf("vault Token in secret %s is nil", vaultTokenFromSecret)
+			}
+			kms.VaultToken = token
+		}
+	}
+	return nil
 }
 
 // GetID is returning correlation ID to KMS configuration.
@@ -213,7 +255,6 @@ func (kms *VaultKMS) DeletePassphrase(key string) error {
 		return fmt.Errorf("could not retrieve vault token to delete the passphrase at %s: %s",
 			key, err)
 	}
-
 	resp, err := kms.send("DELETE", kms.getKeyMetadataURI(key), &vaultToken, nil)
 	if err != nil {
 		return fmt.Errorf("delete passphrase at %s request to vault failed: %s", key, err)
@@ -244,6 +285,9 @@ Vault will verify service account jwt token with Kubernetes and return token
 if the requester is allowed.
 */
 func (kms *VaultKMS) getAccessToken() (string, error) {
+	if kms.VaultToken != "" {
+		return kms.VaultToken, nil
+	}
 	saToken, err := ioutil.ReadFile(serviceAccountTokenPath)
 	if err != nil {
 		return "", fmt.Errorf("service account token could not be read: %s", err)


humblec:
@vasyl-purchel ptal .

internal/util/vault.go
```diff
@@ -208,12 +214,16 @@ func (kms *VaultKMS) SavePassphrase(key, value string) error {
 
 // DeletePassphrase deletes passphrase from Vault.
 func (kms *VaultKMS) DeletePassphrase(key string) error {
-	vaultToken, err := kms.getAccessToken()
-	if err != nil {
-		return fmt.Errorf("could not retrieve vault token to delete the passphrase at %s: %s",
-			key, err)
+	vaultToken := ""
```
vasyl-purchel, line 217: I think it would be cleaner to add this inside getAccessToken() method - if vaultToken auth is in use - just return it, if not - request the token by using k8s auth

humblec: Actually thats what I did at first place. then the check looks something like this: ( which somehow I didnt like :)) 

```
 func (kms *VaultKMS) getAccessToken() (string, error) {
+       if kms.VaultToken != ""{
+               return kms.VaultToken, nil
+       }

```
but I can revert to that..

vasyl-purchel: it looks a lot like memoization solution often used, so I don't see why this wouldn't be the good thing to do :)

humblec: Yeah.. different flavour of choice :).. reverting to first version.. thanks!

humblec: @vasyl-purchel reverted to first model .. ptal.

vasyl-purchel (approved):

humblec:
@Madhu-1 @ShyamsundarR ptal.

examples/kms/vault/kms-config.yaml
```diff
@@ -9,6 +9,7 @@ data:
         "vaultAddress": "http://vault.default.svc.cluster.local:8200",
         "vaultAuthPath": "/v1/auth/kubernetes/login",
         "vaultRole": "csi-kubernetes",
+        "vaultToken": "sample_root_token_id",
```
Madhu-1, line 12: please add E2E testing for this one.

Madhu-1 (commented):
How secure it is to save token in a configmap?

examples/kms/vault/kms-config.yaml
Madhu-1: Add this new configuration to value.yaml and kms-config.yaml

dillaman:
What's the TTL expectation on Vault tokens? I would think the real useful Vault auth method would be its k8s service account integration (i.e. Vault validates that ceph-csi itself is allowed to talk to it).

Madhu-1:
> What's the TTL expectation on Vault tokens?

@dillaman I assume there is no TTL possible with this approach as the token is not getting changed. but what would be the security on the same (if the token is not rotated or regenerated)?

> I would think the real useful Vault auth method would be its k8s service account integration (i.e. Vault validates that ceph-csi itself is allowed to talk to it).

 will it be possible if cephcsi  is talking to an external vault deployed somewhere  outside of kubernetes

vasyl-purchel:
It is possible to have a token that will have a TTL but then you would need to update it in the k8s secrets also. There is also a renewable tokens - so you would have a cron job or smth to renew that token.
it is not very secure though so I would not recommend to use token based auth in production, but for development clusters or tests where you don't want to deal with extra complexity added by k8s auth - it's a good option to have.

humblec:
> It is possible to have a token that will have a TTL but then you would need to update it in the k8s secrets also. There is also a renewable tokens - so you would have a cron job or smth to renew that token.
> it is not very secure though so I would not recommend to use token based auth in production, but for development clusters or tests where you don't want to deal with extra complexity added by k8s auth - it's a good option to have.

@dillaman  we already support kube auth with ceph csi vault driver.  As @vasyl-purchel mentioned, this is an `addon` if someone want to use the token based auth.

examples/kms/vault/kms-config.yaml
humblec: @Madhu-1 this is `kms-config.yaml` , I failed to get which files you are referring ?

dillaman:
> @dillaman we already support kube auth with ceph csi vault driver. As @vasyl-purchel mentioned, this is an `addon` if someone want to use the token based auth.

Good to know -- I thought we only supported the standard username/password login approach. If you add this fixed token approach, I would perhaps document that it's not recommended and that you will realistically need to either keep updating your token or disable the TTL.

examples/kms/vault/kms-config.yaml
Madhu-1: ah correct, still one place its missing https://github.com/ceph/ceph-csi/blob/master/internal/util/vault.go#L57

Madhu-1:
> > @dillaman we already support kube auth with ceph csi vault driver. As @vasyl-purchel mentioned, this is an `addon` if someone want to use the token based auth.
> 
> Good to know -- I thought we only supported the standard username/password login approach. 

No, we never supported the username/password login approach,  we supported only k8s based service account auth.

>If you add this fixed token approach, I would perhaps document that it's not recommended and that you will realistically need to either keep updating your token or disable the TTL.

Yes, it's a good idea to document this one.

examples/kms/vault/kms-config.yaml
humblec: well, its source code comment. but I can add it there too..

humblec: @Madhu-1 regarding e2e, I dont think it make sense to add some tests for wrong token or something like that. Its about vault configuration and we dont want to validate that.  btw, our current E2E example use this anyway.

humblec: @Madhu-1 ` current E2E doesn't use vault user-provided token-based authentication, without E2E (how can we catch if we introduce a regression in this area?)` -> It does use vault token based authentication in this PR test as I mentioned in my previous comment! . That said, with the addition of vault token, the current E2E is using token based authentication. 

However while sending the PR, I thought of adding one more deployment model based on `disabling vault token` in the e2e. But, the deployment happens in a common path ( beforeEach) for our E2E, so ignored it. As  a proof that it has not regressed, I can remove the example config map entry and keep the code in place so that it validates regression too. Is that fine ?

humblec:
@Madhu-1 added the source code comment and also the documentation. PTAL.

Madhu-1 (commented):
LGTM, waiting for CI to pass

Madhu-1 (approved):

humblec:
Thanks @Madhu-1 and @vasyl-purchel .. I am adding `ready-to-merge` based on these 2 approvals..

humblec:
> Thanks @Madhu-1 and @vasyl-purchel .. I am adding `ready-to-merge` based on these 2 approvals..

I could wait some more before I add the label, in case if some one want to chime in.

dillaman:
Why do you need the KMS to resize? The controller can continue to just resize the RBD image and the node server can resize luks since it's already mapped and the LUKS device is open, right?

dillaman:
Doc changes look good to me.

internal/util/vault.go
```diff
@@ -89,13 +94,18 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
...
 	}
-	kms.VaultAuthPath, ok = config["vaultAuthPath"]
-	if !ok || kms.VaultAuthPath == "" {
-		kms.VaultAuthPath = vaultDefaultAuthPath
-	}
-	kms.VaultRole, ok = config["vaultRole"]
-	if !ok || kms.VaultRole == "" {
-		kms.VaultRole = vaultDefaultRole
+
+	kms.VaultToken, ok = config["vaultToken"]
```
ShyamsundarR, line 98: As both auth methods can be mentioned in the configuration, the preference should be to the stronger method first. IOW, if `vaultAuthPath` and `vaultRole` is defined in addition to `vaultToken` we should not use the token by default.

internal/util/vault.go
```diff
@@ -58,6 +58,10 @@ Example JSON structure in the KMS config is,
 		"vaultNamespace": "",
 		"vaultPassphraseRoot": "/v1/secret",
 		"vaultPassphrasePath": "",
+		// `vaultToken` is not very secure or not recommended to use in production,
```
ShyamsundarR, line 61: I assume we are not adding this to `docs/deploy-rbd.md` as this is not a user facing configuration that we want to provide/encourage, right?

examples/kms/vault/kms-config.yaml
ShyamsundarR, line 12: The change here as an example confuses me (at least), as it does not remove the auth path and role, and adds a different auth mechanism (token) into the mix. If we decide to fix order of authentication methods as per the other comment in `InitVaultKMS` then to test tokens or authPath we would need another example kms-config YAML that holds these mutually exclusive.

Also, as it stands, if post this PR we are only testing token based authentication (which is what it looks like on a brief inspection of the code), then from an E2E standpoint we are not testing the user facing code in  `getAccessToken`. I would prefer if we tested user facing options in E2E rather than developer options, if the choice is one or the other.

internal/util/vault.go
humblec: Yes. we are not mentioning it in the deploy-rbd.md .

internal/util/vault.go
humblec: We could do that as well.. Will revert the flow in next patchset.

Madhu-1 (changes requested):
Holding this PR to get clarity on the below question

@humblec the new token-based auth is for only for testing purposes or someone can use it in production also?

humblec:
> Holding this PR to get clarity on the below question
> 
> @humblec the new token-based auth is for only for testing purposes or someone can use it in production also?

@Madhu-1 not recommended for production use, but someone want to use it, they could.

vasyl-purchel:
had a bit of thinking... 🤔 
if someone will want to use it not for testing then token should be stored in secrets and not really in the config map in plain text... so I would recommend to put in docs that token can be used for initial testing only and highly not recommended for production with current implementation.

Madhu-1:
> > Holding this PR to get clarity on the below question
> > @humblec the new token-based auth is for only for testing purposes or someone can use it in production also?
> 
> @Madhu-1 not recommended for production use, but someone want to use it, they could.

if someone is expected to use it in production, IMO we should make sure the changes are done by this PR works (and in the future, no regression should be introduced in this area)
* Add E2E to validate the new configuration
* Move KMS configuration from configmap to secret

humblec:
> > > Holding this PR to get clarity on the below question
> > > @humblec the new token-based auth is for only for testing purposes or someone can use it in production also?
> > 
> > 
> > @Madhu-1 not recommended for production use, but someone want to use it, they could.
> 
> if someone is expected to use it in production, IMO we should make sure the changes are done by this PR works (and in the future, no regression should be introduced in this area)

Well, use it in production or not is upto the user. Thats why we recommend or document it.
 
> 
>     * Add E2E to validate the new configuration
> 
>     * Move KMS configuration from configmap to secret

First of all I dont see a reason to move entire KMS configuration to a secret at all. Its `configuration` and configmap is meant for that. Also in other way its a regression or backward compatibility break. 
Also see below.

>if someone will want to use it not for testing then token should be stored in secrets and not really in the config map in plain text... so I would recommend to put in docs that token can be used for initial testing only and highly not recommended for production with current implementation.

@vasyl-purchel the idea was that, admin  configure or provide it in the `secret` and a highlevel controller like `Rook` can convert it to `configmap`` for CSI pod to consume. If we directly deal with `secret` from CSI level, we are `very limited` as we dont talk to the API server directly, the only option would be embedding `vaultToken` key and value in `node-stage-secret` and `provisioner-secret`.  Is it matching with your thought ?

Madhu-1:
> > > > Holding this PR to get clarity on the below question
> > > > @humblec the new token-based auth is for only for testing purposes or someone can use it in production also?
> > > 
> > > 
> > > @Madhu-1 not recommended for production use, but someone want to use it, they could.
> > 
> > 
> > if someone is expected to use it in production, IMO we should make sure the changes are done by this PR works (and in the future, no regression should be introduced in this area)
> 
> Well, use it in production or not is upto the user. Thats why we recommend or document it.
> 
> > ```
> > * Add E2E to validate the new configuration
> > 
> > * Move KMS configuration from configmap to secret
> > ```
> 
> First of all I dont see a reason to move entire KMS configuration to a secret at all. Its `configuration` and configmap is meant for that. 

why do users want to store the token in configmap (isn't it a security problem?), kubernetes secrets should be used for storing passwords, token, ssh keys, etc am I correct? 

>Also in other way its a regression or backward compatibility break.

How this can be a regression, the configmap is mounted as a file to csi pods. instead of configmap, we need to create a secret and mount it to the pod. for csi its just a JSON file (am i missing something?)

> Also see below.
> 
> > if someone will want to use it not for testing then token should be stored in secrets and not really in the config map in plain text... so I would recommend to put in docs that token can be used for initial testing only and highly not recommended for production with current implementation.
> 
> @vasyl-purchel the idea was that, admin configure or provide it in the `secret` and a highlevel controller like `Rook` can convert it to `configmap` for CSI pod to consume. If we directly deal with `secret` from CSI level, we are `very limited` as we dont talk to the API server directly, the only option would be embedding `vaultToken` key and value in `node-stage-secret` and `provisioner-secret`. Is it matching with your thought ?

humblec:
> > > > > Holding this PR to get clarity on the below question
> > > > > @humblec the new token-based auth is for only for testing purposes or someone can use it in production also?
> > > > 
> > > > 
> > > > @Madhu-1 not recommended for production use, but someone want to use it, they could.
> > > 
> > > 
> > > if someone is expected to use it in production, IMO we should make sure the changes are done by this PR works (and in the future, no regression should be introduced in this area)
> > 
> > 
> > Well, use it in production or not is upto the user. Thats why we recommend or document it.
> > > ```
> > > * Add E2E to validate the new configuration
> > > 
> > > * Move KMS configuration from configmap to secret
> > > ```
> > 
> > 
> > First of all I dont see a reason to move entire KMS configuration to a secret at all. Its `configuration` and configmap is meant for that.
> 
> why do users want to store the token in configmap (isn't it a security problem?), kubernetes secrets should be used for storing passwords, token, ssh keys, etc am I correct?
> 

@Madhu-1 above you were mentioning about `KMS configuration` ( vault address..etc) and not about `vault token` . Especially for `vault token` I was pointing to below communication. is it clear ?

```
@vasyl-purchel the idea was that, admin configure or provide it in the secret and a highlevel controller like Rook can convert it to configmap for CSI pod to consume. If we directly deal with secret from CSI level, we are very limited as we dont talk to the API server directly, the only option would be embedding vaultToken key and value in node-stage-secret and provisioner-secret. Is it matching with your thought ?
```

Madhu-1:
> > > > > > Holding this PR to get clarity on the below question
> > > > > > @humblec the new token-based auth is for only for testing purposes or someone can use it in production also?
> > > > > 
> > > > > 
> > > > > @Madhu-1 not recommended for production use, but someone want to use it, they could.
> > > > 
> > > > 
> > > > if someone is expected to use it in production, IMO we should make sure the changes are done by this PR works (and in the future, no regression should be introduced in this area)
> > > 
> > > 
> > > Well, use it in production or not is upto the user. Thats why we recommend or document it.
> > > > ```
> > > > * Add E2E to validate the new configuration
> > > > 
> > > > * Move KMS configuration from configmap to secret
> > > > ```
> > > 
> > > 
> > > First of all I dont see a reason to move entire KMS configuration to a secret at all. Its `configuration` and configmap is meant for that.
> > 
> > 
> > why do users want to store the token in configmap (isn't it a security problem?), kubernetes secrets should be used for storing passwords, token, ssh keys, etc am I correct?
> 
> @Madhu-1 above you were mentioning about `KMS configuration` ( vault address..etc) and not about `vault token` . Especially for `vault token` I was pointing to below communication. is it clear ?
> 
> ```
> @vasyl-purchel the idea was that, admin configure or provide it in the secret and a highlevel controller like Rook can convert it to configmap for CSI pod to consume. If we directly deal with secret from CSI level, we are very limited as we dont talk to the API server directly, the only option would be embedding vaultToken key and value in node-stage-secret and provisioner-secret. Is it matching with your thought ?
> ```

Why high-level operators convert the secret and store the token in configmap (for me this seems to be a security issue)? what am requesting is instead of using configmap why can't we move from configmap to secret for storing all vault details?

humblec:
> > @vasyl-purchel the idea was that, admin configure or provide it in the secret and a highlevel controller like Rook can convert it to configmap for CSI pod to consume. If we directly deal with secret from CSI level, we are very limited as we dont talk to the API server directly, the only option would be embedding vaultToken key and value in node-stage-secret and provisioner-secret. Is it matching with your thought ?
> > ```
> 
> Why high-level operators convert the secret and store the token in configmap (for me this seems to be a security issue)? what am requesting is instead of using configmap why can't we move from configmap to secret for storing all vault details?

Again, we went with entire vault configuration ( except vault token ( introduced here)) with `Configmap` in first place. Not sure why we didnt have thought about wrapping it in secret at this time?additionally, current deployments with KMS configuration make use of config map enabled templates for KMS. Thats what I was pointing as backward compatibility .etc.  More or less, the configmap is a good fit for `Configuration` values, so vault configuration can be fit into that.. in this case vault address, role..etc and we also use `configmap` for `monior details..` .  

Here we are talking about `vault token` ( considering it can be treat as a sensitive data) . Thats why we are exploring other options as @vasyl-purchel mentioned  to embed it in the secret if someone want to use it in production..etc.

vasyl-purchel:
Yep, I agree that moving whole vault configuration into kubernetes secrets doesn't make much sense - it's not a sensitive data it's a configuration data... for the token - I didn't know rook can be used to create the config map, though my guess is even in that case the sensitive data (vault token) will be available to see in the configmap object... which is less secure then if we would have it in secrets like a passphrase (for non-vault variant)...

though since I still see this as a developer/test only option, it's ok to have it as it is now, just make sure that in documentation it is clearly bad idea to use it in production

humblec:
> Yep, I agree that moving whole vault configuration into kubernetes secrets doesn't make much sense - it's not a sensitive data it's a configuration data... for the token - I didn't know rook can be used to create the config map, though my guess is even in that case the sensitive data (vault token) will be available to see in the configmap object... which is less secure then if we would have it in secrets like a passphrase (for non-vault variant)...

true.. more or less, if someone got access to the configmap in the namespace we operate or to the CSI pods, any way, its compromised to an extent. :+1: 
> 
> though since I still see this as a developer/test only option, it's ok to have it as it is now, just make sure that in documentation it is clearly bad idea to use it in production

Completely agree! Thanks for sharing the thoughts !

ShyamsundarR:
> > Yep, I agree that moving whole vault configuration into kubernetes secrets doesn't make much sense - it's not a sensitive data it's a configuration data... for the token - I didn't know rook can be used to create the config map, though my guess is even in that case the sensitive data (vault token) will be available to see in the configmap object... which is less secure then if we would have it in secrets like a passphrase (for non-vault variant)...
> 
> true.. more or less, if someone got access to the configmap in the namespace we operate or to the CSI pods, any way, its compromised to an extent. +1

The token that is being added by this PR is sensitive information. It should be passed in via **a** secret. For example the current `vaultCAFromSecret` is fetched from the secret.

The config map has configuration information as @vasyl-purchel states, so these are not sensitive in nature.

If we are going to add token support without putting this into a secret, this is not usable.

> 
> > though since I still see this as a developer/test only option, it's ok to have it as it is now, just make sure that in documentation it is clearly bad idea to use it in production
> 
> Completely agree! Thanks for sharing the thoughts !

From a developer/test **only** option we still want e2e to use the supported path, as we are better off testing that in e2e. Which means we would not use the token option in e2e as well.

Which begs the question, why do we need to add token support that is insecure and not usable by users or in e2e?

nixpanic (commented):
Because this is a feature added for testing purposes, it would be good to have a plan to make it ready for production use. Can you add such a plan to docs/design/proposals and keep it updated when modifications to the code are planned/made?

humblec:
> > > Yep, I agree that moving whole vault configuration into kubernetes secrets doesn't make much sense - it's not a sensitive data it's a configuration data... for the token - I didn't know rook can be used to create the config map, though my guess is even in that case the sensitive data (vault token) will be available to see in the configmap object... which is less secure then if we would have it in secrets like a passphrase (for non-vault variant)...
> > 
> > 
> > true.. more or less, if someone got access to the configmap in the namespace we operate or to the CSI pods, any way, its compromised to an extent. +1
> 
> The token that is being added by this PR is sensitive information. It should be passed in via **a** secret. For example the current `vaultCAFromSecret` is fetched from the secret.
> 
> The config map has configuration information as @vasyl-purchel states, so these are not sensitive in nature.
> 
> If we are going to add token support without putting this into a secret, this is not usable.
> 

We could add vault token support in secrets too as we have done for vaultCAFromSecret :

For ex: The configmap will have a below entry

```
"vaultAddress: <blah>
"vaultTokenFromSecret": <vault-token-secret>
```
`vault-token-secret` will be fetched from `secrets` we are provided with and will be used for further operations.

@ShyamsundarR @vasyl-purchel are we in agreement ?

humblec:
Looks like we are in agreement here https://github.com/ceph/ceph-csi/pull/1559#issuecomment-706915448 . Will change this patchset to read the secret key. Thanks !

humblec:
@vasyl-purchel @ShyamsundarR @Madhu-1  @nixpanic please review

internal/util/vault.go
```diff
@@ -58,6 +58,8 @@ Example JSON structure in the KMS config is,
 		"vaultNamespace": "",
 		"vaultPassphraseRoot": "/v1/secret",
 		"vaultPassphrasePath": "",
+                "vaultTokenAuth": true,
```
Madhu-1, line 61: the indentation is not proper here.

internal/util/vault.go
```diff
@@ -89,14 +96,38 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
...
+		}
 	}
-	kms.VaultRole, ok = config["vaultRole"]
-	if !ok || kms.VaultRole == "" {
-		kms.VaultRole = vaultDefaultRole
+	// Check if vaultTokenAuth is enabled and if yes, use token based auth
+	if kms.VaultTokenAuth {
+		// Parse VaultToken from secret
+		vaultTokenFromSecret, ok = config["vaultTokenFromSecret"]
+		if ok && vaultTokenFromSecret != "" {
```
Madhu-1, line 112: what if the token is empty in secret?

e2e/utils.go
```diff
@@ -127,6 +127,35 @@ func updateSecretForEncryption(c kubernetes.Interface) error {
...
 
+// updateVaultTokenForEncryption is an hack to update the secrets created by rook to
+// include the vault token key.
+func updateVaultTokenForEncryption(c kubernetes.Interface) error {
+	secrets, err := c.CoreV1().Secrets(rookNamespace).Get(context.TODO(), rbdProvisionerSecretName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	secrets.Data["vault-token"] = []byte("sample_root_token_id")
```
Madhu-1, line 138: define a const for `sample_root_token_id` and `vault-token`

examples/kms/vault/kms-config.yaml
```diff
@@ -9,6 +9,8 @@ data:
         "vaultAddress": "http://vault.default.svc.cluster.local:8200",
         "vaultAuthPath": "/v1/auth/kubernetes/login",
         "vaultRole": "csi-kubernetes",
+        "vaultTokenAuth": "true",
```
Madhu-1, line 12: we need to validate both token-based and kubernetes secret based encryption. with this change, we are only validating the token-based approach, not the kubernetes secret based.

e2e/utils.go
```diff
@@ -127,6 +127,35 @@ func updateSecretForEncryption(c kubernetes.Interface) error {
 	return nil
 }
 
+// updateVaultTokenForEncryption is an hack to update the secrets created by rook to
+// include the vault token key.
+func updateVaultTokenForEncryption(c kubernetes.Interface) error {
```
Madhu-1, line 132: am not sure how this feature will be integrated with the rook. Rook creates only two secrets (one for provisioner and one for node staging)  for all the users these secrets will be sent to the CSI driver even if the storageclass is not expected to be encrypted. won't there be any security problem with this?

@leseb any idea what's the plan around rook for this one?

e2e/utils.go
```diff
@@ -127,6 +127,35 @@ func updateSecretForEncryption(c kubernetes.Interface) error {
...
+
+	secrets, err = c.CoreV1().Secrets(rookNamespace).Get(context.TODO(), rbdNodePluginSecretName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	secrets.Data["vault-token"] = []byte("sample_root_token_id")
+
+	_, err = c.CoreV1().Secrets(rookNamespace).Update(context.TODO(), secrets, metav1.UpdateOptions{})
+	if err != nil {
```
Madhu-1, line 153: return err we don't need to check it here

Madhu-1 (changes requested):

e2e/utils.go
humblec: >am not sure how this feature will be integrated with the rook.
First of all, rook integration is independent to this repo. :) . CSI driver got its own existence with its deployment model. Rook is just one caller , there could be others too. so the secret creation from rook or its entries are not really a concern here I believe.

>Rook creates only two secrets (one for provisioner and one for node staging) for all the users these secrets will be sent to the CSI driver

Coming to the part of `secrets` will be sent to CSI driver, yes, and thats the general working mechanism for the CSI calls ( ie CO sending secret to CSI driver in many RPCs), also as long as the SC belong to the a specific driver/provisioner, thats by design and that CSI/SP driver is in charge of operating with it. Isnt it ? Also, the secret is not bound for "all" SCs and also its `not` just one. That said, different SCs can have different secrets , that well acceptable! and I would like to point that, this is `not` a new model, even we have `secret` based encryption and we pass `  # Encryption passphrase` in the same way in our implementation today!

e2e/utils.go
humblec: sure

internal/util/vault.go
```diff
@@ -89,14 +96,38 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	if !ok || kms.VaultAddress == "" {
 		return nil, fmt.Errorf("missing 'vaultAddress' for vault KMS %s", kmsID)
 	}
-	kms.VaultAuthPath, ok = config["vaultAuthPath"]
-	if !ok || kms.VaultAuthPath == "" {
-		kms.VaultAuthPath = vaultDefaultAuthPath
+	kms.VaultTokenAuth = false
+	authToken, ok := config["vaultTokenAuth"]
```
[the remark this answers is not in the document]
humblec: that could be done :+1:

e2e/utils.go
humblec: sure.

internal/util/vault.go
humblec: ah.. will correct it.

e2e/utils.go
Madhu-1: @humblec I know all those things, what am trying to understand is the security issue we may face and how this is planned to be integrated with rook

internal/util/vault.go
humblec: sure, we could return `err` in that case :+1:

internal/util/vault.go
```diff
@@ -72,15 +74,20 @@ type VaultKMS struct {
...
 	VaultPassphraseRoot string
 	VaultPassphrasePath string
+	VaultToken          string
+	VaultTokenAuth      bool
 	VaultCAVerify       bool
 	vaultCA             *x509.CertPool
 }
 
 // InitVaultKMS returns an interface to HashiCorp Vault KMS.
+// nolint:gocognit:gocyclo // TODO: reduce complexity
```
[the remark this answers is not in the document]
humblec: hmm.. thought of entire parsing in one place , but it can be addressed too :) .

examples/kms/vault/kms-config.yaml
humblec: This is bit tricky and inviting thoughts. At present, in our general framework umbrella we deploy vault: 

```
var _ = Describe("RBD", func() {
	f := framework.NewDefaultFramework("rbd")
	var c clientset.Interface
	// deploy RBD CSI
	BeforeEach(func() {
		if !testRBD || upgradeTesting {
			Skip("Skipping RBD E2E")
		}
		c = f.ClientSet
		if deployRBD {
			err := createNodeLabel(f, nodeRegionLabel, regionValue)
			if err != nil {
				e2elog.Failf("failed to create node label with error %v", err)
			}
			err = createNodeLabel(f, nodeZoneLabel, zoneValue)
			if err != nil {
				e2elog.Failf("failed to create node label with error %v", err)
			}
			if cephCSINamespace != defaultNs {
				err = createNamespace(c, cephCSINamespace)
				if err != nil {
					e2elog.Failf("failed to create namespace with error %v", err)
				}
			}
			deployRBDPlugin()
		}
		err := createConfigMap(rbdDirPath, f.ClientSet, f)
		if err != nil {
			e2elog.Failf("failed to create configmap with error %v", err)
		}
		err = createRBDStorageClass(f.ClientSet, f, nil, nil)
		if err != nil {
			e2elog.Failf("failed to create storageclass with error %v", err)
		}
		err = createRBDSecret(f.ClientSet, f)
		if err != nil {
			e2elog.Failf("failed to create secret with error %v", err)
		}
		deployVault(f.ClientSet, deployTimeout) ----------------> 1
	})
```

[1] This can be tweaked to test either token based or other based on the input of configmap or based on the manipulation we could do in deploy vault function. But its single run of one type of vault deployment. We could do the same or repeat it to validate both. Otherwise we may have to take out `deployVault` from general umbrella and try to accommodate both kind of deployment as a first function, then the question comes which tests have to bound to these different invocations? all or just some of them?

Any thoughts here @Madhu-1

e2e/utils.go
humblec: @Madhu-1 again my point is that, I dont see any security concern at all and rook integration  discussion is out of the scope here atleast :+1:

internal/util/vault.go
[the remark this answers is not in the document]
humblec: Done.

internal/util/vault.go
humblec: Done.

internal/util/vault.go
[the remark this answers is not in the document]
humblec: @Madhu-1 thinking some more on this, all the configmap attributes are not constants. Also , the usage is just once, so I feel we could leave it as it is .. Is it fine?

e2e/utils.go
leseb: I'm not sure to fully understand the intention behind `updateVaultTokenForEncryption()`, @Madhu-1 are you referring to the secret containing the cephx keys?

humblec:
@Madhu-1 @vasyl-purchel @ShyamsundarR can you please take a look at this ? the e2e comment is not addressed, rest are addressed.

e2e/utils.go
Madhu-1: currently, for RBD rook creates  2 secrets one is `rook-csi-rbd-provisioner` and `rook-csi-rbd-node` the  secret will container cephx keys it looks like below

```yaml
"name": "rook-csi-rbd-provisioner",
"kind": "Secret",
"data": {
 "userID": 'csi-rbd-provisioner',
"userKey": PROVISIONER_KEY
}
```
with the current PR design, the vault token will be added to these secrets it will be

```yaml
"name": "rook-csi-rbd-provisioner",
"kind": "Secret",
"data": {
 "userID": 'csi-rbd-provisioner',
"userKey": PROVISIONER_KEY,
"vaultToken": XXX-XXX-XXXX
}
```
In the downstream, these are the secrets used when a user creates the storageclass. Even if the storageclass is not expected to be encrypted these secrets (created by rook)  will be sent to the CSI driver. won't there be any security problem with this?

Madhu-1: > @Madhu-1 again my point is that, I dont see any security concern at all and rook integration discussion is out of the scope here atleast

as am not sure about the downstream design and keeping Rook out of scope am okay

examples/kms/vault/kms-config.yaml
Madhu-1: We need to keep `vaultTokenAuth` as `false` and move the e2e to a helper function and validate the kube auth first and update the configmap with `vaultTokenAuth` = `true` and call the e2e again, validate the token based.

internal/util/vault.go
Madhu-1: This is not fixed yet, please don't mark it resolved it if it's not fixed.

internal/util/vault.go
```diff
@@ -89,6 +93,14 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	if !ok || kms.VaultAddress == "" {
 		return nil, fmt.Errorf("missing 'vaultAddress' for vault KMS %s", kmsID)
 	}
+	kms.VaultTokenAuth = false
```
Madhu-1, line 96: default is false, why we are setting it back to false here?

internal/util/vault.go
```diff
@@ -89,6 +93,14 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
...
+	err = parseVaultToken(kms, config, secrets)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse 'vaultTokenAuth' for vault <%s> kms config: %s",
+			kmsID, err)
+	}
+
+	// we are falling back to kubernetes auth method
```
Madhu-1, line 103: why we need to fallback to the kubernetes auth method if the token is provided?

e2e/rbd_helper.go
```diff
@@ -139,6 +139,10 @@ func createRBDSecret(c kubernetes.Interface, f *framework.Framework) error {
 	}
 
 	err = updateSecretForEncryption(c)
+	if err != nil {
+		return err
+	}
+	err = updateVaultTokenForEncryption(c)
```
nixpanic, line 145: Because Vault Authentication support can be enabled/disabled, both ways need to be tested. With this change, we are not testing the Kubernetes Auth based authentication anymore?

docs/design/proposals/encrypted-pvc.md
```diff
@@ -127,3 +127,10 @@ data:
 metadata:
   name: ceph-csi-encryption-kms-config
 ```
+
+NOTE:
+
+KMS vault can be authenticated using token auth method or kubernetes auth
```
nixpanic, line 133: Vauls and Kubernetes are names, use **V**ault and **K**ubernetes. Maybe even **K**ubernetes **A**uth

docs/design/proposals/encrypted-pvc.md
```diff
@@ -127,3 +127,10 @@ data:
...
+method. Ceph CSI driver is capable of authenticating with both methods,
```
nixpanic, line 134: Either "Ceph CSI" or "The Ceph CSI driver"

e2e/utils.go
```diff
@@ -127,6 +131,32 @@ func updateSecretForEncryption(c kubernetes.Interface) error {
 	return nil
 }
 
+// updateVaultTokenForEncryption is an hack to update the secrets created by rook to
```
nixpanic, line 134: *... is a hack ...* indicates that this needs to be changed. Add a `FIXME` or `TODO` describing what is needed.

examples/kms/vault/kms-config.yaml
nixpanic, line 12: I agree that defaulting to `false` would be better, specially as you mentioned that this method is not suitable for production deployments. Please also add a comment in the `.yaml` expressing the recommendation to not use this functionality, except for testing.

examples/rbd/secret.yaml
```diff
@@ -13,3 +13,6 @@ stringData:
 
   # Encryption passphrase
   encryptionPassphrase: test_passphrase
+
+  # Vault token
```
nixpanic, line 17: that is a pretty useless comment, it is exactly the same ad the name of the key... Either add value with a comment, or just do not add it at all. Here, you should probably mention it is only used when `vaultTokenAuth` is enabled.

internal/util/vault.go
```diff
@@ -136,6 +149,35 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	return kms, nil
 }
 
+func parseVaultToken(kms *VaultKMS, config, secrets map[string]string) error {
+	var ok bool
```
nixpanic, line 153: use standard coding guidelines/style for multiple variables (I wonder why linters did not complain about this):
```
var (
    ok bool
    err error
    token, authToken, vaultTokenFromSecret string
)
```

internal/util/vault.go
```diff
@@ -136,6 +149,35 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
...
+	var err error
+	var token, authToken, vaultTokenFromSecret string
+	authToken, ok = config["vaultTokenAuth"]
+	if ok {
+		kms.VaultTokenAuth, err = strconv.ParseBool(authToken)
+		if err != nil {
+			return fmt.Errorf("failed to parse 'vaultTokenAuth' for vault  kms config: %v", err)
```
nixpanic, line 160: when returning errors, use `%w` in the formatting string to be able to use `errors.Is()` and similar

internal/util/vault.go
```diff
@@ -89,6 +93,14 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	if !ok || kms.VaultAddress == "" {
 		return nil, fmt.Errorf("missing 'vaultAddress' for vault KMS %s", kmsID)
 	}
+	kms.VaultTokenAuth = false
+	err = parseVaultToken(kms, config, secrets)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse 'vaultTokenAuth' for vault <%s> kms config: %s",
+			kmsID, err)
```
nixpanic, line 100: Errors should be formatted as `%w` when creating new errors

internal/util/vault.go
```diff
@@ -213,7 +255,6 @@ func (kms *VaultKMS) DeletePassphrase(key string) error {
 		return fmt.Errorf("could not retrieve vault token to delete the passphrase at %s: %s",
 			key, err)
 	}
-
 	resp, err := kms.send("DELETE", kms.getKeyMetadataURI(key), &vaultToken, nil)
```
nixpanic, line 258: no need to cleanup unrelated code. If you do this, put it in a separate commit

internal/util/vault.go
```diff
@@ -97,6 +109,7 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	if !ok || kms.VaultRole == "" {
 		kms.VaultRole = vaultDefaultRole
 	}
+
```
nixpanic, line 112: no need to change this here, place it in a cleanup commit if you think this is a useful change

internal/util/vault.go
```diff
@@ -136,6 +149,35 @@ func InitVaultKMS(kmsID string, config, secrets map[string]string) (EncryptionKM
 	return kms, nil
 }
 
+func parseVaultToken(kms *VaultKMS, config, secrets map[string]string) error {
```
nixpanic, line 152: it is better to have this be a method of VaultKMS

internal/util/vault.go
```diff
@@ -244,6 +285,9 @@ Vault will verify service account jwt token with Kubernetes and return token
 if the requester is allowed.
 */
 func (kms *VaultKMS) getAccessToken() (string, error) {
+	if kms.VaultToken != "" {
```
nixpanic, line 288: Should this function not set `kms.VaultToken` too? If it gets set elsewhere, it should probably get explained in a commit.

internal/util/vault.go
```diff
@@ -244,6 +285,9 @@ Vault will verify service account jwt token with Kubernetes and return token
...
+		return kms.VaultToken, nil
```
nixpanic, line 289: Does `kms.VaultToken` not expire?

nixpanic (changes requested):

internal/util/vault.go
leseb: It depends on how it is setup in Vault, the admin can decide to renew it or have it never expire.

nixpanic:
After a lot of considerations and working a bit with the `internal/util/vault.go` code, I have come to a more modular and maintainable approach. Instead of modifying the existing `VaultKMS` type and bolting on additional conflicting functionality, I prefer to add a new KMS type `VaultTokens`. The configuration for this type can be based on the `VaultKMS` type (share helper functions, or a base type).

More details are in [Ceph-CSI: multi-tenancy with Vault tokens](https://hackmd.io/C4ZorZinRhm9FWKW13KWrw?view) document and the [KMS: Vault Tokens](https://github.com/ceph/ceph-csi/projects/8) project.

PR #1572, opened 2020-10-13
Use a GitHub API token for CI jobs

nixpanic:
Sometimes CI jobs fail with errors like
```
[Pipeline] { (detect k8s-1.18 patch release)
[Pipeline] sh
+ ./scripts/get_patch_release.py --version=1.18
Traceback (most recent call last):
  File "./scripts/get_patch_release.py", line 76, in <module>
    main()
  File "./scripts/get_patch_release.py", line 51, in main
    releases = get_releases(json)
  File "./scripts/get_patch_release.py", line 37, in get_releases
    releases.append(release['name'])
TypeError: string indices must be integers
```
This indicates that the API call to GitHub returned something unexpected.

Because the jobs run in the CentOS CI, there can be many API calls from the environment (not only Ceph-CSI), which trigger a rate limit. When the problem happens, debugging showed
```
403 Client Error: rate limit exceeded
```

The rate limit can be prevented when a CI job logs in to use the API. Our @ceph-csi-bot has a "personal access token" that can be used for this. The token is configured in Jenkins and can be obtained in the groovy scripts (Pipelines).

### DIFF
--- a/scripts/get_github_labels.py
+++ b/scripts/get_github_labels.py
@@ -4,9 +4,14 @@ Fetches the labels of an Issue or Pull-Request from GitHub.
 
 Parameters:
  --id=<id>: the number of the the GitHub Issue or Pull-Request
- --has-label=<label>: label to check (exit 0 if set, 1 unset), without
+ --has-label=<label>: label to check (exit 0 if set, 2 unset), without
                       --has-label, all labels of the Issue or Pull-Request
                       get printed.
+
+Exit codes:
+ 0: success
+ 1: any unexpected failure
+ 2: --has-label=<label> was passed, <label> is not set on the PR
 '''
 
 import argparse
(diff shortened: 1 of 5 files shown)


scripts/get_github_labels.py
```diff
@@ -4,14 +4,24 @@
...
 Parameters:
  --id=<id>: the number of the the GitHub Issue or Pull-Request
- --has-label=<label>: label to check (exit 0 if set, 1 unset), without
+ --has-label=<label>: label to check (exit 0 if set, 2 unset), without
                       --has-label, all labels of the Issue or Pull-Request
                       get printed.
+
+Exit codes:
+ 0: success
+ 1: any unexpected failure
```
Madhu-1, line 13: do we need to handle 1 in groovy script incase of any error?

nixpanic: not really, check `!= 0` in the groovy script, which basically means: run e2e if not `0`

Madhu-1 (approved):

PR #1573, opened 2020-10-13
e2e: add snapshotter deploy functionality to minikube script

pkalever:

Just like deploy-rook and teardown-rook, this patch will add
install snapshotter and cleanup snapshotter option to minikube
script.

### DIFF
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -249,6 +249,11 @@ deploy-rook)
     DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
     "$DIR"/rook.sh deploy
     ;;
+install-snapshotter)
+    echo "install snapshot controller"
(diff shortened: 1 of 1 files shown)


Madhu-1 (approved):

nixpanic (approved):
Do you plan to use update the CI jobs that call this script too?

pkalever:
> Do you plan to use update the CI jobs that call this script too?

Yes, also the other places such as documentation.
But that will come in a separate PR.

Thanks, Neils!

pkalever:
@Madhu-1 @nixpanic I have not integrated `install-snapshot.sh delete-crd` assuming that it will be deprecated in the future, because its main purpose is to address alpha CRDs. Is this right?

Are we at a stage to discard this `delete-crd` now ? or do we think alpha CRDs are still in use?
If so do you suggest to add delete-crd to minikube too? so that we can simply replace the install-snapshot.sh usage across all the scripts/doc straight.

Thanks!

PR #1579, opened 2020-10-14
e2e: cleanup Kubernetes version checks

nixpanic:
Reduce redundant code, add a helper to check the Kubernetes version so that tests can run depending on the version that is used.

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -568,3 +568,27 @@ func validatePVCClone(sourcePvcPath, clonePvcPath, clonePvcAppPath string, f *fr
 	wg.Wait()
 	validateRBDImageCount(f, 0)
 }
+
+// k8sVersionGreaterEquals checks the ServerVersion of the Kubernetes cluster
+// and compares it to the major.minor version passed. In case the version of
+// the cluster is equal or higher to major.minor, `true` is returned, `false`
+// otherwise.
+//
+// If fetching the ServerVersion of the Kubernetes cluster fails, the calling
+// test case is marked as `FAILED` and gets aborted.
+//
+// nolint:unparam // currently major is always 1, this can change in the future
+func k8sVersionGreaterEquals(c kubernetes.Interface, major, minor int) bool {
+	v, err := c.Discovery().ServerVersion()
+	if err != nil {
+		e2elog.Failf("failed to get server version with error %v", err)
+		// Failf() marks the case as failure, and returns from the
+		// Go-routine that runs the case. This function will not have a
+		// return value.
+	}
+
+	maj := fmt.Sprintf("%d", major)
+	min := fmt.Sprintf("%d", minor)
+
+	return (v.Major > maj) || (v.Major == maj && v.Minor >= min)
+}
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -425,13 +425,8 @@ var _ = Describe("cephfs", func() {
 			})
 
 			By("Resize PVC and check application directory size", func() {
-				v, err := f.ClientSet.Discovery().ServerVersion()
-				if err != nil {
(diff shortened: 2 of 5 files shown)


e2e/utils.go
```diff
@@ -498,3 +498,17 @@ func oneReplicaDeployYaml(template string) string {
 	var re = regexp.MustCompile(`(\s+replicas:) \d+`)
 	return re.ReplaceAllString(template, `$1 1`)
 }
+
+// nolint:unparam // currently major is always 1, this can change the the future
+func k8sVersionGreaterEquals(c kubernetes.Interface, major, minor int) bool {
```
pkalever, line 503: Naming looks a bit odd to me, a few suggestions:

checkK8sVersionMatchRequired()
checkK8sVersionMeetRequired()
isK8sVersionMatchRequired()
isK8sVersionMeetRequired()

e2e/utils.go
```diff
@@ -498,3 +498,17 @@ func oneReplicaDeployYaml(template string) string {
 	var re = regexp.MustCompile(`(\s+replicas:) \d+`)
 	return re.ReplaceAllString(template, `$1 1`)
 }
+
+// nolint:unparam // currently major is always 1, this can change the the future
```
pkalever, line 502: s/the the future/in the future/ ?

pkalever (approved):
Otherwise looks good to me.

e2e/utils.go
nixpanic: I took the naming from `-ge` (Greater Equals, Bash). `Match` is not correct, as it needs `>=` (pronounced: greater equals). Not sure about `Meet`...

e2e/utils.go
nixpanic: Ah, yes, I'll fix that

e2e/utils.go
```diff
@@ -498,3 +498,17 @@ func oneReplicaDeployYaml(template string) string {
 	var re = regexp.MustCompile(`(\s+replicas:) \d+`)
 	return re.ReplaceAllString(template, `$1 1`)
 }
+
+// nolint:unparam // currently major is always 1, this can change in the future
+func k8sVersionGreaterEquals(c kubernetes.Interface, major, minor int) bool {
+	v, err := c.Discovery().ServerVersion()
+	if err != nil {
+		e2elog.Failf("failed to get server version with error with error %v", err)
```
Madhu-1, line 506: FailF will mark the test case as a failure. this need to be replaced with a log and return will not happen

nixpanic: It is actually intended to fail the case if fetching the ServerVersion errors out. Returning from this function would cause the case to be silently skipped (version not matching), which is not what we want. (OK, the return on the next line is unneeded, I can remove that and leave a comment instead.)

nixpanic:
@mergifyio rebase

e2e/utils.go
nixpanic: @Madhu-1 is the added comment sufficiently clear to explain the working of the new function?

Madhu-1 (approved):

e2e/utils.go
Madhu-1: Sounds good @nixpanic

pkalever: @nixpanic 
s/with error with error/with error/ ?

pkalever (changes requested):
there are duplicate strings in the log msg.

e2e/utils.go
nixpanic: Oh, yes! This is actually coming from the original version checking code that was moved from other files into its own function. I'll fix it.

pkalever (approved):

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

Issue #1588, opened 2020-10-15
ci: minikube can not consume kubernetes v1.19.3 causing CI jobs to fail


nixpanic:
CI jobs that run with the latest Kubernetes 1.19 patch release (v1.19.3) fail. This seems to be an issue with minikube, as it can not deploy the Kubernetes cluster. 

```
starting minikube with kubeadm bootstrapper
* minikube v1.13.0 on Centos 8.2.2004
  - MINIKUBE_VERSION=v1.13.0
! minikube skips various validations when --force is supplied; this may lead to unexpected behavior
* Using the kvm2 driver based on user configuration
* The "kvm2" driver should not be used with root privileges.
* If you are running minikube within a VM, consider using --driver=none:
*   https://minikube.sigs.k8s.io/docs/reference/drivers/none/
* Downloading driver docker-machine-driver-kvm2:
* minikube 1.14.0 is available! Download it: https://github.com/kubernetes/minikube/releases/tag/v1.14.0
* To disable this notice, run: 'minikube config set WantUpdateNotification false'

    > docker-machine-driver-kvm2.sha256: 65 B / 65 B [-------] 100.00% ? p/s 0s    > docker-machine-driver-kvm2: 3.42 MiB / 13.81 MiB [->_____] 24.73% ? p/s ?    > docker-machine-driver-kvm2: 5.77 MiB / 13.81 MiB [-->____] 41.82% ? p/s ?    > docker-machine-driver-kvm2: 7.92 MiB / 13.81 MiB [---->__] 57.32% ? p/s ?    > docker-machine-driver-kvm2: 10.43 MiB / 13.81 MiB  75.54% 11.69 MiB p/s E    > docker-machine-driver-kvm2: 12.32 MiB / 13.81 MiB  89.23% 11.69 MiB p/s E    > docker-machine-driver-kvm2: 13.79 MiB / 13.81 MiB  99.87% 11.69 MiB p/s E    > docker-machine-driver-kvm2: 13.81 MiB / 13.81 MiB  100.00% 13.78 MiB p/s * Downloading VM boot image ...
[...]

PR #1589, opened 2020-10-15
ci: use Kubernetes 1.19.2 for CI jobs

nixpanic:
The latest Kubernetes patch release (v1.19.3) can not be deployed with
minikube. Selecting version 1.19.2 until the problems with minikube are
addressed.

In case the version that is passed with --version=... contains the patch
release already, the latest patch release will not be detected, but the
passed version will be returned (enables forcing a particular version in
CI jobs).

Updates: #1588

### DIFF
--- a/jobs/mini-e2e.yaml
+++ b/jobs/mini-e2e.yaml
@@ -3,7 +3,7 @@
     name: mini-e2e
     k8s_version:
       - '1.18'
-      - '1.19'
+      - '1.19.2'
     jobs:
       - 'mini-e2e_k8s-{k8s_version}'
       - 'mini-e2e-helm_k8s-{k8s_version}'
--- a/scripts/get_patch_release.py
+++ b/scripts/get_patch_release.py
@@ -1,7 +1,10 @@
 #!/usr/bin/python
 '''
 Fetches the Kubernetes releases from GitHub and returns the most recent patch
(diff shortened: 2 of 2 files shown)


Madhu-1:
@nixpanic this is one of the problems we are going to face if we try to deploy the latest available kubernetes versions which are not supported by minikube. how about sticking to a version for both 1.18 and 1.19?

nixpanic:
> @nixpanic this is one of the problems we are going to face if we try to deploy the latest available kubernetes versions which are not supported by minikube. how about sticking to a version for both 1.18 and 1.19?

This has not been an issue before, the jobs have been running with many patch release updates already. Kubernetes wants to release stable patch releases, so deployment should not really be affected with such a minor update. Why is it the case with this patch release, is not clear yet.

We should always test with supported versions, otherwise we'll miss out on unintentional breakage.

Yuggupta27 (approved):
Thanks for the PR. LGTM

nixpanic:
Btw, this is also the [latest stable release](https://storage.googleapis.com/kubernetes-release/release/stable.txt), so being unable to install with minikube is likely a major issue.

Madhu-1 (approved):
to unblock CI lets go ahead with the Fix, Thanks @nixpanic

Issue #1566, opened 2020-10-09
Allow bigger size restore/clone for RBD Blockmode  PVCs


humblec:
# Describe the bug #

CSI spec allows a user to create bigger volumes at restore/clone path. Ideally we should enable it in our driver too. This issue tracks the requirement for RBD Block mode PVCs.

humblec:
Cc @nixpanic , also please feel free to add the details of this requirement :+1:

humblec:
@nixpanic jfyi, I had a quick try on solving this and it looks like I am able to get it working for PVC based Clone with Block Mode. 

The parent volume size is `1GiB` and the request is for `2GiB`: 
```

..I1009 14:22:34.742686       1 utils.go:159] ID: 14 Req-ID: pvc-aa3d2c03-ebab-4289-b08e-8a94ba2eebbe GRPC call: /csi.v1.Controller/CreateVolume
I1009 14:22:34.746710       1 utils.go:160] ID: 14 Req-ID: pvc-aa3d2c03-ebab-4289-b08e-8a94ba2eebbe GRPC request: {"capacity_range":{"required_bytes":2147483648},"name":"pvc-aa3d2c03-ebab-4289-b08e-8a94ba2eebbe","parameters":{"clusterID":"rook-ceph","imageFeatures":"layering","imageFormat":"2","pool":"replicapool"},"secrets":"***stripped***","volume_capabilities":[{"AccessType":{"Block":{}},"access_mode":{"mode":1}}],"volume_content_source":{"Type":{"Volume":{"volume_id":"0001-0009-rook-ceph-0000000000000001-e7cd81a1-0a14-11eb-a983-4a1640963d45"}}}}
..........


[...]

PR #1594, opened 2020-10-19
Resize rbd block

humblec:
Fixes: #1566 

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	v1 "k8s.io/api/core/v1"
 	scv1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -500,7 +502,7 @@ func oneReplicaDeployYaml(template string) string {
 	return re.ReplaceAllString(template, `$1 1`)
 }
 
-func validatePVCClone(sourcePvcPath, clonePvcPath, clonePvcAppPath string, f *framework.Framework) {
+func validatePVCClone(sourcePvcPath, clonePvcPath, clonePvcAppPath, biggerSizeStr string, f *framework.Framework) {
 	var wg sync.WaitGroup
 	totalCount := 10
 	pvc, err := loadPVC(sourcePvcPath)
@@ -516,6 +518,9 @@ func validatePVCClone(sourcePvcPath, clonePvcPath, clonePvcAppPath string, f *fr
 	// validate created backend rbd images
 	validateRBDImageCount(f, 1)
 	pvcClone, err := loadPVC(clonePvcPath)
+	if biggerSizeStr != "" {
+		pvcClone.Spec.Resources.Requests[v1.ResourceStorage] = resource.MustParse(biggerSizeStr)
+	}
 	if err != nil {
 		e2elog.Failf("failed to load PVC with error %v", err)
 	}
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -235,7 +235,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	defer cr.DeleteCredentials()
 
-	rbdVol, err := cs.parseVolCreateRequest(ctx, req)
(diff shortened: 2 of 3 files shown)


internal/rbd/controllerserver.go
```diff
@@ -300,6 +306,16 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, err
 	}
 
+	// we have to resize the RBD volume with block mode request
+	if expansionRequired {
+		util.DebugLog(ctx, "\n requested volume size: %v is greater than the parent volume, increasing the size", newVolSize)
```
Madhu-1, line 311: Why we need debug log here?

Madhu-1:
@humblec are we resizing the rbd block device? if not, can you please update the PR title and provide few details in the PR description.

internal/rbd/controllerserver.go
```diff
@@ -300,6 +306,16 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+		err := rbdVol.resize(newVolSize)
```
Madhu-1, line 312: what if the volume is created successfully and resize fails are we validating the new size and making sure the new requested volume has a new size?

internal/rbd/controllerserver.go
```diff
@@ -188,31 +188,37 @@ func getGRPCErrorForCreateVolume(err error) error {
 	return status.Error(codes.Internal, err.Error())
 }
 
-// validateRequestedVolumeSize validates the request volume size with the
-// source snapshot or volume size, if there is a size mismatches it returns an error.
-func validateRequestedVolumeSize(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cr *util.Credentials) error {
+// source snapshot or volume size, if there is a size missmatch and if the new volume is not block mode
+// it returns an error.
+func validateRequestedVolumeSize(rbdVol *rbdVolume, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cr *util.Credentials, isBlock bool) (bool, error) {
```
Madhu-1, line 193: why we need a bool flag in return here? if we return bool from here we are doing two validation, can we skip this one as do validation in the caller itself?

nixpanic:
> /retest all

CephFS tests might be failing, rebase once #1599 is merged

internal/rbd/controllerserver.go
```diff
@@ -90,7 +90,7 @@ func (cs *ControllerServer) validateVolumeReq(ctx context.Context, req *csi.Crea
 	return nil
 }
 
-func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.CreateVolumeRequest) (*rbdVolume, error) {
+func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.CreateVolumeRequest) (*rbdVolume, bool, error) {
```
nixpanic, line 93: Please document the `bool` for return value. Users of the function do not know what the bool means.

nixpanic: The callers of this function need to be adjusted in the same commit. This is currently not compiling, which means `git bisect` will break.

internal/rbd/controllerserver.go
nixpanic: I do not get it either. The return value needs to be explained in the comment of the function. Please follow go-doc standards too.

internal/rbd/controllerserver.go
```diff
@@ -235,7 +235,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	defer cr.DeleteCredentials()
 
-	rbdVol, err := cs.parseVolCreateRequest(ctx, req)
+	rbdVol, isBlock, err := cs.parseVolCreateRequest(ctx, req)
```
nixpanic, line 238: Maybe introduce a `rbdVol.isBlock()` function or member instead? That prevents the need for the additional boolean return values and parameters

e2e/utils.go
```diff
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
```
nixpanic, line 16: follow standard import guidelines please

nixpanic (changes requested):

humblec:
Revisit this PR>

PR #1595, opened 2020-10-19
util: remove unused CSI server initialization functions

humblec:

### DIFF
--- a/internal/csi-common/utils.go
+++ b/internal/csi-common/utils.go
@@ -82,33 +82,6 @@ func NewControllerServiceCapability(ctrlCap csi.ControllerServiceCapability_RPC_
 	}
 }
 
-// RunNodePublishServer starts node server.
-func RunNodePublishServer(endpoint, hstOption string, d *CSIDriver, ns csi.NodeServer, m bool) {
-	ids := NewDefaultIdentityServer(d)
-
-	s := NewNonBlockingGRPCServer()
-	s.Start(endpoint, hstOption, ids, nil, ns, m)
-	s.Wait()
-}
-
-// RunControllerPublishServer starts controller server.
-func RunControllerPublishServer(endpoint, hstOption string, d *CSIDriver, cs csi.ControllerServer, m bool) {
-	ids := NewDefaultIdentityServer(d)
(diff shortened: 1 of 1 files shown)


nixpanic (changes requested):
Can you add a note in the commit since when/why these are not needed anymore?

Also note that CI in not stable at the moment, #1599 attempts to fix that, no need to retry tests before that one is merged (and this PR is rebased)

nixpanic:
> /retest all

Could you please address the request I made, and rebase this PR afterwards? Without the rebase, it is very likely that this keeps failing.

pkalever:
The go static-check tool is supposed to catch it. Not sure if it's an issue with the tool or we just ignoring the warnings from CI?

nixpanic:
> The go static-check tool is supposed to catch it. Not sure if it's an issue with the tool or we just ignoring the warnings from CI?

What do you mean, what should it catch? If you are talking about detecting functions that are not used, the problem is that these functions are exported in the package (start with a Captial case), so other non-local packages could use them.

pkalever:
> > The go static-check tool is supposed to catch it. Not sure if it's an issue with the tool or we just ignoring the warnings from CI?
> 
> What do you mean, what should it catch? If you are talking about detecting functions that are not used, the problem is that these functions are exported in the package (start with a Captial case), so other non-local packages could use them.

Oops! sorry, I thought this is the dead code used within the repository.

nixpanic:
> Oops! sorry, I thought this is the dead code used within the repository.

It is! But tools will have difficulties detecting it. If there are any tools that can help with that, we should definitely look into that.

humblec:
@nixpanic 
>
Can you add a note in the commit since when/why these are not needed anymore?
>
@nixpanic  I dont know .. Its been there for an year or so. Its an unused functions so I dont know how to answer `why these are not needed anymore` ?

nixpanic (approved):

nixpanic:
@mergifyio rebase

Madhu-1 (approved):

nixpanic:
@mergifyio refresh

PR #1600, opened 2020-10-19
cephfs: replace with a different variable name for builtin `cap` attribute

humblec:

### DIFF
--- a/internal/cephfs/util.go
+++ b/internal/cephfs/util.go
@@ -67,8 +67,8 @@ func (cs *ControllerServer) validateCreateVolumeRequest(req *csi.CreateVolumeReq
 		return status.Error(codes.InvalidArgument, "volume Capabilities cannot be empty")
 	}
 
-	for _, cap := range reqCaps {
-		if cap.GetBlock() != nil {
+	for _, capability := range reqCaps {
+		if capability.GetBlock() != nil {
 			return status.Error(codes.Unimplemented, "block volume not supported")
 		}
 	}
--- a/internal/csi-common/driver.go
+++ b/internal/csi-common/driver.go
@@ -73,8 +73,8 @@ func (d *CSIDriver) ValidateControllerServiceRequest(c csi.ControllerServiceCapa
 		return nil
 	}
 
-	for _, cap := range d.cap {
-		if c == cap.GetRpc().GetType() {
+	for _, capability := range d.cap {
+		if c == capability.GetRpc().GetType() {
 			return nil
 		}
 	}
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -373,8 +373,8 @@ func (cs *ControllerServer) ValidateVolumeCapabilities(
 	ctx context.Context,
 	req *csi.ValidateVolumeCapabilitiesRequest) (*csi.ValidateVolumeCapabilitiesResponse, error) {
 	// Cephfs doesn't support Block volume
-	for _, cap := range req.VolumeCapabilities {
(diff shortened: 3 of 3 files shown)


internal/csi-common/driver.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	klog "k8s.io/klog/v2"
```
Madhu-1, line 25: why this is changed?

internal/cephfs/util.go
```diff
@@ -67,8 +67,8 @@ func (cs *ControllerServer) validateCreateVolumeRequest(req *csi.CreateVolumeReq
 		return status.Error(codes.InvalidArgument, "volume Capabilities cannot be empty")
 	}
 
-	for _, cap := range reqCaps {
```
Madhu-1, line 70: instead of full name, I would like to go with `c`

Madhu-1 (changes requested):

internal/csi-common/driver.go
humblec: the package naming is not required. Its redundant, however I could put it in a seperate commit.

internal/cephfs/util.go
humblec: Why? with `capability` its explicit and clear what we are ranging on.

internal/csi-common/driver.go
humblec: @Madhu-1 do you want me to put it in a seperate commit?

Madhu-1: This alias is present in other places also. I don't see a reason to change it unless some standard says we don't need to add an alias

internal/cephfs/util.go
Madhu-1: https://github.com/golang/go/wiki/CodeReviewComments#variable-names

humblec: `More unusual things and global variables need more descriptive names` -> this is the case here, Isnt it? I think thats why pick `cap` in first place. 

@nixpanic whats your thought here? 
My preference goes with `capability` though to make it clear.

nixpanic: `c` is fine, `req.VolumeCapabilities` is very explicit, and `ReqCaps` does not have the full name either, assuming the reader already knows what `Caps` refers to.

But, `capability` isn't an issue for me.

internal/csi-common/driver.go
nixpanic: I do not see any references changed from `klog` to `v2`. This changes looks unrelated to the renaming f the `cap` variable and should be in a different PR.

nixpanic (changes requested):

internal/cephfs/util.go
humblec: Considering we do have assignments like https://github.com/ceph/ceph-csi/pull/1600/files#diff-9eb942b29a818285935a4ffb1f6584a4d9e86317fbaf07a40a5029b35c2713b7R77 below , keeping it on `capability`. If really needed I can change to `c`.

internal/csi-common/driver.go
humblec: removed `klog` removal from this comment. That can be a separate PR altogether.

nixpanic (approved):

nixpanic:
@mergifyio rebase

Madhu-1 (approved):

PR #1601, opened 2020-10-19
cephfs: fix wrong error check in CreateVolume rollback action

humblec:
Previously the purgeVolume error was ignored due to wrong error variable
check in the createVolume. With this change it checks on the proper error

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -188,11 +188,13 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 			// explictly
 			err = resizeVolume(ctx, volOptions, cr, volumeID(vID.FsSubvolName), volOptions.Size)
 			if err != nil {
-				if purgeErr := purgeVolume(ctx, volumeID(vID.FsSubvolName), cr, volOptions, false); err != nil {
+				purgeErr := purgeVolume(ctx, volumeID(vID.FsSubvolName), cr, volOptions, false)
+				if purgeErr != nil {
 					util.ErrorLog(ctx, "failed to delete volume %s: %v", requestName, purgeErr)
 					// All errors other than ErrVolumeNotFound should return an error back to the caller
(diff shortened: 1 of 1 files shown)


Madhu-1 (approved):
@humblec please provide a description of what was the issue.

internal/cephfs/controllerserver.go
```diff
@@ -188,7 +188,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 			// explictly
 			err = resizeVolume(ctx, volOptions, cr, volumeID(vID.FsSubvolName), volOptions.Size)
 			if err != nil {
-				if purgeErr := purgeVolume(ctx, volumeID(vID.FsSubvolName), cr, volOptions, false); err != nil {
+				if purgeErr := purgeVolume(ctx, volumeID(vID.FsSubvolName), cr, volOptions, false); purgeErr != nil {
```
nixpanic, line 191: These issues can quickly happen. It is much easier to spot when the error assignment and if condition are on two lines. Could you please improve the maintainability that way?

Also, what is the (user visible) problem that this solves? It needs to be added in the commit message.

nixpanic (changes requested):

internal/cephfs/controllerserver.go
humblec: Done.

nixpanic (approved):
Thanks!
Does that mean users would not have noticed this problem, as you didn't answer that question?

humblec:
> Thanks!
> Does that mean users would not have noticed this problem, as you didn't answer that question?

Well, if there was an error in `resizeVolume` previoulsy it would have been always reported that, `failed to delete the volume` and then further error return check would have happened as we were `relying` on `resizeVolume` error previously.

humblec:
@Mergifyio rebase

Madhu-1 (approved):

humblec:
@Mergifyio rebase

PR #1604, opened 2020-10-19
ci: use the credentials-binding plugin for the GitHub API Token

nixpanic:
The standard credentials() function does not seem to do what we need. So
use the credentials-binding Jenkins plugin instead.

See-also: https://www.jenkins.io/doc/pipeline/steps/credentials-binding/


Madhu-1:
@nixpanic CI failure

Madhu-1 (approved):

Issue #1551, opened 2020-10-02
cephfs: replace getSubVolumeInfo() with a go-ceph implementation


nixpanic:
Once #1545 is merged, implement getSubVolumeInfo() with go-ceph language bindings too.

PR #1578, opened 2020-10-14
cephfs: implement getMetadataPool() with go-ceph

nixpanic:
Reduce the number of `ceph fs` command executions, and use go-ceph with a cached/shared connection.

Fixes: #1554
~Do-Not-Merge: depends on #1545, #1581 and #1577, which need to get merged first~

PR #1612, opened 2020-10-20
 cephfs: implement getSubVolumeInfo() with go-ceph

nixpanic:
Reduce the number of `ceph fs` command executions, and use go-ceph with a cached/shared connection.

Fixes: #1551 
~Do-Not-Merge: depends on #1578, which needs to get merged first~

### DIFF
--- a/internal/cephfs/volume.go
+++ b/internal/cephfs/volume.go
@@ -94,7 +94,7 @@ func getVolumeRootPathCeph(ctx context.Context, volOptions *volumeOptions, cr *u
 	return strings.TrimSuffix(stdout, "\n"), nil
 }
 
-func getSubVolumeInfo(ctx context.Context, volOptions *volumeOptions, cr *util.Credentials, volID volumeID) (Subvolume, error) {
+func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, cr *util.Credentials, volID volumeID) (Subvolume, error) {
 	info := Subvolume{}
 	err := execCommandJSON(
 		ctx,
@@ -103,11 +103,11 @@ func getSubVolumeInfo(ctx context.Context, volOptions *volumeOptions, cr *util.C
 		"fs",
 		"subvolume",
 		"info",
-		volOptions.FsName,
+		vo.FsName,
 		string(volID),
 		"--group_name",
-		volOptions.SubvolumeGroup,
-		"-m", volOptions.Monitors,
+		vo.SubvolumeGroup,
+		"-m", vo.Monitors,
 		"-c", util.CephConfigPath,
 		"-n", cephEntityClientPrefix+cr.ID,
 		"--keyfile="+cr.KeyFile)
--- a/internal/cephfs/volume.go
+++ b/internal/cephfs/volume.go
@@ -18,6 +18,7 @@ package cephfs
 
 import (
 	"context"
+	"fmt"
 	"path"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


nixpanic:
@mergifyio rebase

internal/cephfs/volume.go
```diff
@@ -44,23 +44,12 @@ const (
...
-	Mode          int      `json:"mode"`
-	MonAddrs      []string `json:"mon_addrs"`
-	Path          string   `json:"path"`
-	PoolNamespace string   `json:"pool_namespace"`
-	// The subvolume "state" is based on the current state of the subvolume.
-	// It contains one of the following values:
-	// * "complete": subvolume is ready for all operations.
-	// * "snapshot-retained": subvolume is removed but its snapshots are retained.
-	State string `json:"state"`
-	Type  string `json:"type"`
```
humblec, line 62: @nixpanic just wondering is it a good idea to drop these fields. It helps in one way to fill it if needed and also gives us some view on possible options available in the response. I dont have any hard feeling on this, but somehow in a mindset to keep it even if we are not using it.

nixpanic: I left the reference to [cephfs/admin.SubVolumeInfo](https://godoc.org/github.com/ceph/go-ceph/cephfs/admin#SubVolumeInfo) so that it is easy to check what other members are available. Not filling unused members is a little more efficient, cleaner and improves readability a bit.

internal/cephfs/volume.go
```diff
@@ -94,36 +83,38 @@ func getVolumeRootPathCeph(ctx context.Context, volOptions *volumeOptions, cr *u
...
-		"-n", cephEntityClientPrefix+cr.ID,
-		"--keyfile="+cr.KeyFile)
+func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, volID volumeID) (*Subvolume, error) {
+	fsa, err := vo.conn.GetFSAdmin()
+	if err != nil {
+		util.ErrorLog(ctx, "could not get FSAdmin, can not fetch metadata pool for %s:", vo.FsName, err)
+		return nil, err
+	}
+
+	info, err := fsa.SubVolumeInfo(vo.FsName, vo.SubvolumeGroup, string(volID))
```
Madhu-1, line 93: @nixpanic do we get the same error messages what we used to get it when we run CLI commands? can you please check it

nixpanic: Yes, it seems that the Ceph Mgr uses the same functions internally, and error codes and messages are standardized.

Madhu-1 (dismissed):

internal/cephfs/volume.go
```diff
@@ -94,36 +83,38 @@ func getVolumeRootPathCeph(ctx context.Context, volOptions *volumeOptions, cr *u
...
 		}
 
-		return info, err
+		return nil, err
 	}
-	return info, nil
+
+	subvol := Subvolume{
+		// FIXME: info.BytesQuota could be of type fsAdmin.Infinite
+		BytesQuota: int64(info.BytesQuota.(fsAdmin.ByteCount)),
```
phlogistonjohn, line 109: If ceph csi always expects a quota to be set on a subvolume it created you could just do something like:
```
bq, ok := info.BytesQuota.(fsAdmin.ByteCount)
if !ok {
    return nil, fmt.Errorf("some error text")
}
``` 
above.

nixpanic: Oh, yes, I was lazy when I wrote this, and forgot to address the `FIXME` notes.

internal/cephfs/volume.go
```diff
@@ -94,36 +83,45 @@ func getVolumeRootPathCeph(ctx context.Context, volOptions *volumeOptions, cr *u
...
 		}
 
-		return info, err
+		return nil, err
 	}
-	return info, nil
+
+	var bq int64
+	bc, ok := info.BytesQuota.(fsAdmin.ByteCount)
+	if !ok {
```
Madhu-1, line 109: we are not supporting infinite, isn't it? how about returning an error message if this fails.

nixpanic: Sure, that makes things a little simpler. It is not clear from the CSI spec if 'infinite' (or unset capacity) is a common option. Will drop it with the next update.

Madhu-1 (approved):

nixpanic:
@humblec care to re-review?

humblec:
Once again not a great fan of removing the subvolume info members which are unused. considering others are agreeing on removing it, I am sticking to major votes :)

humblec (approved):

PR #1625, opened 2020-10-23
ci: increase deploy-timeout for upgrade-tests-cephfs job

nixpanic:
The upgrade-tests-cephfs fails relative regularly with the following
error during intial deployment:

    timeout waiting for deployment csi-cephfsplugin-provisioner with error error waiting for deployment "csi-cephfsplugin-provisioner" status to match expectation: etcdserver: request timed out

Increasing the timeout to 300 seconds (default 10) might make the job
more reliable.

### DIFF
--- a/upgrade-tests.groovy
+++ b/upgrade-tests.groovy
@@ -109,7 +109,7 @@ node('cico-workspace') {
 			timeout(time: 120, unit: 'MINUTES') {
 				upgrade_args = "--test-cephfs=false --test-rbd=true --upgrade-testing=true"
 				if ("${test_type}" == "cephfs"){
-					upgrade_args = "--test-cephfs=true --test-rbd=false --upgrade-testing=true"
+					upgrade_args = "--deploy-timeout=300 --test-cephfs=true --test-rbd=false --upgrade-testing=true"
 				}
 				ssh "cd /opt/build/go/src/github.com/ceph/ceph-csi && make run-e2e E2E_ARGS=\"--upgrade-version=${csi_upgrade_version} ${upgrade_args}\""
 			}


upgrade-tests.groovy
```diff
@@ -109,7 +109,7 @@ node('cico-workspace') {
 			timeout(time: 120, unit: 'MINUTES') {
 				upgrade_args = "--test-cephfs=false --test-rbd=true --upgrade-testing=true"
```
Yuggupta27, line 110:
```suggestion
				upgrade_args = "--test-cephfs=false --test-rbd=true --deploy-timeout=300 --upgrade-testing=true"
```

Yuggupta27 (changes requested):
should we do simlar for rbd?

upgrade-tests.groovy
```diff
@@ -109,7 +109,7 @@ node('cico-workspace') {
...
 				if ("${test_type}" == "cephfs"){
-					upgrade_args = "--test-cephfs=true --test-rbd=false --upgrade-testing=true"
+					upgrade_args = "--deploy-timeout=300 --test-cephfs=true --test-rbd=false --upgrade-testing=true"
```
Madhu-1, line 112: this won't be helpful the error we are getting is from the kubernetes server. we may need to handle this error in our e2e code.

upgrade-tests.groovy
nixpanic: well, rbd does not seem to fail that often

upgrade-tests.groovy
nixpanic: oh, yes, maybe error handling is not done sufficiently there

nixpanic: #1626 could be a better fix

nixpanic:
Closing in favour of #1626

PR #1640, opened 2020-10-27
ci: Fix shellcheck warnings for prepare.sh

Yuggupta27:
Shellcheck suggests the use of double quotes to prevent
word splitting, hence adding the same.

```
shellcheck --external-sources ./prepare.sh 
In ./prepare.sh line 90:
git clone ${depth} --branch="${base}" "${gitrepo}" "${workdir}"
          ^------^ SC2086: Double quote to prevent globbing and word splitting.
```

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -87,7 +87,7 @@ then
     depth=''
 fi
 
-git clone ${depth} --branch="${base}" "${gitrepo}" "${workdir}"
+git clone "${depth}" --branch="${base}" "${gitrepo}" "${workdir}"
 cd "${workdir}"
 git fetch origin "${ref}:tip/${ref}"
 git checkout "tip/${ref}"


Madhu-1 (approved):

pkalever (approved):

nixpanic (changes requested):
I do not think this works when you pass `--history` on the commandline. Please confirm that I'm wrong.

nixpanic:
Also note that the link in the commit message and PR will result in a 404 error after a while. Better copy/paste the error message that ShellCheck gave.

Yuggupta27:
> I do not think this works when you pass `--history` on the commandline. Please confirm that I'm wrong.

I tested it locally, on passing the `--history` argument, it does the intended and does not use depth while cloning.

nixpanic (approved):

PR #1621, opened 2020-10-22
doc: design document for rbd mirroring

Madhu-1:
This document outlines the internal cephcsi design to handle mirrored RBD images.

This design is an alternative solution to https://github.com/ceph/ceph-csi/pull/1305  for achieving rbd mirroring 

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>
Co-authored-by: ShyamsundarR <srangana@redhat.com>

PR #1641, opened 2020-10-28
rbd: add new controller to regenerate OMAP data

Madhu-1:
In the case of Disaster Recovery failover, the user expected to create the static PVC's. We have planned not to go with the PVC name and namespace for many reasons (as in kubernetes it's planned to support PVC transfer to a new namespace with a different name and with new features coming in like data populator etc). For now, we are planning to go with static PVC's to support async mirroring.
    
During Async mirroring only the RBD images are mirrored to the secondary site, and when the user creates the static PVC's on the failover we need to regenerate the omap data. The volumeHandle in PV spec is an encoded string which contains clusterID and poolID and image UUID, The clusterID and poolID won't remain same on both the clusters, for that cephcsi need to generate the new volumeHandle and its to create a mapping between new volumeHandle and old volumeHandle with that whenever cephcsi gets csi requests it check if the mapping exists it will pull the new volumeHandle and continues other operations.
    
implements #1621    


Madhu-1:
added DNM as I need to run E2E locally (functionality works tested manually)

internal/util/util.go
```diff
@@ -68,13 +68,14 @@ var (
...
-	Endpoint     string // CSI endpoint
-	DriverName   string // name of the driver
-	NodeID       string // node id
-	InstanceID   string // unique ID distinguishing this instance of Ceph CSI
-	PluginPath   string // location of cephcsi plugin
-	DomainLabels string // list of domain labels to read from the node
+	Vtype           string // driver type [rbd|cephfs|liveness|controller]
+	Endpoint        string // CSI endpoint
+	DriverName      string // name of the driver
+	DriverNamespace string // namespace in which driver is deployed
```
nixpanic, line 74: this change, and the part in `main()` can be a separate commit once the functions in the rbd and journal package have been added

e2e/rbd_helper.go
```diff
@@ -72,6 +72,7 @@ func createRBDStorageClass(c kubernetes.Interface, f *framework.Framework, scOpt
 		mOpt := strings.Split(opt, ",")
 		sc.MountOptions = append(sc.MountOptions, mOpt...)
 	}
+	sc.ReclaimPolicy = &retainPolicy
```
nixpanic, line 75: this is also a global variable in `e2e/utils.go`, use `policy` here to refer to the parameter of the function

e2e/utils.go
```diff
@@ -37,6 +37,9 @@ const (
 
 	rookTolBoxPodLabel = "app=rook-ceph-tools"
 	rbdmountOptions    = "mountOptions"
+
+	retainPolicy = v1.PersistentVolumeReclaimRetain
+	deletePolicy = v1.PersistentVolumeReclaimDelete
```
nixpanic, line 43: instead of `deletePolicy`, maybe use `defaultPolicy` or add a comment that delete should be the default for most e2e tests

e2e/pvc.go
```diff
@@ -74,6 +74,73 @@ func createPVCAndvalidatePV(c kubernetes.Interface, pvc *v1.PersistentVolumeClai
 	})
 }
 
+func createPVCAndPV(c kubernetes.Interface, pvc *v1.PersistentVolumeClaim, pv *v1.PersistentVolume) error {
+	_, err := c.CoreV1().PersistentVolumeClaims(pvc.Namespace).Create(context.TODO(), pvc, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	_, err = c.CoreV1().PersistentVolumes().Create(context.TODO(), pv, metav1.CreateOptions{})
```
nixpanic, line 82: if this fails, the PVC was created already. Maybe log that it leaks, or delete it?

e2e/utils.go
```diff
@@ -572,6 +573,112 @@ func validatePVCClone(sourcePvcPath, clonePvcPath, clonePvcAppPath string, f *fr
 	validateRBDImageCount(f, 0)
 }
 
+func validateController(f *framework.Framework, pvcPath, appPath, scPath string) error {
```
nixpanic, line 576: this can probably be placed in a new file, `e2e/rbd_controller.go`. Assuming more tests will be added at a later time, for bugfixes and/or features.

nixpanic, line 576: A comment describing the steps of the validation would be a great help! I am not sure if all the validation done here is intended/needed. Might be helpful to split certain things out in helper functions so that different parts of the validation are easier to understand/review.

internal/rbd/controllerserver.go
```diff
@@ -732,7 +732,6 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		}
 		return nil, err
 	}
-	defer rbdVol.Destroy()
```
nixpanic, line 735: Might be better to remove the instance on line 720. I am not sure how Golang behaves, but the 1st instance might have been replaced by the return value on line 722, causing no `Destroy()` on the 2nd instance. Line 719 can be replaced by `var rbdVol *rbdVolume` as well.

nixpanic (changes requested):

internal/journal/voljournal.go
```diff
@@ -437,18 +437,22 @@ func (conn *Connection) UndoReservation(ctx context.Context,
 // reserveOMapName creates an omap with passed in oMapNamePrefix and a generated <uuid>.
 // It ensures generated omap name does not already exist and if conflicts are detected, a set
 // number of retires with newer uuids are attempted before returning an error.
-func reserveOMapName(ctx context.Context, monitors string, cr *util.Credentials, pool, namespace, oMapNamePrefix string) (string, error) {
+func reserveOMapName(ctx context.Context, monitors string, cr *util.Credentials, pool, namespace, oMapNamePrefix, volUUID string) (string, error) {
```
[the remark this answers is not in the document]
Madhu-1: I have updated the comment for the volUUID, I will see I can move this function to a method with a common structure as a follow-up.

internal/journal/voljournal.go
```diff
@@ -492,17 +498,18 @@ Return values:
 func (conn *Connection) ReserveName(ctx context.Context,
 	journalPool string, journalPoolID int64,
 	imagePool string, imagePoolID int64,
-	reqName, namePrefix, parentName, kmsConf string) (string, string, error) {
+	reqName, namePrefix, parentName, kmsConf, volUUID string) (string, string, error) {
 	// TODO: Take in-arg as ImageAttributes?
```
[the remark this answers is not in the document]
Madhu-1: will fix it as part of the cleanup, am working on cleanup the journal package separately

internal/rbd/rbd_util.go
```diff
@@ -972,6 +984,52 @@ func (rv *rbdVolume) getImageInfo() error {
 	return nil
 }
 
+// imageInfo strongly typed JSON spec for image info.
+type imageInfo struct {
+	Mirroring mirroring `json:"mirroring"`
+}
+
+// parentInfo  spec for parent volume  info.
+type mirroring struct {
```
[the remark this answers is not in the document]
Madhu-1: removed all unused members for now

cmd/cephcsi.go
```diff
@@ -50,9 +56,10 @@ var (
 
 func init() {
 	// common flags
-	flag.StringVar(&conf.Vtype, "type", "", "driver type [rbd|cephfs|liveness]")
+	flag.StringVar(&conf.Vtype, "type", "", "driver type [rbd|cephfs|liveness|controller]")
 	flag.StringVar(&conf.Endpoint, "endpoint", "unix://tmp/csi.sock", "CSI endpoint")
 	flag.StringVar(&conf.DriverName, "drivername", "", "name of the driver")
+	flag.StringVar(&conf.DriverNamespace, "drivernamespace", defaultNS, "namespace in which driver is deployed")
```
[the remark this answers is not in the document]
Madhu-1: it is possible to detect it. it's working as expected or else due to RBAC issue everything will fail

deploy/rbd/kubernetes/csi-provisioner-rbac.yaml
```diff
@@ -73,7 +73,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["configmaps"]
-    verbs: ["get", "list", "watch", "create", "delete"]
+    verbs: ["get", "list", "watch", "create", "update", "delete"]
```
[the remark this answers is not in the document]
Madhu-1: done

internal/rbd/rbd_journal.go
```diff
@@ -466,3 +467,149 @@ func undoVolReservation(ctx context.Context, rbdVol *rbdVolume, cr *util.Credent
 
 	return err
 }
+
+// RegenerateJournal regenerates the omap data for the static volumes, the
+// input parameters imageName, volumeID, pool, journalPool, requestName will be
+// present in the PV.Spec.CSI object based on that we can regenerate the
+// complete omap mapping between imageName and volumeID.
+
+// RegenerateJournal performs below operations
```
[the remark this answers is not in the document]
Madhu-1: will work on cleanup later as am planning to group journal logics?

e2e/pvc.go
Madhu-1: currently, if something fails we are not deleting/cleaning up the resources. the error will be more evident from the Create function will take care of cleanup in followup patches we need to change few big things in the E2E framework

e2e/utils.go
Madhu-1: kept it here as we will add support for cephfs later.

Madhu-1: added comments for better understanding.

internal/rbd/controllerserver.go
Madhu-1: in the case of Error also need to Destroy the context. so I think we can place Destroy next to the struct.

internal/controller/controller.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Config holds the drivername and namespace name.
+type Config struct {
+	DriverName string
+	Namespace  string
+}
+
+// ControllerList holds the list of managers need to be started.
+var ControllerList []ContollerManager
```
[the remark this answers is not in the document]
Madhu-1: Yes as all controllers need to be added to the list we need to export it.

Madhu-1:
CI failed during rook deployment

internal/rbd/rbd_journal.go
```diff
@@ -466,3 +467,149 @@ func undoVolReservation(ctx context.Context, rbdVol *rbdVolume, cr *util.Credent
...
+		}
+	}
+
+	if volumeID != rbdVol.VolID {
+		return j.ReserveNewUUIDMapping(ctx, rbdVol.JournalPool, volumeID, rbdVol.VolID)
+	}
+	return nil
+}
+
+func (rv *rbdVolume) addNewUUIDMapping(ctx context.Context, imagePoolID int64, j *journal.Connection) error {
```
nixpanic, line 598: consider commenting all new functions

internal/rbd/rbd_util.go
```diff
@@ -689,27 +692,41 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
...
-	rbdVol.JournalPool = rbdVol.Pool
+	defer j.Destroy()
 
-	rbdVol.RadosNamespace, err = util.RadosNamespace(util.CsiConfigFile, rbdVol.ClusterID)
+	// check is there any volumeID mapping exists.
+	id, err := j.CheckNewUUIDMapping(ctx, rbdVol.JournalPool, volumeID)
 	if err != nil {
-		return nil, err
+		return rbdVol, fmt.Errorf("failed to get volume id %s mapping %s",
+			volumeID, err)
```
nixpanic, line 710: use `%w` in the `Errorf()` call please. Mostly the format is like `"local message with local variables: original error"`

cmd/cephcsi.go
```diff
@@ -180,11 +187,29 @@ func main() {
...
+		}
 	}
 
 	os.Exit(0)
 }
 
+// initControllers will initialize all the controllers.
+func initControllers() {
+	// Add list of controller here.
+	persistentvolume.Init()
```
nixpanic, line 210: If you call the function `init()`, you do not need to call it explicitly here, just importing the package should be sufficient.

internal/rbd/controllerserver.go
```diff
@@ -622,7 +622,6 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		}
 		return &csi.DeleteVolumeResponse{}, nil
 	}
```
nixpanic, line 624: I am not certain it is a extra `Destroy()`. The initial defer is placed on an object that is empty, and the object/pointer is overwritten by a return value of a function. Better would be to do
```golang
var rbdVol *rbdVolume

rbdVol, err = genVolFromVolID(...)
if err != nil {
    ...
}
defer rbdVol.Destroy()
```

This first destroy in the current code might be completely useless.

charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
```diff
@@ -172,6 +172,42 @@ spec:
...
+            - "--drivername=$(DRIVER_NAME)"
+            - "--drivernamespace=$(DRIVER_NAMESPACE)"
+          env:
+            - name: DRIVER_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: DRIVER_NAME
+              value: {{ .Values.driverName }}
+          securityContext:
```
nixpanic, line 191: these are very high permissions, is that really needed?

charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
```diff
@@ -172,6 +172,42 @@ spec:
...
+            privileged: true
+            capabilities:
+              add: ["SYS_ADMIN"]
+            allowPrivilegeEscalation: true
+          volumeMounts:
```
nixpanic, line 196: I am not sure if the controller really needs all of the following mounts, it does not load kernel modules and does not use devices directly?

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
```diff
@@ -148,6 +148,32 @@ spec:
...
+            - "--v=5"
+            - "--drivername=rbd.csi.ceph.com"
+            - "--drivernamespace=$(DRIVER_NAMESPACE)"
+          env:
+            - name: DRIVER_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          imagePullPolicy: "IfNotPresent"
+          volumeMounts:
```
nixpanic, line 165: No security context as in the Helm chart? Are these mounts really needed?

nixpanic (commented):
LGTM, just recommended some changes to improve maintainability

internal/rbd/rbd_util.go
```diff
@@ -972,6 +984,52 @@ func (rv *rbdVolume) getImageInfo() error {
...
+
+// parentInfo  spec for parent volume  info.
+type mirroring struct {
+	Mode     string `json:"mode"`
+	State    string `json:"state"`
+	GlobalID string `json:"global_id"`
+	Primary  bool   `json:"primary"`
+}
+
+// updateVolWithImageInfo updates provided rbdVolume with information from on-disk data
```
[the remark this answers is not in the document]
Madhu-1: already mentioned where this function is called for easy tracking

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -0,0 +1,167 @@
...
+
+// Init will add the ReconcilePersistentVolume to the list.
+func Init() {
+	// add ReconcilePersistentVolume to the list
+	ctrl.ControllerList = append(ctrl.ControllerList, ReconcilePersistentVolume{})
+}
+
+// Add adds the newPVReconciler.
+func (r ReconcilePersistentVolume) Add(mgr manager.Manager, config ctrl.Config) error {
+	return add(mgr, newPVReconciler(mgr, config))
```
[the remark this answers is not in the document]
Madhu-1: for readability i kept separate function, if required I can add implementation here

cmd/cephcsi.go
Madhu-1: done it explicitly as this should not be called for cephfs or node servers.

internal/rbd/controllerserver.go
Madhu-1: as per the above snippet, the Destroy won't be called in case of an error

charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
Madhu-1: all the container in the provisioner need to have same permission, am not able to find the thread (got this suggestion from kube team)

charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
Madhu-1: removed dev mount for now

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
Madhu-1: removed dev mount and added the missing security context

internal/rbd/controllerserver.go
nixpanic: that is fine, as `genVolFromVolID()` does not return an rbdVolume if an error occurs

charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
nixpanic: ok!

charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
nixpanic: what about the others? /sys and /lib/modules?

internal/rbd/controllerserver.go
Madhu-1: we already have an rbd connection opened we need to close that one right in case of error also. that why I moved the Destroy right after the declaration and even in case of error also `genVolFromVolID` returns the rbdVolume structure

charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
Madhu-1: Removed

nixpanic (changes requested):
I really think the removal of the `rbdVol.Destroy()` is not done correctly. Consider the following code:
```golang
package main

import "fmt"

type myDefer struct {
	name string
}

func (md *myDefer) Destroy() {
	fmt.Printf("destroying %T with name %q\n", md, md.name)
}

func main() {
	fmt.Println("before 1st defer")

	md := &myDefer{"first"}
	defer md.Destroy()
	fmt.Println("after 1st defer")

	fmt.Println("before 2nd allocation")
	md = &myDefer{"second"}
	fmt.Println("before 2nd defer")
	defer md.Destroy()
	fmt.Println("after 2nd defer")
}
```
And, when running:
```
$ go run defer.go
before 1st defer
after 1st defer
before 2nd allocation
before 2nd defer
after 2nd defer
destroying *main.myDefer with name "second"
destroying *main.myDefer with name "first"
```

Comment https://github.com/ceph/ceph-csi/pull/1641/commits/9e4825acf5548cbfcb8e4d3fcaba864ecaf0182e#r528528971 is still the most appropriate solution, I think.

internal/rbd/controllerserver.go
Madhu-1: update the Destory call

nixpanic (dismissed):

nixpanic:
@mergifyio rebase

internal/rbd/controllerserver.go
```diff
@@ -607,9 +607,9 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 	defer cs.OperationLocks.ReleaseDeleteLock(volumeID)
 
 	rbdVol := &rbdVolume{}
```
nixpanic, line 609: golang-ci complains about this, you'll need to use
```golang
var rbdVol *rbdVolume
```

Madhu-1: updated now

nixpanic (approved):

Madhu-1:
Adding ready-to-merge, enhancements can be done in followup PR's

Issue #1646, opened 2020-10-30
Cephfs: Proper Error message is not propagated during DeleteVolume when subvolume has snapshots


Madhu-1:
```text
ceph-0000000000000001-8a672fe1-1a81-11eb-911f-0242ac110007 got omap values: (pool="myfs-metadata", namespace="csi", name="csi.volume.8a672fe1-1a81-11eb-911f-0242ac110007"): map[csi.imagename:csi-vol-8a672fe1-1a81-11eb-911f-0242ac110007 csi.volname:pvc-0e8f98c1-d54f-4c66-8fa2-a987b2c74066]
madhu the stderror is Error ENOTEMPTY: subvolume 'csi-vol-8a672fe1-1a81-11eb-911f-0242ac110007' has snapshots
 and error is an error (exit status 39) occurred while running ceph args: [fs subvolume rm myfs csi-vol-8a672fe1-1a81-11eb-911f-0242ac110007 --group_name csi -m 10.103.71.203:6789 -c /etc/ceph/ceph.conf -n client.csi-cephfs-provisioner --keyfile=***stripped***] 
I1030 07:41:41.153718       1 cephcmds.go:53] ID: 30 Req-ID: 0001-0009-rook-ceph-0000000000000001-8a672fe1-1a81-11eb-911f-0242ac110007 an error (exit status 39) occurred while running ceph args: [fs subvolume rm myfs csi-vol-8a672fe1-1a81-11eb-911f-0242ac110007 --group_name csi -m 10.103.71.203:6789 -c /etc/ceph/ceph.conf -n client.csi-cephfs-provisioner --keyfile=***stripped***]
E1030 07:41:41.153803       1 volume.go:234] ID: 30 Req-ID: 0001-0009-rook-ceph-0000000000000001-8a672fe1-1a81-11eb-911f-0242ac110007 failed to purge subvolume csi-vol-8a672fe1-1a81-11eb-911f-0242ac110007 in fs myfs: an error (exit status 39) occurred while running ceph args: [fs subvolume rm myfs csi-vol-8a672fe1-1a81-11eb-911f-0242ac110007 --group_name csi -m 10.103.71.203:6789 -c /etc/ceph/ceph.conf -n client.csi-cephfs-provisioner --keyfile=***stripped***]
[...]

PR #1652, opened 2020-11-02
cephfs: Provide proper error message on DeleteVolume

Yuggupta27:
In case a subvolume has snapshots, the PV deletion
fails with an unclear error message.
Providing proper error message for the failed request
makes it easier to debug.

In scenario, if subvolume deletion is attempted which is not empty:

Error message received earlier:
```
ID: 39 Req-ID: 0001-000b-myClusterId-0000000000000001-fade3619-1cd9-11eb-b440-0242ac11000e an error (exit status 39) occurred while running ceph args: [fs subvolume rm myfs csi-vol-fade3619-1cd9-11eb-b440-0242ac11000e --group_name 	mysubgrp -m 10.96.144.69:6789 -c /etc/ceph/ceph.conf -n client.admin --keyfile=***stripped***]
```
Error message after patch:

```
ID: 31 Req-ID: 0001-000b-myClusterId-0000000000000001-80ce2451-1ce0-11eb-afd5-0242ac11000e The error is : Error ENOTEMPTY: subvolume 'csi-vol-80ce2451-1ce0-11eb-afd5-0242ac11000e' has snapshots
```

Fixes: https://github.com/ceph/ceph-csi/issues/1646

### DIFF
--- a/internal/cephfs/util.go
+++ b/internal/cephfs/util.go
@@ -34,7 +34,10 @@ import (
 type volumeID string
 
 func execCommandErr(ctx context.Context, program string, args ...string) error {
-	_, _, err := util.ExecCommand(ctx, program, args...)
+	_, stdErr, err := util.ExecCommand(ctx, program, args...)
+	if err != nil {
+		util.ErrorLog(ctx, "stdErr occurred: %s", stdErr)
+	}
 	return err
 }
 


internal/cephfs/util.go
```diff
@@ -34,7 +34,10 @@ import (
 type volumeID string
 
 func execCommandErr(ctx context.Context, program string, args ...string) error {
-	_, _, err := util.ExecCommand(ctx, program, args...)
+	_, stdErr, err := util.ExecCommand(ctx, program, args...)
```
Madhu-1, line 37: How this function was working earlier for other cases?

can you check when we will get an error in stdErr and err? we have e2e tests for negative scenarios also how it was passing earlier?

internal/cephfs/util.go
```diff
@@ -34,7 +34,10 @@ import (
...
+	if err != nil {
```
Madhu-1, line 38: in case of `err!=nil` we will return the stdErr how much helpful it is?

Madhu-1 (changes requested):

Yuggupta27:
Currently none of the conditions are being checked here, and directly error is being returned https://github.com/ceph/ceph-csi/blob/eefaf09adec77a003296a2abfdae84a0a805444c/internal/cephfs/volume.go#L234-L236

Madhu-1:
```go
[🎩︎]mrajanna@localhost ceph-csi $]grep -ri "execCommandErr" .
Binary file ./_output/cephcsi matches
./internal/cephfs/util.go:func execCommandErr(ctx context.Context, program string, args ...string) error {
./internal/cephfs/snapshot.go:	err := execCommandErr(
./internal/cephfs/snapshot.go:	err := execCommandErr(
./internal/cephfs/snapshot.go:	err := execCommandErr(
./internal/cephfs/snapshot.go:	err := execCommandErr(
./internal/cephfs/snapshot.go:	err := execCommandErr(
./internal/cephfs/volumemounter.go:	if err := execCommandErr(ctx, "modprobe", "ceph"); err != nil {
./internal/cephfs/volumemounter.go:	return execCommandErr(ctx, "mount", args[:]...)
./internal/cephfs/volumemounter.go:	if err := execCommandErr(ctx, "mount", "-o", mntOptionSli, from, to); err != nil {
./internal/cephfs/volumemounter.go:		if err := execCommandErr(ctx, "mount", "-o", mntOptionSli, to); err != nil {
./internal/cephfs/volumemounter.go:	if err := execCommandErr(ctx, "umount", mountPoint); err != nil {
./internal/cephfs/volume.go:	err := execCommandErr(ctx, "ceph", arg...)
```
check how this change will impact other functions.

Yuggupta27:
> Currently none of the conditions are being checked here, and directly error is being returned
> 
> https://github.com/ceph/ceph-csi/blob/eefaf09adec77a003296a2abfdae84a0a805444c/internal/cephfs/volume.go#L234-L236

All the conditions are finding the matching string from `err` which is not matching, so the conditions never return true..... Their intended usage should be to check a string.
For eg: The string `Error ENOTEMPTY` is a part of `stderr` and not `err`. So an attempt to find it in `err` will always return false

Madhu-1:
> Currently none of the conditions are being checked here, and directly error is being returned
> 
> https://github.com/ceph/ceph-csi/blob/eefaf09adec77a003296a2abfdae84a0a805444c/internal/cephfs/volume.go#L234-L236

check is  done at https://github.com/ceph/ceph-csi/blob/master/internal/cephfs/controllerserver.go#L352-L361 to return proper error

Madhu-1:
> > Currently none of the conditions are being checked here, and directly error is being returned
> > https://github.com/ceph/ceph-csi/blob/eefaf09adec77a003296a2abfdae84a0a805444c/internal/cephfs/volume.go#L234-L236
> 
> All the conditions are finding the matching string from `err` which is not matching, so the conditions never return true..... Their intended usage should be to check a string.
> For eg: The string `Error ENOTEMPTY` is a part of `stderr` and not `err`. So an attempt to find it in `err` will always return false

please check the errors at https://github.com/ceph/ceph-csi/blob/eefaf09adec77a003296a2abfdae84a0a805444c/internal/cephfs/errors.go#L25-L35 are properly checked? because all use the same function what you have changed now

internal/cephfs/util.go
Yuggupta27: It was passing earlier as it was always returning the [default error](https://github.com/ceph/ceph-csi/blob/master/internal/cephfs/volume.go#L241) in place of specialized (eg: `Error ENOTEMPTY`)

Yuggupta27:
> > > Currently none of the conditions are being checked here, and directly error is being returned
> > > https://github.com/ceph/ceph-csi/blob/eefaf09adec77a003296a2abfdae84a0a805444c/internal/cephfs/volume.go#L234-L236
> > 
> > 
> > All the conditions are finding the matching string from `err` which is not matching, so the conditions never return true..... Their intended usage should be to check a string.
> > For eg: The string `Error ENOTEMPTY` is a part of `stderr` and not `err`. So an attempt to find it in `err` will always return false
> 
> please check the errors at
> 
> https://github.com/ceph/ceph-csi/blob/eefaf09adec77a003296a2abfdae84a0a805444c/internal/cephfs/errors.go#L25-L35
> 
> are properly checked? because all use the same function what you have changed now

Sure Madhu, I'l re-verify the impact

Yuggupta27:
In place of returning stdErr, just logging it if an error occurs might solve the purpose for now; and provide an appropriate message for debugging.

internal/cephfs/util.go
Madhu-1, line 37: @Yuggupta27 using go-ceph binding for purgeVolume should fix the issue can you please check

humblec:
@Yuggupta27 would you be revisiting this PR ?

Yuggupta27:
Closing this as this issue no longer persists. Now a proper error message is returned in case of deletion attempt on a subvolume with snapshots:
```
Events:
  Type     Reason              Age                From                                                                                                    Message
  ----     ------              ----               ----                                                                                                    -------
  Warning  VolumeFailedDelete  12s (x7 over 67s)  cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-6c69c7ddb9-clcnr_3d651f31-9f94-4727-b121-99d313fe6eca  rpc error: code = FailedPrecondition desc = volume has snapshots: rados: ret=-39, Directory not empty: "subvolume 'csi-vol-c4413992-7293-11eb-b771-0242ac110012' has snapshots"
```

PR #1644, opened 2020-10-28
cephfs: replace getCloneInfo() with a go-ceph implementation

nixpanic:
getCloneInfo() does not need to return a full CloneStatus struct that
only has one member. Instead, it can just return the value of the single
member, so the JSON type/struct does not need to be exposed.

This makes the API for getCloneInfo() a little simpler, so it can be
replaced by a go-ceph implementation later on.

As there is little information returned about the clone status, the
function has been renamed to getCloneState() as well.

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)
- `/retest all`: run this in case the CentOS CI failed to start/report any test
  progress or results

</details>

PR #1660, opened 2020-11-04
cephfs: Add 'pending' state for clone status

Yuggupta27:
In certain cases, clone status can be 'pending'.
In that case, the abort error message should be
returned similar to that during 'in-progress'
state. 
This PR adds support for the 'pending' 
state of a clone.

```
sh-4.2# ceph fs clone status ocs-storagecluster-cephfilesystem csi-vol-d33566ec-1e5d-11eb-b02d-0a580a810216 --group_name csi
{
  "status": {
    "state": "pending",
    "source": {
      "volume": "ocs-storagecluster-cephfilesystem",
      "subvolume": "csi-vol-64b5c569-1e5a-11eb-b02d-0a580a810216",
      "snapshot": "csi-vol-d33566ec-1e5d-11eb-b02d-0a580a810216",
      "group": "csi"
    }
  }
}

```

### DIFF
--- a/internal/cephfs/clone.go
+++ b/internal/cephfs/clone.go
@@ -27,6 +27,8 @@ import (
 const (
 	// cephFSCloneFailed indicates that clone is in failed state.
 	cephFSCloneFailed = "failed"
+	// cephFSClonePending indicates that clone is in pending state.
+	cephFSClonePending = "pending"
 	// cephFSCloneInprogress indicates that clone is in in-progress state.
 	cephFSCloneInprogress = "in-progress"
 	// cephFSCloneComplete indicates that clone is in complete state.
@@ -94,6 +96,9 @@ func createCloneFromSubvolume(ctx context.Context, volID, cloneID volumeID, volO
 	case cephFSCloneInprogress:
 		util.ErrorLog(ctx, "clone is in progress for %v", cloneID)
 		return ErrCloneInProgress
+	case cephFSClonePending:
+		util.ErrorLog(ctx, "clone is pending for %v", cloneID)
+		return ErrClonePending
 	case cephFSCloneFailed:
 		util.ErrorLog(ctx, "clone failed for %v", cloneID)
 		cloneFailedErr := fmt.Errorf("clone %s is in %s state", cloneID, clone.Status.State)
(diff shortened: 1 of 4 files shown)


internal/cephfs/clone.go
```diff
@@ -94,6 +96,9 @@ func createCloneFromSubvolume(ctx context.Context, volID, cloneID volumeID, volO
 	case cephFSCloneInprogress:
 		util.ErrorLog(ctx, "clone is in progress for %v", cloneID)
 		return ErrCloneInProgress
+	case cephFSClonePending:
+		util.ErrorLog(ctx, "clone is pending for %v", cloneID)
+		return ErrClonePending
```
Madhu-1, line 101: you should return in-progress here, if not we need to handle inprogress and clonepending both error cases in caller

Madhu-1: sounds good to add both cases i thin

nixpanic:
From the comment it is not clear what the problem is why this is missing. That can be added to the commit message to explain the need for this change.

Also note that #1644 is in progress.

internal/cephfs/clone.go
nixpanic: Maybe add a helper function to make it easier to backoff/retry when cloning is not complete yet.

internal/cephfs/clone.go
```diff
@@ -158,7 +163,7 @@ func createCloneFromSnapshot(ctx context.Context, parentVolOpt, volOptions *volu
 	}
 	defer func() {
 		if err != nil {
-			if !errors.Is(err, ErrCloneInProgress) {
+			if !errors.Is(err, ErrCloneInProgress) || !errors.Is(err, ErrClonePending) {
```
nixpanic, line 166: This check can be don in a helper function, something like `isRetryableCloneErr()`, making it all a little better maintainable.

internal/cephfs/clone.go
```diff
@@ -94,6 +96,9 @@ func createCloneFromSubvolume(ctx context.Context, volID, cloneID volumeID, volO
 	case cephFSCloneInprogress:
 		util.ErrorLog(ctx, "clone is in progress for %v", cloneID)
 		return ErrCloneInProgress
+	case cephFSClonePending:
+		util.ErrorLog(ctx, "clone is pending for %v", cloneID)
```
Madhu-1, line 100: @Yuggupta27 may be in next PR convert this Error logs to Debug logs

Madhu-1 (approved):

internal/cephfs/clone.go
Yuggupta27: Sure Madhu... Will do.

nixpanic (approved):

PR #1661, opened 2020-11-04
ci: fix Blue Ocean status-url for JJB templates

nixpanic:
As ${RUN_DISPLAY_URL} also matches the format for template
substitutions, it can not be used like this for all Jenkins Jobs Builder
files. Instead, $RUN_DISPLAY_URL can be used, as it does not have the
{curly_braces} format and the template engine skips it.

Fixes: 87870369 (ci: link to the job in the Blue Ocean webui)
See-also: [errors while deploying jobs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/jjb-deploy/27/console)

### DIFF
--- a/jobs/mini-e2e.yaml
+++ b/jobs/mini-e2e.yaml
@@ -37,7 +37,7 @@
       lightweight-checkout: true
     triggers:
       - github-pull-request:
-          status-url: ${RUN_DISPLAY_URL}
+          status-url: $RUN_DISPLAY_URL
           status-context: 'ci/centos/mini-e2e/k8s-{k8s_version}'
           # yamllint disable-line rule:line-length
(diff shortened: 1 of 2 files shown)


Madhu-1 (approved):

jobs/mini-e2e.yaml
```diff
@@ -37,7 +37,7 @@
       lightweight-checkout: true
     triggers:
       - github-pull-request:
-          status-url: ${RUN_DISPLAY_URL}
+          status-url: $RUN_DISPLAY_URL
```
Madhu-1, line 40: Don't we need quote for this variable

nixpanic: no, local testing shows this is fine

PR #1663, opened 2020-11-04
 cephfs: replace createSnapshot() with a go-ceph implementation

nixpanic:
Reduce the number of `ceph fs` command executions, and use go-ceph with a cached/shared connection.

~Do-Not-Merge: depends on #1644, which needs to get merged first~

PR #1664, opened 2020-11-04
 cephfs: replace deleteSnapshot() with a go-ceph implementation

nixpanic:
Reduce the number of `ceph fs` command executions, and use go-ceph with a cached/shared connection.

~Do-Not-Merge: depends on #1644 and #1663, which need to get merged first~

### DIFF
--- a/internal/cephfs/snapshot.go
+++ b/internal/cephfs/snapshot.go
@@ -61,18 +61,18 @@ func (vo *volumeOptions) createSnapshot(ctx context.Context, snapID, volID volum
 	return nil
 }
 
-func deleteSnapshot(ctx context.Context, volOptions *volumeOptions, cr *util.Credentials, snapID, volID volumeID) error {
+func (vo *volumeOptions) deleteSnapshot(ctx context.Context, cr *util.Credentials, snapID, volID volumeID) error {
 	args := []string{
 		"fs",
 		"subvolume",
 		"snapshot",
 		"rm",
-		volOptions.FsName,
+		vo.FsName,
 		string(volID),
 		string(snapID),
 		"--group_name",
-		volOptions.SubvolumeGroup,
-		"-m", volOptions.Monitors,
+		vo.SubvolumeGroup,
+		"-m", vo.Monitors,
 		"-c", util.CephConfigPath,
 		"-n", cephEntityClientPrefix + cr.ID,
 		"--keyfile=" + cr.KeyFile,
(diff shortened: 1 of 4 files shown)


internal/cephfs/snapshot.go
```diff
@@ -61,30 +61,16 @@ func (vo *volumeOptions) createSnapshot(ctx context.Context, snapID, volID volum
...
+	if err != nil {
+		util.ErrorLog(ctx, "could not get FSAdmin, can delete snapshot %s from %s/%s: %s", string(snapID), vo.FsName, string(volID), err)
+		return err
 	}
 
-	err := execCommandErr(
-		ctx,
-		"ceph",
-		args[:]...)
+	err = fsa.RemoveSubVolumeSnapshot(vo.FsName, vo.SubvolumeGroup, string(volID), string(snapID))
```
Madhu-1, line 71: use `ForceRemoveSubVolumeSnapshot` so that ceph return success if the snapshot is not present or else we need to handle notfound error here.

internal/cephfs/snapshot.go
```diff
@@ -61,30 +61,16 @@ func (vo *volumeOptions) createSnapshot(ctx context.Context, snapID, volID volum
...
-		volOptions.SubvolumeGroup,
-		"-m", volOptions.Monitors,
-		"-c", util.CephConfigPath,
-		"-n", cephEntityClientPrefix + cr.ID,
-		"--keyfile=" + cr.KeyFile,
-		"--force",
+func (vo *volumeOptions) deleteSnapshot(ctx context.Context, snapID, volID volumeID) error {
+	fsa, err := vo.conn.GetFSAdmin()
+	if err != nil {
+		util.ErrorLog(ctx, "could not get FSAdmin, can delete snapshot %s from %s/%s: %s", string(snapID), vo.FsName, string(volID), err)
```
Madhu-1, line 67:
```suggestion
util.ErrorLog(ctx, "could not get FSAdmin %s", err)
```
Log only error?

internal/cephfs/snapshot.go
```diff
@@ -61,30 +61,16 @@ func (vo *volumeOptions) createSnapshot(ctx context.Context, snapID, volID volum
...
 	}
 
-	err := execCommandErr(
-		ctx,
-		"ceph",
-		args[:]...)
+	err = fsa.RemoveSubVolumeSnapshot(vo.FsName, vo.SubvolumeGroup, string(volID), string(snapID))
 	if err != nil {
-		util.ErrorLog(ctx, "failed to delete subvolume snapshot %s %s(%s) in fs %s", string(snapID), string(volID), err, volOptions.FsName)
+		util.ErrorLog(ctx, "failed to delete subvolume snapshot %s %s(%s) in fs %s", string(snapID), string(volID), err, vo.FsName)
```
Madhu-1, line 73: forgot :(. WE need to change  the logging here also as we done in createsnapshot

util.ErrorLog(ctx, "failed to delete subvolume snapshot %s %s in fs %s with error %s", string(volID),string(snapID), vo.FsName,err)

internal/cephfs/snapshot.go
nixpanic: sure, will do. I guess we assume the req-id and other details are logged at different locations.

internal/cephfs/snapshot.go
nixpanic: will do so

Madhu-1 (approved):

Issue #1636, opened 2020-10-27
Deploying Rook sometimes fails in the CI


nixpanic:
Sometimes deploying Rook fails, but no errors are available. The output looks like this:

```
...
configmap/rook-ceph-operator-config created
deployment.apps/rook-ceph-operator created
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  1187  100  1187    0     0  21196      0 --:--:-- --:--:-- --:--:-- 21196
[Pipeline] }
```

and also
```
...
cephfilesystem.ceph.rook.io/myfs created
cephblockpool.ceph.rook.io/replicapool created
Wait for rook deploy... 0s
...
Checking CEPH cluster state: []
Wait for rook deploy... 45s
The connection to the server 192.168.39.248:8443 was refused - did you specify the right host or port?
error executing jsonpath "{.items[0].status.state}": Error executing template: array index out of bounds: index 0, length 0. Printing more information for debugging the template:
	template was:
		{.items[0].status.state}
	object given to jsonpath engine was:
		map[string]interface {}{"apiVersion":"v1", "items":[]interface {}{}, "kind":"List", "metadata":map[string]interface {}{"resourceVersion":"", "selfLink":""}}
```

More logging when deploying Rook is needed before this problem can be addressed.

Madhu-1:
>Warning Failed 24s kubelet Error: ImagePullBackOff
Normal Pulling 10s (x2 over 25s) kubelet Pulling image "docker.io/ceph/ceph:v14.2.12
Warning Failed 10s (x2 over 25s) kubelet Failed to pull image "docker.io/ceph/ceph:v14.2.12": rpc error: code = Unknown desc = Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
Warning Failed 10s (x2 over 25s) kubelet Error: ErrImagePull

Hitting the rook deployment issue at https://github.com/ceph/ceph-csi/pull/1924

Rakshith-R:
[...]

PR #1668, opened 2020-11-05
cephfs: ignore quota when SubVolumeInfo() returns Infinite

nixpanic:
There is a type-check on BytesQuota after calling SubVolumeInfo() to see
if the value is supported. In case no quota is configured, the value
Infinite is returned. This can not be converted to an int64, so the
original code returned an error.

It seems that attaching/mounting sometimes fails with the following
error:

    FailedMount: MountVolume.MountDevice failed for volume "pvc-0e8fdd18-873b-4420-bd27-fa6c02a49496" : rpc error: code = Internal desc = subvolume csi-vol-0d68d71a-1f5f-11eb-96d2-0242ac110012 has unsupported quota: infinite

By ignoring the quota of Infinite, and not setting a quota in the
Subvolume object, this problem should not happen again.

A (temporary) log of a failed run can be found at [2020-11-05 12:04:25 +0000 GMT](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e_k8s-1.18/runs/66/nodes/67/steps/70/log/?start=0).

Fixes: 6a46b8f17f3 "cephfs: implement getSubVolumeInfo() with go-ceph"

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/cephfs/volume.go
+++ b/internal/cephfs/volume.go
@@ -104,16 +104,20 @@ func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, volID volumeID) (
 		return nil, err
 	}
 
+	subvol := Subvolume{
+		// only set BytesQuota when it is of type ByteCount
+		Path:     info.Path,
+		Features: make([]string, len(info.Features)),
+	}
 	bc, ok := info.BytesQuota.(fsAdmin.ByteCount)
 	if !ok {
-		// info.BytesQuota == Infinite
-		return nil, fmt.Errorf("subvolume %s has unsupported quota: %v", string(volID), info.BytesQuota)
-	}
-
-	subvol := Subvolume{
-		BytesQuota: int64(bc),
-		Path:       info.Path,
-		Features:   make([]string, len(info.Features)),
+		// we ignore info.BytesQuota == Infinite and just continue
+		// without returning quota information
+		if info.BytesQuota != fsAdmin.Infinite {
+			return nil, fmt.Errorf("subvolume %s has unsupported quota: %v", string(volID), info.BytesQuota)
+		}
+	} else {
+		subvol.BytesQuota = int64(bc)
 	}
 	for i, feature := range info.Features {
 		subvol.Features[i] = string(feature)


nixpanic:
#1644 failed with the error in the initial description. But there is an [other CI job failure](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.19/runs/66) there as well:
```
E1105 12:44:31.404610       1 controller.go:1445] delete "pvc-aec82450-d69d-405d-88e3-dac2705dcf87": volume deletion failed: rpc error: code = Internal desc = subvolume csi-vol-90c415a4-1f64-11eb-bcc6-0242ac110012 has unsupported quota: infinite
```

nixpanic:
It is not clear to me why the failure is not consistent. Sometimes the CI jobs fail, but in the majority of the cases they pass...

internal/cephfs/volume.go
```diff
@@ -106,14 +106,20 @@ func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, volID volumeID) (
 
 	bc, ok := info.BytesQuota.(fsAdmin.ByteCount)
 	if !ok {
-		// info.BytesQuota == Infinite
-		return nil, fmt.Errorf("subvolume %s has unsupported quota: %v", string(volID), info.BytesQuota)
+		// we ignore info.BytesQuota == Infinite and just continue
+		// without returning quota information
+		if info.BytesQuota == fsAdmin.Infinite {
```
Madhu-1, line 111: the quota will be set to infinite for the kernel <4.17 which doesn't support quotas.

During Snapshot creation operation we get the subvolume info and return the subvolume size as the snapshot size. if we return  `0` when the user tries to create 1Gb PVC, the external provisioner will validate the size as requested size is greater it will allow for the restore operation, I think it should be fine.

nixpanic: uh, indeed! The `==` is intended to be `!=` here :facepalm:

nixpanic:
> /retest ci/centos/mini-e2e-helm/k8s-1.18

Failed Rook deployment, see #1636

ShyamsundarR (approved):

Madhu-1 (approved):

PR #1669, opened 2020-11-05
journal: fix reading omaps from objects with large key counts

phlogistonjohn:

The implementation of getOMapValues assumed that the number of key-value
    pairs assigned to the object would be close to the number of keys
    being requested. When the number of keys on the object exceeded the
    "listExcess" value the function would fail to read additional keys
    even if they existed in the omap.
    This change sets a large fixed "chunk size" value and keeps reading
    key-value pairs as long as the callback gets called and increments
    the numKeys counter.

This should be completely invisible to the callers as no api has changed, but now the call should fetch all the requested keys, if they exist, even when the number of keys in the omap is large and the number of requested keys is small.

## Related issues ##

Fixes: #issue_number

## Future concerns ##

Maybe someday we'll do a better api in go-ceph and be able to simplify some of the ugliness here. Someday.

### DIFF
--- a/internal/journal/omap.go
+++ b/internal/journal/omap.go
@@ -25,9 +25,10 @@ import (
 	"github.com/ceph/go-ceph/rados"
 )
 
-// listExcess is the number of false-positive key-value pairs we will
-// accept from ceph when getting omap values.
-var listExcess = 32
+// chunkSize is the number of key-value pairs that will be fetched in
+// one call. This is set fairly large to avoid calling into ceph APIs
+// over and over.
+const chunkSize int64 = 512
 
 func getOMapValues(
 	ctx context.Context,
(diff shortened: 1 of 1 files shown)


internal/journal/omap.go
```diff
@@ -51,15 +52,26 @@ func getOMapValues(
...
+			oid, startAfter, prefix, chunkSize,
+			func(key string, value []byte) {
+				numKeys++
+				startAfter = key
+				if want[key] {
+					results[key] = string(value)
+				}
+			},
+		)
+		// if we hit an error. or no new keys were seen, exit the loop
```
nixpanic, line 69: A `.` instead of a `,` in the comment.

nixpanic (approved):
Looks good to me.

The CI complains though, I didn't look why.

phlogistonjohn:
This CI check makes no sense to me:
>
> internal/journal/omap.go:31:25: mnd: Magic number: 512, in <argument> detected (gomnd)
>
> const chunkSize = int64(512)
>
>                        ^

Yes, 512 elements is fairly arbitrary, but how else am I supposed to express an arbitrary value?

Madhu-1 (approved):

PR #1675, opened 2020-11-09
rbd: add support for thick provision

Madhu-1:
Feature to enable thick-provisioning of RBD images:

- [x] add `thickProvision: "true"` to the StorageClass options
- [x] allocate the RBD image by writing zeros to it with go-ceph `WriteSame()`
- [x] e2e test verifying the image is allocated
- [x] allocate extents when expanding an RBD image


Madhu-1:
cc @nixpanic

nixpanic:
An option to enable thick provisioning would be good. However the implementation should probably call `go-ceph.rbd.Image.WriteSame()` (ceph/go-ceph#402) or something like that to fill the image. That is also what the rbd CLI does when passing `--thick-provision`.

examples/rbd/storageclass.yaml
```diff
@@ -29,6 +29,8 @@ parameters:
    # eg: pool: rbdpool
    pool: <rbd-pool-name>
 
+   # Set thickprovision to true if you want thick rbd image.
+   thickprovision: "true"
```
dillaman, line 33:
```suggestion
   thickProvision: "true"
```

Madhu-1: @dillaman I still need to add E2E testing for this new feature. this PR is created to measure the performance

mykaul: Will be happy to see the performance. Do we know what is the optimal block size to pass to Ceph?

Madhu-1: @mykaul  it's currently using default values we are not passing block-size parameter here. @dillaman any suggestion on settting block size? do we need to make it as a parameter is storageclass? or in cephcsi, we can calculate the required block size?

nixpanic: @dillaman would writing the size of the stripe-unit be efficient? Or is there a more appropriate size that can be used with `rbd_writesame()`?

dillaman: Actually, I think using the stripe-unit would be the most inefficient. We use the stripe period in the `rbd` CLI when thick provisioning, which is just the `object-size` multiplied by the `stripe-count`.

nixpanic: oh, wow :roll_eyes: thanks for the clarification!

dillaman: We should probably clarify that I wouldn't realistically expect much (if any) performance difference between thick and thin-provisioned images. The rationale is that bluestore will double-write on small IOs (< 4KiB) and it will perform redirect-on-write for all larger IOs. The benefit for thick-provisioning would be for better space accounting in your cluster when (potentially) overprovisioning.

nixpanic: _better space accounting_ is what the most important change for users is expected to be. We have had reports where overprovisioning caused issues when the cluster ran out of space.

The updated implementation uses go-ceph `WriteSame()` with a `StripeUnit * StripeCount` block-size of zeros. If that is not correct, please let me know.

Thanks!

dillaman: @nixpanic Substitute `StripeCount` for `ObjectSize` (defaults to 4MiB / 1 << `Order`)

nixpanic: @dillaman is there a rbd API that gets the ObjectSize from a created image? Or is it part of the pool? Just wondering how to get the value in case it has been reconfigured somewhere else...

dillaman: @nixpanic Yup, it's in `rbd_image_info_t` exposed via `rbd_stat`

nixpanic: Thanks, that helps. 

> We use the stripe period in the rbd CLI when thick provisioning, which is just the object-size multiplied by the stripe-count.

:point_up: is what I implemented now too.

internal/rbd/rbd_util.go
```diff
@@ -289,6 +297,37 @@ func (rv *rbdVolume) open() (*librbd.Image, error) {
...
+		return err
+	}
+
+	// zeroBlock is the stripe-period: size of the object-size multiplied
+	// by the stripe-count
+	zeroBlock := make([]byte, sc*(1<<st.Order))
+
+	// the actual size of the image as available in the pool, can be
+	// marginally different from the requested image size
+	size := st.Size
```
Madhu-1, line 325: How the size can be different from the requested size here?

examples/rbd/storageclass.yaml
```diff
@@ -29,6 +29,9 @@ parameters:
    # eg: pool: rbdpool
    pool: <rbd-pool-name>
 
+   # Set thickProvision to true if you want RBD images to be fully allocated on
+   # creation (thin provisioning is the default).
+   thickProvision: "false"
```
Madhu-1, line 34: Need to document this at https://github.com/ceph/ceph-csi/blob/master/docs/deploy-rbd.md

internal/rbd/rbd_util.go
```diff
@@ -229,6 +230,13 @@ func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 		return fmt.Errorf("failed to create rbd image: %w", err)
 	}
 
+	if pOpts.ThickProvision {
+		err = pOpts.allocate()
```
Madhu-1, line 234: if the allocation fails here we are not cleaning up the volume but we are returning the error, as the error is returned the entry in the OMAP will be deleted, and then this can leave stale rbd images.

internal/rbd/rbd_util.go
nixpanic: the requested size gets rounded (up) on the mega bytes, so rv.Size (in MB) is different from the size of the RBD image (in bytes)

examples/rbd/storageclass.yaml
nixpanic: ok, will add that!

internal/rbd/rbd_util.go
nixpanic: good catch, will look into that too

nixpanic:
The added e2e test fails to detect allocated extents with `rbd diff` (returns an empty string), however bytes should have been written:
```
I0107 16:05:49.839780       1 rbd_util.go:306] ID: 20 Req-ID: pvc-0057da0f-9e5b-4c23-a3c0-f493948496c3 going to allocate "replicapool/csi-vol-35bb2b09-5102-11eb-87b2-0242ac110013" with 1073741824 bytes, this may take time
I0107 16:05:49.871160       1 rbd_util.go:333] ID: 20 Req-ID: pvc-0057da0f-9e5b-4c23-a3c0-f493948496c3 wrote 1073741824 bytes to "replicapool/csi-vol-35bb2b09-5102-11eb-87b2-0242ac110013"
```

Not trying to use `rados.OpFlagFAdviseSequential` and `rados.OpFlagFAdviseNoCache` to optimize writing a little more.

If this is not sufficient, I'll add `image.Flush()` as well.

nixpanic:
> Now trying to use `rados.OpFlagFAdviseSequential` and `rados.OpFlagFAdviseNoCache` to optimize writing a little more.
> 
> If this is not sufficient, I'll add `image.Flush()` as well.

Flags and `Flush()` do not influence the behaviour, but there is the `rbd_discard_on_zeroed_write_same` Ceph configuration setting that needs disabling.

nixpanic (dismissed):
@Madhu-1 this is ready from my point of view, please have a look.

e2e/rbd_helper.go
```diff
@@ -314,6 +314,38 @@ func deleteBackingRBDImage(f *framework.Framework, pvc *v1.PersistentVolumeClaim
 	return err
 }
 
+func backingRBDImageHasExtents(f *framework.Framework, pvc *v1.PersistentVolumeClaim) (bool, error) {
```
Madhu-1, line 317: what are we validating here? just running `rbd diff` but not checking anything from the command output?

e2e/rbd_helper.go
```diff
@@ -314,6 +314,38 @@ func deleteBackingRBDImage(f *framework.Framework, pvc *v1.PersistentVolumeClaim
...
+		return false, fmt.Errorf("command %q returned error output: %s", cmd, stderr)
+	}
+	if stdout == "" {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func sparifyBackingRBDImage(f *framework.Framework, pvc *v1.PersistentVolumeClaim) error {
```
Madhu-1, line 338: are we just making sure we are able to run sparsify on an image. and also is this command available on all ceph versions?

internal/rbd/rbd_util.go
```diff
@@ -229,6 +230,16 @@ func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 		return fmt.Errorf("failed to create rbd image: %w", err)
 	}
 
+	if pOpts.ThickProvision {
+		err = pOpts.allocate(ctx)
+		if err != nil {
+			// nolint:errcheck // deleteImage() will log errors in
+			// case it fails, no need to log them here again
+			_ = deleteImage(ctx, pOpts, cr)
```
Madhu-1, line 238: can we log an error if delete fails? because if the allocation fails, the image deletion might fail but we will remove the omap entry still this can leave stale resources.

internal/rbd/rbd_util.go
```diff
@@ -229,6 +230,16 @@ func createImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 		return fmt.Errorf("failed to create rbd image: %w", err)
 	}
 
+	if pOpts.ThickProvision {
+		err = pOpts.allocate(ctx)
```
Madhu-1, line 234: Any rough figures on how much allocation time different size rbd image can take?

internal/rbd/rbd_util.go
```diff
@@ -289,6 +300,50 @@ func (rv *rbdVolume) open() (*librbd.Image, error) {
...
+
+	// zeroBlock is the stripe-period: size of the object-size multiplied
+	// by the stripe-count
+	zeroBlock := make([]byte, sc*(1<<st.Order))
+
+	// the actual size of the image as available in the pool, can be
+	// marginally different from the requested image size
+	size := st.Size
+
+	s, err := image.WriteSame(0, size, zeroBlock, rados.OpFlagNone)
```
Madhu-1, line 341: just use st.Size without creating a new variable?

e2e/rbd.go
```diff
@@ -1297,6 +1297,67 @@ var _ = Describe("RBD", func() {
...
+				if err != nil {
+					e2elog.Failf("failed to load PVC with error %v", err)
+				}
+				pvc.Namespace = f.UniqueName
+
+				err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create PVC with error %v", err)
+				}
+				validateRBDImageCount(f, 1)
```
Madhu-1, line 1321: once we create the pvc can we validate the storage consumption size in the ceph cluster? is that what we are doing with `rbd diff` if yes can we compare the allocated size with the expected size.

Madhu-1 (commented):
@nixpanic I left few questions, Overall looks good. Thanks for taking this to completion.

e2e/rbd_helper.go
nixpanic: `rbd diff` returns no output in case the image does not contain any data. There will be a line for each extent when something is written (allocated) there.

e2e/rbd_helper.go
nixpanic: yeah, verify that a thick-provisioned image (with zero's as data) can be sparsified. This is not something that would be done on a deployment, but it validates that the image only contains zero's.

internal/rbd/rbd_util.go
nixpanic: sure. will do!

nixpanic: > sure. will do!

Actually, isn't the comment above the not logging of the error sufficient?

internal/rbd/rbd_util.go
nixpanic: Not much. I expect this to be the same performance as the `rbd create --thick-provision` option. It will be slow, as writing data to the image needs to be done. @gobindadas might be able to share some statistics from that.

internal/rbd/rbd_util.go
nixpanic: sure, I can clean that up

e2e/rbd_helper.go
nixpanic: Added comment to explain the need/usage of the function.

e2e/rbd_helper.go
nixpanic: Added comment to explain the need/usage of the function.

e2e/rbd.go
nixpanic: I think you can only check the storage usage of an RBD image with `rbd diff`. It is possible to count the size of all the extents together and verify that the size of the image matches. But, this goes a little far, I think, it would be sufficient to validate that _some extents_ have been allocated.

internal/rbd/rbd_util.go
Madhu-1: one last thing we need to check is that, when the user request for pvc with a greater size say 100 GB. as this is going to take time, meanwhile what happens if the user deletes the pvc. will that leave any stale PV or rbd image?

e2e/rbd.go
Madhu-1: checking `ceph df --format=json` for the pool help to identify checking the used size before and after creating the image?

e2e/rbd_helper.go
Madhu-1: `rbd sparsify` is available in only the newer ceph version right?

internal/rbd/rbd_util.go
Madhu-1: yeah, sound like we need to improve error logging. not to log in internal functions.

e2e/rbd.go
nixpanic: That assumes there are no other things happening on the pool, I do not think we can assume exclusive access to it.

e2e/rbd_helper.go
nixpanic: It was made available with Nautilus. That is the oldest Ceph release that is still maintained.

e2e/rbd.go
Madhu-1: or even `rbd du` command helps?

nixpanic:
Just added this to the description, it is still missing in the current PR:

- [ ] https://github.com/Mergifyio/mergify-engine/issues/1942

Madhu-1:
@mergifyio rebase

nixpanic (dismissed):
@Madhu-1 could you review (use `ready-to-merge`) this too please?

e2e/rbd_helper.go
```diff
@@ -314,6 +314,46 @@ func deleteBackingRBDImage(f *framework.Framework, pvc *v1.PersistentVolumeClaim
...
+	}
+
+	return true, nil
+}
+
+// sparifyBackingRBDImage runs `rbd sparsify` on the RBD image. Once done, all
+// data blocks that contain zeros are discarded/trimmed/unmapped and do not
+// take up any space anymore. This can be used to verify that an empty, but
+// allocated (with zerofill) extents have been released.
+func sparifyBackingRBDImage(f *framework.Framework, pvc *v1.PersistentVolumeClaim) error {
```
Madhu-1, line 346: `sparifyBackingRBDImage` to `sparsifyBackingRBDImage`

internal/rbd/rbd_util.go
```diff
@@ -1165,13 +1225,42 @@ func (rv *rbdVolume) resize(newSize int64) error {
...
+		offset = st.Size
+	}
+
 	err = image.Resize(uint64(util.RoundOffVolSize(newSize) * helpers.MiB))
 	if err != nil {
 		return err
 	}
 
+	if thick {
+		err = rv.allocate(offset)
```
Madhu-1, line 1251: for expansion of thick image, do we need to first resize the image, allocate and again resize it?

internal/rbd/rbd_util.go
```diff
@@ -1165,13 +1225,42 @@ func (rv *rbdVolume) resize(newSize int64) error {
...
+		if err != nil {
+			resizeErr := image.Resize(uint64(util.RoundOffVolSize(newSize) * helpers.MiB))
+			if resizeErr != nil {
+				err = fmt.Errorf("failed to shrink image (%v) after failed allocation: %w", resizeErr, err)
```
Madhu-1, line 1255: failed to shrink the image? as we don't support shrinking do we need to update it?

e2e/rbd.go
```diff
@@ -1297,6 +1297,67 @@ var _ = Describe("RBD", func() {
...
+				}
+
+				// after sparsifying the image should not have any allocated extents
+				extents, err = backingRBDImageHasExtents(f, pvc)
+				if err != nil {
+					e2elog.Failf("backing RBD image is not thick-provisioned: %v", err)
+				} else if extents {
+					e2elog.Failf("backing RBD image was not sparsified")
+				}
+
```
Madhu-1, line 1344: better to have an E2E for expansion also?

internal/rbd/rbd_util.go
nixpanic: the steps are:

1. resize rbd image (by the provisioner)
2. allocate new extension (by the provisioner)
3. resize the filesystem (on the node)

In case step 2 fails, resize it again to the previous size, so that there is an error returned on the resize, without having additional allocations.

internal/rbd/rbd_util.go
nixpanic: If writing to the RBD image failed, resizing is not complete. So in this case, the additional zeros that have been written (partially) can be chopped off again. There is no filesystem data on the new extension, so reducing the size is safe.

internal/rbd/rbd_util.go
Madhu-1: currently, it does resize to the new image, and it's allocated to the old size. and if the allocation fails it again doing resize to the new size.

I was thinking it should be as follows

* Resize to the new size
* Get the image stat and allocate it to the new size
* If allocation fails shrink to the old size.

please correct me if am wrong here.

anmolsachan:
Would there be any impact on the Kubelet (kubelet_volume_stats_used_bytes) by this? I hope the metric won't report it completely used since the creation of the volume.

nixpanic:
> Would there be any impact on the Kubelet (kubelet_volume_stats_used_bytes) by this? I hope the metric won't report it completely used since the creation of the volume.

No, as long as it is a FilesystemMode PVC, the metrics are gathered from the filesystem on top of the device, not from the underlying block device. For BlockMode PVCs the consumption would be set to 100% as the contents can not be inspected.

e2e/rbd.go
```diff
@@ -1297,6 +1297,67 @@ var _ = Describe("RBD", func() {
 				validateRBDImageCount(f, 0)
 			})
 
+			By("create a thick-provisioned PVC", func() {
+				err := deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				err = createRBDStorageClass(f.ClientSet, f, nil, map[string]string{
```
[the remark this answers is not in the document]
nixpanic: This is how things are done for each modification in the e2e tests. Not sure if all Kubernetes versions allow you to patch StorageClasses either...

e2e/rbd.go
nixpanic: I'll use `rbd du` for the final version of this PR.

internal/rbd/rbd_attach.go
```diff
@@ -248,7 +253,13 @@ func createPath(ctx context.Context, volOpt *rbdVolume, cr *util.Credentials) (s
 		isNbd = true
 	}
 
-	mapArgs = appendDeviceTypeAndOptions(mapArgs, isNbd, volOpt.MapOptions)
+	// check if the image should stay thick-provisioned
+	isThick, err := volOpt.isThickProvisioned()
+	if err != nil {
+		util.WarningLog(ctx, "failed to detect if image %q is thick-provisioned: %v", volOpt.String(), err)
+	}
```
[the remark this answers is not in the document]
nixpanic: If there is a failure getting the metadata of an RBD image, most other operations will not succeed either. I do not think this is a real issue, we probably would not get this far if the RBD image is inaccessible.

Worst case scenario is that it is possible for a component to discard extents of the image. It is unlikely for Filesystem PVCs as `fstrim` is a privileged operation that needs manual execution. `blkdiscard` is probably easier to run, but users would still need to run is intentionally.

internal/rbd/rbd_util.go
nixpanic: I think that is done here? The early `Stat()` call is to get the pre-resize size of the image, so that it can be used as offset for writing zeros for the extension.

e2e/rbd.go
nixpanic: this has been added now too

nixpanic:
@Madhu-1, @humblec this should be pretty much ready to go now. Please check again, thanks!

nixpanic:

```
Feb 13 20:50:43.670: INFO: rbd-80814 app  is in Pending phase expected to be in Running  state (22 seconds elapsed)
Feb 13 20:50:45.676: INFO: ExecWithOptions {Command:[/bin/sh -c rbd ls --format=json --pool=replicapool] Namespace:rook-ceph PodName:rook-ceph-tools-5455675849-l6cfs ContainerName:rook-ceph-tools Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Feb 13 20:50:45.676: INFO: >>> kubeConfig: /root/.kube/config
Feb 13 20:55:58.009: INFO: stdErr occurred: 2021-02-13 20:55:46.581 7fd0a5f06c80  0 monclient(hunting): authenticate timed out after 300

Feb 13 20:55:58.010: FAIL: failed to list rbd images with error failed to list images 2021-02-13 20:55:46.581 7fd0a5f06c80  0 monclient(hunting): authenticate timed out after 300
```

The MON went offline for some reason? [full logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e_k8s-1.20/runs/237/nodes/89/log/)

humblec:
PR looks good to me, retriggering the mini e2e to see previous failure was spurious or not. Will wait for the same.

nixpanic:

Deleting PVCs has been very slow, but I could not identify a real hang in the [logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e_k8s-1.20/runs/241/nodes/85/log/).

nixpanic (approved):
@Madhu-1 @humblec please check the current state, I think it should cover everything that was requested/mentioned.

internal/rbd/rbd_util.go
```diff
@@ -1165,13 +1225,42 @@ func (rv *rbdVolume) resize(newSize int64) error {
...
+
 	err = image.Resize(uint64(util.RoundOffVolSize(newSize) * helpers.MiB))
 	if err != nil {
 		return err
 	}
 
+	if thick {
+		err = rv.allocate(offset)
+		if err != nil {
+			resizeErr := image.Resize(uint64(util.RoundOffVolSize(newSize) * helpers.MiB))
```
Madhu-1, line 1253: @nixpanic the resize as to be done for old size right if allocation fails?

Madhu-1 (commented):
small question, LGTM

internal/rbd/rbd_util.go
nixpanic: Oh, yes, `newSize` is not correct, that should be the old size!

nixpanic:

[job 275](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e_k8s-1.20/runs/275/nodes/85/log/) failed to delete a PVC:
```
Feb 18 08:35:30.919: INFO: waiting for PVC cephfs-80811 in state &PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},} to be deleted (600 seconds elapsed)
Feb 18 08:35:30.925: INFO: waiting for PVC cephfs-80811 in state &PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},} to be deleted (600 seconds elapsed)
Feb 18 08:35:30.930: FAIL: failed to delete PVC or application with error timed out waiting for the condition
```

#1876 has been merged already, it seems the deletion validation needs more improvements.

Madhu-1 (commented):
LGTM

humblec (approved):

humblec:
Adding `ready-to--merge` as we have enough approvals in place.

PR #1678, opened 2020-11-17
rbd: add minsnapshotsonimage flag

Madhu-1:
An rbd image can have a maximum number of snapshots defined by maxsnapshotsonimage On the limit is reached the cephcsi will start flattening the older snapshots and returns the ABORT error message, The Request comes after this as to wait till all the
images are flattened (this will increase the PVC creation time.  Instead of waiting till the maximum snapshots on an RBD image, we can have a soft limit, once the limit reached cephcsi will start flattening the task to break the chain. With this PVC  creation time will only be affected when the hard limit (minsnapshotsonimage) reached.

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -342,10 +342,12 @@ func flattenParentImage(ctx context.Context, rbdVol *rbdVolume, cr *util.Credent
 	return nil
 }
 
-// check snapshots on the rbd image, as we have limit from krbd that
-// an image cannot have more than 510 snapshot at a given point of time.
-// If the snapshots are more than the `maxSnapshotsOnImage` Add a task to
-// flatten all the temporary cloned images.
+// check snapshots on the rbd image, as we have limit from krbd that an image
+// cannot have more than 510 snapshot at a given point of time. If the
+// snapshots are more than the `maxSnapshotsOnImage` Add a task to flatten all
+// the temporary cloned images and return ABORT error message. If the snapshots
+// are more than the `minSnapshotOnImage` Add a task to flatten all the
+// temporary cloned images.
 func flattenTemporaryClonedImages(ctx context.Context, rbdVol *rbdVolume, cr *util.Credentials) error {
 	snaps, err := rbdVol.listSnapshots()
 	if err != nil {
@@ -356,6 +358,7 @@ func flattenTemporaryClonedImages(ctx context.Context, rbdVol *rbdVolume, cr *ut
 	}
 
 	if len(snaps) > int(maxSnapshotsOnImage) {
+		util.DebugLog(ctx, "snapshots count %d on image: %s reached configured hard limit %d", len(snaps), rbdVol, maxSnapshotsOnImage)
 		err = flattenClonedRbdImages(
 			ctx,
 			snaps,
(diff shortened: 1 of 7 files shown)


Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio rebase

internal/rbd/controllerserver.go
```diff
@@ -368,6 +371,20 @@ func flattenTemporaryClonedImages(ctx context.Context, rbdVol *rbdVolume, cr *ut
 		}
 		return status.Errorf(codes.ResourceExhausted, "rbd image %s has %d snapshots", rbdVol, len(snaps))
 	}
+
+	if len(snaps) > int(minSnapshotsOnImageToStartFlatten) {
```
ShyamsundarR, line 375: The soft limit should flatten to only reduce the outstanding count by a value that lowers it below the soft limit, and not all snapshots till the soft limit.

In this case, about 250 snapshots would be flattened, which would take time. The time taken would be lower than the 450 max limit, but still the operation would not be fast enough.

Instead, if we flatten only as many snapshots as required to go below the soft limit, then the repetitive operation cost is amortized across future clones, and each clone operation hence would only be slowed down as much.

So I would suggest we either pass the delta count to `flattenClonedRbdImages` and only flatten as many, or pass in `snaps` list with only a count of snaps that require flattening.

Madhu-1: updated logic to flatten only a minimum number of snapshots once the limit is reached so that the volume creation time should not be affected heavily

ShyamsundarR (dismissed):

scripts/install-helm.sh
```diff
@@ -125,7 +125,7 @@ install_cephcsi_helm_charts() {
...
     # issue when installing ceph-csi-rbd
     kubectl delete cm ceph-csi-config --namespace ${NAMESPACE}
-    "${HELM}" install --namespace ${NAMESPACE} --set provisioner.fullnameOverride=csi-rbdplugin-provisioner --set nodeplugin.fullnameOverride=csi-rbdplugin --set configMapName=ceph-csi-config --set provisioner.podSecurityPolicy.enabled=true --set nodeplugin.podSecurityPolicy.enabled=true --set provisioner.replicaCount=1 ${RBD_CHART_NAME} "${SCRIPT_DIR}"/../charts/ceph-csi-rbd --set topology.enabled=true --set topology.domainLabels="{${NODE_LABEL_REGION},${NODE_LABEL_ZONE}}" --set provisioner.maxSnapshotsOnImage=3
+    "${HELM}" install --namespace ${NAMESPACE} --set provisioner.fullnameOverride=csi-rbdplugin-provisioner --set nodeplugin.fullnameOverride=csi-rbdplugin --set configMapName=ceph-csi-config --set provisioner.podSecurityPolicy.enabled=true --set nodeplugin.podSecurityPolicy.enabled=true --set provisioner.replicaCount=1 ${RBD_CHART_NAME} "${SCRIPT_DIR}"/../charts/ceph-csi-rbd --set topology.enabled=true --set topology.domainLabels="{${NODE_LABEL_REGION},${NODE_LABEL_ZONE}}" --set provisioner.maxSnapshotsOnImage=3 --set provisioner.minSnapshotsOnImage=1
```
nixpanic, line 129: I would recommend setting `minSnapshotsOnImage` to 2, as it would otherwise always hit, and the non-flattening path never gets tested. At least, I think so?

Madhu-1: increased the value to 2.

nixpanic (approved):

ShyamsundarR (approved):

Madhu-1:
@mergifyio rebase

Issue #1645, opened 2020-10-29
Use external library for KMS integration


Madhu-1:
currently, we are using the  HTTP package for the KMS integration. https://github.com/libopenstorage/secrets is an opensource library to support different KMS plugins. I think it's a good idea to use a library to reuse the existing library instead of developing our own package.

cc @leseb

Madhu-1:
@humblec @nixpanic any thoughts?

leseb:
FYI Rook already uses it in its KMS integration here: https://github.com/rook/rook/pull/6474.
The lib is well maintained and they easily accept contributions too.

nixpanic:
> @humblec @nixpanic any thoughts?

Yes, I'm much in favour of this!

PR #1694, opened 2020-11-17
rebase: use libopenstorage/secrets API to access Vault

nixpanic:

This replaces the manual written API access to Vault with the API provided by libopenstorage/secrets. Rook uses the same package to communicate with Vault. Using a maintained package reduces the maintenance burden on Ceph-CSI.

The current feature that uses Vault for secrets should ideally not be broken. Users that consume the current state of the support should transparently be moved to the usage of the new API. This means that the old configuration must remain valid.

## Related issues ##

Fixes: #1645 

---

<details>

  progress or results

</details>


nixpanic:
This PR is marked with `[DNM]` and in **draft** state. There are a few `TODO` and `FIXME` notes that need to be addressed before this is complete.

/cc @leseb

internal/util/vault.go
```diff
@@ -208,130 +211,10 @@ func (kms *VaultKMS) SavePassphrase(key, value string) error {
 
 // DeletePassphrase deletes passphrase from Vault.
 func (kms *VaultKMS) DeletePassphrase(key string) error {
-	vaultToken, err := kms.getAccessToken()
+	err := kms.secrets.DeleteSecret(filepath.Join(kms.vaultPassphrasePath, key), kms.keyContext)
```
Madhu-1, line 214: When `filepath.Join(kms.vaultPassphrasePath, key)` file will be deleted?

Madhu-1 (commented):
can we also have an E2E with vault token-based auth and with certificates later(in next PR's)

leseb (commented):
Looks pretty clean so far :) thanks!

internal/util/vault.go
```diff
@@ -17,15 +17,15 @@ limitations under the License.
...
 	"fmt"
-	"io"
-	"io/ioutil"
-	"net/http"
+	"path/filepath"
 	"strconv"
-	"strings"
+
+	"github.com/hashicorp/vault/api"
+	losSecrets "github.com/libopenstorage/secrets"
```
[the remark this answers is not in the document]
nixpanic: Sure, can do, but it becomes `loss` :-D

internal/util/vault.go
nixpanic: Unsure what you mean? This deletes the secret that is identified with the `filepath.Join(kms.vaultPassphrasePath, key)`. It is just the API replacement of doing the manual HTTP DELETE call.

nixpanic:
Oh no, build errors due to updating external-snapshotter in `vendor/`. This update was not intentional. For some reason local builds (and on manual test system) do not have this problem :thinking:

internal/util/vault.go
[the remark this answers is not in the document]
nixpanic: This was modified to `loss`

internal/util/crypto.go
```diff
@@ -146,8 +146,8 @@ func GetCryptoPassphrase(ctx context.Context, volumeID string, kms EncryptionKMS
 		return passphrase, nil
 	}
 	if _, ok := err.(MissingPassphrase); ok {
-		DebugLog(ctx, "Encryption passphrase is missing for %s. Generating a new one",
-			volumeID)
+		DebugLog(ctx, "Encryption passphrase is missing for %s (%s). Generating a new one",
```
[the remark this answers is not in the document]
nixpanic: This change has been dropped.

internal/util/vault.go
nixpanic: Ok, it seems that the secret gets deleted, but is not destroyed by Vault (it becomes empty). Not sure when the deleted secrets get completely removed though. For this, I have updated the e2e test that checks the contents of the secret.

nixpanic:
This is pretty much complete now. Dropping `[DNM]` and **draft** state. More serious reviews welcome :sparkles:

nixpanic:
Many CI jobs fail while deploying rook (maybe when pulling the image from Docker Hub). When #1712 is merged, this PR can be rebased and hopefully the container image gets pulled from the local CI registry instead.

internal/util/vault.go
```diff
@@ -65,73 +62,110 @@ Example JSON structure in the KMS config is,
...
+	// secret (like /v1/secret/data/<VPP>/key)
+	vaultPassphrasePath string
+
+	secrets    loss.Secrets
+	keyContext map[string]string
 }
 
 // InitVaultKMS returns an interface to HashiCorp Vault KMS.
+//
+// nolint:gocyclo // this si a long function, as it constructs the Vault config
```
Madhu-1, line 77: `%s/si/is`

internal/util/vault.go
```diff
@@ -65,73 +62,110 @@ Example JSON structure in the KMS config is,
...
+		kms.vaultPassphrasePath = vaultDefaultPassphrasePath
+	}
+
 	verifyCA, ok := config["vaultCAVerify"]
 	if ok {
-		kms.VaultCAVerify, err = strconv.ParseBool(verifyCA)
+		var vaultCAVerify bool
+		vaultCAVerify, err = strconv.ParseBool(verifyCA)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse 'vaultCAVerify' for vault <%s> kms config: %s",
```
Madhu-1, line 139: changes not made in this PR can we use `%w` for err instead of `%s`?

internal/util/vault.go
```diff
@@ -65,73 +62,110 @@ Example JSON structure in the KMS config is,
...
-				vaultCAFromSecret)
+		vaultConfig[api.EnvVaultCACert], err = createTempFile("vault-ca-cert", []byte(caPEM))
+		if err != nil {
+			return nil, fmt.Errorf("failed to create temporary file for Vault CA: %w", err)
 		}
-		kms.vaultCA = roots
+		// TODO: delete f.Name() when VaultKMS is destroyed
+	}
+
+	// FIXME: vault.AuthKubernetesTokenPath is not enough? EnvVaultToken needs to be set?
```
Madhu-1, line 158: FixMe is still needed?

internal/util/vault.go
```diff
@@ -184,21 +199,13 @@ func (kms *VaultKMS) GetPassphrase(key string) (string, error) {
 
 // SavePassphrase saves new passphrase in Vault.
 func (kms *VaultKMS) SavePassphrase(key, value string) error {
-	data, err := json.Marshal(map[string]map[string]string{
-		"data": {
+	data := map[string]interface{}{
+		"data": map[string]string{
 			"passphrase": value,
```
Madhu-1, line 204: TODO as an enhancement define `passphrase` as a string variable and use it in all functions?

internal/util/vault.go
```diff
@@ -184,21 +199,13 @@ func (kms *VaultKMS) GetPassphrase(key string) (string, error) {
...
-		return fmt.Errorf("passphrase request data is broken: %s", err)
 	}
 
-	resp, err := kms.request("POST", kms.getKeyDataURI(key), data)
-	if err != nil {
-		return fmt.Errorf("failed to POST passphrase for %s to vault: %s", key, err)
-	}
-	defer resp.Body.Close()
-	err = kms.processError(resp, "save passphrase")
+	err := kms.secrets.PutSecret(filepath.Join(kms.vaultPassphrasePath, key), data, kms.keyContext)
```
Madhu-1, line 208: just return an error? or do logging? currently, only error check is done

internal/util/vault_test.go
```diff
@@ -0,0 +1,56 @@
+/*
+Copyright 2019 The Ceph-CSI Authors.
```
Madhu-1, line 2: 2020?

internal/util/vault.go
```diff
@@ -208,130 +215,74 @@ func (kms *VaultKMS) SavePassphrase(key, value string) error {
...
-	}
-	if kms.vaultCA != nil {
-		tlsConfig.RootCAs = kms.vaultCA
-	}
-	netTransport := &http.Transport{TLSClientConfig: tlsConfig}
-	client := &http.Client{Transport: netTransport}
+	// delete the tmpfile on error
+	defer func() {
+		if err != nil {
+			// ignore error on failure to remove tmpefile (gosec complains)
```
Madhu-1, line 273: tmpefile to tmpfile

Madhu-1 (changes requested):
changes look good, small Nit

Madhu-1:
@nixpanic can you please check does these changes cause any backward compatibility issues?

internal/util/vault.go
nixpanic: ok!

internal/util/vault.go
nixpanic: sure, will do

internal/util/vault.go
nixpanic: Well, 'need' maybe not, but it feels like the configuration is done twice, one option might be sufficient. Will keep it as is, it works for now (any likely needs updating when we want to support multiple Vault tokens instead of the ServiceAccount).

internal/util/vault.go
nixpanic: It is only used 2x, the same as `"data"` (for get/save). I do not think it makes things easier or clearer when there is a variable for these.

internal/util/vault.go
nixpanic: oh, yes, just returning is sufficient, I think

internal/util/vault_test.go
nixpanic: done

internal/util/vault.go
nixpanic: done

nixpanic:
> @nixpanic can you please check does these changes cause any backward compatibility issues?

I do not expect any issues. The configuration options are kept the same, and should get converted to the values the API needs. During development, I have swapped out the passphrase fetching/storing with the original implementation, and with that, things keep working (and after image update/downgrade).

nixpanic:
Hmm, deploying Rook fails often now... #1716 might help in troubleshooting the issues

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio rebase

Madhu-1 (commented):
LGTM just waiting for CI to pass

nixpanic:
@Mergifyio rebase

nixpanic:
All of the `/retest` commands were needed because Rook failed to deploy. Logs are non-conclusive, #1719 has been created to gather more logs. Not all tests fail always, so `/retest` may be sufficient and get the approvals for merging. Work on making the CI more stable is ongoing in any case.

nixpanic:
@mergifyio rebase

nixpanic:
@Madhu-1 @leseb all changes have been done, and CI now passed (no Rook deployment failures....)

Madhu-1 (approved):

Madhu-1 (approved):

tareksha:
hi @nixpanic this change caused ceph-csi to work with KV-v2 secret mounts only :(
https://github.com/libopenstorage/secrets/blob/a1d19aa9713a820d42fb3743ab8563d192d81ff9/vault/vault.go#L115

leseb:
> hi @nixpanic this change caused ceph-csi to work with KV-v2 secret mounts only :( https://github.com/libopenstorage/secrets/blob/a1d19aa9713a820d42fb3743ab8563d192d81ff9/vault/vault.go#L115

The link tells about the secret backend (kv, transit, etc) not the backend version (v1 or v2). So v1 and v2 are supported.

tareksha:
@leseb Why does ceph-csi care about the details of the underlying secrets engine? it simply does the standard read, write and delete operations against vault. I would like to use a custom plugin, not necessarily KV.
libopenstroage has an as-is support for KV-v1 and KV-v2 only.

leseb:
> @leseb Why does ceph-csi care about the details of the underlying secrets engine? it simply does the standard read, write and delete operations against vault. I would like to use a custom plugin, not necessarily KV. libopenstroage has an as-is support for KV-v1 and KV-v2 only.

Not ceph-csi, the libopenstroage does. It does because it needs to know which API path to query when interacting with the KV Secret Engine. For more details, you can see the [API description](https://www.vaultproject.io/api-docs/secret/kv).
I don't think ceph-csi supports any other Secret Engine than KV when Vault is used. But I'll let @nixpanic confirm that.

tareksha:
@leseb the consumers are not aware of details like the libraries that are used by the internal implementation (neither should they, i guess). what they experience eventually is that ceph-csi works with KV engine only.

leseb:
> @leseb the consumers are not aware of details like the libraries that are used by the internal implementation (neither should they, i guess). what they experience eventually is that ceph-csi works with KV engine only.

I've never said the opposite but that's how it is today, ceph-csi only supports the Vault kv secret engine (v1 and v2). So I'm not sure what point you are trying to make here... Initially, you mentioned only vault secret engine v2 being supported and I can tell you this is not the case.

PR #1696, opened 2020-11-18
e2e: set imagePullPolicy to ifNotPresent

Madhu-1:
If the imagePullPolicy is not set and the image tag is empty or latest the image is always pulled. This commit sets the policy to pull images if not present. So that in E2E the image is pulled only once.

### DIFF
--- a/e2e/pod.go
+++ b/e2e/pod.go
@@ -156,6 +156,9 @@ func loadApp(path string) (*v1.Pod, error) {
 	if err != nil {
 		return nil, err
 	}
+	for i := range app.Spec.Containers {
+		app.Spec.Containers[i].ImagePullPolicy = v1.PullIfNotPresent
+	}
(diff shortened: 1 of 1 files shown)


Yuggupta27 (approved):
Great enhancement! LGTM

nixpanic (approved):
I think the default `Always` only causes to pull the image once too, but there is an additional check to see if it needs updating. Using `IfNotPresent` will likely not change much, but it certainly won't hurt either :+1:

Madhu-1:
> I think the default `Always` only causes to pull the image once too, but there is an additional check to see if it needs updating. Using `IfNotPresent` will likely not change much, but it certainly won't hurt either

As we create Many test pods in our E2E, Always will pull the image  whenever a new pod is created and we can easily hit the rate limit issue

PR #1716, opened 2020-11-23
ci: gather logs when deploying Rook fails

nixpanic:
Log a few commands that help troubleshooting Rook deployment issues.
This might need to get extended with more commands.

Updates: #1636

---

<details>

  progress or results

</details>

### DIFF
--- a/scripts/rook.sh
+++ b/scripts/rook.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash -E
 
 ROOK_VERSION=${ROOK_VERSION:-"v1.2.7"}
 ROOK_DEPLOY_TIMEOUT=${ROOK_DEPLOY_TIMEOUT:-300}
@@ -7,6 +7,25 @@ ROOK_BLOCK_POOL_NAME=${ROOK_BLOCK_POOL_NAME:-"newrbdpool"}
 ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
 KUBECTL_RETRY=5
 KUBECTL_RETRY_DELAY=10
+
+trap log_errors ERR
+
+# log_errors is called on exit (see 'trap' above) and tries to provide
+# sufficient information to debug deployment problems
+function log_errors() {
+	# enable verbose execution
+	set -x
+	kubectl get nodes
+	kubectl -n rook-ceph get events
+	kubectl -n rook-ceph describe pods
+	kubectl -n rook-ceph logs -l app=rook-ceph-operator
+	kubectl -n rook-ceph get CephClusters -oyaml
+	kubectl -n rook-ceph get CephFilesystems -oyaml
+	kubectl -n rook-ceph get CephBlockPools -oyaml
+
+	# this function should not return, a fatal error was caught!
+	exit 1
+}
 
 rook_version() {
 	echo "${ROOK_VERSION#v}" | cut -d'.' -f"${1}"


scripts/rook.sh
```diff
@@ -8,6 +8,19 @@ ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
 KUBECTL_RETRY=5
 KUBECTL_RETRY_DELAY=10
 
+trap log_errors EXIT
+
+# log_errors is called on exit (see 'trap' above) and tries to provide
+# sufficient information to debug deployment problems
+function log_errors() {
+	set -x
+	kubectl get events
```
nixpanic, line 17: Maybe the namespace `rook-ceph` needs to be added?

scripts/rook.sh
```diff
@@ -8,6 +8,21 @@ ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
...
 KUBECTL_RETRY_DELAY=10
 
+trap log_errors EXIT
+
+# log_errors is called on exit (see 'trap' above) and tries to provide
+# sufficient information to debug deployment problems
+function log_errors() {
+	# enable verbose execution, disable exit on failure
+	set -x +e
+	kubectl get nodes
```
leseb, line 18: Don't you want the operator logs too?

scripts/rook.sh
```diff
@@ -8,6 +8,21 @@ ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
 KUBECTL_RETRY=5
 KUBECTL_RETRY_DELAY=10
 
+trap log_errors EXIT
+
+# log_errors is called on exit (see 'trap' above) and tries to provide
+# sufficient information to debug deployment problems
+function log_errors() {
+	# enable verbose execution, disable exit on failure
+	set -x +e
```
leseb, line 17: Do you really want  `-e`? I'd continue if anything fails in that function since it's meant to help debugging.

leseb: Nevermind! Read too fast!

scripts/rook.sh
nixpanic: Yes, maybe. Currently most issues are probably due to Docker Pull limits, but the operator logs would be helpful in other cases.

scripts/rook.sh
```diff
@@ -8,6 +8,22 @@ ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
 KUBECTL_RETRY=5
 KUBECTL_RETRY_DELAY=10
 
+trap log_errors EXIT
```
leseb, line 11: If you trap on `ERR` instead you don't need to untrap at the end IIRC.

scripts/rook.sh
```diff
@@ -8,6 +8,22 @@ ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
...
+
+# log_errors is called on exit (see 'trap' above) and tries to provide
+# sufficient information to debug deployment problems
+function log_errors() {
+	# enable verbose execution, disable exit on failure
+	set -x +e
+	kubectl get nodes
+	kubectl -n rook-ceph get events
+	kubectl -n rook-ceph get pods --show-labels
```
Madhu-1, line 20: pod describe the output of pods in the rook-ceph namespace also helps

scripts/rook.sh
nixpanic: That looks nice, I'll use that.

scripts/rook.sh
nixpanic: sure, I can do that instead

scripts/rook.sh
```diff
@@ -222,3 +238,6 @@ Available Commands:
 " >&2
 	;;
 esac
+
+# no major error occured, no need to call the logging function on exit
+trap - EXIT
```
leseb, line 243: This is not needed anymore since you trap for `ERR` now.

nixpanic: duh :man_facepalming:

scripts/rook.sh
```diff
@@ -8,6 +8,22 @@ ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
 KUBECTL_RETRY=5
 KUBECTL_RETRY_DELAY=10
 
+trap log_errors ERR
```
leseb, line 11: Now you need to use `errtrace` (`-E`) in addition to `errexit` (`-e`) to properly trap `ERR`.

scripts/rook.sh
```diff
@@ -8,6 +8,22 @@ ROOK_CEPH_CLUSTER_VERSION="v14.2.10"
...
+function log_errors() {
+	# enable verbose execution
+	set -x
+	kubectl get nodes
+	kubectl -n rook-ceph get events
+	kubectl -n rook-ceph describe pods
+	kubectl -n rook-ceph logs -l app=rook-ceph-operator
+	kubectl -n rook-ceph get CephClusters -oyaml
+	kubectl -n rook-ceph get CephFilesystems -oyaml
+	kubectl -n rook-ceph get CephBlockPools -oyaml
```
leseb, line 24: One last thing I guess, don't you want to `exit 1` after showing the logs so the CI fails?

leseb (changes requested):

scripts/rook.sh
nixpanic: The script runs with `bash -e` so when an error is caught, the script exits with the error value. An additional `exit 1` is not needed (also tested locally with a small script).

scripts/rook.sh
nixpanic: Not sure, with a small test script this seems to work just fine?
```bash
#!/bin/bash -e

trap log_exit ERR

function log_exit() {
	set -x
	echo "oh no"
}

false

exit 0
```

leseb: I think it works because `false` is not in a function. If you have a function that fails this won't work see: https://stackoverflow.com/questions/35800082/how-to-trap-err-when-using-set-e-in-bash

nixpanic: Oh, ok! Yes, failing inside a function indeed makes a difference. Will adapt it with `-E` and exit when the logging function gets called (otherwise it runs 2x, it seems).

Madhu-1 (approved):

Madhu-1:
@leseb do you have any comments? CI is waiting for your approval

leseb (approved):

PR #1719, opened 2020-11-24
ci: add more logging during Rook deployment

nixpanic:
It seems that the new log_errors() function does not get triggered when
the script hits `exit 1` conditions in functions. The functions should
return a non-0 value, not cause an exit of the script.

Updates: #1636

---

<details>

  progress or results

</details>

### DIFF
--- a/scripts/rook.sh
+++ b/scripts/rook.sh
@@ -59,6 +59,12 @@ kubectl_retry() {
             ret=1
             break
         fi
+
+	# log stderr and empty the tmpfile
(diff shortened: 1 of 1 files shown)


Madhu-1 (approved):

Madhu-1:
CI is considering onlydoc change we need to fix it.as this is not a doc change

nixpanic:
> CI is considering this as onlydoc change we need to fix it.as this is not a doc change

Indeed! #1720 should fix that.

PR #1720, opened 2020-11-25
ci: do not mark script changes as doc-only

nixpanic:
Changes in scripts will affect the CI jobs as the scripts are used while
deploying minikube, Rook and other components. These changes need
testing, just like anything else.

---

<details>

  progress or results

</details>

### DIFF
--- a/scripts/skip-doc-change.sh
+++ b/scripts/skip-doc-change.sh
@@ -12,15 +12,12 @@ CHANGED_FILES=$(git diff --name-only "${GIT_SINCE}")
 
 skip=0
 #files to be skipped
-declare -a FILES=(^docs/ .md$ ^scripts/ LICENSE .mergify.yml .github .gitignore .commitlintrc.yml)
+declare -a FILES=(^docs/ .md$ LICENSE .mergify.yml .github .gitignore .commitlintrc.yml)
 
 function check_file_present() {
     local file=$1
     for FILE in "${FILES[@]}"; do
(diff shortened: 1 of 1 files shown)


Yuggupta27 (approved):

Madhu-1 (commented):
we dont have to run E2E for changes in files like gosec.sh, golangci.yaml, build-multi-arch etc.

nixpanic:
> we dont have to run E2E for changes in files like gosec.sh, golangci.yaml, build-multi-arch etc.

For most of them we do! If we change scripts or the configuration, we do need to run the tests with the new configuration. Merging a script/configuration that causes all subsequent e2e tests to fail would be an issue. I am of the opinion that more testing is better, than missing something to test and have to urgently fix (or revert) things.

Madhu-1:
> > we dont have to run E2E for changes in files like gosec.sh, golangci.yaml, build-multi-arch etc.
> 
> For most of them we do! If we change scripts or the configuration, we do need to run the tests with the new configuration. Merging a script/configuration that causes all subsequent e2e tests to fail would be an issue. I am of the opinion that more testing is better, than missing something to test and have to urgently fix (or revert) things.

running static check should be fine but what is the use of running E2E if some changes has done in gosec.sh or golangci.yaml?

Madhu-1:
as we have an option to skip E2E. am currently fine with this change, but we need to make the doc-only detecting script more robust.

Madhu-1 (approved):

Issue #1693, opened 2020-11-17
Enable pre-pulling of container images for CI jobs


nixpanic:
# Describe the feature you'd like to have #

Docker Hub introduced image pull rate limits for anonymous and free accounts. The e2e and other tests that run in the CI sometimes hit the limit and fail to pull images. This ultimately causes the CI jobs to fail.

There should be a list of images that come from Docker Hub, and have these images mirrored in the container registry that our CI jobs use.

# How will we know we have a good solution? (acceptance criteria) #

- [ ] the OpenShift environment where Jenkins runs should automatically pull the container images into the local registry
- [x] CI jobs should pull the images from the local registry
- [x] the e2e tests should automatically consume the pre-pulled images
- [x] while deploying Rook images should be pre-pulled in minikube ([see #1724](https://github.com/ceph/ceph-csi/pull/1724#pullrequestreview-539032714 ))

# Additional context #

[mini-e2e.groovy](https://github.com/ceph/ceph-csi/blob/ci/centos/mini-e2e.groovy) already consumes the `ceph/ceph:v15` image, but it was manually pushed into the local registry.

nixpanic:
Images in the CI registry are not automatically updated. This is still a TODO.

nixpanic:
This is all been done. Images are mirrored in the CI environment, and refreshed nightly.

PR #1724, opened 2020-11-25
ci: push ceph/ceph container image into minikube

nixpanic:
Deploying Ceph with Rook fails as the ceph/ceph:v15 base image can not
be pulled from within the minikube VM. By pushing the image into the VM,
but before deploying Rook, there should be no need to pull the image
from Docker Hub anymore.

Updates: #1693

---

<details>

  progress or results

</details>

### DIFF
--- a/single-node-k8s.sh
+++ b/single-node-k8s.sh
@@ -135,6 +135,12 @@ then
     ./podman2minikube.sh "rook/ceph:${ROOK_VERSION}"
 fi
 
+# Rook also uses ceph/ceph:v15 (build.env:BASE_IMAGE), so push it into the VM
+if [ -n "${BASE_IMAGE}" ] && podman inspect "${BASE_IMAGE}" > /dev/null
+then
+    ./podman2minikube.sh "${BASE_IMAGE}"
+fi
+
 deploy_rook
 
 # running e2e.test requires librados and librbd


nixpanic:
@leseb this might address the issue you mentioned:
```
2020-11-25 17:00:44.274967 E | op-cluster: failed the ceph version check. failed to complete ceph version job: failed to run CmdReporter rook-ceph-detect-version successfully. failed waiting for results ConfigMap rook-ceph-detect-version. timed out waiting for results ConfigMap
```

[job with problem](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.19/detail/mini-e2e-helm_k8s-1.19/192/pipeline#step-76-log-3623):
```
  Warning  Failed     13m (x4 over 14m)     kubelet            Failed to pull image "ceph/ceph:v15": rpc error: code = Unknown desc = Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

single-node-k8s.sh
```diff
@@ -135,6 +135,12 @@ then
     ./podman2minikube.sh "rook/ceph:${ROOK_VERSION}"
 fi
 
+# Rook also uses ceph/ceph:v15 (build.env:BASE_IMAGE), so push it into the VM
+if [ -n "${BASE_IMAGE}" ] && podman inspect "${BASE_IMAGE}" > /dev/null
```
Madhu-1, line 139: Rook not expected to use the base image that cephcsi uses. rook uses different tag of ceph images see https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/cluster.yaml#L24

nixpanic: That actually gets replaced in `scripts/rook.sh`. But yes, it seems that uses `ROOK_CEPH_CLUSTER_VERSION` and not `BASE_IMAGE`. Which is a little weird, as the logs mentions pulling `ceph/ceph:v15` failed.

This might need more investigation as of why the wrong image gets configured/pulled, but for now this will work.

Madhu-1 (approved):

single-node-k8s.sh
Madhu-1: let's keep track of it

nixpanic: Added as a checkbox to #1693

PR #1731, opened 2020-11-26
ci: pull images from local registry directly

nixpanic:
The mirror option of the Docker Registry container is very limited and
prevents updating or manually pushing images to the registry. Instead,
it tries to push the images to the docker.io, which is not what we need.

Instead of using a mirror, the CI registry is now populated with
container images that get pulled and tagged as if they get from Docker
Hub or other locations.

This is more of a manual mirror, as the Docker Registry mirror
functionality is not flexible enough for our usecase (push images
without providing them on docker.io).

Updates: #1693

---

<details>

  progress or results

</details>


nixpanic:
Note that the change to the DeploymentConfig is already applied. Unfortunately the rollout of the updated pod fails in OpenShift due to the Docker Hub pull rate-limit...

Madhu-1 (approved):

PR #1743, opened 2020-11-30
doc: multi-tenancy with Vault tokens

nixpanic:
Design for adding a new KMS type "VaultTokens" that can be used to
configure a Hashicorp Vault service where each tenant has their own
personal token to manage encryptions keys for PVCs.

Updates: #1500

---

<details>

  progress or results

</details>

### DIFF
--- a/docs/design/proposals/encryption-with-vault-tokens.md
+++ b/docs/design/proposals/encryption-with-vault-tokens.md
@@ -1,0 +1,187 @@
+# Multi-tenancy with Vault Tokens
+
+## Current Feature: Vault access with single token
+
+The current feature to [support Hashicorp Vault as a KMS is
+documented](./encrypted-pvc.md):
+
+- Tenants can create PVCs that are encrypted with a unique key (Data Encryption
+  Key, DEK)
+- the key to encrypt/decrypt the PVC is stored in Hashicorp Vault
+- there is a single JSON Web Token (JWT), the Kubernetes ServiceAccount, to
+  access Vault
+- enabling is done in the StorageClass, pointing to a KMS configuration section
+  in a JSON configuration file (Kubernetes ConfigMap)
+
+## High-Level Requirements
+
+The new feature should support multi-tenancy, so that each Tenant can use their
+own Vault Token to access the Key Encryption Key (KEK) and fetch the Data
+Encryption Key (DEK) for PVC encryption:
+
+- each tenant can manage their Vault Token
+- to get the KEK from Vault, each Tenant should provide their personal Vault token
+- Ceph-CSI should use the Vault Token from the Tenant to store the unique key
+  (DEK) for the PVC
+- a Tenant is the owner of a Kubernetes Namespace
+
+## Restrictions to consider
+
+- Tenants can only configure their Kubernetes namespace, their token to access
+  Vault should be located in their namespace (as a Kubernetes Secret)
+- Ceph-CSI can not directly access Kubernetes namespaces, as CSI is an
+  abstraction layer
+- Ceph-CSI needs to talk to a service (sidecar) to access the Vault Token from
+  a Tenant
+- the KMS configuration is a ConfigMap with name
+  `ceph-csi-encryption-kms-config` in the namespace where the Ceph-CSI pods are
+  running
+- the KMS configuration is available for the Ceph-CSI pods at
+  `/etc/ceph-csi-encryption-kms-config/config.json`
+- [example of the
+  configuration](https://github.com/ceph/ceph-csi/blob/master/examples/kms/vault/kms-config.yaml)
+
+## Dependencies
+
+- the name of the Kubernetes Secret needs to be known (configured in the KMS
+  config file)
+- each Tenant (Kubernetes Namespace) should use the same name for the Secret
+  that contains the Vault token
+- the CSI-provisioner sidecar needs to provide the name of the Tenant together
+  with the metadata of the PVC to the Ceph-CSI provisioner
+
+## Implementation Outline
+
+- when creating the PVC the Ceph-CSI provisioner needs to store the Kubernetes
+  Namespace of the PVC in its metadata
+  - stores the `csi.volume.owner` (name of Tenant) in the metadata of the
+    volume and sets it as `rbdVolume.Owner`
+- the Ceph-CSI node-plugin needs to request the Vault Token in the NodeStage
+  CSI operation and create/get the key for the PVC
+- the Ceph-CSI provisioner needs to request the Vault Token to delete the key
+  for the PVC in the VolumeDelete CSI operation
+
+### New sidecar: Fetching Tenant Tokens from Kubernetes
+
+CSI is an abstraction and should not communicate with Kubernetes or other
+Container Orchestration systems directly. Therefore, it is needed to
+communicate with a service that can provide the Vault Token from the Tenants.
+This service can be provided by a sidecar, exposing an endpoint as a
+UNIX-Domain-Socket to communicate through.
+
+This sidecar does not exist at the moment. There are other features within
+Ceph-CSI that will benefit from this sidecar, e.g. Topology support. Because
+Ceph-CSI already uses the Kubernetes API to fetch details from Kubernetes, it
+should be acceptable to fetch the Vault Token configuration for the Tenants the
+same way.
+
+The feature for a sidecar that provides access to the required information from
+Kubernetes and other Container Orchestration frameworks is tracked in
+[#1782](https://github.com/ceph/ceph-csi/issues/1782).
+
+## Configuration Details
+
+- the current KMS configuration file needs extensions for Vault Token support
+- introduce a new KMS-type: VaultTokenKMS (the current VaultKMS uses a
+  Kubernetes ServiceAccount)
+- configuration of the VaultTokenKMS can be very similar to VaultKMS for common
+  settings
+- the configuration can override the defaults for each Tenant separately
+  - Vault Service connection details (address, TLS options, ...)
+  - name of the Kubernetes Secret that can be looked up per tenant
+- the configuration points to a Kubernetes Secret per Tenant that contains the
+  Vault Token
+- the configuration points to an optional Kubernetes ConfigMap per Tenant that
+  contains alternative connection options for the VaultTokenKMS service
+
+### Example of the KMS configuration file for Vault Tokens
+
+The configuration is available in the Ceph-CSI containers as
+`/etc/ceph-csi-encryption-kms-config/config.json`:
+
+```json
+{
+    "vault-with-tokens": {
+        "encryptionKMSType": "vaulttokens",
+        "vaultAddress": "http://vault.default.svc.cluster.local:8200",
+        "vaultBackendPath": "secret/",
+        "vaultTLSServerName": "vault.default.svc.cluster.local",
+        "vaultCAFromSecret": "vault-ca",
+        "vaultCAVerify": "false",
+        "tenantConfigName": "ceph-csi-kms-config",
+        "tenantTokenName": "ceph-csi-kms-token",
+        "tenants": {
+            "my-app": {
+                "vaultAddress": "https://vault.example.com",
+                "vaultCAVerify": "true"
+            },
+            "an-other-app": {
+                "tenantTokenName": "storage-encryption-token"
+            }
+        }
+    }
+}
+```
+
+In the Kubernetes StorageClass, the `kmsID` should be set to
+`vault-with-tokens` in order to select the above configuration.
+
+**Required options**:
+
+- `encryptionKMSType`: should be set to `vaulttokens`
+- `vaultAddress`: should be set to the URL of the Vault service
+
+**Optional options**:
+
+- `vaultBackendPath`: defaults to `secret/`
+- `vaultTLSServerName`: not used when unset
+- `vaultCAFromSecret`: not used when unset
+- `vaultCAVerify`: defaults to `true`
+- `tenantConfigName`: the name of the Kubernetes ConfigMap that contains the
+  Vault connection configuration (can be overridden per Tenant, defaults to
+  `ceph-csi-kms-config`)
+- `tenantTokenName`: the name of the Kubernetes Secret that contains the Vault
+  Token (can be overridden per Tenant, defaults to `ceph-csi-kms-token`)
+- `tenants`: list of Tenants (Kubernetes Namespaces) with their connection
+  configuration that differs from the global parameters
+
+### Configuration stored in the Tenants Kubernetes Namespace
+
+The Vault Token needs to be configured per Tenant. Each Tenant can create,
+modify or delete their own personal Token. The Token is stored in a Kubernetes
+Secret in the Kubernetes Namespace where the PVC is created.
+
+```yaml
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ceph-csi-kms-token
+stringData:
+  token: "sample_root_token_id"
+```
+
+The name `ceph-csi-kms-token` is the default, but can be changed by setting
+`tenantTokenName` in the `/etc/ceph-csi-encryption-kms-config/config.json`
+configuration file.
+
+In addition to the Vault Token that can be configured per Tenant, the
+connection parameters to the Vault Service can be stored in the Tenants
+Kubernetes Namespace as well.
+
+```yaml
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ceph-csi-kms-config
+data:
+  vaultAddress: "https://vault.infosec.example.org"
+  vaultBackendPath: "secret/ceph-csi-encryption/"
+  vaultTLSServerName: "vault.infosec.example.org"
+  vaultCAFromSecret: "vault-infosec-ca"
+  vaultCAVerify: "true"
+```
+
+Only parameters with the `vault`-prefix may be changed in the Kubernetes
+ConfigMap of the Tenant.


docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+- the Ceph-CSI sidecar needs to store the namespace (and name of the Secret?)
+  together with the metadata of the PVC
+
+## Implementation Outline
+
+- the Ceph-CSI node-plugin needs to request the token in the NodeStage CSI
+  operation and create/get the key for the PVC
+- the Ceph-CSI provisioner needs to request the token to delete the key for the
+  PVC in the VolumeDelete CSI operation
+- when creating the PVC the Ceph-CSI provisioner needs to store the namespace
```
Madhu-1, line 56: this need to move to fist point

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+- introduce a new KMS-type: VaultTokenKMS (VaultKMS uses a Kubernetes
+  ServiceAccount)
+- configuration of the VaultTokenKMS can be very similar to VaultKMS for common
+  settings
+- tokens for each tenant can be added to the configuration (insecure & needs
+  admin involvement)
+- the configuration can describe defaults for each tenant
+  - Vault Service connection details (address, TLS options, ...)
+  - default token (in case of a single token for every tenant)
+  - name of the Kubernetes Secret that can be looked up per tenant
```
Madhu-1, line 73: this is optional, right?

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+- the current KMS configuration file needs extensions for Vault Token support
+- introduce a new KMS-type: VaultTokenKMS (VaultKMS uses a Kubernetes
+  ServiceAccount)
+- configuration of the VaultTokenKMS can be very similar to VaultKMS for common
+  settings
+- tokens for each tenant can be added to the configuration (insecure & needs
+  admin involvement)
+- the configuration can describe defaults for each tenant
+  - Vault Service connection details (address, TLS options, ...)
+  - default token (in case of a single token for every tenant)
```
Madhu-1, line 72: the default token is also stored in the secret, the default secret name needs to be specified here?

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+
+```json
+{
+    "vault-with-tokens": {
+        "encryptionKMSType": "vaulttoken",
+        "vaultAddress": "http://vault.default.svc.cluster.local:8200",
+        "vaultCAFromSecret": "vault-ca",
+        "vaultCAVerify": "false",
+        "vaultToken": "@@B64_ENCODED_DEFAULT_TOKEN@@",
+        "tenantConfigName": "@@SECRET_NAME_IN_NAMESPACE@@"
```
Madhu-1, line 88: "," is missing at end

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+  PVC
+- a Tenant is the owner of a Kubernetes Namespace
+
+## Restrictions to consider
+
+- tenants can only configure their Kubernetes namespace, their token to access
+  Vault should be located in their namespace (as a Kubernetes Secret)
+- Ceph-CSI can not directly access Kubernetes namespaces, as CSI is an
+  abstraction layer
+- Ceph-CSI needs to talk to a service (sidecar) to access the token from a
```
Madhu-1, line 31: add some details about the sidecar in this doc (its operation, communication type (GRPC, HTTP etc) and also its deployment strategy

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: The list is not really ordered, but I'll move it up with the next update of the PR.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Implementation Outline
+
+- the Ceph-CSI node-plugin needs to request the token in the NodeStage CSI
+  operation and create/get the key for the PVC
+- the Ceph-CSI provisioner needs to request the token to delete the key for the
+  PVC in the VolumeDelete CSI operation
+- when creating the PVC the Ceph-CSI provisioner needs to store the namespace
+  (and Kubernetes Secret?) of the PVC in its metadata
+  - stores the `csi.volume.owner` (name of Tenant) in the journal and sets it
```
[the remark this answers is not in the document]
nixpanic: Sure, the location does not really matter much to me. There also is some metadata for encryption added to the image, so adding the owner makes sense. Any suggestion for other encyption details? (type, or ID of the KMS maybe?)

#1093 sets it in the journal at the moment, but that is mostly for the e2e testing.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: Yes, this needs more details for sure, will add that.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: There does not need to be a default (global/shared) token. Everything that can be configured in a pre-Tenant secret, can also be configured in the KMS configuration file.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: yes, will phrase things a little more clearer

docs/design/proposals/encryption-with-vault-tokens.md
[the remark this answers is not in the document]
Madhu-1: okay we need to consider both cephfs and rbd both as we may support encryption for cephfs also, keeping it at omap is fine, the omap generator is already generating the required omap, we need to update the controller to add owner also.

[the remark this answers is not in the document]
Madhu-1: we are already setting KMS ID on omap data in ReserveName. what other informations do we need to store?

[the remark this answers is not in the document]
nixpanic: I can not think of anything else either. Owner and KMS ID are the ones that should be sufficient.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+- the name of the Kubernetes Secret needs to be known (configured in the
+  StorageClass, or somewhere else)
+- each tenant (Kubernetes namespace) should use the same name for the Secret
+  that contains the Vault token
+- the Ceph-CSI sidecar needs to store the namespace (and name of the Secret?)
+  together with the metadata of the PVC
+
+## Implementation Outline
+
+- when creating the PVC the Ceph-CSI provisioner needs to store the namespace
```
ShyamsundarR, line 52: I am wondering what would happen when PVC name transfers are in play, as create time namespace would change.

But, in such scenarios when the encryption access token is to be tenant/namespace specific, the same token would need to present/used from the originating namespace prior to the transfer, so it maybe feasible that we just depend on the token from the originating namespace. May need to further factor this in.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+`/etc/ceph-csi-encryption-kms-config/config.json`:
+
+```json
+{
+    "vault-with-tokens": {
+        "encryptionKMSType": "vaulttoken",
+        "vaultAddress": "http://vault.default.svc.cluster.local:8200",
+        "vaultCAFromSecret": "vault-ca",
+        "vaultCAVerify": "false",
+        "vaultToken": "@@B64_ENCODED_DEFAULT_TOKEN@@",
```
ShyamsundarR, line 103: I would suggest we move out all secrets from the config map to a secret within the same namespace as ceph-csi. This way, we map in the secret as well into (say) `/etc/ceph-csi-tokens.json` in the nodeplugin/provisioner deployments as an additional volume. It would keep it clean, as in secrets are in a kubernetes secret and not in a config map.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Implementation Outline
+
+- when creating the PVC the Ceph-CSI provisioner needs to store the namespace
+  (and Kubernetes Secret?) of the PVC in its metadata
+  - stores the `csi.volume.owner` (name of Tenant) in the metadata of the
+    volume and sets it as `rbdVolume.Owner`
+- the Ceph-CSI node-plugin needs to request the token in the NodeStage CSI
+  operation and create/get the key for the PVC
+- the Ceph-CSI provisioner needs to request the token to delete the key for the
```
ShyamsundarR, line 58: The typical delete time issue of originating namespace and hence the token in a secret may be unavailable at delete time, i.e the namespace itself may have been deleted and the delete volume would come in asynchronously.

This is unavoidable, so the delete operation would not be able to garbage collect the key for this PVC? If so, when the same PVC is recreated do we intend to reuse the existing key or a new one (if it is possible to detect that the existing key is stale)?

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+  PVC in the VolumeDelete CSI operation
+
```
ShyamsundarR, line 60: I understand that we need the token secret namespace (name is a configured entity) during delete. But I am not clear where it is being stored. Is it part of the CSI per-volume OMap? I guess the "store" comment above also derives from the same confusion.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+```json
+{
+    "vault-with-tokens": {
+        "encryptionKMSType": "vaulttoken",
+        "vaultAddress": "http://vault.default.svc.cluster.local:8200",
+        "vaultCAFromSecret": "vault-ca",
+        "vaultCAVerify": "false",
+        "vaultToken": "@@B64_ENCODED_DEFAULT_TOKEN@@",
+        "tenantConfigName": "@@SECRET_NAME_IN_NAMESPACE@@",
+        "tenants": {
```
ShyamsundarR, line 105: Do we need this flexibility at all? I.e each tenant token as part of a common store in the ceph-csi namespace.

I ask, because if each tenant has a different token, gathering them in the same store seems counter productive to the security requirement for which independent tokens are created in the first place.

Technically as the ceph-csi namespace is given read access to the secrets it can be argued that the user/admin of the ceph-csi namespace can anyway read all the secrets.

ShyamsundarR (commented):
The sidecar approach and a well defined secret name in the tenant namespace and further deriving the namespace from the create request is a better approach. This makes it easier for users to use the feature as they control the secrets in their namespace and also from the perspective of using a common StorageClass to provision encrypted storage from.

Further comments provided are more clarifications and/or corner cases that need to be handled/thought about.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
+# Multi-tenancy with Vault Tokens
+
+## Current Feature: Vault access with single token
+
+The current feature to [support Hashicorp Vault as a KMS is
+documented](./encrypted-pvc.md):
+
+- tenants can create PVCs that are encrypted with a unique key
+- the key to encrypt/decrypt the PVC is stored in Hashicorp Vault
+- there is a single token (Kubernetes ServiceAccount) to access Vault
```
leseb, line 10: What about the Vault token access using the token auth method and not the Kubernetes Auth?

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+- enabling is done in the StorageClass, pointing to a KMS configuration section
+  in a JSON configuration file (Kubernetes ConfigMap)
+
+## High-Level Requirements
+
+The new feature should support multi-tenancy, so that each Tenant can use their
+own Vault Token for PVC encryption:
+
+- each tenant can manage their Vault token (and PVC keys)
+- to get the key from Vault, each tenant should provide their personal token
```
leseb, line 20:
```suggestion
- to get the key-encryption key from Vault, each tenant should provide their personal Vault token
```

Good to clarify IMO.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+- Ceph-CSI should use the token from the tenant to store the unique key for the
```
leseb, line 21:
```suggestion
- Ceph-CSI should use the Vault token from the tenant to store the unique key-encryption key for the
```

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: That is what we're adding now. This part of the **Current Feature** is about what is supported now, not what is going to be added.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: oh, yes, that probably makes things clearer a little

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+  PVC
+- a Tenant is the owner of a Kubernetes Namespace
```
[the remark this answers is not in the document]
nixpanic: yes, I was considering to add a referral option in the configuration file, such as 
```json
"tenants": {
    "my-app": {
        "vaultToken": "the secret value"
    },
    "companion-app": {
        "owner": "my-app"
    }
}
```
That would need to be configured by the storage admin, and may not be re-configurable by tenants themselves (otherwise any tenant would be able to decrypt PVs from other tenants).

I'd see this as a possible extension, and something that can be added at a later point in time.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,125 @@
...
+- [example of the
+  configuration](https://github.com/ceph/ceph-csi/blob/master/examples/kms/vault/kms-config.yaml)
+
+## Dependencies
+
+- the name of the Kubernetes Secret needs to be known (configured in the
+  StorageClass, or somewhere else)
+- each tenant (Kubernetes namespace) should use the same name for the Secret
+  that contains the Vault token
+- the Ceph-CSI sidecar needs to store the namespace (and name of the Secret?)
```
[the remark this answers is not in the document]
nixpanic: well, as CSI should not talk to the CO directly, we can provide a sidecar for Kubernetes that the provisioner and node-plugin can use to fetch the tenant specific Vault configuration from. Currently there is a feature that talks to Kubernetes already (to fetch node labels), so the plan is to re-use that, and move it to a sidecar (or somewhere else) later.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: Right, namespace transfer would make things difficult. Not sure if there are any CSI operations done to inform the storage backend about this. Definitely sounds like a thing to take into account <somewhere>.

Fortunately LUKS support multiple passphrases for block devices. So we may be able to add a passphrase for the new namespace. But that requires receiving an event of the transfer...

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: That is a good point. There is a dependency on the Secret in the Namespace of the Tenant. When the Namespace is deleted, there is no defined order in the deletion of the Secret or PVCs. Deleting the passphrase for the PV will not be possible when the Secret has been removed. Not sure how this can be avoided or solved.

For the recreation, maybe it is not sufficient to have the owner, but it also needs a UUID of the Kubernetes Namespace (or other unique ID of the tenant). However, people may want to restore their Namespace + PVCs from backup, and in that case we would want to use the pre-existing passphrases again...

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: Yes, there are some attributes stored in OMAP already, this would be an additional one. See #1093 for a 1st implementation.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: Yeah, I'm not very fond if listing secrets in a ConfigMap. Ideally the tokens are not available as a file at all, and rotating tokens would just be automatic by issuing a new API call to get it.

The listing of tokens in the ConfigMap was the initial design, and was kept here for now. It would be the 1st thing to drop from the configuration file :smiley:

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: Tokens will most likely not be stored in the configuration file. That would be ideal. Having the storage admin (that maintains the Ceph-CSI namespace) update/add tokens from tenants is extremely painful. Tenants may have different Vault services that they need to use, so having a per-tenant option to overload options will be helpful.

docs/design/proposals/encryption-with-vault-tokens.md
leseb: Ok, thanks. Let's just replace "token" with "JSON Web Token" (JWT) for more accuracy.
"

docs/design/proposals/encryption-with-vault-tokens.md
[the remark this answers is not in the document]
Madhu-1: AFAIU the sidecar doesn't need to store any information right. the plugin will store the secret name and namespace name on the omap object? which is required for Delete operations.

docs/design/proposals/encryption-with-vault-tokens.md
Madhu-1: This would be possible but the controller needs to be extended to watch for a namespace operation. but if the controller is down during namespace transfer the event may go unnoticed we need to find a way (which might be added in kube) to know successful namespace transfers.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: Tokens are not stored in the ConfigMap anymore, #1759 does not read tokens from the ConfigMap (or file) either.

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,179 @@
...
+- [example of the
+  configuration](https://github.com/ceph/ceph-csi/blob/master/examples/kms/vault/kms-config.yaml)
+
+## Dependencies
+
+- the name of the Kubernetes Secret needs to be known (configured in the KMS
+  config file)
+- each Tenant (Kubernetes Namespace) should use the same name for the Secret
+  that contains the Vault token
+- the Ceph-CSI sidecar needs to store the name of the Tenant together with the
```
Madhu-1, line 50: Are we going to store the secret name also if we support custom secrets also?

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,179 @@
...
+  - Vault Service connection details (address, TLS options, ...)
+  - name of the Kubernetes Secret that can be looked up per tenant
+- the configuration points to a Kubernetes Secret per Tenant that contains the
+  Vault Token
+- the configuration points to an optional Kubernetes ConfigMap per Tenant that
+  contains alternative connection options for the VaultTokenKMS service
+
+### Example of the KMS configuration file for Vault Tokens
+
+The configuration is available in the Ceph-CSI containers as
```
Madhu-1, line 95: this won't be available with the cephcsi container anymore right? this will be moved to the sidecar

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,179 @@
...
+```json
+{
+    "vault-with-tokens": {
+        "encryptionKMSType": "vaulttokens",
+        "vaultAddress": "http://vault.default.svc.cluster.local:8200",
+        "vaultBackendPath": "secret/",
+        "vaultTLSServerName": "vault.default.svc.cluster.local",
+        "vaultCAFromSecret": "vault-ca",
+        "vaultCAVerify": "false",
+        "configName": "ceph-csi-kms-config",
```
Madhu-1, line 107: why we need one more configmap? can we add the vault connection details to the configmap itself?

Madhu-1 (commented):
overall LGTM. Thanks, @nixpanic for the detailed design for encryption

docs/design/proposals/encryption-with-vault-tokens.md
Madhu-1: okay if we store the namespace name. based on that we are going to request a sidecar to get the token.

docs/design/proposals/encryption-with-vault-tokens.md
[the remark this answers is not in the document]
nixpanic: Oh, yes, it is not the sidecar that stores anything, it is the provisioner.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: Just the namespace. The name of the Kubernetes Secret comes from the configuration file.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: This is what we currently support and use already.  If it is moved to a sidecar at a later time, that can always be done. I do not see that as a requirement for the feature.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: not if we allow Tenants to self-service their Vault connection parameters. We'd like to enable Tenants to have a different Vault service from the global configuration.

nixpanic: clarified things a little more, the `configName` is really only per Tenant, and optional. It has now been renamed to `tenantConfigName`, and renamed `tokenName` to `tenantTokenName` so things are matching more.

docs/design/proposals/encryption-with-vault-tokens.md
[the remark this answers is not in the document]
ShyamsundarR: Agree, in the current scheme we can always ask the user to add a per-namespace token.

docs/design/proposals/encryption-with-vault-tokens.md
ShyamsundarR: Ack! I wanted to be sure it was in the CSI metadata stash, IOW the OMap.

docs/design/proposals/encryption-with-vault-tokens.md
ShyamsundarR: Per-tenant option to overload vault service is fine. Thanks for taking care of not adding tokens in the config map :+1:

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Fetching Tenant Tokens from Kubernetes
+
+CSI is an abstraction and should not communicate with Kubernetes or other
+Container Orchestration systems directly. Therefore, it is needed to
+communicate with a service that can provide the Vault Token from the Tenants.
+This service can be provided by a sidecar, exposing an endpoint as a
+UNIX-Domain-Socket to communicate through.
+
+This sidecar does not exist at the moment. There are other features within
```
ShyamsundarR, line 72: (question) The side car is planned as part of the initial feature, or would we currently leverage the plugin to call into k8s for now and add the side car later?

docs/design/proposals/encryption-with-vault-tokens.md
ShyamsundarR: Let's park this in a "known issues/gaps" section for now.

docs/design/proposals/encryption-with-vault-tokens.md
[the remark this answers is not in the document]
ShyamsundarR: In case of RBD mirroring, when we get the static PV created it should have the claimRef which can be used to re-populate the OMap regarding the namespace to fetch the token from?

docs/design/proposals/encryption-with-vault-tokens.md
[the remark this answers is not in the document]
ShyamsundarR: Understood, it is stored in the CSI metadata store (OMap as it stands, and evaluating image meta for RBD at present) by the ceph-csi provisioner plugin.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: The sidecar will come later. It is expected that other features will use it too. For now, we'll connect to Kubernetes directly and fetch the Vault Token (and optional Vault connection details) that way.

ShyamsundarR: Ack! thanks for the update.

docs/design/proposals/encryption-with-vault-tokens.md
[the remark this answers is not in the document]
Madhu-1: Yes, we can make use of a namespace in claimRef which is part of PV spec.

Madhu-1 (approved):
LGTM

docs/design/proposals/encryption-with-vault-tokens.md
ShyamsundarR: Again, let's park this in the "Gaps" section.

@nixpanic do we want to add that section and then merge this in?

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: I have created #1782 for this, and will reference it from the doc.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: Updated the doc with a reference to #1782.

ShyamsundarR (approved):

Madhu-1 (approved):

PR #1755, opened 2020-12-03
CI: Move static checks from Travis CI to github action

Madhu-1:
As we are facing issues with the Travis CI moving all the static checks to GitHub action for better and faster execution.


nixpanic:
Instead of re-doing the installation on Ubuntu, you can run the test in containers with commands like:
```
make containerized-test TARGET=gosec
```
There is a GitHub container registry that we could use for re-using container images. But that can be an enhancement for later.

Developers would be able to run the same tests that way. And we need to maintain test-tool installations only in a single place (`scripts/Dockerfile.test`).

Madhu-1:
> Instead of re-doing the installation on Ubuntu, you can run the test in containers with commands like:
> 
> ```
> make containerized-test TARGET=gosec
> ```
> 
> There is a GitHub container registry that we could use for re-using container images. But that can be an enhancement for later.
> 
> Developers would be able to run the same tests that way. And we need to maintain test-tool installations only in a single place (`scripts/Dockerfile.test`).

okay, let me give it a try with running all tests in containers.

Madhu-1:
@nixpanic I will also try to measure the time taken to run different tests with and without containers. will use a better one in Github actions.

Madhu-1:
Here is the rest of the tests

Run tests within the container #1762 :
* commitlint (pull_request) Successful in 2m
* go-test (pull_request) Successful in 4m
* golangci-lint (pull_request) Successful in 6m
*  gosec (pull_request) Successful in 4m
*  lint-extras (pull_request) Successful in 3m
* mod-check (pull_request) Successful in 6m


Test Logs without running tests in a container #1755:

* lint-extras (pull_request) Failing after 1m 
* commitlint / lint (pull_request) Successful in 44s
* go-test / go-test (pull_request) Successful in 2m
* golangci-lint / golangci-lint (pull_request) Successful in 3m
* gosec / gosec (pull_request) Successful in 2m
* mod-check / mod-check (pull_request) Successful in 15s

cc @nixpanic

nixpanic (changes requested):
Thanks for the tests! My preference goes to #1762 as that reduces the maintenance on the test jobs and developers can run the same tests locally with.

The difference in time of execution for the tests is not alarmingly large. Waiting a few minutes for results should be acceptable. The e2e tests take much longer to run.

nixpanic:
Closing as #1762 has been merged.

PR #1763, opened 2020-12-04
rbd: fix omap generator

Madhu-1:
This PR fixes the below issues.

* If the volume is statically provisioned one which contains `staticVolume="true"` on PV spec, don't generate omap data.
* Undo Reservation if any operations fail after Reserving OMAP.
* If the omap already exists don't try to generate it again.
*  Code cleanup to reduce code complexity.

### DIFF
--- a/internal/controller/persistentvolume/persistentvolume.go
+++ b/internal/controller/persistentvolume/persistentvolume.go
@@ -81,7 +81,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	return nil
 }
 
-func (r *ReconcilePersistentVolume) getCredentials(name, namespace string) (map[string]string, error) {
+func (r *ReconcilePersistentVolume) getCredentials(name, namespace string) (*util.Credentials, error) {
+	var cr *util.Credentials
+
+	if name == "" || namespace == "" {
+		errStr := "secret name or secret namespace is empty"
+		util.ErrorLogMsg(errStr)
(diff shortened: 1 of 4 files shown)


internal/rbd/rbd_journal.go
```diff
@@ -559,6 +559,7 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 			util.ErrorLog(ctx, "failed to add UUID mapping %s: %v", rbdVol, err)
 			return err
 		}
+		return nil
```
nixpanic, line 556: A comment would definitely help here.

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -109,7 +122,15 @@ func (r ReconcilePersistentVolume) reconcilePV(obj runtime.Object) error {
 		volumeHandler := pv.Spec.CSI.VolumeHandle
 		secretName := ""
 		secretNamespace := ""
-
+		// check static volume
+		static, err := checkStaticVolume(pv.Spec.CSI.VolumeAttributes["staticVolume"])
```
nixpanic, line 139: It is safer to pass `pv` or one of its components to `checkStaticVolume()`, and verify in the function that the `"staticVolume"` key does exist.

nixpanic (changes requested):
Looks good, just some small nits to improve.

nixpanic (approved):

PR #1796, opened 2020-12-16
deploy: allow rbd nodeplugin to read ConfigMaps from Tenants

nixpanic:
Tenants can have their own ConfigMap that contains connection parameters
to the Vault Service where the PV encyption keys are located. It is
possible for a Tenant to use a different Vault Service than the one
configured by the Storage Admin who deployed Ceph-CSI.

For this, the node-plugin needs to be able to read the ConfigMap from
the Tenants namespace.

See-also: docs/design/proposals/encryption-with-vault-tokens.md

---

<details>

  progress or results

</details>

### DIFF
--- a/charts/ceph-csi-rbd/templates/nodeplugin-clusterrole.yaml
+++ b/charts/ceph-csi-rbd/templates/nodeplugin-clusterrole.yaml
@@ -15,7 +15,11 @@ rules:
     resources: ["nodes"]
     verbs: ["get"]
 {{- end }}
+  # allow to read Vault Token and connection options from the Tenants namespace
   - apiGroups: [""]
     resources: ["secrets"]
+    verbs: ["get"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
     verbs: ["get"]
 {{- end -}}
--- a/deploy/rbd/kubernetes/csi-nodeplugin-rbac.yaml
+++ b/deploy/rbd/kubernetes/csi-nodeplugin-rbac.yaml
@@ -12,8 +12,12 @@ rules:
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get"]
+  # allow to read Vault Token and connection options from the Tenants namespace
   - apiGroups: [""]
     resources: ["secrets"]
+    verbs: ["get"]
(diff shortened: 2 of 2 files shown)


charts/ceph-csi-rbd/templates/nodeplugin-clusterrole.yaml
```diff
@@ -15,7 +15,8 @@ rules:
     resources: ["nodes"]
     verbs: ["get"]
 {{- end }}
+  # allow to read Vault Token and connection options from the Tenants namespace
   - apiGroups: [""]
-    resources: ["secrets"]
+    resources: ["secrets", "configmaps"]
```
Madhu-1, line 20: This should go as a new resource. adding to the existing resource might  give more access when we add access to only one resouce

```suggestion
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["get"]
- apiGroups: [""]
  resources: ["configmaps"]
  verbs: ["get"]
```

nixpanic: How do you mean _adding to the existing resource might give more access_ ? The comment above clearly states that these two resources are needed for the Vault Tokens functionality. If these resources/verbs are changed in a future PR, we'll see the comment and can judge if the change is valid or needs to be split out.

Madhu-1: Yes if we are updating verb for only configmap later to watch or something like that we need to split the resources and add a new one for configmap, we can add separate resources now only so that we don't need to worry about splitting later

Madhu-1: currently, we are keeping each resource independently in both helm and kubernetes yaml files

nixpanic: > currently, we are keeping each resource independently in both helm and kubernetes yaml files

This is the most compelling reason to do so now. Rules should be documented with the reason why it is needed, but as that is not done yet everywhere, splitting seems okay-ish.

Madhu-1 (approved):

Issue #1764, opened 2020-12-04
Cephfs: use enum to check resize is supported or not


Madhu-1:
currently, we are  using bool pointer to find out the ceph cluster supports resize or not we need to replace it with enum values more information is provided by @phlogistonjohn at https://github.com/ceph/ceph-csi/pull/1735#issuecomment-738083768

PR #1812, opened 2020-12-22
cephfs: use enum to check resize is supported or not

yati1998:
Currently, we are using bool pointer to find out
the ceph cluster supports resize or not. This
commit replaces the bool pointer with enum.

Fixes: #1764

### DIFF
--- a/internal/cephfs/volume.go
+++ b/internal/cephfs/volume.go
@@ -119,9 +119,18 @@ func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, volID volumeID) (
 	return &subvol, nil
 }
 
+type operationState int64
+
+const (
+	unknown operationState = iota
+	supported
+	unsupported
+)
+
 type localClusterState struct {
-	// set true if cluster supports resize functionality.
-	resizeSupported *bool
+	// set the enum value i.e., unknown, supported,
+	// unsupported as per the state of the cluster.
+	resizeState operationState
 	// set true once a subvolumegroup is created
 	// for corresponding cluster.
 	subVolumeGroupCreated bool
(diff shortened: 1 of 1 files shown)


internal/cephfs/volume.go
```diff
@@ -119,9 +119,17 @@ func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, volID volumeID) (
 	return &subvol, nil
 }
 
+type ResizeStatus int64
```
nixpanic, line 122: The type is exported outside of the package (**R**esizeStaus), but the constants are not (**r**esizeUnknown). Is that intentional? I guess the type can stay internal to the package too.

internal/cephfs/volume.go
```diff
@@ -119,9 +119,17 @@ func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, volID volumeID) (
...
+const (
+	resizeUnknown ResizeStatus = iota
+	resizeSupported
+	resizeUnsupported
+)
+
 type localClusterState struct {
 	// set true if cluster supports resize functionality.
-	resizeSupported *bool
+	resizeSupported ResizeStatus
```
nixpanic, line 132: probably better to rename this attribute, and call it `resizeState` so that it does not conflict with the naming of the const.

internal/cephfs/volume.go
```diff
@@ -183,19 +191,15 @@ func (vo *volumeOptions) resizeVolume(ctx context.Context, volID volumeID, bytes
 	}
 	// resize subvolume when either it's supported, or when corresponding
 	// clusterID key was not present.
-	if clusterAdditionalInfo[vo.ClusterID].resizeSupported == nil || *clusterAdditionalInfo[vo.ClusterID].resizeSupported {
-		if clusterAdditionalInfo[vo.ClusterID].resizeSupported == nil {
-			clusterAdditionalInfo[vo.ClusterID].resizeSupported = new(bool)
-		}
+	if clusterAdditionalInfo[vo.ClusterID].resizeSupported == resizeUnknown || clusterAdditionalInfo[vo.ClusterID].resizeSupported == resizeSupported {
```
nixpanic, line 194: it was on a single line, but you may improve readability by putting each statement on a single line

nixpanic (changes requested):
just some minor changes to improve maintainability a little

internal/cephfs/volume.go
yati1998: okay

nixpanic (approved):

internal/cephfs/volume.go
```diff
@@ -119,9 +119,17 @@ func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, volID volumeID) (
...
 }
 
+type resizeState int64
+
+const (
+	resizeUnknown resizeState = iota
+	resizeSupported
+	resizeUnsupported
+)
+
```
Madhu-1, line 129: I would like to change this to operationState and below const to represent an operation which might be helpful when we need the same for other operations

```go
type operationState int64

const (
	unknown operationState = iota
	supported
	unsupported
)
```

internal/cephfs/volume.go
```diff
@@ -119,9 +119,17 @@ func (vo *volumeOptions) getSubVolumeInfo(ctx context.Context, volID volumeID) (
...
+type operationState int64
+
+const (
+	unknown operationState = iota
+	supported
+	unsupported
+)
+
 type localClusterState struct {
 	// set true if cluster supports resize functionality.
```
Madhu-1, line 131: This comment needs to be updated and can you also please change `resizeSupported` to `resizeState`. as it no more a Bool value.

Madhu-1 (approved):

PR #1833, opened 2021-01-15
Explain why deep-flatten isn't available.

dren-dk:

I spent a bit of time researching why why ceph-csi refused to support some very handy image features that I wanted to use,
so I thought it would be helpful to add the explanation that I found.


dren-dk:
Found a better link.

PR #1834, opened 2021-01-15
Explain why certain image features are missing

dren-dk:

I spent a good long while tracking down the reason for some image features being unsupported, so I have added the relevant link to the documentation.


nixpanic (changes requested):
Hi! Thanks for your contribution to the docs, these notes should be very helpful for others.

Could you please create a single commit, and use the following format?
```
doc: explain why certain features are unsupported

Optional description goes here.
And can be multiple lines or paragraphs.

Signed-off-by: Your real Name <email@example.net>
```

When you have the single commit, force-push it to your branch so that the PR gets updated. If you have difficulties with that, I might be able to assist or take care of it for you.

dren-dk:
I think that's it.

nixpanic (approved):

PR #1739, opened 2020-11-30
rbd: add support to create snapshot and clone from snapshot in different pools

Madhu-1:
* Added support to create volumesnapshot in a different pool
   as cephcsi internally creates an rbd image when a user requested a snapshot, cephcsi can create an rbd image in a  different pool

* Added support to create pvc from snapshot to a different pool
   as cephcsi internally creates a clone from a snapshot its possible to clone to a new pool

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

PR #1838, opened 2021-01-20
rbd: add support to create snapshot and clone from snapshot in different pools

Yuggupta27:
Add support to create a snapshot to a different pool.
Also, create a pvc from a snapshot to a different pool.

Co-authored-by: Madhu Rajanna <madhupr007@gmail.com>
Closes: #1739 

---

<details>

  progress or results

</details>

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -958,7 +958,7 @@ var _ = Describe("RBD", func() {
 			By("create ROX PVC clone and mount it to multiple pods", func() {
 				// snapshot beta is only supported from v1.17+
 				if k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
-					err := createRBDSnapshotClass(f)
+					err := createRBDSnapshotClass(f, nil)
 					if err != nil {
 						e2elog.Failf("failed to create storageclass with error %v", err)
 					}
@@ -1093,7 +1093,7 @@ var _ = Describe("RBD", func() {
 				if !k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
 					Skip("pvc restore is only supported from v1.17+")
 				}
-				err := createRBDSnapshotClass(f)
+				err := createRBDSnapshotClass(f, nil)
 				if err != nil {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
@@ -1298,7 +1298,7 @@ var _ = Describe("RBD", func() {
 				// Create a PVC clone and bind it to an app within the namespace
 				// snapshot beta is only supported from v1.17+
 				if k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
-					err = createRBDSnapshotClass(f)
+					err = createRBDSnapshotClass(f, nil)
 					if err != nil {
 						e2elog.Failf("failed to create storageclass with error %v", err)
 					}
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -956,6 +956,103 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("create rbd snapshots in different pool", func() {
+				// snapshot beta is only supported from v1.17+
+				if k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
+					err := deleteResource(rbdExamplePath + "snapshotclass.yaml")
+					if err != nil {
+						e2elog.Failf("failed to delete snapshotclass with error %v", err)
+					}
+					snapshotPool := "snapshot-test"
+					// create pool for snapshots
+					err = createPool(f, snapshotPool)
+					if err != nil {
+						e2elog.Failf("failed to create pool %s with error %v", snapshotPool, err)
+					}
+					// create snapshotclass with pool option
+					param := map[string]string{
+						"pool": snapshotPool,
+					}
+					err = createRBDSnapshotClass(f, param)
+					if err != nil {
+						e2elog.Failf("failed to create snapshotclass with error %v", err)
+					}
+					err = validateSnapshotInDifferentPool(f, snapshotPool, "", defaultRBDPool)
+					if err != nil {
+						e2elog.Failf("failed to validate snapshot in different pool with error %v", err)
+					}
+
+					err = deletePool(snapshotPool, false, f)
+					if err != nil {
+						e2elog.Failf("failed to delete pool %s with error %v", snapshotPool, err)
+					}
+					err = deleteResource(rbdExamplePath + "snapshotclass.yaml")
+					if err != nil {
+						e2elog.Failf("failed to delete snapshotclass with error %v", err)
+					}
+					err = createRBDSnapshotClass(f, nil)
+					if err != nil {
+						e2elog.Failf("failed to create snapshotclass with error %v", err)
+					}
+				}
+			})
+
+			By("create rbd clones in different pool", func() {
+				// snapshot beta is only supported from v1.17+
+				if k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
+					err := deleteResource(rbdExamplePath + "snapshotclass.yaml")
+					if err != nil {
+						e2elog.Failf("failed to delete snapshotclass with error %v", err)
+					}
+					clonePool := "clone-test"
+					// create pool for clones
+					err = createPool(f, clonePool)
+					if err != nil {
+						e2elog.Failf("failed to create pool %s with error %v", clonePool, err)
+					}
+					// create snapshotclass with pool option
+					param := map[string]string{
+						"pool": defaultRBDPool,
+					}
+					err = createRBDSnapshotClass(f, param)
+					if err != nil {
+						e2elog.Failf("failed to create snapshotclass with error %v", err)
+					}
+					cloneSC := "clone-storageclass"
+					param = map[string]string{
+						"pool": clonePool,
+					}
+					// create new storageclass with new pool
+					err = createRBDStorageClass(f.ClientSet, f, cloneSC, nil, param, deletePolicy)
+					if err != nil {
+						e2elog.Failf("failed to create storageclass with error %v", err)
+					}
+					err = validateSnapshotInDifferentPool(f, defaultRBDPool, cloneSC, clonePool)
+					if err != nil {
+						e2elog.Failf("failed to validate clones in different pool with error %v", err)
+					}
+
+					_, err = framework.RunKubectl(cephCSINamespace, "delete", "sc", cloneSC, "--ignore-not-found=true")
+					if err != nil {
+						e2elog.Failf("failed to delete storageclass %s with error %v", cloneSC, err)
+					}
+
+					err = deleteResource(rbdExamplePath + "snapshotclass.yaml")
+					if err != nil {
+						e2elog.Failf("failed to delete snapshotclass with error %v", err)
+					}
+
+					err = deletePool(clonePool, false, f)
+					if err != nil {
+						e2elog.Failf("failed to delete pool %s with error %v", clonePool, err)
+					}
+					err = createRBDSnapshotClass(f, nil)
+					if err != nil {
+						e2elog.Failf("failed to create snapshotclass with error %v", err)
+					}
+				}
+			})
+
 			By("create ROX PVC clone and mount it to multiple pods", func() {
 				// snapshot beta is only supported from v1.17+
 				if k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -181,7 +181,7 @@ var _ = Describe("RBD", func() {
 		if err != nil {
 			e2elog.Failf("failed to create configmap with error %v", err)
 		}
-		err = createRBDStorageClass(f.ClientSet, f, nil, nil, deletePolicy)
+		err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
 		if err != nil {
 			e2elog.Failf("failed to create storageclass with error %v", err)
 		}
@@ -329,7 +329,7 @@ var _ = Describe("RBD", func() {
 				if err != nil {
 					e2elog.Failf("failed to delete storageclass with error %v", err)
 				}
-				err = createRBDStorageClass(f.ClientSet, f, nil, map[string]string{"csi.storage.k8s.io/fstype": "ext4"}, deletePolicy)
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, map[string]string{"csi.storage.k8s.io/fstype": "ext4"}, deletePolicy)
 				if err != nil {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
@@ -343,7 +343,7 @@ var _ = Describe("RBD", func() {
 				if err != nil {
 					e2elog.Failf("failed to delete storageclass with error %v", err)
 				}
-				err = createRBDStorageClass(f.ClientSet, f, nil, nil, deletePolicy)
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
 				if err != nil {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
@@ -354,7 +354,7 @@ var _ = Describe("RBD", func() {
 				if err != nil {
 					e2elog.Failf("failed to delete storageclass with error %v", err)
 				}
-				err = createRBDStorageClass(f.ClientSet, f, nil, map[string]string{"encrypted": "true"}, deletePolicy)
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, map[string]string{"encrypted": "true"}, deletePolicy)
 				if err != nil {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
@@ -368,7 +368,7 @@ var _ = Describe("RBD", func() {
 				if err != nil {
 					e2elog.Failf("failed to delete storageclass with error %v", err)
 				}
-				err = createRBDStorageClass(f.ClientSet, f, nil, nil, deletePolicy)
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
 				if err != nil {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
@@ -383,7 +383,7 @@ var _ = Describe("RBD", func() {
 					"encrypted":       "true",
 					"encryptionKMSID": "vault-test",
 				}
-				err = createRBDStorageClass(f.ClientSet, f, nil, scOpts, deletePolicy)
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, scOpts, deletePolicy)
 				if err != nil {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
@@ -397,7 +397,7 @@ var _ = Describe("RBD", func() {
 				if err != nil {
 					e2elog.Failf("failed to delete storageclass with error %v", err)
 				}
-				err = createRBDStorageClass(f.ClientSet, f, nil, nil, deletePolicy)
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
 				if err != nil {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
@@ -412,7 +412,7 @@ var _ = Describe("RBD", func() {
 					"encrypted":       "true",
 					"encryptionKMSID": "vault-tokens-test",
 				}
-				err = createRBDStorageClass(f.ClientSet, f, nil, scOpts, deletePolicy)
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, scOpts, deletePolicy)
 				if err != nil {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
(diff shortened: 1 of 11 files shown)


e2e/rbd.go
```diff
@@ -1023,17 +1025,114 @@ var _ = Describe("RBD", func() {
...
+					if err != nil {
+						e2elog.Failf("failed to delete snapshotclass with error %v", err)
+					}
+					snapshotPool := "snapshot-test"
+					// create pool for snapshots
+					err = createPool(f, snapshotPool)
+					if err != nil {
+						e2elog.Failf("failed to create pool %s with error %v", snapshotPool, err)
+					}
+					// create snapshotclass to with pool option
```
humblec: `create snapshotclass with pool option`

e2e/rbd.go
```diff
@@ -1023,17 +1025,114 @@ var _ = Describe("RBD", func() {
...
+					if err != nil {
+						e2elog.Failf("failed to delete snapshotclass with error %v", err)
+					}
+					clonePool := "clone-test"
+					// create pool for clones
+					err = createPool(f, clonePool)
+					if err != nil {
+						e2elog.Failf("failed to create pool %s with error %v", clonePool, err)
+					}
+					// create snapshotclass to with pool option
```
humblec: same as above.

e2e/rbd.go
```diff
@@ -1023,17 +1025,114 @@ var _ = Describe("RBD", func() {
...
+					}
+					err = createRBDSnapshotClass(f, param)
+					if err != nil {
+						e2elog.Failf("failed to create snapshotclass with error %v", err)
+					}
+					cloneSC := "clone-storageclass"
+					param = map[string]string{
+						"pool": clonePool,
+					}
+					// create new storageclass to with new pool
```
humblec: same as above

e2e/rbd.go
```diff
@@ -1023,17 +1025,114 @@ var _ = Describe("RBD", func() {
...
+					err = createRBDStorageClass(f.ClientSet, f, cloneSC, nil, param, deletePolicy)
+					if err != nil {
+						e2elog.Failf("failed to create storageclass with error %v", err)
+					}
+					err = validateSnapshotInDifferentPool(f, defaultRBDPool, cloneSC, clonePool)
```
humblec, line 1030: The function here is validateSnapshotInDifferentPool and err says `failed to validate clone in different pool`

e2e/rbd_helper.go
```diff
@@ -225,6 +230,164 @@ func validateImageOwner(pvcPath string, f *framework.Framework) error {
...
+	}
+	pvcClone.Namespace = f.UniqueName
+	// if request is to create clone with different sc
+	if cloneSc != "" {
+		pvcClone.Spec.StorageClassName = &cloneSc
+	}
+	appClone.Namespace = f.UniqueName
+	pvcClone.Spec.DataSource.Name = fmt.Sprintf("%s%d", f.UniqueName, 0)
+
+	// create multiple PVC from same snapshot
```
humblec: s/PVC/PVCs/

e2e/rbd.go
Yuggupta27: Thanks for catching! will update the PR.

e2e/rbd.go
Yuggupta27: The function `validateSnapshotInDifferentPool` refers to creating snapshots in a different pool. It contains parameters (cloneSC,clonePool in this case) that allow the restore operation of the created snapshots in a different pool to verify working across multiple pools.
The error message `failed to validate clone in different pool` refers to the failure of the complete test case which is `create rbd clones in different pool`. Let me know if err message is not apt :)

humblec: yeah I get that, if we are looking for highlevel test function reference, we could log it as `failed to create rbd clones in different pool` too.  I am fine with current form though.

internal/rbd/controllerserver.go
```diff
@@ -800,7 +800,7 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		vol := generateVolFromSnap(rbdSnap)
 		err = vol.Connect(cr)
 		if err != nil {
-			uErr := undoSnapshotCloning(ctx, vol, rbdSnap, vol, cr)
+			uErr := undoSnapshotCloning(ctx, rbdVol, rbdSnap, vol, cr)
```
nixpanic: The commit " rbd: pass parent volume to undoSnapshotCloning function" only changes the variable names on a few locations, it should probably be squashed with an other one?

internal/rbd/rbd_util.go
```diff
@@ -460,19 +460,18 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
...
 		if vol.RbdImageName == "" {
 			return depth, nil
 		}
+		err := vol.openIoctx()
+		if err != nil {
+			return depth, err
+		}
+
 		err = vol.getImageInfo()
+		vol.ioctx.Destroy()
```
nixpanic, line 553: This should probably be done in a `defer`? Or maybe call `vol.Destroy()` instead?

internal/rbd/rbd_util.go
```diff
@@ -630,17 +629,19 @@ func (rv *rbdVolume) checkImageChainHasFeature(ctx context.Context, feature uint
...
 			return false, nil
 		}
+		err := vol.openIoctx()
+		if err != nil {
+			return false, err
+		}
+
 		err = vol.getImageInfo()
+		vol.ioctx.Destroy()
+		vol.ioctx = nil
```
nixpanic, line 729: same here, why not do it in a `defer`, and why not `vol.Destroy()`?

internal/rbd/rbd_journal.go
```diff
@@ -110,6 +110,7 @@ deletion are inverse of each other, and protected by the request name lock, and
 hence any stale omaps are leftovers from incomplete transactions and are hence
 safe to garbage collect.
 */
+// nolint:gocyclo // reduce complexity
```
nixpanic, line 123: please use `FIXME: reduce complexity`. We want it listed in https://www.tickgit.com/browse?repo=github.com/ceph/ceph-csi

internal/rbd/controllerserver.go
```diff
@@ -444,8 +444,17 @@ func (cs *ControllerServer) createVolumeFromSnapshot(ctx context.Context, cr *ut
 
 	// update parent name(rbd image name in snapshot)
 	rbdSnap.RbdImageName = rbdSnap.RbdSnapName
+	parentVol := generateVolFromSnap(rbdSnap)
+	// as we are operating on single cluster reuse the connection
+	parentVol.conn = rbdVol.conn
```
nixpanic: Connections are reference counted, you can not use it like this. Just use `Connect()` with the same parameters and it should work as intended.

internal/rbd/controllerserver.go
```diff
@@ -444,8 +444,17 @@ func (cs *ControllerServer) createVolumeFromSnapshot(ctx context.Context, cr *ut
...
+	err = parentVol.openIoctx()
```
nixpanic: Opening the ioctx is something internal to the rbdVolume struct, it should ideally only be done from `rbdVolume` methods. If it is needed to open the ioctx, do so in `cloneRbdImageFromSnapshot()` or whatever function requires it

e2e/rbd.go
```diff
@@ -167,7 +167,7 @@ var _ = Describe("RBD", func() {
 		if err != nil {
 			e2elog.Failf("failed to create configmap with error %v", err)
 		}
-		err = createRBDStorageClass(f.ClientSet, f, nil, nil, deletePolicy)
+		err = createRBDStorageClass(f.ClientSet, f, "", nil, nil, deletePolicy)
```
nixpanic: instead of an empty string, define a const somewhere (`defaultSCName = ""`), otherwise everyone (or at least I) will always wonder what the `""` means and have to look it up

e2e/rbd.go
```diff
@@ -445,7 +447,7 @@ var _ = Describe("RBD", func() {
 					totalCount := 10
 					wgErrs := make([]error, totalCount)
 					wg.Add(totalCount)
-					err := createRBDSnapshotClass(f, nil))
+					err := createRBDSnapshotClass(f, nil)
```
nixpanic: this part looks like a cleanup of some other commit?

e2e/rbd.go
```diff
@@ -813,7 +813,7 @@ var _ = Describe("RBD", func() {
 				validateRBDImageCount(f, 1, defaultRBDPool)
 				// list RBD images and check if one of them has the same prefix
 				foundIt := false
-				images, err := listRBDImages(f)
+				images, err := listRBDImages(f, defaultRBDPool)
```
nixpanic, line 736: this change should be squashed into the change where the function got the additional parameter

nixpanic (changes requested):

internal/rbd/controllerserver.go
```diff
@@ -958,7 +983,16 @@ func (cs *ControllerServer) doSnapshotClone(ctx context.Context, parentVol *rbdV
 		return ready, cloneRbd, err
 	}
 
-	err = cloneRbd.flattenRbdImage(ctx, cr, false, rbdHardMaxCloneDepth, rbdSoftMaxCloneDepth)
+	forceFlatten := false
+	//  As once the snapshot limit on a parent is reached we cannot flatten the
+	//  images as we dont know the pool in which clones are created,
+	// if the parent image pool and clone pool are different flatten the image
```
humblec, line 960: nit: above is just one space compared to 2 lines above this .. just pointing out

internal/rbd/rbd_journal.go
```diff
@@ -184,25 +185,43 @@ func checkSnapCloneExists(ctx context.Context, parentVol *rbdVolume, rbdSnap *rb
...
 			return false, err
 		}
 	}
 	if err != nil {
 		return false, err
 	}
 
+	//  As once the snapshot limit on a parent is reached we cannot flatten the
+	//  cloned images as we dont know the pool in which clones are created,
+	// if the parent image pool and clone pool are different flatten the image
```
humblec, line 208: same as above

humblec:
`rbd: add support to create snapshot and clone form snapshot in different pools..` pr titile typo on `form` instead of `from` :)

internal/rbd/controllerserver.go
Yuggupta27: This PR does not make any changes to `undoSnapshotCloning()` implementation, just passes parent volume to clean up the snapshots on parent volume. This commit covers all the argument changes for a single function call at all places, thus kept in a single commit. Should I leave it as a commit or squash it with another one?

e2e/rbd.go
Yuggupta27: Done, defined a const for the same :+1:

internal/rbd/controllerserver.go
Yuggupta27: Thanks for the suggestion! Made the required changes.

internal/rbd/rbd_journal.go
Yuggupta27: Done.

internal/rbd/rbd_util.go
```diff
@@ -951,7 +951,7 @@ func (rv *rbdVolume) hasSnapshotFeature() bool {
 }
 
 func (rv *rbdVolume) createSnapshot(ctx context.Context, pOpts *rbdSnapshot) error {
-	util.DebugLog(ctx, "rbd: snap create %s using mon %s", pOpts, pOpts.Monitors)
+	util.DebugLog(ctx, "rbd: snap create %s@%s using mon %s", rv, pOpts.RbdSnapName, pOpts.Monitors)
```
nixpanic: Nit, you can probably use `pOpts.String()` here

internal/rbd/rbd_util.go
```diff
@@ -963,7 +963,7 @@ func (rv *rbdVolume) createSnapshot(ctx context.Context, pOpts *rbdSnapshot) err
 }
 
 func (rv *rbdVolume) deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot) error {
-	util.DebugLog(ctx, "rbd: snap rm %s using mon %s", pOpts, pOpts.Monitors)
+	util.DebugLog(ctx, "rbd: snap rm %s@%s using mon %s", rv, pOpts.RbdSnapName, pOpts.Monitors)
```
nixpanic: `pOpts.String()` here too?

internal/rbd/rbd_util.go
```diff
@@ -495,23 +495,24 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
 			return depth, nil
 		}
+		err := vol.openIoctx()
+		if err != nil {
+			return depth, err
+		}
+
 		err = vol.getImageInfo()
+		vol.Destroy()
+		vol.ioctx = nil
```
nixpanic, line 554: Never use `vol` again after calling `Destroy()` on it. Maybe you need to create a new `vol` inside this loop instead?

Don't set ioctx to nil, that should not be needed. If it is, there is a problem somewhere else.

internal/rbd/rbd_util.go
nixpanic: Same problem here, maybe simply create and destroy the `vol` inside the loop.

These two functions have a lot of the same code. Maybe a cleanup is appropriate? Can you consider using.creating a helper function like
```go
func (rv *rbdVolume) getParent() *rbdVolume {}
```
and then use it like this in the loop:
```golang
currentVol := rv
for {
    vol, err := currentVol.getParent()
    if err != nil {
        ...
    }
    defer vol.Destroy()
    ....
    currentVol = vol
}
```

This prevents the need to modify and reset the `vol` object, which can lead to unexpected failures as the object does not really support that. With an approach like this, the complexity of the code can be reduced quite a bit, I think.

internal/rbd/rbd_util.go
```diff
@@ -1002,7 +1001,7 @@ func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *r
 	}
 
 	util.DebugLog(ctx, logMsg,
-		pSnapOpts, image, rv.imageFeatureSet.Names(), rv.Monitors)
+		parentVol, pSnapOpts.RbdSnapName, rv, rv.imageFeatureSet.Names(), rv.Monitors)
```
nixpanic: can you use `pSnapOpts.String()` here?

nixpanic (changes requested):
Nothing major, but some nits and a consideration for cleanup (can be an own PR later).

At least the last commit text has both co-authored-by and signed-off-by Yug, just list your name once ;-)

Yuggupta27:
> Nothing major, but some nits and a consideration for cleanup (can be an own PR later).

Sure @nixpanic , Just not to complicate things more in this PR, I'll send a separate PR for the cleanup discussed after this PR is merged.

nixpanic:
> > Nothing major, but some nits and a consideration for cleanup (can be an own PR later).
> 
> Sure @nixpanic , Just not to complicate things more in this PR, I'll send a separate PR for the cleanup discussed after this PR is merged.

Ok, please create an issue for that, or post a work-in-progress PR so that there is some guarantee it will get done.

internal/rbd/rbd_util.go
```diff
@@ -963,7 +963,7 @@ func (rv *rbdVolume) createSnapshot(ctx context.Context, pOpts *rbdSnapshot) err
 }
 
 func (rv *rbdVolume) deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot) error {
-	util.DebugLog(ctx, "rbd: snap rm %s using mon %s", pOpts, pOpts.Monitors)
+	util.DebugLog(ctx, "rbd: snap rm %s@%s using mon %s", rv, pOpts.String(), pOpts.Monitors)
```
nixpanic, line 1071: this will look weird. `pOpts.String()` should contain the full rbd image path, including `@` and what comes before it

internal/rbd/rbd_util.go
nixpanic: Not doing it in this PR is sortof ok. But I'd appreciate a `FIXME: ` type of comment.

internal/rbd/controllerserver.go
```diff
@@ -944,6 +944,20 @@ func (cs *ControllerServer) doSnapshotClone(ctx context.Context, parentVol *rbdV
 		return ready, cloneRbd, err
 	}
 
+	defer func() {
+		// If we hit any error below, cleanup the snapshot created on the clone
+		if err != nil {
```
nixpanic, line 925: this if gets nested pretty deep. Is it needed to do the `err != nil` check before calling `errors.Is()`?

internal/rbd/rbd_util.go
```diff
@@ -1002,7 +1010,7 @@ func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *r
 	}
 
 	util.DebugLog(ctx, logMsg,
-		pSnapOpts, image, rv.imageFeatureSet.Names(), rv.Monitors)
+		parentVol, pSnapOpts.String(), rv, rv.imageFeatureSet.Names(), rv.Monitors)
```
nixpanic: logging `parentVol` should not be required, it is supposed to be part of `pSnapOpts.String()`

e2e/rbd_helper.go
```diff
@@ -6,7 +6,10 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
+	"time"
 
+	"github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
```
nixpanic, line 11: Please use `volumesnapshot` instead of `v1beta1` in the code. When we move to a newer version, we would not need to update the `v1beta1` everywhere.

nixpanic (changes requested):
Just a few nits, almost good to go!

humblec:
@Yuggupta27 may be a quick rebase ?

Yuggupta27:

rook-deployment failed

Madhu-1:

deploy snapshotter failed

Yuggupta27:

Rook-deployment failure

humblec:
Moving this out of release 3.3

Madhu-1:
This PR does two below functionalities

* Add support to create a snapshot in a different pool
* Add support to create a clone from a snapshot in a different pool

@Yuggupta27 can we split this PR into two separate PR's? 

Let's first focus on `Add support to create a clone from a snapshot in a different pool` and then `Add support to create a snapshot in a different pool`

Yuggupta27:
> This PR does two below functionalities
> 
> * Add support to create a snapshot in a different pool
> * Add support to create a clone from a snapshot in a different pool
> 
> @Yuggupta27 can we split this PR into two separate PR's?
> 
> Let's first focus on `Add support to create a clone from a snapshot in a different pool` and then `Add support to create a snapshot in a different pool`

Sure @Madhu-1, that will definitely be a better approach. Will send separate PRs for both the features. Thanks for the suggestion!

humblec:
> This PR does two below functionalities
> 
>     * Add support to create a snapshot in a different pool
> 
>     * Add support to create a clone from a snapshot in a different pool
> 
> 
> @Yuggupta27 can we split this PR into two separate PR's?
> 
> Let's first focus on `Add support to create a clone from a snapshot in a different pool` and then `Add support to create a snapshot in a different pool`

imo, The use case we have to cover:

1) Snapshot creation to same pool (Pool A) and clone to different ( Pool B) 
2) Snapshot creation to different pool (Pool B) and clone to same ( Pool B)
3) Snapshot creation to different pool ( Pool B) and clone to different ( Pool A/C)

As a general view, we have to do some cleanup or refactoring around the snapshot and clone code or have to revisit this code once again before getting everything in to make sure snapshot leaks, volume leaks..etc. are not happening. Difficult task though :)

Rakshith-R:
This would need a `DNM` label as snapshot into another pool always causes a flatten on the temp clone image and will leave stales due to https://github.com/ceph/ceph-csi/issues/2327.

Yuggupta27:
> This would need a `DNM` label as snapshot into another pool always causes a flatten on the temp clone image and will leave stales due to #2327.

No issues. This PR is marked as a draft anyway and we don't plan on working on it anytime soon. :+1:

internal/rbd/rbd_util.go
```diff
@@ -1103,12 +1125,18 @@ func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *r
...
 	}
 
+	// As the clone is yet to be created, open the Ioctx.
 	err = rv.openIoctx()
 	if err != nil {
 		return fmt.Errorf("failed to get IOContext: %w", err)
 	}
+	// Destroy only the ioctx, connection is needed for other operations.
+	defer func() {
+		defer rv.ioctx.Destroy()
```
nixpanic, line 1135: No need to do this, keep it open for potential other needs. When `rv.Destroy()` gets called, all attributes of `rv` get cleaned up.

internal/rbd/rbd_util.go
```diff
@@ -1089,7 +1111,7 @@ func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *r
 	}
 
 	util.DebugLog(ctx, logMsg,
-		pSnapOpts, image, rv.imageFeatureSet.Names(), rv.Monitors)
+		pSnapOpts.String(), rv, rv.imageFeatureSet.Names(), rv.Monitors)
```
nixpanic, line 1114: `pSnapOpts` is a pointer to an rbdSnapshot, there should not be a need to call `.String()`

Issue #667, opened 2019-10-04
Qualify or move RBD NBD to full support


humblec:
Eventhough we have support for rbd nbd in CSI , it is not the default or not qualified much. This is a feature request to bring more qualification to rbd-nbd and make it fully supported with RBD CSI.  The main issue we have come across on user space based client/mounter ( fuse, rbd-nbd) is the unavailability or disturbed mounts on workload attached shares/volumes when the CSI pods are respawned. This is the main technical limitation we have atm. 

Reference issue# https://github.com/ceph/ceph-csi/issues/792

Here are the Items needed to move this issue to fixed:

**Ceph CSI v3.4.0 Progress: [Alpha]** 

- [x]  Volume Healer Design doc: https://github.com/ceph/ceph-csi/pull/2275
- [x]  Volume Healer Code Changes: https://github.com/ceph/ceph-csi/pull/2108

**v3.4.0 Beyond:** 

- [ ]  I/O path concern and corruption possibilities which include both file and block mode: 
[comment 1](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677313425) , [comment 2 ]( https://github.com/ceph/ceph-csi/pull/2275/#discussion_r676706321) , [comment 3]( https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677331644)
- [ ] Attacher dependency logic : 
[no need for attacher](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677314730)
- [ ] Kubernetes implementation leak to CSI :  
[csi leak](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677318536)

humblec:
#557

dillaman:
Perhaps the biggest current limitation of NBD is that the rbd-nbd daemon is run within the ceph-csi container. Upgrading the container is not possible w/o killing running workloads.

There are plans to support multiple NBD block devices under a single rbd-nbd process. Additionally, multiple rbd-nbd daemons could be attached to the same block device to provide round-robin IO support (multipath). If rbd-nbd was moved to its own container, was run in a set of at least 2, and an upgrade would only bring down one rbd-nbd daemon concurrently, we could address those concerns.

[...]

PR #1831, opened 2021-01-12
Provide capacity of block-mode RBD volumes

nixpanic:
# Describe what this PR does #

Metrics from block-mode volumes are not available from Kubelet. CephFS and RBD-with-filesystem do return metrics. This adds support for the `capacity` attribute for block-mode volumes.

## Related issues ##

It seems Kubernetes does not call `NodeGetVolumeStats` got block-mode volumes yet. This will get added through https://github.com/kubernetes/kubernetes/pull/97972 _(merged)_

Without the change in Kubernetes, this can not be tested (and we do not have tests for metrics either yet).

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)
- `/retest all`: run this in case the CentOS CI failed to start/report any test
  progress or results

</details>

PR #1811, opened 2020-12-21
rebase: update minikube to latest version

Madhu-1:
As minikube [1.17.1](https://github.com/kubernetes/minikube/releases) is released and updating the minikube to the latest available version.

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

PR #1839, opened 2021-01-21
e2e: add a test case for rbd-nbd mounter

pkalever:

Validate the basic working of rbd-nbd

**Test cases:**
1. Create a PV with rbd-nbd backend and start the application pod using it
2. After the application pod is started, restart the node plugin and expect the IO to fail as the rbd-nbd process is killed
3. After restarting the node plugin, restart the rbd-nbd process by reattaching/re-mapping the device connection and expect the IO from the application pod to continue

### Dependencies ###
- [x] Requires `ceph/ceph:pacific` which is not yet released
- [x] Depends on `nbd.ko` in minikube, see https://github.com/kubernetes/minikube/pull/10217

Updates: #667

### DIFF
--- a/e2e/pod.go
+++ b/e2e/pod.go
@@ -105,17 +105,40 @@ func waitForDeploymentComplete(name, ns string, c kubernetes.Interface, t int) e
 	return nil
 }
 
-func getCommandInPodOpts(f *framework.Framework, c, ns string, opt *metav1.ListOptions) (framework.ExecOptions, error) {
+func findPodAndContainerName(f *framework.Framework, ns, cn string, opt *metav1.ListOptions) (string, string, error) {
+	podList, err := f.PodClientNS(ns).List(context.TODO(), *opt)
+	if err != nil {
+		return "", "", err
+	}
+
+	if len(podList.Items) == 0 {
+		return "", "", errors.New("podlist is empty")
+	}
+
+	if cn != "" {
+		for i := range podList.Items {
+			for j := range podList.Items[i].Spec.Containers {
+				if podList.Items[i].Spec.Containers[j].Name == cn {
+					return podList.Items[i].Name, cn, nil
+				}
+			}
+		}
+		return "", "", errors.New("container name not found")
+	}
+	return podList.Items[0].Name, podList.Items[0].Spec.Containers[0].Name, nil
+}
+
+func getCommandInPodOpts(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (framework.ExecOptions, error) {
 	cmd := []string{"/bin/sh", "-c", c}
-	pods, err := listPods(f, ns, opt)
+	pName, cName, err := findPodAndContainerName(f, ns, cn, opt)
 	if err != nil {
 		return framework.ExecOptions{}, err
 	}
 	return framework.ExecOptions{
 		Command:            cmd,
-		PodName:            pods[0].Name,
+		PodName:            pName,
 		Namespace:          ns,
-		ContainerName:      pods[0].Spec.Containers[0].Name,
+		ContainerName:      cName,
 		Stdin:              nil,
 		CaptureStdout:      true,
 		CaptureStderr:      true,
@@ -170,7 +193,19 @@ func listPods(f *framework.Framework, ns string, opt *metav1.ListOptions) ([]v1.
 }
 
 func execCommandInPod(f *framework.Framework, c, ns string, opt *metav1.ListOptions) (string, string, error) {
-	podOpt, err := getCommandInPodOpts(f, c, ns, opt)
+	podOpt, err := getCommandInPodOpts(f, c, ns, "", opt)
+	if err != nil {
+		return "", "", err
+	}
+	stdOut, stdErr, err := f.ExecWithOptions(podOpt)
+	if stdErr != "" {
+		e2elog.Logf("stdErr occurred: %v", stdErr)
+	}
+	return stdOut, stdErr, err
+}
+
+func execCommandInContainer(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (string, string, error) {
+	podOpt, err := getCommandInPodOpts(f, c, ns, cn, opt)
 	if err != nil {
 		return "", "", err
 	}
@@ -185,7 +220,7 @@ func execCommandInToolBoxPod(f *framework.Framework, c, ns string) (string, stri
 	opt := &metav1.ListOptions{
 		LabelSelector: rookToolBoxPodLabel,
 	}
-	podOpt, err := getCommandInPodOpts(f, c, ns, opt)
+	podOpt, err := getCommandInPodOpts(f, c, ns, "", opt)
 	if err != nil {
 		return "", "", err
 	}
@@ -197,7 +232,7 @@ func execCommandInToolBoxPod(f *framework.Framework, c, ns string) (string, stri
 }
 
 func execCommandInPodAndAllowFail(f *framework.Framework, c, ns string, opt *metav1.ListOptions) (string, string) {
-	podOpt, err := getCommandInPodOpts(f, c, ns, opt)
+	podOpt, err := getCommandInPodOpts(f, c, ns, "", opt)
 	if err != nil {
 		return "", err.Error()
 	}
--- a/e2e/pod.go
+++ b/e2e/pod.go
@@ -204,7 +204,7 @@ func execCommandInPod(f *framework.Framework, c, ns string, opt *metav1.ListOpti
 	return stdOut, stdErr, err
 }
 
-func execCommandInContainer(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (string, string, error) {
+func execCommandInContainer(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (string, string, error) { //nolint:unparam,lll // cn can be used with different inputs later
 	podOpt, err := getCommandInPodOpts(f, c, ns, cn, opt)
 	if err != nil {
 		return "", "", err
@@ -304,7 +304,7 @@ func deletePod(name, ns string, c kubernetes.Interface, t int) error {
 	})
 }
 
-func deletePodWithLabel(label, ns string, skipNotFound bool) error {
+func deletePodWithLabel(label, ns string, skipNotFound bool) error { //nolint:unparam // skipNotFound can be used with different inputs later
 	_, err := framework.RunKubectl(ns, "delete", "po", "-l", label, fmt.Sprintf("--ignore-not-found=%t", skipNotFound))
 	if err != nil {
 		e2elog.Logf("failed to delete pod %v", err)
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -374,6 +374,92 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("perform IO on rbd-nbd volume after nodeplugin restart and expect a failure", func() {
+				err := deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				// Storage class with rbd-nbd mounter
+				err = createRBDStorageClass(f.ClientSet, f, nil, map[string]string{"mounter": "rbd-nbd"}, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+				pvc, err := loadPVC(pvcPath)
+				if err != nil {
+					e2elog.Failf("failed to load PVC with error %v", err)
+				}
+				pvc.Namespace = f.UniqueName
+
+				app, err := loadApp(appPath)
+				if err != nil {
+					e2elog.Failf("failed to load application with error %v", err)
+				}
+
+				app.Namespace = f.UniqueName
+				label := map[string]string{
+					"app": app.Name,
+				}
+				app.Labels = label
+				app.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = pvc.Name
+				err = createPVCAndApp("", f, pvc, app, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create PVC and application with error %v", err)
+				}
+
+				// validate created backend rbd images
+				validateRBDImageCount(f, 1)
+
+				selector, err := getDaemonSetLabelSelector(f, cephCSINamespace, rbdDaemonsetName)
+				if err != nil {
+					e2elog.Failf("failed to get the labels with error %v", err)
+				}
+				// delete rbd nodeplugin pods
+				err = deletePodWithLabel(selector, cephCSINamespace, false)
+				if err != nil {
+					e2elog.Failf("fail to delete pod with error %v", err)
+				}
+
+				// wait for nodeplugin pods to come up
+				err = waitForDaemonSets(rbdDaemonsetName, cephCSINamespace, f.ClientSet, deployTimeout)
+				if err != nil {
+					e2elog.Failf("timeout waiting for daemonset pods with error %v", err)
+				}
+
+				opt := metav1.ListOptions{
+					LabelSelector: fmt.Sprintf("app=%s", app.Name),
+				}
+
+				// FIXME: Fix this behavior, i.e. when the nodeplugin is
+				// restarted, the rbd-nbd processes should be back to life
+				// as rbd-nbd processes are responsible for IO
+
+				// For now to prove this isn't working, write something to
+				// mountpoint and expect a failure as the processes are terminated.
+				filePath := app.Spec.Containers[0].VolumeMounts[0].MountPath + "/test"
+				_, stdErr := execCommandInPodAndAllowFail(f, fmt.Sprintf("echo 'Hello World' > %s", filePath), app.Namespace, &opt)
+				IOErr := fmt.Sprintf("cannot create %s: Input/output error", filePath)
+				if !strings.Contains(stdErr, IOErr) {
+					e2elog.Failf(stdErr)
+				} else {
+					e2elog.Logf("failed IO as expected: %v", stdErr)
+				}
+
+				err = deletePVCAndApp("", f, pvc, app)
+				if err != nil {
+					e2elog.Failf("failed to delete PVC and application with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0)
+				err = deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				err = createRBDStorageClass(f.ClientSet, f, nil, nil, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+			})
+
 			By("create a PVC and bind it to an app with encrypted RBD volume", func() {
 				err := deleteResource(rbdExamplePath + "storageclass.yaml")
 				if err != nil {
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"regexp"
 	"strings"
 
 	. "github.com/onsi/ginkgo" // nolint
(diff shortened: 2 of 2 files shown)


pkalever:
Depends on: https://github.com/kubernetes/minikube/pull/10217

nixpanic:
You may want to have a look at #1840 for an idea how to test it in the CI before the minikube PR is merged+released.

pkalever:
@nixpanic I hope the CI picks the modifications to minikube iso URL as part of this PR testing, is that right?
Or CI will only pick once the changes are merged?

nixpanic:
> @nixpanic I hope the CI picks the modifications to minikube iso URL as part of this PR testing, is that right?
> Or CI will only pick once the changes are merged?

The CI uses `scripts/minikube.sh` from the PR that modifies it, so `--iso-url` will be included in this testing.

nixpanic:
Oh, note that #1831 contains a modification to the `EXTRA_CONFIG` options for minikube, done right. CI jobs will likely fail with the change you made here now.

pkalever:
> Oh, note that #1831 contains a modification to the `EXTRA_CONFIG` options for minikube, done right. CI jobs will likely fail with the change you made here now.

The PR#1831 seems to be in open state now, when it gets merged I shall rebase this, hope that looks like a good plan.
And yeah! while PR#1831 is still in the open state, I don't see a reason why this should fail?

Thanks!

pkalever:
@nixpanic @Madhu-1 @Yuggupta27 any idea why the CI is failing at 'deploy k8s-1.xx and rook' phase?

Thanks!

Madhu-1:
Looks to be a PSP issue https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.19/detail/mini-e2e-helm_k8s-1.19/567/pipeline

pkalever:
> Looks to be a PSP issue https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.19/detail/mini-e2e-helm_k8s-1.19/567/pipeline

Thanks @Madhu-1, now I understand why @nixpanic was warning in the comments above.
At least until https://github.com/ceph/ceph-csi/pull/1831/commits/73d9428760ac2cdcda7979bc871e8e90639be23a is merged, I should copy them to my PR, just to comfort CI.

pkalever:
@nixpanic @Madhu-1 @Yuggupta27 Yay! and the CI finally made it, thanks guys!

nixpanic:
With #1811 the `minikube start` commands in `scripts/minikube.sh` have been adapted. There is an additional `--cni` option now. You'll need to manually rebase this.

e2e/pod.go
```diff
@@ -88,21 +88,43 @@ func waitForDeploymentComplete(name, ns string, c kubernetes.Interface, t int) e
...
+func getCommandInPodOpts(f *framework.Framework, c, ns string, cn string, opt *metav1.ListOptions) (framework.ExecOptions, error) {
+	e2elog.Logf("In getCommandInPodOpts: %v", cn)
+	pn := ""
 	cmd := []string{"/bin/sh", "-c", c}
 	podList, err := f.PodClientNS(ns).List(context.TODO(), *opt)
 	framework.ExpectNoError(err)
 	if len(podList.Items) == 0 {
 		return framework.ExecOptions{}, errors.New("podlist is empty")
 	}
+	found := false
```
humblec: find Container Name in the pod can be wrapped into a new function .

e2e/pod.go
```diff
@@ -111,7 +133,19 @@ func getCommandInPodOpts(f *framework.Framework, c, ns string, opt *metav1.ListO
...
+	}
+	stdOut, stdErr, err := f.ExecWithOptions(podPot)
+	if stdErr != "" {
+		e2elog.Logf("stdErr occurred: %v", stdErr)
+	}
+	return stdOut, stdErr, err
+}
+
+func execCommandInContainer(f *framework.Framework, c, ns string, cn string, opt *metav1.ListOptions) (string, string, error) {
+	podPot, err := getCommandInPodOpts(f, c, ns, cn, opt)
```
humblec: eventhough its not introduced here, Is above (`podPot`) a weird name ? may be we have to rename it as `podOpt` ?

e2e/pod.go
pkalever: Done now!

e2e/pod.go
pkalever: addressed this one too. Thanks!

e2e/pod.go
humblec: Thanks @pkalever :+1:

e2e/pod.go
humblec: Cool. thanks !

e2e/rbd.go
```diff
@@ -334,6 +335,229 @@ var _ = Describe("RBD", func() {
...
+				// # ps -eo 'cmd' | grep [r]bd-nbd
+				// /usr/bin/rbd-nbd --id cephcsi-rbd-node -m svc-name:6789 --keyfile=/tmp/csi/keys/keyfile attach --device /dev/nbd0 pool-name/image-name --try-netlink --reattach-timeout=180
+				opt1 := metav1.ListOptions{
+					LabelSelector: "app in (ceph-csi-rbd, csi-rbdplugin)",
+				}
+				uname, stdErr, err := execCommandInContainer(f, "uname -a", cephCSINamespace, "csi-rbdplugin", &opt1)
+				if err != nil || stdErr != "" {
+					e2elog.Failf("failed to run uname cmd : %v, stdErr: %v ", err, stdErr)
+				}
+				e2elog.Logf("uname 1: %v", uname)
```
humblec: @pkalever instead of `1,2,..etc` may be we can log other marker?

pkalever: Humble, right now, the helm chart e2e tests are failing for rbd-nbd, so I wanted to debug, hence added some extra debug logs. These logs will not be part of the original patch and I shall remove them once everything runs smooth.

Thanks for keeping an eye on this PR!

e2e/rbd.go
```diff
@@ -427,6 +428,136 @@ var _ = Describe("RBD", func() {
...
+				app.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = pvc.Name
+				err = createPVCAndApp("", f, pvc, app, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create PVC and application with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 1)
+
+				// Get the label name of daemonset dynamically, as the label name is not same for helm and non-helm deploys
+				ds, err := f.ClientSet.AppsV1().DaemonSets(cephCSINamespace).Get(context.TODO(), rbdDaemonsetName, metav1.GetOptions{})
```
pkalever: @nixpanic @Madhu-1 JFYI - this way I'm now able to get the label dynamically.

Madhu-1: Thanks @pkalever

Madhu-1: Use this label selector to delete the daemonset pods also?

pkalever: New line number 509?
I'm doing the same right?
'm I missing anything?

Madhu-1: i was referring to https://github.com/ceph/ceph-csi/pull/1839/files#diff-bfe7fe7ada8a43a75c21ddfd9535e10f37d6e55e9a3006ce5a1872f73f3ae198R399-R402

pkalever: Hmm that was part of a different test scenario. But it's a good idea to keep this logic in the utils and reuse it where ever needed. I will fix it, thanks!

pkalever: @Madhu-1 done now. Thanks!

pkalever:
@nixpanic @Madhu-1 please remove the DNM label, and also help review and merge this PR.
The checklist is now ready.

Thanks!

build.env
```diff
@@ -43,7 +43,7 @@ CHANGE_MINIKUBE_NONE_USER=true
 # Rook options
 ROOK_VERSION=v1.4.9
 # Provide ceph image path
-ROOK_CEPH_CLUSTER_IMAGE=docker.io/ceph/ceph:v14.2.12
+ROOK_CEPH_CLUSTER_IMAGE=docker.io/ceph/ceph:v16
```
nixpanic: duplicate commit of #2038 ?

nixpanic (commented):
it seems golang-ci fails, you can run this locally with `make containerized-test` to verify your fix(es)

build.env
pkalever: Yes and this is intentional for satisfying the dependency for this PR at CI. I will remove it once the other PR is merged.

pkalever:
> it seems golang-ci fails, you can run this locally with `make containerized-test` to verify your fix(es)

@nixpanic Is there a way to ignore these errors, the errors don't look like a great deal to me, I think we can safely ignore them for now.

e2e/pod.go
```diff
@@ -105,17 +105,49 @@ func waitForDeploymentComplete(name, ns string, c kubernetes.Interface, t int) e
 	return nil
 }
 
-func getCommandInPodOpts(f *framework.Framework, c, ns string, opt *metav1.ListOptions) (framework.ExecOptions, error) {
+func findContainerName(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (string, string, error) {
+	pn := ""
+	podList, err := f.PodClientNS(ns).List(context.TODO(), *opt)
+	framework.ExpectNoError(err)
```
Madhu-1: return error?

e2e/pod.go
```diff
@@ -105,17 +105,49 @@ func waitForDeploymentComplete(name, ns string, c kubernetes.Interface, t int) e
 	return nil
 }
 
-func getCommandInPodOpts(f *framework.Framework, c, ns string, opt *metav1.ListOptions) (framework.ExecOptions, error) {
+func findContainerName(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (string, string, error) {
```
Madhu-1: the function name should be changed you are finding both pod name and container name here?

e2e/pod.go
```diff
@@ -105,17 +105,49 @@ func waitForDeploymentComplete(name, ns string, c kubernetes.Interface, t int) e
...
+	pn := ""
+	podList, err := f.PodClientNS(ns).List(context.TODO(), *opt)
+	framework.ExpectNoError(err)
+	if len(podList.Items) == 0 {
+		return "", "", errors.New("podlist is empty")
+	}
+	found := false
+	if cn != "" {
+		for i := range podList.Items {
```
Madhu-1, line 119: found is only using if the `cn` is not empty can we move found:=false inside if check?

e2e/pod.go
```diff
@@ -105,17 +105,49 @@ func waitForDeploymentComplete(name, ns string, c kubernetes.Interface, t int) e
...
+			for j := range podList.Items[i].Spec.Containers {
+				if podList.Items[i].Spec.Containers[j].Name == cn {
+					pn = podList.Items[i].Name
+					found = true
+					break
```
Madhu-1: if found  return from here? so that we can avoid the new variable `found` and also an if check?

e2e/rbd.go
```diff
@@ -349,6 +350,228 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("create a PVC and bind it to an app using rbd-nbd mounter", func() {
```
Madhu-1, line 353: can you please enable this E2E also https://github.com/ceph/ceph-csi/blob/devel/e2e/rbd.go#L485-L496?

e2e/rbd.go
```diff
@@ -349,6 +350,228 @@ var _ = Describe("RBD", func() {
...
+				}
+				pvc, err := loadPVC(pvcPath)
+				if err != nil {
+					e2elog.Failf("failed to load PVC with error %v", err)
+				}
+				pvc.Namespace = f.UniqueName
+
+				app, err := loadApp(appPath)
+				if err != nil {
+					e2elog.Failf("failed to  load application with error %v", err)
```
Madhu-1:
```suggestion
e2elog.Failf("failed to load application with error %v", err)
```

e2e/rbd.go
```diff
@@ -349,6 +350,228 @@ var _ = Describe("RBD", func() {
...
+					e2elog.Failf("timeout waiting for daemonset pods with error %v", err)
+				}
+
+				opt := metav1.ListOptions{
+					LabelSelector: fmt.Sprintf("app=%s", app.Name),
+				}
+				// Write something to mountpoint and expect a failure
+				filePath := app.Spec.Containers[0].VolumeMounts[0].MountPath + "/test"
+				_, stdErr := execCommandInPodAndAllowFail(f, fmt.Sprintf("echo 'Hello World' > %s", filePath), app.Namespace, &opt)
+				IOErr := fmt.Sprintf("cannot create %s: Input/output error", filePath)
```
Madhu-1, line 441: Add a FIXME to fix this one as this wont be true when we implement logic to handle daemonset pod restart?

e2e/rbd.go
```diff
@@ -349,6 +350,228 @@ var _ = Describe("RBD", func() {
...
+				} else {
+					e2elog.Logf("failed IO as expected: %v", stdErr)
+				}
+
+				err = deletePVCAndApp("", f, pvc, app)
+				if err != nil {
+					e2elog.Failf("failed to delete PVC and application with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0)
```
Madhu-1, line 453: delete the storageclass here as this is specific to `nbr` adding a test for `krbd` in between might get failed because of this one? and recreate it in next By if you want?

e2e/rbd.go
```diff
@@ -349,6 +350,228 @@ var _ = Describe("RBD", func() {
...
+			})
+
+			By("restart rbd-nbd process on nodeplugin and continue IO after nodeplugin restart", func() {
```
Madhu-1, line 464: Just a question Is this whole test case required? as we are doing some hacks to remount the rbd image which is not the case in the real world and this is expected to handled automatically by cephcsi or other sidecars in daemonset pod?

e2e/pod.go
pkalever: May be findPodAndContainerName() ?

e2e/pod.go
pkalever: This is an assertion, we are returning at line 113

e2e/pod.go
pkalever: Lol, you are right. I think I was called on a Mars mission when I was writing this piece of code.

e2e/rbd.go
pkalever: sure

e2e/rbd.go
pkalever: The subject of the tests remains the same after the automation as well, just that we might remove the lines which are trying to manually bring the process online and override them with the steps to bring the process back to online in the background.

In anyway having this in the CI till then will help us understand the regressions and true failures within rbd-nbd.

e2e/rbd.go
pkalever: done

e2e/rbd.go
pkalever: sure

e2e/rbd.go
pkalever: Cool

pkalever (commented):
Thanks @Madhu-1

e2e/pod.go
Madhu-1: what I meant was as we are returning an error from this function, let's return an error from here and the caller handles it. let's not do both assertion and error returning from the same functions

e2e/pod.go
Madhu-1: Sounds good.

e2e/rbd.go
Madhu-1: currently this PR adds two tests  `perform IO on rbd-nbd volume after nodeplugin restart and expect a failure` and `restart rbd-nbd process on nodeplugin and continue IO after nodeplugin restart`  but when we bring the automatic handling of the restart, one test will be removed and other will be modified to check we are able to read/write the data? IMO having single test seems fine for me. Lets hear from others

pkalever: These are the POC level tests, for someone who is new to rbd-nbd mounter can understand the problem we are dealing here very clearly with the help of these test cases.

In the future when the real story is working we can clear these bits and makeup more useful test cases.

Till then the intention of these tests is to catch some regressions or failures at rbd-nbd only.

e2e/pod.go
```diff
@@ -105,17 +105,40 @@ func waitForDeploymentComplete(name, ns string, c kubernetes.Interface, t int) e
...
+					pn = podList.Items[i].Name
+					return pn, cn, nil
+				}
+			}
+		}
+		return "", "", errors.New("container name not found")
+	}
+	pn = podList.Items[0].Name
+	cn = podList.Items[0].Spec.Containers[0].Name
+	return pn, cn, nil
```
Madhu-1: just return podList.Items[0].Name and podList.Items[0].Spec.Containers[0].Name directly here and above

e2e/pod.go
Madhu-1: @pkalever any comment on this one?

Madhu-1 (commented):
don't have a major comment except for https://github.com/ceph/ceph-csi/pull/1839/files#r624912403 if others are okay we can get it in.

e2e/pod.go
pkalever: Sure.

e2e/pod.go
pkalever: Ah! Got you, I will remove this line, thanks for the clarification.

e2e/pod.go
```diff
@@ -305,7 +305,7 @@ func deletePod(name, ns string, c kubernetes.Interface, t int) error {
 	})
 }
 
-func deletePodWithLabel(label, ns string, skipNotFound bool) error {
+func deletePodWithLabel(label, ns string, skipNotFound bool) error { //nolint:unparam
```
pkalever: @nixpanic this should fix the golint warnings!

pkalever:
@Madhu-1 @nixpanic As we don't see any objections, could you please merge this PR now?

Thanks!

humblec:
@pkalever one quick check: Have we tested this scenario when the actual node reboot or move to `NOT READY` state and come back after some time ?

pkalever:
> @pkalever one quick check: Have we tested this scenario when the actual node reboot or move to `NOT READY` state and come back after some time ?

When a node is rebooted, the application pod might be migrated to a new node or might be scheduled on the same node, in any case, a new NodeStageVolume call is expected to take care of mapping the device.

These tests are mainly focusing on the restart of the node plugin, where there is no NodeStageVolume call for now. We are working on implementing the missing gaps in parallel, where a new init container or a sidecar is supposed to get the list of volume attachments on a node (upon the restart of the node plugin) and make new NodeStageVolume grpc Calls per device for reattaching (remapping) rbd-nbd processes/devices.

Thanks!

humblec:
> > @pkalever one quick check: Have we tested this scenario when the actual node reboot or move to `NOT READY` state and come back after some time ?
> 
> When a node is rebooted, the application pod might be migrated to a new node or might be scheduled on the same node, in any case, a new NodeStageVolume call is expected to take care of mapping the device.

There are some ( corner ) cases, like network error, kubelet goes wrong, node completely goes out..etc which are bit different from volume pov. We have also seen issues popping up with some race conditions from CO to detect and issues RPC calls in between..etc.  The reason for asking node `not ready` scenario testing  was to understand/make sure we are not landing into issues easily when we remount the shares with these options.  So I think its good to check once if we have a setup :)

pkalever:
> There are some ( corner ) cases, like network error, kubelet goes wrong, node completely goes out..etc which are bit different from volume pov. We have also seen issues popping up with some race conditions from CO to detect and issues RPC calls in between..etc. The reason for asking node `not ready` scenario testing was to understand/make sure we are not landing into issues easily when we remount the shares with these options. So I think its good to check once if we have a setup :)

Thanks Humble, Got you. These test cases are less focused on the corner case. But I will make sure to test these corner cases with the logic we are implementing to address node plugin restart automatically.

pkalever:
@Madhu-1 @nixpanic can you please help on taking this PR to merge?
I would like to see them running in the CI which will help us improve the confidence on rbd-nbd.

Thanks!

Madhu-1 (approved):

e2e/pod.go
```diff
@@ -204,7 +204,7 @@ func execCommandInPod(f *framework.Framework, c, ns string, opt *metav1.ListOpti
 	return stdOut, stdErr, err
 }
 
-func execCommandInContainer(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (string, string, error) {
+func execCommandInContainer(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (string, string, error) { //nolint:unparam,lll // cn can be used with different inputs later
```
nixpanic, line 207: We usually put these comments above the line that is affected, not after the line (this line becomes *really* long now).

nixpanic (approved):
It is good to have this tested regularly, so let's get it in.

In case a manual rebase is needed, please correct the formatting of the `// nolint:` comments.

e2e/pod.go
pkalever: I thought putting it above the line will affect the whole function block not just that line in this case?

pkalever:
> In case a manual rebase is needed, please correct the formatting of the `// nolint:` comments.

The recommendation was not to have space in between the slash and `nolint`

e2e/pod.go
nixpanic: Probably, but that is not really an issue.

In general, all `nolint` issues need to be addressed at one point. That means, either the argument is not needed and can be dropped, or the function gets used with other arguments somewhere.

Madhu-1:
@Mergifyio rebase

PR #1844, opened 2021-01-28
rbd: fix issue in ENV variable check

Madhu-1:
Currently, cephcsi is returning an issue if the ENV variable is set, but it should not. This commit fixes the POD_NAMESPACE env variable issue and as well as the KMS_CONFIG_NAME ENV variable.

### DIFF
--- a/internal/util/crypto.go
+++ b/internal/util/crypto.go
@@ -133,11 +133,11 @@ func GetKMS(tenant, kmsID string, secrets map[string]string) (EncryptionKMS, err
 		// If the configmap is not mounted to the CSI pods read the configmap
 		// the kubernetes.
 		namespace := os.Getenv(podNamespace)
-		if namespace != "" {
+		if namespace == "" {
 			return nil, fmt.Errorf("%q is not set", podNamespace)
 		}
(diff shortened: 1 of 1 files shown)


nixpanic (approved):

Madhu-1:
@nixpanic looks like something is wrong with go mod. it's failing with different errors. I rebased https://github.com/ceph/ceph-csi/pull/1675 to check mod error exists only on this PR or in the other PRs also.

nixpanic:
> @nixpanic looks like something is wrong with go mod. it's failing with different errors. I rebased #1675 to check mod error exists only on this PR or in the other PRs also.

This also happens when there is an issue with the repository where the module comes from. Maybe a GitHub glitch?

nixpanic:
> > @nixpanic looks like something is wrong with go mod. it's failing with different errors. I rebased #1675 to check mod error exists only on this PR or in the other PRs also.
> 
> This also happens when there is an issue with the repository where the module comes from. Maybe a GitHub glitch?

The repositories that can not be fetched are different every time. There really seems to be something going on at GitHub, but it is not on their status page :cry:

Yuggupta27 (approved):

Issue #1849, opened 2021-02-01
e2e: add e2e for retain-snapshots feature in cephfs subvolumes


agarwal-mudit:
# Describe the bug #

There is no e2e to test the retain-snapshots feature in cephFS subvolumes.

Add a test case for Bytes Quota case also #1850

agarwal-mudit:
@Madhu-1 Please assign it to me.

agarwal-mudit:
Right now this can not be added because of two reasons:
1. Ceph version: We use upstream ceph version which doesn't support this feature.
2.  #1845

agarwal-mudit:
Already fixed with https://github.com/ceph/ceph-csi/pull/1870

PR #1850, opened 2021-02-01
cephfs: ignore BytesQuota field in case it is not set.

agarwal-mudit:
This can happen when the subvolume is in snapshot-retained state.
We should not return error for such case as it is a valid situation.


nixpanic (approved):

Madhu-1 (commented):
is it possible to have a test case to cover this scenario?

agarwal-mudit:
> is it possible to have a test case to cover this scenario?

Can be done along with https://github.com/ceph/ceph-csi/issues/1849, can't be done before that.

Madhu-1 (approved):

PR #1736, opened 2020-11-30
helm: make kubelet path configurable

Madhu-1:
removed unwanted extra arguments from the helm templates and added a single value kubeletDir to make the kubelet
root-dir configurable.

fixes: #1475

PR #1855, opened 2021-02-02
helm: Fix plugin socket path

mtneug:
PR #1736 made the kubelet path configurable. It also introduced a change in the path to the CSI socket. By default the path is now `/var/lib/kubelet/cephfs.csi.ceph.com/csi.sock` instead of `/var/lib/kubelet/plugins/cephfs.csi.ceph.com/csi.sock`. This PR restores the old default.

### DIFF
--- a/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
+++ b/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
@@ -43,7 +43,7 @@ spec:
           args:
             - "--v={{ .Values.logLevel }}"
             - "--csi-address=/csi/{{ .Values.pluginSocketFile }}"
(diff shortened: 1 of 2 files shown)


Madhu-1 (commented):
@mtneug Thanks for the fix. Can you please take a look at https://github.com/kubernetes-csi/node-driver-registrar#usage  and see the current change matches the recommandation.

mtneug:
@Madhu-1 the CSI driver socket should be placed in `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock`. This PR restores this behaviour (`plugins` directory was missing since #1736).

Madhu-1 (approved):

Yuggupta27 (approved):

Issue #1874, opened 2021-02-17
e2e testing occasionally fails to detect deleted PVC/PV objects


nixpanic:
# Describe the bug #

Sometimes the CI jobs error out while waiting for a PVC/PV to be deleted. This happens for CephFS and for RBD based volumes.

CephFS:
```
Feb 16 17:50:30.882: INFO: waiting for PVC cephfs-6940 in state &PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},} to be deleted (598 seconds elapsed)
Feb 16 17:50:32.868: INFO: waiting for PVC cephfs-6942 in state &PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},} to be deleted (600 seconds elapsed)
Feb 16 17:50:32.874: INFO: waiting for PVC cephfs-6942 in state &PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},} to be deleted (600 seconds elapsed)
Feb 16 17:50:32.882: INFO: waiting for PVC cephfs-6940 in state &PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},} to be deleted (600 seconds elapsed)
Feb 16 17:50:32.981: INFO: waiting for PVC cephfs-6940 in state &PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},} to be deleted (600 seconds elapsed)
Feb 16 17:50:33.000: INFO: failed to delete PVC and app (cephfs-6940): timed out waiting for the condition
Feb 16 17:50:33.000: INFO: failed to delete PVC and app (cephfs-6941): timed out waiting for the condition
Feb 16 17:50:33.000: INFO: failed to delete PVC and app (cephfs-6942): timed out waiting for the condition
Feb 16 17:50:33.000: FAIL: deleting PVCs and apps failed, 3 errors were logged
```

RBD:
```
[...]

PR #1876, opened 2021-02-17
e2e: track deletion of PVC and PV more closely

nixpanic:
When passing a pointer to a PVC and PV, the status of the deleted
objects is not logged correctly. The `PersistentVolumeClaim.Status` and
`PersistedVolume.Status` that is added to the logs contain the status of
the initially created object (reference to the PVC/PV). When the PVC/PV
is removed, there is no guarantee that the object is updated.

Logs show an empty (nullified) `PersistentVolumeClaim.Status`, which is
not helpful. Instead, use the returned PVC/PV from the `Get()` function
and use that for further logging. Even when the `.Status` struct from
the PVC/PV gets wiped, the returned object should have correct details.

Updates: #1874

---

<details>

  progress or results

</details>

### DIFF
--- a/e2e/pvc.go
+++ b/e2e/pvc.go
@@ -94,10 +94,12 @@ func deletePVCAndPV(c kubernetes.Interface, pvc *v1.PersistentVolumeClaim, pv *v
 
 	timeout := time.Duration(t) * time.Minute
 	start := time.Now()
+
+	pvcToDelete := pvc
 	err = wait.PollImmediate(poll, timeout, func() (bool, error) {
 		// Check that the PVC is deleted.
-		e2elog.Logf("waiting for PVC %s in state %s to be deleted (%d seconds elapsed)", pvc.Name, pvc.Status.String(), int(time.Since(start).Seconds()))
-		_, err = c.CoreV1().PersistentVolumeClaims(pvc.Namespace).Get(context.TODO(), pvc.Name, metav1.GetOptions{})
+		e2elog.Logf("waiting for PVC %s in state %s to be deleted (%d seconds elapsed)", pvcToDelete.Name, pvcToDelete.Status.String(), int(time.Since(start).Seconds()))
+		pvcToDelete, err = c.CoreV1().PersistentVolumeClaims(pvcToDelete.Namespace).Get(context.TODO(), pvcToDelete.Name, metav1.GetOptions{})
 		if err == nil {
 			return false, nil
 		}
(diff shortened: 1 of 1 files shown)


e2e/pvc.go
```diff
@@ -94,34 +94,38 @@ func deletePVCAndPV(c kubernetes.Interface, pvc *v1.PersistentVolumeClaim, pv *v
...
-		e2elog.Logf("waiting for PVC %s in state %s to be deleted (%d seconds elapsed)", pvc.Name, pvc.Status.String(), int(time.Since(start).Seconds()))
-		_, err = c.CoreV1().PersistentVolumeClaims(pvc.Namespace).Get(context.TODO(), pvc.Name, metav1.GetOptions{})
+		e2elog.Logf("waiting for PVC %s in state %s to be deleted (%d seconds elapsed)", pvcToDelete.Name, pvcToDelete.Status.String(), int(time.Since(start).Seconds()))
+		pvcToDelete, err = c.CoreV1().PersistentVolumeClaims(pvcToDelete.Namespace).Get(context.TODO(), pvcToDelete.Name, metav1.GetOptions{})
 		if err == nil {
 			return false, nil
 		}
 		if !apierrs.IsNotFound(err) {
-			return false, fmt.Errorf("get on deleted PVC %v failed with error other than \"not found\": %w", pvc.Name, err)
+			return false, fmt.Errorf("get on deleted PVC %v failed with error other than \"not found\": %w", pvcToDelete.Name, err)
```
Madhu-1, line 107: as we are doing get above at line 102, in case of the error the `pvcToDelete.Name` will be empty here?

e2e/pvc.go
```diff
@@ -94,34 +94,38 @@ func deletePVCAndPV(c kubernetes.Interface, pvc *v1.PersistentVolumeClaim, pv *v
...
 
-		_, err = c.CoreV1().PersistentVolumes().Get(context.TODO(), pv.Name, metav1.GetOptions{})
+		pvToDelete, err = c.CoreV1().PersistentVolumes().Get(context.TODO(), pvToDelete.Name, metav1.GetOptions{})
 		if err == nil {
 			return false, nil
 		}
 
 		if !apierrs.IsNotFound(err) {
-			return false, fmt.Errorf("delete PV %v failed with error other than \"not found\": %w", pv.Name, err)
+			return false, fmt.Errorf("delete PV %v failed with error other than \"not found\": %w", pvToDelete.Name, err)
```
Madhu-1, line 128: same here also. incase the error is notFound the pvToDelete object might be nil here?

e2e/pvc.go
nixpanic: Yes, possibly. Or `pvcToDelete` might be `nil`. Either way, it will cause a failure. I opted for a potential panic, but have just reconsidered this and will use `pvc.Name` instead.

e2e/pvc.go
nixpanic: same here, will use `pv.Name` instead

Madhu-1 (approved):

Yuggupta27 (approved):
Thanks for the PR!

humblec (approved):

PR #1877, opened 2021-02-17
rbd: add support for async mirroring

Madhu-1:
we have developed a new volume replication protobuf and [specification](https://github.com/kube-storage/spec) to achieve the Volume replication.

cephcsi will implement the required GRPC services for volume replication. below is the list of API's are implemented in this PR.

* EnableVolumeReplication
* DisableVolumeReplication
* PromoteVolume
* DemoteVolume
* ResyncVolume

A new sidecar controller will be deployed as part of the RBD provisioner pod which will expose the CRD to a user to interact with the Ceph cluster for DR operations.

when a User creates a CR with the PVC name, the new operator will get the required pvc and PV information and send a request to the cephcsi to perform the rbd async operation.

All the above operations are done on the volume handle present in the PV spec. cephcsi is intelligent enough then extract the required information
to get the exact rbd image it has to operate on.

TODO:-

- [ ]  Add usage documentation for RBD async mirroring.
- [ ]  Add documentation for new sidecar deployment.

Above TODO will be done in follow-up PR's.


Madhu-1:
requesting for a review of the initial approach.

Madhu-1:
@dillaman can you please review this one?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	mirroringInfo, err := rbdVol.getImageMirroringInfo(ctx)
+	if err != nil {
+		err = fmt.Errorf("failed to get mirroring info of %s with error %s", rbdVol.String(), err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if mirroringInfo.State == librbd.MirrorImageEnabled {
+		err = rbdVol.disableImageMirroring(ctx, force)
```
ShyamsundarR: Unsure if image is secondary if mirroring can be disabled, may require an additional check  in that regard.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,349 @@
...
+
+// PromoteVolume promotes the volume as primary.
+func (cs *ControllerServer) PromoteVolume(ctx context.Context,
+	req *replication.PromoteVolumeRequest,
+) (*replication.PromoteVolumeResponse, error) {
+	var rbdVol = &rbdVolume{}
+	var err error
+	// ger rbdVolume from VolumeID
+	rbdVol, err = cs.getVolumeFromID(ctx, req.GetVolumeId(), req.GetSecrets())
+	defer cs.cleanup(rbdVol)
```
ShyamsundarR: Protect against parallel operations, because we can, and possibly safer to let one operation on the volume through at a time.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if mirroringInfo.State != librbd.MirrorImageEnabled {
+		return nil, status.Errorf(codes.InvalidArgument, "image is in %d Mode", mirroringInfo.State)
+	}
+
+	// promote secondary to primary or force promote image if its already
+	// primary
+	if !mirroringInfo.Primary || force && mirroringInfo.Primary {
```
ShyamsundarR: I thought we need to force promote in rbd only if the image was not already primary, IOW if if is alreayd primary then there is no promotion to be done. (which basically changes the condition here, but depends on the correctness of my understanding)

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil, status.Error(codes.InvalidArgument, "image is in primary state")
+	}
+
+	err = rbdVol.resyncImage(ctx)
+	if err != nil {
+		util.ErrorLog(ctx, "failed to resync %s with error %s", err.Error())
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	// TODO check the image state and return its ready to use or not
+	mirrorStatus, err := rbdVol.getImageMirroingStatus()
```
ShyamsundarR, line 334: Once we hit resync, isn't there a brief point in time where the image is recreated, IOW missing on the pool for some time till initial image metadata is recreated? The subsequent call may hence get a false negative if so.

ShyamsundarR (commented):
First pass comments, I do not see any surprises.

@dillaman if you could focus on files `internal/rbd/driver.go` and `internal/rbd/mirror.go` it maybe sufficient for now (if you have not already started and filtered it down that is :) )

internal/rbd/replicationcontrollerserver.go
Madhu-1: possible to disable with force option from CLI. added a check now

internal/rbd/replicationcontrollerserver.go
Madhu-1: the lock is taken in the getVolumeFromID function and will be released on the cleanup defer function.  Anything missing here

internal/rbd/replicationcontrollerserver.go
Madhu-1: Yes, you are correct. even with force, the promotion fails if the image is already primary. removed the check now.

internal/rbd/replicationcontrollerserver.go
Madhu-1: yes with some delay the image gets recreated after issuing the resync command. added check for the image not found error to return the image as not ready.

Madhu-1:
> First pass comments, I do not see any surprises.
> 
> @dillaman if you could focus on files `internal/rbd/driver.go` and `internal/rbd/mirror.go` it maybe sufficient for now (if you have not already started and filtered it down that is :) )

@ShyamsundarR `internal/rbd/driver.go` or `internal/rbd/replicationcontrollerserver.go` ?

ShyamsundarR:
> > First pass comments, I do not see any surprises.
> > @dillaman if you could focus on files `internal/rbd/driver.go` and `internal/rbd/mirror.go` it maybe sufficient for now (if you have not already started and filtered it down that is :) )
> 
> @ShyamsundarR `internal/rbd/driver.go` or `internal/rbd/replicationcontrollerserver.go` ?

I looked over them (again first pass), they setup another service endpoint, and was not that bothered w.r.t functionality. Was just **attempting** to help out @dillaman with the core files among the 143 odd files in there (although most are vendor files)

internal/rbd/replicationcontrollerserver.go
ShyamsundarR: Ack! did not look into that function, if that is done then we are good w.r.t serialization of requests.

internal/rbd/replicationcontrollerserver.go
ShyamsundarR: From a use case perspective, I am thinking something like so,

- Provide ability to delete volume when it is secondary, by actually not deleting anything (as the mirroring would recreate the same), IOW fake the result as success as long as image is secondary
- This prevents access via VolRep CR as PV is not longer available (as well as PVC)
- Hence, DisableMirroring would also act to see if it is secondary and if so ignore the request instead of forcing it

The other way to go is to force disable, the assumption (that we need to test) is that the mirroring daemon does not bring the image back due to primary existing and active elsewhere. (also curious how this state is maintained on the secondary)

Madhu-1:
@nixpanic @humblec PTAL

internal/cephfs/driver.go
```diff
@@ -152,7 +152,13 @@ func (fs *Driver) Run(conf *util.Config) {
 	}
 
 	server := csicommon.NewNonBlockingGRPCServer()
-	server.Start(conf.Endpoint, conf.HistogramOption, fs.is, fs.cs, fs.ns, conf.EnableGRPCMetrics)
+	srv := csicommon.Servers{
```
humblec, line 155: Can we add some source code comment about the server like RS ?

internal/cephfs/driver.go
```diff
@@ -152,7 +152,13 @@ func (fs *Driver) Run(conf *util.Config) {
...
+		IS: fs.is,
+		CS: fs.cs,
+		NS: fs.ns,
+		RS: nil,
```
humblec, line 160: Is it that we are not initializing RS server by default ?

internal/rbd/mirror.go
```diff
@@ -0,0 +1,134 @@
...
+	image, err := rv.open()
+	if err != nil {
+		return err
+	}
+	defer image.Close()
+	return image.MirrorDemote()
+}
+
+func (rv *rbdVolume) resyncImage(ctx context.Context) error {
+	util.DebugLog(ctx, "rbd: resync image %s", rv.String())
```
humblec: Good to log all the debug in above functions with rv image/volume name/handle.

internal/rbd/mirror.go
```diff
@@ -0,0 +1,134 @@
...
+	image, err := rv.open()
+	if err != nil {
+		return err
+	}
+	defer image.Close()
+	return image.MirrorResync()
+}
+
+type imageMirrorStatus struct {
+	Name        string `json:"name"`
```
humblec: What is this `name` field stands for ? is it `global_id` ? may be better to document here too.

internal/rbd/mirror.go
```diff
@@ -0,0 +1,134 @@
...
+		"--id", rv.conn.Creds.ID,
+		"--keyfile="+rv.conn.Creds.KeyFile,
+		"-c", util.CephConfigPath,
+		"--format="+"json",
+		"mirror",
+		"image",
+		"status",
+		rv.String())
+	if err != nil {
+		if strings.Contains(stderr, "rbd: error opening image "+rv.RbdImageName+
```
humblec, line 141: Just a question, if the POOL does not exist, we get the same error here but with `pool` name in return instead of `rbd image` ? Isnt it ? if yes, we may fail to capture this error condition.. May be @dillaman can confirm too.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+//  imageMirroringMode is used to indicate the mirroring mode for an RBD image.
+type imageMirroringMode string
+
+const (
+	// imageMirrorModeJournal uses journaling to propagate RBD images between
+	// ceph clusters.
+	imageMirrorModeJournal imageMirroringMode = "journal"
+	// imageMirrorModeSnapshot uses snapshots to propagate RBD images between
+	// ceph clusters.
+	imageMirrorModeSnapshot imageMirroringMode = "snapshot"
```
humblec, line 38: Are we supporting both modes ? I was under the impression that, only snapshot mirroring is supported . If thats the case, may be we can stick with that here.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+	// If the state is up+stopped means the rbd-mirror daemon is
+	// running and stopped means the image is not a target for replication from
+	// another cluster
+	upAndStopped imageMirroingState = "up+stopped"
+)
+
+// mirroringMode is the key to get the imageMirroringMode from parameters.
+var imageMirroringKey = "mirroringMode"
+
+// forceKey is the key to get the force option from parameters.
```
humblec: `parameters` ( the grpc request parameters :)) in reference here can be mentioned precisely too.. so that no confusion comes in which parameters are referred here.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	return rbdVol, err
+}
+
+// getForceOption extracts the force option from the parameters.
+// If not set, the default will be set to false.
+func getForceOption(ctx context.Context, parameters map[string]string) (bool, error) {
+	val, ok := parameters[forceKey]
+	if !ok {
+		util.WarningLog(ctx, "%s is not set in parameters, setting to mirroringMode to default (%v)", forceKey, false)
```
humblec: I am bit confused with above warning. iic, this function is for `force` key and not for `mirroringmode` , Isnt it ?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+		return false, nil
+	}
+	force, err := strconv.ParseBool(val)
+	if err != nil {
+		return false, status.Errorf(codes.Internal, err.Error())
+	}
+	return force, nil
```
humblec, line 107: return `forcekey`?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if mirroringInfo.State == librbd.MirrorImageEnabled {
+		if !force && !mirroringInfo.Primary {
+			return nil, status.Error(codes.InvalidArgument, "image is in non-primary state")
+		}
+		err = rbdVol.disableImageMirroring(ctx, force)
+		if err != nil {
+			util.ErrorLog(ctx, "failed to enable mirroring of %s with error %s", err.Error())
```
humblec: This should be `failed to enable` , Isnt it ?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		// TODO check do we need to return abort after disabling the mirroring
+		return &replication.DisableVolumeReplicationResponse{}, nil
+	}
+
+	if mirroringInfo.State != librbd.MirrorImageDisabled {
```
humblec: If there was an error we would have returned already. If we reach here without error , are we expecting that still there is a condition that make the image to stay in `enabled` state?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+	return &replication.DemoteVolumeResponse{}, nil
+}
+
+// ResyncVolume resync the volume.
+func (cs *ControllerServer) ResyncVolume(ctx context.Context,
+	req *replication.ResyncVolumeRequest,
+) (*replication.ResyncVolumeResponse, error) {
+	var rbdVol = &rbdVolume{}
+	var err error
+	// ger rbdVolume from VolumeID
```
humblec: get?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+		err = fmt.Errorf("failed to get mirroring info of %s with error %s", rbdVol.String(), err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if mirroringInfo.State != librbd.MirrorImageEnabled {
+		return nil, status.Error(codes.InvalidArgument, "image mirroring is not enabled")
+	}
+
+	// return error if the image is still primary
+	if mirroringInfo.Primary {
```
humblec, line 324: If the `MirrorImageEnabled` is false( above if condition check), it cannot be `primary` anymore , Isnt it ? or I am missing some specific scenario here ?

internal/rbd/mirror.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+type imageMirrorStatus struct {
+	Name        string `json:"name"`
+	State       string `json:"state"`
+	Description string `json:"description"`
+	LastUpdate  string `json:"last_update"`
+}
+
+// Fix me once https://github.com/ceph/go-ceph/issues/460 is fixed.
```
nixpanic: please use standard `FIXME:` tagging so that https://www.tickgit.com/browse?repo=github.com/ceph/ceph-csi can be used to find it

internal/rbd/mirror.go
```diff
@@ -0,0 +1,134 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/ceph/ceph-csi/internal/util"
+
+	librbd "github.com/ceph/go-ceph/rbd"
+)
+
+func (rv *rbdVolume) enableImageMirroring(ctx context.Context, mode librbd.ImageMirrorMode) error {
```
nixpanic: ideally comment all of the functions

internal/rbd/controllerserver.go
```diff
@@ -49,6 +49,9 @@ type ControllerServer struct {
 
 	// A map storing all volumes/snapshots with ongoing operations.
 	OperationLocks *util.OperationLock
+
+	// CR holds the credentials which is used to connect to the ceph cluster.
+	CR *util.Credentials
```
nixpanic: should these credentials not be per volume? is it possible to have support for multiple ceph clusters in the same controller, similar to the csi provisioner/plugins?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,359 @@
...
+		return false, nil
+	}
+	force, err := strconv.ParseBool(val)
+	if err != nil {
+		return false, status.Errorf(codes.Internal, err.Error())
+	}
+	return force, nil
+}
+
+func getMirroringMode(ctx context.Context, parameters map[string]string) (librbd.ImageMirrorMode, error) {
```
nixpanic, line 112: make sure to document things, specially functions that receive a map as parameter (a map is usually not a nice API)

Madhu-1: the map is from the request where the user sets the opaque key-value, documented it for now

internal/cephfs/driver.go
Madhu-1: not required for cephfs for now

internal/rbd/mirror.go
Madhu-1: @humblec what extra information you are looking  for logging?

internal/rbd/mirror.go
Madhu-1: as this is specific to the rbd image, the name stands for the rbd image name not global_id

internal/rbd/mirror.go
Madhu-1: If the pool not found the code will fail in decoding the volumeHandle itself we don't hit here.

internal/rbd/replicationcontrollerserver.go
Madhu-1: For now the only snapshot but in future both will be supported its just a placeholder do you want me to remove it?

internal/rbd/replicationcontrollerserver.go
Madhu-1: `forceKey` is the string we need to return the bool value which is force

internal/rbd/replicationcontrollerserver.go
Madhu-1: `failed to disable` corrected it now

internal/rbd/replicationcontrollerserver.go
Madhu-1: can you please elaborate more on this one?

internal/rbd/replicationcontrollerserver.go
Madhu-1: we are returning an error if the mirroring is not enabled. if the mirroring is enabled we are checking it's primary or not.

internal/rbd/controllerserver.go
Madhu-1: good catch! Yes, correct. this cannot be kept in ControllerServer. added new logic for this one.

internal/rbd/mirror.go
Madhu-1: added  comment for name.

internal/rbd/replicationcontrollerserver.go
nixpanic: if it is not used at the moment, better introduce it when there is a need for it

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+	// this image is the target for replication from another storage cluster.
+	upAndReplaying imageMirroingState = "up+replaying"
+	// If the state is up+stopped means the rbd-mirror daemon is
+	// running and stopped means the image is not a target for replication from
+	// another cluster
+	upAndStopped imageMirroingState = "up+stopped"
+)
+
+// mirroringMode is the key to get the imageMirroringMode from parameters.
+var imageMirroringKey = "mirroringMode"
```
nixpanic: is a global variable really needed? or is this meant to be a `const`?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+
+// forceKey is the key to get the force option from parameters.
+var forceKey = "force"
```
nixpanic: same question here, why a global variable?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+
+// getVolumeFromID gets the rbd image details from the volumeID.
+func (cs *ControllerServer) getVolumeFromID(ctx context.Context, volumeID string, secrets map[string]string) (*rbdVolume, *util.Credentials, error) {
```
nixpanic, line 64: This looks like a function that should be part of `internal/rbd/controllerserver.go` and can be used there too.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+func (cs *ControllerServer) cleanup(rbdVol *rbdVolume, cr *util.Credentials) {
+	if cr != nil {
+		// destroy the credential file
+		cr.DeleteCredentials()
+	}
+	if rbdVol != nil {
+		// destroy the cluster connection
+		rbdVol.Destroy()
+		// release the volume lock
+		cs.VolumeLocks.Release(rbdVol.VolID)
```
nixpanic, line 137: After calling `rbdVol.Destroy()`, `rbdVol` should probably not be used anymore. Better release the locks before destroying rbdVol

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+	case imageMirrorModeJournal:
+		mirroringMode = librbd.ImageMirrorModeJournal
+	default:
+		return mirroringMode, status.Errorf(codes.InvalidArgument, "%s %s not supported", imageMirroringKey, val)
+	}
+	return mirroringMode, nil
+}
+
+// cleanup performs below resource cleanup operations.
+func (cs *ControllerServer) cleanup(rbdVol *rbdVolume, cr *util.Credentials) {
```
nixpanic, line 130: this should be a generic function in `controllerserver.go` too, it is not replication specific

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+
+const (
+	// imageMirrorModeJournal uses journaling to propagate RBD images between
+	// ceph clusters.
+	imageMirrorModeJournal imageMirroringMode = "journal"
+	// imageMirrorModeSnapshot uses snapshots to propagate RBD images between
+	// ceph clusters.
+	imageMirrorModeSnapshot imageMirroringMode = "snapshot"
+)
+
```
nixpanic, line 40: I would prefer a type for the ReplicationController, if that is possible, like:

```go
type ReplicationController struct {
    ControllerServer
}
```

then all methods in this file can use

```go
func (rc *ReplicationController) EnableReplication(...) {}
```

This helps with splitting the responsibilities of the different controllers, and makes a cleaner API.

It would even be better to have a separate package for each server component, and an internally consumable RbdVolume API. But things are very messy at the moment, so it is not something that can easily be done now. Please consider preparing for this change in the (near) future.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	if rbdVol != nil {
+		// destroy the cluster connection
+		rbdVol.Destroy()
+		// release the volume lock
+		cs.VolumeLocks.Release(rbdVol.VolID)
+	}
+}
+
+// EnableVolumeReplication enables replication on volume.
```
nixpanic: A comment  for a public API function should describe what it does, and what conditions are important. The current comment is just phrasing the name of the function into human language. can you describe what errors are returned for what conditions?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+		return nil, err
+	}
+
+	mirroringInfo, err := rbdVol.getImageMirroringInfo(ctx)
+	if err != nil {
+		err = fmt.Errorf("failed to get mirroring info of %s with error %s", rbdVol.String(), err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if mirroringInfo.State == librbd.MirrorImageEnabled {
```
nixpanic: instead of the two if-statements, it would be better to use s switch. That way the valid values can get checked, and a default invalid value gets caught automatically. Be prepared for the case that an other state can get returned in future versions.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	return &replication.DisableVolumeReplicationResponse{}, nil
+}
+
+// PromoteVolume promotes the volume as primary.
+func (cs *ControllerServer) PromoteVolume(ctx context.Context,
+	req *replication.PromoteVolumeRequest,
+) (*replication.PromoteVolumeResponse, error) {
+	// get rbdVolume from VolumeID
```
nixpanic: this comment is useless. comments help in addition to the name of the function(block), no need to comment the obvious things.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	mirroringInfo, err := rbdVol.getImageMirroringInfo(ctx)
+	if err != nil {
+		err = fmt.Errorf("failed to get mirroring info of %s with error %s", rbdVol.String(), err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if mirroringInfo.State != librbd.MirrorImageEnabled {
+		return nil, status.Errorf(codes.InvalidArgument, "image is in %d Mode", mirroringInfo.State)
```
nixpanic: This error message is not helpful. Always think about a user reading the error while troubleshooting an issue. Explain what is expected, and what the actual value is. Assist users with hints so that debugging a problem is easier.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
```
nixpanic: same here, what does this error message mean? how does it help a user to diagnose an issue?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	// demote image to secondary
+	if mirroringInfo.Primary {
+		err = rbdVol.demoteImage(ctx)
+		if err != nil {
+			util.ErrorLog(ctx, "failed to demote mirroring of %s with error %s", err.Error())
```
nixpanic: include the name of the rbd image here too? and maybe in all error messages? consider that as a standard requirement

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,351 @@
...
+	if err != nil {
+		// TODO check if the image is not found return valid error message
+		// in case of Resync the image will get deleted and gets recreated and
+		// it takes time for this operation
+		err = fmt.Errorf("failed to get mirroring info of %s with error %s", rbdVol.String(), err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if mirroringInfo.State != librbd.MirrorImageEnabled {
+		return nil, status.Error(codes.InvalidArgument, "image mirroring is not enabled")
```
nixpanic, line 320: not enabled where? in the Ceph Pool, or on the image? detailed errors please

internal/rbd/mirror.go
```diff
@@ -0,0 +1,140 @@
...
+		return err
+	}
+	defer image.Close()
+
+	err = image.MirrorEnable(mode)
+	return err
+}
+
+// disableImageMirroring disables mirroring on an image.
+func (rv *rbdVolume) disableImageMirroring(ctx context.Context, force bool) error {
```
nixpanic: There has been some movement to reduce the passing of a context to rbdVolume methods. It is better to return useful errors with details and have callers log those. Passing a context to internal rbdVolume methods makes it sometimes more difficult to re-use the functions.

internal/rbd/replicationcontrollerserver.go
Madhu-1: Yes can be moved later to do optimization in the rbd controllerserver.go

internal/rbd/replicationcontrollerserver.go
Madhu-1: For now, it is only used here. will do optimization later for cephfs and rbd.

internal/rbd/replicationcontrollerserver.go
Madhu-1: Yes, will try to do it. Later as an enhancement :+1:

internal/rbd/replicationcontrollerserver.go
Madhu-1: as we are doing it image level.  we are returning `image mirroring is not enabled` any suggestion on the error message?

internal/rbd/mirror.go
Madhu-1: Agree, we are already logging the error message if something fails. but without logging with context makes it hard for debugging if we call multiple API if one is successful and one is stuck it's very difficult to debug it.

internal/rbd/replicationcontrollerserver.go
nixpanic: please mark this with a `TODO:` comment

internal/rbd/replicationcontrollerserver.go
nixpanic: Please mark this with a `TODO:` comment.

Madhu-1: done.

internal/rbd/replicationcontrollerserver.go
Madhu-1: done

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+		if err != nil {
+			util.ErrorLog(ctx, "failed to disable mirroring of %s with error %s", err.Error())
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		// TODO check do we need to return abort after disabling the mirroring
+		return &replication.DisableVolumeReplicationResponse{}, nil
+	case librbd.MirrorImageDisabled:
+		// image is already in disabled state
+	default:
+		return nil, status.Errorf(codes.Aborted, "image is in %d Mode", mirroringInfo.State)
```
nixpanic: the error is Aborted, does that mean the mode is not supported? A `%d` digit-mode is not self-explaining. You can be more verbose with the error message.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+	mirroringInfo, err := rbdVol.getImageMirroringInfo(ctx)
+	if err != nil {
+		err = fmt.Errorf("failed to get mirroring info of %s with error %s", rbdVol.String(), err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if mirroringInfo.State != librbd.MirrorImageEnabled {
+		err = rbdVol.enableImageMirroring(ctx, mirroringMode)
+		if err != nil {
+			util.ErrorLog(ctx, "failed to enable mirroring of %s with error %s", err.Error())
```
nixpanic: There is an inconsistency in logging and updateing+returning errors. Soemtimes you gerenera a new `err` with `fmt.Errorf()`, sometimes you log a message with error, and return the original error.

I would opt to extend the error with `%w` in a new error and return it, instead of logging a different message.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	// extract the mirroring mode
+	mirroringMode, err := getMirroringMode(ctx, req.GetParameters())
+	if err != nil {
+		return nil, err
+	}
+
+	mirroringInfo, err := rbdVol.getImageMirroringInfo(ctx)
+	if err != nil {
+		err = fmt.Errorf("failed to get mirroring info of %s with error %s", rbdVol.String(), err)
```
nixpanic: when creating a new updated `err`, use `%w` to include the original error in the message so that `errors.Is()` can be used for checking (in case we would ever need it)

internal/rbd/replicationcontrollerserver.go
Madhu-1: it should be an invalid argument. the int is returned from the go-ceph, would it be good to have int to string conversion in go-ceph itself? or else we need to maintain mapping and need to make changes when something is added or removed in go-ceph.

internal/rbd/replicationcontrollerserver.go
Madhu-1: copy-paste mistake. removed `fmt.Errorf` now just returning the error message what we got.

internal/rbd/replicationcontrollerserver.go
Madhu-1: removed `fmt.Errorf`

internal/cephfs/driver.go
humblec: Better to document it.

internal/rbd/replicationcontrollerserver.go
humblec: Oh. sure.. thanks

internal/rbd/mirror.go
humblec: @Madhu-1 Isnt it good to log the volumehandle too ? so that we can track the request easily against the PVC/PV?

Madhu-1: we don't log the VolumeHandle in cephcsi, as we are logging with `ctx` it logs the request ID which is unique and helps in debugging.

internal/cephfs/driver.go
Madhu-1: done added comment for now.

Madhu-1:
@humblec PTAL

nixpanic:
For things that are marked as TODO in the description of the PR, there should be at least Issues (if not wip PRs) mentioned. We need to track all the work for the next release in the milestone. Please create issues or post changes that you already have as a PR and link them in the description.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.1-0.20190118093823-f849b5445de4
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/hashicorp/vault/api v1.0.5-0.20200902155336-f9d5ce5a171a
+	github.com/kube-storage/spec v0.0.0-20210215170332-968ad288fb6d
```
nixpanic: is there a pre-release that can be tagged and used? that way we can easily see what changes are made in the upstream project in case a future rebase is needed

internal/rbd/mirror.go
nixpanic: Indeed, that is why returning useful and complete error messages is important. Not everyone has debugging enabled, and also in those environments the issue needs to be identified. Callers will likely include the action that failed (disable image mirroring), but this function should probably mention if `open()` or `MirrorDisable()` failed.

We're not doing that consistently everywhere, but maybe we can improve the new code we're adding.

So, I'd recommend `return fmt.Errorf("open of %q failed: %w", rv.String(), err)` or similar.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+	// this image is the target for replication from another storage cluster.
+	upAndReplaying imageMirroingState = "up+replaying"
+	// If the state is up+stopped means the rbd-mirror daemon is
+	// running and stopped means the image is not a target for replication from
+	// another cluster
+	upAndStopped imageMirroingState = "up+stopped"
+)
+
+const (
+	// mirroringMode is the key to get the imageMirroringMode from parameters.
```
nixpanic: `mirroringMode` + `Key` in the comment?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+type imageMirroringMode string
+
+const (
+	// imageMirrorModeSnapshot uses snapshots to propagate RBD images between
+	// ceph clusters.
+	imageMirrorModeSnapshot imageMirroringMode = "snapshot"
+)
+
+// imageMirroingState represents the image mirroring state.
+type imageMirroingState string
```
nixpanic: imageMirroingState -> imageMirroringState (add the missing `r`)

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+			util.ErrorLog(ctx, "failed to disable mirroring of %s with error %s", err.Error())
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		// TODO: check do we need to return abort after disabling the mirroring
+		return &replication.DisableVolumeReplicationResponse{}, nil
+	case librbd.MirrorImageDisabled:
+		// image is already in disabled state
+	default:
+		// TODO: use string instead of int for returning valid error message
+		return nil, status.Errorf(codes.InvalidArgument, "image is in %d Mode", mirroringInfo.State)
```
nixpanic, line 217: @phlogistonjohn should go-ceph offer a way to translate constants like `MirrorImageDisabled` to a string so that consumers can better log it?

go.mod
Madhu-1: Not yet. as we are sticking to a particular commit should be fine. but we will have a release before the cephcsi release so that we can stick to the released version.

internal/rbd/mirror.go
Madhu-1: removed ctx in mirror.go file

internal/rbd/replicationcontrollerserver.go
ShyamsundarR: Just checked NodeStage for any issues when image is not found in this state, and it looks clean. Other requests here would return an error till then and ready state would not be disturbed/updated till resync returns the same, so looks fine.

If ready is set, does that mean initial sync is complete and subsequent incremental sync is in progress? IOW, up+syncing denotes initial rebuild for resync and up+replaying/stopped denotes further snapshot sync?

Asking, as in Ramen we would want to wait for initial full resync to be complete before calling a secondary healthy and ready for any failback/over, and hence the ready response (and subsequently the VolumeReplication CR status) is useful to determine the same.

internal/rbd/replicationcontrollerserver.go
nixpanic: this is fine, sorry for the confusion

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+				Ready: false,
+			}
+			return resp, nil
+		}
+		util.ErrorLog(ctx, err.Error())
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	ready := false
+	state := imageMirroringState(mirrorStatus.State)
+	if state == upAndStopped || state == upAndReplaying {
```
nixpanic, line 349: These are the only two states that are defined in the top of the file, can there be others?

Probably the `ResyncVolumeResponse` should contain a `Status` free-form field so that users can see more details about the status? A single boolean value (not)ready is useful for automation, but we also need to assist users that have to debug (configuration) issues. Or include some events, somehow, so that there is a log tied with the CR with changes in the mirroring state and timestamps (like events of a PVC or Pod).

Madhu-1: Yes good to add it to spec. AFAIK only these two are the state which we need to return success the above state ensures that the image is in good condition to get used.

internal/rbd/mirror.go
```diff
@@ -0,0 +1,155 @@
...
+	Description string `json:"description"`
+	LastUpdate  string `json:"last_update"`
+}
+
+// FIXME: once https://github.com/ceph/go-ceph/issues/460 is fixed use go-ceph.
+// getImageMirroingStatus get the mirroring status of an image.
+func (rv *rbdVolume) getImageMirroingStatus() (*imageMirrorStatus, error) {
+	// rbd mirror image status --format=json info [image-spec | snap-spec]
+	var imgStatus imageMirrorStatus
+	stdout, stderr, err := util.ExecCommand(
```
dillaman, line 128: Nit: Yuck. What's the hold-up on the `go-ceph` change?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+)
+
+// imageMirroringState represents the image mirroring state.
+type imageMirroringState string
+
+const (
+	// If the state is up+replaying, then mirroring is functioning properly.
+	// up means the rbd-mirror daemon is running, and replaying means
+	// this image is the target for replication from another storage cluster.
+	upAndReplaying imageMirroringState = "up+replaying"
```
dillaman, line 48: Nit: the goal should definitely be to not rely on CLI output for determining status when the librbd API provides this as decomposed fields.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+		return nil, err
+	}
+
+	// extract the force option
+	force, err := getForceOption(ctx, req.GetParameters())
+	if err != nil {
+		return nil, err
+	}
+
+	mirroringInfo, err := rbdVol.getImageMirroringInfo()
```
dillaman, line 242: Should you worry about multiple requests running concurrently? I don't see the usual per-image handling lock. At the very least, these should be idempotent so that if you attempt to promote a volume that has already been promoted, you don't get an error. 

Technically I guess this comment applies throughout.

dillaman:
Just leaving some questions/thoughts here on the higher level architecture of the replication controller:

- it's not immediately clear to me how the replication controller is tied into application workloads
- the concept of enable/disable replication, while it maps nicely to RBD/CephFS, it might not map nicely to other storage systems that might have constraints on a PVC prior to enabling/disabling mirroring (i.e. similar in concept to how some storage systems need to know that a PVC will be in a volume group before the creation of the PVC).
- I thought we had talked at one time about a ReplicatedVolume-like CRD?
- while you aren't supporting volume groups nor snapshots right now, will you be able to in the future? i.e. you won't be able to promote an individual RBD volume if it's part of a group, you would need to promote the group.

internal/rbd/replicationcontrollerserver.go
humblec: Good point @dillaman .  The mirroring operations can run in parallel. The only lock taken here is to retrieve volume from the volumeID which is not enough I feel.

Madhu-1: currently taking a lock on the volumeID (encoded volumeID) in all the requests. 

@humblec lock is not only to retrieve the volume it will be released at the last after completing all the mirroring operations. if the lock exists on a volumeID another request will try to get the lock it will fail, return operation already exists error message.

internal/rbd/replicationcontrollerserver.go
Madhu-1: Yes the PR is in progress in go-ceph will switch from CLI to go-ceph once its available

internal/rbd/mirror.go
Madhu-1: work is in progress in go-ceph to get `mirror image status`

Madhu-1:
> Just leaving some questions/thoughts here on the higher level architecture of the replication controller:
> 
> * it's not immediately clear to me how the replication controller is tied into application workloads

In the initial design, the replication controller is not tied to the application workload.

> * the concept of enable/disable replication, while it maps nicely to RBD/CephFS, it might not map nicely to other storage systems that might have constraints on a PVC prior to enabling/disabling mirroring (i.e. similar in concept to how some storage systems need to know that a PVC will be in a volume group before the creation of the PVC).

> * I thought we had talked at one time about a ReplicatedVolume-like CRD?

Yes as we discussed earlier all will be CRD based operation, the operator will make the CSI call to the driver to perform replication operations like (enable/disable mirroring, promote/demote, resync, etc for now) currently it looks like ceph specific the plan is to make it generic in future.

> * while you aren't supporting volume groups nor snapshots right now, will you be able to in the future? i.e. you won't be able to promote an individual RBD volume if it's part of a group, you would need to promote the group.

yes, future plan is to do volume group level, snapshots, etc.

internal/rbd/replicationcontrollerserver.go
Madhu-1: @humblec do you have any suggestion on the locking what extra lock we need to take and how?

dillaman: @Madhu-1 Ah, so the lock is now stashed in the `rbdVol` object and the `cleanup` method will release it?

Madhu-1: the lock is in [ControllerServer structure](https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/controllerserver.go#L53)  as its a global one and the cleanup method will release based on the volumeID

dillaman:
> yes, future plan is to do volume group level, snapshots, etc.

Right, but won't that impact the current design? Is the expectation that you will continue to add new RPC hooks between the replication controller and the CSI for every combination? 

One more thing I thought of is that for RBD at least, if you can create the image and enable mirroring at the same time, it's more efficient since it knows it doesn't have to perform any initial sync. i.e. similar to VolumeGroups, if the CSI CreateVolumeRequest call knows the volume should have mirroring enabled a priori, this optimization can get enabled. Realistically it's probably trivial, but it does highlight how certain storage systems might want to know that replication will be enabled when creating the volume.

Madhu-1:
> > yes, future plan is to do volume group level, snapshots, etc.
> 
> Right, but won't that impact the current design? Is the expectation that you will continue to add new RPC hooks between the replication controller and the CSI for every combination?
> 

Yes was thinking to add RPC hooks when we go for volumegroup level and snapshots.

> One more thing I thought of is that for RBD at least, if you can create the image and enable mirroring at the same time, it's more efficient since it knows it doesn't have to perform any initial sync. i.e. similar to VolumeGroups, if the CSI CreateVolumeRequest call knows the volume should have mirroring enabled a priori, this optimization can get enabled. Realistically it's probably trivial, but it does highlight how certain storage systems might want to know that replication will be enabled when creating the volume.

yes, good idea Thanks @dillaman. we can make it work with some storageclass parameters.  and we can always use CRD for other operations like demote, resync, etc., and also to enable mirroring after PVC created.  will think more about it and update you on this one.

cc @ShyamsundarR

ShyamsundarR:
> > Just leaving some questions/thoughts here on the higher level architecture of the replication controller:
> > 
> > * it's not immediately clear to me how the replication controller is tied into application workloads
> 
> In the initial design, the replication controller is not tied to the application workload.

The application tie up and automation is being worked in parallel. The (broad) user actions in the absence of any multi-cluster application orchestration is to control replication via an added VolRep CR (being worked here: https://github.com/kube-storage/volume-replication-operator ),
- User creates PVCs on one of the clusters (which are part of application)
- User creates VolRep CR pointing to PVCs on the same cluster and requesting them to be "Primary"
    - This enables the replication and also makes this instance Primary by default
    - IOW, the user opts in for PVC replication based on the VolRep CR and hence this is not feeding off the StorageClass during CreateVolume (more on this later, based on other comments)
- If application needs to shift clusters,
    - User restores PVs for application PVCs
    - User creates PVCs and VolRep CR on the new cluster
    - VolRep controller promotes (or force-promotes) the PV backing the PVCs as this is the new desired cluster where the Volume should be primary (auto force-promote is done here as it is a user opt-in to mark this cluster instance as Primary)
    - Application starts to function
- If mirror needs to be reestablished on the older cluster
    - User changes VolRep CR to "Secondary"
    - The controller invokes the required demote->resync(iff it is in split brain) and shifts the instance to Secondary 
- A coordinated hand-off is when VolRep is moved to Secondary on the current running cluster, and the user ensures its status reflects the desire in the CR (i.e it is now secondary), and then VolRep on the peer is moved to Primary (thus avoiding split brain situations)

Automating this workflow with application migration is being taken up separately in https://github.com/RamenDR/ramen 

> 
> > * the concept of enable/disable replication, while it maps nicely to RBD/CephFS, it might not map nicely to other storage systems that might have constraints on a PVC prior to enabling/disabling mirroring (i.e. similar in concept to how some storage systems need to know that a PVC will be in a volume group before the creation of the PVC).

Good point.

My take would be that for storage systems that need replication during volume creation, this is done using a StorageClass parameter. For storage systems that can enable replication later, the creation of VolRep CR is the trigger to enable the same. (by extension for storage systems in the former category, the "enable replication" when a VolRep CR is created becomes a no-op)

For Ceph, it could be an additional optimization that we support in the StorageClass.

I would retain the "enable" replication directive when VolRep CR for a PVC is created, for the latter storage systems.

NOTE: As stated, when we take this to a broader storage vendor eco-system the assumptions may break down

internal/rbd/replicationcontrollerserver.go
humblec: @Madhu-1 the snapshot restore operation lock is free from parent volume operation or iow, the snapshot restore operation is performed without a lock on parent volume atm, Isnt it ? if thats the case, I am wondering while  a mirroring operation on parent volume and snapshot restore operation happens ( in one of the snapshot of this volume) in parallel, is there a possibility of race or weird outcome?

Madhu-1: > @Madhu-1 the snapshot restore operation lock is free from parent volume operation or iow, the snapshot restore operation is performed without a lock on parent volume atm, Isnt it ?

we are taking a lock on the parent volume ID during SnapshotCreate operation here https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/controllerserver.go#L783-L788. During Restore operation we are restoring from a clone that is created on the parent volume During the SnapshotCreate operation.

Yes, it's already a clone do we need to worry about the lock on parent volume Here?

 if thats the case, I am wondering while a mirroring operation on parent volume and snapshot restore operation happens ( in one of the snapshot of this volume) in parallel, is there a possibility of race or weird outcome?

I don't see an issue here. Please let me know if have missed a corner case here.

Madhu-1:
@humblec @nixpanic any more comments on this one? If we don't have any more major comments I would like to get it merged and work on the TODO

internal/rbd/replicationcontrollerserver.go
humblec: May be @dillaman would be the right person to validate above scenario.

humblec:
> @humblec @nixpanic any more comments on this one? If we don't have any more major comments I would like to get it merged and work on the TODO

@Madhu-1 will review the latest version of patchset soon.

internal/rbd/replicationcontrollerserver.go
Madhu-1: Not sure about this one I need to test it out. if rbd allows mapping of the image whose state is up+syncing we can set ready to true. will check with @dillaman.

Madhu-1:
@humblec I have added https://github.com/ceph/ceph-csi/pull/1877#discussion_r589327546 to the TODO list if the issue exists we can take a look at it later. PTAL I would like to get this merge and work on other TODO's

humblec:
> @humblec @nixpanic any more comments on this one? If we don't have any more major comments I would like to get it merged and work on the TODO

One general point I would like to bring up here is that, we are starting one more GRPC service ( replication service https://github.com/ceph/ceph-csi/pull/1877/files#diff-4bc3ac45c059d328076214abbf3b72db3cd475699833d9825fcaa6471a6ccd85R171) with a third party spec/protobuf.  The CSI services based on CSI spec is what this repo driver serves as a fundamental thing.  However this patch adds a (new) replication service which is based on an unknown `spec/protobuf` along with the CSI grpc services ( identity, controller and node).  In one angle we are not sure what/how foreign spec or protobuf is defined in first place or iow, its external to the CSI spec and the same mechanism could get added to standard CSI spec too. Another thought about this approach is that, tomorrow we may have to open for yet another third party SPEC or protobuf , say for example, encryption server which works based on `encryption spec`.etc. or a new request can come in  from the community  to add another GRPC service here which eventually  may complicate the things and maintenance of this project.  I understand that, clubbing the replication service here helps to avoid some code duplication - generating volume/image from volumeid, dealing with OMAP ..etc though. However, its good to hear general opinion about this approach or better suggestions if any  from others too  ( @nixpanic @travisn  @JohnStrunk @dillaman @phlogistonjohn @jarrpa) and accordingly we can move ahead.

Madhu-1:
> > @humblec @nixpanic any more comments on this one? If we don't have any more major comments I would like to get it merged and work on the TODO
> 
> One general point I would like to bring up here is that, we are starting one more GRPC service ( replication service) with a third party spec/protobuf. The CSI services based on CSI spec is what this repo driver serves as a fundamental thing. However this patch adds a (new) replication service which is based on an unknown `spec/protobuf` along with the CSI grpc services ( identity, controller and node). In one angle we are not sure what/how foreign spec or protobuf is defined in first place or iow, its external to the CSI spec and the same mechanism could get added to standard CSI spec too. Another thought about this approach is that, tomorrow we may have to open for yet another third party SPEC or protobuf , say for example, encryption server which works based on `encryption spec`.etc. or a new request can come in from the community to add another GRPC service here which eventually may complicate the things and maintenance of this project. I understand that, clubbing the replication service here helps to avoid some code duplication - generating volume/image from volumeid, dealing with OMAP ..etc though. However, its good to hear general opinion about this approach or better suggestions if any from others too ( @nixpanic @travisn @JohnStrunk @dillaman @phlogistonjohn @jarrpa) and accordingly we can move ahead.

Yes, this is what I explained to you when we meet last time. it not only avoid code duplication, but the cephcsi driver is also the only one who knows how to decode the volumeID to extract the required operation.

Do you see any foresee any issues with clubbing multiple services with a cephcsi driver?

humblec:
> > > @humblec @nixpanic any more comments on this one? If we don't have any more major comments I would like to get it merged and work on the TODO
> > 
> > 
> > One general point I would like to bring up here is that, we are starting one more GRPC service ( replication service) with a third party spec/protobuf. The CSI services based on CSI spec is what this repo driver serves as a fundamental thing. However this patch adds a (new) replication service which is based on an unknown `spec/protobuf` along with the CSI grpc services ( identity, controller and node). In one angle we are not sure what/how foreign spec or protobuf is defined in first place or iow, its external to the CSI spec and the same mechanism could get added to standard CSI spec too. Another thought about this approach is that, tomorrow we may have to open for yet another third party SPEC or protobuf , say for example, encryption server which works based on `encryption spec`.etc. or a new request can come in from the community to add another GRPC service here which eventually may complicate the things and maintenance of this project. I understand that, clubbing the replication service here helps to avoid some code duplication - generating volume/image from volumeid, dealing with OMAP ..etc though. However, its good to hear general opinion about this approach or better suggestions if any from others too ( @nixpanic @travisn @JohnStrunk @dillaman @phlogistonjohn @jarrpa) and accordingly we can move ahead.
> 
> Yes, this is what I explained to you when we meet last time. it not only avoid code duplication, but the cephcsi driver is also the only one who knows how to decode the volumeID to extract the required operation.

@Madhu-1 yeah, same thoughts from me too which was shared in that discussion :). Just making a summary of thoughts to make sure we all are in agreement or opening it up for better thoughts/suggestions if any.

ShyamsundarR:
> > @humblec @nixpanic any more comments on this one? If we don't have any more major comments I would like to get it merged and work on the TODO
> 
> One general point I would like to bring up here is that, we are starting one more GRPC service ( replication service https://github.com/ceph/ceph-csi/pull/1877/files#diff-4bc3ac45c059d328076214abbf3b72db3cd475699833d9825fcaa6471a6ccd85R171) with a third party spec/protobuf. The CSI services based on CSI spec is what this repo driver serves as a fundamental thing. However this patch adds a (new) replication service which is based on an unknown `spec/protobuf` along with the CSI grpc services ( identity, controller and node). In one angle we are not sure what/how foreign spec or protobuf is defined in first place or iow, its external to the CSI spec and the same mechanism could get added to standard CSI spec too. Another thought about this approach is that, tomorrow we may have to open for yet another third party SPEC or protobuf , say for example, encryption server which works based on `encryption spec`.etc. or a new request can come in from the community to add another GRPC service here which eventually may complicate the things and maintenance of this project. I understand that, clubbing the replication service here helps to avoid some code duplication - generating volume/image from volumeid, dealing with OMAP ..etc though. However, its good to hear general opinion about this approach or better suggestions if any from others too ( @nixpanic @travisn @JohnStrunk @dillaman @phlogistonjohn @jarrpa) and accordingly we can move ahead.

I am on board with the current approach, of extending the CSI protocol, defining our own gRPCs, and adding sidecars to operate on the new gRPCs. Although not in the solicited list of members for suggestions, am equally responsible for designing it this way, hence posting my preference.

Madhu-1:
>Checking CEPH cluster state: [Creating]
[Timeout] CEPH cluster not in a healthy state (timeout)

/retest ci/centos/mini-e2e/k8s-1.19

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.1-0.20190118093823-f849b5445de4
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/hashicorp/vault/api v1.0.5-0.20200902155336-f9d5ce5a171a
+	github.com/kube-storage/spec v0.1.0
```
nixpanic, line 13: :+1:

nixpanic (changes requested):
The last commit should not be part of this, it can be included in a downstream-only change instead.

Other than that, it looks good to me.

Madhu-1:
> The last commit should not be part of this, it can be included in a downstream-only change instead.
> 
> Other than that, it looks good to me.

dropped the last commit.

nixpanic (approved):

nixpanic:
I have added the following to #1896 

- [ ] add unit tests for helper functions
- [ ] add e2e testing

There are a lot of things that need addressing that block a v3.3 release. Hopefully getting this merged will make it possible to address each of the items in #1896 soon.

Madhu-1:
@ShyamsundarR @humblec am waiting for one more approval on this PR. Can you please help on this one?

internal/rbd/controllerserver.go
```diff
@@ -41,6 +42,7 @@ const (
 // controller server spec.
 type ControllerServer struct {
 	*csicommon.DefaultControllerServer
+	*replication.UnimplementedControllerServer
```
humblec, line 49: may be its better to document why CS is unimplemented for replication server.

internal/rbd/mirror.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	defer image.Close()
+	err = image.MirrorPromote(force)
+	if err != nil {
+		return fmt.Errorf("failed to promote image %q with error: %w", rv.String(), err)
+	}
+	return nil
+}
+
+// demoteImage promotes image to secondary.
```
humblec: I am not sure `promote/demote` is better here, may be `demote` :)

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+	if rbdVol != nil {
+		// release the volume lock
+		cs.VolumeLocks.Release(rbdVol.VolID)
+		// destroy the cluster connection
+		rbdVol.Destroy()
+	}
+}
+
+// EnableVolumeReplication extracts the RBD volume information from the
+// volumeID,If the image is present it will enable the mirroring based on the
```
humblec: nit: space after comma

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+			return nil, status.Error(codes.InvalidArgument, "image is in non-primary state")
+		}
+		err = rbdVol.disableImageMirroring(force)
+		if err != nil {
+			util.ErrorLog(ctx, err.Error())
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		// TODO: check do we need to return abort after disabling the mirroring
+		return &replication.DisableVolumeReplicationResponse{}, nil
+	case librbd.MirrorImageDisabled:
```
humblec, line 213: Do we need to keep this case as its NOOP?

Madhu-1:
@humblec hopefully this time I have addressed all your review comments. PTAL

internal/rbd/replicationcontrollerserver.go
Madhu-1: Yes, this need to be a NO OP or else this function will return an error message in the default case.

nixpanic (approved):

Madhu-1:
@humblec PTAL

Madhu-1:
@mergifyio rebase

Madhu-1:
E2E 1.19 failed to bring up Rook ceph cluster.

Madhu-1:
@humblec CI passed PTAL

internal/rbd/controllerserver.go
```diff
@@ -41,6 +42,11 @@ const (
 // controller server spec.
 type ControllerServer struct {
 	*csicommon.DefaultControllerServer
+	// added UnimplementedControllerServer as a member of
+	// ControllerServer. if replication specadd more RPC services in the proto
```
humblec: nit: space between spec & add

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+
+	err = rbdVol.resyncImage()
+	if err != nil {
+		util.ErrorLog(ctx, err.Error())
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	// TODO: check the image state and return its ready to use or not
+	mirrorStatus, err := rbdVol.getImageMirroingStatus()
+	if err != nil {
+		// the image gets recreated after issusing resync in that case return
```
humblec: typo on `issuesing`

internal/rbd/replicationcontrollerserver.go
```diff
@@ -0,0 +1,357 @@
...
+		cs.VolumeLocks.Release(rbdVol.VolID)
+		// destroy the cluster connection
+		rbdVol.Destroy()
+	}
+}
+
+// EnableVolumeReplication extracts the RBD volume information from the
+// volumeID, If the image is present it will enable the mirroring based on the
+// user provided information.
+// TODO: create new Replication controller struct for the replication operations.
```
humblec, line 146: @Madhu-1 just to confirm, is this todo still valid?

Madhu-1: Yes, this is valid.

Madhu-1:
@humblec PTAL

Madhu-1:
@humblec @nixpanic PTAL

nixpanic (approved):

humblec:
@ShyamsundarR @dillaman any final comments/review ?

Madhu-1:
@humblec if there are any comments I would like to take up as a follow-up. This PR is created 27 days ago. am going on PTO next week. I would like to take it to completion on priority.

humblec:
> @humblec if there are any comments I would like to take up as a follow-up. This PR is created 27 days ago. am going on PTO next week. I would like to take it to completion on priority.

Sure, as the review was really active till now since start of the PR it looks like its good in shape. As mentioned in previous comment, just waiting for final ack ( with a timeout) from @dillaman and @ShyamsundarR as they were reviewing it actively too.

ShyamsundarR (approved):
LGTM!

Issue #1818, opened 2020-12-30
create cephfs pvc with error 'Operation not permitted'


deadjoker:
# Describe the bug #

I deploy ceph-csi in k8s and use cephfs to provide pvc.
PVC created fail when I use a normal ceph user but succeed if I use admin ceph user.

# Environment details #

- Image/version of Ceph CSI driver : v3.2.0
- OS version: Ubuntu 20.04.1
- Kernel version : 5.4.0-58
- Mounter used for mounting PVC (for cephfs its `fuse` or `kernel`. for rbd its
  `krbd` or `rbd-nbd`) : kernel
- Kubernetes cluster version : v1.20.0
- Containerd version: 1.4.3
- Ceph cluster version : 14.2.8

# Steps to reproduce #

1. create ceph user
`ceph auth caps client.k8sfs mon 'allow r' mgr 'allow rw' mds 'allow rw' osd 'allow rw tag cephfs *=*'`
2. dowload yaml from https://github.com/ceph/ceph-csi/tree/release-v3.2/deploy/cephfs/kubernetes 
3. modify ceph information in csi-config-map.yaml
4. add kms-config.yaml and create from it
```
---
apiVersion: v1
kind: ConfigMap
data:
  config.json: |-
    {}
metadata:
  name: ceph-csi-encryption-kms-config
```
5. add secret.yaml and create from it
```
---
apiVersion: v1
kind: Secret
metadata:
  name: csi-cephfs-secret
  namespace: ceph-csi
stringData:
  # Required for statically provisioned volumes
  #userID: <plaintext ID>
  #userKey: <Ceph auth key corresponding to ID above>

  # Required for dynamically provisioned volumes
  adminID: k8sfs
  adminKey: AQDuM+xfXz0zNRAAnxeJaWdmR2J5I/QxMR9gLQ==
```
6.  add storage class
```
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-cephfs-sc
provisioner: cephfs.csi.ceph.com
parameters:
  # String representing a Ceph cluster to provision storage from.
  # Should be unique across all Ceph clusters in use for provisioning,
  # cannot be greater than 36 bytes in length, and should remain immutable for
  # the lifetime of the StorageClass in use.
  # Ensure to create an entry in the config map named ceph-csi-config, based on
  # csi-config-map-sample.yaml, to accompany the string chosen to
  # represent the Ceph cluster in clusterID below
[...]

PR #1888, opened 2021-02-22
e2e: make use of ceph users created in e2e

Madhu-1:
This commit adds support to create and delete new ceph users to test various functionalities.

### DIFF
--- a/e2e/ceph_user.go
+++ b/e2e/ceph_user.go
@@ -1,0 +1,93 @@
+package e2e
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+// #nosec because of the word `Secret`
+const (
+	// ceph user names
+	keyringRBDProvisionerUsername          = "cephcsi-rbd-provisioner"
+	keyringRBDNodePluginUsername           = "cephcsi-rbd-node"
+	keyringRBDNamespaceProvisionerUsername = "cephcsi-rbd-ns-provisioner"
+	keyringRBDNamespaceNodePluginUsername  = "cephcsi-rbd-ns-node"
+	keyringCephFSProvisionerUsername       = "cephcsi-cephfs-provisioner"
+	keyringCephFSNodePluginUsername        = "cephcsi-cephfs-node"
+	// secret names
+	rbdNodePluginSecretName           = "cephcsi-rbd-node"
+	rbdProvisionerSecretName          = "cephcsi-rbd-provisioner"
+	rbdNamespaceNodePluginSecretName  = "cephcsi-rbd-ns-node"
+	rbdNamespaceProvisionerSecretName = "cephcsi-rbd-ns-provisioner"
+	cephFSNodePluginSecretName        = "cephcsi-cephfs-node"
+	cephFSProvisionerSecretName       = "cephcsi-cephfs-provisioner"
+)
+
+// refer https://github.com/ceph/ceph-csi/blob/devel/docs/capabilities.md#rbd
+// for RBD caps.
+func rbdNodePluginCaps(pool, rbdNamespace string) []string {
+	caps := []string{
+		"mon", "'profile rbd'",
+		"mgr", "'allow rw'",
+	}
+	if rbdNamespace == "" {
+		caps = append(caps, "osd", "'profile rbd'")
+	} else {
+		caps = append(caps, fmt.Sprintf("osd 'profile rbd pool=%s namespace=%s'", pool, rbdNamespace))
+	}
+	return caps
+}
+
+func rbdProvisionerCaps(pool, rbdNamespace string) []string {
+	caps := []string{
+		"mon", "'profile rbd'",
+		"mgr", "'allow rw'",
+	}
+	if rbdNamespace == "" {
+		caps = append(caps, "osd", "'profile rbd'")
+	} else {
+		caps = append(caps, fmt.Sprintf("osd 'profile rbd pool=%s namespace=%s'", pool, rbdNamespace))
+	}
+	return caps
+}
+
+// refer https://github.com/ceph/ceph-csi/blob/devel/docs/capabilities.md#rbd
+// for cephFS caps.
+func cephFSNodePluginCaps() []string {
+	caps := []string{
+		"mon", "'allow r'",
+		"mgr", "'allow rw'",
+		"osd", "'allow rw tag cephfs *=*'",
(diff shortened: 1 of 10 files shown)


humblec:
This could help us to catch issues like #1818 :+1:

e2e/ceph_user.go
```diff
@@ -0,0 +1,95 @@
...
+	// secret names
+	rbdNodePluginSecretName           = "cephcsi-rbd-node"
+	rbdProvisionerSecretName          = "cephcsi-rbd-provisioner"
+	rbdNamespaceNodePluginSecretName  = "cephcsi-rbd-ns-node"
+	rbdNamespaceProvisionerSecretName = "cephcsi-rbd-ns-provisioner"
+	cephFSNodePluginSecretName        = "cephcsi-cephfs-node"
+	cephFSProvisionerSecretName       = "cephcsi-cephfs-provisioner"
+)
+
+func rbdNodePluginCaps(pool, rbdNamespace string) []string {
```
humblec, line 30: Can we define something like below:

 `rbdNodePluginCapsSlice := []string{
		"mon", "'profile rbd'",
		 "mgr", "'allow rw'",
                 "osd", "'profile rbd'",
}`

rbdProvisionerCapsSlice.. etc?  then based on `namespace` passed, we can either return `rbdProvisionerCapsSlice` or `pool and namespace` appended slice.  

In other thought, we can even avoid 2 functions which repeat this action  with same function body.

Madhu-1: Separated it out as we can have different caps for provisioner or nodeplugin

humblec: @Madhu-1 ok. fine . We can have 2 functions with same body. However the caps can be defined with `osd profile rbd ` in it , so that the return  for first namespace check dont need any slice operation in it.

Madhu-1: slice operation is useful when we add more caps to the users in the future. if you don't want it I can return it from there but still, I need to append and return it. I might be missing something can you please help me with some pseudo code?

humblec:
```
rbdNodePluginCapsSlice := []string{ "mon", "'profile rbd'", "mgr", "'allow rw'", "osd", "'profile rbd'", }
rbdProvisionerCapsSlice := []string{ "mon", "'profile rbd'", "mgr", "'allow rw'", "osd", "'profile rbd'", }

rbdNodePluginCaps(pool,rbdnamespace) {
 if rbdnamespace {
  return append(rbdNodePluginCapsSlice , fmt.Sprintf("osd 'profile rbd pool=%s namespace=%s'", pool, rbdNamespace))
}
return rbdNodePluginCapsSlice 
}
same way rbdProvisionerPluginCaps(pool, rbdnamespace){
....
}```

Considering the rbdNodePluginCapsSlice and rbdProvisionerPluginCapsSlice have pretty much same common base, we can avoid 2 slices definition in the top too if needed.

**Comment by Madhu-1** [comment]
@mergifyio rebase

File: e2e/ceph_user.go (hunk 1 above)
Comment (Madhu-1) [reply e2e/ceph_user.go:30]: I see the only difference I see is declaring a global slice. do we need to do it?
and also I would like to avoid using a single slice even if the caps looks similar for now as we test more restricted users/caps having separate slices makes ease of code maintenance

**Comment by Madhu-1** [comment]
@mergifyio rebase

File: e2e/ceph_user.go (hunk 1 above)
Comment (humblec) [reply e2e/ceph_user.go:30]: not only packing to a slice global, it also avoid one append operation in both functions. I can adjust with current version though :)

**Comment by Madhu-1** [comment]
@mergifyio rebase

File: e2e/ceph_user.go [hunk 2]
```diff
@@ -0,0 +1,89 @@
...
+	caps := []string{
```
nixpanic, line 31: I would love to have a link to the description where the requirements for the capabilities are documented. When e2e fails because of capabilities and these get updated here, there should be a reminder to update the docs too.

Maybe document the required capabilities per function, and then have the document refer to the generated godoc for these functions? https://pkg.go.dev/github.com/ceph/ceph-csi/e2e does not show any documentation for e2e now, not sure what would be missing to generate it...

Madhu-1: we have documented the required caps at https://github.com/ceph/ceph-csi/blob/devel/docs/capabilities.md added a link for the same in E2E.

e2e/ceph_user.go
Madhu-1: but if the rbdnamespace is passed it will fail because of duplication osd profile as we are appending for an already existing osd in the CapsSlice

e2e/ceph_user.go
nixpanic: yes, that is the document I referred to. But as the comment is not added to every function (not needed either), it is easy to forget updating it. the other way around, have the comments for the functions describe the need, and refer to the functions from the document would result in an always up-to-date status. But, that requires the functions to be exposed to godoc...

nixpanic (approved):

Madhu-1:
@humblec PTAL

e2e/ceph_user.go
humblec: > but if the rbdnamespace is passed it will fail because of duplication osd profile as we are appending for an already existing osd in the CapsSlice

we dont need that append if namespace is non nill, we can just append `pool` and `namespace`.

Madhu-1: @humblec do you want to change like below, if yes even need to append twice because `'` at the end is important

```go
rbdNodePluginCapsSlice := []string{ "mon", "'profile rbd'", "mgr", "'allow rw'", "osd", "'profile rbd", }
rbdProvisionerCapsSlice := []string{ "mon", "'profile rbd'", "mgr", "'allow rw'", "osd", "'profile rbd", }

rbdNodePluginCaps(pool,rbdnamespace) {
 if rbdnamespace {
  return append(rbdNodePluginCapsSlice , fmt.Sprintf("pool=%s namespace=%s'", pool, rbdNamespace))
}
// add `'` to the end of profile rbd because "'" is important or else the command will fail.
// removing the last object and re-adding `'` and again appending it because complicated
}
}```

humblec: @Madhu-1 as mentioned earlier I am fine with either way ( current version or some more optimized as above ), please let me know the choice, so that we can move ahead with that.

Madhu-1: I would like to go with current approach and change it later if required

humblec (approved):

PR #1893, opened 2021-03-01
Use Mergify's defaults section for configuring common options

nixpanic:
Mergify now supports a `defaults:` section. It can be used to list the common options for actions. By using the `defaults:` section, the configurations file becomes simpler to understand as repetitive options are listed only once.

The updated configuration makes sure to the the @ceph-csi-bot for all actions, and sets the common defaults for the `merge` action.

See-also: https://docs.mergify.io/configuration.html#defaults

---

<details>

  progress or results

</details>

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -5,6 +5,10 @@ defaults:
       bot_account: ceph-csi-bot
     merge:
       bot_account: ceph-csi-bot
+      method: rebase
+      rebase_fallback: merge
+      strict: smart
+      strict_method: rebase
     rebase:
       bot_account: ceph-csi-bot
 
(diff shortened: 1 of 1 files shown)


Madhu-1:
@nixpanic https://github.com/ceph/ceph-csi/pull/1893/checks?check_run_id=2002284145 the mergify rule is invalid can you please check

.mergify.yml
```diff
@@ -39,12 +56,7 @@ pull_request_rules:
       - "status-success=ci/centos/upgrade-tests-rbd"
       - "status-success=DCO"
     actions:
-      merge:
-        method: rebase
-        rebase_fallback: merge
-        bot_account: ceph-csi-bot
-        strict: smart
-        strict_method: rebase
+      merge: {}
```
Madhu-1, line 55: do we need to keep merge action here? if we remove this won't the above default be applied?

nixpanic: yes, I think it is needed, otherwise the action will not get executed

Madhu-1 (approved):

Madhu-1:
Note:- Need manual merging.

humblec (approved):

PR #1914, opened 2021-03-11
rbd: add more logging for NodeUnstageVolume

Madhu-1:
For NodeUnstageVolume it's a two-step process. first, unmount the volume and then unmap the volume. Currently, we are logging only after rbd unmapping is done. sometimes it becomes difficult to debug with the above logging
whether more time is spent in unmount or unmap. This commits adds one more debug log after unmount is done.
with this, we can identify where exactly more time is spent by looking at the logs.


nixpanic (approved):

nixpanic:
**ci/centos/upgrade-tests-rbd** failed with some infrastructure issue:
```
Could not resolve host: github.com; Unknown error
```

Issue #1470, opened 2020-09-11
Allow snapshot creation on an encrypted volume 


humblec:
# Describe the bug #

At present, snapshot creation is blocked on a volume which is already encrypted. The reason to disallow this operation was to avoid the possibilities of the  attack   on encryption key.  

Considering our current snapshot implementation create a new CLONED volume in backend , I feel, we can use a new key for the snapshot volume. However we have to figure out the the possibile solutions on this.

@vasyl-purchel please share your thoughts as well.

vasyl-purchel:
hi @humblec, I haven't followed much on how snapshots are created, but if I'm getting it right, cloning doesn't mean copying over data from one volume to another but actually cloning the volume. in order to use new key it would require to implement re-encryption, which is not currently done... we haven't looked into that yet...

also I'm not sure how snapshot is linked to the original volume... but I believe it shouldn't be an issue if volume is re-encrypted as it will have separate key in kms meaning from encryption point of view it doesn't need to have any relationship with original volume.

dillaman:
Technically, if you don't re-encrypt the full volume after the creation of each snapshot (or similarly for each create-from-snapshot / create-from-volume call), yes, it's possible to brute-force the master encryption key (e.g. known plaintext attacks). 

For example, cloud providers (like GCE) say that snapshots will be "full" snapshots if your volumes are encrypted [1].

[1] https://cloud.google.com/compute/docs/disks/customer-supplied-encryption#create_snapshot

vasyl-purchel:
yes, that's pretty much what I was thinking of also - they also mention you must specify new encryption key, so in case of using kms instead of a single key from k8s secrets we could go similar way as with creating new disk - create new encryption key, make a full copy of existing volume (with all the metadata attached I guess) and then re-encrypt it with a new key

although I know it is possible to brute-force the master key for the first pass probably it can be done without re-encryption with a note on  that in documentation (i.e. if you create one-two copies it  won't make it brute-forceable, but if you create automatically snapshots every day or smth - do it at your own risk :) ) but for feature parity with other storage providers - make sense to have as soon as possible.

humblec:
@vasyl-purchel indeed we are planning to consider the scenario of using KMS with Vault configuration. The kube secret way is not scalable or its not what we want to really consider for enterprise customers/users. So we will be giving or kind of defaulting to KMS+ Vault configuration .. 

This is the process we follow for snapshot creation:

[...]

PR #1917, opened 2021-03-12
rbd: introduce rbdImage as base for rbdVolume and rbdSnapshot

nixpanic:
Because rbdVolume and rbdSnapshot are very similar, they can be based
off a common struct rbdImage that contains the common attributes and
functions.

This makes it possible to re-use functions for snapshots, and prevents
further duplication or code.

Updates: #1470

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -58,84 +58,88 @@ const (
 	thickProvisionMetaKey = ".rbd.csi.ceph.com/thick-provisioned"
 )
 
-// rbdVolume represents a CSI volume and its RBD image specifics.
-type rbdVolume struct {
-	// RbdImageName is the name of the RBD image backing this rbdVolume. This does not have a
-	//   JSON tag as it is not stashed in JSON encoded config maps in v1.0.0
-	// VolID is the volume ID that is exchanged with CSI drivers, identifying this rbdVol
-	// RequestName is the CSI generated volume name for the rbdVolume.  This does not have a
-	//   JSON tag as it is not stashed in JSON encoded config maps in v1.0.0
-	// VolName and MonValueFromSecret are retained from older plugin versions (<= 1.0.0)
-	//   for backward compatibility reasons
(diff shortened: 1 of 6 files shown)


humblec:
I believe there were some reasons to keep the structs different even though most of the characteristics were similar between these. May be @Madhu-1 can add details on why it was kept separately. I am fine with clubbing it to one struct though

internal/rbd/rbd_util.go
```diff
@@ -58,125 +58,129 @@ const (
...
+	// Owner is the creator (tenant, Kubernetes Namespace) of the volume
+	Owner string
+
+	CreatedAt *timestamp.Timestamp
+
+	// conn is a connection to the Ceph cluster obtained from a ConnPool
+	conn *util.ClusterConnection
+	// an opened IOContext, call .openIoctx() before using
+	ioctx *rados.IOContext
+}
```
humblec, line 102: @nixpanic Isnt it good to have a field inside this struct which says the image refers to a snapshot or not ? so that most of the decision can be based on that field. wdyt?

nixpanic: These structs should get extended when there is an opportunity to make the code simpler. I do not immediately see where the code gets simplified, or reduce calls to a "is this image based on a snapshot?" function.

As such a field would not be used by this PR, or the snapshot code is cleaned up, I prefer not to add unused fields now.

nixpanic:
> I believe there were some reasons to keep the structs different even though most of the characteristics were similar between these. May be @Madhu-1 can add details on why it was kept separately. I am fine with clubbing it to one struct though

They are still two structs, this just makes it possible to share common attributes and methods. The new `rbdImage` struct is not created/called anywhere directly, only `rbdVolume` and `rbdSnapshot` are.

Madhu-1 (approved):

internal/rbd/rbd_util.go
humblec: Yeah, we could consider it later. .Thanks!

humblec (approved):

PR #1918, opened 2021-03-12
rbd: copy passphrase for encrypted clones

### LINKED ISSUE #1470: Allow snapshot creation on an encrypted volume

nixpanic:
When a source volume is encrypted, the passphrase needs to be copied and
stored for the newly cloned volume.

This requires the `rbdImage.encryption` component to be configured, and that
is what most of the changes in this PR make possible.

Fixes: #1470

## Notes for reviewers

This PR contains a large number of commits. The easiest to review these is by going to https://github.com/ceph/ceph-csi/pull/1918/commits and clicking the title of the 1st change. Then use the `[next]` button in the top blue bar where the subject of the change is written to go to the next commit.

Each single commit is relatively simple, reviewing all changes at once will be close to impossible.

---

<details>

  progress or results

</details>

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -745,7 +745,37 @@ var _ = Describe("RBD", func() {
 			By("create a PVC-PVC clone and bind it to an app", func() {
 				// pvc clone is only supported from v1.16+
 				if k8sVersionGreaterEquals(f.ClientSet, 1, 16) {
-					validatePVCClone(pvcPath, appPath, pvcSmartClonePath, appSmartClonePath, f)
+					validatePVCClone(pvcPath, appPath, pvcSmartClonePath, appSmartClonePath, false, f)
+				}
+			})
+
+			By("create an encrypted PVC-PVC clone and bind it to an app", func() {
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 16) {
+					Skip("pvc clone is only supported from v1.16+")
+				}
+
+				err := deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				scOpts := map[string]string{
+					"encrypted":       "true",
+					"encryptionKMSID": "secrets-metadata-test",
+				}
+				err = createRBDStorageClass(f.ClientSet, f, nil, scOpts, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+
+				validatePVCClone(pvcPath, appPath, pvcSmartClonePath, appSmartClonePath, true, f)
+
+				err = deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				err = createRBDStorageClass(f.ClientSet, f, nil, nil, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
 			})
 
@@ -762,7 +792,7 @@ var _ = Describe("RBD", func() {
 				}
 				// pvc clone is only supported from v1.16+
 				if v.Major > "1" || (v.Major == "1" && v.Minor >= "16") {
-					validatePVCClone(rawPvcPath, rawAppPath, pvcBlockSmartClonePath, appBlockSmartClonePath, f)
+					validatePVCClone(rawPvcPath, rawAppPath, pvcBlockSmartClonePath, appBlockSmartClonePath, false, f)
 				}
 			})
 			By("create/delete multiple PVCs and Apps", func() {
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -6,9 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
-	"sync"
 
-	"github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
 	. "github.com/onsi/ginkgo" // nolint
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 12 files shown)


nixpanic:

Deploying the snapshot controller failed:
```
snapshotter controller creation failed
```

e2e/rbd.go
```diff
@@ -749,6 +749,36 @@ var _ = Describe("RBD", func() {
...
+				scOpts := map[string]string{
+					"encrypted":       "true",
+					"encryptionKMSID": "secrets-metadata-test",
+				}
+				err = createRBDStorageClass(f.ClientSet, f, nil, scOpts, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+
+				validatePVCClone(pvcPath, appPath, pvcSmartClonePath, appSmartClonePath, f)
```
Madhu-1: can we validate clone is also encrypted to make sure the app gets the encrypted device when using clone PVC?

internal/journal/voljournal.go
```diff
@@ -691,9 +691,22 @@ func (conn *Connection) GetImageAttributes(ctx context.Context, pool, objectUUID
 }
 
 // StoreImageID stores the image ID in omap.
-func (conn *Connection) StoreImageID(ctx context.Context, pool, reservedUUID, imageID string) error {
-	err := setOMapKeys(ctx, conn, pool, conn.config.namespace, conn.config.cephUUIDDirectoryPrefix+reservedUUID,
-		map[string]string{conn.config.csiImageIDKey: imageID})
+func (conn *Connection) StoreImageID(ctx context.Context, pool, reservedUUID, imageID, kmsID, owner string) error {
```
Madhu-1: the function is doing more work than storing the imageID. can we have a separate function to store the encryption ID and owner ?

internal/rbd/controllerserver.go
```diff
@@ -274,7 +271,16 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 				return nil, err
 			}
 		}
-		return buildCreateVolumeResponse(ctx, req, rbdVol)
+
+		if parentVol != nil && parentVol.isEncrypted() {
```
Madhu-1: the same need to be handled for snapshot also?

internal/rbd/rbd_util.go
```diff
@@ -1071,6 +1106,29 @@ func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *r
 		return fmt.Errorf("failed to create rbd clone: %w", err)
 	}
 
+	// delete the cloned image if a next step fails
+	deleteClone := true
+	defer func() {
+		if deleteClone {
+			err = librbd.RemoveImage(rv.ioctx, rv.RbdImageName)
```
Madhu-1, line 1113: use rv.DeleteImage which uses manager tasks to delete the image?

e2e/rbd.go
```diff
@@ -749,6 +749,36 @@ var _ = Describe("RBD", func() {
...
+				}
+				scOpts := map[string]string{
+					"encrypted":       "true",
+					"encryptionKMSID": "secrets-metadata-test",
+				}
+				err = createRBDStorageClass(f.ClientSet, f, nil, scOpts, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+
```
Madhu-1, line 558: Add an e2e to validate encrypted PVC+ snapshot +restore also

internal/rbd/controllerserver.go
nixpanic: no, that does not seem to be required

internal/journal/voljournal.go
nixpanic: Maybe this change is not needed, as `ReserveName()` should set the correct metadata already. I'll try dropping this change and see how it goes.

internal/rbd/rbd_util.go
nixpanic: Not sure that is needed, it is only for an error case that the just created image needs to be removed again.

Because the `deleteImage()` helper uses CephMgr, it also needs access to the Credentials, which are not available here. I think it would clutter the code more than is useful.

internal/rbd/controllerserver.go
```diff
@@ -226,6 +221,7 @@ func validateRequestedVolumeSize(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSnaps
 }
 
 // CreateVolume creates the volume in backend.
+// nolint:gocyclo // encrypted cloning/snapshots added more complexity, this needs to be addressed
```
nixpanic: fwiw, I have an extra commit that cleans up the `nolint` annotation. It will be posted with the next update of this PR.

internal/rbd/controllerserver.go
Madhu-1: what if during creating a clone from snapshot operation the rbd image got created but the encryption information is not stored (maybe in midway the controller restarted?)

internal/rbd/rbd_util.go
Madhu-1: RemoveImage directly from the ceph is time-consuming that's why we brought a manager to handle the image deletion asynchronous. I think this is fine for now.

internal/journal/voljournal.go
Madhu-1: Yes, correct. This is for post metadata reservation, Once the Reservation is done, we need to reserve the imageID at last after creating the image.

nixpanic: Dropping this does not seem to affect the functioning, so that looks good :+1: This change will not be included in the next update.

internal/rbd/controllerserver.go
nixpanic: A next call would return a `VolumeOperationAlreadyExists` error, just like in other places where the provisioning is partially done and aborted. I do not think it is easy to recover from that automatically.

e2e/rbd.go
nixpanic: oh, yes, I was planning to add that, but seem to have forgotton

e2e/rbd.go
nixpanic: I guess so, will see if I can add that when reworking the e2e commit

nixpanic: this has been added now too

e2e/rbd.go
nixpanic: Done!

internal/rbd/controllerserver.go
Madhu-1: I still feel like this is also needed for a snapshot. during vol Exist check for restoring rbd image from snapshot the parentVol is empty we need to check  if the volume encrypted if is restored from a snapshot

Madhu-1:
@nixpanic E2E is failing

nixpanic:
> @nixpanic E2E is failing

```
Apr  8 18:10:15.871: FAIL: failed to create storageclass with error volumesnapshotclasses.snapshot.storage.k8s.io "csi-rbdplugin-snapclass" already exists
```

The Snapshot tests should also delete the snapshotclass once they are done. Will address it right away!

e2e/rbd_helper.go
```diff
@@ -279,6 +268,37 @@ func validateEncryptedPVCAndAppBinding(pvcPath, appPath, kms string, f *framewor
...
+	imageData, err := getImageInfoFromPVC(pvc.Namespace, pvc.Name, f)
+	if err != nil {
+		return err
+	}
+	rbdImageSpec := imageSpec(defaultRBDPool, imageData.imageName)
+
+	return validateEncryptedImage(f, rbdImageSpec, app)
+}
+
+func validateEncryptedImage(f *framework.Framework, rbdImageSpec string, app *v1.Pod) error {
```
humblec, line 285: @nixpanic please add some source code comment about this function.

internal/util/util.go
```diff
@@ -319,3 +319,15 @@ func contains(s []string, key string) bool {
 
 	return false
 }
+
+func getKeys(m map[string]interface{}) []string {
```
humblec, line 326: please add some source code comment.

internal/rbd/clone.go
```diff
@@ -46,7 +46,10 @@ import (
 func (rv *rbdVolume) checkCloneImage(ctx context.Context, parentVol *rbdVolume) (bool, error) {
 	// generate temp cloned volume
 	tempClone := rv.generateTempClone()
+	defer tempClone.Destroy()
+
```
humblec, line 50: nit: may be this extra line is not required

internal/rbd/controllerserver.go
```diff
@@ -274,7 +270,8 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 				return nil, err
 			}
 		}
-		return buildCreateVolumeResponse(ctx, req, rbdVol)
+
```
humblec, line 275: @nixpanic is this new line intentionally kept?

internal/rbd/encryption.go
```diff
@@ -111,6 +111,41 @@ func (ri *rbdImage) setupEncryption(ctx context.Context) error {
 	return nil
 }
 
+func (ri *rbdImage) copyEncryptionConfig(cp *rbdImage) error {
```
humblec, line 118: good to have a source code comment about this newly introduced functiion @nixpanic

internal/rbd/encryption.go
```diff
@@ -111,6 +111,41 @@ func (ri *rbdImage) setupEncryption(ctx context.Context) error {
...
+	// get the unencrypted passphrase
```
humblec, line 124: Isnt this we are getting `crypt` passphrase . Got confused by above comment saying `unencrypted` passphrase

e2e/rbd_helper.go
nixpanic: will do!

internal/util/util.go
nixpanic: ok!

internal/rbd/clone.go
nixpanic: not required, but with the extra line it makes it easier to read

internal/rbd/controllerserver.go
nixpanic: Yes. After a closing `}`, if the next statement is not related to the previous code-block, I prefer an empty line to split the parts of the code.

internal/rbd/encryption.go
nixpanic: Sure, I'll add that.

internal/rbd/encryption.go
nixpanic: yes, `unencrypted` is correct. Passphrases can be encrypted per volume, so a copy of the passphrase for the next volume will get (potentially) encrypted differently. The unencrypted passphrase will remain the same. I hope the comments inside the function make this process clear.

A new comment describing the function will also clarify this more.

internal/rbd/controllerserver.go
Madhu-1:
```go
// check after vol exists from snapshot
if rbdSnap.isEncrypted() {
			rbdSnap.conn = rbdVol.conn.Copy()
			err = rbdSnap.copyEncryptionConfig(&rbdVol.rbdImage)
			if err != nil {
				return nil, fmt.Errorf("failed to clone encryption config: %w", err)
			}
		}
```

internal/rbd/controllerserver.go
humblec: Ok.. Sure..

internal/rbd/controllerserver.go
nixpanic: I think this is now addressed with the added `repairEncryptionConfig()` function.

internal/util/crypto.go
```diff
@@ -243,7 +249,10 @@ func EncryptVolume(ctx context.Context, devicePath, passphrase string) error {
 // OpenEncryptedVolume opens volume so that it can be used by the client.
 func OpenEncryptedVolume(ctx context.Context, devicePath, mapperFile, passphrase string) error {
 	DebugLog(ctx, "Opening device %s with LUKS on %s", devicePath, mapperFile)
-	_, _, err := LuksOpen(devicePath, mapperFile, passphrase)
+	_, stderr, err := LuksOpen(devicePath, mapperFile, passphrase)
+	if err != nil {
+		WarningLog(ctx, "failed to open LUKS device %q: %s", devicePath, stderr)
```
Madhu-1, line 254: this should be an error log not a warning log? logging err with stderror might also be helpfull

Madhu-1 (dismissed):
LGTM

internal/util/crypto.go
nixpanic: I'd leave that up to the caller. The caller can log the returned error with a level it deems appropriate. Only because it is a failure, I didn't think Info-level would be suitable.

humblec:
@nixpanic mostly looks good! , one quick check here  would be updating the encryption doc is needed or not, I feel we can add some notes in the encryption doc about the encrypted clones. wdyt?

internal/rbd/snapshot.go
```diff
@@ -89,7 +89,16 @@ func generateVolFromSnap(rbdSnap *rbdSnapshot) *rbdVolume {
 	vol.RadosNamespace = rbdSnap.RadosNamespace
 	vol.RbdImageName = rbdSnap.RbdSnapName
 	vol.ImageID = rbdSnap.ImageID
-	return vol
+
+	if rbdSnap.isEncrypted() {
+		vol.conn = rbdSnap.conn.Copy()
+		err := rbdSnap.copyEncryptionConfig(&vol.rbdImage)
```
nixpanic: This is wrong for the case where the KMS uses the VolID for encrypting.decrypting the passphrase (SecretMetadataKMS). The encryption of the passphrase for the new rbdVolume is done with an incorrect VolID.

nixpanic:
Now most features manually tested with VaultTokenKMS and SecretMetadataKMS. The 2nd failed in CI, with issues that could be reproduced.

@Madhu-1 @humblec please review again (hopefully last time!)

nixpanic:
> /retest ci/centos/mini-e2e

Deploying snapshot controller failed:
```
snapshotter controller creation failed
```

nixpanic:

Deploying snapshot controller failed:
```
snapshotter controller creation failed
```

nixpanic:
**ci/centos/mini-e2e-helm/k8s-1.20** failed with:
```
Apr  9 17:55:12.145: INFO: Waiting up to 3m0s for all (but 0) nodes to be ready
```

See #1969 

And so did **ci/centos/mini-e2e/k8s-1.19** https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.19/detail/mini-e2e_k8s-1.19/1098/pipeline

nixpanic:
Some e2e tests fail in **STEP: create ROX PVC clone and mount it to multiple pods**. This might be related to the cleanup of the VolumeSnapshotClass that was previously not done.

nixpanic:
> Some e2e tests fail in **STEP: create ROX PVC clone and mount it to multiple pods**. This might be related to the cleanup of the VolumeSnapshotClass that was previously not done.

All ready to go! Create/Delete the SnapshotClass when needed solved the failures.

Madhu-1 (approved):

humblec (approved):

Issue #1896, opened 2021-03-02
Address TODOs for "rbd: add support for async mirroring"


Madhu-1:
This issue is opened to track the TODO's which are pending in #1877 

Below is the list of TODO's

- [ ]  move getVolumeFromID to controllerserver.go and reuse it wherever it's applicable.
- [ ] create a new Replication controller struct for the replication operations.
- [ ] check do we need to return abort after disabling the mirroring
- [ ] use string instead of int for returning a valid error message
- [ ] check if the image is not found return a valid error message
- [ ] check the image state and return its ready to use or not
- [ ] recheck https://github.com/ceph/ceph-csi/pull/1877#discussion_r589327546
- [ ] add unit tests for helper functions
- [ ] add e2e testing

Note: TODO comments already exist in the replicationcontrollerserver.go file.

Madhu-1:
Everything is done as part of #1924 , closing this issue

PR #1924, opened 2021-03-18
rbd: Address rbd mirroring TODO's

Madhu-1:
This PR fixes the TODO's mentioned in #1896 

updates #1896 

Below is the list of TODO's

- [x]  move getVolumeFromID to controllerserver.go and reuse it wherever it's applicable.
- [x] create a new Replication controller struct for the replication operations.
- [x] check do we need to return abort after disabling the mirroring
- [ ] use string instead of int for returning a valid error message
- [x] check if the image is not found return a valid error message
- [x] check the image state and return its ready to use or not
- [ ] recheck https://github.com/ceph/ceph-csi/pull/1877#discussion_r589327546
- [x] add unit tests for helper functions
- [ ] add e2e testing

Note: AFAIK E2E cannot be added as we need two ceph clusters for testing mirroring

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -28,7 +28,6 @@ import (
 
 	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/container-storage-interface/spec/lib/go/csi"
-	"github.com/kube-storage/spec/lib/go/replication"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -42,11 +41,6 @@ const (
 // controller server spec.
 type ControllerServer struct {
 	*csicommon.DefaultControllerServer
-	// added UnimplementedControllerServer as a member of
-	// ControllerServer. if replication spec add more RPC services in the proto
-	// file, then we don't need to add all RPC methods leading to forward
-	// compatibility.
-	*replication.UnimplementedControllerServer
 	// A map storing all volumes with ongoing operations so that additional operations
 	// for that same volume (as defined by VolumeID/volume name) return an Aborted error
 	VolumeLocks *util.VolumeLocks
--- a/internal/rbd/mirror.go
+++ b/internal/rbd/mirror.go
@@ -118,6 +118,12 @@ type imageMirrorStatus struct {
 	State       string `json:"state"` // rbd image state
 	Description string `json:"description"`
 	LastUpdate  string `json:"last_update"`
+	PeerSites   []struct {
+		SiteName    string `json:"site_name"`
+		State       string `json:"state"`
+		Description string `json:"description"`
+		LastUpdate  string `json:"last_update"`
+	} `json:"peer_sites"`
 }
 
 // FIXME: once https://github.com/ceph/go-ceph/issues/460 is fixed use go-ceph.
--- a/internal/rbd/mirror.go
+++ b/internal/rbd/mirror.go
@@ -27,88 +27,88 @@ import (
 )
 
 // enableImageMirroring enables mirroring on an image.
-func (rv *rbdVolume) enableImageMirroring(mode librbd.ImageMirrorMode) error {
-	image, err := rv.open()
+func (ri *rbdImage) enableImageMirroring(mode librbd.ImageMirrorMode) error {
+	image, err := ri.open()
 	if err != nil {
-		return fmt.Errorf("failed to open image %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to open image %q with error: %w", ri.String(), err)
 	}
 	defer image.Close()
 
 	err = image.MirrorEnable(mode)
 	if err != nil {
-		return fmt.Errorf("failed to enable mirroring on %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to enable mirroring on %q with error: %w", ri.String(), err)
 	}
 	return nil
 }
 
 // disableImageMirroring disables mirroring on an image.
-func (rv *rbdVolume) disableImageMirroring(force bool) error {
-	image, err := rv.open()
+func (ri *rbdImage) disableImageMirroring(force bool) error {
+	image, err := ri.open()
 	if err != nil {
-		return fmt.Errorf("failed to open image %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to open image %q with error: %w", ri.String(), err)
 	}
 	defer image.Close()
 
 	err = image.MirrorDisable(force)
 	if err != nil {
-		return fmt.Errorf("failed to disable mirroring on %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to disable mirroring on %q with error: %w", ri.String(), err)
 	}
 	return nil
 }
 
 // getImageMirroringInfo gets mirroring information of an image.
-func (rv *rbdVolume) getImageMirroringInfo() (*librbd.MirrorImageInfo, error) {
-	image, err := rv.open()
+func (ri *rbdImage) getImageMirroringInfo() (*librbd.MirrorImageInfo, error) {
+	image, err := ri.open()
 	if err != nil {
-		return nil, fmt.Errorf("failed to open image %q with error: %w", rv.String(), err)
+		return nil, fmt.Errorf("failed to open image %q with error: %w", ri.String(), err)
 	}
 	defer image.Close()
 
 	info, err := image.GetMirrorImageInfo()
 	if err != nil {
-		return nil, fmt.Errorf("failed to get mirroring info of %q with error: %w", rv.String(), err)
+		return nil, fmt.Errorf("failed to get mirroring info of %q with error: %w", ri.String(), err)
 	}
 	return info, nil
 }
 
 // promoteImage promotes image to primary.
-func (rv *rbdVolume) promoteImage(force bool) error {
-	image, err := rv.open()
+func (ri *rbdImage) promoteImage(force bool) error {
+	image, err := ri.open()
 	if err != nil {
-		return fmt.Errorf("failed to open image %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to open image %q with error: %w", ri.String(), err)
 	}
 	defer image.Close()
 	err = image.MirrorPromote(force)
 	if err != nil {
-		return fmt.Errorf("failed to promote image %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to promote image %q with error: %w", ri.String(), err)
 	}
 	return nil
 }
 
 // demoteImage demotes image to secondary.
-func (rv *rbdVolume) demoteImage() error {
-	image, err := rv.open()
+func (ri *rbdImage) demoteImage() error {
+	image, err := ri.open()
 	if err != nil {
-		return fmt.Errorf("failed to open image %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to open image %q with error: %w", ri.String(), err)
 	}
 	defer image.Close()
 	err = image.MirrorDemote()
 	if err != nil {
-		return fmt.Errorf("failed to demote image %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to demote image %q with error: %w", ri.String(), err)
 	}
 	return nil
 }
 
 // resyncImage resync image to correct the split-brain.
-func (rv *rbdVolume) resyncImage() error {
-	image, err := rv.open()
+func (ri *rbdImage) resyncImage() error {
+	image, err := ri.open()
 	if err != nil {
-		return fmt.Errorf("failed to open image %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to open image %q with error: %w", ri.String(), err)
 	}
 	defer image.Close()
 	err = image.MirrorResync()
 	if err != nil {
-		return fmt.Errorf("failed to resync image %q with error: %w", rv.String(), err)
+		return fmt.Errorf("failed to resync image %q with error: %w", ri.String(), err)
 	}
 	return nil
 }
@@ -122,23 +122,23 @@ type imageMirrorStatus struct {
 
 // FIXME: once https://github.com/ceph/go-ceph/issues/460 is fixed use go-ceph.
 // getImageMirroingStatus get the mirroring status of an image.
-func (rv *rbdVolume) getImageMirroingStatus() (*imageMirrorStatus, error) {
+func (ri *rbdImage) getImageMirroingStatus() (*imageMirrorStatus, error) {
 	// rbd mirror image status --format=json info [image-spec | snap-spec]
 	var imgStatus imageMirrorStatus
 	stdout, stderr, err := util.ExecCommand(
 		context.TODO(),
 		"rbd",
-		"-m", rv.Monitors,
-		"--id", rv.conn.Creds.ID,
-		"--keyfile="+rv.conn.Creds.KeyFile,
+		"-m", ri.Monitors,
+		"--id", ri.conn.Creds.ID,
+		"--keyfile="+ri.conn.Creds.KeyFile,
 		"-c", util.CephConfigPath,
 		"--format="+"json",
 		"mirror",
 		"image",
 		"status",
-		rv.String())
+		ri.String())
 	if err != nil {
-		if strings.Contains(stderr, "rbd: error opening image "+rv.RbdImageName+
+		if strings.Contains(stderr, "rbd: error opening image "+ri.RbdImageName+
 			": (2) No such file or directory") {
 			return nil, util.JoinErrors(ErrImageNotFound, err)
 		}
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -405,11 +405,10 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 
 	mirroringInfo, err := rbdVol.getImageMirroringInfo()
 	if err != nil {
-		// TODO: check if the image is not found return valid error message
 		// in case of Resync the image will get deleted and gets recreated and
-		// it takes time for this operation
+		// it takes time for this operation.
 		util.ErrorLog(ctx, err.Error())
-		return nil, status.Error(codes.Internal, err.Error())
+		return nil, status.Error(codes.Aborted, err.Error())
 	}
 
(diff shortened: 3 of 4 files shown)


internal/rbd/controllerserver.go
```diff
@@ -158,6 +158,41 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
 	return rbdVol, nil
 }
 
+// getVolumeFromID gets the rbd image details from the volumeID.
```
nixpanic: please mention in the comment that when calling this function, a `defer cs.cleanup()` should be done as well

internal/rbd/controllerserver.go
```diff
@@ -158,6 +158,41 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
...
+	if acquired := cs.VolumeLocks.TryAcquire(volumeID); !acquired {
+		util.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volumeID)
+		return nil, cr, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+
+	rbdVol, err := genVolFromVolID(ctx, volumeID, cr, secrets)
+	return rbdVol, cr, err
+}
+
+// cleanup performs below resource cleanup operations.
```
nixpanic: Add a reference back to `cs.getVolumeFromID()`, only after calling that, this function should be called.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -59,6 +59,18 @@ const (
 	forceKey = "force"
 )
 
+// ReplicationServer struct of rbd CSI driver with supported methods of Replication
+// controller server spec.
+type ReplicationServer struct {
```
nixpanic, line 68: Yay, much cleaner! :partying_face:

internal/rbd/replicationcontrollerserver.go
```diff
@@ -190,10 +190,22 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
...
-		// TODO: check do we need to return abort after disabling the mirroring
+		// the image state can be still disabling once we disable the mirroring
+		// check the mirroring is disabled or not
+		mirroringInfo, err = rbdVol.getImageMirroringInfo()
+		if err != nil {
+			util.ErrorLog(ctx, err.Error())
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		if mirroringInfo.State == librbd.MirrorImageDisabling {
+			return nil, status.Errorf(codes.Aborted, "%s is in disabling state", volumeID)
```
nixpanic, line 239: I assume the consumer of the API (kube-storage sidecar?) will do a backoff and check later?

internal/rbd/mirror.go
```diff
@@ -118,6 +118,9 @@ type imageMirrorStatus struct {
 	State       string `json:"state"` // rbd image state
 	Description string `json:"description"`
 	LastUpdate  string `json:"last_update"`
+	PeerSites   []struct {
+		State string `json:"state"`
```
nixpanic: Does this only contain `State`, and no identification of the peer in a particular state? It might be useful to mention which peer is in which state?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -368,8 +368,26 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	ready := false
 	state := imageMirroringState(mirrorStatus.State)
 	if state == upAndStopped || state == upAndReplaying {
+		// Make sure the peer site image state is up and stopped
 		ready = true
+		for _, s := range mirrorStatus.PeerSites {
```
nixpanic, line 441: `PeerSites` describes multiple sites, if there is peer in a incomplete state, you may want to log that peer so users can investigate what the cause could be

internal/rbd/replicationcontrollerserver.go
```diff
@@ -336,9 +336,8 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 
 	mirroringInfo, err := rbdVol.getImageMirroringInfo()
 	if err != nil {
-		// TODO: check if the image is not found return valid error message
 		// in case of Resync the image will get deleted and gets recreated and
-		// it takes time for this operation
+		// it takes time for this operation.
```
nixpanic, line 409: Should that be a `codes.Internal` error? Or, should it be `Aborted` so a retry can be done later?

nixpanic (changes requested):
Mostly nits and questions.

Please use different PRs for addressing different TODO's. A single PR with cleanup and enhancements is more work to review than small PRs that address a single topic. No need to split this PR up, but please do not add additional commits.

internal/rbd/replicationcontrollerserver.go
Madhu-1: Yes as its an Aborted error message the operator will take care of doing backoff check

internal/rbd/replicationcontrollerserver.go
Madhu-1: Aborted makes sense. Thanks for the suggestion.

internal/rbd/replicationcontrollerserver.go
Madhu-1: Added logging now.

internal/rbd/mirror.go
Madhu-1: Added more fields like siteName, Description, and LastUpdate time.

Madhu-1:
> Mostly nits and questions.
> 
> Please use different PRs for addressing different TODO's. A single PR with cleanup and enhancements is more work to review than small PRs that address a single topic. No need to split this PR up, but please do not add additional commits.

will keep this in mind  :+1:

internal/rbd/controllerserver.go
```diff
@@ -158,6 +158,44 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
...
 }
 
+// getVolumeFromID gets the rbd image details from the volumeID and also takes
+// volume lock and creates the new user credential.
+// we need to call cs.cleanup() in defer to remove the lock and credentials.
+func (cs *ControllerServer) getVolumeFromID(ctx context.Context, volumeID string, secrets map[string]string) (*rbdVolume, *util.Credentials, error) {
+	// validate the volume ID
+	cr, err := util.NewUserCredentials(secrets)
+	if err != nil {
+		return nil, nil, status.Error(codes.Internal, err.Error())
```
nixpanic: I wonder if you need to return a gRPC error here. The callers of `cs.getVolumeID()` seem to take the error, inspect it and return their own gRPC version.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -368,8 +368,26 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	ready := false
 	state := imageMirroringState(mirrorStatus.State)
 	if state == upAndStopped || state == upAndReplaying {
+		// Make sure the peer site image state is up and stopped
 		ready = true
+		for _, s := range mirrorStatus.PeerSites {
+			if imageMirroringState(s.State) != upAndStopped {
+				util.DebugLog(ctx, "peer site name=%s mirroring state=%s, description=%s and lastUpdate=%s", s.SiteName, s.State, s.Description, s.LastUpdate)
```
nixpanic: I always think of `DebugLog` for developers, and `DefaultLog` or `InfoLog` and higher for users. Do you want to assist users with troubleshooting potential mirroring issues?

nixpanic (commented):
More nits after an other careful look through the changes

internal/rbd/replicationcontrollerserver.go
Madhu-1: Moved to InfoLog

internal/rbd/controllerserver.go
Madhu-1: Yes removed GRPC errors from this function

nixpanic: It is still there on line 167?

nixpanic: uh, line 173: https://github.com/ceph/ceph-csi/pull/1924/commits/f9fb9b6cd260e40cb54d0d6386bfb6a4b92c2604#diff-3792fcc7313c4d1e0b81c4d3ff7ac714b9d299130641673729196b2a4a1d6f2aR173

nixpanic (changes requested):
LGTM, except for the error return at https://github.com/ceph/ceph-csi/commit/f9fb9b6cd260e40cb54d0d6386bfb6a4b92c2604#diff-3792fcc7313c4d1e0b81c4d3ff7ac714b9d299130641673729196b2a4a1d6f2aR173

Madhu-1:
> LGTM, except for the error return at [f9fb9b6#diff-3792fcc7313c4d1e0b81c4d3ff7ac714b9d299130641673729196b2a4a1d6f2aR173](https://github.com/ceph/ceph-csi/commit/f9fb9b6cd260e40cb54d0d6386bfb6a4b92c2604#diff-3792fcc7313c4d1e0b81c4d3ff7ac714b9d299130641673729196b2a4a1d6f2aR173)

Fixed now

nixpanic (dismissed):

internal/rbd/controllerserver.go
```diff
@@ -158,6 +152,39 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
...
+	cr, err := util.NewUserCredentials(secrets)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if acquired := cs.VolumeLocks.TryAcquire(volumeID); !acquired {
+		return nil, cr, util.ErrOperationExists
+	}
+
+	rbdVol, err := genVolFromVolID(ctx, volumeID, cr, secrets)
```
humblec: @Madhu-1 why dont we unlock or cleanup from this function if we are done with decoding ?

internal/rbd/controllerserver.go
```diff
@@ -1130,6 +1128,13 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, err
 	}
 
+	// lock out volumeID for clone and delete operation
```
humblec: why we are taking expandLock here for clone and delete ?

internal/rbd/controllerserver_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"testing"
+
+	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
+	"github.com/ceph/ceph-csi/internal/util"
+)
+
+func TestControllerServercleanup(t *testing.T) {
+	type fields struct {
```
humblec: somehow I am not in favor of this struct name.. Can we have a better meaningful name please?

internal/rbd/mirror.go
```diff
@@ -118,27 +118,33 @@ type imageMirrorStatus struct {
 	State       string `json:"state"` // rbd image state
 	Description string `json:"description"`
 	LastUpdate  string `json:"last_update"`
+	PeerSites   []struct {
```
humblec, line 121: Can any of the field can be empty?

internal/rbd/controllerserver.go
Madhu-1: the lock needs to exist till the last part of the RPC calls. unlocking or cleanup from this function is not possible

internal/rbd/controllerserver.go
Madhu-1: This is not a new implementation, this is existing and the lock is moved below.

internal/rbd/controllerserver_test.go
Madhu-1: This auto-generated code. can you please suggest what's in your mind?

internal/rbd/mirror.go
Madhu-1: AFAIK, No this cannot be empty as at least one peer site will be present in case of mirroring.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -208,10 +208,22 @@ func (cs *ControllerServer) DisableVolumeReplication(ctx context.Context,
...
+		}
+		if mirroringInfo.State == librbd.MirrorImageDisabling {
+			return nil, status.Errorf(codes.Aborted, "%s is in disabling state", volumeID)
+		}
 		return &replication.DisableVolumeReplicationResponse{}, nil
+	// image is already in disabled state
 	case librbd.MirrorImageDisabled:
-		// image is already in disabled state
+	// image mirroring is still disabling
+	case librbd.MirrorImageDisabling:
```
humblec: Please move `Disabled` and `Disabling` cases above `Enabled`. It will avoid extra operations if image is in any of the disabled/disabling state.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -265,12 +292,27 @@ func (cs *ControllerServer) PromoteVolume(ctx context.Context,
...
+	}
+	rbdVol, cr, err := rs.getVolumeFromID(ctx, volumeID, req.GetSecrets())
+	defer rs.cleanup(rbdVol, cr)
 	if err != nil {
+		switch {
+		case errors.Is(err, util.ErrOperationExists):
+			util.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volumeID)
+			err = status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volumeID)
+		case errors.Is(err, ErrImageNotFound):
+			err = status.Errorf(codes.NotFound, "volume %s not found", volumeID)
```
humblec, line 342: just thinking, `volume not found` or `image not found` is better here?

Madhu-1: as we are considering image as the volume, IMO volume should be fine here

Madhu-1:
>  Warning  Failed     24s                kubelet            Error: ImagePullBackOff
  Normal   Pulling    10s (x2 over 25s)  kubelet            Pulling image "docker.io/ceph/ceph:v14.2.12
 Warning  Failed     10s (x2 over 25s)  kubelet            Failed to pull image "docker.io/ceph/ceph:v14.2.12": rpc error: code = Unknown desc = Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
  Warning  Failed     10s (x2 over 25s)  kubelet            Error: ErrImagePull

Rook deployment failed.

nixpanic (dismissed):

internal/rbd/controllerserver.go
humblec: just to understand this better, the lock is taken inside getVolumeFromID() and unlock happens from the caller buildcreateVolumeResponse. so if the lock is required for the entire RPC call, why dont we take the lock from the caller itself ? @Madhu-1

internal/rbd/replicationcontrollerserver.go
humblec: all the mirror operations log ( rbd_mirror*,)pointing/referencing `image` and here its volume or mix (like in DisableVolumeReplication()) at times. in case of rbd, volume and image is pretty much same, so I was thinking the uniqueness is good

internal/rbd/controllerserver.go
humblec: Lock and unlock from either caller or callee is what I am suggesting.

Madhu-1: This was done to avoid a lot of code duplication which was suggested as an enhancement in the last PR. if required i can change it back. Please let me know  your suggestion

internal/rbd/replicationcontrollerserver.go
Madhu-1: as it's a volumeID which we got from the request, sending an error as the volume Not found. all the functions in mirror.go are referring to the image because after decoding the volumeID we got the image name as those are internal functions and we are logging it as an image. Do you still think the error message we send back to the RPC calls should say `image not found` instead of `volume not found` , if yes,i can do it.

humblec: bit difficult while we interchange the reference, but its fine for now .

internal/rbd/controllerserver.go
humblec: I am in favor of lock and unlock in one/same block scope. Lets check with @nixpanic too

humblec:
@Madhu-1 apart from lock/unlock comment (https://github.com/ceph/ceph-csi/pull/1924#discussion_r604633336 ) this looks good  Thanks

internal/rbd/controllerserver.go
Madhu-1: @nixpanic what do you think?

nixpanic: I prefer to have locks taken and released in the same function. It should be an exception where this is not done.

Maybe introduce a functions like:
```golang
func (cs *ControllerServer) lockVolume(volumeID string) error {
    ...
}

func (cs *ControllerServer) unlockVolume(volumeID string) error {
    ...
}
```

This makes it really easy to re-use the functions everywhere, and set a 
```golang
defer cs.unlockVolume(volID)
```
after obtaining the lock.

I understand/guess that the `cleanup()` function was introduced to reduce the code in certain functions. However, the function moves away from being clear on when/what actions are needed, making maintenance and understanding of the code harder.

If reducing the complexity/size of a function is needed, it is better to introduce (small) helper functions that handle part of the functions work.

Madhu-1: already lock and unlock are smaller functions, introducing one more wrapper around it is not a good idea. I removed the commit which does this enhancement for now for clean code maintenance and reading.

Madhu-1:
> @Madhu-1 apart from lock/unlock comment ([#1924 (comment)](https://github.com/ceph/ceph-csi/pull/1924#discussion_r604633336) ) this looks good Thanks

@humblec This is fixed now, PTAL

humblec:
> > @Madhu-1 apart from lock/unlock comment ([#1924 (comment)](https://github.com/ceph/ceph-csi/pull/1924#discussion_r604633336) ) this looks good Thanks
> 
> @humblec This is fixed now, PTAL

Sure.. thanks!

humblec:
> > > @Madhu-1 apart from lock/unlock comment ([#1924 (comment)](https://github.com/ceph/ceph-csi/pull/1924#discussion_r604633336) ) this looks good Thanks
> > 
> > 
> > @humblec This is fixed now, PTAL
> 
> Sure.. thanks!

LGTM. Approved.. :+1:

humblec (approved):

Madhu-1:
@mergify rebase

agarwal-mudit (approved):

Issue #1921, opened 2021-03-16
Add support for Amazon KMS, with DEKs in metadata


nixpanic:
# Describe the feature you'd like to have #

We would like to support Amazon KMS for encrypting/decrypting the DEKs for volume encryption. As the costs for storing additional keys in the service get billed extra, storing the DEKs in the metadata of the volumes is more cost efficient.

# What is the value to the end user? (why is it a priority?) #

Users will be able to configure AWS KMS for Ceph-CSI volume encryption. This makes it possible to have in-flight encrypted data, and securely stored volume contents on Ceph clusters outside of the control/responsibility of the Ceph-CSI deployer.

# How will we know we have a good solution? (acceptance criteria) #

When the feature is complete:

- users can enable volume encryption in a StorageClass
- the CMK configured in Amazon KMS will be used for encrypting/decrypting the DEKs
- the encrypted DEK for a volume will be stored in the volumes metadata

# Additional context #

Unfortunately testing Amazon KMS in our CI is not possible. That means:

- manual testing is required
- no e2e tests will be added for this feature

# Configuration # 

As a minimal configuration interface, the `ceph-csi-kms-config` ConfigMap and `ceph-csi-kms-token` Secret should be supported, just like the VaultTokensKMS provider already does.

The keys in the `ceph-csi-kms-config` ConfigMap:
```
KMS_PROVIDER: aws-metadata
AWS_REGION
```
and in the `ceph-csi-kms-token` Secret:
```
AWS_CMK
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
```
The keys with the `AWS_` prefix are defined in the [`libopenstorage/secrets/aws` package](https://github.com/libopenstorage/secrets/tree/master/aws) and match the standard AWS environment variables.

nixpanic:
this is done, #1930 has been merged

PR #1927, opened 2021-03-19
util: add API for KMS Provider plugins

nixpanic:
The KMSProvider struct is a simple, extendable type that can be used to
register KMS providers with an internal kmsManager.

Helper functions for creating and configuring KMS providers will also be
located in the new kms.go file. This makes things more modular and
better maintainable.

## Future concerns ##

This work is preparing for adding an other KMS Provider, Amazon KMS through #1921.

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/util/kms.go
+++ b/internal/util/kms.go
@@ -18,7 +18,10 @@ package util
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"os"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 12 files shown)


internal/util/kms.go
```diff
@@ -0,0 +1,242 @@
...
+const (
+	// kmsProviderKey is the name of the KMS provider that is registered at
+	// the kmsManager. This is used in the ConfigMap configuration options.
+	kmsProviderKey = "KMS_PROVIDER"
+	// kmsTypeKey is the name of the KMS provider that is registered at
+	// the kmsManager. This is used in the configfile configuration
+	// options.
+	kmsTypeKey = "encryptionKMSType"
+
+	// podNamespace ENV should be set in the cephcsi container
```
humblec, line 38: @nixpanic why do we need this variable of podNamespace ?

internal/util/kms.go
```diff
@@ -0,0 +1,242 @@
...
+			return nil, fmt.Errorf("could not convert contents "+
+				"of %q to s config section", kmsID)
+		}
+		kmsConfig[kmsID] = section
+	}
+
+	return kmsConfig, nil
+}
+
+func getKMSProvider(config map[string]interface{}) (string, error) {
```
humblec, line 154: better to add source comment.

humblec:
@vasyl-purchel fyi.

internal/util/kms.go
nixpanic: It is used by the Hashicorp Vault Tokens KMS when it is configured with a ConfigMap. This PR just moves it to this file.

internal/util/kms.go
nixpanic: sure, I'll add that

vasyl-purchel:
This is a step in right direction, though I would also consider that in new pacific release there will be luks encryption support: https://docs.ceph.com/en/latest/rbd/rbd-encryption/
Still it looks like keys needs to be stored somewhere but a bunch of boilerplate we added around luks can be removed/simplified

humblec (approved):

nixpanic:
> This is a step in right direction, though I would also consider that in new pacific release there will be luks encryption support: https://docs.ceph.com/en/latest/rbd/rbd-encryption/
> Still it looks like keys needs to be stored somewhere but a bunch of boilerplate we added around luks can be removed/simplified

Nice, thanks for the info! Unfortunately this is only for rbd-nbd, and not krbd. We will also need to provide functional support for Nautilus and Octopus for a while, so moving towards the Ceph integrated solution is not feasible at the moment.

humblec:
@vasyl-purchel any further comments from your end ?

vasyl-purchel:
hi @humblec , no, change looks good to me

humblec:
> hi @humblec , no, change looks good to me

Thanks @vasyl-purchel .

PR #1932, opened 2021-03-23
e2e: verify (non)existence of keys for VaultTokensKMS

nixpanic:
Key existence and removal is only checked for the VaultKMS provider. It
should also be done for the VaultTokensKMS provider.

## Related issues ##

Fixes: https://bugzilla.redhat.com/1941836

---

<details>

  progress or results

</details>

### DIFF
--- a/e2e/rbd_helper.go
+++ b/e2e/rbd_helper.go
@@ -225,6 +225,10 @@ func validateImageOwner(pvcPath string, f *framework.Framework) error {
 	return deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
 }
 
+func kmsIsVault(kms string) bool {
+	return kms == "vault"
+}
+
 func validateEncryptedPVCAndAppBinding(pvcPath, appPath, kms string, f *framework.Framework) error {
 	pvc, app, err := createPVCAndAppBinding(pvcPath, appPath, f, deployTimeout)
 	if err != nil {
@@ -252,9 +256,9 @@ func validateEncryptedPVCAndAppBinding(pvcPath, appPath, kms string, f *framewor
 		return fmt.Errorf("%v not equal to crypt", mountType)
 	}
 
-	if kms == "vault" {
+	if kmsIsVault(kms) || kms == "vaulttokens" {
 		// check new passphrase created
-		_, stdErr := readVaultSecret(imageData.csiVolumeHandle, f)
+		_, stdErr := readVaultSecret(imageData.csiVolumeHandle, kmsIsVault(kms), f)
 		if stdErr != "" {
 			return fmt.Errorf("failed to read passphrase from vault: %s", stdErr)
 		}
@@ -265,9 +269,9 @@ func validateEncryptedPVCAndAppBinding(pvcPath, appPath, kms string, f *framewor
 		return err
 	}
 
-	if kms == "vault" {
+	if kmsIsVault(kms) || kms == "vaulttokens" {
 		// check new passphrase created
-		stdOut, _ := readVaultSecret(imageData.csiVolumeHandle, f)
+		stdOut, _ := readVaultSecret(imageData.csiVolumeHandle, kmsIsVault(kms), f)
 		if stdOut != "" {
 			return fmt.Errorf("passphrase found in vault while should be deleted: %s", stdOut)
 		}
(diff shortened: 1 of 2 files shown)


humblec (dismissed):

nixpanic:
The `readVaultSecret()` function in the e2e tests does not take into account that the path for a secret is different for the `"vaulttokens"`. This needs some improvements.

nixpanic:
The reported bug may not be a bug after all. It seems intentional behaviour of Hashicorp Vault.

After checking the contents of the key that should have been removed, I noticed the `deletion_time`:
```
/ $ vault kv get secret/0001-0011-openshift-storage-0000000000000001-eb00c257-8c8d-11eb-9c37-0a580a800244
====== Metadata ======
Key              Value
---              -----
created_time     2021-03-24T10:44:32.654895768Z
deletion_time    2021-03-24T10:48:23.006111934Z
destroyed        false
version          1
```
There also is no `Data` section, indicating that the contents has been removed.

Depending on the requirements and expectation of customers, they may want to keep this around for some time. It is possible to `undelete` the keys through the Vault tools.

To completely delete/destroy the contents with a command like this:
```
/ $ vault kv metadata delete secret/0001-0011-openshift-storage-0000000000000001-eb00c257-8c8d-11eb-9c37-0a580a800244
Success! Data deleted (if it existed) at: secret/metadata/0001-0011-openshift-storage-0000000000000001-eb00c257-8c8d-11eb-9c37-0a580a800244
```

Subsequent "vault kv list secret" calls do not output the old key anymore.

I am not sure if Hashicorp Vault can be configured to automatically cleanup the metadata of deleted keys.

humblec:
@nixpanic it looks like while we delete, the metadata or version of that key kept in the kv store even after deletion.  https://www.vaultproject.io/docs/commands/kv/delete

```
-versions ([]int: <required>) - The versions to be deleted. The versioned data will not be deleted, but it will no longer be returned in normal get requests.
```

nixpanic:
> @nixpanic it looks like while we delete the metadata or version of that key kept in the kv store even after deletion. https://www.vaultproject.io/docs/commands/kv/delete
> 
> ```
> -versions ([]int: <required>) - The versions to be deleted. The versioned data will not be deleted, but it will no longer be returned in normal get requests.
> ```

Indeed. And we use the [libopenstorage/secrets](https://github.com/libopenstorage/secrets) package, which does not send the `destroy` or `delete metadata` commands. Other projects that use it (like Rook) behave the same way.

humblec (approved):

e2e/rbd_helper.go
```diff
@@ -252,9 +256,9 @@ func validateEncryptedPVCAndAppBinding(pvcPath, appPath, kms string, f *framewor
 		return fmt.Errorf("%v not equal to crypt", mountType)
 	}
 
-	if kms == "vault" {
+	if kmsIsVault(kms) || kms == "vaulttokens" {
```
Madhu-1, line 259: helper function for vaulttokens? similar to `kmsIsVault` `kmsIsVaultToken`?

e2e/utils.go
```diff
@@ -207,9 +212,15 @@ func getMountType(appName, appNamespace, mountPath string, f *framework.Framewor
 //  * issue get request for particular key
 // resulting in stdOut (first entry in tuple) - output that contains the key
 // or stdErr (second entry in tuple) - error getting the key.
-func readVaultSecret(key string, f *framework.Framework) (string, string) {
+func readVaultSecret(key string, usePassphrasePath bool, f *framework.Framework) (string, string) {
```
Madhu-1, line 215: are we setting usePassphrasePath to true or false anywhere? not able to find it

e2e/rbd_helper.go
nixpanic: I added `kmsIsVault()` because golang-ci complained about the 4x usage of `kms == "vault"`. `"vaulttokens"` is not used that much, so I didn't add a helper for that. If you really like one, let me know and I'll update this PR again.

e2e/utils.go
nixpanic: See lines 261 + 274 in e2e/rbd_helper.go, depending on the KMS-ID (`vault`/`vaulttokens`) `usePassphrasePath` is set to `true` or `false`.

Madhu-1 (approved):

e2e/rbd_helper.go
Madhu-1: Okay got it may be when we add more options we can do it. Should not block this PR from getting merged

PR #1938, opened 2021-03-29
[RBD Async DR] Fix resize/Delete volume when it moved to another cluster

### LINKED ISSUE #1936: [RBD Async DR] Not able to resize/Delete volume when it moved to another cluster
we have implemented omap regeneration as part of https://github.com/ceph/ceph-csi/pull/1641

During Async mirroring only the RBD images are mirrored to the secondary site, and when the user creates the static PVCs on the failover we need to regenerate the omap data. The volumeHandle in PV spec is an encoded string that contains clusterID and poolID and image UUID, The clusterID and poolID won't remain the same on both the clusters, for that cephcsi need to generate the new volume handle and its to create a mapping between new volume handle and old volume handle with that whenever cephcsi gets csi requests it check if the mapping exists it will pull the new volume handle and continues other operations.

During few operations like (ControllerExpand, DeleteVolume) we will get the volume handle as the omap is stored in the rados pool we need 
the pool name first to get the volume handle mapping.

I0325 10:58:38.423582       1 utils.go:132] ID: 37 Req-ID: 0001-0009-rook-ceph-0000000000000001-3cadbf3c-8d57-11eb-8968-0242ac110006 GRPC call: /csi.v1.Controller/ControllerExpandVolume
I0325 10:58:38.427109       1 utils.go:133] ID: 37 Req-ID: 0001-0009-rook-ceph-0000000000000001-3cadbf3c-8d57-11eb-8968-0242ac110006 GRPC request: {"capacity_range":{"required_bytes":10737418240},"secrets":"***stripped***","volume_capability":{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":1}},"volume_id":"0001-0009-rook-ceph-0000000000000001-3cadbf3c-8d57-11eb-8968-0242ac110006"}
E0325 10:58:38.438318       1 omap.go:77] ID: 37 Req-ID: 0001-0009-rook-ceph-0000000000000001-3cadbf3c-8d57-11eb-8968-0242ac110006 omap not found (pool="device_health_metrics", namespace="", name="csi.volume.3cadbf3c-8d57-11eb-8968-0242ac110006"): rados: ret=-2, No such file or directory
W0325 10:58:38.439255       1 voljournal.go:636] ID: 37 Req-ID: 0001-0009-rook-ceph-0000000000000001-3cadbf3c-8d57-11eb-8968-0242ac110006 unable to read omap keys: pool or key missing: key not found: rados: ret=-2, No such file or directory
E0325 10:58:38.453466       1 rbd_journal.go:601] ID: 37 Req-ID: 0001-0009-rook-ceph-0000000000000001-3cadbf3c-8d57-11eb-8968-0242ac110006 failed to get image id device_health_metrics/csi-vol-3cadbf3c-8d57-11eb-8968-0242ac110006: image not found: RBD image not found
E0325 10:58:38.454078       1 utils.go:136] ID: 37 Req-ID: 0001-0009-rook-ceph-0000000000000001-3cadbf3c-8d57-11eb-8968-0242ac110006 GRPC error: rpc error: code = NotFound desc = volume ID 0001-0009-rook-ceph-0000000000000001-3cadbf3c-8d57-11eb-8968-0242ac110006 not found

## Suggested implementation

* Omap generator adds volume handle mapping in to new configmap
* Mount the new configmap as a volume to the cephcsi pods
* cephcsi checks for the volume handle mapping in the mounted volume
* If the volume handle mapping exists get the new volume handle
* Based on the new VolumeHandle decode it and use it

Madhu-1:
In the case of the Async DR, the volumeID will not be the same if the clusterID or the PoolID is different, With Earlier implementation, it is expected that the new volumeID mapping is stored in the rados omap pool. In the case of the   ControllerExpand or the DeleteVolume Request, the only volumeID will be sent it's not possible to find the corresponding poolID in the new cluster.

With This Change, it works as below

The csi-rbdplugin-controller will watch for the PV objects, when there are any PV objects created it will check the omap already exists, If the omap doesn't exist it will generate the new volumeID and it checks for the volumeID mapping entry in the new configmap, if the mapping does not exist, it will add the new entry to the configmap. the controller also does a periodic cleanup to remove the stale volumeID mappings from the configmap.

The configmap will be mounted as a volume to the csi-rbdplugin provisioner, when cephcsi receives a request for the volume it will check for the volumeID mapping in the mounted file, if the mapping exists it will
use the new volumeID for further operations.

fixes: #1936

### DIFF
--- a/charts/ceph-csi-rbd/templates/csi-volumeid-mapping.yaml
+++ b/charts/ceph-csi-rbd/templates/csi-volumeid-mapping.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ceph-csi-volumeid-mapping
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ include "ceph-csi-rbd.name" . }}
+    chart: {{ include "ceph-csi-rbd.chart" . }}
+    component: {{ .Values.nodeplugin.name }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  mapping.json: |-
+    []
--- a/charts/ceph-csi-rbd/templates/provisioner-clusterrole.yaml
+++ b/charts/ceph-csi-rbd/templates/provisioner-clusterrole.yaml
@@ -50,7 +50,7 @@ rules:
     verbs: ["update"]
   - apiGroups: [""]
     resources: ["configmaps"]
-    verbs: ["get"]
+    verbs: ["get", "list", "watch", "update"]
 {{- if .Values.provisioner.resizer.enabled }}
   - apiGroups: [""]
     resources: ["persistentvolumeclaims/status"]
--- a/charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
+++ b/charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
@@ -176,6 +176,8 @@ spec:
               mountPath: /etc/ceph-csi-encryption-kms-config/
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
+            - name: volumeid-mapping
+              mountPath: /etc/ceph-csi-volumeid
           resources:
 {{ toYaml .Values.nodeplugin.plugin.resources | indent 12 }}
 {{- if .Values.provisioner.deployController }}
@@ -260,6 +262,9 @@ spec:
           emptyDir: {
             medium: "Memory"
           }
+        - name: volumeid-mapping
+          configMap:
+            name: ceph-csi-volumeid-mapping
 {{- if .Values.provisioner.affinity }}
       affinity:
 {{ toYaml .Values.provisioner.affinity | indent 8 -}}
--- a/deploy/rbd/kubernetes/csi-provisioner-rbac.yaml
+++ b/deploy/rbd/kubernetes/csi-provisioner-rbac.yaml
@@ -54,7 +54,7 @@ rules:
     verbs: ["update"]
   - apiGroups: [""]
     resources: ["configmaps"]
-    verbs: ["get"]
+    verbs: ["get", "list", "watch", "update"]
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
--- a/deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
+++ b/deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
@@ -159,6 +159,8 @@ spec:
               mountPath: /etc/ceph-csi-encryption-kms-config/
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
+            - name: volumeid-mapping
+              mountPath: /etc/ceph-csi-volumeid
         - name: csi-rbdplugin-controller
           securityContext:
             privileged: true
@@ -219,6 +221,9 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: ceph-csi-config
+        - name: volumeid-mapping
+          configMap:
+            name: ceph-csi-volumeid-mapping
         - name: ceph-csi-encryption-kms-config
           configMap:
             name: ceph-csi-encryption-kms-config
--- a/deploy/rbd/kubernetes/csi-volumeid-mapping.yaml
+++ b/deploy/rbd/kubernetes/csi-volumeid-mapping.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ceph-csi-volumeid-mapping
+data:
+  mapping.json: |-
+    []
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"sync"
 
@@ -23,6 +24,7 @@ var (
 	rbdNodePluginRBAC  = "csi-nodeplugin-rbac.yaml"
 	rbdNodePluginPSP   = "csi-nodeplugin-psp.yaml"
 	configMap          = "csi-config-map.yaml"
+	volumeIDMapping    = "csi-volumeid-mapping.yaml"
 	rbdDirPath         = "../deploy/rbd/kubernetes/"
 	rbdExamplePath     = "../examples/rbd/"
 	rbdDeploymentName  = "csi-rbdplugin-provisioner"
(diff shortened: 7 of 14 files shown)


Madhu-1:
@nixpanic @ShyamsundarR @humblec can you please review this one priority?

nixpanic:
Is is not much simpler to set the volumeID/handle in the metadata of the RBD image? What prevents doing that?

I'm not really in favour of having a ConfigMap that basically acts as a database of volumes...

Madhu-1:
> Is is not much simpler to set the volumeID/handle in the metadata of the RBD image? What prevents doing that?
> 

Not that much Easy, to go to the rbd image you need the image name and the pool name which is not possible. In most cases, cephcsi will get a request with volumeID. we decode the volumeId and extract clusterID, pool ID, and retrieves the rbd image name from the omap.

> I'm not really in favour of having a ConfigMap that basically acts as a database of volumes...

Any other suggestions to solve this issue?

humblec:
> Is is not much simpler to set the volumeID/handle in the metadata of the RBD image? What prevents doing that?
> 
> I'm not really in favour of having a ConfigMap that basically acts as a database of volumes...

Agree to the above. Additionally, the csi-rbdplugin-controller watching for PVs, updating configmap and also doing stale cleanup looks very tedious/weird from CSI driver stand point.  imo, these kind of tasks should be outside CSI driver itself( which I am repeating :) from https://github.com/ceph/ceph-csi/pull/1877#issuecomment-795281859) . It seems that,  we are adding more hacks/tweaks here and there to get this async story tackled somehow.

Madhu-1:
> > Is is not much simpler to set the volumeID/handle in the metadata of the RBD image? What prevents doing that?
> > I'm not really in favour of having a ConfigMap that basically acts as a database of volumes...
> 
> Agree to the above. The csi-rbdplugin-controller watching for PVs, updating configmap and also doing stale cleanup looks very tedious/weird from CSI driver stand point. imo, these kind of tasks should be outside CSI driver itself( which I am repeating :) from [#1877 (comment)](https://github.com/ceph/ceph-csi/pull/1877#issuecomment-795281859)) . It seems that, we are adding more hacks/tweaks here and there to get this async story tackled somehow.

csi-rbdplugin-controller is a sidecar. in future, the csidriver will talk to a new sidecar to get the corresponding volumeID with the new RPC spec (this is not there yet)

@humblec am looking for a suggestion as I mention in https://github.com/ceph/ceph-csi/issues/1936 a few days ago.

agarwal-mudit:
@Madhu-1 Will this take care of existing volumes? Meaning, lets say we enable DR on an existing cluster which already has many volumes.

Also, can there be a case when the data in the new config map is corrupted.  Situations like Split brain etc.?

Madhu-1:
> @Madhu-1 Will this take care of existing volumes? Meaning, lets say we enable DR on an existing cluster which already has many volumes.
> 
> Also, can there be a case when the data in the new config map is corrupted. Situations like Split brain etc.?

This is only for the volumes which are moved from one cluster to another cluster, not for the new volume (During failover when the PVC, PV are created for the existing images the entry will get populated in the configmap as cephcsi won't know the pool Name during few operations like Expand and Delete). 
chances are less (not sure we can get split-brain in CM) If the configmap is corrupted i assume it will be same as omap corruption either the read from the mounted volume will fail and we are not able to perform any operations.

nixpanic:
> > I'm not really in favour of having a ConfigMap that basically acts as a database of volumes...
> 
> Any other suggestions to solve this issue?

Can the volume-id be set in a label on the PV, or in some other attribute? It can be a task of the (currently person) operator that creates the static PVs on the secondary site.

Maybe it would be easier if there is a `ReplicatedPersistentVolume` CR that points to a PV and some of the other options? That way, a StorageReplicationController can watch for the `ReplicatedPersistentVolume` objects and store/fetch information there.

Madhu-1:
> > > I'm not really in favour of having a ConfigMap that basically acts as a database of volumes...
> > 
> > 
> > Any other suggestions to solve this issue?
> 
> Can the volume-id be set in a label on the PV, or in some other attribute? It can be a task of the (currently person) operator that creates the static PVs on the secondary site.
> 

This becomes a cephcsi specific fix as the other drivers might work with the same volumeID.

> Maybe it would be easier if there is a `ReplicatedPersistentVolume` CR that points to a PV and some of the other options? That way, a StorageReplicationController can watch for the `ReplicatedPersistentVolume` objects and store/fetch information there.

@nixpanic the problem really exists at the ExpandVolume and DeleteVolume which will get trigger from the external-provisioner.

humblec:
> > > Is is not much simpler to set the volumeID/handle in the metadata of the RBD image? What prevents doing that?
> > > I'm not really in favour of having a ConfigMap that basically acts as a database of volumes...
> > 
> > 
> > Agree to the above. The csi-rbdplugin-controller watching for PVs, updating configmap and also doing stale cleanup looks very tedious/weird from CSI driver stand point. imo, these kind of tasks should be outside CSI driver itself( which I am repeating :) from [#1877 (comment)](https://github.com/ceph/ceph-csi/pull/1877#issuecomment-795281859)) . It seems that, we are adding more hacks/tweaks here and there to get this async story tackled somehow.
> 
> csi-rbdplugin-controller is a sidecar. in future, the csidriver will talk to a new sidecar to get the corresponding volumeID with the new RPC spec (this is not there yet)

I understand its  a sidecar, the point I am making here is, the concern due to  the dependency/references requires for CSI driver to operate due to this change or set of changes in the volumeID  handling.

agarwal-mudit:
> I understand its a sidecar, the point I am making here is, the concern due to the dependency/references requires for CSI driver to operate due to this change or set of changes in the volumeID handling.

@humblec We heard you and no one is denying what you are saying but I guess what Madhu has been asking is for an alternative solution/idea. Will really appreciate if you can help in figuring out the same.

humblec:
> > I understand its a sidecar, the point I am making here is, the concern due to the dependency/references requires for CSI driver to operate due to this change or set of changes in the volumeID handling.
> 
> @humblec We heard you and no one is denying what you are saying but I guess what Madhu has been asking is for an alternative solution/idea. Will really appreciate if you can help in figuring out the same.

@agarwal-mudit @Madhu-1 sure, Does  the available design doc here in this repo reflect the current  implementation ? or it need any update compared to initial plan ? also helpful if there is any integration design doc ( csi driver+ other components of async work) available to refer.

nixpanic:
I favour the solution in #1946 much more than this approach with the ConfigMap.

@humblec could you share your opinion too?

humblec:
> I favour the solution in #1946 much more than this approach with the ConfigMap.
> 
> @humblec could you share your opinion too?

Yeah , definitely `configmap` approach is bit ugly. I will review the solution in 1946 soon. Thanks !

PR #1942, opened 2021-03-29
util: replace empty slice declaration via literal with nil slice

humblec:


nixpanic:
Why is this needed? If there are Golang coding guidelines that recommend this, please link them in the commit message. It probably would be good to include the reference in `docs/development-guide.md` too.

humblec:
> Why is this needed? If there are Golang coding guidelines that recommend this, please link them in the commit message. It probably would be good to include the reference in `docs/development-guide.md` too.

Its a good practice as the compiler knows in advance the nil size and better performance. But its not mandatory to have too. so a rule cant be established on the same.  I may close this PR as a rule cant be emphasized based on this.

PR #1944, opened 2021-03-29
util: replace snake,camel case import from packages

humblec:

### DIFF
--- a/internal/csi-common/server.go
+++ b/internal/csi-common/server.go
@@ -24,8 +24,8 @@ import (
 	"sync"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
-	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
-	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
+	grpcmiddleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	grpcprometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/kube-storage/spec/lib/go/replication"
 	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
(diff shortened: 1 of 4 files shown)


internal/csi-common/server.go
```diff
@@ -24,8 +24,8 @@ import (
 	"sync"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
-	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
-	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
+	grpcmiddleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	grpcprometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
```
Madhu-1, line 28: This is from the package name https://github.com/grpc-ecosystem/go-grpc-prometheus/blob/9abf3eb82b4a313b1a28f370a7ef8ffe0667c681/server_metrics.go#L1 any specific reason to change this one?

humblec: @Madhu-1 the package name may be in that format( snake case and camel cases are not recommended though), havent seen the import name in snake case, have you seen such occurrence?

Madhu-1: the sample code was referred from https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware. if we want to remove it we need to document it in development guide to avoid such occurrence or we can remove the alias let us the package name directly

humblec: Let me rework the patch. There are 3 more instances to be changed. @Madhu-1

nixpanic: > snake case and camel cases are not recommended though

Where is this documented? If there are generic Golang guidelines that we should follow, it would be best to also enable checking for them in the golang-ci configuration.

humblec: @nixpanic I know this is implemented in projects like `kubernetes` where camelcase, snake case imports are not allowed.  Also see https://rakyll.org/style-packages/. Is it enough proof for removing such instances from this repo?

humblec: @nixpanic @Madhu-1 updated all the instances and also updated the commit message. PTAL.

nixpanic: I would like to know how we can prevent this from getting re-introduced in the future. Is there a golang-ci option that we can enable after this commit?

humblec: @nixpanic sure,, I can check whether we can emphasize in the CI somehow and file a PR accordingly.. Can we get this in ?

nixpanic: I think I prefer to see a CI job in combination with this PR. Can you add that as a 2nd commit?

humblec: @nixpanic I still dont know how to catch such instances in CI with a rule . Thats why I said, I have to check on it.

nixpanic: It is not urgent to have this addressed, so it is fine to keep this PR open until there is a CI job to validate this change.

PR #1946, opened 2021-03-30
[RBD Async DR] Fix resize/Delete volume when it moved to another cluster

### LINKED ISSUE #1936: [RBD Async DR] Not able to resize/Delete volume when it moved to another cluster

Madhu-1:
In the case of the Async DR, the volumeID will not be the same if the clusterID or the PoolID is different, With Earlier implementation, it is expected that the new volumeID mapping is stored in the rados omap pool. In the case of the
ControllerExpand or the DeleteVolume Request, the only volumeID will be sent it's not possible to find the corresponding poolID in the new cluster.

With This Change, it works as below

The csi-rbdplugin-controller will watch for the PV objects, when there are any PV objects created it will check the omap already exists, If the omap doesn't exist it will generate the new volumeID and it checks for the volumeID mapping entry in the PV annotation, if the mapping does not exist, it will add the new entry to the PV annotation.

The cephcsi will check for the PV annotations if the omap does not exist if the mapping exists in the PV annotation, it will use the new volumeID for further operations.

fixes: #1936
closes: #1938 (this is an alternative)

### DIFF
--- a/internal/controller/persistentvolume/persistentvolume.go
+++ b/internal/controller/persistentvolume/persistentvolume.go
@@ -41,6 +41,7 @@ import (
 type ReconcilePersistentVolume struct {
 	client client.Client
 	config ctrl.Config
+	Locks  *util.VolumeLocks
 }
 
 var _ reconcile.Reconciler = &ReconcilePersistentVolume{}
@@ -62,6 +63,7 @@ func newPVReconciler(mgr manager.Manager, config ctrl.Config) reconcile.Reconcil
 	r := &ReconcilePersistentVolume{
 		client: mgr.GetClient(),
 		config: config,
+		Locks:  util.NewVolumeLocks(),
 	}
 	return r
 }
@@ -173,6 +175,12 @@ func (r ReconcilePersistentVolume) reconcilePV(obj runtime.Object) error {
 		secretName = pv.Spec.CSI.NodeStageSecretRef.Name
 		secretNamespace = pv.Spec.CSI.NodeStageSecretRef.Namespace
 	}
+
+	// Take lock to process only one volumeHandle at a time.
+	if ok := r.Locks.TryAcquire(pv.Spec.CSI.VolumeHandle); !ok {
+		return fmt.Errorf(util.VolumeOperationAlreadyExistsFmt, pv.Spec.CSI.VolumeHandle)
+	}
+	defer r.Locks.Release(pv.Spec.CSI.VolumeHandle)
 
 	cr, err := r.getCredentials(secretName, secretNamespace)
 	if err != nil {
--- a/internal/controller/persistentvolume/persistentvolume.go
+++ b/internal/controller/persistentvolume/persistentvolume.go
@@ -122,6 +122,24 @@ func checkStaticVolume(pv *corev1.PersistentVolume) (bool, error) {
 	return static, nil
 }
 
+// storeVolumeIDInPV stores the new volumeID in PV object.
+func (r ReconcilePersistentVolume) storeVolumeIDInPV(pv *corev1.PersistentVolume, newVolumeID string) error {
+	if v, ok := pv.Annotations[rbd.PVVolumeHandleAnnotationKey]; ok {
+		if v == newVolumeID {
+			return nil
+		}
+	}
+	if pv.Annotations == nil {
+		pv.Annotations = make(map[string]string)
+	}
+	if pv.Labels == nil {
+		pv.Labels = make(map[string]string)
+	}
+	pv.Labels[rbd.PVReplicatedLabelKey] = rbd.PVReplicatedLabelValue
+	pv.Annotations[rbd.PVVolumeHandleAnnotationKey] = newVolumeID
+	return r.client.Update(context.TODO(), pv)
+}
+
 // reconcilePV will extract the image details from the pv spec and regenerates
 // the omap data.
 func (r ReconcilePersistentVolume) reconcilePV(obj runtime.Object) error {
@@ -163,10 +181,17 @@ func (r ReconcilePersistentVolume) reconcilePV(obj runtime.Object) error {
 	}
 	defer cr.DeleteCredentials()
 
-	err = rbd.RegenerateJournal(imageName, volumeHandler, pool, journalPool, requestName, cr)
+	rbdVolID, err := rbd.RegenerateJournal(imageName, volumeHandler, pool, journalPool, requestName, cr)
 	if err != nil {
 		util.ErrorLogMsg("failed to regenerate journal %s", err)
 		return err
+	}
+	if rbdVolID != volumeHandler {
+		err = r.storeVolumeIDInPV(pv, rbdVolID)
+		if err != nil {
+			util.ErrorLogMsg("failed to store volumeID in PV %s", err)
+			return err
+		}
 	}
 	return nil
 }
--- a/internal/rbd/rbd_journal.go
+++ b/internal/rbd/rbd_journal.go
@@ -25,6 +25,15 @@ import (
 	"github.com/ceph/ceph-csi/internal/util"
 )
 
+const (
+	// PVVolumeHandleAnnotationKey is the annotation key set on the PV object.
+	PVVolumeHandleAnnotationKey = "csi.ceph.io/volume-handle"
+	// PVReplicatedLabelKey is the label key set on PV object.
+	PVReplicatedLabelKey = "csi.ceph.io/replicated-volume"
+	// PVReplicatedLabelValue is the label value set on PV object.
+	PVReplicatedLabelValue = "volume-handle-detected"
+)
+
 func validateNonEmptyField(field, fieldName, structName string) error {
 	if field == "" {
 		return fmt.Errorf("value '%s' in '%s' structure cannot be empty", fieldName, structName)
@@ -479,10 +488,9 @@ func undoVolReservation(ctx context.Context, rbdVol *rbdVolume, cr *util.Credent
 // Extract uuid from volumeID
 // Reserve omap data
 // Generate new volume Handler
-// Create old volumeHandler to new handler mapping
 // The volume handler won't remain same as its contains poolID,clusterID etc
 // which are not same across clusters.
-func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName string, cr *util.Credentials) error {
+func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName string, cr *util.Credentials) (string, error) {
 	ctx := context.Background()
 	var (
 		options map[string]string
@@ -496,7 +504,7 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 
 	err := vi.DecomposeCSIID(rbdVol.VolID)
 	if err != nil {
-		return fmt.Errorf("%w: error decoding volume ID (%s) (%s)",
+		return "", fmt.Errorf("%w: error decoding volume ID (%s) (%s)",
 			ErrInvalidVolID, err, rbdVol.VolID)
 	}
 
@@ -507,13 +515,13 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 	rbdVol.Monitors, _, err = util.GetMonsAndClusterID(options)
 	if err != nil {
 		util.ErrorLog(ctx, "failed getting mons (%s)", err)
-		return err
+		return "", err
 	}
 
 	rbdVol.Pool = pool
 	err = rbdVol.Connect(cr)
 	if err != nil {
-		return err
+		return "", err
 	}
 	rbdVol.JournalPool = journalPool
 	if rbdVol.JournalPool == "" {
@@ -522,13 +530,13 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 	volJournal = journal.NewCSIVolumeJournal("default")
 	j, err := volJournal.Connect(rbdVol.Monitors, rbdVol.RadosNamespace, cr)
 	if err != nil {
-		return err
+		return "", err
 	}
 	defer j.Destroy()
 
 	journalPoolID, imagePoolID, err := util.GetPoolIDs(ctx, rbdVol.Monitors, rbdVol.JournalPool, rbdVol.Pool, cr)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	rbdVol.RequestName = requestName
@@ -538,7 +546,7 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 	imageData, err := j.CheckReservation(
 		ctx, rbdVol.JournalPool, rbdVol.RequestName, rbdVol.NamePrefix, "", kmsID)
 	if err != nil {
-		return err
+		return "", err
 	}
 	if imageData != nil {
 		rbdVol.ReservedID = imageData.ImageUUID
@@ -547,23 +555,23 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 		if rbdVol.ImageID == "" {
 			err = rbdVol.storeImageID(ctx, j)
 			if err != nil {
-				return err
+				return "", err
 			}
 		}
-		err = rbdVol.addNewUUIDMapping(ctx, imagePoolID, j)
-		if err != nil {
-			util.ErrorLog(ctx, "failed to add UUID mapping %s: %v", rbdVol, err)
-			return err
-		}
 		// As the omap already exists for this image ID return nil.
-		return nil
+		rbdVol.VolID, err = util.GenerateVolID(ctx, rbdVol.Monitors, cr, imagePoolID, rbdVol.Pool,
+			rbdVol.ClusterID, rbdVol.ReservedID, volIDVersion)
+		if err != nil {
+			return "", err
+		}
+		return rbdVol.VolID, nil
 	}
 
 	rbdVol.ReservedID, rbdVol.RbdImageName, err = j.ReserveName(
 		ctx, rbdVol.JournalPool, journalPoolID, rbdVol.Pool, imagePoolID,
 		rbdVol.RequestName, rbdVol.NamePrefix, "", kmsID, vi.ObjectUUID, rbdVol.Owner)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	defer func() {
@@ -578,7 +586,7 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 	rbdVol.VolID, err = util.GenerateVolID(ctx, rbdVol.Monitors, cr, imagePoolID, rbdVol.Pool,
 		rbdVol.ClusterID, rbdVol.ReservedID, volIDVersion)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	util.DebugLog(ctx, "re-generated Volume ID (%s) and image name (%s) for request name (%s)",
@@ -586,14 +594,11 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 	if rbdVol.ImageID == "" {
 		err = rbdVol.storeImageID(ctx, j)
 		if err != nil {
-			return err
+			return "", err
 		}
 	}
 
-	if volumeID != rbdVol.VolID {
-		return j.ReserveNewUUIDMapping(ctx, rbdVol.JournalPool, volumeID, rbdVol.VolID)
-	}
-	return nil
+	return rbdVol.VolID, nil
 }
 
 // storeImageID retrieves the image ID and stores it in OMAP.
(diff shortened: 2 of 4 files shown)


Madhu-1:
@nixpanic as discussed one more approach other than  https://github.com/ceph/ceph-csi/pull/1938 to fix this issue

internal/rbd/rbd_journal.go
```diff
@@ -25,6 +25,13 @@ import (
 	"github.com/ceph/ceph-csi/internal/util"
 )
 
+const (
+	// PVVolumeHandleLabel is the annotation key set on the PV object.
+	PVVolumeHandleLabel = "CephCSIVolumeHandle"
```
nixpanic: Labels are commonly formatted like `csi.ceph.io/volume-handle`. All lowercase, with a prefix (`csi.ceph.io/`) for the application/component, and sub-label (`volume-handle`) as real 'label'.

It would be good to adapt such a format here too.

internal/rbd/rbd_util.go
```diff
@@ -881,6 +867,31 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
 	return rbdVol, err
 }
 
+// genVolFromVolID generates a rbdVolume structure from the provided identifier, updating
+// the structure with elements from on-disk image metadata as well.
+func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials, secrets map[string]string) (*rbdVolume, error) {
+	vol, err := generateVolumeFromVolumeID(ctx, volumeID, cr, secrets)
+	if errors.Is(err, util.ErrKeyNotFound) || errors.Is(err, util.ErrPoolNotFound) {
```
nixpanic: Easier to do the `if` the other way around. A small body inside the statement returning `nil, err`, and continue with the rest outside the if-statement like it is the default behaviour.

When an error occurs, ideally return `nil` and an error, not some incomplete `vol` object. Make sure to call `vol.Destroy()` in error cases too.

internal/rbd/rbd_journal.go
```diff
@@ -496,7 +502,7 @@ func RegenerateJournal(imageName, volumeID, pool, journalPool, requestName strin
 
 	err := vi.DecomposeCSIID(rbdVol.VolID)
 	if err != nil {
-		return fmt.Errorf("%w: error decoding volume ID (%s) (%s)",
+		return rbdVol.VolID, fmt.Errorf("%w: error decoding volume ID (%s) (%s)",
```
nixpanic: In error cases, it is better to return an empty string.

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -122,6 +122,40 @@ func checkStaticVolume(pv *corev1.PersistentVolume) (bool, error) {
...
+		}
+	}
+	// check PV label exists
+	if _, ok := pv.Labels[rbd.PVReplicatedLabel]; ok {
+		replicationLabelExists = true
+	}
+	if replicationLabelExists && volumeIDExistInAnnotation {
+		return nil
+	}
+	pv.Labels[rbd.PVReplicatedLabel] = "true"
```
nixpanic: I'm not a big fan of `true` strings, maybe some status like `volume-handle-detected`? Just a nit, not required to change as the value is not *that* important.

nixpanic (commented):
I prefer this approach much more over #1936, thanks for posting this version!

At the moment you're watching for PVs, I hope this can be moved towards the VolumeReplication CR at a later time. From there you can then get the PVC and PV too.

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -122,6 +122,40 @@ func checkStaticVolume(pv *corev1.PersistentVolume) (bool, error) {
 	return static, nil
 }
 
+// storeVolumeIDInPV stores the new volumeID in PV object.
+func (r ReconcilePersistentVolume) storeVolumeIDInPV(pv *corev1.PersistentVolume, newVolumeID string) error {
+	// take a lock to avoid concurrent update of configmap
```
nixpanic: not sure what this comment means

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -122,6 +122,40 @@ func checkStaticVolume(pv *corev1.PersistentVolume) (bool, error) {
 	return static, nil
 }
 
+// storeVolumeIDInPV stores the new volumeID in PV object.
+func (r ReconcilePersistentVolume) storeVolumeIDInPV(pv *corev1.PersistentVolume, newVolumeID string) error {
```
nixpanic, line 128: There probably should be a verification that the PV is a Ceph-CSI one?

internal/rbd/rbd_journal.go
Madhu-1: This is for annotation sorry need to change the name. is it suggested to use `csi.ceph.io/volume-handle` for annotation too?

internal/rbd/rbd_util.go
Madhu-1: Need to return vol even in case of error to destroy the context and to release the lock

internal/rbd/rbd_journal.go
nixpanic: Yes, I think so. Identifying the application/component that set a Label or Annotation seems to be good practice.

internal/rbd/rbd_util.go
nixpanic: really? where is the lock taken then? Ideally it should be released at the same level...

Madhu-1: the lock was taken on the volumeID when and common function is added to cleanup it https://github.com/ceph/ceph-csi/pull/1924/files#diff-3792fcc7313c4d1e0b81c4d3ff7ac714b9d299130641673729196b2a4a1d6f2aR175-R186. if the PR is merged I will rework that one. Will fix it in this PR.

internal/controller/persistentvolume/persistentvolume.go
Madhu-1: The check already exists. the controller will reconcile the only PV created for this Driver.

internal/controller/persistentvolume/persistentvolume.go
Madhu-1: Leftover removing it

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -122,6 +122,39 @@ func checkStaticVolume(pv *corev1.PersistentVolume) (bool, error) {
...
+	if v, ok := pv.Annotations[rbd.PVVolumeHandleAnnotationKey]; ok {
+		if v == newVolumeID {
```
agarwal-mudit, line 130: if this is true, why will it go to Line 141?

Madhu-1: as its already existing in annotation, it will get return at line 129

agarwal-mudit: yeah, then why do we need to check it again in 141

internal/rbd/rbd_journal.go
```diff
@@ -25,6 +25,15 @@ import (
 	"github.com/ceph/ceph-csi/internal/util"
 )
 
+const (
+	// PVVolumeHandleAnnotationKey is the annotation key set on the PV object.
+	PVVolumeHandleAnnotationKey = "csi.ceph.io/volume-handle"
+	// PVReplicatedLabelKey is the label key set on PV object.
+	PVReplicatedLabelKey = "CephCSIReplicatedVolume"
```
nixpanic: Same as with annotations, use `csi.ceph.io/repliacted-volume` to follow the naming other projects do

ShyamsundarR:
Annotations are better than config maps.

@Madhu-1 and me discussed an alternate that provides configuration to the CSI driver for the required mapping, than annotate PVs. Madhu is thinking through the same, and if we discuss and find that a better long term alternative, we would possibly got that route.

Till then, without hopefully confusing matters further, this is an acceptable stand-in for the problem being resolved.

internal/rbd/rbd_journal.go
Madhu-1: done.

nixpanic (dismissed):
CI failures happened due to timeouts.

Just use `@mergifyio rebase` to re-test once a 2nd approval is in.

Madhu-1:
@humblec PTAL

humblec:
> @humblec PTAL

@Madhu-1 do we have any design doc (https://github.com/ceph/ceph-csi/pull/1938#issuecomment-810304643) for this async work in general ( may be one which cover the integration to the other component too ?) apart from the one available here https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/rbd-mirror.md for reference ? the one available here does not carry much information and I am finding difficulty to get the workflow , so the ask.

Madhu-1:
> > @humblec PTAL
> 
> @Madhu-1 do we have any design doc ([#1938 (comment)](https://github.com/ceph/ceph-csi/pull/1938#issuecomment-810304643)) for this async work in general ( may be one which cover the integration to the other component too ?) apart from the one available here https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/rbd-mirror.md for reference ? the one available here does not carry much information and I am finding difficulty to get the workflow , so the ask.

@humblec as i  already explain the currently planned design for you and its clear, this is the doc which I have created initially https://hackmd.io/1Pbj_9rzQ2SwK80GK6-qWg?both. i don't have the LLD

humblec:
> > > @humblec PTAL
> > 
> > 
> > @Madhu-1 do we have any design doc ([#1938 (comment)](https://github.com/ceph/ceph-csi/pull/1938#issuecomment-810304643)) for this async work in general ( may be one which cover the integration to the other component too ?) apart from the one available here https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/rbd-mirror.md for reference ? the one available here does not carry much information and I am finding difficulty to get the workflow , so the ask.
> 
> @humblec as i already explain the currently planned design for you and its clear, this is the doc which I have created initially https://hackmd.io/1Pbj_9rzQ2SwK80GK6-qWg?both. i don't have the LLD

@Madhu-1 looks like many changes  from the initial thoughts keep happening  and was struggling to keep updated on those. Thanks for the link. Hopefully it helps :) .

>https://github.com/ceph/ceph-csi/pull/1946#issuecomment-811115845
>

Are we saying the annotation method ( this PR) will be kept as a `temporary` one ?  if yes, what  are we planning as a permanent solution mentioned by above ?

Madhu-1:
> > > > @humblec PTAL
> > > 
> > > 
> > > @Madhu-1 do we have any design doc ([#1938 (comment)](https://github.com/ceph/ceph-csi/pull/1938#issuecomment-810304643)) for this async work in general ( may be one which cover the integration to the other component too ?) apart from the one available here https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/rbd-mirror.md for reference ? the one available here does not carry much information and I am finding difficulty to get the workflow , so the ask.
> > 
> > 
> > @humblec as i already explain the currently planned design for you and its clear, this is the doc which I have created initially https://hackmd.io/1Pbj_9rzQ2SwK80GK6-qWg?both. i don't have the LLD
> 
> @Madhu-1 looks like many changes from the initial thoughts keep happening and was struggling to keep updated on those. Thanks for the link. Hopefully it helps :) .
> 
> > [#1946 (comment)](https://github.com/ceph/ceph-csi/pull/1946#issuecomment-811115845)
> 
> Are we saying the annotation method ( this PR) will be kept as a `temporary` one ? if yes, what are we planning as a permanent solution mentioned by above ?

Yes am working on a new design need to check the feasibility, it takes some time (not soon) to get the design doc out.

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -161,11 +179,18 @@ func (r ReconcilePersistentVolume) reconcilePV(obj runtime.Object) error {
 		}
 		defer cr.DeleteCredentials()
 
-		err = rbd.RegenerateJournal(imageName, volumeHandler, pool, journalPool, requestName, cr)
```
humblec: @Madhu-1 we call `regeneratejournal` without any locking in place, so while reconciling a PV, if regenrateJournal() was called more than once, Isnt there a possibility of a race scenario ?

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -161,11 +179,18 @@ func (r ReconcilePersistentVolume) reconcilePV(obj runtime.Object) error {
...
+		rbdVolID, err := rbd.RegenerateJournal(imageName, volumeHandler, pool, journalPool, requestName, cr)
 		if err != nil {
 			util.ErrorLogMsg("failed to regenerate journal %s", err)
 			return err
 		}
+		if rbdVolID != volumeHandler {
+			err = r.storeVolumeIDInPV(pv, rbdVolID)
```
humblec: imo, the storeVolumeIDinPV should be a protected call. Otherwise the other parallel request may see a inconsistency in some scenarios.

internal/rbd/rbd_util.go
```diff
@@ -881,6 +867,32 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
...
+	}
+	pvlist, pErr := c.CoreV1().PersistentVolumes().List(context.TODO(), listOpt)
+	if pErr != nil {
+		return vol, pErr
+	}
+	for i := range pvlist.Items {
+		if pvlist.Items[i].Spec.CSI != nil && pvlist.Items[i].Spec.CSI.VolumeHandle == volumeID {
+			if v, ok := pvlist.Items[i].Annotations[PVVolumeHandleAnnotationKey]; ok {
+				util.UsefulLog(ctx, "found new volumeID %s for existing volumeID %s", v, volumeID)
+				return generateVolumeFromVolumeID(ctx, v, cr, secrets)
```
humblec, line 897: @Madhu-1 this is bit confusing. That said, Is it a problem if we see new volume ID for existing Volume ID?  can we add some source code comment about this block logic ?

Madhu-1: @humblec it's not a problem. add more comments

humblec:
@Madhu-1 looks like latest patchset address https://github.com/ceph/ceph-csi/pull/1946#discussion_r607932043,  is this ready for review or you are still polishing it ?

Madhu-1:
> @Madhu-1 looks like latest patchset address [#1946 (comment)](https://github.com/ceph/ceph-csi/pull/1946#discussion_r607932043), is this ready for review or you are still polishing it ?

This is ready for review

internal/rbd/rbd_util.go
```diff
@@ -883,6 +869,38 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
...
+// the structure with elements from on-disk image metadata as well.
+func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials, secrets map[string]string) (*rbdVolume, error) {
+	vol, err := generateVolumeFromVolumeID(ctx, volumeID, cr, secrets)
+	if !errors.Is(err, util.ErrKeyNotFound) && !errors.Is(err, util.ErrPoolNotFound) {
+		return vol, err
+	}
+
+	// If the volume details are not found in the OMAP it can be a mirrored RBD
+	// image and the OMAP is already generated and the volumeHandle might not
+	// be the same in the PV.Spec.CSI.VolumeHandle. Check the PV annotation for
```
humblec, line 882: if the volume details are not in OMAP , it could also be any static PV customer created in that cluster ( without mirroring), Isnt it ? why I am asking this is to confirm, we wont confuse with static PVs created by the admin and the mirrored PVs and thus land into performing different operations. or we are good to distinguish the PVs based on the key set by replication operator just for mirroring ?

internal/rbd/rbd_util.go
```diff
@@ -883,6 +869,38 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
...
+	// the new volumeHandle. If the new volumeHandle is found, generate the RBD
+	// volume structure from the new volumeHandle.
+	c := util.NewK8sClient()
+	listOpt := metav1.ListOptions{
+		LabelSelector: PVReplicatedLabelKey,
+	}
+	pvlist, pErr := c.CoreV1().PersistentVolumes().List(context.TODO(), listOpt)
```
humblec, line 889: This is a heavy operation, that said, listing out all the PVs in the cluster and fetching the volume handle looks to cause performance issues especially in large clusters,.

internal/rbd/rbd_util.go
```diff
@@ -777,9 +778,8 @@ func genSnapFromSnapID(ctx context.Context, rbdSnap *rbdSnapshot, snapshotID str
 	return err
 }
 
-// genVolFromVolID generates a rbdVolume structure from the provided identifier, updating
-// the structure with elements from on-disk image metadata as well.
-func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials, secrets map[string]string) (*rbdVolume, error) {
+// generateVolumeFromVolumeID generates a rbdVolume structure from the provided identifier.
+func generateVolumeFromVolumeID(ctx context.Context, volumeID string, cr *util.Credentials, secrets map[string]string) (*rbdVolume, error) {
```
humblec, line 782: nit: compared to previous scenarios, with the introduction of mirroring, iic, there is a possibility of volumeID being `nil`. If yes, DecompoeCSIID can be  avoided at start of this function if volumeID is nil.

internal/rbd/rbd_util.go
```diff
@@ -883,6 +869,38 @@ func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials,
...
+	c := util.NewK8sClient()
+	listOpt := metav1.ListOptions{
+		LabelSelector: PVReplicatedLabelKey,
+	}
+	pvlist, pErr := c.CoreV1().PersistentVolumes().List(context.TODO(), listOpt)
+	if pErr != nil {
+		return vol, pErr
+	}
+	for i := range pvlist.Items {
+		if pvlist.Items[i].Spec.CSI != nil && pvlist.Items[i].Spec.CSI.VolumeHandle == volumeID {
```
humblec, line 894: if the listOpt is `PVReplicatedLabelKey` why do we want to check whether its a CSI spec below?.

internal/rbd/rbd_util.go
Madhu-1: AFAIC this won't be called for static PVC. and we don't support any operator for static PVC except mounting and unmounting?

internal/rbd/rbd_util.go
Madhu-1: This will be fetched only if the omap is not found and we are fetching based on the selector. as this is a temp solution i don't see a problem with it.

internal/rbd/rbd_util.go
Madhu-1: @humblec can you please suggest in which scenario the volumeId can be nil?

internal/rbd/rbd_util.go
Madhu-1: It's safer check to check for nil pointer before accessing it.

internal/rbd/rbd_util.go
ShyamsundarR: Also, due to the label selector, the static PV created by users would be exempt from the selection. This should not be an issue.

internal/rbd/rbd_util.go
ShyamsundarR: Did a cursory check for `nil` possibility, did not find any.

ShyamsundarR (approved):

nixpanic (approved):

PR #1957, opened 2021-04-06
[RBD DR] Fix force promoteVolume

Madhu-1:
In the case of the DR, the image on the primary site cannot be demoted as the cluster is down, during failover the image needs to be force promoted. RBD returns `Device or resource busy` error message if the image cannot be promoted for the above reason.
Return FailedPrecondition so that the replication operator can send a request to force promote the volume.

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -283,11 +283,6 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 		return nil, err
 	}
-	// extract the force option
-	force, err := getForceOption(ctx, req.GetParameters())
-	if err != nil {
-		return nil, err
-	}
 
 	mirroringInfo, err := rbdVol.getImageMirroringInfo()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Madhu-1:
CI failures are related to deployment

agarwal-mudit (approved):

internal/rbd/replicationcontrollerserver.go
```diff
@@ -301,9 +296,18 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
...
+		err = rbdVol.promoteImage(req.Force)
 		if err != nil {
 			util.ErrorLog(ctx, err.Error())
+			// In case of the DR the image on the primary site cannot be
+			// demoted as the cluster is down, during failover the image need
+			// to be force promoted. RBD returns `Device or resource busy`
+			// error message if the image cannot be promoted for above reason.
+			// Return FailedPrecondition so that replication operator can send
+			// request to force promote the image.
+			if strings.Contains(err.Error(), "Device or resource busy") {
```
humblec, line 308: @Madhu-1  I understand,  this is validated at PromoteImage scenario here, however  just wondering is there a chance for some other scenario which could cause `Device or resource busy` error to be returned on an image while attempted. May be @dillaman can confirm

Madhu-1: I have seen this error when the promotion is not possible because the image was primary on another cluster and cannot be promoted without the force option.

nixpanic (approved):

Madhu-1:
@mergifyio rebase

PR #1959, opened 2021-04-07
cephfs: enable ceph-fuse big_writes by default

### LINKED ISSUE #1928: cephfs: run with fuse_big_writes by default

ceph-fuse in ceph-csi image v3.2.1 is linked against libfuse2:

```
$ kubectl exec -it ceph-csi-cephfs-nodeplugin-nv8c9 -c csi-cephfsplugin -- ceph-fuse --version
ceph version 15.2.8 (bdf3eebcd22d7d0b3dd4d5501bee5bac354d5b55) octopus (stable)
```
```
$ kubectl exec ceph-csi-cephfs-nodeplugin-nv8c9 -c csi-cephfsplugin -- dnf repoquery --requires --resolve ceph-fuse.x86_64
Last metadata expiration check: 0:02:19 ago on Wed Mar 17 15:36:58 2021.
fmt-0:6.2.1-1.el8.x86_64
fuse-0:2.9.7-12.el8.x86_64
fuse-libs-0:2.9.7-12.el8.x86_64
```

By default, the write buffer size in libfuse2 is 2KiB for reasons described here:

https://github.com/libfuse/libfuse/blob/d44bf3a4ac1ce1b56672ac9ae4a5caa3a8ee6844/ChangeLog#L653-L659

`-o big_writes` mount option is used to override this limit. libfuse3 removes `big_writes`, making it enabled by default: https://github.com/libfuse/libfuse/blob/master/ChangeLog.rst#libfuse-300-2016-12-08 , so this change will be incorporated automatically by ceph-fuse as well, see https://github.com/ceph/ceph/pull/34531 , Octopus backport https://github.com/ceph/ceph/pull/34769 .

I propose that until we get ceph-fuse with libfuse3 in the image, `fuse_big_writes` option should be added to ceph.conf

Without enabling this option, the write performance is very poor: https://tracker.ceph.com/issues/44885

Granted, users could add this option into their `mountOptions` or `fuseMountOptions` parameters, but are there reasons to not have it on by default?

Test write performance with and without having `big_writes` enabled while still using libfuse2.

# Additional context #

None.

Rakshith-R:
By default, the write buffer size in libfuse2 is 2KiB `-o big_writes` option is used to override this limit to increase performance.
This commit makes `-o big_writes` option as default for ceph-fuse mounter.

Closes: #1928

### DIFF
--- a/internal/util/cephconf.go
+++ b/internal/util/cephconf.go
@@ -28,6 +28,11 @@ auth_client_required = cephx
 
 # Workaround for http://tracker.ceph.com/issues/23446
 fuse_set_user_groups = false
+
+# ceph-fuse which uses libfuse2 by default has write buffer size of 2KiB
+# adding 'fuse_big_writes = true' option by default to override this limit
+# see https://github.com/ceph/ceph-csi/issues/1928
+fuse_big_writes = true
 `)
 
 const (


Rakshith-R:

Rook deployment failed

nixpanic (dismissed):

Madhu-1:
LGTM. will give some time for @gman0 to review this one.

gman0:
@Rakshith-R maybe a brief comment describing why is this option being added, but otherwise LGTM, thanks!

Madhu-1 (dismissed):

humblec:
The codewise this looks good. However I am not sure whether this option enabling by default good or bad or neutral . Keeping  @idryomov @dillaman in loop,  if they can add some inputs on this

nixpanic (dismissed):
New versions of Ceph enable this by default, and libfuse3 does so as well. Soon (if not already) it will be enabled by default in out container images anyway. I guess  it is good to have, and should not hurt to enable it explicitly.

gman0:
After having a second look, please see this:

https://github.com/ceph/ceph/blob/10a86af09d60a741ea211dc0ebb56ffa4c3c82bd/src/client/fuse_ll.cc#L1262-L1267

When parsing ceph.conf, the ceph-fuse client treats `fuse big writes = true` option as a no-op if it's linked against libfuse >=v3.0.0. For earlier versions it passes `-o big_writes` to libfuse.

Since `-o big_writes` is actually removed from libfuse3, it could fail as it's not a recognized option anymore? I don't have a way to test v3 at hand so I can't say for sure.

Maybe a safer option would be to add `fuse big writes = true` to https://github.com/ceph/ceph-csi/blob/devel/internal/util/cephconf.go#L24-L31 instead of setting it in the mount opts.

Madhu-1:
cc @kotreshhr can you please provide your suggestion on this one from cephfs side?

Madhu-1:
added DNM till we come to a conclusion on this one to avoid merging.

humblec:
> After having a second look, please see this:
> 
> https://github.com/ceph/ceph/blob/10a86af09d60a741ea211dc0ebb56ffa4c3c82bd/src/client/fuse_ll.cc#L1262-L1267
> 
> When parsing ceph.conf, the ceph-fuse client treats `fuse big writes = true` option as a no-op if it's linked against libfuse >=v3.0.0. For earlier versions it passes `-o big_writes` to libfuse.

Nice catch @gman0 :+1: 
> 
> Since `-o big_writes` is actually removed from libfuse3, it could fail as it's not a recognized option anymore? I don't have a way to test v3 at hand so I can't say for sure.
> 
> Maybe a safer option would be to add `fuse big writes = true` to https://github.com/ceph/ceph-csi/blob/devel/internal/util/cephconf.go#L24-L31 instead of setting it in the mount opts.

Yeah, that would be better I believe. Lets get a quick input from ceph folks as well @idryomov  @dillaman @batrick @kotreshhr

batrick:
> > Since `-o big_writes` is actually removed from libfuse3, it could fail as it's not a recognized option anymore? I don't have a way to test v3 at hand so I can't say for sure.
> > Maybe a safer option would be to add `fuse big writes = true` to https://github.com/ceph/ceph-csi/blob/devel/internal/util/cephconf.go#L24-L31 instead of setting it in the mount opts.
> 
> Yeah, that would be better I believe. Lets get a quick input from ceph folks as well @idryomov @dillaman @batrick @kotreshhr

You can set the ceph config harmless. It will just be unused if libfuse >= 3.0.

Madhu-1 (approved):

nixpanic (approved):

PR #1977, opened 2021-04-15
e2e: fix error handling for goerr113 linter

### LINKED ISSUE #1227: Address goerr113 warnings
`goerr113` is a golang linter to check the errors handling expressions.
It warns about using wrapped static errors in place of dynamic at multiple places.
Issue reported:  `err113: do not define dynamic errors, use wrapped static errors instead` 

Steps:
1. Enable `goerr113` linter from `scripts/golangci.yml.in` (if #1157 is merged)
2. Run `make containerized-test TARGET=go-lint`

goerr133 doesn't report any more warnings.

Also, see : #1157

yati1998:
goerr113 is a golang linter to check the errors handling
expressions. It warns about using wrapped static errors in
place of dynamic at multiple places. Issue reported: err113:
do not define dynamic errors, use wrapped static errors instead
So, err113 reports every == and != comparison for exact error type
variables except comparison to nil and io.EOF.
Also, any call of errors.New() and fmt.Errorf() methods are reported
except the calls used to initialise package-level variables and the
fmt.Errorf() calls wrapping the other errors.

Updates: #1227

### DIFF
--- a/e2e/rbd_helper.go
+++ b/e2e/rbd_helper.go
@@ -30,6 +30,40 @@ func rbdOptions(pool string) string {
 	return "--pool=" + pool
 }
 
+type errGettingFsid string
+type errListingRBDNs string
+type errCreatingRBDNs string
+type errMsg string
+type getOmapVal string
+type errReadPassphrase string
+type errPassphraseInVault string
+type errEncryptedState string
+
+func (e errGettingFsid) Error() string {
+	return fmt.Sprintf("error getting fsid %v", string(e))
+}
+func (e errListingRBDNs) Error() string {
+	return fmt.Sprintf("error listing rbd namespace %v", string(e))
+}
+func (e errCreatingRBDNs) Error() string {
+	return fmt.Sprintf("error creating rbd namespace %v", string(e))
+}
+func (e errMsg) Error() string {
+	return fmt.Sprintf("%v", string(e))
+}
+func (e getOmapVal) Error() string {
+	return fmt.Sprintf("failed to getomapval %v", string(e))
+}
+func (e errReadPassphrase) Error() string {
+	return fmt.Sprintf("failed to read passphrase from vault: %s", string(e))
+}
+func (e errPassphraseInVault) Error() string {
+	return fmt.Sprintf("passphrase found in vault while should be deleted: %s", string(e))
+}
+func (e errEncryptedState) Error() string {
+	return fmt.Sprintf("%v not equal to encrypted", string(e))
+}
+
 func createRBDStorageClass(c kubernetes.Interface, f *framework.Framework, scOptions, parameters map[string]string, policy v1.PersistentVolumeReclaimPolicy) error {
 	scPath := fmt.Sprintf("%s/%s", rbdExamplePath, "storageclass.yaml")
 	sc, err := getStorageClass(scPath)
@@ -51,7 +85,7 @@ func createRBDStorageClass(c kubernetes.Interface, f *framework.Framework, scOpt
 		return err
 	}
 	if stdErr != "" {
-		return fmt.Errorf("error getting fsid %v", stdErr)
+		return errGettingFsid(stdErr)
 	}
 	// remove new line present in fsID
 	fsID = strings.Trim(fsID, "\n")
@@ -84,7 +118,7 @@ func createRadosNamespace(f *framework.Framework) error {
 		return err
 	}
 	if stdErr != "" {
-		return fmt.Errorf("error listing rbd namespace %v", stdErr)
+		return errListingRBDNs(stdErr)
 	}
 	if !strings.Contains(stdOut, radosNamespace) {
 		_, stdErr, err = execCommandInToolBoxPod(f,
@@ -93,7 +127,7 @@ func createRadosNamespace(f *framework.Framework) error {
 			return err
 		}
 		if stdErr != "" {
-			return fmt.Errorf("error creating rbd namespace %v", stdErr)
+			return errCreatingRBDNs(stdErr)
 		}
 	}
 	stdOut, stdErr, err = execCommandInToolBoxPod(f,
@@ -102,7 +136,7 @@ func createRadosNamespace(f *framework.Framework) error {
 		return err
 	}
 	if stdErr != "" {
-		return fmt.Errorf("error listing rbd namespace %v", stdErr)
+		return errListingRBDNs(stdErr)
 	}
 
 	if !strings.Contains(stdOut, radosNamespace) {
@@ -112,7 +146,7 @@ func createRadosNamespace(f *framework.Framework) error {
 			return err
 		}
 		if stdErr != "" {
-			return fmt.Errorf("error creating rbd namespace %v", stdErr)
+			return errCreatingRBDNs(stdErr)
 		}
 	}
 	return nil
@@ -176,7 +210,7 @@ func getImageMeta(rbdImageSpec, metaKey string, f *framework.Framework) (string,
 		return "", err
 	}
 	if stdErr != "" {
-		return strings.TrimSpace(stdOut), fmt.Errorf(stdErr)
+		return strings.TrimSpace(stdOut), errMsg(stdErr)
 	}
 	return strings.TrimSpace(stdOut), nil
 }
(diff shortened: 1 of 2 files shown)


e2e/rbd_helper.go
```diff
@@ -30,6 +30,40 @@ func rbdOptions(pool string) string {
 	return "--pool=" + pool
 }
 
+type errGettingFsid string
+type errListingRBDNs string
```
Yuggupta27, line 34:
```suggestion
type errListRBDNs string
```

e2e/rbd_helper.go
```diff
@@ -30,6 +30,40 @@ func rbdOptions(pool string) string {
 	return "--pool=" + pool
 }
 
+type errGettingFsid string
```
Yuggupta27, line 33:
```suggestion
type errGetFsid string
```

e2e/rbd_helper.go
```diff
@@ -30,6 +30,40 @@ func rbdOptions(pool string) string {
...
+type errListingRBDNs string
+type errCreatingRBDNs string
```
Yuggupta27, line 35:
```suggestion
type errCreateRBDNs string
```

e2e/rbd_helper.go
```diff
@@ -30,6 +30,40 @@ func rbdOptions(pool string) string {
...
+type errMsg string
+type getOmapVal string
```
Yuggupta27, line 37: can we replace it with `errGetOmapVal` to match the theme of other errors?

Yuggupta27 (changes requested):
Using Verbs in place of gerunds looks cleaner IMO. Thoughts?

e2e/rbd_helper.go
yati1998: yes. Will change this.

yati1998:
> Using Verbs in place of gerunds looks cleaner IMO. Thoughts?

Yes True. Thanks for the suggestion.

yati1998:
@Madhu-1 @nixpanic would like to get a review from you guys on this PR. Do comment your suggestions. If this is fine with you all, I will follow the same in other files and make the changes.

e2e/rbd_helper.go
```diff
@@ -30,6 +30,40 @@ func rbdOptions(pool string) string {
...
+type errReadPassphrase string
```
nixpanic, line 38: Instead of creating new types, it maybe you can create new static errors like

```golang
var (
    errGetFsid = errors.New("error getting fsid")
)
```

And return that in functions, or use

```golang
return nil, fmt.Errorf("could not fund fsid for %q: %w", rbdVol, errGetFsid)
```

Having new types for all error strings seems a little much to me.

nixpanic (changes requested):

yati1998:
As suggested by @Madhu-1  keeping this at hold for sometimes. It's gonna be lots of changes if we go to resolve the go-linter error in all the e2e files. So Currently, looking for the way to enable goerr113 in only internal files and disable in e2e to avoid these changes.
We can also create static errors for all the errors in one file and use it in all the places.
What's your thought in this @nixpanic @Yuggupta27 .

Rakshith-R:
>So Currently, looking for the way to enable goerr113 in only internal files and disable in e2e to avoid these changes.

The latest version of golangci-lint provides an option to have the following config 
```yaml
issues:
  # Excluding configuration per-path, per-linter, per-text and per-source
  exclude-rules:
    # Exclude known linters from partially hard-vendored code,
    # which is impossible to exclude via "nolint" comments.
    - path: e2e/
      text: "e2e"
      linters:
        - goerr113
```
refernce: https://golangci-lint.run/usage/configuration/#config-file

yati1998:
> > So Currently, looking for the way to enable goerr113 in only internal files and disable in e2e to avoid these changes.
> 
> The latest version of golangci-lint provides an option to have the following config
> 
> ```yaml
> issues:
>   # Excluding configuration per-path, per-linter, per-text and per-source
>   exclude-rules:
>     # Exclude known linters from partially hard-vendored code,
>     # which is impossible to exclude via "nolint" comments.
>     - path: e2e/
>       text: "e2e"
>       linters:
>         - goerr113
> ```
> 
> refernce: https://golangci-lint.run/usage/configuration/#config-file

Thanks @Rakshith-R for the reference.

Yuggupta27:
> As suggested by @Madhu-1 keeping this at hold for sometimes. It's gonna be lots of changes if we go to resolve the go-linter error in all the e2e files. So Currently, looking for the way to enable goerr113 in only internal files and disable in e2e to avoid these changes.
> We can also create static errors for all the errors in one file and use it in all the places.
> What's your thought in this @nixpanic @Yuggupta27 .

Sure, I'm fine with resolving goerr13 warnings only of internal code for now. Just make sure that the issue is open so that TODO for doing the same for e2e is tracked.

nixpanic:
Creating a `error.go` file per package, and place all the static errors (not new error types) in there is probably helpful. It will make it easier to re-use errors and prevents duplication. When you do this, please move existing errors from other files into the `error.go` file too.

Addressing the linter partially (for `internal/`) and enforcing it is definitely a good step. Make sure to add a 

    # TODO: enable linter for all code, see ceph/ceph-csi#1227

comment in the section where it gets excluded. We should still fix the issues later, as all code in the repository should be of equal high qualify.

Also, send lots of small PRs that are quick to get merged. That way you reduce the chance that you need to manually rebase and repost when other changes get included.

yati1998:
> Creating a `error.go` file per package, and place all the static errors (not new error types) in there is probably helpful. It will make it easier to re-use errors and prevents duplication. When you do this, please move existing errors from other files into the `error.go` file too.
> 
> Addressing the linter partially (for `internal/`) and enforcing it is definitely a good step. Make sure to add a
> 
> ```
> # TODO: enable linter for all code, see ceph/ceph-csi#1227
> ```
> 
> comment in the section where it gets excluded. We should still fix the issues later, as all code in the repository should be of equal high qualify.
> 
> Also, send lots of small PRs that are quick to get merged. That way you reduce the chance that you need to manually rebase and repost when other changes get included.

sure @nixpanic . Thanks for your suggestion

e2e/rbd_helper.go
yati1998: @nixpanic in this case suggested by you, we will miss the actual error string that was printed by `stdErr` , hence, I see creating new types will be better.

yati1998:
@nixpanic and @Madhu-1 would like to get your reviews on this PR so that I can proceed with other files. Please ignore the ci error and merge conflicts. I will resolve them after we finalize the way for this. IMO, creating types for the errors and using it where ever needed, is better. 
Creating static errors(as suggested by @nixpanic ) doesn't seem feasible to me, mentioned the reason above.
I can move ahead with this, if you guys are fine with it orelse suggestions are most welcomed.

Madhu-1:
> @nixpanic and @Madhu-1 would like to get your reviews on this PR so that I can proceed with other files. Please ignore the ci error and merge conflicts. I will resolve them after we finalize the way for this. IMO, creating types for the errors and using it where ever needed, is better. Creating static errors(as suggested by @nixpanic ) doesn't seem feasible to me, mentioned the reason above. I can move ahead with this, if you guys are fine with it orelse suggestions are most welcomed.

For me this looks like an extra overhead. am fine not fixing this linter in code base.

yati1998:
> > @nixpanic and @Madhu-1 would like to get your reviews on this PR so that I can proceed with other files. Please ignore the ci error and merge conflicts. I will resolve them after we finalize the way for this. IMO, creating types for the errors and using it where ever needed, is better. Creating static errors(as suggested by @nixpanic ) doesn't seem feasible to me, mentioned the reason above. I can move ahead with this, if you guys are fine with it orelse suggestions are most welcomed.
> 
> For me this looks like an extra overhead. am fine not fixing this linter in code base.

In that case, I would remove the TODO from the linter and close the issue https://github.com/ceph/ceph-csi/issues/1227 with the explanation.

nixpanic:
> In that case, I would remove the TODO from the linter and close the issue #1227 with the explanation.

Instead of removing the TODO comment, leave a shirt description with link to the issue that has the explanation.

We might be able to do better when we introduce new errors.

yati1998:
Closing this PR: https://github.com/ceph/ceph-csi/issues/1227#issuecomment-1068957831

PR #1978, opened 2021-04-15
Fix rbd Volume Resync and omap error handling for DR

Madhu-1:
This PR fixes the below items
* incase if the image is promoted and demoted, the image state will be set to up+unknown if the image on the remote cluster is still in the demoted state.  when the user changes the state from primary to secondary and still the image is in the demoted (secondary) state in the remote cluster. the image state on both the cluster will be in an unknown state.

* rbd: check volumeID in PV if image not found
If the pool or few keys are missing in the omap. The GetImageAttributes function returns a nil error message and few empty items in imageAttributes struct. if the image is not found then use the volumeId present on the PV annotation and check we are able to get the OMAP data from new ID for further operations.

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -872,7 +872,6 @@ func generateVolumeFromVolumeID(ctx context.Context, volumeID string, cr *util.C
 	if err != nil {
 		return rbdVol, err
 	}
-
 	rbdVol.RequestName = imageAttributes.RequestName
 	rbdVol.RbdImageName = imageAttributes.ImageName
 	rbdVol.ReservedID = vi.ObjectUUID
@@ -908,7 +907,7 @@ func generateVolumeFromVolumeID(ctx context.Context, volumeID string, cr *util.C
 // the structure with elements from on-disk image metadata as well.
 func genVolFromVolID(ctx context.Context, volumeID string, cr *util.Credentials, secrets map[string]string) (*rbdVolume, error) {
 	vol, err := generateVolumeFromVolumeID(ctx, volumeID, cr, secrets)
-	if !errors.Is(err, util.ErrKeyNotFound) && !errors.Is(err, util.ErrPoolNotFound) {
+	if !errors.Is(err, util.ErrKeyNotFound) && !errors.Is(err, util.ErrPoolNotFound) && !errors.Is(err, ErrImageNotFound) {
 		return vol, err
 	}
 
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -52,6 +52,10 @@ const (
 	// another cluster
 	upAndStopped imageMirroringState = "up+stopped"
 
+	// If the state is up+unknown means the rbd-mirror daemon is
+	// running and the image is demoted on both the clusters.
+	upAndUnknown imageMirroringState = "up+unknown"
+
 	// If the state is error means image need resync.
 	errorState imageMirroringState = "error"
 )
@@ -375,6 +379,8 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 // ResyncVolume extracts the RBD volume information from the volumeID, If the
 // image is present, mirroring is enabled and the image is in demoted state.
 // If yes it will resync the image to correct the split-brain.
+// nolint:gocyclo // reduce complexity
+// FIXME: reduce complexity.
 func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	req *replication.ResyncVolumeRequest,
 ) (*replication.ResyncVolumeResponse, error) {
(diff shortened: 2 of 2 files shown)


humblec:
@Madhu-1 this is required for release 3.3, Isnt it ?

Madhu-1:
> @Madhu-1 this is required for release 3.3, Isnt it ?

Yes @humblec required for 3.3

internal/util/errors.go
```diff
@@ -33,6 +33,9 @@ var (
 	ErrSnapNameConflict = errors.New("snapshot name conflict")
 	// ErrPoolNotFound is returned when pool is not found.
 	ErrPoolNotFound = errors.New("pool not found")
+
+	// ErrIncompleteDataInOMAP is returned when image details are not found in pool.
+	ErrIncompleteDataInOMAP = errors.New("incomplete data in  OMAP")
```
humblec: nit: a space extra before OMAP.

internal/util/errors.go
```diff
@@ -33,6 +33,9 @@ var (
 	ErrSnapNameConflict = errors.New("snapshot name conflict")
 	// ErrPoolNotFound is returned when pool is not found.
 	ErrPoolNotFound = errors.New("pool not found")
+
+	// ErrIncompleteDataInOMAP is returned when image details are not found in pool.
```
humblec: When image details are not found in OMAP....Isnt it ?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -450,6 +455,19 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		}
 	}
 
+	// when the images are demoted on both clusters and user requests for the
+	// resync of the image, the image mirror state will be unknown state in
+	// both clusters.
+	if state == upAndUnknown {
+		ready = true
```
agarwal-mudit, line 463: We don't need to set it true here.
I am thinking of a scenario when
1. it was false because of the above code
2. comes here and becomes true
3. never goes to 466
this will result in ready becoming "true" even though it should be "false"

Ideally if ready is "false" because of 453, we should not even check this.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -52,6 +52,10 @@ const (
 	// another cluster
 	upAndStopped imageMirroringState = "up+stopped"
 
+	// If the state is up+unknown means the rbd-mirror daemon is
```
agarwal-mudit, line 55: nit: If the state is up+unknown it means that the rbd-mirror daemon is running and the image is demoted on both the clusters.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -450,6 +455,19 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		}
 	}
 
+	// when the images are demoted on both clusters and user requests for the
+	// resync of the image, the image mirror state will be unknown state in
```
agarwal-mudit, line 460: nit: the image mirror state will be "unknown" in both the clusters.

internal/rbd/rbd_util.go
```diff
@@ -872,7 +872,11 @@ func generateVolumeFromVolumeID(ctx context.Context, volumeID string, cr *util.C
 	if err != nil {
 		return rbdVol, err
 	}
-
+	// GetImageAttributes return error as nil if the pool or some keys are missing
+	// in OMAP.validate the data before proceeding.
```
agarwal-mudit: nit: space after the first sentence in the comment

internal/rbd/replicationcontrollerserver.go
Madhu-1: this is set to make it true when the image state is unknown in both cases. it will be set to false even if one of the remote status is not up and unknown.  if the state is set at 453 we are not going to enter this loop.

internal/util/errors.go
Madhu-1: Yes correct. updated it now.

internal/rbd/replicationcontrollerserver.go
agarwal-mudit: Yeah, it will not come here in that case. sorry for the confusion.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -450,6 +455,19 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+	// both clusters.
+	if state == upAndUnknown {
+		ready = true
+		for _, s := range mirrorStatus.PeerSites {
+			if imageMirroringState(s.State) != upAndUnknown {
+				util.UsefulLog(ctx, "peer site name=%s mirroring state=%s, description=%s and lastUpdate=%s", s.SiteName, s.State, s.Description, s.LastUpdate)
```
humblec: @madhu-1, may be add a comma after peer site name?

Madhu-1: sure

internal/rbd/replicationcontrollerserver.go
```diff
@@ -375,6 +379,7 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 // ResyncVolume extracts the RBD volume information from the volumeID, If the
 // image is present, mirroring is enabled and the image is in demoted state.
 // If yes it will resync the image to correct the split-brain.
+// nolint:gocyclo // reduce complexity
```
Yuggupta27, line 382: using `FIXME: reduce complexity.` will allow it to also get listed in https://www.tickgit.com/browse?repo=github.com/ceph/ceph-csi so that it is easier to track.

Yuggupta27 (changes requested):

internal/util/errors.go
```diff
@@ -33,6 +33,9 @@ var (
 	ErrSnapNameConflict = errors.New("snapshot name conflict")
 	// ErrPoolNotFound is returned when pool is not found.
 	ErrPoolNotFound = errors.New("pool not found")
+
```
Yuggupta27: According to the file convention, Newline is not required.

Yuggupta27 (changes requested):

agarwal-mudit (dismissed):

internal/util/errors.go
humblec: @Madhu-1 ^^

Yuggupta27 (approved):

agarwal-mudit (dismissed):

humblec (dismissed):

humblec (approved):

humblec:
@agarwal-mudit @Yuggupta27  ptal

agarwal-mudit (approved):

PR #1991, opened 2021-04-19
rbd: modified logic to check image watchers

Madhu-1:
Before RBD map operation, we do check the watchers on the RBD image. In the case of RWO volume. cephcsi makes sure only one client is using the RBD image. If the rbd image is mirrored, by default mirroring daemon will add a watcher on the image and as we are using go-ceph a watcher will be added as we have opened the image So we will have two watchers on an image if
mirroring is enabled. This holds when the rbd mirror daemon is running, In case if the mirror daemon is not running there will be only one watcher on the rbd image (which is placed by go-ceph image open) we should not block the map operation if the mirroring daemon is not running as its Async mirroring. This commit adds a check to make sure no more than 2 watchers
if the image is mirrored or no more than 1 watcher if it is not mirrored image.

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -439,10 +439,11 @@ func (rv *rbdVolume) isInUse() (bool, error) {
 	// because we opened the image, there is at least one watcher
 	defaultWatchers := 1
 	if rv.Primary {
-		// a watcher will be added by the rbd mirror daemon if the image is primary
+		// if rbd mirror daemon is running, a watcher will be added by the rbd
+		// mirror daemon for mirrored images.
 		defaultWatchers++
 	}
-	return len(watchers) != defaultWatchers, nil
+	return len(watchers) > defaultWatchers, nil
 }
 
 // addRbdManagerTask adds a ceph manager task to execute command


internal/rbd/rbd_util.go
```diff
@@ -439,10 +439,11 @@ func (rv *rbdVolume) isInUse() (bool, error) {
 	// because we opened the image, there is at least one watcher
 	defaultWatchers := 1
 	if rv.Primary {
-		// a watcher will be added by the rbd mirror daemon if the image is primary
+		// if rbd mirror daemon is running, a watcher will be added by the rbd
+		// mirror daemon for mirrored images.
```
nixpanic, line 443: Looks good, this should be an improvement.

However, I wonder of there is an issue when mirroring has not been started (or is currently active), an other rbd client could connect to the image. Should this be something to look into?

nixpanic (approved):

internal/rbd/rbd_util.go
Madhu-1: Yes just had the same discussion with @ShyamsundarR, It's difficult to identify the watcher is added by rbd mirror daemon or not. this issue exists only for RBD mirrored PVC and the watcher is extra protection from the cephcsi side.

Madhu-1:

 Apr 19 13:39:34.710: failed to create vault configmap error running /usr/bin/kubectl --server=https://xx.xx.xx --kubeconfig=/root/.kube/config --namespace=cephcsi-e2e-f731ad1d create --namespace=cephcsi-e2e-f731ad1d -f -:
    Command stdout:
    stderr:
    Error from server: error when creating "STDIN": etcdserver: request timed out
    error:
    exit status 1

internal/rbd/rbd_util.go
ShyamsundarR: Other than this observation LGTM!

ShyamsundarR (approved):

PR #1999, opened 2021-04-21
deploy: parameterize cephfs filesystem name

humblec:
This patch parameterize or make a variable for rook cephfs fs
name which we use in rook deployment

### DIFF
--- a/scripts/rook.sh
+++ b/scripts/rook.sh
@@ -4,6 +4,8 @@ ROOK_VERSION=${ROOK_VERSION:-"v1.4.9"}
 ROOK_DEPLOY_TIMEOUT=${ROOK_DEPLOY_TIMEOUT:-300}
 ROOK_URL="https://raw.githubusercontent.com/rook/rook/${ROOK_VERSION}/cluster/examples/kubernetes/ceph"
 ROOK_BLOCK_POOL_NAME=${ROOK_BLOCK_POOL_NAME:-"newrbdpool"}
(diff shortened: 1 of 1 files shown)


scripts/rook.sh
```diff
@@ -4,6 +4,8 @@ ROOK_VERSION=${ROOK_VERSION:-"v1.4.9"}
 ROOK_DEPLOY_TIMEOUT=${ROOK_DEPLOY_TIMEOUT:-300}
 ROOK_URL="https://raw.githubusercontent.com/rook/rook/${ROOK_VERSION}/cluster/examples/kubernetes/ceph"
 ROOK_BLOCK_POOL_NAME=${ROOK_BLOCK_POOL_NAME:-"newrbdpool"}
+ROOK_FILESYSTEM_NAME=${ROOK_FILESYSTEM_NAME:-"myfs"}
```
Madhu-1, line 7: if you want to create a filesystem with this name, it needs to be patch/sed filesyste-test.yaml file then this will be useful

Madhu-1 (changes requested):

scripts/rook.sh
humblec: That can be choice, but the variable introduction with the default value is fine imo.

Madhu-1: but what is the use of a new variable? IMO the new variable makes sense if we are adding it to the filesystem-test.yaml

Issue #1963, opened 2021-04-08
Validate e2e testing on Kubernetes 1.21


nixpanic:
# Describe the feature you'd like to have #

Kubernetes 1.21 is currently in Beta and we should start testing with this version too.

Admins in the Jenkins environment can manually start the mini-e2e jobs with a given Kubernetes version, doing some pre-check runs.

Once Kubernetes 1.21 has been released, the version should be added as a regular job. The `.mergify.yml` should then also require the tests before merging.

Jobs are available for testing:

* https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.21/activity/
* https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.21/activity/

nixpanic:
#1962 has not been complete, the `scripts/skip-doc-change.sh` script does not detect non-doc changes if there is no PR associated with the CI job run. It will be needed to skip the check when `git_since` is the same as the checked out branch.

nixpanic:
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.20/detail/mini-e2e_k8s-1.20/639/pipeline is a new manual run with Kubernetes 1.21.

nixpanic:
> https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.20/detail/mini-e2e_k8s-1.20/639/pipeline is a new manual run with Kubernetes 1.21.

This failed with some minikube errors during cluster creation. Maybe the new Kubernetes version needs more resources?

nixpanic:
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/k8s-e2e-external-storage/detail/k8s-e2e-external-storage/9/pipeline is an other test, running the k8s-e2e-external-storage job with Kubernetes 1.21.

nixpanic:
[mini-e2e-helm_k8s-1.21/1](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.21/detail/mini-e2e-helm_k8s-1.21/1/pipeline) failed in **STEP: create a PVC and bind it to an app with encrypted RBD volume with VaultKMS**:

[...]

PR #2001, opened 2021-04-21
rbd: return crypt error for the rpc return

humblec:
at present we return the volume connect error if the clone
from snapshot fails when rbdvolume is encrypted, which is incorrect.
This patch correctly return the failed copy encryption error to the
caller

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -811,13 +811,9 @@ func cloneFromSnapshot(ctx context.Context, rbdVol *rbdVolume, rbdSnap *rbdSnaps
 	if rbdVol.isEncrypted() {
 		// FIXME: vol.VolID should be different from rbdVol.VolID
 		vol.VolID = rbdVol.VolID
-		cryptErr := rbdVol.copyEncryptionConfig(&vol.rbdImage)
-		if cryptErr != nil {
-			util.WarningLog(ctx, "failed copy encryption "+
-				"config for %q: %v", vol.String(),
-				rbdSnap.RequestName, cryptErr)
-			return nil, status.Errorf(codes.Internal,
-				err.Error())
+		err = rbdVol.copyEncryptionConfig(&vol.rbdImage)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
 		}
 	}
 


Madhu-1 (dismissed):

nixpanic (changes requested):
Instead of using `cryptErr`, it probably is just as simple to use `err` everywhere. There is no other contents in `err` anyway. Also, consider removing the log.

I think this was copy/pasted from an other location where there could have been two different errors.

humblec:
> Instead of using `cryptErr`, it probably is just as simple to use `err` everywhere. There is no other contents in `err` anyway. Also, consider removing the log.
> Instead of using `cryptErr`, it probably is just as simple to use `err` everywhere. There is no other contents in `err` anyway. Also, consider removing the log.
> 

> 
> I think this was copy/pasted from an other location where there could have been two different errors.

@nixpanic the cloneFromSnapshot() make use of a new error var uErr for undoSnapshotCloning() just above this one and doSnapshot() also distinguish the error to a new var like cryptErr in the same source file. so , imo, keeping a clear or in parity error variable would be better. wdyt? 

I agree we can remove the extra logging :+1: 

> I think this was copy/pasted from an other location where there could have been two different errors.

nixpanic:
> @nixpanic the cloneFromSnapshot() make use of a new error var uErr for undoSnapshotCloning() just above this one and doSnapshot() also distinguish the error to a new var like cryptErr in the same source file. so , imo, keeping a clear or in parity error variable would be better. wdyt?

I prefer to keep things simple. Having a single error variable like we do in most cases is the easiest to understand, and that makes it more difficult to make mistakes like the one fixed in this PR. In the cases where two errors are possible, the 2nd error should be logged, and the 1st error returned. That obviously requires two errors, but all other cases should be able to use a single one.

humblec:
@nixpanic comments addressed . PTAL. Thanks.

internal/rbd/controllerserver.go
```diff
@@ -811,13 +811,9 @@ func cloneFromSnapshot(ctx context.Context, rbdVol *rbdVolume, rbdSnap *rbdSnaps
...
-		cryptErr := rbdVol.copyEncryptionConfig(&vol.rbdImage)
-		if cryptErr != nil {
-			util.WarningLog(ctx, "failed copy encryption "+
-				"config for %q: %v", vol.String(),
-				rbdSnap.RequestName, cryptErr)
-			return nil, status.Errorf(codes.Internal,
-				err.Error())
+		err = rbdVol.copyEncryptionConfig(&vol.rbdImage)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, err.Error())
```
Madhu-1: as no formatting is required, it should be status.Error()

Madhu-1 (dismissed):

internal/rbd/controllerserver.go
humblec: @Madhu-1 sure.. done.. ptal. thanks

nixpanic (approved):

Madhu-1 (approved):

nixpanic:
> /test ci/centos/mini-e2e/k8s-1.21

Just running for stability check, see #1963

PR #2002, opened 2021-04-21
doc: correct the keys in certificate secrets

humblec:

### DIFF
--- a/docs/design/proposals/encryption-with-vault-tokens.md
+++ b/docs/design/proposals/encryption-with-vault-tokens.md
@@ -233,20 +233,19 @@ stringData:
     KNSC2DVVXcCgkqcCgAwIBAgIwewrvx...
 ```
 
-Its also possible that a user can create a single secret for the certificates
-and update the configuration to fetch certificates from a secret.
+It is also possible for a user to create a single Secret that contains both
+the client authentication and update the configuration to fetch the certificate
+and key from the Secret.
 
 ```yaml
 ---
 apiVersion: v1
 kind: secret
 metadata:
-  name: vault-certificates
+  name: vault-client-auth
 stringData:
   cert: |
     MIIC2DCCAcCgAwIBAgIBATANBgkqh...
-  cert: |
-    BATANBgkqcCgAwIBAgIBATANBAwI...
   key: |
     KNSC2DVVXcCgkqcCgAwIBAgIwewrvx...
 ```
--- a/internal/util/vault_tokens.go
+++ b/internal/util/vault_tokens.go
@@ -464,7 +464,7 @@ func (kms *VaultTokensKMS) parseTenantConfig() error {
 		return nil
 	}
 
-	// fetch the ConfigMap from the tanants namespace
+	// fetch the ConfigMap from the tenants namespace
 	c := NewK8sClient()
 	cm, err := c.CoreV1().ConfigMaps(kms.Tenant).Get(context.TODO(),
 		kms.ConfigName, metav1.GetOptions{})


docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -243,9 +243,9 @@ kind: secret
 metadata:
   name: vault-certificates
 stringData:
-  cert: |
+  cert1: |
```
nixpanic: Can there be multiple certificates in the Secret? Or should this Secret contain only a `cert` and `key`?

nixpanic (commented):
Also, why `deploy:` as prefix of the PR if this is only a doc change?

docs/design/proposals/encryption-with-vault-tokens.md
humblec: definitely more than one key:value can be part of the secret. So, its fine.

humblec: regarding more than one cert will be respected by code, I have to check.  @Madhu-1 22ae4a0b16acaa16642dff09101f38d7cceb2a20 looks like changed this key names from `ca.cert` and `tls.cert` to `cert` and `cert` .  Is this valid configuration ? because looking at this commit code, it only look for one key which is `cert` .

humblec:
> Also, why `deploy:` as prefix of the PR if this is only a doc change?

@nixpanic this been addressed. ptal.

docs/design/proposals/encryption-with-vault-tokens.md
nixpanic: I'd like to know the answer of this, before approving the change.

nixpanic (dismissed):

docs/design/proposals/encryption-with-vault-tokens.md
Madhu-1: the secret data part is a map[string]string we cannot have duplication keys, I think we need to remove adding multiple certs to the same secret.

Madhu-1: the key should always remain const `cert` we cannot have cert1 and cert2 here.

humblec: Yeah, we should have uniqueness in key and only one key is respected in code, so adjusted the same. @nixpanic @Madhu-1 ptal

humblec:
@Madhu-1 @nixpanic ptal.

nixpanic (approved):

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -245,8 +245,6 @@ metadata:
 stringData:
   cert: |
     MIIC2DCCAcCgAwIBAgIBATANBgkqh...
```
Madhu-1, line 248: This doesn't seems valid to me as we cannot have both certs in single secret. I think we need to remove this section
@nixpanic wdyt?

nixpanic: Probably the phrasing should be changed if it is confusing? The idea is to have the key and certificate in the same Secret. Maybe rename the combined Secret from `vault-certificates` to `vault-client-auth`. 

The phrase above should then be modified similar from

> a single secret for the certificates and update the configuration to fetch certificates from a secret and update the configuration to fetch certificates

to 

> a single Secret that contains both the client authentication and update the configuration to fetch the certificate and key from the Secret

nixpanic (dismissed):

docs/design/proposals/encryption-with-vault-tokens.md
Madhu-1: sound good!. we can have only one cert in the secret not more than that

humblec:
@nixpanic @Madhu-1 ptal.

Madhu-1 (dismissed):

humblec:
@nixpanic ptal . Thanks

docs/design/proposals/encryption-with-vault-tokens.md
```diff
@@ -233,20 +233,19 @@ stringData:
     KNSC2DVVXcCgkqcCgAwIBAgIwewrvx...
 ```
 
-Its also possible that a user can create a single secret for the certificates
-and update the configuration to fetch certificates from a secret.
+Its also possible that a user can create a single Secret that contains both
```
nixpanic: `Its` -> `It is` (preferred) or `It's`

Maybe better:

> It is also possible for a user to create ...

and end the sentence with a `.`?

nixpanic (commented):
Nit: I don't feel very strongly about it, but when you're addressing documentation shortcomings, it should probably be done right.

docs/design/proposals/encryption-with-vault-tokens.md
humblec: @nixpanic done . PTAL.

nixpanic (approved):

humblec:
@nixpanic can we please add `ready-to-merge` here?

PR #2028, opened 2021-04-28
rbd: mark image as ready when  image states is up+unknown

Madhu-1:
To recover from the split-brain (up+error) state the image needs to be demoted and requested for resync on site-a and then the image on site-b should get demoted. The volume should be marked to ready=true when the image state on both the clusters are up+unknown because during the last snapshot syncing the data gets copied first and then image state on the
site-a changes to up+unknown. If the image state on both the sites are up+unknown consider that complete data is synced as the last snapshot gets exchanged between the clusters.

# Different scenarios for Resync completed

## Trigger resync when image on other site is in promoted state

* create 10 GB of file and validate the data after resync

* Do Failover when the site-a goes down
* Force promotes the image and write data in GiB
* Once the site-a comes back, Demote the image and issue resync
* Demote the image on site-b
* The status will get reflected on the other site when the last
  snapshot sync happens
* The image will go to up+unknown state. and complete data will
  be copied to site a
* Promote the image on site-a and use it

```bash
csi-vol-5633715e-a7eb-11eb-bebb-0242ac110006:
  global_id:   e7f9ec55-06ab-46cb-a1ae-784be75ed96d
  state:       up+unknown
  description: remote image demoted
  service:     a on minicluster1
  last_update: 2021-04-28 07:11:56
  peer_sites:
    name: e47e29f4-96e8-44ed-b6c6-edf15c5a91d6-rook-ceph
    state: up+unknown
    description: remote image demoted
    last_update: 2021-04-28 07:11:41
 ```

## Trigger resync when image on other site is in demoted state

* Do Failover when the site-a goes down
* Force promotes the image on site-b and write data in GiB
* Demote the image on site-b
* Once the site-a comes back, Demote the image on site-a
* The images on both site will go to split-brain state

```bash
csi-vol-37effcb5-a7f1-11eb-bebb-0242ac110006:
  global_id:   115c3df9-3d4f-4c04-93a7-531b82155ddf
  state:       up+error
  description: split-brain
  service:     a on minicluster2
  last_update: 2021-04-28 07:25:41
  peer_sites:
    name: abbda0f0-0117-4425-8cb2-deb4c853da47-rook-ceph
    state: up+error
    description: split-brain
    last_update: 2021-04-28 07:25:26
```
* Issue resync
* The images cannot be resynced because when we issue resync
  on site a the image on site-b was in a demoted state
* To recover from this state (promote and then demote the
  the image on site-b after some time)

```bash
[...]

PR #2045, opened 2021-05-03
Rbd: Fix image flattening when snapshot and parent PVC is deleted

### LINKED ISSUE #2044: RBD: Failed to mount cloned PVC when parent PVC and snapshot is deleted
With kernel v3.10.x failed to mount the PVC when the parent PVC and snapshot is deleted

```
  Normal   Scheduled               10s              default-scheduler        Successfully assigned default/csirbd-demo-pod to minikube
  Normal   SuccessfulAttachVolume  10s              attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-ebfbcb08-41ee-4042-bcb9-24b4d7b096da"
  Warning  FailedMount             3s (x4 over 7s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-ebfbcb08-41ee-4042-bcb9-24b4d7b096da" : rpc error: code = Internal desc = image not found: RBD image not found

```

steps to reproduce the issue

* Create PVC
* Create Snapshot from PVC
* Delete PVC
* Create Clone from snapshot
* Delete snapshot
* Mount PVC clone to application pod
* Delete PVC clone and application pod

Madhu-1:
For flattening we call checkImageChainHasFeature which internally calls to `getImageInfo` returns the parent name even if the parent is in the trash when we try to open the parent image to get its information it fails as the image not found. we should treat the error as nil if the parent is not found.

flatten the image if the deep-flatten feature is present on the images in the chain or if the images in the chain are not zero, as we cannot check the deep-flatten feature of the images which are in the trash.
    
resolves #2044 
resolves #1544 

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -1088,6 +1088,112 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("validate PVC mounting if snapshot and parent PVC are deleted", func() {
+				// snapshot beta is only supported from v1.17+
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
+					Skip("pvc restore is only supported from v1.17+")
+				}
+				err := createRBDSnapshotClass(f)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+				defer func() {
+					err = deleteRBDSnapshotClass()
+					if err != nil {
+						e2elog.Failf("failed to delete VolumeSnapshotClass: %v", err)
+					}
+				}()
+
+				// create PVC and bind it to an app
+				pvc, err := loadPVC(pvcPath)
+				if err != nil {
+					e2elog.Failf("failed to load PVC with error %v", err)
+				}
+
+				pvc.Namespace = f.UniqueName
+				app, err := loadApp(appPath)
+				if err != nil {
+					e2elog.Failf("failed to load application with error %v", err)
+				}
+				app.Namespace = f.UniqueName
+				err = createPVCAndApp("", f, pvc, app, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create PVC and application with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 1)
+
+				snap := getSnapshot(snapshotPath)
+				snap.Namespace = f.UniqueName
+				snap.Spec.Source.PersistentVolumeClaimName = &pvc.Name
+
+				err = createSnapshot(&snap, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create snapshot with error %v", err)
+				}
+				// validate created backend rbd images
+				// parent PVC + snapshot
+				totalImages := 2
+				validateRBDImageCount(f, totalImages)
+				pvcClone, err := loadPVC(pvcClonePath)
+				if err != nil {
+					e2elog.Failf("failed to load PVC with error %v", err)
+				}
+
+				// delete parent PVC
+				err = deletePVCAndApp("", f, pvc, app)
+				if err != nil {
+					e2elog.Failf("failed to delete PVC and application with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 1)
+
+				// create clone PVC
+				pvcClone.Namespace = f.UniqueName
+				err = createPVCAndvalidatePV(f.ClientSet, pvcClone, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create PVC with error %v", err)
+				}
+				// validate created backend rbd images = snapshot + clone
+				totalImages = 2
+				validateRBDImageCount(f, totalImages)
+
+				// delete snapshot
+				err = deleteSnapshot(&snap, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete snapshot with error %v", err)
+				}
+
+				// validate created backend rbd images = clone
+				totalImages = 1
+				validateRBDImageCount(f, totalImages)
+
+				appClone, err := loadApp(appClonePath)
+				if err != nil {
+					e2elog.Failf("failed to load application with error %v", err)
+				}
+				appClone.Namespace = f.UniqueName
+				appClone.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = pvcClone.Name
+
+				// create application
+				err = createApp(f.ClientSet, appClone, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create application with error %v", err)
+				}
+
+				err = deletePod(appClone.Name, appClone.Namespace, f.ClientSet, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete application with error %v", err)
+				}
+				// delete PVC clone
+				err = deletePVCAndValidatePV(f.ClientSet, pvcClone, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete PVC with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0)
+			})
+
 			By("ensuring all operations will work within a rados namespace", func() {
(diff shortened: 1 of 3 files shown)


internal/rbd/rbd_util.go
```diff
@@ -719,7 +719,14 @@ func (rv *rbdVolume) checkImageChainHasFeature(ctx context.Context, feature uint
 		}
 		err = vol.getImageInfo()
 		if err != nil {
-			util.ErrorLog(ctx, "failed to get image info for %s: %s", vol, err)
+			// call to getImageInfo returns the parent name even if the parent
+			// is in the trash, when we try to open the parent image to get its
+			// information it fails because it is already in trash. We should
+			// treat error as nil if the parent is not found.
+			if errors.Is(err, ErrImageNotFound) {
```
Yuggupta27, line 726: Just a question, In case Parent image is not found, why isn't this caught in `getImageInfo()` here? https://github.com/ceph/ceph-csi/blob/8d7c66363daa007c9c28085629b123beb197ece4/internal/rbd/rbd_util.go#L1162

Madhu-1: `parentInfo, err := image.GetParent()`  will return the parent details even if the parent is in trash, and above `ErrNotFound` will get hit only if the parent is not deleted completely. when you try to call `getImageInfo` for parent again as we are checking the images in the chain it will fail with  `ErrImageNotFound` in rv.open() function call

Yuggupta27: Got it! Thanks for explaining :)

e2e/rbd.go
```diff
@@ -1088,6 +1088,111 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("validate PVC mounting if snapshot and parent PVC are deleted", func() {
+				// snapshot beta is only supported from v1.17+
+				if k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
```
nixpanic: Nit: can you add the test like this?
```golang
                                if !k8sVersionGreaterEquals(f.ClientSet, 1, 16) {
                                        Skip("pvc clone is only supported from v1.16+")
                                }
```

That way, indention is a little less, and it makes it easier to drop support for deprecated Kubernetes versions at a later time.

internal/rbd/rbd_util.go
```diff
@@ -726,7 +726,7 @@ func (rv *rbdVolume) checkImageChainHasFeature(ctx context.Context, feature uint
 			if errors.Is(err, ErrImageNotFound) {
 				return false, nil
 			}
-			util.ErrorLog(ctx, "failed to get image info for %s: %s", vol, err)
+			util.ErrorLog(ctx, "failed to get image info for %s: %s", vol.String(), err)
```
nixpanic, line 729: `.String()` gets called implicitly when using the `%s` formatting. I do not think this commit makes a difference?

Madhu-1: No, currently the log looks like below,
```text
21-04-30T23:21:52.581603843Z E0430 23:21:52.581535   24684 rbd_util.go:589] ID: 2896 Req-ID: 0001-0011-rook-ceph-0000000000000001-6fc62615-aa0a-11eb-b36c-0a580a800411 failed to get image info for {%!s(*[]util.TopologyConstrainedPool=<nil>) <nil> map[] csi-snap-bec4c10d-aa09-11eb-b36c-0a580a800411   172.30.41.228:6789,172.30.166.153:6789,172.30.55.186:6789  replicapool   csi-snap-bec4c10d-aa09-11eb-b36c-0a580a800411 %!s(rbd.FeatureSet=257)           %!s(int64=10737418240) %!s(bool=false) %!s(bool=false) %!s(bool=false) %!s(bool=false) <nil>  seconds:1619824736 nanos:697795370 %!s(*util.ClusterConnection=&{0xc0001a71e0 0xc000111620}) %!s(*rados.IOContext=&{0x7f0490078b20})}: image not found: RBD image not found
```
to fix this one am explicitly calling the String() function

Madhu-1: This is because the String function is implemented as a pointer method, let me change it to value type it should also fix the issue. sample code https://play.golang.org/p/2vCy5-a6B3m

Madhu-1: okay looks like am getting a CI error if I consider pass by value
```
hugeParam: ri is heavy (208 bytes); consider passing it by pointer (gocritic)
```

nixpanic: Hmm, maybe it is needed to pass `&vol` to the log function to get the `.String()` automatic?
See https://play.golang.org/p/VlyKSYbO3uv for my modified test.

I do not really care that much though, using `.String()` is fine, as we have that in many other places already.

nixpanic (approved):

Yuggupta27 (approved):

internal/rbd/rbd_util.go
nixpanic: See #2051 for an issue that can be used to cleanup the `.String()` calling.

Madhu-1:

created https://github.com/ceph/ceph-csi/issues/2052 to track it

internal/rbd/rbd_util.go
humblec: @Madhu-1 just to confirm, returning the image info even when an image is in trash can be avoided with any options in Ceph side ? just trying to understand is it on core side or a limitation in above layer.

Madhu-1: the above cannot be avoided as when you ask for parent details using go-ceph `GetParent` function it provides the parent details but it won't return any details like the image is in the trash or not.
It's possible to check the image is in the trash but it will be a costly operator as we need to list all the images in the trash and do a comparison.

Madhu-1:
> /test ci/centos/mini-e2e-helm/k8s-1.20

Failure due to https://github.com/ceph/ceph-csi/issues/2052

internal/rbd/rbd_util.go
humblec: > the above cannot be avoided as when you ask for parent details using go-ceph `GetParent` function it provides the parent details but it won't return any details like the image is in the trash or not.

Ok.. so looks like its a go ceph limitation then? if thats the case, I believe the image in trash or not is kind of a good enhancement we need on GetParent call or related calls. Because it can help us in various code paths in CSI layer and avoid some logic errors/complexities.  wdyt @Madhu-1 ?

Madhu-1: can you open an issue in go-ceph and see if it's possible to get this information?

humblec: > can you open an issue in go-ceph and see if it's possible to get this information?

Sure..!

Madhu-1: @humblec anything else is required on this PR? if not I would like to get it in.

humblec (approved):

internal/rbd/rbd_util.go
humblec: Lets get this in with priority. Thanks :+1:

nixpanic:
btw, is this a replacement for #1544 from @pkalever?

Madhu-1:
> /test ci/centos/mini-e2e/k8s-1.20

`stdErr occurred: 2021-05-04T10:56:41.047+0000 7efe90bf4500  0 monclient(hunting): authenticate timed out after 300` error

Madhu-1:
> /test ci/centos/mini-e2e-helm/k8s-1.20

`May  4 11:55:06.679: FAIL: failed to create storageclass with error error getting fsid 2021-05-04T10:55:00.009+0000 7f344d47f700  0 monclient(hunting): authenticate timed out after 300` error

nixpanic:
No need to retry testing for now, just do a `mergifyio rebase` when #2028 is merged,

Madhu-1:
> btw, is this a replacement for #1544 from @pkalever?

Yes this will replace #1544

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio rebase

Madhu-1:
E2E failed for known errors

Issue #1875, opened 2021-02-17
e2e tests fail with "monclient(hunting): authenticate timed out after 300"


nixpanic:
# Describe the bug #

Occasionally e2e tests fail with messages similar to:
```
Feb 16 16:32:40.944: INFO: ExecWithOptions {Command:[/bin/sh -c rbd ls --format=json --pool=replicapool] Namespace:rook-ceph PodName:rook-ceph-tools-5949d6759-sc49f ContainerName:rook-ceph-tools Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Feb 16 16:32:40.944: INFO: >>> kubeConfig: /root/.kube/config
Feb 16 16:37:46.299: INFO: stdErr occurred: 2021-02-16T16:37:41.815+0000 7f64d00cc500  0 monclient(hunting): authenticate timed out after 300

Feb 16 16:37:46.299: FAIL: failed to list rbd images with error failed to list images 2021-02-16T16:37:41.815+0000 7f64d00cc500  0 monclient(hunting): authenticate timed out after 300
```

It seems `kubectl` is used to call an `rbd` command, which then fails to authenticate to the Ceph cluster.

nixpanic:
Happened at https://github.com/ceph/ceph-csi/pull/1869#issuecomment-780415172

nixpanic:
An other occurrence at https://github.com/ceph/ceph-csi/pull/1873#issuecomment-780080222

nixpanic:
And at https://github.com/ceph/ceph-csi/pull/1675#issuecomment-778749379

nixpanic:
https://github.com/ceph/ceph-csi/pull/1013#issuecomment-780497604 as well

Rakshith-R:
This can be closed since we are not seeing it occur anymore and it was probably due to `mon allowing insecure id` issue https://github.com/rook/rook/issues/7746

PR #2064, opened 2021-05-10
build: update go version to 1.16 in go.mod

humblec:
Make go version latest in the repo

### DIFF
--- a/scripts/check-env.sh
+++ b/scripts/check-env.sh
@@ -60,9 +60,9 @@ else
 fi
 
 # parse the Golang version, return the digit passed as argument
-# 1.13.9 -> go_version 1 -> 1
-# 1.13.9 -> go_version 2 -> 13
-# 1.13.9 -> go_version 3 -> 9
+# 1.16.4 -> go_version 1 -> 1
+# 1.16.4 -> go_version 2 -> 16
+# 1.16.4 -> go_version 3 -> 4
 go_version() {
 	go version | cut -d' ' -f3 | sed 's/^go//' | cut -d'.' -f"${1}"
 }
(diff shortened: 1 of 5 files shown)


Madhu-1 (dismissed):
I don't see a change in build.env,  dont we need to update the development guide?

scripts/check-env.sh
```diff
@@ -60,9 +60,9 @@ else
 fi
 
 # parse the Golang version, return the digit passed as argument
-# 1.13.9 -> go_version 1 -> 1
-# 1.13.9 -> go_version 2 -> 13
-# 1.13.9 -> go_version 3 -> 9
+# 1.16.4 -> go_version 1 -> 1
+# 1.16.4 -> go_version 2 -> 16
+# 1.16.4 -> go_version 3 -> 4
```
nixpanic, line 65: there is really no need to update the example in this script

nixpanic (dismissed):
No objections from me. We do not provide packages that should be used by other projects, so we won't break any consumers.

scripts/check-env.sh
humblec: Yeah. but its ok I believe.

humblec:
@Madhu-1 ptal. Thanks

humblec:
failure due to https://github.com/ceph/ceph-csi/issues/1875

Madhu-1 (dismissed):

humblec:
@Madhu-1 @nixpanic ptal

Madhu-1 (approved):

humblec:
@Madhu-1 as there is no change, can we add `ready-to-merge` here?

Issue #1969, opened 2021-04-09
e2e testing in CI fails with: Waiting up to 3m0s for all (but 0) nodes to be ready


nixpanic:
There are more regularly jobs failing with the following error:

```
Waiting up to 3m0s for all (but 0) nodes to be ready
```

It is unclear why this happens. Minikube seems to become unresponsive (or really slow?) and the e2e test (which is a Kubernetes client) can not continue. Logs from Minikube are not available, as the e2e test fails to get details through the Kubernetes API.

nixpanic:
https://github.com/ceph/ceph-csi/pull/1966#issuecomment-816478728

nixpanic:
After #2058 the jobs seems way more stable. No failures related to the timeout have been reported yet.

nixpanic:
This is not completely fixed yet. #2062 hit this problem :disappointed: 

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/k8s-e2e-external-storage-1.20/51/display/redirect

nixpanic:
In case of a failure, we should probably run a script that gathers:

- status of the minikube VM in libvirt
- dmesg/journal output of the minikube VM
- `top -b -c -n1` from the VM

nixpanic:
New job has started at https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/mini-e2e-helm_k8s-1.20/836/display/redirect . Hopefully the logs give an indication why something goes wrong with minikube.

nixpanic:
Also updating the ceph/ceph:v16 container image in the CI registry to the current v16.2.3 release. Maybe it is better in CPU handling, #2073 suggests that a Ceph process consumes 100% CPU.

nixpanic:
There is also a suspicion that kubernetes 1.19 CI jobs are less affected than the 1.20 jobs...

The statistics for 1.19 shows :sun_behind_small_cloud: , where as 1.20 has a :cloud_with_rain:
 - https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/pipelines

PR #2068, opened 2021-05-11
ci: run system-status.sh in case a job fails

nixpanic:
The new `system-status.sh` script logs the status of the host and the
minikube VM. This gets executed when a CI job fails, and should aid in
troubleshooting spurious failures.

Updates: #1969

---

<details>

  progress or results

</details>

### DIFF
--- a/mini-e2e.groovy
+++ b/mini-e2e.groovy
@@ -171,6 +171,12 @@ node('cico-workspace') {
 		}
 	}
 
+	catch (err) {
+		stage('log system status') {
+			ssh './system-status.sh'
+		}
+	}
+
 	finally {
 		stage('return bare-metal machine') {
 			sh 'cico node done ${CICO_SSID}'
--- a/system-status.sh
+++ b/system-status.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+#
+# Run this script to gather details about the environment where the CI job is
+# running. This can be helpful to identify issues why minikube failed to
+# deploy, or tests encounter problems while running.
+#
+
+function minikube_ssh() {
+    ssh \
+        -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no \
+        -l docker -i "$(minikube ssh-key)" \
+        "$(minikube ip)" "${*}"
+}
+
+function log() {
+    echo "###"
+    echo "### going to execute: ${*}"
+    echo "###"
+    "${@}"
+    echo "###"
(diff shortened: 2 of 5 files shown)


system-status.sh
```diff
@@ -0,0 +1,39 @@
...
+    echo "###"
+}
+
+# get the status of the VM in libvirt
+log virsh list
+
+# status of the minikube Kubernetes cluster
+log minikube status
+
+# get the status of processes in the VM
```
Madhu-1, line 32: want to run minikube logs also?

system-status.sh
```diff
@@ -0,0 +1,39 @@
...
+        -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no \
+        -l docker -i "$(minikube ssh-key)" \
+        "$(minikube ip)" "${*}"
+}
+
+function log() {
+    echo "###"
+    echo "### going to execute: ${*}"
+    echo "###"
+    ${*}
```
Yuggupta27: Add quotes here to avoid whitespace problems?

Yuggupta27 (changes requested):

system-status.sh
nixpanic: will include that

system-status.sh
nixpanic: with quotes the executable gets interpreted like `virsh status` instead of `virsh` with `status` as argument. Will check what is accepted by ShellCheck and works.

mini-e2e.groovy
```diff
@@ -171,6 +171,12 @@ node('cico-workspace') {
 		}
 	}
 
+	catch (err) {
+		stage('log system status') {
+			ssh './system-status.sh'
```
Yuggupta27, line 176: Just a question, is the script executable by default? Or is it safer to use `ssh 'sh system-status.sh'`?

Madhu-1 (approved):

mini-e2e.groovy
nixpanic: Not by default, but the permissions are stored in the git repository. You can check the permissions in the [patch format](https://github.com/ceph/ceph-csi/pull/2068.patch) of this PR, it contains the `0755` that indicates it is executable:
```
diff --git a/system-status.sh b/system-status.sh
new file mode 100755
```

Yuggupta27: Thanks for the clarification!

Yuggupta27 (approved):

PR #2070, opened 2021-05-11
ci: in case of a failure, return error after logging system status

nixpanic:
It seems that it is required to re-throw the error after a catch{..}
block. Without this, and a successful execution of system-status.sh, the
CI jobs get marked as SUCCESS, even when there was a failure.

Fixes: e36155283 "ci: run system-status.sh in case a job fails"

---

<details>

  progress or results

</details>

### DIFF
--- a/k8s-e2e-external-storage.groovy
+++ b/k8s-e2e-external-storage.groovy
@@ -10,6 +10,7 @@ def doc_change = 0
 def k8s_release = 'latest'
 def namespace = 'k8s-e2e-storage-' + UUID.randomUUID().toString().split('-')[-1]
 def ci_registry = 'registry-ceph-csi.apps.ocp.ci.centos.org'
+def failure = null
 
 def ssh(cmd) {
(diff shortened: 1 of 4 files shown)


Madhu-1 (approved):

Yuggupta27:
Can be script be made to always return error? Just because it runs only in case of failure?

Yuggupta27:
Just because it runs only in case of failure?

nixpanic:
> Can be script be made to always return error? Just because it runs only in case of failure?

I guess that could be done, nice idea. It would make the script less re-usable though. Maybe we need to run it after minikube deployment and so we have a functional state we can use to compare against.

Yuggupta27:
> > Can be script be made to always return error? Just because it runs only in case of failure?
> 
> I guess that could be done, nice idea. It would make the script less re-usable though. Maybe we need to run it after minikube deployment and so we have a functional state we can use to compare against.

Agreed. That can be a great use case for the approach used now.

PR #2078, opened 2021-05-12
ci: get the logs from the Ceph cluster pods

nixpanic:
It seems that `/var/log/rook` inside the VM does not contain any files.
Getting the logs from the Pods through kubectl may not be as stable, but
it should get some logs when minikube is still/again available.

Updates: #1969

---

<details>

  progress or results

</details>

### DIFF
--- a/system-status.sh
+++ b/system-status.sh
@@ -44,7 +44,11 @@ log minikube_ssh sudo tar c /var/lib/rook | tar xvO
 
 # gets status of the Rook deployment
 log kubectl -n rook-ceph get pods
-log kubectl -n rook-ceph describe pods
+for POD in $(kubectl -n rook-ceph get pod -l rook_cluster=rook-ceph -o jsonpath='{.items[0].metadata.name}')
+do
+    log kubectl -n rook-ceph describe pod "${POD}"
+    log kubectl -n rook-ceph logs "${POD}"
+done
 log kubectl -n rook-ceph describe CephCluster
 log kubectl -n rook-ceph describe CephBlockPool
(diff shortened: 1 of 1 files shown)


humblec:
> It seems that `/var/log/rook` inside the VM does not contain any files.
> Getting the logs from the Pods through kubectl may not be as stable, but
> it should get some logs when minikube is still/again available.
> 
> Updates: #1969
> Show available bot commands

@nixpanic any example log capture here for reference which proved to be helping?

nixpanic:
> @nixpanic any example log capture here for reference which proved to be helping?

search for `tar c /var/lib/rook` in https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.20/runs/859/nodes/107/log/ , that only outputs `/var/lib/rook` which does not seem to contain any files:

```
###
### going to execute: minikube_ssh sudo tar c /var/lib/rook
###
Warning: Permanently added '192.168.39.72' (ECDSA) to the list of known hosts.
tar: Removing leading `/' from member names
###
### execution finished: minikube_ssh sudo tar c /var/lib/rook
###
var/lib/rook
```

(note that the ordering is a little wrong, writing parts to stdout and stderr)

humblec (approved):

humblec:
various CI failures were tried to study today to see whats causing the CI failure without luck. Looks like definitely we need more logging to understand . Added `ready to merge` as this extra logging could help us in that direction. thanks:!

PR #2098, opened 2021-05-24
cleanup: address ifshort & ineffectual assignment linter issues

Rakshith-R:

-  cleanup: address ineffectual assignement linter issue
- cleanup: address ifshort linter issues

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/rbd/rbd_journal.go
+++ b/internal/rbd/rbd_journal.go
@@ -278,7 +278,7 @@ func (rv *rbdVolume) Exists(ctx context.Context, parentVol *rbdVolume) (bool, er
 
 	// NOTE: Return volsize should be on-disk volsize, not request vol size, so
 	// save it for size checks before fetching image data
-	requestSize := rv.VolSize
+	requestSize := rv.VolSize //nolint:ifshort // FIXME: rename and split function into helpers
 	// Fetch on-disk image attributes and compare against request
 	err = rv.getImageInfo()
 	if err != nil {
--- a/internal/util/errors_test.go
+++ b/internal/util/errors_test.go
@@ -64,8 +64,7 @@ func TestJoinErrors(t *testing.T) {
 	assertErrorIs(x, errFoo, false)
 	assertErrorIs(x, errBar, true)
 	s1 := "w{w{w{x}}: w{w{foo: bar}}}"
-	s2 := w1w2Xw2FooBar.Error()
-	if s1 != s2 {
+	if s2 := w1w2Xw2FooBar.Error(); s1 != s2 {
 		t.Errorf("%s != %s", s1, s2)
 	}
 }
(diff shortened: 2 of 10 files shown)


yati1998 (commented):
Looks Good to me.

internal/rbd/rbd_journal.go
```diff
@@ -278,7 +278,7 @@ func (rv *rbdVolume) Exists(ctx context.Context, parentVol *rbdVolume) (bool, er
 
 	// NOTE: Return volsize should be on-disk volsize, not request vol size, so
 	// save it for size checks before fetching image data
-	requestSize := rv.VolSize
+	requestSize := rv.VolSize //nolint:ifshort
```
nixpanic: Is this really required? Maybe leave a comment that ifshort fails to detect this (little strange) usage.
The function does a lot more than just checking `Exists()`, so maybe it would be good to rename and split the function in helpers that make this `nolint` annotation unneeded. Can be marked as a `FIXME` in the nolint-comment.

internal/rbd/rbd_journal.go
```diff
@@ -278,7 +278,7 @@ func (rv *rbdVolume) Exists(ctx context.Context, parentVol *rbdVolume) (bool, er
 
 	// NOTE: Return volsize should be on-disk volsize, not request vol size, so
 	// save it for size checks before fetching image data
-	requestSize := rv.VolSize
+	requestSize := rv.VolSize //nolint:ifshort // FIXME: rename and split function into helpers
```
Yuggupta27, line 281:
```suggestion
      // FIXME: rename and split function into helpers
	requestSize := rv.VolSize // nolint:ifshort 
```

Yuggupta27 (changes requested):

internal/util/errors_test.go
```diff
@@ -64,8 +64,7 @@ func TestJoinErrors(t *testing.T) {
 	assertErrorIs(x, errFoo, false)
 	assertErrorIs(x, errBar, true)
 	s1 := "w{w{w{x}}: w{w{foo: bar}}}"
-	s2 := w1w2Xw2FooBar.Error()
-	if s1 != s2 {
+	if s2 := w1w2Xw2FooBar.Error(); s1 != s2 {
```
Yuggupta27, line 67: Since two separate variables are used and the declaration of S2 is lengthy by itself, using one-liner here doesn't look so good IMO.

Yuggupta27 (changes requested):

internal/rbd/rbd_journal.go
Rakshith-R: the convention is to have reason for nolint right after nolint comment. https://golangci-lint.run/usage/false-positives/
and `Use //nolint instead of // nolint because machine-readable comments should have no space by Go convention.`

internal/util/errors_test.go
Rakshith-R: It was done to satisfy `ifshort` linter and I think its fine the way it is.

nixpanic: I agree with that too (maybe not her, but other places). But if that is the recommended coding style, I'll accept it. These changes don't make the lines extra long, and I'm sure I'll get used to reading the code this way.

nixpanic (approved):

Madhu-1 (approved):

Issue #2077, opened 2021-05-12
Thick PVC clone does not  make thick PVC


humblec:
# Describe the bug #
When the a thick provisioned PVC is cloned using a thick provisioning enabled storage class of parent PVC, the cloned volume is not thick provisioned. The clone PVC will reach Bound state without thick provisioning.

Also see #2071

nixpanic:
@humblec Do you mean that the newly cloned volume needs to be flattened?

Issue #578, opened 2019-08-29
Add the ability to use exclusive RBD locking to prevent inadvertent multi-node access of an RWO image


ShyamsundarR:
Although CO systems are meant to ensure an RWO volume is attached to a single node, under any (un)availability constraints. It seems prudent to have an additional layer of protection to ensure this is taken care of.

Quoting solutions presented by @dillaman here:
"The best solution is to ensure that k8s will never request the CSI to map a PV to a node when it could already be mapped on an unresponsive node. Barring that, the CSI should enable the exclusive-lock feature on RBD images for RWO volumes at provision time. At attach time, it should break any existing locks prior to mapping via "rbd lock break" (which will blacklist the previous lock owner from the storage cluster), and the map operation should use "rbd device map --exclusive" to ensure that it immediately acquires the lock and will refuse to release it until unmapped / blacklisted."

We would need to analyze what this entails in a CO environment, and what workflows may be needed to remove nodes from a black list (if at all needed).

dillaman:
I want to retract the possibility of using "rbd device map --exclusive". The issue being that when that option is used, krbd will never release the lock. Normally that is exactly what you'd want, except for the case where you are trying to perform some maintenance operation on the image (snapshot create/remove, flatten, ...).

Madhu-1:
we need to do  below steps
During create
* add `--exclusive-lock` feature while creating an image if its RWO
During node stage
* use `rbd map --exclusive`
During unstage
* list the lock on image
* remove the lock
* unmap the rbd device

@dillaman @ShyamsundarR correct me  if am wrong

@dillaman to remove the lock we need  to connect the cluster right, means we need mon endpoints, admin ID and credentials ( is it possible to remove the lock  without that)

dillaman:
> we need to do below steps
> During create
> 
>     * add `--exclusive-lock` feature while creating an image if its RWO
>       During node stage

It's `--image-feature layering,exclusive-lock,......` (there is no `--exclusive-lock` optional)

>     * use `rbd map --exclusive`

Negative -- that was my retraction above. If you use this, you will break the ability to have third-party tools (i.e. `rbd snap create`) function against in-use images.

>       During unstage
> 
>     * list the lock on image
> 
>     * remove the lock
> 
>     * unmap the rbd device

[...]

PR #2101, opened 2021-05-26
Continue thick-provisioning of RBD images on CreateVolume restart

nixpanic:

In case CreateVolume gets interrupted (by provisioner pod restart, or similar) the slow thick-provisioning is aborted. The RBD image will not have the `thick-provisioned` metadata set. Restarting CreateVolume will detect the RBD image, and return the volume which is not completely thick-provisioned.

By validating that the CreateVolume request intends to thick-provision the volume, it is possible to detect the need for the `thick-provisioned` metadata in the RBD image. When the metadata is missing, start allocating the image again.

If the RBD image has watchers (`isInUse()` returns true), assume that the allocating of the image is still in progress. Because thick-provisioning is slow, Kubernetes tends to time-out the initial CreateVolume request(s). It is expected that the `not going to restart thick-provisioning of in-use image` is reported when Kubernetes restarts the CreateVolume request.

## Related issues ##

Fixes: https://bugzilla.redhat.com/1961647

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -19,7 +19,6 @@ package rbd
 import (
 	"context"
 	"errors"
-	"fmt"
 	"strconv"
 
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
@@ -118,13 +117,7 @@ func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	tp := "thickProvision"
-	thick := req.GetParameters()[tp]
-	if thick != "" {
-		if rbdVol.ThickProvision, err = strconv.ParseBool(thick); err != nil {
-			return nil, fmt.Errorf("failed to parse %q: %w", tp, err)
-		}
-	}
+	rbdVol.ThickProvision = cs.isThickProvisionRequest(req)
 
 	rbdVol.RequestName = req.GetName()
 
@@ -322,7 +315,23 @@ func flattenParentImage(ctx context.Context, rbdVol *rbdVolume, cr *util.Credent
 // when the process of creating a volume was interrupted.
 func (cs *ControllerServer) repairExistingVolume(ctx context.Context, req *csi.CreateVolumeRequest,
 	cr *util.Credentials, rbdVol *rbdVolume, rbdSnap *rbdSnapshot) (*csi.CreateVolumeResponse, error) {
-	if rbdSnap != nil {
+	vcs := req.GetVolumeContentSource()
+
+	switch {
+	// normal CreateVolume without VolumeContentSource
+	case vcs == nil:
+		// continue/restart allocating the volume in case it
+		// should be thick-provisioned
+		if cs.isThickProvisionRequest(req) {
+			err := rbdVol.RepairThickProvision()
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+		}
+
+	// rbdVol is a restore from snapshot, rbdSnap is passed
+	case vcs.GetSnapshot() != nil:
+		// restore from snapshot imploes rbdSnap != nil
 		// check if image depth is reached limit and requires flatten
 		err := checkFlatten(ctx, rbdVol, cr)
 		if err != nil {
@@ -333,6 +342,10 @@ func (cs *ControllerServer) repairExistingVolume(ctx context.Context, req *csi.C
 		if err != nil {
 			return nil, err
 		}
+
+	// rbdVol is a clone from an other volume
+	case vcs.GetVolume() != nil:
+		// TODO: is there really nothing to repair on image clone?
 	}
 
 	return buildCreateVolumeResponse(req, rbdVol), nil
@@ -1128,4 +1141,22 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		CapacityBytes:         rbdVol.VolSize,
 		NodeExpansionRequired: nodeExpansion,
 	}, nil
+}
+
+// isThickProvisionRequest returns true in case the request contains the
+// `thickProvision` option set to `true`.
+func (cs *ControllerServer) isThickProvisionRequest(req *csi.CreateVolumeRequest) bool {
+	tp := "thickProvision"
+
+	thick, ok := req.GetParameters()[tp]
+	if !ok || thick == "" {
+		return false
+	}
+
+	thickBool, err := strconv.ParseBool(thick)
+	if err != nil {
+		return false
+	}
+
+	return thickBool
 }
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -257,23 +257,8 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	found, err := rbdVol.Exists(ctx, parentVol)
 	if err != nil {
 		return nil, getGRPCErrorForCreateVolume(err)
-	}
-
-	if found {
-		if rbdSnap != nil {
-			// check if image depth is reached limit and requires flatten
-			err = checkFlatten(ctx, rbdVol, cr)
-			if err != nil {
-				return nil, err
-			}
-
-			err = rbdSnap.repairEncryptionConfig(&rbdVol.rbdImage)
-			if err != nil {
-				return nil, err
-			}
-		}
-
-		return buildCreateVolumeResponse(req, rbdVol), nil
+	} else if found {
+		return cs.repairExistingVolume(ctx, req, cr, rbdVol, rbdSnap)
 	}
 
 	err = validateRequestedVolumeSize(rbdVol, parentVol, rbdSnap, cr)
@@ -330,6 +315,27 @@ func flattenParentImage(ctx context.Context, rbdVol *rbdVolume, cr *util.Credent
 		}
 	}
 	return nil
+}
+
+// repairExistingVolume checks the existing volume or snapshot and makes sure
+// that the state is corrected to what was requested. It is needed to call this
+// when the process of creating a volume was interrupted.
+func (cs *ControllerServer) repairExistingVolume(ctx context.Context, req *csi.CreateVolumeRequest,
+	cr *util.Credentials, rbdVol *rbdVolume, rbdSnap *rbdSnapshot) (*csi.CreateVolumeResponse, error) {
+	if rbdSnap != nil {
+		// check if image depth is reached limit and requires flatten
+		err := checkFlatten(ctx, rbdVol, cr)
+		if err != nil {
+			return nil, err
+		}
+
+		err = rbdSnap.repairEncryptionConfig(&rbdVol.rbdImage)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return buildCreateVolumeResponse(req, rbdVol), nil
 }
 
 // check snapshots on the rbd image, as we have limit from krbd that an image
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -1441,6 +1441,43 @@ func (rv *rbdVolume) isThickProvisioned() (bool, error) {
 	return thick, nil
 }
 
+// RepairThickProvision writes zero bytes to the volume so that it will be
+// completely allocated. In case the volume is already marked as
+// thick-provisioned, nothing will be done.
+func (rv *rbdVolume) RepairThickProvision() error {
+	// if the image has the thick-provisioned metadata, it has been fully
+	// allocated
+	done, err := rv.isThickProvisioned()
+	if err != nil {
+		return fmt.Errorf("failed to repair thick-provisioning of %q: %w", rv, err)
+	} else if done {
+		return nil
+	}
+
+	// in case there are watchers, assume allocating is still happening in
+	// the background (by an other process?)
+	background, err := rv.isInUse()
+	if err != nil {
+		return fmt.Errorf("failed to get users of %q: %w", rv, err)
+	} else if background {
+		return fmt.Errorf("not going to restart thick-provisioning of in-use image %q", rv)
+	}
+
+	// TODO: can this be improved by starting at the offset where
+	// allocating was aborted/restarted?
+	err = rv.allocate(0)
+	if err != nil {
+		return fmt.Errorf("failed to continue thick-provisioning of %q: %w", rv, err)
+	}
+
+	err = rv.setThickProvisioned()
+	if err != nil {
+		return fmt.Errorf("failed to continue thick-provisioning of %q: %w", rv, err)
+	}
+
+	return nil
+}
+
 func (rv *rbdVolume) listSnapshots() ([]librbd.SnapInfo, error) {
 	image, err := rv.open()
 	if err != nil {
--- a/internal/rbd/controllerserver_test.go
+++ b/internal/rbd/controllerserver_test.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2021 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rbd
+
+import (
+	"testing"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+)
+
+func TestIsThickProvisionRequest(t *testing.T) {
+	cs := &ControllerServer{}
+	req := &csi.CreateVolumeRequest{
+		Name: "fake",
+		Parameters: map[string]string{
+			"unkownOption": "not-set",
+		},
+	}
+
+	// pass disabled/invalid values for "thickProvision" option
+	if cs.isThickProvisionRequest(req) {
+		t.Error("request is not for thick-provisioning")
+	}
+
+	req.Parameters["thickProvision"] = ""
+	if cs.isThickProvisionRequest(req) {
+		t.Errorf("request is not for thick-provisioning: %s", req.Parameters["thickProvision"])
+	}
+
+	req.Parameters["thickProvision"] = "false"
+	if cs.isThickProvisionRequest(req) {
+		t.Errorf("request is not for thick-provisioning: %s", req.Parameters["thickProvision"])
+	}
+
+	req.Parameters["thickProvision"] = "off"
+	if cs.isThickProvisionRequest(req) {
+		t.Errorf("request is not for thick-provisioning: %s", req.Parameters["thickProvision"])
+	}
+
+	req.Parameters["thickProvision"] = "no"
+	if cs.isThickProvisionRequest(req) {
+		t.Errorf("request is not for thick-provisioning: %s", req.Parameters["thickProvision"])
+	}
+
+	req.Parameters["thickProvision"] = "**true**"
+	if cs.isThickProvisionRequest(req) {
+		t.Errorf("request is not for thick-provisioning: %s", req.Parameters["thickProvision"])
+	}
+
+	// only "true" should enable thick provisioning
+	req.Parameters["thickProvision"] = "true"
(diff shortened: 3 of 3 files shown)


internal/rbd/controllerserver.go
```diff
@@ -332,6 +310,36 @@ func flattenParentImage(ctx context.Context, rbdVol *rbdVolume, cr *util.Credent
...
+		}
+
+		err = rbdSnap.repairEncryptionConfig(&rbdVol.rbdImage)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// continue/restart allocating the volume in case it
+		// should be thick-provisioned
+		if cs.isThickProvisionRequest(req) {
```
Madhu-1: can we have a check for repairing only fresh volume, not a PVC clone?

internal/rbd/rbd_util.go
```diff
@@ -1441,6 +1441,43 @@ func (rv *rbdVolume) isThickProvisioned() (bool, error) {
 	return thick, nil
 }
 
+// RepairThickProvision writes zero bytes to the volume so that it will be
+// completely allocated. In case the volume is already marked as
+// thick-provisioned, nothing will be done.
+func (rv *rbdVolume) RepairThickProvision() error {
```
Madhu-1, line 1447: If this rewrites all the contents (with zeros) in the existing volume maybe can add a note here to use this carefully?

internal/rbd/rbd_util.go
```diff
@@ -1441,6 +1441,43 @@ func (rv *rbdVolume) isThickProvisioned() (bool, error) {
...
+	} else if done {
+		return nil
+	}
+
+	// in case there are watchers, assume allocating is still happening in
+	// the background (by an other process?)
+	background, err := rv.isInUse()
+	if err != nil {
+		return fmt.Errorf("failed to get users of %q: %w", rv, err)
+	} else if background {
```
Madhu-1, line 1462: do we need to return the ABORT error message in this case? and also it's tricky we don't know who is holding the watcher on this image.

Madhu-1 (commented):
I know it's tricky do you have any plan to add E2E for this case?

internal/rbd/controllerserver.go
nixpanic: I do not think that is needed? Creating from a snapshot will have `rbdSnap` above set, and a clone will already have the thick-provisioned metadata in the image.

We could check if the VolumeContentSource is empty, but would like to understand what corner-case I'm missing.

internal/rbd/rbd_util.go
nixpanic: Is `RepairThickProvision writes zero bytes to the volume` not clear enough?

internal/rbd/rbd_util.go
nixpanic: I thought about that, and had the abort earlier. But I think it needs to be retried, as watchers may expired (provisioner lost its lease and an other took over?).

internal/rbd/controllerserver.go
Madhu-1: Can you please check it. the image metadata will get copied to the cloned images? just now tested it I didn't see image metadata getting copied to clone, Am I missing something here?

```
[root@dhcp53-170 rbd]# kubectl get pv pvc-9533729e-6947-4612-8fb7-58a8b01985fa -oyaml |grep -i imageName
      imageName: csi-vol-57e4646e-be17-11eb-9d80-0242ac110005
[root@dhcp53-170 rbd]# 
[root@dhcp53-170 rbd]# kubectl get pv pvc-efd86ad4-3fb9-40bb-b1f0-91b8cc4974d1 -oyaml |grep -i imageName
      imageName: csi-vol-665bd1ae-be17-11eb-9d80-0242ac110005


sh-4.4# rbd image-meta ls replicapool/csi-vol-665bd1ae-be17-11eb-9d80-0242ac110005    
There are 0 metadata on this image.
sh-4.4# rbd image-meta ls replicapool/csi-vol-57e4646e-be17-11eb-9d80-0242ac110005
There is 1 metadatum on this image:

Key                                  Value
.rbd.csi.ceph.com/thick-provisioned  true 
sh-4.4# rbd image-meta ls replicapool/csi-snap-60cc6470-be17-11eb-9d80-0242ac110005
There are 0 metadata on this image.
sh-4.4# 
```

Currently, I don't think of a corner case but I thought it's good to have an empty Content check.

internal/rbd/rbd_util.go
Madhu-1: I thought of having a warning for using this function or clone. As it overwrites the content. but if you think its fine. we can leave it

internal/rbd/rbd_util.go
Madhu-1: The watcher has some timeout around 5 minutes (i don't recall the exact time). Will this delays the PVC creation by + watcher time and also next if the contents are not zeroed the whole process gets added as a overhead?

internal/rbd/controllerserver.go
nixpanic: hmm, I though this was well tested, also in combination with encryption... will try it out again

nixpanic: Indeed, I probably misremember things. A clone (or even deep-copy) does not take the metadata from the source to the destination.

That makes this approach incomplete....

nixpanic: I guess improvements to the cloning/snapshotting will be handled through #2077. This PR only addresses the restarting of CreateVolume when a thick-provisioning is in progress.

internal/rbd/rbd_util.go
nixpanic: It all depends how the restarting of the provisioner is done, I think. On a normal restart (like upgrade), the Pod gets terminated and librbd will (force) close the connection. I do not expect any real delays because of that. In case the Pod/node becomes unreachable by Kubernetes, the allocating of the image can possibly continue on the non-leading provisioner. Retries on the (new) leading provisioner will assume that provisioning is continuing, which may be correct. If the non-leading provisioner is not doing anything anymore and the librbd connection was not closed, eventually there will be a timeout and re-provisioning will be done.

nixpanic:
> I know it's tricky do you have any plan to add E2E for this case?

I am not sure how to make this a reliable test, ideas welcome.

internal/rbd/rbd_util.go
nixpanic: in my testing, by deleting the leading provisioner pod, the timeout before the watcher is removed is less than 5 minutes. Provisioning is still slow (writing many GB of data), but it continues once there are no more watchers on the image.

internal/rbd/controllerserver.go
nixpanic: @Madhu-1: now only repairing the thick-provisioning for a new (non clone/restore) volume,

internal/rbd/rbd_util.go
nixpanic: For an idea of the timelines, these are logs of the new leading provisioner, starting at the 1st (restarted) CreateVolume request:
```
I0527 08:36:47.637982       1 utils.go:163] ID: 44 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 GRPC call: /csi.v1.Controller/CreateVolume
....
I0527 08:36:47.717254       1 rbd_journal.go:330] ID: 44 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 found existing volume (0001-0011-openshift-storage-0000000000000001-9e38478c-bec6-11eb-933f-0a580a80023c) with image name (csi-vol-9e38478c-bec6-11eb-933f-0a580a80023c) for request (pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7)
E0527 08:36:47.825434       1 utils.go:172] ID: 44 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 GRPC error: rpc error: code = Internal desc = not going to restart thick-provisioning of in-use image "ocs-storagecluster-cephblockpool/csi-vol-9e38478c-bec6-11eb-933f-0a580a80023c"
...
I0527 08:37:03.891402       1 utils.go:163] ID: 49 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 GRPC call: /csi.v1.Controller/CreateVolume
...
I0527 08:37:03.965553       1 rbd_journal.go:330] ID: 49 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 found existing volume (0001-0011-openshift-storage-0000000000000001-9e38478c-bec6-11eb-933f-0a580a80023c) with image name (csi-vol-9e38478c-bec6-11eb-933f-0a580a80023c) for request (pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7)
...
... <gRPC ID 49 does not return yet, allocating continues>
...
I0527 08:39:49.898106       1 utils.go:163] ID: 53 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 GRPC call: /csi.v1.Controller/CreateVolume
...
E0527 08:39:49.898473       1 controllerserver.go:234] ID: 53 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 an operation with the given Volume ID pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 already exists
E0527 08:39:49.898505       1 utils.go:172] ID: 53 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 GRPC error: rpc error: code = Aborted desc = an operation with the given Volume ID pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 already exists
I0527 08:40:21.907534       1 utils.go:163] ID: 54 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 GRPC call: /csi.v1.Controller/CreateVolume
...
E0527 08:40:21.908003       1 controllerserver.go:234] ID: 54 Req-ID: pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 an operation with the given Volume ID pvc-af7278e3-b78d-4ff2-b0a3-028358b465b7 already exists
...
```

All the above messages are expected, allocating the image is still in progress.

internal/rbd/rbd_util.go
```diff
@@ -1441,6 +1441,43 @@ func (rv *rbdVolume) isThickProvisioned() (bool, error) {
...
+	// if the image has the thick-provisioned metadata, it has been fully
+	// allocated
+	done, err := rv.isThickProvisioned()
+	if err != nil {
+		return fmt.Errorf("failed to repair thick-provisioning of %q: %w", rv, err)
+	} else if done {
+		return nil
+	}
+
+	// in case there are watchers, assume allocating is still happening in
```
humblec, line 1457: nit: s/allocating/allocation
btw, If there are no watchers on the image, how the allocation can still happen @nixpanic ?  @idryomov do we expect such conditions in any stage ?

internal/rbd/rbd_util.go
humblec, line 1447: afaict, this function does the complete process again. I mean the writing of zero bytes from start, please correct me if I am wrong here @nixpanic  . If the CO was aborted due to timeout in first or previous cycles, this is going to be mostly the same result in subsequent calls too. In short, we land into `looping` of the requests.  Rather I was thinking is there a way to know the progress of previous attempt and may be proceed from there? if yes, it could help us to speed up the operation and also not landing into the same situation again and again @nixpanic

internal/rbd/controllerserver.go
```diff
@@ -332,6 +310,47 @@ func flattenParentImage(ctx context.Context, rbdVol *rbdVolume, cr *util.Credent
...
+func (cs *ControllerServer) repairExistingVolume(ctx context.Context, req *csi.CreateVolumeRequest,
+	cr *util.Credentials, rbdVol *rbdVolume, rbdSnap *rbdSnapshot) (*csi.CreateVolumeResponse, error) {
+	vcs := req.GetVolumeContentSource()
+
+	switch {
+	// normal CreateVolume without VolumeContentSource
+	case vcs == nil:
+		// continue/restart allocating the volume in case it
+		// should be thick-provisioned
+		if cs.isThickProvisionRequest(req) {
```
Madhu-1, line 325: Just a corner case. If the allocation is completed but failed to set metadata due to restart. we are going to do the allocation again. can we check the used size of the image is equal and if yes just set the metadata or if no continue the allocation?

Madhu-1 (approved):
LGTM, just a suggestion left for possible improvement

internal/rbd/rbd_util.go
nixpanic: _allocating_ - the action is in progress?

The leader provisioner Pod could have lost its lease for some reason, and the previously leading Pod may still be working on it.

internal/rbd/rbd_util.go
nixpanic: Yes, if the process keeps on getting interrupted, there will be little progress made. We might be able to improve things by continuing where the previously interrupted allocating was aborted. There are ways to track progress, and it is possible to detect what parts of an image are allocated. However, that is not trivial and can be done as an enhancement later.

Remember that restarting the allocating is only done when the provisioner was restarted or an other fatal error occurred. Optimizing behaviour for error cases is not a high priority.

internal/rbd/controllerserver.go
nixpanic: Yes, there is a `TODO` noted in the `RepairThickProvision()` function. It would be an optimization for the case where a recurring error happened, so I do not deem this important to address.

idryomov (commented):
> In case the Pod/node becomes unreachable by Kubernetes, the allocating of the image can possibly continue on the non-leading provisioner. Retries on the (new) leading provisioner will assume that provisioning is continuing, which may be correct. If the non-leading provisioner is not doing anything anymore and the librbd connection was not closed, eventually there will be a timeout and re-provisioning will be done.

I'm a little scared by the mention of non-leading provisioner(s) and the general reliance on watches for checking whether the image is in use.  I have raised this in some fencing-related issues and PRs before, but once again: while a presence of a watch can be used to tell whether the image is in use (best effort check), the reverse is not true.  A lack of a watch is not a reliable indicator of pretty much anything -- the image may still be opened and written to!

In this case, what would prevent the non-leading provisioner that is not reachable by k8s from waking up and continuing its zeroing after the image has been (re)zeroed by the now leading provisioner and handed off to the user?  This seems like a potential data corruption scenario to me.

nixpanic:
@idryomov: when a watch is lost (initially added once the image was opened), can the opened image still be used for writing? I expected that the rbd-image-handle would become invalid, and writing returns a failure.

If there is a better recommendation than to check watches on RBD images, than we can surely implement it that way. Until there is a usable alternative, we probably will need to stick with the watchers for now.

internal/rbd/rbd_util.go
humblec: @nixpanic I think the point here is mainly on this: `how we (could land)/landed on more than one watcher` and if we land to a situation how would we identify what caused it, to make  sure or validate whether  is ongoing provisioning is  a genuine attempt happening or a left over from a stale request? If disturbances like leader reelection causes it ( I doubt that could leave such state) , If its from a stale/unmapped request for the CO, the operation may succeed in background , but we are going to retry the same in next request with new leader or whoever called, its going to be an overlap or unwanted second attempt. Suppose the first request caused the watcher to be there, but its no longer functioning , is this watcher get removed automatically after some time ? if not, we cant operate on this volume any more, that would also leave a stale volume in the backend as we are not doing any cleanup here.  I am just trying to understand the workflow in a bad situation , however I think its good to run through so that we are better prepared.

nixpanic: Yes, the watcher gets removed automatically. See the log snippet I added for https://github.com/ceph/ceph-csi/pull/2101#discussion_r640422026 where the 2nd provisioner restarts the CreateVolume request and re-allocates the RBD image. Initially the watch is still present and allocating aborts, but the retries for CreateVolume continue shortly after.

idryomov:
> @idryomov: when a watch is lost (initially added once the image was opened), can the opened image still be used for writing? I expected that the rbd-image-handle would become invalid, and writing returns a failure.

Yes.  A lost watch would be reestablished in due course.  Also, it may not know that the watch is lost for a while, etc.

> 
> If there is a better recommendation than to check watches on RBD images, than we can surely implement it that way. Until there is a usable alternative, we probably will need to stick with the watchers for now.

Ultimately it has to be something that would fence other potential writers at the Ceph level (`ceph osd blocklist`) before taking over.

It seems to be another instance of https://github.com/ceph/ceph-csi/issues/578 and similar issues.  Whereas before the only concern was the entity that mapped the image (since it was the only potential writer in the system), now that the provisioner can also write to the image it needs to be treated as a potential writer and be subjected to the same safeguards.

internal/rbd/rbd_util.go
idryomov: An inactive watch should be cleaned up automatically after 30 seconds.

internal/rbd/rbd_util.go
humblec: @nixpanic do we need to keep a tracker issue for this ?

humblec:
> Ultimately it has to be something that would fence other potential writers at the Ceph level (`ceph osd blocklist`) before taking over.
> 
> It seems to be another instance of #578 and similar issues. Whereas before the only concern was the entity that mapped the image (since it was the only potential writer in the system), now that the provisioner can also write to the image it needs to be treated as a potential writer and be subjected to the same safeguards.

Sure, ultimately `fencing` could help us here. Till we have such mechanism, I think we have to survive with `watchers` . Considering this PR improves the situation of thick provisioning to a good extent, it looks like its good to get this in and implement more safe guards accordingly in near future.  
LGTM. 

@idryomov do you have final comments on this PR ? if not, we will go ahead and merge this :+1: . Thanks for reviewing!

idryomov:
Nothing substantial.

Honestly, the entire scheme with k8s timing out create requests one after another and firing retries at random provisioner instances while zeroing is in progress seems very clumsy to me.  I just hope that thick provisioning isn't going to be widely used because otherwise the current implementation will bite pretty soon.

humblec:
> Nothing substantial.
> 
> Honestly, the entire scheme with k8s timing out create requests one after another and firing retries at random provisioner instances while zeroing is in progress seems very clumsy to me. 
> 
I was wondering if we could resume from where previous attempt halted ( if we can get progression complete : https://github.com/ceph/ceph-csi/pull/2101#discussion_r640450199)  it could surface this problem to an extent. @idryomov is it something we can get supporte/improve  from Ceph side ? especially for operations like `--thick provision`..etc?   

>I just hope that thick provisioning isn't going to be widely used because otherwise the current implementation will bite pretty soon.

As per current reports from the field, It looks like this is going to be used in a heavy extent. This is becoming default choice for VM usecases.

idryomov:
> I was wondering if we could resume from where previous attempt halted ( if we can get progression complete : [#2101 (comment)](https://github.com/ceph/ceph-csi/pull/2101#discussion_r640450199)) it could surface this problem to an extent. @idryomov is it something we can get supporte/improve from Ceph side ? especially for operations like `--thick provision`..etc?

Theoretically you could infer where zeroing got interrupted by looking at the provisioned size (i.e. how many `rbd_data.<image id>` objects are there) or store a marker alongside your `thick-provisioned` metadata after every X gigabytes are zeroed.  But this is just as a potential optimization, not a solution for the data corruption scenario.
> 
> > I just hope that thick provisioning isn't going to be widely used because otherwise the current implementation will bite pretty soon.
> 
> As per current reports from the field, It looks like this is going to be used in a heavy extent. This is becoming default choice for VM usecases.

I wonder why?  Do you have the details or some links I could peruse?

humblec:
> > > I just hope that thick provisioning isn't going to be widely used because otherwise the current implementation will bite pretty soon.
> > 
> > 
> > As per current reports from the field, It looks like this is going to be used in a heavy extent. This is becoming default choice for VM usecases.
> 
> I wonder why? Do you have the details or some links I could peruse?

@oritwas @mykaul @obnoxxx if possible, can you please share the details of thick provision use cases?

nixpanic:
> As per current reports from the field, It looks like this is going to be used in a heavy extent. This is becoming default choice for VM usecases.

My impression is that this is mainly provided for users that need more accurate accounting for storage consumption. When an environment is sufficiently planned for the expected workloads, thin-provisioning would surely be the recommendation. It might be difficult to estimate actual storage consumption for VM workloads, but any deployment should make sure that sufficient capacity is available, thick-provisioned or thin.

Thick-provisioning can be used as a safety-net when an environment does not have sufficient storage capacity for the planned workload. In that case thick-provisioning can prevent workloads from starting and having a storage cluster running out of space (where recovering becomes very difficult).

humblec:
As discussed in today's call, lets go ahead with this interim optimization now and catch up the improvements  on the way..

humblec (approved):

PR #2105, opened 2021-05-28
rbd: add support to create clone in different pools

Yuggupta27:
Add support to restore a pvc into a different pool.

Assuming there are 2 pools namely "replicapool" and "testpool"
Workflow:
1. Create a PVC in a replicapool.
2. For cloning, a snapshot gets created in replicapool.
3. PVC restore operation can be done in a new pool named testpool.

Co-authored-by: Madhu Rajanna <madhupr007@gmail.com>

Updates: #1838

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -142,8 +142,8 @@ func createORDeleteRbdResouces(action string) {
 	}
 }
 
-func validateRBDImageCount(f *framework.Framework, count int) {
-	imageList, err := listRBDImages(f)
+func validateRBDImageCount(f *framework.Framework, count int, pool string) {
+	imageList, err := listRBDImages(f, pool)
 	if err != nil {
 		e2elog.Failf("failed to list rbd images with error %v", err)
 	}
@@ -304,7 +304,7 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to validate owner of pvc with error %v", err)
 				}
 				// validate created backend rbd images
-				validateRBDImageCount(f, 0)
+				validateRBDImageCount(f, 0, defaultRBDPool)
 			})
 
 			By("create a PVC and bind it to an app", func() {
@@ -313,7 +313,7 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to validate pvc and application binding with error %v", err)
 				}
 				// validate created backend rbd images
-				validateRBDImageCount(f, 0)
+				validateRBDImageCount(f, 0, defaultRBDPool)
 			})
 
 			By("create a PVC and bind it to an app with normal user", func() {
@@ -322,7 +322,7 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to validate normal user pvc and application binding with error %v", err)
 				}
 				// validate created backend rbd images
-				validateRBDImageCount(f, 0)
+				validateRBDImageCount(f, 0, defaultRBDPool)
 			})
 
 			By("create a PVC and bind it to an app with ext4 as the FS ", func() {
@@ -339,7 +339,7 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to validate pvc and application binding with error %v", err)
 				}
 				// validate created backend rbd images
-				validateRBDImageCount(f, 0)
+				validateRBDImageCount(f, 0, defaultRBDPool)
 				err = deleteResource(rbdExamplePath + "storageclass.yaml")
 				if err != nil {
 					e2elog.Failf("failed to delete storageclass with error %v", err)
@@ -364,7 +364,7 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to validate pvc and application binding with error %v", err)
 				}
 				// validate created backend rbd images
-				validateRBDImageCount(f, 0)
+				validateRBDImageCount(f, 0, defaultRBDPool)
 				err = deleteResource(rbdExamplePath + "storageclass.yaml")
 				if err != nil {
 					e2elog.Failf("failed to delete storageclass with error %v", err)
@@ -408,7 +408,7 @@ var _ = Describe("RBD", func() {
 				}
 
 				// validate created backend rbd images
-				validateRBDImageCount(f, 1)
+				validateRBDImageCount(f, 1, defaultRBDPool)
 
 				selector, err := getDaemonSetLabelSelector(f, cephCSINamespace, rbdDaemonsetName)
 				if err != nil {
@@ -450,7 +450,7 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to delete PVC and application with error %v", err)
 				}
 				// validate created backend rbd images
-				validateRBDImageCount(f, 0)
+				validateRBDImageCount(f, 0, defaultRBDPool)
 				err = deleteResource(rbdExamplePath + "storageclass.yaml")
 				if err != nil {
 					e2elog.Failf("failed to delete storageclass with error %v", err)
@@ -497,7 +497,7 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to create PVC and application with error %v", err)
 				}
 				// validate created backend rbd images
-				validateRBDImageCount(f, 1)
+				validateRBDImageCount(f, 1, defaultRBDPool)
 
 				selector, err := getDaemonSetLabelSelector(f, cephCSINamespace, rbdDaemonsetName)
 				if err != nil {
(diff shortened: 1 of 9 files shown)


Yuggupta27:

rook deployment failed with
 ```
```

**Review by humblec — DISMISSED** [review]

File: internal/rbd/rbd_util.go [hunk 1]
```diff
@@ -1089,7 +1097,7 @@ func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *r
 	}
 
 	util.DebugLog(ctx, logMsg,
-		pSnapOpts, image, rv.imageFeatureSet.Names(), rv.Monitors)
+		pSnapOpts.String(), rv, rv.imageFeatureSet.Names(), rv.Monitors)
```
nixpanic: there should not be a need to call `.String()`, calls to this function should be automatic with the formatting `%s` or `%q`. Other calls to `.String()` are removed where the object is a pointer to rbdVolume or rbdSnapshot.

internal/rbd/rbd_util.go
```diff
@@ -1103,12 +1111,18 @@ func (rv *rbdVolume) cloneRbdImageFromSnapshot(ctx context.Context, pSnapOpts *r
...
 	}
 
+	// As the clone is yet to be created, open the Ioctx.
 	err = rv.openIoctx()
 	if err != nil {
 		return fmt.Errorf("failed to get IOContext: %w", err)
 	}
+	// Destroy only the ioctx, connection is needed for other operations.
+	defer func() {
+		defer rv.ioctx.Destroy()
```
nixpanic: no need to do this, it will automatically be done when `rv` is destroyed.

internal/rbd/rbd_util.go
```diff
@@ -1044,7 +1044,7 @@ func (rv *rbdVolume) hasSnapshotFeature() bool {
 }
 
 func (rv *rbdVolume) createSnapshot(ctx context.Context, pOpts *rbdSnapshot) error {
-	util.DebugLog(ctx, "rbd: snap create %s using mon %s", pOpts, pOpts.Monitors)
+	util.DebugLog(ctx, "rbd: snap create %s@%s using mon %s", rv, pOpts.String(), pOpts.Monitors)
```
nixpanic: `%s@%s` and passing `rv, pOpts.String()` **should** be the same as just `%s` and `pOpts`. If it is not, that would be a bug (or very large inconsistency) in the caller of `createSnapshot`.

Or, if this is really how a rbdSnapshot does not match with its actual rbd-pool representation, introduce `rbdSnapshot.getSpec()` that returns the `<parentPool>/<parentName>@snapname` format.

internal/rbd/rbd_util.go
```diff
@@ -1056,7 +1056,7 @@ func (rv *rbdVolume) createSnapshot(ctx context.Context, pOpts *rbdSnapshot) err
 }
 
 func (rv *rbdVolume) deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot) error {
-	util.DebugLog(ctx, "rbd: snap rm %s using mon %s", pOpts, pOpts.Monitors)
+	util.DebugLog(ctx, "rbd: snap rm %s@%s using mon %s", rv, pOpts.String(), pOpts.Monitors)
```
nixpanic: see comment for previous commit

internal/rbd/rbd_util.go
```diff
@@ -532,23 +532,26 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
 			return depth, nil
 		}
+		err := vol.openIoctx()
+		if err != nil {
+			return depth, err
+		}
+
 		err = vol.getImageInfo()
+		// FIXME: create and destroy the vol inside the loop.
+		// see https://github.com/ceph/ceph-csi/pull/1838#discussion_r598530807
```
nixpanic, line 552: This really is dangerous. Maybe file a cleanup PR to do this and rebase this PR on top of that.

Or, create an issue for this problem as this really needs correcting urgently.

internal/rbd/rbd_util.go
```diff
@@ -566,6 +566,7 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
 			depth++
 		}
 		vol.RbdImageName = vol.ParentName
+		vol.Pool = vol.ParentPool
```
nixpanic, line 569: Same issue as before, re-using a rbdVolume with different attributes is quite wrong. Needs cleanup urgently.

e2e/rbd_helper.go
```diff
@@ -232,6 +234,162 @@ func kmsIsVault(kms string) bool {
 	return kms == "vault"
 }
 
+// nolint:gocyclo // reduce complexity
```
nixpanic: Please use helper function from the start. Our e2e code should have the same quality as the other components.

e2e/rbd_helper.go
```diff
@@ -232,6 +234,162 @@ func kmsIsVault(kms string) bool {
...
+	for i := 0; i < totalCount; i++ {
+		go func(w *sync.WaitGroup, n int, s v1beta1.VolumeSnapshot) {
+			s.Name = fmt.Sprintf("%s%d", f.UniqueName, n)
+			wgErrs[n] = createSnapshot(&s, deployTimeout)
+			w.Done()
+		}(&wg, i, snap)
+	}
+	wg.Wait()
+
+	failed := 0
```
nixpanic: use something like the following to reduce the complexity:

```go
if failed := logErrors(wgErrs); failed != 0 {
    return fmt.Errorf("creating snapshots failed, %d errors were logged", failed)
}

File: e2e/rbd.go [hunk 9]
```diff
@@ -1212,6 +1212,50 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("create rbd clones in different pool", func() {
+				// snapshot beta is only supported from v1.17+
+				if k8sVersionGreaterEquals(f.ClientSet, 1, 17) {
```
nixpanic: Check the inverse, and use `e2e.Skip()` to continue with the next test. That makes it easier to drop checks like this for old/unsupported Kubernetes versions.

nixpanic (changes requested):

e2e/rbd_helper.go
Yuggupta27: Thanks for the suggestion! Added that commit!

e2e/rbd.go
Yuggupta27: Sounds good.

internal/rbd/rbd_util.go
Yuggupta27: Created an issue for the same https://github.com/ceph/ceph-csi/issues/2144 . Since this issue exists in the current implementation and not something that this PR introduces, IMO we don't need to block this PR for that and we can work on the cleanup after this PR is merged to avoid last-minute breakages and delay. What do you think @nixpanic ?

Yuggupta27:
CI failure due to #2146

internal/rbd/rbd_util.go
humblec: yeah, imo, it can be a followup PR.

internal/rbd/rbd_util.go
```diff
@@ -114,8 +114,10 @@ type rbdVolume struct {
 	Topology            map[string]string
 	// DataPool is where the data for images in `Pool` are stored, this is used as the `--data-pool`
 	// 	 argument when the pool is created, and is not used anywhere else
```
humblec, line 116: @Yuggupta27 not introduced with this patch, but please feel free to correct the `DataPool` explanation intendation.

e2e/rbd_helper.go
```diff
@@ -30,12 +32,15 @@ func rbdOptions(pool string) string {
...
 
-func createRBDStorageClass(c kubernetes.Interface, f *framework.Framework, scOptions, parameters map[string]string, policy v1.PersistentVolumeReclaimPolicy) error {
+func createRBDStorageClass(c kubernetes.Interface, f *framework.Framework, name string, scOptions, parameters map[string]string, policy v1.PersistentVolumeReclaimPolicy) error {
 	scPath := fmt.Sprintf("%s/%s", rbdExamplePath, "storageclass.yaml")
 	sc, err := getStorageClass(scPath)
 	if err != nil {
 		return nil
 	}
+	if name != "" {
+		sc.Name = name
```
humblec, line 42: @Yuggupta27 also keep `sc.Name` in proper spacing compared to outer condition check

internal/rbd/rbd_util.go
Yuggupta27: Sure, I'll send a seperate PR to fix the same

nixpanic (approved):

humblec (approved):

humblec:
@Yuggupta27 As discussed please get the comments addressed in other PR . will get this in for now! Thanks :+1:

Yuggupta27:

rook deployment failure

humblec:
@Mergifyio rebase

internal/rbd/rbd_util.go
Yuggupta27: Here https://github.com/ceph/ceph-csi/pull/2154

PR #2115, opened 2021-06-01
rbd: [DNM] set SetPoolFullTry to allow write fail

### LINKED ISSUE #2128: Failed to create over sized thick provisioning image
Currently For thick provisioning cephcsi creates an image and it writes zeros to the images until the image is full. If the ceph cluster gets full 9occupied more than 85%) during this operation the cephcsi client opened to write zeros hangs and all the subsequent requests for the same volume returns with an error when a user deletes the PVC this will leave a stale rbd image in the cluster.

The above problem can be fixed by #2115  by setting the SetPoolFullTry flag (which allows the delete of rbd images even if the cluster is full and also it returns an error when we are writing zeros when the ceph is full). Even with this fix, we end up Creating/Deleting the image multiple times until the ceph is freed up.  Once the ceph cluster is 85% the cluster becomes read-only and also affects the already mapping client read/write performance. With the help of #2125  we do check for the free space before creating the rbd image and we don't create/delete images frequent for a single PVC request. this will help to avoid the to an extent and will reduces the read/write overread of the already mapped rbd image.

Madhu-1:
when the ceph cluster is full or getting filled when thick provisioning is in progress the write won't fail it will be stuck, with this patch the write will fail when the ceph cluster gets full we can delete the image when write fails.

closes #2128

Note:- PR title and description need to be updated

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -293,7 +293,14 @@ func (ri *rbdImage) openIoctx() error {
 	ioctx.SetNamespace(ri.RadosNamespace)
 	ri.ioctx = ioctx
 
-	return nil
+	err = ri.ioctx.SetPoolFullTry()
+	if err != nil {
+		err = fmt.Errorf("failed to set  SetPoolFullTry IOContext for pool %s: %w", ri.Pool, err)
+	} else {
+		fmt.Println("successfully set the setpoolfulltry option")
+	}
+
+	return err
 }
 
 // getImageID queries rbd about the given image and stores its id, returns
--- a/vendor/github.com/ceph/go-ceph/rados/ioctx_octopus.go
+++ b/vendor/github.com/ceph/go-ceph/rados/ioctx_octopus.go
@@ -1,0 +1,35 @@
+// +build !nautilus
+
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <errno.h>
+// #include <stdlib.h>
+// #include <rados/librados.h>
+import "C"
+
+// SetPoolFullTry makes sure to send requests to the cluster despite
+// the cluster or pool being marked full; ops will either succeed(e.g., delete)
+// or return EDQUOT or ENOSPC.
+//
+// Implements:
+//  void rados_set_pool_full_try(rados_ioctx_t io);
+func (ioctx *IOContext) SetPoolFullTry() error {
+	if err := ioctx.validate(); err != nil {
+		return err
+	}
+	C.rados_set_pool_full_try(ioctx.ioctx)
+	return nil
+}
+
+// UnsetPoolFullTry unsets the flag set by SetPoolFullTry()
+//
+// Implements:
+//  void rados_unset_pool_full_try(rados_ioctx_t io);
+func (ioctx *IOContext) UnsetPoolFullTry() error {
+	if err := ioctx.validate(); err != nil {
+		return err
+	}
+	C.rados_unset_pool_full_try(ioctx.ioctx)
+	return nil
+}


Madhu-1:
@idryomov this was the change I was referring to thick provisioning.

idryomov:
Ah, so this was with go-ceph after all.

I don't see anything wrong at first sight, but I'm not really familiar with go-ceph.  Let me poke around...

Madhu-1:
> Ah, so this was with go-ceph after all.
> 
> I don't see anything wrong at first sight, but I'm not really familiar with go-ceph. Let me poke around...

Thanks @idryomov

leseb (commented):
I could be wrong but it looks like it's already enabled in Ceph https://github.com/ceph/ceph/blob/nautilus/src/tools/rbd/action/Remove.cc#L80, I wonder if it's really needed. @idryomov thoughts?

idryomov:
> I could be wrong but it looks like it's already enabled in Ceph https://github.com/ceph/ceph/blob/nautilus/src/tools/rbd/action/Remove.cc#L80, I wonder if it's really needed. @idryomov thoughts?

It is enabled for remove, not create.  And ceph-csi doesn't generally use the CLI tool except for mapping and unmapping and maybe a couple of other edge cases.

leseb:
> > I could be wrong but it looks like it's already enabled in Ceph https://github.com/ceph/ceph/blob/nautilus/src/tools/rbd/action/Remove.cc#L80, I wonder if it's really needed. @idryomov thoughts?
> 
> It is enabled for remove, not create. And ceph-csi doesn't generally use the CLI tool except for mapping and unmapping and maybe a couple of other edge cases.

That makes sense, I overlooked we were not using the CLI. Thanks!

Madhu-1:
```go
package main

import (
	"flag"
	"fmt"
	"strconv"

	"github.com/ceph/go-ceph/rados"
	"github.com/ceph/go-ceph/rbd"
	"k8s.io/cloud-provider/volume/helpers"
)

const (
	poolname   = "replicapool"
)

var (
	monitors  = flag.String("monitors", "192.168.39.62:6789", "comma seperated MON list")
	id        = flag.String("id", "admin", "ID of user to connect to the Ceph cluster (default `admin`)")
	key       = flag.String("key", "AQBDp71gMFh/IRAA8xH97KQmO/v4UmBxFcSE2A==", "key for user specified in `id`")
	operation = flag.String("operation", "", "type of operation to perform [create|delete]")
	count     = flag.Int("count", 1, "count of operations to perform (default 1)")
)

const (
	// GB - GigaByte size
	GB = 1000 * 1000 * 1000
	// GiB - GibiByte size
	GiB = 1024 * 1024 * 1024
)

// allocate uses the stripe-period of the image to fully allocate (thick
// provision) the image.
func allocate(offset uint64, ioctx *rados.IOContext, imageName string) error {
	image, err := rbd.OpenImage(ioctx, imageName, rbd.NoSnapshot)
	if err != nil {
		return err
	}
	defer image.Close()

	st, err := image.Stat()
	if err != nil {
		return err
	}

	sc, err := image.GetStripeCount()
	if err != nil {
		return err
	}

	// blockSize is the stripe-period: size of the object-size multiplied
	// by the stripe-count
	blockSize := sc * (1 << st.Order)
	zeroBlock := make([]byte, blockSize)

	// the actual size of the image as available in the pool, can be
	// marginally different from the requested image size
	size := st.Size - offset
    fmt.Println("the size of the image",size)
	fmt.Println("the blocksize ",blockSize)

        // In case the remaining space on the volume is smaller than blockSize,
	// write a partial block with WriteAt() after this loop.
	for size > blockSize {
		writeSize := size
		// write a maximum of 1GB per WriteSame() call
		if size > helpers.GiB {
			writeSize = helpers.GiB
		}

		// round down to the size of a zeroBlock
		if (writeSize % blockSize) != 0 {
			writeSize = (writeSize / blockSize) * blockSize
		}

		_, err = image.WriteSame(offset, writeSize, zeroBlock,
			rados.OpFlagNone)
		if err != nil {
			return fmt.Errorf("failed to allocate %d/%d bytes at "+
				"offset %d: %w", writeSize, blockSize, offset, err)
		}

		// write succeeded
		size -= writeSize
		offset += writeSize
	}

	// write the last remaining bytes, in case the image size can not be
	// written with the optimal blockSize
	if size != 0 {
		_, err = image.WriteAt(zeroBlock[:size], int64(offset))
		if err != nil {
			return fmt.Errorf("failed to allocate %d bytes at "+
				"offset %d: %w", size, offset, err)
		}
	}

	return nil
}

func create_conn() (conn *rados.Conn, err error) {
	conn, err = rados.NewConn()
	if err != nil {
		return nil, fmt.Errorf("Error creating NewConn (%v)\n", err)
	}

	baseArgs := []string{
		"-m", *monitors,
		"--id", *id,
		"--key=" + *key,
		"-c", "/etc/ceph/ceph.conf",
	}

	err = conn.ParseCmdLineArgs(baseArgs)
	if err != nil {
		return nil, fmt.Errorf("Error updating connection with args (%v) (%v)", baseArgs, err)
	}

	err = conn.Connect()
	if err != nil {
		return nil, fmt.Errorf("Error connecting to Ceph cluster (%v). Connection args (%+v)", err, baseArgs)
	}

	err = conn.SetConfigOption("rbd_discard_on_zeroed_write_same", "false")
	if err != nil {
		return nil, err
	}
	return
}

func create_rbd() error {
	conn, err := create_conn()
	if err != nil {
		return fmt.Errorf("Error creating connection (%v)", err)
	}
	defer conn.Shutdown()

	ioctx, err := conn.OpenIOContext(poolname)
	if err != nil {
		return fmt.Errorf("Error creating IO context (%v)", err)
	}
	defer ioctx.Destroy()

	err = ioctx.SetPoolFullTry()
	if err != nil {
		err = fmt.Errorf("failed to set  SetPoolFullTry IOContext for pool %s: %w", poolname, err)
	} else {
		fmt.Println("successfully set the setpoolfulltry option")
	}
	image, err := rbd.Create(ioctx, "test", 858993459200, 22)
	if err != nil {
		return fmt.Errorf("Error creating backing image (%v)", err)
	}
	defer image.Close()

	err = allocate(0, ioctx, "test")
	if err != nil {
		return fmt.Errorf("failed to allocate image (%v)", err)
	}
	return nil
}

func delete_rbd() error {
	conn, err := create_conn()
	if err != nil {
		return fmt.Errorf("Error creating connection (%v)", err)
	}
	defer conn.Shutdown()

	ioctx, err := conn.OpenIOContext(poolname)
	if err != nil {
		return fmt.Errorf("Error creating IO context (%v)", err)
	}
	defer ioctx.Destroy()

	err = ioctx.SetPoolFullTry()
	if err != nil {
		err = fmt.Errorf("failed to set  SetPoolFullTry IOContext for pool %s: %w", poolname, err)
	} else {
		fmt.Println("successfully set the setpoolfulltry option")
	}

	//rbd info "${poolname}"/"${omapprefix}"-vol-"${reqid}"
	image := rbd.GetImage(ioctx, "test")
	if image == nil {
		return fmt.Errorf("Error getting image handle (%v)", err)
	}

	err = image.Remove()
	if err != nil {
		return fmt.Errorf("Error removing image (%v)", err)
	}

	return nil
}

func init() {
	flag.Parse()
}

func createRun(count int) error {
	for i := 0; i < count; i++ {
		err := create_rbd()
		if err != nil {
			return fmt.Errorf("Error in create_rbd (%v) (loop %d)", err, i)
		}
	}

	return nil
}

func deleteRun(count int) error {
	for i := 0; i < count; i++ {
		err := delete_rbd()
		if err != nil {
			return fmt.Errorf("Error in delete_rbd (%v) (loop %d)", err, i)
		}
	}

	return nil
}

func main() {
	fmt.Println(*monitors, *id, *key, *operation)
	if *monitors == "" || *id == "" || *key == "" || *operation == "" {
		fmt.Printf("Missing one or more options in command invocation\n")
		flag.PrintDefaults()
		return
	}

	fmt.Printf("Operation: %s\n", *operation)

	switch *operation {
	case "create":
		createRun(*count)
	case "delete":
		deleteRun(*count)
	default:
		fmt.Printf("Operation (%s) not supported\n", *operation)
	}

	fmt.Printf("Operation: %s completed\n", *operation)
	return
}
```

Used the above go-ceph code and still, this is not working for me tried in ceph `v16.2.2` cluster @nixpanic @agarwal-mudit is there anything am missing from go-ceph? 

@idryomov did you get a chance to look into it?

idryomov:
Sorry, it appears to be broken (or at least misbehaving in some cases) on the librbd/RADOS side.  I'm yet to confirm but I'm pretty sure it isn't related to go-ceph.

I'll try to look into it this week.

Madhu-1:
> Sorry, it appears to be broken (or at least misbehaving in some cases) on the librbd/RADOS side. I'm yet to confirm but I'm pretty sure it isn't related to go-ceph.
> 
> I'll try to look into it this week.

Thanks for the update @idryomov

Madhu-1:
> Sorry, it appears to be broken (or at least misbehaving in some cases) on the librbd/RADOS side. I'm yet to confirm but I'm pretty sure it isn't related to go-ceph.
> 
> I'll try to look into it this week.

@idryomov https://github.com/ceph/ceph/pull/43207 is a fix for this one?

idryomov:
Yes, once it lands `SetPoolFullTry()` should do its job.

Madhu-1:
> Yes, once it lands `SetPoolFullTry()` should do its job.

Thanks for the confirmation @idryomov.

Madhu-1:
This is no longer required as we are deprecating thick provisioning.

PR #2125, opened 2021-06-02
rbd: check free size during thick volume creation

### LINKED ISSUE #2128: Failed to create over sized thick provisioning image

Madhu-1:
Added check to validate the request PVC size against 80% of the free size of a pool. This will help us to ensure that we don't try to create a volume that is greater than the available size to an extend.

Note:- when the ceph cluster is  85% full the cluster will become read-only. 80% is chosen as a random value to avoid overcommitting.

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -1172,6 +1172,18 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	// always round up the request size in bytes to the nearest MiB/GiB
 	volSize := util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
 
+	thick, err := rbdVol.isThickProvisioned()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to detect if image %q is thick-provisioned: %v", rbdVol, err)
+	}
+
+	// check are we over committing before resizing the volume.
+	if thick {
+		err = cs.validateThickOverProvisioning(rbdVol)
+		if err != nil {
+			return nil, err
+		}
+	}
 	// resize volume if required
 	if rbdVol.VolSize < volSize {
 		util.DebugLog(ctx, "rbd volume %s size is %v,resizing to %v", rbdVol, rbdVol.VolSize, volSize)
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -91,6 +91,43 @@ func (cs *ControllerServer) validateVolumeReq(ctx context.Context, req *csi.Crea
 	return nil
 }
 
+func getpercent(percent, size int64) int64 {
+	return ((size * percent) / 100)
+}
+
+func (cs *ControllerServer) validateThickOverProvisioning(vol *rbdVolume) error {
+	// LockName taken to validate the ceph pool size
+	poolSizeLock := "poolSizeLock"
+	if acquired := cs.VolumeLocks.TryAcquire(poolSizeLock); !acquired {
+		return status.Errorf(codes.Aborted, util.PoolOperationAlreadyExistsFmt, vol.Pool)
+	}
+	defer cs.VolumeLocks.Release(poolSizeLock)
+
+	imagePool := vol.Pool
+	if vol.DataPool != "" {
+		// use data pool if its set
+		imagePool = vol.DataPool
+	}
+
+	cStats, err := getCephStats(vol.Monitors, *vol.conn.Creds)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+
+	for _, p := range cStats.Pools {
+		if p.Name == imagePool {
+			// Checking 80% of the total available size in the pool so that we
+			// dont over provision in case of thick provisioning.
+			// nolint:gomnd // calculating 80 % of maximum available size in pool.
+			availAllocSize := getpercent(80, p.Stats.MaxAvail)
+			if vol.VolSize > availAllocSize {
+				return status.Errorf(codes.ResourceExhausted, "cannot create volume %s. Requested size %d is greater than available size %d", vol.RequestName, vol.VolSize, availAllocSize)
+			}
+		}
+	}
+	return nil
+}
+
 func (cs *ControllerServer) parseVolCreateRequest(ctx context.Context, req *csi.CreateVolumeRequest) (*rbdVolume, error) {
 	// TODO (sbezverk) Last check for not exceeding total storage capacity
 
@@ -185,7 +222,7 @@ func getGRPCErrorForCreateVolume(err error) error {
 
 // validateRequestedVolumeSize validates the request volume size with the
 // source snapshot or volume size, if there is a size mismatches it returns an error.
-func validateRequestedVolumeSize(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cr *util.Credentials) error {
+func (cs *ControllerServer) validateRequestedVolumeSize(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSnapshot, cr *util.Credentials) error {
 	if rbdSnap != nil {
 		vol := generateVolFromSnap(rbdSnap)
 		err := vol.Connect(cr)
@@ -207,6 +244,14 @@ func validateRequestedVolumeSize(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSnaps
 			return status.Errorf(codes.InvalidArgument, "size mismatches, requested volume size %d and source volume size %d", rbdVol.VolSize, parentVol.VolSize)
 		}
 	}
+
+	if rbdVol.ThickProvision {
+		err := cs.validateThickOverProvisioning(rbdVol)
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
@@ -254,7 +299,7 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return cs.repairExistingVolume(ctx, req, cr, rbdVol, rbdSnap)
 	}
 
-	err = validateRequestedVolumeSize(rbdVol, parentVol, rbdSnap, cr)
+	err = cs.validateRequestedVolumeSize(rbdVol, parentVol, rbdSnap, cr)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -1528,3 +1528,41 @@ func (rv *rbdVolume) getOrigSnapName(snapID uint64) (string, error) {
 
 	return origSnapName, nil
 }
+
+type cephStorageStats struct {
+	Pools []struct {
+		Name  string `json:"name"`
+		ID    int    `json:"id"`
+		Stats struct {
+			MaxAvail int64 `json:"max_avail"`
+		} `json:"stats"`
+	} `json:"pools"`
+}
+
+// FIXME replace this with go-ceph api
+// getCephStats fetches the storage size information.
+func getCephStats(monitor string, cred util.Credentials) (*cephStorageStats, error) {
+	// ceph df --format=json
+	var cephStats cephStorageStats
+
+	stdout, _, err := util.ExecCommand(
+		context.TODO(),
+		"ceph",
+		"-m", monitor,
+		"--id", cred.ID,
+		"--keyfile="+cred.KeyFile,
+		"-c", util.CephConfigPath,
+		"--format="+"json",
+		"df")
+	if err != nil {
+		return nil, err
+	}
+
+	if stdout != "" {
+		err = json.Unmarshal([]byte(stdout), &cephStats)
+		if err != nil {
+			return nil, fmt.Errorf("unmarshal failed (%w), raw buffer response: %s", err, stdout)
+		}
+	}
+	return &cephStats, nil
+}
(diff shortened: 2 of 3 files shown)


Madhu-1:
cc @idryomov @Madhu-1 @leseb

internal/rbd/rbd_util.go
```diff
@@ -1528,3 +1528,40 @@ func (rv *rbdVolume) getOrigSnapName(snapID uint64) (string, error) {
 
 	return origSnapName, nil
 }
+
+type cephStaorageStats struct {
```
agarwal-mudit: You mean cephStorageStats

Rakshith-R:
```suggestion
type cephStorageStats struct {
```

internal/rbd/rbd_util.go
```diff
@@ -1528,3 +1528,40 @@ func (rv *rbdVolume) getOrigSnapName(snapID uint64) (string, error) {
...
+		Stats struct {
+			MaxAvail int64 `json:"max_avail"`
+		} `json:"stats"`
+	} `json:"pools"`
+}
+
+// getCephStats fetches the storage size information.
+func getCephStats(monitor string, cred util.Credentials) (*cephStaorageStats, error) {
+	// ceph df --format=json
+	var cephStats cephStaorageStats
```
Rakshith-R:
```suggestion
	var cephStats cephStorageStats
```

agarwal-mudit:
Will this work for expansion also or only creation?

internal/rbd/rbd_util.go
Madhu-1: my bad. fixed this now.

Madhu-1:
> Will this work for expansion also or only creation?

This is only for creation. Let me see for expansion also.

Rakshith-R:
What if there's two for 5gb thick provisioned pvc requests at around the same time when cluster has maxavai size 10gb ?
I suppose we'll still hit the same issue.

internal/rbd/controllerserver.go
```diff
@@ -91,6 +91,36 @@ func (cs *ControllerServer) validateVolumeReq(ctx context.Context, req *csi.Crea
...
+		return status.Error(codes.Internal, err.Error())
+	}
+	for _, p := range cStats.Pools {
+		if p.Name == vol.Pool {
+			// Checking 80% of the total available size in the pool so that we
+			// dont over provision in case of thick provisioning.
+			// nolint:gomnd // calculating 80 % of maximum available size in pool.
+			availAllocSize := getpercent(80, p.Stats.MaxAvail)
+			if vol.VolSize > availAllocSize {
+				return status.Errorf(codes.ResourceExhausted, "cannot create volume %s. Requested size %d and Maximum available size %d", vol.RequestName, vol.VolSize, availAllocSize)
```
Yuggupta27:
```suggestion
				return status.Errorf(codes.ResourceExhausted, "cannot create volume %s. Requested size %d is greater than Maximum available size %d", vol.RequestName, vol.VolSize, availAllocSize)
```

Yuggupta27 (changes requested):

Madhu-1:
> What if there's two for 5gb thick provisioned pvc requests at around the same time when cluster has maxavai size 10gb ?
> I suppose we'll still hit the same issue.

This will not avoid all the corner cases, this will avoid few corner cases. but with https://github.com/ceph/ceph-csi/pull/2115 we should be able to fix it fully.

nixpanic (changes requested):
I do not think this is the right approach. It will always be racy, multiple volumes can be created at the same time, or volumes can get deleted while other requests are pending.

Hanging is not nice, but with the right alerting in place, a storage admin can trigger cluster expansion (or have other volumes deleted) so that the workload from users continues automatically.

Thick provisioning itself is a workaround for storage accounting, adding a workaround on top of a workaround.... well... we would need extremely good reasons to do so.

Madhu-1:
> I do not think this is the right approach. It will always be racy, multiple volumes can be created at the same time, or volumes can get deleted while other requests are pending.
> 
Yes, this won't solve the issue completely but makes things better.

> Hanging is not nice, but with the right alerting in place, a storage admin can trigger cluster expansion (or have other volumes deleted) so that the workload from users continues automatically.
> 

But with the current problem, either request will hang or we end up creating and deleting the image always until the ceph is freed up.  we thought it's better to fail early instead of doing create/delete which will also cause problem to the rbd images which are already mounted in the cluster.

> Thick provisioning itself is a workaround for storage accounting, adding a workaround on top of a workaround.... well... we would need extremely good reasons to do so.

internal/rbd/controllerserver.go
```diff
@@ -91,6 +91,36 @@ func (cs *ControllerServer) validateVolumeReq(ctx context.Context, req *csi.Crea
...
+		return status.Errorf(codes.Aborted, util.PoolOperationAlreadyExistsFmt, vol.Pool)
+	}
+	defer cs.VolumeLocks.Release(poolSizeLock)
+
+	cStats, err := getCephStats(vol.Monitors, *vol.conn.Creds)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+	for _, p := range cStats.Pools {
+		if p.Name == vol.Pool {
```
idryomov: Does this handle images with a separate data pool?

agarwal-mudit:
> I do not think this is the right approach. It will always be racy, multiple volumes can be created at the same time, or volumes can get deleted while other requests are pending.
> 
> Hanging is not nice, but with the right alerting in place, a storage admin can trigger cluster expansion (or have other volumes deleted) so that the workload from users continues automatically.
> 
> Thick provisioning itself is a workaround for storage accounting, adding a workaround on top of a workaround.... well... we would need extremely good reasons to do so.

I am fine with hanging but the problem here is that once we reach 85% cluster goes to read-only and even if we have alerts placed there is no way to avoid it as we have already issued the command to create a PVC and there is no way we can retract from there.

Moreover, I don't agree with the current design where you create a thin provisioned image and then start filling it. I don't think we do it like this in other storages. When someone requests to creat a thick provisioning volume they expect a guarantee for the space, they either get it or they don't get it. They don't just hang.

nixpanic:
I think we need a clear description (in an issue) about the cases that need to be addressed. Each case needs to be explained individually with a reliable way to prevent it from happening.

Does this address issues where #2115 is not sufficient?

Behaviour needs to be predictable. Users need to be able to understand why volumes can not be created, and what can be done about it. Storage cluster management is not part of Ceph-CSI, we would need to integrate with the cluster manager (Rook or other) to indicate why Ceph-CSI takes a decision that the storage cluster would take differently. Ceph and Kubernetes have a different view in management of storage than is introduced with checking a 80% watermark.

I think there are too few case where this proposed workaround has a benefit. When introducing a workaround, the advantaged need to outweigh the extra burden to keep the workaround in place. But please, document the problem and the proposed solution (with unaddressed corner-cases) in an issue for better discussion.

nixpanic:
> Moreover, I don't agree with the current design where you create a thin provisioned image and then start filling it. I don't think we do it like this in other storages. When someone requests to creat a thick provisioning volume they expect a guarantee for the space, they either get it or they don't get it. They don't just hang.

Hanging while creating a volume is definitely not nice. Allocating the volume should fail, and the partially allocated volume should get cleaned up. This is what #2115 is expected to address.

Madhu-1:
> I think we need a clear description (in an issue) about the cases that need to be addressed. Each case needs to be explained individually with a reliable way to prevent it from happening.
> 

https://github.com/ceph/ceph-csi/issues/2128 captures the issue with more details

> Does this address issues where #2115 is not sufficient?
>

AFAIK not sufficient as it can still cause overhead for already mounted client read/write performance.

> Behaviour needs to be predictable. Users need to be able to understand why volumes can not be created, and what can be done about it. Storage cluster management is not part of Ceph-CSI, we would need to integrate with the cluster manager (Rook or other) to indicate why Ceph-CSI takes a decision that the storage cluster would take differently. Ceph and Kubernetes have a different view in management of storage than is introduced with checking a 80% watermark.
> 

As the Failed error messages will propate the the PVC events i think the user will get fair idea why the request is not satisfied.

> I think there are too few case where this proposed workaround has a benefit. When introducing a workaround, the advantaged need to outweigh the extra burden to keep the workaround in place. But please, document the problem and the proposed solution (with unaddressed corner-cases) in an issue for better discussion.

internal/rbd/controllerserver.go
Madhu-1: Thanks for catching this one, added a check for the data pool also.

agarwal-mudit:
#2115 will still return a failure once there is no space but by the time we would have already messed up the cluster by making it read only as it would be filled >= 85%  So,  #2115 is there for corner cases only.

I agree that storage management is not csi's job and we need to see where this can be moved in future.

But, IMHO the only harm we are doing by implementing this is that we might deny PVC creation for some users even though the storage is available but that cost is worth if it avoids making the cluster unusable. Storage admins can retry the pvc creations in such cases given that we are not expecting extensive use of this feature.

nixpanic:
> #2115 will still return a failure once there is no space but by the time we would have already messed up the cluster by making it read only

Once the image allocation failed, the image should get deleted and space should be free'd up again. There is no permanent read-only scenario in this case?

Many environments will likely have a mix of thin-provisioned (CephFS, object-storage, ..) and thick-provisioned volumes. When the thin-provisioned volumes grow while a thick-provisioned volume is getting allocated, this problem can still occur.

I really doubt that imposing artificial limits for allocating thick-provisioned volumes through Ceph-CSI is sufficient to address the problem.

nixpanic:
> > Does this address issues where #2115 is not sufficient?
> 
> AFAIK not sufficient as it can still cause overhead for already mounted client read/write performance.

This is theoretical only. Once clients start to use their requested volumes, the storage cluster and network needs to handle the load too. I do not think provisioning is very different than the I/O an application does.

Of course writing data requires network bandwidth, CPU and disk I/O. There is a request to the Ceph team about providing an efficient allocation method for RBD images. Once that is available, this problem would be gone.

agarwal-mudit:
> Once the image allocation failed, the image should get deleted and space should be free'd up again. There is no permanent read-only scenario in this case?

I am not sure at what point image allocation will fail, will it return ENOSPC after the cluster hit 85% mark or before? If after then we will hit read-only state.


> Many environments will likely have a mix of thin-provisioned (CephFS, object-storage, ..) and thick-provisioned volumes. When the thin-provisioned volumes grow while a thick-provisioned volume is getting allocated, this problem can still occur.

Yes, for such scenarios we need #2115 
 
> I really doubt that imposing artificial limits for allocating thick-provisioned volumes through Ceph-CSI is sufficient to address the problem.

Agree, but it will solve some problems. Best would have been to change it from ceph side but that's not possible.

nixpanic:
> > I really doubt that imposing artificial limits for allocating thick-provisioned volumes through Ceph-CSI is sufficient to address the problem.
> 
> Agree, but it will solve some problems. Best would have been to change it from ceph side but that's not possible.

It also introduces problems. We will need to maintain this workaround for an error case. The workaround becomes as large as a feature including at least one configuration option, maybe more.

The case when multiple volumes get created at the same time, but are competing for allocation is still not handled. The current proposed solution can only be achieved if there is a single provisioning action happening at the time (which we do not want to enforce). We may need to track if there are thick-provisioning requests in progress, and how much space they have allocated, and still need to allocate, before deciding to allow an additional parallel volume creation.

Why would we want to get into the business of improving an error case (storage cluster becomes full), when we can only address a single cause which most likely happens rarely? I think it is difficult to reason about not fixing other cases, and requiring users to administrate their storage cluster (better/correctly) for those, when a good administrated storage cluster prevents running out of space in the first place. ([Storage Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/#storage-resource-quota) or monitoring available space in the Ceph cluster for example.)

humblec:
We are working here with the kube principles in place which is nothing but keep trying till we get the desired functionality. That's what kube or controller does
in this case wrt the volume creation regardless of thin provisioning or thick provisioning. So bailing out in between while it did not satisfy the request
in a few iterations is not really the (correct) way to go. It is also *not* a viable solution while we think about a dynamic setup where many requests are fired
in parallel...etc, some of them could be 'delete' requests which free up some space.  We can also think about a situation, where storage admin or cluster admin
add more space in between which could eventually satisfy the request. 

The important thing to note here is the entities involved in this transaction.The user who fires the PVC request does not have any visibility to the backend cluster or is not supposed to have it.
The Storage admin who has control over the backend cluster is supposed to 'present' available storage clusters properly to end users or consumers.
That said, while the storage admin opens up a storage class with 'thick provision' enabled, they should make sure it can be satisfied. Obviously they have the choice for presenting such a storage pool
for an end user based on any criteria ( 70%/80% threshold, namespace quota based..etc) and well possible with storage quota feature.  

If Ceph/Backend storage can handle 'stale/frozen' requests like this, it could be a good solution too. 

imo, the workaround or solution should be handled either from the Storage vendor or from storage admin who present the storage to the end user.

humblec:
@Madhu-1 @nixpanic arent we in consensus that we dont need this change?

Madhu-1:
As per discussions, this is not needed. closing this one

PR #2025, opened 2021-04-27
ci: update static check tools

yati1998:
This PR updates the static check tools to
the latest version.
Further needs to resolve all the errors after
updating the version.

Updates: #1586

Signed-off-by: Yati Padia <ypadia@redhat.com>

PR #2126, opened 2021-06-02
cleanup: Modifies Wrapcheck linter

yati1998:
Wrapcheck is a  simple Go linter to check that errors
from external packages are wrapped during return to
help identify the error source during debugging.
This commit addresses the wrapcheck error

Updates:#2025

### DIFF
--- a/e2e/cephfs_helper.go
+++ b/e2e/cephfs_helper.go
@@ -24,7 +24,7 @@ func validateSubvolumegroup(f *framework.Framework, subvolgrp string) error {
 	cmd := fmt.Sprintf("ceph fs subvolumegroup getpath myfs %s", subvolgrp)
 	stdOut, stdErr, err := execCommandInToolBoxPod(f, cmd, rookNamespace)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to exec command in toolbox: %w", err)
 	}
 	if stdErr != "" {
 		return fmt.Errorf("failed to getpath for subvolumegroup %s with error %v", subvolgrp, stdErr)
@@ -106,12 +106,12 @@ func unmountCephFSVolume(f *framework.Framework, appName, pvcName string) error 
 	pod, err := f.ClientSet.CoreV1().Pods(f.UniqueName).Get(context.TODO(), appName, metav1.GetOptions{})
 	if err != nil {
 		e2elog.Logf("Error occurred getting pod %s in namespace %s", appName, f.UniqueName)
-		return err
+		return fmt.Errorf("failed to get pod: %w", err)
 	}
 	pvc, err := f.ClientSet.CoreV1().PersistentVolumeClaims(f.UniqueName).Get(context.TODO(), pvcName, metav1.GetOptions{})
 	if err != nil {
 		e2elog.Logf("Error occurred getting PVC %s in namespace %s", pvcName, f.UniqueName)
-		return err
+		return fmt.Errorf("failed to get pvc: %w", err)
 	}
 	cmd := fmt.Sprintf("umount /var/lib/kubelet/pods/%s/volumes/kubernetes.io~csi/%s/mount", pod.UID, pvc.Spec.VolumeName)
 	_, stdErr, err := execCommandInDaemonsetPod(f, cmd, cephfsDeamonSetName, pod.Spec.NodeName, cephfsContainerName, cephCSINamespace)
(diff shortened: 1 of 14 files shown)


yati1998:
@Madhu-1 created seprate PR for easier merge and also have addressed the comments mentioned by you in https://github.com/ceph/ceph-csi/pull/2025

e2e/cephfs_helper.go
```diff
@@ -24,7 +24,7 @@ func validateSubvolumegroup(f *framework.Framework, subvolgrp string) error {
 	cmd := fmt.Sprintf("ceph fs subvolumegroup getpath myfs %s", subvolgrp)
 	stdOut, stdErr, err := execCommandInToolBoxPod(f, cmd, rookNamespace)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to exec command in toolbox %w", err)
```
nixpanic: I think it is always good to split error reporting with a `:`, like:

```go
fmt.Errorf("failed to exec command in toolbox: %w", err)
```

This also makes `with error` unneeded in other new error returns.

e2e/configmap.go
```diff
@@ -68,14 +68,14 @@ func createConfigMap(pluginPath string, c kubernetes.Interface, f *framework.Fra
...
-			return updateErr
+			return fmt.Errorf("failed to update configmap %w", updateErr)
 		}
 	}
 	if apierrs.IsNotFound(err) {
 		_, err = c.CoreV1().ConfigMaps(cephCSINamespace).Create(context.TODO(), &cm, metav1.CreateOptions{})
 	}
 
-	return err
+	return fmt.Errorf("error=%w", err)
```
nixpanic: This becomes problematic when `err == nil`

yati1998: True, but it shows wrap-check error. will check the alternative in this case.

e2e/configmap.go
```diff
@@ -116,5 +116,5 @@ func createCustomConfigMap(c kubernetes.Interface, pluginPath string, subvolgrpI
 	cm.Namespace = cephCSINamespace
 	// since a configmap is already created, update the existing configmap
 	_, err = c.CoreV1().ConfigMaps(cephCSINamespace).Update(context.TODO(), &cm, metav1.UpdateOptions{})
-	return err
+	return fmt.Errorf("failed to update configmap: %w", err)
```
nixpanic: You will need to change this to

```go
if err != nil {
    return fmt.Errorf("failed to update configmap: %w", err)
}
return nil
```

This is what the e2e jobs cause to fail with:
```
Jun  3 09:32:57.401: failed to create configmap with error failed to update configmap: %!w(<nil>)
```

yati1998: okay got that.

humblec: @nixpanic when update fails, do we really get any error ?

yati1998: @humblec yes, after updating the code as mentioned above, we will get an error. Earlier even if the error was nil due to fmt.Errorf the create configmap was failing.

humblec: hmmm.. some object update errors are ignored from api server iic, any way does not hurt us to have the error return here.

nixpanic (dismissed):

Madhu-1 (dismissed):

Madhu-1 (dismissed):

e2e/pod.go
```diff
@@ -298,7 +298,7 @@ func deletePod(name, ns string, c kubernetes.Interface, t int) error {
 		}
 		e2elog.Logf("%s app  to be deleted (%d seconds elapsed)", name, int(time.Since(start).Seconds()))
 		if err != nil {
-			return false, err
+			return false, fmt.Errorf("failed to get pod: %w", err)
```
humblec: @yati1998 we are interchangeably using `app` and `pod`, I would suggest to keep it in a consistent way like `app`

internal/journal/voljournal.go
```diff
@@ -470,7 +470,7 @@ func reserveOMapName(ctx context.Context, monitors string, cr *util.Credentials,
 				continue
 			}
 
-			return "", err
+			return "", fmt.Errorf("failed to create object: %w", err)
```
humblec: @yati1998  you can better place it as `failed to create omap object for `oMapNamePrefix+iterUUID`

nixpanic (approved):

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

nixpanic:

#1969 caused the issue:
```
Waiting up to 3m0s for all (but 0) nodes to be ready
```

PR #1497, opened 2020-09-21
build: update vendor with latest version

agarwal-mudit:
Updating the version of golang.org/x/text to version v0.3.3.
It fixes the vulnerability in http://golang.org/x/text/encoding/unicode
which leads UTF-16 decoder entering an infinite loop causing
the program to crash or run out of memory.

Signed-off-by: Mudit Agarwal <muagarwa@redhat.com>

PR #2132, opened 2021-06-04
rebase: update golang.org/x/text to 0.3.6

Madhu-1:
update text package to 0.3.6 to fix  CVE-2020-28851

more details at https://github.com/golang/go/issues/42535 and https://bugzilla.redhat.com/show_bug.cgi?id=1922730


agarwal-mudit (approved):

nixpanic (changes requested):
What pulls this in? Can we not update the direct dependency that requires this package?

Madhu-1:
> What pulls this in? Can we not update the direct dependency that requires this package?

different vendors use different versions of text packages as this is a CVE it's good to stick to a version instead of identifying all the dependencies and updating all.

Madhu-1:
we already update the same package previously https://github.com/ceph/ceph-csi/pull/1497

agarwal-mudit:
If its only used for e2e and we don't need to fix this then we can close the PR and bug.
Thanks @nixpanic

nixpanic:
> we already update the same package previously #1497

That probably was not needed either. We should look into updating gomega instead:
```
$ go mod why -m golang.org/x/text
# golang.org/x/text
github.com/ceph/ceph-csi/e2e
github.com/onsi/gomega
github.com/onsi/gomega/matchers
golang.org/x/net/html/charset
golang.org/x/text/encoding
```

As we only pull in the package for e2e testing, I do not see a (strong) need to update the package with a `replace` in go.mod.

PR #2133, opened 2021-06-04
rebase: update packages in go.mod to latest releases

Madhu-1:
updated few packages in `go.mod` to latest available release.


nixpanic (approved):
Thanks!

Is it possible to drop some of the dependencies that are listed as `replace`? That would clean things up even more.

Madhu-1:
> Thanks!
> 
> Is it possible to drop some of the dependencies that are listed as `replace`? That would clean things up even more.

I tried https://github.com/ceph/ceph-csi/blob/devel/go.mod#L40-L41 but the `go mod vendor` failed. still, some work left to update the snapshotter which need E2E changes, I will try to see what can I remove in go.sum

Yuggupta27 (approved):

PR #2095, opened 2021-05-20
e2e: validate encryption keys in KMS

yati1998:
this commit is to validate if the encrypted
keys are created and deleted properly while
pvc-pvc clone images

Updates: #2022

Signed-off-by: Yati Padia <ypadia@redhat.com>

PR #2141, opened 2021-06-07
cleanup: No use of variable validateEncryption

yati1998:
In the function validatePVCSnapshot(...), we don't need
validateEncryption variable as we are passing kms value
which can help us check the value of validateEncryption.
Hence, we can avoid using that.

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -668,7 +668,7 @@ func validatePVCClone(totalCount int, sourcePvcPath, sourceAppPath, clonePvcPath
 }
 
 // nolint:gocyclo,gocognit,nestif // reduce complexity
-func validatePVCSnapshot(totalCount int, pvcPath, appPath, snapshotPath, pvcClonePath, appClonePath, kms string, validateEncryption bool, f *framework.Framework) {
+func validatePVCSnapshot(totalCount int, pvcPath, appPath, snapshotPath, pvcClonePath, appClonePath, kms string, f *framework.Framework) {
 	var wg sync.WaitGroup
 	wgErrs := make([]error, totalCount)
 	chErrs := make([]error, totalCount)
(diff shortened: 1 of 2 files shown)


nixpanic (approved):
Is this part of enabling a new golang-ci linter? If so, mention or link it in the description of the PR.

e2e/utils.go
```diff
@@ -668,7 +668,7 @@ func validatePVCClone(totalCount int, sourcePvcPath, sourceAppPath, clonePvcPath
 }
 
 // nolint:gocyclo,gocognit,nestif // reduce complexity
-func validatePVCSnapshot(totalCount int, pvcPath, appPath, snapshotPath, pvcClonePath, appClonePath, kms string, validateEncryption bool, f *framework.Framework) {
+func validatePVCSnapshot(totalCount int, pvcPath, appPath, snapshotPath, pvcClonePath, appClonePath, kms string, f *framework.Framework) {
```
nixpanic, line 671: You could simplify the changes to the remaining of the function if you add
```go
validateEncryption := (kms != "")
```

yati1998:
> Is this part of enabling a new golang-ci linter? If so, mention or link it in the description of the PR.

No, its just a cleanup from https://github.com/ceph/ceph-csi/pull/2095.

e2e/utils.go
yati1998: isn't it the same as passing the true or false to this variable? As you suggested in https://github.com/ceph/ceph-csi/pull/2095, we can avoid using validateEncryption, if our work is done using kms?

nixpanic: There is no need to pass `validateEncryption` as parameter to the function, but using it inside the function by 'calculating' is fine (and might be more readable).

yati1998: Yeah that's true. I will do that.

humblec (approved):

PR #2145, opened 2021-06-08
ci: pushing artifacts using github actions

### LINKED ISSUE #1781: Add github action to push canary image and helm charts
we are using Travis CI to push the helm charts and the container image when a PR is merged is the branch. It would be good if we move from Travis CI to GitHub action to push these artifacts.

Madhu-1:
As Travis CI `https://travis-ci.org/` is getting shutdown date on June 15th. Either we need to move to a new place https://www.travis-ci.com/ or we can switch to GitHub action to push the image and the helm charts when a PR is merged.

fixes: #1781


Yuggupta27:
ci failure due to #2146

Madhu-1:
@nixpanic @Yuggupta27 looks like Jenkins is down? https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/mini-e2e-helm_k8s-1.19/1444/display/redirect

Yuggupta27:
> @nixpanic @Yuggupta27 looks like Jenkins is down? https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/mini-e2e-helm_k8s-1.19/1444/display/redirect

Yes, Jenkins is down currently and the jobs are inaccessible.

deploy.sh
```diff
@@ -82,41 +50,64 @@ build_push_images() {
...
+	VERSION=${CSI_IMAGE_VERSION//v/} # Set version (without v prefix)
 
-if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
-	build_step "log in to quay.io as user ${QUAY_IO_USERNAME}"
-	# This is a workaround to fix docker permission denied issue during manifest create in Travis CI.
-	# `docker manifest create` fails due to permission denied on `/etc/docker/certs.d/quay.io`
-	# (https://github.com/docker/for-linux/issues/396).
-	sudo chmod o+x /etc/docker
+	# update information in Chart.yaml if the branch is not devel
+	if [ "$BRANCH_NAME" != "devel" ]; then
```
nixpanic: Please use `${BRANCH_NAME}` notation, not variables without `${..}`

deploy.sh
```diff
@@ -73,15 +73,15 @@ push_helm_charts() {
 
 	mkdir -p "$CHARTDIR/csi-charts/docs/$PACKAGE"
 	cp -R "./charts/ceph-csi-$PACKAGE" "$CHARTDIR/csi-charts/docs/$PACKAGE"
-	pushd "$CHARTDIR/csi-charts/docs/$PACKAGE" >/dev/null
+	pushd "$CHARTDIR/csi-charts/docs/$PACKAGE" >/dev/null || exit
```
nixpanic: Would you not need to fail the script when an error occurs? Use `exit 1` instead.

Or use `set -e` in the top of the script to exit on any error in the script.

nixpanic (changes requested):

deploy.sh
Madhu-1: added `set -e` now dropping exit 1

.github/workflows/publish-artifacts.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Login to Quay
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_IO_USERNAME }}
+          password: ${{ secrets.QUAY_IO_PASSWORD }}
+
+      - name: Set build environment based on Git branch name
+        if: github.ref == 'refs/heads/devel'
+        run: echo "BRANCH_NAME=devel" >> $GITHUB_ENV
```
nixpanic, line 28: Instead of hardcoding `devel`, maybe use the branch that is set in the environment `${GITHUB_BASE_REF}`?

nixpanic: Also, what is `$GITHUB_ENV`? It is not listed on https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

deploy.sh
```diff
@@ -82,41 +53,64 @@ build_push_images() {
...
-if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
-	build_step "log in to quay.io as user ${QUAY_IO_USERNAME}"
-	# This is a workaround to fix docker permission denied issue during manifest create in Travis CI.
-	# `docker manifest create` fails due to permission denied on `/etc/docker/certs.d/quay.io`
-	# (https://github.com/docker/for-linux/issues/396).
-	sudo chmod o+x /etc/docker
+	# update information in Chart.yaml if the branch is not devel
+	if [ "${BRANCH_NAME}" != "devel" ]; then
+		# Replace appVersion: canary and version: *-canary with the actual version
+		sed -i "s/\(\s.*canary\)/ $VERSION/" "charts/ceph-csi-$PACKAGE/Chart.yaml"
```
nixpanic: Still many bash variables without `${..}`

nixpanic (changes requested):

.github/workflows/publish-artifacts.yaml
Madhu-1: `GITHUB_BASE_REF` is only available on pull_request not on push. GITHUB_ENV is available here https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

deploy.sh
Madhu-1: these are not added as part of this PR but let me fix it.

.github/workflows/publish-artifacts.yaml
nixpanic: Ah, right, maybe use `${GITHUB_REF}` then? Otherwise there would need to be a way to select the affected `release-v*` branch.

Madhu-1: thought of doing it but again GITHUB_REF needs some striping and the branch is compared with only `devel` in `deploy.sh` IMO we are good for now. I will check about automating the Image name and the helm chart version from the  branch or the tag name (need more work for that one)

.github/workflows/publish-artifacts.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: publish artifacts
+        # podman cannot pull images with both tag and digest
+        # https://github.com/containers/buildah/issues/1407
```
humblec, line 32: @Madhu-1 do we have an issue here because of the limitation mentioned in the podman? I thought we only do `tag` based pulling. Isnt it the case?

humblec:
The change looks good to me :+1: , will wait for couple of comments in the review to be addressed.

.github/workflows/publish-artifacts.yaml
Madhu-1: It was not working earlier this is taken from https://github.com/ceph/ceph-csi/blob/devel/.github/workflows/build-multi-stage.yaml#L15-L17. We do pull based on the `sha`, not the tag.

Madhu-1:
> The change looks good to me  , will wait for couple of comments in the review to be addressed.

All the comments are already addressed/replied to. Did I miss any?

humblec:
> > The change looks good to me  , will wait for couple of comments in the review to be addressed.
> 
> All the comments are already addressed/replied to. Did I miss any?

https://github.com/ceph/ceph-csi/pull/2145#issuecomment-856682589 , I had a query here :)  . Also Niels' change requested still exist.

Madhu-1:
> > > The change looks good to me  , will wait for couple of comments in the review to be addressed.
> > 
> > 
> > All the comments are already addressed/replied to. Did I miss any?
> 
> [#2145 (comment)](https://github.com/ceph/ceph-csi/pull/2145#issuecomment-856682589) , I had a query here :) . Also Niels' change requested still exist.

I hope already replied to all the comments. @nixpanic @humblec PTAL

.github/workflows/publish-artifacts.yaml
humblec: sure, so we need only digest :) and not both ?  in short, do we need that caution/declaimer

Madhu-1: We need both. This is only to build the multistage container image.

humblec (dismissed):

.github/workflows/publish-artifacts.yaml
```diff
@@ -0,0 +1,34 @@
...
+        # use docker to build images
+        run: GITHUB_TOKEN={{ $GITHUB_TOKEN }} CONTAINER_CMD=docker ./deploy.sh
```
nixpanic: Is the `$GITHUB_TOKEN` not set in the environment? Adding it on the commandline might expose it in logs...

nixpanic:
How about removing the travis job with this PR too?

nixpanic:
> How about removing the travis job with this PR too?

Note, the job is removed, but other scripts and references are still there.

Madhu-1:
> > How about removing the travis job with this PR too?
> 
> Note, the job is removed, but other scripts and references are still there.

Removed most of the references

.github/workflows/publish-artifacts.yaml
Madhu-1: No, this should be `secret.GITHUB_TOKEN` as for my testing goes this won't be logged in the runners.

Madhu-1: But let me double-check and confirm this one.

.github/workflows/publish-artifacts.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Set build environment based on Git branch name
+        if: github.ref == 'refs/heads/devel'
+        run: echo "BRANCH_NAME=devel" >> $GITHUB_ENV
+
+      - name: publish artifacts
+        # podman cannot pull images with both tag and digest
+        # https://github.com/containers/buildah/issues/1407
+        # use docker to build images
+        # yamllint disable-line rule:line-length
+        run: GITHUB_TOKEN={{ $secrets.GITHUB_TOKEN }} CONTAINER_CMD=docker ./deploy.sh
```
nixpanic: You have notation `${{ secrets.GITHUB_TOKEN }}` for logging into Quay.io. Which is correct?

.github/workflows/publish-artifacts.yaml
Madhu-1: This won't get printed here is the test https://github.com/Madhu-1/test-github-actions/pull/4/files and https://github.com/Madhu-1/test-github-actions/pull/4/checks?check_run_id=2773891438

nixpanic: thanks for confirming!

.github/workflows/publish-artifacts.yaml
Madhu-1: This is to push helm charts https://github.com/ceph/csi-charts

nixpanic: Yes, but I mean
```
{{ $secrets.GITHUB_TOKEN }}
```
vs
```
${{ secrets.GITHUB_TOKEN }}
```

Madhu-1: Yes, correct. my bad. This is fixed now.

nixpanic (approved):

Madhu-1:
@humblec PTAL

humblec (approved):

Madhu-1:
Manually merging this one as the configurations are changed

Madhu-1:
@mergifyio refresh

Madhu-1:
@mergifyio backport release-v3.3

Issue #2142, opened 2021-06-07
cephfs: efficient read-only snapshots as volumes


gman0:
# Describe the feature you'd like to have #

CephFS exposes snapshots as special, read-only directories of a subvolume. cephfs-csi can already provision writable volumes with snapshots as their data source, where snapshot contents are cloned to the newly created volume. Cloning to snapshot to volume is very expensive in CephFS, as the data needs to be fully copied.

I propose creating read-only (shallow) volumes from snapshots is added to cephfs-csi, where the driver would take advantage of the fact that snapshot's data is already available in a directory, and that directory would be exposed to workloads as a volume. This would make it possible to avoid the expensive cloning process.

# What is the value to the end user? (why is it a priority?) #

What's the point of such read-only volumes?
* **Restore snapshots selectively:** users may want to traverse snapshots, restoring data to a writable volume more selectively instead of restoring the whole snapshot.
* **Volume backup:** users can't backup a live volume, they first need to snapshot it. Once a snapshot is take, it still can't be backed-up, as backup tools usually work with volumes (that are exposed as file-systems) and not snapshots (which might have backend-specific format). What this means is that currently with cephfs-csi we'd have to clone snapshot data twice:
   1. first time, when restoring the snapshot into a volume,
   2. and second time, when transferring that volume into some backup/archive storage (e.g. object store).
   
   The backed-up volume will most likely be thrown away after the backup is finished. That's a lot of wasted work for what we originally wanted to do! Having the ability to create volumes from snapshots cheaply would be a big improvement for this use case.

# How will we know we have a good solution? (acceptance criteria) #

Creating/deleting and mounting/unmounting such read-only volumes should be O(1) regardless of snapshot size.

# Alternatives

* _User's could simply go to `.snap/<snapshot name>` by themselves._
   `.snap` is CephFS-specific detail of how snapshots are exposed. Users / tools may not be aware of this special directory, or it doesn't fit the workflow. The idiomatic way of accessing snapshot contents is by creating a new volume and populating it with snapshot.

# Additional context #

If you agree with the proposed feature, I can add a PR for a `docs/proposals` document where we can discuss this in more detail. After that I can work on its implementation.

Madhu-1:
@gman0 sounds good hope we can see a design doc soon on this one. Also, note we are planning for cephfs mirroring support in cephcsi soon which will directly replicate the volume to remote ceph cluster but looks like this is useful if someone wants to do backup to some other store (object store).

[...]

Issue #742, opened 2019-12-12
Kubernetes sidecar leaders and its impact on controller plugin operations, needs understanding/clarification


ShyamsundarR:
We now will have 4 side cars performing [leader election](https://github.com/ceph/ceph-csi/blob/8a16f740d68b3f31b1955f81f1a4994aa72fd68b/deploy/rbd/kubernetes/v1.14%2B/csi-rbdplugin-provisioner.yaml#L38-L76), and hence, **to my understanding**, each sidecar can determine in isolation that one of the many (two in the default case) available controller plugin instances are the leader for the said sidecar.

This defeats the controller locks against volumes if different operations will reach different controller plugins. This either needs to be understood better (IOW, the problem does not exist) or we may need to work with kubernetes CSI side cars to see how best to ensure all side car leaders land on the same instance.

For example it could mean the provisioner leader instance is on a different node than the snapshotter instance, and so the locks that we take between `DeleteVolume` and `CreateSnapshot` will not contend and can cause unwanted repercussions.

If this is a problem, we need to find ways to mitigate the issue, or look at how we can influence the various sidecars to a single leader (same lock name maybe?).

WoohyungHan:
+1 very interesting problem. But I think `solving by same lock name` can decrease overall performance. Could we use per volume lock? I think there is no influence multiple sidecars working on individual, different volume.

Madhu-1:
can be fixed by running single CSI sidecar images, more info at https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-sig-storage-wg-csi/qZvuOlCME5Y/YzKuurLVBwAJ

Madhu-1:
> +1 very interesting problem. But I think `solving by same lock name` can decrease overall performance. Could we use per volume lock? I think there is no influence multiple sidecars working on individual, different volume.

currently, we are locking per  volume in CSI  container and its in-memory lock, as the lock is in memory and not centralized lock, we can hit the issue like on one provisioner pod we will be deleting the volume and in another provisioner pod we will  be resizing it or cloning from it( possible to hit this issue)

ShyamsundarR:
> > +1 very interesting problem. But I think `solving by same lock name` can decrease overall performance. Could we use per volume lock? I think there is no influence on multiple sidecars working on an individual, different volume.
> 
[...]

PR #414, opened 2019-06-10
Use Deployment with leader election instead of StatefulSet

Madhu-1:
# Describe what this PR does #
Deployment behaves better when a node gets disconnected from the rest of the cluster - new provisioner leader is elected in ~15 seconds, while it may take up to 5 minutes for StatefulSet to start a new replica.

Refer: https://github.com/kubernetes-csi/external-provisioner/commit/52d1fbcf9d52d35053164bb0409aa97dd3e7777a

## Related issues ##
Fixes: #335

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

PR #2148, opened 2021-06-09
proposal: CephFS snapshots as shallow RO volumes

gman0:
This PR adds a proposal document for 'CephFS snapshots as shallow read-only volumes' feature.

Related issue: https://github.com/ceph/ceph-csi/issues/2142

### DIFF
--- a/docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
+++ b/docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
@@ -1,0 +1,272 @@
+# Snapshots as shallow read-only volumes
+
+CSI spec doesn't have a notion of "mounting a snapshot". Instead, the idiomatic
+way of accessing snapshot contents is first to create a volume populated with
+snapshot contents and then mount that volume to workloads.
+
+CephFS exposes snapshots as special, read-only directories of a subvolume
+located in `<subvolume>/.snap`. cephfs-csi can already provision writable
+volumes with snapshots as their data source, where snapshot contents are
+cloned to the newly created volume. However, cloning a snapshot to volume
+is a very expensive operation in CephFS as the data needs to be fully copied.
+When the need is to only read snapshot contents, snapshot cloning is extremely
+inefficient and wasteful.
+
+This proposal describes a way for cephfs-csi to expose CephFS snapshots
+as shallow, read-only volumes, without needing to clone the underlying
+snapshot data.
+
+## Use-cases
+
+What's the point of such read-only volumes?
+
+* **Restore snapshots selectively:** users may want to traverse snapshots,
+  restoring data to a writable volume more selectively instead of restoring
+  the whole snapshot.
+* **Volume backup:** users can't backup a live volume, they first need
+  to snapshot it. Once a snapshot is taken, it still can't be backed-up,
+  as backup tools usually work with volumes (that are exposed as file-systems)
+  and not snapshots (which might have backend-specific format). What this means
+  is that in order to create a snapshot backup, users have to clone snapshot
+  data twice:
+
+  1. first time, when restoring the snapshot into a temporary volume from
+     where the data will be read,
+  1. and second time, when transferring that volume into some backup/archive
+     storage (e.g. object store).
+
+  The temporary backed-up volume will most likely be thrown away after the
+  backup transfer is finished. That's a lot of wasted work for what we
+  originally wanted to do! Having the ability to create volumes from
+  snapshots cheaply would be a big improvement for this use case.
+
+## Alternatives
+
+* _Snapshots are stored in `<subvolume>/.snap`. Users could simply visit this
+  directory by themselves._
+
+  `.snap` is CephFS-specific detail of how snapshots are exposed.
+  Users / tools may not be aware of this special directory, or it may not fit
+  their workflow. At the moment, the idiomatic way of accessing snapshot
+  contents in CSI drivers is by creating a new volume and populating it
+  with snapshot.
+
+## Design
+
+Key points:
+
+* Volume source is a snapshot, volume access mode is `*_READER_ONLY`.
+* No actual new subvolumes are created in CephFS.
+* The resulting volume is a reference to the source subvolume snapshot.
+  This reference would be stored in `Volume.volume_context` map. In order
+  to reference a snapshot, we need subvol name and snapshot name.
+* Mounting such volume means mounting the respective CephFS subvolume
+  and exposing the snapshot to workloads.
+* Let's call a *shallow read-only volume with a subvolume snapshot
+  as its data source* just a *shallow volume* from here on out for brevity.
+
+### Controller operations
+
+Care must be taken when handling life-times of relevant storage resources.
+When a shallow volume is created, what would happen if:
+
+* _Parent subvolume of the snapshot is removed while the shallow volume
+  still exists?_
+
+  This shouldn't be a problem already. The parent volume has either
+  `snapshot-retention` subvol feature in which case its snapshots remain
+  available, or if it doesn't have that feature, it will fail to be deleted
+  because it still has snapshots associated to it.
+* _Source snapshot from which the shallow volume originates is removed while
+  that shallow volume still exists?_
+
+  We need to make sure this doesn't happen and some book-keeping
+  is necessary. Ideally we could employ some kind of reference counting.
+
+#### Reference counting for shallow volumes
+
+As mentioned above, this is to protect shallow volumes, should their source
+snapshot be requested for deletion.
+
+When creating a volume snapshot, a reference tracker (RT), represented by a
+RADOS object, would be created for that snapshot. It would store information
+required to track the references for the backing subvolume snapshot. Upon a
+`CreateSnapshot` call, the reference tracker (RT) would be initialized with a
+single reference record, where the CSI snapshot itself is the first reference
+to the backing snapshot. Each subsequent shallow volume creation would add a
+new reference record to the RT object. Each shallow volume deletion would
+remove that reference from the RT object. Calling `DeleteSnapshot` would remove
+the reference record that was previously added in `CreateSnapshot`.
+
+The subvolume snapshot would be removed from the Ceph cluster only once the RT
+object holds no references. Note that this behavior would permit calling
+`DeleteSnapshot` even if it is still referenced by shallow volumes.
+
+* `DeleteSnapshot`:
+  * RT holds no references or the RT object doesn't exist:
+    delete the backing snapshot too.
+  * RT holds at least one reference: keep the backing snapshot.
+* `DeleteVolume`:
+  * RT holds no references: delete the backing snapshot too.
+  * RT holds at least one reference: keep the backing snapshot.
+
+To enable creating shallow volumes from snapshots that were provisioned by
+older versions of cephfs-csi (i.e. before this feature is introduced),
+`CreateVolume` for shallow volumes would also create an RT object in case it's
+missing. It would be initialized to two: the source snapshot and the newly
+created shallow volume.
+
+##### Concurrent access to RT objects
+
+RADOS API provides access to compound atomic read and write operations. These
+will be used to implement reference tracking functionality, protecting
+modifications of reference records.
+
+#### `CreateVolume`
+
+A read-only volume with snapshot source would be created under these conditions:
+
+1. `CreateVolumeRequest.volume_content_source` is a snapshot,
+1. `CreateVolumeRequest.volume_capabilities[*].access_mode` is any of read-only
+   volume access modes.
+1. Possibly other volume parameters in `CreateVolumeRequest.parameters`
+   specific to shallow volumes.
+
+`CreateVolumeResponse.Volume.volume_context` would then contain necessary
+information to identify the source subvolume / snapshot.
+
+Things to look out for:
+
+* _What's the volume size?_
+
+  It doesn't consume any space on the filesystem. `Volume.capacity_bytes` is
+  allowed to contain zero. We could use that.
+* _What should be the requested size when creating the volume (specified e.g.
+  in PVC)?_
+
+  This one is tricky. CSI spec allows for
+  `CreateVolumeRequest.capacity_range.{required_bytes,limit_bytes}` to be
+  zero. On the other hand,
+  `PersistentVolumeClaim.spec.resources.requests.storage` must be bigger
+  than zero. cephfs-csi doesn't care about the requested size (the volume
+  will be read-only, so it has no usable capacity) and would always set it
+  to zero. This shouldn't case any problems for the time being, but still
+  is something we should keep in mind.
+
+`CreateVolume` and behavior when using volume as volume source (PVC-PVC clone):
+
+| New volume     | Source volume  | Behavior                                                                          |
+|----------------|----------------|-----------------------------------------------------------------------------------|
+| shallow volume | shallow volume | Create a new reference to the parent snapshot of the source shallow volume.       |
+| regular volume | shallow volume | Equivalent for a request to create a regular volume with snapshot as its source.  |
+| shallow volume | regular volume | Such request doesn't make sense and `CreateVolume` should return an error.        |
+
+### `DeleteVolume`
+
+Volume deletion is trivial.
+
+### `CreateSnapshot`
+
+Snapshotting read-only volumes doesn't make sense in general, and should
+be rejected.
+
+### `ControllerExpandVolume`
+
+Same thing as above. Expanding read-only volumes doesn't make sense in general,
+and should be rejected.
+
+## Node operations
+
+Two cases need to be considered:
+
+* (a) Volume/snapshot provisioning is handled by cephfs-csi
+* (b) Volume/snapshot provisioning is handled externally (e.g. pre-provisioned
+  manually, or by OpenStack Manila, ...)
+
+### `NodeStageVolume`, `NodeUnstageVolume`
+
+Here we're mounting the source subvolume onto the node. Subsequent volume
+publish calls then use bind mounts to expose the snapshot directory located in
+`.snap/<SNAPSHOT DIRECTORY NAME>`. Unfortunately, we cannot mount snapshots
+directly because they are not visible during mount time. We need to mount the
+whole subvolume first, and only then perform the binds to target paths.
+
+#### For case (a)
+
+Subvolume paths are normally retrieved by
+`ceph fs subvolume info/getpath <VOLUME NAME> <SUBVOLUME NAME> <SUBVOLUMEGROUP NAME>`,
+which outputs a path like so:
+
+```
+/volumes/<VOLUME NAME>/<SUBVOLUME NAME>/<UUID>
+```
+
+Snapshots are then accessible in:
+
+* `/volumes/<VOLUME NAME>/<SUBVOLUME NAME>/.snap` and
+* `/volumes/<VOLUME NAME>/<SUBVOLUME NAME>/<UUID>/.snap`.
+
+`/volumes/<VOLUME NAME>/<SUBVOLUME NAME>/<UUID>` may be deleted if the source
+subvolume is deleted, but thanks to the `snapshot-retention` feature, snapshots
+in `/volumes/<VOLUME NAME>/<SUBVOLUME NAME>/.snap` will remain to be available.
+
+The CephFS mount should therefore have its root set to the parent of what
+`fs subvolume getpath` returns, i.e. `/volumes/<VOLUME NAME>/<SUBVOLUME NAME>`.
+That way we will have snapshots available regardless of whether the subvolume
+itself still exists or not.
+
+#### For case (b)
+
+For cases where subvolumes are managed externally and not by cephfs-csi, we
+must assume that the cephx user we're given can access only
+`/volumes/<VOLUME NAME>/<SUBVOLUME NAME>/<UUID>` so users won't be able to
+benefit from snapshot retention. Users will need to be careful not to delete
+the parent subvolumes and snapshots while they are associated by these shallow
+RO volumes.
+
+### `NodePublishVolume`, `NodeUnpublishVolume`
+
+Node publish is trivial. We bind staging path to target path as a read-only
+mount.
+
+### `NodeGetVolumeStats`
+
+`NodeGetVolumeStatsResponse.usage[*].available` should be always zero.
+
+## Volume parameters, volume context
+
+This section provides a discussion around determinig what volume parameters and
+volume context parameters will be used to convey necessary information to the
+cephfs-csi driver in order to support shallow volumes.
+
+Volume parameters `CreateVolumeRequest.parameters`:
+
+* Should be "shallow" the default mode for all `CreateVolume` calls that have
+  (a) snapshot as data source and (b) read-only volume access mode? If not,
+  a new volume parameter should be introduced: e.g `isShallow: <bool>`. On the
+  other hand, does it even makes sense for users to want to create full copies
+  of snapshots and still have them read-only?
+
+Volume context `Volume.volume_context`:
+
+* Here we definitely need `isShallow` or similar. Without it we wouldn't be
+  able to distinguish between a regular volume that just happens to have
+  a read-only access mode, and a volume that references a snapshot.
+* Currently cephfs-csi recognizes `subvolumePath` for dynamically provisioned
+  volumes and `rootPath` for pre-previsioned volumes. As mentioned in
+  [`NodeStageVolume`, `NodeUnstageVolume` section](#NodeStageVolume-NodeUnstageVolume),
+  snapshots cannot be mounted directly. How do we pass in path to the parent
+  subvolume?
+  * a) Path to the snapshot is passed in via `subvolumePath` / `rootPath`,
+    e.g.
+    `/volumes/<VOLUME NAME>/<SUBVOLUME NAME>/<UUID>/.snap/<SNAPSHOT NAME>`.
+    From that we can derive path to the subvolume: it's the parent of `.snap`
+    directory.
+  * b) Similar to a), path to the snapshot is passed in via `subvolumePath` /
+    `rootPath`, but instead of trying to derive the right path we introduce
+    another volume context parameter containing path to the parent subvolume
+    explicitly.
+  * c) `subvolumePath` / `rootPath` contains path to the parent subvolume and
+    we introduce another volume context parameter containing name of the
+    snapshot. Path to the snapshot is then formed by appending
+    `/.snap/<SNAPSHOT NAME>` to the subvolume path.


nixpanic:
This looks well thought out :+1: 

Obviously the reference counting is the most important factor for this. It would be the main part where the complexity needs to be handled. I am not sure the additional complexity is worth the benefits.

Would this be something that will be used a lot in certain type of workloads? The use-cases in the document are very clear, but I would like to know more about the actual applications and increase of performance (why is it needed, or is it just nice to have?) for these.

gman0:
@nixpanic I'm discussing this proposal with Kubernetes Data Protection WG still. I'll try to give more concrete examples soon.

gman0:
And just to make it clear, I wouldn't be trying to solve the "multi-master race". At least not as a part of this proposal anyway. There are a few other places in the driver where this is a problem too, no? Or at least I remember reading something along these lines... It would be nice if we could solve it for all these cases at once.

humblec:
@gman0  The  proposal looks good to me :+1:   the shallow RO volume mount for an export make sense to transfer a backup to external source.  One caveat here is fitting this workflow on  data protection w efforts in general which you already taken into account and figuring out .

gman0:
@nixpanic after checking with Data Protection WG, @xing-yang, this proposal should fit in nicely with the planned Volume backup feature - which itself is a part of bigger Application backup feature. It's desirable for a CSI driver to be able to create volumes from snapshots efficiently. All the heavy I/O would be done only when copying the data to the backup/archive storage. This is as opposed to having to create a clone of the snapshot just so that it can be read from when transferring to backup storage. Depending on the amount of data and current load on the Ceph cluster, the increase in performance may be significant.

Right now this is usable with Velero/Restic. We're planning to use this for regular volume/app backups for our users. Later, I'd like to propose to Velero folks to include "volume from snapshot" as an intermediate step when performing volume backups using their [CSI plugin](https://github.com/vmware-tanzu/velero-plugin-for-csi).

gman0:
I've squashed commits for fixing markdown linter errors and added a line saying that solving the race condition for when there are multiple controller plugins running concurrently is out of scope for this proposal. Commit lint is still complaining though, something about subject being empty. I'm not sure what's up with that.

gman0:
@nixpanic does this seem reasonable to you?

gman0:
Is there anything else needed for this PR? @Madhu-1 @ShyamsundarR @humblec PTAL

Madhu-1:
commitlint is failing as the commit description is not present. https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#commit-messages

Madhu-1:
Busy with some work, I will revisit this soon. Thanks, @gman0

gman0:
@Madhu-1 thanks. I can squash the commits that fix the linter errors with the original commit later when we're done with reviews. That should satisfy commitlint too I hope.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+Mutexes won't cover situations where multiple masters may be active though
+(during/after leader election), and modifying _refcount_ would still race in
+this case. Misbehaving master that lost leadership would still be able to
+access resources on the Ceph cluster. Racing on _refcount_ could then cause
+leaking the _refcount_ object itself and/or the subvolume snapshot, or the
+provisioned shallow volume could point to a subvolume snapshot that's already
+been deleted. This is not easily fixable without the individual master replicas
+being aware of the leader election process and is out-of-scope for this
+proposal.
+
```
Madhu-1: Currently, also this problem exists. but this could be a bigger problem as we don't have any lease coordination with the sidecar we are running. as we allow multiple replicas in provisioner deployment. in one pod the external-provisioner might be a leader and in the other pod, the external-snapshotter is acting as a leader?

gman0:
Any updates on reviews on this PR please?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+  This shouldn't be a problem already. The parent volume has either
+  `snapshot-retention` subvol feature in which case its snapshots remain
+  available, or if it doesn't have that feature, it will fail to be deleted
+  because it still has snapshots associated to it.
+* _Source snapshot from which the shallow volume originates is removed while
+  that shallow volume still exists?_
+
+  We need to make sure this doesn't happen and some book-keeping
+  is necessary. Ideally we could employ some kind of reference counting.
```
ShyamsundarR, line 84: I read till here, and this is a useful proposal. I did have similar thoughts earlier but did not take it forward, so thanks @gman0 .

I will review the rest next week. Need to look at this from a CephFS subvolume perspective on snapshot locations and snap path exposed/staged for workloads (just a note to self).

gman0: @ShyamsundarR cool, thanks!

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+
+#### Reference counting for shallow volumes
```
Madhu-1, line 86: Do we need to consider PVC-PVC clone case from PVC creating as shallow volume? to allow disallow the PVC-PVC clone operation?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+Things to look out for:
+
+* _What's the volume size?_
+
+  It doesn't consume any space on the filesystem. `Volume.capacity_bytes` is
+  allowed to contain zero. We could use that.
+* _What should be the requested size when creating the volume (specified e.g.
+  in PVC)?_
+
+  This one is tricky. CSI spec allows for
```
Madhu-1, line 147: Yes even though there is no consumption we need to return the requested size in response to make sure no Error happens at the kubernetes level.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+
+### `CreateSnapshot`
+
+Snapshotting read-only volumes doesn't make sense in general, and should
+be rejected.
+
+### `ControllerExpandVolume`
+
+Same thing as above. Expanding read-only volumes doesn't make sense in general,
+and should be rejected.
```
Madhu-1, line 176: from the kubernetes point of view, it's a new PVC. consider adding CreateVolume from snapshot and CreateVolume from volume case here as well.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+
+Mutexes won't cover situations where multiple masters may be active though
+(during/after leader election), and modifying _refcount_ would still race in
+this case. Misbehaving master that lost leadership would still be able to
+access resources on the Ceph cluster. Racing on _refcount_ could then cause
+leaking the _refcount_ object itself and/or the subvolume snapshot, or the
+provisioned shallow volume could point to a subvolume snapshot that's already
+been deleted. This is not easily fixable without the individual master replicas
+being aware of the leader election process and is out-of-scope for this
+proposal.
```
Madhu-1: Add a note in the document when we implement this one. To use this feature and to avoid issues run Deployment with replica `1`

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+
+* Here we definitely need `isShallow` or similar. Without it we wouldn't be
+  able to distinguish between a regular volume that just happens to have
+  a read-only access mode, and a volume that references a snapshot.
+* Currently cephfs-csi recognizes `subvolumePath` for dynamically provisioned
+  volumes and `rootPath` for pre-previsioned volumes. As mentioned in
+  [`NodeStageVolume`, `NodeUnstageVolume` section](#NodeStageVolume-NodeUnstageVolume),
+  snapshots cannot be mounted directly. How do we pass in path to the parent
+  subvolume?
+  * a) Path to the snapshot is passed in via `subvolumePath` / `rootPath`,
```
Madhu-1, line 260: to keep things simple let's have an extra path for snapshot also and this will be used only in the NodePublish right?

Madhu-1 (commented):
overall LGTM. left few comments

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: Yeah I was thinking about this scenario too. I see that in another commet you already suggested setting Deployment replicas to 1. I'm wondering whether it shouldn't be the default anyway though, and whether it's enough to protect from similar races in general. With Deployments r=1 there still could be more than one replica running concurrently (albeit the others would be in Terminating state), so there's still a chance for a race to occur no? StatefulSets provide the uniqueness guarantee which would take care of this problem altogether.

Whenever ceph-csi needs a mutex to protect Ceph resources from concurrent access, it's usually a good indicator some locking / fencing between controller replicas is also necessary, should a leader switch occur. What leader election provides in sidecars is simply not enough:

https://github.com/kubernetes/client-go/blob/release-14.0/tools/leaderelection/leaderelection.go#L17-L20

How much availability would ceph-csi lose if it were to simply ditch leader election and use a single-replica StatefulSets instead? Please correct me if I'm wrong, but isn't the behavior for resolving failure with and without leader election very similar? 
* With leader election: leader controller plugin crashes / becomes unresponsive, another replica is selected as the new leader.
* Without leader election: controller plugin crashes / becomes unresponsive (signaled by a liveness probe), Kubernetes takes care of recreating the Pod.

In both of those cases some unavailability is present whether it is due to timing out and finding a new leader, or creating a new Pod. Except that multiple replicas + leader election run the risk of race conditions. Maybe we can discuss this more in a separate ticket.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: Good catch!

* request to create **shallow volume** with **shallow volume** as its source: basically a new reference to the parent snapshot,
* request to create **regular volume** with **shallow volume** as its source: equivalent for a request to create a regular volume with snapshot as its source,
* request to create **shallow volume** with **regular volume** as its source: this doesn't make sense and should exit with an error.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: :+1: yes

gman0: When do we look up the path of the snapshot itself?

The dirname has following format: `_<SNAP NAME>_<INODE NUMBER>`. We don't know what it is the directory called unless we actually iterate through `.snap` and find the correct path, i.e. the directory whose name contains `<SNAP NAME>`.

https://github.com/ceph/ceph/blob/a17ebc0406ac5592911fa7f596464c9c848a9c1e/src/mds/snap.cc#L81-L91

When?
* a) In `CreateVolume`. We'd need to mount the volume, look up the snapshot dir, and then unmount the volume. Dirname would be then stored in the response, in `Volume.volume_context["snapshotDirname"]`.
   * Pros: lookup is done only once per volume.
   * Cons: pre-provisioned volumes would have to do the lookup themselves in order for them to populate `Volume.volume_context["snapshotDirname"]`.
* b) In `NodeStageVolume`. `Volume.volume_context["snapshotName"]` must be populated. In this case we'd also need to mount the volume, but we'd need to do that either way. The only extra bit of work is doing the lookup once per PVC per node.
   * Pros: dynamically provisioned and pre-provisioned volumes are treated the same way.
   * Cons: one extra directory lookup once per PVC per node. Probably negligible performance penalty.
* c) Combination of a) and b). `CreateVolume` would populate `Volume.volume_context["snapshotDirname"]`. If it's not present in volume context during `NodeStageVolume` (in case of pre-provisioned volumes), but `snapshotName` is, it would do the lookup then.

Madhu-1: If the subvolume path is `csi-vol-1d17e149-f522-11eb-938d-866c88286c89/e50a2a89-a4fb-4fca-b590-8e8476e8a1a6/` and the snapshot path is present at `csi-vol-1d17e149-f522-11eb-938d-866c88286c89/.snap/csi-snap-f0a7e7be-f524-11eb-938d-866c88286c89` with this we can mount the volume at the stagingPath? what do you think?

Madhu-1: cc @kotreshhr

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
Madhu-1: Yes it's a bit tricky and needs a lot of testing. we moved from statefulset to deployment due to the downtime issues statefulset. and deployment with replica:2/3 got added to have the HA. but with this one, we have a leader election problem. @ShyamsundarR opened an issue for this one https://github.com/ceph/ceph-csi/issues/742 I think we need to check with the CSI team as this can be a problem? but if CSI is stateless this does not matter but for us, it matters as things can go wrong with the counter.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: Ah right, snapshot naming is different in parent dir. Thanks.
```
# pwd
.../volumes/csi/csi-vol-9dba52f1-f54b-11eb-b0aa-0242ac110006
# ls .snap
csi-snap-b7063583-f54b-11eb-b0aa-0242ac110006
```

The problem is when provisioning is managed by an external service (e.g. OpenStack Manila). In such environments the cephx clients have access restricted to the subvolume path only. `.snap` in parent is inaccessible, so they'd need to go to `subvol/.snap`:
```
# pwd
.../volumes/csi/csi-vol-9dba52f1-f54b-11eb-b0aa-0242ac110006/e6bb7726-71f3-4d32-940c-c34ea08ed383
# ls .snap
_csi-snap-b7063583-f54b-11eb-b0aa-0242ac110006_2199023257056
```
And here you can see the `_<SNAP NAME>_<INODE NUMBER>` name format I mentioned in my comment above. Since we don't know the ino num beforehand, *somebody* needs to do a name lookup.

gman0: So for volumes provisioned by cephfs-csi, dirname is trivial to figure out.

Who should do snap dirname lookup for pre-provisioned volumes / volumes managed externally?
* a) The consumer. cephfs-csi would take dirname via volume context. This means that the consumer needs to mount the volume, perform the lookup, unmount it, and then create a pre-provisioned shallow volume with dirname specified.
* b) cephfs-csi. Since cephfs-csi already needs to mount the whole volume, it could easily do the dirname lookup itself. In this case users would only need to supply the snapshot name.

Which do you think makes more sense? b) is obviously a lot more user-friendly -- at the price of a bit more complexity.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: Thanks for the pointer, @Madhu-1 ! I suppose this is the problem you've mentioned https://github.com/ceph/ceph-csi/pull/414 ?

Madhu-1: Yes, correct that's the one.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
Madhu-1: for pre-provisioned volume, i think we can go for `option a` to keep it simple and as the user is manually creating it admin can populate the volumeContext with required snap details so cephcsi will mount it

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+
+## Use-cases
+
+What's the point of such read-only volumes?
+
+* **Restore snapshots selectively:** users may want to traverse snapshots,
+  restoring data to a writable volume more selectively instead of restoring
+  the whole snapshot.
+* **Volume backup:** users can't backup a live volume, they first need
+  to snapshot it. Once a snapshot is take, it still can't be backed-up,
```
humblec: is taken?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+  Users / tools may not be aware of this special directory, or it may not fit
+  their workflow. At the moment, the idiomatic way of accessing snapshot
+  contents in CSI drivers is by creating a new volume and populating it
+  with snapshot.
+
+## Design
+
+Key points:
+
+* Volume source is a snapshot, volume access mode is `*_READER_ONLY`.
```
humblec, line 58: This `*_READER ONLY` is bit complex or an area at kubernetes where things are vague or incomplete. However there are recent improvements on this , but yet to be completed. so I am not sure  is it a good idea to rely on READER ONLY access mode. In other side, considering this is a design proposal may be we are good and later while coming into implementation we can check thoroughly.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,274 @@
...
+* No actual new subvolumes are created in CephFS.
+* The resulting volume is a reference to the source subvolume snapshot.
+  This reference would be demarked in `Volume.volume_context` map.
```
humblec: can you please explain some more on , what is meant here by `demarked` ?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: Oops, that's a typo.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: I didn't want to pollute storage class parameters any more than necessary. I don't have a test implementation of this yet so I'm not sure 100% either about this. If you're ok with this for now, we can see during implementation, like you've suggested.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: I'll use a different word and clarify what the reference looks like.

gman0: https://github.com/ceph/ceph-csi/pull/2148/commits/b68dd5ebbfee3c3fe75a5341ce69b2decd3b0217

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: https://github.com/ceph/ceph-csi/pull/2148/commits/a98a5d30d69c2b25c80edfb413eeb5bd79c7af1a

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: https://github.com/ceph/ceph-csi/pull/2148/commits/b02a288ecad4694cfaf87d770f828c7c9191584a

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: https://github.com/ceph/ceph-csi/pull/2148/commits/4db796a7decf17050bfd66fb3ed9c54ebe0acb68

gman0:
I think I've addressed the comments we had so far, PTAL.

Madhu-1 (commented):
I don't have any major comments except what we already discussed. good to go for implementation we can update the design(minor changes) if required when we start the implementations

@gman0 please squash the commits and make the CI happy :)

gman0:
@Madhu-1 thanks! Commits squashed and all CI jobs have passed.

Madhu-1 (dismissed):

Madhu-1:
@humblec @nixpanic PTAL

haslersn:
> Later, I'd like to propose to Velero folks to include "volume from snapshot" as an intermediate step when performing volume backups using their [CSI plugin](https://github.com/vmware-tanzu/velero-plugin-for-csi).

@gman0 I think that proposal exists: https://github.com/vmware-tanzu/velero/issues/3544

gman0:
@haslersn great, I wasn't aware of this. Thank you for letting us know!

gman0:
@nixpanic @humblec @ShyamsundarR any updates on this proposal? I guess we can always update it later if we need to.

humblec:
> @nixpanic @humblec @ShyamsundarR any updates on this proposal? I guess we can always update it later if we need to.

@gman0  Apologies. Will take a final look at this tomorrow without fail..

gman0:
@humblec thanks! I'm now working on adding a couple of RADOS bindings to go-ceph, for atomic write ops, so that at least the reference counter itself is race-free.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
humblec: The problem is that, eventhough the request is for readonly, the node mount is "read write", are we fine with that?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,297 @@
...
+  with snapshot.
+
+## Design
+
+Key points:
+
+* Volume source is a snapshot, volume access mode is `*_READER_ONLY`.
+* No actual new subvolumes are created in CephFS.
+* The resulting volume is a reference to the source subvolume snapshot.
+  This reference would be stored in `Volume.volume_context` map. In order
```
humblec, line 61: this is specific to node stage request , isnt it ?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,297 @@
...
+  to reference a snapshot, we need subvol name and snapshot name.
+* Mounting such volume means mounting the respective CephFS subvolume
+  and exposing the snapshot to workloads.
```
humblec, line 64: just wondering what if the snapshot get deleted while the subvolume is in use or mounted. because kubelet or ocp does not know at this stage whether the inside snap is getting read by someone in this scenario. did I miss anything here ?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,297 @@
...
+
+As mentioned above, this is to protect shallow volumes, should their source
+snapshot be requested for deletion.
+
+When creating a volume snapshot, a new RADOS object would be created for that
+snapshot. It would contain a single signed integer value that would describe
+the number of references the backing subvolume snapshot has. It would be
+initialized to one, where the CSI snapshot itself is the first reference to the
+backing snapshot. Each subsequent shallow volume creation would increment the
+reference counter, and each shallow volume deletion would decrement it.
```
humblec: if we reference count only on `creation and deletion`, does that handle a situation of "multiple readers" of the same swallow volumes and thus consider deletion when all the references are gone?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: We don't care what users specify as access mode or if it's read-write. It will always be exposed as read-only to workloads. We can make stricter validation and refuse mounting volumes that want to be RW and have snapshot ref in their volume context parameters.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: Well any call that has `volume_context` in request, and when we need to distinguish between volume kinds. But yes, also when staging the volume.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: This is mentioned in [this](https://github.com/ceph/ceph-csi/blob/3edc507537d7ae12933ec750cb5efb6d14397c38/docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md#controller-operations) paragraph:

> When a shallow volume is created, what would happen if parent subvolume of the snapshot is removed while the shallow volume still exists?
>
> This shouldn't be a problem already. The parent volume has either `snapshot-retention` subvol feature in which case its snapshots remain available, or if it doesn't have that feature, it will fail to be deleted because it still has snapshots associated to it.

gman0: So either the snapshot will be kept, or the subvolume deletion will be refused by Ceph. In both cases we should be safe.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: I hope I understood the question correctly, but yes, it doesn't matter how many consumers of the shallow volume there are.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,297 @@
...
+  We need to make sure this doesn't happen and some book-keeping
+  is necessary. Ideally we could employ some kind of reference counting.
+
+#### Reference counting for shallow volumes
+
+As mentioned above, this is to protect shallow volumes, should their source
+snapshot be requested for deletion.
+
+When creating a volume snapshot, a new RADOS object would be created for that
+snapshot. It would contain a single signed integer value that would describe
```
gman0: Integer-based reference counter would break idempotency.

We have no way of distinguishing repeated calls on `CreateVolume`. We'd have duplicate counts. The same problem arises with *Delete ops.

Obviously we still need to track and count references, so this part of the proposal is ok I think, I should have left out the implementation details out though. We may need to store volume IDs (in an omap perhaps) and derive the number of references from that.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: With atomic reference counting / tracking, I think we no longer face the issues mentioned in this thread and in the paragraph regarding race conditions. It no longer matters how many readers/writers modify the references. It also means the requirement for single replica deployments can be lifted.

gman0:
I'll squash and rebase again. @Madhu-1 @humblec can you please take a look on the latest changes? I've reworked the "reference counter" into "reference tracker" - see the comment here https://github.com/ceph/ceph-csi/pull/2148#discussion_r726134474 and here https://github.com/ceph/ceph-csi/pull/2148#discussion_r728029855 .

gman0:
Just FYI, following with the latest findings, I've created a draft implementation of the reference tracker + accompanying program for testing -- albeit in C just for now. Once there's proper support from go-ceph, porting this to Go should be trivial.

https://github.com/gman0/rados-reference-tracker

This solves the problem with maintaining idempotency I mentioned in a comment above.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
```diff
@@ -0,0 +1,272 @@
...
+object holds no references. Note that this behavior would permit calling
+`DeleteSnapshot` even if it is still referenced by shallow volumes.
+
+* `DeleteSnapshot`:
+  * RT holds no references or the RT object doesn't exist:
+    delete the backing snapshot too.
+  * RT holds at least one reference: keep the backing snapshot.
+* `DeleteVolume`:
+  * RT holds no references: delete the backing snapshot too.
+  * RT holds at least one reference: keep the backing snapshot.
```
Madhu-1, line 111: @gman0 why RT is consider during CreateSnapshot and DeleteSnapshot it should be considered only during CreateVolume and DeleteVolume isn't it? if it's a new CreateVolume request increment the RT count and for each DeleteVolume decrement the RT if RT is reached `0` remove the snapshot?

gman0: @Madhu-1 this decouples shallow volumes from their source CSI snapshots - it makes it possible to call `DeleteSnapshot` and succeeding.

Doing what you suggested shouldn't break other aspects of this proposal, so it is possible, but at the cost of `DeleteSnapshot` calls being stuck until there are no shallow volumes associated with the snapshot. This may be an unexpected behavior to users, who would expect `DeleteSnapshot` to finish right away. When users create thick-provisioned volumes from snapshots (i.e. the non-shallow ones), they are able to delete the source snapshot afterwards and they see it succeeding, while the volumes exist. I think this behavior should be mirrored for shallow volumes too, for consistency.

This is possible only when we count references in `CreateSnapshot` and `DeleteSnapshot` procedures too.

Madhu-1: I haven't looked How RT will be implemented but if it's a rados object, I was thinking of the below case.

* when RT key does not exists: During DeleteSnapshot, In this case, return delete the snapshot
* when RT exists but not `0`: In this case DeleteSnapshot checks for RT if there any RT return success as RT was added in the CreateVolume, the DeleteVolume will take care of cleaning up the snapshot once the RT becomes `0`

Madhu-1: this is not a blocker we can take care of it during the implementation.

Madhu-1 (commented):
changes LGTM. once the commits are squashed will approve it.

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
gman0: @Madhu-1 ah I see. This is indeed more of an implementation detail of RT. We can discuss these when I'll push the ref-tracker PR.

gman0:
Thanks, @Madhu-1 ! Squashing changes.

Madhu-1 (approved):

gman0:
@humblec @nixpanic any more comments on this PR? Can we get this merged?

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
humblec: Correct, integer based reference counters are error prone. if the plan is to put it on volumeIDs or other unique entities, we are good @gman0

docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md
humblec: ok.. thanks for confirming

humblec:
> @humblec @nixpanic any more comments on this PR? Can we get this merged?

With the explanations so far , we are good.. More or less, while implementing some adjustments are expected any way .  so lets get this in and start working on the implementation .. Thanks @gman0 ..

humblec (approved):

gman0:
Thank you all for feedback! In order to implement this, it needs a couple of bindings added to go-ceph first -- for the reference tracker component. I'm currently working on them.

Rakshith-R:
> /retest all

Ci seems to be stuck https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/k8s-e2e-external-storage-1.21/detail/k8s-e2e-external-storage-1.21/1451/pipeline

Issue #2123, opened 2021-06-02
Locking in NodeUnpublishVolume significantly slows down pod deletion


mape90:
Case:
Application with multiple pods that have 20 PVs attached to it running on same node. There maybe only 20 PVs but all pods attach these volumes. (20 pods with 20 PVs on single node, all pods attach all 20 PVs) This Requires 400 mount binds and 20 actual cephfs mounts.

Issue:
Deletion of this application can take +15min because volumes are not unmounted and pods hang in Terminating state. Other CSI can handle this in some minutes.

Cause:
NodeUnpublishVolume acquires lock and returns error in case lock is held by another call. At mass deletion of pods (application unistal) NodeUnpublishVolume is called by kubelet ~400times almost immediately. This causes that most of the calls will fail as only first calls can acquire the lock and process the unmount. Others fail and cause back off delay to trigger. So first small patch of PVs are unattached. And all others fail the next round again only part of the volumes succeed and every time calls fail it will take longer and longer.

Actual NodeUnpublishVolume call when successful only takes 100-200ms. So call calls should complete in 1min

Possible solution:
Is the lock absolutely necessary for calling unmount? 
What makes it necessary to lock in NodeUnpublishVolume, as no other CSI implementation has lock. And they call same unmount and os.Remove. Also there is another lock for fuse that is this lock scope.

mape90:
Same issue could happen also for NodePublishVolume, however there seems that kubelet gets throttled by k8s API. This causes that NodePublishVolume are only called 1/sec. Increasing the qps will make the mounting go faster. So there the problem is on kubelet side and not on CSI.

humblec:
> Same issue could happen also for NodePublishVolume, however there seems that kubelet gets throttled by k8s API. This causes that NodePublishVolume are only called 1/sec. Increasing the qps will make the mounting go faster. So there the problem is on kubelet side and not on CSI.

Thanks for the detailed analysis . At umount/unpublish, we take lock ( https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/nodeserver.go#L576 )  to make sure or check any other ongoing operation exist on the volume. But considering the operation of unmount should not cause any harmful operation to the volume itself , I think we can live without locking. But allow some time, we have to think through this .

mape90:
Hi,

[...]

PR #2149, opened 2021-06-10
Get rid of locking at node {publish,unpublish} operations

humblec:
Considering kubelet make sure node operations like unpublish/nodeunpublish
..etc can **not** be called at same time, we dont require extra locking
in nodePublish and nodeUnpublish operation from driver side. Till this commit, we were
operating at NodePublish/NodeUnpublish with a lock and it is now removed.

**RBD**:

- [x] Lock removed from nodePublish
- [x] Lock removed from nodeUnpublish

**CephFS**

- [x] Lock removed from nodePublish
- [x] Lock removed from nodeUnpublish

Fixes: #https://github.com/ceph/ceph-csi/issues/2123
Updates: https://github.com/ceph/ceph-csi/issues/2197

Thanks to @mape90 for reporting and analyzing this issue

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -267,16 +267,9 @@ func (ns *NodeServer) NodeUnpublishVolume(
 	if err = util.ValidateNodeUnpublishVolumeRequest(req); err != nil {
 		return nil, err
 	}
-
-	volID := req.GetVolumeId()
+	// considering kubelet make sure node operations like unpublish/unstage...etc can not be called
+	// at same time, an explicit locking at time of nodeunpublish is not required.
 	targetPath := req.GetTargetPath()
-
-	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
-		util.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volID)
-		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volID)
-	}
-	defer ns.VolumeLocks.Release(volID)
-
 	isMnt, err := util.IsMountPoint(targetPath)
 	if err != nil {
 		if os.IsNotExist(err) {
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -217,11 +217,8 @@ func (ns *NodeServer) NodePublishVolume(
 	targetPath := req.GetTargetPath()
 	volID := req.GetVolumeId()
 
-	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
-		util.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volID)
-		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volID)
-	}
-	defer ns.VolumeLocks.Release(volID)
+	// Considering kubelet make sure the stage and publish operations
+	// are serialized, we dont need any extra locking in nodePublish
 
 	if err := util.CreateMountPoint(targetPath); err != nil {
 		util.ErrorLog(ctx, "failed to create mount point at %s: %v", targetPath, err)
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -611,14 +611,8 @@ func (ns *NodeServer) NodeUnpublishVolume(
 	}
 
 	targetPath := req.GetTargetPath()
-	volID := req.GetVolumeId()
-
-	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
-		util.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volID)
-		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volID)
-	}
-	defer ns.VolumeLocks.Release(volID)
-
+	// considering kubelet make sure node operations like unpublish/unstage...etc can not be called
+	// at same time, an explicit locking at time of nodeunpublish is not required.
 	notMnt, err := mount.IsNotMountPoint(ns.mounter, targetPath)
 	if err != nil {
 		if os.IsNotExist(err) {
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -439,11 +439,8 @@ func (ns *NodeServer) NodePublishVolume(
 	volID := req.GetVolumeId()
 	stagingPath += "/" + volID
 
-	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
-		util.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volID)
-		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volID)
-	}
-	defer ns.VolumeLocks.Release(volID)
+	// Considering kubelet make sure the stage and publish operations
+	// are serialized, we dont need any extra locking in nodePublish
 
 	// Check if that target path exists properly
 	notMnt, err := ns.createTargetMountPath(ctx, targetPath, isBlock)


humblec:
@Madhu-1 , fyi:
```
Step 1/6 : FROM registry.fedoraproject.org/fedora:latest
Get https://registry.fedoraproject.org/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
make: *** [Makefile:207: .test-container-id] Error 1
Error: Process completed with exit code 2.
```

Yuggupta27:
> /retest all

Currently CI is down, so centos-ci tests will not get triggered :(

internal/cephfs/nodeserver.go
```diff
@@ -254,16 +254,11 @@ func (ns *NodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	if err = util.ValidateNodeUnpublishVolumeRequest(req); err != nil {
 		return nil, err
 	}
-
-	volID := req.GetVolumeId()
+	// notice: we have been operating here with a lock while unmounting the volume
+	// which slowdown the unmount operation when there are more pods and delete performed
+	// at almost same time. Considering we have enough safe-guards on CO side, we can
+	// operate outside the critical section.
```
nixpanic: This comment is not needed, and  probably more confusing than it can be. No need to mention issues (slowdown), just explain *why* it is safe to not use locking here.

It would be better to explain the working of NodeUnpublishVolume in the comment above the function, and maybe mention that locking is not required (why?) for unmounting.

nixpanic (commented):
Nits:

- "Getrid" in the subject of the PR is not a word
- `Fix: #...` is not a GitHub annotation in commit messages, use `Fixes: #...`
- Both commit messages should have a description, not only one. You can use the same description for both if you like.

Madhu-1:
@humblec as we are depending on the CO for safeguarding then we don't need locking at the nodepublish also?

nixpanic:
On Thu, Jun 10, 2021 at 12:24:50AM -0700, Madhu Rajanna wrote:
> @humblec as we are depending on the CO for safeguarding then we don't need locking at the nodepublish also?

Agreed! NodeStageVolume should only need this for non-RWX volumes?

Madhu-1:
> On Thu, Jun 10, 2021 at 12:24:50AM -0700, Madhu Rajanna wrote: @humblec as we are depending on the CO for safeguarding then we don't need locking at the nodepublish also?
> Agreed! NodeStageVolume should only need this for non-RWX volumes?

Yes correct, If CO takes care of it then we don't need it in NodeStage and Unstage also.

internal/cephfs/nodeserver.go
humblec: sure... let me address this in a detail or change the comment with the sequence ..etc

humblec:
@nixpanic @Madhu-1 there could be various locks we can get rid of or adjust. The #2197 list out the operations. I would like to keep this one at a time if possible, especially when there is a mix of 'global' and 'local' operations like node*Publish /Node*stage. It also help to seggregate or take an action properly.  NodeUnpublish case is handled here, and followup PRs will be filed accordingly. PTAL.. Thanks !

yati1998 (approved):
LGTM

Madhu-1:
IMO the issue in this area cannot be caught easily. @humblec with how many pods and volumes this has been tested and did you see any warning/error logs in the kubelet after scaling up/down application pods several times?

humblec:
The general test we have with our `e2e` test kind of prove that, there is no `lockup` we come across at least on current tests. Additionally, I have been trying to make use of our scale test CI to run it on top of this PR, however not much luck, the main reason for that is, the CI test suite we have is tightly coupled to `ODF` or `downstream`. Planning to try some more to make it work so that we are sure, there is no 'regression' and also we can do some benchmark on the performance improvement we get with this change.  I will update here how it goes.

humblec:
[Update] The test results are good at my end. However trying to make use of scale testing by taking help from Perf team to really validate the improvement this brings us. Will update here how it goes.

humblec:
FYI, The global mount operations locking has been taken out and the test image is `quay.io/humble/cephcsi:publish` which has been undergoing the testing for a while now.

humblec:
@ekuric would you like to update  your findings as you are helping us with scale testing to verify the improvement with this fix image?

ekuric:
> @ekuric would you like to update your findings as you are helping us with scale testing to verify the improvement with this fix image?

I tested images you provided and I can see below observations

- Create time: It takes approximately 420s ( ~7m ) to start 1000 pods with PVC attached ( from cephfs or ceph rbd storage class ).
-  Delete time: Pods created in previous steps were in project - eg. projectX. For delete tests we ran `oc delete project projectX ` and measured time how long it takes before all devices disappear from storage backend  ( for ceph rbd `rbd ls storagecluster-cephblockpool | wc -l` and for cephfs `ceph fs subvolume ls storagecluster-cephfilesystem csi |grep name | wc -l` ). In both these cases it took between 280 - 300 seconds before all volumes are cleaned on storage side.
- During tests , there were not noticed issues with ceph cluster itself , it was operating without  ( noticeable )  issues. We can for purpose of next release run more extensive tests if necessary.

humblec:
> > @ekuric would you like to update your findings as you are helping us with scale testing to verify the improvement with this fix image?
> 
> I tested images you provided and I can see below observations
> 
>     * Create time: It takes approximately 420s ( ~7m ) to start 1000 pods with PVC attached ( from cephfs or ceph rbd storage class ).
> 
>     * Delete time: Pods created in previous steps were in project - eg. projectX. For delete tests we ran `oc delete project projectX ` and measured time how long it takes before all devices disappear from storage backend  ( for ceph rbd `rbd ls storagecluster-cephblockpool | wc -l` and for cephfs `ceph fs subvolume ls storagecluster-cephfilesystem csi |grep name | wc -l` ). In both these cases it took between 280 - 300 seconds before all volumes are cleaned on storage side.
> 
>     * During tests , there were not noticed issues with ceph cluster itself , it was operating without  ( noticeable )  issues. We can for purpose of next release run more extensive tests if necessary.

This is awesome!!!  :tada: Thanks @ekuric for doing this, much appreciated.  

In short, `1000 pods` delete time has been improved from `980seconds` to `~280` seconds ! Thats huge improvement.  :+1:

humblec:
@Madhu-1 @nixpanic considering we have a good improvement here with some test results, can we get this in ?

internal/rbd/nodeserver.go
```diff
@@ -145,11 +145,8 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 	defer cr.DeleteCredentials()
 
-	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
-		util.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volID)
-		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volID)
-	}
-	defer ns.VolumeLocks.Release(volID)
+	// considering kubelet make sure node operations like publish/stage...etc can not be called
```
Madhu-1: @humblec are we sure that kubelet is taking some central locking when it comes to RWO on multiple nodes. I past we have seen that RWO request comes up on two nodes parallel and if this happens we end up corrupting the data?

Madhu-1 (commented):
Nothing can be caught in review for this kind of PR, Hope everything is tested intensively and everything is verified from kubernetes related to stale mounts, directories, data corruption, etc, and also the node lost scenarios is also tested.

internal/rbd/nodeserver.go
Madhu-1: Discard this one. I was confused with watchers

internal/rbd/nodeserver.go
```diff
@@ -145,11 +145,8 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 	defer cr.DeleteCredentials()
 
-	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
```
pkalever: As its already discussed, I'm sure we all are aware of this, but just to remind again:

Locking is needed by Volume Healer, so at least for NodeStageVolume() we cannot remove it.
See: https://github.com/ceph/ceph-csi/pull/2108

pkalever (changes requested):

internal/rbd/nodeserver.go
humblec: yeah, we are aware. One thing which I can do here is bringing back the nodestage and nodeunstage locks back.. let me refresh the PR with that.

humblec:
> Nothing can be caught in review for this kind of PR, Hope everything is tested intensively and everything is verified from kubernetes related to stale mounts, directories, data corruption, etc, and also the node lost scenarios is also tested.

@Madhu-1  @nixpanic  Sure, btw one change I have done with last refresh is that, brought back the locking in 'global' operations which could help rbd nbd driver:

The delete performance (https://github.com/ceph/ceph-csi/pull/2149#issuecomment-872922124) has been gained even with this version of the patch. so we are good. 

Multilple PVC creation/deletion , App creation, deletion in batches ....etc are performed with this version of the patch and it looks good too, apart from all the e2e tests which has been passed. 

So, requesting to get this in at this earliest , so that more testing can go through with different test suites ( downstream) too.  Thanks :+1:

Madhu-1 (dismissed):
one thing to consider here is also the stale mount/unmount commands for the same volume when we run the commands in parallel. I hope that is also tested and for all the cases we have not seen any issues on the kubelet logs.

humblec:
``` 
Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute

```

nixpanic (dismissed):
@mergifyio rebase

pkalever (approved):
See if there is a need to edit the commit and PR messages. Just remove references about NodeStage and NodeUnstage.

"we were operating at NodePublish/NodeUnpublish/NodeUnstage and NodeStage with a lock and it is now removed."

Rest all looks good to me.

nixpanic:

Failed with #2264 ([logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.20/detail/mini-e2e_k8s-1.20/1659/pipeline)).

Madhu-1 (approved):
As discussed with @humblec we don't have any problem even if we remove the lock and the mount hangs(or takes more times) and kubelet sends `n` number of requests due to timeout.

Madhu-1:
@mergifyio rebase

PR #1974, opened 2021-04-12
rebase: update kubernetes to v1.21.2

Rakshith-R:
# Describe what this PR does #
Updated kubernetes packages to latest release.
- resizefs package has been included into k8s.io/mount-util package. updated code to use the same.
- update controller-runtime package. k8s v1.21.2 needs this update to run. Updated code persistentvolume.go Reconcile() and added context.Context accordingly.
- updated controller-runtime to v9.2.0
- e2e: update e2epv.WaitOnPVandPVC() to match new parameters 

fixes: #1968 

~~Dependency: Need controller-runtime v0.9.0 release https://github.com/kubernetes-sigs/controller-runtime/releases 
(Testing with v0.9.0-alpha.1 )~~

TODO: 

- [x] update controller-runtime package to v0.9.0 ~~(not released yet)~~
 
---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)
- `/retest all`: run this in case the CentOS CI failed to start/report any test
  progress or results

</details>

PR #2153, opened 2021-06-11
various corrections in e2e code

humblec:
This PR covers below items in e2e code

correct createORDeleteCephfsResources() function name for both cephfs and rbd
e2e: correct gosec marker for credentials rule
e2e: use proper variable name for rbdmountoptions 

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -35,7 +35,7 @@ const (
 	vaultPassphrasePath = "ceph-csi/"
 
 	rookToolBoxPodLabel = "app=rook-ceph-tools"
-	rbdmountOptions     = "mountOptions"
+	rbdMountOptions     = "mountOptions"
 
 	retainPolicy = v1.PersistentVolumeReclaimRetain
 	// deletePolicy is the default policy in E2E.
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -24,7 +24,7 @@ import (
 	e2elog "k8s.io/kubernetes/test/e2e/framework/log"
 )
 
-/* #nosec:G101, values not credententials, just a reference to the location.*/
+/* #nosec:G101, values not credentials, just a reference to the location.*/
 const (
 	defaultNs     = "default"
 	defaultSCName = ""
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -55,17 +55,17 @@ func deployCephfsPlugin() {
 		e2elog.Failf("failed to delete nodeplugin rbac %s with error %v", cephfsDirPath+cephfsNodePluginRBAC, err)
 	}
 
-	createORDeleteCephfsResouces("create")
+	createORDeleteCephfsResources("create")
 }
 
 func deleteCephfsPlugin() {
-	createORDeleteCephfsResouces("delete")
+	createORDeleteCephfsResources("delete")
 }
 
-func createORDeleteCephfsResouces(action string) {
+func createORDeleteCephfsResources(action string) {
 	csiDriver, err := ioutil.ReadFile(cephfsDirPath + csiDriverObject)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


Yuggupta27 (approved):
LGTM

e2e/utils.go
```diff
@@ -35,7 +35,7 @@ const (
 	vaultPassphrasePath = "ceph-csi/"
 
 	rookToolBoxPodLabel = "app=rook-ceph-tools"
-	rbdmountOptions     = "mountOptions"
+	rbdMountOptions    = "mountOptions"
```
Yuggupta27:
```suggestion
	rbdMountOptions     = "mountOptions"
```

Yuggupta27: To satisfy gofmt, try aligning it with the above assignment

humblec: it was done while you commenting.. thanks :)

nixpanic (dismissed):
LGTM, but I wonder why codespell did not find the issues. Could you check if the github action can be corrected so that it catches this the next time?

humblec:
> LGTM, but I wonder why codespell did not find the issues. Could you check if the github action can be corrected so that it catches this the next time?

@nixpanic actually the `kubernetes` spell was pointed out by the codespell in github action in  this run, so I think its catching it. Not sure why it was not the case in the previous runs. Have to check, however do we need to hold merging  this PR while checking on the same ?

nixpanic:
> @nixpanic actually the `kubernetes` spell was pointed out by the codespell in github action in this run, so I think its catching it. Not sure why it was not the case in the previous runs. Have to check, however do we need to hold merging this PR while checking on the same ?

None of the other PRs seem to have any errors or warnings in the codespell output. As you're using it to find issues, you seem to be the right person to fix the GitHub action too. Doing so in this PR makes a lot of sense to me, you have the command you used in one of the commit messages.

In case it is really too complex or time consuming to find out why codespell as GitHub action did not caught this, at the very least an issue should be reported about it. It may help to run the GitHub action with some `--verbose` or increased log-level option to aid the investigation.

Rakshith-R:
@nixpanic @humblec , `Error: ./internal/util/aws_metadata.go:96: Kubenetes ==> Kubernetes` was also found in https://github.com/ceph/ceph-csi/pull/1974/checks?check_run_id=2800610364

We're using https://github.com/ceph/ceph-csi/blob/33bf3fae6b38b7aa5ff00a6201b9749f5207a76e/.github/workflows/codespell.yaml#L17
~~i think that is why the codespell is catching this error now.~~
On second thought, the repository hasn't been updated in few months.

Rakshith-R:
okay, I think i found the cause the codespell action uses latest version of https://github.com/codespell-project/codespell (which got bumped to v2.1.0 13 hrs ago)
deduced from https://github.com/codespell-project/actions-codespell/blob/de089481bd65b71b4d02e34ffb3566b6d189333e/requirements.txt#L1. 
So I guess this is to be expected.

nixpanic (approved):
Thanks for looking into this @Rakshith-R !

humblec:
> > @nixpanic actually the `kubernetes` spell was pointed out by the codespell in github action in this run, so I think its catching it. Not sure why it was not the case in the previous runs. Have to check, however do we need to hold merging this PR while checking on the same ?
> 
> None of the other PRs seem to have any errors or warnings in the codespell output. As you're using it to find issues, you seem to be the right person to fix the GitHub action too. Doing so in this PR makes a lot of sense to me, you have the command you used in one of the commit messages.


@nixpanic I think there was a confusion, I didnt find it from my end, the CI caught it on 'this PR run' as mentioned in the previous comment.  Any way, it looks like the codespell we have with github action catching it efficiently compared to previous instances of old version runs of the same :+1:

Madhu-1:
@merigfyio backport release-v3.3

Madhu-1:
@mergifyio backport release-v3.3

Issue #1586, opened 2020-10-15
update all the static check tools to latest version


Madhu-1:
I was going through the golangci-lint new release, looks like we have some static issues its good to run all static tools with latest versions

```go
make go-lint 
cephcsi image settings: quay.io/cephcsi/cephcsi version canary
./scripts/lint-go.sh
INFO [config_reader] Used config file scripts/golangci.yml 
INFO [lintersdb] Active 46 linters: [asciicheck bodyclose deadcode depguard dogsled errcheck exhaustive exportloopref gci gocognit goconst gocritic gocyclo godot gofmt gofumpt goheader goimports golint gomnd gomodguard goprintffuncname gosec gosimple govet ineffassign interfacer lll maligned misspell nakedret nlreturn noctx nolintlint rowserrcheck scopelint sqlclosecheck staticcheck structcheck stylecheck typecheck unconvert unparam unused varcheck whitespace] 
INFO [loader] Using build tags: [nautilus]        
INFO [loader] Go packages loading at mode 575 (compiled_files|exports_file|imports|name|deps|files|types_sizes) took 1.362612936s 
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 21.267923ms 
INFO [linters context/goanalysis] analyzers took 2m30.726592724s with top 10 stages: buildir: 2m1.340534731s, inspect: 3.066803154s, buildssa: 2.741827785s, the_only_name: 2.046159586s, exhaustive: 1.9981591s, gocritic: 1.819492945s, goimports: 1.801243659s, fact_deprecated: 1.495034764s, ctrlflow: 1.397161224s, gofumpt: 1.390896208s 
INFO [linters context/goanalysis] analyzers took 2.318403989s with top 10 stages: buildir: 2.070186758s, U1000: 248.217231ms 
WARN [runner/nolint] Found unknown linters in //nolint directives: gocognit:gocyclo 
INFO [runner/max_same_issues] 508/511 issues with text "return with no blank line before" were hidden, use --max-same-issues 
INFO [runner/max_same_issues] 52/55 issues with text "File is not `gofumpt`-ed" were hidden, use --max-same-issues 
INFO [runner/max_same_issues] 47/50 issues with text "File is not `gci`-ed with -local github.com/ceph/ceph-csi" were hidden, use --max-same-issues 
INFO [runner/max_same_issues] 38/41 issues with text "Top level comment should end in a period" were hidden, use --max-same-issues 
[...]

PR #2155, opened 2021-06-11
cleanup: gocyclo being unused for linter

yati1998:
This commit addresses the following issue:
'nolint:gocyclo // complexity needs to be reduced.'
is unused for linter "gocyclo" (nolintlint)

Updates:#1586

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -591,8 +591,7 @@ func checkContentSource(ctx context.Context, req *csi.CreateVolumeRequest, cr *u
 
 // DeleteVolume deletes the volume in backend and removes the volume metadata
 // from store
-// TODO: make this function less complex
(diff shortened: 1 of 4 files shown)


internal/rbd/controllerserver.go
```diff
@@ -592,7 +592,6 @@ func checkContentSource(ctx context.Context, req *csi.CreateVolumeRequest, cr *u
 // DeleteVolume deletes the volume in backend and removes the volume metadata
 // from store
 // TODO: make this function less complex
```
nixpanic: Dropping the next line causes golint to fail with:
```
Top level comment should end in a period (godot)
```
That means, the previous line should end in a `.`:
```
// TODO: make this function less complex.
```

Need to adjust that here, and on other places.

nixpanic (approved):

Madhu-1 (approved):

PR #2160, opened 2021-06-14
util: returns actual error instead of ErrPoolNotFound

yati1998:
This commit returns actual error returned by the go-ceph API
to the function GetPoolName(..) instead of just returning
ErrPoolNotFound everytime there is an error getting the pool id.
There is a issue reported in which the snapshot creation
takes much more time to reach True state
(i.e., between 2-7 mins) and keeps trying to create with
below error though pool is present:
rpc error: code = NotFound desc = pool not found: pool ID (21)
not found in Ceph cluster.

Since we cannot interpret the actual error for the delay in
snapshot creation, it is required to return the actual error
as well so that we can uderstand the reason.

### DIFF
--- a/internal/util/cephcmds.go
+++ b/internal/util/cephcmds.go
@@ -92,9 +92,11 @@ func GetPoolName(monitors string, cr *Credentials, poolID int64) (string, error)
 	defer connPool.Put(conn)
 
 	name, err := conn.GetPoolByID(poolID)
-	if err != nil {
-		return "", fmt.Errorf("%w: pool ID (%d) not found in Ceph cluster",
+	if errors.Is(err, rados.ErrNotFound) {
+		return "", fmt.Errorf("%w: pool ID(%d) not found in Ceph cluster",
 			ErrPoolNotFound, poolID)
+	} else if err != nil {
+		return "", fmt.Errorf("failed to get pool ID %d: %w", poolID, err)
 	}
 	return name, nil
 }


yati1998:
@nixpanic can you please review this PR, as it is a blocker for 4.8.

Rakshith-R:
@yati1998 how about following similar pattern to the following, 
https://github.com/ceph/ceph-csi/blob/5b7b5f1e3af5a6da0f829e4b7d76f0e369c4a16a/internal/util/cephcmds.go#L74-L80

Can avoid the confusion about `pool not found` error this way.

yati1998:
> @yati1998 how about following similar pattern to the following,
> https://github.com/ceph/ceph-csi/blob/5b7b5f1e3af5a6da0f829e4b7d76f0e369c4a16a/internal/util/cephcmds.go#L74-L80
> 
> Can avoid the confusion about `pool not found` error this way.

true that.

internal/util/cephcmds.go
```diff
@@ -92,9 +92,11 @@ func GetPoolName(monitors string, cr *Credentials, poolID int64) (string, error)
 	defer connPool.Put(conn)
 
 	name, err := conn.GetPoolByID(poolID)
-	if err != nil {
-		return "", fmt.Errorf("%w: pool ID (%d) not found in Ceph cluster",
+	if errors.Is(err, rados.ErrNotFound) {
+		return "", fmt.Errorf("%w: pool (%d) not found in Ceph cluster",
```
Rakshith-R:
```suggestion
		return "", fmt.Errorf("%w: pool ID (%d) not found in Ceph cluster",
```
Keep the `ID` part?

internal/util/cephcmds.go
```diff
@@ -92,9 +92,11 @@ func GetPoolName(monitors string, cr *Credentials, poolID int64) (string, error)
 	defer connPool.Put(conn)
 
 	name, err := conn.GetPoolByID(poolID)
-	if err != nil {
-		return "", fmt.Errorf("%w: pool ID (%d) not found in Ceph cluster",
+	if errors.Is(err, rados.ErrNotFound) {
+		return "", fmt.Errorf("%w: pool ID(%d) not found in Ceph cluster",
 			ErrPoolNotFound, poolID)
+	} else if err != nil {
+		return "", fmt.Errorf("failed to get pool ID: %w", err)
```
nixpanic: Include the pool id in this message too please. It may help while troubleshooting issues.

nixpanic (changes requested):

nixpanic (approved):

Rakshith-R (approved):

agarwal-mudit (approved):

PR #2174, opened 2021-06-17
rbd: add validation for thick restore/clone

### LINKED ISSUE #2176: Only allow Restore/Clone from thick to thick
Allow creation of Clone from thick PVC to thick PVC and Restore of a snapshot of a thick PVC to a thick PVC. this will keep the implementation simple.

 In the future, we can allow thick-to-thin and thin-to-thick operations. will open an issue to allow this functionality(we can consider this in upcoming releases)

Madhu-1:
added validation to allow only Restore of Thick PVC snapshot to a thick clone and creation of a thick clone from thick PVC.
    
### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -222,8 +222,18 @@ func checkValidCreateVolumeRequest(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSna
 		if err != nil {
 			return status.Errorf(codes.InvalidArgument, "cannot restore from snapshot %s: %s", rbdSnap, err.Error())
 		}
+
+		err = rbdSnap.isCompatibleThickProvision(rbdVol)
+		if err != nil {
+			return status.Errorf(codes.InvalidArgument, "cannot restore from snapshot %s: %s", rbdSnap, err.Error())
+		}
 	case parentVol != nil:
 		err = parentVol.isCompatibleEncryption(&rbdVol.rbdImage)
+		if err != nil {
+			return status.Errorf(codes.InvalidArgument, "cannot clone from volume %s: %s", parentVol, err.Error())
+		}
+
+		err = parentVol.isCompatibleThickProvision(rbdVol)
 		if err != nil {
 			return status.Errorf(codes.InvalidArgument, "cannot clone from volume %s: %s", parentVol, err.Error())
 		}
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -895,6 +895,30 @@ func cloneFromSnapshot(ctx context.Context, rbdVol *rbdVolume, rbdSnap *rbdSnaps
 		}
 	}
 
+	// The clone image created during CreateSnapshot has to be marked as thick.
+	// As snapshot and volume both are independent we cannot depend on the
+	// parent volume of the clone to check thick provision during CreateVolume
+	// from snapshot operation because the parent volume can be deleted anytime
+	// after snapshot is created.
+	// TODO: copy thick provision config
+	thick, err := rbdVol.isThickProvisioned()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed checking thick-provisioning of %q: %s", rbdVol, err)
+	}
+
+	if thick {
+		// check the thick metadata is already set on the clone image.
+		thick, err = vol.isThickProvisioned()
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed checking thick-provisioning of %q: %s", vol, err)
+		}
+		if !thick {
+			err = vol.setThickProvisioned()
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "failed mark %q thick-provisioned: %s", vol, err)
+			}
+		}
+	}
 	// if flattening is not needed, the snapshot is ready for use
 	readyToUse := true
 
@@ -983,6 +1007,24 @@ func (cs *ControllerServer) doSnapshotClone(ctx context.Context, parentVol *rbdV
 				"config for %q: %v", cloneRbd, cryptErr)
 			return ready, nil, status.Errorf(codes.Internal,
 				err.Error())
+		}
+	}
+
+	// The clone image created during CreateSnapshot has to be marked as thick.
+	// As snapshot and volume both are independent we cannot depend on the
+	// parent volume of the clone to check thick provision during CreateVolume
+	// from snapshot operation because the parent volume can be deleted anytime
+	// after snapshot is created.
+	// TODO: copy thick provision config
+	thick, err := parentVol.isThickProvisioned()
+	if err != nil {
+		return ready, nil, status.Errorf(codes.Internal, "failed checking thick-provisioning of %q: %s", parentVol, err)
+	}
+
+	if thick {
+		err = cloneRbd.setThickProvisioned()
+		if err != nil {
+			return ready, nil, status.Errorf(codes.Internal, "failed mark %q thick-provisioned: %s", cloneRbd, err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


nixpanic:
In theory we can also call [`rbd_sparsify()`](https://pkg.go.dev/github.com/ceph/go-ceph/rbd#Image.Sparsify) on the new image if it should not be thick-provisioned.

Maybe there should be an issue to track what we actually want to do, and why?

internal/rbd/rbd_util.go
```diff
@@ -1570,3 +1570,19 @@ func (ri *rbdImage) isCompatibleEncryption(dst *rbdImage) error {
...
 }
+
+func (ri *rbdImage) isCompatibleThickProvision(dst *rbdVolume) error {
+	thick, err := ri.isThickProvisioned()
+	if err != nil {
+		return err
+	}
+	switch {
+	case thick && !dst.ThickProvision:
+		return fmt.Errorf("cannot create thin volume from thick volume %q", ri)
```
nixpanic, line 1624: The error message mentions creating, but that may not always be correct. use "does not match" or similar instead?

Madhu-1:
> In theory we can also call [`rbd_sparsify()`](https://pkg.go.dev/github.com/ceph/go-ceph/rbd#Image.Sparsify) on the new image if it should not be thick-provisioned.
> 
> Maybe there should be an issue to track what we actually want to do, and why?

To keep the feature simple for now let's allow only thick-thick will open one for this one and will open another issue to track thick-thin and thin-thick feature

internal/rbd/rbd_util.go
Madhu-1: keeping error message same as https://github.com/ceph/ceph-csi/pull/2175 any suggestion on the error message?

nixpanic: I do not care too much about it :) for now it is only used while creating volumes, so that's fine

humblec: I think current message is fine: we can also change to  something like this:
`Source ( thin/thick) : Destination ( thin/thick) - Unsupported conversion`:

and fill `thin` and `thick` strings accordingly in these cases

Madhu-1: Thanks, @humblec as this is a clear error message and fine let's keep it as it is for now. If required we can change it later.

nixpanic (dismissed):

humblec:
@Madhu-1 please let me know your choice to accept or defer https://github.com/ceph/ceph-csi/pull/2174#discussion_r653297992, accordingly we can move :)

Madhu-1:
> @Madhu-1 please let me know your choice to accept or defer [#2174 (comment)](https://github.com/ceph/ceph-csi/pull/2174#discussion_r653297992), accordingly we can move :)

Already replied at https://github.com/ceph/ceph-csi/pull/2174#discussion_r653298945 :)

humblec (dismissed):

Madhu-1:
@nixpanic @humblec added one more commit to fix some cases where parent PVC of the snapshot is created before Restore operation. PTAL

Madhu-1:

` Error from server: error when creating "STDIN": etcdserver: request timed out`

internal/rbd/controllerserver.go
```diff
@@ -218,15 +218,23 @@ func checkValidCreateVolumeRequest(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSna
 
 	switch {
 	case rbdSnap != nil:
+
```
nixpanic: drop this new line?

internal/rbd/controllerserver.go
```diff
@@ -218,15 +218,23 @@ func checkValidCreateVolumeRequest(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSna
...
 		err = rbdSnap.isCompatibleEncryption(&rbdVol.rbdImage)
 		if err != nil {
 			return status.Errorf(codes.InvalidArgument, "cannot restore from snapshot %s: %s", rbdSnap, err.Error())
 		}
+		// During CreateSnapshot the rbd image will be created with the
```
nixpanic: wow, this is really ugly :cry:

Unfortunatelty I think I understand why this is needed. However, I prefer to have a `isCompatibleThickProvision()` function for the `rbdSnapshot` type instead. That can use the `RbdSnapName` for checking the thick-provisioning state.

(temporary) changing attributes of objects should not be done, it shows lack of proper usage/design of the type in many cases. The `rbdSnapshot` type should probably get corrected, and once that is done, the duplicate `isCompatibleThickProvision()` function can be removed/reconciled.

nixpanic (changes requested):

internal/rbd/controllerserver.go
Madhu-1: Pushed a new commit with a better approach instead of changing attributes on the snapshot we can generate a new temp volume and use it for checking the metadata and we can destroy it once work is done.

internal/rbd/controllerserver.go
```diff
@@ -860,6 +870,30 @@ func cloneFromSnapshot(ctx context.Context, rbdVol *rbdVolume, rbdSnap *rbdSnaps
 		}
 	}
 
+	// The clone image created during CreateSnapshot has to be marked as thick.
+	// As snapshot and volume both are independent we cannot depend on the
+	// parent volume of the clone to check thick provision during CreateVolume
+	// From snapshot operation because the parent volume can be deleted anytime
```
humblec: s/From/from

internal/rbd/controllerserver.go
```diff
@@ -951,6 +985,24 @@ func (cs *ControllerServer) doSnapshotClone(ctx context.Context, parentVol *rbdV
...
```
humblec: s/From/from

internal/rbd/rbd_util.go
```diff
@@ -1570,3 +1570,47 @@ func (ri *rbdImage) isCompatibleEncryption(dst *rbdImage) error {
...
+
+	return nil
+}
+
+// FIXME: merge isCompatibleThickProvision of rbdSnapshot and rbdImage to a single
+// function.
+func (rs *rbdSnapshot) isCompatibleThickProvision(dst *rbdVolume) error {
+	// During CreateSnapshot the rbd image will be created with the
+	// snapshot name. Replacing RbdImageName with RbdSnapName so that we
+	// can check if the image is t	hick provisioned
```
humblec: typo in thick above.

internal/rbd/controllerserver.go
humblec: @Madhu-1 I am not favoring  the latest approach of  an intermediate volume. The reason being an image struct creation in the backend just to hide our code issue :) ..

Madhu-1: @humblec please let me know your suggestion i can do the same.

Madhu-1: we are not creating an image in the backend it's just a generation of the new temp volume structure.

humblec: yes, its just a generation of the new struct, but previously we were flipping a field in the struct, now we are completely creating a new structure.. I am wondering how this favors the other :).. anyway lets get one more opinion on this and decide.

Madhu-1: This approach covers the original concern without modifying the original rbdSnapshot struct and reverts it back. the current temp volume generation is already used in other places, not something new.

nixpanic (approved):

internal/rbd/controllerserver.go
humblec: yep, the entire code path need to be revisited in some time. However considering the priority lets survive with this model. Thanks

humblec (approved):

Madhu-1:

`Error from server: error when creating "STDIN": etcdserver: request timed out`

Madhu-1:

`failed to create PVC with error etcdserver: request timed out`

Issue #2178, opened 2021-06-17
Enhance/Refactor E2E tests 


humblec:
# Describe the feature you'd like to have #

Looking at the E2E, our deployment model for E2E tests is bit clumsy,  Repeated code and the sequential lay out of the code make very difficult to folllow and also make it hard to add more deployment logic in future. This is an enhancement request to make it more modular and clear.

Items planned to address:

- [x] Improve RBD deployment model and make it modular #2180
- [x] Improve CephFS deployment model and make it modular #2180 
- [ ] Improve Vault deployment
- [ ] Make use of wrappers if possible in dpeloyment/test methods
- [ ] Make use of `storage/utils` framework methods for various operations which are kind of readymade for storage operations
- [ ] Try to make use of `LoadManifests()`, `createManifests()`, `PathItems()` methods instad of our own functions.
- [ ] Change object recreation logic in each test for StroageClass
- [ ] Modularize snapshot test cases 
- [ ] revisit each tests and improve the execution time and code modularity
- [x] Get rid of `waitGroup` passing between operations #2215 
- [ ] Dont block on `delete` operations in sender.
  For example: 
  
```
for i := 0; i < totalCount; i++ {
	name := fmt.Sprintf("%s%d", f.UniqueName, i)
	err = deletePVCAndApp(name, f, pvc, app)
	if err != nil {
			e2elog.Failf("failed to delete PVC or application with error %v", err)
	}
}
```

The waitTime goes `totalCount*` * `deploytimeout` in case there is an error in helper func, Instead of this spawn go routines and when we `receive` errors ( in some cases in first error itself)  in a channel , fail//exit.
- [ ] Create a task dispatcher which is capable of emitting and receiving requests/responses asynchronouly -> This could improve the code and also execution time drastically.
- [ ] Get rid of slice of `wgErr` to receive errors from helpers
- [ ] .... more to add

nixpanic:
Indeed! There are also several tests that modify the StorageClass with additional/different options. This should get simplified so the deletion and re-creation of the StorageClass is not so bulky anymore.

I suggest doing something like:

```go
By(...) ... {
    withStorageClass(..., options map[string]string) ... {
        ...
    }
}
```

Where `withStorageClass()` then deletes the existing default StorageClass, and recreates it with the given options, runs the inner function and restores the original StorageClass.

Not sure what you are planning for other deployment options, but maybe you can use the same workflow as the e2e framework does.

humblec:
[...]

PR #2180, opened 2021-06-17
e2e: refactor RBD and CephFS template deployment model

humblec:
e2e: refactor RBD and CephFS template deployment model
    
    At present the RBD and CephFS deployment is really clumsy and have
    code deplication and difficulty to follow. This patch try to group the
    tasks together in more modular fashion and make it flexible to
    extend the same model for both deployment.
    
    This introduces a new interface called deployer with below methods:

        type deployer interface {
            replaceNamespaceInTemplate(string) (string, error)
            runAction(string, string, string) error
    }
        
    
    These methods handle common tasks for the deployment process of RBD and
    CephFS.The individual deployers are free to override the methods or extend
    their own later.  For example:
                    changeTemplateAndRunAction(string, string) error

     is 'deployer' specific ( rbd, cephfs...etc) which handle deployer specific cases.

    Functions removed from both cephfs.go and rbd.go:
        createORDeleteRbdResources()

 
Updates:  https://github.com/ceph/ceph-csi/issues/2178

### DIFF
--- a/e2e/deployer.go
+++ b/e2e/deployer.go
@@ -1,0 +1,44 @@
+package e2e
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+// deployer interface modify the templates on demand and also help
+// to create or delete templates.
+type deployer interface {
+	replaceNamespaceInTemplate(string) (string, error)
+	runAction(string, string, string) error
+}
+
+type deploy struct{}
+
+func (d deploy) replaceNamespaceInTemplate(filePath string) (string, error) {
+	read, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return "", err
+	}
+
+	return strings.ReplaceAll(string(read), "namespace: default", fmt.Sprintf("namespace: %s", cephCSINamespace)), nil
+}
+
+func (d deploy) runAction(template, action, ns string) error {
+	data, err := d.replaceNamespaceInTemplate(template)
+	if err != nil {
+		return fmt.Errorf("failed to read content from %s with error %w", template, err)
+	}
+	if action == "create" {
+		_, err = framework.RunKubectlInput(cephCSINamespace, data, ns, action, "-f", "-")
+	}
+	if action == "delete" {
+		_, err = framework.RunKubectlInput(cephCSINamespace, data, "--ignore-not-found=true", ns, action, "-f", "-")
+	}
+	if err != nil {
+		return fmt.Errorf("failed to %s %q with error %w", action, template, err)
+	}
+	return err
+}
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -55,104 +55,99 @@ var (
 	defaultCloneCount      = 10
 )
 
+// Keep the deployment templates in deployment order.
+var rbdTemplates = []string{
+	rbdDirPath + csiDriverObject,
+	rbdDirPath + rbdProvisioner,
+	rbdDirPath + rbdProvisionerRBAC,
+	rbdDirPath + rbdProvisionerPSP,
+	rbdDirPath + rbdNodePlugin,
+	rbdDirPath + rbdNodePluginRBAC,
+	rbdDirPath + rbdNodePluginPSP,
+}
+
+type rbdDeploy struct {
+	deployer
+	templates []string
+}
+
 func deployRBDPlugin() {
+	// create a rbd deployer
+	rd := rbdDeploy{deployer: deploy{}, templates: rbdTemplates}
+
 	// delete objects deployed by rook
-	data, err := replaceNamespaceInTemplate(rbdDirPath + rbdProvisionerRBAC)
-	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdProvisionerRBAC, err)
-	}
-	_, err = framework.RunKubectlInput(cephCSINamespace, data, "--ignore-not-found=true", ns, "delete", "-f", "-")
-	if err != nil {
-		e2elog.Failf("failed to delete provisioner rbac %s with error %v", rbdDirPath+rbdProvisionerRBAC, err)
-	}
-
-	data, err = replaceNamespaceInTemplate(rbdDirPath + rbdNodePluginRBAC)
-	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdNodePluginRBAC, err)
-	}
-	_, err = framework.RunKubectlInput(cephCSINamespace, data, "delete", "--ignore-not-found=true", ns, "-f", "-")
-	if err != nil {
-		e2elog.Failf("failed to delete nodeplugin rbac %s with error %v", rbdDirPath+rbdNodePluginRBAC, err)
+	deleteRookTemplates := []string{rbdDirPath + rbdProvisionerRBAC, rbdDirPath + rbdNodePluginRBAC}
+	for _, template := range deleteRookTemplates {
+		err := rd.runAction(template, "delete", ns)
+		if err != nil {
+			e2elog.Failf("failed to delete rook template with error %v", err)
+		}
 	}
 
-	createORDeleteRbdResources("create")
+	// Provision RBD from the CSI templates.
+	for _, template := range rd.templates {
+		err := rd.changeTemplateAndRunAction(template, "create")
+		if err != nil {
+			e2elog.Failf("failed to create template with error %v", err)
+		}
+	}
+}
+
+func (d rbdDeploy) changeTemplateAndRunAction(template, action string) error {
+	switch template {
+	case rbdDirPath + csiDriverObject:
+		csiDriver, err := ioutil.ReadFile(template)
+		if err != nil {
+			// changeTemplateAndRunAction is used for upgrade testing as csidriverObject is
+			// newly added, discarding file not found error.
+			if !os.IsNotExist(err) {
+				return fmt.Errorf("failed to read content from (%s) with error %w", template, err)
+			}
+		} else {
+			_, err = framework.RunKubectlInput(cephCSINamespace, string(csiDriver), action, "-f", "-")
+			if err != nil {
+				return fmt.Errorf("failed to %s CSIDriver object (%s) with error %w", action, template, err)
+			}
+		}
+	case rbdDirPath + rbdProvisioner:
+		data, err := d.replaceNamespaceInTemplate(template)
+		if err != nil {
+			return fmt.Errorf("failed to read content from (%s) with error %w", template, err)
+		}
+		data = oneReplicaDeployYaml(data)
+		data = enableTopologyInTemplate(data)
+		_, err = framework.RunKubectlInput(cephCSINamespace, data, action, ns, "-f", "-")
+		if err != nil {
+			return fmt.Errorf("failed to %s rbd provisioner (%s) with error %w", action, template, err)
+		}
+	case rbdDirPath + rbdNodePlugin:
+		data, err := d.replaceNamespaceInTemplate(template)
+		if err != nil {
+			return fmt.Errorf("failed to read content from (%s) with error %w", template, err)
+		}
+		domainLabel := nodeRegionLabel + "," + nodeZoneLabel
+		data = addTopologyDomainsToDSYaml(data, domainLabel)
+		_, err = framework.RunKubectlInput(cephCSINamespace, data, action, ns, "-f", "-")
+		if err != nil {
+			return fmt.Errorf("failed to %s RBD nodeplugin (%s) with error %v", action, template, err)
+		}
+	default:
+		err := d.runAction(template, action, ns)
+		if err != nil {
+			return fmt.Errorf("failed to deploy template with error %v", err)
+		}
+	}
+	return nil
 }
 
 func deleteRBDPlugin() {
-	createORDeleteRbdResources("delete")
-}
-
-func createORDeleteRbdResources(action string) {
-	csiDriver, err := ioutil.ReadFile(rbdDirPath + csiDriverObject)
-	if err != nil {
-		// createORDeleteRbdResources is used for upgrade testing as csidriverObject is
-		// newly added, discarding file not found error.
-		if !os.IsNotExist(err) {
-			e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+csiDriverObject, err)
-		}
-	} else {
-		_, err = framework.RunKubectlInput(cephCSINamespace, string(csiDriver), action, "-f", "-")
+	// delete RBD CSI templates
+	rd := rbdDeploy{deployer: deploy{}, templates: rbdTemplates}
+	for _, template := range rd.templates {
+		err := rd.changeTemplateAndRunAction(template, "delete")
 		if err != nil {
-			e2elog.Failf("failed to %s CSIDriver object with error %v", action, err)
+			e2elog.Failf("failed to delete template with error %v", err)
 		}
-	}
-	data, err := replaceNamespaceInTemplate(rbdDirPath + rbdProvisioner)
-	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdProvisioner, err)
-	}
-	data = oneReplicaDeployYaml(data)
-	data = enableTopologyInTemplate(data)
-	_, err = framework.RunKubectlInput(cephCSINamespace, data, action, ns, "-f", "-")
-	if err != nil {
-		e2elog.Failf("failed to %s rbd provisioner with error %v", action, err)
-	}
-
-	data, err = replaceNamespaceInTemplate(rbdDirPath + rbdProvisionerRBAC)
-	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdProvisionerRBAC, err)
-	}
-	_, err = framework.RunKubectlInput(cephCSINamespace, data, action, ns, "-f", "-")
-	if err != nil {
-		e2elog.Failf("failed to %s provisioner rbac with error %v", action, err)
-	}
-
-	data, err = replaceNamespaceInTemplate(rbdDirPath + rbdProvisionerPSP)
-	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdProvisionerPSP, err)
-	}
-	_, err = framework.RunKubectlInput(cephCSINamespace, data, action, "-f", "-")
-	if err != nil {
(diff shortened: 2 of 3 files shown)


e2e/rbd.go
```diff
@@ -55,104 +55,103 @@ var (
...
-	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdNodePluginRBAC, err)
-	}
-	_, err = framework.RunKubectlInput(cephCSINamespace, data, "delete", "--ignore-not-found=true", ns, "-f", "-")
-	if err != nil {
-		e2elog.Failf("failed to delete nodeplugin rbac %s with error %v", rbdDirPath+rbdNodePluginRBAC, err)
-	}
-
-	createORDeleteRbdResources("create")
+type Deployer interface {
```
nixpanic: comments would be useful here

e2e/rbd.go
```diff
@@ -55,104 +55,103 @@ var (
...
+	replaceNamespaceInTemplate(string) (string,error)
+	runAction(string, string)
+	changeTemplateAndRunAction(string, string)
 }
 
-func deleteRBDPlugin() {
-	createORDeleteRbdResources("delete")
+// Keep the deployment templates in deployment order
+var rbdTemplates = []string{rbdDirPath + csiDriverObject,
```
nixpanic: probably should place the 0th entry on a new line too, to make it easier to read

e2e/rbd.go
```diff
@@ -55,104 +55,103 @@ var (
...
+	data, err := rd.replaceNamespaceInTemplate(Template)
 	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdProvisionerPSP, err)
+		e2elog.Failf("failed to read content from %s with error %v", Template, err)
 	}
-	_, err = framework.RunKubectlInput(cephCSINamespace, data, action, "-f", "-")
+	_, err = framework.RunKubectlInput(cephCSINamespace, data, "--ignore-not-found=true", ns, action, "-f", "-")
 	if err != nil {
-		e2elog.Failf("failed to %s provisioner psp with error %v", action, err)
+		e2elog.Failf("failed to %s '%s' with error %v", action, Template, err)
```
nixpanic: use `%q` for quoting

e2e/rbd.go
```diff
@@ -55,104 +55,103 @@ var (
...
 	}
+}
 
-	data, err = replaceNamespaceInTemplate(rbdDirPath + rbdNodePlugin)
-	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdNodePlugin, err)
+func (rd *rbdDeploy) changeTemplateAndRunAction(Template string, action string) {
```
nixpanic: why pass `Template` with  a capital `T`?

e2e/rbd.go
```diff
@@ -55,104 +55,103 @@ var (
...
+	switch Template {
+	case "rbdDirPath + csiDriverObject":
```
nixpanic: huh? how does this ever match?!

I don't think this is ready for review, marking it **draft** for the time being.

humblec:
@nixpanic just started to prototype it and was about to mark WIP,  thats why I didnt request for review yet, but you were too fast to get in here :)

humblec:
Ok.. Looks like a promising run :) , the failure is on `etcd timeout`:

```
Jun 17 11:22:47.896: FAIL: failed to create vault configmap error running /usr/bin/kubectl --server=https://192.168.39.185:8443 --kubeconfig=/root/.kube/config --namespace=cephcsi-e2e-7a476401 create --namespace=cephcsi-e2e-7a476401 -f -:
Command stdout:
stderr:
Error from server: error when creating "STDIN": etcdserver: request timed out
error:
exit status 1
```

humblec:
The failure is : `failed to create PVC with error etcdserver: request timed out`

humblec:
@nixpanic @Madhu-1  can you take a look at this ? would like to continue refactoring the E2E code, so the ping.

e2e/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+type deploy struct{}
+
+func (d deploy) replaceNamespaceInTemplate(filePath string) (string, error) {
+	read, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return "", err
+	}
+	return strings.ReplaceAll(string(read), "namespace: default", fmt.Sprintf("namespace: %s", cephCSINamespace)), nil
```
Madhu-1, line 26: making namespace or argument makes it more useful

e2e/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (d deploy) runAction(template, action string) {
+	data, err := d.replaceNamespaceInTemplate(template)
+	if err != nil {
+		e2elog.Failf("failed to read content from %s with error %v", template, err)
```
Madhu-1: always return an error. Let  the caller handle the error

e2e/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	_, err = framework.RunKubectlInput(cephCSINamespace, data, "--ignore-not-found=true", ns, action, "-f", "-")
+	if err != nil {
+		e2elog.Failf("failed to %s %q with error %v", action, template, err)
+	}
+}
+
+func (d deploy) changeTemplateAndRunAction(template, action string) {
+	switch template {
+	case rbdDirPath + csiDriverObject, cephfsDirPath + csiDriverObject:
```
Madhu-1: this makes deploy specific to only RBD and cephfs not a generic one?

e2e/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		e2elog.Failf("failed to read content from %s with error %v", template, err)
+	}
+	_, err = framework.RunKubectlInput(cephCSINamespace, data, ns, action, "-f", "-")
+	if err != nil {
+		e2elog.Failf("failed to %s %q with error %v", action, template, err)
+	}
+}
+
+func (d deploy) runActionRookDelete(template, action string) {
```
Madhu-1: if d is not used drop it from the method.

e2e/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+	data, err := d.replaceNamespaceInTemplate(template)
+	if err != nil {
+		e2elog.Failf("failed to read content from %s with error %v", template, err)
+	}
+	_, err = framework.RunKubectlInput(cephCSINamespace, data, "--ignore-not-found=true", ns, action, "-f", "-")
```
Madhu-1: make `ns` as argument

e2e/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		e2elog.Failf("failed to %s %q with error %v", action, template, err)
+	}
+}
+
+func (d deploy) changeTemplateAndRunAction(template, action string) {
+	switch template {
+	case rbdDirPath + csiDriverObject, cephfsDirPath + csiDriverObject:
+		csiDriver, err := ioutil.ReadFile(rbdDirPath + csiDriverObject)
```
Madhu-1: for cephfsDir also its reading rbdDirPath?

Madhu-1 (dismissed):

e2e/deployer.go
humblec: yep..

e2e/deployer.go
humblec: could be done too.

e2e/deployer.go
humblec: I was in a double mind here , that said, as mentioned in the PR description actually we can take out this from common interface and use it seperately, but just very limited cases can cover both atm. so kept that way.  

```

The individual deployers are free to override the methods or extend
their own later. Considering, pretty much all the tasks are covered with
few cases in changeTemplateAndRunAction(), it has been kept inside the
common method, but can be split/override from deployers later if needed.

```

Do we expect any other deployment to come in ? if yes, I can prototype to be in `deployer` specific. @Madhu-1

e2e/deployer.go
humblec: Thanks.. it should have been template,, will correct it.

e2e/deployer.go
Madhu-1: Yes this need to be done for vault also https://github.com/ceph/ceph-csi/blob/devel/e2e/deploy-vault.go

humblec:
@Madhu-1 comments are addressed, ptal , Thanks .

humblec:
@Madhu-1 @nixpanic can you please take a look at this?

humblec:
Can I get some review here please ?

e2e/deployer.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	_, err = framework.RunKubectlInput(cephCSINamespace, data, ns, action, "-f", "-")
+	if err != nil {
+		return fmt.Errorf("failed to %s %q with error %w", action, template, err)
+	}
+
+	return err
+}
+
+func (d deploy) runActionRookDelete(template, action, ns string) error {
```
Madhu-1: the function name is not generic. we are not doing any specific operation related to rook here.

Madhu-1 (commented):
still vault-related are not making use of this interface?

humblec:
> still vault-related are not making use of this interface?

Vault deployment can be handled in  a follow up PR. I have aded it in the todo list.

e2e/deployer.go
humblec: @Madhu-1 any suggestions here ? the idea here is to expand the operations which are Rook specific in that method, so the mention of Rook, may be `Delete` from the name can be taken out.

Madhu-1: why not make use of the above runAction method?

humblec: @Madhu-1 thats why I said above, this method was a placeholder for rook actions as we are seperetely manipulating rook deployment before we start and it also matches with current implementation and give some clarity while invoking the method that, we are doing it on `rook` deployment

Madhu-1: For me, this doesn't look to be rook specific because we are running in  `cephCSINamespace` with `--ignore-not-found=true` instead of `runActionRookDelete` this will be a wrapper to `runAction` something like`runActionDeleteIgnoreNotFound`

humblec: @Madhu-1 now you are talking about the real body of this method or what it do , where I was mentioning about from caller's stand point. One thing to note here is our deployment methods/functions are keep doing 'almost' same actions with some changes here and there, generalizing those can be done in a different way or as a follow up.  The initial attempt here is to get the things seggregated to common methods and make use of it. Optimization/refactoring of the real actions inside functions are next items in the list. I have also renamed the method `runActionRookDelete` to `runRookAction` and also marked the todo for actions/wrappers on existing things in the tracker issue.  Let me know if thats enough?

Madhu-1: still want to make a couple of points here. lets hear from others also

* if we are sure that its a RookAction why to make use of `cephCSINamespace` 
* and we are making use of `--ignore-not-found` it's only specific to delete operations, not for other operations?. please correct me if am wrong.

humblec: @Madhu-1 the thought here is simple, if we think we dont need to keep room for any other `rook` operations apart from `delete` template, we can remove the placeholder. The idea was that, keep a rook specific function which can be enhanced based on requirement in future.. so, getting rid of `rook` specific method is what you want ?

humblec: @Madhu-1 any other comments on existing version ?

Madhu-1 (dismissed):
except for https://github.com/ceph/ceph-csi/pull/2180#discussion_r663825899 everything LGTM. this can be enhanced later

humblec:
> except for [#2180 (comment)](https://github.com/ceph/ceph-csi/pull/2180#discussion_r663825899) everything LGTM. this can be enhanced later

Thanks :+1: , let me get rid of rook method quickly so that no concern :)

humblec:
> > except for [#2180 (comment)](https://github.com/ceph/ceph-csi/pull/2180#discussion_r663825899) everything LGTM. this can be enhanced later
> 
> Thanks +1 , let me get rid of rook method quickly so that no concern :)

@madhu-1 took out rook specific method, and tests are passing... ptal. Thanks !

e2e/rbd.go
```diff
@@ -55,104 +55,92 @@ var (
...
+}
 
-	data, err = replaceNamespaceInTemplate(rbdDirPath + rbdNodePluginRBAC)
-	if err != nil {
-		e2elog.Failf("failed to read content from %s with error %v", rbdDirPath+rbdNodePluginRBAC, err)
-	}
-	_, err = framework.RunKubectlInput(cephCSINamespace, data, action, ns, "-f", "-")
-	if err != nil {
-		e2elog.Failf("failed to %s nodeplugin rbac with error %v", action, err)
+func (d rbdDeploy) changeTemplateAndRunAction(template, action string) {
```
Madhu-1: return error from `changeTemplateAndRunAction` let the caller handle the error message?

Madhu-1 (dismissed):

e2e/rbd.go
humblec: sure. done that way too.  Thanks @Madhu-1

Madhu-1 (approved):

Madhu-1:
@humblec static checks are failing

humblec:
> @humblec static checks are failing

Looks spurious from below and also it pass locally to me.

```

npm ERR! code E403
npm ERR! 403 403 Forbidden - GET https://registry.npmjs.org/@commitlint%2fcli
npm ERR! 403 In most cases, you or one of your dependencies are requesting
npm ERR! 403 a package version that is forbidden by your security policy.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2021-07-06T10_01_28_476Z-debug.log
The command '/bin/sh -c source /build.env     && dnf -y install 	git 	make 	golang 	gcc 	librados-devel 	librbd-devel 	rubygems 	ShellCheck 	yamllint 	npm 	diffutils       python3-prettytable       pylint     && dnf -y update     && dnf -y clean all     && gem install mdl     && curl -sf "https://install.goreleaser.com/github.com/golangci/golangci-lint.sh"        | bash -s -- -b ${GOPATH}/bin "${GOLANGCI_VERSION}"     && curl -sfL "https://raw.githubusercontent.com/securego/gosec/master/install.sh"        | sh -s -- -b $GOPATH/bin "${GOSEC_VERSION}"     && curl -L https://git.io/get_helm.sh | bash -s -- --version "${HELM_VERSION}"     && mkdir /opt/commitlint && pushd /opt/commitlint     && npm init -y     && npm install @commitlint/cli     && popd     && true' returned a non-zero code: 1
make: *** [Makefile:207: .test-container-id] Error 1

Error: Process completed with exit code 2.
```

humblec:
@Madhu-1 I reran it , lets see

humblec:
Considering this is on `e2e` and `deployment`  and been out for  review for 3 weeks, can we add `ready-to-merge`?

e2e/deployer.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// deployer interface modify the templates on demand and also help
+// to create or delete templates.
+type deployer interface {
+	replaceNamespaceInTemplate(string) (string, error)
+	runAction(string, string, string) error
+}
+
+type deploy struct{}
```
nixpanic, line 18: There is a lot of duplication in the deployer for cephfs and rbd, it probably makes sense to combine the two into a single `type deployer`.

When creating the `deployer` instance, pass it the list of templates, maybe with an additional attribute, like:

```go

type deploymentTemplateStage string

var (
    deploymentTemplateStageInit = deploymentTemplateStage("init") // for driver object
    deploymentTemplateStageConfig = deploymentTemplateStage("config")
    deploymentTemplateStageProvisioner = deploymentTemplateStage("provisioner")
    ...
)

type deploymentTemplate struct {
    filename string
    stage deploymentTemplateStage
}

type deployer struct {
    templates []deploymentTemplate
}

cephfsDeployer := deployer{
    templates: []deploymentTemplate{
        {pathToDriverTemplate, deploymentTemplateStageInit},
        {pathToConfigTemplate, deploymentTemplateStageConfig},
        {pathToMoreConfigTemplate, deploymentTemplateStageConfig},
        {pathToProvisionerTemplate, deploymentTemplateStageProvisioner},
    }
}
```

By using defined `deploymentTemplateStage`s or similar, there is no need to have switch/case based on the path of the templates and things become much more re-usable.

nixpanic (changes requested):
just sharing some ideas to clean things up more, and really use a `deployer` API that can be re-used and extended with less effort

nixpanic:
As mentioned over chat, there are several things that I would like to see improved. My previous comment might not have been clear enough, sorry.

- no need for multiple `deploy` types
- `templates` in the `rbdDeploy` and `cephfsDeploy` is not a useful list of filenames to templates, as the `changeTemplateAndRunAction` compares the item with the path?! There should be no need to compare paths if the list of templates is expected to be adaptable, as that always requires changes in the list and the function
- the helper functions for the `deploy` type do not use anything from the `deploy` struct, they can be just stand-alone helper functions
- a possible extension to deploy Vault should definitely not require adding yet an other `deploy` type, a single `deploy` type that can apply templates in different ways (or 'stages' as I called them above) would make things so much simpler to use

In general, refactoring should make the code easier to understand and better to maintain. This current PR increases the line-count and adds obfuscation (by having the filenames of the templates 2x). IMHO this does not contribute to better maintainability of the e2e tests.

humblec:
> As mentioned over chat, there are several things that I would like to see improved. My previous comment might not have been clear enough, sorry.
> 
....

As mentioned over the chat I am making my view here:

*)  The `changeTemplateandRunAction` was in common deployer interface which has been taken out as per the review comment (https://github.com/ceph/ceph-csi/pull/2180#discussion_r658485887), so , kept inside the deployer specific paths, 
*) The current model of embedding interface make use of common mehods whereever possible even from seperate deployers, so no code duplication..
*) `deploy` is a placeholder for operating on the templates and will be transitioning to method receiver  as mentioned over previous comments
*) The logic of 'some' templates are  different  ( for example : provisioner) for rbd and cephfs.
*) Also, as noted in the issue tracker, ideally the deployment of `templates` has to make use of builtin methods like  `f,load/create/manifests..etc eventually which is totally different route altogther. 

Obviously we can squash the phases of optimization in one attempt or go with different model altogether, or have intermediate phases without much disturbances to the code structure we have. The latter is the attempt here, if we want to go with former one, it is possible too and it looks like thats what the request here.

AI : /hold this PR

nixpanic:
fwiw, #2270 is an example that re-models the KMS configuration so that it becomes more modular. Functions use an interface as parameter, and use the methods of the structs to call the implementation. This allows for simple addition of new objects with different properties for the different KMS backends. It prevents switch/cases for each object.

I hope that the example is clear, and you can use a similar approach for deploying templates with different requirements.

PR #2208, opened 2021-06-24
ci: configure `lll` linter to its default value ie 120 chars

humblec:
`lll` linter configuration was made for 180 chars till now due to
long source code signatures/comments we had in the project. This
address it and enabling linter with 120 chars check.
    
### DIFF
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -66,26 +66,58 @@ func init() {
 	flag.IntVar(&conf.PidLimit, "pidlimit", 0, "the PID limit to configure through cgroups")
 	flag.BoolVar(&conf.IsControllerServer, "controllerserver", false, "start cephcsi controller server")
 	flag.BoolVar(&conf.IsNodeServer, "nodeserver", false, "start cephcsi node server")
-	flag.StringVar(&conf.DomainLabels, "domainlabels", "", "list of kubernetes node labels, that determines the topology"+
-		" domain the node belongs to, separated by ','")
+	flag.StringVar(
+		&conf.DomainLabels,
+		"domainlabels",
+		"",
+		"list of kubernetes node labels, that determines the topology"+
+			" domain the node belongs to, separated by ','")
 
 	// cephfs related flags
-	flag.BoolVar(&conf.ForceKernelCephFS, "forcecephkernelclient", false, "enable Ceph Kernel clients on kernel < 4.17 which support quotas")
+	flag.BoolVar(
+		&conf.ForceKernelCephFS,
+		"forcecephkernelclient",
+		false,
+		"enable Ceph Kernel clients on kernel < 4.17 which support quotas")
 
 	// liveness/grpc metrics related flags
 	flag.IntVar(&conf.MetricsPort, "metricsport", 8080, "TCP port for liveness/grpc metrics requests")
-	flag.StringVar(&conf.MetricsPath, "metricspath", "/metrics", "path of prometheus endpoint where metrics will be available")
+	flag.StringVar(
+		&conf.MetricsPath,
+		"metricspath",
+		"/metrics",
+		"path of prometheus endpoint where metrics will be available")
 	flag.DurationVar(&conf.PollTime, "polltime", time.Second*pollTime, "time interval in seconds between each poll")
 	flag.DurationVar(&conf.PoolTimeout, "timeout", time.Second*probeTimeout, "probe timeout in seconds")
 
 	flag.BoolVar(&conf.EnableGRPCMetrics, "enablegrpcmetrics", false, "[DEPRECATED] enable grpc metrics")
-	flag.StringVar(&conf.HistogramOption, "histogramoption", "0.5,2,6",
-		"[DEPRECATED] Histogram option for grpc metrics, should be comma separated value, ex:= 0.5,2,6 where start=0.5 factor=2, count=6")
+	flag.StringVar(
+		&conf.HistogramOption,
+		"histogramoption",
+		"0.5,2,6",
+		"[DEPRECATED] Histogram option for grpc metrics, should be comma separated value, "+
+			"ex:= 0.5,2,6 where start=0.5 factor=2, count=6")
 
-	flag.UintVar(&conf.RbdHardMaxCloneDepth, "rbdhardmaxclonedepth", 8, "Hard limit for maximum number of nested volume clones that are taken before a flatten occurs")
-	flag.UintVar(&conf.RbdSoftMaxCloneDepth, "rbdsoftmaxclonedepth", 4, "Soft limit for maximum number of nested volume clones that are taken before a flatten occurs")
-	flag.UintVar(&conf.MaxSnapshotsOnImage, "maxsnapshotsonimage", 450, "Maximum number of snapshots allowed on rbd image without flattening")
-	flag.UintVar(&conf.MinSnapshotsOnImage, "minsnapshotsonimage", 250, "Minimum number of snapshots required on rbd image to start flattening")
+	flag.UintVar(
+		&conf.RbdHardMaxCloneDepth,
+		"rbdhardmaxclonedepth",
+		8,
+		"Hard limit for maximum number of nested volume clones that are taken before a flatten occurs")
+	flag.UintVar(
+		&conf.RbdSoftMaxCloneDepth,
+		"rbdsoftmaxclonedepth",
+		4,
+		"Soft limit for maximum number of nested volume clones that are taken before a flatten occurs")
+	flag.UintVar(
+		&conf.MaxSnapshotsOnImage,
+		"maxsnapshotsonimage",
+		450,
+		"Maximum number of snapshots allowed on rbd image without flattening")
+	flag.UintVar(
+		&conf.MinSnapshotsOnImage,
+		"minsnapshotsonimage",
+		250,
+		"Minimum number of snapshots required on rbd image to start flattening")
 	flag.BoolVar(&conf.SkipForceFlatten, "skipforceflatten", false,
 		"skip image flattening if kernel support mapping of rbd images which has the deep-flatten feature")
 
--- a/docs/coding.md
+++ b/docs/coding.md
@@ -69,6 +69,33 @@ import (
 * Use log level `ERROR` when something occurs which is fatal to the operation,
   but not to the service or application.
 
+### Wrap long lines
+
+At present, we restrict the number of chars in a line to `120` which is the
+default value for the `lll` linter check we have in CI. If your source code line
+or comment goes beyond this limit please try to organize it in such a way it
+is within the line length limit and help on code reading.
+
+Example:
+
+```
+_, err := framework.RunKubectl(cephCSINamespace, "delete", "cm", "ceph-csi-encryption-kms-config", "--namespace", cephCSINamespace, "--ignore-not-found=true")
+```
+
+Instead of above long function signature, we can organize it to something like below
+which is clear and help on easy code reading.
+
+```
+_, err := framework.RunKubectl(
+            cephCSINamespace,
+            "delete",
+            "cm",
+            "ceph-csi-encryption-kms-config",
+            "--namespace",
+            cephCSINamespace,
+            "--ignore-not-found=true")
+```
+
 ### Mark Down Rules
 
 * MD014 - Dollar signs used before commands without showing output
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -160,7 +160,10 @@ func validateSubvolumePath(f *framework.Framework, pvcName, pvcNamespace, fileSy
 		return err
 	}
 	if subVolumePath != subVolumePathInPV {
-		return fmt.Errorf("subvolumePath %s is not matching the subvolumePath %s in PV", subVolumePath, subVolumePathInPV)
+		return fmt.Errorf(
+			"subvolumePath %s is not matching the subvolumePath %s in PV",
+			subVolumePath,
+			subVolumePathInPV)
 	}
 	return nil
 }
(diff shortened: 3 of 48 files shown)


nixpanic:
please run `make containerized-test` locally and make sure it passes the change to the linter configuration

humblec:
@nixpanic up for review :+1:

scripts/golangci.yml
```diff
@@ -0,0 +1,170 @@
+---
```
Madhu-1: This need to be modified at https://github.com/ceph/ceph-csi/blob/devel/scripts/golangci.yml.in#L113

Madhu-1 (dismissed):

scripts/golangci.yml
humblec: @Madhu-1 didnt I modify it already https://github.com/ceph/ceph-csi/pull/2208/commits/a861f252acb44c9b77afd54a5a7efad41583d2eb#diff-2a662b48bcaf012b73a39baf1703a975ab034cb939366529e9fb36c64943fa8fR112? :)

humblec: @Madhu-1 rebased this PR, can you please take a look at this?  this will have conflict coming in each other PR getting in. So the ping.

e2e/cephfs.go
```diff
@@ -298,7 +306,12 @@ var _ = Describe("cephfs", func() {
 
 			By("create PVC in storageClass with volumeNamePrefix", func() {
 				volumeNamePrefix := "foo-bar-"
-				err := createCephfsStorageClass(f.ClientSet, f, false, map[string]string{"volumeNamePrefix": volumeNamePrefix})
+				err := createCephfsStorageClass(
+					f.ClientSet,
+					f,
+					false,
+					map[string]string{"volumeNamePrefix": volumeNamePrefix},
```
nixpanic: The trailing comma suggests that additional options are possible, but I do not think that `createCephfsStorageClass()` supports that. Is the comma really required?

nixpanic (commented):
Maybe you can document something about coding guidelines? Placing all arguments to many functions make some of the files increase their linecount really a lot. For many, passing a few arguments on the same line would be just as clear without the additional bloat that spreads out the (already long) functions even further.

scripts/golangci.yml
nixpanic: `scripts/golangci.yml` should not be included, it gets re-generated when `scripts/golangci.yml.in` has been modified.

nixpanic (dismissed):

Madhu-1:
Updating the development guide for future is helpful

e2e/cephfs.go
humblec: sure, removed the `traliing comma` here,

humblec:
> Maybe you can document something about coding guidelines? Placing all arguments to many functions make some of the files increase their linecount really a lot. For many, passing a few arguments on the same line would be just as clear without the additional bloat that spreads out the (already long) functions even further.

sure, while the line length goes beyond, it has been split, while splitting: placing one by one instead of some in one line and others in next is not better I believe, as the latter is not consistent either. 

@Madhu-1 @nixpanic  i have updated the development guide too. ptal :+1:

docs/coding.md
```diff
@@ -69,6 +69,34 @@ import (
 * Use log level `ERROR` when something occurs which is fatal to the operation,
   but not to the service or application.
 
+### Wrap long lines
+
+* At present, we restrict the number of chars in a line to `120` which is the
```
nixpanic: no need for the bullet point or indention here, it is the only item in this section of the doc

docs/coding.md
```diff
@@ -69,6 +69,34 @@ import (
...
+  default value for the `lll` linter check we have in CI. If your source code line
+  or comment goes beyond this limit please try to organize it in such a way it
+  is within the line length limit and help on code reading.
+
+  Example:
+
+```
+  _, err := framework.RunKubectl(cephCSINamespace, "delete", "cm", "ceph-csi-encryption-kms-config", "--namespace", cephCSINamespace, "--ignore-not-found=true")
```
nixpanic: this is not indented either, nor the paragraphs below.

cmd/cephcsi.go
```diff
@@ -66,26 +66,66 @@ func init() {
...
-
-	flag.UintVar(&conf.RbdHardMaxCloneDepth, "rbdhardmaxclonedepth", 8, "Hard limit for maximum number of nested volume clones that are taken before a flatten occurs")
-	flag.UintVar(&conf.RbdSoftMaxCloneDepth, "rbdsoftmaxclonedepth", 4, "Soft limit for maximum number of nested volume clones that are taken before a flatten occurs")
-	flag.UintVar(&conf.MaxSnapshotsOnImage, "maxsnapshotsonimage", 450, "Maximum number of snapshots allowed on rbd image without flattening")
-	flag.UintVar(&conf.MinSnapshotsOnImage, "minsnapshotsonimage", 250, "Minimum number of snapshots required on rbd image to start flattening")
+	flag.StringVar(
+		&conf.HistogramOption,
+		"histogramoption",
+		"0.5,2,6",
+		"[DEPRECATED] Histogram option for grpc metrics, should be comma separated value,"+
```
nixpanic: missing space after the comma inside the string at `,"`

internal/cephfs/clone.go
```diff
@@ -134,7 +134,11 @@ func createCloneFromSubvolume(ctx context.Context, volID, cloneID volumeID, volO
 	return nil
 }
 
-func cleanupCloneFromSubvolumeSnapshot(ctx context.Context, volID, cloneID volumeID, parentVolOpt *volumeOptions) error {
+func cleanupCloneFromSubvolumeSnapshot(
+	ctx context.Context,
+	volID, cloneID volumeID,
+	parentVolOpt *volumeOptions,
```
nixpanic: is the comma at the end really required? it looks so incomplete when used with function calls

nixpanic:
In general for PRs of this type, we recommend to send multiple small PRs. A PR for a package like `internal/` is not practical, it will be easier to split it up in PRs per sub-package, or even per (group of) files. That way, it will be easier for others to review, reduce the number of conflicts you need to correct in this PR, and make it easier for other developers to rebase their PRs after parts have been merged.

humblec:
> In general for PRs of this type, we recommend to send multiple small PRs. A PR for a package like `internal/` is not practical, it will be easier to split it up in PRs per sub-package, or even per (group of) files. That way, it will be easier for others to review, reduce the number of conflicts you need to correct in this PR, and make it easier for other developers to rebase their PRs after parts have been merged.

Sure, I can split large commits to groups or subpackage level.. Let me revisit this

humblec:
@nixpanic all comments are addressed and commits are also organized. PTAL :+1:

humblec:
@Madhu-1 ptal

Madhu-1 (approved):

cmd/cephcsi.go
```diff
@@ -66,26 +66,58 @@ func init() {
 	flag.IntVar(&conf.PidLimit, "pidlimit", 0, "the PID limit to configure through cgroups")
 	flag.BoolVar(&conf.IsControllerServer, "controllerserver", false, "start cephcsi controller server")
 	flag.BoolVar(&conf.IsNodeServer, "nodeserver", false, "start cephcsi node server")
-	flag.StringVar(&conf.DomainLabels, "domainlabels", "", "list of kubernetes node labels, that determines the topology"+
-		" domain the node belongs to, separated by ','")
+	flag.StringVar(
```
Madhu-1, line 69: Commit description is not matching the changes , can you please revisit this one and cross check other commits also

Madhu-1 (dismissed):

humblec:
@Madhu-1 corrected, rest of the commit looks good .. Thanks .

e2e/rbd.go
```diff
@@ -331,7 +338,13 @@ var _ = Describe("RBD", func() {
...
 					e2elog.Failf("failed to delete storageclass with error %v", err)
 				}
-				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, map[string]string{"csi.storage.k8s.io/fstype": "ext4"}, deletePolicy)
+				err = createRBDStorageClass(
+					f.ClientSet,
+					f,
+					defaultSCName,
+					nil,
+					map[string]string{"csi.storage.k8s.io/fstype": "ext4"},
+					deletePolicy)
```
Yuggupta27, line 347: Since we are planning to split the parameters into different lines, It becomes difficult to read as it is harder to find the end of the function call. 
Suggested change(here and everywhere else):
In place of closing the function call with 
```
deletePolicy)
```
Closing it out in a separate line will make it more readable.
```
deletePolicy
)
```


E.g. 
```
err = createRBDStorageClass(
					f.ClientSet,
					f,
					defaultSCName,
					nil,
					map[string]string{"csi.storage.k8s.io/fstype": "ext4"},
					deletePolicy
                           )
```
Looks more clear to me than:
```
err = createRBDStorageClass(
					f.ClientSet,
					f,
					defaultSCName,
					nil,
					map[string]string{"csi.storage.k8s.io/fstype": "ext4"},
					deletePolicy)
```

Thoughts?

Yuggupta27 (changes requested):

Madhu-1 (dismissed):

e2e/rbd.go
nixpanic: Yes, when placing all arguments for a function on separate lines, sufficient indenting (automated by `go fmt`) and the closing `)` on a separate line too makes sense to me.

e2e/pod.go
```diff
@@ -128,7 +137,11 @@ func findPodAndContainerName(f *framework.Framework, ns, cn string, opt *metav1.
 	return podList.Items[0].Name, podList.Items[0].Spec.Containers[0].Name, nil
 }
 
-func getCommandInPodOpts(f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (framework.ExecOptions, error) {
+func getCommandInPodOpts(
+	f *framework.Framework,
+	c, ns, cn string,
+	opt *metav1.ListOptions,
```
nixpanic: this is still using the last-argument-has-trailing-comma style

nixpanic (dismissed):
There is currently a mix of

```go
func bla(
    arg int,
    s string,
) error {
```

and

```go
func bla(
    arg int,
    s string) error {
```

I really do not like the trailing comma for function arguments (it is fine when filling a struct or map). So, ideally I prefer the format

```go
func bla(
    arg int,
    s string
) error {
```

but as that was not used anywhere, maybe Golang (or `go fmt`) does not allow it? If that is the case, I would opt for

```go
func bla(
    arg int,
    s string) error {
```

This is a **-1** only because there is isn't a single style used consequently. When changing the style, it should be the same everywhere.

Madhu-1:
> There is currently a mix of
> 
> ```go
> func bla(
>     arg int,
>     s string,
> ) error {
> ```
> 
> and
> 
> ```go
> func bla(
>     arg int,
>     s string) error {
> ```
> 
> I really do not like the trailing comma for function arguments (it is fine when filling a struct or map). So, ideally I prefer the format
> 
> ```go
> func bla(
>     arg int,
>     s string
> ) error {
> ```
> 
> but as that was not used anywhere, maybe Golang (or `go fmt`) does not allow it? If that is the case, I would opt for
> 
> ```go
> func bla(
>     arg int,
>     s string) error {
> ```
> 
> This is a **-1** only because there is isn't a single style used consequently. When changing the style, it should be the same everywhere.

The build will fail for the below format
```
func bla(
    arg int,
    s string
) error {
```
it should be
```
func bla(
    arg int,
    s string,
) error {
```

humblec:
> ```
> func bla(
>     arg int,
>     s string
> ) error {
> ```
> 
> it should be
> 
> ```
> func bla(
>     arg int,
>     s string,
> ) error {
> ```

@nixpanic @Yuggupta27  @Madhu-1 Above was the style in `very first version of the PR`, then changed everything to avoid trailing comma in last arg of functions as per previous discussions and `)` parenthesis brought back to last arg line along with `returns`. Ideally, above is ( which was the first version) is what I have seen in other projects. I dont see a reason for not following that.

Are we in agreement for above version ? 

or we want something like below?

```
func blah(
arg int,
s string) error {
...
```

please ack any of this , its really difficult to keep changing different formats in each version of the PR!

nixpanic:
Functions are not expected to change the number of arguments often, so I would prefer

```go
func blah(
    arg int,
    s string) error {
...
```

for the definition, as well as the places where they get called:

```go
                                        err := blah(
                                            arg,
                                            s)
```

But obviously only when there is a need, and the line length crosses the limit.

humblec:
@nixpanic @Madhu-1 @Yuggupta27 ptal .

e2e/pvc.go
```diff
@@ -141,7 +154,10 @@ func deletePVCAndPV(c kubernetes.Interface, pvc *v1.PersistentVolumeClaim, pv *v
 	})
 }
 
-func getPVCAndPV(c kubernetes.Interface, pvcName, pvcNamespace string) (*v1.PersistentVolumeClaim, *v1.PersistentVolume, error) {
+func getPVCAndPV(
+	c kubernetes.Interface,
+	pvcName, pvcNamespace string,
```
nixpanic: :bell: missed one

nixpanic (dismissed):
argh, **so close**!

e2e/pvc.go
humblec: @nixpanic corrected this too :),, thanks . ptal

nixpanic (approved):

nixpanic:
> /test ci/centos/mini-e2e/k8s-1.20

poking Jenkins after service outage

Madhu-1 (approved):

PR #2213, opened 2021-06-25
cleanup: error is not nil but returns nil

yati1998:
In few places, the error is not nil still it
returns nil.

Updates: #1586

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -311,7 +311,7 @@ func validateNormalUserPVCAccess(pvcPath string, f *framework.Framework) error {
 	}
 	_, stdErr, err := execCommandInPod(f, "echo testing > /target/testing", app.Namespace, &opt)
 	if err != nil {
-		return nil
+		return fmt.Errorf("failed to exec command in pod: %w", err)
 	}
 	if stdErr != "" {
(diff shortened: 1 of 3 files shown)


e2e/utils.go
```diff
@@ -378,7 +378,7 @@ func checkDataPersist(pvcPath, appPath string, f *framework.Framework) error {
 
 	_, stdErr, err := execCommandInPod(f, fmt.Sprintf("echo %s > %s", data, filePath), app.Namespace, &opt)
 	if err != nil {
-		return nil
+		return fmt.Errorf("failed to exec command in pod: %w", err)
```
nixpanic, line 398: When you notice duplication like the above writing of data to a file, you can always create a helper function that returns a single error instead of `stdErr` and `err`. Not required for this PR, but small things like that are good cleanups too!

nixpanic (approved):

e2e/utils.go
yati1998: Yes sure. I will do it another PR.

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

PR #2215, opened 2021-06-28
e2e: remove passing waitgroups for go routines for various operations

humblec:
We are unwantedly passing around waitgroups which is not required.
This address the same inside e2e/*.go and remove waitgroup arg.

Updates: #2178 

### DIFF
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -803,11 +803,11 @@ var _ = Describe("cephfs", func() {
 					snap.Spec.Source.PersistentVolumeClaimName = &pvc.Name
 					// create snapshot
 					for i := 0; i < totalCount; i++ {
-						go func(w *sync.WaitGroup, n int, s snapapi.VolumeSnapshot) {
+						go func(n int, s snapapi.VolumeSnapshot) {
 							s.Name = fmt.Sprintf("%s%d", f.UniqueName, n)
 							wgErrs[n] = createSnapshot(&s, deployTimeout)
-							w.Done()
-						}(&wg, i, snap)
+							wg.Done()
+						}(i, snap)
 					}
 					wg.Wait()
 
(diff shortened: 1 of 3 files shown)


humblec:
@Madhu-1 @nixpanic ptal

nixpanic (commented):
I think this makes it more difficult to move the `go func(...)...` calling into helper functions? It would be nice to call something like
```go
func doOneItemOfLoop(n int, p v1.PersistentVoluemClaim) error { ... }

...

    for i := 0; i < totalCount; i++ {
        go doOneItemOfLoop(i, pvcClone)
    }
```

That would make the code easier to read, and the helper functions can hopefully be re-used.

humblec:
> I think this makes it more difficult to move the `go func(...)...` calling into helper functions? It would be nice to call something like
> 
> ```go
> func doOneItemOfLoop(n int, p v1.PersistentVoluemClaim) error { ... }
> 
> ...
> 
>     for i := 0; i < totalCount; i++ {
>         go doOneItemOfLoop(i, pvcClone)
>     }
> ```
> 
> That would make the code easier to read, and the helper functions can hopefully be re-used.

@nixpanic indeed , I am reworking the communication part in e2e  and getting rid of the stuffs ( error waiting ..etc) one by one which could make the `e2e` test faster.. but in its current form we dont need that `waitgroup` there. its difficult to take everything in one step, so `cleanup+improvement` is the step followed here :) ..

nixpanic:
> @nixpanic indeed , I am reworking the communication part in e2e and getting rid of the stuffs ( error waiting ..etc) one by one which could make the `e2e` test faster.. but in its current form we dont need that `waitgroup` there. its difficult to take everything in one step, so `cleanup+improvement` is the step followed here :) ..

Great! Can you please share the steps in your plan? Maybe create an issue for this with

- [x] each step (#2215)
- [ ] that you (...)
- [ ] are planning (...)
- [ ] to do (...)

That makes reviewing easier, and suggestions like the above maybe unneeded.

humblec:
@nixpanic will add more details, however updated this issue https://github.com/ceph/ceph-csi/issues/2178#issue-923581992 to some more extent.

Madhu-1 (approved):

nixpanic (approved):

humblec:
@Mergifyio rebase

PR #2218, opened 2021-06-29
e2e: check isRetryableAPIError during polling

Madhu-1:
added check to retry if it's an etcd timeout error when the function is polling for the resource.

Note:- This will make things better in some cases where we are polling.

### DIFF
--- a/e2e/errors.go
+++ b/e2e/errors.go
@@ -27,6 +27,7 @@ func isRetryableAPIError(err error) bool {
 		apierrors.IsTooManyRequests(err) || utilnet.IsProbableEOF(err) || utilnet.IsConnectionReset(err) {
 		return true
 	}
+
 	// If the error sends the Retry-After header, we respect it as an explicit confirmation we should retry.
 	if _, shouldRetry := apierrors.SuggestsClientDelay(err); shouldRetry {
 		return true
--- a/e2e/pod.go
+++ b/e2e/pod.go
@@ -81,12 +81,10 @@ func waitForDeploymentComplete(name, ns string, c kubernetes.Interface, t int) e
 	err = wait.PollImmediate(poll, timeout, func() (bool, error) {
 		deployment, err = c.AppsV1().Deployments(ns).Get(context.TODO(), name, metav1.GetOptions{})
 		if err != nil {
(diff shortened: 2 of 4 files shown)


e2e/errors.go
```diff
@@ -27,6 +29,9 @@ func isRetryableAPIError(err error) bool {
 		apierrors.IsTooManyRequests(err) || utilnet.IsProbableEOF(err) || utilnet.IsConnectionReset(err) {
 		return true
 	}
+	if strings.Contains(err.Error(), "etcdserver: request timed out") {
```
nixpanic: Is this really needed? I would expect that this matches `apierrors.IsTimeout(err)`. In case it is needed, a comment that `IsTimeout()` does not match would be suitable.

nixpanic:
This also "Updates: #1969", I think?

e2e/errors.go
Madhu-1: I didn't try. looks like it was not working as expected for the same reason we added this case at https://github.com/ceph/ceph-csi/blob/devel/e2e/pod.go#L85. am moving the check to `isRetryableAPIError` so that this can easily be reused for other calls also.

Madhu-1: I have removed this one hoping this will be covered by either `apierrors.IsTimeout(err)` or `apierrors.IsServerTimeout(err)` if that also does not helps we can bring it back.

nixpanic (approved):

e2e/errors.go
```diff
@@ -27,6 +27,7 @@ func isRetryableAPIError(err error) bool {
 		apierrors.IsTooManyRequests(err) || utilnet.IsProbableEOF(err) || utilnet.IsConnectionReset(err) {
 		return true
 	}
+
```
yati1998, line 30: Maybe we can remove this extra space if not required.

Madhu-1: This is kept for block separation. IMO its not a problem

yati1998: okay.

yati1998 (approved):

Madhu-1:
@mergifyio rebase

humblec (approved):

Issue #2219, opened 2021-06-29
RBD: Image metadata created by cephcsi on rbd images is not  mirrored or copied  when clone is created.


Madhu-1:
Currently, these are the metadata keys used by cephcsi to set on rbd images

```go
// Encryption

// image metadata key for encryption
encryptionMetaKey = ".rbd.csi.ceph.com/encrypted"
// metadataDEK is the key in the image metadata where the (encrypted) DEK is stored.
metadataDEK = ".rbd.csi.ceph.com/dek"

// Thick Provisioner
// image metadata key for thick-provisioning
thickProvisionMetaKey = ".rbd.csi.ceph.com/thick-provisioned"
```

From RBD keys starting with ".rbd" are considered internal and not mirrored and such keys are not copied when cloning or deep-copying locally either.

We need to change the key from `.rbd` to `rbd` so that metadata is available when mirroring is enabled  or when we create internal clones

This issue is to track below items

- [x] Change metadata key  for thick provisioning (#2231)
- [ ] During other operations like (snapshot and clone) copy keys only if the key is starting with `.rbd`

Madhu-1:
More details at https://github.com/ceph/ceph-csi/issues/2121

Madhu-1:
To summarize, because of this one we cannot support mirroring of the Thick, Encrypted PVC created from older cephcsi releases.

PR #2231, opened 2021-06-30
rbd: deprecate .rbd.csi.ceph.com/thick-provisioned metadata key

### LINKED ISSUE #2121: RBD Regional DR: Test/Support mirroring of Thick PVC
As CephCSI supports the creation of Thick PVC. we need to support mirroring same rbd images in Regional DR

Madhu-1:
As image metadata key starting with '.rbd' will not be copied when we do clone or mirroring, deprecating the old key for the same reason use 'csi.ceph.com/thick-provisioned' to set image metadata. set thickProvisionMetaKey only if the parent volume is having a deprecated metadata key set.

updates #2219 
closes:  #2121

### Backward compatibility

* Create Thick PVC from cephcsi 3.3.0
* Upgrade cephcsi from  3.3.0 to latest(custom build from this PR)
* Create Snapshot from Thick PVC-> Verify metadata with the new key is  present on rbd image created during CreateSnapshot operation)
* Restore Snapshot to a thick PVC -> Verify the metadata with new key
* Create PVC-PVC clone from Thick PVC created during 3.3.0 -> Verify  the metadata with a new key on the clone image

### Mirroring

Able to mount/resize the rbd image after the failover to another cluster.

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -1524,10 +1524,10 @@ func (ri *rbdImage) SetMetadata(key, value string) error {
 
 // setThickProvisioned records in the image metadata that it has been
 // thick-provisioned.
-func (rv *rbdVolume) setThickProvisioned() error {
-	err := rv.SetMetadata(thickProvisionMetaKey, "true")
+func (ri *rbdImage) setThickProvisioned() error {
+	err := ri.SetMetadata(thickProvisionMetaKey, "true")
 	if err != nil {
-		return fmt.Errorf("failed to set metadata %q for %q: %w", thickProvisionMetaKey, rv, err)
+		return fmt.Errorf("failed to set metadata %q for %q: %w", thickProvisionMetaKey, ri, err)
 	}
 
 	return nil
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -58,8 +58,12 @@ const (
 	rbdTaskRemoveCmdInvalidString2      = "Error EINVAL: invalid command"
 	rbdTaskRemoveCmdAccessDeniedMessage = "Error EACCES:"
 
-	// image metadata key for thick-provisioning
-	thickProvisionMetaKey = ".rbd.csi.ceph.com/thick-provisioned"
+	// image metadata key for thick-provisioning.
+	// As image metadata key starting with '.rbd' will not be copied when we do
+	// clone or mirroring, deprecating the old key for the same reason use
+	// 'thickProvisionMetaKey' to set image metadata.
+	deprecatedthickProvisionMetaKey = ".rbd.csi.ceph.com/thick-provisioned"
+	thickProvisionMetaKey           = "rbd.csi.ceph.com/thick-provisioned"
 )
 
 // rbdImage contains common attributes and methods for the rbdVolume and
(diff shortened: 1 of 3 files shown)


internal/rbd/rbd_util.go
```diff
@@ -1533,10 +1537,14 @@ func (rv *rbdVolume) setThickProvisioned() error {
 // the expansion can be allocated too.
 func (ri *rbdImage) isThickProvisioned() (bool, error) {
 	value, err := ri.GetMetadata(thickProvisionMetaKey)
-	if err != nil {
+	if err == librbd.ErrNotFound {
+		// check if the image is having deprecated metadata key.
+		value, err = ri.GetMetadata(deprecatedthickProvisionMetaKey)
```
nixpanic, line 1559: If we get here, it may be simpler to set the new key in any case. That way, only one check for the metadata is needed when the image is thin-provisioned.

It also helps with converting old images to the new metadata, and the deprecated key can be used only in this function (or ideally a new helper).

Madhu-1: @nixpanic if the error  is `err == librbd.ErrNotFound` and the metadata key `thickProvisionMetaKey` either it could be thin volume or it might be thick volume with old metadata set. For converting, yes if the value is "true" then we can set the metadata key again which will convert the old images to new metadata. Is that what you are looking for?

```suggestion
func (ri *rbdImage) isThickProvisioned() (bool, error) {
	value, err := ri.GetMetadata(thickProvisionMetaKey)
	if err == librbd.ErrNotFound {
		// check if the image is having deprecated metadata key.
		value, err = ri.GetMetadata(deprecatedthickProvisionMetaKey)
		if err == librbd.ErrNotFound {
			return false, nil
		}
		if value == "true" {
		      err=vi.setThickProvisioned(nil)
		      if err!=nil{
		            return false,err
		      }
		}...
	}
```

nixpanic: When the image is thin-provisioned, setting `thickProvisionMetaKey` to `false` will prevent a 2nd lookup of the metadata on subsequent usage. That makes it also unneeded for `setThickProvisioned()` to check for a parent volume and makes the rest of this PR simpler.

Eventually we can set the metadata (when thin-provisioned) on image creation, if we decide that has a benefit.

Madhu-1: Yes, that makes sense but the only thing is we will be setting image metadata on all images irrespective it's thick or thin. Am fine either way let me know the suggestions. but still, we need to pass parent volume to `setThickProvisioned` to support backward compatibility?

nixpanic: Setting the value would not hurt, so if it is not intentionally set (thin-provisioned), setting it to `false` is fine.

I do not think it will be needed to check if a parent volume has the metadata set when detecting the old key and setting the new one. I can not think of a reason why that is needed, as we already explicitly set the metadata when cloning or restoring from a snapshot.

nixpanic: Still waiting on an update for this. There should not be need to check the parent  volume when getting/setting the thick-provisioned key.

nixpanic (changes requested):

Madhu-1:
Busy with other PR. will revisit this tomorrow

internal/rbd/rbd_util.go
Madhu-1: made changes as per suggestion.

nixpanic (approved):
great, much cleaner! :+1:

humblec (approved):

PR #2232, opened 2021-06-30
e2e: do not create a single-item list

nixpanic:
The deployment of the Vault ConfigMap for the init-scripts job contains
a List with a single Item. This can be cleaned up to just be a ConfigMap
(without the list structure around it).

---

<details>

  progress or results

</details>

### DIFF
--- a/examples/kms/vault/vault.yaml
+++ b/examples/kms/vault/vault.yaml
@@ -60,60 +60,54 @@ spec:
               name: vault-api
 ---
 apiVersion: v1
-items:
-  - apiVersion: v1
-    data:
-      init-vault.sh: |
-        set -x -e
+kind: ConfigMap
+metadata:
+  name: init-scripts
+data:
+  init-vault.sh: |
+    set -x -e
 
-        timeout 300 sh -c 'until vault status; do sleep 5; done'
+    timeout 300 sh -c 'until vault status; do sleep 5; done'
 
-        # login into vault to retrieve token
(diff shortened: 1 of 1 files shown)


examples/kms/vault/vault.yaml
```diff
@@ -60,60 +60,54 @@ spec:
...
-items:
-  - apiVersion: v1
-    data:
-      init-vault.sh: |
-        set -x -e
-
-        timeout 300 sh -c 'until vault status; do sleep 5; done'
+kind: ConfigMap
+metadata:
+  name: init-scripts
```
Yuggupta27, line 65: Should we keep the creationTimestamp to be `null` as the previous yaml? Since it a configmap, it doesn't make much sense to keep track of its creation time. Wdyt?

Yuggupta27 (changes requested):

examples/kms/vault/vault.yaml
Rakshith-R: On the other hand,  it's does not matter at all. Looks good to me ;)

Rakshith-R (approved):

examples/kms/vault/vault.yaml
nixpanic: The `creationTimestamp` gets set automatically if missing, I think. In any case, it does not add any value for the example or tests.

Yuggupta27 (approved):

Madhu-1 (approved):

PR #2244, opened 2021-07-06
e2e: minor changes to allow running e2e on OpenShift

nixpanic:
Most of my manual testing I do on OpenShift. It is useful to run the e2e suite for
validation. However, running on OpenShift fails with different issues. This PR
addresses the problems, and adds arguments that make it possible to run the
tests on an OpenShift deployment with a Ceph cluster configured through Red
Hats OpenShift Container Storage.

This isn't making everything work out of the box though. Some changes to the
files in the `examples/` directory are still needed (like different drivername and
ServiceAccount). But things are getting much closer already.

---

<details>

  progress or results

</details>

### DIFF
--- a/e2e/rbd_helper.go
+++ b/e2e/rbd_helper.go
@@ -59,15 +59,19 @@ func createRBDStorageClass(
 	sc.Parameters["csi.storage.k8s.io/node-stage-secret-namespace"] = cephCSINamespace
 	sc.Parameters["csi.storage.k8s.io/node-stage-secret-name"] = rbdNodePluginSecretName
 
-	fsID, stdErr, err := execCommandInToolBoxPod(f, "ceph fsid", rookNamespace)
-	if err != nil {
-		return err
+	fsID := defaultFSID
+	if fsID == noFSID {
+		var stdErr string
+		fsID, stdErr, err = execCommandInToolBoxPod(f, "ceph fsid", rookNamespace)
+		if err != nil {
+			return err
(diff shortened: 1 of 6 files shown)


e2e/rbd_helper.go
```diff
@@ -59,15 +59,19 @@ func createRBDStorageClass(
...
+		var stdErr string
+		fsID, stdErr, err = execCommandInToolBoxPod(f, "ceph fsid", rookNamespace)
+		if err != nil {
+			return err
+		}
+		if stdErr != "" {
+			return fmt.Errorf("error getting fsid %v", stdErr)
+		}
+		// remove new line present in fsID
+		fsID = strings.Trim(fsID, "\n")
```
Yuggupta27, line 73: Should we trim regardless of whether or not FSID is passed?

nixpanic: I don't think that is needed, users will run `--fsid=openshift-storage` or similar, that usually would not contain additional spaces

nixpanic:
Hmm, this fails on Kubernetes now because the `runAsUser` has been dropped, will need to set a UID that works for both Kubernetes and OpenShift.

```
Failed: Error: container has runAsNonRoot and image will run as root (pod: "vault-bfdd5577b-fvllh_cephcsi-e2e-3a4167ea2926(545b370d-808a-4966-80fd-1e9b86f4a255)", container: vault)
```

PR #2258, opened 2021-07-08
rbd: add support to expand encrypted volume

### LINKED ISSUE #1469: Allow expansion on "encrypted volumes"
# Describe the bug #

At present the expansion of volumes are blocked if the volume is encrypted. Considering expansion is a valid operation on an existing PVC, we should allow it eventhough volume is encrypted. 

I am still not sure, we have to reencrypt the volume after expansion with the same key or we can use different key. But either way this operation can go through. 

@vasyl-purchel any thoughts on this ?

pkalever:

Previously in ControllerExpandVolume() we had a check for encrypted volumes and we use to fail for all expand requests on an encrypted volume. Also for Block VolumeMode PVCs NodeExpandVolume used to be ignored/skipped.                                                            
                                                                            
With these changes, we add support for the expansion of encrypted volumes. Also for raw Block VolumeMode PVCs with Encryption, we call NodeExpandVolume.
                                                                            
That said,                                                                  
With LUKS1, cryptsetup utility doesn't prompt for a passphrase on resizing the crypto mapper device. This is because LUKS1 devices don't use kernel keyring for volume keys.

Whereas, LUKS2 devices use kernel keyring for volume key by default, i.e. cryptsetup utility asks for a passphrase if it detects volume key was previously passed to dm-crypt via kernel keyring service, we are overriding the default by `--disable-keyring` option during cryptsetup open command. So that at the time of crypto mapper device resize we will not be prompted for any passphrase.

Fixes: #1469

### Other notable changes:

* fix bug in DeviceEncryptionStatus() which could lead to failures with unmap in the NodeUnstageVolume path for the encrypted volumes.
* Added testcases

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -597,6 +597,48 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("Resize Encrypted Block PVC and check Device size", func() {
+				err := deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				err = createRBDStorageClass(
+					f.ClientSet,
+					f,
+					defaultSCName,
+					nil,
+					map[string]string{"encrypted": "true"},
+					deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+
+				// FileSystem PVC resize
+				err = resizePVCAndValidateSize(pvcPath, appPath, f)
+				if err != nil {
+					e2elog.Failf("failed to resize filesystem PVC with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0, defaultRBDPool)
+
+				// Block PVC resize
+				err = resizePVCAndValidateSize(rawPvcPath, rawAppPath, f)
+				if err != nil {
+					e2elog.Failf("failed to resize block PVC with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0, defaultRBDPool)
+
+				err = deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+			})
+
 			By("create a PVC and bind it to an app with encrypted RBD volume with VaultKMS", func() {
 				err := deleteResource(rbdExamplePath + "storageclass.yaml")
 				if err != nil {
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -1355,12 +1355,6 @@ func (cs *ControllerServer) ControllerExpandVolume(
 		return nil, status.Error(codes.InvalidArgument, "capacityRange cannot be empty")
 	}
 
-	nodeExpansion := false
-	// Get the nodeexpansion flag set based on the volume mode
-	if req.GetVolumeCapability().GetBlock() == nil {
-		nodeExpansion = true
-	}
-
 	// lock out parallel requests against the same volume ID
 	if acquired := cs.VolumeLocks.TryAcquire(volID); !acquired {
 		util.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volID)
@@ -1368,14 +1362,6 @@ func (cs *ControllerServer) ControllerExpandVolume(
 		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volID)
 	}
 	defer cs.VolumeLocks.Release(volID)
-
-	// lock out volumeID for clone and delete operation
-	if err := cs.OperationLocks.GetExpandLock(volID); err != nil {
-		util.ErrorLog(ctx, err.Error())
-
-		return nil, status.Error(codes.Aborted, err.Error())
-	}
-	defer cs.OperationLocks.ReleaseExpandLock(volID)
 
 	cr, err := util.NewUserCredentials(req.GetSecrets())
 	if err != nil {
@@ -1399,10 +1385,22 @@ func (cs *ControllerServer) ControllerExpandVolume(
 		return nil, err
 	}
 
-	if rbdVol.isEncrypted() {
-		return nil, status.Errorf(codes.InvalidArgument, "encrypted volumes do not support resize (%s)",
-			rbdVol)
+	// NodeExpansion is needed for PersistentVolumes with,
+	// 1. Filesystem VolumeMode with & without Encryption and
+	// 2. Block VolumeMode with Encryption
+	// Hence set nodeExpansion flag based on VolumeMode and Encryption status
+	nodeExpansion := true
+	if req.GetVolumeCapability().GetBlock() != nil && !rbdVol.isEncrypted() {
+		nodeExpansion = false
 	}
+
+	// lock out volumeID for clone and delete operation
+	if err = cs.OperationLocks.GetExpandLock(volID); err != nil {
+		util.ErrorLog(ctx, err.Error())
+
+		return nil, status.Error(codes.Aborted, err.Error())
+	}
+	defer cs.OperationLocks.ReleaseExpandLock(volID)
 
 	// always round up the request size in bytes to the nearest MiB/GiB
 	volSize := util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -875,38 +875,49 @@ func (ns *NodeServer) NodeExpandVolume(
 	}
 	defer ns.VolumeLocks.Release(volumeID)
 
-	devicePath, err := getDevicePath(ctx, volumePath)
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
-	}
-
-	// TODO check size and return success or error
-	volumePath += "/" + volumeID
-	resizer := mount.NewResizeFs(utilexec.New())
-	ok, err := resizer.Resize(devicePath, volumePath)
-	if !ok {
-		return nil, status.Errorf(codes.Internal, "rbd: resize failed on path %s, error: %v", req.GetVolumePath(), err)
-	}
-
-	return &csi.NodeExpandVolumeResponse{}, nil
-}
-
-func getDevicePath(ctx context.Context, volumePath string) (string, error) {
 	imgInfo, err := lookupRBDImageMetadataStash(volumePath)
 	if err != nil {
 		util.ErrorLog(ctx, "failed to find image metadata: %v", err)
 	}
-	device, found := findDeviceMappingImage(
+	devicePath, found := findDeviceMappingImage(
 		ctx,
 		imgInfo.Pool,
 		imgInfo.RadosNamespace,
 		imgInfo.ImageName,
 		imgInfo.NbdAccess)
-	if found {
-		return device, nil
+	if !found {
+		return nil, status.Errorf(codes.Internal,
+			"failed to get device for stagingtarget path %v", volumePath)
 	}
 
-	return "", fmt.Errorf("failed to get device for stagingtarget path %v", volumePath)
+	mapperFile, mapperPath := util.VolumeMapper(volumeID)
+	if imgInfo.Encrypted {
+		// The volume is encrypted, resize an active mapping
+		err = util.ResizeEncryptedVolume(ctx, mapperFile)
+		if err != nil {
+			util.ErrorLog(ctx, "failed to resize device %s, mapper %s: %w",
+				devicePath, mapperFile, err)
+
+			return nil, status.Errorf(codes.Internal,
+				"failed to resize device %s, mapper %s: %v", devicePath, mapperFile, err)
+		}
+		// Use mapper device path for fs resize
+		devicePath = mapperPath
+	}
+
+	if req.GetVolumeCapability().GetBlock() == nil {
+		// TODO check size and return success or error
+		volumePath += "/" + volumeID
+		resizer := mount.NewResizeFs(utilexec.New())
+		var ok bool
+		ok, err = resizer.Resize(devicePath, volumePath)
+		if !ok {
+			return nil, status.Errorf(codes.Internal,
+				"rbd: resize failed on path %s, error: %v", req.GetVolumePath(), err)
+		}
+	}
+
+	return &csi.NodeExpandVolumeResponse{}, nil
 }
 
 // NodeGetCapabilities returns the supported capabilities of the node server.
--- a/internal/util/crypto.go
+++ b/internal/util/crypto.go
@@ -294,7 +294,8 @@ func DeviceEncryptionStatus(ctx context.Context, devicePath string) (mappedDevic
 	if len(lines) < 1 {
 		return "", "", fmt.Errorf("device encryption status returned no stdout for %s", devicePath)
 	}
-	if !strings.HasSuffix(lines[0], " is active.") {
+	// The line will look like: "/dev/mapper/xxx is active and is in use."
+	if !strings.Contains(lines[0], " is active") {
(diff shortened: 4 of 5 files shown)


internal/util/crypto.go
```diff
@@ -28,8 +28,8 @@ import (
 )
 
 const (
-	mapperFilePrefix     = "luks-rbd-"
-	mapperFilePathPrefix = "/dev/mapper"
+	MapperFilePrefix     = "luks-rbd-"
```
nixpanic: Please explain in the commit comment why this is needed. It is every much internal to the encryption functions that are provided in this file. Depending on the need for exposing this, it might be cleaner to add more helper functions so a real API is defined. Exported functions and constants *should* have descriptive comments.

internal/util/crypto.go
```diff
@@ -286,7 +286,7 @@ func DeviceEncryptionStatus(ctx context.Context, devicePath string) (mappedDevic
 	if len(lines) < 1 {
 		return "", "", fmt.Errorf("device encryption status returned no stdout for %s", devicePath)
 	}
-	if !strings.HasSuffix(lines[0], " is active.") {
+	if !strings.Contains(lines[0], " is active") {
```
nixpanic, line 309: hmm, how did this ever work before?!

I like a comment above the comparing of the string, show the (single line) output of what is expected to be returned. Could be an improvement if you have to re-work this PR.

internal/util/crypto.go
```diff
@@ -296,7 +296,7 @@ func DeviceEncryptionStatus(ctx context.Context, devicePath string) (mappedDevic
 			return "", "", fmt.Errorf("device encryption status output for %s is badly formatted: %s",
 				devicePath, lines[i])
 		}
-		if strings.Compare(kv[0], "device") == 0 {
+		if strings.Contains(kv[0], "device") {
```
nixpanic: the value should be `"device"`, why not just compare `kv[0] == "device"`?

docs/deploy-rbd.md
```diff
@@ -63,6 +63,7 @@ make image-cephcsi
...
+| `encryptionType`                                                                                    | no                   | encryption type to use, currently supported types are luks1 and luks2 (defaults to luks2)                                                                                                                                                                                                                                                                                                                                                                                                               |
```
nixpanic: why is this needed? are there still systems around that do not support luks2?

nixpanic (commented):
Not sure if adding support for luks1 is needed, do we want to be compatible with older systems?

Might be good to add a test-case for a Filesystem PVC too, the order of extending block-device and filesystem is important.

internal/util/crypto.go
pkalever: I think, we already have a helper for that, I just discovered it. I will try to use it.

internal/util/crypto.go
pkalever: Sure, that should help.

internal/util/crypto.go
pkalever: Sorry, I initially had a different version. Later I forget to remove this, will fix it up.

docs/deploy-rbd.md
pkalever: It's the other way around. luks1 format remains fully supported. There should be a way to choose b/w available formats, hence it doesn't hurt to be flexible (when cryptosetup and kernel are ok with it, why not cephcsi?). Who knows in the future there could be other types (luks3 or whatever) :-)

pkalever (commented):
> Might be good to add a test-case for a Filesystem PVC too, the order of extending block-device and filesystem is important.

Sure, I will plan to add for FS too, do we want them in the same PR? I guess not.


Thanks @nixpanic, expect a respin soon.

docs/deploy-rbd.md
nixpanic: Adding more features adds to the maintenance burden, so I prefer to only add features when needed.

What version of LUKS is currently the default or standard? If LUKS2 is still new and LUKS1 is more common to be used, then we would need to support both. However, if LUKS1 is mostly replaced by LUKS2, then I do not see a need to support LUKS1.

pkalever: @nixpanic Cryptsetup currentrly supports different encryption operating modes to use with dm-crypt:

```
--type luks for using the default LUKS format version (LUKS1 with cryptsetup < 2.1.0, LUKS2 with cryptsetup ≥ 2.1.0),
--type luks1 for using LUKS1, the most common version of LUKS,
--type luks2 for using LUKS2, the latest available version of LUKS that allows additional extensions,
--type plain for using dm-crypt plain mode,
--type loopaes for a loopaes legacy mode,
--type tcrypt for a TrueCrypt compatibility mode.
```

We only tested ceph-csi with luks1 and luks2, so far. Also, its worth noting that luks1 remains stable, supported "forever"
Having these changes makes us flexible to opt b/w various format types.

nixpanic: I am mostly interested by the default version of LUKS used by cryptsetup on the distributions that we support with Ceph-CSI. We build on CentOS-8 now, and I guess if that provides LUKS2 by default, we do not need to add LUKS1.

The additional tests, and maintenance to make sure we don't break LUKS1 support might not be worth the benefit. If LUKS1 is the current default and we expect users to have provisioned volumes with it, we probably need to maintain support for it, unfortunately.

If there ever will be a need to support different encryption operation modes, I don't know, but I hope single one will be sufficient for us.

Madhu-1: Yes, I agree with the same related to support point of view we don't need to open more options to the user (testing backward compatibility, a regression can become a problem for cephcsi)

If you think this is required please feel free to open a feature request.

e2e/rbd.go
```diff
@@ -713,6 +713,152 @@ var _ = Describe("RBD", func() {
...
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+			})
+
+			By("Resize Encrypted Block PVC and check Device size", func() {
```
Madhu-1: can you please add a test for both filesystem and block and also checking PVC is encrypted or not validation is missing.

internal/rbd/nodeserver.go
```diff
@@ -789,6 +789,24 @@ func (ns *NodeServer) NodeExpandVolume(
...
+		// The volume is encrypted, resize an active mapping
+		err = util.ResizeEncryptedVolume(ctx, mapperFile)
+		if err != nil {
+			util.ErrorLog(ctx, "failed to resize device %s, mapper %s: %w",
+				devicePath, mapperFile, err)
+			return &csi.NodeExpandVolumeResponse{}, err
+		}
+		// Use mapper device path for fs resize
+		devicePath = mapperPath
+	}
```
Madhu-1: This is required only if the device is encrypted?  We are storing the encryption state in the metadata file, Have an if check here to do this only if it's encrypted.

pkalever: The above block is only executed if the device is encrypted right ? mapper is only set for the encrypted devices.

Or did I missed anything in your question?

e2e/rbd.go
pkalever: Sure, I will implement one for Block too

docs/deploy-rbd.md
pkalever: Well, I really didn't follow why we shouldn't support other encryption formats supported by cryptsetup and dm_encrypt, after all Ceph-CSI is operating at the very top of the stack.

Yes, this can be a feature request If that makes us happy here :-)

Will remove this commit in the later respin.

pkalever (commented):
Thanks, Madhu & Niels, expect a respin to address the comment.

internal/rbd/nodeserver.go
Madhu-1: Yes this works, my idea was to use something like existing https://github.com/ceph/ceph-csi/blob/276918db1080e8e42aca19e2ca02312e6e240c9d/internal/rbd/rbd_attach.go#L333-L351 to avoid running `DeviceEncryptionStatus` for non encrypted devices

pkalever: Maybe you are pointing to a different code block here.
But I understand your intention is to suggest getting the encryption status from the StashFile rather than running some real commands. Sure we can save some jiffies I guess.

nixpanic: Encrypted devices have some metadata set, you can run `rbdVolume.checkRbdImageEncrypted()` or similar. Unfortunately there is not `rbdVolume` object created in this function :disappointed:

nixpanic (commented):
some more comments :)

internal/rbd/nodeserver.go
pkalever: Yeah, let's check the stashFile :+1:

e2e/rbd.go
```diff
@@ -713,6 +713,106 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("create a PVC and bind it to an app encrypted RBD volume and validate LUKS format", func() {
```
Madhu-1: Is this extra test case needed?

internal/rbd/controllerserver.go
```diff
@@ -1324,10 +1311,18 @@ func (cs *ControllerServer) ControllerExpandVolume(
 		return nil, err
 	}
 
-	if rbdVol.isEncrypted() {
-		return nil, status.Errorf(codes.InvalidArgument, "encrypted volumes do not support resize (%s)",
-			rbdVol)
+	nodeExpansion := true
+	// Get the nodeexpansion flag set based on the volume mode
```
Madhu-1: can add more details to the comment on what we are doing here?

internal/util/crypto.go
```diff
@@ -255,6 +255,16 @@ func OpenEncryptedVolume(ctx context.Context, devicePath, mapperFile, passphrase
 	return err
 }
 
+// ResizeEncryptedVolume resizes encrypted volume so that it can be used by the client.
+func ResizeEncryptedVolume(ctx context.Context, mapperFile string) error {
+	DebugLog(ctx, "Resizing LUKS device %s", mapperFile)
+	_, stderr, err := LuksResize(mapperFile)
+	if err != nil {
+		WarningLog(ctx, "failed to resize LUKS device %s: %s", mapperFile, stderr)
```
Madhu-1: not sure about places this should be an error log, not the warning log.

Madhu-1 (commented):
no major comments.

e2e/rbd.go
pkalever: This was part of my previous changes, where I also had one for LUKS1 :-)
Thought we can reuse some part of the work. If you feel it adds no value we can drop, but just give it a thought please.

internal/rbd/controllerserver.go
pkalever: sure :+1:

internal/util/crypto.go
pkalever: This was inspired from OpenEncryptedVolume() function above.

Madhu-1: Can you please change this to error log. OpenEncryptedVolume() can be corrected later

pkalever: On it now.

pkalever: Done with both.

pkalever:
@Mergifyio rebase

pkalever:
@Mergifyio rebase

nixpanic:
golangci-lint failure needs to be addressed:
```
e2e/pod.go:220:33: `execCommandInContainer` - `cn` always receives `"csi-rbdplugin"` (unparam)
	f *framework.Framework, c, ns, cn string, opt *metav1.ListOptions) (string, string, error) {
	                               ^
```

internal/util/crypto.go
nixpanic: `OpenEncryptedVolume()` was moved to ErrorLog, this is still WarningLog?

internal/rbd/nodeserver.go
```diff
@@ -841,36 +841,46 @@ func (ns *NodeServer) NodeExpandVolume(
...
-	return "", fmt.Errorf("failed to get device for stagingtarget path %v", volumePath)
+
+	mapperFile, mapperPath := util.VolumeMapper(volumeID)
+	if imgInfo.Encrypted {
+		// The volume is encrypted, resize an active mapping
+		err = util.ResizeEncryptedVolume(ctx, mapperFile)
+		if err != nil {
+			util.ErrorLog(ctx, "failed to resize device %s, mapper %s: %w",
+				devicePath, mapperFile, err)
+			return &csi.NodeExpandVolumeResponse{}, err
```
nixpanic: This looks confusing to me. You return a `NodeExpandVolumeResponse{}` that indicates success, but also an error? If there is an error, return `nil` and a gRPC error with `status.Error*()`

internal/util/crypto.go
```diff
@@ -260,7 +260,7 @@ func ResizeEncryptedVolume(ctx context.Context, mapperFile string) error {
 	DebugLog(ctx, "Resizing LUKS device %s", mapperFile)
 	_, stderr, err := LuksResize(mapperFile)
 	if err != nil {
-		WarningLog(ctx, "failed to resize LUKS device %s: %s", mapperFile, stderr)
+		ErrorLog(ctx, "failed to resize LUKS device %s: %s", mapperFile, stderr)
```
nixpanic, line 269: ah, this part of the commit was expected in the previous one

e2e/rbd.go
nixpanic: I do not think this needs extra validation, we do not care what the default LUKS format is?

nixpanic (changes requested):

internal/util/crypto.go
pkalever: correct

internal/rbd/nodeserver.go
pkalever: good catch!

internal/util/crypto.go
pkalever: Yes, seem to have got confused with commits. adjusted now.

e2e/rbd.go
pkalever: Sure, ok to remove.

pkalever (commented):
Thanks @nixpanic

pkalever:
@Mergifyio rebase

Madhu-1:
@pkalever PR needs manual rebase

pkalever:
> @pkalever PR needs manual rebase

@Madhu-1 done now.

internal/rbd/nodeserver.go
```diff
@@ -875,38 +875,47 @@ func (ns *NodeServer) NodeExpandVolume(
...
+		// Use mapper device path for fs resize
+		devicePath = mapperPath
+	}
+
+	var ok bool
+	if req.GetVolumeCapability().GetBlock() == nil {
+		// TODO check size and return success or error
+		volumePath += "/" + volumeID
+		resizer := mount.NewResizeFs(utilexec.New())
+		ok, err = resizer.Resize(devicePath, volumePath)
```
Madhu-1, line 913: can we have a short variable declaration  here, if not possible have `ok` declaration inside if block

```suggestion
ok, err := resizer.Resize(devicePath, volumePath)
```

Madhu-1 (dismissed):
LGTM, small suggestion

internal/rbd/nodeserver.go
pkalever: I will better move the declaration inside. The suggested code might lead to shadow declaration of `err`?

pkalever: @Madhu-1 done now. Thanks!

Madhu-1 (dismissed):

nixpanic (dismissed):

nixpanic:
@mergifyio rebase

Madhu-1:
@mergifyio rebase

Madhu-1:
@pkalever need to fix golangci-lint issues

pkalever:
> @pkalever need to fix golangci-lint issues

Hopefully done now, it was `nlreturn`, thanks!

Madhu-1:

ETCD timeout

Madhu-1 (approved):

nixpanic:
> > /test ci/centos/upgrade-tests-rbd
> 
> ETCD timeout

How!? Please link to the logs next time so we can see if improving the CI jobs is possible.

nixpanic (approved):

PR #2268, opened 2021-07-11
cleanup: resolves wrapcheck linter

yati1998:
This linter resolves wrapcheck linter
issues.

Updates: #1586

### DIFF
--- a/e2e/configmap.go
+++ b/e2e/configmap.go
@@ -19,7 +19,7 @@ func deleteConfigMap(pluginPath string) error {
 	path := pluginPath + configMap
 	_, err := framework.RunKubectl(cephCSINamespace, "delete", "-f", path, ns)
 	if err != nil {
-		return err
+		return fmt.Errorf("delete configmap: %w", err)
 	}
 	return nil
 }
@@ -57,7 +57,7 @@ func createConfigMap(pluginPath string, c kubernetes.Interface, f *framework.Fra
 	conmap[0].CephFS.SubvolumeGroup = subvolumegroup
 	data, err := json.Marshal(conmap)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to marshal subvolume: %w", err)
 	}
 	cm.Data["config.json"] = string(data)
 	cm.Namespace = cephCSINamespace
@@ -73,9 +73,12 @@ func createConfigMap(pluginPath string, c kubernetes.Interface, f *framework.Fra
 	}
 	if apierrs.IsNotFound(err) {
 		_, err = c.CoreV1().ConfigMaps(cephCSINamespace).Create(context.TODO(), &cm, metav1.CreateOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to create configmap: %w", err)
+		}
 	}
 
-	return err
+	return nil
 }
 
 // createCustomConfigMap provides multiple clusters information.
@@ -110,7 +113,7 @@ func createCustomConfigMap(c kubernetes.Interface, pluginPath string, subvolgrpI
 	}
 	data, err := json.Marshal(conmap)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to marshal cluster info: %w", err)
 	}
 	cm.Data["config.json"] = string(data)
 	cm.Namespace = cephCSINamespace
--- a/e2e/rbd_helper.go
+++ b/e2e/rbd_helper.go
@@ -92,7 +92,7 @@ func createRBDStorageClass(
 	}
 	sc.ReclaimPolicy = &policy
 	_, err = c.StorageV1().StorageClasses().Create(context.TODO(), &sc, metav1.CreateOptions{})
-	return err
+	return fmt.Errorf("failed to create storageclass: %w", err)
 }
 
 func createRadosNamespace(f *framework.Framework) error {
@@ -149,7 +149,7 @@ func createRBDSecret(f *framework.Framework, secretName, userName, userKey strin
 	sc.StringData["userKey"] = userKey
 	sc.Namespace = cephCSINamespace
 	_, err = f.ClientSet.CoreV1().Secrets(cephCSINamespace).Create(context.TODO(), &sc, metav1.CreateOptions{})
-	return err
+	return fmt.Errorf("failed to create secret: %w", err)
 }
 
 type imageInfoFromPVC struct {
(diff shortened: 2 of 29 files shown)


yati1998:
We still have errors with wrap check linter. I Will make a different PR for that otherwise it will become difficult to review.

e2e/configmap.go
```diff
@@ -19,7 +19,7 @@ func deleteConfigMap(pluginPath string) error {
 	path := pluginPath + configMap
 	_, err := framework.RunKubectl(cephCSINamespace, "delete", "-f", path, ns)
 	if err != nil {
-		return err
+		return fmt.Errorf("delete configmap: %w", err)
```
Madhu-1, line 22: failed to delete configmap?

e2e/configmap.go
```diff
@@ -57,7 +57,7 @@ func createConfigMap(pluginPath string, c kubernetes.Interface, f *framework.Fra
 	conmap[0].CephFS.SubvolumeGroup = subvolumegroup
 	data, err := json.Marshal(conmap)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to marshal subvolume: %w", err)
```
Madhu-1, line 60: failed to marshal clusterInfo

e2e/rbd_helper.go
```diff
@@ -92,7 +92,7 @@ func createRBDStorageClass(
 	}
 	sc.ReclaimPolicy = &policy
 	_, err = c.StorageV1().StorageClasses().Create(context.TODO(), &sc, metav1.CreateOptions{})
-	return err
+	return fmt.Errorf("failed to create storageclass: %w", err)
```
Madhu-1, line 95: need `if err!=nil` check here? err can be nil

e2e/rbd_helper.go
```diff
@@ -149,7 +149,7 @@ func createRBDSecret(f *framework.Framework, secretName, userName, userKey strin
 	sc.StringData["userKey"] = userKey
 	sc.Namespace = cephCSINamespace
 	_, err = f.ClientSet.CoreV1().Secrets(cephCSINamespace).Create(context.TODO(), &sc, metav1.CreateOptions{})
-	return err
+	return fmt.Errorf("failed to create secret: %w", err)
```
Madhu-1, line 152: same if check is required `err` can be nil

e2e/snapshot.go
```diff
@@ -149,7 +149,7 @@ func createRBDSnapshotClass(f *framework.Framework) error {
 		return err
 	}
 	_, err = sclient.VolumeSnapshotClasses().Create(context.TODO(), &sc, metav1.CreateOptions{})
-	return err
+	return fmt.Errorf("failed to create volume snapshotclass: %w", err)
```
Madhu-1, line 152: same if check is required here also `err` can be nil

e2e/snapshot.go
```diff
@@ -160,7 +160,11 @@ func deleteRBDSnapshotClass() error {
 	if err != nil {
 		return err
 	}
-	return sclient.VolumeSnapshotClasses().Delete(context.TODO(), sc.Name, metav1.DeleteOptions{})
+	err = sclient.VolumeSnapshotClasses().Delete(context.TODO(), sc.Name, metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to delete volume snapshotclass: %w", err)
```
Madhu-1, line 165: `volume snapshotclass` to `volumesnapshotclass`

internal/cephfs/controllerserver.go
```diff
@@ -60,7 +60,7 @@ func (cs *ControllerServer) createBackingVolume(
 	if sID != nil {
 		if err = cs.OperationLocks.GetRestoreLock(sID.SnapshotID); err != nil {
 			util.ErrorLog(ctx, err.Error())
-			return status.Error(codes.Aborted, err.Error())
+			return fmt.Errorf("failed to get restore lock: %w", status.Error(codes.Aborted, err.Error()))
```
Madhu-1, line 63: as we are returning GRPC error message why this is required?

Madhu-1 (commented):
Haven't looked at the code where we are returning the GRPC error code.

internal/cephfs/clone.go
```diff
@@ -227,13 +228,13 @@ func (vo *volumeOptions) getCloneState(ctx context.Context, volID volumeID) (cep
 			vo.FsName,
 			string(volID),
 			err)
-		return cephFSCloneError, err
+		return cephFSCloneError, fmt.Errorf("failed to get FSAdmin: %w", err)
```
nixpanic, line 231: Include some more details with every error, here you can use `"failed to get FSAdmin for volume %q: %w", vo, err`. I recommend to do something similar like this for other functions too.

internal/cephfs/controllerserver.go
nixpanic, line 63: This might actually be problematic. The `status.Error()` returns a grpc-error, not sure if (or how) that is retained when it is re-formatted with `%w`. I really dislike it that some functions return grpc-errors, and others don't. IMHO only CSI related calls should return grpc-errors, and helper functions should stick to 'normal' error. Or at least helpers that return grpc-errors should have a comment at the top of the function about it.

Madhu-1: I think for now we can disable this static check and we can bring it back once we address returning proper error messages handling from internal and outer functions. @nixpanic any thought?

yati1998: @Madhu-1  @nixpanic We have many other similar changes. So shall I disable this linter or make further changes?
IMO we can merge the changes that don't have `status.Error()` and disable the linter. Later on, when we have proper error messages, we can work on this. If you want I can open an issue for the same or add TODO while disabling the linter.
Let me know how shall I proceed with this.

nixpanic: The `status.Error*()` calls should only be done in the `controllerserver.go` and `nodeserver.go` files, I think. If there is an annotation like `// nolint:...` that you can place in the top of those files, and enable the linter for everything else, that would be my preference.

You might need to cleanup some of the error returning, as I suspect that some other files return gRPC errors with `status.Error*()` too, those should become normal errors, and the calling function should convert them to gRPC errors. This might not be a trivial change though. Place such changes in their own commit (per function, or group of functions) and maybe even their own PRs.

yati1998: @nixpanic as there are other corner cases to be considered as well, disabling this for now and will add a TODO.

yati1998:
Closing this as it's not required.

PR #2271, opened 2021-07-12
cleanup: resolves gofumpt linter issues

yati1998:
This commit runs the command
"gofumpt -l -w ." to resolve the gofumpt
linter issues. Gofumpt is a modified fork
of gofmt. It enforces a stricter format than gofmt.

Updates: #1586


Madhu-1 (changes requested):
Remove vendor changes we don't need this one?

e2e/rbd.go
```diff
@@ -1945,10 +1944,9 @@ var _ = Describe("RBD", func() {
 						}
 					}()
 
-					var pvc = &v1.PersistentVolumeClaim{}
-					pvc, err = loadPVC(pvcPath)
-					if err != nil {
-						e2elog.Failf("failed to load PVC with error %v", err)
+					pvc, pvcErr := loadPVC(pvcPath)
```
nixpanic, line 1947: I'm surprised the change of naming `err` is suggested. It seems this is the only location where that is done?

nixpanic (changes requested):
no concerns, except for the `vendor/` directory changes.

It might be simpler to split this PR into multiple smaller ones in order to reduce the number of conflicts it might cause. Consider doing that in case you notice it needs several manual rebases.

e2e/rbd.go
yati1998: No No. Actually the static-check was failing for this code, as the `err` here was overlapping the one declared above.

PR #2272, opened 2021-07-13
rbd: rework image chain iteration

### LINKED ISSUE #2144: cleanup: rework approach to traverse rbd image chain
In current approach for iteration of chained images in functions like [getCloneDepth()](https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/rbd_util.go#L530) and [checkImageChainHasFeature()](https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/rbd_util.go#L701),
we re-use and update an rbdVolume with different attributes which is quite wrong.
https://github.com/ceph/ceph-csi/blob/7b5c78ec7c36b98433d19956efba2bf061dc3428/internal/rbd/rbd_util.go#L563

Rework the loop such that:
we create and destroy a new volume inside the loop and use it for iteration and not just change parameters of the existing volume.

With this, we might be able to reduce complexity and it will minimize the redundant code in both functions.

See https://github.com/ceph/ceph-csi/pull/1838#discussion_r598530807
cc @nixpanic

Yuggupta27:
Improve the image chain iteration by creating and
deleting a volume inside the loop and using it for
traversing, in place of just updating parameters of
the existing volume.

Fixes: #2144 

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -593,6 +593,31 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 	return nil
 }
 
+// getParentVolume returns the parent volume of the caller volume.
+func (rv *rbdVolume) getParentVolume() (*rbdVolume, error) {
+	parentVol := rbdVolume{}
+	rbdImage, err := rv.open()
+	if err != nil {
+		return nil, err
+	}
+	defer rbdImage.Close()
+	parentInfo, err := rbdImage.GetParent()
+	if err != nil {
+		// Caller should decide whether not finding
+		// the parent is an error or not.
+		if errors.Is(err, librbd.ErrNotFound) {
+			return &parentVol, nil
+		}
+		return nil, err
+	}
+	parentVol.RbdImageName = parentInfo.Image.ImageName
+	parentVol.Pool = parentInfo.Image.PoolName
+	parentVol.Monitors = rv.Monitors
+	parentVol.RadosNamespace = rv.RadosNamespace
+	parentVol.conn = rv.conn.Copy()
+	return &parentVol, nil
+}
+
 func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
 	var depth uint
 	vol := rbdVolume{}
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -820,21 +820,17 @@ func (rv *rbdVolume) checkImageChainHasFeature(ctx context.Context, feature uint
 	vol.Monitors = rv.Monitors
 	vol.RbdImageName = rv.RbdImageName
 	vol.conn = rv.conn.Copy()
+	currVol := &vol
 
 	for {
-		if vol.RbdImageName == "" {
+		if currVol.RbdImageName == "" {
 			return false, nil
 		}
-		err := vol.openIoctx()
+		err := currVol.openIoctx()
 		if err != nil {
 			return false, err
 		}
-
-		err = vol.getImageInfo()
-		// FIXME: create and destroy the vol inside the loop.
-		// see https://github.com/ceph/ceph-csi/pull/1838#discussion_r598530807
-		vol.ioctx.Destroy()
-		vol.ioctx = nil
+		parentVol, err := currVol.getParentVolume()
 		if err != nil {
 			// call to getImageInfo returns the parent name even if the parent
 			// is in the trash, when we try to open the parent image to get its
@@ -843,15 +839,14 @@ func (rv *rbdVolume) checkImageChainHasFeature(ctx context.Context, feature uint
 			if errors.Is(err, ErrImageNotFound) {
 				return false, nil
 			}
-			util.ErrorLog(ctx, "failed to get image info for %s: %s", vol.String(), err)
-
+			util.ErrorLog(ctx, "failed to get parent image info for %s: %s", currVol.String(), err)
 			return false, err
 		}
-		if f := vol.hasFeature(feature); f {
+		defer parentVol.Destroy()
+		if f := currVol.hasFeature(feature); f {
 			return true, nil
 		}
-		vol.RbdImageName = vol.ParentName
-		vol.Pool = vol.ParentPool
+		currVol = parentVol
 	}
 }
 
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -595,27 +595,35 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 
 // getParentVolume returns the parent volume of the caller volume.
 func (rv *rbdVolume) getParentVolume() (*rbdVolume, error) {
-	parentVol := rbdVolume{}
-	rbdImage, err := rv.open()
+	rbdImageConn, err := rv.open()
 	if err != nil {
 		return nil, err
 	}
-	defer rbdImage.Close()
-	parentInfo, err := rbdImage.GetParent()
+	defer rbdImageConn.Close()
+	parentInfo, err := rbdImageConn.GetParent()
 	if err != nil {
 		// Caller should decide whether not finding
 		// the parent is an error or not.
 		if errors.Is(err, librbd.ErrNotFound) {
-			return &parentVol, nil
+			return &rbdVolume{}, nil
 		}
 		return nil, err
 	}
-	parentVol.RbdImageName = parentInfo.Image.ImageName
-	parentVol.Pool = parentInfo.Image.PoolName
-	parentVol.Monitors = rv.Monitors
-	parentVol.RadosNamespace = rv.RadosNamespace
-	parentVol.conn = rv.conn.Copy()
-	return &parentVol, nil
+	// parentVol.RbdImageName = parentInfo.Image.ImageName
+	// parentVol.Pool = parentInfo.Image.PoolName
+	// parentVol.Monitors = rv.Monitors
+	// parentVol.RadosNamespace = rv.RadosNamespace
+	// parentVol.conn = rv.conn.Copy()
+	// return &parentVol, nil
+	return &rbdVolume{
+		rbdImage: rbdImage{
+			RbdImageName:   parentInfo.Image.ImageName,
+			Pool:           parentInfo.Image.PoolName,
+			Monitors:       rv.Monitors,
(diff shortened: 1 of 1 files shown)


internal/rbd/rbd_util.go
```diff
@@ -570,6 +570,30 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
 }
 
+func (rv *rbdVolume) getImageParent() (*rbdVolume, error) {
+	parentVol := rbdVolume{}
+	rbdImage, err := rv.open()
+	if err != nil {
+		return nil, err
+	}
+	parentInfo, err := rbdImage.GetParent()
+	if err != nil {
```
Madhu-1, line 604:
```suggestion
if err != nil {
		// Caller should decide whether not finding
		// the parent is an error or not.
		if errors.Is(err, librbd.ErrNotFound) {
                       // parentVol.RbdImageName="" by default?
			return &parentVol, nil
		}
		return nil, err

}
parentVol.RbdImageName = parentInfo.Image.ImageName
parentVol.Pool = parentInfo.Image.PoolName
parentVol.Monitors = rv.Monitors
parentVol.conn = rv.conn.Copy()
return &parentVol, nil
```

Yuggupta27: Thanks for the suggestion! Updated with changes.

internal/rbd/rbd_util.go
```diff
@@ -578,36 +600,34 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
...
-
-		err = vol.getImageInfo()
-		// FIXME: create and destroy the vol inside the loop.
-		// see https://github.com/ceph/ceph-csi/pull/1838#discussion_r598530807
-		vol.ioctx.Destroy()
-		vol.ioctx = nil
+		parentVol, err := currVol.getImageParent()
+		defer parentVol.Destroy()
+		currVol.ioctx.Destroy()
+		currVol.ioctx = nil
```
Madhu-1: why this is required here?

internal/rbd/rbd_util.go
```diff
@@ -578,36 +600,34 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
...
 			return depth, err
 		}
-
-		err = vol.getImageInfo()
-		// FIXME: create and destroy the vol inside the loop.
-		// see https://github.com/ceph/ceph-csi/pull/1838#discussion_r598530807
-		vol.ioctx.Destroy()
-		vol.ioctx = nil
+		parentVol, err := currVol.getImageParent()
+		defer parentVol.Destroy()
```
Madhu-1: incase of `err` you are returning nil for parentVol calling `nil.Destory` will panic?

Madhu-1 (changes requested):

internal/rbd/rbd_util.go
```diff
@@ -570,6 +570,28 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 	return nil
 }
 
+func (rv *rbdVolume) getImageParent() (*rbdVolume, error) {
```
nixpanic: Please document what it does. All functions ideally have a comment.

Maybe rename to `getParentVolume()`?

internal/rbd/rbd_util.go
```diff
@@ -600,36 +600,34 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
...
-
+	currVol := vol
+	defer currVol.Destroy()
 	for {
-		if vol.RbdImageName == "" {
+		if currVol.RbdImageName == "" {
 			return depth, nil
 		}
-		err := vol.openIoctx()
+		err := currVol.openIoctx()
```
nixpanic, line 643: Why is opening the iocontext needed? I expect that to be done at the place where it is used.

internal/rbd/rbd_util.go
nixpanic: Don't destroy the iocontext here, it should automatically be done when the rbdVolume is not needed anymore. If there is a leak, the `defer currVol.Destroy()` needs to be placed at the tight spot (somewhere inside the for loop).

internal/rbd/rbd_util.go
```diff
@@ -600,36 +600,34 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
...
+			util.ErrorLog(ctx, "failed to check depth on image %s: %s", &currVol, err)
 			return depth, err
 		}
-		if vol.ParentName != "" {
+		if parentVol.RbdImageName != "" {
 			depth++
 		}
-		vol.RbdImageName = vol.ParentName
-		vol.Pool = vol.ParentPool
+		currVol = *parentVol
```
nixpanic, line 661: The previous `currVol` is no referenced here anymore, and gets garbage collected.

internal/rbd/rbd_util.go
```diff
@@ -600,36 +600,34 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
 	vol.Monitors = rv.Monitors
 	vol.RbdImageName = rv.RbdImageName
 	vol.conn = rv.conn.Copy()
-
+	currVol := vol
```
nixpanic, line 637: maybe assign `&vol` so that `currVol` becomes a `*rbdVolume` and you can assign `parentVol` directly to it?

internal/rbd/rbd_util.go
```diff
@@ -790,21 +790,22 @@ func (rv *rbdVolume) checkImageChainHasFeature(ctx context.Context, feature uint
 	vol.Monitors = rv.Monitors
 	vol.RbdImageName = rv.RbdImageName
 	vol.conn = rv.conn.Copy()
+	currVol := vol
```
nixpanic, line 832: basically the same comments as the previous commit

nixpanic (changes requested):

internal/rbd/rbd_util.go
Yuggupta27: Sure

internal/rbd/rbd_util.go
Yuggupta27: Done

internal/rbd/rbd_util.go
Yuggupta27: Nice catch, now Destroy is called only if err is nil

internal/rbd/rbd_util.go
```diff
@@ -603,36 +603,31 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
 	vol.Monitors = rv.Monitors
 	vol.RbdImageName = rv.RbdImageName
 	vol.conn = rv.conn.Copy()
-
+	currVol := &vol
```
nixpanic: use `vol := &rbdVolume{}` instead?

There needs to be a `defer vol.Destroy()` somewhere (or `currVol.Destroy()`, but `vol.Destroy()` will be less confusing).

internal/rbd/rbd_util.go
nixpanic: I still do not understand why this is needed. Is it required for calling `getParentVolume()`? If so, it should be done in the function there.

internal/rbd/rbd_util.go
```diff
@@ -603,36 +603,31 @@ func (rv *rbdVolume) getCloneDepth(ctx context.Context) (uint, error) {
...
+		parentVol, err := currVol.getParentVolume()
 		if err != nil {
 			// if the parent image is moved to trash the name will be present
 			// in rbd image info but the image will be in trash, in that case
 			// return the found depth
 			if errors.Is(err, ErrImageNotFound) {
 				return depth, nil
 			}
-			util.ErrorLog(ctx, "failed to check depth on image %s: %s", &vol, err)
+			util.ErrorLog(ctx, "failed to check depth on image %s: %s", &currVol, err)
```
nixpanic, line 655: `currVol` is already a `*rbdVolume`, using `&currVol` will likely print the pointer to `currVol` and not the result of `currVol.String()` as expected

nixpanic (changes requested):

internal/rbd/rbd_util.go
```diff
@@ -570,6 +570,31 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
+			return &parentVol, nil
+		}
+		return nil, err
+	}
+	parentVol.RbdImageName = parentInfo.Image.ImageName
+	parentVol.Pool = parentInfo.Image.PoolName
+	parentVol.Monitors = rv.Monitors
+	parentVol.RadosNamespace = rv.RadosNamespace
+	parentVol.conn = rv.conn.Copy()
+	return &parentVol, nil
```
pkalever:
```suggestion
	return &rbdVolume{
               rbdImage: rbdImage{
                       RbdImageName:   parentInfo.Image.ImageName,
                       Pool:           parentInfo.Image.PoolName,
                       Monitors:       rv.Monitors,
                       RadosNamespace: rv.RadosNamespace,
                       conn:           rv.conn.Copy(),
               },
       }, nil
```

internal/rbd/rbd_util.go
Madhu-1: Is this needed because the parent and the image pool are different? the openIoctx is per pool?

internal/rbd/rbd_util.go
```diff
@@ -793,14 +809,14 @@ func (rv *rbdVolume) checkImageChainHasFeature(ctx context.Context, feature uint
 			if errors.Is(err, ErrImageNotFound) {
 				return false, nil
 			}
-			util.ErrorLog(ctx, "failed to get image info for %s: %s", vol.String(), err)
+			util.ErrorLog(ctx, "failed to get parent image info for %s: %s", currVol.String(), err)
 			return false, err
 		}
-		if f := vol.hasFeature(feature); f {
+		defer parentVol.Destroy()
```
Madhu-1: defer in for loop as gotchas https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01

Madhu-1:
@Yuggupta27 am reopening this PR as this is still a pending one.

Madhu-1:
@Yuggupta27 any update here?

Yuggupta27:
Hi, I would not be able to currently persue this improvement due to other commitments. Closing this for now.

PR #2275, opened 2021-07-13
doc: add design doc for volume healer

### LINKED ISSUE #667: Qualify or move RBD NBD to full support

pkalever:

The Volume Healer at its minimal form.

Please note this is just a reference for someone new to volume healer, you can build more context by familiarizing yourself with the code that we have already pushed at https://github.com/ceph/ceph-csi/pull/2108


pkalever:
@humblec PTAL, thanks!

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+  the criteria of PV Bound, mounter type
+- Build the StagingPath where rbd images PVC is mounted, based on the
+  KUBELET path and PV object
+- Construct the NodeStageVolume() request and send Request to CSI Driver.
+- The NodeStageVolume() has a way to identify calls received from the
+  healer and when executed from the healer context, it just runs in the
+  minimal required form, where it fetches the previously mapped device to
+  the image, and the respective secrets and finally ensures to bringup the
+  respective process back to life. Thus enabling IO to continue.
+
```
humblec, line 82: @gnufied @jsafrane as part of addressing `disconnect` of volumes with the user space mounter in ceph side, we are planning to introduce a healer in our driver. However just to make sure we are *not* breaking any workflow of kube CSI client with this addition, we would like to get a second pair of eyes on this. That said, the healer look for `va` objects and run or file `modified` `nodestage calls while the plugin container comes up.   Above list the workflow of healer some more in detail with our current design. 

Do you foresee any issues with this approach ?  please feel free to ask if you have any extra questions on this :).. Thanks in advance.

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+rbd driver to mount the rbd images onto the application node. Here on
+at cephcsi we will also have a userspace way of mounting the RBD images,
+via RBD-NBD.
+
+[Rbd-nbd](https://docs.ceph.com/en/latest/man/8/rbd-nbd/) is a client for
+RADOS block device (rbd) images like the existing rbd kernel module. It
+will map an rbd image to an NBD (Network Block Device) device, allowing
+access to it as a regular local block device.
+
+![csi-rbdnbd](https://user-images.githubusercontent.com/12432241/125400940-23e75d00-e3d0-11eb-806c-bab09318de54.png)
```
nixpanic: include the images in this PR too please, ideally in SVG format

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
+# RBD NBD VOLUME HEALER
+
+- [RBD NBD VOLUME HEALER](#rbd-nbd-volume-healer)
+  - [Rbd Nbd](#rbd-nbd)
+    - [Advantages of userspace mounters](#advantages-of-userspace-mounters)
+    - [Side effects of userspace mounters](#side-effects-of-userspace-mounters)
+  - [Volume Healer](#volume-healer)
+    - [More thoughts](#more-thoughts)
+
+## Rbd Nbd
```
nixpanic: You have a mix of Rbd, RBD and rbd. Maybe settle on one form of capitalization? Same for NBD.

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+access to it as a regular local block device.
+
+![csi-rbdnbd](https://user-images.githubusercontent.com/12432241/125400940-23e75d00-e3d0-11eb-806c-bab09318de54.png)
+
+It’s worth making a note that the rbd-nbd processes will run on the
+client-side, which is inside the `csi-rbdplugin` node plugin.
+
+### Advantages of userspace mounters
+
+- Adding changes to client logic to accommodate new features will be easy
```
nixpanic: I assume that this is out of scope for ceph-csi, make it clear that it is for the Ceph project, compare to krbd?

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+- Container upgrades will be independent of the host node, which means if
```
nixpanic, line 33: rebooting is not needed now either, we expect `rbd.ko` to be available on the host, just like `nbd.ko`.

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+  there any changes in the client logic, we don’t have to reboot the node
+  as the changes will be shipped inside the container.
+- We will be a better citizen in K8s by switching to the userspace model.
```
nixpanic: really? why is this better?

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+- Unlike krbd, rbd-nbd uses librbd user-space library that gets most of the
+  development focus, and hence rbd-nbd will be future rich.
+- Unlike rbd-nbd, krbd exhibits additional drawback, being entirely kernel
+  space impacts fault-tolerance as any kernel panic affects a whole node
```
nixpanic: the NBD kernel module has this same problem?

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+    - [Side effects of userspace mounters](#side-effects-of-userspace-mounters)
+  - [Volume Healer](#volume-healer)
+    - [More thoughts](#more-thoughts)
+
+## Rbd Nbd
+
+The RBD CSI plugin will provision new RBD images and attach and mount those
+to workloads. Currently, the default mounter is krbd, which uses the kernel
+rbd driver to mount the rbd images onto the application node. Here on
+at cephcsi we will also have a userspace way of mounting the RBD images,
```
nixpanic: use Ceph-CSI please

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Advantages of userspace mounters
+
+- Adding changes to client logic to accommodate new features will be easy
+- Container upgrades will be independent of the host node, which means if
+  there any changes in the client logic, we don’t have to reboot the node
+  as the changes will be shipped inside the container.
+- We will be a better citizen in K8s by switching to the userspace model.
+- Unlike krbd, rbd-nbd uses librbd user-space library that gets most of the
+  development focus, and hence rbd-nbd will be future rich.
```
nixpanic: **will be _more feature_ rich** (not future)

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,94 @@
...
+
+Since the rbd-nbd processes run per volume map on the client side i.e.
+inside the `csi-rbdplugin` node plugin, a restart of the node plugin will
+terminate all the rbd-nbd processes, and there is no way to restore
+these processes back to life currently, which could lead to IO errors
+on all the application pods.
+
+![nodeplugin-restart](https://user-images.githubusercontent.com/12432241/125401399-ba1b8300-e3d0-11eb-9db2-e00f11a96af8.png)
+
+This is where the volume Healer could help.
```
nixpanic: **V**olume **H**ealer?

docs/design/proposals/rbd-volume-healer.md
pkalever: If that confused you, please feel free to suggest a betterment, I mean to say:

If there are any new features in the rbd, for example, RBD_FEATURE_LAYERING, RBD_FEATURE_STRIPINGV2, RBD_FEATURE_EXCLUSIVE_LOCK, RBD_FEATURE_OBJECT_MAP, RBD_FEATURE_FAST_DIFF, RBD_FEATURE_DEEP_FLATTEN, etc

One has to implement this in the Linux kernel rbd driver and wait for the feature to be available in the kernel, once they are available one has to upgrade the kernel and reboot the host node. Thankfully this is not needed for userspace mounters as we get them in userspace via librbd without needing to reboot the node.

docs/design/proposals/rbd-volume-healer.md
pkalever: I hope my previous comment summarizes it well.

docs/design/proposals/rbd-volume-healer.md
pkalever: Yes, but I see it's mostly unchanged much from years and I consider it to be much stable now.
NBD is much generic and there are many eyes on it compared to the rbd driver, also there won't be any vendor-specific logic inside the driver.

docs/design/proposals/rbd-volume-healer.md
pkalever: OK

docs/design/proposals/rbd-volume-healer.md
pkalever: oops! ok

docs/design/proposals/rbd-volume-healer.md
pkalever: Yes

docs/design/proposals/rbd-volume-healer.md
pkalever: users who don't have much understanding of the ceph or storage, might not understand the advantages of userspace mounter. Does it hurt to capture these details as an intro to `what and why userspace mounter`?

docs/design/proposals/rbd-volume-healer.md
pkalever: OK

docs/design/proposals/rbd-volume-healer.md
pkalever: What benefit will it serve here?

pkalever: I have refreshed the PR to reflect the requested changes, but I'm still curious to know why we wanted to persist them locally.

nixpanic: If we ever need to adjust the image, or do something similar, having the source available for the image is useful

pkalever:
@humblec @nixpanic @Madhu-1 given the feature is landed already, can we pull this?

docs/design/proposals/rbd-volume-healer.md
nixpanic: it surely does not hurt, but currently the text is not clear where it becomes easier to accommodate new features. You mean that it is easier to add features to rbd-nbd as that is released regularly with Ceph, and more difficult to add features to the kernel rbd module as that is part of the Linux kernel release schedule? (I understand it is about adding features to rbd-nbd here, and not ceph-csi, but that might be clarified a little too.)

docs/design/proposals/rbd-volume-healer.md
nixpanic: sure, I get that part, but that is just repeating the previous point? 'container upgrades' would be ceph-csi new version, I guess, so mentioning something about new rbd features might help.

docs/design/proposals/rbd-volume-healer.md
nixpanic: if that is a follow-up on the previous point, maybe add `because of that ...`

docs/design/proposals/rbd-volume-healer.md
nixpanic: please mention that in the text too then :)

nixpanic (changes requested):
I'd like to get this included. Added some clarification on my comments. When a reader is not intimately familiar with RBD and Ceph-CSI, some of the statements will likely be confusing and appear incomplete.

pkalever:
> I'd like to get this included. Added some clarification on my comments. When a reader is not intimately familiar with RBD and Ceph-CSI, some of the statements will likely be confusing and appear incomplete.

I agree @nixpanic, hope with the latest respin it's better now.

nixpanic (dismissed):
the codespell github action is preventing this from getting merged, you may want to add a commit that adjusts the configuration in `.github/workflows/codespell.yaml` to skip the ` docs/design/proposals/images` directory.

pkalever:
> the codespell github action is preventing this from getting merged, you may want to add a commit that adjusts the configuration in `.github/workflows/codespell.yaml` to skip the ` docs/design/proposals/images` directory.

Addressed the same, thanks @nixpanic

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,102 @@
...
+    - [Advantages of userspace mounters](#advantages-of-userspace-mounters)
+    - [Side effects of userspace mounters](#side-effects-of-userspace-mounters)
+  - [Volume Healer](#volume-healer)
+    - [More thoughts](#more-thoughts)
+
+## Rbd nbd
+
+The rbd CSI plugin will provision new rbd images and attach and mount those
+to workloads. Currently, the default mounter is krbd, which uses the kernel
+rbd driver to mount the rbd images onto the application node. Here on
```
humblec: its pod

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,102 @@
...
+  - If the NodeUnstage fails with an operation already going on, in the
+    next NodeUnstage the volume gets unmounted
+- What if the PVC is deleted?
+  - If the PVC is deleted, the volume attachment list might already got
+    refreshed and entry will be skipped/deleted at the healer.
+  - For any reason, If the request bails out with Error NotFound, skip the
+    PVC, assuming it might have deleted or the NodeUnstage might have
+    already happened.
+- The Volume healer currently works with rbd-nbd, but the design can
+  accomidate other userspace mounters (may be cepfs fuse).
```
humblec: may be good to have a note on its support level too.

Madhu-1: cepfs fuse/ cephfs fuse

docs/design/proposals/rbd-volume-healer.md
pkalever: OK

docs/design/proposals/rbd-volume-healer.md
Madhu-1: `accomidate` to `accommodate`

pkalever: We have https://github.com/ceph/ceph-csi/blob/devel/docs/rbd-nbd.md to discuss on the support matrix.

pkalever: I also think that the design doc is not a good fit for support level details.

pkalever: OK

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,102 @@
...
+  - If the NodeUnstage fails with an operation already going on, in the
+    next NodeUnstage the volume gets unmounted
+- What if the PVC is deleted?
+  - If the PVC is deleted, the volume attachment list might already got
+    refreshed and entry will be skipped/deleted at the healer.
+  - For any reason, If the request bails out with Error NotFound, skip the
+    PVC, assuming it might have deleted or the NodeUnstage might have
+    already happened.
+- The Volume healer currently works with rbd-nbd, but the design can
+  accommodate other userspace mounters (may be cepfs fuse).
```
Madhu-1: typo `cepfs fuse` missed this one?

pkalever: Somehow Github didn't reflect this (browser cache?), updating now.

Madhu-1 (dismissed):

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,102 @@
...
+- Container upgrades will be independent of the host node, which means if
+  there are any new features with rbd-nbd, we don’t have to reboot the node
+  as the changes will be shipped inside the container.
+- Because the container upgrades are host node independent, we will be a
+  better citizen in K8s by switching to the userspace model.
+- Unlike krbd, rbd-nbd uses librbd user-space library that gets most of the
+  development focus, and hence rbd-nbd will be feature-rich.
+- Unlike rbd-nbd, krbd exhibits additional drawback, being entirely kernel
+  space impacts fault-tolerance as any kernel panic affects a whole node
+  not only a single pod that is using rbd storage. Thanks to the rbd-nbd’s
```
humblec: considering rbd nbd running in the node, isnt this equally applicable for rbd nbd as well ?

docs/design/proposals/rbd-volume-healer.md
humblec: yeah , above support matrix is good enough.

humblec:
Any limitations to list out here in the design doc ?

docs/design/proposals/rbd-volume-healer.md
pkalever: There is a difference, the krbd is a complete kernel and vendor-specific driver which needs changes on every feature basis. On the other hand, NBD is a generic driver, while all the vendor-specific logic sits in the userspace.

humblec: I should have been more clear, I was pointing out `kernel panic` situation.

pkalever:
> Any limitations to list out here in the design doc ?

such as?

docs/design/proposals/rbd-volume-healer.md
pkalever: Do you see this as a better fit?

```
Being entirely kernel space impacts fault-tolerance as any kernel panic  
affects a whole node not only a single pod that is using rbd storage.    
Thanks to the rbd-nbd’s userspace design, we are less bothered here, the 
krbd is a complete kernel and vendor-specific driver which needs changes 
on every feature basis, on the other hand, rbd-nbd depends on NBD generic
driver, while all the vendor-specific logic sits in the userspace. It's  
worth taking note that NBD generic driver is mostly unchanged much from  
years and consider it to be much stable. Also given NBD is a generic     
driver there will be many eyes on it compared to the rbd driver.         
```

docs/design/proposals/rbd-volume-healer.md
gnufied: No amount of "healing" is going to fix potential data loss that could happen because application simply failed to write data to underlying mount point. Are we sure this is what we want?

gnufied: Not to mention, this would make driver very much kube specific. kinda defeats the purpose of CSI, no?

nixpanic (dismissed):

docs/design/proposals/rbd-volume-healer.md
humblec: @pkalever not sure, the only thing which I was pointing here is any `panic` situation in the kernel where mount happened equally applies the krbd or nbd as long as `nbd` runs inside the pod, does that make sense ? so I am not sure we have to specifically mention this `panic` case rather start from here directly .. ie ` krbd is a complete kernel and vendor-specific driver ....` , wdyt?

pkalever: The intention was to highlight the fact that rbd-nbd gets IO logic in userspace while krbd has to handle them in kernel completely. And there is an advantage for carrying the vendor-specific logic in the userspace rather than keeping it in the kernel space, frequently changing logic might be more prone to bugs which might lead to panics.

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,102 @@
...
+RADOS block device (rbd) images like the existing rbd kernel module. It
+will map an rbd image to an nbd (Network Block Device) device, allowing
+access to it as a regular local block device.
+
+![csi-rbd-nbd](./images/csi-rbd-nbd.svg)
+
+It’s worth making a note that the rbd-nbd processes will run on the
+client-side, which is inside the `csi-rbdplugin` node plugin.
+
+### Advantages of userspace mounters
```
jsafrane, line 28: This chapter trades stability (of kernel) for features (of rbd-nbd).

What a filesystem (xfs, ext4) does when rbd-nbd is not running at the time an application issues `write()` there? How long do they wait for any user space daemon to come back? What is the probability of loosing data and/or corrupting the filesystem?

IMO all write errors lead to remounting the volume as read-only and requiring remount (to replay logs and potentially fix the filesystem). Is this going to happen basically randomly when rbd-nbd pods are deleted and take too long to restart?

A block device managed by user space will be much more error prone, is it really worth the trade?

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,102 @@
...
+This is where the Volume healer could help.
+
+## Volume healer
+
+Volume healer runs on the start of rbd node plugin and runs within the
+node plugin driver context.
+
+Volume healer does the below,
+
+- Get the Volume attachment list for the current node where it is running
```
jsafrane, line 69: Why Ceph RBD uses attacher at all? It does not provide 3rd party attach.

docs/design/proposals/rbd-volume-healer.md
```diff
@@ -0,0 +1,102 @@
...
+- Filter the volume attachments list through matching driver name and
+  status attached
+- For each volume attachment get the respective PV information and check
+  the criteria of PV Bound, mounter type
+- Build the StagingPath where rbd images PVC is mounted, based on the
+  KUBELET path and PV object
```
jsafrane, line 75: This leaks Kubernetes implementation details to CSI drivers. CSI drivers should be able to run outside of Kubernetes.
Can you store a state of the driver on the host (say in `/var/lib/rbd-csi/*`, available in the driver pod as HostPath volume)? With a file for each rbd-nbd to restore after restart (or whatever you need, but let it be somewhat transactional to survive pod/machine restarts safely).

docs/design/proposals/rbd-volume-healer.md
humblec: @pkalever yeah, I get that, so this `krbd is a complete kernel and vendor specific ..` line is actually conveying that message.  Any way, please let me know if you would like just adjust above or go ahead with current version.  I am fine with any.

pkalever: Thanks @humblec, I have adjusted the previous version with the version mentioned in this comment.

humblec (dismissed):

docs/design/proposals/rbd-volume-healer.md
humblec: Thanks a lot @pkalever .. looks good :+1:

docs/design/proposals/rbd-volume-healer.md
jsafrane: +1, this feels very error prone. I like the kernel implementation, because it's rock stable.

humblec:
Adding `DNM` as we just received feedback from @jsafrane and @gnufied on this design and have to discuss on the same.

docs/design/proposals/rbd-volume-healer.md
humblec: @jsafrane True! and we really want to get rid of the attacher and that was one of my concern as raised here  ( Point 2) https://github.com/ceph/ceph-csi/pull/2108#issuecomment-875734208

docs/design/proposals/rbd-volume-healer.md
pkalever: Yes our CI already caught some of these problems when we initially injected few random tests. We are in discussion with the ceph core team if it is a good idea to freeze the filesystem as part of detach operation with the help of rbd-nbd quience hooks. Thereafter attempting to perform any modifications to the frozen fs will be blocked waiting until the fs is unfrozen which will be achieved by attaching the process back.

docs/design/proposals/rbd-volume-healer.md
pkalever: Given the changes are already merged and it's an alpha release, I will open an issue for this to make further improvements and target any better suggestions for the beta.

humblec:
Considering we have  very valid comments/concerns  here,   we have to continue discussions on this : 
To track these comments, here is the list which I think capture  those  comments.. 

1) I/O path concern and corruption possibilities which include both file and block mode: 
[comment 1](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677313425) , [comment 2 ]( https://github.com/ceph/ceph-csi/pull/2275/#discussion_r676706321) , [comment 3]( https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677331644)

2) Attacher dependency logic : 
[no need for attacher](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677314730)

3) Kubernetes implementation leak to CSI :  
[csi leak](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677318536)

Considering we already got the implementation based on this design merged and today is the `3.4.0` release cut, either we have to get this in and track these concerns in seperate issues or keep this design open and continue the discussions here.

@Madhu-1 @nixpanic thoughts?

docs/design/proposals/rbd-volume-healer.md
nixpanic: Using userspace clients for storage is definitely more complex than using the kernel equivalents. However, we would like to persuade this feature, as there are benefits as well. The current available feature (since the beginning of ceph-csi) has been labelled alpha. There are many improvements needed in both Ceph-CSI as well as in the Ceph rbd-nbd component. @pkalever is working on both, and https://github.com/ceph/ceph-csi/projects/13 contains the tasks that still are pending.

The main painpoint users can be affected with, is the cancelled I/O while the block-device is temporarily unavailable. This will cause filesystems to abort and remount themselves as read-only. There is no practical recovery from such a state. To prevent this, there are discussions with the Ceph team to run `fsfreeze` before detaching the userspace process, and unfreezing the filesystem afterwards. How this can be done for non-filesystem users remains to be seen (device-mapper can do it, maybe there is a generic syscall/ioctl). @pkalever, please create an issue so that we can track the Ceph discussions (add URLs to mails/... there).

Even with this volume healer, ceph-csi will keep the rbd-nbd feature in Alpha state. But, we do hope to get a little closer into gaining confidence to run real workloads on top of it.

docs/design/proposals/rbd-volume-healer.md
nixpanic: @pkalever please add a link to the discussion (ideally tracked in a ceph-csi issue)

docs/design/proposals/rbd-volume-healer.md
nixpanic: Placing the required details in our own host-path volume is not so much an issue, we can change that.

However, we would also require the credentials to access the Ceph cluster. In addition to security concerns, it is possible that these expire and are not updated (would get updated in the Kubernetes Secrets from the StorageClass). So at least some part of the healer is dependent on Kubernetes at the moment (calling the NodeStageVolume). We're all not happy about the Kubernetes dependency, but for a few other features we already have that as well :disappointed:

nixpanic:
> Considering we already got the implementation based on this design merged and today is the `3.4.0` release cut, either we have to get this in and track these concerns in seperate issues or keep this design open and continue the discussions here.

Nothing major is blocking the merging of this PR from my point of view. Any concerns should be collected as issues and worked on as part of the [rbd-nbd support project](https://github.com/ceph/ceph-csi/projects/13).

Madhu-1:
> Considering we have very valid comments/concerns here, we have to continue discussions on this :
> To track these comments, here is the list which I think capture those comments..
> 
> 1. I/O path concern and corruption possibilities which include both file and block mode:
>    [comment 1](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677313425) , [comment 2 ](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r676706321) , [comment 3](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677331644)
> 2. Attacher dependency logic :
>    [no need for attacher](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677314730)
> 3. Kubernetes implementation leak to CSI :
>    [csi leak](https://github.com/ceph/ceph-csi/pull/2275/#discussion_r677318536)
> 
> Considering we already got the implementation based on this design merged and today is the `3.4.0` release cut, either we have to get this in and track these concerns in seperate issues or keep this design open and continue the discussions here.
> 
> @Madhu-1 @nixpanic thoughts?

Not a blocker for now we can tracks these in different issues in nbd board

Madhu-1 (approved):

docs/design/proposals/rbd-volume-healer.md
pkalever: > What a filesystem (xfs, ext4) does when rbd-nbd is not running at the time an application issues write() there? How long do they wait for any user space daemon to come back?

When rbd-nbd process is terminated, the IO will wait until the nbd NBD_ATTR_TIMEOUT is reached. NBD_ATTR_TIMEOUT can be set by `--io-timeout` option of rbd-nbd tool, It specifies how long the IO should wait to get handled before bailing out. When we set the value to zero, it should keep the requeueing logic enabled, Specifying `--io-timeout` 0 tells the nbd driver to not abort the request and instead see if it can be restarted on another socket.

We can set NBD_ATTR_DEAD_CONN_TIMEOUT which controls, how long the device should be held waiting for the userspace process to come back to life. This can be set with rbd-nbd option `--reattach-timeout`

> What is the probability of loosing data and/or corrupting the filesystem?
> IMO all write errors lead to remounting the volume as read-only and requiring remount (to replay logs and potentially fix the filesystem). Is this going to happen basically randomly when rbd-nbd pods are deleted and take too long to restart?

Data/Metadata loss will probably happen only when the nodeplugin pod doesn't come back in time after a restart/upgrade, again this should be extremely rare IMHO.

> A block device managed by userspace will be much more error prone, is it really worth the trade?

I agree, but we can make it work better for sure, for example, gluster-block storage (openshift container storage 3) is based on tcmu-runner userspace block service and our users are happy with it.

PR #2287, opened 2021-07-14
ci: do not use Groovy string interpolation for credentials

nixpanic:
Jenkins warns in the output of CI jobs about the following:

    Warning: A secret was passed to "sh" using Groovy String interpolation, which is insecure.
        Affected argument(s) used the following variable(s): [CREDS_PASSWD, CREDS_USER]
        See https://jenkins.io/redirect/groovy-string-interpolation for details.

Variable with 'single quotes' and without the {curly brackets} are
expecred to not be affected. There is some indirection in the strings
passed to the `sh` function, so this approach might not fix it?

---

<details>

  progress or results

</details>

### DIFF
--- a/ci-job-validation.groovy
+++ b/ci-job-validation.groovy
@@ -8,8 +8,26 @@ def git_since = 'ci/centos'
 def base = ''
 def doc_change = 0
 // private, internal container image repository
-def cached_image = 'registry-ceph-csi.apps.ocp.ci.centos.org/ceph-csi'
+def ci_registry = 'registry-ceph-csi.apps.ocp.ci.centos.org'
+def cached_image = 'ceph-csi'
 def use_pulled_image = 'USE_PULLED_IMAGE=yes'
+
+def ssh(cmd) {
+	sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} '${cmd}'"
+}
+
+def podman_login(registry, username, passwd) {
+	ssh "podman login --authfile=~/.podman-auth.json --username=${username} --password='${passwd}' ${registry}"
+}
+
+// podman_pull pulls image from the source (CI internal) registry, and tags it
+// as unqualified image name and into the destination registry. This prevents
+// pulling from the destination registry.
+//
+// Images need to be pre-pushed into the source registry, though.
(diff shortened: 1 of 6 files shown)


ci-job-validation.groovy
```diff
@@ -8,9 +8,27 @@ def git_since = 'ci/centos'
...
+def ci_registry = 'registry-ceph-csi.apps.ocp.ci.centos.org'
+def cached_image = 'ceph-csi'
 def use_pulled_image = 'USE_PULLED_IMAGE=yes'
 
+def ssh(cmd) {
+	sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} '${cmd}'"
+}
+
+def podman_login(registry, username, passwd) {
+	ssh "podman login --authfile=~/.podman-auth.json --username=${username} --password='${passwd}' ${registry}"
```
Madhu-1, line 20: `'` is not used for username and registry. `'` is required for password?

ci-job-validation.groovy
```diff
@@ -8,9 +8,27 @@ def git_since = 'ci/centos'
 def base = ''
 def doc_change = 0
 // private, internal container image repository
-def cached_image = 'registry-ceph-csi.apps.ocp.ci.centos.org/ceph-csi'
+def ci_registry = 'registry-ceph-csi.apps.ocp.ci.centos.org'
+def cached_image = 'ceph-csi'
 def use_pulled_image = 'USE_PULLED_IMAGE=yes'
 
+def ssh(cmd) {
+	sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} '${cmd}'"
```
Madhu-1, line 16: same here for `cmd` `'` is needed?

ci-job-validation.groovy
nixpanic: usernames will not contain spaces or other special characters, a passwd might.

ci-job-validation.groovy
nixpanic: if the command contains special characters (`;` or `&&` and so on), only the 1st part of the command would be sent over to the remote machine

Madhu-1 (approved):

Issue #2118, opened 2021-06-02
Add clusterID and poolID configuration to support failover in Regional DR


Madhu-1:
```0001-0009-rook-ceph-0000000000000002-b0285c97-a0ce-11eb-8c66-0242ac110002```

The Above is the volumeID is sent back in the response to the CreateVolume operation and will be added as a volume handle in the PV spec. CO (Kubernetes) uses above as the identifier for further operations on the volume/PVC. 

The VolumeID is encoded as below

```
0001 -->	                         [csi_id_version=1:4byte] + [-:1byte]
0009 -->	                          [length of clusterID=1:4byte] + [-:1byte]
rook-ceph -->	                          [clusterID:36bytes (MAX)] + [-:1byte]
0000000000000002 -->	                  [poolID:16bytes] + [-:1byte]
b0285c97-a0ce-11eb-8c66-0242ac110002 --> [ObjectUUID:36bytes]

Total of constant field lengths, including '-' field separators would hence be,
4+1+4+1+1+16+1+36 = 64
```

When mirroring is enabled the volume which is `csi-vol-ObjectUUID` is mirrored to the other cluster.

> `csi-vol` is const name and over has the option to override it in storageclass.

 During the Disaster Recovery (failover operation) the PVC and PV will be recreated on the other cluster when cephcsi receives the request for operations like (NodeStage, ExpandVolume, DeleteVolume, etc) the volumeId is sent in the request which will help to identify the volume.

```yaml=
apiVersion: v1
kind: ConfigMap
data:
  config.json: |-
    [
      {
       "clusterID": "rook-ceph",
        "radosNamespace": "<rados-namespace>",
        "monitors": [
          "192.168.39.82:6789"
        ],
        "cephFS": {
          "subvolumeGroup": "<subvolumegroup for cephfs volumes>"
        }
      },
            {
       "clusterID": "fs-id",
        "radosNamespace": "<rados-namespace>",
        "monitors": [
          "192.168.39.83:6789"
        ],
        "cephFS": {
          "subvolumeGroup": "<subvolumegroup for cephfs volumes>"
        }
      }
    ]
metadata:
  name: ceph-csi-config
```

During CSI operations, CephCSI will decode the volumeID and gets the monitor configuration from the configmap, and from the pool ID it will get the pool Name and retrieves the OMAP data stored in the rados OMAP, and finally checks the volume is present in that pool.

We have two below problems here

* The clusterID can be different
[...]

PR #2292, opened 2021-07-15
doc: add design doc for clusterid poolid mapping

Madhu-1:
added design doc to handle volumeID mapping in case of the failover in the Disaster Recovery.

update #2118

### DIFF
--- a/docs/design/proposals/clusterid-mapping.md
+++ b/docs/design/proposals/clusterid-mapping.md
@@ -1,0 +1,176 @@
+# Design to handle clusterID and poolID for DR
+
+During disaster recovery/migration of a cluster, as part of the failover, the
+kubernetes artifacts like deployment, PVC, PV, etc will be restored to a new
+cluster by the admin. Even if the kubernetes objects are restored the
+corresponding RBD/CephFS subvolume cannot be retrieved during CSI operations as
+the clusterID and poolID are not the same in both clusters. Let's see the
+problem in more detail below.
+
+`0001-0009-rook-ceph-0000000000000002-b0285c97-a0ce-11eb-8c66-0242ac110002`
+
+The above is the sample volumeID sent back in response to the CreateVolume
+operation and added as a volumeHandle in the PV spec. CO (Kubernetes) uses
+above as the identifier for other operations on the volume/PVC.
+
+The VolumeID is encoded as,
+
+```text
+0001 -->                              [csi_id_version=1:4byte] + [-:1byte]
+0009 -->                              [length of clusterID=1:4byte] + [-:1byte]
+rook-ceph -->                         [clusterID:36bytes (MAX)] + [-:1byte]
+0000000000000002 -->                  [poolID:16bytes] + [-:1byte]
+b0285c97-a0ce-11eb-8c66-0242ac110002 --> [ObjectUUID:36bytes]
+Total of constant field lengths, including '-' field separators would hence be,
+4+1+4+1+1+16+1+36 = 64
+```
+
+When mirroring is enabled volume which is `csi-vol-ObjectUUID` is mirrored to
+the other cluster.
+
+> `csi-vol` is const name and over has the option to override it in
+> storageclass.
+
+During the Disaster Recovery (failover operation) the PVC and PV will be
+recreated on the other cluster. When Ceph-CSI receives the request for
+operations like (NodeStage, ExpandVolume, DeleteVolume, etc) the volumeID is
+sent in the request which will help to identify the volume.
+
+```yaml=
+apiVersion: v1
+kind: ConfigMap
+data:
+  config.json: |-
+    [
+      {
+       "clusterID": "rook-ceph",
+       "radosNamespace": "<rados-namespace>",
+       "monitors": [
+         "192.168.39.82:6789"
+       ],
+       "cephFS": {
+         "subvolumeGroup": "<subvolumegroup for cephfs volumes>"
+       }
+      },
+      {
+       "clusterID": "fs-id",
+       "radosNamespace": "<rados-namespace>",
+       "monitors": [
+         "192.168.39.83:6789"
+       ],
+       "cephFS": {
+         "subvolumeGroup": "<subvolumegroup for cephfs volumes>"
+       }
+      }
+    ]
+metadata:
+  name: ceph-csi-config
+```
+
+During CSI/Replication operations, Ceph-CSI will decode the volumeID and gets
+the monitor configuration from the configmap and by the poolID will get the
+pool Name and retrieves the OMAP data stored in the rados OMAP and finally
+check the volume is present in the pool.
+
+## Problems with volumeID Replication
+
+* The clusterID can be different
+  * as the clusterID is the namespace where rook is deployed, the Rook might be
+    deployed in the different namespace on a secondary cluster
+  * In standalone Ceph-CSI the clusterID is fsID and fsID is unique per
+    cluster
+
+* The poolID can be different
+  * PoolID which is encoded in the volumeID won't remain the same across
+    clusters
+
+To solve this problem we need to have a new mapping between clusterID's and the
+poolID's.
+
+Example configmap Need to be created before failover to `site2-storage` from
+`site1-storage` and `site3-storage`.
+
+```yaml=
+apiVersion: v1
+kind: ConfigMap
+data:
+  cluster-mapping.json: |-
+  [{
+    "clusterIDMapping": {
+    "site1-storage" (clusterID on site1): "site2-storage" (clusterID on site2)
+   },
+    "RBDPoolIDMapping": [{
+    "1" (poolID on site1): "2" (poolID on site2),
+    "11": "12"
+   }],
+    "CephFSFsIDMapping": [{
+    "13" (FsID on site1): "34" (FsID on site2),
+    "3": "4"
+   }]
+  }, {
+   "clusterIDMapping": {
+   "site3-storage"  (clusterID on site3): "site2-storage" (clusterID on site2)
+   },
+   "RBDPoolIDMapping": [{
+   "5" (poolID on site3): "2" (poolID on site2),
+   "16": "12"
+   }],
+   "CephFSFsIDMapping": [{
+   "3"(FsID on site3): "34" (FsID on site2),
+   "4": "4"
+   }]
+ }]
+metadata:
+  name: ceph-csi-config
+```
+
+**Note:-** the configmap will be mounted as a volume to the CSI (provisioner
+and node plugin) pods.
+
+The above configmap will get created as it is or updated (if new Pools are
+created on the existing cluster) with new entries when the admin choose to
+failover/failback the cluster.
+
+Whenever Ceph-CSI receives a CSI/Replication request it will first decode the
+volumeHandle and try to get the required OMAP details. If it is not able to
+retrieve the poolID or clusterID details from the decoded volumeHandle, Ceph-CSI
+will check for the clusterID and PoolID mapping.
+
+If the old volumeID
+`0001-00013-site1-storage-0000000000000001-b0285c97-a0ce-11eb-8c66-0242ac110002`
+contains the `site1-storage` as the clusterID, now Ceph-CSI will look for the
+corresponding clusterID `site2-storage` from the above configmap. If the
+clusterID mapping is found now Ceph-CSI will look for the poolID mapping ie
+mapping between `1` and `2`.
+
+Example:- pool with the same name exists on both the clusters with different IDs
+Replicapool with ID `1` on site1 and Replicapool with ID `2` on site2.
+
+After getting the required mapping Ceph-CSI has the required information to get
+more details from the rados OMAP. If we have multiple clusterID mapping it will
+loop through all the mapping and checks the corresponding pool to get the OMAP
+data. If the clusterID mapping does not exist Ceph-CSI will return a `Not
+Found` error message to the caller.
+
+After failover to the cluster `site2-storage`, the admin might have created new
+PVCs on the primary cluster `site2-storage`. Later after recovering the
+cluster `site1-storage`, the admin might choose to failback from
+`site2-storage` to `site1-storage`. Now admin needs to copy all the newly
+created kubernetes artifacts to the failback cluster. For clusterID mapping, the
+admin needs to copy the above-created configmap `ceph-clusterid-mapping` to
+the failback cluster. When Ceph-CSI receives a CSI/Replication request for
+the volumes created on the `site2-storage` it will decode the volumeID and
+retrieves the clusterID ie `site2-storage`. In the above configmap
+`ceph-clusterid-mapping` the `site2-storage` is the value and `site1-storage`
+is the key in the `clusterIDMapping` entry.
+
+Ceph-CSI will check both `key` and `value` to check the clusterID mapping. If it
+is found in `key` it will consider `value` as the  corresponding mapping, if it
+is found in `value` place it will treat `key` as the corresponding mapping and
+retrieves all the poolID details of the cluster.
+
+This mapping on the remote cluster is only required when we are doing a
+failover operation from the primary cluster to a remote cluster. The existing
+volumes that are created on the remote cluster does not require
+any mapping as the volumeHandle already contains the required information about
+the local cluster (clusterID, poolID etc).


docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,140 @@
+# Design to handle clusterID and poolID for DR
+
+`0001-0009-rook-ceph-0000000000000002-b0285c97-a0ce-11eb-8c66-0242ac110002`
+
+The Above is the sample volumeID sent back in response to the CreateVolume operation
```
humblec: s/Above/above

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,140 @@
+# Design to handle clusterID and poolID for DR
+
```
nixpanic, line 2: please include a short paragraph with the problem statement as an introduction

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,140 @@
...
+metadata:
+  name: ceph-csi-config
+```
+
+During CSI/Replication operations, Ceph-CSI will decode the volumeID and gets
+the monitor configuration from the configmap and the poolID it will get the
+pool Name and retrieves the OMAP data stored in the rados OMAP and finally
+check the volume is present in the pool.
+
+We have two problems!
```
nixpanic: make this a sub-title, and phrase it more like "Two Problems with volumeID Replication" or so

nixpanic (commented):
It is unclear to me if the ConfigMap is expected to be the same or different on the two clusters. Ideally the contents is the same, and a user can just copy/paste it. This would be needed if it is not a one-way-only 'migration', but a fallback with new created volumes is expected to function. More details about this would be welcome.

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,140 @@
...
+the poolID's.
+
+```yaml=
+apiVersion: v1
+kind: ConfigMap
+data:
+  config.json: |-
+    [
+      {
+       "rook-ceph"(cluster1): "kube-system"(cluster2),
```
nixpanic: An existing namespace like `kube-system` is probably not nice to re-use. Maybe use `site1-storage` and `site2-storage`, or give them names like `storage-amsterdam` and `storage-bangalore`

Madhu-1:
> It is unclear to me if the ConfigMap is expected to be the same or different on the two clusters. Ideally the contents is the same, and a user can just copy/paste it. This would be needed if it is not a one-way-only 'migration', but a fallback with new created volumes is expected to function. More details about this would be welcome.

In most of the cases the content of the configmap is not the same as the pooldid's and clusterid are different. and added some details about it.

nixpanic:
> In most of the cases the content of the configmap is not the same as the pooldid's and clusterid are different. and added some details about it.

That is unfortunate. Could it be made the same?

I am also missing an example or explanation of how >2 clusters can be configured. Even if that is not something that is expected to be functional soon, it should be considered as a long-term goal and the design should be prepared (easily extended) for it.

Madhu-1:
> > In most of the cases the content of the configmap is not the same as the pooldid's and clusterid are different. and added some details about it.
> 
> That is unfortunate. Could it be made the same?
> 

Not sure why it's required, This is by design and expected to be different and this configuration will be updated when the admin creates/deletes the pools or when he adds new ceph cluster details to the cephcsi configmap where cephcsi has a mapping between the clusterID and monitor details.

> I am also missing an example or explanation of how >2 clusters can be configured. Even if that is not something that is expected to be functional soon, it should be considered as a long-term goal and the design should be prepared (easily extended) for it.

The example in the design doc contains the mapping when cephcsi is configured to work with multiple ceph clusters.

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,150 @@
...
+During disaster recovery/migration of a cluster, the application and the
+storage will be moved to a new cluster. When the kubernetes storage objects
+like  PV,PVC etc will be restored by the admin as part of the failover. Even if
+the kubernetes objects are restored the corresponding RBD/CephFS subvolume
+cannot be retrieved during CSI operations as the clusterID and poolID are not
+same in both clusters. Lets see the problem in more detail below.
+
+`0001-0009-rook-ceph-0000000000000002-b0285c97-a0ce-11eb-8c66-0242ac110002`
+
+The Above is the sample volumeID sent back in response to the CreateVolume
```
humblec: s/Above/above

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,150 @@
+# Design to handle clusterID and poolID for DR
+
+During disaster recovery/migration of a cluster, the application and the
+storage will be moved to a new cluster. When the kubernetes storage objects
```
humblec: @Madhu-1 `When the ....` , this sentence looks incomplete

nixpanic:
> In most of the cases the content of the configmap is not the same as the pooldid's and clusterid are different. and added some details about it.

I understand that the Cluster-ID and Pool-IDs will most likely be different. That is not what I meant to comment on.

My concern is that the ConfigMap for the mapping of the IDs is different for the different sites. This makes it error-prone and difficult to use. A format that makes it possible for the ConfigMap to be the same on both sites would be much preferred.

In the latest update, `site3-storage` shows up in the ConfigMap as well. I appreciate the addition! For me, the format becomes even more difficult to read, and I wonder where the mapping between `site1-storage` and `site3-storage` is. This can be calculated of course. But if there are even more sites, it becomes difficult to see if all combinations are covered.

Would it not make sense to use an ordered list for the mappings? More like a table so that it becomes
```yaml=
apiVersion: v1
kind: ConfigMap
metadata:
  name: ceph-clusterid-mapping
data:
  config.json: |-
  [{
    "clusterIDMapping": ["site1-storage", "site2-storage", "site3-storage"]
   },
    "RBDPoolIDMapping": [
      ["1", "2", "5"],
      ["11", "12", "16"]
   ],
...
```
This looks a little simpler to maintain to me. Validation can also just check the number of entries in `clusterIDMapping` and the number of entries in the other mappings.

Madhu-1:
> > In most of the cases the content of the configmap is not the same as the pooldid's and clusterid are different. and added some details about it.
> 
> I understand that the Cluster-ID and Pool-IDs will most likely be different. That is not what I meant to comment on.
> 
> My concern is that the ConfigMap for the mapping of the IDs is different for the different sites. This makes it error-prone and difficult to use. A format that makes it possible for the ConfigMap to be the same on both sites would be much preferred.
> 
> In the latest update, `site3-storage` shows up in the ConfigMap as well. I appreciate the addition! For me, the format becomes even more difficult to read, and I wonder where the mapping between `site1-storage` and `site3-storage` is. This can be calculated of course. But if there are even more sites, it becomes difficult to see if all combinations are covered.
> 
> Would it not make sense to use an ordered list for the mappings? More like a table so that it becomes
> 
> ```
> apiVersion: v1
> kind: ConfigMap
> metadata:
>   name: ceph-clusterid-mapping
> data:
>   config.json: |-
>   [{
>     "clusterIDMapping": ["site1-storage", "site2-storage", "site3-storage"]
>    },
>     "RBDPoolIDMapping": [
>       ["1", "2", "5"],
>       ["11", "12", "16"]
>    ],
> ...
> ```
> 
> This looks a little simpler to maintain to me. Validation can also just check the number of entries in `clusterIDMapping` and the number of entries in the other mappings.

IMO it will be more problematic to edit if we have more clusters and more pools. with the current proposed change, the configmap entry will grow it will always have 1:1 mapping and be easy to update without touching the entry of the other clusters.

@humblec @ShyamsundarR any suggestions.

If all agree that the suggestion is fine am happy to incorporate the changes.

nixpanic:
> IMO it will be more problematic to edit if we have more clusters and more pools. with the current proposed change, the configmap entry will grow it will always have 1:1 mapping and be easy to update without touching the entry of the other clusters.

The problem I see is that there is a 1:N mapping, and that should somehow be accommodated in a form that does not require multiple 1:1 mappings.

humblec:
I agree to the point that, the clusterID and PoolID consistency between the sites is difficult to maintain, especially when the admins differ and when the action is manual. Maintaining the current schema with different clusterIDs and Its own pool list in separate block looks to be difficult to maintain as well when it comes to 1:N scenario as noted in above comments. 

I have an additional query here:
```
After getting the required mapping Ceph-CSI has the
required information to get more details from the rados OMAP.
```
@Madhu-1 what is referred here as `required information` above ? and how it is  used for any of subsequent  operation ?

Madhu-1:
> I agree to the point that, the clusterID and PoolID consistency between the sites is difficult to maintain, especially when the admins differ and when the action is manual. Maintaining the current schema with different clusterIDs and Its own pool list in separate block looks to be difficult to maintain as well when it comes to 1:N scenario as noted in above comments.
> 

@humblec are you suggesting using the json format specified from Niels above?

> I have an additional query here:
> 
> ```
> After getting the required mapping Ceph-CSI has the
> required information to get more details from the rados OMAP.
> ```
> 
> @Madhu-1 what is referred here as `required information` above ? and how it is used for any of subsequent operation ?

required information is rados namespace and the pool ID/Name in which cephcsi need to look for the rados objects it stores  in the pool (like volumeID, owner, kmsID,volume name prefix etc)

humblec:
> >   
> > @Madhu-1 what is referred here as `required information` above ? and how it is used for any of subsequent operation ?
> 
> required information is whatever CSI stores in the rados objects in the pool (like volumeID, owner, kmsID, etc)

Only `volumeID` in use in context of replication, Isnt it ?

Madhu-1:
> > > @Madhu-1 what is referred here as `required information` above ? and how it is used for any of subsequent operation ?
> > 
> > 
> > required information is whatever CSI stores in the rados objects in the pool (like volumeID, owner, kmsID, etc)
> 
> Only `volumeID` in use in context of replication, Isnt it ?

Yes, but other objects are required for CSI calls like Mount/Expand etc.

humblec:
Few thoughts I am brainstorming here is :

1) Having same configmap across clusters schema helps , it also help on automation:  In short, we can do the parsing of the clusterID and pick subjected Pool ID from next entry mapping
2) A marker like `local` or `remote` in existing config map OR having a separate config file for additional operations 
3) A mechanism of extending current volumehandle with extra information like `imageid `, `kms` or a encoded string of a `struct` we want. So that by maintainting backeward compatibility we can fallback..etc

Madhu-1:
@nixpanic @humblec updated the doc as per discussion PTAL.

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,170 @@
...
+poolID's.
+
+Sample configmap Need to be created before failover to `site2-storage` from
+`site1-storage` and `site3-storage`.
+
+```yaml=
+apiVersion: v1
+kind: ConfigMap
+data:
+  config.json: |-
```
nixpanic: can you rename this to `cluster-mapping.json` or something that prevents confusion with the existing `config.json`

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,170 @@
...
+  }, {
+   "clusterIDMapping": {
+   "site3-storage"  (clusterID on site3): "site2-storage" (clusterID on site2)
+   },
+   "RBDPoolIDMapping": [{
+    "5" (poolID on site3): "2" (poolID on site2),
+    "16": "12"
+   }],
+   "CephFSPoolIDMapping": [{
+   "3"(poolID on site1): "34" (poolID on site2),
```
nixpanic: 1st entry is for `site-3`?

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,170 @@
...
+
+* The poolID can be different
+  * PoolID which is encoded in the volumeID won't remain the same across
+    clusters
+
+To solve this problem we need to have a new mapping between clusterID's and the
+poolID's.
+
+Sample configmap Need to be created before failover to `site2-storage` from
+`site1-storage` and `site3-storage`.
```
nixpanic, line 91: was this configuration not sufficient for all sites?

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,170 @@
...
+
+Whenever Ceph-CSI receives a CSI/Replication request it will first decode the
+volumeHandle and try to get the required OMAP details if it is not able to
+retrieve receives a `Not Found` error message and Ceph-CSI will check for the
+clusterID mapping. If the old volumeID
+`0001-00013-site1-storage-0000000000000001-b0285c97-a0ce-11eb-8c66-0242ac110002`
+contains the `site1-storage` as the clusterID, now Ceph-CSI will look for the
+corresponding clusterID `site2-storage` from the above configmap. If the
+clusterID mapping is found now Ceph-CSI will look for the poolID mapping ie
+mapping between `1` and `2`. Example:- pool with name exists on both the
```
nixpanic: put example on its own line, and indent the contents of the example?

nixpanic (commented):
A few nits, and missed a copy/paste update in the example. Looks good otherwise.

docs/design/proposals/clusterid-mapping.md
Madhu-1: Yes it was a typo :)

docs/design/proposals/clusterid-mapping.md
Madhu-1: Yes, this is sufficient. Add comment to consider this as example configmap and the same configmap can be created on the clusters before failing over.

nixpanic (dismissed):

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+clusterID mapping.
+
+If the old volumeID
+`0001-00013-site1-storage-0000000000000001-b0285c97-a0ce-11eb-8c66-0242ac110002`
+contains the `site1-storage` as the clusterID, now Ceph-CSI will look for the
+corresponding clusterID `site2-storage` from the above configmap. If the
+clusterID mapping is found now Ceph-CSI will look for the poolID mapping ie
+mapping between `1` and `2`.
+
+Example:- pool with name exists on both the clusters with different IDs
```
Yuggupta27:
```suggestion
Example:- pool with the same name exists on both the clusters with different IDs
```

Yuggupta27 (changes requested):

nixpanic (dismissed):

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+
+```yaml=
+apiVersion: v1
+kind: ConfigMap
+data:
+  config.json: |-
+    [
+      {
+       "clusterID": "rook-ceph",
+        "radosNamespace": "<rados-namespace>",
```
humblec: looks like an extra space intendation in this line and later lines.

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+        "monitors": [
+          "192.168.39.82:6789"
+        ],
+        "cephFS": {
+          "subvolumeGroup": "<subvolumegroup for cephfs volumes>"
+        }
+      },
+            {
+       "clusterID": "fs-id",
+        "radosNamespace": "<rados-namespace>",
```
humblec: same here

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+metadata:
+  name: ceph-csi-config
+```
+
+During CSI/Replication operations, Ceph-CSI will decode the volumeID and gets
+the monitor configuration from the configmap and the poolID will get the
+pool Name and retrieves the OMAP data stored in the rados OMAP and finally
+check the volume is present in the pool.
+
+## Two Problems with volumeID Replication
```
humblec: better to have a title `Problems` .. instead of `Two Problems`

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+
+* The clusterID can be different
+  * as the clusterID is the namespace where rook is deployed, the Rook might be
+    deployed in the different namespace on a secondary cluster
+  * In standalone Ceph-CSI the clusterID is fsID. and fsID is unique per
```
humblec: remove `.`

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+**Note:-** the configmap will be mounted as a volume to the CSI (provisioner
+and node plugin) pods.
+
+The above configmap will get created as it is or updated (if new Pools are
+created on the existing cluster) with new entries when the admin choose to
+failover/failback the cluster.
+
+Whenever Ceph-CSI receives a CSI/Replication request it will first decode the
+volumeHandle and try to get the required OMAP details if it is not able to
+retrieve receives a `Not Found` error message and Ceph-CSI will check for the
```
humblec: Above sentence has to be reworded, retrieve receives... looks incomplete

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+       }
+      }
+    ]
+metadata:
+  name: ceph-csi-config
+```
+
+During CSI/Replication operations, Ceph-CSI will decode the volumeID and gets
+the monitor configuration from the configmap and the poolID will get the
+pool Name and retrieves the OMAP data stored in the rados OMAP and finally
```
humblec, line 72: may be something like , `by the poolID will get the ...` ?

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+
+**Note:-** the configmap will be mounted as a volume to the CSI (provisioner
+and node plugin) pods.
+
+The above configmap will get created as it is or updated (if new Pools are
+created on the existing cluster) with new entries when the admin choose to
+failover/failback the cluster.
+
+Whenever Ceph-CSI receives a CSI/Replication request it will first decode the
+volumeHandle and try to get the required OMAP details if it is not able to
```
humblec: I think above sentence need to be rephrased.

docs/design/proposals/clusterid-mapping.md
```diff
@@ -0,0 +1,176 @@
...
+Replicapool with ID `1` on site1 and Replicapool with ID `2` on site2.
+
+After getting the required mapping Ceph-CSI has the required information to get
+more details from the rados OMAP. If we have multiple clusterID mapping it will
+loop through all the mapping and checks the corresponding pool to get the OMAP
+data. If the clusterID mapping does not exist Ceph-CSI will return a `Not
+Found` error message to the caller.
+
+After failover to the cluster `site2-storage`, the admin might have created new
+PVC's on the primary cluster `site2-storage`. Later after recovering the
```
humblec: may be PVC's/PVCs

docs/design/proposals/clusterid-mapping.md
Madhu-1: @humblec please suggest the changes you are thinking of

Madhu-1:
@humblec PTAL. Thanks

docs/design/proposals/clusterid-mapping.md
humblec: @Madhu-1 after `OMAP details ` you may close that sentence or concatenate it properly

Madhu-1: @humblec done PTAL

humblec (approved):

Madhu-1:
@nixpanic PTAL

nixpanic (approved):

PR #2294, opened 2021-07-16
ci: add the root user to the libvirt group

nixpanic:
When starting a minikube VM, there are always warning messages like
this:

    X libvirt group membership check failed:
    user is not a member of the appropriate libvirt group

The CI jobs run as root, so minikube works just fine. By adding the root
user to the libvirt group, the warning is hopefully removed.

Note that the libvirt group may not exist yet, as the packages will get
installed in a later stage. This change also adds a check to create the
libvirt group in case it is missing.

---

<details>

  progress or results

</details>

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -80,6 +80,10 @@ set -x
 
 dnf -y install git podman make
 
+# minikube wants the user (everything runs as root) to be in the libvirt group
+getent group libvirt || groupadd --system libvirt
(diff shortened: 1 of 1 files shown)


Madhu-1 (approved):

Yuggupta27 (approved):
LGTM! Was facing this locally some time back too.

Issue #2204, opened 2021-06-24
rbd-nbd: occasionally file system entering into read-only on nodeplugin restart


Madhu-1:
Requires below PR from ceph:

- [ ] https://github.com/ceph/ceph/pull/42609
```

[1mSTEP[0m: restart rbd-nbd process on nodeplugin and continue IO after nodeplugin restart
Jun 23 15:02:27.261: INFO: Running '/usr/bin/kubectl --server=https://192.168.39.63:8443 --kubeconfig=/root/.kube/config --namespace=cephcsi-e2e-aec55366 --namespace=cephcsi-e2e-aec55366 delete -f -'
Jun 23 15:02:27.387: INFO: stderr: "warning: deleting cluster-scoped resources, not scoped to the provided namespace\n"
Jun 23 15:02:27.387: INFO: stdout: "storageclass.storage.k8s.io \"csi-rbd-sc\" deleted\n"
Jun 23 15:02:27.394: INFO: ExecWithOptions {Command:[/bin/sh -c ceph fsid] Namespace:rook-ceph PodName:rook-ceph-tools-7467d8bf8-vjzwh ContainerName:rook-ceph-tools Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Jun 23 15:02:27.394: INFO: >>> kubeConfig: /root/.kube/config
Jun 23 15:02:29.405: INFO: Waiting up to &PersistentVolumeClaim{ObjectMeta:{rbd-pvc  rbd-3274    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Spec:PersistentVolumeClaimSpec{AccessModes:[ReadWriteOnce],Resources:ResourceRequirements{Limits:ResourceList{},Requests:ResourceList{storage: {{1073741824 0} {<nil>} 1Gi BinarySI},},},VolumeName:,Selector:nil,StorageClassName:*csi-rbd-sc,VolumeMode:nil,DataSource:nil,},Status:PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},},} to be in Bound state
Jun 23 15:02:29.406: INFO: waiting for PVC rbd-pvc (0 seconds elapsed)
Jun 23 15:02:31.409: INFO: waiting for PVC rbd-pvc (2 seconds elapsed)
[...]

Issue #2264, opened 2021-07-09
CI jobs often fail with minikube error `X Exiting due to GUEST_PROVISION`


nixpanic:
# Describe the bug #

When runninc CI jobs, minikube regularly fails with the following error:
```
* Failed to start kvm2 VM. Running "minikube delete" may fix it: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute

X Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute
```

This prevents the job from continuing, and a `/retest ...` is needed.

# Logs #

From [mini-e2e_k8s-1.20/1523](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.20/detail/mini-e2e_k8s-1.20/1523/pipeline):
```
* Creating kvm2 VM (CPUs=8, Memory=14336MB, Disk=20000MB) ...
* Deleting "minikube" in kvm2 ...
! StartHost failed, but will try again: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute
* Creating kvm2 VM (CPUs=8, Memory=14336MB, Disk=20000MB) ...
* Failed to start kvm2 VM. Running "minikube delete" may fix it: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute

X Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute
```

nixpanic:
https://github.com/ceph/ceph-csi/pull/2343#issuecomment-891825471 failed again ([logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.19/detail/mini-e2e-helm_k8s-1.19/2298/pipeline))

nixpanic:
[...]

PR #2298, opened 2021-07-16
e2e: sync data from rbd-nbd mount

pkalever:

Until we have a real fix, just to avoid occasionally file system entering into read-only on nodeplugin restart, lets sync data from the application pod.

Updates: #2204

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -465,6 +465,21 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to create PVC and application with error %v", err)
 				}
 
+				appOpt := metav1.ListOptions{
+					LabelSelector: fmt.Sprintf("app=%s", app.Name),
+				}
+				// TODO: Remove this once we ensure that rbd-nbd can sync data
+				// from Filesystem layer to backend rbd image as part of its
+				// detach or SIGTERM signal handler
+				_, stdErr, err := execCommandInPod(
+					f,
+					fmt.Sprintf("sync %s", app.Spec.Containers[0].VolumeMounts[0].MountPath),
+					app.Namespace,
+					&appOpt)
+				if err != nil || stdErr != "" {
+					e2elog.Failf("failed to sync, err: %v, stdErr: %v ", err, stdErr)
+				}
+
 				// validate created backend rbd images
 				validateRBDImageCount(f, 1, defaultRBDPool)
 
@@ -537,9 +552,6 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to poll: %v", err)
 				}
 
-				appOpt := metav1.ListOptions{
-					LabelSelector: fmt.Sprintf("app=%s", app.Name),
-				}
 				filePath := app.Spec.Containers[0].VolumeMounts[0].MountPath + "/test"
 				_, stdErr, err = execCommandInPod(
 					f,


e2e/rbd.go
```diff
@@ -539,14 +539,26 @@ var _ = Describe("RBD", func() {
...
 				filePath := app.Spec.Containers[0].VolumeMounts[0].MountPath + "/test"
-				_, stdErr, err = execCommandInPod(
-					f,
-					fmt.Sprintf("echo 'Hello World' > %s", filePath),
-					app.Namespace,
-					&appOpt)
-				if err != nil || stdErr != "" {
-					e2elog.Failf("failed to write IO, err: %v, stdErr: %v ", err, stdErr)
-				}
+				err = wait.PollImmediate(poll, timeout, func() (bool, error) {
```
nixpanic: might be useful to place this in a helper function `writeInPodWithRetry()`?

e2e/rbd.go
```diff
@@ -539,14 +539,26 @@ var _ = Describe("RBD", func() {
...
+					_, stdErr, err = execCommandInPod(
+						f,
+						fmt.Sprintf("echo 'Hello World' > %s", filePath),
+						app.Namespace,
+						&appOpt)
+					if err != nil {
+						e2elog.Failf("failed to run write command: %v", err)
```
nixpanic: this will cause the test to fail, can it happen that `err` is non-nil, and the filesystem or block-device is read-only?

e2e/rbd.go
```diff
@@ -539,14 +539,26 @@ var _ = Describe("RBD", func() {
...
+					}
+					if stdErr != "" {
+						readOnlyErr := fmt.Sprintf("cannot create %s: Read-only file system", filePath)
```
nixpanic: If a filesystem became read-only, you probably need to remount it read-writable again? Or is that automatically done?

nixpanic (changes requested):

e2e/rbd.go
pkalever: Sure, that should make this less complex and modular.

e2e/rbd.go
pkalever: I think I need to reorder this check:

I need to check for stdErr first and then if that is equal to "Read-only ..." then retry, if stdErr != "Read-only" then err should bailout.

Thanks, I think I recorded at the last moment.

e2e/rbd.go
pkalever: Honestly speaking, I tried to reproduce this several times locally, but couldn't hit it.
I expect it to resolve automatically, but I'm not entirely sure!

pkalever (commented):
Thanks @nixpanic

e2e/rbd.go
pkalever: oh! Given this is handling special cases like retry for Read-Only error but not for others, this might not be helpful?

e2e/rbd.go
pkalever: @trociny any thoughts here, have you hit [this](https://github.com/ceph/ceph-csi/issues/2204) before?

trociny: As I understand, the "read-only file system" error is returned not by rbd/nbd layer but the upper filesystem error: when it previously encounters IO error it makes the filesystem read-only. And it does not resolve automatically and needs remount. I don't think you can't safely remount while still running processes that use the filesystem. So you would need to restart those too. I.e. effectively you would need to restart the pod anyway.

I would rather investigate the root cause why it entered that R/O state after reattach to try to avoid this. Looking at the dmesg for kernel error messages may be useful here.

One hypothesis is that you are still terminating (restarting) rbd-nbd process using `kill -9` command (which I seemed to see you were using previously). In this case any pending i/o does not have a chance to complete and probably the I/O error is returned for it (though I would expect it might hung forever).

Another hypothesis that there is some I/O timeout inside the kernel that was hit when the device was in detached state and the error was returned, though I am not aware of such timeout.

Anyway, the kernel logs seem the most useful here to trace this.

pkalever: > As I understand, the "read-only file system" error is returned not by rbd/nbd layer but the upper filesystem error: when it previously encounters IO error it makes the filesystem read-only. And it does not resolve automatically and needs remount. I don't think you can't safely remount while still running processes that use the filesystem. So you would need to restart those too. I.e. effectively you would need to restart the pod anyway.

OK.

> 
> I would rather investigate the root cause why it entered that R/O state after reattach to try to avoid this. Looking at the dmesg for kernel error messages may be useful here.
> 

Yes, this is the reason why I tried to reproduce this, but with no luck. I will still give one more attempt later.


> One hypothesis is that you are still terminating (restarting) rbd-nbd process using `kill -9` command (which I seemed to see you were using previously). In this case any pending i/o does not have a chance to complete and probably the I/O error is returned for it (though I would expect it might hung forever).
> 

Glad you brought this valid point here. I thought of bringing it into the team discussions when we initially discussed about it, but completely forget about this later.

By default, Kubernetes will send the SIGTERM signal to the processes and wait for `terminationGracePeriodSeconds` (default 30 seconds) before force`SIGKILL` killing the process.

So there is a chance that the `rbd-nbd` process might get `SIGKILL`, if the Grace Period is expired.

@nixpanic @Madhu-1
Maybe for nbd, 
1. we can adjust `terminationGracePeriodSeconds` to 120 seconds in the [nodeplugin](https://github.com/ceph/ceph-csi/blob/devel/deploy/rbd/kubernetes/csi-rbdplugin.yaml)  to satisfy the test case and see if we can reproduce this again.
2. Add a preStop hook inside the [nodeplugin](https://github.com/ceph/ceph-csi/blob/devel/deploy/rbd/kubernetes/csi-rbdplugin.yaml) and wait for rbd-nbd processes to terminate while keeping the `terminationGracePeriodSeconds` high.

I like the latter one which is safer for applications data.

> Another hypothesis that there is some I/O timeout inside the kernel that was hit when the device was in detached state and the error was returned, though I am not aware of such timeout.
> 

Sure.

> Anyway, the kernel logs seem the most useful here to trace this.

I will also try to reproduce it locally and get back on this.

Thanks for the help @trociny.

pkalever:
@nixpanic could you please mark this PR as `DNM`, we would want to finish the discussion first and then as part of this PR, we can make changes accordingly. The current changes don't look right at the moment.

e2e/rbd.go
pkalever: @trociny as you know, we cannot expect the container to wait forever to finish the pending IO so that the rbd-nbd process is done with wait_io_finish().

What is the suggested timeout wait for the rbd-nbd processes to finish the pending IO?

trociny: I don't know. Theoretically, there might be some issue with the ceph cluster, and in this case the io will get stuck until the issue is resolved (possibly forever).

Your 30 sec timeout might be good enough (although I might increase it to 60 sec). If it is really 30 sec before `SIGKILL` is sent, then it does not look like  it was the case in the issue #2204, because if I interpret the log correctly the restart was not more than 18 sec:
```
Jun 23 15:02:45.485: INFO: map command running before restart, mapCmd: /usr/bin/rbd-nbd --id cephcsi-rbd-node -m rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789 --keyfile=/tmp/csi/keys/keyfile-113061691 map replicapool/csi-vol-a5bccd9c-d42b-11eb-991e-4664e1c21ce8 --try-netlink --reattach-timeout=180
...
Jun 23 15:03:02.638: INFO: attach command running after restart, runningAttachCmd: /usr/bin/rbd-nbd --id cephcsi-rbd-node -m rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789 --keyfile=/tmp/csi/keys/keyfile-test attach --device /dev/nbd0 replicapool/csi-vol-a5bccd9c-d42b-11eb-991e-4664e1c21ce8 --try-netlink --reattach-timeout=180
```

trociny: Basically, sending `SIGKILL` eventually means that you should rather restart the pod instead of trying reattach the device.

pkalever: > I don't know. Theoretically, there might be some issue with the ceph cluster, and in this case the io will get stuck until the issue is resolved (possibly forever).

Even in that case, ceph might define an IO timeout (grace period) and fail the IO (I guess) if the IO doesn't get any response. I'm saying this from my experience with gluster storage (ignore, I'm not a ceph expert anyways!)

> 
> Your 30 sec timeout might be good enough (although I might increase it to 60 sec). If it is really 30 sec before `SIGKILL` is sent, then it does not look like it was the case in the issue #2204, because if I interpret the log correctly the restart was not more than 18 sec:
> 
> ```
> Jun 23 15:02:45.485: INFO: map command running before restart, mapCmd: /usr/bin/rbd-nbd --id cephcsi-rbd-node -m rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789 --keyfile=/tmp/csi/keys/keyfile-113061691 map replicapool/csi-vol-a5bccd9c-d42b-11eb-991e-4664e1c21ce8 --try-netlink --reattach-timeout=180
> ...
> Jun 23 15:03:02.638: INFO: attach command running after restart, runningAttachCmd: /usr/bin/rbd-nbd --id cephcsi-rbd-node -m rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789 --keyfile=/tmp/csi/keys/keyfile-test attach --device /dev/nbd0 replicapool/csi-vol-a5bccd9c-d42b-11eb-991e-4664e1c21ce8 --try-netlink --reattach-timeout=180
> ```

@trociny the termination grace period countdown actually starts much ahead, even before invoking the preStop hook and not once the SIGTERM signal will be sent. So if the `terminationGracePeriodSeconds` is 30 seconds, that actually doesn't mean the process gets 30 seconds after SIGTERM is sent.

That said,
Can we consider NBD_ATTR_TIMEOUT (`--io-timeout`) as the max wait time for the rbd-nbd process here (once it received the SIGTERM)? if so, what would it be?


> Basically, sending SIGKILL eventually means that you should rather restart the pod instead of trying reattach the device.

Yes this should be expected.

We cannot expect rbd-nbd to behave sanely when we shut it down ungracefully. Hence give it some time to gracefully handle the IO (wait for pending IO to finish) and proceed further.


cc: @idryomov @lxbsz


Thanks Again.

trociny: > Can we consider NBD_ATTR_TIMEOUT (--io-timeout) as the max wait time for the rbd-nbd process here (once it received the SIGTERM)? if so, what would it be?

Yes, it looks like a good idea. If `--io-timeout` is exceeded then there is no much sense to wait for the process to gracefully shutdown. By default rbd-nbd does not set it, and I am not sure what the kernel uses in this case. If I remember correctly it was 30 sec when I looked at it.

Anyway, if you hit this timeout then there is no much sense in reattaching -- you will need to restart the pod.

pkalever: Thanks @trociny

@idryomov @trociny @lxbsz
what is the grace period timeout for IO in ceph (for rbd volumes) after which any pending IO will return failure (if not handled)?

May be this could be a factor to decide on `--io-timeout` for `rbd-nbd` and based on these numbers we can adjust `terminationGracePeriodSeconds` for pod termination.

trociny: On the rados level the IO will hung forever (until it is served).

lxbsz: > Thanks @trociny
> 
> @idryomov @trociny @lxbsz
> what is the grace period timeout for IO in ceph (for rbd volumes) after which any pending IO will return failure (if not handled)?

It should be `rados_osd_op_timeout`, the default value is 0, means unlimited. For this maybe you can check the use the same logic in tcmu-runner's rbd.c handler, more detail please see `timer_check_and_set_def()`.

> 
> May be this could be a factor to decide on `--io-timeout` for `rbd-nbd` and based on these numbers we can adjust `terminationGracePeriodSeconds` for pod termination.

pkalever: Thanks for your valuable suggestions @trociny and @lxbsz, I will send a patch to improve rbd-nbd.

In the meanwhile to work around this issue at the ceph-csi level, I have sent https://github.com/ceph/ceph-csi/pull/2313
This should help reduce the frequency of hitting such Read-Only mounts for now.

trociny: Just a note. Although improving rbd-nbd to optionally change `rados_osd_op_timeout` looks like a good idea, I don't see how it will resolve (or improve considerably) this particular case. Instead of hanging forever the io will fail, so your nbd device is in an error state. The timeout error will be returned to the filesystem and it will enter the RO state.

pkalever: Exactly!

This will help us improve the coordination b/w the backend (max expected wait time for IO to finish) and pod ungrateful termination wait time. Obviously, if you look at this case, it just waits for less than 30 seconds and assumes IO will not be finished and give up. Having a tunable for `rados_osd_op_timeout` (300-600 sec?) at rbd-nbd will help csi understand time factor better (because CSI will set that tunable at the initial place by invoking `rbd-nbd` with something like `--radosop-timeout=300`)

Why do we want to wait forever? There should be a timeout for IO to finish or acknowledge that it's failing. If you look at this case, we cannot wait for rbd-nbd forever because that will lead to hanging for pod termination forever.

I also wonder, if the IO wait time is set to infinite, how does one identify the missed writes/reads? Because we won't be able to say a given IO failed by now or is expected to finish soon, won't it be better to drop them after a max timeout?

Thanks again!

trociny: Just dropping a timed out IO would mean that the application that issued it would never get the response, i.e. it would still waiting forever but now without any chance to recover. Returning the timeout error to the upper layer most likely will make the application to fail. When waiting forever for a rados op to complete there is a chance that it is a temporarily problem with Ceph (which is usually the case in prod clusters but not in the test) that will eventually resolve and the IO successfully complete.

I am not trying to say that waiting forever is better than aborting with timeout. It is just not much difference. I.e. now you have to `kill -9` to "simulate" the timeout. When you add the timeout you will not need to `kill -9` but you will still need to recover from this messy state, and just reattaching will not work. You will need to restart the pod, and it will work only if the cluster is healthy again.

idryomov: I don't think messing with `rados_osd_op_timeout` is a good idea.  If a per-I/O timeout is desired, I think it should be configured on the client side, that is on the individual `/dev/nbdX` device in the block layer.

pkalever: Thanks @trociny I totally get you.

@idryomov Hmm, ok then, how long the pod should wait here for rbd-nbd to finish the pending IO? If you look at https://github.com/ceph/ceph-csi/pull/2313, I thought of giving rbd-nbd 600 sec, do you think this should be good ? Or do we have any better way to fix this then?

idryomov: Well, any timeout can be just a millisecond too short ;)  That said, 600 seconds is probably too long.

As Mykola stressed, whether we configure a timeout inside/from `rbd-nbd` or whether that timeout comes from the outside in the form of a `SIGKILL` after `terminationGracePeriodSeconds`, the result is going to be more or less the same -- lost I/O and a wedged mount.  In the absence of being able to refuse nodeplugin pod restart (e.g. somehow tell k8s to back off and retry later), CSI needs to be prepared for that as it is going to happen one way or another.

pkalever: @idryomov Yes, thanks for summarizinging it all :-)

Well, I hope its clear `terminationGracePeriodSeconds` means the max wait time that a pod should wait before encountering SIGKILL. Just in case if all the `rbd-nbd` processes are done within say 10 seconds, then pod deleting will happen immediately i.e. it won't wait or `terminationGracePeriodSeconds` to happen.

That said, if `600` seconds sounds long, please suggest a good timeout here.

idryomov: Maybe 120 seconds like was mentioned earlier?  It's just that 10 minutes is practically an eternity by k8s standards -- I think the default eviction timeout is like 5 minutes?

Honestly, as noted in https://github.com/ceph/ceph-csi/pull/2313 I'm not sure touching `terminationGracePeriodSeconds` is the right thing to do at all.

pkalever: > Maybe 120 seconds like was mentioned earlier? It's just that 10 minutes is practically an eternity by k8s standards -- I think the default eviction timeout is like 5 minutes?

Thanks for the catch. Yes, from the docs it looks like the default but can be configured.
Yes, we should be good to adopt 120 - 180 seconds.

> Honestly, as noted in #2313 I'm not sure touching `terminationGracePeriodSeconds` is the right thing to do at all.

I understand, as discussed here and there, this should be a workaround fix to give some time for rbd-nbd to handle pending IO and also note down the behavior with the new timeout from the ceph-csi CI.

We should definitely give more thoughts around this, and here is the tracker https://tracker.ceph.com/issues/51820 

Thanks @idryomov

trociny:
@pkalever Forgot to mention it earlier, but actually, the rbd-nbd does not require any modifications if you want to change  `rados_osd_op_timeout`. As it is a ceph-specific config option there are several ways you can pass it to the rbd-nbd's rados client:

1) set it in ceph.conf
2) provide as `--rados-osd-op-timeout=X` command line option when starting rbd-nbd (you can pass any ceph client specific config option this way)
3) via `export CEPH_ARGS="--rados-osd-op-timeout=X"`

But as it was said earlier it does not look like it will solve the issue.

pkalever:
> @pkalever Forgot to mention it earlier, but actually, the rbd-nbd does not require any modifications if you want to change `rados_osd_op_timeout`. As it is a ceph-specific config option there are several ways you can pass it to the rbd-nbd's rados client:
> 
> 1. set it in ceph.conf
> 2. provide as `--rados-osd-op-timeout=X` command line option when starting rbd-nbd (you can pass any ceph client specific config option this way)
> 3. via `export CEPH_ARGS="--rados-osd-op-timeout=X"`
> 
> But as it was said earlier it does not look like it will solve the issue.

Yes, but thanks for mentioning the various ways to pass the option, that helps me someday :-)

@trociny we also think that it might be possible that the IO is cached at the filesystem or the block layer level. And to flush the data from FS->Block->RBD-NBD->CEPH(OSD) we are thinking something like this https://github.com/ceph/ceph-csi/pull/2313#discussion_r675575745

It will be helpful to get your thoughts :-)

trociny:
> @trociny we also think that it might be possible that the IO is cached at the filesystem or the block layer level. And to flush the data from FS->Block->RBD-NBD->CEPH(OSD) we are thinking something like this [#2313 (comment)](https://github.com/ceph/ceph-csi/pull/2313#discussion_r675575745)

As @idryomov already mentioned, when detaching the rbd-nbd stops accepting io from the filesystem layer (and then waits for outstanding io to complete). So flush/freeze is not necessary here: the cached io will just wait for the re-attach, and actually flushing will increase the detach time. It might make sense to flush caches for safety reasons, just in case the re-attach fails. But I am not sure how useful it is.

pkalever:
> As @idryomov already mentioned, when detaching the rbd-nbd stops accepting io from the filesystem layer (and then waits for outstanding io to complete). So flush/freeze is not necessary here: the cached io will just wait for the re-attach, and actually flushing will increase the detach time. It might make sense to flush caches for safety reasons, just in case the re-attach fails. But I am not sure how useful it is.

I understand that on a detach rbd-nbd will stop accepting IO and will wait for pending IO to finish, but what happens to the write accepted at the FS and block layers but not reached to rbd-nbd daemon yet?


Yes, data consistency is priority 1, so flushing caches will ensure that the critical writes are not missed.

Maybe we should  follow the below order:
1. Freeze -> Ensure to flush the data in all the layers from FS till Ceph OSD
2. Detach -> Becomes safer, as the FS is in a frozen state, any further attempts to write/modify to the frozen filesystem will block waiting for the filesystem to be unfrozen
3. Reattach -> Make the device available again
4. Unfreeze -> Unblocks IO and continue
 
Also by freezing the mounts, we can allow snapshotting the images which are in the detached state too, the images will be data consistent anyways (probably at the quiesce hook script we need to tweak to avoid attempting to freeze if the mount point is already frozen)

Thanks again!

e2e/rbd.go
```diff
@@ -465,6 +465,21 @@ var _ = Describe("RBD", func() {
...
 
+				appOpt := metav1.ListOptions{
+					LabelSelector: fmt.Sprintf("app=%s", app.Name),
+				}
+				// TODO: Remove this once we ensure that rbd-nbd can sync data
+				// from Filesystem layer to backend rbd image as part of its
+				// detach or SIGTERM signal handler
+				_, stdErr, err := execCommandInPod(
+					f,
+					fmt.Sprintf("cd %s && sync", app.Spec.Containers[0].VolumeMounts[0].MountPath),
```
nixpanic: Not sure if changing the directory makes a difference. Consider running `sync <mountpath>` instead?

pkalever: Sure, that looks like a good idea.

Madhu-1 (approved):

nixpanic (approved):

pkalever:
ci/centos/upgrade-tests-cephfs:

Hit by: https://github.com/ceph/ceph-csi/issues/2264
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/upgrade-tests-cephfs/detail/upgrade-tests-cephfs/2483/pipeline

pkalever:
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.19/runs/2253/nodes/98/log/?start=0

Test: STEP: perform IO on rbd-nbd volume after nodeplugin restart

```

Jul 29 14:19:42.784: INFO: ExecWithOptions {Command:[/bin/sh -c rpm -qa | grep rbd-nbd] Namespace:cephcsi-e2e-aab37efa1a91 PodName:csi-rbdplugin-nlwxm ContainerName:csi-rbdplugin Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Jul 29 14:19:42.784: INFO: >>> kubeConfig: /root/.kube/config
Jul 29 14:19:43.532: INFO: rbd-nbd package version: rbd-nbd-16.2.5-0.el8.x86_64

Jul 29 14:19:43.538: INFO: ExecWithOptions {Command:[/bin/sh -c ps -eo 'cmd' | grep [r]bd-nbd] Namespace:cephcsi-e2e-aab37efa1a91 PodName:csi-rbdplugin-nlwxm ContainerName:csi-rbdplugin Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Jul 29 14:19:43.538: INFO: >>> kubeConfig: /root/.kube/config
Jul 29 14:19:43.749: INFO: rbd-nbd process is not running yet: command terminated with exit code 1
[...]
Jul 29 14:29:43.972: INFO: ExecWithOptions {Command:[/bin/sh -c ps -eo 'cmd' | grep [r]bd-nbd] Namespace:cephcsi-e2e-aab37efa1a91 PodName:csi-rbdplugin-nlwxm ContainerName:csi-rbdplugin Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Jul 29 14:29:43.972: INFO: >>> kubeConfig: /root/.kube/config
Jul 29 14:29:44.217: INFO: rbd-nbd process is not running yet: command terminated with exit code 1
Jul 29 14:29:44.217: FAIL: timed out waiting for the rbd-nbd process: rbd-nbd process is not running yet: command terminated with exit code 1

Logs:

Jul 29 14:29:44.438: INFO: Logs of cephcsi-e2e-aab37efa1a91/csi-rbdplugin-nlwxm:csi-rbdplugin on node minikube

Jul 29 14:29:44.438: INFO: STARTLOG
[...]
I0729 13:19:40.538141   35700 rbd_healer.go:75] sending nodeStageVolume for volID: 0001-0024-afaacd7b-f981-4b72-be41-de4915138654-0000000000000001-96252ac2-f06f-11eb-90eb-46e64d0ef0dd, stagingPath: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-296fd51d-a4b2-4f04-9605-d7306e49c3e7/globalmount
[...]
I0729 13:20:06.059451   35700 rbd_attach.go:280] rbd: map mon rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789
I0729 13:20:06.255082   35700 nodeserver.go:146] rbd volID: 0001-0024-afaacd7b-f981-4b72-be41-de4915138654-0000000000000001-96252ac2-f06f-11eb-90eb-46e64d0ef0dd was successfully attached to device: /dev/nbd0
```


[rbd-nbd-process-missing.log](https://github.com/ceph/ceph-csi/files/6906845/rbd-nbd-process-missing.log)

pkalever:
ci/centos/mini-e2e-helm/k8s-1.19 failed:

@nixpanic request to take a look at the log once.

I see the healer has done its job of attaching the rbd-nbd process from the rbd nodeplugin pod logs. But looks like the `ps | grep` couldn't find it?

Also why are the nodeplugin time stamp and e2e time stamp differ by an hour?

nixpanic:

Failed to restart rbd-nbd ([logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.19/detail/mini-e2e-helm_k8s-1.19/2263/pipeline/)):
```
[1mSTEP[0m: perform IO on rbd-nbd volume after nodeplugin restart
...
Jul 30 14:43:30.017: INFO: ExecWithOptions {Command:[/bin/sh -c ps -eo 'cmd' | grep [r]bd-nbd] Namespace:cephcsi-e2e-fc7f9c5015bd PodName:csi-rbdplugin-d6s9w ContainerName:csi-rbdplugin Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Jul 30 14:43:30.017: INFO: >>> kubeConfig: /root/.kube/config
Jul 30 14:43:30.269: INFO: rbd-nbd process is not running yet: command terminated with exit code 1
Jul 30 14:43:30.269: FAIL: timed out waiting for the rbd-nbd process: rbd-nbd process is not running yet: command terminated with exit code 1
```

nixpanic:

Failed to deploy CephFS provisioner ([logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/upgrade-tests-cephfs/detail/upgrade-tests-cephfs/2493/pipeline/)):
```
Jul 30 14:30:03.102: INFO: ExecWithOptions {Command:[/bin/sh -c ceph auth get-or-create-key client.cephcsi-cephfs-provisioner mon 'allow r' mgr 'allow rw' osd 'allow rw tag cephfs metadata=*'] Namespace:rook-ceph PodName:rook-ceph-tools-7467d8bf8-4j8pq ContainerName:rook-ceph-tools Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Jul 30 14:30:03.102: INFO: >>> kubeConfig: /root/.kube/config
Jul 30 14:30:10.121: FAIL: failed to create user cephcsi-cephfs-provisioner with error etcdserver: request timed out
```

Issue #2199, opened 2021-06-23
[Tracker ] Move listed features from Alpha -> Beta and Beta-> GA


humblec:
# Describe the feature you'd like to have #

At present, we have listed some features in Alpha state, however some of them are qualified already and also deployed in production setups.  we have to lift these features from Alpha/Beta to Beta/GA state in upcoming Ceph CSI releases. 

`RBD`

||Features   | Feature Status | CSI Driver Version | CSI Spec Version | Ceph Cluster Version | Kubernetes Version |
| ------ | --------------------------------------------------------- | -------------- | ------------------ | ---------------- | -------------------- | ------------------ |
|Provision File Mode ROX volume from snapshot              | Alpha          | >= v3.0.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.17.0         |
| Provision File Mode ROX volume from another volume        | Alpha          | >= v3.0.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.16.0         |
 | Provision Block Mode ROX volume from snapshot             | Alpha          | >= v3.0.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.17.0         |
| Provision Block Mode ROX volume from another volume       | Alpha          | >= v3.0.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.16.0         |
 | Creating and deleting snapshot                            | Alpha          | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.17.0         |
| Provision volume from snapshot                            | Alpha          | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.17.0         |
| Provision volume from another volume                      | Alpha          | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.16.0         |
| Expand volume                                             | Beta           | >= v2.0.0          | >= v1.1.0        | Nautilus (>=14.0.0)  | >= v1.15.0         |
| Metrics Support                                           | Beta           | >= v1.2.0          | >= v1.1.0        | Nautilus (>=14.0.0)  | >= v1.15.0         |
 Topology Aware Provisioning Support                       | Alpha          | >= v2.1.0          | >= v1.1.0        | Nautilus (>=14.0.0)  | >= v1.14.0         |

`CephFS`

[...]

PR #2304, opened 2021-07-19
doc: lift snapshot creation/deletion and new volume creation support to Beta

humblec:
Since kubernetes 1.17 snapshot functionality is at Beta state
and external snapshotter 3.0.3. Since v4.0.0 of snapshotter controller
and kubernetes v1.20, the snapshot functionality has been lifted to GA.
Ceph CSI drivers have been supporting this feature for last few releases
and users are heavily using this feature since then.

We also have good amount of e2e test case which cover snapshot creation
and deletion and other operations on the same. With that, this PR
proposes of lifting this feature support to Beta with v3.4.0 version.

This PR also lift clone functionality to Beta stage
Updates: https://github.com/ceph/ceph-csi/issues/2199

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ for its support details.
 |        | Provision Block Mode ROX volume from another volume       | Alpha          | >= v3.0.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.16.0         |
 |        | Creating and deleting snapshot                            | Beta           | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.17.0         |
 |        | Provision volume from snapshot                            | Beta           | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.17.0         |
-|        | Provision volume from another volume                      | Alpha          | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.16.0         |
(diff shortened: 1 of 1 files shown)


nixpanic:
Cloning and restoring from snapshot with thick-provisioning enabled need a newer Ceph version. Older versions did not copy zero-filled contents, but made the volumes thin-provisioned. Is that something that should be noted somewhere?

humblec:
> Cloning and restoring from snapshot with thick-provisioning enabled need a newer Ceph version. Older versions did not copy zero-filled contents, but made the volumes thin-provisioned. Is that something that should be noted somewhere?

@nixpanic good point.. I think we can keep a seperate entry for thick provisioning cloning and restoring in the matrix and mark existing ones as thin provisioned . wdyt?

humblec:
> > Cloning and restoring from snapshot with thick-provisioning enabled need a newer Ceph version. Older versions did not copy zero-filled contents, but made the volumes thin-provisioned. Is that something that should be noted somewhere?
> 
> @nixpanic good point.. I think we can keep a seperate entry for thick provisioning cloning and restoring in the matrix and mark existing ones as thin provisioned . wdyt?

or keep a diclaimer under the fields we have in the matrix and add a note below the matrix about this, that can also be done.

humblec:
@nixpanic as discussed , we will add a seperate section for non CSI spec specific features in a seperate matrix. With that in mind, this is good to go, ptal . Thanks :+1:

nixpanic (approved):

yati1998 (approved):

PR #2306, opened 2021-07-20
doc: explicitly mention volume/pv metrics feature

humblec:
we have volume PV metrics support available in our driver along
with the grpc metrics support (EnableGRPCMetrics) we added in between
for csi operations. The latter is getting deprecated and the current
mention in the support matrix on metrics support confuse many.
This PR explictly mention this support in the docs to volume/PV metrics


yati1998 (approved):

nixpanic (approved):

humblec:
Adding `ready-to-merge` as discussed in the CSI call and based on the change it is bringing in

PR #2313, opened 2021-07-22
deploy: adjust terminationGracePeriodSeconds to help rbd-nbd

pkalever:

As part of pod termination, Kubernetes initially sends SIGTERM to rbd-nbd and waits for ~30 sec i.e default `terminationGracePeriodSeconds` and then sends SIGKILL as part of ungraceful termination sequence. This is where some of the IO goes unhandled and mount point may turn out to read-only.

rbd-nbd expects some time for it to handle pending IO after SIGTERM is issued, currently at the rados level `rados_osd_op_timeout` is unlimited (wait forever), this means we never know when the IO will finish, if rbd-nbd can set the IO ops some timeout via `rados_osd_op_timeout` (in the rbd-nbd) then its possible for us at ceph-csi to exactly define time to wait for rbd-nbd at rbd nodeplugin.

But till we have the fix in rbd-nbd, to avoid the frequency of hitting such issue at the csi level, we should consider increasing `terminationGracePeriodSeconds` to a good value from the current 30 sec.

`terminationGracePeriodSeconds` field of a Pod spec instructs Kubernetes to wait for a given number of seconds before killing a container after termination. This leaves the processes time to free resources and flush any important data.

updates: #2204 

### DIFF
--- a/deploy/rbd/kubernetes/csi-rbdplugin.yaml
+++ b/deploy/rbd/kubernetes/csi-rbdplugin.yaml
@@ -19,6 +19,7 @@ spec:
       # to use e.g. Rook orchestrated cluster, and mons' FQDN is
       # resolved through k8s service, set dns policy to cluster first
       dnsPolicy: ClusterFirstWithHostNet
+      terminationGracePeriodSeconds: 180
       containers:
         - name: driver-registrar
           # This is necessary only for systems with SELinux, where
--- a/charts/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
+++ b/charts/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
@@ -35,6 +35,7 @@ spec:
       # to use e.g. Rook orchestrated cluster, and mons' FQDN is
       # resolved through k8s service, set dns policy to cluster first
       dnsPolicy: ClusterFirstWithHostNet
+      terminationGracePeriodSeconds: {{ .Values.nodeplugin.terminationGracePeriodSeconds }}
       containers:
         - name: driver-registrar
           # This is necessary only for systems with SELinux, where
--- a/charts/ceph-csi-rbd/values.yaml
+++ b/charts/ceph-csi-rbd/values.yaml
@@ -53,6 +53,7 @@ nodeplugin:
   priorityClassName: system-node-critical
   # if you are using rbd-nbd client set this value to OnDelete
   updateStrategy: RollingUpdate
+  terminationGracePeriodSeconds: 180
 
   httpMetrics:
     # Metrics only available for cephcsi/cephcsi => 1.2.0


pkalever:
@Madhu-1 @nixpanic IMHO, we need to have this at `3.4.0`. Thanks!

deploy/rbd/kubernetes/csi-rbdplugin.yaml
```diff
@@ -19,6 +19,7 @@ spec:
       # to use e.g. Rook orchestrated cluster, and mons' FQDN is
       # resolved through k8s service, set dns policy to cluster first
       dnsPolicy: ClusterFirstWithHostNet
+      terminationGracePeriodSeconds: 600
```
idryomov: Although 600 seconds is probably too long, increasing this timeout from the default 30 seconds would certainly help in many cases.  But, as mentioned in https://github.com/ceph/ceph-csi/pull/2298, it is just a workaround and can still lead to lost I/O and user workloads getting stuck.

At the risk of sounding like a broken record: with `rbd-nbd` mounter, the nodeplugin pod is no longer a "simple" CSI pod where all of the I/O logic/state is contained in the kernel.  It is in the I/O path and must be treated specially.  For upgrade in particular, it feels like a better approach would be to ask k8s to not schedule any new nodeplugin-dependent pods on the node, gracefully drain existing or already scheduled nodeplugin-dependent pods and then upgrade the nodeplugin pod.  This seems less error prone and more general -- for example, it would also work for `ceph-fuse` which doesn't have an equivalent of `rbd-nbd attach` (at least AFAIK).

pkalever: > Although 600 seconds is probably too long, increasing this timeout from the default 30 seconds would certainly help in many cases. But, as mentioned in #2298, it is just a workaround and can still lead to lost I/O and user workloads getting stuck.

Thanks @idryomov 

Yes, it is also good to understand that the probability of holding IO for longer than 600 sec is very less, and there is no way to avoid this completely as far as I understand, unless we agree to wait forever, which is very impractical here IMO.

> At the risk of sounding like a broken record: with `rbd-nbd` mounter, the nodeplugin pod is no longer a "simple" CSI pod where all of the I/O logic/state is contained in the kernel. It is in the I/O path and must be treated specially. For upgrade in particular, it feels like a better approach would be to ask k8s to not schedule any new nodeplugin-dependent pods on the node, gracefully drain existing or already scheduled nodeplugin-dependent pods and then upgrade the nodeplugin pod. This seems less error prone and more general -- for example, it would also work for `ceph-fuse` which doesn't have an equivalent of `rbd-nbd attach` (at least AFAIK).

Yes. The rbd nodeplugin is a daemon-set and hence with the daemon-sets upgrade strategy the ordering of old and new pods going down and coming up respectively will be promising.

```
DaemonSet Update Strategy

DaemonSet has two update strategy types:

OnDelete:
With OnDelete update strategy, after you update a DaemonSet template, new DaemonSet pods will
only be created when you manually delete old DaemonSet pods. This is the same behavior of
DaemonSet in Kubernetes version 1.5 or before.

RollingUpdate:
This is the default update strategy.
With RollingUpdate update strategy, after you update a DaemonSet template, old DaemonSet pods
will be killed, and new DaemonSet pods will be created automatically, in a controlled fashion. At most
one pod of the DaemonSet will be running on each node during the whole update process.
```

idryomov: Yes, `OnDelete` would be part of what I'm thinking about.  It looks like most of this is already documented at https://github.com/ceph/ceph-csi/blob/devel/docs/ceph-csi-upgrade.md but currently it is a manual process.  Has any thought been put into automating this in some higher-level operator (e.g. OCS operator)?

I mean, upgrade the most important (perhaps even the only?) use case we are chasing here, right?  We can't do much about an ungraceful restart of the nodeplugin pod, but automating upgrades seems feasible to me.

pkalever: I'm not up to date about the automation effort.
@Madhu-1 @nixpanic can you please chime in?

nixpanic:
> rbd-nbd expects some time for it to handle pending IO after SIGTERM is issued, currently at the rados level `rados_osd_op_timeout` is unlimited (wait forever), this means we never know when the IO will finish, if rbd-nbd can set the IO ops some timeout via `rados_osd_op_timeout` (in the rbd-nbd) then its possible for us at ceph-csi to exactly define time to wait for rbd-nbd at rbd nodeplugin.

If this is the issue in the test, would calling `sync` after writing the file be a solution?

deploy/rbd/kubernetes/csi-rbdplugin.yaml
nixpanic: we have seen issues where PodNetworking (as opposed to HostNetworking) is a problem. As that changes the (Linux) network namespace of the connections, restarting the ceph-csi pods might prevent further operations on the mountpoints (basically prevents unmounting, I thought). For this to work out, updates might require to drain all Pods that use Ceph-CSI volumes from the node where the update is taking place. To me it is unclear what component should be coordinating the update of Ceph-CSI, often it is Rook that deploys Ceph-CSI, so that is the main candidate from my understanding.

Draining application Pods before updating is quite a big hammer, but it would be an approach that works for this as well.

pkalever:
> > rbd-nbd expects some time for it to handle pending IO after SIGTERM is issued, currently at the rados level `rados_osd_op_timeout` is unlimited (wait forever), this means we never know when the IO will finish, if rbd-nbd can set the IO ops some timeout via `rados_osd_op_timeout` (in the rbd-nbd) then its possible for us at ceph-csi to exactly define time to wait for rbd-nbd at rbd nodeplugin.
> 
> If this is the issue in the test, would calling `sync` after writing the file be a solution?

Yes, we need to ensure that we are flushing everything pending (as part of handling SIGTERM) something like what we have done before with tcmu-runner here: https://github.com/open-iscsi/tcmu-runner/blob/master/rbd.c#L1478

I have opened a tracker to double-check the code that we are not assuming anything here: https://tracker.ceph.com/issues/51820

@idryomov @trociny. might have more insight.

nixpanic:
I really only meant `sync` in our e2e tests, as I expect rbd-nbd to flush its devices before exiting. #2204 is an instability in the tests that might be fixed that way.

For a real world workload, applications might continuously write, so at one point there needs to be a stop in accepting I/O, and caching it for a while before the next rbd-nbd has taken over. Device timeouts before the filesystem receives an error and goes read-only can usually be specified under `/sys/block/<bdev>/device/...` I think. Or maybe rbd-nbd has a parameter for it? Tuning that might be needed too?

deploy/rbd/kubernetes/csi-rbdplugin.yaml
idryomov: Yes, our recent troubles with pod networking is why I brought this up.  While tweaking timeouts might help `rbd-nbd` in this specific case, this is a general problem and we need a general solution.  I agree that draining is a quite big hammer, but it seems to be the only approach that is in line with k8s.

That said I'm not really a k8s person, so I may be missing something.  Just trying to avoid the pile up of ad-hoc workarounds that don't really solve the problem...

idryomov:
> I expect rbd-nbd to flush its devices before exiting

When rbd-nbd receives `SIGTERM`, it stops accepting I/O.  I think the current assumption that Prasanna is working off of is that k8s sends `SIGKILL` before `rbd-nbd` is able to process all outstanding in-flight I/O.

deploy/rbd/kubernetes/csi-rbdplugin.yaml
pkalever: @idryomov

In a short discussion with @nixpanic we came across the below idea, and we need your thoughts on this :-)

On the other comment, Niels thinks that the data is probably not reaching the rbd-nbd maybe it might be possible that writes are cached at filesystem or in the block layer too, that way issuing the sync from the mount might help push them to rbd-nbd tray at least. But again `sync ` might be possible to exclusively run in our e2e, but we cannot ensure that to happen with real workloads. For this, to happen probably we should pass `-o sync` at mount time, but that will reduce the IO performance drastically.

We are curious to know, just like the quiesce hook which executes `fsfreeze` on the mount before starting to take a snapshot of an image and then `fsfreeze -u` after the snapshot is taken, can we run the quiesce hook for attach and detach operations? i.e as part of the quiescing, call `fsfreeze` before the detach is called, and then allow reattach the process then invoke `fsfreeze -u`?

We understand that the freezing + flushing of the pending inflight IO takes time too, but we should be in a better place to ensure that the data is not cached/waiting at the other layers. So if that causes pods to get terminated before flushing data is complete at rbd-nbd, there is a good reason to extend the default pod termination grace timeout (or use the application lifecycle hooks)

humblec:
```
As part of pod termination, Kubernetes initially sends SIGTERM to rbd-nbd and waits for ~30 sec i.e default terminationGracePeriodSeconds and then sends SIGKILL as part of ungraceful termination sequence. This is where some of the IO goes unhandled and mount point may turn out to read-only.

rbd-nbd expects some time for it to handle pending IO after SIGTERM is issued, currently at the rados level rados_osd_op_timeout is unlimited (wait forever), this means we never know when the IO will finish, if rbd-nbd can set the IO ops some timeout via rados_osd_op_timeout (in the rbd-nbd) then its possible for us at ceph-csi to exactly define time to wait for rbd-nbd at rbd nodeplugin.

But till we have the fix in rbd-nbd, to avoid the frequency of hitting such issue at the csi level, we should consider increasing terminationGracePeriodSeconds to a good value from the current 30 sec.
```
I am sorry to say that, I am completely against this change.  That said, we should not increase the timeout to this proposed higher value. The main reason being:

*) This means, an application hang of `180` seconds+ ( 3minutes+) before kubelet can make next scheduling decision
*) The change is * NOT * NBD specific at all, we are pushing node plugin container timeout to a higer value which  affect all other scenarios even in absense of 'NBD' shares in place !
*) Also, here we are fixing a value blindly without any assurance of it can solve the issue.

As an additional note, a pod can go wrong in different  situations  ( like memory/cpu/resource crunch) ..etc and just not limited to `upgrade time`.

pkalever:
>>As part of pod termination, Kubernetes initially sends SIGTERM to rbd-nbd and waits for ~30 sec i.e default >>terminationGracePeriodSeconds and then sends SIGKILL as part of ungraceful termination sequence. This is where some of >>the IO goes unhandled and mount point may turn out to read-only.

>>rbd-nbd expects some time for it to handle pending IO after SIGTERM is issued, currently at the rados level >>rados_osd_op_timeout is unlimited (wait forever), this means we never know when the IO will finish, if rbd-nbd can set the IO >>ops some timeout via rados_osd_op_timeout (in the rbd-nbd) then its possible for us at ceph-csi to exactly define time to wait >>for rbd-nbd at rbd nodeplugin.

>>But till we have the fix in rbd-nbd, to avoid the frequency of hitting such issue at the csi level, we should consider increasing >>terminationGracePeriodSeconds to a good value from the current 30 sec.

>I am sorry to say that, I am completely against this change. That said, we should not increase the timeout to this proposed higher >value. The main reason being:

>*) This means, an application hang of 180 seconds+ ( 3minutes+) before kubelet can make next scheduling decision
>*) The change is * NOT * NBD specific at all, we are pushing node plugin container timeout to a higer value which affect all other >scenarios even in absense of 'NBD' shares in place !

I hope we are clear that the pod deletion is delayed only until the rbd-nbd process is waiting on IO, which means let's say for example if the rbd-nbd process is done (flushing the pending IO) in 2sec, then the pod can continue to terminate on the very next second.

This delay doesn't mean anything for non-rbd-nbd mounters (krbd).

So we only wanted to wait here when rbd-nbd processes are waiting for IO to finish at the OSD level. Because not waiting might cost us data loss.

>*) Also, here we are fixing a value blindly without any assurance of it can solve the issue.

We are fixing the time bounded by pod eviction timeout i.e. 5m.

Yes, of course, this was just to add some time for rbd-nbd to finish its job and I never said this is the real fix, it was not even a complete workaround solution too. This was the least that we can do while our solution is still baking.

It might also be very much possible the inflight IO is still pending at the fs or block layers. To rule this out completely we are working on ways to fix it., some initial idea is updated here #2313 (comment)

Fixing this is going to take some time and we shouldn't expect that to happen at 3.4.0 we can target it for the next release.

That all said, no objections from me waiting for the actual fix.

humblec:
> 
> > *) This means, an application hang of 180 seconds+ ( 3minutes+) before kubelet can make next scheduling decision
> > *) The change is * NOT * NBD specific at all, we are pushing node plugin container timeout to a higer value which affect all other >scenarios even in absense of 'NBD' shares in place !
> 
> I hope we are clear that the pod deletion is delayed only until the rbd-nbd process is waiting on IO, which means let's say for example if the rbd-nbd process is done (flushing the pending IO) in 2sec, then the pod can continue to terminate on the very next second.
> 
> This delay doesn't mean anything for non-rbd-nbd mounters (krbd).

Not true :) It is equally applicable to **any** process in the nodeplugin pod and its `responsiveness` to the outside events.

> 
> So we only wanted to wait here when rbd-nbd processes are waiting for IO to finish at the OSD level. Because not waiting might cost us data loss.

The dataloss and `RO` mount is a concern.

pkalever:
> > I hope we are clear that the pod deletion is delayed only until the rbd-nbd process is waiting on IO, which means let's say for example if the rbd-nbd process is done (flushing the pending IO) in 2sec, then the pod can continue to terminate on the very next second.
> > This delay doesn't mean anything for non-rbd-nbd mounters (krbd).
> 
> Not true :) It is equally applicable to **any** process in the nodeplugin pod and its `responsiveness` to the outside events.
>

I think there is a difference in understanding my statement. Try to answer this, and I hope you will get it. What process can be in stuck/hung in the state for more than ~30sec in a non-rbd-nbd mounter scenario?

humblec:
>I think there is a difference in understanding my statement. Try to answer this, and I hope you will get it. What process can be in stuck/hung in the state for more than ~30sec in a non-rbd-nbd mounter scenario?
>

@pkalever  we never know and thats why we have that timeout in first place even before nbd. Increasing that is risky as kublet get stalled during this period for next action.

humblec:
I think one solution we could think of here is possibilities of `prestop` hook.  If nodeplugin want to do any 'bookkeeping' kind of work, the right approach would be looking at getting it done by the hook mechanism. The hook  is called immediately before a container is terminated due to an API request or management event such as a liveness/startup probe failure, preemption, resource contention and others. From there the grace termination period count down starts. Thoughts ?

PR #2316, opened 2021-07-23
rbd: rename encryption metadata keys to enable mirroring

### LINKED ISSUE #2219: RBD: Image metadata created by cephcsi on rbd images is not  mirrored or copied  when clone is created.

nixpanic:
The new MigrateMetadata() function can be used to get the metadata of an
image with a deprecated and new key. Renaming metadata keys can be done
easily this way.

A default value will be set in the image metadata when it is missing
completely. But if the deprecated key was set, the data is stored under
the new key and the deprecated key is removed.

RBD image metadata keys that start with '.rbd' are expected to be
internal to RBD itself and are not mirrored to remote sites. Renaming
the keys (dropping the '.rbd' prefix) and using the new
MigrateMetadata() function now makes the keys available on remote sites
too.

Closes: #2219

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/rbd/encryption.go
+++ b/internal/rbd/encryption.go
@@ -53,16 +53,18 @@ const (
 	rbdImageRequiresEncryption = rbdEncryptionState("requiresEncryption")
 
 	// image metadata key for encryption.
-	encryptionMetaKey = ".rbd.csi.ceph.com/encrypted"
+	encryptionMetaKey    = "rbd.csi.ceph.com/encrypted"
+	oldEncryptionMetaKey = ".rbd.csi.ceph.com/encrypted"
 
 	// metadataDEK is the key in the image metadata where the (encrypted)
 	// DEK is stored.
-	metadataDEK = ".rbd.csi.ceph.com/dek"
+	metadataDEK    = "rbd.csi.ceph.com/dek"
+	oldMetadataDEK = ".rbd.csi.ceph.com/dek"
 )
 
 // checkRbdImageEncrypted verifies if rbd image was encrypted when created.
 func (ri *rbdImage) checkRbdImageEncrypted(ctx context.Context) (rbdEncryptionState, error) {
-	value, err := ri.GetMetadata(encryptionMetaKey)
+	value, err := ri.MigrateMetadata(oldEncryptionMetaKey, encryptionMetaKey, string(rbdImageEncryptionUnknown))
 	if errors.Is(err, librbd.ErrNotFound) {
 		util.DebugLog(ctx, "image %s encrypted state not set", ri)
 
@@ -317,7 +319,7 @@ func (ri *rbdImage) FetchDEK(volumeID string) (string, error) {
 		return "", fmt.Errorf("volume %q can not fetch DEK for %q", ri, volumeID)
 	}
 
-	return ri.GetMetadata(metadataDEK)
+	return ri.MigrateMetadata(oldMetadataDEK, metadataDEK, "")
 }
 
 // RemoveDEK does not need to remove the DEK from the metadata, the image is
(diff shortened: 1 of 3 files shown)


internal/rbd/rbd_util.go
```diff
@@ -1593,23 +1593,62 @@ func (ri *rbdImage) SetMetadata(key, value string) error {
...
+		return value, nil
+	} else if !errors.Is(err, librbd.ErrNotFound) {
+		return "", err
+	}
+
+	// migrate contents from oldKey to newKey
+	removeOldKey := false
+	value, err = ri.GetMetadata(oldKey)
+	if errors.Is(err, librbd.ErrNotFound) {
+		removeOldKey = true
```
Madhu-1: this should be set if `err==nil`?

internal/rbd/encryption.go
```diff
@@ -53,16 +53,18 @@ const (
...
 
 	// image metadata key for encryption.
-	encryptionMetaKey = ".rbd.csi.ceph.com/encrypted"
+	encryptionMetaKey    = "csi.ceph.com/encrypted"
+	oldEncryptionMetaKey = ".rbd.csi.ceph.com/encrypted"
 
 	// metadataDEK is the key in the image metadata where the (encrypted)
 	// DEK is stored.
-	metadataDEK = ".rbd.csi.ceph.com/dek"
+	metadataDEK    = "csi.ceph.com/dek"
```
Madhu-1: Do you think to keep the key consistent we need to change https://github.com/ceph/ceph-csi/blob/4e6d9be82629d5d12dc88963acfcd1f48ffc3676/internal/rbd/rbd_util.go#L65-L66 also?

Madhu-1 (changes requested):
Need E2E change also https://github.com/ceph/ceph-csi/blob/4e6d9be82629d5d12dc88963acfcd1f48ffc3676/e2e/rbd_helper.go#L497

internal/rbd/encryption.go
nixpanic: hmm, in the commit text of 332a47a100e you mentioned the name of the key without `rbd.` prefix, but in the code only the initial `.` was dropped. I don't really care that much, but consistency is probably nice, will add `rbd.` in front  of the key.

internal/rbd/rbd_util.go
nixpanic: yes!

Madhu-1 (dismissed):

internal/rbd/rbd_util.go
Madhu-1: Need to use `switch` to fix static check failure.

internal/rbd/encryption.go
Madhu-1: Looks like I missed that one. the only restriction is that we should not have key starting with `.rbd`.  `rbd` is fine :+1:

internal/rbd/rbd_util.go
nixpanic: Indeed, but just rewriting to a switch/case does not work either it seems (esimonov/ifshort#17). Now modified the code to work around both linter errors.

humblec:
lgtm. @Madhu-1 any further comments here?

Madhu-1 (approved):

humblec (approved):

PR #2348, opened 2021-08-03
ci: disable commitlint mergify rule

Madhu-1:
currently, PR merging is blocked due to the commitlint issue. disabling commitlint on the release branches now. more details at
https://github.com/ceph/ceph-csi/pull/2342


Yuggupta27 (approved):
LGTM

yati1998 (approved):

humblec:
@Madhu-1 dont we need to just disable it on `release-*` branches ?

Madhu-1:
> @Madhu-1 dont we need to just disable it on `release-*` branches ?

We have different mergify rules for different branches. This PR is disabling on the release branches only.

nixpanic (approved):

humblec:
> > @Madhu-1 dont we need to just disable it on `release-*` branches ?
> 
> We have different mergify rules for different branches. This PR is disabling on the release branches only.

@Madhu-1 sure, I was asking that, cant we keep it disabled only in a particular branch and its section , for example : on `release-34` just disable it on https://github.com/ceph/ceph-csi/blob/release-v3.4/.mergify.yml#L294 and thus ... instead of disabling it on `devel` for all the branches.

Madhu-1:
> > > @Madhu-1 dont we need to just disable it on `release-*` branches ?
> > 
> > 
> > We have different mergify rules for different branches. This PR is disabling on the release branches only.
> 
> @Madhu-1 sure, I was asking that, cant we keep it disabled only in a particular branch and its section , for example : on `release-34` just disable it on https://github.com/ceph/ceph-csi/blob/release-v3.4/.mergify.yml#L294 and thus ... instead of disabling it on `devel` for all the branches.

AFAIK this is not the way Mergify works. It always picks the configuration from the default branch. Adding rules to release branch won't help, if it works happy to make changes

nixpanic:
> AFAIK this is not the way Mergify works. It always picks the configuration from the default branch. Adding rules to release branch won't help, if it works happy to make changes

Yes, I think that is correct. We only need to adjust the configuration in the default branch (`devel` in our case). There is no use in changing the configuration in any of the other branches.

humblec (approved):

PR #2394, opened 2021-08-10
rbd: fix rbd-nbd io-timeout to never abort

pkalever:

* Fix rbd-nbd io-timeout to match reattach-timeout

With the tests at CI, it kind of looks like that the IO is timing out after 30 seconds (default with rbd-nbd). Since we have tweaked reattach-timeout to 300 seconds at ceph-csi, we need to explicitly set io-timeout on the device too, as it doesn't make any sense to keep `io-timeout < reattach-timeout`
                                                                            
Hence we set io-timeout for rbd nbd to 0. Specifying io-timeout 0 tells the nbd driver to not abort the request and instead see if it can be restarted on another socket.                                                

---
* At our e2e, use io-timeout conditionally based on kernel version             
                                                                      
We need https://www.mail-archive.com/linux-block@vger.kernel.org/msg38060.html inorder to use `--io-timeout=0`. This patch is part of kernel v5.4. Since minikube doesn't have a v5.4 kernel yet, let's use io-timeout value conditionally based on the kernel version at our e2e.                       

---

Thanks @idryomov for the timeout suggestion.

Updates: #2204

Suggested-by: Ilya Dryomov <idryomov@redhat.com>

---

<details>

  progress or results

</details>

### DIFF
--- a/e2e/pod.go
+++ b/e2e/pod.go
@@ -406,3 +406,22 @@ func calculateSHA512sum(f *framework.Framework, app *v1.Pod, filePath string, op
 
 	return checkSum, nil
 }
+
+// getKernelVersionFromDaemonset gets the kernel version from the specified container.
+func getKernelVersionFromDaemonset(f *framework.Framework, ns, dsn, cn string) (string, error) {
+	selector, err := getDaemonSetLabelSelector(f, ns, dsn)
+	if err != nil {
+		return "", err
+	}
+
+	opt := metav1.ListOptions{
+		LabelSelector: selector,
+	}
+
+	kernelRelease, stdErr, err := execCommandInContainer(f, "uname -r", ns, cn, &opt)
+	if err != nil || stdErr != "" {
+		return "", err
+	}
+
+	return kernelRelease, nil
+}
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -57,6 +57,8 @@ var (
 	appBlockSmartClonePath = rbdExamplePath + "block-pod-clone.yaml"
 	snapshotPath           = rbdExamplePath + "snapshot.yaml"
 	defaultCloneCount      = 10
+
+	nbdMapOptions = "debug-rbd=20"
 )
 
 func deployRBDPlugin() {
@@ -178,6 +180,7 @@ func validateRBDImageCount(f *framework.Framework, count int, pool string) {
 var _ = Describe("RBD", func() {
 	f := framework.NewDefaultFramework("rbd")
 	var c clientset.Interface
+	var kernelRelease string
 	// deploy RBD CSI
 	BeforeEach(func() {
 		if !testRBD || upgradeTesting {
@@ -232,6 +235,27 @@ var _ = Describe("RBD", func() {
 			e2elog.Failf("failed to create node secret with error %v", err)
 		}
 		deployVault(f.ClientSet, deployTimeout)
+
+		// wait for provisioner deployment
+		err = waitForDeploymentComplete(rbdDeploymentName, cephCSINamespace, f.ClientSet, deployTimeout)
+		if err != nil {
+			e2elog.Failf("timeout waiting for deployment %s with error %v", rbdDeploymentName, err)
+		}
+
+		// wait for nodeplugin deamonset pods
+		err = waitForDaemonSets(rbdDaemonsetName, cephCSINamespace, f.ClientSet, deployTimeout)
+		if err != nil {
+			e2elog.Failf("timeout waiting for daemonset %s with error %v", rbdDaemonsetName, err)
+		}
+
+		kernelRelease, err = getKernelVersionFromDaemonset(f, cephCSINamespace, rbdDaemonsetName, "csi-rbdplugin")
+		if err != nil {
+			e2elog.Failf("failed to get the kernel version with error %v", err)
+		}
+		// default io-timeout=0, needs kernel >= 5.4
+		if !util.CheckKernelSupport(kernelRelease, nbdZeroIOtimeoutSupport) {
+			nbdMapOptions = "debug-rbd=20,io-timeout=330"
+		}
 	})
 
 	AfterEach(func() {
@@ -302,20 +326,6 @@ var _ = Describe("RBD", func() {
 
 	Context("Test RBD CSI", func() {
 		It("Test RBD CSI", func() {
-			By("checking provisioner deployment is running", func() {
-				err := waitForDeploymentComplete(rbdDeploymentName, cephCSINamespace, f.ClientSet, deployTimeout)
-				if err != nil {
-					e2elog.Failf("timeout waiting for deployment %s with error %v", rbdDeploymentName, err)
-				}
-			})
-
-			By("checking nodeplugin deamonset pods are running", func() {
-				err := waitForDaemonSets(rbdDaemonsetName, cephCSINamespace, f.ClientSet, deployTimeout)
-				if err != nil {
-					e2elog.Failf("timeout waiting for daemonset %s with error %v", rbdDaemonsetName, err)
-				}
-			})
-
 			// test only if ceph-csi is deployed via helm
 			if helmTest {
 				By("verify PVC and app binding on helm installation", func() {
@@ -412,7 +422,7 @@ var _ = Describe("RBD", func() {
 					nil,
 					map[string]string{
 						"mounter":    "rbd-nbd",
-						"mapOptions": "debug-rbd=20",
+						"mapOptions": nbdMapOptions,
 					},
 					deletePolicy)
 				if err != nil {
@@ -435,10 +445,6 @@ var _ = Describe("RBD", func() {
 			})
 
 			By("Resize rbd-nbd PVC and check application directory size", func() {
-				kernelRelease, err := util.GetKernelVersion()
-				if err != nil {
-					e2elog.Failf("failed to get kernel version with error %v", err)
-				}
 				if util.CheckKernelSupport(kernelRelease, nbdResizeSupport) {
 					err := deleteResource(rbdExamplePath + "storageclass.yaml")
 					if err != nil {
@@ -452,7 +458,7 @@ var _ = Describe("RBD", func() {
 						nil,
 						map[string]string{
 							"mounter":    "rbd-nbd",
-							"mapOptions": "debug-rbd=20",
+							"mapOptions": nbdMapOptions,
 						},
 						deletePolicy)
 					if err != nil {
@@ -497,7 +503,7 @@ var _ = Describe("RBD", func() {
 					nil,
 					map[string]string{
 						"mounter":    "rbd-nbd",
-						"mapOptions": "debug-rbd=20",
+						"mapOptions": nbdMapOptions,
 					},
 					deletePolicy)
 				if err != nil {
@@ -651,7 +657,7 @@ var _ = Describe("RBD", func() {
 					nil,
 					map[string]string{
 						"mounter":    "rbd-nbd",
-						"mapOptions": "debug-rbd=20",
+						"mapOptions": nbdMapOptions,
 						"encrypted":  "true",
 					},
 					deletePolicy)
@@ -1009,7 +1015,7 @@ var _ = Describe("RBD", func() {
 						map[string]string{
 							"imageFeatures": "layering,journaling,exclusive-lock",
 							"mounter":       "rbd-nbd",
-							"mapOptions":    "debug-rbd=20",
+							"mapOptions":    nbdMapOptions,
 						},
 						deletePolicy)
 					if err != nil {
--- a/e2e/rbd_helper.go
+++ b/e2e/rbd_helper.go
@@ -40,6 +40,28 @@ var nbdResizeSupport = []util.KernelVersion{
 	}, // standard 5.3+ versions
 }
 
+// To use `io-timeout=0` we need
+// www.mail-archive.com/linux-block@vger.kernel.org/msg38060.html
+// nolint:gomnd // numbers specify Kernel versions.
+var nbdZeroIOtimeoutSupport = []util.KernelVersion{
+	{
+		Version:      5,
+		PatchLevel:   4,
+		SubLevel:     0,
+		ExtraVersion: 0,
+		Distribution: "",
+		Backport:     false,
+	}, // standard 5.4+ versions
+	{
+		Version:      4,
+		PatchLevel:   18,
+		SubLevel:     0,
+		ExtraVersion: 305,
+		Distribution: ".el8",
+		Backport:     true,
+	}, // CentOS 8.4
+}
+
 func imageSpec(pool, image string) string {
 	if radosNamespace != "" {
 		return pool + "/" + radosNamespace + "/" + image
--- a/internal/rbd/rbd_attach.go
+++ b/internal/rbd/rbd_attach.go
@@ -48,9 +48,16 @@ const (
 	rbdUnmapCmdNbdMissingMap  = "rbd-nbd: %s is not mapped"
 	rbdMapConnectionTimeout   = "Connection timed out"
 
-	defaultNbdReAttachTimeout = 300
+	defaultNbdReAttachTimeout = 300 /* in seconds */
 
-	useNbdNetlink  = "try-netlink"
+	// The default way of creating nbd devices via rbd-nbd is through the
+	// legacy ioctl interface, to take advantage of netlink features we
+	// should specify `try-netlink` flag explicitly.
+	useNbdNetlink = "try-netlink"
+
+	// `reattach-timeout` of rbd-nbd is to tweak NBD_ATTR_DEAD_CONN_TIMEOUT.
+	// It specifies how long the device should be held waiting for the
+	// userspace process to come back to life.
 	setNbdReattach = "reattach-timeout"
 )
 
--- a/internal/rbd/rbd_attach.go
+++ b/internal/rbd/rbd_attach.go
@@ -49,6 +49,7 @@ const (
 	rbdMapConnectionTimeout   = "Connection timed out"
 
 	defaultNbdReAttachTimeout = 300 /* in seconds */
+	defaultNbdIOTimeout       = 0   /* do not abort the requests */
 
 	// The default way of creating nbd devices via rbd-nbd is through the
 	// legacy ioctl interface, to take advantage of netlink features we
@@ -59,6 +60,10 @@ const (
 	// It specifies how long the device should be held waiting for the
 	// userspace process to come back to life.
 	setNbdReattach = "reattach-timeout"
+
+	// `io-timeout` of rbd-nbd is to tweak NBD_ATTR_TIMEOUT. It specifies
+	// how long the IO should wait to get handled before bailing out.
+	setNbdIOTimeout = "io-timeout"
 )
 
 var hasNBD = false
@@ -256,6 +261,9 @@ func appendDeviceTypeAndOptions(cmdArgs []string, isNbd, isThick bool, userOptio
 		if !strings.Contains(userOptions, setNbdReattach) {
 			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("%s=%d", setNbdReattach, defaultNbdReAttachTimeout))
 		}
+		if !strings.Contains(userOptions, setNbdIOTimeout) {
+			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
+		}
 	}
 	if isThick {
 		// When an image is thick-provisioned, any discard/unmap/trim
@@ -279,6 +287,9 @@ func appendRbdNbdCliOptions(cmdArgs []string, userOptions string) []string {
 	}
 	if !strings.Contains(userOptions, setNbdReattach) {
 		cmdArgs = append(cmdArgs, fmt.Sprintf("--%s=%d", setNbdReattach, defaultNbdReAttachTimeout))
+	}
+	if !strings.Contains(userOptions, setNbdIOTimeout) {
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
 	}
 	if userOptions != "" {
 		options := strings.Split(userOptions, ",")


pkalever:
Thanks, @trociny for highlighting it.

pkalever:
Most of the tests stuck after 'return bare-metal machine' state.

internal/rbd/rbd_attach.go
```diff
@@ -239,6 +241,9 @@ func appendDeviceTypeAndOptions(cmdArgs []string, isNbd, isThick bool, userOptio
 		if !strings.Contains(userOptions, setNbdReattach) {
 			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("%s=%d", setNbdReattach, defaultNbdReAttachTimeout))
 		}
+		if !strings.Contains(userOptions, setNbdIOTimeout) {
+			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
```
Madhu-1, line 265: Question:
@pkalever will there be any impact on pod graceful termination and force termination because of this one?

pkalever: @Madhu-1 Nope this shouldn't hurt the termination sequence of the pod lifecycle. This timeout is supplied to the nbd driver via rbd-nbd process, It means how long the cmds should wait before timing out (like: when the rbd-nbd process is down, wait for it to serve IO)


But it's good to get some confirmation from @idryomov about the values we are using i.e reattach-timeout=300 and io-timeout=$reattach-timeout + 30, has any further objections.

cc: @lxbsz

nixpanic (dismissed):
makes sense to me, thanks!

Madhu-1 (commented):
LGTM will wait for response for https://github.com/ceph/ceph-csi/pull/2394/files#r686495332 and merge it

internal/rbd/rbd_attach.go
idryomov: This seems wrong to me.  I think this would cause in-flight requests to not be requeued when rbd-nbd is respawned,  see https://github.com/torvalds/linux/commit/2c272542baee2c3b9e8e3a260db81227ccefe8b5.  By setting `io-timeout` to the value of `reattach-timeout` or higher we would effectively disable that `nbd_requeue_cmd()` call.

idryomov (commented):
I haven't tested yet but I think we should always set `io-timeout` to 0.  This should keep the requeueing logic in action and work around the timeout handling bugs present in all kernels (where the old value just hangs around and the first mapping is different from subsequent mappings).

internal/rbd/rbd_attach.go
pkalever: I have tested the different settings of io-timeout values like greater than, less than, equal to reattach timeout value keeping reattach timeout at 300 fixed, and honestly with the latest nbd driver available at the master  I had not seen any difference in the io timeout/fail behavior.

Here are the observations captured: https://hackmd.io/@FFELd2RJT76Uqp-3M-e4zw/Sys8bLfeY

I'm planning to test the behaviour on the minikube kernel version to understand the behaviour between the kernels.

---

I hope now you might sense the need for https://github.com/ceph/ceph/pull/42609

Thanks!

CC: @trociny @lxbsz

internal/rbd/rbd_attach.go
```diff
@@ -48,10 +48,12 @@ const (
...
 
-	defaultNbdReAttachTimeout = 300
+	defaultNbdReAttachTimeout = 300 /* in seconds */
+	defaultNbdIOTimeout       = defaultNbdReAttachTimeout + 30
 
-	useNbdNetlink  = "try-netlink"
-	setNbdReattach = "reattach-timeout"
+	useNbdNetlink   = "try-netlink"
+	setNbdReattach  = "reattach-timeout"
+	setNbdIOTimeout = "io-timeout"
```
humblec, line 66: @pkalever can you please add some details about these timeout values as source code comments ?

internal/rbd/rbd_attach.go
idryomov: On the contrary, so far I don't see the need for https://github.com/ceph/ceph/pull/42609 at all.

I don't understand what you were testing though.  If you kill rbd-nbd with `reattach-timeout` set, eventually in-flight I/O would be failed no matter what the value of `io-timeout` is.  I think the test should have included `rbd device attach` at e.g. 3 minute mark with the expectation that nothing gets timed out, the filesystem stays read-write and dd continues to execute after reattach.  I would expect `io-timeout=30` to be disqualified immediately and with enough test rounds you should be able to see the difference between `io-timeout=300`/`io-timeout=330` and `io-timeout=0`.

idryomov: Nit: get dmesg dumps with raw jiffie timestamps (i.e. none of that `Aug12 06:52` and `+0.004259` stuff).  Given that we are investigating 30 second timeouts, a timestamp such as `Aug12 06:52` that doesn't include seconds at all is pretty useless.

pkalever: > On the contrary, so far I don't see the need for [ceph/ceph#42609](https://github.com/ceph/ceph/pull/42609) at all.

We don't do -osync writes from mount, which means we may end up with some IO pending at filesystem/block layers while the rbd-nbd goes detached and we shouldn't expect the pending IO in the filesystem/block layers to be handled as part of SIGTERM handler of rbd-nbd.

1. The quiesce will ensure that the data will be pushed to the rbd-nbd layer.
2. It will also block/wait the IO on the mount while the rbd-nbd is offline.
3. Also just in case if the process is not attached in the reattachtimeout, the copy is ensured to be consistent (corruption free)
4. Snapshots can be taken in a consistent way even when the rbd-nbd process is offline.

> 
> I don't understand what you were testing though. If you kill rbd-nbd with `reattach-timeout` set, eventually in-flight I/O would be failed no matter what the value of `io-timeout` is. I think the test should have included `rbd device attach` at e.g. 3 minute mark with the expectation that nothing gets timed out, the filesystem stays read-write and dd continues to execute after reattach.

I was testing a case where the attach doesn't happen in the reattach-timeout and wanted to note how IO behaves in various io-timeouts used.

I expect the process attach to handle the IO which is not timed out yet, but what about the IO which is timed out already (30 secs is too low, I have not seen them timing out in 30secs though)? We cannot expect the process to be back in 30 seconds and cope up with the pending IO. And we are uncertain how to handle the read-only mounts which are perhaps happening after the IO timeouts. Certainly, we want to avoid this situation.

> I would expect `io-timeout=30` to be disqualified immediately and with enough test rounds you should be able to see the difference between `io-timeout=300`/`io-timeout=330` and `io-timeout=0`.

I will test more with the dmesgs timestamps suggestions you mentioned. Also include tests with attach next time.

Thanks @idryomov.

idryomov: > > On the contrary, so far I don't see the need for [ceph/ceph#42609](https://github.com/ceph/ceph/pull/42609) at all.
> 
> We don't do -osync writes from mount, which means we may end up with some IO pending at filesystem/block layers while the rbd-nbd goes detached and we shouldn't expect the pending IO in the filesystem/block layers to be handled as part of SIGTERM handler of rbd-nbd.

Right, that I/O would get processed after rbd-nbd is reattached.

> 
>     1. The quiesce will ensure that the data will be pushed to the rbd-nbd layer.
> 
>     2. It will also block/wait the IO on the mount while the rbd-nbd is offline.
> 
>     3. Also just in case if the process is not attached in the reattachtimeout, the copy is ensured to be consistent (corruption free)
> 
>     4. Snapshots can be taken in a consistent way even when the rbd-nbd process is offline.

This is true but I don't see much value in smoothing out the "reattach didn't happen" case, especially if it is achieved by allowing more or less arbitrary code execution within the signal handler.  The only reason we are bothering with all of this (`rbd device attach`, `io-timeout`, etc) is to make reattach not just possible but also reliable.  It should "just" work and if something goes wrong, it is not that different from an abrupt termination (e.g. if rbd-nbd gets a SIGKILL from k8s because it doesn't react to SIGTERM in time or if rbd-nbd actually crashes due to a bug in librbd or the daemon itself).

Is there a reason for treating the "reattach didn't happen" case specially?

nixpanic: Currently this is the most important requirement

> 
> 2. It will also block/wait the IO on the mount while the rbd-nbd is offline.
> 

When the re-attach is not fast enough, or the filesystem decides do write to the block device while the rbd-nbd process is unavailable, the (ext4) filesystem detects a problem with the block device, and decides to become read-only. There is no reasonable way for a Pod to recover from that.

Before the rbd-nbd process does a detach, it would be best to call `fsfreeze` and block I/O from going to the block-device. Upon attach an unfreeze should be done. It is ideal if `rbd-nbd` support this, but we can consider an alternative where Ceph-CSI tries to handle it.

pkalever: Well honestly, I do not see a good reason why rbd-nbd shouldn't do it, the way like snapshots.

That said, ceph-csi cannot handle it, because although the rbd-nbd process and the application mount resides on the same node they both run in two different pods, so ceph-csi cannot coordinate b/w freeze and detach. Since the rbd-nbd receives the signal in any case, it can very well handle it.

If we have to do it from the `preStop` hook, at scale, it will be a lot of serializing of all the freezes tasks per PV. Since there will be one rbd-nbd process per volume, the fsfreeze will be run parallelly in case rbd-nbd can handle it, that way we will not hit the termination grace period of Kubernetes.

idryomov: > When the re-attach is not fast enough, or the filesystem decides do write to the block device while the rbd-nbd process is unavailable, the (ext4) filesystem detects a problem with the block device, and decides to become read-only. There is no reasonable way for a Pod to recover from that.

Even with `io-timeout` set to 0?  I don't think ext4 detects anything by itself.  IIUC the problem was that that we weren't setting `io-timeout` and the block layer was returning an error after 30 seconds.  This PR should resolve that.

As for what happens if reattach doesn't occur within `reattach-timeout`, if the filesystem is frozen, wouldn't the user application just hang in D state?  How would you recover from that?

pkalever: > This is true but I don't see much value in smoothing out the "reattach didn't happen" case, especially if it is achieved by allowing more or less arbitrary code execution within the signal handler. The only reason we are bothering with all of this (`rbd device attach`, `io-timeout`, etc) is to make reattach not just possible but also reliable. It should "just" work and if something goes wrong, it is not that different from an abrupt termination (e.g. if rbd-nbd gets a SIGKILL from k8s because it doesn't react to SIGTERM in time or if rbd-nbd actually crashes due to a bug in librbd or the daemon itself).

1. Once we have the rbd-nbd doing the freeze, the idea is to improve the termination sequence of the rbd nodeplugin pod by adding the preHooks such a way that Kubernetes waits for all the rbd-nbd processes on the pod before doing any ungrateful shutdown, yes I remember some of ceph-csi team conversations and it really make sense!

2. Bug in librbd or daemon is not acceptable and should be resolved once we root cause it :-)

> 
> Is there a reason for treating the "reattach didn't happen" case specially?

Because, that is the worst that can happen in a normal work flow, as far as we can think.

Thanks!

pkalever: @idryomov when io-timeout is set to 0, I do not see any difference that is getting affected.


```
[root@linux-vm1 build]# ./bin/rbd-nbd map rbd-pool/image0 --try-netlink --reattach-timeout 300 --io-timeout 0
/dev/nbd0

[root@linux-vm1 build]# cat /sys/block/nbd0/queue/io_timeout 
30000
```

This is why I treat it to behave the same as default 30 seconds.

idryomov: > * Once we have the rbd-nbd doing the freeze, the idea is to improve the termination sequence of the rbd nodeplugin pod by adding the preHooks such a way that Kubernetes waits for all the rbd-nbd processes on the pod before doing any ungrateful shutdown, yes I remember some of ceph-csi team conversations and it really make sense!

I'm not following.  What happens after all those filesystems get frozen?  Who is going to unfreeze them to clean up user workloads stuck in uninterruptible sleep?

> 
> * Bug in librbd or daemon is not acceptable and should be resolved once we root cause it :-)

But reattach not completing within the 5-minute timeout is?  I guess I'm missing why it seems to be considered "normal work flow".

idryomov: > @idryomov when io-timeout is set to 0, I do not see any difference that is getting affected.
> 
> ```
> [root@linux-vm1 build]# ./bin/rbd-nbd map rbd-pool/image0 --try-netlink --reattach-timeout 300 --io-timeout 0
> /dev/nbd0
> 
> [root@linux-vm1 build]# cat /sys/block/nbd0/queue/io_timeout 
> 30000
> ```
> 
> This is why I treat it to behave the same as default 30 seconds.

Ah, this is not the actual timeout.  This shows how often the timeout handler would fire.  With `io-timeout` set to 0, it should output `Possible stuck request` messages every 30 seconds but never actually time out anything.  Super confusing but that seems to be by design ;)

pkalever: > I'm not following. What happens after all those filesystems get frozen? Who is going to unfreeze them to clean up user workloads stuck in uninterruptible sleep?

In a normal workflow the next `attach` should unfreeze the mount. In case if its too late for `attach` to succeed, then the Ceph-CSI Healer should unfreeze the mount points on a recognized exit code from `rbd-nbd attach` failure.

BTW - Ceph-CSI Healer is responsible to do attaching of the rbd-nbd volumes after a pod restart.

> But reattach not completing within the 5-minute timeout is? I guess I'm missing why it seems to be considered "normal work flow".

I get we shouldn't hit 5 mins timeout most of the times, but we cannot expect that to happen all the time for the practical realworld usecases, and so this will be an extra layer of protection. After all it can be an optional feature at rbd-nbd for generic use case.

> > This is why I treat it to behave the same as default 30 seconds.
> 
> Ah, this is not the actual timeout. This shows how often the timeout handler would fire.

Do you mean, its a bug in kernel? You remember we tested this before with various `--io-timeouts` which were actually getting reflected fine at `/sys/block/nbd0/queue/io_timeout `?

Here: https://github.com/ceph/ceph-csi/issues/2204#issuecomment-894033738

> With `io-timeout` set to 0, it should output `Possible stuck request` messages every 30 seconds but never actually time out anything. Super confusing but that seems to be by design ;)

I will have to test it (very soon), but IIUC you are saying if we set  `--io-timeouts` to `0` the nbd driver will consider it and hence should print `Possible stuck request`, but it is probably it is not just getting reflected at `/sys/block/nbd0/queue/io_timeout `?


Thanks @idryomov.

idryomov: Yes.  `queue/io_timeout` shows the value for the timeout handler.  In the timeout handler, the driver can make a choice to abort the request or to ignore the timeout (perhaps initiating some sort of recovery, etc).  Specifying `--io-timeout 0` tells the nbd driver to _not_ abort the request and instead see if it can be restarted on another socket.

> You remember we tested this before with various --io-timeouts which were actually getting reflected fine at `/sys/block/nbd0/queue/io_timeout` ?

A value other than 0 is reflected correctly.  0 is special and doesn't get reflected because it is a bit of a hack in the nbd driver.

idryomov: > In case if its too late for `attach` to succeed, then the Ceph-CSI Healer should unfreeze the mount points on a recognized exit code from `rbd-nbd attach` failure.

Unfreeze without a daemon attached?  What would handle the I/O from the unfreezed filesystem?

pkalever: > Yes. `queue/io_timeout` shows the value for the timeout handler. In the timeout handler, the driver can make a choice to abort the request or to ignore the timeout (perhaps initiating some sort of recovery, etc). Specifying `--io-timeout 0` tells the nbd driver to _not_ abort the request and instead see if it can be restarted on another socket.

Yes, I see this now:

```
[Thu Aug 19 05:12:33 2021] nbd0: detected capacity change from 0 to 2097152
[Thu Aug 19 05:12:47 2021] EXT4-fs (nbd0): mounted filesystem with ordered data mode. Opts: (null). Quota mode: none.
[Thu Aug 19 05:12:58 2021] block nbd0: Receive control failed (result -32)
[Thu Aug 19 05:12:58 2021] block nbd0: Dead connection, failed to find a fallback
[Thu Aug 19 05:13:28 2021] block nbd0: Possible stuck request 00000000e47aef47: control (read@1073676288,4096B). Runtime 30 seconds
[Thu Aug 19 05:13:58 2021] block nbd0: Possible stuck request 00000000e47aef47: control (read@1073676288,4096B). Runtime 60 seconds
[Thu Aug 19 05:14:22 2021] block nbd0: reconnected socket
[Thu Aug 19 05:14:29 2021] block nbd0: Possible stuck request 00000000e47aef47: control (read@1073676288,4096B). Runtime 90 seconds
[Thu Aug 19 05:16:06 2021] block nbd0: NBD_DISCONNECT
[Thu Aug 19 05:16:06 2021] block nbd0: Disconnected due to user request.
[Thu Aug 19 05:16:06 2021] block nbd0: shutting down sockets
```
The `queue/io_timeout` value was quite misleading here.

> 
> > You remember we tested this before with various --io-timeouts which were actually getting reflected fine at `/sys/block/nbd0/queue/io_timeout` ?
> 
> A value other than 0 is reflected correctly. 0 is special and doesn't get reflected because it is a bit of a hack in the nbd driver.

Not an important one, but I wish to fix it someday :-)

> > In case if its too late for `attach` to succeed, then the Ceph-CSI Healer should unfreeze the mount points on a recognized exit code from `rbd-nbd attach` failure.
> 
> Unfreeze without a daemon attached? What would handle the I/O from the unfreezed filesystem?

This will be the same as the case where we set `--io-timeout=0` and cannot attach the process within the `--reattach-timeout`. I expect in both cases the IO should bailout and it's time to map the image freshly i.e. the application pod needs a restart.

That said, probably the next question would be, what are we going to solve by quiescing?

* You guessed it, we guaranty that the data is not corrupted and rbd-nbd is reliable over any pod restarts. If we do not use quiesce some of the filesystem data/metadata might be stuck in the fs/block/kernel layers, in case if the process attach doesn't happen in time, the missing writes might lead to a corrupted backend. On the other hand, by quiescing, we ensure the data will be pushed from fs/block till backend, hence even if the application pod needs a restart, we will be quite confident the data is in a good shape.

Thanks again!

nixpanic: Looks like setting `io-timeout` to `0` is what we need. We just want the I/O to be queued until the rbd-nbd process reattaches.

pkalever:
/retest ci/centos/upgrade-tests-rbd

---
```
```

internal/rbd/rbd_attach.go
pkalever: @nixpanic Yes, I have updated the PR with the same today.

pkalever: @nixpanic Yes, I have updated the PR with the same today. We also definitely need quiescing for the reason mentioned in the above comment.

nixpanic (dismissed):

nixpanic:
The **lint-extras** GitHub action fails (just like `make containerized-test`), which will be fixed once #2421 is merged and this PR is rebased.

internal/rbd/rbd_attach.go
nixpanic: We would not need `fsfreeze` in case the block-layer does not report errors but just queues the I/O. This should work for both filesystem- and block-mode volumes.

Quiescing takes time (flushing outstanding I/O) and delays restarting the node-plugin Pod. I am not convinced there is a good reason to flushing I/O, as we do expect the pod to restart. Assuming there are issues with restarting, or re-attaching doesn't give me confidence in the solution...

pkalever: Given the scenario where the process is killed(TERM) and brought back to life while the device is held, is a bit special, It will not be a big surprise to arm rbd-nbd with additional safety.

IMHO, at any cost, we shouldn't be in a place that will lead us to corruption. Data loss is secondary to me, which we cannot avoid at the worst.

nixpanic:
@mergifyio rebase

pkalever:
```
[1mSTEP[0m: checking provisioner deployment is running
[1mSTEP[0m: checking nodeplugin deamonset pods are running
Aug 19 11:56:10.916: INFO: Waiting up to 10m0s for all daemonsets in namespace 'cephcsi-e2e-6856a6b083be' to start
Aug 19 11:56:10.919: INFO: 1 / 1 pods ready in namespace 'cephcsi-e2e-6856a6b083be' in daemonset 'csi-cephfsplugin' (0 seconds elapsed)
[1mSTEP[0m: verify PVC and app binding on helm installation
Aug 19 11:56:10.929: INFO: Waiting up to &PersistentVolumeClaim{ObjectMeta:{csi-cephfs-pvc  cephfs-3274    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Spec:PersistentVolumeClaimSpec{AccessModes:[ReadWriteMany],Resources:ResourceRequirements{Limits:ResourceList{},Requests:ResourceList{storage: {{1073741824 0} {<nil>} 1Gi BinarySI},},},VolumeName:,Selector:nil,StorageClassName:*csi-cephfs-sc,VolumeMode:nil,DataSource:nil,DataSourceRef:nil,},Status:PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},},} to be in Bound state
Aug 19 11:56:10.929: INFO: waiting for PVC csi-cephfs-pvc (0 seconds elapsed)
Aug 19 11:56:12.944: INFO: waiting for PVC csi-cephfs-pvc (2 seconds elapsed)
Aug 19 11:56:12.951: INFO: Waiting for PV pvc-8f823be1-4b73-4575-bb13-cf4eb66fad91 to bind to PVC csi-cephfs-pvc
Aug 19 11:56:12.951: INFO: Waiting up to timeout=10m0s for PersistentVolumeClaims [csi-cephfs-pvc] to have phase Bound
Aug 19 11:56:12.954: INFO: PersistentVolumeClaim csi-cephfs-pvc found and phase=Bound (2.92652ms)
Aug 19 11:56:12.954: INFO: Waiting up to 10m0s for PersistentVolume pvc-8f823be1-4b73-4575-bb13-cf4eb66fad91 to have phase Bound
Aug 19 11:56:12.957: INFO: PersistentVolume pvc-8f823be1-4b73-4575-bb13-cf4eb66fad91 found and phase=Bound (3.080234ms)
Aug 19 11:56:12.971: INFO: Waiting up to csi-cephfs-demo-pod to be in Running state
Aug 19 12:01:56.989: INFO: Waiting for pod csi-cephfs-demo-pod to be deleted
Aug 19 12:11:57.001: FAIL: failed to validate CephFS pvc and application binding with error timed out waiting for the condition
```
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.22/runs/94/nodes/101/log/?start=0


```
[1mSTEP[0m: Resize rbd-nbd PVC and check application directory size
E0819 10:57:22.628963   83435 util.go:245] kernel 4.18.0-305.12.1.el8_4.x86_64 does not support required features
[1mSTEP[0m: perform IO on rbd-nbd volume after nodeplugin restart
Aug 19 10:57:22.629: INFO: waiting for kubectl (delete -f [] args %!s(MISSING)) to finish
Aug 19 10:57:22.629: INFO: Running '/usr/bin/kubectl --server=https://192.168.39.112:8443 --kubeconfig=/root/.kube/config --namespace=cephcsi-e2e-7b8169a8 delete -f -'
Aug 19 10:57:22.762: INFO: stderr: "warning: deleting cluster-scoped resources, not scoped to the provided namespace\n"
Aug 19 10:57:22.762: INFO: stdout: "storageclass.storage.k8s.io \"csi-rbd-sc\" deleted\n"
Aug 19 10:57:22.769: INFO: ExecWithOptions {Command:[/bin/sh -c ceph fsid] Namespace:rook-ceph PodName:rook-ceph-tools-7467d8bf8-kp6qz ContainerName:rook-ceph-tools Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Aug 19 10:57:22.769: INFO: >>> kubeConfig: /root/.kube/config
Aug 19 10:57:24.919: INFO: Waiting up to &PersistentVolumeClaim{ObjectMeta:{rbd-pvc  rbd-1318    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Spec:PersistentVolumeClaimSpec{AccessModes:[ReadWriteOnce],Resources:ResourceRequirements{Limits:ResourceList{},Requests:ResourceList{storage: {{1073741824 0} {<nil>} 1Gi BinarySI},},},VolumeName:,Selector:nil,StorageClassName:*csi-rbd-sc,VolumeMode:nil,DataSource:nil,DataSourceRef:nil,},Status:PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},},} to be in Bound state
Aug 19 10:57:24.919: INFO: waiting for PVC rbd-pvc (0 seconds elapsed)
Aug 19 10:57:26.926: INFO: waiting for PVC rbd-pvc (2 seconds elapsed)
Aug 19 10:57:26.933: INFO: Waiting for PV pvc-091a6bbf-7b9e-41c2-86a2-ea7d74b4e95e to bind to PVC rbd-pvc
Aug 19 10:57:26.933: INFO: Waiting up to timeout=10m0s for PersistentVolumeClaims [rbd-pvc] to have phase Bound
Aug 19 10:57:26.935: INFO: PersistentVolumeClaim rbd-pvc found and phase=Bound (2.387417ms)
Aug 19 10:57:26.935: INFO: Waiting up to 10m0s for PersistentVolume pvc-091a6bbf-7b9e-41c2-86a2-ea7d74b4e95e to have phase Bound
Aug 19 10:57:26.937: INFO: PersistentVolume pvc-091a6bbf-7b9e-41c2-86a2-ea7d74b4e95e found and phase=Bound (2.394599ms)
Aug 19 10:57:26.949: INFO: Waiting up to csi-rbd-demo-pod to be in Running state
Aug 19 10:57:46.991: INFO: ExecWithOptions {Command:[/bin/sh -c sync /var/lib/www/html] Namespace:rbd-1318 PodName:csi-rbd-demo-pod ContainerName:web-server Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Aug 19 10:57:46.991: INFO: >>> kubeConfig: /root/.kube/config
Aug 19 10:57:47.184: INFO: stdErr occurred: sync: error syncing '/var/lib/www/html': Input/output error

Aug 19 10:57:47.184: FAIL: failed to sync, err: command terminated with exit code 1, stdErr: sync: error syncing '/var/lib/www/html': Input/output error
```

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e_k8s-1.20/runs/2161/nodes/87/log/?start=0

pkalever:
/retest ci/centos/mini-e2e/k8s-1.20

---

```

```

internal/rbd/rbd_attach.go
pkalever: Yesterday there was a discussion around this in the CSI standup meeting, and here is the summary of the discussion:

1. If we do not quiesce on detach, then there is a chance of data corruption, hence we should provide quiesce.
2. For setups where the writes are slow and hence heavily cached on fs (like a few MBs, GBs), the flushing involved as part of fsfreeze might take time and might lead us to hit `reattach-timeout` (note: this can also be reconfigured from current 300sec to higher). Hence we should have the option to disable quiesce on detach at Ceph-CSI.

The pod termination sequence should be improved with the `preHooks`, the hook should ideally wait for `rbd-nbd` processes.

nixpanic: > 
> 1. If we do not quiesce on detach, then there is a chance of data corruption, hence we should provide quiesce.
> 

Note that for applications there probably is little difference with or without quiescing. The application does not get informed about the need to flush writes, it can still be interrupted (blocked/paused) while it is writing data.

When updating the DaemonSet for the rbd node-plugin, **all** RBD volumes will start to flush their data when quiescing is done. This will cause a load spike on the Ceph cluster and network.

pkalever: Agreed @nixpanic. This should be almost the same load that takes place if we have to snapshot all the PVs in one go!

Madhu-1 (dismissed):

internal/rbd/rbd_attach.go
```diff
@@ -48,10 +48,22 @@ const (
 	rbdUnmapCmdNbdMissingMap  = "rbd-nbd: %s is not mapped"
 	rbdMapConnectionTimeout   = "Connection timed out"
 
-	defaultNbdReAttachTimeout = 300
+	defaultNbdReAttachTimeout = 300 /* in seconds */
+	defaultNbdIOTimeout       = 0   /* do not abort the requests */
 
-	useNbdNetlink  = "try-netlink"
+	// The default way of creating nbd devices via rbd-nbd is though the
```
humblec: through?

internal/rbd/rbd_attach.go
```diff
@@ -48,10 +48,22 @@ const (
...
+	// legacy iotcl interface, to take advantage of netlink features we
```
humblec: ioctl.

internal/rbd/rbd_attach.go
```diff
@@ -48,10 +48,22 @@ const (
...
+	// should specify `try-netlink` flag explicitly.
+	useNbdNetlink = "try-netlink"
+
+	// `reattach-timeout` of rbd-nbd is to tweaks NBD_ATTR_DEAD_CONN_TIMEOUT.
```
humblec: tweak

internal/rbd/rbd_attach.go
```diff
@@ -48,10 +48,22 @@ const (
...
+	// It specifies how long the device should be hold waiting for the
```
humblec: held?

internal/rbd/rbd_attach.go
```diff
@@ -48,10 +48,22 @@ const (
...
+	// userspace process to come back to life.
 	setNbdReattach = "reattach-timeout"
+
+	// `io-timeout` of rbd-nbd is to tweak NBD_ATTR_TIMEOUT. It specifies
+	// how long the IO should wait to get handled before bailing out.
```
humblec, line 65: it would be better if we can also add some reference to the mentoned timeout attributes from ceph core doc urls.

pkalever: There are no reference docs AFAIK.

pkalever:
@humblec fixed the typos now. Thanks, Humble.

humblec:
> @humblec fixed the typos now. Thanks, Humble.

Thanks @pkalever , code looks good to me :+1: .   Hope @idryomov is fine with the updated values of these timeouts too.

trociny (commented):
LGTM

pkalever:

---
```
```

Madhu-1 (dismissed):

humblec (dismissed):

pkalever:
@nixpanic @Madhu-1 @humblec did anything change recently:


https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.20/runs/2165/nodes/97/log/?start=0

```
[1mSTEP[0m: create a PVC and bind it to an app using rbd-nbd mounter with encryption
Aug 23 09:44:20.406: INFO: waiting for kubectl (delete -f [] args %!s(MISSING)) to finish
Aug 23 09:44:20.406: INFO: Running '/usr/bin/kubectl --server=https://192.168.39.116:8443 --kubeconfig=/root/.kube/config --namespace=cephcsi-e2e-fb8312b32bd4 delete -f -'
Aug 23 09:44:20.540: INFO: stderr: "warning: deleting cluster-scoped resources, not scoped to the provided namespace\n"
Aug 23 09:44:20.540: INFO: stdout: "storageclass.storage.k8s.io \"csi-rbd-sc\" deleted\n"
Aug 23 09:44:20.547: INFO: ExecWithOptions {Command:[/bin/sh -c ceph fsid] Namespace:rook-ceph PodName:rook-ceph-tools-7467d8bf8-dfzkx ContainerName:rook-ceph-tools Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:true Quiet:false}
Aug 23 09:44:20.547: INFO: >>> kubeConfig: /root/.kube/config
Aug 23 09:44:22.739: INFO: Waiting up to &PersistentVolumeClaim{ObjectMeta:{rbd-pvc  rbd-8081    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Spec:PersistentVolumeClaimSpec{AccessModes:[ReadWriteOnce],Resources:ResourceRequirements{Limits:ResourceList{},Requests:ResourceList{storage: {{1073741824 0} {<nil>} 1Gi BinarySI},},},VolumeName:,Selector:nil,StorageClassName:*csi-rbd-sc,VolumeMode:nil,DataSource:nil,DataSourceRef:nil,},Status:PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},},} to be in Bound state
Aug 23 09:44:22.740: INFO: waiting for PVC rbd-pvc (0 seconds elapsed)
Aug 23 09:44:24.762: INFO: waiting for PVC rbd-pvc (2 seconds elapsed)
Aug 23 09:44:24.769: INFO: Waiting for PV pvc-b0af166e-ac0a-4efd-97ed-24eb3b780a17 to bind to PVC rbd-pvc
Aug 23 09:44:24.769: INFO: Waiting up to timeout=10m0s for PersistentVolumeClaims [rbd-pvc] to have phase Bound
Aug 23 09:44:24.772: INFO: PersistentVolumeClaim rbd-pvc found and phase=Bound (2.485568ms)
Aug 23 09:44:24.772: INFO: Waiting up to 10m0s for PersistentVolume pvc-b0af166e-ac0a-4efd-97ed-24eb3b780a17 to have phase Bound
Aug 23 09:44:24.774: INFO: PersistentVolume pvc-b0af166e-ac0a-4efd-97ed-24eb3b780a17 found and phase=Bound (2.787388ms)
Aug 23 09:44:24.788: INFO: Waiting up to csi-rbd-demo-pod to be in Running state
Aug 23 09:54:24.801: FAIL: failed to validate encrypted pvc with error timed out waiting for the condition
```


I remember this succeeded in the previous run.

nixpanic:
I do not remember seeing such a problem earlier. Relevant snippet from the logs:
```
I0823 08:44:32.513339   37638 utils.go:176] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a GRPC call: /csi.v1.Node/NodeStageVolume
I0823 08:44:32.513835   37638 utils.go:180] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a GRPC request: {"secrets":"***stripped***","staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-b0af166e-ac0a-4efd-97ed-24eb3b780a17/globalmount","volume_capability":{"AccessType":{"Mount":{"fs_type":"ext4","mount_flags":["discard"]}},"access_mode":{"mode":1}},"volume_context":{"clusterID":"f2414235-9b8f-4c48-83c0-ada58316ad77","csi.storage.k8s.io/pv/name":"pvc-b0af166e-ac0a-4efd-97ed-24eb3b780a17","csi.storage.k8s.io/pvc/name":"rbd-pvc","csi.storage.k8s.io/pvc/namespace":"rbd-8081","encrypted":"true","imageFeatures":"layering","imageName":"csi-vol-50665b36-03ee-11ec-b76a-4ae391743d3a","journalPool":"replicapool","mapOptions":"debug-rbd=20","mounter":"rbd-nbd","pool":"replicapool","storage.kubernetes.io/csiProvisionerIdentity":"1629707901029-8081-rbd.csi.ceph.com","thickProvision":"false"},"volume_id":"0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a"}
I0823 08:44:32.514077   37638 rbd_util.go:1202] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a setting disableInUseChecks: false image features: [layering] mounter: rbd-nbd
I0823 08:44:32.515457   37638 omap.go:86] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a got omap values: (pool="replicapool", namespace="", name="csi.volume.50665b36-03ee-11ec-b76a-4ae391743d3a"): map[csi.imageid:116ae54a3729 csi.imagename:csi-vol-50665b36-03ee-11ec-b76a-4ae391743d3a csi.volname:pvc-b0af166e-ac0a-4efd-97ed-24eb3b780a17 csi.volume.encryptKMS:default csi.volume.owner:rbd-8081]
E0823 08:44:32.515674   37638 util.go:245] kernel 4.19.182 does not support required features
I0823 08:44:33.119147   37638 cephcmds.go:60] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a command succeeded: rbd [device list --format=json --device-type nbd]
I0823 08:44:33.233619   37638 rbd_attach.go:298] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a rbd: map mon rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789
I0823 08:44:34.034478   37638 cephcmds.go:60] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a command succeeded: rbd [--id cephcsi-rbd-node -m rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789 --keyfile=***stripped*** map replicapool/csi-vol-50665b36-03ee-11ec-b76a-4ae391743d3a --device-type nbd --options try-netlink --options reattach-timeout=300 --options io-timeout=0 --options debug-rbd=20]
I0823 08:44:34.034720   37638 nodeserver.go:381] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a rbd image: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a/replicapool was successfully mapped at /dev/nbd0
I0823 08:44:34.056078   37638 encryption.go:78] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a image replicapool/csi-vol-50665b36-03ee-11ec-b76a-4ae391743d3a encrypted state metadata reports "encryptionPrepared"
I0823 08:44:34.057202   37638 mount_linux.go:446] Attempting to determine if disk "/dev/nbd0" is formatted using blkid with args: ([-p -s TYPE -s PTTYPE -o export /dev/nbd0])
I0823 08:44:34.061575   37638 mount_linux.go:449] Output: ""
I0823 08:44:34.061617   37638 crypto.go:265] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a Encrypting device /dev/nbd0 with LUKS
E0823 08:44:35.222328   37638 encryption.go:196] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a failed to encrypt volume replicapool/csi-vol-50665b36-03ee-11ec-b76a-4ae391743d3a: failed to encrypt device /dev/nbd0 with LUKS: an error (exit status 1) occurred while running cryptsetup args: [-q luksFormat --type luks2 --hash sha256 /dev/nbd0 -d /dev/stdin]
I0823 08:44:35.816624   37638 cephcmds.go:60] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a command succeeded: rbd [unmap /dev/nbd0 --device-type nbd --options try-netlink --options reattach-timeout=300 --options io-timeout=0]
E0823 08:44:35.816967   37638 utils.go:185] ID: 8 Req-ID: 0001-0024-f2414235-9b8f-4c48-83c0-ada58316ad77-0000000000000001-50665b36-03ee-11ec-b76a-4ae391743d3a GRPC error: rpc error: code = Internal desc = failed to encrypt rbd image replicapool/csi-vol-50665b36-03ee-11ec-b76a-4ae391743d3a: failed to encrypt volume replicapool/csi-vol-50665b36-03ee-11ec-b76a-4ae391743d3a: failed to encrypt device /dev/nbd0 with LUKS: an error (exit status 1) occurred while running cryptsetup args: [-q luksFormat --type luks2 --hash sha256 /dev/nbd0 -d /dev/stdin]
```

nixpanic:
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.20/runs/2165/nodes/106/log/?start=0

While using `pvc-b0af166e-ac0a-4efd-97ed-24eb3b780a17`:
```
Aug 23 08:44:35 minikube kernel: block nbd0: Connection timed out
Aug 23 08:44:35 minikube kernel: block nbd0: shutting down sockets
Aug 23 08:44:35 minikube kernel: print_req_error: 2 callbacks suppressed
Aug 23 08:44:35 minikube kernel: print_req_error: I/O error, dev nbd0, sector 20736
Aug 23 08:44:35 minikube kernel: block nbd0: Connection timed out
Aug 23 08:44:35 minikube kernel: print_req_error: I/O error, dev nbd0, sector 20992
Aug 23 08:44:35 minikube kernel: block nbd0: Connection timed out
Aug 23 08:44:35 minikube kernel: print_req_error: I/O error, dev nbd0, sector 21248
Aug 23 08:44:35 minikube kernel: block nbd0: Connection timed out
Aug 23 08:44:35 minikube kernel: print_req_error: I/O error, dev nbd0, sector 21504
Aug 23 08:44:35 minikube kernel: block nbd0: Connection timed out
Aug 23 08:44:35 minikube kernel: print_req_error: I/O error, dev nbd0, sector 21760
Aug 23 08:44:35 minikube kernel: block nbd0: Connection timed out
Aug 23 08:44:35 minikube kernel: print_req_error: I/O error, dev nbd0, sector 22016
Aug 23 08:44:35 minikube kernel: block nbd0: Connection timed out
Aug 23 08:44:35 minikube kernel: print_req_error: I/O error, dev nbd0, sector 22272
Aug 23 08:44:35 minikube kernel: block nbd0: NBD_DISCONNECT
Aug 23 08:44:35 minikube kernel: block nbd0: Send disconnect failed -32
```

It seems that `rbd-nbd` hit some kind of timeout?

idryomov:
Kernel 4.19.182 doesn't handle `io-timeout=0` correctly because it lacks https://github.com/torvalds/linux/commit/2da22da573481cc4837e246d0eee4d518b3f715e.  In general, the nbd driver in older kernels is quite broken so you may need to constrain nbd tests to e.g. CentOS 8.4.

pkalever:
> Kernel 4.19.182 doesn't handle `io-timeout=0` correctly because it lacks [torvalds/linux@2da22da](https://github.com/torvalds/linux/commit/2da22da573481cc4837e246d0eee4d518b3f715e). In general, the nbd driver in older kernels is quite broken so you may need to constrain nbd tests to e.g. CentOS 8.4.

Thanks Ilya!

Another option could be,
As we know, the e2e run on minikube which doesn't have v5.4 (release holding the patch)
Given nbd module is a dynamic insertable module in minikube kernels, maybe we can build a custom one with this patch on top of 4.19.182 and insert it as part of the test setup for nbd tests? Or even build a custom minikube iso with this patch?

idryomov:
> Hope @idryomov is fine with the updated values of these timeouts too.

Yup, I'm fine with `io-timeout=0` -- it was me who suggested it after all ;)

However, I still don't see the need for quiesce on detach.  In particular, I don't think this statement is true:

> If we do not quiesce on detach, then there is a chance of data corruption, hence we should provide quiesce.

It isn't directly related to this PR though -- just so happened that the discussion for https://github.com/ceph/ceph/pull/42609 is happening here.

idryomov:
> As we know, the e2e run on minikube which doesn't have v5.4 (release holding the patch)
> Given nbd module is a dynamic insertable module in minikube kernels, maybe we can build a custom one with this patch on top of 4.19.182 and insert it as part of the test setup for nbd tests? Or even build a custom minikube iso with this patch?

It isn't just this patch.  If you going to treat nbd job failures seriously, my advice would be to get a minikube image with as recent kernel as possible.  Case in point: in upstream rbd CI we are seeing some failures with 5.4 LTS kernels but not with CentOS 8.4.

pkalever:
> > Hope @idryomov is fine with the updated values of these timeouts too.
> 
> Yup, I'm fine with `io-timeout=0` -- it was me who suggested it after all ;)

My fault, I should have shared the credits with you in the commit msg! I will do it in the next respin. Sorry!

> 
> However, I still don't see the need for quiesce on detach. In particular, I don't think this statement is true:

Maybe I didn't explain it well all this time. Let me put that as a question and see:

I want to solve a case where the `attach` doesn't happen in `reattach-timeout`, so that even if the application pod has to restart, it will have a corruption-free backend image.

The reasons for not reattaching in the specified timeout could be anything beyond our control.

* Upgrade failures (cannot download image)
* Container crash could be a bug in ceph-csi or ...
* Network issue downloading the image
* Resource highly used on the node
* docker/podman bug?

We can go on with practical reasons.

Do we have any better solution or suggestions to handle this case other than ceph/ceph#42609 (quiesce on detach) ?


> 
> > If we do not quiesce on detach, then there is a chance of data corruption, hence we should provide quiesce.
> 
> It isn't directly related to this PR though -- just so happened that the discussion for [ceph/ceph#42609](https://github.com/ceph/ceph/pull/42609) is happening here.

Agree!


Thanks :-)

e2e/rbd.go
```diff
@@ -232,6 +234,15 @@ var _ = Describe("RBD", func() {
 			e2elog.Failf("failed to create node secret with error %v", err)
 		}
 		deployVault(f.ClientSet, deployTimeout)
+
+		kernelRelease, err := util.GetKernelVersion()
```
nixpanic: You'll get the kernel version where the e2e (client) executable runs, not the kernel version that is used by the csi-rbdplugin pod.

There are different ways you can get the kernel version of the node(s) where the containers are running:

a. exec into a pod and run `uname`
b. use the k8s API to list the nodes and inspect [the details](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#nodesysteminfo-v1-core)

nixpanic (changes requested):

e2e/rbd.go
pkalever: Hmm, good catch Niels, then what we are doing here: https://github.com/ceph/ceph-csi/blob/devel/e2e/rbd.go#L442
might be also wrong then! I will fix this at both places.

Thanks, @nixpanic.

pkalever: Done now!

e2e/rbd.go
```diff
@@ -57,6 +57,9 @@ var (
 	appBlockSmartClonePath = rbdExamplePath + "block-pod-clone.yaml"
 	snapshotPath           = rbdExamplePath + "snapshot.yaml"
 	defaultCloneCount      = 10
+
+	kernelRelease = ""
+	nbdMapOptions = "debug-rbd=20"
```
humblec, line 61: would be good to define a constant here? no hard opinion though.

e2e/rbd_helper.go
```diff
@@ -952,3 +966,22 @@ func waitToRemoveImagesFromTrash(f *framework.Framework, poolName string, t int)
 
 	return err
 }
+
+// getKernelVersionFromConatiner gets the kernel version from the specified container.
```
humblec: container

e2e/rbd.go
pkalever: Could you elaborate I doubt I got you?

e2e/rbd_helper.go
pkalever: I didn't get this too?

Madhu-1: typo in getKernelVersionFromConatiner

pkalever: Ah! Thanks Madhu and Humble

pkalever: Done now.

e2e/rbd.go
pkalever: constant for debug level?

pkalever: If yes, Nah! this is good enough.

e2e/rbd_helper.go
```diff
@@ -952,3 +966,22 @@ func waitToRemoveImagesFromTrash(f *framework.Framework, poolName string, t int)
 
 	return err
 }
+
+// getKernelVersionFromContainer gets the kernel version from the specified container.
+func getKernelVersionFromContainer(f *framework.Framework, ns, dsn, cn string) (string, error) {
```
Madhu-1: it should have been`getKernelVersionFromRbdplugin`.

Madhu-1 (dismissed):

e2e/rbd_helper.go
pkalever: Ah! I should have moved it to utils?

pkalever: or rather into e2e/pod.go?

pkalever: Done now, thanks @Madhu-1 
Moved it into e2e/pod.go

e2e/pod.go
```diff
@@ -406,3 +406,22 @@ func calculateSHA512sum(f *framework.Framework, app *v1.Pod, filePath string, op
 
 	return checkSum, nil
 }
+
+// getKernelVersionFromContainer gets the kernel version from the specified container.
+func getKernelVersionFromContainer(f *framework.Framework, ns, dsn, cn string) (string, error) {
```
Madhu-1: sorry, I did not convey what I meant. This function is not generic (pod, deployment, etc)  it is specific to the daemonset pod the function name should be `getKernelVersionFromDaemonset`.

Not e:-if we have kubernetes clusters with multiple nodes (with different kernel versions) things might not work as expected. 

This can be enhanced as follows (maybe as a follow-up patch)

* Check all the nodes kernel versions
* If any of the nodes is having expected kernel version
* schedule the application pod to the specific node

e2e/rbd.go
```diff
@@ -57,6 +57,9 @@ var (
 	appBlockSmartClonePath = rbdExamplePath + "block-pod-clone.yaml"
 	snapshotPath           = rbdExamplePath + "snapshot.yaml"
 	defaultCloneCount      = 10
+
+	kernelRelease = ""
```
nixpanic: kernel release does not need to  be a global variable, I think?

e2e/rbd.go
```diff
@@ -232,6 +235,15 @@ var _ = Describe("RBD", func() {
 			e2elog.Failf("failed to create node secret with error %v", err)
 		}
 		deployVault(f.ClientSet, deployTimeout)
+
+		kernelRelease, err = util.GetKernelVersion()
```
nixpanic: don't introduce a mistake, and correct it in the next commit, consider adding the helper function in one commit, and using it here instead (or combine the two in a single commit)

e2e/rbd_helper.go
```diff
@@ -40,6 +40,20 @@ var nbdResizeSupport = []util.KernelVersion{
...
+// nolint:gomnd // numbers specify Kernel versions.
+var nbdZeroIOtimeoutSupport = []util.KernelVersion{
+	{
+		Version:      5,
+		PatchLevel:   4,
+		SubLevel:     0,
+		ExtraVersion: 0,
+		Distribution: "",
+		Backport:     false,
+	}, // standard 5.4+ versions
```
nixpanic, line 54: Didn't you mention that the CentOS 8.4 kernel was also good? That can probably be added as a backport kernel, and running the e2e tests on OpenShift with current RHEL CoreOS should then work with the right timeouts.

e2e/rbd.go
pkalever: We run all the rbd tests in one go right?
What is the problem with reusing the kernel release?

e2e/rbd.go
pkalever: I see it this way:
Oh! The mistake was not introduced by this PR, it was part of the resize nbd device test already, I was trying to reorder things to reuse them.

e2e/rbd_helper.go
pkalever: Yes, we could, Do you know what kernel version does Centos 8.4 run?

e2e/pod.go
pkalever: Got it, I can rename it to getKernelVersionFromDaemonset.

Yeah, right! we can add such intelligence, for now, let's switch the timeout based on the kernel versions.
In the future, we can improve it (I will raise an issue to track your suggestions)

Thanks Madhu.

pkalever: Here: https://github.com/ceph/ceph-csi/issues/2432

e2e/rbd_helper.go
pkalever: From here: https://en.wikipedia.org/wiki/CentOS
I see it as `4.18.0-305`

Will add the same, thanks Niels.

e2e/rbd.go
pkalever: I think I got you!

e2e/rbd.go
pkalever: I have reordered the things now.

pkalever:
@Madhu-1 @nixpanic hope I addressed everying. PTAL. Thanks!

nixpanic (approved):

Madhu-1 (approved):

idryomov:
Nit: change the title of the PR to "rbd: fix rbd-nbd io-timeout to never abort".

pkalever:
> Nit: change the title of the PR to "rbd: fix rbd-nbd io-timeout to never abort".

Thanks @idryomov done now :-)

idryomov:
> I want to solve a case where the `attach` doesn't happen in `reattach-timeout`, so that even if the application pod has to restart, it will have a corruption-free backend image.

I assume by backend image you mean filesystem image?  The filesystem image should always be corruption-free because filesystems are built to withstand sudden power loss or kernel panics.  Unless you foresee reattach not completing in time continuously and on a regular basis, there is nothing wrong with occasionally losing in-flight I/O and going through journal recovery on the next mount.

> 
> The reasons for not reattaching in the specified timeout could be anything beyond our control.
> 
>     * Upgrade failures (cannot download image)

Does the download of the new image start after detach (i.e. "hanging" in-flight I/O)?  That seems wrong even if everything in between (container registry, network, etc) were 100% reliable.

> 
>     * Container crash could be a bug in ceph-csi or ...

How is it different from a crash with no reattach in sight?

> 
>     * Network issue downloading the image

Already mentioned above.

> 
>     * Resource highly used on the node

If the node is so oversubscribed that rbd-nbd process fails to start, the user workload is already impacted.

> 
>     * docker/podman bug?

Again, how is it different from a "regular" crash?

> 
> 
> We can go on with practical reasons.
> 
> Do we have any better solution or suggestions to handle this case other than [ceph/ceph#42609](https://github.com/ceph/ceph/pull/42609) (quiesce on detach) ?

I guess I just don't see why this case needs handling.  I'm working off of the assumption that reattach failures would be (or at least are expected to be) as rare as rbd-nbd crashes or kernel panics.  Even if the download of the new image happens after detach (instead of before as it should), how often do you expect the upgrade to fail?

PR #2396, opened 2021-08-11
cleanup: run codespell on containerized testing

### LINKED ISSUE #2229: Run `codespell` on local contiainerized testing

`codespell` is run as a GitHub Action on PRs. This test is not included when developers run `make containerized-test` before creating a new PR.

Developers would not get surprised by simple CI failures that are expected to have been run locally before creating a PR or updating one.

`make containerized-test` should run `codespell` in a similar way as the GitHub Action does.

# Additional context #

The GitHub Action is in [`.github/workflows/codespell.yaml`](https://github.com/ceph/ceph-csi/blob/devel/.github/workflows/codespell.yaml).

yati1998:
This commit adds a new target codespell to the
make containerized-test.

Fixes: #2229

### DIFF
--- a/.github/workflows/codespell.yaml
+++ b/.github/workflows/codespell.yaml
@@ -14,9 +14,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: codespell
-        uses: codespell-project/actions-codespell@master
-        with:
-          skip: .git,./vendor,./docs/design/proposals/images
-          check_filenames: true
-          ignore_words_list: ExtraVersion,extraversion,ba
-          check_hidden: true
+        run: CONTAINER_CMD=docker make containerized-test TARGET=codespell
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ endif
 
 all: cephcsi
 
-.PHONY: go-test static-check mod-check go-lint lint-extras gosec commitlint
+.PHONY: go-test static-check mod-check go-lint lint-extras gosec commitlint codespell
 ifeq ($(CONTAINERIZED),no)
 # include mod-check in non-containerized runs
 test: go-test static-check mod-check
@@ -95,7 +95,7 @@ else
 # exclude mod-check for containerized runs (CI runs it separately)
 test: go-test static-check
 endif
-static-check: check-env go-lint lint-extras gosec
+static-check: check-env codespell go-lint lint-extras gosec
 
 go-test: TEST_COVERAGE ?= $(shell . $(CURDIR)/build.env ; echo $${TEST_COVERAGE})
 go-test: GO_COVER_DIR ?= $(shell . $(CURDIR)/build.env ; echo $${GO_COVER_DIR})
@@ -139,6 +139,8 @@ func-test:
 check-env:
 	@./scripts/check-env.sh
 
+codespell:
+	codespell --config scripts/codespell.conf
 #
 # commitlint will do a rebase on top of GIT_SINCE when REBASE=1 is passed.
 #
--- a/scripts/Dockerfile.test
+++ b/scripts/Dockerfile.test
@@ -29,6 +29,7 @@ RUN source /build.env \
 	librbd-devel \
 	rubygems \
 	ShellCheck \
+      codespell \
 	yamllint \
 	npm \
 	diffutils \
--- a/scripts/codespell.conf
+++ b/scripts/codespell.conf
@@ -1,0 +1,4 @@
+[codespell]
+skip = .git,./vendor,./docs/design/proposals/images
+ignore-words-list = ExtraVersion,extraversion,ba
+check-filenames = true


scripts/Dockerfile.test
```diff
@@ -29,6 +29,7 @@ RUN source /build.env \
 	librbd-devel \
 	rubygems \
 	ShellCheck \
+   codespell \
```
Yuggupta27: correct indent

Yuggupta27 (changes requested):

scripts/Dockerfile.test
yati1998: Dont know, but it's fine in my local file

yati1998:
@nixpanic do review. JFYI, codespell doesn't have --config option. Hence, added the required configuration in the command itself.

Yuggupta27 (changes requested):
I might be missing something but I don't see codespell running in the job.. Can you please point me to the job where it is running? (I see it getting installed, but don't see it running)

Makefile
```diff
@@ -87,7 +87,7 @@ endif
 
 all: cephcsi
 
-.PHONY: go-test static-check mod-check go-lint lint-extras gosec commitlint
+.PHONY: go-test static-check mod-check go-lint lint-extras gosec commitlint codespell
```
nixpanic, line 90: It probably would be good to add `codespell` to `static-check` too, so that `make containerized-test` includes it automatically. With the current state, one needs to run `make containerized-test TARGET=codespell`.

nixpanic (changes requested):

nixpanic:
oh, and to run it, update `.github/workflows/codespell.yaml` similar to (most of) the other workflows that run `make containerized-test TARGET=...`

Makefile
Yuggupta27: True, and we have disabled `make containerized-test ` in the ci so it would never run for us

.github/workflows/codespell.yaml
```diff
@@ -20,3 +20,4 @@ jobs:
           check_filenames: true
           ignore_words_list: ExtraVersion,extraversion,ba
           check_hidden: true
+        run: CONTAINER_CMD=docker make containerized-test TARGET=codespell
```
nixpanic, line 17: I do not think this is correct. You probably should just copy one of the other workflows and replace `TARGET=...` with `TARGET=codespell`. There should be no need to have the codespell options listed in this file anymore, as those are now set in the `Makefile`.

nixpanic (changes requested):
I think there is a problem with the workflow change, `codespell` is not listed in https://github.com/ceph/ceph-csi/pull/2396/checks anymore, and it should run inside the container now....

Makefile
```diff
@@ -139,6 +139,8 @@ func-test:
 check-env:
 	@./scripts/check-env.sh
 
+codespell:
+	codespell --skip=.git,./vendor,./docs/design/proposals/images -L ExtraVersion,extraversion,ba
```
Madhu-1: Move all these configurations to a configfile https://github.com/codespell-project/codespell#using-a-config-file?

.github/workflows/codespell.yaml
Madhu-1: Yes this yaml file show remain same as other yaml file in the workflows directory with only Target and name change.

Madhu-1:
> I think there is a problem with the workflow change, `codespell` is not listed in https://github.com/ceph/ceph-csi/pull/2396/checks anymore, and it should run inside the container now....

This is because of the error
 `
a step cannot have both the `uses` and `run` keys
`

Makefile
yati1998: Yes I have checked that, but running codespell with config flag shows error: `unrecognized arguments: --config=`

.github/workflows/codespell.yaml
yati1998: @nixpanic @Madhu-1 still there is an issue removing uses section, as the flags check_filename and check_hidden are not being mentioned in the command running in Makefile. Not sure if it's important to have that?

nixpanic: I don't think that is much of a problem. For me, it is more important that the checks in CI can be run locally with the same configuration.

yati1998: I have updated the PR. Please do review it.

Makefile
nixpanic: I guess the argument parsing of codespell isn't that great. There seems to be a `--config` option, but without the `=`:
```
$ codespell -h | grep config
                 [-A LINES] [-B LINES] [-C LINES] [--config CONFIG]
  --config CONFIG       path to config file.
```

You might be able to use `--config scripts/codespell.conf`

yati1998: oh, okay. Thanks.

yati1998: I am not sure which version you are using @nixpanic  but for me it still fails.
`[yatipadia@192 ceph-csi]$ codespell -h | grep config
[yatipadia@192 ceph-csi]$ 
`

Madhu-1:
```
[🎩︎]mrajanna@localhost volume-replication-operator $]codespell -h |grep config
                 [-A LINES] [-B LINES] [-C LINES] [--config CONFIG]
  --config CONFIG       path to config file.
[🎩︎]mrajanna@localhost volume-replication-operator $]codespell --version
2.0.0
```
latest is 2.1.0 https://github.com/codespell-project/codespell/releases/tag/v2.1.0

nixpanic: we verified things a little more, and the next update to this PR will use the `--config` option :smiley:

yati1998: Done now.

nixpanic (dismissed):

scripts/codespell.conf
```diff
@@ -0,0 +1,5 @@
+ 
+[codespell]
+skip = .git,./vendor,./docs/design/proposals/images
+ignore-words-list = ExtraVersion,extraversion,ba
+check-filenames = true
```
Madhu-1: @yati1998 GitHub is giving a warning about `no newline at end of the file`

scripts/codespell.conf
```diff
@@ -0,0 +1,5 @@
+
```
Madhu-1: New line is required?

Madhu-1 (commented):
No major comments. Left few questions

scripts/codespell.conf
yati1998: No it's not required. I will remove it

yati1998: No, will remove it

Madhu-1 (approved):

nixpanic (approved):

PR #2404, opened 2021-08-12
RBD Async DR: Handle DisableReplication/DeleteVolume for mirrored secondary image

Madhu-1:
This PR does the follow

* Return succuss for DisableVolumeReplication if the image is healthy secondary

If the image is in a secondary state and it's up+replaying means it's a healthy secondary and the image is primary somewhere in the remote cluster and the local image is getting replayed.  Return success for the Disabling mirroring as we cannot disable the mirroring on the secondary state, when the image on the remote site gets disabled the image on all the remote (secondary)
will get auto-deleted. This helps in garbage collecting the volume replication Kubernetes artifacts

* Cleanup OMAP data During DeleteVolume for secondary image
    
If the image is in a secondary state and it's up+replaying means it's a healthy secondary and the image is primary somewhere in the remote cluster and the local image is getting replayed. Delete the OMAP data generated as we cannot delete the secondary image. When the image on the primary cluster gets deleted/mirroring disabled, the image on all the remote (secondary) clusters will get auto-deleted. This helps in garbage collecting the OMAP, PVC, and PV objects after failback operation.

More details about this at https://github.com/csi-addons/volume-replication-operator/issues/100

TODO:

- [x] E2E testing

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -831,6 +831,51 @@ func (cs *ControllerServer) DeleteVolume(
 	}
 	defer cs.VolumeLocks.Release(rbdVol.RequestName)
 
+	return cleanupRBDImage(ctx, rbdVol, cr)
+}
+
+// cleanupRBDImage removes the rbd image and OMAP metadata associated with it.
+func cleanupRBDImage(ctx context.Context,
+	rbdVol *rbdVolume, cr *util.Credentials) (*csi.DeleteVolumeResponse, error) {
+	mirroringInfo, err := rbdVol.getImageMirroringInfo()
+	if err != nil {
+		util.ErrorLog(ctx, err.Error())
+
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	// Cleanup only omap data if the following condition is met
+	// Mirroring is enabled on the image
+	// Local image is secondary
+	// Local image is in up+replaying state
+	if mirroringInfo.State == librbd.MirrorImageEnabled && !mirroringInfo.Primary {
+		// If the image is in a secondary state and its up+replaying means its
+		// an healthy secondary and the image is primary somewhere in the
+		// remote cluster and the local image is getting replayed. Delete the
+		// OMAP data generated as we cannot delete the secondary image. When
+		// the image on the primary cluster gets deleted/mirroring disabled,
+		// the image on all the remote (secondary) clusters will get
+		// auto-deleted. This helps in garbage collecting the OMAP, PVC and PV
+		// objects after failback operation.
+		localStatus, rErr := rbdVol.getLocalState()
+		if rErr != nil {
+			return nil, status.Error(codes.Internal, rErr.Error())
+		}
+		if localStatus.Up && localStatus.State == librbd.MirrorImageStatusStateReplaying {
+			if err = undoVolReservation(ctx, rbdVol, cr); err != nil {
+				util.ErrorLog(ctx, "failed to remove reservation for volume (%s) with backing image (%s) (%s)",
+					rbdVol.RequestName, rbdVol.RbdImageName, err)
+
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+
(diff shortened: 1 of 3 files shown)


internal/rbd/controllerserver.go
```diff
@@ -843,6 +849,45 @@ func (cs *ControllerServer) DeleteVolume(
...
+				util.ErrorLog(ctx, "failed to remove reservation for volume (%s) with backing image (%s) (%s)",
+					rbdVol.RequestName, rbdVol.RbdImageName, err)
+
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+
+			return &csi.DeleteVolumeResponse{}, nil
+		}
+		util.ErrorLog(ctx,
+			"secondary image status is up=%t and state=%s",
```
Yuggupta27, line 874: `secondary image status is up=%t` -> `secondary image status=%t` ?

Madhu-1: i thought as its a bool value the status=true or false does not make sense so add up state for more readibility

Yuggupta27: Sure, This seems better :+1:

internal/rbd/controllerserver.go
```diff
@@ -831,6 +831,51 @@ func (cs *ControllerServer) DeleteVolume(
 	}
 	defer cs.VolumeLocks.Release(rbdVol.RequestName)
 
+	return cleanupRBDImage(ctx, rbdVol, cr)
+}
+
+// cleanupRBDImage removes the rbd image and OMAP metadata associated with it.
+func cleanupRBDImage(ctx context.Context,
+	rbdVol *rbdVolume, cr *util.Credentials) (*csi.DeleteVolumeResponse, error) {
+	mirroringInfo, err := rbdVol.getImageMirroringInfo()
```
ShyamsundarR, line 840: (question) So this would return no error when image mirroring is not enabled? Got till the go-ceph call, but wanted to validate nevertheless. And if image mirroring is not enabled the mirroringInfo  struct will have the right values.

ShyamsundarR (approved):

internal/rbd/controllerserver.go
Madhu-1: Yes, correct. it won't return an error message, the struct will be filled with the right values.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -321,31 +321,56 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
...
-		// the image state can be still disabling once we disable the mirroring
-		// check the mirroring is disabled or not
-		mirroringInfo, err = rbdVol.getImageMirroringInfo()
-		if err != nil {
-			util.ErrorLog(ctx, err.Error())
+	return &replication.DisableVolumeReplicationResponse{}, nil
+}
 
-			return nil, status.Error(codes.Internal, err.Error())
+func disableVolumeReplication(rbdVol *rbdVolume,
```
nixpanic, line 333: Is there a reason why this is not a function of `rbdVolume`? You could have the `mirroringInfo` argument as part of the `rbdVolume` struct, and drop the parameter here.

Madhu-1: Made `getLocalState` as the methods of the `rbdVolume` but didn't made `disableVolumeReplication`  a method of `rbdVolume` as this function returns the appropriate GRPC error codes and IMO the core function which is the methods of rbdVolume should not deal with GRPC error codes directly. and as this is an extension to core rbd functions didn't add `mirroringInfo` as the member of the `rbdVolume` structure.

Am planning to rework core files with interfaces and breaking major functions into smaller helper functions to write more unit testings. I will work on creating a new structure that embeds the `rbdImage` core structure and moves these helper functions as an interface which will help us to write more unit testing.

nixpanic (approved):

Madhu-1:
@mergifyio rebase

PR #2393, opened 2021-08-10
e2e: improve the debug logs for rbd-nbd

pkalever:
# Describe what this PR does #
 
Ceph’s logging levels operate on a scale of 1 to 20, where 1 is terse and 20 is verbose.
 
Format:
debug-{subsystem} = {log-level}
 
Setting `rbd` loglevel to 20 at our rbd-nbd e2e tests.
 
Also see:

- https://github.com/ceph/ceph-csi/pull/2406
- https://github.com/ceph/ceph-csi/pull/2407

 
 
Signed-off-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)
- `/retest all`: run this in case the CentOS CI failed to start/report any test
  progress or results

</details>

PR #2406, opened 2021-08-12
rbd: rbd-nbd logging the Ceph-CSI way

pkalever:

- Provide host-path for rbd-nbd logging, to enable the logging and persist the logs
- One logfile per device/volume
- Add ability to customize the logDir, default: /var/log/ceph
- Update the meta stash with logDir details
- Use the same to remove logfile on unStage/unmap to be space-efficient
- Provide variable to alter host-path location for ceph clients via helm
- Update rbd-nbd doc about log path details

---

### Summary (behavior with this series)

If you are using the default rbd nodeplugin daemonset and StorageClass templates then `cephLogDir` will be `/var/log/ceph`, this directory will be a host-path and the default log file path will be `/var/log/ceph/rbd-nbd-<volID>.log`. rbd-nbd creates a log file per volume under the `cephLogDir` path on NodeStage(map) and removed the same on the respective NodeUnstage(unmap).

In case if you need a customized log path, you should do the below:

- Edit the daemonset templates to change the `cephLogDir`
  - If you are using helm charts, then you can use key `cephLogDir`

    ```
    helm install --set cephLogDir=/var/log/ceph-csi/my-dir
    ```

  - For standard templates edit [csi-rbdplugin.yaml](https://github.com/ceph/ceph-csi/blob/devel/deploy/rbd/kubernetes/csi-rbdplugin.yaml) to update `hostPath` for `ceph-logdir`, also edit psp [csi-nodeplugin-psp.yaml](https://github.com/ceph/ceph-csi/blob/devel/deploy/rbd/kubernetes/csi-nodeplugin-psp.yaml) to update `pathPrefix` spec entries.
- Update the StorageClass with the customized log directory path
  - Now update rbd StorageClass for `cephLogDir`, for example

    ```
    cephLogDir: "/var/log/prod-A-logs"
    ```

`NOTE`:

- On uninstall make sure to delete `cephLogDir` on host manually to freeup
  some space just in case if there are any uncleaned log files.
- In case if you do not need the rbd-nbd logging to persistent, then just
  update the StorageClass for `cephLogDir` to use a non-persistent path.

---
### Log file naming (with this series)

```
$ ls /var/log/ceph/                                                                                              
rbd-nbd-0001-0024-fed5480a-f00f-417a-a51d-31d8a8144c03-0000000000000003-eba90b33-0156-11ec-a30b-4678a93686c2.log
rbd-nbd-0001-0024-fed5480a-f00f-417a-a51d-31d8a8144c03-0000000000000003-ebdc64a1-0156-11ec-a30b-4678a93686c2.log
rbd-nbd-0001-0024-fed5480a-f00f-417a-a51d-31d8a8144c03-0000000000000003-ec034226-0156-11ec-a30b-4678a93686c2.log
rbd-nbd-0001-0024-fed5480a-f00f-417a-a51d-31d8a8144c03-0000000000000003-ec55a4a7-0156-11ec-a30b-4678a93686c2.log
rbd-nbd-0001-0024-fed5480a-f00f-417a-a51d-31d8a8144c03-0000000000000003-f1111e4e-0155-11ec-a30b-4678a93686c2.log
``` 

#### Also see:

- https://github.com/ceph/ceph-csi/pull/2393
- https://github.com/ceph/ceph-csi/pull/2407

---

<details>

  progress or results

</details>


pkalever:
Most of them failed at deploy, probably something is wrong with this patch, not sure, let me test it once again.

Madhu-1:
Do you need to add this Directory to PSP?
 
https://github.com/ceph/ceph-csi/blob/devel/deploy/rbd/kubernetes/csi-nodeplugin-psp.yaml#L39
https://github.com/ceph/ceph-csi/blob/devel/deploy/rbd/kubernetes/csi-provisioner-psp.yaml#L31
https://github.com/ceph/ceph-csi/blob/devel/charts/ceph-csi-rbd/templates/provisioner-psp.yaml#L34
https://github.com/ceph/ceph-csi/blob/devel/charts/ceph-csi-rbd/templates/nodeplugin-psp.yaml#L43

pkalever:
> Do you need to add this Directory to PSP?
> 
> https://github.com/ceph/ceph-csi/blob/devel/deploy/rbd/kubernetes/csi-nodeplugin-psp.yaml#L39
> https://github.com/ceph/ceph-csi/blob/devel/deploy/rbd/kubernetes/csi-provisioner-psp.yaml#L31
> https://github.com/ceph/ceph-csi/blob/devel/charts/ceph-csi-rbd/templates/provisioner-psp.yaml#L34
> https://github.com/ceph/ceph-csi/blob/devel/charts/ceph-csi-rbd/templates/nodeplugin-psp.yaml#L43

Oh! probably this is why the tests were failing!?
Thanks @Madhu-1 will fix it.

charts/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
```diff
@@ -124,6 +124,8 @@ spec:
               mountPropagation: "Bidirectional"
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
+            - name: rbd-nbd-logdir
```
nixpanic: This is not only for `rbd-nbd` logs, but for other Ceph logs as well (in case enabled). Maybe call this `ceph-logdir` instead?

pkalever: @nixpanic  I mean that!
on the csi-rbdplugin the only ceph client logs will be from rbd-nbd atleast now!?

nixpanic: yes, only rbd-nbd logs at the moment, at least if there is no other manual configuration set. I think it is possible to reconfigure clients to write logs to their own files under `/var/log/ceph`, including the clients created by (go-ceph) libraries. There is just no need to br rbd-nbd specific in for generic paths like this.

humblec: @pkalever afaik, mounting a host dir for an application to log while running in container/pod is not a practice in kube/ocp land.  The pod restarts are handlded with same kube logging and fetch mechanism for almost all type of containers , this looks to be a non ideal approach

pkalever: @humblec 
Yes, I'm aware of this and there is an option to rbd-nbd push the logs to stderr (like what rook does https://github.com/rook/rook/blob/master/pkg/operator/ceph/config/defaults.go#L45).

But we are running rbd-nbd and csi-rbd plugin components of ceph-csi as part of the same container. The log formats of rbd plugin and rbd-nbd are different, hence it will not be wise to club the logs.

@nixpanic @Madhu-1 thoughts are welcome.

Thanks!

Madhu-1: AFAIK we don't persist the logs on the node in any of the ceph components in rook. is this required only for E2E or for upstream users also?

if this need to be the default
* This consumes the storage on the nodes
* Need steps to uninstall (removal of this directory)
* or instead of the default, we can document the steps to configure it? 

If this is only for E2E
* Update the daemonset templates in the E2E

pkalever: > AFAIK we don't persist the logs on the node in any of the ceph components in rook. is this required only for E2E or for upstream users also?

Probably they should persist the ceph server logs somewhere, without which debugging issue will not be possible. So I think in some way the logs will still be persistent somewhere. As an old school fellow, from my experience with gluster, we used to bind mount just like what I did in this PR.

> 
> if this need to be the default

Yes, this is not only for our e2e.

> 
> * This consumes the storage on the nodes

Yes but, IMHO rbd-nbd logs with default log level will not be a lot, probably a few KBs.

> * Need steps to uninstall (removal of this directory)
> * or instead of the default, we can document the steps to configure it?

sounds good, we can update the documentation about editing the sections of the template?

Thanks!

pkalever: @nixpanic I have respin changing name to `ceph-logdir`

Madhu-1: >sounds good, we can update the documentation about editing the sections of the template?

@pkalever sounds good.  is it something you are taking care of in this PR?

pkalever: Sure Madhu, it make sense to get it done within the same PR.
I will update here: https://github.com/ceph/ceph-csi/blob/devel/docs/rbd-nbd.md#configuration

pkalever: @Madhu-1 PTAL. Thanks!

internal/rbd/rbd_attach.go
```diff
@@ -239,6 +241,9 @@ func appendDeviceTypeAndOptions(cmdArgs []string, isNbd, isThick bool, userOptio
 		if !strings.Contains(userOptions, setNbdReattach) {
 			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("%s=%d", setNbdReattach, defaultNbdReAttachTimeout))
 		}
+		if !strings.Contains(userOptions, setLogFile) {
```
Madhu-1: IMO we should not default this one. not useful for everyone. but as you mentioned this can be configured in the storageclass hope that is enough.

pkalever: AFAIK, there is no way to set the log-dir for rbd clients, there is only a way to set log-file by the complete path.
I want to have a consistent naming for rbd-nbd log file like: rbd-nbd.log, unlike ceph-client.$id.log

If you don't like it feel free to show up :+1: I can remove it.

Madhu-1: No, what I meant was not to set default logging if someone does not specify it in the storageclass.
one more question, If multiple clients write to the same file how do identify which pod/client the log belongs to?

nixpanic:
I have a suggesting to consider for an alternative solution, one that is more container friendly:

- use a shared volume where the `rbd-nbd.log` is stored
- add a side-car to the daemonset that runs `tail -f` on the log from rbd-nbd (retry to run `tail` in case the log does not exist yet)

This makes it possible to get the logs with standard kubernetes calls, and there is no permanent storage on the node needed. It is possible to use this in any environment, not only our e2e.

If there is a log that `rbd map` and similar calls use, there would be the additional benefit that better debugging of Ceph CLI calls becomes possible (just use the same log file).

internal/rbd/rbd_attach.go
pkalever: > No, what I meant was not to set default logging if someone does not specify it in the storageclass.

So, I get you right. We don't need a standard logfile name like: rbd-nbd.log ? You mean to use the ceph-client.admin.log name'ing?

> one more question, If multiple clients write to the same file how do identify which pod/client the log belongs to?

This is something I'm working on currently, ideally rbd-nbd should prefix the device number (like nbd0, nbd1 .. etc) which is not the case currently. But in the future expect it logs with the device index as prefix.

Thanks!

pkalever:
> I have a suggesting to consider for an alternative solution, one that is more container friendly:
> 
>     * use a shared volume where the `rbd-nbd.log` is stored

Do you mean, we don't have to create a host path ourself? Or is it an alternative like the latest changes which gives provision to tweak a custom logfile path from storageclass?

> 
>     * add a side-car to the daemonset that runs `tail -f` on the log from rbd-nbd (retry to run `tail` in case the log does not exist yet)
> 
> 
> This makes it possible to get the logs with standard kubernetes calls, and there is no permanent storage on the node needed. It is possible to use this in any environment, not only our e2e.
> 
> If there is a log that `rbd map` and similar calls use, there would be the additional benefit that better debugging of Ceph CLI calls becomes possible (just use the same log file).

Yes, this should be an alternative to address to meat the K8s standards.

nixpanic:
On Wed, Aug 18, 2021 at 05:52:15AM -0700, Prasanna Kumar Kalever wrote:
> > I have a suggesting to consider for an alternative solution, one that is more container friendly:
> > 
> >     * use a shared volume where the `rbd-nbd.log` is stored
> 
> Do you mean, we don't have to create a host path ourself? Or is it an alternative like the latest changes which gives provision to tweak a custom logfile path from storageclass?

Yes, no `hostPath` but a volume with type `emptyDir`. It can be placed
on /var/log/ceph or whatever the default location is for logs.

Madhu-1:
> On Wed, Aug 18, 2021 at 05:52:15AM -0700, Prasanna Kumar Kalever wrote: > I have a suggesting to consider for an alternative solution, one that is more container friendly: > > * use a shared volume where the `rbd-nbd.log` is stored Do you mean, we don't have to create a host path ourself? Or is it an alternative like the latest changes which gives provision to tweak a custom logfile path from storageclass?
> Yes, no `hostPath` but a volume with type `emptyDir`. It can be placed on /var/log/ceph or whatever the default location is for logs.

If we use emtpyDir again the logs won't persist for multiple restarts? Can we take the below root for this one?

* Log by default to /var/log/ceph
* Log the rbd-nbd to a different path if it's provided in the storageclass
* template changes to different log path
* update storageclass/documentation for logfile path configuration
* Update documentation for the uninstall

pkalever:
> > On Wed, Aug 18, 2021 at 05:52:15AM -0700, Prasanna Kumar Kalever wrote: > I have a suggesting to consider for an alternative solution, one that is more container friendly: > > * use a shared volume where the `rbd-nbd.log` is stored Do you mean, we don't have to create a host path ourself? Or is it an alternative like the latest changes which gives provision to tweak a custom logfile path from storageclass?
> > Yes, no `hostPath` but a volume with type `emptyDir`. It can be placed on /var/log/ceph or whatever the default location is for logs.
> 
> If we use emtpyDir again the logs won't persist for multiple restarts? Can we take the below root for this one?
> 
> * Log by default to /var/log/ceph

Log to **/var/log/ceph/** by default, anyone doesn't need it persisting then just edit the storageclass and change it to a directory that is not bind-mounted.

Example: Maybe
`mapOptions: log-file=/var/lohg/ceph-csi/rbd-nbd.log`

> * Log the rbd-nbd to a different path if it's provided in the storageclass
> * template changes to different log path

That's already handled. I will also add variables for the helm charts case which allows to alter the hostpath location.

> * update storageclass/documentation for logfile path configuration
> * Update documentation for the uninstall

OK.


@nixpanic any objections?

Apparently, @Madhu-1 checked rook code and at rook, ceph server logs are well persistent in a hostpath. They also have a `log-collector` sidecar container which logroates the ceph logs at the hostpath directory. Moving forward we can use the same hostpath used by rook and take advantage of logrotation.

pkalever:
@Mergifyio rebase

pkalever:
@Madhu-1 @nixpanic I have updated this PR as per our discussion, please take a look.

Thanks!

internal/rbd/nodeserver.go
```diff
@@ -824,7 +828,8 @@ func (ns *NodeServer) NodeUnstageVolume(
 		imgInfo.NbdAccess,
 		imgInfo.Encrypted,
 		req.GetVolumeId(),
-		imgInfo.UnmapOptions); err != nil {
+		imgInfo.UnmapOptions,
```
Madhu-1: Maybe as a cleanup:
Looks like the number of arguments of this function is growing. can we add detachRBDImageOrDeviceSpec as method to `imgInfo`? or make `imgInfo` as an argument instead of passing individual parameters?

internal/rbd/rbd_attach.go
```diff
@@ -417,6 +424,13 @@ func detachRBDImageOrDeviceSpec(
 
 		return fmt.Errorf("rbd: unmap for spec (%s) failed (%w): (%s)", imageOrDeviceSpec, err, stderr)
 	}
+	if isNbd && logDir != "" {
+		logFile := fmt.Sprintf("%s/rbd-nbd-%s.log", logDir, volumeID)
```
Madhu-1: Can we have a common helper function to return the log file path?

Madhu-1 (commented):
no major objects. small suggestions

internal/rbd/nodeserver.go
pkalever: There are also other callers of detachRBDImageOrDeviceSpec() which doesn't use imgInfo. Hence it wouldn't be ideal.
Just have a check Madhu.

internal/rbd/rbd_attach.go
pkalever: Sure :+1:

pkalever: Done now. Ready for review again :-)

pkalever:
/retest ci/centos/mini-e2e/k8s-1.21

---

```
! StartHost failed, but will try again: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute
```

internal/rbd/rbd_util.go
```diff
@@ -2002,3 +2006,16 @@ func (ri *rbdImage) addSnapshotScheduling(
 
 	return nil
 }
+
+// getCephClientLogFileName compiles the complete log file path based on inputs.
+func getCephClientLogFileName(id, logDir, prefix string) string {
```
Madhu-1, line 2011: can you please add unit testing for this one and do we need to have prefix check here?

Madhu-1 (commented):
LGTM no major comments

docs/rbd-nbd.md
```diff
@@ -28,6 +29,41 @@ client-side, which is inside the `csi-rbdplugin` node plugin.
...
 in the StorageClass.
 
+### Configuring logging path
+
+If you are using the default rbd nodeplugin daemonset and StorageClass
+templates then `cephLogDir` will be `/var/log/ceph`, this directory will be
+a host-path and the default log file path will be
+`/var/log/ceph/rbd-nbd-<volID>.log`. rbd-nbd creates a log file per volume
+under the `cephLogDir` path on NodeStage(map) and removed the same on
+the respective NodeUnstage(unmap).
```
humblec, line 39: @pkalever just to understand, I think the main reason for this patch set was that, somehow we were looking for rbd nbd logs to be persisted. if `nodestage` and `nodeunstage` create and remove the log dir, I am not sure how that helps?  if the plan is to get the logs while staged, why we need hostdir? Am i missing something here?

pkalever: This was according to the discussion in our previous standup call.

Yes, we clean the logs in `unmap` and not on `detach`. Why do we want to analyze the logs of successful volumes? Mainly we will be looking into the logs when we see some issue with the volume and those logs will not be removed as there will not be a unmap on the odd ones or fail to map in the first place.

The idea here is to be more space-efficient.

internal/rbd/rbd_util.go
pkalever: @Madhu-1 wanted to make this function look simple and not return any error and then handle it exclusively from the caller, I'm worried that kills the purpose of making this simplified. So just in case if anyone passes prefix as nothing, use "ceph-" which is quite common for all ceph logs. Again that's just me!

Madhu-1 (dismissed):

internal/rbd/rbd_util.go
Madhu-1: Looks like You are passing prefix as rbd-nbd always that's why the other question

internal/rbd/rbd_attach.go
```diff
@@ -384,7 +382,7 @@ func detachRBDImageOrDeviceSpec(
 	ctx context.Context,
 	imageOrDeviceSpec string,
 	isImageSpec, isNbd, encrypted bool,
-	volumeID, unmapOptions string) error {
+	volumeID, unmapOptions, logDir string) error {
```
nixpanic: This really gets ugly, if passing an existing structure is not practical, create a new `detachRBDImageArgs` with all the needed parameters. That makes calling this function so much simpler to understand.

pkalever: Done now, PTAL.

internal/rbd/rbd_util.go
nixpanic: Please add a test to `internal/rbd/rbd_util_test.go` so that some of the expected results are documented and validated automatically.

pkalever: Since @Madhu-1 and @nixpanic both are asking for this, I have added the unit test for getCephClientLogFileName().

PTAL now.

pkalever:

---
```
Aug 23 16:44:11.387: FAIL: failed to write IO, err: command terminated with exit code 2, stdErr: /bin/sh: 1: cannot create /var/lib/www/html/test: Read-only file system

```

Madhu-1 (approved):

nixpanic (approved):

PR #2431, opened 2021-08-23
rbd: add EnsureImageCleanup to ensure image cleanup from trash 

### LINKED ISSUE #1728: Rbd images left in the trash
During the RBD volume deletion, we are doing 3 operation 
1. Find image present in the pool
2. Move the image to the trash
3. Add a task to remove the image from the trash

In case if the 3rd step fails cephcsi will return an internal server error and the external-provisioner will retry again, now cephcsi will get a request, it will try to find the image in the pool now as the image is not present it will assume that image not present and it will continue with the omap cleanup

Do cephcsi need to check images in the trash and add a task to remove them from trash or let ceph take care of removing images in the trash (this is functionality available in ceph where ceph will clean up images in the trash after some time?). 

@dillaman what would be the best approach we need to take here?

Rakshith-R:
After moving moving image to trash, if `trash remove` step fails,
then external-provisioner will issue subsequent requests, in which
image will be absent in pool( will be in trash) and omap cleanup will
be done with stale image left in trash with no `trash remove` step on it.

To avoid this scenario list trash images and find corresponding id for given
image name and add a task to flatten when we encounter a ErrImageNotFound.

Fixes: ceph#1728

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -2761,6 +2761,109 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("validate image deletion when it is moved to trash", func() {
+				// make sure pool is empty
+				validateRBDImageCount(f, 0, defaultRBDPool)
+
+				err := createRBDSnapshotClass(f)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass: %v", err)
+				}
+				defer func() {
+					err = deleteRBDSnapshotClass()
+					if err != nil {
+						e2elog.Failf("failed to delete VolumeSnapshotClass: %v", err)
+					}
+				}()
+
+				pvc, err := loadPVC(pvcPath)
+				if err != nil {
+					e2elog.Failf("failed to load pvc: %v", err)
+				}
+				pvc.Namespace = f.UniqueName
+
+				err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create pvc: %v", err)
+				}
+
+				pvcSmartClone, err := loadPVC(pvcSmartClonePath)
+				if err != nil {
+					e2elog.Failf("failed to load pvcSmartClone: %v", err)
+				}
+				pvcSmartClone.Namespace = f.UniqueName
+
+				err = createPVCAndvalidatePV(f.ClientSet, pvcSmartClone, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create pvc: %v", err)
+				}
+
+				snap := getSnapshot(snapshotPath)
+				snap.Namespace = f.UniqueName
+				snap.Spec.Source.PersistentVolumeClaimName = &pvc.Name
+				err = createSnapshot(&snap, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create snapshot: %v", err)
+				}
+
+				smartCloneImageData, err := getImageInfoFromPVC(pvcSmartClone.Namespace, pvcSmartClone.Name, f)
+				if err != nil {
+					e2elog.Failf("failed to get ImageInfo from pvc: %v", err)
+				}
+
+				imageList, err := listRBDImages(f, defaultRBDPool)
+				if err != nil {
+					e2elog.Failf("failed to list rbd images: %v", err)
+				}
+				for _, imageName := range imageList {
+					if imageName == smartCloneImageData.imageName {
+						// do not move smartclone image to trash to test
+						// temporary image clone cleanup.
+						continue
+					}
+					_, _, err = execCommandInToolBoxPod(f,
+						fmt.Sprintf("rbd snap purge %s %s", rbdOptions(defaultRBDPool), imageName), rookNamespace)
+					if err != nil {
+						e2elog.Failf(
+							"failed to snap purge %s %s: %v",
+							imageName,
+							rbdOptions(defaultRBDPool),
+							err)
+					}
+					_, _, err = execCommandInToolBoxPod(f,
+						fmt.Sprintf("rbd trash move %s %s", rbdOptions(defaultRBDPool), imageName), rookNamespace)
+					if err != nil {
+						e2elog.Failf(
+							"failed to move rbd image %s %s to trash: %v",
+							imageName,
+							rbdOptions(defaultRBDPool),
+							err)
+					}
+				}
+
+				err = deleteSnapshot(&snap, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete snapshot: %v", err)
+				}
+
+				err = deletePVCAndValidatePV(f.ClientSet, pvcSmartClone, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete pvc: %v", err)
+				}
+
+				err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete pvc: %v", err)
+				}
+
+				validateRBDImageCount(f, 0, defaultRBDPool)
+
+				err = waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to validate rbd images in trash %s: %v", rbdOptions(defaultRBDPool), err)
+				}
+			})
+
 			By("validate stale images in trash", func() {
 				err := waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
 				if err != nil {
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -565,6 +565,26 @@ func (rv *rbdVolume) getTrashPath() string {
 	return trashPath + "/" + rv.ImageID
 }
 
+// ensureImageCleanup finds image in trash and if found removes it
+// from trash.
+func (rv *rbdVolume) ensureImageCleanup(ctx context.Context) error {
+	trashInfoList, err := librbd.GetTrashList(rv.ioctx)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to list images in trash: %v", err)
+
+		return err
+	}
+	for _, val := range trashInfoList {
+		if val.Name == rv.RbdImageName {
+			rv.ImageID = val.Id
+
+			return trashRemoveImage(ctx, rv, rv.conn.Creds)
+		}
+	}
+
+	return nil
+}
+
 // deleteImage deletes a ceph image with provision and volume options.
 func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
 	image := pOpts.RbdImageName
@@ -597,6 +617,12 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 		return err
 	}
 
+	return trashRemoveImage(ctx, pOpts, cr)
+}
+
+// trashRemoveImage adds a task to trash remove an image using ceph manager if supported,
+// otherwise removes the image from trash.
+func trashRemoveImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
 	// attempt to use Ceph manager based deletion support if available
 
 	args := []string{
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -1383,7 +1383,12 @@ func (cs *ControllerServer) DeleteSnapshot(
 
 	err = rbdVol.getImageInfo()
 	if err != nil {
-		if !errors.Is(err, ErrImageNotFound) {
+		if errors.Is(err, ErrImageNotFound) {
+			err = rbdVol.ensureImageCleanup(ctx)
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+		} else {
 			log.ErrorLog(ctx, "failed to delete rbd image: %s/%s with error: %v", rbdVol.Pool, rbdVol.VolName, err)
 
 			return nil, status.Error(codes.Internal, err.Error())
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -776,8 +776,13 @@ func (cs *ControllerServer) checkErrAndUndoReserve(
 		return &csi.DeleteVolumeResponse{}, nil
 	}
 
-	// All errors other than ErrImageNotFound should return an error back to the caller
-	if !errors.Is(err, ErrImageNotFound) {
+	if errors.Is(err, ErrImageNotFound) {
+		err = rbdVol.ensureImageCleanup(ctx)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	} else {
+		// All errors other than ErrImageNotFound should return an error back to the caller
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
@@ -924,8 +929,13 @@ func cleanupRBDImage(ctx context.Context,
 	tempClone := rbdVol.generateTempClone()
 	err = deleteImage(ctx, tempClone, cr)
 	if err != nil {
-		// return error if it is not ErrImageNotFound
-		if !errors.Is(err, ErrImageNotFound) {
+		if errors.Is(err, ErrImageNotFound) {
+			err = tempClone.ensureImageCleanup(ctx)
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+		} else {
+			// return error if it is not ErrImageNotFound
 			log.ErrorLog(ctx, "failed to delete rbd image: %s with error: %v",
 				tempClone, err)
 


internal/rbd/rbd_util.go
```diff
@@ -564,6 +564,16 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 
 		return err
 	}
+	defer func() {
```
nixpanic: Indention looks really weird in GitHub while reviewing this change... I wonder if there is some strange space/tab mismatch here.

nixpanic (commented):
Looks good in general, but what is the expectation towards the user when the image is restored from trash? As the encryption passphrase has been deleted, the image can not be used anymore... The expectations should probably be documented.

Rakshith-R:
> Looks good in general, but what is the expectation towards the user when the image is restored from trash? As the encryption passphrase has been deleted, the image can not be used anymore... The expectations should probably be documented.

@nixpanic 
There's actually no way to recover/cancel a deletion request on a PVC as far as my findings go.
This image is certianly to be deleted in subsequent request from external-provisioner(which will definitely be fired).

nixpanic:
On Mon, Aug 23, 2021 at 05:53:13AM -0700, Rakshith R wrote:
> > Looks good in general, but what is the expectation towards the user when the image is restored from trash? As the encryption passphrase has been deleted, the image can not be used anymore... The expectations should probably be documented.
> 
> @nixpanic 
> There's actually no way to recover/cancel a deletion request on a PVC as far as my findings go.
> This image is certianly to be deleted in subsequent request from external-provisioner(which will definitely be fired).

Right, that is what I assumed too. Documenting this in the commit
message or in the comments of the function would be good.

internal/rbd/rbd_util.go
Rakshith-R: >Indention looks really weird in GitHub while reviewing this change... I wonder if there is some strange space/tab mismatch here.

Double checked, looks fine to me. Maybe because of nesting depth, it looks that way.

Rakshith-R:
> Right, that is what I assumed too. Documenting this in the commit message or in the comments of the function would be good.

Documented the behavior in both the comment and commit message. Hope it is clearer now.

internal/rbd/rbd_util.go
```diff
@@ -564,6 +564,19 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
+		// restore image from trash to avoid leaking stale image.
+		// If `trash remove` step fails, then in subsequent request, image should be
+		// found in the pool for deletion to resume normally. Hence, Restore the image
+		// from trash on err from `trash remove` step.
+		if err != nil {
```
Madhu-1: Thanks for giving it a try, but this fix will still leave the stale images if the function inside defer fails or never gets called due to restart, etc

The way I was thinking to fix this one was

* If the image Not found in the pool do below operations as part of delete
* As we store the imageID check do we have any image in the trash with this ID
* In some cases we don't store ID and we just move the image to the trash in that case check do we have an image with a name in the Trash
* If the Image is found add a task to remove it from the trash
* If the image is not found return notFound or success as expected.

Madhu-1 (commented):
Please add an E2E for this one by moving the image to the trash and try to delete the Volume.

internal/rbd/rbd_util.go
Rakshith-R: > Thanks for giving it a try, but this fix will still leave the stale images if the function inside defer fails or never gets called due to restart, etc

This mainly did focus on just errors from `trash remove` api calls.
 pod restart at a particular step is highly unlikely.

> 
> The way I was thinking to fix this one was
> 
> * If the image Not found in the pool do below operations as part of delete
> * As we store the imageID check do we have any image in the trash with this ID

We can directly restore image from trash here with imageID(if imageID != "") 
if found, send back an error and let another subsequent request handle it gracefully.
else, continue with cleanup.

> * In some cases we don't store ID and we just move the image to the trash in that case check do we have an image with a name in the Trash

IMO, we have been using and storing image ID for a very long time now and this is a cleanup bug patch and covering latest use cases is sufficient. 

The following steps increase complexity and add too many api calls.
(trash ls, find image, construct trash path, add task to remove  )

> * If the Image is found add a task to remove it from the trash
> * If the image is not found return notFound or success as expected.

@Madhu-1 , let me know what you think.

Madhu-1: > > Thanks for giving it a try, but this fix will still leave the stale images if the function inside defer fails or never gets called due to restart, etc
> 
> This mainly did focus on just errors from `trash remove` api calls.
> pod restart at a particular step is highly unlikely.


Not sure why we want to fix it half not the full as we know the bug. IMO we should not work on only one scenario

> > The way I was thinking to fix this one was
> > 
> > * If the image Not found in the pool do below operations as part of delete
> > * As we store the imageID check do we have any image in the trash with this ID
> 
> We can directly restore image from trash here with imageID(if imageID != "")
> if found, send back an error and let another subsequent request handle it gracefully.
> else, continue with cleanup.

Why do you want to restore the image which is already in the trash (Have you verified how much time restore can take
depends on the image size? if it takes less time we can consider that but that is not enough to handle all the corner cases)


> > * In some cases we don't store ID and we just move the image to the trash in that case check do we have an image with a name in the Trash
> 
> IMO, we have been using and storing image ID for a very long time now and this is a cleanup bug patch and covering latest use cases is sufficient.
> 

Are we storing the imageID of the temp clone which we create during the PVC-PVC clone? and also consider the case where image creation got succeeded but next operation like storing the imageID in omap fails etc

> The following steps increase complexity and add too many api calls.
> (trash ls, find image, construct trash path, add task to remove )
> 

Am open to better suggestion to fix it completely not partially.
 
> > * If the Image is found add a task to remove it from the trash
> > * If the image is not found return notFound or success as expected.
> 
> @Madhu-1 , let me know what you think.

internal/rbd/rbd_util.go
```diff
@@ -597,6 +597,28 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
+			pOpts.ImageID = val.Id
+
+			return TrashRemoveImage(ctx, pOpts, cr)
+		}
+	}
+
+	return nil
+}
+
+func TrashRemoveImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
```
Madhu-1: don't export these functions keep it internal.

internal/rbd/rbd_util.go
```diff
@@ -597,6 +597,28 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 		return err
 	}
 
+	return TrashRemoveImage(ctx, pOpts, cr)
+}
+
+func EnsureImageCleanup(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
```
Madhu-1: make this a method of rbdVolume?

e2e/rbd.go
```diff
@@ -2761,6 +2761,110 @@ var _ = Describe("RBD", func() {
...
+				err = deletePVCAndValidatePV(f.ClientSet, pvcSmartClone, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete pvc: %v", err)
+				}
+
+				err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete pvc: %v", err)
+				}
+
```
Madhu-1, line 2858: add validation for `0` images in the pool

Rakshith-R:
@mergifyio rebase

Rakshith-R:
> @Mergifyio rebase


3 ci e2e failed due to timeout, #2264, http error and rebased on top of devel.
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/mini-e2e-helm_k8s-1.20/2767/display/redirect
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/mini-e2e-helm_k8s-1.21/1432/display/redirect
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/mini-e2e_k8s-1.21/1447/display/redirect

@ceph/ceph-csi-contributors PTAL.

Rakshith-R:

failed due to #2264 

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/upgrade-tests-rbd/detail/upgrade-tests-rbd/3366/pipeline

Rakshith-R:


panic: test timed out after 1h30m0s
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.20/detail/mini-e2e-helm_k8s-1.20/2786/pipeline

Rakshith-R:
/retest ci/centos/mini-e2e/k8s-1.21

panic: test timed out after 1h30m0s
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.21/detail/mini-e2e_k8s-1.21/1466/pipeline

e2e/rbd.go
```diff
@@ -2761,6 +2761,112 @@ var _ = Describe("RBD", func() {
...
+				validateRBDImageCount(f, 0, defaultRBDPool)
+
+				err := createRBDSnapshotClass(f)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass with error %v", err)
+				}
+				defer func() {
+					err = deleteRBDSnapshotClass()
+					if err != nil {
+						e2elog.Failf("failed to delete VolumeSnapshotClass: %v", err)
```
Madhu-1: `with error` is not required?

e2e/rbd.go
```diff
@@ -2761,6 +2761,112 @@ var _ = Describe("RBD", func() {
...
+					var stdErr string
+					if imageName == smartCloneImageData.imageName {
+						// do not move smartclone image to trash to test
+						// temporary image clone cleanup.
+						continue
+					}
+					_, stdErr, err = execCommandInToolBoxPod(f,
+						fmt.Sprintf("rbd snap purge %s %s", rbdOptions(defaultRBDPool), imageName), rookNamespace)
+					if err != nil || stdErr != "" {
+						e2elog.Logf(
```
Madhu-1: make it as Failf or else the below command fails, it becomes hard to find the issue.

e2e/rbd.go
```diff
@@ -2761,6 +2761,112 @@ var _ = Describe("RBD", func() {
...
+							"failed to snap purge %s/%s: %v stdErr: %v",
+							defaultRBDPool,
```
Madhu-1: log  `rbdOptions(defaultRBDPool)` instead of defaultRBDPool

e2e/rbd.go
```diff
@@ -2761,6 +2761,112 @@ var _ = Describe("RBD", func() {
...
+							imageName,
+							err,
+							stdErr)
+					}
+					_, stdErr, err = execCommandInToolBoxPod(f,
+						fmt.Sprintf("rbd trash move %s %s", rbdOptions(defaultRBDPool), imageName), rookNamespace)
+					if err != nil || stdErr != "" {
+						e2elog.Failf(
+							"failed to move rbd image %s/%s to trash: %v stdErr: %v",
```
Madhu-1: log  `rbdOptions(defaultRBDPool)` instead of defaultRBDPool

e2e/rbd.go
```diff
@@ -2761,6 +2761,112 @@ var _ = Describe("RBD", func() {
...
+				err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete pvc: %v", err)
+				}
+
+				validateRBDImageCount(f, 0, defaultRBDPool)
+
+				err = waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to validate rbd images in pool %s trash with error %v", defaultRBDPool, err)
```
Madhu-1: same here as  well log  `rbdOptions(defaultRBDPool)` instead of defaultRBDPool? and remove `with error`?

internal/rbd/rbd_util.go
```diff
@@ -565,6 +565,24 @@ func (rv *rbdVolume) getTrashPath() string {
 	return trashPath + "/" + rv.ImageID
 }
 
+func (rv *rbdVolume) ensureImageCleanup(ctx context.Context) error {
+	trashInfoList, err := librbd.GetTrashList(rv.ioctx)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to list images in trash: %w", err)
```
Madhu-1: `%w` is not needed for logging it should be `%v`

internal/rbd/rbd_util.go
```diff
@@ -597,6 +615,10 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 		return err
 	}
 
+	return trashRemoveImage(ctx, pOpts, cr)
+}
+
+func trashRemoveImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
```
Madhu-1, line 625: can we make this a method of rbdVolume? if we do that we can avoid passing the two parameters to the function

Madhu-1 (changes requested):
few nits. changes LGTM

internal/rbd/rbd_util.go
Rakshith-R: > can we make this a method of rbdVolume? if we do that we can avoid passing the two parameters to the function

I doubt we copied cr into pOpts everywhere. 

I'll send a pr to use go-ceph rbd task api for flatten and trash remove, which will remove the need for passing `cr` everywhere and make it a method on rbdVolume then(+ some cleanup too).

e2e/rbd.go
```diff
@@ -2761,6 +2761,109 @@ var _ = Describe("RBD", func() {
...
+				pvc.Namespace = f.UniqueName
+
+				err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create pvc: %v", err)
+				}
+
+				pvcSmartClone, err := loadPVC(pvcSmartClonePath)
+				if err != nil {
+					e2elog.Failf("failed to load pvcSmartClone : %v", err)
```
Madhu-1: `pvcSmartClone : %v` to `pvcSmartClone: %v`

e2e/rbd.go
```diff
@@ -2761,6 +2761,109 @@ var _ = Describe("RBD", func() {
...
+				err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete pvc: %v", err)
+				}
+
+				validateRBDImageCount(f, 0, defaultRBDPool)
+
+				err = waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to validate rbd images in trash %q: %v", rbdOptions(defaultRBDPool), err)
```
Madhu-1: be consistent with formatting  use `%s` here

Rakshith-R:
@mergifyio rebase

internal/rbd/rbd_util.go
```diff
@@ -565,6 +565,24 @@ func (rv *rbdVolume) getTrashPath() string {
 	return trashPath + "/" + rv.ImageID
 }
 
+func (rv *rbdVolume) ensureImageCleanup(ctx context.Context) error {
```
humblec, line 570: good to add some source code comment here

internal/rbd/rbd_util.go
```diff
@@ -565,6 +565,26 @@ func (rv *rbdVolume) getTrashPath() string {
 	return trashPath + "/" + rv.ImageID
 }
 
+// ensureImageCleanup finds image in trash and if found removes it
+// from trash.
+func (rv *rbdVolume) ensureImageCleanup(ctx context.Context) error {
+	trashInfoList, err := librbd.GetTrashList(rv.ioctx)
```
humblec, line 571: as we are calling `ensureImageCleanup` from more than caller, I think reading the trash list can be racy. Isnt it ?

Rakshith-R: > as we are calling `ensureImageCleanup` from more than caller, I think reading the trash list can be racy. Isnt it ? 

Does not matter,  we are just checking if that particular image is in trash and if it is, add a task/ remove from trash. 

And delete operation on pvc/ snapshot is already protected by locks.

humblec: so if its there and before the caller issue the task to remove from trash, if another thread did add a task , the first one going to fail?

Rakshith-R: > so if its there and before the caller issue the task to remove from trash, if another thread did add a task , the first one going to fail? 

Due to delete op locking, two threads will never be acting on same image. 

If there is a task already added, new cmd to add task will be mapped to the same task without any errors or failures.

humblec: yeah, even snap operation is on the lock, so should be safe here.

humblec (approved):

Madhu-1 (approved):

yati1998 (approved):

Rakshith-R:

failed due to #2264 
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.22/detail/mini-e2e-helm_k8s-1.22/798/pipeline

Rakshith-R:

Failed due to #2264 
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.22/detail/mini-e2e-helm_k8s-1.22/799/pipeline

nixpanic:

it seems there was some issue with the MON(s):
```
failed to validate rbd images in trash --pool=replicapool: failed to list images in trash 2021-10-20T13:52:59.494+0000 7fc1070f3700 -1 monclient: get_auth_request but no auth handler is set up
```
possibly the [logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.22/detail/mini-e2e-helm_k8s-1.22/803/pipeline/110) contain more details

Rakshith-R:
@mergifyio rebase

Issue #2438, opened 2021-08-25
Start deprecating previous CSI releases


humblec:
# Describe the bug #

We dont have a clear policy on which ceph csi releases are supported and what are deprecated, I would like to propose below support policy and would like to get opinion on this :

1) Deprecate or mark all the `N-1` major releases as deprecated

For ex: 
Considering we have release `3.X` now, mark all the releases `<v2.0.0` as deprecated and remove from the support. 

@nixpanic @Madhu-1 @agarwal-mudit  any thoughts on this ?

Madhu-1:
Sounds good to me to deprecate all releases less than N-1.

Yuggupta27:
> # Describe the bug
> We dont have a clear policy on which ceph csi releases are supported and what are deprecated, I would like to propose below support policy and would like to get opinion on this :
> 
> 1. Deprecate or all the `N-1` major releases as depreacted
> 
> For ex:
> Considering we have release `3.X` now, mark all the releases `<v2.0.0` as deprecated and remove from the support.
> 
> @nixpanic @Madhu-1 @agarwal-mudit any thoughts on this ?

Sure, we can deprecate N-1 and suggest an upgrade for the same :+1:

humblec:
Also what about `minor releases`? I think supporting last 2 or 3 minor releases are good enough , Isnt it ?

For example, on supported major versions, last 2 minor releases are supported. 

In that sense, with the release of `3.4.0` , we drop support for `3.1.0` ? and only support from `3.2.0`?

Madhu-1:
> Also what about `minor releases`? I think supporting last 2 or 3 minor releases are good enough , Isnt it ?
> 
> For example, on supported major versions, last 2 minor releases are supported.
> 
> In that sense, with the release of `3.4.0` , we drop support for `3.1.0` ? and only support from `3.2.0`?

IMO Minor releases will always be backward compatible and no breaking changes and we should support all the minor releases for the supported major releases. the deprecation should be only for major releases.

humblec:
also what about the `patch releases`?  considering patch releases are CVE fixes mostly, may be only last 2 patch releases in support?

nixpanic:
The process to deprecate a release should be part of the process that is used for releasing a new version. This should also include adding/removing `backport-to*` labels, updating Mergify rules, ...

Madhu-1:
@humblec anything pending on this one or we can close it?

humblec:
[...]

PR #2440, opened 2021-08-25
doc: update support matrix for deprecated ceph csi releases

humblec:
As discussed in https://github.com/ceph/ceph-csi/issues/2438
    we are marking ceph csi release support to N.(x-1) release versions.
    
    N = latest major release
    x = latest minor release
    
    This address the release version support matrix based on the
    same.

updates #2438 

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,13 @@ support other orchestration environments in the future.
 
 NOTE:
 
-- **`csiv0.3`** is deprecated with release of **`csi v1.1.0`**
+The supported window of Ceph CSI versions  is known as "N.(x-1)":
+(N (Latest major release) . (x (Latest minor release) - 1)).
+
+For example, if Ceph CSI latest major version is `3.4.0` today, support is
+provided for the versions above `3.3.0`. If users are running an unsupported
+Ceph CSI version, they will be asked to upgrade when requesting support for the
+cluster.
 
 ## Support Matrix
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,7 +55,11 @@ support other orchestration environments in the future.
 
 NOTE:
 
-- **`csiv0.3`** is deprecated with release of **`csi v1.1.0`**
+The supported window of Ceph CSI versions  is known as "N-1": (N (Latest major release) - 1 (major versions)).
+For example, if ceph csi latest major version is `3.0.0` today, support is provided for the versions above `2.0.0`
```
Madhu-1: ceph csi to `Ceph CSI`

Madhu-1 (changes requested):
update PR with fixes #2438?

Madhu-1 (dismissed):
update PR with fixes #2438?

README.md
humblec: done

Madhu-1 (dismissed):

README.md
```diff
@@ -55,7 +55,14 @@ support other orchestration environments in the future.
 
 NOTE:
 
-- **`csiv0.3`** is deprecated with release of **`csi v1.1.0`**
+The supported window of Ceph CSI versions  is known as "N-1":
+(N (Latest major release) - 1 (major versions)).
+
+For example, if Ceph CSI latest major version is `3.0.0` today, support is
+provided for the versions above `2.0.0`. If customers are running an unsupported
```
nixpanic: not `customers`, but `users`

README.md
```diff
@@ -55,7 +55,14 @@ support other orchestration environments in the future.
...
+Ceph CSI version, they will be asked to upgrade when requesting support for the
+cluster. Clusters running unsupported Ceph CSI releases are not covered by the
+support policies.
```
nixpanic: What `support policies`? Sounds like there is an SLA of some kind somewhere... If a policy is documented, link to it please.

nixpanic (dismissed):

README.md
humblec: @nixpanic done.

README.md
humblec: it was a general comment , however removed that line to avoid confusion.

humblec:
ptal @ceph/ceph-csi-contributors @ceph/ceph-csi-maintainers

Madhu-1 (commented):
LGTM, reserving approval right for @nixpanic as he requested changes

yati1998 (dismissed):

README.md
```diff
@@ -55,7 +55,13 @@ support other orchestration environments in the future.
...
 NOTE:
 
-- **`csiv0.3`** is deprecated with release of **`csi v1.1.0`**
+The supported window of Ceph CSI versions  is known as "N-1":
+(N (Latest major release) - 1 (major versions)).
+
+For example, if Ceph CSI latest major version is `3.0.0` today, support is
+provided for the versions above `2.0.0`. If users are running an unsupported
+Ceph CSI version, they will be asked to upgrade when requesting support for the
+cluster.
```
nixpanic, line 64: This effectively means we need to update v2.0, v2.1, v3.0, v3.1, .. for bugfixes and security issues (including vendored packages). This is definitely not what we do at the moment.

Should we not just stick to what  we currently do? Which is `3.x` and `3.x-1`, so we're at `release-3.4` now, meaning we'll maintain `release-3.3` until we have a version from `release-3.5`?

humblec: @nixpanic what you are pointing out here is the `minor` version support where the PR currently describe `major` version support.  I can add minor version support too with `N-1` , however the question is about the `major versions` and what is the policy for them , based on that only we can deprecate previous major versions?

nixpanic: I do not think we maintain the major version `2` anymore, eventhough we now released `3.4`. The current phrasing suggests we maintain (and fix bugs etc.) for `2.x`, which is definitely not the case.

Deprecating everything before `3.3` makes sense to me. That is effectively the only version we're still maintaining in some form (very minimal maintenance).

humblec: I am fine with that :+1: , adjusting the PR accordingly.

humblec: @nixpanic PR updated as per above. ptal

nixpanic (dismissed):
looks much better to me, and matches the reality at the moment

yati1998 (approved):

nixpanic (approved):

PR #2443, opened 2021-08-26
deploy: enable podInfoOnMount flag in csidriver object

### LINKED ISSUE #2441: Add more metadata by enabling features on csidriver object

There are some flags we can enable on csiDriver object which could help us to get more information on various requests our driver receive, for example, enable `podinfoonmount` which could help on getting more details like pod  namespace/UID..etc

https://kubernetes-csi.github.io/docs/pod-info.html

This information could be handy while we operate later.  With this issue we try to add this enhancements

humblec:
This flag enables CSI driver to get the pod name, namespace, uid
and service account as part of the nodepublish request.
This come handy while we enhance features like encryption, namespace
transfer...etc. Enabling this flag as part of this patchset. This is
similar to the extra metadata CSI driver receive as part of PVC but
for pod.

Fixes https://github.com/ceph/ceph-csi/issues/2441

### DIFF
--- a/charts/ceph-csi-cephfs/templates/csidriver-crd.yaml
+++ b/charts/ceph-csi-cephfs/templates/csidriver-crd.yaml
@@ -8,4 +8,4 @@ metadata:
   name: {{ .Values.driverName }}
 spec:
   attachRequired: true
-  podInfoOnMount: false
+  podInfoOnMount: true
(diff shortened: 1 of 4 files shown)


Madhu-1 (changes requested):
What is the use-case of passing this information to the nodeplugin?  as we discussed in yesterday's call this is not going to help encryption.  I would like to see a feature that is using this information along with this PR or else simply we are enabling this option but it's of no use until there is a feature to use it.

nixpanic (changes requested):
We need this during `NodeStageVolume`, at the time we fetch the encryption passphrase and call `cryptsetup`. `NodePublishVolume` is too late, it only bind-mounts the attached/decrypted/mounted filesystem in a location where the container can access it.

PR #2445, opened 2021-08-26
rbd: skip volumeattachment processing if pv marked for deletion

humblec:
if the volumeattachment has been fetched but marked for deletion
the nbd healer dont want to process further on this pv. This patch
adds a check for pv is marked for deletion and if so, make the
healer skip processing the same

### DIFF
--- a/internal/rbd/rbd_healer.go
+++ b/internal/rbd/rbd_healer.go
@@ -153,9 +153,8 @@ func runVolumeHealer(ns *NodeServer, conf *util.Config) error {
 
 			continue
 		}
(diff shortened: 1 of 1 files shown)


pkalever (approved):
From k8s doc:

`DeletionTimestamp`:
* This field is set by the server when a graceful deletion is
requested by the user, and is not directly settable by a client.
* If not set, graceful deletion of the object has not been requested.
* Once the deletionTimestamp is set, this value may not be unset or
be set further into the future

The changes LGTM.

nixpanic (approved):

PR #2461, opened 2021-08-31
ci: ignore k8s.io/kubernetes dependencies

nixpanic:
These dependencies are pulled in by k8s.io/kubernetes with version
v0.0.0. It is therefore required to use 'replace' in go.mod to select a
compatible version of the additional k8s.io packages.

Dependabot does not seem to update packages listed in 'replace', only
under 'require'. That means, the version updates done by Dependabot do
not have any effect, as the contents is replaced with a different
version anyway. Ignoring these packages prevents the creation of
non-functional PRs.

Closes: #2459 
Closes: #2458 
Closes: #2456

---

<details>

  progress or results

</details>


nixpanic:
This intentionally does not include `k8s.io/kubernetes` as the PR creates a notification that an update can be done. It is up to a contributor to update the PR, or replace it with an alternative.

.github/dependabot.yml
```diff
@@ -9,3 +9,30 @@ updates:
       - rebase
     commit-message:
       prefix: "rebase"
+    ignore:
+      # k8s.io/kubernetes prevents auto-updating these
+      - dependency-name: "k8s.io/api"
```
Rakshith-R, line 14:
```suggestion
      - dependency-name: "k8s.io/*"
```

This will work too?

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#specifying-dependencies-and-versions-to-ignore

nixpanic: it might work, but we use `k8s.io/klog` and `k8s.io/utils` as well (maybe more?), and those can be updated separately

Rakshith-R: Okay 👍

humblec: kind of agree to both, ideally its good to get all the kube dependencies ( which include klog , utils..etc) fetched in single update ,  so `*` make sense but I am not sure the regular expression works.  We can try this model too.

Rakshith-R (approved):

.github/dependabot.yml
nixpanic: @humblec , `k8s.io/klog` and `k8s.io/utils` are versioned independently of `k8s.io/kubernetes`, so updating them together with `k8s.io/kubernetes` is not particularly  useful.

humblec (approved):

PR #2471, opened 2021-09-01
cephfs: save space in the cluster by avoiding roundoff

humblec:
CephFS controller round off ( by util.RoundOffBytes) on the size
passed in from the provisioner at present, however this can
cause wastage of space in the cluster. The CephFS filesystem is
capable of creating volume with the exact size passed in. so no
need of Roundoff here before we reach out to the cluster

```
$ ceph fs subvolume create <vol_name> <subvol_name> [--size
<size_in_bytes>]...
```

Clone scenario:
    
This shouldnt be an issue as cephfs keep the parent size while the
clone is performed, we also have an extra measure at clone to make
the requested size is achieved by resize if the backend is not
having the requested size. This shouldnt be a case falling into with
this patch though.

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -186,7 +186,12 @@ func (cs *ControllerServer) CreateVolume(
 	defer volOptions.Destroy()
 
 	if req.GetCapacityRange() != nil {
-		volOptions.Size = util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
+		// we were doing round off ( by util.RoundOffBytes) to the size passed
+		// in from the provisioner before Ceph CSI v3.4.0, however this can
+		// cause wastage of space in the cluster. The CephFS filesystem is capable
+		// of creating volume with the exact size passed in. so no need of Roundoff
+		// here before we reach out to the cluster
+		volOptions.Size = req.GetCapacityRange().GetRequiredBytes()
 	}
 
 	parentVol, pvID, sID, err := checkContentSource(ctx, req, cr)


internal/cephfs/controllerserver.go
```diff
@@ -185,7 +185,12 @@ func (cs *ControllerServer) CreateVolume(
 	defer volOptions.Destroy()
 
 	if req.GetCapacityRange() != nil {
-		volOptions.Size = util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
+		// we were doing round off ( by util.RoundOffBytes) to the size passed
+		// in from the provisioner before ceph csi v3.4.0, however this can
```
Madhu-1: `ceph csi` to `Ceph CSI` and after this Fix, if anyone uses an older version of external-provisioner they will face the same issue as before. are we documenting to use provisioner version which does not roundoff anymore?

can you please open an issue with more details about this one like https://github.com/kubernetes-csi/external-provisioner/issues/199?

Madhu-1 (changes requested):
Can you please add an E2E to validate the size like 1.1Mi or 1.2Gib etc?

Same changes are required for cephfs resizing and RBD createVolume and resize also. could be a separate PR's.

internal/cephfs/controllerserver.go
humblec: @Madhu-1 I would rather see this change in this way , regardless of the roundoff or adjustments external provisioner does , from next version of ceph csi onwards,  we try to create the same size of the volume which has been requested by the sidecar request.  On return path again, the roundoff or adjustments done by sidecar or console or some client is again left to them. In short, cephfs driver here act like a dump and serve what has been requested and no internal manipulation in the driver code  in between. That seems like a right approach to me.  wdyt?

internal/cephfs/controllerserver.go
```diff
@@ -185,7 +185,12 @@ func (cs *ControllerServer) CreateVolume(
...
+		// cause wastage of space in the cluster. The CephFS filesystem is capable
+		// of creating volume with the exact size passed in. so no need of Roundoff
+		// here before we reach out to the cluster
+		volOptions.Size = req.GetCapacityRange().GetRequiredBytes()
```
nixpanic, line 194: There might be an issue where a volume was created with a previous version, and a new volume cloned from the original volume with the same size given? If  the original volume's size was rounded up, and the clone request contains a smaller size (not round-up), will that work?

humblec: Shouldnt be an issue as the clone keep the parent size.  I have also explained this in the commit message.

humblec:
removing the milestone tracker from this issue.

Madhu-1:
Hello there! There hasn't been any activity on this PR for quite some time, so I have removed the "keepalive" label. To avoid it being closed as stale shortly, please revisit and give it some attention. Thank you!

PR #2472, opened 2021-09-01
cephfs: no need to check for zero volume size

humblec:
At present there is a 'todo' to check for zero volume size
    in the createVolume request which in unwanted, ie the pvc
    creation with size 0 fail from the kubernetes api validation itself:
    
    For ex:
    
    ```
    ..spec.resources[storage]: Invalid value: "0": must be greater than zero```
    ```
    so we dont need any extra check in the controller server

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Card](https://goreportcard.com/badge/github.com/ceph/ceph-csi)](https://goreport
 - [Ceph CSI](#ceph-csi)
   - [Overview](#overview)
   - [Project status](#project-status)
-  - [Supported CO platforms](#supported-co-platforms)
+  - [Known to work CO platforms](#known-to-work-co-platforms)
   - [Support Matrix](#support-matrix)
     - [Ceph-CSI features and available versions](#ceph-csi-features-and-available-versions)
     - [CSI spec and Kubernetes version compatibility](#csi-spec-and-kubernetes-version-compatibility)
@@ -47,11 +47,18 @@ NOTE:
 
 Status: **GA**
 
-## Supported CO platforms
+## Known to work CO platforms
 
 Ceph CSI drivers are currently developed and tested **exclusively** on Kubernetes
-environments. There is work in progress to make this CO independent and thus
-support other orchestration environments in the future.
+environments.
+
+| Ceph CSI Version | Container Orchestrator Name | Version Tested|
+| -----------------| --------------------------- | --------------|
+| v3.4.0 | Kubernetes | v1.20, v1.21, v1.22|
+| v3.3.0 | Kubernetes | v1.20, v1.21, v1.22|
+
+There is work in progress to make this CO independent and thus
+support other orchestration environments (Nomad, Mesos..etc) in the future.
 
 NOTE:
 
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -188,7 +188,6 @@ func (cs *ControllerServer) CreateVolume(
 	if req.GetCapacityRange() != nil {
 		volOptions.Size = util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
 	}
-	// TODO need to add check for 0 volume size
 
 	parentVol, pvID, sID, err := checkContentSource(ctx, req, cr)
 	if err != nil {


nixpanic (dismissed):
would be good to give a reference to the CSI spec and/or the implementation where the creation of a 0 size volume is prevented.

internal/cephfs/controllerserver.go
```diff
@@ -187,7 +187,6 @@ func (cs *ControllerServer) CreateVolume(
 	if req.GetCapacityRange() != nil {
 		volOptions.Size = util.RoundOffBytes(req.GetCapacityRange().GetRequiredBytes())
 	}
-	// TODO need to add check for 0 volume size
```
Madhu-1, line 191: From CSI spec

```
// This field is OPTIONAL. This allows the CO to specify the capacity
  // requirement of the volume to be provisioned. If not specified, the
  // Plugin MAY choose an implementation-defined capacity range. If
  // specified it MUST always be honored, even when creating volumes
  // from a source; which MAY force some backends to internally extend
  // the volume after creating it.
  CapacityRange capacity_range = 2;
```

are we sure that no CO is implementing the `0` size? because we have some users for kubernetes and Nomad CO

humblec:
> would be good to give a reference to the CSI spec and/or the implementation where the creation of a 0 size volume is prevented.

@nixpanic  Its actually at CO side, the pvc creation fail like this:

```
# kubectl create -f check-zero.yaml 
The PersistentVolumeClaim "rbd-pvc-zero" is invalid: spec.resources[storage]: Invalid value: "0": must be greater than zero```

File: internal/cephfs/controllerserver.go (hunk 1 above)
Comment (nixpanic) [reply internal/cephfs/controllerserver.go:191]: And Mesos too!

Comment (humblec) [reply internal/cephfs/controllerserver.go:191]: @madhu-1 https://github.com/ceph/ceph-csi/pull/2472#issuecomment-910122520 is on Kube CO , reg#` nomad`,  I went through available articles  to an extent,  but was not able to find any particular reference on `zero` sized PVCs.  I am aware from a couple of recently reported nomad issues here in this repo that,  ceph csi is conusumed with Nomad too, eventhough we havent really paid an attention specific to `nomad` CO in many places of the code, and if we start accounting Nomad as a CO ,  Iam in an impression that, many places to revisit in ceph csi and check the different possibilities. with that in mind , do we really need to worry about `nomad` at this stage ? also, do you see any reference form Nomad where they create `0` sized PVCs?  Cc @nixpanic

**Comment by nixpanic** [comment]
Not sure a `kubectl` check is sufficient, it really should be some CSI standard...

File: internal/cephfs/controllerserver.go (hunk 1 above)
Comment (humblec) [reply internal/cephfs/controllerserver.go:191]: Also as an additional reference to CSI spec, see https://github.com/container-storage-interface/spec/issues/488

Comment (Madhu-1) [reply internal/cephfs/controllerserver.go:191]: I really don't know about the nomad but at the same time, I don't think we need to consider removing the TODO unless this is mentioned in the CSI spec.

Comment (humblec) [reply internal/cephfs/controllerserver.go:191]: @nixpanic @madhu-1 one other thought I have here is to mention the CO we support in this project . Considering we are doing or testing only Kube , may be we have to report it in the support matrix too.  once we started to qualify `nomad` or some other `CO` we can update the matrix. Please let me know what do you think , accordingly I can update the support matrix too .

Comment (Madhu-1) [reply internal/cephfs/controllerserver.go:191]: Yes that looks good 👍

Comment (nixpanic) [reply internal/cephfs/controllerserver.go:191]: I would not like to see 'support' for specific COs. Maybe list 'known to work COs' and list the versions we use in CI and known products? That way, users that manage to configure Hashicorp Nomad, Apache Mesos or Docker Swarm could add the versions they use.

Comment (humblec) [reply internal/cephfs/controllerserver.go:191]: yeah, that make sense, let me file a PR with `known to work COs`.  :+1:

Comment (humblec) [reply internal/cephfs/controllerserver.go:191]: @nixpanic @Madhu-1 added `Known to Work` COs and tested matrix. PTAL. Thanks

**Comment by humblec** [comment]
@Madhu-1 @nixpanic ptal.. thanks !

File: README.md [hunk 2]
```diff
@@ -47,11 +47,18 @@ NOTE:
...
 
 Ceph CSI drivers are currently developed and tested **exclusively** on Kubernetes
-environments. There is work in progress to make this CO independent and thus
-support other orchestration environments in the future.
+environments.
+
+| Ceph CSI Version | Container Orchestrator Name | Version Tested|
+| -----------------| --------------------------- | --------------|
+| v3.4.0 | Kubernetes | v1.20, v1.21, v.22|
+| v3.3.0 | Kubernetes | v1.20, v1.21, v.22|
```
nixpanic: should we not list OpenShift too?

nixpanic (dismissed):
The commit still mentions

> creation with size 0 fail from the api validation itself

which might only be correct for Kubernetes, it is better to mention that

README.md
Madhu-1: typo v22 to v1.22

README.md
```diff
@@ -47,11 +47,18 @@ NOTE:
 
 Status: **GA**
 
-## Supported CO platforms
+## Known to work CO platforms
```
Madhu-1, line 50: markdown table at line 11 needs to be correct for this new change

README.md
humblec: @nixpanic actually I thought about it, but considering we dont have a CI for that, I left it out. if we are in agreement to list, we can list that too..

humblec: I think its better to not list in this upstream repo, as we dont have a CI to verify against . @nixpanic @Madhu-1 wdyt?

Madhu-1: Sounds good to me and also along with CI we need to add openshift specific deployment/Readme changes to run cephcsi in openshift also

Madhu-1 (dismissed):

README.md
nixpanic: I think it is fine to mention OpenShift, users should be able to deploy ceph-csi on it. If there are changes needed to the Helm Charts or `deploy/` examples, we should probably document/add that. Not sure what needs changing though, I usually deploy with Rook, and that seems to work fine?

nixpanic (dismissed):
> The commit still mentions
> 
> > creation with size 0 fail from the api validation itself
> 
> which might only be correct for Kubernetes, it is better to mention that

This still needs addressing?

README.md
Madhu-1: Rook has the required SCC changes which are required to run in openshift but we have not documented the changes in cephcsi. IMO we can add OCP to the list once we add the required documentation

humblec: I agree with @Madhu-1 here, compared to Rook repo, we only list kube templates here and cover it in the documentation . Once we have examples and documentation we could add openshift.

Madhu-1 (approved):

humblec:
@nixpanic may be final look ? :)

humblec:
@nixpanic ptal.. thanks !

yati1998 (approved):

PR #2473, opened 2021-09-01
ci: do not run commitlint GitHub Action for dependabot

nixpanic:
The commit messages that Dependabot (@app/dependabot) creates are not
always accepted by the commitlint check. The configuration for
dependabot does not give a lot of options to customize the message, so
instead of adjusting the message to pass commitlint, just skip the check
(Mergify does not require commitlint status either).

See-also: #2460

---

<details>

  progress or results

</details>

### DIFF
--- a/.github/workflows/commitlint.yaml
+++ b/.github/workflows/commitlint.yaml
@@ -8,6 +8,7 @@ on:
 jobs:
   commitlint:
     name: commitlint
+    if: ${{ github.actor != 'dependabot[bot]' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


Madhu-1 (dismissed):

.github/workflows/commitlint.yaml
```diff
@@ -8,6 +8,7 @@ on:
 jobs:
   commitlint:
     name: commitlint
+    if: ${{ github.actor != 'app/dependabot' }}
```
Rakshith-R:
```suggestion
    if: ${{ github.actor != 'dependabot[bot]' }}
```
@nixpanic Can you please add reference to where you found example for `'app/dependabot'`. 
This website mentions it as `'dependabot[bot]'`
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions

Rakshith-R (changes requested):

.github/workflows/commitlint.yaml
nixpanic: I was wondering about that too!

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context mentions for `github.actor`:

> The login of the user that initiated the workflow run.

And that seems to be `app/dependabot` according to the PRs that the bot created.

The link you gave has more examples specifically for dependabot, so I'd trust that more too. Will change it!

Rakshith-R (approved):

Madhu-1 (approved):

humblec (approved):

PR #2476, opened 2021-09-01
util: make ceph.conf as configurable during installation

### LINKED ISSUE #1799: ceph.conf is deleted/changed at cephcsi startup
Looks like the cephcsi plugin is changing the ceph.conf. This is not really desirable, nobody is going to expect that a csi plugin is going to change your /etc/ceph/ceph.conf file. 

Before running the csi plugin
```
[@ceph]# ls -arlt
total 28

-rw-r--r--   1 root root  170 Feb 16  2019 rbdmap
-rw-r--r--   1 root root 6189 Sep 25 09:30 ceph.conf
drwxr-xr-x 110 root root 8192 Dec 16 12:04 ..
drwxr-xr-x   2 root root   89 Dec 16 12:23 .
-rw-r--r--   1 root root    0 Dec 16 17:07 keyring
```
After running the csi plugin with
`/usr/libexec/csi/csiceph -endpoint unix:///tmp/csiceph.sock -nodeid $HOSTNAME -type rbd -drivername ceph-csi-v3.2.0-rbd -alsologtostderr -v 99`

```
[@ceph]# ls -arlt
total 24
-rw-r--r--   1 root root  170 Feb 16  2019 rbdmap
drwxr-xr-x 110 root root 8192 Dec 16 12:04 ..
drwxr-xr-x   2 root root   89 Dec 16 12:23 .
-rw-r--r--   1 root root    0 Dec 16 17:09 keyring
-rw-r--r--   1 root root  182 Dec 16 17:09 ceph.conf
```

### LINKED ISSUE #1815: Support ceph cluster running without cephx

I want to use ceph-csi on cluster running without cephx due to performance reasons.

Disabling cephx can be helpful to squeeze extra performance when doing massive small IO with ceph. Obviously there are security implications for the cluster, and it should only be done in specific scenarios.

Standard features like creating, destroying, mounting, unmounting, etc work.

# Additional context #

Adding the following to ```ceph-csi-rbd-nodeplugin``` daemonset and ```ceph-csi-rbd-provisioner``` deployment seems to make everything work, so this should be simple to support.
```
        lifecycle:
          postStart:
            exec:
              command:
              - sh
              - -c
              - |
                sleep 5 &&
                sed -i 's/cephx/none/' /etc/ceph/ceph.conf &
```

Madhu-1:
Moved the ceph.conf to a new configmap so that users can customize the ceph.conf before/after installation of cephcsi.

fixes: #1815
fixes: #1799

Note:- this is backward compatible with older templates also

up for initial review.

TODO:-

- [x] E2E update

### DIFF
--- a/charts/ceph-csi-cephfs/templates/ceph-conf.yaml
+++ b/charts/ceph-csi-cephfs/templates/ceph-conf.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ceph-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ include "ceph-csi-cephfs.name" . }}
+    chart: {{ include "ceph-csi-cephfs.chart" . }}
+    component: {{ .Values.nodeplugin.name }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  ceph.conf: |
+{{ tpl .Values.cephconf . | indent 4 }}
+  keyring: ""
--- a/charts/ceph-csi-cephfs/values.yaml
+++ b/charts/ceph-csi-cephfs/values.yaml
@@ -275,6 +275,26 @@ secret:
   adminID: <plaintext ID>
   adminKey: <Ceph auth key corresponding to ID above>
 
+# This is a sample configmap that helps define a Ceph configuration as required
+# by the CSI plugins.
+# Sample ceph.conf available at
+# https://github.com/ceph/ceph/blob/master/src/sample.ceph.conf Detailed
+# documentation is available at
+# https://docs.ceph.com/en/latest/rados/configuration/ceph-conf/
+cephconf: |
+  [global]
+    auth_cluster_required = cephx
+    auth_service_required = cephx
+    auth_client_required = cephx
+
+    # Workaround for http://tracker.ceph.com/issues/23446
+    fuse_set_user_groups = false
+
+    # ceph-fuse which uses libfuse2 by default has write buffer size of 2KiB
+    # adding 'fuse_big_writes = true' option by default to override this limit
+    # see https://github.com/ceph/ceph-csi/issues/1928
+    fuse_big_writes = true
+
 #########################################################
 # Variables for 'internal' use please use with caution! #
 #########################################################
--- a/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
+++ b/deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
@@ -143,6 +143,8 @@ spec:
               readOnly: true
             - name: host-dev
               mountPath: /dev
+            - name: ceph-config
+              mountPath: /etc/ceph/
             - name: ceph-csi-config
               mountPath: /etc/ceph-csi-config/
             - name: keys-tmp-dir
@@ -181,6 +183,9 @@ spec:
         - name: host-dev
           hostPath:
             path: /dev
+        - name: ceph-config
+          configMap:
+            name: ceph-config
         - name: ceph-csi-config
           configMap:
             name: ceph-csi-config
--- a/docs/deploy-cephfs.md
+++ b/docs/deploy-cephfs.md
@@ -157,6 +157,12 @@ within the Ceph CSI plugin pods. To add a specific Ceph clusters configuration
 details, refer to [Creating CSI configuration](../examples/README.md#creating-csi-configuration)
 for more information.
 
+**Deploy Ceph configuration ConfigMap for CSI pods:**
+
+```bash
+kubectl create -f ../example/ceph-config.yaml
+```
+
 **Deploy CSI sidecar containers:**
 
 ```bash
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -28,7 +28,7 @@ var (
 	cephfsDeamonSetName   = "csi-cephfsplugin"
 	cephfsContainerName   = "csi-cephfsplugin"
 	cephfsDirPath         = "../deploy/cephfs/kubernetes/"
-	cephfsExamplePath     = "../examples/cephfs/"
+	cephfsExamplePath     = examplePath + "cephfs/"
 	subvolumegroup        = "e2e"
 	fileSystemName        = "myfs"
 )
@@ -74,6 +74,19 @@ func createORDeleteCephfsResources(action kubectlAction) {
 		err = retryKubectlInput(cephCSINamespace, action, string(csiDriver), deployTimeout)
 		if err != nil {
 			e2elog.Failf("failed to %s CSIDriver object with error %v", action, err)
+		}
+	}
+	cephConf, err := ioutil.ReadFile(examplePath + cephConfconfigMap)
+	if err != nil {
+		// createORDeleteCephfsResources is used for upgrade testing as cephConfConfigmap is
+		// newly added, discarding file not found error.
+		if !os.IsNotExist(err) {
+			e2elog.Failf("failed to read content from %s with error %v", examplePath+cephConfconfigMap, err)
+		}
+	} else {
+		err = retryKubectlInput(cephCSINamespace, action, string(cephConf), deployTimeout)
+		if err != nil {
+			e2elog.Failf("failed to %s ceph-conf configmap object with error %v", action, err)
 		}
 	}
 	data, err := replaceNamespaceInTemplate(cephfsDirPath + cephfsProvisioner)
--- a/examples/ceph-conf.yaml
+++ b/examples/ceph-conf.yaml
@@ -1,0 +1,28 @@
+---
+# This is a sample configmap that helps define a Ceph configuration as required
+# by the CSI plugins.
+
+# Sample ceph.conf available at
+# https://github.com/ceph/ceph/blob/master/src/sample.ceph.conf Detailed
+# documentation is available at
+# https://docs.ceph.com/en/latest/rados/configuration/ceph-conf/
+apiVersion: v1
+kind: ConfigMap
+data:
+  ceph.conf: |
+    [global]
+    auth_cluster_required = cephx
+    auth_service_required = cephx
+    auth_client_required = cephx
+
+    # Workaround for http://tracker.ceph.com/issues/23446
+    fuse_set_user_groups = false
+
+    # ceph-fuse which uses libfuse2 by default has write buffer size of 2KiB
+    # adding 'fuse_big_writes = true' option by default to override this limit
+    # see https://github.com/ceph/ceph-csi/issues/1928
+    fuse_big_writes = true
+  # keyring is a required key and its value should be empty
+  keyring: |
+metadata:
+  name: ceph-config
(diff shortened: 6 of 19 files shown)


examples/ceph-conf.yaml
```diff
@@ -0,0 +1,27 @@
...
+    auth_client_required = cephx
+
+    # Workaround for http://tracker.ceph.com/issues/23446
+    fuse_set_user_groups = false
+
+    # ceph-fuse which uses libfuse2 by default has write buffer size of 2KiB
+    # adding 'fuse_big_writes = true' option by default to override this limit
+    # see https://github.com/ceph/ceph-csi/issues/1928
+    fuse_big_writes = true
+  keyring: |-
```
nixpanic: Maybe not include a `keyring`?  It might be sufficient to only map the `/etc/ceph/ceph.conf` file from the ConfigMap, and be able to create additional files through the provisioner/node-plugin?

Consider using https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#add-configmap-data-to-a-specific-path-in-the-volume for that.

deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
```diff
@@ -184,6 +186,9 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: ceph-csi-config
+        - name: ceph-config
+          configMap:
+            name: ceph-config
```
nixpanic, line 188: maybe only include `ceph.conf`, and no other ConfigMap entries?

internal/util/cephconf.go
```diff
@@ -71,7 +76,11 @@ if any ceph commands fails it will log below error message
 */
 // createKeyRingFile creates the keyring files to fix above error message logging.
 func createKeyRingFile() error {
-	_, err := os.Create(keyRing)
+	var err error
+	// create keyring file if it does not exist to support backward compatibility
+	if _, err = os.Stat(keyRing); os.IsNotExist(err) {
+		_, err = os.Create(keyRing)
```
nixpanic, line 82: I am not sure this  works for directories where a ConfigMap is mounted? Should the mapped ConfigMap not be read-only?

charts/ceph-csi-cephfs/templates/ceph-conf.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    app: {{ include "ceph-csi-cephfs.name" . }}
+    chart: {{ include "ceph-csi-cephfs.chart" . }}
+    component: {{ .Values.nodeplugin.name }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  ceph.conf: |-
+{{ tpl .Values.cephconf . | indent 4 }}
+  keyring: ""
```
nixpanic, line 15: drop `keyring` if you can

nixpanic (changes requested):
looks all pretty reasonable, but I hope you can remove the empty `keyring` entries from the ConfigMap

nixpanic:
[Rook provides an 'override' option for `ceph.conf`](https://rook.io/docs/rook/v1.7/ceph-advanced-configuration.html#custom-cephconf-settings), maybe aligning the functionality with that can be considered?

examples/ceph-conf.yaml
Madhu-1: I tried this one when the configmap is mounted the directory becomes readonly and cannot create a keyring file for that reason I added keyring as an entry to the configmap. let me try once again

Madhu-1:
> [Rook provides an 'override' option for `ceph.conf`](https://rook.io/docs/rook/v1.7/ceph-advanced-configuration.html#custom-cephconf-settings), maybe aligning the functionality with that can be considered?

will check this one when integrating this with Rook.

internal/util/cephconf.go
Madhu-1: No this will be only to backward compatibility or else if the configmap contains only ceph.conf and even if the volume is mounted with readOnly:false it fails here.

docs/deploy-cephfs.md
```diff
@@ -48,7 +48,7 @@ that should be resolved in v14.2.3.
...
 | `--nodeid`                | _empty_                     | This node's ID                                                                                                                                                                                                                                                                       |
-| `--type`                  | _empty_                     | Driver type: `[rbd/cephfs]`. If the driver type is set to  `rbd` it will act as a `rbd plugin` or if it's set to `cephfs` will act as a `cephfs plugin`                                                                                                                                        |
+| `--type`                  | _empty_                     | Driver type: `[rbd/cephfs]`. If the driver type is set to  `rbd` it will act as a `rbd plugin` or if it's set to `cephfs` will act as a `cephfs plugin`                                                                                                                              |
```
nixpanic: I wonder what changed on this line...

examples/ceph-conf.yaml
```diff
@@ -0,0 +1,27 @@
...
+    auth_client_required = cephx
+
+    # Workaround for http://tracker.ceph.com/issues/23446
+    fuse_set_user_groups = false
+
+    # ceph-fuse which uses libfuse2 by default has write buffer size of 2KiB
+    # adding 'fuse_big_writes = true' option by default to override this limit
+    # see https://github.com/ceph/ceph-csi/issues/1928
+    fuse_big_writes = true
+  keyring: |
```
nixpanic, line 26: maybe add a comment that `keyring` can (or should?) stay empty so that Ceph-CSI can write to it?

docs/deploy-cephfs.md
Madhu-1: identification correction by vscode, I can remove it.

examples/ceph-conf.yaml
Madhu-1: sure adding it.

nixpanic (dismissed):

charts/ceph-csi-cephfs/values.yaml
```diff
@@ -275,6 +275,26 @@ secret:
...
+# documentation is available at
+# https://docs.ceph.com/en/latest/rados/configuration/ceph-conf/
+cephconf: |
+  [global]
+    auth_cluster_required = cephx
+    auth_service_required = cephx
+    auth_client_required = cephx
+
+    # Workaround for http://tracker.ceph.com/issues/23446
+    fuse_set_user_groups = false
```
humblec, line 291: just wondering this is still valid , the referenced issue has been fixed already ? @kotreshhr any thoughts here?

charts/ceph-csi-cephfs/values.yaml
```diff
@@ -275,6 +275,26 @@ secret:
...
+
+    # ceph-fuse which uses libfuse2 by default has write buffer size of 2KiB
+    # adding 'fuse_big_writes = true' option by default to override this limit
+    # see https://github.com/ceph/ceph-csi/issues/1928
+    fuse_big_writes = true
```
humblec, line 296: we are explicitly setting this mount option while fuse mounter is used via https://github.com/ceph/ceph-csi/pull/1959, so this is not valid either?

humblec (changes requested):

charts/ceph-csi-cephfs/values.yaml
Madhu-1: Not related to this PR? Can be tracked seperately?

e2e/cephfs.go
```diff
@@ -76,6 +76,19 @@ func createORDeleteCephfsResources(action kubectlAction) {
 			e2elog.Failf("failed to %s CSIDriver object with error %v", action, err)
 		}
 	}
+	cephConf, err := ioutil.ReadFile(examplePath + cephConfconfigMap)
+	if err != nil {
+		// createORDeleteCephfsResources is used for upgrade testing as cephConfConfigmap is
+		// newly added, discarding file not found error.
```
humblec, line 82: can we adjust or use a different method for upgrade testing, otherwise we will not be able to get the exact error if the file is really missing in new version ?

internal/util/cephconf.go
```diff
@@ -50,11 +50,16 @@ func createCephConfigRoot() error {
...
 // ceph related CLIs.
 func WriteCephConfig() error {
-	if err := createCephConfigRoot(); err != nil {
+	var err error
+	if err = createCephConfigRoot(); err != nil {
 		return err
 	}
 
-	err := ioutil.WriteFile(CephConfigPath, cephConfig, 0o600)
+	// create config file if it does not exist to support backward compatibility
```
humblec, line 58: may be I am missing something here, isnt a default file exist in the container bundle ? do we need to create it explicitly?

Madhu-1:
@mergifyio rebase

internal/util/cephconf.go
Madhu-1: `// create config file if it does not exist to support backward compatibility` this is to support backward compatibility where user can still use these changes with older images or it will become a breaking change.

e2e/cephfs.go
Madhu-1: I don't want to rework the framework on this PR and this is not something newly introduced this is following the existing model https://github.com/ceph/ceph-csi/blob/devel/e2e/rbd.go#L92-L104. if the file is missing then the pods won't come up and the pod describes will clearly mention that what is the issue.

deploy/cephfs/kubernetes/csi-cephfsplugin-provisioner.yaml
```diff
@@ -184,6 +186,9 @@ spec:
         - name: ceph-csi-config
           configMap:
             name: ceph-csi-config
+        - name: ceph-config
```
humblec, line 186: nit: better to mention this volume before ceph-csi-config to preserve the order..

Madhu-1: @humblec what order can be more specific here?

Madhu-1: you mean in other files the ceph-config is mentioned before ceph-csi-config?

Madhu-1: done

e2e/cephfs.go
humblec: I doubt thats the case, if the pod fails due to missing configuration, its unlikcly the describe list the cause : ), are you sure on that ?

Madhu-1: Yes as the configmap is already mentioned  on the deployment/daemonset yaml, the pod will fail with unmounted volumes

humblec (approved):

e2e/cephfs.go
humblec: I think there is a confusion, I was pointing it out the exact file missing case and not really the volume mounted via manifest. but any way, considering this is in our e2e,  its fine..

yati1998 (approved):
LGTM

Madhu-1:
@mergifyio rebase

Madhu-1:

Failed due to https://github.com/ceph/ceph-csi/issues/2204

PR #2478, opened 2021-09-01
rbd-nbd: provide alternatives to preserve the ceph log files

pkalever:

* **Fix cephLogDir passing to storageclass via helm**

   `cephLogDir`: is a storage class option that is passed to rbd-nbd daemon.
   `cephLogDirHostPath`: is a nodeplugin daemonset level option that helps in using the right host-path while bind-mounting

* **provide alternatives to preserve the ceph log files**
  Currently, we delete the ceph client log file on unmap/detach.

  This patch provides additional alternatives for users who would like to persist the log files.

  **Strategies:**
  
  `remove`: delete log file on unmap/detach
  `compress`: compress the log file to gzip on unmap/detach, in case there exists a `.gz` file from previous map/unmap of the same volume, then override the previous log with new log.
  `preserve`: preserve the log file in text format

  Note that the default strategy will be remove on unmap, and these options can be tweaked from the storage class

  Compression size details example:

  **On Map:** (with debug-rbd=20)
  $ ls -lh
  -rw-r--r-- 1 root root 526K Sep  1 18:15 rbd-nbd-0001-0024-fed5480a-f00f-417a-a51d-31d8a8144c03-0000000000000003-d2e89c87-0b4d-11ec-8ea6-160f128e682d.log

  **On unmap:**
  $ ls -lh
  -rw-r--r-- 1 root root  33K Sep  1 18:15 rbd-nbd-0001-0024-fed5480a-f00f-417a-a51d-31d8a8144c03-0000000000000003-d2e89c87-0b4d-11ec-8ea6-160f128e682d.gz

---

<details>

  progress or results

</details>


internal/util/log/log.go
```diff
@@ -154,3 +159,38 @@ func TraceLog(ctx context.Context, message string, args ...interface{}) {
 		klog.InfoDepth(1, logMessage)
 	}
 }
+
+// GzipLogFile convert log file from text format gzip compressed format.
+func GzipLogFile(pathToFile string) error {
```
nixpanic: Consider placing this in a new file in the same directory.

You also should add a unit-test for new helper functions (in the same commit). Demonstrate that a new file was created and is in gzip format. I wonder if the original file should get removed/truncated?

internal/util/log/log.go
```diff
@@ -154,3 +159,38 @@ func TraceLog(ctx context.Context, message string, args ...interface{}) {
...
+	defer tf.Close() // #nosec:G307, error on close is not critical here
+
+	// Create a Reader and use ReadAll to get all the bytes from the file.
+	reader := bufio.NewReader(tf)
+	content, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return err
+	}
+
+	// Replace txt extension with gz extension.
```
nixpanic: there is no `txt` extension, bur `.log`?

internal/rbd/rbd_attach.go
```diff
@@ -462,9 +464,24 @@ func detachRBDImageOrDeviceSpec(
 	}
 	if dArgs.isNbd && dArgs.logDir != "" {
 		logFile := getCephClientLogFileName(dArgs.volumeID, dArgs.logDir, "rbd-nbd")
-		if err = os.Remove(logFile); err != nil {
-			log.WarningLog(ctx, "failed to remove logfile: %s, error: %v",
-				logFile, err)
+		switch logStrategyLookup[dArgs.logStatergy] {
```
nixpanic: this switch adds quite some complexity to the existing function, move it to a helper that handles the log strategy? Also makes it easier to write a unit-test for that.

docs/rbd-nbd.md
```diff
@@ -38,31 +38,39 @@ a host-path and the default log file path will be
...
 
-In case if you need a customized log path, you should do below:
+- There are different strategies to maintain the logs
+  - `remove`: delete log file on unmap/detach (default behaviour)
+  - `compress`: compress the log file to gzip on unmap/detach
+  - `preserve`: preserve the log file in text format
 
-- Edit the daemonset templates to change the ceph log directory host-path
-  - If you are using helm charts, then you can use key `cephLogDirHostPath`
+  You can tweak the log strategies though `cephLogStatergy` option from the
```
nixpanic: typo in `cephLogStatergy`

I always prefer to have Kubernetes components capitalized as they do in their docs, meaning StorageClass, DaemonSet and so on.

nixpanic (dismissed):
looks like a useful functionality, I hope the e2e tests can be extended to consume this too

unit-tests are missing for new helper functions, please add those

internal/rbd/rbd_attach.go
```diff
@@ -462,9 +464,24 @@ func detachRBDImageOrDeviceSpec(
...
+		case compress:
+			if err = log.GzipLogFile(logFile); err != nil {
```
Madhu-1: the time taken here depends on the file size? if yes can we do it in go routine to avoid affecting the GRPC call performance?

internal/util/log/log.go
```diff
@@ -154,3 +159,38 @@ func TraceLog(ctx context.Context, message string, args ...interface{}) {
...
+func GzipLogFile(pathToFile string) error {
+	tf, err := os.Open(pathToFile) // #nosec:G304, file inclusion via variable.
+	if err != nil {
+		return err
+	}
+	defer tf.Close() // #nosec:G307, error on close is not critical here
+
+	// Create a Reader and use ReadAll to get all the bytes from the file.
+	reader := bufio.NewReader(tf)
+	content, err := ioutil.ReadAll(reader)
```
Madhu-1: use ioutil.ReadFile?

internal/util/log/log.go
pkalever: Yes! The original file will get removed to save the space with the `compress` strategy.

pkalever:
> looks like a useful functionality, I hope the e2e tests can be extended to consume this too

yes, that is the intention behind this PR.

internal/rbd/rbd_attach.go
```diff
@@ -397,6 +399,30 @@ func waitForrbdImage(ctx context.Context, backoff wait.Backoff, volOptions *rbdV
 	return err
 }
 
+// statergicActionOnLogFile act on log file based on cephLogStrategy.
+func statergicActionOnLogFile(ctx context.Context, logStrategy, logFile string) {
```
humblec: @pkalever may be you meant strategicAction.. ?

docs/rbd-nbd.md
```diff
@@ -31,38 +31,46 @@ in the StorageClass.
...
 
-In case if you need a customized log path, you should do below:
+- There are different strategies to maintain the logs
+  - `remove`: delete log file on unmap/detach (default behaviour)
+  - `compress`: compress the log file to gzip on unmap/detach
+  - `preserve`: preserve the log file in text format
 
-- Edit the daemonset templates to change the `cephLogDir`
-  - If you are using helm charts, then you can use key `cephLogDir`
+  You can tweak the log strategies though `cephLogStrategy` option from the
```
humblec: through

internal/rbd/rbd_attach.go
```diff
@@ -397,6 +399,30 @@ func waitForrbdImage(ctx context.Context, backoff wait.Backoff, volOptions *rbdV
...
+	switch logStrategyLookup[logStrategy] {
+	case compress:
+		go func(logFile string) {
+			var err error
+			if err = log.GzipLogFile(logFile); err != nil {
+				log.WarningLog(ctx, "failed to compress logfile: %s, error: %v", logFile, err)
+			}
+			if err = os.Remove(logFile); err != nil {
+				log.WarningLog(ctx, "failed to remove logfile: %s, error: %v", logFile, err)
```
humblec: why compress also delete @pkalever ?

pkalever: The GzipLogFile() file creates a new file which will be a compressed format of the source and it expects the user to delete the original one!

internal/rbd/rbd_attach.go
pkalever: right!

internal/rbd/rbd_attach.go
pkalever: @humblec @nixpanic @Madhu-1  should we remove the original file inside  GzipLogFile() then ?

pkalever: yeah! Moved `os.Remove` part into GzipLogFile(), so that this looks clean and avoid confusion. Thanks!

internal/util/log/log_utils_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestGzipLogFile(t *testing.T) {
+	t.Parallel()
+	logFile, err := ioutil.TempFile("/tmp", "rbd-*.log")
```
nixpanic: instead of hardcoding `/tmp`, consider using [`t.TempDir()`](https://pkg.go.dev/testing#T.TempDir)

internal/rbd/rbd_attach.go
```diff
@@ -397,6 +399,27 @@ func waitForrbdImage(ctx context.Context, backoff wait.Backoff, volOptions *rbdV
 	return err
 }
 
+// strategicActionOnLogFile act on log file based on cephLogStrategy.
+func strategicActionOnLogFile(ctx context.Context, logStrategy, logFile string) {
```
nixpanic: can you add a unit-test for this too?

internal/rbd/rbd_attach.go
```diff
@@ -397,6 +399,27 @@ func waitForrbdImage(ctx context.Context, backoff wait.Backoff, volOptions *rbdV
...
+			}
+		}(logFile)
+	case remove:
+		if err := os.Remove(logFile); err != nil {
+			log.WarningLog(ctx, "failed to remove logfile: %s, error: %v", logFile, err)
+		}
+	case preserve:
+		// do nothing
+	default:
+		log.ErrorLog(ctx, "unknown cephLogStrategy option: '%s', hint: 'remove'|'compress'|'preserve'", logStrategy)
```
nixpanic: instead of quoting `%s`, use `%q` as that creates a quoted string?

internal/rbd/rbd_util.go
```diff
@@ -117,6 +118,20 @@ type rbdImage struct {
 	ioctx *rados.IOContext
 }
 
+type logFileStrategy uint64
```
nixpanic: what is the reason to use an `uint64` for this? it seems simpler to just have `logFileStrategy` as a `string`?

internal/util/log/log_utils_test.go
```diff
@@ -0,0 +1,46 @@
+/*
+Copyright 2019 ceph-csi authors.
```
humblec: 2021?

internal/rbd/rbd_util.go
humblec: yep, I was thinking in the similar line,  so we dont need the enum (https://github.com/ceph/ceph-csi/pull/2478/files#diff-7bf15962b917754fd100b82b2f1d8a11b4ec2dc9885f6c0177adf93ed9d2b613R123)  and also map.

internal/rbd/rbd_util.go
```diff
@@ -2011,3 +2015,25 @@ func getCephClientLogFileName(id, logDir, prefix string) string {
...
 	return fmt.Sprintf("%s/%s-%s.log", logDir, prefix, id)
 }
+
+// strategicActionOnLogFile act on log file based on cephLogStrategy.
+func strategicActionOnLogFile(logStrategy, logFile string) error {
+	var err error
+
+	switch strings.ToLower(logStrategy) {
+	case "compress":
+		if err = log.GzipLogFile(logFile); err != nil {
```
pkalever, line 2035: @Madhu-1 I tried changing this to:
```
go func(logFile string) {                                                                     
        var err error                                                                         
        if err = log.GzipLogFile(logFile); err != nil {                                       
                log.WarningLog(ctx, "failed to compress logfile: %s, error: %v", logFile, err)
        }                                                                                     
}(logFile)                                                                                    
```

But it didn't work! An immediate check for .gz file after calling `strategicActionOnLogFile()` in the unit tests `TestStrategicActionOnLogFile()` return err :-(

Madhu-1: May be we can check with some timeout or with forloop?

examples/rbd/storageclass.yaml
```diff
@@ -73,8 +73,16 @@ parameters:
    # By default host-path /var/log/ceph of node is bind-mounted into
    # csi-rbdplugin pod at /var/log/ceph mount path. See docs/rbd-nbd.md
    # for available configuration options.
+   # This is to configure target bindmount path used inside container.
```
Madhu-1: >This is to configure target bindmount path used inside container

This does not reflect that this is for logging?

add a note this is only for rbd-nbd?

internal/util/log/log_utils.go
```diff
@@ -0,0 +1,54 @@
...
+	gf, err := os.OpenFile(newExt, os.O_CREATE|os.O_RDWR, 0o600) // #nosec:G304, file inclusion via variable.
+	if err != nil {
+		return err
+	}
+	defer gf.Close() // #nosec:G307, error on close is not critical here
+
+	// Write compressed data.
+	w := gzip.NewWriter(gf)
+	defer w.Close()
+	if _, err = w.Write(content); err != nil {
```
Madhu-1, line 45: If write fails. Delete the opened file?

Madhu-1 (commented):
No major comments. unless moving the Gzip to a goroutine to avoid performance issues.

examples/rbd/storageclass.yaml
pkalever: We didn't want this to be used by rbd-nbd only, maybe in the future ceph-fuse can use it too.

Sure I will consider changing this to:
```
   # This is to configure target bindmount path used inside container for ceph clients logging.
```

sounds fine?

internal/rbd/rbd_util.go
pkalever: Fixed it now.

pkalever: Fixed it now.

internal/util/log/log_utils.go
pkalever: Yes!

pkalever (commented):
Thanks @Madhu-1, respinning in a bit.

Madhu-1 (commented):
Please fix CI failures.

pkalever:
> Please fix CI failures.

@Madhu-1 Done now

internal/util/log/log_utils.go
```diff
@@ -0,0 +1,52 @@
...
+		os.Remove(newExt) // #nosec:G104, not important error to handle
+
+		return err
```
Madhu-1: This will result in below case and IMO this should be fine

 if the application is mounted and unmount and log file is generated and preserved and again if the application is mount and unmounted the log file and the preserved file will be overridden and will have only the latest content.

Madhu-1: @pkalever can  you please update the doc for the same?

internal/rbd/rbd_util.go
```diff
@@ -2011,3 +2015,23 @@ func getCephClientLogFileName(id, logDir, prefix string) string {
...
+
+// strategicActionOnLogFile act on log file based on cephLogStrategy.
+func strategicActionOnLogFile(ctx context.Context, logStrategy, logFile string) {
+	var err error
+
+	switch strings.ToLower(logStrategy) {
+	case "compress":
+		if err = log.GzipLogFile(logFile); err != nil {
+			log.ErrorLog(ctx, "failed to compress logfile %q: %v", logFile, err)
+		}
```
Madhu-1, line 2037: can this be moved to a new goroutine. if it takes more time depends on the data of the log file the performance will be affected.

Madhu-1: in other words, IMO strategicActionOnLogFile can be moved to go-routine.

pkalever: This is Awful! Actually, my respin was to address your previous comment on similar lines, and I made the changes locally and tested the unit pieces. It is strange it is not reflecting here. Sorry for jumping the gun and marking your previous comment as resolved :-(

internal/util/log/log_utils.go
pkalever: Yes! Also, I'm adding the access rights as 0644 and opening with O_TRUNC to override the file size.

pkalever:
@Madhu-1 @humblec please reveiw.

pkalever:
@Mergifyio rebase

Madhu-1 (approved):

pkalever:
@Yuggupta27 @Rakshith-R @yati1998 please review.

yati1998 (approved):

Rakshith-R (approved):

Madhu-1:
@mergifyio rebase

PR #2482, opened 2021-09-02
cleanup: introduce populateRbdVol(), to fill rbdVol from stage req

humblec:
At present the nodeStageVolume() handle many logic of filling rbdvol
struct based on the request received and this method is complex to
follow. with this patch, filling or populating rbvol struct has been
segregated and handled hence make the stage functions' job easy.

### DIFF
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -149,28 +149,14 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
-// NodeStageVolume mounts the volume to a staging path on the node.
-// Implementation notes:
-// - stagingTargetPath is the directory passed in the request where the volume needs to be staged
-//   - We stage the volume into a directory, named after the VolumeID inside stagingTargetPath if
-//    it is a file system
-//   - We stage the volume into a file, named after the VolumeID inside stagingTargetPath if it is
-//    a block volume
-// - Order of operation execution: (useful for defer stacking and when Unstaging to ensure steps
-//	are done in reverse, this is done in undoStagingTransaction)
-//   - Stash image metadata under staging path
-//   - Map the image (creates a device)
-//   - Create the staging file/directory under staging path
-//   - Stage the device (mount the device mapped for image)
-// TODO: make this function less complex.
-// nolint:gocyclo,cyclop // reduce complexity
-func (ns *NodeServer) NodeStageVolume(
+// populateRbdVol update the fields in rbdVolume struct based on the request it received.
+func populateRbdVol(
 	ctx context.Context,
-	req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
-	if err := util.ValidateNodeStageVolumeRequest(req); err != nil {
-		return nil, err
-	}
-
+	req *csi.NodeStageVolumeRequest,
+	cr *util.Credentials) (*rbdVolume, error) {
+	var err error
+	var j *journal.Connection
+	volID := req.GetVolumeId()
 	isBlock := req.GetVolumeCapability().GetBlock() != nil
 	disableInUseChecks := false
 	// MULTI_NODE_MULTI_WRITER is supported by default for Block access type volumes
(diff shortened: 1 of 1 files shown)


internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,65 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
...
 
+// populateVolOptions update the fields in rbdVolume struct based on the request it received.
+func (ns *NodeServer) populateVolOptions(
+	ctx context.Context,
+	req *csi.NodeStageVolumeRequest,
+	disableInUseChecks bool,
+	isStaticVol bool,
+	volID string, cr *util.Credentials) (*rbdVolume, error) {
+	var err error
+	volOptions, err := genVolFromVolumeOptions(ctx, req.GetVolumeContext(), req.GetSecrets(), disableInUseChecks)
```
nixpanic: `volOptions` is not really a suitable name better call it `rbdVol` or `rv` like done in other places.

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,65 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
+// populateVolOptions update the fields in rbdVolume struct based on the request it received.
+func (ns *NodeServer) populateVolOptions(
+	ctx context.Context,
+	req *csi.NodeStageVolumeRequest,
+	disableInUseChecks bool,
+	isStaticVol bool,
```
nixpanic: the whole request is passed, is it not possible to detect `isStaticVol` from that?

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,65 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
...
+		err = vi.DecomposeCSIID(volID)
+		if err != nil {
+			err = fmt.Errorf("error decoding volume ID (%s): %w", volID, err)
+
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		j, connErr := volJournal.Connect(volOptions.Monitors, volOptions.RadosNamespace, cr)
+		if connErr != nil {
+			log.ErrorLog(ctx, "failed to establish cluster connection: %v", connErr)
```
nixpanic: consider updating `connErr` just like is done for other error cases

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,65 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
+// populateVolOptions update the fields in rbdVolume struct based on the request it received.
+func (ns *NodeServer) populateVolOptions(
```
nixpanic: Can this function be re-used easier when it isn't a method of `NodeServer`? `ns` does not seem to be used in the function, so tying it to `NodeServer` is sub-optimal.

nixpanic (dismissed):

internal/rbd/nodeserver.go
humblec: Done. I was more inclined to keep it as  a method though.

internal/rbd/nodeserver.go
humblec: done.

internal/rbd/nodeserver.go
humblec: Can be done,  and its done :)

internal/rbd/nodeserver.go
humblec: changed and kept it with rbdvol

humblec:
@nixpanic comments are addressed. PTAL. Thanks !

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,65 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
+// populateVolOptions update the fields in rbdVolume struct based on the request it received.
+func populateRbdVol(
+	ctx context.Context,
+	req *csi.NodeStageVolumeRequest,
+	disableInUseChecks bool,
+	volID string, cr *util.Credentials) (*rbdVolume, error) {
```
Madhu-1: volID is part of req, reusing avoid the number of arguments.

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,65 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
...
+	var err error
+	var j *journal.Connection
+	volOptions, err := genVolFromVolumeOptions(ctx, req.GetVolumeContext(), req.GetSecrets(), disableInUseChecks)
```
Madhu-1: `disableInUseChecks` is used only in genVolFromOptions.  we can move the calculation of `disableInUseChecks` to this function it no need to be an argument.

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,65 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
+// populateVolOptions update the fields in rbdVolume struct based on the request it received.
+func populateRbdVol(
```
Madhu-1, line 153: `populateVolOptions` to `populateRbdVol`? function name and comment is not matching

Madhu-1 (dismissed):

internal/rbd/nodeserver.go
humblec: done..

internal/rbd/nodeserver.go
humblec: Done..

internal/rbd/nodeserver.go
humblec: Corrected. ptal .. thanks @Madhu-1

Madhu-1 (commented):
>internal/rbd/nodeserver.go:251: File is not `gofumpt`-ed (gofumpt)

Need to fix CI

humblec:
> internal/rbd/nodeserver.go

@madhu-1 done.. thanks.

Madhu-1 (dismissed):

humblec:
@Mergifyio rebase

internal/rbd/nodeserver.go
nixpanic: it has not changed, still `volOptions` and not `rbdVol`?

humblec: @nixpanic can you refresh and see ?  it is `rbdVol` https://github.com/ceph/ceph-csi/pull/2482/files#diff-2f84daaad04728739d84bfee31afc54913845fde94c8d612079dc36f45bdb5f6R153

humblec:
@nixpanic  all the comment are addressed, ptal

internal/rbd/nodeserver.go
nixpanic: On line 181 the variable is still confusingly called `volOptions`, not `rv` or `rbdVol`. Refreshing does not seem to be sufficient.

internal/rbd/nodeserver.go
```diff
@@ -280,6 +226,74 @@ func (ns *NodeServer) NodeStageVolume(
...
+
+	isStaticVol := isStaticVolume(req.GetVolumeContext())
+
+	// throw error when imageFeatures parameter is missing or empty
+	// for backward compatibility, ignore error for non-static volumes from older cephcsi version
+	if imageFeatures, ok := req.GetVolumeContext()["imageFeatures"]; checkImageFeatures(imageFeatures, ok, isStaticVol) {
+		return nil, status.Error(codes.InvalidArgument, "missing required parameter imageFeatures")
+	}
+
+	volOptions, err := populateRbdVol(ctx, req, cr)
```
nixpanic: `volOptions` here as well, the object does not contain only options, but has a lot of methods too. It is clearer to rename it to `rv` or `rbdVol` as it is not a volume-options object independently of cephfs/rbd (hopefully we get there is in the future, but we're still far away from that).

humblec: moved the variable to `rv` , ptal.. thanks @nixpanic

internal/rbd/nodeserver.go
humblec: changed those too..

internal/rbd/nodeserver.go
humblec: please wait.. let me rerun the tests.

humblec: should be good now;

nixpanic (approved):
Thanks!

Madhu-1 (approved):

PR #2484, opened 2021-09-02
rbd: use String() method of MirrorImageState in return error

humblec:
MirrorImageState (type C.rbd_mirror_image_state_t) has a string
method which can be used while returning error in the replication
controller. Previously, we were using int return in the error which
is not the proper usage.

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -324,8 +324,7 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 	case librbd.MirrorImageEnabled:
 		return disableVolumeReplication(rbdVol, mirroringInfo, force)
 	default:
-		// TODO: use string instead of int for returning valid error message
-		return nil, status.Errorf(codes.InvalidArgument, "image is in %d Mode", mirroringInfo.State)
+		return nil, status.Errorf(codes.InvalidArgument, "image is in %s Mode", mirroringInfo.State)
 	}
 
 	return &replication.DisableVolumeReplicationResponse{}, nil


internal/rbd/replicationcontrollerserver.go
```diff
@@ -324,8 +324,7 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 	case librbd.MirrorImageEnabled:
 		return disableVolumeReplication(rbdVol, mirroringInfo, force)
 	default:
-		// TODO: use string instead of int for returning valid error message
-		return nil, status.Errorf(codes.InvalidArgument, "image is in %d Mode", mirroringInfo.State)
+		return nil, status.Errorf(codes.InvalidArgument, "image is in %s Mode", mirroringInfo.State.String())
```
Rakshith-R:
```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "image is in %s Mode", mirroringInfo.State)
```
String() func is implicitly called ?

Rakshith-R (dismissed):

internal/rbd/replicationcontrollerserver.go
humblec: true.   no explicit call required..

Rakshith-R (approved):
LGTM, thanks!

Madhu-1 (approved):

yati1998 (approved):

humblec:
@Mergifyio rebase

Rakshith-R:
> @Mergifyio rebase

@humblec , can you please grant permissions to let mergifyio rebase the pr? 
 https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

humblec:
@Mergifyio rebase

PR #2496, opened 2021-09-08
rbd: set vaultAuthNamespace to vaultNamespace if empty

Madhu-1:
When we read the `csi-kms-connection-details` configmap vaultAuthNamespace might not be set when we do the
conversion the vaultAuthNamespace might be set to empty key and this commits check for the empty value of
vaultAuthNamespace and set the vaultAuthNamespace to` vaultNamespace`.

setting empty value for vaultAuthNamespace happened due  to Marshalling at 
https://github.com/ceph/ceph-csi/blob/e99dd3dea4759887a051eda486ef1df5149ad689/internal/kms/vault_tokens.go#L136-L139.

### DIFF
--- a/internal/kms/vault.go
+++ b/internal/kms/vault.go
@@ -192,10 +192,15 @@ func (vc *vaultConnection) initConnection(config map[string]interface{}) error {
 	if errors.Is(err, errConfigOptionInvalid) {
 		return err
 	}
-	vaultAuthNamespace := vaultNamespace // optional, same as vaultNamespace
+	vaultAuthNamespace := ""
 	err = setConfigString(&vaultAuthNamespace, config, "vaultAuthNamespace")
 	if errors.Is(err, errConfigOptionInvalid) {
 		return err
+	}
+	// if the vaultAuthNamespace key is present and value is empty in config, set
+	// the optional vaultNamespace.
+	if vaultAuthNamespace == "" {
+		vaultAuthNamespace = vaultNamespace
 	}
 	// set the option if the value was not invalid
 	if firstInit || !errors.Is(err, errConfigOptionMissing) {


internal/kms/vault.go
```diff
@@ -192,11 +192,16 @@ func (vc *vaultConnection) initConnection(config map[string]interface{}) error {
 	if errors.Is(err, errConfigOptionInvalid) {
 		return err
 	}
-	vaultAuthNamespace := vaultNamespace // optional, same as vaultNamespace
+	vaultAuthNamespace := ""
 	err = setConfigString(&vaultAuthNamespace, config, "vaultAuthNamespace")
 	if errors.Is(err, errConfigOptionInvalid) {
 		return err
 	}
+	// if the vaultAuthNamespace key is present and value is empty in config, set
```
humblec, line 200: bit confused about this, previously we were setting `vaultDefaultNamespace` as `vaultNamespace` and move it as a `default` value for `vaultAuthNamespace` .  if `vaultAuthNamespace` key is present and value is empty, dont we actually have to return in `errConfigOptionInvalid` saying invalid ?

Madhu-1: It gets added due to the Marshalling done at https://github.com/ceph/ceph-csi/blob/devel/internal/kms/vault_tokens.go#L136-L139. returning Error causes an issue to the users who are creating the encrypted PVC without `vaultAuthNamespace`. Currently, we don't check for empty values for the configurations it could be a good improvement if we want to do it. May be as a follow-up PR.

Rakshith-R: okay, this makes it crystal clear, good catch !:sweat_smile:

Rakshith-R: @Madhu-1 , add a line about to the commit msg/ pr description please?  
`Marshalling done at https://github.com/ceph/ceph-csi/blob/devel/internal/kms/vault_tokens.go#L136-L139. `

Madhu-1: Done

Rakshith-R (approved):

humblec:
https://github.com/ceph/ceph-csi/pull/2496#discussion_r704201245. -> yeah, I think we have to improve the config check and error return functions to take care the invalid inputs, otherwise its error prone and could cause various issues like this :+1:

internal/kms/vault.go
humblec: lets get this in.. thanks !

humblec (approved):

Issue #2509, opened 2021-09-15
Tracker issue for intree rbd to ceph csi online migration


humblec:
# Describe the bug #

This issue track the few changes required to support in tree rbd migration to ceph csi ( Project # https://github.com/ceph/ceph-csi/projects/18) , this allows an in tree rbd user to do  online migration without changing any existing templates. This feature is tracked in kubernetes for v1.23 as alpha under https://github.com/kubernetes/enhancements/pull/2963. 

Demo : part 1 shows PVCs created before enabling migration and part2 shows PVCs created after migration. 
https://drive.google.com/file/d/1yfcngCEyrOPDG5tOg7gva0wVJqNELRGc/view

As discussed in CSI calls, currently identified changes are:

- [x] Get clusterid from mon #https://github.com/ceph/ceph-csi/pull/2512
- considering the clusterID field is a required one for csi , but intree SC has monitors field as a required thing we have to figure out the clusterID from the monitors passed in the request.  The node stage and unstage operations should work without any issues as those will be tracked as static volumes for the csi driver.
- [x] Use a volume handle passed in and use it for operations on existing volumes. #https://github.com/ceph/ceph-csi/pull/2523
The volume handle would include the migration version, mon, pool and image which could be used for operations like delete. Once we have these information available with us its well possible to delete the existing in tree PVs. The volume handle can 
The current format of the volume handle looks like this:

```
mig_mons-<monitors-hash>_image-<imagename>_<pool-hash>
```
Admin supposed to create a clusterID based on the monitors hash ( md5sum) in the csi config map and keep the monitors under this configuration before enabling the migration. While CSI driver receive the volume handle it will look at the configmap and figure out the mons to do the operations.  
- [x] use/fill required secret data in the incoming migration secret request #https://github.com/ceph/ceph-csi/pull/2562

The initial plan was to ask the admin to adjust the secret as a pre-upgrade step for migration which avoid this change in the CSI driver, however taking out admin burden is the main ask from kubernetes community so this change is introduced.
"key" field value will be picked up from the migraion secret to "UserKey" field. "adminId" field value will be picked up from the migration secret to "UserID" field.  if `adminId` field is nil or not set, `UserID` field will be filled with default value ie `admin`.The above logic get activated only when the secret is a migration secret, otherwise skipped to the normal workflow as we have today.

ToDo:
- [x] Add design documentation #2609 
- [x] Validate or add NBD case accordingly.  
[...]

PR #2512, opened 2021-09-16
util: get clusterID for the passed in mon string

humblec:
as part of migration support, the clusterID has to be fetched
from passed in mon. Because the intree RBD storage class only
got monitor and not `clusterID` parameter support. However, in
CSI, SC has the `clusterID` parameter support but not mon. Due
to that we have to fetch the clusterID from config file for the
passed in mon and use it in our operations. This adds a helper
function to retrieve clusterID from passed in mon string.

Updates https://github.com/ceph/ceph-csi/issues/2509

### DIFF
--- a/e2e/staticpvc.go
+++ b/e2e/staticpvc.go
@@ -189,6 +189,102 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
 	return err
 }
 
+func validateRBDStaticMigrationPV(f *framework.Framework, appPath string, isBlock bool) error {
+	opt := make(map[string]string)
+	var (
+		rbdImageName = "test-static-pv"
+		pvName       = "pv-name"
+		pvcName      = "pvc-name"
+		namespace    = f.UniqueName
+		// minikube creates default class in cluster, we need to set dummy
+		// storageclass on PV and PVC to avoid storageclass name mismatch
+		sc = "storage-class"
+	)
+
+	c := f.ClientSet
+	mons, err := getMons(rookNamespace, c)
+	if err != nil {
+		return fmt.Errorf("failed to get mons: %w", err)
+	}
+	mon := strings.Join(mons, ",")
+	size := "4Gi"
+	// create rbd image
+	cmd := fmt.Sprintf(
+		"rbd create %s --size=%d --image-feature=layering %s",
+		rbdImageName,
+		4096,
+		rbdOptions(defaultRBDPool))
+
+	_, e, err := execCommandInToolBoxPod(f, cmd, rookNamespace)
+	if err != nil {
+		return err
+	}
+	if e != "" {
+		return fmt.Errorf("failed to create rbd image %s", e)
+	}
+
+	opt["migration"] = "true"
+	opt["monitors"] = mon
+	opt["imageFeatures"] = "layering"
+	opt["pool"] = defaultRBDPool
+	opt["staticVolume"] = strconv.FormatBool(true)
+	opt["imageName"] = rbdImageName
+	pv := getStaticPV(
+		pvName,
+		rbdImageName,
+		size,
+		rbdNodePluginSecretName,
+		cephCSINamespace,
+		sc,
+		"rbd.csi.ceph.com",
+		isBlock,
+		opt)
+
+	_, err = c.CoreV1().PersistentVolumes().Create(context.TODO(), pv, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("PV Create API error: %w", err)
+	}
+
+	pvc := getStaticPVC(pvcName, pvName, size, namespace, sc, isBlock)
+
+	_, err = c.CoreV1().PersistentVolumeClaims(pvc.Namespace).Create(context.TODO(), pvc, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("PVC Create API error: %w", err)
+	}
+	// bind pvc to app
+	app, err := loadApp(appPath)
+	if err != nil {
+		return err
+	}
+
+	app.Namespace = namespace
+	app.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = pvcName
+	err = createApp(f.ClientSet, app, deployTimeout)
+	if err != nil {
+		return err
+	}
+
+	err = deletePod(app.Name, app.Namespace, f.ClientSet, deployTimeout)
+	if err != nil {
+		return err
+	}
+
+	err = c.CoreV1().PersistentVolumeClaims(pvc.Namespace).Delete(context.TODO(), pvc.Name, metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to delete pvc: %w", err)
+	}
+
+	err = c.CoreV1().PersistentVolumes().Delete(context.TODO(), pv.Name, metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to delete pv: %w", err)
+	}
+
+	cmd = fmt.Sprintf("rbd rm %s %s", rbdImageName, rbdOptions(defaultRBDPool))
+	_, _, err = execCommandInToolBoxPod(f, cmd, rookNamespace)
+
+	return err
+}
+
 // nolint:gocyclo,cyclop // reduce complexity
 func validateCephFsStaticPV(f *framework.Framework, appPath, scPath string) error {
 	opt := make(map[string]string)
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -149,6 +149,25 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
+// getClusterIDFromMigrationVolume fills the clusterID for the passed in monitors.
+func getClusterIDFromMigrationVolume(parameters map[string]string) (string, error) {
+	var err error
+	var rclusterID string
+	mons := parameters["monitors"]
+	for _, m := range strings.Split(mons, ",") {
+		rclusterID, err = util.GetClusterIDFromMon(m)
+		if err != nil && !errors.Is(err, util.ErrMissingConfigForMonitor) {
+			return "", err
+		}
+
+		if rclusterID != "" {
+			return rclusterID, nil
+		}
+	}
+
+	return "", err
+}
+
 // populateRbdVol update the fields in rbdVolume struct based on the request it received.
 func populateRbdVol(
 	ctx context.Context,
@@ -188,6 +207,10 @@ func populateRbdVol(
 	// get rbd image name from the volume journal
 	// for static volumes, the image name is actually the volume ID itself
 	if isStaticVol {
+		if req.GetVolumeContext()[intreeMigrationKey] == intreeMigrationLabel {
+			// if migration static volume, use imageName as volID
+			volID = req.GetVolumeContext()["imageName"]
+		}
 		rv.RbdImageName = volID
 	} else {
 		var vi util.CSIIdentifier
@@ -267,6 +290,16 @@ func (ns *NodeServer) NodeStageVolume(
 		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volID)
 	}
 	defer ns.VolumeLocks.Release(volID)
+
+	// Check this is a migration request because in that case, unlike other node stage requests
+	// it will be missing the clusterID, so fill it by fetching it from config file using mon.
+	if req.GetVolumeContext()[intreeMigrationKey] == intreeMigrationLabel && req.VolumeContext[util.ClusterIDKey] == "" {
+		cID, cErr := getClusterIDFromMigrationVolume(req.GetVolumeContext())
+		if cErr != nil {
+			return nil, status.Error(codes.Internal, cErr.Error())
+		}
+		req.VolumeContext[util.ClusterIDKey] = cID
+	}
 
 	stagingParentPath := req.GetStagingTargetPath()
 	stagingTargetPath := stagingParentPath + "/" + volID
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -73,6 +73,10 @@ const (
 	// thick provisioned or thin provisioned.
 	thickProvisionMetaData = "true"
 	thinProvisionMetaData  = "false"
+
+	// these are the migration label key and value for parameters in volume context.
+	intreeMigrationKey   = "migration"
+	intreeMigrationLabel = "true"
 )
 
 // rbdImage contains common attributes and methods for the rbdVolume and
--- a/internal/util/csiconfig.go
+++ b/internal/util/csiconfig.go
@@ -33,7 +33,7 @@ const (
 	CsiConfigFile = "/etc/ceph-csi-config/config.json"
 
 	// ClusterIDKey is the name of the key containing clusterID.
-	clusterIDKey = "clusterID"
+	ClusterIDKey = "clusterID"
 )
 
 // ClusterInfo strongly typed JSON spec for the below JSON structure.
@@ -154,7 +154,7 @@ func GetMonsAndClusterID(ctx context.Context, clusterID string, checkClusterIDMa
 
 // GetClusterID fetches clusterID from given options map.
 func GetClusterID(options map[string]string) (string, error) {
-	clusterID, ok := options[clusterIDKey]
+	clusterID, ok := options[ClusterIDKey]
 	if !ok {
 		return "", ErrClusterIDNotSet
 	}
@@ -168,11 +168,8 @@ func GetClusterID(options map[string]string) (string, error) {
 // else error.
 func GetClusterIDFromMon(mon string) (string, error) {
 	clusterID, err := readClusterInfoWithMon(CsiConfigFile, mon)
-	if err != nil {
-		return "", err
-	}
 
-	return clusterID, nil
+	return clusterID, err
 }
 
 func readClusterInfoWithMon(pathToConfig, mon string) (string, error) {
@@ -193,6 +190,12 @@ func readClusterInfoWithMon(pathToConfig, mon string) (string, error) {
 	}
 
 	for _, cluster := range config {
+		// as the same mons can fall into different clusterIDs with
+		// different radosnamespace configurations, we are bailing out
+		// if radosnamespace configuration is found for this cluster
+		if cluster.RadosNamespace != "" {
+			continue
+		}
 		for _, m := range cluster.Monitors {
 			if m == mon {
 				return cluster.ClusterID, nil
@@ -200,5 +203,5 @@ func readClusterInfoWithMon(pathToConfig, mon string) (string, error) {
 		}
 	}
 
-	return "", fmt.Errorf("missing configuration of cluster ID for mon %q", mon)
+	return "", ErrMissingConfigForMonitor
 }
--- a/internal/util/csiconfig.go
+++ b/internal/util/csiconfig.go
@@ -161,3 +161,44 @@ func GetClusterID(options map[string]string) (string, error) {
 
 	return clusterID, nil
 }
+
+// GetClusterIDFromMon will be called with a mon string to fetch
+// clusterID based on the passed in mon string. If passed in 'mon'
+// string has been found in the config the clusterID is returned,
+// else error.
+func GetClusterIDFromMon(mon string) (string, error) {
+	clusterID, err := readClusterInfoWithMon(CsiConfigFile, mon)
+	if err != nil {
+		return "", err
+	}
+
+	return clusterID, nil
+}
+
+func readClusterInfoWithMon(pathToConfig, mon string) (string, error) {
+	var config []ClusterInfo
+
+	// #nosec
+	content, err := ioutil.ReadFile(pathToConfig)
+	if err != nil {
+		err = fmt.Errorf("error fetching configuration file %q: %w", pathToConfig, err)
+
+		return "", err
+	}
+
+	err = json.Unmarshal(content, &config)
+	if err != nil {
+		return "", fmt.Errorf("unmarshal failed (%w), raw buffer response: %s",
+			err, string(content))
+	}
+
+	for _, cluster := range config {
+		for _, m := range cluster.Monitors {
+			if m == mon {
+				return cluster.ClusterID, nil
+			}
+		}
+	}
+
+	return "", fmt.Errorf("missing configuration of cluster ID for mon %q", mon)
+}
--- a/internal/util/csiconfig_test.go
+++ b/internal/util/csiconfig_test.go
@@ -34,7 +34,7 @@ func cleanupTestData() {
 	os.RemoveAll(basePath)
 }
 
-//  nolint:gocyclo,cyclop // TODO: make this function less complex.
+// nolint:gocyclo,cyclop // TODO: make this function less complex.
 func TestCSIConfig(t *testing.T) {
 	t.Parallel()
 	var err error
--- a/internal/util/csiconfig_test.go
+++ b/internal/util/csiconfig_test.go
@@ -34,7 +34,7 @@ func cleanupTestData() {
 	os.RemoveAll(basePath)
 }
 
-// TODO: make this function less complex.
+//  nolint:gocyclo,cyclop // TODO: make this function less complex.
 func TestCSIConfig(t *testing.T) {
 	t.Parallel()
 	var err error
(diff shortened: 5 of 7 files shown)


Madhu-1 (dismissed):
@humblec why do you need clusterID? without complete implementation, this PR is not complete.
can you please provide more details in the commit message and PR description?

humblec:
> @humblec why do you need clusterID? without complete implementation, this PR is not complete.
> can you please provide more details in the commit message and PR description?

sure, let me add details, its only in issue.. adding it

humblec:
> @humblec why do you need clusterID? without complete implementation, this PR is not complete.
> can you please provide more details in the commit message and PR description?

@Madhu-1 done.. ptal.  thanks ..

Madhu-1 (commented):
@humblec can you please complete the implementation instead of adding the helper function which is not going to get called anywhere in the code. and also if you have the mon IP then why do you need the clusterID and what you are going to do with clusterID?

Rakshith-R (dismissed):
@humblec , I think we can simulate the usage of this function/ online migration in our ci by using static rbd pv/pvc with such a volumeHandle (formed as described in the issue). 

By doing this we will be sure of the usage of util helper and the process of mounting, unmounting works. 

Can you please add such e2e test case[s].  
And add links to online migration docs and steps prescribed from kublet's side too.

edit; clusterID -> volumehandle

humblec:
> @humblec can you please complete the implementation instead of adding the helper function which is not going to get called anywhere in the code. and also if you have the mon IP then why do you need the clusterID and what you are going to do with clusterID?

@Madhu-1 ptal at the latest patchset, the top commit add nodestage request functionality for migrated volumes, the clusterID has to be retrieved from the monitors passed in and use it to get the volumes treated as static volume and rest of the static volume operations go through..

humblec:
> @humblec , I think we can simulate the usage of this function/ online migration in our ci by using static rbd pv/pvc with such a volumeHandle (formed as described in the issue).
> 
> By doing this we will be sure of the usage of util helper and the process of mounting, unmounting works.
> 
> Can you please add such e2e test case[s].
> And add links to online migration docs and steps prescribed from kublet's side too.
> 
> edit; clusterID -> volumehandle

@Rakshith-R ptal : so for migrated volumes, apart from today's static volume treatment, we just need to fill the clusterID from the available monitors, rest of the operations should be pretty straightforward.

internal/rbd/nodeserver.go
```diff
@@ -264,6 +282,17 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 	defer ns.VolumeLocks.Release(volID)
 
+	// Check this is a migration request because in that case, unlike other node stage requests
+	// it will be missing the clusterID, so fill it by fetching it from config file using mon.
+	if req.GetVolumeContext()["migration"] == "true" && req.VolumeContext["clusterID"] == "" {
+		cID, err := fillClusterIDFromMigrationReq(req.GetVolumeContext())
```
Madhu-1: do error checking first and return if there are any errors.

internal/rbd/nodeserver.go
```diff
@@ -264,6 +282,17 @@ func (ns *NodeServer) NodeStageVolume(
...
+		if cID != "" {
+			req.VolumeContext["clusterID"] = cID
```
Madhu-1: i don't think it's a good approach to modify the existing request context. with this design looks like no other option to avoid minimal code changes.

internal/rbd/nodeserver.go
```diff
@@ -188,6 +201,11 @@ func populateRbdVol(
 	// get rbd image name from the volume journal
 	// for static volumes, the image name is actually the volume ID itself
 	if isStaticVol {
+		if req.GetVolumeContext()["migration"] == "true" {
```
Madhu-1: why we are adding both migration and staticVolume to the attributes? this is just to treat rbd images as static one even though its migrated?

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,19 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
+func fillClusterIDFromMigrationReq(parameters map[string]string) (string, error) {
+	var err error
+	rclusterID := ""
+	mons := parameters["monitors"]
+	for _, m := range strings.Split(mons, ",") {
+		rclusterID, err = util.GetClusterIDFromMon(util.CsiConfigFile, m)
```
Madhu-1: please check for specific error messages (const) if the error is as expected then continue or else return the error.

internal/rbd/nodeserver.go
```diff
@@ -188,6 +201,11 @@ func populateRbdVol(
 	// get rbd image name from the volume journal
 	// for static volumes, the image name is actually the volume ID itself
 	if isStaticVol {
+		if req.GetVolumeContext()["migration"] == "true" {
+			// if migration static volume, use imageName as volID
+			rv.RbdImageName = req.GetVolumeContext()["imageName"]
+			volID = rv.RbdImageName
```
Madhu-1: why not send volID as the rbdImage name itself from the kubelet instead of doing a hack here? are you planning to add support for image deletion also?

internal/rbd/nodeserver.go
```diff
@@ -188,6 +201,11 @@ func populateRbdVol(
 	// get rbd image name from the volume journal
 	// for static volumes, the image name is actually the volume ID itself
 	if isStaticVol {
+		if req.GetVolumeContext()["migration"] == "true" {
+			// if migration static volume, use imageName as volID
+			rv.RbdImageName = req.GetVolumeContext()["imageName"]
```
Madhu-1: this looks redundant to me.

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,19 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
+func fillClusterIDFromMigrationReq(parameters map[string]string) (string, error) {
```
Madhu-1: this can cause issues if the radosNamespace configuration is present for this clusterID.

internal/rbd/nodeserver.go
Madhu-1: use predefined clusterID const from `csiConfig.go`

Madhu-1 (dismissed):
@humblec I would like to see below things to make it complete

* Fix CI issue
* Add documentation for this one
* Are you planning to add the DeletionVolume option also?
* Add E2E test cases to cover this migration case with static PVC (krbd and nbd)

Madhu-1:
@pkalever can you please take a look at it from an nbd point of view?

internal/rbd/nodeserver.go
humblec: sure, done. thanks.

internal/rbd/nodeserver.go
humblec: sure, done..

internal/rbd/nodeserver.go
humblec: yep, an extra marker to make its a migrated static volume.

internal/rbd/nodeserver.go
humblec: actually any error or iow, except `nil` for error is a signal to bail out , so avoiding an extra check for a specific error here.

internal/rbd/nodeserver.go
humblec: Yep, exactly , thats the plan to add deletion but only for migration static volume and not for other static volume (atleast in first version) :+1:

internal/rbd/nodeserver.go
humblec: the rv struct should have it , but kind of double confirmaion

humblec:
> * Fix CI issue 
>
 Done
> * Add documentation for this one
>
Will cover it in a follow up PR , adding to todo tracker
> * Are you planning to add the DeletionVolume option also?:
>
Yes, will cover in a follow up PR 
> * Add E2E test cases to cover this migration case with static PVC (krbd and nbd)
>
 Done.

@Madhu-1 @Rakshith-R all comments are addressed , ptal.. Thanks.

humblec:
```
Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute
```

e2e/staticpvc.go
```diff
@@ -189,6 +189,109 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
...
+	opt["monitors"] = mon
+	if !checkImgFeat {
+		opt["imageFeatures"] = "layering"
+	}
+	opt["pool"] = defaultRBDPool
+	opt["staticVolume"] = strconv.FormatBool(true)
+	opt["imageName"] = rbdImageName
+	if radosNamespace != "" {
+		opt["radosNamespace"] = radosNamespace
+	}
```
Madhu-1: Does kubernetes support create rbd image with radosNamespace?

e2e/staticpvc.go
```diff
@@ -189,6 +189,109 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
...
+	if err != nil {
+		return err
+	}
+	if e != "" {
+		return fmt.Errorf("failed to create rbd image %s", e)
+	}
+
+	opt["migration"] = "true"
+	opt["monitors"] = mon
+	if !checkImgFeat {
```
Madhu-1: This value is always false do we need this one?

e2e/staticpvc.go
```diff
@@ -189,6 +189,109 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
...
+	c := f.ClientSet
+	mon := "rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789"
+
+	size := "4Gi"
+	// create rbd image
+	cmd := fmt.Sprintf(
+		"rbd create %s --size=%d --image-feature=layering %s",
+		rbdImageName,
+		4096,
+		rbdOptions(defaultRBDPool))
```
Madhu-1, line 216: kubernetes supports v1 and v2 images you might need to add tests for both?

e2e/staticpvc.go
```diff
@@ -189,6 +189,109 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
...
+		pvName       = "pv-name"
+		pvcName      = "pvc-name"
+		namespace    = f.UniqueName
+		// minikube creates default class in cluster, we need to set dummy
+		// storageclass on PV and PVC to avoid storageclass name mismatch
+		sc = "storage-class"
+	)
+
+	c := f.ClientSet
+	mon := "rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789"
```
Madhu-1: please retrieve mon from the ceph cluster instead of hardcoding it.

e2e/staticpvc.go
```diff
@@ -189,6 +189,109 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
...
+	}
+
+	opt["migration"] = "true"
+	opt["monitors"] = mon
+	if !checkImgFeat {
+		opt["imageFeatures"] = "layering"
+	}
+	opt["pool"] = defaultRBDPool
+	opt["staticVolume"] = strconv.FormatBool(true)
+	opt["imageName"] = rbdImageName
```
Madhu-1, line 231: this test cover krbd. kubernetes also supports nbd driver? are we not supporting it?

internal/rbd/nodeserver.go
Madhu-1: this comment is not addressed and the function name can be `GetClusterIDFromMigrationVolume`

internal/rbd/nodeserver.go
Madhu-1: If there is an error in reading the configuration file when we need to continue in the for loop?

internal/rbd/nodeserver.go
Madhu-1: But do we need this double-check here? IMO it not needed

Madhu-1 (dismissed):

internal/rbd/nodeserver.go
humblec: shouldnt be an issue in this case.

e2e/staticpvc.go
humblec: dont see that  option, so will remove this.

internal/rbd/nodeserver.go
Madhu-1: @humblec why it's not an issue can you please explain?

e2e/staticpvc.go
humblec: just kept it for parity with other static test

e2e/staticpvc.go
humblec: atleast not required now, can revisit later if really needed.

e2e/staticpvc.go
humblec: I actually did that yeterday but it didnt go through as mon was referred like this. regardless this is not a concern as its admin's responsibility to make sure the mapping exist. also unit tests are also in place to prove that things are working pretty good.

e2e/staticpvc.go
humblec: considering both are in alpha, I will add nbd case as todo for now.

e2e/staticpvc.go
humblec: regardless, also adding a TODO comment for later.

e2e/staticpvc.go
humblec: Done., added in todo

internal/rbd/nodeserver.go
humblec: @Madhu-1 radosnamespace wont be passed in this migration case.

e2e/staticpvc.go
Rakshith-R: https://github.com/ceph/ceph-csi/blob/22bb31df196ba72cb967a9f9d98e6e277315aa6f/e2e/utils.go#L69 

Mons can be fetched using this helper ?
(even though i think the service will lead to the same ip format each time.)

internal/util/csiconfig.go
```diff
@@ -154,10 +154,55 @@ func GetMonsAndClusterID(ctx context.Context, clusterID string, checkClusterIDMa
...
 
 	return clusterID, nil
 }
+
+// GetClusterIDFromMon will be called with a mon string to fetch
+// clusterID based on the passed in mon string. If passed in 'mon'
+// string has been found in the config the clusterID is returned,
+// else error.
+func GetClusterIDFromMon(pathToConfig, mon string) (string, error) {
+	clusterID, err := readClusterInfoWithMon(pathToConfig, mon)
```
Rakshith-R:
```suggestion
func GetClusterIDFromMon(mon string) (string, error) {
	clusterID, err := readClusterInfoWithMon(CsiConfigFile, mon)
```

internal/rbd/nodeserver.go
Rakshith-R:
```suggestion
		rclusterID, err = util.GetClusterIDFromMon(m)
```

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,21 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 	return nil
 }
 
+// fillClusterIDFromMigrationReq fills the clusterID for the passed in monitors.
```
Rakshith-R: Move this function to util, as it not rbd specific ?

internal/util/csiconfig.go
```diff
@@ -154,10 +154,55 @@ func GetMonsAndClusterID(ctx context.Context, clusterID string, checkClusterIDMa
...
+	if err != nil {
+		return "", err
+	}
+
+	if clusterID == "" {
```
Rakshith-R: move this check to `readClusterInfoWithMon`, 

that way unit testcases for `readClusterInfoWithMon` will be sufficient.

Rakshith-R (dismissed):

internal/rbd/nodeserver.go
humblec: the migration is only applicable for rbd now, so its fine to keep it here.

e2e/staticpvc.go
```diff
@@ -189,6 +189,109 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
 	return err
 }
 
+func validateRBDStaticMigrationPV(f *framework.Framework, appPath string, isBlock, checkImgFeat bool) error {
```
pkalever: @humblec can we also consider improving this to take `mounter-type` parameter and act accordingly?
if so, it will be fairly simple to call this with krbd and rbd-nbd mounters from e2e/rbd.go.

Thanks!

internal/util/csiconfig.go
humblec: done..

internal/util/csiconfig.go
humblec: Done

e2e/staticpvc.go
humblec: @pkalever yeah, can be done, will keep it for todo. thanks :+1:

internal/rbd/nodeserver.go
humblec: imo, it looks like its not populated at this stage, I could be wrong, but making sure or assuring should not be of no harm as well.

e2e/staticpvc.go
pkalever: [Just repeating here for context]
If the tests are not part of the same #PR, then probably it means we cannot validate `get clusterID` for the NBD case.
if that is fine with others, then we can create an issue (or add to some tracker) for persistence (set bot rule to keep-alive) and move on.

e2e/staticpvc.go
Madhu-1: This will fail if someone runs  with  different mons, suggest to use the already existing helper function

e2e/staticpvc.go
Madhu-1: This code will never get hit then why do we need it?

internal/rbd/nodeserver.go
Madhu-1: even if you don't pass the radosnamespace will be taken from the existing configmap and https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/rbd_util.go#L1178-L1181 this is going to get called and things will fail and if you have multiple clusterID with the same mons again not sure things will work as each clusterID's can/might contain different radosNamespaces

internal/rbd/nodeserver.go
Madhu-1: you are doing 
```
rv.RbdImageName = req.GetVolumeContext()["imageName"]
volID = rv.RbdImageName
```
inside the if check and again after if check there is code to assign `volID` to ` rv.RbdImageName`

this can be `volID=req.GetVolumeContext()["imageName"]` You don't need to assign it the imageName and volID and again assign volID to imageName

e2e/staticpvc.go
humblec: yeah, will try to figure it out later as its not a priority one atleast for now.

internal/rbd/nodeserver.go
```diff
@@ -149,6 +149,27 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
...
+func fillClusterIDFromMigrationReq(parameters map[string]string) (string, error) {
+	var err error
+	var rclusterID string
+	mons := parameters["monitors"]
+	for _, m := range strings.Split(mons, ",") {
+		rclusterID, err = util.GetClusterIDFromMon(m)
+		if err != nil {
+			if (strings.Contains(err.Error(), util.ErrConfFileFetchFailed)) ||
+				(strings.Contains(err.Error(), util.ErrUnmarshalFailed)) {
+				return "", err
```
Madhu-1: you can check ` "missing configuration of cluster ID for mon %q", mon`(deing const in errors.go) and return error 
```suggestion
if err != nil && !errors.Is(err, util.ErrorMissingConfiguration){
return "",err
}
```

Madhu-1 (dismissed):

Rakshith-R (commented):
LGTM, I'll let madhu finish his review.

humblec:
> LGTM, I'll let madhu finish his review.

Thanks.. adding the last 2 comments from madhu. :+1:

e2e/staticpvc.go
Madhu-1: but why merge/push a code that will break if someone runs tests locally?

humblec: @Madhu-1 addressed it.. thanks.

humblec:
@Madhu-1 @Rakshith-R all  the comments are addressed, ptal. thanks!

internal/rbd/nodeserver.go
humblec: Done.. @Madhu-1

internal/rbd/nodeserver.go
Madhu-1: @humblec is this not addressed yet, function name need to be changed

internal/util/csiconfig_test.go
```diff
@@ -34,7 +34,7 @@ func cleanupTestData() {
 	os.RemoveAll(basePath)
 }
 
-// TODO: make this function less complex.
+// nolint:gocyclo,cyclop // reduce complexity
```
Madhu-1: Please bring back the ToDo which can be used for tracking the todos in project

internal/util/errors.go
```diff
@@ -35,6 +35,8 @@ var (
 	ErrPoolNotFound = errors.New("pool not found")
 	// ErrClusterIDNotSet is returned when cluster id is not set.
 	ErrClusterIDNotSet = errors.New("clusterID must be set")
+	// ErrMissingConfigForMon is returned when clusterID is not found for the mon.
+	ErrMissingConfigForMon = errors.New("missing configuration of cluster ID for mon")
```
Madhu-1: For monitor

internal/util/csiconfig_test.go
```diff
@@ -132,4 +132,29 @@ func TestCSIConfig(t *testing.T) {
 	if err != nil || content != "mon4,mon5,mon6" {
 		t.Errorf("Failed: want (%s), got (%s) (%v)", "mon4,mon5,mon6", content, err)
 	}
+
+	data = "[{\"clusterID\":\"" + clusterID2 + "\",\"monitors\":[\"mon1\",\"mon2\",\"mon3\"]}," +
+		"{\"clusterID\":\"" + clusterID1 + "\",\"monitors\":[\"mon4\",\"mon5\",\"mon6\"]}]"
```
Madhu-1, line 137: Add a test case for radosnamespace

internal/rbd/nodeserver.go
humblec: Done.

internal/util/csiconfig_test.go
humblec: done.

humblec:
@Madhu-1 @Rakshith-R pending or all the comments are addressed. ptal.

humblec:
@Madhu-1 @Rakshith-R can you please take a final look at this PR ?

e2e/staticpvc.go
```diff
@@ -189,6 +189,106 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
...
+	// bind pvc to app
+	app, err := loadApp(appPath)
+	if err != nil {
+		return err
+	}
+
+	app.Namespace = namespace
+	app.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = pvcName
+	if checkImgFeat {
+		err = createAppErr(f.ClientSet, app, deployTimeout, "missing required parameter imageFeatures")
```
Madhu-1: this part of the code will never get hit as you are passing `checkImgFeat` as always false. please remove `checkImgFeat` argument from the function

internal/rbd/rbd_util.go
```diff
@@ -72,6 +72,10 @@ const (
 	// thick provisioned or thin provisioned.
 	thickProvisionMetaData = "true"
 	thinProvisionMetaData  = "false"
+
+	// these are the migration label key and value for params in volume context.
```
Madhu-1: ` params in volume context` to `parameters in volume context`

internal/util/csiconfig.go
```diff
@@ -154,10 +154,54 @@ func GetMonsAndClusterID(ctx context.Context, clusterID string, checkClusterIDMa
...
+	}
+
+	err = json.Unmarshal(content, &config)
+	if err != nil {
+		return "", fmt.Errorf("unmarshal failed (%w), raw buffer response: %s",
+			err, string(content))
+	}
+
+	for _, cluster := range config {
+		if cluster.RadosNamespace != "" {
```
Madhu-1, line 196: add  a comment why we are continuing in this case.

internal/util/errors.go
```diff
@@ -35,6 +35,8 @@ var (
 	ErrPoolNotFound = errors.New("pool not found")
 	// ErrClusterIDNotSet is returned when cluster id is not set.
 	ErrClusterIDNotSet = errors.New("clusterID must be set")
+	// ErrMissingConfigForMonitor is returned when clusterID is not found for the mon.
+	ErrMissingConfigForMonitor = errors.New("missing configuration of cluster ID for mon")
```
Madhu-1: `missing configuration of cluster ID for mon` to `missing configuration of a cluster ID for monitor`

Madhu-1 (commented):
LGTM. small nits

internal/util/errors.go
humblec: Done.. thanks ..

internal/util/csiconfig.go
humblec: sure, done .. thanks..

internal/rbd/rbd_util.go
humblec: done.. thanks..

e2e/staticpvc.go
humblec: done. thanks..

e2e/staticpvc.go
```diff
@@ -189,6 +189,102 @@ func validateRBDStaticPV(f *framework.Framework, appPath string, isBlock, checkI
 	return err
 }
 
+func validateRBDStaticMigrationPV(f *framework.Framework, appPath string, isBlock, checkImgFeat bool) error {
```
Madhu-1: `checkImgFeat` not used in this function, needs to be removed.

humblec: :+1: , done.. thanks

internal/util/csiconfig.go
```diff
@@ -154,10 +154,57 @@ func GetMonsAndClusterID(ctx context.Context, clusterID string, checkClusterIDMa
...
+// clusterID based on the passed in mon string. If passed in 'mon'
+// string has been found in the config the clusterID is returned,
+// else error.
+func GetClusterIDFromMon(mon string) (string, error) {
+	clusterID, err := readClusterInfoWithMon(CsiConfigFile, mon)
+	if err != nil {
+		return "", err
+	}
+
+	return clusterID, nil
```
Rakshith-R:
```suggestion

	return clusterID, err
```

Rakshith-R (dismissed):
@humblec , it looks good to me
just a single small change to remove redundant check.

internal/util/csiconfig.go
humblec: Done.

humblec: @Rakshith-R thanks.

Madhu-1 (approved):

Rakshith-R (approved):

humblec:
@Mergifyio rebase

PR #2514, opened 2021-09-16
rbd: added RBD features support for krbd

### LINKED ISSUE #2513: Support for all Ceph RBD features
Currently RBD csi isn't support all possible image features, but features support by kernel

```console
[root@k8s-worker-node0]# rbd --id kubernetes create kubernetes_rbd/kubernetes-dynamic-pvc-test --size=10G
[root@k8s-worker-node0]# rbd --id kubernetes info kubernetes_rbd/kubernetes-dynamic-pvc-test
rbd image 'kubernetes-dynamic-pvc-test':
        size 10 GiB in 2560 objects
        order 22 (4 MiB objects)
        snapshot_count: 0
        id: 58201ff55f0060
        block_name_prefix: rbd_data.58201ff55f0060
        format: 2
        features: layering, exclusive-lock, object-map, fast-diff, deep-flatten
        op_features:
        flags:
        create_timestamp: Thu Sep 16 16:50:08 2021
        access_timestamp: Thu Sep 16 16:50:08 2021
        modify_timestamp: Thu Sep 16 16:50:08 2021
[root@k8s-worker-node0]# rbd --id kubernetes map kubernetes_rbd/kubernetes-dynamic-pvc-test
/dev/rbd3
[root@k8s-worker-node0]# rbd --id kubernetes showmapped
id  pool            namespace  image                                                        snap  device
0   kubernetes_rbd             kubernetes-dynamic-pvc-0c5c7599-d9c3-11eb-bf5b-aeada6b0a7bc  -     /dev/rbd0
1   kubernetes_rbd             kubernetes-dynamic-pvc-95b13e19-6a15-11eb-86da-8efbeed4ee16  -     /dev/rbd1
2   kubernetes_rbd             kubernetes-dynamic-pvc-9d6d76f5-a822-11eb-9646-82e54736c1fe  -     /dev/rbd2
3   kubernetes_rbd             kubernetes-dynamic-pvc-test                                  -     /dev/rbd3
[root@k8s-worker-node0]# dmesg -TL | grep rbd | grep features | tail -n3
[Fri Aug 27 14:17:35 2021] rbd: rbd0: capacity 32212254720 features 0x1
[Wed Sep  8 17:15:29 2021] rbd: rbd2: capacity 5368709120 features 0x1
[Thu Sep 16 16:51:45 2021] rbd: rbd3: capacity 10737418240 features 0x3d
```

So we need allow for `storageClass.imageFeatures`: `layering,object-map,fast-diff,deep-flatten,exclusive-lock`

k0ste:
Added support for `object-map, fast-diff`

This PR added support for all currently supported [krbd](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/block/rbd.c#n112) image features:

```
#define RBD_FEATURES_ALL	(RBD_FEATURE_LAYERING |		\
				 RBD_FEATURE_STRIPINGV2 |	\
				 RBD_FEATURE_EXCLUSIVE_LOCK |	\
				 RBD_FEATURE_OBJECT_MAP |	\
				 RBD_FEATURE_FAST_DIFF |	\
				 RBD_FEATURE_DEEP_FLATTEN |	\
				 RBD_FEATURE_DATA_POOL |	\
				 RBD_FEATURE_OPERATIONS)

/* Features supported by this (client software) implementation. */

#define RBD_FEATURES_SUPPORTED	(RBD_FEATURES_ALL)
```

The features is described in [man rbd](https://docs.ceph.com/en/latest/man/8/rbd/#cmdoption-rbd-image-feature)

```
Specifies which RBD format 2 feature should be enabled when creating an image. Multiple features can be enabled by repeating this option multiple times. The following features are supported:

· layering: layering support
· striping: striping v2 support
· exclusive-lock: exclusive locking support
· object-map: object map support (requires exclusive-lock)
· fast-diff: fast diff calculations (requires object-map)
· deep-flatten: snapshot flatten support
· journaling: journaled IO support (requires exclusive-lock)
· data-pool: erasure coded pool support
```

## Related issues ##

Fixes: https://github.com/ceph/ceph-csi/issues/2513


examples/rbd/storageclass.yaml
```diff
@@ -33,10 +33,11 @@ parameters:
...
    # (required) RBD image features, CSI creates image with image-format 2
-   # CSI RBD currently supports `layering`, `journaling`, `exclusive-lock`
-   # features. If `journaling` is enabled, must enable `exclusive-lock` too.
-   # imageFeatures: layering,journaling,exclusive-lock
-   imageFeatures: layering
+   # CSI RBD currently supports `layering`, `journaling`, `exclusive-lock`,
+   # `object-map`, `fast-diff` features. If `journaling` is enabled, must
+   # enable `exclusive-lock` too.
+   # imageFeatures: layering,journaling,exclusive-lock,object-map,fast-diff
+   imageFeatures: "layering"
```
pkalever, line 41: We might want to do changes in the helm charts too?

internal/rbd/rbd_util_test.go
```diff
@@ -86,33 +86,41 @@ func TestValidateImageFeatures(t *testing.T) {
...
 			},
 			true,
-			"feature exclusive-lock requires rbd-nbd for mounter",
+			"feature fast-diff requires object-map to be set",
 		},
 		{
 			"layering,exclusive-loc,journaling",
 			&rbdVolume{
-				Mounter: rbdNbdMounter,
+				Mounter: rbdDefaultMounter,
```
Madhu-1, line 126: why this change is made? this is a negative case mounter should not matter.

internal/rbd/rbd_util_test.go
```diff
@@ -86,33 +86,41 @@ func TestValidateImageFeatures(t *testing.T) {
...
+		{
+			"layering,journaling",
+			&rbdVolume{
+				Mounter: rbdDefaultMounter,
+			},
 			true,
 			"feature journaling requires exclusive-lock to be set",
 		},
 		{
-			"layering,exclusive-lock,journaling",
```
Madhu-1, line 100: What is the reason for modifying the existing tests not adding new ones?

internal/rbd/rbd_util.go
```diff
@@ -174,7 +174,18 @@ var supportedFeatures = map[string]imageFeature{
 		needRbdNbd: false,
 	},
 	librbd.FeatureNameExclusiveLock: {
-		needRbdNbd: true,
+		needRbdNbd: false,
```
Madhu-1: reason for removing nbd here?

pkalever (changes requested):
It will be great if we can add e2e tests for each of these features with krbd and rbd-nbd mounters?

Madhu-1 (changes requested):
@k0ste Thanks for the contribution please take care of the below items

* Update the existing documentations
* Update storageclass in  helm charts
* Add E2E for CreateVolume, CreateSnapshot, PVC-PVC Clone, PVC restore, etc for new image features  added
*

internal/rbd/rbd_util.go
k0ste: The reason is:

```
  Warning  ProvisioningFailed    8s (x8 over 70s)   rbd.csi.ceph.com_ceph-csi-rbd-provisioner-6b9f54d9df-4slxv_a61dd25a-888b-4104-8e9b-e5ec09125b38  failed to provision volume with StorageClass "csi-rbd-sc": rpc error: code = InvalidArgument desc = feature exclusive-lock requires rbd-nbd for mount
```

In ancient times this feature was supported only by librbd, but currently support with krbd

k0ste:
> * Update the existing documentations

Updated
> * Update storageclass in  helm charts

Updated
> * Add E2E for CreateVolume, CreateSnapshot, PVC-PVC Clone, PVC restore, etc for new image features  added

Test added

e2e/rbd.go
```diff
@@ -1051,6 +1051,40 @@ var _ = Describe("RBD", func() {
...
+					if err != nil {
+						e2elog.Failf("failed to validate pvc and application binding with error %v", err)
+					}
+					err = deleteResource(rbdExamplePath + "storageclass.yaml")
+					if err != nil {
+						e2elog.Failf("failed to delete storageclass with error %v", err)
+					}
+					err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
+					if err != nil {
+						e2elog.Failf("failed to create storageclass with error %v", err)
```
Madhu-1: Please add snapshot, create PVC from the snapshot, and PVC-PVC clone test also.

Madhu-1 (commented):
@k0ste CI is failing can you please check

tomiles:
@k0ste Thanks for this PR, looking forward to this merging. We've recently hit this issue too of using the features already  supported in the kernel but not being allowed by ceph-csi.

pkalever:
Honestly! I'm not quite convinced with the approach that we are taking in using various features of krbd. The feature set will be dependent on the kernel version (rbd driver loaded).

We should check for krbd supported features first and if the feature is available use krbd else we should fall back to rbd-nbd.

tomiles:
I agree that detecting it would be even better. But without this PR we are just blocking features supported in most of the current LTS distros kernel versions which also isn’t a good approach.

k0ste:
> We should check for krbd supported features first and if the feature is available use krbd else we should fall back to rbd-nbd

I think better is fail, because krbd & rbd-nbd performance is dramatically different. Admin should get error and make a self decision - use nbd or update kernel

pkalever:
> I think better is fail, because krbd & rbd-nbd performance is dramatically different. Admin should get error and make a self decision - use nbd or update kernel

Do you have any benchmarking results comparing both?
Right, architecture wise it makes sense to agree that the nbd performance would be a bit low, but I don't think they should be too bad or incomparable with krbd.

pkalever:
FYI: for detecting the runtime krbd features supported, here is the PR: https://github.com/ceph/ceph-csi/pull/2556

humblec:
> > We should check for krbd supported features first and if the feature is available use krbd else we should fall back to rbd-nbd
> 
> I think better is fail, because krbd & rbd-nbd performance is dramatically different. Admin should get error and make a self decision - use nbd or update kernel

@k0ste agreed, however if admin is flexible to fallback, may be the SC option ( `fallbackNbd`)  as in #2556 helps? just checking what is the normal practice/desire from admins  in different setups in these kind of scenarios .

e2e/rbd.go
```diff
@@ -1479,6 +1514,7 @@ var _ = Describe("RBD", func() {
 					f)
 			})
 
+
```
nixpanic: no need to add the extra line here

e2e/rbd.go
```diff
@@ -1515,6 +1551,43 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+
+			By("create a PVC-PVC clone with object-map,fast-diff image-features and bind it to an app", func() {
+				// pvc clone is only supported from v1.16+
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 16) {
+					Skip("pvc clone is only supported from v1.16+")
```
nixpanic: it seems that these `Skip()` calls cause skipping of all the following `By()` tests as well. Instead log a warning and return from the `By()` function.

Ceph-CSI does not try to support Kubernetes < 1.16 anymore, so this check can be remove completely.

e2e/rbd.go
```diff
@@ -1466,6 +1466,41 @@ var _ = Describe("RBD", func() {
 				},
 			)
 
+			By(
+				"create a PVC and Bind it to an app with object-map,fast-diff image-features and default mounter",
+				func() {
+					err := deleteResource(rbdExamplePath + "storageclass.yaml")
+					if err != nil {
+						e2elog.Failf("failed to delete storageclass with error %v", err)
```
nixpanic: nit: Please drop the `with error` wording, we're trying to remove the redundancy of reporting an error and additionally mentioning `with error`. I understand many existing tests still include it, so it would be nice if you can change it for your tests in an update of this PR.

e2e/rbd.go
```diff
@@ -1720,6 +1793,7 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+
```
nixpanic: Please check for other additional unneeded empty lines.

nixpanic (commented):
Looks reasonable to me in general. Just a few nits mentioned related to the e2e tests.

internal/rbd/rbd_util.go
```diff
@@ -206,7 +206,18 @@ var (
...
+		},
+		librbd.FeatureNameObjectMap: {
+			needRbdNbd: false,
+			dependsOn:  []string{librbd.FeatureNameExclusiveLock},
+		},
+		librbd.FeatureNameFastDiff: {
+			needRbdNbd: false,
+			dependsOn:  []string{librbd.FeatureNameObjectMap},
+		},
+		librbd.FeatureNameDeepFlatten: {
```
Madhu-1: I don't see e2e for deepflatten feature can you please include this in e2e also?

e2e/rbd.go
```diff
@@ -1515,6 +1551,43 @@ var _ = Describe("RBD", func() {
...
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
+				}
+				err = createRBDStorageClass(
+					f.ClientSet,
+					f,
+					defaultSCName,
+					nil,
+					map[string]string{
+						"imageFeatures": "layering,exclusive-lock,object-map,fast-diff",
```
humblec: may be add `deep-flatten` too ?

e2e/rbd.go
```diff
@@ -1515,6 +1551,43 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+
+			By("create a PVC-PVC clone with object-map,fast-diff image-features and bind it to an app", func() {
+				// pvc clone is only supported from v1.16+
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 16) {
```
humblec: This is no longer needed, please see #2617 , in short, the version check below 1.20 is not needed any more, so you can remove this.

Madhu-1:
@mergifyio rebase

Madhu-1:
@k0ste do you need any help in moving this forward? if you are busy I can cherry-pick your commit and take this to completion by this week. please let me know what do you think?

internal/rbd/rbd_util.go
```diff
@@ -206,7 +206,18 @@ var (
 			needRbdNbd: false,
 		},
 		librbd.FeatureNameExclusiveLock: {
-			needRbdNbd: true,
+			needRbdNbd: false,
```
Madhu-1, line 209: as there is no mandatory requirement for the nbd mounted here, unit tests also need to be corrected for this one.

internal/rbd/rbd_util.go
Madhu-1: @k0ste any specific reason for adding deep-flatten image feature? currently deep flatten is handled internally by cephcsi to make clones and snapshots independent of each other

Madhu-1 (changes requested):
need to adjust the go-tests and fix go linter errors

Madhu-1:
I have tested this on my local minikube machine,  as expected the mapping fails as minikube uses a lower kernel version (4.9.x).


```text
rbd info replicapool/csi-vol-c3f1eb8a-5650-11ec-af4c-ca30c5bfe2cb
rbd image 'csi-vol-c3f1eb8a-5650-11ec-af4c-ca30c5bfe2cb':
	size 1 GiB in 256 objects
	order 22 (4 MiB objects)
	snapshot_count: 0
	id: 11cbfe59d66d
	block_name_prefix: rbd_data.11cbfe59d66d
	format: 2
	features: layering, exclusive-lock, object-map, fast-diff
	op_features: 
	flags: 
	create_timestamp: Mon Dec  6 04:55:42 2021
	access_timestamp: Mon Dec  6 04:55:42 2021
	modify_timestamp: Mon Dec  6 04:55:42 2021

```


* Pod describe ouput
```text
 Warning  FailedMount             0s (x4 over 4s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-47d81c8c-76de-420e-8c63-a130641dbec4" : rpc error: code = Internal desc = rbd: map failed with error an error (exit status 6) occurred while running rbd args: [--id csi-rbd-node -m 10.104.3.155:6789 --keyfile=***stripped*** map replicapool/csi-vol-c3f1eb8a-5650-11ec-af4c-ca30c5bfe2cb --device-type krbd --options noudev], rbd error output: rbd: sysfs write failed
```

* dmesg logs from the minikube node

```log
$ uname -r
4.19.202

$ dmesg
...
[  892.675164] libceph: mon0 10.104.3.155:6789 session established
[  892.675246] libceph: mon0 10.104.3.155:6789 socket closed (con state OPEN)
[  892.675262] libceph: mon0 10.104.3.155:6789 session lost, hunting for new mon
[  892.675594] libceph: mon0 10.104.3.155:6789 session established
[  892.675693] libceph: client4642 fsid e65ffd0f-8425-443f-97e8-276410d951cb
[  892.683899] rbd: image csi-vol-c3f1eb8a-5650-11ec-af4c-ca30c5bfe2cb: image uses unsupported features: 0x18
[  924.899178] libceph: mon0 10.104.3.155:6789 session established
[  924.899249] libceph: mon0 10.104.3.155:6789 socket closed (con state OPEN)
[  924.899265] libceph: mon0 10.104.3.155:6789 session lost, hunting for new mon
[  924.899601] libceph: mon0 10.104.3.155:6789 session established
[  924.899705] libceph: client4657 fsid e65ffd0f-8425-443f-97e8-276410d951cb
[  924.905023] rbd: image csi-vol-c3f1eb8a-5650-11ec-af4c-ca30c5bfe2cb: image uses unsupported features: 0x18
$ 
```

 @k0ste if you can remove the e2e  changes we can continue on reviews other changes

internal/rbd/rbd_util.go
k0ste: What you mean? `needRbdNbd` is already `false`

internal/rbd/rbd_util.go
k0ste: Deep-flatten may be needed by external administration features, out of Kubernetes scope

k0ste:
> @k0ste if you can remove the e2e changes we can continue on reviews other changes

What exactly e2e changes? I was included everything tips that your team left🤔

internal/rbd/rbd_util.go
Madhu-1: as I mentioned earlier deep flattening is already handled by cephcsi, I think we can have a separate PR with the use-case that it will solve.

internal/rbd/rbd_util.go
Madhu-1:
```=== RUN   TestGetCephClientLogFileName/test_for_empty_id
=== CONT  TestValidateImageFeatures
    rbd_util_test.go:176: 
        	Error Trace:	rbd_util_test.go:176
        	Error:      	Error message not equal:
        	            	expected: "feature exclusive-lock requires rbd-nbd for mounter"
        	            	actual  : "feature journaling requires rbd-nbd for mounter"
        	Test:       	TestValidateImageFeatures
    rbd_util_test.go:176: 
        	Error Trace:	rbd_util_test.go:176
        	Error:      	Error message not equal:
        	            	expected: "feature exclusive-lock requires rbd-nbd for mounter"
        	            	actual  : "feature journaling requires rbd-nbd for mounter"
        	Test:       	TestValidateImageFeatures
--- FAIL: TestValidateImageFeatures (0.00s)
```
please check the [go test](https://github.com/ceph/ceph-csi/runs/4427733123?check_suite_focus=true#step:3:1544) failure you will get what needs to be changed

Madhu-1:
> > @k0ste if you can remove the e2e changes we can continue on reviews other changes
> 
> What exactly e2e changes? I was included everything tips that your team leftthinking

the newly added e2e test will fail in our e2e environment as we are using a minikube to create a kubernetes cluster. please check https://github.com/ceph/ceph-csi/pull/2514#issuecomment-986443038

k0ste:
`deep-flatten` was removed as requested

Madhu-1:
Please run `CONTAINER_CMD=docker make containerized-test TARGET=go-test` and `run: CONTAINER_CMD=docker make containerized-test TARGET=go-lint` locally to see what is the issue if you see any go-test or go-lint test failure.

k0ste:
Some CI test failed due:
```
kernel 4.19.202 does not support required features
```

Madhu-1:
> Some CI test failed due:
> 
> ```
> kernel 4.19.202 does not support required features
> ```

CI is not going to pass as minikube comes with a kernel version of `4.19.202` which does not support fast-diff,obj-map... image feature. i tested this one and mentioned the same at https://github.com/ceph/ceph-csi/pull/2514#issuecomment-986443038 and also you need to remove the e2e changes you are adding as I mentioned here https://github.com/ceph/ceph-csi/pull/2514#issuecomment-986545025.

k0ste:
> > Some CI test failed due:
> > ```
> > kernel 4.19.202 does not support required features
> > ```
> 
> CI is not going to pass as minikube comes with a kernel version of `4.19.202` which does not support fast-diff,obj-map... image feature. i tested this one and mentioned the same at [#2514 (comment)](https://github.com/ceph/ceph-csi/pull/2514#issuecomment-986443038) and also you need to remove the e2e changes you are adding as I mentioned here [#2514 (comment)](https://github.com/ceph/ceph-csi/pull/2514#issuecomment-986545025).

e2e changes was removed

internal/rbd/rbd_util_test.go
```diff
@@ -88,29 +88,53 @@ func TestValidateImageFeatures(t *testing.T) {
...
+				Mounter: rbdDefaultMounter,
+			},
+			false,
+			"",
+		},
 		{
 			"layering,journaling",
 			&rbdVolume{
-				Mounter: rbdNbdMounter,
+				Mounter: rbdDefaultMounter,
```
pkalever, line 102: Why do we need this change?

pkalever (changes requested):
IMO, we shouldn't have removed e2e tests rather defended the test with `util.CheckKernelSupport` for the minimum required version, that way the CI will not bailout. 

@k0ste can this be reverted with added defence check?

cc: @Madhu-1

Madhu-1:
> IMO, we shouldn't have removed e2e tests rather defended the test with `util.CheckKernelSupport` for the minimum required version, that way the CI will not bailout.

Yes, but I don't think it's useful as this will never get executed in near time as we are still on 4.9 and what we need is 5.3 and moreover, we need to rework that one to check the kernel version on all nodes and schedule pod for that node.

internal/rbd/rbd_util_test.go
Madhu-1: no harm in this one.

Madhu-1 (approved):
LGTM

Madhu-1:
@mergifyio rebase

pkalever:
> Yes, but I don't think it's useful as this will never get executed in near time as we are still on 4.9 and what we need is 5.3 and moreover, we need to rework that one to check the kernel version on all nodes and schedule pod for that node.

Soon minikube will have 5.x, minkube community is working on it. I just wanted to avoid rework on it later.

@k0ste I don't insist on having the e2e, but please create an issue for this so that we don't miss the e2e coverage later.

pkalever (approved):
Well, the changes LGTM. My only concern is on the e2e tests. Assuming @k0ste will open an issue for e2e tests later. Approving this now.

PR #2529, opened 2021-09-21
rbd: utilize cookie support from rbd for nbd

pkalever:

**Problem:**
On remap/attach of device (i.e. nodeplugin restart), there is no way for rbd-nbd to defend if the backend storage is matching with the initial backend storage.

Say, if an initial map request for backend "pool1/image1" got mapped to /dev/nbd0 and the userspace process is terminated (on nodeplugin restart). A next remap/attach (nodeplugin start) request within reattach-timeout is allowed to use /dev/nbd0 for a different backend "pool1/image2"

For example, an operation like below could be dangerous:
```
$ sudo rbd-nbd map --try-netlink rbd-pool/ext4-image
/dev/nbd0
$ sudo blkid /dev/nbd0
/dev/nbd0: UUID="bfc444b4-64b1-418f-8b36-6e0d170cfc04" TYPE="ext4"
$ sudo pkill -15 rbd-nbd   <-- nodeplugin terminate
$ sudo rbd-nbd attach --try-netlink --device /dev/nbd0 rbd-pool/xfs-image
/dev/nbd0
$ sudo blkid /dev/nbd0
/dev/nbd0: UUID="d29bf343-6570-4069-a9ea-2fa156ced908" TYPE="xfs"
```

**Solution:**
rbd-nbd/kernel now provides a way to keep some metadata in sysfs to identify between the device and the backend, so that when a remap/attach request is made, rbd-nbd can compare and avoid such dangerous operations.

With the provided solution, as part of the initial map request, backend cookie (ceph-csi VOLID) can be stored in the sysfs per device config, so that on a remap/attach request rbd-nbd will check and validate if the backend per device cookie matches with the initial map backend with the help of cookie.

At Ceph-csi we use VOLID as device cookie, which will be unique, we pass the VOLID as cookie at map and use the same at the time of attach, that way rbd-nbd can identify backends and their matching devices.

**Requires:**
https://github.com/ceph/ceph/pull/41323
https://lkml.org/lkml/2021/4/29/274

- [x] Require PR https://github.com/ceph/ceph/pull/41323

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/rbd/rbd_attach.go
+++ b/internal/rbd/rbd_attach.go
@@ -250,7 +250,7 @@ func attachRBDImage(ctx context.Context, volOptions *rbdVolume, device string, c
 	return devicePath, err
 }
 
-func appendNbdDeviceTypeAndOptions(cmdArgs []string, isThick bool, userOptions string) []string {
+func appendNbdDeviceTypeAndOptions(cmdArgs []string, isThick bool, userOptions, cookie string) []string {
 	cmdArgs = append(cmdArgs, "--device-type", accessTypeNbd)
 
 	isUnmap := CheckSliceContains(cmdArgs, "unmap")
@@ -263,6 +263,10 @@ func appendNbdDeviceTypeAndOptions(cmdArgs []string, isThick bool, userOptions s
 		}
 		if !strings.Contains(userOptions, setNbdIOTimeout) {
 			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
+		}
+
+		if hasNBDCookieSupport {
+			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("cookie=%s", cookie))
 		}
 
 		if isThick {
@@ -309,7 +313,7 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, isThick bool, userOptions 
 
 // appendRbdNbdCliOptions append mandatory options and convert list of useroptions
 // provided for rbd integrated cli to rbd-nbd cli format specific.
-func appendRbdNbdCliOptions(cmdArgs []string, userOptions string) []string {
+func appendRbdNbdCliOptions(cmdArgs []string, userOptions, cookie string) []string {
 	if !strings.Contains(userOptions, useNbdNetlink) {
 		cmdArgs = append(cmdArgs, fmt.Sprintf("--%s", useNbdNetlink))
 	}
@@ -318,6 +322,9 @@ func appendRbdNbdCliOptions(cmdArgs []string, userOptions string) []string {
 	}
 	if !strings.Contains(userOptions, setNbdIOTimeout) {
 		cmdArgs = append(cmdArgs, fmt.Sprintf("--%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
+	}
+	if hasNBDCookieSupport {
+		cmdArgs = append(cmdArgs, fmt.Sprintf("--cookie=%s", cookie))
 	}
 	if userOptions != "" {
 		options := strings.Split(userOptions, ",")
@@ -362,11 +369,11 @@ func createPath(ctx context.Context, volOpt *rbdVolume, device string, cr *util.
 		// TODO: use rbd cli for attach/detach in the future
 		cli = rbdNbdMounter
 		mapArgs = append(mapArgs, "attach", imagePath, "--device", device)
-		mapArgs = appendRbdNbdCliOptions(mapArgs, volOpt.MapOptions)
+		mapArgs = appendRbdNbdCliOptions(mapArgs, volOpt.MapOptions, volOpt.VolID)
 	} else {
 		mapArgs = append(mapArgs, "map", imagePath)
 		if isNbd {
-			mapArgs = appendNbdDeviceTypeAndOptions(mapArgs, isThick, volOpt.MapOptions)
+			mapArgs = appendNbdDeviceTypeAndOptions(mapArgs, isThick, volOpt.MapOptions, volOpt.VolID)
 		} else {
 			mapArgs = appendKRbdDeviceTypeAndOptions(mapArgs, isThick, volOpt.MapOptions)
 		}
@@ -476,7 +483,7 @@ func detachRBDImageOrDeviceSpec(
 
 	unmapArgs := []string{"unmap", dArgs.imageOrDeviceSpec}
 	if dArgs.isNbd {
-		unmapArgs = appendNbdDeviceTypeAndOptions(unmapArgs, false, dArgs.unmapOptions)
+		unmapArgs = appendNbdDeviceTypeAndOptions(unmapArgs, false, dArgs.unmapOptions, dArgs.volumeID)
 	} else {
 		unmapArgs = appendKRbdDeviceTypeAndOptions(unmapArgs, false, dArgs.unmapOptions)
 	}
--- a/internal/rbd/rbd_attach.go
+++ b/internal/rbd/rbd_attach.go
@@ -67,10 +67,13 @@ const (
 	setNbdIOTimeout = "io-timeout"
 )
 
-var hasNBD = false
+var (
+	hasNBD              = true
+	hasNBDCookieSupport = false
+)
 
 func init() {
-	hasNBD = checkRbdNbdTools()
+	setRbdNbdToolFeatures()
 }
 
 // rbdDeviceInfo strongly typed JSON spec for rbd device list output (of type krbd).
@@ -193,27 +196,30 @@ func waitForPath(ctx context.Context, pool, namespace, image string, maxRetries 
 	return "", false
 }
 
-// Check if rbd-nbd tools are installed.
-func checkRbdNbdTools() bool {
+// set features available with rbd-nbd, and NBD module loaded status.
+func setRbdNbdToolFeatures() {
 	// check if the module is loaded or compiled in
 	_, err := os.Stat(fmt.Sprintf("/sys/module/%s", moduleNbd))
 	if os.IsNotExist(err) {
 		// try to load the module
 		_, _, err = util.ExecCommand(context.TODO(), "modprobe", moduleNbd)
 		if err != nil {
-			log.ExtendedLogMsg("rbd-nbd: nbd modprobe failed with error %v", err)
-
-			return false
+			hasNBD = false
+			log.WarningLogMsg("rbd-nbd: nbd modprobe failed with error %v", err)
 		}
 	}
-	if _, _, err := util.ExecCommand(context.TODO(), rbdTonbd, "--version"); err != nil {
-		log.ExtendedLogMsg("rbd-nbd: running rbd-nbd --version failed with error %v", err)
 
-		return false
+	stdout, stderr, err := util.ExecCommand(context.TODO(), rbdTonbd, "--help")
+	if err != nil || stderr != "" {
+		hasNBD = false
+		log.WarningLogMsg("running rbd-nbd --help failed with error:%v, stderr:%s", err, stderr)
 	}
-	log.ExtendedLogMsg("rbd-nbd tools were found.")
 
-	return true
+	if strings.Contains(stdout, "--cookie") {
+		hasNBDCookieSupport = true
+	}
+
+	log.DefaultLog("NBD module loaded: %t, rbd-nbd supported features, cookie: %t", hasNBD, hasNBDCookieSupport)
 }
 
 func attachRBDImage(ctx context.Context, volOptions *rbdVolume, device string, cr *util.Credentials) (string, error) {


pkalever:
`DNM` waiting for https://github.com/ceph/ceph/pull/41323

internal/util/util.go
```diff
@@ -353,3 +354,15 @@ func CallStack() string {
 
 	return string(stack)
 }
+
+// IsVersionSatisfied check if available version satisfies required version
+func IsVersionSatisfied(required, available string) bool {
```
nixpanic: Please add a unit test for this function in `internal/util/util_test.go`

nixpanic: Also, the commit message mentions "package versions"... what packages? Are these RPM, Golang or other versions?

internal/rbd/rbd_attach.go
```diff
@@ -193,8 +198,8 @@ func waitForPath(ctx context.Context, pool, namespace, image string, maxRetries
 	return "", false
 }
 
-// Check if rbd-nbd tools are installed.
-func checkRbdNbdTools() bool {
+// Check if rbd-nbd tools are installed and returns its version if possible.
```
nixpanic: In what case would it not be possible to get the version? If there is no known case, just return the version or an error. No need for returning an additional bool.

You could split this change from the cookie modification, and have a separate commit for it.

internal/rbd/rbd_attach.go
```diff
@@ -259,6 +269,10 @@ func appendNbdDeviceTypeAndOptions(cmdArgs []string, isThick bool, userOptions s
 			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
 		}
 
+		if cookie != "" && util.IsVersionSatisfied(rbdNbdCookieSupportVersion, rbdNbdVersion) {
```
nixpanic: move this to a helper function, like `cmdArgs = appendNbdCookieOption(cmdArgs, cookie)` or such

nixpanic (changes requested):
Looks like a good change to have. Left a few minor comments.

Have you considered inspecting the output of `rbd-nbd --help` to see if there is support for :cookie: ? That would make it possible to get support enabled in case some Ceph vendor backports the feature.

internal/rbd/rbd_attach.go
```diff
@@ -203,17 +208,22 @@ func checkRbdNbdTools() bool {
...
-	if _, _, err := util.ExecCommand(context.TODO(), rbdTonbd, "--version"); err != nil {
-		log.ExtendedLogMsg("rbd-nbd: running rbd-nbd --version failed with error %v", err)
-
-		return false
+	var rbdNbdVer string
+	stdout, stderr, err := util.ExecCommand(context.TODO(), rbdTonbd, "--version")
+	if err != nil || stderr != "" {
+		log.ExtendedLogMsg("rbd-nbd: running rbd-nbd --version failed with error:%v, stderr:%s", err, stderr)
+		return false, ""
+	} else if stdout != "" {
```
Madhu-1: any specific reason to have this check? instead of that we split stdout and check its length to make sure it's>3?

internal/util/util.go
```diff
@@ -353,3 +354,15 @@ func CallStack() string {
...
+	v1, _ := version.NewVersion(required)
+	v2, _ := version.NewVersion(available)
+
+	if v2.LessThan(v1) {
+		return false
+	}
+
+	return true
```
Madhu-1: `return !v2.LessThan(v1)` ? or `!(v2.LessThan(v1)==true`?

internal/rbd/rbd_attach.go
```diff
@@ -313,6 +327,9 @@ func appendRbdNbdCliOptions(cmdArgs []string, userOptions string) []string {
 	if !strings.Contains(userOptions, setNbdIOTimeout) {
 		cmdArgs = append(cmdArgs, fmt.Sprintf("--%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
 	}
+	if cookie != "" && util.IsVersionSatisfied(rbdNbdCookieSupportVersion, rbdNbdVersion) {
```
Madhu-1: check version only one and store it globally instead of checking it always?

internal/util/util.go
```diff
@@ -353,3 +354,15 @@ func CallStack() string {
 
 	return string(stack)
 }
+
+// IsVersionSatisfied check if available version satisfies required version
+func IsVersionSatisfied(required, available string) bool {
+	v1, _ := version.NewVersion(required)
+	v2, _ := version.NewVersion(available)
```
Madhu-1: are we discarding the error here?

internal/rbd/rbd_attach.go
```diff
@@ -65,12 +65,17 @@ const (
 	// `io-timeout` of rbd-nbd is to tweak NBD_ATTR_TIMEOUT. It specifies
 	// how long the IO should wait to get handled before bailing out.
 	setNbdIOTimeout = "io-timeout"
+
+	// rbd-nbd version with per device cookie support.
+	// TODO: Fix with right version when cookie changes merge.
+	rbdNbdCookieSupportVersion = "17.0.0"
```
Madhu-1: instead of depending on the version is it possible to check cookie is present in some help text of the CLI command?

pkalever: I was trying to address this in a more generic way, considering we might have to use the same utilities for other options in the future, today we are checking for `cookie` tomorrow we might want to check for other options or features.

internal/rbd/rbd_attach.go
pkalever: This was there already and I didn't want to break it, hence continued. I don't think there would be a possible case where the version cannot be returned.

Sure I will consider making the requested changes in the next respin.

internal/rbd/rbd_attach.go
pkalever: As also suggested by Niels at `checkRbdNbdTools`, let's also safely ignore a case that doesn't return the version?

internal/rbd/rbd_attach.go
pkalever: Ok

internal/rbd/rbd_attach.go
pkalever: Aren't we doing the same now? `rbdNbdVersion` is collected once and used everytime.

internal/util/util.go
pkalever: https://github.com/hashicorp/go-version#version-parsing-and-comparison

Should I rather stick to the binary version rather than the package version?

internal/util/util.go
pkalever: Yeah, I don't know how useful it would be to handle them. Not seen anything in examples handling them gracefully:
https://github.com/hashicorp/go-version#version-parsing-and-comparison

I can add one if you want.

internal/util/util.go
pkalever: Yes, looks simple I can pick one.

pkalever (commented):
@nixpanic @Madhu-1 I think I was on the other side of the coin thinking about reusing the `rbd-nbd --version` for future options instead of looking for `--cookie` or some other option individually, which could be a per option exec.

Also, the utility function introduced `IsVersionSatisfied` is generic enough for reusability.

That said, I would still like to know your preference among `rbd-nbd --version` or `rbd-nbd --help | grep cookie`


Thanks!

internal/rbd/rbd_attach.go
pkalever: OH! Do you mean the comparison part as well?

Like move this part to global:

```
isNbdCookie = util.IsVersionSatisfied(rbdNbdCookieSupportVersion, rbdNbdVersion)

```

internal/util/util.go
pkalever: Done now, changed it to "util: provide utility to parse and compare versions"

internal/util/util.go
pkalever: done will respin

internal/rbd/rbd_attach.go
pkalever: after the recent changes, looks like this is not needed for me.

internal/rbd/rbd_attach.go
Madhu-1: Won't this become problematic if someone backports the fix to another version? something like a downstream version?

internal/rbd/rbd_attach.go
```diff
@@ -193,35 +204,46 @@ func waitForPath(ctx context.Context, pool, namespace, image string, maxRetries
...
-		log.ExtendedLogMsg("rbd-nbd: running rbd-nbd --version failed with error %v", err)
+	stdout, stderr, err := util.ExecCommand(context.TODO(), rbdTonbd, "--version")
+	if err != nil || stderr != "" {
+		log.WarningLogMsg("rbd-nbd: running rbd-nbd --version failed with error:%v, stderr:%s", err, stderr)
+
+		return ""
+	}
+	ver := strings.Split(stdout, " ")
+	// ex: ceph version 15.2.14 cd3bb7e87a2f62c1b862ff3fd8b1eec13391a5be) octopus (stable)
+	rbdNbdVer := ver[2]
```
Madhu-1: this can cause panic if the ver length is not 3? need a  length check?

internal/rbd/rbd_attach.go
```diff
@@ -193,35 +204,46 @@ func waitForPath(ctx context.Context, pool, namespace, image string, maxRetries
...
+	return rbdNbdVer
 }
 
 func attachRBDImage(ctx context.Context, volOptions *rbdVolume, device string, cr *util.Credentials) (string, error) {
 	var err error
 
 	image := volOptions.RbdImageName
 	useNBD := false
-	if volOptions.Mounter == rbdTonbd && hasNBD {
+	if volOptions.Mounter == rbdTonbd {
```
Madhu-1: we no longer default to krbd if nbd is not present?

internal/rbd/rbd_attach.go
pkalever: Upstream: I think this will be considered as a feature, and hence I don't think ceph will backport it to minor z-stream releases.

Downstream: In, downstream we should change this to something else ?

internal/rbd/rbd_attach.go
pkalever: oops! right, I thought I added one based on your previous comment, will fix it.

internal/rbd/rbd_attach.go
Madhu-1: it good to reuse the same code for both upstream and downstream instead of having downstream only change and I have seen few important bug fixes are backported from octopus to nautilus etc. and also if someone is maintaining their own ceph downstream it might break. just trying to cover all cases here instead of depending on the version?

internal/rbd/rbd_attach.go
pkalever: Not sure on this part, but how does it matter for us here (just curious)?

Madhu-1: if the nbd module was not available on the node we use krbd to map the image by default with this change we don't default to anything we always try to use nbd if it's specified in storageclass?

internal/rbd/rbd_attach.go
pkalever: @nixpanic @Madhu-1 I understand your concern, and it kind of makes sense to me in a different angle, I was just thinking on the reusability part.

Okay! I will grep for `cookie` for now to satisfies your review :-)

BTW, do we want to keep the `IsVersionSatisfied` util func as I have also added `TestIsVersionSatisfied` may be for later use or just discard that work? (I'm just trying to save some future cycles)

Thanks!

Madhu-1: Thank you :+1:  if it's not used anywhere IMO it should be removed.

internal/rbd/rbd_attach.go
pkalever: > by default with this change we don't default to anything we always try to use nbd if it's specified in storageclass?

I'm sorry, I might be missing something, which part of the code changes from this PR is leading to the behaviour.

Madhu-1: `if volOptions.Mounter == rbdTonbd && hasNBD {` to `if volOptions.Mounter == rbdTonbd`

pkalever: this is only to avoid checking if rbd-nbd is installed or not (which is redundant) rest remain as it is IMHO.

nixpanic: Previously there NBD was only used when `hasNBD` was set. So, in the case where `checkRbdNbdTools()` returned false, a StorageClass configured to use NBD would fallback to krbd instead. This is a change in behaviour, and we do not know if there are any users that rely on it.

pkalever: right!
so the question is
1. should we have `nbdLoaded` variable that checks if the module is loaded, if not fall back to `krbd` ?
2. Or just fail if the storage class asks for `rbd-nbd` and `lsmod` doesn't list it?

We can still avoid hasNBD checking for rbd-nbd binary availability on the container which is redundant.

pkalever: @nixpanic @Madhu-1 check the latest updates, I have left hasNBD for module loaded case, and added `hasNBDCookieSupport` separately. Hope that will keep the old style of fallback intact.

Thanks!

internal/rbd/rbd_attach.go
```diff
@@ -193,27 +194,30 @@ func waitForPath(ctx context.Context, pool, namespace, image string, maxRetries
 	return "", false
 }
 
-// Check if rbd-nbd tools are installed.
-func checkRbdNbdTools() bool {
+// setRbdNbdToolFeatures set features available with rbd-nbd.
+func setRbdNbdToolFeatures() {
 	// check if the module is loaded or compiled in
+	hasNBD = true
```
nixpanic: Default is `false`, this temporary changes it to `true`, after which it may become `false` again. That isn't very clean. Better only set it once to `true` after successful detection.

internal/rbd/rbd_attach.go
```diff
@@ -68,9 +68,10 @@ const (
 )
 
 var hasNBD = false
+var hasNBDCookieSupport = false
```
nixpanic: use different formatting please:
```go
var (
    hasNBD = false
    hasNBDCookieSupport = false
)
```

internal/rbd/rbd_attach.go
```diff
@@ -259,6 +263,10 @@ func appendNbdDeviceTypeAndOptions(cmdArgs []string, isThick bool, userOptions s
 			cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
 		}
 
+		if cookie != "" && hasNBDCookieSupport {
```
nixpanic: instead of repeating this if statement, add a helper:
```go
cmdArgs = appendRbdNbdCookie(cmdArgs, cookie)

func appendRbdNbdCookie(...) []string {
    if cookie != "" && hasNBDCookieSupport {
        cmdArgs = ....
    }
    return cmdArgs
}
```

internal/rbd/rbd_attach.go
```diff
@@ -313,6 +321,9 @@ func appendRbdNbdCliOptions(cmdArgs []string, userOptions string) []string {
 	if !strings.Contains(userOptions, setNbdIOTimeout) {
 		cmdArgs = append(cmdArgs, fmt.Sprintf("--%s=%d", setNbdIOTimeout, defaultNbdIOTimeout))
 	}
+	if cookie != "" && hasNBDCookieSupport {
```
nixpanic: prevent this duplication by adding a `appendRbdNbdCookie()` helper?

internal/rbd/rbd_attach.go
pkalever: Ok

internal/rbd/rbd_attach.go
pkalever: sure

internal/rbd/rbd_attach.go
pkalever: I don't see much value @nixpanic 

The syntax of cmdArgs in both cases are different

one is:

```
if cookie != "" && hasNBDCookieSupport {
        cmdArgs = append(cmdArgs, "--options", fmt.Sprintf("cookie=%s", cookie))
}

```

and the other is:
```
if cookie != "" && hasNBDCookieSupport {
        cmdArgs = append(cmdArgs, fmt.Sprintf("--cookie=%s", cookie))
}
```

pkalever: We need to defent further with a bool again, I would prefer keep it as it is just like others in the respective functions.

internal/rbd/rbd_attach.go
pkalever: Same as above.

pkalever (commented):
Thanks @nixpanic

internal/rbd/rbd_attach.go
pkalever: @nixpanic actually we don't have to check for empty `cookie` as it is passed from VolID which will never be empty?

pkalever:
tests are failing:

```
[1mSTEP[0m: create a PVC and bind it to an app with normal user
Sep 30 06:45:48.183: INFO: Waiting up to &PersistentVolumeClaim{ObjectMeta:{rbd-1318  rbd-1318    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Spec:PersistentVolumeClaimSpec{AccessModes:[ReadWriteOnce],Resources:ResourceRequirements{Limits:ResourceList{},Requests:ResourceList{storage: {{1073741824 0} {<nil>} 1Gi BinarySI},},},VolumeName:,Selector:nil,StorageClassName:*csi-rbd-sc,VolumeMode:nil,DataSource:nil,DataSourceRef:nil,},Status:PersistentVolumeClaimStatus{Phase:,AccessModes:[],Capacity:ResourceList{},Conditions:[]PersistentVolumeClaimCondition{},},} to be in Bound state
Sep 30 06:45:48.183: INFO: waiting for PVC rbd-1318 (0 seconds elapsed)
Sep 30 06:45:50.192: INFO: waiting for PVC rbd-1318 (2 seconds elapsed)
Sep 30 06:45:50.199: INFO: Waiting for PV pvc-626fafad-908a-4a53-88de-cce70479d2b1 to bind to PVC rbd-1318
Sep 30 06:45:50.200: INFO: Waiting up to timeout=10m0s for PersistentVolumeClaims [rbd-1318] to have phase Bound
Sep 30 06:45:50.203: INFO: PersistentVolumeClaim rbd-1318 found and phase=Bound (3.129072ms)
Sep 30 06:45:50.203: INFO: Waiting up to 10m0s for PersistentVolume pvc-626fafad-908a-4a53-88de-cce70479d2b1 to have phase Bound
Sep 30 06:45:50.205: INFO: PersistentVolume pvc-626fafad-908a-4a53-88de-cce70479d2b1 found and phase=Bound (2.568213ms)
Sep 30 06:45:50.217: INFO: Waiting up to pod-run-as-non-root to be in Running state
Sep 30 06:55:50.229: FAIL: failed to validate normal user pvc and application binding with error timed out waiting for the condition
```
This doesn't look like an issue related to this PR

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.20/runs/2619/nodes/109/log/?start=0

internal/rbd/rbd_attach.go
```diff
@@ -193,27 +196,32 @@ func waitForPath(ctx context.Context, pool, namespace, image string, maxRetries
...
 	_, err := os.Stat(fmt.Sprintf("/sys/module/%s", moduleNbd))
 	if os.IsNotExist(err) {
 		// try to load the module
 		_, _, err = util.ExecCommand(context.TODO(), "modprobe", moduleNbd)
 		if err != nil {
-			log.ExtendedLogMsg("rbd-nbd: nbd modprobe failed with error %v", err)
+			log.WarningLogMsg("rbd-nbd: nbd modprobe failed with error %v", err)
 
-			return false
+			return false // failed to load nbd module
```
Madhu-1: no need to return bool from here? instead of that set `hasNBD` the same way you are handling the `hasNBDCookieSupport`variable

Madhu-1 (changes requested):
small suggestion. LGTM. requesting changes to hold this one till we have the next ceph release

@pkalever you also mention we have dependencies with ceph and kernel also. what if ceph has --cookie flag but the kernel does not support it. do you see any problem?

internal/rbd/rbd_attach.go
pkalever: Done now.

pkalever:
> small suggestion. LGTM. requesting changes to hold this one till we have the next ceph release

Maybe you mean until the cookie PR is merged in ceph repo!?
`DNM` is already in place + there is a waiting task on the ceph PR which should keep this on track

> 
> @pkalever you also mention we have dependencies with ceph and kernel also. what if ceph has --cookie flag but the kernel does not support it. do you see any problem?

If the kernel doesn't support `cookie` then rbd-nbd (ceph) will just ignore it even if the user (ceph-csi) passes the `--cookie` option, hence it's safe and doesn't hurt to pass `--cookie` even if the kernel doesn't support it

internal/rbd/rbd_attach.go
```diff
@@ -193,27 +196,34 @@ func waitForPath(ctx context.Context, pool, namespace, image string, maxRetries
...
-		log.ExtendedLogMsg("rbd-nbd: running rbd-nbd --version failed with error %v", err)
 
-		return false
+	if isModuleLoaded {
+		hasNBD = true
+	}
+
+	stdout, stderr, err := util.ExecCommand(context.TODO(), rbdTonbd, "--help")
+	if err != nil || stderr != "" {
+		log.WarningLogMsg("running rbd-nbd --help failed with error:%v, stderr:%s", err, stderr)
```
nixpanic, line 215: if running the command fails, does it still make sense to set `hasNBD` to `true` as done above? Maybe set it after running the command and returning from the function on a failure?

pkalever: Sure :+1:

Madhu-1: isModuleLoaded should be set to false here also? am also thinking do we need an extra variable? can we just set `hasNBD` to false when running modproble or os.stat and `--help` fails?

internal/rbd/rbd_attach.go
```diff
@@ -193,27 +196,34 @@ func waitForPath(ctx context.Context, pool, namespace, image string, maxRetries
 	return "", false
 }
 
-// Check if rbd-nbd tools are installed.
-func checkRbdNbdTools() bool {
+// set features available with rbd-nbd, and NBD module loaded status.
+func setRbdNbdToolFeatures() {
 	// check if the module is loaded or compiled in
+	isModuleLoaded := true
```
Madhu-1: if os.Stat fails with any other error than IsNotExists `isModuleLoaded` remains `true`
`

internal/rbd/rbd_attach.go
pkalever: I used hasNBD, later I thought the reviewers didn't like it.

Madhu-1: i think the request from Niels is to set `hasNBD` to false if running `--help` command fails

internal/rbd/rbd_attach.go
pkalever: Yes it should be set to true as before. We should just make sure /sys/module/nbd exists.

pkalever: > if os.Stat fails with any other error than IsNotExists isModuleLoaded remains true

Yes, this is intentional.

Madhu-1 (approved):

pkalever:
@Mergifyio rebase

pkalever:
@ceph/ceph-csi-contributors this PR is ready for review/merge. Thanks!

pkalever:
@nixpanic could you also please help review this?

Thanks!

nixpanic (approved):

pkalever:
```

```

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.20/detail/mini-e2e_k8s-1.20/2998/pipeline

pkalever:
```


```

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/upgrade-tests-cephfs/detail/upgrade-tests-cephfs/3579/pipeline

PR #2546, opened 2021-09-29
RBD: Fix encrypted restore with different kms config

Rakshith-R:
- rbd: modify copyEncryptionConfig to accept copyOnlyPassphrase arg 

`During PVC snapshot/clone both kms config and passphrase needs to copied,
while for PVC restore only passphrase needs to be copied to dest rbdvol
since destination storageclass may have another kms config.`

- e2e: modify validatePVCSnapshot() to use restoreSCName & restoreKMS

- e2e: add testcase for PVC restore from vaultKMS to vaultTenantSAKMS 
- e2e: add test for PVC restore from vaultKMS to userSecretMetadataKMS

Example of failed e2e run when vaultTenantSAKMS is used as restore SC 
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.20/detail/mini-e2e_k8s-1.20/2627/pipeline

search for `secret ID` , mount fails because it can't find encryption passphrase.

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -1102,7 +1102,8 @@ var _ = Describe("RBD", func() {
 						snapshotPath,
 						pvcClonePath,
 						appClonePath,
-						noKMS,
+						noKMS, noKMS,
+						defaultSCName,
 						f)
 				}
 			})
@@ -1169,7 +1170,11 @@ var _ = Describe("RBD", func() {
 					e2elog.Failf("failed to create storageclass with error %v", err)
 				}
 
-				validatePVCSnapshot(1, pvcPath, appPath, snapshotPath, pvcClonePath, appClonePath, vaultKMS, f)
+				validatePVCSnapshot(1,
+					pvcPath, appPath, snapshotPath, pvcClonePath, appClonePath,
+					vaultKMS, vaultKMS,
+					defaultSCName,
+					f)
 
 				err = deleteResource(rbdExamplePath + "storageclass.yaml")
 				if err != nil {
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -1186,6 +1186,63 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("Validate PVC restore from vaultKMS to vaultTenantSAKMS", func() {
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 16) {
+					Skip("pvc clone is only supported from v1.16+")
+				}
+				restoreSCName := "restore-sc"
+				err := deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass: %v", err)
+				}
+				scOpts := map[string]string{
+					"encrypted":       "true",
+					"encryptionKMSID": "vault-test",
+				}
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, scOpts, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass: %v", err)
+				}
+
+				scOpts = map[string]string{
+					"encrypted":       "true",
+					"encryptionKMSID": "vault-tenant-sa-test",
+				}
+				err = createRBDStorageClass(f.ClientSet, f, restoreSCName, nil, scOpts, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass: %v", err)
+				}
+
+				err = createTenantServiceAccount(f.ClientSet, f.UniqueName)
+				if err != nil {
+					e2elog.Failf("failed to create ServiceAccount: %v", err)
+				}
+				defer deleteTenantServiceAccount(f.UniqueName)
+
+				validatePVCSnapshot(1,
+					pvcPath, appPath, snapshotPath, pvcClonePath, appClonePath,
+					vaultKMS, vaultTenantSAKMS,
+					restoreSCName, f)
+
+				err = retryKubectlArgs(cephCSINamespace, kubectlDelete, deployTimeout, "storageclass", restoreSCName)
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass %q: %v", restoreSCName, err)
+				}
+
+				err = deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass: %v", err)
+				}
+
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0, defaultRBDPool)
+
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
+				if err != nil {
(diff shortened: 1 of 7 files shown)


internal/rbd/encryption.go
```diff
@@ -135,8 +135,11 @@ func (ri *rbdImage) copyEncryptionConfig(cp *rbdImage) error {
 		return fmt.Errorf("failed to fetch passphrase for %q: %w",
 			ri, err)
 	}
-
-	cp.encryption, err = util.NewVolumeEncryption(ri.encryption.GetID(), ri.encryption.KMS)
+	if copyOnlyPassphrase {
+		cp.encryption, err = util.NewVolumeEncryption(cp.encryption.GetID(), cp.encryption.KMS)
+	} else {
+		cp.encryption, err = util.NewVolumeEncryption(ri.encryption.GetID(), ri.encryption.KMS)
+	}
```
Madhu-1: instead of this new variable, the same can be done if we compare KMS/ID?

Rakshith-R: Actually cp.encrypyion is `nil` at places, so we cannot do that.
(Since it is reused for snapshotting and cloning (+intermediate steps))
And to have very minimal changes to be able to backport to 3.4.

Rakshith-R:

failed due to #2264 
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.20/detail/mini-e2e-helm_k8s-1.20/2627/pipeline

internal/rbd/encryption.go
```diff
@@ -135,8 +135,11 @@ func (ri *rbdImage) copyEncryptionConfig(cp *rbdImage) error {
 		return fmt.Errorf("failed to fetch passphrase for %q: %w",
 			ri, err)
 	}
-
-	cp.encryption, err = util.NewVolumeEncryption(ri.encryption.GetID(), ri.encryption.KMS)
+	if copyOnlyPassphrase {
+		cp.encryption, err = util.NewVolumeEncryption(cp.encryption.GetID(), cp.encryption.KMS)
```
nixpanic: this seems wrong, setting `cp.encryption` based on the contents of `cp.encryption` is weird. It would need at minimum a very clear comment explaining why this is needed, and works at all.

e2e/utils.go
```diff
@@ -867,6 +871,29 @@ func validatePVCSnapshot(
...
+							"failed to get image info for %s namespace=%s volumehandle=%s error=%w",
+							name,
+							p.Namespace,
+							imageData.csiVolumeHandle,
+							sErr)
+					} else {
+						// check new passphrase created
+						stdOut, stdErr := restoreKMS.getPassphrase(f, imageData.csiVolumeHandle)
+						if stdOut != "" {
+							e2elog.Logf("successfully read the passphrase from vault: %s", stdOut)
```
nixpanic: this refers to "vault", but that is not the only passphrase storage that cam be configured? Is the error of `getPassphrase()` not clear enough?

e2e/rbd.go
```diff
@@ -1145,6 +1145,63 @@ var _ = Describe("RBD", func() {
...
 			})
 
+			By("Validate PVC restore from vaultKMS to vaultTenantSAKMS", func() {
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 16) {
+					Skip("pvc clone is only supported from v1.16+")
+				}
+				restoreSCName := "restore-sc"
+				err := deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass with error %v", err)
```
nixpanic: new error logging should ideally not include the repetitive/obvious `with error`, but only `<message>: err`.

e2e/utils.go
Rakshith-R: >this refers to "vault", but that is not the only passphrase storage that cam be configured? Is the error of getPassphrase() not clear enough?

The stdOut is not required here, removed it.

e2e/rbd.go
Rakshith-R: done

internal/rbd/encryption.go
Rakshith-R: >this seems wrong, setting cp.encryption based on the contents of cp.encryption is weird. It would need at minimum a very clear comment explaining why this is needed, and works at all.


yeah, it was redundant :facepalm: , fixed now, PTAL
destination rbdvol.encryption for snapshot restore PVC will be set/Initialized earlier, so its not required

internal/rbd/encryption.go
```diff
@@ -123,7 +123,7 @@ func (ri *rbdImage) setupEncryption(ctx context.Context) error {
 // rbdImage to the passed argument. This function re-encrypts the passphrase
 // from the original, so that both encrypted passphrases (potentially, depends
 // on the DEKStore) have different contents.
-func (ri *rbdImage) copyEncryptionConfig(cp *rbdImage) error {
+func (ri *rbdImage) copyEncryptionConfig(cp *rbdImage, copyOnlyPassphrase bool) error {
```
nixpanic, line 130: it would help to explain in the above comment when to set `copyOnlyPassphrase` to `true`, and what it exactly does.

internal/rbd/encryption.go
```diff
@@ -135,8 +135,9 @@ func (ri *rbdImage) copyEncryptionConfig(cp *rbdImage) error {
 		return fmt.Errorf("failed to fetch passphrase for %q: %w",
 			ri, err)
 	}
-
-	cp.encryption, err = util.NewVolumeEncryption(ri.encryption.GetID(), ri.encryption.KMS)
+	if !copyOnlyPassphrase {
```
nixpanic, line 143: add empty lines around this if-block to increase readability?

Move the error check and `SetDEKStore` inside the if-statement too?

e2e/utils.go
```diff
@@ -1353,6 +1353,7 @@ func retryKubectlFile(namespace string, action kubectlAction, filename string, t
 // retryKubectlArgs takes a namespace and action telling kubectl what to do
 // with the passed arguments. This function retries until no error occurred, or
 // the timeout passed.
+// nolint:unparam // retryKubectlArgs will be used with kubectlDelete arg later on.
```
nixpanic, line 1361: should `deleteResource` use this function then?

nixpanic (commented):
just a few nits, looks pretty complete to me

e2e/utils.go
Rakshith-R: its using `retryKubectlInput`  
https://github.com/ceph/ceph-csi/blob/90d246fc55142fbcfe9ef8fd34c4a20526c00ce0/e2e/utils.go#L126

Rakshith-R:

failed due to #2264 

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/upgrade-tests-rbd/detail/upgrade-tests-rbd/3258/pipeline

nixpanic (approved):

Madhu-1 (approved):

Issue #2541, opened 2021-09-28
Expose SCC's through ceph-csi package


leseb:
# Describe the feature you'd like to have #

The ceph-csi project is imported by several projects, typically Kubernetes operators.
When running on environments like OpenShift, Security Context Constraints need to be imported for Ceph-CSI to operate normally.
It would be nice if ceph-csi could export its own SCCs through its API (one of the packages). An exported function that returns the minimum SCCs.

# What is the value to the end-user? (why is it a priority?) #

This will reduce friction for consumers of Ceph-CSI who don't need to know or hardcore the SCCs in their own code. They can just pull the required SCC from any given branch.

# How will we know we have a good solution? (acceptance criteria) #

SCCs can be imported from ceph-csi.

# Additional context #

Rook-Ceph does that already https://github.com/rook/rook/blob/dcb88a71d1668f1b9a74c4af3fd2053fc39c0045/pkg/apis/ceph.rook.io/v1/scc.go#L30

Madhu-1:
is it only to export SCC or pod/service templates also?

leseb:
> is it only to export SCC or pod/service templates also?

I'll open a different one for the templates :-)

nixpanic:
Could this easily be used to generate the deployment/Helm yaml files too? That would be ideal!

Madhu-1:
Yes, we need to unmarshal the file and expose that as a function in the cephcsi package. main note here we don't have CI to test the SCC atleast

nixpanic:
What I meant is that I hope we can maintain a single location (as a public API) where the SCC (and other templates) are maintained. When these are updated, we would need to run a small tool that takes the templates from the API and generates the files under `charts/` and `deploy/`.

Madhu-1:
Sounds good :+1:

leseb:
> Could this easily be used to generate the deployment/Helm yaml files too? That would be ideal!

Yes definitely, I have opened https://github.com/ceph/ceph-csi/issues/2542 separately, we can merge both if you want.

nixpanic:
Hmm, I have been looking at this, and considered giving it a try. But there is no reference to `SecurityContextConstraints` or `security.openshift.io` in the Ceph-CSI repository. Where is a SCC for Ceph-CSI maintained now?

leseb:
[...]

PR #2551, opened 2021-09-30
build: add API for generating OpenShift SecurityContextConstraints

nixpanic:
This new Go module allows other projects (like Rook) to consume
deployment details (the SCC to get started) directly from Ceph-CSI.

tools/yamlgen is a minimal example of how the Ceph-CSI api/deploy
package can be used to generate the SecurityContextConstraints
deployment object.

Further enhancements should make it possible to generate other
deployment files from the API, possibly based on parameters passed to
the yamlgen tool.

Updates: #2541

---

<details>

  progress or results

</details>


api/deploy/ocp/scc.yaml
```diff
@@ -0,0 +1,47 @@
...
+supplementalGroups:
+  type: RunAsAny
+# The type of volumes which are mounted to csi pods
+volumes:
+  - configMap
+  - projected
+  - emptyDir
+  - hostPath
+users:
+  # A user needs to be added for each rook service account.
```
leseb: It's generic right? So no need to mention rook.

api/deploy/ocp/scc.yaml
```diff
@@ -0,0 +1,47 @@
...
+  # This assumes running in the default sample "rook-ceph" namespace.
```
leseb: ditto

api/deploy/ocp/scc.yaml
```diff
@@ -0,0 +1,47 @@
...
+  # If other namespaces or service accounts are configured, they need to be
+  # updated here.
+  - "system:serviceaccount:{{ .Namespace }}:{{ .Prefix }}csi-rbd-plugin-sa"
```
leseb, line 39:
```suggestion
  - "system:serviceaccount:{{ .Namespace }}:{{ .Prefix }}-csi-rbd-plugin-sa"
```

So we don't have to include `-` in the prefix?

api/deploy/ocp/scc.go
```diff
@@ -0,0 +1,72 @@
...
+var SecurityContextConstraintsDefaults = SecurityContextConstraintsValues{
+	Namespace: "ceph-csi",
+	Prefix:    "",
+}
+
+// NewSecurityContextConstraints takes a namespace and the name of the deployer
+// (like Rook) and returns a SecurityContextConstraints object that can be
+// deployed on OpenShift.
+//
+// The deployer parameter (when not an empty string) is used as a prefix for
```
leseb, line 54: Can we clarify the deployer parameter? Especially if we want to include an `-` in it? Right now it's unclear how the `user` in the SCC will turn into.

api/deploy/ocp/scc_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+func TestNewSecurityContextConstraints(t *testing.T) {
+	var scc *secv1.SecurityContextConstraints
+
+	getSCC := func() {
+		scc = NewSecurityContextConstraints(SecurityContextConstraintsDefaults)
+	}
+
+	require.NotPanics(t, getSCC)
+	require.NotNil(t, scc)
```
leseb: How about adding more validation? Especially for `users` and when a deployer is passed.

api/deploy/ocp/scc.yaml
nixpanic: oh, yes, that is just a left-over from copy/pasting

api/deploy/ocp/scc.yaml
nixpanic: not everything might want to have a prefix, so the `-` would be optional

api/deploy/ocp/scc.go
nixpanic: Right. I'll make it so that if no `deployer` is passed, the prefix is empty, otherwise the prefix will be set to `<deployer>-` with the `-`.

api/deploy/ocp/scc_test.go
nixpanic: Sure, this is just to get started :smiley:

api/deploy/ocp/scc.go
```diff
@@ -0,0 +1,90 @@
...
+// template.
+type SecurityContextConstraintsValues struct {
+	// Namespace contains the OpenShift Namespace where the SCC will be
+	// used.
+	Namespace string
+	// Deployer refers to the Operator that creates the SCC and
+	// ServiceAccounts. This is an optional option.
+	Deployer  string
+
+	// Prefix is used internally only, might be verwritten
```
leseb:
```suggestion
	// Prefix is used internally only, might be overwritten
```

leseb: I think the comment is confusing. "used internally only", if so, then why is it exported? But then "might be overwritten" so I believe it contradicts the previous statement.

nixpanic: OK, I'll hide it from the API and make it a private thing.

api/deploy/ocp/scc.go
```diff
@@ -0,0 +1,103 @@
...
+	SecurityContextConstraintsValues
+
+	// Prefix is based on SecurityContextConstraintsValues.Deployer.
+	Prefix string
+}
+
+// NewSecurityContextConstraintsYAMLNewSecurityContextConstraintsYAML returns a
+// YAML string where the variables in the template have been replaced by the
+// values set in the SecurityContextConstraintsValues.
+func NewSecurityContextConstraintsYAML(values SecurityContextConstraintsValues) string {
```
Madhu-1: NewSecurityContextConstraintsYAMLNewSecurityContextConstraintsYAML to NewSecurityContextConstraintsYAML

api/deploy/ocp/scc.yaml
```diff
@@ -0,0 +1,43 @@
+---
+kind: SecurityContextConstraints
+apiVersion: security.openshift.io/v1
+metadata:
+  name: "{{ .Prefix }}ceph-csi"
```
Madhu-1, line 5: make this name completely configurable? instead of just supporting the prefix?

api/deploy/ocp/scc.yaml
```diff
@@ -0,0 +1,43 @@
...
+  type: RunAsAny
+# The type of volumes which are mounted to csi pods
+volumes:
+  - configMap
+  - projected
+  - emptyDir
+  - hostPath
+users:
+  # A user needs to be added for each service account.
+  - "system:serviceaccount:{{ .Namespace }}:{{ .Prefix }}csi-rbd-plugin-sa"
```
Madhu-1, line 39: what if someone deploys cephcsi with a different SA names?

tools/yamlgen/main.go
```diff
@@ -0,0 +1,37 @@
...
+# /!\ DO NOT MODIFY THIS FILE
+#
+# This file has been automatically generated by Ceph-CSI yamlgen.
+# The source for the contents can be found in the api/deploy directory, make
+# your modifications there.
+#`
+
+func main() {
+	fmt.Println(header)
+	fmt.Print(ocp.NewSecurityContextConstraintsYAML(ocp.SecurityContextConstraintsDefaults))
```
Madhu-1: can we directly write files to a specific location here itself instead of printing and redirecting the output in a makefile?

api/deploy/ocp/scc.go
nixpanic: Ah, yes, will fix that.

api/deploy/ocp/scc.yaml
nixpanic: We could, but not sure it is helpful. This is a Ceph-CSI piece, so it should be fine to have `ceph-csi` as part of the name?

The API always offers an option to set a different name if users really want to. They can call `NewSecurityContextConstraints()` and set the `.Name` of the object.

api/deploy/ocp/scc.yaml
nixpanic: Users may want to do that. These deployment artifacts are more of an example than anything else, users may want to modify them before deploying. Any deployment application (like Rook) could change the contents of the generated object as well. At the moment, I do not see a benefit in a more complex approach that allows configurable names (or selection per component like cephfs/rbd).

tools/yamlgen/main.go
nixpanic: yeah, I was planning that for a follow-up, but can include a start of that now too.

api/deploy/ocp/scc.yaml
Madhu-1: sounds good for now.

api/deploy/ocp/scc.yaml
Madhu-1: NewSecurityContextConstraints replace prefix with  deployer still we will get ceph-csi as  suffix I think it should be fine

tools/yamlgen/main.go
Madhu-1: if we do that we are good to go on this one from my point.

api/deploy/ocp/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/ceph/ceph-csi/api/deploy/ocp
```
Madhu-1: @nixpanic we will have one more go.mod for the kubernetes specific folder also? instead of that is it makes it easy to have single go.mod for api's?

nixpanic: Yes, it is intentional to have `go.mod` for a Kubernetes package too. If there would be an other container platform that we want to support this way, it will prevent pulling in dependencies from different platforms.

But, on the other hand, golang might be smart enough to include only the dependencies that are needed. So, I don't really have a strong preference.

Madhu-1: For easy maintenance  IMO we  can have single go.mod in api/deploy folder

api/deploy/ocp/scc.go
```diff
@@ -0,0 +1,103 @@
...
+		panic(err)
+	}
+
+	return scc
+}
+
+// internalSecurityContextConstraintsValues extends
+// SecurityContextConstraintsValues with some private attributes that may get
+// set based on other values.
+type internalSecurityContextConstraintsValues struct {
```
Madhu-1, line 74: as it's unexported can someone make use of it?

api/deploy/ocp/scc.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	tmpl, err := template.New("SCC").Parse(securityContextConstraints)
+	if err != nil {
+		panic(err)
+	}
+	err = tmpl.Execute(&buf, internalValues)
+	if err != nil {
+		panic(err)
+	}
```
Madhu-1: should we return the error let the caller handle the error message?

Makefile
```diff
@@ -160,6 +163,14 @@ cephcsi: check-env
 e2e.test: check-env
 	go test $(GO_TAGS) -mod=vendor -c ./e2e
 
+#
+# Update the generated deploy/scc.yaml when the template changed.
+# This requires running 'go mod vendor' so update the API files under the
+# vendor/ directory.
+deploy/scc.yaml: api/deploy/ocp/scc.yaml
```
Madhu-1: this should be some generic command (`generate-templates`) that will generate all artifacts?

Madhu-1 (commented):
LGTM. left few questions/suggestions

api/deploy/ocp/scc.go
nixpanic: no, it is internal only, used to gather private values based on the contents of the public SecurityContextConstraintsValues struct.

api/deploy/ocp/scc.go
nixpanic: Not sure how this can fail?

Makefile
nixpanic: Yes, and probably it should be maintained in a Makefile under the `deploy/` directory as all yaml files need to be listed there.

api/deploy/ocp/go.mod
nixpanic: sure, I'll move it around.

api/deploy/ocp/scc.go
Madhu-1: Yeah even not sure, but let's not panic and return the error message let the caller handle it?

Makefile
Madhu-1: sounds good :+1:

Madhu-1 (approved):

yati1998 (approved):

nixpanic:
There are now changes related to e2e, tests can be skipped.

PR #2556, opened 2021-10-05
rbd: detect krbd features in runtime and fallback to nbd

### LINKED ISSUE #478: Automatically attempt to use rbd-nbd when krbd cannot support RBD image features

If an image has features enabled that are not supported by krbd, the attacher should automatically attempt to use rbd-nbd. The end user shouldn't need to know when to set the `mounter` StorageClass attribute [1]

The implementation details (and limitations of krbd) will be hidden from the user, facilitating the use of the driver.

An image with features enabled that are not supported by krbd will still be properly attached to a pod. For example, if the `journaling` feature is enabled in support of RBD mirroring.

[1] https://github.com/ceph/ceph-csi/blob/master/examples/rbd/storageclass.yaml#L35

pkalever:

Currently, we recognize and warn for the provided image features based on our prior intelligence at ceph-csi (i.e based on supportedFeatures map and validateImageFeatures) at image/PV creation time. It might be very much possible that the cluster is heterogeneous i.e. the PV creation and application container might both be on different nodes with different kernel versions (krbd driver versions).

This PR adds a mechanism to check for the supported krbd features during mount time, if the krbd driver doesn't have the specified image feature then it will fall back to rbd-nbd mounter.

This PR also provides an option to disable the auto fallback to nbd mounter via storageclass `fallbackToNBD` parameter.

## Related issues ##

Fixes: #478 

---

<details>

  progress or results

</details>


nixpanic:
Where you mention "realtime", you mean "runtime". Please change the comments accordingly.

internal/rbd/rbd_util.go
```diff
@@ -210,6 +213,85 @@ var supportedFeatures = map[string]imageFeature{
...
+	if err != nil {
+		return supported, err
+	}
+
+	features, err := hexStringToInteger(attr)
+	if err != nil {
+		return supported, err
+	}
+
+	if strings.EqualFold(fname, librbd.FeatureNameLayering) {
```
nixpanic: Better not hard-code all features here, let go-ceph handle that with the `FeatureSet` object? Then use `fs.Names()` to get a list of all supported features to match against?

internal/rbd/rbd_util_test.go
```diff
@@ -283,3 +283,43 @@ func TestStrategicActionOnLogFile(t *testing.T) {
...
+			featureName: "xyz",
+			isSupported: false,
+			err:         librbd.ErrNotFound,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			supported, err := isKrbdFeatureSupported(tt.featureName)
```
nixpanic: This now is dependent on the system where the tests are running. If the rbd kernel module is not (and can not be) loaded, this will always fail. You have to think of an other way to test the functions, without depending on the rbd kernel module.

internal/rbd/nodeserver.go
```diff
@@ -127,6 +127,23 @@ func isStaticVolume(parameters map[string]string) bool {
 	return staticVol
 }
 
+// isFallbackToNBDRequest checks if fallbackToNBD is set or not.
+func isFallbackToNBDRequest(parameters map[string]string) bool {
+	tp := "fallbackToNBD"
```
nixpanic: what does `tp` stand for here?

internal/rbd/rbd_util_test.go
pkalever: I agree. Should we discard the unit test?

internal/rbd/rbd_util.go
pkalever: I had switched to switch case now. I will check on `FeatureSet`, thanks!

internal/rbd/rbd_util_test.go
pkalever: Or is there a way to `modprobe rbd` prior to running unit tests from the CI scripts?

nixpanic: Always have unit tests for helper functions. Maybe initialize the supported features once with a global variable, and set the variable to something pre-defined in the tests.

unit tests should be able to run as non-root while developing ceph-csi.

pkalever: Apologies, I didn't follow completely,
It is `getKrbdSupportedFeatures` job to get supported features (load rbd module, read the /sys file ...) now this unit test is for `isKrbdFeatureSupported`. How does defining the global variable for supported features is going to help? I mean how do I bypass the call to `getKrbdSupportedFeatures` ?

pkalever:
@Mergifyio rebase

pkalever:
@ceph/ceph-csi-contributors 
Is there any known issue with the CI?
1. the tests are failing at build artifacts itself
2. the go-test, go-sec, commitlint, codespell ...etc tests are not even got triggered

Madhu-1:
> @ceph/ceph-csi-contributors Is there any known issue with the CI?
> 
> 1. the tests are failing at build artifacts itself
> 2. the go-test, go-sec, commitlint, codespell ...etc tests are not even got triggered

Nothing am aware of

Madhu-1:
@mergifyio rebase

Madhu-1:
@pkalever jenkins is back again. can you please force push to start Github actions?

pkalever:
> @pkalever jenkins is back again. can you please force push to start Github actions?

Done now, thanks @Madhu-1

internal/rbd/rbd_util_test.go
nixpanic: yes, the tests can not rely on `getKrbdSupportedFeatures` (which only needs to be called once anyway). So, I'd suggest checking a global variable in `isKrbdFeatureSupported` (which could be returned by `getKrbdSupportedFeatures` if set earlier). Unit tests can then set the global variable, making `getKrbdSupportedFeatures` do (almost) nothing, and the test will work everywhere.

nixpanic:
```
=== CONT  TestIsKrbdFeatureSupported/supported_feature
W1006 06:27:58.852937    8434 rbd_util.go:230] modprobe failed with error an error (exec: "modprobe": executable file not found in $PATH) occurred while running modprobe args: [rbd]
    rbd_util_test.go:321: isKrbdFeatureSupported(layering) got(err) = false(an error (exec: "modprobe": executable file not found in $PATH) occurred while running modprobe args: [rbd]), expected(err) = layering(<nil>)
```
the **go-test** job fails due to this.


I suggest something like
```go
var krbdSupportedFeatures uint64

func getKrbdSupportedFeatures() (uint64, error) {
    if krbdSupportedFeatures != 0 {
        return krbdSupportedFeatures, nil
    }

    // detect features from /sys/... and set krbdSupportedFeatures
}

// unit testing
func TestIsKrbdFeatureSupported() {
    krbdSupportedFeatures = 0x12345

    supported := isKrbdFeatureSupported("layering")
}
```

internal/rbd/rbd_util.go
pkalever: Done now.

internal/rbd/rbd_util_test.go
pkalever: Sure, Thanks applied now.

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,106 @@ type migrationVolID struct {
...
+		librbd.FeatureNameExclusiveLock: {
+			needRbdNbd: true,
+		},
+		librbd.FeatureNameJournaling: {
+			needRbdNbd: true,
+			dependsOn:  []string{librbd.FeatureNameExclusiveLock},
+		},
+	}
+
+	krbdSupportedFeaturesAttr string
```
nixpanic: any reason to not have this as an uint or librbd.FeatureSet?

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,106 @@ type migrationVolID struct {
...
+		return supported, err
+	}
+
+	found := false
+	for _, featureName := range imageFeatureSet.Names() {
+		if !strings.EqualFold(fname, featureName) {
+			continue
+		}
+		found = true
+		if (uint64(imageFeatureSet) & features) != 0 {
```
nixpanic: this does not need to be inside the for-loop?

Maybe only list the features that are not supported in case there are any? Do not run the for-loop when all required features are supported.

You might be able to get the missing features with something like this:
```
librbd.FeatureSet(uint64(imageFeatureSet) & ~features).Names()
```

internal/rbd/rbd_util_test.go
```diff
@@ -283,3 +284,44 @@ func TestStrategicActionOnLogFile(t *testing.T) {
...
+		},
+		{
+			name:        "unrecognized feature",
+			featureName: "xyz",
+			isSupported: false,
+			err:         librbd.ErrNotFound,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
nixpanic: Please do not  overload the `tt` variable, it is very ugly and I hope other tests get cleaned up too. Use `tc := tt` and use `tc` afterwards, or something like that.

internal/rbd/rbd_util.go
pkalever: Well,  I'm treating this as something returnable from getKrbdSupportedFeatures().
`getKrbdSupportedFeatures` read attr from the /sys and return as a string which is later used by hexStringToInteger() to convert into unit.

If you want krbdSupportedFeaturesAttr as unit, then I should 

getKrbdSupportedFeatures() (uint64, error) {
  //read from /sys
  return hexStringToInteger(strAttr), nil
}

is that what you want?
I'm fine with the string, it's just a matter of code preference.

internal/rbd/rbd_util_test.go
pkalever: Done.

internal/rbd/rbd_util.go
pkalever: > librbd.FeatureSet(uint64(imageFeatureSet) & ~features).Names()

internal/rbd/rbd_util.go:288:21: invalid character U+007E '~'
internal/rbd/rbd_util.go:290:90: cannot call pointer method on rbd.FeatureSet(uint64(imageFeatureSet) & ~features)
internal/rbd/rbd_util.go:290:90: cannot take the address of rbd.FeatureSet(uint64(imageFeatureSet) & ~features)
make: *** [Makefile:161: cephcsi] Error 2

pkalever: After started using FeatureSet, I see there is some duplication in this function, I will cleanup and refresh tomorrow.

charts/ceph-csi-rbd/README.md
```diff
@@ -150,6 +150,7 @@ charts and their default values.
 | `storageClass.pool`                            | Ceph pool into which the RBD image shall be created                                                                                                  | `replicapool`                                      |
 | `storageClass.thickProvision`                  | Specifies whether thick provision should be enabled                                                                                                  | `false`                                            |
 | `storageclass.imageFeatures`                   | Specifies RBD image features                                                                                                                         | `layering`                                         |
+| `storageclass.fallbackToNBD`                   | Specifies whether to fallback to rbd-nbd mounter when imageFeatures are not found at the time of mount                                               | `true`                                             |
```
Madhu-1: it should be something called autofackback? if NBD is not supporting we can fall back to kbrd? not sure about the case it can happen but we can be safe?

pkalever:
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.21/detail/mini-e2e-helm_k8s-1.21/1438/pipeline

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.20/detail/mini-e2e_k8s-1.20/2787/pipeline


```


```

pkalever:
@Madhu-1 @nixpanic PTAL. Thanks!

internal/rbd/nodeserver.go
```diff
@@ -127,6 +127,23 @@ func isStaticVolume(parameters map[string]string) bool {
 	return staticVol
 }
 
+// isFallbackToNBDRequest checks if fallbackToNBD is set or not.
+func isFallbackToNBDRequest(parameters map[string]string) bool {
```
nixpanic: the 1st commit adds `isFallbackToNBDRequest`, but does not use it or has a unit-test for it

internal/rbd/nodeserver.go
```diff
@@ -296,6 +296,12 @@ func fallBackToNBD(ctx context.Context, imageFeatures string) (bool, error) {
 	if supported {
 		return false, nil
 	}
+
+	if !isFallbackToNBD {
+		log.ErrorLog(ctx, "unsupported krbd features in: %v, set `fallbackToNBD:true` or fix krbd driver", arr)
+
+		return false, librbd.ErrNotFound // not supported by krbd driver on host
```
nixpanic: better return a ErrNotSupported or similar, ErrNotFound is not much telling, and maybe even confusing when the previous error is logged

nixpanic (changes requested):
just a few nits, thanks!

internal/rbd/nodeserver.go
```diff
@@ -127,6 +127,23 @@ func isStaticVolume(parameters map[string]string) bool {
...
+		return true // default value
+	}
+
+	fallbackBool, err := strconv.ParseBool(fallback)
+	if err != nil {
+		return false
+	}
+
+	return fallbackBool
+}
```
Madhu-1:
```suggestion
func isFallbackToNBDRequest(parameters map[string]string) bool {
	fallback := true // default value
	optionName := "fallbackToNBD"

	if f, ok := parameters[optionName]; ok {
		fallback, _ = strconv.ParseBool(f) // discard error as we are not considering the errorcase here
	}
	return fallback
}
```

internal/rbd/nodeserver.go
```diff
@@ -321,11 +373,21 @@ func (ns *NodeServer) NodeStageVolume(
...
-	rv, err := populateRbdVol(ctx, req, cr)
+	useNBD := false
+	if req.GetVolumeContext()["mounter"] != rbdNbdMounter {
+		isFallbackToNBD := isFallbackToNBDRequest(req.GetVolumeContext())
+		useNBD, err = fallBackToNBD(ctx, imageFeatures, isFallbackToNBD)
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
Madhu-1: move this logic to populateRbdVol so we can avoid adding one more input parameter to the function?

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,97 @@ type migrationVolID struct {
...
+			dependsOn:  []string{librbd.FeatureNameExclusiveLock},
+		},
+	}
+
+	krbdSupportedFeaturesAttr string
+)
+
+// getKrbdSupportedFeatures load the module if needed and return supported
+// features attribute as a string.
+func getKrbdSupportedFeatures() (string, error) {
```
Madhu-1, line 222: we should avoid getting the supported features for each request as it can contribute to performance. get this once when starting the CSI driver and store it in a global variable?

internal/rbd/nodeserver.go
```diff
@@ -255,6 +280,33 @@ func populateRbdVol(
...
+	supported, err := isKrbdFeatureSupported(imageFeatureSet)
+	if err != nil {
+		log.ErrorLog(ctx, "failed fallBackToNBD(%v), err: %w", arr, err)
+
+		return false, err
+	}
+	if supported {
+		return false, nil
+	}
```
Madhu-1:
```suggestion
supported, err := isKrbdFeatureSupported(imageFeatureSet)
	if err != nil || !supported {
		return false, err
	}
```

Madhu-1 (changes requested):
apologies reviewed it a couple of days back but  forgot to post my review

internal/rbd/nodeserver.go
pkalever: Missed in juggling the code as part of addressing reviews. Will fix it.

internal/rbd/nodeserver.go
pkalever: Will take care

charts/ceph-csi-rbd/README.md
pkalever: In storage class, only one mounter is supported. Hence I would strongly feel, the fallback option should be per mounter type. If needed we can provide another option like `fallBackToKrbd` which should be used with `nbd` mounter.

In the future, I would like to see more cleanups for mounter-specific options for better clarification, for example: mapOptions naming specific to mounter type, like nbdMapOptions, etc ..

internal/rbd/nodeserver.go
pkalever: Sure

internal/rbd/nodeserver.go
pkalever: Makesense :+1:

internal/rbd/rbd_util.go
pkalever: This will be a good improvement, sure :+1:

internal/rbd/nodeserver.go
pkalever: Sure

charts/ceph-csi-rbd/README.md
Madhu-1: the main idea is to avoid adding more parameters to the storageclass.

pkalever: Ok. For krbd we check if the image features are supported and if not supported fallback to rbd-nbd, I'm not sure what would be the fallback motive from nbd to krbd? From the nodeplugin point of view when do we want to fallback from nbd to krbd?

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,85 @@ type migrationVolID struct {
...
+	// check if the module is loaded or compiled in
+	_, err := os.Stat(krbdSupportedFeaturesFile)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			log.ErrorLogMsg("stat on: %s failed, err: %v", krbdSupportedFeaturesFile, err)
```
nixpanic: nit: log errors without referring to the `err` variable. Suggestion: `stat on %q failed: %v`

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,85 @@ type migrationVolID struct {
...
+
+			return "", err
+		}
+		// try to load the module
+		_, _, err = util.ExecCommand(context.TODO(), "modprobe", rbdDefaultMounter)
+		if err != nil {
+			log.WarningLogMsg("modprobe failed with error %v", err)
```
nixpanic: same here, use something like `modprobe failed with: %v` instead

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,85 @@ type migrationVolID struct {
...
+
+			return "", err
+		}
+	}
+	val, err := ioutil.ReadFile(krbdSupportedFeaturesFile)
+	if err != nil {
+		log.ErrorLogMsg("ReadFile %s failed, err: %v", krbdSupportedFeaturesFile, err)
```
nixpanic: and here too. Note that `ReadFile` is a go function, and not user friendly phrased. Maybe `reading file %q failed: %v`?

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,85 @@ type migrationVolID struct {
...
+}
+
+// hexStringToInteger convert hex value to uint.
+func hexStringToInteger(hexString string) (uint64, error) {
+	// trim 0x prefix
+	numberStr := strings.TrimPrefix(strings.ToLower(hexString), "0x")
+
+	output, err := strconv.ParseUint(numberStr, 16, 64)
+	if err != nil {
+		log.ErrorLogMsg("strconv.ParseUint(%s) failed, err: %v", numberStr, err)
```
nixpanic: again, phrase it as something that makes sense to a user too

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,85 @@ type migrationVolID struct {
...
+func isKrbdFeatureSupported(imageFeatures string) bool {
+	arr := strings.Split(imageFeatures, ",")
+	log.DefaultLog("checking for ImageFeatures: %v", arr)
+	imageFeatureSet := librbd.FeatureSetFromNames(arr)
+
+	supported := true
+	for _, featureName := range imageFeatureSet.Names() {
+		if (uint64(librbd.FeatureSetFromNames(strings.Split(featureName, " "))) & krbdFeatures) == 0 {
+			supported = false
+			log.ErrorLogMsg("krbd feature: %s not supported", featureName)
```
nixpanic: use `%q` instead of `%s` so that the featurename gets quoted

internal/rbd/nodeserver.go
nixpanic: This would benefit from a unit-test. Our test coverage should get better, adding tests for small functions like this prevents more maintenance backlog.

nixpanic (commented):
just some nits, not much of real concerns

internal/rbd/nodeserver_test.go
```diff
@@ -56,3 +56,40 @@ func TestGetStagingPath(t *testing.T) {
...
+		{
+			name:         "fallbackToNBD => false",
+			scParameters: map[string]string{"fallbackToNBD": "false"},
+			expect:       false,
+		},
+		{
+			name:         "fallbackToNBD => empty",
+			scParameters: map[string]string{},
+			expect:       true, // default
+		},
```
nixpanic, line 96: Could have added a value that can not be converted to a bool, like `"sometimes"` and verify that the default is taken.

Also, an empty value (empty string) is different than a value that is not set at all.

nixpanic (dismissed):
I'm ok with this change, the unit-test for `isFallbackToNBDRequest()` could be improved.

The new option in the StorageClass is maybe not so great, but I also would not know what to suggest otherwise.

internal/rbd/nodeserver_test.go
pkalever: Hmm, sure I will take care. Thanks!

pkalever: @nixpanic I hope this addresses all the cases now. Thanks for the suggestion.

pkalever:
@nixpanic @Madhu-1 @ceph/ceph-csi-contributors 

Please suggest some generic naming for the currently proposed `fallbackToNBD` option. The intention is to reuse the user-facing SC option for priority-based switching between available mounters in the future. 

Please note, currently, we didn't identify the best case to switch from NBD to Krbd, I welcome your thoughts on suggestions. The NBD driver not loaded is much like a prerequisite task just like the KRBD driver case IMHO, hence I'm not considering it.


Here are some confusing names, please help me pick one.

May be `switchMounter` `autoSwitchMounter` `switchToOtherMounter` `fallbackToOtherMounter` `autoFallback` `fallbackMounter` `redirectMounter` `shiftMounter` :-)

Thanks!

Madhu-1:
Am fine with any generic name. `autoFallback` looks good to me.

nixpanic:
> Am fine with any generic name. `autoFallback` looks good to me.

:+1: or something like `tryNextMounter` or such

internal/rbd/rbd_util_test.go
```diff
@@ -283,3 +283,40 @@ func TestStrategicActionOnLogFile(t *testing.T) {
...
+			isSupported: false,
+		},
+	}
+	for _, tt := range tests {
+		tc := tt
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			var err error
+			krbdSupportedFeaturesAttr := "0x1"
+			krbdFeatures, err = hexStringToInteger(krbdSupportedFeaturesAttr) // initialize krbdFeatures
```
nixpanic, line 313: As `krbdFeatures` is a global variable, tests should probably not run in parallel.

pkalever:
> Am fine with any generic name. `autoFallback` looks good to me.

The only thing I'm not happy about this is that it doesn't refer to `mounter` anywhere, like autoFallback about what?

> +1 or something like tryNextMounter or such

And this looks good, with a minor change: `tryOtherMounters`

Thanks @Madhu-1 @nixpanic

charts/ceph-csi-rbd/templates/storageclass.yaml
```diff
@@ -15,6 +15,9 @@ parameters:
   pool: {{ .Values.storageClass.pool }}
   imageFeatures: {{ .Values.storageClass.imageFeatures }}
   thickProvision: {{ .Values.storageClass.thickProvision | quote}}
+{{- if .Values.storageClass.tryOtherMounters }}
+  tryOtherMounters: "{{ .Values.storageClass.tryOtherMounters }}"
```
Madhu-1: use `{{ .Values.storageClass.tryOtherMounters | quote}} ` to keep it similar to other options in SC

charts/ceph-csi-rbd/values.yaml
```diff
@@ -282,6 +282,16 @@ storageClass:
...
   imageFeatures: layering
 
+  # (optional) Specifies whether to try other mounters in case if the current
+  # mounter fails to mount the rbd image for any reason. True means fallback
+  # to next mounter (default)
+  # Note: tryOtherMounters is currently useful to fallback from krbd to rbd-nbd
+  # in case if any of the specified imageFeatures is not supported by krbd
+  # driver on node scheduled for application pod launch, but in the future this
+  # should work with any mounter type.
+  # tryOtherMounters: true
```
Madhu-1: IMO we should removed it from line 293 and keep it as optional one not to enable it by default for now

internal/rbd/nodeserver.go
```diff
@@ -128,6 +128,21 @@ func isStaticVolume(parameters map[string]string) bool {
 	return staticVol
 }
 
+// isTryOtherMountersRequest checks if tryOtherMounters is set or not.
+func isTryOtherMountersRequest(parameters map[string]string) bool {
+	fallback := true // default value
```
Madhu-1: the default should be false? and try other mounters only if it's requested?

internal/rbd/nodeserver.go
```diff
@@ -222,10 +237,28 @@ func populateRbdVol(
 		rv.RbdImageName = imageAttributes.ImageName
 	}
 
+	krbdSupported := false
+	if req.GetVolumeContext()["mounter"] != rbdNbdMounter {
+		krbdSupported = isKrbdFeatureSupported(req.GetVolumeContext()["imageFeatures"])
```
Madhu-1: @pkalever is it possible to log with the context in isKrbdFeatureSupported or return an error message from it saying which feature is not supported? this is because logging with context to identify the error belongs to which request.

internal/rbd/nodeserver.go
```diff
@@ -222,10 +237,28 @@ func populateRbdVol(
...
+		if !krbdSupported && !isTryOtherMountersRequest(req.GetVolumeContext()) {
+			log.ErrorLog(ctx, "unsupported krbd Feature, set `tryOtherMounters:true` or fix krbd driver")
+			err = fmt.Errorf("unsupported krbd Feature")
```
Madhu-1:
```suggestion
return nil, status.Errorf(codes.Internal, "unsupported krbd Feature")
```

Madhu-1 (commented):
LGTM left some questions and suggestions.

charts/ceph-csi-rbd/values.yaml
pkalever: Madhu, I remember we discussed this before and decided to keep it true by default.

I request that this should be the way to promote NBD use case as KRBD is the default now. Very few of the users are aware of the backend mounter and there should be a very very minimal number of users who do not want to use NBD.

internal/rbd/nodeserver.go
pkalever: As mentioned above, I would strongly feel and request to keep it true by default. We have given an option just to turn-off if someone really don't like NBD in anycase.

internal/rbd/nodeserver.go
pkalever: Logging with context: yes
return feature-not-supported: why? we already log it somewhere and I hope that should be good enough, won't it?

Madhu-1: i meant to use https://github.com/ceph/ceph-csi/blob/devel/internal/util/log/log.go#L69 because as cephcsi can get many requests it becomes difficult to identify the error message that belongs to which request if it is not logged with context/request-id

internal/rbd/nodeserver.go
Madhu-1: again this will be a new behavior change right? IMO we should not make it default until nbd is production-ready in cephcsi in other words supported is GA'ed in cephcsi.

internal/rbd/nodeserver.go
pkalever: Got it, I will rather supply ctx to isKrbdFeatureSupported instead of returning an un-supported feature.

Thanks Madhu.

internal/rbd/nodeserver.go
pkalever: @Madhu-1 hmmm, yes seems like a valid concern for me with alpha support of nbd. 

@nixpanic do you got any further thoughts on this? 

Thanks.

pkalever: @Madhu-1 I'm convinced on this, have submitted a respin with tryOtherMounters: false by default. PTAL.

Thanks!

charts/ceph-csi-rbd/templates/storageclass.yaml
```diff
@@ -15,6 +15,9 @@ parameters:
   pool: {{ .Values.storageClass.pool }}
   imageFeatures: {{ .Values.storageClass.imageFeatures }}
   thickProvision: {{ .Values.storageClass.thickProvision | quote}}
+{{- if .Values.storageClass.tryOtherMounters }}
+  tryOtherMounters: "{{ .Values.storageClass.tryOtherMounters | quote}}"
```
Madhu-1: `"{{ .Values.storageClass.tryOtherMounters | quote}}"` to `{{ .Values.storageClass.tryOtherMounters | quote}}`

charts/ceph-csi-rbd/values.yaml
```diff
@@ -282,6 +282,16 @@ storageClass:
...
 
+  # (optional) Specifies whether to try other mounters in case if the current
+  # mounter fails to mount the rbd image for any reason. True means fallback
+  # to next mounter, default is set to false.
+  # Note: tryOtherMounters is currently useful to fallback from krbd to rbd-nbd
+  # in case if any of the specified imageFeatures is not supported by krbd
+  # driver on node scheduled for application pod launch, but in the future this
+  # should work with any mounter type.
+  # tryOtherMounters: false
+  tryOtherMounters: false
```
Madhu-1:
```suggestion
tryOtherMounters: false
```

internal/rbd/nodeserver.go
```diff
@@ -128,6 +128,21 @@ func isStaticVolume(parameters map[string]string) bool {
...
 
+// isTryOtherMountersRequest checks if tryOtherMounters is set or not.
+func isTryOtherMountersRequest(parameters map[string]string) bool {
+	fallback := false // default value
+	optionName := "tryOtherMounters"
+
+	if f, ok := parameters[optionName]; ok {
+		var err error
+		if fallback, err = strconv.ParseBool(f); err != nil {
+			return false // for cases like, not parsable or empty string
```
Madhu-1: log the error for better debugging? can you do break or return false. using break will avoid extra return from here. i will leave it to you

internal/rbd/rbd_util.go
```diff
@@ -197,17 +200,85 @@ type migrationVolID struct {
...
+	if err != nil {
+		if !os.IsNotExist(err) {
+			log.ErrorLogMsg("stat on %q failed: %v", krbdSupportedFeaturesFile, err)
+
+			return "", err
+		}
+		// try to load the module
+		_, _, err = util.ExecCommand(context.TODO(), "modprobe", rbdDefaultMounter)
+		if err != nil {
+			log.WarningLogMsg("modprobe failed with: %v", err)
```
Madhu-1:
```suggestion
log.ErrorLogMsg("modprobe failed: %v", err)
```

Madhu-1 (commented):
few nits. LGTM

charts/ceph-csi-rbd/values.yaml
pkalever: If you haven't paid attention, most of the other options in this file have examples @Madhu-1 . Which is why I just kept it this way. If you want me to still remove this, I will add a cleanup patch for others first and then rebase this patch, let me know which is good.

Madhu-1: yeah, I noticed in other places also. IMO either it should be commented or set by in value.yaml not both. anything is fine for me

pkalever: commented will anyway means false, as it is not mandatory (optional) will leave it commented. Thanks!

internal/rbd/nodeserver.go
pkalever: I will send a change adding log, but break is not possible as it is not within a loop or switch

Madhu-1 (dismissed):

pkalever:
@Madhu-1 I have moved unit tests for parseBoolOption() into the commit where we introduced it. I guess that is the only change. Could you please replace your ack?

Thanks!

charts/ceph-csi-rbd/values.yaml
```diff
@@ -289,6 +289,15 @@ storageClass:
...
   imageFeatures: layering
 
+  # (optional) Specifies whether to try other mounters in case if the current
+  # mounter fails to mount the rbd image for any reason. True means fallback
+  # to next mounter, default is set to false.
+  # Note: tryOtherMounters is currently useful to fallback from krbd to rbd-nbd
+  # in case if any of the specified imageFeatures is not supported by krbd
+  # driver on node scheduled for application pod launch, but in the future this
+  # should work with any mounter type.
+  # tryOtherMounters: false
```
humblec, line 299: @pkalever its always from krbd to nbd, Isnt it ?  not from nbd to krbd ?

humblec: if above is the case we could have actually stayed with better `values` for this parameter?  may be a string which talks which mounters to be tried like `trynbd, tryblah,.etc` , if its always from krbd to nbd , I doubt "tryOthermounters" make sense, imo, fallbacktoNbd was better.

pkalever: @humblec Currently, it is krbd to nbd. But as this is a user-facing option it kept generic to provide the flexibity in such a way that we can use the same to fallback from nbd to krbd too without any changes to storageClass. But for that we need to identify in which cases we should fallback from nbd to krbd, we have not come across anything for now.

Madhu-1 (approved):

nixpanic (approved):

PR #2557, opened 2021-10-06
rbd: use clusterID from the volumeContext when handling nodestage request for migration

humblec:
previously we were retriving clusterID using the monitors field
in the volume context at node stage code path. however it is possible to
retrieve clusterID from the volumeContext passed with  nodeStage request

Additional Info:
https://github.com/ceph/ceph-csi/pull/2523#issuecomment-931130464
    

internal/rbd/nodeserver.go
```diff
@@ -291,13 +273,19 @@ func (ns *NodeServer) NodeStageVolume(
...
-		cID, cErr := getClusterIDFromMigrationVolume(req.GetVolumeContext()["monitors"])
-		if cErr != nil {
-			return nil, status.Error(codes.Internal, cErr.Error())
+	// it will be missing the clusterID, so fill it by fetching it from config file using monitors hash.
+	if isMigrationContext(req.GetVolumeContext()) {
+		// verify and parse volID
+		if isMigrationVolID(volID) {
+			parsedMigHandle, pErr := parseMigrationVolID(volID)
+			if pErr != nil {
+				log.ErrorLog(ctx, "failed to parse the migration handle, err :%v ", pErr)
```
nixpanic: just use `failed to parse the migration handle: %v` as error format, the last component after `:` is expected to be the error message

internal/rbd/migration_test.go
```diff
@@ -171,3 +171,42 @@ func TestParseMigrationVolID(t *testing.T) {
...
+			false,
+		},
+		{
+			"nil clusterID set",
+			map[string]string{"migrtion": "true", "clusterID": ""},
+			true,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
nixpanic: I really hate that we're doing it this way. It is very ugly to overload a `tt` variable like this. Please don't introduce these kind of things for new tests, use a different variable instead.

humblec: well, its just a variable name' ? as long as "memory" is concerned it is totally different location :). No hard opinion to change the variable name though. can be changed.

internal/rbd/nodeserver.go
humblec: done

internal/rbd/nodeserver.go
```diff
@@ -291,13 +273,19 @@ func (ns *NodeServer) NodeStageVolume(
...
-		cID, cErr := getClusterIDFromMigrationVolume(req.GetVolumeContext()["monitors"])
-		if cErr != nil {
-			return nil, status.Error(codes.Internal, cErr.Error())
+	// it will be missing the clusterID, so fill it by fetching it from config file using monitors hash.
+	if isMigrationContext(req.GetVolumeContext()) {
+		// verify and parse volID
+		if isMigrationVolID(volID) {
+			parsedMigHandle, pErr := parseMigrationVolID(volID)
+			if pErr != nil {
+				log.ErrorLog(ctx, "failed to parse the migration handle: %v ", pErr)
```
nixpanic: drop the trailing space at `%v "`

nixpanic (commented):
I don't have any concerns with this change. It is not clear to me why this is better or what bug it addresses. That could be explained a little more in the commit message.

internal/rbd/nodeserver.go
humblec: done.

humblec:
> I don't have any concerns with this change. It is not clear to me why this is better or what bug it addresses. That could be explained a little more in the commit message.

my bad, this is actually a small change which marked as a followup item https://github.com/ceph/ceph-csi/pull/2523#issuecomment-931130464.. Added this information in the PR description too. sorry for missing it in first place @nixpanic

nixpanic (dismissed):

humblec:
@Mergifyio rebase

yati1998 (dismissed):

humblec:
@Mergifyio rebase

Madhu-1 (dismissed):
Please remove GetClusterIDFromMon and   readClusterInfoWithMon which is not required anymore https://github.com/ceph/ceph-csi/pull/2512/files#diff-b9c67ae3efee5ec6b2d1889405ffaf967a4808f3d9a12a287ac6faa9ba19d762R169-R207 .
Please also make required changes in E2E to remove monitor information from PV.CSI  object as  its not required anymore.

internal/rbd/nodeserver.go
```diff
@@ -292,13 +274,19 @@ func (ns *NodeServer) NodeStageVolume(
...
-		cID, cErr := getClusterIDFromMigrationVolume(req.GetVolumeContext()["monitors"])
-		if cErr != nil {
-			return nil, status.Error(codes.Internal, cErr.Error())
+	// it will be missing the clusterID, so fill it by fetching it from config file using monitors hash.
+	if isMigrationContext(req.GetVolumeContext()) {
+		// verify and parse volID
+		if isMigrationVolID(volID) {
+			parsedMigHandle, pErr := parseMigrationVolID(volID)
+			if pErr != nil {
+				log.ErrorLog(ctx, "failed to parse the migration handle: %v", pErr)
```
Madhu-1: `failed to parse the migration handle`  to `failed to parse volumeID`

internal/rbd/nodeserver.go
```diff
@@ -292,13 +274,19 @@ func (ns *NodeServer) NodeStageVolume(
...
+
+				return nil, status.Error(codes.Internal, pErr.Error())
+			}
+			// store clusterID
+			req.VolumeContext[util.ClusterIDKey] = parsedMigHandle.clusterID
```
Madhu-1: as you already know the clusterID can't we just add that to volumeContext in kubernetes itself?

Madhu-1 (commented):
@humblec sorry for  it late, i have a question PTAL

humblec:
> Please remove GetClusterIDFromMon and readClusterInfoWithMon which is not required anymore https://github.com/ceph/ceph-csi/pull/2512/files#diff-b9c67ae3efee5ec6b2d1889405ffaf967a4808f3d9a12a287ac6faa9ba19d762R169-R207 . Please also make required changes in E2E to remove monitor information from PV.CSI object as its not required anymore.

@Madhu-1 I kept that out there in case if we want to use the same in future , but I can remove that as well.  Done.

internal/rbd/nodeserver.go
humblec: @Madhu-1 its possible on both ways, however kept this way as volID considered to be CSI  handler which can be depended for the operations.

Madhu-1: in nodeStage operation, we retrieve the clusterID from the context. IMO we should take the same route as it's possible now.

humblec: @Madhu-1 if really needed I can do that, wdyt?  why I am in bit hesitation here is, it also need changes in the migration Context func ...etc. Can that be done later? I can mark it as todo for now . Is that fine ?

Madhu-1: instead of adding complexity am more inclined to do this as it's going to be minimal change.

humblec: well, this patch submission was done to use mons hash as in the previous PR review comment, now the change request is actually different :).

Madhu-1: @humblec I do not always think about all the possibilities sometimes we suggesting things and when actually reviewing PR's sometimes I will try to think about the possibilities. if we get good approach IMO not a bad to change in between instead of changing later :)

humblec: `@humblec I do not always think about all the possibilities sometimes we suggesting things and when actually reviewing PR's sometimes I will try to think about the possibilities. if we get good approach IMO not a bad to change in between instead of changing later :)` - hmmm..as mentioned both works, any way addressing this comment then.

humblec:
`Error from server: error when creating "https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/v4.0.0/client/config/crd/snapshot.storage.k8s.io_volumesnapshotcontents.yaml": etcdserver: request timed out
script returned exit code 1`

humblec:
I have to adjust the commit message and little bit of splitting the last commit.. on it

humblec:
> I have to adjust the commit message and little bit of splitting the last commit.. on it

Done..

humblec:
@Mergifyio rebase

humblec:
Can you please review this ?

yati1998 (commented):
LGTM. Will approve once any of the maintainers approve it.

internal/rbd/migration_test.go
```diff
@@ -156,17 +156,56 @@ func TestParseMigrationVolID(t *testing.T) {
...
+		},
+		{
+			"clusterID set",
+			map[string]string{"migration": "true", "clusterID": "something"},
+			true,
+		},
+		{
+			"nil clusterID set",
+			map[string]string{"migration": "true", "clusterID": ""},
+			true,
```
Madhu-1: this test seems redundant as there is no clusteID validation in `isMigrationContext` can you please remove this test?

internal/rbd/nodeserver.go
```diff
@@ -291,14 +273,14 @@ func (ns *NodeServer) NodeStageVolume(
...
-		cID, cErr := getClusterIDFromMigrationVolume(req.GetVolumeContext()["monitors"])
-		if cErr != nil {
-			return nil, status.Error(codes.Internal, cErr.Error())
+	// Check this is a migration request and verify clusterID in the volumeContext.
+	if isMigrationContext(req.GetVolumeContext()) {
+		// verify and parse volID.
+		if req.GetVolumeContext()[util.ClusterIDKey] == "" {
+			log.ErrorLog(ctx, "clusterID is nil in the volume context")
+
+			return nil, status.Error(codes.InvalidArgument, "missing required parameter clusterID")
```
Madhu-1: IMO this check can be removed as the clusterID validation will be done when populating the rbdVolume just like other NodeStageRequest, we don't need to tread this as special

Madhu-1 (commented):
Currently, we have 3 e2e tests and we can move the duplicate test for Block and See if you can adjust other tests to validate Image deletion.

* validate RBD migration+static Block PVC Deletion
* validate RBD migration+static FileSystem PVC
* validate RBD migration+static Block PVC

internal/rbd/nodeserver.go
humblec: Thought an extra check here is good for fail early, but removing it.

humblec:
@Madhu-1 addressed the final comments. ptal.. thanks..

internal/rbd/migration.go
```diff
@@ -32,6 +32,12 @@ func isMigrationVolID(volHash string) bool {
 		strings.Contains(volHash, migImageNamePrefix) && strings.Contains(volHash, migMonPrefix)
 }
 
+// isMigrationContext validate the passed in request volumeContext/map carry
+// migration specific parameters.
+func isMigrationContext(vc map[string]string) bool {
+	return vc[intreeMigrationKey] == intreeMigrationLabel
+}
```
Madhu-1: this is not required anymore and can be removed?

humblec: @Madhu-1 done. ptal

Madhu-1 (approved):

yati1998 (approved):

humblec:
@Mergifyio rebase

PR #2569, opened 2021-10-14
e2e: correct secretnames in the SC for radosNamespace test

humblec:
eventhough the secrets are created with a different user , the custom
parameters are set with the wrong secrets. This patch corrects the same

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -2405,6 +2405,7 @@ var _ = Describe("RBD", func() {
 				}
 
 				updateConfigMap("e2e-ns")
+				radosNamespace = "e2e-ns"
 				// create rbd provisioner secret
 				key, err := createCephUser(
 					f,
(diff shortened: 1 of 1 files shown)


e2e/rbd.go
```diff
@@ -2438,11 +2438,11 @@ var _ = Describe("RBD", func() {
 				param := make(map[string]string)
 				// override existing secrets
 				param["csi.storage.k8s.io/provisioner-secret-namespace"] = cephCSINamespace
-				param["csi.storage.k8s.io/provisioner-secret-name"] = rbdProvisionerSecretName
+				param["csi.storage.k8s.io/provisioner-secret-name"] = rbdNamespaceProvisionerSecretName
```
nixpanic, line 2442: this is confusing, is the `cephCSINamespace` always the same as the `rbdNamespace`?

humblec: These custom secrets are created under `CephCSINamespace` so the namspace in SC has been set , and  these secrets carry custom users.  

Which `rbdNamespace` value you are referring here @nixpanic ? in e2e we only have below namespaces:
```
       cephCSINamespace string
	rookNamespace    string
	radosNamespace   string
```

nixpanic: I'm referring to the 1st part of `rbdNamespaceProvisionerSecretName`. It seems confusing to expect the secret named `rbdNamespaceProvisionerSecretName` to be located in the `cephCSINamespace` namespace.

Madhu-1: rbdNamespace refers to the rados rbd namespace in the ceph cluster and cephCSINamespace is kubernetes namespace.

Madhu-1 (dismissed):

e2e/rbd.go
humblec: > I'm referring to the 1st part of `rbdNamespaceProvisionerSecretName`. It seems confusing to expect the secret named `rbdNamespaceProvisionerSecretName` to be located in the `cephCSINamespace` namespace.

yeah, I agree to an extent that , it could have named some more better. More apted string was 'radosNamespace.....' may be in this context. I can change that if really required though. 

@nixpanic @Madhu-1 wdyt?

humblec:
hmm.. looks like some more changes are required here. will revisit those.. 
````
E1014 08:28:12.741061  158599 rbd_util.go:658] ID: 4 Req-ID: 0001-0024-58194f6b-c88c-49dd-81c1-070cd66699e6-0000000000000002-a1082954-2cc8-11ec-90e9-3aea6ce63e2b failed to check depth on image replicapool/csi-vol-a1082954-2cc8-11ec-90e9-3aea6ce63e2b: rbd: ret=-1, Operation not permitted
```

e2e/rbd.go
Madhu-1: Yeah that looks better

nixpanic: right, `rbdNamespaceProvisionerSecretName` :arrow_right: `radosNamespaceProvisionerSecretName` makes things clearer

nixpanic: just wondering if this is still going to get addressed...

PR #2572, opened 2021-10-14
rbd: check local image description for split-brain

Madhu-1:
In some corner cases like `re-player shutdown,` the local image will not be in an error state. It would
be also worth considering the `description` field to make sure about split-brain.

Note:- This extra check is suggested by the rbd team in a BZ https://bugzilla.redhat.com/show_bug.cgi?id=2012143 to cover a corner case

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -634,8 +634,7 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		ready = checkRemoteSiteStatus(ctx, mirrorStatus)
 	}
 
-	// resync only if the image is in error state
-	if strings.Contains(localStatus.State.String(), string(errorState)) {
+	if resyncRequired(localStatus) {
 		err = rbdVol.resyncImage()
 		if err != nil {
 			log.ErrorLog(ctx, err.Error())
@@ -663,4 +662,21 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	}
 
 	return resp, nil
+}
+
+// resyncRequired returns true if local image is in split-brain state and image
+// needs resync.
+func resyncRequired(localStatus librbd.SiteMirrorImageStatus) bool {
+	// resync is required if the image is in error state or the description
+	// contains split-brain message.
+	// In some corner cases like `re-player shutdown` the local image will not
+	// be in an error state. It would be also worth considering the `description`
+	// field to make sure about split-brain.
+	splitBrain := "split-brain"
+	if strings.Contains(localStatus.State.String(), string(errorState)) ||
+		strings.Contains(localStatus.Description, splitBrain) {
+		return true
+	}
+
+	return false
 }


Madhu-1:
>X Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute

>Oct 14 13:25:04.332: FAIL: failed to write IO, err: command terminated with exit code 2, stdErr: /bin/sh: 1: cannot create /var/lib/www/html/test: Read-only file system


Tests are failing for the above reason not related to this change. will retest/rebase later

ShyamsundarR (dismissed):
This change by itself is fine.

Further observations are as follows,
- We should report no errors from resync iff the state from RBD is "resyncing"
- For example if state from RBD is "starting_replay" we would report nil error and !ready, which will result in VolRep reporting Resyncing as true
- We need to figure out from RBD as to which states are the ones for which we can report Resync as progressing (but not ready, as we know the ready state), and report it only when we reach those states

The above needs some input from RBD mirroring for clarity and possibly can be a separate PR.

@Madhu-1 do we want a separate issue to discuss with RBD engineers, or should we leverage this PR?

Madhu-1:
> The above needs some input from RBD mirroring for clarity and possibly can be a separate PR.

Yes, that needs discussion with the RBD team, will take it separately.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -634,8 +634,13 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
 	}
 
-	// resync only if the image is in error state
-	if strings.Contains(localStatus.State.String(), string(errorState)) {
+	// resync if the image is in error state or the description contains
+	// split-brain message.
+	// In some corner cases like `re-player shutdown` the local image will not
+	// be in an error state. It would be also worth considering the `description`
+	// field to make sure about split-brain.
+	if strings.Contains(localStatus.State.String(), string(errorState)) ||
```
nixpanic: I think it makes sense to move this tot a small helper function, maybe `localStatus.ResyncNeeded()` or `localStatus.MayResync()`. This reduces the inline comments and hence size+complexity of the function.

Madhu-1: @nixpanic added a new helper function PTAL

yati1998 (approved):
LGTM

internal/rbd/replicationcontrollerserver.go
```diff
@@ -664,3 +663,20 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+// resyncRequired returns true if local image is in split-brain state and image
+// needs resync.
+func resyncRequired(localStatus librbd.SiteMirrorImageStatus) bool {
+	// resync is required if the image is in error state or the description
+	// contains split-brain message.
+	// In some corner cases like `re-player shutdown` the local image will not
+	// be in an error state. It would be also worth considering the `description`
+	// field to make sure about split-brain.
+	splitBrain := "split-brain"
+	if strings.Contains(localStatus.State.String(), string(errorState)) ||
```
nixpanic, line 676: I would expect some linter to complain about this if-return-true-else-return-false kind of formatting. A single return statement would be shorter and not more difficult to read.

nixpanic (approved):

internal/rbd/replicationcontrollerserver.go
Madhu-1: No linter complain anything related to that.  if CI fails will introduce a variable to hold the value and will have single return

Madhu-1:

failed to deploy ceph [cluster](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/k8s-e2e-external-storage-1.22/detail/k8s-e2e-external-storage-1.22/241/pipeline)

Madhu-1:

`X Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute`

PR #2170, opened 2021-06-16
testing: pass --delete-on-failure when starting minikube

nixpanic:
Sometimes there is an unclear error while starting the Minikube VM.
Possibly this can be worked around by passing the --delete-on-failure
option:

    If set, delete the current cluster if start fails and try again. Defaults to false.

See-also: https://minikube.sigs.k8s.io/docs/commands/start/

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)
- `/retest all`: run this in case the CentOS CI failed to start/report any test
  progress or results

</details>

PR #2577, opened 2021-10-18
e2e: do not delete the namespace and contents after a failure

nixpanic:
When a failure occurs, by default the test namespace is removed. This
makes it impossible to fetch the logs of the containers where the
failure was discovered. Pass --delete-namespace-on-failure=false as an
additional argument to the `run-e2e` make target, so that the namespace
is kept.

---

<details>

  progress or results

</details>

### DIFF
--- a/mini-e2e-helm.groovy
+++ b/mini-e2e-helm.groovy
@@ -180,7 +180,7 @@ node('cico-workspace') {
 		}
 		stage('run e2e') {
 			timeout(time: 120, unit: 'MINUTES') {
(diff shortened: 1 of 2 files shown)


Yuggupta27 (approved):

Yuggupta27:
We also sometimes see failures in `minikube start` during a CI run; I see a flag where minikube can re-run in case of a failure during `minikube start`. 
Flag: `--delete-on-failure` , source: https://minikube.sigs.k8s.io/docs/commands/start/#options

Should we add it? What do you think?

nixpanic:
> We also sometimes see failures in `minikube start` during a CI run; I see a flag where minikube can re-run in case of a failure during `minikube start`. Flag: `--delete-on-failure` , source: https://minikube.sigs.k8s.io/docs/commands/start/#options

We already have that, see #2170

Madhu-1 (approved):

Issue #2579, opened 2021-10-18
Cephfs `kubelet_volume_stats_used_bytes` only updates on pod start


bbatha:
# Describe the bug #

When using cephfs volumes I would like for the `kubelet_volume_stats` metrics especially `kubelet_volume_stats_used_bytes` to update periodically when the pod is running. Currently it only updates when a pod mounting the volume starts. This makes monitoring disk full conditions impossible.

# Environment details #

- Image/version of Ceph CSI driver : 3.4.0
- Helm chart version : 3.4.0
- Kernel version : `5.4.0-65-generic #73~18.04.1-Ubuntu`
- Mounter used for mounting PVC (for cephFS its `fuse` or `kernel`. for rbd its
  `krbd` or `rbd-nbd`) : kernel
- Kubernetes cluster version : 1.20
- Ceph cluster version : Octopus

# Steps to reproduce #

1. Create a pod mounting a cephfs pvc. Check prometheus for metrics.
1. Add data to the volume. Prometheus doesn't update
1. Delete pod and recreate. Prometheus is updated

# Actual results #

`kubelet_volume_stats_used_bytes` remains constant for the lifetime of the pod

# Expected behavior #

`kubelet_volume_stats_used_bytes` updates with the current usage of the volume. This could be a polled value.

# Logs #

```
I1018 18:32:55.481991       1 utils.go:176] ID: 500 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I1018 18:32:55.482111       1 utils.go:180] ID: 500 GRPC request: {"volume_id":"0001-001a-gradient-processing-shared-0000000000000001-bf9e449d-eb3e-11eb-9032-a6458f361b6e","volume_path":"/var/lib/kubelet/pods/81e44048-484d-45ff-ba47-49f4684094b2/volumes/kubernetes.io~csi/pvc-71c4a7ec-32fa-4a1a-b2e0-95631dd77df1/mount"}
I1018 18:32:55.493996       1 mount_linux.go:173] Cannot run systemd-run, assuming non-systemd OS
I1018 18:32:55.494044       1 mount_linux.go:174] systemd-run failed with: exit status 1
I1018 18:32:55.494067       1 mount_linux.go:175] systemd-run output: System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to create bus connection: Host is down
[...]

PR #2586, opened 2021-10-19
util: NodeGetVolumeStatsResponse.Usage may not contain negative values

nixpanic:
Following the CSI specification, values that are included in the
VolumeUsage MUST NOT be negative. However, CephFS seems to return -1 for
the number of inodes that are available. Instead of returning a
negative value, set it to 0 so that it will not get included in the
encoded JSON response.

Updates: #2579
See-also: https://github.com/container-storage-interface/spec/blob/5b0d4540158a260cb3347ef1c87ede8600afb9bf/spec.md?plain=1#L2477-L2487

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/csi-common/utils.go
+++ b/internal/csi-common/utils.go
@@ -263,17 +263,30 @@ func FilesystemNodeGetVolumeStats(ctx context.Context, targetPath string) (*csi.
 	return &csi.NodeGetVolumeStatsResponse{
 		Usage: []*csi.VolumeUsage{
 			{
-				Available: available,
-				Total:     capacity,
-				Used:      used,
+				Available: requirePositive(available),
+				Total:     requirePositive(capacity),
+				Used:      requirePositive(used),
 				Unit:      csi.VolumeUsage_BYTES,
 			},
 			{
-				Available: inodesFree,
-				Total:     inodes,
-				Used:      inodesUsed,
+				Available: requirePositive(inodesFree),
+				Total:     requirePositive(inodes),
+				Used:      requirePositive(inodesUsed),
 				Unit:      csi.VolumeUsage_INODES,
 			},
 		},
 	}, nil
+}
+
+// requirePositive returns the value for `x` when it is greater or equal to 0,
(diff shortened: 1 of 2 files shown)


internal/csi-common/utils.go
```diff
@@ -263,17 +263,30 @@ func FilesystemNodeGetVolumeStats(ctx context.Context, targetPath string) (*csi.
...
+				Total:     requirePositive(inodes),
+				Used:      requirePositive(inodesUsed),
 				Unit:      csi.VolumeUsage_INODES,
 			},
 		},
 	}, nil
 }
+
+// requirePositive returns the value for `x` when it is greater or equal to 0,
+// or returns 0 in the acse `x` is negative.
```
humblec, line 282: case..

yati1998 (approved):

Madhu-1 (approved):

Rakshith-R:


    Oct 20 06:11:16.357: failed to create PVC with error failed to get pvc: rpc error: code = Unknown desc = OK: HTTP status code 200; transport: missing content-type field

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.20/detail/mini-e2e-helm_k8s-1.20/2844/pipeline

humblec:
>However, CephFS seems to return -1 for
the number of inodes that are available. Instead of returning a
negative value

@nixpanic is this an expected behaviour or a bug or issue at CephFS side ? do we need to follow up with CephFS folks on that ?

nixpanic:
> > However, CephFS seems to return -1 for
> > the number of inodes that are available. Instead of returning a
> > negative value
> 
> @nixpanic is this an expected behaviour or a bug or issue at CephFS side ? do we need to follow up with CephFS folks on that ?

Not all filesystems provide available inodes, they can be dynamically allocated on-demand and not included in the pre-created (mkfs time) metadata of the filesystem. This is not a bug, but rather a feature of some filesystems (they can never run out of inodes).

PR #2589, opened 2021-10-25
rbd: note that thick-provisioning is deprecated

nixpanic:
Thick-provisioning was introduced to make accounting of assigned space
for volumes easier. When thick-provisioned volumes are the only consumer
of the Ceph cluster, this works fine. However, it is unlikely that this
is the case. Instead, accounting of the requested (thin-provisioned)
size of volumes is much more practical as different types of volumes can
be tracked.

OpenShift already provides cluster-wide quotas, which can combine
accounting of requested volumes by grouping different StorageClasses.

In addition to the difficult practise of allowing only thick-provisioned
RBD backed volumes, the performance makes thick-provisioning
troublesome. As volumes need to be completely allocated, data needs to
be written to the volume. This can take a long time, depending on the
size of the volume. Provisioning, cloning and snapshotting becomes very
much noticeable, and because of the additional time consumption, more
prone to failures.

---

<details>

  progress or results

</details>


internal/rbd/controllerserver.go
```diff
@@ -1517,5 +1521,11 @@ func isThickProvisionRequest(parameters map[string]string) bool {
 		return false
 	}
 
+	if logThickProvisioningDeprecation {
```
Madhu-1, line 1524: @nixpanic is it good to log when starting the csi driver(if we do so we don't need to cover create, expand etc?) ? as we are logging only once not sure it will help if the user if he misses this one (when having more PVC's operations going on).

nixpanic: I am not sure there is any value in logging more than once. We could also do logging every N-th time, or based on a timeout if you prefer something like that?

Not all users configure thick-provisioning (it is not enabled by default), so logging it when starting the CSI-driver is not suitable IMHO and might even cause confusion.

Madhu-1: nothing coming to my mind as it's a tricky one. am fine with the current change. @ceph/contributors any suggestions?

pkalever: It won't matter to me much if we log the deprecation msg about thick-provisioning even if one doesn't set it currently (actually it might help someone who has plans to use it in the future)

Hence I would rather just log it at driver initialization time.

nixpanic: I prefer to not log it by default, users might assume it is enabled globally somehow and they need to take action. This would be incorrect, as nobody is expected to use the feature.

pkalever: I mean a log like:

"thick-provisioning is **deprecated** and will be removed in a future release, this is for your information in case if you are using it currently or planing to use it in the future."

.. will not hurt.

humblec:
`Thick-provisioning was introduced to make accounting of assigned space
for volumes easier. `

 In  general thick provisiioning helps to reduce storage latency..etc, and a useful thing to have in use cases like vm templating.etc. so may be good to add that as well to the description.   also a small description about the issues/problems we encountered ( with pointers .).etc would be good ? so that we have some justification about why we are deprecating.

nixpanic:
> `Thick-provisioning was introduced to make accounting of assigned space for volumes easier. `
> 
> In general thick provisiioning helps to reduce storage latency..etc, and a useful thing to have in use cases like vm templating.etc. so may be good to add that as well to the description. also a small description about the issues/problems we encountered ( with pointers .).etc would be good ? so that we have some justification about why we are deprecating.

Any latency related to thick-provisioning depends on the storage system used. With Ceph, there is no difference, there is no performance benefit when thick-provisioning is used. This was known in advance, so the only benefit was related to the accounting of space consumption. We never mentioned any other benefit, so I wonder why we would need to remind people about that now.

humblec:
> > `Thick-provisioning was introduced to make accounting of assigned space for volumes easier. `
> > In general thick provisiioning helps to reduce storage latency..etc, and a useful thing to have in use cases like vm templating.etc. so may be good to add that as well to the description. also a small description about the issues/problems we encountered ( with pointers .).etc would be good ? so that we have some justification about why we are deprecating.
> 
> Any latency related to thick-provisioning depends on the storage system used. With Ceph, there is no difference, there is no performance benefit when thick-provisioning is used. This was known in advance, so the only benefit was related to the accounting of space consumption. We never mentioned any other benefit, so I wonder why we would need to remind people about that now.

Agree, but just wondering what was the justifications we have provided while introducing it in first place :) , if it was only accounting of space consumption, I think thats good enough.

nixpanic:
> Agree, but just wondering what was the justifications we have provided while introducing it in first place :) , if it was only accounting of space consumption, I think thats good enough.

Yes, it was only introduced to make accounting of requested space easier.

humblec (dismissed):

internal/rbd/controllerserver.go
```diff
@@ -1517,5 +1521,11 @@ func isThickProvisionRequest(parameters map[string]string) bool {
...
+		log.WarningLog(context.TODO(), "thick-provisioning is "+
```
Madhu-1: use `WarningLogMsg` as no context need to be passed?

Madhu-1 (commented):
small nit.

internal/rbd/controllerserver.go
nixpanic: oh, yes, will do!

Madhu-1 (approved):

nixpanic:

Failed to start minikube VM ([logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.22/detail/mini-e2e-helm_k8s-1.22/830/pipeline)), #2264

yati1998 (approved):
LGTM

PR #2594, opened 2021-10-26
rbd Regional DR: rcheck local image state for resyncing

Madhu-1:
below are the local states of the mirrored image

1.  "unknown"  -> If the image is in an error state  means data is completely synced
2. "error" -> If the image is in an error state means it needs resync
3. "syncing"
4. "starting_replay"
5. "replaying" -> If the image is in a replaying state means its replaying from primary source
6. "stopping_replay"
7. "stopped" 

If the resync is successfully started which means the image will be in a "replaying" state we can consider "replaying" state to report resync successfully going on the state.
we are discarding the intermediate states like "syncing", "starting_replay" and "stopping_replay".

This addresses the comment https://github.com/ceph/ceph-csi/pull/2572#pullrequestreview-779833923

Note to review: This is the suggestion approved by the RBD team. https://bugzilla.redhat.com/show_bug.cgi?id=2012143#c17

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -532,7 +532,6 @@ func checkRemoteSiteStatus(ctx context.Context, mirrorStatus *librbd.GlobalMirro
 // ResyncVolume extracts the RBD volume information from the volumeID, If the
 // image is present, mirroring is enabled and the image is in demoted state.
 // If yes it will resync the image to correct the split-brain.
-// FIXME: reduce complexity.
 func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	req *replication.ResyncVolumeRequest,
 ) (*replication.ResyncVolumeResponse, error) {
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -513,9 +513,10 @@ func checkRemoteSiteStatus(ctx context.Context, mirrorStatus *librbd.GlobalMirro
 	for _, s := range mirrorStatus.SiteStatuses {
 		log.UsefulLog(
 			ctx,
-			"peer site mirrorUUID=%s, mirroring state=%s, description=%s and lastUpdate=%s",
+			"peer site mirrorUUID=%q, daemon up=%t, mirroring state=%q, description=%q and lastUpdate=%d",
 			s.MirrorUUID,
-			s.State.String(),
+			s.Up,
+			s.State,
 			s.Description,
 			s.LastUpdate)
 		if s.MirrorUUID != "" {
@@ -611,8 +612,9 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	lastUpdateTime := time.Unix(localStatus.LastUpdate, 0).UTC()
 	log.UsefulLog(
 		ctx,
-		"local image mirroring state=%s, description=%s and lastUpdate=%s",
-		localStatus.State.String(),
+		"local status: daemon up=%t, image mirroring state=%q, description=%q and lastUpdate=%s",
+		localStatus.Up,
+		localStatus.State,
 		localStatus.Description,
 		lastUpdateTime)
 
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -523,16 +523,15 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 func checkRemoteSiteStatus(ctx context.Context, mirrorStatus *librbd.GlobalMirrorImageStatus) bool {
 	ready := true
 	for _, s := range mirrorStatus.SiteStatuses {
+		log.UsefulLog(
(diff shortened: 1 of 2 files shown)


internal/rbd/replicationcontrollerserver.go
```diff
@@ -633,6 +633,14 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	if state == unknown && localStatus.Up {
 		ready = checkRemoteSiteStatus(ctx, mirrorStatus)
 	}
+	// convert the last update time to UTC
+	lastUpdateTime := time.Unix(localStatus.LastUpdate, 0).UTC()
+	log.UsefulLog(
+		ctx,
+		"image mirroring state=%s, description=%s and lastUpdate=%s",
```
nixpanic: would it make sense to use `%q` to get the quoted strings?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -664,6 +668,25 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+	// cephCSI can return error message and volume replication operator can
+	// mark the VolumeReplication status as not resyncing for the volume.
+
+	// If the state is Replaying means the resync is going on.
+	// Once the volume on remote cluster is demoted and resync
+	// is completed the image state will be moved to UNKNOWN .
+	if localStatus.State != librbd.MirrorImageStatusStateReplaying &&
+		localStatus.State != librbd.MirrorImageStatusStateUnknown {
+		return fmt.Errorf(
+			"not resyncing. image is in %s state",
```
nixpanic: probably use `%q` here too

internal/rbd/replicationcontrollerserver.go
```diff
@@ -633,6 +633,14 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	if state == unknown && localStatus.Up {
 		ready = checkRemoteSiteStatus(ctx, mirrorStatus)
 	}
+	// convert the last update time to UTC
+	lastUpdateTime := time.Unix(localStatus.LastUpdate, 0).UTC()
+	log.UsefulLog(
+		ctx,
+		"image mirroring state=%s, description=%s and lastUpdate=%s",
+		localStatus.State.String(),
```
nixpanic: `.String()` should be used automatically when formatting strings, there is no need to call it?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -523,16 +523,15 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 func checkRemoteSiteStatus(ctx context.Context, mirrorStatus *librbd.GlobalMirrorImageStatus) bool {
 	ready := true
 	for _, s := range mirrorStatus.SiteStatuses {
+		log.UsefulLog(
+			ctx,
+			"peer site mirrorUUID=%s, mirroring state=%s, description=%s and lastUpdate=%s",
```
nixpanic: Use `%q`? I understand the code got moved, but you can improve it as well :smiley:

internal/rbd/replicationcontrollerserver.go
```diff
@@ -695,10 +684,11 @@ func resyncRequired(localStatus librbd.SiteMirrorImageStatus) bool {
 	// be in an error state. It would be also worth considering the `description`
 	// field to make sure about split-brain.
 	splitBrain := "split-brain"
-	if strings.Contains(localStatus.State.String(), string(errorState)) ||
+	resyncReq := false
```
nixpanic: why introduce the additional variable? There are lots of function that have multiple `return` statements, which seems to be pretty common in Go.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -532,7 +532,6 @@ func checkRemoteSiteStatus(ctx context.Context, mirrorStatus *librbd.GlobalMirro
 // ResyncVolume extracts the RBD volume information from the volumeID, If the
 // image is present, mirroring is enabled and the image is in demoted state.
 // If yes it will resync the image to correct the split-brain.
-// FIXME: reduce complexity.
 func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
```
nixpanic, line 535: The commit message mentions `complexity of ResyncVolume is not reduced`. Do you mean `now` instead of `not`?

nixpanic (commented):
mostly nits for your consideration, will have a closer look later

internal/rbd/replicationcontrollerserver.go
Madhu-1: yes it was by mistake fixing now

internal/rbd/replicationcontrollerserver.go
Madhu-1: wanted to avoid multiple returns. as its common reverting back to 2 returns.

ShyamsundarR (approved):
LGTM!

yati1998 (approved):

nixpanic (approved):

PR #2602, opened 2021-10-28
util: set defaults for Vault config before converting

nixpanic:
When using UPPER_CASE formatting for the HashiCorp Vault KMS
configuration, a missing `VAULT_DESTROY_KEYS` will cause the option to
be set to "false". The default for the option is intended for be "true".

This is a difference in behaviour between the `vaultDestroyKeys` and
`VAULT_DESTROY_KEYS` options. Both should use a default of "true" when
the configuration does not set the option explicitly.

By setting the default options in the `standardVault` struct before
unmarshalling the configuration in it, the default values will be
retained for the missing configuration options.

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/kms/vault.go
+++ b/internal/kms/vault.go
@@ -43,7 +43,7 @@ const (
 	vaultDefaultRole           = "csi-kubernetes"
 	vaultDefaultNamespace      = ""
 	vaultDefaultPassphrasePath = ""
-	vaultDefaultCAVerify       = "true"
+	vaultDefaultCAVerify       = true
 	vaultDefaultDestroyKeys    = "true"
 )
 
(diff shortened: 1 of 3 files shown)


yati1998 (approved):
LGTM.

nixpanic:
> /retest ci/centos/mini-e2e/k8s-1.22

Failed with #2204 ([logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.22/detail/mini-e2e_k8s-1.22/839/pipeline))

Madhu-1 (approved):

internal/kms/vault.go
```diff
@@ -43,7 +43,7 @@ const (
 	vaultDefaultRole           = "csi-kubernetes"
 	vaultDefaultNamespace      = ""
 	vaultDefaultPassphrasePath = ""
-	vaultDefaultCAVerify       = "true"
+	vaultDefaultCAVerify       = true
```
humblec, line 46: @nixpanic this is bit confusing actually, vaultDefaultCAVerify take it as bool and DestroyKeys with string but in form form of boolean  "true". Cant we make it in uniform manner?

Issue #2587, opened 2021-10-19
Add inline volume support for   Ceph CSI driver 


humblec:
# Describe the feature you'd like to have #

Traditionally, volumes that are backed by CSI drivers can only be used with a PersistentVolume and PersistentVolumeClaim object combination. Two different Kubernetes features allow volumes to follow the Pod's lifecycle: CSI ephemeral volumes and generic ephemeral volumes.

In both features, the volumes are specified directly in the pod specification for ephemeral use cases. At runtime, nested inline volumes follow the ephemeral lifecycle of their associated pods where Kubernetes and the driver handle all phases of volume operations as pods are created and destroyed.

CSI ephemeral Inline volume spec example:

```
kind: Pod
apiVersion: v1
metadata:
  name: my-csi-app
spec:
  containers:
    - name: my-frontend
      image: busybox
      volumeMounts:
      - mountPath: "/data"
        name: my-csi-inline-vol
      command: [ "sleep", "1000000" ]
  volumes:
    - name: my-csi-inline-vol
      csi:
        driver: <drivername>
        volumeAttributes:
          something: something
```

General inline ephemeral support example:

```
kind: Pod
apiVersion: v1
metadata:
  name: some-pod
spec:
  containers:
     ...
  volumes:
    - name: scratch-volume
      ephemeral:
        volumeClaimTemplate:
          metadata:
            labels:
              type: my-frontend-volume
          spec:
            accessModes: [ "ReadWriteOnce" ]
            storageClassName: "scratch-storage-class"
            resources:
              requests:
                storage: 1Gi
```

What new functionality do you want?

Add inline volume support for Ceph CSI driver

# What is the value to the end user? (why is it a priority?) #

The volume lifecycle can be controlled by the POD spec

# Additional context #

Design doc for the feature support has been captured here:
https://hackmd.io/SXcB4VizTKaI4E7efOV_xw

Madhu-1:
@humblec why have one more tracker as we already have an issue opened for this one? can we close this one and make use of #1461?

humblec:
> @humblec why have one more tracker as we already have an issue opened for this one? can we close this one and make use of #1461?

That issue has been filed only for cephfs and this tracker supposed to cover ceph csi driver in general and also kind of umbrella tracker for the feature.  #1461 can be added as a subtracker in that sense.

PR #2603, opened 2021-10-28
 rbd: add generic ephemeral volume validation

humblec:
rbd: add generic ephemeral volume validation
    
    This commit adds the validation of csi RBD driver to work with
    ephemeral volume support. With ephemeral volume support a user can
    specify ephemeral volumes in its pod spec and tie the lifecycle
    of the PVC with the POD.
    
    An example POD spec looks like this:
    
    kind: Pod
    apiVersion: v1
    metadata:
      name: my-app
       spec:
      containers:
        - name: csi-rbd-demo-pod
          image: busybox
          volumeMounts:
            - mountPath: "/scratch"
              name: mypvc
          command: [ "sleep", "1000000" ]
      volumes:
        - name: mypvc
          ephemeral:
            volumeClaimTemplate:
              spec:
                accessModes: [ "ReadWriteOnce" ]
                storageClassName: "csi-rbd-sc"
                resources:
                  requests:
                    storage: 1Gi

Additional Ref # https://github.com/rook/rook/pull/9055

Updates #2587 

### DIFF
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -279,6 +279,7 @@ var _ = Describe("cephfs", func() {
 			appClonePath := cephFSExamplePath + "pod-restore.yaml"
 			appSmartClonePath := cephFSExamplePath + "pod-clone.yaml"
 			snapshotPath := cephFSExamplePath + "snapshot.yaml"
+			appEphemeralPath := cephFSExamplePath + "pod-ephemeral.yaml"
 
 			By("checking provisioner deployment is running", func() {
 				err := waitForDeploymentComplete(cephFSDeploymentName, cephCSINamespace, f.ClientSet, deployTimeout)
@@ -312,6 +313,37 @@ var _ = Describe("cephfs", func() {
 					}
 				})
 			}
+			By("verify generic ephemeral volume support", func() {
+				// generic ephemeral volume support is beta since v1.21.
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 21) {
+					Skip("generic ephemeral volume only supported from v1.21+")
+				}
+				err := createCephfsStorageClass(f.ClientSet, f, true, nil)
+				if err != nil {
+					e2elog.Failf("failed to create CephFS storageclass: %v", err)
+				}
+				// create application
+				app, err := loadApp(appEphemeralPath)
+				if err != nil {
+					e2elog.Failf("failed to load application: %v", err)
+				}
+				app.Namespace = f.UniqueName
+				err = createApp(f.ClientSet, app, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create application: %v", err)
+				}
+				validateSubvolumeCount(f, 1, fileSystemName, subvolumegroup)
+				// delete pod
+				err = deletePod(app.Name, app.Namespace, f.ClientSet, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete application: %v", err)
+				}
+				validateSubvolumeCount(f, 0, fileSystemName, subvolumegroup)
+				err = deleteResource(cephFSExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete CephFS storageclass: %v", err)
+				}
+			})
 
 			By("check static PVC", func() {
 				scPath := cephFSExamplePath + "secret.yaml"
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -57,6 +57,7 @@ var (
 	appClonePath           = rbdExamplePath + "pod-restore.yaml"
 	appSmartClonePath      = rbdExamplePath + "pod-clone.yaml"
 	appBlockSmartClonePath = rbdExamplePath + "block-pod-clone.yaml"
+	appEphemeralPath       = rbdExamplePath + "pod-ephemeral.yaml"
 	snapshotPath           = rbdExamplePath + "snapshot.yaml"
 	defaultCloneCount      = 10
 
@@ -367,6 +368,36 @@ var _ = Describe("RBD", func() {
 					}
 				})
 			}
+			By("verify generic ephemeral volume support", func() {
+				// generic ephemeral volume support is supported from 1.21
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 21) {
+					Skip("generic ephemeral volume only supported from v1.21+")
+				}
+				// create application
+				app, err := loadApp(appEphemeralPath)
+				if err != nil {
+					e2elog.Failf("failed to load application: %v", err)
+				}
+				app.Namespace = f.UniqueName
+				err = createApp(f.ClientSet, app, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create application: %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 1, defaultRBDPool)
+				err = deletePod(app.Name, app.Namespace, f.ClientSet, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete application: %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0, defaultRBDPool)
+				// validate images in trash
+				err = waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to validate rbd images in pool %s trash: %v", defaultRBDPool, err)
+				}
+			})
+
 			// todo: may be remove the below deletion test later once the migration nodestage tests are adjusted
 			// also to have deletion validation through the same.
 			By("validate RBD migration+static Block PVC Deletion", func() {
--- a/examples/rbd/pod-ephemeral.yaml
+++ b/examples/rbd/pod-ephemeral.yaml
@@ -1,0 +1,23 @@
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: csi-rbd-demo-ephemeral-pod
+spec:
+  containers:
+    - name: web-server
+      image: docker.io/library/nginx:latest
+      volumeMounts:
+        - mountPath: /myspace
+          name: mypvc
+  volumes:
+    - name: mypvc
+      ephemeral:
+        volumeClaimTemplate:
+          spec:
+            accessModes:
+              - ReadWriteOnce
+            storageClassName: csi-rbd-sc
+            resources:
+              requests:
+                storage: 1Gi
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -160,7 +160,7 @@ CSI_RESIZER_VERSION=${CSI_RESIZER_VERSION:-"v1.2.0"}
 CSI_NODE_DRIVER_REGISTRAR_VERSION=${CSI_NODE_DRIVER_REGISTRAR_VERSION:-"v2.2.0"}
 
 #feature-gates for kube
-K8S_FEATURE_GATES=${K8S_FEATURE_GATES:-"ExpandCSIVolumes=true"}
+K8S_FEATURE_GATES=${K8S_FEATURE_GATES:-""}
 
 #extra-config for kube https://minikube.sigs.k8s.io/docs/reference/configuration/kubernetes/
 EXTRA_CONFIG_PSP="--extra-config=apiserver.enable-admission-plugins=PodSecurityPolicy --addons=pod-security-policy"
--- a/examples/cephfs/pod-ephemeral.yaml
+++ b/examples/cephfs/pod-ephemeral.yaml
@@ -1,0 +1,23 @@
+---
+kind: Pod
+apiVersion: v1
+metadata:
+  name: csi-cephfs-demo-ephemeral-pod
+spec:
+  containers:
+    - name: web-server
+      image: docker.io/library/nginx:latest
+      volumeMounts:
+        - mountPath: /myspace
+          name: mypvc
+  volumes:
+    - name: mypvc
+      ephemeral:
+        volumeClaimTemplate:
+          spec:
+            accessModes:
+              - ReadWriteOnce
+            storageClassName: csi-cephfs-sc
+            resources:
+              requests:
+                storage: 1Gi


e2e/rbd.go
```diff
@@ -367,6 +368,28 @@ var _ = Describe("RBD", func() {
 					}
 				})
 			}
+			By("verify ephemeral volume support", func() {
+				// create application
+				app, err := loadApp(appEphemeralPath)
+				if err != nil {
+					e2elog.Failf("failed to load application with error %v", err)
```
Madhu-1: remove with error from all places

e2e/rbd.go
```diff
@@ -367,6 +368,28 @@ var _ = Describe("RBD", func() {
...
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 1, defaultRBDPool)
+
+				err = deletePod(app.Name, app.Namespace, f.ClientSet, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete application with error %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0, defaultRBDPool)
```
Madhu-1, line 393: please validate image in trash

examples/rbd/pod-ephemeral.yaml
```diff
@@ -0,0 +1,22 @@
...
+      volumeMounts:
+        - mountPath: "/myspace"
+          name: mypvc
+  volumes:
+    - name: mypvc
+      ephemeral:
+        volumeClaimTemplate:
+          spec:
+            accessModes: ["ReadWriteOnce"]
+            storageClassName: "csi-rbd-sc"
```
Madhu-1: quotes not required for mountPath and storageclass and also for access mode? please keep the same as in other yamls.

e2e/rbd.go
```diff
@@ -367,6 +368,28 @@ var _ = Describe("RBD", func() {
 					}
 				})
 			}
+			By("verify ephemeral volume support", func() {
```
Madhu-1: generic ephemeral volume support?

e2e/rbd.go
```diff
@@ -367,6 +368,28 @@ var _ = Describe("RBD", func() {
...
+				// create application
+				app, err := loadApp(appEphemeralPath)
```
Madhu-1, line 377: CI will fail for 20 as the feature gate is still alpha?

* add kube validation check here
* enable feature gate for kube 1.20?

Madhu-1 (dismissed):

Madhu-1:
please add same e2e for cephfs also

e2e/rbd.go
humblec: running the tests only >= 21 is good enough I believe. wdyt?

Madhu-1: my suggestion is to run from 20+ as we test on kube 1.20,21, and 22 also

humblec:
@Madhu-1 added tests for cephfs too and enabled feature gate only for kube versions >=1.20.  PTAL.

scripts/minikube.sh
```diff
@@ -202,6 +211,14 @@ up)
 
     disable_storage_addons
 
+    #  get kubernetes version we are operating on and accordingly enable feature gates
+    KUBE_MAJOR=$(kube_version 1)
+    KUBE_MINOR=$(kube_version 2)
+    if  [ "${KUBE_MAJOR}" -eq 1 ] && [ "${KUBE_MINOR}" -ge 20 ];
```
nixpanic: 1. not sure you can compare strings with `-eq` or `-ge`
2. no `;` needed when `then` is on the next  line

scripts/minikube.sh
```diff
@@ -202,6 +211,14 @@ up)
...
+    then
+    # if kubernetes version is greater than 1.20 enable Generic Ephemeral Volume feature gate
+    K8S_FEATURE_GATES="GenericEphemeralVolume=true"
```
nixpanic: `K8S_FEATURE_GATES` is already set at line 163, this overrides that option. Should the variable not be extended with the feature gate instead?

nixpanic (changes requested):
It is simpler to included the example .yaml files with the e2e commits so that they are self contained and really belong together. There is then no need to include the full yaml in the commit message anymore, cleaning things up a little.

humblec:
> It is simpler to included the example .yaml files with the e2e commits so that they are self contained and really belong together. There is then no need to include the full yaml in the commit message anymore, cleaning things up a little.

sure, squashed commits of example yaml to the validation commit.

scripts/minikube.sh
humblec: the `-eq or ge` is working as expected, I have removed unwanted `;` in the last refresh.

scripts/minikube.sh
humblec: my plan was to remove the expandCSIVolumes feature gate as a follow up as it no longer wanted, will file a seperate PR for the same, however its better to be in appended string form here, so that in future if there is already set value, it should carry that in final string.

scripts/minikube.sh
```diff
@@ -202,6 +211,14 @@ up)
 
     disable_storage_addons
 
+    #  get kubernetes version we are operating on and accordingly enable feature gates
+    KUBE_MAJOR=$(kube_version 1)
+    KUBE_MINOR=$(kube_version 2)
+    if  [ "${KUBE_MAJOR}" -eq 1 ] && [ "${KUBE_MINOR}" -ge 20 ]
+    then
+    # if kubernetes version is greater than 1.20 enable Generic Ephemeral Volume feature gate
+    K8S_FEATURE_GATES="${K8S_FEATURE_GATES},GenericEphemeralVolume=true"
```
nixpanic: if `K8S_FEATURE_GATES` is an empty string, maybe the extra `,` can cause issues?

scripts/minikube.sh
```diff
@@ -160,7 +169,7 @@ CSI_RESIZER_VERSION=${CSI_RESIZER_VERSION:-"v1.2.0"}
 CSI_NODE_DRIVER_REGISTRAR_VERSION=${CSI_NODE_DRIVER_REGISTRAR_VERSION:-"v2.2.0"}
 
 #feature-gates for kube
-K8S_FEATURE_GATES=${K8S_FEATURE_GATES:-"ExpandCSIVolumes=true"}
+K8S_FEATURE_GATES=${K8S_FEATURE_GATES:-""}
```
nixpanic, line 163: Now you are dropping `ExpandCSIVolumes` without comment. It is better to do that in a stand-alone commit. As a cleanup, it can be included in this PR, no need to create a separate one.

nixpanic (dismissed):

scripts/minikube.sh
humblec: sure, put it as a seperate commit here in this PR itself.

scripts/minikube.sh
humblec: as discussed, its not causing issue rather its going through, so will keep this version.

humblec:
@Mergifyio rebase

humblec:
@Mergifyio refresh

humblec:
Test failure https://github.com/ceph/ceph-csi/issues/2610

humblec:
@Madhu-1 can you please review this ?

e2e/cephfs.go
```diff
@@ -312,6 +313,36 @@ var _ = Describe("cephfs", func() {
...
 				})
 			}
+			By("verify generic ephemeral volume support", func() {
+				// generic ephemeral volume support is beta since v1.21.
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 21) {
+					Skip("generic ephemeral volume only supported from v1.21+")
+				}
+				err := createCephfsStorageClass(f.ClientSet, f, true, nil)
+				if err != nil {
+					e2elog.Failf("failed to create CephFS storageclass with error %v", err)
```
Madhu-1: `failed to create CephFS storageclass with error %v` to `failed to create CephFS storageclass: %v` `with error` formatting needs to be adjusted in other places also

e2e/cephfs.go
```diff
@@ -312,6 +313,36 @@ var _ = Describe("cephfs", func() {
...
+				// create application
+				app, err := loadApp(appEphemeralPath)
+				if err != nil {
+					e2elog.Failf("failed to load application: %v", err)
+				}
+				app.Namespace = f.UniqueName
+				err = createApp(f.ClientSet, app, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create application: %v", err)
+				}
```
Madhu-1, line 334: add a subvolumeCount validation here to make sure subvolume is present in backend?

e2e/rbd.go
```diff
@@ -367,6 +368,40 @@ var _ = Describe("RBD", func() {
...
+				err = waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to validate rbd images in pool %s trash: %v", defaultRBDPool, err)
+				}
+				err = waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to validate rbd images in pool %s trash: %v", defaultRBDPool, err)
+				}
+			})
+
```
Madhu-1:
```suggestion
// validate images in trash
				err = waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
				if err != nil {
					e2elog.Failf("failed to validate rbd images in pool %s trash: %v", defaultRBDPool, err)
				}
```

scripts/minikube.sh
```diff
@@ -202,6 +211,14 @@ up)
 
     disable_storage_addons
 
+    #  get kubernetes version we are operating on and accordingly enable feature gates
+    KUBE_MAJOR=$(kube_version 1)
+    KUBE_MINOR=$(kube_version 2)
+    if  [ "${KUBE_MAJOR}" -eq 1 ] && [ "${KUBE_MINOR}" -ge 21 ]
+    then
+    # if kubernetes version is greater than 1.21 enable Generic Ephemeral Volume feature gate
+    K8S_FEATURE_GATES="${K8S_FEATURE_GATES},GenericEphemeralVolume=true"
```
Madhu-1: as it is beta in 1.21+ already enabled by default. we don't need to enable it explicitly?

e2e/cephfs.go
```diff
@@ -312,6 +313,36 @@ var _ = Describe("cephfs", func() {
 					}
 				})
 			}
+			By("verify generic ephemeral volume support", func() {
+				// generic ephemeral volume support is beta since v1.21.
+				if !k8sVersionGreaterEquals(f.ClientSet, 1, 21) {
```
Madhu-1, line 318: it's skipping for 1.20 but as per https://github.com/ceph/ceph-csi/pull/2603#discussion_r739026205 it's agreed to run on  1.20 also?

humblec: this was deliberate, considering this is beta and supported since 1.21, this should be good I believe.

scripts/minikube.sh
humblec: @Madhu-1 if this script is executed for versions less than 1.21, we dont need to enable this feature gate, thats why tihs check in place.

e2e/rbd.go
humblec: on more thoughts, considering other tests are enabled for beta for example, snapshot/clone...etc it looks like its fine to enable only on versions 1.21. also next release is anyway going to throw away 1.20 from here, so should be good I believe. Last patch refresh adjusted the same.

humblec:
@nixpanic @Madhu-1 all the comments are addressed, ptal.. thanks in advance.

e2e/cephfs.go
nixpanic: agreed, but in that case it is not needed to enable the featue gate in `minikube.sh`?

humblec: it was added considering the script can be used independently for kube deployments, but may be here, we can skip that and only have e2e check. adjusted the PR..

scripts/minikube.sh
humblec: it was added considering the script can be used independently for kube deployments, but may be here, we can skip that and only have e2e check. adjusted the PR..

examples/rbd/pod-ephemeral.yaml
```diff
@@ -0,0 +1,22 @@
...
+      image: docker.io/library/nginx:latest
+      volumeMounts:
+        - mountPath: /myspace
+          name: mypvc
+  volumes:
+    - name: mypvc
+      ephemeral:
+        volumeClaimTemplate:
+          spec:
+            accessModes: [ReadWriteOnce]
```
nixpanic: The other examples use the cleaner (IMHO) format of

```yaml
accessModes:
  - ReadWriteOnce
```

it is probably better to use the same formatting in all examples we have

nixpanic (dismissed):
just a nit about the yaml formatting of the examples, looks good to me otherwise

examples/rbd/pod-ephemeral.yaml
```diff
@@ -0,0 +1,23 @@
+--- 
+apiVersion: v1
+kind: Pod
+metadata:
+  name: csi-rbd-demo-ephemeral-pod
+spec:
+  containers:
+      image: "docker.io/library/nginx:latest"
```
Madhu-1: remove `""` from image name

examples/rbd/pod-ephemeral.yaml
```diff
@@ -0,0 +1,23 @@
...
+      ephemeral:
+        volumeClaimTemplate:
+          spec:
+            accessModes:
+              - ReadWriteOnce
+            resources:
+              requests:
+                storage: 1Gi
+            storageClassName: csi-rbd-sc
+      name: mypvc
```
Madhu-1: move the name to line 14  and keep it similar to cephfs

humblec:
@Mergifyio rebase

humblec:
@nixpanic @Madhu-1 ptal.. thanks.

Madhu-1 (dismissed):

humblec:
rebased to devel, and also a minor change ( as pointed out by @Rakshith-R in ci logs)  from `appEphemeralPath = cephFSExamplePath + "pod-ephemeral.yaml"` to `appEphemeralPath := cephFSExamplePath + "pod-ephemeral.yaml"`

Madhu-1 (approved):

Rakshith-R (approved):

humblec:
@Mergifyio rebase

nixpanic:
> /retest ci/centos/upgrade-tests

For some reason the CentOS jobs were not listed, this causes a refresh from Jenkins in the CentOS CI.

humblec:
@Mergifyio rebase

humblec:
@Mergifyio rebase

humblec:
@Mergifyio refresh

humblec:
@Mergifyio rebase

Rakshith-R:
@mergifyio rebase

humblec:
@Mergifyio rebase

humblec:
@Mergifyio rebase

PR #2631, opened 2021-11-10
doc: lift minimum Ceph Cluster Version to Octopus

Rakshith-R:

Support for ceph Nautilus is being removed since it is
at end of life and minimum required version is now set
to ceph Octopus.

Refer: https://docs.ceph.com/en/latest/releases/

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,28 +79,28 @@ for its support details.
 
 | Plugin | Features                                                  | Feature Status | CSI Driver Version | CSI Spec Version | Ceph Cluster Version | Kubernetes Version |
 | ------ | --------------------------------------------------------- | -------------- | ------------------ | ---------------- | -------------------- | ------------------ |
-| RBD    | Dynamically provision, de-provision Block mode RWO volume | GA             | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.14.0         |
-|        | Dynamically provision, de-provision Block mode RWX volume | GA             | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.14.0         |
-|        | Dynamically provision, de-provision File mode RWO volume  | GA             | >= v1.0.0          | >= v1.0.0        | Nautilus (>=14.0.0)  | >= v1.14.0         |
-|        | Provision File Mode ROX volume from snapshot              | Alpha          | >= v3.0.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.17.0         |
-|        | Provision File Mode ROX volume from another volume        | Alpha          | >= v3.0.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.16.0         |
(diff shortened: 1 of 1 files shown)


humblec:
@Rakshith-R 

Nautilus last release (14.2.22)  happened this year on `2021-06-30` , I dont think dropping its support is a good idea. Also Ceph has not moved this release to Archived so far.

humblec (changes requested):
Nautilus last release (14.2.22) happened this year on 2021-06-30 , I dont think dropping its support is a good idea. Also Ceph has not moved this release to Archived so far.

README.md
```diff
@@ -79,28 +79,28 @@ for its support details.
...
-|        | Volume/PV Metrics of File Mode Volume                     | Beta           | >= v1.2.0          | >= v1.1.0        | Nautilus (>=14.0.0)  | >= v1.15.0         |
-|        | Volume/PV Metrics of Block Mode Volume                    | Beta           | >= v1.2.0          | >= v1.1.0        | Nautilus (>=14.0.0)  | >= v1.21.0         |
-|        | Topology Aware Provisioning Support                       | Alpha          | >= v2.1.0          | >= v1.1.0        | Nautilus (>=14.0.0)  | >= v1.14.0         |
-| CephFS | Dynamically provision, de-provision File mode RWO volume  | Beta           | >= v1.1.0          | >= v1.0.0        | Nautilus (>=14.2.2)  | >= v1.14.0         |
-|        | Dynamically provision, de-provision File mode RWX volume  | Beta           | >= v1.1.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.14.0         |
-|        | Dynamically provision, de-provision File mode ROX volume  | Alpha          | >= v3.0.0          | >= v1.0.0        | Nautilus (>=v14.2.2) | >= v1.14.0         |
```
nixpanic, line 98: I do not think this is the right way to do it. Users can still pull older ceph-csi container images and use that with older Ceph versions. It probably is better to add a table where the supported ceph versions and ceph-csi versions are listed. The next ceph-csi version will not be supported on Nautilus anymore, current ceph-csi releases will not be changed.

Add a link to the Ceph releases as well, that should make things easier to understand and track

Madhu-1: we need to start adding a new column here to represent the compatibility between cephcsi and ceph.

Rakshith-R: |   CSI Driver Version   | Ceph Cluster Version  |
|----|---|
| >=v3.5.0 | Octopus (>=v15.0.0) |
| >=v1.0.0 | Nautilus  (>=v14.0.0) |

Something like this?

nixpanic: Yes, something like that. Or mention the Ceph version used in the container image? That way users can check themselves if it is compatible with whatever old/new Ceph cluster they use.

Madhu-1: yes, something like that, and also good to mention that check older cephcsi releases if you are looking for support of older maintained ceph version.

Madhu-1:
@Rakshith-R refresh or close this PR?

PR #2633, opened 2021-11-11
rbd: use go-ceph rbd admin task api for 'trash remove' & 'flatten'

### LINKED ISSUE #2186: Use go-ceph for ceph task manager commands
Currently, we are using RBD CLI for the below two manager functionalities

* Add task to fatten the rbd image
* Add task to move the image to the trash

go-ceph issue https://github.com/ceph/go-ceph/issues/516 to track the same work. Once go-ceph API's are available we should switch from  CLI to go-ceph

Thanks, @phlogistonjohn for finding this one and opening the go-ceph issue :)

Rakshith-R:
- rebase: use go-ceph v0.13.0
- rbd: use go-ceph rbd admin task api instead of cli   
- rbd: remove redundant rbdVolume.connect() in flattneRbdImage() 
- rbd: remove redundant util.Credentials arg from flattenRbdImage() 
- rbd: refractor deleteImage() & trashRemoveImage() 

Closes: #2186 

New Mgr Not Supported errors thrown by go-ceph
- `API call not implemented server-side: No handler found for 'rbd task add trash remove'` (rook 1.2.0, ceph v:14.2.2)
- `rados: ret=-13, Permission denied: "access denied: does your client key have mgr caps? See http://docs.ceph.com/en/latest/mgr/administrator/#client-authentication"` (when user secret does not have mgr caps)

### DIFF
--- a/build.env
+++ b/build.env
@@ -13,7 +13,7 @@ CSI_IMAGE_VERSION=canary
 
 # Ceph version to use
 BASE_IMAGE=docker.io/ceph/ceph:v16
-CEPH_VERSION=octopus
+CEPH_VERSION=pacific
 
 # standard Golang options
 GOLANG_VERSION=1.17.5
--- a/internal/rbd/migration.go
+++ b/internal/rbd/migration.go
@@ -82,7 +82,7 @@ func deleteMigratedVolume(ctx context.Context, parsedMigHandle *migrationVolID, 
 		return err
 	}
 	defer rv.Destroy()
-	err = deleteImage(ctx, rv, cr)
+	err = rv.deleteImage(ctx)
 	if err != nil {
 		log.ErrorLog(ctx, "failed to delete rbd image: %s, err: %v", rv, err)
 	}
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -56,9 +56,8 @@ const (
 	// Output strings returned during invocation of "ceph rbd task add remove <imagespec>" when
 	// command is not supported by ceph manager. Used to check errors and recover when the command
 	// is unsupported.
-	rbdTaskRemoveCmdInvalidString1      = "no valid command found"
-	rbdTaskRemoveCmdInvalidString2      = "Error EINVAL: invalid command"
-	rbdTaskRemoveCmdAccessDeniedMessage = "Error EACCES:"
+	rbdTaskRemoveCmdInvalidString       = "No handler found"
+	rbdTaskRemoveCmdAccessDeniedMessage = "access denied:"
 
 	// image metadata key for thick-provisioning.
 	// As image metadata key starting with '.rbd' will not be copied when we do
@@ -595,43 +594,27 @@ func isNotMountPoint(mounter mount.Interface, stagingTargetPath string) (bool, e
 	return isNotMnt, err
 }
 
-// addRbdManagerTask adds a ceph manager task to execute command
-// asynchronously. If command is not found returns a bool set to false
-// example arg ["trash", "remove","pool/image"].
-func addRbdManagerTask(ctx context.Context, pOpts *rbdVolume, arg []string) (bool, error) {
-	args := []string{"rbd", "task", "add"}
-	args = append(args, arg...)
-	log.DebugLog(
-		ctx,
-		"executing %v for image (%s) using mon %s, pool %s",
-		args,
-		pOpts.RbdImageName,
-		pOpts.Monitors,
-		pOpts.Pool)
-	supported := true
-	_, stderr, err := util.ExecCommand(ctx, "ceph", args...)
-	if err != nil {
-		switch {
-		case strings.Contains(stderr, rbdTaskRemoveCmdInvalidString1) &&
-			strings.Contains(stderr, rbdTaskRemoveCmdInvalidString2):
-			log.WarningLog(
-				ctx,
-				"cluster with cluster ID (%s) does not support Ceph manager based rbd commands"+
-					"(minimum ceph version required is v14.2.3)",
-				pOpts.ClusterID)
-			supported = false
-		case strings.HasPrefix(stderr, rbdTaskRemoveCmdAccessDeniedMessage):
-			log.WarningLog(ctx, "access denied to Ceph MGR-based rbd commands on cluster ID (%s)", pOpts.ClusterID)
-			supported = false
-		default:
-			log.WarningLog(ctx, "uncaught error while scheduling a task (%v): %s", err, stderr)
-		}
-	}
-	if err != nil {
-		err = fmt.Errorf("%w. stdError:%s", err, stderr)
+// isCephMgrSupported determines if the cluster has support for MGR based operation
+// depending on the error.
+func isCephMgrSupported(ctx context.Context, clusterID string, err error) bool {
+	switch {
+	case err == nil:
+		return true
+	case strings.Contains(err.Error(), rbdTaskRemoveCmdInvalidString):
+		log.WarningLog(
+			ctx,
+			"cluster with cluster ID (%s) does not support Ceph manager based rbd commands"+
+				"(minimum ceph version required is v14.2.3)",
+			clusterID)
+
+		return false
+	case strings.Contains(err.Error(), rbdTaskRemoveCmdAccessDeniedMessage):
+		log.WarningLog(ctx, "access denied to Ceph MGR-based rbd commands on cluster ID (%s)", clusterID)
+
+		return false
 	}
 
-	return supported, err
+	return true
 }
 
 // getTrashPath returns the image path for trash operation.
@@ -704,14 +687,17 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
 func trashRemoveImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
 	// attempt to use Ceph manager based deletion support if available
 
-	args := []string{
-		"trash", "remove",
-		pOpts.getTrashPath(),
-		"--id", cr.ID,
-		"--keyfile=" + cr.KeyFile,
-		"-m", pOpts.Monitors,
+	ra, err := pOpts.conn.GetRBDAdmin()
+	if err != nil {
+		return err
 	}
-	rbdCephMgrSupported, err := addRbdManagerTask(ctx, pOpts, args)
+
+	log.DebugLog(ctx, "rbd: adding task to remove image %s with id %s from trash", pOpts, pOpts.ImageID)
+
+	ta := ra.Task()
+	_, err = ta.AddTrashRemove(admin.NewImageSpec(pOpts.Pool, pOpts.RadosNamespace, pOpts.ImageID))
+
+	rbdCephMgrSupported := isCephMgrSupported(ctx, pOpts.ClusterID, err)
 	if rbdCephMgrSupported && err != nil {
 		log.ErrorLog(ctx, "failed to add task to delete rbd image: %s, %v", pOpts, err)
 
@@ -725,6 +711,8 @@ func trashRemoveImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credential
 
 			return err
 		}
+	} else {
+		log.DebugLog(ctx, "rbd: successfully added task to move image %s with id %s to trash", pOpts, pOpts.ImageID)
 	}
 
 	return nil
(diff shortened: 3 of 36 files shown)


Rakshith-R:

failed due to #2264 
https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/k8s-e2e-external-storage-1.22/detail/k8s-e2e-external-storage-1.22/531/pipeline

build.env
```diff
@@ -13,7 +13,7 @@ CSI_IMAGE_VERSION=canary
 
 # Ceph version to use
 BASE_IMAGE=docker.io/ceph/ceph:v16
-CEPH_VERSION=octopus
+BUILD_TAGS=ceph_preview
```
humblec: can we document the ceph_preview thing ?

humblec (changes requested):

build.env
Rakshith-R: >can we document the ceph_preview thing ?

```suggestion
# ceph_preview tag is required to access go-ceph APIs which are under preview.
BUILD_TAGS=ceph_preview
```

@humblec is this sufficient ?

internal/util/connection.go
```diff
@@ -123,6 +123,16 @@ func (cc *ClusterConnection) GetRBDAdmin() (*ra.RBDAdmin, error) {
 	return ra.NewFromConn(cc.conn), nil
 }
 
+// GetTaskAdmin returns TaskAdmin to add tasks on rbd volumes.
```
Madhu-1: `rbd volumes` to `rbd images`

Madhu-1 (commented):
LGTM . @Rakshith-R have you verified its working with older ceph releases where we don't have manager command support?

build.env
Madhu-1: add a link to https://github.com/ceph/go-ceph/blob/master/docs/api-stability.md#preview here

Rakshith-R:
> LGTM . @Rakshith-R have you verified its working with older ceph releases where we don't have manager command support?

Yes, I ran it with ceph 14.2.2 where there's rbd task support, works as expected.

Madhu-1:
> > LGTM . @Rakshith-R have you verified its working with older ceph releases where we don't have manager command support?
> 
> Yes, I ran it with ceph 14.2.2 where there's rbd task support, works as expected.

you mean no rbd task support?

Rakshith-R:
> > > LGTM . @Rakshith-R have you verified its working with older ceph releases where we don't have manager command support?
> > 
> > 
> > Yes, I ran it with ceph 14.2.2 where there's rbd task support, works as expected.
> 
> you mean no rbd task support?

yes :sweat_smile:  
*ceph 14.2.2 where there's * *no* * rbd task support, works as expected.

Madhu-1 (dismissed):

build.env
nixpanic: I do not think we should set `ceph_preview` as default. Make sure that APIs from go-ceph are stable and part of the latest release.

nixpanic (changes requested):
Suggestion to squash the last two patches together (drop `cr` from `flattenRbdImage()`).

Having `ceph_preview` as default in the `BUILD_TAGS` is not very clean, we should probably wait until go-ceph provides a version where there is no need to set the build tag.

Madhu-1 (dismissed):

Rakshith-R:
Resolved/ satisfied golangci linter, ptal thanks

build.env
Rakshith-R: `ceph_preview` is removed and go-ceph is upgraded to v0.13.0 in this pr instead

internal/rbd/rbd_util.go
```diff
@@ -700,14 +683,17 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
+	ra, err := pOpts.conn.GetRBDAdmin()
+	if err != nil {
+		return err
 	}
-	rbdCephMgrSupported, err := addRbdManagerTask(ctx, pOpts, args)
+
+	log.DebugLog(ctx, "rbd: adding task to remove image %s with id %s from trash", pOpts, pOpts.ImageID)
+
+	ta := ra.Task()
+	_, err = ta.AddTrashRemove(admin.NewImageSpec(pOpts.Pool, pOpts.RadosNamespace, pOpts.ImageID))
```
nixpanic: Please introduce a helper to get the `ImageSpec` from `rbdImage`. There should be only a single location where `admin.NewImageSpec()` is called.

```go
func (ri *rbdImage) getImageSpec() *admin.NewImageSpec {
    return admin.NewImageSpec(ri.Pool, ri.RadosNamespace, ri.ImageID)
}
```

internal/rbd/rbd_util.go
```diff
@@ -700,14 +683,17 @@ func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) er
...
-		"--id", cr.ID,
-		"--keyfile=" + cr.KeyFile,
-		"-m", pOpts.Monitors,
+	ra, err := pOpts.conn.GetRBDAdmin()
+	if err != nil {
+		return err
 	}
-	rbdCephMgrSupported, err := addRbdManagerTask(ctx, pOpts, args)
+
+	log.DebugLog(ctx, "rbd: adding task to remove image %s with id %s from trash", pOpts, pOpts.ImageID)
```
nixpanic: I am a big fan of using `%q` for quoted values in logs, it increases readability a little

nixpanic (changes requested):
just a nit to improve readability and maintenance of this change, looks good otherwise.

Rakshith-R:
@mergifyio rebase

internal/rbd/rbd_util.go
Rakshith-R: > Please introduce a helper to get the `ImageSpec` from `rbdImage`. There should be only a single location where `admin.NewImageSpec()` is called.
> 
> ```go
> func (ri *rbdImage) getImageSpec() *admin.NewImageSpec {
>     return admin.NewImageSpec(ri.Pool, ri.RadosNamespace, ri.ImageID)
> }
> ```

Actually task required imageID and flatten requires imageName, at go-ceph level it was decided to have single `imageSpec` but for ceph it is `imageIDSpec` and `imageSpec`. 

We would still need two methods, one for id and one for name. 
I prefer to keep it the same as it is now.

internal/rbd/rbd_util.go
```diff
@@ -651,74 +624,77 @@ func (rv *rbdVolume) ensureImageCleanup(ctx context.Context) error {
...
 		}
 	}
 
 	return nil
 }
 
 // deleteImage deletes a ceph image with provision and volume options.
-func deleteImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
-	image := pOpts.RbdImageName
+func (rv *rbdVolume) deleteImage(ctx context.Context) error {
```
Madhu-1, line 641: make this a method of `rbdImage` struct?

internal/rbd/rbd_util.go
```diff
@@ -651,74 +624,77 @@ func (rv *rbdVolume) ensureImageCleanup(ctx context.Context) error {
...
 	}
 
-	return trashRemoveImage(ctx, pOpts, cr)
+	return rv.trashRemoveImage(ctx)
 }
 
 // trashRemoveImage adds a task to trash remove an image using ceph manager if supported,
 // otherwise removes the image from trash.
-func trashRemoveImage(ctx context.Context, pOpts *rbdVolume, cr *util.Credentials) error {
+func (rv *rbdVolume) trashRemoveImage(ctx context.Context) error {
```
Madhu-1, line 677: same here make this method of `rbdImage` struct?

Rakshith-R:
@mergifyio rebase

Rakshith-R:
@Madhu-1 , moving those two methods to rbdImage struct would require moving a few more methods(getImageID, ensureTrashCleanup ...) onto rbdImage struct. 

I would prefer it be done in another pr, maybe we can decide on which methods should be on rbdImage and rbdVolume struct through a issue maybe before making the changes?

@ceph/ceph-csi-contributors PTAL

Rakshith-R:
@mergifyio rebase

internal/rbd/rbd_util.go
nixpanic: ah

> We would still need two methods, one for id and one for name.

That is a little sad. The current approach is hard to read, and (as you see with my comment). What is the reason for passing different parameters to `admin.NewImageSpec()`? In case one of the parameters is not set, can a helper function not detect that?

internal/rbd/rbd_util.go
nixpanic: Still pending? if there is a reason not to do that, document it.

Rakshith-R: Opened issue to track it https://github.com/ceph/ceph-csi/issues/2742

internal/rbd/rbd_util.go
Rakshith-R: > ah
> 
> > We would still need two methods, one for id and one for name.
> 
> That is a little sad. The current approach is hard to read, and (as you see with my comment). What is the reason for passing different parameters to `admin.NewImageSpec()`?

To avoid two almost similar types at go-ceph level 😂

> In case one of the parameters is not set, can a helper function not detect that?

That would be moving the logic inside a function with a boot parameter.

There's no way to automatically detect it. 
If another level of abstraction with helper function is needed, I'll add it but anyway we would have to decide whether it is imageid(trash ops) or imagename.

nixpanic (dismissed):

Madhu-1 (dismissed):

Madhu-1:
@humblec you have requested changes on this PR. PTAL

internal/rbd/migration.go
```diff
@@ -112,6 +112,14 @@ func genVolFromMigVolID(ctx context.Context, migVolID *migrationVolID, cr *util.
 
 		return nil, err
 	}
+	defer rv.Destroy()
+	// if connected , delete it
```
humblec: please remove the extra space.

internal/rbd/migration.go
```diff
@@ -112,6 +112,14 @@ func genVolFromMigVolID(ctx context.Context, migVolID *migrationVolID, cr *util.
...
+	err = rv.deleteImage(ctx)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to delete rbd image : %s, err: %v", rv.RbdImageName, err)
```
humblec: rbd image:

humblec (changes requested):

internal/rbd/rbd_util.go
```diff
@@ -828,14 +826,13 @@ func (rv *rbdVolume) flattenRbdImage(
 		return nil
 	}
 
-	ra, err := rv.conn.GetRBDAdmin()
+	log.DebugLog(ctx, "rbd: adding task to flatten image %s", rv)
```
humblec: image: ?

internal/rbd/rbd_util.go
```diff
@@ -871,10 +871,6 @@ func (rv *rbdVolume) flattenRbdImage(
 			"task manager does not support flatten,image will be flattened once hardlimit is reached: %v",
 			err)
 		if forceFlatten || depth >= hardlimit {
-			err = rv.Connect(cr)
```
humblec, line 876: The commit has a typo on `flattne`

internal/rbd/migration.go
nixpanic: ok!

internal/rbd/rbd_util.go
nixpanic: that is reported as #2742

internal/rbd/rbd_util.go
nixpanic: using `%q` instead, `:` is usually followed by an error

Madhu-1 (dismissed):

humblec (dismissed):

Rakshith-R:
few lines creeped in to migration.go during rebase, fixed that. 
PTAL

Madhu-1 (approved):

humblec:
> few lines creeped in to migration.go during rebase, fixed that. PTAL

Thanks @Rakshith-R , I will let @nixpanic to confirm once, as I heard previous changes were tested too.

nixpanic (approved):

humblec (approved):

Issue #2327, opened 2021-07-27
RBD: getImageInfo() on rbd images on which flattening is in progress throws `ErrImageNotFound`


Rakshith-R:
Running `getImageInfo()` may return `ErrImageNotFound` when rbd images are undergoing flattening operation. 
 https://github.com/ceph/ceph-csi/blob/43f753760b0800beb023de066ff33e850fb4189e/internal/rbd/rbd_util.go#L1292

We need to try to do cleanup when we encounter `ErrImageNotFound` since it maybe unreliable in such cases

<details>
<summary>Ceph exec logs</summary>

```
root@rook-ceph-tools-7b96766574-wh7sf /]# ceph rbd task add flatten replicapool/tmp
{"sequence": 9, "id": "e2a2df32-3edb-48c4-b538-a41985232e99", "message": "Flattening image replicapool/tmp", "refs": {"action": "flatten", "pool_name": "replicapool", "pool_namespace": "", "image_name": "tmp", "image_id": "41c9e3608f0b"}}
^[[A[root@rook-ceph-tools-7b96766574-wh7sf /]# rbd info replicapool/tmp
rbd image 'tmp':
        size 1 GiB in 256 objects
        order 22 (4 MiB objects)
        snapshot_count: 0
        id: 41c9e3608f0b
        block_name_prefix: rbd_data.41c9e3608f0b
        format: 2
        features: layering, deep-flatten, operations
        op_features: clone-child
        flags: 
        create_timestamp: Tue Jul 27 11:29:40 2021
        access_timestamp: Tue Jul 27 11:29:40 2021
        modify_timestamp: Tue Jul 27 11:29:40 2021
        parent: replicapool/csi-vol-673a80bb-eea1-11eb-80f6-0242ac110006@18f674b3-62f5-4f3c-b248-add99476c0c0
        overlap: 1 GiB
[root@rook-ceph-tools-7b96766574-wh7sf /]# rbd info replicapool/tmp
rbd image 'tmp':
        size 1 GiB in 256 objects
        order 22 (4 MiB objects)
        snapshot_count: 0
        id: 41c9e3608f0b
        block_name_prefix: rbd_data.41c9e3608f0b
        format: 2
        features: layering, deep-flatten, operations
        op_features: clone-child
        flags: 
        create_timestamp: Tue Jul 27 11:29:40 2021
        access_timestamp: Tue Jul 27 11:29:40 2021
        modify_timestamp: Tue Jul 27 11:29:40 2021
[...]

PR #2634, opened 2021-11-11
cephfs: detect corrupted FUSE mounts and try to restore them

### LINKED ISSUE #2616: Improve detection and recovery of broken ceph-fuse mounts

Broken FUSE-based mounts is a well known problem in CSI. While we can't guarantee that the fuse cephfs client won't fail (bugs in ceph-fuse or node plugin, OOM, upgrades, ...), we can make it easier for users to recover from this situation.

Currently, cephfs-csi mounts a cephfs volume in `NodeStageVolume` only, and then binds this to a target path in `NodePublishVolume`. If the Node Plugin container was restarted, it will lose the original FUSE mount as well as all the bindmounts, rendering all current and future Pods consuming the volume unable to access the volume. The only way for users to get out of this situation is to trigger `NodeUnstageVolume`, so that the volume can be properly remounted later. This would probably involve restarting the node, because at least in cephfs-csi v3.4.0, `NodeUnpublishVolume` can't finish with broken mounts ([`util.IsMountPoint()`](https://github.com/ceph/ceph-csi/blob/b95f3cdcbc1d3659afd48551e96744fd988db6bd/internal/cephfs/nodeserver.go#L294-L303) will keep failing).

Let's try to recover mountpoints in both `NodeStageVolume` and `NodePublishVolume`. In both we would check for broken mountpoints (`stat()` and check for `ENOTCONN`) and remount if necessary. Remounting is safe, because all Pods consuming the volume(s) have their mounts broken anyway.

I can work on this.

It would be possible to recover a ceph-fuse mount without needing to trigger `NodeUnstageVolume`->`NodeStageVolume`. The next Pod to mount the volume would be also the one to trigger mount recovery.

Note that existing Pods would remain in a bad state, only new ones would have their mounts working. Existing Pods would need to `NodeUnpublishVolume`->`NodePublishVolume`, which could be achieved only by deleting these Pods. We can discuss this or some other automatic mount reconciliation in a separate ticket.

* Create an app Pod that mounts a CephFS volume with `mounter: fuse`.
* Delete the Node plugin Pod (or kill the ceph-fuse process inside the container).
* Observe the app Pod not being able to access the volumes anymore (ends with an error, e.g. `Transport endpoint is not connected`).
* Create another app Pod, observe the volume being accessible again.

gman0:

This PR adds detection and recovery of corrupted mounts to `Node{Stage,Unstage}Volume` and `Node{Publish,Unpublish}Volume` procedures.

When a corruption is detected during volume stage/publish, the volume in question will be unmounted and mounted again, restoring it for workloads on the node. Note that these remounts are done only in case of FUSE.

When a corruption is detected during volume unstage/unpublish, the mountpoint will be unmounted as opposed to just returning an error.

The actual mount corruption is detected by `IsCorruptedMnt` here:

https://github.com/kubernetes/kubernetes/blob/ba944971f66210d09273ee779380d3e116f75880/staging/src/k8s.io/mount-utils/mount_helper_unix.go#L39-L57

In my experience, when a FUSE process exits, accessing the mountpoint results in `ENOTCONN`. The function above includes many other error codes. The more conservative way would be to check only for `ENOTCONN` for now and later add more errno's to the list of recognized errors. On the other hand, using the `IsCorruptedMnt` function catches errors that I haven't been able to trigger with my limited testing methods. Reviewers can decide which one is more sensible.

## Related issues ##

Fixes: #2616

## Future concerns ##

Mount recovery provided by this PR recovers mounts only in newly created Pods.

## Appendix - testing

This PR was tested by following steps:
* create a PVC with `mounter: fuse`,
* create a Deployment with two replicas mounting that PVC,
* delete the ceph-csi-cephfs Node plugin Pod,
* observe that running `$ stat /mnt` in the workload Pods results in an error: `stat: cannot stat '/mnt': Transport endpoint is not connected`,
* delete one of the Deployment replicas. This triggers `NodeUnpublishVolume` and `NodePublishVolume`, which in turn triggers mount recovery,
* observe that running `$ stat /mnt` in the new replica Pod succeeds.

### Before this PR

```
$ kubectl logs -l app=ceph-csi-cephfs,component=nodeplugin -c csi-cephfsplugin
I1111 13:38:51.373446       1 cephcsi.go:167] Driver version: v3.4.0 and Git version: 94ef181bc866c8ccaf507de66f213b26608309ff
I1111 13:38:51.373941       1 cephcsi.go:185] Initial PID limit is set to -1
I1111 13:38:51.374005       1 cephcsi.go:191] Reconfigured PID limit to -1 (max)
I1111 13:38:51.374011       1 cephcsi.go:212] Starting driver type: cephfs with name: cephfs.csi.ceph.com
I1111 13:38:51.387400       1 volumemounter.go:87] loaded mounter: kernel
I1111 13:38:51.400493       1 volumemounter.go:98] loaded mounter: fuse
I1111 13:38:51.401342       1 server.go:131] Listening for connections on address: &net.UnixAddr{Name:"//csi/csi.sock", Net:"unix"}
I1111 13:38:52.165347       1 utils.go:176] ID: 1 GRPC call: /csi.v1.Identity/GetPluginInfo
I1111 13:38:52.168260       1 utils.go:180] ID: 1 GRPC request: {}
I1111 13:38:52.168296       1 identityserver-default.go:38] ID: 1 Using default GetPluginInfo
I1111 13:38:52.168379       1 utils.go:187] ID: 1 GRPC response: {"name":"cephfs.csi.ceph.com","vendor_version":"v3.4.0"}
I1111 13:38:52.744387       1 utils.go:176] ID: 2 GRPC call: /csi.v1.Node/NodeGetInfo
I1111 13:38:52.744466       1 utils.go:180] ID: 2 GRPC request: {}
I1111 13:38:52.744479       1 nodeserver-default.go:60] ID: 2 Using default NodeGetInfo
I1111 13:38:52.744620       1 utils.go:187] ID: 2 GRPC response: {"accessible_topology":{},"node_id":"127.0.0.1"}
I1111 13:39:01.848210       1 utils.go:176] ID: 3 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC call: /csi.v1.Node/NodePublishVolume
I1111 13:39:01.848650       1 utils.go:180] ID: 3 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC request: {"secrets":"***stripped***","staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/globalmount","target_path":"/var/lib/kubelet/pods/656f7d40-7c02-4790-bf81-eb7c1d753fd3/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount","volume_capability":{"AccessType":{"Mount":{}},"access_mode":{"mode":5}},"volume_context":{"monitors":"188.184.74.8:6789","mounter":"fuse","provisionVolume":"false","rootPath":"/volumes/_nogroup/6dc56738-5c8a-42d4-bfc1-901a65cc84eb/c2bd88ca-f269-4d4c-8cf7-a5eb48e3eaa9"},"volume_id":"8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c"}
I1111 13:39:01.857917       1 mount_linux.go:173] Cannot run systemd-run, assuming non-systemd OS
I1111 13:39:01.857947       1 mount_linux.go:174] systemd-run failed with: exit status 1
I1111 13:39:01.857959       1 mount_linux.go:175] systemd-run output: System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to create bus connection: Host is down
I1111 13:39:01.864268       1 cephcmds.go:60] ID: 3 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c command succeeded: mount [-o bind,_netdev /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/globalmount /var/lib/kubelet/pods/656f7d40-7c02-4790-bf81-eb7c1d753fd3/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount]
I1111 13:39:01.864311       1 nodeserver.go:266] ID: 3 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c cephfs: successfully bind-mounted volume 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c to /var/lib/kubelet/pods/656f7d40-7c02-4790-bf81-eb7c1d753fd3/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount
I1111 13:39:01.864384       1 utils.go:187] ID: 3 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC response: {}
I1111 13:39:02.753991       1 utils.go:176] ID: 4 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC call: /csi.v1.Node/NodeUnpublishVolume
I1111 13:39:02.754148       1 utils.go:180] ID: 4 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC request: {"target_path":"/var/lib/kubelet/pods/5bc657be-656f-406b-bb23-962218259034/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount","volume_id":"8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c"}
I1111 13:39:02.762338       1 mount_linux.go:173] Cannot run systemd-run, assuming non-systemd OS
I1111 13:39:02.762359       1 mount_linux.go:174] systemd-run failed with: exit status 1
I1111 13:39:02.762369       1 mount_linux.go:175] systemd-run output: System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to create bus connection: Host is down
E1111 13:39:02.762485       1 utils.go:185] ID: 4 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC error: rpc error: code = Internal desc = stat /var/lib/kubelet/pods/5bc657be-656f-406b-bb23-962218259034/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount: transport endpoint is not connected
I1111 13:39:03.357951       1 utils.go:176] ID: 5 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC call: /csi.v1.Node/NodeUnpublishVolume
I1111 13:39:03.358107       1 utils.go:180] ID: 5 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC request: {"target_path":"/var/lib/kubelet/pods/5bc657be-656f-406b-bb23-962218259034/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount","volume_id":"8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c"}
I1111 13:39:03.365459       1 mount_linux.go:173] Cannot run systemd-run, assuming non-systemd OS
I1111 13:39:03.365477       1 mount_linux.go:174] systemd-run failed with: exit status 1
I1111 13:39:03.365485       1 mount_linux.go:175] systemd-run output: System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to create bus connection: Host is down
E1111 13:39:03.365533       1 utils.go:185] ID: 5 Req-ID: 8f9e7e53-4c25-4e74-9e1a-6507e32bbc5c GRPC error: rpc error: code = Internal desc = stat /var/lib/kubelet/pods/5bc657be-656f-406b-bb23-962218259034/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount: transport endpoint is not connected
... repeated calls to NodeUnpublishVolume that never succeed ...
```

* `NodePublishVolume` allegedly succeeded, but the Pod won't come up as it end ups in CrashLoopBackOff state. `kubectl describe` reports following error:
   `  Warning  Failed     81s (x5 over 2m44s)   kubelet            Error: failed to start container "test": Error response from daemon: error while creating mount source path '/var/lib/kubelet/pods/656f7d40-7c02-4790-bf81-eb7c1d753fd3/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount': mkdir /var/lib/kubelet/pods/656f7d40-7c02-4790-bf81-eb7c1d753fd3/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount: file exists`
* Running `$ stat /var/lib/kubelet/pods/656f7d40-7c02-4790-bf81-eb7c1d753fd3/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount` on the node reveals the actual problem:
   `stat: cannot statx '/var/lib/kubelet/pods/656f7d40-7c02-4790-bf81-eb7c1d753fd3/volumes/kubernetes.io~csi/pvc-d89978df-6549-4fb0-a72b-0b610c1a8e0c/mount': Transport endpoint is not connected`
* `NodeUnpublishVolume` will never succeed because [`util.IsMountPoint()`](https://github.com/ceph/ceph-csi/blob/b95f3cdcbc1d3659afd48551e96744fd988db6bd/internal/cephfs/nodeserver.go#L294-L303) will keep failing. Admin must unmount the volume on the node manually or restart the node.

### Node plugin logs with this PR

```
$ kubectl logs -l app=ceph-csi-cephfs,component=nodeplugin -c csi-cephfsplugin
I1111 08:58:17.802425       1 cephcsi.go:168] Driver version: canary and Git version: af752dd38f8e74d84987ab14f2ac489a10bb635c
I1111 08:58:17.802753       1 cephcsi.go:186] Initial PID limit is set to -1
I1111 08:58:17.802834       1 cephcsi.go:192] Reconfigured PID limit to -1 (max)
I1111 08:58:17.802859       1 cephcsi.go:217] Starting driver type: cephfs with name: cephfs.csi.ceph.com
I1111 08:58:17.814902       1 volumemounter.go:79] loaded mounter: kernel
I1111 08:58:17.826562       1 volumemounter.go:90] loaded mounter: fuse
I1111 08:58:17.827027       1 server.go:131] Listening for connections on address: &net.UnixAddr{Name:"//csi/csi.sock", Net:"unix"}
I1111 08:58:18.611849       1 utils.go:177] ID: 1 GRPC call: /csi.v1.Identity/GetPluginInfo
I1111 08:58:18.614095       1 utils.go:181] ID: 1 GRPC request: {}
I1111 08:58:18.614124       1 identityserver-default.go:38] ID: 1 Using default GetPluginInfo
I1111 08:58:18.614190       1 utils.go:188] ID: 1 GRPC response: {"name":"cephfs.csi.ceph.com","vendor_version":"canary"}
I1111 08:58:18.652409       1 utils.go:177] ID: 2 GRPC call: /csi.v1.Node/NodeGetInfo
I1111 08:58:18.652472       1 utils.go:181] ID: 2 GRPC request: {}
I1111 08:58:18.652494       1 nodeserver-default.go:60] ID: 2 Using default NodeGetInfo
I1111 08:58:18.652598       1 utils.go:188] ID: 2 GRPC response: {"accessible_topology":{},"node_id":"127.0.0.1"}
I1111 08:58:41.227645       1 utils.go:177] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 GRPC call: /csi.v1.Node/NodePublishVolume
I1111 08:58:41.227909       1 utils.go:181] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 GRPC request: {"secrets":"***stripped***","staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/globalmount","target_path":"/var/lib/kubelet/pods/e1c0eb2c-95c4-4fca-85fc-abb185ecd5a9/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount","volume_capability":{"AccessType":{"Mount":{}},"access_mode":{"mode":5}},"volume_context":{"monitors":"188.184.74.8:6789","mounter":"fuse","provisionVolume":"false","rootPath":"/volumes/_nogroup/0a8a7a03-61a9-4bda-bd47-067605c11b7a/a00fcddc-617d-4695-9fa7-0885855b3a31"},"volume_id":"49670f1b-26de-4feb-887f-b74eb6ab4488"}
I1111 08:58:41.227986       1 nodeserver.go:528] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 cephfs: detected corrupted FUSE mount in staging target path /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/globalmount during NodePublishVolume, remounting
I1111 08:58:41.239076       1 cephcmds.go:62] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 command succeeded: umount [/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/globalmount]
I1111 08:58:41.239276       1 nodeserver.go:193] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 cephfs: mounting volume 49670f1b-26de-4feb-887f-b74eb6ab4488 with Ceph FUSE driver
I1111 08:58:41.295946       1 cephcmds.go:62] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 command succeeded: ceph-fuse [/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/globalmount -m 188.184.74.8:6789 -c /etc/ceph/ceph.conf -n client.pvc-1e705b75-af6e-4827-ab45-90f2563539f2 --keyfile=***stripped*** -r /volumes/_nogroup/0a8a7a03-61a9-4bda-bd47-067605c11b7a/a00fcddc-617d-4695-9fa7-0885855b3a31 -o nonempty]
I1111 08:59:04.413678       1 cephcmds.go:62] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 command succeeded: mount [-o bind,_netdev /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/globalmount /var/lib/kubelet/pods/e1c0eb2c-95c4-4fca-85fc-abb185ecd5a9/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount]
I1111 08:59:04.413750       1 nodeserver.go:341] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 cephfs: successfully bind-mounted volume 49670f1b-26de-4feb-887f-b74eb6ab4488 to /var/lib/kubelet/pods/e1c0eb2c-95c4-4fca-85fc-abb185ecd5a9/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount
I1111 08:59:04.413937       1 utils.go:188] ID: 3 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 GRPC response: {}
I1111 08:59:04.495789       1 utils.go:177] ID: 4 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 GRPC call: /csi.v1.Node/NodeUnpublishVolume
I1111 08:59:04.495927       1 utils.go:181] ID: 4 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 GRPC request: {"target_path":"/var/lib/kubelet/pods/2df62d3a-b329-4f4a-8483-e98a563463eb/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount","volume_id":"49670f1b-26de-4feb-887f-b74eb6ab4488"}
E1111 08:59:04.495982       1 nodeserver.go:360] ID: 4 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 stat failed: stat /var/lib/kubelet/pods/2df62d3a-b329-4f4a-8483-e98a563463eb/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount: transport endpoint is not connected
I1111 08:59:04.495994       1 nodeserver.go:374] ID: 4 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 cephfs: detected corrupted mount in publish target path /var/lib/kubelet/pods/2df62d3a-b329-4f4a-8483-e98a563463eb/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount, trying to unmount anyway
I1111 08:59:04.504291       1 cephcmds.go:62] ID: 4 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 command succeeded: umount [/var/lib/kubelet/pods/2df62d3a-b329-4f4a-8483-e98a563463eb/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount]
I1111 08:59:04.504397       1 nodeserver.go:396] ID: 4 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 cephfs: successfully unbounded volume 49670f1b-26de-4feb-887f-b74eb6ab4488 from /var/lib/kubelet/pods/2df62d3a-b329-4f4a-8483-e98a563463eb/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount
I1111 08:59:04.504473       1 utils.go:188] ID: 4 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 GRPC response: {}
```

---

<details>

  progress or results

</details>

### DIFF
--- a/deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
+++ b/deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
@@ -100,6 +100,8 @@ spec:
               mountPath: /etc/ceph-csi-config/
             - name: keys-tmp-dir
               mountPath: /tmp/csi/keys
+            - name: ceph-csi-mountinfo
+              mountPath: /csi/mountinfo
         - name: liveness-prometheus
           securityContext:
             privileged: true
@@ -164,6 +166,10 @@ spec:
           emptyDir: {
             medium: "Memory"
           }
+        - name: ceph-csi-mountinfo
+          hostPath:
+            path: /var/lib/kubelet/plugins/cephfs.csi.ceph.com/mountinfo
+            type: DirectoryOrCreate
 ---
 # This is a service to expose the liveness metrics
 apiVersion: v1
--- a/docs/ceph-fuse-corruption.md
+++ b/docs/ceph-fuse-corruption.md
@@ -1,0 +1,45 @@
+# ceph-fuse: detection of corrupted mounts and their recovery
+
+Mounts managed by ceph-fuse may get corrupted by e.g. the ceph-fuse process
+exiting abruptly, or its parent Node Plugin container being terminated, taking
+down its child processes with it.
+
+This may manifest in concerned workloads like so:
+
+```
+# mount | grep fuse
+ceph-fuse on /cephfs-share type fuse.ceph-fuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other)
+# ls /cephfs-share
+ls: /cephfs-share: Socket not connected
+```
+
+or,
+
+```
+# stat /home/kubelet/pods/ae344b80-3b07-4589-b1a1-ca75fa9debf2/volumes/kubernetes.io~csi/pvc-ec69de59-7823-4840-8eee-544f8261fef0/mount: transport endpoint is not connected
+```
+
+This feature allows CSI CephFS plugin to be able to detect if a ceph-fuse mount
+is corrupted during the volume publishing phase, and will attempt to recover it
+for the newly scheduled pod. Pods that already reside on a node whose
+ceph-fuse mountpoints were broken may still need to be restarted, however.
+
+## Detection
+
+A mountpoint is deemed corrupted if `stat()`-ing it returns one of the
+following errors:
+
+* `ENOTCONN`
+* `ESTALE`
+* `EIO`
+* `EACCES`
+* `EHOSTDOWN`
+
+## Recovery
+
+Once a mountpoint corruption is detected, its recovery is performed by
+remounting the volume associated with it.
+
+Recovery is attempted only if `/csi/mountinfo` directory is made available to
+CSI CephFS plugin (available by default in the Helm chart and Kubernetes
+manifests).
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -280,6 +280,7 @@ var _ = Describe("cephfs", func() {
 		It("Test CephFS CSI", func() {
 			pvcPath := cephFSExamplePath + "pvc.yaml"
 			appPath := cephFSExamplePath + "pod.yaml"
+			deplPath := cephFSExamplePath + "deployment.yaml"
 			appRWOPPath := cephFSExamplePath + "pod-rwop.yaml"
 			pvcClonePath := cephFSExamplePath + "pvc-restore.yaml"
 			pvcSmartClonePath := cephFSExamplePath + "pvc-clone.yaml"
@@ -497,6 +498,134 @@ var _ = Describe("cephfs", func() {
 				err = validatePVCAndAppBinding(pvcPath, appPath, f)
 				if err != nil {
 					e2elog.Failf("failed to validate CephFS pvc and application binding: %v", err)
+				}
+				err = deleteResource(cephFSExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete CephFS storageclass: %v", err)
+				}
+			})
+
+			By("verifying that ceph-fuse recovery works for new pods", func() {
+				err := deleteResource(cephFSExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete CephFS storageclass: %v", err)
+				}
+				err = createCephfsStorageClass(f.ClientSet, f, true, map[string]string{
+					"mounter": "fuse",
+				})
+				if err != nil {
+					e2elog.Failf("failed to create CephFS storageclass: %v", err)
+				}
+				replicas := int32(2)
+				pvc, depl, err := validatePVCAndDeploymentAppBinding(
+					f, pvcPath, deplPath, f.UniqueName, &replicas, deployTimeout,
+				)
+				if err != nil {
+					e2elog.Failf("failed to create PVC and Deployment: %v", err)
+				}
+				deplPods, err := listPods(f, depl.Namespace, &metav1.ListOptions{
+					LabelSelector: fmt.Sprintf("app=%s", depl.Labels["app"]),
+				})
+				if err != nil {
+					e2elog.Failf("failed to list pods for Deployment: %v", err)
+				}
+
+				doStat := func(podName string) (stdErr string, err error) {
+					_, stdErr, err = execCommandInContainerByPodName(
+						f,
+						fmt.Sprintf("stat %s", depl.Spec.Template.Spec.Containers[0].VolumeMounts[0].MountPath),
+						depl.Namespace,
+						podName,
+						depl.Spec.Template.Spec.Containers[0].Name,
+					)
+
+					return stdErr, err
+				}
+				ensureStatSucceeds := func(podName string) error {
+					stdErr, statErr := doStat(podName)
+					if statErr != nil || stdErr != "" {
+						return fmt.Errorf(
+							"expected stat to succeed without error output ; got err %w, stderr %s",
+							statErr, stdErr,
+						)
+					}
+
+					return nil
+				}
+
+				pod1Name, pod2Name := deplPods[0].Name, deplPods[1].Name
+
+				// stat() ceph-fuse mountpoints to make sure they are working.
+				for i := range deplPods {
+					err = ensureStatSucceeds(deplPods[i].Name)
+					if err != nil {
+						e2elog.Failf(err.Error())
+					}
+				}
+				// Kill ceph-fuse in cephfs-csi node plugin Pods.
+				nodePluginSelector, err := getDaemonSetLabelSelector(f, cephCSINamespace, cephFSDeamonSetName)
+				if err != nil {
+					e2elog.Failf("failed to get node plugin DaemonSet label selector: %v", err)
+				}
+				_, stdErr, err := execCommandInContainer(
+					f, "killall -9 ceph-fuse", cephCSINamespace, "csi-cephfsplugin", &metav1.ListOptions{
+						LabelSelector: nodePluginSelector,
+					},
+				)
+				if err != nil {
+					e2elog.Failf("killall command failed: err %v, stderr %s", err, stdErr)
+				}
+				// Verify Pod podName2 that stat()-ing the mountpoint results in ENOTCONN.
+				stdErr, err = doStat(pod2Name)
+				if err == nil || !strings.Contains(stdErr, "not connected") {
+					e2elog.Failf(
+						"expected stat to fail with 'Transport endpoint not connected' or 'Socket not connected'; got err %v, stderr %s",
+						err, stdErr,
+					)
+				}
+				// Delete podName2 Pod. This serves two purposes: it verifies that deleting pods with
+				// corrupted ceph-fuse mountpoints works, and it lets the replicaset controller recreate
+				// the pod with hopefully mounts working again.
+				err = deletePod(pod2Name, depl.Namespace, c, deployTimeout)
+				if err != nil {
+					e2elog.Failf(err.Error())
+				}
+				// Wait for the second Pod to be recreated.
+				err = waitForDeploymentComplete(c, depl.Name, depl.Namespace, deployTimeout)
+				if err != nil {
+					e2elog.Failf(err.Error())
+				}
+				// List Deployment's pods again to get name of the new pod.
+				deplPods, err = listPods(f, depl.Namespace, &metav1.ListOptions{
+					LabelSelector: fmt.Sprintf("app=%s", depl.Labels["app"]),
+				})
+				if err != nil {
+					e2elog.Failf("failed to list pods for Deployment: %v", err)
+				}
+				for i := range deplPods {
+					if deplPods[i].Name != pod1Name {
+						pod2Name = deplPods[i].Name
+
+						break
+					}
+				}
+				if pod2Name == "" {
+					podNames := make([]string, len(deplPods))
+					for i := range deplPods {
+						podNames[i] = deplPods[i].Name
+					}
+					e2elog.Failf("no new replica found ; found replicas %v", podNames)
+				}
+				// Verify Pod podName3 has its ceph-fuse mount working again.
+				err = ensureStatSucceeds(pod2Name)
+				if err != nil {
+					e2elog.Failf(err.Error())
+				}
+
+				// Delete created resources.
+				err = deletePVCAndDeploymentApp(f, pvc, depl)
+				if err != nil {
+					e2elog.Failf("failed to delete PVC and Deployment: %v", err)
 				}
 				err = deleteResource(cephFSExamplePath + "storageclass.yaml")
 				if err != nil {
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -191,19 +191,12 @@ func createPVCAndApp(
 	return err
 }
 
-// createPVCAndDeploymentApp creates pvc and deployment, if name is not empty
-// same will be set as pvc and app name.
+// createPVCAndDeploymentApp creates pvc and deployment.
 func createPVCAndDeploymentApp(
 	f *framework.Framework,
-	name string,
 	pvc *v1.PersistentVolumeClaim,
 	app *appsv1.Deployment,
 	pvcTimeout int) error {
-	if name != "" {
-		pvc.Name = name
-		app.Name = name
-		app.Spec.Template.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = name
-	}
 	err := createPVCAndvalidatePV(f.ClientSet, pvc, pvcTimeout)
 	if err != nil {
 		return err
@@ -213,19 +206,50 @@ func createPVCAndDeploymentApp(
 	return err
 }
 
-// DeletePVCAndDeploymentApp deletes pvc and deployment, if name is not empty
-// same will be set as pvc and app name.
-func deletePVCAndDeploymentApp(
+// validatePVCAndDeploymentAppBinding creates PVC and Deployment, and waits until
+// all its replicas are Running. Use `replicas` to override default number of replicas
+// defined in `deploymentPath` Deployment manifest.
+func validatePVCAndDeploymentAppBinding(
 	f *framework.Framework,
-	name string,
-	pvc *v1.PersistentVolumeClaim,
-	app *appsv1.Deployment) error {
-	if name != "" {
-		pvc.Name = name
-		app.Name = name
-		app.Spec.Template.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = name
+	pvcPath string,
+	deploymentPath string,
+	namespace string,
+	replicas *int32,
+	pvcTimeout int,
+) (*v1.PersistentVolumeClaim, *appsv1.Deployment, error) {
+	pvc, err := loadPVC(pvcPath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to load PVC: %w", err)
+	}
+	pvc.Namespace = namespace
+
+	depl, err := loadAppDeployment(deploymentPath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to load Deployment: %w", err)
+	}
+	depl.Namespace = f.UniqueName
+	if replicas != nil {
+		depl.Spec.Replicas = replicas
 	}
 
+	err = createPVCAndDeploymentApp(f, pvc, depl, pvcTimeout)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	err = waitForDeploymentComplete(f.ClientSet, depl.Name, depl.Namespace, deployTimeout)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return pvc, depl, nil
+}
+
+// DeletePVCAndDeploymentApp deletes pvc and deployment.
+func deletePVCAndDeploymentApp(
+	f *framework.Framework,
+	pvc *v1.PersistentVolumeClaim,
+	app *appsv1.Deployment) error {
 	err := deleteDeploymentApp(f.ClientSet, app.Name, app.Namespace, deployTimeout)
 	if err != nil {
 		return err
--- a/internal/cephfs/fuserecovery.go
+++ b/internal/cephfs/fuserecovery.go
@@ -1,0 +1,273 @@
+/*
+Copyright 2022 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cephfs
+
+import (
+	"context"
+
+	"github.com/ceph/ceph-csi/internal/cephfs/mounter"
+	"github.com/ceph/ceph-csi/internal/cephfs/store"
+	fsutil "github.com/ceph/ceph-csi/internal/cephfs/util"
+	"github.com/ceph/ceph-csi/internal/util"
+	"github.com/ceph/ceph-csi/internal/util/log"
+
+	mountutil "k8s.io/mount-utils"
+)
+
+type (
+	mountState int
+)
+
+const (
+	msUnknown mountState = iota
+	msNotMounted
+	msMounted
+	msCorrupted
+
+	// ceph-fuse fsType in /proc/<PID>/mountinfo.
+	cephFuseFsType = "fuse.ceph-fuse"
+)
+
+func (ms mountState) String() string {
+	return [...]string{
+		"UNKNOWN",
+		"NOT_MOUNTED",
+		"MOUNTED",
+		"CORRUPTED",
+	}[int(ms)]
+}
+
+func getMountState(path string) (mountState, error) {
+	isMnt, err := util.IsMountPoint(path)
+	if err != nil {
+		if util.IsCorruptedMountError(err) {
+			return msCorrupted, nil
+		}
+
+		return msUnknown, err
+	}
+
+	if isMnt {
+		return msMounted, nil
+	}
+
+	return msNotMounted, nil
+}
+
+func findMountinfo(mountpoint string, mis []mountutil.MountInfo) int {
+	for i := range mis {
+		if mis[i].MountPoint == mountpoint {
+			return i
+		}
+	}
+
+	return -1
+}
+
+// Ensures that given mountpoint is of specified fstype.
+// Returns true if fstype matches, or if no such mountpoint exists.
+func validateFsType(mountpoint, fsType string, mis []mountutil.MountInfo) bool {
+	if idx := findMountinfo(mountpoint, mis); idx > 0 {
+		mi := mis[idx]
+
+		if mi.FsType != fsType {
+			return false
+		}
+	}
+
+	return true
+}
+
+// tryRestoreFuseMountsInNodePublish tries to restore staging and publish
+// volume moutpoints inside the NodePublishVolume call.
+//
+// Restoration is performed in following steps:
+// 1. Detection: staging target path must be a working mountpoint, and target
+//    path must not be a corrupted mountpoint (see getMountState()). If either
+//    of those checks fail, mount recovery is performed.
+// 2. Recovery preconditions:
+//    * NodeStageMountinfo is present for this volume,
+//    * if staging target path and target path are mountpoints, they must be
+//      managed by ceph-fuse,
+//    * VolumeOptions.Mounter must evaluate to "fuse".
+// 3. Recovery:
+//    * staging target path is unmounted and mounted again using ceph-fuse,
+//    * target path is only unmounted; NodePublishVolume is then expected to
+//      continue normally.
+func (ns *NodeServer) tryRestoreFuseMountsInNodePublish(
+	ctx context.Context,
+	volID fsutil.VolumeID,
+	stagingTargetPath string,
+	targetPath string,
+	volContext map[string]string,
+) error {
+	// Check if there is anything to restore.
+
+	stagingTargetMs, err := getMountState(stagingTargetPath)
+	if err != nil {
+		return err
+	}
+
+	targetMs, err := getMountState(targetPath)
+	if err != nil {
+		return err
+	}
+
+	if stagingTargetMs == msMounted && targetMs != msCorrupted {
+		// Mounts seem to be fine.
+		return nil
+	}
+
+	// Something is broken. Try to proceed with mount recovery.
+
+	log.WarningLog(ctx, "cephfs: mount problem detected when publishing a volume: %s is %s, %s is %s; attempting recovery",
+		stagingTargetPath, stagingTargetMs, targetPath, targetMs)
+
+	// NodeStageMountinfo entry must be present for this volume.
+
+	var nsMountinfo *fsutil.NodeStageMountinfo
+
+	if nsMountinfo, err = fsutil.GetNodeStageMountinfo(volID); err != nil {
+		return err
+	} else if nsMountinfo == nil {
+		log.WarningLog(ctx, "cephfs: cannot proceed with mount recovery because NodeStageMountinfo record is missing")
+
+		return nil
+	}
+
+	// Check that the existing stage and publish mounts for this volume are
+	// managed by ceph-fuse, and that the mounter is of the FuseMounter type.
+	// Then try to restore them.
+
+	var (
+		volMounter mounter.VolumeMounter
+		volOptions *store.VolumeOptions
+	)
+
+	procMountInfo, err := util.ReadMountInfoForProc("self")
+	if err != nil {
+		return err
+	}
+
+	if !validateFsType(stagingTargetPath, cephFuseFsType, procMountInfo) ||
+		!validateFsType(targetPath, cephFuseFsType, procMountInfo) {
+		// We can't restore mounts not managed by ceph-fuse.
+		log.WarningLog(ctx, "cephfs: cannot proceed with mount recovery on non-FUSE mountpoints")
+
+		return nil
+	}
+
+	volOptions, err = ns.getVolumeOptions(ctx, volID, volContext, nsMountinfo.Secrets)
+	if err != nil {
+		return err
+	}
+
+	volMounter, err = mounter.New(volOptions)
+	if err != nil {
+		return err
+	}
+
+	if _, ok := volMounter.(*mounter.FuseMounter); !ok {
+		// We can't restore mounts with non-FUSE mounter.
+		log.WarningLog(ctx, "cephfs: cannot proceed with mount recovery with non-FUSE mounter")
+
+		return nil
+	}
+
+	// Try to restore mount in staging target path.
+	// Unmount and mount the volume.
+
+	if stagingTargetMs != msMounted {
+		if err := mounter.UnmountVolume(ctx, stagingTargetPath); err != nil {
+			return err
+		}
+
+		if err := ns.mount(
+			ctx,
+			volMounter,
+			volOptions,
+			volID,
+			stagingTargetPath,
+			nsMountinfo.Secrets,
+			nsMountinfo.VolumeCapability,
+		); err != nil {
+			return err
+		}
+	}
+
+	// Try to restore mount in target path.
+	// Only unmount the bind mount. NodePublishVolume should then
+	// create the bind mount by itself.
+
+	if err := mounter.UnmountVolume(ctx, targetPath); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Try to restore FUSE mount of the staging target path in NodeStageVolume.
+// If corruption is detected, try to only unmount the volume. NodeStageVolume
+// should be able to continue with mounting the volume normally afterwards.
+func (ns *NodeServer) tryRestoreFuseMountInNodeStage(
+	ctx context.Context,
+	mnt mounter.VolumeMounter,
+	stagingTargetPath string,
+) error {
+	// Check if there is anything to restore.
+
+	stagingTargetMs, err := getMountState(stagingTargetPath)
+	if err != nil {
+		return err
+	}
+
+	if stagingTargetMs != msCorrupted {
+		// Mounts seem to be fine.
+		return nil
+	}
+
+	// Something is broken. Try to proceed with mount recovery.
+
+	log.WarningLog(ctx, "cephfs: mountpoint problem detected when staging a volume: %s is %s; attempting recovery",
+		stagingTargetPath, stagingTargetMs)
+
+	// Check that the existing stage mount for this volume is  managed by
+	// ceph-fuse, and that the mounter is FuseMounter. Then try to restore them.
+
+	procMountInfo, err := util.ReadMountInfoForProc("self")
+	if err != nil {
+		return err
+	}
+
+	if !validateFsType(stagingTargetPath, cephFuseFsType, procMountInfo) {
+		// We can't restore mounts not managed by ceph-fuse.
+		log.WarningLog(ctx, "cephfs: cannot proceed with mount recovery on non-FUSE mountpoints")
+
+		return nil
+	}
+
+	if _, ok := mnt.(*mounter.FuseMounter); !ok {
+		// We can't restore mounts with non-FUSE mounter.
+		log.WarningLog(ctx, "cephfs: cannot proceed with mount recovery with non-FUSE mounter")
+
+		return nil
+	}
+
+	// Restoration here means only unmounting the volume.
+	// NodeStageVolume should take care of the rest.
+	return mounter.UnmountVolume(ctx, stagingTargetPath)
+}
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -47,11 +47,10 @@ type NodeServer struct {
 
 func getCredentialsForVolume(
 	volOptions *store.VolumeOptions,
-	req *csi.NodeStageVolumeRequest) (*util.Credentials, error) {
+	secrets map[string]string) (*util.Credentials, error) {
 	var (
-		err     error
-		cr      *util.Credentials
-		secrets = req.GetSecrets()
+		err error
+		cr  *util.Credentials
 	)
 
 	if volOptions.ProvisionVolume {
@@ -64,7 +63,7 @@ func getCredentialsForVolume(
 	} else {
 		// The volume is pre-made, credentials are in node stage secrets
 
-		cr, err = util.NewUserCredentials(req.GetSecrets())
+		cr, err = util.NewUserCredentials(secrets)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get user credentials from node stage secrets: %w", err)
 		}
@@ -73,11 +72,38 @@ func getCredentialsForVolume(
 	return cr, nil
 }
 
+func (ns *NodeServer) getVolumeOptions(
+	ctx context.Context,
+	volID fsutil.VolumeID,
+	volContext,
+	volSecrets map[string]string,
+) (*store.VolumeOptions, error) {
+	volOptions, _, err := store.NewVolumeOptionsFromVolID(ctx, string(volID), volContext, volSecrets)
+	if err != nil {
+		if !errors.Is(err, cerrors.ErrInvalidVolID) {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		volOptions, _, err = store.NewVolumeOptionsFromStaticVolume(string(volID), volContext)
+		if err != nil {
+			if !errors.Is(err, cerrors.ErrNonStaticVolume) {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+
+			volOptions, _, err = store.NewVolumeOptionsFromMonitorList(string(volID), volContext, volSecrets)
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+		}
+	}
+
+	return volOptions, nil
+}
+
 // NodeStageVolume mounts the volume to a staging path on the node.
 func (ns *NodeServer) NodeStageVolume(
 	ctx context.Context,
 	req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
-	var volOptions *store.VolumeOptions
 	if err := util.ValidateNodeStageVolumeRequest(req); err != nil {
 		return nil, err
 	}
@@ -94,28 +120,18 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 	defer ns.VolumeLocks.Release(req.GetVolumeId())
 
-	volOptions, _, err := store.NewVolumeOptionsFromVolID(ctx, string(volID), req.GetVolumeContext(), req.GetSecrets())
+	volOptions, err := ns.getVolumeOptions(ctx, volID, req.GetVolumeContext(), req.GetSecrets())
 	if err != nil {
-		if !errors.Is(err, cerrors.ErrInvalidVolID) {
-			return nil, status.Error(codes.Internal, err.Error())
-		}
-
-		// gets mon IPs from the supplied cluster info
-		volOptions, _, err = store.NewVolumeOptionsFromStaticVolume(string(volID), req.GetVolumeContext())
-		if err != nil {
-			if !errors.Is(err, cerrors.ErrNonStaticVolume) {
-				return nil, status.Error(codes.Internal, err.Error())
-			}
-
-			// get mon IPs from the volume context
-			volOptions, _, err = store.NewVolumeOptionsFromMonitorList(string(volID), req.GetVolumeContext(),
-				req.GetSecrets())
-			if err != nil {
-				return nil, status.Error(codes.Internal, err.Error())
-			}
-		}
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 	defer volOptions.Destroy()
+
+	mnt, err := mounter.New(volOptions)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to create mounter for volume %s: %v", volID, err)
+
+		return nil, status.Error(codes.Internal, err.Error())
+	}
 
 	// Check if the volume is already mounted
 
@@ -133,7 +149,16 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 
 	// It's not, mount now
-	if err = ns.mount(ctx, volOptions, req); err != nil {
+
+	if err = ns.mount(
+		ctx,
+		mnt,
+		volOptions,
+		fsutil.VolumeID(req.GetVolumeId()),
+		req.GetStagingTargetPath(),
+		req.GetSecrets(),
+		req.GetVolumeCapability(),
+	); err != nil {
 		return nil, err
 	}
 
@@ -142,11 +167,16 @@ func (ns *NodeServer) NodeStageVolume(
 	return &csi.NodeStageVolumeResponse{}, nil
 }
 
-func (*NodeServer) mount(ctx context.Context, volOptions *store.VolumeOptions, req *csi.NodeStageVolumeRequest) error {
-	stagingTargetPath := req.GetStagingTargetPath()
-	volID := fsutil.VolumeID(req.GetVolumeId())
-
-	cr, err := getCredentialsForVolume(volOptions, req)
+func (*NodeServer) mount(
+	ctx context.Context,
+	mnt mounter.VolumeMounter,
+	volOptions *store.VolumeOptions,
+	volID fsutil.VolumeID,
+	stagingTargetPath string,
+	secrets map[string]string,
+	volCap *csi.VolumeCapability,
+) error {
+	cr, err := getCredentialsForVolume(volOptions, secrets)
 	if err != nil {
 		log.ErrorLog(ctx, "failed to get ceph credentials for volume %s: %v", volID, err)
 
@@ -154,20 +184,13 @@ func (*NodeServer) mount(ctx context.Context, volOptions *store.VolumeOptions, r
 	}
 	defer cr.DeleteCredentials()
 
-	m, err := mounter.New(volOptions)
-	if err != nil {
-		log.ErrorLog(ctx, "failed to create mounter for volume %s: %v", volID, err)
-
-		return status.Error(codes.Internal, err.Error())
-	}
-
-	log.DebugLog(ctx, "cephfs: mounting volume %s with %s", volID, m.Name())
+	log.DebugLog(ctx, "cephfs: mounting volume %s with %s", volID, mnt.Name())
 
 	readOnly := "ro"
 
-	if req.VolumeCapability.AccessMode.Mode == csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY ||
-		req.VolumeCapability.AccessMode.Mode == csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY {
-		switch m.(type) {
+	if volCap.AccessMode.Mode == csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY ||
+		volCap.AccessMode.Mode == csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY {
+		switch mnt.(type) {
 		case *mounter.FuseMounter:
 			if !csicommon.MountOptionContains(strings.Split(volOptions.FuseMountOptions, ","), readOnly) {
 				volOptions.FuseMountOptions = util.MountOptionsAdd(volOptions.FuseMountOptions, readOnly)
@@ -179,7 +202,7 @@ func (*NodeServer) mount(ctx context.Context, volOptions *store.VolumeOptions, r
 		}
 	}
 
-	if err = m.Mount(ctx, stagingTargetPath, cr, volOptions); err != nil {
+	if err = mnt.Mount(ctx, stagingTargetPath, cr, volOptions); err != nil {
 		log.ErrorLog(ctx,
 			"failed to mount volume %s: %v Check dmesg logs if required.",
 			volID,
@@ -201,8 +224,9 @@ func (ns *NodeServer) NodePublishVolume(
 		return nil, err
 	}
 
+	stagingTargetPath := req.GetStagingTargetPath()
 	targetPath := req.GetTargetPath()
-	volID := req.GetVolumeId()
+	volID := fsutil.VolumeID(req.GetVolumeId())
 
 	// Considering kubelet make sure the stage and publish operations
 	// are serialized, we dont need any extra locking in nodePublish
@@ -238,8 +262,8 @@ func (ns *NodeServer) NodePublishVolume(
 
 	if err = mounter.BindMount(
 		ctx,
-		req.GetStagingTargetPath(),
-		req.GetTargetPath(),
+		stagingTargetPath,
+		targetPath,
 		req.GetReadonly(),
 		mountOptions); err != nil {
 		log.ErrorLog(ctx, "failed to bind-mount volume %s: %v", volID, err)
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -135,6 +135,10 @@ func (ns *NodeServer) NodeStageVolume(
 
 	// Check if the volume is already mounted
 
+	if err = ns.tryRestoreFuseMountInNodeStage(ctx, mnt, stagingTargetPath); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to try to restore FUSE mounts: %v", err)
+	}
+
 	isMnt, err := util.IsMountPoint(stagingTargetPath)
 	if err != nil {
 		log.ErrorLog(ctx, "stat failed: %v", err)
@@ -163,6 +167,25 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 
 	log.DebugLog(ctx, "cephfs: successfully mounted volume %s to %s", volID, stagingTargetPath)
+
+	if _, isFuse := mnt.(*mounter.FuseMounter); isFuse {
+		// FUSE mount recovery needs NodeStageMountinfo records.
+
+		if err = fsutil.WriteNodeStageMountinfo(volID, &fsutil.NodeStageMountinfo{
+			VolumeCapability: req.GetVolumeCapability(),
+			Secrets:          req.GetSecrets(),
+		}); err != nil {
+			log.ErrorLog(ctx, "cephfs: failed to write NodeStageMountinfo for volume %s: %v", volID, err)
+
+			// Try to clean node stage mount.
+			if unmountErr := mounter.UnmountVolume(ctx, stagingTargetPath); unmountErr != nil {
+				log.ErrorLog(ctx, "cephfs: failed to unmount %s in WriteNodeStageMountinfo clean up: %v",
+					stagingTargetPath, unmountErr)
+			}
+
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	}
 
 	return &csi.NodeStageVolumeResponse{}, nil
 }
@@ -237,11 +260,33 @@ func (ns *NodeServer) NodePublishVolume(
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if err := ns.tryRestoreFuseMountsInNodePublish(
+		ctx,
+		volID,
+		stagingTargetPath,
+		targetPath,
+		req.GetVolumeContext(),
+	); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to try to restore FUSE mounts: %v", err)
+	}
+
 	if req.GetReadonly() {
 		mountOptions = append(mountOptions, "ro")
 	}
 
 	mountOptions = csicommon.ConstructMountOptions(mountOptions, req.GetVolumeCapability())
+
+	// Ensure staging target path is a mountpoint.
+
+	if isMnt, err := util.IsMountPoint(stagingTargetPath); err != nil {
+		log.ErrorLog(ctx, "stat failed: %v", err)
+
+		return nil, status.Error(codes.Internal, err.Error())
+	} else if !isMnt {
+		return nil, status.Errorf(
+			codes.Internal, "staging path %s for volume %s is not a mountpoint", stagingTargetPath, volID,
+		)
+	}
 
 	// Check if the volume is already mounted
 
@@ -284,11 +329,14 @@ func (ns *NodeServer) NodeUnpublishVolume(
 	if err = util.ValidateNodeUnpublishVolumeRequest(req); err != nil {
 		return nil, err
 	}
+
 	// considering kubelet make sure node operations like unpublish/unstage...etc can not be called
 	// at same time, an explicit locking at time of nodeunpublish is not required.
 	targetPath := req.GetTargetPath()
 	isMnt, err := util.IsMountPoint(targetPath)
 	if err != nil {
+		log.ErrorLog(ctx, "stat failed: %v", err)
+
 		if os.IsNotExist(err) {
 			// targetPath has already been deleted
 			log.DebugLog(ctx, "targetPath: %s has already been deleted", targetPath)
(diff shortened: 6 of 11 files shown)


nixpanic:
The description looks good, but I did not look into the code in detail yet. Is it possible to add an e2e test for this? You have documented the steps to reproduce, so I guess it should be possible...

gman0:
@nixpanic sure. In this PR, or in a separate one after this PR is merged?

nixpanic:
> @nixpanic sure. In this PR, or in a separate one after this PR is merged?

Ideally a 2nd commit in this PR.

gman0:
Sorry I was busy with other tasks so I didn't have time to check the CI failures.

I'd like to bring your attention to two things in this PR, both revolving around the fact that it's trying to remount volumes in `NodePublishVolume`. The way things are currently laid out simply don't allow for easy, backward-compatible transition to "recoverable FUSE-backed mounts":

* `NodePublishVolume` would now need secrets passed in the request. Storage classes would need to specify `csi.storage.k8s.io/node-publish-secret-name` and `csi.storage.k8s.io/node-publish-secret-namespace` to reference the secret. This can be made opt-in, and we'd offer mount recovery only if this secret reference is present.
* [`NodeStageVolumeRequest.volume_capability`](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodestagevolume) may be different from [`NodePublishVolumeRequest.volume_capability`](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume). `NodeStageVolume` would have to remember volume capability (e.g. in a file, in an emptyDir volume), and use this in case of mount recovery. Then it would need to "forget" this information in `NodeUnstageVolume`. This too can be made opt-in, similar to the point above.

Both of these points now make it less attractive to have this feature because of the added complexity. Unfortunately, support for FUSE is so bad that I still think it's worth it. What do you think @nixpanic @humblec @Madhu-1?

Madhu-1:
@gman0 is going to work if the node is restarted right? this won't solve the `Transport endpoint not connected` broken mount when only the csi cephfsplugin daemonset pod is restarted, isn't it?

Madhu-1:
@gman0 for Node restart scenario I was thinking below steps (theoretical one not tested)

* For fuse:- During NodeStage Check mount already exists if yes check its corrupted. if yes do unmount and continue with the NodeStage process

* During NodePublish check the target path is mountpoint if yes check it's corrupted. if it's a corrupted return error let the cleanup happen at NodeStage? (technically this should not happen because the cleanup is already done in nodestage which is issued before nodepublish)

at least how am looking at solving the fuse problem is 

* Get required fixes in the fuse client in ceph fs
* Introduce a healer in cephcsi as we did for rbd-nbd to fix the userspace client issue.

gman0:
@Madhu-1 

For your first comment:

> is going to work if the node is restarted right?

Restarting the node repairs broken mounts regardless of this patch. Kubelet would invoke `NodeStageVolume` and `NodePublishVolume` when it comes back up. And there are no broken mounts after restarts (they are not persisted across reboots) so there's nothing to fix for this case...

> this won't solve the Transport endpoint not connected broken mount when only the csi cephfsplugin daemonset pod is restarted, isn't it?

That's not the point of this PR. I've mentioned that both here and in https://github.com/ceph/ceph-csi/issues/2616:

> Mount recovery provided by this PR recovers mounts only in newly created Pods.

and

> Note that existing Pods would remain in a bad state, only new ones would have their mounts working. Existing Pods would need to NodeUnpublishVolume->NodePublishVolume, which could be achieved only by deleting these Pods.

The point is that after csi-cephfsplugin container restarts, or one or more ceph-fuse processes crash, new Pods that get scheduled on the affected node can still work. With this fix, new Pods consuming the volumes will not get `ENOTCONN` or similar IO errors anymore (so no _Transport endpoint not connected_) caused by severed FUSE connections. Existing Pods on the node would need to be restarted (i.e. deleted and created again) however.

> During NodeStage Check mount already exists if yes check its corrupted. if yes do unmount and continue with the NodeStage process

For your second comment:

> During NodeStage Check mount already exists if yes check its corrupted. if yes do unmount and continue with the NodeStage process

Yes, this is already done in this PR.

> During NodePublish check the target path is mountpoint if yes check it's corrupted. if it's a corrupted return error let the cleanup happen at NodeStage?

I think you're implying this would be done as a part of this "volume healer" component? Because otherwise without it, this would mean the volume in question would need to achieve `NodeUnstageVolume->NodeStageVolume` in their lifetime cycle. That's not always straightforward.

> (technically this should not happen because the cleanup is already done in nodestage which is issued before nodepublish)

Seems like this is also pertains to the "volume healer".

> Get required fixes in the fuse client in ceph fs

Is there any ETA on this? This problem (and with FUSE in general) has been here from the beginning.

> Introduce a healer in cephcsi as we did for rbd-nbd to fix the userspace client issue.

I'm only now getting familiar with this. What I understood is that as a part of driver initialization, it checks for existing volume attachments on the node managed by NBD, it then gathers enough metadata from PVs to recreate `NodeStageVolumeRequest`s and calls `NodeStageVolume` for the concerned volumes. There are a couple of things I'm not sure about though. I don't think this is the right place to discuss it (if we do, please let's do it in a separate ticket), but I'll go over this just to move this PR forward -- to decide whether we want to aim more in the direction of rbd-nbd or not.

* It circumvents the standard lifecycle of a volume on a node. The CSI spec says `NodeStageVolume` is called before any `NodePublishVolume` for the volume. Granted, it doesn't say anything about calling it some time after again, and since all calls must be idempotent anyway, invoking it again must be safe. And the driver probably knows what it's doing. I get why it's been implemented like this -- the procedure does the thing that was needed, so you're just reusing it. Having said that, the CSI spec gives `NodeStageVolume` a specific meaning that the "volume healer" component doesn't necessarily adhere to IMO.
* It's run only once. What happens if just a single NBD process dies (due to a bug for example)? With current setup, the whole Node plugin would need to be manually restarted to trigger restoring the mounts.
* This approach is too opinionated in regards to what the application consuming the volumes is doing. Imagine a scenario where it keeps opened files/directories for a long periods of time, and any IO error it encounters just triggers a retry to perform that read/write again (meaning it will never attempt to refresh these file handles). The file handles have been invalidated by the remount, and the application would never recover from this. The only safe way to restore mounts is to recreate the Pods, and let kubelet trigger mounting of volumes. If you do anything less than this, you can't guarantee the application is actually healthy -- even when you have restored the mounts. We shouldn't assume anything about what the application is doing. It lost its storage, and needs to be restarted. The same thing would happen with the kernel client by the way, just the application "restart" would be triggered by kernel panic :P
* For block devices this approach probably works in some cases. However, with FUSE it's not as straightforward. Haven't we tried basically this exact thing in https://github.com/ceph/ceph-csi/pull/282 and it turned out to not work at all? Maybe we could make it sort of work with the use of `nsenter` to propagate the restored bind mounts to the respective containers, but my point above still holds.

With all this said, I think we need some other reconciliation mechanism. For cephfs-csi we can start by restoring mounts when we can (during `NodePublishVolume` if necessary) for now. After that, the next step should be to somehow trigger restarts of Pods consuming the volumes (ideally as a part of kubelet's volume health monitoring and reconciliation, but until then this would have to be some external component).

gman0:
I've made a couple of changes that I hope we can discuss.

* @pkalever in my previous comment https://github.com/ceph/ceph-csi/pull/2634#issuecomment-976457353 I mentioned that the volume healer in its current state won't cover all the cases we would need to make the aftermath of ceph-fuse failures a bit easier to clean. Apologies for not speaking up on this beforehand, but I'm getting familiar with its codebase only as of recently. In https://github.com/ceph/ceph-csi/issues/2616#issuecomment-963044853 you mentioned adding support for some reattach feature. Could you please describe it a bit more? I assume this would have to be some component running in kernel so that it's not affected by issues that userspace introduces? Also, would this include somehow persisting opened file handles across ceph-fuse crashes? These all seem like major changes to me that would take a lot of time develop and test (and eventually land in kernel?). For these reasons, and the ones I mentioned in the comment above, I'm not convinced we should be going the volume healer way in its current form.
* Added mount info (created during `NodeStageVolume` and deleted in `NodeUnstageVolume`). This is currently persisted in a host path volume. It's used to restore mounts in staging target path during `NodePublishVolume` calls.
* Now passing in publish volume secrets. This needs updating storage classes.

These latest changes are all WIP. Please do let me know if you have other ideas how to go about this.

gman0:
Hi all and happy new year! @pkalever @nixpanic @Madhu-1 @humblec could you please take a look at this, if what was discussed makes sense and we can continue with this PR or not?

gman0:
@Madhu-1 would you prefer if I split this PR into smaller chunks, starting with the "less controversial" ones?

Madhu-1:
@gman0 sorry busy with something else. will take a look at this PR soon. No need for splitting.

humblec:
@gman0 apologies for the delay here, can you please refresh this PR and lets get this moving without any further delay..

gman0:
Thanks @humblec for reminding. I'll refresh this in the coming days.

gman0:
@humblec @Madhu-1 @pkalever I've refreshed this PR, and split it into smaller commits for hopefully easier review. Please see commit descriptions too, where I tried to go into more details behind the reasoning for each change. I'll squash them and add tests once we agree on the general approach for this.

internal/cephfs/nodeserver.go
```diff
@@ -94,31 +123,25 @@ func (ns *NodeServer) NodeStageVolume(
...
-				return nil, status.Error(codes.Internal, err.Error())
-			}
-		}
+		return nil, status.Error(codes.Internal, err.Error())
 	}
-	defer volOptions.Destroy()
 
 	// Check if the volume is already mounted
 
+	if err = ns.tryRestoreFuseMountInNodeStage(ctx, m, stagingTargetPath); err != nil {
```
Madhu-1: @gman0 if the volume is already mounted on the nodes. kubelet will not issue the nodeStage again. as we are trying to restore here, when the Nodestage will be issued to correct this one? Only when the node is rebooted this will get triggered or even this can happen in other cases also?

internal/cephfs/nodeserver.go
```diff
@@ -94,31 +123,25 @@ func (ns *NodeServer) NodeStageVolume(
...
+	}
+	defer volOptions.Destroy()
 
-		// gets mon IPs from the supplied cluster info
-		volOptions, _, err = core.NewVolumeOptionsFromStaticVolume(string(volID), req.GetVolumeContext())
-		if err != nil {
-			if !errors.Is(err, cerrors.ErrNonStaticVolume) {
-				return nil, status.Error(codes.Internal, err.Error())
-			}
+	m, err := mounter.New(volOptions)
```
Madhu-1: instead of `m` use `mnt` for a variable name?

internal/cephfs/nodeserver.go
```diff
@@ -133,42 +156,57 @@ func (ns *NodeServer) NodeStageVolume(
...
+		req.GetStagingTargetPath(),
+		req.GetSecrets(),
+		req.GetVolumeCapability(),
+	); err != nil {
 		return nil, err
 	}
 
 	log.DebugLog(ctx, "cephfs: successfully mounted volume %s to %s", volID, stagingTargetPath)
 
+	if err = fsutil.WriteNodeStageMountinfo(volID, &fsutil.NodeStageMountinfo{
```
Madhu-1: I think cleanup is missing in the current code also but if the write fails we need to umount the volume to make sure no stale resources are left?

internal/cephfs/nodeserver.go
```diff
@@ -422,3 +509,258 @@ func (ns *NodeServer) NodeGetVolumeStats(
 
 	return nil, status.Errorf(codes.InvalidArgument, "targetpath %q is not a directory or device", targetPath)
 }
+
```
Madhu-1: move below changes to a new file fuse-recovery.go?

internal/cephfs/util/mountinfo.go
```diff
@@ -0,0 +1,140 @@
+/*
+Copyright 2021 The Ceph-CSI Authors.
```
Madhu-1: 2022

internal/cephfs/nodeserver.go
```diff
@@ -422,3 +509,258 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+	// Preconditions for mount recovery in NodePublishVolume calls:
+	// * volume secrets are present,
+	// * NodeStageMountinfo entry is present for this volume.
+
+	var (
+		nsMountinfo *fsutil.NodeStageMountinfo
+	)
+
+	if volSecrets == nil {
```
Madhu-1: check secrets first to fail fast?

Madhu-1 (commented):
@gman0 changes look okay to me. but am not clear on who will issue the NodeStage and Nodepublish to recover from the corruption?  is it possible to add an e2e for this one?

gman0:
Thank you @Madhu-1 for review! I'll address all your comments tomorrow.

internal/cephfs/nodeserver.go
```diff
@@ -73,11 +73,40 @@ func getCredentialsForVolume(
...
+	volSecrets map[string]string,
+) (*core.VolumeOptions, error) {
+	volOptions, _, err := core.NewVolumeOptionsFromVolID(ctx, string(volID), volContext, volSecrets)
+	if err != nil {
+		if !errors.Is(err, cerrors.ErrInvalidVolID) {
+			return nil, err
+		}
+
+		// gets mon IPs from the supplied cluster info
+		volOptions, _, err = core.NewVolumeOptionsFromStaticVolume(string(volID), volContext)
```
humblec: @gman0 dont we have `clusterID` here from the volContext and just use it with `util.GetMon` or existing function to retriev the monitors?

internal/cephfs/nodeserver.go
humblec: The general flow (iic ) goes like this: it  will try to fetch and restore if there is a record, otherwise skip and do the staging as in normal workflow.

internal/cephfs/nodeserver.go
```diff
@@ -346,14 +416,25 @@ func (ns *NodeServer) NodeUnstageVolume(
 
 	isMnt, err := util.IsMountPoint(stagingTargetPath)
 	if err != nil {
+		log.ErrorLog(ctx, "stat failed: %v", err)
```
humblec, line 406: please have the path where stat or mountpoint attempted too.

internal/cephfs/nodeserver.go
```diff
@@ -288,19 +348,29 @@ func (ns *NodeServer) NodeUnpublishVolume(
 	if err = util.ValidateNodeUnpublishVolumeRequest(req); err != nil {
 		return nil, err
 	}
+
 	// considering kubelet make sure node operations like unpublish/unstage...etc can not be called
 	// at same time, an explicit locking at time of nodeunpublish is not required.
 	targetPath := req.GetTargetPath()
 	isMnt, err := util.IsMountPoint(targetPath)
 	if err != nil {
+		log.ErrorLog(ctx, "stat failed: %v", err)
```
humblec, line 338: same here , record the path.

internal/cephfs/nodeserver.go
humblec: Isnt this the clean up https://github.com/ceph/ceph-csi/pull/2634/files#diff-091d7d0d011bb3b3c7ef3d64b375bb563052397a9a97ac7adfa4c94169367f6cR449 ?

humblec (commented):
@gman0 Pretty much everything in place , cool. Have few comments, ptal.. thanks :+1:

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
```diff
@@ -164,6 +166,10 @@ spec:
           emptyDir: {
             medium: "Memory"
           }
+        - name: ceph-csi-mountinfo
+          hostPath:
+            path: /var/lib/kubelet/plugins/cephfs.csi.ceph.com/mountinfo
```
humblec, line 171: can we get this path from the `driverName` var?

gman0:
@Madhu-1 

> who will issue the NodeStage and Nodepublish to recover from the corruption?

Ideally kubelet would -- using `NodeGetVolumeStats` triggered by `volume_condition`, but it doesn't offer this functionality at the moment. The alternative would be to have some external component watching and triggering recovery, however we don't have that either. So for the time being users are the ones who have to restart their Pods in order for their mounts to start working again.

I believe I mentioned this in some other comment above, but the main point of this PR isn't really to fix mounts in existing Pods on the node. For the lack of better term: they are screwed and must be deleted (and let k8s start them again).

The improvement done here is that whenever **new** Pods get scheduled on the node and they try to mount the same volume(s) that had their `staging_target_path` mountpoints broken, they can actually succeed.

Without this PR, the only way to currently recover from these broken mounts is to restart the node because that volume is rendered un-mountable. Users can't unmount such volumes either, see [this](https://github.com/ceph/ceph-csi/blob/0311eb5f447762a581194d649ea4d703da104fb8/internal/cephfs/nodeserver.go#L265-L276) and [this](https://github.com/ceph/ceph-csi/blob/0311eb5f447762a581194d649ea4d703da104fb8/internal/cephfs/nodeserver.go#L319-L329): the case when `util.IsMountPoint` returns an error due to broken mounts is not handled -- kubelet will just keep on issuing `Node{Unstage,Unpublish}Volume` calls but never succeeding. So without being able unmount volumes in existing Pods on the node, and not being able to successfully run new Pods there, the only option is to restart it.

We already discussed that the end goal would be to support reconnecting in the FUSE kernel module, but that is not trivial. This PR is just a first step in making userspace mounts a bit more reliable.

> is it possible to add an e2e for this one?

Of course! I will try to add it.

internal/cephfs/nodeserver.go
gman0: @Madhu-1

> kubelet will not issue the nodeStage again.

That's right, this is probably a dead code with kubelet at the moment. I've added it here just for the sake of "correctness": if NodeStage sees something is broken, it will try to fix it.

@humblec 

Workflow in `NodeStageVolume`:
* Is `staging_target_path` corrupted?
   * Yes: unmount `staging_target_path`.
   * No: fall through.
* Mount `staging_target_path`.

Workflow in `NodePublishVolume`:
* Is `staging_target_path` corrupted or unmounted, or `target_path` corrupted?
   * Yes: check for mountinfo record, and volume secrets. Try to remount `staging_target_path` and only unmount `target_path` (the bindmount will be restored in the next step)
   * No, or some of the prerequisites are not present: fall through.
* Is `staging_target_path` ok? If no, exit with error.
* Bindmount `staging_target_path` to `target_path`.

internal/cephfs/nodeserver.go
gman0: @Madhu-1 is correct, I'll take a look.

internal/cephfs/nodeserver.go
gman0: Sure, unfortunately it will probably need to stay in the same package because it's using some unexported functions? Refactoring that would be messy I think.

internal/cephfs/nodeserver.go
gman0: So currently the logs would be like this:

* `cephfs: mount problem detected when publishing a volume: %s is %s, %s is %s; attempting recovery`
* If no secrets:
   * `cephfs: cannot proceed with mount recovery because volume secrets are missing in the request`
* If no mountinfo record:
   * `cephfs: cannot proceed with mount recovery because NodeStageMountinfo record is missing`

IMHO this gives users a bit more idea about whats going on and why they will not be getting any mount recovery, instead of just exiting with an error "Transport endpoint is not connected" (this could be improved too of course).

Please let me know if you'd still prefer to fail fast (in which case I'll try to improve the err message).

internal/cephfs/nodeserver.go
gman0: GitHub's in-line comments are a bit weird... Are you referring to [`core.NewVolumeOptionsFromStaticVolume()`](https://github.com/ceph/ceph-csi/blob/0311eb5f447762a581194d649ea4d703da104fb8/internal/cephfs/core/volumeoptions.go#L438-L505)? The `// gets mon IPs from the supplied cluster info` comment is wrong, that function does a lot more than just retrieving mon IPs.

I just copy-pasted what was there before refactoring to minimize changes, but I can change the descriptions because they are indeed a bit misleading.

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
gman0: This is just plain YAML document, so I'm not sure that's possible. I did use `driverName` in the Helm chart though.

internal/cephfs/nodeserver.go
gman0: Actually, the error returned by `os.Stat()` (that's what `util.IsMountPoint` calls) already includes the path:

```
E1111 08:59:04.495982       1 nodeserver.go:360] ID: 4 Req-ID: 49670f1b-26de-4feb-887f-b74eb6ab4488 stat failed: stat /var/lib/kubelet/pods/2df62d3a-b329-4f4a-8483-e98a563463eb/volumes/kubernetes.io~csi/pvc-1e705b75-af6e-4827-ab45-90f2563539f2/mount: transport endpoint is not connected

```

internal/cephfs/nodeserver.go
Madhu-1: it should be fine. I was looking to optimize it so that we don't make unnecessary calls

Madhu-1:
>Ideally kubelet would -- using NodeGetVolumeStats triggered by volume_condition, but it doesn't offer this functionality at the moment. The alternative would be to have some external component watching and triggering recovery, however we don't have that either. So for the time being users are the ones who have to restart their Pods in order for their mounts to start working again.

https://github.com/csi-addons/kubernetes-csi-addons/pull/94 should help us to recover from this one until kubelet implements this one.

gman0:
I've tried to address the comments (and errors reported by "static-check") and added an e2e test. PTAL.

gman0:
The CI seems to be failing on [`createApp` in generic volume test](https://github.com/ceph/ceph-csi/blob/14f16ff95b4e3643af82236ec9ebb65a94ef0fb3/e2e/cephfs.go#L324-L340).

gman0:
A rather strange error:

```
2688   │ • Failure [600.333 seconds]
2689   │ cephfs
2690   │ /go/src/github.com/ceph/ceph-csi/e2e/cephfs.go:191
2691   │   Test CephFS CSI
2692   │   /go/src/github.com/ceph/ceph-csi/e2e/cephfs.go:279
2693   │     Test CephFS CSI [It]
2694   │     /go/src/github.com/ceph/ceph-csi/e2e/cephfs.go:280
2695   │ 
2696   │     Feb 16 12:51:34.061: killall command failed: err podlist is empty, stderr 
2697   │ 
2698   │     /go/src/github.com/ceph/ceph-csi/e2e/cephfs.go:575
```

https://github.com/ceph/ceph-csi/blob/e08007f30073c374923b40c0dddcb95218088a23/e2e/cephfs.go#L568-L576

I'm not sure how this could **sometimes** return empty list.

Other errors seem to be related to timeouts. @Madhu-1 @humblec do you think increasing time limit could help?

deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
humblec: actually it can be picked from ENV var even here.

Madhu-1:
> A rather strange error:
> 
> ```
> 2688   │ • Failure [600.333 seconds]
> 2689   │ cephfs
> 2690   │ /go/src/github.com/ceph/ceph-csi/e2e/cephfs.go:191
> 2691   │   Test CephFS CSI
> 2692   │   /go/src/github.com/ceph/ceph-csi/e2e/cephfs.go:279
> 2693   │     Test CephFS CSI [It]
> 2694   │     /go/src/github.com/ceph/ceph-csi/e2e/cephfs.go:280
> 2695   │ 
> 2696   │     Feb 16 12:51:34.061: killall command failed: err podlist is empty, stderr 
> 2697   │ 
> 2698   │     /go/src/github.com/ceph/ceph-csi/e2e/cephfs.go:575
> ```
> 
> https://github.com/ceph/ceph-csi/blob/e08007f30073c374923b40c0dddcb95218088a23/e2e/cephfs.go#L568-L576
> 
> I'm not sure how this could **sometimes** return empty list.
> 
> Other errors seem to be related to timeouts. @Madhu-1 @humblec do you think increasing time limit could help?

I think it's failing in helm tests because the labels are different in helm templates. get the deployment with the name and get labels from it instead of hardcoding it.

gman0:
@Madhu-1 thanks! I changed it as you suggested.

gman0:
FYI all of the latest CI failures exhibit the issue described in https://github.com/ceph/ceph-csi/issues/2889.

gman0:
The latest failures are now mostly related resize failures, where `df | grep | awk` that's run in the app container couldn't get a correct reading of volume size. In https://github.com/ceph/ceph-csi/pull/2634/commits/64ec5b5fac6330dd40f68b01cec57195a09d90d3 I used Alpine-based distro, and `df` behaves differently apparently: if devname is too long, it seems to put it on a separate line, which screws up processing it with `awk`. I'll switch to something glibc-based.

gman0:
error from ci/centos/mini-e2e-helm/k8s-1.23:

```• Failure [1754.575 seconds]

RBD

/go/src/github.com/ceph/ceph-csi/e2e/rbd.go:223

  Test RBD CSI

  /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:370

    Test RBD CSI [It]

    /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:371



    Feb 18 22:44:45.335: backend images not matching kubernetes resource count,image count 20 kubernetes resource count 21

    backend image Info:

     [csi-vol-3f0238c1-910c-11ec-85a1-da752a949301 csi-vol-49fe7d70-910c-11ec-85a1-da752a949301 csi-vol-49fe7d70-910c-11ec-85a1-da752a949301-temp csi-vol-4a2d481d-910c-11ec-85a1-da752a949301-temp csi-vol-4a4968cc-910c-11ec-85a1-da752a949301 csi-vol-4a4968cc-910c-11ec-85a1-da752a949301-temp csi-vol-4a688b50-910c-11ec-85a1-da752a949301 csi-vol-4a688b50-910c-11ec-85a1-da752a949301-temp csi-vol-4a844d45-910c-11ec-85a1-da752a949301 csi-vol-4a844d45-910c-11ec-85a1-da752a949301-temp csi-vol-4a97ee89-910c-11ec-85a1-da752a949301 csi-vol-4a97ee89-910c-11ec-85a1-da752a949301-temp csi-vol-5065a19c-910c-11ec-85a1-da752a949301 csi-vol-5065a19c-910c-11ec-85a1-da752a949301-temp csi-vol-50c86f31-910c-11ec-85a1-da752a949301 csi-vol-50c86f31-910c-11ec-85a1-da752a949301-temp csi-vol-517d9f94-910c-11ec-85a1-da752a949301 csi-vol-517d9f94-910c-11ec-85a1-da752a949301-temp csi-vol-51ace001-910c-11ec-85a1-da752a949301 csi-vol-51ace001-910c-11ec-85a1-da752a949301-temp]



    /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:214
```

gman0:
@Madhu-1 @humblec tests seem to be passing. Can you please take a look so I clean up the commits? I'll try to do without https://github.com/ceph/ceph-csi/pull/2634/commits/315b406841d3ebc171259c919f2e2ae90e89c6ec (https://github.com/ceph/ceph-csi/issues/2889#issuecomment-1046475237).

gman0:
Rebased and all tests but one (https://github.com/ceph/ceph-csi/issues/2327, [CI run](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.23/detail/mini-e2e-helm_k8s-1.23/755/pipeline)) passing. https://github.com/ceph/ceph-csi/pull/2634/commits/611d7d3ec6eb512114927794665c540e9fa4cafd seems to fix https://github.com/ceph/ceph-csi/issues/2889

humblec:
> @Madhu-1 @humblec tests seem to be passing. Can you please take a look so I clean up the commits? I'll try to do without [315b406](https://github.com/ceph/ceph-csi/commit/315b406841d3ebc171259c919f2e2ae90e89c6ec) ([#2889 (comment)](https://github.com/ceph/ceph-csi/issues/2889#issuecomment-1046475237)).

Last a couple of days, I was on SL, so the delay.. 
Will review this today, meanwhile you may correct the commit lint failure @gman0

gman0:
Ok thanks, I hope you're doing better now. I'll do that.

gman0:
I'll rebase and clean the commits once https://github.com/ceph/ceph-csi/pull/2902 is merged.

gman0:
I've cleaned up the commits and added some brief docs. I wonder if "recovery" shouldn't be called just "remount" -- because that's all this PR is doing. I wouldn't want to give users false expectations from it.

e2e/utils.go
```diff
@@ -191,19 +191,12 @@ func createPVCAndApp(
 	return err
 }
 
-// createPVCAndDeploymentApp creates pvc and deployment, if name is not empty
-// same will be set as pvc and app name.
+// createPVCAndDeploymentApp creates pvc and deployment.
```
gman0, line 194: Apologies for these not exactly relevant changes (in rbd.go as well), but they were asked for by some linter in `static-check`.

gman0:
ci/centos/mini-e2e-helm/k8s-1.22 failed with

```
[91m[1m• Failure [2344.624 seconds][0m
RBD
[90m/go/src/github.com/ceph/ceph-csi/e2e/rbd.go:223[0m
  Test RBD CSI
  [90m/go/src/github.com/ceph/ceph-csi/e2e/rbd.go:370[0m
    [91m[1mTest RBD CSI [It][0m
    [90m/go/src/github.com/ceph/ceph-csi/e2e/rbd.go:371[0m

    [91mFeb 24 21:17:53.923: backend images not matching kubernetes resource count,image count 22 kubernetes resource count 21
    backend image Info:
     [csi-vol-239cfc6a-95b7-11ec-a0c7-669db1b289d4 csi-vol-25b1d3a7-95b7-11ec-a0c7-669db1b289d4 csi-vol-25b1d3a7-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-25cfc03a-95b7-11ec-a0c7-669db1b289d4 csi-vol-25cfc03a-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-25ee62bd-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-26281dc3-95b7-11ec-a0c7-669db1b289d4 csi-vol-26281dc3-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-2652076a-95b7-11ec-a0c7-669db1b289d4 csi-vol-2652076a-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-2664eee4-95b7-11ec-a0c7-669db1b289d4 csi-vol-2664eee4-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-267e5602-95b7-11ec-a0c7-669db1b289d4 csi-vol-267e5602-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-26a492a3-95b7-11ec-a0c7-669db1b289d4 csi-vol-26a492a3-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-2a14ca7c-95b7-11ec-a0c7-669db1b289d4 csi-vol-2a14ca7c-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-2aa32621-95b7-11ec-a0c7-669db1b289d4 csi-vol-2aa32621-95b7-11ec-a0c7-669db1b289d4-temp csi-vol-2ac71f8d-95b7-11ec-a0c7-669db1b289d4 csi-vol-2ac71f8d-95b7-11ec-a0c7-669db1b289d4-temp][0m
```

CI run: https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.22/detail/mini-e2e-helm_k8s-1.22/2187/pipeline

gman0:
Some infra error?

ci/centos/mini-e2e-helm/k8s-1.22 https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.22/detail/mini-e2e-helm_k8s-1.22/2188/pipeline

```
cico-workspace-z0nz1 was marked offline: Node is being removed

Unable to create live FilePath for cico-workspace-z0nz1
```

gman0:
All tests passed, @Madhu-1 @humblec @nixpanic . Can you please take a look?

internal/cephfs/fuserecovery.go
```diff
@@ -0,0 +1,283 @@
...
+// 2. Recovery preconditions:
+//    * volume secrets are present,
+//    * NodeStageMountinfo is present for this volume,
+//    * if staging target path and target path are mountpoints, they must be
+//      managed by ceph-fuse,
+//    * VolumeOptions.Mounter must evaluate to "fuse".
+// 3. Recovery:
+//    * staging target path is unmounted and mounted again using ceph-fuse,
+//    * target path is only unmounted; NodePublishVolume is then expected to
+//      continue normally.
```
humblec, line 110: @gman0 if we `also record nodeStageSecret` information in this record path for a volume, cant we forget about the requirement of filling `nodePublishSecret` and consuming it? because the recovery in any of the above workflow `contact` or `get` a reference about staging information from this record. Please correct if I missed anything here.

humblec:
> All tests passed, @Madhu-1 @humblec @nixpanic . Can you please take a look?

@gman0 I did look at this PR on friday and it looks  good to go ! :+1: and also thanks for pulling this!.

I have one question about the possibility of avoiding `nodePublishSecret` , apart from that this looks good to me.

Also about "recovery Vs remount` :  both can serve, but may be for an end user `remount` convey the message easy.

ps ( or a general comment about the approach of solving this, as we have reference to NBD case ) # 

The `metadata record` and trying to recover from the record is a common model followed or suggested mechanism. Pretty much or  this is the same model which I have been keep proposing for NBD case too and still proposing against the `Healer`. Eventhough we can continue that discussions in NBD issue/PRs , I just thought of mentioning it here too.

e2e/utils.go
humblec: its fine :)

internal/cephfs/fuserecovery.go
gman0: @humblec I remember there were discussions about safety implications of storing secrets on the node in the past, and I'm not aware there have been any changes in this regard, but please do let me know if you know more. Not having to push the secrets into `NodePublishVolume` calls would be a lot nicer, I completely agree, but I'm not sure we have any safe way to do that.

Even though what ceph-csi is currently doing with the secrets file is not completely bulletproof either, storing the secrets in mountinfo would be a lot worse (it's a `hostPath` volume) I think.

gman0:
@humblec 

> The metadata record and trying to recover from the record is a common model followed or suggested mechanism. Pretty much or this is the same model which I have been keep proposing for NBD case too and still proposing against the Healer. Eventhough we can continue that discussions in NBD issue/PRs , I just thought of mentioning it here too.

Please see this comment https://github.com/ceph/ceph-csi/pull/2634#issuecomment-976457353 -- I've tried to summarize some initial thoughts there. But anyway we can discuss it elsewhere if everyone is ok with that.

gman0:
Can this PR get some more attention? It's been open since November of last year.

humblec:
> Can this PR get some more attention? It's been open since November of last year.

@gman0 as mentioned a couple of days back,  I am all good with this PR! approved too. I was waiting for one more revert from the team which is missing atm. Let me try to ping and get the second approval. A couple of resources are away this week, so the delay.

humblec (dismissed):

internal/cephfs/fuserecovery.go
humblec: The main thought process behind getting rid of nodepublish was that, its a completely new input for the driver at this stage, so I was thinking some way to get rid of it and may be stick to what is known till now for the driver/end user. Actually storing the secret information in the node may not be that bad as even today if someone has access to the host volumes, there is security issue.

humblec:
Cc @ceph/ceph-csi-contributors can you spend some time on the review of this PR considering its been hanging there for some time now?

gman0:
Thanks @humblec! The code freeze for v3.6 is the 20th of March as it seems, so I just want to make sure we have some time reserve should more changes be needed, if this PR is to be merged in time for the release.

internal/cephfs/fuserecovery.go
gman0: Could someone with root access to the node (or at least to the node plugin containers) be able to somehow mine the password out of ceph-fuse process memory? What I'm getting at is if it's reasonably easy to retrieve the cephx key using this channel, I'd agree with you that it might not make that much difference if we stored it in a file.

> its a completely new input for the driver at this stage,

Yes, it is very ugly, and I too wish we could get rid of it.

gman0: Is there someone from Ceph we could ping to verify this claim?

gman0: I tried asking on ceph-users mailing list:

https://lists.ceph.io/hyperkitty/list/ceph-users@ceph.io/thread/TMLSACFXFBDTUCSVRHQGDORRCJWE24AQ/

humblec:
> Thanks @humblec! The code freeze for v3.6 is the 20th of March as it seems, so I just want to make sure we have some time reserve should more changes be needed, if this PR is to be merged in time for the release.

Sure, and thanks for raising the concern! We will get this in this week :+1:

internal/cephfs/fuserecovery.go
humblec: Thanks,! at present we have one reply which talks about core dumping  as you noted other than the root access.

gman0: @humblec I've opened https://github.com/ceph/ceph-csi/issues/2924 and I think I can move forward on this. If anyone thinks otherwise please do let me know.

humblec: Sure, as its possible to retrieve the cephx keys pretty easily ( ref #issue 2924) , I think we could think about storing the secret info even in the record file and get rid of using one other secret in SC which also enable us to have no difference in SC param list compared to previous version and avoid recreating SCs..etc.

gman0: Done, mount info now stores secrets from `NodeStageVolumeRequest` in a file (0600, owned by root).

gman0:
ci/centos/mini-e2e-helm/k8s-1.21:

again this error:

```

• Failure [704.771 seconds]

RBD

/go/src/github.com/ceph/ceph-csi/e2e/rbd.go:223

  Test RBD CSI

  /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:370

    Test RBD CSI [It]

    /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:371



    Mar  8 17:44:50.080: backend images not matching kubernetes resource count,image count 22 kubernetes resource count 21

    backend image Info:

     [csi-vol-4a31f8d6-9f07-11ec-93b0-8ae793334235 csi-vol-5bb7ce8c-9f07-11ec-93b0-8ae793334235 csi-vol-5bb7ce8c-9f07-11ec-93b0-8ae793334235-temp csi-vol-5bd64941-9f07-11ec-93b0-8ae793334235 csi-vol-5bd64941-9f07-11ec-93b0-8ae793334235-temp csi-vol-5c128411-9f07-11ec-93b0-8ae793334235-temp csi-vol-5c313bba-9f07-11ec-93b0-8ae793334235 csi-vol-5c313bba-9f07-11ec-93b0-8ae793334235-temp csi-vol-5d95cfcf-9f07-11ec-93b0-8ae793334235 csi-vol-5d95cfcf-9f07-11ec-93b0-8ae793334235-temp csi-vol-5f067573-9f07-11ec-93b0-8ae793334235 csi-vol-5f067573-9f07-11ec-93b0-8ae793334235-temp csi-vol-5f3b72fc-9f07-11ec-93b0-8ae793334235 csi-vol-5f3b72fc-9f07-11ec-93b0-8ae793334235-temp csi-vol-5f585149-9f07-11ec-93b0-8ae793334235 csi-vol-5f585149-9f07-11ec-93b0-8ae793334235-temp csi-vol-5f63c1b2-9f07-11ec-93b0-8ae793334235 csi-vol-5f63c1b2-9f07-11ec-93b0-8ae793334235-temp csi-vol-635883e4-9f07-11ec-93b0-8ae793334235 csi-vol-635883e4-9f07-11ec-93b0-8ae793334235-temp csi-vol-63b66685-9f07-11ec-93b0-8ae793334235 csi-vol-63b66685-9f07-11ec-93b0-8ae793334235-temp]



    /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:214
```

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.21/detail/mini-e2e-helm_k8s-1.21/2944/pipeline

gman0:
ci/centos/mini-e2e/k8s-1.22:

Both cephfs and rbd tests failed.

* cephfs: failed on "verifying that ceph-fuse recovery works for new pods" test -- the second replica of the deployment never became ready (`Error: context deadline exceeded`), but I can see `NodePublishVolume` for both pods succeed quickly, so the problem must have occurred somewhere else...
* rbd:

```
Mar  8 17:41:31.171: INFO: waiting for kubectl (create -f args []) to finish
Mar  8 17:41:31.171: INFO: Running '/usr/bin/kubectl --server=https://192.168.39.163:8443 --kubeconfig=/root/.kube/config --namespace=cephcsi-e2e-6ae669b3 create -f -'
Mar  8 17:41:31.398: INFO: rc: 1
Mar  8 17:41:31.398: INFO: will run kubectl (create) args ([]) again (0 seconds elapsed)
Mar  8 17:41:31.398: FAIL: failed to create ceph-conf configmap object: failed to run kubectl: error running /usr/bin/kubectl --server=https://192.168.39.163:8443 --kubeconfig=/root/.kube/config --namespace=cephcsi-e2e-6ae669b3 create -f -:
Command stdout:

stderr:
Error from server (Forbidden): error when creating "STDIN": configmaps "ceph-config" is forbidden: unable to create new content in namespace cephcsi-e2e-6ae669b3 because it is being terminated

error:
exit status 1
````

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.22/detail/mini-e2e_k8s-1.22/2184/pipeline/

File: docs/ceph-fuse-corruption.md [hunk 13]
```diff
@@ -0,0 +1,45 @@
+# ceph-fuse: detection of corrupted mounts and their recovery
```
Madhu-1, line 1: @gman0 am trying to understand why this doc is needed, is it going to help users?

e2e/cephfs.go
```diff
@@ -504,6 +505,135 @@ var _ = Describe("cephfs", func() {
...
+						return fmt.Errorf(
+							"expected stat to succeed without error output ; got err %w, stderr %s",
+							statErr, stdErr,
+						)
+					}
+
+					return nil
+				}
+
+				podName1, podName2 := deplPods[0].Name, deplPods[1].Name
```
Madhu-1: use pod1Name and pod2Name here for better readability?

e2e/cephfs.go
```diff
@@ -504,6 +505,135 @@ var _ = Describe("cephfs", func() {
...
+					e2elog.Failf(err.Error())
+				}
+				// List Deployment's pods again to get name of the new pod.
+				deplPods, err = listPods(f, depl.Namespace, &metav1.ListOptions{
+					LabelSelector: fmt.Sprintf("app=%s", depl.Labels["app"]),
+				})
+				if err != nil {
+					e2elog.Failf("failed to list pods for Deployment: %v", err)
+				}
+				var podName3 string
```
Madhu-1: no need for new variable reuse podName2 here?

internal/cephfs/util/mountinfo.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+// nodeStageMountinfoRecord describes a single
+// record of mountinfo of a staged volume.
+// encoding/json-friendly format.
+// Only for internal use for marshaling and unmarshaling.
+type nodeStageMountinfoRecord struct {
+	VolumeCapabilityProtoJSON string            `json:",omitempty"`
+	MountOptions              []string          `json:",omitempty"`
+	Secrets                   map[string]string `json:",omitempty"`
```
Madhu-1, line 46: why store the secrets on the nodePath what is the problem with getting it from the Nodepublish secret?

docs/ceph-fuse-corruption.md
gman0: It was largely to inform users about the need of node publish secrets, but after discussion with @humblec, and some concrete examples in https://github.com/ceph/ceph-csi/issues/2924 with might not need them anymore, and so this document is probably also not necessary.

internal/cephfs/util/mountinfo.go
gman0: @Madhu-1 Please see the comments in https://github.com/ceph/ceph-csi/pull/2634#pullrequestreview-894655827. To sum this up:
* Users would need to recreate their storage classes after upgrade to take advantage of this feature.
* The same is true for existing PVs that would need to be recreated. This would mean taking the application using these PVs down.
* Storing the keyrings on the node doesn't seem to be an issue. Should an adversary user get root access to the node, it is trivial to acquire the keys anyway. See https://github.com/ceph/ceph-csi/issues/2924 for more details. Instead of having this "false feeling of safety", we could focus on limiting the "blast radius" a cephx user of a volume has. But this is out of scope for this PR.

docs/ceph-fuse-corruption.md
gman0: Also, it describes what the recovery is doing and limits what users can expect from it. I think it still serves this purpose. Please let me know if you'd still like me to remove it.

gman0:
ci/centos/mini-e2e-helm/k8s-1.22

```
24963   │ • Failure [2188.486 seconds]
24964   │ RBD
24965   │ /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:223
24966   │   Test RBD CSI
24967   │   /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:370
24968   │     Test RBD CSI [It]
24969   │     /go/src/github.com/ceph/ceph-csi/e2e/rbd.go:371
24970   │ 
24971   │     Mar  9 12:24:41.985: backend images not matching kubernetes resource count,image count 22 kubernetes resource count 21
24972   │     backend image Info:
24973   │      [csi-vol-cb21f231-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-cd5f9a22-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-cd5f9a22-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-cd7e2653-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-cd9d188a-9fa3-11ec-bf54-e69348dbc8c3 csi-
        │ vol-cd9d188a-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-cdd964bf-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-cdd964bf-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-ce3005a7-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-ce3005a7-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-c
        │ e50af5f-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-ce50af5f-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-ce723aa9-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-ce723aa9-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-cea8f414-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-cea8f414-9f
        │ a3-11ec-bf54-e69348dbc8c3-temp csi-vol-d632f265-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-d632f265-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-d6a68d79-9fa3-11ec-bf54-e69348dbc8c3 csi-vol-d6a68d79-9fa3-11ec-bf54-e69348dbc8c3-temp csi-vol-d7092d80-9fa3-11
        │ ec-bf54-e69348dbc8c3 csi-vol-d7092d80-9fa3-11ec-bf54-e69348dbc8c3-temp]
```

https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.22/detail/mini-e2e-helm_k8s-1.22/2231/pipeline

/retest ci/centos/mini-e2e-helm/k8s-1.22

internal/cephfs/util/mountinfo.go
Madhu-1: as the fuse plugin is not production-ready and it's a new storageclass for the fuse mounter type why worry about adding new publish secrets to the storageclass.

IMO we don't need to worry about supporting the existing PVs.

docs/ceph-fuse-corruption.md
Madhu-1: as we can handle this thing internally not sure this is needed and we will [automate the recovery](https://github.com/csi-addons/kubernetes-csi-addons/pull/94) also in the future this will go away, isn't it? am looking will this be staying for a long time or will go away after some time. am fine if all agree to keep this doc.

internal/cephfs/util/mountinfo.go
gman0: Passing the secret to `NodePublishVolume` is no different to having it stored on the node security-wise. It's only a problem if a user has access to root account of the node, but that would be a problem with or without passing in extra secrets, so it doesn't really make any difference. Things like credentials revocation (so that the mountinfo file would possibly contain stale credentials) is also of no concern, because you'd still have to fully unmount the volume from the node, so that `NodeUnstageVolume` is triggered -- which in turn removes the mountinfo file.

For those deployments that do use FUSE (despite it's official ceph-csi status), adding new secrets only causes trouble for existing deployments.

By the way, mountinfo is written only when FUSE mounter is used:

https://github.com/ceph/ceph-csi/blob/1023d9720d01e95f0c239c8df812fa0d04e689e9/internal/cephfs/nodeserver.go#L171-L188

Maybe I'm not thinking of other cases though... Can you explain your concerns?

Madhu-1: am not sure what else major problem this can lead to and one this I can think of now itself is that this will also block us from updating the secret details username/keyring value. am trying to avoid a bigger problem when we GA this one.

One use case to save the secrets is to support already created PVC. we can document users to delete and recreate the PV object with publish reference also. IMO no one is using cephfs fuse in production yet.

gman0: > and one this I can think of now itself is that this will also block us from updating the secret details username/keyring value.

I've touched on this in my last comment.

>  Things like credentials revocation (so that the mountinfo file would possibly contain stale credentials) is also of no concern, because you'd still have to fully unmount the volume from the node, so that NodeUnstageVolume is triggered

If you change client name or the key, you need to remount. Remounting will refresh the mountinfo file.

> IMO no one is using cephfs fuse in production yet.

We are, in many clusters.

docs/ceph-fuse-corruption.md
gman0: @Madhu-1 

> as we can handle this thing internally not sure this is needed and we will automate the recovery

That might be true, but I think it's also important to let users know what exactly they can expect from the recovery, that it's not exactly a silver bullet, and has its limitations.

> also in the future this will go away, isn't it?

Well that depends on whether the FUSE kernel module, libfuse, and FUSE clients will be able to support reconnecting.

internal/cephfs/util/mountinfo.go
gman0: > am trying to avoid a bigger problem when we GA this one.

Actually adding new secrets is an API-facing breaking change, whose necessity might change in the future. Right now we can have this feature without needing to change the API (the secrets requirements in this case). Should the needs change, we can always add them later... it's always much harder to remove things though.

Madhu-1 (approved):
Am still not okay with storing the creds on the hostPath, as it's already discussed and @humblec already agreed for this one. I hope all the possibilities are considered like what problem it can cause etc.

humblec (approved):

luanshuo:
@gman0  Hi, there have some trouble,can you help me?

in our cluster, mount targetpath have error,after delete ceph-csi daemonset pod
image

we have mountinfo 
image

why no trying to unmount and mount ?

ceph-csi: 3.6.2
ceph: 15.2.13
k8s: 1.21.8

Madhu-1:
@luanshuo above logs says its NodeGetVolumeStats, please open a new issue with complete logs if you are looking for a help.

PR #2638, opened 2021-11-15
ci: avoid preleased bits

pkalever:

Skip testing with prereleased content, for example
v1.23.0-alpha.4
v1.23.0-alpha.3
v1.23.0-alpha.2
v1.23.0-alpha.1
v1.22.0-rc.0
v1.22.0-beta.2
v1.22.0-beta.1
v1.22.0-beta.0
v1.22.0-alpha.3
v1.22.0-alpha.2

---

<details>

  progress or results

</details>

### DIFF
--- a/scripts/get_patch_release.py
+++ b/scripts/get_patch_release.py
@@ -77,7 +77,9 @@ def get_releases(gh_releases):
     '''
     releases = list()
     for release in gh_releases:
-        releases.append(release['name'].split()[-1])
+        # ignore prereleases
+        if not release['prerelease']:
+            releases.append(release['name'].split()[-1])
     return releases
 
 


nixpanic:
What is the reason to skip testing with pre-releases? We usually do want to test with pre-releases for the next Kubernetes version. The script should pick the most recent release by default, and not some older release (pre-release, or minor release for that matter).

pkalever:
> What is the reason to skip testing with pre-releases? We usually do want to test with pre-releases for the next Kubernetes version. The script should pick the most recent release by default, and not some older release (pre-release, or minor release for that matter).

Picking prerelease bits means we are attempting to run ceph-csi on non-stable k8s branches. Which is testing a PR on a non-stable k8s branch.

I think running new PRs on non-stable k8s versions doesn't make sense to me. We should rather consider running nightly or weekly jobs with ceph-csi `devel` (merged ones) to catch any anomalies or changes in k8s.

nixpanic:
> The script should pick the most recent release by default, and not some older release (pre-release, or minor release for that matter).

:point_up: is important. The script prints out the version that it detected and configured for testing. If it detects a pre-release version where a stable version is available, we need to check why that happens.

For upcoming kubernetes versions, there are only alpha/beta/rc releases, so we do need to be able to test with those.

Madhu-1:
@pkalever we need to test with the most recent release(pre/patch/rc etc)  to catch the bugs earlier. say for example if we have 1.22.0 release but if we still testing with 1.22.0 pre or RC release then it's a problem? this PR is addressing this issue?

pkalever:
> @pkalever we need to test with the most recent release(pre/patch/rc etc) to catch the bugs earlier. say for example if we have 1.22.0 release but if we still testing with 1.22.0 pre or RC release then it's a problem? this PR is addressing this issue?

With these changes, if a k8s release is marked as prerelease like beta, alpha, and RC it will not be picked. My intention was to avoid testing ceph-csi new PRs with unstable k8s releases so that any anomalies in k8s might make us think it could be an issue with the new PR submitted at ceph-csi.

IMHO the k8s vs ceph-csi compatibility issues need to be checked with nightly or weekly jobs but not with regular PR-based checks.

nixpanic:
We only enable automatic testing of PRs with Kubernetes versions that are released. The CI jobs can be triggered manually with Kubernetes versions that are in alpha/beta/rc.

Automated CI runs, and required CI job passing should always use stable versions of Kubernetes. If you have seen occasions where this is not the case, we need to investigate why/how that happened. In that case, please link to jobs where non-stable releases were used.

pkalever:
Closing this based on https://github.com/ceph/ceph-csi/pull/2638#issuecomment-971347649

Will keep track for any non-stable k8s installations in the test beds.

Thanks @nixpanic & @Madhu-1

PR #2647, opened 2021-11-17
rbd: add a workaround to fix rbd snapshot scheduling

Madhu-1:
currently, we have a bug in the rbd mirror scheduling module. After doing failover and failback the scheduling is not getting updated and the mirroring snapshots are not getting created periodically as per the scheduling interval. This PR workaround this one by doing the below operations
    
* Create a dummy (csi-vol-dummy-ceph fsID) image per cluster and this image should be easily identified.
    
* During Promote operation on any image enables the mirroring on the dummy image. when we enable the mirroring on the dummy image the pool will get updated and the scheduling will be reconfigured.
    
* During Demote operation on any image disables the mirroring on the dummy image. the disabled need to be done to enable the mirroring again when we get the promote request to make the image as the primary
    
* When the DR is no more needed, this image needs to be  manually cleanup for now as we don't want to add a check
in the existing DeleteVolume code path for deleting dummy images as it impacts the performance of the DeleteVolume workflow.
 

Moved to add scheduling to the promote operation as scheduling need to be added when the image is promoted and this is the correct method of adding the scheduling to make the scheduling take place.

More details at  https://bugzilla.redhat.com/show_bug.cgi?id=2019161

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -117,12 +117,10 @@ func getMirroringMode(ctx context.Context, parameters map[string]string) (librbd
 	return mirroringMode, nil
 }
 
-// getSchedulingDetails gets the mirroring mode and scheduling details from the
+// validateSchedulingDetails gets the mirroring mode and scheduling details from the
 // input GRPC request parameters and validates the scheduling is only supported
 // for snapshot mirroring mode.
-func getSchedulingDetails(parameters map[string]string) (admin.Interval, admin.StartTime, error) {
-	admInt := admin.NoInterval
-	adminStartTime := admin.NoStartTime
+func validateSchedulingDetails(parameters map[string]string) error {
 	var err error
 
 	val := parameters[imageMirroringKey]
@@ -134,43 +132,49 @@ func getSchedulingDetails(parameters map[string]string) (admin.Interval, admin.S
 	// an optional parameter.
 	case "":
 	default:
-		return admInt, adminStartTime, status.Error(codes.InvalidArgument, "scheduling is only supported for snapshot mode")
+		return status.Error(codes.InvalidArgument, "scheduling is only supported for snapshot mode")
 	}
 
 	// validate mandatory interval field
 	interval, ok := parameters[schedulingIntervalKey]
 	if ok && interval == "" {
-		return admInt, adminStartTime, status.Error(codes.InvalidArgument, "scheduling interval cannot be empty")
+		return status.Error(codes.InvalidArgument, "scheduling interval cannot be empty")
 	}
-	adminStartTime = admin.StartTime(parameters[schedulingStartTimeKey])
+	adminStartTime := admin.StartTime(parameters[schedulingStartTimeKey])
 	if !ok {
 		// startTime is alone not supported it has to be present with interval
 		if adminStartTime != "" {
-			return admInt, admin.NoStartTime, status.Errorf(codes.InvalidArgument,
+			return status.Errorf(codes.InvalidArgument,
 				"%q parameter is supported only with %q",
 				schedulingStartTimeKey,
 				schedulingIntervalKey)
 		}
 	}
 	if interval != "" {
-		admInt, err = validateSchedulingInterval(interval)
+		err = validateSchedulingInterval(interval)
 		if err != nil {
-			return admInt, admin.NoStartTime, status.Error(codes.InvalidArgument, err.Error())
+			return status.Error(codes.InvalidArgument, err.Error())
 		}
 	}
 
-	return admInt, adminStartTime, nil
+	return nil
+}
+
+// getSchedulingDetails returns scheduling interval and scheduling startTime.
+func getSchedulingDetails(parameters map[string]string) (admin.Interval, admin.StartTime) {
+	return admin.Interval(parameters[schedulingIntervalKey]),
+		admin.StartTime(parameters[schedulingStartTimeKey])
 }
 
 // validateSchedulingInterval return the interval as it is if its ending with
 // `m|h|d` or else it will return error.
-func validateSchedulingInterval(interval string) (admin.Interval, error) {
+func validateSchedulingInterval(interval string) error {
 	re := regexp.MustCompile(`^\d+[mhd]$`)
 	if re.MatchString(interval) {
-		return admin.Interval(interval), nil
+		return nil
 	}
 
-	return "", errors.New("interval specified without d, h, m suffix")
+	return errors.New("interval specified without d, h, m suffix")
 }
 
 // EnableVolumeReplication extracts the RBD volume information from the
@@ -189,7 +193,7 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 	}
 	defer cr.DeleteCredentials()
 
-	interval, startTime, err := getSchedulingDetails(req.GetParameters())
+	err = validateSchedulingDetails(req.GetParameters())
 	if err != nil {
 		return nil, err
 	}
@@ -235,19 +239,6 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 
 			return nil, status.Error(codes.Internal, err.Error())
 		}
-	}
-
-	if interval != "" {
-		err = rbdVol.addSnapshotScheduling(interval, startTime)
-		if err != nil {
-			return nil, err
-		}
-		log.DebugLog(
-			ctx,
-			"Added scheduling at interval %s, start time %s for volume %s",
-			interval,
-			startTime,
-			rbdVol)
 	}
 
 	return &replication.EnableVolumeReplicationResponse{}, nil
@@ -435,6 +426,20 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 
 			return nil, status.Error(codes.Internal, err.Error())
 		}
+	}
+
+	interval, startTime := getSchedulingDetails(req.GetParameters())
+	if interval != admin.NoInterval {
+		err = rbdVol.addSnapshotScheduling(interval, startTime)
+		if err != nil {
+			return nil, err
+		}
+		log.DebugLog(
+			ctx,
+			"Added scheduling at interval %s, start time %s for volume %s",
+			interval,
+			startTime,
+			rbdVol)
 	}
 
 	return &replication.PromoteVolumeResponse{}, nil
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -23,6 +23,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ceph/ceph-csi/internal/util"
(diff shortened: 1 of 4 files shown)


Madhu-1:
Manual testing works for me. For now added DNM as its getting tested by @ShyamsundarR and @BenamarMk

internal/rbd/replicationcontrollerserver.go
```diff
@@ -420,6 +511,15 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 
 	// promote secondary to primary
 	if !mirroringInfo.Primary {
+		var mode librbd.ImageMirrorMode
+		mode, err = getMirroringMode(ctx, req.GetParameters())
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to get mirroring node %s", err.Error())
+		}
+		err = enableMirroringOnDummyImage(rbdVol, mode)
```
BenamarMk: rbdVol is not the dummy volume.  We will be enabling mirroring on a none dummy volume if am reading this correctly.

BenamarMk: Also, enabling mirroring for the dummy volume should be done after we promote the real volume.

Madhu-1: `enableMirroringOnDummyImage` takes the rbdVol as input and creates a new dummy struct and replaces the name and calls the mirror operation on that dummy image.

Madhu-1: @BenamarMk is it the same for disabling mirroring on dummy image also? it should be done after demoting the real image?

Madhu-1: fixed for enabling. let me know your input for disable also.

BenamarMk: Disable is fine.  YOu don't need to do anything to it.  So basically the flow is like this:
1. Promote
2. enable dummy
And then
1. Disable dummy
2. Demote

internal/rbd/replicationcontrollerserver.go
```diff
@@ -437,6 +529,30 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 	}
 
+	var mode librbd.ImageMirrorMode
```
BenamarMk, line 540: Shouldn't the whole block from 532 all the way to 540 be inside the block above? IOW, don't you need to put it at line 529?

Madhu-1: The reason for keeping it out of the loop is if the csidriver is restarted after promote and before enabling the mirroring on the dummy image on the next request the if check will be skipped as the image is already primary

internal/rbd/replicationcontrollerserver.go
Madhu-1: okay, then the current changes are fine. Thanks for confirming @BenamarMk

BenamarMk (approved):
It looks good to me @Madhu-1. Thanks for coding it so quickly.  I will now test it.

Madhu-1:

`Nov 17 15:36:49.574: FAIL: failed to create user cephcsi-cephfs-node with error etcdserver: request timed out`

internal/rbd/replicationcontrollerserver.go
```diff
@@ -495,6 +611,10 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 
 	// demote image to secondary
 	if mirroringInfo.Primary {
+		err = disableMirroringOnDummyImage(rbdVol)
```
ShyamsundarR, line 622: We need to ensure dummy image is primary (even if split-brain) for the enable/disable schedule to not error out. IOW, the dummy image can be in a split-brain scenario forever, but needs to be primary for these operations.

ShyamsundarR: Ignore @BenamarMk pointed out that we create different dummy images per cluster, so that would always be primary for the local cluster allowing enable/disable calls.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -437,6 +529,30 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 	}
 
+	var mode librbd.ImageMirrorMode
+	mode, err = getMirroringMode(ctx, req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get mirroring node %s", err.Error())
```
humblec: node/mode

internal/rbd/replicationcontrollerserver.go
```diff
@@ -237,20 +260,89 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
...
 
-	if interval != "" {
-		err = rbdVol.addSnapshotScheduling(interval, startTime)
+	return &replication.EnableVolumeReplicationResponse{}, nil
+}
+
+// getdummyImageName returns the csi-vol-dummy+cluster FSID as the image name.
+// each cluster should have a unique dummy image created. choosing the cluster
+// FSID for the same reason.
+func getdummyImageName(conn *util.ClusterConnection) (string, error) {
```
humblec: getDummyImageName

humblec:
One general question here @Madhu-1 , whlie the failback happen and if the dummy image exist, do we foresee any issues?

Madhu-1:
> One general question here @Madhu-1 , whlie the failback happen and if the dummy image exist, do we foresee any issues?

Do you mean the dummy image created by the user? it's a workaround I don't think the user will create any real images with the same name for using it.

humblec:
> > One general question here @Madhu-1 , whlie the failback happen and if the dummy image exist, do we foresee any issues?
> 
> Do you mean the dummy image created by the user? it's a workaround I don't think the user will create any real images with the same name for using it.

not that, a dummy image exist in the cluster and during failover we create another one for second cluster, Now during failback, the first cluster's dummy image exist already or not cleaned up, Thinking that,  I was asking the existence of dummy image in the first cluster can be a problem while the failback is established and images started to get promoted/demoted.

Madhu-1:
> > > One general question here @Madhu-1 , whlie the failback happen and if the dummy image exist, do we foresee any issues?
> > 
> > 
> > Do you mean the dummy image created by the user? it's a workaround I don't think the user will create any real images with the same name for using it.
> 
> not that, a dummy image exist in the cluster and during failover we create another one for second cluster,

The dummy image is unique per cluster it's created with the fsid as the prefix.

 >Now during fallback,  the first cluster's dummy image exist already or not cleaned up, Thinking that, I was asking the existence of dummy image in the first cluster can be a problem while the failback is established and images started to get promoted/demoted.

i don't see any problem here. during fallback the dummy image mirroring will be disabled we are not going to cleanup or do anything. we operator on the dummy images which is created for that cluster not the dummy image of the other cluster.

humblec:
> > > > One general question here @Madhu-1 , whlie the failback happen and if the dummy image exist, do we foresee any issues?
> > > 
> > > 
> > > Do you mean the dummy image created by the user? it's a workaround I don't think the user will create any real images with the same name for using it.
> > 
> > 
> > not that, a dummy image exist in the cluster and during failover we create another one for second cluster,
> 
> The dummy image is unique per cluster it's created with the fsid as the prefix.

Yeah, thats correct, what I meant above was same, may be `during failover..` was unwanted.. :) 

> 
> > Now during fallback,  the first cluster's dummy image exist already or not cleaned up, Thinking that, I was asking the existence of dummy image in the first cluster can be a problem while the failback is established and images started to get promoted/demoted.
> 
> i don't see any problem here. during fallback the dummy image mirroring will be disabled we are not going to cleanup or do anything. we operator on the dummy images which is created for that cluster not the dummy image of the other cluster.

Sure, if we dont expect any problem in this scenario, we are good.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -237,20 +260,89 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
...
-		log.DebugLog(
-			ctx,
-			"Added scheduling at interval %s, start time %s for volume %s",
-			interval,
-			startTime,
-			rbdVol)
+		dummyVol := rbdVol
+		dummyVol.RbdImageName = imgName
+		err = createImage(ctx, dummyVol, dummyVol.conn.Creds)
+		if err != nil && !strings.Contains(err.Error(), "File exists") {
```
humblec, line 297: @Madhu-1 one quick question here, if the image exist, do we need to return error? rather cant we treat it as success ?

Madhu-1: Yes its already exists we need to treat it as a success. Example csi deployment restarts etc.

humblec:
lgtm.. considering the importance of this PR we have to take this in at the earliest and can do further enhancements later.. @Madhu-1 will we wait for testing result from @BenamarMk @ShyamsundarR before merge ?  

CC @ceph/ceph-csi-contributors ptal

Madhu-1:
> lgtm.. considering the importance of this PR we have to take this in at the earliest and can do further enhancements later.. @Madhu-1 will we wait for testing result from @BenamarMk @ShyamsundarR before merge ?
> 

Yes waiting for the testing result.  but want to close up on review as changes look good to @BenamarMk  and tested locally.

> CC @ceph/ceph-csi-contributors ptal

internal/rbd/replicationcontrollerserver.go
```diff
@@ -437,6 +537,30 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 	}
 
+	var mode librbd.ImageMirrorMode
+	mode, err = getMirroringMode(ctx, req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get mirroring mode %s", err.Error())
+	}
+	err = enableMirroringOnDummyImage(rbdVol, mode)
```
ShyamsundarR, line 545: For the second image that we promote, this enable will become a No-OP? IOW we need to disable/enable mirroring on the dummy image for each promote of a real image, such that the real image gets scheduled.

In the current form the second enable would become a no-OP is what I think. @BenamarMk ?

Madhu-1: Yes, correct its a no-op. the mirroring is enabled on dummy image only once for all promote operations. The same goes for disable also

BenamarMk: I think I tested that and it worked. I will test it again to see if refresh happens when the dummy image is enabled for mirroring while it is already ENABLED.  But if I am wrong, then we can always do this:

Promote PV:
```
1. disable mirroring Dummy
2. promote PV
3. enable mirroring Dummy
```

Demote PV:
```
1. disable mirroring Dummy
2. demote PV
```

humblec (approved):

humblec:
> > lgtm.. considering the importance of this PR we have to take this in at the earliest and can do further enhancements later.. @Madhu-1 will we wait for testing result from @BenamarMk @ShyamsundarR before merge ?
> 
> Yes waiting for the testing result. but want to close up on review as changes look good to @BenamarMk and tested locally.
> 
> > CC @ceph/ceph-csi-contributors ptal

@Madhu-1 it looks like we are still testing this, However I have placed approval to avoid any unnecessary delay once we are done with the testing to merge.

ShyamsundarR (changes requested):
There are a few corner cases... hopefully this prevents the auto-merge

humblec:
> There are a few corner cases... hopefully this prevents the auto-merge

@ShyamsundarR there is already DNM, so that should be enough to block the merge.

Madhu-1:
closing in favor of https://github.com/ceph/ceph-csi/pull/2656

PR #2609, opened 2021-11-02
doc: add migration design documentation

humblec:
This commit adds migration design doc which carry information about
the required changes and design for rbd intree to csi migration. 

Additional info: 
This doc is pretty much the replica of tracker issue and the PRs merged ( with e2e in place) on this effort.

Updates https://github.com/ceph/ceph-csi/issues/2509
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

PR #2562, opened 2021-10-11
rbd: migration secret support

humblec:
rbd: parse migration secret and set it for CSI driver operations
    
    The intree secret has a data field called "key" which is the base64
    admin secret key. The ceph CSI driver currently expect the secret to
    contain data field "UserKey" for the equivalant. The CSI driver also
    expect the "UserID" field which is not available in the in-tree secret
    by deafult. This missing userID will be filled (if the username differ
    than 'admin') in the migration secret as 'adminId' field in the
    migration request, this commit adds the logic to parse this migration
    secret as below:
    
    "key" field value will be picked up from the migraion secret to "UserKey"
    field.
    "adminId" field value will be picked up from the migration secret to "UserID"
    field
    
    if `adminId` field is nil or not set, `UserID` field will be filled with
    default value ie `admin`.The above logic get activated only when the secret
    is a migration secret, otherwise skipped to the normal workflow as we have
    today.

Note:
E2E tests are added to make sure migration kind of secret is passed in tests.

Additional note:
The initial plan was to ask the admin to adjust the secret as a pre-upgrade step
for migration which avoid this change in the CSI driver, however taking out admin burden
is the main ask from kubernetes community so this change is introduced. None of the
existing workflow is broken for CSI driver with this change.

Updates #2509 

Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

PR #2648, opened 2021-11-18
rbd: update ControllerExpandVolume with migration secret support

humblec:
This PR address below items from the tracker/todo: 
- [x] e2e: adjust migration e2e tests and introduce helper functions
    This commit adjust existing migration e2e tests to a couple of tests
    to cover the scenarios. The seperate filesystem and block tests have
    been shrinked to single one and also introduced a couple of helper
    functions to setup and teardown migraition specific secret,configmap
    and sc. The static pv function has been renamed to a general name
    while the tests were adjusted.
- [x]  rbd: add migration secret support to controllerserver functions
    This commit adds the migration secret request validation to expand,
    create controller functions.
- [x]  e2e: remove unparam linter escapes from getStaticPV and getStaticPVC
     This `unparam` linter escape is no longer needed and CI is failing
    if we keep there. This commit remove the same and make CI happy.
       
Updates #2509 

### DIFF
--- a/e2e/migration.go
+++ b/e2e/migration.go
@@ -4,103 +4,12 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
-	"strconv"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 )
-
-func validateRBDStaticMigrationPVDeletion(f *framework.Framework, appPath, scName string, isBlock bool) error {
-	opt := make(map[string]string)
-	var (
-		rbdImageName        = "kubernetes-dynamic-pvc-e0b45b52-7e09-47d3-8f1b-806995fa4412"
-		pvName              = "pv-name"
-		pvcName             = "pvc-name"
-		namespace           = f.UniqueName
-		sc                  = scName
-		provisionerAnnKey   = "pv.kubernetes.io/provisioned-by"
-		provisionerAnnValue = "rbd.csi.ceph.com"
-	)
-
-	c := f.ClientSet
-	PVAnnMap := make(map[string]string)
-	PVAnnMap[provisionerAnnKey] = provisionerAnnValue
-	mons, err := getMons(rookNamespace, c)
-	if err != nil {
-		return fmt.Errorf("failed to get mons: %w", err)
-	}
-	mon := strings.Join(mons, ",")
-	size := staticPVSize
-	// create rbd image
-	cmd := fmt.Sprintf(
-		"rbd create %s --size=%s --image-feature=layering %s",
-		rbdImageName,
-		staticPVSize,
-		rbdOptions(defaultRBDPool))
-
-	_, stdErr, err := execCommandInToolBoxPod(f, cmd, rookNamespace)
-	if err != nil {
-		return err
-	}
-	if stdErr != "" {
-		return fmt.Errorf("failed to create rbd image %s", stdErr)
-	}
-
-	opt["migration"] = "true"
-	opt["clusterID"] = getMonsHash(mon)
-	opt["imageFeatures"] = staticPVImageFeature
-	opt["pool"] = defaultRBDPool
-	opt["staticVolume"] = strconv.FormatBool(true)
-	opt["imageName"] = rbdImageName
-
-	// Make volumeID similar to the migration volumeID
-	volID := composeIntreeMigVolID(mon, rbdImageName)
-	pv := getStaticPV(
-		pvName,
-		volID,
-		size,
-		rbdNodePluginSecretName,
-		cephCSINamespace,
-		sc,
-		provisionerAnnValue,
-		isBlock,
-		opt,
-		PVAnnMap,
-		deletePolicy)
-
-	_, err = c.CoreV1().PersistentVolumes().Create(context.TODO(), pv, metav1.CreateOptions{})
-	if err != nil {
-		return fmt.Errorf("PV Create API error: %w", err)
-	}
-
-	pvc := getStaticPVC(pvcName, pvName, size, namespace, sc, isBlock)
-
-	_, err = c.CoreV1().PersistentVolumeClaims(pvc.Namespace).Create(context.TODO(), pvc, metav1.CreateOptions{})
-	if err != nil {
-		return fmt.Errorf("PVC Create API error: %w", err)
-	}
-	// bind pvc to app
-	app, err := loadApp(appPath)
-	if err != nil {
-		return err
-	}
-
-	app.Namespace = namespace
-	app.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = pvcName
-	err = createApp(f.ClientSet, app, deployTimeout)
-	if err != nil {
-		return err
-	}
-
-	err = deletePVCAndApp("", f, pvc, app)
-	if err != nil {
-		return fmt.Errorf("failed to delete PVC and application: %w", err)
-	}
-
-	return err
-}
 
 // composeIntreeMigVolID create a volID similar to intree migration volID
 // the migration volID format looks like below
@@ -190,6 +99,7 @@ func createMigrationUserSecretAndSC(f *framework.Framework, scName string) error
 	return nil
 }
 
+// createMigrationSC create a SC with migration specific secrets and clusterid.
 func createMigrationSC(f *framework.Framework, scName string) error {
 	err := deleteResource(rbdExamplePath + "storageclass.yaml")
 	if err != nil {
@@ -203,6 +113,13 @@ func createMigrationSC(f *framework.Framework, scName string) error {
 	param["csi.storage.k8s.io/controller-expand-secret-name"] = rbdMigrationProvisionerSecretName
 	param["csi.storage.k8s.io/node-stage-secret-namespace"] = cephCSINamespace
 	param["csi.storage.k8s.io/node-stage-secret-name"] = rbdMigrationNodePluginSecretName
+	mons, err := getMons(rookNamespace, f.ClientSet)
+	if err != nil {
+		return fmt.Errorf("failed to get mons: %w", err)
+	}
+	mon := strings.Join(mons, ",")
+	param["migration"] = "true"
+	param["clusterID"] = getMonsHash(mon)
 	err = createRBDStorageClass(f.ClientSet, f, scName, nil, param, deletePolicy)
 	if err != nil {
 		return fmt.Errorf("failed to create storageclass: %w", err)
@@ -270,6 +187,43 @@ func deleteProvNodeMigrationSecret(f *framework.Framework, provisionerSecret, no
 		if err != nil {
 			return fmt.Errorf("failed to delete node secret: %w", err)
 		}
+	}
+
+	return nil
+}
+
+// setupMigrationCMSecretAndSC create custom configmap, secret and SC for migration tests.
+func setupMigrationCMSecretAndSC(f *framework.Framework, scName string) error {
+	c := f.ClientSet
+	if scName == "" {
+		scName = defaultSCName
+	}
+	err := generateClusterIDConfigMapForMigration(f, c)
+	if err != nil {
+		return fmt.Errorf("failed to generate clusterID configmap: %w", err)
+	}
+
+	err = createMigrationUserSecretAndSC(f, scName)
+	if err != nil {
+		return fmt.Errorf("failed to create storageclass: %w", err)
+	}
+
+	return nil
+}
+
+// tearDownMigrationSetup deletes custom configmap and secret.
+func tearDownMigrationSetup(f *framework.Framework) error {
+	err := deleteConfigMap(rbdDirPath)
+	if err != nil {
+		return fmt.Errorf("failed to delete configmap: %w", err)
+	}
+	err = createConfigMap(rbdDirPath, f.ClientSet, f)
+	if err != nil {
+		return fmt.Errorf("failed to create configmap: %w", err)
+	}
+	err = deleteProvNodeMigrationSecret(f, true, true)
+	if err != nil {
+		return fmt.Errorf("failed to delete migration users and Secrets associated: %w", err)
 	}
 
 	return nil
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -402,37 +402,58 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
-			// todo: may be remove the below deletion test later once the migration nodestage tests are adjusted
-			// also to have deletion validation through the same.
-			By("validate RBD migration+static Block PVC Deletion", func() {
-				err := generateClusterIDConfigMapForMigration(f, c)
+			By("validate RBD migration PVC", func() {
+				err := setupMigrationCMSecretAndSC(f, "")
 				if err != nil {
-					e2elog.Failf("failed to generate clusterID configmap: %v", err)
+					e2elog.Failf("failed to setup migration prerequisites: %v", err)
 				}
-
-				// create a sc with different migration secret
-				err = createMigrationUserSecretAndSC(f, "migrationsc")
+				err = validatePVCAndAppBinding(pvcPath, appPath, f)
 				if err != nil {
-					e2elog.Failf("failed to create storageclass: %v", err)
-				}
-				err = validateRBDStaticMigrationPVDeletion(f, rawAppPath, "migrationsc", true)
-				if err != nil {
-					e2elog.Failf("failed to validate rbd migrated static block pv: %v", err)
+					e2elog.Failf("failed to validate pvc and application binding: %v", err)
 				}
 				// validate created backend rbd images
 				validateRBDImageCount(f, 0, defaultRBDPool)
-				err = deleteConfigMap(rbdDirPath)
+
+				// Block PVC resize
+				err = resizePVCAndValidateSize(rawPvcPath, rawAppPath, f)
 				if err != nil {
-					e2elog.Failf("failed to delete configmap: %v", err)
-				}
-				err = createConfigMap(rbdDirPath, f.ClientSet, f)
-				if err != nil {
-					e2elog.Failf("failed to create configmap: %v", err)
+					e2elog.Failf("failed to resize block PVC: %v", err)
 				}
 
-				err = deleteProvNodeMigrationSecret(f, true, true)
+				// FileSystem PVC resize
+				err = resizePVCAndValidateSize(pvcPath, appPath, f)
 				if err != nil {
-					e2elog.Failf("failed to delete migration users and Secrets associated: %v", err)
+					e2elog.Failf("failed to resize filesystem PVC: %v", err)
+				}
+				err = deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete storageclass: %v", err)
+				}
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
+				if err != nil {
+					e2elog.Failf("failed to create storageclass: %v", err)
+				}
+				err = tearDownMigrationSetup(f)
+				if err != nil {
+					e2elog.Failf("failed to tear down migration setup: %v", err)
+				}
+			})
+
+			By("validate RBD migration+static FileSystem", func() {
+				err := setupMigrationCMSecretAndSC(f, "migrationsc")
+				if err != nil {
+					e2elog.Failf("failed to setup migration prerequisites: %v", err)
+				}
+				// validate filesystem pvc mount
+				err = validateRBDStaticMigrationPVC(f, appPath, "migrationsc", false)
+				if err != nil {
+					e2elog.Failf("failed to validate rbd migrated static file mode pvc: %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0, defaultRBDPool)
+				err = tearDownMigrationSetup(f)
+				if err != nil {
+					e2elog.Failf("failed to tear down migration setup: %v", err)
 				}
 				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
 				if err != nil {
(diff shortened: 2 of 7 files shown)


humblec:
The tests are failing due to the encryption vault failure.  However this PR specific tests have passed.  Appreciate review .. Thanks

internal/rbd/controllerserver.go
```diff
@@ -271,19 +271,27 @@ func checkValidCreateVolumeRequest(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSna
...
 	// the credentials to any of the utility functions.
-	cr, err := util.NewUserCredentials(req.GetSecrets())
+	secrets := req.GetSecrets()
+	if util.IsMigrationSecret(secrets) {
+		secrets, err = util.ParseAndSetSecretMapFromMigSecret(secrets)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+	cr, err := util.NewUserCredentials(secrets)
```
Madhu-1: why this change is needed?

humblec: this is required when we added migration secret support https://github.com/ceph/ceph-csi/pull/2562 , otherwise validation of the request will fail. @Madhu-1

Madhu-1: why you are adding it to createVolume? the migration is only to support the existing volume, isn't it?

humblec: I am not adding anything here @Madhu-1, not sure what is the confusion here. Migration is for existing intree storage classes. As pointed out in all the migration documentations it is the very basic thing expected from migration,  for ex: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/csi-migration.md
```
Dynamically Provisioned volumes will continue to be provisioned with the in-tree PersistentVolumeSource. The CSI external-provisioner to pick up the in-tree PVC's when migration is turned on and provision using the CSI Drivers;...
```

Madhu-1: Do you mean to say that the user will still create the CSI volumes using storageclass which contains the intree details?

humblec: it is upto the admin , however mostly what I heard was that, new CSI SCs will be given to the users thus new volumes will be on CSI SC as anyway they are in migration path.

Madhu-1: IMHO we should not support this one in cephcsi and just stick to the plan to treat migration as static only add documentation to use new SC for csi volumes.

humblec: This is very well understood and clear that new volumes has to go through and its a must requirement. We were always in plan and  never deviated from it . Also, this was  discussed in  CSI calls. `Static` treatment is only applicable to the existing intree volumes.

Madhu-1: @humblec can you please share with me the link to a  chat/doc where this is discussed in csi call? at least am not aware of any discussion. for migration, we are getting everything in bit and piece and this was not part of your design doc. now, this is news for me, and am not aware of it.

humblec: >for migration, we are getting everything in bit and piece and this was not part of your design doc. now, this is news for me, and am not aware of it.
>
@Madhu-1 it was discussed  in csi stand up calls, may be you were unavailable on those calls. It was/is part of the tracker issue, design doc ( with above link in place)..etc and there was also an explicit  mention of  provisioning and other operations in the design doc which was removed when you asked to replace that section by this comment https://github.com/ceph/ceph-csi/pull/2609/files#r744505983 . That section was replaced by the review comment  ask as it was repeating the same thing.. It is really surprising that,  it is a news . Also, this is not breaking change at all!! All existing functionality works and helping users if they opted for migration.

Madhu-1: @nixpanic are you aware of this discussion?

humblec: @Madhu-1 *at least* I remember discussing these as parf of these dates `2021-09-13 , 2021-09-14` in CSI standup calls. https://docs.google.com/document/d/1K1aerdMpraIh56-skdoEoVF9RZrO4NUcbHtjN-f3u1s/edit#heading=h.z6gjxa34z215 .  Atleast @Rakshith-R should be aware of this..Also I have provided updates in other stand up calls too. More or less, I still didnt get what is the `real concern` here? it is not breaking any existing functionality and rather helping users !!!

Madhu-1: @humblec at least I was not aware of any of these till I review this PR. the reason for asking to remove the duplicate content from the design doc here is that https://github.com/ceph/ceph-csi/pull/2609/files#r744505983 does not want to have the same exact copy-paste from the kubernetes design doc in cephcsi. if we get the design PR in first and then the implementation we can avoid these discussions and also we are doing multiple PR to achieve this one we can avoid doing cleanup introduced in the last PR like we are doing in this PR hope you got my concern

e2e/migration.go
```diff
@@ -203,6 +113,13 @@ func createMigrationSC(f *framework.Framework, scName string) error {
 	param["csi.storage.k8s.io/controller-expand-secret-name"] = rbdMigrationProvisionerSecretName
 	param["csi.storage.k8s.io/node-stage-secret-namespace"] = cephCSINamespace
 	param["csi.storage.k8s.io/node-stage-secret-name"] = rbdMigrationNodePluginSecretName
+	mons, err := getMons(rookNamespace, f.ClientSet)
+	if err != nil {
+		return fmt.Errorf("failed to get mons: %w", err)
+	}
+	mon := strings.Join(mons, ",")
+	param["migration"] = "true"
+	param["clusterID"] = getMonsHash(mon)
```
Madhu-1, line 122: is this a hack for e2e? because i don't believe the intree SC carry these parameters

internal/rbd/controllerserver.go
Rakshith-R: @Madhu-1 *at least* I remember discussing these as parf of these dates `2021-09-13 , 2021-09-14` in CSI standup calls. https://docs.google.com/document/d/1K1aerdMpraIh56-skdoEoVF9RZrO4NUcbHtjN-f3u1s/edit#heading=h.z6gjxa34z215 .  Atleast @Rakshith-R should be aware of this.

I don't remember discussing create or resize of volume using intree storage class at that point in time.

>.Also I have provided updates in other stand up calls too. More or less, I still didnt get what is the `real concern` here? it is not breaking any existing functionality and rather helping users !!! 


IMO, We should be encouraging the usage of csi storageclass.

Rakshith-R: >@Madhu-1 *at least* I remember discussing these as parf of these dates `2021-09-13 , 2021-09-14` in CSI standup calls. https://docs.google.com/document/d/1K1aerdMpraIh56-skdoEoVF9RZrO4NUcbHtjN-f3u1s/edit#heading=h.z6gjxa34z215 .  Atleast @Rakshith-R should be aware of this.
> 

 I don't remember discussing create or resize of volume using intree storage class at that point in time.
 
> >.Also I have provided updates in other stand up calls too. More or less, I still didnt get what is the `real concern` here? it is not breaking any existing functionality and rather helping users !!! 
> 



 IMO, We should be encouraging the usage of csi storageclass.
> 
> 
> 
>  
> 
>

e2e/rbd.go
```diff
@@ -367,37 +367,65 @@ var _ = Describe("RBD", func() {
...
-				err = createConfigMap(rbdDirPath, f.ClientSet, f)
+				err = tearDownMigrationSetup(f)
 				if err != nil {
-					e2elog.Failf("failed to create configmap with error %v", err)
+					e2elog.Failf("failed to tear down migration setup: %v", err)
 				}
+			})
 
-				err = deleteProvNodeMigrationSecret(f, true, true)
+			By("validate RBD migration+static FileSystem and Block PVC", func() {
```
Madhu-1: do we have Block PVC support for rbd in intree?

e2e/rbd.go
```diff
@@ -367,37 +367,65 @@ var _ = Describe("RBD", func() {
...
+				if err != nil {
+					e2elog.Failf("failed to validate rbd migrated static file mode pvc: %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0, defaultRBDPool)
+				// validate block pvc mount
+				err = validateRBDStaticMigrationPVC(f, rawAppPath, "migrationsc", true)
 				if err != nil {
-					e2elog.Failf("failed to delete migration users and Secrets associated with error %v", err)
+					e2elog.Failf("failed to validate rbd migrated static block pv with error %v", err)
```
Madhu-1: remove with error

internal/rbd/controllerserver.go
humblec: >@humblec at least I was not aware of any of these till I review this PR. the reason for asking to remove the duplicate content from the design doc here is that https://github.com/ceph/ceph-csi/pull/2609/files#r744505983 does not want to have the same exact copy-paste from the kubernetes design doc in cephcsi. if we get the design PR in first and then the implementation we can avoid these discussions and also we are doing multiple PR to achieve this one we can avoid doing cleanup introduced in the last PR like we are doing in this PR hope you got my concern
>

@Madhu-1 nw at all, I am glad that you asked the clarification. once again I was also surprised by reading this review comment. We can definitely make the process better if there are flaws , end of the day its all about helping users with the project.  :+1:

Madhu-1 (dismissed):
a couple of things here

* cant we wait for kubernetes to release with the translation library? the whole e2e is a hack to test the migration not sure it works the way that is expected from the migration side. 
* I really don't see any value in supporting the dynamic provisioning for the intree storageclass. IMO we should document that for new dynamic PVC use new csi storageclass.\

internal/rbd/controllerserver.go
humblec: >I don't remember discussing create or resize of volume using intree storage class at that point in time.' .

@Rakshith-R please refresh :), resize came in late, but new volumes and annotation hack was discussed :).

e2e/rbd.go
humblec: not neeed, I can remove that.

internal/rbd/controllerserver.go
Madhu-1: @humblec IMHO design discussion should include the maintainers :)

e2e/rbd.go
humblec: sure.

e2e/migration.go
humblec: its a hack for e2e to simulate the situation of migration SC.

internal/rbd/controllerserver.go
Madhu-1: IMO the whole process is a complete mess for me at least. if we don't have proper path things become hard at least to understand the goal and to make decisions.

humblec:
>   cant we wait for kubernetes to release with the translation library? the whole e2e is a hack to test the migration not sure it works the way that is expected from the migration side.
>

@Madhu-1   I added the e2e as it was the request since the very first PR of this effort. Indeed its good to have till we have 1.23 release is out. Once its out I will remove it.. 

>  I really don't see any value in supporting the dynamic provisioning for the intree storageclass. IMO we should document that for new dynamic PVC use new csi storageclass.\
>

Sure, we can document CSI SCs are recommend for new volumes in the doc. However we dont have a legitimate justification to block provisioning if the users really wanted to use intree SC. Also, I am afraid that, this is not an option at this stage as its the very basic requirement from kubernetes which is very well documented/set the expectations on since its start, also the change is very minor and not effecting/breaking any "existing" functionality at all !!    Also, as mentioned earlier, admin can control which SCs to submit to users...etc in a migration enabled cluster and its upto him and based on his requirement. Also it is a common thought process for all migration enabled plugin setups.

internal/rbd/controllerserver.go
humblec: @Madhu-1 indeed the process for feature enhancement can be improved.. I struggled on few topics too in the past. I can take an action item to have a discussion among the team/community here about the process and document the process here in the repo from next release onwards.. :+1:

e2e/rbd.go
humblec: done.

e2e/rbd.go
humblec: done.

internal/rbd/controllerserver.go
Rakshith-R: > .@Rakshith-R please refresh :), resize came in late, but new volumes and annotation hack was discussed :).

@humblec  As far as i remember we mainly discussed mapping required for vol-id changes. 

I only supported the node operations for older intree pvc as they are crucial,
I still don't feel we should support other functions and instead should encourage users to switch to **csi storageclass** going forward. :)

humblec: > > .@Rakshith-R please refresh :), resize came in late, but new volumes and annotation hack was discussed :).
> 
> @humblec As far as i remember we mainly discussed mapping required for vol-id changes.
> 
the vol id part  was for existing volumes, in that call new vols and annotations to identify or distinguish the volid were also discussed. any way there is no point of going back :)

>I only supported the node operations for older intree pvc as they are crucial,
>
for older intree PVCs we are only supporting node operations and then delete volume we are able to support.. creation is not for existing volumes, it is for existing SC.

internal/rbd/controllerserver.go
```diff
@@ -271,19 +271,27 @@ func checkValidCreateVolumeRequest(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSna
...
-	if err := cs.validateVolumeReq(ctx, req); err != nil {
+	var err error
+	if err = cs.validateVolumeReq(ctx, req); err != nil {
 		return nil, err
 	}
 
 	// TODO: create/get a connection from the the ConnPool, and do not pass
 	// the credentials to any of the utility functions.
-	cr, err := util.NewUserCredentials(req.GetSecrets())
+	secrets := req.GetSecrets()
```
nixpanic: Because this code is duplicated in multiple places now, you should use a helper function for it. Maybe introduce
```
cr, err := util.NewUserCredentialsWithMigration(req.GetSecrets())
```
as a variation on `util.NewUserCredentials()`.

internal/rbd/controllerserver.go
```diff
@@ -96,7 +96,7 @@ func (cs *ControllerServer) validateVolumeReq(ctx context.Context, req *csi.Crea
 
 func (cs *ControllerServer) parseVolCreateRequest(
 	ctx context.Context,
-	req *csi.CreateVolumeRequest) (*rbdVolume, error) {
+	req *csi.CreateVolumeRequest, secrets map[string]string) (*rbdVolume, error) {
```
nixpanic: This is quite ugly, the `CreateVolumeRequest` already contains secrets. Preventing to pass extra (or replacement) `secrets` would be much better. If there is really no other reasonable way to do this, it needs to be well documented for the function.

internal/rbd/controllerserver.go
```diff
@@ -1432,7 +1440,8 @@ func (cs *ControllerServer) DeleteSnapshot(
 func (cs *ControllerServer) ControllerExpandVolume(
 	ctx context.Context,
 	req *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
-	if err := cs.Driver.ValidateControllerServiceRequest(csi.ControllerServiceCapability_RPC_EXPAND_VOLUME); err != nil {
+	var err error
+	if err = cs.Driver.ValidateControllerServiceRequest(csi.ControllerServiceCapability_RPC_EXPAND_VOLUME); err != nil {
```
nixpanic: You can now just assign `err` on the previous line. It reduces the length of this line and makes it more readable.

internal/rbd/controllerserver.go
nixpanic: The design doc should be updated to include what changes are specific for Ceph-CSI. That includes mentioning functions like `ParseAndSetSecretMapFromMigSecret()`, and when that needs to be used.

For migration, I was expecting that the old in-tree PVs would be migrated to new CSI PVs. But it seems there is no real migration of that  kind. It currently looks like in-tree PVs are not _migrated_, but rather support for the in-tree format is added to Ceph-CSI. This is a little confusing...

The design doc should probably describe in a little more detail what CSI procedures need support for the old in-tree PVs for _migration_, and how that functionality is added to Ceph-CSI. A separate PR for the (updated) design doc might be appropriate, and that should be approved/merged before implementing the needed changes (ideally).

internal/rbd/controllerserver.go
humblec: @nixpanic sure, let me check , considering the secret parsing and using done in many places ( I mean here and there, ie sometimes from helper directly on request, sometimes direct usage of passed cred ..etc) I noticed some difficulty.. but let me recheck again.

humblec:
>The design doc should be updated to include what changes are specific for Ceph-CSI. That includes mentioning functions like ParseAndSetSecretMapFromMigSecret(), and when that needs to be used.
>

Yeah, will add more details..

>For migration, I was expecting that the old in-tree PVs would be migrated to new CSI PVs. But it seems there is no real migration of that kind. It currently looks like in-tree PVs are not migrated, but rather support for the in-tree format is added to Ceph-CSI. This is a little confusing...

Thats correct .

>The design doc should probably describe in a little more detail what CSI procedures need support for the old in-tree PVs for migration, and how that functionality is added to Ceph-CSI. A separate PR for the (updated) design doc might be appropriate, and that should be approved/merged before implementing the needed changes (ideally).
>

I will add details to the doc. Thanks :+1:

internal/rbd/controllerserver.go
humblec: > If there is really no other reasonable way to do this, it needs to be well documented for the function.
>
as we decided to not change the request itself,  this is the reasonable way to do this. I have also documented parseVolCreateRequest() with the details.. thanks.

internal/rbd/controllerserver.go
humblec: Sure.. done .. Thanks.

internal/rbd/controllerserver.go
```diff
@@ -1456,13 +1470,21 @@ func (cs *ControllerServer) ControllerExpandVolume(
...
+	}
+
+	cr, err := util.NewUserCredentials(secrets)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	defer cr.DeleteCredentials()
 
-	rbdVol, err := genVolFromVolID(ctx, volID, cr, req.GetSecrets())
+	rbdVol, err := genVolFromVolID(ctx, volID, cr, secrets)
```
nixpanic: These `secrets` are only needed for KMS (encryption passphrase in the linked StorageClass Secret), so the migration secrets (cluster access credentials) are not needed.

```go
	secrets := req.GetSecrets()
	if util.IsMigrationSecret(secrets) {
		secrets, err = util.ParseAndSetSecretMapFromMigSecret(secrets)
		if err != nil {
		}
	}

	cr, err := util.NewUserCredentials(secrets)
```

can therefore be replaced by a single  helper like

```go
cr, err := util.NewUserCredentialsWithMigration(req.GetSecrets());
```

or something along those lines. When the helper does not detect `IsMigrationSecret()`, it can return `NewUserCredentials()` directly. Such a single helper is much cleaner and easier to maintain copy/pasted code in different locations. It also reduces the :spaghetti: code sprinkled over multiple places for migration support.

humblec: Sure, @nixpanic , will adopt that model here.

internal/rbd/controllerserver.go
humblec: converted to above model and all tests are passing :+1:  , ptal .. thanks @nixpanic

internal/rbd/controllerserver.go
humblec: Done.

internal/rbd/controllerserver.go
humblec: Added https://github.com/ceph/ceph-csi/pull/2679 ..

internal/rbd/controllerserver.go
```diff
@@ -94,6 +94,11 @@ func (cs *ControllerServer) validateVolumeReq(ctx context.Context, req *csi.Crea
 	return nil
 }
 
+// parseVolCreateRequest take create volume `request` argument and `secret map` as its args.
+// the secretmap will be passed to genVolFromVolumeOptions() function further. Even though
+// the request carry the secretmap, we dont directly consume it from the request due to the
+// possibility that the secret can be of the form migration secret. As we are not touching the
+// request from caller in any stage, this model has been kept.
```
nixpanic: what does `this model has been kept` mean in this comment?

internal/rbd/controllerserver.go
```diff
@@ -1455,13 +1453,11 @@ func (cs *ControllerServer) ControllerExpandVolume(
 		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volID)
 	}
 	defer cs.VolumeLocks.Release(volID)
-
-	cr, err := util.NewUserCredentials(req.GetSecrets())
+	cr, err := util.NewUserCredentialsWithMigration(req.GetSecrets())
```
nixpanic, line 1445: empty lines  between different sections make the code more readable, please do not remove them

internal/rbd/controllerserver.go
```diff
@@ -1455,13 +1453,11 @@ func (cs *ControllerServer) ControllerExpandVolume(
...
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 	defer cr.DeleteCredentials()
-
```
nixpanic, line 1455: same here, keep empty lines between different code-blocks please

internal/util/credentials.go
```diff
@@ -145,11 +145,31 @@ func ParseAndSetSecretMapFromMigSecret(secretmap map[string]string) (map[string]
 // of a migration volume request. The migration secret carry a field
```
nixpanic, line 145: update the comment with the new function name too

internal/util/credentials.go
```diff
@@ -145,11 +145,31 @@ func ParseAndSetSecretMapFromMigSecret(secretmap map[string]string) (map[string]
...
 	// it is good to have this check in place, also it gives clear error about this
 	// was hit on migration request compared to general one.
 	return len(passedSecretMap) != 0 && passedSecretMap[migUserKey] != ""
 }
+
+// NewUserCredentialsWithMigration takes secret map from the request and validate it is
+// a migration secret, if yes, it continues to create CR from it after parsing the migration
+// secret. If it is not a migration it will continue the attempt to create credentials from it
+// without parsing the secret. This function returns credentials and error.
+func NewUserCredentialsWithMigration(passedSecretMap map[string]string) (*Credentials, error) {
```
nixpanic: that is `passedSecretMap`? Do you mean 'passed' as in the argument to the function? That seems rather redundant, and just `secrets` works fine everywhere else too.

humblec: done.

internal/util/credentials.go
humblec: Done.

internal/rbd/controllerserver.go
humblec: Done.

internal/rbd/controllerserver.go
humblec: Done.

internal/rbd/controllerserver.go
humblec: forgot to remove this comment, it made sense only with previous version, no longer needed above explanation. Cleaned up now.. Thanks

humblec:
@nixpanic all comments are addressed, ptal. Thanks :+1:

internal/util/credentials.go
```diff
@@ -141,15 +141,35 @@ func ParseAndSetSecretMapFromMigSecret(secretmap map[string]string) (map[string]
...
 	// was hit on migration request compared to general one.
-	return len(passedSecretMap) != 0 && passedSecretMap[migUserKey] != ""
+	return len(secrets) != 0 && secrets[migUserKey] != ""
+}
+
+// NewUserCredentialsWithMigration takes secret map from the request and validate it is
+// a migration secret, if yes, it continues to create CR from it after parsing the migration
+// secret. If it is not a migration it will continue the attempt to create credentials from it
+// without parsing the secret. This function returns credentials and error.
+func NewUserCredentialsWithMigration(secrets map[string]string) (*Credentials, error) {
```
nixpanic, line 161: this is a new helper function, so there should probably be a unit test for that?

nixpanic (commented):
Looks ok to me. There were lots of discussions with @Rakshith-R and @Madhu-1 , so I would like to see their approvals too.

internal/util/credentials.go
humblec: actually this function make use of 2 other helpers which already covered in the unit test. The part which is not covered with the unit test is for the core function `newUserWithCredentials` which existed for some time , but we havent added any unit tests because of its nature like creation of keyfile inside the tmp dir. so I think may be unit test is not required for this new helper function..

humblec:
> Looks ok to me. There were lots of discussions with @Rakshith-R and @Madhu-1 , so I would like to see their approvals too.

Thanks @nixpanic :+1: 

@Madhu-1 @Rakshith-R can you please review this PR for the latest version? I have also added more details to the doc PR.

Madhu-1 (commented):
Kubernetes 1.23 is planned for Dec 7, IMO we can remove all the e2e hacks we don't need to rework again after some time and do real e2e tests to make sure everything is working fine?

humblec:
> Kubernetes 1.23 is planned for Dec 7, IMO we can remove all the e2e hacks we don't need to rework again after some time and do real e2e tests to make sure everything is working fine?

@Madhu-1 this PR is not introducing any extra e2e tests but  rather it reduced the e2e tests from 3 to 2.  Considering this PR has already passed 2 weeks here , can we get this in ? also no more hack is required on e2e further. Once kube release is available we can consume but imo we dont have to wait for that.

Madhu-1:
> > Kubernetes 1.23 is planned for Dec 7, IMO we can remove all the e2e hacks we don't need to rework again after some time and do real e2e tests to make sure everything is working fine?
> 
> @Madhu-1 this PR is not introducing any extra e2e tests but rather it reduced the e2e tests from 3 to 2. Considering this PR has already passed 2 weeks here , can we get this in ? also no more hack is reduced on e2e further. Once kube release is available we can consume but imo we dont have to wait for that.

@nixpanic  what's your thought on this one?

nixpanic (approved):

humblec:
Thanks a lot @nixpanic :+1:

humblec:
@Madhu-1 can you please review/approve this , I would like to wrap this without further delay..

Madhu-1:
@humblec  can you please don't dismiss the review requests. if any changes are requested and someone else comes and approves the PR it will get merged.

Madhu-1 (approved):

PR #2656, opened 2021-11-18
rbd: add a workaround to fix rbd snapshot scheduling

ShyamsundarR:
currently, we have a bug in the rbd mirror scheduling module. After doing failover and failback the scheduling is not getting updated and the mirroring snapshots are not getting created periodically as per the scheduling interval. This PR workaround this one by doing the below operations

- Create a dummy (csi-vol-dummy-ceph fsID) image per cluster and this image should be easily identified.

- During Promote operation on any image enables the mirroring on the dummy image. when we enable the mirroring on the dummy image the pool will get updated and the scheduling will be reconfigured.

- During Demote operation on any image disables the mirroring on the dummy image. the disabled need to be done to enable the mirroring again when we get the promote request to make the image as the primary

- When the DR is no more needed, this image needs to be manually cleanup for now as we don't want to add a check in the existing DeleteVolume code path for deleting dummy images as it impacts the performance of the DeleteVolume workflow.

Moved to add scheduling to the promote operation as scheduling need to be added when the image is promoted and this is the correct method of adding the scheduling to make the scheduling take place.

More details at https://bugzilla.redhat.com/show_bug.cgi?id=2019161
Signed-off-by: Madhu Rajanna madhupr007@gmail.com
Signed-off-by: Shyamsundar Ranganathan <srangana@redhat.com>

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -117,12 +117,10 @@ func getMirroringMode(ctx context.Context, parameters map[string]string) (librbd
 	return mirroringMode, nil
 }
 
-// getSchedulingDetails gets the mirroring mode and scheduling details from the
+// validateSchedulingDetails gets the mirroring mode and scheduling details from the
 // input GRPC request parameters and validates the scheduling is only supported
 // for snapshot mirroring mode.
-func getSchedulingDetails(parameters map[string]string) (admin.Interval, admin.StartTime, error) {
-	admInt := admin.NoInterval
-	adminStartTime := admin.NoStartTime
+func validateSchedulingDetails(parameters map[string]string) error {
 	var err error
 
 	val := parameters[imageMirroringKey]
@@ -134,43 +132,49 @@ func getSchedulingDetails(parameters map[string]string) (admin.Interval, admin.S
 	// an optional parameter.
 	case "":
 	default:
-		return admInt, adminStartTime, status.Error(codes.InvalidArgument, "scheduling is only supported for snapshot mode")
+		return status.Error(codes.InvalidArgument, "scheduling is only supported for snapshot mode")
 	}
 
 	// validate mandatory interval field
 	interval, ok := parameters[schedulingIntervalKey]
 	if ok && interval == "" {
-		return admInt, adminStartTime, status.Error(codes.InvalidArgument, "scheduling interval cannot be empty")
+		return status.Error(codes.InvalidArgument, "scheduling interval cannot be empty")
 	}
-	adminStartTime = admin.StartTime(parameters[schedulingStartTimeKey])
+	adminStartTime := admin.StartTime(parameters[schedulingStartTimeKey])
 	if !ok {
 		// startTime is alone not supported it has to be present with interval
 		if adminStartTime != "" {
-			return admInt, admin.NoStartTime, status.Errorf(codes.InvalidArgument,
+			return status.Errorf(codes.InvalidArgument,
 				"%q parameter is supported only with %q",
 				schedulingStartTimeKey,
 				schedulingIntervalKey)
 		}
 	}
 	if interval != "" {
-		admInt, err = validateSchedulingInterval(interval)
+		err = validateSchedulingInterval(interval)
 		if err != nil {
-			return admInt, admin.NoStartTime, status.Error(codes.InvalidArgument, err.Error())
+			return status.Error(codes.InvalidArgument, err.Error())
 		}
 	}
 
-	return admInt, adminStartTime, nil
+	return nil
+}
+
+// getSchedulingDetails returns scheduling interval and scheduling startTime.
+func getSchedulingDetails(parameters map[string]string) (admin.Interval, admin.StartTime) {
+	return admin.Interval(parameters[schedulingIntervalKey]),
+		admin.StartTime(parameters[schedulingStartTimeKey])
 }
 
 // validateSchedulingInterval return the interval as it is if its ending with
 // `m|h|d` or else it will return error.
-func validateSchedulingInterval(interval string) (admin.Interval, error) {
+func validateSchedulingInterval(interval string) error {
 	re := regexp.MustCompile(`^\d+[mhd]$`)
 	if re.MatchString(interval) {
-		return admin.Interval(interval), nil
+		return nil
 	}
 
-	return "", errors.New("interval specified without d, h, m suffix")
+	return errors.New("interval specified without d, h, m suffix")
 }
 
 // EnableVolumeReplication extracts the RBD volume information from the
@@ -189,7 +193,7 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 	}
 	defer cr.DeleteCredentials()
 
-	interval, startTime, err := getSchedulingDetails(req.GetParameters())
+	err = validateSchedulingDetails(req.GetParameters())
 	if err != nil {
 		return nil, err
 	}
@@ -235,19 +239,6 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 
 			return nil, status.Error(codes.Internal, err.Error())
 		}
-	}
-
-	if interval != "" {
-		err = rbdVol.addSnapshotScheduling(interval, startTime)
-		if err != nil {
-			return nil, err
-		}
-		log.DebugLog(
-			ctx,
-			"Added scheduling at interval %s, start time %s for volume %s",
-			interval,
-			startTime,
-			rbdVol)
 	}
 
 	return &replication.EnableVolumeReplicationResponse{}, nil
@@ -435,6 +426,20 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 
 			return nil, status.Error(codes.Internal, err.Error())
 		}
+	}
+
+	interval, startTime := getSchedulingDetails(req.GetParameters())
+	if interval != admin.NoInterval {
+		err = rbdVol.addSnapshotScheduling(interval, startTime)
+		if err != nil {
+			return nil, err
+		}
+		log.DebugLog(
+			ctx,
+			"Added scheduling at interval %s, start time %s for volume %s",
+			interval,
+			startTime,
+			rbdVol)
 	}
 
 	return &replication.PromoteVolumeResponse{}, nil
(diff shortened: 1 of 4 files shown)


internal/rbd/replicationcontrollerserver.go
```diff
@@ -303,52 +302,33 @@ func createDummyImage(ctx context.Context, rbdVol *rbdVolume) error {
...
+	if err != nil {
+		return err
 	}
+	dummyVol := rbdVol
+	dummyVol.RbdImageName = imgName
 
-	return nil
-}
+	dummyImageOpsLock.Lock()
+	err = dummyVol.disableImageMirroring(false)
```
BenamarMk, line 318: (nit) Should we ignore any errors here?

BenamarMk (approved):
LGTM. I will test the change.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -237,20 +259,78 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
...
+	err = dummyVol.enableImageMirroring(mirroringMode)
+	if err != nil {
+		return err
+	}
+
+	err = dummyVol.addSnapshotScheduling(admin.Interval("1m"), admin.NoStartTime)
+	if err != nil {
+		return err
+	}
+	dummyImageOpsLock.Unlock()
```
Madhu-1: Unlock should be on defer? If there is any error in above steps the unlock never happens and we wount be able to get lock again?

BenamarMk (commented):
@ShyamsundarR I tested your patch and it works as expected.  So it is fine other than the legitimate comment from @Madhu-1. We should be able to merge it once that is fixed.

internal/rbd/replicationcontrollerserver.go
ShyamsundarR: Thanks, fixed now.

ShyamsundarR:
Latest push was to add the "." to the end of the comment to address linter failure.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -237,20 +259,78 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
...
+	if err != nil {
+		return err
+	}
+
+	err = dummyVol.enableImageMirroring(mirroringMode)
+	if err != nil {
+		return err
+	}
+
+	err = dummyVol.addSnapshotScheduling(admin.Interval("1m"), admin.NoStartTime)
```
ShyamsundarR, line 328: Technically like create image we can do this call to set the schedule only once and not always. But not optimizing this yet as tests around the DR workflow is completed with this patch.

internal/rbd/replicationcontrollerserver.go
ShyamsundarR: I am thinking we do not, say we ignore it, and hence the enable becomes a no-op as it was never disabled, and causes the schedule to not trigger, we would be back at square one.

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio rebase

internal/rbd/replicationcontrollerserver.go
Madhu-1: @ShyamsundarR if csi driver is restarted after disable and before enable will it be a problem? Won't we get any error here?

ShyamsundarR: Multiple disables on a disabled image is not throwing errors in the CLI, validated the same, we should be fine under that assumption.

Madhu-1: Yes correct. no problem here

Madhu-1 (approved):

internal/rbd/replicationcontrollerserver.go
```diff
@@ -63,6 +64,18 @@ const (
...
 )
 
+type operation string
+
+var (
+	// pool+"/"+key to check dummy image is created.
+	dummyImageCreated operation = "dummyImageCreated"
+	// Read write lock to ensure that only one operation is happening at a time.
+	operationLock = sync.Map{}
+
```
humblec, line 74: nit:extra line not needed, but we can live with that.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -437,6 +517,32 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 	}
 
+	var mode librbd.ImageMirrorMode
+	mode, err = getMirroringMode(ctx, req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get mirroring mode %s", err.Error())
+	}
+
+	log.DebugLog(ctx, "Attempting to tickle dummy image for restarting RBD schedules")
```
humblec, line 526: Attempting/attempting

Madhu-1:
>I1119 03:03:23.221198       1 connection.go:187] GRPC error: rpc error: code = InvalidArgument desc = invalid encryption kms configuration: failed connecting to Vault: failed to get the authentication token: Put "http://vault.cephcsi-e2e-ab3babb4a6ad.svc.cluster.local:8200/v1/auth/kubernetes/login": dial tcp: lookup vault.cephcsi-e2e-ab3babb4a6ad.svc.cluster.local: no such host

encryption tests are failing. not related to this PR.

Madhu-1:
@humblec looks like cephcsi CI is broken due to a vault issue. am planning to merge this one manually. WDYT?

Madhu-1:
> > I1119 03:03:23.221198       1 connection.go:187] GRPC error: rpc error: code = InvalidArgument desc = invalid encryption kms configuration: failed connecting to Vault: failed to get the authentication token: Put "http://vault.cephcsi-e2e-ab3babb4a6ad.svc.cluster.local:8200/v1/auth/kubernetes/login": dial tcp: lookup vault.cephcsi-e2e-ab3babb4a6ad.svc.cluster.local: no such host
> 
> encryption tests are failing. not related to this PR.

https://github.com/ceph/ceph-csi/issues/2657

Madhu-1:
@humblec are you expecting a change for the above comments as part of this PR?

humblec:
> @ShyamsundarR I tested your patch and it works as expected. So it is fine other than the legitimate comment from @Madhu-1. We should be able to merge it once that is fixed.

Thanks @BenamarMk for confirming. 

>@humblec are you expecting a change for the above comments as part of this PR?

@Madhu-1  yeah, the CI failure is on encryption, considering the urgency of this PR, I am fine to merge manually :+1: , I have a couple of minor comments though.

Madhu-1:
> > @ShyamsundarR I tested your patch and it works as expected. So it is fine other than the legitimate comment from @Madhu-1. We should be able to merge it once that is fixed.
> 
> Thanks @BenamarMk for confirming.
> 
> > @humblec are you expecting a change for the above comments as part of this PR?
> 
> @Madhu-1 yeah, the CI failure is on encryption, considering the urgency of this PR, I am fine to merge manually +1 , I have a couple of minor comments though.

@humblec we can address that in a follow-up PR as it's not a blocker for this PR?

humblec:
@Madhu-1 @ShyamsundarR compared to previous PR, I couldnt get the corner case this PR fixed though. Appreciated if we can have a quick mention some where in this PR.  It looks to me that, now the dummy image scheduling has been set to short/1minute period for other images to rearrange the reschedule  ( may be more iterations for them to reschedule) compared to previous version, I could be wrong.

humblec:
> > > @ShyamsundarR I tested your patch and it works as expected. So it is fine other than the legitimate comment from @Madhu-1. We should be able to merge it once that is fixed.
> > 
> > 
> > Thanks @BenamarMk for confirming.
> > > @humblec are you expecting a change for the above comments as part of this PR?
> > 
> > 
> > @Madhu-1 yeah, the CI failure is on encryption, considering the urgency of this PR, I am fine to merge manually +1 , I have a couple of minor comments though.
> 
> @humblec we can address that in a follow-up PR as it's not a blocker for this PR?

sure, nw  go ahead

humblec (approved):

Madhu-1:
> @Madhu-1 @ShyamsundarR compared to previous PR, I couldnt get the corner case this PR fixed though. Appreciated if we can have a quick mention some where in this PR. It looks to me that, now the dummy image scheduling has been set to short/1minute period for other images to rearrange the reschedule ( may be more iterations for them to reschedule) compared to previous version, I could be wrong.

the previous PR was not resetting the scheduling for all the rbd images the enable and disable was done only on one PromoteVolume. To reset the scheduling the enable and disable of mirroring need to be done on all the images. more detail are in this PR description

BenamarMk:
Tested an app with multiple PVCs (for the same app) and failover/relocation multiple times and the fix worked. 
I have noticed while testing (kind of) an issue that might resolve/or workaround the fsck issue (BZ:2021460).  I'll investigate next.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -437,6 +517,32 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
...
+	err = tickleMirroringOnDummyImage(rbdVol, mode)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to enable mirroring on dummy image %s", err.Error())
+	}
+
+	interval, startTime := getSchedulingDetails(req.GetParameters())
+	if interval != admin.NoInterval {
+		err = rbdVol.addSnapshotScheduling(interval, startTime)
```
BenamarMk, line 534: @ShyamsundarR @Madhu-1 we have a bug in this line. `rbdVol` was changed to `rbdVol dummy` object in line 527 changed.  In this line, we still point to the same dummy rbdVol.  Adding schedule here is adding it to the wrong image.  I didn't catch this in my testing because my real image also has a schedule of 1m.  If you change the real image schedule to something else, you will see the dummy image schedule added to "something else".

@Madhu-1 It is an easy fix.  Can you make the change?

Madhu-1: got it. tested the fix making a PR.

BenamarMk: Thanks @Madhu-1 for providing the fix so quickly.  I have tested patch #2669 and it works as intended.

Issue #2671, opened 2021-11-23
rbd: implement spec for identity discovery


Rakshith-R:
CSI-Addons Nodes and CSI-drivers need to provide a way that makes it
possible to discover supported features. The Identity Service is a
specification that allows the CSI-Addons side-car to provide basic
information about itself and the capabilities the CSI-driver implements.

More details at https://github.com/csi-addons/spec/pull/28

PR #2683, opened 2021-11-29
rbd: add framework for CSI-Addons Identity Service

nixpanic:
Add a new CSI-Addons Server and empty Identity Service for the RBD
plugin. The implementation of the Identity Service procedure calls will
be done in other PRs.

Updates: #2671

---

<details>

  progress or results

</details>


internal/csi-addons/server/server.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// CSIAddonsServer is the gRPC server that listens on an endpoint (UNIX domain
+// socket) where the CSI-Addons requests come in.
+type CSIAddonsServer struct {
+	// URL components to listen on the UNIX domain socket
+	scheme string
+	path   string
+
+	// state of the CSIAddonsServer
```
humblec, line 48: below field does not carry any  `state`  in general.. so wondering are there better explanation on below fields. :)

internal/csi-addons/server/server.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	// setup the UNIX domain socket
+	if e := os.Remove(cas.path); e != nil && !os.IsNotExist(e) {
+		return fmt.Errorf("failed to remove %q: %w", cas.path, e)
+	}
+
+	listener, err := net.Listen(cas.scheme, cas.path)
+	if err != nil {
+		return fmt.Errorf("failed to listen on %q: %w", cas.path, err)
```
humblec, line 102: %s would be good, isnt it ?

cmd/cephcsi.go
```diff
@@ -128,6 +128,9 @@ func init() {
 	flag.BoolVar(&conf.Version, "version", false, "Print cephcsi version information")
 	flag.BoolVar(&conf.EnableProfiling, "enableprofiling", false, "enable go profiling")
 
+	// CSI-Addons configuration
+	flag.StringVar(&conf.CSIAddonsEndpoint, "csi-addons-endpoint", "unix://tmp/csi-addons.sock", "CSI-Addons endpoint")
```
Madhu-1, line 132: question:- Do we need to start one more server? can't we serve the new Spec in the existing server?

internal/csi-addons/server/server.go
nixpanic: for what? the path is nicer to be quoted, and the error needs `%w` so that `errors.Is()` can be used by the caller if needed.

cmd/cephcsi.go
nixpanic: We could, but that does not look very clean to me. It are two distinct projects (csi, csi-addons), and maybe at one point (parts of) the csi-addons get included in the CSI spec. Keeping the separate prevents confusion and maybe even potential conflicts in the future.

Madhu-1 (approved):

yati1998 (approved):

PR #2684, opened 2021-11-30
rbd: check local image state during promote operation

Madhu-1:
rbd mirroring CLI calls are async and it doesn't wait for the operation to be completed. ex:- `rbd mirror image enable`
it will enable the mirroring on the image but it doesn't ensure that the image is mirroring enabled and healthy
primary. The same goes for the promote volume also. This commits adds a check-in PromoteVolume to make sure the image is in a healthy state i.e `up+stopped`.

**note:-** 
* Not considering any intermediate states to make sure the image is completely healthy before responding success to the RPC call.
* With https://github.com/csi-addons/volume-replication-operator/pull/126 only PromoteVolume will be sent to CSI driver if the VR is already marked as primary previously. Adding local image state check in PromoteVolume to take care of failover and fallback.

### DIFF
--- a/internal/rbd/errors.go
+++ b/internal/rbd/errors.go
@@ -42,4 +42,6 @@ var (
 	ErrMissingImageNameInVolID = errors.New("rbd image name information can not be empty in volID")
 	// ErrDecodeClusterIDFromMonsInVolID is returned when mons hash decoding on migration volID.
 	ErrDecodeClusterIDFromMonsInVolID = errors.New("failed to get clusterID from monitors hash in volID")
+	// ErrUnHealthyMirroredImage is returned when mirrored image is not healthy.
+	ErrUnHealthyMirroredImage = errors.New("mirrored image is not healthy")
 )
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -536,6 +536,11 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 	}
 
+	err = checkHealthyPrimary(ctx, rbdVol)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
 	var mode librbd.ImageMirrorMode
 	mode, err = getMirroringMode(ctx, req.GetParameters())
 	if err != nil {
@@ -563,6 +568,35 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 	}
 
 	return &replication.PromoteVolumeResponse{}, nil
+}
+
+// checkHealthyPrimary checks if the image is a healhty primary or not.
+// healthy primary image will be in up+stopped state, for states other
+// than this it returns an error message.
+func checkHealthyPrimary(ctx context.Context, rbdVol *rbdVolume) error {
+	mirrorStatus, err := rbdVol.getImageMirroringStatus()
+	if err != nil {
+		return err
+	}
+	localStatus, err := mirrorStatus.LocalStatus()
+	if err != nil {
+		// LocalStatus can fail if the local site status is not found in
+		// mirroring status. Log complete sites status to debug why getting
+		// local status failed
+		log.ErrorLog(ctx, "mirroring status is %+v", mirrorStatus)
+
+		return fmt.Errorf("failed to get local status: %w", err)
+	}
+
+	if !localStatus.Up && localStatus.State != librbd.MirrorImageStatusStateStopped {
+		return fmt.Errorf("%s %w. State is up=%t, state=%q",
+			rbdVol,
+			ErrUnHealthyMirroredImage,
+			localStatus.Up,
+			localStatus.State)
+	}
+
+	return nil
 }
 
 // DemoteVolume extracts the RBD volume information from the


internal/rbd/mirror.go
```diff
@@ -69,6 +69,24 @@ func (ri *rbdImage) getImageMirroringInfo() (*librbd.MirrorImageInfo, error) {
 	return info, nil
 }
 
+// checkHealthyPrimary checks if the image is in up+stopped state.
+func (ri *rbdImage) checkHealthyPrimary() error {
+	mirrorStatus, err := ri.getImageMirroringStatus()
+	if err != nil {
+		return fmt.Errorf("failed to get mirroring info of %q with error: %w", ri, err)
```
yati1998: `fmt.Errorf("failed to get mirroring info of %q: %w", ri, err)` , `with error` is not required.

yati1998 (commented):
LGTM

Madhu-1:
below is the output from VR condition when the image is not healthy primary

```yaml
conditions:
    - lastTransitionTime: "2021-11-30T06:40:24Z"
      message: ""
      observedGeneration: 8
      reason: FailedToPromote
      status: "False"
      type: Completed
    - lastTransitionTime: "2021-11-30T06:40:24Z"
      message: ""
      observedGeneration: 8
      reason: Error
      status: "True"
      type: Degraded
    - lastTransitionTime: "2021-11-30T06:35:58Z"
      message: ""
      observedGeneration: 8
      reason: NotResyncing
      status: "False"
      type: Resyncing
    lastCompletionTime: "2021-11-30T06:35:58Z"
    lastStartTime: "2021-11-26T05:50:08Z"
    message: 'rpc error: code = Internal desc = replicapool/csi-vol-b2f4ee8a-4e7c-11ec-9b91-0242ac110005
      is not healthy. Image state is up=false, state="unknown"'
```

internal/rbd/replicationcontrollerserver.go
```diff
@@ -546,6 +551,32 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 	return &replication.PromoteVolumeResponse{}, nil
 }
 
+// checkHealthyPrimary checks if the image is in up+stopped state.
```
nixpanic: I assume that this function returns no error when the image is Healthy and Primary. How does that relate to `up+stopped`?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -546,6 +551,32 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
...
+		// LocalStatus can fail if the local site status is not found in
+		// mirroring status. Log complete sites status to debug why getting
+		// local status failed
+		log.ErrorLog(ctx, "mirroring status is %+v", mirrorStatus)
+
+		return fmt.Errorf("failed to get local status: %w", err)
+	}
+
+	if !localStatus.Up && localStatus.State != librbd.MirrorImageStatusStateStopped {
+		return fmt.Errorf("%s is not healthy. Image state is up=%t, state=%q",
```
nixpanic: There is no `%w` here, can there be an error that callers can check with `errors.Is()`?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -517,6 +517,11 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 	}
 
+	err = checkHealthyPrimary(ctx, rbdVol)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
```
nixpanic, line 541: If there is an expected delay (as I understand from the commit message), should this not get retried?

internal/rbd/replicationcontrollerserver.go
Madhu-1: up+stopped state means the image is healthy and primary, if the image is not in up+stopped state it returns an error message.

internal/rbd/replicationcontrollerserver.go
Madhu-1: AFAIK I don't think it's needed. no caller is going to extract the error message for comparison. this is directly sent to GRPC

internal/rbd/replicationcontrollerserver.go
Madhu-1: Yes, it will be retried by the caller for us it's an internal server error. even after promoting the system as not reached the state what we expected.

internal/rbd/replicationcontrollerserver.go
nixpanic: Right. It probably helps to clarify that in the comments too.

internal/rbd/replicationcontrollerserver.go
nixpanic: It was just a suggestion, not a requested change. By using `%w` and pre-defined errors functions are easier to re-use. It might be useful in the future, hence the consideration.

internal/rbd/replicationcontrollerserver.go
Madhu-1: added a comment :+1:

internal/rbd/replicationcontrollerserver.go
Madhu-1: added new error and returned with `%w` for future.

nixpanic (dismissed):

internal/rbd/errors.go
```diff
@@ -42,4 +42,6 @@ var (
 	ErrMissingImageNameInVolID = errors.New("rbd image name information can not be empty in volID")
 	// ErrDecodeClusterIDFromMonsInVolID is returned when mons hash decoding on migration volID.
 	ErrDecodeClusterIDFromMonsInVolID = errors.New("failed to get clusterID from monitors hash in volID")
+	// ErrUnHealthyImage is returned when image is not healthy.
+	ErrUnHealthyImage = errors.New("image is not healthy")
```
humblec: @Madhu-1 do we need to add `mirrored image...` instead of `image` , so that it clear that, this error is specifically for mirroring enabled one?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -517,6 +517,11 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 	}
 
+	err = checkHealthyPrimary(ctx, rbdVol)
```
humblec, line 539: may be I am missing something here: so the check happens and if it fails we return error , I mean this is not a blocking operation.

Madhu-1: sorry, i didn't get you. can you please explain a bit more?

internal/rbd/errors.go
Madhu-1: i can do it.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -546,6 +551,35 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 	return &replication.PromoteVolumeResponse{}, nil
 }
 
+// checkHealthyPrimary checks if the is a healhty primary or not.
```
humblec: nit: typo. `if the is....` -

yati1998: `checks if the "image" is healhty primary or not.`

internal/rbd/replicationcontrollerserver.go
```diff
@@ -546,6 +551,35 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
...
+// healthy primary image will be in up+stopped state, for states other
+// than this it returns an error message.
+func checkHealthyPrimary(ctx context.Context, rbdVol *rbdVolume) error {
+	mirrorStatus, err := rbdVol.getImageMirroringStatus()
+	if err != nil {
+		return err
```
yati1998, line 579: `return err` -> `return fmt.Errorf("failed to get mirror status: %w", err)` ?

Madhu-1: `getImageMirroringStatus` already returning the same error message. doing this will add a duplicate error message.

yati1998 (approved):
LGTM

nixpanic (approved):

internal/rbd/replicationcontrollerserver.go
humblec: @Madhu-1 I was mainly asking that, once we return an error do we come back here and try again in next iteration? because I was thinking about a scenario where the `up+stopped` was done with a delay in backend , meanwhile this check was performed and failed.

Madhu-1: once we return the error as it's an RPC call the VR operator will reissue it till the PromoteVolume succeeds. the whole idea of adding this check is to make sure the primary image is healthy before marking it as primary in  VR.

humblec (approved):

Issue #2675, opened 2021-11-23
rbd: removed a workaround to fix rbd snapshot scheduling


Madhu-1:
Removed to code workaround introduced in https://github.com/ceph/ceph-csi/pull/2656 to workaround the rbd snapshot scheduling bug.  once the rbd fixes the issue we will remove this in cephcsi

PR needs to be reverted back

- [ ] #2656 
- [ ] #2669
- [ ] #2694

humblec:
@Madhu-1 iic, this is done, isnt it ?

Madhu-1:
@humblec this is a tracker to remove the workaround in cephcsi. once ceph identifies the issue will remove the workaround.

humblec:
Thanks!

Madhu-1:
https://github.com/ceph/ceph-csi/pull/3413 fix it

PR #2694, opened 2021-12-07
rbd: recreate dummy image with new image features and 1MiB size

Madhu-1:
we added a workaround for rbd scheduling by creating a dummy image in #2656. with the fix, we are creating a dummy image of the size of the first actual rbd image which is sent in EnableVolumeReplication request if the actual rbd image size is 1Tib we are creating a dummy image of 1TiB which is not good. even though it's a thin-provisioned rbd image this is causing issues for the transfer of the snapshot during the mirroring operation. The current changes include recreating the rbd image with 1Mib size which is the smaller supported size in rbd.

The dummy image will be created with a 1MiB size. during the snapshot transfer operation, the 1MiB will be transferred even if the dummy image does not contain any data. adding the new image features `fast-diff,layering,obj-map,exclusive-lock` on the dummy image will ensure that only the diff is transferred to the remote cluster.

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -300,23 +300,46 @@ func getOperationName(poolName string, optName operation) string {
 // createDummyImage creates a dummy image as a workaround for the rbd
 // scheduling problem.
 func createDummyImage(ctx context.Context, rbdVol *rbdVolume) error {
+	var err error
+	var imgName string
+
+	dummyImageOpsLock.Lock()
+	defer dummyImageOpsLock.Unlock()
 	optName := getOperationName(rbdVol.Pool, dummyImageCreated)
 	if _, ok := operationLock.Load(optName); !ok {
 		// create a dummy image
-		imgName, err := getDummyImageName(rbdVol.conn)
+		imgName, err = getDummyImageName(rbdVol.conn)
 		if err != nil {
 			return err
 		}
 		dummyVol := *rbdVol
 		dummyVol.RbdImageName = imgName
+		// create 1MiB dummy image. 1MiB=1048576 bytes
+		dummyVol.VolSize = 1048576
 		err = createImage(ctx, &dummyVol, dummyVol.conn.Creds)
-		if err != nil && !strings.Contains(err.Error(), "File exists") {
-			return err
+		if err != nil {
+			if strings.Contains(err.Error(), "File exists") {
+				err = repairDummyImage(ctx, &dummyVol)
+			}
 		}
-		operationLock.Store(optName, true)
+		if err == nil {
+			operationLock.Store(optName, true)
+		}
 	}
 
-	return nil
+	return err
+}
+
+// repairDummyImage deletes and recreates the dummy image.
+func repairDummyImage(ctx context.Context, dummyVol *rbdVolume) error {
+	// deleting and recreating the dummy image will not impact anything as its
+	// a workaround to fix the scheduling problem.
+	err := deleteImage(ctx, dummyVol, dummyVol.conn.Creds)
+	if err != nil {
+		return err
+	}
+
+	return createImage(ctx, dummyVol, dummyVol.conn.Creds)
 }
 
 // tickleMirroringOnDummyImage disables and reenables mirroring on the dummy image, and sets a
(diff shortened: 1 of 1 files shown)


Madhu-1:
cc @BenamarMk

internal/rbd/replicationcontrollerserver.go
```diff
@@ -300,23 +300,56 @@ func getOperationName(poolName string, optName operation) string {
 // createDummyImage creates a dummy image as a workaround for the rbd
 // scheduling problem.
 func createDummyImage(ctx context.Context, rbdVol *rbdVolume) error {
+	var err error
+	var imgName string
 	optName := getOperationName(rbdVol.Pool, dummyImageCreated)
 	if _, ok := operationLock.Load(optName); !ok {
```
ShyamsundarR, line 309: To prevent parallel dummy image creation or size update, should we just add the:
```
	dummyImageOpsLock.Lock()
	defer dummyImageOpsLock.Unlock()
```
here as well?

Previously image exists was ignored, and so a race to create would not touch the image. Now as it is being recreated (even when size is the same) it is potentially possible that the first call is now attempting the tickle of the schedule when the subsequent call is deleting the image.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -300,23 +300,56 @@ func getOperationName(poolName string, optName operation) string {
...
-		if err != nil && !strings.Contains(err.Error(), "File exists") {
-			return err
+		if err != nil {
+			if strings.Contains(err.Error(), "File exists") {
+				err = repairDummyImage(ctx, &dummyVol)
+			}
 		}
+	}
+
+	if err == nil {
```
ShyamsundarR: Although harmless outside the Load for the same, should this move inside the if block above?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -300,23 +300,56 @@ func getOperationName(poolName string, optName operation) string {
...
+		features := librbd.FeatureSetFromNames(f)
+		dummyVol.imageFeatureSet = features
+		// create 1MiB dummy image. 1MiB=1048576 bytes
+		dummyVol.VolSize = 1048576
 		err = createImage(ctx, &dummyVol, dummyVol.conn.Creds)
-		if err != nil && !strings.Contains(err.Error(), "File exists") {
-			return err
+		if err != nil {
+			if strings.Contains(err.Error(), "File exists") {
+				err = repairDummyImage(ctx, &dummyVol)
```
BenamarMk, line 330: If we repair the dummy image here, then we will be repairing it for as long as the workaround is in place. IOW, we will be repairing even the one that does not need repairing.  Can we not repair an existing dummy image and just live with it?  The defective dummy image only exists in a few test systems, and they will go away once these test systems are rebuilt (or we can delete them manually). Thereafter, we will never need to repair them, whereas if we leave this call here, we will always delete and recreate the dummy image.  This is not a big deal and we can live with it as is, knowing that this is a temporary fix. But I guess, the more we reduce the stress on RBD, the better.

Madhu-1: as it is a dummy thin-provisioned 1Mib  image it won't create must stress on the rbd side

Below is the length checks that need to be added if we don't recreate the rbd image

* if the rbd image already  exists, get the details
* check image size is 1Mib if not resize the rbd image
* check rbd image is has required image features
* if the image does not have the required features update the image features on the image

the whole dummy image code will be removed from cephcsi once rbd fixes the scheduling problem. as this is for the workaround and repair will only happen when cephcsi is restarted/ during first-time replication. we can go with a simple workaround.

BenamarMk: I agree.

BenamarMk (approved):

ShyamsundarR (approved):
@BenamarMk take a look, things look fine otherwise.

agarwal-mudit (approved):

nixpanic (approved):
Looks OK to me. I am very interested to see an issue (linked with all these workaround fixes) that tracks the Ceph fix, so that we can drop the dummy-image workaround after that. The fixes for this workaround are getting more complex and (still) untested in e2e.

Madhu-1:
> Looks OK to me. I am very interested to see an issue (linked with all these workaround fixes) that tracks the Ceph fix, so that we can drop the dummy-image workaround after that. The fixes for this workaround are getting more complex and (still) untested in e2e.

updated TODO list to track the PR's for workaround on https://github.com/ceph/ceph-csi/issues/2675

Issue #2672, opened 2021-11-23
rbd: implement spec for CSI-Addons Operation ReclaimSpace


Rakshith-R:
When executing a ReclaimSpace operation the storage consumption of the
volume is tried to be reduced. The Storage Provider can implement this
by calling functions like fstrim, blkdiscard or similar on the
volume.

More details at https://github.com/csi-addons/spec/pull/30

```
// Controller holds the RPC method for running discard operations on the
// inactive (not required to be staged/published) volume.
service Controller {
  // ControllerReclaimSpace is a procedure that gets called on the CSI
  // Controller.
  rpc ControllerReclaimSpace (ControllerReclaimSpaceRequest)
  returns (ControllerReclaimSpaceResponse) {}
}

// Node holds the RPC method for running discard operations on the active
// (staged/published) volume.
service Node {
  // NodeReclaimSpace is a procedure that gets called on the CSI NodePlugin.
  rpc NodeReclaimSpace (NodeReclaimSpaceRequest)
  returns (NodeReclaimSpaceResponse) {}
}
```

humblec:
@Rakshith-R anything pending here , specifically from ceph csi repo side? please revisit the state as its marked for 3.5 release.

Rakshith-R:
> @Rakshith-R anything pending here , specifically from ceph csi repo side? please revisit the state as its marked for 3.5 release.

I think it's done, but I'll let @nixpanic confirm there's nothing left.

nixpanic:
> @Rakshith-R anything pending here , specifically from ceph csi repo side? please revisit the state as its marked for 3.5 release.

The functionality is available, and tested manually. There is no automated e2e verification yet, so this issue has not been closed either.

PR #2702, opened 2021-12-09
rbd: move driver component into the rbd-driver package

nixpanic:
The rbd package contains several functions that can be used by
CSI-Addons Service implementations. Unfortunately it is not possible to
do this, as the rbd-driver needs to import the csi-addons/rbd package to
provide the CSI-Addons server. This causes a circular import when
services use the rbd package:

 - rbd/driver.go import csi-addons/rbd
 - csi-addons/rbd import rbd (including the driver)

By moving rbd/driver.go into its own package, the circular import can be
prevented.

Updates: #2672 

---

<details>

  progress or results

</details>

### DIFF
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -27,7 +27,7 @@ import (
 	"github.com/ceph/ceph-csi/internal/controller"
 	"github.com/ceph/ceph-csi/internal/controller/persistentvolume"
 	"github.com/ceph/ceph-csi/internal/liveness"
-	"github.com/ceph/ceph-csi/internal/rbd"
+	rbddriver "github.com/ceph/ceph-csi/internal/rbd/driver"
 	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/log"
 
@@ -222,7 +222,7 @@ func main() {
 	case rbdType:
 		validateCloneDepthFlag(&conf)
 		validateMaxSnaphostFlag(&conf)
-		driver := rbd.NewDriver()
+		driver := rbddriver.NewDriver()
 		driver.Run(&conf)
 
 	case cephFSType:
--- a/internal/rbd/globals.go
+++ b/internal/rbd/globals.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2021 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rbd
+
+import (
+	"fmt"
+
+	"github.com/ceph/ceph-csi/internal/journal"
+)
+
+const (
+	// volIDVersion is the version number of volume ID encoding scheme.
+	volIDVersion uint16 = 1
+)
+
+var (
+	// CSIInstanceID is the instance ID that is unique to an instance of CSI, used when sharing
+	// ceph clusters across CSI instances, to differentiate omap names per CSI instance.
+	CSIInstanceID = "default"
+
+	// volJournal and snapJournal are used to maintain RADOS based journals for CO generated
+	// VolumeName to backing RBD images.
+	volJournal  *journal.Config
+	snapJournal *journal.Config
+	// rbdHardMaxCloneDepth is the hard limit for maximum number of nested volume clones that are taken before flatten
+	// occurs.
+	rbdHardMaxCloneDepth uint
+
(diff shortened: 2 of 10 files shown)


internal/rbd/globals.go
```diff
@@ -0,0 +1,96 @@
...
+	switch name {
+	case "rbdHardMaxCloneDepth":
+		rbdHardMaxCloneDepth = value
+	case "rbdSoftMaxCloneDepth":
+		rbdSoftMaxCloneDepth = value
+	case "maxSnapshotsOnImage":
+		maxSnapshotsOnImage = value
+	case "minSnapshotsOnImageToStartFlatten":
+		minSnapshotsOnImageToStartFlatten = value
+	case "krbdFeatures":
```
Madhu-1, line 70: this looks overburdened for maintenance as we need to add two strings when we add new configurations. one is for the case and one in the caller need to send the name string. define consts for these variables?

internal/rbd/globals.go
```diff
@@ -0,0 +1,96 @@
...
+		krbdFeatures = value
+	default:
+		msg := fmt.Sprintf("BUG: can not set unknown variable %q", name)
```
Madhu-1: no need to have a new short variable declaration just do `panic(fmt.Sprintf("BUG: can not set unknown variable %q", name))`

internal/rbd/globals.go
nixpanic: That is partially intentional! I hope it prevents others from adding more variables like this.

Ideally these variables gets moved into the ControllerServer and NodeServer objects, so that they can pass it on to the calls/rbdImage that need it. In the end, these are settings that should be applied to a particular Ceph cluster or StorageClass, and can be different for other environments. Configuring these globally does not seem to be right.

internal/rbd/globals.go
nixpanic: sure, I'll modify it

cmd/cephcsi.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/ceph/ceph-csi/internal/controller"
 	"github.com/ceph/ceph-csi/internal/controller/persistentvolume"
 	"github.com/ceph/ceph-csi/internal/liveness"
-	"github.com/ceph/ceph-csi/internal/rbd"
+	rbddriver "github.com/ceph/ceph-csi/internal/rbd-driver"
```
Madhu-1: forgot to post this comment in the last review.

can this be `github.com/ceph/ceph-csi/internal/rbd/driver` IMO moving just driver .go to a new folder outside or rbd packages don't look good?

nixpanic: sure, that should be possible too

internal/rbd/rbd_util.go
```diff
@@ -1145,9 +1141,11 @@ func generateVolumeFromVolumeID(
 	return rbdVol, err
 }
 
-// genVolFromVolID generates a rbdVolume structure from the provided identifier, updating
+// GenVolFromVolID generates a rbdVolume structure from the provided identifier, updating
 // the structure with elements from on-disk image metadata as well.
-func genVolFromVolID(
+//
+// nolint // returns non-exported *rbdVolume, which is fine
+func GenVolFromVolID(
```
Madhu-1, line 1148: this is exported because you want to use this in csi-addons code?

Madhu-1 (approved):

internal/rbd/rbd_util.go
nixpanic: Yes! I want to use this in `internal/csi-addons/rbd` and with the `driver.go` in the `rbd` package the circular import is happening (not allowed by Go).

nixpanic:

[logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.20/detail/mini-e2e-helm_k8s-1.20/3495/pipeline) contain the following Ceph error:
```
Dec  9 11:44:10.564: FAIL: failed to validate owner of pvc: failed to getomapval 2021-12-09T11:44:10.542+0000 7f1469c82700 -1 monclient: handle_auth_reply_more no authorizer?
```

internal/rbd/rbd_util.go
```diff
@@ -1145,9 +1145,11 @@ func generateVolumeFromVolumeID(
 	return rbdVol, err
 }
 
-// genVolFromVolID generates a rbdVolume structure from the provided identifier, updating
+// GenVolFromVolID generates a rbdVolume structure from the provided identifier, updating
 // the structure with elements from on-disk image metadata as well.
-func genVolFromVolID(
+//
+// nolint // returns non-exported *rbdVolume, which is fine
```
Rakshith-R, line 1147: The function`GenVolFromVolID`, if imported and used, will return a variable with a undefined type if `rbdVolume` is not exported ? 

Edit: seems to be possible, found a reasonable explanation of why its allowed here https://stackoverflow.com/a/33601042

Yuggupta27 (approved):
lgtm

Rakshith-R (approved):

Madhu-1:
CI failure is due to `X Exiting due to GUEST_PROVISION: Failed to start host: creating host: create: Error creating machine: Error in driver during machine creation: IP not available after waiting: machine minikube didn't return IP after 1 minute`

PR #2699, opened 2021-12-08
rebase: update kubernetes to v1.23.0

Madhu-1:
updating go dependency to latest kubernetes released version i.e v1.23.0

closes https://github.com/ceph/ceph-csi/issues/2698
closes #2590 
depends-on: #2703

note:- other dependencies are auto-updated when I ran `go mod` commands

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

PR #2703, opened 2021-12-10
rebase: update golang version to 1.17.5

Madhu-1:
This commit updates the golang to the latest available release i.e 1.17.5

Updating golang to 1.7.x will unblock https://github.com/ceph/ceph-csi/pull/2699

### DIFF
--- a/build.env
+++ b/build.env
@@ -16,7 +16,7 @@ BASE_IMAGE=docker.io/ceph/ceph:v16
 CEPH_VERSION=octopus
 
 # standard Golang options
-GOLANG_VERSION=1.16.4
+GOLANG_VERSION=1.17.5
 GO111MODULE=on
 
 # commitlint version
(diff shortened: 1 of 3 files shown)


Madhu-1:
adding backport label as we need to support 3.4 for a while

nixpanic:
> adding backport label as we need to support 3.4 for a while

Could you explain why this is needed for 3.4? We usually don't backport features, only bugfixes...

nixpanic (changes requested):
I think the golang version in go.mod needs to be adjusted too?

nixpanic:
> I think the golang version in go.mod needs to be adjusted too?

If you want to keep that as part of #2699, that is fine too.

Madhu-1:
> I think the golang version in go.mod needs to be adjusted too?

not a blocker for now. will adjust when updating the  go.mod even with 1.16 in go.mod everything works fine.

Madhu-1:
> > I think the golang version in go.mod needs to be adjusted too?
> 
> If you want to keep that as part of #2699, that is fine too.

yes will adjust it as part of #2699

Madhu-1:
> > adding backport label as we need to support 3.4 for a while
> 
> Could you explain why this is needed for 3.4? We usually don't backport features, only bugfixes...

yes, correct this is not needed. dropping it

nixpanic (approved):

Madhu-1:
@mergifyio rebase

Yuggupta27 (approved):

PR #2716, opened 2021-12-16
rbd: add support for bigger size restore/clone PVC

### LINKED ISSUE #1438: Allow bigger size restore/clone for RBD FileMode PVCs
# Describe the bug #

CSI spec allows a user to create bigger volumes at restore/clone path. Ideally we should enable it in our driver too. This issue tracks the requirement for RBD.

### LINKED ISSUE #1566: Allow bigger size restore/clone for RBD Blockmode  PVCs

Madhu-1:
This PR allows the user to create the bigger PVC from the PVC and restore a snapshot to a bigger size PVC. and did require refactoring to avoid linter issues.

fixes: #1566
fixes: #1438

### DIFF
--- a/e2e/clone.go
+++ b/e2e/clone.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2021 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2elog "k8s.io/kubernetes/test/e2e/framework/log"
+)
+
+func validateBiggerCloneFromPVC(f *framework.Framework,
+	pvcPath,
+	appPath,
+	pvcClonePath,
+	appClonePath string) error {
+	const (
+		size    = "1Gi"
+		newSize = "2Gi"
+	)
+	pvc, err := loadPVC(pvcPath)
+	if err != nil {
+		return fmt.Errorf("failed to load PVC: %w", err)
+	}
+	label := make(map[string]string)
+	pvc.Namespace = f.UniqueName
+	pvc.Spec.Resources.Requests[v1.ResourceStorage] = resource.MustParse(size)
+	app, err := loadApp(appPath)
+	if err != nil {
+		return fmt.Errorf("failed to load app: %w", err)
+	}
+	label[appKey] = appLabel
+	app.Namespace = f.UniqueName
+	app.Labels = label
+	opt := metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", appKey, label[appKey]),
+	}
+	err = createPVCAndApp("", f, pvc, app, deployTimeout)
+	if err != nil {
+		return fmt.Errorf("failed to create pvc and application: %w", err)
+	}
+
+	pvcClone, err := loadPVC(pvcClonePath)
+	if err != nil {
+		e2elog.Failf("failed to load PVC: %v", err)
+	}
+	pvcClone.Namespace = f.UniqueName
+	pvcClone.Spec.DataSource.Name = pvc.Name
+	pvcClone.Spec.Resources.Requests[v1.ResourceStorage] = resource.MustParse(newSize)
+	appClone, err := loadApp(appClonePath)
+	if err != nil {
+		e2elog.Failf("failed to load application: %v", err)
+	}
+	appClone.Namespace = f.UniqueName
+	appClone.Labels = label
+	err = createPVCAndApp("", f, pvcClone, appClone, deployTimeout)
+	if err != nil {
+		return fmt.Errorf("failed to create pvc clone and application: %w", err)
+	}
+	err = deletePVCAndApp("", f, pvc, app)
+	if err != nil {
+		return fmt.Errorf("failed to delete pvc and application: %w", err)
+	}
+	if pvcClone.Spec.VolumeMode == nil || *pvcClone.Spec.VolumeMode == v1.PersistentVolumeFilesystem {
+		err = checkDirSize(appClone, f, &opt, newSize)
+		if err != nil {
+			return err
+		}
+	}
+
+	if pvcClone.Spec.VolumeMode != nil && *pvcClone.Spec.VolumeMode == v1.PersistentVolumeBlock {
+		err = checkDeviceSize(appClone, f, &opt, newSize)
+		if err != nil {
+			return err
+		}
+	}
+	err = deletePVCAndApp("", f, pvcClone, appClone)
+	if err != nil {
+		return fmt.Errorf("failed to delete pvc and application: %w", err)
+	}
+
+	return nil
+}
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2021 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package e2e
 
 import (
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -3444,6 +3444,77 @@ var _ = Describe("RBD", func() {
 				})
 			})
 
+			By("clone PVC to a bigger size PVC", func() {
+				By("clone PVC to bigger size encrypted PVC with VaultKMS", func() {
+					scOpts := map[string]string{
+						"encrypted":       "true",
+						"encryptionKMSID": "vault-test",
+					}
+					err := createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, scOpts, deletePolicy)
+					if err != nil {
+						e2elog.Failf("failed to create storageclass: %v", err)
+					}
+					defer func() {
+						err = deleteResource(rbdExamplePath + "storageclass.yaml")
+						if err != nil {
+							e2elog.Failf("failed to delete storageclass: %v", err)
+						}
+					}()
+
+					// validate filesystem mode PVC
+					err = validateBiggerCloneFromPVC(f,
+						pvcPath,
+						appPath,
+						pvcSmartClonePath,
+						appSmartClonePath)
+					if err != nil {
+						e2elog.Failf("failed to validate bigger size clone: %v", err)
+					}
+					// validate block mode PVC
+					err = validateBiggerCloneFromPVC(f,
+						rawPvcPath,
+						rawAppPath,
+						pvcBlockSmartClonePath,
+						appBlockSmartClonePath)
+					if err != nil {
+						e2elog.Failf("failed to validate bigger size clone: %v", err)
+					}
+				})
+
+				By("clone PVC to bigger size pvc", func() {
+					err := createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
+					if err != nil {
+						e2elog.Failf("failed to create storageclass: %v", err)
+					}
+					// validate filesystem mode PVC
+					err = validateBiggerCloneFromPVC(f,
+						pvcPath,
+						appPath,
+						pvcSmartClonePath,
+						appSmartClonePath)
+					if err != nil {
+						e2elog.Failf("failed to validate bigger size clone: %v", err)
+					}
+					// validate block mode PVC
+					err = validateBiggerCloneFromPVC(f,
+						rawPvcPath,
+						rawAppPath,
+						pvcBlockSmartClonePath,
+						appBlockSmartClonePath)
+					if err != nil {
+						e2elog.Failf("failed to validate bigger size clone: %v", err)
+					}
+				})
+
+				By("validate image deletion", func() {
+					validateRBDImageCount(f, 0, defaultRBDPool)
+					err := waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
+					if err != nil {
+						e2elog.Failf("failed to validate rbd images in pool %s trash: %v", defaultRBDPool, err)
+					}
+				})
+			})
+
 			// Make sure this should be last testcase in this file, because
 			// it deletes pool
 			By("Create a PVC and delete PVC when backend pool deleted", func() {
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -60,6 +60,8 @@ var (
 	appClonePath           = rbdExamplePath + "pod-restore.yaml"
 	appSmartClonePath      = rbdExamplePath + "pod-clone.yaml"
 	appBlockSmartClonePath = rbdExamplePath + "block-pod-clone.yaml"
+	pvcBlockRestorePath    = rbdExamplePath + "pvc-block-restore.yaml"
+	appBlockRestorePath    = rbdExamplePath + "pod-block-restore.yaml"
 	appEphemeralPath       = rbdExamplePath + "pod-ephemeral.yaml"
 	snapshotPath           = rbdExamplePath + "snapshot.yaml"
 	deployFSAppPath        = e2eTemplatesPath + "rbd-fs-deployment.yaml"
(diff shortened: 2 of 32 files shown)


internal/rbd/controllerserver.go
```diff
@@ -1397,6 +1394,33 @@ func (cs *ControllerServer) DeleteSnapshot(
...
+	rbdVol := generateVolFromSnap(rbdSnap)
+	err := rbdVol.Connect(cr)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+	defer rbdVol.Destroy()
+	// cleanup the image from trash if the error is image not found.
+	err = rbdVol.ensureImageCleanup(ctx)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to delete rbd image: %s/%s with error: %v", rbdVol.Pool, rbdVol.VolName, err)
```
nixpanic: while moving the code around, you are allowed to clean logging functions up too, just log `rbdVol` as `%q` so that the standard formatting of the image name is used (including optional rados namespace).

Same counts for the below logging of `rbdSnap`.

internal/rbd/controllerserver.go
```diff
@@ -588,6 +596,14 @@ func (cs *ControllerServer) createVolumeFromSnapshot(
 
 	log.DebugLog(ctx, "create volume %s from snapshot %s", rbdVol.RequestName, rbdSnap.RbdSnapName)
 
+	// resize the volume if the size is different
+	if rbdSnap.SizeBytes != rbdVol.VolSize {
+		err = rbdVol.resize(rbdVol.VolSize)
```
nixpanic: Reducing size is dangerous, that should not be done.

Consider placing these resize checks/actions in a helper function:

```go
err = rbdVol.Expand(rbdSnap.SizeBytes)
...

// Expand resizes the rbdImage to the given size, in case the rbdImage is smaller.
func (ri *rbdImage) Expand(size int) error {
    if ri.VolSize >= size {
         return nil
    }

    return ri.resize(size)
}
```

internal/rbd/nodeserver.go
```diff
@@ -435,24 +435,13 @@ func (ns *NodeServer) stageTransaction(
 	}
 	transaction.isMounted = true
 
-	// resize if its fileSystemType static volume.
-	if staticVol && !isBlock {
-		var ok bool
-		resizer := mount.NewResizeFs(utilexec.New())
-		ok, err = resizer.NeedResize(devicePath, stagingTargetPath)
+	// As we are suppoting the restore of a volume to a bigger size and
```
nixpanic: typo `suppoting`

internal/rbd/nodeserver.go
```diff
@@ -463,6 +452,46 @@ func (ns *NodeServer) stageTransaction(
...
+		if err != nil {
+			log.ErrorLog(ctx, "failed to resize device %s, mapper %s: %w",
+				mapperPath, mapperFile, err)
+
+			return status.Errorf(codes.Internal,
+				"failed to resize device %s, mapper %s: %v", mapperPath, mapperFile, err)
+		}
+		// Use mapper device path for fs resize
+		devicePath = mapperPath
+	}
```
nixpanic, line 461: empty lines between sections of code make things more readable, please use that to please the eyes of your fellow developers

internal/rbd/nodeserver.go
```diff
@@ -463,6 +452,46 @@ func (ns *NodeServer) stageTransaction(
...
+	resizer := mount.NewResizeFs(utilexec.New())
+	ok, err = resizer.NeedResize(devicePath, stagingTargetPath)
+	if err != nil {
+		return status.Errorf(codes.Internal,
+			"need resize check failed on devicePath %s and staingPath %s, error: %v",
+			devicePath,
+			stagingTargetPath,
+			err)
+	}
+	if ok {
```
nixpanic: In most flows we return early in case there is an error or nothing needs to be done. In such a coding style checking for `!ok` on this line and returning fits better. It also reduces the indention of the `resizer.Resize()` call, making the flow a little simpler to follow when just glancing over the code.

e2e/rbd.go
```diff
@@ -3383,6 +3385,114 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("restore snapshot to a bigger size PVC", func() {
```
nixpanic, line 3359: what is the benefit of these two nested `By()` calls?

nixpanic (changes requested):
Mostly some small coding improvements to make understanding and maintaining the code a little easier. Nothing else outstanding, as far as I am concerned.

internal/rbd/controllerserver.go
Madhu-1: the smaller size is not allowed for now in kubernetes  (CO). will send a follow-up patch to introduce this check for cephcsi (for both cephfs and rbd) to disallow that operation as it might cause some problems like data loss etc.

e2e/rbd.go
Madhu-1: Having nested `by` helps to break larger testing into smaller pieces for better readability.

internal/rbd/controllerserver.go
nixpanic: Indeed **NEVER** reduce the size of an RBD volume.

Changing quotas for CephFS would be acceptable, in case the current consumption does not hit the limit.

Madhu-1: https://github.com/ceph/ceph-csi/issues/2718

Madhu-1: will do some testing for cephfs how it works the reducing size of a bigger size subvolume and will send follow-up for this one. opened #2718 for tracking it.

internal/rbd/rbd_util.go
```diff
@@ -1053,9 +1053,33 @@ func genSnapFromSnapID(
...
+	if err != nil {
+		return fmt.Errorf("failed to update snapshot details for %q: %w", rbdSnap, err)
+	}
+
 	return err
 }
 
+// updateSnapshotDetails will copies the details from the rbdVolume to the
+// rbdSnapshot. example copying size from rbdVolume to rbdSnapshot.
+func updateSnapshotDetails(rbdSnap *rbdSnapshot) error {
```
nixpanic, line 1070: Not sure how I missed this in the preview review round...

Can `getImageInfo()` not be a function for `rbdImage` instead? That would make things much simpler.

internal/rbd/controllerserver.go
nixpanic: I still would like to see the check for resizing in a helper function. This if+resize block is repeated a lot.

Madhu-1:
@pkalever PTAL mainly for encrypted PVC expansion.

Madhu-1:
something is wrong. marking it as draft as i need to test all flows manuallly

internal/rbd/rbd_util.go
nixpanic: hmm, this has not been addressed yet?

internal/rbd/controllerserver.go
```diff
@@ -1397,6 +1394,39 @@ func (cs *ControllerServer) DeleteSnapshot(
...
+	}
+	defer rbdVol.Destroy()
+
+	err = rbdVol.openIoctx()
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+
+	// cleanup the image from trash if the error is image not found.
+	err = rbdVol.ensureImageCleanup(ctx)
```
nixpanic, line 1423: maybe make `ensureImageCleanup()` a function  of `rbdImage` so that you can call it on the `rbdSnap` directly?

internal/rbd/clone.go
```diff
@@ -181,6 +181,16 @@ func (rv *rbdVolume) createCloneFromImage(ctx context.Context, parentVol *rbdVol
 		return err
 	}
 
+	// resize the image to the requested size is greater than the current size
+	if rv.VolSize > parentVol.VolSize {
```
nixpanic: repetitive code blocks, really should be done with a helper function

nixpanic (changes requested):

internal/rbd/rbd_util.go
Madhu-1: unfortunately, this is still needed as rbdImage is created with RbdSnapName and I moved the size to rbdImage struct.

internal/rbd/clone.go
Madhu-1: should be fixed now

internal/rbd/controllerserver.go
Madhu-1: moved to a helper function

internal/rbd/controllerserver.go
Madhu-1: This needs more changes than expected as i need to update multiple functions which am not touching in this PR. will revisit when working on the codebase cleanup.

e2e/rbd.go
```diff
@@ -3338,6 +3340,199 @@ var _ = Describe("RBD", func() {
...
+				By("validate image deletion", func() {
+					validateRBDImageCount(f, 0, defaultRBDPool)
+					err := waitToRemoveImagesFromTrash(f, defaultRBDPool, deployTimeout)
+					if err != nil {
+						e2elog.Failf("failed to validate rbd images in pool %s trash: %v", defaultRBDPool, err)
+					}
+				})
+			})
+
+			By("clone PVC to a bigger size PVC", func() {
```
humblec, line 3463: Instead of adding a new test can we add one more clone case to the "exisitng" clone test ?

e2e/rbd.go
```diff
@@ -3338,6 +3340,199 @@ var _ = Describe("RBD", func() {
...
+				By("clone PVC to bigger size pvc", func() {
```
humblec: why 2 `By`s sequentially here?

humblec (changes requested):

e2e/rbd.go
Madhu-1: all clone tests need not go through the resize tests adding a separate test so that this can be reused for cephfs.

e2e/rbd.go
Madhu-1: https://github.com/ceph/ceph-csi/pull/2716#discussion_r770352888 already discussed here

internal/rbd/rbd_util.go
```diff
@@ -107,6 +107,9 @@ type rbdImage struct {
 	// identifying this rbd image
 	VolID string `json:"volID"`
 
+	// VolSize is the size of the RBD image backing this rbdVolume.
```
nixpanic: nit: this is now part of `rbdImage`, not `rbdVolume`

internal/rbd/controllerserver.go
```diff
@@ -1001,7 +1001,7 @@ func (cs *ControllerServer) CreateSnapshot(
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	rbdSnap.RbdImageName = rbdVol.RbdImageName
-	rbdSnap.SizeBytes = rbdVol.VolSize
+	rbdSnap.VolSize = rbdVol.VolSize
```
nixpanic, line 1027: this commit has a typo in the commit message (maintaing).

internal/rbd/controllerserver.go
nixpanic: ok, thanks!

internal/rbd/rbd_util.go
```diff
@@ -1703,6 +1703,16 @@ func cleanupRBDImageMetadataStash(metaDataPath string) error {
 	return nil
 }
 
+// expand checks if the requestedVolume size and the existing image size both
+// are same. If they are same, it returns nil else it resizes the image.
+func (rv *rbdVolume) expand() error {
```
nixpanic, line 1716: I was expecting to see `expand()` to have a `size` parameter... I guess this works too.

internal/rbd/nodeserver.go
```diff
@@ -427,32 +427,93 @@ func (ns *NodeServer) stageTransaction(
...
+	var ok bool
+
+	// if its a non encrypted block device we dont need any expansion
+	if isBlock && !transaction.isEncrypted {
+		return nil
+	}
+
+	resizer := mount.NewResizeFs(utilexec.New())
+
+	if transaction.isEncrypted {
```
nixpanic, line 465: a helper like `resizeEncryptedDevice()` might be good to have?

e2e/clone.go
```diff
@@ -0,0 +1,84 @@
+package e2e
```
nixpanic, line 17: probably should add a copyright/license header?

internal/rbd/rbd_util.go
```diff
@@ -127,6 +127,12 @@ type rbdImage struct {
 	RequestName string
 	NamePrefix  string
 
+	// ParentName represents the parent image name of the image.
+	ParentName string
+	// Parent Pool is the pool that contains the parent image.
+	ParentPool      string
+	ImageFeatureSet librbd.FeatureSet
```
nixpanic, line 134: why export `ImageFeatureSet`?

internal/rbd/rbd_util.go
Madhu-1: instead of passing one more input using struct fields internally

Madhu-1: instead of passing one more input using struct fields internally

e2e/clone.go
Madhu-1: we dont have copyright/license for the e2e tests

internal/rbd/rbd_util.go
Madhu-1: without this one, the golint was failing and most of the members of rbdImage are exported and it's for a reason?

e2e/rbd.go
```diff
@@ -3338,6 +3340,199 @@ var _ = Describe("RBD", func() {
...
+						rawAppPath,
+						snapshotPath,
+						pvcBlockRestorePath,
+						appBlockRestorePath)
+					if err != nil {
+						e2elog.Failf("failed to validate restore bigger size clone: %v", err)
+					}
+				})
+
+				By("restore snapshot to bigger size encrypted PVC with VaultKMS", func() {
```
humblec, line 3407: `s/VaultKMS/Vault KMS` .

examples/rbd/pod-block-restore.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
humblec, line 1: instead of adding new yaml here, can we stick or use exisitng one `pod-with-raw-block-volume` ?

examples/rbd/pvc-block-restore.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
humblec, line 1: same case here `rbd-pvc-restore`

e2e/clone.go
nixpanic: but we should?

internal/rbd/rbd_util.go
nixpanic: I hope that in the future we do not need to have the members exported, as that invites to not have a nice `rbdImage` API. It is fine for now.

nixpanic (dismissed):
No real concerns. It would be good to have the copyright/licence header in new e2e files, but that is not a blocker.

e2e/clone.go
Madhu-1: will add to all the e2e files in follow-up.

Madhu-1:
@nixpanic added Copyright for all files inside the e2e folder.

Madhu-1:
@humblec PTAL there  is a requested change from you

humblec:
Adding 3.5 milestone here for the tracker.

humblec:
mostly looks good.. few minor comments.

examples/rbd/pvc-block-restore.yaml
Madhu-1: rbd-pvc-restore is a filesystem PVC this is for BlockPVC

examples/rbd/pod-block-restore.yaml
Madhu-1: as new PVC is added keeping a different pod name

e2e/rbd.go
Madhu-1: this is a copy-paste from `By("create a PVC and bind it to an app with encrypted RBD volume with VaultKMS", func() {`

Madhu-1: `By("Validate PVC restore from vaultKMS to vaultTenantSAKMS", func() {`

humblec: ok.. then we can keep the same format

examples/rbd/pod-block-restore.yaml
humblec: yeah, thats what I was trying to avoid. As no changes in the PVC yamls. so I am checkign why we need new PVC and POD. Cant we use existing ones?

examples/rbd/pvc-block-restore.yaml
humblec: it was a typo-- , I mean we have existing FS mode and Block Mode PVC yamls which can be reused..

Madhu-1: there is no existing block PVC restore.

examples/rbd/pod-block-restore.yaml
Madhu-1: not possible as we don't have the existing yamls in place, i don't see it as a major concern/blocker for this one

Madhu-1:
@humblec PTAL

humblec:
As discussed this looks good to me .. thanks . :+1:

humblec (approved):

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio rebase

Yuggupta27 (approved):

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio rebase

PR #2733, opened 2021-12-21
rbd: reset dummy image id

Madhu-1:
dummy image rbdVolume struct is derived from the actual one rbdVolume of the volumeID sent in the EnableVolumeReplication request. and the dummy rbdVolume struct contains the image id of the actual volume because of that when we are repairing the dummy image the image is sent to trash but not deleted due to the wrong image ID. resetting
the image id will make sure the image id is fetching from the ceph cluster and the same image id will be used for manager operation.


Madhu-1:
added `skip-e2e` label as the change is in `replicationcontrollerserver.go` and this is not covered in E2E testing.

yati1998 (approved):

nixpanic (approved):

PR #2738, opened 2021-12-22
rbd: Implement FenceClusterNetwork and UnfenceClusterNetwork

Yuggupta27:
Add implementation for `FenceClusterNetwork` and `UnfenceClusterNetwork` grpc calls.

-  `FenceClusterNetwork`: allows to blocks access to a CIDR block by
creating a network fence using blocklisting.
> Ceph operation: `ceph osd blocklist add <ip>`

- `UnfenceClusterNetwork`: allows unblocking the access to a
CIDR block by removing it from the network fence.
> Ceph operation: `ceph osd blocklist rm <ip>`

Since the controller is not available yet, e2e will be added later.
Currently, all the grpcs have been manually tested and verified.

---

<details>

  progress or results

</details>


internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,149 @@
...
+
+// validateFenceClusterNetworkReq checks the sanity of FenceClusterNetworkRequest.
+func validateFenceClusterNetworkReq(req *fence.FenceClusterNetworkRequest) error {
+	if len(req.GetCidrs()) == 0 {
+		return status.Error(codes.InvalidArgument, "cidr block cannot be empty")
+	}
+
+	options := req.GetParameters()
+	if value, ok := options["clusterID"]; !ok || value == "" {
+		return status.Error(codes.InvalidArgument, "missing or empty cluster ID to provision volume from")
```
Madhu-1: error message need to be corrected.

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,149 @@
...
+	defer cr.DeleteCredentials()
+
+	nwFence, err := genNetworkFenceRequest(ctx, req.Cidrs, req.GetParameters())
+	if err != nil {
+		return nil, err
+	}
+
+	err = util.AddNetworkFence(ctx, nwFence, cr)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to fence cidr block %s: %s", nwFence.Cidr, err.Error())
```
Madhu-1: use %q for CIDR?

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,149 @@
...
+		return status.Error(codes.InvalidArgument, "cidr block cannot be empty")
+	}
+
+	options := req.GetParameters()
+	if value, ok := options["clusterID"]; !ok || value == "" {
+		return status.Error(codes.InvalidArgument, "missing or empty cluster ID to provision volume from")
+	}
+
+	return nil
+}
```
Madhu-1: This seem to be the duplicate of validateFenceClusterNetworkReq have a common function instead of duplicating it?

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to blocklist IP %s: %v", ip, stderr)
+	}
+
+	return nil
+}
+
+// AddNetworkFence blocks access for all the IPs in the IP range mentioned via the CIDR block
+// using a network fence.
+func AddNetworkFence(ctx context.Context, nwFence *NetworkFence, cr *Credentials) error {
```
Madhu-1: make this as method of NetworkFence struct

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+	// fetch the list of IPs from a CIDR block
+	hosts, err := getIPRange(nwFence.Cidr)
+	if err != nil {
+		return fmt.Errorf("failed to convert cidr block to corresponding IP range")
```
Madhu-1: add Cird to the error message.

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+		"--id", cr.ID,
+		"--keyfile=" + cr.KeyFile,
+		"-m", nwFence.Monitors,
+	}
+	cmd := []string{"osd", "blocklist", "add"}
+	cmd = append(cmd, arg...)
+
+	_, stderr, err := ExecCommand(ctx, "ceph", cmd...)
+	if err != nil {
+		return fmt.Errorf("failed to blocklist IP %s: %v", ip, stderr)
```
Madhu-1: use %q for ip and  `err` is missing when only stderr is getting returned

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// getIpRange returns a list of IPs from the IP range
+// corresponding to a CIDR block.
+func getIPRange(netw string) ([]string, error) {
```
Madhu-1: add unit testing for this one

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+	// convert string to IPNet struct
+	var infinity uint32 = 0xffffffff
+	_, ipv4Net, err := net.ParseCIDR(netw)
+	if err != nil {
+		return nil, err
+	}
+	// convert IPNet struct mask and address to uint32
+	mask := binary.BigEndian.Uint32(ipv4Net.Mask)
+	// start IP address
+	start := binary.BigEndian.Uint32(ipv4Net.IP)
```
Madhu-1: can you do something like this https://go.dev/play/p/m8TNTtygK0 to get range in CIRD

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	return cidrList
+}
+
+// GetCidr returns the CIDR as a string for further processing.
+func GetCidr(cidrs Cidrs) (string, error) {
+	cidrList := cidrs.CidrList()
+	if len(cidrList) < 1 {
+		return "", fmt.Errorf("the cidr cannot be empty")
```
Madhu-1: use error.New

internal/util/fencing.go
Madhu-1: add unit testing for all the function which does not make  any network call

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+		err = nwFence.addCephBlocklist(ctx, host, cr)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// getIpRange returns a list of IPs from the IP range
```
humblec: should be getIPRange

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+
+	_, stderr, err := ExecCommand(ctx, "ceph", cmd...)
+	if err != nil {
+		return fmt.Errorf("failed to unblock IP %s: %v", ip, stderr)
+	}
+
+	return nil
+}
+
+func RemoveNetworkFence(ctx context.Context, nwFence *NetworkFence, cr *Credentials) error {
```
humblec: add documentation.

internal/util/fencing.go
```diff
@@ -0,0 +1,142 @@
...
+	var cidrList []string
+	for _, cidr := range cidrs {
+		cidrList = append(cidrList, cidr.Cidr)
+	}
+
+	return cidrList
+}
+
+// GetCidr returns the CIDR as a string for further processing.
+func GetCidr(cidrs Cidrs) (string, error) {
```
humblec: GetCIDR make sense I believe

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, err
+	}
+
+	clusterID, err := util.GetClusterID(fenceOptions)
+	if err != nil {
+		return nil, err
+	}
+	nwFence.Monitors, _, err = util.GetMonsAndClusterID(ctx, clusterID, false)
+	if err != nil {
+		log.ErrorLog(ctx, "failed getting mons (%s)", err)
```
humblec: I believe there is aleady a error code for this exist.

internal/util/fencing.go
Yuggupta27: Added unit tests for the required files.

internal/util/fencing.go
Yuggupta27: Done :+1:

internal/csi-addons/rbd/network_fence.go
Yuggupta27: Done, Now added a common function for the same :+1:

Yuggupta27:
Tested the `FenceClusterNetwork` and `UnfenceClusterNetwork` grpcs locally using a custom grpc client for NetworkFence operation written on top of  https://github.com/ceph/ceph-csi/pull/2745.
Everything seems to be working as expected :)

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,130 @@
...
+	*fence.UnimplementedFenceControllerServer
+}
+
+func NewFenceControllerServer() *FenceControllerServer {
+	return &FenceControllerServer{}
+}
+
+func (fcs *FenceControllerServer) RegisterService(server grpc.ServiceRegistrar) {
+	fence.RegisterFenceControllerServer(server, fcs)
+}
```
Madhu-1: add comment for exported types and methods

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,130 @@
...
+
+// validateFenceClusterNetworkReq checks the sanity of FenceClusterNetworkRequest.
+func validateNetworkFenceReq(fenceClients []*fence.CIDR, options map[string]string) error {
+	if len(fenceClients) == 0 {
+		return status.Error(codes.InvalidArgument, "cidr block cannot be empty")
+	}
+
+	if value, ok := options["clusterID"]; !ok || value == "" {
+		return status.Error(codes.InvalidArgument, "missing or empty cluster ID")
```
Madhu-1: `cluster ID` to `clusterID`

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return nil, err
+	}
+	clusterID, err := util.GetClusterID(fenceOptions)
+	if err != nil {
+		return nil, err
+	}
+	nwFence.Monitors, _, err = util.GetMonsAndClusterID(ctx, clusterID, false)
+	if err != nil {
+		log.ErrorLog(ctx, "failed getting mons (%s)", err)
```
Madhu-1: `log.ErrorLog(ctx, "failed getting mons (%s)", err)` to `log.ErrorLog(ctx, "failed to get monitors for clusterID %q: %s", err)`

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,130 @@
...
+
+		return nil, err
+	}
+
+	return nwFence, nil
+}
+
+// FenceClusterNetwork blocks access to a CIDR block by creating a network fence.
```
Madhu-1, line 58: comment can be better to explain what actually happens.

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	defer cr.DeleteCredentials()
+
+	nwFence, err := genNetworkFenceRequest(ctx, req.Cidrs, req.GetParameters())
+	if err != nil {
+		return nil, err
+	}
+	err = nwFence.AddNetworkFence(ctx, cr)
```
Madhu-1: add a new line for readability

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,130 @@
...
+	defer cr.DeleteCredentials()
+
+	nwFence, err := genNetworkFenceRequest(ctx, req.Cidrs, req.GetParameters())
+	if err != nil {
+		return nil, err
+	}
+
+	err = nwFence.RemoveNetworkFence(ctx, cr)
+	if err != nil {
+		return nil, err
```
Madhu-1: return error with details same as FenceClusterNetwork

internal/csi-addons/rbd/network_fence_test.go
```diff
@@ -0,0 +1,59 @@
...
+	controller := NewFenceControllerServer()
+
+	req := &fence.FenceClusterNetworkRequest{
+		Parameters: map[string]string{},
+		Secrets:    nil,
+		Cidrs:      nil,
+	}
+
+	_, err := controller.FenceClusterNetwork(context.TODO(), req)
+	assert.Error(t, err)
```
Madhu-1, line 40: assert for the specific error messages what you are expecting.

internal/util/fencing.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/ceph/ceph-csi/internal/util/log"
+
+	"github.com/csi-addons/spec/lib/go/fence"
+	"github.com/golang/protobuf/ptypes/timestamp"
+)
+
+type NetworkFence struct {
+	Cidr      []string
+	Monitors  string
+	CreatedAt *timestamp.Timestamp
```
Madhu-1: why this is required?

internal/util/fencing.go
```diff
@@ -0,0 +1,163 @@
...
+	// At any point in this time, the IPs can be unblocked by an UnfenceClusterReq.
+	// This needs to be updated once ceph provides functionality for the same.
+	blocklistTime := "157784760"
+	cmd := []string{"osd", "blocklist", "add", ip, blocklistTime}
+	cmd = append(cmd, arg...)
+	_, _, err := ExecCommand(ctx, "ceph", cmd...)
+	if err != nil {
+		return fmt.Errorf("failed to blocklist IP %q: %w", ip, err)
+	}
+	log.DebugLog(ctx, "blocklisted ip %s", ip)
```
Madhu-1: blocklisted id %q successfully?

internal/util/fencing.go
```diff
@@ -0,0 +1,163 @@
...
+
+// AddNetworkFence blocks access for all the IPs in the IP range mentioned via the CIDR block
+// using a network fence.
+func (nwFence *NetworkFence) AddNetworkFence(ctx context.Context, cr *Credentials) error {
+	// for each cidr block, convert it into a range of IPs so as to perform blocklisting operation.
+	for _, cidr := range nwFence.Cidr {
+		// fetch the list of IPs from a CIDR block
+		hosts, err := getIPRange(cidr)
+		if err != nil {
+			return fmt.Errorf("failed to convert cidr block %s to corresponding IP range", cidr)
```
Madhu-1: return the actual `err` with %w?

internal/util/fencing.go
```diff
@@ -0,0 +1,163 @@
...
+		if ip[j] > 0 {
+			break
+		}
+	}
+}
+
+type Cidrs []*fence.CIDR
+
+// CidrList converts a CIDR struct list to a list.
+func CidrList(cidrs Cidrs) ([]string, error) {
```
Madhu-1: it should be getCIDR?

internal/util/fencing.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	if len(cidrList) < 1 {
+		return nil, errors.New("the cidr cannot be empty")
+	}
+
+	return cidrList, nil
+}
+
+// removeCephBlocklist removes an IP from ceph osd blocklist.
+func (nwFence *NetworkFence) removeCephBlocklist(ctx context.Context, ip string, cr *Credentials) error {
```
Madhu-1: keep the receiver with a short name in all places it can be called as `nf`

internal/util/fencing.go
```diff
@@ -0,0 +1,163 @@
...
+	// for each cidr block, convert it into a range of IPs so as to undo blocklisting operation.
+	for _, cidr := range nwFence.Cidr {
+		// fetch the list of IPs from a CIDR block
+		hosts, err := getIPRange(cidr)
+		if err != nil {
+			return fmt.Errorf("failed to convert cidr block %s to corresponding IP range", cidr)
+		}
+		// remove ceph blocklist for each IP in the range mentioned by the CIDR
+		for _, host := range hosts {
+			log.DebugLog(ctx, "unblocking ip %s", host)
```
Madhu-1: is this logging required?

vendor/modules.txt
```diff
@@ -91,9 +91,11 @@ github.com/container-storage-interface/spec/lib/go/csi
 # github.com/csi-addons/replication-lib-utils v0.2.0
 ## explicit; go 1.15
 github.com/csi-addons/replication-lib-utils/protosanitizer
-# github.com/csi-addons/spec v0.1.2-0.20211220083702-c779b23cf97c
+# github.com/csi-addons/spec v0.1.2-0.20211220115741-32fa508dadbe
 ## explicit
+github.com/csi-addons/spec/lib/go/fence
 github.com/csi-addons/spec/lib/go/identity
+github.com/csi-addons/spec/lib/go/reclaimspace
```
Madhu-1, line 98: why reclaim is getting added here?

internal/util/fencing.go
Yuggupta27: Nice catch, added earlier just for personal debugging. Updated now

internal/util/fencing.go
Yuggupta27: Done :+1:

internal/csi-addons/rbd/network_fence.go
Yuggupta27: Added some more context around the same.

internal/util/fencing.go
Yuggupta27: Added now :+1:

vendor/modules.txt
Yuggupta27: As discussed with @nixpanic , in go version 1.17, it seems to be adding those dependencies, which it was not added in the reclaimspace PR as it was implemented in a previous version.
As suggested by @nixpanic, added a new commit to cleanup the vendor.
Thanks!

internal/csi-addons/rbd/network_fence_test.go
Yuggupta27: I think any error other than `nil` should be considered a test fail. Thoughts?

Madhu-1: unit tests should test for the specific error message in most of the cases to make sure things are working as expected.

Madhu-1 (dismissed):

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/ceph/ceph-csi/internal/util"
+	"github.com/ceph/ceph-csi/internal/util/log"
+
+	"github.com/csi-addons/spec/lib/go/fence"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// FenceControllerServer struct of rbd CSI driver with supported methods
```
humblec, line 29: structure?  below it has mentioned as `structure` , I like `struct ` everywhere as it is the type.

internal/csi-addons/rbd/network_fence_test.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// TestUnfenceClusterNetwork is a minimal test for the
+// UnfenceClusterNetwork() procedure. During unit-testing, there is no Ceph
```
humblec: please keep same or close line wrap in these functions.

internal/util/fencing.go
```diff
@@ -0,0 +1,160 @@
...
+
+// addCephBlocklist adds an IP to ceph osd blocklist.
+func (nf *NetworkFence) addCephBlocklist(ctx context.Context, ip string, cr *Credentials) error {
+	arg := []string{
+		"--id", cr.ID,
+		"--keyfile=" + cr.KeyFile,
+		"-m", nf.Monitors,
+	}
+	// TODO: add blocklist till infinity.
+	// Currently, ceph does not provide the functionality to blocklist IPs
```
humblec: is there any tracker from ceph for this, if yes, please mention.

humblec (dismissed):
few minor comments

internal/csi-addons/rbd/network_fence.go
Yuggupta27: Thanks, Updated now

internal/util/fencing.go
Yuggupta27: I am not aware of any tracker from ceph atm. But will let @Madhu-1 confirm

Madhu-1: Nothing as of now

internal/csi-addons/rbd/identity.go
```diff
@@ -82,6 +82,12 @@ func (is *IdentityServer) GetCapabilities(
 						Type: identity.Capability_ReclaimSpace_OFFLINE,
 					},
 				},
+			}, &identity.Capability{
+				Type: &identity.Capability_NetworkFence_{
+					NetworkFence: &identity.Capability_NetworkFence{
+						Type: identity.Capability_NetworkFence_NETWORK_FENCE,
```
humblec, line 88: hmmm.. the repeatation of capability string cant be avoided ? I refer to NetworkFence_NETWORK_FENCE

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,137 @@
...
+	defer cr.DeleteCredentials()
+
+	nwFence, err := genNetworkFenceRequest(ctx, req.Cidrs, req.GetParameters())
+	if err != nil {
+		return nil, err
+	}
+
+	err = nwFence.RemoveNetworkFence(ctx, cr)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to unfence cidr block %q: %s", nwFence.Cidr, err.Error())
```
humblec: imo, in all errors its good to mention CIDR instead of cidr.

internal/util/fencing.go
```diff
@@ -0,0 +1,160 @@
...
+		"-m", nf.Monitors,
+	}
+	cmd := []string{"osd", "blocklist", "rm", ip}
+	cmd = append(cmd, arg...)
+
+	_, stdErr, err := ExecCommand(ctx, "ceph", cmd...)
+	if err != nil {
+		return fmt.Errorf("failed to unblock IP %q: %v %w", ip, stdErr, err)
+	}
+	log.DebugLog(ctx, "unblocked ip %s successfully", ip)
```
humblec: s/ip/IP

internal/rbd/driver/driver.go
```diff
@@ -218,6 +218,11 @@ func (r *Driver) setupCSIAddonsServer(conf *util.Config) error {
 		r.cas.RegisterService(rs)
 	}
 
+	if conf.IsControllerServer {
```
nixpanic: this should be included in the above `if conf.IsControllerServer`

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,109 @@
+/*
+Copyright 2021 The Ceph-CSI Authors.
```
nixpanic: may correct the year in the next revision

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,109 @@
...
+
+// genNetworkFenceRequest returns a networkFence struct object from the Network fence/unfence request.
+func genNetworkFenceRequest(
+	ctx context.Context, cidrs []*fence.CIDR, fenceOptions map[string]string) (*util.NetworkFence, error) {
+	var err error
+	nwFence := &util.NetworkFence{}
+
+	nwFence.Cidr, err = util.GetCIDR(cidrs)
+	if err != nil {
+		return nil, err
```
nixpanic: Descriptive errors are very helpful for debugging. Not always needed for local helpers in the same package, but should be added for calling functions from other packages. Consider this comment for all `return nil, err` lines.

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	clusterID, err := util.GetClusterID(fenceOptions)
+	if err != nil {
+		return nil, err
+	}
+	nwFence.Monitors, _, err = util.GetMonsAndClusterID(ctx, clusterID, false)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to get monitors for clusterID %q: %s", clusterID, err)
```
nixpanic: why log the error here, but not in other error situations?

internal/util/fencing.go
```diff
@@ -0,0 +1,160 @@
...
+	for ip := netIP.Mask(ipnet.Mask); ipnet.Contains(ip); inc(ip) {
+		hosts = append(hosts, ip.String())
+	}
+
+	return hosts, nil
+}
+
+// inc is an helper function for getIPRange for incrementing
+// IP values to return all IPs in a range.
+func inc(ip net.IP) {
```
nixpanic: This is in the `util` package, you should really use a better name for this function. You also need to describe what it exactly does, give some examples in the comment.

Does this work with IPv6?

internal/util/fencing.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	if len(cidrList) < 1 {
+		return nil, errors.New("the cidr cannot be empty")
+	}
+
+	return cidrList, nil
+}
+
+// removeCephBlocklist removes an IP from ceph osd blocklist.
+func (nf *NetworkFence) removeCephBlocklist(ctx context.Context, ip string, cr *Credentials) error {
```
nixpanic: this really should be in a different package, it is fencing related, not a generic utility

nixpanic (changes requested):
just some initial remarks, not completely reviewed yet

internal/csi-addons/rbd/network_fence.go
Yuggupta27: Thanks for the suggestion! Updated now

internal/util/fencing.go
Yuggupta27: Yes, it works with IPv6 too. 
Coming to the function `inc`, it doesn't have an identity of its own and is required by only `getIPRange()` to help with iterating IPs to find out the range depicted by the CIDR block. Additionally, updated its name to `incIP` (short for increment IP) which is a tad more descriptive.

internal/util/fencing.go
Yuggupta27: Sure, I added them to utils as these functions will be utilized by cephfs in the future too. 

Alternatively, we can have a `common` folder inside `csi-addons` folder which can accumulate `fencing.go` .
Proposed directory structure: `ceph-csi/internal/csi-addons/common/`
Thoughts?

Yuggupta27: Also, as suggested by @Madhu-1 , we can have a `fence` folder inside `internal`.
Proposed directory structure(2): `ceph-csi/internal/fence/`

internal/util/fencing.go
Yuggupta27: Updated now

internal/csi-addons/rbd/identity.go
Yuggupta27: Here NETWORK_FENCE indicates that the CSI-driver provides RPCs for NetworkFence operation. 
For unsupported, we go with `Capability_NetworkFence_UNKNOWN`.
See the [csi-addons spec](https://pkg.go.dev/github.com/csi-addons/spec@v0.1.2-0.20211220115741-32fa508dadbe/lib/go/identity#Capability_NetworkFence_NETWORK_FENCE) for more info on the same.

internal/csi-addons/rbd/network_fence.go
```diff
@@ -0,0 +1,109 @@
...
+
+// FenceClusterNetwork blocks access to a CIDR block by creating a network fence.
+// It adds the range of IPs to the osd blocklist, which helps ceph in denying access
+// to the malicious clients to prevent data corruption.
+func (fcs *FenceControllerServer) FenceClusterNetwork(
+	ctx context.Context,
+	req *fence.FenceClusterNetworkRequest) (*fence.FenceClusterNetworkResponse, error) {
+	err := validateNetworkFenceReq(req.GetCidrs(), req.Parameters)
+	if err != nil {
+		return nil, err
```
nixpanic: I am not very fond of the mixed error returns in this function. Helper functions should probably just return an `error`, and this function wraps that in a `status.Error`. For each non-`status.error` return I now have to check what the helper function returns.

internal/csi-addons/rbd/network_fence_test.go
nixpanic: You might be able to check for `status.InvalidArgument`, but not much more? Not sure if there is much value in that.

internal/util/fencing.go
nixpanic: As this really is only useful for csi-addons, can it not be `ceph-csi/internal/csi-addons/fencing/network.go`?

internal/util/fencing.go
```diff
@@ -0,0 +1,160 @@
...
+func incIP(ip net.IP) {
+	for j := len(ip) - 1; j >= 0; j-- {
+		ip[j]++
+		if ip[j] > 0 {
+			break
+		}
+	}
+}
+
+type Cidrs []*fence.CIDR
```
nixpanic: Missing a comment for this type. I thought there was a Go check that required all exported types to be commented...

internal/util/fencing.go
Yuggupta27: Sure, looks good to me! Will update

internal/util/fencing_test.go
```diff
@@ -0,0 +1,54 @@
...
+		cidr        string
+		expectedIPs []string
+	}{
+		{
+			cidr:        "10.23.56.67/32",
+			expectedIPs: []string{"10.23.56.67"},
+		},
+		{
+			cidr:        "62.76.47.12/30",
+			expectedIPs: []string{"62.76.47.12", "62.76.47.13", "62.76.47.14", "62.76.47.15"},
```
nixpanic: please include IPv6 addresses too

Yuggupta27: included IPv6 in test cases too :+1:

internal/csi-addons/rbd/network_fence.go
Yuggupta27: Thanks for pointing it out! Updated it to be more consistent now

internal/util/fencing.go
Yuggupta27: Added now :+1:

internal/csi-addons/rbd/network_fence.go
```diff
@@ -109,3 +109,31 @@ func (fcs *FenceControllerServer) FenceClusterNetwork(
...
 	return &fence.FenceClusterNetworkResponse{}, nil
 }
+
+// UnfenceClusterNetwork unblocks the access to a CIDR block by removing the network fence.
+func (fcs *FenceControllerServer) UnfenceClusterNetwork(
+	ctx context.Context,
+	req *fence.UnfenceClusterNetworkRequest) (*fence.UnfenceClusterNetworkResponse, error) {
+	err := validateNetworkFenceReq(req.GetCidrs(), req.Parameters)
+	if err != nil {
+		return nil, err
```
nixpanic: you updated the error returns in the previous commit, you probably want to do that here too

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,162 @@
...
+
+package networkfence
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+
+	"github.com/ceph/ceph-csi/internal/util"
```
nixpanic, line 22: ceph-csi places its own package at the end of the imports

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/ceph/ceph-csi/internal/util/log"
+
+	"github.com/csi-addons/spec/lib/go/fence"
+)
+
+type NetworkFence struct {
```
nixpanic, line 31: make sure all exported types have documentation

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,162 @@
...
+	Cidr     []string
+	Monitors string
+}
+
+// addCephBlocklist adds an IP to ceph osd blocklist.
+func (nf *NetworkFence) addCephBlocklist(ctx context.Context, ip string, cr *util.Credentials) error {
```
nixpanic: instead of passing credentials for all calls, add a `NewNetworkFence(*util.Credentials)` function and add a non-exported `cr *util.Credentials` to the struct.

internal/csi-addons/networkfence/fencing_test.go
```diff
@@ -0,0 +1,58 @@
...
+	tests := []struct {
+		cidr        string
+		expectedIPs []string
+	}{
+		{
+			cidr:        "10.23.56.67/32",
+			expectedIPs: []string{"10.23.56.67"},
+		},
+		{
+			cidr:        "62.76.47.12/30",
```
nixpanic: why did you pick this address? Use addresses from https://datatracker.ietf.org/doc/html/rfc1918

internal/csi-addons/networkfence/fencing_test.go
```diff
@@ -0,0 +1,58 @@
...
+			expectedIPs: []string{"62.76.47.12", "62.76.47.13", "62.76.47.14", "62.76.47.15"},
+		},
+		{
+			cidr:        "2001:4860:4860::8888/127",
```
nixpanic: Use a range from the "unique local addresses" as in https://datatracker.ietf.org/doc/html/rfc4193

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,162 @@
...
+	for ip := netIP.Mask(ipnet.Mask); ipnet.Contains(ip); incIP(ip) {
+		hosts = append(hosts, ip.String())
+	}
+
+	return hosts, nil
+}
+
+// incIP is an helper function for getIPRange() for incrementing
+// IP values to return all IPs in a range.
+func incIP(ip net.IP) {
```
nixpanic, line 130: I am still not sure how this works... IP-addresses can contain `0`, like `192.168.0.123`. Combined with a netmask like `255.255.0.0`, `192.168.1.0` should be a valid adddress, I think?

internal/csi-addons/networkfence/fencing_test.go
```diff
@@ -0,0 +1,58 @@
...
+	for _, tt := range tests {
+		ts := tt
+		t.Run(ts.cidr, func(t *testing.T) {
+			t.Parallel()
+			got, err := getIPRange(ts.cidr)
+			if err != nil {
+				t.Errorf("error occurred in getting ip range")
+			}
+			sort.Strings(got)
+			for i := 0; i < len(ts.expectedIPs); i++ {
```
nixpanic: this limits the check, there should also not be more IPs than that are expected

Yuggupta27: Good catch! Updated now

internal/csi-addons/networkfence/fencing_test.go
Yuggupta27: Updated now :+1:

internal/csi-addons/networkfence/fencing_test.go
Yuggupta27: Updated now  :+1:

internal/csi-addons/networkfence/fencing.go
Yuggupta27: Added a new function for the same as suggested :+1:

internal/csi-addons/rbd/network_fence.go
Yuggupta27: Added :+1:

internal/csi-addons/networkfence/fencing.go
Yuggupta27: On validating with other files in the repo, the import sequence seems to be correct to me. Let me know if I have missed anything?

internal/csi-addons/networkfence/fencing.go
Yuggupta27: Added now :+1:

internal/csi-addons/networkfence/fencing.go
Yuggupta27: Yup! it is a valid address and as per the code, indexes at `0` in 192.168.0.123 are returned during the pre-condition before `inc` is even called. Also, to demonstrate this, I have added the same test case in the unit test as well now. I can also add any other test case that you would like to validate.

internal/csi-addons/networkfence/fencing_test.go
```diff
@@ -0,0 +1,58 @@
...
+		cidr        string
+		expectedIPs []string
+	}{
+		{
+			cidr:        "192.168.1.0/30",
+			expectedIPs: []string{"192.168.1.0", "192.168.1.1"},
+		},
+		{
+			cidr:        "10.0.0.0/30",
+			expectedIPs: []string{"10.0.0.0", "10.0.0.1", "10.0.0.2", "10.0.0.3"},
```
nixpanic, line 35: These validations are incomplete. A /30 network should always return the same amount of IP-addresses. In the 1st case it returns only 2, in the second case it returns 4. There should be a check that **all** IP-addresses are returned, none may be missing, and not one more than expected.

internal/csi-addons/networkfence/fencing_test.go
```diff
@@ -0,0 +1,58 @@
...
+			cidr:        "fd4a:ecbc:cafd:4e49::/127",
+			expectedIPs: []string{"fd4a:ecbc:cafd:4e49::", "fd4a:ecbc:cafd:4e49::1"},
+		},
+	}
+	for _, tt := range tests {
+		ts := tt
+		t.Run(ts.cidr, func(t *testing.T) {
+			t.Parallel()
+			got, err := getIPRange(ts.cidr)
+			if err != nil {
```
nixpanic: use `assert.NoError(t, err)` to simplify things

internal/csi-addons/networkfence/fencing_test.go
```diff
@@ -0,0 +1,58 @@
...
+				t.Errorf("error occurred in getting ip range")
+			}
+			sort.Strings(got)
+			for i := 0; i != len(ts.expectedIPs); i++ {
```
nixpanic: Use [`assert.ElementsMatch()`](https://pkg.go.dev/github.com/stretchr/testify@v1.7.0/assert#ElementsMatch) instead, much simpler.

nixpanic (changes requested):
changes needed for the unit-tests, give me a confidence boost that it all works as expected :)

The commit `rbd: add network fencing utils` should be placed before the service implementation. It should be possible to build (and test) ceph-csi after each single commit. If that is not possible, `git bisect` will fail and finding the change that introduced an error becomes more difficult.

internal/csi-addons/networkfence/fencing_test.go
Yuggupta27: Oh no, that was an overlook :(  I forgot to update the CIDR to `/31` and just updated the expected range. Corrected now. Thanks for catching that :)

Yuggupta27: Also added the suggested check

Yuggupta27:
> changes needed for the unit-tests, give me a confidence boost that it all works as expected :)
> 
> The commit `rbd: add network fencing utils` should be placed before the service implementation. It should be possible to build (and test) ceph-csi after each single commit. If that is not possible, `git bisect` will fail and finding the change that introduced an error becomes more difficult.

Thanks for the suggestions, the test looks much cleaner now! Also rearranged commits as required :+1:

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,191 @@
...
+		if ip[j] > 0 {
+			break
+		}
+	}
+}
+
+// Cidrs is a list of CIDR structs.
+type Cidrs []*fence.CIDR
+
+// GetCIDR converts a CIDR struct list to a list.
```
humblec, line 142: This is confusing. List is not  a type per se.

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,191 @@
...
+func GetCIDR(cidrs Cidrs) ([]string, error) {
+	var cidrList []string
+	for _, cidr := range cidrs {
+		cidrList = append(cidrList, cidr.Cidr)
+	}
+	if len(cidrList) < 1 {
```
humblec, line 148: cant we check the input first , instead of appending then validating the size?

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,191 @@
...
+		"-m", nf.Monitors,
+	}
+	cmd := []string{"osd", "blocklist", "rm", ip}
+	cmd = append(cmd, arg...)
+
+	_, stdErr, err := util.ExecCommand(ctx, "ceph", cmd...)
+	if err != nil {
+		return fmt.Errorf("failed to unblock IP %q: %v %w", ip, stdErr, err)
+	}
+	log.DebugLog(ctx, "unblocked IP %s successfully", ip)
```
humblec: IP:

internal/csi-addons/networkfence/fencing.go
Yuggupta27: We initially validate the input here https://github.com/ceph/ceph-csi/pull/2738/files?authenticity_token=Yo7Z0aohQClJB_sNzHyUqNlP3Wi2tnWhknWiR27KveWX0VuLWFixeQ5VFmkogs-5tljCXW9ylI2_vj7WoHXR1A&file-filters%5B%5D=.go&file-filters%5B%5D=.txt#diff-829c4b5ccf3d560ea770c3c8a49458fb3639ab035fc64bd630a5380d7e6712beR47

internal/csi-addons/networkfence/fencing.go
Yuggupta27: The [spec](https://github.com/csi-addons/spec/blob/main/lib/go/fence/fence.pb.go) has defined the CIDR as a separate message 
```
type CIDR struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// CIDR block
	Cidr string `protobuf:"bytes,1,opt,name=cidr,proto3" json:"cidr,omitempty"`
}
```

Here CIDR is just one of the components which we require.

The Message that CEPH-CSI receives is a list of CIDR structs(shown above); from which, for our processing we need only `CIDR.Cidr` field.
So that function helps in achieving that step and gives a just a list of CIDR strings.

internal/csi-addons/networkfence/fencing.go
Yuggupta27: I feel it looks good as then `successfully` will look weird after `:`
e.g: `unblocked IP: 10.2.4.6 successfully`

What do you think?

humblec: may be %q goes better

internal/csi-addons/networkfence/fencing_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	for _, tt := range tests {
+		ts := tt
+		t.Run(ts.cidr, func(t *testing.T) {
+			t.Parallel()
+			got, err := getIPRange(ts.cidr)
+			assert.NoError(t, err)
+
+			// validate if number of IPs in the range is same as expected, if not, fail.
+			if len(got) == len(ts.expectedIPs) {
```
nixpanic: No need to do this check. If you use `assert.ElementsMatch()` it will be checked too. A useful error message is displayed when there is an inconsistency. Use `ElementsMatchf()` if you want to provide an additional error message.

nixpanic (dismissed):
just a nit that can be used to clean things up in the unit-tests even more.

The added capability could be part of the last commit `rbd: add NetworkFence operation`, so that a calling service would not expect fencing to be available until the service is registered (for test-building).

internal/csi-addons/networkfence/fencing_test.go
Yuggupta27: Added now :+1:

internal/csi-addons/networkfence/fencing.go
Yuggupta27: Updated :+1:

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,191 @@
...
+type NetworkFence struct {
+	Cidr     []string
+	Monitors string
+	cr       *util.Credentials
+}
+
+// NewNetworkFence returns a networkFence struct object from the Network fence/unfence request.
+func NewNetworkFence(
+	ctx context.Context, cr *util.Credentials, cidrs []*fence.CIDR,
+	fenceOptions map[string]string) (*NetworkFence, error) {
```
humblec, line 42: @Yuggupta27 I think this should be multiline as mentioned in the development/coding guide.

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,191 @@
...
+	var err error
```
humblec, line 43: you can remove this and use err from line 42 with short declaration ?

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,191 @@
...
+		"--keyfile=" + nf.cr.KeyFile,
+		"-m", nf.Monitors,
+	}
+	// TODO: add blocklist till infinity.
+	// Currently, ceph does not provide the functionality to blocklist IPs
+	// for infinite time. As a workaround, add a blocklist for 5 YEARS to
+	// represent infinity from ceph-csi side.
+	// At any point in this time, the IPs can be unblocked by an UnfenceClusterReq.
+	// This needs to be updated once ceph provides functionality for the same.
+	blocklistTime := "157784760"
```
humblec: I would like to make this as a constant

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	log.DebugLog(ctx, "blocklisted IP %q successfully", ip)
+
+	return nil
+}
+
+// AddNetworkFence blocks access for all the IPs in the IP range mentioned via the CIDR block
+// using a network fence.
+func (nf *NetworkFence) AddNetworkFence(ctx context.Context) error {
+	// for each cidr block, convert it into a range of IPs so as to perform blocklisting operation.
```
humblec: cidr/CIDR

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, err
+	}
+	for ip := netIP.Mask(ipnet.Mask); ipnet.Contains(ip); incIP(ip) {
+		hosts = append(hosts, ip.String())
+	}
+
+	return hosts, nil
+}
+
+// incIP is an helper function for getIPRange() for incrementing
```
humblec, line 128: It can be mentioned as  generic helper too.

internal/csi-addons/rbd/network_fence_test.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// TestUnfenceClusterNetwork is a minimal test for the UnfenceClusterNetwork()
+// procedure. During unit-testing, there is no Ceph cluster available, so actual
+// operations can not be performed.
+func TestUnfenceClusterNetwork(t *testing.T) {
+	t.Parallel()
+
+	controller := NewFenceControllerServer()
+
```
humblec, line 49: may be we can remove this new line.

internal/csi-addons/rbd/network_fence_test.go
```diff
@@ -0,0 +1,59 @@
...
+	_, err := controller.FenceClusterNetwork(context.TODO(), req)
+	assert.Error(t, err)
+}
+
+// TestUnfenceClusterNetwork is a minimal test for the UnfenceClusterNetwork()
+// procedure. During unit-testing, there is no Ceph cluster available, so actual
+// operations can not be performed.
+func TestUnfenceClusterNetwork(t *testing.T) {
+	t.Parallel()
+
```
humblec: no need for new line here I feel

internal/csi-addons/networkfence/fencing.go
humblec: Thanks.

internal/csi-addons/networkfence/fencing.go
Yuggupta27: Done :+1:

internal/csi-addons/networkfence/fencing.go
Yuggupta27: Updated now :+1:

internal/csi-addons/networkfence/fencing.go
Yuggupta27: Cant do that... as `nwFence.CIDR is already declared and not being re-declared. It will throw error

internal/csi-addons/networkfence/fencing.go
Yuggupta27: This function is completely dependent on getCIDR as it is just an iterator. Thus, I have mentioned the same there.

internal/csi-addons/rbd/network_fence_test.go
Yuggupta27: Since it is starting a new server, the function looks cluttered without it. Removed new lines at other places :+1:

internal/csi-addons/networkfence/fencing_test.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	for _, tt := range tests {
+		ts := tt
+		t.Run(ts.cidr, func(t *testing.T) {
+			t.Parallel()
+			got, err := getIPRange(ts.cidr)
+			assert.NoError(t, err)
+
+			// validate if number of IPs in the range is same as expected, if not, fail.
+			assert.Equal(t, ts.expectedIPs, got)
```
nixpanic: nit: I guess this could throw false errors, in case the order of the lists changes. https://pkg.go.dev/github.com/stretchr/testify@v1.7.0/assert#ElementsMatch would prevent that.

nixpanic (dismissed):

internal/csi-addons/networkfence/fencing_test.go
Yuggupta27: Thanks! Updated

nixpanic (dismissed):

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,194 @@
...
+
+// AddNetworkFence blocks access for all the IPs in the IP range mentioned via the CIDR block
+// using a network fence.
+func (nf *NetworkFence) AddNetworkFence(ctx context.Context) error {
+	// for each CIDR block, convert it into a range of IPs so as to perform blocklisting operation.
+	for _, cidr := range nf.Cidr {
+		// fetch the list of IPs from a CIDR block
+		hosts, err := getIPRange(cidr)
+		if err != nil {
+			return fmt.Errorf("failed to convert cidr block %s to corresponding IP range: %w", cidr, err)
```
humblec: nit: cidr/CIDR

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	log.DebugLog(ctx, "unblocked IP %q successfully", ip)
+
+	return nil
+}
+
+// RemoveNetworkFence unblocks access for all the IPs in the IP range mentioned via the CIDR block
+// using a network fence.
+func (nf *NetworkFence) RemoveNetworkFence(ctx context.Context) error {
+	// for each cidr block, convert it into a range of IPs so as to undo blocklisting operation.
```
humblec: nit: cidr/CIDR

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,194 @@
...
+	for _, cidr := range nf.Cidr {
+		// fetch the list of IPs from a CIDR block
+		hosts, err := getIPRange(cidr)
+		if err != nil {
+			return fmt.Errorf("failed to convert cidr block %s to corresponding IP range", cidr)
```
humblec: same here

internal/csi-addons/networkfence/fencing.go
```diff
@@ -0,0 +1,194 @@
...
+type Cidrs []*fence.CIDR
+
+// GetCIDR converts a CIDR struct list to a list.
+func GetCIDR(cidrs Cidrs) ([]string, error) {
+	var cidrList []string
+	for _, cidr := range cidrs {
+		cidrList = append(cidrList, cidr.Cidr)
+	}
+	if len(cidrList) < 1 {
+		return nil, errors.New("the cidr cannot be empty")
```
humblec: cidr/CIDR

Yuggupta27:
ci/centos/mini-e2e/k8s-1.22

Madhu-1 (approved):

nixpanic (approved):

humblec (approved):

PR #2740, opened 2021-12-22
deploy: add CSI-Addons endpoint

nixpanic:
The --csi-addons-endpoint= option has been added recently, but was not
configured in the deployment files yet. The socket was incorrectly
created as `/csi-addons.sock`, by correcting the URL to the socket, the
socket now gets created as `/csi/csi-addons.sock` in the same directory
as other sockets.

If an endpoint is a UNIX Domain Socket, the format needs to be
`unix:///path/to/socket`. The `unix://` URL format allows an
authentication provider to be added directly after the `//`. If there is
no authentication provider needed, the field can remain empty. After the
authetication provider, the full path needs to be specified, starting
with a `/`. This means that URLs to a UDS should start with `unix:///`
in normal cases.

---

<details>

  progress or results

</details>


deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
```diff
@@ -118,6 +118,7 @@ spec:
             - "--type=rbd"
             - "--controllerserver=true"
             - "--endpoint=$(CSI_ENDPOINT)"
+            - "--csi-addons-endpoint=$(CSI_ADDONS_ENDPOINT)"
```
Madhu-1, line 121: Need same changes in helm chart also

cmd/cephcsi.go
```diff
@@ -129,7 +129,7 @@ func init() {
 	flag.BoolVar(&conf.EnableProfiling, "enableprofiling", false, "enable go profiling")
 
 	// CSI-Addons configuration
-	flag.StringVar(&conf.CSIAddonsEndpoint, "csi-addons-endpoint", "unix://tmp/csi-addons.sock", "CSI-Addons endpoint")
+	flag.StringVar(&conf.CSIAddonsEndpoint, "csi-addons-endpoint", "unix:///csi/csi-addons.sock", "CSI-Addons endpoint")
```
Madhu-1: not related to this PR but looks like this new parameter missing in doc https://github.com/ceph/ceph-csi/blob/devel/docs/deploy-rbd.md#configuration

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
yati1998: @nixpanic and @Madhu-1 do you wish to add the changes for the helm chart in the same PR or different? Otherwise looks good to me.

nixpanic: yes, I know! Just being lazy and updated it only here for easy copy/pasting in my test environment.

cmd/cephcsi.go
nixpanic: right! I'll add it there  too

deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml
nixpanic: Done now.

cmd/cephcsi.go
nixpanic: this has been addressed too.

humblec (approved):

yati1998 (commented):
LGTM, will wait for @Madhu-1 to review it.

Madhu-1 (approved):

yati1998 (approved):

Rakshith-R (approved):

PR #2747, opened 2022-01-04
rbd: change the configmap of HPCS/KP key names to reflect the IBM string

humblec:
considering IBM has different crypto services (ex: SKLM) in place, its
good to keep the configmap key names with below format

`IBM_KP_...` instead of `KP_..`

so that in future, if we add more crypto services from IBM we can keep
similar schema specific to that specific service from IBM.

Ex: `IBM_SKLM_...`

### DIFF
--- a/internal/kms/keyprotect.go
+++ b/internal/kms/keyprotect.go
@@ -33,7 +33,7 @@ const (
 
 	// keyProtectMetadataDefaultSecretsName is the default name of the Kubernetes Secret
 	// that contains the credentials to access the Key Protect KMS. The name of
-	// the Secret can be configured by setting the `KMS_SECRET_NAME`
+	// the Secret can be configured by setting the `IBM_KP_SECRET_NAME`
 	// option.
 	//
 	// #nosec:G101, value not credential, just references token.
@@ -43,17 +43,16 @@ const (
 	// the credentials to access the Key ProtectKMS.
 	//
 	// #nosec:G101, no hardcoded secret, this is a configuration key.
-	keyProtectSecretNameKey = "KMS_SECRET_NAME"
-	keyProtectRegionKey     = "KP_REGION"
+	keyProtectSecretNameKey = "IBM_KP_SECRET_NAME"
+	keyProtectRegionKey     = "IBM_KP_REGION"
 
-	keyProtectServiceInstanceID = "KP_SERVICE_INSTANCE_ID"
+	keyProtectServiceInstanceID = "IBM_KP_SERVICE_INSTANCE_ID"
 	// The following options are part of the Kubernetes Secrets.
 	// #nosec:G101, no hardcoded secrets, only configuration keys.
-	keyProtectServiceAPIKey   = "KP_SERVICE_API_KEY"
-	KeyProtectCustomerRootKey = "KP_CUSTOMER_ROOT_KEY"
-
-	keyProtectSessionToken = "KP_SESSION_TOKEN"
-	keyProtectCRK          = "KP_CRK_ARN"
+	keyProtectServiceAPIKey   = "IBM_KP_SERVICE_API_KEY"
+	KeyProtectCustomerRootKey = "IBM_KP_CUSTOMER_ROOT_KEY"
+	keyProtectSessionToken    = "IBM_KP_SESSION_TOKEN" //nolint:gosec // only configuration key
+	keyProtectCRK             = "IBM_KP_CRK_ARN"
 )
 
 var _ = RegisterProvider(Provider{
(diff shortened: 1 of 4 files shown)


Madhu-1 (dismissed):

internal/kms/keyprotect.go
```diff
@@ -44,16 +44,16 @@ const (
...
 	// The following options are part of the Kubernetes Secrets.
 	// #nosec:G101, no hardcoded secrets, only configuration keys.
-	keyProtectServiceAPIKey   = "KP_SERVICE_API_KEY"
-	KeyProtectCustomerRootKey = "KP_CUSTOMER_ROOT_KEY"
+	keyProtectServiceAPIKey   = "IBM_KP_SERVICE_API_KEY"
+	KeyProtectCustomerRootKey = "IBM_KP_CUSTOMER_ROOT_KEY"
 
-	keyProtectSessionToken = "KP_SESSION_TOKEN"
-	keyProtectCRK          = "KP_CRK_ARN"
+	keyProtectSessionToken = "IBM_KP_SESSION_TOKEN" //nolint:gosec
```
pkalever: why do we need nolint here?

humblec: if not, we get a linter error.

pkalever: @humblec 
Well! We didn't see any linter issues before(as per the diff), I'm might be definitely missing something here.
Could you please paste what the linter error looks like?

nixpanic: Indeed, I wonder about that too. Why were there earlier no lint errors, or, why now?

humblec:
```
[/home/hchiramm/go/src/github.com/ceph/ceph-csi/internal/kms/keyprotect.go:55] - G101: Potential hardcoded credentials (Confidence: LOW, Severity: HIGH)
 keyProtectSessionToken = "IBM_KP_SESSION_TOKEN"
``` 
this is what I got locally, may be the gosec or linter version difference.. I can try to remove the tag from the PR and see CI here picks it up.

build.env
```diff
@@ -23,7 +23,7 @@ GO111MODULE=on
 COMMITLINT_VERSION=latest
 
 # static checks and linters
-GOLANGCI_VERSION=v1.39.0
+GOLANGCI_VERSION=v1.42.0
```
Madhu-1: please revert this change. we are going to get a lot of linter errors for this one.

humblec: yeah, done. I added it to check the CI here picks the errors I get locally, reverted this commit

Madhu-1 (dismissed):

humblec:
https://github.com/ceph/ceph-csi/runs/4701490805?check_suite_focus=true
``` 
level=info msg="[runner] Issues before processing: 485, after processing: 1"
internal/kms/keyprotect.go:55:2: G101: Potential hardcoded credentials (gosec)
	keyProtectSessionToken = "IBM_KP_SESSION_TOKEN"
	
```

This is the golang CI failure here when I removed the tag :) 
@pkalever @nixpanic

internal/kms/keyprotect.go
pkalever: I think gosec is reporting error when the consts contain "Secret" or "token" naming, see https://github.com/securego/gosec/issues/295

humblec: isnt it the same thing solved by the tag?

pkalever: Yes, but now we know the reason. Hence can make better comments for the readers.

humblec: Well, I added that tag when I got the error as mentioned in the previous comment. :) , also the main query was that, why it was not giving error previously ( based on the diff)?, which is not answered by above gosec validation comment.  The close answer I could think of here is it is because of golang version difference.

humblec:
@Madhu-1 based on above discussions, I had to make it failing in CI with the last refresh and had to bring back the first version of the PR. please bring back the approval.

Madhu-1:
> @Madhu-1 based on above discussions, I had to make it failing in CI with the last refresh and had to bring back the first version of the PR. please bring back the approval.

not sure it's needed. it could have been tested locally and posted as a comment.

Madhu-1 (approved):

humblec:
> > @Madhu-1 based on above discussions, I had to make it failing in CI with the last refresh and had to bring back the first version of the PR. please bring back the approval.
> 
> not sure it's needed. it could have been tested locally and posted as a comment.

 It was tested locally as mentioned here https://github.com/ceph/ceph-csi/pull/2747#discussion_r778015905 , then had to make sure it matches the CI too.. so the attempt.

yati1998 (approved):

internal/kms/keyprotect.go
pkalever: @humblec 
It's easily understandable the new gosec is giving some new error, but why is the question for me! the comment "// only configuration key" doesn't make any sense to me.

IMHO comment should be more like, why we are using the constant name having "Secret" in it, ensuring we are aware of it, as the error is more caution oriented!!!

PR #2748, opened 2022-01-04
e2e: make recreateCSIRBDPods() function to a generic one

humblec:
This commit make recreateCSIRBDPods function to be a general one
so that it can be consumed by more clients.

Updates https://github.com/ceph/ceph-csi/issues/2509

### DIFF
--- a/e2e/pod.go
+++ b/e2e/pod.go
@@ -395,3 +395,22 @@ func getKernelVersionFromDaemonset(f *framework.Framework, ns, dsn, cn string) (
 
 	return kernelRelease, nil
 }
+
+// recreateCSIPods delete the daemonset and deployment pods based on the selectors passed in.
+func recreateCSIPods(f *framework.Framework, podLabels, daemonsetName, deploymentName string) error {
+	err := deletePodWithLabel(podLabels, cephCSINamespace, false)
+	if err != nil {
+		return fmt.Errorf("failed to delete pods with labels (%s): %w", podLabels, err)
+	}
+	// wait for csi pods to come up
+	err = waitForDaemonSets(daemonsetName, cephCSINamespace, f.ClientSet, deployTimeout)
+	if err != nil {
+		return fmt.Errorf("timeout waiting for daemonset pods: %w", err)
+	}
+	err = waitForDeploymentComplete(f.ClientSet, deploymentName, cephCSINamespace, deployTimeout)
+	if err != nil {
+		return fmt.Errorf("timeout waiting for deployment to be in running state: %w", err)
+	}
+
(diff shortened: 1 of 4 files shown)


e2e/pod.go
```diff
@@ -395,3 +395,22 @@ func getKernelVersionFromDaemonset(f *framework.Framework, ns, dsn, cn string) (
 
 	return kernelRelease, nil
 }
+
+func recreateCSIPods(f *framework.Framework, podLabel, daemonsetName, deploymentName string) error {
```
nixpanic: Comments for the parameters might be useful. Also a description containing the behaviour (what is expected if the pods already/partially exist?) would be welcome.

e2e/pod.go
```diff
@@ -395,3 +395,22 @@ func getKernelVersionFromDaemonset(f *framework.Framework, ns, dsn, cn string) (
...
+	err := deletePodWithLabel(podLabel,
+		cephCSINamespace, false)
```
nixpanic: this can be on the previous line

e2e/pod.go
```diff
@@ -395,3 +395,22 @@ func getKernelVersionFromDaemonset(f *framework.Framework, ns, dsn, cn string) (
...
+	if err != nil {
+		return fmt.Errorf("failed to delete pods with labels: %w", err)
```
nixpanic: `with labels`? which labels? it is good to include those details in the error.

e2e/utils.go
```diff
@@ -60,6 +60,8 @@ const (
 	appLabel = "write-data-in-pod"
 
 	noError = ""
+	// labels of rbd pods.
```
nixpanic: Mention that this is used for listing/deleting the pods. The `rbdPodLabel` name suggests it is a single label (which could be set on a Pod), not a selector.

nixpanic (changes requested):
When cleaning up things, it is better to improve things all-over, otherwise there is little benefit if the code shuffling.

e2e/pod.go
humblec: Done

e2e/utils.go
humblec: Done

e2e/pod.go
humblec: Done

e2e/pod.go
humblec: Done. it issue `delete` operation on the POD and rest is same as what CO does today.

nixpanic (approved):

yati1998 (approved):

PR #2749, opened 2022-01-05
[wip ]e2e: remove podUnknown state case condition in waitForPodInRunningState

humblec:
podUnknown state is a long time deprecated value and we dont need it.

-- snip--
// PodUnknown means that for some reason the state of the pod could not be obtained, typically due
// to an error in communicating with the host of the pod.
// Deprecated: It isn't being set since 2015 (74da3b14b0c0f658b3bb8d2def5094686d0e9095)

--/snip--


nixpanic:
This as `[wip]` in the subject, assuming it is not ready for review/merging, marking as **draft**

Issue #2725, opened 2021-12-20
Reverify all the KMS struct exports and unexport if not needed


humblec:
# Describe the bug #

At present all the KMS structs ( SecretKMS,AWSMetadataKMS...etc) where EncryptDEK(), DecryptDEK, ..etc methods are implemented are exposed. However considering all KMS structs reside in the same package (ie kms) we may not need to expose this struct, 
This issue track the revalidation and once confirmed all the KMS structs can be unexported. 

Ex:

```
// SecretsKMS is default KMS implementation that means no KMS is in use.
type SecretsKMS struct {
	IntegratedDEK

	passphrase string
}

```

Additional Ref# https://github.com/ceph/ceph-csi/pull/2723#discussion_r772288543

Cc @nixpanic

PR #2723, opened 2021-12-20
IBM HPCS/Key Protect KMS integration 

humblec:
rbd: Implement Key Protect KMS integration for Ceph CSI
    
This commit adds the support for HPCS/Key Protect IBM KMS service
to Ceph CSI service. EncryptDEK() and DecryptDEK() of RBD volumes are
done with the help of key protect KMS server by wrapping and unwrapping
the DEK and by using the DEKStoreMetadata.
    

Fixes: # https://github.com/ceph/ceph-csi/issues/2585

What is covered: 
- [x] KMS integration interface
- [x] Documentation about the design consideration
- [x] Dependency update
- [x] Additionally this has been tested against Key Protect service hosted in an IBM cloud and below is the integration test result logs:
```
   

--------------- Application POD Mounting ------------
...
/pvc/name":"testkp","csi.storage.k8s.io/pvc/namespace":"openshift-storage","encrypted":"true","encryptionKMSID":"kp-metadata-test","imageFeatures":"layering","imageName":"csi-vol-948fe374-5f0d-11ec-a42d-0a580a84001c","journalPool":"ocs-storagecluster-cephblockpool","pool":"ocs-storagecluster-cephblockpool","storage.kubernetes.io/csiProvisionerIdentity":"1639727227436-8081-openshift-storage.rbd.csi.ceph.com","thickProvision":"false"},"volume_id":"0001-0011-...
.......
I1217 07:59:58.080408  257478 crypto.go:199] ID: 6 Req-ID: 0001-0011-openshift-storage-0000000000000001-948fe374-5f0d-11ec-a42d-0a580a84001c Encrypting device "/dev/rbd0"	 with LUKS
I1217 08:00:04.518933  257478 crypto.go:258] ID: 6 Req-ID: 0001-0011-openshift-storage-0000000000000001-948fe374-5f0d-11ec-a42d-0a580a84001c "/dev/mapper/luks-rbd-0001-0011-openshift-storage-0000000000000001-948fe374-5f0d-11ec-a42d-0a580a84001c" is not an active LUKS device (an error (exit status 4) occurred while running cryptsetup args: [status luks-rbd-0001-0011-openshift-storage-0000000000000001-948fe374-5f0d-11ec-a42d-0a580a84001c]):
[...]

PR #2750, opened 2022-01-06
rbd: unexport existing KMS structs for vault, AWS and KeyProtect

humblec:
At present the KMS structs are exported and ideally we should be
able to work without exporting the same.

Updates https://github.com/ceph/ceph-csi/issues/2725 

### DIFF
--- a/internal/kms/kms.go
+++ b/internal/kms/kms.go
@@ -359,20 +359,20 @@ type DEKStore interface {
 	RemoveDEK(volumeID string) error
 }
 
-// IntegratedDEK is a DEKStore that can not be configured. Either the KMS does
+// integratedDEK is a DEKStore that can not be configured. Either the KMS does
 // not use a DEK, or the DEK is stored in the KMS without additional
 // configuration options.
-type IntegratedDEK struct{}
+type integratedDEK struct{}
 
-func (i IntegratedDEK) RequiresDEKStore() DEKStoreType {
+func (i integratedDEK) RequiresDEKStore() DEKStoreType {
 	return DEKStoreIntegrated
 }
 
-func (i IntegratedDEK) EncryptDEK(volumeID, plainDEK string) (string, error) {
+func (i integratedDEK) EncryptDEK(volumeID, plainDEK string) (string, error) {
 	return plainDEK, nil
 }
 
-func (i IntegratedDEK) DecryptDEK(volumeID, encyptedDEK string) (string, error) {
+func (i integratedDEK) DecryptDEK(volumeID, encyptedDEK string) (string, error) {
 	return encyptedDEK, nil
 }
 
--- a/internal/kms/secretskms.go
+++ b/internal/kms/secretskms.go
@@ -50,7 +50,7 @@ const (
 
 // SecretsKMS is default KMS implementation that means no KMS is in use.
 type SecretsKMS struct {
-	IntegratedDEK
+	integratedDEK
 
 	passphrase string
 }
--- a/internal/kms/secretskms.go
+++ b/internal/kms/secretskms.go
@@ -48,8 +48,8 @@ const (
 	metadataSecretNamespaceKey = "secretNamespace"
 )
 
-// SecretsKMS is default KMS implementation that means no KMS is in use.
-type SecretsKMS struct {
+// secretsKMS is default KMS implementation that means no KMS is in use.
+type secretsKMS struct {
 	integratedDEK
 
 	passphrase string
@@ -60,7 +60,7 @@ var _ = RegisterProvider(Provider{
 	Initializer: newSecretsKMS,
 })
 
-// newSecretsKMS initializes a SecretsKMS that uses the passphrase from the
+// newSecretsKMS initializes a secretsKMS that uses the passphrase from the
 // secret that is configured for the StorageClass. This KMS provider uses a
 // single (LUKS) passhprase for all volumes.
 func newSecretsKMS(args ProviderInitArgs) (EncryptionKMS, error) {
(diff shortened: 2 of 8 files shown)


nixpanic:
This is just a cleanup, there is no requirement in including this in 3.5.0. We should focus on other PRs that add functionality or fix bugs instead. The 3.5.0 release is already a day too late.

humblec:
Rebased to devel.. ptal @nixpanic

humblec:
@nixpanic @Rakshith-R can you take a look at this ?

internal/kms/secretskms.go
```diff
@@ -49,7 +49,7 @@ const (
 )
 
 // SecretsKMS is default KMS implementation that means no KMS is in use.
```
Rakshith-R:
```suggestion
// secretsKMS is default KMS implementation that means no KMS is in use.
```
change the caps in struct comment too, other places too

internal/kms/secretskms.go
```diff
@@ -69,43 +69,43 @@ func newSecretsKMS(args ProviderInitArgs) (EncryptionKMS, error) {
...
-func (kms SecretsKMS) RemoveDEK(key string) error {
+// secretsKMS.
+func (kms secretsKMS) RemoveDEK(key string) error {
 	return nil
 }
 
-// SecretsMetadataKMS is a KMS based on the SecretsKMS, but stores the
+// SecretsMetadataKMS is a KMS based on the secretsKMS, but stores the
 // Data-Encryption-Key (DEK) in the metadata of the volume.
 type SecretsMetadataKMS struct {
```
Rakshith-R:
```suggestion
type secretsMetadataKMS struct {
```

Unexport this struct too ?

Rakshith-R (commented):
unexport `type DEKStoreType string`, `type IntegratedDEK struct{}` and `type DEKStore interface` too ?

internal/kms/secretskms.go
humblec: @Rakshith-R done.. ptal,

internal/kms/secretskms.go
humblec: @Rakshith-R done

humblec:
> unexport `type DEKStoreType string`, `type IntegratedDEK struct{}` and `type DEKStore interface` too ?

@Rakshith-R done other both except the interface, which need by outside package?

Rakshith-R:
> > unexport `type DEKStoreType string`, `type IntegratedDEK struct{}` and `type DEKStore interface` too ?
> 
> @Rakshith-R done other both except the interface, which need by outside package?

:+1:, there seems to be a stray file `internal/kms/.keyprotect.go.swp` @humblec

humblec:
> > > unexport `type DEKStoreType string`, `type IntegratedDEK struct{}` and `type DEKStore interface` too ?
> > 
> > 
> > @Rakshith-R done other both except the interface, which need by outside package?
> 
> +1, there seems to be a stray file `internal/kms/.keyprotect.go.swp` @humblec

ah. multiple window switching .. removed it .. thanks @Rakshith-R

internal/kms/kms.go
```diff
@@ -316,11 +316,11 @@ func GetDefaultKMS(secrets map[string]string) (EncryptionKMS, error) {
 type EncryptionKMS interface {
 	Destroy()
 
-	// RequiresDEKStore returns the DEKStoreType that is needed to be
+	// RequiresDEKStore returns the dekStoreType that is needed to be
```
nixpanic: This interface is public, so the return type should be public too?

internal/kms/kms.go
```diff
@@ -334,18 +334,18 @@ type EncryptionKMS interface {
 	DecryptDEK(volumeID, encyptedDEK string) (string, error)
 }
 
-// DEKStoreType describes what DEKStore needs to be configured when using a
+// dekStoreType describes what DEKStore needs to be configured when using a
 // particular KMS. A KMS might support different DEKStores depending on its
 // configuration.
-type DEKStoreType string
+type dekStoreType string
```
nixpanic: this is part of the public API. I think it is better to export it, just like the consts  below.

nixpanic (changes requested):

internal/kms/kms.go
humblec: ok.. lets keep it exported. thanks

internal/kms/kms.go
humblec: better to be public

nixpanic:
> At present the KMS structs are exported and ideally we should be able to work without exporting the same.
> 
> Updates #2750

I am wondering if there is a goal that you want to reach with this. The linked #2750 refers to this PR, and that is not helpful. Depending on the goal, `IntegratedKMS` may need to stay exported.

humblec:
> > At present the KMS structs are exported and ideally we should be able to work without exporting the same.
> > Updates #2750
> 
> I am wondering if there is a goal that you want to reach with this. The linked #2750 refers to this PR, and that is not helpful. Depending on the goal, `IntegratedKMS` may need to stay exported.

https://github.com/ceph/ceph-csi/issues/2725 is the correct issue link which came out of a review comment   (https://github.com/ceph/ceph-csi/pull/2723#discussion_r772288543). Corrected the link now, may be we will keep `IntegratedDEK` unexported and open it up if we really need? that sound fine to me.

humblec:
@nixpanic @Rakshith-R ptal.. thanks .

nixpanic (approved):

yati1998 (approved):

PR #2760, opened 2022-01-11
cephfs: log error message if clone fails

Madhu-1:
During CreateVolume from snapshot/volume, it's difficult to identify if the clone is failed and a new clone is created. In case
of clone failure logging the error message for better debugging.

### DIFF
--- a/internal/cephfs/core/fsjournal.go
+++ b/internal/cephfs/core/fsjournal.go
@@ -124,6 +124,11 @@ func CheckVolExists(ctx context.Context,
 			return nil, cerrors.ErrClonePending
 		}
 		if cloneState == cephFSCloneFailed {
+			log.ErrorLog(ctx,
+				"clone failed, deleting subvolume clone. vol=%s, subvol=%s subvolgroup=%s",
+				volOptions.FsName,
+				vid.FsSubvolName,
+				volOptions.SubvolumeGroup)
 			err = volOptions.PurgeVolume(ctx, fsutil.VolumeID(vid.FsSubvolName), true)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


internal/cephfs/core/fsjournal.go
```diff
@@ -124,6 +124,11 @@ func CheckVolExists(ctx context.Context,
 			return nil, cerrors.ErrClonePending
 		}
 		if cloneState == cephFSCloneFailed {
```
pkalever, line 126: Will be better if we log the `cloneState` outside this block of code?

Madhu-1: it might not be helpful as the clone takes time to complete with that we will see more in-progress logs. and we are already returning pending state to the user and it's logged as GRPC error.

internal/cephfs/core/fsjournal.go
```diff
@@ -124,6 +124,11 @@ func CheckVolExists(ctx context.Context,
...
+			log.ErrorLog(ctx,
```
yati1998, line 127: "clone failed, deleting subvolume clone. vol=%s, subvol=%s subvolgroup=%s". Maybe specifying the names will be more clearer?

Madhu-1: i can do that but we dont follow any specific format for cephfs logging.

yati1998: This seems more clear to be but yeah it's fine for me either way.

yati1998 (approved):

nixpanic (approved):

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio refresh

Madhu-1:
@Mergifyio refresh

Issue #2718, opened 2021-12-16
Add check to block smaller size PVC creation from snapshot/clone


Madhu-1:
Even though this operation is blocked from the kubernetes csi-external-provisioner. as cephcsi should be independent of CO it is good to validate the size at cephcsi. some details at https://github.com/ceph/ceph-csi/pull/2716#discussion_r770329308

PR #2766, opened 2022-01-12
rbd: reject small size restore/clone from snapshot/volume

Madhu-1:
As per the CSI standard, the size is an optional parameter, as we are allowing the clone/restore to a bigger size today we need to block the clone/restore to a smaller size as its a have side effects like data corruption, etc.
    
Note:- Even though this check is present in the kubernetes sidecar as CSI is CO independent adding the check here.
    
updates: #2718
    

nixpanic (approved):

Rakshith-R (approved):

Madhu-1:
@mergifyio rebase

PR #2788, opened 2022-01-17
cephfs: disallow creating small volumes from snapshot/volume

### LINKED ISSUE #2718: Add check to block smaller size PVC creation from snapshot/clone

Madhu-1:
as per the CSI standard, the size is an optional parameter, as we are allowing the clone to a bigger size today we need to block the clone to a smaller size as its a have side effects like data corruption, etc.

Note:- Even though this check is present in the kubernetes sidecar as CSI is CO independent adding the check here.

fixes: #2718

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -148,6 +148,35 @@ func checkContentSource(
 	return nil, nil, nil, status.Errorf(codes.InvalidArgument, "not a proper volume source %v", volumeSource)
 }
 
+// checkValidCreateVolumeRequest checks if the request is valid
+// CreateVolumeRequest by inspecting the request parameters.
+func checkValidCreateVolumeRequest(
+	vol,
+	parentVol *store.VolumeOptions,
+	pvID *store.VolumeIdentifier,
+	sID *store.SnapshotIdentifier) error {
+	switch {
+	case pvID != nil:
+		if vol.Size < parentVol.Size {
+			return fmt.Errorf(
+				"cannot clone from volume %s: volume size %d is smaller than source volume size %d",
+				pvID.VolumeID,
+				parentVol.Size,
+				vol.Size)
+		}
+	case sID != nil:
+		if vol.Size < parentVol.Size {
+			return fmt.Errorf(
+				"cannot restore from snapshot %s: volume size %d is smaller than source volume size %d",
+				sID.SnapshotID,
+				parentVol.Size,
+				vol.Size)
+		}
+	}
+
+	return nil
+}
+
 // CreateVolume creates a reservation and the volume in backend, if it is not already present.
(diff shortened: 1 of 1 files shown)


Madhu-1:
@mergifyio rebase

Madhu-1:
`Jan 17 11:36:11.447: FAIL: failed to create storageclass: etcdserver: request timed out`

Madhu-1:
@ceph/ceph-csi-maintainers @ceph/contributors need a review if anyone has free time.

Madhu-1:
@mergifyio rebase

humblec:
Actually this will never happen from Kube CO and also on Openshift. In our documentation we have mentioned, only those as tested or Known to Work https://github.com/ceph/ceph-csi#known-to-work-co-platforms .. The  confusion this could bring here is, if we started to think about other COs , then many areas have to be looked upon in various  code paths..

Madhu-1:
> Actually this will never happen from Kube CO and also on Openshift. 
In our documentation we have mentioned, only those as tested or Known to Work https://github.com/ceph/ceph-csi#known-to-work-co-platforms .. The confusion this could bring here is, if we started to think about other COs , then many areas have to be looked upon in various code paths..

we never mentioned that this will won't work or we should only fix it as per the kubernetes/OCP requirement. most of the time we will try to align with the standard CSI specification.

Madhu-1:
@nixpanic do you have any comments on this one?

internal/cephfs/controllerserver.go
```diff
@@ -199,6 +234,11 @@ func (cs *ControllerServer) CreateVolume(
 		defer parentVol.Destroy()
 	}
 
+	err = checkValidCreateVolumeRequest(volOptions, parentVol, pvID, sID)
+	if err != nil {
+		return nil, err
```
nixpanic: This just returns the error, instead of converting it to a gRPC error. The `checkValidCreateVolumeRequest()` function takes care of that, but it would be cleaner to do it here.

Ideally top-level gRPC functions are the only ones that return gRPC errors, all other functions should just return plain errors so that they can be re-used outside of the direct gRPC flow.

nixpanic (changes requested):

nixpanic (dismissed):

yati1998 (dismissed):
LGTM

yati1998 (approved):

nixpanic (approved):

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio refresh

PR #2420, opened 2021-08-18
ci: tag ceph image with quay.io

Madhu-1:
added if-else check to check the image starting with quay.io. If yes, pull and tag the image with quay.io

currently, https://github.com/ceph/ceph-csi/pull/2419 is failing with the below error message 

```
+ ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@n26.dusty.ci.centos.org 'podman pull --authfile=~/.podman-auth.json registry-ceph-csi.apps.ocp.ci.centos.org/quay.io/ceph/ceph:v16 && podman tag registry-ceph-csi.apps.ocp.ci.centos.org/quay.io/ceph/ceph:v16 quay.io/ceph/ceph:v16 docker.io/quay.io/ceph/ceph:v16'

Warning: Permanently added 'n26.dusty.ci.centos.org,172.19.2.90' (ECDSA) to the list of known hosts.

Trying to pull registry-ceph-csi.apps.ocp.ci.centos.org/quay.io/ceph/ceph:v16...

Error: Error initializing source docker://registry-ceph-csi.apps.ocp.ci.centos.org/quay.io/ceph/ceph:v16: Error reading manifest v16 in registry-ceph-csi.apps.ocp.ci.centos.org/quay.io/ceph/ceph: manifest unknown: manifest unknown

script returned exit code 125
```

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

**Note:-**  just changed in one place. If the changes look okay i will make changes in other places also

PR #2797, opened 2022-01-19
ci: tag ceph image with quay.io

Madhu-1:
added if-else check to check the image starting with quay.io. If yes, pull and tag the image with quay.io

i cannot reopen https://github.com/ceph/ceph-csi/pull/2420, this is a replacement PR.

updates #2797

### DIFF
--- a/mirror/images.txt
+++ b/mirror/images.txt
@@ -6,6 +6,9 @@
 # format: <full-image-spec> <short-name>
 #
 
+# ceph-csi devel
+quay.io/ceph/ceph:v16		ceph/ceph:v16
+
 # ceph-csi v3.4
 docker.io/ceph/ceph:v16		ceph/ceph:v16
 docker.io/rook/ceph:v1.6.2	rook/ceph:v1.6.2
--- a/containerized-tests.groovy
+++ b/containerized-tests.groovy
@@ -137,10 +137,19 @@ node('cico-workspace') {
 						script: 'ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} \'source /opt/build/go/src/github.com/ceph/ceph-csi/build.env && echo ${BASE_IMAGE}\'',
 						returnStdout: true
 					).trim()
-					def d_io_regex = ~"^docker.io/"
 
-					// base_image is like ceph/ceph:v15 or docker.io/ceph/ceph:v15, strip "docker.io/"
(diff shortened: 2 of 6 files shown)


Madhu-1:
@nixpanic to answer your  questions at  #2420

>Might be easier to use short names (without quay.io or docker.io) in build.env and other placed?

If we do this the image will get pulled from the default docker hub registry?

>Things might be easier, and only a change to the images.txt could work.

as the retagging is happed with docker.io in CI jobs we might not able able to find image with quay.io and re-pull happens and it takes times

mirror/images.txt
```diff
@@ -6,6 +6,9 @@
 # format: <full-image-spec> <short-name>
 #
 
+# ceph-csi devel
```
nixpanic, line 9: this is not needed, quay.io does not have pull limits like docker

remove the docker.io/ceph/ceph entry below.

nixpanic (changes requested):
I think this does more than needed. When using images from quay. there is no need to mirror them. Use `quay.io/ceph/ceph:v16` everywhere, instead of `ceph/ceph:v16`.

How does Rook refer to the images? If that is the abbreviated name, only then you may need to do the re-tagging with `podman_pull`.

Madhu-1:
> I think this does more than needed. When using images from the quay. there is no need to mirror them. Use `quay.io/ceph/ceph:v16` everywhere, instead of `ceph/ceph:v16`.
> 

Yes, it can be removed as we can pull from the quay.io always but for that, we need to update the Rook version in build.env once rook is updated remove changes in the centos branch to pre-pull the ceph image from the internal registry?

> How does Rook refer to the images? If that is the abbreviated name, only then you may need to do the re-tagging with `podman_pull`.

in Rook 1.6.2 which we use in cephcsi now refers to it as `ceph/ceph:v16` https://github.com/rook/rook/blob/v1.6.2/cluster/examples/kubernetes/ceph/cluster-test.yaml#L29 with Rook 1.7 this is referred to as `quay.io/ceph/ceph:v16`.

nixpanic:
If all the required versions are on quay.io, pre-pull them from there, and re-tag them as `ceph/ceph:v16`. The references in `images.txt` can be removed.

We _could_ mirror images locally, that would reduce internet bandwidth, and possibly speed things up a little. But that is not required and can be done in a different PR.

Madhu-1:
closing this one as things are working fine. will revisit when required.

PR #2812, opened 2022-01-24
rbd: change the keyprotect metadata name to `ibmkeyprotect`

humblec:
To be consistent with other components and also to explictly
state it belong to `ibm keyprotect` service introducing this
change

https://github.com/rook/rook/pull/9615/files#diff-2f97d4e20aa3611cfe1fec4777fb663abcf866d94ed58d33220a82a7229e25f4R53

### DIFF
--- a/internal/kms/keyprotect.go
+++ b/internal/kms/keyprotect.go
@@ -23,14 +23,15 @@ import (
 	"fmt"
 
 	"github.com/ceph/ceph-csi/internal/util/k8s"
+	"github.com/ceph/ceph-csi/internal/util/log"
 
 	kp "github.com/IBM/keyprotect-go-client"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	kmsTypeKeyProtectMetadata = "kp-metadata"
-
+	kmsTypeKeyProtectMetadata    = "ibmkeyprotect"
+	kmsTypeKeyProtectMetadataOld = "kp-metadata"
 	// keyProtectMetadataDefaultSecretsName is the default name of the Kubernetes Secret
 	// that contains the credentials to access the Key Protect KMS. The name of
 	// the Secret can be configured by setting the `IBM_KP_SECRET_NAME`
@@ -61,6 +62,21 @@ var _ = RegisterProvider(Provider{
 	UniqueID:    kmsTypeKeyProtectMetadata,
 	Initializer: initKeyProtectKMS,
 })
+
+// RegisterProvider for kmsTypeKeyProtectMetadataOld is kept here for backward compatibility.
+var _ = RegisterProvider(Provider{
+	UniqueID:    kmsTypeKeyProtectMetadataOld,
+	Initializer: initKeyProtectKMSOld,
+})
+
+// initKeyProtectKMSOld is the wrapper with a warning log.
+func initKeyProtectKMSOld(args ProviderInitArgs) (EncryptionKMS, error) {
+	log.WarningLogMsg("%q is deprecated provider for IBM key Protect,"+
+		"use new provider name %q in the configuration, proceeding with %q",
+		kmsTypeKeyProtectMetadataOld, kmsTypeKeyProtectMetadata, kmsTypeKeyProtectMetadata)
+
+	return initKeyProtectKMS(args)
+}
 
 // KeyProtectKMS store the KMS connection information retrieved from the kms configmap.
 type KeyProtectKMS struct {
--- a/examples/kms/vault/csi-kms-connection-details.yaml
+++ b/examples/kms/vault/csi-kms-connection-details.yaml
@@ -60,9 +60,9 @@ data:
       "IBM_KP_SECRET_NAME": "ceph-csi-aws-credentials",
       "AWS_REGION": "us-west-2"
     }
-  kp-metadata-test: |-
+  ibmkeyprotect-test: |-
     {
-      "KMS_PROVIDER": "kp-metadata",
+      "KMS_PROVIDER": "ibmkeyprotect",
       "IBM_KP_SECRET_NAME": "ceph-csi-kp-credentials",
       "IBM_KP_SERVICE_INSTANCE_ID": "7abef064-01dd-4237-9ea5-8b3890970be3",
       "IBM_KP_BASE_URL": "https://us-south.kms.cloud.ibm.com",
--- a/examples/kms/vault/kms-config.yaml
+++ b/examples/kms/vault/kms-config.yaml
@@ -91,8 +91,8 @@ data:
         "encryptionKMSType": "metadata",
         "secretName": "storage-encryption-secret"
       },
-      "kp-metadata-test": {
-        "encryptionKMSType": "kp-metadata",
+      "ibmkeyprotect-test": {
+        "encryptionKMSType": "ibmkeyprotect",
         "secretName": "ceph-csi-kp-credentials",
         "keyProtectRegionKey": "us-south-2",
         "keyProtectServiceInstanceID": "7abef064-01dd-4237-9ea5-8b3890970be3"


internal/kms/keyprotect.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	kmsTypeKeyProtectMetadata = "kp-metadata"
+	kmsTypeKeyProtectMetadata = "ibmkeyprotect"
```
nixpanic: This breaks deployments in 3.5 that use this.

nixpanic (changes requested):
Update the documentation too? And make sure that it is backported to 3.5 before anyone uses it.

humblec:
> Update the documentation too? And make sure that it is backported to 3.5 before anyone uses it.

@nixpanic this was not mentioend in the doc, so we are good :)

internal/kms/keyprotect.go
humblec: sure, added backporting label. @nixpanic

Rakshith-R:
Can we keep the old name and add the new proposed name alongside the old one?


Both the names will point to the same init function and that should be okay.

humblec:
> Can we keep the old name and add the new proposed name alongside the old one?
> 
> Both the names will point to the same init function and that should be okay.

@Rakshith-R can be done to be utter sure we are not breaking.. @nixpanic wdyt?

nixpanic:
> > Can we keep the old name and add the new proposed name alongside the old one?
> > Both the names will point to the same init function and that should be okay.
> 
> @Rakshith-R can be done to be utter sure we are not breaking.. @nixpanic wdyt?

There is a release with this KMS provider, so better not break it. Maybe add a warning and recommend using the new name in the configuration?

humblec:
> > > Can we keep the old name and add the new proposed name alongside the old one?
> > > Both the names will point to the same init function and that should be okay.
> > 
> > 
> > @Rakshith-R can be done to be utter sure we are not breaking.. @nixpanic wdyt?
> 
> There is a release with this KMS provider, so better not break it. Maybe add a warning and recommend using the new name in the configuration?

Sure.. Done.. ptal @nixpanic .. thanks.

internal/kms/keyprotect.go
```diff
@@ -22,15 +22,17 @@ import (
 	"errors"
 	"fmt"
 
+	"k8s.io/klog/v2"
+
 	"github.com/ceph/ceph-csi/internal/util/k8s"
 
 	kp "github.com/IBM/keyprotect-go-client"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
Rakshith-R:
```suggestion
	"github.com/ceph/ceph-csi/internal/util/k8s"

	"k8s.io/klog/v2"
	kp "github.com/IBM/keyprotect-go-client"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

Rakshith-R (changes requested):
Just a tiny nit of reordering import lines.
Other things look good.

internal/kms/keyprotect.go
humblec: Sure, addressed it.. Thanks @Rakshith-R

Rakshith-R (dismissed):

internal/kms/keyprotect.go
```diff
@@ -62,6 +63,20 @@ var _ = RegisterProvider(Provider{
...
 
+// RegisterProvider for kmsTypeKeyProtectMetadataOld is kept here for backward compatibility.
+var _ = RegisterProvider(Provider{
+	UniqueID:    kmsTypeKeyProtectMetadataOld,
+	Initializer: initKeyProtectKMSOld,
+})
+
+// initKeyProtectKMSOld is the wrapper with a warning log.
+func initKeyProtectKMSOld(args ProviderInitArgs) (EncryptionKMS, error) {
+	klog.Warning("%s is deprecated provider for IBM key Protect, use new provider name %s in the configuration",
```
Madhu-1: why use klog here? use internal logs what cephcsi implements.

internal/kms/keyprotect.go
```diff
@@ -23,14 +23,15 @@ import (
...
+	"k8s.io/klog/v2"
 	kp "github.com/IBM/keyprotect-go-client"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	kmsTypeKeyProtectMetadata = "kp-metadata"
-
+	kmsTypeKeyProtectMetadata    = "ibmkeyprotect"
+	kmsTypeKeyProtectMetadataOld = "kp-metadata"
```
Madhu-1, line 34: this needs to be taken care  at `getSecrets` function or else the return message in getSecrets seems wrong.

internal/kms/keyprotect.go
```diff
@@ -62,6 +63,20 @@ var _ = RegisterProvider(Provider{
 	Initializer: initKeyProtectKMS,
 })
 
+// RegisterProvider for kmsTypeKeyProtectMetadataOld is kept here for backward compatibility.
```
Madhu-1, line 66: open an issue for taking this one to remove it in 3.6

Madhu-1 (changes requested):

internal/kms/keyprotect.go
humblec: This is addressed in a different way now , ptal @Madhu-1

internal/kms/keyprotect.go
humblec: https://github.com/ceph/ceph-csi/issues/2816

internal/kms/keyprotect.go
```diff
@@ -62,6 +63,21 @@ var _ = RegisterProvider(Provider{
...
+var _ = RegisterProvider(Provider{
+	UniqueID:    kmsTypeKeyProtectMetadataOld,
+	Initializer: initKeyProtectKMSOld,
+})
+
+// initKeyProtectKMSOld is the wrapper with a warning log.
+func initKeyProtectKMSOld(args ProviderInitArgs) (EncryptionKMS, error) {
+	log.WarningLog(context.TODO(), "%s is deprecated provider for IBM key Protect,"+
```
Madhu-1: use `WarningLogMsg` as no actual context is passed here.

Madhu-1 (commented):
small nit

internal/kms/keyprotect.go
humblec: @Madhu-1 done. thanks. ptal.

internal/kms/keyprotect.go
```diff
@@ -62,6 +63,21 @@ var _ = RegisterProvider(Provider{
...
 
+// RegisterProvider for kmsTypeKeyProtectMetadataOld is kept here for backward compatibility.
+var _ = RegisterProvider(Provider{
+	UniqueID:    kmsTypeKeyProtectMetadataOld,
+	Initializer: initKeyProtectKMSOld,
+})
+
+// initKeyProtectKMSOld is the wrapper with a warning log.
+func initKeyProtectKMSOld(args ProviderInitArgs) (EncryptionKMS, error) {
+	log.WarningLogMsg("%s is deprecated provider for IBM key Protect,"+
```
Madhu-1: instead of `%s` `%q` could have been better for logging.

Madhu-1 (dismissed):

internal/kms/keyprotect.go
```diff
@@ -78,14 +94,20 @@ type KeyProtectKMS struct {
...
 	sessionToken      string
 	crk               string
+	providerName      string
 }
 
 func initKeyProtectKMS(args ProviderInitArgs) (EncryptionKMS, error) {
 	kms := &KeyProtectKMS{
 		namespace: args.Namespace,
 	}
+	// storing providername.
```
nixpanic: This does not explain why the provider name needs to be stored. A comment about that would be helpful (in the struct too).

If this is only for the logging in `getSecrets()`, it would be sufficient to be clearer in the warning log and mention `"ibmkeyprotect"` is used.

humblec: @nixpanic , I have kept only logging model now and mentioned in the log message as we are proceeding with new provider name.. should be good for now. ptal.

nixpanic (approved):

Rakshith-R (approved):

humblec:
@Mergifyio refresh

humblec:
@Mergifyio refresh

Issue #852, opened 2020-03-04
Refractor codebase


Madhu-1:
currently, it's difficult to write a Unit testing for the current code base, if we move all cephfs and rbd and rados function to a new interface and methods it will allow us to mock the things are helping us to write the unit testing for the current codebase

the functions which can be moved to the different interfaces  and to seperate files also example
```
GetOMapValue()
getImageInfo()
createImage()
deleteImage()
etc...
```

Madhu-1:
* Move files to a specific folder based on the operations
* Break larger functions in the smaller functions
* Implement interfaces for all the functions/methods which does go-ceph/CLI calls
  - This helps in adding more unit testings
* Change error implementations
  - Internal functions should not return GRPC error codes/messages

Madhu-1:
Reopening as planning to work on this one.

Madhu-1:
@humblec @nixpanic @pkalever @Yuggupta27 @yati1998 @Rakshith-R please add your suggestions here so that we can consider for refactoring

nixpanic:
I would really like to see separate packages for:

* k8s usage: helper functions that do not return k8s objects directly (1st step in moving to a k8s sidecar)
* provisioner and node-plugin: separated from `internal/rbd` and `internal/cephfs` API
* (future) `cmd/tool`: using ` internal/rbd` and `internal/cephfs` API

nixpanic:
And of course, the KMS API should move to its own `internal/kms` package too. With each provider placed in `internal/kms/<provider>`.

humblec:
@Madhu-1 as I brought up in earlier meetings and had some discussions on, a couple of items I have been working on this regard are :
[1] Refactoring operational framework  and [2] e2e refactoring:

The first should help us to address transactional chain for the operation and help us to improve each transaction we have in place..etc, the design https://hackmd.io/TvrswhUUQDGwVk1vFLY7Pw
The second should help us to have more flexible e2e tests..  #2178 

so apart from the points raised above, I would like to do these 2 things  which should help us to improve the overall code base :+1:

Madhu-1:
Yes, agree am not tracking E2E on this one. so won't makes any changes to E2E as part of this issue. Updated the Issue title

humblec:
> Yes, agree am not tracking E2E on this one. so won't makes any changes to E2E as part of this issue. Updated the Issue title

Ok. If thats the case operational framwork improvement is something which I have been working on which could be part of this larger issue.

[...]

PR #2815, opened 2022-01-25
cleanup: use interface in filesystem.go

Madhu-1:
Currently, we are using methods and all the methods make a network call to fetch details from the ceph clusters, it's difficult to write test cases for these functions, if we move to the interfaces we can make use of mock to write unit testing for the caller functions.

updates https://github.com/ceph/ceph-csi/issues/852

Making an initial PR with only one file to get the review on the idea, the whole change to the internal core functions are available at https://github.com/Madhu-1/ceph-csi/commit/378a3a45196af019fcae794a33562888a9e9a9dc.

### DIFF
--- a/internal/cephfs/core/filesystem.go
+++ b/internal/cephfs/core/filesystem.go
@@ -25,76 +25,102 @@ import (
 	"github.com/ceph/ceph-csi/internal/util/log"
 )
 
-func (vo *VolumeOptions) getFscID(ctx context.Context) (int64, error) {
-	fsa, err := vo.conn.GetFSAdmin()
+// FileSystem is the interface that holds the signature of filesystem methods
+// that interacts with CephFS filesystem API's.
+type FileSystem interface {
+	// GetFscID returns the ID of the filesystem with the given name.
+	GetFscID(context.Context, string) (int64, error)
+	// GetMetadataPool returns the metadata pool name of the filesystem with the given name.
+	GetMetadataPool(context.Context, string) (string, error)
+	// GetFsName returns the name of the filesystem with the given ID.
+	GetFsName(context.Context, int64) (string, error)
+}
+
+// fileSystem is the implementation of FileSystem interface.
+type fileSystem struct {
+	conn *util.ClusterConnection
+}
+
+// NewFileSystem returns a new instance of fileSystem.
+func NewFileSystem(conn *util.ClusterConnection) FileSystem {
+	return &fileSystem{
+		conn: conn,
+	}
+}
+
+// GetFscID returns the ID of the filesystem with the given name.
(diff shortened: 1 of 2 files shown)


internal/cephfs/core/filesystem.go
```diff
@@ -25,76 +25,99 @@ import (
...
+}
+
+// NewFileSystem returns a new instance of fileSystem.
+func NewFileSystem(conn *util.ClusterConnection) FileSystem {
+	return &fileSystem{
+		conn: conn,
+	}
+}
+
+// GetFscID returns the ID of the filesystem with the given name.
```
nixpanic, line 51: The documentation for functions from an interface can be placed at the interface. In case there are multiple implementation of the interface, there is no need to repeat the generic description.

Details for the implementation specifics of this function should of course be documented here, instead of at the interface level.

Madhu-1: Done.

internal/cephfs/core/filesystem.go
```diff
@@ -25,76 +25,102 @@ import (
...
+	// GetFscID returns the ID of the filesystem with the given name.
+	GetFscID(context.Context, string) (int64, error)
+	// GetMetadataPool returns the metadata pool name of the filesystem with the given name.
+	GetMetadataPool(context.Context, string) (string, error)
+	// GetFsName returns the name of the filesystem with the given ID.
+	GetFsName(context.Context, int64) (string, error)
+}
+
+// fileSystem is the implementation of FileSystem interface.
+type fileSystem struct {
```
nixpanic, line 40: Thinking about this a little more, could the `VolumeOptions` type not just implement this interface? I think the `VolumeOptions` struct is misnamed anyway, it probably should be a struct called `CephFSVolume` or something...

Madhu-1: the final idea is to have independent struck like  Volume, Snapshot, etc, and move the duplicate or just use the core structs in VolumeOptions how we do for rbdImage and rbdVolume. as there is no dependency with VolumeOptions and Filesystem just making use of empty struct here.

nixpanic (approved):

Madhu-1:
@mergifyio rebase

Madhu-1:

`Jan 27 05:07:58.935: INFO: Unexpected error while creating namespace: Post "https://192.168.39.20:8443/api/v1/namespaces": dial tcp 192.168.39.20:8443: connect: connection refused`

yati1998 (approved):
LGTM

Madhu-1:
@mergifyio refresh

Madhu-1:
@mergifyio refresh

PR #2819, opened 2022-01-25
ci: use sidecar versions from the build.env 

humblec:
The sidecar images in minikube deployment will be fetched from build.env and used/validated accordingly.

Testing ( with all versions set):
```
validating sidecar controller configuration
```
Testing without a sidecar version set:
```
validating sidecar controller configuration
CSI_RESIZER_VERSION version is empty, make sure build.env has set this sidecar version

```

### DIFF
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -124,6 +124,21 @@ function validate_container_cmd() {
     fi
 }
 
+# validate csi sidecar image version
+function validate_sidecar() {
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+# shellcheck disable=SC1091
+    source "${SCRIPT_DIR}/../build.env"
+
+    sidecars=(CSI_ATTACHER_VERSION CSI_SNAPSHOTTER_VERSION CSI_PROVISIONER_VERSION CSI_RESIZER_VERSION CSI_NODE_DRIVER_REGISTRAR_VERSION)
+    for sidecar in "${sidecars[@]}"; do
+        if [[ -z "${!sidecar}" ]]; then
+	   echo "${sidecar}" version is empty, make sure build.env has set this sidecar version
+	   exit 1
+    fi
+done
+}
+
 # Storage providers and the default storage class is not needed for Ceph-CSI
 # testing. In order to reduce resources and potential conflicts between storage
 # plugins, disable them.
@@ -160,13 +175,6 @@ if [[ "${VM_DRIVER}" == "kvm2" ]] || [[ "${VM_DRIVER}" == "hyperkit" ]]; then
 else
     DISK_CONFIG=""
 fi
-
-#configure csi sidecar version
-CSI_ATTACHER_VERSION=${CSI_ATTACHER_VERSION:-"v3.2.1"}
-CSI_SNAPSHOTTER_VERSION=${CSI_SNAPSHOTTER_VERSION:-"v4.1.1"}
-CSI_PROVISIONER_VERSION=${CSI_PROVISIONER_VERSION:-"v2.2.2"}
-CSI_RESIZER_VERSION=${CSI_RESIZER_VERSION:-"v1.2.0"}
-CSI_NODE_DRIVER_REGISTRAR_VERSION=${CSI_NODE_DRIVER_REGISTRAR_VERSION:-"v2.2.0"}
 
 # configure csi image version
 CSI_IMAGE_VERSION=${CSI_IMAGE_VERSION:-"canary"}
@@ -290,6 +298,8 @@ cephcsi)
     copy_image_to_cluster "${CEPHCSI_IMAGE_REPO}"/cephcsi:"${CSI_IMAGE_VERSION}" "${CEPHCSI_IMAGE_REPO}"/cephcsi:"${CSI_IMAGE_VERSION}"
     ;;
 k8s-sidecar)
+    echo "validating sidecar's image version"
+    validate_sidecar
     echo "copying the kubernetes sidecar images"
     copy_image_to_cluster "${K8S_IMAGE_REPO}/csi-attacher:${CSI_ATTACHER_VERSION}" "${K8S_IMAGE_REPO}/csi-attacher:${CSI_ATTACHER_VERSION}"
     copy_image_to_cluster "${K8S_IMAGE_REPO}/csi-snapshotter:${CSI_SNAPSHOTTER_VERSION}" "${K8S_IMAGE_REPO}/csi-snapshotter:${CSI_SNAPSHOTTER_VERSION}"


Madhu-1 (commented):
This is supposed to be overridden from the build.env, if we start updating sidecar is both places it's a double work. We should stick to updating it in only one place.

humblec:
> This is supposed to be overridden from the build.env, if we start updating sidecar is both places it's a double work. We should stick to updating it in only one place.

Yeah, however its is possible to use this `scripts/minikube.sh` locally  , where the version update make sense. 

As we keep the versions in the script its indeed better to keep the latest versions in place. Also sidecar updates are not going to happen very frequently.. Mostly once with a kube release.  iow, there is no point to keep very old versions in the script either..

Madhu-1 (commented):
Do source build.env in minikube.sh and use it? and this avoids duplicate works if we maintain image versios in multiple places

humblec:
See sourcing can be used, but my point is simple, if we keep versions, keep latest versions in place, and if we dont want to maintain the versions separately in this file,  there is no point to keep the very old versions out there. Also sourcing make the dependency to build.env , if thats the case we dont even need mention of these versions here in this file.

Madhu-1:
> See sourcing can be used, but my point is simple, if we keep versions, keep latest versions in place, and if we dont want to maintain the versions separately in this file, there is no point to keep the very old versions out there. Also sourcing make the dependency to build.env , if thats the case we dont even need mention of these versions here in this file.

I don't like the idea of updating it in multiple places which is completely useless. if you think it can be removed please remove and add a validation to check the image versions are set.

humblec:
So, locally it works, however in CI, the `build.env` path is different... any thoughts  here @ceph/ceph-csi-contributors

Madhu-1:
> So, locally it works, however in CI, the `build.env` path is different... any thoughts here @ceph/ceph-csi-contributors

source build.env like below?

```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
# shellcheck source=build.env
source "${SCRIPT_DIR}/../build.env"
```

humblec:
> > So, locally it works, however in CI, the `build.env` path is different... any thoughts here @ceph/ceph-csi-contributors
> 
> source build.env like below?
> 
> ```
> SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
> # shellcheck source=build.env
> source "${SCRIPT_DIR}/../build.env"
> ```

I knew this existed some where but was not able to fetch it .. Thanks.. should be good now. :+1:

nixpanic (changes requested):
The PR description and commit message suggest that this is for deploying (by Helm or `deploy/` yaml files). However this is for CI only?

Maybe it would be good to validate that the deployment files actually have the same versions of the sidecars as listed in `build.env`. Currently the versions in `build.env` may be different in `charts/` and `deploy/`.

humblec:
> The PR description and commit message suggest that this is for deploying (by Helm or `deploy/` yaml files). However this is for CI only?
> 
> Maybe it would be good to validate that the deployment files actually have the same versions of the sidecars as listed in `build.env`. Currently the versions in `build.env` may be different in `charts/` and `deploy/`.

I will keep this PR for CI.. Updated commit and PR description.

humblec:
@nixpanic kept this PR for CI only and updated. ptal.. thanks.

scripts/minikube.sh
```diff
@@ -162,11 +162,21 @@ else
...
-CSI_SNAPSHOTTER_VERSION=${CSI_SNAPSHOTTER_VERSION:-"v4.1.1"}
-CSI_PROVISIONER_VERSION=${CSI_PROVISIONER_VERSION:-"v2.2.2"}
-CSI_RESIZER_VERSION=${CSI_RESIZER_VERSION:-"v1.2.0"}
-CSI_NODE_DRIVER_REGISTRAR_VERSION=${CSI_NODE_DRIVER_REGISTRAR_VERSION:-"v2.2.0"}
+
+# as we are no longer keeping the sidecar versions here, try to fetch it from build.env file
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+source "${SCRIPT_DIR}/../build.env"
+
+sidecars=(CSI_ATTACHER_VERSION CSI_SNAPSHOTTER_VERSION CSI_PROVISIONER_VERSION CSI_RESIZER_VERSION CSI_NODE_DRIVER_REGISTRAR_VERSION)
```
Madhu-1: I don't think we need to check the sidecars version by default, this script is used for many tasks. creating minikube cluster. deploying Rook. copying sidecar images. creating pools etc. this check should only be called if someone is using this to copy the sidecar images.

Madhu-1 (changes requested):

scripts/minikube.sh
```diff
@@ -162,11 +162,21 @@ else
...
+for controller in "${sidecars[@]}"; do
+	if [[ "${!controller}" ]]; then
```
nixpanic: For checking empty strings, use `-z`.
No need for listing the versions, that is just clutter IMHO.
This check is not always needed, only for deploying?

scripts/minikube.sh
```diff
@@ -162,11 +162,21 @@ else
...
+		echo "${controller}" : "${!controller}"
+	else
+		echo "${controller}" version is empty, make sure build.env has set this sidecar version
+               exit 1
```
nixpanic: indention needs to be fixed

nixpanic (changes requested):

scripts/minikube.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash -e
-
+# shellcheck disable=SC1091
```
nixpanic: please only do this where needed, not for the whole script

nixpanic (dismissed):

scripts/minikube.sh
humblec: Done.

scripts/minikube.sh
humblec: sure, used `-z`

scripts/minikube.sh
humblec: Done.

humblec:
@nixpanic @Madhu-1 all comments are addressed. ptal .. thanks.

scripts/minikube.sh
```diff
@@ -124,6 +124,21 @@ function validate_container_cmd() {
...
 }
 
+# configure csi sidecar version
+# shellcheck disable=SC1091
+function validate_sidecar() {
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+source "${SCRIPT_DIR}/../build.env"
+
+sidecars=(CSI_ATTACHER_VERSION CSI_SNAPSHOTTER_VERSION CSI_PROVISIONER_VERSION CSI_RESIZER_VERSION CSI_NODE_DRIVER_REGISTRAR_VERSION)
+for controller in "${sidecars[@]}"; do
```
Madhu-1: use sidecar instead of controller for variable name?

scripts/minikube.sh
```diff
@@ -124,6 +124,21 @@ function validate_container_cmd() {
     fi
 }
 
+# configure csi sidecar version
+# shellcheck disable=SC1091
+function validate_sidecar() {
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
```
Madhu-1: function indentation is wrong compared to other functions in the file

scripts/minikube.sh
```diff
@@ -124,6 +124,21 @@ function validate_container_cmd() {
     fi
 }
 
+# configure csi sidecar version
+# shellcheck disable=SC1091
```
Madhu-1: why disable check for the whole function?

humblec:
@nixpanic @Madhu-1 comments are addressed. ptal. thanks.

humblec:
@Madhu-1 @nixpanic ptal. thanks.

humblec:
@Madhu-1 @nixpanic can you please revisit this PR as you have changes requested here?

scripts/minikube.sh
```diff
@@ -290,6 +298,8 @@ cephcsi)
     copy_image_to_cluster "${CEPHCSI_IMAGE_REPO}"/cephcsi:"${CSI_IMAGE_VERSION}" "${CEPHCSI_IMAGE_REPO}"/cephcsi:"${CSI_IMAGE_VERSION}"
     ;;
 k8s-sidecar)
+    echo "validating sidecar controller configuration"
```
Madhu-1: `echo "validating sidecar controller configuration"` to  `echo "validating sidecar images version"`

scripts/minikube.sh
```diff
@@ -124,6 +124,21 @@ function validate_container_cmd() {
     fi
 }
 
+# configure csi sidecar version
+function validate_sidecar() {
```
Madhu-1, line 128: comment and function name is not matching rename this to configure_sidecar or update the comment?

scripts/minikube.sh
```diff
@@ -124,6 +124,21 @@ function validate_container_cmd() {
...
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+# shellcheck disable=SC1091
```
Madhu-1, line 130: the comment is not indented is this expected?

Madhu-1 (commented):
LGTM. left few suggestions

humblec:
@Rakshith-R @yati1998 ptal.. thanks

yati1998 (approved):
LGTM

Rakshith-R (approved):

humblec:
@Madhu-1 ptal.. thanks.

Madhu-1 (approved):

humblec:
@Mergifyio rebase

Rakshith-R:
@Mergifyio rebase

humblec:
@Mergifyio refresh

humblec:
@Mergifyio rebase

PR #2820, opened 2022-01-25
ci: remove duplicate DCO validation

humblec:
This commit remove the duplicate DCO in default rule.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -37,7 +37,6 @@ queue_rules:
               - "status-success=ci/centos/mini-e2e/k8s-1.23"
               - "status-success=ci/centos/upgrade-tests-cephfs"
               - "status-success=ci/centos/upgrade-tests-rbd"
-              - "status-success=DCO"
           - and:
               - base=release-v3.4
               - "status-success=codespell"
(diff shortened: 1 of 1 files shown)


humblec:
we are indidually handling the merge rules for devel and release branches, so may be this is not required, however having it in default should also be fine i think. `dependabot` commits are exception, so if we add this in the default rule, it could cause issues. however would like to seek the thoughts here.

Madhu-1 (changes requested):
This will block PR's created by dependent bot as  commitlint doesn't passes

.mergify.yml
```diff
@@ -26,6 +26,7 @@ queue_rules:
               - "status-success=golangci-lint"
               - "status-success=mod-check"
               - "status-success=lint-extras"
+              - "status-success=commitlint"
```
nixpanic: The commit message does not change when rebasing the PR. Once Mergify decides to put it in this queue, the commitlint check has been validated already (with dependabot commits skipped). There is no need to include commitlint here again (like that, DCO could be removed here too).

nixpanic (changes requested):

.mergify.yml
humblec: yeah , true.. let me just remove DCO for now and the rule conditions can be validated later..

Madhu-1 (commented):
update PR description.

Madhu-1 (changes requested):
Remove DOC in other `and` conditions too

Madhu-1 (approved):

Madhu-1:
Manual merge is required. merge this once PR's in the queue gets merged https://dashboard.mergify.com/github/ceph/repo/ceph-csi/queues

nixpanic (approved):

PR #2789, opened 2022-01-17
deploy: update csi-snapshotter sidecar to v5.0.1

humblec:
This release of snapshotter has a breaking change as mentioned
in the release note:

Refer# https://github.com/kubernetes-csi/external-snapshotter/releases/tag/v5.0.0

RBAC rules are also updated with this commit.

Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

PR #2822, opened 2022-01-25
deploy: make the snapshotter version in parity

humblec:
The build.env versions are not matching with current version
we have in the deployment artifacts, this correct the same.


Madhu-1 (changes requested):
This can be taken care of at https://github.com/ceph/ceph-csi/pull/2789.

humblec:
THis is no longer rquired and I am closing this one

PR #2833, opened 2022-01-31
implement VOLUME_CONDITION capability

### LINKED ISSUE #2794: Implement `VOLUME_CONDITION` capability

Healer Controller checks if a volume is still mounted and usable. To check whether a volume is usable, the CSI driver shall check if filesystem is corrupted, whether there are bad blocks, etc. in this RPC.

Instead of adding a new RPC, we can leverage the existing NodeGetVolumeStats RPC.

In the NodeGetVolumeStatsRequest, there are volume_id, volume_path, and staging_target_path. CSI driver can figure out whether a volume is still mounted based on these parameters.

A new message volume_condition will be added to NodeGetVolumeStatsResponse. A new Node capability `VOLUME_CONDITION` will be added to indicate whether a CSI driver has implemented this function. In a volume_condition message, there is a boolean parameter abnormal indicating whether the volume is normal or not and a message that describes the details of the volume condition.

# What new functionality do you want?

Provide a way to check the volume health status

# What is the value to the end-user? (why is it a priority?) #

The healer controller can leverage this to auto-fix the volume without user intervention.

# Additional context #

Read more at https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1432-volume-health-monitor#add-node-volume-health-function

pkalever:

Node Volume Health checks if a volume is still mounted and usable. To check whether a volume is usable, the CSI driver shall check if filesystem is corrupted, whether there are bad blocks, etc. in this RPC.                                                                      
```                                                                   
// VolumeCondition represents the current condition of a volume.      
                                                                      
message VolumeCondition {                                             
  option (alpha_message) = true;                                      
                                                                      
  // Normal volumes are available for use and operating optimally.    
  // An abnormal volume does not meet these criteria.                 
  // This field is REQUIRED.                                          
  bool abnormal = 1;                                                  
                                                                      
  // The message describing the condition of the volume.              
  // This field is REQUIRED.                                          
  string message = 2;                                                 
}                                                                     
```                                                                   
                                                                      
## Related issues ##

Fixes: #2794

## Logs ##
**With Feature Gates enabled:**
[csi-rbdplugin.log](https://github.com/ceph/ceph-csi/files/8940528/csi-rbdplugin.log)
[metrics.log](https://github.com/ceph/ceph-csi/files/8940529/metrics.log)
**With Feature Gates disabled:**
[csi-rbdplugin_without_fg.log](https://github.com/ceph/ceph-csi/files/8940708/csi-rbdplugin_without_fg.log)
[metrics_without_fg.log](https://github.com/ceph/ceph-csi/files/8940709/metrics_without_fg.log)

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -603,6 +603,13 @@ func (ns *NodeServer) NodeGetCapabilities(
 			{
 				Type: &csi.NodeServiceCapability_Rpc{
 					Rpc: &csi.NodeServiceCapability_RPC{
+						Type: csi.NodeServiceCapability_RPC_VOLUME_CONDITION,
+					},
+				},
+			},
+			{
+				Type: &csi.NodeServiceCapability_Rpc{
+					Rpc: &csi.NodeServiceCapability_RPC{
 						Type: csi.NodeServiceCapability_RPC_GET_VOLUME_STATS,
 					},
 				},
@@ -623,22 +630,32 @@ func (ns *NodeServer) NodeGetVolumeStats(
 	ctx context.Context,
 	req *csi.NodeGetVolumeStatsRequest,
 ) (*csi.NodeGetVolumeStatsResponse, error) {
-	var err error
-	targetPath := req.GetVolumePath()
-	if targetPath == "" {
-		err = fmt.Errorf("targetpath %v is empty", targetPath)
-
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+	volumePath := req.GetStagingTargetPath()
+	if volumePath == "" {
+		if volumePath = req.GetVolumePath(); volumePath == "" {
+			return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+		}
+	} else {
+		volumePath = volumePath + "/" + req.GetVolumeId()
 	}
 
-	stat, err := os.Stat(targetPath)
+	var volCondition csi.VolumeCondition
+	stat, err := os.Stat(volumePath)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to get stat for targetpath %q: %v", targetPath, err)
+		volCondition.Abnormal = true
+		volCondition.Message = fmt.Sprintf("failed to get stat for volumepath %q: %v", volumePath, err)
+
+		return &csi.NodeGetVolumeStatsResponse{
+			VolumeCondition: &volCondition,
+		}, nil
 	}
 
 	if stat.Mode().IsDir() {
-		return csicommon.FilesystemNodeGetVolumeStats(ctx, ns.Mounter, targetPath)
+		res, err := csicommon.FilesystemNodeGetVolumeStats(ctx, ns.Mounter, volumePath)
+		res.VolumeCondition = &volCondition
+
+		return res, err
 	}
 
-	return nil, status.Errorf(codes.InvalidArgument, "targetpath %q is not a directory or device", targetPath)
+	return nil, status.Errorf(codes.InvalidArgument, "volumePath %q is not a directory or device", volumePath)
 }
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -45,6 +45,8 @@ type NodeServer struct {
 	// A map storing all volumes with ongoing operations so that additional operations
 	// for that same volume (as defined by VolumeID) return an Aborted error
 	VolumeLocks *util.VolumeLocks
+
+	StagingPath string
 }
 
 // stageTransaction struct represents the state a transaction was when it either completed
@@ -1109,6 +1111,13 @@ func (ns *NodeServer) NodeGetCapabilities(
 			{
 				Type: &csi.NodeServiceCapability_Rpc{
 					Rpc: &csi.NodeServiceCapability_RPC{
+						Type: csi.NodeServiceCapability_RPC_VOLUME_CONDITION,
+					},
+				},
+			},
+			{
+				Type: &csi.NodeServiceCapability_Rpc{
+					Rpc: &csi.NodeServiceCapability_RPC{
 						Type: csi.NodeServiceCapability_RPC_EXPAND_VOLUME,
 					},
 				},
@@ -1221,31 +1230,100 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
 	return false
 }
 
+// extractPVNameFromPath extracts PV name from path containing PV name.
+func extractPVNameFromPath(path string) string {
+	str := strings.Split(path, "/pvc-")[1]
+	str = strings.Split(str, "/")[0]
+
+	return "pvc-" + str
+}
+
 // NodeGetVolumeStats returns volume stats.
 func (ns *NodeServer) NodeGetVolumeStats(
 	ctx context.Context,
 	req *csi.NodeGetVolumeStatsRequest,
 ) (*csi.NodeGetVolumeStatsResponse, error) {
-	var err error
-	targetPath := req.GetVolumePath()
-	if targetPath == "" {
-		err = fmt.Errorf("targetpath %v is empty", targetPath)
-
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+	var (
+		imgInfo rbdImageMetadataStash
+		err     error
+	)
+	var metadataPath string
+	volumePath := req.GetStagingTargetPath()
+	if volumePath == "" {
+		if volumePath = req.GetVolumePath(); volumePath == "" {
+			return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+		}
+		pvName := extractPVNameFromPath(volumePath)
+		metadataPath = ns.StagingPath + pvName + "/globalmount"
+	} else {
+		metadataPath = volumePath
+		volumePath = volumePath + "/" + req.GetVolumeId()
 	}
 
-	stat, err := os.Stat(targetPath)
+	imgInfo, err = lookupRBDImageMetadataStash(metadataPath)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to get stat for targetpath %q: %v", targetPath, err)
+		log.ErrorLog(ctx, "failed to find image metadata at stagingPath %q: %v", volumePath, err)
+
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	return checkBlockDeviceStatus(ctx, volumePath, &imgInfo, ns.Mounter)
+}
+
+// checkBlockDeviceStatus checks the block device current status.
+func checkBlockDeviceStatus(
+	ctx context.Context,
+	volumePath string,
+	imgInfo *rbdImageMetadataStash,
+	mounter mount.Interface,
+) (*csi.NodeGetVolumeStatsResponse, error) {
+	var volCondition csi.VolumeCondition
+
+	stat, err := os.Stat(volumePath)
+	if err != nil {
+		volCondition.Abnormal = true
+		volCondition.Message = fmt.Sprintf("failed to get stat for volumepath %q: %v", volumePath, err)
+
+		return &csi.NodeGetVolumeStatsResponse{
+			VolumeCondition: &volCondition,
+		}, nil
+	}
+
+	// userspace mounters
+	var stdout, stderr string
+	if imgInfo.NbdAccess {
+		stdout, stderr, err = util.ExecCommand(ctx, "rbd-nbd", "list-mapped")
+		if err != nil || stderr != "" {
+			log.ErrorLog(ctx, "running rbd-nbd list-mapped failed with error:%v, stderr:%s", err, stderr)
+
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		if !strings.Contains(stdout, imgInfo.DevicePath) && !strings.Contains(stdout, imgInfo.ImageName) {
+			volCondition.Abnormal = true
+			volCondition.Message = fmt.Sprintf("rbd-nbd process not found for image: %s, device: %s",
+				imgInfo.String(), imgInfo.DevicePath)
+
+			return &csi.NodeGetVolumeStatsResponse{
+				VolumeCondition: &volCondition,
+			}, nil
+		}
+	}
+
+	var res *csi.NodeGetVolumeStatsResponse
 	if stat.Mode().IsDir() {
-		return csicommon.FilesystemNodeGetVolumeStats(ctx, ns.Mounter, targetPath)
+		res, err = csicommon.FilesystemNodeGetVolumeStats(ctx, mounter, volumePath)
+		res.VolumeCondition = &volCondition
+
+		return res, err
 	} else if (stat.Mode() & os.ModeDevice) == os.ModeDevice {
-		return blockNodeGetVolumeStats(ctx, targetPath)
+		res, err = blockNodeGetVolumeStats(ctx, volumePath)
+		res.VolumeCondition = &volCondition
+
+		return res, err
 	}
 
-	return nil, fmt.Errorf("targetpath %q is not a block device", targetPath)
+	return nil, fmt.Errorf("volumePath %q is not a block device", volumePath)
 }
 
 // blockNodeGetVolumeStats gets the metrics for a `volumeMode: Block` type of
--- a/internal/rbd/driver/driver.go
+++ b/internal/rbd/driver/driver.go
@@ -140,6 +140,7 @@ func (r *Driver) Run(conf *util.Config) {
 		if err != nil {
 			log.FatalLogMsg("failed to start node server, err %v\n", err)
 		}
+		r.ns.StagingPath = conf.StagingPath
 		var attr string
 		attr, err = rbd.GetKrbdSupportedFeatures()
 		if err != nil && !errors.Is(err, os.ErrNotExist) {


internal/cephfs/nodeserver.go
```diff
@@ -403,22 +410,29 @@ func (ns *NodeServer) NodeGetCapabilities(
...
-	var err error
-	targetPath := req.GetVolumePath()
-	if targetPath == "" {
-		err = fmt.Errorf("targetpath %v is empty", targetPath)
-
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+	// Get volume path
+	// With Kubernetes version>=v1.19.0, stats request carries volume_path and
+	// staging_target_path, what csi requires is staging_target_path.
+	volumePath := req.GetStagingTargetPath()
```
Madhu-1, line 633: with kubernetes 1.23 i see it contains only the volumePath, not the stagingPath, first check for volumePath if it's empty check the staging path?

```bash
I0201 05:53:39.161788       1 utils.go:195] ID: 32 GRPC request: {"volume_id":"0001-0009-rook-ceph-0000000000000001-2538aeb8-8312-11ec-921d-021bb3f04b38","volume_path":"/var/lib/kubelet/pods/42f2f237-971e-4ab3-b743-332bc16f4eee/volumes/kubernetes.io~csi/pvc-0e32af29-a511-4161-8519-f39dcb3c4ed8/mount"}
```

internal/rbd/nodeserver.go
```diff
@@ -1190,26 +1197,56 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
+		return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+	}
 
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+	stagingParentPath := req.GetStagingTargetPath()
+	imgInfo, err := lookupRBDImageMetadataStash(stagingParentPath)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to find image metadata, at stagingPath: %s, err: %v", stagingParentPath, err)
+
+		return nil, err
```
Madhu-1: incase of failure need to return GRPC error message

internal/rbd/nodeserver.go
```diff
@@ -1190,26 +1197,56 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
-		return nil, status.Errorf(codes.InvalidArgument, "failed to get stat for targetpath %q: %v", targetPath, err)
+		volCondition.Abnormal = true
+		volCondition.Message = fmt.Sprintf("failed to get stat for volumepath %q: %v", volumePath, err)
+	}
+
+	// userspace mounters
+	if imgInfo.NbdAccess {
+		stdout, stderr, err := util.ExecCommand(ctx, "ps", "aux", "|", "grep", "[r]bd-nbd")
+		if err != nil || stderr != "" {
+			log.ErrorLog(ctx, "running ps failed with error:%v, stderr:%s", err, stderr)
```
Madhu-1: incase of failure need to return GRPC error message

internal/cephfs/nodeserver.go
pkalever: You must be checking an internal structure?

See https://github.com/kubernetes/kubernetes/pull/84832/files#diff-4d92d6f31bff887bff3db01184c2977481dc68c5725dee8209238df1d71b9f03R3863

I can see this is part of v1.18.0-alpha.0-274-g0513ee7d6cc 


I remember we used similar stuff here https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/nodeserver.go#L990

pkalever: In fact I see it exist since v1.17.0 as per git history

```
[0] pkalever 😎 kubernetes✨ git contain 72bcec4e4d516dcac227606541f990af13fb2918 | grep v1.17.0
v1.17.0
v1.17.0-beta.2
v1.17.0-rc.1
v1.17.0-rc.2
```

Madhu-1: >I remember we used similar stuff here https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/nodeserver.go#L990

You are referring to NodeExpandVolume, not NodeGetVolumeStats?

pkalever: Sure, lets use VolumePath because kubelet still uses volumePath in its NodeGetVolumeStatsRequest.

internal/rbd/nodeserver.go
```diff
@@ -1190,26 +1197,53 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
+		imgInfo rbdImageMetadataStash
+		err     error
+	)
+	volumePath := req.GetStagingTargetPath()
+	if volumePath == "" {
+		if volumePath = req.GetVolumePath(); volumePath == "" {
+			return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+		}
+	} else {
+		imgInfo, err = lookupRBDImageMetadataStash(volumePath)
```
Madhu-1: volumePath will be statingTargetPath+volumeID incase if its a stagingpath https://github.com/ceph/ceph-csi/blob/28fef9b3797e41f277d11d9a3bde11e971994503/internal/rbd/nodeserver.go#L284-L285?

internal/rbd/nodeserver.go
```diff
@@ -1190,26 +1197,53 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
+	stat, err := os.Stat(volumePath)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to get stat for targetpath %q: %v", targetPath, err)
+		volCondition.Abnormal = true
+		volCondition.Message = fmt.Sprintf("failed to get stat for volumepath %q: %v", volumePath, err)
+	}
+
+	// userspace mounters
+	if imgInfo.NbdAccess {
+		stdout, stderr, err := util.ExecCommand(ctx, "ps", "aux", "|", "grep", "[r]bd-nbd")
```
Madhu-1: is there any simplification for this one? if we have more nbd mounts this will return a big list of mounts?

Madhu-1: @pkalever, not all base images we use for cephcsi will have `ps` installed by default (downstream)?

pkalever: We depend on many such utilities and I would expect it to be present. I think `ps` is part of `procps-ng` package, can we fix it with the container images?

In case `ps` is not available then we need to search all the `/proc/*/cmdline` one at a time and that would be time-consuming. 

There is also another way to do this via `rbd-nbd list-mapped` command, but maybe lets conclude if the unavailability of `ps` is acceptable for us, because maybe others might use it (if not now maybe in the future)

cc: @nixpanic

Madhu-1: we don't have `ps` package in downstream cephcsi containers. if we want it we need to make it as required.

```
[🎩︎]mrajanna@fedora logs $]oc exec -it csi-rbdplugin-provisioner-58995c7c7b-hrchr -c csi-rbdplugin -- sh
sh-4.4# ps
sh: ps: command not found
sh-4.4#
```

pkalever: Ok, let's then mark it as required?

pkalever: Using pstree for the time being, lets see what others got to say.

internal/rbd/nodeserver.go
```diff
@@ -1190,26 +1197,54 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
 func (ns *NodeServer) NodeGetVolumeStats(
 	ctx context.Context,
 	req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
-	var err error
-	targetPath := req.GetVolumePath()
-	if targetPath == "" {
-		err = fmt.Errorf("targetpath %v is empty", targetPath)
+	var (
```
nixpanic, line 1246: please split checking condition and stats in helper functions

internal/rbd/nodeserver.go
```diff
@@ -1190,26 +1197,54 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
+	volumePath := req.GetStagingTargetPath()
+	if volumePath == "" {
+		if volumePath = req.GetVolumePath(); volumePath == "" {
+			return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+		}
+	} else {
+		volumePath = volumePath + "/" + req.GetVolumeId()
+		imgInfo, err = lookupRBDImageMetadataStash(volumePath)
+		if err != nil {
+			log.ErrorLog(ctx, "failed to find image metadata, at stagingPath: %s, err: %v", volumePath, err)
```
nixpanic: use common formatting of errors, like `"failed to find image metadata at stagingPath %q: %v", volumePath, err`

internal/rbd/nodeserver.go
```diff
@@ -1190,26 +1197,54 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
 
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
 	}
 
-	stat, err := os.Stat(targetPath)
+	var volCondition *csi.VolumeCondition
+	stat, err := os.Stat(volumePath)
```
nixpanic, line 1282: `stat` is not a sufficient check for everything. In case a filesystem went read-only after a network outage, `stat` will still report success.

`stat` on a block-device will probably always succeed?

internal/rbd/nodeserver.go
```diff
@@ -1190,26 +1197,54 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to get stat for targetpath %q: %v", targetPath, err)
+		volCondition.Abnormal = true
+		volCondition.Message = fmt.Sprintf("failed to get stat for volumepath %q: %v", volumePath, err)
+	}
+
+	// userspace mounters
+	if imgInfo.NbdAccess {
+		stdout, stderr, err := util.ExecCommand(ctx, "pstree", "--arguments", "|", "grep", "[r]bd-nbd")
```
nixpanic: This is rather ugly. Is there no way to validate that rbd-nbd is running for this particular volume?

internal/rbd/nodeserver.go
```diff
@@ -1218,7 +1253,10 @@ func (ns *NodeServer) NodeGetVolumeStats(
 // connect to the Ceph cluster.
 //
 // TODO: https://github.com/container-storage-interface/spec/issues/371#issuecomment-756834471
-func blockNodeGetVolumeStats(ctx context.Context, targetPath string) (*csi.NodeGetVolumeStatsResponse, error) {
+func blockNodeGetVolumeStats(
+	ctx context.Context,
+	targetPath string,
+	volCondition *csi.VolumeCondition) (*csi.NodeGetVolumeStatsResponse, error) {
```
nixpanic: This feels wrong. No need to pass `*csi.VolumeCondition` if it is not consumed. Just set the `VolumeCondition` in the function that calls `blockNodeGetVolumeStats()` instead.

nixpanic (changes requested):
A commit message should describe what is implemented, not a copy/paste of a protocol structure.

Using `stat` is not sufficient to detect errors, consider documenting limitations of the `VOLUME_CONDITION`.

internal/rbd/nodeserver.go
pkalever: ### Problem with Writes:

**FS:**
Detecting if the accessMode is ReadOnly might not be possible with the info available at NodeGetVolumeStatsRequest{}, if we try to read that from mounts, then if the mounts turn into read-only we will get wrong info.

**Block:**
On the other side, we cannot write data into a block volume, because that could overwrite some important data.

### Problem with Reads:


**FS:**
We don't know which file to read unless we write something into the root dir of mount and read the same, as mentioned above writing data is not easy.

**Block:**
I think Reads on block volume can be done, without any hiccups, but again I just don't want to address 1/4 cases.

internal/rbd/nodeserver.go
pkalever: I know, this is just a handy way! I will see if we can fetch the image spec details with `rbd-nbd list-mapped` command.

internal/rbd/nodeserver.go
pkalever: @nixpanic please let me know if you think of something better?

nixpanic: For RBD devices, hopefully some rbd command can check the status of the mapped device.

I do not know what would be suitable for good CephFS health checking.

I'd recommend to get in touch with the Ceph developers and ask for their recommendation. Ideally use their developers mailinglist, so that we can link to your request and their responses in follow-up PRs.

internal/rbd/nodeserver.go
pkalever:
```
[root@minikube /]# rbd-nbd list-mapped                              
id       pool         namespace  image                                         snap  device   
343256   replicapool             csi-vol-b135ce39-7cd5-11eb-8fd5-6aabb2d74ddd  -     /dev/nbd0
2132286  replicapool             csi-vol-0a0f037a-8af4-11eb-8fd5-6aabb2d74ddd  -     /dev/nbd1

```
I have used list-mapped now.

internal/rbd/nodeserver.go
pkalever: @idryomov is there any better way to detect possible filesystem errors/bad blocks with CephFS and RBD in both file and block modes? This VOLUME_CONDITION capability should check if the volume is mounted and usable.

idryomov: I'm aware of anything similar to what you seem to be looking for here.  If the device/filesystem is mapped/mounted, it's expected to work (unless its client instance was blocklisted).  If the cluster is broken, unreachable, etc, we expect that to be a temporary condition and block I/O until it's resolved.  So from that POV, a mapped/mounted volume is always usable.

pkalever: Thanks @idryomov :+1:

internal/rbd/nodeserver.go
```diff
@@ -1227,26 +1234,70 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
 	}
 
-	stat, err := os.Stat(targetPath)
+	return checkBlockDeviceStatus(ctx, volumePath, &imgInfo)
+}
+
+func checkBlockDeviceStatus(
```
Rakshith-R, line 1274: Add a descriptive comment on what the function does ?

internal/cephfs/nodeserver.go
```diff
@@ -490,22 +497,28 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+		volumePath = volumePath + "/" + req.GetVolumeId()
 	}
 
-	stat, err := os.Stat(targetPath)
+	var volCondition *csi.VolumeCondition
+	stat, err := os.Stat(volumePath)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to get stat for targetpath %q: %v", targetPath, err)
+		volCondition.Abnormal = true
+		volCondition.Message = fmt.Sprintf("failed to get stat for volumepath %q: %v", volumePath, err)
```
Madhu-1, line 646: missing return here? if stats is failing what do we need to check?

internal/rbd/nodeserver.go
```diff
@@ -1227,26 +1234,70 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+			return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+		}
+	} else {
+		volumePath = volumePath + "/" + req.GetVolumeId()
+		imgInfo, err = lookupRBDImageMetadataStash(volumePath)
+		if err != nil {
+			log.ErrorLog(ctx, "failed to find image metadata at stagingPath %q: %v", volumePath, err)
 
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+			return nil, status.Error(codes.NotFound, err.Error())
```
Madhu-1: why this is changed to `NotFound` error code?

internal/rbd/nodeserver.go
```diff
@@ -1227,26 +1234,70 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+func checkBlockDeviceStatus(
+	ctx context.Context,
+	volumePath string,
+	imgInfo *rbdImageMetadataStash,
+) (*csi.NodeGetVolumeStatsResponse, error) {
+	var volCondition *csi.VolumeCondition
+	stat, err := os.Stat(volumePath)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to get stat for targetpath %q: %v", targetPath, err)
+		volCondition.Abnormal = true
```
Madhu-1, line 1284: if `stat` is failing why do we need to go ahead and check for other things?

internal/rbd/nodeserver.go
pkalever: not only the return changes the whole code changes, when lookup is failing what error code is suitable ?

Madhu-1: Not found is not suitable for all errors. if error is `ErrMissingStash` you can consider it as not for. keeping Internal error is also fine.

pkalever: Done now.

Madhu-1 (commented):
changes LGTM, as we dont have CI for this one @pkalever can you paste the csi plugin logs and kubelet metrics for this one to make sure everything is working fine?

pkalever:
@Mergifyio rebase

internal/rbd/nodeserver.go
nixpanic: There are some quicks done in `posix_fs_health_check()` (from the [GlusterFS sources](https://github.com/gluster/glusterfs/blob/devel/xlators/storage/posix/src/posix-helpers.c)). XFS seems to behave different from ext4, so that probably needs some checking too.

Ideally read/write PVCs are tested by reading and writing, read-only PVCs by reading, only. It might make sense to write to a (hidden) file, store a timestamp and compare that after reading. Or just `stat()` the root of the filesystem. Anything you do needs a timeout, as the operation might be blocked in case network is unavailable or something. It is crucial for the `VOLUME_CONDITION` to be marked unhealthy, and not return a failure while requesting the condition.

pkalever:
@Madhu-1 please find the logs requested
**With Feature Gates enabled:**
[csi-rbdplugin.log](https://github.com/ceph/ceph-csi/files/8940528/csi-rbdplugin.log)
[metrics.log](https://github.com/ceph/ceph-csi/files/8940529/metrics.log)
**With Feature Gates disabled:**
[metrics_without_fg.log](https://github.com/ceph/ceph-csi/files/8940709/metrics_without_fg.log)
[csi-rbdplugin_without_fg.log](https://github.com/ceph/ceph-csi/files/8940708/csi-rbdplugin_without_fg.log)

internal/rbd/nodeserver.go
pkalever: unfortunately, reading and writing are not possible, please go through https://github.com/ceph/ceph-csi/pull/2833/files#r799183365 and see if that answers it.

nixpanic: It should be possible to read the mount options,and it it is not read-only, you should be able to write to a file on the filesystem. Document the filename you are using well, so that users can find the name in the documentation.

Without writing to a file, and reading the contents (timestamp?) afterwards, the checks would not be very useful as only few issues would be detected.

Users consuming block-mode volumes are on their own regarding this, sadly.

nixpanic: There is also a move to use a subdir for the actual PV, and not the root of the filesystem. That gives the CSI-driver an option to store metadata (for fscrypt) in the root of the volume, outside of the view of the consumers. Doing this for a health-check file makes sense too.

pkalever:
@Mergifyio rebase

internal/rbd/nodeserver.go
pkalever: @nixpanic I had opened a new issue to track improvements.
As discussed, let us start with a simple stat() and we will improve it with https://github.com/ceph/ceph-csi/issues/3200 soon with another PR.

pkalever:
@Mergifyio rebase

pkalever:
@Mergifyio rebase

pkalever:
@Mergifyio rebase

internal/rbd/nodeserver.go
```diff
@@ -1221,31 +1230,100 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
 	return false
 }
 
+// extractPVNameFromPath extracts PV name from path containing PV name.
```
nixpanic, line 1233: This needs a bit more explaining with examples, and ideally unit-tests. The implementation is probably very Kubernetes specific now too?

internal/rbd/nodeserver.go
```diff
@@ -1221,31 +1230,100 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
-	targetPath := req.GetVolumePath()
-	if targetPath == "" {
-		err = fmt.Errorf("targetpath %v is empty", targetPath)
+	var (
+		imgInfo rbdImageMetadataStash
+		err     error
+	)
+	var metadataPath string
+	volumePath := req.GetStagingTargetPath()
+	if volumePath == "" {
```
nixpanic, line 1252: Have you seen it happen that `GetStagingTargetPath()` is empty? I wonder if there isn't a nicer way to find the location of the metadata json file

internal/rbd/nodeserver.go
```diff
@@ -1221,31 +1230,100 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
 	} else if (stat.Mode() & os.ModeDevice) == os.ModeDevice {
-		return blockNodeGetVolumeStats(ctx, targetPath)
+		res, err = blockNodeGetVolumeStats(ctx, volumePath)
+		res.VolumeCondition = &volCondition
+
+		return res, err
 	}
 
-	return nil, fmt.Errorf("targetpath %q is not a block device", targetPath)
+	return nil, fmt.Errorf("volumePath %q is not a block device", volumePath)
```
nixpanic, line 1326: errors that get returned from this function should be GRPC errors (`status.Error()`)

internal/rbd/nodeserver.go
```diff
@@ -1221,31 +1230,100 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
-		return nil, status.Errorf(codes.InvalidArgument, "failed to get stat for targetpath %q: %v", targetPath, err)
+		volCondition.Abnormal = true
+		volCondition.Message = fmt.Sprintf("failed to get stat for volumepath %q: %v", volumePath, err)
+
+		return &csi.NodeGetVolumeStatsResponse{
+			VolumeCondition: &volCondition,
+		}, nil
+	}
+
+	// userspace mounters
```
nixpanic, line 1292: It would be cleaner to have this check in a helper function like `checkHealthNBD(volume, path)`. The function can return an error that can then be used for marking the volume unhealthy.

nixpanic (changes requested):

Madhu-1:
Hello there! There hasn't been any activity on this PR for quite some time, so I have removed the "keepalive" label. To avoid it being closed as stale shortly, please revisit and give it some attention. Thank you!

PR #2859, opened 2022-02-04
build: disable removed Apache Arrow repository

nixpanic:
The CentOS 8 repository for Apache Arrow has been removed. This causes
container-image builds fail with the following error:

    Errors during downloading metadata for repository 'apache-arrow-centos':
      - Status code: 404 for https://apache.jfrog.io/artifactory/arrow/centos/8/x86_64/repodata/repomd.xml (IP: 54.190.66.70)
    Error: Failed to download metadata for repo 'apache-arrow-centos': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried

The Ceph base image has `arrow/centos/8` configured, maybe Apache Arrow
offers a CentOS Stream 8 repository now? Once the Ceph container-image
has been updated, the repository can be enabled again.

Ceph-CSI does not depend on Apache Arrow, so there is no functional
change by disabling the repository.

See-also: ceph/ceph-container#1990

---

<details>

  progress or results

</details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -203,6 +203,7 @@ ifeq ($(USE_PULLED_IMAGE),no)
 .devel-container-id: GOARCH ?= $(shell go env GOARCH 2>/dev/null)
 .devel-container-id: .container-cmd scripts/Dockerfile.devel
 	[ ! -f .devel-container-id ] || $(CONTAINER_CMD) rmi $(CSI_IMAGE_NAME):devel
+	$(RM) .devel-container-id
 	$(CONTAINER_CMD) build $(CPUSET) --build-arg BASE_IMAGE=$(BASE_IMAGE) --build-arg GOARCH=$(GOARCH) -t $(CSI_IMAGE_NAME):devel -f ./scripts/Dockerfile.devel .
 	$(CONTAINER_CMD) inspect -f '{{.Id}}' $(CSI_IMAGE_NAME):devel > .devel-container-id
 else
@@ -216,6 +217,7 @@ ifeq ($(USE_PULLED_IMAGE),no)
 # create a (cached) container image with dependencies for testing cephcsi
 .test-container-id: .container-cmd build.env scripts/Dockerfile.test
 	[ ! -f .test-container-id ] || $(CONTAINER_CMD) rmi $(CSI_IMAGE_NAME):test
+	$(RM) .test-container-id
 	$(CONTAINER_CMD) build $(CPUSET) --build-arg GOARCH=$(GOARCH) -t $(CSI_IMAGE_NAME):test -f ./scripts/Dockerfile.test .
 	$(CONTAINER_CMD) inspect -f '{{.Id}}' $(CSI_IMAGE_NAME):test > .test-container-id
 else
--- a/deploy/cephcsi/image/Dockerfile
+++ b/deploy/cephcsi/image/Dockerfile
@@ -23,6 +23,11 @@ RUN source /build.env && \
 # test if the downloaded version of Golang works (different arch?)
 RUN ${GOROOT}/bin/go version && ${GOROOT}/bin/go env
 
+# FIXME: Ceph does not need Apache Arrow anymore, some container images may
+# still have the repository enabled. Disabling the repository can be removed in
+# the future, see https://github.com/ceph/ceph-container/pull/1990 .
+RUN dnf config-manager --disable apache-arrow-centos || true
+
 RUN dnf -y install \
 	librados-devel librbd-devel \
 	/usr/bin/cc \
--- a/scripts/Dockerfile.devel
+++ b/scripts/Dockerfile.devel
@@ -17,6 +17,11 @@ RUN source /build.env \
     && mkdir -p /usr/local/go \
     && curl https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-${GOARCH}.tar.gz | tar xzf - -C ${GOROOT} --strip-components=1
 
+# FIXME: Ceph does not need Apache Arrow anymore, some container images may
+# still have the repository enabled. Disabling the repository can be removed in
+# the future, see https://github.com/ceph/ceph-container/pull/1990 .
+RUN dnf config-manager --disable apache-arrow-centos || true
+
 RUN dnf -y install \
 	git \
 	make \


nixpanic:
This should address failing builds like https://github.com/ceph/ceph-csi/runs/5063254699?check_suite_focus=true

humblec (approved):

humblec:
@Rakshith-R ptal

yati1998 (approved):

mykaul:
I *think* this issue was fixed in Ceph containers (https://github.com/ceph/ceph-container/pull/1990 ?), but it does raise (again) the question of which containers of ceph-csi do actually need to be built on top of Ceph (And which can use minimal UBI, for example!)

Rakshith-R:
> I _think_ this issue was fixed in Ceph containers ([ceph/ceph-container#1990](https://github.com/ceph/ceph-container/pull/1990) ?), 

There is no image release  with the fix yet: https://quay.io/repository/ceph/ceph?tag=v16&tab=tags.
(The released images are a week old)

Madhu-1:
> I _think_ this issue was fixed in Ceph containers ([ceph/ceph-container#1990](https://github.com/ceph/ceph-container/pull/1990) ?), but it does raise (again) the question of which containers of ceph-csi do actually need to be built on top of Ceph (And which can use minimal UBI, for example!)

We have only one container image for cephcsi (cephfs and rbd). We do need ceph libraries ( librdbd-devel and librados-devel) for go-ceph and golang installed on the container where we are building the cephcsi images.

mykaul:
> > I _think_ this issue was fixed in Ceph containers ([ceph/ceph-container#1990](https://github.com/ceph/ceph-container/pull/1990) ?), but it does raise (again) the question of which containers of ceph-csi do actually need to be built on top of Ceph (And which can use minimal UBI, for example!)
> 
> We have only one container image for cephcsi (cephfs and rbd). We do need ceph libraries ( librdbd-devel and librados-devel) for go-ceph and golang installed on the container where we are building the cephcsi images.

Interesting, I thought we've had some side containers as well, but I don't see it:
[ykaul@ykaul glusterfs]$ podman image list
REPOSITORY                                        TAG            IMAGE ID      CREATED       SIZE
quay.io/cephcsi/cephcsi                           v3.5.1         a68e15f8572a  2 weeks ago   1.3 GB
registry.access.redhat.com/ubi9-beta/ubi-micro    9.0.0beta-455  dc252c7a3afc  3 months ago  26.5 MB
registry.access.redhat.com/ubi9-beta/ubi-micro    latest         dc252c7a3afc  3 months ago  26.5 MB
registry.access.redhat.com/ubi9-beta/ubi          latest         28b0a4b69d9b  3 months ago  229 MB
registry.access.redhat.com/ubi9-beta/ubi-minimal  latest         fcf4fa2ad9f7  3 months ago  97.5 MB


I do believe we should look at how we reduce the size a bit from 1.3GB (or did I pick up the wrong image?)

Madhu-1:
> > > I _think_ this issue was fixed in Ceph containers ([ceph/ceph-container#1990](https://github.com/ceph/ceph-container/pull/1990) ?), but it does raise (again) the question of which containers of ceph-csi do actually need to be built on top of Ceph (And which can use minimal UBI, for example!)
> > 
> > 
> > We have only one container image for cephcsi (cephfs and rbd). We do need ceph libraries ( librdbd-devel and librados-devel) for go-ceph and golang installed on the container where we are building the cephcsi images.
> 
> Interesting, I thought we've had some side containers as well, but I don't see it: 

Do you mean csi sidecar images or something else?

>[ykaul@ykaul glusterfs]$ podman image list REPOSITORY TAG IMAGE ID CREATED SIZE quay.io/cephcsi/cephcsi v3.5.1 a68e15f8572a 2 weeks ago 1.3 GB registry.access.redhat.com/ubi9-beta/ubi-micro 9.0.0beta-455 dc252c7a3afc 3 months ago 26.5 MB registry.access.redhat.com/ubi9-beta/ubi-micro latest dc252c7a3afc 3 months ago 26.5 MB registry.access.redhat.com/ubi9-beta/ubi latest 28b0a4b69d9b 3 months ago 229 MB registry.access.redhat.com/ubi9-beta/ubi-minimal latest fcf4fa2ad9f7 3 months ago 97.5 MB
> 
> I do believe we should look at how we reduce the size a bit from 1.3GB (or did I pick up the wrong image?)

Yes, that is the correct image. but the 1.27 Gib is from the ceph base image. and we just add cephcsi binary (61 Mib) to the [final image](https://github.com/ceph/ceph-csi/blob/devel/deploy/cephcsi/image/Dockerfile#L56-L70). So most of the size contribution is from the base image.  

we need to check is it possible to improve the overall size of the image we are shipping.

mykaul:
> 
> Yes, that is the correct image. but the 1.27 Gib is from the ceph base image. and we just add cephcsi binary (61 Mib) to the [final image](https://github.com/ceph/ceph-csi/blob/devel/deploy/cephcsi/image/Dockerfile#L56-L70). So most of the size contribution is from the base image.
> 
> we need to check is it possible to improve the overall size of the image we are shipping.

Indeed. 2 aspects here - the main one is security actually - fewer packages, smaller attack surface, less need to update. The other is the experience of syncing such large images. Less of an issue (but helps CI a bit).

mykaul:
(BTW, Rook is also using Ceph images as base - how come they did not need to fix it as well?)

Madhu-1:
> (BTW, Rook is also using Ceph images as base - how come they did not need to fix it as well?)


AFAIK They don't install any packages on the base image and they build the image on the machine as there is no dependency on ceph libraries when building the rook binary and they [copy](https://github.com/rook/rook/blob/4c637858676dad5b5c7e2f3ddf2d87caa6b5d53b/images/ceph/Dockerfile#L29) it to the rook image. @travisn @leseb can you confirm is this correct?

travisn:
> > (BTW, Rook is also using Ceph images as base - how come they did not need to fix it as well?)
> 
> AFAIK They don't install any packages on the base image and they build the image on the machine as there is no dependency on ceph libraries when building the rook binary and they [copy](https://github.com/rook/rook/blob/4c637858676dad5b5c7e2f3ddf2d87caa6b5d53b/images/ceph/Dockerfile#L29) it to the rook image. @travisn @leseb can you confirm is this correct?

Correct, Rook just pulls the base image from `quay.io/ceph/ceph` and we basically just add our rook binary.

PR #2877, opened 2022-02-12
helm: make ceph.conf ConfigMap name configurable

### LINKED ISSUE #2858: [chart] ConfigMap conflict between ceph-csi-rbd and ceph-csi-cephfs
# Describe the bug #

When deploying both **ceph-csi-rbd** and **ceph-csi-cephfs** in the same namespace, there is a conflict on ConfigMap ceph-config because the resource name is hard coded:
https://github.com/ceph/ceph-csi/blob/de2489ed7d3afb1322c12ad6ec4f5cc57fe38f81/charts/ceph-csi-rbd/templates/ceph-conf.yaml#L4
https://github.com/ceph/ceph-csi/blob/de2489ed7d3afb1322c12ad6ec4f5cc57fe38f81/charts/ceph-csi-cephfs/templates/ceph-conf.yaml#L4

Right now this issue is blocking the upgrade from ~~3.3.1~~ 3.4.0 to 3.5.1 in my cluster.

# Expected behavior #

Similar solution as CSI config: to suppress creation https://github.com/ceph/ceph-csi/blob/de2489ed7d3afb1322c12ad6ec4f5cc57fe38f81/charts/ceph-csi-rbd/templates/csiplugin-configmap.yaml#L1
And/or have a configurable name: https://github.com/ceph/ceph-csi/blob/de2489ed7d3afb1322c12ad6ec4f5cc57fe38f81/charts/ceph-csi-rbd/templates/csiplugin-configmap.yaml#L5

Please let me know which solution is acceptable and I'll be happy to send a PR.

Thank you!

sgissi:
At both RBD and CephFS helm charts, the ConfigMap name for ceph.conf was hardcoded and led to conflicts. Default setting is backwards compatible.

Fixes: #2858.

Signed-off-by: Silvio Gissi <silvio@gissilabs.com>


charts/ceph-csi-cephfs/README.md
```diff
@@ -133,6 +133,7 @@ charts and their default values.
 | `driverName`                                   | Name of the csi-driver                                                                                                                               | `cephfs.csi.ceph.com`                              |
 | `configMapName`                                | Name of the configmap which contains cluster configuration                                                                                           | `ceph-csi-config`                                  |
 | `externallyManagedConfigmap`                   | Specifies the use of an externally provided configmap                                                                                                | `false`                                            |
+| `cephConfigMapName`                            | Name of the configmap which contains ceph.conf configuration                                                                                           | `ceph-config`                                  |
```
Madhu-1: can you rename this variable to `cephConfConfigMapName`?

charts/ceph-csi-cephfs/values.yaml
```diff
@@ -319,3 +319,5 @@ configMapName: ceph-csi-config
 # configMapKey:
 # Use an externally provided configmap
 externallyManagedConfigmap: false
+# Name of the configmap used for ceph.conf
+cephConfigMapName: ceph-config
```
Madhu-1: change the configmap name to avoid upgrade problems or else the user needs to pass different names for both cephfs and rbd by default?

Madhu-1 (changes requested):
changes look good. left a comment for better variable naming.

charts/ceph-csi-cephfs/values.yaml
sgissi: I think the name is consistent with other deployment options, plus it stays backwards compatible for users upgrading. Only people with multiple deployments of cephfs and/or rbd in the same namespace will need to rename.

Personally I would prefer that all resources were named after the release name to avoid external conflicts but that is a much bigger (and disruptive) change.

charts/ceph-csi-cephfs/README.md
sgissi: The reason I picked this name is that it can include other Ceph config files in the future beyond ceph.conf plus it keeps the variable name short. I can change it though.

charts/ceph-csi-cephfs/values.yaml
Madhu-1: Yes, this is a good point for the upgrades. we can discard this one.

charts/ceph-csi-cephfs/README.md
Madhu-1: It gets complicated. we can have more configmaps its not a problem.

sgissi: Understood, I just pushed a new commit to rename the variable. Thanks for the feedback.

Madhu-1 (changes requested):
@sgissi changes look good. can you squash the commits into 1?

sgissi:
Commits squashed into one.

charts/ceph-csi-cephfs/templates/ceph-conf.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: ceph-config
+  name: {{ .Values.cephConfConfigMapName | quote }}
```
Madhu-1, line 4: i assume `quote` is not required but  should not be a problem

Madhu-1 (approved):

Rakshith-R (approved):

Rakshith-R:
@mergifyio rebase

PR #2883, opened 2022-02-15
helm-charts: Add selinuxMount flag to enable/disable /etc/selinux host mount

### LINKED ISSUE #2876: nodeplugin failed to mkdir "/etc/selinux"
# Describe the bug #
Unable to update or deploy from fresh the ceph-ccsi-cephfs plugin because on error failed to mkdir "/etc/selinux": mkdir /etc/selinux: read-only file system

# Environment details #

- Image/version of Ceph CSI driver : default for helm chart 3.5.1
- Helm chart version : 3.5.1
- Kernel version : 5.15.16
  `krbd` or `rbd-nbd`) : default
- Kubernetes cluster version : v1.23.1
- Ceph cluster version : 16.2.7

# Steps to reproduce #

1. Setup details: virtualized Talos cluster
1. Deployment to trigger the issue: 
   $ helm install  ceph-csi-cephfs ceph-csi/ceph-csi-cephfs --values values.yaml
      values.yam contains csiConfig, prom scrape, change in configMapName
1. See error
   spec: failed to generate spec: failed to mkdir "/etc/selinux": mkdir /etc/selinux: read-only file system

# Expected behavior #

Probably a flag in the values to disable selinux mounts on non-selinux suupporting distros.

Aste88:

Add selinuxMount flag to enable/disable /etc/selinux host mount inside pods to support selinux-enabled filesystems

Yes, selinuxMount default is true (the previous behaviour)

None I'm aware of

## Related issues ##

Fixes: #2876

### DIFF
--- a/charts/ceph-csi-cephfs/README.md
+++ b/charts/ceph-csi-cephfs/README.md
@@ -156,6 +156,7 @@ charts and their default values.
 | `secret.name`                                  | Specifies the cephFS secret name                                                                                                                     | `csi-cephfs-secret`                                |
 | `secret.adminID`                               | Specifies the admin ID of the cephFS secret                                                                                                          | `<plaintext ID>`                                   |
 | `secret.adminKey`                              | Specifies the key that corresponds to the adminID                                                                                                    | `<Ceph auth key corresponding to ID above>`        |
+| `selinuxMount`                                | Mount the host /etc/selinux inside pods to support selinux-enabled filesystems                                                                                                      | `true`                                            |
 
 ### Command Line
 
--- a/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
+++ b/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
@@ -112,9 +112,11 @@ spec:
               name: host-mount
             - mountPath: /sys
               name: host-sys
+{{- if .Values.selinuxMount }}
             - mountPath: /etc/selinux
               name: etc-selinux
               readOnly: true
+{{- end }}
             - mountPath: /lib/modules
               name: lib-modules
               readOnly: true
@@ -176,9 +178,11 @@ spec:
         - name: host-sys
           hostPath:
             path: /sys
+{{- if .Values.selinuxMount }}
         - name: etc-selinux
           hostPath:
             path: /etc/selinux
+{{- end }}
         - name: host-mount
           hostPath:
             path: /run/mount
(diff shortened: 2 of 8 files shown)


charts/ceph-csi-cephfs/README.md
```diff
@@ -156,6 +156,7 @@ charts and their default values.
 | `secret.name`                                  | Specifies the cephFS secret name                                                                                                                     | `csi-cephfs-secret`                                |
 | `secret.adminID`                               | Specifies the admin ID of the cephFS secret                                                                                                          | `<plaintext ID>`                                   |
 | `secret.adminKey`                              | Specifies the key that corresponds to the adminID                                                                                                    | `<Ceph auth key corresponding to ID above>`        |
+| `selinuxMounnt`                                | Mount the host /etc/selinux inside pods to support selinux-enabled filesystems                                                                                                      | `true`                                            |
```
Madhu-1: `selinuxMounnt` to `selinuxMount`

Madhu-1 (changes requested):
@Aste88 Thanks for the PR. can you fix DCO and commit lint problem. please refer to https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#code-contribution-workflow for any help

Madhu-1 (approved):

Rakshith-R (approved):

humblec:
@Aste88 @Madhu-1 @Rakshith-R @nixpanic I doubt this was the right fix we need , That said, here we are defaulting to `true` and consuming it from the host. If the PODs are running in host which does not support, this could cause issues? or did I miss anything in helm deployment here?

Madhu-1:
@humblec this was done to support the backward compatibility and now the user is having an option to turn it on and off which was not present at all.

humblec:
> @humblec this was done to support the backward compatibility and now the user is having an option to turn it on and off which was not present at all.

Getting a knob done is fine, but a couple of things , like 
1) Default to `true
2) HostPath 

All new deployments will try to pick the hostmount now being this value default to true and if we are on NON selinux system it could cause issues.

Instead of this, I was thinking we could have gone ahead with `DirectoryOrCreate` instead of HostPath.. I believe its the better solution . Please let me know what you folks think.. I can drop a PR accordingly.

Madhu-1:
> > @humblec this was done to support the backward compatibility and now the user is having an option to turn it on and off which was not present at all.
> 
> Getting a knob done is fine, but a couple of things , like
> 
> 1. Default to `true
> 2. HostPath
> 
> All new deployments will try to pick the hostmount now being this value default to true and if we are on NON selinux system it could cause issues.
> 

Yes correctly its the existing behaviour for both helm and deployments with kube templates.

> Instead of this, I was thinking we could have gone ahead with `DirectoryOrCreate` instead of HostPath.. I believe its the better solution . Please let me know what you folks think. I can drop a PR accordingly.

 Is it allowed to create a Directory on `/etc` folder on the nodes? If possible looks good.

PR #2884, opened 2022-02-16
cephfs: Fix umounting not working when the client has been evicted

### LINKED ISSUE #970: When the network fluctuates, the mount will report an error
# Describe the bug #
When the network is suddenly disconnected and then restored, the file mount becomes unavailable. 
**Here is the error report**
> E0423 03:19:03.235751       1 utils.go:161] ID: 59294 GRPC error: rpc error: code = Internal desc = stat /var/lib/kubelet/pods/37124b57-9d0d-4aea-8e42-488db9ae2e3d/volumes/kubernetes.io~csi/pvc-minio-0-copy/mount: permission denied

**the file on a physical machine**
image

# Environment details #

- Image/version of Ceph CSI driver
2.0.1
- Kubernetes cluster version
1.16
- Kernel
4.19.12-1.el7.elrepo.x86_64
# Steps to reproduce #

1. A file is mounted by csi, keeping read and write
2. Shut down the network
3. recover the network

# Expected behavior #

When the network is restored, the file is also restored
or
Manual recovery, but there is no need to restart the machine

klausenbusk:

If the CephFS client is evicted, ex: due to the node being disconnected
from the network for more than 300 seconds[1], the kernel will return
"Permission Denied" when trying to access the mount[2]. In that case we
should just ignore the error and unmount anyways.

This was tested with the kernel's CephFS driver (4.19.202).

[1] https://docs.ceph.com/en/latest/cephfs/eviction/#automatic-client-eviction
[1] https://github.com/torvalds/linux/blob/26291c54e111ff6ba87a164d85d4a4e134b7315c/fs/ceph/mds_client.c#L2809-L2820

This was tested with minikube as described here (`storageClassName` changed to `ceph-filesystem`): https://github.com/ceph/ceph-csi/issues/970#issuecomment-1022227639.

I have only tested this with the kernel's CephFS driver. I'm not sure how the FUSE driver reacts if it gets evicted.

A alternative approach is using the `recover_session=clean` mount option introduced in [Linux v5.4](https://ceph.io/en/news/blog/2020/automatic-cephfs-recovery-after-blacklisting/), in that case the kernel driver would establish a new session.

## Related issues ##

Fixes: #970

## Future concerns ##

In case the client is evicted without k8s triggering `NodeUnpublishVolume`, the mount is "broken" and can't be used. I guess that could happen if the pod hasn't been evicted but the CephFS client has ([`pod-eviction-timeout`](https://kubernetes.io/docs/concepts/architecture/nodes/#condition) higher than [`session_autoclose`](https://docs.ceph.com/en/latest/cephfs/eviction/#automatic-client-eviction)?

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -265,14 +265,19 @@ func (ns *NodeServer) NodeUnpublishVolume(
 	targetPath := req.GetTargetPath()
 	isMnt, err := util.IsMountPoint(targetPath)
 	if err != nil {
-		if os.IsNotExist(err) {
-			// targetPath has already been deleted
-			log.DebugLog(ctx, "targetPath: %s has already been deleted", targetPath)
+		// The ceph module returns "permission denied" if the client has been evicted
+		if os.IsPermission(err) {
+			isMnt = true
+		} else {
+			if os.IsNotExist(err) {
+				// targetPath has already been deleted
+				log.DebugLog(ctx, "targetPath: %s has already been deleted", targetPath)
 
-			return &csi.NodeUnpublishVolumeResponse{}, nil
+				return &csi.NodeUnpublishVolumeResponse{}, nil
+			}
+
+			return nil, status.Error(codes.Internal, err.Error())
 		}
-
-		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if !isMnt {
 		if err = os.RemoveAll(targetPath); err != nil {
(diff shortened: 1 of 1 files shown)


klausenbusk:
I'm not sure what is going wrong in the failed builds. Is it just flakiness?

Madhu-1:
> I'm not sure what is going wrong in the failed builds. Is it just flakiness?

Yes, it's flaky not related to this PR.

humblec:
Thanks for the PR @klausenbusk !

FUSE mounts get  broken and we are fixing or trying to recover it via #2634

gman0:
Just FYI, this is what I'm doing in https://github.com/ceph/ceph-csi/pull/2634 as well.

https://github.com/ceph/ceph-csi/blob/e08007f30073c374923b40c0dddcb95218088a23/internal/cephfs/nodeserver.go#L409-L429

`IsCorruptedMountError` then calls:

https://github.com/kubernetes/kubernetes/blob/df2768123d8d4d3854ec06bb67e352dcb1b4bd08/staging/src/k8s.io/mount-utils/mount_helper_unix.go#L42-L62

gman0:
There are couple more error codes to catch than just `EACCES`, so I'd suggest either using that function for recognizing corrupted mounts I mentioned above, or waiting for #2634.

klausenbusk:
> There are couple more error codes to catch than just `EACCES`, so I'd suggest either using that function for recognizing corrupted mounts I mentioned above, or waiting for #2634.

Oh, I totally missed that PR. Looking at the code it looks like it will indeed solve this issue and offer a more complete fix. Your PR is making good progress, so I'll close this PR.

PR #2893, opened 2022-02-21
add atomic reference tracker

gman0:

RT, reference tracker, is key-based implementation of a reference counter.
Unlike an integer-based counter, RT counts references by tracking unique
keys. This allows accounting in situations where idempotency must be
preserved. It guarantees there will be no duplicit increments or decrements
of the counter.

It is directly needed by the [snapshot-backed shallow volumes for CephFS](https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/cephfs-snapshot-shallow-ro-vol.md#reference-counting-for-shallow-volumes), but is general enough for any other use-cases that need accounting.

Includes unit tests. They use "fake RADOS" interface, but I have run them under real Ceph locally too, and everything worked as expected.

## Related issues ##

Part of https://github.com/ceph/ceph-csi/issues/2142

## Future concerns ##

The RADOS objects handled by this RT are versioned. Should the object layout ever need to change / some backwards-incompatible implementation change, it should be possible to safely do so.

---

<details>

  progress or results

</details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,21 +46,18 @@ endif
 
 GO_PROJECT=github.com/ceph/ceph-csi
 
+CEPH_VERSION ?= $(shell . $(CURDIR)/build.env ; echo $${CEPH_VERSION})
+# TODO: ceph_preview tag may be removed with go-ceph 0.16.0
+GO_TAGS_LIST ?= $(CEPH_VERSION) ceph_preview
+
 # go build flags
 LDFLAGS ?=
 LDFLAGS += -X $(GO_PROJECT)/internal/util.GitCommit=$(GIT_COMMIT)
 # CSI_IMAGE_VERSION will be considered as the driver version
 LDFLAGS += -X $(GO_PROJECT)/internal/util.DriverVersion=$(CSI_IMAGE_VERSION)
+GO_TAGS ?= -tags=$(shell echo $(GO_TAGS_LIST) | tr ' ' ',')
 
 BASE_IMAGE ?= $(shell . $(CURDIR)/build.env ; echo $${BASE_IMAGE})
-
-ifndef CEPH_VERSION
-	CEPH_VERSION = $(shell . $(CURDIR)/build.env ; echo $${CEPH_VERSION})
-endif
-ifdef CEPH_VERSION
-	# pass -tags to go commands (for go-ceph build constraints)
-	GO_TAGS = -tags=$(CEPH_VERSION)
-endif
 
 # passing TARGET=static-check on the 'make containerized-test' or 'make
 # containerized-build' commandline will run the selected target instead of
@@ -109,8 +106,12 @@ mod-check: check-env
 	@echo 'running: go mod verify'
 	@go mod verify && [ "$(shell sha512sum go.mod)" = "`sha512sum go.mod`" ] || ( echo "ERROR: go.mod was modified by 'go mod verify'" && false )
 
-scripts/golangci.yml: build.env scripts/golangci.yml.in
-	sed "s/@@CEPH_VERSION@@/$(CEPH_VERSION)/g" < scripts/golangci.yml.in > scripts/golangci.yml
+scripts/golangci.yml: scripts/golangci.yml.in
+	rm -f scripts/golangci.yml.buildtags.in
+	for tag in $(GO_TAGS_LIST); do \
+		echo "    - $$tag" >> scripts/golangci.yml.buildtags.in ; \
+	done
+	sed "/@@BUILD_TAGS@@/r scripts/golangci.yml.buildtags.in" scripts/golangci.yml.in | sed '/@@BUILD_TAGS@@/d' > scripts/golangci.yml
 
 go-lint: scripts/golangci.yml
 	./scripts/lint-go.sh
--- a/internal/util/reftracker/errors/errors.go
+++ b/internal/util/reftracker/errors/errors.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2022 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package errors
+
+import (
+	goerrors "errors"
+	"fmt"
+
+	"github.com/ceph/go-ceph/rados"
+	"golang.org/x/sys/unix"
+)
+
+// ErrObjectOutOfDate is an error returned by RADOS read/write ops whose
+// rados_*_op_assert_version failed.
+var ErrObjectOutOfDate = goerrors.New("object is out of date since the last time it was read, try again later")
+
+// UnexpectedReadSize formats an error message for a failure due to bad read
+// size.
+func UnexpectedReadSize(expectedBytes, actualBytes int) error {
+	return fmt.Errorf("unexpected size read: expected %d bytes, got %d",
+		expectedBytes, actualBytes)
+}
+
+// UnknownObjectVersion formats an error message for a failure due to unknown
+// reftracker object version.
+func UnknownObjectVersion(unknownVersion uint32) error {
+	return fmt.Errorf("unknown reftracker version %d", unknownVersion)
+}
+
+// FailedObjectRead formats an error message for a failed RADOS read op.
+func FailedObjectRead(cause error) error {
+	if cause != nil {
+		return fmt.Errorf("failed to read object: %w", TryRADOSAborted(cause))
+	}
+
+	return nil
+}
+
+// FailedObjectRead formats an error message for a failed RADOS read op.
+func FailedObjectWrite(cause error) error {
+	if cause != nil {
+		return fmt.Errorf("failed to write object: %w", TryRADOSAborted(cause))
+	}
+
+	return nil
+}
+
+// TryRADOSAborted tries to extract rados_*_op_assert_version from opErr.
+func TryRADOSAborted(opErr error) error {
+	if opErr == nil {
+		return nil
+	}
+
+	var radosOpErr rados.OperationError
+	if !goerrors.As(opErr, &radosOpErr) {
+		return opErr
+	}
+
+	// nolint:errorlint // Can't use errors.As() because rados.radosError is private.
+	errnoErr, ok := radosOpErr.OpError.(interface{ ErrorCode() int })
+	if !ok {
+		return opErr
+	}
+
+	errno := errnoErr.ErrorCode()
+	if errno == -int(unix.EOVERFLOW) || errno == -int(unix.ERANGE) {
+		return ErrObjectOutOfDate
+	}
+
+	return nil
+}
--- a/internal/util/reftracker/radoswrapper/fakerados.go
+++ b/internal/util/reftracker/radoswrapper/fakerados.go
@@ -1,0 +1,551 @@
+/*
+Copyright 2022 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package radoswrapper
+
+import (
+	"fmt"
+
+	"github.com/ceph/go-ceph/rados"
+	"golang.org/x/sys/unix"
+)
+
+type (
+	FakeObj struct {
+		Oid    string
+		Ver    uint64
+		Xattrs map[string][]byte
+		Omap   map[string][]byte
+		Data   []byte
+	}
+
+	FakeRados struct {
+		Objs map[string]*FakeObj
+	}
+
+	FakeIOContext struct {
+		LastObjVersion uint64
+		Rados          *FakeRados
+	}
+
+	FakeWriteOp struct {
+		IoCtx *FakeIOContext
+
+		steps map[fakeWriteOpStepExecutorIdx]fakeWriteOpStepExecutor
+		oid   string
+	}
+
+	FakeReadOp struct {
+		IoCtx *FakeIOContext
+
+		steps map[fakeReadOpStepExecutorIdx]fakeReadOpStepExecutor
+		oid   string
+	}
+
+	fakeWriteOpStepExecutorIdx int
+	fakeReadOpStepExecutorIdx  int
+
+	fakeWriteOpStepExecutor interface {
+		operate(w *FakeWriteOp) error
+	}
+
+	fakeReadOpStepExecutor interface {
+		operate(r *FakeReadOp) error
+	}
+
+	fakeRadosError int
+)
+
+const (
+	fakeWriteOpAssertVersionExecutorIdx fakeWriteOpStepExecutorIdx = iota
+	fakeWriteOpRemoveExecutorIdx
+	fakeWriteOpCreateExecutorIdx
+	fakeWriteOpSetXattrExecutorIdx
+	fakeWriteOpWriteFullExecutorIdx
+	fakeWriteOpRmOmapKeysExecutorIdx
+	fakeWriteOpSetOmapExecutorIdx
+
+	fakeReadOpAssertVersionExecutorIdx fakeReadOpStepExecutorIdx = iota
+	fakeReadOpReadExecutorIdx
+	fakeReadOpGetOmapValuesByKeysExecutorIdx
+)
+
+var (
+	_ IOContextW = &FakeIOContext{}
+
+	// fakeWriteOpStepExecutorOrder defines fixed order in which the write ops are performed.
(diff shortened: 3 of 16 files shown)


gman0:
Please see a somewhat concrete example of how this would be used here https://github.com/ceph/ceph-csi/issues/2142#issuecomment-1047112415.

Makefile
```diff
@@ -46,22 +46,19 @@ endif
 
 GO_PROJECT=github.com/ceph/ceph-csi
 
+CEPH_VERSION ?= $(shell . $(CURDIR)/build.env ; echo $${CEPH_VERSION})
+# TODO: ceph_preview tag may be removed with go-ceph 0.16.0
+GO_TAGS_LIST ?= $(CEPH_VERSION) ceph_preview
```
Madhu-1, line 51: using preview API by default was not accepted see https://github.com/ceph/ceph-csi/pull/2633#discussion_r751977715 we need a stable API as this will be used in production?

internal/util/rt/errors/errors.go
```diff
@@ -0,0 +1,86 @@
+/*
```
Madhu-1: use full name instead of `rt` for the folder name?

internal/util/rt/errors/errors.go
```diff
@@ -0,0 +1,86 @@
...
+*/
+
+package errors
+
+import (
+	goerrors "errors"
+	"fmt"
+
+	"golang.org/x/sys/unix"
+
```
Madhu-1: no new line is required as both are third part library?

internal/util/rt/errors/errors.go
```diff
@@ -0,0 +1,86 @@
...
+		return fmt.Errorf("failed to read RT object: %w", TryRADOSAborted(cause))
+	}
+
+	return nil
+}
+
+// FailedRTObjectRead formats an error message for a failed RADOS read op.
+func FailedRTObjectWrite(cause error) error {
+	if cause != nil {
+		return fmt.Errorf("failed to read RT object: %w", TryRADOSAborted(cause))
```
Madhu-1: Failed to write?

internal/util/rt/radoswrapper/fakerados.go
```diff
@@ -0,0 +1,556 @@
...
+		if err := e.operate(r); err != nil {
+			return err
+		}
+	}
+
+	if obj, err := r.IoCtx.getObj(oid); err == nil {
+		r.IoCtx.LastObjVersion = obj.Ver
+	}
+
+	fmt.Printf("READ %+v\n", r.IoCtx.Rados.Objs)
```
Madhu-1: not required?

internal/util/rt/radoswrapper/fakerados.go
```diff
@@ -0,0 +1,556 @@
...
+limitations under the License.
+*/
+
+package radoswrapper
+
+import (
+	"fmt"
+
+	"golang.org/x/sys/unix"
+
```
Madhu-1: no new line required here

internal/util/rt/v1/add.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	goerrors "errors"
+
+	"github.com/ceph/ceph-csi/internal/util/rt/errors"
+	"github.com/ceph/ceph-csi/internal/util/rt/radoswrapper"
+	"github.com/ceph/ceph-csi/internal/util/rt/reftype"
+)
+
+// Atomically adds refs to an existing RT object.
```
Madhu-1: instead of operation-specific file names can we have everything in one file with interfaces and methods which will help us to mock in when writing the Unit test cases.

Madhu-1 (commented):
Will take a look at it soon.

Makefile
gman0: Here's a list of preview go-ceph APIs that this PR relies on:
* `rados.ReadOp.AssertVersion()`
* `rados.ReadOp.GetOmapValuesByKeys()` and its return value `rados.ReadOpOmapGetValsByKeysStep struct`
* `rados.ReadOp.Read()`
* `rados.WriteOp.AssertVersion()`
* `rados.WriteOp.Remove()`
* `rados.WriteOp.SetXattr()`

These are just bare bindings for librados for functions that are stable -- not really any brand new functionality that would be in a big risk of breaking in following two go-ceph releases (the time needed to graduate an API from preview to stable). Is this not acceptable too?

internal/util/rt/errors/errors.go
gman0: Renaming to `reftracker` because `referencetracker` seems a bit too long I think. Please let me know if you had something else in mind.

internal/util/rt/errors/errors.go
gman0: `goimports` requires packages from the standard library to be separated from other third party packages.

gman0:
```
$ make containerized-test
internal/util/reftracker/errors/errors.go:21: File is not `gofumpt`-ed (gofumpt)
	"fmt"
```

```
$ gofumpt -d errors.go
diff -u errors.go.orig errors.go
--- errors.go.orig	2022-02-22 16:32:31.395203394 +0100
+++ errors.go	2022-02-22 16:32:31.395203394 +0100
@@ -19,6 +19,7 @@
 import (
 	goerrors "errors"
 	"fmt"
+
 	"golang.org/x/sys/unix"
 
 	"github.com/ceph/go-ceph/rados"
```

Makefile
gman0: So how do we do this, can anybody please let me know? Can I continue prototyping other parts of #2142 while this is in review, or are you postponing this PR until `ceph_preview` build tag is no longer needed?

All of these bindings used in this PR are for RADOS primitives, which are very unlikely to change.

Madhu-1: am fine with this one as it was not accepted in earlier PR i mentioned it here. @humblec @nixpanic?

humblec: fine with me too. :+1:

humblec:
lgtm. will wait for others to take final look at this. 

Cc @ceph/ceph-csi-maintainers

internal/util/reftracker/errors/errors.go
```diff
@@ -0,0 +1,86 @@
...
+
+package errors
+
+import (
+	goerrors "errors"
+	"fmt"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/ceph/go-ceph/rados"
```
Madhu-1: no new line is required as both are third party https://github.com/ceph/ceph-csi/blob/devel/internal/util/util.go#L31

internal/util/reftracker/radoswrapper/fakerados.go
```diff
@@ -0,0 +1,552 @@
...
+limitations under the License.
+*/
+
+package radoswrapper
+
+import (
+	"fmt"
+
+	"golang.org/x/sys/unix"
+
```
Madhu-1: same here

internal/util/reftracker/radoswrapper/interface.go
```diff
@@ -0,0 +1,62 @@
...
+	WriteFull(b []byte)
+	SetOmap(pairs map[string][]byte)
+	RmOmapKeys(keys []string)
+	AssertVersion(v uint64)
+
+	Operate(oid string) error
+	Release()
+}
+
+// WriteOpW is a wrapper around rados.ReadOp.
```
Madhu-1: comment need to be updated?

internal/util/reftracker/radoswrapper/interface.go
```diff
@@ -0,0 +1,62 @@
...
+type ReadOpW interface {
+	Read(offset uint64, buffer []byte) *rados.ReadOpReadStep
+	GetOmapValuesByKeys(keys []string) ReadOpOmapGetValsByKeysStepW
+	AssertVersion(v uint64)
+
+	Operate(oid string) error
+	Release()
+}
+
+// WriteOpW is a wrapper around rados.ReadOpOmapGetValsByKeysStep.
```
Madhu-1: comment need to be updated?

internal/util/reftracker/radoswrapper/radoswrapper.go
```diff
@@ -0,0 +1,133 @@
...
+
+var _ IOContextW = &IOContext{}
+
+func NewIOContext(ioctx *rados.IOContext) IOContextW {
+	return &IOContext{
+		IOContext: ioctx,
+	}
+}
+
+func (c *IOContext) GetLastVersion() (uint64, error) {
```
Madhu-1, line 51: Need comments for all the exported methods.

internal/util/reftracker/reftracker.go
```diff
@@ -0,0 +1,162 @@
...
+	ioctx radoswrapper.IOContextW,
+	rtName string,
+	refs map[string]struct{},
+) (created bool, err error) {
+	if err = validateAddInput(rtName, refs); err != nil {
+		return false, err
+	}
+
+	// Read reftracker version.
+
```
Madhu-1, line 139: no new line required

internal/util/reftracker/reftracker.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+
+			return true, nil
+		}
+
+		return false, fmt.Errorf("failed to read reftracker version: %w", err)
+	}
+
+	// Add references to reftracker object.
+
```
Madhu-1, line 155: no new line required.

internal/util/reftracker/reftracker.go
```diff
@@ -0,0 +1,162 @@
...
+	case v1.Version:
+		err = v1.Add(ioctx, rtName, gen, refs)
+		if err != nil {
+			err = fmt.Errorf("failed to add refs: %w", err)
+		}
+	default:
+		err = errors.UnknownObjectVersion(rtVer)
+	}
+
+	return false, err
```
Madhu-1, line 171: should we return `true` in case `v1.Add` is successful?

internal/util/reftracker/reftracker.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Remove atomically removes refs from reference tracker named `rtName`.
+// If the reftracker object holds no references after this removal, the whole
+// object is deleted too. This is also reflected in the return value.
+// If the reftracker object doesn't exist, (true, nil) is returned.
+func Remove(
+	ioctx radoswrapper.IOContextW,
+	rtName string,
+	refs map[string]reftype.RefType,
+) (deleted bool, err error) {
```
Madhu-1: no need for named returns? same in other places also

internal/util/reftracker/version/version_test.go
```diff
@@ -0,0 +1,111 @@
...
+*/
+
+package version
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/ceph/ceph-csi/internal/util/reftracker/radoswrapper"
```
Madhu-1: import order need to be updated as per https://github.com/ceph/ceph-csi/blob/devel/docs/coding.md#imports

internal/util/reftracker/reftracker.go
gman0: I think it helps with readability to be explicit about what this boolean represents. It returns `true` only if the whole object was deleted, not if a reference in the object was removed. I think it's an important distinction, and having it embedded in the function signature makes this a bit more obvious I hope.

The comment of the function also mentions removal and deletion as two separate things:
> If the reftracker object holds no references after this **removal**, the whole
> object is **deleted** too.

I'll leave it for now, but please let me know if you insist on removing it anyway.

internal/util/reftracker/reftracker.go
gman0: Similary to what I've mentioned in https://github.com/ceph/ceph-csi/pull/2893/files/4f7800a48ff563bc7c6d759bdd7a7f9b2fa5bb2a#r831982758, I hoped that using named returns (`(created bool, err error)`) would help users notice what the returned boolean represents. It returns `true` only if the object was created by the `Add()` call. In other cases (when it already exists, and refs are simply added to it) it's not created, so it must return `false`.

internal/util/reftracker/radoswrapper/radoswrapper.go
gman0: I'll add comments in the respective interfaces instead of implementations. Let me know if you want me to add comments to implementations instead.

internal/util/reftracker/reftracker.go
gman0: Comments separated with blank lines above and below are describing sections of code. Without having a new line separating it from the code, it would look like it's describing only that particular line it was adjacent to. I tried to stay consistent with this rule throughout this PR.

I'm leaving it like this for now, please let me know if you insist on changing it.

internal/util/reftracker/reftracker.go
gman0: I'll try to make the comment clearer though.

internal/util/reftracker/radoswrapper/radoswrapper.go
Madhu-1: Yes should be fine.

internal/util/reftracker/reftracker.go
Madhu-1: We dont use named returns anymore (we stopped long back). mixing named/naked returns is not a good idea for the whole repo.

internal/util/reftracker/reftracker.go
Madhu-1: Yes, that helps. return true only if the object is created not for  adding reference.

internal/util/reftracker/radoswrapper/interface.go
```diff
@@ -0,0 +1,106 @@
...
+	SetXattr(name string, value []byte)
+
+	// WriteFull writes a given byte slice as the whole object,
+	// atomically replacing it.
+	WriteFull(b []byte)
+
+	// SetOmap appends the map `pairs` to the omap `oid`.
+	SetOmap(pairs map[string][]byte)
+
+	// RmOmapKeys removes the specified `keys` from the omap `oid`.
```
humblec, line 62: should have been explicit with the name like `RemoveOmapKeys` , this (`RmOmap..`) is close to convey the meaning though.

internal/util/reftracker/reftracker.go
```diff
@@ -0,0 +1,164 @@
...
+//
+// It is stored persistently as a RADOS object, and is safe to use with
+// multiple concurrent writers, and across different nodes of a cluster.
+
+// Add atomically adds refs to reference tracker named `rtName`.
+// If the reftracker object doesn't exist yet, it is created. This is also
+// reflected in the return value.
+func Add(
+	ioctx radoswrapper.IOContextW,
+	rtName string,
```
humblec, line 131: An example `rtName` and `refs` entries in source code comments help too.

internal/util/reftracker/reftracker.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+	default:
+		err = errors.UnknownObjectVersion(rtVer)
+	}
+
+	return deleted, err
+}
+
+var (
+	errNoName = goerrors.New("missing reftracker name")
```
humblec: `errNoRTName` make it clear?

internal/util/reftracker/reftype/reftype.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+
+	"github.com/ceph/ceph-csi/internal/util/reftracker/errors"
+)
+
+// RefType describes type of the reftracker reference.
+type RefType int32
+
+const (
+	refTypeSize = 4
```
humblec: do we need `int32` here ? or lesser size will also suffice ?

humblec:
@gman0 this looks good to me, I got a few minor comments/questions though.

internal/util/reftracker/radoswrapper/interface.go
gman0: All these methods are actually just copy&paste of go-ceph's RADOS interface ([`IOContext.RmOmapKeys`](https://github.com/ceph/go-ceph/blob/6a4d7eb0bdf7872d0890bfaf2a3de70c05d892ed/rados/omap.go#L191-L197)) -- I wanted to keep them exactly as they are in the library since this is just a wrapper. Let me know if this makes sense and is ok to keep it, or you want me to change it.

internal/util/reftracker/reftype/reftype.go
gman0: Yeah I don't suppose there will be much more elements in the type, and `int8` should suffice, so I'll change it to that. Let me know if you disagree.

humblec: sure, if we can suffice with lesser, its good to consume the same, at times little bit of buffer could also help, I am even fine between 8 and 32 ( if at all needed or we anticipate).  The 4 bytes looked heavy. so any lesser size I am fine .. :+1:

internal/util/reftracker/radoswrapper/interface.go
humblec: sure, lets keep with the same/existing name then :+1:

internal/util/reftracker/reftracker.go
gman0: @humblec since `Add()` and `Remove()` manipulate state, I thought it might be useful to also document how the state changes over time, see https://github.com/ceph/ceph-csi/pull/2893/files#diff-69e3db766e37eaa1c0fef3163906ed7e6256e091a512d48dc441dc2c66347f31R42-R123

Let me know if you have any remarks on this.

Madhu-1 (approved):

internal/util/reftracker/reftracker.go
humblec: Thanks for adding that really helpful !! @gman0 :+1:

humblec (approved):

gman0:
Hmm, is the merge stuck?

Rakshith-R:
> Hmm, is the merge stuck?

https://github.com/ceph/ceph-csi/pull/2893/checks?check_run_id=5688271877
It's second in the queue,

Rakshith-R:
@mergifyio refresh 
@mergifyio rebase

gman0:
@mergifyio refresh

PR #2911, opened 2022-03-01
util: configure tenants vaultAuthNamespace if not set

nixpanic:
When a tenant provides a configuration that includes the
`vaultNamespace` option, the `vaultAuthNamespace` option is still taken
from the global configuration. This is not wanted in all cases, as the
`vaultAuthNamespace` option defauls to the `vaultNamespace` option which
the tenant may want to override as well.

The following behaviour is now better defined:

1. no `vaultAuthNamespace` in the global configuration:
   A tenant can override the `vaultNamespace` option and that will also
   set the `vaultAuthNamespace` option to the same value.

2. `vaultAuthNamespace` and `vaultNamespace` in the global configuration:
   When both options are set to different values in the global
   configuration, the tenant `vaultNamespace` option will not override
   the global `vaultAuthNamespace` option. The tenant can configure
   `vaultAuthNamespace` with a different value if required.

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/kms/vault_tokens.go
+++ b/internal/kms/vault_tokens.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ceph/ceph-csi/internal/util/k8s"
 
 	"github.com/hashicorp/vault/api"
+	loss "github.com/libopenstorage/secrets"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


rachael-george:
> 1. no `vaultAuthNamespace` in the global configuration:
>    A tenant can override the `vaultNamespace` option and that will also
>    set the `vaultAuthNamespace` option to the same value.

This sounds good to me

> 2. `vaultAuthNamespace` and `vaultNamespace` in the global configuration:
>    When both options are set to different values in the global
>    configuration, the tenant `vaultNamespace` option will not override
>    the global `vaultAuthNamespace` option. The tenant can configure
>    `vaultAuthNamespace` with a different value if required.

Will the behavior be same if both `vaultAuthNamespace` and `vaultNAmespace` have the same value explicitly  set in the global configmap? The tenant would have to configure each separately to override?

nixpanic:
> Will the behavior be same if both vaultAuthNamespace and vaultNAmespace have the same value explicitly set in the global configmap? The tenant would have to configure each separately to override?

If the global configuration contains both `vaultAuthNamespace` and `vaultNamespace` with the same value, the configuration of the tenant with `vaultNamespace` will override both options. The global configuration will behave the same as if only `vaultNamespace` was set (which causes `vaultAuthNamespace` to be set implicitly).

Madhu-1 (approved):

Rakshith-R (approved):

Issue #1563, opened 2020-10-08
Enable CephFS encryption for CephFS subvolumes


humblec:
# Describe the feature you'd like to have #

CephFS make use of `fscrypt` based encryption or it will be available soon.
Ref# https://lwn.net/Articles/829448/

We have to think about the possibilities of integration for the subvolumes we provision and manage  from CSI.

**CSI requirements can be summarized to below:**

```
*) Each PVC is mapped to a subvolume in the backend
*) Each PVC has to be encrypted with a different key
     I believe in the `fscrypt` case, we will have a custom protector for “CSI” which is common for all the PVCs but will have a “seperate” protector key for each subvolume

*) The keys will be stored in a KMS engine and retrieved at time of mounting the volume to a workload/POD
*) The key retrieved from the KMS engine will be used for encrypting and decrypting the volume

*) Operations like “expansion”, “snapshot” , “Cloning to a new subvolume” on an existing PVC ..etc has to work without hiccups. 
*) Once the snapshot or clone is in place the parent/source volume can be deleted at any time. IOW, no dependency exist to the source/original volume used for snapshot/clone .
```

I would like to capture the thoughts or would like to discuss on  `fscrypt+ cephfs ` implmentation against satisfying above requirements from Ceph CSI side.

@batrick @dillaman @vasyl-purchel @ShyamsundarR @johnStrunk

laptimus:
Thanks Madhu for the prompt reply and reopening this issue.

We are at a point of making critical design choices for our application. We will be going with CepfFS hoping that KMS encryption support will be added soon to the CephFS provisioning and storageclass.

regards

laptimus:
Hello Madhu

Any update on this issue

regards

Madhu-1:
@jtlayton @batrick @vshankar @kotreshhr  CephFS already supports fscrypt based encryption?

jtlayton:
No, it's still a work in progress. We're hoping to have it ready to ship for Quincy release.

laptimus:
Thanks @Madhu-1 @jtlayton

laptimus:
@Madhu-1 @jtlayton - What is the bug number for this issue on tracker.ceph.com

thanks

Madhu-1:
https://tracker.ceph.com/issues/46690 is the one, lets get confirmation from Jeff.

irq0:
Hello!

[...]

PR #2912, opened 2022-03-01
doc: add proposal for CephFS fscrypt integration

irq0:
Add proposal document covering key management integration
of Ceph CSI and https://github.com/google/fscrypt

Updates: #1563
Signed-off-by: Marcel Lauhoff <marcel.lauhoff@suse.com>

### DIFF
--- a/docs/design/proposals/cephfs-fscrypt.md
+++ b/docs/design/proposals/cephfs-fscrypt.md
@@ -1,0 +1,284 @@
+# Ceph Filesystem fscrypt Support
+
+## Problem Description
+
+As a Ceph Container Storage Interface (Ceph CSI) user, I want a cloud native way
+to manage keys and enable encryption on Ceph Filesystem (CephFS) volumes.
+
+In order to access encrypted volumes without Ceph CSI, this can be done by unlocking
+volumes with user space tools.
+
+## Background
+
+*fscrypt* or *FSCrypt* is a Linux Kernel feature that allows the filesystem
+to support the transparent encryption of files and directories. Local
+filesystems like ext4 and F2FS (Flash-Friendly File System) support this feature
+already.
+
+Work is in progress to add fscrypt support to CephFS for filesystem-level encryption.
+
+- [FSCrypt Kernel Documentation](https://www.kernel.org/doc/html/latest/filesystems/fscrypt.html)
+- Management Tools
+  - [`fscrypt`](https://github.com/google/fscrypt)
+  - [`fscryptctl`](https://github.com/google/fscryptctl)
+- [Ceph Feature Tracker: "Add fscrypt support to the kernel CephFS client"](https://tracker.ceph.com/issues/46690)
+- [`fscrypt` design document](https://goo.gl/55cCrI)
+
+**NOTE:** In this document, fscrypt refers to the filesystem-level encryption
+feature, while `fscrypt` specifically refers to the user space tool.
+
+## Terminology
+
+- *FSCrypt*, *fscrypt* - Linux Kernel filesystem-level encryption feature
+- `fscrypt` (code formatted) - User space tool manage keys and encryption policies
+- `fscryptctl` (code formatted) - Low-level user space tool manage
+  keys and encryption policies
+- *subvolume* - CephFS subvolume
+- *unlocking* - Using a key to make an encrypted filesystem
+  accessible in plain text
+- *protector* (fscrypt) - A single method or secret plus data used to
+  derive a *protector key*. Example: user login passphrase
+- *protector key* (fscrypt) - A symmetric key derived from an external
+  source. Used by a *policy* to unwrap a *policy key*
+- *policy key* (fscrypt) - An encryption key passed to the kernel to
+  unlock a directory
+- *policy* (fscrypt) - A collection of directories protected and unlocked as a unit
+- *KMS* - Key management system
+
+## User Visible Change
+
+Similar to the existing RADOS Block Device (RBD) encryption support,
+we propose adding encryption support in the configuration and Key
+Management Service (KMS) integration.
+
+In this example, a user may enable encryption using storage class keys similar
+to RBD. Ceph CSI then configures and unlocks the persistent volumes and CephFS
+subvolumes.
+
+Due to the way `fscrypt` stores metadata, subvolumes have a regular root
+directory containing  a `/.fscrypt` directory and a
+`/ceph-csi-encrypted` directory. The first contains `fscrypt` metadata; the
+latter is the fscrypt-enabled directory made that is accessible to pods.
+
+Example configuration using a secrets-based KMS:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cephfs-storage-encryption-secret
+stringData:
+  encryptionPassphrase: verysecretpassword
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-cephfs-sc-encrypted
+provisioner: cephfs.csi.ceph.com
+parameters:
+  clusterID: <cluster-id>
+  fsName: cephfs
+
+  encrypted: "true"
+  encryptionKMSID: "user-ns-secrets-metadata"
+
+  csi.storage.k8s.io/provisioner-secret-name: csi-cephfs-secret
+  csi.storage.k8s.io/provisioner-secret-namespace: default
+  csi.storage.k8s.io/controller-expand-secret-name: csi-cephfs-secret
+  csi.storage.k8s.io/controller-expand-secret-namespace: default
+  csi.storage.k8s.io/node-stage-secret-name: csi-cephfs-secret
+  csi.storage.k8s.io/node-stage-secret-namespace: default
+reclaimPolicy: Delete
+allowVolumeExpansion: true
+mountOptions:
+  - debug
+```
+
+The change will leverage the existing Ceph CSI KMS and support any
+integration now available to RBD encryption
+
+## Implementation
+
+We suggest to leverage the encryption features in Ceph CSI and integrate
+that with `fscrypt`, a Go tool, and the library for key management and configuration
+of the fscrypt kernel feature.
+
+Ceph CSI and `fscrypt` have a lot of overlap between their key management
+features. The Key Management section will go into detail on how and where keys
+are managed.
+
+- Ceph CSI provides the user facing configuration and access to key
+  management systems
+- `fscrypt` handles key derivation, storage of wrapped keys and metadata
+
+The current CephFS subvolume root will remain untouched with the exception that
+the subvolume root is not bind mounted into the pod, but rather a well-known
+subdirectory. The root will contain a `/.fscrypt` directory managed by `fscrypt`.
+
+`fscrypt` requires access to a mounted filesystem and therefore the encryption setup
+must take place in the `NodeStageVolume` request handler instead of `CreateVolume`.
+This is the same case for RBD. The set up will take place right between
+subvolume mount and bind mount to the container namespace.
+
+Additional checks after unlocking will ensure that a container operates on an
+unlocked encrypted directory and never on directory that has fscrypt enabled.
+
+### Key Management
+
+```mermaid
+graph LR
+    vault[Vault, Default] -->|data encryption key| csi_kms_integrated
+    secrets[Passphrase<br>K8s secrets, AWS, IBM Key Protect] -->|passphrase| csi_kms_metadata
+
+
+    subgraph ceph_csi_kms[Ceph CSI KMS]
+        csi_kms_metadata['metadata' type DEK store]
+        csi_kms_integrated['integrated' type DEK store]
+
+    end
+
+
+    subgraph protector[fscrypt protector]
+        custom_passphrase[CustomPasswordSource];
+        raw_passphrase[RawKeySource];
+
+        protector_metadata[(/.fscrypt/protectors)];
+
+        protector_metadata -->|wrapped key| custom_passphrase;
+        protector_metadata -->|wrapped key| raw_passphrase;
+
+    end
+
+    subgraph policy[fscrypt policy]
+        policy_unwrap
+        policy_metadata[(/.fscrypt/policies)]
+
+        policy_metadata -->|wrapped key| policy_unwrap
+    end
+
+
+    csi_kms_metadata -->|DEK| custom_passphrase
+    csi_kms_integrated -->|DEK| raw_passphrase
+
+    custom_passphrase -->|protector key| policy_unwrap
+    raw_passphrase -->|protector key| policy_unwrap
+
+    policy_unwrap -->|policy key| kernel[Kernel API]
+
+```
+
+The diagram shows the keys flowing from Ceph CSI to the Kernel API unlocking
+a directory. On the way, key material from Ceph CSI passes two key derivation
+steps in fscrypt:
+
+- *protectors* and
+- *policies*
+
+`fscrypt` supports multiple *protectors*. These may source secrets from login
+passwords, custom passwords or soon Ceph CSI. Unlocking a protector
+yields a *protector key* that is then used to unlock a *policy*.
+
+A *policy* may unlock multiple directories. In our case there will be
+only a single policy for a single well-known directory on the subvolume root.
+A policy is used to derive a *policy key*, which is passed to the Kernel API
+along with other settings, such as the desired encryption algorithm.
+
+Going back to the beginning of the diagram and looking at the interface between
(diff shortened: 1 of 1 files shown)


humblec:
Cc @jtlayton @kotreshhr  appreciated your view on this proposal to integrate CephFS encryption with CSI.

docs/design/proposals/cephfs-fscrypt.md
```diff
@@ -0,0 +1,276 @@
...
+
+Ceph CSI and `fscrypt` have a lot of overlap between their key management
+features. The Key Management section will go into detail on how and where keys
+are managed.
+
+  - Ceph CSI provides the user facing configuration and access to key
+    management systems
+  - `fscrypt` handles key derivation, storage of wrapped keys and metadata
+
+The current CephFS subvolume root will remain untouched with the exception that
```
humblec, line 114: are we referring to SubVoumeGroup here as `subvolume root` or its the `filesystem` ?

docs/design/proposals/cephfs-fscrypt.md
```diff
@@ -0,0 +1,276 @@
...
+the subvolume root is not bind mounted into the pod, but rather a well-known
+subdirectory. The root will contain a `/.fscrypt` directory managed by `fscrypt`.
+
+`fscrypt` requires access to a mounted filesystem and therefore the encryption setup
+must take place in the `NodeStageVolume` request handler instead of `CreateVolume`.
```
humblec, line 119: Dont we need to create a `protector key` at time of `createvolume` and store ?

docs/design/proposals/cephfs-fscrypt.md
```diff
@@ -0,0 +1,276 @@
...
+
+To be compatible with `fscrypt`, this directory requires support as
+well. The downside of this is that we lose the CephFS subvolume root to
+metadata and encrypted data will reside under a well-known
+subdirectory (for example, `/ceph-csi-encrypted`).
+
+
+## Dependencies
+
+The proposed change is tailored to CephFS and requires CephFS support
```
humblec, line 221: Looks like one of the CephFS string ( from ...and requires.... ) has to be replaced...

humblec:
@irq0 first of all this is a well formatted/written design doc , Thanks for that.. :+1: 
I have a couple of comments on the design or proposal 
Also, can we also list the dependencies or prerequisites to support this feature in Ceph CSI? 

ps # the diagram is not rendering properly for me, so I may be missed the `peotector key` generation part which I asked in one comment.

docs/design/proposals/cephfs-fscrypt.md
irq0: The protector keys are stored wrappend in the '/.fscrypt' directory on the volume. We therefore need filesystem access which we don't have in CreateVolume, hence NodeStageVolume.

docs/design/proposals/cephfs-fscrypt.md
irq0: Subvolume root refers to the root directory of the subvolume. What gets mapped into a pod once mounted.

On my testsetup this would be:
```
$ bin/ceph fs subvolume info a csi-vol-4fa8245c-9b00-11ec-bdd8-eed58c1c7c89 csi
{
...
    "path": "/volumes/csi/csi-vol-4fa8245c-9b00-11ec-bdd8-eed58c1c7c89/32a6c9c4-d83a-482b-8abf-e6b0ea676f3a",
...
}

$ sudo mount -t ceph -o name=admin,secret=.. 192.168.122.1:40687:/volumes/csi/csi-vol-4fa8245c-9b00-11ec-bdd8-eed58c1c7c89/32a6c9c4-d83a-482b-8abf-e6b0ea676f3a /subvolume_root 
```

docs/design/proposals/cephfs-fscrypt.md
irq0: A lot of CephFS in that sentence :).  It tries to express that this is change is agnostic to CephFS, except for the fact that it is part of Ceph CSI's CephFS integration.

irq0:
> @irq0 first of all this is a well formatted/written design doc , Thanks for that.. +1 I have a couple of comments on the design or proposal Also, can we also list the dependencies or prerequisites to support this feature in Ceph CSI?

Added runtime and build dependencies to the Dependencies section

> 
> ps # the diagram is not rendering properly for me, so I may be missed the `peotector key` generation part which I asked in one comment.

Odd. Maybe the mermaid live editor works better:

https://mermaid.live/edit#pako:eNqVVE1v4yAQ_SuIS3Ylt72ukqpSm_QQuZWiWuoe7MiiMGmsxIAAb2S1_e8LxgnBdbVbLNnD8N58PDBvmAoGeIpfFZFb9PBUcGTHH9LsTf7s3glawMYZa3RxcfPOiCEIOFWtNJXgaAftO6K6Kne1LituwAYywHwYDVSB0fmKaC23imi4flE36S99XEnQ7e8sQcu7R5RCi1ZKGKB9JnkihQQ1GOIqKLh7uhTNiy-dgtyWPS6f2wmaZ0uUPmZrD3RjGCafHK0JMq0EtLhPkTZCwQgpNJdPgj1C9FTgbKRK6TsUKt_oTsPgOU_Z2Fh1GRTI553H6XgQimWiURTWs8BQ5HAOfyIHq2eABeApXRDhx9VlX8zVaVX__Bet26ODbUoC60_BsOrZ9_hxD7OBkLGMYl_RNmjYTc8E9I6y4S7BJ_do526pAtt31HZM-FzzSKJo44cnrotgD8uIWjEhnLBAiQU6pRgG8n_PUfAv64yj_QfJ0yKnZ3UeT9mB4rDP0-6DbldLKydOcA2qJhWz98ybi1Jgs4UaCjy1JvPXS4EL_mGhjbQ6wT2rbBl4uiF7DQkmjRFZyymeGtXAEbSoiJWo7lEffwFbzbJp

docs/design/proposals/cephfs-fscrypt.md
```diff
@@ -0,0 +1,284 @@
...
+provisioner: cephfs.csi.ceph.com
+parameters:
+  clusterID: <cluster-id>
+  fsName: cephfs
+
+  encrypted: "true"
+  secretName: "cephfs-storage-encryption-secret"
+  encryptionKMSID: "user-ns-secrets-metadata"
+  secretNamespace: "default"
+  encryptionKMSType: "metadata"
```
Rakshith-R:
```suggestion
  encrypted: "true"
  encryptionKMSID: "user-ns-secrets-metadata"
```

The rest of the config will go into a configmap like [csi-kms-connection-details.yaml](https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/csi-kms-connection-details.yaml) or [kms-config.yaml](https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/kms-config.yaml)

I think we can reuse the entire [kms package](https://github.com/ceph/ceph-csi/tree/devel/internal/kms) for cephfs too.

Rakshith-R (commented):
jfyi There is a feature request on cephfs for metadata capability similar to rbd 
https://tracker.ceph.com/issues/54472

humblec:
Waiting for CephFS team input on this l.. hopefully we will have it soon! thanks !

docs/design/proposals/cephfs-fscrypt.md
```diff
@@ -0,0 +1,281 @@
...
+xattrs on a mounted CephFS filesystem.
+
+A prototype showing this approach is available: https://github.com/irq0/ceph-csi/tree/wip/fscrypt
+
+Benefits:
+- Simpler key wrapping
+- No `/.fscrypt` on the subvolume root
+
+Drawbacks
+- `fscryptctl` is a C tool and does not lend itself to be integrated into Ceph CSI
```
jtlayton, line 261: fwiw: `fscryptctl` is very simple and just calls a bunch of ioctls. You could probably drive those ioctls from Go as well.

jtlayton (commented):
This all looks reasonable to me. I'm afraid I don't have enough intimate knowledge of Ceph CSI to know what the best method to use is.

humblec:
@irq0 can you please correct the CI linter failure ?

docs/design/proposals/cephfs-fscrypt.md
```diff
@@ -0,0 +1,281 @@
...
+Similar to the existing RADOS Block Device (RBD) encryption support,
+we propose adding encryption support in the configuration and Key Management Service (KMS)
+integration.
+
+In this example, a user may enable encryption using storage class keys similar
+to RBD. Ceph CSI then configures and unlocks the persistent volumes and CephFS
+subvolumes.
+
+Due to the way `fscrypt` stores metadata, subvolumes have a regular root
+directory containing  a `/.fscrypt` directory and a
```
jtlayton, line 59: FWIW, the `.fscrypt` directory is one of the main reasons I wanted the CSI team to think about this. This directory is not generally an issue with local filesystems like ext4, as admins almost always mount the root of the fileystem, and you can generally ensure that this directory is available to userland applications.

Contrast that with something like NFS or Ceph, where mounting a subdirectory of an export is very common. You could end up in a situation where you've run `fscrypt setup` on an upper-level directory but then someone mounts a subdirectory of it. The `.fscrypt` directory won't be available on the client at that point.

The placement of the .fscrypt directory is crucial if you intend to use the fscrypt binary under the hood. I wonder if it may even be better to do something like store the info that's currently in `.fscrypt` directly in a RADOS objects instead, which would sort of sidestep the whole issue. That would mean a major overhaul for fscrypt, however (or a rewrite).

humblec:
@irq0 can you please revisit the comments from Jeff and the linter failures?

docs/design/proposals/cephfs-fscrypt.md
irq0: Yes, it is not a big drawback. Using the policy functions from fscrypt would also be an option. (https://github.com/google/fscrypt/blob/master/metadata/policy.go)

docs/design/proposals/cephfs-fscrypt.md
irq0: I'll split the anwer into the .fscrypt directory and RADOS object alternative.

I think we should limit the discussion to the Ceph CSI context.
We basically have two options:
1. Integrate Ceph CSI with the `fscrypt` tooling as is (this design
   doc)
2. Create another key management option that is specific to CephFS
   (Alternative [ceph-csi-ksm]; Fscrypt Metadata on RADOS below)

- The `.fscrypt` directory is necessary for 1.
- 1. and 2. can coexist on the same filesystem, since they are
  user land key management systems that ultimately set a fscrypt
  policy in the kernel. 

I'm in favor with 1. Why? It would enable users to manually get a
secret from K8s secrets or Vault, mount a CephFS somewhere and use the
`fscrypt` tool without changes with that secret to unlock the
encrypted data. No Ceph specific tools required.

## The .fscrypt metadata directory and Ceph CSI

To make things a bit more concrete, here is what a freshly created
PV/PVC looks like on an otherwise empty CephFS with an implementation
based on the doc:

```
# mount -t ceph -o name=admin,secret=$(bin/ceph auth get-key client.admin) \
   $(bin/ceph mon dump --format json | jq -r '.mons[] | .addr' | sed -e 's/\/0//'):/ \
   /mnt
# find /mnt
/mnt
/mnt/volumes
/mnt/volumes/_csi:csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd.meta
/mnt/volumes/csi
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/703b0bb0-0d92-4c98-a1d2-ef943a4260ad
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/703b0bb0-0d92-4c98-a1d2-ef943a4260ad/.fscrypt
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/703b0bb0-0d92-4c98-a1d2-ef943a4260ad/.fscrypt/protectors
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/703b0bb0-0d92-4c98-a1d2-ef943a4260ad/.fscrypt/protectors/46ef26343bfaa137
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/703b0bb0-0d92-4c98-a1d2-ef943a4260ad/.fscrypt/policies
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/703b0bb0-0d92-4c98-a1d2-ef943a4260ad/.fscrypt/policies/451ca267026dd6704b85efdccef305f8
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/703b0bb0-0d92-4c98-a1d2-ef943a4260ad/ceph-csi-encrypted
/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/.meta
```

This would add a `.fscrpyt` directory on each subvolume / PVC. To
manually use `fscrypt` a user would have to mount
`/mnt/volumes/csi/csi-vol-cab93367-afff-11ec-89f2-620077d1dfcd/703b0bb0-0d92-4c98-a1d2-ef943a4260ad`

Since Ceph CSI owns the subvolumes below CSI anyway, I don't think
there is a drawback to this.

We could move the `.fscrypt` directory as high as the filesystem or
the csi subvolume group. Basically having it next to the "meta" files.

## Fscrypt Metadata on RADOS
An implementation could be as straight forward as using two objects,
one for protectors one for policies. Each with an object map mapping
keyid to a fscrypt metadata blob. The entries would directly correspond to the files in
/.fscrypt/{protectors,policies}/$keyid.

The problem would be access control in the spirit of `fscrypt`.
`/.fscrypt` is basically set up like `/tmp` (sticky). Created by root,
but non-root users are allowed to add/remove their own protectors.

The fscrypt design doc lists the following requirements for a metadata store:

> Metadata Requirements
> There are a few properties that we want our metadata storage to have.
>
> For a filesystem that is set up for encryption, a user can create an
> encrypted directory (and its associated metadata) without being root
>
> Any user with access to an encrypted directory (via standard UNIX
> permissions) and the correct credentials can unlock the directory,
> regardless of who set up the directory.
>
> A non-root user cannot delete another user’s metadata, which would
> make the files corresponding to that metadata unreadable (see above
> Threat Model).
>
> An encrypted directory can be protected with a Protector whose data
> is on another filesystem. This is necessary to protect a folder on a
> USB drive with a user’s login password, for example.
>
> We do NOT require that any user be able to set up encryption on a
> filesystem, as this involves making privileged changes to the system.

I'm not familiar enough with the Ceph FS POSIX user translation to
RADOS auth, but I suspect user access to the fscrypt metadata
objects won't be easy.

I agree, this is a major overhaul or a new Ceph FS specific key management solution.

jtlayton: > The problem would be access control in the spirit of fscrypt. /.fscrypt is basically set up like /tmp (sticky). Created by root, but non-root users are allowed to add/remove their own protectors.

Good point. That would be tricky to replicate in bare RADOS with cephx creds. You'd probably have to layer some enforcement on top. Blech.

That said...one thing that the local fscrypt filesystem folks don't need to worry about is idmapping. I imagine consistent uid/gid mapping is a must for k8s nodes so I'll assume that's not a problem here.

> We could move the .fscrypt directory as high as the filesystem or the csi subvolume group. Basically having it next to the "meta" files.

The key point here is that the `.fscrypt` dir has to be consistently reachable, so you need to consistently mount the same cephfs directory. If you're dealing with a pretty flat hierarchy where every tenant is mounting his own subvolume, you should be ok. If, on the other hand, you have a situation where some clients mount at a higher point in the directory tree then things get more iffy.

I imagine Ceph CSI keeps things fairly flat though, so you should be OK.

Having multiple fscrypt directories (one for each subvolume) seems like the best approach, IMO. Allowing access to the keys in there should be safe (since they are just part of the overall KDF), but it'd be better to deny access to it when tenants don't need it.

nixpanic (approved):
Thanks for joining todays standup call @irq0!

As discussed during the call, it would be a great start to have this function on RBD volumes with ext4. Once CephFS in the kernel supports this, it can be enabled for CephFS too. In the mean time, including and stabilizing fscrypt within Ceph-CSI can already get started.

Note that our e2e suite uses minikube, and hence the minikube-iso should have a kernel with fscrypt enabled ([minikube kernel config](https://github.com/kubernetes/minikube/blob/master/deploy/iso/minikube-iso/board/minikube/x86_64/linux_x86_64_defconfig)).

humblec (approved):

Issue #2913, opened 2022-03-01
Provisioning of NFS exported volumes


nixpanic:
# Describe the feature you'd like to have #

It would be useful to have volumes provisioned that can be exported over NFS. Ceph already has the ability to [manage NFS-Ganesha and export CephFS directories (or subvolumes)](https://docs.ceph.com/en/latest/mgr/nfs/).

# What is the value to the end user? (why is it a priority?) #

This would allow the following use-cases to be addressed:

- clients that can not reach the OSDs/MONs, but could use a single TCP-port to connect to the Ceph managed NFS-server
- worker-nodes that do not have support for CephFS (NFS is more standard)

# How will we know we have a good solution? (acceptance criteria) #

Minimally the following CSI operations should be supported:

- `CreateVolume`
- `DeleteVolume`

# Proof of Concept #

A [work-in-progress NFS-provisioner](https://github.com/nixpanic/ceph-csi/tree/wip/nfs-provisioner) is available, with a container image at `quay.io/nixpanic/cephcsi:nfs`. Changes/additions are mostly under `internal/nfs/` and `examples/nfs/`. Scripts and .yaml deployment files are included in the branch for testing and deploying with Rook. (Testing is done on OpenShift Data Foundation, as those deployments are readily available to me.)

# Tasks #

- [x] design document: #2914
- [x] NFS-provisioner component: #2948
- [x] deployment artifacts: #2980
- [x] e2e testing: #3036
- [ ] validate error handling, CSI-procedures should be idempotent
  - [ ] `CreateVolume`
  - [x] `DeleteVolume`: #3071
- [x] use go-ceph API instead of `ceph` command: #2977, ceph/go-ceph#655
- [ ] documentation
- [ ] deployment with helm
- [ ] examples: #2963

Rakshith-R:
items are still yet to be done.

PR #2914, opened 2022-03-01
doc: proposal for providing NFS-exports

nixpanic:
The design for NFS-provising of volumes explains how the existing NFS
CSI Node-Plugin can be used, and what the steps are to create the
NFS-export with Ceph Mgr.

Updates: #2913

---

<details>

  progress or results

</details>

### DIFF
--- a/docs/design/proposals/nfs.md
+++ b/docs/design/proposals/nfs.md
@@ -1,0 +1,70 @@
+# Dynamic provisioning of NFS volumes
+
+Ceph has [support for NFS-Ganesha to export directories][ceph_mgr_nfs] on
+CephFS. This can be used to export CephFS based volumes over NFS.
+
+## Node-Plugin for mounting NFS-exports
+
+The Kubernetes CSI community provides and maintains a [NFS CSI][nfs_csi]
+driver. This driver can be used as a Node-Plugin so that NFS CSI volumes can be
+mounted. When a CSI volume has the `server` and `share`
+[parameters][nfs_csi_params], the Node-Plugin will be able to mount the
+NFS-export.
+
+## Exporting CephFS based volumes over NFS
+
+Ceph-CSI already creates CephFS volumes, that can be mounted over the native
+CephFS protocol. A new provisioner in Ceph-CSI can create CephFS volumes, and
+include the required [NFS CSI parameters][nfs_csi_params] so that the [NFS CSI
+driver][nfs_csi] can mount the CephFS volume over NFS.
+
+The provisioner that handles the CSI requests for NFS volume can call the [Ceph
+Mgr commands to export/unexport][ceph_mgr_nfs] the CephFS volumes. The CephFS
+volumes would be internally managed by the NFS provisioner, and only be exposed
+as NFS CSI volumes towards the consumers.
+
+### `CreateVolume` CSI operation
+
+When the Ceph-CSI NFS provisioner is requested to create a NFS CSI volume, the
+following steps need to be taken:
+
+1. create a CephFS volume, use the CephFS `CreateVolume` call or other internal
+   API
+1. call the Ceph Mgr API to export the CephFS volume with NFS-Ganesha
+1. return the NFS CSI volume, with `server` and `share` parameters (other
+   parameters that are useful for CephFS volume management may be kept)
+
+The 2nd step requires a NFS-cluster name for the Ceph Mgr call(s). The name of
+the NFS-cluster as managed by Ceph should be provided in the parameters of the
+`CreateVolume` operation. For Kubernetes that means the parameters is set as an
+option in the `StorageClass`.
+
+The `server` parameter of the volume is an other option that is managed by the
+Ceph (or Rook) infrastructure. This parameter is also required to be provided
+in the `CreateVolume` parameters.
+
+Removing the NFS-export for the volume (or other operations) requires the name
+of the NFS-cluster, as it is needed for the Ceph Mgr API. Like other parameters
+of the CephFS volume, it will be needed to store the NFS-cluster name in the
+OMAP journalling.
+
+### `DeleteVolume` CSI operation
+
+The `DeleteVolume` operation only receives the `volume_id` parameter, which
+is to be used by the CSI Controller (provisioner) to locate the backing volume.
+The `DeleteVolume` operation for the CephFS provisioner already knows how to
+delete volumes by ID.
+
+In order to remove the exported volume from the NFS-cluster, the operation
+needs to fetch the name of the NFS-cluster from the journal where it was stored
+during `CreateVolume`.
+
+### Additional CSI operations
+
+`CreateVolume` and `DeleteVolume` are the only required operations for the CSI
+Controller (provisioner). Additional features as they are supported by CephFS
+can forward the operations to the CephFS provisioner at a later time.
+
+[ceph_mgr_nfs]: https://docs.ceph.com/en/latest/mgr/nfs/
+[nfs_csi]: https://github.com/kubernetes-csi/csi-driver-nfs
+[nfs_csi_params]: https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/docs/driver-parameters.md


docs/design/proposals/nfs.md
```diff
@@ -0,0 +1,65 @@
...
+When the Ceph-CSI NFS provisioner is requested to create a NFS CSI volume, the
+following steps need to be taken:
+
+1. create a CephFS volume, use the CephFS `CreateVolume` call or other internal
+   API
+1. call the Ceph Mgr API to export the CephFS volume with NFS-Ganesha
+1. return the NFS CSI volume, with `server` and `share` parameters (other
+   parameters that are useful for CephFS volume management may be kept)
+
+The 2nd step requires a NFS-cluster name for the Ceph Mgr call(s). The name of
```
humblec, line 37: In above listing we only have step 1 @nixpanic

docs/design/proposals/nfs.md
```diff
@@ -0,0 +1,65 @@
...
+the NFS-cluster as managed by Ceph should be provided in the parameters of the
```
humblec, line 38: Can we have some reference to NFS-cluster here?

docs/design/proposals/nfs.md
```diff
@@ -0,0 +1,65 @@
...
+`CreateVolume` operation. For Kubernetes container orchistrators that means the
```
humblec: I think either we have to mention just `Kubernetes` or use `Kubernetes container orchetrator`

docs/design/proposals/nfs.md
```diff
@@ -0,0 +1,65 @@
...
+parameters is set as an option in the `StorageClass`.
+
+The `server` parameter of the volume is an other option that is managed by the
+Ceph (or Rook) infrastructure. This parameter is also required to be provided
+in the `CreateVolume` parameters.
```
humblec, line 44: What this `server` refers to ? if its mon IP there is a chance for the dynamic change . Or is it a static IP which always clients can connect to ? @nixpanic

humblec:
@nixpanic apart from the comments above, one thing which I see missing here is the UID/GID/root squash permission confifguration .. How is it handled ? can we mention that too ?

docs/design/proposals/nfs.md
nixpanic: it lists 3x `1.`  as that is the required formatting for lint-markdown.

docs/design/proposals/nfs.md
nixpanic: `server` is the parameter as mentioned for the NFS CSI NodePlugin, it should be the hostname (or IP-addresse) to the Ceph managed NFS-cluster.

docs/design/proposals/nfs.md
nixpanic: it is the 1st link on line 3. Ceph (optionally through Rook) will set that up.

docs/design/proposals/nfs.md
nixpanic: It is probably more like `Kubernetes based container orchestrators`?

docs/design/proposals/nfs.md
humblec: Yeah, I am trying to get clarification on which host name it refers ? because in general case there may be a VIP in front which can cope up with some back end failures.  Or is it the case that, `server` is not an option coming from SC at all in any stage? rather filled by Ceph CSI based on Ceph mgr return to export CephFS share as NFS server?  @nixpanic

docs/design/proposals/nfs.md
humblec: well, it says `step 2` below and no reference to that, cant we change the number `1` to have an order ?

docs/design/proposals/nfs.md
Madhu-1: as it kubernetes based I would call it `Kubernetes`

docs/design/proposals/nfs.md
```diff
@@ -0,0 +1,65 @@
...
+### `DeleteVolume` CSI operation
+
+The `DeleteVolume` operation only receives the `volume_id` parameter, which
+is to be used by the CSI Controller (provisioner) to locate the backing volume.
+The `DeleteVolume` operation for the CephFS provisioner already knows how to
+delete volumes by ID.
+
+Removing the NFS-export for the volume requires the name of the NFS-cluster, as
+it is needed for the Ceph Mgr API. Like other parameters of the CephFS volume,
+it will be needed to store the NFS-cluster name in the OMAP journalling.
```
Madhu-1: Mention this in CreateVolume to store this information?

docs/design/proposals/nfs.md
Rakshith-R: image

it looks fine when it is rendered.

humblec: @Rakshith-R thats interesting, I was going through without rendering and noticed all belongs to "1" :thinking:

docs/design/proposals/nfs.md
nixpanic: No, Ceph-CSI should not deploy the NFS-cluster, that is something Rook or the Ceph administrator should do. Ceph-CSI will just use Ceph Mgr to add new exported volumes to the existing Ceph managed NFS-Ganesha(s).

The `server` is therefore configured by Rook or an other operator (automated or human). It refers to the hostname that the NFS-cluster (or single NFS-Ganesha instance) is reachable by. A VIP, or some other load-balanced high-available mechanism is the same for the mounting clients, they do not care.

docs/design/proposals/nfs.md
nixpanic: yes, will do that

nixpanic:
> @nixpanic apart from the comments above, one thing which I see missing here is the UID/GID/root squash permission confifguration .. How is it handled ? can we mention that too ?

What does CSI need to do with UID/GID/root squashing? Containers usually get their own mapped UID/GIDs, just like on other filesystems.

Rakshith-R (approved):

humblec:
> > @nixpanic apart from the comments above, one thing which I see missing here is the UID/GID/root squash permission confifguration .. How is it handled ? can we mention that too ?
> 
> What does CSI need to do with UID/GID/root squashing? Containers usually get their own mapped UID/GIDs, just like on other filesystems.

what I was trying to confirm here is the root or normal access on the share once exported, have seen that, the NFS shares permission denied errors as the volume access granted from server based on the UID/GID settings. so wondering is it a volume exported with all access in place or restricted to some userID/GIDs ?

nixpanic:
> > > @nixpanic apart from the comments above, one thing which I see missing here is the UID/GID/root squash permission confifguration .. How is it handled ? can we mention that too ?
> > 
> > 
> > What does CSI need to do with UID/GID/root squashing? Containers usually get their own mapped UID/GIDs, just like on other filesystems.
> 
> what I was trying to confirm here is the root or normal access on the share once exported, have seen that, the NFS shares permission denied errors as the volume access granted from server based on the UID/GID settings. so wondering is it a volume exported with all access in place or restricted to some userID/GIDs ?

There is no special configuration needed as such. It will just be mounted as root through the node-plugin, after which the permissions (uid/gid) for the container are configured. This is works the same as for CephFS, for all I know.

Madhu-1 (commented):
LGTM. @humblec do you have any concerns on this one?

humblec:
> > > > @nixpanic apart from the comments above, one thing which I see missing here is the UID/GID/root squash permission confifguration .. How is it handled ? can we mention that too ?
> > > 
> > > 
> > > What does CSI need to do with UID/GID/root squashing? Containers usually get their own mapped UID/GIDs, just like on other filesystems.
> > 
> > 
> > what I was trying to confirm here is the root or normal access on the share once exported, have seen that, the NFS shares permission denied errors as the volume access granted from server based on the UID/GID settings. so wondering is it a volume exported with all access in place or restricted to some userID/GIDs ?
> 
> There is no special configuration needed as such. It will just be mounted as root through the node-plugin, after which the permissions (uid/gid) for the container are configured. This is works the same as for CephFS, for all I know.

ok, lets give a try with default configurations in place, in case if there are permission issues, we can look into this further.

docs/design/proposals/nfs.md
humblec: well, this server IP or server address is static and can not be changed, so the concern . If upper layers manage it properly, it should work though.

humblec (approved):

Madhu-1 (approved):

PR #2933, opened 2022-03-10
cephfs: go with default permissions while creating subvolumes

humblec:
While creating subvolumes, CephFS driver set the mode to `777`
and pass it along to go ceph apis which cause the subvolume
permission to be on 777, however if we create a subvolume
directly in the ceph cluster, the default permission bits are
set which is 755 for the subvolume. This commit try to stick
to the default behaviour even while creating the subvolume via
CSI driver.

### DIFF
--- a/internal/cephfs/core/volume.go
+++ b/internal/cephfs/core/volume.go
@@ -39,12 +39,6 @@ import (
 // taken through this additional cluster information.
 var clusterAdditionalInfo = make(map[string]*localClusterState)
 
-const (
-	// modeAllRWX can be used for setting permissions to Read-Write-eXecute
-	// for User, Group and Other.
-	modeAllRWX = 0o777
-)
-
 // Subvolume holds subvolume information. This includes only the needed members
 // from fsAdmin.SubVolumeInfo.
 type Subvolume struct {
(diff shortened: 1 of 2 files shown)


internal/cephfs/core/volume.go
```diff
@@ -231,7 +231,6 @@ func (s *subVolumeClient) CreateVolume(ctx context.Context) error {
 
 	opts := fsAdmin.SubVolumeOptions{
 		Size: fsAdmin.ByteCount(s.Size),
-		Mode: modeAllRWX,
```
Madhu-1, line 234: remove `modeAllRWX` variable which is of no use as we are not setting the mode anymore.

humblec: Sure @Madhu-1 ..

Madhu-1:
>Mar 10 15:44:06.999: INFO: stdErr occurred: /bin/sh: /target/testing: Permission denied
Mar 10 15:44:06.999: FAIL: failed to validate normal user CephFS pvc and application binding: failed to exec command in pod: command terminated with exit code 1

@humblec with this change normal user application is failing to access the mount

Madhu-1:
>Mar 10 15:44:06.999: INFO: stdErr occurred: /bin/sh: /target/testing: Permission denied
Mar 10 15:44:06.999: FAIL: failed to validate normal user CephFS pvc and application binding: failed to exec command in pod: command terminated with exit code 1

@humblec with this change normal user application is failing to access the mount

humblec:
Yep, have to go through this in detail and revert.. thanks.

PR #2945, opened 2022-03-16
build: remove cache while building container image

humblec:
Reduce size of the container image by removing the cache in deploy
    and devel container.

Additional Note: Noticed around 120MB size reduction

### DIFF
--- a/deploy/cephcsi/image/Dockerfile
+++ b/deploy/cephcsi/image/Dockerfile
@@ -33,6 +33,8 @@ RUN dnf -y install \
 	/usr/bin/cc \
 	make \
 	git \
+	&& dnf clean all \
+	&& rm -rf /var/cache/yum \
     && true
 
 ENV GOROOT=${GOROOT} \
--- a/scripts/Dockerfile.devel
+++ b/scripts/Dockerfile.devel
@@ -29,6 +29,8 @@ RUN dnf -y install \
 	librados-devel \
 	librbd-devel \
     && dnf -y update \
+    && dnf clean all \
+    && rm -rf /var/cache/yum \
     && true
 
 WORKDIR "/go/src/github.com/ceph/ceph-csi"


Madhu-1:
>Reduce size of the container image

what is the size its getting reduced to?

nixpanic:
This is the `:devel` container-image, what is the benefit of the reduced size? The image itself is not available in a registry, nobody downloads it.

humblec:
> This is the `:devel` container-image, what is the benefit of the reduced size? The image itself is not available in a registry, nobody downloads it.

@nixpanic both has been pushed. ptal . thanks

deploy/cephcsi/image/Dockerfile
```diff
@@ -33,6 +33,8 @@ RUN dnf -y install \
 	/usr/bin/cc \
 	make \
 	git \
+	&& dnf clean all \
+	&& rm -rf /var/cache/yum \
```
nixpanic, line 37: This is just a temporary image, only used for compiling the executable. This cleanup does not have any effect on the resulting image. In fact, it will just slow things down a little...

scripts/Dockerfile.devel
```diff
@@ -29,6 +29,8 @@ RUN dnf -y install \
 	librados-devel \
 	librbd-devel \
     && dnf -y update \
+    && dnf clean all \
+    && rm -rf /var/cache/yum \
```
nixpanic, line 33: This image is not pushed to any (public) registry. Only locally on developers machines and in the CI. This is fine if the savings are important.

nixpanic (changes requested):
> > This is the `:devel` container-image, what is the benefit of the reduced size? The image itself is not available in a registry, nobody downloads it.
> 
> @nixpanic both has been pushed. ptal . thanks

We still do not push these reduced size images anywhere. Please describe clearly what the benefit is (developers machines only?).

deploy/cephcsi/image/Dockerfile
humblec: considering we do have only one layer ( as its in same RUN command) with reduced size build layer in the image,  I expect there is a small effect. why you say no effect in this context and it can slow things down?

scripts/Dockerfile.devel
humblec: I am not sure what is referred as `savings` here, if its about the packages in cache , I would like to ask, why we need that?

deploy/cephcsi/image/Dockerfile
nixpanic: The only advantage to add the extra commands to clean the DNF caches, is that (potentially) cached layers on a build machine can be re-used. It could save some storage (as it uses the cached layer) on a developers machine, but automated build environments will rebuild all layers from the same Dockerfile anyway.

So, is the time that `dnf clean all` needs to cleanup worth the few (?) MBs of storage in a developers environment?

scripts/Dockerfile.devel
nixpanic: We don't need it, but I have not experienced any issues by having it either. As this is a final image for developers, cleaning the packages is fine IMHO. Not sure if deleting `/var/cache/yum` causes issues in case someone wants to manually run the container and install some other tools for extra debugging?

humblec: The extra debugging in container land mostly by attaching the debugging container and doing stuffs, so I dont think thats has to be considered here.

deploy/cephcsi/image/Dockerfile
humblec: I think so.. more or less, keeping a cache inside a container layer is not a practice in general.

humblec:
@nixpanic can you please revisit this PR ?

nixpanic (approved):
This is fine by me. The commit message could be clearer, as it only reduces the size of a temporary layer in the deploy container-image, and removes the DNF cache from the `:devel` container-image. This is something that only helps developers, and is not user visible.

Madhu-1 (approved):

yati1998 (approved):

Rakshith-R:
@mergifyio refresh
@mergifyio rebase

Rakshith-R:
@mergifyio refresh

nixpanic:
Building the container-image worked, and most CI jobs have passed now. Added labels to skip further CI jobs, and automatic retry.

nixpanic:
> /retest ci/centos/mini-e2e-helm/k8s-1.23

For some reason, this did not get picked up?

nixpanic:
@Mergifyio refresh

nixpanic:
@Mergifyio refresh

humblec:
Looks like Megifyio is not helping here. :(

humblec:
@Mergifyio refresh

Rakshith-R:
@mergifyio rebase

PR #2953, opened 2022-03-21
ci: add release 3.5 image details to the mirror image list

humblec:
The release 3.5 image details were missing and this commit add the
same.

### DIFF
--- a/mirror/images.txt
+++ b/mirror/images.txt
@@ -9,18 +9,12 @@
 # ceph-csi devel
 docker.io/rook/ceph:v1.8.2	rook/ceph:v1.8.2
 
-
 # ceph-csi v3.5
-docker.io/ceph/ceph:v16		ceph/ceph:v16
-docker.io/rook/ceph:v1.6.2	rook/ceph:v1.6.2
-
-
-# ceph-csi v3.4
-docker.io/ceph/ceph:v16		ceph/ceph:v16
+quay.io/ceph/ceph:v16		ceph/ceph:v16
 docker.io/rook/ceph:v1.6.2	rook/ceph:v1.6.2
 
 # ceph-csi v3.3
-docker.io/ceph/ceph:v15		ceph/ceph:v15
+quay.io/ceph/ceph:v15		ceph/ceph:v15
 docker.io/rook/ceph:v1.4.9	rook/ceph:v1.4.9
 
 # e2e testing
--- a/mirror/images.txt
+++ b/mirror/images.txt
@@ -8,6 +8,7 @@
 
 # ceph-csi devel
 docker.io/rook/ceph:v1.8.2	rook/ceph:v1.8.2
+quay.io/ceph/ceph:v17		ceph/ceph:v17
 
 # ceph-csi v3.5
 quay.io/ceph/ceph:v16		ceph/ceph:v16
--- a/mirror/images.txt
+++ b/mirror/images.txt
@@ -9,6 +9,12 @@
 # ceph-csi devel
 docker.io/rook/ceph:v1.8.2	rook/ceph:v1.8.2
 
+
+# ceph-csi v3.5
+docker.io/ceph/ceph:v16		ceph/ceph:v16
+docker.io/rook/ceph:v1.6.2	rook/ceph:v1.6.2
+
+
 # ceph-csi v3.4
 docker.io/ceph/ceph:v16		ceph/ceph:v16
 docker.io/rook/ceph:v1.6.2	rook/ceph:v1.6.2


humblec:
@nixpanic not sure how ci was passing without the 3.5 entries in the release 3.5 branch. 

Also one thing noted here is, this image references images from docker hub instead of quay.io . If that need to be changed as well, please mention, I can add the same to this PR.

humblec:
> @nixpanic not sure how ci was passing without the 3.5 entries in the release 3.5 branch.
> 
> Also one thing noted here is, this image references images from docker hub instead of quay.io . If that need to be changed as well, please mention, I can add the same to this PR.

I have added that too as a second commit..

humblec:
> > @nixpanic not sure how ci was passing without the 3.5 entries in the release 3.5 branch.
> > Also one thing noted here is, this image references images from docker hub instead of quay.io . If that need to be changed as well, please mention, I can add the same to this PR.
> 
> I have added that too as a second commit..

In the scripts of CI its referring completely to `docker.io` , may be its kept that way to be in parity with `rook` images?

mirror/images.txt
```diff
@@ -9,12 +9,18 @@
 # ceph-csi devel
 docker.io/rook/ceph:v1.8.2	rook/ceph:v1.8.2
 
+
```
nixpanic: please remove the additional empty lines. Just a single empty line between section suffices.

mirror/images.txt
```diff
@@ -9,12 +9,18 @@
...
+# ceph-csi v3.5
+quay.io/cephcsi/ceph:v16	ceph/ceph:v16
```
nixpanic: this looks incorrect, we do not use `quay.io/cephcsi/`, the images are located under `quay.io/ceph/`?

mirror/images.txt
```diff
@@ -9,12 +9,18 @@
...
+docker.io/rook/ceph:v1.6.2	rook/ceph:v1.6.2
+
+
 # ceph-csi v3.4
-docker.io/ceph/ceph:v16		ceph/ceph:v16
+quay.io/cephcsi/ceph:v16	ceph/ceph:v16
```
nixpanic: there is no need to duplicate things, that will probably cause images to pushed to the CI mirror multiple times (and overwrites), so it just causes extra network traffic.

nixpanic (changes requested):
> In the scripts of CI its referring completely to `docker.io` , may be its kept that way to be in parity with `rook` images?

Ceph did not push to Quay.io when we started to use the mirroring. Docker Hub has the pull restrictions, Quay.io does not.

mirror/images.txt
humblec: sure, removed duplicated instances too @nixpanic

mirror/images.txt
nixpanic: you marked this as resolved, but I still see the additional empty lines in this commit?

mirror/images.txt
```diff
@@ -9,18 +9,12 @@
...
 
-
 # ceph-csi v3.5
-docker.io/ceph/ceph:v16		ceph/ceph:v16
-docker.io/rook/ceph:v1.6.2	rook/ceph:v1.6.2
-
-
-# ceph-csi v3.4
-docker.io/ceph/ceph:v16		ceph/ceph:v16
+quay.io/ceph/ceph:v16		ceph/ceph:v16
```
nixpanic, line 14: please keep the values aligned (with tabs)

mirror/images.txt
```diff
@@ -8,6 +8,8 @@
 
 # ceph-csi devel
 docker.io/rook/ceph:v1.8.2	rook/ceph:v1.8.2
+quay.io/ceph/ceph:v17		ceph/ceph:v17
+
```
nixpanic, line 12: again an unneeded empty line, please remove those.

nixpanic (changes requested):

mirror/images.txt
humblec: not sure there is an inconsistency with tabs , however it looks to be rendering correctly to me https://github.com/ceph/ceph-csi/blob/511ae20d106ff5e3fd08ec2ed533e9b6e357c555/mirror/images.txt

nixpanic: thanks, seems the GitHub webui/PR does not show it correctly for some reason.

nixpanic (approved):
The commits still add the extra empty lines, and a next one removes them. Not very clean, but the result looks good now.

Madhu-1 (approved):

PR #2972, opened 2022-03-29
rbd: increase force promote timeout to 2 minutes

Madhu-1:
Increase the timeout to 2 minutes to give enough time for rollback to complete.
As rollback is performed by the force-promote command it, at times, may take more than a minute (based on dirty blocks that need to be rolled back approximately) to rollback.

The added extra 1 minute is useful though to avoid multiple calls to complete the rollback and in extreme corner cases to avoid failures in the first instance of the call when the mirror watcher is not yet removed (post scaling down the RBD mirror instance)

More context at https://bugzilla.redhat.com/show_bug.cgi?id=2009735#c33. 

### DIFF
--- a/internal/rbd/mirror.go
+++ b/internal/rbd/mirror.go
@@ -88,8 +88,8 @@ func (ri *rbdImage) promoteImage(force bool) error {
 	return nil
 }
 
-// forcePromoteImage promotes image to primary with force option with 1 minute
-// timeout. If there is no response within 1 minute,the rbd CLI process will be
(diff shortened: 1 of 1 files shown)


Madhu-1:
Skipping E2E as there is no E2E for Replication and this part of code will not get hit in the normal workflow.

ShyamsundarR (approved):

yati1998 (approved):

nixpanic (approved):

PR #2979, opened 2022-03-31
ci: correct prefix path for centos mirroring repository

humblec:
as a solution to the image referencing issue, we are prefixing
the string `quay.io` as a tag in mirror image list.


nixpanic (approved):
Thanks! All non-docker.io registries should probably use this format. It is required for images that are specified in a deployment with a registry-prefix.

yati1998 (approved):

PR #2983, opened 2022-03-31
util: add support to run commands with nsenter

Madhu-1:
add support to run rbd map and mount -t commands with the nsenter.

The complete design of the pod/multus network is added here https://github.com/rook/rook/blob/master/design/ceph/multus-network.md#csi-pods

WIP Rook design PR https://github.com/rook/rook/pull/9903

cc @leseb

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -50,6 +50,42 @@ kubectl replace -f ./csi-config-map-sample.yaml
 Storage class and snapshot class, using `<cluster-id>` as the value for the
 option `clusterID`, can now be created on the cluster.
 
+## Running CephCSI with pod networking
+
+The current problem with Pod Networking, is when a CephFS/RBD volume is mounted
+in a pod using Ceph CSI and then the CSI CephFS/RBD plugin is restarted or
+terminated (e.g. by restarting or deleting its DaemonSet), all operations on
+the volume become blocked, even after restarting the CSI pods.
+
+The only workaround is to restart the node where the Ceph CSI plugin pod was
+restarted. This can be mitigated by running the `rbd map`/`mount -t` commands
+in a different network namespace which does not get deleted when the CSI
+CephFS/RBD plugin is restarted or terminated.
+
+If someone wants to run the CephCSI with the pod networking they can still do
+by setting the `netNamespaceFilePath`. If this path is set CephCSI will execute
+the `rbd map`/`mount -t` commands after entering the [network
+namespace](https://man7.org/linux/man-pages/man7/network_namespaces.7.html)
+specified by `netNamespaceFilePath` with the
+[nsenter](https://man7.org/linux/man-pages/man1/nsenter.1.html) command.
+
+`netNamespaceFilePath` should point to the network namespace of some
+long-running process, typically it would be a symlink to
+`/proc/<long running process id>/ns/net`.
+
+The long-running process can also be another pod which is a Daemonset which
+never restarts. This Pod should only be stopped and restarted when a node is
+stopped so that volume operations do not become blocked. The new DaemonSet pod
+can contain a single container, responsible for holding its pod network alive.
+It is used as a passthrough by the CephCSI plugin pod which when mounting or
+mapping will use the network namespace of this pod.
+
+Once the pod is created get its PID and create a symlink to
+`/proc/<PID>/ns/net` in the hostPath volume shared with the csi-plugin pod and
+specify the path in the `netNamespaceFilePath` option.
+
+*Note* This Pod should have `hostPID: true` in the Pod Spec.
+
 ## Deploying the storage class
 
 Once the plugin is successfully deployed, you'll need to customize
--- a/internal/cephfs/mounter/fuse.go
+++ b/internal/cephfs/mounter/fuse.go
@@ -65,12 +65,20 @@ func mountFuse(ctx context.Context, mountPoint string, cr *util.Credentials, vol
 	if volOptions.FsName != "" {
 		args = append(args, "--client_mds_namespace="+volOptions.FsName)
 	}
+	var (
+		stderr string
+		err    error
+	)
 
-	_, stderr, err := util.ExecCommand(ctx, "ceph-fuse", args[:]...)
+	if volOptions.NetNamespaceFilePath != "" {
+		_, stderr, err = util.ExecuteCommandWithNSEnter(ctx, volOptions.NetNamespaceFilePath, "ceph-fuse", args[:]...)
+	} else {
+		_, stderr, err = util.ExecCommand(ctx, "ceph-fuse", args[:]...)
+	}
+
 	if err != nil {
 		return fmt.Errorf("%w stderr: %s", err, stderr)
 	}
-
 	// Parse the output:
 	// We need "starting fuse" meaning the mount is ok
 	// and PID of the ceph-fuse daemon for unmount
--- a/internal/cephfs/mounter/kernel.go
+++ b/internal/cephfs/mounter/kernel.go
@@ -51,7 +51,16 @@ func mountKernel(ctx context.Context, mountPoint string, cr *util.Credentials, v
 
 	args = append(args, "-o", optionsStr)
 
-	_, stderr, err := util.ExecCommand(ctx, "mount", args[:]...)
+	var (
+		stderr string
+		err    error
+	)
+
+	if volOptions.NetNamespaceFilePath != "" {
+		_, stderr, err = util.ExecuteCommandWithNSEnter(ctx, volOptions.NetNamespaceFilePath, "mount", args[:]...)
+	} else {
+		_, stderr, err = util.ExecCommand(ctx, "mount", args[:]...)
+	}
 	if err != nil {
 		return fmt.Errorf("%w stderr: %s", err, stderr)
 	}
--- a/internal/rbd/rbd_attach.go
+++ b/internal/rbd/rbd_attach.go
@@ -457,8 +457,17 @@ func createPath(ctx context.Context, volOpt *rbdVolume, device string, cr *util.
 		mapArgs = append(mapArgs, "--read-only")
 	}
 
-	// Execute map
-	stdout, stderr, err := util.ExecCommand(ctx, cli, mapArgs...)
+	var (
+		stdout string
+		stderr string
+		err    error
+	)
+
+	if volOpt.NetNamespaceFilePath != "" {
+		stdout, stderr, err = util.ExecuteCommandWithNSEnter(ctx, volOpt.NetNamespaceFilePath, cli, mapArgs...)
+	} else {
+		stdout, stderr, err = util.ExecCommand(ctx, cli, mapArgs...)
+	}
 	if err != nil {
 		log.WarningLog(ctx, "rbd: map error %v, rbd output: %s", err, stderr)
 		// unmap rbd image if connection timeout
--- a/internal/util/cephcmds.go
+++ b/internal/util/cephcmds.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"os/exec"
 	"time"
 
@@ -31,6 +32,47 @@ import (
 
 // InvalidPoolID used to denote an invalid pool.
 const InvalidPoolID int64 = -1
+
+// ExecuteCommandWithNSEnter executes passed in program with args with nsenter
+// and returns separate stdout and stderr streams. In case ctx is not set to
+// context.TODO(), the command will be logged after it was executed.
+func ExecuteCommandWithNSEnter(ctx context.Context, netPath, program string, args ...string) (string, string, error) {
+	var (
+		sanitizedArgs = StripSecretInArgs(args)
+		stdoutBuf     bytes.Buffer
+		stderrBuf     bytes.Buffer
+	)
+
+	// check netPath exists
+	if _, err := os.Stat(netPath); err != nil {
+		return "", "", fmt.Errorf("failed to get stat for %s %w", netPath, err)
+	}
+	//  nsenter --net=%s -- <program> <args>
+	args = append([]string{fmt.Sprintf("--net=%s", netPath), "--", program}, args...)
+	cmd := exec.Command("nsenter", args...) // #nosec:G204, commands executing not vulnerable.
+
+	cmd.Stdout = &stdoutBuf
+	cmd.Stderr = &stderrBuf
+
+	err := cmd.Run()
+	stdout := stdoutBuf.String()
+	stderr := stderrBuf.String()
+
+	if err != nil {
+		err = fmt.Errorf("an error (%w) occurred while running %s args: %v", err, program, sanitizedArgs)
+		if ctx != context.TODO() {
+			log.UsefulLog(ctx, "%s", err)
+		}
+
+		return stdout, stderr, err
+	}
+
+	if ctx != context.TODO() {
+		log.UsefulLog(ctx, "command succeeded: %s %v", program, sanitizedArgs)
+	}
+
+	return stdout, stderr, nil
+}
 
 // ExecCommand executes passed in program with args and returns separate stdout
 // and stderr streams. In case ctx is not set to context.TODO(), the command
--- a/internal/util/csiconfig.go
+++ b/internal/util/csiconfig.go
@@ -49,6 +49,8 @@ type ClusterInfo struct {
 		// SubvolumeGroup contains the name of the SubvolumeGroup for CSI volumes
 		SubvolumeGroup string `json:"subvolumeGroup"`
 	} `json:"cephFS"`
+	// symlink filepath for the network namespace where we need to execute commands.
+	NetNamespaceFilePath string `json:"netNamespaceFilePath"`
 }
 
 // Expected JSON structure in the passed in config file is,
@@ -160,4 +162,13 @@ func GetClusterID(options map[string]string) (string, error) {
 	}
 
 	return clusterID, nil
+}
+
+func GetNetNamespaceFilePath(pathToConfig, clusterID string) (string, error) {
+	cluster, err := readClusterInfo(pathToConfig, clusterID)
+	if err != nil {
+		return "", err
+	}
+
+	return cluster.NetNamespaceFilePath, nil
 }
--- a/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
+++ b/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
@@ -31,6 +31,7 @@ spec:
       priorityClassName: {{ .Values.nodeplugin.priorityClassName }}
 {{- end }}
       hostNetwork: true
+      hostPID: true
       # to use e.g. Rook orchestrated cluster, and mons' FQDN is
       # resolved through k8s service, set dns policy to cluster first
       dnsPolicy: ClusterFirstWithHostNet
--- a/charts/ceph-csi-cephfs/values.yaml
+++ b/charts/ceph-csi-cephfs/values.yaml
@@ -27,6 +27,7 @@ serviceAccounts:
 #       - "<MONValue2>"
 #     cephFS:
 #       subvolumeGroup: "csi"
+#     netNamespaceFilePath: "{{ .kubeletDir }}/plugins/{{ .driverName }}/net"
 csiConfig: []
 
 # Set logging level for csi containers.
--- a/charts/ceph-csi-rbd/values.yaml
+++ b/charts/ceph-csi-rbd/values.yaml
@@ -25,6 +25,7 @@ serviceAccounts:
 #     monitors:
 #       - "<MONValue1>"
 #       - "<MONValue2>"
+#     netNamespaceFilePath: "{{ .kubeletDir }}/plugins/{{ .driverName }}/net"
 csiConfig: []
 
 # Configuration details of clusterID,PoolID and FscID mapping
--- a/deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
+++ b/deploy/cephfs/kubernetes/csi-cephfsplugin.yaml
@@ -15,6 +15,7 @@ spec:
       serviceAccountName: cephfs-csi-nodeplugin
       priorityClassName: system-node-critical
       hostNetwork: true
+      hostPID: true
       # to use e.g. Rook orchestrated cluster, and mons' FQDN is
       # resolved through k8s service, set dns policy to cluster first
       dnsPolicy: ClusterFirstWithHostNet
--- a/examples/csi-config-map-sample.yaml
+++ b/examples/csi-config-map-sample.yaml
@@ -20,6 +20,12 @@ kind: ConfigMap
 # NOTE: Make sure you don't add radosNamespace option to a currently in use
 # configuration as it will cause issues.
 # The field "cephFS.subvolumeGroup" is optional and defaults to "csi".
+# The <netNamespaceFilePath#> fields are the various network namespace
+# path for the Ceph cluster identified by the <cluster-id>, This will be used
+# by the CSI plugin to execute the rbd map/unmap and mount -t commands in the
+# network namespace specified by the <netNamespaceFilePath#>.
+# If a CSI plugin is using more than one Ceph cluster, repeat the section for
+# each such cluster in use.
 # NOTE: Changes to the configmap is automatically updated in the running pods,
 # thus restarting existing pods using the configmap is NOT required on edits
 # to the configmap.
@@ -37,6 +43,7 @@ data:
       {
         "clusterID": "<cluster-id>",
         "radosNamespace": "<rados-namespace>",
+        "netNamespaceFilePath": "<kubeletRootPath>/plugins/rbd.csi.ceph.com/net",
         "monitors": [
           "<MONValue1>",
           "<MONValue2>",
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -126,6 +126,11 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 	defer volOptions.Destroy()
 
+	volOptions.NetNamespaceFilePath, err = util.GetNetNamespaceFilePath(util.CsiConfigFile, volOptions.ClusterID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
 	mnt, err := mounter.New(volOptions)
 	if err != nil {
 		log.ErrorLog(ctx, "failed to create mounter for volume %s: %v", volID, err)
--- a/internal/cephfs/store/volumeoptions.go
+++ b/internal/cephfs/store/volumeoptions.go
@@ -50,6 +50,8 @@ type VolumeOptions struct {
 	ProvisionVolume    bool   `json:"provisionVolume"`
 	KernelMountOptions string `json:"kernelMountOptions"`
 	FuseMountOptions   string `json:"fuseMountOptions"`
+	// Network namespace file path to execute nsenter command
+	NetNamespaceFilePath string
 
 	// conn is a connection to the Ceph cluster obtained from a ConnPool
 	conn *util.ClusterConnection
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -332,6 +332,10 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 	defer rv.Destroy()
 
+	rv.NetNamespaceFilePath, err = util.GetNetNamespaceFilePath(util.CsiConfigFile, rv.ClusterID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
 	if isHealer {
 		err = healerStageTransaction(ctx, cr, rv, stagingParentPath)
 		if err != nil {
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -156,6 +156,8 @@ type rbdVolume struct {
 	LogStrategy        string
 	VolName            string
 	MonValueFromSecret string
+	// Network namespace file path to execute nsenter command
+	NetNamespaceFilePath string
 	// RequestedVolSize has the size of the volume requested by the user and
 	// this value will not be updated when doing getImageInfo() on rbdVolume.
 	RequestedVolSize   int64
--- a/internal/util/csiconfig_test.go
+++ b/internal/util/csiconfig_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package util
 
 import (
+	"encoding/json"
 	"os"
 	"testing"
 )
@@ -136,5 +137,71 @@ func TestCSIConfig(t *testing.T) {
 	err = os.WriteFile(basePath+"/"+csiClusters, []byte(data), 0o600)
 	if err != nil {
 		t.Errorf("Test setup error %s", err)
+	}
+}
+
+func TestGetNetNamespaceFilePath(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name      string
+		clusterID string
+		want      string
+	}{
+		{
+			name:      "get NetNamespaceFilePath for cluster-1",
+			clusterID: "cluster-1",
+			want:      "/var/lib/kubelet/plugins/rbd.ceph.csi.com/cluster1-net",
+		},
+		{
+			name:      "get NetNamespaceFilePath for cluster-2",
+			clusterID: "cluster-2",
+			want:      "/var/lib/kubelet/plugins/rbd.ceph.csi.com/cluster2-net",
+		},
+		{
+			name:      "when NetNamespaceFilePath is empty",
+			clusterID: "cluster-3",
+			want:      "",
+		},
+	}
+
+	csiConfig := []ClusterInfo{
+		{
+			ClusterID:            "cluster-1",
+			Monitors:             []string{"ip-1", "ip-2"},
+			NetNamespaceFilePath: "/var/lib/kubelet/plugins/rbd.ceph.csi.com/cluster1-net",
+		},
+		{
+			ClusterID:            "cluster-2",
+			Monitors:             []string{"ip-3", "ip-4"},
+			NetNamespaceFilePath: "/var/lib/kubelet/plugins/rbd.ceph.csi.com/cluster2-net",
+		},
+		{
+			ClusterID: "cluster-3",
+			Monitors:  []string{"ip-5", "ip-6"},
+		},
+	}
+	csiConfigFileContent, err := json.Marshal(csiConfig)
+	if err != nil {
+		t.Errorf("failed to marshal csi config info %v", err)
+	}
+	tmpConfPath := t.TempDir() + "/ceph-csi.json"
+	err = os.WriteFile(tmpConfPath, csiConfigFileContent, 0o600)
+	if err != nil {
+		t.Errorf("failed to write %s file content: %v", CsiConfigFile, err)
+	}
+	for _, tt := range tests {
+		ts := tt
+		t.Run(ts.name, func(t *testing.T) {
+			t.Parallel()
+			got, err := GetNetNamespaceFilePath(tmpConfPath, ts.clusterID)
+			if err != nil {
+				t.Errorf("GetNetNamespaceFilePath() error = %v", err)
+
+				return
+			}
+			if got != ts.want {
+				t.Errorf("GetNetNamespaceFilePath() = %v, want %v", got, ts.want)
+			}
+		})
 	}
 }


nixpanic:
A much easier way is to run the whole `cephcsi` executable (from the daemonset) with `nsenter` as prefix. That way only Rook needs to adapt the `.yaml` files.

internal/util/cephcmds.go
```diff
@@ -32,6 +32,38 @@ import (
 // InvalidPoolID used to denote an invalid pool.
 const InvalidPoolID int64 = -1
 
+func ExecCommandWithSenter(ctx context.Context, netPath, program string, args ...string) (string, string, error) {
```
nixpanic: name of the function should probably be `ExecuteCommandWithNSEnter` (missing `N`).

cmd/cephcsi.go
```diff
@@ -131,6 +131,7 @@ func init() {
 	flag.BoolVar(&conf.Version, "version", false, "Print cephcsi version information")
 	flag.BoolVar(&conf.EnableProfiling, "enableprofiling", false, "enable go profiling")
 
+	flag.BoolVar(&conf.SupportNsenter, "supportnsenter", false, "Use nsenter to map and mount -t operation")
```
nixpanic: Instead of a `bool`, I would prefer to have an option like `--network-pid=${pid}` which then translates to a namespace-path like `/proc/${pid}/ns/net`. This should make it easy to pass the parameter as an (environment) variable, and no need to add a new option in the configuration file/ConfigMap.

internal/cephfs/driver.go
```diff
@@ -123,6 +123,7 @@ func (fs *Driver) Run(conf *util.Config) {
 			log.FatalLogMsg(err.Error())
 		}
 		fs.ns = NewNodeServer(fs.cd, conf.Vtype, topology)
+		fs.ns.SupportNsenter = conf.SupportNsenter
```
nixpanic: I'd recommend to use `NS` as capitals, here `SupportNSEnter`.

nixpanic (changes requested):
Actually, there will be a different network namespace on each worker node. So there is some configuration needed that contains the name of the worker node and the PID of the long-running process.

That means, it will not be easy to pass the PID on the commandline in a deamonset, as it is different per worker node. Maybe the configuration should be placed in a ConfigMap, and the key in the ConfigMap is the hostname of the worker-node, so that the PID can be obtained from the ConfigMap as environment variable and passed on to the `cephcsi` executable? Not sure of that is possible though...

Madhu-1:
> A much easier way is to run the whole `cephcsi` executable (from the daemonset) with `nsenter` as prefix. That way only Rook needs to adapt the `.yaml` files.

As you already noticed this will not works if multiple ceph clusters are in different multus network that is the reason for adding it to the CSI configmap. the admin need to add network path to the configmap and cephcsi will use it to execute commands.

cmd/cephcsi.go
Madhu-1: This is just to say we support nsenter or not and the actual configuration will be fetched from the CSI configmap which has the clusterID and mon mapping. This helps to support multiple ceph clusters in different multus network

Madhu-1:
> Actually, there will be a different network namespace on each worker node. So there is some configuration needed that contains the name of the worker node and the PID of the long-running process.
> 
> That means, it will not be easy to pass the PID on the commandline in a deamonset, as it is different per worker node. Maybe the configuration should be placed in a ConfigMap, and the key in the ConfigMap is the hostname of the worker-node, so that the PID can be obtained from the ConfigMap as environment variable and passed on to the `cephcsi` executable? Not sure of that is possible though...



> Actually, there will be a different network namespace on each worker node. So there is some configuration needed that contains the name of the worker node and the PID of the long-running process.
> 
> That means, it will not be easy to pass the PID on the commandline in a deamonset, as it is different per worker node. Maybe the configuration should be placed in a ConfigMap, and the key in the ConfigMap is the hostname of the worker-node, so that the PID can be obtained from the ConfigMap as environment variable and passed on to the `cephcsi` executable? Not sure of that is possible though...

the smylink will be created from the holder pod on each of the node hostpath to the known path and that information will be added to the configmap and cephcsi will fetch the path from the configmap and use that to execute the nsenter command.

Madhu-1:
Validation is done with the below steps

# After using the nsenter

* get the po -o wide output and note the IP of the holder pod and the rbdplugin pod (it should be pod ip)

```
$ kuberc get po -owide
NAME                                            READY   STATUS      RESTARTS        AGE   IP              NODE       NOMINATED NODE   READINESS GATES
csi-cephfsplugin-mq4t7                          3/3     Running     1 (22h ago)     23h   192.168.39.33   minikube   <none>           <none>
csi-cephfsplugin-provisioner-5dc9cbcc87-x496h   6/6     Running     4 (6h52m ago)   23h   10.244.0.28     minikube   <none>           <none>
csi-plugin-holder-tjfpv                         1/1     Running     0               15m   10.244.0.37     minikube   <none>           <none>
csi-rbdplugin-provisioner-d47c55d7c-qpqcp       6/6     Running     4 (6h52m ago)   24h   10.244.0.26     minikube   <none>           <none>
csi-rbdplugin-ztktc                             3/3     Running     0               19s   10.244.0.38     minikube   <none>           <none>
rook-ceph-mds-myfs-a-7cb97b4796-f7cnr           1/1     Running     10 (28h ago)    43h   10.244.0.12     minikube   <none>           <none>
rook-ceph-mds-myfs-b-75657896f8-p6fd4           1/1     Running     9 (28h ago)     43h   10.244.0.13     minikube   <none>           <none>
rook-ceph-mgr-a-844f4f47b8-hgzd2                1/1     Running     6 (28h ago)     43h   10.244.0.7      minikube   <none>           <none>
rook-ceph-mon-a-59d69ff7d8-tjtmm                1/1     Running     0               25h   10.244.0.20     minikube   <none>           <none>
rook-ceph-osd-0-99c4fd57d-6k799                 1/1     Running     0               25h   10.244.0.24     minikube   <none>           <none>
rook-ceph-osd-prepare-minikube-pq2rl            0/1     Completed   0               25h   10.244.0.23     minikube   <none>           <none>
rook-ceph-tools-696bff5bc6-gqgw9                1/1     Running     0               43h   10.244.0.8      minikube   <none>        
```
*  Create the application pod
```
[🎩︎]mrajanna@fedora rbd $]kubectl get po
NAME              READY   STATUS    RESTARTS   AGE
csirbd-demo-pod   1/1     Running   0          28s
```
* Check the watcher on the rbd image (IP should be of holder pod)
```
# rbd status output

sh-4.4$ rbd status replicapool/csi-vol-cc2ee14a-b0dc-11ec-be75-fac2bf87c48c
Watchers:
	watcher=10.244.0.37:0/2548114391 client.126292 cookie=18446462598732840961
```

* Exec into the application pod and  write some data
```
[🎩︎]mrajanna@fedora rbd $]kubectl exec -it csirbd-demo-pod -- sh
# df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay          28G  7.9G   19G  31% /
tmpfs            64M     0   64M   0% /dev
tmpfs           2.0G     0  2.0G   0% /sys/fs/cgroup
/dev/vda1        28G  7.9G   19G  31% /etc/hosts
shm              64M     0   64M   0% /dev/shm
/dev/rbd0       976M  2.6M  907M   1% /var/lib/www/html
tmpfs           3.9G   12K  3.9G   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs           2.0G     0  2.0G   0% /proc/acpi
tmpfs           2.0G     0  2.0G   0% /proc/scsi
tmpfs           2.0G     0  2.0G   0% /sys/firmware
# cd /var/lib/www/html
# touch a
# cat a
# 
```

* Bring down cephcsi pod and read and write data

```
[🎩︎]mrajanna@fedora ceph-csi $]kuberc get ds csi-rbdplugin -oyaml >ds.yaml
[🎩︎]mrajanna@fedora ceph-csi $]kuberc delete ds csi-rbdplugin
daemonset.apps "csi-rbdplugin" deleted
[🎩︎]mrajanna@fedora ceph-csi $]kuberc get po
NAME                                            READY   STATUS      RESTARTS        AGE
csi-cephfsplugin-mq4t7                          3/3     Running     1 (22h ago)     23h
csi-cephfsplugin-provisioner-5dc9cbcc87-x496h   6/6     Running     4 (6h55m ago)   23h
csi-plugin-holder-tjfpv                         1/1     Running     0               18m
csi-rbdplugin-provisioner-d47c55d7c-qpqcp       6/6     Running     4 (6h55m ago)   24h
rook-ceph-mds-myfs-a-7cb97b4796-f7cnr           1/1     Running     10 (28h ago)    43h
rook-ceph-mds-myfs-b-75657896f8-p6fd4           1/1     Running     9 (28h ago)     43h
rook-ceph-mgr-a-844f4f47b8-hgzd2                1/1     Running     6 (28h ago)     43h
rook-ceph-mon-a-59d69ff7d8-tjtmm                1/1     Running     0               25h
rook-ceph-osd-0-99c4fd57d-6k799                 1/1     Running     0               25h
rook-ceph-osd-prepare-minikube-pq2rl            0/1     Completed   0               25h
rook-ceph-tools-696bff5bc6-gqgw9                1/1     Running     0               43h
[🎩︎]mrajanna@fedora ceph-csi $]
```

* Write data to application pod

```
# touch b
# echo testing>a
# cat a
testing
# ls
a  b  lost+found
```
 
* Bring up the cephcsi pod and do delete pod

```
[🎩︎]mrajanna@fedora ceph-csi $]kubectl apply -f ds.yaml 
daemonset.apps/csi-rbdplugin created
[🎩︎]mrajanna@fedora ceph-csi $]kuberc get po
NAME                                            READY   STATUS      RESTARTS        AGE
csi-cephfsplugin-mq4t7                          3/3     Running     1 (22h ago)     23h
csi-cephfsplugin-provisioner-5dc9cbcc87-x496h   6/6     Running     4 (6h56m ago)   23h
csi-plugin-holder-tjfpv                         1/1     Running     0               19m
csi-rbdplugin-4rttw                             3/3     Running     0               3s
csi-rbdplugin-provisioner-d47c55d7c-qpqcp       6/6     Running     4 (6h56m ago)   24h
rook-ceph-mds-myfs-a-7cb97b4796-f7cnr           1/1     Running     10 (29h ago)    43h
rook-ceph-mds-myfs-b-75657896f8-p6fd4           1/1     Running     9 (29h ago)     43h
rook-ceph-mgr-a-844f4f47b8-hgzd2                1/1     Running     6 (29h ago)     43h
rook-ceph-mon-a-59d69ff7d8-tjtmm                1/1     Running     0               25h
rook-ceph-osd-0-99c4fd57d-6k799                 1/1     Running     0               25h
rook-ceph-osd-prepare-minikube-pq2rl            0/1     Completed   0               25h
rook-ceph-tools-696bff5bc6-gqgw9                1/1     Running     0               43h
[🎩︎]mrajanna@fedora ceph-csi $]
```
* Delete application pod
```
[🎩︎]mrajanna@fedora rbd $]kubectl delete -f pod.yaml 
pod "csirbd-demo-pod" deleted
[🎩︎]mrajanna@fedora rbd $]kubectl get po
No resources found in default namespace.
```
* Check watchers are deleted on rbd image
```
sh-4.4$ rbd status replicapool/csi-vol-cc2ee14a-b0dc-11ec-be75-fac2bf87c48c
Watchers: none
sh-4.4$
```

## cephfs validation after the nsenter

* Get the po -o wide output and note the IP of holder pod and the rbdplugin pod (it should be pod ip)

```
[🎩︎]mrajanna@fedora ceph-csi $]kuberc get po -owide
NAME                                            READY   STATUS      RESTARTS       AGE     IP            NODE       NOMINATED NODE   READINESS GATES
csi-cephfsplugin-fjmkf                          3/3     Running     0              77s     10.244.0.41   minikube   <none>           <none>
csi-cephfsplugin-provisioner-5dc9cbcc87-x496h   6/6     Running     4 (7h ago)     24h     10.244.0.28   minikube   <none>           <none>
csi-plugin-holder-tjfpv                         1/1     Running     0              23m     10.244.0.37   minikube   <none>           <none>
csi-rbdplugin-4rttw                             3/3     Running     0              4m13s   10.244.0.40   minikube   <none>           <none>
csi-rbdplugin-provisioner-d47c55d7c-qpqcp       6/6     Running     4 (7h ago)     24h     10.244.0.26   minikube   <none>           <none>
rook-ceph-mds-myfs-a-7cb97b4796-f7cnr           1/1     Running     10 (29h ago)   43h     10.244.0.12   minikube   <none>           <none>
rook-ceph-mds-myfs-b-75657896f8-p6fd4           1/1     Running     9 (29h ago)    43h     10.244.0.13   minikube   <none>           <none>
rook-ceph-mgr-a-844f4f47b8-hgzd2                1/1     Running     6 (29h ago)    43h     10.244.0.7    minikube   <none>           <none>
rook-ceph-mon-a-59d69ff7d8-tjtmm                1/1     Running     0              25h     10.244.0.20   minikube   <none>           <none>
rook-ceph-osd-0-99c4fd57d-6k799                 1/1     Running     0              25h     10.244.0.24   minikube   <none>           <none>
rook-ceph-osd-prepare-minikube-pq2rl            0/1     Completed   0              25h     10.244.0.23   minikube   <none>           <none>
rook-ceph-tools-696bff5bc6-gqgw9                1/1     Running     0              43h     10.244.0.8    minikube   <none>           <none>
[🎩︎]mrajanna@fedora ceph-csi $]
```

* Create the application pod

```
[🎩︎]mrajanna@fedora cephfs $]kubectl create -f pod.yaml 
pod/csicephfs-demo-pod created
[🎩︎]mrajanna@fedora cephfs $]kubectl get po
NAME                 READY   STATUS    RESTARTS   AGE
csicephfs-demo-pod   1/1     Running   0          36s
```

* Check the client IP of the subvolume, it should be of the holder pod

```
sh-4.4$ ceph tell mds.0 client ls
2022-04-01T11:07:40.189+0000 7f660b7fe700  0 client.126313 ms_handle_reset on v2:10.244.0.12:6800/1837830945
2022-04-01T11:07:40.221+0000 7f660b7fe700  0 client.126315 ms_handle_reset on v2:10.244.0.12:6800/1837830945
[
    {
        "id": 126305,
        "entity": {
            "name": {
                "type": "client",
                "num": 126305
            },
            "addr": {
                "type": "v1",
                "addr": "10.244.0.37:0",
                "nonce": 2489648140
            }
        },
        "state": "open",
        "num_leases": 0,
        "num_caps": 1,
        "request_load_avg": 1,
        "uptime": 42.091375360999997,
        "requests_in_flight": 0,
        "num_completed_requests": 0,
        "num_completed_flushes": 0,
        "reconnecting": false,
        "recall_caps": {
            "value": 0,
            "halflife": 60
        },
        "release_caps": {
            "value": 0,
            "halflife": 60
        },
        "recall_caps_throttle": {
            "value": 0,
            "halflife": 1.5
        },
        "recall_caps_throttle2o": {
            "value": 0,
            "halflife": 0.5
        },
        "session_cache_liveness": {
            "value": 0.90734644607347803,
            "halflife": 300
        },
        "cap_acquisition": {
            "value": 0,
            "halflife": 10
        },
        "delegated_inos": [],
        "inst": "client.126305 v1:10.244.0.37:0/2489648140",
        "completed_requests": [],
        "prealloc_inos": [],
        "client_metadata": {
            "client_features": {
                "feature_bits": "0x00000000000001ff"
            },
            "metric_spec": {
                "metric_flags": {
                    "feature_bits": "0x"
                }
            },
            "entity_id": "csi-cephfs-node",
            "hostname": "csi-cephfsplugin-fjmkf",
            "kernel_version": "4.19.202",
            "root": "/volumes/csi/csi-vol-c76c80b4-b1ab-11ec-8810-ba246b5fb0a2/6a6b131d-e94b-46ca-acc4-8772be0a92db"
        }
    },

```

* Exec in to the application pod and write the data

```
[🎩︎]mrajanna@fedora cephfs $]kubectl exec -it pod/csicephfs-demo-pod -- sh
# df -h
Filesystem                                                                                                         Size  Used Avail Use% Mounted on
overlay                                                                                                             28G  7.9G   19G  30% /
tmpfs                                                                                                               64M     0   64M   0% /dev
tmpfs                                                                                                              2.0G     0  2.0G   0% /sys/fs/cgroup
/dev/vda1                                                                                                           28G  7.9G   19G  30% /etc/hosts
shm                                                                                                                 64M     0   64M   0% /dev/shm
10.99.252.147:6789:/volumes/csi/csi-vol-c76c80b4-b1ab-11ec-8810-ba246b5fb0a2/6a6b131d-e94b-46ca-acc4-8772be0a92db  1.0G     0  1.0G   0% /var/lib/www/html
tmpfs                                                                                                              3.9G   12K  3.9G   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs                                                                                                              2.0G     0  2.0G   0% /proc/acpi
tmpfs                                                                                                              2.0G     0  2.0G   0% /proc/scsi
tmpfs                                                                                                              2.0G     0  2.0G   0% /sys/firmware
# cd /var/lib/www/html
# touch a
# ls a
a
```
* delete the cephfs daemonset pod

```
[🎩︎]mrajanna@fedora ceph-csi $]kuberc get ds csi-cephfsplugin -oyaml >cephfs-ds.yaml
[🎩︎]mrajanna@fedora ceph-csi $]kuberc delete ds csi-cephfsplugin
daemonset.apps "csi-cephfsplugin" deleted
[🎩︎]mrajanna@fedora ceph-csi $]kuberc get po
NAME                                            READY   STATUS      RESTARTS       AGE
csi-cephfsplugin-provisioner-5dc9cbcc87-x496h   6/6     Running     4 (7h3m ago)   24h
csi-plugin-holder-tjfpv                         1/1     Running     0              26m
csi-rbdplugin-4rttw                             3/3     Running     0              7m5s
csi-rbdplugin-provisioner-d47c55d7c-qpqcp       6/6     Running     4 (7h3m ago)   24h
rook-ceph-mds-myfs-a-7cb97b4796-f7cnr           1/1     Running     10 (29h ago)   43h
rook-ceph-mds-myfs-b-75657896f8-p6fd4           1/1     Running     9 (29h ago)    43h
rook-ceph-mgr-a-844f4f47b8-hgzd2                1/1     Running     6 (29h ago)    43h
rook-ceph-mon-a-59d69ff7d8-tjtmm                1/1     Running     0              25h
rook-ceph-osd-0-99c4fd57d-6k799                 1/1     Running     0              25h
rook-ceph-osd-prepare-minikube-pq2rl            0/1     Completed   0              25h
rook-ceph-tools-696bff5bc6-gqgw9                1/1     Running     0              43h
[🎩︎]mrajanna@fedora ceph-csi $]
```
* exec into the application pod and check we are still able to access the application
```
[🎩︎]mrajanna@fedora cephfs $]kubectl exec -it pod/csicephfs-demo-pod -- sh
# df -h
Filesystem                                                                                                         Size  Used Avail Use% Mounted on
overlay                                                                                                             28G  7.9G   19G  30% /
tmpfs                                                                                                               64M     0   64M   0% /dev
tmpfs                                                                                                              2.0G     0  2.0G   0% /sys/fs/cgroup
/dev/vda1                                                                                                           28G  7.9G   19G  30% /etc/hosts
shm                                                                                                                 64M     0   64M   0% /dev/shm
10.99.252.147:6789:/volumes/csi/csi-vol-c76c80b4-b1ab-11ec-8810-ba246b5fb0a2/6a6b131d-e94b-46ca-acc4-8772be0a92db  1.0G     0  1.0G   0% /var/lib/www/html
tmpfs                                                                                                              3.9G   12K  3.9G   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs                                                                                                              2.0G     0  2.0G   0% /proc/acpi
tmpfs                                                                                                              2.0G     0  2.0G   0% /proc/scsi
tmpfs                                                                                                              2.0G     0  2.0G   0% /sys/firmware
# cd /var/lib/www/html
# touch a
# ls a
a
# touch b
# echo testing>b
# cat b
testing
# 
```

* Bring back the cephfs daemonset pod
```
[🎩︎]mrajanna@fedora ceph-csi $]kubectl apply -f cephfs-ds.yaml 
daemonset.apps/csi-cephfsplugin created
[🎩︎]mrajanna@fedora ceph-csi $]
🎩︎]mrajanna@fedora ceph-csi $]kuberc get po -owide
NAME                                            READY   STATUS      RESTARTS       AGE   IP            NODE       NOMINATED NODE   READINESS GATES
csi-cephfsplugin-m4ktc                          3/3     Running     0              18s   10.244.0.43   minikube   <none>           <none>
csi-cephfsplugin-provisioner-5dc9cbcc87-x496h   6/6     Running     4 (7h4m ago)   24h   10.244.0.28   minikube   <none>           <none>
csi-plugin-holder-tjfpv                         1/1     Running     0              27m   10.244.0.37   minikube   <none>           <none>
csi-rbdplugin-4rttw                             3/3     Running     0              8m    10.244.0.40   minikube   <none>           <none>
csi-rbdplugin-provisioner-d47c55d7c-qpqcp       6/6     Running     4 (7h4m ago)   24h   10.244.0.26   minikube   <none>           <none>
rook-ceph-mds-myfs-a-7cb97b4796-f7cnr           1/1     Running     10 (29h ago)   43h   10.244.0.12   minikube   <none>           <none>
rook-ceph-mds-myfs-b-75657896f8-p6fd4           1/1     Running     9 (29h ago)    43h   10.244.0.13   minikube   <none>           <none>
rook-ceph-mgr-a-844f4f47b8-hgzd2                1/1     Running     6 (29h ago)    43h   10.244.0.7    minikube   <none>           <none>
rook-ceph-mon-a-59d69ff7d8-tjtmm                1/1     Running     0              25h   10.244.0.20   minikube   <none>           <none>
rook-ceph-osd-0-99c4fd57d-6k799                 1/1     Running     0              25h   10.244.0.24   minikube   <none>           <none>
rook-ceph-osd-prepare-minikube-pq2rl            0/1     Completed   0              25h   10.244.0.23   minikube   <none>           <none>
rook-ceph-tools-696bff5bc6-gqgw9                1/1     Running     0              43h   10.244.0.8    minikube   <none>           <none>
```

* delete the application
```
[🎩︎]mrajanna@fedora cephfs $]kubectl deleectl delete -f pod.yaml 
pod "csicephfs-demo-pod" deleted
```
* check the client of the subvolume
```
sh-4.4$ ceph tell mds.0 client ls
2022-04-01T11:11:03.118+0000 7f82d17fa700  0 client.126319 ms_handle_reset on v2:10.244.0.12:6800/1837830945
2022-04-01T11:11:03.165+0000 7f82d17fa700  0 client.126321 ms_handle_reset on v2:10.244.0.12:6800/1837830945
[]
```

nixpanic:
I wonder if this requires Rook at all? Is it not possible to take a very similar approach and deploy the `csi-holder` daemonset as part of the Ceph-CSI repository, or maybe even through CSI-Addons?

Madhu-1:
> I wonder if this requires Rook at all? Is it not possible to take a very similar approach and deploy the `csi-holder` daemonset as part of the Ceph-CSI repository, or maybe even through CSI-Addons?

Yes should be possible but we need more validation with upgrades etc,  this will be only used if the NetworkFilePath is set in the configmap or else everything remains the same as the backward compatible. once this becomes stable we can switch to the pod network and get rid of the host networking. Let's take this as a new issue?

nixpanic:
> Let's take this as a new issue?

Sure! This is a rather important change with a complex background. It should be documented in a design proposal in any case too. It needs to explain the reason for `nsenter`, and how the symlink setup/detection is expected to work. With such a description, I can better review the change and verify that it does what is intended.

leseb:
I agree that the Holder DaemonSet could be deployed by an addon instead of relying on Rook. Especially if we want to make this patch generic, by generic I mean that it removes the need of HostNetwork, if we get ride of that we solve all our problems :)

nixpanic:
> I agree that the Holder DaemonSet could be deployed by an addon instead of relying on Rook. Especially if we want to make this patch generic, by generic I mean that it removes the need of HostNetwork, if we get ride of that we solve all our problems :)

Not only Ceph-CSI is affected with this, any CSI-driver that does network mounting. If the solution is generic, other drivers will be able to benefit from this as well. Thinking of the csi-driver-nfs at the moment, as that is a new feature we're adding.

leseb:
> > I agree that the Holder DaemonSet could be deployed by an addon instead of relying on Rook. Especially if we want to make this patch generic, by generic I mean that it removes the need of HostNetwork, if we get ride of that we solve all our problems :)
> 
> Not only Ceph-CSI is affected with this, any CSI-driver that does network mounting. If the solution is generic, other drivers will be able to benefit from this as well. Thinking of the csi-driver-nfs at the moment, as that is a new feature we're adding.

We have already discussed that NFS is impacted by this within the Rook team especially when we saw https://github.com/rook/rook/pull/9927. If the solution is articulated by Rook it's fine but Rook is supposed to "only" deploy Ceph-CSI or NFS (whatever). However, the original form is deployment should be coming from the driver itself.

What are you thinking about exactly when you say that NFS is also impacted? If we want something generic perhaps the CSI spec itself needs to change but each driver implements its own deployment so yes NFS is free to re-use the same approach if they want to.

nixpanic:
> What are you thinking about exactly when you say that NFS is also impacted? If we want something generic perhaps the CSI spec itself needs to change but each driver implements its own deployment so yes NFS is free to re-use the same approach if they want to.

If the solution is generic enough, maybe it can be adopted by Kubernetes and the CSI Specification at one point. Until then, we might be able to provide some API and the CSI-holder Pods from withing CSI-Addons. The CSI-Addons are a little like an experimentation/incubation before proposing changes to the CSI Specification and the implementation in Kubernetes CSI.

idryomov:
> If the solution is generic enough, maybe it can be adopted by Kubernetes and the CSI Specification at one point.

This solution is just a workaround for the fact that Kubernetes doesn't support sharing network namespaces between pods -- nothing more, nothing less.  As such it may be met with resistance in wider Kubernetes communities ;)

In CSI, the issue of sudden pod termination or restarts/upgrades is certainly not unique to the Ceph kernel client.  But I don't know how many other projects have attempted to tackle it.  Take Rancher's Longhorn as an example: I haven't actually used it so I could be wrong but I think its "solution" is to basically kill the workload pod and then rely on Kubernetes to spawn it back up somewhere else, together with a restarted/upgraded I/O-handling backend pod.

internal/util/csiconfig.go
```diff
@@ -49,6 +49,8 @@ type ClusterInfo struct {
 		// SubvolumeGroup contains the name of the SubvolumeGroup for CSI volumes
 		SubvolumeGroup string `json:"subvolumeGroup"`
 	} `json:"cephFS"`
+	// symlink filepath for the network where we need to execute commands.
+	NetworkFilePath string `json:"networkFilePath"`
```
idryomov: Nit: perhaps `NetNamespaceFilePath` to make it immediately obvious what it is about?

internal/util/cephcmds.go
```diff
@@ -32,6 +32,41 @@ import (
...
+// and returns separate stdout and stderr streams. In case ctx is not set to
+// context.TODO(), the command will be logged after it was executed.
+func ExecuteCommandWithNSEnter(ctx context.Context, netPath, program string, args ...string) (string, string, error) {
+	var (
+		sanitizedArgs = StripSecretInArgs(args)
+		stdoutBuf     bytes.Buffer
+		stderrBuf     bytes.Buffer
+	)
+	args = append([]string{fmt.Sprintf("--net=%s", netPath), program}, args...)
+	cmd := exec.Command("nsenter", args...) // #nosec:G204, commands executing not vulnerable.
```
idryomov, line 52: I would make it `nsenter --net=%s -- <program> <args>` to prevent `nsenter` from trying to interpret some program argument as an option.  It did that in the past and IIRC that got fixed but let's stay on the safe side.

Madhu-1: Thanks @idryomov sounds like a good idea, will update the PR

internal/cephfs/mounter/fuse.go
```diff
@@ -47,7 +47,10 @@ var (
 
 type FuseMounter struct{}
 
-func mountFuse(ctx context.Context, mountPoint string, cr *util.Credentials, volOptions *store.VolumeOptions) error {
+func mountFuse(ctx context.Context,
```
nixpanic: I am looking at this, and wonder what changed in addition to the formatting? If it is really formatting only, please put that in a different commit or PR.

Same counts for the function signature updates elsewhere.

internal/cephfs/mounter/fuse.go
```diff
@@ -65,10 +68,25 @@ func mountFuse(ctx context.Context, mountPoint string, cr *util.Credentials, vol
...
-
-	_, stderr, err := util.ExecCommand(ctx, "ceph-fuse", args[:]...)
-	if err != nil {
-		return fmt.Errorf("%w stderr: %s", err, stderr)
+	var (
+		stderr string
+		err    error
+	)
+
+	netPath, nErr := util.GetNetNamespaceFilePath(util.CsiConfigFile, volOptions.ClusterID)
```
nixpanic: would it not make sense to detect this in `NodeStageVolume` and set the `netPath` as part of the `VolumeOptions` struct? The `netPath` is related to how the connection is made, similar to the MONs and other options.

internal/cephfs/mounter/volumemounter.go
```diff
@@ -99,7 +99,10 @@ func LoadAvailableMounters(conf *util.Config) error {
 }
 
 type VolumeMounter interface {
-	Mount(ctx context.Context, mountPoint string, cr *util.Credentials, volOptions *store.VolumeOptions) error
+	Mount(ctx context.Context,
```
nixpanic: I'm missing what changed, except for the formatting.

internal/rbd/rbd_attach.go
```diff
@@ -417,8 +423,21 @@ func createPath(ctx context.Context, volOpt *rbdVolume, device string, cr *util.
...
 	}
 
-	// Execute map
-	stdout, stderr, err := util.ExecCommand(ctx, cli, mapArgs...)
+	var (
+		stdout string
+		stderr string
+		err    error
+	)
+	netPath, nErr := util.GetNetNamespaceFilePath(util.CsiConfigFile, volOpt.ClusterID)
```
nixpanic: Please set `netPath` as part of the `rbdVolume` object creation, in `NodeStageVolume` or similar.

internal/util/cephcmds.go
```diff
@@ -32,6 +32,42 @@ import (
...
+// and returns separate stdout and stderr streams. In case ctx is not set to
+// context.TODO(), the command will be logged after it was executed.
+func ExecuteCommandWithNSEnter(ctx context.Context, netPath, program string, args ...string) (string, string, error) {
+	var (
+		sanitizedArgs = StripSecretInArgs(args)
+		stdoutBuf     bytes.Buffer
+		stderrBuf     bytes.Buffer
+	)
+	//  nsenter --net=%s -- <program> <args>
+	args = append([]string{fmt.Sprintf("--net=%s", netPath), "--", program}, args...)
```
nixpanic, line 51: In order to improve security, please check that the `netPath` exists. The path is read from a ConfigMap, which could contain some unexpected strings. Not sure what happens if `netPath` is set to `/proc/self/ns/net -- /bin/date ; echo DO NOT DO THIS`

nixpanic (changes requested):
Missing some documentation as to what the `netNamespaceFilePath` should point to, in case it is set. It should be a symlink that points to `/proc/<pid-of-some-process/ns/net`, which is rather critical information.

internal/cephfs/mounter/fuse.go
Madhu-1: It was leftover from my previous changes, removed now

internal/cephfs/mounter/fuse.go
Madhu-1: Yes it should be possible, updated code for it.

internal/util/cephcmds.go
Madhu-1: added a check for the path exists, if self is given it will be executed in its own namespace?

nixpanic: yes, `self` is a symlink to the process' own PID.

nixpanic (commented):
Changes look good, and clean now. Thanks!

Still missing a description and example of where the symlink should point to. The things that creates the symlink, and the container using it probably require `hostPID: true`? It would be good to document these requirements too.

Madhu-1:
> Changes look good, and clean now. Thanks!
> 
> Still missing a description and example of where the symlink should point to. The things that creates the symlink, and the container using it probably require `hostPID: true`? It would be good to document these requirements too.

working on the Document and template changes, which should be there in some time.

Madhu-1:
@idryomov @leseb PTAL

examples/README.md
```diff
@@ -50,6 +50,39 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+the volume become blocked, even after restarting the CSI pods.
+
+The only workaround is to restart the node where the Ceph CSI plugin pod was
+restarted.  This can be mitigated by running the `rbd map`/`mount -t` commands
+in a different namespace which does not get deleted and recreated often.
+
+If someone wants to run the CephCSI with the pod networking they can still do
+by setting the `netNamespaceFilePath`. If this path is set CephCSI will execute
+the `rbd map`/`mount -t` commands by entering the network namespace specified by
+`netNamespaceFilePath` with the nsenter.
```
nixpanic: > ... with the `nsenter` command.

Or just

> ... with `nsenter`.

Might be good to reference `man 7 namespaces` and/or `man 7 network_namespaces` for additional information.

Madhu-1: added link to man page

examples/README.md
```diff
@@ -50,6 +50,41 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+## Running CephCSI with pod networking
+
+The current problem with Pod Networking, is when a CephFS/RBD volume is mounted
+in a pod using Ceph CSI and then the CSI CephFS/RBD plugin is restarted or
+terminated (e.g. by restarting or deleting its DaemonSet), all operations on
+the volume become blocked, even after restarting the CSI pods.
+
+The only workaround is to restart the node where the Ceph CSI plugin pod was
+restarted.  This can be mitigated by running the `rbd map`/`mount -t` commands
+in a different namespace which does not get deleted and recreated often.
```
idryomov: The namespace doesn't ever get recreated:
```suggestion
in a different network namespace which does not get deleted when the CSI CephFS/RBD plugin is restarted or terminated.
```

examples/README.md
```diff
@@ -50,6 +50,41 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+
+If someone wants to run the CephCSI with the pod networking they can still do
+by setting the `netNamespaceFilePath`. If this path is set CephCSI will execute
+the `rbd map`/`mount -t` commands by entering the [network
```
idryomov:
```suggestion
the `rbd map`/`mount -t` commands after entering the [network
```

examples/README.md
```diff
@@ -50,6 +50,41 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+namespace](https://man7.org/linux/man-pages/man7/network_namespaces.7.html)
+specified by `netNamespaceFilePath` with the
+[nsenter](https://man7.org/linux/man-pages/man1/nsenter.1.html) command.
+
+`netNamespaceFilePath` is typical of the network namespace of a long-running
```
idryomov:
```suggestion
`netNamespaceFilePath` should point to the network namespace of some long-running
```

examples/README.md
```diff
@@ -50,6 +50,41 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+process it usually `/proc/<long running process id>/ns/net`
```
idryomov:
```suggestion
process, typically it would be a symlink to `/proc/<long running process id>/ns/net`.
```

examples/README.md
```diff
@@ -50,6 +50,41 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+
+The long-running process can also be another pod which is a (Daemonset) which
```
idryomov:
```suggestion
The long-running process can also be another pod which is a DaemonSet which
```

examples/README.md
```diff
@@ -50,6 +50,41 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+never restarts. This Pod should only be stopped and restarted when a node is
+stopped so that volume operations do not become blocked. The new DaemonSet pod
+can contain a single container, the container responsible for holding
+filesystem mounts and mapped block devices. It is used as a passthrough by the
```
idryomov: It wouldn't "hold" mounts or block devices -- in fact they don't need holding.  It's all about the pod network:
```suggestion
can contain a single container, responsible for holding its pod network alive.
It is used as a passthrough by the
```

examples/README.md
```diff
@@ -50,6 +50,41 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+CephCSI plugin pod which when mounting or mapping will use the network
+namespace of this pod.
+
+Once the pod is created get the PID of it and create a symlink of
+`/proc/$$/ns/net` in the hostPath volume shared with the csi-plugin pod and
```
idryomov:
```suggestion
Once the pod is created get its PID and create a symlink to
`/proc/<PID>/ns/net` in the hostPath volume shared with the csi-plugin pod and
```

internal/cephfs/mounter/fuse.go
```diff
@@ -65,10 +65,21 @@ func mountFuse(ctx context.Context, mountPoint string, cr *util.Credentials, vol
...
+	if volOptions.NetNamespaceFilePath != "" {
+		_, stderr, err = util.ExecuteCommandWithNSEnter(ctx, volOptions.NetNamespaceFilePath, "ceph-fuse", args[:]...)
+		if err != nil {
+			return fmt.Errorf("%w stderr: %s", err, stderr)
+		}
+	} else {
+		_, stderr, err = util.ExecCommand(ctx, "ceph-fuse", args[:]...)
+		if err != nil {
+			return fmt.Errorf("%w stderr: %s", err, stderr)
+		}
```
idryomov: `err != nil` check could be moved outside to avoid repeating it.

internal/cephfs/mounter/kernel.go
```diff
@@ -51,9 +51,21 @@ func mountKernel(ctx context.Context, mountPoint string, cr *util.Credentials, v
...
+	if volOptions.NetNamespaceFilePath != "" {
+		_, stderr, err = util.ExecuteCommandWithNSEnter(ctx, volOptions.NetNamespaceFilePath, "mount", args[:]...)
+		if err != nil {
+			return fmt.Errorf("%w stderr: %s", err, stderr)
+		}
+	} else {
+		_, stderr, err = util.ExecCommand(ctx, "mount", args[:]...)
+		if err != nil {
+			return fmt.Errorf("%w stderr: %s", err, stderr)
+		}
```
idryomov: ditto

idryomov (approved):
Some nits, mostly to make the explanation clearer.  Otherwise LGTM!

Madhu-1:
> Some nits, mostly to make the explanation clearer. Otherwise LGTM!

Thanks, @idryomov for the valuable feedback. all comments are addreeed.

nixpanic (approved):

Madhu-1:
> /retest ci/centos/mini-e2e/k8s-1.22

Rook deployment failure

Madhu-1:

cephfs deployment failure `5 14:22:18.237: timeout waiting for deployment csi-rbdplugin-provisioner: error waiting for deployment "csi-rbdplugin-0provisioner" status to match desired state: `

examples/README.md
```diff
@@ -50,6 +50,42 @@ kubectl replace -f ./csi-config-map-sample.yaml
...
+stopped so that volume operations do not become blocked. The new DaemonSet pod
+can contain a single container, responsible for holding its pod network alive.
+It is used as a passthrough by the CephCSI plugin pod which when mounting or
+mapping will use the network namespace of this pod.
+
+Once the pod is created get its PID and create a symlink to
+`/proc/<PID>/ns/net` in the hostPath volume shared with the csi-plugin pod and
+specify the path in the `netNamespaceFilePath` option.
+
+*Note* This Pod should have `hostPID: true` in the Pod Spec.
```
pkalever, line 87: @Madhu-1 will it be worth mentioning the cons of this model may be as a `FIXME`, for example, unable to upgrade, binary copies issues etc?

pkalever:
@Madhu-1 The PR looks great, but I think an e2e test case would be a great addition if possible.
Thanks!

examples/README.md
Madhu-1: i think we dont need to worry about that one. if required as we progress will add it.

internal/util/cephcmds.go
```diff
@@ -32,6 +33,47 @@ import (
...
+		stderrBuf     bytes.Buffer
+	)
+
+	// check netPath exists
+	if _, err := os.Stat(netPath); err != nil {
+		return "", "", fmt.Errorf("failed to get stat for %s %w", netPath, err)
+	}
+	//  nsenter --net=%s -- <program> <args>
+	args = append([]string{fmt.Sprintf("--net=%s", netPath), "--", program}, args...)
+	cmd := exec.Command("nsenter", args...) // #nosec:G204, commands executing not vulnerable.
```
pkalever, line 52: @Madhu-1 do we have `util-linux-core` (provides nsenter) package always ?

Madhu-1:
> @Madhu-1 The PR looks great, but I think an e2e test case would be a great addition if possible. Thanks!

adding E2E is not easy as we need to to start a new holder pod and get the required information like PID and update the configmap. that's the reason i passed the manual testing results in  https://github.com/ceph/ceph-csi/pull/2983#issuecomment-1085784808

internal/util/cephcmds.go
Madhu-1: Yes we have this in base ceph image already

pkalever:
> > @Madhu-1 The PR looks great, but I think an e2e test case would be a great addition if possible. Thanks!
> 
> adding E2E is not easy as we need to to start a new holder pod and get the required information like PID and update the configmap. that's the reason i passed the manual testing results in [#2983 (comment)](https://github.com/ceph/ceph-csi/pull/2983#issuecomment-1085784808)

Understood, may be at least we open an issue to track it for future?

Madhu-1:
> > > @Madhu-1 The PR looks great, but I think an e2e test case would be a great addition if possible. Thanks!
> > 
> > 
> > adding E2E is not easy as we need to to start a new holder pod and get the required information like PID and update the configmap. that's the reason i passed the manual testing results in [#2983 (comment)](https://github.com/ceph/ceph-csi/pull/2983#issuecomment-1085784808)
> 
> Understood, may be at least we open an issue to track it for future?

https://github.com/ceph/ceph-csi/issues/2999 here it is

pkalever:
> > > > @Madhu-1 The PR looks great, but I think an e2e test case would be a great addition if possible. Thanks!
> > > 
> > > 
> > > adding E2E is not easy as we need to to start a new holder pod and get the required information like PID and update the configmap. that's the reason i passed the manual testing results in [#2983 (comment)](https://github.com/ceph/ceph-csi/pull/2983#issuecomment-1085784808)
> > 
> > 
> > Understood, may be at least we open an issue to track it for future?
> 
> #2999 here it is

Thanks @Madhu-1

internal/util/cephcmds.go
pkalever: Thanks for confirming.

pkalever (approved):
LGTM, thanks!

examples/README.md
idryomov: > @Madhu-1 will it be worth mentioning the cons of this model may be as a `FIXME`, for example, unable to upgrade, binary copies issues etc?

I don't think there are any upgrade issues and there is certainly no binary copying going on with this approach.  The long-running pod could literally be just a `pause` process.

pkalever: AH! That sounds great and intresting too!

nixpanic:
@mergifyio refresh

nixpanic:
@mergifyio rebase

nixpanic:
**ci/centos/mini-e2e-helm/k8s-1.21** failed with:
```
Apr  6 18:11:55.418: backend images not matching kubernetes resource count,image count 22 kubernetes resource count 21
```

Madhu-1:
@mergifyio refresh

Madhu-1:
> /test ci/centos/mini-e2e-helm/k8s-1.21

Rook deployment failure.

nixpanic:
@mergifyio refresh

PR #2898, opened 2022-02-22
rbd: update volume/snapshot-volume details on the image as metadata

pkalever:
# Describe what this PR does #

 This helps Monitoring solutions without access to Kubernetes clusters to display the details of the PV/PVC/PVCNameSpace and snapshot-name/snapshot-namespace/snapshotcontent-name in their dashboard.

## Is there anything that requires special attention ##

Scenarios covered:
* Validate metadata is present after Create of a PV
* Validate metadata is set after Provisioner pod is restarted
* Validate metadata is updated after a PVC is deleted and PV is reattached to a new PVC
* Validate metadata is updated after a PVC is deleted and PV is reattached to a new PVC on a different namespace
* Validate metadata is set on snapshot create

## Related issues ##
Fixes: #2874

## Other details ##
Currently, this PR uses K8s CO specific keywords
```
// PV metadata keys
pvcNameKey      = "csi.storage.k8s.io/pvc/name"
pcNamespaceKey = "csi.storage.k8s.io/pvc/namespace"
pvNameKey       = "csi.storage.k8s.io/pv/name"

// snapshot metadata keys                                              
volSnapNameKey        = "csi.storage.k8s.io/volumesnapshot/name"       
volSnapNamespaceKey   = "csi.storage.k8s.io/volumesnapshot/namespace"  
volSnapContentNameKey = "csi.storage.k8s.io/volumesnapshotcontent/name"

```
we might want to use generalized naming for keywords, let's discuss it on this PR.

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)
- `/retest all`: run this in case the CentOS CI failed to start/report any test
  progress or results

</details>

PR #2991, opened 2022-04-05
e2e: set --extra-create-metadata=false

pkalever:

set "-extra-create-metadata=false" on
"create a Block mode PVC-PVC clone and bind it to an app"
to avoid ErrImageNotFound when rbd images are undergoing flattening
operation.

WorkAround: #2327

---

<details>

  progress or results

</details>


Madhu-1:
@pkalever this should be part of https://github.com/ceph/ceph-csi/pull/2898 and it's not fixing the problem of #2327 rather it just avoiding it.

nixpanic (changes requested):
Instead of reading the yaml files from the `deploy/` directory, extend the functionality of `api/` and use that.

pkalever:
> Instead of reading the yaml files from the `deploy/` directory, extend the functionality of `api/` and use that.

Where did you see that we are reading from `deploy/` directory? And can you explain a bit on which api you are referent to extend here? (may be elaborate a bit more please)

e2e/rbd.go
```diff
@@ -2026,6 +2028,41 @@ var _ = Describe("RBD", func() {
 				if err != nil {
 					e2elog.Failf("failed to get server version: %v", err)
 				}
+
+				deploymentPatch, err := json.Marshal(map[string]interface{}{
```
nixpanic, line 2032: Why use a patch, and not just `Get()` and `Update()`. There should not be a need to list all `args` that way.

pkalever: Because we need this for both helm and non-helm deployments.

nixpanic: You can just use the Kubernetes API to get the Deployment object, modify the `args` and send the updated object to the API server. No need for JSON and untyped maps of strings.

pkalever: Sure, I can give it a shot, thanks for the hint

pkalever:
@Madhu-1 The tests are failing even with this workaround:

```
Apr  7 12:00:18.848: FAIL: backend images not matching kubernetes resource count,image count 2 kubernetes resource count 1
```

charts/ceph-csi-rbd/values.yaml
```diff
@@ -128,13 +128,7 @@ provisioner:
...
   replicaCount: 3
   strategy:
-    # RollingUpdate strategy replaces old pods with new ones gradually,
-    # without incurring downtime.
-    type: RollingUpdate
-    rollingUpdate:
-      # maxUnavailable is the maximum number of pods that can be
-      # unavailable during the update process.
-      maxUnavailable: 50%
+    type: Recreate
```
nixpanic, line 131: What is the reason to change this? The comments from commit ea088d40be are very descriptive, not sure this is an improvement?

pkalever: Oh! sorry, this is just a workaround. Currently, when we set Strategy type to Rollingupdate, the Patch command is working differently, i.e while the old pod is still Running a new pod is getting created and waiting in the pending state, the old pod is not getting terminated (unless someone manually kill it) and the new pod is waiting for old one to get terminated, obviously, the status of pod replica is set to 2 out of which 1 is not ready. This scenario is leading to other failures in the CI, such as waiting for provisioner 1/2 state.

That said, when the `Rollingupdate: {}` attribute is set initially k8s won't give us the provision to patch/apply it to `null`, only manual edit is allowed.

That all said, this is just an experiment PR for now. Which is anyway not working as expected.

nixpanic: > That all said, this is just an experiment PR for now. Which is anyway not working as expected.

Ok! In that case, I'll mark this as a **Draft** for the time being.

pkalever:
@Madhu-1 I hope we are good to close this PR then?

Madhu-1:
> @Madhu-1 I hope we are good to close this PR then?

Yes we can close this one for now.

PR #3008, opened 2022-04-11
rbd: consider remote image health for primary

Madhu-1:
To consider the image is healthy during the Promote operation currently we are checking only the image state on the primary site. If the network is flaky or the remote site is down the image health is not as expected. To make sure the image is healthy
across the clusters check the state on both local and remote clusters.

some details: https://bugzilla.redhat.com/show_bug.cgi?id=2014495

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -615,8 +615,9 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 }
 
 // checkHealthyPrimary checks if the image is a healhty primary or not.
-// healthy primary image will be in up+stopped state, for states other
-// than this it returns an error message.
+// healthy primary image will be in up+stopped state in local cluster and
+// up+replaying in the remote clusters, for states other than this it returns
+// an error message.
 func checkHealthyPrimary(ctx context.Context, rbdVol *rbdVolume) error {
 	mirrorStatus, err := rbdVol.getImageMirroringStatus()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


yati1998 (commented):
LGTM

ShyamsundarR (approved):

Madhu-1:
@humblec @nixpanic Need approval from maintainers.

nixpanic (approved):
The linked BZ contains only very few public details. Please consider opening things up a little more.

ShyamsundarR:
@Madhu-1 
Was running some tests with the entire stack for DR, the issue that crops up with this secondary image check is that the VR is never marked Primary post a failover, IOW this condition remains false:
```
      reason: FailedToPromote
      status: "False"
      type: Completed
```

The reason being the image reports:
```
csi-vol-851eac94-f402-11ec-9b37-0242ac11000e:
  global_id:   461b4974-eabd-42c4-89c8-eb3870f42137
  state:       up+stopped
  description: local image is primary
  service:     a on hub
  last_update: 2022-06-24 21:23:17
  peer_sites:
    name: 77ae0133-c330-4c67-922b-77d8e9afd0b2
    state: down+stopped
    description: local image is primary
    last_update: 2022-06-24 21:23:34
```

Overall the reason for adding this was to ensure local image once enabled is marked for replication, instead of assuming that it is, hence the local check will suffice for now. In the (near) future as we add a periodic update to local status based on image info, that will give a clearer picture of replication state and last replicated instance time etc.

IOW, with this PR failover breaks and hence needs to be reverted.

nixpanic:
This should be reported as an issue!

Ideally some e2e testing caught this... We're still waiting on getting contributions for that though :cry: 

Unit tests are very little as well. It should be easier to improve in that area. Different "image reports" like the above can be stored in tests, after parsing there can be a checn on what actions need to be taken.

PR #3012, opened 2022-04-11
rbd: check nbd tool features only for rbd driver

Madhu-1:
calling setRbdNbdToolFeatures inside an init gets called in main.go for both cephfs and rbd driver. instead of calling it in init function call this in rbd driver.go as this is specific to rbd.

### DIFF
--- a/internal/rbd/driver/driver.go
+++ b/internal/rbd/driver/driver.go
@@ -153,6 +153,8 @@ func (r *Driver) Run(conf *util.Config) {
 			log.FatalLogMsg(err.Error())
 		}
 		rbd.SetGlobalInt("krbdFeatures", krbdFeatures)
+
+		rbd.SetRbdNbdToolFeatures()
 	}
 
 	if conf.IsControllerServer {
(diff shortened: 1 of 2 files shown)


nixpanic (changes requested):
This only needs to be done for the node-plugin, the controller/provisioner does not need to check for it

Madhu-1:
> This only needs to be done for the node-plugin, the controller/provisioner does not need to check for it

is controller/provisioner checking this one? i am calling this function https://github.com/ceph/ceph-csi/blob/3ae65c6b22aa1c9c2d0e72cc11682e68a8e615f3/internal/rbd/driver/driver.go#L136-L158 if its a nodesever only. am i missing anything?

nixpanic (approved):

Rakshith-R (approved):

pkalever (approved):
Good catch Madhu :+1:

Madhu-1:
Rook deployment failed in 3 out of 4 tests. adding retry label as this should not cause any functional regressions.

Madhu-1:
@mergifyio refresh

PR #3037, opened 2022-04-15
nfs: return gRPC status from CephFS CreateVolume failure

nixpanic:
The NFS Controller returns a non-gRPC error in case the CreateVolume
call for the CephFS volume fails. It is better to return the gRPC-status
of the error that the CephFS Controller sets.

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/nfs/controller/controllerserver.go
+++ b/internal/nfs/controller/controllerserver.go
@@ -18,7 +18,6 @@ package controller
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/ceph/ceph-csi/internal/cephfs"
 	"github.com/ceph/ceph-csi/internal/cephfs/store"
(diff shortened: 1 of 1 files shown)


internal/nfs/controller/controllerserver.go
```diff
@@ -75,7 +74,7 @@ func (cs *Server) CreateVolume(
 	req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	res, err := cs.backendServer.CreateVolume(ctx, req)
 	if err != nil {
-		return res, fmt.Errorf("failed to create CephFS volume: %w", err)
+		return res, status.Errorf(status.Code(err), "failed to create CephFS volume: %v", err)
```
Madhu-1: just return err from here?

nixpanic: sure! The same is already done on `DeleteVolume`, so it makes sense.

Madhu-1 (approved):

Rakshith-R (approved):

PR #3039, opened 2022-04-18
separate netNamespaceFilePath for cephfs and RBD

Madhu-1:
Currently, netNamespaceFilePath is the path of the core ceph configuration in the ceph-csi configmap and this makes the configuration and automation really hard for the users for the below reasons

* A single netNamespaceFilePath will be used for both cephfs and rbd. Need to have a shared hostPath where the pid symlink exists between the cephfs and rbd pod.
* When we define a new holder pod the taints and toleration need to be added separately, this introduces new configurations at the consumer level and adds complexity for maintenance.
 
As this is more of an rbd and cephfs specific configuration good to have it under the cephfs and rbd umbrella in ceph configurations.

As we introduced the RBD specific section in the config file, moving the radosNamespace to the RBD as its RBD specific option and keeping radosNamespace still at the global level to be backward compatible.

### DIFF
--- a/internal/util/csiconfig.go
+++ b/internal/util/csiconfig.go
@@ -46,6 +46,8 @@ type ClusterInfo struct {
 	Monitors []string `json:"monitors"`
 	// CephFS contains CephFS specific options
 	CephFS struct {
+		// symlink filepath for the network namespace where we need to execute commands.
+		NetNamespaceFilePath string `json:"netNamespaceFilePath"`
 		// SubvolumeGroup contains the name of the SubvolumeGroup for CSI volumes
 		SubvolumeGroup string `json:"subvolumeGroup"`
 	} `json:"cephFS"`
@@ -181,4 +183,14 @@ func GetRBDNetNamespaceFilePath(pathToConfig, clusterID string) (string, error) 
 	}
 
 	return cluster.RBD.NetNamespaceFilePath, nil
+}
+
+// GetCephFSNetNamespaceFilePath returns the netNamespaceFilePath for CephFS volumes.
+func GetCephFSNetNamespaceFilePath(pathToConfig, clusterID string) (string, error) {
+	cluster, err := readClusterInfo(pathToConfig, clusterID)
+	if err != nil {
+		return "", err
+	}
+
+	return cluster.CephFS.NetNamespaceFilePath, nil
 }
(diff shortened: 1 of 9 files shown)


internal/util/csiconfig.go
```diff
@@ -164,11 +176,21 @@ func GetClusterID(options map[string]string) (string, error) {
...
 }
 
-func GetNetNamespaceFilePath(pathToConfig, clusterID string) (string, error) {
+func GetRBDNetNamespaceFilePath(pathToConfig, clusterID string) (string, error) {
+	cluster, err := readClusterInfo(pathToConfig, clusterID)
+	if err != nil {
+		return "", err
+	}
+
+	return cluster.RBD.NetNamespaceFilePath, nil
```
leseb, line 185: What if `NetNamespaceFilePath` is empty? Is it handled somewhere?

internal/util/csiconfig.go
```diff
@@ -164,11 +176,21 @@ func GetClusterID(options map[string]string) (string, error) {
...
+
+// GetCephFSNetNamespaceFilePath returns the netNamespaceFilePath for CephFS volumes.
+func GetCephFSNetNamespaceFilePath(pathToConfig, clusterID string) (string, error) {
 	cluster, err := readClusterInfo(pathToConfig, clusterID)
 	if err != nil {
 		return "", err
 	}
 
-	return cluster.NetNamespaceFilePath, nil
+	return cluster.CephFS.NetNamespaceFilePath, nil
```
leseb, line 195: ditto

internal/util/csiconfig.go
Madhu-1: handled here https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/rbd_attach.go#L463-L467

internal/util/csiconfig.go
Madhu-1: https://github.com/ceph/ceph-csi/blob/devel/internal/cephfs/mounter/kernel.go#L59-L62 and https://github.com/ceph/ceph-csi/blob/devel/internal/cephfs/mounter/fuse.go#L73-L77

leseb (approved):
LGTM! Tested and validated with Rook in latest https://github.com/rook/rook/pull/9925

Madhu-1:
@mergifyio rebase

nixpanic (approved):

Madhu-1:
@mergifyio rebase

Rakshith-R (approved):

PR #3081, opened 2022-05-04
rbd: use `vaultAuthPath` variable name in error msg

Rakshith-R:
Before the change, the error msg was the following:
```
failed to set VAULT_AUTH_MOUNT_PATH in Vault config: path is empty
```
`vaultAuthPath` is the actual variable name set by the
user. The error message will now be the following:
```
failed to set 'vaultAuthPath' in vault config: path is empty
```

### DIFF
--- a/internal/kms/vault.go
+++ b/internal/kms/vault.go
@@ -352,7 +352,7 @@ func initVaultKMS(args ProviderInitArgs) (EncryptionKMS, error) {
 
 	kms.vaultConfig[vault.AuthMountPath], err = detectAuthMountPath(vaultAuthPath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to set %s in Vault config: %w", vault.AuthMountPath, err)
+		return nil, fmt.Errorf("failed to set \"vaultAuthPath\" in Vault config: %w", err)
 	}
 
 	vaultRole := vaultDefaultRole


internal/kms/vault.go
```diff
@@ -352,7 +352,7 @@ func initVaultKMS(args ProviderInitArgs) (EncryptionKMS, error) {
 
 	kms.vaultConfig[vault.AuthMountPath], err = detectAuthMountPath(vaultAuthPath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to set %s in Vault config: %w", vault.AuthMountPath, err)
+		return nil, fmt.Errorf("failed to set %s in Vault config: %w", "vaultAuthPath", err)
```
Madhu-1:
```suggestion
return nil, fmt.Errorf("failed to set vaultAuthPath in Vault config: %w", err)
```

Madhu-1 (dismissed):

internal/kms/vault.go
```diff
@@ -352,7 +352,7 @@ func initVaultKMS(args ProviderInitArgs) (EncryptionKMS, error) {
 
 	kms.vaultConfig[vault.AuthMountPath], err = detectAuthMountPath(vaultAuthPath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to set %s in Vault config: %w", vault.AuthMountPath, err)
+		return nil, fmt.Errorf("failed to set vaultAuthPath in Vault config: %w", err)
```
pkalever: I would prefer:
```suggestion
		return nil, fmt.Errorf("failed to set 'vaultAuthPath' in Vault config: %w", err)
```

internal/kms/vault.go
nixpanic: When quotes are used, use `"` like `%q` would do

internal/kms/vault.go
nixpanic: strange, I tried to reply here, but it ended up at a previous comment?

>  When quotes are used, use `"` like `%q` would do

internal/kms/vault.go
```diff
@@ -352,7 +352,7 @@ func initVaultKMS(args ProviderInitArgs) (EncryptionKMS, error) {
 
 	kms.vaultConfig[vault.AuthMountPath], err = detectAuthMountPath(vaultAuthPath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to set %s in Vault config: %w", vault.AuthMountPath, err)
+		return nil, fmt.Errorf("failed to set \"vaultAuthPath\" in Vault config: %w", err)
```
Rakshith-R, line 355: done PTAL @pkalever @nixpanic @Madhu-1

nixpanic (approved):

nixpanic:
We don't really need e2e testing for this, added the label to skip them a next rebase/retry

Madhu-1 (approved):

yati1998 (approved):

pkalever (approved):

Rakshith-R:
@mergifyio rebase

PR #3102, opened 2022-05-09
ci: fix commitlint problem

Madhu-1:
Still seeing the issue of the commitlint as below

fatal: unsafe repository ('/go/src/github.com/ceph/ceph-csi' is owned by someone else) To add an exception for this directory,
call:

git config --global --add safe.directory /go/src/github.com/ceph/ceph-csi

This is not fixed yet happening at https://github.com/ceph/ceph-csi/runs/6345425683?check_suite_focus=true#step:3:1419

Note:
```
Setting this in Dockerfile which is used for the tests instead of Makefile as the issue is not seen in the local environment and we run tests inside the container in the CI.
```

### DIFF
--- a/.github/workflows/commitlint.yaml
+++ b/.github/workflows/commitlint.yaml
@@ -17,7 +17,6 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ github.event.pull_request.head.sha }}
-          set-safe-directory: false
       - name: commitlint
         # yamllint disable-line rule:line-length
         run: make containerized-test CONTAINER_CMD=docker TARGET=commitlint GIT_SINCE="origin/${GITHUB_BASE_REF}"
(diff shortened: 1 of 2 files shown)


Rakshith-R (changes requested):
Let's add it within the script

https://github.com/ceph/ceph-csi/blob/9533889b643f8274e8a0aae5583b022ddb1f5ee3/Makefile#L153

The step needs to be executed within the container
https://github.com/ceph/ceph-csi/runs/6346056638?check_suite_focus=true

Madhu-1:
> Let's add it within the script
> 
> https://github.com/ceph/ceph-csi/blob/9533889b643f8274e8a0aae5583b022ddb1f5ee3/Makefile#L153

Am still testing this one. I don't see the problem in the local setup. if adding to a workflow does not help. will add it there as the last option.

yati1998 (dismissed):

Rakshith-R (approved):

yati1998 (approved):

nixpanic (approved):

Madhu-1:
@mergifyio backport release-v3.6

PR #3104, opened 2022-05-09
ci: add `ci/skip/multi-arch-build` label

nixpanic:
If the `ci/skip/multi-arch-build` label is set on a PR, the GitHub
Workflow only builds for the local architecture. This makes it possible
to merge PRs faster.

---

<details>

  progress or results

</details>

### DIFF
--- a/.github/workflows/build-multi-stage.yaml
+++ b/.github/workflows/build-multi-stage.yaml
@@ -15,7 +15,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: multi-arch-build
+        # yamllint disable-line rule:line-length
+        if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip/multi-arch-build') }}
         # podman cannot pull images with both tag and digest
         # https://github.com/containers/buildah/issues/1407
         # use docker to build images
         run: CONTAINER_CMD=docker ./scripts/build-multi-arch-image.sh
+      - name: single-arch-build
+        # yamllint disable-line rule:line-length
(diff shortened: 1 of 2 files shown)


.github/workflows/build-multi-stage.yaml
```diff
@@ -15,7 +15,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: multi-arch-build
+        # yamllint disable-line rule:line-length
+        if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip/multi-arch-build') }}
```
Madhu-1, line 19: @nixpanic instead of running a single arch how about doing `exit 0` or `return 0` which makes it even faster and really helps in the doc change or yaml change?

nixpanic: I considered something like that too. It seems this just take ~3 minutes now, so from my point of view that is acceptable. Do you want me to change this?

Madhu-1: should be fine :+1:

Madhu-1 (approved):

Rakshith-R (approved):

humblec:
@nixpanic its a good addition.. however what are the classifications here to build only for local arch?

humblec (approved):

nixpanic:
> @nixpanic its a good addition.. however what are the classifications here to build only for local arch?

What do you mean with "classifications"?

nixpanic:
Adding `DNM` label, so that this does not get merged while an other PR is almost done testing

PR #3133, opened 2022-05-24
cephfs: set PV/PVC and clusterName details on the subvolume as metadata

### LINKED ISSUE #2875: Maintain metadata information with CephFS subvolume

Maintain PV/PVC/NameSpace mapping with the CephFS subvolume  metadata

This helps Monitoring solutions without access to Kubernetes clusters to display the details of the PV/PVC/NameSpace in their dashboard.

Validate metadata is present after Create of a PV
Validate metadata is updated after Provisioner pod is restarted
Validate metadata is updated after a PVC deleted and PV is reattached to a new PVC

pkalever:

This helps Monitoring solutions without access to Kubernetes clusters to display the details of the PV/PVC/NameSpace in their dashboard.         

~~Needs: https://github.com/ceph/go-ceph/pull/691
Just for the CI to pass I have included go-ceph custom branch and hence `DNM`, but the reviews can continue for other commits.~~

## Related issues ##

Fixes: #2875

---

<details>

  progress or results

</details>


pkalever:
@Mergifyio rebase

e2e/cephfs.go
```diff
@@ -402,6 +406,80 @@ var _ = Describe(cephfsType, func() {
...
+				if err != nil {
+					e2elog.Failf("failed to load PVC: %v", err)
+				}
+				pvc.Namespace = f.UniqueName
+
+				err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to create PVC: %v", err)
+				}
+
```
Madhu-1, line 423: add validateSubvolumeCount and  validateOmapCount validations

e2e/cephfs.go
```diff
@@ -402,6 +406,80 @@ var _ = Describe(cephfsType, func() {
...
+				if err != nil || stdErr != "" {
+					e2elog.Failf("failed to get PVC name %s/%s %s: err=%v stdErr=%q",
+						fileSystemName, subvol[0], pvcNameKey, err, stdErr)
+				}
+				pvcName = strings.TrimSuffix(pvcName, "\n")
+				if pvcName != pvc.Name {
+					e2elog.Failf("expected pvcName %q got %q", pvc.Name, pvcName)
+				}
+
+				pvcNamespace, stdErr, err := execCommandInToolBoxPod(f,
```
Madhu-1: move this to a helper function as we need to do same validation for PVC restore and PVC clone tests

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,86 @@
...
+// ForceRemoveMetadata attempt to forcefully remove custom metadata set
+// on the subvolume in a volume using the metadata key.
+func (s *subVolumeClient) ForceRemoveMetadata(key string) error {
+	fsa, err := s.conn.GetFSAdmin()
+	if err != nil {
+		return err
+	}
+
+	return fsa.ForceRemoveMetadata(s.FsName, s.SubvolumeGroup, s.VolID, key)
+}
```
Madhu-1: where this is getting called?

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,86 @@
...
+
+// ListMetadata lists custom metadata (key-value pairs) set on the subvolume
+// in a volume.
+func (s *subVolumeClient) ListMetadata() (map[string]string, error) {
```
Madhu-1: same here as well? if this is not getting called please remove it.

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	return nil
+}
+
+// UnsetAllMetadata unset all the metadata from arg keys on subvolume.
+func (s *subVolumeClient) UnsetAllMetadata(keys []string) error {
+	for _, key := range keys {
+		err := s.RemoveMetadata(key)
+		// TODO: replace string comparison with errno.
```
Madhu-1, line 118: please open an issue to track the TODO

e2e/cephfs.go
```diff
@@ -402,6 +406,80 @@ var _ = Describe(cephfsType, func() {
 				}
 			})
 
+			By("create a PVC and check PVC/PV metadata on CephFS subvolume", func() {
```
Madhu-1, line 408: Add e2e for pvc restore and pvc clone.

Madhu-1 (changes requested):
requesting changes to avoid accidental merge as go-ceph changes are not available in the go-ceph release yet.

e2e/cephfs.go
```diff
@@ -402,6 +406,80 @@ var _ = Describe(cephfsType, func() {
...
+					e2elog.Failf("expected pvName %q got %q", pvcObj.Spec.VolumeName, pvName)
+				}
+
+				err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					e2elog.Failf("failed to delete PVC: %v", err)
+				}
+				err = deleteResource(cephFSExamplePath + "storageclass.yaml")
+				if err != nil {
+					e2elog.Failf("failed to delete CephFS storageclass: %v", err)
```
Madhu-1, line 464: to test these changes in we also need a new ceph release and also we need to make sure we support older ceph clusters also. need extra validation to check if ceph cluster supports these API's or not.

e2e/cephfs.go
pkalever: This will be part of a different PR: https://github.com/ceph/ceph-csi/pull/3168

Madhu-1 (commented):
We need to support older ceph versions that do not support metadata operations. if ceph doesnot support metadata operations we need to skip setting  metadata

pkalever:
> We need to support older ceph versions that do not support metadata operations. if ceph doesnot support metadata operations we need to skip setting metadata

Not just Quincy, the patches were backported to Pacific and Octopus I guess. I think ideally we should check on all those versions, right?

As we also have snapshot metadata-related changes in a different PR https://github.com/ceph/ceph-csi/pull/3168, is this something that we can cover in a separate PR?

e2e/cephfs_helper.go
```diff
@@ -212,6 +212,39 @@ func listCephFSSubVolumes(f *framework.Framework, filesystem, groupname string)
...
+		fmt.Sprintf("ceph fs subvolume metadata ls %s %s --group_name=%s --format=json", filesystem, subvolume, groupname),
+		rookNamespace)
+	if err != nil {
+		return metadata, err
+	}
+	if stdErr != "" {
+		return metadata, fmt.Errorf("error listing subvolume metadata %v", stdErr)
+	}
+
+	err = json.Unmarshal([]byte(stdout), &metadata)
```
pkalever, line 240: @Madhu-1 @nixpanic do you see anything wrong here, unmarshal is failing for some reason?

In the previous run, I had added a log to print stdout and this is how it looks:

```
Jul 18 14:48:08.414: INFO: ExecWithOptions: execute(POST https://192.168.39.245:8443/api/v1/namespaces/rook-ceph/pods/rook-ceph-tools-6944779c47-z9q6b/exec?command=%2Fbin%2Fsh&command=-c&command=ceph+fs+subvolume+metadata+ls+myfs+csi-vol-3ca56d9c-06a0-11ed-85d9-aedfd69586e0+--group_name%3De2e+--format%3Djson&container=rook-ceph-tools&container=rook-ceph-tools&stderr=true&stdout=true)
Jul 18 14:48:10.685: INFO: stdout=
{
    "csi.ceph.com/cluster/name": "k8s-cluster-1",
    "csi.storage.k8s.io/pv/name": "pvc-8552822b-6172-447b-9065-5a0d5954929a",
    "csi.storage.k8s.io/pvc/name": "csi-cephfs-pvc",
    "csi.storage.k8s.io/pvc/namespace": "cephfs-8081"
}
 stdErr= err=json: cannot unmarshal object into Go value of type []e2e.cephfsSubvolumeMetadata
Jul 18 14:48:10.685: FAIL: failed to list subvolume metadata: json: cannot unmarshal object into Go value of type []e2e.cephfsSubvolumeMetadata
```

pkalever: I think I got the problem

pkalever:
> > We need to support older ceph versions that do not support metadata operations. if ceph doesnot support metadata operations we need to skip setting metadata
> 
> Not just Quincy, the patches were backported to Pacific and Octopus I guess. I think ideally we should check on all those versions, right?

Maybe we can just look at the ceph CLI for available metadata options and decide whether the ceph version supports it or not? 

 
 As we also have snapshot metadata-related changes in a different PR #3168, is this something that we can cover in a separate PR?

Madhu-1:
> > > We need to support older ceph versions that do not support metadata operations. if ceph doesnot support metadata operations we need to skip setting metadata
> > 
> > 
> > Not just Quincy, the patches were backported to Pacific and Octopus I guess. I think ideally we should check on all those versions, right?
> 
> Maybe we can just look at the ceph CLI for available metadata options and decide whether the ceph version supports it or not?
> 
> As we also have snapshot metadata-related changes in a different PR #3168, is this something that we can cover in a separate PR?

We need to check on the server side  same like https://github.com/ceph/ceph-csi/blob/devel/internal/cephfs/core/volume.go#L35-L40 and skip metadata if its not supported

pkalever:
> We need to check on the server side same like https://github.com/ceph/ceph-csi/blob/devel/internal/cephfs/core/volume.go#L35-L40 and skip metadata if its not supported

Thanks for the hint, that helped! Please take a look at the changes.

pkalever:

[Logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.21/runs/4589/nodes/107/log/?start=0)

pkalever:
/retest ci/centos/mini-e2e-helm/k8s-1.23
[Logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.23/runs/2445/nodes/107/log/?start=0)

pkalever:
/retest ci/centos/mini-e2e/k8s-1.22
[Logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e_k8s-1.22/runs/3612/nodes/89/log/?start=0)

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,177 @@
...
+func (s *subVolumeClient) GetMetadata(key string) (string, error) {
+	// keyPresent checks whether corresponding clusterID key is present in clusterAdditionalInfo
+	var keyPresent bool
+	// verify if corresponding clusterID key is present in the map,
+	// and if not, initialize with default values(false).
+	if _, keyPresent = clusterAdditionalInfo[s.clusterID]; !keyPresent {
+		clusterAdditionalInfo[s.clusterID] = &localClusterState{}
+	}
+	// resize subvolume when either it's supported, or when corresponding
+	// clusterID key was not present.
```
Madhu-1: Need to update the comment.

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+		var invalid fsAdmin.NotImplementedError
+		// In case the error is other than invalid command return error to the caller.
+		if !errors.As(err, &invalid) {
+			return "", err
+		}
+	}
+	clusterAdditionalInfo[s.clusterID].subVolMetadataState = unsupported
+
+	return "", fmt.Errorf("operation GetMetadata is not supported")
```
Madhu-1: use errors.New

Madhu-1:
```suggestion
notSupErr:=errors.New("operation GetMetadata is not supported")
if  clusterAdditionalInfo[s.clusterID].subVolMetadataState== unsupported {
	return "", notSupErr
}
fsa, err := s.conn.GetFSAdmin()
if err != nil {
	return "", err
}
res, err := fsa.GetMetadata(s.FsName, s.SubvolumeGroup, s.VolID, key)
if err != nil {
  var invalid fsAdmin.NotImplementedError
 // In case the error is other than invalid command return error to the caller.
  if errors.As(err, &invalid) {
  	clusterAdditionalInfo[s.clusterID].subVolMetadataState = unsupported
  	  err=notSupErr
  }
  return "",err
  }
clusterAdditionalInfo[s.clusterID].subVolMetadataState = supported
return res, nil
}
```

Above suggestions makes it more readable?

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,177 @@
...
+	// keyPresent checks whether corresponding clusterID key is present in clusterAdditionalInfo
+	var keyPresent bool
+	// verify if corresponding clusterID key is present in the map,
+	// and if not, initialize with default values(false).
+	if _, keyPresent = clusterAdditionalInfo[s.clusterID]; !keyPresent {
+		clusterAdditionalInfo[s.clusterID] = &localClusterState{}
+	}
+	// resize subvolume when either it's supported, or when corresponding
+	// clusterID key was not present.
+	if clusterAdditionalInfo[s.clusterID].subVolMetadataState == unknown ||
```
Madhu-1: Move the logic of checking support to a helper function and reuse it?

internal/cephfs/core/metadata.go
pkalever: surely does.

internal/cephfs/core/metadata.go
pkalever: No, your previous suggestion for code change is good.

internal/cephfs/core/metadata.go
```diff
@@ -29,34 +32,82 @@ const (
 // GetMetadata gets custom metadata on the subvolume in a volume based on key
 // name.
 func (s *subVolumeClient) GetMetadata(key string) (string, error) {
+	notSupErr := errors.New("operation GetMetadata is not supported")
+	if clusterAdditionalInfo[s.clusterID].subVolMetadataState == unsupported {
```
nixpanic: this is a very long if-statement, and it repeats itself often. Much cleaner to have a helper function like
```go
func (s *subVolumeClient) supportsSubVolMetadata() bool {}
```

internal/cephfs/core/metadata.go
```diff
@@ -29,34 +32,82 @@ const (
...
+		return "", notSupErr
+	}
 	fsa, err := s.conn.GetFSAdmin()
 	if err != nil {
 		return "", err
 	}
+	res, err := fsa.GetMetadata(s.FsName, s.SubvolumeGroup, s.VolID, key)
+	if err != nil {
+		var invalid fsAdmin.NotImplementedError
```
nixpanic: this part should really not be copy/pasted. place is in the helper function mentioned above.

is this not simpler for checking the error type? No need to cast the error into `invalid` is it is not used anywhere.
```go
if errors.Is(err, fsAdmin.NotImplementedError) {
```

pkalever: If we use it directly we see this compilation error:
```
# github.com/ceph/ceph-csi/internal/cephfs/core
internal/cephfs/core/metadata.go:35:42: fsAdmin.NotImplementedError (type) is not an expression
internal/cephfs/core/metadata.go:79:22: fsAdmin.NotImplementedError (type) is not an expression
internal/cephfs/core/metadata.go:105:21: fsAdmin.NotImplementedError (type) is not an expression
make: *** [Makefile:160: cephcsi] Error 2
```

internal/cephfs/core/metadata.go
```diff
@@ -17,46 +17,97 @@ limitations under the License.
...
 const (
 	// clusterNameKey cluster Key, set on RBD image.
 	clusterNameKey = "csi.ceph.com/cluster/name"
 )
 
+func (s *subVolumeClient) supportsSubVolMetadata(err error) bool {
+	var invalid fsAdmin.NotImplementedError
+	if clusterAdditionalInfo[s.clusterID].subVolMetadataState == unsupported {
+		return false
+	} else if err != nil && errors.Is(err, &invalid) {
```
nixpanic: just use `errors.Is(err, fsAdmin.NotImplementedError)`, no need to assign it to `invalid` if you don't use that variable.

internal/cephfs/core/metadata.go
```diff
@@ -17,46 +17,97 @@ limitations under the License.
...
+	}
+	clusterAdditionalInfo[s.clusterID].subVolMetadataState = supported
+
+	return true
+}
+
 // GetMetadata gets custom metadata on the subvolume in a volume based on key
 // name.
 func (s *subVolumeClient) GetMetadata(key string) (string, error) {
+	notSupErr := errors.New("operation GetMetadata is not supported")
```
nixpanic: `notSupErr` can be a package wide variable, call it `ErrSubVolMetadataNotSupported` so that others can use `errors.Is()` to handle it.

reduce this to
```
if !s.supportsSubVolMetadata(nil) {
    return "", ErrSubVolMetadataNotSupported
}
```

internal/cephfs/core/metadata.go
pkalever:
```
[0] pkalever 😎 ceph-csi✨ make
GOOS=linux go build -tags=quincy,ceph_preview,ceph_ci_untested -mod vendor -a -ldflags '-X github.com/ceph/ceph-csi/internal/util.GitCommit=310cbc2025acc7a7703ce67582f1060c13f3d735 -X github.com/ceph/ceph-csi/internal/util.DriverVersion=canary' -o _output/cephcsi ./cmd/
# github.com/ceph/ceph-csi/internal/cephfs/core
internal/cephfs/core/metadata.go:35:41: fsAdmin.NotImplementedError (type) is not an expression
make: *** [Makefile:160: cephcsi] Error 2
[2] pkalever 👿 ceph-csi✨ git diff
diff --git a/internal/cephfs/core/metadata.go b/internal/cephfs/core/metadata.go
index 2dfdcf786..33be13419 100644
--- a/internal/cephfs/core/metadata.go
+++ b/internal/cephfs/core/metadata.go
@@ -30,10 +30,9 @@ const (
 )
 
 func (s *subVolumeClient) supportsSubVolMetadata(err error) bool {
-       var invalid fsAdmin.NotImplementedError
        if clusterAdditionalInfo[s.clusterID].subVolMetadataState == unsupported {
                return false
-       } else if err != nil && errors.Is(err, &invalid) {
+       } else if err != nil && errors.Is(err, fsAdmin.NotImplementedError) {
                // In case the error is other than invalid command return error to the caller.
                clusterAdditionalInfo[s.clusterID].subVolMetadataState = unsupported

```

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	clusterAdditionalInfo[s.clusterID].subVolMetadataState = supported
+
+	return true
+}
+
+// GetMetadata gets custom metadata on the subvolume in a volume based on key
+// name.
+func (s *subVolumeClient) GetMetadata(key string) (string, error) {
+	if !s.supportsSubVolMetadata(nil) {
```
Madhu-1: I dont think it's a better one. passing nil and err and making the decision on that one is not required; if this is required, have a new helper function to check the clusterAdditionalInfo and return the required error message.

Madhu-1 (commented):
changes LGTM  but i have a concern on the way we are handing clusterAdditionalInfo check

internal/cephfs/core/metadata.go
pkalever: I just optimized it a bit, may be passing `nil` directly made you think something is not right, will be okay if I make it to:

```
func (s *subVolumeClient) supportsSubVolMetadata(err error) bool {
        var invalid fsAdmin.NotImplementedError
        if clusterAdditionalInfo[s.clusterID].subVolMetadataState == unsupported {
                return false
        } else if err != nil && errors.Is(err, &invalid) {
                // In case the error is other than invalid command return error to the caller.
                clusterAdditionalInfo[s.clusterID].subVolMetadataState = unsupported

                return false
        }
        clusterAdditionalInfo[s.clusterID].subVolMetadataState = supported

        return true
}

// GetMetadata gets custom metadata on the subvolume in a volume based on key
// name.
func (s *subVolumeClient) GetMetadata(key string) (string, error) {
        var err error
        if !s.supportsSubVolMetadata(err) {
                return "", ErrSubVolMetadataNotSupported
        }
        fsa, err := s.conn.GetFSAdmin()
        if err != nil {
                return "", err
        }
        res, err := fsa.GetMetadata(s.FsName, s.SubvolumeGroup, s.VolID, key)
        if !s.supportsSubVolMetadata(err) {
                return "", ErrSubVolMetadataNotSupported
        }
        if err != nil {
                return "", err
        }

        return res, nil
}
```

pkalever: Logically there is nothing wrong here!!

nixpanic: I think I prefer two separate helpers to make their functioning clearer and easier to understand.

`s.isUnsupportedSubVolMetadata(err)`, pass the error, set either `supported` or `unsupported` and return 

`s.supportsSubVolMetadata()`, no error, just returns a bool. If the support was not checked before, it can do the checking if needed, or return `true` when in doubt. If it does checking, it should use `s.isUnsupportedSubVolMetadata(err)` to cache the support.

internal/cephfs/controllerserver.go
```diff
@@ -311,6 +315,15 @@ func (cs *ControllerServer) CreateVolume(
 			}
 		}
 
+		if !volOptions.BackingSnapshot {
+			// Set metadata on restart of provisioner pod when subvolume exist
+			metadata := k8s.GetVolumeMetadata(req.GetParameters())
```
Madhu-1: we can move this outside of if block and use it, For now we are calling `k8s.GetVolumeMetadata(req.GetParameters())` twice in createVolume

internal/cephfs/controllerserver.go
```diff
@@ -658,7 +679,7 @@ func (cs *ControllerServer) ControllerExpandVolume(
 
 	RoundOffSize := util.RoundOffCephFSVolSize(req.GetCapacityRange().GetRequiredBytes())
 
-	volClient := core.NewSubVolume(volOptions.GetConnection(), &volOptions.SubVolume, volOptions.ClusterID)
+	volClient := core.NewSubVolume(volOptions.GetConnection(), &volOptions.SubVolume, volOptions.ClusterID, "")
```
Madhu-1: why not pass clusterName instead of an empty string in all cases, as we will have clusterName?

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	fsAdmin "github.com/ceph/go-ceph/cephfs/admin"
+)
+
+const (
+	// clusterNameKey cluster Key, set on RBD image.
```
Madhu-1: this is for the cephfs subvolume, not RBD image?  reuse the clusterNameKey we have for RBD instead of defining separate one for cephfs?

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	err = fsa.SetMetadata(s.FsName, s.SubvolumeGroup, s.VolID, key, value)
+	if !s.isUnsupportedSubVolMetadata(err) {
+		return ErrSubVolMetadataNotSupported
+	}
+	if err != nil {
+		return err
+	}
+
+	return nil
```
Madhu-1: just return err should be enough?

internal/cephfs/core/metadata.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	err = fsa.RemoveMetadata(s.FsName, s.SubvolumeGroup, s.VolID, key)
+	if !s.isUnsupportedSubVolMetadata(err) {
+		return ErrSubVolMetadataNotSupported
+	}
+	if err != nil {
+		return err
+	}
+
+	return nil
```
Madhu-1: just return err?

Madhu-1 (commented):
some nits

internal/cephfs/controllerserver.go
pkalever: Sure I can pass wherever available.

Madhu-1 (dismissed):

Madhu-1 (dismissed):

Rakshith-R (dismissed):

pkalever:
@Mergifyio rebase

pkalever:
/retest ci/centos/upgrade-tests-cephfs

[Logs](https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/rest/organizations/jenkins/pipelines/upgrade-tests-cephfs/runs/6238/nodes/93/log/?start=0)

pkalever:
@Mergifyio rebase

Rakshith-R (approved):

Madhu-1 (approved):

PR #3154, opened 2022-06-02
cephfs: skip NetNamespaceFilePath if the volume is pre-provisioned

### LINKED ISSUE #3153: cephfs: GetCephFSNetNamespaceFilePath fails if config.json is missing
# Describe the bug #

`GetCephFSNetNamespaceFilePath` fails if `config.json` (supplied by `ceph-csi-config` ConfigMap) is missing. Specifically, the code shown below expects the config to be there:

https://github.com/ceph/ceph-csi/blob/2655d2edca323c94d25bd64b44306d7997921fe0/internal/cephfs/nodeserver.go#L129-L134

https://github.com/ceph/ceph-csi/blob/2655d2edca323c94d25bd64b44306d7997921fe0/internal/util/csiconfig.go#L188-L196

In deployments where the config is missing, it makes mounting volumes impossible. `NodeStageVolume` always fails.

This may be true for rbd-csi too.

# Steps to reproduce #

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: my-cephfs-pv
spec:
  accessModes:
  - ReadWriteMany
  capacity:
    storage: 1Gi
  csi:
    driver: cephfs.csi.ceph.com
    volumeHandle: my-cephfs-volume
    nodeStageSecretRef:
      name: csi-cephfs-secret
      namespace: default
    volumeAttributes:
      provisionVolume: "false"
      monitors: "..."
      rootPath: "..."

      # clusterID is not specified. We're passing monitor IPs directly with `monitors` parameter.

---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-cephfs-pvc
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
  volumeName: my-cephfs-pv
  storageClassName: ""
---
apiVersion: v1
kind: Pod
metadata:
  name: my-cephfs-pod
spec:
  containers:
    - name: web-server
      image: nginx
      imagePullPolicy: IfNotPresent
      volumeMounts:
        - name: mypvc
          mountPath: /var/lib/www
  volumes:
    - name: mypvc
      persistentVolumeClaim:
        claimName: my-cephfs-pvc
        readOnly: false
```

# Actual results #

Mounting the volume fails with:

```
E0515 08:52:55.398271       1 utils.go:200] ID: 10 Req-ID: ... GRPC error: rpc error: code = Internal desc = missing configuration for cluster ID ""
```

# Expected behavior #

Mounting should have succeeded despite the missing config.

# Additional context #

* `netNamespaceFilePath` should be treated as an optional parameter. Skip it silently, if it cannot be retrieved.
* I'm wondering why is this even part of the per-cluster config? Is the net ns file path expected to be different across different cluster configurations? Maybe we could pass it to the driver via cmd args / env variable too?

Madhu-1:
In the case of pre-provisionVolume the clusterID is not set in the volume context as the clusterID is missing
we cannot extract the NetNamespaceFilePath from the configuration file. For static volume and dynamically provisioned volume, the clusterID is set.

Note:- This is a special case to support mounting PV without clusterID parameter.

closes #3153 

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -126,12 +126,18 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 	defer volOptions.Destroy()
 
-	volOptions.NetNamespaceFilePath, err = util.GetCephFSNetNamespaceFilePath(
-		util.CsiConfigFile,
-		volOptions.ClusterID)
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+	// Skip extracting NetNamespaceFilePath if the clusterID is empty.
+	// In case of pre-provisioned volume the clusterID is not set in the
+	// volume context.
+	if volOptions.ClusterID != "" {
+		volOptions.NetNamespaceFilePath, err = util.GetCephFSNetNamespaceFilePath(
+			util.CsiConfigFile,
+			volOptions.ClusterID)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
 	}
+
 	mnt, err := mounter.New(volOptions)
 	if err != nil {
 		log.ErrorLog(ctx, "failed to create mounter for volume %s: %v", volID, err)


Rakshith-R:
What is a `pre-provisionVolume` ?

internal/cephfs/nodeserver.go
```diff
@@ -126,12 +126,20 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 	defer volOptions.Destroy()
 
-	volOptions.NetNamespaceFilePath, err = util.GetCephFSNetNamespaceFilePath(
-		util.CsiConfigFile,
-		volOptions.ClusterID)
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+	// In case of pre-provisionVolume volume the clusterID is not set in the
```
gman0:
```suggestion
	// In case of pre-provisioned volume the clusterID is not set in the
```

Madhu-1:
> What is a `pre-provisionVolume` ?

it was there in 1.0 https://github.com/ceph/ceph-csi/blob/csi-v1.0/examples/cephfs/storageclass.yaml#L12-L19 we kept it on request.

internal/cephfs/nodeserver.go
```diff
@@ -126,12 +126,20 @@ func (ns *NodeServer) NodeStageVolume(
...
-		util.CsiConfigFile,
-		volOptions.ClusterID)
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+	// In case of pre-provisioned volume the clusterID is not set in the
+	// volume context as the clusterID is missing we cannot extract the
+	// NetNamespaceFilePath from the configuration file.
+	// For static volume and dynamically provisioned volume the clusterID is set.
+	// skip extracting NetNamespaceFilePath if the volume is pre-provisioned.
+	if _, ok := req.GetVolumeContext()["provisionVolume"]; !ok {
```
gman0:
```suggestion
	if volOptions.ClusterID != "" {
```

I'm wondering if this would work too? Since it's only pre-provisioned volumes that don't set `clusterID`, this also implies that `provisionVolume: "false"` (otherwise [`getVolumeOptions()`](https://github.com/ceph/ceph-csi/blob/3c0e1b4970cd1213687977e86d8c5d5817e0946c/internal/cephfs/nodeserver.go#L75-L101) would have failed).

This would make the comment explaining why this is needed a lot shorter too. And it does exactly what we need: `GetCephFSNetNamespaceFilePath()` requires `clusterID` to be present -- so call it only if it is in fact set.

Madhu-1: Yes this also works, updated it

gman0 (approved):
/lgtm

Thank you, @Madhu-1!

Rakshith-R (approved):

nixpanic (approved):

Issue #3103, opened 2022-05-09
Cleanup bug in createVolumeFromSnapshot() and CreateSnapshot()


pkalever:
# Describe the bug #

1. In createVolumeFromSnapshot(), if the unsetAllMetadata or copyEncryptionConfig or expand fails we need to remove the clone. 
2. Same in CreateSnapshot() too 

See:
https://github.com/ceph/ceph-csi/pull/3005#discussion_r867690813
https://github.com/ceph/ceph-csi/pull/3005#discussion_r867691871

Madhu-1:
@yati1998 PTAL

yati1998:
Closing this issue since the [fix](https://github.com/ceph/ceph-csi/pull/3155) is merged.

PR #3155, opened 2022-06-02
rbd: remove the clone in case of failure

yati1998:
This commit removes the clone incase
unsetAllMetadata or copyEncryptionConfig or
expand fails for createVolumeFromSnapshot
and CreateSnapshot.
It also removes the clone in case of
any failure in createCloneFromImage.

issue: #3103

### DIFF
--- a/internal/rbd/clone.go
+++ b/internal/rbd/clone.go
@@ -138,6 +138,16 @@ func (rv *rbdVolume) createCloneFromImage(ctx context.Context, parentVol *rbdVol
 		return err
 	}
 
+	defer func() {
+		if err != nil {
+			log.DebugLog(ctx, "Removing clone image %q", rv)
+			errDefer := rv.deleteImage(ctx)
+			if errDefer != nil {
+				log.ErrorLog(ctx, "failed to delete clone image %q: %v", rv, errDefer)
+			}
+		}
+	}()
+
 	err = rv.getImageID()
 	if err != nil {
 		log.ErrorLog(ctx, "failed to get volume id %s: %v", rv, err)
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -644,6 +644,17 @@ func (cs *ControllerServer) createVolumeFromSnapshot(
 
 		return err
 	}
+
+	defer func() {
+		if err != nil {
+			log.DebugLog(ctx, "Removing clone image %q", rbdVol)
+			errDefer := rbdVol.deleteImage(ctx)
+			if errDefer != nil {
+				log.ErrorLog(ctx, "failed to delete clone image %q: %v", rbdVol, errDefer)
+			}
+		}
+	}()
+
 	err = rbdVol.unsetAllMetadata(k8s.GetSnapshotMetadataKeys())
 	if err != nil {
 		log.ErrorLog(ctx, "failed to unset snapshot metadata on rbd image %q: %v", rbdVol, err)
@@ -1141,6 +1152,16 @@ func (cs *ControllerServer) CreateSnapshot(
 	// Update the metadata on snapshot not on the original image
 	rbdVol.RbdImageName = rbdSnap.RbdSnapName
 	rbdVol.ClusterName = cs.ClusterName
+
+	defer func() {
+		if err != nil {
+			log.DebugLog(ctx, "Removing clone image %q", rbdVol)
+			errDefer := rbdVol.deleteImage(ctx)
+			if errDefer != nil {
+				log.ErrorLog(ctx, "failed to delete clone image %q: %v", rbdVol, errDefer)
+			}
+		}
+	}()
 
 	err = rbdVol.unsetAllMetadata(k8s.GetVolumeMetadataKeys())
 	if err != nil {


internal/rbd/controllerserver.go
```diff
@@ -591,9 +591,22 @@ func (cs *ControllerServer) createVolumeFromSnapshot(
 
 		return err
 	}
+	deleteClone := true
+
```
pkalever, line 657: May be move the defer here and set the `deleteClone = false` once all the operations succeeds

```
    deleteClone := true                                                              
    defer func() {                                                                   
        if deleteClone {                                                             
            err = librbd.RemoveImage(rbdVol.ioctx, rbdVol.RbdImageName)                      
            if err != nil {                                                          
                log.ErrorLog(ctx, "failed to delete clone image %q: %v", rbdVol, err)
            }                                                                        
        }                                                                            
    }()                                                                              
    err = rbdVol.unsetAllMetadata(k8s.GetSnapshotMetadataKeys())
    [...]
    err = parentVol.copyEncryptionConfig(&rbdVol.rbdImage, true)
    [...]
    err = rbdVol.expand()
    [...]
    deleteClone = false

    return nil
}
```

internal/rbd/controllerserver.go
```diff
@@ -610,9 +634,21 @@ func (cs *ControllerServer) createVolumeFromSnapshot(
...
+				}
+			}
+		}()
 		return err
 	}
+	// Success! Do not delete the cloned image now :)
+	deleteClone = false
 
 	return nil
 }
```
pkalever: You need to do similar fix in CreateSnapshot()

yati1998: yeah will add it in different commit

internal/rbd/controllerserver.go
Madhu-1: you dont need a variable. do `if err!=nil` check in defer and delete the clone on failure.

internal/rbd/controllerserver.go
Madhu-1: please take care of pvc-pvc clone case also.

yati1998:
@Madhu-1 @pkalever @nixpanic would like to get review on this.

internal/rbd/clone.go
```diff
@@ -138,6 +138,16 @@ func (rv *rbdVolume) createCloneFromImage(ctx context.Context, parentVol *rbdVol
 		return err
 	}
 
+	defer func() {
+		if err != nil {
+			log.DebugLog(ctx, "Removing the clone image %q:", rv)
+			errDefer := librbd.RemoveImage(rv.ioctx, rv.RbdImageName)
```
nixpanic: why not call `rv.deleteImage(ctx)`?

internal/rbd/controllerserver.go
```diff
@@ -591,6 +591,17 @@ func (cs *ControllerServer) createVolumeFromSnapshot(
 
 		return err
 	}
+
+	defer func() {
+		if err != nil {
+			log.DebugLog(ctx, "Removing the clone image %q:", rbdVol)
+			errDefer := librbd.RemoveImage(rbdVol.ioctx, rbdVol.RbdImageName)
```
nixpanic: same here... is there a reason to call `librbd.RemoveImage()` instead of the more complete `rbdVol.deleteImage()`?

nixpanic (changes requested):
in general, the ControllerServer should not call librbd functions directly, only the methods of an `rbdImage` object.

internal/rbd/controllerserver.go
yati1998: No specific reason, will update

internal/rbd/controllerserver.go
```diff
@@ -1135,6 +1146,16 @@ func (cs *ControllerServer) CreateSnapshot(
 	rbdVol.RbdImageName = rbdSnap.RbdSnapName
 	rbdVol.ClusterName = cs.ClusterName
 
+	defer func() {
+		if err != nil {
+			log.DebugLog(ctx, "Removing the clone image %q:", rbdVol)
+			errDefer := librbd.RemoveImage(rbdVol.ioctx, rbdVol.RbdImageName)
```
nixpanic: why does this call `librbd.RemoveImage()` and not `rbdVol.deleteImage()`?

nixpanic (dismissed):

Madhu-1 (approved):

internal/rbd/clone.go
```diff
@@ -138,6 +138,16 @@ func (rv *rbdVolume) createCloneFromImage(ctx context.Context, parentVol *rbdVol
 		return err
 	}
 
+	defer func() {
+		if err != nil {
+			log.DebugLog(ctx, "Removing the clone image %q:", rv)
```
Madhu-1:
```suggestion
log.DebugLog(ctx, "Removing the clone image %q", rv)
```

internal/rbd/controllerserver.go
```diff
@@ -638,6 +638,17 @@ func (cs *ControllerServer) createVolumeFromSnapshot(
 
 		return err
 	}
+
+	defer func() {
+		if err != nil {
+			log.DebugLog(ctx, "Removing the clone image %q:", rbdVol)
```
Madhu-1:
```suggestion
log.DebugLog(ctx, "Removing the clone image %q", rbdVol)
```

internal/rbd/controllerserver.go
```diff
@@ -638,6 +638,17 @@ func (cs *ControllerServer) createVolumeFromSnapshot(
...
+			errDefer := rbdVol.deleteImage(ctx)
+			if errDefer != nil {
+				log.ErrorLog(ctx, "failed to delete the clone image %q: %v", rbdVol, errDefer)
```
Madhu-1:
```suggestion
log.ErrorLog(ctx, "failed to delete clone image %q: %v", rbdVol, errDefer)
```

internal/rbd/clone.go
```diff
@@ -138,6 +138,16 @@ func (rv *rbdVolume) createCloneFromImage(ctx context.Context, parentVol *rbdVol
...
+			errDefer := rv.deleteImage(ctx)
+			if errDefer != nil {
+				log.ErrorLog(ctx, "failed to delete the clone image %q: %v", rv, errDefer)
```
Madhu-1:
```suggestion
log.ErrorLog(ctx, "failed to delete clone image %q: %v", rv, errDefer)
```

Madhu-1 (changes requested):
didn't notice it earlier. some nits

Madhu-1 (approved):

nixpanic (approved):

PR #3157, opened 2022-06-02
ci: use golangci-lint --fix instead of go-fmt in pre-commit

irq0:
`make test` no longer relies on go-fmt, but rather gofumpt. To catch as
many linters as possible use golangci-lint just like make test does.

Signed-off-by: Marcel Lauhoff <marcel.lauhoff@suse.com>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,11 +7,11 @@ repos:
     hooks:
       - id: check-signoff
 
-  # Catch gofmt issues, if any.
-  - repo: git://github.com/dnephin/pre-commit-golang
-    rev: v0.3.5
+  - repo: https://github.com/golangci/golangci-lint
+    rev: v1.46.2
     hooks:
-      - id: go-fmt
+      - id: golangci-lint
+        entry: golangci-lint run --config=scripts/golangci.yml --fix --new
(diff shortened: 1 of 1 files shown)


yati1998 (approved):
LGTM

Yuggupta27 (commented):
The main idea behind just checking the `gofmt` errors in pre-commit stage was to keep the pre-commit not too strict.
But we can go with checking all possible golang-ci-lint-issues too :+1:

.pre-commit-config.yaml
```diff
@@ -7,11 +7,11 @@ repos:
...
 
-  # Catch gofmt issues, if any.
-  - repo: git://github.com/dnephin/pre-commit-golang
-    rev: v0.3.5
+  - repo: https://github.com/golangci/golangci-lint
+    rev: v1.46.2
     hooks:
-      - id: go-fmt
+      - id: golangci-lint
+        entry: golangci-lint run --config=scripts/golangci.yml --fix --new
```
nixpanic, line 14: `scripts/golangci.yml` is generated, you will need to run `make scripts/golangci.yml` first

nixpanic (changes requested):

PR #3158, opened 2022-06-02
RFC: RBD+CephFS fscrypt support #1563

irq0:
Add fscrypt integration with the Ceph CSI KMS. Supports Ceph FS and ext4 on RBD storage classes.
Implementation is based on the design doc https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/cephfs-fscrypt.md.

Testing this is currently non trivial. ext4 requires kernel support (see https://github.com/google/fscrypt#runtime-dependencies; Ceph FS requires a custom built kernel and a Ceph MDS from main. Unfortunately minikube currently doesn't have neither.

k8s config and tools to run this with a vstart cluster are in https://github.com/irq0/dev-ceph-csi-fscrypt-config


humblec:
Thanks @irq0 !

go.mod
```diff
@@ -42,6 +43,11 @@ require (
 	sigs.k8s.io/controller-runtime v0.11.0-beta.0.0.20211208212546-f236f0345ad2
 )
 
+require (
```
nixpanic: I do not understand why this is placed in a new `require` block. If it is done by tooling :shrug: , otherwise it should be fine to just have it combined in other blocks.

vendor/modules.txt
```diff
@@ -212,6 +212,15 @@ github.com/golang/protobuf/ptypes/wrappers
 # github.com/golang/snappy v0.0.4
 ## explicit
 github.com/golang/snappy
+# github.com/google/fscrypt v0.3.3
```
nixpanic, line 215: you will also need to `go mod vendor` the updates and additions. Ideally in a separate commit, as that makes reviewing a little easier.

internal/kms/kms.go
```diff
@@ -93,6 +93,15 @@ func GetKMS(tenant, kmsID string, secrets map[string]string) (EncryptionKMS, err
 	return kmsManager.buildKMS(tenant, kmsConfig, secrets)
 }
 
+func GetKMSTestDummy(kmsID string) EncryptionKMS {
```
nixpanic, line 96: if this is only used for unit testing, can it be placed in `kms_test.go` instead? Maybe add a
```go
type ProviderTest struct {
	Provider
	CreateTestDummy TestDummyFunc
}
```
in the `kms_test.go` for that as well? Not sure if that is sufficient though...

internal/cephfs/fscrypt_test.go
```diff
@@ -0,0 +1,37 @@
+package cephfs
```
nixpanic: please include the licence/copyright header

internal/cephfs/fscrypt_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
```
nixpanic: use `"context"` instead?

internal/cephfs/nodeserver.go
```diff
@@ -105,10 +100,21 @@ func (ns *NodeServer) getVolumeOptions(
 	return volOptions, nil
 }
 
+func MaybeFscryptUnlock(ctx context.Context, volOptions *store.VolumeOptions,
```
nixpanic: New (exported) functions ideally have comments.

Does it make sense to have this as a method of `store.VolumeOptions`?

internal/util/fscrypt.go
```diff
@@ -1,4 +1,4 @@
-package cephfs
+package util
```
nixpanic: The `util` package is really large already. If you can move this into `util/encryption` or something similar, please do!

internal/rbd/controllerserver.go
```diff
@@ -1469,7 +1469,7 @@ func (cs *ControllerServer) ControllerExpandVolume(
 	// 2. Block VolumeMode with Encryption
 	// Hence set nodeExpansion flag based on VolumeMode and Encryption status
 	nodeExpansion := true
-	if req.GetVolumeCapability().GetBlock() != nil && !rbdVol.isEncrypted() {
+	if req.GetVolumeCapability().GetBlock() != nil && !rbdVol.isBlockEncrypted() {
```
nixpanic, line 1559: the commit message contains

> include the world 'block'

You probably mean s/world/word/

internal/rbd/nodeserver.go
```diff
@@ -685,6 +699,17 @@ func (ns *NodeServer) NodePublishVolume(
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
 
+	fileEncrypted, err := util.IsEncrypted(req.GetVolumeContext())
+	if err != nil {
+		return nil, err
```
nixpanic: this function needs to return a gRPC error. Functions from the `util` package should only return 'plain' errors.

internal/rbd/nodeserver.go
```diff
@@ -700,7 +725,8 @@ func (ns *NodeServer) mountVolumeToStagePath(
 	ctx context.Context,
 	req *csi.NodeStageVolumeRequest,
 	staticVol bool,
-	stagingPath, devicePath string) error {
+	stagingPath, devicePath string, fileEncryption bool,
```
nixpanic: coding guidelines suggest to place new parameters of different types on the next line

internal/util/fscrypt.go
```diff
@@ -210,15 +217,49 @@ func initializeAndUnlock(ctx context.Context, fscryptContext *fscryptactions.Con
...
-	_, err := xattr.Get(directoryPath, "ceph.fscrypt.auth")
-	if err != nil {
-		return fmt.Errorf("error reading ceph.fscrypt.auth xattr on %q: %w", directoryPath, err)
+	switch filesystem {
+	case "cephfs":
+		_, err := xattr.Get(directoryPath, "ceph.fscrypt.auth")
+		if err != nil {
+			return fmt.Errorf("error reading ceph.fscrypt.auth xattr on %q: %w", directoryPath, err)
+		}
+	case "ext4":
```
nixpanic: maybe this should be the default case?

internal/util/fscrypt.go
```diff
@@ -210,15 +217,49 @@ func initializeAndUnlock(ctx context.Context, fscryptContext *fscryptactions.Con
...
+func IsDirectoryUnlockedFscrypt(directoryPath, filesystem string) error {
 	if _, err := fscryptmetadata.GetPolicy(directoryPath); err != nil {
 		return fmt.Errorf("no fscrypt policy set on directory %q: %w", directoryPath, err)
 	}
 
-	_, err := xattr.Get(directoryPath, "ceph.fscrypt.auth")
-	if err != nil {
-		return fmt.Errorf("error reading ceph.fscrypt.auth xattr on %q: %w", directoryPath, err)
+	switch filesystem {
+	case "cephfs":
```
nixpanic: above you use `"ceph"`, here you check `"cephfs"`... Which is the right name? Might be better to make it a constant :wink:

internal/util/fscrypt.go
```diff
@@ -210,15 +217,49 @@ func initializeAndUnlock(ctx context.Context, fscryptContext *fscryptactions.Con
 	return nil
 }
 
+func getInodeEncryptedAttribute(p string) (bool, error) {
```
nixpanic: a little comment on what this function checks would be nice

nixpanic (commented):
Thanks for the PR! Some initial comments for now, I'll try to keep following the progress you're making.

internal/util/fscrypt.go
irq0: Makes sense, but let me double check if other filesystems with fscrypt support set the inode encryption flag.

internal/util/fscrypt.go
irq0: Good catch! It should be "ceph", as that is what the kernel calls cephfs.

internal/kms/kms.go
irq0: Good idea. I'll look into that.

internal/util/fscrypt.go
irq0: Checked: ext4 and f2fs set the inode encrypted flag. Ceph FS doesn't support inode flags. Changing the ext4 to the default case.

internal/kms/kms.go
irq0: Turned out a little bit different, but I managed to move the test dummy stuff out of the Provider struct

irq0:
I renamed the branch. Apparently this closes attached PRs :). New PR is https://github.com/ceph/ceph-csi/pull/3310

PR #3007, opened 2022-04-11
rbd: set CO cluster Name as metadata on the image

pkalever:
# Describe what this PR does #

This change helps read the CO cluster name from the environment variable
and if the variable exists ceph-csi will set the same on the RBD images.

We need users to set cluster name in the env at deployment time so that
ceph-csi will know about the cluster name info.

## Related issues ##

Fixes: #2973

Signed-off-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)
- `/retest all`: run this in case the CentOS CI failed to start/report any test
  progress or results

</details>

PR #3160, opened 2022-06-03
ci: check more reviews in retest action

nixpanic:
It seems PullRequests.ListReviews() returns 30 reviews by default. This
is practical for pagination, but not so much for a GitHub action. There
is the rare occasion that the number of reviews if higher than 30. If
that is the case, the retest action does not capture the latest reviews,
which is where the APPROVED status is set.

By increasing the number of results to 100 per page, we should have
sufficient space for many more reviews and automated retest-ing.

---

<details>

  progress or results

</details>


nixpanic:
#3007 is a current problematic PR

Madhu-1 (approved):

yati1998 (approved):

PR #2507, opened 2021-09-14
rbd: Report errors when a resync maybe in progress

ShyamsundarR:
Currently we return a !ready status if an image
is not found when a replication resync is issued.

We also return a !ready just post issuing a resync.

The change is to ensure we return errors in these
cases for the caller to retry the operation till
we can determine we are actually resyncing, and then
return !ready with nil errors.

Part of addressing:
  https://github.com/csi-addons/volume-replication-operator/issues/101

Signed-off-by: Shyamsundar Ranganathan <srangana@redhat.com>

PR #3206, opened 2022-06-23
rbd: issue resync only if the force flag is set

### LINKED ISSUE #3209: [Replication] Data loss due to resync because of the spurious split-brain errors reported by RBD
During relocation when demoting the current primary,  the image will move the secondary and will be in split-brain state for sometime
As there are spurious split-brain errors reported by RBD, the resync will attempt to resync from the "known" secondary and that will cause data loss, From VolRep stack is aware that it is a relocate and hence can prevent any resync call in such situations, which is a required fix

Brief details about the fix:
- CSI ResyncVolume will not issue a resync till the force option is provided, and report split-brain errors till then
- VolRep will take an additional "autoresync" flag for resync, that can be used to drive the "force" request to CSI ResyncVolume
   - IOW: spec.autoresync: true/false in VolRep which is done in https://github.com/csi-addons/volume-replication-operator/pull/133
- VRG/Ramen will accept action (or equivalent) to understand that the current move from Primary->Secondary is a relocate and hence pass autoresync "false" to VolRep
- DRPC will update VRG action for the above based on its action

Expected behaviour:
- A relocation may still get stuck based on the RBD spurious errors, but will prevent the data loss
  - If the RBD error is transient, then the periodic RBD mirror deamon restart will help relocation progress
  - If the RBD error is non-transient, it will remain stuck, but with other RBD fixes this situation would be rectified

Madhu-1:
During failover we do demote the volume on the primary as the image is still not promoted yet on the remote cluster,
there are spurious split-brain errors reported by RBD, the Cephcsi resync will attempt to resync from the "known" secondary and that will cause data loss

fixes https://github.com/ceph/ceph-csi/issues/3209

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -856,18 +856,11 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		ready = checkRemoteSiteStatus(ctx, mirrorStatus)
 	}
 
-	if resyncRequired(localStatus) {
-		err = rbdVol.resyncImage()
-		if err != nil {
-			log.ErrorLog(ctx, err.Error())
+	err = resyncVolume(localStatus, rbdVol, req.Force)
+	if err != nil {
+		log.ErrorLog(ctx, err.Error())
 
-			return nil, status.Error(codes.Internal, err.Error())
-		}
-
-		// If we issued a resync, return a non-final error as image needs to be recreated
-		// locally. Caller retries till RBD syncs an initial version of the image to
-		// report its status in the resync request.
-		return nil, status.Error(codes.Unavailable, "awaiting initial resync due to split brain")
+		return nil, err
 	}
 
 	err = checkVolumeResyncStatus(localStatus)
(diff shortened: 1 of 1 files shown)


ShyamsundarR (approved):

Madhu-1:
Skipping CI as we dont have E2E for DR workflow

Rakshith-R (approved):

internal/rbd/replicationcontrollerserver.go
```diff
@@ -887,6 +880,29 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+		}
+		err := rbdVol.resyncImage()
+		if err != nil {
+			return status.Error(codes.Internal, err.Error())
+		}
+
+		// If we issued a resync, return a non-final error as image needs to be recreated
+		// locally. Caller retries till RBD syncs an initial version of the image to
+		// report its status in the resync request.
+		return status.Error(codes.Unavailable, "awaiting initial resync due to split brain")
```
nixpanic, line 900: `UNAVAILABLE` does not seem to be an error listed in [spec](https://github.com/csi-addons/spec/tree/main/replication#error-scheme-4)? Should this be `ABORTED` instead?

nixpanic (changes requested):

internal/rbd/replicationcontrollerserver.go
Madhu-1: Not part of this PR, it was added earlier

Madhu-1: As discussed will update the spec for the new error code

nixpanic: Thanks, it seems this is intentional https://github.com/ceph/ceph-csi/pull/2507#pullrequestreview-754755312

Madhu-1: Yes correct.

nixpanic (approved):

Madhu-1:
> /retest all

retesting all as skip-ci label is added

PR #3203, opened 2022-06-23
rebase: update minikube to v1.26.0

pkalever:
# Describe what this PR does #

A new stable release of minikube is available, let's switch to it.
https://github.com/kubernetes/minikube/releases/tag/v1.26.0

Signed-off-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)
- `/retest all`: run this in case the CentOS CI failed to start/report any test
  progress or results

</details>

PR #3214, opened 2022-06-27
ci: prevent panic in retest action on `nil` strings

nixpanic:
In case a PullRequest does not have a MergeableState set, it will be
`nil`. Dereferencing the pointer will cause a Go panic, and the action
won't work as intended.

---

<details>

  progress or results

</details>

### DIFF
--- a/actions/retest/main.go
+++ b/actions/retest/main.go
@@ -116,7 +116,7 @@ func main() {
 		log.Fatalf("failed to list pull requests %v\n", err)
 	}
 	for _, re := range req {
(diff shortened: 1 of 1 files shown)


Rakshith-R (approved):
Thanks

nixpanic:
The original failure can be seen in https://github.com/ceph/ceph-csi/runs/7068151815?check_suite_focus=true

Madhu-1 (approved):

nixpanic:
Added `DNM` label so that #3203 can get merged first

PR #3216, opened 2022-06-27
revert: rbd: consider remote image health for primary

### LINKED ISSUE #3215: [Replication] Failed to move image to primary after Failover
Was running some tests with the entire stack for DR, and the issue that crops up with this secondary image check is that the VR is never marked Primary post a failover, IOW this condition remains false:

```
      reason: FailedToPromote
      status: "False"
      type: Completed
```
The reason being the image reports:

```
csi-vol-851eac94-f402-11ec-9b37-0242ac11000e:
  global_id:   461b4974-eabd-42c4-89c8-eb3870f42137
  state:       up+stopped
  description: local image is primary
  service:     a on hub
  last_update: 2022-06-24 21:23:17
  peer_sites:
    name: 77ae0133-c330-4c67-922b-77d8e9afd0b2
    state: down+stopped
    description: local image is primary
    last_update: 2022-06-24 21:23:34
```
Overall the reason for adding this was to ensure the local image once enabled is marked for replication, instead of assuming that it is, hence the local check will suffice for now. In the (near) future as we add a periodic update to local status based on image info, that will give a clearer picture of replication state and last replicated instance time, etc.

Originally posted at https://github.com/ceph/ceph-csi/pull/3008#issuecomment-1166001328 by @ShyamsundarR

Madhu-1:
When the image is force promoted to primary on the cluster the remote image might not be in a replaying
state because due to the split-brain state. This PR reverts back the commit c3c87f2ef33e8d8ad08d7d9f28b59d1aedc4ef31. Which we added to check the remote image status.

closes: #3215 

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -581,7 +581,12 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 		}
 	}
 
-	err = checkHealthyPrimary(ctx, rbdVol)
+	mirrorStatus, err := rbdVol.getImageMirroringStatus()
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	err = checkHealthyPrimary(ctx, rbdVol, mirrorStatus)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
@@ -618,11 +623,7 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 // checkHealthyPrimary checks if the image is a healhty primary or not.
 // healthy primary image will be in up+stopped state, for states other
 // than this it returns an error message.
-func checkHealthyPrimary(ctx context.Context, rbdVol *rbdVolume) error {
-	mirrorStatus, err := rbdVol.getImageMirroringStatus()
-	if err != nil {
-		return err
-	}
+func checkHealthyPrimary(ctx context.Context, rbdVol *rbdVolume, mirrorStatus *librbd.GlobalMirrorImageStatus) error {
 	localStatus, err := mirrorStatus.LocalStatus()
 	if err != nil {
 		// LocalStatus can fail if the local site status is not found in
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -633,7 +633,7 @@ func checkHealthyPrimary(ctx context.Context, rbdVol *rbdVolume) error {
 		return fmt.Errorf("failed to get local status: %w", err)
 	}
 
-	if !localStatus.Up && localStatus.State != librbd.MirrorImageStatusStateStopped {
+	if !localStatus.Up || localStatus.State != librbd.MirrorImageStatusStateStopped {
 		return fmt.Errorf("%s %w. State is up=%t, state=%q",
 			rbdVol,
 			ErrUnHealthyMirroredImage,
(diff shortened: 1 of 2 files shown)


Madhu-1:
skipping CI as we dont have E2E testing for Mirroring

ShyamsundarR (dismissed):
LGTM, tested with similar changes as well...

nixpanic (changes requested):
By reverting this, won't you re-introduce the problem that the change addressed? The original commit points to https://bugzilla.redhat.com/show_bug.cgi?id=2014495.

Can you explain why this is not an issue?

I also expect to see some unit testing for different image/mirroring statuses. This code seems fragile enough that breakage is not easily spotted or detected. The change that introduced the regression was only detected months after is has been merged. Preventing breaking this code in the future should be of the highest priority.

Madhu-1:
> By reverting this, won't you re-introduce the problem that the change addressed? The original commit points to https://bugzilla.redhat.com/show_bug.cgi?id=2014495.
> 

It was not a problem, it's an enhancement we did, as this is getting reverted back I will move the BZ to the next release (downstream )

> Can you explain why this is not an issue?
>

As I mentioned above its an enhancement we tried to due but introduced a regression in a corner case. AFAIK, so revert back is fine 
 
> I also expect to see some unit testing for different image/mirroring statuses. This code seems fragile enough that breakage is not easily spotted or detected. 

Unit testing is not possible as `checkHealthyPrimary` makes a ceph call to get the mirroring image details.

> The change that introduced the regression was only detected months after is has been merged. Preventing breaking this code in the future should be of the highest priority.

not all the cases can be covered in the happy path testing, as you know testing DR is always hard we get new bugs in different areas when we started exploring different areas.

nixpanic (commented):
> > By reverting this, won't you re-introduce the problem that the change addressed? The original commit points to https://bugzilla.redhat.com/show_bug.cgi?id=2014495.
> 
> It was not a problem, it's an enhancement we did, as this is getting reverted back I will move the BZ to the next release (downstream )

Ok! It would be good to have the enhancement documented in this project too. Please create an issue for that.

> > Can you explain why this is not an issue?
> 
> As I mentioned above its an enhancement we tried to due but introduced a regression in a corner case. AFAIK, so revert back is fine

Thanks for considering, that was not clear to me by reading the PR description or commit message.

> > I also expect to see some unit testing for different image/mirroring statuses. This code seems fragile enough that breakage is not easily spotted or detected.
> 
> Unit testing is not possible as `checkHealthyPrimary` makes a ceph call to get the mirroring image details.

It is important to find a way for testing this, at least minimally. A regression was introduced because of missing tests, and there is no guarantee that a future _enhancement_ like this will not break things again.

> > The change that introduced the regression was only detected months after is has been merged. Preventing breaking this code in the future should be of the highest priority.
> 
> not all the cases can be covered in the happy path testing, as you know testing DR is always hard we get new bugs in different areas when we started exploring different areas.

Unit testing is not only about happy-path testing. Specially for features that can not be tested in e2e (yet), unit-tests are important. Can you consider re-structuring the code a bit, so that the mirror status and planned actions can be validated?

nixpanic (dismissed):
much appreciated, thanks!

Madhu-1:
> much appreciated, thanks!

Fixed CI test failure, PTAL

ShyamsundarR (approved):

nixpanic (approved):

nixpanic:
@mergifyio requeue

Madhu-1:
@mergifyio refresh

Issue #2890, opened 2022-02-18
cephfs: Cannot run systemd-run, assuming non-systemd OS


discostur:
Hi,

i'm running cephfs via ceph-csi and on some nodes i get the following errror in the ceph-csi-cephfs-nodeplugin container:

```bash
I0218 03:20:47.221224       1 utils.go:202] ID: 103 GRPC response: {"usage":[{"available":104857600,"total":104857600,"unit":1},{"total":43645,"unit":2,"used":43646}]}
I0218 03:20:49.927809       1 utils.go:191] ID: 104 GRPC call: /csi.v1.Node/NodeGetCapabilities
I0218 03:20:49.927865       1 utils.go:195] ID: 104 GRPC request: {}
I0218 03:20:49.927936       1 utils.go:202] ID: 104 GRPC response: {"capabilities":[{"Type":{"Rpc":{"type":1}}},{"Type":{"Rpc":{"type":2}}},{"Type":{"Rpc":{"type":5}}}]}
I0218 03:20:49.928764       1 utils.go:191] ID: 105 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0218 03:20:49.928809       1 utils.go:195] ID: 105 GRPC request: {"volume_id":"0001-0024-01578d80-6c97-46ba-9327-cb2b13980916-0000000000000001-8cb28e6b-2366-11eb-b0b4-0a580a2a02cd","volume_path":"/var/lib/kubelet/pods/84d81409-4c10-4763-82a4-3c1c234424bc/volumes/kubernetes.io~csi/pvc-05cdecd4-8155-4e34-8bf0-1a9f221931e3/mount"}
I0218 03:20:49.935100       1 mount_linux.go:218] Cannot run systemd-run, assuming non-systemd OS
I0218 03:20:49.935119       1 mount_linux.go:219] systemd-run output: System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to create bus connection: Host is down
, failed with: exit status 1
``` 

I get this error on some nodes - on others not. Not sure what the error is related to or if it has any impact on provisioning volumes. 

So my question is if the error is OK or if anything is going wrong here?

Thanks

# Environment details #

- Helm chart version : v3.4.0 / v3.5.1
- Kernel version : AlmaLinux 8.5 / 4.18.0-348.2.1.el8_5.x86_64 
- Mounter used for mounting PVC (for cephFS its `fuse` or `kernel`. for rbd its
[...]

PR #3225, opened 2022-06-29
logging: reduce systemd related errors while mounting

nixpanic:
There are regular reports that identify a non-error as the cause of
failures. The Kubernetes mount-utils package has detection for systemd
based environments, and if systemd is unavailable, the following error
is logged:

    Cannot run systemd-run, assuming non-systemd OS
    systemd-run output: System has not been booted with systemd as init
    system (PID 1). Can't operate.
    Failed to create bus connection: Host is down, failed with: exit status 1

Because of the `failed` and `exit status 1` error message, users might
assume that the mounting failed. This does not need to be the case. The
container-images that the Ceph-CSI projects provides, do not use
systemd, so the error will get logged with each mount attempt.

By using the newer MountSensitiveWithoutSystemd() function from the
mount-utils package where we can, the number of confusing logs get
reduced.

See-also: #2890

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -773,7 +773,7 @@ func (ns *NodeServer) mountVolumeToStagePath(
 
 	if isBlock {
 		opt = append(opt, "bind")
-		err = diskMounter.Mount(devicePath, stagingPath, fsType, opt)
+		err = diskMounter.MountSensitiveWithoutSystemd(devicePath, stagingPath, fsType, opt, nil)
 	} else {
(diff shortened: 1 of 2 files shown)


humblec (changes requested):
`mount-utils package where we can, the number of confising logs get
reduced.` -> confusing ..

humblec:
lgtm. if you can correct the commitlint failure.

humblec (approved):

yati1998 (approved):

Rakshith-R:
@mergifyio rebase

PR #3233, opened 2022-07-06
cephfs: Support mount option on nodeplugin

takmatsu:

This PR adds supporting mount option on nodeplugin.

I know currently StorageClass or Volume can have same configuration.
But sometimes several options of them are almost same. so I feel adding same options on all storage class is a bit useless.
and also StorageClass can't be changed basically though mounting should be the responsibility of nodeplugin side.
So I believe this PR should be convenient also for other users.

## my typical example option
* `recover_session=clean`

no breaking changes

## Related issues ##

None

## Future concerns ##

None

---

<details>

  progress or results

</details>


charts/ceph-csi-cephfs/values.yaml
```diff
@@ -102,6 +102,9 @@ nodeplugin:
   # on kernel < 4.17 which support quotas
   # forcecephkernelclient: true
 
+  kernelmountoptions: ""
```
nixpanic, line 107: Add a little documentation, maybe with an example?

nixpanic:
This might be useful, but it would be good to also have this for RBD so that the two have very similar features.

Can you give your practical example in this PR, explain what options you would like to configure this way?

nixpanic:
Thanks for the update! Could you squash the 2nd commit in the 1st, please?

Looks good to me otherwise :+1:

Rakshith-R (approved):

yati1998 (approved):

nixpanic (approved):

nixpanic:
@mergifyio rebase

PR #3244, opened 2022-07-13
ci: use host-networking while creating CephCluster

pkalever:

You can verify by checking your mon IP's and host IP, both should be the same.

This will be useful if we are creating multiple clusters and wanted to enable mirroring on them.

### DIFF
--- a/scripts/rook.sh
+++ b/scripts/rook.sh
@@ -64,17 +64,18 @@ function deploy_rook() {
 	kubectl_retry create -f "${TEMP_DIR}/operator.yaml"
 	# Override the ceph version which rook installs by default.
 	if [ -z "${ROOK_CEPH_CLUSTER_IMAGE}" ]; then
-		kubectl_retry create -f "${ROOK_URL}/cluster-test.yaml"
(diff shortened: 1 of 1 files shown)


scripts/rook.sh
```diff
@@ -64,17 +64,18 @@ function deploy_rook() {
 	kubectl_retry create -f "${TEMP_DIR}/operator.yaml"
 	# Override the ceph version which rook installs by default.
 	if [ -z "${ROOK_CEPH_CLUSTER_IMAGE}" ]; then
-		kubectl_retry create -f "${ROOK_URL}/cluster-test.yaml"
+		curl -o "${TEMP_DIR}"/cluster-test.yaml "${ROOK_URL}/cluster-test.yaml"
+		sed -i "s/healthCheck:/network:\n    provider: host\n  healthCheck:/g" "${TEMP_DIR}"/cluster-test.yaml
```
Madhu-1, line 68: not  always useful as extra firewall configurations need to be done if someone is installing minikube on host machines.

Madhu-1: good to have this functionality behind a flag to control host networking.

Issue #3246, opened 2022-07-18
Reduce messages related to systemd in the logs


nixpanic:
# Describe the feature you'd like to have #

kubernetes/kubernetes#111083 already reduced the systemd related logging a lot. CephFS does not have any systemd logging at all anymore.\

RBD calls the `k8s.io/mount-utils` function `FormatAndMount()`, which still causes some systemd log messages. With the enhancement in kubernetes/kubernetes#111218, a `NewWithoutSystemd()` function has been proposed, and Ceph-CSI could benefit from that.

# What is the value to the end user? (why is it a priority?) #

No more confusing systemd log messages, see #2890 for an example of a user that reports an issue about it.

# How will we know we have a good solution? (acceptance criteria) #

No more systemd log messages. Ceph-CSI does not want to use systemd in any case.

# Additional context #

Example of the changes that are required once kubernetes/kubernetes#111218 has been merged+released:

```patch
diff --git a/internal/rbd/driver/driver.go b/internal/rbd/driver/driver.go
index 5019d02cc..2fed92be0 100644
--- a/internal/rbd/driver/driver.go
+++ b/internal/rbd/driver/driver.go
@@ -73,7 +73,7 @@ func NewReplicationServer(c *rbd.ControllerServer) *rbd.ReplicationServer {
 
 // NewNodeServer initialize a node server for rbd CSI driver.
 func NewNodeServer(d *csicommon.CSIDriver, t string, topology map[string]string) (*rbd.NodeServer, error) {
-       mounter := mount.New("")
+       mounter := mount.NewWithoutSystemd("")
 
        return &rbd.NodeServer{
                DefaultNodeServer: csicommon.NewDefaultNodeServer(d, t, topology),
diff --git a/internal/util/util.go b/internal/util/util.go
index 4c36e3679..3207fa148 100644
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -321,7 +321,7 @@ func checkDirExists(p string) bool {
 
 // IsMountPoint checks if the given path is mountpoint or not.
 func IsMountPoint(p string) (bool, error) {
-       dummyMount := mount.New("")
+       dummyMount := mount.NewWithoutSystemd("")
        notMnt, err := dummyMount.IsLikelyNotMountPoint(p)
        if err != nil {
                return false, err
@@ -344,7 +344,7 @@ func ReadMountInfoForProc(proc string) ([]mount.MountInfo, error) {
 
 // Mount mounts the source to target path.
[...]

PR #3247, opened 2022-07-18
cleanup: create k8s.io/mount-utils Mounter only once

nixpanic:
Recently the k8s.io/mount-utils package added more runtime dectection.
When creating a new Mounter, the detect is run every time. This is
unfortunate, as it logs a message like the following:

```
mount_linux.go:283] Detected umount with safe 'not mounted' behavior
```

This message might be useful, so it probably good to keep it.

In Ceph-CSI there are various locations where Mounter instances are
created. Moving that to the DefaultNodeServer type reduces it to a
single place. Some utility functions need to accept the additional
parameter too, so that has been modified as well.

See-also: kubernetes/kubernetes#109676
Updates: #3246

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/rbd/driver/driver.go
+++ b/internal/rbd/driver/driver.go
@@ -29,7 +29,6 @@ import (
 	"github.com/ceph/ceph-csi/internal/util/log"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
-	mount "k8s.io/mount-utils"
 )
 
 // Driver contains the default identity,node and controller struct.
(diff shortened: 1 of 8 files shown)


internal/rbd/driver/driver.go
```diff
@@ -73,11 +72,8 @@ func NewReplicationServer(c *rbd.ControllerServer) *rbd.ReplicationServer {
 
 // NewNodeServer initialize a node server for rbd CSI driver.
 func NewNodeServer(d *csicommon.CSIDriver, t string, topology map[string]string) (*rbd.NodeServer, error) {
-	mounter := mount.New("")
-
 	return &rbd.NodeServer{
 		DefaultNodeServer: csicommon.NewDefaultNodeServer(d, t, topology),
-		Mounter:           mounter,
```
Madhu-1, line 80: we need this change for rbd right? or else it will panic when we try to access mounter

nixpanic: no, that should be set by the DefaultNodeServer... not sure why this isn't working as I expected :thinking:

Having a look now.

nixpanic: Dropping the overloading `Mounter` attribute from `rbd.NodeServer` should work :crossed_fingers:

Rakshith-R:
@mergifyio rebase

Rakshith-R:
@mergifyio rebase

Rakshith-R (approved):

Madhu-1 (approved):

PR #3265, opened 2022-07-26
deploy: enable HonorPVReclaimPolicy feature gate

### LINKED ISSUE #3230: Enable HonorPVReclaimPolicy feature gate
Enable the HonorPVReclaimPolicy feature gate supported by latest version of csi-provisioner sidecar.
https://github.com/kubernetes-csi/external-provisioner/releases/tag/v3.2.0

Currently, it is disabled due to nfs error handling issue, once it is handled in nfs, we can enable the feature gate.
Check out the comments for more details.: https://github.com/ceph/ceph-csi/pull/3186#issuecomment-1174679852

yati1998:
This commit enables the HonorPVReclaimPolicy feature
gate.

fixes: https://github.com/ceph/ceph-csi/issues/3230

Signed-off-by: Yati Padia [ypadia@redhat.com](mailto:ypadia@redhat.com)

### DIFF
--- a/deploy/nfs/kubernetes/csi-nfsplugin-provisioner.yaml
+++ b/deploy/nfs/kubernetes/csi-nfsplugin-provisioner.yaml
@@ -47,9 +47,7 @@ spec:
             - "--timeout=150s"
             - "--leader-election=true"
             - "--retry-interval-start=500ms"
-            # TODO: enable the feature-gate once error has been handled in nfs.
-            # Issue #3230
-            # - "--feature-gates=HonorPVReclaimPolicy=true"
+            - "--feature-gates=HonorPVReclaimPolicy=true"
             - "--prevent-volume-mode-conversion=true"
           env:
             - name: ADDRESS
--- a/e2e/nfs.go
+++ b/e2e/nfs.go
@@ -373,7 +373,7 @@ var _ = Describe("nfs", func() {
 			By("verify RWOP volume support", func() {
 				err := createNFSStorageClass(f.ClientSet, f, false, nil)
 				if err != nil {
-					e2elog.Failf("failed to create CephFS storageclass: %v", err)
+					e2elog.Failf("failed to create NFS storageclass: %v", err)
 				}
 				pvc, err := loadPVC(pvcRWOPPath)
 				if err != nil {
@@ -411,40 +411,40 @@ var _ = Describe("nfs", func() {
 				validateSubvolumeCount(f, 0, fileSystemName, defaultSubvolumegroup)
 				err = deleteResource(nfsExamplePath + "storageclass.yaml")
 				if err != nil {
-					e2elog.Failf("failed to delete CephFS storageclass: %v", err)
+					e2elog.Failf("failed to delete NFS storageclass: %v", err)
 				}
 			})
 
 			By("create a storageclass with pool and a PVC then bind it to an app", func() {
 				err := createNFSStorageClass(f.ClientSet, f, false, nil)
 				if err != nil {
-					e2elog.Failf("failed to create CephFS storageclass: %v", err)
+					e2elog.Failf("failed to create NFS storageclass: %v", err)
 				}
 				err = validatePVCAndAppBinding(pvcPath, appPath, f)
 				if err != nil {
-					e2elog.Failf("failed to validate CephFS pvc and application binding: %v", err)
+					e2elog.Failf("failed to validate NFS pvc and application binding: %v", err)
 				}
 				err = deleteResource(nfsExamplePath + "storageclass.yaml")
 				if err != nil {
-					e2elog.Failf("failed to delete CephFS storageclass: %v", err)
+					e2elog.Failf("failed to delete NFS storageclass: %v", err)
 				}
 			})
 
 			By("create a PVC and bind it to an app", func() {
 				err := createNFSStorageClass(f.ClientSet, f, false, nil)
 				if err != nil {
-					e2elog.Failf("failed to create CephFS storageclass: %v", err)
+					e2elog.Failf("failed to create NFS storageclass: %v", err)
 				}
 				err = validatePVCAndAppBinding(pvcPath, appPath, f)
 				if err != nil {
-					e2elog.Failf("failed to validate CephFS pvc and application  binding: %v", err)
+					e2elog.Failf("failed to validate NFS pvc and application  binding: %v", err)
 				}
 			})
 
 			By("create a PVC and bind it to an app with normal user", func() {
 				err := validateNormalUserPVCAccess(pvcPath, f)
 				if err != nil {
-					e2elog.Failf("failed to validate normal user CephFS pvc and application binding: %v", err)
+					e2elog.Failf("failed to validate normal user NFS pvc and application binding: %v", err)
 				}
 			})
 
--- a/internal/nfs/controller/errors.go
+++ b/internal/nfs/controller/errors.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2022 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrNotConnected is returned when components from the NFS
+	// ControllerServer can not connect to the Ceph cluster or NFS-Ganesha
+	// service.
+	ErrNotConnected = errors.New("not connected")
+
+	// ErrNotFound is a generic error that is the parent of other "not
+	// found" failures. Callers can check if something was "not found",
+	// even if the actual error is more specific.
+	ErrNotFound = errors.New("not found")
+
+	// ErrExportNotFound is returned by components that communicate with the
+	// NFS-Ganesha service, and have identified that the NFS-Export does
+	// not exist (anymore). This error is also a ErrNotFound.
+	ErrExportNotFound = fmt.Errorf("NFS-export %w", ErrNotFound)
+
+	// ErrFilesystemNotFound is returned in case the filesystem
+	// does not exist.
+	ErrFilesystemNotFound = fmt.Errorf("filesystem %w", ErrNotFound)
+)
--- a/internal/nfs/controller/volume.go
+++ b/internal/nfs/controller/volume.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -124,7 +125,7 @@ func (nv *NFSVolume) GetExportPath() string {
 // a new NFS-export for the volume on the Ceph managed NFS-server.
 func (nv *NFSVolume) CreateExport(backend *csi.Volume) error {
 	if !nv.connected {
-		return fmt.Errorf("can not created export for %q: not connected", nv)
+		return fmt.Errorf("can not created export for %q: %w", nv, ErrNotConnected)
 	}
 
 	fs := backend.VolumeContext["fsName"]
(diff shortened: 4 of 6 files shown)


deploy/nfs/kubernetes/csi-nfsplugin-provisioner.yaml
```diff
@@ -49,7 +49,7 @@ spec:
             - "--retry-interval-start=500ms"
             # TODO: enable the feature-gate once error has been handled in nfs.
             # Issue #3230
-            # - "--feature-gates=HonorPVReclaimPolicy=true"
+            - "--feature-gates=HonorPVReclaimPolicy=true"
```
nixpanic, line 50: the comment above can be removed now

e2e/nfs.go
```diff
@@ -371,80 +371,82 @@ var _ = Describe("nfs", func() {
...
-				err := createNFSStorageClass(f.ClientSet, f, false, nil)
-				if err != nil {
-					e2elog.Failf("failed to create CephFS storageclass: %v", err)
-				}
-				pvc, err := loadPVC(pvcRWOPPath)
-				if err != nil {
-					e2elog.Failf("failed to load PVC: %v", err)
-				}
-				pvc.Namespace = f.UniqueName
+				if k8sVersionGreaterEquals(f.ClientSet, 1, 22) {
```
nixpanic: we're not testing with 1.21 anymore, this if-check can be removed so that the test is always enabled

internal/nfs/controller/errors.go
```diff
@@ -0,0 +1,43 @@
...
+	ErrNotFound = errors.New("not found")
+
+	// ErrExportNotFound is returned by components that communicate with the
+	// NFS-Ganesha service, and have identified that the NFS-Export does
+	// not exist (anymore). This error is also a ErrNotFound.
+	ErrExportNotFound = fmt.Errorf("NFS-export %w", ErrNotFound)
+
+	// ErrFilesystemNotFound is returned in case the filesystem
+	// does not exist.
+	ErrFilesystemNotFound = fmt.Errorf("filesystem not found")
```
nixpanic: use the same format as `ErrExportNotFound`, include `%w` and `ErrNotFound`. That makes it possible for callers to check only for `ErrNotFound` and not all `Err...NotFound` variations
```go
err := ErrFilesystemNotFound
if err != nil && !errors.Is(err, ErrNotFound) {
    ...
```

internal/nfs/controller/volume.go
```diff
@@ -251,18 +254,21 @@ func (nv *NFSVolume) deleteExportCommand(cmd, nfsCluster string) []string {
...
-		return "", fmt.Errorf("can not get NFS-cluster for %q: not connected", nv)
+		return "", fmt.Errorf("can not get NFS-cluster for %q: %w", nv, ErrNotConnected)
 	}
 
 	fs := fscore.NewFileSystem(nv.conn)
 	fsName, err := fs.GetFsName(nv.ctx, nv.fscID)
 	if err != nil {
-		return "", fmt.Errorf("failed to get filesystem name for ID %x: %w", nv.fscID, err)
+		if errors.Is(err, util.ErrPoolNotFound) {
+			return "", fmt.Errorf("filesystem not found %w: %v", ErrFilesystemNotFound, err)
```
nixpanic: `ErrFilesystemNotFound` contains the text "filesystem not found" already, maybe use
```go
fmt.Errorf("%w for ID %x: %v", ErrFilesystemNotFound, nv.fscID, err)
```

nixpanic (changes requested):
just a few nits to make it all a little cleaner

nixpanic (dismissed):

internal/nfs/controller/volume.go
```diff
@@ -251,18 +254,21 @@ func (nv *NFSVolume) deleteExportCommand(cmd, nfsCluster string) []string {
...
 	}
 
 	fs := fscore.NewFileSystem(nv.conn)
 	fsName, err := fs.GetFsName(nv.ctx, nv.fscID)
 	if err != nil {
-		return "", fmt.Errorf("failed to get filesystem name for ID %x: %w", nv.fscID, err)
+		if errors.Is(err, util.ErrPoolNotFound) {
+			return "", fmt.Errorf("%w for ID %x: %v", ErrFilesystemNotFound, nv.fscID, err)
+		}
+		return "", fmt.Errorf("filesystem name for ID %x %w: %v", nv.fscID, ErrNotFound, err)
```
nixpanic: golangci-lint requires an empty line before this last `return`

nixpanic (dismissed):

internal/nfs/controller/volume.go
```diff
@@ -274,7 +281,7 @@ func (nv *NFSVolume) getNFSCluster() (string, error) {
 
 	clusterName, err := j.FetchAttribute(nv.ctx, mdPool, nv.objectUUID, clusterNameKey)
 	if err != nil {
-		return "", fmt.Errorf("failed to get cluster name: %w", err)
+		return "", fmt.Errorf("cluster name for %q %w: %v", nv.objectUUID, ErrNotFound, err)
```
Madhu-1: This wont be true in all causes we should return `ErrNotFound` only if `FetchAttribute` is returning `ErrKeyNotFound` or `ErrPoolNotFound`

internal/nfs/controller/volume.go
```diff
@@ -283,18 +290,18 @@ func (nv *NFSVolume) getNFSCluster() (string, error) {
...
 	if !nv.connected {
-		return fmt.Errorf("can not set NFS-cluster for %q: not connected", nv)
+		return fmt.Errorf("can not set NFS-cluster for %q: %w", nv, ErrNotConnected)
 	}
 
 	fs := fscore.NewFileSystem(nv.conn)
 	fsName, err := fs.GetFsName(nv.ctx, nv.fscID)
 	if err != nil {
-		return fmt.Errorf("failed to get filesystem name for ID %x: %w", nv.fscID, err)
+		return fmt.Errorf("filesystem name for ID %x %w: %v", nv.fscID, ErrNotFound, err)
```
Madhu-1: same here we need below check isnt it

```
if errors.Is(err, util.ErrPoolNotFound) {
			return "", fmt.Errorf("%w for ID %x: %v", ErrFilesystemNotFound, nv.fscID, err)
		}

		return "", fmt.Errorf("filesystem name for ID %x %w: %v", nv.fscID, ErrNotFound, err)
```

internal/nfs/controller/volume.go
```diff
@@ -251,18 +254,22 @@ func (nv *NFSVolume) deleteExportCommand(cmd, nfsCluster string) []string {
...
+			return "", fmt.Errorf("%w for ID %x: %v", ErrFilesystemNotFound, nv.fscID, err)
+		}
+
+		return "", fmt.Errorf("filesystem name for ID %x %w: %v", nv.fscID, ErrNotFound, err)
 	}
 
 	mdPool, err := fs.GetMetadataPool(nv.ctx, fsName)
 	if err != nil {
-		return "", fmt.Errorf("failed to get metadata pool for %q: %w", fsName, err)
+		return "", fmt.Errorf("metadata pool for %q %w: %v", fsName, ErrNotFound, err)
```
Madhu-1, line 270: Do we need to return `ErrNotFound` for all the error cases or return `ErrNotFound` only if `ErrPoolNotFound` error is returned?

internal/nfs/controller/volume.go
```diff
@@ -283,18 +290,18 @@ func (nv *NFSVolume) getNFSCluster() (string, error) {
...
 	fsName, err := fs.GetFsName(nv.ctx, nv.fscID)
 	if err != nil {
-		return fmt.Errorf("failed to get filesystem name for ID %x: %w", nv.fscID, err)
+		return fmt.Errorf("filesystem name for ID %x %w: %v", nv.fscID, ErrNotFound, err)
 	}
 
 	mdPool, err := fs.GetMetadataPool(nv.ctx, fsName)
 	if err != nil {
-		return fmt.Errorf("failed to get metadata pool for %q: %w", fsName, err)
+		return fmt.Errorf("metadata pool for %q %w: %v", fsName, ErrNotFound, err)
```
Madhu-1, line 308: same here for error handling we should return `ErrNotFound` for pool not found error not for all the error cases isn't it?

Madhu-1 (changes requested):
need some clarification on error handling. requesting changes to avoid accidental merge.

internal/nfs/controller/volume.go
yati1998: yeah missed this. will update

internal/nfs/controller/volume.go
yati1998: As far as I understood, for now, we are returning this for all cases, and in case we don't find the pool or in the above case the filesystem we can be specific about it. @nixpanic please correct me if I am wrong.

internal/nfs/controller/volume.go
```diff
@@ -251,18 +254,22 @@ func (nv *NFSVolume) deleteExportCommand(cmd, nfsCluster string) []string {
...
 
 	fs := fscore.NewFileSystem(nv.conn)
 	fsName, err := fs.GetFsName(nv.ctx, nv.fscID)
 	if err != nil {
-		return "", fmt.Errorf("failed to get filesystem name for ID %x: %w", nv.fscID, err)
+		if errors.Is(err, util.ErrPoolNotFound) {
+			return "", fmt.Errorf("%w for ID %x: %v", ErrFilesystemNotFound, nv.fscID, err)
+		}
+
+		return "", fmt.Errorf("filesystem name for ID %x %w: %v", nv.fscID, ErrNotFound, err)
```
nixpanic: note that `ErrNotFound` will cause the caller to assume that the NFS-volume does not exist (anymore). If there is some other error (permissions?), it may not be correct. It is probably better to not include `ErrNotFound`, in the non-specific error return.

internal/nfs/controller/volume.go
nixpanic: I do not have a strong preference for one approach or another. It is true that an error fetching the metadata pool now will assume that the NFS-volume does not exist. Even in the case when the error is different from `ErrPoolNotFound`.

But, I also agree that it is cleaner to do the same checking as done for GetFsName() on line 261. In case the Pool does can not be found, return a `ErrNotFound` and otherwise the `err` that was returned from the call (with additional `fmt.Errorf(..)` context).

The same theory applies to the error checking further below.

nixpanic (changes requested):

yati1998:
@nixpanic the changes look good to me now, Thanks for the changes. @Madhu-1 would like to get your review.

internal/nfs/controller/volume.go
Madhu-1: This is still not addressed yet?

nixpanic (dismissed):

internal/nfs/controller/volume.go
nixpanic: hmm, I thought it was? it seems I missed this one

nixpanic: It should be addressed now too.

nixpanic (approved):

internal/nfs/controller/volume.go
```diff
@@ -273,8 +280,10 @@ func (nv *NFSVolume) getNFSCluster() (string, error) {
 	defer j.Destroy()
 
 	clusterName, err := j.FetchAttribute(nv.ctx, mdPool, nv.objectUUID, clusterNameKey)
-	if err != nil {
-		return "", fmt.Errorf("failed to get cluster name: %w", err)
+	if err != nil && errors.Is(err, util.ErrPoolNotFound) || errors.Is(err, util.ErrKeyNotFound) {
```
Rakshith-R, line 283:
```suggestion
	if err != nil && (errors.Is(err, util.ErrPoolNotFound) || errors.Is(err, util.ErrKeyNotFound)) {
```

A small nit,
but this should not matter, going by this example
I guess we can strip `err !=nil &&` check completely.
https://go.dev/play/p/e97541eA7HO

Rakshith-R (approved):

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

nixpanic:
> /retest ci/centos/k8s-e2e-external-storage/1.23

Mininkube failed to start the VM

nixpanic:
> /retest ci/centos/k8s-e2e-external-storage/1.23
> /retest ci/centos/mini-e2e-helm/k8s-1.23

Both  were stuck while requesting a system for testing.

nixpanic:
@mergifyio rebase

nixpanic:

Failed to install packages in the VM

PR #3269, opened 2022-07-27
rbd: remove checkHealthyPrimary check

### LINKED ISSUE #3268: RBD DR: Images are not marked as healthy after forced failover
After Failover of workloads to the secondary cluster when the primary cluster is down, RBD Image Is not marked healthy, and VR resources are not promoted to the Primary, IN VolumeReplication, the `CURRENT STATE` remains Unknown and doesn't change to Primary.

This happens because the primary cluster went down, and we have force promoted the image on the secondary cluster. and the image stays in up+stopping_replay or could be any other states as in https://github.com/ceph/go-ceph/blob/f5d1adf26ffafc2a3d3fc6cbc802aa278f684ad9/rbd/mirror.go#L327-L335. Currently assumption was that the image will always be `up+stopped` . But the image will be in `up+stopped` only for planned failover and it could be in any other state if its a forced failover. For this reason, we need to remove the check for https://github.com/ceph/ceph-csi/blob/33d4d54dbe91b2498d2b5ed93383d62acda4ae3b/internal/rbd/replicationcontrollerserver.go#L623-L647 from the PromoteVolume RPC call.

Madhu-1:
After Failover of workloads to the secondary cluster when the primary cluster is down,
RBD Image is not marked healthy, and VR resources are not promoted to the Primary; in VolumeReplication, the `CURRENT STATE` remains Unknown and doesn't change to Primary.

This happens because the primary cluster went down, and we have force promoted the image on the
secondary cluster. and the image stays in up+stopping_replay or could be any other states. The assumption was that the image would always be `up+stopped`. But the image will be in `up+stopped` only for a planned failover, and it could be in any other state if it's a forced failover. For this reason, removing checkHealthyPrimary from the PromoteVolume RPC call.

fixes #3268 


idryomov (approved):
Ack from the RBD perspective

nixpanic (approved):

nixpanic:
@Madhu-1 as this is a bugfix, backport it to release-3.6?

Madhu-1:
> @Madhu-1 as this is a bugfix, backport it to release-3.6?

Need to backport it to 3.6. adding Label. Thanks @nixpanic

Rakshith-R (approved):

pkalever (approved):
Looks reasonable to me.

PR #3291, opened 2022-08-08
rbd: Report remote peer readiness if Up and status.Unknown

ShyamsundarR:
Current code uses an !A && !B condition incorrectly to
test A:Up and B:status for a remote peer image.

This should be !A || !B as we require both conditions to
be in the specified state (Up: true, and status Unknown).

This is corrected by this commit, and further fixes:
- check and return ready only when a remote site is
found in the status output
- check if all peer sites are ready, if multiple are found
and return ready appropriately

Signed-off-by: Shyamsundar Ranganathan <srangana@redhat.com>

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -680,6 +680,7 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 // It returns true if the state of the remote cluster is up and unknown.
 func checkRemoteSiteStatus(ctx context.Context, mirrorStatus *librbd.GlobalMirrorImageStatus) bool {
 	ready := true
+	found := false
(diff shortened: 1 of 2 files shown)


Madhu-1 (dismissed):

Madhu-1:
>internal/rbd/replicationcontrollerserver_test.go:310: File is not `gofmt`-ed with `-s` (gofmt)

@ShyamsundarR lint failure

ShyamsundarR:
Executed e2e disaster recovery testing, including ramen orchestration and this code change across deploy->failover->relocate->delete of an application with multiple PVs and also when one of the clusters is down. The change does not impact the regular workflow.

Adding this note here for confidence level on the patch during reviews.

Madhu-1 (approved):

Rakshith-R (approved):

PR #3346, opened 2022-08-26
RFC: Use minikube prerelease ISO on CI

irq0:
Now that minikube fscrypt kernel support is merged (https://github.com/kubernetes/minikube/pull/14783), I'm looking for a way to run https://github.com/ceph/ceph-csi/pull/3310 on the Ceph CSI CI with that ISO. A updated ISO was build and is available for download: https://storage.googleapis.com/minikube-builds/iso/14783/minikube-v1.26.1-1661377864-14783-amd64.iso

 Would this be a way forward until the next minikube release?

### DIFF
--- a/build.env
+++ b/build.env
@@ -39,6 +39,7 @@ HELM_VERSION=v3.9.2
 
 # minikube settings
 MINIKUBE_VERSION=v1.26.1
+MINIKUBE_ISO_URL=https://storage.googleapis.com/minikube-builds/iso/14783/minikube-v1.26.1-1661377864-14783-amd64.iso
 VM_DRIVER=none
 CHANGE_MINIKUBE_NONE_USER=true
 
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -150,6 +150,7 @@ function disable_storage_addons() {
 # configure minikube
 MINIKUBE_ARCH=${MINIKUBE_ARCH:-"amd64"}
 MINIKUBE_VERSION=${MINIKUBE_VERSION:-"latest"}
+MINIKUBE_ISO_URL=${MINIKUBE_ISO_URL:-""}
 KUBE_VERSION=${KUBE_VERSION:-"latest"}
 CONTAINER_CMD=${CONTAINER_CMD:-"docker"}
 MEMORY=${MEMORY:-"4096"}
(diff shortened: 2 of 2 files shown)


Rakshith-R:
@irq0 
Add these commits as part of #3310, to check if ci for new encryption changes passes ?

nixpanic (approved):
I like the feature, and am sure @pkalever appreciates it too. Going to default for a pre-release for everything is not that great though... At the very least, we'll need an issue to track removing the default pre-release URL once the next minikube version is available.

We use pre-releases sometimes, so I don't think it's a blocking issue.

build.env
```diff
@@ -39,6 +39,7 @@ HELM_VERSION=v3.9.2
 
 # minikube settings
 MINIKUBE_VERSION=v1.26.1
+MINIKUBE_ISO_URL=https://storage.googleapis.com/minikube-builds/iso/14783/minikube-v1.26.1-1661377864-14783-amd64.iso
```
Madhu-1, line 42: If this doesn't get updated and CI works fine, am good with this one.

Madhu-1:
> Now that minikube fscrypt kernel support is merged ([kubernetes/minikube#14783](https://github.com/kubernetes/minikube/pull/14783)), I'm looking for a way to run #3310 on the Ceph CSI CI with that ISO. A updated ISO was build and is available for download: https://storage.googleapis.com/minikube-builds/iso/14783/minikube-v1.26.1-1661377864-14783-amd64.iso
> 
> Would this be a way forward until the next minikube release?

The best option is to add it as a new commit in the PR which needs this support.

build.env
pkalever: I think @Madhu-1 concern is vaild, can we copy this to some local file server @nixpanic (I remember we supplied a native link with NBD patches before) ?

nixpanic: We don't really want to host ISO images ourselves. If this link stays available until a next minikube version contains the fscrypt feature, I'd prefer to using this.

irq0:
> @irq0 Add these commits as part of #3310, to check if ci for new encryption changes passes ?

Tried, but it seems that single-node-k8s.sh from the ci branch doesn't forward the new env var. https://github.com/ceph/ceph-csi/pull/3351 should handle that

irq0:
> > @irq0 Add these commits as part of #3310, to check if ci for new encryption changes passes ?
> 
> Tried, but it seems that single-node-k8s.sh from the ci branch doesn't forward the new env var. #3351 should handle that

Thanks for merging #3351. I added the two commits to the last push to #3310 and the CI system picked up the new options. The ISO is listed in the minikube logs output of the 'log system status' jenkins node. I'm closing this one in favor #3310.

PR #3361, opened 2022-09-01
rbd: change default FsGroupPolicy to "File" for RBD CSI driver

humblec:
This commit change the default fsgroup policy for csi driver object
to "File" type which is the better/correct setting for the CSI volumes
as it works regardless of fstype or access mode

We have been using default value which is "ReadWriteOnceWithFSType"

With this change backward compatibility should be preserved.

### DIFF
--- a/charts/ceph-csi-rbd/templates/csidriver-crd.yaml
+++ b/charts/ceph-csi-rbd/templates/csidriver-crd.yaml
@@ -9,3 +9,4 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: false
+  fsGroupPolicy: File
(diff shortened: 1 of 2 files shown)


Madhu-1:
> This commit change the default fsgroup policy for csi driver object to "File" type which is the better/correct setting for the CSI volumes. We have been using default value ie "ReadWriteOnceWithFSType". with this change backward compatibility should be preserved.
> 

@humblec please provide some more details why this is better or correct? it helps in reviewing and merging this one.

humblec:
> > This commit change the default fsgroup policy for csi driver object to "File" type which is the better/correct setting for the CSI volumes. We have been using default value ie "ReadWriteOnceWithFSType". with this change backward compatibility should be preserved.
> 
> @humblec please provide some more details why this is better or correct? it helps in reviewing and merging this one.

@Madhu-1 File works regardless of fstype or access mode and the better setting for NAS type.

Madhu-1 (approved):

Rakshith-R (approved):

PR #3363, opened 2022-09-05
rbd: map only primary image

Madhu-1:
If the image is mirroring enabled and primary consider it for mapping. If the image is mirroring enabled but not primary yet. return error message until the image is marked as primary.

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -138,8 +138,6 @@ type rbdImage struct {
 	// an opened IOContext, call .openIoctx() before using
 	ioctx *rados.IOContext
 
-	// Primary represent if the image is primary or not.
-	Primary bool
 	// Set metadata on volume
 	EnableMetadata bool
 }
@@ -512,6 +510,10 @@ func (ri *rbdImage) open() (*librbd.Image, error) {
 
 // isInUse checks if there is a watcher on the image. It returns true if there
 // is a watcher on the image, otherwise returns false.
+// In case of mirroring, the image should be primary to check watchers if the
+// image is secondary it returns an error.
+// isInUse is called with exponential backoff to check the image is used by
+// anyone else the returned bool value is discarded if its a RWX access.
 func (ri *rbdImage) isInUse() (bool, error) {
 	image, err := ri.open()
 	if err != nil {
@@ -532,11 +534,16 @@ func (ri *rbdImage) isInUse() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	ri.Primary = mirrorInfo.Primary
+
+	if mirrorInfo.State == librbd.MirrorImageEnabled && !mirrorInfo.Primary {
+		// Mapping secondary image can cause issues.returning error as the
+		// bool value is discarded if it its RWX access.
+		return false, fmt.Errorf("cannot map image %s it is not primary", ri)
+	}
 
 	// because we opened the image, there is at least one watcher
 	defaultWatchers := 1
-	if ri.Primary {
+	if mirrorInfo.Primary {
 		// if rbd mirror daemon is running, a watcher will be added by the rbd
 		// mirror daemon for mirrored images.
 		defaultWatchers++


Madhu-1:
@idryomov can you please take a look at this one?

internal/rbd/rbd_util.go
```diff
@@ -532,7 +532,13 @@ func (ri *rbdImage) isInUse() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	ri.Primary = mirrorInfo.Primary
+
+	if mirrorInfo.State == librbd.MirrorImageEnabled {
+		if !ri.Primary {
+			return false, fmt.Errorf("image %s is not primary", ri)
```
nixpanic: Should this not return `true`?

It would be good to update the comment of the function too. It explains only about watchers, and a note about mirroring (requiring it to be the primary) would be good.

Madhu-1: This function returns false for all the error cases.

nixpanic: Not being primary is not an error. It is a valid state. Just like there could be more watchers than expected. The image is just "in use", and should not be mapped on the node.

nixpanic: If you prefer to report details about why the image is "in use", consider cleaning up this function by only returning an error, and not an additional bool.

Madhu-1: >If you prefer to report details about why the image is "in use", consider cleaning up this function by only returning an error, and not an additional bool.

This requires more changes as we try to map with exponential backoff. altered function to log error if the image is not primary and return true,nil to try in exponential backoff.

nixpanic: ah, ok! in that case, document the behaviour in the description of the function too :-)

nixpanic (dismissed):

internal/rbd/rbd_util.go
```diff
@@ -532,7 +534,15 @@ func (ri *rbdImage) isInUse() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	ri.Primary = mirrorInfo.Primary
+
+	if mirrorInfo.State == librbd.MirrorImageEnabled {
+		if !ri.Primary {
```
idryomov: Shouldn't this be `!mirrorInfo.Primary`?

internal/rbd/rbd_util.go
```diff
@@ -532,7 +534,15 @@ func (ri *rbdImage) isInUse() (bool, error) {
...
+			log.ErrorLog(ctx, "cannot map image %s it is not primary", ri)
+
+			return true, nil
+		}
+		ri.Primary = mirrorInfo.Primary
```
idryomov: Can `isInUse()` be called with `ri.Primary` set to `true`?  Previously, `ri.Primary` was reset unconditionally, with this change it's no longer the case.

idryomov (changes requested):

internal/rbd/rbd_util.go
Madhu-1: Now I remember why I made this change, The bool value (watchers) are/is discarded if the request is for RWX volume. Mapping secondary can cause issues that's why we are introducing this check, reverting this change PTAL

internal/rbd/rbd_util.go
Madhu-1: :facepalm: yes you are correct. am missed it somehow

internal/rbd/rbd_util.go
Madhu-1: Removed ri.Primary as it's not used anywhere else except this function.

internal/rbd/rbd_util.go
```diff
@@ -532,11 +534,18 @@ func (ri *rbdImage) isInUse() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	ri.Primary = mirrorInfo.Primary
+
+	if mirrorInfo.State == librbd.MirrorImageEnabled {
+		if !mirrorInfo.Primary {
```
idryomov: These could be combined into a single check for better readability and less indentation.

nixpanic (approved):
Looks much cleaner and easier to understand now. Thanks for adding the additional comments too!

idryomov (approved):
This looks fine to me in isolation -- the error message is improved -- but I don't think this change is a proper fix for the issue of the DR stack submitting `NodeStageVolume` requests concurrently with `PromoteVolume` requests.

Madhu-1:
> This looks fine to me in isolation -- the error message is improved -- but I don't think this change is a proper fix for the issue of the DR stack submitting `NodeStageVolume` requests concurrently with `PromoteVolume` requests.

Yes, this is not a fix for it completely. this is to ensure that we don't try to map the rbd image, which is not primary yet. the DR Orchestration should take care of it

Rakshith-R (approved):

nixpanic:
This can be rebased with Mergify once #3353 is merged.

Madhu-1:
@mergifyio rebase

PR #3377, opened 2022-09-07
rbd: create token and use it for vault SA everytime possible

### LINKED ISSUE #3360: Supporting TokenRequest API for vault tenant SA & K8s versions < 1.24

Currently, ceph-csi supports using the TokenRequest API in vault tenant SA approach from K8s 1.24+. However, for K8s version older than 1.24 we rely on secret references obtained from SA, even though the TokenRequest API is supported. 

Proposal would be to use TokenRequest API by default and fall back to the secret reference approach if we aren’t able to create the token for any reason. Related conversation on PR that introduced create token approach for K8s 1.24 https://github.com/ceph/ceph-csi/pull/3174/files#r898205177

Better security as ceph-csi would rely on tokens that have a bounded lifetime. Quoting K8s document for reference:
https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets 

> Since 1.22, this type of Secret is no longer used to mount credentials into Pods, and obtaining tokens via the TokenRequest API is recommended instead of using service account token Secret objects. Tokens obtained from the TokenRequest API are more secure than ones stored in Secret objects, because they have a bounded lifetime and are not readable by other API clients. You can use the kubectl create token command to obtain a token from the TokenRequest API.
>
> You should only create a service account token Secret object if you can't use the TokenRequest API to obtain a token, and the security exposure of persisting a non-expiring token credential in a readable API object is acceptable to you.

saiprashanth173:
Signed-off-by: Prashanth Dintyala <vdintyala@nvidia.com>

Create short lived service account tokens and use them for vault SA approach even for K8s version <1.24; fall back to using token obtained from services account's secret reference (if exists). 
  
## Related issues ##
Fixes: #3360 

---

<details>

  progress or results

</details>

### DIFF
--- a/internal/kms/vault_sa.go
+++ b/internal/kms/vault_sa.go
@@ -23,8 +23,6 @@ import (
 	"io/ioutil"
 	"os"
 
-	"github.com/ceph/ceph-csi/internal/util/k8s"
-
 	"github.com/libopenstorage/secrets/vault"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -39,9 +37,6 @@ const (
 	// should be available in the Tenants namespace. This ServiceAccount
 	// will be used to connect to Hashicorp Vault.
 	vaultTenantSAName = "ceph-csi-vault-sa"
-	// Kubernetes version which requires ServiceAccount token creation.
-	// Kubernetes 1.24 => 1 * 1000 + 24.
-	kubeMinVersionForCreateToken = 1024
 )
 
 /*
@@ -298,6 +293,15 @@ func (kms *vaultTenantSA) getToken() (string, error) {
 			kms.tenantSAName, err)
 	}
 
+	// From Kubernetes v1.24+, secret for service account tokens are not
+	// automatically created. Trying to fetch tokens from service account secret references will fail
+	// refer: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md \
+	// #no-really-you-must-read-this-before-you-upgrade-1.
+	token, err := kms.createToken(sa, c)
+	if err == nil {
+		return token, nil
+	}
+
 	for _, secretRef := range sa.Secrets {
 		secret, sErr := c.CoreV1().Secrets(kms.Tenant).Get(context.TODO(), secretRef.Name, metav1.GetOptions{})
 		if sErr != nil {
@@ -310,7 +314,7 @@ func (kms *vaultTenantSA) getToken() (string, error) {
 		}
 	}
 
-	return kms.createToken(sa, c)
+	return "", fmt.Errorf("failed to find/create ServiceAccount token %s/%s: %w", kms.Tenant, kms.tenantSAName, err)
 }
 
 // getTokenPath creates a temporary directory structure that contains the token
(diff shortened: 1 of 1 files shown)


internal/kms/vault_sa.go
```diff
@@ -298,6 +295,16 @@ func (kms *vaultTenantSA) getToken() (string, error) {
...
+	// From Kubernetes v1.24+, secret for service account tokens are not
+	// automatically created. Trying to fetch tokens from service account secret references will fail
+	// refer: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md \
+	// #no-really-you-must-read-this-before-you-upgrade-1.
+	token, err := kms.createToken(sa, c)
+	if err == nil {
+		return token, nil
+	}
+
+	log.WarningLogMsg("Failed to create SA token, err: %v. Proceeding to fetch token from secret references...", err)
```
Madhu-1: log this warning only once, or the log will be flooded if kubernetes doesn't support token creation.

saiprashanth173: I can skip logging this message if that makes sense. I am not sure if there is a way to log it only once, do you have some suggestions. Please let me know.

Rakshith-R: Yes I think we can skip logging completely. 

Rest of the pr looks good to me,

Thanks !

saiprashanth173: Thanks for your feedback. Removed logging :)

Madhu-1 (dismissed):

internal/kms/vault_sa.go
```diff
@@ -298,6 +293,15 @@ func (kms *vaultTenantSA) getToken() (string, error) {
...
 	}
 
+	// From Kubernetes v1.24+, secret for service account tokens are not
+	// automatically created. Trying to fetch tokens from service account secret references will fail
+	// refer: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md \
+	// #no-really-you-must-read-this-before-you-upgrade-1.
+	token, err := kms.createToken(sa, c)
+	if err == nil {
+		return token, nil
+	}
```
Rakshith-R, line 303:
```suggestion
	token, createTokenErr := kms.createToken(sa, c)
	if createTokenErr == nil {
		return token, nil
	}
```

Rakshith-R (changes requested):
I think we would be completely missing the err which may have happened during createToken.
if it was something that would be helpful. 

I'd prefer we make the requested changes, so we can capture both the errors if they occur, 

wdyt @Madhu-1 @saiprashanth173 ?

internal/kms/vault_sa.go
Madhu-1: not sure what the change above will cover. If we want to be more specific. Continue to fetch token from the SA only if kubernetes doesn't support TokeRequest API else return all other errors.

internal/kms/vault_sa.go
```diff
@@ -310,7 +314,7 @@ func (kms *vaultTenantSA) getToken() (string, error) {
 		}
 	}
 
-	return kms.createToken(sa, c)
+	return "", fmt.Errorf("failed to find token in ServiceAccount %s/%s", kms.Tenant, kms.tenantSAName)
```
Rakshith-R:
```suggestion
	err = fmt.Errorf("failed to find token in ServiceAccount %s/%s", kms.Tenant, kms.tenantSAName)
	if createTokenErr != nil {
		err = fmt.Errorf(%w and failed to create token: %w",err, createTokenErr)
	}
	return "", err
```

internal/kms/vault_sa.go
Rakshith-R: apologies, the other part of the requested changes was not posted by mistake.

Rakshith-R (changes requested):

internal/kms/vault_sa.go
saiprashanth173: Great catch! Yes it makes sense to include this error as part of the final error. I updated the code to include it in the error being returned.

internal/kms/vault_sa.go
saiprashanth173: We only reach here if createTokenErr is not `nil`. I felt `if` here is redundant. Included it as part of final error.

Rakshith-R (approved):
Thanks ! ,
LGTM

Madhu-1 (approved):

PR #3379, opened 2022-09-09
rbd: implments getVolumeReplicationInfo

yati1998:
This commit implements getVolumeReplicationInfo
to get the last sync time and update it in volume
replication CR.

Signed-off-by: yati1998 <ypadia@redhat.com>


internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +835,116 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	return resp, nil
 }
 
+// GetVolumeReplicationInfo extracts the RBD volume information from the volumeID, If the
+// image is present, mirroring is enabled and the image is in primary state.
+// If yes it will get the volume replication information.
+func (rs *ReplicationServer) GetVolumeReplicationInfo(ctx context.Context,
+	req *replication.GetVolumeReplicationInfoRequest,
```
nixpanic, line 842: move all parameters on a new line, it is a recommendation in our own coding standard

internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +835,116 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+	}
+
+	localStatus, err := mirrorStatus.LocalStatus()
+	if err != nil {
+		log.ErrorLog(ctx, err.Error())
+
+		return nil, fmt.Errorf("failed to get local status: %w", err)
+	}
+
+	description := localStatus.Description
```
nixpanic, line 908: Is this some JSON format? You could use `JSON.Unmarshal()` and parse it in a much safer way.

internal/rbd/replicationcontrollerserver.go
yati1998: just followed the format of other operations defined above and all seems to be this way.

internal/rbd/replicationcontrollerserver.go
nixpanic: Parsing should be done in a helper function, which can easily be unit-tested. Make sure that is done where you can. Include real, good and bad examples of the `localStatus.Description` contents for the tests.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +835,116 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	return resp, nil
 }
 
+// GetVolumeReplicationInfo extracts the RBD volume information from the volumeID, If the
+// image is present, mirroring is enabled and the image is in primary state.
+// If yes it will get the volume replication information.
```
nixpanic: comments are repeating themselves, 1x is sufficient :wink:

nixpanic (changes requested):
The parsing of `localStatus.Description` needs improvements and testing.

Note that there is a typo in the subject of the commit+PR, please correct that with a next version of this change too.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +835,122 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+	lastSyncTime := timestamppb.New(lastUpdateTime)
+
+	resp := &replication.GetVolumeReplicationInfoResponse{
+		LastSyncTime: lastSyncTime,
+	}
+
+	return resp, nil
+}
+
+func getLocalSnapshotTime(description string) (string, error) {
```
nixpanic: Please include a unit-test and examples of how the description is formatted. Without knowing the format, it is not possible to judge if the parsing is correct.

And really, it looks like parsing JSON, why not use `json.Unmarshal()` for it?

internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +835,122 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+		return nil, fmt.Errorf("failed to get local status: %w", err)
+	}
+
+	description := localStatus.Description
+	localSnapTime, err := getLocalSnapshotTime(description)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get local snapshot time: %w", err)
+	}
+
+	localSnapTime = localSnapTime[:len(localSnapTime)-1]
```
nixpanic: You just got the time from a function, why isn't the return format in something that you can use directly? Ideally no further parsing/converting is needed here anymore.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +836,118 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+
+// This function gets the local snapshot time from the description
+// of localStatus and converts it into required type.
+func getLastSyncTime(description string) (*timestamppb.Timestamp, error) {
+	// Format of the description will be as followed:
+	// description = "replaying,{"bytes_per_second":0.0,
+	// "bytes_per_snapshot":149504.0,"local_snapshot_timestamp":1662655501
+	// ,"remote_snapshot_timestamp":1662655501}"
+	splittedString := strings.SplitN(description, ",", 2)
+	type localStatus struct {
```
ShyamsundarR, line 933: This should potentially be optional (`omitempty`)to handle the case where there is no lastSyncTime reported, which would also default the time returned here to "0" (or the oldest timestamp). As a result if there is not lastSyncTime the rest of the stack and the VolumeReplication resource would also return "0".

Currently, if not made optional we would fail this call and keep returning errors to the caller, possibly causing frequent reconciles.

nixpanic: I think `omitempty` is only used while creating JSON, not when parsing it back?

The default (if missing) is set to `0`, like an uninitialized int. If there is a need to differentiate between `0` and unset, it can be done like
```go
type localStatus struct {
	LocalSnapshotTime *int64 `json:"local_snapshot_timestamp"`
}
```
With that `LocalSnapshotTime` will either be `nil` or set to the value that was returned by Ceph.

ShyamsundarR: We should be fine then, as long as missing key is not an error.

internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +836,118 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+		LocalSnapshotTime int64 `json:"local_snapshot_timestamp"`
+	}
+
+	var localSnapTime localStatus
+	err := json.Unmarshal([]byte(splittedString[1]), &localSnapTime)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal description: %w", err)
+	}
+	localSnapshotTime := localSnapTime.LocalSnapshotTime
+	if localSnapshotTime == 0 {
```
ShyamsundarR: This should also not return an error (as discussed with @yati1998 ), as understood a missing localSnapshotTime implies it was never synced even once, and so the reported time should be "0", which would be the beginning of epoch.

internal/rbd/replicationcontrollerserver_test.go
```diff
@@ -432,3 +434,34 @@ func TestCheckRemoteSiteStatus(t *testing.T) {
...
+			"valid description",
+			//nolint:lll // sample output cannot be split into multiple lines.
+			`replaying,{"bytes_per_second":0.0,"bytes_per_snapshot":149504.0,"local_snapshot_timestamp":1662655501,"remote_snapshot_timestamp":1662655501}`,
+			timestamppb.New(time.Unix(1662655501, 0)),
+		},
+		{
+			"invalid description",
+			`replaying,{"bytes_per_second":0.0,"bytes_per_snapshot":149504.0,"remote_snapshot_timestamp":1662655501}`,
+			nil,
+		},
```
nixpanic, line 465: please also add a test-case where `description` is an empty string (`""`), and contains some invalid JSON  (like missing `}`).

internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +836,116 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+		LocalSnapshotTime int64 `json:"local_snapshot_timestamp"`
+	}
+
+	var localSnapTime localStatus
+	err := json.Unmarshal([]byte(splittedString[1]), &localSnapTime)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal description: %w", err)
+	}
+	localSnapshotTime := localSnapTime.LocalSnapshotTime
+	// convert the local snapshot time to UTC
```
nixpanic: This seems like a crucial thing? Is the timestamp that gets returned by `GetVolumeReplicationInfo` expected to be in UTC? I do not remember reading that in the spec, if it is not there, the notes in the spec need to be amended.

nixpanic (changes requested):
few minor things, looks good to me otherwise

internal/rbd/replicationcontrollerserver.go
yati1998: no it is not, timestamp returned by `GetVolumeReplicationInfo` is `*timestamppb.Timestamp`, it's just an intermediate conversion.

yati1998:
addressed all the comments, can you please add your final reviews @nixpanic @Rakshith-R @ShyamsundarR @Madhu-1

internal/rbd/replicationcontrollerserver.go
```diff
@@ -834,6 +836,117 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	return resp, nil
 }
 
+// GetVolumeReplicationInfo extracts the RBD volume information from the volumeID, If the
+// image is present, mirroring is enabled and the image is in primary state.
+// If yes it will get the volume replication information.
```
Rakshith-R:
```suggestion
// GetVolumeReplicationInfo extracts the RBD volume information from the volumeID, If the
// image is present, mirroring is enabled and the image is in primary state.
```

nixpanic (approved):

Rakshith-R (approved):

ShyamsundarR (approved):
LGTM! Thanks...

nixpanic:
This PR can be rebased with Mergifyio and have the `ci/retry/e2e` label added once #3380 is merged.

Madhu-1:
@yati1998, even though its a new RPC planning to backport this to 3.7?

yati1998:
> @yati1998, even though its a new RPC planning to backport this to 3.7?

IMO we can have it in next release, maybe minor. But if you guys think to backport it, I can send a PR.

yati1998:
> @yati1998, even though its a new RPC planning to backport this to 3.7?

IMO we can have it in next release, maybe minor. But if you guys think to backport it, I can send a PR.

Rakshith-R:
adding skip e2e, since we dont have any e2e for volrep operation.

Rakshith-R:
@mergifyio rebase

Rakshith-R:
@mergifyio refresh

PR #3174, opened 2022-06-09
rbd: create token and use it for vault SA

Rakshith-R:
create the token if kubernetes version is 1.24+ and use it for vault sa.
    
Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>
Signed-off-by: Rakshith R <rar@redhat.com>

Resolves: #3135

PR #3394, opened 2022-09-14
doc: add proposal for fetching ceph credentials from KMS

saiprashanth173:
Add proposal to support fetching Ceph credentials from KMS

Signed-off-by: Prashanth Dintyala <vdintyala@nvidia.com>

Ceph-CSI currently relies on K8s secrets for ceph user (provisioner, node-stage, expand-secret)  credentials. Given that ceph-csi already supports managing volume encryption keys through vault and other KMS, it should be possible to extend the existing implementation to fetch Ceph user credentials as well.

This PR has a high level design proposal for this feature. 

Related slack discussion with @Madhu-1  @nixpanic: https://cephcsi.slack.com/archives/CJV6G2DFT/p1662485493014329 

---

<details>

  progress or results

</details>

### DIFF
--- a/docs/design/proposals/creds-from-kms.md
+++ b/docs/design/proposals/creds-from-kms.md
@@ -1,18 +1,37 @@
 # Fetching Ceph user credentials from KMS
 
-## Proposal:
+## Proposal
+
 Ceph-CSI supports several KMS implementations for storing and
 fetching DEKs (Data Encryption Keys) to support encryption of RBD volumes.
 Focus of this proposal is to extend the existing KMS implementation to fetch
 Ceph User credentials used for creating, deleting, mapping and resizing RBD volumes.
 
-## Background:
+### Benefits
 
-### Credentials Management:
+- For Ceph-CSI setup where Ceph is deployed externally, either as
+  a standalone or through Rook, one needs to manually obtain the Ceph user keyring(s)
+  and create the K8s required secrets. With this feature,
+  sensitive keyring(s) can be stored securely in an
+  external secret management system such as Hashicorp Vault,
+  AWS Secret Manager, etc. and have Ceph-CSI pull these
+  sensitive keys directly from them.
+- Provides additional security in comparison with K8s secrets,
+  which are relatively easy to access and
+  store sensitive secrets as unencrypted based64 encoded text.
 
-Ceph-CSI relies on K8s CSI components for fetching Ceph user credentials.
+### Drawback
+
+Adds additional overhead as on each CSI RPC call
+a request is made to KMS for fetching credentials.
+
+## Background
+
+### Credentials Management
+
+Ceph-CSI relies on K8s CSI components for obtaining the Ceph user credentials.
 K8s CSI uses the following reserved StorageClass parameters for fetching
-credentials and passes them to Ceph-CSI.
+credentials and passes them to Ceph-CSI RPC calls.
 
 ```yaml
 csi.storage.k8s.io/controller-expand-secret-name
@@ -23,20 +42,27 @@ csi.storage.k8s.io/node-stage-secret-name
 csi.storage.k8s.io/node-stage-secret-namespace
 ```
 
-### KMS Implementation:
-Ceph-CSI utilizes the encryptionKMSID StorageClass parameter and corresponding config
-entry in KMS ConfigMap (eg. [ceph-csi-encryption-kms-config](https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/kms-config.yaml)) to integrate with the KMS.
+### KMS Implementation
 
-## Extending existing implementation:
+Ceph-CSI utilizes the `encryptionKMSID` StorageClass parameter and
+corresponding config entry in KMS ConfigMap
+(eg. [ceph-csi-encryption-kms-config](https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/kms-config.yaml))
+to integrate with the KMS.
 
-> ### Deviation from existing approach:
-> * *StorageClass* parameters cannot be used for credentials as
-> they are not passed in all CSI requests e.g DeleteVolumeRequest.
-> * Similarly, using K8s namespace as tenant name for KMS wouldn’t
+## Extending existing implementation
+
+> ### Deviation from existing approach
+>
+> - *StorageClass* parameters, such as `encryptionKMSID`,
+> cannot be used for credentials as
+> they are not passed in all CSI requests e.g. DeleteVolumeRequest.
+> - Similarly, using K8s namespace as tenant name for KMS wouldn’t
 > be possible as this information not included as part of all CSI requests.
 
-* KMS ID is expected to be present as part of provisioner, node-stage, controller-expand secrets, along with Ceph userID.
-```
+- KMS ID is expected to be present as part of provisioner,
+  node-stage, controller-expand secrets, along with Ceph userID.
+
+```yaml
 apiVersion: v1
 kind: Secret
 metadata:
@@ -46,8 +72,9 @@ stringData:
   kmsID: <kms-id>
  ```
 
-
-* A new ConfigMap named ceph-csi-creds-kms-config similar to [ceph-csi-encryption-kms-config](https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/kms-config.yaml) will be added.
+- A new ConfigMap named ceph-csi-creds-kms-config similar to
+  [ceph-csi-encryption-kms-config](https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/kms-config.yaml)
+  will be added.
 
 ```yaml
 apiVersion: v1
@@ -59,10 +86,13 @@ metadata:
     name: ceph-csi-creds-kms-config
 ```
 
-* For KMS that rely on Tenant namespace, a new optional field name "tenantNamespace" is added in addition to existing KMS config.
-If "tenantNamespace" is not provided, Ceph-CSI namespace will be used to lookup required ServiceAccount, ConfigMap or Secrets.
+- For KMS that rely on Tenant namespace, a new optional field name "tenantNamespace"
+  is added in addition to existing KMS config.
+  If "tenantNamespace" is not provided, Ceph-CSI namespace will be used to lookup
+  required ServiceAccount, ConfigMap or Secrets.
 
 **Example KMS config:**
+
 ```json
     "vault-tenant-sa-test": {
         "credsKMSType": "vaulttenantsa",
@@ -77,13 +107,40 @@ If "tenantNamespace" is not provided, Ceph-CSI namespace will be used to lookup 
         "tenantNamespace": "any-tenant"
     }
 ```
-* Credential KMS config wouldn’t support nested "tenant" config that is provided in the "tenant" field,
-i.e. each tenant is expected to have their own entry in the KMS ConfigMap.
 
-## Code Changes:
-* New interface named CredStore will be introduced
+- Credential KMS config wouldn’t support nested tenant config
+  provided using the `tenant` field,
+  which implies each tenant is expected to have their own entry in the KMS ConfigMap.
 
-```code
+### Backward Compatibility
+
+KMS integration will only be enabled when `ceph-csi-creds-kms-config` ConfigMap exists
+and secrets contain the `kmsID` key. In case where secrets contain
+both `kmsID` and `userKey`
+the keyring provided secrets will be used for creating the credential object.
+
+### Integration with Rook
+
+Rook integration for KMS support for credentials would be similar
+PVC encryption.
+A new option `CSI_ENABLE_KMS_CREDS` will be added to Rook operator.
+When it is set to `true`, Rook will create
+Ceph-CSI Deployments and DaemonSets that mount the Creds KMS ConfigMap.
+
+At the time of writing this proposal, Rook only
+supports storing Ceph keyring as secrets.
+This [issue](https://github.com/rook/rook/issues/6374)
+tracks adding support for storing
+all Rook secrets in Vault. Till there is progress on this ticket,
+it is assumed that an external script will be
+responsible for creating the Ceph user and putting the keyring
+in a KMS. Rook will only deploy Ceph-CSI with correct config.
+
+## Code Changes
+
+- New interface named CredStore will be introduced
+
+```go
 // CredStore allows KMS instances to implement a modular backend for Creds
 // storage.
 type CredStore interface {
@@ -92,5 +149,5 @@ type CredStore interface {
 }
 ```
 
-* Provider Initializers will be refactored, if required,
-to accommodate both volume encryption and credential storage requirement.
+- Provider Initializers will be refactored, if required,
+  to accommodate both volume encryption and credential storage requirement.
--- a/docs/design/proposals/creds-from-kms.md
+++ b/docs/design/proposals/creds-from-kms.md
@@ -6,7 +6,7 @@ Ceph-CSI supports several KMS implementations for storing and
 fetching DEKs (Data Encryption Keys) used for RBD volume encryption.
 The focus of this proposal is to leverage existing KMS implementation
 to support obtaining Ceph user keyring used while
-creating, deleting, mapping and resizing RBD volumes.
+creating, deleting, mapping and resizing volumes.
 
 ### Benefits
 
@@ -25,16 +25,16 @@ creating, deleting, mapping and resizing RBD volumes.
 
 - Adds additional overhead as on each CSI RPC call
   a request is made to KMS for fetching credentials.
-- Possible risk of hitting the KMS rate limit
+- Possible risk of hitting the KMS rate limit.
 
 ## Extending existing implementation
 
-- KMS ID is  provided as part of the provisioner,
+- KMS ID is provided as part of the provisioner,
   node-stage, and controller-expand secrets, along with Ceph `userID`, as shown below.
   Having this information as part of *StorageClass* parameters,
   like using the `encryptionKMSID` key for volume encryption,
   wouldn't work as these parameters are not passed
-  to all CSI RPCs, e.g. `DeleteVolume`.
+  to all CSI RPCs, e.g. `DeleteVolume`, `ControllerExpand` etc.
 
 ```yaml
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


humblec:
@saiprashanth173 thanks for the proposal! 

Looking into this, I am really confused and here is why.


>Ceph-CSI relies on K8s CSI components for fetching Ceph user credentials.
K8s CSI uses the following reserved StorageClass parameters for fetching
credentials and passes them to Ceph-CSI.
csi.storage.k8s.io/controller-expand-secret-name
csi.storage.k8s.io/controller-expand-secret-namespace
csi.storage.k8s.io/provisioner-secret-name
csi.storage.k8s.io/provisioner-secret-namespace
csi.storage.k8s.io/node-stage-secret-name
csi.storage.k8s.io/node-stage-secret-namespace
>


These secrets and its management has nothing to do with the CSI driver itself. This is the mechanism co-ordinated from CO ( ex: Kube) based on the CSI spec . We ( as in Ceph CSI drivers) just consume it in the path! It is not something like "go and fetch" from "secrets" from CSI driver pov. These are provided as part of the RPC call to the driver like any other SC-> CSI driver combo. 
These entries in the SC is common for all CSI drivers and nothing specific to Ceph CSI driver thus Ceph CSI driver has no role here. 

> StorageClass parameters cannot be used for credentials as
> they are not passed in all CSI requests e.g DeleteVolumeRequest.
>

This is well intended and its not supposed to be used and it works. 

Regardless, keeping the Ceph user information in KMS/Kube object is something we have to get the discussions on and we dont have to clobber it with above secrets..etc. In that light, adding more details on "use case" , "advantages compared to existing model" , "security aspects" (if any), "backward compatibliity" , "integration to rook as it comes in the workflow"  ...etc would be value added to the proposal.

saiprashanth173:
@humblec  thanks for your feedback. 

Sorry it took me some time to get back, I was out of the office last week. 

 >  These secrets and its management has nothing to do with the CSI driver itself. This is the mechanism co-ordinated from CO ( ex: Kube) based on the CSI spec . We ( as in Ceph CSI drivers) just consume it in the path! It is not something like "go and fetch" from "secrets" from CSI driver pov. These are provided as part of the RPC call to the driver like any other SC-> CSI driver combo.
These entries in the SC is common for all CSI drivers and nothing specific to Ceph CSI driver thus Ceph CSI driver has no role here.

I initially added the background section to give some context on how the current flow works. My intention was not to comment on the mechanism, it was just to summarize it. I removed the background section for now as it appears to add confusion and isn't solving the intended purpose.

> This is well intended and its not supposed to be used and it works.

Just to clarify, I was trying to set context for why `kmsID` in case of credentials should be a part `expand-secret`, `provisioner-secret` or `node-stage-secret` instead of specifying it as a StorageClass parameter, like we do in the case of volume encryption. If you feel this isn't very clear, I can restructure this section a bit. Please let me know. 


> Regardless, keeping the Ceph user information in KMS/Kube object is something we have to get the discussions on and we dont have to clobber it with above secrets..etc. In that light, adding more details on "use case" , "advantages compared to existing model" , "security aspects" (if any), "backward compatibliity" , "integration to rook as it comes in the workflow" ...etc would be value added to the proposal.


I tried to add some more details to the proposal. I will try to refine the proposal further. Please let me know if you have any other suggestions on the latest version. Happy to make further changes based on your feedback :)

saiprashanth173:
@Madhu-1 @humblec @nixpanic do you have any suggestion/comments on the proposal? Please let me know.

Madhu-1:
@saiprashanth173 is this still WIP design doc?

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,130 @@
+# Fetching Ceph user credentials from KMS
+
+## Proposal
+
+Ceph-CSI supports several KMS implementations for storing and
+fetching DEKs (Data Encryption Keys) used for RBD volume encryption.
+The focus of this proposal is to leverage existing KMS implementation
+to support obtaining Ceph user keyring used while
+creating, deleting, mapping and resizing RBD volumes.
```
Madhu-1: we dont need to stick with RBD volumes here as this can be used with cephfs also?

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Benefits
+
+- For scenarios where Ceph is deployed externally, either as
+  a standalone or through Rook, one needs to obtain the Ceph keyring(s)
+  and manually create the K8s secrets needed by Ceph-CSI. With this feature,
```
Madhu-1, line 15: This is true only if CephCSI is depoyed externally, when Rook deploys the ceph cluster it creates the required secrets for cephcsi.

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,130 @@
...
+  keys directly from them.
+- More secure compared to K8s secrets,
+  which are relatively easy to access and
+  store sensitive secrets as unencrypted based64 encoded text.
+
+### Drawbacks
+
+- Adds additional overhead as on each CSI RPC call
+  a request is made to KMS for fetching credentials.
+- Possible risk of hitting the KMS rate limit
```
Madhu-1: Wont this be a major drawback?

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Extending existing implementation
+
+- KMS ID is  provided as part of the provisioner,
```
Madhu-1:
```suggestion
- KMS ID is provided as part of the provisioner,
```

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,130 @@
...
+  node-stage, and controller-expand secrets, along with Ceph `userID`, as shown below.
+  Having this information as part of *StorageClass* parameters,
+  like using the `encryptionKMSID` key for volume encryption,
+  wouldn't work as these parameters are not passed
+  to all CSI RPCs, e.g. `DeleteVolume`.
```
Madhu-1: ControllerExpand RPC also

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,130 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: csi-rbd-provisioner
+stringData:
+  userID: <ceph-user-id>
+  kmsID: <kms-id>
+ ```
+
+- A new ConfigMap with default name `ceph-csi-creds-kms-config` similar to
```
Madhu-1: why we need extra kms configmap cant we use existing kms configuration?

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,130 @@
...
+apiVersion: v1
+kind: ConfigMap
+data:
+    config.json: |-
+        {<kms-id>: <config>}
+metadata:
+    name: ceph-csi-creds-kms-config
+```
+
+- For KMS that rely on tenants' namespace to obtain required
```
Madhu-1, line 53: this is not something specific to this feature right?

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,130 @@
...
+  which implies each tenant is expected to have their own entry in the KMS ConfigMap.
+
+### Backward Compatibility
+
+KMS integration will only be enabled when `ceph-csi-creds-kms-config` ConfigMap exists
+and CSI secrets contain the `kmsID` key. In case where secrets contain
+both `kmsID` and `userKey`
+the keyring provided secrets will be used for creating the credential object.
+
+### Integration with Rook
```
Madhu-1, line 89: @travisn can you please check this one?

Madhu-1 (commented):
Overall looks good but as mention this is kind of overhead and have performance impact if someone chooses security over performance its good for them.

IMHO again it wont solve this issue, if the kubernetes secrets are compromised anyone can gain access to KMS(as all the KMS details are stored in kubernetes configmap and secret) and get the ceph cluster access.

Rakshith-R:
> Overall looks good but as mention this is kind of overhead and have performance impact if someone chooses security over performance its good for them.
>
> IMHO again it wont solve this issue, if the kubernetes secrets are compromised anyone can gain access to KMS(as all the KMS details are stored in kubernetes configmap and secret) and get the ceph cluster access.

+1

docs/design/proposals/creds-from-kms.md
saiprashanth173: I agree that this feature would directly fit into external setup scenario. However, I would see this as a valid case even when Rook deploys Ceph cluster. I see that Rook creates the following secrets by default, and the user caps are quite broad:
```
rook-csi-cephfs-node
rook-csi-cephfs-provisioner
rook-csi-rbd-node
rook-csi-rbd-provisioner
``` 

However, if we were to support more advanced use case such as soft multi-tenancy. It might make sense to have a ceph users per tenant and provide permissions to only access specific pool(s). 

Also, if Rook plans to add support for https://github.com/rook/rook/issues/6374. It would be a valid case even in normal scenario.

docs/design/proposals/creds-from-kms.md
saiprashanth173: I feel KMS rate limit is a valid concern for anything, including RBD volume encryption, using the KMS implementation. One possible approach to mitigate this would be to use an in-memory cache of Ceph keyrings, given that keyrings don't change too often we only fetch them when the keyring doesn't exist already or is invalid (i.e. on failed ceph auth).

docs/design/proposals/creds-from-kms.md
saiprashanth173: Technically, we don't need a new config here. Given that structure of encryption and creds KMS config entries slight differ, specially due to the newly added `tenantNamespace` key and missing tenant sub config in case of creds, I thought it might make sense to keep them separate. 

I can change this section to use the same configuration if that makes sense. Please let me know what you think.

docs/design/proposals/creds-from-kms.md
saiprashanth173: Sorry, I don't follow the question completely. 

The need to add "tenantNamespace" in config is specific to this feature, as we don't have the tenant namespace available in all RPC calls. RBD volume encryption uses the tenants' namespace for all its operations.

docs/design/proposals/creds-from-kms.md
Madhu-1: >However, if we were to support more advanced use case such as soft multi-tenancy. It might make sense to have a ceph users per tenant and provide permissions to only access specific pool(s). 

Admin can still create new users with different caps using cephclient CRD and use that with custom storageclasses.

docs/design/proposals/creds-from-kms.md
Madhu-1: :+1: the in-memory cache need to be deleted for few cases like 
* auth issues
* connection issues

docs/design/proposals/creds-from-kms.md
Madhu-1: Yes please

docs/design/proposals/creds-from-kms.md
Madhu-1: add the new field to the existing struct and reuse it, lets avoid too many configurations

Madhu-1:
> > Overall looks good but as mention this is kind of overhead and have performance impact if someone chooses security over performance its good for them.
> > IMHO again it wont solve this issue, if the kubernetes secrets are compromised anyone can gain access to KMS(as all the KMS details are stored in kubernetes configmap and secret) and get the ceph cluster access.
> 
> +1

@saiprashanth173 can you please comment on this one, this helps to take this forward.

nixpanic:
> > > Overall looks good but as mention this is kind of overhead and have performance impact if someone chooses security over performance its good for them.
> > > IMHO again it wont solve this issue, if the kubernetes secrets are compromised anyone can gain access to KMS(as all the KMS details are stored in kubernetes configmap and secret) and get the ceph cluster access.
> > 
> > 
> > +1
> 
> @saiprashanth173 can you please comment on this one, this helps to take this forward.

This is also what I wonder mostly about. The advantage of placing the Ceph credentials in a KMS is not clear to me. Kubernetes still stores credentials to access the KMS, so getting to the Ceph credentials is still possible without much additional effort?

saiprashanth173:
> IMHO again it wont solve this issue, if the kubernetes secrets are compromised anyone can gain access to KMS(as all the KMS details are stored in kubernetes configmap and secret) and get the ceph cluster access.

Sorry, I missed commenting on this. 

I mostly agree that this solution wouldn't solve the issue completely for the KMS integrations that require storing auth tokens in K8s secrets. If we take vault service account based approach, it would make it more difficult (if not impossible) to access credentials compared to K8s secrets, for following reasons:
1. In K8s 1.24+, service account token secrets are not auto-generated and starting Ceph-CSI 3.7 we always attempt to create a service account token and use it with vault. These tokens are short-lived, and creating them would require additional privileges, e.g. RBAC with `serviceaccounts/token`|`create` . Related: https://github.com/ceph/ceph-csi/pull/3174 https://github.com/ceph/ceph-csi/pull/3377 
2. Even with cases where a service account token is compromised, the impact will be restricted only to the vault path which the vault role linked to the service account has access to.

I believe storing secrets in KMS can provide additional level of security. Even with compromised access tokens, most KMS support additional security tightening measures which imposes restrictions on who can access the KMS endpoints, making it slightly more difficult to access sensitive secrets.

Additionally, I see this feature solving another issue, specifically for external Ceph setups, where the process of sharing and rotating keyrings can be made more seamless and secure.

Please let me know what you think. Happy to discuss this further.

docs/design/proposals/creds-from-kms.md
saiprashanth173: >This is true only if CephCSI is depoyed externally, when Rook deploys the ceph cluster it creates the required secrets for cephcsi. 

Sorry, I might have misunderstood your comment. 

In this point, I was only referring CephCSI being deployed externally:
>For scenarios where Ceph is deployed externally, either as
  a standalone or through Rook,

`For scenarios where Ceph is deployed externally,` by this, I mean Ceph cluster is not in the same K8s cluster as Ceph-CSI. `either as
  a standalone or through Rook` by this I mean the external Ceph cluster may have been deployed using Rook or some other tool.  

If you feel the sentence isn't clear enough, I can change that a bit. Please let me know.

> Admin can still create new users with different caps using cephclient CRD and use that with custom storageclasses.

Thanks for letting me know. You are right, for Rook based Ceph and CephCSI deployments this approach might not be applicable, unless Rook supports storing keyrings directly in a KMS.

docs/design/proposals/creds-from-kms.md
saiprashanth173: I added a point on in-memory cache under `Code Changes` sections. 

Do you mind elaborating a bit on what you mean by `connection issues` in this context?

docs/design/proposals/creds-from-kms.md
saiprashanth173: I updated this section for reusing the existing KMS configuration.

docs/design/proposals/creds-from-kms.md
saiprashanth173: Makes sense. I updated the section.

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,131 @@
...
+- Possible risk of hitting the KMS rate limit.
+
+## Extending existing implementation
+
+- KMS ID is provided as part of the provisioner,
+  node-stage, and controller-expand secrets, along with Ceph `userID`, as shown below.
+  Having this information as part of *StorageClass* parameters,
+  like using the `encryptionKMSID` key for volume encryption,
+  wouldn't work as these parameters are not passed
+  to all CSI RPCs, e.g. `DeleteVolume`, `ControllerExpand` etc.
```
nixpanic, line 37: The parameters in the StorageClass are copied into the PersistentVolume.CSIDriver parameters. I think it would be cleaner to not use the common secrets in the StorageClass for this, but directly refer to a `kmsID` with potentially additional parameters like `userID`.

docs/design/proposals/creds-from-kms.md
```diff
@@ -0,0 +1,131 @@
...
+  KMS ConfigMap, very similar to the volume encryption config found in
+  [cph-csi-encryption-kms-config](https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/kms-config.yaml).
+
+- For KMS that rely on tenants' namespace to obtain required
+  ServiceAccount, ConfigMap, or Secrets,
+  an optional field, "tenantNamespace" will be added to the KMS config.
+  As the tenant namespace isn't available in all CSI requests,
+  value from provided in this field will be used instead.
+  If "tenantNamespace" field is absent, Ceph-CSI namespace
+  will be used as default.
```
nixpanic, line 59: I think this is not very practical in the end. Why have a _default_ user for some operations, and use the tenant for others? Using different Ceph credentials for different operations on a single volume looks rather confusing. Usually you would also try to prevent normal users from accessing the Ceph cluster, so those users should not need to have access to Ceph credentials.

nixpanic (commented):
I agree that this proposal could make it easier to deny access (or rotate tokens) for certain users/accounts if credentials are required to access a KMS and fetch the Ceph credentials from there. Careful auditing of the access logs in the KMS might prevent requiring rotating the Ceph credentials after someone obtained the short-living KMS credentials.

It would also become possible (not sure about the _smart_ or _recommended_ practice) to share Ceph credentials from a single location (KMS) over multiple Kubernetes clusters.

With that, I'm supportive of the feature and some users might be interested in it.

docs/design/proposals/creds-from-kms.md
saiprashanth173: @nixpanic to clarify my point:
1. the `tenantNamespace` here is only used for getting everything required by the KMS to fetch the Ceph user credentials. One such example is obtaining the KMS access token, which could be creating a service account token for vault SA or fetching tokens from K8s secret for some other KMS. 
2. Any volume associated with a storage class would always use same Ceph credentials for all its operations. I see Ceph credentials as pair of `userID and userKey`, where userID is obtained from `csi.storage.k8s.io/*-secret-name`  and `userKey` is fetched from the KMS. The user will never change until we change `userID` and KMS config. 
3. The KMS will only use one namespace for all the operations. If the `tenantNamespace` is present in the KMS config, it will always use that. If not, it always looks in Ceph-CSI namespace as default. Again, this is only for KMS authentication so that we can fetch the required Ceph credentials.

docs/design/proposals/creds-from-kms.md
saiprashanth173: It is not very clear to me on how we can obtain the `PersistentVolume.CSIDriver`  parameters from calls like `DeleteVolume`. 

Looking at the DeleteVolume method and DeleteVolumeRequest struct:

```golang
func (cs *ControllerServer) DeleteVolume(
	ctx context.Context,
	req *csi.DeleteVolumeRequest,
) (*csi.DeleteVolumeResponse, error)

// ....

type DeleteVolumeRequest struct {
	// The ID of the volume to be deprovisioned.
	// This field is REQUIRED.
	VolumeId string `protobuf:"bytes,1,opt,name=volume_id,json=volumeId,proto3" json:"volume_id,omitempty"`
	// Secrets required by plugin to complete volume deletion request.
	// This field is OPTIONAL. Refer to the `Secrets Requirements`
	// section on how to use this field.
	Secrets              map[string]string `protobuf:"bytes,2,rep,name=secrets,proto3" json:"secrets,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
	XXX_NoUnkeyedLiteral struct{}          `json:"-"`
	XXX_unrecognized     []byte            `json:"-"`
	XXX_sizecache        int32             `json:"-"`
}
```

I only see that we can get VolumeId as part of this request. Are you suggesting we get the `PersistentVolume` object here to obtain these parameters? If yes, wouldn't that require an additional API call?

Do you mind helping me understand this better?

nixpanic (approved):

PR #3396, opened 2022-09-19
rbd: ignore extra data in kmip credential secret

Rakshith-R:
This commit removes check for unsupported option
for `kmip` KMS since user may store extra data here and
it does not affect functionality of the feature.

### DIFF
--- a/internal/kms/kmip.go
+++ b/internal/kms/kmip.go
@@ -313,9 +313,6 @@ func (kms *kmipKMS) getSecrets() (map[string]string, error) {
 		switch k {
 		case kmipClientKey, kmipCLientCert, kmipCACert, kmipUniqueIdentifier:
 			config[k] = string(v)
-		default:
-			return nil, fmt.Errorf("unsupported option for KMS "+
-				"provider %q: %s", kmsTypeKMIP, k)
 		}
 	}
 


Rakshith-R:
Skipping e2e since there is no e2e for kmip yet

humblec (approved):

humblec:
@Rakshith-R ideally we could just warn and log instead of erroring out from default case..

Rakshith-R:
> @Rakshith-R ideally we could just warn and log instead of erroring out from default case.. 

I don't think that'll be required too,
We don't need to bother about extra data that may have been set, it's not required for kmip functionality at all.

humblec:
> > @Rakshith-R ideally we could just warn and log instead of erroring out from default case..
> 
> I don't think that'll be required too, We don't need to bother about extra data that may have been set, it's not required for kmip functionality at all.

than the functionality, its for educating the admin/user..  its not required to have that 'field` filled in is why we are not stopping the operation or not coming out. However its good to log most of the times too.

internal/kms/kmip.go
```diff
@@ -313,9 +313,6 @@ func (kms *kmipKMS) getSecrets() (map[string]string, error) {
 		switch k {
 		case kmipClientKey, kmipCLientCert, kmipCACert, kmipUniqueIdentifier:
 			config[k] = string(v)
-		default:
```
Madhu-1, line 316: I dont see any reason to make kmip as special here. This default check exists for all other kms as well. If we allow adding more data to the secret specific to csi there are chances for human mistakes when they update/recreate the secrets. why can't admin create an extra secrete specific to csi itself without any unwanted data?

SanjalKatiyar: In HPCS as well, one Secret is getting created and is shared across nobaa/rook/ceph-csi (but yeah I agree, difference there is that no extra key is added during runtime).
This (Thales) Secret have other info as well which is still common between noobaa/rook/ceph-csi, is it alright to replicate those information twice when we have an option to re-use the same Secret ?
Shouldn't operator give an error if required information is not present in Secret and just neglect extra information which it do not need (it should be safe to neglect if not needed) ?
@Madhu-1 wdyt ?

SanjalKatiyar: btw admin can create separate Secrets, I am just confirming if we should or not...

Madhu-1: it's good to have a dedicated secret for the below reasons
* To avoid human errors 
* Keeps the secret data very minimal and even more secured if it is compromised as we are going to use secret per service it's not going to affect all the services.
* Avoid multiple services depending on one global secret

SanjalKatiyar: ack ! thanks !

Rakshith-R:
> it's good to have a dedicated secret for the below reasons
> 
> * To avoid human errors
> * Keeps the secret data very minimal and even more secured if it is compromised as we are going to use secret per service it's not going to affect all the services.
> * Avoid multiple services depending on one global secret


👍 closing this pr

PR #3415, opened 2022-10-06
ci: script changes to run minikube with none driver

Madhu-1:
install required crio to run minikube with none driver and added if check to exclude minikube arguments which doesnot required for none driver.
    
### DIFF
--- a/scripts/minikube.sh
+++ b/scripts/minikube.sh
@@ -1,4 +1,5 @@
 #!/bin/bash -e
+set -o xtrace
 
 #Based on ideas from https://github.com/rook/rook/blob/master/tests/scripts/minikube.sh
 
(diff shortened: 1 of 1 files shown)


nixpanic:
I think you also need to modify https://github.com/ceph/ceph-csi/blob/ci/centos/single-node-k8s.sh

Madhu-1:
> I think you also need to modify https://github.com/ceph/ceph-csi/blob/ci/centos/single-node-k8s.sh

Yes for now hardcoding the script to start cluster with none driver for testing, if everything goes well i move some installation and configurtion to above script

scripts/minikube.sh
```diff
@@ -176,6 +177,14 @@ else
     DISK_CONFIG=""
 fi
 
+EXTRA_MINIKUBE_ARGS=${EXTRA_MINIKUBE_ARGS:-"--container-runtime=cri-o"}
+
+function install_crio() {
+    dnf install jq conntrack docker -y
```
nixpanic: please do not use Docker, use Podman instead.

Isn't there a repository that provides CRIO RPMs?

nixpanic:
@mergifyio rebase

nixpanic:
#3419 and #3420 can hopefully replace this PR.

PR #3441, opened 2022-10-17
e2e: generate unique name for PVC

### LINKED ISSUE #2533: [E2E] generate unique names for PVC/PV and pod Names
With the current E2E it's difficult to identify the problem based on the PVC name or the pod name as we reuse the same name in multiple tests. It's good to generate a unique name each time which helps in debugging.

Madhu-1:
generate unique name for PVC and snapshots which helps in debugging E2E test failures.

fixes #2533

### DIFF
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -342,6 +342,7 @@ var _ = Describe(cephfsType, func() {
 					e2elog.Failf("failed to load application: %v", err)
 				}
 				app.Namespace = f.UniqueName
+				app.Spec.Volumes[0].PersistentVolumeClaim.ClaimName = pvc.Name
 				baseAppName := app.Name
 
 				err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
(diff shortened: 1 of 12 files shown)


nixpanic:
Is `f.UniqueName` unique enough? I thought that was a unique name for the whole test suite, so it would be the same for all tests?

Madhu-1:
> Is `f.UniqueName` unique enough? I thought that was a unique name for the whole test suite, so it would be the same for all tests?

AFAIK its unique for the whole suite not for individual tests `By()`

nixpanic:
> AFAIK its unique for the whole suite not for individual tests `By()`

Indeed, so it does not make it (much) easier to track a PVC for a particular test? When you're reworking this, does it not make sense to make the name unique per `By()`?

Madhu-1:
> > AFAIK its unique for the whole suite not for individual tests `By()`
> 
> Indeed, so it does not make it (much) easier to track a PVC for a particular test? When you're reworking this, does it not make sense to make the name unique per `By()`?

it was hard to debug it earlier as the name is not unique per test `By`. I will try keep it that way now a unique name is generated for each PVC.

PR #3456, opened 2022-10-19
Cleanup snapshots before removing rbd image

### LINKED ISSUE #3416: Delete snapshots before deleting RBD image
I want image snapshots to be deleted on image delete request. So basically smth like this:
```
# rbd snap purge IMAGE_ID
# rbd rm IMAGE_ID
```

It is important because images which have snapshots cannot be deleted from ceph and get stuck in trash. If you try to delete k8s pvc(pv) which has rbd snapshots in ceph, the image won't be actually deleted from ceph and will get stuck in trash. There will be a lot of messages in ceph mgr logs like this:
```
debug 2022-09-30T10:07:30.214+0000 7fb618bba700  0 [rbd_support INFO root] execute_trash_remove: task={"sequence": 47220, "id": "701e47ae-fb41-4838-ae32-a26e802d0097", "message": "Removing image csi/45e917746337a from trash", "refs": {"action": "trash remove", "pool_name": "csi", "pool_namespace": "", "image_id": "45e917746337a"}, "retry_message": "[errno 39] RBD image has snapshots (error deleting image from trash)", "retry_attempts": 9, "retry_time": "2022-09-30T10:07:28.411597", "in_progress": true, "progress": 0.0}
debug 2022-09-30T10:07:30.230+0000 7fb618bba700  0 [rbd_support ERROR root] execute_task: [errno 39] RBD image has snapshots (error deleting image from trash)
```
 
The image isn't getting stuck in trash.

# Additional context #
As I can see this problem was solved in Ceph dashboard some time ago:

https://tracker.ceph.com/issues/36404

So they delete snapshots and then delete image.

Ceph CSI stops provisioning new volumes once ceph trash is full. k8s pvcs are just waiting in Provision state and logs are full of messages like:
```
I0930 10:06:28.594040       1 rbd_util.go:610] ID: 105522 Req-ID: 0001-0024-3bbd557f-6326-493f-a98d-3491181374a8-0000000000000003-bab0957c-3a77-11ed-a1f6-9234d9da000a rbd: delete production-island10bab0957c-3a77-11ed-a1f6-9234d9da000a-temp using mon 10.25.37.17:6789,10.25.37.18:6789,10.25.37.12:6789,10.25.37.13:6789,10.25.37.11:6789, pool csi
E0930 10:06:28.619336       1 rbd_util.go:591] ID: 105522 Req-ID: 0001-0024-3bbd557f-6326-493f-a98d-3491181374a8-0000000000000003-bab0957c-3a77-11ed-a1f6-9234d9da000a failed to list images in trash: rbd: ret=-34, Numerical result out of range
E0930 10:06:28.619402       1 utils.go:200] ID: 105522 Req-ID: 0001-0024-3bbd557f-6326-493f-a98d-3491181374a8-0000000000000003-bab0957c-3a77-11ed-a1f6-9234d9da000a GRPC error: rpc error: code = Internal desc = rbd: ret=-34, Numerical result out of range
```

ruslanloman:
Hello!

Ceph csi couldn't remove pvc with snapshots that was created directly in
ceph using rbd or other clients. As a result, the rbd image remains permanently in the trash.
As solution we can cleanup snapshots before removing rbd image.

## Related issues ##

Fixes: #3416

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -652,6 +652,13 @@ func (ri *rbdImage) deleteImage(ctx context.Context) error {
 		return err
 	}
 
+	// Make sure there are no snapshots for the image before removing, otherwise it's stuck in the trash.
+	err = ri.purgeSnapshots(ctx)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to purge image snapshots: %s, error: %v", ri, err)
+		return err
+	}
+
 	rbdImage := librbd.GetImage(ri.ioctx, image)
 	err = rbdImage.Trash(0)
 	if err != nil {
@@ -1440,6 +1447,30 @@ func (ri *rbdImage) deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot) erro
 	}
 
 	return err
+}
+
+func (ri *rbdImage) purgeSnapshots(ctx context.Context) error {
+	rbdSnap := &rbdSnapshot{}
+
+	snaps, err := ri.listSnapshots()
+	if err != nil {
+		log.DebugLog(ctx, "rbd: failed to get list of snapshots %v", err)
+	}
+
+	for _, snap := range snaps {
+		log.DebugLog(ctx, "rbd: found image snapshot %s", snap.Name)
+		rbdSnap.RbdSnapName = snap.Name
+		rbdSnap.Monitors = ri.Monitors
+
+		err = ri.deleteSnapshot(ctx, rbdSnap)
+		if err != nil {
+			log.ErrorLog(ctx, "failed to delete rbd image snapshot: %s, error: %v", ri, err)
+
+			return err
+		}
+	}
+
+	return nil
 }
 
 func (rv *rbdVolume) cloneRbdImageFromSnapshot(


internal/rbd/rbd_util.go
```diff
@@ -1442,6 +1449,30 @@ func (ri *rbdImage) deleteSnapshot(ctx context.Context, pOpts *rbdSnapshot) erro
 	return err
 }
 
+func (ri *rbdImage) purgeSnapshots(ctx context.Context) error {
+	rbdSnap := &rbdSnapshot{}
+
+	snaps, err := ri.listSnapshots()
+	if err != nil {
+		log.DebugLog(ctx, "rbd: failed to get list of snapshots %v", err)
```
nixpanic, line 1457: return the error?

nixpanic (commented):
I am not sure this is what all users want. Some users might still want to use the snapshot, even when someone/thing else deleted the volume. Having the snapshot available, and the parent rbd image in the trash makes it possible to undo unintentional deletions (for example).

Madhu-1:
> ceph using rbd or other clients. As a result, the rbd image remains permanently in the trash.

why cephcsi need to handle the user created snapshots? if user is creating the snapshot its user responsibility to delete it also https://github.com/ceph/ceph-csi/issues/3416, having this feature can also lead to similar changes for others also like Replication, cephfs subvolume deletion etc (just an example).

Rakshith-R (changes requested):
We still had discussions going on the issue thread?

>why does cephcsi need to handle the user created snapshots? if user is creating the snapshot its user responsibility to delete it also

+1
We cannot guarantee such behavior, when another external entity acts on cephcsi managed images.

We cannot purge snapshots of a pvc having k8s snapshot and clone. These preserve the link between parent and child

ruslanloman:
Hi, thank you for your comments!

With this PR i'm trying to find some solution for this issue. Because right now cephcsi couldn't remove the ceph rdb image with snapshots and we can consider it as a bug.
There are 3 solutions that i have in my mind.
1. Purge snapshots before image removing (this approach is  in PR)
2. In case we are trying to remove pvc with rbd snapshots then we just return an error and let the client clean it up manually.
3. Same as item 1 but we can add a config option to allow snapshot cleanup.

nixpanic:
> With this PR i'm trying to find some solution for this issue. Because right now cephcsi couldn't remove the ceph rdb image with snapshots and we can consider it as a bug. There are 3 solutions that i have in my mind.

I don't think this is a bug, I would probably call it _intentional behaviour_. It might be unexpected for some users, but Ceph-CSI should not, and does not delete volumes or snapshots that were created by others.

>     1. Purge snapshots before image removing (this approach is  in PR)
> 
>     2. In case we are trying to remove pvc with rbd snapshots then we just return an error and let the client clean it up manually.
> 
>     3. Same as item 1 but we can add a config option to allow snapshot cleanup.

I think option 2 is the most reasonable. Ceph-CSI should not (never) be able to get blamed for data-loss, and wrongly configuring a "force snapshot removal" option invites that (even if it is an admin/user error).

Option 2 makes sure that the user/admin notices their unbalanced usage of Ceph-CSI and other tools. The user/admin should then decide to correct the cooperation of the tools/procedures.

ruslanloman:
Hi @nixpanic I got your point. I'm going to close this PR as invalid.

PR #3362, opened 2022-09-02
e2e: adjust e2e test to run on kubernetes 1.25

Madhu-1:
removed podsecurity feature-gate and adjusted e2e tests to use privileged pod security for the e2e namespace

Depends-On: #3365

PR #3471, opened 2022-10-26
ci: reduce comments in ci action

Madhu-1:
We dont need to run individual tests to start the Jenkins jobs, We already have a regex to start similar tests are different kubernetes versions.

See https://github.com/ceph/ceph-csi/pull/3362#issuecomment-1291597227 for test result

### DIFF
--- a/.github/workflows/pull-request-commentor.yaml
+++ b/.github/workflows/pull-request-commentor.yaml
@@ -15,79 +15,30 @@ jobs:
     permissions:
       pull-requests: write
     steps:
-      - name: Add comment to trigger external storage tests for Kubernetes 1.22
+      - name: Add comment to trigger external storage tests for Kubernetes
         uses: peter-evans/create-or-update-comment@v2
         with:
           issue-number: ${{ github.event.pull_request.number }}
           body: |
-            /test ci/centos/k8s-e2e-external-storage/1.22
+            /test ci/centos/k8s-e2e-external-storage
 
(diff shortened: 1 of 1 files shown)


Madhu-1:
Note here is this will also start all extra jobs which are not required for CI. we can decide Do we need it or we can close it.

Madhu-1:
> Note here is this will also start all extra jobs which are not required for CI. we can decide Do we need it or we can close it.

This is also similar to `/test all` :dagger:

.github/workflows/pull-request-commentor.yaml
```diff
@@ -15,79 +15,30 @@ jobs:
...
       pull-requests: write
     steps:
-      - name: Add comment to trigger external storage tests for Kubernetes 1.22
+      - name: Add comment to trigger external storage tests for Kubernetes
         uses: peter-evans/create-or-update-comment@v2
         with:
           issue-number: ${{ github.event.pull_request.number }}
           body: |
-            /test ci/centos/k8s-e2e-external-storage/1.22
+            /test ci/centos/k8s-e2e-external-storage
```
nixpanic, line 23: this starts all versions of the tests, also the ones that are not required. it'll cause more load on the CI, and that is what we try to reduce.

nixpanic (changes requested):

Madhu-1:
Closing based on https://github.com/ceph/ceph-csi/pull/3471#discussion_r1005491158

PR #3478, opened 2022-10-28
rbd: return error if last sync time not present

Madhu-1:
As per the csiaddon spec, last sync time is a required parameter in the GetVolumeReplicationInfo if we are failed to parse the description, we return an error message instead of nil which is an empty response.

Note:- skipping E2E as we dont have CI for DR

### DIFF
--- a/internal/rbd/replicationcontrollerserver.go
+++ b/internal/rbd/replicationcontrollerserver.go
@@ -804,13 +804,14 @@ func getLastSyncTime(description string) (*timestamppb.Timestamp, error) {
 	// description = "replaying,{"bytes_per_second":0.0,
 	// "bytes_per_snapshot":149504.0,"local_snapshot_timestamp":1662655501
 	// ,"remote_snapshot_timestamp":1662655501}"
-	// In case there is no local snapshot timestamp we can pass the default value
+	// In case there is no local snapshot timestamp return an error as the
+	// LastSyncTime is required.
 	if description == "" {
-		return nil, nil
+		return nil, errors.New("empty description")
 	}
 	splittedString := strings.SplitN(description, ",", 2)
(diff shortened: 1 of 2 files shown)


nixpanic (approved):

humblec (approved):

Madhu-1:
Added DNM for now, as @yati1998 needs to check a few things.

Madhu-1:
Thinking More about it, keeping the LastSyncTime as optional makes more sense than a generic CSI Driver as well. Update the spec to make it optional.

Madhu-1:
https://github.com/csi-addons/spec/pull/47 and https://github.com/csi-addons/kubernetes-csi-addons/pull/260 are alternative for this one, @yati1998 PTAL

ShyamsundarR:
> Thinking More about it, keeping the LastSyncTime as optional makes more sense than a generic CSI Driver as well. Update the spec to make it optional.

The intention should not be to error (unless there is actually an error). For example if remote status is missing, then determining the lastSyncTime is not possible in this reconcile, and would be reported as epoch (or 0).

If we error out because the volume is not synced yet or the remote is not reporting a sync time (and report the errors on the resource), cases where remote is down or due to failover is still Primary etc. will result in VolRep resource potentially reporting errors and not a healthy primary state and cause issues.

Overall, I think we expect a lastSyncTime, and if we are unable to determine the same it is epoch OR prior lastSyncTime value for the resource, iff it is from the same generation as the current CR (which is applicable at the VR layer and not at the gRPC).

IOW, what does the caller (VolRep reconciler in this case) do when this errors out is critical IMHO.

Madhu-1:
Closing this one as per discussion and https://github.com/csi-addons/kubernetes-csi-addons/pull/260 solves the problem

PR #3519, opened 2022-11-11
rebase: make use of latest go-ceph library ( v1.18.0)

humblec:
- [x]     rebase: go-ceph version has been updated to v1.18.0
- [x]      docs: update/correct development guide

### DIFF
--- a/docs/development-guide.md
+++ b/docs/development-guide.md
@@ -46,7 +46,7 @@ it is **highly** encouraged to:
 * Set up a pre-commit hook to catch issues locally.
 
    ```console
-   pip install pre-commit==2.5.1
+   pip install pre-commit
    pre-commit install
    ```
 
@@ -338,7 +338,7 @@ fresh PRs, rebase of PRs and force pushing changes to existing PRs.
 In case of failed we already documented steps  to manually
 [retrigger](#retriggering-the-ci-jobs) the CI jobs. Sometime the tests might be
 flaky which required manually retriggering always. We have newly added a github
-action which runs periodically to retest the failed PR's. Below are the cretia
+action which runs periodically to retest the failed PR's. Below are the criteria
 for auto retesting the failed PR.
 
 * Analyze the logs and make sure its a flaky test.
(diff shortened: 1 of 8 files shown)


Madhu-1 (commented):
Changes LGTM, please keep dependent changes in one commit, one commit for makefile and go changes?

humblec:
> Changes LGTM, please keep dependent changes in one commit, one commit for makefile and go changes?



> Changes LGTM, please keep dependent changes in one commit, one commit for makefile and go changes?

I kept the preview tag removal out of this PR. it looks like some where in go-ceph it still uses. will look into that in detail and act accordingly.  @Madhu-1 ptal.. thanks

Madhu-1 (dismissed):

docs/development-guide.md
```diff
@@ -46,7 +46,7 @@ it is **highly** encouraged to:
 * Set up a pre-commit hook to catch issues locally.
 
    ```console
-   pip install pre-commit==2.5.1
+   pip install pre-commit==2.20.0
```
nixpanic: does it need to be versioned, why not use the latest?

vendor/github.com/ceph/go-ceph/common/admin/nfs/admin.go
```diff
@@ -1,5 +1,5 @@
-//go:build !(nautilus || octopus) && ceph_preview
-// +build !nautilus,!octopus,ceph_preview
+//go:build !(nautilus || octopus)
+// +build !nautilus,!octopus
```
nixpanic, line 2: Are there still things for which we depend on `ceph_preview`? If not, remove it from the build-tags.

humblec: @nixpanic this has to be removed from go-ceph vendor and not here ?

docs/development-guide.md
humblec: need not be, I put it without the version in that case.

vendor/github.com/ceph/go-ceph/common/admin/nfs/admin.go
nixpanic: I think we add the `ceph_preview` build-tag in our scripts. If we don't do that anymore, that's fine too.

humblec: @nixpanic as mentioned here https://github.com/ceph/ceph-csi/pull/3519#issuecomment-1311491189 , the preview thing has to be checked and afacit, removing it cause the ci to fail. I will take this up as followup thing as mentioned.

Madhu-1 (approved):

yati1998 (approved):

PR #3526, opened 2022-11-12
rebase: fix CVEs in the image

humblec:
This commit update dependencies which is required to fix below CVEs.
    
    CVE-2022-27664
    CVE-2022-27191


humblec:
@Madhu-1 this PR should fix most of the ceph csi image specific vulnarabilities reported in the scan. Can you take a look at this in priority?

nixpanic (changes requested):
Most of these are simple rebases, they should not be combined in the same PR with fixes for CVE's. For a CVE fix, you'll need to state the CVE numbers as well.

api/go.mod
```diff
@@ -1,10 +1,32 @@
 module github.com/ceph/ceph-csi/api
 
-go 1.16
+go 1.18
```
nixpanic: This requires consumers of the API to use Go 1.18. Ideally we provide the API for the lpwest, still supported Go version?

humblec: no.., 1.16 and 1.17 are unsupported or already hit EOL.

humblec:
> Most of these are simple rebases, they should not be combined in the same PR with fixes for CVE's. For a CVE fix, you'll need to state the CVE numbers as well.

@nixpanic I have mentioned which CVEs and kept only that change in the PR. ptal.

humblec:
> Most of these are simple rebases, they should not be combined in the same PR with fixes for CVE's. For a CVE fix, you'll need to state the CVE numbers as well.

Indeed these are different areas and small. But I have already put into seperate commits eventhough its a single PR. More or less, I didnt see value running different CI tests and wasting resources for each of this change , so combined into one PR.  Regardless I am splitting this to different PRs now.

nixpanic:
Why the urgency for these?

CVE-2022-27664: http server issue after `GOAWAY`, can that happen when metrics are enabled?
CVE-2022-27191: ssh server issue... Ceph-CSI does not run a ssh-server

humblec:
> Why the urgency for these?
> 
> [CVE-2022-27664](https://github.com/advisories/GHSA-69cg-p879-7622): http server issue after `GOAWAY`, can that happen when metrics are enabled? [CVE-2022-27191](https://github.com/advisories/GHSA-8c26-wmh5-6g9v): ssh server issue... Ceph-CSI does not run a ssh-server

@nixpanic all the security reports against Ceph CSI report this vulnerability.  Unfortunately even security reports run on projects which consume Ceph CSI image also list down these vulnerabilities and this has been reported by those maintainers too. One example here is CNCF project Rook.

nixpanic:
Sure, lots of security scanners report issues with container images because vulnerable code is included. But if the code is not used, there is no urgency to update it. This isn't only for Go packages, but also for the RPMs that come with the OS base container image.

Explaining why a CVE needs to be fixed is important for prioritization, and shows that reported CVEs are taking seriously. A quick glance does not give me a hint that this can be exploited, so for me this has the same priority as a normal rebase of a dependency.

nixpanic (approved):

Rakshith-R (approved):

nixpanic:
@mergifyio rebase

PR #3529, opened 2022-11-14
build: update packages in api directory to the latest

humblec:
this commit update the packages in API directory and also update the mention of go version in the same. The go version 1.16 and 1.17 have already hit EOL.


api/go.mod
```diff
@@ -1,10 +1,32 @@
 module github.com/ceph/ceph-csi/api
 
-go 1.16
+go 1.18
```
nixpanic, line 3: I don't think we want to require Go 1.18 for projects that import this api. Please use an older supported version of Go.

nixpanic (changes requested):

api/go.mod
humblec: @nixpanic as mentioned in PR description , 1.16 and 1.17 are EOL . which version you are referring by "older supported version of Go" ?

nixpanic: I had to search where to find those details. In future, include the link to the EOL notice for a version please, that makes things much easier to verify.

nixpanic (approved):

Rakshith-R (approved):

humblec:
@Mergifyio rebase

nixpanic:
@humblec why was the `ok-to-test` label removed and re-added? This caused double the CI jobs, which will now cause additional timeouts and requires more retries later :sob:

humblec:
> @humblec why was the `ok-to-test` label removed and re-added? This caused double the CI jobs, which will now cause additional timeouts and requires more retries later sob

@nixpanic  noticed 6 jobs atleast stuck.. so removed the `ok-to-test` , may be I would have manually retriggered it. :(

nixpanic:
> @nixpanic noticed 6 jobs atleast stuck.. so removed the `ok-to-test` , may be I would have manually retriggered it. :(

Please define **stuck**.

Depending on the way it is "stuck" a manual abort in Jenkins can free-up a worker node, or prevent one from being allocated for a test where the results already have been cancelled.

Issue #3545, opened 2022-11-21
Few E2E tests are getting skipped in devel branch


Madhu-1:
> @Madhu-1 this keeps failing, maybe there is an issue with it?

I see a problem with tests in the devel branch here https://github.com/ceph/ceph-csi/blob/0f0957164ec882a91d11bf1c880e1c0777abfb01/e2e/rbd.go#L4592-L4610. You can see the ceph users are deleted. Now we are trying to run one more test, `pvcDeleteWhenPoolNotFound` which creates the PVC and deletes the PVC as the user is deleted. The test should never work. I feel somewhere something went wrong in the devel branch. A bunch of tests are getting skipped, and we are not running all the tests for sure. Because of some missing patch in the release-3.7 branch, this is getting caught now, I fixed this in the release-3.7 branch. We need to check why the full test suite is not running/why tests are getting skipped.

_Originally posted by @Madhu-1 in https://github.com/ceph/ceph-csi/issues/3531#issuecomment-1321884300_

PR #3555, opened 2022-11-28
e2e: return return from By() in tests

Madhu-1:
Return from an individual test with enclosed By() does not work as expected. For that reason, we need to use if/else statements to skip the test or run the tests.

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -923,8 +923,6 @@ var _ = Describe("RBD", func() {
 				if err != nil {
 					if rwopMayFail(err) {
 						e2elog.Logf("RWOP is not supported: %v", err)
-
-						return
 					}
 					e2elog.Failf("failed to create PVC: %v", err)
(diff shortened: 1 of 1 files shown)


nixpanic:
> Return from an individual test with enclosed By() does not work as expected. For that reason, we need to use if/else statements to skip the test or run the tests.

@Madhu-1 do you know if this is documented somewhere? I don't understand how an early return is different from an implicit return at the end of a function.

Madhu-1:
> > Return from an individual test with enclosed By() does not work as expected. For that reason, we need to use if/else statements to skip the test or run the tests.
> 
> @Madhu-1 do you know if this is documented somewhere? I don't understand how an early return is different from an implicit return at the end of a function.

@nixpanic am testing this out https://github.com/ceph/ceph-csi/issues/3545. Busy with something else, I need to debug this and understand the cause of skipping tests.

PR #3570, opened 2022-12-08
ci: use new `duffy` command for CentOS CI

nixpanic:
The `cico` command is getting deprecated, and the new `duffy` command should be used instead. The new command requires a `~/.config/duffy` file that contains an API endpoint and credentials.

See-also: https://sigs.centos.org/guide/ci/#installing-and-configuring-duffy-client

---

<details>

  progress or results

</details>

### DIFF
--- a/commitlint.groovy
+++ b/commitlint.groovy
@@ -1,9 +1,22 @@
 def cico_retries = 16
 def cico_retry_interval = 60
+def duffy_pool = 'virt-ec2-t2-centos-8s-x86_64'
 def ci_git_repo = 'https://github.com/ceph/ceph-csi'
 def ci_git_branch = 'ci/centos'
 def ref = "devel"
 def git_since = 'origin/devel'
+
+def create_duffy_config() {
+	writeFile(
+		file: '/home/jenkins/.config/duffy',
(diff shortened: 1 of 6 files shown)


nixpanic:
A manually started testing job can be found here: https://jenkins-ceph-csi.apps.ocp.ci.centos.org/job/ndevos-testing-new-duffy/36/console

humblec (approved):

commitlint.groovy
```diff
@@ -1,10 +1,23 @@
 def cico_retries = 16
 def cico_retry_interval = 60
+def duffy_pool = 'virt-ec2-t2-centos-8s-x86_64'
```
humblec, line 3: @nixpanic from where we get this value options like `virt-ec2-*` ?

Madhu-1 (approved):

commitlint.groovy
nixpanic: That is the pool the legacy `cico` command got the systems from. I found this pool in the output of the `duffy list-sessions` command when some of our CI jobs were running. There are many more pools, and they are probably documented somewhere, but this was easier/quicker.

PR #3572, opened 2022-12-12
[wip]rebase: adopt logging framework change and update to kube 1.26

humblec:
- [x]  This commit update the dependency to 1.26.0 version along with
    other client modules.
- [x] log framework package is part of core framework with this kube version update.


Rakshith-R:
Was the first commit required to get second commit to work ? 

The reason for the first commit of e2elog to framework.Log change is not clear to me. 
:+1:  I'll take a look again once it is out of [wip]

Issue #2595, opened 2021-10-26
Update Static check tool to latest version


yati1998:
The current version for golangci-lint in ceph-csi is v1.39.0, but we have new changes introduced in the latest releases of golangci-lint causing some static errors.

```
make go-lint
cephcsi image settings: quay.io/cephcsi/cephcsi version canary
sed "s/@@CEPH_VERSION@@/octopus/g" < scripts/golangci.yml.in > scripts/golangci.yml
./scripts/lint-go.sh
INFO [config_reader] Used config file scripts/golangci.yml 
INFO [lintersdb] Active 62 linters: [asciicheck bodyclose cyclop deadcode depguard dogsled durationcheck errcheck errorlint exhaustive exportloopref forcetypeassert gocognit goconst gocritic gocyclo godot gofmt gofumpt goheader goimports golint gomnd gomodguard goprintffuncname gosec gosimple govet ifshort importas ineffassign interfacer lll makezero maligned misspell nakedret nestif nilerr nlreturn noctx nolintlint paralleltest predeclared promlinter revive rowserrcheck scopelint sqlclosecheck staticcheck structcheck stylecheck tagliatelle thelper tparallel typecheck unconvert unparam unused varcheck wastedassign whitespace] 
INFO [loader] Using build tags: [octopus]         
INFO [loader] Go packages loading at mode 575 (name|types_sizes|compiled_files|files|imports|deps|exports_file) took 40.304118615s 
WARN [runner] The linter 'scopelint' is deprecated (since v1.39.0) due to: The repository of the linter has been deprecated by the owner.  Replaced by exportloopref. 
WARN [runner] The linter 'interfacer' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner.  
WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive. 
WARN [runner] The linter 'maligned' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner.  Replaced by govet 'fieldalignment'. 
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 15.673235ms 
INFO [linters context] importas settings found, but no aliases listed. List aliases under alias: key. 
[...]

PR #3580, opened 2022-12-13
cleanup: fix ireturn linter

riya-singhal31:
This PR fix the ireturn linter errors caused by golangci-lint 1.43.

For reference- https://github.com/ceph/ceph-csi/issues/2595

Signed-off-by: riya-singhal31 <rsinghal@redhat.com>

### DIFF
--- a/internal/cephfs/core/filesystem.go
+++ b/internal/cephfs/core/filesystem.go
@@ -42,7 +42,7 @@ type fileSystem struct {
 }
 
 // NewFileSystem returns a new instance of fileSystem.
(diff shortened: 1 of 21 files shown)


nixpanic:
Can this be fixed in a different way? Adding the `nolint:ireturn` annotation just skips the check for those functions...

Rakshith-R (changes requested):
Please add description of linter in commit message. 
Explanation of how this commit fixes the linter errors
and why this fix was chosen.

PR #3617, opened 2023-01-23
rbd: modify `ensureImageCleanup()` to be best effort

### LINKED ISSUE #3593: parent volumes are allowed to be removed from pv and pvc despite existing snapshots
# Describe the bug #

The parent pvc can be deleted without issues with existing snapshots from the kubernetes cluster. It will disappear from the pvc and pv list, get moved to the trash on the ceph and the ceph will keep trying to remove it forever but fails to be removed due to existing snapshots.

# Environment details #

- Image/version of Ceph CSI driver : canary
- Helm chart version : canary
- Kernel version : 5.15.0-52-generic
  `krbd` or `rbd-nbd`) : krbd
- Kubernetes cluster version : 1.23.12
- Ceph cluster version : 17.2.5

# Steps to reproduce #

1. Create a pvc on the kubernetes. Create snapshots from them, in my case it's 10 snapshots.
2. Delete the parent volume from the kubernetes with kubectl delete pvc which in my case will remove the pv as well.
3. The parent volume will disappear from the pvc and pv list also it will be moved to the trash where it will be attempted to be deleted
4. The ceph manager will report the following error:
`debug 2023-01-05T10:56:42.406+0000 7f59ddf24700  0 [rbd_support INFO root] execute_trash_remove: task={"sequence": 803528, "id": "71caadb1-a589-4c48-a162-153ce9f1d1ed", "message": "Removing image production-island11-overcloud/3127fdcbed096c from trash", "refs": {"action": "trash remove", "pool_name": "production-island11-overcloud", "pool_namespace": "", "image_id": "3127fdcbed096c"}, "retry_message": "[errno 39] RBD image has snapshots (error deleting image from trash)", "retry_attempts": 856, "retry_time": "2023-01-05T10:56:42.204701", "in_progress": true, "progress": 0.0}
`

# Expected behavior #

The expected behaviour is to not allow the parent to get deleted until all the snapshots were cleaned up and not to put it to the trash, because ceph will attempt to keep trying to delete and fails due to existing snapshots.

# Logs #

If the issue is in PVC creation, deletion, cloning please attach complete logs
of below containers.

- csi-provisioner and csi-rbdplugin/csi-cephfsplugin container logs from the
  provisioner pod.
`debug 2023-01-05T10:56:42.406+0000 7f59ddf24700  0 [rbd_support ERROR root] execute_task: [errno 39] RBD image has snapshots (error deleting image from trash)
debug 2023-01-05T10:56:42.406+0000 7f59ddf24700  0 [rbd_support INFO root] execute_task: task={"sequence": 803528, "id": "71caadb1-a589-4c48-a162-153ce9f1d1ed", "message": "Removing image production-island11-overcloud/3127fdcbed096c from trash", "refs": {"action": "trash remove", "pool_name": "production-island11-overcloud", "pool_namespace": "", "image_id": "3127fdcbed096c"}, "retry_message": "[errno 39] RBD image has snapshots (error deleting image from trash)", "retry_attempts": 856, "retry_time": "2023-01-05T10:56:42.204701"}
debug 2023-01-05T10:56:42.406+0000 7f59ddf24700  0 [rbd_support INFO root] execute_trash_remove: task={"sequence": 803528, "id": "71caadb1-a589-4c48-a162-153ce9f1d1ed", "message": "Removing image production-island11-overcloud/3127fdcbed096c from trash", "refs": {"action": "trash remove", "pool_name": "production-island11-overcloud", "pool_namespace": "", "image_id": "3127fdcbed096c"}, "retry_message": "[errno 39] RBD image has snapshots (error deleting image from trash)", "retry_attempts": 856, "retry_time": "2023-01-05T10:56:42.204701", "in_progress": true, "progress": 0.0}
`

Thanks in advance for looking into this!

Rakshith-R:
The function `ensureImageCleanup()`, called when images to be deleted are not found, ensures that these images if in trash have a task on them to be removed from trash.

This scenario is an extreme edge case and it is an expensive operation to list images in trash depending on the number of images in trash and , therefore this commit modifies code to only log the errors rather than returning it which causes a retry of the whole request.

Closes: #3593

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -842,7 +842,8 @@ func (cs *ControllerServer) checkErrAndUndoReserve(
 	if errors.Is(err, ErrImageNotFound) {
 		err = rbdVol.ensureImageCleanup(ctx)
 		if err != nil {
-			return nil, status.Error(codes.Internal, err.Error())
+			log.ErrorLog(ctx, "failed to ensure image %q is cleaned up",
+				rbdVol.RbdImageName, err)
 		}
 	} else {
 		// All errors other than ErrImageNotFound should return an error back to the caller
@@ -1002,7 +1003,8 @@ func cleanupRBDImage(ctx context.Context,
 		if errors.Is(err, ErrImageNotFound) {
 			err = tempClone.ensureImageCleanup(ctx)
 			if err != nil {
-				return nil, status.Error(codes.Internal, err.Error())
+				log.ErrorLog(ctx, "failed to ensure image %q is cleaned up",
+					tempClone.RbdImageName, err)
 			}
 		} else {
 			// return error if it is not ErrImageNotFound
@@ -1498,9 +1500,8 @@ func cleanUpImageAndSnapReservation(ctx context.Context, rbdSnap *rbdSnapshot, c
 	// cleanup the image from trash if the error is image not found.
 	err = rbdVol.ensureImageCleanup(ctx)
 	if err != nil {
-		log.ErrorLog(ctx, "failed to delete rbd image: %q with error: %v", rbdVol.Pool, rbdVol.VolName, err)
-
-		return status.Error(codes.Internal, err.Error())
+		log.ErrorLog(ctx, "failed to ensure image %q is cleaned up",
+			rbdVol.RbdImageName, err)
 	}
 	err = undoSnapReservation(ctx, rbdSnap, cr)
 	if err != nil {


Madhu-1:
> The function `ensureImageCleanup()`, called when images to be deleted are not found, ensures that these images if in trash have a task on them to be removed from trash.
> 
> This scenario is an extreme edge case and it is an expensive operation to list images in trash depending on the number of images in trash and , therefore this commit modifies code to only log the errors rather than returning it which causes a retry of the whole request.
> 
> Closes: #3593
> 
> Signed-off-by: Rakshith R [rar@redhat.com](mailto:rar@redhat.com)

@Rakshith-R I agree this is a best effort, but this can leave the stale resources in the cluster what is the measure to remove the stale resources?
There was a bug in the go-ceph code which was using less value for the list operation, it's been a workaround with the larger value. IMO the actual fix should come from the go-ceph, not at the cephcsi level.

Rakshith-R:
> @Rakshith-R I agree this is a best effort, but this can leave the stale resources in the cluster what is the measure to remove the stale resources? There was a bug in the go-ceph code which was using less value for the list operation, it's been a workaround with the larger value. IMO the actual fix should come from the go-ceph, not at the cephcsi level.

IMO, I see a better trade off in  very rarely (this is an extreme edge case) leaving behind a stale resource **in trash** rather than blocking all delete requests &/ repeating such expensive operation (listing large number of images in trash) which also ends up slowing down other operations too(create,clone,snapshot, etc.,). 

let me know what you think,

nixpanic (commented):
Not sure about this... If Ceph Mgr is unreachable for a brief time, could this problem not also happen? Without returning an error to the caller, there will not be a retry (by Kubernetes) to delete the image.

Rakshith-R:
:+1: 
We'll have to find another way then, closing this pr

PR #3624, opened 2023-01-27
rbd: discover if StagingTargetPath in NodeExpandVolume exists

### LINKED ISSUE #3623: staging_path is optional parameter in the NodeExpandVolumeRequest RPC
Cephcsi rbd NodePlugin expects the staging_path in the NodeExpandVolume RPC call or else it will fail to find the device information [1]. As it's an optional parameter  [2] we cannot assume it will be sent in all the Requests, cephcsi should be able to work without optional parameters. 

[1] https://github.com/ceph/ceph-csi/blob/f852873e16af6d186c56bf83e4aabfc4a9272bfe/internal/rbd/nodeserver.go#L1039-L1072

[2] https://github.com/container-storage-interface/spec/blob/bd1cb6579c9ed791a6d9980fa3c0ef11600cc273/csi.proto#L1604-L1613

Madhu-1:
The StagingTargetPath is an optional entry in NodeExpandVolumeRequest, We cannot expect it to be permanently set, and at the same time, cephcsi depended on the StaingTargetPath to retrieve some metadata
information.

This commit will check all the mount ref and identifies the stagingTargetPath by reviewing the image-meta.json file that exists. This is a costly operation as we need to loop through all the mounts and check image-meta.json in each mount, but this happens only if the StaingTargetPath needs to be set in the NodeExpandVolumeRequest.

fixes #3623

Note :- gofmt tool I am using also fixed the indentation problem in the comment section in nodeserver.go

### DIFF
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -269,11 +269,11 @@ func populateRbdVol(
 // Implementation notes:
 // - stagingTargetPath is the directory passed in the request where the volume needs to be staged
 //   - We stage the volume into a directory, named after the VolumeID inside stagingTargetPath if
-//    it is a file system
+//     it is a file system
 //   - We stage the volume into a file, named after the VolumeID inside stagingTargetPath if it is
-//    a block volume
-// - Order of operation execution: (useful for defer stacking and when Unstaging to ensure steps
-//	are done in reverse, this is done in undoStagingTransaction)
+//     a block volume
+//   - Order of operation execution: (useful for defer stacking and when Unstaging to ensure steps
+//     are done in reverse, this is done in undoStagingTransaction)
 //   - Stash image metadata under staging path
 //   - Map the image (creates a device)
 //   - Create the staging file/directory under staging path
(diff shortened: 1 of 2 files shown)


internal/rbd/nodeserver.go
```diff
@@ -1026,6 +1026,31 @@ func (ns *NodeServer) NodeUnstageVolume(
 	return &csi.NodeUnstageVolumeResponse{}, nil
 }
 
+// getStagingPath returns the staging path for the volume from the volume path.
+/*
```
nixpanic: mixing of comment styles is a little odd.

internal/rbd/nodeserver.go
```diff
@@ -1042,8 +1067,18 @@ func (ns *NodeServer) NodeExpandVolume(
...
+		// stagingTargetPath is optional parameter in NodeExpandVolumeRequest request
+		// Kubernetes will not send the staging_target_path in the volume_path in some cases
+		// Refer https://github.com/kubernetes/kubernetes/issues/115343
+		var err error
+		volumePath, err = ns.getStagingPath(volumePath)
+		if err != nil {
+			// If stagingTargetPath is not found in volumePath then use
+			// volumePath
+			volumePath = req.GetVolumePath()
+			log.ErrorLog(ctx, "failed to find stagingTargetPath from volumePath %v: %v", volumePath, err)
```
nixpanic: This isn't always an error, and there is an alternative value used... Maybe log as info?

nixpanic (changes requested):
just some nits, nothing serious

nixpanic (dismissed):

internal/rbd/nodeserver.go
```diff
@@ -1042,8 +1065,18 @@ func (ns *NodeServer) NodeExpandVolume(
...
+		// Kubernetes will not send the staging_target_path in the volume_path in some cases
+		// Refer https://github.com/kubernetes/kubernetes/issues/115343
+		var err error
+		volumePath, err = ns.getStagingPath(volumePath)
+		if err != nil {
+			// If stagingTargetPath is not found in volumePath then use
+			// volumePath
+			volumePath = req.GetVolumePath()
+			log.UsefulLog(ctx, "failed to find stagingTargetPath from volumePath %v: %v", volumePath, err)
+		}
```
Rakshith-R: @Madhu-1 
Did you intend the code to the as follows ? 

Otherwise the `volumePath` being sent to `getStagingPath()` is empty from the check on line 1066

```suggestion
		var err error
		volumePath, err = ns.getStagingPath(req.GetVolumePath())
		if err != nil {
			// If stagingTargetPath is not found in volumePath then use
			// volumePath
			volumePath = req.GetVolumePath()
			log.UsefulLog(ctx, "failed to find stagingTargetPath from volumePath %v: %v", volumePath, err)
		}
```

Rakshith-R (changes requested):

internal/rbd/nodeserver.go
Madhu-1: Yes that correct, fixed now.

Rakshith-R (dismissed):

nixpanic (dismissed):

nixpanic:
@mergifyio rebase

Madhu-1:
@mergifyio requeue

Madhu-1:
Mergify dropped the approvals after rebase :(

Rakshith-R (approved):

nixpanic (approved):

nixpanic:
> Mergify dropped the approvals after rebase :(

I've seen that happen somewhere else too. But mostly the approvals are kept.

Issue #3336, opened 2022-08-23
Release 3.8.0 tracker


humblec:
# Describe the bug #

This issue tracks release v3.8 of Ceph CSI driver 
The milestone is available here https://github.com/ceph/ceph-csi/milestone/15

- [ ] Code Freeze : 10 Feb 2023
- [ ] GA : 15 Feb 2023

Features:
- [x] #3310 
- [x] read_from_replica support for better performance @Rakshith-R 
- [x] https://github.com/ceph/ceph-csi/issues/757 
- [x] https://github.com/ceph/ceph-csi/issues/3603
- [x] https://github.com/ceph/ceph-csi/issues/3616
- [x] https://github.com/ceph/ceph-csi/issues/2637 
- [x] https://github.com/ceph/ceph-csi/pull/3639

humblec:
@ceph/ceph-csi-contributors any features we would like to track it for this release ? If yes, please list out

yati1998:
https://github.com/ceph/ceph-csi/issues/3603 ?

humblec:
> #3603 ?

Yeah , thanks ! added

Rakshith-R:
Release notes regarding read affinity and 
removal of option to run cephcsi being run as both controller and node servers when either of the option is provided.
https://github.com/ceph/ceph-csi/pull/3639#discussion_r1095685719

humblec:
> Release notes regarding read affinity and removal of option to run cephcsi being run as both controller and node servers when either of the option is provided. [#3639 (comment)](https://github.com/ceph/ceph-csi/pull/3639#discussion_r1095685719)

Sure.. added on the tracker

humblec:
@riya-singhal31 are we targetting shallow volumes for 3.8 ?

riya-singhal31:
> @riya-singhal31 are we targetting shallow volumes for 3.8 ?

Yes

Madhu-1:
Closing as the 3.8 release is out, and @Rakshith-R will take care of creating new milestone, and issue trackers for 3.9

PR #3639, opened 2023-02-02
rbd: add capability to automatically enable read affinity 

Rakshith-R:

- util: add new cmdline args to enable read affinity

This commit adds new cmdline args `enable-read-affinity`
and `crush-location-labels` to facilitate new feature of
using crush location labels from node labels to enable
read affinity using ceph's `read_from_replica=localize`
feature. This feature redirect reads to the nearest OSD.
This will be currently used only during rbd map cmd.

- rbd: add capability to automatically enable read affinity

This commit makes use of crush location labels from node
labels to supply `crush_location` and `read_from_replica=localize`
options during rbd map cmd. Using these options, ceph
will be able to redirect reads to the closest OSD,
improving performance.

- docs: add documentation regarding read affinity

This commit adds documentation about read affinity supported
for rbd volumes.

### DIFF
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -79,8 +79,15 @@ func init() {
 		&conf.DomainLabels,
 		"domainlabels",
 		"",
-		"list of kubernetes node labels, that determines the topology"+
+		"list of Kubernetes node labels, that determines the topology"+
 			" domain the node belongs to, separated by ','")
+	flag.BoolVar(&conf.EnableReadAffinity, "enable-read-affinity", false, "enable read affinity")
+	flag.StringVar(
+		&conf.CrushLocationLabels,
+		"crush-location-labels",
+		"",
+		"list of Kubernetes node labels, that determines the"+
+			" CRUSH location the node belongs to, separated by ','")
 
 	// cephfs related flags
 	flag.BoolVar(
--- a/deploy/rbd/kubernetes/csi-rbdplugin.yaml
+++ b/deploy/rbd/kubernetes/csi-rbdplugin.yaml
@@ -68,6 +68,15 @@ spec:
             # and pass the label names below, for CSI to consume and advertise
             # its equivalent topology domain
             # - "--domainlabels=failure-domain/region,failure-domain/zone"
+            #
+            # Options to enable read affinity.
+            # If enabled Ceph CSI will fetch labels from kubernetes node and
+            # pass `read_from_replica=localize,crush_location=type:value` during
+            # rbd map command. refer:
+            # https://docs.ceph.com/en/latest/man/8/rbd/#kernel-rbd-krbd-options
+            # for more details.
+            # - "--enable-read-affinity=true"
+            # - "--crush-location-labels=topology.io/zone,topology.io/rack"
           env:
             - name: POD_IP
               valueFrom:
--- a/docs/deploy-rbd.md
+++ b/docs/deploy-rbd.md
@@ -48,6 +48,8 @@ make image-cephcsi
 | `--skipforceflatten`     | `false`                       | skip image flattening on kernel < 5.2 which support mapping of rbd images which has the deep-flatten feature                                                                                                                                                                         |
 | `--maxsnapshotsonimage`  | `450`                         | Maximum number of snapshots allowed on rbd image without flattening                                                                                                                                                                                                                  |
 | `--setmetadata`          | `false`                       | Set metadata on volume                                                                                                                                                                                                                                                               |
+| `--enable-read-affinity` | `false`                       | enable read affinity                                                                                                                                                                                                                                                                 |
+| `--crush-location-labels`| _empty_                       | Kubernetes node labels that determine the CRUSH location the node belongs to, separated by ','                                                                                                                                                                                       |
 
 **Available volume parameters:**
 
@@ -195,6 +197,29 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
 **Deploy Helm Chart:**
 
 [See the Helm chart readme for installation instructions.](../charts/ceph-csi-rbd/README.md)
+
+## Read Affinity using crush locations for RBD volumes
+
+Ceph CSI supports mapping RBD volumes with krbd options
+`"read_from_replica=localize,crush_location=type1:value1|type2:value2"` to
+allow serving reads from the most local OSD (according to OSD locations as
+defined in the CRUSH map).
+Refer [krbd-options](https://docs.ceph.com/en/latest/man/8/rbd/#kernel-rbd-krbd-options)
+for more details.
+
+This can be enabled by adding labels to Kubernetes nodes like
+`"topology.io/region=east"` and `"topology.io/zone=east-zone1"` and
+passing command line arguments `"--enable-read-affinity=true"` and
+`"--crush-location-labels=topology.io/zone,topology.io/region"` to Ceph CSI
+RBD daemonset pod "csi-rbdplugin" container, resulting in Ceph CSI adding
+`"--options read_from_replica=localize,crush_location=zone:east-zone1|region:east"`
+krbd options during rbd map operation.
+If enabled, this option will be added to all RBD volumes mapped by Ceph CSI.
+Well known labels can be found
+[here](https://kubernetes.io/docs/reference/labels-annotations-taints/).
+
+>Note: Label values will have all its dots `"."` normalized with dashes `"-"`
+in order for it to work with ceph CRUSH map.
 
 ## Encryption for RBD volumes
 
--- a/internal/rbd/driver/driver.go
+++ b/internal/rbd/driver/driver.go
@@ -71,11 +71,19 @@ func NewReplicationServer(c *rbd.ControllerServer) *rbd.ReplicationServer {
 }
 
 // NewNodeServer initialize a node server for rbd CSI driver.
-func NewNodeServer(d *csicommon.CSIDriver, t string, topology map[string]string) (*rbd.NodeServer, error) {
-	return &rbd.NodeServer{
+func NewNodeServer(
+	d *csicommon.CSIDriver,
+	t string,
+	topology map[string]string,
+	crushLocationMap map[string]string,
+) (*rbd.NodeServer, error) {
+	ns := rbd.NodeServer{
 		DefaultNodeServer: csicommon.NewDefaultNodeServer(d, t, topology),
 		VolumeLocks:       util.NewVolumeLocks(),
-	}, nil
+	}
+	ns.SetReadAffinityMapOptions(crushLocationMap)
+
+	return &ns, nil
 }
 
 // Run start a non-blocking grpc controller,node and identityserver for
@@ -84,9 +92,10 @@ func NewNodeServer(d *csicommon.CSIDriver, t string, topology map[string]string)
 // This also configures and starts a new CSI-Addons service, by calling
 // setupCSIAddonsServer().
 func (r *Driver) Run(conf *util.Config) {
-	var err error
-	var topology map[string]string
-
+	var (
+		err                        error
+		topology, crushLocationMap map[string]string
+	)
 	// update clone soft and hard limit
 	rbd.SetGlobalInt("rbdHardMaxCloneDepth", conf.RbdHardMaxCloneDepth)
 	rbd.SetGlobalInt("rbdSoftMaxCloneDepth", conf.RbdSoftMaxCloneDepth)
@@ -128,6 +137,13 @@ func (r *Driver) Run(conf *util.Config) {
 			})
 	}
 
+	if conf.EnableReadAffinity {
+		crushLocationMap, err = util.GetCrushLocationMap(conf.CrushLocationLabels, conf.NodeID)
+		if err != nil {
+			log.FatalLogMsg(err.Error())
+		}
+	}
+
 	// Create GRPC servers
 	r.ids = NewIdentityServer(r.cd)
 
@@ -136,7 +152,7 @@ func (r *Driver) Run(conf *util.Config) {
 		if err != nil {
 			log.FatalLogMsg(err.Error())
 		}
-		r.ns, err = NewNodeServer(r.cd, conf.Vtype, topology)
+		r.ns, err = NewNodeServer(r.cd, conf.Vtype, topology, crushLocationMap)
 		if err != nil {
 			log.FatalLogMsg("failed to start node server, err %v\n", err)
 		}
@@ -162,17 +178,6 @@ func (r *Driver) Run(conf *util.Config) {
 		log.WarningLogMsg("replication service running on controller server is deprecated " +
 			"and replaced by CSI-Addons, see https://github.com/ceph/ceph-csi/issues/3314 for more details")
 		r.rs = NewReplicationServer(r.cs)
-	}
-	if !conf.IsControllerServer && !conf.IsNodeServer {
-		topology, err = util.GetTopologyFromDomainLabels(conf.DomainLabels, conf.NodeID, conf.DriverName)
-		if err != nil {
-			log.FatalLogMsg(err.Error())
-		}
-		r.ns, err = NewNodeServer(r.cd, conf.Vtype, topology)
-		if err != nil {
-			log.FatalLogMsg("failed to start node server, err %v\n", err)
-		}
-		r.cs = NewControllerServer(r.cd)
 	}
 
 	s := csicommon.NewNonBlockingGRPCServer()
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -45,6 +45,8 @@ type NodeServer struct {
 	// A map storing all volumes with ongoing operations so that additional operations
 	// for that same volume (as defined by VolumeID) return an Aborted error
 	VolumeLocks *util.VolumeLocks
+	// readAffinityMapOptions contains map options to enable read affinity.
+	readAffinityMapOptions string
 }
 
 // stageTransaction struct represents the state a transaction was when it either completed
@@ -143,7 +145,7 @@ func healerStageTransaction(ctx context.Context, cr *util.Credentials, volOps *r
 // this function also receive the credentials and secrets args as it differs in its data.
 // The credentials are used directly by functions like voljournal.Connect() and other functions
 // like genVolFromVolumeOptions() make use of secrets.
-func populateRbdVol(
+func (ns *NodeServer) populateRbdVol(
 	ctx context.Context,
 	req *csi.NodeStageVolumeRequest,
 	cr *util.Credentials,
@@ -250,6 +252,7 @@ func populateRbdVol(
 	if err != nil {
 		return nil, err
 	}
+	ns.appendReadAffinityMapOptions(rv)
 
 	rv.VolID = volID
 
@@ -263,6 +266,19 @@ func populateRbdVol(
 	}
 
 	return rv, err
+}
+
+// appendReadAffinityMapOptions appends readAffinityMapOptions to mapOptions
+// if mounter is rbdDefaultMounter and readAffinityMapOptions is not empty.
+func (ns NodeServer) appendReadAffinityMapOptions(rv *rbdVolume) {
+	switch {
+	case ns.readAffinityMapOptions == "" || rv.Mounter != rbdDefaultMounter:
+		return
+	case rv.MapOptions != "":
+		rv.MapOptions += "," + ns.readAffinityMapOptions
+	default:
+		rv.MapOptions = ns.readAffinityMapOptions
+	}
 }
 
 // NodeStageVolume mounts the volume to a staging path on the node.
@@ -318,7 +334,7 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 
 	isStaticVol := parseBoolOption(ctx, req.GetVolumeContext(), staticVol, false)
-	rv, err := populateRbdVol(ctx, req, cr)
+	rv, err := ns.populateRbdVol(ctx, req, cr)
 	if err != nil {
 		return nil, err
 	}
@@ -1348,4 +1364,23 @@ func getDeviceSize(ctx context.Context, devicePath string) (uint64, error) {
 	}
 
 	return size, nil
+}
+
+func (ns *NodeServer) SetReadAffinityMapOptions(crushLocationMap map[string]string) {
+	if len(crushLocationMap) == 0 {
+		return
+	}
+
+	var b strings.Builder
+	b.WriteString("read_from_replica=localize,crush_location=")
+	first := true
+	for key, val := range crushLocationMap {
+		if first {
+			b.WriteString(fmt.Sprintf("%s:%s", key, val))
+			first = false
+		} else {
+			b.WriteString(fmt.Sprintf("|%s:%s", key, val))
+		}
+	}
+	ns.readAffinityMapOptions = b.String()
 }
--- a/internal/util/crushlocation.go
+++ b/internal/util/crushlocation.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2023 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"strings"
+
+	"github.com/ceph/ceph-csi/internal/util/log"
+)
+
+// GetCrushLocationMap returns the crush location map, determined from
+// the crush location labels and their values from the CO system.
+// Expects crushLocationLabels in arg to be in the format "[prefix/]<name>,[prefix/]<name>,...",.
+// Returns map of crush location types with its array of associated values.
+func GetCrushLocationMap(crushLocationLabels, nodeName string) (map[string]string, error) {
+	if crushLocationLabels == "" {
+		return nil, nil
+	}
+
+	nodeLabels, err := k8sGetNodeLabels(nodeName)
+	if err != nil {
+		return nil, err
+	}
+
+	return getCrushLocationMap(crushLocationLabels, nodeLabels), nil
+}
+
+// getCrushLocationMap returns the crush location map, determined from
+// the crush location labels and node labels.
+func getCrushLocationMap(crushLocationLabels string, nodeLabels map[string]string) map[string]string {
+	labelsToRead := strings.Split(crushLocationLabels, labelSeparator)
+	log.DefaultLog("CRUSH location labels passed for processing: %+v", labelsToRead)
+
+	labelsIn := make(map[string]bool, len(labelsToRead))
+	for _, label := range labelsToRead {
+		labelsIn[label] = true
+	}
+
+	// Determine values for requested labels from node labels
+	crushLocationMap := make(map[string]string, len(labelsIn))
+	for key, value := range nodeLabels {
+		if _, ok := labelsIn[key]; !ok {
+			continue
+		}
+		// label found split name component and store value
+		nameIdx := strings.IndexRune(key, keySeparator)
+		crushLocationType := strings.TrimSpace(key[nameIdx+1:])
+		if crushLocationType == "hostname" {
+			// ceph defaults to "host" while Kubernetes uses "hostname" as key.
+			crushLocationType = "host"
+		}
+		// replace "." with "-" to satisfy ceph crush map.
+		value = strings.Replace(strings.TrimSpace(value), ".", "-", -1)
+		crushLocationMap[crushLocationType] = value
+	}
+
+	if len(crushLocationMap) == 0 {
+		return nil
+	}
+	log.DefaultLog("list of CRUSH location processed: %+v", crushLocationMap)
+
+	return crushLocationMap
+}
--- a/internal/util/crushlocation_test.go
+++ b/internal/util/crushlocation_test.go
@@ -1,0 +1,113 @@
+/*
+Copyright 2023 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_getCrushLocationMap(t *testing.T) {
+	t.Parallel()
+	type input struct {
+		crushLocationLabels string
+		nodeLabels          map[string]string
+	}
+	tests := []struct {
+		name string
+		args input
+		want map[string]string
+	}{
+		{
+			name: "empty crushLocationLabels",
+			args: input{
+				crushLocationLabels: "",
+				nodeLabels:          map[string]string{},
+			},
+			want: nil,
+		},
+		{
+			name: "empty nodeLabels",
+			args: input{
+				crushLocationLabels: "topology.io/zone,topology.io/rack",
+				nodeLabels:          map[string]string{},
+			},
+			want: nil,
+		},
+		{
+			name: "matching crushlocation and node labels",
+			args: input{
+				crushLocationLabels: "topology.io/zone,topology.io/rack",
+				nodeLabels: map[string]string{
+					"topology.io/zone": "zone1",
+				},
+			},
+			want: map[string]string{"zone": "zone1"},
+		},
+		{
+			name: "multuple matching crushlocation and node labels",
+			args: input{
+				crushLocationLabels: "topology.io/zone,topology.io/rack",
+				nodeLabels: map[string]string{
+					"topology.io/zone": "zone1",
+					"topology.io/rack": "rack1",
+				},
+			},
+			want: map[string]string{"zone": "zone1", "rack": "rack1"},
+		},
+		{
+			name: "no match between crushlocation and node labels",
+			args: input{
+				crushLocationLabels: "topology.io/zone,topology.io/rack",
+				nodeLabels: map[string]string{
+					"topology.io/region": "region1",
+				},
+			},
+			want: nil,
+		},
+		{
+			name: "check crushlocation value replacement to satisfy ceph requirement",
+			args: input{
+				crushLocationLabels: "topology.io/zone,topology.io/rack",
+				nodeLabels: map[string]string{
+					"topology.io/zone": "south.east.1",
+				},
+			},
+			want: map[string]string{"zone": "south-east-1"},
+		},
+		{
+			name: "hostname key should be replaced with host",
+			args: input{
+				crushLocationLabels: "topology.io/zone,topology.io/hostname",
+				nodeLabels: map[string]string{
+					"topology.io/hostname": "worker-1",
+				},
+			},
+			want: map[string]string{"host": "worker-1"},
+		},
+	}
+	for _, tt := range tests {
+		currentTT := tt
+		t.Run(currentTT.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t,
+				currentTT.want,
+				getCrushLocationMap(currentTT.args.crushLocationLabels, currentTT.args.nodeLabels))
+		})
+	}
+}
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -147,6 +147,10 @@ type Config struct {
 
 	// Cluster name
 	ClusterName string
+
+	// Read affinity related options
+	EnableReadAffinity  bool   // enable OSD read affinity.
+	CrushLocationLabels string // list of CRUSH location labels to read from the node.
 }
 
 // ValidateDriverName validates the driver name.
--- a/internal/rbd/nodeserver_test.go
+++ b/internal/rbd/nodeserver_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestGetStagingPath(t *testing.T) {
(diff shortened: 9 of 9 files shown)


docs/deploy-rbd.md
```diff
@@ -196,6 +198,18 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
 
 [See the Helm chart readme for installation instructions.](../charts/ceph-csi-rbd/README.md)
 
+## Read Affinity using crush locations for RBD volumes
+
+CephCSI supports mapping RBD volumes with krbd options
```
humblec: Ceph CSI

docs/deploy-rbd.md
```diff
@@ -196,6 +198,18 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
...
+`"read_from_replica=localize,crush_location=type1:value1|type2:value2"` to achieve
+reads from the most local serving OSD.
+Refer [krbd-options](https://docs.ceph.com/en/latest/man/8/rbd/#kernel-rbd-krbd-options) for more details.
+
+This can be enabled by adding labels to kubernetes nodes in the format `"topology.io/zone=east"` and passing
+cmdline arguments `"--enable-read-affinity=true"` and `"--crush-location-labels=topology.io/zone,topology.io/region"`
+to CephCSI RBD daemonset pod "csi-rbdplugin" container, resulting in CephCSI adding
```
humblec: same here

internal/rbd/rbd_attach.go
```diff
@@ -388,6 +388,35 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, userOptions string) []stri
 	return cmdArgs
 }
 
+// appendCruchLocationMapOptions parses crushLocationMap to the format
```
humblec: appendcrush..

internal/rbd/rbd_attach.go
```diff
@@ -388,6 +388,35 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, userOptions string) []stri
...
+		if crushLocation != "" {
+			crushLocation = crushLocation + fmt.Sprintf("|%s:%s", key, val)
+		} else {
+			crushLocation = fmt.Sprintf("%s:%s", key, val)
+		}
+	}
+	if crushLocation == "" {
+		// if no crush locations are found, return the mapOptions as it is.
+		return mapOptions
+	}
```
humblec: looks like this section can be improved..

internal/csi-common/driver.go
```diff
@@ -33,7 +33,10 @@ type CSIDriver struct {
 	nodeID  string
 	version string
 	// topology constraints that this nodeserver will advertise
-	topology     map[string]string
+	topology map[string]string
+	// crushLocationMap that will be used to enable read affinity.
+	CrushLocationMap map[string]string
```
nixpanic: does this need to be exported? Ideally it isn't.

internal/rbd/rbd_util.go
```diff
@@ -177,6 +177,8 @@ type rbdVolume struct {
 	RequestedVolSize   int64
 	DisableInUseChecks bool
 	readOnly           bool
+	// CrushLocationMap contains details required to enable read affinity.
+	CrushLocationMap map[string]string
```
nixpanic: This is not a rbdVolume property, it should not be added here.

internal/rbd/rbd_attach.go
```diff
@@ -306,7 +306,7 @@ func getMapOptions(req *csi.NodeStageVolumeRequest, rv *rbdVolume) error {
 		return err
 	}
 	if rv.Mounter == rbdDefaultMounter {
-		rv.MapOptions = krbdMapOptions
+		rv.MapOptions = appendCruchLocationMapOptions(krbdMapOptions, rv.CrushLocationMap)
```
nixpanic: instead of `rv.CrushLocationMap`, use `ns.CrushLocationMap` and make it a function of the NodeServer.

nixpanic (changes requested):
How can this be tested? Some steps/guidance and complete examples for that should probably be documented too.

cmd/cephcsi.go
```diff
@@ -81,6 +81,13 @@ func init() {
 		"",
 		"list of kubernetes node labels, that determines the topology"+
 			" domain the node belongs to, separated by ','")
+	flag.BoolVar(&conf.EnableReadAffinity, "enable-read-affinity", false, "enable read affinity")
+	flag.StringVar(
```
Madhu-1, line 85: Please add a document that setting this will use the same affinity to all the ceph clusters that cephcsi is connected.

internal/csi-common/driver.go
Rakshith-R: > does this need to be exported? Ideally it isn't.

Yes, it needs to exported to be accessible to individual drivers to use in rpc function call.

internal/rbd/rbd_util.go
Rakshith-R: > This is not a rbdVolume property, it should not be added here.

removed it, its much simpler now.

Rakshith-R:
> How can this be tested? Some steps/guidance and complete examples for that should probably be documented too.

I've documented it here https://github.com/ceph/ceph-csi/pull/3639/files#diff-ffd3d77379c6dad83a2a35da3df2161ac6e62ee31396ce0657ba06662ef0b965 .

We just need to add labels to kubernetes node, set those cmdline args for rbd nodeplugin daemonset container and
required options will be passed during rbd map command.

The rendered form will look like this https://github.com/ceph/ceph-csi/blob/37925274b7de1150ca0adc79f703698d89d539ad/docs/deploy-rbd.md#read-affinity-using-crush-locations-for-rbd-volumes

@idryomov @travisn , Can you please take a look too?

docs/deploy-rbd.md
```diff
@@ -196,6 +198,18 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
...
 [See the Helm chart readme for installation instructions.](../charts/ceph-csi-rbd/README.md)
 
+## Read Affinity using crush locations for RBD volumes
+
+Ceph CSI supports mapping RBD volumes with krbd options
+`"read_from_replica=localize,crush_location=type1:value1|type2:value2"` to achieve
+reads from the most local serving OSD.
+Refer [krbd-options](https://docs.ceph.com/en/latest/man/8/rbd/#kernel-rbd-krbd-options) for more details.
+
+This can be enabled by adding labels to kubernetes nodes in the format `"topology.io/zone=east"` and passing
```
Madhu-1: `kubernetes` to `Kubernetes`

docs/deploy-rbd.md
```diff
@@ -196,6 +198,18 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
...
+cmdline arguments `"--enable-read-affinity=true"` and `"--crush-location-labels=topology.io/zone,topology.io/region"`
+to Ceph CSI RBD daemonset pod "csi-rbdplugin" container, resulting in Ceph CSI adding
+`"--options read_from_replica=localize,crush_location=zone:east"` krbd options during rbd map operation.
```
Madhu-1: Add link/predefined expected label keys here https://kubernetes.io/docs/reference/labels-annotations-taints/

internal/rbd/rbd_attach.go
```diff
@@ -388,6 +388,32 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, userOptions string) []stri
...
+		}
+	}
+
+	newMapOptions := fmt.Sprintf("read_from_replica=localize,crush_location=%s", crushLocation)
+	if mapOptions != "" {
+		newMapOptions = mapOptions + "," + newMapOptions
+	}
+
+	return newMapOptions
+}
```
Madhu-1:
```suggestion
func appendCrushLocationMapOptions(mapOptions string, crushLocationMap map[string]string) string {
	if len(crushLocationMap) == 0 {
		return mapOptions
	}

	var b strings.Builder
	b.WriteString("read_from_replica=localize,crush_location=")
	first := true
	for key, val := range crushLocationMap {
		if first {
			b.WriteString(fmt.Sprintf("%s:%s", key, val))
			first = false
		} else {
			b.WriteString(fmt.Sprintf("|%s:%s", key, val))
		}
	}

	if mapOptions != "" {
		b.WriteString("," + mapOptions)
	}

	return b.String()
}
```

internal/rbd/rbd_attach_test.go
```diff
@@ -104,3 +106,62 @@ func TestParseMapOptions(t *testing.T) {
...
 }
+
+func Test_appendCrushLocationMapOptions(t *testing.T) {
+	type args struct {
+		mapOptions       string
+		crushLocationMap map[string]string
+	}
+	tests := []struct {
+		name string
+		args args
```
Madhu-1: Define mapOptions and crushLocationMap here, no need for one more struct

internal/util/crushlocation.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, err
+	}
+
+	return getCrushLocationMap(crushLocationLabels, nodeLabels), nil
+}
+
+// getCrushLocationMap returns the crush location map, determined from
+// the crush location labels and node labels.
+func getCrushLocationMap(crushLocationLabels string, nodeLabels map[string]string) map[string]string {
+	labelsToRead := strings.SplitN(crushLocationLabels, labelSeparator, -1)
```
Madhu-1: instead of SplitN you can use Split?

internal/util/crushlocation.go
```diff
@@ -0,0 +1,72 @@
...
+	log.DefaultLog("crush location labels passed for processing: %+v", labelsToRead)
+
+	labelsIn := make(map[string]bool)
```
Madhu-1:
```suggestion
labelsIn := make(map[string]bool),len(labelsToRead))
```

internal/util/crushlocation.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, label := range labelsToRead {
+		labelsIn[label] = true
+	}
+
+	// Determine values for requested labels from node labels
+	crushLocationMap := make(map[string]string)
```
Madhu-1: same here initialize to len of labelsToRead

internal/util/crushlocation_test.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+func Test_getCrushLocationMap(t *testing.T) {
+	type args struct {
+		crushLocationLabels string
+		nodeLabels          map[string]string
+	}
+	tests := []struct {
+		name string
+		args args
```
Madhu-1: No need for one more args struct.

internal/rbd/rbd_attach.go
Madhu-1: Is this going to remain the same for cephfs also? if yes, we can move this to util and reuse

Madhu-1 (changes requested):
Good to add an e2e test for this one. To make sure the labels are set on the rbd mapped device.

internal/util/crushlocation.go
Rakshith-R: maps don't have their length predefined.
arrays do.

internal/util/crushlocation.go
Rakshith-R: same as above
https://github.com/ceph/ceph-csi/pull/3639#discussion_r1095519562

internal/rbd/rbd_attach_test.go
Rakshith-R: > Define mapOptions and crushLocationMap here, no need for one more struct

This is unit test cases, most parts are auto generated. 
I think its clear this way and I don't see problem with sticking with this/ advantage to changing the structs.

internal/util/crushlocation_test.go
Rakshith-R: same as above

internal/util/crushlocation.go
Madhu-1: I didn't get what you mean by `maps don't have their length predefined.` the ask is to initialize the map for space of predefined length as `labelsToRead` as you are not going to more items than that.

internal/rbd/rbd_attach_test.go
Madhu-1: Auto-generated code depends from tool to tool. I see that there is little value in having more struct. If you see a value, please explain it so we can consider and decide on it.

internal/rbd/rbd_attach.go
Rakshith-R: I think the current function code is simple enough.

there's no options `read_from_replica=localize,crush_location=key:val` for cephfs atm, we can modify later based on its availability.

https://docs.ceph.com/en/quincy/man/8/mount.ceph/#basic

docs/deploy-rbd.md
Rakshith-R: > Add link/predefined expected label keys here https://kubernetes.io/docs/reference/labels-annotations-taints/

We don't have a restriction at cephcsi level to have such expectations, user can use any labels he wants.

cmd/cephcsi.go
Rakshith-R: > Please add a document that setting this will use the same affinity to all the ceph clusters that cephcsi is connected.

I think that is implied. There's no statement informing users otherwise.

internal/rbd/rbd_attach.go
Madhu-1: am looking at more optimized and readable code and also simple now but can having simple code can become complicated in the future.

cmd/cephcsi.go
Madhu-1: Having a note doesn't hurt and it helps users to understand the impact of it.

docs/deploy-rbd.md
Madhu-1: But still, you need to have a format to follow where the exact format is documented. i can add  a label is `topology.io/zone/test=east` or any arbitrary value, if we dont have any predefined strict rules defined on the format user can set any values he like in different formats

internal/rbd/rbd_attach_test.go
Rakshith-R: It clearly defines what are the individual arguments instead of cramming them in a single struct.

Madhu-1: sounds good. dont use the same name for both the struct member and the struct name

internal/rbd/driver/driver.go
```diff
@@ -168,7 +182,7 @@ func (r *Driver) Run(conf *util.Config) {
 		if err != nil {
```
Rakshith-R: @Madhu-1 @nixpanic 
I am removing this segment of code.
https://github.com/ceph/ceph-csi/blob/650c522ce016bcff86d590006d120f7b829fb93d/internal/rbd/driver/driver.go#L166-L176

changes to controllerserver and nodeserver standalone if statements are not propagated to this part from a long time (see krbd feature and replication sever), 
and I don't think this section is necessary. 

The complexity of this function will go down and it'll be much clearer.
wdyt?

Madhu-1: sounds good to remove it not required and make its less complex

internal/util/crushlocation.go
Rakshith-R: :sweat_smile:  most programming language does not have this kind of predefined length for map type...
Its added now.

internal/rbd/driver/driver.go
```diff
@@ -163,17 +177,6 @@ func (r *Driver) Run(conf *util.Config) {
...
-		topology, err = util.GetTopologyFromDomainLabels(conf.DomainLabels, conf.NodeID, conf.DriverName)
-		if err != nil {
-			log.FatalLogMsg(err.Error())
-		}
-		r.ns, err = NewNodeServer(r.cd, conf.Vtype, topology)
-		if err != nil {
-			log.FatalLogMsg("failed to start node server, err %v\n", err)
-		}
-		r.cs = NewControllerServer(r.cd)
-	}
```
Madhu-1, line 176: This takes out the possibility of running cephcsi as both controller and provisioner. good to add this to release notes

Madhu-1 (dismissed):
LGTM, good to see if you can add an E2E for this one to make sure functionality works and never breaks in future.

internal/rbd/driver/driver.go
Rakshith-R: done added a note in tracker issue https://github.com/ceph/ceph-csi/issues/3336#issuecomment-1415731640 

opened todo issue for e2e and helm options.
https://github.com/ceph/ceph-csi/issues/3642

internal/rbd/rbd_attach.go
```diff
@@ -388,6 +388,33 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, userOptions string) []stri
...
+// "--read_from_replica=localize,--crush_location=key1:value1|key3:value2"
+// and appends it to mapOptions.
+func appendCrushLocationMapOptions(mapOptions string, crushLocationMap map[string]string) string {
+	if len(crushLocationMap) == 0 {
+		// if no crush locations are found, return the mapOptions as it is.
+		return mapOptions
+	}
+
+	var b strings.Builder
+	b.WriteString("read_from_replica=localize,crush_location=")
```
humblec: may be use the input string as `const` ?

Rakshith-R: > may be use the input string as `const` ?

This currently used at only one place, let's do so when we'll be using it at multiple places.

deploy/rbd/kubernetes/csi-rbdplugin.yaml
```diff
@@ -68,6 +68,15 @@ spec:
             # and pass the label names below, for CSI to consume and advertise
             # its equivalent topology domain
             # - "--domainlabels=failure-domain/region,failure-domain/zone"
+            #
+            # Options to enable read affinity.
+            # If enabled Ceph CSI will fetch labels from kubernetes node
```
idryomov:
```suggestion
            # If enabled Ceph CSI will fetch labels from kubernetes node and
```

cmd/cephcsi.go
```diff
@@ -81,6 +81,13 @@ func init() {
...
+		&conf.CrushLocationLabels,
+		"crush-location-labels",
+		"",
+		"list of kubernetes node labels, that determines the"+
+			" crush location the node belongs to, separated by ','")
```
idryomov: CRUSH is an abbreviation so it should be capitalized in the option descriptions and documentation:
```suggestion
			" CRUSH location the node belongs to, separated by ','")
```

internal/util/crushlocation_test.go
```diff
@@ -0,0 +1,103 @@
...
+				},
+			},
+			want: map[string]string{"zone": "zone1", "rack": "rack1"},
+		},
+		{
+			name: "no match between crushlocation and node labels",
+			args: input{
+				crushLocationLabels: "topology.io/zone,topology.io/rack",
+				nodeLabels: map[string]string{
+					"topology.io/region": "zone1",
```
idryomov:
```suggestion
					"topology.io/region": "region1",
```

internal/util/crushlocation.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Determine values for requested labels from node labels
+	crushLocationMap := make(map[string]string, len(labelsIn))
+	for key, value := range nodeLabels {
+		if _, ok := labelsIn[key]; ok {
+			// label found split name component and store value
+			nameIdx := strings.IndexRune(key, keySeparator)
+			crushLocationType := strings.TrimSpace(key[nameIdx+1:])
+			// replace "." with "-" to satisfy ceph crush map.
+			value = strings.Replace(strings.TrimSpace(value), ".", "-", -1)
```
idryomov: Is this really needed?  Are dots in zone and region names expected?

If nothing else, these kinds of transformations should be highlighted in the documentation.

docs/deploy-rbd.md
```diff
@@ -196,6 +198,24 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
...
+Refer [krbd-options](https://docs.ceph.com/en/latest/man/8/rbd/#kernel-rbd-krbd-options)
+for more details.
+
+This can be enabled by adding labels to Kubernetes nodes in the format `"topology.io/zone=east"`
+and passing cmdline arguments `"--enable-read-affinity=true"` and
+`"--crush-location-labels=topology.io/zone,topology.io/region"`
+to Ceph CSI RBD daemonset pod "csi-rbdplugin" container, resulting in Ceph CSI adding
+`"--options read_from_replica=localize,crush_location=zone:east"` krbd options during
+rbd map operation.
+If enabled, this option will be added to all RBD volumes mapped by Ceph CSI.
```
idryomov, line 217: It would be nice to also note that, for this setting to be effective, the OSDs should be assigned to the same  zones, regions, etc respectively on the cluster side (in the CRUSH map).  Just declaring that a particular node belongs to a particular zone on the client side is insufficient.

internal/rbd/rbd_attach.go
```diff
@@ -388,6 +388,33 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, userOptions string) []stri
 	return cmdArgs
 }
 
+// appendCrushLocationMapOptions parses crushLocationMap to the format
+// "--read_from_replica=localize,--crush_location=key1:value1|key3:value2"
```
idryomov: Actual parsing happens in `getCrushLocationMap`:
```suggestion
// appendCrushLocationMapOptions uses crushLocationMap to the format
// "read_from_replica=localize,crush_location=key1:value1|key2:value2"
```

internal/rbd/rbd_attach_test.go
```diff
@@ -104,3 +106,62 @@ func TestParseMapOptions(t *testing.T) {
...
+				crushLocationMap: map[string]string{
+					"region": "west",
+				},
+			},
+			want: "read_from_replica=localize,crush_location=region:west",
+		},
+		{
+			name: "filled mapOptions and crushLocationMap",
+			args: input{
+				mapOptions: "read-only=true",
```
idryomov: It doesn't matter for this unit test, but all krbd options use underscores (so `read_only`,  not `read-only`).  Someone grepping for `mapOptions` examples in the future may get confused.

internal/util/crushlocation.go
Rakshith-R: > Is this really needed? Are dots in zone and region names expected?

yes there maybe dots in the values, there's no such restriction on kubernetes side.

> If nothing else, these kinds of transformations should be highlighted in the documentation.

This is adapted from rook code which said the same thing,
I'll add a note in documentation.

internal/rbd/rbd_attach.go
Rakshith-R: changed it to `converts` now

internal/rbd/rbd_attach_test.go
Rakshith-R: https://docs.ceph.com/en/latest/man/8/rbd/#kernel-rbd-krbd-options 

changed it to `notrim` instead to avoid confusion

docs/deploy-rbd.md
Rakshith-R: I've modified the first sentence to elaborate on it, 
But users should find more info on ceph krbd map options site which is linked in second sentence.

Rakshith-R:
Thanks for the review @idryomov, 
I've made changes, ptal

docs/deploy-rbd.md
```diff
@@ -48,6 +48,8 @@ make image-cephcsi
...
 | `--setmetadata`          | `false`                       | Set metadata on volume                                                                                                                                                                                                                                                               |
+| `--enable-read-affinity` | `false`                       | enable read affinity                                                                                                                                                                                                                                                                 |
+| `--crush-location-labels`| _empty_                       | Kubernetes node labels that determines the CRUSH location the node belongs to, separated by ','                                                                                                                                                                                      |
```
idryomov:
```suggestion
| `--crush-location-labels`| _empty_                       | Kubernetes node labels that determine the CRUSH location the node belongs to, separated by ','                                                                                                                                                                                      |
```

docs/deploy-rbd.md
```diff
@@ -196,6 +198,28 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
 
 [See the Helm chart readme for installation instructions.](../charts/ceph-csi-rbd/README.md)
 
+## Read Affinity using crush locations for RBD volumes
+
+Ceph CSI supports mapping RBD volumes with krbd options
+`"read_from_replica=localize,crush_location=type1:value1|type2:value2"` to
+optimize for reads from the most local serving OSD (OSD assigned in the
+same ceph CRUSH map location i.e.,region,zone,rack,etc).
```
idryomov: The CRUSH location doesn't have to be the _same_, the proximity gets calculated:
```suggestion
Ceph CSI supports mapping RBD volumes with krbd options
`"read_from_replica=localize,crush_location=type1:value1|type2:value2"` to
allow serving reads from the most local OSD (according to OSD locations as
defined in the CRUSH map).
```

docs/deploy-rbd.md
```diff
@@ -196,6 +198,28 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
...
+Refer [krbd-options](https://docs.ceph.com/en/latest/man/8/rbd/#kernel-rbd-krbd-options)
+for more details.
+
+This can be enabled by adding labels to Kubernetes nodes in the format `"topology.io/zone=east"`
+and passing cmdline arguments `"--enable-read-affinity=true"` and
+`"--crush-location-labels=topology.io/zone,topology.io/region"`
+to Ceph CSI RBD daemonset pod "csi-rbdplugin" container, resulting in Ceph CSI adding
+`"--options read_from_replica=localize,crush_location=zone:east"` krbd options during
+rbd map operation.
```
idryomov: I'd extend the example to two CRUSH locations:
```suggestion
This can be enabled by adding labels to Kubernetes nodes like
`"topology.io/region=east"` and `"topology.io/zone=east-zone1"` and
passing cmdline arguments `"--enable-read-affinity=true"` and
`"--crush-location-labels=topology.io/zone,topology.io/region"` to Ceph CSI
RBD daemonset pod "csi-rbdplugin" container, resulting in Ceph CSI adding
`"--options read_from_replica=localize,crush_location=zone:east-zone1|region:east"`
krbd options during rbd map operation.
```

internal/rbd/rbd_attach.go
```diff
@@ -388,6 +388,33 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, userOptions string) []stri
 	return cmdArgs
 }
 
+// appendCrushLocationMapOptions converts crushLocationMap to the format
+// "--read_from_replica=localize,--crush_location=key1:value1|key2:value2"
```
idryomov:
```suggestion
// "read_from_replica=localize,crush_location=key1:value1|key2:value2"
```

internal/rbd/rbd_attach_test.go
```diff
@@ -104,3 +106,62 @@ func TestParseMapOptions(t *testing.T) {
...
+				crushLocationMap: map[string]string{
+					"region": "west",
+				},
+			},
+			want: "read_from_replica=localize,crush_location=region:west",
+		},
+		{
+			name: "filled mapOptions and crushLocationMap",
+			args: input{
+				mapOptions: "notrim=true",
```
idryomov: `notrim` is just a flag, it doesn't take any arguments:
```suggestion
				mapOptions: "notrim",
```

internal/rbd/rbd_attach_test.go
```diff
@@ -104,3 +106,62 @@ func TestParseMapOptions(t *testing.T) {
...
+			args: input{
+				mapOptions: "",
+				crushLocationMap: map[string]string{
+					"region": "west",
+				},
+			},
+			want: "read_from_replica=localize,crush_location=region:west",
+		},
+		{
+			name: "filled mapOptions and crushLocationMap",
```
idryomov: Need an additional test for multiple CRUSH locations (i.e. for appending via `|`).

Rakshith-R: > Need an additional test for multiple CRUSH locations (i.e. for appending via `|`).

since we are using map, the function may yield multiple output (position of key:value pair maybe interchanged ). Modified the unit test case to handle this scenario. 
ptal

docs/deploy-rbd.md
```diff
@@ -196,6 +198,29 @@ The Helm chart is located in `charts/ceph-csi-rbd`.
...
+Ceph CSI supports mapping RBD volumes with krbd options
+`"read_from_replica=localize,crush_location=type1:value1|type2:value2"` to
+allow serving reads from the most local OSD (according to OSD locations as
+defined in the CRUSH map).
+Refer [krbd-options](https://docs.ceph.com/en/latest/man/8/rbd/#kernel-rbd-krbd-options)
+for more details.
+
+This can be enabled by adding labels to Kubernetes nodes like
+`"topology.io/region=east"` and `"topology.io/zone=east-zone1"` and
+passing cmdline arguments `"--enable-read-affinity=true"` and
```
humblec: command line

internal/util/crushlocation.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return getCrushLocationMap(crushLocationLabels, nodeLabels), nil
+}
+
+// getCrushLocationMap returns the crush location map, determined from
+// the crush location labels and node labels.
+func getCrushLocationMap(crushLocationLabels string, nodeLabels map[string]string) map[string]string {
+	labelsToRead := strings.Split(crushLocationLabels, labelSeparator)
+	log.DefaultLog("CRUSH location labels passed for processing: %+v", labelsToRead)
```
humblec, line 46: do we need to log this, I think its better to skip ?

internal/util/util.go
```diff
@@ -147,6 +147,10 @@ type Config struct {
 
 	// Cluster name
 	ClusterName string
+
+	// Read affinity related options
+	EnableReadAffinity  bool   // enable read affinity.
```
humblec: enable OSD read affinity ?

internal/util/crushlocation.go
Rakshith-R: > do we need to log this, I think its better to skip ?

Let's have this, it'll be only logged once when the nodeplugin pod starts anyway.

idryomov (approved):
Ack from the RBD perspective

Madhu-1 (dismissed):

Rakshith-R:
As @idryomov  pointed out here https://github.com/rook/rook/pull/11631#discussion_r1098471171

We need consider hostname to host conversion too, PTAL.

Rakshith-R:
moved code around, golangi linter was complaining about too much code inside conditional statement.
https://github.com/ceph/ceph-csi/compare/c3d85ba9d5c9785190d86a997d8ea02167eb75b4..491f19916aa735ff02d3935143fefa65f57c9451

internal/util/crushlocation.go
```diff
@@ -0,0 +1,77 @@
...
+	crushLocationMap := make(map[string]string, len(labelsIn))
+	for key, value := range nodeLabels {
+		if _, ok := labelsIn[key]; !ok {
+			continue
+		}
+		// label found split name component and store value
+		nameIdx := strings.IndexRune(key, keySeparator)
+		crushLocationType := strings.TrimSpace(key[nameIdx+1:])
+		if crushLocationType == "hostname" {
+			// ceph defaults to "host" while kubernetes uses "hostname" as key.
```
humblec: s/kubernetes/Kubernetes

cmd/cephcsi.go
```diff
@@ -81,6 +81,13 @@ func init() {
 		"",
 		"list of kubernetes node labels, that determines the topology"+
```
humblec: s/kubernetes/Kubernetes

cmd/cephcsi.go
```diff
@@ -81,6 +81,13 @@ func init() {
...
 			" domain the node belongs to, separated by ','")
+	flag.BoolVar(&conf.EnableReadAffinity, "enable-read-affinity", false, "enable read affinity")
+	flag.StringVar(
+		&conf.CrushLocationLabels,
+		"crush-location-labels",
+		"",
+		"list of kubernetes node labels, that determines the"+
```
humblec: same here

internal/util/util.go
```diff
@@ -147,6 +147,10 @@ type Config struct {
 
 	// Cluster name
 	ClusterName string
+
+	// Read affinity related options
+	EnableReadAffinity  bool   // enable OSD read affinity.
+	CrushLocationLabels string // list of crush location labels to read from the node.
```
humblec: s/crush to CRUSH

humblec (commented):
nit: we can address later too.

humblec (dismissed):

cmd/cephcsi.go
nixpanic: A single Ceph-CSI deployment supports multiple Ceph clusters... What is the reason to have this as a commandline parameter, and not a setting in the StorageClass or the ConfigMap?

internal/csi-common/driver.go
nixpanic: Why is then this different from `topology`?

Rakshith-R: > Why is then this different from `topology`?

Currently processing required for topology is identical to cephfs and rbd (topology feature is itself in beta stage with lot of TODOs), 
Whereas processing required for CrushLocation maybe different for cephfs and rbd,(only rbd support read affinity at the moment.). Therefore, its better to let individual drivers handle it.

cmd/cephcsi.go
Rakshith-R: > A single Ceph-CSI deployment supports multiple Ceph clusters... What is the reason to have this as a commandline parameter, and not a setting in the StorageClass or the ConfigMap?

Its simpler to have it as comandline parameter.
In most of the cases, clusters will share the same crush map characteristics and this design
should be sufficient.

StorageClasses are immutable, therefore existing rbd volumes will not be able to use this feature.

The current implementation covers most of the cases and there should be no need to have this in
a configmap with unique labels per cluster. If such a need arises in the future, it will be easy to
override cmdline labels and prefer labels per cluster in configmap. 

One more reason is of performance, using cmdline args will enable us to restart the container and
fetch the labels only at the time of restart instead of having to fetch it for each map/mount requests.

Rakshith-R:
Everytime I press on re-request review from someone less, 
github seems to be removing request for you @idryomov. 
I apologise if this spamming your inbox :(
Screenshot from 2023-02-09 10-47-01

Screenshot from 2023-02-09 10-49-58

cmd/cephcsi.go
nixpanic: > Its simpler to have it as comandline parameter.

For who? You as a developer, or administrators deploying Ceph-CSI? I'd guess a StorageClass+ConfigMap option would be much easier for administrators.

I have no idea if CRUSH maps are commonly the same on different clusters. My assumption would be that it is highly unlikely, and they would be different per cluster.

IMHO, putting this change in the StorageClass would be suitable, as a SC already refers to a particular cluster.

In case of users that upgrade and want to benefit from this feature, placing it in the ConfigMap should work too?

Fetching labels of nodes needs to be dynamic anyway, I guess? What if a cluster changes, nodes get removed, added, relocated? Autoscaling a Kubernetes cluster is a thing, so fetching the labels only at startup of the provisioner isn't correct.

Rakshith-R: > > Its simpler to have it as comandline parameter.
> 
> For who? You as a developer, or administrators deploying Ceph-CSI? I'd guess a StorageClass+ConfigMap option would be much easier for administrators.

Both, I'd like to think of this feature as set and forget.
Topology of cluster with existing nodes is rarely changed. OSDs are assinged using topology labels on the node.
rbd volumes use the same (with cephcsi's help) to figure out their nearest OSD for read affinity.

> 
> I have no idea if CRUSH maps are commonly the same on different clusters. My assumption would be that it is highly unlikely, and they would be different per cluster.

https://docs.ceph.com/en/quincy/rados/operations/crush-map/#types-and-buckets

We really only care about the CRUSH map type and value, not the exact structure.
 If the cluster is situated in kubernetes, then 
OSD are assigned according to topology labels of the node. This is the same set of topology that will be used by cephcsi to extract the crush location and pass it on to the device while mapping.

CRUSH maps will be not be very different from each other among ceph clusters running on the same kubernetes cluster.

> 
> IMHO, putting this change in the StorageClass would be suitable, as a SC already refers to a particular cluster.

Existing pvc/storageclasses will not be able to take advantage of this feature. 

> 
> In case of users that upgrade and want to benefit from this feature, placing it in the ConfigMap should work too?
> 
> Fetching labels of nodes needs to be dynamic anyway, I guess? What if a cluster changes, nodes get removed, added, relocated? Autoscaling a Kubernetes cluster is a thing, so fetching the labels only at startup of the provisioner isn't correct.

This is happening in the nodeplugin pod and will be used during mapping.
In any of the scenario described above, cephcsi nodeplugin pod will get restarted and we will fetch it again. 
Each nodeplugin pod cares only about its node's labels.

+ avoids repetition of same labels in each storageclass / configmap per cluster section.

Rakshith-R:
@mergifyio rebase

cmd/cephcsi.go
nixpanic: I guess we can always move it to the ConfigMap if more flexibility is required.

Right, of course it is only used in the node-plugin, so changes to the number of nodes in the Kubernetes cluster is handled automatically.

internal/csi-common/driver.go
nixpanic: The `CSIDriver` struct does not expose any attributes at the moment. It would be better to provide access/get function(s) for it instead. Sortof what you have in `internal/util/crushlocation.go`, but maybe even further and generate the `crush_location=` rbd commandline option?

internal/rbd/rbd_attach.go
```diff
@@ -388,6 +388,33 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, userOptions string) []stri
...
+func appendCrushLocationMapOptions(mapOptions string, crushLocationMap map[string]string) string {
+	if len(crushLocationMap) == 0 {
+		// if no crush locations are found, return the mapOptions as it is.
+		return mapOptions
+	}
+
+	var b strings.Builder
+	b.WriteString("read_from_replica=localize,crush_location=")
+	first := true
+	for key, val := range crushLocationMap {
```
nixpanic: `key` and `val` may not contain `,`, else things will go wrong. This is use-input, so you'll need to check it more carefully. There might be other characters that are forbidden.

internal/rbd/rbd_attach.go
```diff
@@ -388,6 +388,33 @@ func appendKRbdDeviceTypeAndOptions(cmdArgs []string, userOptions string) []stri
 	return cmdArgs
 }
 
+// appendCrushLocationMapOptions converts crushLocationMap to the format
+// "read_from_replica=localize,crush_location=key1:value1|key2:value2"
+// and appends it to mapOptions.
+func appendCrushLocationMapOptions(mapOptions string, crushLocationMap map[string]string) string {
```
nixpanic: Why generate this string on every mount? It can be generated once at startup, because the crushLocationMap never changes while the node-plugin is running. It would be much nicer if the NodeServer has a function to return the map, or maybe even the generated string.

internal/rbd/nodeserver.go
```diff
@@ -318,7 +319,7 @@ func (ns *NodeServer) NodeStageVolume(
 	}
 
 	isStaticVol := parseBoolOption(ctx, req.GetVolumeContext(), staticVol, false)
-	rv, err := populateRbdVol(ctx, req, cr)
+	rv, err := populateRbdVol(ctx, req, cr, ns.Driver.CrushLocationMap)
```
nixpanic: `populateRbdVol()` could be a function of `NodeServer`, so that you do not need to pass extra arguments.

nixpanic (changes requested):
Current way of how it is written feels like a little too hacky. You can more cleanly extend the NodeServer struct with functions, so that passing the crush map locations isn't needed that much. It'll make the code easier to understand and maintain.

I think my suggestions require relatively small changes, but should cleanup the flow quite a bit.

internal/rbd/rbd_attach.go
Rakshith-R: > `key` and `val` may not contain `,`, else things will go wrong. This is use-input, so you'll need to check it more carefully. There might be other characters that are forbidden.

@idryomov 
Any suggestion on validating this ? 
It'll be good to know what characters are valid and what's not in key:val parts

Rakshith-R: kubernetes already has a check in place, so we should be good.

```
$ k label node minikube hello.io/he="abcd ede"
error: invalid label value: "hello.io/he=abcd ede": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')

$ k label node minikube "hello.io/hn dsfds"="abcdede"
The Node "minikube" is invalid: metadata.labels: Invalid value: "hello.io/hn dsfds": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')
```

Rakshith-R:
Moved stuff around a bit as niels suggested ,simplified the code changes and tested it.
ptal

Madhu-1:
@Rakshith-R CI failure

Rakshith-R:
> @Rakshith-R CI failure

Thanks, fixed it.

nixpanic (approved):
Thanks, much cleaner than the previous versions of the PR :clap:

Madhu-1 (approved):

nixpanic:
> /retest ci/centos/mini-e2e/k8s-1.25

Logs are at https://jenkins-ceph-csi.apps.ocp.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.25/detail/mini-e2e_k8s-1.25/259/pipeline

Rakshith-R:
@mergifyio refresh

PR #3651, opened 2023-02-07
cephfs: use shallow volumes for the ROX accessMode

### LINKED ISSUE #3603: cephfs: use shallow volumes for the ROX accessModes

This is not a new functionality rather asking to make the [cephfs shallow volumes](https://github.com/ceph/ceph-csi/blob/devel/docs/cephfs-snapshot-backed-volumes.md) as default feature for ROX volumes.

Currently ROX volumes need to be created with a new storageclass. This is not very useful for DR usecase. In a cluster with many subvolumesgroup and filesystem, we cannot ask admin/user to create exact copy of storageclasses with extra parameter `backingSnapshot: "true"`. In most of the backup system/Data movers the operators/admin will create a clones from a snapshot and move the data from source to the destination. In most of the cases the source volumes is created only for the read access to move the data.

The end user need not to care about creating a new storageclass to get shallow volumes feature.

Current cephfs clone is not something useful in a cluster with many volumes and subvolumes. Even as its a ROX accessMode 
the user is asking to give a volume with a readonly access. Instead of creating a clones (full copy) like today we can just create a shallow volume and give it to the user.

@gman0 @ShyamsundarR @BenamarMk @humblec @nixpanic any thoughts?

riya-singhal31:
this commit makes shallow volume as default feature for ROX volumes.

Signed-off-by: riya-singhal31 <rsinghal@redhat.com>

Fixes: #3603


internal/cephfs/store/volumeoptions.go
```diff
@@ -235,6 +235,7 @@ func NewVolumeOptions(
 	opts.Monitors = strings.Join(clusterData.Monitors, ",")
 	opts.SubvolumeGroup = clusterData.CephFS.SubvolumeGroup
 	opts.Owner = k8s.GetOwner(volOptions)
+	opts.BackingSnapshot = true
```
Madhu-1: This will impact the volume creation for other access modes, the Request is to support backingSnapshot for only ROX modes. For other access modes, we should create a subvolume.

Madhu-1 (changes requested):

internal/cephfs/store/volumeoptions.go
```diff
@@ -236,6 +236,16 @@ func NewVolumeOptions(
 	opts.SubvolumeGroup = clusterData.CephFS.SubvolumeGroup
 	opts.Owner = k8s.GetOwner(volOptions)
 
+	// backingSnapshot will be used only with read-only access modes
+	volCaps := req.GetVolumeCapabilities()
```
nixpanic: Could you make this a helper function so that testing is easier?

```go
    // default BackingSnapshot to true only for ReadOnly volume requests
    opt.BackingSnapshot = isVolumeCreateRO(req)
    ...
}

func isVolumeCreateRO(req *csi.CreateVolumeRequest) bool {
    ...
}
```

That way, you can add a unit test for `isVolumeCreateRO()`, document it's behaviour and reduce the complexity of `NewVolumeOptions()`.

internal/cephfs/store/volumeoptions.go
```diff
@@ -323,6 +325,19 @@ func NewVolumeOptions(
 	return &opts, nil
 }
 
+func isVolumeCreateRO(req *csi.CreateVolumeRequest) bool {
```
nixpanic: can you add a unit-tests in `internal/cephfs/store/volumeoptions_test.go` for this function?

riya-singhal31: Yes, sure. I'll update the PR.

internal/cephfs/store/volumeoptions_test.go
```diff
@@ -0,0 +1,82 @@
+package store
```
nixpanic, line 17: please include the header as in other files in this directory

riya-singhal31: Thanks, updated.

internal/cephfs/store/volumeoptions.go
```diff
@@ -235,6 +235,8 @@ func NewVolumeOptions(
 	opts.Monitors = strings.Join(clusterData.Monitors, ",")
 	opts.SubvolumeGroup = clusterData.CephFS.SubvolumeGroup
 	opts.Owner = k8s.GetOwner(volOptions)
+	// default BackingSnapshot to true only for ReadOnly volume requests
+	opts.BackingSnapshot = IsVolumeCreateRO(req.VolumeCapabilities)
```
Rakshith-R: The datasource needs to be a snapshot for this to work.

Rakshith-R (changes requested):

internal/cephfs/store/volumeoptions.go
riya-singhal31: Here we are just specifying the value for backing snapshot based on the access mode of volume.
The data source should be handled by user, while creating the yaml of volume. right?

Rakshith-R: > Here we are just specifying the value for backing snapshot based on the access mode of volume. The data source should be handled by user, while creating the yaml of volume. right?

This will cause a error when data source is pvc.

We should default backingSnapshot option to true if the datasource is a snapshot and request is for a RO volume.

riya-singhal31: Thanks Rakshith, updated.

internal/cephfs/store/volumeoptions.go
```diff
@@ -323,6 +327,32 @@ func NewVolumeOptions(
...
+				csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY:
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func IsDataSourceSnapshot(volumeSource *csi.VolumeContentSource) bool {
```
Rakshith-R: VolumeSource can itself be nil too,
for a bare createPVC, this can lead to a panic here

e2e/cephfs.go
```diff
@@ -1794,8 +1918,7 @@ var _ = Describe(cephfsType, func() {
 					framework.Failf("failed to create PVC and app: %v", err)
 				}
 
-				// Snapshot-backed volume shouldn't contribute to total subvolume count.
-				validateSubvolumeCount(f, 1, fileSystemName, subvolumegroup)
+				validateSubvolumeCount(f, 2, fileSystemName, subvolumegroup)
 
 				// Deleting snapshot before deleting pvcClone should succeed. It will be
 				// deleted once all volumes that are backed by this snapshot are gone.
```
Rakshith-R, line 1924:
```suggestion


```

check snapshot count right after deletesnapshoy

Rakshith-R (dismissed):
Thanks !

e2e/cephfs.go
```diff
@@ -1617,8 +1617,8 @@ var _ = Describe(cephfsType, func() {
 						if err != nil {
 							framework.Failf("failed to delete CephFS storageclass: %v", err)
 						}
+
```
yati1998, line 1620: Is this extra line required? IMO, you can remove this.

yati1998 (changes requested):
@riya-singhal31 , just a small nit. else looks good to me.

Madhu-1 (dismissed):

Rakshith-R:
@mergifyio rebase

Rakshith-R:
@yati1998 Can you re-review this pr ?

yati1998 (dismissed):

Rakshith-R:
@mergifyio rebase

riya-singhal31:
@Rakshith-R, @Madhu-1, @yati1998, @nixpanic  Can you please re-review the PR.

Madhu-1 (approved):

nixpanic (approved):

yati1998 (approved):

Madhu-1:
> /retest all

dont use restart all it will start many unwanted CI jobs

docs/deploy-cephfs.md
```diff
@@ -70,7 +70,7 @@ you're running it inside a k8s cluster and find the config itself).
...
 | `snapshotNamePrefix`                                                                                | no             | Prefix to use for naming snapshots (defaults to `csi-snap-`)                                                                                                                                                            |
-| `backingSnapshot`                                                                                   | no             | Boolean value. The PVC shall be backed by the CephFS snapshot specified in its data source. `pool` parameter must not be specified. (defaults to `false`)                                                               |
+| `backingSnapshot`                                                                                   | no             | Boolean value. The PVC shall be backed by the CephFS snapshot specified in its data source. `pool` parameter must not be specified. (defaults to `true`)                                                               |
```
haslersn, line 73: @riya-singhal31 Here you say the default is `true`. However, I looked a bit at the code and it seems there are 3 options:

- `backingSnapshot` is `true`, then the volume is backed by a snapshot or, if that's not possible, an error is returned.
- `backingSnapshot` is `false`, then the volume is never backed by a snapshot.
- `backingSnapshot` is not set, then the volume is backed by a snapshot if supported (i.e., `IsShallowVolumeSupported(req)` returns true.

Is my interpretation of the source code correct? If yes, then you cannot say that the default is `true`, because the default behavior deviates from the behavior in the `true`/`false` cases.

PR #3653, opened 2023-02-07
rename configmaps to match the csi type (rbd or cephfs), fix for 3652

### LINKED ISSUE #3652: deploying rbd and cephfs fails due to same configmap name
# Describe the bug #

When deploying with helm both rbd and cephfs charts, the configmap ceph-config has the same name so the chart deployed after the other fails as it cannot use the configmap name :

> Helm install failed: rendered manifests contain a resource that already exists. Unable to continue with install: ConfigMap "ceph-config" in namespace "storage" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "ceph-csi-cephfs": current value is "ceph-csi-rbd"

# Steps to reproduce #

deploy both helm chart for rbd and cephfs
the cephfs chart deployment fails due to the above error.

# Expected behavior #

both helm chart deployment should succeed

kifeo:
rename configmaps to match the csi type (rbd or cephfs)

This fixes bug https://github.com/ceph/ceph-csi/issues/3652

## Related issues ##

Fixes: #3652 

---

<details>

  progress or results

</details>

### DIFF
--- a/charts/ceph-csi-cephfs/values.yaml
+++ b/charts/ceph-csi-cephfs/values.yaml
@@ -314,10 +314,10 @@ kubeletDir: /var/lib/kubelet
 # Name of the csi-driver
 driverName: cephfs.csi.ceph.com
 # Name of the configmap used for state
-configMapName: ceph-csi-config
+configMapName: ceph-csi-config-cephfs
 # Key to use in the Configmap if not config.json
 # configMapKey:
 # Use an externally provided configmap
 externallyManagedConfigmap: false
 # Name of the configmap used for ceph.conf
-cephConfConfigMapName: ceph-config
+cephConfConfigMapName: ceph-config-cephfs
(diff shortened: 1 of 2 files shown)


Madhu-1:
@kifeo Please have a look at https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#code-contribution-workflow to fix CI errors.

charts/ceph-csi-cephfs/values.yaml
```diff
@@ -314,10 +314,10 @@ kubeletDir: /var/lib/kubelet
...
 # Name of the configmap used for state
-configMapName: ceph-csi-config
+configMapName: ceph-csi-config-cephfs
 # Key to use in the Configmap if not config.json
 # configMapKey:
 # Use an externally provided configmap
 externallyManagedConfigmap: false
 # Name of the configmap used for ceph.conf
-cephConfConfigMapName: ceph-config
+cephConfConfigMapName: ceph-config-cephfs
```
Madhu-1, line 323: This is a problem with the upgrade change. We need to check what are the problems by changing the default names.

Issue #3397, opened 2022-09-19
fsGroupPolicy field is immutable


kub3let:
# Describe the bug #

I'm trying to upgrade a cluster running Ceph CSI 3.7.0 to 3.7.1

# Environment details #

- Image/version of Ceph CSI driver : 3.7.0
- Helm chart version : 3.7.0
- Kernel version : 5.10.0-17-amd64
- Mounter used for mounting PVC (for cephFS its `fuse` or `kernel`. for rbd its
  `krbd` or `rbd-nbd`) : n/a
- Kubernetes cluster version : k3s v1.24.4+k3s1
- Ceph cluster version : 16.2.7

# Steps to reproduce #

Setup cluster with Ceph CSI 3.7.0 use helm to upgrade to 3.7.1

# Actual results #

Helm upgrade fails with message:

UPGRADE FAILED: cannot patch \"rbd.csi.ceph.com\" with kind CSIDriver: CSIDriver.storage.k8s.io \"rbd.csi.ceph.com\" is invalid: spec.fsGroupPolicy: Invalid value: \"File\": field is immutable"

```
 FAILED! => {"changed": false, "command": "/usr/sbin/helm --version=v3.7.1 upgrade -i --reset-values --create-namespace -f=/tmp/tmp2sr2me9a.yml ceph-csi ceph-csi/ceph-csi-rbd", "msg": "Failure when executing Helm command. Exited 1.\nstdout: \nstderr: Error: UPGRADE FAILED: cannot patch \"rbd.csi.ceph.com\" with kind CSIDriver: CSIDriver.storage.k8s.io \"rbd.csi.ceph.com\" is invalid: spec.fsGroupPolicy: Invalid value: \"File\": field is immutable\n", "stderr": "Error: UPGRADE FAILED: cannot patch \"rbd.csi.ceph.com\" with kind CSIDriver: CSIDriver.storage.k8s.io \"rbd.csi.ceph.com\" is invalid: spec.fsGroupPolicy: Invalid value: \"File\": field is immutable\n", "stderr_lines": ["Error: UPGRADE FAILED: cannot patch \"rbd.csi.ceph.com\" with kind CSIDriver: CSIDriver.storage.k8s.io \"rbd.csi.ceph.com\" is invalid: spec.fsGroupPolicy: Invalid value: \"File\": field is immutable"], "stdout": "", "stdout_lines": []}
```

# Expected behavior #

upgrade should succeed ?

# Additional context #

[...]

PR #3655, opened 2023-02-07
doc: Add basic upgrade documentation for Helm Charts

Syphdias:
Without this patch the READMEs for the Helm Charts do not provide any documentation on how to upgrade to a newer version. There is at least one known issue when updating to a newer versions that is unavoidable as of writing. There is a workaround for the issue which should be documented in the upgrade section.

This is a problem because currently the only way to find this workaround is to go through closed GitHub issues. These might not be around at the time someone needs this information. Furthermore the issue should be communicated to the operator before it occurs.

This patch adds basic documentation for updating the Helm repository, and upgrading the installed release of the Helm Chart. How values can be set is not part of the documentation. If an operator used custom values, e.g. for the secret, they probably already know how to deal with setting values. However, the docs still remind the reader to take values into account.
Reusing the installed values (`--reuse-values`) has lead to problems in past, which is why it is explicitly discouraged. An example for this would be the value `logLevel` which was changed to `sidecarLogLevel`. Reusing values lead to `.Values.sidecarLogLevel` being empty and the `csi-provisioner` not being started due to invalid value `-v=""`. Comparing new values with set values is encouraged.

The workaround for issue #3397 from GitHub is being addressed in the section Know Issues Upgrading.

Signed-off-by: Christian Kugler <syphdias+git@gmail.com>

~~Do you have any questions?~~
~~Is the change backward compatible?~~

I am not very familiar with Helm 2.x. In this case I think the commands for helm 3.x work just the same on 2.x. Some confirmation on that would be nice.

~~Provide any external context for the change, if any.~~

## Related issues ##
#3397

### DIFF
--- a/charts/ceph-csi-cephfs/README.md
+++ b/charts/ceph-csi-cephfs/README.md
@@ -50,6 +50,36 @@ After installation succeeds, you can get a status of Chart
 helm status "ceph-csi-cephfs"
 ```
 
+### Upgrade Chart
+
+If you want to upgrade your Chart, use the following commands.
+
+```bash
+helm repo update ceph-csi
+helm upgrade --namespace ceph-csi-cephfs ceph-csi-cephfs ceph-csi/ceph-csi-cephfs
+```
+
+For upgrading to a specific version, provide the flag `--version` and the
+version.
+
+**Do not forget to include your values**, if they differ from the default values.
+We recommend not to use `--reuse-values` in case there are new defaults AND
+compare your currently used values with the new default values.
+
+#### Known Issues Upgrading
+
+- When upgrading to version >=3.7.0, you might encounter an error that the
+  CephFS CSI Driver cannot be updated. Please refer to
+  [issue](https://github.com/ceph/ceph-csi/issues/3397) for more details.
+  This is due to the CSIDriver resource not being updatable. To work around this
+  you can delete the CSIDriver object by running:
+
+  ```bash
+  kubectl delete csidriver cephfs.csi.ceph.com
+  ```
+
+  Then rerun your `helm upgrade` command.
+
 ### Delete Chart
 
 If you want to delete your Chart, use this command
--- a/charts/ceph-csi-rbd/README.md
+++ b/charts/ceph-csi-rbd/README.md
@@ -50,6 +50,36 @@ After installation succeeds, you can get a status of Chart
 helm status "ceph-csi-rbd"
 ```
 
+### Upgrade Chart
+
+If you want to upgrade your Chart, use the following commands.
+
+```bash
+helm repo update ceph-csi
+helm upgrade --namespace ceph-csi-rbd ceph-csi-rbd ceph-csi/ceph-csi-rbd
+```
+
+For upgrading to a specific version, provide the flag `--version` and the
+version.
+
+**Do not forget to include your values**, if they differ from the default values.
+We recommend not to use `--reuse-values` in case there are new defaults AND
+compare your currently used values with the new default values.
+
+#### Known Issues Upgrading
+
+- When upgrading to version >=3.7.0, you might encounter an error that the
+  RBD CSI Driver cannot be updated. Please refer to
+  [issue](https://github.com/ceph/ceph-csi/issues/3397) for more details.
+  This is due to the CSIDriver resource not being updatable. To work around this
+  you can delete the CSIDriver object by running:
+
+  ```bash
+  kubectl delete csidriver rbd.csi.ceph.com
+  ```
+
+  Then rerun your `helm upgrade` command.
+
 ### Delete Chart
 
 If you want to delete your Chart, use this command


charts/ceph-csi-cephfs/README.md
```diff
@@ -50,6 +50,32 @@ After installation succeeds, you can get a status of Chart
...
+For upgrading to a specific version, provide the flag `--version` and the
+version.
+
+**Do not forget to include your values**, if they differ from the default values.
+We recommend not to use `--reuse-values` in case there are new defaults AND
+compare your currently used values with the new default values.
+
+#### Known Issues Upgrading
+- When upgrading to version >=3.7.0, you might encounter an error that the
+  CSIDriver cannot be updated (#3397). This is due to the CSIDriver resource not
```
Madhu-1: Provide a link to the issue instead of just the issue number.

charts/ceph-csi-cephfs/README.md
```diff
@@ -50,6 +50,32 @@ After installation succeeds, you can get a status of Chart
...
+  being updatable. To work around this you can delete the CSIDriver object by
+  running:
+  ```bash
+  kubectl delete csidriver rbd.csi.ceph.com
```
Madhu-1: cephfs.csi.ceph.com

Madhu-1 (commented):
Thanks, @Syphdias, for taking this up.

Madhu-1:
Please refer to https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#code-contribution-workflow to fix CI issue.

charts/ceph-csi-rbd/README.md
```diff
@@ -50,6 +50,32 @@ After installation succeeds, you can get a status of Chart
...
+For upgrading to a specific version, provide the flag `--version` and the
+version.
+
+**Do not forget to include your values**, if they differ from the default values.
+We recommend not to use `--reuse-values` in case there are new defaults AND
+compare your currently used values with the new default values.
+
+#### Known Issues Upgrading
+- When upgrading to version >=3.7.0, you might encounter an error that the
+  CSIDriver cannot be updated (#3397). This is due to the CSIDriver resource not
```
humblec: CSIDriver can be mentioned as `CSI Driver` or even mention  as `RBD CSI Driver`

Syphdias: I went for `CSIDriver` because this is what the k8s resource is named – or `csidriver.storage.k8s.io` in its longer form. https://kubernetes-csi.github.io/docs/csi-driver-object.html

I'll change it in the first instance but in the next two instances I explicitly mean the k8s resource/object which uses the concatenated spelling. Is this acceptable?

charts/ceph-csi-cephfs/README.md
```diff
@@ -50,6 +50,34 @@ After installation succeeds, you can get a status of Chart
...
+version.
+
+**Do not forget to include your values**, if they differ from the default values.
+We recommend not to use `--reuse-values` in case there are new defaults AND
+compare your currently used values with the new default values.
+
+#### Known Issues Upgrading
+
+- When upgrading to version >=3.7.0, you might encounter an error that the
+  CephFS CSI Driver cannot be updated (https://github.com/ceph/ceph-csi/issues/3397).
```
Madhu-1:
```suggestion
CephFS CSI Driver cannot be updated. Please refer to [issue](https://github.com/ceph/ceph-csi/issues/3397) for more details.
```

Madhu-1 (commented):
small nit, change look good to me.

charts/ceph-csi-cephfs/README.md
Syphdias: I also used a bare URL, which I need to fix.

Syphdias: I assume the dedent was an accident?

Madhu-1: Yes correct , suggestion is to use link instead of just issue URL

Syphdias: Rerunning `make containerized-test TARGET=lint-extras` before pushing this time… – hopefully for the last time :smile:

Syphdias: `2 chart(s) linted, 0 chart(s) failed` :tada:

humblec (approved):

charts/ceph-csi-cephfs/README.md
Madhu-1: Thanks for testing it locally :100:

Madhu-1 (approved):

charts/ceph-csi-cephfs/README.md
Syphdias: I could have avoided half the comments, if I did in the first place

nixpanic:
@mergifyio queue

nixpanic:
@mergifyio unqueue

This somehow has the `DNM` label?

PR #3693, opened 2023-02-24
kms: Add semaphore to limit concurrency

### LINKED ISSUE #3472: RBD: OOMKills occurs when secret metadata encryption type is used with multiple PVC create request.
I tested secret based encryption with 3.7.1 i dont see any crash with below limits

```
Limits:
      cpu:     500m
      memory:  256Mi
    Requests:
      cpu:     250m
      memory:  256Mi
```
```
[🎩︎]mrajanna@fedora rbd $]kubectl get pvc
NAME      STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS      AGE
claim0    Bound    pvc-81cb47af-dfbd-4360-b444-7636e8a2c359   1Gi        RWO            rook-ceph-block   4m48s
claim1    Bound    pvc-c1c20239-db18-4c56-b9da-8745b0046428   1Gi        RWO            rook-ceph-block   4m47s
claim10   Bound    pvc-7d6e66d8-ceea-4e0c-9775-64aa84b1548b   1Gi        RWO            rook-ceph-block   4m46s
claim2    Bound    pvc-3a61744c-2d0e-46c1-9d8c-3b0f5f49574c   1Gi        RWO            rook-ceph-block   4m47s
claim3    Bound    pvc-aa2613fb-4db8-4254-801b-8d9d72e83979   1Gi        RWO            rook-ceph-block   4m47s
claim4    Bound    pvc-00fbe104-1809-4a11-8c39-9e3ceee9d5c9   1Gi        RWO            rook-ceph-block   4m47s
claim5    Bound    pvc-7ad36255-755b-4bdd-a88c-4bbf695e8b69   1Gi        RWO            rook-ceph-block   4m47s
claim6    Bound    pvc-bcd48a37-3d8d-47ce-b780-511020690397   1Gi        RWO            rook-ceph-block   4m47s
claim7    Bound    pvc-1fac5dcf-1668-489a-8799-16630b74e971   1Gi        RWO            rook-ceph-block   4m47s
claim8    Bound    pvc-949006fd-5b47-4e9c-acc1-3808894245f8   1Gi        RWO            rook-ceph-block   4m46s
claim9    Bound    pvc-05ab1ba5-4c8f-41a5-ab09-c042b6089b23   1Gi        RWO            rook-ceph-block   4m46s
```

but when i tested with metadata type encryption i can see the crash, this confirms we have a memory leak?

_Originally posted by @Madhu-1 in https://github.com/ceph/ceph-csi/issues/3402#issuecomment-1278691078_

lentzi90:
generateCipher is memory heavy, so to avoid OOM situations, a semaphore is added to limit concurrency here.

Fixes: #3472

NOTE: I think the original issue is already resolved (see discussion in the issue). I pushed this anyway because I had already done the work and hoping that it can be useful in other situations or just for discussion. Feel free to close!

---

<details>

  progress or results

</details>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,6 +47,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.14.4
 )
 
+require golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4
+
 require (
 	github.com/ansel1/merry v1.6.2 // indirect
 	github.com/ansel1/merry/v2 v2.0.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -1257,6 +1257,7 @@ golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9/go.mod h1:RxMgew5VJxzue5/jJ
 golang.org/x/sync v0.0.0-20201207232520-09787c993a3a/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sync v0.0.0-20210220032951-036812b2e83c/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
+golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4 h1:uVc8UZUe6tr40fFVnUP5Oj+veunVezqYl9z7DYw9xzw=
 golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sys v0.0.0-20180823144017-11551d06cbcc/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
 golang.org/x/sys v0.0.0-20180830151530-49385e6e1522/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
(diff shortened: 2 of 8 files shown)


go.mod
```diff
@@ -47,6 +47,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.14.4
 )
 
+require golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4
```
nixpanic, line 50: nit, this could be part of the above `require` block

nixpanic (commented):
I think this is good to have, it adds to the stability of ceph-csi when multiple volumes are created at the same time. Volume creation is not a critical part of a workflow when running applications, a slight delay while generating the cipher should be acceptable for increased stability.

nixpanic:
**commitlint** fails because `kms` is not a valid prefix for commits. Use `util` instead?

trociny:
@lentzi90 Is there any estimate how "heavy" the `generateCipher` may be?

Because from the comment [1] of the related issue, the oom-killer was invoked by `cryptsetup` process, which is called by `EncryptVolume`, i.e. after `generateCipher`. And it looks like the memory limit was reached due to spawning many `cryptsetup` processes simultaneously and this patch will not prevent this (although it might throttle it a little and make the problem more difficult to reproduce).

It seems to me that a solution could be to limit the number of  NodeStageVolume concurrent calls. Looking at the code, I suppose it is in this loop [2], where we spawn all NodeStageVolume calls concurrently [3] and then just wait for them to complete. I think it could do it in configurable chunks, e.g. not more than 10 calls simultaneously.

[1] https://github.com/ceph/ceph-csi/issues/3472#issuecomment-1442051527
[2] https://github.com/ceph/ceph-csi/blob/991c21f7fdeacef76ac08911d219435d89a1c119/internal/rbd/rbd_healer.go#L179
[3] https://github.com/ceph/ceph-csi/blob/991c21f7fdeacef76ac08911d219435d89a1c119/internal/rbd/rbd_healer.go#L219

lentzi90:
@trociny like I mentioned on the issue already, I think this is a different issue. I just pushed the code that I already had for the original issue (which seems to be fixed already).
If you (or anyone else) would like to take this over and push a new PR I would be very happy. Unfortunately I won't have time to work on this in the near time.

PR #3716, opened 2023-03-13
doc: add additional provisioner name

riya-singhal31:
this commit adds another provisioner name, which is deployed via rook.


Madhu-1 (changes requested):
we dont need this comment, as this examples are specific to standalone cephcsi deployment

riya-singhal31:
> we dont need this comment, as this examples are specific to standalone cephcsi deployment

Thanks for the review Madhu, closing the PR in that case.

PR #3782, opened 2023-04-26
doc: use the Ceph Slack instance and not our silo'ed own one

nixpanic:
Currently the Ceph-CSI community is on the 'free' Slack instance at https://cephcsi.slack.com. The Ceph project uses a Slack instance that we can use for Ceph-CSI as well. In order to integrate more with other Ceph projects, we should ideally be active on the same Slack instance.

For now, we have `#ceph-csi` as only channel on the https://ceph-storage-slack.com, we can add more channels if needed.

See-also: https://ceph.io/en/community/connect/

---

<details>

  progress or results

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,8 +198,8 @@ welcome and encouraged to join.
 
 Please use the following to reach members of the community:
 
-- Slack: Join our [slack channel](https://cephcsi.slack.com) to discuss
-  anything related to this project. You can join the slack by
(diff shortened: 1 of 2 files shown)


yati1998 (approved):

karthik-us (approved):

Madhu-1 (approved):

humblec:
@nixpanic we also need to notify the existing users in the current instance to migrate..etc

PR #3854, opened 2023-05-24
fix: Invalid "invalid encryption kms configuration" error

### LINKED ISSUE #3759: Invalid "invalid encryption kms configuration" error
# Describe the bug #

After upgrading to v3.8.0, the following error appeared whereas I did not configure encryption:

```
csi-rbdplugin E0419 07:35:12.775698    6851 utils.go:210] ID: 69 Req-ID: 0001-0009-rook-ceph-0000000000000001-4aef4df9-b4a3-11ed-9485-b2ff204cea03 GRPC error: rpc error: code = Internal desc = invalid encryption kms configuration: missing encryption passphrase in secrets
```

Rolling back to v3.7.0 solved the issue.

This might be due to dc7ba684e3ba16a0f62569f815be5261162432e3 which removed and `else` clause.
So before (in v3.7.0) the `initKMS` procedure was interrupted when no specific KMS configuration is present (which is my case).
Now (in v3.8.0) a non-nil error is returned: 

```
diff --git a/internal/rbd/encryption.go b/internal/rbd/encryption.go
index c72113288..ea65f14aa 100644
--- a/internal/rbd/encryption.go
+++ b/internal/rbd/encryption.go
@@ -309,8 +309,6 @@ func (ri *rbdImage) initKMS(ctx context.Context, volOptions, credentials map[str
        kmsID, encType, err := ParseEncryptionOpts(ctx, volOptions, rbdDefaultEncryptionType)
        if err != nil {
                return err
-       } else if kmsID == "" {
-               return nil
        }
```

# Environment details #

- Image/version of Ceph CSI driver : v3.8.0
- Kernel version : 5.4.0
  `krbd` or `rbd-nbd`) : krbd
- Kubernetes cluster version : 1.25.4
- Ceph cluster version : 17.2.6

# Steps to reproduce #

1. Have the following StorageClass configuration:

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    meta.helm.sh/release-name: rook-ceph-cluster
    meta.helm.sh/release-namespace: rook-ceph
    storageclass.kubernetes.io/is-default-class: "false"
  creationTimestamp: "2023-02-23T16:11:55Z"
  labels:
    app.kubernetes.io/managed-by: Helm
  name: ceph-block
  resourceVersion: "795634683"
  uid: fb7f65ef-0bff-49a0-bb52-5699bde7ce92
parameters:
  clusterID: rook-ceph
  csi.storage.k8s.io/controller-expand-secret-name: rook-csi-rbd-provisioner
  csi.storage.k8s.io/controller-expand-secret-namespace: rook-ceph
  csi.storage.k8s.io/fstype: ext4
  csi.storage.k8s.io/node-stage-secret-name: rook-csi-rbd-node
  csi.storage.k8s.io/node-stage-secret-namespace: rook-ceph
  csi.storage.k8s.io/provisioner-secret-name: rook-csi-rbd-provisioner
  csi.storage.k8s.io/provisioner-secret-namespace: rook-ceph
  encrypted: "false"
  imageFeatures: layering
  imageFormat: "2"
  pool: ceph-blockpool
provisioner: rook-ceph.rbd.csi.ceph.com
reclaimPolicy: Retain
volumeBindingMode: Immediate
```
2. Create a pods with a PVC using this storage class

# Actual results #

The error about an "invalid encryption kms" is raised (as described above)

# Expected behavior #

No error while mounting Ceph image

# Logs #

"csi-rbdplugin" logs:

```
W0419 07:32:17.161034    6851 rbd_attach.go:241] kernel version "5.4.0-137-generic" doesn't support cookie feature
E0419 07:33:10.655993    6851 utils.go:210] ID: 23 Req-ID: 0001-0009-rook-ceph-0000000000000001-4aef4df9-b4a3-11ed-9485-b2ff204cea03 GRPC error: rpc error: code = Internal desc = invalid encryption kms configuration: missing encryption passphrase in secrets
```

riya-singhal31:
this commit adds the validation for encryption
value as false, and sets the type as none

Fixes: #3759

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -2872,6 +2872,56 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("create storageClass with encrypted as false", func() {
+				err := deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					framework.Failf("failed to delete storageclass: %v", err)
+				}
+				err = createRBDStorageClass(
+					f.ClientSet,
+					f,
+					defaultSCName,
+					nil,
+					map[string]string{"encrypted": "false"},
+					deletePolicy)
+				if err != nil {
+					framework.Failf("failed to create storageclass: %v", err)
+				}
+				// set up PVC
+				pvc, err := loadPVC(pvcPath)
+				if err != nil {
+					framework.Failf("failed to load PVC: %v", err)
+				}
+				pvc.Namespace = f.UniqueName
+				err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					framework.Failf("failed to create PVC: %v", err)
+				}
+
+				// validate created backend rbd images
+				validateRBDImageCount(f, 1, defaultRBDPool)
+				validateOmapCount(f, 1, rbdType, defaultRBDPool, volumesType)
+
+				// clean up after ourselves
+				err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					framework.Failf("failed to  delete PVC: %v", err)
+				}
+				// validate created backend rbd images
+				validateRBDImageCount(f, 0, defaultRBDPool)
+				validateOmapCount(f, 0, rbdType, defaultRBDPool, volumesType)
+
+				err = deleteResource(rbdExamplePath + "storageclass.yaml")
+				if err != nil {
+					framework.Failf("failed to delete storageclass: %v", err)
+				}
+				err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, nil, deletePolicy)
+				if err != nil {
+					framework.Failf("failed to create storageclass: %v", err)
+				}
+			})
+
+
 			By("validate RBD static FileSystem PVC", func() {
 				err := validateRBDStaticPV(f, appPath, false, false)
 				if err != nil {
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -2921,7 +2921,6 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
-
 			By("validate RBD static FileSystem PVC", func() {
 				err := validateRBDStaticPV(f, appPath, false, false)
 				if err != nil {
--- a/internal/rbd/encryption.go
+++ b/internal/rbd/encryption.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"strconv"
 
 	kmsapi "github.com/ceph/ceph-csi/internal/kms"
 	"github.com/ceph/ceph-csi/internal/util"
@@ -341,7 +342,8 @@ func ParseEncryptionOpts(
 		encrypted, kmsID string
 	)
 	encrypted, ok = volOptions["encrypted"]
-	if !ok {
+	val, _ := strconv.ParseBool(encrypted)
+	if !ok || !val{
 		return "", util.EncryptionTypeNone, nil
 	}
 	kmsID, err = util.FetchEncryptionKMSID(encrypted, volOptions["encryptionKMSID"])
--- a/internal/rbd/encryption.go
+++ b/internal/rbd/encryption.go
@@ -110,7 +110,7 @@ func (ri *rbdImage) isFileEncrypted() bool {
 }
 
 func IsFileEncrypted(ctx context.Context, volOptions map[string]string) (bool, error) {
-	_, encType, err := ParseEncryptionOpts(ctx, volOptions, util.EncryptionTypeInvalid)
+	_, encType, err := ParseEncryptionOpts(volOptions, util.EncryptionTypeInvalid)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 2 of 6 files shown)


internal/rbd/encryption.go
```diff
@@ -341,7 +342,8 @@ func ParseEncryptionOpts(
 		encrypted, kmsID string
 	)
 	encrypted, ok = volOptions["encrypted"]
-	if !ok {
+	val, _ := strconv.ParseBool(encrypted)
```
Madhu-1: I'm not sure why parsing is necessary when the encrypted key isn't configured

Madhu-1 (changes requested):
Please add an E2E for it.

riya-singhal31:
> Please add an E2E for it.

Yes, adding an unit test for it.

internal/rbd/encryption.go
nixpanic: No need to parse in case `!ok`, but it is possible that `encrypted: false` is set. Checking the boolean value of `encrypted` is important.

Maybe you prefer something like this?

```go
encrypted, ok = volOptions["encrypted"]
if !ok {
    return ....
} else if val, _ := strconv.ParseBool(encrypted): !val {
    return ...
}
```

I am not in favor of two the same `return` statement though. Converting the `encrypted` string to a boolean, even when not required, is not a huge performance loss, and I think the code will be cleaner.

internal/rbd/encryption_test.go
```diff
@@ -0,0 +1,86 @@
...
+			},
+			fallbackType: util.EncryptionTypeBlock,
+			expectedKMS:  "",
+			expectedEnc:  util.EncryptionTypeNone,
+			expectedErr:  nil,
+		},
+		{
+			testName: "Valid Encryption Option With KMS ID",
+			volOptions: map[string]string{
+				"encrypted":       "true",
```
nixpanic, line 68: can you add an other test with `encrypted` set to an invalid string (not `true` or `false`)?

riya-singhal31: Done, thanks!

Madhu-1:
@riya-singhal31 CI problem, you can run the tests locally by running commands like `CONTAINER_CMD=docker make containerized-test TARGET=go-lint` you just need to adjust the TARGET to run different tests

internal/rbd/encryption.go
riya-singhal31: We will need to follow the 2nd, so as to handle the case when the encrypted is not configured.

riya-singhal31:
> @riya-singhal31 CI problem, you can run the tests locally by running commands like `CONTAINER_CMD=docker make containerized-test TARGET=go-lint` you just need to adjust the TARGET to run different tests

Sure, thanks Madhu.

e2e/rbd.go
```diff
@@ -2872,6 +2872,56 @@ var _ = Describe("RBD", func() {
 				}
 			})
 
+			By("create storageClass with encrypted as false", func() {
+				encrypted := "false"
```
Madhu-1: no need for a new variable just set this in map at line 2886

internal/rbd/encryption.go
```diff
@@ -344,6 +344,13 @@ func ParseEncryptionOpts(
 	if !ok {
 		return "", util.EncryptionTypeNone, nil
 	}
+	val, er := strconv.ParseBool(encrypted)
+	if er != nil {
+		return "", util.EncryptionTypeInvalid, fmt.Errorf("strconv.ParseBool: parsing %s: invalid syntax", encrypted)
```
Madhu-1: Instead of defining our own error just return the error `er` here?

internal/rbd/encryption_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestParseEncryptionOpts(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		testName     string
+		volOptions   map[string]string
+		fallbackType util.EncryptionType
+		expectedKMS  string
+		expectedEnc  util.EncryptionType
+		expectedErr  string
```
Madhu-1: Instead of a string, you can just check error is expected or not with a bool value here.

riya-singhal31: Thanks for the review, Madhu.
Updated the PR with mentioned changes.

Madhu-1 (dismissed):

internal/rbd/encryption.go
```diff
@@ -341,8 +341,14 @@ func ParseEncryptionOpts(
 		encrypted, kmsID string
 	)
 	encrypted, ok = volOptions["encrypted"]
-	val, _ := strconv.ParseBool(encrypted)
-	if !ok || !val{
+	if !ok {
+		return "", util.EncryptionTypeNone, nil
+	}
+	val, er := strconv.ParseBool(encrypted)
```
nixpanic: you can re-use both `ok` and `err` variables here. No need to introduce two new ones.

nixpanic (changes requested):
just a small nit to make the code a little cleaner

riya-singhal31:
> just a small nit to make the code a little cleaner

Thanks for the suggestion Niels, updated.

nixpanic (approved):

riya-singhal31:
Hi @Madhu-1 ,
With the recent commit, the approval got removed. So, requesting you to please approve it again if you find it good.

Madhu-1 (approved):

nixpanic:
@mergifyio queue

riya-singhal31:
cherry-pick release-4.13

riya-singhal31:
/cherry-pick-release-4.13

riya-singhal31:
git cherry-pick -4.13

Issue #3314, opened 2022-08-18
rbd: complete migration of volrep to csi-addons


Rakshith-R:
TODO's:

- [x] move logic to `internal/csi-addons/rbd`
- [x] address [comment](https://github.com/ceph/ceph-csi/pull/3608/files#r1092953730)
- [x] remove replication service from controller server
- [x] add minimal unit tests like `internal/csi-addons/rbd/reclaimspace_test.go`
- [x] https://github.com/ceph/ceph-csi/pull/3349

_Originally posted by @nixpanic in https://github.com/ceph/ceph-csi/issues/3305#issuecomment-1216245713_

yati1998:
@Rakshith-R I guess all the task here is done already. Can you confirm and close the issue, if done?

Rakshith-R:
> @Rakshith-R I guess all the task here is done already. Can you confirm and close the issue, if done?

No, 3 out of 4 TODOs are still not done.

@riya-singhal31  do you wanna take up this issue ?

yati1998:
> > @Rakshith-R I guess all the task here is done already. Can you confirm and close the issue, if done?
> 
> No, 3 out of 4 TODOs are still not done.
> cool
> @riya-singhal31 do you wanna take up this issue ?
> yeah would be a good one

riya-singhal31:
Closes with - https://github.com/ceph/ceph-csi/pull/3608

Rakshith-R:
> Closes with - #3608

you mean updates,

Make sure to address this [comment](https://github.com/ceph/ceph-csi/pull/3608/files#r1092953730) too.
three more tasks are remaining including the above one.

riya-singhal31:
>>add minimal unit tests like internal/csi-addons/rbd/reclaimspace_test.go 
There is already a test file for replication.go, moved it to csi-addons. Do we need to add some more tests to it?
cc @Rakshith-R

Rakshith-R:
> > add minimal unit tests like internal/csi-addons/rbd/reclaimspace_test.go

> There is already a test file for replication.go, moved it to csi-addons. Do we need to add some more tests to it?
> cc @Rakshith-R

They need to be added after completion of task 3 `remove replication service from controller server`
refer https://github.com/ceph/ceph-csi/blob/devel/internal/csi-addons/rbd/reclaimspace_test.go.

Madhu-1:
@riya-singhal31 any updates here?

riya-singhal31:
[...]

PR #3608, opened 2023-01-12
cleanup: migration of volrep to csi-addons

riya-singhal31:
This commit moves the volrep logic(replication controller server) from internal/rbd to internal/csi-addons/rbd.

Signed-off-by: riya-singhal31 <rsinghal@redhat.com>

PR #3884, opened 2023-06-05
rbd: remove grpc and csi-addons import from rbd/replication.go

riya-singhal31:
this commit removes grpc import from replication.go and replaced it with usual errors

updates - https://github.com/ceph/ceph-csi/issues/3314

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -326,7 +326,12 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 	case librbd.MirrorImageDisabling:
 		return nil, status.Errorf(codes.Aborted, "%s is in disabling state", volumeID)
 	case librbd.MirrorImageEnabled:
-		return corerbd.DisableVolumeReplication(rbdVol, mirroringInfo, force)
+		err = rbdVol.DisableVolumeReplication(mirroringInfo, force)
+		if err != nil {
+			return nil, getGRPCError(err)
+		}
+
+		return &replication.DisableVolumeReplicationResponse{}, nil
 	default:
 		return nil, status.Errorf(codes.InvalidArgument, "image is in %s Mode", mirroringInfo.State)
 	}
@@ -627,9 +632,7 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 
 	err = rbdVol.ResyncVol(localStatus, req.Force)
 	if err != nil {
-		log.ErrorLog(ctx, err.Error())
-
-		return nil, err
+		return nil, getGRPCError(err)
 	}
 
 	err = checkVolumeResyncStatus(localStatus)
@@ -647,6 +650,32 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	}
 
 	return resp, nil
+}
+
+func getGRPCError(err error) error {
+	if err == nil {
+		return status.Error(codes.OK, codes.OK.String())
+	}
+
+	errorStatusMap := map[error]codes.Code{
+		corerbd.ErrFetchingLocalState:          codes.Internal,
+		corerbd.ErrResyncImageFailed:           codes.Internal,
+		corerbd.ErrDisableImageMirroringFailed: codes.Internal,
+		corerbd.ErrFetchingMirroringInfo:       codes.Internal,
+		corerbd.ErrInvalidArgument:             codes.InvalidArgument,
+		corerbd.ErrAborted:                     codes.Aborted,
+		corerbd.ErrFailedPrecondition:          codes.FailedPrecondition,
+		corerbd.ErrUnavailable:                 codes.Unavailable,
+	}
+
+	for e, code := range errorStatusMap {
+		if errors.Is(err, e) {
+			return status.Error(code, err.Error())
+		}
+	}
+
+	// Handle any other non nil error not listed in the map
+	return status.Error(codes.Unknown, err.Error())
 }
 
 // GetVolumeReplicationInfo extracts the RBD volume information from the volumeID, If the
--- a/internal/csi-addons/rbd/replication_test.go
+++ b/internal/csi-addons/rbd/replication_test.go
@@ -18,6 +18,7 @@ package rbd
 
 import (
 	"context"
+	"errors"
 	"reflect"
 	"strings"
 	"testing"
@@ -27,6 +28,9 @@ import (
 
 	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/ceph/go-ceph/rbd/admin"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
@@ -491,6 +495,75 @@ func TestValidateLastSyncTime(t *testing.T) {
 			if !ts.AsTime().Equal(tt.timestamp.AsTime()) {
 				t.Errorf("getLastSyncTime() %v, expected %v", ts, tt.timestamp)
 			}
+		})
+	}
+}
+
+func TestGetGRPCError(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name        string
+		err         error
+		expectedErr error
+	}{
+		{
+			name:        "FetchingLocalStateFailed",
+			err:         corerbd.ErrFetchingLocalState,
+			expectedErr: status.Error(codes.Internal, corerbd.ErrFetchingLocalState.Error()),
+		},
+		{
+			name:        "ResyncImageFailed",
+			err:         corerbd.ErrResyncImageFailed,
+			expectedErr: status.Error(codes.Internal, corerbd.ErrResyncImageFailed.Error()),
+		},
+		{
+			name:        "DisableImageMirroringFailed",
+			err:         corerbd.ErrDisableImageMirroringFailed,
+			expectedErr: status.Error(codes.Internal, corerbd.ErrDisableImageMirroringFailed.Error()),
+		},
+		{
+			name:        "FetchingMirroringInfoFailed",
+			err:         corerbd.ErrFetchingMirroringInfo,
+			expectedErr: status.Error(codes.Internal, corerbd.ErrFetchingMirroringInfo.Error()),
+		},
+		{
+			name:        "InvalidArgument",
+			err:         corerbd.ErrInvalidArgument,
+			expectedErr: status.Error(codes.InvalidArgument, corerbd.ErrInvalidArgument.Error()),
+		},
+		{
+			name:        "Aborted",
+			err:         corerbd.ErrAborted,
+			expectedErr: status.Error(codes.Aborted, corerbd.ErrAborted.Error()),
+		},
+		{
+			name:        "FailedPrecondition",
+			err:         corerbd.ErrFailedPrecondition,
+			expectedErr: status.Error(codes.FailedPrecondition, corerbd.ErrFailedPrecondition.Error()),
+		},
+		{
+			name:        "Unavailable",
+			err:         corerbd.ErrUnavailable,
+			expectedErr: status.Error(codes.Unavailable, corerbd.ErrUnavailable.Error()),
+		},
+		{
+			name:        "InvalidError",
+			err:         errors.New("some error"),
+			expectedErr: status.Error(codes.Unknown, "some error"),
+		},
+		{
+			name:        "NilError",
+			err:         nil,
+			expectedErr: status.Error(codes.OK, "ok string"),
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			result := getGRPCError(tt.err)
+			assert.Equal(t, tt.expectedErr, result)
 		})
 	}
 }
--- a/internal/rbd/errors.go
+++ b/internal/rbd/errors.go
@@ -45,4 +45,22 @@ var (
 	// ErrLastSyncTimeNotFound is returned when last sync time is not found for
 	// the image.
 	ErrLastSyncTimeNotFound = errors.New("last sync time not found")
+	// ErrFailedPrecondition is returned when operation is rejected because the system is not in a state
+	// required for the operation's execution.
+	ErrFailedPrecondition = errors.New("system is not in a state required for the operation's execution")
+	// ErrUnavailable is returned when the image needs to be recreated
+	// locally and may be corrected by retrying with a backoff.
+	ErrUnavailable = errors.New("image needs to be recreated")
+	// ErrAborted is returned when the operation is aborted.
+	ErrAborted = errors.New("operation got aborted")
+	// ErrInvalidArgument is returned when the client specified an invalid argument.
+	ErrInvalidArgument = errors.New("invalid arguments provided")
+	// ErrFetchingLocalState is returned when the operation to fetch local state fails.
+	ErrFetchingLocalState = errors.New("failed to get local state")
+	// ErrDisableImageMirroringFailed is returned when the operation to disable image mirroring fails.
+	ErrDisableImageMirroringFailed = errors.New("failed to disable image mirroring")
+	// ErrFetchingMirroringInfo is returned when the operation to fetch mirroring info of image fails.
+	ErrFetchingMirroringInfo = errors.New("failed to get mirroring info of image")
+	// ErrResyncImageFailed is returned when the operation to resync the image fails.
+	ErrResyncImageFailed = errors.New("failed to resync image")
 )
--- a/internal/rbd/replication.go
+++ b/internal/rbd/replication.go
@@ -18,12 +18,10 @@ package rbd
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	librbd "github.com/ceph/go-ceph/rbd"
-	"github.com/csi-addons/spec/lib/go/replication"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 func (rv *rbdVolume) ResyncVol(localStatus librbd.SiteMirrorImageStatus, force bool) error {
@@ -31,19 +29,18 @@ func (rv *rbdVolume) ResyncVol(localStatus librbd.SiteMirrorImageStatus, force b
 		// If the force option is not set return the error message to retry
 		// with Force option.
 		if !force {
-			return status.Errorf(codes.FailedPrecondition,
-				"image is in %q state, description (%s). Force resync to recover volume",
-				localStatus.State, localStatus.Description)
+			return fmt.Errorf("%w: image is in %q state, description (%s). Force resync to recover volume",
+				ErrFailedPrecondition, localStatus.State, localStatus.Description)
 		}
 		err := rv.resyncImage()
 		if err != nil {
-			return status.Error(codes.Internal, err.Error())
+			return fmt.Errorf("%w: failed to resync image: %w", ErrResyncImageFailed, err)
 		}
 
 		// If we issued a resync, return a non-final error as image needs to be recreated
 		// locally. Caller retries till RBD syncs an initial version of the image to
 		// report its status in the resync request.
-		return status.Error(codes.Unavailable, "awaiting initial resync due to split brain")
+		return fmt.Errorf("%w: awaiting initial resync due to split brain", ErrUnavailable)
 	}
 
 	return nil
@@ -85,10 +82,10 @@ func resyncRequired(localStatus librbd.SiteMirrorImageStatus) bool {
 	return false
 }
 
-func DisableVolumeReplication(rbdVol *rbdVolume,
+func (rv *rbdVolume) DisableVolumeReplication(
 	mirroringInfo *librbd.MirrorImageInfo,
 	force bool,
-) (*replication.DisableVolumeReplicationResponse, error) {
+) error {
 	if !mirroringInfo.Primary {
 		// Return success if the below condition is met
 		// Local image is secondary
@@ -102,32 +99,30 @@ func DisableVolumeReplication(rbdVol *rbdVolume,
 		// disabled the image on all the remote (secondary) clusters will get
 		// auto-deleted. This helps in garbage collecting the volume
 		// replication Kubernetes artifacts after failback operation.
-		localStatus, rErr := rbdVol.GetLocalState()
+		localStatus, rErr := rv.GetLocalState()
 		if rErr != nil {
-			return nil, status.Error(codes.Internal, rErr.Error())
+			return fmt.Errorf("%w: %w", ErrFetchingLocalState, rErr)
 		}
 		if localStatus.Up && localStatus.State == librbd.MirrorImageStatusStateReplaying {
-			return &replication.DisableVolumeReplicationResponse{}, nil
+			return nil
 		}
 
-		return nil, status.Errorf(codes.InvalidArgument,
-			"secondary image status is up=%t and state=%s",
-			localStatus.Up,
-			localStatus.State)
+		return fmt.Errorf("%w: secondary image status is up=%t and state=%s",
+			ErrInvalidArgument, localStatus.Up, localStatus.State)
 	}
-	err := rbdVol.DisableImageMirroring(force)
+	err := rv.DisableImageMirroring(force)
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return fmt.Errorf("%w: %w", ErrDisableImageMirroringFailed, err)
 	}
 	// the image state can be still disabling once we disable the mirroring
 	// check the mirroring is disabled or not
-	mirroringInfo, err = rbdVol.GetImageMirroringInfo()
+	mirroringInfo, err = rv.GetImageMirroringInfo()
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return fmt.Errorf("%w: %w", ErrFetchingMirroringInfo, err)
 	}
 	if mirroringInfo.State == librbd.MirrorImageDisabling {
-		return nil, status.Errorf(codes.Aborted, "%s is in disabling state", rbdVol.VolID)
+		return fmt.Errorf("%w: %q is in disabling state", ErrAborted, rv.VolID)
 	}
 
-	return &replication.DisableVolumeReplicationResponse{}, nil
+	return nil
 }


internal/rbd/errors.go
```diff
@@ -45,4 +45,16 @@ var (
...
 	// the image.
 	ErrLastSyncTimeNotFound = errors.New("last sync time not found")
+	// ErrFailedPrecondition is returned when operation is rejected because the system is not in a state 
+	// required for the operation's execution.
+	ErrFailedPrecondition = errors.New("operation is rejected because the system is not in a state required for the operation's execution.")
+    // ErrUnavailable is returned when the service is currently unavailable.
+	// This is a most likely a transient condition and may be corrected
+	// by retrying with a backoff. Note that it is not always safe to retry
+	// non-idempotent operations.
+	ErrUnavailable = errors.New("service is currently unavailable")
```
yati1998: ErrUnavailable could be ErrUnavailableService for better readability.

internal/rbd/errors.go
```diff
@@ -45,4 +45,16 @@ var (
 	// ErrLastSyncTimeNotFound is returned when last sync time is not found for
 	// the image.
 	ErrLastSyncTimeNotFound = errors.New("last sync time not found")
+	// ErrFailedPrecondition is returned when operation is rejected because the system is not in a state 
+	// required for the operation's execution.
+	ErrFailedPrecondition = errors.New("operation is rejected because the system is not in a state required for the operation's execution.")
+    // ErrUnavailable is returned when the service is currently unavailable.
```
yati1998: please check the indentation.

Rakshith-R (changes requested):
The linked issue does not require this change.
Can you provide the reason for making this change ? 
grpc codes & status are used for a reason.

riya-singhal31:
> The linked issue does not require this change. Can you provide the reason for making this change ? grpc codes & status are used for a reason.

This is the update for - https://github.com/ceph/ceph-csi/pull/3608/files#r1092953730 
This commit involves cleanup of grpc package and will be adding another commit to remove csi-addons.

Rakshith-R:
> > The linked issue does not require this change. Can you provide the reason for making this change ? grpc codes & status are used for a reason.
> 
> This is the update for - https://github.com/ceph/ceph-csi/pull/3608/files#r1092953730 This commit involves cleanup of grpc package and will be adding another commit to remove csi-addons.

The main devel branch should get only complete prs going from one stable state to another.
Please mark the pr ready for review (draft state otherwise) until all commits are in.
Make sure the functions(in csiaddons package) which in turn handle these errors definitely need to return the correct corresponding grpc status and codes.

nixpanic (changes requested):
This is a good start for cleaning up, thanks!

You'll need to check the callers of the functions, and map the new errors to some gRPC codes. The CSI protocol expects a gRPC code+error to be returned from the upper API functions. Have a look at the CSI spec to see what code is suitable for the error in the particular API (`CreateVolume` or `DeleteVolume` may return a different error to its caller for the same internal failure).

internal/rbd/errors.go
riya-singhal31: Updated, thanks!

riya-singhal31:
> > > The linked issue does not require this change. Can you provide the reason for making this change ? grpc codes & status are used for a reason.
> > 
> > 
> > This is the update for - https://github.com/ceph/ceph-csi/pull/3608/files#r1092953730 This commit involves cleanup of grpc package and will be adding another commit to remove csi-addons.
> 
> The main devel branch should get only complete prs going from one stable state to another. Please mark the pr ready for review (draft state otherwise) until all commits are in. Make sure the functions(in csiaddons package) which in turn handle these errors definitely need to return the correct corresponding grpc status and codes.

Yes, I have used the similar error definitions as the corresponding grpc status and codes.

internal/csi-addons/rbd/replication.go
```diff
@@ -326,7 +326,13 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 	case librbd.MirrorImageDisabling:
 		return nil, status.Errorf(codes.Aborted, "%s is in disabling state", volumeID)
 	case librbd.MirrorImageEnabled:
-		return corerbd.DisableVolumeReplication(rbdVol, mirroringInfo, force)
+		err = rbdVol.DisableVolumeReplication(mirroringInfo, force)
+		if err != nil {
+			return nil, err
```
nixpanic: this should not return `err` directly, but a gRPC error with a suitable code. `DisableVolumeReplication()` did not seem to have done that in all cases, that was not correct. With this refactoring it becomes clear, and you should address that.

internal/rbd/replication.go
```diff
@@ -18,32 +18,31 @@ package rbd
...
 )
 
 func (rv *rbdVolume) ResyncVol(localStatus librbd.SiteMirrorImageStatus, force bool) error {
 	if resyncRequired(localStatus) {
 		// If the force option is not set return the error message to retry
 		// with Force option.
 		if !force {
-			return status.Errorf(codes.FailedPrecondition,
-				"image is in %q state, description (%s). Force resync to recover volume",
+			fmt.Errorf("image is in %q state, description (%s). Force resync to recover volume",
```
nixpanic: this does not look righ, you may want to assign it to `err` and return that?

```
internal/rbd/replication.go:32:14: unusedresult: result of fmt.Errorf call not used (govet)
			fmt.Errorf("image is in %q state, description (%s). Force resync to recover volume",
			          ^
```

riya-singhal31:
This now removes the replication service from internal/rbd and returns gRPC code+error in internal/csi-addons/rbd.
Thanks, PTAL.
@nixpanic @Rakshith-R

internal/rbd/errors.go
```diff
@@ -45,4 +45,16 @@ var (
...
+	ErrFailedPrecondition = errors.New("system is not in a state required for the operation's execution")
+	// ErrServiceUnavailable is returned when the service is currently unavailable.
+	// This is a most likely a transient condition and may be corrected
+	// by retrying with a backoff. Note that it is not always safe to retry
+	// non-idempotent operations.
+	ErrServiceUnavailable = errors.New("service is currently unavailable")
+	// ErrAborted is returned when the operation is aborted.
+	ErrAborted = errors.New("operation got aborted")
+	// ErrInvalidArgument is returned when the client specified an invalid argument.
+	ErrInvalidArgument = errors.New("invalid arguments provided")
```
Madhu-1: where are these error messages are used?

internal/csi-addons/rbd/replication.go
```diff
@@ -326,7 +326,23 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
...
+			return nil, status.Error(codes.Internal, err.Error())
+		case strings.Contains(err.Error(), "secondary image status is"):
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		case strings.Contains(err.Error(), "failed to disable image mirroring"):
+			return nil, status.Error(codes.Internal, err.Error())
+		case strings.Contains(err.Error(), "failed to get mirroring info of image"):
+			return nil, status.Error(codes.Internal, err.Error())
+		case strings.Contains(err.Error(), "in disabling state"):
+			return nil, status.Error(codes.Aborted, err.Error())
+		}
```
Madhu-1: You should define an error and return it and use `errors.IS/As` for comparison instead of string comparison.

Madhu-1 (changes requested):

internal/rbd/replication.go
```diff
@@ -19,31 +19,31 @@ package rbd
...
 
 func (rv *rbdVolume) ResyncVol(localStatus librbd.SiteMirrorImageStatus, force bool) error {
 	if resyncRequired(localStatus) {
 		// If the force option is not set return the error message to retry
 		// with Force option.
 		if !force {
-			return status.Errorf(codes.FailedPrecondition,
-				"image is in %q state, description (%s). Force resync to recover volume",
-				localStatus.State, localStatus.Description)
+		 	fmt.Errorf("image is in %q state, description (%s). Force resync to recover volume", localStatus.State, localStatus.Description)
```
nixpanic: These `fmt.Errorf(....)` statements return an error, and a lot of them are not used/returned.

You can probably do something like this which will be better:

```go
return fmt.Errorf("%w: image is in %q state, description (%s). Force resync to recover volume", ErrFailedPrecondition, localStatus.State, localStatus.Description)
```
or
```go
err = fmt.Errorf("%w: image is in %q state, description (%s). Force resync to recover volume", ErrFailedPrecondition, localStatus.State, localStatus.Description)
return err
```
By using `%w` and pasing `ErrFailedPrecondition` for the formatting, you get the error message in the error string, and the type of the error will still be `ErrFailedPrecondition`.

internal/rbd/replication.go
```diff
@@ -19,31 +19,31 @@ package rbd
...
+			return ErrFailedPrecondition
 		}
 		err := rv.resyncImage()
 		if err != nil {
-			return status.Error(codes.Internal, err.Error())
+			fmt.Errorf("failed to resync image")
+			return err
```
nixpanic: What type of error is returned here? Maybe it makes sense to provide a comment at the top of the function that describes the different errors? The caller of the function can then easily figure out what errors need to be caught and handled differently.

internal/csi-addons/rbd/replication.go
```diff
@@ -326,7 +326,23 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 	case librbd.MirrorImageDisabling:
 		return nil, status.Errorf(codes.Aborted, "%s is in disabling state", volumeID)
 	case librbd.MirrorImageEnabled:
-		return corerbd.DisableVolumeReplication(rbdVol, mirroringInfo, force)
+		err = rbdVol.DisableVolumeReplication(mirroringInfo, force)
+		if err == nil {
+			return &replication.DisableVolumeReplicationResponse{}, nil
+		}
+		switch {
+		case strings.Contains(err.Error(), "failed to get local state"):
```
nixpanic: Instead of checking for contents of the string, use `errors.Is()` as function to check if an error is of a particular type.

https://github.com/ceph/ceph-csi/blob/40888f01b66245b3b15042b62903ae742cae494a/internal/rbd/clone.go#L58-L88

internal/rbd/replication.go
riya-singhal31: Yes Niels, I have already done this here (in 2nd commit), 			
return fmt.Errorf("image is in %q state, description (%s). Force resync to recover volume: %w",

riya-singhal31:
Thanks for the comments, have addressed all the suggested modification, requesting you to PTAL,
@Madhu-1 @nixpanic

internal/csi-addons/rbd/replication.go
```diff
@@ -326,7 +326,24 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
...
+				return nil, status.Error(codes.Internal, err.Error())
+			case errors.Is(err, corerbd.ErrInvalidArgument):
+				return nil, status.Error(codes.InvalidArgument, err.Error())
+			case errors.Is(err, corerbd.ErrDisableImageMirroringFailed):
+				return nil, status.Error(codes.Internal, err.Error())
+			case errors.Is(err, corerbd.ErrMirroringInfoNotFound):
+				return nil, status.Error(codes.Internal, err.Error())
+			case errors.Is(err, corerbd.ErrAborted):
+				return nil, status.Error(codes.Aborted, err.Error())
+			}
```
nixpanic: please add a `default` case, otherwise uncaught errors will get ignored

internal/csi-addons/rbd/replication.go
```diff
@@ -627,9 +644,14 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
-
-		return nil, err
+		switch {
+		case errors.Is(err, corerbd.ErrFailedPrecondition):
+			return nil, status.Error(codes.FailedPrecondition, err.Error())
+		case errors.Is(err, corerbd.ErrResyncImageFailed):
+			return nil, status.Error(codes.Internal, err.Error())
+		case errors.Is(err, corerbd.ErrServiceUnavailable):
+			return nil, status.Error(codes.Unavailable, err.Error())
+		}
```
nixpanic: same here, needs a `default:` to catch unknown errors

nixpanic (changes requested):
You can squash both commits into one, as the 2nd almost redoes everything from the 1st. 

Applying only the 1st will likely not result in a functional ceph-csi executable. Only if the commits are in such an order and independent enough so that applying one at the time, run tests, apply the next, run tests, keeps working as it should, commits can be split. (That is the best way, but sometimes difficult to get right. It is required for `git blame` which can be used for bug hunting.)

riya-singhal31:
> You can squash both commits into one, as the 2nd almost redoes everything from the 1st.
> 
> Applying only the 1st will likely not result in a functional ceph-csi executable. Only if the commits are in such an order and independent enough so that applying one at the time, run tests, apply the next, run tests, keeps working as it should, commits can be split. (That is the best way, but sometimes difficult to get right. It is required for `git blame` which can be used for bug hunting.)

Updated, thanks.

riya-singhal31:
Update: To reduce the cyclomatic complexity, changed the switch statements to storing the values in map.

internal/csi-addons/rbd/replication.go
```diff
@@ -649,6 +665,17 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	return resp, nil
 }
 
+func getError(err error, errorStatusMap map[error]codes.Code) error {
```
nixpanic: New helper functions like this require a unit-test to prevent them from breaking in the future. Please add one.

internal/csi-addons/rbd/replication.go
```diff
@@ -627,9 +639,13 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 
 	err = rbdVol.ResyncVol(localStatus, req.Force)
 	if err != nil {
-		log.ErrorLog(ctx, err.Error())
+		errorStatusMap := map[error]codes.Code{
```
nixpanic: There could be a signle `errorStatusMap` internally to the `getError()` function. You use it twice, and there is no overlap in the keys.

nixpanic (changes requested):

internal/csi-addons/rbd/replication.go
riya-singhal31: Added the unit test, thank you!

internal/csi-addons/rbd/replication.go
```diff
@@ -326,7 +326,12 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 	case librbd.MirrorImageDisabling:
 		return nil, status.Errorf(codes.Aborted, "%s is in disabling state", volumeID)
 	case librbd.MirrorImageEnabled:
-		return corerbd.DisableVolumeReplication(rbdVol, mirroringInfo, force)
+		err = rbdVol.DisableVolumeReplication(mirroringInfo, force)
+		if err == nil {
```
nixpanic: nit: the usual check is for `err != nil`, and have the success path continue after the if-statement

nixpanic (dismissed):

internal/rbd/errors.go
```diff
@@ -45,4 +45,24 @@ var (
...
 	// the image.
 	ErrLastSyncTimeNotFound = errors.New("last sync time not found")
+	// ErrFailedPrecondition is returned when operation is rejected because the system is not in a state
+	// required for the operation's execution.
+	ErrFailedPrecondition = errors.New("system is not in a state required for the operation's execution")
+	// ErrServiceUnavailable is returned when the service is currently unavailable.
+	// This is a most likely a transient condition and may be corrected
+	// by retrying with a backoff. Note that it is not always safe to retry
+	// non-idempotent operations.
+	ErrServiceUnavailable = errors.New("service is currently unavailable")
```
Rakshith-R: Please adapt this generic definition to the context.

Just `ErrUnavailable` is fine. 
In this context
```
// If we issued a resync, return a non-final error as image needs to be recreated
		// locally. Caller retries till RBD syncs an initial version of the image to
		// report its status in the resync request.
		return status.Error(codes.Unavailable, "awaiting initial resync due to split brain")
```

Something like Send this err to trigger a retry from the caller side.

internal/rbd/replication.go
```diff
@@ -102,32 +99,30 @@ func DisableVolumeReplication(rbdVol *rbdVolume,
...
 	// check the mirroring is disabled or not
-	mirroringInfo, err = rbdVol.GetImageMirroringInfo()
+	mirroringInfo, err = rv.GetImageMirroringInfo()
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return fmt.Errorf("%w: %w", ErrMirroringInfoNotFound, err)
 	}
 	if mirroringInfo.State == librbd.MirrorImageDisabling {
-		return nil, status.Errorf(codes.Aborted, "%s is in disabling state", rbdVol.VolID)
+		return fmt.Errorf("%w: %s is in disabling state", ErrAborted, rv.VolID)
```
Rakshith-R:
```suggestion
		return fmt.Errorf("%w: %q is in disabling state", ErrAborted, rv.VolID)
```

internal/rbd/replication.go
```diff
@@ -102,32 +99,30 @@ func DisableVolumeReplication(rbdVol *rbdVolume,
 		// disabled the image on all the remote (secondary) clusters will get
 		// auto-deleted. This helps in garbage collecting the volume
 		// replication Kubernetes artifacts after failback operation.
-		localStatus, rErr := rbdVol.GetLocalState()
+		localStatus, rErr := rv.GetLocalState()
 		if rErr != nil {
-			return nil, status.Error(codes.Internal, rErr.Error())
+			return fmt.Errorf("%w: %w", ErrLocalStateNotFound, rErr)
```
Rakshith-R:
```suggestion
			return fmt.Errorf("%w: failed to ... : %w", ErrLocalStateNotFound, rErr)
```
To maintain consistency and better tracing.

internal/rbd/replication.go
```diff
@@ -102,32 +99,30 @@ func DisableVolumeReplication(rbdVol *rbdVolume,
...
-			localStatus.Up,
-			localStatus.State)
+		return fmt.Errorf("%w: secondary image status is up=%t and state=%s",
+			ErrInvalidArgument, localStatus.Up, localStatus.State)
 	}
-	err := rbdVol.DisableImageMirroring(force)
+	err := rv.DisableImageMirroring(force)
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return fmt.Errorf("%w: %w", ErrDisableImageMirroringFailed, err)
```
Rakshith-R, line 115:
```suggestion
		return fmt.Errorf("%w: failed to ... : %w", ErrDisableImageMirroringFailed, err)
```

internal/rbd/replication.go
```diff
@@ -102,32 +99,30 @@ func DisableVolumeReplication(rbdVol *rbdVolume,
...
 	}
 	// the image state can be still disabling once we disable the mirroring
 	// check the mirroring is disabled or not
-	mirroringInfo, err = rbdVol.GetImageMirroringInfo()
+	mirroringInfo, err = rv.GetImageMirroringInfo()
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return fmt.Errorf("%w: %w", ErrMirroringInfoNotFound, err)
```
Rakshith-R:
```suggestion
		return fmt.Errorf("%w: failed to ...  : %w", ErrMirroringInfoNotFound, err)
```

internal/csi-addons/rbd/replication_test.go
```diff
@@ -494,3 +498,42 @@ func TestValidateLastSyncTime(t *testing.T) {
...
 }
+
+func TestGetError(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name        string
+		err         error
+		expectedErr error
+	}{
+		{
```
Rakshith-R, line 509: Unit test should cover all possible cases.

internal/csi-addons/rbd/replication.go
```diff
@@ -649,6 +652,28 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+func getError(err error) error {
+	errorStatusMap := map[error]codes.Code{
+		corerbd.ErrLocalStateNotFound:          codes.Internal,
+		corerbd.ErrInvalidArgument:             codes.InvalidArgument,
+		corerbd.ErrDisableImageMirroringFailed: codes.Internal,
+		corerbd.ErrMirroringInfoNotFound:       codes.Internal,
+		corerbd.ErrAborted:                     codes.Aborted,
+		corerbd.ErrFailedPrecondition:          codes.FailedPrecondition,
+		corerbd.ErrResyncImageFailed:           codes.Internal,
+		corerbd.ErrServiceUnavailable:          codes.Unavailable,
```
Rakshith-R:
```suggestion
		corerbd.ErrLocalStateNotFound:          codes.Internal,
		corerbd.ErrResyncImageFailed:           codes.Internal,
		corerbd.ErrDisableImageMirroringFailed: codes.Internal,
		corerbd.ErrMirroringInfoNotFound:       codes.Internal,
		corerbd.ErrInvalidArgument:             codes.InvalidArgument,
		corerbd.ErrAborted:                     codes.Aborted,
		corerbd.ErrFailedPrecondition:          codes.FailedPrecondition,
		corerbd.ErrServiceUnavailable:          codes.Unavailable,
```
nit: better to have it ordered.

internal/rbd/replication.go
riya-singhal31: Thanks for the comment Rakshith, just to clarify : We already have this definition in errors.go -> failed to get local state
Did you meant to put that(failed to.. statement) twice here, as it will be retrieved from ErrLocalStateNotFound too. 
@Rakshith-R

Rakshith-R: In that case of repeating" failed to ..."
It's fine.

and @riya-singhal31 failed to get is not equivalent to not found.

Please change the error name to something like `ErrFetching..` instead of `Err..NotFound`.
NotFound in the error definition is misleading IMO.

riya-singhal31: Updated, thanks. PTAL

internal/csi-addons/rbd/replication.go
riya-singhal31: Thanks Niels, updated. PTAL

nixpanic (dismissed):

internal/rbd/errors.go
```diff
@@ -45,4 +45,22 @@ var (
...
+	// ErrInvalidArgument is returned when the client specified an invalid argument.
+	ErrInvalidArgument = errors.New("invalid arguments provided")
+	// ErrFetchingLocalState is returned when the local state is not found.
+	ErrFetchingLocalState = errors.New("failed to get local state")
+	// ErrDisableImageMirroringFailed is returned when the operation to disable image mirroring is failed.
+	ErrDisableImageMirroringFailed = errors.New("failed to disable image mirroring")
+	// ErrFetchingMirroringInfo is returned when mirroring info of image is not found.
+	ErrFetchingMirroringInfo = errors.New("failed to get mirroring info of image")
+	// ErrResyncImageFailed is returned when the image is failed to resync.
+	ErrResyncImageFailed = errors.New("failed to resync image")
```
Rakshith-R:
```suggestion
	// ErrFetchingLocalState is returned when the operation to fetch local state fails.
	ErrFetchingLocalState = errors.New("failed to get local state")
	// ErrDisableImageMirroringFailed is returned when the operation to disable image mirroring fails.
	ErrDisableImageMirroringFailed = errors.New("failed to disable image mirroring")
	// ErrFetchingMirroringInfo is returned when the operation to fetch mirroring info of image fails.
	ErrFetchingMirroringInfo = errors.New("failed to get mirroring info of image")
	// ErrResyncImageFailed is returned when the operation to resync the image fails.
	ErrResyncImageFailed = errors.New("failed to resync image")
```
nit

internal/rbd/replication.go
Rakshith-R: This is still pending.

Rakshith-R (commented):
it almost looks good to me.
just a few nits.

riya-singhal31:
> it almost looks good to me. just a few nits.

Thanks, Updated.
@Rakshith-R PTAL.

Rakshith-R (dismissed):

Rakshith-R:
@Madhu-1 ,
Can you please take a look, since you've requested changes ?

internal/csi-addons/rbd/replication.go
```diff
@@ -649,6 +652,28 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	return resp, nil
 }
 
+func getError(err error) error {
```
```suggestion
func getGRPCError(err error) error {
```

Madhu-1 (commented):
small nit, LGTM

internal/csi-addons/rbd/replication.go
```diff
@@ -649,6 +652,28 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+	}
+
+	for e, code := range errorStatusMap {
+		if errors.Is(err, e) {
+			return status.Error(code, err.Error())
+		}
+	}
+
+	// Handle any other error not listed in the map
+	return status.Error(codes.Unknown, err.Error())
```
Madhu-1, line 678: This is panic if the passed err is nil?

riya-singhal31: No, this is a kind of defualt statement in switch. If the passed err is not among the ones listed in map and is not nil.

Madhu-1: can you add a unit test for a nil error?

riya-singhal31: We are calling this function only if the passed error is not nil, that's why we don't have any handling for nil here.

riya-singhal31: Adding a unit test for nil, it will send the unknown error otherwise, which is not expected.

Madhu-1: Please add it as a documentation to the function and also mention this function will panic if you pass the non-nil error

nixpanic: If `err` is `nil`, the this should probably return a gRPC `code.OK` with `code.OK.String()` as message.

riya-singhal31: Updated the function with nil check and added the same case in unit test so now it doesn't need a documentation for it.

nixpanic (dismissed):

Rakshith-R (commented):
lgtm

internal/csi-addons/rbd/replication.go
```diff
@@ -649,6 +652,32 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+
+	for e, code := range errorStatusMap {
+		if errors.Is(err, e) {
+			return status.Error(code, err.Error())
+		}
+	}
+
+	if err == nil {
+		return status.Error(codes.OK, codes.OK.String())
+	}
```
Madhu-1: This should be the first check, move this to line 656

riya-singhal31: Done, thanks!

internal/csi-addons/rbd/replication.go
```diff
@@ -649,6 +652,32 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+		corerbd.ErrUnavailable:                 codes.Unavailable,
+	}
+
+	for e, code := range errorStatusMap {
+		if errors.Is(err, e) {
+			return status.Error(code, err.Error())
+		}
+	}
+
+	// Handle any other error nol nil error not listed in the map
```
```suggestion
// Handle any other error non nil error not listed in the map
```

nixpanic (approved):

Madhu-1 (approved):

Rakshith-R:
@mergifyio queue

PR #3894, opened 2023-06-09
rbd: get lastsyncbytes and lastsycduration for volrep

### LINKED ISSUE #3889: update getVolumeReplicationInfo to reflect more metrics/info
rbd mirror image status contains the lastsyncbytes and lastsyncduration. ceph-csi extracts those data and sends the response back through the getVolumeReplicationInfo.

Here's the output example for the same:

```

rbd mirror image status  replicapool/csi-vol-9384c8a6-ea23-46d4-9f67-8eff1e444035 csi-vol-9384c8a6-ea23-46d4-9f67-8eff1e444035:   global_id:   e55504e0-324d-46b0-a9ba-1ea67fa39e59   state:       up+replaying   description: replaying, {"bytes_per_second":0.0,"bytes_per_snapshot":81920.0,"last_snapshot_bytes":81920,"last_snapshot_sync_seconds":0,"local_snapshot_timestamp":1684675261,"remote_snapshot_timestamp":1684675261,"replay_state":"idle"}   service:     a on dr2   last_update: 2023-05-22 11:38:03   peer_sites:     name: c6b80b8c-4308-42bc-851f-23002cfad6ee     state: up+starting_replay     description: starting replay     last_update: 2023-05-22 11:38:05

```

yati1998:
This commit get more information from the description like lastsyncbytes and lastsyncduration and send them as a response of getvolumereplicationinfo request.


internal/csi-addons/rbd/replication.go
```diff
@@ -794,6 +805,68 @@ func getLastSyncTime(description string) (*timestamppb.Timestamp, error) {
...
+	if description == "" {
+		return 0, nil
+	}
+	splittedString := strings.SplitN(description, ",", 2)
+	if len(splittedString) == 1 {
+		return 0, nil
+	}
+	type localStatus struct {
+		LocalSnapshotBytes int64 `json:"last_snapshot_bytes"`
+	}
```
nixpanic: This code is repeated for `getLastSyncDuration()`, please put it in a helper function that returns a `localStatus`.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ceph/ceph-csi
 
-go 1.20
+go 1.19
```
nixpanic: watch out, this should not have been included.

yati1998:
cc @ShyamsundarR

internal/csi-addons/rbd/replication.go
```diff
@@ -756,42 +759,68 @@ func RemoteStatus(gmis *librbd.GlobalMirrorImageStatus) (librbd.SiteMirrorImageS
...
-		return nil, fmt.Errorf("empty local snapshot timestamp: %w", corerbd.ErrLastSyncTimeNotFound)
+		return nil, nil, 0, fmt.Errorf("empty local snapshot timestamp: %w", corerbd.ErrLastSyncTimeNotFound)
 	}
 
 	lastUpdateTime := time.Unix(localSnapTime.LocalSnapshotTime, 0)
 	lastSyncTime := timestamppb.New(lastUpdateTime)
 
-	return lastSyncTime, nil
+	var localSnapBytes localStatus
+	errb := json.Unmarshal([]byte(splittedString[1]), &localSnapBytes)
```
nixpanic: no need to `json.Unmarshal()` again, you already have `localStatus` which you can use

internal/csi-addons/rbd/replication.go
```diff
@@ -756,42 +759,68 @@ func RemoteStatus(gmis *librbd.GlobalMirrorImageStatus) (librbd.SiteMirrorImageS
...
-	// "bytes_per_snapshot":149504.0,"local_snapshot_timestamp":1662655501
-	// ,"remote_snapshot_timestamp":1662655501}"
+	// description = `replaying, {"bytes_per_second":0.0,"bytes_per_snapshot":81920.0,
+	// "last_snapshot_bytes":81920,"last_snapshot_sync_seconds":0,
+	// "local_snapshot_timestamp":1684675261,
+	// "remote_snapshot_timestamp":1684675261,"replay_state":"idle"}`
+	// In case there is no last snapshot bytes returns 0 as the
+	// LastSyncBytes is optional.
+	// In case there is no last snapshot sync seconds, it returns nil as the
+	// LastSyncDuration is optional.
```
Rakshith-R, line 803: Was `lastSync[Byte|duration]` added recently to ceph ? 
if so, please add ceph tracker/ version from which this was added for reference in comment.

have you tested this call on a ceph cluster version that does not provide these information ?

internal/csi-addons/rbd/replication.go
```diff
@@ -756,42 +759,68 @@ func RemoteStatus(gmis *librbd.GlobalMirrorImageStatus) (librbd.SiteMirrorImageS
...
+
+	var localSnapDuration localStatus
+	errd := json.Unmarshal([]byte(splittedString[1]), &localSnapDuration)
+	if errd != nil {
+		return nil, nil, 0, fmt.Errorf("failed to unmarshal last snap sync seconds: %w", errd)
+	}
+
+	lastUpdateDuration := time.Unix(localSnapDuration.LocalSnapshotDuration, 0)
+	lastUpdateDuration1 := time.Until(lastUpdateDuration)
+	lastSyncDuration := durationpb.New(lastUpdateDuration1)
```
Rakshith-R: Please use better variable naming and explain this logic in a comment above.

internal/csi-addons/rbd/replication.go
```diff
@@ -756,42 +759,68 @@ func RemoteStatus(gmis *librbd.GlobalMirrorImageStatus) (librbd.SiteMirrorImageS
...
+
+	return lastSyncTime, lastSyncDuration, localSnapBytes.LocalSnapshotBytes, nil
+
```
Rakshith-R: nit: extra newline.

Rakshith-R (changes requested):
Have you tested this end to end in a live cluster ? 

Can you please provide the new CR yaml with status and logs of each container? 

Please test it also with ceph cluster version that does not have this change(if these variables were added recently)?

internal/csi-addons/rbd/replication.go
```diff
@@ -756,42 +759,59 @@ func RemoteStatus(gmis *librbd.GlobalMirrorImageStatus) (librbd.SiteMirrorImageS
...
 	}
 	splittedString := strings.SplitN(description, ",", 2)
 	if len(splittedString) == 1 {
-		return nil, fmt.Errorf("no local snapshot timestamp: %w", corerbd.ErrLastSyncTimeNotFound)
+		return nil, nil, 0, fmt.Errorf("no snapshot details: %w", corerbd.ErrLastSyncTimeNotFound)
 	}
 	type localStatus struct {
-		LocalSnapshotTime int64 `json:"local_snapshot_timestamp"`
+		LocalSnapshotTime     int64 `json:"local_snapshot_timestamp"`
+		LocalSnapshotBytes    int64 `json:"last_snapshot_bytes"`
```
nixpanic: the naming of these members are not matching the json contents. Note the difference in `Local` and `last_`. Please rename them to match the json, it is a little confusing otherwise.

yati1998:
> Have you tested this end to end in a live cluster ?
>  testing is stuck still de to some image issue, but it shouldn't be any issue, as this follows the same flow as lastsynctime. But Still working on to get working examples and logs.
Meanwhile, feel free to review the PR if any other changes are required.

internal/csi-addons/rbd/replication.go
Rakshith-R: The above query was not answered.

The localStatus struct which is used for unmarshalling will give default 0 value when the field is not present ? 
Check this with unit test cases.

We will need to handle missing values.
The current code workflow does not do that.

Rakshith-R (changes requested):

internal/csi-addons/rbd/replication.go
yati1998: Yeah, the 0 value of lastSyncBytes states undefined files, which has been mentioned in csi-addons. and if the value is 0 the lastsyncbytes in volumereplication operator will not be updated.

Rakshith-R: Similar logic needs to be applied for lastSyncDuration below too?

internal/csi-addons/rbd/replication.go
```diff
@@ -756,42 +759,59 @@ func RemoteStatus(gmis *librbd.GlobalMirrorImageStatus) (librbd.SiteMirrorImageS
 	return ss, err
 }
 
-// This function gets the local snapshot time from the description
-// of localStatus and converts it into required type.
-func getLastSyncTime(description string) (*timestamppb.Timestamp, error) {
+// This function gets the local snapshot time, last sync snapshot seconds
+// and last sync bytes from the description of localStatus and convert
+// it into required types.
+func getLastSyncInfo(description string) (*timestamppb.Timestamp, *durationpb.Duration, int64, error) {
```
Madhu-1: create a local struct and return it in response, this avoids returning many values

internal/csi-addons/rbd/replication.go
yati1998: yes, thats done.

internal/csi-addons/rbd/replication_test.go
```diff
@@ -438,58 +439,114 @@ func TestCheckRemoteSiteStatus(t *testing.T) {
...
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			ts, err := getLastSyncTime(tt.description)
+			teststruct, err := getLastSyncInfo(tt.description)
 			if err != nil && !strings.Contains(err.Error(), tt.expectedErr) {
 				// returned error
-				t.Errorf("getLastSyncTime() returned error, expected: %v, got: %v",
+				t.Errorf("getLastSyncInfo() returned error, expected: %v, got: %v",
```
nixpanic, line 544: Including `tt.name` in the logs helps when there is a failure.

https://github.com/ceph/ceph-csi/actions/runs/5338117669/jobs/9675045736?pr=3894#step:3:1194
https://github.com/ceph/ceph-csi/actions/runs/5338117669/jobs/9675045736?pr=3894#step:3:1262

the order of the tests may not be the same as  the log message, I think?
https://github.com/ceph/ceph-csi/actions/runs/5338117669/jobs/9675045736?pr=3894#step:3:1271

nixpanic (dismissed):

internal/csi-addons/rbd/replication.go
```diff
@@ -756,42 +765,68 @@ func RemoteStatus(gmis *librbd.GlobalMirrorImageStatus) (librbd.SiteMirrorImageS
...
 
-	lastUpdateTime := time.Unix(localSnapTime.LocalSnapshotTime, 0)
+	// converts localSnapshotTime of type int64 to time.Time
+	lastUpdateTime := time.Unix(localSnapInfo.LocalSnapshotTime, 0)
 	lastSyncTime := timestamppb.New(lastUpdateTime)
 
-	return lastSyncTime, nil
+	// converts localSnapshotDuration of type int64 to time.Time
+	lastDurationTime := strconv.Itoa(int(localSnapInfo.LastSnapshotDuration)) + "s"
+	// converts time.time to time.Duration
```
```suggestion
// converts localSnapshotDuration of type int64 to string format with appended `s` seconds required  for time.ParseDuration
lastDurationTime := fmt.Sprintf("%ds",localSnapInfo.LastSnapshotDuration)
// parse Duration from the lastDurationTime string
```

Madhu-1 (commented):
small nit

internal/csi-addons/rbd/replication.go
Rakshith-R: My very first question in this thread is still unanswered. 

Still does not handle the case where duration does not exist in the description.

Current workflow will return 0s which is misleading IMO.

internal/csi-addons/rbd/replication.go
yati1998: done

nixpanic (dismissed):

internal/csi-addons/rbd/replication.go
yati1998: That's what is mentioned in the comments, o/nil indicates the value of this field is undefined. We can't return any error as it's optional, and in case duration is not there it will be set to nil, 0 in case of lastsyncbytes. These values are sent by ceph-csi and in csi-addons we are keeping it nil only, which will be updates to the status.

internal/csi-addons/rbd/replication_test.go
```diff
@@ -438,58 +440,118 @@ func TestCheckRemoteSiteStatus(t *testing.T) {
...
-			"description with no JSON",
-			`replaying`,
-			nil,
-			corerbd.ErrLastSyncTimeNotFound.Error(),
+			name: "description without last_snapshot_sync_seconds",
+			//nolint:lll // sample output cannot be split into multiple lines.
+			description: `replaying, {"bytes_per_second":0.0,"bytes_per_snapshot":81920.0,"last_snapshot_bytes":81920,"local_snapshot_timestamp":1684675261,"remote_snapshot_timestamp":1684675261,"replay_state":"idle"}`,
+			info: localstruct{
+				lastsyncduration: nil,
+				lastsynctime:     timestamppb.New(time.Unix(1684675261, 0)),
```
Rakshith-R: This should be nil as your comment in 
replicaiton.go line 779.

```
	// In case there is no last snapshot sync seconds, it returns nil as the
	// LastSyncDuration is optional.
```

internal/csi-addons/rbd/replication.go
Rakshith-R: check.
https://github.com/ceph/ceph-csi/pull/3894/files#r1238240556

internal/csi-addons/rbd/replication_test.go
yati1998: `lastsyncduration: nil,` yes its nil only right?

internal/csi-addons/rbd/replication_test.go
```diff
@@ -438,58 +440,118 @@ func TestCheckRemoteSiteStatus(t *testing.T) {
...
 			}
-			if !ts.AsTime().Equal(tt.timestamp.AsTime()) {
-				t.Errorf("getLastSyncTime() %v, expected %v", ts, tt.timestamp)
+			if teststruct.lastsynctime.GetSeconds() != tt.info.lastsynctime.GetSeconds() {
+				t.Errorf("name: %v, getLastSyncInfo() %v, expected %v", tt.name, teststruct.lastsynctime, tt.info.lastsynctime)
+			}
+			if teststruct.lastsyncduration.GetSeconds() != tt.info.lastsyncduration.GetSeconds() {
+				t.Errorf("name: %v, getLastSyncInfo() %v, expected %v", tt.name, teststruct.lastsyncduration,
+					tt.info.lastsyncduration)
+			}
```
Rakshith-R:
```suggestion
			if teststruct.lastsyncduration == nil &&  tt.info.lastsyncduration !=nil {
				t.Errorf("name: %v, getLastSyncInfo() %v, expected %v", tt.name, teststruct.lastsyncduration,
					tt.info.lastsyncduration)
			}
			if teststruct.lastsyncduration.GetSeconds() != tt.info.lastsyncduration.GetSeconds() {
				t.Errorf("name: %v, getLastSyncInfo() %v, expected %v", tt.name, teststruct.lastsyncduration,
					tt.info.lastsyncduration)
			}
```

Add another check here, 
a nil duration GetSeconds is giving 0 that is why the `missing last_sync_duration ` case is not getting caught here.

Rakshith-R (changes requested):

internal/csi-addons/rbd/replication_test.go
yati1998: Couldn't get the purpose of the extra check, can you please elaborate?
If the last_sync_duration is missing, the function returns nil and so does the testcase has and GetSeconds of both will be 0 and if it is not, the error will be caught.

Rakshith-R: Have you checked it with a println function ? 

teststruct.lastsyncduration.GetSeconds() gives a 0 when teststruct.lastsyncduration is nil.

yati1998: yes, I know that, but what is the issue with it?

nixpanic (dismissed):

yati1998:
@Rakshith-R and @nixpanic please review

internal/csi-addons/rbd/replication.go
```diff
@@ -82,6 +83,12 @@ type ReplicationServer struct {
 	*corerbd.ControllerServer
 }
 
+type localstruct struct {
```
nixpanic: Naming a struct `localstruct` is not very practical, can you give it a better name? On the other hand, it seems to contain the same fields as `replication.GetVolumeReplicationInfoResponse`, you could just use that instead as well.

internal/csi-addons/rbd/replication.go
```diff
@@ -785,42 +782,71 @@ func RemoteStatus(gmis *librbd.GlobalMirrorImageStatus) (librbd.SiteMirrorImageS
...
+	// In case there is no last snapshot sync seconds, it returns nil as the
+	// LastSyncDuration is optional.
 	// In case there is no local snapshot timestamp return an error as the
 	// LastSyncTime is required.
+
+	var response replication.GetVolumeReplicationInfoResponse
+
 	if description == "" {
-		return nil, fmt.Errorf("empty description: %w", corerbd.ErrLastSyncTimeNotFound)
+		return &response, fmt.Errorf("empty description: %w", corerbd.ErrLastSyncTimeNotFound)
```
Rakshith-R:
```suggestion
		return nil, fmt.Errorf("empty description: %w", corerbd.ErrLastSyncTimeNotFound)
```

Return nil where ever you are encountering error

Rakshith-R (changes requested):

Rakshith-R (approved):

Rakshith-R:
@mergifyio rebase

nixpanic (approved):

Issue #2295, opened 2021-07-16
SELinux-related mount options are silently ignored


alexcern:
# Describe the bug #

SELinux-related `kernelMountOptions` for a cephfs mount (typically `-o context=...`) are silently ignored.

# Environment details #

- Image/version of Ceph CSI driver : 3.2.1
- Helm chart version : 3.2.1
- Kernel version : 5.10.19-200.fc33.x86_64 (FCOS)
- Mounter used for mounting PVC (for cephfs its `fuse` or `kernel`. for rbd its
  `krbd` or `rbd-nbd`) : cephfs / kernel
- Kubernetes cluster version : 1.20
- Ceph cluster version : ?

# Steps to reproduce #

1. Consider a cephfs volume with existing SELinux labels on files
1. Create cephfs-csi PV with `kernelMountOptions: context=system_u:object_r:container_file_t:s0` (Using kernel mounter; fuse mounter doesn't support SElinux labels on files)
1. Run a pod mounting that PV as `/cephfs`

# Actual results #

`mount` shows something like this:

```
<redacted>:6789,<redacted>:6789,<redacted>:6789:/volumes/_nogroup/8693b026-39af-4c90-af60-1e8290cc3c79 on /cephfs type ceph (rw,relatime,seclabel,name=pvc-db621df2-199d-4c23-b16f-bcc0f06d8b14,secret=<hidden>,acl)
```

Notice the `context` option is missing and we have `seclabel` instead.

In the pod, `ls -lZ /cephfs` shows different SELinux file labels than what we specified in the mount options

# Expected behavior #

The SELinux file labels should be exactly the context specified in the mount options (here `system_u:object_r:container_file_t:s0`)

`mount` should show something like the following, reflecting our mount option:

```
<redacted>:6789,<redacted>:6789,<redacted>:6789:/volumes/_nogroup/8693b026-39af-4c90-af60-1e8290cc3c79 on /cephfs type ceph (rw,relatime,context=system_u:object_r:container_file_t:s0,name=pvc-db621df2-199d-4c23-b16f-bcc0f06d8b14,secret=<hidden>,acl)
```

# Logs #

# Additional context #

AFAICS this happens because `libmount` silently drops selinux-related mount options if it doesn't detect that selinux is enabled: https://github.com/karelzak/util-linux/blob/17fc869/libmount/src/context_mount.c#L291

[...]

PR #3902, opened 2023-06-13
Add support & e2e for mountOptions & efficient selinux relabelling support

### LINKED ISSUE #757: Honor mount options in NodeStage RPC as well
Currently it was decided[1] not to honor mount options in NodeStage RPC request, and we mount the volume with default options.

While discussing regarding the `_netdev` options and what ceph-csi does with mount options with @gnufied, it was brought to my attention that we should honor mount options in the NodeStage request itself.

Opening this issue to add justifications around the same, and to decide and implement honoring mount options during NodeStage request.

[1] Issue discussion:
  - https://github.com/ceph/ceph-csi/pull/576#issuecomment-528030567
  - https://github.com/ceph/ceph-csi/pull/576#issuecomment-528793122

### LINKED ISSUE #3855: Add support for Efficient SELinux volume relabeling

Efficient SELinux volume relabeling
refer:  https://kubernetes.io/blog/2023/04/18/kubernetes-1-27-efficient-selinux-relabeling-beta/

Faster mounting of pods with specific access modes in clusters with SELinux enabled OS.

PVC which suffered from delay in mounting due to selinux relabelling issue should no longer occur
for specific access modes.

Tasks:
- Add capabilities ReadWriteOncePod and SELinuxMountReadWriteOncePod.
- set `seLinuxMount: true` in csidriver object.
- complete https://github.com/ceph/ceph-csi/issues/757

Rakshith-R:
- pv.Spec.MountOptions will now be added during NodePublish as well as NodeStage Request.
- `SeLinuxMount: true` option is set in csidriver object. 
- e2e testcases is added to test presence of mountOptions during mounting by all three csidriver nodeplugins.

Note: 
- refer comment https://github.com/ceph/ceph-csi/issues/757#issuecomment-1590758956
- Automatic addition of selinux context in mountOptions cannot be tested in ci since minikube
is not selinux enabled.
- Added e2e closely mimics the selinux context that kublet would've passed. https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/csi_mock/csi_selinux_mount.go#L77

refer: https://kubernetes.io/blog/2023/04/18/kubernetes-1-27-efficient-selinux-relabeling-beta/

Resolves: #3855 #757

### DIFF
--- a/e2e/cephfs_helper.go
+++ b/e2e/cephfs_helper.go
@@ -66,6 +66,12 @@ func createCephfsStorageClass(
 	if err != nil {
 		return err
 	}
+	// TODO: remove this once the ceph-csi driver release-v3.9 is completed
+	// and upgrade tests are done from v3.9 to devel.
+	// The mountOptions from previous are not compatible with NodeStageVolume
+	// request.
+	sc.MountOptions = []string{}
+
 	sc.Parameters["fsName"] = fileSystemName
 	sc.Parameters["csi.storage.k8s.io/provisioner-secret-namespace"] = cephCSINamespace
 	sc.Parameters["csi.storage.k8s.io/provisioner-secret-name"] = cephFSProvisionerSecretName
--- a/e2e/pod.go
+++ b/e2e/pod.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/kubernetes/pkg/client/conditions"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	frameworkPod "k8s.io/kubernetes/test/e2e/framework/pod"
 )
 
 const errRWOPConflict = "node has pod using PersistentVolumeClaim with the same name and ReadWriteOncePod access mode."
@@ -164,6 +165,28 @@ func execCommandInDaemonsetPod(
 	f *framework.Framework,
 	c, daemonsetName, nodeName, containerName, ns string,
 ) (string, error) {
+	podName, err := getDaemonsetPodOnNode(f, daemonsetName, nodeName, ns)
+	if err != nil {
+		return "", err
+	}
+
+	cmd := []string{"/bin/sh", "-c", c}
+	podOpt := e2epod.ExecOptions{
+		Command:       cmd,
+		Namespace:     ns,
+		PodName:       podName,
+		ContainerName: containerName,
+		CaptureStdout: true,
+		CaptureStderr: true,
+	}
+
+	_ /* stdout */, stderr, err := execWithRetry(f, &podOpt)
+
+	return stderr, err
+}
+
+// getDaemonsetPodOnNode returns the name of a daemonset pod on a particular node.
+func getDaemonsetPodOnNode(f *framework.Framework, daemonsetName, nodeName, ns string) (string, error) {
 	selector, err := getDaemonSetLabelSelector(f, ns, daemonsetName)
 	if err != nil {
 		return "", err
@@ -187,19 +210,7 @@ func execCommandInDaemonsetPod(
 		return "", fmt.Errorf("%s daemonset pod on node %s in namespace %s not found", daemonsetName, nodeName, ns)
 	}
 
-	cmd := []string{"/bin/sh", "-c", c}
-	podOpt := e2epod.ExecOptions{
-		Command:       cmd,
-		Namespace:     ns,
-		PodName:       podName,
-		ContainerName: containerName,
-		CaptureStdout: true,
-		CaptureStderr: true,
-	}
-
-	_ /* stdout */, stderr, err := execWithRetry(f, &podOpt)
-
-	return stderr, err
+	return podName, nil
 }
 
 // listPods returns slice of pods matching given ListOptions and namespace.
@@ -535,6 +546,76 @@ func validateRWOPPodCreation(
 	}
 
 	app.Name = baseAppName
+	err = deletePVCAndApp("", f, pvc, app)
+	if err != nil {
+		return fmt.Errorf("failed to delete PVC and application: %w", err)
+	}
+
+	return nil
+}
+
+// verifySeLinuxMountOption verifies the SeLinux context MountOption added to PV.Spec.MountOption
+// is successfully used by nodeplugin during mounting by checking for its presence in the
+// nodeplugin container logs.
+func verifySeLinuxMountOption(
+	f *framework.Framework,
+	pvcPath, appPath, daemonSetName, cn, ns string,
+) error {
+	mountOption := "context=\"system_u:object_r:container_file_t:s0:c0,c1\""
+
+	// create PVC
+	pvc, err := loadPVC(pvcPath)
+	if err != nil {
+		return fmt.Errorf("failed to load pvc: %w", err)
+	}
+	pvc.Namespace = f.UniqueName
+	err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+	if err != nil {
+		return fmt.Errorf("failed to create PVC: %w", err)
+	}
+	// modify PV spec.MountOptions
+	pv, err := getBoundPV(f.ClientSet, pvc)
+	if err != nil {
+		return fmt.Errorf("failed to get PV: %w", err)
+	}
+	pv.Spec.MountOptions = []string{mountOption}
+
+	// update PV
+	_, err = f.ClientSet.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to update pv: %w", err)
+	}
+
+	app, err := loadApp(appPath)
+	if err != nil {
+		return fmt.Errorf("failed to load application: %w", err)
+	}
+	app.Namespace = f.UniqueName
+	err = createApp(f.ClientSet, app, deployTimeout)
+	if err != nil {
+		return fmt.Errorf("failed to create application: %w", err)
+	}
+
+	pod, err := f.ClientSet.CoreV1().Pods(f.UniqueName).Get(context.TODO(), app.Name, metav1.GetOptions{})
+	if err != nil {
+		framework.Logf("Error occurred getting pod %s in namespace %s", app.Name, f.UniqueName)
+
+		return fmt.Errorf("failed to get pod: %w", err)
+	}
+
+	nodepluginPodName, err := getDaemonsetPodOnNode(f, daemonSetName, pod.Spec.NodeName, ns)
+	if err != nil {
+		return fmt.Errorf("failed to get daemonset pod on node: %w", err)
+	}
+	logs, err := frameworkPod.GetPodLogs(context.TODO(), f.ClientSet, ns, nodepluginPodName, cn)
+	if err != nil {
+		return fmt.Errorf("failed to get pod logs from container %s/%s/%s : %w", ns, nodepluginPodName, cn, err)
+	}
+
+	if !strings.Contains(logs, mountOption) {
+		return fmt.Errorf("mount option %s not found in logs: %s", mountOption, logs)
+	}
+
 	err = deletePVCAndApp("", f, pvc, app)
 	if err != nil {
 		return fmt.Errorf("failed to delete PVC and application: %w", err)
--- a/deploy/cephfs/kubernetes/csidriver.yaml
+++ b/deploy/cephfs/kubernetes/csidriver.yaml
@@ -14,3 +14,4 @@ spec:
   attachRequired: false
   podInfoOnMount: false
   fsGroupPolicy: File
+  seLinuxMount: true
--- a/deploy/nfs/kubernetes/csidriver.yaml
+++ b/deploy/nfs/kubernetes/csidriver.yaml
@@ -13,5 +13,6 @@ metadata:
 spec:
   attachRequired: false
   fsGroupPolicy: File
+  seLinuxMount: true
   volumeLifecycleModes:
     - Persistent
--- a/deploy/rbd/kubernetes/csidriver.yaml
+++ b/deploy/rbd/kubernetes/csidriver.yaml
@@ -13,4 +13,5 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: false
+  seLinuxMount: true
   fsGroupPolicy: File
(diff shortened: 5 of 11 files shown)


e2e/pod.go
```diff
@@ -542,3 +553,72 @@ func validateRWOPPodCreation(
...
+	}
+	// modify PV spec.MountOptions
+	pv, err := getBoundPV(f.ClientSet, pvc)
+	if err != nil {
+		framework.Failf("failed to get PV: %v", err)
+	}
+	pv.Spec.MountOptions = []string{mountOption}
+
+	// update PV
+	_, err = f.ClientSet.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
```
Madhu-1, line 584: create one ctx and reuse it.

e2e/pod.go
```diff
@@ -542,3 +553,72 @@ func validateRWOPPodCreation(
...
+		return fmt.Errorf("failed to get daemonset pod on node: %w", err)
+	}
+	logs, err := frameworkPod.GetPodLogs(context.TODO(), f.ClientSet, ns, nodepluginPodName, cn)
+	if err != nil {
+		return fmt.Errorf("failed to get pod logs from container %s/%s/%s : %w", ns, nodepluginPodName, cn, err)
+	}
+
+	if !strings.Contains(logs, mountOption) {
+		return fmt.Errorf("mount option %s not found in logs: %s", mountOption, logs)
+	}
```
Madhu-1, line 617: verifying with logs doesn't ensure that the mount option is set, you need to inspect the mountpoint to check whether the mount option is applied on the mountpath or not.

Madhu-1 (commented):
Need to remove WIP from PR title

e2e/pod.go
Rakshith-R: > verifying with logs doesn't ensure that the mount option is set, you need to inspect the mountpoint to check whether the mount option is applied on the mountpath or not.

@Madhu-1 do you have the cmd to do that ?

Madhu-1: you should be able to get it with `mount |grep <path>`.  see if you can reuse `checkMountOptions` function which we have already.

Rakshith-R: Thanks!
 I'll re request review once ci is green with this change

e2e/utils.go
```diff
@@ -769,13 +769,32 @@ func pvcDeleteWhenPoolNotFound(pvcPath string, cephFS bool, f *framework.Framewo
...
 	}
 
 	pvc.Namespace = f.UniqueName
+	err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+	if err != nil {
+		framework.Failf("failed to create PVC: %v", err)
+	}
+	if len(mountOptions) > 0 {
+		// modify PV spec.MountOptions
+		pv, err := getBoundPV(f.ClientSet, pvc)
```
nixpanic:
```
e2e/utils.go:786:7: shadow: declaration of "err" shadows declaration at line 774 (govet)
		pv, err := getBoundPV(f.ClientSet, pvc)
```

e2e/pod.go
Rakshith-R: @Madhu-1  the selinux context is not shown in mount options but it is passed 
I guess that context is only visible on selinux enabled system. 
```
[38;5;9m[FAILED] failed to verify mount options: mount option context="system_u:object_r:container_file_t:s0:c0,c1" not found in 10.102.50.233:6789:/volumes/e2e/csi-vol-761f01c7-565d-4674-a154-809ba79de1c2/6551c48e-aff1-4e84-8b0f-a908984ac37f on /var/lib/www type ceph (rw,relatime,seclabel,name=cephcsi-cephfs-node,secret=<hidden>,fsid=00000000-0000-0000-0000-000000000000,acl,mds_namespace=myfs)
  [0m
```

https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.26/runs/369/nodes/100/log/?start=0 

ctrl+f: "context="
```
  I0614 10:22:00.902816   27952 cephcmds.go:105] ID: 21 Req-ID: 0001-0024-dadd4530-ed10-4771-8fcf-b55843578b77-0000000000000001-761f01c7-565d-4674-a154-809ba79de1c2 command succeeded: mount [-t ceph rook-ceph-mon-a.rook-ceph.svc.cluster.local:6789:/volumes/e2e/csi-vol-761f01c7-565d-4674-a154-809ba79de1c2/6551c48e-aff1-4e84-8b0f-a908984ac37f /var/lib/kubelet/plugins/kubernetes.io/csi/cephfs.csi.ceph.com/f400b7bc4e6e1ae0d881d5d12e9ff8a70cb54e6f642e6c2a79e3fb4b3ac58050/globalmount -o name=cephcsi-cephfs-node,secretfile=/tmp/csi/keys/keyfile-111413977,mds_namespace=myfs,context="system_u:object_r:container_file_t:s0:c0,c1",_netdev]
```


I've reverted back to just checking the logs. 
The selinux context label is unique and since the mount is successful, I think it should be sufficient for now.

e2e/utils.go
Rakshith-R: > ```
> e2e/utils.go:786:7: shadow: declaration of "err" shadows declaration at line 774 (govet)
> 		pv, err := getBoundPV(f.ClientSet, pvc)
> ```

Current version of pr does not suffer from this error. 
ptal.

e2e/pod.go
nixpanic: it would be much better to validate both in this case. Check the logs for the full `context=....` option, and use `findmnt` to check if the `context=` option was passed.

nixpanic: something like

```
$ findmnt --json --output VFS-OPTIONS --source ...
```

Rakshith-R: @nixpanic 
I don't think its possible in our ci. 
This issue is the same as the one in https://github.com/ceph/ceph-csi/issues/2295#issue-946287829.

> AFAICS this happens because libmount silently drops selinux-related mount options if it doesn't detect that selinux is enabled: https://github.com/karelzak/util-linux/blob/17fc869/libmount/src/context_mount.c#L291

Madhu-1 (dismissed):

Rakshith-R:
Just modified the function name and definition + rebased on top of devel.

PTAL
Thanks,

Madhu-1 (dismissed):

e2e/pod.go
nixpanic: ok, so minikube does not use SELinux :disappointed:

e2e/pod.go
```diff
@@ -542,3 +553,73 @@ func validateRWOPPodCreation(
...
 }
+
+// verifySeLinuxMountOption verifies the SeLinux context MountOption added to PV.Spec.MountOption
+// is successfully used by nodeplugin during mounting by checking for its presence in the
+// nodeplugin container logs.
+func verifySeLinuxMountOption(
+	f *framework.Framework,
+	pvcPath, appPath, daemonSetName, cn, ns string,
+) error {
+	mountOption := "context=\"system_u:object_r:container_file_t:s0:c0,c1\""
```
nixpanic, line 564: so. minikube does not have SELinux, and the contents of the  `context=` would therefore not matter, as it is ignored. What happens when you run this test on a system where SELinux is enforced, will mounting (or `createApp()` below) fail?

Rakshith-R: > so. minikube does not have SELinux, and the contents of the `context=` would therefore not matter, as it is ignored. What happens when you run this test on a system where SELinux is enforced, will mounting (or `createApp()` below) fail?

No, this label would be applied on mount point during mounting. 
Depending on the pod security context selinux options(matches the passed context or not), CRI relabelling may or may not change the already applied label in general case after mounting by csidriver.

But with the KEP, feature gates and RWOP, kublet will pass this context as mountOption automatically. 
So this e2e tries to mimic the same. 
https://kubernetes.io/blog/2023/04/18/kubernetes-1-27-efficient-selinux-relabeling-beta/#mounting-with-selinux-context

Rakshith-R: There's one case where selinux relabelling is not done by CRI, when selinux type is "spc_t" (privileged container).

In openshift, I created a cephfs pvc with selinux label mountOption and mounted it with such a pod.
```
root@csicephfs-demo-pod-3:/#  ls -liaZ /var/lib/www/html
total 0
1099511627785 drwxr-xr-x. 2 root root system_u:object_r:container_file_t:s0:c244,c505  0 Jun 15 06:55 .
     98598060 drwxr-xr-x. 3 root root system_u:object_r:container_file_t:s0:c193,c596 18 Jun 15 07:01 ..
```
Here we can see clearly that context of the mount location is different from the other directories which have the pod's default context.

So this confirms passing context in mountOptions definitely works as intended.

nixpanic: My concern was that if mounting fails, running e2e on a different environment fails as well, which would not be intended. Thanks for checking!

nixpanic (dismissed):

nixpanic:
@mergifyio queue

Rakshith-R:
k8s 1.24 ci tests will fail since selinuxMount in csidriver object is not defined.
(feature gates selinuxMountPod went alpha in k8s v1.25)

@Madhu-1 @nixpanic 
k8s 1.24 will be EOL at end of next month 
Can we remove release 1.24 from ci ? 
https://kubernetes.io/releases/#release-v1-24

```
[FAILED] failed to create resource: failed to create resource "../deploy/cephfs/kubernetes/csidriver.yaml": failed to run kubectl: error running /usr/bin/kubectl --server=https://192.168.49.2:8443/ --kubeconfig=/root/.kube/config --namespace=cephcsi-e2e-c9811c74 create -f -:

  Command stdout:
  stderr:

  error: error validating "STDIN": error validating data: ValidationError(CSIDriver.spec): unknown field "seLinuxMount" in io.k8s.api.storage.v1.CSIDriverSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

external-storage tests are failing due to presence of debug in here https://github.com/ceph/ceph-csi/blob/devel/scripts/k8s-storage/sc-cephfs.yaml.in#L19. 
I'll make this change after we decide on the above 1.24 removal

Madhu-1:
> nuxMount in csidriver object is not defined

+1 to Remove support for 1.24 as its near to EOL. we already have a release which support 1.24 and I think we don't make a release before this month end.

nixpanic:
> k8s 1.24 will be EOL at end of next month
> Can we remove release 1.24 from ci ?

I think that is fine, we do not want to support 1.24 with the next release. As we don't make major changes to the active release branches, I do not think we introduce regressions.

Why would **ci/centos/mini-e2e-helm/k8s-1.24** not fail though?

Rakshith-R:
Edited sc for external storage to remove debug mountoptions.
This external storage e2e passed beyond sc deployment.

https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/k8s-e2e-external-storage-1.26/detail/k8s-e2e-external-storage-1.26/334/pipeline

@mergifyio rebase

This pr is ready to be reviewed again, ptal, thanks.

Rakshith-R:
@mergifyio rebase

nixpanic (dismissed):

e2e/pod.go
```diff
@@ -542,3 +553,73 @@ func validateRWOPPodCreation(
...
+
+	// create PVC
+	pvc, err := loadPVC(pvcPath)
+	if err != nil {
+		framework.Failf("failed to load pvc: %v", err)
```
Madhu-1: It would be best to return an error from this point. We shouldn't let the internal function fail due to this error. Instead, we should return an error in all cases and let the caller function handle the failure.

Madhu-1 (commented):
small nit. LGTM

e2e/pod.go
Rakshith-R: > It would be best to return an error from this point. We shouldn't let the internal function fail due to this error. Instead, we should return an error in all cases and let the caller function handle the failure.

Thanks !
missed it while abstracting the calls to a helper function.
its fixed now.

e2e/pod.go
```diff
@@ -542,3 +553,73 @@ func validateRWOPPodCreation(
...
+	pvc.Namespace = f.UniqueName
+	err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+	if err != nil {
+		return fmt.Errorf("failed to create PVC: %w", err)
+	}
+	// modify PV spec.MountOptions
+	pv, err := getBoundPV(f.ClientSet, pvc)
+	if err != nil {
+		fmt.Errorf("failed to get PV: %w", err)
+	}
```
Madhu-1, line 580: Return is missing here.

Madhu-1 (dismissed):

Rakshith-R:
Raised issue for the failed ci. It is not related to this pr.
https://github.com/ceph/ceph-csi/issues/3910

nixpanic (dismissed):

e2e/cephfs_helper.go
```diff
@@ -66,6 +66,12 @@ func createCephfsStorageClass(
 	if err != nil {
 		return err
 	}
+	// TODO: remove this once the ceph-csi driver release-v3.9 is completed
+	// and upgrade tests are done from v3.9 to devel.
+	// The mountOptions from previous are not compatible with NodeStageVolume
+	// request.
+	sc.MountOptions = []string{}
+
```
Rakshith-R, line 74: CephFS upgrade tests pick up cephfs sc from previous release branches which still
have `debug` mountoption in sc yamls. 
Adding this workaround untill we set upgrade test from 3.9 to devel.

I'll add a note regarding this breaking change in release notes too.
(users will need to recreate cephfs storage without debug option and 
remove it from pvs too.)


added issue to track this one https://github.com/ceph/ceph-csi/issues/3911

I'll ask re-review requests after all the ci is green.
(:sweat_smile: this pr seems to be giving more ci issues than expected.)

Rakshith-R:
> /retest ci/centos/mini-e2e-helm/k8s-1.26

reported a new issue with logs here https://github.com/ceph/ceph-csi/issues/3913

Rakshith-R:
Ci is finally green! 🎊
PTAL

Madhu-1 (approved):

nixpanic (approved):

PR #3930, opened 2023-06-23
rbd: (lastSyncInfo)handle last sync duration being empty case

Rakshith-R:
This commit modifies code to handle last sync duration being empty & 0,returning nil & 0 on encountering it respectively. Earlier both case returned 0. Test case is added for the same too.

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -814,9 +814,9 @@ func getLastSyncInfo(description string) (*replication.GetVolumeReplicationInfoR
 		return nil, fmt.Errorf("no snapshot details: %w", corerbd.ErrLastSyncTimeNotFound)
 	}
 	type localStatus struct {
-		LocalSnapshotTime    int64 `json:"local_snapshot_timestamp"`
-		LastSnapshotBytes    int64 `json:"last_snapshot_bytes"`
-		LastSnapshotDuration int64 `json:"last_snapshot_sync_seconds"`
+		LocalSnapshotTime    int64  `json:"local_snapshot_timestamp"`
+		LastSnapshotBytes    int64  `json:"last_snapshot_bytes"`
+		LastSnapshotDuration *int64 `json:"last_snapshot_sync_seconds"`
 	}
 
 	var localSnapInfo localStatus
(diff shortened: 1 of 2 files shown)


internal/csi-addons/rbd/replication_test.go
```diff
@@ -512,6 +512,18 @@ func TestValidateLastSyncInfo(t *testing.T) {
 			},
 			expectedErr: "",
 		},
+		{
+			name: "description with last_snapshot_sync_seconds = 0",
+			//nolint:lll // sample output cannot be split into multiple lines.
+			description: `replaying, {"bytes_per_second":0.0,"bytes_per_snapshot":81920.0,"last_snapshot_sync_seconds":0,
+			"last_snapshot_bytes":81920,"local_snapshot_timestamp":1684675261,"remote_snapshot_timestamp":1684675261,"replay_state":"idle"}`,
```
Madhu-1, line 519: am not sure we can have this case from Ceph, where we will have last_snapshot_bytes but duration is `0`

Madhu-1 (approved):

internal/csi-addons/rbd/replication_test.go
Rakshith-R: > am not sure we can have this case from Ceph, where we will have last_snapshot_bytes but duration is `0`

I did not focus on other parameters.
We can ignore that parameter for this test case since parameters are not interlinked here, this test case only focuses on duration.

internal/csi-addons/rbd/replication.go
```diff
@@ -814,9 +814,9 @@ func getLastSyncInfo(description string) (*replication.GetVolumeReplicationInfoR
 		return nil, fmt.Errorf("no snapshot details: %w", corerbd.ErrLastSyncTimeNotFound)
 	}
 	type localStatus struct {
-		LocalSnapshotTime    int64 `json:"local_snapshot_timestamp"`
-		LastSnapshotBytes    int64 `json:"last_snapshot_bytes"`
-		LastSnapshotDuration int64 `json:"last_snapshot_sync_seconds"`
+		LocalSnapshotTime    int64  `json:"local_snapshot_timestamp"`
+		LastSnapshotBytes    int64  `json:"last_snapshot_bytes"`
+		LastSnapshotDuration *int64 `json:"last_snapshot_sync_seconds"`
```
nixpanic, line 819: if `last_snapshot_sync_seconds` can be empty when Ceph returns this, can `last_snapshot_bytes` not also be empty?

Is it known what Ceph versions return incomplete results?

Rakshith-R: > if `last_snapshot_sync_seconds` can be empty when Ceph returns this, can `last_snapshot_bytes` not also be empty?

Maybe but currently it does not matter since there's no *int64 type in protoc3 sadly. 

Anyway csiaddons defines bytes 0 as empty and sets it to nil in status. 
This pr just rectifies the nil case for last sync duration. 
 
 https://github.com/csi-addons/kubernetes-csi-addons/blob/main/controllers/replication.storage/volumereplication_controller.go#L390-L393
 
> Is it known what Ceph versions return incomplete results?
I asked the same too in previous pr but didn't get an answer.

riya-singhal31 (approved):

PR #4002, opened 2023-07-18
fix when rbdVol is nil, panic occurred

huangzynn:
image
When rbdVol is nil, panic occurred
add if  rbdVol != nil

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -1078,7 +1078,9 @@ func (cs *ControllerServer) CreateSnapshot(
 
 	// Fetch source volume information
 	rbdVol, err := GenVolFromVolID(ctx, req.GetSourceVolumeId(), cr, req.GetSecrets())
-	defer rbdVol.Destroy()
+	if rbdVol != nil {
+		defer rbdVol.Destroy()
+	}
 	if err != nil {
 		switch {
 		case errors.Is(err, ErrImageNotFound):


nixpanic:
Do you have steps to reproduce this problem? It would be good to have a test case, or at least be able to validate the fix.

internal/rbd/controllerserver.go
```diff
@@ -1078,7 +1078,9 @@ func (cs *ControllerServer) CreateSnapshot(
 
 	// Fetch source volume information
 	rbdVol, err := GenVolFromVolID(ctx, req.GetSourceVolumeId(), cr, req.GetSecrets())
-	defer rbdVol.Destroy()
+	if rbdVol != nil {
+		defer rbdVol.Destroy()
```
nixpanic, line 1082: this could be moved after `if err != nil`, no need to check if `rbdVol != nil` in that case.

Rakshith-R (changes requested):
Thanks for the contribution,
Please fix your commit message by referring to https://github.com/ceph/ceph-csi/blob/b1d1d3bf75b8becb427ec494dd6449f7b5bce115/docs/development-guide.md#commit-messages .

huangzynn:
Such as:
image
image
image

nixpanic:
Hi @huangzynn , in order to fix the commitlint issue, please make these changes:

1. squash the commits into one
2. change the subject of the commit to `rbd: move CreateSnapshot defer rbdVol.Destroy() after err != nil`
3. force push the single commit to your GitHub branch

Many thanks!

nixpanic:
> Do you have steps to reproduce this problem? It would be good to have a test case, or at least be able to validate the fix.

So, by reading the screenshots, is it correct to say that the problem can be triggered when creating a VolumeSnapshot of a non-existing PVC?

huangzynn:
> > Do you have steps to reproduce this problem? It would be good to have a test case, or at least be able to validate the fix.
> 
> So, by reading the screenshots, is it correct to say that the problem can be triggered when creating a VolumeSnapshot of a non-existing PVC?
yes

internal/rbd/controllerserver.go
```diff
@@ -1092,6 +1091,7 @@ func (cs *ControllerServer) CreateSnapshot(
 
 		return nil, err
 	}
+	defer rbdVol.Destroy()
```
Madhu-1: This can cause a cleanup problem as rbdVol is returned even in case of error. IMO moving this to line 1081 with a check for nil is okay.

huangzynn: That's what I thought from the beginning

Madhu-1:
@huangzynn as suggested here https://github.com/ceph/ceph-csi/blob/b1d1d3bf75b8becb427ec494dd6449f7b5bce115/docs/development-guide.md#commit-messages you need to update the commit to fix the commitlint CI problem. @nixpanic posted a sample for you above. you are missing `rbd:` prefix in commit message. job output here https://github.com/ceph/ceph-csi/actions/runs/5596968399/jobs/10234777959?pr=4002

internal/rbd/controllerserver.go
nixpanic: `rbdVol` should never be returned in a case of an error. Either `err != nil` or `rbdVol != nil`. The additional if-statement should be useless, and in effect is the same as having the `defer` below the error handling. It adds a bit of complexity, and makes me wonder if `rbdVol == nil` should not be handled further down in the function as well?

nixpanic:
It seems not trivial to reproduce this, when I try, I get the following

```
Status:
  Error:
    Message:     Failed to create snapshot content with error snapshot controller failed to update invalid-pvc-pr-4002 on API server: cannot get claim from snapshot
    Time:        2023-07-19T10:55:43Z
  Ready To Use:  false
Events:
  Type     Reason                         Age    From                 Message
  ----     ------                         ----   ----                 -------
  Warning  SnapshotContentCreationFailed  2m39s  snapshot-controller  Failed to create snapshot content with error snapshot controller failed to update invalid-pvc-pr-4002 on API server: cannot get claim from snapshot
```

Maybe the PCV needs to exist to pass the snapshot-controller check, and get deleted before Ceph-CSI acts on it?

huangzynn:
> It seems not trivial to reproduce this, when I try, I get the following
> 
> ```
> Status:
>   Error:
>     Message:     Failed to create snapshot content with error snapshot controller failed to update invalid-pvc-pr-4002 on API server: cannot get claim from snapshot
>     Time:        2023-07-19T10:55:43Z
>   Ready To Use:  false
> Events:
>   Type     Reason                         Age    From                 Message
>   ----     ------                         ----   ----                 -------
>   Warning  SnapshotContentCreationFailed  2m39s  snapshot-controller  Failed to create snapshot content with error snapshot controller failed to update invalid-pvc-pr-4002 on API server: cannot get claim from snapshot
> ```
> 
> Maybe the PCV needs to exist to pass the snapshot-controller check, and get deleted before Ceph-CSI acts on it?


What is your snapshot controller version？
My snapshot-controller is v6.2.2

nixpanic:
> My snapshot-controller is v6.2.2

My cluster has v6.2.1... not sure if there is a big difference between those.

humblec:
@huangzynn can you please provide the below details from this cluster.

- [x] The exact yaml of volume snapshot object and `#kubectl get pvc` output from the namespace
- [x] The snapshot controller and external snapshotter version
- [x] Kubernetes cluster and Ceph CSI driver version

huangzynn:
> * kubectl get pvc

image
image
csi-snapshotter:v6.2.2
snapshot-controller:v6.2.2
cephcsi:v3.6.2
Kubernetes:v1.21.5

PR #4053, opened 2023-08-16
rebase: bump k8s.io dependencies to v0.26.7

karthik-us:
Bumping the k8s.io dependencies to v0.26.7. The main intention
is to update the client-go library dependency for kube 1.27+,
where aggregated discovery is enabled. With client-go lower than
v0.26.4 in some cases it is possible that the discovery response
comes with the malformed response with nil GVK leading to Crash
Loop Back Off state.

Upstream kubernetes issue:
https://github.com/kubernetes/kubernetes/pull/116603


karthik-us:
@Madhu-1 PTAL.

Madhu-1:
@karthik-us can you please update the description about reason for this change

go.mod
```diff
@@ -180,12 +180,12 @@ replace (
 	//
 	// k8s.io/kubernetes depends on these k8s.io packages, but unversioned
 	//
-	k8s.io/api => k8s.io/api v0.26.1
+	k8s.io/api => k8s.io/api v0.26.7
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.26.1
-	k8s.io/apimachinery => k8s.io/apimachinery v0.26.1
+	k8s.io/apimachinery => k8s.io/apimachinery v0.26.7
 	k8s.io/apiserver => k8s.io/apiserver v0.26.1
 	k8s.io/cli-runtime => k8s.io/cli-runtime v0.26.1
```
nixpanic: Is there a reason to not update all k8s.io packages at once?

Madhu-1: Yes we can change it as well 👍🏻 as the bug is in the client-go package it's only updated and not sure if updating other dependencies might cause some other regression. @nixpanic whats your suggestion?

nixpanic: I guess it is recommended to update to the latest versions when possible. That will include bugfixes and so on. Mixing different versions is mst likely less well tested than using the same version for everything. In case we hit an unidentified regression, we can always downgrade again. I hope that the projects we use have good CI, and regressions are kept to a minimum.

karthik-us: Thanks @nixpanic & @Madhu-1, will update all the packages in that case.

karthik-us:
> @karthik-us can you please update the description about reason for this change

Done.

go.mod
karthik-us: Done.

Madhu-1 (approved):

nixpanic (approved):

riya-singhal31 (approved):

Madhu-1:
@mergifyio queue

PR #4061, opened 2023-08-21
ci: run versioned k8s jobs only on selected branches

### LINKED ISSUE #4060: Consider running tests with different kubernetes version in different cephcsi branches
> @nixpanic as per official doc it's not reached eol yet https://kubernetes.io/releases/

Maybe we should consider testing with different Kubernetes versions for different branches? The GitHub Commenter action that we use with `ok-to-test`  might be able to support comments depending on the branch? It would speed up the e2e a little, as test jobs for 4 Kubernetes versions will take quite some time (not sure if they can all run in parallel due to CI quota).

_Originally posted by @nixpanic in https://github.com/ceph/ceph-csi/issues/4057#issuecomment-1685788519_

nixpanic:
By using a matrix strategy with excluding certain branches and Kubernetes versions, the number of CI jobs per PullRequest should stay limited.

Closes: #4060

---

<details>

</details>

### DIFF
--- a/.github/workflows/pull-request-commentor.yaml
+++ b/.github/workflows/pull-request-commentor.yaml
@@ -8,111 +8,70 @@ on:
       - "release-v*"
     types:
       - labeled
+
+permissions:
+  pull-requests: write
+
 jobs:
-  add-comment:
+  branch-k8s-selection:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        branch: [release-v3.8, release-v3.9, devel]
+        k8s: ["1.25", "1.26", "1.27", "1.28"]
+        exclude:
+          # the next Ceph-CSI version will not be tested with old Kubernetes
+          - k8s: "1.25"
+            branch: devel
+          # Ceph-CSI <= 3.9 was released before Kubernetes 1.28
+          - k8s: "1.28"
+            branch: release-v3.8
+          - k8s: "1.28"
+            branch: release-v3.9
+
+    if: >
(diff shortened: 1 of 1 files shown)


nixpanic:
This can only be tested once merged. The GitHub Workflow that is triggered when labelling comes from the default (`devel`) branch of the repository.

.github/workflows/pull-request-commentor.yaml
```diff
@@ -8,111 +8,70 @@ on:
...
-            /test ci/centos/mini-e2e-helm/k8s-1.26
-
-      - name: Add comment to trigger helm E2E tests for Kubernetes 1.27
-        uses: peter-evans/create-or-update-comment@v3
-        with:
-          token: ${{ secrets.CEPH_CSI_BOT_TOKEN }}
-          issue-number: ${{ github.event.pull_request.number }}
-          body: |
-            /test ci/centos/mini-e2e-helm/k8s-1.27
+    if: >
```
Madhu-1, line 33: @nixpanic why do we need this if check here?

nixpanic: We had the if statement already, so that it filters on the `ok-to-test` label and skips PRs that are closed. The check has been extended to filter on the target branch of the PR as well.

Each job has an if-statement like this, mostly for the check on the `ok-to-test` label.

Madhu-1 (approved):

riya-singhal31 (approved):

PR #4068, opened 2023-08-22
ci: exclude branches from the testing matrix for `ok-to-test` comments

nixpanic:
It seems that `matrix.*` parameters can not be used in the if-statement for a job. Now using the `exclude:` parameter with a more dynamically constructed value for the branch. If the value for the branch is not part of the initial branch list, the value will not be excluded, so the jobs are expected to run.

---

<details>

</details>

### DIFF
--- a/.github/workflows/pull-request-commentor.yaml
+++ b/.github/workflows/pull-request-commentor.yaml
@@ -21,19 +21,30 @@ jobs:
         branch: [release-v3.8, release-v3.9, devel]
         k8s: ["1.25", "1.26", "1.27", "1.28"]
         exclude:
+          # The exclude items below are dynamically constructed. If the
+          # base_ref matches the given branch, it will be excluded from the
+          # test matrix.
+
           # the next Ceph-CSI version will not be tested with old Kubernetes
           - k8s: "1.25"
-            branch: devel
+            branch: >
+              ${{ "devel" == github.base_ref
+              && github.base_ref || "no-exclude" }}
+
           # Ceph-CSI <= 3.9 was released before Kubernetes 1.28
           - k8s: "1.28"
-            branch: release-v3.8
+            branch: >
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-commentor.yaml
```diff
@@ -21,19 +21,30 @@ jobs:
...
+          # The exclude items below are dynamically constructed. If the
+          # base_ref matches the given branch, it will be excluded from the
+          # test matrix.
+
           # the next Ceph-CSI version will not be tested with old Kubernetes
           - k8s: "1.25"
-            branch: devel
+            branch: >
+              ${{ "devel" == github.base_ref
+              && github.base_ref || "no-exclude" }}
```
Madhu-1, line 32: @nixpanic can you point me to the reference to no-exclude? i tried to find one but didn't get any, how about below version 

```yaml
   strategy:
      matrix:
        k8s: ["1.25", "1.26", "1.27", "1.28"]
        branch:
          - github.base_ref
        exclude:
          # the next Ceph-CSI version will not be tested with old Kubernetes
          - k8s: "1.25"
            branch: devel
          # Ceph-CSI <= 3.9 was released before Kubernetes 1.28
          - k8s: "1.28"
            branch: release-v3.8
          - k8s: "1.28"
            branch: release-v3.9

    if: >
      (github.event.label.name == 'ok-to-test' &&
      github.event.pull_request.merged != true )
```

nixpanic: `no-exclude` is just a placeholder. The `branch:` value is expanded to the name of the branch, or the non-existing branch `no-exclude`. If the `branch:` argument is a non-existing branch, the Kubernetes version will not be excluded (hence included and tested).

Madhu-1 (approved):

riya-singhal31 (approved):

PR #4076, opened 2023-08-24
rbd: Dont depend on image state to issue resync

Madhu-1:
Previously we were dependent on the image mirror state to issue the resync command, What we came to know is that we cannot depend on the image mirror state as the state can change anytime from one to another, We are following the below steps to fix this problem.

* During the Demote volume store the image creation timestamp. 
* During Resync do the below operation
    * Check the image creation timestamp stored during the Demote operation and the current creation timestamp during Resync and check both are equal and it for force resync then issue resync
    * If the image on both sides is not in the unknown state, check last_snapshot_timestamp on the local mirror description,  if it's present send volumeReady as false, or else return an error message.
    * If both the images are in up+unknown the send volumeReady as true.

Storing the image creation timestamp on the rbd image as metadata and it gets cleaned up when the image is recreated as it gets synced up from the primary cluster.

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -50,6 +50,11 @@ const (
 	// imageMirrorModeJournal uses journaling to propagate RBD images between
 	// ceph clusters.
 	imageMirrorModeJournal imageMirroringMode = "journal"
+
+	// imageCreationTimeKey is the key to get/set the image creation timestamp
+	// on the image metadata. The key is starting with `.rbd` so that it will
+	// not get replicated to remote cluster.
+	imageCreationTimeKey = ".rbd.image.creation_time"
 )
 
 const (
@@ -480,6 +485,14 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 
 		return nil, err
 	}
+
+	creationTime, err := rbdVol.GetImageCreationTime()
+	if err != nil {
+		log.ErrorLog(ctx, err.Error())
+
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
 	mirroringInfo, err := rbdVol.GetImageMirroringInfo()
 	if err != nil {
 		log.ErrorLog(ctx, err.Error())
@@ -497,6 +510,17 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 
 	// demote image to secondary
 	if mirroringInfo.Primary {
+		// store the image creation time for resync
+		_, err = rbdVol.GetMetadata(imageCreationTimeKey)
+		if err != nil && errors.Is(err, librbd.ErrNotFound) {
+			err = rbdVol.SetMetadata(imageCreationTimeKey, timestampToString(creationTime))
+		}
+		if err != nil {
+			log.ErrorLog(ctx, err.Error())
+
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
 		err = rbdVol.DemoteImage()
 		if err != nil {
 			log.ErrorLog(ctx, err.Error())
@@ -538,6 +562,8 @@ func checkRemoteSiteStatus(ctx context.Context, mirrorStatus *librbd.GlobalMirro
 // ResyncVolume extracts the RBD volume information from the volumeID, If the
 // image is present, mirroring is enabled and the image is in demoted state.
 // If yes it will resync the image to correct the split-brain.
+//
+//nolint:gocyclo,cyclop // TODO: reduce complexity
 func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 	req *replication.ResyncVolumeRequest,
 ) (*replication.ResyncVolumeResponse, error) {
@@ -578,7 +604,7 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		// it takes time for this operation.
 		log.ErrorLog(ctx, err.Error())
 
-		return nil, status.Error(codes.Aborted, err.Error())
+		return nil, status.Errorf(codes.Aborted, err.Error())
 	}
 
 	if mirroringInfo.State != librbd.MirrorImageEnabled {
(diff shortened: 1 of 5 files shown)


Madhu-1:
Looks like yum update on base image is broken :(

internal/csi-addons/rbd/replication.go
```diff
@@ -637,14 +658,28 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		ready = checkRemoteSiteStatus(ctx, mirrorStatus)
 	}
 
-	err = rbdVol.ResyncVol(localStatus, req.Force)
+	err = rbdVol.GetImageInfo()
```
nixpanic: `GetImageInfo()` only returns an error? Can you not return whatever information you need? Checking attributes like `rbdVol.CreatedAt` is dangerous, specially because of the required calling of `GetImageInfo()` (that probably was a reason the function was not public).

Maybe introduce `rbdVol.GetCreationTime()` that internally calls `getImageInfo()` if the creation time is not set yet?

internal/csi-addons/rbd/replication.go
```diff
@@ -40,6 +41,12 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+// imageResyncMap is used to store the volumeID and image creation time for resync.
+// The volumeID gets added to the map when the DemoteVolume RPC is called and
+// removed from the map when the PromoteVolume RPC is called or
+// DisableVolumeReplication RPC is called.
+var imageResyncMap = sync.Map{}
```
nixpanic: How does this get filled when the driver starts? Is it a problem if the state is not known from the beginning?

internal/csi-addons/rbd/replication.go
Madhu-1: @nixpanic yes it internally updates the image details in local structure and returns an error if that fails. am not sure we should start introduces functions to get a field of the rbdVol.

internal/csi-addons/rbd/replication.go
Madhu-1: This gets filled when cephcsi gets the Replication RPC calls, it's not a problem we might end up issuing rbd Resync (in case the CSI pods get restarted)

nixpanic (changes requested):
`GetImgeInfo()` really should be an internal function of `rbdVolume`. A _Get...()_ function should return something in addition to an error.

Using attributes from rbdVolume outside of its own package is not very clean. Use functions that return requires attributes instead (so that the can be populated/checked in rbdVolume functions).

internal/csi-addons/rbd/replication.go
nixpanic: Maybe rename `GetImageInfo()` to `UpdateImageInfo()`, but accessing attributes of rbdVolume outside of the rbs package is not very clean. There is no guarantee that the attribute has been set (unless you remember to call `getImageInfo()`, which is rather error prone.

internal/csi-addons/rbd/replication.go
nixpanic: ok, good to mention that in the comments too.

Madhu-1: done, updated the comment 👍🏻

Madhu-1:
> `GetImgeInfo()` really should be an internal function of `rbdVolume`. A _Get...()_ function should return something in addition to an error.
> 
> Using attributes from rbdVolume outside of its own package is not very clean. Use functions that return requires attributes instead (so that the can be populated/checked in rbdVolume functions).

Added one extra function to return CreationTimestamp, @nixpanic PTAL

internal/csi-addons/rbd/replication.go
```diff
@@ -659,6 +698,29 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+	ctx context.Context,
+	volumeID string,
+	currentImageTime *timestamppb.Timestamp,
+) (bool, error) {
+	savedImageTime, ok := imageResyncMap.Load(volumeID)
+	if !ok {
+		return false, errors.New("failed to get image creation time")
+	}
+
+	log.UsefulLog(ctx, "savedImageTime=%v, currentImageTime=%v", savedImageTime, currentImageTime)
```
nixpanic: nit: remove this, or phrase it a little nicer, or move it to debug level.

internal/rbd/replication.go
```diff
@@ -126,3 +98,23 @@ func (rv *rbdVolume) DisableVolumeReplication(
 
 	return nil
 }
+
+func (rv *rbdVolume) CheckImageIsPrimary() error {
```
nixpanic: I am confused by the naming of the function. I expected a true/false return in case the image is (not) primary...

Instead, if the image is primary, it returns a `ErrInvalidArgument`?

What is the intention of this function, and can it be made clearer by renaming and/or returning a bool? A comment with a description of the function would also help.

Madhu-1: things got messed up when I tried to refractor, for now removed refractoring as it looks complicated to do as part of this PR.

internal/csi-addons/rbd/replication.go
Madhu-1: i removed this logic and using image meta for this operation so that we will not have problem with restart

internal/csi-addons/rbd/replication.go
```diff
@@ -637,14 +663,53 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+	}
+
+	parts := strings.Fields(savedImageTime)
+	if len(parts) != 2 {
+		return nil, status.Errorf(codes.Internal, "failed to parse image creation time: %s", savedImageTime)
+	}
+	if len(strings.Split(parts[0], ":")) != 2 || len(strings.Split(parts[1], ":")) != 2 {
+		return nil, status.Errorf(codes.Internal, "failed to parse image creation time: %s", savedImageTime)
+	}
+	secondsStr := strings.Split(parts[0], ":")[1]
```
yati1998: In my opinion it would be to good to have a different function to parse the creation time and get the required format, we can also further add unit test for it. What do you think @Madhu-1 ?

nixpanic: yes, definitely a good suggestion!

internal/csi-addons/rbd/replication.go
```diff
@@ -497,6 +510,17 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 
 	// demote image to secondary
 	if mirroringInfo.Primary {
+		// store the image creation time for resync
+		_, err = rbdVol.GetMetadata(imageCreationTimeKey)
+		if err != nil && errors.Is(err, librbd.ErrNotFound) {
+			err = rbdVol.SetMetadata(imageCreationTimeKey, creationTime.String())
```
nixpanic: Consider using your own `timestampToString()` and `timestampFromString()` functions. There is no guarantee that `creationTime.String()` produces the same string format across package versions.

nixpanic (changes requested):

internal/rbd/rbd_util.go
```diff
@@ -1558,6 +1558,19 @@ func (rv *rbdVolume) setImageOptions(ctx context.Context, options *librbd.ImageO
 	return nil
 }
 
+// getImageCreationTime returns the creation time of the image. if the image
```
nixpanic: getImageCreationTime -> GetImageCreationTime

internal/csi-addons/rbd/replication.go
```diff
@@ -637,14 +663,36 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+		return nil, status.Errorf(codes.Internal, "failed to get image info: %s", err.Error())
 	}
 
-	err = checkVolumeResyncStatus(localStatus)
+	// image creation time is stored in the image metadata. it looks like
+	// `"seconds:1692879841 nanos:631526669"`
+	savedImageTime, err := rbdVol.GetMetadata(imageCreationTimeKey)
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return nil, status.Errorf(codes.Internal, "failed to get image creation time: %s", err.Error())
```
nixpanic: consider including the image name and key in the error message, if something goes wrong it gives a hint to check while troubleshooting

internal/csi-addons/rbd/replication.go
```diff
@@ -637,14 +663,36 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		ready = checkRemoteSiteStatus(ctx, mirrorStatus)
 	}
 
-	err = rbdVol.ResyncVol(localStatus, req.Force)
+	creationTime, err := rbdVol.GetImageCreationTime()
 	if err != nil {
-		return nil, getGRPCError(err)
+		return nil, status.Errorf(codes.Internal, "failed to get image info: %s", err.Error())
```
nixpanic: update the error message to include the name of the image?

nixpanic (approved):

riya-singhal31 (approved):

riya-singhal31:
@Mergifyio queue

PR #4089, opened 2023-08-29
doc: add design doc for RBD QoS

### LINKED ISSUE #521: librbd QoS settings for RBD based PVs

The ability to set [librbd QoS settings](https://docs.ceph.com/docs/nautilus/rbd/rbd-config-ref/#rbd-qos-settings) on a PV to limit how much IO can be consumed from the Ceph Cluster. 

The exactly limits would be informed through the storage-class configuration. Ideally we would support three different types of limits:

1. static rbd_qos_iops_limit and rbd_qos_bps_limit per volume
2. dynamic rbd_qos_iops_limit and rbd_qos_bps_limit per volume as a function of the PV size (eg. 3 IOPS per GB, 100 MB/s per TB with a configurable rbd_qos_schedule_tick_min.

A PVC could specify the number of IOPs from storage classes of the second type, but it would adjust the capacity requested based on the above ratio configured in the storage class definition.

# What is the value to the end user? #

Many users were frustrated by IO noisy neighbor issues in early Ceph deployments that were catering to OpenStack environments. Folks started to implement QEMU throttling at the virtio-blk/scsi and this became much more manageable. [Capacity based IOPs](https://docs.openstack.org/cinder/latest/admin/blockstorage-capacity-based-qos.html) further improved on the situation by providing familiar a public cloud like experience (vs static per volume limits).

We want Kubernetes and OpenShift users to have improved noisy neighbor isolation too!

# How will we know we have a good solution? #

1. Configure ceph-csi to use [nbd-rbd approach](https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/csi/example/rbd/storageclass.yaml#L28).
2. Provision volume from storage class as configured above.
3. CSI provisioner would set limit on RBD image.
4. fio test against PV would confirm that IOPs limits were being enforced.

Once resize work is finished, we'll need to ensure new limits are applied when a volume is re-sized.

Madhu-1:
Add design doc for QoS for rbd devices mapped with both krbd and rbd-nbd

closes: #521

Thanks to @idryomov for helping me to understand and experiment on this one 🎉

### DIFF
--- a/docs/design/proposals/rbd-qos.md
+++ b/docs/design/proposals/rbd-qos.md
@@ -1,0 +1,271 @@
+# Design Doc for RBD QoS using cgroup v2
+
+## Introduction
+
+The RBD QoS (Quality of Service) design aims to address the issue of IO noisy
+neighbor problems encountered in early Ceph deployments catering to OpenStack
+environments. These problems were effectively managed by implementing QEMU
+throttling at the virtio-blk/scsi level. To further enhance this,
+capacity-based IOPS were introduced, providing a more dynamic experience
+similar to public cloud environments.
+
+The challenge arises in virtual environments, where a noisy neighbor can lead
+to performance degradation for other instances sharing the same resources.
+Although it's uncommon to observe noisy neighbor issues in Kubernetes
+environments backed by Ceph storage, the possibility exists. The existing QoS
+support with rbd-nbd doesn't apply to krbd, and as rbd-nbd isn't suitable for
+container production workloads, a solution is needed for krbd.
+
+To mitigate resource starvation issues, setting QoS at the device level through
+cgroup v2 when enabled becomes crucial. This approach guarantees that I/O
+capacity isn't overcommitted and is fairly distributed among workloads.
+
+## Dependency
+
+* cgroup v2 must be enabled on the Node
+* We might have Kubernetes dependency as well
+* Container runtime dependency that supports cgroupv2
+
+## Manual steps for implementing RBD QoS in a Kubernetes Cluster
+
+```bash
+[$] ssh root@node1
+sh-4.4# chroot /host
+sh-5.1# cat /proc/partitions
+major minor  #blocks  name
+
+ 259        0  125829120 nvme0n1
+ 259        1       1024 nvme0n1p1
+ 259        2     130048 nvme0n1p2
+ 259        3     393216 nvme0n1p3
+ 259        4  125303791 nvme0n1p4
+ 259        6   52428800 nvme2n1
+   7        0  536870912 loop0
+ 259        5  536870912 nvme1n1
+ 252        0   52428800 rbd0
+sh-5.1#
+```
+
+Once the rbd device is mapped on the node we get the device's major and minor
+number we need to set the io limit on the device but we need to find the right
+cgroup file where we need to set the limit
+
+Kubernetes/Openshift creates a custom cgroup hierarchy for the pods it created
+but start is `/sys/fs/cgroup`  folder
+
+```bash
+sh-5.1# cd /sys/fs/cgroup/
+sh-5.1# ls
+cgroup.controllers cgroup.subtree_control cpuset.mems.effective  io.stat   memory.reclaim   sys-kernel-debug.mount
+cgroup.max.depth cgroup.threads  dev-hugepages.mount    kubepods.slice  memory.stat   sys-kernel-tracing.mount
+cgroup.max.descendants cpu.pressure  dev-mqueue.mount       machine.slice  misc.capacity   system.slice
+cgroup.procs  cpu.stat  init.scope        memory.numa_stat  sys-fs-fuse-connections.mount user.slice
+cgroup.stat  cpuset.cpus.effective io.pressure        memory.pressure  sys-kernel-config.mount
+```
+
+`kubepods.slice` is the starting point and it contains multiple slices
+
+```bash
+sh-5.1# cd kubepods.slice
+sh-5.1# ls
+cgroup.controllers cpuset.cpus    hugetlb.2MB.rsvd.max       memory.pressure
+cgroup.events  cpuset.cpus.effective   io.bfq.weight        memory.reclaim
+cgroup.freeze  cpuset.cpus.partition   io.latency        memory.stat
+cgroup.kill  cpuset.mems    io.max        memory.swap.current
+cgroup.max.depth cpuset.mems.effective   io.pressure        memory.swap.events
+cgroup.max.descendants hugetlb.1GB.current   io.stat        memory.swap.high
+cgroup.procs  hugetlb.1GB.events   kubepods-besteffort.slice      memory.swap.max
+cgroup.stat  hugetlb.1GB.events.local  kubepods-burstable.slice      memory.zswap.current
+cgroup.subtree_control hugetlb.1GB.max    kubepods-pod2b38830b_c2d6_4528_8935_b1c08511b1e3.slice  memory.zswap.max
+cgroup.threads  hugetlb.1GB.numa_stat   memory.current       misc.current
+cgroup.type  hugetlb.1GB.rsvd.current  memory.events        misc.max
+cpu.idle  hugetlb.1GB.rsvd.max   memory.events.local       pids.current
+cpu.max   hugetlb.2MB.current   memory.high        pids.events
+cpu.max.burst  hugetlb.2MB.events   memory.low        pids.max
+cpu.pressure  hugetlb.2MB.events.local  memory.max        rdma.current
+cpu.stat  hugetlb.2MB.max    memory.min        rdma.max
+cpu.weight  hugetlb.2MB.numa_stat   memory.numa_stat
+cpu.weight.nice  hugetlb.2MB.rsvd.current  memory.oom.group
+```
+
+Based on the QoS of the pod, either our application pod will end up in the
+above `kubepods-besteffort.slice` or `kubepods-burstable.slice` or
+`kubepods.slice` (Guaranteed QoS) cgroup. The 3 QoS classes are defined
+[here](https://kubernetes.io/docs/concepts/workloads/pods/pod-QoS/#quality-of-service-classes)
+
+To identify the right cgroup file, we need pod UUID and container UUID from the
+`pod yaml` output
+
+```bash
+[$]kubectl get po csi-rbd-demo-pod -oyaml |grep uid
+  uid: cdf7b785-4eb7-44f7-99cc-ef53890f4dfd
+[$]kubectl get po csi-rbd-demo-pod -oyaml |grep -i containerID
+  - containerID: cri-o://77e57fbbc0f0630f41f9f154f4b5fe368b6dcf7bef7dcd75a9c4b56676f10bc9
+[$]kubectl get po csi-rbd-demo-pod -oyaml |grep -i qosClass
+  qosClass: BestEffort
+```
+
+Now check in the `kubepods-besteffort.slice` and identify the right path using
+pod UID and container UID
+
+Before that check `io.max` on the application pod and see if there is any limit
+
+```bash
+[$]kubectl exec -it csi-rbd-demo-pod -- sh
+sh-4.4# cat /sys/fs/cgroup/io.max
+sh-4.4#
+```
+
+Come back to the Node and find the right cgroup scope
+
+```bash
+sh-5.1# cd kubepods-besteffort.slice/kubepods-besteffort-podcdf7b785_4eb7_44f7_99cc_ef53890f4dfd.slice/crio-77e57fbbc0f0630f41f9f154f4b5fe368b6dcf7bef7dcd75a9c4b56676f10bc9.scope/
+
+
+sh-5.1# echo "252:0 wbps=1048576" > io.max
+sh-5.1# cat io.max
+252:0 rbps=max wbps=1048576 riops=max wiops=max
+```
+
+Now go back to the application pod and check if we have the right limit set
+
+```bash
+[$]kubectl exec -it csi-rbd-demo-pod -- sh
+sh-4.4# cat /sys/fs/cgroup/io.max
+252:0 rbps=max wbps=1048576 riops=max wiops=max
+sh-4.4#
+```
+
+Note:- We can only support the QoS that cgroup v2 io controller supports, this
+means that cumulative read+write QoS limits won't be supported.
+
+Below are the configurations that will be supported
+
+|  Parameter     |  Description     |
+|  ---  |  ---  |
+|  MaxReadIOPS     | Max read IO operations per second      |
+|  MaxWriteIOPS     | Max write IO operations per second      |
+|  MaxReadBytesPerSecond     |  Max read bytes per second     |
+|  MaxWriteBytesPerSecond     |  Max write bytes per second     |
+
+## Different approaches
+
+The above solution can be implemented using 3 different approaches.
+
+### 1. QoS using new parameters in RBD StorageClass
+
+```yaml
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+   name: csi-rbd-sc
+provisioner: rbd.csi.ceph.com
+parameters:
+  MaxReadIOPS: ""
+  MaxWriteIOPS: ""
+  MaxReadBytesPerSecond: ""
+  MaxWriteBytesPerSecond: ""
+```
+
+#### Implementation for StorageClass QoS
+
+1. Create new storageClass with new parameters for QoS
+1. Modify CSIDriver object to pass pod details to the NodePublishVolume CSI
+   procedure
+1. During NodePublishVolume CSI procedure
+   * Retrieve the QoS configuration from the volumeContext in NodePublishRequest
+   * Identify the rbd device using the NodeStageVolumePath
+   * Get the pod UUID from the NodeStageVolume
+   * Set io.max file in all the containers in the pod
+
+#### Drawbacks of StorageClass QoS
+
+1. No way to update the QoS at runtime
+1. Need to take a backup and restore to New QoS StorageClass
+1. Delete and Recreate the PV object
+
+### 2. QoS using parameters in VolumeAttributeClass
+
+```yaml
+apiVersion: storage.k8s.io/v1alpha1
+kind: VolumeAttributesClass
+metadata:
+  name: silver
+parameters:
+  MaxReadIOPS: ""
+  MaxWriteIOPS: ""
+  MaxReadBytesPerSecond: ""
+  MaxWriteBytesPerSecond: ""
+```
+
+VolumeAttributesClassName is a new parameter in the PVC object the user can
+choose from and this can also be updated or removed later.
+
+This new VolumeAttributeClass is designed to keep storage that supports setting
+QoS at the storage level which means setting some configuration at the storage
+(like QoS for nbd)
+
+#### Implementation of VolumeAttributeClass QoS
+
+1. Modify CSIDriver object to pass pod details to the NodePublishVolume CSI
+   procedure
+1. Add support in Ceph-CSI to expose ModifyVolume CSI procedure
+1. Ceph-CSI will store QoS in the rbd image metadata
+1. During NodeStage operation retrieve the image metadata and store it in
+   stagingPath
+1. Whenever a new pod comes in apply the QoS
+
+#### Drawbacks of VolumeAttributeClass QoS
+
+One problem with above is all application need to be scaled downed and scaled
+up to get the new QoS value even though its changed in the PVC object, this is
+sometime impossible as it will have downtime.
+
+### 3. QoS using parameters in VolumeAttributeClass with NodePublish Secret
+
+1. Modify CSIDriver object to pass pod details to the NodePublishVolume CSI
+   procedure
+1. Add support in Ceph-CSI to expose ModifyVolume CSI procedure
+1. Ceph-CSI will store QoS in the rbd image metadata
+1. During NodePublishVolume operation retrieve the QoS from image metadata
+1. Whenever a new pod comes in apply the QoS
+
+This solution addresses the aforementioned issue, but it requires a secret to
+communicate with the ceph cluster. Therefore, we must create a new
+PublishSecret for the storageClass, which may be beneficial when Kubernetes
+eventually enables Node operations.
+
+Both options 2 and 3 are contingent upon changes to the CSI spec and Kubernetes
+support. Additionally,
+[VolumeAttributeClass](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/3751-volume-attributes-class/README.md)
+is currently being developed within the Kubernetes realm and will initially be
+in the Alpha stage. Consequently, it will be disabled by default.
+
+#### Advantages of QoS using VolumeAttributeClass
+
+1. No Restore/Clone operation is required to change the QoS
+1. Easily QoS can be changed for existing PVC only with second approach not
+   with third as it needs new secret.
+
+### Hybrid Approach
+
+Considering the advantages and drawbacks, we can use StorageClass and
+VolumeAttributeClass to support QoS, with VolumeAttributeClass taking
+precedence over StorageClass. This approach offers a flexible solution that
+accounts for dynamic changes while addressing the challenges of existing
+approaches.
+
+### References
+
+Some of the useful links that helped me to understand cgroup v2 and how to set
+QoS on the device.
+
+* [Kubernetes cgroup v2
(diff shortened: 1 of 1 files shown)


Madhu-1:
@travisn PTAL

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
+# Design Doc for RBD QOS using cgroup v2
```
nixpanic: QoS is probably the more common abbreviation?

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+sh-5.1# cd /sys/fs/cgroup/
+sh-5.1# ls
+cgroup.controllers cgroup.subtree_control cpuset.mems.effective  io.stat   memory.reclaim   sys-kernel-debug.mount
+cgroup.max.depth cgroup.threads  dev-hugepages.mount    kubepods.slice  memory.stat   sys-kernel-tracing.mount
+cgroup.max.descendants cpu.pressure  dev-mqueue.mount       machine.slice  misc.capacity   system.slice
+cgroup.procs  cpu.stat  init.scope        memory.numa_stat  sys-fs-fuse-connections.mount user.slice
+cgroup.stat  cpuset.cpus.effective io.pressure        memory.pressure  sys-kernel-config.mount
+```
+
+`kubepods.slice`` is the starting point and it contains multiple hierarchy
```
nixpanic: one ` too many

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+
+sh-5.1# echo "252:0 wbps=1048576" > io.max
+sh-5.1# cat io.max
+252:0 rbps=max wbps=1048576 riops=max wiops=max
+```
+
+Now go back to the application pod and check if we have the right limit set
+
+```bash
+[$]oc exec: oc exec -it csi-rbd-demo-pod -- sh
```
nixpanic: use `kubectl` examples?

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+sh-4.4#
+```
+
+Note:- We can only support the QOS that cgroup v2 io controller supports.
+
+Below are the configurations that will be supported
+
+|  Parameter     |  Description     |
+|  ---  |  ---  |
+|  MaxReadIOOperationsPerSecond     | Max read IO operations per second      |
```
nixpanic: just use IOPS, that is common enough, IOOperationsPerSecond sounds very weird.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+|  MaxReadIOOperationsPerSecond     | Max write IO operations per second      |
+|  MaxReadBytesPerSecond     |  Max read bytes per second     |
+|  MaxWriteBytesPerSecond     |  Max write bytes per second     |
+
+## Different approaches
+
+The above solution can be implemented in CephCSI using 3 different approaches.
```
nixpanic: the name of the project is Ceph-CSI :wink:

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### Drawbacks of VolumeAttributeClass QOS
+
+One problem with above is all application need to be scaled downed and scaled
+up to get the new QOS value even though its changed in the PVC object, this is
+sometime impossible as it will have downtime.
+
+### VolumeAttributeClass with NodePublish Secret
+
+1. Modify CSIDriver object to pass pod details to the NodePublish Volume
```
nixpanic: maybe: ... `NodePublishVolume` CSI procedure.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+1. Add support in cephCSI to expose ModifyVolumeRPC call
```
nixpanic: maybe: ... `ModifyVolume` CSI procedure.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+1. CephCSI will store QOS in the rbd image metadata
+1. During NodePublish operation retrieve the QOS from image metadata
```
nixpanic: `NodePublishVolume`

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+1. Whenever a new pod comes in apply the QOS
+
+This solution addresses the aforementioned issue, but it requires a secret to
+communicate with the ceph cluster. Therefore, we must create a new
+PublishSecret for the storageClass, which may be beneficial when Kubernetes
+eventually enables Node operations.
+
+Both options 2 and 3 are contingent upon changes to the CSI spec and Kubernetes
+support. Additionally, VolumeAttributeClass is currently being developed within
+the Kubernetes realm and will initially be in the Alpha stage. Consequently, it
```
nixpanic: add inline link to the KEP please

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+VolumeAttributesClassName is a new parameter in the PVC object the user can
+choose from and this can also be updated or removed later.
+
+This new VolumeAttributeClass is designed to keep storage that supports setting
+QOS at the storage level which means setting some configuration at the storage
+(like QOS for nbd)
+
+#### Implementation of VolumeAttributeClass QOS
+
+1. Modify CSIDriver object to pass pod details to the NodePublish Volume RPC
```
nixpanic: maybe: ... `NodePublishVolume` CSI procedure.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+1. Add support in cephCSI to expose ModifyVolume RPC
```
nixpanic: cephCSI -> Ceph-CSI

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+
+1. No Restore/Clone operation is required to change the QOS
+1. Easily QOS can be changed for existing PVC only with second approach not
+   with third as it needs new secret.
+
+#### Hybrid Approach
+
+Considering the advantages and drawbacks, we can use StorageClass and
+VolumeAttributeClass to support QoS, with VolumeAttributeClass taking
+precedence over StorageClass.
```
nixpanic: Did you consider adding something like `VolumeAttributeClass` to CSI-Addons? That would allow changing the parameters similar to `ModifyVolume` already.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,269 @@
...
+deployments by providing a way to set I/O limits on devices through cgroupv2
+for both krbd and rbd-nbd mapped devices. Different approaches involving
+StorageClass and VolumeAttributesClass have been discussed, each with its own
+advantages and drawbacks. The hybrid approach offers a flexible solution that
+accounts for dynamic changes while addressing the challenges of existing
+approaches.
+
+### References
+
+* <https://kubernetes.io/docs/concepts/architecture/cgroups/>
```
nixpanic: please link with a subject/description of what is at the URL

docs/design/proposals/rbd-qos.md
Madhu-1: removed it now :D

docs/design/proposals/rbd-qos.md
Madhu-1: @nixpanic not as of today as kubernetes is working on it, if not we can look for adding it in future

docs/design/proposals/rbd-qos.md
Madhu-1: @nixpanic any specific format you are looking at, provide a suggestion.

nixpanic: something like

```markdown
* [Kubernetes cgroup v2 Architecture](https://kubernetes.io/docs/concepts/architecture/cgroups/)
```

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,277 @@
...
+1. Whenever a new pod comes in apply the QoS
+
+This solution addresses the aforementioned issue, but it requires a secret to
+communicate with the ceph cluster. Therefore, we must create a new
+PublishSecret for the storageClass, which may be beneficial when Kubernetes
+eventually enables Node operations.
+
+Both options 2 and 3 are contingent upon changes to the CSI spec and Kubernetes
+support. Additionally,
+[VolumeAttributeClass](https://github.com/kubernetes/enhancements/pull/3780) is
```
nixpanic: a link to the actual doc is probably easier for people to read https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/3751-volume-attributes-class/README.md

docs/design/proposals/rbd-qos.md
nixpanic: I've now checked the KEP and it seems to have noted Alpha status for Kubernetes 1.28. I am not sure implementing additional custom parameters in the StorageClass is the right approach if the feature lands in Kubernetes  soon. Having the parameters in the StorageClass means maintaining support for at least 2 configuration directives, which adds work on the testing and maintenance.

Ideally we only need to implement and maintain the `VolumeAttributeClass` CSI procedures that Kubernetes is implementing.  A single standard mechanism benefits us as well as Ceph-CSI users. Putting efforts in getting the Kubernetes work done is important, and if the Ceph-CSI contributors can assists with that, all the better!

Madhu-1: @nixpanic this is deferred for Kubernetes 1.29 https://github.com/kubernetes/enhancements/issues/3751#issuecomment-1640690606 and am not sure it will happen (hope it will happen), and as it goes with Alpha, Beta Feature cycle it takes more time for users to use this in production and not everyone will agree to upgrade to latest kubernetes for this one. I think it better to have both solutions and IMO it is not a maintenance and testing burden for this one as its minimal one, later once VolumeAttributeClass is stable we can deprecate the support for storageclass Qos.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,277 @@
...
+|  MaxReadIOPS     | Max read IO operations per second      |
+|  MaxWriteIOPS     | Max write IO operations per second      |
+|  MaxReadBytesPerSecond     |  Max read bytes per second     |
+|  MaxWriteBytesPerSecond     |  Max write bytes per second     |
+
+## Different approaches
+
+The above solution can be implemented using 3 different approaches.
+
+### QoS using new parameters in RBD StorageClass
```
Rakshith-R:
```suggestion
### 1. QoS using new parameters in RBD StorageClass
```

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,277 @@
...
+   * Get the pod UUID from the NodeStageVolume
+   * Set io.max file in all the containers in the pod
+
+#### Drawbacks of StorageClass QoS
+
+1. No way to update the QoS at runtime
+1. Need to take a backup and restore to New QoS StorageClass
+1. Delete and Recreate the PV object
+
+### Add support for VolumeAttributeClass
```
Rakshith-R:
```suggestion
### 2. QoS using parameters in VolumeAttributeClass
```

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,277 @@
...
+   stagingPath
+1. Whenever a new pod comes in apply the QoS
+
+#### Drawbacks of VolumeAttributeClass QoS
+
+One problem with above is all application need to be scaled downed and scaled
+up to get the new QoS value even though its changed in the PVC object, this is
+sometime impossible as it will have downtime.
+
+### VolumeAttributeClass with NodePublish Secret
```
Rakshith-R:
```suggestion
### 3. QoS using parameters in VolumeAttributeClass with NodePublish Secret
```

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,277 @@
...
+1. Ceph-CSI will store QoS in the rbd image metadata
+1. During NodePublishVolume operation retrieve the QoS from image metadata
+1. Whenever a new pod comes in apply the QoS
+
+This solution addresses the aforementioned issue, but it requires a secret to
+communicate with the ceph cluster. Therefore, we must create a new
+PublishSecret for the storageClass, which may be beneficial when Kubernetes
+eventually enables Node operations.
+
+Both options 2 and 3 are contingent upon changes to the CSI spec and Kubernetes
```
Rakshith-R, line 239: Adding number to option will help differentiating the different options we have. 
It was a bit confusing to differentiate the option using only font size.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,277 @@
...
+support. Additionally,
+[VolumeAttributeClass](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/3751-volume-attributes-class/README.md)
+is currently being developed within the Kubernetes realm and will initially be
+in the Alpha stage. Consequently, it will be disabled by default, but it does
+offer the following advantages.
```
Rakshith-R:
```suggestion
in the Alpha stage. Consequently, it will be disabled by default.

#### Advantages of QoS using VolumeAttributeClass
```

Let's highlight by adding another heading here.

Rakshith-R (commented):
Thanks !
Left a few comments to highlight options and small nits.

The design doc looks very complete and captures everything perfectly to me.

docs/design/proposals/rbd-qos.md
Rakshith-R: > @nixpanic this is deferred for Kubernetes 1.29 [kubernetes/enhancements#3751 (comment)](https://github.com/kubernetes/enhancements/issues/3751#issuecomment-1640690606) and am not sure it will happen (hope it will happen), and as it goes with Alpha, Beta Feature cycle it takes more time for users to use this in production and not everyone will agree to upgrade to latest kubernetes for this one. I think it better to have both solutions and IMO it is not a maintenance and testing burden for this one as its minimal one, later once VolumeAttributeClass is stable we can deprecate the support for storageclass Qos.

+1

Rakshith-R (dismissed):

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,278 @@
...
+cpu.max.burst  hugetlb.2MB.events   memory.low        pids.max
+cpu.pressure  hugetlb.2MB.events.local  memory.max        rdma.current
+cpu.stat  hugetlb.2MB.max    memory.min        rdma.max
+cpu.weight  hugetlb.2MB.numa_stat   memory.numa_stat
+cpu.weight.nice  hugetlb.2MB.rsvd.current  memory.oom.group
+```
+
+Based on the QoS of the pod, either our application pod will end up in the
+above besteffort.slice or burstable.slice cgroup or Guaranteed cgroup, The 3
+QoS class is defined
```
idryomov: Nit: "QoS classes are defined ...".  Also you probably meant to have a dot instead of comma above.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,278 @@
...
+cpu.max   hugetlb.2MB.current   memory.high        pids.events
+cpu.max.burst  hugetlb.2MB.events   memory.low        pids.max
+cpu.pressure  hugetlb.2MB.events.local  memory.max        rdma.current
+cpu.stat  hugetlb.2MB.max    memory.min        rdma.max
+cpu.weight  hugetlb.2MB.numa_stat   memory.numa_stat
+cpu.weight.nice  hugetlb.2MB.rsvd.current  memory.oom.group
+```
+
+Based on the QoS of the pod, either our application pod will end up in the
+above besteffort.slice or burstable.slice cgroup or Guaranteed cgroup, The 3
```
idryomov: So for `Burstable` and `BestEffort`, a scope under `kubepods-burstable.slice` and `kubepods-besteffort.slice` would be adjusted.  How would `Guaranteed` be handled?

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,278 @@
...
+Now go back to the application pod and check if we have the right limit set
+
+```bash
+[$]kubectl exec -it csi-rbd-demo-pod -- sh
+sh-4.4# cat /sys/fs/cgroup/io.max
+252:0 rbps=max wbps=1048576 riops=max wiops=max
+sh-4.4#
+```
+
+Note:- We can only support the QoS that cgroup v2 io controller supports.
```
idryomov: Probably worth calling out that this means that cumulative read+write QoS limits won't be supported.

docs/design/proposals/rbd-qos.md
Madhu-1: `Guaranteed` will end up in the `kubepods.slice` itself, like `kubepodspod2b38830b_c2d6_4528_8935_b1c08511b1e3.slice` pod here.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,279 @@
...
+```bash
+[$]kubectl get po csi-rbd-demo-pod -oyaml |grep uid
+  uid: cdf7b785-4eb7-44f7-99cc-ef53890f4dfd
+[$]kubectl get po csi-rbd-demo-pod -oyaml |grep -i containerID
+  - containerID: cri-o://77e57fbbc0f0630f41f9f154f4b5fe368b6dcf7bef7dcd75a9c4b56676f10bc9
+[$]kubectl get po csi-rbd-demo-pod -oyaml |grep -i qosClass
+  qosClass: BestEffort
+```
+
+Now check in the besteffor.slice and identify the right path using pod UID and
```
idryomov:
```suggestion
Now check kubepods-besteffort.slice and identify the right path using pod UID and
```

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,279 @@
...
+ 252        0   52428800 rbd0
+sh-5.1#
+```
+
+Once the rbd device is mapped on the node we get the device's major and minor
+number we need to set the io limit on the device but we need to find the right
+cgroup file where we need to set the limit
+
+Kubernetes/Openshift creates a custom cgroup hierarchy for the pods it created
+but start is `sys/fs/cgroup`  folder
```
idryomov:
```suggestion
but start is `/sys/fs/cgroup`  folder
```

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,279 @@
...
+cpu.max   hugetlb.2MB.current   memory.high        pids.events
+cpu.max.burst  hugetlb.2MB.events   memory.low        pids.max
+cpu.pressure  hugetlb.2MB.events.local  memory.max        rdma.current
+cpu.stat  hugetlb.2MB.max    memory.min        rdma.max
+cpu.weight  hugetlb.2MB.numa_stat   memory.numa_stat
+cpu.weight.nice  hugetlb.2MB.rsvd.current  memory.oom.group
+```
+
+Based on the QoS of the pod, either our application pod will end up in the
+above besteffort.slice or burstable.slice cgroup or kubepods.slice (Guaranteed
```
idryomov:
```suggestion
above kubepods-besteffort.slice or kubepods-burstable.slice cgroup or kubepods.slice (Guaranteed
```

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,279 @@
...
+sh-5.1# cd /sys/fs/cgroup/
+sh-5.1# ls
+cgroup.controllers cgroup.subtree_control cpuset.mems.effective  io.stat   memory.reclaim   sys-kernel-debug.mount
+cgroup.max.depth cgroup.threads  dev-hugepages.mount    kubepods.slice  memory.stat   sys-kernel-tracing.mount
+cgroup.max.descendants cpu.pressure  dev-mqueue.mount       machine.slice  misc.capacity   system.slice
+cgroup.procs  cpu.stat  init.scope        memory.numa_stat  sys-fs-fuse-connections.mount user.slice
+cgroup.stat  cpuset.cpus.effective io.pressure        memory.pressure  sys-kernel-config.mount
+```
+
+`kubepods.slice` is the starting point and it contains multiple hierarchy
```
idryomov: Perhaps you meant "slices" instead "hierarchy"?  The plural of "hierarchy" is "hierarchies".

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Conclusion
+
+The RBD QoS design aims to mitigate IO noisy neighbor issues in Ceph
+deployments by providing a way to set I/O limits on devices through cgroupv2
+for both krbd and rbd-nbd mapped devices. Different approaches involving
+StorageClass and VolumeAttributesClass have been discussed, each with its own
+advantages and drawbacks. The hybrid approach offers a flexible solution that
+accounts for dynamic changes while addressing the challenges of existing
+approaches.
```
idryomov, line 257: Not sure if this paragraph brings any value -- this is not an academic paper ;)

A sentence like "This would account for dynamic changes while ..." can be added to the "Hybrid approach" section and I would ditch the entire "Conclusion" section otherwise.

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,279 @@
+# Design Doc for RBD QoS using cgroup v2
+
+## Introduction
+
+The RBD QoS (Quality of Service) design aims to address the issue of IO noisy
+neighbor problems encountered in early Ceph deployments catering to OpenStack
+environments. These problems were effectively managed by implementing QEMU
+throttling at the virtio-blk/scsi level. To further enhance this,
+capacity-based IOPs were introduced, providing a more dynamic experience
```
idryomov:
```suggestion
capacity-based IOPS were introduced, providing a more dynamic experience
```

idryomov (commented):
Just a few nits, otherwise LGTM!

docs/design/proposals/rbd-qos.md
Madhu-1: yes thats correct, its slices. Fixed it now

docs/design/proposals/rbd-qos.md
Madhu-1: Removed the Conclusion :D

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,271 @@
...
+cpu.max   hugetlb.2MB.current   memory.high        pids.events
+cpu.max.burst  hugetlb.2MB.events   memory.low        pids.max
+cpu.pressure  hugetlb.2MB.events.local  memory.max        rdma.current
+cpu.stat  hugetlb.2MB.max    memory.min        rdma.max
+cpu.weight  hugetlb.2MB.numa_stat   memory.numa_stat
+cpu.weight.nice  hugetlb.2MB.rsvd.current  memory.oom.group
+```
+
+Based on the QoS of the pod, either our application pod will end up in the
+above `kubepods-besteffort.slice` or `kubepods-besteffort.slice` or
```
idryomov: One of these used to be burstable ;)
```suggestion
above `kubepods-besteffort.slice` or `kubepods-burstable.slice` or
```

Madhu-1: copy paste mistake, fixed it

idryomov (approved):

docs/design/proposals/rbd-qos.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Advantages of QoS using VolumeAttributeClass
+
+1. No Restore/Clone operation is required to change the QoS
+1. Easily QoS can be changed for existing PVC only with second approach not
+   with third as it needs new secret.
+
+### Hybrid Approach
+
+Considering the advantages and drawbacks, we can use StorageClass and
```
travisn, line 253: Is it possible for users to override the QoS in the storage class with the VolumeAttributeClass settings? This sounds good assuming the admin controls which VolumeAttributeClass is available for the user to specify in their PVC. Otherwise, the user could override admin settings in the storage class.

Madhu-1: both storageclass and VolumeAttributeClass are admin created ones and the user will not have access to create one, Yes users can update the VolumeAttributeClass in the PVC later creation but he/she should get the name of the VolumeAttributeClass from the admin.

Rakshith-R (approved):

nixpanic (approved):

travisn (approved):

Madhu-1:
@mergifyio queue

PR #4097, opened 2023-09-02
rbd: discard not found error from GetMetadata

Madhu-1:
During ResyncVolume call, discard not found error from GetMetadata API. If the image gets resynced the local image creation time will be lost, if the key is not present in the image metadata then we can assume that the image is already resynced.


yati1998 (approved):
LGTM

riya-singhal31 (approved):

Rakshith-R (approved):

Madhu-1:
@mergifyio queue

Issue #4098, opened 2023-09-04
csi-rbdplugin crashes when decoding volume ID(CSI identifier) failed


crazytaxii:
# Describe the bug #

csi-rbdplugin will crash when decoding volume ID failed due to nil pointer

# Environment details #

- Image/version of Ceph CSI driver : v3.6.2
- Helm chart version :
- Kernel version : 4.18.0
- Mounter used for mounting PVC (for cephFS its `fuse` or `kernel`. for rbd its
  `krbd` or `rbd-nbd`) : krbd
- Kubernetes cluster version : v1.27.2
- Ceph cluster version : 16.2.13

problematic code:

```go
	rbdVol, err := GenVolFromVolID(ctx, volumeID, cr, req.GetSecrets())
	defer rbdVol.Destroy()
	if err != nil {
		return cs.checkErrAndUndoReserve(ctx, err, volumeID, rbdVol, cr)
	}
```

> `defer` should be placed under error processing

```[tasklist]
### Tasks
```

Madhu-1:
@crazytaxii Do you have any test cases or the steps to produce this panic?

crazytaxii:
the [code](https://github.com/ceph/ceph-csi/blob/v3.9.0/internal/rbd/controllerserver.go#L924-L928) itself is incorrect:

```go
	rbdVol, err := GenVolFromVolID(ctx, volumeID, cr, req.GetSecrets())
	defer rbdVol.Destroy()
	if err != nil {
		return cs.checkErrAndUndoReserve(ctx, err, volumeID, rbdVol, cr)
	}
```

it should be:

```go
	rbdVol, err := GenVolFromVolID(ctx, volumeID, cr, req.GetSecrets())
	if err != nil {
		return cs.checkErrAndUndoReserve(ctx, err, volumeID, rbdVol, cr)
	}
	defer rbdVol.Destroy()
```

PR #4099, opened 2023-09-04
fixed csi-rbdplugin crashes when decoding volume ID(CSI identifier) failed

crazytaxii:

csi-rbdplugin will crash when decoding volume ID failed due to nil pointer

## Related issues ##

https://github.com/ceph/ceph-csi/issues/4098
Closes: #4002

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -922,7 +922,11 @@ func (cs *ControllerServer) DeleteVolume(
 	}
 
 	rbdVol, err := GenVolFromVolID(ctx, volumeID, cr, req.GetSecrets())
-	defer rbdVol.Destroy()
+	defer func() {
+		if rbdVol != nil {
+			rbdVol.Destroy()
+		}
+	}()
 	if err != nil {
 		return cs.checkErrAndUndoReserve(ctx, err, volumeID, rbdVol, cr)
 	}


Madhu-1:
Duplicate of https://github.com/ceph/ceph-csi/pull/4002

crazytaxii:
> Duplicate of #4002

I think my solution is better.

nixpanic (dismissed):

internal/rbd/controllerserver.go
```diff
@@ -922,10 +922,10 @@ func (cs *ControllerServer) DeleteVolume(
 	}
 
 	rbdVol, err := GenVolFromVolID(ctx, volumeID, cr, req.GetSecrets())
-	defer rbdVol.Destroy()
 	if err != nil {
 		return cs.checkErrAndUndoReserve(ctx, err, volumeID, rbdVol, cr)
 	}
+	defer rbdVol.Destroy()
```
Madhu-1: Please note that this case lead to the memory leak problem as rbdVol is returned even in case of error from GenVolFromVolID function

nixpanic: If there is an error, `GenVolFromVolID()` should not return a connected `rbdVol`... If that is the case, `GenVolFromVolID()` should be corrected too.

crazytaxii: > Please note that this case lead to the memory leak problem as rbdVol is returned even in case of error from GenVolFromVolID function

Oh I see, maybe GenVolFromVolID funtion returns error though, the connection has established.

nixpanic (approved):

karthik-us (approved):
LGTM.

riya-singhal31 (approved):

riya-singhal31:
@Mergifyio queue

nixpanic:
@mergifyio rebase

PR #4165, opened 2023-10-06
deploy: support for read affinity options per cluster

### LINKED ISSUE #4161: common: add support for read affinity options per cluster-id from configmap

Currently, read affinity options need to be passed through cmdline args and apply to volumes of all the clusters.

Since cephcsi handles multiple cluster which have may different set of crush location key-value pairs, we need to move
these options to be consumed per cluster-id basis from configmap.

Users can configure different crush location for different cluster.
Ex: crush location labels for internal ceph cluster maybe different from a external ceph cluster.

Concerned options are successfully picked up from configmap per cluster-id.

iPraveenParihar:
- **deploy: support for read affinity options per cluster**
Implemented the capability to include read affinity options 
for individual clusters within the ceph-csi-config ConfigMap. 
This allows users to configure the crush location for each 
cluster separately. The read affinity options specified in 
the ConfigMap will supersede those provided via command 
line arguments.

- **util: moved GetNodeLabels() under internal/util/k8s**
- **util: added read affinity related functions and unit testcases**
- **util: added RunsOnKubernetes() function**

Fixes: #4161 

---

<details>

</details>

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -5,3 +5,8 @@
 - Removed the deprecated grpc metrics flag's in [PR](https://github.com/ceph/ceph-csi/pull/4225)
 
 ## Features
+
+RBD
+
+- Support for configuring read affinity for individuals cluster within the ceph-csi-config
+  ConfigMap in [PR](https://github.com/ceph/ceph-csi/pull/4165)
--- a/charts/ceph-csi-rbd/values.yaml
+++ b/charts/ceph-csi-rbd/values.yaml
@@ -27,6 +27,11 @@ serviceAccounts:
 #       - "<MONValue2>"
 #     rbd:
 #       netNamespaceFilePath: "{{ .kubeletDir }}/plugins/{{ .driverName }}/net"
+#     readAffinity:
+#       enabled: true
+#       crushLocationLabels:
+#         - topology.kubernetes.io/region
+#         - topology.kubernetes.io/zone
 csiConfig: []
 
 # Configuration details of clusterID,PoolID and FscID mapping
--- a/deploy/csi-config-map-sample.yaml
+++ b/deploy/csi-config-map-sample.yaml
@@ -32,6 +32,10 @@ kind: ConfigMap
 # path for the Ceph cluster identified by the <cluster-id>, This will be used
 # by the RBD CSI plugin to execute the rbd map/unmap in the
 # network namespace specified by the "rbd.netNamespaceFilePath".
+# The "readAffinity" fields are used to enable read affinity and pass the crush
+# location map for the Ceph cluster identified by the cluster <cluster-id>,
+# enabling this will add
+# "read_from_replica=localize,crush_location=<label:value>" to the map option.
 # If a CSI plugin is using more than one Ceph cluster, repeat the section for
 # each such cluster in use.
 # NOTE: Changes to the configmap is automatically updated in the running pods,
@@ -66,6 +70,15 @@ data:
         }
         "nfs": {
           "netNamespaceFilePath": "<kubeletRootPath>/plugins/nfs.csi.ceph.com/net",
+        },
+        "readAffinity": {
+          "enabled": "false",
+          "crushLocationLabels": [
+            "<Label1>",
+            "<Label2>"
+            ...
+            "<Label3>"
+          ]
         }
       }
     ]
--- a/docs/deploy-rbd.md
+++ b/docs/deploy-rbd.md
@@ -47,7 +47,7 @@ make image-cephcsi
 | `--maxsnapshotsonimage`  | `450`                         | Maximum number of snapshots allowed on rbd image without flattening                                                                                                                                                                                                                  |
 | `--setmetadata`          | `false`                       | Set metadata on volume                                                                                                                                                                                                                                                               |
 | `--enable-read-affinity` | `false`                       | enable read affinity                                                                                                                                                                                                                                                                 |
-| `--crush-location-labels`| _empty_                       | Kubernetes node labels that determine the CRUSH location the node belongs to, separated by ','                                                                                                                                                                                       |
+| `--crush-location-labels`| _empty_                       | Kubernetes node labels that determine the CRUSH location the node belongs to, separated by ','.<br>`Note: These labels will be replaced if crush location labels are defined in the ceph-csi-config ConfigMap for the specific cluster.`                                                                                                                                                                                       |
 
 **Available volume parameters:**
 
@@ -221,6 +221,12 @@ krbd options during rbd map operation.
 If enabled, this option will be added to all RBD volumes mapped by Ceph CSI.
 Well known labels can be found
 [here](https://kubernetes.io/docs/reference/labels-annotations-taints/).
+
+Read affinity can be configured for individual clusters within the
+`ceph-csi-config` ConfigMap. This allows configuring the crush location labels
+for each ceph cluster separately. The crush location labels specified in the
+ConfigMap will supersede  those provided via command line argument
+`--crush-location-labels`.
 
 >Note: Label values will have all its dots `"."` normalized with dashes `"-"`
 in order for it to work with ceph CRUSH map.
--- a/internal/rbd/driver/driver.go
+++ b/internal/rbd/driver/driver.go
@@ -26,6 +26,7 @@ import (
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
 	"github.com/ceph/ceph-csi/internal/rbd"
 	"github.com/ceph/ceph-csi/internal/util"
+	"github.com/ceph/ceph-csi/internal/util/k8s"
 	"github.com/ceph/ceph-csi/internal/util/log"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -68,14 +69,14 @@ func NewControllerServer(d *csicommon.CSIDriver) *rbd.ControllerServer {
 func NewNodeServer(
 	d *csicommon.CSIDriver,
 	t string,
-	topology map[string]string,
-	crushLocationMap map[string]string,
+	nodeLabels, topology, crushLocationMap map[string]string,
 ) (*rbd.NodeServer, error) {
 	ns := rbd.NodeServer{
-		DefaultNodeServer: csicommon.NewDefaultNodeServer(d, t, topology),
-		VolumeLocks:       util.NewVolumeLocks(),
+		DefaultNodeServer:         csicommon.NewDefaultNodeServer(d, t, topology),
+		VolumeLocks:               util.NewVolumeLocks(),
+		NodeLabels:                nodeLabels,
+		CLIReadAffinityMapOptions: util.ConstructReadAffinityMapOption(crushLocationMap),
 	}
-	ns.SetReadAffinityMapOptions(crushLocationMap)
 
 	return &ns, nil
 }
@@ -87,8 +88,8 @@ func NewNodeServer(
 // setupCSIAddonsServer().
 func (r *Driver) Run(conf *util.Config) {
 	var (
-		err                        error
-		topology, crushLocationMap map[string]string
+		err                                    error
+		nodeLabels, topology, crushLocationMap map[string]string
 	)
 	// update clone soft and hard limit
 	rbd.SetGlobalInt("rbdHardMaxCloneDepth", conf.RbdHardMaxCloneDepth)
@@ -125,11 +126,15 @@ func (r *Driver) Run(conf *util.Config) {
 			})
 	}
 
-	if conf.EnableReadAffinity {
-		crushLocationMap, err = util.GetCrushLocationMap(conf.CrushLocationLabels, conf.NodeID)
+	if k8s.RunsOnKubernetes() {
+		nodeLabels, err = k8s.GetNodeLabels(conf.NodeID)
 		if err != nil {
 			log.FatalLogMsg(err.Error())
 		}
+	}
+
+	if conf.EnableReadAffinity {
+		crushLocationMap = util.GetCrushLocationMap(conf.CrushLocationLabels, nodeLabels)
 	}
 
 	// Create GRPC servers
@@ -140,7 +145,7 @@ func (r *Driver) Run(conf *util.Config) {
 		if err != nil {
 			log.FatalLogMsg(err.Error())
 		}
-		r.ns, err = NewNodeServer(r.cd, conf.Vtype, topology, crushLocationMap)
+		r.ns, err = NewNodeServer(r.cd, conf.Vtype, nodeLabels, topology, crushLocationMap)
 		if err != nil {
 			log.FatalLogMsg("failed to start node server, err %v\n", err)
 		}
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -45,8 +45,10 @@ type NodeServer struct {
 	// A map storing all volumes with ongoing operations so that additional operations
 	// for that same volume (as defined by VolumeID) return an Aborted error
 	VolumeLocks *util.VolumeLocks
-	// readAffinityMapOptions contains map options to enable read affinity.
-	readAffinityMapOptions string
+	// NodeLabels stores the node labels
+	NodeLabels map[string]string
+	// CLIReadAffinityMapOptions contains map options passed through command line to enable read affinity.
+	CLIReadAffinityMapOptions string
 }
 
 // stageTransaction struct represents the state a transaction was when it either completed
@@ -258,11 +260,10 @@ func (ns *NodeServer) populateRbdVol(
 		rv.Mounter = rbdNbdMounter
 	}
 
-	err = getMapOptions(req, rv)
+	err = ns.getMapOptions(req, rv)
 	if err != nil {
 		return nil, err
 	}
-	ns.appendReadAffinityMapOptions(rv)
 
 	rv.VolID = volID
 
@@ -280,14 +281,14 @@ func (ns *NodeServer) populateRbdVol(
 
 // appendReadAffinityMapOptions appends readAffinityMapOptions to mapOptions
 // if mounter is rbdDefaultMounter and readAffinityMapOptions is not empty.
-func (ns NodeServer) appendReadAffinityMapOptions(rv *rbdVolume) {
+func (rv *rbdVolume) appendReadAffinityMapOptions(readAffinityMapOptions string) {
 	switch {
-	case ns.readAffinityMapOptions == "" || rv.Mounter != rbdDefaultMounter:
+	case readAffinityMapOptions == "" || rv.Mounter != rbdDefaultMounter:
 		return
 	case rv.MapOptions != "":
-		rv.MapOptions += "," + ns.readAffinityMapOptions
+		rv.MapOptions += "," + readAffinityMapOptions
 	default:
-		rv.MapOptions = ns.readAffinityMapOptions
+		rv.MapOptions = readAffinityMapOptions
 	}
 }
 
@@ -1377,23 +1378,4 @@ func getDeviceSize(ctx context.Context, devicePath string) (uint64, error) {
 	}
 
 	return size, nil
-}
-
-func (ns *NodeServer) SetReadAffinityMapOptions(crushLocationMap map[string]string) {
-	if len(crushLocationMap) == 0 {
-		return
-	}
-
-	var b strings.Builder
-	b.WriteString("read_from_replica=localize,crush_location=")
-	first := true
-	for key, val := range crushLocationMap {
-		if first {
-			b.WriteString(fmt.Sprintf("%s:%s", key, val))
-			first = false
-		} else {
-			b.WriteString(fmt.Sprintf("|%s:%s", key, val))
-		}
-	}
-	ns.readAffinityMapOptions = b.String()
 }
--- a/internal/util/crushlocation.go
+++ b/internal/util/crushlocation.go
@@ -23,20 +23,15 @@ import (
 )
 
 // GetCrushLocationMap returns the crush location map, determined from
-// the crush location labels and their values from the CO system.
+// the crush location labels and their values from the node labels passed in arg.
 // Expects crushLocationLabels in arg to be in the format "[prefix/]<name>,[prefix/]<name>,...",.
 // Returns map of crush location types with its array of associated values.
-func GetCrushLocationMap(crushLocationLabels, nodeName string) (map[string]string, error) {
+func GetCrushLocationMap(crushLocationLabels string, nodeLabels map[string]string) map[string]string {
 	if crushLocationLabels == "" {
-		return nil, nil
+		return nil
 	}
 
-	nodeLabels, err := k8sGetNodeLabels(nodeName)
-	if err != nil {
-		return nil, err
-	}
-
-	return getCrushLocationMap(crushLocationLabels, nodeLabels), nil
+	return getCrushLocationMap(crushLocationLabels, nodeLabels)
 }
 
 // getCrushLocationMap returns the crush location map, determined from
--- a/internal/util/csiconfig.go
+++ b/internal/util/csiconfig.go
@@ -62,6 +62,11 @@ type ClusterInfo struct {
 		// symlink filepath for the network namespace where we need to execute commands.
 		NetNamespaceFilePath string `json:"netNamespaceFilePath"`
 	} `json:"nfs"`
+	// Read affinity map options
+	ReadAffinity struct {
+		Enabled             bool     `json:"enabled"`
+		CrushLocationLabels []string `json:"crushLocationLabels"`
+	} `json:"readAffinity"`
 }
 
 // Expected JSON structure in the passed in config file is,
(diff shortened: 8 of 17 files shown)


internal/rbd/nodeserver.go
```diff
@@ -1397,8 +1406,28 @@ func getDeviceSize(ctx context.Context, devicePath string) (uint64, error) {
...
+	ns.readAffinityMapOptions = constructReadAffinityMapOption(crushLocationMap)
+}
+
+func (ns *NodeServer) getReadAffinityMapOptionFromConfigMap(rv *rbdVolume) string {
+	crushLocationLabels, err  := util.GetReadAffinityOptions(util.CsiConfigFile, rv.ClusterID)
+	if err != nil {
+		log.FatalLogMsg(err.Error())
+	}
+
+	crushLocationMap, err := util.GetCrushLocationMap(crushLocationLabels, ns.Driver.NodeID)
```
Rakshith-R: This function call k8s api, which is an expensive operation, to get node labels.
Performance will severely be impacted if we perform such a call on each nodeStage request. 

Please fetch the node labels when the driver starts and store it.

iPraveenParihar:
@Rakshith-R @Madhu-1 
PR is ready for a review

internal/rbd/nodeserver.go
```diff
@@ -45,8 +45,9 @@ type NodeServer struct {
 	// A map storing all volumes with ongoing operations so that additional operations
 	// for that same volume (as defined by VolumeID) return an Aborted error
 	VolumeLocks *util.VolumeLocks
-	// readAffinityMapOptions contains map options to enable read affinity.
-	readAffinityMapOptions string
+	NodeLabels  map[string]string
```
nixpanic: please document this new map with a comment

internal/rbd/nodeserver.go
```diff
@@ -280,14 +283,14 @@ func (ns *NodeServer) populateRbdVol(
 
 // appendReadAffinityMapOptions appends readAffinityMapOptions to mapOptions
 // if mounter is rbdDefaultMounter and readAffinityMapOptions is not empty.
-func (ns NodeServer) appendReadAffinityMapOptions(rv *rbdVolume) {
+func appendReadAffinityMapOptions(rv *rbdVolume, readAffinityMapOptions string) {
```
nixpanic: this can now be a function of the `rbdVolume` struct:
```go
func (rv *rbdVolume) appendReadAffinityMapOptions(readAffinityMapOptions string)
```
which would be cleaner

internal/util/crushlocation.go
```diff
@@ -26,16 +26,11 @@ import (
 // the crush location labels and their values from the CO system.
 // Expects crushLocationLabels in arg to be in the format "[prefix/]<name>,[prefix/]<name>,...",.
 // Returns map of crush location types with its array of associated values.
-func GetCrushLocationMap(crushLocationLabels, nodeName string) (map[string]string, error) {
+func GetCrushLocationMap(crushLocationLabels string, nodeLabels map[string]string) (map[string]string, error) {
```
nixpanic: the comment is not correct anymore. The nodeLabels are not retrieved from the CO system within this function.

internal/util/topology.go
```diff
@@ -34,7 +34,7 @@ const (
 	labelSeparator string = ","
 )
 
-func k8sGetNodeLabels(nodeName string) (map[string]string, error) {
+func K8sGetNodeLabels(nodeName string) (map[string]string, error) {
```
nixpanic: maybe you can move this function to a file under `internal/util/k8s/`?

internal/rbd/nodeserver.go
```diff
@@ -1396,9 +1399,46 @@ func getDeviceSize(ctx context.Context, devicePath string) (uint64, error) {
 	return size, nil
 }
 
-func (ns *NodeServer) SetReadAffinityMapOptions(crushLocationMap map[string]string) {
+func (ns *NodeServer) SetcmdReadAffinityMapOptions(crushLocationMap map[string]string) {
+	ns.cmdReadAffinityMapOptions = constructReadAffinityMapOption(crushLocationMap)
```
nixpanic: Is there a good reason to keep `constructReadAffinityMapOption()` as a separate function? Why not make it

```go
func (ns *NodeServer) constructReadAffinityMapOption(crushLocationMap map[string]string)
```

or something similar?

internal/rbd/nodeserver.go
nixpanic: > Please fetch the node labels when the driver starts and store it.

- Is this a problem for a running Ceph-CSI setup when nodes are added later on?
- What happens if the Crush location map is modified?

internal/rbd/nodeserver.go
iPraveenParihar: Certainly, we can make `cmdReadAffinityMapOptions` a public field. This would allow us to directly use it as follows:

```go
ns.cmdReadAffinityMapOptions = ns.constructReadAffinityMapOption(crushLocationMap)
```
This change would eliminate the need for the `SetcmdReadAffinityMapOptions()` method.

internal/rbd/nodeserver.go
Rakshith-R: > > Please fetch the node labels when the driver starts and store it.
> 
> * Is this a problem for a running Ceph-CSI setup when nodes are added later on?

No, We just use this in daemonset nodeplugin pod. 
So, when a new node is added, a daemonset pod starts on it and we'll pick up the labels that time.
There is a requirement currently that when node labels are updated, they'll need to restart nodepugin pods andthe mounted volumes for the changed topology to take affect.

> * What happens if the Crush location map is modified?

If you meant Crush Location map modified by means of node topology label update, it is the same as above.
If you meant changes were done to the configmap to update labels, they'll need to restart the nodeplugin pod too.


I expect these changes to be very rare and possibly untouched after initial creation.

internal/rbd/driver/driver.go
```diff
@@ -125,8 +126,13 @@ func (r *Driver) Run(conf *util.Config) {
 			})
 	}
 
+	nodeLabels, err = k8s.GetNodeLabels(conf.NodeID)
+	if err != nil {
+		log.FatalLogMsg(err.Error())
+	}
```
Madhu-1: This code will cause problems for non-kubernetes CO. do this check only if the labels are passed?

internal/rbd/nodeserver.go
```diff
@@ -1396,9 +1400,42 @@ func getDeviceSize(ctx context.Context, devicePath string) (uint64, error) {
 	return size, nil
 }
 
-func (ns *NodeServer) SetReadAffinityMapOptions(crushLocationMap map[string]string) {
+// getReadAffinityMapOptionFromConfigMap retrieves read affinity map options for cluster `clusterID`.
+func (ns *NodeServer) getReadAffinityMapOptionFromConfigMap(clusterID string) string {
+	crushLocationLabels, err := util.GetReadAffinityOptions(util.CsiConfigFile, clusterID)
+	if err != nil {
+		log.FatalLogMsg(err.Error())
```
Madhu-1: it should not be Fatal you need to return the error message from the function for all cases

internal/util/k8s/node.go
```diff
@@ -0,0 +1,23 @@
+package k8s
```
Madhu-1, line 17: add missing license code

internal/util/k8s/node.go
```diff
@@ -0,0 +1,23 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func GetNodeLabels(nodeName string) (map[string]string, error) {
+	client, err := NewK8sClient()
+	if err != nil {
+		return nil, fmt.Errorf("can not get node %q information, failed "+
+			"to connect to Kubernetes: %w", nodeName, err)
+	}
```
Madhu-1, line 31: can you check is it possible to re-use the client instead of creating one always

internal/rbd/driver/driver.go
iPraveenParihar: Will move this under 
```go
if conf.EnableReadAffinity {
  ...
}
```

and will make `--enable-ready-affinity` command line flag to be mandatory for enabling read affinity.

Madhu-1: i was thinking do we really need to have this flag, if the labels are set assume that you will enable read-affinity like we do currently?

iPraveenParihar: > This code will cause problems for non-kubernetes CO. do this check only if the labels are passed?

@Madhu-1 we can have a new `--container-orchestrator` flag _(default - kubernetes)_. So we will be fetching node labels only when the flag is set to _kubernetes_?

Or this way
> and will make --enable-ready-affinity command line flag to be mandatory for enabling read affinity.

Madhu-1: i would vote for new `--container-orchestrator` which might be useful to decide when to make kube API calls. @nixpanic @Rakshith-R thought?

Rakshith-R: > i was thinking do we really need to have this flag, if the labels are set assume that you will enable read-affinity like we do currently?

labels maybe set in the configmap instead. 

Let's have the read-affinity flag set in cmdline as a pre-requisite. 
This setting can be overridden by individual cluster id in configmap.

> i would vote for new `--container-orchestrator` which might be useful to decide when to make kube API calls. @nixpanic @Rakshith-R thought?

I think this entails changes at a lot of places and is out of scope for this pr.

> and will make --enable-ready-affinity command line flag to be mandatory for enabling read affinity.

This option seems sufficient for me.

charts/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
```diff
@@ -87,8 +87,7 @@ spec:
 {{- if .Values.nodeplugin.profiling.enabled }}
             - "--enableprofiling={{ .Values.nodeplugin.profiling.enabled }}"
 {{- end }}
-            - "--enable-read-affinity={{ and .Values.readAffinity .Values.readAffinity.enabled }}"
```
Madhu-1: any specific reason to remove `enable-read-affinity` flag?

charts/ceph-csi-rbd/values.yaml
```diff
@@ -282,7 +287,6 @@ topology:
 # readAffinity:
 # Enable read affinity for RBD volumes. Recommended to
 # set to true if running kernel 5.8 or newer.
-# enabled: false
```
Madhu-1: lets not remove it, it can cause breaking change for existing clusters

cmd/cephcsi.go
```diff
@@ -81,13 +81,13 @@ func init() {
...
 		"list of Kubernetes node labels, that determines the topology"+
 			" domain the node belongs to, separated by ','")
-	flag.BoolVar(&conf.EnableReadAffinity, "enable-read-affinity", false, "enable read affinity")
 	flag.StringVar(
 		&conf.CrushLocationLabels,
 		"crush-location-labels",
 		"",
 		"list of Kubernetes node labels, that determines the"+
 			" CRUSH location the node belongs to, separated by ','")
+	flag.StringVar(&conf.ContainerOrchestrator, "container-orchestrator", "kubernetes", "container orchestrator")
```
Madhu-1: my suggestion was to add this extra flag but not to remove the `EnableReadAffinity` existing flag :)

docs/deploy-rbd.md
```diff
@@ -224,6 +223,12 @@ If enabled, this option will be added to all RBD volumes mapped by Ceph CSI.
 Well known labels can be found
 [here](https://kubernetes.io/docs/reference/labels-annotations-taints/).
 
+Read affinity can be configured for individual clusters within the
+`ceph-csi-config` ConfigMap. This allows configuring the crush location labels
+for each cluster separately. The crush location labels specified in the
```
Madhu-1: `for each cluster separately` to `for each ceph cluster separately`

internal/rbd/driver/driver.go
```diff
@@ -125,8 +126,13 @@ func (r *Driver) Run(conf *util.Config) {
 			})
 	}
 
-	if conf.EnableReadAffinity {
-		crushLocationMap, err = util.GetCrushLocationMap(conf.CrushLocationLabels, conf.NodeID)
+	nodeLabels, err = k8s.GetNodeLabels(conf.NodeID)
```
Madhu-1: call this function only if `ContainerOrchestrator` is kubernetes

internal/rbd/nodeserver.go
```diff
@@ -1396,9 +1397,12 @@ func getDeviceSize(ctx context.Context, devicePath string) (uint64, error) {
 	return size, nil
 }
 
-func (ns *NodeServer) SetReadAffinityMapOptions(crushLocationMap map[string]string) {
+// ConstructReadAffinityMapOption constructs a read affinity map option based on the provided crushLocationMap.
+// It appends crush location labels in the format
+// "read_from_replica=localize,crush_location=label1:value1|label2:value2|...".
+func ConstructReadAffinityMapOption(crushLocationMap map[string]string) string {
```
Madhu-1: If this function can be reused move this to util as we need this for cephfs as well?

internal/rbd/nodeserver.go
```diff
@@ -1412,5 +1416,39 @@ func (ns *NodeServer) SetReadAffinityMapOptions(crushLocationMap map[string]stri
...
 	}
-	ns.readAffinityMapOptions = b.String()
+
+	return b.String()
+}
+
+// getReadAffinityMapOptions retrieves the readAffinityMapOptions from the CSI config file if it exists.
+// If not, it falls back to returning the `CLIReadAffinityMapOptions` from the command line.
+// If neither of these options is available, it returns an empty string.
+func (ns *NodeServer) getReadAffinityMapOptions(clusterID string) (string, error) {
```
Madhu-1: Move this to until as this can be reused for cephfs asl well?

internal/util/k8s/node.go
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright 2020 The CephCSI Authors.
```
Madhu-1: 2023?

internal/rbd/nodeserver.go
iPraveenParihar: Yea, I thought to move this to util during CephFS work.

cmd/cephcsi.go
```diff
@@ -88,6 +88,8 @@ func init() {
 		"",
 		"list of Kubernetes node labels, that determines the"+
 			" CRUSH location the node belongs to, separated by ','")
+	flag.StringVar(&conf.ContainerOrchestrator, "container-orchestrator", "kubernetes",
```
Madhu-1:
```suggestion
flag.StringVar(&conf.ContainerOrchestrator, "container-orchestrator", "kubernetes", "container orchestrator where cephcsi is running")
```

deploy/csi-config-map-sample.yaml
```diff
@@ -66,6 +66,15 @@ data:
         }
         "nfs": {
           "netNamespaceFilePath": "<kubeletRootPath>/plugins/nfs.csi.ceph.com/net",
+        },
```
Madhu-1, line 73: please add more details about this configuration in this file comments

docs/deploy-rbd.md
```diff
@@ -48,8 +48,7 @@ make image-cephcsi
...
 | `--maxsnapshotsonimage`  | `450`                         | Maximum number of snapshots allowed on rbd image without flattening                                                                                                                                                                                                                  |
 | `--setmetadata`          | `false`                       | Set metadata on volume                                                                                                                                                                                                                                                               |
-| `--enable-read-affinity` | `false`                       | enable read affinity                                                                                                                                                                                                                                                                 |
```
Madhu-1, line 49: This required to be set?

docs/deploy-rbd.md
```diff
@@ -215,7 +214,7 @@ for more details.
 
 This can be enabled by adding labels to Kubernetes nodes like
 `"topology.io/region=east"` and `"topology.io/zone=east-zone1"` and
-passing command line arguments `"--enable-read-affinity=true"` and
```
Madhu-1: Need to revert this back?

internal/util/k8s/client.go
```diff
@@ -46,5 +52,9 @@ func NewK8sClient() (*kubernetes.Clientset, error) {
 		return nil, fmt.Errorf("failed to create client: %w", err)
 	}
 
+	if kubeclient == nil {
```
Madhu-1: This check is not required here?

internal/util/read_affinity.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+
+	return b.String()
+}
+
+// GetReadAffinityMapOptions retrieves the readAffinityMapOptions from the CSI config file if it exists.
+// If not, it falls back to returning the `cliReadAffinityMapOptions` from the command line.
+// If neither of these options is available, it returns an empty string.
+func GetReadAffinityMapOptions(
```
Madhu-1, line 50: Add unit testing for this one?

internal/util/read_affinity.go
```diff
@@ -0,0 +1,82 @@
...
+		return "", nil
+	}
+
+	if configCrushLocationLabels == "" {
+		return cliReadAffinityMapOptions, nil
+	}
+
+	crushLocationMap, err := GetCrushLocationMap(configCrushLocationLabels, nodeLabels)
+	if err != nil {
+		log.FatalLogMsg(err.Error())
```
Madhu-1: log as error and a return err is missing here?

Madhu-1 (commented):
If possible add a test case where crush rules are set in configmap as well.

Madhu-1 (dismissed):
LGTM

internal/util/read_affinity.go
```diff
@@ -0,0 +1,76 @@
+// /*
```
nixpanic: nit: the commenting looks very weird. can you use the same as what is done in other files?

cmd/cephcsi.go
```diff
@@ -88,6 +88,8 @@ func init() {
...
+		"container orchestrator where cephcsi is running")
```
nixpanic: It would be better to auto-detect this. You can check for the `KUBERNETES_SERVICE_HOST` environment variable, for example. When an other CO is supported, detection for that can be added.

A `config.GetContainerOrchestrator()` function can take care of the detection, and return a string constant for Kubernetes (in the `internal/util` package):

```go
type ContainerOrchestrator string

const (
    ContainerOrchestratorKubernetes = ContainerOrchestrator("kubernetes")
)

func (c *Config) GetContainerOrchestrator() ContainerOrchestrator {
    // do detection
    return ContainerOrchestratorKubernetes
}
```

nixpanic (commented):
Some minor changes to improve things. It would be nice to split the commit in several ones, one commit for a particular functionality addition, structure them like:

1. moving of code, from one package to an other (like `GetNodeLabels()`)
2. new util functions
3. using the new util functions
4. deployment changes
5. e2e tests (but that may be hard for this PR)

cmd/cephcsi.go
iPraveenParihar: This will be out of scope for this PR and we can take up this in another PR.

nixpanic: I think it would be better to do it in this PR, so that there is no need to add the `container-orchestrator` parameter, and remove it afterwards (breaks backwards compatibility).

internal/rbd/driver/driver.go
```diff
@@ -125,13 +126,17 @@ func (r *Driver) Run(conf *util.Config) {
 			})
 	}
 
-	if conf.EnableReadAffinity {
-		crushLocationMap, err = util.GetCrushLocationMap(conf.CrushLocationLabels, conf.NodeID)
+	if conf.ContainerOrchestrator == "kubernetes" {
```
nixpanic: `"kubernetes" ` should be a constant, and if set to something else, a warning needs to be emitted.

internal/util/csiconfig.go
```diff
@@ -209,3 +214,20 @@ func GetNFSNetNamespaceFilePath(pathToConfig, clusterID string) (string, error)
 
 	return cluster.NFS.NetNamespaceFilePath, nil
 }
+
+// GetCrushLocationLabels returns the crushLocationLabels from csi config for the given clusterID
+// If `readAffinity.enabled` is set to true, else returns an empty string.
+func GetCrushLocationLabels(pathToConfig, clusterID string) (bool, string, error) {
```
nixpanic, line 215: Please document the `bool` return value as well.

internal/util/csiconfig_test.go
```diff
@@ -365,3 +365,101 @@ func TestGetNFSNetNamespaceFilePath(t *testing.T) {
...
+	tmpConfPath := t.TempDir() + "/ceph-csi.json"
+	err = os.WriteFile(tmpConfPath, csiConfigFileContent, 0o600)
+	if err != nil {
+		t.Errorf("failed to write %s file content: %v", CsiConfigFile, err)
+	}
+	for _, tt := range tests {
+		tc := tt
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			_, got, err := GetCrushLocationLabels(tmpConfPath, tc.clusterID)
```
nixpanic: if there is no need to check the returned bool, don't return it at all. Or validate here what you're expecting.

cmd/cephcsi.go
iPraveenParihar: What about the default value for the flag? 
I was considering setting the flag's default value to an empty string. This way, if the user doesn't pass a value for the flag, we can automatically detect it. Does that sound good?

internal/util/csiconfig_test.go
iPraveenParihar: returned bool value is required and validated in the calling function - `GetReadAffinityMapOptions()` under `internal/util/read_affinity.go`. 

```go
configReadAffinityEnabled, configCrushLocationLabels, err = GetCrushLocationLabels(CsiConfigFile, clusterID)
if err != nil {
  return "", err
}

if !configReadAffinityEnabled {
  return "", nil
}

if configCrushLocationLabels == "" {
  return cliReadAffinityMapOptions, nil
}
```

nixpanic: great, that is an excellent reason to validate the returned bool here too!

cmd/cephcsi.go
nixpanic: Ideally users (or other deployment systems like Rook) do not need to pass a flag at all, so auto-detection should be the default.

If auto-detection is the default, then there is no good reason to have the flag anymore :smiley:

nixpanic: Maybe add a `func RunsOnKubernetes() bool` to the `internal/util/k8s` package?

internal/rbd/driver/driver.go
nixpanic: if `ContainerOrchestrator` is only used here, consider replacing it with `internal/util/k8s.RunsOnKubernetes()` or similar

internal/rbd/nodeserver.go
nixpanic: I have no idea how often locations/labels are added to a crush map :man_shrugging: If it is clearly documented that changing labels of a node requires restarting, then thats fine.

internal/util/csiconfig_test.go
nixpanic: I still think it would be good to verify the returned bool here.

internal/util/k8s/client.go
```diff
@@ -56,3 +56,11 @@ func NewK8sClient() (*kubernetes.Clientset, error) {
 
 	return client, nil
 }
+
+// RunsOnKubernetes checks if the application is running within a Kubernetes cluster
+// by inspecting the presence of the KUBERNETES_SERVICE_HOST environment variable.
+func RunsOnKubernetes() bool {
```
nixpanic, line 62: yay, auto-detection!

Can you move this commit before the one that calls this function? It should be possible to at least build the ceph-csi executable after each commit (otherwise `git bisect` breaks).

iPraveenParihar: Done

internal/util/csiconfig_test.go
iPraveenParihar: Done.

nixpanic (dismissed):
Thanks for all the corrections and improvements!

Madhu-1 (dismissed):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

nixpanic:
It seems the **verify readAffinity support** e2e tests fails.

iPraveenParihar:
> It seems the **verify readAffinity support** e2e tests fails.

Had forgotten to update the csi config in e2e, made the changes.
PTAL.

Madhu-1 (dismissed):

yati1998 (dismissed):

PendingReleaseNotes.md
```diff
@@ -3,3 +3,8 @@
 ## Breaking Changes
 
 ## Features
+
+RBD
```
iPraveenParihar, line 9: @Madhu-1,
added to PendingReleaseNotes.md

Madhu-1 (approved):

Rakshith-R (approved):

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio queue

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

PR #4193, opened 2023-10-16
ci: add capability to run specific e2e tests

karthik-us:
The current e2e test runs will trigger all 3 types of tests which takes a lot of time to complete. If the patch impacts only a single type, then having an option to trigger only that particular e2e test can save
considerable amount of time based on the test type.

This can also help in reducing the time in testing when a patch failed on a particular e2e test type and triggering only that e2e with the fix for the failure or re-trigger only that test again to check for false positives.

### DIFF
--- a/jobs/mini-e2e.yaml
+++ b/jobs/mini-e2e.yaml
@@ -10,6 +10,10 @@
           only_run_on_request: true
       - '1.28':
           only_run_on_request: true
+    test_type:
+      - 'cephfs'
+      - 'rbd'
+      - 'nfs'
     jobs:
       - 'mini-e2e_k8s-{k8s_version}'
       - 'mini-e2e-helm_k8s-{k8s_version}'
@@ -32,6 +36,10 @@
           name: k8s_version
           default: '{k8s_version}'
           description: Kubernetes version to deploy in the test cluster.
+      - string:
+          name: test_type
+          default: '{test_type}'
+          description: Mentions whether tests run for rbd/cephfs/nfs.
     pipeline-scm:
       scm:
         - git:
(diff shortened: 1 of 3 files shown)


Madhu-1:
I assume this will be a breaking change, This might need some more enhancement to detect which tests need to be run and also might require changes in the pull-request commenter as well.

jobs/mini-e2e.yaml
```diff
@@ -32,6 +36,10 @@
           name: k8s_version
           default: '{k8s_version}'
           description: Kubernetes version to deploy in the test cluster.
+      - string:
+          name: test_type
+          default: '{test_type}'
```
Rakshith-R: default should be empty strings ?

jobs/mini-e2e.yaml
```diff
@@ -46,7 +54,7 @@
           status-url: $RUN_DISPLAY_URL
           status-context: 'ci/centos/mini-e2e/k8s-{k8s_version}'
           # yamllint disable-line rule:line-length
-          trigger-phrase: '/(re)?test ci/centos/mini-e2e(/k8s-{k8s_version})?'
+          trigger-phrase: '/(re)?test ci/centos/mini-e2e(/k8s-{k8s_version}-{test_type})?'
```
Rakshith-R:
```suggestion
          trigger-phrase: '/(re)?test ci/centos/mini-e2e(/k8s-{k8s_version}(test_type-{test_type})?)?'
```
it should be optional and with prefix test-type ?

Rakshith-R (commented):
This will be an awesome change to have for debugging.
But to merge a pr I would expect we would need to run the full suite, since we may
have cross dependencies between drivers.

jobs/mini-e2e.yaml
karthik-us: Done.

jobs/mini-e2e.yaml
karthik-us: Yes, the intention was to keep it optional only. Updated now with the `/test_type-{test_type}` format.

karthik-us:
> I assume this will be a breaking change, This might need some more enhancement to detect which tests need to be run and also might require changes in the pull-request commenter as well.

Since it is made optional, it should not break the existing flow. We will call it on a need basis for debugging purpose as Rakshith mentioned. The pull-request-commentor may not need any change. When it gets the `ok-to-test` label, the whole test suite runs as usual to make sure no cross dependency is broken.

Rakshith-R (approved):
LGTM, thanks !

Madhu-1 (approved):

PR #4125, opened 2023-09-13
cephfs: add support for VolumeCondition in NodeGetVolumeStats operation

nixpanic:
This is my work-in-progress state for supporting `VOLUME_CONDITION`, starting with CephFS.

Updates: #1356

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

- `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

PR #4200, opened 2023-10-17
CephFS: prevent hanging NodeGetVolumeStats on `stat()` syscall when an MDS is slow

nixpanic:
When an MDS is slow, the `NodeGetVolumeStats` CSI procedure may hang for a while. In case the MDS does not respond at all (or the reply is lost), the `stat()` syscall may hang indefinitely.

Because Kubelet retries the `NodeGetVolumeStats` CSI procedure regularly, the number of hanging `stat()` syscalls can pike up. Each hanging call will consume a thread (go routine), and this may eventually starve the processing capabilities of the CephFS CSI-nodeplugin container.

The commits in this PR come from #4125, which is planned to have a little more features (like a separate `data/` directory for the contents of the volume).

## Manual Testing

1. create a PVC and start a Pod that uses it
2. let it run for a few minutes, confirm that `NodeGetVolumeStats` calls are in the CephFS node-plugin logs
3. scale down the MDS(s), to simulate a failure
4. check the CephFS node-plugin logs and see if there are unhealthy messages for `NodeGetVolumeStats` calls

Results (see `GRPC response`):
```
I1024 12:31:13.033160       1 utils.go:195] ID: 11 GRPC call: /csi.v1.Node/NodeGetCapabilities
I1024 12:31:13.033195       1 utils.go:206] ID: 11 GRPC request: {}
I1024 12:31:13.033274       1 utils.go:212] ID: 11 GRPC response: {"capabilities":[{"Type":{"Rpc":{"type":1}}},{"Type":{"Rpc":{"type":2}}},{"Type":{"Rpc":{"type":4}}},{"Type":{"Rpc":{"type":5}}}]}
I1024 12:31:13.034119       1 utils.go:195] ID: 12 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I1024 12:31:13.034261       1 utils.go:206] ID: 12 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000001-af951451-f063-4a67-bf7c-8e788bdc587c","volume_path":"/var/lib/kubelet/pods/b2a6ae50-16a9-4326-83d5-58aafcb70f90/volumes/kubernetes.io~csi/pvc-66c3bfe6-1046-438b-8070-60b070b71149/mount"}
I1024 12:31:13.035323       1 utils.go:212] ID: 12 GRPC response: {"usage":[{"available":1073741824,"total":1073741824,"unit":1}],"volume_condition":{"message":"volume is in a healthy condition"}}

I1024 12:32:33.878002       1 utils.go:195] ID: 13 GRPC call: /csi.v1.Node/NodeGetCapabilities
I1024 12:32:33.878281       1 utils.go:206] ID: 13 GRPC request: {}
I1024 12:32:33.878379       1 utils.go:212] ID: 13 GRPC response: {"capabilities":[{"Type":{"Rpc":{"type":1}}},{"Type":{"Rpc":{"type":2}}},{"Type":{"Rpc":{"type":4}}},{"Type":{"Rpc":{"type":5}}}]}
I1024 12:32:33.879443       1 utils.go:195] ID: 14 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I1024 12:32:33.879502       1 utils.go:206] ID: 14 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000001-af951451-f063-4a67-bf7c-8e788bdc587c","volume_path":"/var/lib/kubelet/pods/b2a6ae50-16a9-4326-83d5-58aafcb70f90/volumes/kubernetes.io~csi/pvc-66c3bfe6-1046-438b-8070-60b070b71149/mount"}
I1024 12:32:33.879549       1 utils.go:212] ID: 14 GRPC response: {"volume_condition":{"abnormal":true,"message":"health-check has not responded for 119.215949 seconds"}}
I1024 12:34:17.335423       1 utils.go:195] ID: 15 GRPC call: /csi.v1.Node/NodeGetCapabilities
I1024 12:34:17.335461       1 utils.go:206] ID: 15 GRPC request: {}
I1024 12:34:17.335542       1 utils.go:212] ID: 15 GRPC response: {"capabilities":[{"Type":{"Rpc":{"type":1}}},{"Type":{"Rpc":{"type":2}}},{"Type":{"Rpc":{"type":4}}},{"Type":{"Rpc":{"type":5}}}]}
I1024 12:34:17.336271       1 utils.go:195] ID: 16 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I1024 12:34:17.336350       1 utils.go:206] ID: 16 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000001-af951451-f063-4a67-bf7c-8e788bdc587c","volume_path":"/var/lib/kubelet/pods/b2a6ae50-16a9-4326-83d5-58aafcb70f90/volumes/kubernetes.io~csi/pvc-66c3bfe6-1046-438b-8070-60b070b71149/mount"}
I1024 12:34:17.336442       1 utils.go:212] ID: 16 GRPC response: {"volume_condition":{"abnormal":true,"message":"health-check has not responded for 222.672824 seconds"}}
I1024 12:35:55.149537       1 utils.go:195] ID: 17 GRPC call: /csi.v1.Node/NodeGetCapabilities
I1024 12:35:55.149583       1 utils.go:206] ID: 17 GRPC request: {}
I1024 12:35:55.149667       1 utils.go:212] ID: 17 GRPC response: {"capabilities":[{"Type":{"Rpc":{"type":1}}},{"Type":{"Rpc":{"type":2}}},{"Type":{"Rpc":{"type":4}}},{"Type":{"Rpc":{"type":5}}}]}
I1024 12:35:55.150294       1 utils.go:195] ID: 18 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I1024 12:35:55.150331       1 utils.go:206] ID: 18 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000001-af951451-f063-4a67-bf7c-8e788bdc587c","volume_path":"/var/lib/kubelet/pods/b2a6ae50-16a9-4326-83d5-58aafcb70f90/volumes/kubernetes.io~csi/pvc-66c3bfe6-1046-438b-8070-60b070b71149/mount"}
I1024 12:35:55.150431       1 utils.go:212] ID: 18 GRPC response: {"volume_condition":{"abnormal":true,"message":"health-check has not responded for 320.486780 seconds"}}
```

Scale up the MDS(s) again, see that the volume becomes healthy again:
```
I1024 12:39:03.241203       1 utils.go:195] ID: 21 GRPC call: /csi.v1.Node/NodeGetCapabilities
I1024 12:39:03.241236       1 utils.go:206] ID: 21 GRPC request: {}
I1024 12:39:03.241308       1 utils.go:212] ID: 21 GRPC response: {"capabilities":[{"Type":{"Rpc":{"type":1}}},{"Type":{"Rpc":{"type":2}}},{"Type":{"Rpc":{"type":4}}},{"Type":{"Rpc":{"type":5}}}]}
I1024 12:39:03.242094       1 utils.go:195] ID: 22 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I1024 12:39:03.242132       1 utils.go:206] ID: 22 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000001-af951451-f063-4a67-bf7c-8e788bdc587c","volume_path":"/var/lib/kubelet/pods/b2a6ae50-16a9-4326-83d5-58aafcb70f90/volumes/kubernetes.io~csi/pvc-66c3bfe6-1046-438b-8070-60b070b71149/mount"}
I1024 12:39:03.242220       1 utils.go:212] ID: 22 GRPC response: {"volume_condition":{"abnormal":true,"message":"health-check has not responded for 448.545782 seconds"}}
I1024 12:40:40.223909       1 utils.go:195] ID: 23 GRPC call: /csi.v1.Node/NodeGetCapabilities
I1024 12:40:40.223962       1 utils.go:206] ID: 23 GRPC request: {}
I1024 12:40:40.224044       1 utils.go:212] ID: 23 GRPC response: {"capabilities":[{"Type":{"Rpc":{"type":1}}},{"Type":{"Rpc":{"type":2}}},{"Type":{"Rpc":{"type":4}}},{"Type":{"Rpc":{"type":5}}}]}
I1024 12:40:40.224870       1 utils.go:195] ID: 24 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I1024 12:40:40.224989       1 utils.go:206] ID: 24 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000001-af951451-f063-4a67-bf7c-8e788bdc587c","volume_path":"/var/lib/kubelet/pods/b2a6ae50-16a9-4326-83d5-58aafcb70f90/volumes/kubernetes.io~csi/pvc-66c3bfe6-1046-438b-8070-60b070b71149/mount"}
I1024 12:40:40.225807       1 utils.go:212] ID: 24 GRPC response: {"usage":[{"available":1073741824,"total":1073741824,"unit":1}],"volume_condition":{"message":"volume is in a healthy condition"}}
```

---

<details>

</details>

### DIFF
--- a/docs/design/proposals/volume-condition.md
+++ b/docs/design/proposals/volume-condition.md
@@ -1,0 +1,25 @@
+# Support for CSI `VolumeCondition` aka Volume Health Checker
+
+## health-checker API
+
+Under `internal/health-checker`  the Manager for health-checking is
+implemented. The Manager can start a checking process for a given path, return
+the (un)healthy state and stop the checking process when the volume is not
+needed anymore.
+
+The Manager is responsible for creating a suitable checker for the requested
+path. If the path is a block-device, the BlockChecker should be created. For a
+filesystem path (directory), the FileChecker is appropriate.
+
+## CephFS
+
+The health-checker writes to the file `csi-volume-condition.ts` in the root of
+the volume. This file contains a JSON formatted timestamp.
+
+A new `data` directory is introduced for newly created volumes. During the
+`NodeStageVolume` call the root of the volume is mounted, and the `data`
+directory is bind-mounted inside the container when `NodePublishVolume` is
+called.
+
+The `data` directory makes it possible to place Ceph-CSI internal files in the
+root of the volume, without that the user/application has access to it.
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -29,6 +29,7 @@ import (
 	"github.com/ceph/ceph-csi/internal/cephfs/store"
 	fsutil "github.com/ceph/ceph-csi/internal/cephfs/util"
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
+	hc "github.com/ceph/ceph-csi/internal/health-checker"
 	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/fscrypt"
 	"github.com/ceph/ceph-csi/internal/util/log"
@@ -47,6 +48,7 @@ type NodeServer struct {
 	VolumeLocks        *util.VolumeLocks
 	kernelMountOptions string
 	fuseMountOptions   string
+	healthChecker      hc.Manager
 }
 
 func getCredentialsForVolume(
@@ -228,6 +230,8 @@ func (ns *NodeServer) NodeStageVolume(
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 
+		ns.startSharedHealthChecker(ctx, req.GetVolumeId(), stagingTargetPath)
+
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
@@ -270,7 +274,22 @@ func (ns *NodeServer) NodeStageVolume(
 		}
 	}
 
+	ns.startSharedHealthChecker(ctx, req.GetVolumeId(), stagingTargetPath)
+
 	return &csi.NodeStageVolumeResponse{}, nil
+}
+
+// startSharedHealthChecker starts a health-checker on the stagingTargetPath.
+// This checker can be shared between multiple containers.
+//
+// TODO: start a FileChecker for read-writable volumes that have an app-data subdir.
+func (ns *NodeServer) startSharedHealthChecker(ctx context.Context, volumeID, dir string) {
+	// The StatChecker works for volumes that do not have a dedicated app-data
+	// subdirectory, or are read-only.
+	err := ns.healthChecker.StartSharedChecker(volumeID, dir, hc.StatCheckerType)
+	if err != nil {
+		log.WarningLog(ctx, "failed to start healthchecker: %v", err)
+	}
 }
 
 func (ns *NodeServer) mount(
@@ -479,7 +498,8 @@ func (ns *NodeServer) NodePublishVolume(
 
 	// Ensure staging target path is a mountpoint.
 
-	if isMnt, err := util.IsMountPoint(ns.Mounter, stagingTargetPath); err != nil {
+	isMnt, err := util.IsMountPoint(ns.Mounter, stagingTargetPath)
+	if err != nil {
 		log.ErrorLog(ctx, "stat failed: %v", err)
 
 		return nil, status.Error(codes.Internal, err.Error())
@@ -491,7 +511,7 @@ func (ns *NodeServer) NodePublishVolume(
 
 	// Check if the volume is already mounted
 
-	isMnt, err := util.IsMountPoint(ns.Mounter, targetPath)
+	isMnt, err = util.IsMountPoint(ns.Mounter, targetPath)
 	if err != nil {
 		log.ErrorLog(ctx, "stat failed: %v", err)
 
@@ -545,6 +565,10 @@ func (ns *NodeServer) NodeUnpublishVolume(
 	// considering kubelet make sure node operations like unpublish/unstage...etc can not be called
 	// at same time, an explicit locking at time of nodeunpublish is not required.
 	targetPath := req.GetTargetPath()
+
+	// stop the health-checker that may have been started in NodeGetVolumeStats()
+	ns.healthChecker.StopChecker(req.GetVolumeId(), targetPath)
+
 	isMnt, err := util.IsMountPoint(ns.Mounter, targetPath)
 	if err != nil {
 		log.ErrorLog(ctx, "stat failed: %v", err)
@@ -599,6 +623,9 @@ func (ns *NodeServer) NodeUnstageVolume(
 	}
 
 	volID := req.GetVolumeId()
+
+	ns.healthChecker.StopSharedChecker(volID)
+
 	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
 		log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volID)
 
@@ -673,6 +700,13 @@ func (ns *NodeServer) NodeGetCapabilities(
 			{
 				Type: &csi.NodeServiceCapability_Rpc{
 					Rpc: &csi.NodeServiceCapability_RPC{
+						Type: csi.NodeServiceCapability_RPC_VOLUME_CONDITION,
+					},
+				},
+			},
+			{
+				Type: &csi.NodeServiceCapability_Rpc{
+					Rpc: &csi.NodeServiceCapability_RPC{
 						Type: csi.NodeServiceCapability_RPC_SINGLE_NODE_MULTI_WRITER,
 					},
 				},
@@ -694,6 +728,35 @@ func (ns *NodeServer) NodeGetVolumeStats(
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	// health check first, return without stats if unhealthy
+	healthy, msg := ns.healthChecker.IsHealthy(req.GetVolumeId(), targetPath)
+
+	// If healthy and an error is returned, it means that the checker was not
+	// started. This could happen when the node-plugin was restarted and the
+	// volume is already staged and published.
+	if healthy && msg != nil {
+		// Start a StatChecker for the mounted targetPath, this prevents
+		// writing a file in the user-visible location. Ideally a (shared)
+		// FileChecker is started with the stagingTargetPath, but we can't
+		// get the stagingPath from the request easily.
+		// TODO: resolve the stagingPath like rbd.getStagingPath() does
+		err = ns.healthChecker.StartChecker(req.GetVolumeId(), targetPath, hc.StatCheckerType)
+		if err != nil {
+			log.WarningLog(ctx, "failed to start healthchecker: %v", err)
+		}
+	}
+
+	// !healthy indicates a problem with the volume
+	if !healthy {
+		return &csi.NodeGetVolumeStatsResponse{
+			VolumeCondition: &csi.VolumeCondition{
+				Abnormal: true,
+				Message:  msg.Error(),
+			},
+		}, nil
+	}
+
+	// warning: stat() may hang on an unhealthy volume
 	stat, err := os.Stat(targetPath)
 	if err != nil {
 		if util.IsCorruptedMountError(err) {
--- a/internal/health-checker/filechecker.go
+++ b/internal/health-checker/filechecker.go
@@ -1,0 +1,118 @@
+/*
+Copyright 2023 ceph-csi authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthchecker
+
+import (
+	"errors"
+	"os"
+	"path"
+	"time"
+)
+
+type fileChecker struct {
+	checker
+
+	// filename contains the filename that is used for checking.
+	filename string
+}
+
+func newFileChecker(dir string) ConditionChecker {
+	fc := &fileChecker{
+		filename: path.Join(dir, "csi-volume-condition.ts"),
+	}
+	fc.initDefaults()
+
+	fc.checker.runChecker = func() {
+		fc.isRunning = true
+
+		ticker := time.NewTicker(fc.interval)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-fc.commands: // STOP command received
+				fc.isRunning = false
+
+				return
+			case now := <-ticker.C:
+				err := fc.writeTimestamp(now)
+				if err != nil {
+					fc.mutex.Lock()
+					fc.healthy = false
+					fc.err = err
+					fc.mutex.Unlock()
+
+					continue
+				}
+
+				ts, err := fc.readTimestamp()
+				if err != nil {
+					fc.mutex.Lock()
+					fc.healthy = false
+					fc.err = err
+					fc.mutex.Unlock()
+
+					continue
+				}
+
+				// verify that the written timestamp is read back
+				if now.Compare(ts) != 0 {
+					fc.mutex.Lock()
+					fc.healthy = false
+					fc.err = errors.New("timestamp read from file does not match what was written")
+					fc.mutex.Unlock()
+
+					continue
+				}
+
+				// run health check, write a timestamp to a file, read it back
+				fc.mutex.Lock()
+				fc.healthy = true
+				fc.err = nil
+				fc.lastUpdate = ts
+				fc.mutex.Unlock()
+			}
+		}
+	}
+
+	return fc
+}
+
+// readTimestamp reads the JSON formatted timestamp from the file.
+func (fc *fileChecker) readTimestamp() (time.Time, error) {
+	var ts time.Time
+
+	data, err := os.ReadFile(fc.filename)
+	if err != nil {
+		return ts, err
+	}
+
+	err = ts.UnmarshalJSON(data)
+
+	return ts, err
+}
+
+// writeTimestamp writes the timestamp to the file in JSON format.
+func (fc *fileChecker) writeTimestamp(ts time.Time) error {
+	data, err := ts.MarshalJSON()
+	if err != nil {
+		return err
+	}
+
+	//nolint:gosec // allow reading of the timestamp for debugging
+	return os.WriteFile(fc.filename, data, 0o644)
+}
--- a/internal/health-checker/manager.go
+++ b/internal/health-checker/manager.go
@@ -1,0 +1,208 @@
+/*
+Copyright 2023 ceph-csi authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthchecker
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+)
+
+// CheckerType describes the type of health-check that needs to be done.
+type CheckerType uint64
+
+const (
+	// StatCheckerType uses the stat() syscall to validate volume health.
+	StatCheckerType = iota
+	// FileCheckerType writes and reads a timestamp to a file for checking the
+	// volume health.
+	FileCheckerType
+)
+
+// Manager provides the API for getting the health status of a volume. The main
+// usage is requesting the health status by volumeID.
+//
+// When the Manager detects that a new volumeID is used for checking, a new
+// instance of a ConditionChecker is created for the volumeID on the given
+// path, and started.
+//
+// Once the volumeID is not active anymore (when NodeUnstageVolume is called),
+// the ConditionChecker needs to be stopped, which can be done by
+// Manager.StopChecker().
+type Manager interface {
+	// StartChecker starts a health-checker of the requested type for the
+	// volumeID using the path. The path usually is the publishTargetPath, and
+	// a unique path for this checker. If the path can be used by multiple
+	// containers, use the StartSharedChecker function instead.
+	StartChecker(volumeID, path string, ct CheckerType) error
+
+	// StartSharedChecker starts a health-checker of the requested type for the
+	// volumeID using the path. The path usually is the stagingTargetPath, and
+	// can be used for multiple containers.
+	StartSharedChecker(volumeID, path string, ct CheckerType) error
+
+	StopChecker(volumeID, path string)
+	StopSharedChecker(volumeID string)
+
+	// IsHealthy locates the checker for the volumeID and path. If no checker
+	// is found, `true` is returned together with an error message.
+	// When IsHealthy runs into an internal error, it is assumed that the
+	// volume is healthy. Only when it is confirmed that the volume is
+	// unhealthy, `false` is returned together with an error message.
+	IsHealthy(volumeID, path string) (bool, error)
+}
+
+// ConditionChecker describes the interface that a health status reporter needs
+// to implement. It is used internally by the Manager only.
+type ConditionChecker interface {
+	// start runs a the health checking function in a new go routine.
+	start()
+
+	// stop terminates a the health checking function that runs in a go
+	// routine.
+	stop()
+
+	// isHealthy returns the status of the volume, without blocking.
+	isHealthy() (bool, error)
+}
+
+type healthCheckManager struct {
+	checkers sync.Map // map[volumeID]ConditionChecker
+}
+
+func NewHealthCheckManager() Manager {
+	return &healthCheckManager{
+		checkers: sync.Map{},
+	}
+}
+
+func (hcm *healthCheckManager) StartSharedChecker(volumeID, path string, ct CheckerType) error {
+	return hcm.createChecker(volumeID, path, ct, true)
+}
+
+func (hcm *healthCheckManager) StartChecker(volumeID, path string, ct CheckerType) error {
+	return hcm.createChecker(volumeID, path, ct, false)
+}
+
+// createChecker decides based on the CheckerType what checker to start for
+// the volume.
+func (hcm *healthCheckManager) createChecker(volumeID, path string, ct CheckerType, shared bool) error {
+	switch ct {
+	case FileCheckerType:
+		return hcm.startFileChecker(volumeID, path, shared)
+	case StatCheckerType:
+		return hcm.startStatChecker(volumeID, path, shared)
+	}
+
+	return nil
+}
+
+// startFileChecker initializes the fileChecker and starts it.
+func (hcm *healthCheckManager) startFileChecker(volumeID, path string, shared bool) error {
+	workdir := filepath.Join(path, ".csi")
+	err := os.Mkdir(workdir, 0o755)
+	if err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed to created workdir %q for health-checker: %w", workdir, err)
+	}
+
+	cc := newFileChecker(workdir)
+
+	return hcm.startChecker(cc, volumeID, path, shared)
+}
+
+// startStatChecker initializes the statChecker and starts it.
+func (hcm *healthCheckManager) startStatChecker(volumeID, path string, shared bool) error {
+	cc := newStatChecker(path)
+
+	return hcm.startChecker(cc, volumeID, path, shared)
+}
+
+// startChecker adds the checker to its map and starts it.
+// Shared checkers are key'd by their volumeID, whereas non-shared checkers
+// are key'd by theit volumeID+path.
+func (hcm *healthCheckManager) startChecker(cc ConditionChecker, volumeID, path string, shared bool) error {
+	key := volumeID
+	if shared {
+		key = fallbackKey(volumeID, path)
+	}
+
+	// load the 'old' ConditionChecker if it exists, otherwise store 'cc'
+	old, ok := hcm.checkers.LoadOrStore(key, cc)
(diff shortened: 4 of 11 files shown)


internal/cephfs/nodeserver.go
```diff
@@ -711,8 +744,38 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+
+	err = ns.healthChecker.StartChecker(hcDir)
+	if err != nil {
+		log.WarningLog(ctx, "failed to start health-checker for path %q: %v", hcDir, err)
+	}
+}
+
+func getHealthCheckPath(stagingTargetPath string) (string, error) {
+	dir := path.Join(stagingTargetPath, ".csi")
+	err := os.Mkdir(dir, 0o700)
```
nixpanic: `os.Mkdir()` or any other check that the directory on the CephFS mountpoint exists is problematic. The NodeGetVolumeStats call hangs when there is no available MDS (to cause intentional failures).

The directory for the timestamp should be created/checked once, when the health-checker is started (just after mounting, likely to succeed).

internal/cephfs/nodeserver.go
```diff
@@ -270,6 +272,9 @@ func (ns *NodeServer) NodeStageVolume(
 		}
 	}
 
+	err = ns.healthChecker.StartChecker(req.GetVolumeId(), stagingTargetPath)
+	log.WarningLog(ctx, "failed to start healtchecker: %v", err)
+
```
Madhu-1: What if the volume is already mounted (case where csi plugin just restarted)

internal/cephfs/nodeserver.go
```diff
@@ -599,6 +606,9 @@ func (ns *NodeServer) NodeUnstageVolume(
 	}
 
 	volID := req.GetVolumeId()
+
+	ns.healthChecker.StopChecker(volID)
```
Madhu-1: This should happen right before umount but after all other operations right?

internal/cephfs/nodeserver.go
```diff
@@ -694,6 +711,18 @@ func (ns *NodeServer) NodeGetVolumeStats(
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	// health check first, return without stats if unhealthy
+	healthy, msg := ns.healthChecker.IsHealthy(req.GetVolumeId())
```
Madhu-1: what happens if the NodeGetVolumeStats is called right after the NodeStageVolume?

internal/cephfs/nodeserver.go
```diff
@@ -711,7 +740,12 @@ func (ns *NodeServer) NodeGetVolumeStats(
 	}
 
 	if stat.Mode().IsDir() {
-		return csicommon.FilesystemNodeGetVolumeStats(ctx, ns.Mounter, targetPath, false)
+		res, err := csicommon.FilesystemNodeGetVolumeStats(ctx, ns.Mounter, targetPath, false)
+		if err != nil {
+			return nil, err
+		}
+
+		return res, nil
```
Madhu-1: we dont need this check and can have `return csicommon.FilesystemNodeGetVolumeStats(ctx, ns.Mounter, targetPath, false)` as it was before

internal/health-checker/filechecker.go
```diff
@@ -0,0 +1,186 @@
...
+	"time"
+)
+
+// command is what is sent through the channel to terminate the go routine.
+type command string
+
+const (
+	// stopCommand is sent through the channel to stop checking.
+	stopCommand = command("STOP")
+)
```
Madhu-1: can make use of empty struct to signal between go routines?

internal/health-checker/filechecker.go
```diff
@@ -0,0 +1,186 @@
...
+			// a command was reveived, need to stop checking
+			return true
+		default:
+			// continue with checking
+		}
+
+		time.Sleep(time.Second)
+	}
+
+	return false
```
Madhu-1: use time.Ticker for this one?

internal/health-checker/filechecker.go
```diff
@@ -0,0 +1,186 @@
...
+
+// writeTimestamp writes the timestamp to the file in JSON format.
+func (fc *fileChecker) writeTimestamp(ts time.Time) error {
+	data, err := ts.MarshalJSON()
+	if err != nil {
+		return err
+	}
+
+	//nolint:gosec // allow reading of the timestamp for debugging
+	return os.WriteFile(fc.filename, data, 0o644)
```
Madhu-1, line 117: will this also get blocked like os.Stat as well or we will receive an error?

internal/cephfs/nodeserver.go
```diff
@@ -694,6 +711,18 @@ func (ns *NodeServer) NodeGetVolumeStats(
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	// health check first, return without stats if unhealthy
+	healthy, msg := ns.healthChecker.IsHealthy(req.GetVolumeId())
+	if !healthy {
+		return &csi.NodeGetVolumeStatsResponse{
```
Madhu-1, line 751: What happens for the already running workload if cephcsi is upgraded for this one version?

docs/design/proposals/volume-condition.md
```diff
@@ -0,0 +1,25 @@
...
+path. If the path a s block-device, the BlockChecker should be created. For a
+filesystem path (directory), the FileChecker is appropriate.
+
+## CephFS
+
+The health-checker writes to the file `csi-volume-condition.ts` in the root of
+the volume. This file containse a JSON formatted timestamp.
+
+A new `data` directory is introduced for newly created volumes. During the
+`NodeStageVolume` call the root of the volume is mounted, and the `data`
```
Madhu-1, line 20: Will there be any problem for the static PVC or the existing PVC which already have data folder created in it? i don't think so but checking

internal/cephfs/nodeserver.go
```diff
@@ -270,6 +272,9 @@ func (ns *NodeServer) NodeStageVolume(
 		}
 	}
 
+	err = ns.healthChecker.StartChecker(req.GetVolumeId(), stagingTargetPath)
```
Madhu-1: I might be missing the code where We create the data directory and mount that to the target path from stagingPath.

internal/health-checker/filechecker.go
```diff
@@ -0,0 +1,186 @@
...
+	// Without such a check, a single slow write/read could trigger actions
+	// to recover an unhealthy volume already.
+	//
+	// It is required to check, in case the write or read in the go routine
+	// is blocked.
+
+	delay := time.Since(fc.lastUpdate)
+	if delay > (fc.interval + fc.timeout) {
+		fc.healthy = false
+		fc.err = fmt.Errorf("health-check has not responded for %f seconds", delay.Seconds())
```
Madhu-1: Will there be any problem with multiple goroutines trying to update the same variables?

internal/health-checker/manager.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (hcm *healthCheckManager) StartChecker(volumeID, path string) error {
+	workdir := filepath.Join(path, ".csi")
+	err := os.Mkdir(workdir, 0o755)
+	if err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed to created workdir %q for health-checker: %w", workdir, err)
+	}
+
+	cc := newFileChecker(workdir)
```
Madhu-1, line 123: This always starts a FileChecker what about BlockChecker for future?

internal/health-checker/manager.go
```diff
@@ -0,0 +1,121 @@
...
+	if ok {
+		// 'old' was loaded, cast it to ConditionChecker
+		_, ok = old.(ConditionChecker)
+		if !ok {
+			return fmt.Errorf("failed to cast cc to ConditionChecker for volume-id %q", volumeID)
+		}
+	} else {
+		// 'cc' was stored, start it only once
+		cc.start()
+	}
```
Madhu-1, line 155: This can happen only for Static PVC? because the NodeStageVolume will be called only once per volumeID

internal/health-checker/manager.go
```diff
@@ -0,0 +1,121 @@
...
+func (hcm *healthCheckManager) StartChecker(volumeID, path string) error {
+	workdir := filepath.Join(path, ".csi")
+	err := os.Mkdir(workdir, 0o755)
+	if err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed to created workdir %q for health-checker: %w", workdir, err)
+	}
+
+	cc := newFileChecker(workdir)
+
+	// load the 'old' ConditionChecker if it exists, otherwuse store 'cc'
```
Madhu-1: `otherwuse` to `otherwise`

internal/health-checker/manager.go
```diff
@@ -0,0 +1,121 @@
...
+	// 'old' was loaded, cast it to ConditionChecker
+	cc, ok := old.(ConditionChecker)
+	if !ok {
+		// failed to cast, should not be possible
+		return
+	}
+	cc.stop()
+}
+
+func (hcm *healthCheckManager) IsHealthy(volumeID string) (bool, error) {
```
Madhu-1: How manager behaves when the same volumeID is used for both cephfs/NFS and rbd?

docs/design/proposals/volume-condition.md
nixpanic: Existing PVC's would not have the option in the VolumeContext, so there will be a fall-back mechanism.

The current implementation uses a hidden `.csi/` directory in the root of the volume, which is basically what is used for volumes that do not have the extra (coming) option in the VolumeContext.

internal/health-checker/filechecker.go
nixpanic: That is possible, but what is the advantage?

I've chosen this way to make it extendable if needed, it does not matter at the moment what is sent over the channel.

internal/health-checker/filechecker.go
nixpanic: yes, I'll do, that makes it a little cleaner.

internal/health-checker/filechecker.go
nixpanic: Not really. There is a very small race in case one go-routine does not enter this if-statement, and an other does + updates the `fc.err` before the 1st go-routines exits the function. I don't think it is a realistic failure scenario to account for.

A same way can happen with this function and concurrent `runChecker()`. But it also looks quite an edge case.

If you insist, I can add some locking around updating/reading `fc.healthy` and `fc.err`.

internal/health-checker/filechecker.go
nixpanic: Yes. The behavior for all (I guess) syscalls will be the same. Either blocking, or returning an unexpected error (like permission denied after fencing).

internal/health-checker/manager.go
nixpanic: When a `BlockChecker` is introduced, the `Manager` interface will need some modifications in any case.

internal/health-checker/manager.go
nixpanic: `NodeStageVolume` might have failed/timedout after the checker was started. Calling `NodeStageVolume` again should be a no-op in that case, this just ensures some idempotency.

internal/health-checker/manager.go
nixpanic: That should not matter, this Manager runs in a node-plugin, and those are dedicated for cephfs/rbd/nfs.

I am also not sure how you could get the same VolumeID for different storage backends?

internal/cephfs/nodeserver.go
nixpanic: That is not part of this PR. This only addresses the async health-checking. #4125 was started with the extra `data` subdir in mind, and it will be remodelled to extend this PR.

internal/cephfs/nodeserver.go
nixpanic: In that case, there it no health-checker.

Possibly `NodeGetVolumeStats` can be extended to start the checker to catch this.

internal/cephfs/nodeserver.go
nixpanic: Nothing. The `VolumeCondition` is not used anywhere directly. [Kubernetes can report it](https://kubernetes.io/docs/concepts/storage/volume-health-monitoring/), but does not do so by default (yet).

internal/cephfs/nodeserver.go
nixpanic: ah, yes. It was modified in an earlier version to set `res.VolumeCondition = ...` after getting the stats. ... that wasn't my smartest move, as getting the stats got hung, so the `VolumeCondition` was never set and returned :man_facepalming:

docs/design/proposals/volume-condition.md
```diff
@@ -0,0 +1,25 @@
+# Support for CSI `VolumeCondition` aka Volume Health Checker
+
+## health-checker API
+
+Under `internal/health-checker`  the Manager for health-checking is
+implemented. The Manager can start a checking process for a given path, return
+the (un)healthy state and stop the checking process when the volume is not
+needed anymore.
+
+The Manager is responsible for creating a suitble checker for the requested
```
riya-singhal31: minor nits,  suitable

docs/design/proposals/volume-condition.md
```diff
@@ -0,0 +1,25 @@
...
+path. If the path a s block-device, the BlockChecker should be created. For a
```
riya-singhal31:
```suggestion
path. If the path is block-device, the BlockChecker should be created. For a
```

docs/design/proposals/volume-condition.md
```diff
@@ -0,0 +1,25 @@
...
+filesystem path (directory), the FileChecker is appropriate.
+
+## CephFS
+
+The health-checker writes to the file `csi-volume-condition.ts` in the root of
+the volume. This file containse a JSON formatted timestamp.
```
riya-singhal31:
```suggestion
the volume. This file contains a JSON formatted timestamp.
```

nixpanic:
@Madhu-1 and @riya-singhal31 , I have manually tested this again (by scaling MDS's down/up) and it still works as before. Could you review again and mark the addressed comments as resolved? Many thanks!

internal/cephfs/nodeserver.go
```diff
@@ -270,6 +272,9 @@ func (ns *NodeServer) NodeStageVolume(
...
+	log.WarningLog(ctx, "failed to start healtchecker: %v", err)
```
riya-singhal31:
```suggestion
	log.WarningLog(ctx, "failed to start healthchecker: %v", err)
```

internal/cephfs/nodeserver.go
Madhu-1: what i mean was if the cephcsi is upgraded we might return the volume as unhealthy for the already mounted pvc because health checker will only gets started during NodeStageVolume? same question for cephcsi restart as well. `IsHealthy` returns error if there is no health check found

internal/health-checker/filechecker.go
Madhu-1: In most cases empty structs are preferred to signal in the go routines, that's why the suggestion

internal/health-checker/manager.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func NewHealthCheckManager() Manager {
+	return &healthCheckManager{
+		checkers: sync.Map{},
+	}
+}
+
+func (hcm *healthCheckManager) StartChecker(volumeID, path string) error {
+	workdir := filepath.Join(path, ".csi")
```
Madhu-1, line 117: what if the PVC already has this path or the application doesn't expect any folders inside the mounted directory?

internal/cephfs/nodeserver.go
Madhu-1: The same check needs to be done at line 233 where we are returning success if the directory is already mounted

internal/cephfs/nodeserver.go
Madhu-1: I think we need this , can we have a tracker for this one?

Madhu-1 (commented):
Left some questions, Mostly LGTM

internal/cephfs/nodeserver.go
nixpanic: Done.

internal/cephfs/nodeserver.go
nixpanic: It should be done early, yes. Unmounting may not be possible if the checker writes/reads from the mountpoint at the same time.

internal/cephfs/nodeserver.go
nixpanic: Done.

internal/cephfs/nodeserver.go
nixpanic: Done.

internal/cephfs/nodeserver.go
nixpanic: yes, makes sense. #4219 has been created for this.

internal/cephfs/nodeserver.go
nixpanic: Before running the 1st check, the volume is regarded healthy. It is the assumption that when mounting works, the volume is still healthy shortly after.

internal/cephfs/nodeserver.go
nixpanic: Right, tracked in #4219.

internal/health-checker/filechecker.go
nixpanic: Is that a recommendation that is documented anywhere? I have seen several examples with integers and booleans, or structs that contain one or more members...

internal/health-checker/manager.go
nixpanic: Other filesystems also maintain their own metadata in the root, like `lost+found` on ext4. Applications that (do not) expect certain filesystem internal metadata should be quite rare, as they are not very portable.

Would you prefer to have an option per volume to use a different directory, or disable it?

internal/health-checker/filechecker.go
nixpanic: Done.

internal/health-checker/manager.go
Madhu-1: Yes, I agree on that one, this is something we are introducing, but not all the application owners/developers will be aware of it. IMHO we should not write or touch anything in the path that is provided to the application pods, I would prefer the below

* Mount the subvolume to the stagingPath and create a new directory as in the design doc and mount that directory to the targetPath
* Do health check only inside the stagingPath which is not available to the application pod

internal/cephfs/nodeserver.go
```diff
@@ -270,6 +275,9 @@ func (ns *NodeServer) NodeStageVolume(
 		}
 	}
 
+	err = ns.healthChecker.StartChecker(req.GetVolumeId(), stagingTargetPath)
+	log.WarningLog(ctx, "failed to start healthchecker: %v", err)
```
Madhu-1: this should be done inside `if err !=nil` check

internal/cephfs/nodeserver.go
```diff
@@ -228,6 +230,9 @@ func (ns *NodeServer) NodeStageVolume(
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 
+		err = ns.healthChecker.StartChecker(req.GetVolumeId(), stagingTargetPath)
+		log.WarningLog(ctx, "failed to start healthchecker: %v", err)
```
Madhu-1: this should be done inside `if err !=nil` check

internal/health-checker/manager.go
```diff
@@ -0,0 +1,121 @@
...
+		return
+	}
+	cc.stop()
+}
+
+func (hcm *healthCheckManager) IsHealthy(volumeID string) (bool, error) {
+	// load the 'old' ConditionChecker if it exists
+	old, ok := hcm.checkers.Load(volumeID)
+	if !ok {
+		return true, fmt.Errorf("no ConditionChecker for volume-id: %s", volumeID)
```
Madhu-1: This error will not checked  as we are returning volume as Healthy and also the error message

internal/health-checker/manager.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	// 'old' was loaded, cast it to ConditionChecker
+	cc, ok := old.(ConditionChecker)
+	if !ok {
+		return true, fmt.Errorf("failed to cast cc to ConditionChecker for volume-id %q", volumeID)
```
Madhu-1, line 194: This error will not checked as we are returning volume as Healthy and also the error message

internal/health-checker/manager.go
nixpanic: For legacy volumes, we could also do a minimal `stat()` on the root of the volume, instead of writing/reading a timestamp. It might not be able to detect all failures scenarios, but it would be a non-intrusive check.

Newly created volumes can have their own `data` subdirectory that is mapped for use inside of the container, leaving the root of the volume for contents that is not visible to applications. That can be done in PR #4125 once this one is accepted.

Madhu-1: >For legacy volumes, we could also do a minimal stat() on the root of the volume, instead of writing/reading a timestamp. It might not be able to detect all failures scenarios, but it would be a non-intrusive check.

It's a temporary problem, once the node gets rebooted or the application gets restagted we will move to the new format. We can live with it.

>Newly created volumes can have their own data subdirectory that is mapped for use inside of the container, leaving the root of the volume for contents that is not visible to applications. That can be done in PR https://github.com/ceph/ceph-csi/pull/4125 once this one is accepted.

Sounds good 👍🏻

internal/health-checker/manager.go
nixpanic: If there is an error in the internals of the health-checker, it is assumed that the volume is healthy. Only when there really is a problem with the volume, the status is reported as unhealthy.

Madhu-1: Can we log and also document this one? we might need to have better logging to check whether the health checker is running and what was the latest run details (by looking at the logs), maybe we can do it as an enhancement later.

internal/health-checker/filechecker.go
nixpanic: Locking has been added.

internal/health-checker/manager.go
nixpanic: Added `statChecker` and will use that in the nodeserver for now. `fileChecker` is still included so that the refactoring with a base `checker` makes sense.

internal/health-checker/manager.go
nixpanic: Added a check for that now, will start a checker in that case too.

internal/health-checker/manager.go
nixpanic: I've added more docs about it now.

internal/cephfs/nodeserver.go
nixpanic: Done!

internal/cephfs/nodeserver.go
nixpanic: done!

internal/cephfs/nodeserver.go
nixpanic: this is actually addressed in the latest revision of this PR too

nixpanic:
Lots of manual testing done. including the following combinations:

pod uses PVC and

1. is healthy
2. csi-cephfsplugin pods restarted, volume stays healthy
3. MDS' are stopped, volume becomes unhealthy
4. MDS' restarted, volume becomes healthy again
5. csi-cephfsplugin pods restarted when MDS' are unavailable, volume stays unhealthy
6. csi-cephfsplugin restarted while unhealthy, MDS' restarted, volume becomes healthy again

internal/cephfs/nodeserver.go
```diff
@@ -694,6 +728,34 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
 
+	// health check first, return without stats if unhealthy
+	healthy, msg := ns.healthChecker.IsHealthy(req.GetVolumeId(), targetPath)
+
+	// If healthy and an error is returned, it means that the checker was not
+	// started. This could happen when the node-plugin was restarted and the
+	// volume is already staged and published.
+	if healthy && msg != nil {
+		// Start a StatChecker for the mounted targetPath, yhis prevents
+		// writing a file in the user-visible location. Ideally a (shared)
```
Madhu-1, line 739: `yhis` to `this`

riya-singhal31 (dismissed):
Thanks Niels, LGTM

internal/cephfs/nodeserver.go
```diff
@@ -694,6 +728,34 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+		// FileChecker is started with the stagingTargetPath, but we can't
+		// get the stagingPath from the request.
+		err = ns.healthChecker.StartChecker(req.GetVolumeId(), targetPath, hc.StatCheckerType)
```
Madhu-1, line 743: can we use something like below to get the stagingPath from targetPath 
https://github.com/ceph/ceph-csi/blob/c09700b5688847cfa2efad0ae1dab5cc14e521ed/internal/rbd/nodeserver.go#L1079-L1094

Madhu-1, line 743: i also see a very small race for the below case

* During NodeUnpublish cephcsi stopped the health checker
* Before umount operation we got a `NodeGetVolumeStats` RPC call which inturn started a health checker
* Now we have a stale health check which will be cleanedup only during NodeUnstage (i think)

Is it possible to have such a case?

Madhu-1 (commented):
LGTM, left a question and small nit

internal/cephfs/nodeserver.go
nixpanic: I do not expect the CO to request `NodeUnpublish` and then still use the same `publishTargetPath` in a later `NodeGetVolumeStats`. If a CO does that, then yes, there would be a stale health checker (if it was started in a `NodeGetVolumeStats` call, and not in `NodeStageVolume`).

nixpanic: Something like this could be an enhancement. I am not sure if it is worth the effort and additional complexity though.

When a FileChecker is used, it would make more sense as that could detect the existence of the file holding the timestamp.

nixpanic: Adding a TODO comment with reference to `rbd.getStagingPath()`

nixpanic:
@Madhu-1 I have addressed your latest concerns as well. Please check again, thanks!

Madhu-1 (approved):
LGTM, Thanks @nixpanic

riya-singhal31 (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

PR #4202, opened 2023-10-18
cephfs: remove snapshot protect/unprotect

### LINKED ISSUE #4169: Remove code for Protect/Unprotect snapshot in CephFS
Based on the supported ceph version, check the feasibility of removing the protect/unprotect code https://github.com/ceph/ceph-csi/blob/cba5402ac27eb0e732d09743bf3480d0ad5a9e1b/internal/cephfs/core/snapshot.go#L177-L244

iPraveenParihar:

This commit eliminates the code for protecting and unprotecting
snapshots, as the functionality to protect and unprotect snapshots
is being deprecated.

Is the change backward compatible? Yes

Are there concerns around backward compatibility? No

Fixes: #4169

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -750,7 +750,7 @@ func (cs *ControllerServer) ControllerExpandVolume(
 // CreateSnapshot creates the snapshot in backend and stores metadata
 // in store
 //
-//nolint:gocognit,gocyclo,cyclop // golangci-lint did not catch this earlier, needs to get fixed late
+//nolint:gocyclo,cyclop // golangci-lint did not catch this earlier, needs to get fixed late
 func (cs *ControllerServer) CreateSnapshot(
 	ctx context.Context,
 	req *csi.CreateSnapshotRequest,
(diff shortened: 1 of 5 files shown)


internal/cephfs/controllerserver.go
```diff
@@ -830,54 +830,19 @@ func (cs *ControllerServer) CreateSnapshot(
 	}
 	defer cs.VolumeLocks.Release(sourceVolID)
 	snapName := req.GetName()
-	sid, snapInfo, err := store.CheckSnapExists(ctx, parentVolOptions, cephfsSnap, cs.ClusterName, cs.SetMetadata, cr)
+	sid, _, err := store.CheckSnapExists(ctx, parentVolOptions, cephfsSnap, cs.ClusterName, cs.SetMetadata, cr)
```
nixpanic: This is the only usage of `CheckSnapExists()`. Please add a 2nd commit to cleanup the function, and not return the unused `snapInfo` anymore.

internal/cephfs/controllerserver.go
```diff
@@ -830,54 +830,19 @@ func (cs *ControllerServer) CreateSnapshot(
...
 
-	// check are we able to retrieve the size of parent
-	// ceph fs subvolume info command got added in 14.2.10 and 15.+
-	// as we are not able to retrieve the parent size we are rejecting the
-	// request to create snapshot.
-	// TODO: For this purpose we could make use of cached clusterAdditionalInfo
-	// too.
 	volClient := core.NewSubVolume(parentVolOptions.GetConnection(), &parentVolOptions.SubVolume,
 		parentVolOptions.ClusterID, cs.ClusterName, cs.SetMetadata)
 	info, err := volClient.GetSubVolumeInfo(ctx)
```
nixpanic, line 840: Now missing an error check?

internal/cephfs/core/clone.go
```diff
@@ -73,43 +72,20 @@ func (s *subVolumeClient) CreateCloneFromSubvolume(
...
-			if err != nil {
-				log.ErrorLog(ctx, "failed to delete snapshot %s %v", snapshotID, err)
-			}
-		}
 
+	// if cloneErr is not nil we will unprotect the snapshot and delete the snapshot
+	var cloneErr error
+
+	defer func() {
 		if cloneErr != nil {
```
Madhu-1, line 75: Not related to this PR, if the Expand fails or the DeleteSnapshot fails we also need to remove the clone and the snapshot, which is missing, can you please address that as well?

iPraveenParihar: Will address this in a separate PR.
Created a issue to track - #4218

internal/cephfs/controllerserver.go
```diff
@@ -866,18 +852,8 @@ func (cs *ControllerServer) CreateSnapshot(
 
 	metadata := k8s.GetSnapshotMetadata(req.GetParameters())
 	if sid != nil {
-		// check snapshot is protected
-		protected := true
 		snapClient := core.NewSnapshot(parentVolOptions.GetConnection(), sid.FsSnapshotName,
```
Madhu-1: Move this client creation to inside if statement?

nixpanic: that won't work, snapclient is also used online 885 (a little below this if)

nixpanic (dismissed):

internal/cephfs/controllerserver.go
Madhu-1: i dont think its used elsewhere inside the `if sid` the client is only used to set metadata for already found snapshot. please let me know if am missing something

Madhu-1: client creation can happen inside `if len(metadata) != 0 {`

iPraveenParihar: > i dont think its used elsewhere inside the `if sid` the client is only used to set metadata for already found snapshot. please let me know if am missing something

yes, thats correct. @nixpanic mentioned the same

> client creation can happen inside if len(metadata) != 0 {

I'll move under this if

Madhu-1 (commented):
One last comment, missed that in last review 🤦🏻

internal/cephfs/controllerserver.go
```diff
@@ -1102,12 +1071,6 @@ func (cs *ControllerServer) DeleteSnapshot(
 	}
 	snapClient := core.NewSnapshot(volOpt.GetConnection(), sid.FsSnapshotName,
```
Madhu-1: Move this snapClient to inside `if needsDelete {` statement as well.

Madhu-1 (approved):

nixpanic (approved):

nixpanic:
@mergifyio rebase

PR #4203, opened 2023-10-19
rebase: update csi spec to latest release

Madhu-1:
updating csi spec to v1.9.0 release.


internal/csi-common/controllerserver-default.go
```diff
@@ -94,3 +94,11 @@ func (cs *DefaultControllerServer) ControllerGetVolume(
 ) (*csi.ControllerGetVolumeResponse, error) {
 	return nil, status.Error(codes.Unimplemented, "")
 }
+
+// ControllerModifyVolume modifies existing volume.
+func (cs *DefaultControllerServer) ControllerModifyVolume(
```
nixpanic: this is so ugly... Do you know why `DefaultControllerServer` does not extend/contain the `csi.UnimplementedControllerServer` so that all these functions can be removed?

Madhu-1: Good catch! this was missing in two places, add it and also removed unwanted methods as well.

nixpanic (approved):

riya-singhal31 (approved):

riya-singhal31:
@Mergifyio queue

nixpanic:
Hmm, all **k8s-1.26** jobs failed. That can hardly be coincidence.

But, I guess the issue is fixed. Deploying failed because downloading [kubectl.sha256](https://dl.k8s.io/release/v1.26.10/bin/linux/amd64/kubectl?checksum=file:https://dl.k8s.io/release/v1.26.10/bin/linux/amd64/kubectl.sha256) fails in all cases. It works for me.

nixpanic:
All **k8s-1.26** CI jobs failed again, with downloading `kubectl` and `kubectl.sha256`. I can download both URLs on my laptop, over IPv6 and IPv4. There might be a broken proxy for dl.k8s.io that serves the contents to the systems in the CentOS CI. We'll have to retry the jobs later again.

PR #4231, opened 2023-11-06
cephfs: addressed few todo

riya-singhal31:
[1]This commit updates the go-ceph version to use the latest branch
[2] This commit addressed few todo, replacing string comparison with ErrNotExist code of cephFS and rbd


internal/cephfs/core/metadata.go
```diff
@@ -152,16 +152,18 @@ func (s *subVolumeClient) UnsetAllMetadata(keys []string) error {
...
 		err := s.removeMetadata(key)
 		// If setMetadata is not supported return nil
 		if errors.Is(err, ErrSubVolMetadataNotSupported) {
 			return nil
 		}
-		// TODO: replace string comparison with errno.
-		if err != nil && !strings.Contains(err.Error(), "No such file or directory") {
-			return fmt.Errorf("failed to unset metadata key %q on subvolume %v: %w", key, s, err)
+		if err != nil {
+			if errors.As(err, &errno) && errno == syscall.ENOENT {
```
Madhu-1: @riya-singhal31 please add an E2E to ensure this working as expected or please test this manually by removing the metadata in the backend.

internal/rbd/rbd_util.go
```diff
@@ -2141,18 +2142,23 @@ func (rv *rbdVolume) setAllMetadata(parameters map[string]string) error {
 
 // unsetAllMetadata unset all the metadata from arg keys on RBD image.
 func (rv *rbdVolume) unsetAllMetadata(keys []string) error {
+	var errno syscall.Errno
 	for _, key := range keys {
 		err := rv.RemoveMetadata(key)
-		// TODO: replace string comparison with errno.
-		if err != nil && !strings.Contains(err.Error(), "No such file or directory") {
-			return fmt.Errorf("failed to unset metadata key %q on %q: %w", key, rv, err)
+		if err != nil {
```
Madhu-1: please add an E2E to ensure this working as expected or please test this manually by removing the metadata in the backend.

internal/cephfs/core/metadata.go
riya-singhal31: @Madhu-1 I'm replacing this errno with the actual errors that are returned by go-ceph, like for ENOENT they have "ErrNotExist "https://github.com/ceph/go-ceph/blob/master/rbd/errors.go#L77 .
But for cephFS, they have the error as private https://github.com/ceph/go-ceph/blob/master/cephfs/errors.go#L64, so sending a patch there to make it a public and then will modify this PR to direct compare the errors using errors.Is(), that might look more better.

riya-singhal31:
Updated the PR, just a minor replacement of string check with error code. 
cc: @Madhu-1 , @nixpanic

PR #4243, opened 2023-11-07
build: make sure `nfs-utils` is installed

nixpanic:
Ceph is minimizing their container-images, which can cause the
`nfs-utils` package to be dropped. As Ceph-CSI supports mounting NFS, it
needs the `/sbin/mount.nfs` executable, so install the package (or a
no-op if it is installed already).

See-also: https://rook-io.slack.com/archives/C46Q5UC05/p1699188662893109


riya-singhal31 (approved):

Rakshith-R (approved):

Rakshith-R:
I think its better we run the entire ci just to be sure nothing breaks with this change.

riya-singhal31:
@Mergifyio queue

PR #4247, opened 2023-11-08
fix : discard flattening rbdFlattenNoParent message

huangzynn:
fix : discard flattening rbdFlattenNoParent message is error if the image does not have any parent

image


Rakshith-R (changes requested):
The current error message is accurate.
Can you explain why you don't think so ?

huangzynn:
> The current error message is accurate. Can you explain why you don't think so ?

https://github.com/ceph/ceph-csi/blob/ffaab69631f064a0b8b19a46c3eedfde5a619219/internal/rbd/rbd_util.go#L836-L838

Because this judgment comes from the error message returned by Ceph

PR #4263, opened 2023-11-16
util: include request-IDs in all gRPC calls for the Controller

nixpanic:
Snapshot procedures do not seem to contain the `Req-ID:` prefix in the
logs anymore (or weren't they there at all?) for some reason. This adds
them back.

### DIFF
--- a/internal/csi-common/server.go
+++ b/internal/csi-common/server.go
@@ -97,7 +97,7 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, srv Servers) {
 		klog.Fatalf("Failed to listen: %v", err)
 	}
 
-	server := grpc.NewServer()
+	server := grpc.NewServer(NewMiddlewareServerOption(false))
 	s.server = server
 
 	if srv.IS != nil {


internal/csi-common/server.go
```diff
@@ -97,7 +97,7 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, srv Servers) {
 		klog.Fatalf("Failed to listen: %v", err)
 	}
 
-	server := grpc.NewServer()
+	server := grpc.NewServer(NewMiddlewareServerOption(false))
```
Madhu-1, line 100: we can remove `metrics` argument from `NewMiddlewareServerOption` as it always set to false but it can be in separate PR.

Madhu-1 (approved):

iPraveenParihar (approved):

Madhu-1:
@mergifyio queue

internal/csi-common/server.go
iPraveenParihar: Removed in #4264

PR #4273, opened 2023-11-21
rbd: use ImageID where possible, add ParentID to prevent confusion

nixpanic:
`librbd.OpenImageById()` works if the image is in the trash, so it makes
it possible to get the parent of the image.

When a new volume is not created yet, the ImageID should not be set to
the ID of the snapshot.

In some places the ImageID is used as the ID of the parent. That is very
confusing and prone to errors. Instead, fetch the right ImageID where
possible, and set ParentID for referencing to parent images.

---

<details>

</details>

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -1017,9 +1017,10 @@ func cleanupRBDImage(ctx context.Context,
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


yati1998:
@nixpanic does this PR still needs to be included?

PR #4291, opened 2023-11-28
support nbd on euler or arm

muxuelanKK:

For euler or arm，command  “rbd device list --format=json --device-type=nbd” can be response slowly, often more than 5 miniutes. So  we use rbd-nbd list-mapped instead.

Do you have any questions?

### DIFF
--- a/internal/rbd/rbd_attach.go
+++ b/internal/rbd/rbd_attach.go
@@ -143,10 +143,22 @@ type detachRBDImageArgs struct {
 // getDeviceList queries rbd about mapped devices and returns a list of deviceInfo
 // It will selectively list devices mapped using krbd or nbd as specified by accessType.
 func getDeviceList(ctx context.Context, accessType string) ([]deviceInfo, error) {
-	// rbd device list --format json --device-type [krbd|nbd]
-	stdout, _, err := util.ExecCommand(ctx, rbd, "device", "list", "--format="+"json", "--device-type", accessType)
-	if err != nil {
-		return nil, fmt.Errorf("error getting device list from rbd for devices of type (%s): %w", accessType, err)
+	var (
+		stdout string
+		err    error
+	)
+
+	if accessType == accessTypeKRbd {
+		// rbd device list --format json --device-type [krbd|nbd]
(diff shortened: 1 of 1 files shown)


nixpanic:
I think this looks reasonable. Could you update the title of the commit to

    rbd: support nbd on euler or arm

so that **commitlint** will accept it? You'll need to force-push that change to your branch that you used to create this PR from.

Thanks!

internal/rbd/rbd_attach.go
```diff
@@ -352,29 +355,27 @@ func attachRBDImage(ctx context.Context, volOptions *rbdVolume, device string, c
 }
 
 func appendNbdDeviceTypeAndOptions(cmdArgs []string, userOptions, cookie string) []string {
-	cmdArgs = append(cmdArgs, "--device-type", accessTypeNbd)
-
 	isUnmap := CheckSliceContains(cmdArgs, "unmap")
 	if !isUnmap {
 		if !strings.Contains(userOptions, useNbdNetlink) {
-			cmdArgs = append(cmdArgs, "--options", useNbdNetlink)
+			cmdArgs = append(cmdArgs, "--"+useNbdNetlink)
```
nixpanic, line 370: My main concern with this patch is that the `--options` parameter seems to be dropped. Is that intentional, and does it keep working with the currently maintained Ceph versions?

nixpanic:
Inspired by this PR, I've created #4364 which will cause a little conflict, sorry!

nixpanic:
> /test ci/centos/mini-e2e/k8s-1.29

This passed, so I think all still works. Ideally it is more efficient now too.

@pkalever can you have a quick look and check if it doesn't break anything?

nixpanic (approved):
CI job passed, so functionality should be OK.

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

PR #4302, opened 2023-12-05
added permission to get nodes for rbd

### LINKED ISSUE #4306: [RBAC] Helm Chart 3.10.0: csi-rbdplugin container enters CrashLoopBackoff with failed to get node error
# Describe the bug #

Hello,
After upgrading the helm chart from 3.9.0 to 3.10.0, the container `csi-rbdplugin` crashes in a loop with a permissions error to fetch the nodes resources.
```
F1206 16:31:23.921143       1 driver.go:131] failed to get node "xxxxxxxx" information: nodes "xxxxxxxx" is forbidden: User "system:serviceaccount:ceph-csi-rbd:cph-cs-rbd-ceph-csi-rbd-provisioner" cannot get resource "nodes" in API group "" at the cluster scope
```
This is probably related to changes from #4165, but the RBAC rules for the service account do not match the new requirement to fetch the nodes labels.

# Environment details #

- Image/version of Ceph CSI driver : 3.10.0
- Helm chart version : 3.10.0
- Kernel version : 5.15.0-89-generic
  `krbd` or `rbd-nbd`) :
- Kubernetes cluster version : 1.27.3
- Ceph cluster version :

# Steps to reproduce #

1. Deploy helm chart ceph-csi/ceph-csi-rbd with version 3.10.0. 
2. Make sure `rbac.create: true` is set in values.yaml

# Actual results #

Pod `csi-rbd-provisioner` is in CrashLoopBackup state due to failure in `csi-rbdplugin` container

# Expected behavior #

The container and pod should not crash

# Logs #

csi-rbdplugin:
```
I1206 16:31:23.907599       1 cephcsi.go:191] Driver version: v3.10.0 and Git version: 24ae2a7a062b3e58746bb9cc6d5737e37a7e771c
I1206 16:31:23.907720       1 cephcsi.go:223] Starting driver type: rbd with name: rbd.csi.ceph.com
I1206 16:31:23.907750       1 driver.go:94] Enabling controller service capability: CREATE_DELETE_VOLUME
I1206 16:31:23.907755       1 driver.go:94] Enabling controller service capability: CREATE_DELETE_SNAPSHOT
I1206 16:31:23.907761       1 driver.go:94] Enabling controller service capability: CLONE_VOLUME
I1206 16:31:23.907765       1 driver.go:94] Enabling controller service capability: EXPAND_VOLUME
I1206 16:31:23.907770       1 driver.go:107] Enabling volume access mode: SINGLE_NODE_WRITER
I1206 16:31:23.907774       1 driver.go:107] Enabling volume access mode: MULTI_NODE_MULTI_WRITER
I1206 16:31:23.907778       1 driver.go:107] Enabling volume access mode: SINGLE_NODE_SINGLE_WRITER
I1206 16:31:23.907918       1 driver.go:107] Enabling volume access mode: SINGLE_NODE_MULTI_WRITER
F1206 16:31:23.921143       1 driver.go:131] failed to get node "xxxxxxx" information: nodes "xxxxxxx" is forbidden: User "system:serviceaccount:ceph-csi-rbd:cph-cs-rbd-ceph-csi-rbd-provisioner" cannot get resource "nodes" in API group "" at the cluster scope
```

# Additional context #

I don't think it's the same issue as #4298. Setting `readAffinity.enabled` to `true` or `false` doesn't change the issue.

nemcikjan:

Based on the implementation in rbd driver the nodeplugin should have by default permission to get nodes in order to retrieve the node labels

* Related rbd driver implementation can be found [here](https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/driver/driver.go#L128C1-L133C3)

## Future concerns ##

**Checklist:**

  guidelines in the [developer

---

<details>

</details>

### DIFF
--- a/internal/rbd/driver/driver.go
+++ b/internal/rbd/driver/driver.go
@@ -125,7 +125,7 @@ func (r *Driver) Run(conf *util.Config) {
 			})
 	}
 
-	if k8s.RunsOnKubernetes() {
+	if k8s.RunsOnKubernetes() && conf.IsNodeServer {
 		nodeLabels, err = k8s.GetNodeLabels(conf.NodeID)
 		if err != nil {
 			log.FatalLogMsg(err.Error())
--- a/charts/ceph-csi-cephfs/templates/nodeplugin-clusterrole.yaml
+++ b/charts/ceph-csi-cephfs/templates/nodeplugin-clusterrole.yaml
@@ -11,9 +11,7 @@ metadata:
     heritage: {{ .Release.Service }}
     {{- with .Values.commonLabels }}{{ toYaml . | trim | nindent 4 }}{{- end }}
 rules:
-{{- if and .Values.readAffinity .Values.readAffinity.enabled }}
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get"]
-{{- end }}
 {{- end -}}
--- a/charts/ceph-csi-rbd/templates/nodeplugin-clusterrole.yaml
+++ b/charts/ceph-csi-rbd/templates/nodeplugin-clusterrole.yaml
@@ -11,11 +11,9 @@ metadata:
     heritage: {{ .Release.Service }}
     {{- with .Values.commonLabels }}{{ toYaml . | trim | nindent 4 }}{{- end }}
 rules:
-{{- if .Values.topology.enabled }}
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get"]
-{{- end }}
   # allow to read Vault Token and connection options from the Tenants namespace
   - apiGroups: [""]
     resources: ["secrets"]


Rakshith-R:
Thanks for the pr ! 

@iPraveenParihar after the new changes to obtain node labels if the pods are running in k8s cluster, we'll need rbac to get nodes regardless of topology or read affinity right ?

Rakshith-R:
@nemcikjan Can you remove the surrounding conditions here too ? 
https://github.com/ceph/ceph-csi/blob/ec05ae6e5f096c5f8be990968f950a6915152b3e/charts/ceph-csi-cephfs/templates/nodeplugin-clusterrole.yaml#L14

nemcikjan:
> @nemcikjan Can you remove the surrounding conditions here too ?
> 
> https://github.com/ceph/ceph-csi/blob/ec05ae6e5f096c5f8be990968f950a6915152b3e/charts/ceph-csi-cephfs/templates/nodeplugin-clusterrole.yaml#L14

@Rakshith-R I can but I looked also into the cephfs driver implementation and got the impression that it's not necessary for the cephfs rbac but it's possible that I missed something

Rakshith-R:
> > @nemcikjan Can you remove the surrounding conditions here too ?
> > https://github.com/ceph/ceph-csi/blob/ec05ae6e5f096c5f8be990968f950a6915152b3e/charts/ceph-csi-cephfs/templates/nodeplugin-clusterrole.yaml#L14
> 
> @Rakshith-R I can but I looked also into the cephfs driver implementation and got the impression that it's not necessary for the cephfs rbac but it's possible that I missed something

Yes, we're only checking if cephcsi is running inside k8s before trying to fetch the node labels now. 
Therefore, That rbac is needed regardless of the read affinity condition

iPraveenParihar:
> Thanks for the pr !
> 
> @iPraveenParihar after the new changes to obtain node labels if the pods are running in k8s cluster, we'll need rbac to get nodes regardless of topology or read affinity right ?

yes, we can remove the read affinity condition.

XtremeOwnageDotCom:
Just a note, I also had to update the provisioner role, in my cluster. It apparently also needs the ability to get nodes.

https://github.com/ceph/ceph-csi/blob/devel/charts/ceph-csi-rbd/templates/provisioner-clusterrole.yaml

The exact same resolution, would apply to it.

nemcikjan:
@XtremeOwnageDotCom I reviewed the implementation and it doesn't seem that the rbd provisioner needs that permission at all based on the condition [here](https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/driver/driver.go#L142C1-L142C1). Maybe @Rakshith-R or @iPraveenParihar could confirm.

iPraveenParihar:
> Just a note, I also had to update the provisioner role, in my cluster. It apparently also needs the ability to get nodes.
> 
> https://github.com/ceph/ceph-csi/blob/devel/charts/ceph-csi-rbd/templates/provisioner-clusterrole.yaml
> 
> The exact same resolution, would apply to it.



In RBD, we need to add the NodeServer condition as done in CephFS
https://github.com/ceph/ceph-csi/blob/23091689434a39317e4790e2312cb8fcf19800dc/internal/rbd/driver/driver.go#L128-L133

CephFS,
https://github.com/ceph/ceph-csi/blob/23091689434a39317e4790e2312cb8fcf19800dc/internal/cephfs/driver.go#L112-L118

Rakshith-R (commented):
Looks good to me, 
Can you please squash both the commits into one ?

internal/rbd/driver/driver.go
```diff
@@ -125,25 +125,21 @@ func (r *Driver) Run(conf *util.Config) {
...
 		if err != nil {
 			log.FatalLogMsg(err.Error())
 		}
+		nodeLabels, err = k8s.GetNodeLabels(conf.NodeID)
+		if err != nil {
+			log.FatalLogMsg(err.Error())
+		}
+		if conf.EnableReadAffinity {
+			crushLocationMap = util.GetCrushLocationMap(conf.CrushLocationLabels, nodeLabels)
+		}
```
Rakshith-R:
```suggestion
	if conf.IsNodeServer {
		topology, err = util.GetTopologyFromDomainLabels(conf.DomainLabels, conf.NodeID, conf.DriverName)
		if err != nil {
			log.FatalLogMsg(err.Error())
		}
		if k8s.RunsOnKubernetes() {
		        nodeLabels, err = k8s.GetNodeLabels(conf.NodeID)
		        if err != nil {
			        log.FatalLogMsg(err.Error())
		        }
		}
		if conf.EnableReadAffinity {
			crushLocationMap = util.GetCrushLocationMap(conf.CrushLocationLabels, nodeLabels)
		}
```

The current change will prevent nodeserver from running in non-K8s environment.
Let's check runsOnK8s only while getting node labels.

Rakshith-R (changes requested):

Rakshith-R:
> Looks good to me, Can you please squash both the commits into one ?

The last commit came after I pressed enter for this review. 
just one more change on the latest commit. 

Thanks

nemcikjan:
@Rakshith-R should I then squash all the commits?

Rakshith-R:
> @Rakshith-R should I then squash all the commits?

Just the first two since they are so similar
and modify the third commit according to https://github.com/ceph/ceph-csi/pull/4302#discussion_r1418683065

internal/rbd/driver/driver.go
nemcikjan: @Rakshith-R Actually, this is tricky, because `util.GetTopologyFromDomainLabels` internally also fetches node labels using `k8s.GetNodeLabels` that's why I merged the condition. I suggest to extend `util.GetTopologyFromDomainLabels` by new parameter nodeLabels so we can separately check if the controller is running in k8s and fetch the node labels and then just reuse them

Rakshith-R: > @Rakshith-R Actually, this is tricky, because `util.GetTopologyFromDomainLabels` internally also fetches node labels using `k8s.GetNodeLabels` that's why I merged the condition. I suggest to extend `util.GetTopologyFromDomainLabels` by new parameter nodeLabels so we can separately check if the controller is running in k8s and fetch the node labels and then just reuse them

@nemcikjan 
We'll now be having nodes get rbac by default for nodeserver so let's just make sure GetNodeLabels is covered inside conf.nodeserver struct for backport.

We can make this enhancement in follow-up pr.

nemcikjan: @Rakshith-R OK, I'll make appropriate changes to the scope of this PR

Rakshith-R: > Actually, this is tricky, because `util.GetTopologyFromDomainLabels` internally also fetches node labels using `k8s.GetNodeLabels` that's why I merged the condition. I suggest to extend `util.GetTopologyFromDomainLabels` by new parameter nodeLabels so we can separately check if the controller is running in k8s and fetch the node labels and then just reuse them

That part should be fine since we are now adding rbac always for nodeserver.
let's just make this change to keep GetNodeLabels inside conf.nodeserver  for backports.
We'll make that enhancement in a follow up pr. 

Please check the commitlint failure too 
https://github.com/ceph/ceph-csi/actions/runs/7126377653/job/19405449656?pr=4302

Thanks

Rakshith-R: @nemcikjan ^

Rakshith-R (dismissed):
Thanks

internal/rbd/driver/driver.go
```diff
@@ -125,17 +125,6 @@ func (r *Driver) Run(conf *util.Config) {
...
-		if err != nil {
-			log.FatalLogMsg(err.Error())
-		}
-	}
-
-	if conf.EnableReadAffinity {
-		crushLocationMap = util.GetCrushLocationMap(conf.CrushLocationLabels, nodeLabels)
-	}
-
 	// Create GRPC servers
```
Rakshith-R:
```suggestion
	if k8s.RunsOnKubernetes() && conf.IsNodeServer {
		nodeLabels, err = k8s.GetNodeLabels(conf.NodeID)
		if err != nil {
			log.FatalLogMsg(err.Error())
		}
	}
	if conf.EnableReadAffinity {
		crushLocationMap = util.GetCrushLocationMap(conf.CrushLocationLabels, nodeLabels)
	}
	
	// Create GRPC servers
```


@nemcikjan  golangci lint is complaining that the below nest if is too complex
https://github.com/ceph/ceph-csi/actions/runs/7139935245/job/19446163983?pr=4302

Let's just add conf.NodeServer check in the same place.

Rakshith-R (commented):
ci failure https://github.com/ceph/ceph-csi/actions/runs/7139935245/job/19446163983?pr=4302

Rakshith-R (approved):
Thanks !

riya-singhal31 (approved):

Rakshith-R:
@mergifyio rebase
@mergifyio queue

Rakshith-R:
@mergifyio queue

PR #4321, opened 2023-12-15
cephFS: fix fetchIP to support more formats

riya-singhal31:
the output of` ceph tell mds.0 client ls`  returns inst to be prefixed by v1 or similar.
So adding the support for same in FetchIP, so that it can fetch correct IP from the inst of that format.

example - clientInfo:  "client.24152 v1:100.64.0.7:0/3658550259"
			expectedIP:  "100.64.0.7"

### DIFF
--- a/internal/csi-addons/networkfence/fencing.go
+++ b/internal/csi-addons/networkfence/fencing.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


nixpanic (approved):

yati1998 (approved):

riya-singhal31:
Thanks for the early review @nixpanic , @yati1998

riya-singhal31:
@Mergifyio queue

PR #4337, opened 2023-12-20
revert: revert to 3.10-canary

riya-singhal31:
This commit makes template changes for
3.10 release branch to use 3.10-canary tag.

PR #4340, opened 2023-12-20
revert: use v3.10-canary as work-in-progress version again

nixpanic:
v3.10.1 has been released, we should not update the built
container-images with the same version. Use v3.10-canary for everything
that is merged in the release-v3.10 branch before we release a next
version.


nixpanic:
This is actually a duplicate of #4337

PR #4341, opened 2023-12-21
build: use go 1.21 in go.mod

nixpanic:
Some packages still have issues because older Go versions fail to tidy
go.mod because some modules require go 1.21. `go mod tidy` has now been
run with go 1.21.5, so everything is clean again.


nixpanic:
I've updated the `:devel` container in the CI, and now CI is running correctly for #4323.

PR #4364, opened 2024-01-05
cleanup: simplify `rbdGetDeviceList()`

nixpanic:
The `rbdGetDeviceList()` function uses two very similar types for
converting krbd and NBD device information from JSON. There is no need
to use this distinction, and callers of `rbdGetDeviceList()` should not
need to care about it either.

By introducing a `deviceInfo` interface with Get-functions, the
`rbdGetDeviceList()` function becomes a little simpler, with a clearly
defined API for the returned list.

### DIFF
--- a/internal/rbd/rbd_attach.go
+++ b/internal/rbd/rbd_attach.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/rbd/rbd_attach.go
```diff
@@ -91,25 +90,43 @@ var (
...
 
+type deviceInfo interface {
+	GetName() string
+	GetPool() string
+	GetRadosNamespace() string
+	GetDevice() string
+}
+
 // rbdDeviceInfo strongly typed JSON spec for rbd device list output (of type krbd).
 type rbdDeviceInfo struct {
```
Rakshith-R, line 101: We no longer have to save ID ? 
It's not used anywhere ?

nixpanic: Indeed, it is not used anywhere for attaching/mounting.

Madhu-1 (approved):

Rakshith-R (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@Mergifyio queue

PR #4378, opened 2024-01-16
cephfs: do not run `modprobe` if support is compiled into the kernel

### LINKED ISSUE #4376: Kernel mounter refuses to mount when cephfs is built into kernel
# Describe the bug #

When running on a kernel configured with `CONFIG_CEPH_FS=y` (not `=m`), the nodeplugin refuses to mount.

The `mountKernel` function [begins](https://github.com/ceph/ceph-csi/blob/devel/internal/cephfs/mounter/kernel.go#L35) with a call to `modprobe ceph`, even if the `/proc/filesystems` list already includes `ceph`.  If that modprobe command fails, it doesn't even try to mount the volume.

When the kernel has enabled cephfs as a builtin feature (not as a module), `modprobe` won't find a file named `ceph.ko`, and thus, it doesn't work.

# Environment details #

- Image/version of Ceph CSI driver : 3.10.1
- Helm chart version : ceph-csi-cephfs-3.10.1
- Kernel version : 6.6.6
- Mounter used for mounting PVC : kernel
- Kubernetes cluster version : 1.29.0
- Ceph cluster version : 18.2.1

# Steps to reproduce #

1. Build a custom kernel for your k8s nodes, which builds cephfs support directly into the kernel image.
2. Install it on your k8s nodes.
3. Make a cephfs PVC and attach it to a pod.
4. Witness the destruction.

# Actual results #

Pods fail to start.  The node plugin logs talk about modprobe failing.

# Expected behavior #

It mounts the thing and everything works.

# Logs #

```
I0116 12:10:16.999298    1227 nodeserver.go:312] ID: 18 Req-ID: 0001-0024-b5d35fba-7048-11ed-9ed9-001e0651162c-0000000000000006-a2253aa3-ef73-498b-acd5-89b97138e206 cephfs: mounting volume 0001-0024-b5d35fba-7048-11ed-9ed9-001e0651162c-0000000000000006-a2253aa3-ef73-498b-acd5-89b97138e206 with Ceph kernel client
I0116 12:10:17.000508    1227 cephcmds.go:98] ID: 18 Req-ID: 0001-0024-b5d35fba-7048-11ed-9ed9-001e0651162c-0000000000000006-a2253aa3-ef73-498b-acd5-89b97138e206 an error (exit status 1) occurred while running modprobe args: [ceph]
E0116 12:10:17.000518    1227 nodeserver.go:322] ID: 18 Req-ID: 0001-0024-b5d35fba-7048-11ed-9ed9-001e0651162c-0000000000000006-a2253aa3-ef73-498b-acd5-89b97138e206 failed to mount volume 0001-0024-b5d35fba-7048-11ed-9ed9-001e0651162c-0000000000000006-a2253aa3-ef73-498b-acd5-89b97138e206: an error (exit status 1) occurred while running modprobe args: [ceph] Check dmesg logs if required.
E0116 12:10:17.000533    1227 utils.go:169] ID: 18 Req-ID: 0001-0024-b5d35fba-7048-11ed-9ed9-001e0651162c-0000000000000006-a2253aa3-ef73-498b-acd5-89b97138e206 GRPC error: rpc error: code = Internal desc = an error (exit status 1) occurred while running modprobe args: [ceph]
```

# Additional context #

I was able to prove that modprobe is the only issue, by applying a hacky workaround.  I exec'ed into the node plugin container and replaced the `modprobe` command with a symlink to `/bin/true`.  After doing this, the node plugin mounted the cephfs volume successfully and the pods started, and everything seems to work great.

I had built my kernel this way in an attempt to optimize the kernel I use on k8s nodes, to minimize the attack surface and simplify the build/boot processes.  I think these are worthwhile goals, and while building ceph as a module is extremely common, ceph csi's dependency on that is a bit problematic.

Maybe turn the error into a warning?  Or check if `ceph` is in `/proc/filesystems` before doing the modprobe?  Or try mounting anyway before returning the error?

nixpanic:
By reading the contents of /proc/filesystems, and checking if "ceph" is
included there, running "modprobe ceph" can be skipped.

Fixes: #4376

### DIFF
--- a/internal/cephfs/mounter/kernel.go
+++ b/internal/cephfs/mounter/kernel.go
@@ -19,6 +19,8 @@ package mounter
 import (
 	"context"
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/ceph/ceph-csi/internal/cephfs/store"
 	"github.com/ceph/ceph-csi/internal/util"
@@ -27,13 +29,47 @@ import (
 const (
 	volumeMounterKernel = "kernel"
 	netDev              = "_netdev"
+	kernelModule        = "ceph"
 )
 
-type KernelMounter struct{}
+// testErrorf can be set by unit test for enhanced error reporting.
+var testErrorf = func(fmt string, args ...any) { /* do nothing */ }
 
-func mountKernel(ctx context.Context, mountPoint string, cr *util.Credentials, volOptions *store.VolumeOptions) error {
-	if err := execCommandErr(ctx, "modprobe", "ceph"); err != nil {
-		return err
+type KernelMounter interface {
+	Mount(
+		ctx context.Context,
+		mountPoint string,
+		cr *util.Credentials,
+		volOptions *store.VolumeOptions,
+	) error
+
+	Name() string
+}
+
+type kernelMounter struct {
+	// needsModprobe indicates that the ceph kernel module is not loaded in
+	// the kernel yet (or compiled into it)
+	needsModprobe bool
+}
+
+func NewKernelMounter() KernelMounter {
+	return &kernelMounter{
+		needsModprobe: !filesystemSupported(kernelModule),
+	}
+}
+
+func (m *kernelMounter) mountKernel(
+	ctx context.Context,
+	mountPoint string,
+	cr *util.Credentials,
+	volOptions *store.VolumeOptions,
+) error {
+	if m.needsModprobe {
+		if err := execCommandErr(ctx, "modprobe", kernelModule); err != nil {
+			return err
+		}
+
+		m.needsModprobe = false
 	}
 
 	args := []string{
@@ -68,7 +104,7 @@ func mountKernel(ctx context.Context, mountPoint string, cr *util.Credentials, v
 	return err
 }
 
-func (m *KernelMounter) Mount(
+func (m *kernelMounter) Mount(
 	ctx context.Context,
 	mountPoint string,
 	cr *util.Credentials,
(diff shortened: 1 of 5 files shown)


internal/cephfs/mounter/kernel.go
```diff
@@ -27,13 +29,45 @@ import (
...
+func (m *kernelMounter) mountKernel(
+	ctx context.Context,
+	mountPoint string,
+	cr *util.Credentials,
+	volOptions *store.VolumeOptions,
+) error {
+	if m.needsModprobe {
+		if err := execCommandErr(ctx, "modprobe", kernelModule); err != nil {
+			return err
+		}
```
Madhu-1, line 70: i think we can optimize it  by moving this modeprobe to `NewKernelMounter` so that we run it only once when driver starts not always WDYT?

nixpanic: There is a check to see if the driver is loaded when `NewKernelMounter()` is called. That sets `m.needsModprobe`. It is done so that `NewKernelMounter()` does not need to return an error, keeping other changes minimal.

Madhu-1: yeah  i agree but this works byt when where a mountKernel is called and if `needsModprobe` not set we keep on running `modeprobe`, but running that can be avoided if we set `needsModprobe` to `true` on successful run of modprobe,

internal/cephfs/mounter/kernel.go
```diff
@@ -27,13 +29,45 @@ import (
 const (
 	volumeMounterKernel = "kernel"
 	netDev              = "_netdev"
+	kernelModule        = "ceph"
 )
 
-type KernelMounter struct{}
+// testErrorf can be set by unit test for enhanced error reporting.
+var testErrorf = func(fmt string, args ...any) { /* do nothing */ }
```
Madhu-1, line 36: `testErrorf` might not be the right name for non-test file and also `fmt` can be replaced to something else as its a core package name.

internal/cephfs/mounter/kernel.go
Madhu-1: we can take it as an enhancement as well.

Madhu-1 (dismissed):

internal/cephfs/mounter/kernel.go
nixpanic: not sure I understand, `fmt` is just a string?

Madhu-1: what i mean is that fmt is also a package name, we might get into a static check problem if we use the same name for some variable name as well.

internal/cephfs/mounter/kernel.go
nixpanic: Ah, yes, if `modprobe` succeeded, it should be set to `true`

nixpanic: this is done now, please review again.

internal/cephfs/mounter/kernel.go
```diff
@@ -27,13 +29,47 @@ import (
...
+	mountPoint string,
+	cr *util.Credentials,
+	volOptions *store.VolumeOptions,
+) error {
+	if m.needsModprobe {
+		if err := execCommandErr(ctx, "modprobe", kernelModule); err != nil {
+			return err
+		}
+
+		m.needsModprobe = false
```
Madhu-1, line 72: i think this needs to be set to true? also this is not of much use with the current code because mounter.New() is called for every NodeStage RPC call. we need to move this to Driver initialization time as it needs to run only one time.

nixpanic: once the module is loaded, it will be listed in `/proc/filesystems`, so eventually `modprobe` is only run once

Madhu-1 (approved):

Rakshith-R (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

PR #4402, opened 2024-01-30
ci: define `i` as type `int` in for loop

nixpanic:
The output of the for loop seemt to be `groovy.lang.IntRange`, which is
not very useful. Maybe it is required to define the variable of type
`int`?


nixpanic:
I don't know why **ci/centos/jjb-validate** fails, this PR is not related to it.

Madhu-1 (approved):

nixpanic:
I hope I fixed the CI issue, but that can only be tested once the PR is merged :crossed_fingers:

PR #4491, opened 2024-03-12
cleanup: don't return an internal type from VolumeGroupJournal.Connect()

nixpanic:
The VolumeGroupJournal interface does not need to return anything except for a potential error. Any instance that implements the VolumeGroupJournal interface can be used to call all functions.

---

<details>

</details>

### DIFF
--- a/internal/journal/volumegroupjournal.go
+++ b/internal/journal/volumegroupjournal.go
@@ -36,7 +36,7 @@ type VolumeGroupJournal interface {
 	Connect(
 		monitors,
 		namespace string,
-		cr *util.Credentials) (*volumeGroupJournalConfig, error)
+		cr *util.Credentials) error
 	// Destroy frees any resources and invalidates the journal connection.
 	Destroy()
 	// SetNamespace sets the namespace for the journal.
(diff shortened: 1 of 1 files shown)


internal/journal/volumegroupjournal.go
```diff
@@ -115,14 +115,14 @@ func (sgj *volumeGroupJournalConfig) Connect(
 	monitors,
 	namespace string,
 	cr *util.Credentials,
-) (*volumeGroupJournalConfig, error) {
```
Madhu-1, line 118: we might get into problem when we serve many requests for different pools because all the journaling interface are initialized once and when are using different pool or ceph users we get connection for each one and that's the reason for returning which helps in initializing the connections, we cannot call Connect for same global object from multiple places. for now lets keep this and adjust it later or remove connect from the interface and modify it to take the input struct and return interface from it.

nixpanic: Currently it returns a pointer to itself, not a new instance of `volumeGroupJournalConfig`.

Why would someone not create multiple VolumeGroupJournal instances for different pools? If the connection/credentials to the Ceph cluster are the same, they will share the ClusterConnection already.

Madhu-1: probably we might need to address it when required, For now approving it as there is no consumer of this yet

Madhu-1 (approved):

Rakshith-R (approved):

nixpanic:
@mergifyio queue

PR #4513, opened 2024-03-21
rbd: check parent image while mirroring of child

Rakshith-R:
This commit adds code to check parent image while enabling mirroring on child image. It errors out if the parent is in trash or not mirror enabled.
This enables better communication with the user on why mirroring cannot be enabled on the child image.

### DIFF
--- a/internal/rbd/mirror.go
+++ b/internal/rbd/mirror.go
@@ -33,6 +33,39 @@ func (ri *rbdImage) EnableImageMirroring(mode librbd.ImageMirrorMode) error {
 	}
 	defer image.Close()
 
+	parentInfo, err := image.GetParent()
+	if err != nil {
+		return fmt.Errorf("failed to get parent info of image %q: %w",
+			ri, err)
+	}
+
+	if parentInfo.Image.Trash {
+		return fmt.Errorf("failed to enable mirroring on image %q:"+
+			" parent %q is in trash",
+			parentInfo.Image.ImageID, ri)
+	}
+
+	parent, err := ri.getParent()
+	if err != nil {
+		return fmt.Errorf("failed to enable mirroring on image %q:"+
+			" failed to get parent: %w", ri, err)
+	}
+
+	if parent != nil {
+		parentMirroringInfo, err := parent.GetImageMirroringInfo()
+		if err != nil {
+			return fmt.Errorf("failed to enable mirroring on image %q:"+
+				" failed to get mirroring info of parent %q: %w",
+				ri, parent, err)
+		}
+
+		if parentMirroringInfo.State != librbd.MirrorImageEnabled {
+			return fmt.Errorf("failed to enable mirroring on image %q:"+
+				"parent image %q is not enabled for mirroring",
+				ri, parent)
+		}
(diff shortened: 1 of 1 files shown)


internal/rbd/mirror.go
```diff
@@ -33,6 +33,39 @@ func (ri *rbdImage) EnableImageMirroring(mode librbd.ImageMirrorMode) error {
 	}
 	defer image.Close()
 
+	parentInfo, err := image.GetParent()
+	if err != nil {
```
Madhu-1, line 37: if there is no parent for the image this will fail with ` librbd.ErrNotFound` error isn't it?

internal/rbd/mirror.go
```diff
@@ -33,6 +33,39 @@ func (ri *rbdImage) EnableImageMirroring(mode librbd.ImageMirrorMode) error {
...
+		return fmt.Errorf("failed to get parent info of image %q: %w",
+			ri, err)
+	}
+
+	if parentInfo.Image.Trash {
+		return fmt.Errorf("failed to enable mirroring on image %q:"+
+			" parent %q is in trash",
+			parentInfo.Image.ImageID, ri)
```
Madhu-1, line 45: returning ImageID is not useful, if possible return parent name or just discard parent id or log `ImageName`

internal/rbd/mirror.go
```diff
@@ -33,6 +33,39 @@ func (ri *rbdImage) EnableImageMirroring(mode librbd.ImageMirrorMode) error {
...
+	}
+
+	parent, err := ri.getParent()
```
Madhu-1, line 48: lets add `Trash` to the `rbdImage` struct and make only one call `getParent()` (need to populate Trash value as well) which gives all the details. we don't need to call `getParent` multiple times.

internal/rbd/mirror.go
```diff
@@ -33,6 +33,39 @@ func (ri *rbdImage) EnableImageMirroring(mode librbd.ImageMirrorMode) error {
...
+	if parent != nil {
+		parentMirroringInfo, err := parent.GetImageMirroringInfo()
+		if err != nil {
+			return fmt.Errorf("failed to enable mirroring on image %q:"+
+				" failed to get mirroring info of parent %q: %w",
+				ri, parent, err)
+		}
+
+		if parentMirroringInfo.State != librbd.MirrorImageEnabled {
+			return fmt.Errorf("failed to enable mirroring on image %q:"+
```
Madhu-1, line 63: we need to return Invalid RPC code when parent in not mirror enabled or when the parent is in trash?

Rakshith-R: https://github.com/csi-addons/spec/tree/main/replication#enablevolumereplication

https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume-errors

So I'll modify csi addons spec EnableVolReplication call errors' INVALID_ARGUEMENT description to include something similar to createvolume_errors  indicating we cannot enable replication on certain volumes.

Rakshith-R:
Will replace this with a new pr.

PR #4566, opened 2024-04-18
util: add support to configure mirror daemon count

### LINKED ISSUE #4312: rbd: add support for adjusting the number of RBD mirror image watchers
Since ceph-csi release v3.3.1 the workflow of RBD volume atttachment considers 2 watchers to check on the image when mirroring primary is true (https://github.com/ceph/ceph-csi/pull/1993).

If you operate your ceph cluster environment with more than one rbd-mirror daemon (e.g. for high availibility and/or the cluster is streched over more than one datacenter) you maybe want to adjust the number of image watchers to check on the image.

It's required to run at least 2 rbd-mirror daemons if you want to ensure non-disruptive RBD mirroring. This is not a problem for the "in use" logic of mirrored images because only 1 rbd-mirror client watches the image usually.
But we observed volume attachment issues with 2 rbd-mirror clients when a mirror daemon leader becomes temporarily unresponsive (for example due to heavy load in the peering cluster).
That's not an serious issue for the rbd mirroring because of redundancy but meanwhile for the volume attachment of re-scheduled stateful pods.

In addition in some cases you even want to run more than 2 rbd-mirror daemons to balance the leaders over multiple RBD pools. This setup is not possible with the current implementation because you have 2 watchers on every image with 3 daemons for example (under normal operation).

Current implementation tolerates only one watcher if mirroring primary is true (https://github.com/ceph/ceph-csi/blob/v3.10.0/internal/rbd/rbd_util.go#L553-L559). I think it would be very helpful to make it possible adjusting the number of rbd-mirror watchers in more complex ceph environments.

Madhu-1:

Currently, we are assuming that only one rbd mirror daemon running on the ceph cluster but that is not true for many cases and it can be more than one, this PR makes this a configurable parameter.

Do you have any questions?

yes 

No

Fixes: #4312

### DIFF
--- a/internal/util/csiconfig.go
+++ b/internal/util/csiconfig.go
@@ -45,6 +45,7 @@ const (
 	"clusterID": "<cluster-id>",
 	"rbd": {
 		"radosNamespace": "<rados-namespace>"
+		"mirrorDaemonCount": 1
 	},
 	"monitors": [
 		"<monitor-value>",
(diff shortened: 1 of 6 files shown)


internal/util/csiconfig.go
```diff
@@ -45,6 +45,7 @@ const (
 	"clusterID": "<cluster-id>",
 	"rbd": {
 		"radosNamespace": "<rados-namespace>"
+		"mirrorDaemonCount": "1"
```
nixpanic: Great test case, but unfortunately missing in the unit tests.

Please add a test-case that expects an error when the argument is a number as `"string"` (maybe convert it?) or even an invalid number.

nixpanic (changes requested):

nixpanic (approved):

Rakshith-R (approved):

Rakshith-R:
@megifyio rebase
@mergifyio queue

Madhu-1:
@mergifyio rebase

Madhu-1:
@Mergifyio queue

PR #4578, opened 2024-04-23
ci: do not require Kubernetes 1.26 tests for the devel branch

nixpanic:
Kubernetes 1.26 is End-Of-Life, the next version of Ceph-CSI will not
support that anymore.


.mergify.yml
```diff
@@ -67,25 +64,6 @@ queue_rules:
               - "status-success=ci/centos/mini-e2e/k8s-1.28"
               - "status-success=ci/centos/upgrade-tests-cephfs"
               - "status-success=ci/centos/upgrade-tests-rbd"
-          - and:
```
Rakshith-R: Let's keep this ? 
We'll need to re add when k8s test versions differs between devel and release branches?

nixpanic: I think we can do that when there is a need. Currently things (rules to queue, rules to merge) are not in-sync, this PR corrects that too.

Madhu-1: @nixpanic we need to readd everything in that case,. keep as its and remove 1.26 from devel for sure and if we are not planning to run that on release-3.11 we can drop it from above section. lets not remove the entire section for devel.

nixpanic: Ok! I've added the rule for the devel branch in that case, so that everything is the same everywhere.

Madhu-1 (approved):

iPraveenParihar:
@nixpanic, 
can we drop 1.26 for v3.11.0?
https://github.com/ceph/ceph-csi/blob/4b49d2898f5f034f76a8b7df7dcde785a954805c/README.md?plain=1#L57-L62

Rakshith-R (approved):

Rakshith-R:
> @nixpanic, can we drop 1.26 for v3.11.0?
> 
> https://github.com/ceph/ceph-csi/blob/4b49d2898f5f034f76a8b7df7dcde785a954805c/README.md?plain=1#L57-L62

That release is already out, I think we can remove it for 3.11.1 when it happens

Rakshith-R:
@mergifyio queue

PR #4584, opened 2024-04-25
build: prevent "Operation not supported" failures while extracting Golang in CI jobs

nixpanic:
It seems GitHub has an issue with the Fedora 40 container image, extracting the Golang tarball fails. The Fedora 39 image does not have this problem, so use that for the time being.

### DIFF
--- a/scripts/Dockerfile.test
+++ b/scripts/Dockerfile.test
@@ -8,7 +8,7 @@
 # little different.
 #
 
-FROM registry.fedoraproject.org/fedora:latest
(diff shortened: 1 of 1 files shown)


Madhu-1:
@nixpanic this CI passed, we need to backport this to 3.10 and 3.11 as well.

Madhu-1 (approved):

iPraveenParihar (approved):

iPraveenParihar:
@Mergifyio queue

Issue #4569, opened 2024-04-19
update kubernetes dependency to 1.30


Madhu-1:
Kubernetes 1.30 is released  https://kubernetes.io/blog/2024/04/17/kubernetes-v1-30-release/ let's make use of it in cephcsi

- [ ]  Make Kubernetes E2E test mandatory in CI Run
- [x] Update Mergify rules to consider tests on kube 1.30
- [x] Drop un maintained kube releases
https://github.com/ceph/ceph-csi/pull/4589
- [x] go dependency update
https://github.com/ceph/ceph-csi/pull/4593
- [ ] Readme update

Madhu-1:
@iPraveenParihar anything pending on this one?

iPraveenParihar:
> @iPraveenParihar anything pending on this one?

@Madhu-1, I have updated tasks as completed with PRs linked in the Issue description.
- For k8s E2E tests mandatory in CI run, we do have conditions in mergify right?
https://github.com/ceph/ceph-csi/blob/aa2cc67650958e263637e9d5f3bf936dae105c30/.mergify.yml#L75-L78
did you mean anything else?

- For Readme update, I don't see anything to be done. 
As we only updated k8s 1.30 in devel, not in release v3.11. Or did I miss anything? 🤔

Madhu-1:
Yes i meant the same, in that case can we close it if all the required works are completed?

iPraveenParihar:
Yes, good to close.

PR #4593, opened 2024-04-29
build: use Go 1.22

nixpanic:
Kubernetes 1.30 packages require Go 1.22 or newer.

With a new Go version, **golangci-lint** reports more issues, these are addressed as well.

golangci-lint reports these:

    The copy of the 'for' variable "kmsID" can be deleted (Go 1.22+)
    (copyloopvar)

golangci-lint warns about this:

    for loop can be changed to use an integer range (Go 1.22+)
    (intrange)

gosec warns about it like this:

    G601: Implicit memory aliasing in for loop. (gosec)

Related-to: #4569

### DIFF
--- a/e2e/errors_test.go
+++ b/e2e/errors_test.go
@@ -75,7 +75,6 @@ Error from server (AlreadyExists): error when creating "STDIN": deployments.apps
 		},
 	}
 	for _, tt := range tests {
-		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			if got := getStdErr(tt.errString); got != tt.expected {
--- a/internal/cephfs/groupcontrollerserver_test.go
+++ b/internal/cephfs/groupcontrollerserver_test.go
@@ -106,13 +106,12 @@ func TestControllerServer_validateCreateVolumeGroupSnapshotRequest(t *testing.T)
 		},
 	}
 	for _, tt := range tests {
-		ts := tt
-		t.Run(ts.name, func(t *testing.T) {
+		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			err := cs.validateCreateVolumeGroupSnapshotRequest(ts.args.ctx, ts.args.req)
-			if ts.wantErr {
+			err := cs.validateCreateVolumeGroupSnapshotRequest(tt.args.ctx, tt.args.req)
+			if tt.wantErr {
 				c := status.Code(err)
-				if c != ts.code {
+				if c != tt.code {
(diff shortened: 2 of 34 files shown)


e2e/errors_test.go
```diff
@@ -75,7 +75,6 @@ Error from server (AlreadyExists): error when creating "STDIN": deployments.apps
 		},
 	}
 	for _, tt := range tests {
-		tt := tt
```
Rakshith-R, line 78: I think this is required to be done since we run tests in parallel.
let's rename the in loop variable `currentTT := tt`

internal/cephfs/groupcontrollerserver_test.go
```diff
@@ -106,13 +106,12 @@ func TestControllerServer_validateCreateVolumeGroupSnapshotRequest(t *testing.T)
 		},
 	}
 	for _, tt := range tests {
-		ts := tt
-		t.Run(ts.name, func(t *testing.T) {
+		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
```
Rakshith-R, line 110: Same as above comment.
We'll need the local in loop variable since the test will be run in parallel.

internal/cephfs/nodeserver_test.go
```diff
@@ -143,28 +143,27 @@ func Test_setMountOptions(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		tc := tt
```
Rakshith-R, line 146: Same as above comment.
We'll need the local in loop variable since the test will be run in parallel.

Rakshith-R (changes requested):
Only this linter commit needs a another approach.

https://github.com/ceph/ceph-csi/pull/4593/commits/0fc86dfdb2a1f99caea7d41790cc6d27c6aa0519

e2e/errors_test.go
nixpanic: This is actually something that is a new feature in Golang 1.22, which makes the extra assigning not needed anymore. https://go.dev/blog/loopvar-preview has more details about it, as well as the **References to loop variables** section in the [Go 1.22 release notes](https://tip.golang.org/doc/go1.22).

Rakshith-R: understood, That's awesome ! :D

Rakshith-R (approved):

yati1998 (approved):

nixpanic:
@mergifyio queue

nixpanic:
@mergifyio requeue

nixpanic:
> /retest ci/centos/mini-e2e-helm/k8s-1.28

Failed with a mismatch of backend images

PR #4655, opened 2024-06-03
docs: proposal for providing PV key rotation

black-dragon74:
The design and implementation details for rotating the encryption keys for volumes.


docs/design/proposals/pv-key-rotation.md
```diff
@@ -0,0 +1,55 @@
...
+- Create a `rbdvolume` object using volume ID, this is done by `GenVolFromVolID`.
+- Fetch the current key from the KMS, it is needed for subsequent LUKS operations.
+- Get the device path for the volume by calling `waitForPath` as all LUKS operations require the device path.
+- Add the fetched key to LUKS slot 1, this will serve as a backup of the key.
+- Generate a new key and store it locally. It will be updated in the KMS at later steps.
+- Remove the exsitng key from slot 0 upon verifying that the key in KMS == the key in slot 0.
+- Add new key to slot 0 and then call `LuksVerifyKey` to verify that the slot was successfully updated.
+- Update the new key in the KMS.
+- Fetch the key again and verify that the key in KMS == the new key we generated.
+- We can now remove the backup key from slot 1.
```
Rakshith-R: Let's keep the process the same as the one implemented here https://github.com/rook/rook/blob/ff448e11d8836388e976d0b7de74f517cd76e275/pkg/daemon/ceph/osd/key_rotation.go#L31

docs/design/proposals/pv-key-rotation.md
```diff
@@ -0,0 +1,55 @@
...
+
+The `VolumeEncryption` struct is modified to make `generateNewEncryptionPassphrase` a public member function.
+
+A metadata is set on the RBD image to indicate that the image is being processed for keyrotation. Presence of this metadata will prevent the same image being processed again.
+
+The following steps are followed to process the device for key rotation:
+
+- Create a `rbdvolume` object using volume ID, this is done by `GenVolFromVolID`.
+- Fetch the current key from the KMS, it is needed for subsequent LUKS operations.
+- Get the device path for the volume by calling `waitForPath` as all LUKS operations require the device path.
```
Rakshith-R: How are you going to determine the device path ?

docs/design/proposals/pv-key-rotation.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Implementation Details
+
+The encryption key rotation request will contain with it the volume ID, credentials and secrets.
+
+These values are then used to call `GenVolFromVolID` to get the rbdVolume structure.
+
+The `VolumeEncryption` struct is modified to make `generateNewEncryptionPassphrase` a public member function.
+
+A metadata is set on the RBD image to indicate that the image is being processed for keyrotation. Presence of this metadata will prevent the same image being processed again.
```
Rakshith-R: What's the purpose of setting metadata and which server controller or node will set this / respond to this ? 
When will the rotation be triggered?

docs/design/proposals/pv-key-rotation.md
```diff
@@ -0,0 +1,55 @@
...
+
+The following new methods are added to `cryptsetup.go` for handling the key rotation.
+
+- `LuksAddKey`: Adds a new key to specified LUKS slot
+- `LuksRemoveKey`: Removes the specified key from its slot using `luksKillSlot`
+- `LuksVerifyKey`: Verifies that the given key exists in the given slot using `luksChangeKey`.
+
+### Implementation Details
+
+The encryption key rotation request will contain with it the volume ID, credentials and secrets.
```
Rakshith-R: credentials and secrets are one and the same here ?

docs/design/proposals/pv-key-rotation.md
```diff
@@ -0,0 +1,55 @@
...
+
+These values are then used to call `GenVolFromVolID` to get the rbdVolume structure.
+
+The `VolumeEncryption` struct is modified to make `generateNewEncryptionPassphrase` a public member function.
+
+A metadata is set on the RBD image to indicate that the image is being processed for keyrotation. Presence of this metadata will prevent the same image being processed again.
+
+The following steps are followed to process the device for key rotation:
+
+- Create a `rbdvolume` object using volume ID, this is done by `GenVolFromVolID`.
```
Madhu-1: please separate the controller and node-plugin process here and also mention how you will handle RWX PVC as well.

Rakshith-R (dismissed):

docs/design/proposals/pv-key-rotation.md
```diff
@@ -0,0 +1,55 @@
+# Encryption Key Rotation
```
Madhu-1: update the file name `rbd-pv-key-rotation` as this is specific to rbd.

Madhu-1 (changes requested):

docs/design/proposals/pv-key-rotation.md
black-dragon74: You mean the wordings of this doc? As the steps are essentially the same.

docs/design/proposals/pv-key-rotation.md
black-dragon74: As per my current understanding, calling `waitForPath` defined in `rbd_attach.go` can get us the device path. Am I wrong to assume that?

docs/design/proposals/pv-key-rotation.md
black-dragon74: We will register this service only on the node server hence the node server will set the metadata. Before doing that it will ensure that it is not already present.

The purpose of setting this metadata is to serve as a indication to avoid processing of the same PVC while it is already being processed. Might be extraneous though. If so, we can ditch this step.

The rotation will be triggered when the node server receives the RPC call from csi-addons. Which will be in turn be triggered by the cron schedule defined in the annotated storage class.

docs/design/proposals/pv-key-rotation.md
black-dragon74: Ack.

docs/design/proposals/pv-key-rotation.md
black-dragon74: I will update this. We use the secrets to generate the credentials. The request will contain volumeID and secrets.

docs/design/proposals/pv-key-rotation.md
black-dragon74: Ack.

docs/design/proposals/pv-key-rotation.md
Rakshith-R: The volume/operation locks which we have will serve that purpose ? 

I guess metadata is required if we have rbd mounted in multiple nodes but even with metadata set if controller pings another nodeplugin after first request then the status will get updated to complete?

docs/design/proposals/pv-key-rotation.md
Rakshith-R: LuksVerify function is not used in the above linked function.
Let's keep it the same so as to avoid extra work and having different steps in each place.

black-dragon74: Ack.

docs/design/proposals/pv-key-rotation.md
black-dragon74: In case of RWX volumes we will send the request to only one of the nodes, as the underlying device is the same and doing it on one node will do it for all.

Ditching the metadata part then :)

Rakshith-R: > In case of RWX volumes we will send the request to only one of the nodes, as the underlying device is the same and doing it on one node will do it for all.
> 
> Ditching the metadata part then :)

Yea, one more problem will be how to decide which node to target and send requests only to that.

black-dragon74: What I have so far is to get a list of vol attachments and see if the node advertises KeyRotate identity. If so, we send the request to it. We can safely ignore the rest of the nodes in the attachment list. This will work for both type of volumes as we need to find out the node to send rpc requests to anyways. WDYT?

Rakshith-R: My main concern is request timing out and then we may choose another node in a follow up request. Operation locks cannot help in this case.

Madhu-1: what if we have a a recovery key generated by the controller and store it in the KMS temporarily until the existing key is updated in KMS? (we still need to check and see if this is a feasible and accepted solution)

black-dragon74: Just a thought, what about we requeue the reconcile and the next time we do not forward the request to the node that timed out? Leaning towards this as we won't have to change the existing KMS impl this way and we'll have much less processing on csi driver side of things.

Madhu-1: You need to think about the side effect it going to have multiple key or key not matching which is in KMS with the one actual in-use.

docs/design/proposals/rbd-pv-key-rotation.md
```diff
@@ -0,0 +1,57 @@
...
+- `LuksRemoveKey`: Removes the specified key from its slot using `luksKillSlot`
+- `LuksVerifyKey`: Verifies that the given key exists in the given slot using `luksChangeKey`.
+
+### Implementation Details
+
+The encryption key rotation request will contain with it the volume ID and secrets.
+
+The secrets are used to generate the credentials for autheticating aginst a ceph cluster.
+
+These values are then used to call `GenVolFromVolID` to get the rbdVolume structure.
```
Madhu-1: `aginst` to `against`

docs/design/proposals/rbd-pv-key-rotation.md
```diff
@@ -0,0 +1,57 @@
...
+- Generate a new key and store it locally. It will be updated in the KMS at later steps.
+- Remove the exsitng key from slot 0 upon verifying that the key in KMS == the key in slot 0.
+- Add new key to slot 0.
+- Update the new key in the KMS.
+- Fetch the key again and verify that the key in KMS == the new key we generated.
+- We can now remove the backup key from slot 1.
+
+These order of the above steps guarantees that we always have one key that can unlock the encrypted volume.
+
+In case of RWX volumes, we send the request to ONLY one of the nodes where the volume is mounted. As the underlying device is same, rotating the key on any one of the nodes will suffice and prevent us from additional performance overheads.
```
Madhu-1: This is not yet decided right? Please mark RWX as out of scope/future from this one.

docs/design/proposals/rbd-pv-key-rotation.md
```diff
@@ -0,0 +1,57 @@
...
+## Document Terminology
+
+- Encryption Key: The passphrase that is used to encrypt and open the device.
+- LUKS: The specification used by dm-crypt to process encrypted volumes on linux.
+
+## Proposed Solution
+
+The proposed solution in this document, is to address the periodic rotation of encryption keys for encrypted volumes.
+
+This document outlines the rotation steps for PVCs backed by RBD and will be updated with other volume types as they are supported.
```
Madhu-1: `will be updated with other volume types as they are supported` what is other volume types here?

black-dragon74: Cephfs backed volumes encrypted using fscrypt.

Madhu-1: This doc is for rbd only as the doc name suggest, we don't need to cover cephfs specific things in this doc

Madhu-1 (dismissed):

Madhu-1:
@black-dragon74 please fix the CI failures

docs/design/proposals/rbd-pv-key-rotation.md
```diff
@@ -0,0 +1,77 @@
+# Encryption Key Rotation
+
+## Proposal
+
+Subject of this proposal is to add support for rotation of
+encryption keys (KEKs) for encrypted volumes in Ceph-CSI.
+
+Support for rotating keys on RWX/ROX volumes is out of scope for now
+and shall be added later.
```
nixpanic: Also add that filesystem encryption with `fscrypt` is out of scope, this is only for block-device encryption with LUKS. The RBD backend supports both, CephFS only supports `fscrypt`, NFS supports none.

docs/design/proposals/rbd-pv-key-rotation.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Document Terminology
+
+- Encryption Key: The passphrase that is used to encrypt and open the device.
+- LUKS: The specification used by dm-crypt to process encrypted volumes on linux.
+
+## Proposed Solution
+
+The proposed solution in this document, is to address the periodic rotation
```
nixpanic: How is this a proposal for `periodic` rotation?  There is no scheduler that does the key rotation at certain intervals/periods. This is a description of what the CSI-Addons API requests will do, who/whatever the API calls at what intervals/periods is out of scope for the implementation of the API.

docs/design/proposals/rbd-pv-key-rotation.md
```diff
@@ -0,0 +1,77 @@
...
+of encryption keys for encrypted volumes.
+
+This document outlines the rotation steps for PVCs backed by RBD.
+
+### Implementation Summary
+
+This feature builds upon the foundation laid by encrypted pvcs.
+
+An existing storage class can be annotated with
```
nixpanic: this is not part of Ceph-CSI, you can probably refer to the documentation in the Kubernetes CSI-Addons repository for these details.

docs/design/proposals/rbd-pv-key-rotation.md
```diff
@@ -0,0 +1,77 @@
...
+- Create a `rbdvolume` object using volume ID,
+  this is done by `GenVolFromVolID`.
+- Fetch the current key from the KMS, it is needed for
+  subsequent LUKS operations.
+- Get the device path for the volume by calling `waitForPath` as all LUKS
+  operations require the device path.
+- Add the fetched key to LUKS slot 1, this will serve as a backup of the key.
+- Generate a new key and store it locally. It will be updated
+  in the KMS at later steps.
+- Remove the exsitng key from slot 0 upon verifying that the
```
nixpanic: typo: exsitng

docs/design/proposals/rbd-pv-key-rotation.md
```diff
@@ -0,0 +1,77 @@
...
+  key in KMS == the key in slot 0.
+- Add new key to slot 0.
+- Update the new key in the KMS.
+- Fetch the key again and verify that the
+  key in KMS == the new key we generated.
+- We can now remove the backup key from slot 1.
+
+These order of the above steps guarantees that we always have one key
```
nixpanic: note that the key in the KMS can always be used to unlock the volume, a small but important detail :wink:

Madhu-1 (approved):

nixpanic (approved):

nixpanic:
@mergifyio queue

PR #4671, opened 2024-06-10
rbd: use internal as default error code in getGRPCError()

Rakshith-R:

This commit replaces `codes.Unknown` with `codes.Internal` as the default error code in getGRPCError()
since it much better suited.

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -776,14 +776,10 @@ func getGRPCError(err error) error {
 	}
 
 	errorStatusMap := map[error]codes.Code{
-		corerbd.ErrFetchingLocalState:          codes.Internal,
-		corerbd.ErrResyncImageFailed:           codes.Internal,
-		corerbd.ErrDisableImageMirroringFailed: codes.Internal,
-		corerbd.ErrFetchingMirroringInfo:       codes.Internal,
-		corerbd.ErrInvalidArgument:             codes.InvalidArgument,
-		corerbd.ErrAborted:                     codes.Aborted,
-		corerbd.ErrFailedPrecondition:          codes.FailedPrecondition,
-		corerbd.ErrUnavailable:                 codes.Unavailable,
+		corerbd.ErrInvalidArgument:    codes.InvalidArgument,
+		corerbd.ErrAborted:            codes.Aborted,
+		corerbd.ErrFailedPrecondition: codes.FailedPrecondition,
+		corerbd.ErrUnavailable:        codes.Unavailable,
 	}
 
 	for e, code := range errorStatusMap {
@@ -792,8 +788,8 @@ func getGRPCError(err error) error {
 		}
 	}
 
-	// Handle any other non nil error not listed in the map
-	return status.Error(codes.Unknown, err.Error())
+	// Handle any other non nil error not listed in the map as internal error
+	return status.Error(codes.Internal, err.Error())
 }
 
 // GetVolumeReplicationInfo extracts the RBD volume information from the volumeID, If the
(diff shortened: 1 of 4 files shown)


nixpanic:
Hi @Rakshith-R,

Can you explain why `Internal` is **much better suited** than `Unknown` for an unrecognized error?
Does this influence the behavior of CSI callers on these errors?

Rakshith-R:
> Hi @Rakshith-R,
> 
> Can you explain why `Internal` is **much better suited** than `Unknown` for an unrecognized error? Does this influence the behavior of CSI callers on these errors?

Unknown and Internal both will get retries from the callers.

We don't use Unknown anywhere else. 
Most command execution failures are treated as internal and not unknown elsewhere in cephcsi.
We should default to that instead of Unknown.

Rakshith-R:
> > Hi @Rakshith-R,
> > Can you explain why `Internal` is **much better suited** than `Unknown` for an unrecognized error? Does this influence the behavior of CSI callers on these errors?
> 
> Unknown and Internal both will get retries from the callers.
> 
> We don't use Unknown anywhere else. Most command execution failures are treated as internal and not unknown elsewhere in cephcsi. We should default to that instead of Unknown.

https://github.com/ceph/ceph-csi/commit/cdaa9264eb4ee1698efa99ab48224049e7f7c64f

This commit introduced unknown grpc error code. 
Without using internal as default, we would need to export each error message and add it to the list every time a new code block it added.

nixpanic:
> > > Hi @Rakshith-R,
> > > Can you explain why `Internal` is **much better suited** than `Unknown` for an unrecognized error? Does this influence the behavior of CSI callers on these errors?
> > 
> > 
> > Unknown and Internal both will get retries from the callers.
> > We don't use Unknown anywhere else. Most command execution failures are treated as internal and not unknown elsewhere in cephcsi. We should default to that instead of Unknown.
> 
> [cdaa926](https://github.com/ceph/ceph-csi/commit/cdaa9264eb4ee1698efa99ab48224049e7f7c64f)
> 
> This commit introduced unknown grpc error code. Without using internal as default, we would need to export each error message and add it to the list every time a new code block it added.

Except that it is easy to forget to add a new error code (we should have unit tests for that), considering a matching gRPC error code is probably better than always returning `Internal`?

This change is not bad, but I am not sure how it improves things. In order to prevent the `Unknown` error from being returned, maybe log a warning marked as bug with call-stack so that it needs to be addressed?

Rakshith-R:
> > > > Hi @Rakshith-R,
> > > > Can you explain why `Internal` is **much better suited** than `Unknown` for an unrecognized error? Does this influence the behavior of CSI callers on these errors?
> > > 
> > > 
> > > Unknown and Internal both will get retries from the callers.
> > > We don't use Unknown anywhere else. Most command execution failures are treated as internal and not unknown elsewhere in cephcsi. We should default to that instead of Unknown.
> > 
> > 
> > [cdaa926](https://github.com/ceph/ceph-csi/commit/cdaa9264eb4ee1698efa99ab48224049e7f7c64f)
> > This commit introduced unknown grpc error code. Without using internal as default, we would need to export each error message and add it to the list every time a new code block it added.
> 
> Except that it is easy to forget to add a new error code (we should have unit tests for that), considering a matching gRPC error code is probably better than always returning `Internal`?
> 
> This change is not bad, but I am not sure how it improves things. In order to prevent the `Unknown` error from being returned, maybe log a warning marked as bug with call-stack so that it needs to be addressed?

I think the whole way of creating constants of errors in rbd package, exporting them and using them in csiaddons package should be used only when you need a grpc error code that is not internal.

Adopting the export and use method for internal code too would be way too tedious according to me.

Most of the places we error out with internal error code. We don't use unknown code at all in other places.
Please take a look, I've replaced all the error messages which used internal error code.

nixpanic (approved):
Error constants that are used only once don't make much sense to me either.

iPraveenParihar (approved):

Rakshith-R:
@mergifyio queue

PR #4687, opened 2024-06-18
journal: Update group journal to accomudate rbd group

Madhu-1:
Updated the group journal to remove snapshot-specific things from it and make it generic as it can be reused for rbd as well and also added a new key to the volume journal to store the groupID which is required to track the groupID of the volume from which we can identify the rbd image belongs to which group.


nixpanic (approved):
Looks useful to me. I'm not completely sure yet how to use this within rbd grouping, but I hope to be able to  try it out soon.

Rakshith-R (approved):

Madhu-1:
@mergifyio queue

PR #4697, opened 2024-06-27
Set object lock for volumes for cephfs encryption

### LINKED ISSUE #4654: Concurrency issue when pvc is mounted to multiple pods
# Describe the bug #

This is an extension/reopen of [https://github.com/ceph/ceph-csi/issues/4592](url). Our test showed that when creating a deployment of replica 2, with high chance that one pod will fail to mount the pvc. 
# Environment details #

Same as [https://github.com/ceph/ceph-csi/issues/4592](url)

# Steps to reproduce #

1. Change the pod yaml in issue 4592 to a deployment
1. Make replicas to 2(or high enough so pods will be running on different nodes)

# Actual results #
Only one pod will be running

# Expected behavior #
All pods will be running

Sunnatillo:

Do you have any questions?

For example:

## Related issues ##

Closes: #4688
Fixes: #4654

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -23,6 +23,8 @@ import (
 	"os"
 	"path"
 	"strings"
+	"syscall"
+	"time"
 
 	cerrors "github.com/ceph/ceph-csi/internal/cephfs/errors"
 	"github.com/ceph/ceph-csi/internal/cephfs/mounter"
@@ -127,13 +129,70 @@ func maybeUnlockFileEncryption(
 	stagingTargetPath string,
 	volID fsutil.VolumeID,
 ) error {
-	if volOptions.IsEncrypted() {
-		log.DebugLog(ctx, "cephfs: unlocking fscrypt on volume %q path %s", volID, stagingTargetPath)
+	if !volOptions.IsEncrypted() {
+		return nil
+	}
 
-		return fscrypt.Unlock(ctx, volOptions.Encryption, stagingTargetPath, string(volID))
+	// Define Mutex Lock variables
+	lockName := string(volID) + "-mutexLock"
+	lockDesc := "Lock for " + string(volID)
+	lockDuration := 150 * time.Second
+	// Generate a consistent lock cookie for the client using hostname and process ID
+	lockCookie := generateLockCookie()
+	var flags byte = 0
+
+	log.DebugLog(ctx, "Creating lock for the following volume ID %s", volID)
+
+	ioctx, err := volOptions.GetConnection().GetIoctx(volOptions.MetadataPool)
+	if err != nil {
+		log.ErrorLog(ctx, "Failed to create ioctx: %s", err)
+
+		return err
+	}
+	defer ioctx.Destroy()
+
+	res, err := ioctx.LockExclusive(volOptions.VolID, lockName, lockCookie, lockDesc, lockDuration, &flags)
+	if res != 0 {
+		switch res {
+		case -int(syscall.EBUSY):
+			return fmt.Errorf("Lock is already held by another client and cookie pair for %v volume", volID)
+		case -int(syscall.EEXIST):
+			return fmt.Errorf("Lock is already held by the same client and cookie pair for %v volume", volID)
+		default:
+			return fmt.Errorf("Failed to lock volume ID %v: %w", volID, err)
+		}
+	}
+	log.DebugLog(ctx, "Lock successfully created for volume ID %s", volID)
+
+	log.DebugLog(ctx, "cephfs: unlocking fscrypt on volume %q path %s", volID, stagingTargetPath)
+	err = fscrypt.Unlock(ctx, volOptions.Encryption, stagingTargetPath, string(volID))
+	if err != nil {
+		return err
+	}
+
+	ret, err := ioctx.Unlock(string(volID), lockName, lockCookie)
+	switch ret {
+	case 0:
+		log.DebugLog(ctx, "Lock %s successfully released ", lockName)
+	case -int(syscall.ENOENT):
+		log.DebugLog(ctx, "Lock is not held by the specified %s, %s pair", lockCookie, lockName)
+	default:
+		log.ErrorLog(ctx, "Failed to release following lock, this will lead to orphan lock %s: %v",
+			lockName, err)
 	}
 
 	return nil
+}
+
+// generateLockCookie generates a consistent lock cookie for the client.
+func generateLockCookie() string {
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown-host"
+	}
+	pid := os.Getpid()
+
+	return fmt.Sprintf("%s-%d", hostname, pid)
 }
 
 // maybeInitializeFileEncryption initializes KMS and node specifics, if volContext enables encryption.


internal/cephfs/nodeserver.go
```diff
@@ -127,13 +135,72 @@ func maybeUnlockFileEncryption(
...
+		"This is some pod here",
+		retryoptions,
+	)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to aquire lock %s: %v", lockName, err)
+
+		return err
+	}
+
+	if created {
```
nixpanic: Usually we write code in a way that handles errors or unexpected results, and bails out, only when everything succeeds it continues in the function.

```go
if !created {
    return fmt.Errorf("There is already one file system with name %s", string(volID))
}

defer func() {
...
```

nixpanic (commented):
just a nit to improve the readability of the code a little, looks quite good already

internal/cephfs/nodeserver.go
```diff
@@ -127,13 +128,47 @@ func maybeUnlockFileEncryption(
...
+	lockName := string(volID) + "-mutexLock"
+	lockCookie := lockName + "-coockie"
+	lockDesc := "Lock for " + string(volID)
+	lockDuration := 40 * time.Second
+	var flags byte = 0
+
+	log.DebugLog(ctx, "Creating lock for the following volume ID %s", volID)
+
+	ioctx, err := volOptions.GetConnection().GetIoctx(volOptions.MetadataPool)
+	if err != nil {
```
Madhu-1, line 147: return error from here?

internal/cephfs/nodeserver.go
```diff
@@ -127,13 +128,47 @@ func maybeUnlockFileEncryption(
...
+	}
+	defer ioctx.Destroy()
+
+	res, err := ioctx.LockExclusive(volOptions.VolID, lockName, string(lockCookie), lockDesc, lockDuration, &flags)
```
Madhu-1:
```suggestion
res, err := ioctx.LockExclusive(volOptions.VolID, lockName, lockCookie, lockDesc, lockDuration, &flags)
```
lockCookie is already a string

internal/cephfs/nodeserver.go
```diff
@@ -127,13 +128,47 @@ func maybeUnlockFileEncryption(
...
 
+	if volOptions.IsEncrypted() == false {
+		return nil
+	}
+
+	// Define Mutex Lock variables
+	lockName := string(volID) + "-mutexLock"
+	lockCookie := lockName + "-coockie"
+	lockDesc := "Lock for " + string(volID)
+	lockDuration := 40 * time.Second
```
Madhu-1: should we have some higher value for this one like (150S) default is 180S in kubelet? and make it configurable so user can adjust this one based on the network connectivity with ceph cluster?

internal/cephfs/nodeserver.go
```diff
@@ -127,13 +128,47 @@ func maybeUnlockFileEncryption(
...
+
+	if res == 0 {
+		log.DebugLog(ctx, "cephfs: unlocking fscrypt on volume %q path %s", volID, stagingTargetPath)
 		return fscrypt.Unlock(ctx, volOptions.Encryption, stagingTargetPath, string(volID))
 	}
+	defer func() {
+		log.DebugLog(ctx, "Releasing following lock  %s", lockName)
 
-	return nil
+		_, err = ioctx.Unlock(string(volID), lockName, lockCookie)
```
Madhu-1: wont we get into the problem if we try to unlock if someone else is already holding the lock? in case of RWX and multiple nodes

internal/cephfs/nodeserver.go
```diff
@@ -127,13 +128,47 @@ func maybeUnlockFileEncryption(
...
+
+	res, err := ioctx.LockExclusive(volOptions.VolID, lockName, string(lockCookie), lockDesc, lockDuration, &flags)
+	if err != nil {
+		return fmt.Errorf("Failed to lock volume ID %v: %s", volID, err)
+	}
+	log.DebugLog(ctx, "Lock successfully created for the volume ID %s", volID)
+
+	if res == 0 {
+		log.DebugLog(ctx, "cephfs: unlocking fscrypt on volume %q path %s", volID, stagingTargetPath)
 		return fscrypt.Unlock(ctx, volOptions.Encryption, stagingTargetPath, string(volID))
```
Madhu-1: dont we need to release lock once the fscrypt.Unlock is done?

internal/cephfs/nodeserver.go
```diff
@@ -127,13 +128,47 @@ func maybeUnlockFileEncryption(
...
-	return nil
+		_, err = ioctx.Unlock(string(volID), lockName, lockCookie)
+
+		if err != nil {
+			log.ErrorLog(ctx, "failed to release following lock, this will lead to orphan lock %s: %v",
+				lockName, err)
+		}
+	}()
+
+	return fmt.Errorf("There is already one file system with name %s", string(volID))
```
Madhu-1: Need to update the error message

internal/cephfs/nodeserver.go
Sunnatillo: Done

internal/cephfs/nodeserver.go
Sunnatillo: Hey. Thank you for the review. I have addressed most of your comments.

internal/cephfs/nodeserver.go
Sunnatillo: Should we add flag to make it configurable?

nixpanic: My preference is to have a high timeout (150 seconds like it is now should be plenty). Adding a configuration option if really needed can be done later. But hopefully this is sufficient, I expect users that need to have a higher timeout to have more severe problems than failures to setup encryption.

Madhu-1: works for me as well

Madhu-1:
@Sunnatillo can you please check the CI failures?

internal/cephfs/nodeserver.go
```diff
@@ -127,10 +133,55 @@ func maybeUnlockFileEncryption(
...
+	log.DebugLog(ctx, "Creating lock for the following volume ID %s", volID)
+
+	ioctx, err := volOptions.GetConnection().GetIoctx(volOptions.MetadataPool)
+	if err != nil {
+		log.ErrorLog(ctx, "Failed to create ioctx: %s", err)
+	}
+	defer ioctx.Destroy()
+
+	res, err := ioctx.LockExclusive(volOptions.VolID, lockName, lockCookie, lockDesc, lockDuration, &flags)
+	switch res {
```
Madhu-1: can we make use of https://pkg.go.dev/syscall? and also need to return errors instead of returning nil until the fscrypt.Unlock is done or how to ensure that unlock is successful?

Madhu-1: i think we might need to make use of metadata to ensure it's in progress or is done along with the lock.

internal/cephfs/nodeserver.go
```diff
@@ -127,10 +133,55 @@ func maybeUnlockFileEncryption(
...
+		log.DebugLog(ctx, "Lock successfully created for volume ID %s", volID)
+	case -C.EBUSY:
+		log.DebugLog(ctx, "Lock is already held by another client and cookie pair")
+	case -C.EEXIST:
+		log.DebugLog(ctx, "Lock is already held by the same client and cookie pair")
+	default:
+		return fmt.Errorf("Failed to lock volume ID %v: %w", volID, err)
+	}
+
+	if res == 0 {
```
nixpanic: instead of positive check, use
```go
if res != 0 {
    return err // can this happen? err == nil
}

// Success! Continue with the function :)
```

internal/cephfs/nodeserver.go
```diff
@@ -127,10 +133,55 @@ func maybeUnlockFileEncryption(
...
+		log.DebugLog(ctx, "cephfs: unlocking fscrypt on volume %q path %s", volID, stagingTargetPath)
+		err = fscrypt.Unlock(ctx, volOptions.Encryption, stagingTargetPath, string(volID))
+		if err != nil {
+			return err
+		} else {
```
nixpanic: no need for the `else {`, just continue with the function and loose one indention level (makes it easier to read and understand the code).

internal/cephfs/nodeserver.go
nixpanic: @Sunnatillo please check the above to improve readability of the code.

internal/cephfs/nodeserver.go
nixpanic: if creation of the ioctx fails, an error should be returned

internal/cephfs/nodeserver.go
nixpanic: I do not see how the syscall package can be used here.

internal/cephfs/nodeserver.go
Sunnatillo: Thank you for the review @nixpanic. 

`res` can be equal to `0`, `C.EBUSY` and `-C.EEXIST`. Only in the case of  `res == 0` it should try to unlock fscrypt on volume.  In my opinion `if res != 0 ` is not suitable here.

internal/cephfs/nodeserver.go
Sunnatillo: Thank you for the reviews @Madhu-1. 

Sorry, I might misunderstood you.  `fscrypt.Unlock` returns an error if it is not successful. 

Could you please elaborate more regarding https://pkg.go.dev/syscall?

Madhu-1: i mean like below

```go
if errors.Is(err, syscall.EEXIST) {
```

if we do like above we don't need to import c to this file

internal/cephfs/nodeserver.go
nixpanic: @Sunnatillo , the thing I wanted to suggest is to return early from the function when `res != 0`. In the case where `res == 0` the function can continue (and you can drop one tab indention). I think this makes the code a little easier to understand.

internal/cephfs/nodeserver.go
nixpanic: @Sunnatillo, this is the last error that still needs to be handled. If getting the ioctx fails, there is no need to continue.

internal/cephfs/nodeserver.go
```diff
@@ -127,10 +129,54 @@ func maybeUnlockFileEncryption(
...
+	if res != 0 {
+		switch res {
+		case -int(syscall.EBUSY):
+			log.DebugLog(ctx, "Lock is already held by another client and cookie pair")
+		case -int(syscall.EEXIST):
+			log.DebugLog(ctx, "Lock is already held by the same client and cookie pair")
+		default:
+			return fmt.Errorf("Failed to lock volume ID %v: %w", volID, err)
+		}
+		return err
```
nixpanic: **golangci-lint** wants an empty line before this return statement :disappointed:

internal/cephfs/nodeserver.go
```diff
@@ -127,10 +129,54 @@ func maybeUnlockFileEncryption(
...
+	log.DebugLog(ctx, "Creating lock for the following volume ID %s", volID)
+
+	ioctx, err := volOptions.GetConnection().GetIoctx(volOptions.MetadataPool)
+	if err != nil {
+		log.ErrorLog(ctx, "Failed to create ioctx: %s", err)
+	}
+	defer ioctx.Destroy()
+
+	res, err := ioctx.LockExclusive(volOptions.VolID, lockName, lockCookie, lockDesc, lockDuration, &flags)
+	if res != 0 {
```
Madhu-1, line 155: https://github.com/ceph/go-ceph/blob/1046b034a1f618f67acd3c6523482917e27c7113/rados/ioctx.go#L545-L558 Looking at this one, we need to return an error in case of EBUSY and EEXIST at line 161 it return nil, we need to have return in those cases as well, @nixpanic can you please check this one?

nixpanic: EBUSY and EEXIST are expected return values, not an error while calling the function. I do not think go-ceph should return an error in those cases.

Madhu-1: sorry i meant not in the go-ceph from here cephcsi should return an error so that if the lock is held on one node we should keep retrying on another node until it's done, or we just return nil (which is happening here) assuming that the other client will unlock fscrypt?

Makefile
```diff
@@ -162,7 +162,7 @@ commitlint:
 .PHONY: cephcsi
 cephcsi: check-env
 	if [ ! -d ./vendor ]; then (go mod tidy && go mod vendor); fi
-	GOOS=linux go build $(GO_TAGS) -mod vendor -a -ldflags '$(LDFLAGS)' -o _output/cephcsi ./cmd/
+	GOOS=linux go build $(GO_TAGS) -buildvcs=false -mod vendor -a -ldflags '$(LDFLAGS)' -o _output/cephcsi ./cmd/
```
nixpanic: this now got added, is that intentional?

Sunnatillo: My bad.  it is not intentional

Sunnatillo: removed now.

internal/cephfs/nodeserver.go
nixpanic: ah, yes, now I see what you mean. If EBUSY/EEXISTS is returned from `LockExclusive()`, `err` will be `nil`. That `err` is then returned on line 162, and the caller of this function assumes all is OK (because `err` is `nil`). This is not correct, and instead of returning `err` on line 162, a new error should be returned. Ideally each `case` in the below switch sets `err` to a new unique value (mentioning something similar as what is in the debug log).

internal/cephfs/nodeserver.go
```diff
@@ -127,15 +129,70 @@ func maybeUnlockFileEncryption(
...
+	lockDuration := 150 * time.Second
+	// Generate a consistent lock cookie for the client using hostname and process ID
+	lockCookie := generateLockCookie()
+	var flags byte = 0
+
+	log.DebugLog(ctx, "Creating lock for the following volume ID %s", volID)
+
+	ioctx, err := volOptions.GetConnection().GetIoctx(volOptions.MetadataPool)
+	if err != nil {
+		log.ErrorLog(ctx, "Failed to create ioctx: %s", err)
```
Madhu-1, line 148: return err is missing here, everything else LGTM

Sunnatillo: added

nixpanic: Thanks for your persistence!

nixpanic (dismissed):

Madhu-1 (dismissed):
Thank you @Sunnatillo 🎉

nixpanic:
@mergifyio rebase

Sunnatillo:
Thank you for reviews @nixpanic, @Madhu-1. 

I rebased to latest devel branch.

nixpanic (approved):

iPraveenParihar:
@nixpanic, I see `fscrypt.Unlock` call for RBD fscrypt. Do we need to do the same here as well?
do we need to
https://github.com/ceph/ceph-csi/blob/e71a95fece24de8de8f1de51201a7314af4f3480/internal/rbd/nodeserver.go#L467-L474

nixpanic:
> @nixpanic, I see `fscrypt.Unlock` call for RBD fscrypt. Do we need to do the same here as well?

No, for RBD that is not needed. RBD with a filesystem is never attached to more than one workernode at the same time, so initialization/unlocking of the filesystem will always be on a single node. It is different with ReadWriteMany RBD volumes in block-mode with encryption. The LUKS calls would benefit from a similar locking.

iPraveenParihar (approved):
Thanks!

internal/cephfs/nodeserver.go
```diff
@@ -127,15 +129,72 @@ func maybeUnlockFileEncryption(
...
+			return fmt.Errorf("Lock is already held by the same client and cookie pair for %v volume", volID)
+		default:
+			return fmt.Errorf("Failed to lock volume ID %v: %w", volID, err)
+		}
+	}
+	log.DebugLog(ctx, "Lock successfully created for volume ID %s", volID)
+
+	log.DebugLog(ctx, "cephfs: unlocking fscrypt on volume %q path %s", volID, stagingTargetPath)
+	err = fscrypt.Unlock(ctx, volOptions.Encryption, stagingTargetPath, string(volID))
+	if err != nil {
```
NymanRobin, line 169: Won’t this leave a hanging lock in case the fscrypt fails to unlock the filesystem?

I think the ”ceph unlock functionality” could be moved to a separate function and this function pushed to the defer stack after the exclusiveLock is successful?

NymanRobin (commented):
Thanks a lot for fixing this 🎉

But I have a slight concern in the scenario when facrypt.Unlock fails. What do you think?

internal/cephfs/nodeserver.go
Sunnatillo: What about `lockDuration` variable on the `LockExclusive` function? When duration expires it will automatically unlock?

NymanRobin: Yes, this I understand. But it cause unnecessary delays in my opinion that could easily be avoided. My point is that usually this is operation that would take seconds start to take minutes due to one failure

Sunnatillo: @nixpanic, @Madhu-1 WDYT?

Madhu-1: Yes make sense to avoid unwanted delay in this case,As suggested `ioctx.Unlock` can be called defer right after the successful Lock.

NymanRobin: I created an issue for this so it is not forgotten: https://github.com/ceph/ceph-csi/issues/4715

z2000l:
Is there a config to use the lock? We built an image from devel branch (https://github.com/ceph/ceph-csi/tree/devel) and tried it out. However pods failed to come up and we got the following error:

Warning  FailedMount             15s (x6 over 31s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-4642810a-7f00-4ae0-8583-ddbcee6ed314" : rpc error: code = Internal desc = Failed to lock volume ID 0001-0009-rook-ceph-00000000000000ad-d71d2949-db34-434e-8008-5497dad5c70b: rados: ret=-1, Operation not permitted

NymanRobin:
I have tested this and encountered the same error as @z2000l. @Sunnatillo, could you please document any additional settings required or anything we might be overlooking?

internal/cephfs/nodeserver.go
```diff
@@ -127,15 +129,72 @@ func maybeUnlockFileEncryption(
...
+	ioctx, err := volOptions.GetConnection().GetIoctx(volOptions.MetadataPool)
+	if err != nil {
+		log.ErrorLog(ctx, "Failed to create ioctx: %s", err)
 
-		return fscrypt.Unlock(ctx, volOptions.Encryption, stagingTargetPath, string(volID))
+		return err
+	}
+	defer ioctx.Destroy()
+
+	res, err := ioctx.LockExclusive(volOptions.VolID, lockName, lockCookie, lockDesc, lockDuration, &flags)
```
NymanRobin, line 154: ~~When reading the code in go-ceph it seems to be that the object that we want to lock should exist and this won't create a new object? Are we sure there exist an object in rados corresponding to `VolID`, my initial assumption is no but I am not 100% confident, without further investigation~~
```
// LockExclusive takes an exclusive lock on an object.
func (ioctx *IOContext) LockExclusive(oid, name, cookie, desc string, duration time.Duration, flags *byte) (int, error) {
```
Edit: The object seems not to be needed

NymanRobin: Continuing my dig here, I added the following logging before the LockExclusive

```
_, err = ioctx.Stat(volOptions.VolID)
if err != nil {
  log.ErrorLog(ctx, "The error when object is not here: %s", err)
  log.ErrorLog(ctx, "Object %s does not exist", volOptions.VolID)
}
```

I get these logs:
```
E0724 09:03:43.715019       1 nodeserver.go:156] ID: 142 Req-ID: 0001-0009-rook-ceph-0000000000000001-489ccf33-13a5-40fc-8460-7dd866bc44de The error when object is not here: rados: ret=-2, No such file or directory
E0724 09:03:43.715051       1 nodeserver.go:157] ID: 142 Req-ID: 0001-0009-rook-ceph-0000000000000001-489ccf33-13a5-40fc-8460-7dd866bc44de Object csi-vol-489ccf33-13a5-40fc-8460-7dd866bc44de does not exist
E0724 09:03:43.715477       1 utils.go:240] ID: 142 Req-ID: 0001-0009-rook-ceph-0000000000000001-489ccf33-13a5-40fc-8460-7dd866bc44de GRPC error: rpc error: code = Internal desc = Failed to lock volume ID 0001-0009-rook-ceph-0000000000000001-489ccf33-13a5-40fc-8460-7dd866bc44de: rados: ret=-1, Operation not permitted
```
This makes me highly suspicious... I however wonder why in Ceph they do not return -2, in case a lock is tried to be acquired on a object that does not exist? Maybe they have not thought of this scenario and it could be reported?I guess next step in this journey to the core would be to create the object first and see if the lock can the be acquired :smile: 

Edit: Actual problem is comments below, the object seems to not be needed

NymanRobin: Now I know at least why the -1 comes, and it is correct that the permissions are not there!
After adding the following to the authentication I can acquire the lock:
```
ceph auth caps client.csi-cephfs-node \
  mon 'allow *' \
  mgr 'allow *' \
  osd 'allow *' \
  mds 'allow *'
```

But I think what is actually needed is the following:
`ceph auth caps client.csi-cephfs-node rados 'allow rw'`

NymanRobin: Okay to conclude this it seems the lock can be created without the object existing so that is all good and my ramblings from above can be ignored!

Except some more permission for the client.csi-cephfs-node  is needed to be able to create the lock

However it is extremely slow currently for me, bringing up five pods with 3 replicas and after 5 minutes I have only one replica of each. The error reported is the following:

```
  Warning  FailedMount             3m32s (x2 over 5m47s)  kubelet                  Unable to attach or mount volumes: unmounted volumes=[my-volume], unattached volumes=[my-volume kube-api-access-8cmvd]: timed out waiting for the condition
  Warning  FailedMount             81s (x11 over 7m35s)   kubelet                  MountVolume.MountDevice failed for volume "pvc-6afa784b-dc4f-44fb-bed9-8c6fe985c7e8" : rpc error: code = Internal desc = Lock is already held by another client and cookie pair for 0001-0009-rook-ceph-0000000000000001-c7f5d8f9-f9ca-4af8-8457-7ffd1bfb4437 volume
  Warning  FailedMount             74s                    kubelet                  Unable to attach or mount volumes: unmounted volumes=[my-volume], unattached volumes=[kube-api-access-8cmvd my-volume]: timed out waiting for the condition

```

This is the status of the pods after 8m46s:
```
$ kubectl get pods -n default 
NAME                            READY   STATUS              RESTARTS   AGE
deploy-pod-1-7f7789fd5f-92x4f   1/1     Running             0          8m46s
deploy-pod-1-7f7789fd5f-99qmp   1/1     Running             0          8m46s
deploy-pod-1-7f7789fd5f-dnr79   0/1     ContainerCreating   0          8m46s
deploy-pod-2-58c654874c-j2mvp   0/1     ContainerCreating   0          8m46s
deploy-pod-2-58c654874c-jh2wp   1/1     Running             0          8m46s
deploy-pod-2-58c654874c-svjws   1/1     Running             0          8m46s
deploy-pod-3-6674fc79f-rbtk2    1/1     Running             0          8m46s
deploy-pod-3-6674fc79f-w452p    1/1     Running             0          8m46s
deploy-pod-3-6674fc79f-w8wqb    0/1     ContainerCreating   0          8m46s
deploy-pod-4-6c958fb787-cqxlb   1/1     Running             0          8m46s
deploy-pod-4-6c958fb787-kvgbl   0/1     ContainerCreating   0          8m46s
deploy-pod-4-6c958fb787-q794c   1/1     Running             0          8m46s
deploy-pod-5-86b9bb47-5vmrq     0/1     ContainerCreating   0          8m45s
deploy-pod-5-86b9bb47-dd6v7     1/1     Running             0          8m45s
deploy-pod-5-86b9bb47-g5jvg     0/1     ContainerCreating   0          8m45s
```

After a whopping 13 minutes all the pods were finally running!

Should we actually poll the lock when trying to acquire, I think that might make it faster

Sunnatillo: Thank you @NymanRobin for testing. 
I also tested. it is slow. 
It took me 8 minutes to 6 replicas to be ready. 
We should definitely improve it.

NymanRobin:
I can write up a ticket for the performance and for the permissions!

internal/cephfs/nodeserver.go
NymanRobin: I can write up a ticket for the performance and for the permissions!

PR #4731, opened 2024-07-25
cephfs: correct the id for cephfs locks

### LINKED ISSUE #4727: Improve performance in encrypted cephfs provisioning
#  The bug #

Currently provisioning of encrypted cephfs filesystems are slow, takes several minutes to provision a deployment with a attached filesystem. 

# Potential solution #
I think performance could be increased by polling the lock state as currently we let the reconcile loop do all the heavy lifting. So we will always need to wait at least as many loops as there is replicas. However the lock is very short-lived so we could simply wait in a loop and retry acquiring it with some interval for x-amount of times before giving up.

Also should observe if there is any inefficiencies in releasing the lock as these always amount up to a huge amount of time in case of a hanging lock. Not sure if there is any though

Additional the lock is also needed only on the initial setup so we could keep track of locked filesystems in a omap and check if it is already locked, but this brings quite a lot of complexity and not sure what the actual speed improvement would be

NymanRobin:
There was a discrepancy between the objectId when creating the lock and when releasing the lock this causes every lock to hang.

Fixes: #4727 

This removes the performance concerns discussed in this [review thread](https://github.com/ceph/ceph-csi/pull/4697#discussion_r1690157131), after this change the pods with this encrypted filesystems comes up in seconds as expected.

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -151,7 +151,7 @@ func maybeUnlockFileEncryption(
 	}
 	defer ioctx.Destroy()
 
-	res, err := ioctx.LockExclusive(volOptions.VolID, lockName, lockCookie, lockDesc, lockDuration, &flags)
+	res, err := ioctx.LockExclusive(string(volID), lockName, lockCookie, lockDesc, lockDuration, &flags)
 	if res != 0 {
 		switch res {
 		case -int(syscall.EBUSY):


Sunnatillo (commented):
LGTM

Sunnatillo:
Thank you for taking care of it.

iPraveenParihar (dismissed):
Thanks!

nixpanic (dismissed):

nixpanic:
@mergifyio rebase

nixpanic (approved):

yati1998 (approved):

PR #4739, opened 2024-07-29
[WIP] Implement support for volume group mirroring

Madhu-1:
This PR does the required changes so that the volume group can also implement the Mirror interface and we do need to make many changes in the replication code to work differently for the mirroring of an image or a mirroring of a group.

Note:- The last commit to add use go-ceph is commented out, please don't review it, Will remove that code and make changes to the mirror implementation of the group. except that other commits can be reviewed.

Depends-on: ceph/go-ceph#1010

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -29,11 +29,13 @@ import (
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
 	"github.com/ceph/ceph-csi/internal/rbd"
 	corerbd "github.com/ceph/ceph-csi/internal/rbd"
+	rbd_group "github.com/ceph/ceph-csi/internal/rbd/group"
 	"github.com/ceph/ceph-csi/internal/rbd/types"
 	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/log"
 
 	librbd "github.com/ceph/go-ceph/rbd"
+
 	"github.com/ceph/go-ceph/rbd/admin"
 	"github.com/csi-addons/spec/lib/go/replication"
 	"google.golang.org/grpc"
@@ -785,13 +787,14 @@ func getGRPCError(err error) error {
 	}
 
 	errorStatusMap := map[error]codes.Code{
-		corerbd.ErrImageNotFound:      codes.NotFound,
-		util.ErrPoolNotFound:          codes.NotFound,
-		corerbd.ErrInvalidArgument:    codes.InvalidArgument,
-		corerbd.ErrFlattenInProgress:  codes.Aborted,
-		corerbd.ErrAborted:            codes.Aborted,
-		corerbd.ErrFailedPrecondition: codes.FailedPrecondition,
-		corerbd.ErrUnavailable:        codes.Unavailable,
+		corerbd.ErrImageNotFound:         codes.NotFound,
+		util.ErrPoolNotFound:             codes.NotFound,
+		corerbd.ErrInvalidArgument:       codes.InvalidArgument,
+		corerbd.ErrFlattenInProgress:     codes.Aborted,
+		corerbd.ErrAborted:               codes.Aborted,
+		corerbd.ErrFailedPrecondition:    codes.FailedPrecondition,
+		corerbd.ErrUnavailable:           codes.Unavailable,
+		rbd_group.ErrRBDGroupUnAvailable: codes.Unavailable,
 	}
 
 	for e, code := range errorStatusMap {
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -257,9 +257,9 @@ func validateSchedulingInterval(interval string) error {
 func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 	req *replication.EnableVolumeReplicationRequest,
 ) (*replication.EnableVolumeReplicationResponse, error) {
-	volumeID := csicommon.GetIDFromReplication(req)
-	if volumeID == "" {
-		return nil, status.Error(codes.InvalidArgument, "empty volume ID in request")
+	reqID := csicommon.GetIDFromReplication(req)
+	if reqID == "" {
+		return nil, status.Error(codes.InvalidArgument, "empty ID in request")
 	}
 	cr, err := util.NewUserCredentials(req.GetSecrets())
 	if err != nil {
@@ -272,24 +272,23 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 		return nil, err
 	}
 
-	if acquired := rs.VolumeLocks.TryAcquire(volumeID); !acquired {
-		log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volumeID)
+	if acquired := rs.VolumeLocks.TryAcquire(reqID); !acquired {
+		log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, reqID)
 
-		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volumeID)
+		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, reqID)
 	}
-	defer rs.VolumeLocks.Release(volumeID)
+	defer rs.VolumeLocks.Release(reqID)
 
 	mgr := rbd.NewManager(rs.driverInstance, req.GetParameters(), req.GetSecrets())
 	defer mgr.Destroy(ctx)
 
-	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
+	volumes, mirror, err := mgr.GetMirrorSource(ctx, reqID, req.GetReplicationSource())
 	if err != nil {
+		log.ErrorLog(ctx, "failed to get mirror source with id %q: %v", reqID, err)
+
 		return nil, getGRPCError(err)
 	}
-	mirror, err := rbdVol.ToMirror()
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
-	}
+	defer destoryVolumes(ctx, volumes)
 
 	// extract the mirroring mode
 	mirroringMode, err := getMirroringMode(ctx, req.GetParameters())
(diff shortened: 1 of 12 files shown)


internal/rbd/manager.go
```diff
@@ -124,7 +124,19 @@ func (mgr *rbdManager) GetVolumeByID(ctx context.Context, id string) (types.Volu
 
 	volume, err := GenVolFromVolID(ctx, id, creds, mgr.secrets)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get volume from id %q: %w", id, err)
+		switch {
```
nixpanic: this commit (and maybe others) can be put in a separate PR for earlier merging and keeping follow-ups smaller if you like

internal/rbd/manager.go
```diff
@@ -295,3 +295,43 @@ func (mgr *rbdManager) DeleteVolumeGroup(ctx context.Context, vg types.VolumeGro
...
+
+func (mgr *rbdManager) GetMirrorSource(ctx context.Context, volumeID string,
+	rep *replication.ReplicationSource,
+) ([]types.Volume, types.Mirror, error) {
+	// Backward compatibility: if rep is nil, we assume that the sidecar is still old and
+	// setting only volumeId not the replication source.
+	if rep == nil {
+		rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
+		if err != nil {
+			return nil, nil, err
```
nixpanic: consider adding a description to the returned `err`

internal/csi-addons/rbd/volumegroup.go
```diff
@@ -117,6 +117,21 @@ func (vs *VolumeGroupServer) CreateVolumeGroup(
 
 	log.DebugLog(ctx, "VolumeGroup %q has been created: %+v", req.GetName(), vg)
 
+	// extract the flatten mode
+	flattenMode, err := getFlattenMode(ctx, req.GetParameters())
+	if err != nil {
+		return nil, err
```
nixpanic: This needs to return a grpc error. `getFlattenMode` is not appropriate to do so.

internal/csi-addons/rbd/volumegroup.go
```diff
@@ -343,17 +343,40 @@ func (vs *VolumeGroupServer) ModifyVolumeGroupMembership(
...
 				err)
 		}
 
+		//nolint:gocritic // need to call .Destroy() for all volumes
+		defer vol.Destroy(ctx)
+		volumes[i] = vol
+	}
+
+	// extract the flatten mode
+	flattenMode, err := getFlattenMode(ctx, req.GetParameters())
```
nixpanic: same here, `getFlattenMode()` probably does not return a grpc error.

internal/csi-addons/rbd/volumegroup.go
```diff
@@ -343,17 +343,40 @@ func (vs *VolumeGroupServer) ModifyVolumeGroupMembership(
...
+	if err != nil {
+		return nil, err
+	}
+	// Flatten the image if the flatten mode is set to FlattenModeForce
+	// before adding it to the volume group.
+	for _, vol := range volumes {
+		err = vol.HandleParentImageExistence(ctx, flattenMode)
+		if err != nil {
+			return nil, getGRPCError(err)
```
nixpanic: consider adding more details to the error, include at least the volume name/id

internal/csi-addons/rbd/replication.go
```diff
@@ -304,7 +304,7 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 		return nil, err
 	}
 
-	info, err := mirror.GetMirroringInfo()
+	info, err := mirror.GetMirroringInfo(ctx)
```
nixpanic: can be in an other PR too

internal/rbd/group/volume_group.go
```diff
@@ -151,18 +151,17 @@ func GetVolumeGroup(
 	return vg, nil
 }
 
-// String makes it easy to include the volumeGroup object in log and error
-// messages.
+// String returns the image-spec (pool/{namespace}/{name}) format of the group.
```
nixpanic: can be in an other PR too

internal/rbd/group/volume_group.go
```diff
@@ -465,3 +470,300 @@ func (vg *volumeGroup) RemoveVolume(ctx context.Context, vol types.Volume) error
...
+	return nil
+}
+
+func (vg *volumeGroup) ToMirror() (types.Mirror, error) {
+	return vg, nil
+}
+
+// GroupStatus is a wrapper around librbd.MirrorGroupInfo that contains the
+// group mirror info.
+type GroupStatus struct {
```
nixpanic: why not just do it like
```go
type GroupStatus *librbd.MirrorGroupInfo
```

internal/csi-addons/rbd/volumegroup.go
Madhu-1: getFlattenMode is returning grpc error same is already used in replication.

internal/rbd/manager.go
Madhu-1: Let me see if i can do that 👍🏻

internal/csi-addons/rbd/volumegroup.go
Madhu-1: let me check that if HandleParentImageExistence already including it

internal/rbd/group/volume_group.go
Madhu-1: >Methods defined on the original type are not available to the new type unless you explicitly defined

if we create a type i need to implement duplication methods of MirrorGroupInfo here which is again not much useful, lets keep it this way which allow us to use methods of librbd struct without any problem

Madhu-1:
@nixpanic This is ready for one more round of review :)

internal/rbd/types/manager.go
```diff
@@ -47,4 +49,8 @@ type Manager interface {
 	// DeleteVolumeGroup removes VolumeGroup from the backend storage and
 	// any details from the journal.
 	DeleteVolumeGroup(ctx context.Context, vg VolumeGroup) error
+
+	// GetMirrorSource returns the source of the mirror for the given volume or group.
+	GetMirrorSource(ctx context.Context, volumeID string,
+		rep *replication.ReplicationSource) ([]Volume, Mirror, error)
```
nixpanic, line 79: it is not clear to me why `[]Volume` is returned. Should that not be either a single Volume or a VolumeGroup?

nixpanic: would it make sense to have a `Mirror.GetSource()` function so that the API is easier to understand?

internal/rbd/group/volume_group.go
```diff
@@ -469,3 +474,296 @@ func (vg *volumeGroup) ListVolumes(ctx context.Context) ([]types.Volume, error)
 func (vg *volumeGroup) ToMirror() (types.Mirror, error) {
 	return vg, nil
 }
+
+func (vg *volumeGroup) EnableMirroring(ctx context.Context, mode librbd.ImageMirrorMode) error {
```
nixpanic: I wonder if it is worth to simplify things a little by placing this in a separate file `group_mirror.go`?

```go
type volumeGroupMirror *volumeGroup // or make it a struct with extra members?

func (vgm *volumeGroupMirror) EnableMirroring(ctx context.Context, mode librbd.ImageMirrorMode) error {
    ...
```

This would keep the VolumeGroup interface smal(ler). I am considering a similar approach for VolumeGroupSnapshot.

internal/rbd/group/volume_group.go
```diff
@@ -469,3 +474,296 @@ func (vg *volumeGroup) ListVolumes(ctx context.Context) ([]types.Volume, error)
...
+	if err != nil {
+		return nil, err
+	}
+
+	info, err := librbd.GetMirrorGroupInfo(ioctx, name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get volume group mirroring info %q: %w", vg, err)
+	}
+
+	return GroupInfo{MirrorGroupInfo: info}, nil
```
nixpanic: you could also do

```go
type GroupInfo *librbd.MirrorGroupInfo
```

and cast it here

```go
return GroupInfo(info)
```

That looks a little simpler to me.

internal/rbd/group/volume_group.go
```diff
@@ -469,3 +474,296 @@ func (vg *volumeGroup) ListVolumes(ctx context.Context) ([]types.Volume, error)
...
+}
+
+// GlobalGroupMirrorStatus is a wrapper around librbd.GlobalGroupMirrorImageStatus that contains the
+// global mirror group status.
+type GlobalMirrorGroupStatus struct {
+	*librbd.GlobalMirrorGroupStatus
+}
+
+func (status GlobalMirrorGroupStatus) GetState() string {
+	return status.GlobalMirrorGroupStatus.Info.State.String()
```
nixpanic: could any of the members be `nil`?

internal/rbd/group/volume_group.go
```diff
@@ -469,3 +474,296 @@ func (vg *volumeGroup) ListVolumes(ctx context.Context) ([]types.Volume, error)
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GroupInfo is a wrapper around librbd.MirrorGroupInfo that contains the
+// group mirror info.
+type GroupInfo struct {
```
nixpanic: If you want to use a struct, consider exposing an interface instead of the struct. That encourages a cleaner API towards consumers.

internal/csi-addons/rbd/replication.go
```diff
@@ -280,15 +280,12 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 	mgr := rbd.NewManager(rs.csiID, req.GetParameters(), req.GetSecrets())
 	defer mgr.Destroy(ctx)
 
-	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
+	volumes, mirror, err := mgr.GetMirrorSource(ctx, volumeID, req.GetReplicationSource())
```
Rakshith-R: let's call `volumeID` something else like reqID or replicationID.
It can then be either referring to a single volume or group (or later on snapshot).

internal/rbd/types/manager.go
Rakshith-R: Let's put everything related to processing volumes within functions in mirror type ?

Madhu-1: >would it make sense to have a Mirror.GetSource() function so that the API is easier to understand?

@nixpanic I put it with the Manager because that is the one who is generating the volume and group from an ID, moving this to Mirroring

>it is not clear to me why []Volume is returned. Should that not be either a single Volume or a VolumeGroup?

This is required because we need to work on all the volumes if its a group, i am trying to avoid type casting and doing the ListVolumes and generating the volumes again that contains the connections, let me know if you have any suggestions

> Let's put everything related to processing volumes within functions in mirror type ?

I mostly added all the mirroring specific things to mirror type but there are again generic one as well that the reason i left it in volume type as it makes more sense.

internal/rbd/group/volume_group.go
Madhu-1: Yes but again this will have problem like https://github.com/ceph/ceph-csi/pull/4739#discussion_r1697232084 and adding new field to GroupInfo need more refractoring, IMO this should be okay but the struct name should not be exported(taken care of it)

internal/rbd/group/volume_group.go
Madhu-1: Not possible as the members are not pointers

internal/rbd/manager.go
```diff
@@ -294,3 +295,98 @@ func (mgr *rbdManager) DeleteVolumeGroup(ctx context.Context, vg types.VolumeGro
...
+		var mir types.Mirror
+		mir, err = rbdVol.ToMirror()
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to convert volume %s to mirror: %w", rbdVol, err)
+		}
+
+		return []types.Volume{rbdVol}, mir, nil
+	}
+	switch {
+	case rep.GetVolume() != nil:
```
Rakshith-R:
```suggestion
	switch {
	// Backward compatibility: if rep is nil, we assume that the sidecar is still old and
	// set only volumeId, else rely on replicationSource.
	case rep == nil || rep.GetVolume() != nil:
```

internal/rbd/types/manager.go
Rakshith-R: :+1: 
We can check on volumes and refactor it later on.
Everything LGTM
Only concern now is ceph's API stability and go-ceph pr going in.

https://github.com/ceph/go-ceph/pull/1010#issuecomment-2223542910
https://github.com/ceph/ceph/pull/53793#discussion_r1670404359

PR #4743, opened 2024-07-30
RBD: improvments to existing code to better align with volumegroup

Madhu-1:
This PR contains the commits taken out of https://github.com/ceph/ceph-csi/pull/4739 to make the other PR smaller. This contains lot of improvement which is going to help for future implementation and this also takes care of handling the flattening before adding the image to the group.

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -282,16 +282,7 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 
 	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
 	if err != nil {
-		switch {
-		case errors.Is(err, corerbd.ErrImageNotFound):
-			err = status.Errorf(codes.NotFound, "volume %s not found", volumeID)
-		case errors.Is(err, util.ErrPoolNotFound):
-			err = status.Errorf(codes.NotFound, "pool %s not found for %s", rbdVol.GetPoolName(), volumeID)
-		default:
-			err = status.Errorf(codes.Internal, err.Error())
-		}
-
-		return nil, err
+		return nil, getGRPCError(err)
 	}
 	mirror, err := rbdVol.ToMirror()
 	if err != nil {
@@ -361,16 +352,7 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 
 	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
 	if err != nil {
-		switch {
-		case errors.Is(err, corerbd.ErrImageNotFound):
-			err = status.Errorf(codes.NotFound, "volume %s not found", volumeID)
-		case errors.Is(err, util.ErrPoolNotFound):
-			err = status.Errorf(codes.NotFound, "pool %s not found for %s", rbdVol.GetPoolName(), volumeID)
-		default:
-			err = status.Errorf(codes.Internal, err.Error())
-		}
-
-		return nil, err
+		return nil, getGRPCError(err)
 	}
 	mirror, err := rbdVol.ToMirror()
 	if err != nil {
@@ -438,16 +420,7 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 
 	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
 	if err != nil {
-		switch {
-		case errors.Is(err, corerbd.ErrImageNotFound):
-			err = status.Errorf(codes.NotFound, "volume %s not found", volumeID)
-		case errors.Is(err, util.ErrPoolNotFound):
-			err = status.Errorf(codes.NotFound, "pool %s not found for %s", rbdVol.GetPoolName(), volumeID)
-		default:
-			err = status.Errorf(codes.Internal, err.Error())
-		}
-
-		return nil, err
+		return nil, getGRPCError(err)
 	}
 	mirror, err := rbdVol.ToMirror()
 	if err != nil {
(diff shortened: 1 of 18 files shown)


internal/rbd/mirror.go
```diff
@@ -61,7 +61,7 @@ func (rv *rbdVolume) HandleParentImageExistence(
 
 	parent, err := rv.getParent()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get parent image %s: %w", rv, err)
```
Rakshith-R:
```suggestion
		return fmt.Errorf("failed to get parent of image %s: %w", rv, err)
```

internal/csi-addons/rbd/volumegroup.go
```diff
@@ -341,17 +341,40 @@ func (vs *VolumeGroupServer) ModifyVolumeGroupMembership(
...
+		if err != nil {
 			return nil, status.Errorf(
 				codes.NotFound,
 				"failed to find a volume with CSI ID %q: %v",
 				id,
 				err)
 		}
 
+		//nolint:gocritic // need to call .Destroy() for all volumes
+		defer vol.Destroy(ctx)
```
Rakshith-R: This deferred function is called when the outer function is exiting right ? 
not when the loop ends or all iterations are complete ? 
(I maybe mistaken)

internal/rbd/group/volume_group.go
```diff
@@ -151,18 +151,17 @@ func GetVolumeGroup(
 	return vg, nil
 }
 
-// String makes it easy to include the volumeGroup object in log and error
-// messages.
+// String returns the image-spec (pool/{namespace}/{name}) format of the group.
 func (vg *volumeGroup) String() string {
```
Rakshith-R, line 155: tagging @nixpanic to verify is vg.id is not used anywhere from the Stringer call.

Rakshith-R (commented):
tiny nits

internal/csi-addons/rbd/volumegroup.go
Madhu-1: Yes, you are correct, this gets stacked up as it's called in for loop. removed it

internal/csi-addons/rbd/volumegroup.go
```diff
@@ -117,6 +117,19 @@ func (vs *VolumeGroupServer) CreateVolumeGroup(
...
+	flattenMode, err := getFlattenMode(ctx, req.GetParameters())
+	if err != nil {
+		return nil, err
+	}
+	// Flatten the image if the flatten mode is set to FlattenModeForce
+	// before adding it to the volume group.
+	for _, v := range volumes {
+		err = v.HandleParentImageExistence(ctx, flattenMode)
+		if err != nil {
+			return nil, getGRPCError(err)
```
nixpanic, line 135: include a bit more details in the error maybe? Mention the VolumeGroup and Volume?

internal/csi-addons/rbd/volumegroup.go
nixpanic: an other option, maybe cleaner?

```go
volumes := make([]types.Volume, len(toAdd))
defer func() {
    // need to call .Destroy() for all volumes
    for _, vol := range volume {
        vol.Destroy(ctx)
    }
}()

for i, id := range toAdd {
    ...
```

internal/csi-addons/rbd/replication.go
```diff
@@ -284,9 +284,9 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 	if err != nil {
 		switch {
 		case errors.Is(err, corerbd.ErrImageNotFound):
-			err = status.Errorf(codes.NotFound, "volume %s not found", volumeID)
+			err = status.Errorf(codes.NotFound, err.Error())
```
nixpanic: code duplication for `NotFound` error cases, consider using `fallthrough`:

```go
switch {
    case errors.Is(err, corerbd.ErrImageNotFound):
        fallthrough
    case errors.Is(err, util.ErrPoolNotFound):
        err = status.Errorf(codes.NotFound, err.Error())
...
```

this code block is repeated really a lot, it is probably worth it to put it in a helper function.

internal/rbd/group/volume_group.go
nixpanic: it shouldn't, `vg.GetID()` is the preferred way, or in some cases `vg.id` is safe.

internal/csi-common/utils.go
```diff
@@ -131,6 +131,9 @@ func GetIDFromReplication(req interface{}) string {
 		if reqID == "" {
 			reqID = r.GetVolumeId() //nolint:nolintlint,staticcheck // req.VolumeId is deprecated
 		}
+		if reqID == "" {
+			reqID = src.GetVolumegroup().GetVolumeGroupId()
```
nixpanic: can `GetVolumegroup()` return nil?

nixpanic (changes requested):
few minor things that should be improved. Thanks for splitting these commits out of the larger one!

internal/csi-addons/rbd/volumegroup.go
Madhu-1: i did think and reverted back, just before you reviewed the PR, will bring it back

internal/csi-addons/rbd/volumegroup.go
Madhu-1: added volumegroup, volume details are already returned by HandleParentImageExistence

internal/csi-common/utils.go
Madhu-1: yes, missed this one, fixed now, added a test for it as well

internal/rbd/manager.go
```diff
@@ -124,7 +124,18 @@ func (mgr *rbdManager) GetVolumeByID(ctx context.Context, id string) (types.Volu
...
 	volume, err := GenVolFromVolID(ctx, id, creds, mgr.secrets)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get volume from id %q: %w", id, err)
+		switch {
+		case errors.Is(err, ErrImageNotFound):
+			err = fmt.Errorf("volume %s not found: %w", id, err)
+
+			return nil, err
+		case errors.Is(err, util.ErrPoolNotFound):
+			err = fmt.Errorf("pool %s not found for %s: %w", volume.Pool, id, err)
```
nixpanic, line 133: `volume.Pool` seems weird, the `Volume` interface does not have a `Pool` attribute.

On a second look, the `GenVolFromVolID()` function returns an `rbdVolume`, so it should work.

nixpanic (approved):

Rakshith-R (approved):

Rakshith-R:
@mergifyio queue

nixpanic:
@mergifyio requeue

PR #4750, opened 2024-08-06
rbd: VolumeGroupReplicationContent controller to regenerate the OMAP data

iPraveenParihar:
# Describe what this PR does #

This commit adds new controller that watches for the
VolumeGroupReplicationContent and regenerates the OMAP data if
it doesn't exists.

**Checklist:**

* [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the [developer guide](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#commit-messages).
* [x] Reviewed the developer guide on [Submitting a Pull
  Request](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#development-workflow)
* [x] [Pending release notes](https://github.com/ceph/ceph-csi/blob/devel/PendingReleaseNotes.md) updated with breaking and/or notable changes for the next major release.
* [x] Documentation has been updated, if necessary.
* [x] Unit tests have been added, if necessary.
* [x] Integration tests have been added, if necessary.

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

* `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

PR #4753, opened 2024-08-07
journal: store CreationTime for VolumeGroups

nixpanic:
VolumeGroupSnapshots expect to have a CreationTime attribute. This can not be obtained by any (go-ceph) API calls, so it need to be stored in the journal.

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -520,7 +520,7 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	creationTime, err := rbdVol.GetCreationTime()
+	creationTime, err := rbdVol.GetCreationTime(ctx)
 	if err != nil {
 		log.ErrorLog(ctx, err.Error())
 
@@ -693,7 +693,7 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		ready = checkRemoteSiteStatus(ctx, sts.GetAllSitesStatus())
 	}
 
-	creationTime, err := rbdVol.GetCreationTime()
+	creationTime, err := rbdVol.GetCreationTime(ctx)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to get image info for %s: %s", rbdVol, err.Error())
 	}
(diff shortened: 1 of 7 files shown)


internal/rbd/snapshot.go
```diff
@@ -118,6 +121,16 @@ func (rbdSnap *rbdSnapshot) toVolume() *rbdVolume {
 	}
 }
 
+func (rbdSnap *rbdSnapshot) ToCSI(ctx context.Context) (*csi.Snapshot, error) {
```
Rakshith-R, line 146: Let's make use of it [here](https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/controllerserver.go#L1238) ?

internal/csi-addons/rbd/replication.go
```diff
@@ -746,8 +746,8 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 }
 
 // timestampToString converts the time.Time object to string.
-func timestampToString(st *timestamppb.Timestamp) string {
-	return fmt.Sprintf("seconds:%d nanos:%d", st.GetSeconds(), st.GetNanos())
+func timestampToString(st *time.Time) string {
+	return fmt.Sprintf("seconds:%d nanos:%d", st.Unix(), st.Nanosecond())
```
Madhu-1, line 750: can we check on this one, Does Unix is equivalent to GetSeconds()?

nixpanic: Yes
- https://pkg.go.dev/google.golang.org/protobuf/types/known/timestamppb#Timestamp
- https://pkg.go.dev/time#Time.Unix

internal/rbd/snapshot.go
nixpanic: oh, sure, this function actually was not required to be included, it is part of my VolumeGroupSnapshot work. I'll include it here so that it all gets a little cleaner.

Rakshith-R (approved):

nixpanic:
@iPraveenParihar this adds the creation time to the journal, maybe you need to add that to #4750 as well?

nixpanic:
@mergifyio rebase

iPraveenParihar (approved):
Thanks!

nixpanic:
@mergifyio queue

PR #4794, opened 2024-08-21
rbd: move common functions for VolumeGroup structs into own type

nixpanic:
Another PR that prepares for VolumeGroupSnapshot support. Refactoring common functions and attributes only, not introducing any functional change.

Many functions that are implemented for the volumeGroup type can be
shared with the (coming) volumeGroupSnapshot type. Move these functions
into a commonVolumeGroup type, so that volumeGroup and
volumeGroupSnapshot can inherit them.

### DIFF
--- a/internal/rbd/group/util.go
+++ b/internal/rbd/group/util.go
@@ -1,0 +1,248 @@
+/*
+Copyright 2024 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


internal/rbd/group/util.go
```diff
@@ -0,0 +1,248 @@
...
+		cvg.credentials = nil
+	}
+
+	log.DebugLog(ctx, "destroyed volume group instance with id %q", cvg.id)
+}
+
+// getVolumeGroupAttributes fetches the attributes from the journal, sets some
+// of the common values for the VolumeGroup and returns the attributes struct
+// for further consumption (like checking the VolumeMap).
+func (cvg *commonVolumeGroup) getVolumeGroupAttributes(ctx context.Context) (*journal.VolumeGroupAttributes, error) {
```
Madhu-1, line 123: why this is not an exposed function? are we going to have wrapper to get each attributes from the omap? in that case wont we making more than required API calls?

nixpanic: The VolumeGroup struct (and VolumeGroupSnapshot struct) can just call it. Ideally those structs do not expose any internals like the journalling.

Madhu-1 (approved):

Rakshith-R (approved):

Madhu-1:
@mergifyio queue

PR #4836, opened 2024-09-12
ci: tarball downloads have moved to dl.k8s.io

nixpanic:
storage.googleapis.com does not contain the latest releases. All
tarballs are now available from dl.k8s.io.


black-dragon74:
Same as https://github.com/ceph/ceph-csi/pull/4835?

PR #4840, opened 2024-09-16
cleanup: remove unneeded `updateSnapshotDetails()` function

nixpanic:
`updateSnapshotDetails()` just calls `getImageInfo()` on an `rbdVolume`
created from the `rbdSnapshot`. `getImageInfo()` is a function of the
base `rbdImage` struct, so there really is no need for this indirection.

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -790,7 +790,7 @@ func (cs *ControllerServer) createBackingImage(
 			}
 		}
 	}()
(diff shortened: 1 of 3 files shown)


iPraveenParihar (dismissed):

Rakshith-R (dismissed):

nixpanic:
@mergifyio queue

nixpanic:
Something isn't right here... Maybe the ImageID of the snapshot/parent should not get updated?

PR #4868, opened 2024-09-25
util: use protobuf encoding for core k8s apis

Nikhil-Ladha:
For core K8s API objects like Pods, Nodes, etc., we can use protobuf encoding which reduces CPU consumption related to (de)serialization, reduces overall latency of the API call, reduces memory footprint, reduces the amount of work performed by the GC and results in quicker propagation of objects to event handlers of shared informers.

### DIFF
--- a/internal/util/k8s/client.go
+++ b/internal/util/k8s/client.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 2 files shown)


internal/util/k8s/client.go
```diff
@@ -47,6 +48,7 @@ func NewK8sClient() (*kubernetes.Clientset, error) {
 			return nil, fmt.Errorf("failed to get cluster config: %w", err)
 		}
 	}
+	cfg.ContentType = runtime.ContentTypeProtobuf
```
Madhu-1, line 51: there is a PV [controller](https://github.com/ceph/ceph-csi/tree/devel/internal/controller), can you please if we need any changes to that as well?

Nikhil-Ladha: Updated the code for that controller.

nixpanic (approved):

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

PR #4879, opened 2024-10-01
ci: use private retest github action

Nikhil-Ladha:
Use private retest github action instead of pinning to a single branch/commit hash.

Closes: #4878

### DIFF
--- a/.github/workflows/retest.yaml
+++ b/.github/workflows/retest.yaml
@@ -14,9 +14,11 @@ jobs:
     if: github.repository == 'ceph/ceph-csi'
     runs-on: ubuntu-latest
     steps:
-      # path to the retest action
-      # yamllint disable-line rule:line-length
-      - uses: ceph/ceph-csi/actions/retest@28dc64dcae3cec8d11d84bdf525bda0ef757c688  # devel
+      - name: Checkout the ceph-csi respository
+        # yamllint disable-line rule:line-length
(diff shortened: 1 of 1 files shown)


nixpanic (approved):
Should be good, thanks!

.github/workflows/retest.yaml
```diff
@@ -14,9 +14,10 @@ jobs:
     if: github.repository == 'ceph/ceph-csi'
     runs-on: ubuntu-latest
     steps:
-      # path to the retest action
-      # yamllint disable-line rule:line-length
-      - uses: ceph/ceph-csi/actions/retest@28dc64dcae3cec8d11d84bdf525bda0ef757c688  # devel
+      - name: Checkout the ceph-csi respository
+        uses: actions/checkout@v4
```
Madhu-1: we should not be using v4 (it should be a hash) here isn't it?

nixpanic: indeed, use this:

```yaml
uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938 # v4.2.0
```

nixpanic (changes requested):

.github/workflows/retest.yaml
Nikhil-Ladha: Ah, right.
Updated it now :)

Madhu-1 (approved):

nixpanic (approved):

nixpanic:
@mergifyio queue

PR #4502, opened 2024-03-15
rbd: VolumeGroupSnapshot support

nixpanic:
Add support for VolumeGroupSnapshots in RBD. The last two commits enable the feature in the Group Controller Server and expose the capability. All other commits provide the functionality through the `rbd.Manager` interface.

Currently there is no Ceph container-image release that provides the required librbd features. Building from this PR will not provide support for VolumeGroupSnapshot yet, the base container-image needs to be set to Ceph CI main branch (`quay.ceph.io/ceph-ci/ceph:main`) for that. A test-image based on Ceph main can be found at `quay.io/nixpanic/cephcsi:pr_4739`.

Notable changes:

- `internal/rbd_types` package with interfaces so that objects can be passed around cleaner
- `internal/rbd/volume.go` implementing the new `Volume` interface for `rbdImage`
- `internal/rbd_group` package for all RBD-group functionalities
- `internal/rbd/group_controller.go` for all CSI VolumeGroup service procedures

Depends-on: #4794
Depends-on: #4870
Depends-on: #4871
Depends-on: #4884
Depends-on: #4898
Depends-on: #4902
Depends-on: #4904
Depends-on: #4915

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

* `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

PR #4885, opened 2024-10-03
rbd: do not return an error when deleting a non-existing image

nixpanic:
Deleting an image that has been already remove, should not be reported
as a failure. This improves the idempotency of the `rbdImage.Delete()`
function, making it easier for callers to detect real errors.

## Related issues ##

Found while working on #4502.

---

<details>

</details>

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -671,7 +671,7 @@ func (ri *rbdImage) Delete(ctx context.Context) error {
 
 	rbdImage := librbd.GetImage(ri.ioctx, image)
 	err = rbdImage.Trash(0)
-	if err != nil {
+	if err != nil && !errors.Is(err, librbd.ErrNotFound) {
 		log.ErrorLog(ctx, "failed to delete rbd image: %s, error: %v", ri, err)
 
 		return err
@@ -694,7 +694,7 @@ func (ri *rbdImage) trashRemoveImage(ctx context.Context) error {
 	_, err = ta.AddTrashRemove(admin.NewImageSpec(ri.Pool, ri.RadosNamespace, ri.ImageID))
 
 	rbdCephMgrSupported := isCephMgrSupported(ctx, ri.ClusterID, err)
-	if rbdCephMgrSupported && err != nil {
+	if rbdCephMgrSupported && err != nil && !errors.Is(err, librbd.ErrNotFound) {
 		log.ErrorLog(ctx, "failed to add task to delete rbd image: %s, %v", ri, err)
 
 		return err


iPraveenParihar (approved):

Rakshith-R:
We need to not have this check in order to trigger https://github.com/ceph/ceph-csi/blob/611e5d920e1623a2f58e2eff8bbb15cb3c21213d/internal/rbd/rbd_util.go#L623

nixpanic:
> We need to not have this check in order to trigger
> 
> https://github.com/ceph/ceph-csi/blob/611e5d920e1623a2f58e2eff8bbb15cb3c21213d/internal/rbd/rbd_util.go#L623

Sorry, I am not sure what you mean by that... `ensureImageCleanup` can cause a panic without those error checks?

internal/rbd/rbd_util.go
```diff
@@ -694,7 +694,7 @@ func (ri *rbdImage) trashRemoveImage(ctx context.Context) error {
 	_, err = ta.AddTrashRemove(admin.NewImageSpec(ri.Pool, ri.RadosNamespace, ri.ImageID))
 
 	rbdCephMgrSupported := isCephMgrSupported(ctx, ri.ClusterID, err)
-	if rbdCephMgrSupported && err != nil {
+	if rbdCephMgrSupported && err != nil && !errors.Is(err, librbd.ErrNotFound) {
```
Madhu-1, line 697: @nixpanic is `AddTrashRemove` Mrg API is also support this error message?

Rakshith-R:
> > We need to not have this check in order to trigger
> > https://github.com/ceph/ceph-csi/blob/611e5d920e1623a2f58e2eff8bbb15cb3c21213d/internal/rbd/rbd_util.go#L623
> 
> Sorry, I am not sure what you mean by that... `ensureImageCleanup` can cause a panic without those error checks?

If the image does not exist, then the flow should not have reached that point at all.
Image not being found would've lead to ensureImageCleanup in actual delete procedure.
I think there are multiple processes acting on same image that's causing this error/panic in your case ??

internal/rbd/rbd_util.go
nixpanic: yes, it seemed so. I have been testing with the Ceph main branch, maybe things changed there?

nixpanic:
> > > We need to not have this check in order to trigger
> > > https://github.com/ceph/ceph-csi/blob/611e5d920e1623a2f58e2eff8bbb15cb3c21213d/internal/rbd/rbd_util.go#L623
> > 
> > 
> > Sorry, I am not sure what you mean by that... `ensureImageCleanup` can cause a panic without those error checks?
> 
> If the image does not exist, then the flow should not have reached that point at all. Image not being found would've lead to ensureImageCleanup in actual delete procedure. I think there are multiple processes acting on same image that's causing this error/panic in your case ??

I don't think it multiple processes are trying to delete the same volume at the same time. However a 2nd `DeleteVolume` seems to come immediately after the 1st completed (https://github.com/kubernetes-csi/external-provisioner/issues/1235 might be related?).

Madhu-1:
> > > > We need to not have this check in order to trigger
> > > > https://github.com/ceph/ceph-csi/blob/611e5d920e1623a2f58e2eff8bbb15cb3c21213d/internal/rbd/rbd_util.go#L623
> > > 
> > > 
> > > Sorry, I am not sure what you mean by that... `ensureImageCleanup` can cause a panic without those error checks?
> > 
> > 
> > If the image does not exist, then the flow should not have reached that point at all. Image not being found would've lead to ensureImageCleanup in actual delete procedure. I think there are multiple processes acting on same image that's causing this error/panic in your case ??
> 
> I don't think it multiple processes are trying to delete the same volume at the same time. However a 2nd `DeleteVolume` seems to come immediately after the 1st completed ([kubernetes-csi/external-provisioner#1235](https://github.com/kubernetes-csi/external-provisioner/issues/1235) might be related?).

we do have in memory lock to ensure that this should not happen.

@Rakshith-R any further comments on this one?

Rakshith-R:
> > > > > We need to not have this check in order to trigger
> > > > > https://github.com/ceph/ceph-csi/blob/611e5d920e1623a2f58e2eff8bbb15cb3c21213d/internal/rbd/rbd_util.go#L623
> > > > 
> > > > 
> > > > Sorry, I am not sure what you mean by that... `ensureImageCleanup` can cause a panic without those error checks?
> > > 
> > > 
> > > If the image does not exist, then the flow should not have reached that point at all. Image not being found would've lead to ensureImageCleanup in actual delete procedure. I think there are multiple processes acting on same image that's causing this error/panic in your case ??
> > 
> > 
> > I don't think it multiple processes are trying to delete the same volume at the same time. However a 2nd `DeleteVolume` seems to come immediately after the 1st completed ([kubernetes-csi/external-provisioner#1235](https://github.com/kubernetes-csi/external-provisioner/issues/1235) might be related?).
> 
> we do have in memory lock to ensure that this should not happen.
> 
> @Rakshith-R any further comments on this one?

Yea, I think We'll encounter this error only when another process puts the image to trash while this process reaches the `rbd.Trash(0)` or `ra.addTrashRemove` call. The only reason a process gets to that stage is when a rbd image is found.
 
We should treat such occurrences as a race condition happening somewhere in the code and fix that instead of the current error check to avoid it.

(sorry, missed this comment earlier)

Rakshith-R:
@nixpanic I think we can close this pr ?

nixpanic:
I have not seen this issue anymore, closing.

PR #4898, opened 2024-10-10
rbd: add feature check to see if GroupSnapGetInfo is available

nixpanic:
The go-ceph rbd package provides the GroupSnapGetInfo function, but it may return ErrUnsupported when called. Returning this error after advertising the support for VolumeGroupSnapshot seems ugly.

In order to advertise support for VolumeGroupSnapshot, SupportsGroupSnapGetInfo() can be used, which detects the required C function of librbd.

## Related issues ##

To be consumed in #4502 

---

<details>

</details>

### DIFF
--- a/internal/rbd/features/features.go
+++ b/internal/rbd/features/features.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2024 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package features
+
+/*
+#cgo LDFLAGS: -lrbd
+#include <rbd/librbd.h>
+*/
+import "C"
+
+import (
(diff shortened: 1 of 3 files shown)


Rakshith-R (approved):

internal/rbd/features/features.go
```diff
@@ -0,0 +1,58 @@
...
+// SupportsGroupSnapGetInfo detects if librbd has the rbd_group_snap_get_info
+// function.
+func SupportsGroupSnapGetInfo() (bool, error) {
+	groupGetSnapInfoOnce.Do(func() {
+		// make sure librbd.so.x is loaded, might not (yet) be the case
+		// if no rbd functions are called
+		var opts C.rbd_image_options_t
+		//nolint:gocritic // ignore result of rbd_image_options functions
+		C.rbd_image_options_create(&opts)
+		C.rbd_image_options_destroy(opts)
```
Madhu-1, line 45: @nixpanic do we need to create/destory the image? can you please add a comment about a need for it as well?

nixpanic: this is just to call an RBD function that won't cause any harm.

RBD-image-options can be created and destroyed without accessing the Ceph cluster. You can see the options as an in-memory list of features. The options are used when an image is created, and features of the image can be set in the options object and passed to `rbd.CreateImage()` and similar functions.

nixpanic: Also, the comment is just above it:
```
		// make sure librbd.so.x is loaded, might not (yet) be the case
		// if no rbd functions are called
```

Madhu-1 (approved):

nixpanic:
@mergifyio queue

PR #4899, opened 2024-10-10
e2e: add tests for RBD VolumeGroupSnapshots

nixpanic:
Support for RBD VolumeGroupSnapshots is coming soon. However the implementation depends on changes in librbd, which are currently only available in the Ceph master branch.

Only when librbd in the Ceph-CSI container has support for the new function, the tests should te attempred

## Related issues ##

This is no a 100% strict dependency, as current Ceph releases do not have the required functions yet. It is expected that #4502 gets merged before Ceph has a stable release (non master branch builds).

Sortof-depends-on: #4502

Example YAML files that are used by the test are posted as #4901.

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

* `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

PR #4901, opened 2024-10-11
doc: add VolumeGroupSnapshot examples for RBD

nixpanic:
By labelling the PersistentVolumeClaim with `group: test`, the label
selector in the VolumeGroupSnapshot will create a snapshot of an RBD
group containing the single volume (or more if there are other volumes
in the same namespace with that label).

Once the VolumeGroupSnapshot is created, each volume in the group will
have a VolumeSnapshot that can be restored.

These files are also used by #4899 and eventually #4502, but the test is skipped until Ceph has a
new release with the required feature.

### DIFF
--- a/examples/rbd/groupsnapshot.yaml
+++ b/examples/rbd/groupsnapshot.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
+kind: VolumeGroupSnapshot
+metadata:
+  name: rbd-groupsnapshot
+spec:
+  source:
+    selector:
+      matchLabels:
+        # The PVCs will need to have this label for it to be
+        # included in the VolumeGroupSnapshot
+        group: test
+  volumeGroupSnapshotClassName: csi-rbdplugin-groupsnapclass
--- a/examples/rbd/groupsnapshotclass.yaml
+++ b/examples/rbd/groupsnapshotclass.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
+kind: VolumeGroupSnapshotClass
+metadata:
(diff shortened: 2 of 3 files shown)


examples/rbd/groupsnapshotclass.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
```
iPraveenParihar, line 2: @nixpanic, I see example values are openshift specific.
Can you make it similar to cephfs groupsnapshotclass.yaml example file?

nixpanic: ah, yes, it seems I missed some changes for that

nixpanic: corrected this, please check again, thanks!

iPraveenParihar (dismissed):
Thanks!

yati1998 (dismissed):

examples/rbd/groupsnapshot.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
+kind: VolumeGroupSnapshot
+metadata:
+  name: new-groupsnapshot-demo-2
```
Rakshith-R:
```suggestion
  name: rbd-groupsnapshot
```
LGTM,
Just  a tiny nit
Let's keep naming conventions the same.

nixpanic: Sure, definitely nicer.

iPraveenParihar (approved):

Rakshith-R (approved):

nixpanic:
@mergifyio queue

PR #4912, opened 2024-10-17
rbd: Add timeout for cryptsetup commands

black-dragon74:

This PR introduces a new wrapper around LUKS utility functions that allow us to have context aware subsequent operations, `util.NewLuksWrapperWithContext`. This context is shared across the functions exposed by this wrapper.

A sample use case is to have a context that times out after certain period of time.

# How to Test #

We can simulate a timeout by replacing the cryptsetup binary with a wrapper script that introduces some delay.

- Exec into the pod where the cryptsetup binary is located (rbd-plugin in case of key rotation)
- `mv /usr/sbin/cryptsetup /usr/sbin/cryptsetup_orig`
- Create a new `/usr/sbin/cryptsetup` and paste the following:
```sh
#!/usr/bin/env bash

set -m

sleep 140

exec /usr/sbin/cryptsetup_orig "$@"
```
- `chmod +x /usr/sbin/cryptsetup`
- When the process is killed due to timeout you will see a log like:
```log
timeout occurred while running cryptsetup args: [...]
```

### DIFF
--- a/internal/rbd/encryption.go
+++ b/internal/rbd/encryption.go
@@ -26,6 +26,7 @@ import (
 
 	kmsapi "github.com/ceph/ceph-csi/internal/kms"
 	"github.com/ceph/ceph-csi/internal/util"
+	"github.com/ceph/ceph-csi/internal/util/cryptsetup"
 	"github.com/ceph/ceph-csi/internal/util/lock"
 	"github.com/ceph/ceph-csi/internal/util/log"
 
@@ -475,8 +476,14 @@ func (rv *rbdVolume) RotateEncryptionKey(ctx context.Context) error {
 	// Lock params
 	lockName := rv.VolID + "-mutexlock"
 	lockDesc := "Key rotation mutex lock for " + rv.VolID
-	lockDuration := 3 * time.Minute
 	lockCookie := rv.VolID + "-enc-key-rotate"
+
+	// Keep this a little more than ExecutionTimeout to have some buffer
+	// for cleanup. If this lock is a part of some gRPC call, the client
+	// should always timeout after the lockDuration to avoid issues.
+	lockDuration := cryptsetup.ExecutionTimeout + 30*time.Second
+	timedCtx, cancel := context.WithTimeout(ctx, cryptsetup.ExecutionTimeout)
+	defer cancel()
 
 	// Acquire the exclusive lock based on vol id
 	lck := lock.NewLock(rv.ioctx, rv.VolID, lockName, lockCookie, lockDesc, lockDuration)
@@ -500,8 +507,11 @@ func (rv *rbdVolume) RotateEncryptionKey(ctx context.Context) error {
 		return fmt.Errorf("failed to fetch the current passphrase for %q: %w", rv, err)
 	}
 
+	// Create a new luks wrapper
+	luks := cryptsetup.NewLUKSWrapper(timedCtx)
+
 	// Step 2: Add current key to slot 1
-	err = util.LuksAddKey(devicePath, oldPassphrase, oldPassphrase, luksSlot1)
+	err = luks.AddKey(devicePath, oldPassphrase, oldPassphrase, luksSlot1)
 	if err != nil {
 		return fmt.Errorf("failed to add curr key to luksSlot1: %w", err)
 	}
@@ -513,20 +523,20 @@ func (rv *rbdVolume) RotateEncryptionKey(ctx context.Context) error {
 		return fmt.Errorf("failed to generate a new passphrase: %w", err)
 	}
 
-	err = util.LuksAddKey(devicePath, oldPassphrase, newPassphrase, luksSlot0)
+	err = luks.AddKey(devicePath, oldPassphrase, newPassphrase, luksSlot0)
 	if err != nil {
 		return fmt.Errorf("failed to add the new key to luksSlot0: %w", err)
 	}
 
 	// Step 4: Add the new key to KMS
-	err = rv.blockEncryption.StoreCryptoPassphrase(ctx, rv.VolID, newPassphrase)
+	err = rv.blockEncryption.StoreCryptoPassphrase(timedCtx, rv.VolID, newPassphrase)
 	if err != nil {
 		return fmt.Errorf("failed to update the new key into the KMS: %w", err)
 	}
 
 	// Step 5: Remove the old key from slot 1
 	// We use the newPassphrase to authenticate LUKS here
-	err = util.LuksRemoveKey(devicePath, newPassphrase, luksSlot1)
+	err = luks.RemoveKey(devicePath, newPassphrase, luksSlot1)
 	if err != nil {
 		return fmt.Errorf("failed to remove the backup key from luksSlot1: %w", err)
 	}
--- a/internal/util/cephcmds.go
+++ b/internal/util/cephcmds.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/ceph/ceph-csi/internal/util/log"
+	"github.com/ceph/ceph-csi/internal/util/stripsecrets"
 
 	"github.com/ceph/go-ceph/rados"
 )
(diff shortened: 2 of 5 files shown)


nixpanic (commented):
This looks reasonable to me. Did you manage to test this somehow, and see how everything gets corrected by a different container?

Madhu-1:
> This looks reasonable to me. Did you manage to test this somehow, and see how everything gets corrected by a different container?

we need to ensure that the CLI command we start is killed and doesn't continue to execute, please test it the changes to ensure that it works as expected and also provide some test results on how this is tested

internal/util/cryptsetup.go
```diff
@@ -18,18 +18,26 @@ package util
...
 
 	"github.com/ceph/ceph-csi/internal/util/file"
 	"github.com/ceph/ceph-csi/internal/util/log"
 )
 
-// Limit memory used by Argon2i PBKDF to 32 MiB.
-const cryptsetupPBKDFMemoryLimit = 32 << 10 // 32768 KiB
+const (
+	// Maximum time to wait for cryptsetup commands to complete.
+	CryptSetupExecutionTimeout = 3 * time.Minute
```
Madhu-1: Lets keep the timeout which is almost less than the GRPC timeout between the sidecar and the cephcsi so that we try and if it didn't work as will kill the process before the next GRPC kicks in on any other nodes/same node.

black-dragon74:
> This looks reasonable to me. Did you manage to test this somehow, and see how everything gets corrected by a different container?

Trying but not able to yet. I am thinking of having a wrapper to cryptsetup that sleeps for some time and then tries to execute the intended command. The key is to kill the process in midst of operations.

I will test it and report.

internal/util/cryptsetup.go
black-dragon74: I like the idea. Let's keep them a minute apart then? Accounting for any delays that might occur in sending gRPC request and also to keep some buffer period after sending SIGKILL?

Madhu-1: minute might be too much, we can reduce it by 30 seconds to be on safer side, @nixpanic thoughts?

nixpanic:
> Trying but not able to yet. I am thinking of having a wrapper to cryptsetup that sleeps for some time and then tries to execute the intended command. The key is to kill the process in midst of operations.

I also think the only way reproduce this, is by adding a wrapper for the command. It helps  to make the command do something that can be verified after aborting, maybe something like this (untested):

```bash
#!/bin/sh
#
# Assume a 1GB block-device, write unencrypted data at 500MB offset.
#

# run normal crypsetup, it will be quick and exits cleanly
cryptsetup.real $@

# run dd in a loop, small I/O with sync to make it slower
# only runs the loop if /tmp/testing exists
while [ -e /tmp/testing ]
do
    # set correct device and verify parameters
    dd if=/dev/urandom of=/dev/rbd... bs=1 count=100m offset=500m
done
```

Make sure to name the script `cryptsetup` and place is in the location where `cryptsetup` normally lives. The original `cryptsetup` executable should be renamed to `cryptsetup.real` so that it still can be executed.

nixpanic:
Hey @black-dragon74 , I'll mark this as a draft for now, otherwise I keep checking it for updates. Please move it back as ready for review once you have tested it. Thanks!

Madhu-1 (dismissed):
LGTM

internal/util/cryptsetup.go
```diff
@@ -18,22 +18,41 @@ package util
...
+const (
+	// Maximum time to wait for cryptsetup commands to complete.
+	CryptSetupExecutionTimeout = 2*time.Minute + 30*time.Second
+
+	// Limit memory used by Argon2i PBKDF to 32 MiB.
+	cryptsetupPBKDFMemoryLimit = 32 << 10 // 32768 KiB
+)
+
+// LuksWrapper is a struct that provides a context-aware wrapper around cryptsetup commands.
+type LuksWrapper struct {
```
nixpanic: nice approach. Please call this LUKSWrapper as LUKS is the correct spelling.

internal/util/cryptsetup.go
```diff
@@ -18,22 +18,41 @@ package util
...
+	ctx context.Context
+}
+
+// NewLuksWrapperWithContext creates a new LuksWrapper instance with the provided context.
+// The context is used to control the lifetime of the cryptsetup commands executed by the LuksWrapper.
+func NewLuksWrapperWithContext(ctx context.Context) *LuksWrapper {
```
nixpanic: Same here, `NewLUKSWrapper` would be ok. Also no need to add the `WithContext` as there is only one `New..` call.

To prevent issues with incorrect use, you could do the  following as well:
```go
type LUKSWrapper interface {
    Format(devicePath, passphrase string) (string, string, error)
    Open(...
    ...
}

type luksWrapper struct {
    ctx context.Context
}

// verify that luksWrapper implements the LUKSWrapper interface
var _ LUKSWrapper = &luksWrapper{}

func NewLuksWrapper(ctx context.Context) LuksWrapper {
    return &luksWrapper{ctx: ctx}
}
```

This way, nbody outside the `internal/util` package can create stucts of `luksWrapper` without setting a context.

As this LUKS API is much cleaner (already without the above suggestion), consider moving it all to a `internal/util/cryptsetup` package. Currently `internal/util` has too many things. Moving can be done in a separate PR later on too.

internal/util/cryptsetup.go
```diff
@@ -18,22 +18,41 @@ package util
...
+	return &LuksWrapper{ctx: ctx}
+}
 
 // LuksFormat sets up volume as an encrypted LUKS partition.
-func LuksFormat(devicePath, passphrase string) (string, string, error) {
-	return execCryptsetupCommand(
+func (l *LuksWrapper) LuksFormat(devicePath, passphrase string) (string, string, error) {
```
nixpanic: `LuksFormat` and other functions are now part of a LUKS type, no need to include `Luks` in every function name. Instead, rename the functions like `func (l *luksWrapper) Format(...`

nixpanic (changes requested):
Renaming the type and functions make it better, please take care of that.

The suggestion with the interface and moving to a separate package are optional and can be done in a follow-up PR if you prefer.

internal/util/cryptsetup.go
black-dragon74: I have updated the PR with suggested changes. Should I move `StripSecrets` inside `util/log`? Makes more sense there (plus, we will have one less package).

I skipped `var _ LUKSWrapper = &luksWrapper{}` as `NewLUKSWrapper` is expected to return a type that implements `LUKSWrapper` so the interface verification would be done here. Correct?

internal/rbd/encryption.go
```diff
@@ -500,8 +505,13 @@ func (rv *rbdVolume) RotateEncryptionKey(ctx context.Context) error {
 		return fmt.Errorf("failed to fetch the current passphrase for %q: %w", rv, err)
 	}
 
+	// Create a new luks wrapper
+	timeoutCtx, cancel := context.WithTimeout(ctx, cryptsetup.ExecutionTimeout)
+	luks := cryptsetup.NewLUKSWrapper(timeoutCtx)
+	defer cancel()
+
```
Rakshith-R: This should be done before obtaining locks at line 481.
Pass the same context to get/put cryptopassphrase calls as well.

Rakshith-R (dismissed):
LGTM, just need to start context before obtaining locks and use the same context for get/put calls to kms.

internal/rbd/encryption.go
black-dragon74: Done!

nixpanic (approved):

nixpanic:
@mergifyio rebase

internal/util/cryptsetup.go
nixpanic: Yes, that is fine. The error message might be a little more confusing in case there is a mismatch between the interface/struct, but nothing blocking this imho.

Great cleanup, looks much tidier this way.

Madhu-1 (approved):

nixpanic:
@mergifyio queue

PR #4918, opened 2024-10-28
cephfs: Log clone progress during a clone operation

### LINKED ISSUE #4813: [Cephfs] Log clone progress
cephfs is now providing clone progress status https://github.com/ceph/ceph/pull/54620/files, we need to update go-ceph and csi for the same, This is very helpful for users to understand what's going on when cloning

Nikhil-Ladha:
Add clone progress report in the clone status operation and report the same in the logs for user's reference.

Fixes: #4813 

**Checklist:**

  guidelines in the [developer
* [X] [Pending release
* [X] Documentation has been updated, if necessary.
* [X] Integration tests have been added, if necessary.

---

<details>

</details>

### DIFF
--- a/internal/cephfs/core/clone.go
+++ b/internal/cephfs/core/clone.go
@@ -28,16 +28,17 @@ import (
 
 // cephFSCloneState describes the status of the clone.
 type cephFSCloneState struct {
-	state    admin.CloneState
-	errno    string
-	errorMsg string
+	state          admin.CloneState
+	progressReport admin.CloneProgressReport
+	errno          string
+	errorMsg       string
 }
 
 // CephFSCloneError indicates that fetching the clone state returned an error.
-var CephFSCloneError = cephFSCloneState{}
+var CephFSCloneError = &cephFSCloneState{}
 
 // ToError checks the state of the clone if it's not cephFSCloneComplete.
-func (cs cephFSCloneState) ToError() error {
+func (cs *cephFSCloneState) ToError() error {
 	switch cs.state {
 	case admin.CloneComplete:
 		return nil
@@ -52,6 +53,14 @@ func (cs cephFSCloneState) ToError() error {
 	}
 
 	return nil
+}
+
+func (cs *cephFSCloneState) GetProgressReport() admin.CloneProgressReport {
+	return admin.CloneProgressReport{
+		PercentageCloned: cs.progressReport.PercentageCloned,
+		AmountCloned:     cs.progressReport.AmountCloned,
+		FilesCloned:      cs.progressReport.FilesCloned,
+	}
 }
 
 // CreateCloneFromSubvolume creates a clone from a subvolume.
@@ -87,7 +96,7 @@ func (s *subVolumeClient) CreateCloneFromSubvolume(
 		return err
 	}
 
-	var cloneState cephFSCloneState
+	var cloneState *cephFSCloneState
 	cloneState, err = s.GetCloneState(ctx)
 	if err != nil {
 		log.ErrorLog(ctx, "failed to get clone state: %v", err)
@@ -157,7 +166,7 @@ func (s *subVolumeClient) CreateCloneFromSnapshot(
 			}
 		}
 	}()
-	var cloneState cephFSCloneState
+	var cloneState *cephFSCloneState
 	// avoid err variable shadowing
 	cloneState, err = s.GetCloneState(ctx)
 	if err != nil {
@@ -182,7 +191,7 @@ func (s *subVolumeClient) CreateCloneFromSnapshot(
 }
 
 // GetCloneState returns the clone state of the subvolume.
-func (s *subVolumeClient) GetCloneState(ctx context.Context) (cephFSCloneState, error) {
+func (s *subVolumeClient) GetCloneState(ctx context.Context) (*cephFSCloneState, error) {
 	fsa, err := s.conn.GetFSAdmin()
 	if err != nil {
 		log.ErrorLog(
@@ -209,10 +218,11 @@ func (s *subVolumeClient) GetCloneState(ctx context.Context) (cephFSCloneState, 
 		errStr = failure.ErrStr
 	}
 
-	state := cephFSCloneState{
-		state:    cs.State,
-		errno:    errno,
-		errorMsg: errStr,
+	state := &cephFSCloneState{
+		state:          cs.State,
+		progressReport: cs.ProgressReport,
+		errno:          errno,
+		errorMsg:       errStr,
 	}
 
 	return state, nil
--- a/internal/cephfs/store/fsjournal.go
+++ b/internal/cephfs/store/fsjournal.go
@@ -128,6 +128,17 @@ func CheckVolExists(ctx context.Context,
 		}
 		err = cloneState.ToError()
 		if errors.Is(err, cerrors.ErrCloneInProgress) {
+			progressReport := cloneState.GetProgressReport()
+			// append progress report only if the progress report parameters are present.
+			if progressReport.PercentageCloned != "" {
+				err = fmt.Errorf("%w. progress report: percentage cloned=%s, amount cloned=%s, files cloned=%s",
+					err,
+					progressReport.PercentageCloned,
+					progressReport.AmountCloned,
+					progressReport.FilesCloned)
+				log.ErrorLog(ctx, err.Error())
+			}
+
 			return nil, err
 		}
 		if errors.Is(err, cerrors.ErrClonePending) {
--- a/internal/cephfs/core/clone_test.go
+++ b/internal/cephfs/core/clone_test.go
@@ -28,11 +28,11 @@ import (
 func TestCloneStateToError(t *testing.T) {
 	t.Parallel()
 	errorState := make(map[cephFSCloneState]error)
-	errorState[cephFSCloneState{fsa.CloneComplete, "", ""}] = nil
-	errorState[CephFSCloneError] = cerrors.ErrInvalidClone
-	errorState[cephFSCloneState{fsa.CloneInProgress, "", ""}] = cerrors.ErrCloneInProgress
-	errorState[cephFSCloneState{fsa.ClonePending, "", ""}] = cerrors.ErrClonePending
-	errorState[cephFSCloneState{fsa.CloneFailed, "", ""}] = cerrors.ErrCloneFailed
+	errorState[cephFSCloneState{fsa.CloneComplete, fsa.CloneProgressReport{}, "", ""}] = nil
+	errorState[*CephFSCloneError] = cerrors.ErrInvalidClone
+	errorState[cephFSCloneState{fsa.CloneInProgress, fsa.CloneProgressReport{}, "", ""}] = cerrors.ErrCloneInProgress
+	errorState[cephFSCloneState{fsa.ClonePending, fsa.CloneProgressReport{}, "", ""}] = cerrors.ErrClonePending
+	errorState[cephFSCloneState{fsa.CloneFailed, fsa.CloneProgressReport{}, "", ""}] = cerrors.ErrCloneFailed
 
 	for state, err := range errorState {
 		require.ErrorIs(t, state.ToError(), err)
--- a/internal/cephfs/core/volume.go
+++ b/internal/cephfs/core/volume.go
@@ -75,7 +75,7 @@ type SubVolumeClient interface {
 	// CreateCloneFromSubVolume creates a clone from the subvolume.
 	CreateCloneFromSubvolume(ctx context.Context, parentvolOpt *SubVolume) error
 	// GetCloneState returns the clone state of the subvolume.
-	GetCloneState(ctx context.Context) (cephFSCloneState, error)
+	GetCloneState(ctx context.Context) (*cephFSCloneState, error)
 	// CreateCloneFromSnapshot creates a clone from the subvolume snapshot.
 	CreateCloneFromSnapshot(ctx context.Context, snap Snapshot) error
 	// CleanupSnapshotFromSubvolume removes the snapshot from the subvolume.


Nikhil-Ladha:
cc @Madhu-1 @nixpanic

internal/cephfs/store/fsjournal.go
```diff
@@ -128,6 +128,14 @@ func CheckVolExists(ctx context.Context,
 		}
 		err = cloneState.ToError()
 		if errors.Is(err, cerrors.ErrCloneInProgress) {
+			progressReport := cloneState.GetProgressReport()
```
Madhu-1, line 131: it would be good to surface this error to the user as an error message not everyone will have access/check the log message

internal/cephfs/core/clone.go
```diff
@@ -54,6 +55,14 @@ func (cs cephFSCloneState) ToError() error {
 	return nil
 }
 
+func (cs cephFSCloneState) GetProgressReport() admin.CloneProgressReport {
+	return admin.CloneProgressReport{
```
Madhu-1, line 59: cant we return `cs.progressReport` in other works do we need this helper function?

Madhu-1 (commented):
what happens if we are using older version of the ceph that doesn't support the new fields? i assume it will be empty just want to confirm on it.

internal/cephfs/core/clone.go
Nikhil-Ladha: I initially opted for that, but since the struct is not exposed(public) to functions outside the `core` pkg, we can't directly access the members of the struct.
Hence, a helper function.

Nikhil-Ladha:
> what happens if we are using older version of the ceph that doesn't support the new fields? i assume it will be empty just want to confirm on it.

Yes, it will be empty.

internal/cephfs/store/fsjournal.go
Nikhil-Ladha: Done

Madhu-1: i dont see any update to add progressReport to the error message, can you please point me to the code where we are adding/returning progressReport as part of GRPC error message?

Nikhil-Ladha: Sorry, I misunderstood the previous comment.
Updated the code to return the progress report values as part of `err` value of the function call.
Please check now.

internal/cephfs/store/fsjournal.go
```diff
@@ -128,6 +128,14 @@ func CheckVolExists(ctx context.Context,
...
+			err = fmt.Errorf("%s. progress report: percentage cloned=%s, amount cloned=%s, files cloned=%s",
```
Madhu-1: use %w for err as the caller can still check the error using unwrapping.

Madhu-1 (commented):
can you please test it out and paste the PVC describe output, just to make sure we are surfacing it to user, feel free to add E2E if you think it can be done as well

Madhu-1 (dismissed):
LGTM, Lets get the testing results

Nikhil-Ladha:
Testing results:

PVC Events:
```
Events:
  Type     Reason                 Age                From                                                                                                             Message
  ----     ------                 ----               ----                                                                                                             -------
  Normal   ExternalProvisioning   68s (x2 over 68s)  persistentvolume-controller                                                                                      Waiting for a volume to be created either by the external provisioner 'rook-ceph.cephfs.csi.ceph.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
  Warning  ProvisioningFailed     68s                rook-ceph.cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-f5df68b95-9stgl_d702fa82-d9f0-4f54-a5b9-3bd7b2cdd7d5  failed to provision volume with StorageClass "rook-cephfs": rpc error: code = Aborted desc = clone from snapshot is pending
  Warning  ProvisioningFailed     67s                rook-ceph.cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-f5df68b95-9stgl_d702fa82-d9f0-4f54-a5b9-3bd7b2cdd7d5  failed to provision volume with StorageClass "rook-cephfs": rpc error: code = Aborted desc = clone from snapshot is already in progress. progress report: percentage cloned=0%, amount cloned=306/0, files cloned=3/1
  Normal   Provisioning           66s (x3 over 68s)  rook-ceph.cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-f5df68b95-9stgl_d702fa82-d9f0-4f54-a5b9-3bd7b2cdd7d5  External provisioner is provisioning volume for claim "default/cephfs-pvc-clone"
  Normal   ProvisioningSucceeded  58s                rook-ceph.cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-f5df68b95-9stgl_d702fa82-d9f0-4f54-a5b9-3bd7b2cdd7d5  Successfully provisioned volume pvc-70f1e45e-7d31-4e52-babe-8ff083860505
```

CSI Logs:
```
I1028 09:34:09.854647       1 fsjournal.go:137] ID: 38 Req-ID: pvc-70f1e45e-7d31-4e52-babe-8ff083860505 clone from snapshot is already in progress. progress report: percentage cloned=0%!,(MISSING) amount cloned=306/0, files cloned=3/1
E1028 09:34:09.854817       1 utils.go:271] ID: 38 Req-ID: pvc-70f1e45e-7d31-4e52-babe-8ff083860505 GRPC error: rpc error: code = Aborted desc = clone from snapshot is already in progress. progress report: percentage cloned=0%, amount cloned=306/0, files cloned=3/1
```

I am unable to understand what does this "MISSING" indicates in the log, can someone please help?

iPraveenParihar:
> I am unable to understand what does this "MISSING" indicates in the log, can someone please help?

`MISSING` in log messages is occurred when there is mismatch between format specifiers and the arguments provided.
 something, like fmt.Printf("key1: %s, key2: %s", "value1") should log `MISSING`. 
 
 In your PR, I don't see that to be the case 🤔.  You could verify that once with you latest PR changes?

Nikhil-Ladha:
> > I am unable to understand what does this "MISSING" indicates in the log, can someone please help?
> 
> `MISSING` in log messages is occurred when there is mismatch between format specifiers and the arguments provided. something, like fmt.Printf("key1: %s, key2: %s", "value1") should log `MISSING`.
> 
> In your PR, I don't see that to be the case 🤔. You could verify that once with you latest PR changes?

The above output is using the latest changes only.

internal/cephfs/store/fsjournal.go
```diff
@@ -128,6 +128,14 @@ func CheckVolExists(ctx context.Context,
 		}
 		err = cloneState.ToError()
 		if errors.Is(err, cerrors.ErrCloneInProgress) {
+			progressReport := cloneState.GetProgressReport()
+			err = fmt.Errorf("%w. progress report: percentage cloned=%s, amount cloned=%s, files cloned=%s",
```
Madhu-1: >progress report: percentage cloned=, amount cloned=, files cloned= 

We might need to put it behind some if-check to avoid this kind of log in unsupported version(it looks ugly).

Nikhil-Ladha: Added a check for it.

internal/cephfs/store/fsjournal.go
```diff
@@ -128,7 +128,20 @@ func CheckVolExists(ctx context.Context,
...
+				return nil, err
+			} else {
+				err = fmt.Errorf("%w. progress report: percentage cloned=%s, amount cloned=%s, files cloned=%s",
+					err,
+					progressReport.PercentageCloned,
+					progressReport.AmountCloned,
+					progressReport.FilesCloned)
+				log.UsefulLog(ctx, err.Error())
+
+				return nil, err
```
Madhu-1:
```suggestion
if progressReport.PercentageCloned == "" && progressReport.AmountCloned == "" &&
  progressReport.FilesCloned == "" {
     return nil, err
}
msg:="progress report:"
if progressReport.PercentageCloned != "" {
msg=fmt.Sprintf("%s percentage cloned=%s,progressReport.PercentageCloned)
}
if  progressReport.AmountCloned != "" {
msg=fmt.Sprintf("%s amount cloned=%s,progressReport.AmountCloned)
}
if progressReport.FilesCloned != "" {
msg=fmt.Sprintf("%s files cloned=%s=%s,progressReport.FilesCloned)
}
err = fmt.Errorf("%w. msg)
log.UsefulLog(ctx, err.Error())

return nil, err
```

can we do like this?

Nikhil-Ladha: I doubt that only one of them would ever be reported, it will either be all or none.
Even if the values are empty it would be "0" and not empty. So, I don't think we need something elaborative like this.

Madhu-1: we need to be sure on it , because its a string not int to show `0`,

Nikhil-Ladha: Yes, I have verified it. Even if the values are 0 (i.e, for empty directories, ), the output will be 0 wrapped in string.
Ref: https://github.com/ceph/ceph/blob/main/src/pybind/mgr/volumes/fs/stats_util.py#L26-L74

Madhu-1 (dismissed):

internal/cephfs/core/clone.go
```diff
@@ -28,16 +28,17 @@ import (
...
-	errorMsg string
+	state          admin.CloneState
+	progressReport admin.CloneProgressReport
+	errno          string
+	errorMsg       string
 }
 
 // CephFSCloneError indicates that fetching the clone state returned an error.
-var CephFSCloneError = cephFSCloneState{}
+var CephFSCloneError = &cephFSCloneState{}
```
Rakshith-R, line 38: Why is the change from value to reference required ?

Rakshith-R (commented):
just few small nits.
rest looks good to me.

internal/cephfs/core/clone.go
Nikhil-Ladha: Because adding a new struct member to the `cephFSCloneState` was making the size of the struct too big, the CI was complaining for the same [here](https://github.com/ceph/ceph-csi/actions/runs/11548533844/job/32140093334). That's the reason, we need to have this change.

Rakshith-R: Understood, thanks!

Rakshith-R (commented):
:facepalm: Sorry,
didn't click add review on this requested change.

Please take a look, small nit.

Nikhil-Ladha:
> 🤦 Sorry, didn't click add review on this requested change.
> 
> Please take a look, small nit.

I don't see the code for which you are suggesting the change, can you please re-highlight?

internal/cephfs/store/fsjournal.go
```diff
@@ -128,7 +128,20 @@ func CheckVolExists(ctx context.Context,
...
+			}
```
Rakshith-R:
```suggestion
			progressReport := cloneState.GetProgressReport()
			// append progress report only if all the parameters are present.
			if (progressReport.PercentageCloned == "" || progressReport.AmountCloned == "" ||
				progressReport.FilesCloned == "") {
				err = fmt.Errorf("%w. progress report: percentage cloned=%s, amount cloned=%s, files cloned=%s",
					err,
					progressReport.PercentageCloned,
					progressReport.AmountCloned,
					progressReport.FilesCloned)
			}
			log.ErrorLog(ctx, err.Error())

			return nil, err
```

Rakshith-R: @Nikhil-Ladha 

github should not allow me to comment if i have pending review comments  
:facepalm:
sorry for the confusion.

Nikhil-Ladha: I was wondering that if one is not empty so neither of them would be, so we can just add a check for `PercentageCloned` instead of all 3 params. WDYT @Madhu-1 @Rakshith-R ?

Rakshith-R: This new approach works for me.

Nikhil-Ladha: Updated the code accordingly.

Nikhil-Ladha: @Madhu-1 if we are good with this logic, can we go ahead and merge this?

Madhu-1 (dismissed):

iPraveenParihar (dismissed):
LGTM
Thanks!

internal/cephfs/store/fsjournal.go
```diff
@@ -128,7 +128,18 @@ func CheckVolExists(ctx context.Context,
...
+			if progressReport.PercentageCloned != "" {
+				err = fmt.Errorf("%w. progress report: percentage cloned=%s, amount cloned=%s, files cloned=%s",
+					err,
+					progressReport.PercentageCloned,
+					progressReport.AmountCloned,
+					progressReport.FilesCloned)
+				log.UsefulLog(ctx, err.Error())
+
+				return nil, err
+			}
```
Rakshith-R:
```suggestion
			}
			log.ErrorLog(ctx, err.Error())

			return nil, err
```

Rakshith-R (changes requested):

internal/cephfs/store/fsjournal.go
Nikhil-Ladha: It would be logically incorrect to term it as `error` here.
Since, there is no `error` as such with the clone, rather it is just `in progress` and should get complete in sometime. So, I don't think we should log it as an `error` log.

Rakshith-R: It is returned as a error. I would be fine with using debuglog, we don't use useful log anywhere else.
And your recent change no longer returns err if progress is missing. This needs to be changed as well. Observe the change in curly bracket

Nikhil-Ladha: Ah, right.
Updated the code now, please take a look.

Rakshith-R (approved):

iPraveenParihar (approved):

iPraveenParihar:
@Mergifyio rebase

iPraveenParihar:
@Mergifyio queue

PR #4941, opened 2024-11-04
rbd: check if mirror image has parent image

yati1998:
since enable mirroring fails if it has parent image this commit adds a check for that and returns
suitable error which can be picke by VR status

### DIFF
--- a/internal/rbd/mirror.go
+++ b/internal/rbd/mirror.go
@@ -91,6 +91,10 @@ func (ri *rbdImage) EnableMirroring(_ context.Context, mode librbd.ImageMirrorMo
 
 	err = image.MirrorEnable(mode)
 	if err != nil {
+		if ri.ParentName != "" {
+			return fmt.Errorf("failed to enable mirroring as image %q has a parent %q", ri.ImageID, ri.ParentName)
+		}
+
 		return fmt.Errorf("failed to enable mirroring on %q with error: %w", ri, err)
 	}
 


iPraveenParihar:
@yati1998, parent image check is already handled in `HandleParentImageExistence()`
https://github.com/ceph/ceph-csi/blob/c4a1e3a958002a13b1c8a15ad56c6f5195a3eaa7/internal/csi-addons/rbd/replication.go#L309-L322

yati1998:
> @yati1998, parent image check is already handled in `HandleParentImageExistence()`
> 
> https://github.com/ceph/ceph-csi/blob/c4a1e3a958002a13b1c8a15ad56c6f5195a3eaa7/internal/csi-addons/rbd/replication.go#L309-L322

okay, missed it, it seems we need to skip returning grpc error and instead return some meaningful error
cc @nixpanic any suggestions on how we can improvise the error?

internal/rbd/mirror.go
```diff
@@ -91,6 +91,10 @@ func (ri *rbdImage) EnableMirroring(_ context.Context, mode librbd.ImageMirrorMo
 
 	err = image.MirrorEnable(mode)
 	if err != nil {
+		if ri.ParentName != "" {
+			return fmt.Errorf("failed to enable mirroring as image %q has a parent %q", ri.ImageID, ri.ParentName)
```
nixpanic, line 95: don't forget to include the original error too

nixpanic:
> okay, missed it, it seems we need to skip returning grpc error and instead return some meaningful error

gRPC functions should return a gRPC error witch `codes.*` as a status and a message. These are sent over the gRPC protocol back to the caller. The message that is part of the gRPC error should contain as many details as posssible, but user understandable. In the end, these should land in a `.Status` entry.

yati1998:
I dont think this change is required as parent existence check it anyways verified.
Closing this PR

PR #4944, opened 2024-11-05
Implement GetFenceClients to return the ceph clusterID and the client Address

### LINKED ISSUE #4913: Add annotations with ceph client address

Add annotations with ceph client addresses to relevant k8s objects.

To reliably blocklist a client during a node split-brain, we need to stash the client's address somewhere, so that an operator (e.g. Rook) can eventually pick it up.

### How to get the client's address?
Depending on network setup, nodeplugin's node IP may be different than what ends up connecting to the Ceph cluster. [`rados_getaddrs`](https://docs.ceph.com/en/reef/rados/api/librados/#c.rados_getaddrs) (https://github.com/ceph/go-ceph/pull/1037) could be used to reliably retrieve the IP that Ceph sees.

### Where to store the annotation?
Annotating the nodeplugin's node should be sufficient if we're aiming for blocklisting the whole node.

### When to annotate?
Nodeplugin receives cluster credentials only when it receives the respective RPC calls, so we'll need to do it there. This unfortunately makes the driver Kuberentes aware.

### What to annotate?
To be figured out. In general we need a mapping between Ceph cluster ID and client IP. If the IP of clients on a single node is expected to change (e.g. what if the nodeplugin is deployed with `hostNetwork: false` ?), we may need to keep track of more IPs for one cluster ID entry.

Another option would be to keep track of the full entity address (IP + nonce) -- in that case it might make more sense to annotate the VolumeAttachment object instead?

Node fencing / client blocklisting can be performed reliably.

---
If nobody else is working on this topic I can send a PR once we agree on all the details mentioned above.

Madhu-1:
This PR implements a csiaddons RPC calls to return the ceph fsID and the client IP that is used to connect to the ceph cluster from the rbd plugin daemonset pod.

closes: #4913


Madhu-1:
# Test result

```bash
sh-5.1# csi-client 
Enter the path for the Unix socket: unix:///csi/csi-addons.sock
Using socket path: unix:///csi/csi-addons.sock
Server response: id:"a815fe8e-eabd-4e87-a6e8-78cebfb67d08"  addresses:{cidr:"10.244.0.1"}
sh-5.1# cat /sys/devices/rbd/0/
block/        client_id     config_info   features      major         name          pool          pool_ns       refresh       snap_id       uevent        
client_addr   cluster_fsid  current_snap  image_id      minor         parent        pool_id       power/        size          subsystem/    
sh-5.1# cat /sys/devices/rbd/0/client_addr 
10.244.0.1:0/353261872
sh-5.1# cat /sys/devices/rbd/0/cluster_fsid 
a815fe8e-eabd-4e87-a6e8-78cebfb67d08
sh-5.1# 
```

wrote  a small client to get the details, this is how it looks like

nixpanic:
`cidr:"10.244.0.1"` is not in correct CIDR notation, it is a plain IP-address. Maybe call it differently, or make it `cidr:"10.244.0.1/32"`.

Madhu-1:
> `cidr:"10.244.0.1"` is not in correct CIDR notation, it is a plain IP-address. Maybe call it differently, or make it `cidr:"10.244.0.1/32"`.

added `/32` to make it CIDR.

Rakshith-R (commented):
Cephfs nodeplugin needs this  NF server too ?

Madhu-1:
> hfs nodeplugin needs this NF server too

This is only for rbd now, we will extend it to cephfs later on if required.

Rakshith-R (approved):
Thanks,
LGTM

nixpanic (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

acolombier:
I'm getting this error when trying to run the E2E. I assume `librados-dev` needs to be at a certain version. Could we make the required version explicit [in the doc](https://github.com/ceph/ceph-csi/blob/d06497cf329db11309f966821b8727944b452d28/docs/development-guide.md#L27)?
```
# github.com/ceph/ceph-csi/internal/util
internal/util/connection.go:162:17: cc.conn.GetAddrs undefined (type *rados.Conn has no field or method GetAddrs)
FAIL    github.com/ceph/ceph-csi/e2e [build failed]
FAIL
```

Madhu-1:
> I'm getting this error when trying to run the E2E. I assume `librados-dev` needs to be at a certain version. Could we make the required version explicit [in the doc](https://github.com/ceph/ceph-csi/blob/d06497cf329db11309f966821b8727944b452d28/docs/development-guide.md#L27)?
> 
> ```
> # github.com/ceph/ceph-csi/internal/util
> internal/util/connection.go:162:17: cc.conn.GetAddrs undefined (type *rados.Conn has no field or method GetAddrs)
> FAIL    github.com/ceph/ceph-csi/e2e [build failed]
> FAIL
> ```

The error seems to be coming from go-ceph. can you please specify which exact command is failing.  https://docs.ceph.com/en/pacific/rados/api/librados/#c.rados_getaddrs is supported in older version as well. its not something new

PR #4946, opened 2024-11-06
rbd: cleanup inconsistent state in `reserveSnap()` after a failure

### LINKED ISSUE #4945: rbd: `reserveSnap()` should cleanup its state in case of a failure
`reserveSnap()` could set certain attributes of `rbdSnap`, but can fail later on. In case of a failure, attributes set in `rbdSnap` may not be consistent. The function should cleanup its state in case of a failure (`UndoReserveName()`), and only change rbdSnap attributes on success.

https://github.com/ceph/ceph-csi/blob/d5849a4801fa9f3383cefc4f96e18fcad420dc80/internal/rbd/rbd_journal.go#L403-L421

_Originally posted by @iPraveenParihar in https://github.com/ceph/ceph-csi/pull/4502#discussion_r1814796388_

nixpanic:
`reserveSnap()` can potentially fail halfway through, in that case it
needs to undo the snapshot reservation and restore modified attributes
of the snapshot.

Fixes: #4945

### DIFF
--- a/internal/rbd/rbd_journal.go
+++ b/internal/rbd/rbd_journal.go
@@ -388,6 +388,18 @@ func (ri *rbdImage) repairImageID(ctx context.Context, j *journal.Connection, fo
 func reserveSnap(ctx context.Context, rbdSnap *rbdSnapshot, rbdVol *rbdVolume, cr *util.Credentials) error {
 	var err error
 
+	// restore original values in case of an error
+	origReservedID := rbdSnap.ReservedID
+	origRbdSnapName := rbdSnap.RbdSnapName
+	origVolID := rbdSnap.VolID
+	defer func() {
+		if err != nil {
+			rbdSnap.ReservedID = origReservedID
+			rbdSnap.RbdSnapName = origRbdSnapName
+			rbdSnap.VolID = origVolID
+		}
+	}()
+
 	journalPoolID, imagePoolID, err := util.GetPoolIDs(ctx, rbdSnap.Monitors, rbdSnap.JournalPool, rbdSnap.Pool, cr)
 	if err != nil {
 		return err
@@ -405,6 +417,17 @@ func reserveSnap(ctx context.Context, rbdSnap *rbdSnapshot, rbdVol *rbdVolume, c
 		ctx, rbdSnap.JournalPool, journalPoolID, rbdSnap.Pool, imagePoolID,
 		rbdSnap.RequestName, rbdSnap.NamePrefix, rbdVol.RbdImageName, kmsID, rbdSnap.ReservedID, rbdVol.Owner,
 		"", encryptionType)
+	defer func() {
+		// only undo the reservation when an error occurred
+		if err == nil {
+			return
+		}
+
+		undoErr := undoSnapReservation(ctx, rbdSnap, cr)
+		if undoErr != nil {
+			log.WarningLog(ctx, "failed undoing reservation of snapshot %q: %v", rbdSnap, undoErr)
+		}
+	}()
 	if err != nil {
 		return err
 	}


internal/rbd/rbd_journal.go
```diff
@@ -388,6 +388,18 @@ func (ri *rbdImage) repairImageID(ctx context.Context, j *journal.Connection, fo
...
+	origReservedID := rbdSnap.ReservedID
+	origRbdSnapName := rbdSnap.RbdSnapName
+	origVolID := rbdSnap.VolID
+	defer func() {
+		if err != nil {
+			rbdSnap.ReservedID = origReservedID
+			rbdSnap.RbdSnapName = origRbdSnapName
+			rbdSnap.VolID = origVolID
+		}
+	}()
```
Madhu-1, line 401: @nixpanic any specific reason to preserve and restore these values? i assume if the reservation is done properly these values will be set and not overridden to any values isnt it?

nixpanic: there are these two calls that overwrite the attributes in any case (success of error):
```
rbdSnap.ReservedID, rbdSnap.RbdSnapName, err = j.ReserveName(...
```
and
```
rbdSnap.VolID, err = util.GenerateVolID(...
```
So, if something goes wrong, it looks nicer to reset the attributes that may have been modified by a failing call.

nixpanic: Setting those attributes only at the end of the function is tricky too, `undoSnapReservation()` expects them to be set in order to clean things up in the journal.

internal/rbd/rbd_journal.go
```diff
@@ -405,6 +417,17 @@ func reserveSnap(ctx context.Context, rbdSnap *rbdSnapshot, rbdVol *rbdVolume, c
 		ctx, rbdSnap.JournalPool, journalPoolID, rbdSnap.Pool, imagePoolID,
 		rbdSnap.RequestName, rbdSnap.NamePrefix, rbdVol.RbdImageName, kmsID, rbdSnap.ReservedID, rbdVol.Owner,
 		"", encryptionType)
+	defer func() {
+		// only undo the reservation when an error occurred
+		if err == nil {
+			return
+		}
+
+		undoErr := undoSnapReservation(ctx, rbdSnap, cr)
```
Rakshith-R, line 426: This should be and is already called outside after reserveSnap() calls ?

https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/controllerserver.go#L1194
https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/controllerserver.go#L1194

nixpanic: That is not sufficient in  case `reserveSnap()` had an internal error. That is the reason why this change was suggested in the 1st place. A failing function should cleanup its own internal partially-completed state on a failure, `reserveSnap()` did not do that yet.

nixpanic:
@Madhu-1 @Rakshith-R , any other concerns?

internal/rbd/rbd_journal.go
Madhu-1: @nixpanic the ReserveID and RbdSnapName names are not set earlier its only set by calling the `j.ReserveName` function, IMO we dont need to preserve anything and also  calling the defer incase of error should be enough.  same goes for volID as well.

Rakshith-R (approved):

Rakshith-R:
> @Madhu-1 @Rakshith-R , any other concerns?

LGTM.

internal/rbd/rbd_journal.go
nixpanic: @Madhu-1 if they are not expected to be set, shall I add a check and error message in case they are not an empty string?

Madhu-1: currently we dont have any case where they are set before calling this function. there are callers doing undo reservation and that will fail or not required

* Call Reserve to do reservation
* Undo if there is any error

we updated the current code to do undo internally if something fails.

Doesnt hurt to have this code but this will result in extra error/logs when we try to call the Undo function multiple times.


Let me know what do you prefer

1. Merge the current changes
2. Add a check to ensure the values are not set before calling `reserveSnap` internal function
3. Dont revert back and allow override.

nixpanic: The current PR as it is shows what a function should do/restore. There are no callers that will call `undoSnapReservation()` in case `reserveSnap()` fails, so there should not be additional logs for that.

When a function returns an error, there should be no need for the caller to undo whatever internals the function may, or may not, have done. Everything internal to the function should stay there. For the more complex cases, a special `undo` function can be returned, similar to what [`Manager.getGroupUUID()`](https://github.com/ceph/ceph-csi/blob/cea8bf8110531284d272582b9c536ffdb2ece82b/internal/rbd/manager.go#L117-L166) does.

Merging the current PR as it is should be fine IMHO, @Madhu-1, but I am also fine with improving it some way you recommend.

Madhu-1: https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/controllerserver.go#L1194
https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/controllerserver.go#L1194 might trigger an additional log

Madhu-1 (approved):

internal/rbd/rbd_journal.go
nixpanic: No, it would not. In case there is an error returned by `reserveSnap()`, the `defer` where `undoSnapReservation()` is called will never be reached.

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

PR #4949, opened 2024-11-11
rbd: add locking for VolumeGroupSnapshot operations

nixpanic:
Add VolumeGroupLocks in the CSI Controller Server so that operations are
protected against concurrent requests for the same VolumeGroupSnapshot.

### DIFF
--- a/internal/rbd/group_controllerserver.go
+++ b/internal/rbd/group_controllerserver.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/ceph/ceph-csi/internal/rbd/types"
+	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/log"
 )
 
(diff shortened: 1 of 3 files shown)


Rakshith-R (approved):

internal/rbd/group_controllerserver.go
```diff
@@ -166,15 +175,25 @@ func (cs *ControllerServer) DeleteVolumeGroupSnapshot(
 	// 1. verify that all snapshots in the request are all snapshots in the group
 	// 2. delete the group
 
+	groupSnapshotID := req.GetGroupSnapshotId()
```
Rakshith-R, line 178: https://github.com/ceph/ceph-csi/blob/8dadf26537bb6108c9a8a83ab8dda962cf56dc70/internal/rbd/group_controllerserver.go#L213

Get call needs to have locks too ? 
to prevent delete and get calls at the same time.

Rakshith-R (changes requested):

internal/rbd/group_controllerserver.go
nixpanic: Ah, right, I decided it would not be needed. But you're right, I'll add it there too.

Madhu-1 (approved):

Rakshith-R (approved):
Thanks !

Rakshith-R:
@mergifyio queue

PR #4970, opened 2024-11-20
Prevent dataloss due to the concurrent RPC calls (occurrence is very low)

### LINKED ISSUE #4966: Race condition in RBD node plugin may lead to total data loss while unmounting
# Describe the bug #

Recently, we encountered a total data loss in a few of our RBD volumes.

From the logs of the RBD nodeplugin, we inferred that the following condition occurred (all these steps can be correlated to the rbd-nodeplugin logs, attached to this issue):

- Attempted to bindmount volume from its staging path to its target path at `/var/lib/kubelet/pods/.../volumes/kubernetes.io~csi/.../mount`.

- We were seeing an initial `/csi.v1.Node/NodePublishVolume`, followed (for reasons unknown) by a _second_ `/csi.v1.Node/NodePublishVolume` call exactly two minutes after that. We're presuming that this was the CSI timeout[^csi-timeout] after which the kubelet attempts to mount the volume again.

- This causes the node plugin to bindmount the volume _again_ over the target volume, **WITHOUT** unmounting the previous bindmount.

- Shortly after, the volume is unmounted again. kubelet invokes a `/csi.v1.Node/NodeUnpublishVolume` call to the node plugin.

- The node plugin `unmount`s the volume, and calls `os.RemoveAll` on the target path. **HOWEVER**, since `mount` was invoked twice on that volume, there is still an active bind mount to the actual RBD volume in that directory, causing `os.RemoveAll` to recursively remove the entire contents of that volume.

- RIP :skull: 

# Environment details #

- Image/version of Ceph CSI driver : v3.12.2
- Helm chart version : 3.12.2
- Kernel version : 6.10.11+bpo-amd64 (debian bookworm-backports)
- Mounter used for mounting PVC (for cephFS its fuse or kernel. for rbd its krbd or rbd-nbd) : krbd
- Kubernetes cluster version : v1.30.5
- Ceph cluster version : v16.2.14

# Steps to reproduce #

1. Unknown. Assistance in further debugging might be required.

Presumably:

1. Schedule a Pod mounting an RBD volume
2. Have kubelet crash+restart shortly after mounting the volume, for example due to OOM
3. Terminate the Pod, causing kubelet to unmount (and involuntarily purge) the volume

# Actual results #

Total data loss in the RBD volume, due to the RBD node plugin running a recursive `os.RemoveAll` on a still-mounted volume.

# Expected behavior #

Mounting and unmounting (even under load and occasional error conditions) should work without data corruption.

# Logs #

Attached to this issue are our logs from the ceph-csi-rbd-nodeplugin from the time of the incident (newest entries at the top).

[Explore-logs-2024-11-18 19_34_03.txt](https://github.com/user-attachments/files/17814805/Explore-logs-2024-11-18.19_34_03.txt)

# Additional context #

## Quick&dirty workaround

We implemented a quick workaround by replacing the `os.RemoveAll` call with a regular `os.Remove`, which we can also provide as PR if required. Our reasoning for this was as follows:

- If the volume was correctly unmounted, the target path will be an empty directory, and a simple `os.Remove` call will suffice to remove it correctly.
- If the volume was _not_ correctly unmounted, the target path will still contain the volume's contents, and the `os.Remove` call will fail with an error -- which is very preferable to an unexpected data loss. In this case, we would expect the `NodeUnpublishVolume` call to be (eventually) retried and to (also eventually) succeed.

## "The kubelet should make sure that this does not happen"

From reading the code comments around the `os.RemoveAll` call in question, it seems to be presumed that the kubelet will guard against concurrent volume operations:

https://github.com/ceph/ceph-csi/blob/3e9b438e7c3a756dd4c6160a7c793d97ab92053f/internal/rbd/nodeserver.go#L917-L918

This does not seem to be the case. In fact, the CSI spec[^csi-spec] explicitly recognizes that (emphasis mine) "in some circumstances, the CO MAY lose state (for example when the CO crashes and restarts), and **MAY issue multiple calls simultaneously for the same volume.** The Plugin, SHOULD handle this as gracefully as possible, and MAY return this error code to reject secondary calls."

Furthermore, the CephFS node plugin seems to already follow this recommendation, and both checks if the target path is actually a mount point and only uses a simple `os.Remove` call instead of `os.RemoveAll` (although even that might be susceptible to race conditions):

https://github.com/ceph/ceph-csi/blob/3e9b438e7c3a756dd4c6160a7c793d97ab92053f/internal/cephfs/nodeserver.go#L642-L645

It should also be noted that the CephFS implementation also checks in its `NodePublishVolume` implementation if the target directory is already a mountpoint, and just returns early in this case (which would also prevent issues like these):

https://github.com/ceph/ceph-csi/blob/3e9b438e7c3a756dd4c6160a7c793d97ab92053f/internal/cephfs/nodeserver.go#L551-L562

## Suggested mitigations

- In `NodeUnpublishVolume`: Use `os.RemoveAll` _only_ (if at all) after asserting that the target path is _not_ still a mountpoint. Note that this alone might still be potentially racy.
- In `NodePublishVolume`: Test if the target path is already bindmounted to the staging path, and return early without mounting it again.
- To be entirely sure, it might also be advisable to add a (per-volume) locking mechanism protecting all volume operations, to ensure that no operations run in parallel.

---

Credits for discovering this issue and the associated debugging go to @hensur and @Lucaber.

[^csi-timeout]: https://github.com/kubernetes/kubernetes/blob/cf480a3a1a9cb22f3439c0a7922822d9f67f31b5/pkg/volume/csi/csi_plugin.go#L54
[^csi-spec]: https://github.com/container-storage-interface/spec/blob/master/spec.md#error-scheme

Madhu-1:
This PR includes series for commits for the following actions

* Introduce in memory lock for NodePublish and UnPublish (we should not be dependent on the CO and make assumption that CO makes the calls serial for the same volID and the targetPath)
* Use os.Remove instead of os.Removeall to remove the empty directory, os.Removeall ends up removing everything in the path

Thanks to @martin-helmich  @hensur and @Lucaber for the detailed analysis 

fixes: #4966

### DIFF
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -709,8 +709,12 @@ func (ns *NodeServer) NodePublishVolume(
 	volID := req.GetVolumeId()
 	stagingPath += "/" + volID
 
-	// Considering kubelet make sure the stage and publish operations
-	// are serialized, we dont need any extra locking in nodePublish
+	if acquired := ns.VolumeLocks.TryAcquire(targetPath); !acquired {
+		log.ErrorLog(ctx, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+
+		return nil, status.Errorf(codes.Aborted, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+	}
+	defer ns.VolumeLocks.Release(targetPath)
 
 	// Check if that target path exists properly
 	notMnt, err := ns.createTargetMountPath(ctx, targetPath, isBlock)
@@ -914,8 +918,14 @@ func (ns *NodeServer) NodeUnpublishVolume(
 	}
 
 	targetPath := req.GetTargetPath()
-	// considering kubelet make sure node operations like unpublish/unstage...etc can not be called
-	// at same time, an explicit locking at time of nodeunpublish is not required.
+
+	if acquired := ns.VolumeLocks.TryAcquire(targetPath); !acquired {
+		log.ErrorLog(ctx, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+
+		return nil, status.Errorf(codes.Aborted, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+	}
+	defer ns.VolumeLocks.Release(targetPath)
+
 	isMnt, err := ns.Mounter.IsMountPoint(targetPath)
 	if err != nil {
 		if os.IsNotExist(err) {
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -938,7 +938,7 @@ func (ns *NodeServer) NodeUnpublishVolume(
 		return nil, status.Error(codes.NotFound, err.Error())
 	}
 	if !isMnt {
-		if err = os.RemoveAll(targetPath); err != nil {
+		if err = os.Remove(targetPath); err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 
@@ -949,7 +949,7 @@ func (ns *NodeServer) NodeUnpublishVolume(
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if err = os.RemoveAll(targetPath); err != nil {
+	if err = os.Remove(targetPath); err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -494,6 +494,13 @@ func (ns *NodeServer) NodePublishVolume(
 	targetPath := req.GetTargetPath()
 	volID := fsutil.VolumeID(req.GetVolumeId())
 
+	if acquired := ns.VolumeLocks.TryAcquire(targetPath); !acquired {
+		log.ErrorLog(ctx, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+
+		return nil, status.Errorf(codes.Aborted, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+	}
+	defer ns.VolumeLocks.Release(targetPath)
+
 	volOptions := &store.VolumeOptions{}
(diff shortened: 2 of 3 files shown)


internal/rbd/nodeserver.go
```diff
@@ -709,8 +709,12 @@ func (ns *NodeServer) NodePublishVolume(
...
-	// are serialized, we dont need any extra locking in nodePublish
+	if acquired := ns.VolumeLocks.TryAcquire(targetPath); !acquired {
+		log.ErrorLog(ctx, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+
+		return nil, status.Errorf(codes.Aborted, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+	}
+	defer ns.VolumeLocks.Release(targetPath)
 
 	// Check if that target path exists properly
 	notMnt, err := ns.createTargetMountPath(ctx, targetPath, isBlock)
```
martin-helmich, line 720: If I see it correctly, `createTargetMountPath` still uses the (possibly unreliable?) `IsLikelyNotMountPoint` under the hood; should this be adjusted to the same `IsMountPoint` that the cephfs implementation uses? Or can this remain as-is, as discussed in https://github.com/ceph/ceph-csi/issues/4966#issuecomment-2488145516?

martin-helmich (approved):
Wow, that was quick, thanks. 🤩🚀  This is also pretty much the fix that we had in mind for this. 👍

internal/rbd/nodeserver.go
Madhu-1: Yes thats correct, it checks the sources, The main problem is with lock and the os.RemoveAll, if os.Remove fails we will get another RPC for unpublish which will trigger `umount` if its actually a mount.  For now lets keep it simple with what we  have today.

Rakshith-R (approved):

nixpanic (approved):
LGTM.

It would be good to backport this to active releases, I think?

Madhu-1:
@mergifyio queue

PR #4996, opened 2024-12-03
rbd: Use assume_storage_prezeroed when formatting

### LINKED ISSUE #4948: Pass assume_storage_prezeroed option when formatting ext4 if possible

Instead of passing `lazy_itable_init=1` and `lazy_journal_init=1` to `mkfs.ext4`, Ceph CSI should pass `assume_storage_prezeroed=1` which is stronger and allows the filesystem to skip inode table zeroing completely instead of simply doing it lazily (it's a superset of `lazy_itable_init=1` and `lazy_journal_init=1`).  As before with `lazy_itable_init=1` and `lazy_journal_init=1`, this should be limited to dynamically provisioned volumes -- a case where Ceph CSI can guarantee that `mkfs.ext4` invocation immediately follows the creation of the RBD image.

`assume_storage_prezeroed` option became available in e2fsprogs 1.47.0 last year, so it should probably be "discovered" similar to `xfsSupportsReflink()`.

Freshly created RBD volumes would consume less space.  Quoting `assume_storage_prezeroed` implementation patch:

```
    - Avoiding zeroing out the inode table and journal reduces the
      initial metadata space allocation from 0.48% to 0.01%.
    - Lazy inode table zeroing results in a further 1.45% of logical
      volume space getting allocated for inode tables, even if no file
      data is added to the filesystem. With assume_storage_prezeroed,
      the metadata allocation remains at 0.01%.
```

Run before and after tests that would create and attach batches of e.g. 100G, 500G, 1T and 5T RBD volumes, noting space usage as reported by `ceph df`.  The volumes would need to be attached for a while to allow for lazy inode table zeroing to complete in the before case.

black-dragon74:

Instead of passing `lazy_itable_init=1` and `lazy_journal_init=1` to
`mkfs.ext4`, pass `assume_storage_prezeroed=1` which is
stronger and allows the filesystem to skip inode table zeroing
completely instead of simply doing it lazily.

The support for this flag is checked by trying to format a fake
temporary image with `mkfs.ext4` and checking its `STDERR`.

Closes: #4948

### DIFF
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -26,6 +26,7 @@ import (
 
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
 	"github.com/ceph/ceph-csi/internal/util"
+	"github.com/ceph/ceph-csi/internal/util/file"
 	"github.com/ceph/ceph-csi/internal/util/fscrypt"
 	"github.com/ceph/ceph-csi/internal/util/log"
 
@@ -45,6 +46,11 @@ type NodeServer struct {
 	// A map storing all volumes with ongoing operations so that additional operations
 	// for that same volume (as defined by VolumeID) return an Aborted error
 	VolumeLocks *util.VolumeLocks
+
+	// ext4HasPrezeroedSupport indicates whether the ext4 filesystem has support for pre-zeroed blocks.
+	ext4HasPrezeroedSupport featureFlag
+	// xfsHasReflinkSupport indicates whether the xfs filesystem has support for reflink.
+	xfsHasReflinkSupport featureFlag
 }
 
 // stageTransaction struct represents the state a transaction was when it either completed
@@ -61,11 +67,16 @@ type stageTransaction struct {
 	devicePath string
 }
 
+// featureFlag represents a type for defining feature flags within the RBD node server.
+type featureFlag int
+
 const (
-	// values for xfsHasReflink.
-	xfsReflinkUnset int = iota
-	xfsReflinkNoSupport
-	xfsReflinkSupport
+	// featureFlagUnset represents the default value for a feature flag.
+	featureFlagUnset featureFlag = iota
+	// featureFlagNoSupport represents the value for a feature flag when the feature is not supported.
+	featureFlagNoSupport
+	// featureFlagSupport represents the value for a feature flag when the feature is supported.
+	featureFlagSupport
 
 	staticVol        = "staticVolume"
 	volHealerCtx     = "volumeHealerContext"
@@ -94,15 +105,6 @@ var (
 			Distribution: ".el8",
 			Backport:     true,
 		}, // RHEL 8.2
-	}
-
-	// xfsHasReflink is set by xfsSupportsReflink(), use the function when
-	// checking the support for reflink.
-	xfsHasReflink = xfsReflinkUnset
-
-	mkfsDefaultArgs = map[string][]string{
-		"ext4": {"-m0", "-Enodiscard,lazy_itable_init=1,lazy_journal_init=1"},
-		"xfs":  {"-K"},
 	}
 
 	mountDefaultOpts = map[string][]string{
@@ -793,7 +795,7 @@ func (ns *NodeServer) mountVolumeToStagePath(
 	}
 
 	if existingFormat == "" && !staticVol && !readOnly && !isBlock {
-		args := mkfsDefaultArgs[fsType]
+		args := ns.getMkfsArgs(fsType)
 
 		// if the VolumeContext contains "mkfsOptions", use those as args instead
 		volumeCtx := req.GetVolumeContext()
@@ -1292,8 +1294,8 @@ func (ns *NodeServer) processEncryptedDevice(
 // argument. In case it is supported, return true.
 func (ns *NodeServer) xfsSupportsReflink() bool {
 	// return cached value, if set
-	if xfsHasReflink != xfsReflinkUnset {
-		return xfsHasReflink == xfsReflinkSupport
+	if ns.xfsHasReflinkSupport != featureFlagUnset {
+		return ns.xfsHasReflinkSupport == featureFlagSupport
 	}
 
 	// run mkfs.xfs in the same namespace as formatting would be done in
@@ -1303,15 +1305,63 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
 	if err != nil {
 		// mkfs.xfs should fail with an error message (and help text)
 		if strings.Contains(string(out), "reflink=0|1") {
-			xfsHasReflink = xfsReflinkSupport
+			ns.xfsHasReflinkSupport = featureFlagSupport
 
 			return true
 		}
 	}
 
-	xfsHasReflink = xfsReflinkNoSupport
+	ns.xfsHasReflinkSupport = featureFlagNoSupport
 
 	return false
+}
+
+// ext4SupportsPrezeroed checks if the ext4 filesystem supports the
+// "assume_storage_prezeroed" option. It does this by creating a temporary
+// image file, attempting to format it with the ext4 filesystem using the
+// "assume_storage_prezeroed" option, and checking the output for errors.
+func (ns *NodeServer) ext4SupportsPrezeroed() bool {
+	if ns.ext4HasPrezeroedSupport != featureFlagUnset {
+		return ns.ext4HasPrezeroedSupport == featureFlagSupport
+	}
+
+	ctx := context.TODO()
+	tempImgFile, err := os.CreateTemp(os.TempDir(), "prezeroed.img")
+	if err != nil {
+		log.WarningLog(ctx, "failed to create temporary image file: %v", err)
+
+		return false
+	}
+	defer os.Remove(tempImgFile.Name())
+
+	if err = file.CreateSparseFile(tempImgFile, 1); err != nil {
+		log.WarningLog(ctx, "failed to create sparse file: %v", err)
+
+		return false
+	}
+
+	diskMounter := &mount.SafeFormatAndMount{Interface: ns.Mounter, Exec: utilexec.New()}
+
+	// `-n` Causes mke2fs to not actually create a file system
+	// but display what it would do if it were to create a file system.
+	out, err := diskMounter.Exec.Command(
+		"mkfs.ext4",
+		"-n",
+		"-Eassume_storage_prezeroed=1",
+		tempImgFile.Name(),
+	).CombinedOutput()
+	if err != nil {
+		if strings.Contains(string(out), "Bad option(s) specified") {
+			ns.ext4HasPrezeroedSupport = featureFlagNoSupport
+		}
+
+		// if the error is not due to the missing option, we can't be sure
+		// if the option is supported or not, return false optimistically
+		return false
+	}
+	ns.ext4HasPrezeroedSupport = featureFlagSupport
+
+	return true
 }
 
 // NodeGetVolumeStats returns volume stats.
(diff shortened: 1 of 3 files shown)


internal/rbd/nodeserver.go
```diff
@@ -101,7 +101,7 @@ var (
 	xfsHasReflink = xfsReflinkUnset
 
 	mkfsDefaultArgs = map[string][]string{
-		"ext4": {"-m0", "-Enodiscard,lazy_itable_init=1,lazy_journal_init=1"},
+		"ext4": {"-m0", "-Enodiscard,assume_storage_prezeroed=1"},
```
nixpanic: make sure that you only do this when `mkfs.ext4` supports the option, you'll need to add some form of detection, ideally with something like `sync.Once`.

idryomov: > `assume_storage_prezeroed` option became available in e2fsprogs 1.47.0 last year, so it should probably be "discovered" similar to `xfsSupportsReflink()`.

My suggestion to do it similar to `xfsSupportsReflink()` isn't viable because `mkfs.ext4` doesn't include extended options in its help output, but trying with `assume_storage_prezeroed=1` and falling back to `lazy_itable_init=1,lazy_journal_init=1` upon parsing out `Bad option(s) specified: assume_storage_prezeroed` error should work.

black-dragon74: I and Madhu had a discussion on this. Since this flag is part of e2fsprogs >= 1.47 and we have control over which version of it we bundle in the container, do we really need this check?

Sidenote: Detection is possible by parsing the semver of e2fsprogs. It is standard flag so it must be there?

idryomov: The check for the version isn't always reliable in the face of downstream backports.

Madhu-1: as @idryomov mentioned we cannot depend on the version check, as the base image includes the required version, we could say what is supported by this version but that could be a problem if someone wants to build/test custom images. let's depend on the error check and fall back that works for all the cases.

internal/rbd/nodeserver.go
```diff
@@ -67,6 +67,10 @@ const (
 	xfsReflinkNoSupport
 	xfsReflinkSupport
 
+	ext4PrezeroedUnset
```
nixpanic: restart counting again, use `ext4PrezeroedUnset = itoa`

internal/rbd/nodeserver.go
```diff
@@ -748,6 +756,7 @@ func (ns *NodeServer) NodePublishVolume(
 	return &csi.NodePublishVolumeResponse{}, nil
 }
 
+//nolint:cyclop,gocyclo // function is required to have multiple conditional checks
```
nixpanic: This is rather ugly. Maybe it is not required if you add a function like
```go
args := ns.GetMKFSDefaultArgs(fsType)
```
a function like that can add the `assume_storage_prezeroed=1` option if support is detected.

internal/rbd/nodeserver.go
```diff
@@ -792,8 +801,12 @@ func (ns *NodeServer) mountVolumeToStagePath(
 		readOnly = true
 	}
 
+	//nolint:nestif // readability
 	if existingFormat == "" && !staticVol && !readOnly && !isBlock {
 		args := mkfsDefaultArgs[fsType]
+		if fsType == "ext4" && ns.ext4SupportsPrezeroed() {
+			args = []string{"-m0", "-Enodiscard,assume_storage_prezeroed=1"}
```
nixpanic: this drops the default options from line 112, is that intentional?

internal/rbd/nodeserver.go
```diff
@@ -1314,6 +1327,50 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
+// ext4SupportsPrezeroed checks if the ext4 filesystem supports the
+// "assume_storage_prezeroed" option. It does this by creating a temporary
+// image file, attempting to format it with the ext4 filesystem using the
+// "assume_storage_prezeroed" option, and checking the output for errors.
+func (ns *NodeServer) ext4SupportsPrezeroed() bool {
+	if ext4HasPrezeroedSupport != ext4PrezeroedUnset {
+		return ext4HasPrezeroedSupport == ext4PrezeroedSupport
+	}
+
+	tempImage := "/tmp/prezeroed.img"
```
nixpanic: use a temporary filename, please

internal/rbd/nodeserver.go
```diff
@@ -1314,6 +1327,50 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
+	ctx := context.TODO()
+	_, _, err := util.ExecCommand(ctx, "dd", "if=/dev/zero", "of="+tempImage, "bs=1M", "count=1")
```
nixpanic: Instead of executing a `dd` command, open the file, seek to 1mb and write a singled zero byte. This makes a sparse file, which should be much quicker and can easily be done with plain Go.

nixpanic (changes requested):

internal/rbd/nodeserver.go
black-dragon74: go doesn't consider `iota` twice in the same const block :(

internal/rbd/nodeserver.go
black-dragon74: Yes, `lazy_itable_init=1,lazy_journal_init=1` are not required when using `assume_storage_prezeroed=1`. Rest of the options (`-m0, -Enodiscard`) from line 112 are included.

internal/rbd/nodeserver.go
black-dragon74: Thank you for this insight :)

internal/rbd/nodeserver.go
black-dragon74: Done

internal/rbd/nodeserver.go
```diff
@@ -1397,3 +1449,42 @@ func getDeviceSize(ctx context.Context, devicePath string) (uint64, error) {
...
+		return []string{"-K"}
+	default:
+		log.WarningLogMsg(fmt.Sprintf("unknown fsType: %q, using default mkfs options", fsType))
+
+		return []string{}
+	}
+}
+
+// createSpareFile makes `file` a sparse file of size `sizeMB`.
+func createSparseFile(file *os.File, sizeMB int64) error {
```
nixpanic: this is not specific to the nodeserver, move it to the internal/util package please, and add a unit-test for it

internal/rbd/nodeserver.go
```diff
@@ -1314,6 +1317,55 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
+	}
+
+	diskMounter := &mount.SafeFormatAndMount{Interface: ns.Mounter, Exec: utilexec.New()}
+
+	// `-n` Causes mke2fs to not actually create a file system
+	// but display what it would do if it were to create a file system.
+	out, err := diskMounter.Exec.Command(
+		"mkfs.ext4",
+		"-n",
+		"-Enodiscard,assume_storage_prezeroed=1",
```
nixpanic: I think it is nice to have the options mentioned as a const somewhere, they are here, but also in `getMkfsArgs()`.

Maybe keep the default map, and use `ext4.prezeroed` as fstype to get the options?

internal/rbd/nodeserver.go
```diff
@@ -100,10 +104,9 @@ var (
 	// checking the support for reflink.
 	xfsHasReflink = xfsReflinkUnset
 
-	mkfsDefaultArgs = map[string][]string{
-		"ext4": {"-m0", "-Enodiscard,lazy_itable_init=1,lazy_journal_init=1"},
-		"xfs":  {"-K"},
-	}
+	// ext4HasPrezeroed is set by ext4SupportsPrezeroed(), use the function when
+	// checking the support for assume_storage_prezeroed.
+	ext4HasPrezeroedSupport = ext4PrezeroedUnset
```
nixpanic: I understand you copied the `xfsHasReflink` idea. But I think it is cleaner if the `NodeServer` struct has a field for these kind of detectable parameters. Have you thought about doing it that way?

nixpanic (commented):
looks good, just a few nits and potential improvements

internal/rbd/nodeserver.go
```diff
@@ -1314,6 +1317,55 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
 	return false
 }
 
+// ext4SupportsPrezeroed checks if the ext4 filesystem supports the
+// "assume_storage_prezeroed" option. It does this by creating a temporary
+// image file, attempting to format it with the ext4 filesystem using the
+// "assume_storage_prezeroed" option, and checking the output for errors.
+func (ns *NodeServer) ext4SupportsPrezeroed() bool {
```
Madhu-1, line 1323: how about if we check if this is supported or not in the driver.go(if we are stating it as nodeplugin) before starting the node plugin and store it in the Nodeserver struct like its suggested here https://github.com/ceph/ceph-csi/pull/4996/files#r1910308642?

internal/rbd/nodeserver.go
idryomov: > I think it is nice to have the options mentioned as a const somewhere, they are here

I don't have an opinion about it being expressed as a const, but I think it should be a single option.  This function is supposed to be checking whether ext4 supports "prezeroed", so it should be just that -- pass only `assume_storage_prezeroed=1`, `nodiscard` isn't relevant here.

internal/rbd/nodeserver.go
black-dragon74: I refactored both the XFS and ext4 detections to use struct fields. PTAL.

internal/rbd/nodeserver.go
black-dragon74: Removed the `nodiscard` flag.

internal/rbd/nodeserver.go
black-dragon74: As the result for this is cached, let's check it when we need it instead of doing it preemptively? This way also has the added benefit of keeping `driver.go` simple. WDYT?

internal/rbd/nodeserver.go
```diff
@@ -1303,17 +1305,65 @@ func (ns *NodeServer) xfsSupportsReflink() bool {
...
+	if ns.ext4HasPrezeroedSupport != featureFlagUnset {
+		return ns.ext4HasPrezeroedSupport == featureFlagSupport
+	}
+
+	ctx := context.TODO()
+	tempImgFile, err := os.CreateTemp(os.TempDir(), "prezeroed.img")
```
nixpanic, line 1329: Although unlikely, it is possible that this runs more than once at the same time. This is eventually called while attaching+mounting the RBD-image.

It is possible that the file is created and then:
- more than one thread passes here
- one thread finishes the function and deletes the file
- other threads continue to execute the `mkfs.ext4` command, with the file missing
- the other threads will fail with an unexpected error

The `prezeroed.img` filename should be unique, use `os.CreateTemp()`. Or you can use an approach that builds on `sync.Once`.

internal/util/file/file_test.go
```diff
@@ -98,3 +98,51 @@ func TestCreateTempFile_WithLargeContent(t *testing.T) {
...
+		{"WithValidSize", 10, false},
+		{"WithZeroSize", 0, true},
+		{"WithNegativeSize", -1, true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			file, err := os.CreateTemp("", "test-sparse-")
```
nixpanic: place this in the `t.TempDir()` directory for automatic cleanup, no need to have to remove the file in a `defer`.

internal/rbd/nodeserver.go
black-dragon74: This is just a suffix. Each filename will be different.

internal/rbd/nodeserver.go
nixpanic: I think this is sufficient for now. It could be placed in some `internal/util/...` package, and keep a single initialization with `sync.Once` there. That could be an improvement, but I don't think it is really needed at the moment.

internal/util/file/file_test.go
nixpanic: Rebased and fixed this nit for you :-)

nixpanic (approved):

Madhu-1 (approved):

nixpanic:
@mergifyio queue

PR #5001, opened 2024-12-04
rbd: return group not found error for Get,Delete RPC calls

### LINKED ISSUE #4999: ControllerGetVolumeGroup and DeleteVolumeGroup is not idempotent
# Describe the bug #
Ideally, `ControllerGetVolumeGroup` and `DeleteVolumeGroup` should be idempotent i.e, if the volume group is not found in the backend then the `ControllerVolumeGroup` shall return status code as `NotFound` and if the volume group has been already deleted from the backend then the `DeleteVolumeGroup` shall return the response as `OK`. But, either of them is not the case now.

The `ControllerVolumeGroup` RPC call returns `OK` response with the volume group id , even if the volume group doesn't exists. And, the `DeleteVolumeGroup` RPC call returns `BUG: name not set` error while trying to delete an already deleted volume group.

# Steps to reproduce #
1. Use the csi-addons cli and try to fetch an non-existing volume group
2. Use the csi-addons cli and try to delete an already deleted volume group

# Actual results #
1. A non-existent volume group is returned with an OK reponse.
2. An already deleted volume group fails with an error.

# Expected behavior #
1. A non-existent volume group should return a `NotFound` status code.
2. An already deleted volume group should return an OK response.

Nikhil-Ladha:
We should return `NotFound` GRPC status if the group doesn't exists for ControllerGetVolumeGroup RPC call.
And, an empty/OK response for DeleteVolumeGroup if the group doesn't exists

Fixes: #4999 

---

<details>

</details>

### DIFF
--- a/internal/csi-addons/rbd/volumegroup.go
+++ b/internal/csi-addons/rbd/volumegroup.go
@@ -18,10 +18,12 @@ package rbd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 
 	"github.com/ceph/ceph-csi/internal/rbd"
+	"github.com/ceph/ceph-csi/internal/rbd/group"
 	"github.com/ceph/ceph-csi/internal/rbd/types"
 	"github.com/ceph/ceph-csi/internal/util/log"
 
@@ -192,9 +194,15 @@ func (vs *VolumeGroupServer) DeleteVolumeGroup(
 	// resolve the volume group
 	vg, err := mgr.GetVolumeGroupByID(ctx, req.GetVolumeGroupId())
 	if err != nil {
+		if errors.Is(err, group.ErrRBDGroupNotFound) {
+			log.DebugLog(ctx, "VolumeGroup %q doesn't exists", req.GetVolumeGroupId())
+
+			return &volumegroup.DeleteVolumeGroupResponse{}, nil
+		}
+
 		return nil, status.Errorf(
-			codes.NotFound,
-			"could not find volume group %q: %s",
+			codes.Internal,
+			"could not fetch volume group %q: %s",
 			req.GetVolumeGroupId(),
 			err.Error())
 	}
--- a/internal/rbd/group/util.go
+++ b/internal/rbd/group/util.go
@@ -145,6 +145,12 @@ func (cvg *commonVolumeGroup) getVolumeGroupAttributes(ctx context.Context) (*jo
 		attrs = &journal.VolumeGroupAttributes{}
 	}
 
+	if attrs.GroupName == "" || attrs.CreationTime == nil {
+		log.ErrorLog(ctx, "volume group with id %v not found", cvg.id)
+
+		return nil, ErrRBDGroupNotFound
+	}
+
 	cvg.requestName = attrs.RequestName
 	cvg.name = attrs.GroupName
 	cvg.creationTime = attrs.CreationTime
--- a/internal/rbd/group/volume_group.go
+++ b/internal/rbd/group/volume_group.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	"github.com/ceph/go-ceph/rados"
+	librados "github.com/ceph/go-ceph/rados"
 	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/csi-addons/spec/lib/go/volumegroup"
@@ -31,7 +32,10 @@ import (
 	"github.com/ceph/ceph-csi/internal/util/log"
 )
 
-var ErrRBDGroupNotConnected = errors.New("RBD group is not connected")
+var (
+	ErrRBDGroupNotConnected = fmt.Errorf("%w: RBD group is not connected", librados.ErrNotConnected)
+	ErrRBDGroupNotFound     = fmt.Errorf("%w: RBD group not found", librbd.ErrNotFound)
+)
 
 // volumeGroup handles all requests for 'rbd group' operations.
 type volumeGroup struct {


internal/csi-addons/rbd/volumegroup.go
```diff
@@ -192,6 +194,12 @@ func (vs *VolumeGroupServer) DeleteVolumeGroup(
 	// resolve the volume group
 	vg, err := mgr.GetVolumeGroupByID(ctx, req.GetVolumeGroupId())
 	if err != nil {
+		if errors.Is(err, group.ErrRBDGroupNotFound) {
```
iPraveenParihar, line 197: same error check needs to be added for `/ControllerGetVolumeGroup`?

internal/csi-addons/rbd/volumegroup.go
```diff
@@ -192,6 +194,12 @@ func (vs *VolumeGroupServer) DeleteVolumeGroup(
...
+			log.DebugLog(ctx, "VolumeGroup %q doesn't exists", req.GetVolumeGroupId())
+
+			return &volumegroup.DeleteVolumeGroupResponse{}, nil
+		}
+
 		return nil, status.Errorf(
```
iPraveenParihar, line 203: Can you modify this error message? It also says could not find volume group.

internal/rbd/group/volume_group.go
```diff
@@ -82,6 +85,12 @@ func GetVolumeGroup(
 		return nil, fmt.Errorf("failed to get volume attributes for id %q: %w", vg, err)
 	}
 
+	if attrs.GroupName == "" || attrs.CreationTime == nil {
```
iPraveenParihar: we could make use of `vg.GetName()` & `vg.GetCreationTime()` ?

iPraveenParihar (dismissed):
Please paste the testing logs.

internal/csi-addons/rbd/volumegroup.go
Nikhil-Ladha: We are already returning `NotFound` status code in case of an error there, so no update needed for ControllerGetVolumeGroup RPC.

internal/rbd/group/volume_group.go
Nikhil-Ladha: Right, we can do that.
Updated the code to use only `vg.GetName`

internal/csi-addons/rbd/volumegroup.go
Nikhil-Ladha: Done

Nikhil-Ladha:
> Please paste the testing logs.

```
[nladha@dhcp53-204 ~]$ kubectl exec -it deployments/csi-rbdplugin-provisioner -c csi-addons -n rook-ceph -- csi-addons --endpoint unix:///csi/csi-addons.sock -operation ControllerGetVolumeGroup -volumegroupid 0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa -secret rook-ceph/rook-csi-rbd-provisioner
E1204 08:42:48.273772      11 volumegroup.go:174] Failed to get volume group: rpc error: code = NotFound desc = could not find volume group "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa": failed to get volume group with id "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa": RBD group not found
ERROR: failed to execute "ControllerGetVolumeGroup": rpc error: code = NotFound desc = could not find volume group "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa": failed to get volume group with id "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa": RBD group not found
command terminated with exit code 1
[nladha@dhcp53-204 ~]$ kubectl exec -it deployments/csi-rbdplugin-provisioner -c csi-addons -n rook-ceph -- csi-addons --endpoint unix:///csi/csi-addons.sock -operation DeleteVolumeGroup -volumegroupid 0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa -secret rook-ceph/rook-csi-rbd-provisioner
Volume Group Deleted.
```

csi-rbdplugin-provisioner logs
```
I1204 08:42:48.177599       1 utils.go:266] ID: 24 GRPC call: /volumegroup.Controller/ControllerGetVolumeGroup
I1204 08:42:48.180222       1 utils.go:267] ID: 24 GRPC request: {"secrets":"***stripped***","volume_group_id":"0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa"}
I1204 08:42:48.260968       1 util.go:102] ID: 24 object for volume group "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa" has been initialized
E1204 08:42:48.270666       1 omap.go:212] ID: 24 omap not found (pool="replicapool", namespace="", name="csi.volume.group.71f1dc66-e334-4b9b-9f7c-536f4c3f03fa"): rados: ret=-2, No such file or directory
W1204 08:42:48.271027       1 volumegroupjournal.go:409] ID: 24 unable to read omap values: pool missing: key not found: rados: ret=-2, No such file or directory
E1204 08:42:48.271583       1 volume_group.go:89] ID: 24 volume group with id 0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa not found
E1204 08:42:48.271747       1 utils.go:271] ID: 24 GRPC error: rpc error: code = NotFound desc = could not find volume group "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa": failed to get volume group with id "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa": RBD group not found
I1204 08:43:29.919432       1 utils.go:266] ID: 25 GRPC call: /volumegroup.Controller/DeleteVolumeGroup
I1204 08:43:29.920103       1 utils.go:267] ID: 25 GRPC request: {"secrets":"***stripped***","volume_group_id":"0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa"}
I1204 08:43:29.921395       1 util.go:102] ID: 25 object for volume group "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa" has been initialized
E1204 08:43:29.924151       1 omap.go:212] ID: 25 omap not found (pool="replicapool", namespace="", name="csi.volume.group.71f1dc66-e334-4b9b-9f7c-536f4c3f03fa"): rados: ret=-2, No such file or directory
W1204 08:43:29.924554       1 volumegroupjournal.go:409] ID: 25 unable to read omap values: pool missing: key not found: rados: ret=-2, No such file or directory
E1204 08:43:29.924648       1 volume_group.go:89] ID: 25 volume group with id 0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa not found
I1204 08:43:29.925068       1 volumegroup.go:194] ID: 25 VolumeGroup "0001-0009-rook-ceph-0000000000000002-71f1dc66-e334-4b9b-9f7c-536f4c3f03fa" doesn't exists
I1204 08:43:29.925668       1 utils.go:273] ID: 25 GRPC response: null
```

internal/csi-addons/rbd/volumegroup.go
Madhu-1: If not found it should be a success response.

internal/rbd/group/volume_group.go
```diff
@@ -82,6 +85,12 @@ func GetVolumeGroup(
 		return nil, fmt.Errorf("failed to get volume attributes for id %q: %w", vg, err)
 	}
 
+	if _, err := vg.GetName(ctx); err != nil {
```
Madhu-1: This doesnt seems to be the right fix. i see that `getVolumeGroupAttributes` we are comparing with `librbd.ErrNotFound` i dont think it ever returns librbd error message, either it returns the pooNotfound or the key not found error message, which need to be returned and the caller need to check for it like we are doing in other delete volume or delete snapshot RPC call

internal/csi-addons/rbd/volumegroup.go
Nikhil-Ladha: In ControllerGetVolumeGroup, if the group doesn't exists, we should return NotFound only, right?
Else, how would we know that the volume group doesn't exists?

internal/rbd/group/volume_group.go
Nikhil-Ladha: I don't think it ever returns either of the 2 errors as well. We just log in case either of these 2 errors are found, and continue to update the attributes struct.
Can you please link the other delete RPC call where we are actually doing it that way?

Nikhil-Ladha: So, we return PoolNotFound error if we fail to get the pool, i.e, it is not found in the cluster based on the value in the param.
But, I don't see that we are actually returning the `ErrKeyNotFound` error, apart from the `FetchAttribute` function that is used in the nfs controller.
So, not sure if I am missing something in the design aspect or is this actually a bug?

internal/csi-addons/rbd/volumegroup.go
Madhu-1: Yes correct in ControllerGetVolumeGroup its notFound error but in DeleteVolumeGroup it should be success, this current change is for DeleteVolumeGroup RPC call isnt it?

Nikhil-Ladha: Yes, and we are returning a success reponse only.
`return nil, status.Errorf(codes.OK, "volume group %q doesn't exists", req.GetVolumeGroupId())`
The `codes.OK` basically, leads to a `nil` error only.

internal/rbd/group/volume_group.go
Madhu-1: Yes thats correct, i missed it but there are couple of things still bothering me

1. The `errors.Is(err, librbd.ErrNotFound) ` check above is never going to get hit.
2. The `GetName` returns error message as bug which is not the case as its expected.
3. https://github.com/ceph/ceph-csi/blob/a32ba13045f4f433d4c00f568c67209556428a76/internal/csi-addons/rbd/volumegroup.go#L193-L200 this is wrong as we cannot assume the error as not found

internal/csi-addons/rbd/volumegroup.go
Madhu-1: Yes but we need to follow similar flow what we are following for other Delete operation, we return emptyResp,nil for better readability `&volumegroup.DeleteVolumeGroupResponse{}, nil` like its returned at the end.

internal/rbd/group/volume_group.go
Nikhil-Ladha: 1. I agree, even I don't see that erorr being returned anywhere. What error shall we check for instead?
2. Becuase, it is used while CreateVolumeGroup RPC call to veirfy that the name was set correctly when the VolumeGroup is being created. But, as it's expected bug we are handling it in the parent function. How else do you think we should handle this case?
3. Agreed, that's why I have updated that code to return the error as `InternalError` above.

Madhu-1: lets hear from @nixpanic

internal/csi-addons/rbd/volumegroup.go
Nikhil-Ladha: Ack, I will update it to return an empty response instead.

internal/rbd/group/volume_group.go
nixpanic: `librbd.ErrNotFound` can be returned by go-ceph operations. In recent go-ceph, all `ErrNotFound` (in the rbd, cephfs and rados packages) map to `-C.ENOENT`. These errors can be mixed, and still can be compared to eachother as long as the cause (not found errno) is the same.

Ideally all errors that are returned include the underlying error with `%w` formatting. That way, a single error can be of multiple types, and comparing them is less prone to problems.

Some go-ceph functions are composed out of multiple C calls to librados and/or to librbd/libcephfs. It is not trivial to know which error is returned when, just "not found" is sufficient and should be treated the same way for librados and librbd.

A similar problem surfaces sometimes in Ceph-CSI. Some operations return the received error directly, others replace it with their own variant. This makes it a good source of bugs, as checking for the error type is non-trivial.

The safest would be to have dedicated errors per package, and only return those (maybe only for exported functions?). This can be done by starting to use the `%w` formatting, and even including matching errors with the `%w` formatting in the exported error variables.

```go
var (
	ErrRBDGroupNotConnected = fmt.Errorf("%w: RBD group is not connected", librados.ErrNotConnected)
	ErrRBDGroupNotFound     = fmt.Errorf("%w: RBD group not found", librbd.ErrNotFound)
)
```

This way, `err = ErrRBDGroupNotFound; errors.Is(err, librbd.ErrNotFound)` will still be true, and changes to the rest of the code can be done piece by piece. Eventually only `errors.Is(err, group.ErrRBDGroupNotFound)` would then be used outside of the `internal/rbd/group` package.

Nikhil-Ladha: I am wondering if having a name check that return the `librbd.ErrNotFound` error from the `getVolumeAttributes` function would be a better choice? In that way, we can just compare the `ErrRBDGroupNotFound ` error from that function and return?

internal/csi-addons/rbd/volumegroup.go
Nikhil-Ladha: Done

internal/rbd/group/volume_group.go
Nikhil-Ladha: I have updated the code to reflect my above statement, please check again once.

internal/rbd/group/util.go
```diff
@@ -145,6 +146,12 @@ func (cvg *commonVolumeGroup) getVolumeGroupAttributes(ctx context.Context) (*jo
 		attrs = &journal.VolumeGroupAttributes{}
 	}
 
+	if attrs.GroupName == "" || attrs.CreationTime == nil {
+		log.ErrorLog(ctx, "volume group with id %v not found", cvg.id)
+
+		return nil, librbd.ErrNotFound
```
nixpanic: Don't return a `librbd` error, only use `librbd` errors for checking. Return the new `ErrRBDGroupNotFound` instead. `mgr.GetVolumeGroupByID()` checks for `group.ErrRBDGroupNotFound`, which is more specific than `librbd.ErrNotFound` and therefor will not match it.

nixpanic (changes requested):
The approach looks good to me. Just a small mistake that needs correcting.

internal/rbd/group/util.go
Nikhil-Ladha: Ack, updated the code.

nixpanic (approved):

Madhu-1 (approved):
A followup PR is required to update the DeleteVolumeGroupSnapshot and GetvolumeGroupsnapshot RPC to make use of ErrRBDGroupNotFound to return success and not found and even in ControllerGetVolumeGroup RPC as well

Nikhil-Ladha:
> A followup PR is required to update the DeleteVolumeGroupSnapshot and GetvolumeGroupsnapshot RPC to make use of ErrRBDGroupNotFound to return success and not found and even in ControllerGetVolumeGroup RPC as well

ControllerGetVolumeGroup is taken care off in this PR itself, only the GroupSnapshot RPCs need to be updated.
i.e, The ControllerGetVolumeGroup RPC calls GetVolumeGroupByID [1] function,  that internally calls GetVolumeGroup which returns the `ErrRBDGroupNotFound` error, if the group doesn't exists. And, in the parent `ControllerGetVolumeGroup` RPC call, we return the `NotFound` status.code if the err != nil [3].

[1] https://github.com/ceph/ceph-csi/blob/6c0305c8b071e0e9aeb4455629b03964d5ac4d1a/internal/csi-addons/rbd/volumegroup.go#L434
[2] https://github.com/ceph/ceph-csi/blob/6c0305c8b071e0e9aeb4455629b03964d5ac4d1a/internal/rbd/manager.go#L224
[3] https://github.com/ceph/ceph-csi/blob/6c0305c8b071e0e9aeb4455629b03964d5ac4d1a/internal/csi-addons/rbd/volumegroup.go#L435

Madhu-1:
> > A followup PR is required to update the DeleteVolumeGroupSnapshot and GetvolumeGroupsnapshot RPC to make use of ErrRBDGroupNotFound to return success and not found and even in ControllerGetVolumeGroup RPC as well
> 
> ControllerGetVolumeGroup is taken care off in this PR itself, only the GroupSnapshot RPCs need to be updated.

i was not able to find the code for it, can you please point me to the code.

Nikhil-Ladha:
@Madhu-1 updated my previous comment with links, please check now.

nixpanic:
@mergifyio queue

Nikhil-Ladha:
@Mergifyio backport release-v3.13

PR #5008, opened 2024-12-10
build: include `--nobest` while upgrading the devel container packages

nixpanic:
On occasion there are inconsistencies in the CentOS or Ceph RPM
repositories. This can cause `dnf update` to fail when there are
conflicting versions of a package. By adding the `--nobest` option,
packages can get upgraded to an older version, but not the latest.

### DIFF
--- a/scripts/Dockerfile.devel
+++ b/scripts/Dockerfile.devel
@@ -35,7 +35,7 @@ RUN dnf -y install \
 	librados-devel \
         libcephfs-devel \
 	librbd-devel \
-    && dnf -y update \
+    && dnf -y --nobest update \
     && dnf clean all \
     && rm -rf /var/cache/yum \
     && true


black-dragon74:
Maybe we should take this opportunity to make it like:

```Dockerfile
RUN dnf -y --nobest update \
      && dnf -y install \
    git \
    make \
    gcc \
    librados-devel \
    libcephfs-devel \
    librbd-devel \
    && dnf clean all \
    && rm -rf /var/cache/yum \
    && true
```

Rakshith-R (approved):

nixpanic:
> Maybe we should take this opportunity to make it like:
> 
> ```dockerfile
> RUN dnf -y --nobest update \
>       && dnf -y install \
>     git \
>     make \
>     gcc \
>     librados-devel \
>     libcephfs-devel \
>     librbd-devel \
>     && dnf clean all \
>     && rm -rf /var/cache/yum \
>     && true
> ```

The current extra indention is added as the packages are part of the `dnf -y install` command. All stand-alone commands start with an indention and `&&`, so I think it is nicely aligned already.

black-dragon74 (approved):

Rakshith-R:
@mergifyio queue

PR #5016, opened 2024-12-12
rbd: support QoS based on capacity for rbd volume

YiteGu:
1. QoS provides settings for rbd volume read/write iops and read/write bandwidth.
2. All QoS parameters are placed in the SC, send QoS parameters from SC to Cephcsi through PVC create request.
3. We need provide QoS parameters in the SC as below:
   - BaseReadIops
   - BaseWriteIops
   - BaseReadBytesPerSecond
   - BaseWriteBytesPerSecond
   - ReadIopsPerGiB
   - WriteIopsPerGiB
   - ReadBpsPerGiB
   - WriteBpsPerGiB
   - BaseVolSizeBytes 
There are 4 base qos parameters among them, when users apply for a volume capacity equal to or less than BaseVolSizebytes, use base qos limit. For the portion of capacity exceeding BaseVolSizebytes, QoS will be increased in steps set per GB. If the step size parameter per GB is not provided, only base QoS limit will be used and not associated with capacity size.
4. If PVC has resize request, adjust the QoS limit according to the QoS parameters after resizing.

Do you have any questions?

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/docs/rbd/deploy.md
+++ b/docs/rbd/deploy.md
@@ -73,6 +73,15 @@ make image-cephcsi
 | `stripeUnit`                                                                                        | no                   | stripe unit in bytes                                                                                                                                                                                                                                                                               |
 | `stripeCount`                                                                                       | no                   | objects to stripe over before looping                                                                                                                                                                                                                                                              |
 | `objectSize`                                                                                        | no                   | object size in bytes                                                                                                                                                                                                                                                                               |
+| `BaseReadIops`                                                                                      | no                   | the base limit of read operations per second |
+| `BaseWriteIops`                                                                                     | no                   | the base limit of write operations per second |
+| `BaseReadBytesPerSecond`                                                                            | no                   | the base limit of read bytes per second |
+| `BaseWriteBytesPerSecond`                                                                           | no                   | the base limit of write bytes per second |
+| `ReadIopsPerGiB`                                                                                    | no                   | the limit of read operations per GiB |
+| `WriteIopsPerGiB`                                                                                   | no                   | the limit of write operations per GiB |
+| `ReadBpsPerGiB`                                                                                     | no                   | the limit of read bytes per GiB |
+| `WriteBpsPerGiB`                                                                                    | no                   | the limit of write bytes per GiB |
+| `BaseVolSizeBytes`                                                                                  | no                   | the min size of volume what use to calculate qos beased on capacity |
 | `extraDeploy` | no | array of extra objects to deploy with the release |
 
 **NOTE:** An accompanying CSI configuration file, needs to be provided to the
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -27,6 +27,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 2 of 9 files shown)


internal/rbd/controllerserver.go
```diff
@@ -231,6 +231,22 @@ func (cs *ControllerServer) parseVolCreateRequest(
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	// Get QosParameters from SC if qos configuration existing in SC
```
nixpanic, line 234: Can you move this par to a function of the rbdVolume struct (in a new file `qos.go`)?

```go
func (rv *rbdVolume) SetQOS(params map[string]string) error {
	baseQosVolSize := ""
	if v, ok := params[baseVolSizeBytes]; ok && v != "" {
		baseQosVolSize = v
	}

	rbdQosParameters := parseQosParams(params)
	for _, param := range rbdQosParameters {
		if param.provide {
			err = calcQosBasedOnCapacity(ctx, rbdVol, *param, baseQosVolSize)
			if err != nil {
				return err
			}
		}
	}

	return nil
}
```

and then use it like

```go
err = rbdVol.SetQOS(req.GetParameters())
if err != nil {
    return status.Error(codes.InvalidArgument, err.Error())
}
```

internal/rbd/controllerserver.go
```diff
@@ -243,6 +259,32 @@ func (cs *ControllerServer) parseVolCreateRequest(
 	return rbdVol, nil
 }
 
+func parseQosParams(
```
nixpanic: move this function to a new file `qos.go` please

internal/rbd/controllerserver.go
```diff
@@ -424,6 +466,19 @@ func (cs *ControllerServer) CreateVolume(
 		return nil, err
 	}
 
+	// Set rbd qos for image
```
nixpanic: move this to a function of the rbdVolume struct too (in `qos.go`)

internal/rbd/controllerserver.go
```diff
@@ -1615,6 +1675,33 @@ func (cs *ControllerServer) ControllerExpandVolume(
 	}, nil
 }
 
+func (cs *ControllerServer) adjustRbdImageQos(
```
nixpanic: `cs` is unused in this function, move it to the rbdVolume struct as well please.

internal/rbd/rbd_util.go
```diff
@@ -81,6 +81,35 @@ const (
 
 	// clusterNameKey cluster Key, set on RBD image.
 	clusterNameKey = "csi.ceph.com/cluster/name"
+
```
nixpanic: `rbd_util.go` is really large already, new features like QoS benefit from being placed as much as possible in a new `qos.go` file. Please move everything there, thanks!

internal/rbd/rbd_util.go
```diff
@@ -81,6 +81,35 @@ const (
...
+	baseReadBytesPerSecond  = "BaseReadBytesPerSecond"
+	baseWriteBytesPerSecond = "BaseWriteBytesPerSecond"
+	readIopsPerGB           = "ReadIopsPerGB"
+	writeIopsPerGB          = "WriteIopsPerGB"
+	readBpsPerGB            = "ReadBpsPerGB"
+	writeBpsPerGB           = "WriteBpsPerGB"
+	baseVolSizeBytes        = "BaseVolSizeBytes"
+
+	// Qos type name of rbd image.
+	rbdQosReadIopsLimit  = "rbd_qos_read_iops_limit"
```
nixpanic: drop the `rbd` prefix of the variables if you can

internal/rbd/rbd_util.go
```diff
@@ -81,6 +81,35 @@ const (
 
 	// clusterNameKey cluster Key, set on RBD image.
 	clusterNameKey = "csi.ceph.com/cluster/name"
+
+	// Qos parameters name of StorageClass.
+	baseReadIOPS            = "BaseReadIOPS"
+	baseWriteIOPS           = "BaseWriteIOPS"
+	baseReadBytesPerSecond  = "BaseReadBytesPerSecond"
+	baseWriteBytesPerSecond = "BaseWriteBytesPerSecond"
+	readIopsPerGB           = "ReadIopsPerGB"
```
nixpanic: some variables and values have **IOPS** (my preference), some have **Iops**. Please use one format everywhere.

internal/rbd/controllerserver.go
```diff
@@ -424,6 +466,19 @@ func (cs *ControllerServer) CreateVolume(
 		return nil, err
 	}
 
+	// Set rbd qos for image
+	if len(rbdVol.QosParameters) != 0 && rbdVol.Mounter == rbdNbdMounter {
```
nixpanic: It probably makes sense to store the QosParameters even when the mounter isn't NBD. With the upcoming ControllerModifyVolume CSI call, users may possibly switch between different mounters in the future.

There is also interest in configuring IOPS for block devices (through `/sys`) when the kernel mounter is used. Basing the IOPS configuration on what NBD uses would be most practical.

internal/rbd/rbd_util.go
```diff
@@ -212,6 +244,14 @@ type migrationVolID struct {
 	clusterID string
 }
 
+type rbdQos struct {
```
nixpanic: remove the rbd prefixes here too please, this is all in the rbd package no need for the duplication

nixpanic (commented):
Thanks for the PR! This is really a nice feature to have. A couple of comments that need to be addressed to improve the maintainability and usability.

- move all qos related functions, structs, constants to a new `qos.go` file
- all new functions that receive a *rbdVolume as argument should become a function of the rbdVolume struct (see example for the `SetQOS` function)
- new parameters for the StorageClass need to be added to the documentation
- documentation should explain that the QoS settings are currently only used in combination with the NBD mounter
- ideally add unit tests for functions that do not interact with a Ceph cluster
- ideally add some form or e2e testing, maybe set QoS parameters in the StorageClass, and verify that those are set on an RBD-image?

Thanks again, this is a great start!

internal/rbd/controllerserver.go
YiteGu: done

internal/rbd/controllerserver.go
YiteGu: done

internal/rbd/controllerserver.go
YiteGu: done

internal/rbd/controllerserver.go
YiteGu: done

internal/rbd/rbd_util.go
YiteGu: done

internal/rbd/rbd_util.go
YiteGu: done

internal/rbd/rbd_util.go
YiteGu: Unified use `Iops` format

internal/rbd/rbd_util.go
YiteGu: done

internal/rbd/controllerserver.go
YiteGu: Get, as long as SC provides QoS parameters, we should set them for it. But, we need to clarify that currently only NBD is supported.

internal/rbd/qos_test.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	for k, v := range rv.QosParameters {
+		switch k {
+		case readIopsLimit:
+			if v != tests[baseReadIops] {
+				t.Errorf("SetQOS: %s: %s, want %s", k, v, tests[baseReadIops])
+			}
+		case writeIopsLimit:
+			if v != tests[baseWriteIops] {
+				t.Errorf("SetQOS: %s: %s, want %s", k, v, tests[baseReadIops])
```
nixpanic: this should probably be `tests[baseWriteIops]`

internal/rbd/controllerserver.go
```diff
@@ -424,6 +430,19 @@ func (cs *ControllerServer) CreateVolume(
 		return nil, err
 	}
 
+	// Apply Qos parameters to rbd image.
```
nixpanic: The CreateVolume function is already at the limit of the complexity (number of if/else and loop statements). Maybe it makes sense to move these two parts to the `createBackingImage()` function?

https://github.com/ceph/ceph-csi/actions/runs/12390057314/job/34584308943?pr=5016#step:3:841

internal/rbd/controllerserver.go
```diff
@@ -424,6 +430,19 @@ func (cs *ControllerServer) CreateVolume(
...
+	err = rbdVol.ApplyQOS(ctx)
+	if err != nil {
+		log.DebugLog(ctx, "failed apply QOS for rbd image: %v", err)
+
+		return nil, status.Error(codes.Internal, err.Error())
```
nixpanic: note that a CSI function like `CreateVolume` needs to return a `status.Error()` or `status.Errorf()` generated type of error. Functions like `createBackingImage()` should return a 'normal' error. This is not in-sync with the error returned from `SaveQOS()` in the call below.

nixpanic:
Looks quite complete, many thanks for adding the unit and e2e tests!

nixpanic:
> /test ci/centos/mini-e2e/k8s-1.31

not sure why that did not work, but the CI is running now (takes ~2 hours to complete): https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/job/mini-e2e_k8s-1.31/283/display/redirect

internal/rbd/qos_test.go
YiteGu: Yup :-)

YiteGu:
> Looks quite complete, many thanks for adding the unit and e2e tests!

Write it briefly to verify that e2e testing is available. More scenes need to be added. :-)

internal/rbd/controllerserver.go
YiteGu: Yes, I full agree.

internal/rbd/controllerserver.go
YiteGu: Get

nixpanic (dismissed):

docs/rbd/deploy.md
```diff
@@ -73,6 +73,15 @@ make image-cephcsi
...
+| `BaseReadIops`                                                                                      | no                   | the base limit of read operations per second |
+| `BaseWriteIops`                                                                                     | no                   | the base limit of write operations per second |
+| `BaseReadBytesPerSecond`                                                                            | no                   | the base limit of read bytes per second |
+| `BaseWriteBytesPerSecond`                                                                           | no                   | the base limit of write bytes per second |
+| `ReadIopsPerGB`                                                                                     | no                   | the limit of read operations per GiB |
```
Madhu-1: in the key `GB` is used and in the comments its mentioned `GiB` is that correct?

docs/rbd/deploy.md
```diff
@@ -73,6 +73,15 @@ make image-cephcsi
...
+| `WriteIopsPerGB`                                                                                    | no                   | the limit of write operations per GiB |
+| `ReadBpsPerGB`                                                                                      | no                   | the limit of read bytes per GiB |
+| `WriteBpsPerGB`                                                                                     | no                   | the limit of write bytes per GiB |
+| `BaseVolSizeBytes`                                                                                  | no                   | the min size of volume what use to calc qos beased on capacity |
```
Madhu-1: `calc` to `calculate`

docs/rbd/deploy.md
```diff
@@ -73,6 +73,15 @@ make image-cephcsi
...
 | `stripeCount`                                                                                       | no                   | objects to stripe over before looping                                                                                                                                                                                                                                                              |
 | `objectSize`                                                                                        | no                   | object size in bytes                                                                                                                                                                                                                                                                               |
+| `BaseReadIops`                                                                                      | no                   | the base limit of read operations per second |
```
Madhu-1, line 76: can all these operations can be supported for `krbd`?

e2e/rbd.go
```diff
@@ -4643,6 +4644,130 @@ var _ = Describe("RBD", func() {
 				validateOmapCount(f, 0, rbdType, defaultRBDPool, volumesType)
 			})
 
+			By("validate rbd image qos", func() {
+				qosParameters := map[string]string{
+					"BaseReadIops":            "2000",
+					"BaseWriteIops":           "1000",
+					"BaseReadBytesPerSecond":  "209715200",
+					"BaseWriteBytesPerSecond": "104857600",
```
Madhu-1: please define a var for numeric values as its used in two places to have single source

e2e/rbd.go
```diff
@@ -4643,6 +4644,130 @@ var _ = Describe("RBD", func() {
...
+
+				// create PVC
+				pvc, err = loadPVC(pvcPath)
+				if err != nil {
+					framework.Failf("failed to load PVC: %v", err)
+				}
+
+				pvc.Namespace = f.UniqueName
+				pvc.Spec.Resources.Requests[v1.ResourceStorage] = resource.MustParse("100Gi")
+				wants = map[string]string{
```
Madhu-1, line 4769: what will happen for cloned PVC or PVC created from snapshot, are they going to get the same values or different values?

Can you please add a test case of that as well

e2e/rbd.go
```diff
@@ -4643,6 +4644,130 @@ var _ = Describe("RBD", func() {
...
+				err = validateQOS(f, pvc, wants)
+				if err != nil {
+					framework.Failf("failed to validate qos: %v", err)
+				}
+
+				// delete pvc
+				err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+				if err != nil {
+					framework.Failf("failed to delete PVC: %v", err)
+				}
```
Madhu-1, line 4917: validate the omap/imageCount etc to avoid leaks

internal/rbd/controllerserver.go
```diff
@@ -766,6 +773,21 @@ func (cs *ControllerServer) createBackingImage(
 		return status.Error(codes.Internal, err.Error())
 	}
 
+	// Apply Qos parameters to rbd image.
+	err = rbdVol.ApplyQOS(ctx)
+	if err != nil {
+		log.DebugLog(ctx, "failed apply QOS for rbd image: %v", err)
```
Madhu-1: log it as error

internal/rbd/controllerserver.go
```diff
@@ -766,6 +773,21 @@ func (cs *ControllerServer) createBackingImage(
...
+
+		return status.Error(codes.Internal, err.Error())
+	}
+	// Save Qos parameters from SC in Image medatate, we will use it while resize volume.
+	err = rbdVol.SaveQOS(ctx, scParams)
+	if err != nil {
+		log.DebugLog(ctx, "failed save QOS for rbd image: %v", err)
```
Madhu-1: log it as error

internal/rbd/controllerserver.go
```diff
@@ -1605,6 +1627,13 @@ func (cs *ControllerServer) ControllerExpandVolume(
 		if err != nil {
 			log.ErrorLog(ctx, "failed to resize rbd image: %s with error: %v", rbdVol, err)
 
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		// adjust rbd qos after resize volume.
+		err = rbdVol.AdjustQOS(ctx)
+		if err != nil {
+			log.DebugLog(ctx, "failed adjust QOS for rbd image")
```
Madhu-1: log as error

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+		baseReadIops:            {readIopsLimit, "", readIopsPerGB, "", false},
+		baseWriteIops:           {writeIopsLimit, "", writeIopsPerGB, "", false},
+		baseReadBytesPerSecond:  {readBpsLimit, "", readBpsPerGB, "", false},
+		baseWriteBytesPerSecond: {writeBpsLimit, "", writeBpsPerGB, "", false},
+	}
+	for k, v := range scParams {
+		if qos, ok := rbdQosParameters[k]; ok && v != "" {
+			qos.baseLimit = v
+			qos.provide = true
+			for _k, _v := range scParams {
```
Madhu-1: please dont use `_k` and `_v`

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+	rbdQosParameters := map[string]*qosSpec{
+		baseReadIops:            {readIopsLimit, "", readIopsPerGB, "", false},
+		baseWriteIops:           {writeIopsLimit, "", writeIopsPerGB, "", false},
+		baseReadBytesPerSecond:  {readBpsLimit, "", readBpsPerGB, "", false},
+		baseWriteBytesPerSecond: {writeBpsLimit, "", writeBpsPerGB, "", false},
+	}
+	for k, v := range scParams {
+		if qos, ok := rbdQosParameters[k]; ok && v != "" {
+			qos.baseLimit = v
+			qos.provide = true
```
Madhu-1: can we rename `provide` to something more accurate? like present etc?

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+			log.ErrorLog(ctx, "failed to set rbd qos, %s: %s. %v", k, v, err)
+
+			return err
+		}
+	}
+
+	return nil
+}
+
+func calcQosBasedOnCapacity(
```
Madhu-1: Can this also be a method of `rbdVol` or `rbdImage`?

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+					return err
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func getRbdImageQOS(
```
Madhu-1: make this as method of `rbdVol`?

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+			continue
+		} else if err != nil {
+			log.ErrorLog(ctx, "failed to get metadata: %s. %v", k, err)
+
+			return nil, err
+		}
+		perGBLimit, err := rbdVol.GetMetadata(param.rbdQosPerGBType)
+		if errors.Is(err, librbd.ErrNotFound) {
+			// rbdQosPerGBType does not exist, set it empty.
+			perGBLimit = ""
```
Madhu-1: Do we need explict setting it to empty string here , AFAIK it returns empty string in case of an error

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+		} else if err != nil {
+			log.ErrorLog(ctx, "failed to get metadata: %s. %v", param.rbdQosPerGBType, err)
+
+			return nil, err
```
Madhu-1: i think we dont need if and else if check here. we can have single check `err !=nil && !errors.Is(err, librbd.ErrNotFound)`

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	baseVolSize, err := rbdVol.GetMetadata(baseQosVolSize)
+	if errors.Is(err, librbd.ErrNotFound) {
+		// rbdBaseQosVolSize does not exist, set it empty.
+		baseVolSize = ""
+	} else if err != nil {
+		log.ErrorLog(ctx, "failed to get metadata: %s. %v", baseQosVolSize, err)
+
+		return nil, err
+	}
```
Madhu-1: same here as well we dont need two checks

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+
+	return rbdQosParameters, nil
+}
+
+func (rv *rbdVolume) AdjustQOS(
+	ctx context.Context,
+) error {
+	rbdQosParameters, err := getRbdImageQOS(ctx, rv)
+	if err != nil {
+		log.ErrorLog(ctx, "get rbd image qos failed")
```
Madhu-1: log the error and also why we are only logging here not in other places in this function?

internal/rbd/qos.go
```diff
@@ -0,0 +1,281 @@
...
+					qos.perGBLimit = _v
+				}
+			}
+		}
+	}
+
+	return rbdQosParameters
+}
+
+func (rv *rbdVolume) SetQOS(
```
Madhu-1, line 90: i think we can use `rbImage` instead of `rbdVolume` for most of the methods getting added in this PR.

Madhu-1 (changes requested):

examples/rbd/storageclass.yaml
```diff
@@ -162,6 +162,38 @@ parameters:
    # stripeCount: <>
    # (optional) The object size in bytes.
    # objectSize: <>
+
+   # rbd volume QoS.
```
Madhu-1, line 166: why not use https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/#the-volumeattributesclass-api as these values are not immutable as anything mentioned in the SC is mutable?

docs/rbd/deploy.md
YiteGu: Yes, these operations can be supported for `krbd`, but we need to set QoS by cgroup after get QoS limit result by `calcQosBasedOnCapacity`. if we support burst QoS, `krbd` can not be support.

docs/rbd/deploy.md
YiteGu: sorry, there is confuse, we should use unit of bytes(IEC).

e2e/rbd.go
YiteGu: cloned PVC or PVC created from snapshot get the same values after my testing on the ceph.

docs/rbd/deploy.md
YiteGu: done

e2e/rbd.go
YiteGu: updated

e2e/rbd.go
YiteGu: done

internal/rbd/controllerserver.go
YiteGu: done

internal/rbd/controllerserver.go
YiteGu: done

internal/rbd/controllerserver.go
YiteGu: done

internal/rbd/qos.go
YiteGu: I eliminated nesting `for`

internal/rbd/qos.go
YiteGu: updated

internal/rbd/qos.go
YiteGu: updated

internal/rbd/qos.go
YiteGu: done

internal/rbd/qos.go
YiteGu: updated

internal/rbd/qos.go
YiteGu: updated

internal/rbd/qos.go
YiteGu: updated

internal/rbd/qos.go
YiteGu: this log deleted

internal/rbd/qos.go
YiteGu: I don't have any particular opinion on this point, I respect your suggestion. :)

examples/rbd/storageclass.yaml
YiteGu: About this:
1. At the beginning of our discussion on this functional requirement, SC was our object of discussion. According to the discussion content, I placed these QoS parameters in SC.
2. In my opinion, VolumeAttributesClass is also nice solution. Existing SC parameters mutable is indeed an advantage. In addition, VolumeAttributesClass need higher k8s version support, SC does not have this requirement.

YiteGu:
change: @Madhu-1 @nixpanic 
1. Consider the two cases create pvc from snapshot or create pvc from pvc, the clone pvc size maybe greater than parent. So, we should adjust qos after expand pvc. On the create pvc from snapshot case, apply qos way as same as normal create. On the create pvc from pvc case, adjust qos after expand. 
2. we use `RequestedVolSize` to calculate qos values.
3. add e2e test what the pvc create size greater than  parent.

internal/rbd/qos.go
nixpanic: I do not think there is a benefit for that. when using `rbdImage` instead of `rbdVolume`, the `rbdSnapshot` struct inherits these functions (snapshots contain the metadata already, afaik). But there is currently no way for an `rbdSnapshot` to be mapped/attached, so these settings won't be used during runtime.

A possible enhancement with the `ModifyVolume` CSI procedure won't be applicable to snapshots either.

nixpanic (dismissed):
Thanks! Looks good to me.

@Madhu-1 , can you take a look again too?

internal/rbd/controllerserver.go
```diff
@@ -766,6 +773,21 @@ func (cs *ControllerServer) createBackingImage(
...
 	}
 
+	// Apply Qos parameters to rbd image.
+	err = rbdVol.ApplyQOS(ctx)
+	if err != nil {
+		log.ErrorLog(ctx, "failed apply QOS for rbd image: %v", err)
+
+		return status.Error(codes.Internal, err.Error())
+	}
+	// Save Qos parameters from SC in Image medatate, we will use it while resize volume.
```
Madhu-1: `medatate` to `metadata`

internal/rbd/controllerserver.go
```diff
@@ -1605,6 +1627,13 @@ func (cs *ControllerServer) ControllerExpandVolume(
 		if err != nil {
 			log.ErrorLog(ctx, "failed to resize rbd image: %s with error: %v", rbdVol, err)
 
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		// adjust rbd qos after resize volume.
+		err = rbdVol.AdjustQOS(ctx)
+		if err != nil {
+			log.ErrorLog(ctx, "failed adjust QOS for rbd image")
```
Madhu-1: log the error message

internal/rbd/controllerserver.go
YiteGu: done

internal/rbd/controllerserver.go
YiteGu: done

Madhu-1 (approved):

internal/rbd/qos.go
YiteGu: Get :)

nixpanic (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

PR #5024, opened 2024-12-16
util: return correct status code for VolumeGroupSnapshot

Nikhil-Ladha:
Fix status codes that are returned for Get/Delete RPC calls for VolumeGroup/VolumeGroupSnapshot.

### DIFF
--- a/internal/cephfs/store/volumegroup.go
+++ b/internal/cephfs/store/volumegroup.go
@@ -172,6 +172,12 @@ func NewVolumeGroupOptionsFromID(
 		return nil, nil, err
 	}
 
+	if groupAttributes.GroupName == "" {
+		log.ErrorLog(ctx, "volume group snapshot with id %v not found", volumeGroupSnapshotID)
+
+		return nil, nil, cerrors.ErrGroupNotFound
+	}
+
 	vgs.RequestName = groupAttributes.RequestName
 	vgs.FsVolumeGroupSnapshotName = groupAttributes.GroupName
 	vgs.VolumeGroupSnapshotID = volumeGroupSnapshotID
--- a/internal/csi-addons/rbd/volumegroup.go
+++ b/internal/csi-addons/rbd/volumegroup.go
@@ -195,7 +195,7 @@ func (vs *VolumeGroupServer) DeleteVolumeGroup(
 	vg, err := mgr.GetVolumeGroupByID(ctx, req.GetVolumeGroupId())
 	if err != nil {
 		if errors.Is(err, group.ErrRBDGroupNotFound) {
-			log.DebugLog(ctx, "VolumeGroup %q doesn't exists", req.GetVolumeGroupId())
+			log.ErrorLog(ctx, "VolumeGroup %q doesn't exists", req.GetVolumeGroupId())
 
 			return &volumegroup.DeleteVolumeGroupResponse{}, nil
 		}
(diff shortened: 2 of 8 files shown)


Nikhil-Ladha:
@Madhu-1 updated the ControllerGetVolumeGroup as well, to differentiate the case of internal errors vs actual not found errors, please take a look now.

internal/cephfs/store/volumegroup.go
```diff
@@ -172,6 +172,12 @@ func NewVolumeGroupOptionsFromID(
 		return nil, nil, err
 	}
 
+	if groupAttributes.GroupName == "" || groupAttributes.CreationTime == nil {
```
Madhu-1: GroupName is enough we should not be checking both with `or` flag.

internal/csi-addons/rbd/volumegroup.go
```diff
@@ -433,9 +433,19 @@ func (vs *VolumeGroupServer) ControllerGetVolumeGroup(
 	// resolve the volume group
 	vg, err := mgr.GetVolumeGroupByID(ctx, req.GetVolumeGroupId())
 	if err != nil {
+		if errors.Is(err, group.ErrRBDGroupNotFound) {
+			log.DebugLog(ctx, "VolumeGroup %q doesn't exists", req.GetVolumeGroupId())
```
Madhu-1: log it as `error`

internal/rbd/group/group_snapshot.go
```diff
@@ -69,6 +71,12 @@ func GetVolumeGroupSnapshot(
 
 	attrs, err := vgs.getVolumeGroupAttributes(ctx)
 	if err != nil {
+		if errors.Is(err, librbd.ErrNotFound) {
```
Madhu-1: why we can get this error condition met? i don't think `getVolumeGroupAttributes` directly return librbd error can you please check on this one again?

internal/rbd/group_controllerserver.go
```diff
@@ -229,10 +238,20 @@ func (cs *ControllerServer) GetVolumeGroupSnapshot(
 
 	groupSnapshot, err := mgr.GetVolumeGroupSnapshotByID(ctx, groupSnapshotID)
 	if err != nil {
+		if errors.Is(err, group.ErrRBDGroupNotFound) {
+			log.DebugLog(ctx, "VolumeGroupSnapshot %q doesn't exists", groupSnapshotID)
```
Madhu-1: log as error

Madhu-1 (changes requested):

internal/rbd/group/group_snapshot.go
Nikhil-Ladha: We are adding the `librbd.ErrNotFound` error to `ErrRBDGroupNotFound` error here: https://github.com/ceph/ceph-csi/blob/c7d54ab7767e04d597e891d6307249dd5a8cec79/internal/rbd/group/volume_group.go#L37, therefore, this check would work (I have tested as well)

internal/csi-addons/rbd/volumegroup.go
Nikhil-Ladha: We are logging it as error in previous functions, hence just added it as a `DebugLog` instead so that only if required these should be logged. Do you think we should log it anyway?

Madhu-1: it should be fine we should log it as error here as well.

internal/rbd/group/group_snapshot.go
Madhu-1: lets compare this with `ErrRBDGroupNotFound` instead of `librbd.ErrNotFound` as `getVolumeGroupAttributes` is returning `ErrRBDGroupNotFound` ?

Nikhil-Ladha: Right, missed that change.
Updated it now.

Madhu-1 (approved):

iPraveenParihar (approved):

iPraveenParihar:
@Nikhil-Ladha, do we need to back port this to `release-v3.13`?

Nikhil-Ladha:
Yeah, we should backport this along with the previous PR.

Nikhil-Ladha:
@Mergifyio backport release-v3.13

Nikhil-Ladha:
I don't seem to have permissions to do so :/
Can you please trigger the command for this PR as well #5001 ?

iPraveenParihar:
@Mergifyio queue

iPraveenParihar:
> I don't seem to have permissions to do so :/
Can you please trigger the command for this PR as well https://github.com/ceph/ceph-csi/pull/5001 ?

Sure, lets merge this first then will backport both in order.

PR #5071, opened 2025-01-10
rebase: update csi-resizer to v1.13.1

nixpanic:
The Kubernetes CSI community has released an updated external-resizer
sidecar that fixes some issues and adds a few improvements.

See: https://github.com/kubernetes-csi/external-resizer/releases/tag/v1.13.1


Madhu-1 (dismissed):

nixpanic:
It seems the new container-image is not available in the registry yet.

nixpanic:
> It seems the new container-image is not available in the registry yet.

v1.13.0 failed to build and could not be pushed to the release repository, v1.13.1 contains the release-tools fixes and is now available.

Madhu-1 (dismissed):

Rakshith-R (dismissed):

nixpanic:
@mergifyio queue

nixpanic:
Interesting failure in the CI:
```
  [FAIL] External Storage [Driver: rbd.csi.ceph.com] [Testpattern: Dynamic PV (default fs)(allowExpansion)] volume-expand [It] Verify if offline PVC expansion works
```
Offline PVC expansion is not expected to work if the PVC has a filesystem on it :thinking:

Madhu-1:
@nixpanic https://kubernetes.slack.com/archives/C09QZFCE5/p1736798710626099 it looks like its broken. why its running offline expansion for ceph-csi, should we disable it as we don't support it?

nixpanic:
> @nixpanic https://kubernetes.slack.com/archives/C09QZFCE5/p1736798710626099 it looks like its broken. why its running offline expansion for ceph-csi, should we disable it as we don't support it?

Yeah, I also wonder why it would run. Will look into it soonish. Maybe it is because `controllerExpansion: true` is set in `scripts/k8s-storage/driver-rbd.yaml`.

scripts/k8s-storage/driver-rbd.yaml
```diff
@@ -50,8 +50,8 @@ DriverInfo:
     # Support for volume limits
     volumeLimits: false
 
-    # Support for volume expansion in controllers
-    controllerExpansion: true
+    # Support for volume expansion in controllers (only for RBD with BlockMode)
+    controllerExpansion: false
```
Madhu-1: @nixpanic we still need this to set to true for filesystem as well as it needs both controllerExpansion and filesystem resize as well?

Madhu-1: https://github.com/kubernetes/kubernetes/blob/ec2e0de35a298363872897e5904501b029817af3/test/e2e/storage/framework/testdriver.go#L193 this need to be set to false as its enabled by default?

nixpanic: yeah, the options are a bit confusing, and their documentation could be better... I now found an [offlineExpansion](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/framework/testdriver.go#L193) option, maybe that is more suitable.

scripts/k8s-storage/driver-rbd.yaml
```diff
@@ -56,6 +56,9 @@ DriverInfo:
     # Support for volume expansion in nodes
     nodeExpansion: true
 
+    # supports offline volume expansion
+    offlineExpansion: false
```
Madhu-1, line 60: @nixpanic same need to be done for cephfs and nfs driver as well?

Madhu-1 (approved):
LGTM

scripts/k8s-storage/driver-rbd.yaml
nixpanic: no, CephFS and NFS can be expanded while the volume is not mounted.

iPraveenParihar (approved):

Rakshith-R:
@mergifyio queue

PR #5090, opened 2025-01-17
cephfs: upgrading mount syntax

MageekChiu:
[The old syntax](https://docs.ceph.com/en/pacific/man/8/mount.ceph/) is almost deprecated, and there are reasons to upgrade it to [the new one](https://docs.ceph.com/en/reef/man/8/mount.ceph/)
 - old syntax is lack of fsid param, which is critical for debugging and observability
 - mds_namespace is deprecated, so it might be inappropriate to continue using it
 - kernel will try new syntax first and then the old one(check with mount -v), it is a waste to use the old one
 
From the ubuntu manpage，[20.04 LTS](https://manpages.ubuntu.com/manpages/focal/man8/mount.ceph.8.html) supports the old syntax while [22.04 LTS](https://manpages.ubuntu.com/manpages/jammy/man8/mount.ceph.8.html) supports the new one.

### DIFF
--- a/internal/cephfs/mounter/kernel.go
+++ b/internal/cephfs/mounter/kernel.go
@@ -72,25 +72,24 @@ func (m *kernelMounter) mountKernel(
 		m.needsModprobe = false
 	}
 
+	fsID, err := volOptions.GetFSID()
+	if err != nil {
+		return fmt.Errorf("failed to get fsID, stop mounting: %w", err)
+	}
+
 	args := []string{
 		"-t", "ceph",
-		fmt.Sprintf("%s:%s", volOptions.Monitors, volOptions.RootPath),
+		fmt.Sprintf("%s@%s.%s=%s", cr.ID, fsID, volOptions.FsName, volOptions.RootPath),
 		mountPoint,
 	}
 
-	optionsStr := fmt.Sprintf("name=%s,secretfile=%s", cr.ID, cr.KeyFile)
-	mdsNamespace := ""
-	if volOptions.FsName != "" {
-		mdsNamespace = "mds_namespace=" + volOptions.FsName
-	}
-	optionsStr = util.MountOptionsAdd(optionsStr, mdsNamespace, volOptions.KernelMountOptions, netDev)
+	optionsStr := fmt.Sprintf("mon_addr=%s,secretfile=%s", strings.ReplaceAll(volOptions.Monitors, ",", "/"), cr.KeyFile)
+
+	optionsStr = util.MountOptionsAdd(optionsStr, volOptions.KernelMountOptions, netDev)
 
 	args = append(args, "-o", optionsStr)
 
-	var (
-		stderr string
-		err    error
-	)
+	var stderr string
 
 	if volOptions.NetNamespaceFilePath != "" {
 		_, stderr, err = util.ExecuteCommandWithNSEnter(ctx, volOptions.NetNamespaceFilePath, "mount", args[:]...)
--- a/internal/cephfs/store/volumeoptions.go
+++ b/internal/cephfs/store/volumeoptions.go
@@ -103,6 +103,19 @@ func (vo *VolumeOptions) Destroy() {
 	}
 }
 
+func (vo *VolumeOptions) GetFSID() (string, error) {
+	if vo.conn == nil {
+		return "", errors.New("cluster not connected yet")
+	}
+
+	fsID, err := vo.conn.GetFSID()
+	if err != nil {
+		return "", err
+	}
+
+	return fsID, nil
+}
+
 func validateNonEmptyField(field, fieldName string) error {
 	if field == "" {
 		return fmt.Errorf("parameter '%s' cannot be empty", fieldName)
@@ -682,6 +695,17 @@ func NewVolumeOptionsFromMonitorList(
 		opts.BackingSnapshot = true
 	}
 
+	cr, err := util.NewAdminCredentials(secrets)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer cr.DeleteCredentials()
+
+	err = opts.Connect(cr)
+	if err != nil {
+		return nil, nil, err
+	}
+
 	return &opts, &vid, nil
 }
 
@@ -730,7 +754,7 @@ func NewVolumeOptionsFromStaticVolume(
 		return nil, nil, err
 	}
 
-	if err = extractOptionalOption(&opts.FsName, "fsName", options); err != nil {
+	if err = extractOption(&opts.FsName, "fsName", options); err != nil {
 		return nil, nil, err
 	}
 
@@ -759,6 +783,17 @@ func NewVolumeOptionsFromStaticVolume(
 
 	if opts.BackingSnapshotID != "" {
 		opts.BackingSnapshot = true
+	}
+
+	cr, err := util.NewAdminCredentials(secrets)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer cr.DeleteCredentials()
+
+	err = opts.Connect(cr)
+	if err != nil {
+		return nil, nil, err
 	}
 
 	return &opts, &vid, nil
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -482,14 +482,6 @@ var _ = Describe(cephfsType, func() {
 				}
 			})
 
-			By("check static PVC with without FsName", func() {
-				scPath := cephFSExamplePath + "secret.yaml"
-				err := validateCephFsStaticPV(f, appPath, scPath, "")
-				if err != nil {
-					framework.Failf("failed to validate CephFS static pv without filesystem name: %v", err)
-				}
-			})
-
 			By("create a storageclass with pool and a PVC then bind it to an app", func() {
 				err := createCephfsStorageClass(f.ClientSet, f, true, nil)
 				if err != nil {


nixpanic:
Many thanks @MageekChiu!

Could you update the PR description with a reference that explains the changes? It would be useful to know how recent these changes are, and if there could be a problem when users have an older version deployed.

It seems the commit message contains a few long lines. Please edit it and keep them under 80 characters.

@kotreshhr, maybe you have a reference to the new/changed mount options handy?

MageekChiu:
@nixpanic Greate advices, thank you. 
I‘ve edited the PR description and the commit message.

Cool homepage by the way, need to add one myself 😄.

nixpanic (dismissed):
Looks good to me, thanks!

internal/cephfs/mounter/kernel.go
```diff
@@ -74,16 +74,13 @@ func (m *kernelMounter) mountKernel(
 
 	args := []string{
 		"-t", "ceph",
-		fmt.Sprintf("%s:%s", volOptions.Monitors, volOptions.RootPath),
+		fmt.Sprintf("%s@%s.%s=%s", cr.ID, volOptions.ClusterID, volOptions.FsName, volOptions.RootPath),
```
iPraveenParihar: @MageekChiu, The `volOptions.ClusterID` does not represent the actual Ceph cluster ID. Instead, Ceph-CSI uses it as a unique identifier to map the cluster configuration between the clusterID parameter specified in the StorageClass and the clusterID defined in the `ceph-csi-config` ConfigMap.  This field does not necessarily have to be the cluster FSID required by the mount option.

nixpanic: Good catch! `internal/util.GetMonsAndClusterID()` should be used for obtaining the real cluster-id.

iPraveenParihar: @nixpanic 🤔, even that just returns the clusterID from the csi configFile.  The right method is `internal/util.GetFSID()`

https://github.com/ceph/ceph-csi/blob/4e02fae9212fa70d782ff8c6420a0c56be5c3487/internal/util/connection.go#L118-L124

MageekChiu: @iPraveenParihar  @nixpanic thanks for your opinions, very helpful.

Ceph support multiple fs nowadays, but I don't see any parameter indicates our csi supports that.
So we only support getting fsid like bellow?
```
conn := &util.ClusterConnection{}
err = conn.Connect(monitors, cr)
if err != nil {
    return nil, status.Errorf(codes.Internal, "failed to connect to MONs %q: %s", monitors, err)
}
defer conn.Destroy()

fsID, err := conn.GetFSID()
if err != nil {
    return nil, status.Errorf(codes.Internal, "failed to get cephfs id: %s", err)
}
```

By the way, the fsid of the default and only one fs is indeed the clusterID itself.

If we don't intend to support multiple fs, then `internal/util.GetMonsAndClusterID()` should be enough.

nixpanic: > @nixpanic 🤔, even that just returns the clusterID from the csi configFile.

Isn't that the `ceph-csi-config` ConfigMap?

MageekChiu: Oh, storageclass parameter `fsName` indicate csi support multiple fs.
How about we just add one more parameter `fsid` to the storageclass to eliminate the need to call the api of ceph cluster?
fsid can be omitted, which is also supported by the new syntax
`mount -t ceph name@.fs_name=/ /mnt/mycephfs -o mon_addr=1.2.3.4`

MageekChiu: > Ceph support multiple fs nowadays, but I don't see any parameter indicates our csi supports that.

My mistake, the fsName in storageclass does that, I'll do the refactoring.

And for clarity, FsID is [UUID for a ceph cluster](https://docs.ceph.com/en/reef/man/8/ceph/#fsid) rather than an id for a CephFS instance(file system)

kotreshhr:
> Many thanks @MageekChiu!
> 
> Could you update the PR description with a reference that explains the changes? It would be useful to know how recent these changes are, and if there could be a problem when users have an older version deployed.
> 
> It seems the commit message contains a few long lines. Please edit it and keep them under 80 characters.
> 
> @kotreshhr, maybe you have a reference to the new/changed mount options handy?

I see the old and new mount options are already linked in the PR. Yes, the kernel tries the new syntax first and falls back to old if not available. I think this should be fine but tagging the kernel developer @Markuze @vshankar to confirm if this breaks anything based on the version it got introduced.

internal/cephfs/store/volumeoptions.go
```diff
@@ -46,6 +46,7 @@ type VolumeOptions struct {
 	RequestName  string
 	NamePrefix   string
 	ClusterID    string
+	FsID         string
```
nixpanic: Instead of always fetching the `FsID`, can you add a `func (vo *VolumeOptions) GetFSID() (string, error)` function?

This can then be used where needed, currently only in `NewKernelMounter()`.

By extending the `NewVolumeOptions()` function, it grows (too) large, and the golang-ci linter fails due to that.

iPraveenParihar: @nixpanic,I believe you may have missed this—we already have `GetFSID()` defined in the `ClusterConnection` struct. Since `ClusterConnection` is a member of the `VolumeOptions` struct, it can be accessed via `vo.conn.GetFSID()`.

nixpanic:
Hi @MageekChiu, the PR looks good to me. Can you squash all commits into a single one and force-push your branch? That makes it possible for the @Mergify bot can do it's work and get this in soon.

Thanks!

internal/cephfs/mounter/kernel.go
```diff
@@ -72,24 +72,25 @@ func (m *kernelMounter) mountKernel(
 		m.needsModprobe = false
 	}
 
+	fsID, err := volOptions.GetFSID()
```
iPraveenParihar, line 75: why can't we use existing `GetFSID()` instead?
```suggestion
	fsID, err := volOptions.GetConnection().GetFSID()
```

iPraveenParihar (changes requested):

internal/cephfs/mounter/kernel.go
nixpanic: `volOptions.GetConnection()` can return `nil`, so it is not really safe. A new `GetFSID()` would prevent any incorrect usage, now, and possibly in the future.

MageekChiu: Yeah, I think the checking is needed, `Connect` and `Destroy` do the checking too.
And I've squashed the commits.

nixpanic (dismissed):

internal/cephfs/mounter/kernel.go
iPraveenParihar: > volOptions.GetConnection() can return nil, so it is not really safe. A new GetFSID() would prevent any incorrect usage, now, and possibly in the future.

when volOptions struct is created, we make sure we have the connection set. There is no way `GetConnection()` could return nil.

IMO, `GetConnection()` should be handling the nil check, or the `GetConnection()` caller should check for nil and proceed.

 as we have similar usage at some places. _for example_ -  https://github.com/ceph/ceph-csi/blob/72b9d5ac54d71e0f034a18a38d410a2da329c5e0/internal/cephfs/nodeserver.go#L145-L153

nixpanic: Sure, improving `GetConnection()` would be nice too. I don't think that needs to be done in this PR though.

nixpanic: > Sure, improving `GetConnection()` would be nice too. I don't think that needs to be done in this PR though.

That can be done through issue #5129.

nixpanic:
> /test ci/centos/mini-e2e/k8s-1.32

There has been a new Ceph release yesterday, building Ceph-CSI fails until the new container-image can install required RPMs (librados-devel etc...). :cry:

nixpanic:
@mergifyio rebase

@iPraveenParihar , is there anything blocking this PR from your point of view?

iPraveenParihar (dismissed):

nixpanic:
@mergifyio queue

nixpanic:
It looks like the **ci/centos/mini-e2e-helm/k8s-1.30** and **ci/centos/mini-e2e/k8s-1.32** CI jobs failed during `check static PVC with FsName`, with
```
  I0213 09:41:19.618304   22892 utils.go:266] ID: 100 Req-ID: pv-name GRPC call: /csi.v1.Node/NodeStageVolume
  I0213 09:41:19.618350   22892 utils.go:267] ID: 100 Req-ID: pv-name GRPC request: {"secrets":"***stripped***","staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/cephfs.csi.ceph.com/a0dd44413b10ee8f81629ecf349e45d15232fdfb63323715b60e6bbdb7f50dae/globalmount","volume_capability":{"access_mode":{"mode":"SINGLE_NODE_MULTI_WRITER"},"mount":{}},"volume_context":{"clusterID":"d1738887-c6ec-4570-bd3c-85bd6e09a8ab","fsName":"myfs","rootPath":"/volumes/testGroup/testSubVol/13f7794a-ff38-40fd-a599-91863996ac33","staticVolume":"true"},"volume_id":"pv-name"}
  I0213 09:41:19.618516   22892 volumemounter.go:126] requested mounter: , chosen mounter: kernel
  I0213 09:41:19.618594   22892 nodeserver.go:358] ID: 100 Req-ID: pv-name cephfs: mounting volume pv-name with Ceph kernel client
  I0213 09:41:19.618641   22892 crushlocation.go:41] CRUSH location labels passed for processing: [topology.kubernetes.io/region topology.kubernetes.io/zone]
  I0213 09:41:19.618656   22892 crushlocation.go:73] list of CRUSH location processed: map[region:east zone:east-zone1]
  E0213 09:41:19.618671   22892 nodeserver.go:368] ID: 100 Req-ID: pv-name failed to mount volume pv-name: failed to get fsID, stop mounting: cluster not connected yet Check dmesg logs if required.
  E0213 09:41:19.618694   22892 utils.go:271] ID: 100 Req-ID: pv-name GRPC error: rpc error: code = Internal desc = failed to get fsID, stop mounting: cluster not connected yet
```

MageekChiu:
I think this error is happening because `NewVolumeOptionsFromStaticVolume` did not call `Connect(cr *util.Credentials)` , should I add it?
And maybe `NewVolumeOptionsFromMonitorList ` as well.
Thank you @nixpanic @iPraveenParihar

iPraveenParihar:
> I think this error is happening because `NewVolumeOptionsFromStaticVolume` did not call `Connect(cr *util.Credentials)` , should I add it?
> And maybe `NewVolumeOptionsFromMonitorList ` as well.

@MageekChiu, yes the `Connect()` call is missing in both above mentioned calls. Please add it.

MageekChiu:
The fs_name is mandatory for new syntax. I'll have to find out all the possible missing fs_name.

MageekChiu:
Hi @nixpanic, I'm wondering why I‘m not able to open the detail page of the failed tests now?

MageekChiu:
Guess we'll be removing the test ["check static PVC with without FsName"](https://github.com/ceph/ceph-csi/blob/e536333e4c828d090e92b4b153e0548588241ce2/e2e/cephfs.go#L485C8-L485C44) since fsName is mandatory now. And we should also check fsName with `extractOption` rather than `extractOptionalOption`.

nixpanic:
> Hi @nixpanic, I'm wondering why I‘m not able to open the detail page of the failed tests now?

That can happen when the job results are garbage collected (after a week and/or many other test runs). I have started the job again, let's see how it goes now.

nixpanic (approved):

iPraveenParihar (approved):

iPraveenParihar:
@Mergifyio rebase

iPraveenParihar:
@Mergifyio queue

iPraveenParihar:
seems like the `ok-to-test` label is not working here?
trying to remove and add manually, if that works.

iPraveenParihar:
> seems like the `ok-to-test` label is not working here? trying to remove and add manually, if that works.

Add Comment action fails
```
peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043 is not allowed to be used in ceph/ceph-csi. Actions in this workflow must be: within a repository owned by ceph, created by GitHub, or matching the following: ....
```

adding comments manually for now.

iPraveenParihar:
ci/centos/mini-e2e-helm/k8s-1.31

iPraveenParihar:
@Mergifyio requeue

iPraveenParihar:
@Mergifyio requeue

PR #5092, opened 2025-01-20
cleanup: use `rbd.Manager` for fetching volumes in CSI-Addons operations

nixpanic:
ReclaimSpace and EncryptionKeyRotation now use the `Volume` type instead of the (supposed to be) internal `rbdVolume`.

---

<details>

</details>

### DIFF
--- a/internal/csi-addons/rbd/encryptionkeyrotation.go
+++ b/internal/csi-addons/rbd/encryptionkeyrotation.go
@@ -32,11 +32,15 @@ import (
 
 type EncryptionKeyRotationServer struct {
 	*ekr.UnimplementedEncryptionKeyRotationControllerServer
+	driver  string
 	volLock *util.VolumeLocks
 }
 
-func NewEncryptionKeyRotationServer(volLock *util.VolumeLocks) *EncryptionKeyRotationServer {
-	return &EncryptionKeyRotationServer{volLock: volLock}
+func NewEncryptionKeyRotationServer(driver string, volLock *util.VolumeLocks) *EncryptionKeyRotationServer {
+	return &EncryptionKeyRotationServer{
+		driver:  driver,
+		volLock: volLock,
+	}
 }
 
 func (ekrs *EncryptionKeyRotationServer) RegisterService(svc grpc.ServiceRegistrar) {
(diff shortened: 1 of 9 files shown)


internal/rbd/types/volume.go
```diff
@@ -25,9 +25,48 @@ import (
...
+	// checking snapshots limit and clone depth limit and flatten it
+	// if required.
+	PrepareVolumeForSnapshot(ctx context.Context, cr *util.Credentials) error
+
+	// HandleParentImageExistence checks the image's parent.
+	// if the parent image does not exist and is not in trash, it returns nil.
+	// if the flattenMode is FlattenModeForce, it flattens the image itself.
+	// if the parent image is in trash, it returns an error.
+	// if the parent image exists and is not enabled for mirroring, it returns an error.
+	HandleParentImageExistence(ctx context.Context, flattenMode FlattenMode) error
```
Rakshith-R: This is required for csi addons replication enable operation & group add operation only and not for CSI spec calls.

Rakshith-R (dismissed):
LGTM, thanks !

Madhu-1 (dismissed):

Madhu-1:
@mergifyio queue

internal/csi-addons/rbd/encryptionkeyrotation.go
```diff
@@ -32,11 +32,15 @@ import (
 
 type EncryptionKeyRotationServer struct {
 	*ekr.UnimplementedEncryptionKeyRotationControllerServer
+	driver  string
```
iPraveenParihar: this is `csiID`, naming it as driver causes confusion to me. And also everywhere else we have `csiID` naming lets stick to that?

internal/csi-addons/rbd/reclaimspace.go
```diff
@@ -37,13 +37,18 @@ import (
 // of CSI-addons reclaimspace controller service spec.
 type ReclaimSpaceControllerServer struct {
 	*rs.UnimplementedReclaimSpaceControllerServer
+
+	driver      string
```
iPraveenParihar: here as well

iPraveenParihar (changes requested):
requested changes

internal/csi-addons/rbd/encryptionkeyrotation.go
nixpanic: `csiID` is a very strangely called variable. In the CSI Specification it is the `name` returned in the [IdentityService.GetPluginInfo operation](https://github.com/container-storage-interface/spec/blob/master/spec.md#getplugininfo). It is not so much an ID, it is more the name of the driver (for its current deployment). Even in Kubernetes `name` is what is used, see `deploy/rbd/kubernetes/csidriver.yaml`.

`csiID` is also used for variables of the `util.CSIIdentifier` type, and that is why I prefer a different variable name for the driver-instance.

I'll send a cleanup to rename the `csiID` variable for the driver-instance, and will call it `driverName` for clarity.

nixpanic: Oh, wait, in most cases it is actually the `instanceID` for the driver to allow multi-tenancy on the same Ceph cluster (it is used as the location of the journal where Ceph-CSI stores its metadata). I'll call it `driverInstance` instead.

iPraveenParihar (approved):
LGTM. Thanks!

Rakshith-R (approved):

Madhu-1 (approved):

Madhu-1:
@mergifyio queue

PR #4464, opened 2024-02-27
fscrypt: create a new blank key sized according to the passphrase

mgfritch:
Padding a passphrase with null chars to arrive at a 32-byte length
later forces a user to also pass null chars via the term when
attempting to manually unlock a subvolume via the fscrypt cli tools.

This also had a side-effect of truncating any longer length passphrase
down to a shorter 32-byte length.

fixup for:
https://github.com/ceph/ceph-csi/commit/cfea8d756231e7eda5150e33a7f5711cfcf3aafb
https://github.com/ceph/ceph-csi/commit/dd0e1988c08e75ce9564854eab93bb0826a9532d

## Is there anything that requires special attention ##

Backward compatibility is provided to unlock subvolumes that might have been created using the old style null-padded passphrase

This PR also includes an additional fix for a related issue where fscrypt was allowed to indefinitely retry the `keyFn` after an auth failure, which can block subsequent requests to the ceph csi driver.

## Related issues ##

n/a

## Future concerns ##

List items that are not part of the PR and do not impact it's
functionality, but are work items that can be taken up subsequently.

**Checklist:**

* [x] **Commit Message Formatting**: Commit titles and messages follow
  guidelines in the [developer
  guide](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#commit-messages).
* [x] Reviewed the developer guide on [Submitting a Pull
  Request](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#development-workflow)
* [ ] [Pending release
  notes](https://github.com/ceph/ceph-csi/blob/devel/PendingReleaseNotes.md)
  updated with breaking and/or notable changes for the next major release.
* [x] Documentation has been updated, if necessary.
* [ ] Unit tests have been added, if necessary.
* [ ] Integration tests have been added, if necessary.

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

* `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

PR #5115, opened 2025-01-29
Add better failure handling in case CephFS with fscrypt fails

nixpanic:
It seems that fscrypt expects a key with exactly 32 bytes. In order to
use a random length key from a KMS, either repeat the key until the
length is reached, or trim the key when needed.

See: https://github.com/google/fscrypt/tree/v0.3.4#using-a-raw-key-protector
Test-container-image: `quay.io/nixpanic/cephcsi:rhstor-6834`

---

<details>

</details>

### DIFF
--- a/internal/util/fscrypt/fscrypt.go
+++ b/internal/util/fscrypt/fscrypt.go
@@ -217,12 +217,14 @@ func initializeAndUnlock(
 	}
 
 	protector, err := fscryptactions.CreateProtector(fscryptContext, protectorName, keyFn, owner)
-	if err != nil {
+	if err != nil && protector != nil {
 		log.ErrorLog(ctx, "fscrypt: protector name=%s create failed: %v. reverting.", protectorName, err)
 		if revertErr := protector.Revert(); revertErr != nil {
 			return revertErr
 		}
 
+		return err
+	} else if err != nil {
 		return err
 	}
 
--- a/internal/util/fscrypt/fscrypt.go
+++ b/internal/util/fscrypt/fscrypt.go
@@ -60,7 +60,8 @@ var policyV2Support = []util.KernelVersion{
 
 // error values
 var (
-	ErrBadAuth = errors.New("key authentication check failed")
+	ErrBadAuth         = errors.New("key authentication check failed")
+	ErrEmptyPassphrase = errors.New("empty passphrase given")
 )
 
 func AppendEncyptedSubdirectory(dir string) string {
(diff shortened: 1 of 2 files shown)


internal/util/fscrypt/fscrypt.go
```diff
@@ -217,12 +241,14 @@ func initializeAndUnlock(
 	}
 
 	protector, err := fscryptactions.CreateProtector(fscryptContext, protectorName, keyFn, owner)
-	if err != nil {
+	if err != nil && protector != nil {
```
Madhu-1, line 244: i looked at the CreateProtector it will always return the protector as nil when there is an error, we don't need a check for the protector if there is an error as nothing is to be reverted for that protector we can remove below code `protector.Revert()`  as well

internal/util/fscrypt/fscrypt.go
```diff
@@ -94,14 +95,35 @@ func getPassphrase(ctx context.Context, encryption util.VolumeEncryption, volID
 	return passphrase, nil
 }
 
+// resizePassphrase makes sure that the given passphrase will be of [size]
+// bytes. In case the passphrase is shorter, it will be repeated as many times
+// as needed. When a passphrase is (or becomes) longer than the requested
+// [size], the passphrase in truncated.
+func resizePassphrase(passphrase string, size int) (string, error) {
```
Madhu-1, line 102: i assume this function is added to make sure we always get the size of 32 which is required what will happen if the passphrase size is 8 or something even if its added it will not become 32?

Madhu-1 (commented):
32 limit is required for `SourceType_raw_key` and it going to be same requirement for `SourceType_custom_passphrase` because in cephcsi both are supported, The current change might affect both? i have checked the complete code but just mentioned so that we consider it as well. 


@nixpanic in CI we have encryption disabled due to dependency, do we still need to keep it disabled or we can enable it as its good time to do it (if possible)

internal/util/fscrypt/fscrypt.go
nixpanic: See the added unit test `TestResizePassphrase`, I think it works as described, no matter what size of password is used.

internal/util/fscrypt/fscrypt.go
nixpanic: Ah, good, thanks!

internal/util/fscrypt/fscrypt.go
Madhu-1: Can we add a case where the passphrase length is 3 and the size is 7 and validate returned passphrase length is matching the size?

internal/util/fscrypt/fscrypt.go
```diff
@@ -112,9 +134,11 @@ func createKeyFuncFromVolumeEncryption(
 			return nil, err
 		}
 
-		if keySize < 0 {
-			keySize = len(passphrase)
+		passphrase, err = resizePassphrase(passphrase, keySize)
+		if err != nil {
+			return nil, err
```
Madhu-1, line 139: can we also check wont this change break the change added in https://github.com/ceph/ceph-csi/pull/4464?

internal/util/fscrypt/fscrypt.go
nixpanic: Sure, added them now.

internal/util/fscrypt/fscrypt.go
nixpanic: yeah, I still need to figure out how the different protector is created, and then only do the password resizing for the _raw key protector(s)_

PR #5116, opened 2025-01-30
rbd: log formatting and go-ceph deprecation in `constructImageOptions`

nixpanic:
When a `dataPool` is passed while creating a volume, there is a
`%!s(MISSING)` piece added to a debug log message. By using
`fmt.Sprintf()` instead of concatinating the string, this should be gone
now.

go-ceph also deprecated `Rbd` prefix from go-ceph `rbd.ImageOption*`
long ago, so we should not use those anymore.

---

<details>

</details>


Madhu-1 (approved):

iPraveenParihar (approved):

nixpanic:
@mergifyio queue

PR #5137, opened 2025-02-11
build: reinstall the ceph-release package

nixpanic:
With the new Ceph container-build process, the .repo files for DNF are
removed. This means that the build containers can not install required
dependencies anymore, like librados-devel and librbd-devel.

By reinstalling the ceph-release package from the Ceph distribution
server, the .repo files are restored and the required Ceph packages can
be installed again.

Disabled NFS e2e test reported in #5141.


Madhu-1 (approved):
@nixpanic can you please open a tracker for nfs bug?

yati1998 (approved):

PR #5184, opened 2025-02-28
RBD: check for volume group existence

iPraveenParihar:

This PR adds a check for Volume Group existence in `generateVolumeGroup` call.

**Checklist:**

---

<details>

</details>

### DIFF
--- a/internal/csi-addons/rbd/encryptionkeyrotation.go
+++ b/internal/csi-addons/rbd/encryptionkeyrotation.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 
 	"github.com/ceph/ceph-csi/internal/rbd"
+	rbderrors "github.com/ceph/ceph-csi/internal/rbd/errors"
 	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/log"
 
@@ -68,7 +69,7 @@ func (ekrs *EncryptionKeyRotationServer) EncryptionKeyRotate(
 	rbdVol, err := mgr.GetVolumeByID(ctx, volID)
 	if err != nil {
 		switch {
-		case errors.Is(err, util.ErrImageNotFound):
+		case errors.Is(err, rbderrors.ErrImageNotFound):
 			err = status.Errorf(codes.NotFound, "volume ID %s not found", volID)
 		case errors.Is(err, util.ErrPoolNotFound):
 			log.ErrorLog(ctx, "failed to get backend volume for %s: %v", volID, err)
--- a/internal/rbd/errors/errors.go
+++ b/internal/rbd/errors/errors.go
@@ -14,11 +14,19 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package rbd
+package rbderrors
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+
+	librados "github.com/ceph/go-ceph/rados"
+	librbd "github.com/ceph/go-ceph/rbd"
+)
 
 var (
+	// ErrImageNotFound is returned when image name is not found in the cluster on the given pool and/or namespace.
+	ErrImageNotFound = errors.New("image not found")
 	// ErrSnapNotFound is returned when snap name passed is not found in the list of snapshots for the
 	// given image.
 	ErrSnapNotFound = errors.New("snapshot not found")
@@ -55,4 +63,8 @@ var (
 	ErrInvalidArgument = errors.New("invalid arguments provided")
 	// ErrImageInUse is returned when the image is in use.
 	ErrImageInUse = errors.New("image is in use")
+	// ErrGroupNotConnected is returned when the RBD group is not connected.
+	ErrGroupNotConnected = fmt.Errorf("%w: RBD group is not connected", librados.ErrNotConnected)
+	// ErrGroupNotFound is returned when group is not found in the cluster on the given pool and/or namespace.
+	ErrGroupNotFound = fmt.Errorf("%w: RBD group not found", librbd.ErrNotFound)
 )
--- a/internal/rbd/group/util.go
+++ b/internal/rbd/group/util.go
@@ -66,7 +66,7 @@ type commonVolumeGroup struct {
 }
 
 // generateVolumeGroup generates a commonVolumeGroup structure from the volumeGroup identifier.
-func (cvg *commonVolumeGroup) generateVolumeGroup(csiID util.CSIIdentifier) error {
+func (cvg *commonVolumeGroup) generateVolumeGroup(ctx context.Context, csiID util.CSIIdentifier) error {
 	mons, err := util.Mons(util.CsiConfigFile, csiID.ClusterID)
 	if err != nil {
 		return fmt.Errorf("failed to get MONs for cluster id %q: %w", csiID.ClusterID, err)
@@ -86,16 +86,25 @@ func (cvg *commonVolumeGroup) generateVolumeGroup(csiID util.CSIIdentifier) erro
 	cvg.namespace = namespace
 	cvg.pool = pool
 
+	// Check if the volume group exists in the journal.
+	_, err = cvg.getVolumeGroupAttributes(ctx)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
 // generateVolumeGroupFromMapping checks the clusterID and poolID mapping and
 // generates commonVolumeGroup structure for the mapped clusterID and poolID.
+// If the mapping is not found, it returns ErrGroupNotFound.
 func (cvg *commonVolumeGroup) generateVolumeGroupFromMapping(
 	ctx context.Context,
 	csiID util.CSIIdentifier,
 	mapping *[]util.ClusterMappingInfo,
 ) error {
+	var volumeGroupGenerationError error
+
 	mcsiID := csiID
 	existingClusterID := csiID.ClusterID
 	existingPoolID := strconv.FormatInt(csiID.LocationID, 10)
@@ -126,18 +135,18 @@ func (cvg *commonVolumeGroup) generateVolumeGroupFromMapping(
 						return err
 					}
 					mcsiID.LocationID = mPID
-					err = cvg.generateVolumeGroup(mcsiID)
-					if ShouldRetryVolumeGroupGeneration(err) {
-						continue
+					volumeGroupGenerationError = cvg.generateVolumeGroup(ctx, mcsiID)
+					if !ShouldRetryVolumeGroupGeneration(volumeGroupGenerationError) {
+						return volumeGroupGenerationError
 					}
 
-					return err
+					log.DebugLog(ctx, "volume group not found for poolID mapping %s: %v", cvg.id, mappedPoolID)
 				}
 			}
 		}
 	}
 
-	return util.ErrPoolNotFound
+	return rbderrors.ErrGroupNotFound
 }
 
 func (cvg *commonVolumeGroup) initCommonVolumeGroup(
@@ -146,6 +155,7 @@ func (cvg *commonVolumeGroup) initCommonVolumeGroup(
 	csiDriver string,
 	creds *util.Credentials,
 ) error {
+	var volumeGroupGenerationError error
 	csiID := util.CSIIdentifier{}
 
 	err := csiID.DecomposeCSIID(id)
@@ -160,15 +170,15 @@ func (cvg *commonVolumeGroup) initCommonVolumeGroup(
 	cvg.objectUUID = csiID.ObjectUUID
 	// cvg.monitors, cvg.namespace, cvg.pool are set in generateVolumeGroup
 
-	err = cvg.generateVolumeGroup(csiID)
+	volumeGroupGenerationError = cvg.generateVolumeGroup(ctx, csiID)
 	// If the error is not a retryable error, return from here.
-	if err != nil && !ShouldRetryVolumeGroupGeneration(err) {
-		return err
+	if volumeGroupGenerationError != nil && !ShouldRetryVolumeGroupGeneration(volumeGroupGenerationError) {
+		return volumeGroupGenerationError
 	}
 
-	// If the error is a retryable error, we should try to get the cluster mapping
-	// and generate the volume group from the mapping.
-	if ShouldRetryVolumeGroupGeneration(err) {
+	// If Volume Group doesn't exists or the error is a retryable error,
+	// we should try to get the cluster mapping and generate the volume group from the mapping.
+	if ShouldRetryVolumeGroupGeneration(volumeGroupGenerationError) {
 		mapping, err := util.GetClusterMappingInfo(csiID.ClusterID)
 		if err != nil {
 			return err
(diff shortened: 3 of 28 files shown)


internal/rbd/group/util.go
```diff
@@ -85,7 +85,10 @@ func (cvg *commonVolumeGroup) generateVolumeGroup(csiID util.CSIIdentifier) erro
 	cvg.namespace = namespace
 	cvg.pool = pool
 
-	return nil
+	// Check if the volume group exists, by fetching the volume group attributes.
+	_, err = cvg.getVolumeGroupAttributes(context.Background())
```
nixpanic: move this to a new `Exists(context.Context) (bool, error)` function in the VolumeGroup interface, and call it only when/where needed

internal/util/errors.go
```diff
@@ -66,5 +72,5 @@ func ShouldRetryVolumeGeneration(err error) bool {
 	return (errors.Is(err, ErrKeyNotFound) ||
 		errors.Is(err, ErrPoolNotFound) ||
 		errors.Is(err, ErrImageNotFound) ||
-		errors.Is(err, rados.ErrPermissionDenied))
+		errors.Is(err, rados.ErrPermissionDenied) || errors.Is(err, ErrRBDGroupNotFound))
```
nixpanic: please keep the coding style the same, add the new check on its own line

internal/rbd/group/util.go
iPraveenParihar: Thanks!, for the early review 😀

iPraveenParihar: Done

internal/util/errors.go
```diff
@@ -66,5 +72,6 @@ func ShouldRetryVolumeGeneration(err error) bool {
 	return (errors.Is(err, ErrKeyNotFound) ||
 		errors.Is(err, ErrPoolNotFound) ||
 		errors.Is(err, ErrImageNotFound) ||
-		errors.Is(err, rados.ErrPermissionDenied))
+		errors.Is(err, rados.ErrPermissionDenied) ||
+		errors.Is(err, ErrRBDGroupNotFound))
```
Rakshith-R: Why is group not found error here?

Please create another function for retrying volgroup generation

Rakshith-R (changes requested):

internal/util/errors.go
```diff
@@ -40,6 +42,10 @@ var (
 	ErrClusterIDNotSet = errors.New("clusterID must be set")
 	// ErrMissingConfigForMonitor is returned when clusterID is not found for the mon.
 	ErrMissingConfigForMonitor = errors.New("missing configuration of cluster ID for monitor")
+	// ErrRBDGroupNotFound is returned when group name is not found in the cluster on the given pool and/or namespace.
```
nixpanic: The `internal/util` package should ideally not have references to RBD, CephFS or NFS. There needs to be a good explanation and advantage to do so. I do not see neither at the moment.

Possibly a `internal/rbd/errors` package would be more appropriate, you could collect all RBD errors there? But it really depends on the reason why you introduced this change.

internal/rbd/group/util.go
```diff
@@ -85,7 +85,9 @@ func (cvg *commonVolumeGroup) generateVolumeGroup(csiID util.CSIIdentifier) erro
 	cvg.namespace = namespace
 	cvg.pool = pool
 
-	return nil
+	_, err = cvg.Exists(context.Background())
```
nixpanic: Don't call this here, call it at a location that wants to check if the volume group exists.

internal/rbd/group/util.go
```diff
@@ -417,3 +419,13 @@ func (cvg *commonVolumeGroup) GetCreationTime(ctx context.Context) (*time.Time,
 
 	return cvg.creationTime, nil
 }
+
+// Exists checks if the volume group exists in the journal.
+func (cvg *commonVolumeGroup) Exists(ctx context.Context) (bool, error) {
```
nixpanic: Add this function to the VolumeGroup interface, so that any caller can use it.

internal/util/errors.go
```diff
@@ -66,5 +72,31 @@ func ShouldRetryVolumeGeneration(err error) bool {
...
+//   - rados.ErrPermissionDenied: Permissions to access the pool is denied.
+//
+// If any of these errors are encountered, the function returns `true`, indicating
+// that the volumegroup search should continue because of known error. Otherwise, it
+// returns `false`, meaning the search should stop.
+//
+// This helper function is used in scenarios where multiple attempts may be made
+// to retrieve or generate volumegroup information, and we want to gracefully handle
+// specific failure cases while retrying for others.
+func ShouldRetryVolumeGroupGeneration(err error) bool {
```
nixpanic: add a unit-test for this please

internal/util/errors.go
```diff
@@ -66,5 +72,31 @@ func ShouldRetryVolumeGeneration(err error) bool {
...
+		errors.Is(err, ErrRBDGroupNotFound))
+}
+
+// ShouldRetryVolumeGroupGeneration determines whether the process of finding or generating
+// volumegroups should continue based on the type of error encountered.
+//
+// It checks if the given error matches any of the following known errors:
+//   - ErrPoolNotFound: The rbd pool where the volumegroup/omap is expected doesn't exist.
+//   - ErrRBDGroupNotFound: The volumegroup doesn't exist in the rbd pool.
+//   - rados.ErrPermissionDenied: Permissions to access the pool is denied.
```
nixpanic: Why retry on PermissionDenied? This usually is a pretty final error.

iPraveenParihar: The same check is being done for RBD volumes - _fyi: https://github.com/ceph/ceph-csi/commit/88b964fe18d37af5b53cea00e8d785159fb61f59_

nixpanic: Thanks for the pointer. That commit message contains a description of the _why_, ideally that is included in the comment, but in the commit message it is ok too.

nixpanic:
@iPraveenParihar , the new `Exists()` function in the interface causes the **golangci-lint** to fail. As all interfaces have a `Destroy()` function, you could move that one to a base interface (`journalledObject` maybe) in a separate commit.

iPraveenParihar:
Waiting on @Nikhil-Ladha, for the testing result.

Nikhil-Ladha:
> Waiting on @Nikhil-Ladha, for the testing result.

Tested with different pool ids on both the clusters (and the cluster had other pools too) and, everything looks good.
We can probably optimize the code to avoid dual checks for omap data, but it can be a part of different PR as well.

iPraveenParihar:
> Tested with different pool ids on both the clusters (and the cluster had other pools too) and, everything looks good. We can probably optimize the code to avoid dual checks for omap data, but it can be a part of different PR as well.

Thanks @Nikhil-Ladha for the testing! I could think of using the `rbd_group_get_id` from [ceph/ceph](https://github.com/ceph/ceph/blob/812add83d5c135d230f5fc12b41dfa20a5d1dfa1/src/librbd/librbd.cc#L7124C16-L7124C32) but not yet available in go-ceph for now. We could pick it up once we have it in go-ceph.

internal/csi-addons/rbd/encryptionkeyrotation.go
```diff
@@ -21,6 +21,7 @@ import (
 	"errors"
 
 	"github.com/ceph/ceph-csi/internal/rbd"
+	rbd_errors "github.com/ceph/ceph-csi/internal/rbd/errors"
```
Madhu-1: lets not use `_` use `rbderrors`

internal/rbd/errors/errors.go
```diff
@@ -55,4 +63,8 @@ var (
 	ErrInvalidArgument = errors.New("invalid arguments provided")
 	// ErrImageInUse is returned when the image is in use.
 	ErrImageInUse = errors.New("image is in use")
+	// ErrRBDGroupNotConnected is returned when the RBD group is not connected.
+	ErrRBDGroupNotConnected = fmt.Errorf("%w: RBD group is not connected", librados.ErrNotConnected)
+	// ErrRBDGroupNotFound is returned when group name is not found in the cluster on the given pool and/or namespace.
```
Madhu-1:
```suggestion
// ErrRBDGroupNotFound is returned when group is not found in the cluster on the given pool and/or namespace.
```

Madhu-1 (dismissed):
LGTM

internal/rbd/group/util.go
```diff
@@ -125,18 +133,27 @@ func (cvg *commonVolumeGroup) generateVolumeGroupFromMapping(
...
 
-					return err
+					return volumeGroupGenerationError
 				}
 			}
 		}
 	}
 
-	return util.ErrPoolNotFound
+	return volumeGroupGenerationError
```
Rakshith-R: I think this is too complex.
Can you try to simplify it ?

Rakshith-R (changes requested):

internal/rbd/errors/errors.go
```diff
@@ -55,4 +63,8 @@ var (
 	ErrInvalidArgument = errors.New("invalid arguments provided")
 	// ErrImageInUse is returned when the image is in use.
 	ErrImageInUse = errors.New("image is in use")
+	// ErrRBDGroupNotConnected is returned when the RBD group is not connected.
+	ErrRBDGroupNotConnected = fmt.Errorf("%w: RBD group is not connected", librados.ErrNotConnected)
```
nixpanic: This is in a RBD sub-package, drop the RBD part of all variables, make it `ErrGroupNotConnected`.

internal/rbd/types/group.go
```diff
@@ -37,15 +37,15 @@ type journalledObject interface {
 
 	// GetClusterID returns the ID of the cluster of the object.
 	GetClusterID(ctx context.Context) (string, error)
+
+	// Destroy frees the resources used by the VolumeGroup.
```
nixpanic: VolumeGroup => object

nixpanic (changes requested):

internal/rbd/group/util.go
Rakshith-R: @iPraveenParihar Can't Exists() be called inside generatevolgroup ? 

I think the Exists logic can be fit inside generateVolGroup

iPraveenParihar: I think @nixpanic had asked to move it out from genereateVolumeGroup  - https://github.com/ceph/ceph-csi/pull/5184#discussion_r1975320212

internal/rbd/group/util.go
Rakshith-R: > Don't call this here, call it at a location that wants to check if the volume group exists.

@iPraveenParihar @nixpanic 
I don't think we need to have a function Exists at all. Its called at only two places, that too only after generateVolGroup call 

https://github.com/ceph/ceph-csi/blob/9eb4f61d015696171986f6167a512658d09aec2c/internal/rbd/group/util.go#L474-L482

internal/rbd/group/util.go
iPraveenParihar: @Rakshith-R, I tried making it simpler by moving the group exist check code to `generateVolumeGroup` function.

Rakshith-R (dismissed):
LGTM, Thanks

internal/csi-addons/rbd/encryptionkeyrotation.go
```diff
@@ -21,6 +21,7 @@ import (
 	"errors"
 
 	"github.com/ceph/ceph-csi/internal/rbd"
+	rbderrors "github.com/ceph/ceph-csi/internal/rbd/errors"
```
nixpanic, line 24: maybe it makes sense to move common RBD errors to `internal/rbd/types/errors.go`? That is what I decided to do with #5221.

internal/rbd/group/util.go
```diff
@@ -86,16 +86,25 @@ func (cvg *commonVolumeGroup) generateVolumeGroup(csiID util.CSIIdentifier) erro
 	cvg.namespace = namespace
 	cvg.pool = pool
 
+	// Check if the volume group exists in the journal.
+	_, err = cvg.getVolumeGroupAttributes(ctx)
```
nixpanic, line 90: Does this need to be called here? The initialization of the struct does not need to guarantee that the volume group is recorded in the journal.

internal/rbd/group/util.go
```diff
@@ -86,16 +86,25 @@ func (cvg *commonVolumeGroup) generateVolumeGroup(csiID util.CSIIdentifier) erro
...
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
 // generateVolumeGroupFromMapping checks the clusterID and poolID mapping and
 // generates commonVolumeGroup structure for the mapped clusterID and poolID.
+// If the mapping is not found, it returns ErrRBDGroupNotFound.
```
nixpanic: the comment still contains _ErrRBDGroupNotFound_ -> _ErrGroupNotFound_

internal/rbd/group/util.go
iPraveenParihar: I tried to keep it similar to what we have for _[generateVolumeFromVolumeID](https://github.com/ceph/ceph-csi/blob/91baf7681d84fdd450648004764ee1bea419207d/internal/rbd/rbd_util.go#L1145)_.  Here to, we initialise the rbdVol struct and fetch details from journal. Calling outside, made it little complex for readability.

nixpanic: Ok. I just wonder about possible side-effects in case a volumegroup does not exist, or was in the process of getting deleted. I think it should be fine, but maybe you can check that?

internal/csi-addons/rbd/encryptionkeyrotation.go
iPraveenParihar: IMO, `rbd/errors` is more suitable than `rbd/types` _(which should be just for the interface and struct definition?)_

internal/rbd/group/util.go
iPraveenParihar: >  just wonder about possible side-effects in case a volumegroup does not exist

`initCommonVolumeGroup` is only been called after a VolumeGroup is created, we are good here.

> was in the process of getting deleted

operation locks will prevent this case, but actually I don't see operationa locks on `/DeleteVolumeGroup` & `/CreateVolumeGroup`, was that missed or not required at all?

Nikhil-Ladha: > initCommonVolumeGroup is only been called after a VolumeGroup is created, we are good here.

Actually, we can. For example, if we try to delete an already deleted (or non-existent) volume. Though, looking at the code it would still fail with the same error i.e, `ErrGroupNotFound` but after going through some more function calls.

internal/csi-addons/rbd/encryptionkeyrotation.go
nixpanic: That is fine with me too. I just wanted to locate the errors that are documented at the types/interfaces together. I' can move it to `internal/rbd/errors` too.

internal/rbd/group/util.go
iPraveenParihar: Yes, that should be fine.

Rakshith-R (approved):

nixpanic (approved):

nixpanic:
@mergifyio rebase

Issue #4633, opened 2024-05-23
e2e: corrupted mount not attempting to recover.


iPraveenParihar:
> > replacing, `k8s.io/mount-utils => k8s.io/mount-utils v0.29.3`
> 
> this works!, So, will add a `TODO: update once fixed in mount-utils` and visit later? @Madhu-1

@iPraveenParihar Thank you, lets get it merged and we will revisit mount-utils, lets check what is changed in mount-utils release notes.

_Originally posted by @Madhu-1 in https://github.com/ceph/ceph-csi/issues/4614#issuecomment-2126600054_

iPraveenParihar:
There is a change in the IsLikelyNotMountPoint method: https://github.com/kubernetes/mount-utils/compare/v0.29.3...v0.30.2

From the method description, it is clear that IsLikelyNotMountPoint is not suitable for detecting bind mounts: https://github.com/kubernetes/mount-utils/blob/dfd453c575c09ba21aa9f45bf9a9ddb8724c5e0b/mount_linux.go#L498-L513

ceph-csi uses this method for fuse mount recovery to determine if the path is a valid MountPoint or not. https://github.com/ceph/ceph-csi/blob/202f43c82d63d37bd765dea7ecf8de4037eca7b6/internal/util/util.go#L327-L335

we need alternative method to be used?
thoughts @nixpanic @Madhu-1 ?

nixpanic:
Probably you can delete this function and call `x.mounter.IsMountPoint(...)` directly? I don't see why the indirection is beneficial for anything.

iPraveenParihar:
> Probably you can delete this function and call `x.mounter.IsMountPoint(...)` directly? I don't see why the indirection is beneficial for anything.

`IsMountPoint(...)` is an expensive operation as it enumerates mount points on the Node and then does the evaluate operation on each mount point.
https://github.com/kubernetes/mount-utils/blob/dfd453c575c09ba21aa9f45bf9a9ddb8724c5e0b/mount.go#L72-L79

may be because of that, they chose `IsLikelyNotMountPoint()`?

Madhu-1:
> > Probably you can delete this function and call `x.mounter.IsMountPoint(...)` directly? I don't see why the indirection is beneficial for anything.
> 
[...]

PR #5188, opened 2025-03-04
build: move e2e dependencies into `e2e/go.mod`

nixpanic:
Several packages are only used while running the e2e suite. These
packages are less important to update, as the they can not influence the
final executable that is part of the Ceph-CSI container-image.

By moving these dependencies out of the main Ceph-CSI go.mod, it is
easier to identify if a reported CVE affects Ceph-CSI, or only the
testing (like most of the Kubernetes CVEs).


e2e/rbd.go
```diff
@@ -23,7 +23,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/ceph/ceph-csi/internal/util"
+	"github.com/ceph/ceph-csi/pkg/util/crypto"
```
Rakshith-R, line 26: This line should have been included in the previous commit.

nixpanic: Ah! yes.

black-dragon74:
If we do not have plans to use `pkg` for anything else, may I suggest a top level `utils` instead?

nixpanic:
> If we do not have plans to use `pkg` for anything else, may I suggest a top level `utils` instead?

`pkg` is a common name for go projects, so I decided to take that. In the future, we may be able to move the internal `rbd.Manager` to a public `pkg` package too, so that other tools (rook cli, or other debugging/troubleshooting tools) can use it.

black-dragon74 (dismissed):

Rakshith-R (dismissed):

nixpanic:
@mergifyio queue

nixpanic:
Might depend on #5193, as testing with the previous Rook version seems to be more stable?

nixpanic:
> Might depend on #5193, as testing with the previous Rook version seems to be more stable?

Included the Rook downgrade in this PR too now, just to see if the CI jobs pass.

nixpanic:
> /retest ci/centos/mini-e2e/k8s-1.32

Failed with:
```
[FAILED] failed to validate omap count for rados ls --pool=myfs-metadata --namespace=csi | grep -v default | grep -v csi.volume.group. | grep -c ^csi.volume.: expected omap object count 4, got 5
```
[logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.32/detail/mini-e2e_k8s-1.32/109/pipeline)

iPraveenParihar:
> > /retest ci/centos/mini-e2e/k8s-1.32
> 
> Failed with:
> 
> ```
> [FAILED] failed to validate omap count for rados ls --pool=myfs-metadata --namespace=csi | grep -v default | grep -v csi.volume.group. | grep -c ^csi.volume.: expected omap object count 4, got 5
> ```
> 
> [logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.32/detail/mini-e2e_k8s-1.32/109/pipeline)


@nixpanic, all jobs have failed at the CephFS test `verifying that ceph-fuse recovery works for new pods`. The NFS test failures are due to stale entries from the failed CephFS test case, so they can be ignored.

I've seen this test case fail previously with mount-utils versions later than v0.29.3 (reference: https://github.com/ceph/ceph-csi/issues/4633). To address this, I had added k8s.io/mount-utils v0.29.3 in the replace section, but it's missing from the main go.mod file, which I suspect is causing the failure. Could you try adding the replace directive in the main go.mod file and see if that resolves the issue?

nixpanic:
> @nixpanic, all jobs have failed at the CephFS test `verifying that ceph-fuse recovery works for new pods`. The NFS test failures are due to stale entries from the failed CephFS test case, so they can be ignored.
> 
> I've seen this test case fail previously with mount-utils versions later than v0.29.3 (reference: #4633). To address this, I had added k8s.io/mount-utils v0.29.3 in the replace section, but it's missing from the main go.mod file, which I suspect is causing the failure. Could you try adding the replace directive in the main go.mod file and see if that resolves the issue?

Thanks for the suggestion, @iPraveenParihar! The replacing of `mount-utils` is done in the `e2e/go.mod` file, but not anymore in main `go.mod`. I have added that now again, lets see ho things go :crossed_fingers:

nixpanic:
> Thanks for the suggestion, @iPraveenParihar! The replacing of `mount-utils` is done in the `e2e/go.mod` file, but not anymore in main `go.mod`. I have added that now again, lets see ho things go 🤞

Seems to pass with the old mount-utils package!

black-dragon74 (approved):

Rakshith-R (approved):

Rakshith-R:
@mergifyio queue

PR #5203, opened 2025-03-07
util: do not use `mount-utils.IsLikelyNotMountPoint` anymore

### LINKED ISSUE #4633: e2e: corrupted mount not attempting to recover.

nixpanic:
`IsLikelyNotMountPoint()` is an optimized version for `IsMountPoint()`
which can not detect all type of mounts (anymore). The slower
`IsMountPoint()` is more safe to use. This can cause a slight
performance regression in the case there are many mountpoints on the
system, but correctness is more important than speed while mounting.

Fixes: #4633


Madhu-1:
@nixpanic we are using `IsLikelyNotMountPoint` in two other places (nfs,rbd) can we replace it as well?

nixpanic:
> @nixpanic we are using `IsLikelyNotMountPoint` in two other places (nfs,rbd) can we replace it as well?

Sure, done!

internal/nfs/nodeserver/nodeserver.go
```diff
@@ -200,7 +200,7 @@ func (ns *NodeServer) mountNFS(
 		err    error
 	)
 
-	notMnt, err := ns.Mounter.IsLikelyNotMountPoint(mountPoint)
+	notMnt, err := ns.Mounter.IsMountPoint(mountPoint)
```
Madhu-1: we should change `notMnt` to `Ismnt` or something like that and we should reverse the check for it because `IsLikelyNotMountPoint` was returning its a mount point or not as `!` operation of IsMountPoint

Madhu-1 (changes requested):

internal/nfs/nodeserver/nodeserver.go
nixpanic: Ah, yes :facepalm:

nixpanic: This should have been addressed now. Thanks for catching!

internal/nfs/nodeserver/nodeserver.go
```diff
@@ -200,19 +200,19 @@ func (ns *NodeServer) mountNFS(
...
-	notMnt, err := ns.Mounter.IsLikelyNotMountPoint(mountPoint)
+	isMnt, err := ns.Mounter.IsMountPoint(mountPoint)
 	if err != nil {
 		if os.IsNotExist(err) {
 			err = os.MkdirAll(mountPoint, defaultMountPermission)
 			if err != nil {
 				return err
 			}
-			notMnt = true
+			isMnt = false
```
Madhu-1, line 210: i think this check is not required, because of an error, isMnt will be set to false by default

internal/rbd/nodeserver.go
```diff
@@ -880,9 +880,9 @@ func (ns *NodeServer) mountVolume(ctx context.Context, stagingPath string, req *
 
 func (ns *NodeServer) createTargetMountPath(ctx context.Context, mountPath string, isBlock bool) (bool, error) {
```
Madhu-1, line 886: can we update the `createTargetMountPath` to return isMnt to check it simple, returning `!` and doing `!` check on the function return value looks so confusing :(

Madhu-1 (commented):
LGTM, if possible please see if you can address the nits.

internal/rbd/nodeserver.go
nixpanic: Sure, I can clean that up. It confused me a little too.

nixpanic: @Madhu-1 , please have an other look. The 2nd commit simplifies `createTargetMountPath()`.

Madhu-1 (approved):
LGTM

iPraveenParihar (approved):

Rakshith-R:
@mergifyio queue

PR #5206, opened 2025-03-10
rbd: use ListChildrenAttributes() instead of ListChildren()

### LINKED ISSUE #5173: rbd: Avoid flattening child images which are in trash
# Describe the bug #

Currently, `ListChildren()` goceph function lists all child images regardless of their existence in trash.
https://github.com/ceph/ceph-csi/blob/43b150f14d2858f2bdcd268d8fac16377a98b8a4/internal/rbd/rbd_util.go#L2039

https://github.com/ceph/go-ceph/blob/e84ad6c5a35c0514773539c3462af300c3b50584/rbd/snapshot_nautilus.go#L138

We need to make use of new go-ceph API `ListChildrenWithParams` API in order to distinguish child images in trash and avoid trying to flatten them. 
https://github.com/ceph/go-ceph/pull/1079

Rakshith-R:

This commit modifies listSnapAndChildren() to make use of ListChildrenAttributes() instead of ListChildren() which allows us to filter out images in trash.
This commit also order the alive images so that temp clone images are followed by images backing volume snapshots so that temp clone images are flattened first.

Yes

No

## Related issues ##

Fixes: #5173 

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -590,7 +590,7 @@ func (cs *ControllerServer) repairExistingVolume(ctx context.Context, req *csi.C
 // are more than the `minSnapshotOnImage` Add a task to flatten all the
 // temporary cloned images.
 func flattenTemporaryClonedImages(ctx context.Context, rbdVol *rbdVolume, cr *util.Credentials) error {
-	snaps, children, err := rbdVol.listSnapAndChildren()
+	snapAndChildrenInfo, err := rbdVol.listSnapAndChildren()
 	if err != nil {
 		if errors.Is(err, util.ErrImageNotFound) {
 			return status.Error(codes.InvalidArgument, err.Error())
@@ -598,6 +598,17 @@ func flattenTemporaryClonedImages(ctx context.Context, rbdVol *rbdVolume, cr *ut
 
 		return status.Error(codes.Internal, err.Error())
 	}
+	children := make([]string, 0)
+	// order the temp clone images first followed by the volume snapshot
+	// images so that the temp clone images are flattened first.
+	// This is done in order to:
+	// - increase number of snapshots which can be supported for
+	// changed block tracking(rbd snap diff).
+	// - favor scalability since multiple PVCs can be restored from same
+	// volumesnapshot versus just one PVC-PVC clone.
+	children = append(children, snapAndChildrenInfo.TempCloneChildren...)
+	children = append(children, snapAndChildrenInfo.VolumeSnapshotChildren...)
+	snaps := snapAndChildrenInfo.SnapInfoList
 
 	if len(snaps) > int(maxSnapshotsOnImage) {
 		log.DebugLog(
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -84,6 +84,10 @@ const (
 
 	// clusterNameKey cluster Key, set on RBD image.
 	clusterNameKey = "csi.ceph.com/cluster/name"
+
+	// Suffix added to the temp cloned image name.
+	// This will always be (rbd image name + "-temp").
+	tempImageSuffix = "-temp"
 )
 
 // rbdImage contains common attributes and methods for the rbdVolume and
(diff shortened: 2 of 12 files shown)


internal/rbd/rbd_util.go
```diff
@@ -2038,12 +2041,28 @@ func (ri *rbdImage) listSnapAndChildren() ([]librbd.SnapInfo, []string, error) {
 		return nil, nil, err
 	}
 
-	// ListChildren() returns pools, images, err.
-	_, children, err := image.ListChildren()
+	// ListChildrenAttributes() returns child ImageSpec list, err.
```
nixpanic: you can drop this comment, I think the function is rather clear

internal/rbd/rbd_util.go
```diff
@@ -2038,12 +2041,28 @@ func (ri *rbdImage) listSnapAndChildren() ([]librbd.SnapInfo, []string, error) {
...
 		return nil, nil, err
 	}
 
+	children := make([]string, 0)
+	volSnapChildren := make([]string, 0)
+	for _, child := range childImageSpecList {
+		if child.Trash {
+			continue
+		}
+		if strings.HasSuffix(child.ImageName, "-temp") {
```
nixpanic: Maybe consider adding a `isTempImage(child)` function?

internal/rbd/rbd_util.go
```diff
@@ -2038,12 +2041,28 @@ func (ri *rbdImage) listSnapAndChildren() ([]librbd.SnapInfo, []string, error) {
...
+			children = append(children, child.ImageName)
+		} else {
+			volSnapChildren = append(volSnapChildren, child.ImageName)
+		}
+	}
+	// order the temp clone images first followed by the volume snapshots
+	// so that the temp clones are flattened first.
```
nixpanic: This is the 2nd comment about `temp clones are flattened first`. I don't get why that is the case. If that is a requirement of the returned list, mention something about the order. Or wherever the list is consumed, use some other way of inspecting the list. Might be cleaner to return two lists, so that the distinction is very clear.

Rakshith-R: > This is the 2nd comment about `temp clones are flattened first`. I don't get why that is the case. If that is a requirement of the returned list, mention something about the order. Or wherever the list is consumed, use some other way of inspecting the list. Might be cleaner to return two lists, so that the distinction is very clear.

I've removed it from the function definition, since we still are returning full list of child images which are not in trash.

I've expanded the comment in this section with reasons why such ordering is done so it can be understood in the future.

nixpanic: Thanks, that makes things a little bit clearer. I don't understand (yet) if, and why callers need to care about the order in the list that is returned. If callers need to care, or benefit from it, why not return separate lists? Or, a struct like `{ snaps, tempChildren, volSnapChildren }` so that callers of the function are very aware of the differences, and can decide how to handle them?

Rakshith-R: > Thanks, that makes things a little bit clearer. I don't understand (yet) if, and why callers need to care about the order in the list that is returned. If callers need to care, or benefit from it, why not return separate lists? Or, a struct like `{ snaps, tempChildren, volSnapChildren }` so that callers of the function are very aware of the differences, and can decide how to handle them?

I've moved the logic to order images into the caller and added a struct for returned parameters.
PTAL

internal/rbd/rbd_util.go
```diff
@@ -2025,26 +2029,49 @@ func (ri *rbdImage) DisableDeepFlatten() error {
 	return image.UpdateFeatures(librbd.FeatureDeepFlatten, false)
 }
 
-// listSnapAndChildren returns list of names of snapshots and child images.
-func (ri *rbdImage) listSnapAndChildren() ([]librbd.SnapInfo, []string, error) {
+type SnapAndChildrenInfo struct {
```
nixpanic: No need to export this, use `snapAndChildrenInfo` instead.

nixpanic (commented):
Thanks, makes it much cleaner. Just a nit that needs addressing.

internal/rbd/controllerserver.go
```diff
@@ -590,14 +590,23 @@ func (cs *ControllerServer) repairExistingVolume(ctx context.Context, req *csi.C
...
 		return status.Error(codes.Internal, err.Error())
 	}
+	// order the temp clone images first followed by the volume snapshot
+	// images so that the temp clone images are flattened first.
+	// This is done in order to:
+	// - increase number of snapshots which can be supported for
+	// changed block tracking(rbd snap diff).
+	// - favor scalability since multiple PVCs can be restored from same
+	// volumesnapshot versus just one PVC-PVC clone.
+	children := append(snapAndChildrenInfo.TempCloneChildren, snapAndChildrenInfo.VolumeSnapshotChildren...)
```
nixpanic: **golangci-lint** complains about this. `append()` may modify the 1st parameter (`snapAndChildrenInfo`), which could cause confusion later on in case it is used again ([docs](https://pkg.go.dev/builtin#append)).

Instead, you could do it like this:
```go
children := make([]string, len(snapAndChildrenInfo.TempCloneChildren) + len(snapAndChildrenInfo.VolumeSnapshotChildren))
children = append(children, snapAndChildrenInfo.TempCloneChildren...)
children = append(children, snapAndChildrenInfo.VolumeSnapshotChildren...)
```

nixpanic (dismissed):

Rakshith-R:
> **golangci-lint** complains about this. `append()` may modify the 1st parameter (`snapAndChildrenInfo`), which could cause confusion later on in case it is used again ([docs](https://pkg.go.dev/builtin#append)).
> 
> Instead, you could do it like this:
> 
> ```go
> children := make([]string, len(snapAndChildrenInfo.TempCloneChildren) + len(snapAndChildrenInfo.VolumeSnapshotChildren))
> children = append(children, snapAndChildrenInfo.TempCloneChildren...)
> children = append(children, snapAndChildrenInfo.VolumeSnapshotChildren...)
> ```

It needs to be initialized with zero length :(
ci passes now
ptal

nixpanic (approved):

yati1998 (approved):

Rakshith-R:
@mergifyio queue

Rakshith-R:
Upgrade tests are failing due to 
https://github.com/ceph/ceph-csi/pull/5211

Rakshith-R:
@mergifyio requeue

PR #5253, opened 2025-04-03
e2e: `Skip()` inside a test case causes other cases to be skipped

nixpanic:
The Ginkgo `Skip()` function does not only skip the current `By()`, but
it's parent `It()` function. That means other `By()` tests will not be
run after a `Skip()` was called.

Replacing the `Skip()` with an early return of the function makes sure
that all `By()` tests are executed.


Rakshith-R (approved):

iPraveenParihar (approved):

nixpanic:
@mergifyio queue

nixpanic:
@mergifyio rebase

iPraveenParihar:
@Mergifyio queue

Issue #5282, opened 2025-04-17
Failing to mount cephfs volumes with invalid argument


Starttoaster:
# Describe the bug #

I upgraded ceph-csi in my clusters to 3.14.0 last week, for both rbd and cephfs. I began seeing cephfs volumes failing to mount with an error from nodeplugin stating `invalid argument`.

I'm not sure what the invalid argument actually is, the logs don't quite make that clear. But I noted this old Issue exhibited a similar logline https://github.com/ceph/ceph-csi/issues/2254

And I noted in the 3.14.0 release this PR https://github.com/ceph/ceph-csi/pull/5090 changed the cephfs mount syntax. Not sure if it's related, but seemed suspicious.

# Environment details #

- Image/version of Ceph CSI driver : 3.14.0
- Helm chart version : 3.14.0
- Kernel version : `5.15.0-135-generic` on the k8s nodes
- Mounter used for mounting PVC (for cephFS its `fuse` or `kernel`. for rbd its
  `krbd` or `rbd-nbd`) : we didn't specify it, so admittedly I'm not sure. I'll update this when I find out
- Kubernetes cluster version : 1.32.3
- Ceph cluster version : 18.2.2

# Steps to reproduce #

1. Setup details: 'Deploy the latest release (3.14.0) of ceph-csi for cephfs with mostly default settings', values file below:

```yaml
csiConfig:
  - clusterID: $CEPH_CLUSTER_ID
    monitors: $CEPH_MONITORS
    readAffinity:
      enabled: true
      crushLocationLabels:
        - topology.kubernetes.io/region
        - topology.kubernetes.io/zone

readAffinity:
  enabled: true
  crushLocationLabels:
    - topology.kubernetes.io/region
    - topology.kubernetes.io/zone

nodeplugin:
  httpMetrics:
    containerPort: 8187

storageClass:
  create: true
  name: cephfs-nvme
  clusterID: $CEPH_CLUSTER_ID
  fsName: $CEPHFS_FS
  pool: $CEPHFS_POOL

secret:
  create: true
  adminID: $CEPHFS_USERNAME
  adminKey: $CEPHFS_KEY

kubeletDir: /var/lib/k0s/kubelet
```

2. Create a PVC with ReadWriteMany, and a Pod that mounts it.
3. See error

# Actual results #

ceph-csi failed to mount these volumes, receiving an invalid argument error.

# Expected behavior #

cephfs would mount my volumes.

# Logs #

[...]

PR #5288, opened 2025-04-22
helm: mark userID and userKey required in secret

black-dragon74:

This patch marks `userID` and `userKey` required in helm values.

A release note has been added for the same as well.

Ref:  https://github.com/ceph/ceph-csi/issues/5282

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -2,6 +2,9 @@
 
 ## Breaking Changes
 
+- Support for `secret.adminID` and `secret.adminKey` has been deprecated and removed
+  from Helm charts. Please use `secret.userID` and `secret.userKey` instead.
+
 ## Features
 
 ## NOTE
--- a/charts/ceph-csi-cephfs/templates/secret.yaml
+++ b/charts/ceph-csi-cephfs/templates/secret.yaml
@@ -14,6 +14,6 @@ metadata:
     heritage: {{ .Release.Service }}
     {{- with .Values.commonLabels }}{{ toYaml . | trim | nindent 4 }}{{- end }}
 stringData:
-  userID: {{ .Values.secret.userID }}
-  userKey: {{ .Values.secret.userKey }}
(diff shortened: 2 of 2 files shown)


PendingReleaseNotes.md
```diff
@@ -2,6 +2,9 @@
 
 ## Breaking Changes
 
+- Support for `secret.adminID` and `secret.adminKey` has been deprecated and removed.
```
nixpanic: I think it is only removed from the Helm charts? It is not a breaking change if users did not use Helm to deploy? That could be explained a little better.

charts/ceph-csi-cephfs/templates/secret.yaml
```diff
@@ -14,6 +14,6 @@ metadata:
     heritage: {{ .Release.Service }}
     {{- with .Values.commonLabels }}{{ toYaml . | trim | nindent 4 }}{{- end }}
 stringData:
-  userID: {{ .Values.secret.userID }}
-  userKey: {{ .Values.secret.userKey }}
+  userID: {{ required "A valid userID must be specified in secret" .Values.secret.userID }}
+  userKey: {{ required "A valid userKey must be specified in secret" .Values.secret.userKey }}
```
Madhu-1, line 18: should we add adminID and adminKey back in this release and deprecate it and Remove it in next major release?

thoughts?

black-dragon74: That would make sense actually. This got overlooked in the previous release. Breaking changes should not be a part of minor/patch versions.

Madhu-1 (approved):

nixpanic:
Does this report the new _required_ message, even if `adminID` is set in the `values.yaml` of the current (before attempting to update) deployment?

black-dragon74:
> Does this report the new _required_ message, even if `adminID` is set in the `values.yaml` of the current (before attempting to update) deployment?

Yes.

nixpanic (approved):

nixpanic:
@mergifyio queue

PR #5294, opened 2025-04-28
doc: Update cephfs caps

### LINKED ISSUE #4728: Update the capabilities needed for client.csi-cephfs-node in case of encryption
# Describe the bug #

Currently the standard capabilities set in the [capabilities.md ](https://github.com/ceph/ceph-csi/blob/devel/docs/capabilities.md) is not enough to acquire a lock in rados that is needed when running. This can be seen when the LockExclusive is called rados returns the following error code

`Failed to lock volume ID 0001-0009-rook-ceph-0000000000000001-489ccf33-13a5-40fc-8460-7dd866bc44de: rados: ret=-1, Operation not permitted`

Some discussion and a naive solution how to circumvent this can be found, in this [PR discussion ](https://github.com/ceph/ceph-csi/pull/4697#discussion_r1689605829)

# The Solution #

It needs to be figured out what to add to the OSD caps to give only the necessary permission to do the lock operation. Further this should then be documented in the `capabilities.md `.

flx5:
When cephfs is used with encryption an exclusive lock is acquired. This needs the execute permission on the metadata pool.

Fixes #4728

## Future concerns ##
- It might be necessary to add a e2e test for an encrypted cephfs to avoid issues like #4728 in the future.

**Checklist:**

  guidelines in the [developer
* [x] [Pending release

---

<details>

</details>

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -28,7 +28,7 @@ capabilities:
 
 ```
 mgr "allow rw"
-osd "allow rw tag cephfs metadata=cephfs, allow rw tag cephfs data=cephfs"
+osd "allow rwx tag cephfs metadata=cephfs, allow rw tag cephfs data=cephfs"
 mds "allow r fsname=cephfs path=/volumes, allow rws fsname=cephfs path=/volumes/csi"
 mon "allow r fsname=cephfs"
 ```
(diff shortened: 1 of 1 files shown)


docs/capabilities.md
```diff
@@ -28,7 +28,7 @@ capabilities:
 
 ```
 mgr "allow rw"
-osd "allow rw tag cephfs metadata=cephfs, allow rw tag cephfs data=cephfs"
+osd "allow rwx tag cephfs metadata=cephfs, allow rw tag cephfs data=cephfs"
```
iPraveenParihar, line 31: If am correct, the `allow x` permission is specifically needed when CephFS encryption is in use. If encryption is not configured, this permission can be omitted from the OSD capabilities. Can we add this as NOTE section?

iPraveenParihar (changes requested):

docs/capabilities.md
nixpanic: You are correct. Currently it is only required when CephFS encryption is used. However, this is needed for using RADOS locks, which might get used in other places too. A note about it might be useful, but it can get outdated easily too. Once deployed, users will likely not change the capabilities later on, adding an encrypted CephFS storage class after a while is surely more common.

I am fine with the current enhancement, it will prevent others from hitting problems with this.

nixpanic (approved):

iPraveenParihar (approved):
Thanks!

PR #5302, opened 2025-05-02
Implement ListSnapshots RPC for CephFS and RBD

black-dragon74:

This patch implements `ListSnapshots` RPC which is required by the `external-snapshotter` to properly update status for `VolumeSnapshots/Contents`

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -974,6 +974,62 @@ func (cs *ControllerServer) validateSnapshotReq(ctx context.Context, req *csi.Cr
 	return nil
 }
 
+func (cs *ControllerServer) ListSnapshots(
+	ctx context.Context,
+	req *csi.ListSnapshotsRequest,
+) (*csi.ListSnapshotsResponse, error) {
+	if err := cs.Driver.ValidateControllerServiceRequest(
+		csi.ControllerServiceCapability_RPC_LIST_SNAPSHOTS); err != nil {
+		log.ErrorLog(ctx, "invalid list snapshot req: %v", protosanitizer.StripSecrets(req))
+
+		return nil, err
+	}
+
+	cr, err := util.NewAdminCredentials(req.GetSecrets())
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	defer cr.DeleteCredentials()
+
+	snapshotID := req.GetSnapshotId()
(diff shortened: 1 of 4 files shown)


internal/cephfs/controllerserver.go
```diff
@@ -974,6 +974,62 @@ func (cs *ControllerServer) validateSnapshotReq(ctx context.Context, req *csi.Cr
...
+	cr, err := util.NewAdminCredentials(req.GetSecrets())
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	defer cr.DeleteCredentials()
+
+	snapshotID := req.GetSnapshotId()
+	if snapshotID == "" {
+		return nil, status.Error(codes.InvalidArgument, "snapshot ID cannot be empty")
+	}
```
Nikhil-Ladha, line 997: Technically, we can't error out here because the field is `optional`.
So, if the snapshotID is missing, the csi-driver should be listing all the available snapshots in the backend.
Ref: https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go#L3527

black-dragon74: That is a problem for us as without the snapshot ID we won't really know where to list snapshots from :(

internal/rbd/controllerserver.go
```diff
@@ -356,6 +357,56 @@ func checkValidCreateVolumeRequest(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSna
...
+	}
+
+	mgr := NewManager(cs.Driver.GetInstanceID(), nil, req.GetSecrets())
+	defer mgr.Destroy(ctx)
+
+	snap, err := mgr.GetSnapshotByID(ctx, req.GetSnapshotId())
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "failed to find snapshot with ID %q: %s", req.GetSnapshotId(), err.Error())
+	}
+	defer snap.Destroy(ctx)
```
Nikhil-Ladha, line 378: Same as above

internal/cephfs/controllerserver.go
Nikhil-Ladha: And that's the reason I think, either we shouldn't implement this at all (since it's an optional PRC) or don't error out if the id is not present because it goes against the RPC definition of the field being optional.

PR #5315, opened 2025-05-19
rbd: improve logging for rpc calls

Nikhil-Ladha:
Added logging of reqID for volume group rpc calls. Also, added logs for replication rpc calls
which are helpful during debugging of issues related to failover/relocate.

**Checklist:**

  guidelines in the [developer
* [X] [Pending release
* [X] Documentation has been updated, if necessary.
* [X] Integration tests have been added, if necessary.

---

<details>

</details>

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -306,6 +306,7 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	log.UsefulLog(ctx, "mirror state is %s", info.GetState())
 	if info.GetState() != librbd.MirrorImageEnabled.String() {
 		err = rbdVol.HandleParentImageExistence(ctx, flattenMode)
 		if err != nil {
@@ -371,6 +372,7 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	log.UsefulLog(ctx, "mirror state is %s", info.GetState())
 	switch info.GetState() {
 	// image is already in disabled state
 	case librbd.MirrorImageDisabled.String():
@@ -433,6 +435,7 @@ func (rs *ReplicationServer) PromoteVolume(ctx context.Context,
 
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	log.UsefulLog(ctx, "mirror state is %s, primary: %s", info.GetState(), info.IsPrimary())
 
 	if info.GetState() != librbd.MirrorImageEnabled.String() {
 		return nil, status.Errorf(
@@ -533,6 +536,7 @@ func (rs *ReplicationServer) DemoteVolume(ctx context.Context,
 
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	log.UsefulLog(ctx, "mirror state is %s, primary: %s", info.GetState(), info.IsPrimary())
 
 	if info.GetState() != librbd.MirrorImageEnabled.String() {
 		return nil, status.Errorf(
--- a/internal/csi-common/utils.go
+++ b/internal/csi-common/utils.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/csi-addons/spec/lib/go/replication"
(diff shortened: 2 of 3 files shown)


internal/csi-addons/rbd/replication.go
```diff
@@ -306,6 +306,7 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	log.UsefulLog(ctx, "mirror state is %s", info.GetState())
```
Madhu-1, line 309: log rbd volume as well

Madhu-1: please take care of it in other places as well.

internal/csi-common/utils.go
```diff
@@ -222,18 +223,23 @@ func getReqID(req interface{}) string {
...
-	case *replication.DisableVolumeReplicationRequest:
-		reqID = GetIDFromReplication(r)
-	case *replication.PromoteVolumeRequest:
-		reqID = GetIDFromReplication(r)
-	case *replication.DemoteVolumeRequest:
-		reqID = GetIDFromReplication(r)
-	case *replication.ResyncVolumeRequest:
-		reqID = GetIDFromReplication(r)
-	case *replication.GetVolumeReplicationInfoRequest:
+	case *replication.EnableVolumeReplicationRequest,
```
Madhu-1, line 226: As this is the pattern followed for all other RPC calls, lets keep it as it is unless some linter complains about it.

internal/rbd/mirror.go
```diff
@@ -183,6 +187,8 @@ func (rm *rbdMirror) Promote(ctx context.Context, force bool) error {
 		return fmt.Errorf("failed to promote image %q with error: %w", rm, err)
 	}
 
+	log.DebugLog(ctx, "image %q has been promoted", rm)
```
Madhu-1, line 190: Please make similar changes for group as well in downstream branch

internal/csi-common/utils.go
Nikhil-Ladha: Yes, the change is because of the `gocyclo` complain from the golangci-linter.

internal/rbd/mirror.go
Nikhil-Ladha: It is already there for group, so we are good there.

internal/csi-addons/rbd/replication.go
Nikhil-Ladha: We are already tracking the volumeID/groupID in the RPC call with this change, so I think we can skip logging the rbdVolume here.
Because, then we need to add a conditional logic for volume vs group calls

Madhu-1: >Because, then we need to add a conditional logic for volume vs group calls

Having the names provide mode tails at the log level, we dont need to reverse trace the details from the request ID. We need to work on this one, please take care of it when we add VGR support upstream,

Madhu-1 (approved):

internal/csi-common/utils.go
```diff
@@ -222,18 +223,23 @@ func getReqID(req interface{}) string {
...
+
+	// VolumeGroup
+	case *volumegroup.CreateVolumeGroupRequest:
+		reqID = r.GetName()
+	case *volumegroup.ModifyVolumeGroupMembershipRequest:
+		reqID = r.GetVolumeGroupId()
+	case *volumegroup.DeleteVolumeGroupRequest:
+		reqID = r.GetVolumeGroupId()
+	case *volumegroup.ControllerGetVolumeGroupRequest:
+		reqID = r.GetVolumeGroupId()
```
Rakshith-R, line 242:
```suggestion
	case *volumegroup.ModifyVolumeGroupMembershipRequest,  *volumegroup.DeleteVolumeGroupRequest, *volumegroup.ControllerGetVolumeGroupRequest:
		reqID = r.GetVolumeGroupId()
```

Rakshith-R (changes requested):
Just a small nit, 
everything else looks good to me.

internal/csi-common/utils.go
Nikhil-Ladha: Yeah, this doesn't work since we have different interfaces that is calling the method.
Initially thought of this way, but then had to drop it.

Rakshith-R (approved):
Thanks

Rakshith-R:
@mergifyio queue

Nikhil-Ladha:
@mergifyio requeue

Issue #5346, opened 2025-05-30
Add support for Changed Block Tracking feature


Rakshith-R:
# Describe the feature you'd like to have #

Implement [Snapshot Metadata Service RPCs](https://github.com/container-storage-interface/spec/blob/master/spec.md#snapshot-metadata-service-rpcs) to support
changed block tracking feature 

kep: https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/3314-csi-changed-block-tracking

# What is the value to the end user? (why is it a priority?) #

Efficient backup can be done by only backing up allocated blocks in a snapshot and changed blocks between two snapshots.

# How will we know we have a good solution? (acceptance criteria) #

- [x] Implement  [Snapshot Metadata Service RPCs](https://github.com/container-storage-interface/spec/blob/master/spec.md#snapshot-metadata-service-rpcs) https://github.com/ceph/ceph-csi/pull/5347
- [x] Add middleware for logging for the new streaming server https://github.com/ceph/ceph-csi/pull/5411
- [x] Add new sidecar deployment capability in CephCSIOperator (Depends on https://github.com/ceph/ceph-csi-operator/pull/274)
- [ ] Enable changed block tracking tests in external k8s suite  (depends-on https://github.com/kubernetes/kubernetes/pull/130918)

PR #5347, opened 2025-05-30
 rbd: add support for changed block tracking RPCs

Rakshith-R:
rbd: add support for changed block tracking RPCs

This commit adds support for GetMetadataAllocated
and GetMetadataDelta RPCs.
[CSI Snapshot Metadata Service RPCs](https://github.com/container-storage-interface/spec/blob/master/spec.md#snapshot-metadata-service-rpcs)

Updates: #5346 

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -5,6 +5,7 @@
 ## Features
 
 - helm: Support VolumeSnapshotClass and VolumeGroupSnapshotClass
+- rbd: add support for [CSI Snapshot Metadata Service RPCs](https://github.com/container-storage-interface/spec/blob/master/spec.md#snapshot-metadata-service-rpcs)
 
 ## NOTE
 
--- a/docs/rbd/deploy.md
+++ b/docs/rbd/deploy.md
@@ -550,3 +550,36 @@ kubectl edit cm ceph-config
 
 This will update the `ceph.conf` of the underlying ceph cluster to enable
 librbd logs in `csi-rbdplugin` container.
+
+## Changed Block Tracking (CBT)
+
+> **Warning**: Requires Ceph version that supports RBD snap diff by ID feature.
+For details, see ceph tracker:
+["diff-iterate by snap ID"](https://tracker.ceph.com/issues/65720).
+
+Ceph-CSI implements Changed Block Tracking (CBT) for RBD volumes using
+the SnapshotMetadataService (SMS) APIs. This feature enables efficient
+and reliable differential backup of data stored in CSI volumes.
+
+The feature is exposed through the CSI Controller Server's
+SnapshotMetadataService APIs with two primary operations:
+
+* `GetMetadataAllocated`:
+
+   * Streams metadata about allocated blocks in a snapshot
+   * Useful for full backup operations
+   * Returns block ranges that contain actual data
+
+* `GetMetadataDelta`:
+
+   * Streams metadata about block differences between two snapshots
+   * Optimal for incremental backups
+   * Returns only blocks that changed between snapshots
+
+Additional Resources:
+
+* CSI Differential Snapshot for Block Volumes KEP:
+  [kep-3314](https://github.com/kubernetes/enhancements/issues/3314)
+
+* External Snapshot Metadata sidecar project:
+  [repository](https://github.com/kubernetes-csi/external-snapshot-metadata)
--- a/internal/rbd/errors/errors.go
+++ b/internal/rbd/errors/errors.go
@@ -68,3 +68,11 @@ var (
 	// ErrGroupNotFound is returned when group is not found in the cluster on the given pool and/or namespace.
 	ErrGroupNotFound = fmt.Errorf("%w: RBD group not found", librbd.ErrNotFound)
 )
+
+// ErrorCode is an interface that defines a method to return an error code.
+// This interface is used to provide a consistent way to retrieve error codes
+// from functions that is expected to return an error with specific codes.
+type ErrorCode interface {
+	// ErrorCode returns the error code.
+	ErrorCode() int
+}
--- a/internal/rbd/features/features.go
+++ b/internal/rbd/features/features.go
@@ -31,6 +31,10 @@ var (
 	groupGetSnapInfoOnce      sync.Once
 	errGroupGetSnapInfo       error
 	groupGetSnapInfoSupported = false
+
+	rbdSnapDiffByIDOnce      sync.Once
+	errRBDSnapDiffByID       error
+	rbdSnapDiffByIDSupported = false
 )
 
 // SupportsGroupSnapGetInfo detects if librbd has the rbd_group_snap_get_info
@@ -55,4 +59,28 @@ func SupportsGroupSnapGetInfo() (bool, error) {
 	})
 
 	return groupGetSnapInfoSupported, errGroupGetSnapInfo
+}
+
+// SupportsRBDSnapDiffByID detects if librbd has the rbd_diff_iterate3
+// function.
+func SupportsRBDSnapDiffByID() (bool, error) {
+	rbdSnapDiffByIDOnce.Do(func() {
+		// make sure librbd.so.x is loaded, might not (yet) be the case
+		// if no rbd functions are called
+		var opts C.rbd_image_options_t
+		//nolint:gocritic // ignore result of rbd_image_options functions
+		C.rbd_image_options_create(&opts)
+		C.rbd_image_options_destroy(opts)
+
+		// check for rbd_diff_iterate3() in loaded libs/symbols
+		errRBDSnapDiffByID = dlsym("rbd_diff_iterate3")
+
+		if errRBDSnapDiffByID == nil {
+			rbdSnapDiffByIDSupported = true
+		} else if strings.Contains(errRBDSnapDiffByID.Error(), "undefined symbol") {
+			errRBDSnapDiffByID = nil
+		}
+	})
+
+	return rbdSnapDiffByIDSupported, errRBDSnapDiffByID
 }
--- a/internal/rbd/identityserver.go
+++ b/internal/rbd/identityserver.go
@@ -21,6 +21,7 @@ import (
 
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
 	"github.com/ceph/ceph-csi/internal/rbd/features"
+	"github.com/ceph/ceph-csi/internal/util/log"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 )
@@ -62,7 +63,9 @@ func (is *IdentityServer) GetPluginCapabilities(
 
 	// GroupSnapGetInfo is used within the VolumeGroupSnapshot implementation
 	vgsSupported, err := features.SupportsGroupSnapGetInfo()
-	if err == nil && vgsSupported {
+	if err != nil {
+		log.WarningLog(ctx, "error checking for group snapshot support: %s", err)
+	} else if vgsSupported {
 		gcs := csi.PluginCapability{
 			Type: &csi.PluginCapability_Service_{
 				Service: &csi.PluginCapability_Service{
@@ -71,6 +74,21 @@ func (is *IdentityServer) GetPluginCapabilities(
 			},
 		}
 		caps = append(caps, &gcs)
+	}
+
+	// RBDSnapDiffByID is used within the SnapshotMetadata service
+	rbdSnapDiffByIDSupported, err := features.SupportsRBDSnapDiffByID()
+	if err != nil {
+		log.WarningLog(ctx, "error checking for snapshot diff by ID support: %s", err)
+	} else if rbdSnapDiffByIDSupported {
+		snapDiffCaps := csi.PluginCapability{
+			Type: &csi.PluginCapability_Service_{
+				Service: &csi.PluginCapability_Service{
+					Type: csi.PluginCapability_Service_SNAPSHOT_METADATA_SERVICE,
+				},
+			},
+		}
+		caps = append(caps, &snapDiffCaps)
 	}
 
 	return &csi.GetPluginCapabilitiesResponse{
--- a/internal/rbd/sms_controllerserver.go
+++ b/internal/rbd/sms_controllerserver.go
@@ -1,0 +1,253 @@
+/*
+Copyright 2025 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rbd
+
+import (
+	"context"
+	"errors"
+
+	rbderrors "github.com/ceph/ceph-csi/internal/rbd/errors"
+	"github.com/ceph/ceph-csi/internal/rbd/features"
+	"github.com/ceph/ceph-csi/internal/util/log"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	// defaultMaxResults is the internal limit for max results in metadata streaming
+	// This is set to a reasonable value to prevent excessive memory usage
+	// and ensure efficient processing of metadata requests.
+	// TODO: Consider making this configurable via a cmdline flag.
+	defaultMaxResults int32 = 128
+)
+
+// normalizeMaxResults normalizes the maxResults parameter. It handles zero values,
+// applies internal limits, and provides logging for debugging.
+func normalizeMaxResults(ctx context.Context, requestedMaxResults int32) int32 {
+	// Handle zero case - user wants default behavior
+	if requestedMaxResults == 0 {
+		log.DebugLog(ctx, "maxResults was 0, using internal default: %d", defaultMaxResults)
+
+		return defaultMaxResults
+	}
+
+	// Apply internal max limit for memory management
+	if requestedMaxResults > defaultMaxResults {
+		log.DebugLog(ctx, "maxResults limited from %d to %d (internal max limit)",
+			requestedMaxResults, defaultMaxResults)
+
+		return defaultMaxResults
+	}
+	log.DebugLog(ctx, "using requested maxResults: %d", requestedMaxResults)
+
+	return requestedMaxResults
+}
+
+func validateMetadataAllocatedReq(req *csi.GetMetadataAllocatedRequest) error {
+	if req.GetSnapshotId() == "" {
+		return status.Error(codes.InvalidArgument, "snapshot ID cannot be empty")
+	}
+
+	if req.GetSecrets() == nil || len(req.GetSecrets()) == 0 {
+		return status.Error(codes.InvalidArgument, "secrets cannot be nil or empty")
+	}
+
+	if req.GetMaxResults() < 0 {
+		return status.Error(codes.InvalidArgument, "max results cannot be negative")
+	}
+
+	if req.GetStartingOffset() < 0 {
+		return status.Error(codes.OutOfRange, "starting offset cannot be negative")
+	}
+
+	return nil
+}
+
+// GetMetadataAllocated streams the allocated block metadata for a snapshot.
+func (cs *ControllerServer) GetMetadataAllocated(
+	req *csi.GetMetadataAllocatedRequest,
+	stream csi.SnapshotMetadata_GetMetadataAllocatedServer,
+) error {
+	ctx := stream.Context()
+	rbdSnapDiffByIDSupported, err := features.SupportsRBDSnapDiffByID()
+	if err != nil {
+		log.ErrorLog(ctx, "error checking for snapshot diff by ID support: %s", err)
+
+		return status.Error(codes.Internal, err.Error())
+	} else if !rbdSnapDiffByIDSupported {
+		log.ErrorLog(ctx, "snapshot diff by ID feature is not supported")
+
+		return status.Error(codes.Unimplemented, "snapshot diff by ID feature is not supported")
+	}
+
+	err = validateMetadataAllocatedReq(req)
+	if err != nil {
+		return err
+	}
+
+	snapshotID := req.GetSnapshotId()
+
+	mgr := NewManager(cs.Driver.GetInstanceID(), nil, req.GetSecrets())
+	defer mgr.Destroy(ctx)
+
+	rbdSnap, err := mgr.GetSnapshotByID(ctx, snapshotID)
+	if err != nil {
+		if errors.Is(err, rbderrors.ErrImageNotFound) {
+			log.ErrorLog(ctx, "snapshot %q not found: %v", snapshotID, err)
+
+			return status.Errorf(codes.NotFound, "snapshot %q not found: %v", snapshotID, err)
+		}
+
+		return status.Error(codes.Internal, err.Error())
+	}
+	defer rbdSnap.Destroy(ctx)
+
+	volSize := rbdSnap.GetSize()
+
+	startingOffset := req.GetStartingOffset()
+	if startingOffset >= volSize {
+		return status.Errorf(codes.OutOfRange, "starting offset %d is out of range for volume size %d",
+			startingOffset, volSize)
+	}
+
+	maxResults := normalizeMaxResults(ctx, req.GetMaxResults())
+
+	sendResponse := func(changedBlocks []*csi.BlockMetadata) error {
+		return stream.Send(&csi.GetMetadataAllocatedResponse{
+			BlockMetadataType:   csi.BlockMetadataType_VARIABLE_LENGTH,
+			VolumeCapacityBytes: volSize,
+			BlockMetadata:       changedBlocks,
+		})
+	}
+
+	err = rbdSnap.ProcessMetadata(ctx, startingOffset, maxResults, nil, sendResponse)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to stream metadata allocated: %v", err)
+
+		return status.Error(codes.Internal, err.Error())
+	}
+	log.DebugLog(ctx, "successfully streamed metadata allocated")
+
+	return nil
+}
+
+func validateMetadataDeltaReq(req *csi.GetMetadataDeltaRequest) error {
+	if req.GetBaseSnapshotId() == "" {
+		return status.Error(codes.InvalidArgument, "base snapshot ID cannot be empty")
+	}
+
+	if req.GetTargetSnapshotId() == "" {
+		return status.Error(codes.InvalidArgument, "target snapshot ID cannot be empty")
+	}
+
+	if req.GetSecrets() == nil || len(req.GetSecrets()) == 0 {
+		return status.Error(codes.InvalidArgument, "secrets cannot be nil or empty")
+	}
+
+	if req.GetMaxResults() < 0 {
+		return status.Error(codes.InvalidArgument, "max results cannot be negative")
+	}
+
+	if req.GetStartingOffset() < 0 {
+		return status.Error(codes.OutOfRange, "starting offset cannot be negative")
+	}
+
+	return nil
+}
+
+// GetMetadataDelta streams the block metadata delta between two snapshots.
+func (cs *ControllerServer) GetMetadataDelta(
+	req *csi.GetMetadataDeltaRequest,
+	stream csi.SnapshotMetadata_GetMetadataDeltaServer,
+) error {
+	ctx := stream.Context()
+	rbdSnapDiffByIDSupported, err := features.SupportsRBDSnapDiffByID()
+	if err != nil {
+		log.ErrorLog(ctx, "failed to check for snapshot diff by ID support: %s", err)
+
+		return status.Error(codes.Internal, err.Error())
+	} else if !rbdSnapDiffByIDSupported {
+		log.ErrorLog(ctx, "snapshot diff by ID feature is not supported")
+
+		return status.Error(codes.Unimplemented, "snapshot diff by ID feature is not supported")
+	}
+
+	err = validateMetadataDeltaReq(req)
+	if err != nil {
+		return err
+	}
+
+	baseSnapshotID := req.GetBaseSnapshotId()
+	targetSnapshotID := req.GetTargetSnapshotId()
+
+	mgr := NewManager(cs.Driver.GetInstanceID(), nil, req.GetSecrets())
+	defer mgr.Destroy(ctx)
+
+	baseRBDSnap, err := mgr.GetSnapshotByID(ctx, baseSnapshotID)
+	if err != nil {
+		if errors.Is(err, rbderrors.ErrImageNotFound) {
+			log.ErrorLog(ctx, "snapshot %q not found: %v", baseSnapshotID, err)
+
+			return status.Errorf(codes.NotFound, "snapshot %q not found: %v", baseSnapshotID, err)
+		}
+
+		return status.Error(codes.Internal, err.Error())
+	}
+	defer baseRBDSnap.Destroy(ctx)
+
+	targetRBDSnap, err := mgr.GetSnapshotByID(ctx, targetSnapshotID)
+	if err != nil {
+		if errors.Is(err, rbderrors.ErrImageNotFound) {
+			log.ErrorLog(ctx, "snapshot %q not found: %v", targetSnapshotID, err)
+
+			return status.Errorf(codes.NotFound, "snapshot %q not found: %v", targetSnapshotID, err)
+		}
+
+		return status.Error(codes.Internal, err.Error())
+	}
+	defer targetRBDSnap.Destroy(ctx)
+
+	volSize := targetRBDSnap.GetSize()
+
+	startingOffset := req.GetStartingOffset()
+	if startingOffset >= volSize {
+		return status.Errorf(codes.OutOfRange, "starting offset %d is out of range for volume size %d",
+			startingOffset, volSize)
+	}
+
+	maxResults := normalizeMaxResults(ctx, req.GetMaxResults())
+
+	sendResponse := func(changedBlocks []*csi.BlockMetadata) error {
+		return stream.Send(&csi.GetMetadataDeltaResponse{
+			BlockMetadataType:   csi.BlockMetadataType_VARIABLE_LENGTH,
+			VolumeCapacityBytes: volSize,
+			BlockMetadata:       changedBlocks,
+		})
+	}
+
+	err = targetRBDSnap.ProcessMetadata(ctx, startingOffset, maxResults, baseRBDSnap, sendResponse)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to stream metadata delta: %v", err)
+
+		return status.Error(codes.Internal, err.Error())
+	}
+	log.DebugLog(ctx, "successfully streamed metadata delta")
+
+	return nil
+}
--- a/internal/rbd/snap_diff.go
+++ b/internal/rbd/snap_diff.go
@@ -1,0 +1,201 @@
+/*
+Copyright 2025 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rbd
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	rbderrors "github.com/ceph/ceph-csi/internal/rbd/errors"
+	"github.com/ceph/ceph-csi/internal/rbd/types"
+	"github.com/ceph/ceph-csi/internal/util/log"
+
+	librbd "github.com/ceph/go-ceph/rbd"
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc/codes"
+)
+
+// ProcessMetadata streams the block metadata for a snapshot.
+// If baseRBDSnap is provided, it calculates the delta between the
+// current snapshot and the base snapshot.
+func (rbdSnap *rbdSnapshot) ProcessMetadata(
+	ctx context.Context,
+	startingOffset int64,
+	maxResults int32,
+	baseSnap types.Snapshot,
+	sendResponse types.MetadataCallback,
+) error {
+	var fromSnapID uint64
+
+	image, err := rbdSnap.open()
+	if err != nil {
+		return fmt.Errorf("failed to open image %q: %w", rbdSnap, err)
+	}
+	defer func() {
+		cErr := image.Close()
+		if cErr != nil {
+			log.WarningLog(ctx, "resource leak, failed to close image: %v", cErr)
+		}
+	}()
+
+	snapID, err := rbdSnap.getRBDSnapID(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get snapshot ID of %q: %w", rbdSnap, err)
+	}
+
+	err = image.SetSnapByID(snapID)
+	if err != nil {
+		return fmt.Errorf("failed to set snapshot %q by ID %d: %w",
+			rbdSnap, snapID, err)
+	}
+
+	if baseSnap != nil {
+		baseRBDSnap, err := rbdSnapFromSnapshot(baseSnap)
+		if err != nil {
+			return fmt.Errorf("failed to convert base snapshot to rbdSnapshot: %w", err)
+		}
+		fromSnapID, err = baseRBDSnap.getRBDSnapID(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get ID of base snapshot ID: %w", err)
+		}
+	}
+
+	luksHeaderPadding, err := rbdSnap.getLuksHeaderSizeMetadata()
+	if err != nil {
+		return fmt.Errorf("failed to get LUKS header size from metadata: %w", err)
+	}
+	// We need to adjust the starting offset to account for the LUKS header padding.
+	startingOffset += int64(luksHeaderPadding)
+
+	// If the starting offset is greater than the volume size after adding padding,
+	// we can return early since there are no changed blocks to report.
+	if startingOffset > rbdSnap.VolSize {
+		return nil
+	}
+
+	changedBlocks := make([]*csi.BlockMetadata, 0)
+	cb := createDiffIterateByIDCB(
+		ctx,
+		&changedBlocks,
+		maxResults,
+		luksHeaderPadding,
+		sendResponse,
+	)
+	diffIterateByIDConfig := librbd.DiffIterateByIDConfig{
+		Offset:   uint64(startingOffset),
+		Length:   uint64(rbdSnap.VolSize),
+		Callback: cb,
+	}
+	if baseSnap != nil {
+		// If a base snapshot is provided, set the FromSnapID to
+		// the base snapshot ID to get the delta between the two snapshots.
+		diffIterateByIDConfig.FromSnapID = fromSnapID
+	}
+
+	diffIterateErr := image.DiffIterateByID(diffIterateByIDConfig)
+	err = handleDiffIterateError(diffIterateErr)
+	if err != nil {
+		return fmt.Errorf("failed to get diff: %w", err)
+	}
+
+	if len(changedBlocks) != 0 {
+		// Send any remaining changed blocks after the loop.
+		err = sendResponse(changedBlocks)
+		if err != nil {
+			return fmt.Errorf("failed to send response: %w", err)
+		}
+	}
+	// Successfully completed the diff operation.
+	return nil
+}
+
+// handleDiffIterateError checks the error returned by DiffIterateByID.
+// It handles specific error codes and returns a formatted error message
+// for unrecognized error codes.
+func handleDiffIterateError(err error) error {
+	if err == nil {
+		// No error, return nil.
+		return nil
+	}
+	// Check if the error implements ErrorCode to handle specific error codes.
+	var errCode rbderrors.ErrorCode
+	ok := errors.As(err, &errCode)
+	if !ok {
+		// The error does not implement ErrorCode, return a generic internal error.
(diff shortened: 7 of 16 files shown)


internal/rbd/controllerserver.go
```diff
@@ -1709,3 +1709,67 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
 
 	return &csi.ControllerUnpublishVolumeResponse{}, nil
 }
+
+// GetSnapshotMetadata streams the allocated block metadata for a snapshot.
```
Nikhil-Ladha:
```suggestion
// GetMetadataAllocated streams the allocated block metadata for a snapshot.
```

internal/rbd/controllerserver.go
```diff
@@ -1709,3 +1709,67 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
...
+
+	return rbdSnap.getMetadataAllocated(ctx, stream, req.GetStartingOffset(), req.GetMaxResults())
+}
+
+// GetMetadataDelta streams the block metadata delta between two snapshots.
+func (cs *ControllerServer) GetMetadataDelta(
+	req *csi.GetMetadataDeltaRequest,
+	stream csi.SnapshotMetadata_GetMetadataDeltaServer,
+) error {
+	ctx := context.WithValue(stream.Context(), "Base-Snapshot-ID", req.GetBaseSnapshotId())
```
Nikhil-Ladha: `Base Snapshot ID`

internal/rbd/controllerserver.go
```diff
@@ -1709,3 +1709,67 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
...
+	defer cr.DeleteCredentials()
+
+	baseSnapshotID := req.GetBaseSnapshotId()
+	if baseSnapshotID == "" {
+		return status.Error(codes.InvalidArgument, "base snapshot ID cannot be empty")
+	}
+
+	targetSnapshotID := req.GetTargetSnapshotId()
+	if targetSnapshotID == "" {
+		return status.Error(codes.InvalidArgument, "base snapshot ID cannot be empty")
```
Nikhil-Ladha:
```suggestion
		return status.Error(codes.InvalidArgument, "target snapshot ID cannot be empty")
```

Nikhil-Ladha (changes requested):

internal/rbd/controllerserver.go
```diff
@@ -1709,3 +1709,67 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
 
 	return &csi.ControllerUnpublishVolumeResponse{}, nil
 }
+
```
Madhu-1: Lets move this to a new file as its a new functionality, This fine is already having 1770 lines of code.

internal/rbd/controllerserver.go
```diff
@@ -1709,3 +1709,67 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
...
+	defer cr.DeleteCredentials()
+
+	snapshotID := req.GetSnapshotId()
+	if snapshotID == "" {
+		return status.Error(codes.InvalidArgument, "snapshot ID cannot be empty")
+	}
+
+	rbdSnap, err := genSnapFromSnapID(ctx, snapshotID, cr, req.GetSecrets())
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
```
Madhu-1: what about the NotFound error?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+)
+
+func (rbdSnap *rbdSnapshot) getMetadataAllocated(
+	ctx context.Context,
+	stream csi.SnapshotMetadata_GetMetadataAllocatedServer,
+	startingOffset int64,
+	maxResults int32) error {
+	image, err := rbdSnap.open()
+	if err != nil {
+		return err
```
Madhu-1: return GRPC error message

Madhu-1: Take care of this in other places in this function

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	err = image.SetSnapByID(snapID)
+	if err != nil {
+		return fmt.Errorf("failed to set snapshot %q by ID %d: %w", rbdSnap, snapID, err)
+	}
+
+	changedBlocks := make([]*csi.BlockMetadata, 0)
+	diffIterateByIDConfig := librbd.DiffIterateByIDConfig{
+		Offset: uint64(startingOffset),
```
Madhu-1: No validation is required for startingOffSet, what is the error message returned for invalid offSet?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+		Length: uint64(rbdSnap.VolSize),
+		Callback: func(offset, sizeBytes uint64, _ int, _ interface{}) int {
+			select {
+			case <-stream.Context().Done():
+				return streamDoneErrCode
+			default:
+				changedBlocks = append(
```
Madhu-1: Dont we get into the memory problem, If we have larger changedBlocks as we keep on appending to it?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+				VolumeCapacityBytes: rbdSnap.VolSize,
+				BlockMetadata:       changedBlocks,
+			})
+			if err != nil {
+				log.ErrorLog(ctx, "failed to send response: %v", err)
+
+				return sendResponseErrCode
+			}
+			// Reset the changedBlocks slice for the next batch
+			changedBlocks = make([]*csi.BlockMetadata, 0)
```
Madhu-1: Can we have a fixed number of memory allocated per batch so that we dont get into OOM problem?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+*/
+
+package rbd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ceph/ceph-csi/internal/util/log"
+	librbd "github.com/ceph/go-ceph/rbd"
```
Madhu-1, line 28: a new line is missing

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+		case streamDoneErrCode:
+			// The stream was closed by the client.
+			// This is not an error, just a normal termination of the stream.
+			// Log the closure for debugging purposes.
+			log.DebugLog(ctx, "The stream was closed")
+
+			return nil
+		case sendResponseErrCode:
+			// An error occurred while sending the response.
+			log.ErrorLog(ctx, "failed to send response: %w", err)
```
Madhu-1: log as %s, take care of it in other places as well

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	err = image.DiffIterateByID(diffIterateByIDConfig)
+	if err != nil {
+		errno, ok := err.(interface{ ErrorCode() int })
+		if !ok {
+			// The error does not implement ErrorCode, return a generic internal error.
+			log.ErrorLog(ctx, "failed to get diff: %w", err)
+
+			return status.Error(codes.Internal, fmt.Sprintf("failed to get diff: $w", err))
```
Madhu-1: cant we use status.Errorf here?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+		case sendResponseErrCode:
+			// An error occurred while sending the response.
+			log.ErrorLog(ctx, "failed to send response: %w", err)
+
+			return status.Error(codes.Internal, "failed to send response")
+		default:
+			// An unexpected error occurred.
+			log.ErrorLog(ctx, "failed to get diff: %w, error code: %d", err, errno)
+
+			return status.Error(codes.Internal, fmt.Sprintf("failed to get diff: %w, error code: %d", err, errno))
```
Madhu-1: use status.Errorf

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	}
+	// If no error, we successfully completed the diff operation.
+	if len(changedBlocks) > 0 {
```
Madhu-1: Can we have a function to send the response and can be reused here and above like

```suggestion
send := func() error {
		if len(changedBlocks) == 0 {
			return nil
		}
		log.DebugLog(ctx, "Sending %d changed blocks to the stream", len(changedBlocks))
		if err := stream.Send(&csi.GetMetadataAllocatedResponse{
			BlockMetadataType:   csi.BlockMetadataType_VARIABLE_LENGTH,
			VolumeCapacityBytes: rbdSnap.VolSize,
			BlockMetadata:       changedBlocks,
		}); err != nil {
			log.ErrorLog(ctx, "failed to send response: %v", err)
			return status.Error(codes.Internal, fmt.Sprintf("failed to send response: %v", err))
		}
		changedBlocks = changedBlocks[:0] // Reset slice
		return nil
	}
```

internal/rbd/controllerserver.go
```diff
@@ -1709,3 +1709,67 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
...
+	}
+
+	targetSnapshotID := req.GetTargetSnapshotId()
+	if targetSnapshotID == "" {
+		return status.Error(codes.InvalidArgument, "base snapshot ID cannot be empty")
+	}
+
+	baseRBDSnap, err := genSnapFromSnapID(ctx, baseSnapshotID, cr, req.GetSecrets())
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
```
Madhu-1: Need to handle NotFound?

internal/rbd/controllerserver.go
```diff
@@ -1709,3 +1709,67 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
...
+	}
+	defer baseRBDSnap.Destroy(ctx)
+
+	targetRBDSnap, err := genSnapFromSnapID(ctx, targetSnapshotID, cr, req.GetSecrets())
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
```
Madhu-1: Need to handle NotFound?

internal/rbd/identityserver.go
```diff
@@ -73,6 +73,19 @@ func (is *IdentityServer) GetPluginCapabilities(
 		caps = append(caps, &gcs)
 	}
 
+	// RBDSnapDiffByID is used within the SnapshotMetadata service
+	rbdSnapDiffByIDSupported, err := features.SupportsRBDSnapDiffByID()
+	if err == nil && rbdSnapDiffByIDSupported {
```
Madhu-1: This is same as `SupportsGroupSnapGetInfo` (error is not logged) can we log the error for debug purpose?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,263 @@
...
+			return status.Error(codes.Internal, err.Error())
+		}
+	}
+	log.DebugLog(ctx, "successfully completed GetMetadataAllocated operation")
+
+	// Successfully completed the diff operation.
+	return nil
+}
+
+func (rbdSnap *rbdSnapshot) getMetadataDelta(
```
Madhu-1: This looks similar to above and have lot of duplication . Can we rewrite this to avoid duplication of the core logic? That will help us to keep the outer implementation simple

Madhu-1 (changes requested):

Rakshith-R:
Thanks for the quick detailed review

The new main modifications are:
- moved the new calls to snap_diff.go file
- debug logs for feature support check errors
- refactored to have a single flow with sendStreamResponse handling which response to send allocated or delta
- modified return statement and ensured they are formatted and have correct error codes
- added validation for req parameters
- added check for snapshot's existence + returns correct error code now
- added additional checks for startingOffset (add out of range error code as specified in the [spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#getmetadataallocated-errors))
- additional checks for maxResults and internal value to limit memory usage ([spec](spec allows us to send lower number of blocks as well))

failing golang lintci is being handled in https://github.com/ceph/ceph-csi/pull/5351

Please take a look again

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,344 @@
...
+
+	ctx := stream.Context()
+	cr, err := util.NewUserCredentials(req.GetSecrets())
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+	defer cr.DeleteCredentials()
+
+	snapshotID := req.GetSnapshotId()
+	if snapshotID == "" {
```
iPraveenParihar: not needed? already handled in `validateMetadataAllocatedReq`

internal/rbd/features/features.go
```diff
@@ -29,8 +29,11 @@ import (
 
 var (
 	groupGetSnapInfoOnce      sync.Once
+	rbdSnapDiffByIDOnce       sync.Once
 	errGroupGetSnapInfo       error
+	errRBDSnapDiffByID        error
 	groupGetSnapInfoSupported = false
+	rbdSnapDiffByIDSupported  = false
```
nixpanic: nit: I prefer to keep the related variables grouped together, like:
```go
	groupGetSnapInfoOnce      sync.Once
	errGroupGetSnapInfo       error
	groupGetSnapInfoSupported = false

	rbdSnapDiffByIDOnce       sync.Once
	errRBDSnapDiffByID        error
	rbdSnapDiffByIDSupported  = false
```

internal/rbd/features/features.go
```diff
@@ -56,3 +59,25 @@ func SupportsGroupSnapGetInfo() (bool, error) {
 
 	return groupGetSnapInfoSupported, errGroupGetSnapInfo
 }
+
+func SupportsRBDSnapDiffByID() (bool, error) {
```
nixpanic, line 66: every public/exported function should have a comment above it describing what it does.

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,344 @@
...
+const (
+	// TODO: Make internalMaxResults configurable via a flag or config option.
+	// internalMaxResults is the maximum number of results to return in a single
+	// response. This is used to limit the number of changed blocks that needs
+	// to be stored in memory before sending a response to the stream.
+	internalMaxResults = 20
+
+	// streamDoneErrCode is a custom error code used to indicate that the
+	// stream was closed by the client.
+	streamDoneErrCode = -1
```
nixpanic: Is there a reason to not use `google.golang.org/grpc/codes` constants? Or, maybe base it on return values of `read()`, like `EOF` and so on?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	if baseRBDSnap != nil {
+		// If a base snapshot is provided, set the FromSnapID to
+		// the base snapshot ID to get the delta between the two snapshots.
+		diffIterateByIDConfig.FromSnapID = fromSnapID
+	}
+
+	err = image.DiffIterateByID(diffIterateByIDConfig)
+	if err != nil {
+		errno, ok := err.(interface{ ErrorCode() int })
```
nixpanic: Isn't there a go-ceph error type that can be used for typecasting instead of an anonymous interface? Or, at least move the type-casting out into it's own function with a descriptive comment.

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,344 @@
...
+
+	if req.GetStartingOffset() < 0 {
+		return status.Error(codes.OutOfRange, "starting offset cannot be negative")
+	}
+
+	return nil
+}
+
+// GetMetadataAllocated streams the allocated block metadata for a snapshot.
+func (cs *ControllerServer) GetMetadataAllocated(
```
nixpanic: Please don't mix functions for rbdImage and ControllerServer in the same file. Move the ControllerServer SMS functions (and helpers) into a separate file.

Eventually, after much more refactoring, the ControllerServer should only use the (internal) rbdManager API and opaque Volume type.

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,344 @@
...
+		return status.Error(codes.Internal, err.Error())
+	}
+	defer cr.DeleteCredentials()
+
+	snapshotID := req.GetSnapshotId()
+	if snapshotID == "" {
+		return status.Error(codes.InvalidArgument, "snapshot ID cannot be empty")
+	}
+
+	rbdSnap, err := genSnapFromSnapID(ctx, snapshotID, cr, req.GetSecrets())
```
nixpanic: Use the `rbdManager` to get this, it should be(come) the only entry point to get Volumes and Snapshots. You will need to extend the Snapshot interface/type for the diff/metadata in a clean way too.

internal/rbd/snapshot.go
```diff
@@ -400,3 +400,31 @@ func (rbdSnap *rbdSnapshot) SetVolumeGroup(ctx context.Context, cr *util.Credent
...
+// getRBDSnapID returns the associated rbd snapshot ID.
+func (rbdSnap *rbdSnapshot) getRBDSnapID(ctx context.Context) (uint64, error) {
+	vol := rbdSnap.toVolume()
+	err := vol.Connect(rbdSnap.conn.Creds)
+	if err != nil {
+		return 0, err
+	}
+	defer vol.Destroy(ctx)
+
+	image, err := vol.open()
```
nixpanic, line 430: Is there really way to get the `ParentInfo` through `rbdSnap`? Re-using `conn.Creds` is dangerous, as it may get free'd on the `vol.Destroy()`.

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,344 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	// TODO: Make internalMaxResults configurable via a flag or config option.
+	// internalMaxResults is the maximum number of results to return in a single
+	// response. This is used to limit the number of changed blocks that needs
+	// to be stored in memory before sending a response to the stream.
+	internalMaxResults = 20
```
nixpanic: Isn't 20 very small?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,344 @@
...
+	ctx context.Context,
+	stream metadataStream,
+	startingOffset int64,
+	maxResults int32,
+	baseRBDSnap *rbdSnapshot,
+) error {
+	var fromSnapID uint64
+	image, err := rbdSnap.open()
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to open image %q: %s", rbdSnap, err)
```
nixpanic: `rbdSnap` functions should not return gRPC errors. gRPC errors should only be created in `ControllerServer` or other gRPC servers.

nixpanic (commented):
1st pass going through this. Needs cleanup and clear splitting of internal RBD things and ControllerServer Snapshot usage. Extend the `types.Snapshot` interface with required functions, so that ControllerServer does not need to use `rbdSnap` types directly.

internal/rbd/snapshot.go
Rakshith-R: > Is there really way to get the `ParentInfo` through `rbdSnap`? Re-using `conn.Creds` is dangerous, as it may get free'd on the `vol.Destroy()`.

The image underneath rbdSnap object is actually the parent PVC, so we have to convert it into rbdVol to access the child rbd image. It's done at a lot of places currently too.

Switched to using `conn.Copy()` so we should be good.
```
// Copy adds an extra reference count to the used ConnEntry and returns the
// *rados.Conn if it was found.
```
(Will be pushed in next pr update)

internal/rbd/snap_diff.go
Rakshith-R: > Isn't 20 very small?

I'm fine with any value here on the lower
The CSI spec gives us freedom to send lower number than asked for.

We'll have a followup to make this configurable.

Rakshith-R: Any suggestions ?

nixpanic: I expect that the size of a single metadata for a changed block is relatively small. Start with something like 128, so that the number of repeated calls is reduced (will be less efficient?).

A configuration option is definitely good to do at a later point in time. Does the sidecar provide metrics about the number of changed blocks returned per diff? With that, a more educated guess for a default can be made.

PendingReleaseNotes.md
```diff
@@ -4,6 +4,8 @@
 
 ## Features
 
+- rbd: add support for [CSI Snapshot Metadata Service RPCs](https://github.com/container-storage-interface/spec/blob/master/spec.md#snapshot-metadata-service-rpcs)
+
```
nixpanic, line 9: add something about it to the rbd documentation too?

internal/rbd/errors/errors.go
```diff
@@ -68,3 +68,8 @@ var (
 	// ErrGroupNotFound is returned when group is not found in the cluster on the given pool and/or namespace.
 	ErrGroupNotFound = fmt.Errorf("%w: RBD group not found", librbd.ErrNotFound)
 )
+
+type ErrorCode interface {
+	// Code returns the error code.
```
nixpanic: Comment `Code` -> `ErrorCode`.

Can you also add a comment for the interface itself, with a small explanation how (which kind of errors) it can be used?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,214 @@
...
+		})
+	default:
+		return errors.New("invalid stream type")
+	}
+}
+
+// StreamMetadata streams the block metadata for a snapshot.
+// If baseRBDSnap is provided, it calculates the delta between the
+// current snapshot and the base snapshot.
+func (rbdSnap *rbdSnapshot) StreamMetadata(
```
nixpanic: I have the feeling this function does too much. The `rbdSnapshot` struct should probably not send anything, but instead return `[]*csi.BlockMetadata` to the caller which can then send it out? Would that not make it a cleaner implementation?

internal/rbd/snapshot.go
nixpanic: > It's done at a lot of places currently too.

Yes, but that does not make it good practice :vomiting_face: 

> Switched to using conn.Copy() so we should be good.

Better, still ugly. But I think that is as good as we can do for now.

internal/rbd/types/snapshot.go
```diff
@@ -37,4 +38,19 @@ type Snapshot interface {
...
+	GetVolSize() int64
+
+	// GetRBDSnapID retrieves the RBD snapshot ID for the snapshot.
+	GetRBDSnapID(ctx context.Context) (uint64, error)
+
+	// StreamMetadata streams the block metadata for a snapshot.
+	// If baseRBDSnap is provided, the delta between the
+	// current snapshot and the base snapshot is streamed.
+	StreamMetadata(ctx context.Context,
+		stream grpc.ServerStream,
```
nixpanic: `grpc.ServerStream` really isn't nice here, see my other comment.

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	err = image.SetSnapByID(snapID)
+	if err != nil {
+		return fmt.Errorf("failed to set snapshot %q by ID %d: %w",
+			rbdSnap, snapID, err)
+	}
+
+	if baseRBDSnap != nil {
+		fromSnapID, err = baseRBDSnap.GetRBDSnapID(ctx)
```
nixpanic: Instead of exporting the `GetRBDSnapID()` function in the Snapshot interface, it can be an internal function too. In this case, typecasting `baseRBDSnap` to a `*rbdSnapshot` would be ok.

Consider a helper function for typecasting:
```go
func rbdSnapFromSnapshot(snap types.Snapshot) (*rbdSnapshot, error) {}
```

internal/rbd/types/snapshot.go
```diff
@@ -37,4 +38,19 @@ type Snapshot interface {
 
 	// SetVolumeGroup sets the CSI volume group ID in the snapshot.
 	SetVolumeGroup(ctx context.Context, creds *util.Credentials, vgID string) error
+
+	// GetVolSize returns the size of the snapshot in bytes.
+	GetVolSize() int64
+
+	// GetRBDSnapID retrieves the RBD snapshot ID for the snapshot.
+	GetRBDSnapID(ctx context.Context) (uint64, error)
```
nixpanic: This should be a non-backend-specific naming, drop RBD from it. Ideally remove it from the interface, see a previous comment.

internal/rbd/snap_diff.go
Rakshith-R: > I expect that the size of a single metadata for a changed block is relatively small. Start with something like 128, so that the number of repeated calls is reduced (will be less efficient?).

:+1: 

 
> A configuration option is definitely good to do at a later point in time. Does the sidecar provide metrics about the number of changed blocks returned per diff? With that, a more educated guess for a default can be made.

Currently only metrics regarding start & end time of requests is added 
@Nikhil-Ladha ^

internal/rbd/snap_diff.go
Rakshith-R: > I have the feeling this function does too much. 


Broke down this function to include multiple helpers

>The `rbdSnapshot` struct should probably not send anything, but instead return `[]*csi.BlockMetadata` to the caller which can then send it out? Would that not make it a cleaner implementation?

It's a streaming API, so we would be sending more than once.
Used this as inspiration to instead send down a func that only takes `[]*csi.BlockMetadata` as input and send stream response internally.

docs/rbd/deploy.md
```diff
@@ -524,3 +524,36 @@ module is enabled on the nodes running ceph-csi attachers.
...
+
+   * Streams metadata about allocated blocks in a snapshot
+   * Useful for full backup operations
+   * Returns block ranges that contain actual data
+
+* `GetMetadataDelta`:
+
+   * Streams metadata about block differences between two snapshots
+   * Optimal for incremental backups
+   * Returns only blocks that changed between snapshots
```
Madhu-1, line 577: This doc is for someone who wants to deploy and test it out, can we add relevant steps for the same?

internal/rbd/identityserver.go
```diff
@@ -72,6 +73,25 @@ func (is *IdentityServer) GetPluginCapabilities(
...
+			Type: &csi.PluginCapability_Service_{
+				Service: &csi.PluginCapability_Service{
+					Type: csi.PluginCapability_Service_SNAPSHOT_METADATA_SERVICE,
+				},
+			},
+		}
+		caps = append(caps, &snapDiffCaps)
+	}
+	if err != nil {
+		log.DebugLog(ctx, "error checking for snapshot diff by ID support: %s", err)
```
Madhu-1: log as error/warning?

internal/rbd/sms_controllerserver.go
```diff
@@ -0,0 +1,220 @@
...
+	req *csi.GetMetadataAllocatedRequest,
+	stream csi.SnapshotMetadata_GetMetadataAllocatedServer,
+) error {
+	if err := validateMetadataAllocatedReq(req); err != nil {
+		return err
+	}
+
+	snapshotID := req.GetSnapshotId()
+
+	ctx := context.WithValue(stream.Context(), log.ReqID, req.GetSnapshotId())
```
Madhu-1: log.ReqID will not be set for this one unless its added to middle ware? and also set the SnapshotID in the middleware itself?

internal/rbd/sms_controllerserver.go
```diff
@@ -0,0 +1,220 @@
...
+) error {
+	if err := validateMetadataDeltaReq(req); err != nil {
+		return err
+	}
+
+	baseSnapshotID := req.GetBaseSnapshotId()
+	targetSnapshotID := req.GetTargetSnapshotId()
+
+	ctx := context.WithValue(stream.Context(), log.ReqID, targetSnapshotID)
+	ctx = context.WithValue(ctx, "Base Snapshot ID", baseSnapshotID)
```
Madhu-1: set this in middleware?

internal/rbd/sms_controllerserver.go
```diff
@@ -0,0 +1,220 @@
...
+
+	baseRBDSnap, err := mgr.GetSnapshotByID(ctx, baseSnapshotID)
+	if errors.Is(err, rbderrors.ErrImageNotFound) {
+		log.ErrorLog(ctx, "snapshot %q not found: %v", baseSnapshotID, err)
+
+		return status.Errorf(codes.NotFound, "snapshot %q not found: %v", baseSnapshotID, err)
+	}
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
```
Madhu-1:
```suggestion
if err != nil {
    if errors.Is(err, rbderrors.ErrImageNotFound) {
		log.ErrorLog(ctx, "snapshot %q not found: %v", baseSnapshotID, err)

		return status.Errorf(codes.NotFound, "snapshot %q not found: %v", baseSnapshotID, err)
	}
		return status.Error(codes.Internal, err.Error())
	}

it would be nice to have one block for the error handling, take care of it in other places as well.

File: internal/rbd/snap_diff.go [hunk 40]
```diff
@@ -0,0 +1,195 @@
...
+) error {
+	var fromSnapID uint64
+	image, err := rbdSnap.open()
+	if err != nil {
+		return fmt.Errorf("failed to open image %q: %w", rbdSnap, err)
+	}
+	defer func() {
+		cErr := image.Close()
+		if cErr != nil {
+			log.WarningLog(ctx, "resource leak, failed to close image: %v", cErr)
```
Madhu-1, line 52: log it as error?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,195 @@
...
+	if baseSnap != nil {
+		// If a base snapshot is provided, set the FromSnapID to
+		// the base snapshot ID to get the delta between the two snapshots.
+		diffIterateByIDConfig.FromSnapID = fromSnapID
+	}
+
+	diffIterateErr := image.DiffIterateByID(diffIterateByIDConfig)
+	err = handleDiffIterateError(ctx, diffIterateErr)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to get diff: %s", err)
```
Madhu-1: log it at the caller? we dont need to log this at multiple places in this function?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,195 @@
...
+	log.DebugLog(ctx, "successfully completed GetMetadata operation")
+
+	// Successfully completed the diff operation.
+	return nil
+}
+
+// handleDiffIterateError checks the error returned by DiffIterateByID.
+// It handles specific error codes and returns a formatted error message
+// for unrecognized error codes.
+func handleDiffIterateError(ctx context.Context, err error) error {
```
Madhu-1: let the caller log it and we dont need to pass the ctx?

internal/rbd/snapshot.go
```diff
@@ -400,3 +400,48 @@ func (rbdSnap *rbdSnapshot) SetVolumeGroup(ctx context.Context, cr *util.Credent
...
 }
+
+// GetVolSize returns the size of the snapshot in bytes.
+func (rbdSnap *rbdSnapshot) GetVolSize() int64 {
+	return rbdSnap.VolSize
+}
+
+// rbdSnapFromSnapshot converts a types.Snapshot to a *rbdSnapshot.
+// It returns an error if the snapshot is nil or not of type *rbdSnapshot.
+func rbdSnapFromSnapshot(snap types.Snapshot) (*rbdSnapshot, error) {
```
Madhu-1, line 411: can we add a function to Manager interface to return `rbdSnapshot` we dont need to have implementation here we can directly pass `*rbdSnapshot` to the `StreamMetadata` method

Madhu-1 (commented):
some nits, overall LGTM

internal/rbd/snapshot.go
Rakshith-R: > can we add a function to Manager interface to return `rbdSnapshot` we dont need to have implementation here we can directly pass `*rbdSnapshot` to the `StreamMetadata` method


>*rbdSnapshot to the StreamMetadata method
this was the implementation in one of my earlier iteration

The current approach was suggested by @nixpanic in 

https://github.com/ceph/ceph-csi/pull/5347#discussion_r2128818985

internal/rbd/sms_controllerserver.go
Rakshith-R: > log.ReqID will not be set for this one unless its added to middle ware? and also set the SnapshotID in the middleware itself?

Yes, its not logged :disappointed: ,

we need a new middle ware for streamserver interceptor, current one does not work these stream rpcs.
I'll send a separate pr for that.

internal/rbd/sms_controllerserver.go
Rakshith-R: same as above
https://github.com/ceph/ceph-csi/pull/5347#discussion_r2166324902

Will add a separate pr for stream interceptor.

internal/rbd/snapshot.go
nixpanic: > can we add a function to Manager interface to return rbdSnapshot

No, the idea of the manager is to separate the CSI components from the RBD internals. Clean APIs for all the things make it a much more maintainable project. Eventually...

internal/rbd/snap_diff.go
nixpanic: I think of errors as fatal, a warning for a resource leak (unlikely to happen) is more suitable.

internal/rbd/sms_controllerserver.go
Madhu-1: Please open a tracker of it, so that we dont miss it

docs/rbd/deploy.md
```diff
@@ -527,7 +527,7 @@ If custom image is built for the rbd-plugin instance, make sure that it contains
 
 ## Changed Block Tracking (CBT)
 
-> **Warning**: Requires Ceph version that supports RBD snap diff by ID feature.
```
Nikhil-Ladha: These changes should be part of the previous `doc: add documentation for CBT feature` commit.

internal/rbd/sms_controllerserver.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	// defaultMaxResults is the internal limit for max results in metadata streaming
+	// This is set to a reasonable value to prevent excessive memory usage
+	// and ensure efficient processing of metadata requests.
+	// TODO: Consider making this configurable via a cmdline flag.
```
Nikhil-Ladha, line 36: Please have an issue created for this, so it's easier to track.

docs/rbd/deploy.md
Rakshith-R: > This doc is for someone who wants to deploy and test it out, can we add relevant steps for the same?

Since ceph-csi-operator is the one deploying the sidecar, I'll add a link here to that doc when its ready over there.

internal/rbd/identityserver.go
Rakshith-R: > log as error/warning?

Added it as warning since cephcsi can still function without these features.

internal/rbd/sms_controllerserver.go
Rakshith-R: https://github.com/ceph/ceph-csi/issues/5401
created

Rakshith-R:
updated the code with
- modification as per reviews
- added logic to handle luksHeader if it exits by adding it to initial offset and subtracting it from offset that will be returned to client.

PTAL

Madhu-1 (dismissed):

Rakshith-R:
https://github.com/ceph/ceph-csi/actions/runs/15966099828/job/45026731182?pr=5347
```
//nolint:unused // Unused code will be used in future.
```
forgot to remove this nolint comment :face_exhaling: 
removed it and rebased on top of devel

PTAL

Madhu-1 (dismissed):

Nikhil-Ladha (commented):
lgtm

Nikhil-Ladha (commented):
LGTM

Madhu-1:
@nixpanic Any final comments before merging as you have reviewed it

internal/rbd/identityserver.go
```diff
@@ -72,6 +73,25 @@ func (is *IdentityServer) GetPluginCapabilities(
 		}
 		caps = append(caps, &gcs)
 	}
+	if err != nil {
+		log.WarningLog(ctx, "error checking for group snapshot support: %s", err)
+	}
+
+	// RBDSnapDiffByID is used within the SnapshotMetadata service
+	rbdSnapDiffByIDSupported, err := features.SupportsRBDSnapDiffByID()
+	if err == nil && rbdSnapDiffByIDSupported {
```
nixpanic: you can write this a little simpler:

```go
	if err != nil {
		log.WarningLog(ctx, "error checking for snapshot diff by ID support: %s", err)
	} else if rbdSnapDiffByIDSupported {
		snapDiffCaps := csi.PluginCapability{
			Type: &csi.PluginCapability_Service_{
				Service: &csi.PluginCapability_Service{
					Type: csi.PluginCapability_Service_SNAPSHOT_METADATA_SERVICE,
				},
			},
		}
		caps = append(caps, &snapDiffCaps)
	}
```

internal/rbd/sms_controllerserver.go
```diff
@@ -0,0 +1,226 @@
...
+			requestedMaxResults, defaultMaxResults)
+
+		return defaultMaxResults
+	}
+	log.DebugLog(ctx, "using requested maxResults: %d", requestedMaxResults)
+
+	return requestedMaxResults
+}
+
+func validateMetadataAllocatedReq(req *csi.GetMetadataAllocatedRequest) error {
```
nixpanic, line 62: Consider including `SupportsRBDSnapDiffByID()` in the validate functions. VolumeGroupSnapshot calls still arrived at the controller, even when the support for it was not returned in the capabilities.

Or (cleaner?) check at the start of each procedure, like in `GetMetadataAllocated()`.

internal/rbd/sms_controllerserver.go
```diff
@@ -0,0 +1,226 @@
...
+
+		return status.Error(codes.Internal, err.Error())
+	}
+	defer rbdSnap.Destroy(ctx)
+
+	volSize := rbdSnap.GetVolSize()
+
+	startingOffset := req.GetStartingOffset()
+	if startingOffset >= volSize {
+		return status.Error(codes.OutOfRange, "starting offset is out of range")
```
nixpanic: consider including the requested offset/size

internal/rbd/sms_controllerserver.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	maxResults := normalizeMaxResults(ctx, req.GetMaxResults())
+
+	sendResponse := func(changedBlocks []*csi.BlockMetadata) error {
+		return stream.Send(&csi.GetMetadataAllocatedResponse{
+			BlockMetadataType:   csi.BlockMetadataType_VARIABLE_LENGTH,
```
nixpanic, line 133: is this variable? I assumed RBD uses fixed sized block?

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil
+	}
+
+	changedBlocks := make([]*csi.BlockMetadata, 0)
+	cb := createDiffIterateByIDCB(
+		ctx,
+		sendResponse,
+		&changedBlocks,
+		maxResults,
+		luksHeaderPadding,
```
nixpanic, line 96: should this not be `startingOffset`?

internal/rbd/snap_diff.go
nixpanic: Looks easier to understand already, thanks.

I understand that passing `sendResponse` makes things easier. It ties CSI+rbd functionality very closely together, which isn't very nice. But `librbd.DiffIterateByIDConfig{}` needs the callback, so I'm also not sure how to split the CSI/rbd things more. This is fine for now, maybe we can come up with an alternative approach later.

internal/rbd/snap_diff.go
```diff
@@ -0,0 +1,198 @@
...
+	ctx context.Context,
+	sendResponse types.SMSSendResponse,
+	changedBlocks *[]*csi.BlockMetadata,
+	maxResults int32,
+	luksHeaderPadding uint64,
+) librbd.DiffIterateCallback {
+	return func(offset, sizeBytes uint64, _ int, _ interface{}) int {
+		select {
+		case <-ctx.Done():
+			return int(codes.Canceled)
```
nixpanic, line 173: This callback should probably return go-ceph/rbd error values? Internally in go-ceph `getError(..)` is called, and that may convert some non-0 value as a real error. gRPC error codes are different from C errno values.

This returns a `librbd.DiffIterateCallback` function, so it needs to follow the librbd specs.

internal/rbd/snapshot.go
```diff
@@ -400,3 +400,48 @@ func (rbdSnap *rbdSnapshot) SetVolumeGroup(ctx context.Context, cr *util.Credent
 
 	return nil
 }
+
+// GetVolSize returns the size of the snapshot in bytes.
```
nixpanic: Just call this `GetSize()`, no need for the weird `Vol` part.

internal/rbd/types/snapshot.go
```diff
@@ -25,6 +25,9 @@ import (
 	"github.com/ceph/ceph-csi/internal/util"
 )
 
+// SMSSendResponse is a function type for sending responses in the SMS controller server.
+type SMSSendResponse func([]*csi.BlockMetadata) error
```
nixpanic: maybe call this `MetadataCallback`, there is no direct relation to the SMS service (it is just the only thing using it, now)

internal/rbd/types/snapshot.go
```diff
@@ -37,4 +40,16 @@ type Snapshot interface {
...
 	// SetVolumeGroup sets the CSI volume group ID in the snapshot.
 	SetVolumeGroup(ctx context.Context, creds *util.Credentials, vgID string) error
+
+	// GetVolSize returns the size of the snapshot in bytes.
+	GetVolSize() int64
+
+	// StreamMetadata streams the block metadata for a snapshot.
+	// If baseRBDSnap is provided, the delta between the
+	// current snapshot and the base snapshot is streamed.
+	StreamMetadata(ctx context.Context,
```
nixpanic: like `SMSSendResponse`, maybe call this function `ProcessMetadata` or something more generic?

internal/rbd/snap_diff.go
Rakshith-R: > should this not be `startingOffset`?

No, offset is passed below as part of config( 3 lines below).

luksHeaderPadding in order to adjust the offset sent to the user for encrypted volumes.

internal/rbd/snap_diff.go
Rakshith-R: > This callback should probably return go-ceph/rbd error values? Internally in go-ceph `getError(..)` is called, and that may convert some non-0 value as a real error. gRPC error codes are different from C errno values.
> 
> This returns a `librbd.DiffIterateCallback` function, so it needs to follow the librbd specs.

Actually the callback function is free to decide the error code, which is again interpreted by the user who has written the callback function.

librbd just calls this function and passed the error code as is to the caller if non-zero.

Rakshith-R: Example in go-ceph tests https://github.com/ceph/go-ceph/blob/4209816079604c4f56adbf91db7ec466aca07bb4/rbd/diff_iterate_test.go#L321-L340

internal/rbd/sms_controllerserver.go
Rakshith-R: > is this variable? I assumed RBD uses fixed sized block?

https://github.com/container-storage-interface/spec/blob/master/spec.md#metadata-format
```
The FIXED_LENGTH style requires a fixed value for the size_bytes field in all tuples in a given sequence.
The VARIABLE_LENGTH style does not constrain the value of the size_bytes field in a sequence.
```

```
/ # /snapshot-metadata-lister -max-results 10 -starting-offset 0 -namespace default -kubeconfig=“” -snapshot snapshot-1
Record#   VolCapBytes  BlockMetadataType   ByteOffset     SizeBytes
------- -------------- ----------------- -------------- --------------
      1      536870912   VARIABLE_LENGTH              0        4194304
/ # /snapshot-metadata-lister -max-results 10 -starting-offset 123213 -namespace default -kubeconfig=“” -snapshot snapshot-1
Record#   VolCapBytes  BlockMetadataType   ByteOffset     SizeBytes
------- -------------- ----------------- -------------- --------------
      1      536870912   VARIABLE_LENGTH         123213        4071091
```

ceph has fixed sized blocks for each object internally but sizeBytes is not constant in all cases. :thinking:

internal/rbd/snap_diff.go
nixpanic: ah, ok. And `codes.OK` is equal to 0?

internal/rbd/snap_diff_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (e *mockError) Error() string {
+	return fmt.Sprintf("%s: code %d", e.msg, e.code)
+}
+
+func (e *mockError) ErrorCode() int {
+	return e.code
+}
+
+func newMockError(code int, msg string) *mockError {
```
nixpanic, line 41: what about

```go
if code == 0 {
    return nil
}
```

That way, you can add a test case for `codes.OK` as well.

internal/rbd/types/snapshot.go
```diff
@@ -37,4 +40,16 @@ type Snapshot interface {
...
 	// SetVolumeGroup sets the CSI volume group ID in the snapshot.
 	SetVolumeGroup(ctx context.Context, creds *util.Credentials, vgID string) error
+
+	// GetSize returns the size of the snapshot in bytes.
+	GetSize() int64
+
+	// ProcessMetadata processes the block metadata for a snapshot.
+	// If baseRBDSnap is provided, the delta between the
+	// current snapshot and the base snapshot is processed.
+	ProcessMetadata(ctx context.Context,
```
nixpanic, line 50: a common coding practice is to have the callback as last parameter for functions (with optional callback arguments after that)

Rakshith-R: :+1: ,callback function are at the end now.

internal/rbd/snap_diff_test.go
Rakshith-R: > what about
> 
> ```go
> if code == 0 {
>     return nil
> }
> ```
> 
> That way, you can add a test case for `codes.OK` as well.

Added a codes.ok case inside handleDiffIterateError() function and corresponding test case is added too.

nixpanic (approved):

Madhu-1 (approved):

Madhu-1:
@Nikhil-Ladha PTAL as you have requested changes

Nikhil-Ladha (approved):

Madhu-1:
@mergifyio queue

Madhu-1:
@Mergifyio rebase

Madhu-1:
@mergifyio queue

Rakshith-R:
@mergifyio requeue

PR #5353, opened 2025-06-05
csi: validate incoming VolumeGroupSnapshot requests

nixpanic:
Add missing validation for VolumeGroupSnapshot requests.

### DIFF
--- a/.commitlintrc.yml
+++ b/.commitlintrc.yml
@@ -31,6 +31,7 @@ rules:
     - - build
       - cephfs
       - ci
+      - csi
       - csiaddons
       - cleanup
       - deploy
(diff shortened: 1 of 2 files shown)


Rakshith-R (approved):

Nikhil-Ladha (approved):
Wondering if "rbd" would have been a more appropriate commit title type for this change 🤔

nixpanic:
> Wondering if "rbd" would have been a more appropriate commit title type for this change 🤔

It's actually both. This validation isn't really _rbd_ specific, only that it is in the RBD group-controller. Because it is more of a CSI-verification, I through _csi_ as component would be better.

nixpanic:
@mergifyio queue

PR #5364, opened 2025-06-13
doc: add `v3` symlink to get current versions in pkg.go.dev

nixpanic:
https://pkg.go.dev/github.com/ceph/ceph-csi does not list the current
v3.14.0 release. It seems to come from the fact that a v3 release of a
Go module expects a v3/go.mod file in the repository. Instead of moving
all files from the root of the repository to a new `v3` directory, make
a symlink so that https://github.com/ceph/ceph-csi/v3/go.mod resolves
correctly (eventhough the go.mod does not reference /v3).

Steps to add a package (also has a suggestion for a version):
- https://pkg.go.dev/about#adding-a-package

The Go proxy protocol is used to track available versions, and this would be a request for the v3.14.0 version:
- `https://proxy.golang.org/github.com/ceph/ceph-csi/v3/@v/v3.14.0.info`

That results in an error like
```
not found: github.com/ceph/ceph-csi/v3@v3.14.0: invalid version: go.mod has non-.../v3 module path "github.com/ceph/ceph-csi" (and .../v3/go.mod does not exist) at revision v3.14.0
```

Adding the `v3` symlink to the root of the directory might be sufficient to get the documentation generated.


Madhu-1 (approved):

Nikhil-Ladha (approved):

Nikhil-Ladha:
@mergifyio queue

nixpanic:
@mergifyio backport release-v3.14

PR #5369, opened 2025-06-18
csiaddons: add Status, StatusMessage in GetVolumeReplicationInfo resp

Nikhil-Ladha:

Update csi-addons/spec pkg version to include the latest RPC changes and add `Status` and `StatusMessage` in GetVolumeReplicationInfo response.


internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +945,37 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
+
+		return replication.GetVolumeReplicationInfoResponse_UNKNOWN, "", err
+	}
+
+	localState := localStatus.GetState()
+
+	// check if mirroring state is up/down
+	isUp := localStatus.IsUP()
+
+	if isUp {
```
nixpanic: how about flipping this to `if !isUp`? Return early on error conditions makes the code a little cleaner. It would cause the large `switch` to not be indented anymore.

internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +945,37 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
+		case librbd.MirrorImageStatusStateUnknown.String():
+			return replication.GetVolumeReplicationInfoResponse_UNKNOWN, "mirroring is in unknown state", nil
+		case librbd.MirrorImageStatusStateError.String():
+			return replication.GetVolumeReplicationInfoResponse_ERROR, "mirroring is in error state", nil
+		case librbd.MirrorImageStatusStateReplaying.String():
+			return replication.GetVolumeReplicationInfoResponse_HEALTHY, "image/group is secondary", nil
+		case librbd.MirrorImageStatusStateStopped.String():
+			return replication.GetVolumeReplicationInfoResponse_HEALTHY, "image/group is primary", nil
+		default:
+			return replication.GetVolumeReplicationInfoResponse_UNKNOWN, "mirroring is in intermediatory state", nil
```
Madhu-1: How about returning the message present in the description as it contains the more meaningful details required for debugging. passing own message might not be helpful here.

internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +945,37 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
+		}
+	}
+
+	return replication.GetVolumeReplicationInfoResponse_DEGRADED, "Mirroring is down", nil
```
Madhu-1: set own message from cephcsi only if we dont find any message in the description.

internal/csi-addons/rbd/replication.go
Nikhil-Ladha: Ack, that can be done.
Though, only for the case of secondary I was thinking of passing it manually as the description for a secondary cluster is only "replaying", wdyt?

Madhu-1: Having native details returned from ceph helps in debugging. Lets add whatever the details we get from ceph , when we dont get anything we can have ceph-csi specific generic response.

Nikhil-Ladha: Updated.

internal/csi-addons/rbd/replication.go
Nikhil-Ladha: Makes sense, updated the condition now.

internal/csi-addons/rbd/replication.go
```diff
@@ -43,7 +43,7 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-// imageMirroringMode is used to indicate the mirroring mode for an RBD image.
+// imageMirroringMode is used to indicate the mirroring mode for an RBdesc image.
```
Madhu-1: typo here

internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +945,38 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
+	case librbd.MirrorImageStatusStateUnknown.String():
+		return replication.GetVolumeReplicationInfoResponse_UNKNOWN, desc, nil
+	case librbd.MirrorImageStatusStateError.String():
+		return replication.GetVolumeReplicationInfoResponse_ERROR, desc, nil
+	case librbd.MirrorImageStatusStateReplaying.String():
+		return replication.GetVolumeReplicationInfoResponse_HEALTHY, desc, nil
+	case librbd.MirrorImageStatusStateStopped.String():
+		return replication.GetVolumeReplicationInfoResponse_HEALTHY, desc, nil
+	default:
+		return replication.GetVolumeReplicationInfoResponse_UNKNOWN, desc, nil
```
Madhu-1: instead of multiple return, set a variable and return at the end?

also add a unit testing for this one?

internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +945,38 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
 	return nil
 }
+
+func getCurrentReplicationStatus(ctx context.Context,
+	mirrorGlobalStatus types.GlobalStatus,
+) (replication.GetVolumeReplicationInfoResponse_Status, string, error) {
+	// get the state of replication
+	localStatus, err := mirrorGlobalStatus.GetLocalSiteStatus()
+	if err != nil {
+		log.ErrorLog(ctx, err.Error())
```
Madhu-1: Lets log it at the caller? so that we dont need to pass the ctx

internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +945,38 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
+
+		return replication.GetVolumeReplicationInfoResponse_UNKNOWN, "", err
+	}
+
+	localState := localStatus.GetState()
+	desc := localStatus.GetDescription()
```
Madhu-1: No need for extra variable return `localStatus.GetDescription()` ?

Madhu-1 (commented):
small nits.

internal/csi-addons/rbd/replication.go
Nikhil-Ladha: This way it is cleaner I feel, otherwise we need to call the function at all the places which is a bit ugly IMO.
Probably, we can remove the `localState` var as it is used only at one place.

Rakshith-R: Do we really need the entire description in statusMessage field ? 
It may get too long ? 
Maybe we are looking for something specific within the description ? 

https://github.com/ceph/ceph-csi/blob/a3c4b2a54ca272d9998107ca0d105489f253d3d3/internal/csi-addons/rbd/replication_test.go#L228

Madhu-1: Yes that correct, we dont need the complete description, first part of it is required

nixpanic: That state should probably be obtained through the `types.SyncInfo` interface, so extend that here:

https://github.com/ceph/ceph-csi/blob/a3c4b2a54ca272d9998107ca0d105489f253d3d3/internal/rbd/mirror.go#L447

Nikhil-Ladha: @nixpanic have the `GetState` method already handles the parsing of `up+stopped` and such states and returns only the `state` as `stopped` from it, so I don't think we need another method to parse that under `types.SyncInfo`

Nikhil-Ladha:
Updatd the code and also added unit tests for the function, ptal.

internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +947,35 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
+	if !isUp {
+		return replication.GetVolumeReplicationInfoResponse_DEGRADED, desc, nil
+	}
+
+	resp := replication.GetVolumeReplicationInfoResponse_UNKNOWN
+	switch localStatus.GetState() {
+	case librbd.MirrorImageStatusStateError.String():
+		resp = replication.GetVolumeReplicationInfoResponse_ERROR
+	case librbd.MirrorImageStatusStateReplaying.String():
+		fallthrough
```
Madhu-1: instead of fallthrough add mutliplecase to avoid problems in future? where we need to add more states?

internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +947,35 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
+	case librbd.MirrorImageStatusStateStopped.String():
+		resp = replication.GetVolumeReplicationInfoResponse_HEALTHY
+	}
```
Madhu-1, line 982: add a default case and log it, This helps for debugging in future?

Madhu-1 (commented):
small nit, LGTM

internal/csi-addons/rbd/replication.go
Nikhil-Ladha: Fallthrough was only for replaying and stopped state.
We don't need `healthy` state for any other state as of now.

internal/csi-addons/rbd/replication.go
Nikhil-Ladha: Well, if we log a default case here then we would have to pass the ctx (again) 😅

Madhu-1: now make sense to log it for debugging

internal/csi-addons/rbd/replication.go
Madhu-1: add a comment if you still need to keep fallthrough, its easy to make mistakes after few years it will go unnoticed if we add another case b/w.

internal/csi-addons/rbd/replication.go
Nikhil-Ladha: Ack, added a log.

internal/csi-addons/rbd/replication.go
Nikhil-Ladha: Had to remove fallthrough as golangci-lint was complaining, but added a comment for future ref.

Madhu-1 (approved):

internal/csi-addons/rbd/replication.go
```diff
@@ -937,3 +947,39 @@ func checkVolumeResyncStatus(ctx context.Context, localStatus types.SiteStatus)
...
+	// check if mirroring state is up/down
+	isUp := localStatus.IsUP()
+
+	if !isUp {
+		return replication.GetVolumeReplicationInfoResponse_DEGRADED, desc, nil
+	}
+
+	resp := replication.GetVolumeReplicationInfoResponse_UNKNOWN
+	switch localStatus.GetState() {
+	case librbd.MirrorImageStatusStateError.String():
```
nixpanic, line 972: **nit, for future cleanup:** `librbd` should ideally only be use in the `internal/rbd` package, everything else should consume the `internal/rbd/types.Volume` (sub-)interface.

Nikhil-Ladha: Hmm, that is something that we could plan to do in a different PR as a cleanup. Though, I am not sure how feasible that is 🤔

Rakshith-R (approved):

Rakshith-R:
@mergifyio queue

PR #5409, opened 2025-07-03
doc: design doc for non-graceful node shutdown

iPraveenParihar:

doc: design doc for non-graceful node shutdown 

When `ControllerUnpublishVolume` is called without the node first having cleaned up the volume (via `NodeUnstageVolume` and `NodeUnpublishVolume`), the CSI driver has no opportunity to revoke the node’s access to the volume.The node may still hold active mounts, open file handles, or client sessions. This can lead to data corruption due to writes from disconnected yet still-active clients.

**Checklist:**

* [ ] **Commit Message Formatting**: Commit titles and messages follow guidelines in the [developer guide](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#commit-messages).
* [ ] Reviewed the developer guide on [Submitting a Pull Request](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#development-workflow)

---

<details>

</details>

### DIFF
--- a/docs/design/proposals/non-graceful-node-shutdown.md
+++ b/docs/design/proposals/non-graceful-node-shutdown.md
@@ -1,0 +1,173 @@
+# Non graceful node shutdown
+
+In Kubernetes, when a node becomes dysfunctional or is intentionally drained,
+taints `node.kubernetes.io/out-of-service` may be applied to mark the node
+as unavailable _refer [k8s doc](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#non-graceful-node-shutdown)_.
+This results in the forceful deletion of pods scheduled on the node and the
+cleanup of associated VolumeAttachment objects. However, this forceful cleanup
+sequence bypasses key CSI lifecycle calls most notably, the
+`NodeUnpublishVolume` and `NodeUnstageVolume` entry points are not invoked
+prior to `ControllerUnpublishVolume` call.
+
+## Problem
+
+When `ControllerUnpublishVolume` is called without the node first having
+cleaned up the volume (via `NodeUnpublishVolume` and `NodeUnstageVolume`),
+the CSI driver has no opportunity to revoke the node's access to the volume.
+The node may still hold active mounts, open file handles, or client sessions.
+This can lead to data corruption as applications may still be running on the
+broken node with active client sessions, even though the node is marked as out
+of service.
+
+## Important Note
+
+> ⚠️ **WARNING**: When a node becomes out of service, its mounts and device
+> mappings will persist until the node undergoes a complete power lifecycle
+> (includes shutdown and startup). To prevent data inconsistency or corruption,
+> administrators **MUST NOT** remove the `node.kubernetes.io/out-of-service`
+> taint until the node has successfully completed a full power cycle. Removing
+> the taint prematurely may leave stale device states, active client sessions,
+> or lingering mounts, which can lead to serious data integrity issues.
+
+## Proposed solution
+
+To ensure safe volume reuse and prevent stale client access during node
+disruptions, the proposed solution is to track the client address during
+the `NodeStageVolume()` operation and store it in the image or subvolume
+metadata under the key: `.[rbd|cephfs].csi.ceph.com/clientAddress/<NodeId>`.
+
+> Note: This metadata should not be copied when creating clones, snapshots,
+> or mirror images of the volume.
+
+This stored address can then be used:
+
+* In `ControllerUnpublishVolume()` to blocklist or evict the client if the node
+  has the `out-of-service` taints, ensuring it no longer has access to the
+  volume.
+
+* In `ControllerPublishVolume()` to remove the client from the blocklist,
+  allowing the volume to be accessed again when the node recovers.
+
+## Implementation Details
+
+### ControllerPublishVolume()
+
+* Check if the metadata key `.[rbd|cephfs].csi.ceph.com/clientAddress/<NodeId>`
+  exists (`NodeId` from the request)
+
+* If it exists:
+   * Remove the client from the blocklist:
+
+    ```
+    ceph osd blocklist rm <clientAddress>
+    ```
+
+   * Remove the metadata key after removing from blocklist
+
+    ```
+    # RBD
+    rbd image-meta remove <pool>/<namespace>/<image> .rbd.csi.ceph.com/clientAddress/<NodeId>
+
+    # CephFS
+    ceph fs subvolume metadata rm <vol_name> <sub_name> .cephfs.csi.ceph.com/clientAddress/<NodeId> [<group_name>]
+    ```
+
+### NodeStageVolume()
+
+* **Applicable to**: RBD and CephFS
+
+* Before volume staging:
+   * Retrieve the client address:
+
+   * Always set it in the image/subvolume metadata:
+    (`NodeId` from the plugin container argument `--nodeid`)
+
+    ```
+    .[rbd|cephfs].csi.ceph.com/clientAddress/<NodeId>: <clientAddress>
+    ```
+
+   * Continue with Volume staging
+
+### ControllerUnpublishVolume()
+
+* Retrieve the Node object using the provided `NodeId` from the request.
+
+* If the node has the `out-of-service` taint:
+   * Check for the metadata key:
+
+    ```
+    .[rbd|cephfs].csi.ceph.com/clientAddress/<NodeId>
+    ```
+
+   * If present, proceed to revoke client access:
+      * **For RBD**:
+
+      ```
+      # Add client to blocklist with extended duration to prevent stale client access.
+      # The blocklist must persist until we can confirm the node has gone through
+      # a complete power cycle, as premature expiration could lead to data corruption
+      ceph osd blocklist add <clientAddress> 157784760
+      ```
+
+      * **For CephFS**:
+         * List active clients and match against `clientAddress` to get the `clientId`.
+            * Evict the client and blocklist the address:
+
+        ```
+        ceph tell mds.* client evict id=<clientId>
+        # Add client to blocklist with extended duration to prevent stale client access.
+        # The blocklist must persist until we can confirm the node has gone through
+        # a complete power cycle, as premature expiration could lead to data corruption
+        ceph osd blocklist add <clientAddress> 157784760
+        ```
+
+* If the node doesn't have the `out-of-service` taint:
+   * Remove the `.[rbd|cephfs].csi.ceph.com/clientAddress/<NodeId>` metadata
+      key
+
+    ```
+    # RBD
+    rbd image-meta remove <pool>/<namespace>/<image> .rbd.csi.ceph.com/clientAddress/<NodeId>
+
+    # CephFS
+    ceph fs subvolume metadata rm <vol_name> <sub_name> .cephfs.csi.ceph.com/clientAddress/<NodeId> [<group_name>]
+    ```
+
+### Workaround for older PVs
+
+**Problem**: `ControllerPublishVolume()`/`ControllerUnpublishVolume()` requires
+controller-publish-secret. The secret needed to access the Ceph cluster may be
+missing for older PVs if the following parameters were not specified in their
+corresponding StorageClass at the time of provisioning:
+
+```
+csi.storage.k8s.io/controller-publish-secret-name
+csi.storage.k8s.io/controller-publish-secret-namespace
+```
+
+**Solution 1**: Fallback to default secrets if available in csi-config-map
+ConfigMap.
+
+```yaml
+[
+  {
+    "clusterID": "my-cluster",
+    "rbd": {
+      ...
+      "controllerPublishSecretRef": {
+        "secretName": "publish-secret-name",
+        "secretNamespace": "publish-secret-namespace"
+      },
+      ...
+    },
+    "cephfs": {
+      ...
+      "controllerPublishSecretRef": {
+        "secretName": "publish-secret-name",
+        "secretNamespace": "publish-secret-namespace"
+      },
+      ...
+    }
+  }
+]
+```


docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,127 @@
+# Non graceful node shutdown
+
+In Kubernetes, when a node becomes unhealthy or is intentionally drained,
+taints `node.kubernetes.io/out-of-service` may be applied to mark the node
```
nixpanic, line 4: link to the Kubernetes admin guide please

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,127 @@
+# Non graceful node shutdown
+
+In Kubernetes, when a node becomes unhealthy or is intentionally drained,
```
nixpanic: _unhealthy_ is also used for volumes, maybe call it _disfunctional_?

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,127 @@
...
+notably, the `NodeUnpublishVolume` and `NodeUnstageVolume` entry points are not
+invoked prior to `ControllerUnpublishVolume` call.
+
+## Problem
+
+When `ControllerUnpublishVolume` is called without the node first having cleaned
+up the volume (via `NodeUnstageVolume` and `NodeUnpublishVolume`), the CSI driver
+has no opportunity to revoke the node’s access to the volume.The node may still
+hold active mounts, open file handles, or client sessions. This can lead to data
+corruption due to writes from disconnected yet still-active clients.
```
nixpanic: _disconnected_ isn't really correct, as a disconnected client can not reach the storage. A client may re-connect later on, causing havoc. Or, the node may be somehow broken, but apps may still be running, this is the major concern.

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Proposed solution
+
+To ensure safe volume reuse and prevent stale client access during node disruptions,
+the proposed solution is to track the client address during the `NodeStageVolume()`
+operation and store it in the image or subvolume metadata under the key:
+`csi.storage.k8s.io/clientAddress/<NodeId>`.
```
nixpanic: `csi.storage.k8s.io` isn't appropriate, use `csi.ceph.io` or something similar instead

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,127 @@
...
+
+### ControllerPublishVolume()
+
+- Check if the metadata key `csi.storage.k8s.io/clientAddress/<NodeId>`
+exists(`NodeId` from the request)
+- If it exists:
+  - Remove the client from the blocklist:
+
+    ```bash
+    ceph osd blocklist rm <clientAddress+nonce>
```
nixpanic: also remove the metadata key?

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,127 @@
...
+controller-publish-secret. The secret needed to access the Ceph cluster may be
+missing for older PVs if the following parameters were not specified in their
+corresponding StorageClass at the time of provisioning:
+
+```
+csi.storage.k8s.io/controller-publish-secret-name
+csi.storage.k8s.io/controller-publish-secret-namespace
+```
+
+**Solution 1**: Fetch secrets somehow using volumeID.
```
nixpanic: how?

docs/design/proposals/non-graceful-node-shutdown.md
Madhu-1: Ensure that this is not copied to the clones and snapshots.

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,127 @@
...
+### NodeStageVolume()
+
+- **Applicable to**: RBD and CephFS
+- Before volume staging
+  - Retrieve the client address:
+  - Always set it in the image/subvolume metadata:
+    (`NodeId` from the plugin container argument `--nodeid`)
+
+    ```
+    csi.storage.k8s.io/clientAddress/<NodeId>: <clientAddress+nonce>
```
Madhu-1: This steps is missing the details to get the clientAdress+nonce.

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,127 @@
...
+
+  - If present, proceed to revoke client access:
+    - **For RBD**:
+
+      ```bash
+      ceph osd blocklist add <clientAddress+nonce>
+      ```
+
+    - **For CephFS**:
+      - List active clients and match against `clientAddress` to get the `clientId`.
```
Madhu-1: Are we not going to store it in the metadata?

docs/design/proposals/non-graceful-node-shutdown.md
iPraveenParihar: @Madhu-1, does this get us the required address?
https://github.com/ceph/ceph-csi/blob/cfb3979d6b8363c00d4470b796da542acc6f7df4/internal/util/connection.go#L154-L163

docs/design/proposals/non-graceful-node-shutdown.md
Madhu-1: How about below

* For RBD using something like `.rbd.csi.ceph.com` so that we dont need to worry about the replication but we still need to take care of clone/snapshots
* For cephfs use `.cephfs.csi.ceph.com` ?

docs/design/proposals/non-graceful-node-shutdown.md
Madhu-1: it will give us the IP address not nonce

iPraveenParihar: Ahh okay, from comment I thought we get the IP+nonce

https://github.com/ceph/ceph-csi/blob/cfb3979d6b8363c00d4470b796da542acc6f7df4/internal/csi-addons/rbd/network_fence.go#L153-L163


For RBD,  we can get the clientAddres+nonce from below node path
```
[root@c1-m03 /]# cat /sys/devices/rbd/0/client_addr
192.168.39.34:0/2458835906
```
suggested by @nixpanic

docs/design/proposals/non-graceful-node-shutdown.md
iPraveenParihar: ➕  `.rbd.csi.ceph.com` and `.cephfs.csi.ceph.com` will be consistent with other metadata key as well.
https://github.com/ceph/ceph-csi/blob/cfb3979d6b8363c00d4470b796da542acc6f7df4/internal/rbd/encryption.go#L57-L66

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,136 @@
...
+invoked prior to `ControllerUnpublishVolume` call.
+
+## Problem
+
+When `ControllerUnpublishVolume` is called without the node first having cleaned
+up the volume (via `NodeUnstageVolume` and `NodeUnpublishVolume`), the CSI driver
+has no opportunity to revoke the node's access to the volume. The node may still
+hold active mounts, open file handles, or client sessions. This can lead to data
+corruption as applications may still be running on the broken node with active
+client sessions, even though the node is marked as out of service.
```
idryomov: https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#non-graceful-node-shutdown says that `node.kubernetes.io/out-of-service` taint should be added only if "the node is already in shutdown or power off state (not in the middle of restarting)".  How can there be active client sessions and applications still be running there if the node is physically powered off?

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,136 @@
...
+
+    ```
+    .csi.ceph.io/clientAddress/<NodeId>
+    ```
+
+  - If present, proceed to revoke client access:
+    - **For RBD**:
+
+      ```bash
+      ceph osd blocklist add <clientAddress+nonce>
```
idryomov: A "problem" with OSD blocklist entries is that by default they expire after 1 hour.  The expiration timeout is in seconds and can be changed by passing an additional integer argument.  For example, `ceph osd blocklist add 1.2.3.4/1234 86400` would persist the blocklist entry for 1 day.

If the goal here is to accommodate scenarios where the admin applies the out-of-service taint without knowing that the node is actually powered off, the blocklist entry needs to persist until it becomes known for sure that the node went through (an equivalent of) a power cycle.  Allowing the blocklist entry to expire before that point can lead to "stale client access" and therefore data corruption.

docs/design/proposals/non-graceful-node-shutdown.md
Madhu-1: @idryomov we are trying to cover the worst case where we lost access to the nodes and we want to block the client on that node to avoid the problems to be on safer side. When the taint is added the kubernetes assumes that node is already in power off state and starts moving the work load to other nodes, we are trying to have more fencing mechanism to ensure that we are fencing off the node to avoid the access to the client (if they run or ever comes back when the taint still exists)

docs/design/proposals/non-graceful-node-shutdown.md
Madhu-1: we will blocklist of the max period (like 5 years) or indefinite time. @iPraveenParihar can you please update the time for that as well.

iPraveenParihar: _just for referenece: https://github.com/ceph/ceph-csi/blob/cfb3979d6b8363c00d4470b796da542acc6f7df4/internal/csi-addons/networkfence/fencing.go#L256-L261_

docs/design/proposals/non-graceful-node-shutdown.md
idryomov: What about the point when the taint is removed?  Since the admin can apply the taint without knowing that the node is actually powered off (and that being precisely what this design is trying to guard against), I'm wondering whether the admin can also remove the taint whenever they feel like it?

Put differently, is it enforced that the node goes through (an equivalent of) a power cycle before the taint is removed?

Madhu-1: When the node is recovered and admin can access the node,  If the node is available (kubelet+api-server) ensure that all pods are removed from the node (to keep it consistent with the ETCD data of pods) and no new pods are scheduled on the node. 

When everything is fine the admin can remove the taint and mark the node available for new pods scheduling. The expectation is that the admin will remove the taints when there are no pods running on the node.  As we are implementing this in the RPC spec. CSI takes care of unfencing (when a pod is scheduled for the node) and  allow the future operations.

idryomov: Does the term "recovered" include a power cycle?

No pods running on the node is very different from (an equivalent of) a power cycle.  Let's say the taint gets added on an active/running node with a bunch of mapped RBD devices and the corresponding mounts.  Later, the admin realizes that they screwed up and decides to remove the taint.  What is supposed to ensure that those mounts are teared down before Ceph CSI unfences?

Madhu-1: In this case, when the taint is added the kubelet will send the RPC calls (NodeUnpublish,NodeUnstage,ControllerUnpublish) This is very cephCSI blocklists the IP (which means the the mount is already teared down by the kubelet and when the taint is remvoed ceph-csi will unfence in the next ControllerPublish where there are no mounts exists and it will be followed by the NodeStage and NodePublish to map the image on the node.

idryomov: Hrm, I may have muddied the waters with "an active/running node".  I'm trying to explore the scenario described in this doc -- one where `NodeUnstageVolume` and `NodeUnpublishVolume` calls either aren't issued at all or don't make it and only `ControllerUnpublishVolume` goes in effect.  If the node isn't powered off, the mounts would still be there and would be left behind, right?  My understanding is that only blocklisting would occur as part of handling `ControllerUnpublishVolume`.

I'm missing what exactly happens later, when the node becomes reachable to the admin/Kubernetes.  If the power cycle isn't required to take place, who is responsible for tearing down the mounts that were left behind?

(What I meant by "an active/running node" is that the node being unreachable to the admin/Kubernetes doesn't imply that it's unable to talk to the Ceph cluster.  Being able to reach even just one of the OSDs is enough for the client to be considered active from the Ceph POV.)

Madhu-1: > Hrm, I may have muddied the waters with "an active/running node". I'm trying to explore the scenario described in this doc -- one where `NodeUnstageVolume` and `NodeUnpublishVolume` calls either aren't issued at all or don't make it and only `ControllerUnpublishVolume` goes in effect. If the node isn't powered off, the mounts would still be there and would be left behind, right? My understanding is that only blocklisting would occur as part of handling `ControllerUnpublishVolume`.
> 

Yes thats a correct.

> I'm missing what exactly happens later, when the node becomes reachable to the admin/Kubernetes. If power cycle isn't required to take place, who is responsible for tearing down the mounts that were left behind?

The kubelet will take care of it as soon the node comes back. And later its admin responsibility to power of the node as well.  i hope @iPraveenParihar already tested this case as well.

idryomov: > The kubelet will take care of it as soon the node comes back.

Can you go into more detail here?  How exactly does kubelet do that?  If one of the steps there fails for some reason, is the admin prevented from removing the taint?

> And later its admin responsibility to power of the node as well.

... but this isn't enforced?

Madhu-1: > > The kubelet will take care of it as soon the node comes back.
> 
> Can you go into more detail here? How exactly does kubelet do that? If one of the steps there fails for some reason, is the admin prevented from removing the taint?
> 

I havent looked into the kubelet code but we can check on that one @iPraveenParihar PTAL. Not yet, kubernetes doesnt have any checks before removing the taint, its left to the admin to validate everything and remove the taint (we can also document if anything is required from our side as well) 
> > And later its admin responsibility to power of the node as well.
> 
> ... but this isn't enforced?

yes nothing is enforced in kubernetes its only documented step.

iPraveenParihar: From what I have tested...

> Let's say the taint gets added on an active/running node with a bunch of mapped RBD devices and the corresponding mounts. Later, the admin realizes that they screwed up and decides to remove the taint. What is supposed to ensure that those mounts are teared down before Ceph CSI unfences?

### Scenario 1: Node is healthy (Node is Ready, kubelet is running), taints are removed without reboot.

- If a taint is added to a healthy node where the kubelet is functional, existing pods are marked for deletion but stay in a Terminating or Error state.

- These pods aren’t forcefully deleted because the node's status is not NotReady.

- NodeUnpublishVolume and NodeUnstageVolume are not called, since the CSI nodeplugin pods are also removed.

- ControllerUnpublishVolume is not triggered either — VolumeAttachment objects stay intact until pods are forcefully deleted.

- New pods scheduled to another node will fail with attach errors due to the image still being mapped.

- The admin realizes that they screwed up and decides to remove the taint, nodeplugin pods come back up, kubelet issues the pending NodeUnpublishVolume/NodeUnstageVolume calls for the pod/s in terminating/error state, mappings are removed, and new pods on the other node can finally attach the volume successfully.

> I'm trying to explore the scenario described in this doc -- one where NodeUnstageVolume and NodeUnpublishVolume calls either aren't issued at all or don't make it and only ControllerUnpublishVolume goes in effect. If the node isn't powered off, the mounts would still be there and would be left behind, right? My understanding is that only blocklisting would occur as part of handling ControllerUnpublishVolume.

### Scenario 2: Node is unresponsive (Node is NotReady, kubelet is not running), taints are removed without reboot

- The unresponsive node is tainted.

- Pods are forcefully deleted. 

- VolumeAttachment deletion triggers ControllerUnpublishVolume, which blocks the RBD image.

- New pods are scheduled on other nodes and can attach without issue.

- Here, now if the node comes back without a reboot and Kubelet is running, NodeUnpublishVolume/NodeUnstageVolume would not be called as the pod/s is already deleted. Since the node hasn’t rebooted, the original RBD device mappings and mount points still persist.


> The kubelet will take care of it as soon the node comes back. And later its admin responsibility to power of the node as well. i hope @iPraveenParihar already tested this case as well.

@Madhu-1, the mounts and mappings will persist unless the node is rebooted. That’s why we discussed that the admin should only remove the taint after the node has come back from a successful shutdown. This ensures clean teardown and avoids stale device state.

idryomov: > Here, now if the node comes back without a reboot and Kubelet is running, NodeUnpublishVolume/NodeUnstageVolume would not be called as the pod/s is already deleted. Since the node hasn’t rebooted, the original RBD device mappings and mount points still persist.

This is exactly what I suspected and why I kept probing here ;)  This should be obvious by now but I want to restate it just in case: if the node a) isn't properly cleaned up (i.e. all mounts and RBD device mappings are teared down via `NodeUnpublishVolume`/`NodeUnstageVolume`) or b) doesn't go through a power cycle to waive any concerns around cleanup, removing the blocklist entry in `ControllerPublishVolume` can lead to data corruption.  This was added after my earlier comment
```
       # The blocklist must persist until we can confirm the node has gone through
       # a complete power cycle, as premature expiration could lead to data corruption
```
but not in the main body of the doc and previously wasn't mentioned at all.  I'd suggest highlighting this in a separate paragraph.

Madhu-1: Thanks @idryomov for probing on this one to ensure we dont miss anything and we document it :) , @iPraveenParihar Thanks for confirming as my memory is really old for this as this exists for many releases and i tested very long back. We need to document this in our documents what is the exception from the ceph-csi  to avoid any problems related to data.

docs/design/proposals/non-graceful-node-shutdown.md
iPraveenParihar: > @Madhu-1, does this get us the required address?
> 
> https://github.com/ceph/ceph-csi/blob/cfb3979d6b8363c00d4470b796da542acc6f7df4/internal/util/connection.go#L154-L163

I tried this, but it returns a different nonce, IP being same.
The method mentioned earlier in previous comment only applies to block devices. (`/sys/devices/rbd/0/client_addr`).
I haven't found any other way to retrieve the IP and nonce together except calling rbd status/ceph tell API calls.

That said, I don't see any difference between blocklisting by just IP vs IP+nonce.  Since the primary goal is to block the access from the tainted node, so blocklisting by IP alone should suffice?

^^ @Madhu-1 @nixpanic @Rakshith-R

docs/design/proposals/non-graceful-node-shutdown.md
iPraveenParihar: we could store the clientId for CephFs subvolumes, that would add one extra metadata entry but could save us from listing the active clients and matching  operation for each subvolume request.

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,149 @@
...
+This results in the forceful deletion of pods scheduled onthe node and the
+cleanup of associated VolumeAttachment objects. However, this forceful cleanup
+sequence bypasses key CSI lifecycle calls most notably, the `NodeUnpublishVolume`
+and `NodeUnstageVolume` entry points are not invoked prior to `ControllerUnpublishVolume`
+call.
+
+## Problem
+
+When `ControllerUnpublishVolume` is called without the node first having cleaned
+up the volume (via `NodeUnstageVolume` and `NodeUnpublishVolume`), the CSI driver
```
Madhu-1: `(via `NodeUnstageVolume` and `NodeUnpublishVolume`)` to `(via `NodeUnpublishVolume` and `NodeUnstageVolume`)`

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,149 @@
...
+### ControllerPublishVolume()
+
+- Check if the metadata key `.[rbd|cephfs].csi.ceph.com/clientAddress/<NodeId>` exists
+(`NodeId` from the request)
+
+- If it exists:
+    - Remove the client from the blocklist:
+
+    ```bash
+    ceph osd blocklist rm <clientAddress+nonce>
```
Madhu-1: remove the nonce as we will go with ip blocklisting?

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,149 @@
...
+    ```
+
+    - Remove the metadata key after removing from blocklist
+
+    ```
+    # RBD
+    rbd image-meta remove <pool>/<image> .rbd.csi.ceph.com/clientAddress/<NodeId>
```
Madhu-1: add rbd namespace as well to the command

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Workaround for older PVs
+
+**Problem**: `ControllerPublishVolume()`/`ControllerUnpublishVolume()` requires
+controller-publish-secret. The secret needed to access the Ceph cluster may be
+missing for older PVs if the following parameters were not specified in their
+corresponding StorageClass at the time of provisioning:
+
+```
+csi.ceph.io/controller-publish-secret-name
```
Madhu-1: why we have `csi.ceph.io`?

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,149 @@
...
+    "cephfs": {
+      ...
+      "controllerPublishSecretRef": {
+        "secretName": "publish-secret-name",
+        "secretNamespace": "publish-secret-namespace"
+      },
+      ...
+    }
+  }
+]
```
Madhu-1, line 172: Can you please leave a note that the node has to go through the power lifecycle before removing the taint or else we are going to have data inconsistency/corruption problem

docs/design/proposals/non-graceful-node-shutdown.md
iPraveenParihar: mistake should be `csi.storage.k8s.io/controller-publish-secret-name`

docs/design/proposals/non-graceful-node-shutdown.md
iPraveenParihar: Added as Important note above.

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,173 @@
...
+controller-publish-secret. The secret needed to access the Ceph cluster may be
+missing for older PVs if the following parameters were not specified in their
+corresponding StorageClass at the time of provisioning:
+
+```
+csi.storage.k8s.io/controller-publish-secret-name
+csi.storage.k8s.io/controller-publish-secret-namespace
+```
+
+**Solution 1**: Fallback to default secrets if available in csi-config-map
```
iPraveenParihar, line 148: @Madhu-1, are we good with this solution to address the older PVCs?

Madhu-1: Yes we are good with this one 👍🏻

Rakshith-R (dismissed):
LGTM

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,173 @@
...
+The node may still hold active mounts, open file handles, or client sessions.
+This can lead to data corruption as applications may still be running on the
+broken node with active client sessions, even though the node is marked as out
+of service.
+
+## Important Note
+
+> ⚠️ **WARNING**: When a node becomes out of service, its mounts and device
+mappings will persist until the node goes through a complete power lifecycle
+(shutdown and restart). To prevent data inconsistency or corruption,
```
Madhu-1: we dont need restart here, just mention that power cycle( includes shutdown and start it )

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,173 @@
...
+administrators **MUST NOT** remove the `node.kubernetes.io/out-of-service`
+taint until the node has successfully completed a full shutdown and restart
+cycle. This ensures proper cleanup of stale device state and prevents data
```
Madhu-1: `full shutdown and restart cycle` to `full power cycle`

docs/design/proposals/non-graceful-node-shutdown.md
```diff
@@ -0,0 +1,173 @@
...
+corruption from lingering mounts or active client sessions.
```
Madhu-1:
```suggestion
Removing the taint prematurely may leave stale device state, active client sessions, or lingering mounts, which can lead to serious data integrity issues.
```

Madhu-1 (commented):
small nits, LGTM

Madhu-1 (approved):

Rakshith-R (approved):

Rakshith-R:
@mergifyio rebase

Rakshith-R:
@mergifyio requeue

PR #5416, opened 2025-07-08
ci: dont upload the snyk report to the github

Madhu-1:
There is no need to have the report pushed to the release artifacts and also we need to provide extra permission to write it, As it was never used.

PS: we can reenable this feature later (if required) on and someone need to check this and as well as golang snyk as well and add support for the same.


nixpanic (approved):

iPraveenParihar (approved):

Madhu-1:
@mergifyio queue

PR #5422, opened 2025-07-10
doc: clarifiy NVMe-oF design a little more

nixpanic:
After collecting feedback, it has been confirmed that the Ceph-CSI/RBD
component can be used to create the RBD-images.

The NVMe Subsystems may not have been pre-created, so Ceph-CSI should
verify if the subsystem exists, and create it if needed.

### DIFF
--- a/docs/design/proposals/nvme-of.md
+++ b/docs/design/proposals/nvme-of.md
@@ -64,13 +64,13 @@ functionality in the future.
 The Controller-plugin should implement the following CSI procedures:
 
 - `CreateVolume` / `DeleteVolume`
-  1. Create RBD-image on the Ceph cluster (see Considerations)
-  1. Create the NVMe-subsystem (see Considerations)
-  1. Expose the RBD-image through the NVMe-oF Gateway
+   1. Create RBD-image on the Ceph cluster through Ceph-CSI/RBD
+   1. Create the NVMe-subsystem (if it does not exist yet)
+   1. Expose the RBD-image through the NVMe-oF Gateway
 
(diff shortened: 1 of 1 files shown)


Rakshith-R (approved):

docs/design/proposals/nvme-of.md
```diff
@@ -80,8 +80,10 @@ Administrator settable parameters, this extends the parameters that can be set
 for RBD volumes (set in a StorageClass):
 
 - _all parameters for the Ceph-CSI RBD driver_
-- NVMe-oF Gateway (or gateway group?)
-- NVMe subsystem (or possibly detected based on Kubernetes Namespace?)
+- NVMe-oF Gateway Group (hostname, could be a Kubernetes Service)
+- NVMe subsystem name
+   - if given, the name for the subsystem should be used
+   - if not given, base the name on the owner (Kubernetes Namespace)
```
Madhu-1, line 86: can we make this as mandatory in the SC so that we dont need to worry about it in other CO? or add a comment in other CO its a mandatory field

nixpanic: I think it can be optional in a StorageClass, as we can detect the owner in Kubernetes clusters. For other platforms (Docker Swarm, Hashicorp Nomad, ..) it should be mandatory. I think it is sufficiently clear in the document?

Madhu-1: That last part that its mandatory for other CO was not mentioned, as this also need more PR for open questions, it can be addressed later

Madhu-1 (approved):

Madhu-1:
@mergifyio queue

Issue #5370, opened 2025-06-18
Add support for NVMe-oF volumes backed by RBD


nixpanic:
# Describe the feature you'd like to have #

It would be great to have a Ceph-CSI provisioner for creating RBD volumes and have them accessible over NVMe-oF.

# What is the value to the end user? (why is it a priority?) #

Users that can not use native RBD-clients (like MS Windows hosts) can still use Ceph RBD.

# How will we know we have a good solution? (acceptance criteria) #

- [ ] Ceph-CSI should be able to create and delete RBD-images that are accessible over NVMe-oF
- [ ] simple deployment files under `deploy/` are added
- [x] e2e testing should deploy the Ceph nvme-gateway and run e2e (#5614)
- [ ] integrate deploying the NVMe-oF controller- and node-plugin with ceph-csi-operator (ceph/ceph-csi-operator#375)

# Additional context #

- [ ] Rook needs to deploy the NVMe-oF gateway rook/rook#16020
- [x] There needs to be a tool or API included in the Ceph base container image for configuring ([doc](https://docs.ceph.com/en/latest/rbd/nvmeof-target-configure/))

gadididi:
the first nvmeof-csi PR : https://github.com/ceph/ceph-csi/pull/5452 
provides the provisioner. 
there are some known issue \ open tasks that **must** be fixed in the next PR:

- [x]   Change the create\delete order: **create** subsystem -> listener -> ns -> add host (opposite direction for deletion)
- [x]   If some resource exists, don’t raise failure (need to discuss which resources)
- [x] Add more listeners to sc (temporary solution until the gateway does it automatically)
- [ ]  Add more GW's IP to sc 
- [ ]  ~Check ceph manager to get IPs list?~ the gateway will handle listeners in the future
- [x]   Create util-nvmeof package and move util function to `nvmeof_util.go` file
- [ ]   Makes create\Delete calls fullly idempotent 
- [x] cleanup resources when CreateVolume fails (#5538)
- [x]   Add ns UUID into `NVMeoFVolumeData` , for NodeStage
- [x]   Add support for RADOS namespace ( pool/**namespace**/rbd_image_name) (ceph/ceph-nvmeof#1458)
- [x] remove the generated gRPC/protobuf files (once ceph/ceph-nvmeof#1449 is merged)
[...]

PR #5452, opened 2025-07-27
nvmeof: add nvmeof provisioner ability

gadididi:

This PR introduces **NVMe-oF CSI driver support** for Ceph-CSI, enabling RBD volumes to be exposed through NVMe-oF protocol instead of traditional kernel RBD mapping.

## Components Added:

- **NVMe-oF Controller Server**: Complete wrapper around RBD CSI that orchestrates NVMe-oF resource creation and cleanup
- **NVMe-oF Gateway Client**: Full gRPC client implementation for managing NVMe-oF subsystems, namespaces, listeners, and hosts
- **Clean Domain Architecture**: Separate structs for CSI operations vs protobuf wire protocol with proper abstraction layers
- **Comprehensive Metadata Management**: Stores all NVMe-oF metadata (subsystem NQN, namespace ID, host NQN, listener details, gateway addresses) in RBD image metadata for reliable cleanup
- **Integration Tests**: Docker-based tests for validating NVMe-oF gateway operations with configurable environment

### **CreateVolume**:
1. **Validate request** with comprehensive parameter validation  
2. **Create RBD volume** through backend CSI driver
3. **Setup NVMe-oF resources**:
   - Create/ensure subsystem exists
   - Create namespace in subsystem
   - Create listeners on specified addresses
   - Add host to subsystem
4. **Store metadata** in both VolumeContext (for NodeServer) and RBD metadata (for cleanup)

### **DeleteVolume**:
1. **Retrieve stored metadata** from RBD image
2. **Cleanup NVMe-oF resources** in proper order:
   - Remove host from subsystem
   - Delete namespace
   - Delete listeners
   - Delete subsystem (if empty)
3. **Delete RBD volume**

### ⚠️ **Prototype Status - Controller Operations Only** ⚠️

This is a **working prototype** implementing the **controller-side operations**. The purpose of this PR is to:
- Demonstrate complete NVMe-oF CSI controller integration
- Validate the clean architecture approach with proper separation of concerns
- Get feedback on metadata management and cleanup strategies

### What's Working:
- ✅ **Complete CreateVolume/DeleteVolume** operations with full NVMe-oF setup
- ✅ **Production-ready NVMe-oF Gateway integration** (subsystem, namespace, listener, host management)
- ✅ **Robust metadata storage/retrieval** with dual storage (VolumeContext + RBD metadata)
- ✅ **Clean separation of management vs data plane** (gateway addresses vs listener addresses)
- ✅ **Integration tests** with configurable environment variables
- ✅ **Proper resource lifecycle management** with idempotent operations

### What's Missing/TODO:
- ❌ **NodeServer implementation** (NodeStageVolume, NodePublishVolume, NodeUnpublishVolume)
- ❌ **NVMe-oF initiator operations** (nvme connect/disconnect on nodes)
- ❌ **Multi-gateway support (HA)**
- ❌ **Multi-listeners support**
- ❌ **MTLS Grpc**
- ❌ **inband-authentication **
- ❌ **Advanced features** (volume expansion?, snapshots?, cloning?)

### Key Design Decisions Needing Review:
- **Architecture Pattern**: Controller-only setup vs full CSI implementation
- **Metadata Strategy**: Dual storage (VolumeContext + RBD metadata) for reliability

### Configuration Model:
StorageClass parameters support flexible deployment scenarios:
```yaml
parameters:
  subsystemNQN: "nqn.2016-06.io.ceph:subsystem.production"
  hostNQN: "nqn.2016-06.io.ceph:host.k8s-node1"
  nvmeofGatewayAddress: "10.242.64.100"    # Management network
  nvmeofGatewayPort: "5500"                # gRPC port
  listenerIpAddress: "10.242.64.32"       # Data network  
  listenerPort: "4420"                    # NVMe-oF port
  listenerHostname: "ceph-nvmeof-gw1"
```

## Related issues\PR\discussions  ##

1. https://github.com/ceph/ceph-csi/issues/5370
2. https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/nvme-of.md

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>


internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,275 @@
+package controller
```
nixpanic, line 17: Note that new files should have the copyright header.

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,356 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/protobuf/proto"
+)
+
+// Config holds gateway client configuration.
+type GatewayConfig struct {
+	Address        string
+	Port           string
+	ConnectTimeout int // seconds
```
nixpanic: Make attributes as accurate as possible. `Port` would probably be an `int`, and `ConnectTimeout` a `time.Duration`.

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,356 @@
...
+		return fmt.Errorf("failed to connect to Gateway gRPC server: %w", err)
+	}
+
+	c.conn = conn
+	c.client = NewGatewayClient(conn)
+
+	return nil
+}
+
+// TODO- maybe move to util-nvmeof package?
```
nixpanic, line 408: It is definitely fine to have a `internal/nvmeof/util` package if there are multiple shared functions. It makes it cleaner for unit-testing as well.

internal/nvmeof/tests/Dockerfile.nvmeof-test
```diff
@@ -0,0 +1,45 @@
+# Dockerfile.nvmeof-test
```
nixpanic, line 1: Complex testing like this should be in a separate commit (of the same PR). Simple unit-tests for utility functions can be in the same commit as where the function is introduced. Splitting large commits into smaller chunks makes it easier to review.

gadididi: fixed

internal/nvmeof/nvmeof.go
gadididi: is it ok I will do it in the next PR? I will create `internal/nvmeof/util` folder ?

internal/nvmeof/nvmeof.go
gadididi: I removed the `ConnectTimeout` var, because the first implementation of rpc client I used deprecated code:

```golang
// Connect to Gateway gRPC server.
func (c *GatewayRpcClient) connect() error {
    target := fmt.Sprintf("%s:%s", c.config.Address, c.config.Port)

    ctx, cancel := context.WithTimeout(context.Background(), time.Duration(c.GatewayConfig.ConnectTimeout))
    defer cancel()

    conn, err := grpc.DialContext(ctx, target, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
    if err != nil {
        return fmt.Errorf("failed to connect to Gateway gRPC server: %w", err)
    }

    c.conn = conn
    c.client = NewGatewayClient(conn)

    return nil
}
``` 

the `grpc.DialContext()` shows me warning: 
_grpc.DialContext is deprecated: use NewClient instead.  Will be supported throughout 1.x._
so I changed it to :
```golang
	// Create connection using new gRPC API
	conn, err := grpc.NewClient(target,
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
```
I will change the Port to be  `uint32`

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,526 @@
...
+	if err != nil {
+		log.ErrorLog(ctx, "failed to create RBD volume: %v", err)
+
+		return nil, err
+	}
+
+	backend := res.GetVolume()
+	volumeID := backend.GetVolumeId()
+	rbdImageName := res.GetVolume().GetVolumeContext()["imageName"]
+	rbdPoolName := res.GetVolume().GetVolumeContext()["pool"]
```
nixpanic, line 94: In addition to rbd-image name and the rados pool, there can be a rados namespace as well. That should probably be included in the `createNVMeoFResources()` function too.

internal/nvmeof/nvmeof.go
nixpanic: sure, no problem at all

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,397 @@
...
+	GatewayAddress
+}
+
+// GatewayClient wraps the gRPC client and connection details.
+type GatewayRpcClient struct {
+	config *GatewayConfig
+	conn   *grpc.ClientConn
+	client GatewayClient // protobuf client
+
+	// Thread-safe serial generation
```
nixpanic: Not sure this is needed, why does it need to be thread safe? If the same serial is used twice while communicating to the gateway, would one of the two not just fail? (and then retried by the container platform)

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+// Close closes the gateway connection.
+func (c *GatewayRpcClient) Close() error {
+	if c.conn != nil {
+		return c.conn.Close()
+	}
+
+	// Clear config data
+	c.config = nil
```
nixpanic, line 105: This is only reached a 2nd time the function is called. It should be set to `nil` immediately after closing.

My preference would be to check, return early if nothing needs to be done:
```go
conn := c.conn // get a reference so that concurrent calls don't panic

if conn == nil {
    return
}

c.conn = nil
conn.Close()

File: internal/nvmeof/controller/controllerserver.go [hunk 8]
```diff
@@ -0,0 +1,526 @@
...
+
+	// Get RBD volume
+	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find volume with ID %q: %w", volumeID, err)
+	}
+	defer rbdVol.Destroy(ctx)
+
+	// Retrieve all metadata including gateway management info
+	metadataKeys := map[string]*string{
```
nixpanic: Is there a benefit of using `*string` here? It isn't very common to use pointers to a string in Go, it makes the code a little more complex.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,526 @@
...
+func (cs *Server) validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
+	// Validate required parameters
+	params := req.GetParameters()
+	requiredParams := []string{
+		"subsystemNQN", "hostNQN", "nvmeofGatewayAddress", "nvmeofGatewayPort",
+		"listenerIpAddress", "listenerPort", "listenerHostname",
+	}
+	for _, param := range requiredParams {
+		if params[param] == "" {
+			return status.Error(codes.InvalidArgument, "missing required parameter: "+param)
```
nixpanic: Maybe don't return a gRPC error, only return gRPC errors in the top `CreateVolume` and `DeleteVolume` functions?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,526 @@
...
+	return nvmeofData, nil
+}
+
+// getGatewayConfigFromRequest extracts gateway configuration from request parameters.
+func (cs *Server) getGatewayConfigFromRequest(req *csi.CreateVolumeRequest) (*nvmeof.GatewayConfig, error) {
+	params := req.GetParameters()
+
+	address := params["nvmeofGatewayAddress"]
+	if address == "" {
+		return nil, status.Error(codes.InvalidArgument, "nvmeofGatewayAddress parameter is required")
```
nixpanic: No need for a gRPC error here either, I think.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,526 @@
...
+			Address: *metadataKeys["nvmeof.GatewayAddress"],
+			Port:    *metadataKeys["nvmeof.GatewayPort"],
+		},
+	}
+
+	return nvmeofData, nil
+}
+
+// getGatewayConfigFromRequest extracts gateway configuration from request parameters.
+func (cs *Server) getGatewayConfigFromRequest(req *csi.CreateVolumeRequest) (*nvmeof.GatewayConfig, error) {
```
nixpanic: Functions that do not use `cs *Server` should not be a function of the struct either, just make it a simple stand-alone helper function.

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,397 @@
...
+type NVMeoFVolumeData struct {
+	SubsystemNQN          string
+	NamespaceID           uint32
+	HostNQN               string
+	ListenerInfo          ListenerDetails
+	GatewayManagementInfo GatewayAddress
+}
+
+// Config holds gateway client configuration.
+type GatewayConfig struct {
```
nixpanic: Why not use `GatewayAddress` where you now have `GatewayConfig` structs? Just a single member in the struct makes it look a little too much indirect.

internal/nvmeof/tests/Dockerfile.nvmeof-test
```diff
@@ -0,0 +1,45 @@
...
+    netcat-openbsd
+
+# Set working directory
+WORKDIR /app
+
+# Copy all source code
+COPY . .
+
+# Download dependencies
+RUN go mod download
```
nixpanic: not sure this is needed, all dependencies are expected to be in the `vendor/` directory already

internal/nvmeof/nvmeof.go
gadididi: I wanted to distinguish between listener (IP + port + listener hostname )and gw config (IP + port) because each of them has different purpose, so I will do like this:
```Go

type GatewayAddress struct {
    Address string 
    Port    uint32 
}

type ListenerDetails struct {
    GatewayAddress
    Hostname string 
}

type GatewayConfig = GatewayAddress
```

internal/nvmeof/controller/controllerserver.go
gadididi: @nixpanic , In proto file there is no field RADOS namespace:
```Go

type NamespaceAddReq struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	RbdPoolName       string  `protobuf:"bytes,1,opt,name=rbd_pool_name,json=rbdPoolName,proto3" json:"rbd_pool_name,omitempty"`
	RbdImageName      string  `protobuf:"bytes,2,opt,name=rbd_image_name,json=rbdImageName,proto3" json:"rbd_image_name,omitempty"`
	SubsystemNqn      string  `protobuf:"bytes,3,opt,name=subsystem_nqn,json=subsystemNqn,proto3" json:"subsystem_nqn,omitempty"`
	Nsid              *uint32 `protobuf:"varint,4,opt,name=nsid,proto3,oneof" json:"nsid,omitempty"`
	BlockSize         uint32  `protobuf:"varint,5,opt,name=block_size,json=blockSize,proto3" json:"block_size,omitempty"`
	Uuid              *string `protobuf:"bytes,6,opt,name=uuid,proto3,oneof" json:"uuid,omitempty"`
	Anagrpid          *int32  `protobuf:"varint,7,opt,name=anagrpid,proto3,oneof" json:"anagrpid,omitempty"`
	CreateImage       *bool   `protobuf:"varint,8,opt,name=create_image,json=createImage,proto3,oneof" json:"create_image,omitempty"`
	Size              *uint64 `protobuf:"varint,9,opt,name=size,proto3,oneof" json:"size,omitempty"`
	Force             *bool   `protobuf:"varint,10,opt,name=force,proto3,oneof" json:"force,omitempty"`
	NoAutoVisible     *bool   `protobuf:"varint,11,opt,name=no_auto_visible,json=noAutoVisible,proto3,oneof" json:"no_auto_visible,omitempty"`
	TrashImage        *bool   `protobuf:"varint,12,opt,name=trash_image,json=trashImage,proto3,oneof" json:"trash_image,omitempty"`
	DisableAutoResize *bool   `protobuf:"varint,13,opt,name=disable_auto_resize,json=disableAutoResize,proto3,oneof" json:"disable_auto_resize,omitempty"`
	ReadOnly          *bool   `protobuf:"varint,14,opt,name=read_only,json=readOnly,proto3,oneof" json:"read_only,omitempty"`
}
```

do I need to concatenate it like:

Without namespace
`rbd create pool/image-name --size 1G`

With namespace  
`rbd create pool/namespace/image-name --size 1G`

need to check if nvmeof repo support that, if does not, I will open an issue

internal/nvmeof/nvmeof.go
gadididi: right, my mistake , i think maybe also set to nil the client? like this:

```Go
// Close closes the gateway connection.
func (c *GatewayRpcClient) Close() error {
	if c == nil {
		return nil
	}
	var err error
	// Close connection if it exists
	if c.conn != nil {
		err = c.conn.Close()
		c.conn = nil
	}
	// Clear client reference
	c.client = nil
	// Clear config to prevent reuse
	c.config = nil
	return err
}
```
what do you think?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -260,7 +266,7 @@ func (cs *Server) createNVMeoFResources(
 
 	// Step 5: Create listeners
 	err = gateway.CreateListener(ctx, nvmeofData.SubsystemNQN, nvmeofData.ListenerInfo.Address,
```
nixpanic: Maybe reduce the arguments for this function, more like this:
```go
err = gateway.CreateListener(ctx, nvmeofData.SubsystemNQN, nvmeofData.ListenerInfo)
```
All other parameters are part of `nvmeofData.ListenerInfo`, passing that struct looks simpler.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -330,19 +332,21 @@ func (cs *Server) cleanupNVMeoFResources(
...
 	}
+	listenerPortStr := strconv.FormatUint(uint64(data.ListenerInfo.Port), 10)
+	GatewayManagementInfoPortStr := strconv.FormatUint(uint64(data.GatewayManagementInfo.Port), 10)
 
 	volume.VolumeContext["nvmeof.SubsystemNQN"] = data.SubsystemNQN
 	volume.VolumeContext["nvmeof.NamespaceID"] = strconv.FormatUint(uint64(data.NamespaceID), 10)
 	volume.VolumeContext["nvmeof.HostNQN"] = data.HostNQN
 	volume.VolumeContext["nvmeof.ListenerIpAddress"] = data.ListenerInfo.Address
-	volume.VolumeContext["nvmeof.ListenerPort"] = data.ListenerInfo.Port
+	volume.VolumeContext["nvmeof.ListenerPort"] = listenerPortStr
```
nixpanic: Instead of `strconv.FormatUint`, you could also do `fmt.Sprintf("%d", data.ListenerInfo.Port)` here and below.

internal/nvmeof/nvmeof.go
nixpanic: Yes, that is probably good to do. Many structs in Ceph-CSI have a `Destroy()` call that free's up resources. As this isn't a single `close()` anymore, maybe rename the function?

internal/nvmeof/nvmeof.go
```diff
@@ -44,36 +43,37 @@ type ListenerDetails struct {
...
+	config    *GatewayConfig
+	conn      *grpc.ClientConn
+	client    GatewayClient // protobuf client
+	maxSerial *big.Int      // Maximum allowed serial number for subsystems.
 }
 
+// TODO: Move this to a util package?
+const (
+	MaxAllowedSubsystemSerialNumber = 99999999999999
+	SerialOffset                    = 2 // prevent 0 and 1 as serial numbers
```
nixpanic: Should this offset be stored somewhere? What happens if the controller restarts, and the offset is reset back to it's initial value?

internal/nvmeof/nvmeof.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/tests/Dockerfile.nvmeof-test
gadididi: I removed that line

internal/nvmeof/nvmeof.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: fixed (also at `DeleteListener`)

internal/nvmeof/controller/controllerserver.go
gadididi: why?
I took it from here:
https://stackoverflow.com/questions/10105935/how-to-convert-an-int-value-to-string-in-go

I read that `strconv.FormatUint` is better than `fmt.Sprintf`

internal/nvmeof/nvmeof.go
gadididi: as we talked,
it's actually a constant offset for random number generation, not a counter , I will change the name to be more understandable

internal/nvmeof/controller/controllerserver.go
gadididi: @nixpanic , opened an issue:
https://github.com/ceph/ceph-nvmeof/issues/1458

internal/nvmeof/nvmeof.go
```diff
@@ -157,12 +157,19 @@ func (gw *GatewayRpcClient) DeleteNamespace(ctx context.Context, subsystemNQN st
...
 	}
-	if status.GetStatus() != 0 {
-		return fmt.Errorf("gateway NamespaceDelete returned error: %s", status.GetErrorMessage())
+	if status.GetStatus() == 0 {
+		log.DebugLog(ctx, "Namespace deleted successfully: %d", namespaceID)
+
+		return nil
 	}
-	log.DebugLog(ctx, "Namespace deleted successfully: %d", namespaceID)
+	if isNotFoundError(status.GetErrorMessage()) && status.GetStatus() == 2 { // ENOENT
```
nixpanic: Just checking for `status.GetStatus() == 2` or even better, `status.GetStatus() == syscall.ENOENT` is probably sufficient? Log the `status.GetErrorMessage()` for more details would then be useful.

Comparing error messages is always a little ugly, there is no guarantee the message stays the same (or isn't translated into an other language).

gadididi: @nixpanic ,I will change to `syscall.ENOENT` , about the strings compare,  I know, I dont know what is the best solution for it.
The best thing I can think is check if the subsystem exists (there is grcp call for it) if not , so 100% there are no ns, host and listener, and no need to call delete for those, what do you think?

nixpanic: Ideally you drop the `isNotFoundError()` function and rely on the `ENOENT` status.

Deletion of the subsystem will not happen often, it is more likely that a volume is deleted, and other volumes are still available. There is no need to enhance performance for listener/subsystem deletion.

internal/nvmeof/controller/controllerserver.go
nixpanic: It looks like there is a performance benefit with `strconv.FormatUint`, which is nice. It isn't really a critical path here, so I don't think anyone will notice. `fmt.Sprintf()` looks so much simpler and shorter, it has my preference still.

The way you have it is fine, no need to change it. Thanks for pointing out the difference.

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,428 @@
...
+			return nil, fmt.Errorf("failed to connect to gateway: %w", err)
+		}
+	}
+
+	return client, nil
+}
+
+// Close closes the gateway connection.
+func (c *GatewayRpcClient) Destroy() error {
+	if c == nil {
```
nixpanic: I do not think this can happen. If something calls `nil.Destroy()` there is a failure before we get here.

```go
var client *GatewayRpcClient
client = nil
client.Destroy() // panic!
```

nixpanic (commented):
No major changes needed from my point of view. There are quite a few TODO's listed, those are all improvements we can work out later.

For an initial proof-of-concept this looks good to me. With your additional work-in-progress branch(es) I have been able to get this up and running :partying_face: 

```
$ kubectl describe pvc
Name:          nvmeof-test-pvc
Namespace:     default
StorageClass:  ocs-storagecluster-ceph-nvmeof
Status:        Bound
Volume:        pvc-3a0994d7-6457-4966-977e-dc756b871af7
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: nvmeof.csi.ceph.com
               volume.kubernetes.io/storage-provisioner: nvmeof.csi.ceph.com
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      1Gi
Access Modes:  RWO
VolumeMode:    Filesystem
Used By:       <none>
Events:
  Type    Reason                 Age   From                                                                                                    Message
  ----    ------                 ----  ----                                                                                                    -------
  Normal  Provisioning           9s    nvmeof.csi.ceph.com_csi-nvmeofplugin-provisioner-745f6d9947-f56rd_3479ebb6-df20-477d-a7bb-737129b80352  External provisioner is provisioning volume for claim "default/nvmeof-test-pvc"
  Normal  ExternalProvisioning   9s    persistentvolume-controller                                                                             Waiting for a volume to be created either by the external provisioner 'nvmeof.csi.ceph.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
  Normal  ProvisioningSucceeded  8s    nvmeof.csi.ceph.com_csi-nvmeofplugin-provisioner-745f6d9947-f56rd_3479ebb6-df20-477d-a7bb-737129b80352  Successfully provisioned volume pvc-3a0994d7-6457-4966-977e-dc756b871af7
```

internal/nvmeof/nvmeof.go
nixpanic: Also, the comment is still about `Close()`, but the function was renamed to `Destroy()`.

yati1998:
I went through the code, looks good to me from the demo we had.  The missing parts already have TODO tag in it, so we are good to have them in upcoming PR.

internal/nvmeof/nvmeof.go
nixpanic: I've fixed this up for you :smiley:

nixpanic (dismissed):

nixpanic:
No shared code was modified, and we don't have e2e testing yet, so skip them.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+		}
+	}
+	// Delete RBD volume through backend
+
+	return cs.backendServer.DeleteVolume(ctx, req)
+}
+
+// validateCreateVolumeRequest validates the incoming request for nvmeof.
+// the rest of the parameters are validated by RBD.
+func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
```
Madhu-1, line 144: does this supports all the parameters present in the RBD storageclass as well? If not we need to add a validation to block it. Like erasure coding. rbd namespace etc

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+	rbdImageName := res.GetVolume().GetVolumeContext()["imageName"]
+	rbdPoolName := res.GetVolume().GetVolumeContext()["pool"]
+
+	log.DebugLog(ctx, "RBD volume created successfully: pool=%s, image=%s, id=%s",
+		rbdPoolName, rbdImageName, volumeID)
+
+	// Step 2: Setup NVMe-oF resources
+	nvmeofData, err := createNVMeoFResources(ctx, req, rbdPoolName, rbdImageName)
+	if err != nil {
+		log.ErrorLog(ctx, "NVMe-oF resource setup failed for volume %s: %v", volumeID, err)
```
Madhu-1: `failed for volumeID` instead of `failed for volume`

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+
+		return nil, err
+	}
+
+	backend := res.GetVolume()
+	volumeID := backend.GetVolumeId()
+	rbdImageName := res.GetVolume().GetVolumeContext()["imageName"]
+	rbdPoolName := res.GetVolume().GetVolumeContext()["pool"]
+
+	log.DebugLog(ctx, "RBD volume created successfully: pool=%s, image=%s, id=%s",
```
Madhu-1: This is already present in the `CreateVolume` log message. this seems to be redundant can you please validate this?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+		// TODO: Implement cleanup of RBD volume on NVMe-oF failure
+
+		return nil, status.Errorf(codes.Internal, "NVMe-oF setup failed: %v", err)
+	}
+
+	// step 3: Populate volume context for NodeServer
+	populateVolumeContext(backend, nvmeofData)
+
+	// Step 4: Store NVMe-oF metadata in the volume context
+	if err := cs.storeNVMeoFMetadata(ctx, req, volumeID, nvmeofData); err != nil {
```
Madhu-1, line 109: TODO to implement cleanup in case of failure?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+		return nil, err // Error already formatted with proper status code
+	}
+
+	return &csi.CreateVolumeResponse{Volume: backend}, nil
+}
+
+func (cs *Server) DeleteVolume(
```
Madhu-1, line 117: Please add a comment for exported function

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+	if err != nil {
+		return status.Errorf(codes.Aborted, "failed to find volume with ID %q: %v", volumeID, err)
+	}
+	defer rbdVol.Destroy(ctx)
+
+	GatewayManagementInfoPortStr := strconv.FormatUint(uint64(nvmeofData.GatewayManagementInfo.Port), 10)
+	ListenerInfoPortStr := strconv.FormatUint(uint64(nvmeofData.ListenerInfo.Port), 10)
+
+	// Prepare all metadata entries
+	metadata := map[string]string{
```
Madhu-1, line 405: All the metadata keys need to be start with `.rbd` so that in future we dont need to have the metadata copied to remote cluster if we support cloning as well?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+
+	// Get RBD volume
+	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
+	if err != nil {
+		return status.Errorf(codes.Aborted, "failed to find volume with ID %q: %v", volumeID, err)
+	}
+	defer rbdVol.Destroy(ctx)
+
+	GatewayManagementInfoPortStr := strconv.FormatUint(uint64(nvmeofData.GatewayManagementInfo.Port), 10)
+	ListenerInfoPortStr := strconv.FormatUint(uint64(nvmeofData.ListenerInfo.Port), 10)
```
Madhu-1:
```suggestion
gatewayManagementInfoPortStr := strconv.FormatUint(uint64(nvmeofData.GatewayManagementInfo.Port), 10)
listenerInfoPortStr := strconv.FormatUint(uint64(nvmeofData.ListenerInfo.Port), 10)
```

lets not use upper case to variable names

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+	return nil
+}
+
+// populateVolumeContext adds NVMe-oF information to volume context for NodeServer.
+func populateVolumeContext(volume *csi.Volume, data *nvmeof.NVMeoFVolumeData) {
+	if volume.VolumeContext == nil {
+		volume.VolumeContext = make(map[string]string)
+	}
+	listenerPortStr := strconv.FormatUint(uint64(data.ListenerInfo.Port), 10)
+	GatewayManagementInfoPortStr := strconv.FormatUint(uint64(data.GatewayManagementInfo.Port), 10)
```
Madhu-1: `GatewayManagementInfoPortStr` to `gatewayManagementInfoPortStr`

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+		return nil, fmt.Errorf("failed to find volume with ID %q: %w", volumeID, err)
+	}
+	defer rbdVol.Destroy(ctx)
+
+	// Retrieve all metadata including gateway management info
+	metadata := make(map[string]string)
+
+	// Required metadata keys
+	requiredKeys := []string{
+		"nvmeof.SubsystemNQN",
```
Madhu-1: define a const for these keys as they are used in multiple places?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+		metadata[key] = value
+	}
+
+	// Parse namespace ID
+	nsid, err := strconv.ParseUint(metadata["nvmeof.NamespaceID"], 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("invalid namespace ID: %w", err)
+	}
+
+	ListenerInfoPort, err := strconv.ParseUint(metadata["nvmeof.ListenerPort"], 10, 32)
```
Madhu-1: `ListenerInfoPort` to `listenerInfoPort`

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid listener port: %w", err)
+	}
+	GatewayPort, err := strconv.ParseUint(metadata["nvmeof.GatewayPort"], 10, 32)
```
Madhu-1: `GatewayPort` to `gatewayPort`

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+		Port:    nvmeofData.GatewayManagementInfo.Port,
+	}
+
+	gateway, err := nvmeof.NewGatewayRpcClient(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to management gateway %s:%d: %w",
+			config.Address, config.Port, err)
+	}
+
+	log.DebugLog(ctx, "Connected to gateway for cleanup: %s:%s",
```
Madhu-1: use `%d` for port

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+	if volumeID == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "empty volume ID in request")
+	}
+	// Get NVMe-oF metadata for cleanup
+	nvmeofData, err := cs.getNVMeoFMetadata(ctx, req, volumeID)
+	if err != nil {
+		log.DebugLog(ctx, "No NVMe-oF metadata found, skipping NVMe-oF cleanup: %v", err)
+	} else {
+		// Clean up NVMe-oF resources
+		if err := cleanupNVMeoFResources(ctx, req, nvmeofData); err != nil {
```
Madhu-1, line 131: cleanup is not required even in case partial key is missing in metadata? i assume it wont happen but checking for clarity. shouldnt we just cleanup the resouces if required details are found in metadata?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+	nvmeofData *nvmeof.NVMeoFVolumeData,
+) error {
+	// Create RBD manager
+	mgr := rbdutil.NewManager(cs.backendServer.Driver.GetInstanceID(), nil, req.GetSecrets())
+	defer mgr.Destroy(ctx)
+
+	// Get RBD volume
+	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
+	if err != nil {
+		return status.Errorf(codes.Aborted, "failed to find volume with ID %q: %v", volumeID, err)
```
Madhu-1, line 397: i think `Aborted` is not the right error message here. it could be not found

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+
+	return cs.backendServer.DeleteVolume(ctx, req)
+}
+
+// validateCreateVolumeRequest validates the incoming request for nvmeof.
+// the rest of the parameters are validated by RBD.
+func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
+	// Validate required parameters
+	params := req.GetParameters()
+	requiredParams := []string{
```
Madhu-1, line 147: define const as they are used in multiple places

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+// TODO - need to support multiple listeners.
+// TODO - need to fallback cleanup if any step fails.
+func createNVMeoFResources(
+	ctx context.Context,
+	req *csi.CreateVolumeRequest,
+	rbdPoolName,
+	rbdImageName string,
+) (*nvmeof.NVMeoFVolumeData, error) {
+	// Step 1: Connect to gateway
+	gateway, err := connectGateway(req)
```
Madhu-1: Move this to below as we dont need to connect if there is any issue with input. and also input validation need to be done at at beginning even before creating the rbd image?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+		"nvmeof.GatewayPort":    GatewayManagementInfoPortStr,
+	}
+
+	// Store all metadata entries
+	for key, value := range metadata {
+		if value == "" {
+			log.WarningLog(ctx, "Skipping empty metadata value for key: %s", key)
+
+			continue
+		}
```
Madhu-1, line 427: Can this happen where there can be keys not no values for these?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+			Address: metadata["nvmeof.GatewayAddress"],
+			Port:    uint32(GatewayPort),
+		},
+	}
+
+	return nvmeofData, nil
+}
+
+// getGatewayConfigFromRequest extracts gateway configuration from request parameters.
+func getGatewayConfigFromRequest(req *csi.CreateVolumeRequest) (*nvmeof.GatewayConfig, error) {
```
Madhu-1: pass only parameters not the complete request as input

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+	}
+
+	return &nvmeof.GatewayConfig{
+		Address: address,
+		Port:    uint32(port),
+	}, nil
+}
+
+// connectGateway creates and connects a gateway client for this operation.
+func connectGateway(req *csi.CreateVolumeRequest) (*nvmeof.GatewayRpcClient, error) {
```
Madhu-1: pass only parameters as input as we dont need the complete request?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,563 @@
...
+	gateway, err := nvmeof.NewGatewayRpcClient(config)
+	if err != nil {
+		return nil, fmt.Errorf("gateway connection failed: %w", err)
+	}
+
+	return gateway, nil
+}
+
+// connectGatewayForDelete creates gateway connection using stored management address.
+func connectGatewayForDelete(
```
Madhu-1: i see that we have `connectGateway` and `connectGatewayForDelete` can we have single generic connect that accepts the config as input and return the gateway client?

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,424 @@
...
+package nvmeof
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+	"math/big"
+	"syscall"
+
+	"github.com/ceph/ceph-csi/internal/util/log"
```
Madhu-1: Need to change the import order as this should be at last

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,424 @@
...
+
+// NewGatewayRpcClient creates a new gateway client.
+func NewGatewayRpcClient(config *GatewayConfig) (*GatewayRpcClient, error) {
+	client := &GatewayRpcClient{
+		config:    config,
+		maxSerial: big.NewInt(MaxAllowedSubsystemSerialNumber - RandomNumberOffset),
+	}
+
+	// Only connect if config is provided
+	if config != nil {
```
Madhu-1: what is the use of returning an unconnected client? in all other functions we need to check if the client is empty or not before accessing it. wont it be better to return an error if config is empty?

internal/nvmeof/gateway.pb.go
```diff
@@ -0,0 +1,6917 @@
+//
+//  Copyright (c) 2021 International Business Machines
```
Madhu-1: Couple of questions

1. Why we dont have proto file for this one and it looks odd to me that we are generating the pb file and keeping it here. usually the repo should provide library for these proto files.
2. There are no comments on the struct fields, it hard to understand what each field in request and response are. 
3. No makefile changes required to build this one?
4. How do we keep it updated always if we manually build it?

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,424 @@
...
+		// TrashImage:        nil,
+		// DisableAutoResize: nil,
+		// ReadOnly:          nil,
+	}
+
+	resp, err := gw.client.NamespaceAdd(ctx, req)
+	if err != nil {
+		return 0, fmt.Errorf("failed to create namespace for %s/%s: %w", poolName, imageName, err)
+	}
+	if resp.GetStatus() != 0 {
```
Madhu-1, line 137: This looks odd to me, where we didnt get the error from GRPC server rather we dont nil response and actual response structure contains the response code and message. where i can read about these status? i didnt find any details in the above .pb.go file

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,424 @@
...
+
+	req := &DeleteSubsystemReq{
+		SubsystemNqn: subsystemNQN,
+	}
+
+	status, err := gw.client.DeleteSubsystem(ctx, req)
+	if err != nil {
+		return fmt.Errorf("failed to delete subsystem %s: %w", subsystemNQN, err)
+	}
+	if status.GetStatus() != 0 {
```
Madhu-1: dont we need not found error check here?

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,424 @@
...
+	log.DebugLog(ctx, "Checking if subsystem %s exists on gateway %s", subsystemNQN, gw.config)
+
+	req := &ListSubsystemsReq{
+		SubsystemNqn: nil,
+	}
+	resp, err := gw.client.ListSubsystems(ctx, req)
+	if err != nil {
+		return false, fmt.Errorf("failed to list subsystems: %w", err)
+	}
+	if resp.GetStatus() != 0 {
```
Madhu-1, line 240: dont we need to return err unconditionally if the GetStatus()!=0 from here?

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,424 @@
...
+	status, err := gw.client.NamespaceDelete(ctx, req)
+	if err != nil {
+		return fmt.Errorf("failed to delete namespace %d: %w", namespaceID, err)
+	}
+	if status.GetStatus() == 0 {
+		log.DebugLog(ctx, "Namespace deleted successfully: %d", namespaceID)
+
+		return nil
+	}
+	if status.GetStatus() == int32(syscall.ENOENT) { // ENOENT
```
Madhu-1, line 163: Wont the GRPC server is idempotent and return success if the namespace doesnt exists? This is question because i see that for Create there is no Already exists error check but we have only for delete.

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,424 @@
...
+		return nil
+	}
+	errorMsg := resp.GetErrorMessage()
+	if resp.GetStatus() == int32(syscall.ENOENT) { // ENOENT
+		log.DebugLog(ctx, "Listener %s already deleted from subsystem %s (not found)", listenerInfo.Address, subsystemNQN)
+
+		return nil // Listener already deleted, no error
+	}
+
+	return fmt.Errorf("gateway DeleteListener returned error (status=%d): %s", resp.GetStatus(), errorMsg)
```
Madhu-1: return `resp.GetErrorMessage()` no need for a new variable

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,424 @@
...
+		return nil
+	}
+	errorMsg := resp.GetErrorMessage()
+	if resp.GetStatus() == int32(syscall.ENOENT) { // ENOENT
+		log.DebugLog(ctx, "Host %s already removed from subsystem %s (not found)", hostNQN, subsystemNQN)
+
+		return nil
+	}
+
+	return fmt.Errorf("gateway RemoveHost returned error (status=%d): %s", resp.GetStatus(), errorMsg)
```
Madhu-1: return resp.GetErrorMessage() no need for a new variable

Madhu-1 (changes requested):
Thanks for all the work 👍🏻 The PR is in good shape.

internal/nvmeof/controller/controllerserver.go
nixpanic: Good point. Some parameters are not possible yet (like RADOS namespace), others are hopefully not required and can be made optional. These TODO's are mentioned in https://github.com/ceph/ceph-csi/issues/5370#issuecomment-3163939133 where we will follow-up on shortly too.

internal/nvmeof/gateway.pb.go
nixpanic: These files will be removed once ceph/ceph-nvmeof#1449 is merged.

internal/nvmeof/nvmeof.go
nixpanic: The [`gateway.proto`](https://github.com/ceph/ceph-nvmeof/blob/devel/control/proto/gateway.proto) is part of the **ceph-nvmeof** repository. Unfortunately, there is little documentation about it.

internal/nvmeof/controller/controllerserver.go
nixpanic: Indeed, will remove it.

internal/nvmeof/controller/controllerserver.go
nixpanic: yes, is part of the idempotent TODO mentioned in #5370.

internal/nvmeof/controller/controllerserver.go
nixpanic: Yes, indeed. These parameters are probably different on a mirrored cluster.

internal/nvmeof/controller/controllerserver.go
nixpanic: Done.

internal/nvmeof/controller/controllerserver.go
nixpanic: Addressed with a `String()` function for printing the address and port in a simpler way.

internal/nvmeof/controller/controllerserver.go
nixpanic: There is not really a better error than `Aborted`. This is called from `CreateVolume` and `NotFound` indicates that the `volume_source` was not found (for snapshot/cloning).

`Aborted` seems like the most appropriate error to me. The chance that the RBD-image was created, and after it was not found anymore is pretty slim in any case.

internal/nvmeof/controller/controllerserver.go
nixpanic: yes, done.

internal/nvmeof/controller/controllerserver.go
nixpanic: swapped step 2 & 1, thanks for the suggestion

internal/nvmeof/controller/controllerserver.go
nixpanic: Done.

internal/nvmeof/controller/controllerserver.go
nixpanic: Done.

internal/nvmeof/nvmeof.go
nixpanic: yuck, I still don't like this awkwardness in Ceph-CSI guidelines. Changed it now though.

internal/nvmeof/controller/controllerserver.go
nixpanic: That would basically just be this call:
```go
gateway, err := nvmeof.NewGatewayRpcClient(config)
```

I am not sure how to make it simpler/cleaner. Can we improve this later on, if really needed?

internal/nvmeof/nvmeof.go
nixpanic: `config` is never `nil` in the current code, so dropped this check.

internal/nvmeof/nvmeof.go
nixpanic: Added now. Not all procedures are completely idempotent yet, we'll fix that in follow-up PRs too.

internal/nvmeof/nvmeof.go
nixpanic: No, the gateway returns errors for objects that exist already, and for non-existing objects that are being removed.

internal/nvmeof/nvmeof.go
nixpanic: yeah, simplified it now.

internal/nvmeof/controller/controllerserver.go
nixpanic: Not at the moment, but ideally some of these parameters become optional with future improvements.

internal/nvmeof/controller/controllerserver.go
Madhu-1: @nixpanic i dont see its addressed, These are still string, Is this also a TODO?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -0,0 +1,574 @@
...
+	}
+
+	if len(namespaces.GetNamespaces()) == 0 {
+		log.DebugLog(ctx, "Subsystem %s is empty, deleting", subsystemNQN)
+		err = gateway.DeleteSubsystem(ctx, subsystemNQN)
+		if err != nil {
+			return fmt.Errorf("failed to delete empty subsystem: %w", err)
+		}
+		log.DebugLog(ctx, "Empty subsystem %s deleted", subsystemNQN)
+	} else {
```
Madhu-1: This check can be done first `if len(namespaces.GetNamespaces()) != 0 return nil` so that we can avoid else statement

internal/nvmeof/controller/controllerserver.go
Madhu-1: Sounds good, its good to have single common function for connect

internal/nvmeof/nvmeof.go
Madhu-1: This is as per the documentation we have and you have updated this documentation, there is a mix of import in this PR, where i some pace internal packages are placed before third party and after in some places

internal/nvmeof/nvmeof.go
Madhu-1: Any plan to open a tracker to improve the documentation for consumers? or else it will become hard to understand the input and outputs

internal/nvmeof/nvmeof.go
```diff
@@ -0,0 +1,417 @@
...
+		Trsvcid:  &listenerInfo.Port,
+		Adrfam:   &adrfam, // Assuming IPv4, can be configurable // TODO - make it configurable
+		Force:    &force,
+	}
+
+	resp, err := gw.client.DeleteListener(ctx, req)
+	if err != nil {
+		return fmt.Errorf("failed to delete listener %s from subsystem %s: %w", listenerInfo.Address, subsystemNQN, err)
+	}
+	if resp.GetStatus() == 0 {
```
Madhu-1, line 331: it would be nice to return error early and avoid multiple if if we check `resp.GetStatus() != 0 && resp.GetStatus() != int32(syscall.ENOENT)` same is required in other Delete calls as we..

Madhu-1 (commented):
@nixpanic the Plan is to get this PR merged and later remove the pb files and integration test file when ceph PR is merged?

Madhu-1:
> @nixpanic the Plan is to get this PR merged and later remove the pb files and integration test file when ceph PR is merged?

Based on the response, will LGTM, no need to address the new review comments, if we are planning to merge it. it can be a followup PR

internal/nvmeof/nvmeof.go
gadididi: @Madhu-1 , Hi , according to this guide: https://github.com/ceph/ceph-csi/blob/devel/docs/coding.md
third party must be located in the middle, but in the entire project I see some files with third-party in the middle and some files in the end, what is correct? 
in our case, nvmeof.go and  controllerserver.go have third-party in the end, and nvmeof_test.go has in the middle.

Madhu-1: @gadididi it should be in middle as per documentation, we didnt get time to address this in current codebase, we need to follow the standard specified in the document for new files.

internal/nvmeof/nvmeof.go
gadididi: @Madhu-1 , I can do it , but i want a specific check to handle "no exist" case. 

what I can do is something like this:

```go

	resp, err := gw.client.DeleteListener(ctx, req)
	if err != nil {
		return fmt.Errorf("failed to delete listener %s from subsystem %s: %w", listenerInfo.Address, subsystemNQN, err)
	}

	// Early return for error cases (anything other than success or not-found)
	if resp.GetStatus() != 0 && resp.GetStatus() != int32(syscall.ENOENT) {
		return fmt.Errorf("gateway DeleteListener returned error (status=%d): %s", resp.GetStatus(), resp.GetErrorMessage())
	}

	// Not-found cases
    if resp.GetStatus() == int32(syscall.ENOENT) {
		log.DebugLog(ctx, "Listener %s already deleted from subsystem %s (not found)", listenerInfo.Address, subsystemNQN)
         return nil
	}
   // Handle success 
   log.DebugLog(ctx, "Listener deleted successfully: %s from subsystem %s", listenerInfo.Address, subsystemNQN)
	return nil

```

but maybe it makes it more complicated, what do you think?

internal/nvmeof/nvmeof.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/nvmeof.go
Madhu-1: i assume this is for logging, IMO we dont need to log if it is success or NotFound because its same for CSI where the deletion is successful. if really required we can log the status code at the end which helps for debugging. its not worth to add a if statement to log different error messages.

gadididi: I'd like to keep the distinction because logging "deleted successfully" when it was already gone feels inaccurate. The extra if helps with debugging to know if we actually deleted something vs. it was already missing.

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

gadididi:
Hi, @Madhu-1 , @nixpanic ,

I pushed 2 new commits about generated files. 

### Update: Using upstream protobuf files from ceph-nvmeof

I've updated the PR to use the generated protobuf files from upstream `ceph-nvmeof` instead of maintaining local copies in our repository.

### What changed:
- Added dependency on `github.com/ceph/ceph-nvmeof/lib/go/nvmeof` which now contains the generated `gateway.pb.go` and `gateway_grpc.pb.go` files
- Removed our locally generated protobuf files 
- Vendored the upstream files into `vendor/github.com/ceph/ceph-nvmeof/lib/go/nvmeof/`
- Updated imports to use the upstream package: `import pb "github.com/ceph/ceph-nvmeof/lib/go/nvmeof"`

### Commands I ran:
```bash
go get github.com/ceph/ceph-nvmeof@devel
go mod tidy
go mod vendor
```
**Note about Go version change:
The Go version in go.mod was automatically updated from 1.24.0 to 1.24.4 because the upstream ceph-nvmeof module requires Go 1.24.4**

Madhu-1 (dismissed):
LGTM, Lets get it merged to devel once we have 3.15.0 release

Madhu-1:
Added DNM to avoid accidental merge, please remove it once we have new release

Madhu-1:
Thank you @gadididi for all the work 🎉

Madhu-1:
i would suggest to drop https://github.com/ceph/ceph-csi/pull/5452/commits/24951fb0f7c5ab7f6dd8b2254c9aa9cf4ac941dc commit as its removing the proto file from internal folder and remove pb file from https://github.com/ceph/ceph-csi/pull/5452/commits/1c430a5ff895a9284f40cad7be994953ac5b788a this commit

gadididi:
thank you! 🙂
yes, Niels told me wait until will be  new release 
I will do rebase interactive and rearrange the commits (squash https://github.com/ceph/ceph-csi/commit/24951fb0f7c5ab7f6dd8b2254c9aa9cf4ac941dc and https://github.com/ceph/ceph-csi/commit/1c430a5ff895a9284f40cad7be994953ac5b788a)

nixpanic (approved):

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

`release-v3.15` has been branched, this can now get merged in the `devel` branch

PR #5488, opened 2025-08-12
rbd: adjust qos parameters start with small case

YiteGu:

we should change this to small to align this with other parameters.

---

<details>

</details>

### DIFF
--- a/docs/rbd/deploy.md
+++ b/docs/rbd/deploy.md
@@ -73,19 +73,19 @@ make image-cephcsi
 | `stripeUnit`                                                                                        | no                   | stripe unit in bytes                                                                                                                                                                                                                                                                               |
 | `stripeCount`                                                                                       | no                   | objects to stripe over before looping                                                                                                                                                                                                                                                              |
 | `objectSize`                                                                                        | no                   | object size in bytes                                                                                                                                                                                                                                                                               |
-| `BaseIops`                                                                                          | no                   | the base limit of operations per second |
-| `BaseReadIops`                                                                                      | no                   | the base limit of read operations per second |
-| `BaseWriteIops`                                                                                     | no                   | the base limit of write operations per second |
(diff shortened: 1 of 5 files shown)


iPraveenParihar:
@YiteGu, Can you please add a message in PR description why is this change needed and same in commit message?

YiteGu:
> we should change this to small to align this with other parameters.

updated. the purpose is very simple. change QOS params to small to align this with other parameters.

nixpanic:
This isn't backwards compatible, is that a problem? Existing StorageClasses with the old parameters won't keep functioning when Ceph-CSI is updated to use the new parameters.

Madhu-1:
> This isn't backwards compatible, is that a problem? Existing StorageClasses with the old parameters won't keep functioning when Ceph-CSI is updated to use the new parameters.



> This isn't backwards compatible, is that a problem? Existing StorageClasses with the old parameters won't keep functioning when Ceph-CSI is updated to use the new parameters.

@nixpanic ceph-csi never advertized nbd as supported mounter, its currently kept as alpha feature only https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/rbd-nbd.md#status, This change is required to keep all the parameter keys to be inconsistent format.

Madhu-1 (approved):
We can add a release notes to update these keys to switch to the new one.

nixpanic (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

YiteGu:
> We can add a release notes to update these keys to switch to the new one.

Can an existing PVC be switched to a new storageclass by modifying the storageClassName?

Madhu-1:
> > We can add a release notes to update these keys to switch to the new one.
> 
> Can an existing PVC be switched to a new storageclass by modifying the storageClassName?

its not possible as these attributes are stored in the PV object, best option is to implement VolumeAttributeClass for this one to change the existing options.

PR #5492, opened 2025-08-12
nvmeof: initial node-plugin functionality

gadididi:

This PR implements the CSI node plugin functionality for NVMe-oF (NVMe over Fabrics) support in Ceph-CSI. It adds the core node-side operations required for mounting NVMe-oF volumes to pods.

The implementation includes:
- **NodeStageVolume**: Connects to NVMe-oF targets and mounts volumes to staging paths
- **NodePublishVolume**: Bind mounts staged volumes to pod directories
- **NodeUnstageVolume**: Unmounts 
- **NodeUnpublishVolume**: Unmounts volumes from pod directories
- **Metadata persistence**: Stores connection info to support idempotent operations

**Architecture decisions:**
- The implementation follows patterns from the RBD node plugin for consistency
- Uses UUID-based device paths (`/dev/disk/by-id/nvme-uuid.*`) for stability across reconnects
- Implements two-stage mounting (device → staging → pod) as per CSI specification

**Backward compatibility:**
- This is a new feature addition with no impact on existing RBD/CephFS functionality
- All NVMe-oF code is isolated in separate packages

**Review focus areas:**
- Mount option handling (particularly `_netdev` usage)
- Error handling and cleanup in failure scenarios
- Idempotency of stage/unstage operations

## Related issues\PR\discussions ##

This is part of the ongoing NVMe-oF support implementation. Related PRs will follow for:
- Controller plugin implementation  https://github.com/ceph/ceph-csi/pull/5452
- Volume provisioning and deletion #TODO
- Kubernetes deployment manifests #TODO
- https://github.com/ceph/ceph-csi/issues/5370
https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/nvme-of.md

## Future concerns ##

- Currently, the load nvme-tcp kernel module is located in NodeStageVolume, after driver.go file will be merged in another PR, this part will move to node-plugin creation place
- Implement reference counting for shared volumes (it will be probably in GW side in OMAP - it is under discussion) 

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/nvmeof/nodeserver/nodeserver.go
+++ b/internal/nvmeof/nodeserver/nodeserver.go
@@ -1,0 +1,656 @@
+/*
+Copyright 2025 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodeserver
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	mount "k8s.io/mount-utils"
+	utilexec "k8s.io/utils/exec"
+
+	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
+	"github.com/ceph/ceph-csi/internal/nvmeof"
+	"github.com/ceph/ceph-csi/internal/util"
+	"github.com/ceph/ceph-csi/internal/util/log"
+)
+
+// NodeServer struct of ceph nvmeof driver with supported methods of CSI
+// node server spec.
+type NodeServer struct {
+	csicommon.DefaultNodeServer
+	// A map storing all volumes with ongoing operations so that additional operations
+	// for that same volume (as defined by VolumeID) return an Aborted error.
+	volumeLocks *util.IDLocker
+
+	initiator nvmeof.NVMeInitiator
+}
+
+// ConnectionInfo holds NVMe-oF connection details.
+type NvmeConnectionInfo struct {
+	SubsystemNQN  string                  `json:"subsystemNQN"`
+	NamespaceID   uint32                  `json:"namespaceID"`
+	NamespaceUUID string                  `json:"namespaceUUID"`
+	Listeners     []nvmeof.GatewayAddress `json:"listeners"`
+	HostNQN       string                  `json:"hostNQN,omitempty"`
+	Transport     string                  `json:"transport"`
+}
+
+// stageTransaction struct represents the state a transaction was when it either completed
+// or failed
+// this transaction state can be used to rollback the transaction.
+type stageTransaction struct {
+	// isStagePathCreated represents whether the mount path to stage the volume on was created or not
+	isStagePathCreated bool
+	// isMounted represents if the volume was mounted or not
+	isMounted bool
+	// devicePath represents the path where nvmeof device is mapped
+	devicePath string
+}
+
+const (
+	nvmeofSubsystemNQN  = "SubsystemNQN"
+	nvmeofNamespaceID   = "NamespaceID"
+	nvmeofNamespaceUUID = "NamespaceUUID"
+	nvmeofListeners     = "listeners"
+	nvmeofHostNQN       = "HostNQN"
+	defaultTransport    = "tcp"
+)
+
+// NewNodeServer initialize a node server for ceph CSI driver.
+func NewNodeServer(
+	d *csicommon.CSIDriver,
+	nodeID,
+	t string,
+) *NodeServer {
+	// Create NVMe initiator
+	nvmeInitiator := nvmeof.NewNVMeInitiator()
+	ns := &NodeServer{
+		DefaultNodeServer: *csicommon.NewDefaultNodeServer(d, t, "", map[string]string{}, map[string]string{}),
+		initiator:         nvmeInitiator,
+		volumeLocks:       util.NewIDLocker(),
+	}
+
+	return ns
+}
+
+// NodeGetCapabilities returns the supported capabilities of the node server.
+func (ns *NodeServer) NodeGetCapabilities(
+	ctx context.Context,
+	req *csi.NodeGetCapabilitiesRequest,
+) (*csi.NodeGetCapabilitiesResponse, error) {
+	return &csi.NodeGetCapabilitiesResponse{
+		Capabilities: []*csi.NodeServiceCapability{
+			{
+				Type: &csi.NodeServiceCapability_Rpc{
+					Rpc: &csi.NodeServiceCapability_RPC{
+						Type: csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME,
+					},
+				},
+			},
+			{
+				Type: &csi.NodeServiceCapability_Rpc{
+					Rpc: &csi.NodeServiceCapability_RPC{
+						Type: csi.NodeServiceCapability_RPC_SINGLE_NODE_MULTI_WRITER,
+					},
+				},
+			},
+		},
+	}, nil
+}
+
+// NodeStageVolume mounts the volume to a staging path on the node.
+func (ns *NodeServer) NodeStageVolume(
+	ctx context.Context,
+	req *csi.NodeStageVolumeRequest,
+) (*csi.NodeStageVolumeResponse, error) {
+	var err error
+	if err = util.ValidateNodeStageVolumeRequest(req); err != nil {
+		return nil, err
+	}
+
+	volumeID := req.GetVolumeId()
+	cr, err := util.NewUserCredentialsWithMigration(req.GetSecrets())
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	defer cr.DeleteCredentials()
+	if acquired := ns.volumeLocks.TryAcquire(volumeID); !acquired {
+		log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volumeID)
+
+		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer ns.volumeLocks.Release(volumeID)
+
+	volumeContext := req.GetVolumeContext()
+	stagingTargetPath := getStagingTargetPath(req)
+
+	// Check if stagingPath is already mounted
+	isNotMnt, err := isNotMountPoint(ns.Mounter, stagingTargetPath)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	} else if !isNotMnt {
+		log.DebugLog(ctx, "nvmeof: volume %s is already mounted to %s, skipping", volumeID, stagingTargetPath)
+
+		return &csi.NodeStageVolumeResponse{}, nil
+	}
+
+	// Parse volume context
+	connectionInfo, err := ns.getNvmeConnection(volumeContext, req.GetPublishContext())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid volume context: %v", err)
+	}
+
+	// TODO: MOVE THE LOADING KERNEL MODULE TO DRIVE.GO!!!
+	// Load kernel modules first
+	if err := ns.initiator.LoadKernelModules(ctx); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to load kernel modules: %v", err)
+	}
+
+	// perform the actual staging and if this fails, have undoStagingTransaction
+	// cleans up for us
+	txn, err := ns.stageTransaction(ctx, req, connectionInfo)
+	defer func() {
+		if err != nil {
+			ns.undoStagingTransaction(ctx, req, txn)
+		}
+	}()
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	log.DebugLog(ctx, "nvmeof: successfully staged volume %s to stagingTargetPath %s",
+		volumeID, stagingTargetPath)
+
+	return &csi.NodeStageVolumeResponse{}, nil
+}
+
+// NodePublishVolume mounts the volume mounted to the staging target path to the target
+// path.
+func (ns *NodeServer) NodePublishVolume(
+	ctx context.Context,
+	req *csi.NodePublishVolumeRequest,
+) (*csi.NodePublishVolumeResponse, error) {
+	err := util.ValidateNodePublishVolumeRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	targetPath := req.GetTargetPath()
+	stagingPath := req.GetStagingTargetPath()
+	volID := req.GetVolumeId()
+	isBlock := req.GetVolumeCapability().GetBlock() != nil
+
+	// Add volume ID to staging path (this is where NodeStage mounted it)
+	stagingPath = stagingPath + "/" + volID
+
+	if acquired := ns.volumeLocks.TryAcquire(targetPath); !acquired {
+		log.ErrorLog(ctx, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+
+		return nil, status.Errorf(codes.Aborted, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+	}
+	defer ns.volumeLocks.Release(targetPath)
+
+	// Check if that target path exists properly
+	isMnt, err := ns.createTargetMountPath(ctx, targetPath, isBlock)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if isMnt {
+		return &csi.NodePublishVolumeResponse{}, nil
+	}
+
+	// Publish Path
+	err = ns.mountVolume(ctx, stagingPath, req)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	log.DebugLog(ctx, "nvmeof: successfully mounted stagingPath %s to targetPath %s", stagingPath, targetPath)
+
+	return &csi.NodePublishVolumeResponse{}, nil
+}
+
+// NodeUnpublishVolume unmounts the volume from the target path.
+func (ns *NodeServer) NodeUnpublishVolume(
+	ctx context.Context,
+	req *csi.NodeUnpublishVolumeRequest,
+) (*csi.NodeUnpublishVolumeResponse, error) {
+	err := util.ValidateNodeUnpublishVolumeRequest(req)
+	if err != nil {
+		return nil, err
+	}
+	targetPath := req.GetTargetPath()
+
+	if acquired := ns.volumeLocks.TryAcquire(targetPath); !acquired {
+		log.ErrorLog(ctx, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+
+		return nil, status.Errorf(codes.Aborted, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+	}
+	defer ns.volumeLocks.Release(targetPath)
+
+	isMnt, err := ns.Mounter.IsMountPoint(targetPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// targetPath has already been deleted
+			log.DebugLog(ctx, "targetPath: %s has already been deleted", targetPath)
+
+			return &csi.NodeUnpublishVolumeResponse{}, nil
+		}
+
+		return nil, status.Error(codes.NotFound, err.Error())
+	}
+	if !isMnt {
+		if err = os.Remove(targetPath); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		return &csi.NodeUnpublishVolumeResponse{}, nil
+	}
+
+	if err = ns.Mounter.Unmount(targetPath); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if err = os.Remove(targetPath); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	log.DebugLog(ctx, "nvmeof: successfully unbound volume %s from %s", req.GetVolumeId(), targetPath)
+
+	return &csi.NodeUnpublishVolumeResponse{}, nil
+}
+
+// NodeUnstageVolume unstages the volume from the staging path.
+func (ns *NodeServer) NodeUnstageVolume(
+	ctx context.Context,
+	req *csi.NodeUnstageVolumeRequest,
+) (*csi.NodeUnstageVolumeResponse, error) {
+	var err error
+	if err = util.ValidateNodeUnstageVolumeRequest(req); err != nil {
+		return nil, err
+	}
+
+	volumeID := req.GetVolumeId()
+	if acquired := ns.volumeLocks.TryAcquire(volumeID); !acquired {
+		log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volumeID)
+
+		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer ns.volumeLocks.Release(volumeID)
+
+	stagingTargetPath := getStagingTargetPath(req)
+
+	isMnt, err := ns.Mounter.IsMountPoint(stagingTargetPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
+		// Continue on ENOENT errors as we may still have the nvmeof device mapped
+		isMnt = false
+	}
+	if isMnt {
+		// Unmounting the image
+		err = ns.Mounter.Unmount(stagingTargetPath)
+		if err != nil {
+			log.ExtendedLog(ctx, "failed to unmount stagingPath: %s with error: %v", stagingTargetPath, err)
+
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		log.DebugLog(ctx, "successfully unmounted volume (%s) from staging path (%s)",
+			req.GetVolumeId(), stagingTargetPath)
+	}
+
+	if err = os.Remove(stagingTargetPath); err != nil {
+		// Any error is critical as Staging path is expected to be empty by Kubernetes, it otherwise
+		// keeps invoking Unstage. Hence any errors removing files within this path is a critical
+		// error
+		if !os.IsNotExist(err) {
+			log.ErrorLog(ctx, "failed to remove staging target path (%s): (%v)", stagingTargetPath, err)
+
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	}
+	log.DebugLog(ctx, "successfully removed staging path (%s)", stagingTargetPath)
+
+	return &csi.NodeUnstageVolumeResponse{}, nil
+}
+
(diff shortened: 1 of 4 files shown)


internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,702 @@
...
+	ctx context.Context,
+	req *csi.NodeStageVolumeRequest,
+	stagingPath, devicePath string,
+) error {
+	diskMounter := &mount.SafeFormatAndMount{Interface: ns.Mounter, Exec: utilexec.New()}
+	volumeCap := req.GetVolumeCapability()
+	isBlock := req.GetVolumeCapability().GetBlock() != nil
+
+	// Start with base mount options
+	mountOptions := []string{"_netdev"} // TODO : do i need it??
```
nixpanic: this is just playing nice with other tools on the host that may need to unmount the filesystem before the network is taken down (during a graceful reboot, for example)

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,702 @@
...
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if isMnt {
+		return &csi.NodePublishVolumeResponse{}, nil
+	}
+
+	// Publish Path
+	err = ns.mountVolume(ctx, stagingPath, req)
```
Madhu-1, line 226: return error (not GRPC) from `mountVolume` and add GRPC error here at higher level?

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,194 @@
...
+			return expectedPath, nil
+		}
+	}
+
+	return "", fmt.Errorf("device path with uuid: %s not found", uuid)
+}
+
+// Helper to format UUID with dashes.
+// TODO: move to nvmeof util.
+func formatUUID(uuid string) string {
```
nixpanic: This should probably use https://pkg.go.dev/github.com/google/uuid#Parse and UUID.String().

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,194 @@
...
+		// Add HostNQN only if specified
+		if req.HostNQN != "" {
+			args = append(args, "--hostnqn", req.HostNQN)
+		}
+
+		stdout, stderr, err := util.ExecCommandWithTimeout(ctx, connectTimeout, "nvme", args...)
+		// Execute connection
+		if err != nil {
+			log.WarningLog(ctx, "Failed to connect to %s:%s - stdout: %s, stderr: %s", address, port, stdout, stderr)
+			lastErr = err // TODO: fail or continue to next?
```
nixpanic: continue with the next connection if one fails, that is why the gateway can support HA

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,194 @@
...
+	log.DebugLog(ctx, "Loading NVMe-oF kernel modules")
+	var stderr string
+	modules := []string{
+		"nvme_tcp",
+		"nvme_fabrics",
+	}
+
+	for _, module := range modules {
+		// check if the module is loaded or compiled in
+		_, err := os.Stat("/sys/module/" + module)
```
nixpanic: We have this check/loading in multiple driver variations now, it should be moved to a util function.

#5564 adds `kmod.Modprobe()` for this.

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		log.DebugLog(ctx, "Successfully connected to subsystem %s via %s:%s",
+			req.SubsystemNQN, address, port)
+
+		return true, nil
+	}
+
+	// TODO: what to do if the list is empty or failed to connect to all ip? error or warning?
+	return false, fmt.Errorf("failed to connect to any gateway: %w", lastErr)
```
nixpanic: Failing to connect to all gateways should return an error. Even if one gateway is connected, it would be successful.

internal/nvmeof/subsystem_tracker.go
```diff
@@ -0,0 +1,308 @@
...
+// loadState loads the subsystem state from disk.
+func (t *fileBasedTracker) loadState() error {
+	// Check if state file exists
+	if _, err := os.Stat(t.stateFile); os.IsNotExist(err) {
+		log.DefaultLog("State file %s does not exist, starting with empty state", t.stateFile)
+
+		return nil
+	}
+
+	// Read state file
```
nixpanic: While reading and writing the file, there should be a file-lock used. It is unlikely, but technically possible that two node-plugins run on the same system (possibly with different drivernames). When one process is writing, an other reading process may read incomplete/incorrect data.

Maybe instead of the Mutex in the `fileBasedTracker`, a file lock can be used?

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -644,17 +643,27 @@ func (ns *NodeServer) connectToSubsystem(ctx context.Context, info *NvmeConnecti
 		}
 	}()
 
+	discoverReq := &nvmeof.DiscoveryRequest{
```
nixpanic: I wonder if the gateway should not discover the listeners. It can then return the list of listeners in the `PublishContext` so that it can be used in `NodeStageVolume`.

internal/nvmeof/subsystem_tracker.go
nixpanic: Replaced the mutex with `github.com/gofrs/flock`

internal/nvmeof/nodeserver/nodeserver.go
gadididi: @nixpanic , did you mean the `nvme` command should not discover the listeners? 
I created new PR: https://github.com/ceph/ceph-csi/pull/5589
passing of multiple listeners via `PublishContext`.
we can discuss about it (if run discover is good anyway, or redundant)

internal/nvmeof/nvmeof_initiator.go
gadididi: ok, I will add printing of what actually GWs we managed connect to. 
UPDATE: fixed.

internal/nvmeof/nodeserver/nodeserver.go
gadididi: UPDATE: discover function is removed.

internal/nvmeof/nvmeof_initiator.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
nixpanic: Yes, you need this. Please drop the comment, or improve it.

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+		return vr.GetStagingTargetPath() + "/" + vr.GetVolumeId()
+	case *csi.NodeUnstageVolumeRequest:
+		return vr.GetStagingTargetPath() + "/" + vr.GetVolumeId()
+	}
+
+	return ""
+}
+
+func (ns *NodeServer) readHostNQN() (string, error) {
+	data, err := os.ReadFile("/etc/nvme/hostnqn")
```
nixpanic: What to do is this file does not exist? This function runs inside the nodeplugin container, and if the container-image contains this file, all nodeplugins will use the same HostNQN.

Maybe pass the HostNQN from the commandline into the NodeServer struct?

gadididi: @nixpanic , I forgot to fix it. 
We agreed to create unique UUID per node-ID by generate_UUID(node_ID) in `ControllerPublishVolume()`, then pass it via ControlPublishContext to` NodeStageVolume()`, right?

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+func (ns *NodeServer) NodeGetCapabilities(
+	ctx context.Context,
+	req *csi.NodeGetCapabilitiesRequest,
+) (*csi.NodeGetCapabilitiesResponse, error) {
+	return &csi.NodeGetCapabilitiesResponse{
+		Capabilities: []*csi.NodeServiceCapability{
+			{
+				Type: &csi.NodeServiceCapability_Rpc{
+					Rpc: &csi.NodeServiceCapability_RPC{
+						Type: csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME,
```
Madhu-1, line 108: NodeServiceCapability_RPC_SINGLE_NODE_MULTI_WRITER also required to be added here

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+		return nil, err
+	}
+
+	// Read actual host NQN from system
+	hostNQN, err := ns.readHostNQN()
+	if err != nil {
+		log.ErrorLog(ctx, "Failed to read host NQN: %v", err)
+		// Fallback to default behavior or return error
+
+		return nil, fmt.Errorf("failed to read host NQN: %w", err)
```
Madhu-1: return GRPC error

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+	}
+
+	volumeID := req.GetVolumeId()
+	volumeContext := req.GetVolumeContext()
+
+	stagingParentPath := req.GetStagingTargetPath()
+	stagingTargetPath := getStagingTargetPath(req)
+
+	// Check if stagingPath is already mounted
+	isNotMnt, err := isNotMountPoint(ns.Mounter, stagingTargetPath)
```
Madhu-1, line 150: volume locks are not required here?

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+	}
+
+	log.DebugLog(ctx, "nvmeof: successfully staged volume %s to stagingTargetPath %s",
+		volumeID, stagingTargetPath)
+
+	return &csi.NodeStageVolumeResponse{}, nil
+}
+
+// NodePublishVolume mounts the volume mounted to the device path to the target
+// path.
```
Madhu-1:
```suggestion
// NodePublishVolume mounts the volume mounted to the staging target path to the target
// path.
```

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
+		// Continue on ENOENT errors as we may still have the nvmeof device mapped
+		isMnt = false
+	}
+	if isMnt {
+		// Unmounting the image
+		err = ns.Mounter.Unmount(stagingTargetPath)
+		if err != nil {
+			log.ExtendedLog(ctx, "failed to unmount targetPath: %s with error: %v", stagingTargetPath, err)
```
Madhu-1: `failed to unmount stagingPath`

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+	isMnt, err := mounter.IsMountPoint(stagingTargetPath)
+	if os.IsNotExist(err) {
+		err = nil
+	}
+
+	return !isMnt, err
+}
+
+// parseVolumeContext extracts connection information from volume context.
+func (ns *NodeServer) parseVolumeContext(volumeContext, publishContext map[string]string) (*NvmeConnectionInfo, error) {
```
Madhu-1: better to name this as GetNvmeConnection?

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+func (ns *NodeServer) cleanupVolumeMetadata(stagingTargetPath string) error {
+	metadataFile := filepath.Join(stagingTargetPath, stashFileName)
+	if err := os.Remove(metadataFile); err != nil {
+		return fmt.Errorf("failed to cleanup stashed JSON data (%s): %w", metadataFile, err)
+	}
+
+	return nil
+}
+
+// createStageMountPoint.
```
Madhu-1: incomplete comment

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,762 @@
...
+func (ns *NodeServer) createStageMountPoint(ctx context.Context, mountPath string, isBlock bool) error {
```
Madhu-1, line 561: return non grpc error and let the caller return grpc error?

internal/nvmeof/nodeserver/nodeserver.go
Madhu-1: This file exists on all the system by default and who generates it?

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,189 @@
...
+type nvmeInitiator struct{}
+
+// NewNVMeInitiator creates a new NVMe-oF initiator.
+func NewNVMeInitiator() NVMeInitiator {
+	return &nvmeInitiator{}
+}
+
+// LoadKernelModules ensures required kernel modules are loaded.
+func (ni *nvmeInitiator) LoadKernelModules(ctx context.Context) error {
+	log.DebugLog(ctx, "Loading NVMe-oF kernel modules")
```
Madhu-1: log which modules we are loading

Madhu-1 (commented):
over all LGTM, there are few duplication between this and rbd node server code, good to see if we can refactor both to a internal library for later use.

internal/nvmeof/nodeserver/nodeserver.go
gadididi: @Madhu-1 the The `/etc/nvme/hostnqn` file is created by the nvme-cli package. But, I am going to remove this approach. Instead, I am going to provide the node_id as hostNQN, because this `/etc/nvme/hostnqn` file is the same in every worker-node, because every worker-node uses the same prebuilt image..

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nvmeof_initiator.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: UPDATE:
I completely remove `readHostNQN()` function and the override function `NodeGetInfo()`.

in the hostnqn files you can see `nqn.2014-08.org.nvmexpress:uuid:<uuid>`:
1. `nqn.2014-08.org.nvmexpress` is a fixed prefix defined by the NVMe specification for UUID-based NQNs.
2. `2014-08` = August 2014 (when this UUID-based NQN format was standardized)
3. `org.nvmexpress` = The NVMe Express organization's reverse domain

I see in the GW code we dont have to provide the uuid at all.
So, we can just provide the hsotnqn with the hardcoded value `nqn.2014-08.org.nvmexpress:` (as standard hostnqn prefix)  
and in the part `uuid:<uuid>` (it is  `<user-part>` part) we put the node_ID value.
I will open another fix PR in `ControllerPublishVolume()` for this approach.

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: added, also added `VolumeLocks *util.VolumeLocks` property in `NodeServer` struct.

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,734 @@
...
+	"github.com/ceph/ceph-csi/internal/util/log"
+)
+
+// NodeServer struct of ceph nvmeof driver with supported methods of CSI
+// node server spec.
+type NodeServer struct {
+	csicommon.DefaultNodeServer
+	// A map storing all volumes with ongoing operations so that additional operations
+	// for that same volume (as defined by VolumeID) return an Aborted error
+	VolumeLocks *util.VolumeLocks
```
nixpanic: Please make `VolumeLocks` local to the package, rename it to `volumeLocks`. It also needs to be initialized inside `NewNodeServer()`.

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,734 @@
...
+	if err != nil {
+		log.ErrorLog(ctx, "failed to get namespace uuid %s: %v", info.NamespaceUUID, err)
+
+		return "", fmt.Errorf("failed to get namespace device: %w", err)
+	}
+
+	return devicePath, nil
+}
+
+// disconnectFromSubsystem just mock, need to be removed later.
```
nixpanic: Can this be removed now? Or, you can improve the comment by explaining when _later_ is.

nixpanic (commented):
Looks good to me, just a few small comments. Thanks!

internal/nvmeof/nodeserver/nodeserver.go
gadididi: @nixpanic , yes I can remove it. 
if I remove it, this line `connectionInfo, err := lookupNvmeVolumeMetadataStash(stagingParentPath)`
is no needed anymore and all the stashed metadata (that we save in the `NodeStageVolume` step)  is redundant. 
what do you think? 
should I remove `stashNvmeVolumeMetadata` , `lookupNvmeVolumeMetadataStash` , right?

nixpanic: Yes, ideally we do not need any metadata stored. Removing the stashing/file would be nice.

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed, load\save metadata functions and disconnect function are removed now.

nixpanic (dismissed):

internal/nvmeof/nodeserver/nodeserver.go
nixpanic: Thanks!

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,631 @@
...
+
+	isMnt, err := ns.Mounter.IsMountPoint(stagingTargetPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
+		// Continue on ENOENT errors as we may still have the nvmeof device mapped
+		isMnt = false
+	}
+	if isMnt {
```
Madhu-1, line 314: similar to RBD  we need to have lock on Publish/unpublish and NodeUnstage to avoid concurrent operations and also dependency on the CO to send the request serially.

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,631 @@
...
+		}
+
+		return false, nil
+	}
+
+	// block volume checks
+	// #nosec
+	pathFile, err := os.OpenFile(mountPath, os.O_CREATE|os.O_RDWR, 0o750)
+	if err != nil {
+		log.DebugLog(ctx, "Failed to create mountPath:%s with error: %v", mountPath, err)
```
Madhu-1: this should be error log

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,631 @@
...
+
+		return false, fmt.Errorf("failed to create mount file %q: %w", mountPath, err)
+	}
+	if err = pathFile.Close(); err != nil {
+		log.DebugLog(ctx, "Failed to close mountPath:%s with error: %v", mountPath, err)
```
Madhu-1: this should be error log

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,631 @@
...
+
+		return false, fmt.Errorf("failed to close mount file %q: %w", mountPath, err)
+	}
+
+	return false, nil
+}
+
+// isNotMountPoint checks whether MountPoint does not exists and
+// also discards error indicating mountPoint exists.
```
Madhu-1: `also discards error indicating mountPoint does not exists`

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -0,0 +1,631 @@
...
+func isNotMountPoint(mounter mount.Interface, stagingTargetPath string) (bool, error) {
+	isMnt, err := mounter.IsMountPoint(stagingTargetPath)
+	if os.IsNotExist(err) {
+		err = nil
+	}
+
+	return !isMnt, err
+}
+
+// getNvmeConnection extracts connection information from volume context.
```
Madhu-1: need to update comment to also mentioned about publishContext as well

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,173 @@
...
+
+		// Add HostNQN only if specified
+		if req.HostNQN != "" {
+			args = append(args, "--hostnqn", req.HostNQN)
+		}
+
+		stdout, stderr, err := util.ExecCommandWithTimeout(ctx, connectTimeout, "nvme", args...)
+		// Execute connection
+		if err != nil {
+			log.WarningLog(ctx, "Failed to connect to %s - stdout: %s, stderr: %s", listener, stdout, stderr)
```
Madhu-1, line 113: we need to log err as well here?

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,173 @@
...
+	log.DebugLog(ctx, "Disconnecting from NVMe-oF subsystem %s", subsystemNQN)
+
+	// Disconnect using nvme disconnect with NQN
+	args := []string{"disconnect", "-n", subsystemNQN}
+	stdout, stderr, err := util.ExecCommandWithTimeout(ctx, disconnectTimeout, "nvme", args...)
+	if err != nil {
+		log.WarningLog(ctx, "Failed to disconnect from %s - stdout: %s, stderr: %s", subsystemNQN, stdout, stderr)
+	}
+
+	log.DebugLog(ctx, "Successfully disconnected from subsystem %s", subsystemNQN)
```
Madhu-1: return inside to avoid the success debug log in case of error?

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,173 @@
...
+
+// DisconnectSubsystem disconnects from an NVMe-oF subsystem using NQN.
+func (ni *nvmeInitiator) DisconnectSubsystem(ctx context.Context, subsystemNQN string) error {
+	log.DebugLog(ctx, "Disconnecting from NVMe-oF subsystem %s", subsystemNQN)
+
+	// Disconnect using nvme disconnect with NQN
+	args := []string{"disconnect", "-n", subsystemNQN}
+	stdout, stderr, err := util.ExecCommandWithTimeout(ctx, disconnectTimeout, "nvme", args...)
+	if err != nil {
+		log.WarningLog(ctx, "Failed to disconnect from %s - stdout: %s, stderr: %s", subsystemNQN, stdout, stderr)
```
Madhu-1: log err as well?

Madhu-1 (commented):
small nits, We should start adding E2E setup/tests after this one as we have basic functionality now once this PR gets merged.

nixpanic:
> We should start adding E2E setup/tests after this one as we have basic functionality now once this PR gets merged.

The controller and nodeplugin still need to get the initialization in `internal/nvmeof/driver/` and `cmd/cephcsi.go`. When those are added, we can get started with automated testing. #5614 contains a start for that, it deploys the ceph-nvmeof gateway and provides a place for writing the tests in `e2e/nvmeof.go`.

internal/nvmeof/nvmeof_initiator.go
gadididi: basically, it is ok if we cloud not disconnect from subsystem because the ControllerPublish will remove the hostnqn (by remove host grpc) either way.
but in our approach, we dont run disconnect at all. I missed to delete this function. 
I am deleting it now.

internal/nvmeof/nvmeof_initiator.go
gadididi: this function `DisconnectSubsystem()` had been removed

internal/nvmeof/nvmeof_initiator.go
gadididi: We don't want to treat here like an error, because it tries in for loop to connect to each listener (=GW) if in some reason some GWs were down, it is ok as long as at least one of them is running. we want to support `High Availability`

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nodeserver/nodeserver.go
gadididi: thanks, fixed

Madhu-1 (approved):

internal/nvmeof/nvmeof_initiator.go
```diff
@@ -0,0 +1,154 @@
...
+// nvmeInitiator implements NVMeInitiator interface.
+type nvmeInitiator struct{}
+
+// NewNVMeInitiator creates a new NVMe-oF initiator.
+func NewNVMeInitiator() NVMeInitiator {
+	return &nvmeInitiator{}
+}
+
+// LoadKernelModules ensures required kernel modules are loaded.
+func (ni *nvmeInitiator) LoadKernelModules(ctx context.Context) error {
```
Rakshith-R, line 68: We should call this function from NewNodeServer since this function needs to called only once during first stage call ?

gadididi: @Rakshith-R ,yes correct, but there is no `driver.go` file yet. (Currently it calls every time NodestageVolune is called .. but not for long time) 
I added the TODO in this file:
`// TODO: MOVE THE LOADING KERNEL MODULE TO DRIVE.GO!!!`
the next PR will be` driver.go` file , where we call `NewNodeServer` and there I will call `LoadKernelModules` function

nixpanic (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

nixpanic:
We don't have e2e yet, skipping the two failed ones for k8s 1.34.

nixpanic:
@mergifyio requeue

PR #5497, opened 2025-08-13
util: cache the controller publish secret

black-dragon74:

This patch introduces a caching mechanism based on the shared informers for ControllerPublish secrets.

Do you have any questions?

- Should we watch for all namespaces?
- Is the default cache re-sync interval of 10 minutes ideal?

### DIFF
--- a/internal/util/k8s/secrets.go
+++ b/internal/util/k8s/secrets.go
@@ -18,15 +18,173 @@ package k8s
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"os"
+	"os/signal"
+	"sync"
+	"sync/atomic"
+	"syscall"
+	"time"
 
+	"github.com/ceph/ceph-csi/internal/util/log"
+
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
 )
+
+// cachedSecret is a string representation of a Kubernetes
+// secret's data.
+type cachedSecret struct {
+	data map[string]string
+}
+
+// secretCache is a thread safe cache for secrets.
+// /!\ The `cache` must be read/modified with the lock held.
+type secretCache struct {
+	// cache stores secrets identified by their hash.
+	// The cacheKey function defines how each hash is generated.
+	cache map[string]cachedSecret
+	sync.RWMutex
+	// A boolean indicating whether the secret watcher is running.
+	running atomic.Bool
+}
+
+// newSecretCache returns a new instance of secretCache.
+func newSecretCache() *secretCache {
+	return &secretCache{
+		cache: make(map[string]cachedSecret),
+	}
+}
+
+// cachedSecrets is a local instance of secretCache.
+var cachedSecrets = newSecretCache()
+
+// cacheKey is a helper to create sha256 hash based on a secret's name
+// and namespace. The hash is then used as an index for secretCache.
+func (sc *secretCache) cacheKey(ns, name string) string {
+	input := fmt.Sprintf("%s/%s", ns, name)
+	h := sha256.Sum256([]byte(input))
+
+	// Only use 16bytes of the hash as key, collisions are astronimically improbable.
+	// UUIDv4 is also 128bits and is used as a standard without worries of collision.
+	return hex.EncodeToString(h[:16])
+}
+
+// startSecretWatcher handles events on Secrets informer and updates
+// or invalidates the secretCache accordingly.
+func (sc *secretCache) startSecretWatcher(stopCh chan struct{}) error {
+	client, err := NewK8sClient()
+	if err != nil {
+		return fmt.Errorf("failed to connect to Kubernetes: %w", err)
+	}
+
+	factory := informers.NewSharedInformerFactoryWithOptions(
+		client,
+		// TODO: The sync interval below may need to be made configurable by the user.
+		time.Minute*10,
+	)
+
+	informer := factory.Core().V1().Secrets().Informer()
+	if _, err = informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(oldObj, newObj any) {
+			if newSecret, ok := newObj.(*corev1.Secret); ok {
+				_ = cachedSecrets.updateCache(newSecret, false)
+			}
+		},
+		DeleteFunc: func(obj any) {
+			if secret, ok := obj.(*corev1.Secret); ok {
+				cachedSecrets.deleteFromCache(secret)
+			}
+		},
+	}); err != nil {
+		return fmt.Errorf("failed to add event handler for secrets due to: %w", err)
+	}
+
+	go informer.Run(stopCh)
+
+	return nil
+}
+
+// deleteFromCache deletes a secret from secretCache.
+func (sc *secretCache) deleteFromCache(secret *corev1.Secret) {
+	key := sc.cacheKey(secret.Namespace, secret.Name)
+
+	sc.Lock()
+	defer sc.Unlock()
+
+	delete(sc.cache, key)
+}
+
+// updateCache updates a secret in secretCache if it is already present
+// or if force it set to true. It returns the secret's data if successful.
+func (sc *secretCache) updateCache(secret *corev1.Secret, force bool) map[string]string {
+	cKey := sc.cacheKey(secret.Namespace, secret.Name)
+
+	// We should cache this secret only if it was requested in the first place i.e force == true
+	// For all subsequent updates where this function is called, it is set to false
+	sc.RLock()
+	if _, requested := sc.cache[cKey]; !requested && !force {
+		sc.RUnlock()
+
+		return nil
+	}
+	sc.RUnlock()
+
+	secretData := make(map[string]string, len(secret.Data))
+	for key, value := range secret.Data {
+		secretData[key] = string(value)
+	}
+
+	cs := cachedSecret{
+		data: secretData,
+	}
+
+	sc.Lock()
+	defer sc.Unlock()
+
+	sc.cache[cKey] = cs
+
+	return cs.data
+}
 
 // GetSecret retrieves a Kubernetes Secret by its name and namespace.
 // It returns a map of key-value pairs contained in the Secret's data.
 // If the Secret does not exist or cannot be accessed, it returns an error.
 func GetSecret(secretName, secretNamespace string) (map[string]string, error) {
+	// Start the watcher if not already running, only once
+	if cachedSecrets.running.CompareAndSwap(false, true) {
+		stopCh := make(chan struct{})
+		if err := cachedSecrets.startSecretWatcher(stopCh); err != nil {
+			log.ErrorLogMsg("failed to start secret watcher: %v", err)
+		}
+
+		// Cleanup the channel
+		go func() {
+			termChan := make(chan os.Signal, 1)
+			signal.Notify(termChan, syscall.SIGINT, syscall.SIGTERM)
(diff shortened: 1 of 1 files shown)


internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,90 @@ package k8s
...
+	// Return from cache if present
+	cacheLock.RLock()
+	if entry, found := secretCache[key]; found {
+		cacheLock.RUnlock()
+
+		return entry.data, nil
+	}
+	cacheLock.RUnlock()
+
+	// The entry was not found in cache, fetch it from the API server
```
Rakshith-R: Let's start the watcher in first call to getSecret.
(It is currently started in omap generator.)

This way the watcher only starts in cephcsi container in case it tries to fetch secrets.

The approach looks good to me.

black-dragon74: Done. PTAL

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,110 @@ package k8s
...
+// It should preferably be started early in the bootstraping process.
+func StartSecretWatcher(stopCh chan struct{}) error {
+	client, err := NewK8sClient()
+	if err != nil {
+		return fmt.Errorf("failed to connect to Kubernetes: %w", err)
+	}
+
+	factory := informers.NewSharedInformerFactoryWithOptions(
+		client,
+		time.Minute*10, // FIXME: We could optionally restrict this to specific NS.
```
Rakshith-R: you can remove this FIXME, secrets maybe in any namespace

Rakshith-R (commented):
just a small nit,
please open the pr for review

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,135 @@ package k8s
...
+	"sync/atomic"
+	"time"
 
+	"github.com/ceph/ceph-csi/internal/util/log"
+
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
```
nixpanic: controller-runtime is very large, I don't think you need it here.

The `signals` package is only used for this:
```go
			<-signals.SetupSignalHandler().Done()
```

I think you can just write an empty struct to the channel too?
```go
<-struct{}{}
```

Or, consider replacing `stopCh chan struct{}` by `stopCh chan bool` and write `<-true` or something to it.

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,135 @@ package k8s
...
+)
+
+type cachedSecret struct {
```
nixpanic, line 41: Why introduce this type, and not use the Kubernetes Secret type?

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,135 @@ package k8s
...
+	data map[string]string
+}
+
+var (
+	secretCache    = make(map[string]cachedSecret)
```
nixpanic: please document the format of the key, and/or point to `cacheKey()`

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,135 @@ package k8s
...
+	return nil
+}
+
+// deleteSecretFromCache deletes a secret from secretCache.
+func deleteSecretFromCache(secret *corev1.Secret) {
+	key := cacheKey(secret.Namespace, secret.Name)
+
+	cacheLock.Lock()
+	delete(secretCache, key)
+	cacheLock.Unlock()
```
nixpanic: please use the more common pattern like
```go
	cacheLock.Lock()
	defer cacheLock.Unlock()

	delete(secretCache, key)
```

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,135 @@ package k8s
...
+		secretData[key] = string(value)
+	}
+
+	cs := cachedSecret{
+		data: secretData,
+	}
+
+	cacheLock.Lock()
+	secretCache[cKey] = cs
+	cacheLock.Unlock()
```
nixpanic: Here too, use `defer` for unlocking.

nixpanic:
> Do you have any questions?
> 
>     * Should we watch for all namespaces?

Only the namespaces+secrets that are requested. There will be many Secrets in a Kubernetes cluster, we definitely should not cache all of them. Even all Secrets in the namespace where Ceph-CSI is deployed may be way too much. We also don't want auditing to fire alarms when Ceph-CSI is caching secrets it would not use.

>     * Is the default cache re-sync interval of 10 minutes ideal?

A non-changeable default is never ideal. For some users it will be fine, others may want/need to tune it. Ideally there is an option to adjust the behavior.

internal/util/k8s/secrets.go
black-dragon74: To avoid iterating, casting and copying `secret.Data`

internal/util/k8s/secrets.go
black-dragon74: Ah, I got this mixed up with read locks, which can/should be released sequentially when possible. Fixed.

internal/util/k8s/secrets.go
black-dragon74: The idea was to avoid leaking the unclosed channel. Handled it via `signal.Notify`.

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,152 @@ package k8s
...
+	if watcherRunning.CompareAndSwap(false, true) {
+		stopCh := make(chan struct{})
+		if err := StartSecretWatcher(stopCh); err != nil {
+			log.ErrorLogMsg("failed to start secret watcher: %v", err)
+		}
+
+		// Cleanup the channel
+		go func() {
+			termChan := make(chan os.Signal, 1)
+			signal.Notify(termChan, os.Interrupt, syscall.SIGTERM)
```
nixpanic: mixing `os.` and `syscall.` signal specifications looks a bit weird. Can you use only signals from the `syscall` package?

black-dragon74: Done. Fun fact, `os.Interrupt` is internally defined as `syscall.SIGINT`. Why does `os.Kill` (`syscall.SIGKILL` makes sense as it is a system signal) even exist though, AFAIK it can't be caught/trapped on *NIX systems, no?

internal/util/k8s/secrets.go
nixpanic: I guess nothing uses `secret.StringData`, but only the base64 encoded values?

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,152 @@ package k8s
...
+type cachedSecret struct {
+	data map[string]string
+}
+
+var (
+	// secretCache stores secrets in memory, identified by their hash.
+	// The cacheKey function defines how each hash is generated.
+	secretCache    = make(map[string]cachedSecret)
+	cacheLock      sync.RWMutex
+	watcherRunning atomic.Bool
```
nixpanic: There is one thing I don't like (not really object to) about this. There are now three new package wide variables, and quite a few functions that are only for the secret cache. To clean this up, and have fewer variables/functions in the package, you can make it more object oriented.

```go
type secretCache struct {
    cache    map[string]cachedSecret
   lock sync.RWMutex
   running atomic.Bool
}

var cachedSecrets = newSecretCache()

func newSecretCache() *secretCache {
    ...
}

func (sc *secretCache) cacheKey(ns, name string) string {
    ...
}
```

This way, smart editors and the godoc documentation group everything much better together.

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,152 @@ package k8s
...
+	input := fmt.Sprintf("%s/%s", ns, name)
+	h := sha256.Sum256([]byte(input))
+
+	// Only use 16bytes
+	return hex.EncodeToString(h[:16])
+}
+
+// StartSecretWatcher handles events on Secrets informer and updates
+// or invalidates the secretCache accordingly.
+func StartSecretWatcher(stopCh chan struct{}) error {
```
nixpanic: this does not need to be a public function, call it `startSecretWatcher()`.

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,152 @@ package k8s
...
+	client, err := NewK8sClient()
+	if err != nil {
+		return fmt.Errorf("failed to connect to Kubernetes: %w", err)
+	}
+
+	factory := informers.NewSharedInformerFactoryWithOptions(
+		client,
+		time.Minute*10,
```
nixpanic, line 88: Make this a constant, and document a **TODO** that it may need to be configurable.

nixpanic (commented):
Looks good, just a few nits that should be corrected. The OOP change would be nice, but isn't absolutely required.

internal/util/k8s/secrets.go
black-dragon74: Done. PTAL :)

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,172 @@ package k8s
...
+// cachedSecrets is a local instance of secretCache.
+var cachedSecrets = newSecretCache()
+
+// cacheKey is a helper to create sha256 hash based on a secret's name
+// and namespace. The hash is then used as an index for secretCache.
+func (sc *secretCache) cacheKey(ns, name string) string {
+	input := fmt.Sprintf("%s/%s", ns, name)
+	h := sha256.Sum256([]byte(input))
+
+	// Only use 16bytes
```
nixpanic: I understand that the code only uses 16 bytes. The comment should better explain _why_ only 16 bytes should be used. Is there no issue with potential abbreviated-hash-collisions?

internal/util/k8s/secrets.go
```diff
@@ -18,15 +18,172 @@ package k8s
...
 )
 
+// cachedSecret is a string representation of a Kubernetes
+// secret's data.
+type cachedSecret struct {
+	data map[string]string
+}
+
+// secretCache is a thread safe cache for secrets.
+// /!\ The `cache` must be read/modified with the lock held.
```
nixpanic, line 46: Just wondering why you decided to use `sync.RWMutex` and not `sync.Map`. Can you explain that a little?

nixpanic (commented):
Looks good. Just waiting for a response on the outstanding questions/comments.

internal/util/k8s/secrets.go
black-dragon74: `sync.Map` is not strongly typed and would require type casting.

internal/util/k8s/secrets.go
black-dragon74: Added a comment. For a cache key, 128bits is collision is improbable until order of 10^9 entries. We'd be good with more than a billion entries.

internal/util/k8s/secrets.go
black-dragon74: `map[string][]byte` vs `map[string]string`. While storing in cache we iterate over the bytes and cast them to string. Storing them like that is more convenient as the code elsewhere expects a `map[string]string`

Rakshith-R (approved):
Thanks !

Rakshith-R:
@mergifyio rebase

iPraveenParihar (approved):
Thanks!

iPraveenParihar:
@Mergifyio rebase

iPraveenParihar:
@Mergifyio queue

iPraveenParihar:
@Mergifyio queue

iPraveenParihar:
https://github.com/Mergifyio queue

iPraveenParihar:
@Mergifyio refresh

PR #5498, opened 2025-08-13
rbd: consider lastSyncTimeNotFound as image not syncing

Rakshith-R:

This commit modifies code to consider image as not syncing when the lastSyncTime is not found.

Currently replication's Resync call fails if the lastSyncTime of an image is not found
when deciding if an image is in syncing state or not.
Instead, we should treat missing lastSyncTime as the image not being
in Syncing state.

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -716,6 +716,8 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 			err.Error())
 	}
 
+	// If the savedImageTime is not empty,
+	// then the image might be in need of resync.
 	if savedImageTime != "" {
 		st, sErr := timestampFromString(savedImageTime)
 		if sErr != nil {
@@ -723,12 +725,23 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
 		}
 		log.DebugLog(ctx, "image %s, savedImageTime=%v, currentImageTime=%v", rbdVol, st, creationTime)
 
+		// Fetch the last sync info from the local status in order
+		// to determine if the image is syncing or not.
 		syncInfo, sErr := localStatus.GetLastSyncInfo(ctx)
-		if sErr != nil {
+		if sErr != nil && !errors.Is(sErr, rbderrors.ErrLastSyncTimeNotFound) {
 			return nil, status.Errorf(codes.Internal, "failed to get last sync info: %s", sErr.Error())
 		}
+		// consider the image to be not syncing if either
+		// - the last sync time was not found
+		// or
+		// - the sync info indicates the image is not syncing
+		isNotSyncing := errors.Is(sErr, rbderrors.ErrLastSyncTimeNotFound) || !syncInfo.IsSyncing()
 
-		if req.GetForce() && st.Equal(*creationTime) && !syncInfo.IsSyncing() {
+		// image needs to be resynced if
+		// - force option is set
+		// - image creation time is equal to the saved image creation time
+		// - image is not already in syncing state
+		if req.GetForce() && st.Equal(*creationTime) && isNotSyncing {
 			err = mirror.Resync(ctx)
 			if err != nil {
 				return nil, getGRPCError(err)


internal/csi-addons/rbd/replication.go
```diff
@@ -724,11 +724,16 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
 
 		syncInfo, sErr := localStatus.GetLastSyncInfo(ctx)
-		if sErr != nil {
+		if sErr != nil && !errors.Is(sErr, rbderrors.ErrLastSyncTimeNotFound) {
 			return nil, status.Errorf(codes.Internal, "failed to get last sync info: %s", sErr.Error())
 		}
 
-		if req.GetForce() && st.Equal(*creationTime) && !syncInfo.IsSyncing() {
+		// consider the image is not syncing if either
+		// - the last sync time was not found or
```
Madhu-1: please add a comment that when last sync time is not found for example when in split brain state etc. it will help us in future why we used this check.

internal/csi-addons/rbd/replication.go
```diff
@@ -724,11 +724,16 @@ func (rs *ReplicationServer) ResyncVolume(ctx context.Context,
...
+		// - the sync info indicates the image is not syncing
+		IsNotSyncing := (errors.Is(sErr, rbderrors.ErrLastSyncTimeNotFound) || !syncInfo.IsSyncing())
```
```suggestion
isNotSyncing := errors.Is(sErr, rbderrors.ErrLastSyncTimeNotFound) || !syncInfo.IsSyncing()
```

Madhu-1 (changes requested):
small nit

internal/csi-addons/rbd/replication.go
Rakshith-R: > please add a comment that when last sync time is not found for example when in split brain state etc. it will help us in future why we used this check.

I've added example as when images on both peered clusters are in secondary state.

The actual check is to skip resync if the image is syncing.

The last sync time not found error can be considered equivalent to the image is not in syncing state.

Madhu-1: >I've added example as when images on both peered clusters are in secondary state.

I think this is wrong, because if the images are secondary state on both cluster, both will be in up+unknown state (we have check for this even before this state). This is more of an error state when not in healthy secondary state.  when both are in secondary state its not possible to resync as there is no primary image.

Rakshith-R: > > I've added example as when images on both peered clusters are in secondary state.
> 
> I think this is wrong, because if the images are secondary state on both cluster, both will be in up+unknown state (we have check for this even before this state). This is more of an error state when not in healthy secondary state. when both are in secondary state its not possible to resync as there is no primary image.

resync with both images in secondary is a no-op AFAIK.

Rakshith-R: > > I've added example as when images on both peered clusters are in secondary state.
> 
> I think this is wrong, because if the images are secondary state on both cluster, both will be in up+unknown state (we have check for this even before this state). This is more of an error state when not in healthy secondary state. when both are in secondary state its not possible to resync as there is no primary image.

The behavior of resync execution when both images are secondary is not changed by this pr.

Madhu-1: i think you are missing my point, my request is to add a comment not to change any code flow. For example that the last sync might be missing when image is in up+error or split brain state

Madhu-1 (dismissed):
its good to add a comment if not possible, this PR need to be revisited later when we get into problem to understand the reason

nixpanic:
please describe in the commit message and PR what kind of problem can be seen without this change

Rakshith-R:
Added a lot of comments at various steps
and describe the issue seen without this change.

PTAL

nixpanic (approved):

Madhu-1 (approved):

Madhu-1:
@mergifyio queue

PR #5500, opened 2025-08-14
rbd: return replication status and status message in case of error

Nikhil-Ladha:

We should return the status and status message even if we fail to fetch the local status of the volume/group so that the error/state can be propagated to the VR/VGR in csi-addons.
For ex, if the local mirror status is `down+unknown` when rbd is not able to fetch the mirror status, we should report the same on the VR/VGR i.e, the `state` being `unknown` and the message being `status not found`.

Also, it seems more logical to return a `FailedPreCondition` error code if `sync info not found` error is hit.

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -916,7 +916,7 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(ctx context.Context,
 	lastSyncInfo, err := remoteStatus.GetLastSyncInfo(ctx)
 	if err != nil {
 		if errors.Is(err, rbderrors.ErrLastSyncTimeNotFound) {
-			return nil, status.Errorf(codes.NotFound, "failed to get last sync info: %v", err)
+			return nil, status.Errorf(codes.FailedPrecondition, "failed to get last sync info: %v", err)
 		}
 
 		return nil, status.Errorf(codes.Internal, "failed to get last sync info: %v", err)
@@ -933,12 +933,7 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(ctx context.Context,
 		resp.LastSyncDuration = durationpb.New(*lastDuration)
 	}
 
-	replicationStatus, statusMessage, err := getCurrentReplicationStatus(ctx, mirrorStatus)
-	if err != nil {
-		log.ErrorLog(ctx, err.Error())
-
-		return nil, status.Errorf(codes.Internal, "failed to get local status: %v", err)
-	}
+	replicationStatus, statusMessage := getCurrentReplicationStatus(ctx, mirrorStatus)
 
 	resp.Status = replicationStatus
 	resp.StatusMessage = statusMessage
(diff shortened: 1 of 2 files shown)


internal/csi-addons/rbd/replication.go
```diff
@@ -922,9 +929,9 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(ctx context.Context,
 
 	replicationStatus, statusMessage, err := getCurrentReplicationStatus(ctx, mirrorStatus)
 	if err != nil {
+		// log error and return response so that we can surface replication status
+		// and message in csi-addons
 		log.ErrorLog(ctx, err.Error())
```
Rakshith-R: return response is missing below

Nikhil-Ladha: We are returning in the end, which would contain the status and message returned from the `getCurrentReplicationStatus` for the error condition.

Rakshith-R: > We are returning in the end, which would contain the status and message returned from the `getCurrentReplicationStatus` for the error condition.

It's not a good practice to rely on return values when err !=nil,
I suggest you complete remove the err return value from getCurrentReplicationStatus and 
always log the Status and message and return it in response.

internal/csi-addons/rbd/replication.go
```diff
@@ -903,7 +903,14 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(ctx context.Context,
...
 		if errors.Is(err, rbderrors.ErrLastSyncTimeNotFound) {
-			return nil, status.Errorf(codes.NotFound, "failed to get last sync info: %v", err)
+			// log error and return response so that we can surface replication status
+			// and message in csi-addons
+			log.ErrorLog(ctx, "failed to get last sync info: %v", err)
+
+			return &replication.GetVolumeReplicationInfoResponse{
+				Status:        replication.GetVolumeReplicationInfoResponse_UNKNOWN,
+				StatusMessage: "sync info not found",
+			}, nil
```
nixpanic: Initially there was a `codes.NotFound`. The way I read [the error codes from the spec](https://github.com/csi-addons/spec/tree/main/replication#error-scheme-5), `FAILED_PRECONDITION` would be more appropriate. No need to fake an error in a `GetVolumeReplicationInfoResponse`.

nixpanic:
Looks reasonable to me, do you want to adjust the commit message and the description of this PR to clarify things a little more? Specially in the PR description, an example of what was missing, and how it will look like afterwards would be very helpful.

Nikhil-Ladha:
> Looks reasonable to me, do you want to adjust the commit message and the description of this PR to clarify things a little more? Specially in the PR description, an example of what was missing, and how it will look like afterwards would be very helpful.

Sure, updated the description to add some more context.
Hopefully, it looks better now :)

Rakshith-R (approved):

nixpanic (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio queue

PR #5525, opened 2025-08-19
nvmeof: prevent concurrent requests for the same volume

nixpanic:
On occasion the container orchitrator (Kubernetes) will retry CSI
operations that have timed out. In certain failure scenarios, these
operations are actually still active, and just blocked (or very slow) at
the Ceph cluster.

A 2nd try of the same operation should not be attempted, as it mau cause
conflicts and will likely result in the slow/hanging situation again.

By protecting concurrent operations for the same volume and reporting a
failure, users at least can see that something is malfunctioning.

Updates: #5370


Rakshith-R (approved):

Nikhil-Ladha (approved):

Nikhil-Ladha:
@Mergifyio queue

PR #5560, opened 2025-09-01
util: fix read affinity activation by command flags

npdgm:

Two mechanisms are offered to enable read affinity and configure CRUSH location labels:
- command flags `--enable-read-affinity` and `--crush-location-labels`
- cluster configuration attributes `readAffinity.enabled` and `readAffinity.crushLocationLabels`

It is expected that both could be used to enable the feature, with the possibility of one having precedence over the other.

When parsing the cluster configuration, a missing `readAffinity.enabled` key gets unmarshalled as a non-nullable bool and set to `false`. As a result, `GetReadAffinityMapOptions()` never considered CLI flags when the JSON file was missing read affinity settings.

This PR fixes the fallback logic so that command flags get some authority on enabling read affinity.
Further improvement would require `readAffinity.enabled` to be unmarshalled as a nullable type.

Do you have any questions?
No

Yes

Maybe

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/util/read_affinity.go
+++ b/internal/util/read_affinity.go
@@ -62,11 +62,7 @@ func GetReadAffinityMapOptions(
 		return "", err
 	}
 
-	if !configReadAffinityEnabled {
-		return "", nil
-	}
-
-	if configCrushLocationLabels == "" {
+	if !configReadAffinityEnabled || configCrushLocationLabels == "" {
 		return cliReadAffinityMapOptions, nil
 	}
 


iPraveenParihar:
@npdgm, Actually there was plan to remove/deprecate the CLI read affinity options if I remember correctly.

@Rakshith-R ^

rdegez:
Hi (I work with @npdgm)
Whatever solution you choose will be fine but the status-quo is not OK since it leave you thinking that read-locality is enabled when using the helm-chart values (and the logs will even display things about the crush_location mapping from k8s label) whereas in fact it's not!  
Hence this patch.
Cheers,

Rakshith-R:
:thinking:  https://github.com/ceph/ceph-csi/blob/79c4e5b587d6a3f4e58e394422796bc4ff4d63fb/api/deploy/kubernetes/csi-config-map.go#L35 
ReadAffinity is a struct here, that's causing the problem.

I think we should just document that users need to enable ReadAffinity in configmap per clusterID too 
but crushlocation labels could be borrowed either from configmap or cmdline.

https://github.com/ceph/ceph-csi/blob/79c4e5b587d6a3f4e58e394422796bc4ff4d63fb/docs/rbd/deploy.md#read-affinity-using-crush-locations-for-rbd-volumes

We definitely want to give priority to per cluster options in configmap.

npdgm:
In light of these comments, I agree that CLI indicators should indeed be retired. Configurations involving multiple clusters require individual settings for consistency.
So feel free to reject this PR if you don't want to introduce a change of behaviour before flags are removed.
Cheers

PR #5569, opened 2025-09-05
rbd: add support for radosNamespace from StorageClass parameters

cybercoder:
This change allows specifying the RADOS namespace directly in the StorageClass parameters using the 'radosNamespace' parameter, instead of only reading it from the cluster configuration file.

The implementation:
- Checks for 'radosNamespace' parameter in StorageClass first
- Falls back to existing behavior (config file) if not specified
- Maintains backward compatibility

Example StorageClass usage:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: storageclass1
  annotations:
    storageclass.kubernetes.io/is-default-class: "false"
provisioner: rbd.csi.ceph.com
parameters:
   pool: pool1
   radosNamespace: ns1
   imageFeatures: layering
   imageFormat: "2"
   csi.storage.k8s.io/provisioner-secret-name: ceph-pool1-secret
   csi.storage.k8s.io/provisioner-secret-namespace: ceph-csi
   csi.storage.k8s.io/controller-expand-secret-name: ceph-pool1-secret
   csi.storage.k8s.io/controller-expand-secret-namespace: ceph-csi
   csi.storage.k8s.io/node-stage-secret-name: ceph-pool1-secret
   csi.storage.k8s.io/node-stage-secret-namespace: ceph-csi
reclaimPolicy: Retain
allowVolumeExpansion: true
mountOptions:
   - discard

```
Fixes: ability to set RADOS namespace per StorageClass

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -1407,9 +1407,13 @@ func genVolFromVolumeOptions(
 		return nil, err
 	}
 
-	rbdVol.RadosNamespace, err = util.GetRBDRadosNamespace(util.CsiConfigFile, rbdVol.ClusterID)
-	if err != nil {
-		return nil, err
+	if namespace, ok := volOptions["radosNamespace"]; ok && namespace != "" {
+		rbdVol.RadosNamespace = namespace
+	} else {
+		rbdVol.RadosNamespace, err = util.GetRBDRadosNamespace(util.CsiConfigFile, rbdVol.ClusterID)
+		if err != nil {
+			return nil, err
+		}
 	}
 	if rbdVol.Mounter, ok = volOptions["mounter"]; !ok {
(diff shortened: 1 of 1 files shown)


Rakshith-R (changes requested):
Hey,

It was conscious design choice to put radosNamespace and cephfs subvolumegroup in configmap rather than in storageclass. 
We would like to keep it that way for the following reason.

RadosNamespace is referred to in multiple RPC calls where only VolumeHandle is available for the driver depend on.

Therefore, just putting RadosNamespace in storageclass -> VolumeContext is insufficient.

Corbeau3000:
It there a way to use differents namespaces (ns1, ns2, ns3) within a single pool? Actually I can't figure out a way, so I am stuck using a single pool/namespace per StorageClass.

PR #5595, opened 2025-09-15
nvmeof: fix remove host in ControllerUnpublishVolume

gadididi:
validate if the host is not needed before remove

**Fix little bug:** 
NVMe-oF CSI driver incorrectly removes hosts from subsystems during ControllerUnpublishVolume even when the host is still using other namespaces in the same subsystem, causing connectivity issues for remaining volumes.

**solution:**
Host should only be removed from subsystem when no namespaces are actively using it.

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/nvmeof/controller/controllerserver.go
+++ b/internal/nvmeof/controller/controllerserver.go
@@ -565,6 +565,20 @@ func unpublishResources(ctx context.Context, data *nvmeof.NVMeoFVolumeData, node
 		}
 	}()
 
+	// check if there are more namespaces in this subsystem with this host, if so, do not remove the host.
+	namespaces, err := gateway.ListNamespaces(ctx, subsystemNQN)
+	if err != nil {
+		return fmt.Errorf("failed to list namespaces for subsystem %s: %w", subsystemNQN, err)
+	}
+	for _, ns := range namespaces.GetNamespaces() {
+		for _, host := range ns.GetHosts() {
+			if host == hostNQN {
+				log.DebugLog(ctx, "Host %s is still using namespace %s, not removing", hostNQN, ns.GetNsid())
+
+				return nil
+			}
+		}
+	}
 	// Remove host from subsystem
 	if err := gateway.RemoveHost(ctx, subsystemNQN, hostNQN); err != nil {
 		return fmt.Errorf("failed to remove host %s from subsystem %s: %w",


iPraveenParihar (approved):

internal/nvmeof/controller/controllerserver.go
```diff
@@ -565,6 +565,20 @@ func unpublishResources(ctx context.Context, data *nvmeof.NVMeoFVolumeData, node
...
+		for _, host := range ns.GetHosts() {
+			if host == hostNQN {
+				log.DebugLog(ctx, "Host %s is still using namespace %s, not removing", hostNQN, ns.GetNsid())
+
+				return nil
+			}
+		}
+	}
 	// Remove host from subsystem
 	if err := gateway.RemoveHost(ctx, subsystemNQN, hostNQN); err != nil {
```
Rakshith-R, line 583: If multiple namespaces are tied to a single host,
then we need some kind of rw locking mechanism ? 

Just simply checking for namespaces in host at a single point in time will not work ? 

What if controllerPublish for another volume is running in parallel
and passes this point in flow ?

https://github.com/ceph/ceph-csi/blob/3b552ee245c3e226e0ce994f879a65dc9c48bb1a/internal/nvmeof/nvmeof.go#L309-L313

Rakshith-R (changes requested):

internal/nvmeof/controller/controllerserver.go
gadididi: @Rakshith-R ,I think you raised good point,
you are absolutely right.
the exact scenario that can make issue:
```md
Initial state: Host connected with 1 namespace (Volume A)

Thread 1 (UnpublishVolume A): Lists namespaces → sees 1 namespace (Volume A)
Thread 2 (CreateVolume B): Creates namespace B → now there are 2 namespaces total
Thread 2 (PublishVolume B): Calls AddHost → gets EEXIST → continues
Thread 1 (continues from step 1): Based on its earlier check showing 1 namespace, decides to remove host
Thread 1: Removes host
Result: Volume B's namespace exists but has no host access
```
another example(less severe):
```md
Initial state: Host is connected to subsystem with 2 namespaces (Volume A, Volume B)

Thread 1 (UnpublishVolume A): Checks namespaces → sees 2 namespaces → decides NOT to remove host
Thread 2 (UnpublishVolume B): Checks namespaces → sees 2 namespaces → decides NOT to remove host
Thread 1: Removes namespace A → now only Volume B remains
Thread 2: Removes namespace B → now NO namespaces remain
Result: Host is still in subsystem but has NO namespaces to access (orphaned host)
```

The entire code (including add\remove subsystem\ns\listener\host) for now does not support multithread.
I added it here: https://github.com/ceph/ceph-csi/issues/5370#issuecomment-3163939133
I will address this supporting in the next PRs

Rakshith-R (approved):

Rakshith-R:
@mergifyio rebase

Rakshith-R:
@mergifyio rebase

Rakshith-R:
@mergifyio queue

PR #5605, opened 2025-09-26
nvmeof: prevent concurrent adding/removing of hosts to the gateway

nixpanic:
When a host is added to or removed from the gateway, a lock protects
concurrent adding/removing of the same host in ControllerPublishVolume
or ControllerUnpublishVolume.

In case a 2nd gRPC operation is blocked due to the locking, an error is
returned to the caller. The caller will have to retry the operation at a
later time (usually Kubernetes retries with a very short delay).

Modifications to a subsystem in the gateway are now done under a lock,
so that only a single CreateVolume or DeleteVolume changes the
configuration at a time. This prevents conflicts where a DeleteVolume
removes listeners or even the subsystem from the gateway, when a
CreateVolume assumed everything was in place, so skipped the creation of
the subsystem and adding of listeners.

The VolumeLock structs and NewVoluemLocks() function provide a locking
API that isn't restricted to volumes only. The implementation is very
generic, and can be used by other parts of the code that need to lock
operations based on others resources.

By renaming VolumeLocks to IDLocker (it is already in the idlocker.go
file), the genericness of the locking API is expressed better.

Users of the API are recommended to call their IDLocker object with an
understandable name. VolumeLocks for locking resources related to
volumes, HostLocks for locking resources related to hosts, and so on.

Updates: #5370

### DIFF
--- a/internal/nvmeof/controller/controllerserver.go
+++ b/internal/nvmeof/controller/controllerserver.go
@@ -48,6 +48,10 @@ type Server struct {
 	// the gateway during ControllerPublishVolume and ControllerUnpublishVolume.
 	hostLocks *util.VolumeLocks
 
+	// subsystemLocks prevents concurrent calls from deleting an "empty but not empty
+	// anymore" subsystem (and listeners).
+	subsystemLocks *util.VolumeLocks
+
 	// backendServer handles the RBD requests
 	backendServer *rbd.ControllerServer
 }
@@ -55,9 +59,10 @@ type Server struct {
 // NewControllerServer initialize a controller server for nvmeof CSI driver.
 func NewControllerServer(d *csicommon.CSIDriver) (*Server, error) {
 	return &Server{
-		volumeLocks:   util.NewVolumeLocks(),
-		hostLocks:     util.NewVolumeLocks(),
-		backendServer: rbddriver.NewControllerServer(d),
+		volumeLocks:    util.NewVolumeLocks(),
+		hostLocks:      util.NewVolumeLocks(),
+		subsystemLocks: util.NewVolumeLocks(),
+		backendServer:  rbddriver.NewControllerServer(d),
 	}, nil
 }
 
@@ -133,7 +138,7 @@ func (cs *Server) CreateVolume(
 	rbdPoolName := res.GetVolume().GetVolumeContext()["pool"]
 
 	// Step 2: Setup NVMe-oF resources
-	nvmeofData, err := createNVMeoFResources(ctx, req, rbdPoolName, rbdImageName)
+	nvmeofData, err := cs.createNVMeoFResources(ctx, req, rbdPoolName, rbdImageName)
 	if err != nil {
 		log.ErrorLog(ctx, "NVMe-oF resource setup failed for volumeID %s: %v", volumeID, err)
 
@@ -145,7 +150,7 @@ func (cs *Server) CreateVolume(
 			return
 		}
 
-		cleanupErr := cleanupNVMeoFResources(ctx, nvmeofData)
+		cleanupErr := cs.cleanupNVMeoFResources(ctx, nvmeofData)
 		if cleanupErr != nil {
 			log.ErrorLog(ctx, "failed to cleanup NVMe-oF resources for volume  %q: %v", volumeID, cleanupErr)
 		}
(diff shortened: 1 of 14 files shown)


internal/nvmeof/controller/controllerserver.go
```diff
@@ -45,6 +45,10 @@ type Server struct {
 	// for that same volume (as defined by VolumeID/volume name) return an Aborted error
 	volumeLocks *util.VolumeLocks
```
gadididi, line 46: should not we add also `OperationLocks *util.OperationLock` property?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -45,6 +45,10 @@ type Server struct {
...
 
+	// hostLocks protects against concurrently adding hosts to, and removing hosts from
+	// the gateway during ControllerPublishVolume and ControllerUnpublishVolume.
+	hostLocks *util.VolumeLocks
```
gadididi: I am just wondering about this scenario: 
1. there is subsystem "A" user1 uses it.
2. user1 finishes to use and delete the pvc. `DleteVolume()` is called, but does not delete the subsystem yet. (it searches and see no more namespaces left in this subsystem, so this subsystem is going to removed)
3. in the meantime user2 create pvc with the same subsystem. `CreateVolume()` is called, it sees the subsystem exists, so go to create namespace.
4. back to user1 , it deletes the subsystem.
5. back to user2 , it tries to create the namespace but got an error, because the subsystem does not exist.

I think because it failed, the  `CreateVolume()` will be called again and will be pass successfully. but maybe we should add locking for each resource? (subsystem, namespace and listener) as well?

gadididi (commented):
LGTM!
I left 2 questions

internal/nvmeof/controller/controllerserver.go
nixpanic: I considered that, but the operations are pre-defined and used in combination with the volume-id. In our case, we need locking per host-id, independent of the volume-id.

The name _VolumeLocks_ is probably the most confusing, in the end it is just a map that hold a state for a key.  The key usually is a volume-id, but it can be a host-id just as well.

internal/nvmeof/controller/controllerserver.go
nixpanic: Yes, great you mention this. It is something I was considering too! I'll have a look and add locking for those operations as well.

internal/nvmeof/controller/controllerserver.go
gadididi: I see. Maybe add (in controllerserver.go file) 
```go
type HostIdLocks = util.VolumeLocks 
```
or something like that, to make it clear. what do you think?

nixpanic: I'm actually considering to rename `util.VolumeLocks` to `util.IDLocker` :rofl:

internal/util/idlocker.go
```diff
@@ -31,6 +31,9 @@ const (
 
 	// TargetPathOperationAlreadyExistsFmt string format to return for concurrent operation on target path.
 	TargetPathOperationAlreadyExistsFmt = "an operation with the given target path %s already exists"
+
+	// HostOperationAlreadyExistsFmt string format to return for concurrent operation.
```
Madhu-1:
```suggestion
// HostOperationAlreadyExistsFmt string format to return for concurrent host operation.
```

comment seems to be too generic

internal/nvmeof/controller/controllerserver.go
nixpanic: This has been done now. A lock per subsystem that is held while creating/deleting the subsystem, listeners and namespace.

internal/util/idlocker.go
nixpanic: addressed this too.

internal/nvmeof/controller/controllerserver.go
nixpanic: Renamed the VolumeLocks struct to IDLocker in a separate commit.

internal/nvmeof/controller/controllerserver.go
Madhu-1: Will there be a case like below

* User 1 delete the PVC and sub system deletion fails for some reason
* User 2 create the PVC and a namespace is created
* when i User 1 RPC is triggered. it will return success response and only deletes the namespace 

am i right?

Madhu-1 (approved):

internal/nvmeof/controller/controllerserver.go
nixpanic: Yes, that should be handled correctly.

If the 2nd user attempts to create a volume while the subsystem is locked by the 1st, the 2nd will get an error and should retry. When the 1st fails during its operation, it will retry as well (causing further cleanup).

Rakshith-R (approved):
Thanks

PR #5641, opened 2025-10-09
nvmeof: add examples and deployment files

nixpanic:
Examples and deployment files for NVMe-oF. These are used with the e2e testing
as well.

See-also: #5672

PR #5645, opened 2025-10-09
ci: add nvmeof test-type support for triggering CI jobs

nixpanic:
- **ci: add test_type=nvmeof for running only the NVMe-oF tests**
- **ci: drop unneeded `test_type-` prefix from trigger phrase**

Relates-to: #5641

### DIFF
--- a/mini-e2e-helm.groovy
+++ b/mini-e2e-helm.groovy
@@ -198,15 +198,15 @@ node('cico-workspace') {
 		}
 		stage('run e2e') {
 			timeout(time: 150, unit: 'MINUTES') {
-				def t_type = ""
+				def t_type = "" // test all by default
 				if ("${test_type}" == "cephfs"){
 					t_type = '--test-cephfs=true --test-rbd=false --test-nfs=false'
 				} else if ("${test_type}" == "rbd"){
 					t_type = '--test-rbd=true --test-cephfs=false --test-nfs=false'
 				} else if ("${test_type}" == "nfs"){
 					t_type = '--test-nfs=true --test-cephfs=false --test-rbd=false'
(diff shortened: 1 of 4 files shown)


mini-e2e-helm.groovy
```diff
@@ -198,15 +198,15 @@ node('cico-workspace') {
...
 				if ("${test_type}" == "cephfs"){
 					t_type = '--test-cephfs=true --test-rbd=false --test-nfs=false'
 				} else if ("${test_type}" == "rbd"){
 					t_type = '--test-rbd=true --test-cephfs=false --test-nfs=false'
 				} else if ("${test_type}" == "nfs"){
 					t_type = '--test-nfs=true --test-cephfs=false --test-rbd=false'
-				} else {
-					t_type = '--test-rbd=true --test-cephfs=true --test-nfs=true'
+				} else if ("${test_type}" == "nvmeof"){
+					t_type = '--test-nvmeof=true --test-nfs=false --test-cephfs=false --test-rbd=false'
```
Madhu-1, line 209: Instead of depending on the defaults in e2e tets. lets disable nvmeof tets if its only for cephfs or rbd or nfs not to run it by default?

nixpanic: Well, passing `--test-nvmeof=false` won't work for older branches that do not have the `--test-nvmeof` parameter. I've done it this way so that the other branches do not need changes to the e2e parameters.

But yeah, the nvmeof tests would run even if only one of the other types are selected. it is not optimal.

Madhu-1 (approved):

Rakshith-R (approved):

PR #5711, opened 2025-10-27
rbd: make block encryption cipher configurable  

### LINKED ISSUE #5686: rbd: Enable Configurable Ciphers and Key Sizes for RBD Encryption

Background: We want to be able to use different ciphers for dmcrypt disk encryption, not only the default option AES-XTS. 

Problem: Currenlty disk encryption done through dmcrypt in ceph-csi RBD volume encryption can only be with AES-XTS even though dmcrypt supports different ciphers. 

Because different ciphers have different effective key sizes it makes sense to also make the key size configurable. For example, the linux kernel implements AES-XTS with a key size of 512 whereas other AES-based ciphers are implemented with a key size of 128 or 256. 

When AES-XTS does not provide sufficient security, it would be useful to be able to change the cipher to better align with the security requirements of the use case. 

Add a list of criteria that should be met for this feature to be useful.

- If ciphers are configurable for RBD volume encryption encryption other than the default AES-XTS
- If the key size is configurable
- If the node does not support a configured cipher and key size pair, then an error message containing this information must be displayed
- If no cipher is specified the current default cipher AES-XTS should still be the default 

# Additional context #

## Idea behind implemention 

Main logic behind LUKS encryption is found in /internal/util/cryptsetup/cryptsetup.go

So we need to get arguments from storage class RBD volume configuration to the cryptsetup calls. 

More importantly, since the LUKS header stores the chosen cipher, it only needs to be specified when the RBD volume is formatted for the first time. There is no need to specify the cipher on subsequent openings.

Greenpepper15:

The PR adds three new block encryption (RBD encryption) configuration options to customize the encryptions means used.

The option "encryptionCipher" configures the cipher with which the block device is encrypted. The option "encryptionKeySize" is about configuring the key size.

Lastly the option "integrityMode" configures the algorithms the block encryption should use to verify the integrity of consumed data. When  this option is not used then no integrity verification happens. This is implemented via the kernel module dmIntegrity.  Essentially when cryptsetup receives the “integrityMode” flag then it will set up a dmcrypt mapper device and a dmIntegrity mapper device. 

When the integrity option is not used IO stack looks like this 

App (Pod) —> dmcrypt mapper —> rbd device 

With the integrity option it will look like this

App (Pod) —> dmcrypt mapper —> dmIntegrity mapper —> rbd device 

In this setup the dmcrypt device would encrypt data and compute authentication tags (for example hmacs). Meanwhile dmIntegrity would be tasked with storing the encrypted data and corresponding authentication tags on the block device.

## Motivation ## 

I want to be able to customize means to employ encryption to match a required security level. Since I added an allowlist for cipher strings not any algorithm can be deployed (which hinders configurability) but it is the most easiest and secure means to validate user input. 

Additionally I want to integrate means of verifying the integrity of consumed data because it’s becoming increasingly required from a compliance standpoint. For example the recently released Digital Operational Resilience Act (DORA) legislation from the EU demands that data confidentiality and integrity need to be established. I see the usage of dmIntegrity as the easiest means to achieve this. 

Do you have any questions?

- In the PR a resize operation is not possible when integrity protection is enabled. This is because in older kernel releases (older than 5.7) it was not possible
- I made an allowlist for the “encryptionCipher” and “integrityMode” configurations due to security concerns. Do you also think that this allowlist makes sense? 

- The biggest hurdle for backwards compatibility will be the dmIntegrity. For example in this PR I forbid the usage of resize when the integrity option is used because on older linux kernel versions the resize key word was not supported. 

These sources illustrate why integrating an integrity mode makes sense:  

- [Digital Operational Resilience Act](https://www.eiopa.europa.eu/digital-operational-resilience-act-dora_en) (DORA)
    -  Article 5 (2) (b) “[The management body shall] put in place policies that aim to ensure the maintenance of high standards of availability, authenticity, integrity and confidentiality, of data;”
     
- [BSI TR-02102-1 "Kryptographische Verfahren: Empfehlungen und Schlüssellängen" Version: 2025-01](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR02102/BSI-TR-02102.pdf?__blob=publicationFile&v=13)
    - Germany's Federal Office for Information Security
    - Documents talks about when encryption is used means to establish that the ciphertext was not altered is crucial 
    - “Die Verwendung einer Volume-Verschlüsselungslösung allein ist nur empfehlenswert, wenn diese einen wirksamen kryptographischen Schutz vor Manipulation der Daten beinhaltet [...]”
    - Translation: The use of a volume encryption solution alone is only recommended if it includes effective cryptographic protection against data manipulation.

- Here is useful [dmintegrity documentation](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/DMIntegrity)

## Related issues ##

Resolves ceph#5686

## Future concerns ##

- Resize is currently not supported when an integrity mode is enabled. However since a resize is possible in dmIntegrity with newer linux kernel versions (larger than [5.7](https://gitlab.com/cryptsetup/cryptsetup/-/issues/388#note_955610140)) we could support it
- The new configuration options could also be used to make file system encryption more customizable

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>


Greenpepper15:
Hi @nixpanic or @Madhu-1 , since you are both very familiar with this repository, I would appreciate your feedback on my recent changes.

I know you're busy, so if it would be helpful or save time, I am happy to schedule a quick meeting to walk you through the modifications. Maybe we can use the next Dev standup slot. Please let me know what works best for you.

Madhu-1:
> Hi @nixpanic or @Madhu-1 , since you are both very familiar with this repository, I would appreciate your feedback on my recent changes.
> 
> I know you're busy, so if it would be helpful or save time, I am happy to schedule a quick meeting to walk you through the modifications. Maybe we can use the next Dev standup slot. Please let me know what works best for you.

@Greenpepper15 Thanks for the work 🎉 , requested for reviews on this one.  Lets continue the discussion in the PR itself, if required we can have a this as a topic in one of the standup.

examples/rbd/storageclass.yaml
```diff
@@ -135,6 +135,14 @@ parameters:
    # Only applicable for block encryption.
    # A string is expected here, i.e. "128", not 128.
    # encryptionKeySize: "128"
+
+   # (optional) Instruct the plugin verify the integrity when data is read.
+   # By default it is disabled. If encrypted is false then this flag will be ignored. 
+   # Specify the integrity verification mode via the dm-Integrity specification.
+   # Selecting this option prevents the image from being resized..
```
iPraveenParihar:
```suggestion
   # Selecting this option prevents the image from being resized.
```

examples/rbd/storageclass.yaml
```diff
@@ -122,26 +122,31 @@ parameters:
    # A string is expected here, i.e. "true", not true.
    # encrypted: "true"
 
-   # (optional) Instruct the plugin it has to encrypt the volume with this specific cipher.
-   # By default it is disabled. If encrypted is false then this flag will be ignored. 
+   # (optional) Instruct the plugin it has to encrypt the volume
```
iPraveenParihar, line 134: I think these changes are part of previous commit.

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -54,19 +54,59 @@ const (
...
+func (e *EncryptionOptions) IntegrityMode() *string {
+	return e.integrityMode
 }
 
-func (e *EncryptionOptions) SetAeadMode(aead string) {
-	e.aeadMode = new(string)
-	*e.aeadMode = aead
+func (e *EncryptionOptions) SetintegrityMode(integrity string) error {
+	if err := integrityModeValidator.validate(integrity); err != nil {
+		return fmt.Errorf("cannot set integrity mode'%s': %w", integrity, err)
```
iPraveenParihar:
```suggestion
		return fmt.Errorf("cannot set integrity mode '%s': %w", integrity, err)
```

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -54,19 +54,59 @@ const (
...
+}
+
+var (
+	cipherValidator = &validator{
+		name: "cipher",
+		allowedValues: map[string]struct{}{
+			"aes-xts-plain64":     {},
+			"serpent-xts-plain64": {},
+			"twofish-xts-plain64": {},
+			"aegis128-random":     {},
```
iPraveenParihar: do you think we need to validate keysize for these ciphers?
Like what happens If a user provides cipher `aegis128-random` and 256 as keysize? Do the luks.Format() fails?

examples/rbd/storageclass.yaml
```diff
@@ -122,6 +122,19 @@ parameters:
...
 
+   # (optional) Instruct the plugin it has to encrypt the volume with this specific cipher.
+   # By default it is disabled. If encrypted is false then this flag will be ignored. 
+   # Specify cipher according to the cryptsetup specification.
+   # Only applicable for block encryption.
+   # encryptionCipher: "aes-cbc-essiv:sha256"
+
+   # (optional) Instruct the plugin to use a certain key size to encrypt data.
+   # Most ciphers support key sizes of 128 or 256 bit. 
+   # By default it is disabled. If encrypted is false and encryptionCipher is not set
```
iPraveenParihar:
```suggestion
   # By default it is disabled. If encrypted is false or encryptionCipher is not set
```

internal/rbd/encryption.go
```diff
@@ -367,6 +374,26 @@ func (ri *rbdImage) initKMS(ctx context.Context, volOptions, credentials map[str
 	return nil
 }
 
+func ParseCipherOptions(volOptions map[string]string) (*cryptsetup.EncryptionOptions, error) {
```
iPraveenParihar: please add unit test for this.

iPraveenParihar (commented):
Could you please add e2e test with these options enabled in StorageClass?

internal/util/cryptsetup/cryptsetup.go
Greenpepper15: Yes if a key size is used that is not compatible with the cipher than luks.format() will fail. 

The aegis128-random case is an easy case because if any other key size than 128 is used  it will fail. But I am unsure if we should do it to all possible ciphers because than we would couple our implementation with the cipher suites available on the host system. 

At the end the behavior would be the same. Either our sanity check will generate an error or luks.format() will generate an error. 

But far as I can see from my local testing the quality of the error message from luks.format() varies widely. 
For example this 
`sudo cryptsetup luksFormat --type luks2 --cipher aegis128-random --integrity aead --key-size 256 /dev/vdb`
results in 
`Cipher aegis128-random (key size 256 bits) is not available.`
which is good. 

But something like 
`sudo cryptsetup luksFormat --type luks2 --cipher aes-xts-random --integrity hmac-sha256 --key-size 1024 /dev/vdb`
(xts modes only allow key sizes of 256 and 512 on my system) 
results in 
`device-mapper: reload ioctl on temporary-cryptsetup-fa35e15a-2c7f-4b6f-aea2-02ff3bfe7624 (253:1) failed: Invalid argument`
which will is not very expressive. 

Based on this I think a Ceph-csi local (cipher, key size) pair validation seems like a good idea. If people want to support different key sizes they can open a small PR. 

What do you think?

Greenpepper15:
Yes I am working on adding the e2e tests and unit tests.

internal/util/crypto.go
```diff
@@ -159,6 +166,10 @@ func (ve *VolumeEncryption) GetID() string {
 	return ve.id
 }
 
+func (ve *VolumeEncryption) CipherOptions() *cryptsetup.EncryptionOptions {
```
nixpanic, line 170: new functions and stucts ideally get a small comment, here and in `cryptsetup.go`

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -54,9 +54,37 @@ const (
...
+type EncryptionOptions struct {
+	cipher  string
+	keysize *uint
+}
+
+func (e *EncryptionOptions) Keysize() *uint {
+	return e.keysize
+}
+
+func (e *EncryptionOptions) SetKeySize(keysize string) error {
```
nixpanic, line 392: `SetKeySize` and above `Keysize`, please pick one way of Camel[cC]asing.

internal/rbd/nodeserver.go
```diff
@@ -69,8 +69,8 @@ type stageTransaction struct {
 	isMounted bool
 	// isBlockEncrypted represents if the volume was encrypted or not
 	isBlockEncrypted bool
-	// isAEADEncrypted represents if the volume was encrypted using an AEAD cipher
-	isAEADEncrypted bool
+	// isIntegrityProtected represents if the volume was establishes the cryptographic integrity of read data
+	isIntegrityProtected bool
```
nixpanic, line 74: In a previous commit you added `isAEADEncrypted`, and now you are replacing it with `isIntegrityProtected`. It might be intentional to show the development, but I wonder about the usefulness. Does it make this PR simpler, or more complex?

internal/util/cryptsetup/cryptsetup.go
nixpanic: What about keeping a _known good_ list, and reporting a warning about an unknown combination? That way users may send a PR for additional _known good_ combinations, and are not completely surprised when `luksFormat` spits out an incomprehensible error.

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -54,19 +54,59 @@ const (
...
-func (e *EncryptionOptions) AeadMode() *string {
-	return e.aeadMode
+func (e *EncryptionOptions) IntegrityMode() *string {
+	return e.integrityMode
 }
 
-func (e *EncryptionOptions) SetAeadMode(aead string) {
-	e.aeadMode = new(string)
-	*e.aeadMode = aead
+func (e *EncryptionOptions) SetintegrityMode(integrity string) error {
```
nixpanic: watch the case usage of `IntegrityMode()` and `SetintegrityMode()` :eyeglasses:

docs/rbd/deploy.md
```diff
@@ -68,6 +68,9 @@ make image-cephcsi
...
 | `encrypted`                                                                                         | no                   | disabled by default, use `"true"` to enable either LUKS or fscrypt encryption on PVC and `"false"` to disable it. **Do not change for existing storageclasses**                                                                                                                                                      |
+| `encryptionCipher`                                                                                  | no                   | set the cipher that is used for the volume encryption. **Supported Keywords**: `aes-xts-plain64` (default), `serpent-xts-plain64`, `twofish-xts-plain64`, `aegis128-random` (requires `integrityMode`: `aead` and `encryptionKeySize`: `"128"`)                                                                            |
```
nixpanic: This commit has a little too long lines in the commit message, **commitlint** fails because of that.

nixpanic (commented):
Thanks for the PR, looks quite ok in general. Missing unit+e2e tests are known and worked on, great!

It would be good to point to the cryptsetup documentation somewhere, in the hope it contains a list of valid/recommended/working combinations for ciphers and keysizes.

internal/rbd/nodeserver.go
Greenpepper15: Midway I decided to change the name midway because isIntegrityProtected covers more of the semantic I want to cover. 
In that structure we already have a member that indicate if the image is encrypted (`isBlockEncrypted`) therefore indicating that something is `isBlockEncrypted` and `AEADEncrypted` both indicate that the image is encrypted, but the AEAD parts adds the semantic that the integrity is also protected. Hence I decided to rename the member to `isIntegrityProtected`. Now the `isBlockEncrypted` and `isIntegrityProtected ` is true than we know the image confidently and integrity is protected.

internal/util/cryptsetup/cryptsetup.go
Greenpepper15: Yeah that sounds like a good compromise. I will implement this.

internal/rbd/nodeserver.go
nixpanic: Thanks for the additional explanation, all good :+1:

internal/rbd/encryption.go
```diff
@@ -367,6 +374,34 @@ func (ri *rbdImage) initKMS(ctx context.Context, volOptions, credentials map[str
...
```
Madhu-1: lets not expose this function make it `parseCipherOptions`

internal/rbd/encryption.go
```diff
@@ -367,6 +374,34 @@ func (ri *rbdImage) initKMS(ctx context.Context, volOptions, credentials map[str
...
+	cipher, cipherOk := volOptions["encryptionCipher"]
+	keysize, keysizeOk := volOptions["encryptionKeySize"]
+	integrity, integrityOk := volOptions["integrityMode"]
+	if !cipherOk {
+		return nil, nil
+	}
+	opts := &cryptsetup.EncryptionOptions{}
+	if cipherOk {
```
Madhu-1: i think we dont need this check because we are already returning at 381

internal/rbd/encryption.go
```diff
@@ -367,6 +374,34 @@ func (ri *rbdImage) initKMS(ctx context.Context, volOptions, credentials map[str
...
+		if err := opts.SetCipher(cipher); err != nil {
+			return nil, fmt.Errorf("failed to set cipher: %w", err)
+		}
+	}
+	// when cipher is not set keysize is not used
+	if cipherOk && keysizeOk {
+		if err := opts.SetKeySize(keysize); err != nil {
+			return nil, fmt.Errorf("failed to set key size: %w", err)
+		}
+	}
```
Madhu-1:
```suggestion
if keysize, ok := volOptions["encryptionKeySize"];ok{
if err := opts.SetKeySize(keysize); err != nil {
			return nil, fmt.Errorf("failed to set key size: %w", err)
     }
}
```

lets group it together and have local loop, same for integrity as well

internal/rbd/encryption_test.go
```diff
@@ -96,3 +96,127 @@ func TestParseEncryptionOpts(t *testing.T) {
...
+				actual := actualEncOptions.Cipher()
+				comparePtr(t, "Cipher", tt.expectedCipher, &actual)
+				comparePtr(t, "IntegrityMode", tt.expectedIntegrity, actualEncOptions.IntegrityMode())
+				comparePtr(t, "Key Size", tt.expectedKeysize, actualEncOptions.Keysize())
+			},
+		)
+	}
+}
+
+func comparePtr[T comparable](t *testing.T, fieldName string, expected, actual *T) {
```
Madhu-1: can we use https://pkg.go.dev/github.com/stretchr/testify/assert package for it? i think with this we can reduce lot of code in test file

internal/rbd/nodeserver.go
```diff
@@ -558,6 +560,10 @@ func (ns *NodeServer) stageTransaction(
 			return transaction, err
 		}
 		transaction.isBlockEncrypted = true
+		// Cryptsetup does not support Resize when Integrity protection is enabled
+		if cipherOpts := volOptions.blockEncryption.CipherOptions(); cipherOpts != nil && cipherOpts.IntegrityMode() != nil {
```
Madhu-1, line 565: Not related to this check but do i understand this correct where we cannot resize the device? what about the expansion in NodeExpandVolume, dont we need to block the volume expansion for this volumes at the early stage at controller level itself?

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -54,9 +54,92 @@ const (
...
+type validator struct {
+	allowedValues map[string]struct{}
+	name          string
+}
+
+func (v *validator) validate(value string) error {
+	_, found := v.allowedValues[value]
+	if !found {
+		return fmt.Errorf("%s not allowed: %s", v.name, value)
+	}
```
Madhu-1:
```suggestion
if _, ok := v.allowedValues[value];!ok {
		return fmt.Errorf("%s not allowed: %s", v.name, value)
	}
```

internal/util/cryptsetup/cryptsetup.go
Madhu-1: `SetIntegrityMode`

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -54,9 +54,92 @@ const (
...
+func (e *EncryptionOptions) IntegrityMode() *string {
+	return e.integrityMode
+}
+
+func (e *EncryptionOptions) SetintegrityMode(integrity string) error {
+	if err := integrityModeValidator.validate(integrity); err != nil {
+		return fmt.Errorf("cannot set integrity mode'%s': %w", integrity, err)
+	}
+	e.integrityMode = new(string)
+	*e.integrityMode = integrity
```
Madhu-1: can we simply do below and let go handle to allocation

`e.integrityMode=&mode`

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -54,9 +54,92 @@ const (
...
+
+	return nil
+}
+
+func (e *EncryptionOptions) Keysize() *uint {
```
Madhu-1: `KeySize`

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -54,9 +54,92 @@ const (
...
+	return e.keysize
+}
+
+func (e *EncryptionOptions) SetKeySize(keysize string) error {
+	parsedVal, err := strconv.ParseUint(keysize, 10, 64)
```
Madhu-1: 32 is enough here right or do we really need 64?

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -79,15 +162,25 @@ func NewLUKSWrapper(ctx context.Context) LUKSWrapper {
...
 		"--hash",
 		"sha256",
+	}
+	if cipherOptions != nil {
+		args = append(args, "--cipher", cipherOptions.Cipher())
+		if cipherOptions.IntegrityMode() != nil {
+			args = append(args, "--integrity", *cipherOptions.IntegrityMode())
+		}
+		if cipherOptions.Keysize() != nil {
+			args = append(args, "--key-size", strconv.FormatUint(uint64(*cipherOptions.Keysize()), 10))
```
Madhu-1: use fmt.Sprintf instead of conversion and formatting?

Madhu-1 (commented):
overall Looks good to me, @Greenpepper15 Thanks for the great work 🎉 Lets add the missing E2E testing for it.

Greenpepper15:
Hey I got a question concerning the e2e tests because I ran into a cyclic dependency. 

For the e2e test I want to expose a function in /internal/util/cryptsetup/cryptsetup.go that returns all the allowed Ciphers to the e2e tests. 

Then I can do something like 
```go
err = createRBDStorageClass(
       ...
	map[string]string{"encrypted": "true",
		"encryptionType":   crypto.EncryptionTypeBlock.String(),
		"encryptionCipher":  cryptsetup.GetAllowedCiphers()[0],   <-- Return an allowed cipher that I definitely know is valid because it comes from the source
	},
	deletePolicy)
```

But the e2e tests get the cryptsetup functionality through the file /e2e/vendor/github.com/ceph/ceph-csi/internal/until/crypsetup/cryptsetup.go does not have the new GetAllowedCiphers() defined because it references an old version of crypetup.go. 

Should I go something like this for now 

```go
err = createRBDStorageClass(
       ...
	map[string]string{"encrypted": "true",
		"encryptionType":   crypto.EncryptionTypeBlock.String(),
		"encryptionCipher":  "aes-xts-plain64",   <-- Insert a static string that I know will work now but may change in a later version
	},
	deletePolicy)
```

And then later open another PR to replace the static assignments with a GetAllowedCiphers() call?

Madhu-1:
@Greenpepper15 have you tried running `go mod tidy` in the e2e folder and see if that works?

Greenpepper15:
Yes I just tired and it works. Thanks for the advice. 
But `go mod tidy` changes a lot of files in the e2e/vendor directory not just the cryptsetup.go file. Is that acceptable?

Madhu-1:
> Yes I just tired and it works. Thanks for the advice. But `go mod tidy` changes a lot of files in the e2e/vendor directory not just the cryptsetup.go file. Is that acceptable?

Yes that should be fine

e2e/utils.go
```diff
@@ -906,6 +908,69 @@ func checkMountOptions(pvcPath, appPath string, f *framework.Framework, mountFla
...
+	}
+	if stdErr != "" {
+		return nil, fmt.Errorf("failed command to get cryptsetup status from %s %s", mountPath, stdErr)
+	}
+
+	return parseLuksStatus(stdOut)
+}
+
+// PraseLuksStatus parses parts of the output of a "cryptsetup luksStatus <device>" command
+func parseLuksStatus(dump string) (luksStatus *cryptsetup.LuksStatus, err error) {
```
Greenpepper15: In the e2e tests I get the encryption state of the PV via `cryptsetup status` to compare it against the desired encryption state. 

Should I put this function into the cryptsetup.go file instead? I initially put it here because only the e2e tests use it but since the cryptsetup.go file encapsulates all the `cryptsetup` logic I think it would be better suited for it. 
What do you think?

e2e/rbd_helper.go
```diff
@@ -519,7 +582,26 @@ func validateEncryptedPVCAndAppBinding(pvcPath, appPath string, kms kmsConfig, f
...
 		return err
 	}
-
+	options, err := getEncryptionOptionsFromPV(pvc.Namespace, pvc.Name, f)
+	if err != nil {
+		return fmt.Errorf("failed to get encryption options for pv %q: %w",
+			imageData.pvName, err)
+	}
+	if options != nil {
+		luksStatus, err := getLuksStatusFromMount(imageData.pvName, app.Name, f)
```
Greenpepper15, line 593: Here you can see the largest semantical difference between my encryption tests and the previous encryption tests. 

In the previous e2e block encryption tests the desired encryption state is retrieved from the attribute `encrypted` from the rbd image metadata. Then it compares the actual pv encryption state with the `getMountType()` function (which in the background calls `lsblk -o TYPE,MOUNTPOINT | grep '%s' | awk '{print $1}'`). 

In my tests this is also done but in addition the tests try to retrieve the `encryptionCipher`, `encryptionKeySize` and `integrityMode` attributes from the pv volume attributes (`pv.Spec.CSI.VolumeAttributes`) as the desired encryption state. If these are present then `cryptsetup status` is called on the pv to retrieve the actual encryption state. After some parsing, these two encryption states are compared. 

Do you think this methodology is okay?

internal/rbd/nodeserver.go
Greenpepper15: The main reason why I did this was because a few lines after this we have this piece of code inside the `stageTransaction()` function. 

```go
	if !volOptions.readOnly {
		// As we are supporting the restore of a volume to a bigger size and
		// creating bigger size clone from a volume, we need to check filesystem
		// resize is required, if required resize filesystem.
		// in case of encrypted block PVC resize only the LUKS device.
		err = resizeNodeStagePath(ctx, isBlock, transaction, req.GetVolumeId(), stagingTargetPath)
		if err != nil {
			return transaction, err
		}
	}
```
Here we are also allowed to stop the resize operation in the event that the volume is read only (I stop the resize operation in the first few lines in the `resizeNodeStagePath()` function with the `isIntegrityProtected` attribute). That is why I thought it was fine to stop it here.

internal/util/cryptsetup/cryptsetup.go
Greenpepper15: Yeah, should be more than enough.

internal/util/cryptsetup/cryptsetup.go
Greenpepper15: Because if I use fmt.Sprinft() the test scripts fails with:

```
../internal/util/cryptsetup/cryptsetup.go:309:38: integer-format: fmt.Sprintf can be replaced with faster strconv.FormatUint (perfsprint)
                        args = append(args, "--key-size", fmt.Sprintf("%d", *cipherOptions.KeySize()))
```
Sadly strconv.FormatUint() only works with uint64. This is why I converted it.

internal/rbd/nodeserver.go
Madhu-1: This should be fine, i assume we will not support resizing PVC when IntegrityMode is set in sc right?

e2e/rbd_helper.go
Madhu-1: Yes should be fine, we can take some decision of what need to be tested and how based on the attributes.

e2e/utils.go
Madhu-1: sounds good 👍🏻

Greenpepper15:
Can somebody trigger the e2e tests? I tested them on my Cluster but I want to see if they work in the CI environment too.

nixpanic:
@Greenpepper15 , the tests pass, but I do not see the test-cases you added being executed :thinking: 

```
$ curl 'https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.34-rbd/runs/1/nodes/100/log/?start=0&download=true' > pr5711-e2e.log
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 4236k    0 4236k    0     0  2818k      0 --:--:--  0:00:01 --:--:-- 2818k
$ grep STEP pr5711-e2e.log
...
```

Greenpepper15:
Yes when I look at the [test output](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.34-rbd/runs/1/nodes/100/log/?start=0) I also cannot find the output for my tests. At least I saw that the previous encryption tests succeeded. 

Does anyone know the reason? On my debug cluster I tested with this test configuration and it worked for me. 

```
# inside the /e2e directory 
go test . --test-cephfs=false --test-rbd=true --upgrade-testing=false --deploy-rbd=true --deploy-cephfs=false  --timeout=20m --deploy-timeout=20
```

Or it is because of the merge conflict that is yet to be resolved?

nixpanic:
> Does anyone know the reason? On my debug cluster I tested with this test configuration and it worked for me.

Maybe the CI job fails to checkout your  change because there is a merge conflict?

From https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e-helm_k8s-1.34-rbd/detail/mini-e2e-helm_k8s-1.34-rbd/1/pipeline/26 :

```
+ cd /home/jenkins/build/ceph-csi
+ git fetch origin pull/5711/merge
From https://github.com/ceph/ceph-csi
 * branch                refs/pull/5711/merge -> FETCH_HEAD
+ git checkout -b pull/5711/merge FETCH_HEAD
Switched to a new branch 'pull/5711/merge'
```

When I execute the same steps, I also do not have your changes to the `./e2e/` files when I go through `git log --stat`.

Greenpepper15:
```
<<<<<<< feature/5686/rbd-configurable-cipher
	golang.org/x/net v0.45.0 // indirect
=======
	golang.org/x/net v0.46.0 // indirect
>>>>>>> devel
```
This is only conflict in the merge conflict. I guess if I choose my version it will run without any issues. But we don't want a downgrade version on devel. 

How about I choose my changes for now and later somebody else calls `go mod tidy` and updates the corresponding files to resolve this issue?

nixpanic:
I think the issue is that there is a merge conflict. It probably does not matter how you address it. Most practical would be to rebase on top of the latest `devel` branch, and use versions from there. That will be needed later on in any case too.

Once the conflict is resolved, try to leave the `/test ...` message in a comment (only that single line in the comment), maybe you are allowed to start the CI job yourself too.

Greenpepper15:
The test failed with this error 

  E1110 20:00:02.940630   35903 crypto.go:238] ID: 244 Req-ID: 0001-0024-3674dcb1-48bf-45a6-8638-8fbb36b033b4-0000000000000004-c95798e0-5f4e-4956-9b64-da1138bba3e3 failed to encrypt device "/dev/rbd0" with LUKS (an error (exit status 1) occurred while running cryptsetup args: [-q luksFormat --type luks2 --hash sha256 --cipher aegis128-random --integrity aead --key-size 128 --luks2-metadata-size 4096k --luks2-keyslots-size 8192k --pbkdf-memory 32768 /dev/rbd0 -d -]): **_Cipher aegis128-random (key size 128 bits) is not available._**

This happens when the aegis128 kernel module is not loaded. I will temporarily remove this test for tonight and run the tests again to see the other tests do not fail. 

Could someone do a `sudo modprobe aegis128` on the worker node(s)?

nixpanic:
> Could someone do a `sudo modprobe aegis128` on the worker node(s)?

There is a `Modprobe()` helper at `internal/util/kmod/modprobe.go`, the node-plugin should be able to load the module that way. Ideally `cryptsetup` loads the module automatically... But that seems not be be done, so you have to think of a way to detect what modules are required for a cipher :thinking:

internal/rbd/nodeserver.go
Greenpepper15: Yes exactly.

internal/util/cryptsetup/cryptsetup.go
Greenpepper15: Hey can you check out my recommendation system? 
It is implemented in the `GetRecommendation()` function in cryptsetup.go. 

I would be especially interested in feedback on how I report the recommendation to users. I report it via the UsefulLog function. You guys have other ideas? 

```go
// EncryptVolume encrypts provided device with LUKS.
func EncryptVolume(ctx context.Context, devicePath, passphrase string, cipher *cryptsetup.EncryptionOptions) error {
	log.DebugLog(ctx, "Encrypting device %q	 with LUKS", devicePath)
	if cipher != nil {
		recommendation := cryptsetup.GetRecommendation(*cipher)
		log.UsefulLog(ctx, "current encryption configuration is: %v", recommendation)
	}
	_, stdErr, err := luks.Format(devicePath, passphrase, cipher)
...
```

e2e/utils.go
Greenpepper15: This parsing code now is in the cryptsetup. go file. I also added a few tests for it.

internal/rbd/encryption.go
```diff
@@ -615,3 +620,19 @@ func (rv *rbdVolume) RotateEncryptionKey(ctx context.Context) error {
 	// Return error accordingly.
 	return nil
 }
+
+// Loads the necessary cipher modules to enable encryption with specific ciphers.
+func SetEncryptionFeatures() {
+	_, err := os.Stat("/sys/module/" + aegisKernelModule)
```
nixpanic: This `stat` isn't needed, `kmod.Modprobe()` does that too already.

Is there a reasonable way to only load the module when needed?

nixpanic:
From [the logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e-helm_k8s-1.34-rbd/runs/5/nodes/100/log/?start=0).

This is unfortunate:

```
  E1114 13:14:34.421166   37619 crypto.go:242] ID: 331 Req-ID: 0001-0024-b7e4c8eb-90e2-4e53-8130-485b4b948c6b-0000000000000004-7a021e5e-78bd-4878-8968-670ce9f077a8 failed to encrypt device "/dev/rbd0" with LUKS (an error (exit status 1) occurred while running cryptsetup args: [-q luksFormat --type luks2 --hash sha256 --cipher aegis128-random --integrity aead --key-size 128 --luks2-metadata-size 4096k --luks2-keyslots-size 8192k --pbkdf-memory 32768 /dev/rbd0 -d -]): Cipher aegis128-random (key size 128 bits) is not available.
```

The e2e tests run on a CentOS Stream 9 system (minikube with podman backend, not it's own VM). Is the kernel module not available on this CentOS version?

```
  W1114 12:48:18.857770   37619 modprobe.go:49] modprobe of kernel module "aegis128" failed (an error (exit status 1) occurred while running modprobe args: [aegis128]): modprobe: FATAL: Module aegis128 not found in directory /lib/modules/5.14.0-612.el9.x86_64
  E1114 12:48:18.857796   37619 encryption.go:633] modprobe of kernel module "aegis128" failed (an error (exit status 1) occurred while running modprobe args: [aegis128]): modprobe: FATAL: Module aegis128 not found in directory /lib/modules/5.14.0-612.el9.x86_64
  I1114 12:48:18.857805   37619 encryption.go:635] AEGIS-128 kernel module loaded.
  I1114 12:48:18.857812   37619 encryption.go:637] AEGIS-128 kernel module already loaded.
```

nixpanic:
> The e2e tests run on a CentOS Stream 9 system (minikube with podman backend, not it's own VM). Is the kernel module not available on this CentOS version?

To answer my own question: No, the aegis128 kernel module is not available on CentOS Stream 9 or even on CentOS Stream 10 (not in the `kernel-modules-core` and `kernel-modules-extras` packages).

It won't be trivial to add e2e testing for this :disappointed:

Greenpepper15:
Okay that is good to know. 
**Then I will drop the support for the aegis cipher**. As far as know from my research using this mode is experimental anyway in DmIntegrity and not recommend. 

The problem with using aegis or something like AES-GCM is that it is not possible stop verifying the authentication tags. Once you have this cipher to MUST verify the integrity when decrypting (because the decryption process is tied to integrity verification). 
When using something like aes-xts with an HMAC the HMAC verification can be skipped in case of an error. This is why these kind of compound methods are recommended. 

The "only" upside to aegis is that it is typically faster than doing something like an aes-xts with an HMAC. 
Or should I leave the aegis option but we do not test for it?

internal/rbd/encryption_test.go
Greenpepper15: I rewrote this test suite. You can check it out.

Greenpepper15:
Hey guys please give me some comment if we should drop aegis128 support or not. That would be at last thing (at least from my side) that I need to implement.

And please give some feedback if the existing e2e test suites are enough for you. If not please specify a test scenario you would like to see then I will add it.

Madhu-1:
> Hey guys please give me some comment if we should drop aegis128 support or not. That would be at last thing (at least from my side) that I need to implement.
> 

@Greenpepper15 yes lets drop it, we can introduce it in new PR if required.
 
> And please give some feedback if the existing e2e test suites are enough for you. If not please specify a test scenario you would like to see then I will add it.

Will take sometime and review it soon. If its using existing test and doing all the tests like data consistency/snapshot/clone/resize etc we are good.

Madhu-1:
@Greenpepper15 it looks there are commints in this PR not related to this feature

Greenpepper15:
Yes which one do you mean? 

As far as I know it should only be [this one](https://github.com/ceph/ceph-csi/pull/5711/commits/a1bee63e3e7a85a716783f6ce8154597cde0b8f3). Should I drop this and open a new PR or should I edit the PR?

nixpanic:
> Or should I leave the aegis option but we do not test for it?

You mention it is experimental, so having the option is nice for users to try it out. No CI test for it seems needed, as CentOS does not provide the kernel module (yet). You can keep the test, but disable it (return immediately inside the `By()`) and leave a `// TODO: ..` comment about it.

Greenpepper15:
@nixpanic I already removed the tests. I think removing aegis128 like that is fine for now.

Greenpepper15:
Hey guys the [build for arm64](https://github.com/ceph/ceph-csi/actions/runs/19473993705/job/55794816988) cannot finish. You guys have any clue why it cannot finish the build?

Greenpepper15:
Hey is anyone working on reviewing the code? 
Otherwise is there anything I can provide to help you guys?

Madhu-1:
@Greenpepper15 maintainers are little busy with other priority work, we will get back to this one, at the meantime this PR contains 41 commits (there are extra commits) not related to this PR, can you please rebase this PR and keep only relevant commits?

Greenpepper15:
Rebase done.

e2e/rbd_helper.go
```diff
@@ -294,6 +294,69 @@ func getImageInfoFromPVC(pvcNamespace, pvcName string, f *framework.Framework) (
 	return imageData, nil
 }
 
+func getEncryptionOptionsFromPV(pvcNamespace, pvcName string, f *framework.Framework) (options *cryptsetup.EncryptionOptions, err error) {
```
Madhu-1: lets not use the named return

e2e/rbd_helper.go
```diff
@@ -294,6 +294,69 @@ func getImageInfoFromPVC(pvcNamespace, pvcName string, f *framework.Framework) (
...
+	c := f.ClientSet.CoreV1()
+	pvc, err := c.PersistentVolumeClaims(pvcNamespace).Get(context.TODO(), pvcName, metav1.GetOptions{})
+	if err != nil {
+		return options, fmt.Errorf("failed to get pvc: %w", err)
+	}
+	pv, err := c.PersistentVolumes().Get(context.TODO(), pvc.Spec.VolumeName, metav1.GetOptions{})
+	if err != nil {
+		return options, fmt.Errorf("failed to get pv: %w", err)
+	}
+	if pv.Spec.CSI.VolumeAttributes == nil {
```
Madhu-1: check for CSI as its a pointer before accessing it

internal/util/crypto.go
Madhu-1: Can we add a comment for the exported function

internal/util/cryptsetup/crypsetup_test.go
```diff
@@ -0,0 +1,363 @@
+/*
+Copyright 2019 The Ceph-CSI Authors.
```
Madhu-1: 2025

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -52,18 +53,479 @@ const (
...
+	cipher        string
+	keysize       uint
+	integrityMode *string
+	// if this is set we have to subtract it from the
+	// keysize to get actual key size used for encryption
+	intgrityKeySize *uint
+	sectorSize      *uint
+}
+
+func (l *LuksStatus) Cipher() string {
```
Madhu-1, line 118: Add comment for all the exported functions in this file as well

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -52,18 +53,479 @@ const (
...
+
+func minLevel(l1, l2 RecommendationLevel) RecommendationLevel {
+	if levelScores[l1] < levelScores[l2] {
+		return l1
+	}
+
+	return l2
+}
+
+// CheckCombination valides the (cipher, key size, integrity mode) tuple.
```
Madhu-1: start the comment with the function name

internal/util/cryptsetup/cryptsetup.go
```diff
@@ -52,18 +53,479 @@ const (
...
+}
+
+// ParseLuksStatus parses parts of the output of a "cryptsetup luksStatus <device>" command.
+func ParseLuksStatus(dump string) (*LuksStatus, error) {
+	var (
+		luksStatus *LuksStatus
+		err        error
+	)
+	scanner := bufio.NewScanner(strings.NewReader(dump))
+	initLuksStatus := func() {
```
Madhu-1, line 457: instead of this helper cant we dont `luksStatus := &LuksStatus{}` during variable declaration?

Madhu-1 (commented):
Its good to see lot of E2E tests for this one 🎉 small nits

Can we also have a followup for below block the PVC resize when its using integrity mode

what happens when someone try to clone or restore the PVC to higher value, we need to block it as well right?

internal/util/cryptsetup/cryptsetup.go
Greenpepper15: Because to check if any parsing happened at all I verify if the luksStatus variable is still nil.

```go
if luksStatus == nil {
    return nil, fmt.Errorf("parsing failed with status: %s", dump)
}
```

If luksStatus is nil then parsing failed.

internal/util/cryptsetup/cryptsetup.go
Greenpepper15: Addressed please take a look.

Greenpepper15:
> Its good to see lot of E2E tests for this one 🎉 small nits
> 
> Can we also have a followup for below block the PVC resize when its using integrity mode
> 
> what happens when someone try to clone or restore the PVC to higher value, we need to block it as well right?

Let's say it like that with this 
```go
// resizeNodeStagePath resizes the device if its encrypted and it also resizes
// the stagingTargetPath if filesystem needs resize.
func resizeNodeStagePath(...)
     ...
	if (isBlock && !transaction.isBlockEncrypted) || transaction.isIntegrityProtected {
		log.WarningLog(ctx, "resizing not possible or not necessary with selected options")

		return nil
	}
     ...
```
catches all the resize tries as far as I saw from my testing (I also looked at the lone or restore the PVC to higher value case).

The question is if we should stop a resize under these configuration sooner (on the controllerServer) or if we go into an error state if a resize happens under these circumstances. What do you think?

Madhu-1:
> > Its good to see lot of E2E tests for this one 🎉 small nits
> > Can we also have a followup for below block the PVC resize when its using integrity mode
> > what happens when someone try to clone or restore the PVC to higher value, we need to block it as well right?
> 
> Let's say it like that with this
> 
> ```go
> // resizeNodeStagePath resizes the device if its encrypted and it also resizes
> // the stagingTargetPath if filesystem needs resize.
> func resizeNodeStagePath(...)
>      ...
> 	if (isBlock && !transaction.isBlockEncrypted) || transaction.isIntegrityProtected {
> 		log.WarningLog(ctx, "resizing not possible or not necessary with selected options")
> 
> 		return nil
> 	}
>      ...
> ```
> 
> catches all the resize tries as far as I saw from my testing (I also looked at the lone or restore the PVC to higher value case).
> 
> The question is if we should stop a resize under these configuration sooner (on the controllerServer) or if we go into an error state if a resize happens under these circumstances. What do you think?

Catching early is better to give a nice user experience lets take it as a followup PR.

@Greenpepper15 we dont squash and merge all commits into 1, can you please cleanup the commits, there are few commits that are addressing the review comments and also few changes added in one commit and removed it another commit.

@nixpanic @Rakshith-R PTAL

Greenpepper15:
I cleaned up the commit history. Is there anything I overlooked?

Madhu-1 (dismissed):

Madhu-1:
@Greenpepper15 can you please rebase the PR

Greenpepper15:
> @Greenpepper15 can you please rebase the PR

Done.

Madhu-1 (dismissed):

nixpanic (dismissed):

nixpanic:
@mergifyio queue

Thanks for being persistent, @Greenpepper15!

Greenpepper15:
```
 Summarizing 1 Failure:
  [FAIL] External Storage [Driver: cephfs.csi.ceph.com] [Testpattern: Dynamic PV (default fs)] provisioning [It] should provision storage with pvc data source in parallel [Slow] [Slow]
  k8s.io/kubernetes/test/e2e/framework/volume/fixtures.go:604

Ran 34 of 7041 Specs in 1410.754 seconds
FAIL! -- 33 Passed | 1 Failed | 0 Pending | 7007 Skipped
--- FAIL: TestE2E (1410.95s)
FAIL
```
Oh do I need to review something or does this have to do with the CI process? I don't think I did any changes that would impact Cephfs.

Madhu-1:
@mergifyio requeue

Madhu-1:
> ```
>  Summarizing 1 Failure:
>   [FAIL] External Storage [Driver: cephfs.csi.ceph.com] [Testpattern: Dynamic PV (default fs)] provisioning [It] should provision storage with pvc data source in parallel [Slow] [Slow]
>   k8s.io/kubernetes/test/e2e/framework/volume/fixtures.go:604
> 
> Ran 34 of 7041 Specs in 1410.754 seconds
> FAIL! -- 33 Passed | 1 Failed | 0 Pending | 7007 Skipped
> --- FAIL: TestE2E (1410.95s)
> FAIL
> ```
> 
> Oh do I need to review something or does this have to do with the CI process? I don't think I did any changes that would impact Cephfs.

seems flaky, restarted the CI

Madhu-1:
@mergifyio queue

Greenpepper15:
> VolumeStatus:nil,},} to be in Bound state
>   I1211 11:33:43.046505 76117 pvc.go:63] waiting for PVC block-pvc-clone (0 seconds elapsed)
>   I1211 11:33:43.062861 76117 pvc.go:85] PVC Event: Provisioning - External provisioner is provisioning volume for claim "rbd-5848/block-pvc-clone"
>   I1211 11:33:43.062972 76117 pvc.go:85] PVC block-pvc-clone Event: ExternalProvisioning - Waiting for a volume to be created either by the external provisioner 'rbd.csi.ceph.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
>   I1211 11:33:43.062985 76117 pvc.go:85] PVC block-pvc-clone Event: ProvisioningSucceeded - Successfully provisioned volume pvc-ba74e087-8c1b-41e7-b019-87da0f7c24bc
>   I1211 11:33:45.047502 76117 pvc.go:63] waiting for PVC block-pvc-clone (2 seconds elapsed)
>   I1211 11:33:45.066074 76117 pvc.go:85] PVC block-pvc-clone Event: Provisioning - External provisioner is provisioning volume for claim "rbd-5848/block-pvc-clone"
>   I1211 11:33:45.066110 76117 pvc.go:85] PVC block-pvc-clone Event: ExternalProvisioning - Waiting for a volume to be created either by the external provisioner 'rbd.csi.ceph.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
>   I1211 11:33:45.066120 76117 pvc.go:85] PVC block-pvc-clone Event: ProvisioningSucceeded - Successfully provisioned volume pvc-ba74e087-8c1b-41e7-b019-87da0f7c24bc
>   I1211 11:33:45.066127 76117 pvc.go:85] PVC block-pvc-clone Event: ExternalProvisioning - Waiting for a volume to be created either by the external provisioner 'rbd.csi.ceph.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
>   I1211 11:33:45.066135 76117 pvc.go:85] PVC block-pvc-clone Event: Provisioning - External provisioner is provisioning volume for claim "rbd-5848/block-pvc-clone"
>   I1211 11:33:47.048192 76117 pvc.go:63] waiting for PVC block-pvc-clone (4 seconds elapsed)
>   I1211 11:33:47.052092 76117 pv.go:468] Waiting for PV pvc-61a82984-4d74-4846-9ef0-06a19c48c7d1 to bind to PVC block-pvc-clone
>   I1211 11:33:47.052133 76117 pv.go:790] Waiting up to timeout=10m0s for PersistentVolumeClaims [block-pvc-clone] to have phase Bound
>   I1211 11:33:47.053532 76117 pv.go:801] PersistentVolumeClaim block-pvc-clone found and phase=Bound (1.380058ms)
>   I1211 11:33:47.053558 76117 pv.go:763] Waiting up to 10m0s for PersistentVolume pvc-61a82984-4d74-4846-9ef0-06a19c48c7d1 to have phase Bound
>   I1211 11:33:47.055061 76117 pv.go:771] PersistentVolume pvc-61a82984-4d74-4846-9ef0-06a19c48c7d1 found and phase=Bound (1.488699ms)
>   I1211 11:33:47.062877   76117 warnings.go:110] "Warning: would violate PodSecurity \"restricted:latest\": allowPrivilegeEscalation != false (container \"centos\" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container \"centos\" must set securityContext.capabilities.drop=[\"ALL\"]), runAsNonRoot != true (pod or container \"centos\" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container \"centos\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")"
>   I1211 11:33:47.062941 76117 pod.go:377] Waiting up to 10m0s for pod-with-block-volume-clone to be in Running state
>   I1211 11:33:53.068335 76117 pod.go:440] Waiting for pod pod-with-raw-block-volume to be deleted
> Sending interrupt signal to process
> script returned exit code 255

okay this time it also seems be a time sensitive failure. Only this test uses cloning "validate encrypted (with selected cipher and integrity mode) PVC Clone chained with depth 2". Was it the cause of the failure?

Madhu-1:
@Greenpepper15 this might be related to the new update in the sidecar which considers the invalid_argument and try to requeue it after 5 minutes, @iPraveenParihar PTAL, we might need to add `--retry-interval-max=30s` as a new PR

Madhu-1:
Discard my above comments, i thought its related to the PVC but it looks like app, @Greenpepper15 we have not seen the issue on any other CI runs, it might be  problem with this PR itself.

Madhu-1:
We need to increase the test time it looks like 2:30 min is not enough :(

Madhu-1:
Add a new commit to increase the timeout https://github.com/ceph/ceph-csi/blob/468b76cba2b75258b1e09f1cead0f343d58fa7d6/build.env#L74

Greenpepper15:
Done. Let's try it out. 

But why do you think this issue arises? When I manually trigger the RBD e2e tests this does not seem to happen. 

Since AEAD encryption with dmIntegrity and DmCrypt consumes more resources I guess increasing the timeout might help. 

Otherwise maybe the resource freeing is buggy on larger setups. I will see after we run the tests again.

Greenpepper15:
So what is the current state of things? Should we wait to proceed till the holidays pass?

Madhu-1 (approved):

nixpanic (approved):
:snowman_with_snow:

nixpanic:
@mergifyio queue

Greenpepper15:
Yay thanks for the help people!
I look forward to contributing to the project again.

Issue #5694, opened 2025-10-21
nvmeof: QoS Support for NVMe-oF CSI Driver


gadididi:
### Overview
Add QoS (Quality of Service) support for NVMe-oF namespaces, allowing users to control IOPS and bandwidth limits both at volume creation and during runtime.

#### 1. QoS at Volume Creation (StorageClass)
Set initial QoS limits via StorageClass parameters. If omitted, namespaces remain unlimited.

**Example StorageClass:**
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-nvmeof-standard
provisioner: nvmeof.csi.ceph.com
parameters:
  pool: mypool
  # Optional QoS parameters
  qosRwIopsPerSecond: "10000"
  qosRwMegabytesPerSecond: "100"
  qosReadMegabytesPerSecond: "150"
  qosWriteMegabytesPerSecond: "50"
```

**Implementation:** Modify `ControllerCreateVolume()` to:
- Parse QoS parameters from StorageClass
- Call `ns set_qos` API after namespace creation (if parameters exist)
- Handle missing parameters gracefully (no QoS = unlimited)

#### 2. Runtime QoS Modification (VolumeAttributesClass)
Enable QoS changes on existing volumes without recreation using CSI `ControllerModifyVolume()`.

**Example VolumeAttributesClass:**
```yaml
apiVersion: storage.k8s.io/v1beta1
kind: VolumeAttributesClass
metadata:
  name: high-performance
driverName: nvmeof.csi.ceph.com
parameters:
  qosRwIopsPerSecond: "50000"
  qosRwMegabytesPerSecond: "500"
```
Now you want to hook the pvc with the VAC, you should do: 
**Usage:**
```bash
# Apply QoS to existing PVC
kubectl patch pvc my-pvc -p '{"spec":{"volumeAttributesClassName":"high-performance"}}'
```

**Implementation:** Add `ControllerModifyVolume()` RPC to:
- Parse QoS parameters from VolumeAttributesClass
- Call `ns set_qos` via GRPC to the GW
- Deploy `csi-resizer` sidecar to monitor VAC changes

### Supported QoS Parameters
All parameters map directly to NVMe-oF gateway `ns set_qos` command:
- `qosRwIopsPerSecond` - R/W IOPS limit (0 = unlimited)
- `qosRwMegabytesPerSecond` - R/W bandwidth limit
[...]

PR #5714, opened 2025-10-29
nvmeof: add qos ability

gadididi:

Adds QoS (Quality of Service) support for NVMe-oF CSI driver, allowing users to set IOPS and bandwidth limits on volumes at creation time and modify them dynamically at runtime.

**New StorageClass parameters:**
- `rwIosPerSecond` - Read/Write IOPS limit
- `rwMbytesPerSecond` - Read/Write bandwidth limit (MB/s)
- `rMbytesPerSecond` - Read bandwidth limit (MB/s)
- `wMbytesPerSecond` - Write bandwidth limit (MB/s)

**Key decisions:**
- Uses NVMe-oF-specific parameter names (not RBD QoS keys) because the architectures are different: details [here](https://github.com/ceph/ceph-csi/issues/5694
)
- Implements `ControllerModifyVolume()` RPC for runtime QoS changes via VolumeAttributesClass (GA at K8s 1.34+ 
you can read more about it here: [volume-attributes-classes)](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/)
- Adds stub `ControllerExpandVolume()` with `EXPAND_VOLUME` capability to satisfy csi-resizer requirements (csi-resizer needs this capability to start, even when only using it for modify operations)
- Uses `GetControllerPublishSecretRef()` for credential retrieval in `ControllerModifyVolume()` as a temporary solution (TODO: implement `GetControllerExpandSecretRef()` with proper NVMeoF support in ceph-csi-config)

**Credential handling note:**
Since CSI spec doesn't define `csi.storage.k8s.io/controller-modify-secret-name` parameter and csi-resizer doesn't pass secrets to `ControllerModifyVolume()` according to [storageclass-secrets,](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#storageclass-secrets) . The implementation retrieves credentials from ceph-csi-config ConfigMap using the existing `controllerPublishSecretRef` field. This follows the same pattern as other CSI operations.

**Example workflow:**

1. Create StorageClass with:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-nvmeof-qos
provisioner: nvmeof.csi.ceph.com
parameters:
  clusterID: "cluster-1"
  pool: "mypool"
  # ... other params
```

2. Create default QoS

```yaml 
apiVersion: storage.k8s.io/v1
kind: VolumeAttributesClass
metadata:
  name: default
driverName: nvmeof.csi.ceph.com
parameters:
  rwIopsPerSecond: "5000"
```

3. Create PVC (gets 10000 IOPS, 100 MB/s):
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
  storageClassName: csi-nvmeof-qos
  volumeAttributesClassName: default  # ← Start with default (5000 IOPS)
```

4. Later, define VolumeAttributesClass for runtime modification:
```yaml
apiVersion: storage.k8s.io/v1alpha1
kind: VolumeAttributesClass
metadata:
  name: high-performance
driverName: nvmeof.csi.ceph.com
parameters:
  nvmeof_rw_ios_per_second: "50000"
  nvmeof_rw_mbytes_per_second: "500"
```

5. Modify existing PVC (triggers ControllerModifyVolume):
```bash
kubectl patch pvc my-pvc -p '{"spec":{"volumeAttributesClassName":"high-performance"}}'
```

Result: Volume QoS updated to 50000 IOPS, 500 MB/s without recreation or downtime.

NOTE: 
- Adds csi-resizer sidecar to controller deployment according to this: [resizer](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/#resizer)

## Related issues ##

https://github.com/ceph/ceph-csi/issues/5694

## Future concerns ##
- Add proper NVMeoF struct to ceph-csi-config-map ClusterInfo with dedicated `controllerExpandSecretRef` field **OR** Add `controllerExpandSecretRef` for the existing `RBD` struct. (because NVMeoF is just a wrapper to RBD). 
- Implement `GetControllerExpandSecretRef()` to replace temporary use of `GetControllerPublishSecretRef()`

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/nvmeof/controller/controllerserver.go
+++ b/internal/nvmeof/controller/controllerserver.go
@@ -312,6 +312,22 @@ func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
 	if err != nil {
 		return fmt.Errorf("invalid listeners parameter: %w", err)
 	}
+	// Validate QoS parameters - cannot mix RBD and NVMe-oF QoS
+	mutableParams := req.GetMutableParameters()
+
+	// check for RBD QoS parameters in both params and mutableParams
+	if hasRBDQoS := rbd.HasQoSParams(params); hasRBDQoS {
+		return errors.New("setting RBD QoS parameters on NVMe-oF volumes is not supported")
+	}
+	if hasRBDQoS := rbd.HasQoSParams(mutableParams); hasRBDQoS {
+		return errors.New("setting RBD QoS parameters on NVMe-oF volumes is not supported")
+	}
+
+	// It take the mutableParams value from the volumeAttributesClassName in the PersistentVolumeClaim yaml.
+	_, err = parseQoSParameters(mutableParams)
+	if err != nil {
+		return fmt.Errorf("invalid NVMe-oF QoS parameters: %w", err)
+	}
 
 	return nil
 }
@@ -349,6 +365,44 @@ func parseListeners(listenersJSON string) ([]nvmeof.ListenerDetails, error) {
 	}
 
 	return listeners, nil
+}
+
+// parseQoSParameters extracts and parses QoS parameters from the given map.
+func parseQoSParameters(params map[string]string) (*nvmeof.NVMeoFQosVolume, error) {
+	qos := &nvmeof.NVMeoFQosVolume{}
+	hasAnyQoS := false
+
+	parseParam := func(key, name string, dest **uint64) error {
+		if val, exists := params[key]; exists && val != "" {
+			parsed, err := strconv.ParseUint(val, 10, 64)
+			if err != nil {
+				return fmt.Errorf("invalid %s: %w", name, err)
+			}
+			*dest = &parsed
+			hasAnyQoS = true
+		}
+
+		return nil
+	}
+
+	if err := parseParam(nvmeof.RwIosPerSecond, nvmeof.RwIosPerSecond, &qos.RwIosPerSecond); err != nil {
+		return nil, err
+	}
+	if err := parseParam(nvmeof.RwMbytesPerSecond, nvmeof.RwMbytesPerSecond, &qos.RwMbytesPerSecond); err != nil {
+		return nil, err
+	}
+	if err := parseParam(nvmeof.RMbytesPerSecond, nvmeof.RMbytesPerSecond, &qos.RMbytesPerSecond); err != nil {
+		return nil, err
+	}
+	if err := parseParam(nvmeof.WMbytesPerSecond, nvmeof.WMbytesPerSecond, &qos.WMbytesPerSecond); err != nil {
+		return nil, err
+	}
+
+	if !hasAnyQoS {
+		return nil, nil
+	}
+
+	return qos, nil
 }
 
 // ensureSubsystem checks if the subsystem exists, and creates it if not.
@@ -434,7 +488,6 @@ func cleanupEmptySubsystem(
 }
 
 // createNVMeoFResources sets up the NVMe-oF resources for the given RBD volume.
-// TODO - need to support multiple listeners.
 func (cs *Server) createNVMeoFResources(
 	ctx context.Context,
 	req *csi.CreateVolumeRequest,
@@ -464,6 +517,16 @@ func (cs *Server) createNVMeoFResources(
 			Address: params["nvmeofGatewayAddress"],
 			Port:    uint32(nvmeofGatewayPort),
 		},
+	}
+	// extract Qos parameters if any
+	mutableParams := req.GetMutableParameters()
+	// It take the mutableParams value from the volumeAttributesClassName in the PersistentVolumeClaim yaml.
+	// We already verified in the validateCreateVolumeRequest that there is no RBD QoS
+	nvmeofQoS, err := parseQoSParameters(mutableParams)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to parse NVMe-oF QoS parameters: %v", err)
+
+		return nil, fmt.Errorf("failed to parse QoS parameters: %w", err)
 	}
 
 	// Step 2: Connect to gateway
@@ -504,6 +567,15 @@ func (cs *Server) createNVMeoFResources(
 	}
 	log.DebugLog(ctx, "Namespace created: %s/%s with NSID: %d", rbdPoolName, rbdImageName, nsid)
 	nvmeofData.NamespaceID = nsid
+
+	// Step 5: Set QoS limits if any
+	if nvmeofQoS != nil {
+		log.DebugLog(ctx, "Setting QoS limits: %s", nvmeofQoS)
+		if err := gateway.SetQoSLimitsForNamespace(ctx, nvmeofData.SubsystemNQN, nvmeofData.NamespaceID,
+			*nvmeofQoS); err != nil {
+			return nil, fmt.Errorf("setting QoS limits failed: %w", err)
+		}
+	}
 
 	uuid, err := gateway.GetUUIDBySubsystemAndNameSpaceID(ctx, nvmeofData.SubsystemNQN, nvmeofData.NamespaceID)
 	if err != nil {
--- a/internal/nvmeof/controller/controllerserver.go
+++ b/internal/nvmeof/controller/controllerserver.go
@@ -29,6 +29,7 @@ import (
 
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
 	"github.com/ceph/ceph-csi/internal/nvmeof"
+	nvmeoferrors "github.com/ceph/ceph-csi/internal/nvmeof/errors"
 	"github.com/ceph/ceph-csi/internal/rbd"
 	rbdutil "github.com/ceph/ceph-csi/internal/rbd"
 	rbddriver "github.com/ceph/ceph-csi/internal/rbd/driver"
@@ -293,6 +294,42 @@ func (cs *Server) ControllerUnpublishVolume(
 	return &csi.ControllerUnpublishVolumeResponse{}, nil
 }
 
+// ControllerModifyVolume modifies the volume's QoS parameters.
+func (cs *Server) ControllerModifyVolume(
+	ctx context.Context,
+	req *csi.ControllerModifyVolumeRequest,
+) (*csi.ControllerModifyVolumeResponse, error) {
+	volumeID := req.GetVolumeId()
+	params := req.GetMutableParameters()
+
+	// Step 1: Acquire volume lock
+	if acquired := cs.volumeLocks.TryAcquire(volumeID); !acquired {
+		log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volumeID)
+
+		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer cs.volumeLocks.Release(volumeID)
+
+	// Step 2: Parse QoS parameters from mutable_parameters
+	hasRBDQoS := rbd.HasQoSParams(params)
+	if hasRBDQoS {
+		log.ErrorLog(ctx, "Cannot set RBD QoS parameters on NVMe-oF volumes")
+
+		return nil, status.Error(codes.InvalidArgument, "cannot set RBD QoS parameters on NVMe-oF volumes")
+	}
+	nvmeofQoS, err := parseQoSParameters(params)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to parse NVMe-oF QoS parameters: %v", err)
+
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse QoS parameters: %v", err)
+	}
+	if nvmeofQoS != nil {
+		return cs.modifyNVMeoFQoS(ctx, req, nvmeofQoS)
+	}
+
+	return &csi.ControllerModifyVolumeResponse{}, nil
+}
+
 // validateCreateVolumeRequest validates the incoming request for nvmeof.
 // the rest of the parameters are validated by RBD.
 func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
@@ -403,6 +440,85 @@ func parseQoSParameters(params map[string]string) (*nvmeof.NVMeoFQosVolume, erro
 	}
 
 	return qos, nil
+}
+
+// modifyNVMeoFQoS handles NVMe-oF gateway QoS modification.
+func (cs *Server) modifyNVMeoFQoS(
+	ctx context.Context,
+	req *csi.ControllerModifyVolumeRequest,
+	qos *nvmeof.NVMeoFQosVolume,
+) (*csi.ControllerModifyVolumeResponse, error) {
+	volumeID := req.GetVolumeId()
+
+	// Step 1: Get secrets
+
+	// Since ControllerModifyVolume doesn't receive volume context and dont have option to take secrets
+	// because there is no "csi.storage.k8s.io/controller-modify-secret-name" field in the SC !,
+	// the full solution for it is to use GetControllerExpandSecretRef but there is no such function yet.
+	// TODO: change the call to GetControllerExpandSecretRef once it is implemented.
+	secrets := req.GetSecrets()
+	if secrets == nil {
+		secretName, secretNamespace, err := util.GetControllerPublishSecretRef(volumeID, util.RBDType)
+		if err != nil {
+			log.ErrorLog(ctx, "Failed to get secret reference: %v", err)
+
+			return nil, status.Errorf(codes.Internal, "failed to get secret reference: %v", err)
+		}
+
+		secrets, err = k8s.GetSecret(secretName, secretNamespace)
+		if err != nil {
+			log.ErrorLog(ctx, "Failed to get secret from k8s: %v", err)
+
+			return nil, status.Errorf(codes.Internal, "failed to get secret: %v", err)
+		}
+	}
+
+	// Step 2: Get NVMe-oF metadata
+	nvmeofData, err := cs.getNVMeoFMetadata(ctx, secrets, volumeID)
+	if err != nil {
+		log.ErrorLog(ctx, "Failed to get NVMe-oF metadata: %v", err)
+
+		return nil, nvmeoferrors.ToGRPCError(err)
+	}
+
+	// Step 3: Connect to gateway
+	config := &nvmeof.GatewayConfig{
+		Address: nvmeofData.GatewayManagementInfo.Address,
+		Port:    nvmeofData.GatewayManagementInfo.Port,
+	}
+	gateway, err := connectGateway(ctx, config)
+	if err != nil {
+		log.ErrorLog(ctx, "Gateway connection failed: %v", err)
+
+		return nil, status.Errorf(codes.Unavailable, "gateway connection failed: %v", err)
+	}
+	defer func() {
+		if closeErr := gateway.Destroy(); closeErr != nil {
+			log.ErrorLog(ctx, "Failed to close gateway connection: %v", closeErr)
+		}
+	}()
+
+	// Step 4: Apply NVMe-oF QoS via gateway
+	log.DebugLog(ctx, "Setting QoS for subsystem=%s, nsid=%d", nvmeofData.SubsystemNQN, nvmeofData.NamespaceID)
+
+	err = gateway.SetQoSLimitsForNamespace(ctx, nvmeofData.SubsystemNQN, nvmeofData.NamespaceID, *qos)
+	if err != nil {
+		// Check if error is EEXIST (RBD QoS already set)
+		if errors.Is(err, nvmeoferrors.ErrRbdQoSExists) {
+			log.ErrorLog(ctx, "RBD QoS already configured on volume")
+
+			return nil, status.Error(codes.InvalidArgument,
+				"RBD QoS already configured on this volume, cannot set NVMe-oF gateway QoS")
+		}
+
+		log.ErrorLog(ctx, "Failed to set QoS limits: %v", err)
+
+		return nil, status.Errorf(codes.Internal, "failed to set QoS limits: %v", err)
+	}
+
+	log.DebugLog(ctx, "Successfully modified NVMe-oF QoS for volume %s", volumeID)
+
+	return &csi.ControllerModifyVolumeResponse{}, nil
 }
 
 // ensureSubsystem checks if the subsystem exists, and creates it if not.
--- a/internal/nvmeof/controller/controllerserver.go
+++ b/internal/nvmeof/controller/controllerserver.go
@@ -282,7 +282,7 @@ func (cs *Server) ControllerUnpublishVolume(
 	if err != nil {
 		log.ErrorLog(ctx, "failed to get NVMe-oF metadata for volumeID %s: %v", volumeID, err)
 
-		return nil, status.Errorf(codes.Internal, "failed to get NVMe-oF metadata: %v", err)
+		return nil, nvmeoferrors.ToGRPCError(err)
 	}
 
 	// Unpublish NVMe-oF resources
@@ -799,7 +799,8 @@ func (cs *Server) getNVMeoFMetadata(
 	// Get RBD volume
 	rbdVol, err := mgr.GetVolumeByID(ctx, volumeID)
 	if err != nil {
-		return nil, fmt.Errorf("failed to find volume with ID %q: %w", volumeID, err)
+		return nil, fmt.Errorf("%w: failed to find volume with ID %q: %w",
+			nvmeoferrors.ErrMetadataNotFound, volumeID, err)
 	}
 	defer rbdVol.Destroy(ctx)
 
@@ -820,10 +821,12 @@ func (cs *Server) getNVMeoFMetadata(
 	for _, key := range requiredKeys {
 		value, err := rbdVol.GetMetadata(key)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get %s: %w", key, err)
+			return nil, fmt.Errorf("%w: failed to get %s: %w",
+				nvmeoferrors.ErrMetadataNotFound, key, err)
 		}
 		if value == "" {
-			return nil, fmt.Errorf("metadata %s is empty", key)
+			return nil, fmt.Errorf("%w: metadata %s is empty",
+				nvmeoferrors.ErrMetadataNotFound, key)
 		}
 		metadata[key] = value
 	}
@@ -831,18 +834,21 @@ func (cs *Server) getNVMeoFMetadata(
 	// Parse namespace ID
 	nsid, err := strconv.ParseUint(metadata[toRBDMetadataKey(vcNamespaceID)], 10, 32)
 	if err != nil {
-		return nil, fmt.Errorf("invalid namespace ID: %w", err)
+		return nil, fmt.Errorf("%w: invalid namespace ID: %w",
+			nvmeoferrors.ErrMetadataCorrupted, err)
 	}
 
 	gatewayPort, err := strconv.ParseUint(metadata[toRBDMetadataKey(vcGatewayPort)], 10, 32)
 	if err != nil {
-		return nil, fmt.Errorf("invalid gateway port: %w", err)
+		return nil, fmt.Errorf("%w: invalid gateway port: %w",
+			nvmeoferrors.ErrMetadataCorrupted, err)
 	}
 
 	// Parse listeners from JSON
 	var listeners []nvmeof.ListenerDetails
 	if err := json.Unmarshal([]byte(metadata[toRBDMetadataKey(vcListeners)]), &listeners); err != nil {
-		return nil, fmt.Errorf("failed to parse listeners JSON: %w", err)
+		return nil, fmt.Errorf("%w: failed to parse listeners JSON: %w",
+			nvmeoferrors.ErrMetadataCorrupted, err)
 	}
 
 	// Construct NVMe-oF volume data
--- a/internal/nvmeof/controller/controllerserver.go
+++ b/internal/nvmeof/controller/controllerserver.go
@@ -330,6 +330,18 @@ func (cs *Server) ControllerModifyVolume(
 	return &csi.ControllerModifyVolumeResponse{}, nil
 }
 
+// ControllerExpandVolume handles volume expansion requests.
+// For now it only updates the capacity in the response as NVMe-oF
+// this must be added because ControllerModifyVolume requires the sidecar csi-resizer. and
+// csi-resizer searches for the capacity ControllerServiceCapability_RPC_EXPAND_VOLUME.
+// In the future, if NVMe-oF gateway supports volume expansion, the logic must be added here.
+func (cs *Server) ControllerExpandVolume(
+	ctx context.Context,
+	req *csi.ControllerExpandVolumeRequest,
+) (*csi.ControllerExpandVolumeResponse, error) {
+	return nil, status.Errorf(codes.Unimplemented, "ControllerExpandVolume is not implemented for NVMe-oF volumes")
+}
+
 // validateCreateVolumeRequest validates the incoming request for nvmeof.
 // the rest of the parameters are validated by RBD.
 func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
--- a/internal/nvmeof/nvmeof.go
+++ b/internal/nvmeof/nvmeof.go
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/protobuf/proto"
 
+	nvmeoferrors "github.com/ceph/ceph-csi/internal/nvmeof/errors"
 	"github.com/ceph/ceph-csi/internal/util/log"
 )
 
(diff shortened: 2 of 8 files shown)


internal/nvmeof/qos.go
```diff
@@ -0,0 +1,64 @@
...
+
+// NVMeoFQosVolume holds the QoS parameters for an NVMe-oF volume.
+type NVMeoFQosVolume struct {
+	RwIosPerSecond    *uint64
+	RwMbytesPerSecond *uint64
+	RMbytesPerSecond  *uint64
+	WMbytesPerSecond  *uint64
+}
+
+// QoS parameter keys that will be shown in the storage class.
```
nixpanic: nit: "storage class" is Kubernetes specific, this would work with any container orchistrator. Maybe mention the "ModifyVolume" CSI procedure too?

internal/nvmeof/qos.go
```diff
@@ -0,0 +1,64 @@
...
+package nvmeof
+
+import (
+	"fmt"
+	"strings"
+)
+
+// NVMeoFQosVolume holds the QoS parameters for an NVMe-oF volume.
+type NVMeoFQosVolume struct {
+	RwIosPerSecond    *uint64
```
nixpanic: Please comment each attribute with their non-abbreviated name.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -351,6 +351,44 @@ func parseListeners(listenersJSON string) ([]nvmeof.ListenerDetails, error) {
 	return listeners, nil
 }
 
+// parseQoSParameters extracts and parses QoS parameters from the given map.
+func parseQoSParameters(params map[string]string) (*nvmeof.NVMeoFQosVolume, error) {
```
nixpanic, line 420: Can you add unit tests for this function?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -293,6 +293,112 @@ func (cs *Server) ControllerUnpublishVolume(
...
+		log.DebugLog(ctx, "No QoS parameters provided for volume %s", volumeID)
+
+		return &csi.ControllerModifyVolumeResponse{}, nil
+	}
+	log.DebugLog(ctx, "Parsed QoS parameters: %s", qos)
+
+	// Step 3: Get NVMe-oF metadata for the volume
+
+	// Since ControllerModifyVolume doesn't receive volume context and dont have option to take secrets
+	// because there is no "csi.storage.k8s.io/controller-modify-secret-name" field in the SC !,
```
nixpanic: `ControllerModifyVolume` can have secrets according to the specification: https://github.com/container-storage-interface/spec/blob/master/spec.md#controllermodifyvolume

Do you mean that the resizer sidecar does not pass the secret from Kubernetes to the CSI-driver? Is there an issue for that at https://github.com/kubernetes-csi/external-resizer , or can we provide a PR for it?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -399,6 +399,21 @@ func (cs *Server) ControllerModifyVolume(
...
+// ControllerExpandVolume handles volume expansion requests.
+// For now it only updates the capacity in the response as NVMe-oF
+// this must be added because ControllerModifyVolume requires the sidecar csi-resizer. and
+// csi-resizer searches for the capacity ControllerServiceCapability_RPC_EXPAND_VOLUME.
+// In the future, if NVMe-oF gateway supports volume expansion, the logic must be added here.
+func (cs *Server) ControllerExpandVolume(
+	ctx context.Context,
+	req *csi.ControllerExpandVolumeRequest,
+) (*csi.ControllerExpandVolumeResponse, error) {
+	return &csi.ControllerExpandVolumeResponse{
```
nixpanic: instead of faking an action, can you return a gRPC UNIMPLEMENTED error?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -399,6 +399,21 @@ func (cs *Server) ControllerModifyVolume(
 	return &csi.ControllerModifyVolumeResponse{}, nil
 }
 
+// ControllerExpandVolume handles volume expansion requests.
+// For now it only updates the capacity in the response as NVMe-oF
+// this must be added because ControllerModifyVolume requires the sidecar csi-resizer. and
+// csi-resizer searches for the capacity ControllerServiceCapability_RPC_EXPAND_VOLUME.
+// In the future, if NVMe-oF gateway supports volume expansion, the logic must be added here.
```
nixpanic, line 337: This sounds like a shortcoming of the external-resizer sidecar. Only ModifyVolume should be allowed. But there seems to be a required check on volume resize: https://github.com/kubernetes-csi/external-resizer/blob/master/pkg/resizer/csi_resizer.go#L59

internal/nvmeof/controller/controllerserver.go
gadididi: Yes, I saw it has that, but when I ran it the secret was `nil`, so I thought maybe I need to add into StorageClass 

```yml
  csi.storage.k8s.io/controller-modify-secret-name: csi-rbd-secret
  csi.storage.k8s.io/controller-modify-secret-namespace: default
```

like I have for **`ControllerPublishVolume()`** 

```yml
  csi.storage.k8s.io/controller-publish-secret-name: csi-rbd-secret
  csi.storage.k8s.io/controller-publish-secret-namespace: default
```
but what I see here:
https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#storageclass-secrets
there is no for modify. maybe I am wrong, but I could not find any place that mentioned that..


So, I implemented like `ControllerUnPublishVolume()`  here:  https://github.com/ceph/ceph-csi/blob/f01ab138265cac37a9c6a9893daa7a0edb8663ec/internal/nvmeof/controller/controllerserver.go#L268

About the resizer, I am not sure how it should pass the secret. I will look at it and see if it is unimplemented\issue or something else..

nixpanic: Secrets are indeed not fetched, they are always `nil`:

https://github.com/kubernetes-csi/external-resizer/blob/cba05cb7cc4d06d7c06ae95dc8941fdcb75bd67e/pkg/modifier/csi_modifier.go#L84-L89

I do not think it is possible to get secrets as part of the `ControllerModifyVolume` call, so these need to be resolved through an other way (like the new default credentials).

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/qos.go
gadididi: fixed

internal/nvmeof/qos.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
nixpanic: maybe kubernetes-csi/external-resizer#544 is acceptable, would you be able to test with that container-image?

internal/nvmeof/controller/controllerserver.go
Madhu-1: @nixpanic i think we should open a tracker with the external-resizer as there could be drivers that support ModiyVolume but not Expand volume.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -293,6 +294,78 @@ func (cs *Server) ControllerUnpublishVolume(
...
 
+// ControllerModifyVolume modifies the volume's QoS parameters.
+func (cs *Server) ControllerModifyVolume(
+	ctx context.Context,
+	req *csi.ControllerModifyVolumeRequest,
+) (*csi.ControllerModifyVolumeResponse, error) {
+	volumeID := req.GetVolumeId()
+	params := req.GetMutableParameters()
+
+	log.DebugLog(ctx, "ControllerModifyVolume: volumeID=%s, params=%+v", volumeID, params)
```
Madhu-1: This should  be logged in the Server.go please add this new RPC to https://github.com/ceph/ceph-csi/blob/fd202c7cf1a9619d528351769c108a41b17480f5/internal/csi-common/utils.go#L210

internal/nvmeof/controller/controllerserver.go
```diff
@@ -293,6 +294,78 @@ func (cs *Server) ControllerUnpublishVolume(
...
+
+	case hasNVMeoFQoS:
+		// NVMe-oF QoS only
+		return cs.modifyNVMeoFQoS(ctx, req, nvmeofQoS)
+
+	default:
+		// Should never reach here
+		log.ErrorLog(ctx, "Unexpected QoS parameter state")
+
+		return nil, status.Error(codes.Internal, "unexpected QoS parameter state")
```
Madhu-1: invalid InvalidArgument?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -312,6 +385,17 @@ func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
...
+	hasRBDQoS := rbd.HasQoSParams(params)
+	nvmeofQoS, err := parseQoSParameters(params)
+	if err != nil {
+		return fmt.Errorf("invalid NVMe-oF QoS parameters: %w", err)
+	}
+	hasNVMeoFQoS := (nvmeofQoS != nil)
+
+	if hasRBDQoS && hasNVMeoFQoS {
+		return errors.New("cannot specify both RBD and NVMe-oF QoS parameters")
+	}
```
Madhu-1, line 379: These should be MutableParameters not the parameters during the CreateVolume isnt it?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -351,6 +435,131 @@ func parseListeners(listenersJSON string) ([]nvmeof.ListenerDetails, error) {
...
+		return nil, err
+	}
+	if err := parseParam(nvmeof.RMbytesPerSecond, nvmeof.RMbytesPerSecond, &qos.RMbytesPerSecond); err != nil {
+		return nil, err
+	}
+	if err := parseParam(nvmeof.WMbytesPerSecond, nvmeof.WMbytesPerSecond, &qos.WMbytesPerSecond); err != nil {
+		return nil, err
+	}
+
+	if !hasAnyQoS {
```
Madhu-1, line 450: we dont need this variable right? we can simply return qos which will be nil if nothing is set

internal/nvmeof/controller/controllerserver.go
```diff
@@ -351,6 +435,131 @@ func parseListeners(listenersJSON string) ([]nvmeof.ListenerDetails, error) {
...
+		}
+	}()
+
+	// Step 4: Acquire subsystem lock
+	if acquired := cs.subsystemLocks.TryAcquire(nvmeofData.SubsystemNQN); !acquired {
+		log.ErrorLog(ctx, "Subsystem %s is already locked", nvmeofData.SubsystemNQN)
+
+		return nil, status.Errorf(codes.Aborted, "subsystem %s operation already in progress", nvmeofData.SubsystemNQN)
+	}
+	defer cs.subsystemLocks.Release(nvmeofData.SubsystemNQN)
```
Madhu-1: would it make sense to acquire lock and than connect to sub system?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -351,6 +435,131 @@ func parseListeners(listenersJSON string) ([]nvmeof.ListenerDetails, error) {
...
+
+	// Step 5: Apply NVMe-oF QoS via gateway
+	log.DebugLog(ctx, "Setting QoS for subsystem=%s, nsid=%d", nvmeofData.SubsystemNQN, nvmeofData.NamespaceID)
+
+	err = gateway.SetQoSLimitsForNamespace(ctx, nvmeofData.SubsystemNQN, nvmeofData.NamespaceID, *qos)
+	if err != nil {
+		// Check if error is EEXIST (RBD QoS already set)
+		if errors.Is(err, os.ErrExist) {
```
Madhu-1: can we use cephCSI specific error (declare it in error.go) instead of using os.ExxExist?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -351,6 +435,131 @@ func parseListeners(listenersJSON string) ([]nvmeof.ListenerDetails, error) {
...
+			log.ErrorLog(ctx, "RBD QoS already configured on volume")
+
+			return nil, status.Error(codes.InvalidArgument,
```
Madhu-1, line 522: this need to be success when it already configured with same values if not it need to be changed right?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -465,6 +673,20 @@ func (cs *Server) createNVMeoFResources(
 			Port:    uint32(nvmeofGatewayPort),
 		},
 	}
+	// extract Qos parameters if any
+	nvmeofQoS, err := parseQoSParameters(params)
```
Madhu-1: this should be MutableParameters not the regular one

internal/nvmeof/controller/controllerserver.go
gadididi: @Madhu-1, I saw there is field named  `MutableParameters` in `CreateVolumeRequest` struct. 
but I saw in rbd code it took the rbd QoS parameters from `req.GetParameters()`
here:
https://github.com/ceph/ceph-csi/blob/fd202c7cf1a9619d528351769c108a41b17480f5/internal/rbd/controllerserver.go#L437
then the flow comes to here:
https://github.com/ceph/ceph-csi/blob/fd202c7cf1a9619d528351769c108a41b17480f5/internal/rbd/controllerserver.go#L823-L830

So, if I understand correctly the rbd's QoS params are located in the SC. I wonder if put the nvmeof's QoS params in SC as `MutableParameters` , but I decided to do like rbd..

 what do you think?

internal/nvmeof/controller/controllerserver.go
gadididi: same topic\question as here: https://github.com/ceph/ceph-csi/pull/5714/files/359a34bb20db086c5e6a19200c6e710a24c69655#r2509807057

internal/nvmeof/controller/controllerserver.go
gadididi: my bad, will fix it

internal/nvmeof/controller/controllerserver.go
gadididi: 🤔 
maybe I am missing something, but:
```go
qos := &nvmeof.NVMeoFQosVolume{}  // ← This creates a NON-NIL pointer to an empty struct (that contains 4 nil pointers)
```
so, I can remove the `hasAnyQoS` , but I still need to add to this function:
```go
	// Return nil if no QoS parameters were set
	if qos.RwIosPerSecond == nil && qos.RwMbytesPerSecond == nil &&
		qos.RMbytesPerSecond == nil && qos.WMbytesPerSecond == nil {
		return nil, nil
	}
```

internal/nvmeof/controller/controllerserver.go
gadididi: @Madhu-1 , the "already exists" means there is RBD QoS in that namespace(=rbd image), and if the user tries also set QoS for the nvmeof namespace (that effects on the spdk, not directly on the rbd image) it should fail.

the logic is:
```md
1. Parse mutable_parameters
2. Check if contains RBD QoS params? 
   - YES: 
     - Check if ALSO contains NVMe-oF QoS params?
       - YES: Return INVALID_ARGUMENT (can't have both)
       - NO: Call RBD ControllerModifyVolume, return
   - NO: Continue with NVMe-oF QoS
3. Apply NVMe-oF QoS via gateway
4. If gateway returns EEXIST: Return INVALID_ARGUMENT (RBD QoS already set)
5. If other gateway error: Return INTERNAL
```

the current RBD controller does not have   `ControllerModifyVolume` implementation. later in nvmeof `ControllerModifyVolume` we will call to rbd `ControllerModifyVolume` if only rbd QoS params are provided.

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: addressed

internal/nvmeof/controller/controllerserver.go
gadididi: Good catch!

internal/nvmeof/controller/controllerserver.go
nixpanic: Maybe the subsystem lock is not even needed here?

internal/nvmeof/qos.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// QoS parameter keys that user can set.
+// these parameters are used to configure the QoS limits for NVMe-oF volumes.
+// They correspond to the fields in NVMeoFQosVolume struct.
+const (
+	RwIosPerSecond    = "nvmeof_rw_ios_per_second"
+	RwMbytesPerSecond = "nvmeof_rw_mbytes_per_second"
+	RMbytesPerSecond  = "nvmeof_r_mbytes_per_second"
+	WMbytesPerSecond  = "nvmeof_w_mbytes_per_second"
```
nixpanic: note that we do not have any parameters in the StorageClass that contain under_scores. Please use camelCase format for parameter keys. There is no need for the `nvmeof` prefix either, these parameters should only be documented for the NVMe-oF controller/provisioner.

internal/nvmeof/controller/controllerserver.go
nixpanic: created https://github.com/kubernetes-csi/external-resizer/pull/545 for that, but a **draft** for now as I have not tested it yet

internal/nvmeof/controller/controllerserver.go
nixpanic: The RBD implementation does not use `ControllerModifyVolume` (yet). For `ControllerModifyVolume` the `mutable_parameters` in the StorageClass should be used (relatively new addition). Prevent using `parameters` in the StorageClass for QoS.

internal/nvmeof/controller/controllerserver.go
nixpanic: I think it is simpler with the `hasAnyQoS` bool. You need to allocate a `qos` struct so that the values can be set in it.

internal/nvmeof/controller/controllerserver.go
Madhu-1: @gadididi Lets remove it as it gets logged as part of the RPC request it self?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -293,6 +294,78 @@ func (cs *Server) ControllerUnpublishVolume(
...
+		// Both types present - invalid
+		log.ErrorLog(ctx, "Cannot mix RBD and NVMe-oF QoS parameters")
+
+		return nil, status.Error(codes.InvalidArgument, "cannot mix RBD and NVMe-oF QoS parameters")
+
+	case !hasRBDQoS && !hasNVMeoFQoS:
+		// No QoS parameters - nothing to do
+		log.DebugLog(ctx, "No QoS parameters provided for volume %s", volumeID)
+
+		return &csi.ControllerModifyVolumeResponse{}, nil
```
Madhu-1: Question:- If the values are unset or removed from the VAC, dont we need to remove these QOS from the volume as well. In other words how a user can remove QOS?

internal/nvmeof/controller/controllerserver.go
Madhu-1: yes we should move to the mutable as per the CSI spec. lets open a tracker for it for nbd QOS as well.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -351,6 +435,131 @@ func parseListeners(listenersJSON string) ([]nvmeof.ListenerDetails, error) {
...
+				return fmt.Errorf("invalid %s: %w", name, err)
+			}
+			*dest = &parsed
+			hasAnyQoS = true
+		}
+
+		return nil
+	}
+
+	if err := parseParam(nvmeof.RwIosPerSecond, nvmeof.RwIosPerSecond, &qos.RwIosPerSecond); err != nil {
```
Madhu-1, line 437: what will be the behavior if `0` is set for these keys. is it considered as infinite or disabled?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -351,6 +435,131 @@ func parseListeners(listenersJSON string) ([]nvmeof.ListenerDetails, error) {
...
+			return nil, status.Errorf(codes.Internal, "failed to get secret: %v", err)
+		}
+	}
+
+	// Step 2: Get NVMe-oF metadata
+	nvmeofData, err := cs.getNVMeoFMetadata(ctx, secrets, volumeID)
+	if err != nil {
+		log.ErrorLog(ctx, "Failed to get NVMe-oF metadata: %v", err)
+
+		return nil, status.Errorf(codes.NotFound, "NVMe-oF metadata not found: %v", err)
```
Madhu-1: this is not always NotFound error right? it could be internal server error as well

internal/nvmeof/controller/controllerserver.go
Madhu-1: lets remove it if its not required

internal/nvmeof/controller/controllerserver.go
Madhu-1: got it, i assume if we reissue the call again for the NVMe-of, the call will not fail.

internal/nvmeof/controller/controllerserver.go
Madhu-1: for Nvme lets do it right. use mutable parameters only

internal/nvmeof/qos.go
Madhu-1: yes lets use CephCSI specifc keys for user facing configuration and translate it internall to keep the keys standard

internal/nvmeof/controller/controllerserver.go
gadididi: I checked, we can remove it

internal/nvmeof/controller/controllerserver.go
gadididi: yes, correct

internal/nvmeof/controller/controllerserver.go
gadididi: Ok, so now I am going to change just `nvmeofQoS, err := parseQoSParameters(mutableParams)`
and leave this line `hasRBDQoS := rbd.HasQoSParams(params)` as is ( I add TODO, once RBD will implement the ControllerModifyVolume, we can change )

I read how it works:
**There is no section in SC that called `mutable_parameters` like `parameters`**

there will be default VolumeAttributesClass . and on the fly the user can modify with other VACs:

### example
1. StorageClass (no QoS):
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-nvmeof-sc
provisioner: nvmeof.csi.ceph.com
parameters:
  pool: mypool
  clusterID: "..."
  # No QoS here!
```
2. VolumeAttributesClass "default" (5000 IOPS):
```yaml
apiVersion: storage.k8s.io/v1
kind: VolumeAttributesClass
metadata:
  name: default
driverName: nvmeof.csi.ceph.com
parameters:
  rwIopsPerSecond: "5000"
```

**create SC and default VAC**

3. PVC (uses "default"):
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-pvc
spec:
  storageClassName: csi-nvmeof-sc
  volumeAttributesClassName: default  # ← Start with default (5000 IOPS)
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
```
**Volume is created with 5000 IOPS** 

**Later: Modify to High Performance**
5. Create "high-performance" VolumeAttributesClass (10000 IOPS):
```yaml
apiVersion: storage.k8s.io/v1
kind: VolumeAttributesClass
metadata:
  name: high-performance
driverName: nvmeof.csi.ceph.com
parameters:
  rwIopsPerSecond: "10000"
  rwMbytesPerSecond: "100"
```
6. Patch the PVC to use "high-performance":

```bash
bashkubectl patch pvc my-pvc -p '{"spec":{"volumeAttributesClassName":"high-performance"}}'
```

Once RBD controller will implement `ControllerModifyVolume()` some PVC can have VAC of default rbd QoS , and some can have VAC of default nvmeof QoS (what the user wants) , and some with nothing (unlimited) and the `req.GetMutableParameters()` will return `nil`  .

internal/nvmeof/qos.go
gadididi: Ok , I will use camelCase format.
FYI, these params will not be shown in the SC anymore, only in VAC yaml file.
I added an example in this PR here:
https://github.com/ceph/ceph-csi/pull/5714#discussion_r2517646297
 
 I think if we have 2 different types of VACs , for rbd and for nvmeof (once the rbd will implement the `modify`). 
 the VAC has field `driverName: nvmeof.csi.ceph.com`, it means only PVCs that created by nvmeof driver can be affected by this VAC. **BUT** I can create VAC for "nvmeof PVC" that has rbd qos params. 
 So I must somehow distinguish between these vars.
 I can remove the nvmeof prefix, to be e.g. `RwIosPerSecond` but it is very general, is not? 
 in this VAC I need to know what params are for nvmeof and what for rbd.
 with the nvmeof prefix it will be clearner IMO , what do you think? @nixpanic , @Madhu-1

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/qos.go
Madhu-1: @gadididi lets take a step back, do we need to support RBD QOS for the rbd images created using nvmeof csi driver? what are the advantages of it? if there is no value we should not support it, just support QOS using nvmeof only

internal/nvmeof/controller/controllerserver.go
Madhu-1: Yes thats correct expectation from me as well.

internal/nvmeof/qos.go
gadididi: @Madhu-1, I did not see specific requirement for that. but I think if there is already implementation of QoS in the RBD controller, and there is an option to set rbd QoS on nvmeof volumes (via rbd cli on the rbd image), maybe is it good to have this option here (in nvmeof controller too)

UPDATE:
I was wrong about the RBD QoS 
I asked my manager Aviv, and he said, it is not recommended to have both together (as I said before) **but also it is not recommended to allow to the user to set QoS on rbd , when the user uses nvmeof volumes**. the purpose of the nvmeof is expose higher level of setting 

So, what I am going to do:


```md
1. Parse mutable_parameters
2. Check if contains RBD QoS params?
   - YES: Return INVALID_ARGUMENT (RBD QoS not supported for NVMe-oF volumes)
   - NO: Continue
3. Check if contains NVMe-oF QoS params?
   - NO: Return SUCCESS (no-op, nothing to modify)
   - YES: Continue
4. Apply NVMe-oF QoS via gateway
5. If gateway error EXIST: means rbd QoS was set (not recommended for the user , but he did it ) return ErrorExistst
6. if gateway return another error :  Return INTERNAL
7. else Return SUCCESS
```

what do you think?

nixpanic: >  I can remove the nvmeof prefix, to be e.g. RwIosPerSecond but it is very general, is not?

Yes, the name of the parameter is general. But the StorageClass is for NVMe-oF, so it is still specific enough that way.

internal/nvmeof/nvmeof.go
```diff
@@ -167,6 +168,43 @@ func (gw *GatewayRpcClient) DeleteNamespace(ctx context.Context, subsystemNQN st
...
+) error {
+	log.DebugLog(ctx, "Setting QoS limits on namespace %d in subsystem %s", namespaceID, subsystemNQN)
+
+	req := &pb.NamespaceSetQosReq{
+		SubsystemNqn:      subsystemNQN,
+		Nsid:              namespaceID,
+		RwIosPerSecond:    qos.RwIosPerSecond,
+		RwMbytesPerSecond: qos.RwMbytesPerSecond,
+		RMbytesPerSecond:  qos.RMbytesPerSecond,
+		WMbytesPerSecond:  qos.WMbytesPerSecond,
```
nixpanic, line 186: are all these parameters required? if a parameter is not set in `qos`, there should not be a default 0 configured. Instead, the configuration should not overwrite the missing parameter, but not set that at all.

internal/nvmeof/qos.go
Madhu-1: sounds good and it also keep very minimal and exactly what we need to have

internal/nvmeof/controller/controllerserver.go
gadididi: it is considered as infinite. but "disable QoS" means remove the limitation, so I think it is same terms

internal/nvmeof/controller/controllerserver.go
gadididi: @Madhu-1 , Removing `volumeAttributesClassName` doesn't trigger `ControllerModifyVolume()`
So, how to remove QoS (means set to unlimited all params), 2 options:

1. create removal_qos VAC with empty params. once I get it (in the ControllerModifyVolume()) by patching the new removal_qos VAC , I see no params at all - means set to unlimited.

2.  create removal_qos VAC with 0 params to all 4 fields. once I get it (in the ControllerModifyVolume()) by patching the new removal_qos VAC , it calls as always to the grpc set_qos .

currently code support option 2. if I get empty VAC (no params) it finishes with no error
IMO option 2 is better. what do you think?

internal/nvmeof/nvmeof.go
gadididi: e.g. if `qos.RwIosPerSecond` was not provided in the VAC, it means the value is nil (not 0)  
When you pass nil, the gateway won't receive that field at all (every field here are optional), so it won't overwrite existing values.

internal/nvmeof/controller/controllerserver.go
gadididi: fixed ( I added some errors in `errors.go` file ) and handling it in `getNVMeoFMetadata()` function

internal/nvmeof/controller/controllerserver.go
nixpanic: This was merged. The `ControllerExpandVolume` procedure can be kept until the external-resizer has a release. (Or maybe by that time nvmeof implements resizing too.)

gadididi: @nixpanic , Aviv told me we should implement the `ControllerExpandVolume()` (resize nvmeof volume). So, in any case we need this function.

internal/nvmeof/controller/controllerserver.go
gadididi: I added

internal/nvmeof/controller/controllerserver.go
Madhu-1: >Removing volumeAttributesClassName doesn't trigger ControllerModifyVolume()

Is this something supported? we need to test with kubernetes 1.34, If yes yes option 2 make sense to me.  we need to document these behaviour to set the right expectations.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -293,6 +294,54 @@ func (cs *Server) ControllerUnpublishVolume(
...
+	// Step 2: Parse QoS parameters from mutable_parameters
+	hasRBDQoS := rbd.HasQoSParams(params)
+	if hasRBDQoS {
+		log.ErrorLog(ctx, "Cannot set RBD QoS parameters on NVMe-oF volumes")
+
+		return nil, status.Error(codes.InvalidArgument, "cannot set RBD QoS parameters on NVMe-oF volumes")
+	}
+	nvmeofQoS, err := parseQoSParameters(params)
+	if err != nil {
+		log.ErrorLog(ctx, "There are QOS params and it failed to parse QoS parameters: %v", err)
```
Madhu-1:
```suggestion
log.ErrorLog(ctx, "failed to parse NVMe-oF QoS parameters: %v", err)
```

internal/nvmeof/controller/controllerserver.go
```diff
@@ -312,6 +361,22 @@ func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
 	if err != nil {
 		return fmt.Errorf("invalid listeners parameter: %w", err)
 	}
+	// Validate QoS parameters - cannot mix RBD and NVMe-oF QoS
+	mutableParams := req.GetMutableParameters()
+	// TODO: currently the RBD QoS params are located in the StorageClass parameters (params),
+	// once RBD will implement the ControllerModifyVolume
+	// and we move the QoS params to mutable_parameters, we should change here to
+	// hasRBDQoS := rbd.HasQoSParams(mutableParams).
```
Madhu-1: Lets remove this TODO?

lets add similar check for the mutableParams as well to fail the Create it self?

Madhu-1 (commented):
small nits, LGTM

internal/nvmeof/controller/controllerserver.go
nixpanic: yes,`ControllerExpandVolume` needs to be implemented too, but in a separate PR :smile:

internal/nvmeof/controller/controllerserver.go
nixpanic: > create removal_qos VAC with 0 params to all 4 fields

Call the VAC **unlimited** or similar (in the documentation or examples), makes it easier for users to understand.

internal/nvmeof/controller/controllerserver.go
gadididi: fixed

internal/nvmeof/controller/controllerserver.go
gadididi: fixed, checks both mutableParams and Params

Madhu-1 (dismissed):

nixpanic (dismissed):

nixpanic:
@mergifyio queue

gadididi:
@mergifyio requeue

gadididi:
@mergifyio queue

gadididi:
@mergifyio requeue

Madhu-1 (approved):

nixpanic (approved):

nixpanic:
@mergifyio queue

PR #5722, opened 2025-11-04
helm: inject POD_NAMESPACE for KMIP secret lookup

### LINKED ISSUE #5721: Helm chart: missing POD_NAMESPACE env causes KMIP secret lookup to fail
### Summary
When deploying the `ceph-csi-rbd` Helm chart v3.15.0 with a KMIP-backed encrypted StorageClass, PVC
provisioning fails: the controller logs `failed to get Secret /ceph-csi-kmip-credentials`. Root cause
is that the chart never injects the `POD_NAMESPACE` env var, so the KMIP provider builds a secret lookup
with an empty namespace.

### Environment
- Chart: ceph-csi-rbd 3.15.0
- Ceph-CSI image: quay.io/cephcsi/cephcsi:v3.15.0 (default)
- Kubernetes: v1.30.2 or higher

### Steps to reproduce

1. Install the chart with a KMIP KMS config (e.g., `encryptionKMSConfig` entry for `kmip-test` and a
StorageClass using `encrypted: "true"` and `encryptionKMSID: kmip-test`).
2. Create a PVC against the encrypted StorageClass.
3. Watch the provisioner pod events/logs.

### Observed result

PVC remains `Pending`. Events repeat:

failed to provision volume with StorageClass "csi-rbd-enc-sc": rpc error: code = InvalidArgument desc
= invalid encryption kms configuration: failed to get secrets: failed to get Secret /ceph-csi-kmip-
credentials: the server could not find the requested resource

The controller request contains no namespace (`/ceph-csi-kmip-credentials`).

### Why it happens

`internal/kms/kms.go` requires the `POD_NAMESPACE` env var ([v3.15.0#L40-L41](https://github.com/ceph/ceph-csi/blob/v3.15.0/internal/kms/kms.go#L40-L41)). 

`getPodNamespace()` reads it and errors when empty ([#L106-L114](https://github.com/ceph/ceph-csi/blob/v3.15.0/internal/kms/kms.go#L106-L114)), and the namespace flows into the provider init args ([#L247-L252](https://github.com/ceph/ceph-csi/blob/v3.15.0/internal/kms/kms.go#L247-L252)). 

The KMIP provider stores that value ([`kmip.go`#L78-L90](https://github.com/ceph/ceph-csi/blob/v3.15.0/internal/kms/kmip.go#L78-L90)) and uses it to fetch the KMIP credentials secret ([#L250-L261(https://github.com/ceph/ceph-csi/blob/v3.15.0/internal/kms/kmip.go#L250-L261)).

The plain manifests provided in the repo include `POD_NAMESPACE`, but the Helm templates do not.
After manually adding the env var to `ceph-csi-rbd-provisioner` (and nodeplugin) deployments, the PVC
provisions successfully.

### Expected result
Helm-based deployments should succeed without manual patches; the chart should set `POD_NAMESPACE`
wherever the KMS code depends on it.

### Proposed fix
Add the downward API env entry to the relevant containers in `templates/provisioner-deployment.yaml`
(csi-rbdplugin, csi-rbdplugin-controller, liveness) and `templates/nodeplugin-daemonset.yaml` (csi-
rbdplugin, liveness). 

Example:

```yaml
- name: POD_NAMESPACE
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: metadata.namespace
```

imusmanmalik:
Align the Helm deployments for both RBD and CephFS with the static manifests by wiring the downward API `POD_NAMESPACE` env into every CSI container that reads the KMS config. Without it the KMIP provider falls back to an empty namespace and fails to fetch `ceph-csi-kmip-credentials`.

Fixes: #5721

- inject the downward API `POD_NAMESPACE` env into all CSI containers in the RBD and CephFS Helm charts, so the KMIP provider resolves the correct namespace
- bring the Helm charts in line with the static manifests, which already set this env var, preventing encrypted PVC provisioning failures when KMIP is enabled
- verified with `helm lint charts/ceph-csi-rbd` and `helm lint charts/ceph-csi-cephfs`

No backward-compatibility concerns: the env var comes from the pod namespace and is already expected by the Ceph-CSI binaries. Helm users without KMIP see no behavioural change.

## Related issues ##

Fixes: #5721

## Future concerns ##

None.

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
+++ b/charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
@@ -90,6 +90,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
             - name: DRIVER_NAME
               value: {{ .Values.driverName }}
             - name: NODE_ID
--- a/charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
+++ b/charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
@@ -115,6 +115,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
             - name: DRIVER_NAME
               value: {{ .Values.driverName }}
             - name: NODE_ID
--- a/charts/ceph-csi-cephfs/templates/provisioner-deployment.yaml
+++ b/charts/ceph-csi-cephfs/templates/provisioner-deployment.yaml
@@ -110,6 +110,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
             - name: DRIVER_NAME
               value: {{ .Values.driverName }}
             - name: NODE_ID
--- a/charts/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
+++ b/charts/ceph-csi-rbd/templates/nodeplugin-daemonset.yaml
@@ -88,6 +88,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
             - name: DRIVER_NAME
               value: {{ .Values.driverName }}
             - name: NODE_ID


charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
```diff
@@ -178,6 +182,10 @@ spec:
           env:
             - name: CSI_ENDPOINT
               value: "unix:///csi/{{ .Values.pluginSocketFile }}"
+            - name: POD_NAMESPACE
```
iPraveenParihar: why do we need it for `liveness-prometheus` container ?

charts/ceph-csi-rbd/templates/provisioner-deployment.yaml
```diff
@@ -310,6 +314,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: POD_NAMESPACE
```
iPraveenParihar: I don't think we need here as well.

iPraveenParihar (changes requested):
small nit.
Thanks!

charts/ceph-csi-cephfs/templates/nodeplugin-daemonset.yaml
imusmanmalik: the liveness-prometheus sidecar is just the cephcsi binary in liveness mode, so it never talks to the KMS. I’ve dropped POD_NAMESPACE from both liveness containers (and from the RBD controller helper) so the charts now mirror what’s shipped under deploy/rbd/kubernetes/csi-rbdplugin-provisioner.yaml, where only the main plugin has that env var. Thanks for calling it out!

Madhu-1 (changes requested):
Please squash the changes to a single commit and also ensure that same changes are present in non helm based yamls as well

imusmanmalik:
> Please squash the changes to a single commit and also ensure that same changes are present in non helm based yamls as well

@Madhu-1 Changes are already present in manifests under `deploy` so i believe we don't need to change anything there. It was other way around changes were not present in helm templates.

Madhu-1 (approved):

iPraveenParihar (approved):

iPraveenParihar:
@Mergifyio rebase

iPraveenParihar:
@Mergifyio queue

PR #5725, opened 2025-11-05
rbd: remove forcePromote function

### LINKED ISSUE #5349: remove timeout for the force promote command in rbd
# Describe the bug #

When the timeout was added, rbd mirroring was still in it's early stage and we needed a. workaround for the force promote to not get stuck and get it called again if it is not executed within a stipulated timeframe.
But, with the current enhancements around rbd and osds the timeout option can be remove for the force promote command and we should wait for the command to complete and return the result. If it's taking more time, that means the image is very large and it actually requires time to rollback the image from scratch.

Nikhil-Ladha:

Remove `ForcePromote` function as we don't need a timeout for force promote to work anymore. The issue has been fixed in rbd and force promote is supposed to work without the timeout.

Fixes: #5349

### DIFF
--- a/internal/rbd/mirror.go
+++ b/internal/rbd/mirror.go
@@ -27,7 +27,6 @@ import (
 
 	rbderrors "github.com/ceph/ceph-csi/internal/rbd/errors"
 	"github.com/ceph/ceph-csi/internal/rbd/types"
-	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/log"
 )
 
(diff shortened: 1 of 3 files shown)


internal/rbd/mirror.go
```diff
@@ -187,40 +186,12 @@ func (rm *rbdMirror) Promote(ctx context.Context, force bool) error {
...
-		"rbd",
-		promoteArgs...,
-	)
-	if err != nil {
-		return fmt.Errorf("failed to promote image %q with error: %w", rm, err)
-	}
-
-	if stderr != "" {
-		return fmt.Errorf("failed to promote image %q with stderror: %s", rm, stderr)
+	if force {
```
Madhu-1: instead of if/else logging, log the force flag as well

Nikhil-Ladha: Done.

Madhu-1 (approved):

iPraveenParihar (approved):

iPraveenParihar:
Since the changes are in RBD mirroring code and we don't yet have mirroring related e2e. We can skip running e2e.

iPraveenParihar:
@Mergifyio rebase

Issue #5757, opened 2025-11-16
Add support for CSI volume expansion in the NVMe-oF CSI driver


gadididi:
# Describe the feature you'd like to have

Add support for CSI volume expansion (ControllerExpandVolume and NodeExpandVolume) in the NVMe-oF CSI driver to allow dynamic resizing of persistent volumes.

# What is the value to the end user? (why is it a priority?)

Users can expand their persistent volumes without downtime or manual intervention. When applications need more storage capacity, administrators can simply update the PVC size in Kubernetes, and the CSI driver will handle the resize operation automatically through the NVMe-oF gateway.

# How will we know we have a good solution? (acceptance criteria)

-  CSI driver implements `ControllerExpandVolume` RPC call
-  CSI driver implements `NodeExpandVolume` RPC call for filesystem expansion
-  Driver advertises `EXPAND_VOLUME` controller capability and `EXPAND_VOLUME` node capability
-  Driver advertises `VolumeExpansion.ONLINE` capability for online expansion
-  Volume expansion works for both block and filesystem modes
-  PVC expansion from size of X to size of Y completes successfully without pod restart
-  Filesystem is automatically expanded after namespace resize
-  Idempotent: expanding to same or smaller size returns success without error

# Additional context

The NVMe-oF gateway already supports namespace resizing via the `namespace_resize` gRPC call, which internally calls SPDK's `bdev_rbd_resize` to resize both the RBD image and the NVMe namespace. This feature request is to expose this functionality through the CSI interface.

gadididi:
this feature was merged:
https://github.com/ceph/ceph-csi/pull/5792

PR #5792, opened 2025-11-24
nvmeof: expansion volume feature

gadididi:
# Describe what this PR does

Implements CSI `ControllerExpandVolume` and `NodeExpandVolume` for NVMe-oF volumes to support online volume expansion.

## Is there anything that requires special attention

**ControllerExpandVolume:**
- Calls directly to rbd `ControllerExpandVolume()` . 

**NodeExpandVolume:**
- Search for the by the current staging path the device path the pod is mounted to . 
- Resizes filesystem using `mount.NewResizeFs()` (supports ext4, xfs, btrfs)
- Skips filesystem resize for block mode volumes

FYI:
1. The NVMe-oF gateway supports namespace resizing via the ResizeNamespace GRPC call, which internally calls SPDK's `bdev_rbd_resize` to resize both the RBD image and the NVMeoF namespace. - BUT best practice is to resize directly the rbd itself. 

TODO:
 - add some unit tests 

## Related issues ##

https://github.com/ceph/ceph-csi/issues/5757

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/nvmeof/controller/controllerserver.go
+++ b/internal/nvmeof/controller/controllerserver.go
@@ -76,6 +76,12 @@ func (cs *Server) ControllerGetCapabilities(
 	return cs.backendServer.ControllerGetCapabilities(ctx, req)
 }
 
+// ValidateControllerServiceRequest uses the RBD backendServer driver field.
+// It checks whether the controller service request is valid.
+func (cs *Server) ValidateControllerServiceRequest(capability csi.ControllerServiceCapability_RPC_Type) error {
+	return cs.backendServer.Driver.ValidateControllerServiceRequest(capability)
+}
+
 // ValidateVolumeCapabilities checks whether the volume capabilities requested
 // are supported.
 func (cs *Server) ValidateVolumeCapabilities(
@@ -330,16 +336,18 @@ func (cs *Server) ControllerModifyVolume(
 	return &csi.ControllerModifyVolumeResponse{}, nil
 }
 
-// ControllerExpandVolume handles volume expansion requests.
-// For now it only updates the capacity in the response as NVMe-oF
-// this must be added because ControllerModifyVolume requires the sidecar csi-resizer. and
-// csi-resizer searches for the capacity ControllerServiceCapability_RPC_EXPAND_VOLUME.
-// In the future, if NVMe-oF gateway supports volume expansion, the logic must be added here.
 func (cs *Server) ControllerExpandVolume(
 	ctx context.Context,
 	req *csi.ControllerExpandVolumeRequest,
 ) (*csi.ControllerExpandVolumeResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "ControllerExpandVolume is not implemented for NVMe-oF volumes")
+	err := cs.ValidateControllerServiceRequest(csi.ControllerServiceCapability_RPC_EXPAND_VOLUME)
+	if err != nil {
+		log.ErrorLog(ctx, "invalid expand volume req: %v", err)
+
+		return nil, err
+	}
+	// expand volume is handled by rbd backend server
+	return cs.backendServer.ControllerExpandVolume(ctx, req)
 }
 
 // validateCreateVolumeRequest validates the incoming request for nvmeof.
--- a/internal/nvmeof/nodeserver/nodeserver.go
+++ b/internal/nvmeof/nodeserver/nodeserver.go
@@ -121,6 +121,13 @@ func (ns *NodeServer) NodeGetCapabilities(
 					},
 				},
 			},
+			{
+				Type: &csi.NodeServiceCapability_Rpc{
+					Rpc: &csi.NodeServiceCapability_RPC{
+						Type: csi.NodeServiceCapability_RPC_EXPAND_VOLUME,
+					},
+				},
+			},
 		},
 	}, nil
 }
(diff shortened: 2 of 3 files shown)


nixpanic:
> * The NVMe-oF gateway supports namespace resizing via the ResizeNamespace GRPC call, which internally calls SPDK's `bdev_rbd_resize` to resize both the RBD image and the NVMeoF namespace.
> 
> * If someone change externally the rbd image. automatically the namespace will be changed according to the rbd image. So maybe we should reconsider this approach: The NVMeoF's `ControllerExpandVolume()` will call to rbd's  `ControllerExpandVolume()`. that's it. for NodeExpandVolume, there is need to implement it, because the device is exposed as "nvme device" (and not "rbd device")

I do not have a preference for one way or the other. Maybe the gateway approach is more future proof, in case the gateway or spdk uses the size of the image at a later point in time?

internal/nvmeof/controller/controllerserver.go
```diff
@@ -330,16 +337,134 @@ func (cs *Server) ControllerModifyVolume(
...
+	if secrets == nil {
+		/*	Reuse the existing function - it's just getting Ceph credentials from k8s secret.
+			GetControllerPublishSecretRef name is misleading here.
+			another option is to create a new function like this one:
+
+				func GetControllerSecretRef(volumeId, driverType string) (string, string, error) {
+					return GetControllerPublishSecretRef(volumeId, driverType)
+				}
+		*/
+		secretName, secretNamespace, err := util.GetControllerPublishSecretRef(req.GetVolumeId(), util.RBDType)
```
nixpanic: The `ControllerExpandVolumeRequest` should contain secrets already, use those.

https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerexpandvolume

internal/nvmeof/controller/controllerserver.go
```diff
@@ -330,16 +337,134 @@ func (cs *Server) ControllerModifyVolume(
...
+			return nil, status.Errorf(codes.Internal, "failed to resize NVMe-oF namespace: %v", err)
+		}
+
+		log.DebugLog(ctx, "Successfully resized NVMe-oF namespace for volume %s", volumeID)
+	} else {
+		log.DebugLog(ctx, "No resize required for NVMe-oF namespace for volume %s", volumeID)
+	}
+
+	return &csi.ControllerExpandVolumeResponse{
+		CapacityBytes:         int64(namespaceSizeBytes),
```
nixpanic: does the gateway not round-up the size? The RBD-image has a certain block-size, so possibly the image a little larger than requested?

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -356,6 +357,78 @@ func (ns *NodeServer) NodeUnstageVolume(
...
+	}
+
+	if acquired := ns.volumeLocks.TryAcquire(volumeID); !acquired {
+		log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volumeID)
+
+		return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer ns.volumeLocks.Release(volumeID)
+
+	namespaceUUID, err := lookupNvmeVolumeMetadataStash(volumePath)
```
nixpanic: Is this the only thing where you need the metadata stash? Is there no reasonable syscall that we can use to identify the device by the `req.GetVolumePath()`?

If there is no need for the metadata stash, that would be quite nice.

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -732,3 +805,40 @@ func getStagingTargetPath(req interface{}) string {
 
 	return ""
 }
+
+// getStagingPath returns the base staging path for a volume from its publish path.
+// For K8s < v1.19, NodeExpandVolume may only receive the publish path (where the pod sees the volume).
```
nixpanic: I assume this is mostly copied from the rbd nodeplugin. There is no support for Kuberneres 1.19 and older, there is no need to take that version into account anymore.

internal/nvmeof/nodeserver/nodeserver.go
gadididi: I think I can retrieve the device path from `mount` e.g. `/dev/disk/by-id/nvme-uuid.12345678-1234-1234-1234-123456789abc`, 
then I can parse and get the UUID. 

but I was wondering, if the device is not mounted, it will fail. and metadata always succeed, what do you think?

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -332,11 +346,89 @@ func (ns *NodeServer) NodeUnstageVolume(
...
+		// See: https://github.com/kubernetes/kubernetes/issues/115343
+		var err error
+		volumePath, err = ns.getStagingPath(req.GetVolumePath())
+		if err != nil {
+			// If stagingTargetPath is not found in volumePath then use
+			// volumePath
+			volumePath = req.GetVolumePath()
+			log.UsefulLog(ctx, "failed to find stagingTargetPath from volumePath %v: %v", volumePath, err)
+		}
+	}
```
Madhu-1: Lets stick to the newer kubernetes supported version like > 1.30

internal/nvmeof/nodeserver/nodeserver.go
Madhu-1: If the device is not mounted means that the NodeStage is not done, if its not done we need need to resize, we can return error until that time.

internal/nvmeof/nodeserver/nodeserver.go
Madhu-1: Lets remove this one as well

Madhu-1 (commented):
small nits LGTM

internal/nvmeof/nodeserver/nodeserver.go
gadididi: So, I am going to remove the stashing part, and just make small search:
something like this:

```Go
func getDeviceFromMount(mounter mount.Interface, mountPath string) (string, error) {
	// Get mount info
	mountInfos, err := mounter.List()
	if err != nil {
		return "", err
	}

	for _, mi := range mountInfos {
		if mi.Device == mountPath {
			// mi.Device is the device path
			// e.g., /dev/disk/by-id/nvme-uuid.12345678-1234-1234-1234-123456789abc
			return mi.Device, nil
		}
	}

	return "", fmt.Errorf("mount not found for path %s", mountPath)
}
```
when `mountPath` param is the `req.GetStagingTargetPath()`

Madhu-1: Will search be a costly operations when we have too many mounts in the system? if we will get into performance penalty lets stick to this one only. @nixpanic thoughts?

nixpanic: IMHO stashing details is really ugly. There can be thousands of mounts on a container host, so the loop might take long. However, compared to the I/O done for resizing (metadata updates on the filesystem), this loop won't be significant. It is not like resizing is an operation that is done constantly.

My preference is to detect the device according to the path.

internal/nvmeof/controller/controllerserver.go
nixpanic: you may remove the commented out code :smile:

gadididi: @nixpanic , did you mean remove this entire section:
```go
	if secrets == nil {
		/*	Reuse the existing function - it's just getting Ceph credentials from k8s secret.
			GetControllerPublishSecretRef name is misleading here.
			another option is to create a new function like this one:

				func GetControllerSecretRef(volumeId, driverType string) (string, string, error) {
					return GetControllerPublishSecretRef(volumeId, driverType)
				}
		*/
		secretName, secretNamespace, err := util.GetControllerPublishSecretRef(req.GetVolumeId(), util.RBDType)
		if err != nil {
			log.ErrorLog(ctx, "controller expand secret not found: %v", err)

			return nil, status.Errorf(codes.Internal, "failed to get secrets: %v", err)
		}

		secrets, err = k8s.GetSecret(secretName, secretNamespace)
		if err != nil {
			return nil, status.Errorf(codes.Internal, "failed to get secrets: %v", err)
		}
	}

```
right?

nixpanic: yes, correct. `secrets` should not be `nil` (in the near future).

internal/nvmeof/controller/controllerserver.go
gadididi: good catch I will check it. After size changes, I must return the current new size and not the requested size

gadididi: update: nvmeof  `ControllerExpandVolume()` will call to rbd `ControllerExpandVolume()`

internal/nvmeof/nodeserver/nodeserver.go
gadididi: fixed

internal/nvmeof/nvmeof.go
```diff
@@ -205,6 +205,35 @@ func (gw *GatewayRpcClient) SetQoSLimitsForNamespace(
 		status.GetStatus(), status.GetErrorMessage())
 }
 
+// ResizeNamespace resizes a NVMeoF namespace to a new size.
+func (gw *GatewayRpcClient) ResizeNamespace(
```
nixpanic: I think you can drop this commit now too.

nixpanic (commented):
looks good to me, just need to drop the 1st commit as it is unused?

internal/nvmeof/nvmeof.go
gadididi: removed

internal/nvmeof/nvmeof.go
```diff
@@ -238,6 +238,38 @@ func (gw *GatewayRpcClient) GetUUIDBySubsystemAndNameSpaceID(
 	return "", fmt.Errorf("namespace with nsid %d not found", namespaceID)
 }
 
+// GetNamespaceSize gets the size of a namespace in bytes.
+func (gw *GatewayRpcClient) GetNamespaceSize(
```
Madhu-1: Lets remove it as its not used anywhere?

gadididi: removed

Madhu-1 (approved):

nixpanic (approved):

PR #5799, opened 2025-11-25
rbd: add block volume stats support

### LINKED ISSUE #5471: Returned current usage for block volume metrics
Today due to missing secrets we are not able to query ceph/rbd to provide the current usage metrics of the block volume. As we recently introduced a mechanism to provide the secrets per cluster ID to connect to ceph. We can use that mechanism to get the secret and connect to ceph and get the actual usage metrics.

This is a workaround for the solution as there is no progress on adding the secrets to get NodeGetVolumeStats RPC call.

Rakshith-R:
This commit adds support for providing rbd
block mode volume stats for NodeGetVolumeStats
call using DiffIterate API.
To compensate for lack of secrets in NodeGetVolumeStats call. The newly introduced ControllerPublishSecretRef is used to get credentials to connect to ceph
cluster.

Do you have any questions?

For example:

## Related issues ##

Fixes: #issue_number

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -7,5 +7,6 @@
 - set nodeId:userId mapping in metadata [PR](https://github.com/ceph/ceph-csi/pull/5445)
    - refer design doc for more details - [here](https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/userID-mapping.md)
 - cephfs-csi: fix mounting alternate filesystem when mounting by monitor lists [PR](https://github.com/ceph/ceph-csi/pull/5643)
+- rbd: add block volume stats support [PR](https://github.com/ceph/ceph-csi/pull/5799)
 
 ## NOTE
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -38,6 +38,7 @@ import (
 	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/file"
 	"github.com/ceph/ceph-csi/internal/util/fscrypt"
+	"github.com/ceph/ceph-csi/internal/util/k8s"
 	"github.com/ceph/ceph-csi/internal/util/log"
 	"github.com/ceph/ceph-csi/pkg/util/kernel"
 )
@@ -1486,6 +1487,13 @@ func (ns *NodeServer) NodeGetVolumeStats(
 	req *csi.NodeGetVolumeStatsRequest,
 ) (*csi.NodeGetVolumeStatsResponse, error) {
 	var err error
+	volumeId := req.GetVolumeId()
+	if volumeId == "" {
+		err = fmt.Errorf("volumeID %v is empty", volumeId)
+
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
 	targetPath := req.GetVolumePath()
 	if targetPath == "" {
 		err = fmt.Errorf("targetpath %v is empty", targetPath)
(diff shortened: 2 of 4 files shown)


internal/rbd/nodeserver.go
```diff
@@ -1512,32 +1522,99 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+		// treat it as a static volume if decomposition fails.
+		// Static volumes do not support diff iterate to measure usage.
+		isDiffPossible = false
+	}
 
-		return nil, status.Error(codes.Internal, err.Error())
+	secretName, secretNamespace, err := util.GetControllerPublishSecretRef(volumeId, util.RBDType)
+	if err != nil {
+		// diff iterate to measure usage is not possible without secrets.
+		isDiffPossible = false
```
black-dragon74, line 1560: Let's also log it? Would help in debugging.

internal/rbd/nodeserver.go
```diff
@@ -1512,32 +1522,99 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+			},
+		}, nil
+	}
+
+	secrets, err := k8s.GetSecret(secretName, secretNamespace)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get node volume stats secret from k8s: %w", err)
+	}
+
+	credentials, err := util.NewAdminCredentials(secrets)
```
black-dragon74, line 1575: The RPC `NodeGetVolumeStats` would be called frequently and this will create a temp file leading to IO latency/wear on each call. Can we optimize this somehow?

internal/rbd/nodeserver.go
Rakshith-R: > Let's also log it? Would help in debugging.

NodeGetVolumeStats is a very frequest call like every <~2 mins,
Logging this would easily fill up the entire log.

At this point, I don't think lot of users have added publishsecret to the configmap too.

internal/rbd/nodeserver.go
Rakshith-R: > The RPC `NodeGetVolumeStats` would be called frequently and this will create a temp file leading to IO latency/wear on each call. Can we optimize this somehow?

This feature is disabled by default.
When enabled, we will reach here only for RBD block mode volumes.

Let's get this in and after extensive testing make such perf optimizations :)

internal/rbd/nodeserver.go
black-dragon74: While I agree with that, we would never know why `isDiffPossible` is not possible (no pun intended) xD

Rakshith-R: > While I agree with that, we would never know why `isDiffPossible` is not possible (no pun intended) xD

agreed :laughing:  , 
VolumeID could easily be made from logs to indicate static volume or not, 
+ added the following

`GetControllerPublishSecretRef` now returns a known error `ErrConfigNotFound` in case secret is not present so we can distinguish between absence of secret and an error while parsing and finding it.
Hope its better now

Madhu-1:
@Rakshith-R why do we need to have this optional not has a default feature?

Rakshith-R:
> @Rakshith-R why do we need to have this optional not has a default feature?

Performance reasons, 
NodeGetVolumeStats will get called every 2 minutes on each mounted volume, diff iterate command linearly grows with image size, and we do not exactly know the impact until we do extensive testing.

That's why I would like to have a flag and keep it disabled by default for now..

Madhu-1:
> > @Rakshith-R why do we need to have this optional not has a default feature?
> 
> Performance reasons, NodeGetVolumeStats will get called every 2 minutes on each mounted volume, diff iterate command linearly grows with image size, and we do not exactly know the impact until we do extensive testing.
> 
> That's why I would like to have a flag and keep it disabled by default for now..

I would suggest the opposite way, Dont add the flag as part of this PR, Lets enable it by default and get it tested, based on the results we can have additional PR to configure it, IMO we should move away from introducing new flag for every feature and have something like (--features-set=key:value,key:value) just like the kubernetes sidecar, it helps us to easily deprecate or support the features instead of adding/removing new commandline agruments always) which might break the releases/upgrades

Rakshith-R:
> > > @Rakshith-R why do we need to have this optional not has a default feature?
> > 
> > 
> > Performance reasons, NodeGetVolumeStats will get called every 2 minutes on each mounted volume, diff iterate command linearly grows with image size, and we do not exactly know the impact until we do extensive testing.
> > That's why I would like to have a flag and keep it disabled by default for now..
> 
> I would suggest the opposite way, Dont add the flag as part of this PR, Lets enable it by default and get it tested, based on the results we can have additional PR to configure it, IMO we should move away from introducing new flag for every feature and have something like (--features-set=key:value,key:value) just like the kubernetes sidecar, it helps us to easily deprecate or support the features instead of adding/removing new commandline agruments always) which might break the releases/upgrades

Agreed, removed the single cmdline flag for now.
We should introduced a feature set flag instead to avoid the mess.

Please take a look.

internal/rbd/nodeserver.go
```diff
@@ -1512,32 +1520,102 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+	secretName, secretNamespace, err := util.GetControllerPublishSecretRef(volumeId, util.RBDType)
+	if errors.Is(err, util.ErrConfigNotFound) {
+		// diff iterate to measure usage is not possible without secrets.
+		isDiffPossible = false
+	} else if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	// If diff iterate is not possible, return only the size of the block device.
+	if !isDiffPossible {
```
nixpanic, line 1566: This could be its own function and can be called/returned on the line where `isDiffPossible` is set to `false` above.

Splitting it makes maintaining this a little easier.

internal/rbd/nodeserver.go
nixpanic: I think creating a temporary file won't be a noticeable delay compared to the diff-iterating of the image.

nixpanic:
> NodeGetVolumeStats will get called every 2 minutes on each mounted volume, diff iterate command linearly grows with image size, and we do not exactly know the impact until we do extensive testing.

Do you have performance numbers of diff-iterate on images with different sizes and snapshot-chains?

PendingReleaseNotes.md
```diff
@@ -7,5 +7,6 @@
 - set nodeId:userId mapping in metadata [PR](https://github.com/ceph/ceph-csi/pull/5445)
    - refer design doc for more details - [here](https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/userID-mapping.md)
 - cephfs-csi: fix mounting alternate filesystem when mounting by monitor lists [PR](https://github.com/ceph/ceph-csi/pull/5643)
+- rbd: rbd: add block volume stats support [PR](https://github.com/ceph/ceph-csi/pull/5799)
```
Madhu-1:
```suggestion
- rbd: add block volume stats support [PR](https://github.com/ceph/ceph-csi/pull/5799)
```

internal/rbd/rbd_util.go
```diff
@@ -2338,3 +2338,39 @@ func (rv *rbdVolume) PrepareVolumeForSnapshot(ctx context.Context, cr *util.Cred
 
 	return nil
 }
+
+// usedBytes returns the logical used bytes of the rbd image by iterating over
```
Madhu-1: getUsedBytes returns

internal/rbd/rbd_util.go
```diff
@@ -2338,3 +2338,39 @@ func (rv *rbdVolume) PrepareVolumeForSnapshot(ctx context.Context, cr *util.Cred
...
+
+				return 0
+			},
+		})
+	if err != nil {
+		return 0, fmt.Errorf("failed to iterate over rbd image %s: %w", rv, err)
+	}
+
+	end := time.Now()
+	log.DebugLog(ctx, "DiffIterate on image of size %d bytes took %v seconds",
```
Madhu-1: log image details as well which helps for  debugging

Madhu-1 (commented):
Can we please add test case to ensure we are getting expected results?

Rakshith-R:
> Can we please add test case to ensure we are getting expected results?

Existing NodeGetVolumeStats implementation do not have test cases too.
I think we can have a test case that verifies the results directly from kubelet metrics generated from this call.
This will be added in a follow-up pr..
Opened an issue to track it https://github.com/ceph/ceph-csi/issues/5804

internal/rbd/nodeserver.go
Rakshith-R: > This could be its own function and can be called/returned on the line where `isDiffPossible` is set to `false` above.
> 
> Splitting it makes maintaining this a little easier.

Created helped to getblockmetrics.
IsDiffPossible is set to false at two places, Therefore, I think its cleaner to check and call in a different if statement.

Rakshith-R:
> > NodeGetVolumeStats will get called every 2 minutes on each mounted volume, diff iterate command linearly grows with image size, and we do not exactly know the impact until we do extensive testing.
> 
> Do you have performance numbers of diff-iterate on images with different sizes and snapshot-chains?

Tests were run once on these volume usage :
```
1G - 35ms
25G - 900ms
50G - 1.8s
100G - 3.5s
150G - 4.2s
```
Tests were also conducted on snapshot-chains but it worked as expected

The duration may vary and we would need very extensive testing to arrive at an approximate.

Madhu-1 (approved):

nixpanic (approved):

nixpanic:
@mergifyio queue

PR #5807, opened 2025-11-26
util: fix cache miss when secret watcher is not running

black-dragon74:

This patch addresses the following:

- Retry starting the secret watcher if it fails.
- Do not read from cache if watcher is not running.
- Reduce the work under locks.

### DIFF
--- a/internal/util/k8s/secrets.go
+++ b/internal/util/k8s/secrets.go
@@ -124,16 +124,7 @@ func (sc *secretCache) deleteFromCache(secret *corev1.Secret) {
 func (sc *secretCache) updateCache(secret *corev1.Secret, force bool) map[string]string {
 	cKey := sc.cacheKey(secret.Namespace, secret.Name)
 
-	// We should cache this secret only if it was requested in the first place i.e force == true
-	// For all subsequent updates where this function is called, it is set to false
-	sc.RLock()
-	if _, requested := sc.cache[cKey]; !requested && !force {
-		sc.RUnlock()
-
-		return nil
-	}
-	sc.RUnlock()
-
+	// populate data outside the lock
 	secretData := make(map[string]string, len(secret.Data))
 	for key, value := range secret.Data {
 		secretData[key] = string(value)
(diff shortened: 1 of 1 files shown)


internal/util/k8s/secrets.go
```diff
@@ -124,16 +124,7 @@ func (sc *secretCache) deleteFromCache(secret *corev1.Secret) {
...
-	// For all subsequent updates where this function is called, it is set to false
-	sc.RLock()
-	if _, requested := sc.cache[cKey]; !requested && !force {
-		sc.RUnlock()
-
-		return nil
-	}
-	sc.RUnlock()
-
+	// populate date outside the lock
```
nixpanic: _date_ probably should be _data_?

nixpanic (dismissed):

internal/util/k8s/secrets.go
black-dragon74: Fixed xD

internal/util/k8s/secrets.go
```diff
@@ -143,9 +134,15 @@ func (sc *secretCache) updateCache(secret *corev1.Secret, force bool) map[string
...
 
+	// We should cache this secret only if it was requested in the first place i.e force == true
+	// For all subsequent updates where this function is called, it is set to false
 	sc.Lock()
 	defer sc.Unlock()
 
+	if _, requested := sc.cache[cKey]; !requested && !force {
+		return nil
+	}
+
```
Rakshith-R:
```suggestion
	// We should cache this secret only if it was already present in the first place i.e force == true
	// For all subsequent updates where this function is called, it is set to false
	sc.Lock()
	defer sc.Unlock()

	if _, present := sc.cache[cKey]; !found && !force {
		return nil
	}

```
I was confused by the use of `requested`, can we instead use `present`  ?

black-dragon74: If it helps, it will only be present if it was requested by us in the first place. And in that step we are just checking that. 

We ignore events for all other secrets.

Rakshith-R (approved):

nixpanic (approved):

nixpanic:
@mergifyio queue

PR #5871, opened 2025-12-18
rbd: allow configuring deleted volume retention time (#5830)

### LINKED ISSUE #5830: allow configuring deleted volume retention time
ceph supports moving volumes into the trash to be deleted at a configurable time later.
This is very useful to recover from accidentally deleted images

It appears ceph-csi does not allow to configure retention time for deleted volumes, it is hardcoded to 0 seconds
https://github.com/ceph/ceph-csi/blob/devel/internal/rbd/rbd_util.go#L710

tested with v3.14.1
```
rbd ls | grep csi-vol-f3c2954b-1ede-40b4-aada-215719a2fafe
csi-vol-f3c2954b-1ede-40b4-aada-215719a2fafe
kubectl delete pvc name
rbd trash ls
# empty
```

It would be nice if volumes created by ceph-csi could always have a retention time on deleted volumes. Having a single global value for all managed volumes would be sufficient.

# What is the value to the end user? #

It is easy to accidentally delete a persistentvolumeclaim with a volume with reclaimpolicy: delete, directly or indirectly via e.g. namespace deletion.

It would be valuable that this could be recovered by restoring the volume on the ceph store from the trash and restoring the persistentvolume object that that has been backed up via other means.

reclaimpolicy: retain provides this option but in multitenant clusters retaining persistentvolumes introduces additional complexity in handling permissions for reclaiming and cleanup of old volumes.
It also does not protect from accidentally deleting the persistentvolume itself (though that is less likely to happen on accident).

Ady0333:
# Description#

- This PR introduces a configurable option to delay the permanent deletion of RBD volumes by moving them to the RBD trash with a specified deferment time.
- 
- Previously, the CSI driver invoked `rbdImage.Trash(0)` with a hard-coded deferment value of `0`, meaning deleted volumes were immediately eligible for purge or removal depending on the backend configuration. There was no way to specify a delay from the CSI side to allow recovery of accidentally deleted volumes.
- 
- This change allows administrators to configure a safety window (for example, one hour) during which deleted volumes remain in the RBD trash and can be recovered before being permanently removed.

- No special attention is required for normal usage.

Do you have any questions?
- No open questions at this time.

Yes, the change is backward compatible.

No. The default value of the new configuration is `0`, which preserves the existing behavior.

- This change aligns with Ceph RBD’s native support for trash deferment.
- It allows Ceph-CSI users to better leverage Ceph’s safety mechanisms when managing persistent volumes in Kubernetes environments.
- No CSI spec changes are required for this enhancement.

## Related issues ##

Fixes: #5830 

## Tests Performed

- Attempted local execution of RBD unit tests:
  - go test ./internal/rbd/...

  These fail locally due to missing Ceph native libraries
  (librbd, librados), required by go-ceph CGO bindings.

- Attempted CGO-disabled execution:
  - CGO_ENABLED=0 go test ./internal/rbd/features

  This fails as expected because feature detection relies on
  CGO-enabled Ceph symbols.

- Provided a detailed validation guide describing how to
  verify the behavior in a Ceph-backed Kubernetes environment.

Full verification is expected to be covered by CI.

**Checklist:**

* [✔️] **Commit Message Formatting**: Commit titles and messages follow
  guidelines in the [developer
* [ ✔️] Reviewed the developer guide on [Submitting a Pull
* [ ] [Pending release

### DIFF
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -151,6 +151,12 @@ func init() {
 		"Minimum number of snapshots required on rbd image to start flattening")
 	flag.BoolVar(&conf.SkipForceFlatten, "skipforceflatten", false,
 		"skip image flattening if kernel support mapping of rbd images which has the deep-flatten feature")
+	flag.UintVar(
+		&conf.RbdTrashMaxDelay,
+		"rbdtrashmaxdelay",
+		0,
+		"Storage class parameter 'trashMaxDelay' value used for a delay (in seconds)"+
+			" before processing the deletion of the rbd image.")
 
 	flag.BoolVar(&conf.Version, "version", false, "Print cephcsi version information")
 	flag.BoolVar(&conf.EnableProfiling, "enableprofiling", false, "enable go profiling")
--- a/cmd/cephcsi.go
+++ b/cmd/cephcsi.go
@@ -151,12 +151,13 @@ func init() {
 		"Minimum number of snapshots required on rbd image to start flattening")
 	flag.BoolVar(&conf.SkipForceFlatten, "skipforceflatten", false,
 		"skip image flattening if kernel support mapping of rbd images which has the deep-flatten feature")
-	flag.UintVar(
+	flag.DurationVar(
 		&conf.RbdTrashMaxDelay,
 		"rbdtrashmaxdelay",
 		0,
-		"Storage class parameter 'trashMaxDelay' value used for a delay (in seconds)"+
-			" before processing the deletion of the rbd image.")
+		"Delay before permanently deleting RBD images from trash. When set to a positive duration, "+
+			"deleted volumes are moved to Ceph trash and retained for this period. Users are responsible "+
+			"for purging trash after the retention period.")
 
 	flag.BoolVar(&conf.Version, "version", false, "Print cephcsi version information")
 	flag.BoolVar(&conf.EnableProfiling, "enableprofiling", false, "enable go profiling")
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"time"
 
 	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -208,6 +209,14 @@ func (cs *ControllerServer) parseVolCreateRequest(
 	err = rbdVol.initKMS(ctx, req.GetParameters(), req.GetSecrets())
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	if delay, ok := req.GetParameters()["trashMaxDelay"]; ok {
+		d, err := time.ParseDuration(delay)
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "invalid trashMaxDelay: %v", err)
+		}
+		rbdVol.TrashMaxDelay = d
 	}
 
 	rbdVol.RequestName = req.GetName()
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -445,6 +445,9 @@ func (cs *ControllerServer) CreateVolume(
 
 	// Set Metadata on PV Create
 	metadata := k8s.GetVolumeMetadata(req.GetParameters())
+	if delay, ok := req.GetParameters()["trashMaxDelay"]; ok {
+		metadata[trashMaxDelayKey] = delay
+	}
 	err = rbdVol.setAllMetadata(metadata)
 	if err != nil {
 		if deleteErr := rbdVol.Delete(ctx); deleteErr != nil {
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
-	"time"
 
 	librbd "github.com/ceph/go-ceph/rbd"
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -209,14 +208,6 @@ func (cs *ControllerServer) parseVolCreateRequest(
 	err = rbdVol.initKMS(ctx, req.GetParameters(), req.GetSecrets())
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
-	}
-
-	if delay, ok := req.GetParameters()["trashMaxDelay"]; ok {
-		d, err := time.ParseDuration(delay)
-		if err != nil {
-			return nil, status.Errorf(codes.InvalidArgument, "invalid trashMaxDelay: %v", err)
-		}
-		rbdVol.TrashMaxDelay = d
 	}
 
 	rbdVol.RequestName = req.GetName()
@@ -454,9 +445,6 @@ func (cs *ControllerServer) CreateVolume(
 
 	// Set Metadata on PV Create
 	metadata := k8s.GetVolumeMetadata(req.GetParameters())
-	if delay, ok := req.GetParameters()["trashMaxDelay"]; ok {
-		metadata[trashMaxDelayKey] = delay
-	}
 	err = rbdVol.setAllMetadata(metadata)
 	if err != nil {
 		if deleteErr := rbdVol.Delete(ctx); deleteErr != nil {
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -93,13 +93,6 @@ const (
 	// Suffix added to the temp cloned image name.
 	// This will always be (rbd image name + "-temp").
 	tempImageSuffix = "-temp"
-
-	// trashMaxDelayKey is the key used to store the trash max delay in metadata.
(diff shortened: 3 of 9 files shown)


Ady0333:
Hello maintainers,
I’ve raised this PR to address issue #5830.
DCO has been fixed and the PR is ready for review whenever convenient.

Thank you for your time and for maintaining ceph-csi.

nixpanic:
Looks reasonable, but I wonder if it makes more sense to have this as a parameter in the StorageClass (and tunable by a VolumeAttributesClass)?

Ady0333:
Thank you for the feedback @nixpanic . I agree that making this tunable via StorageClass makes more sense.

I've updated the implementation to support trashMaxDelay as a parameter in the StorageClass. This value is now stored in the image metadata upon creation and used during deletion. If the parameter isn't present, it falls back to the existing global configuration for backward compatibility.

I've pushed the changes. Could you please review them when you have a moment?

internal/rbd/rbd_util.go
```diff
@@ -707,7 +710,25 @@ func (ri *rbdImage) Delete(ctx context.Context) error {
...
+			if v, parseErr := strconv.ParseUint(val, 10, 64); parseErr == nil {
+				trashDelay = uint(v)
+			}
+		}
+	} else if !errors.Is(err, librbd.ErrNotFound) {
+		log.WarningLog(ctx, "failed to open rbd image %q to check metadata: %v", ri, err)
+	}
+
+	// rbdTrashMaxDelay is the delay in seconds to delete the image from trash
+	err = rbdImage.Trash(uint64(trashDelay))
```
juliantaylor: this expects a duration not uint64

internal/rbd/rbd_util.go:731:23: cannot use uint64(trashDelay) (value of type uint64) as "time".Duration value in argument to rbdImage.Trash

internal/rbd/rbd_util.go
```diff
@@ -707,7 +710,25 @@ func (ri *rbdImage) Delete(ctx context.Context) error {
...
+		if metaErr == nil {
+			if v, parseErr := strconv.ParseUint(val, 10, 64); parseErr == nil {
+				trashDelay = uint(v)
+			}
+		}
+	} else if !errors.Is(err, librbd.ErrNotFound) {
+		log.WarningLog(ctx, "failed to open rbd image %q to check metadata: %v", ri, err)
+	}
+
+	// rbdTrashMaxDelay is the delay in seconds to delete the image from trash
```
juliantaylor: the ri.trashRemoveImage in line 742 will remove the item from the thrash explicitly
not sure why that is done, but with a retention time that probably should be skipped

juliantaylor (commented):
using a standard duration string to specify the retention should better than a unitless intenter

internal/rbd/controllerserver.go
```diff
@@ -445,6 +445,9 @@ func (cs *ControllerServer) CreateVolume(
 
 	// Set Metadata on PV Create
 	metadata := k8s.GetVolumeMetadata(req.GetParameters())
+	if delay, ok := req.GetParameters()["trashMaxDelay"]; ok {
+		metadata[trashMaxDelayKey] = delay
+	}
```
nixpanic: Storing this in the metadata (visible in the Ceph dashboard) isn't really appropriate. It is better to store it in the image-attributes, similar to `rbdVol.Owner`.

Ady0333:
Hello @nixpanic and @juliantaylor !! Greetings !
I apologize for the mistakes caused by me. I will work on the PR and the code again and make sure that there will be no further mistakes from my side.

Also thankyou guys for investing your time and reviewing my PR.

nixpanic:
> I apologize for the mistakes caused by me. I will work on the PR and the code again and make sure that there will be no further mistakes from my side.

Don't worry about making mistakes, that is how we learn!

Thanks for taking on the challenge :clap: Let us know if you have any doubts, questions or anything else we can assist you with.

Rakshith-R:
Hey @Ady0333 ,

CephCSI follows this design for volume deletion https://github.com/ceph/ceph-csi/blob/004532cf4668844bb0df41f1f947bab72125a54c/docs/design/proposals/rbd-snap-clone.md#delete-a-volume-pvc 

Move a image to trash and then add a task to remove it from trash.

The reason to add a task is because an image with linked child images cannot be deleted until all its children are either deleted or flattened.

Have you tested
- what will happen when you do move an image to trash with a particular duration and then a trash remove task runs on it within that duration? 

----

> the ri.trashRemoveImage in line 742 will remove the item from the thrash explicitly not sure why that is done, but with a retention time that probably should be skipped

@juliantaylor Retention time just guarantees that the image will be protected for sad duration, it does not make any comments regarding immediate deletion after the duration is completed.

We should definitely add a task to remove the image from trash to make sure this runs and does not leave stale.
Or User should assume responsibility for trash cleanup which I don't think is optimal.

juliantaylor:
>  We should definitely add a task to remove the image from trash to make sure this runs and does not leave stale.
Or User should assume responsibility for trash cleanup which I don't think is optimal.

imo purging the thrash should be the responsibility of ceph itself, not ceph-csi. so far I know it has a configurable trash purging schedule and the age you give images when moving to the thrash ist defines its minimum time it is retained (not the maximum)

Ady0333:
Hello @nixpanic and @juliantaylor !!
I’ve pushed an updated commit addressing all review feedback.

The implementation now:
- stores trashMaxDelay as an RBD image xattr (not metadata or CSI journal),
- parses duration strings using time.ParseDuration, and
- relies on Ceph to manage trash purging when a delay is set, preserving legacy behavior for zero-delay volumes.

Please take another look when you have a chance.

cmd/cephcsi.go
```diff
@@ -151,6 +151,12 @@ func init() {
 		"Minimum number of snapshots required on rbd image to start flattening")
 	flag.BoolVar(&conf.SkipForceFlatten, "skipforceflatten", false,
 		"skip image flattening if kernel support mapping of rbd images which has the deep-flatten feature")
+	flag.UintVar(
```
juliantaylor: flag.DurationVar should give you a Duration directly.

juliantaylor (commented):
I assume the omap values are still removed on pv deletion and would need to be reconstructed for a restore?
https://github.com/alvistack/ceph-ceph-csi/blob/devel/docs/resource-cleanup.md
this which is fine for me, as they can be externally backed up similar to the kubernetes objects.


Adding xattrs with the remove delay to the image itself feels overengineered to me, for the use case of data recovery it would be sufficient to just always pass the configured retention time to the Trash function, regardless of what the configuration was when the image was created.

Rakshith-R (commented):
Hey @Ady0333 ,

Let's keep this simple for now.

Use only cmd line arg `"rbdtrashmaxdelay"` input,
If it is set to a valid positive number, use it to direct send the image to Trash with that number and return immediately (no need to remove it from trash)

If not, retain current behavior.

Please add e2e for this new feature, 

Add a section in the end here https://github.com/ceph/ceph-csi/blob/982f0934732f9f03ed6393201197caf1c4225e1c/docs/rbd/deploy.md#L4
regarding this new feature explicitly mentioning that users need to take care of purging the trash while using this option.

With a note in https://github.com/ceph/ceph-csi/blob/982f0934732f9f03ed6393201197caf1c4225e1c/PendingReleaseNotes.md#L4

Ady0333:
Thankyou @Rakshith-R and @juliantaylor for the detailed feedback!!

That makes sense — I agree that keeping this as a simple, delete-time behavior is preferable. I’ll update the implementation to rely solely on the --rbdtrashmaxdelay CLI flag and always pass the configured value directly to the Trash() call, without persisting any retention-related metadata on the image.

I’ll also clean up the flag handling (using flag.DurationVar) and remove any unnecessary logic so that the behavior is:

delay > 0 → move image to trash with deferment

delay ≤ 0 → retain current behavior

Additionally, I’ll add documentation in docs/rbd/deploy.md clarifying the new option and explicitly noting that users are responsible for purging the RBD trash, along with an entry in PendingReleaseNotes.md.

Regarding E2E coverage, I’ll take a look at what can be reasonably added given the current test infrastructure, and if needed, follow up with a separate issue/PR.

Thanks again for the guidance — I’ll push an updated revision shortly.

Ady0333:
@Rakshith-R and @juliantaylor, I have updated the changes please go through them whenever you guys get time.

juliantaylor:
I have tested the patch applied against v3.15.1 and deleted images correctly get moved to the trash.

A minor concern is maybe the name of the argument, the max is a bit misleading as the maximum time it stays in the trash depends on the setup of the purging schedule (or manual purges) on the ceph side, trashmindelay may be better as it should be the minimum retention time for scheduled purges as long as it does not exceed the ceph side configured maximum retention time.
Maybe just name it trashdelay or imagetrashdelay with a note in the documentation that the actual retention time depends on ceph cluster settings.

Ady0333:
Thanks for testing and confirming the behavior @juliantaylor. 

Good point on the naming — I agree that max is misleading given that actual retention depends on Ceph-side purge scheduling. I’ll rename the flag to something clearer (e.g. --rbdtrashdelay) and update the documentation to clarify that it represents a minimum deferment before purge eligibility, with actual retention governed by Ceph settings.

I’ll push an updated revision shortly.

juliantaylor:
any updates on this?

PR #5899, opened 2026-01-06
rbd: fetch the volumeIds from VGRContent's status for omap data

Nikhil-Ladha:

Currently, we fetch the volumeId's from VGRContent.Spec.Source which has all the volumes that should be part of the group. But, it might happen that the initial modify call might fail due to some reason like network/config issue. In that case, the volumegroup omap will already be updated with the volumeIds and subsequent ModifyVolumeGroup call will fail as the volumes before and after Modify operation is same.
Therefore, we should fetch the volumeIds from the VGRContent status, which only gets updated post a succesfull modification.

### DIFF
--- a/internal/controller/volumegroup/volumegroupreplicationcontent.go
+++ b/internal/controller/volumegroup/volumegroupreplicationcontent.go
@@ -207,7 +207,16 @@ func (r *ReconcileVGRContent) reconcileVGRContent(ctx context.Context, obj runti
 
 	reqName := vgrc.Name
 	groupHandle := vgrc.Spec.VolumeGroupReplicationHandle
-	volumeIds := vgrc.Spec.Source.VolumeHandles
+	volumeRefs := vgrc.Status.PersistentVolumeRefList
+	volumeIds := make([]string, len(volumeRefs))
+	for i := range volumeRefs {
+		pv := &corev1.PersistentVolume{}
+		err := r.client.Get(ctx, types.NamespacedName{Name: volumeRefs[i].Name}, pv)
+		if err != nil {
+			return fmt.Errorf("failed to get persistentvolume %s: %w", volumeRefs[i].Name, err)
+		}
+		volumeIds[i] = pv.Spec.CSI.VolumeHandle
+	}
 
 	if groupHandle == "" {
 		return errors.New("volume group replication handle is empty")


internal/controller/volumegroup/volumegroupreplicationcontent.go
```diff
@@ -207,7 +207,16 @@ func (r *ReconcileVGRContent) reconcileVGRContent(ctx context.Context, obj runti
 
 	reqName := vgrc.Name
 	groupHandle := vgrc.Spec.VolumeGroupReplicationHandle
-	volumeIds := vgrc.Spec.Source.VolumeHandles
+	volumeRefs := vgrc.Status.PersistentVolumeRefList
+	volumeIds := []string{}
```
nixpanic: you can be a little more efficient:
```go
volumeIds := make([]string, len(volumeRefs))
...
   volumeIds[i] = pv.Spec.CSI.VolumeHandle
```

this prevents the need for `append()` to extend or re-allocate a new `volumeIds` slice.

nixpanic (commented):
I assume that when fetching the PV fails, a new reconcile is scheduled?

Nikhil-Ladha:
> I assume that when fetching the PV fails, a new reconcile is scheduled?

Yes, it will trigger a new reconcile.

internal/controller/volumegroup/volumegroupreplicationcontent.go
Nikhil-Ladha: Makes sense, updated the code now.

nixpanic (dismissed):

internal/controller/volumegroup/volumegroupreplicationcontent.go
```diff
@@ -207,7 +207,16 @@ func (r *ReconcileVGRContent) reconcileVGRContent(ctx context.Context, obj runti
...
 	reqName := vgrc.Name
 	groupHandle := vgrc.Spec.VolumeGroupReplicationHandle
-	volumeIds := vgrc.Spec.Source.VolumeHandles
+	volumeRefs := vgrc.Status.PersistentVolumeRefList
+	volumeIds := make([]string, len(volumeRefs))
+	for i := range volumeRefs {
+		pv := &corev1.PersistentVolume{}
+		err := r.client.Get(ctx, types.NamespacedName{Name: volumeRefs[i].Name}, pv)
+		if err != nil {
+			return fmt.Errorf("volume %s not found", volumeRefs[i].Name)
```
Madhu-1: Wrap the original error. The error may not be always not found,  keep it generic

Madhu-1 (changes requested):

internal/controller/volumegroup/volumegroupreplicationcontent.go
Nikhil-Ladha: Done.

nixpanic (approved):

Madhu-1 (approved):

iPraveenParihar:
@Mergifyio rebase

iPraveenParihar:
@Mergifyio queue

PR #5971, opened 2026-01-23
ci: delay checking ci/skip/multi-arch-build label

nixpanic:
When a GitHub Event is created, the labels of the PR are set to the
current values. Mergify will only set the labels, or copy the labels to
the PR once the PR is created. That means the initial GitHub Event does
not have the right labels set yet.

In order to really skip the multi-arch-build, and run single-arch-build
instead, the label needs to be checked from the PR, instead of compared
from the labels in the GitHub Event.

### DIFF
--- a/.github/workflows/build-multi-stage.yaml
+++ b/.github/workflows/build-multi-stage.yaml
@@ -19,16 +19,37 @@ jobs:
     name: multi-arch-build
     runs-on: ubuntu-latest
     steps:
+      - name: delay a minute to have labels added
+        run: sleep 1m
+      - name: detect single-arch-build
+        # yamllint disable-line rule:line-length
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd  # v8.0.0
+        id: detect-single-arch-build
+        with:
+          script: |
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


nixpanic:
Tested a lot, and this seems to work now. See the attempts of the last runs at https://github.com/ceph/ceph-csi/actions/runs/21289467990/job/61279172064?pr=5971

.github/workflows/build-multi-stage.yaml
```diff
@@ -19,16 +19,29 @@ jobs:
     name: multi-arch-build
     runs-on: ubuntu-latest
     steps:
+      - name: detect single-arch-build
+        id: detect-single-arch-build
+        # wait 1 minute because setting labels may be delayed
+        run: |
+          sleep 1m
+          PR=${{ github.event.pull_request.number }}
```
black-dragon74: Why not use: https://github.com/actions/github-script?

nixpanic: I was not aware that existed :smile: 

Will have a look at it and see if that makes things easier.

nixpanic: @black-dragon74 , please have a look again. Now using github-script and testeb with and without the label.

black-dragon74 (approved):
Thank you!

Rakshith-R (approved):

PR #5974, opened 2026-01-26
nvmeof: treat "connecting" state as valid in path detection

### LINKED ISSUE #5964: Already Connected in Nodeserver
```bash
E0122 12:39:03.281079  105266 cephcmds.go:157] ID: 33 Req-ID: 0001-0011-openshift-storage-0000000000000002-a617ce6d-9918-4451-98e9-a2096c57a456 an error 
(exit status 1) and stderror (already connected) occurred while running nvme args: 
[connect -t tcp -n nqn.2016-06.io.ceph:subsystem.test-integration -a 10.128.2.52 -s 4420 -l 1800 \
--hostnqn nqn.2014-08.org.nvmexpress:gdidi-gwfsm-worker-3-ngbpx 
W0122 12:39:03.281121  105266 nvmeof_initiator.go:203] ID: 33 Req-ID: 
0001-0011-openshift-storage-0000000000000002-a617ce6d-9918-4451-98e9-a2096c57a456 Failed to connect to 10.128.2.52:4420 - stdout: , stderr: already connected
```
It should not raise an error\print error. 
Need to add already connected handling in the nodeserver.

gadididi:
When checking if a path to a gateway already exists, treat both "live" and "connecting" states as
valid connections that should not be re-attempted.

The "connecting" state indicates the NVMe kernel
is actively trying to establish or re-establish
a connection, which occurs in scenarios like:

- Initial connection establishment
- Gateway temporarily unavailable and kernel retrying
- Subsystem\Host deleted and recreated on the gateway

The kernel's ctrl_loss_tmo mechanism will continue retry attempts for up to 30 minutes
( by `-l 1800` param in `nvme connect` command).
Attempting nvme connect while a path is in
"connecting" state results in "already connected"
errors and can cause volume attachment failures
during create/delete cycles.

By treating "connecting" as a valid state,
we allow the kernel's retry logic to handle
reconnection automatically without interference.

## Related issues ##

Fixes: https://github.com/ceph/ceph-csi/issues/5964

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/nvmeof/nvmeof_initiator.go
+++ b/internal/nvmeof/nvmeof_initiator.go
@@ -158,7 +158,7 @@ func (ni *nvmeInitiator) ConnectSubsystem(ctx context.Context, req *ConnectReque
 
 		// Check if already connected to this specific gateway
 		if req.HostNQN != "" && existingConnections != nil {
-			if existingConnections.hasLivePathToGateway(
+			if existingConnections.hasPathToGateway(
 				req.SubsystemNQN, req.HostNQN, listener.Address, portStr) {
 				log.DebugLog(ctx, "Already connected to subsystem %s via %s:%s with HostNQN %s",
 					req.SubsystemNQN, listener.Address, portStr, req.HostNQN)
@@ -247,8 +247,8 @@ func listSubsystems(ctx context.Context) (nvmeHostConnections, error) {
 	return hosts, nil
 }
 
-// hasLivePathToGateway checks if a live path exists to the specified gateway.
-func (nhc nvmeHostConnections) hasLivePathToGateway(subsystemNQN, hostNQN,
+// hasPathToGateway checks if a path exists to the specified gateway.
+func (nhc nvmeHostConnections) hasPathToGateway(subsystemNQN, hostNQN,
 	gatewayIP, gatewayPort string,
 ) bool {
 	for _, host := range nhc {
@@ -261,10 +261,25 @@ func (nhc nvmeHostConnections) hasLivePathToGateway(subsystemNQN, hostNQN,
 				continue
 			}
 
+			// loop through paths to find matching path
 			for _, path := range subsys.Paths {
+				// Check if the path matches the gateway IP and port
+				// and is in a usable state:
+				// - "live": connection is active and working
+				// - "connecting": kernel is actively trying to (re)connect
+				//
+				// The "connecting" state occurs when:
+				// 1. Initial connection is being established
+				// 2. Connection lost and kernel is retrying (ctrl_loss_tmo in effect)
+				// 3. Subsystem was deleted/recreated on the gateway
+				//
+				// In all cases, the kernel's retry mechanism handles reconnection
+				// for up to ctrl_loss_tmo seconds, so we should not attempt another
+				// connection which would fail with "already connected" error.
 				if path.Address.Traddr == gatewayIP &&
 					path.Address.Trsvcid == gatewayPort &&
-					path.State == "live" {
+					(path.State == "live" ||
+						path.State == "connecting") {
 					return true
 				}
 			}
--- a/internal/nvmeof/nvmeof_test.go
+++ b/internal/nvmeof/nvmeof_test.go
@@ -317,7 +317,7 @@ func TestHasLivePathToGateway(t *testing.T) {
 			hostNQN:      "nqn.2014-08.org.nvmexpress:uuid:test-host",
 			gatewayIP:    "10.129.2.45",
 			gatewayPort:  "4420",
-			want:         false,
+			want:         true,
 		},
 		{
 			name:         "wrong gateway IP",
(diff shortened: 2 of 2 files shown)


internal/nvmeof/nvmeof_initiator.go
```diff
@@ -261,10 +261,25 @@ func (nhc nvmeHostConnections) hasLivePathToGateway(subsystemNQN, hostNQN,
...
+				// 3. Subsystem was deleted/recreated on the gateway
+				//
+				// In all cases, the kernel's retry mechanism handles reconnection
+				// for up to ctrl_loss_tmo seconds, so we should not attempt another
+				// connection which would fail with "already connected" error.
 				if path.Address.Traddr == gatewayIP &&
 					path.Address.Trsvcid == gatewayPort &&
-					path.State == "live" {
+					(path.State == "live" ||
+						path.State == "connecting") {
```
Madhu-1, line 282: Do we need to retry if its in connecting state until it goes to live?

gadididi: No, we don't need to retry. Here's why:
The kernel is already doing the retrying for us. When a path is in "connecting" state, the NVMe kernel driver is actively attempting to establish the connection and will keep retrying automatically for up to 30 minutes (based on the `ctrl_loss_tmo` we set with `-l 1800` in nvme connect command).

From the CSI driver's perspective, our job is just to ensure a connection attempt has been initiated for the next step- mounting. 
Once we see a path in either "live" or "connecting" state for our subsystem, we know the kernel has it under control.
What happens in practice:
this function is called from the nodeserver in `NodeStageVolume()`. right after `ControllerPublishVolume()` added (if was need to add) the host (by GRPC AddHost to nvmeof GW)

If the path is "live" --> great, connection is working
If the path is "connecting" --> it means there is a path but it sis not connected (probably due to remove last nvmeof ns , So the `ControllerUnPublishVolume()` removed the host and the subsystem) kernel is retrying, and when the `ControllerPublishVolume()` added the host again, it becomes available but it is too fast get into this check. the nvme will automatically transition to "live" and the namespace will appear.

another example.
1. there was connection with live path for some nvmeof ns . then the admin deletes that pod and the pvc. the host and the subsystem both are deleted. 
2. 30 minutes passed
3. admin wants to create new pvc and then new pod. 
4. the `nvme subsys-list` will not see connecting or live. 
5. nvme connect will rerun again


actually I could anyway to rerun connect, without this checking. if it is "live" or "connecting" I would get an error "Already Connected" So basically, I can parse the error code and stderr and see if this is the case, move on, and do not raise an error.
@Madhu-1 what do you think?

Madhu-1: A couple of questions

* if its in the connecting state, will the application be able to read/write? I assume no. What is the use of mounting such PVC if it cannot be used?
* Will there be any problem during NodeUnstage? 

>actually I could anyway to rerun connect, without this checking. if it is "live" or "connecting" I would get an error "Already Connected" So basically, I can parse the error code and stderr and see if this is the case, move on, and do not raise an error.

If  the `nvme subsys-list` is not reliable, we should rerun the connect and discard known errors

gadididi: > if its in the connecting state, will the application be able to read/write? I assume no. What is the use of mounting such PVC if it cannot be used?

The "connecting" state we're checking happens right after ControllerPublishVolume stage, and before the mount stage.
Here's the timeline:

1. CreateVolume - Volume is created on the gateway
2. ControllerPublishVolume - Host is added to the subsystem on the gateway
3. NodeStageVolume - This is where we check for existing connections.
  At this point, we call nvme list-subsys to see what's already connected
  If we see "connecting", it means a previous operation already initiated the connection (some prev PVC that's already deleted) 

4. The kernel continues establishing the connection in the background
5. Before we actually mount, we wait for the namespace device to appear using `GetNamespaceDeviceByUUID()` which retries until the device is accessible
6. Only when the device exists and is accessible do we proceed to format and mount it

So, the key point is: We don't mount a PVC that's stuck in "connecting" state. We just avoid running nvme connect again (which would fail with "already connected"). We still wait for the device to become available before mounting.

> If the nvme subsys-list is not reliable, we should rerun the connect and discard known errors

I dont think it is not reliable.
at some runs the  `nvme subsys-list` already shows `live` . I think it is about timing. but for sure (and easier way)  to parse the stderr is fit solution. 
But I would avoid parsing the string in stderr in some reason the next nvme-tool version will return another sting.

nixpanic (approved):

internal/nvmeof/nvmeof_initiator.go
nixpanic: I'm happy with this approach for now. If there is a need to always try to connect, we can do that at a later point in time too.

Madhu-1 (approved):

Madhu-1:
@mergifyio queue

Issue #6037, opened 2026-02-09
nvmeof: ListListener grpc call returns empty


gadididi:
## Describe the bug #

When we are creating a subsystem with AutoListener feature,  the nListListener grpc call (that is called right after) returns empty listener list.
But if we query (via GW cli) it returns right answer.

## Environment details #

- Image/version of Ceph CSI driver : `canary`
- Kubernetes cluster version : non relevant 
- Ceph cluster version : ceph `main` (which has `AutoListener` support)

## Steps to reproduce #

1. create rook env and deploy the main ceph version
2. deploy nvmeof csi provisioner
3. create pvc
4. you will see in the log the return list of listeners from the GW is empty. 
5. deploy the node plugin 
**6. create test-pod - you will get an error-- the list of listeners is empty!** 

## Actual results #
There is no error handling in the provisioner side because we cannot know which listeners (GWs) really exist in this cluster.
Then just when you create test-pod you will catch this error. 

Describe what happened

# Expected behavior #

## Logs #

- csi-provisioner: 

```bash 
I0208 15:25:22.975469       1 utils.go:350] ID: 15 Req-ID: pvc-6ad5a7a7-3c82-4afd-87a2-e0be3651968f GRPC call: /csi.v1.Controller/CreateVolume
I0208 15:25:22.975594       1 utils.go:351] ID: 15 Req-ID: pvc-6ad5a7a7-3c82-4afd-87a2-e0be3651968f GRPC request: {"capacity_range":{"required_bytes":134217728},"name":"pvc-6ad5a7a7-3c82-4afd-87a2-e0be3651968f","parameters":{"clusterID":"rook-ceph","csi.storage.k8s.io/pv/name":"pvc-6ad5a7a7-3c82-4afd-87a2-e0be3651968f","csi.storage.k8s.io/pvc/name":"nvmeof-external-volume","csi.storage.k8s.io/pvc/namespace":"default","imageFeatures":"layering,deep-flatten,exclusive-lock,object-map,fast-diff","imageFormat":"2","networkMask":"10.128.0.0/14","nvmeofGatewayAddress":"172.30.113.58","nvmeofGatewayPort":"5500","pool":"nvmeof","subsystemNQN":"nqn.2016-06.io.spdk:cnode1.rook-ceph"},"secrets":"***stripped***","volume_capabilities":[{"access_mode":{"mode":"SINGLE_NODE_WRITER"},"mount":{"fs_type":"ext4"}}]}
..
..
[...]

PR #6038, opened 2026-02-09
nvmeof: add retry mechanism for listenerList

gadididi:

adding retry mechanism for listenerList grpc call
because there is unknown (under investigation) issue in the nvmeof gw, when user creates subsystem with autoListener option (it creates the listener automatically) but then, when the controller wants to retrieve the listeners list, it gets empty list. there is some
delay in the GW.
FYI- the listeners list comes from the ceph mon command `nvme-gw listeners`  . 

## Related issues ##

fixed: 
https://github.com/ceph/ceph-csi/issues/6037

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/nvmeof/nvmeof.go
+++ b/internal/nvmeof/nvmeof.go
@@ -23,6 +23,7 @@ import (
 	"math/big"
 	"syscall"
 
+	"github.com/avast/retry-go/v4"
 	pb "github.com/ceph/ceph-nvmeof/lib/go/nvmeof"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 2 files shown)


internal/nvmeof/nvmeof.go
```diff
@@ -536,6 +537,30 @@ func ConvertListenersFromProto(protoListeners []*pb.ListenerInfo) []ListenerDeta
...
+				return nil, fmt.Errorf("failed to list auto-created listeners: %w", err)
+			}
+
+			if len(autoListeners.GetListeners()) == 0 {
+				return nil, fmt.Errorf("no auto-listeners found for subsystem %s", subsystemNQN)
+			}
+
+			return ConvertListenersFromProto(autoListeners.GetListeners()), nil
+		},
+		retry.Attempts(6), // ~100ms, 200ms, 400ms, 800ms, 1.6s, 3.2s = ~6.3s total
```
Madhu-1, line 560: @gadididi will this work in all the clusters where we have 100 pvc etc or if ceph cluster is under some small stress? because retries are always very tricky

gadididi: I will try to check it soon

gadididi:
This PR was work around for issue when user uses network_mask -> means the nvmeof GW creates the listener automatically -> needs to call lister_list to get the list and send it to nodeserver (he needs to know what IP connect to) -> the GW retrieve the "auto listeners" from monitor which has some delay. 
needs to investigate it, but this PR is not the solution for it.

PR #6048, opened 2026-02-10
util: add support for GKLM KMS over KMIP

black-dragon74:

This patch adds support for GKLM KMS using the KMIP protocol.

Details specific to GKLM:
- Get RPC to encrypt and decrypt the KEK for LUKS.
- KMS' `USE_CRYPTO_RPC` is used to distinguish between which RPC to
  call.
  - If `USE_CRYPTO_RPC` is true, `Encrypt/Decrypt` RPCs are called.
  - If `USE_CRYPTO_RPC` is false, `Get` RPC is called to fetch the key
    by its UUID and encryption/decryption is done locally.

### DIFF
--- a/docs/design/proposals/encryption-with-gklm.md
+++ b/docs/design/proposals/encryption-with-gklm.md
@@ -1,0 +1,155 @@
+# Encrypted volumes with IBM GKLM
+
+IBM Guardium Key Lifecycle Manager (GKLM) is a centralized key management system
+that helps manage encryption keys and certificates across the enterprise. GKLM
+supports the Key Management Interoperability Protocol (KMIP), an industry-standard
+protocol for communication with key management systems. To support this KMS
+integration in Ceph-CSI and enable GKLM users to perform volume encryption
+operations, the following details are provided.
+
+## Connection to IBM GKLM service
+
+GKLM uses the KMIP protocol for communication. The following parameters/values
+can be used to establish the connection to the GKLM service from the CSI driver
+and to perform encryption operations:
+
+```text
+* KMS_PROVIDER
+Must be set to "kmip" to use the KMIP protocol.
+
+* USE_CRYPTO_RPC
+Indicates whether cryptographic operations should be handled by the KMS server.
+The KMS provider must support the Encrypt and Decrypt RPC methods defined
+in the KMIP specification.
+For GKLM, this setting MUST be `false`.
+
+* KMIP_ENDPOINT
+The GKLM server endpoint address (e.g., "gklm-server:5696").
+
+* KMIP_SECRET_NAME
+Name of the Kubernetes Secret containing the credentials for
+communicating with the GKLM server. Defaults to "ceph-csi-kmip-credentials".
+
+* KMS_SERVICE_NAME (optional)
+A unique name for the key management service within the project.
+
+* TLS_SERVER_NAME (optional)
+The endpoint server name. Useful when the GKLM endpoint does not have
+a DNS entry. This SAN on the returned certificate must match this.
+
+* READ_TIMEOUT (optional)
+Network read timeout, in seconds. The default value is 10.
+
+* WRITE_TIMEOUT (optional)
+Network write timeout, in seconds. The default value is 10.
+```
+
+### Values provided in the connection Secret
+
+The Secret contains TLS certificates for secure communication with the GKLM
+server and the unique identifier of the encryption key:
+
+```text
+* CA_CERT
+The GKLM server's certificate that is marked as key serving certificate.
+
+* CLIENT_CERT
+Client certificate that will be used to connect to the GKLM server.
+
+* CLIENT_KEY
+Client key that will be used to connect to the GKLM server.
+
+* UNIQUE_IDENTIFIER
+UUID of the symmetric key pair in GKLM to use for encryption/decryption.
+```
+
+The Ceph-CSI KMS plugin interface for GKLM will read the Secret name from the
+KMS ConfigMap and fetch these values.
+
+### Values provided in the ConfigMap
+
+The KMS configuration is stored in a ConfigMap (typically named
+`ceph-csi-encryption-kms-config`) with the following structure:
+
+```json
+{
+  "gklm-test": {
+    "KMS_PROVIDER": "kmip",
+    "KMS_SERVICE_NAME": "gklm",
+    "USE_CRYPTO_RPC": "false",
+    "KMIP_ENDPOINT": "sklmapp.sklm.svc.cluster.local:5696",
+    "KMIP_SECRET_NAME": "ceph-csi-kmip-credentials",
+    "TLS_SERVER_NAME": "sklmapp.sklm.svc.cluster.local",
+    "READ_TIMEOUT": 10,
+    "WRITE_TIMEOUT": 10
+  }
+}
+```
+
+### Storage class configuration
+
+As with other KMS integrations, the StorageClass must be configured for
+encryption with the `encryptionKMSID` parameter matching the key in the KMS ConfigMap:
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-rbd-sc-gklm
+parameters:
+  encrypted: "true"
+  encryptionKMSID: "gklm-test" # This matches the JSON key for GKLM config above
+  # ... other parameters
+```
+
+## Prerequisites
+
+Before using GKLM for volume encryption, you must:
+
+1. Create a symmetric key pair (256-bit) in the GKLM server
+1. Obtain the UUID of this key
+1. Provide this UUID as the `UNIQUE_IDENTIFIER` in the Kubernetes Secret
+1. Ensure `USE_CRYPTO_RPC` is set to `"false"` in the KMS ConfigMap
+
+## Volume Encrypt and Decrypt Operations
+
+GKLM integration differs from traditional KMIP implementations in how it handles
+encryption and decryption:
+
+This approach uses GKLM's `Get` RPC operations instead of the
+`Encrypt/Decrypt` operations used by other KMIP servers. The actual
+encryption/decryption of the DEK happens locally using the key
+fetched from GKLM by its UUID.
+
+**Note:** The above will only apply if `USE_CRYPTO_RPC` is set to false in
+the KMS ConfigMap.
+
+### Encryption Process
+
+1. The CSI driver connects to the GKLM server using KMIP protocol and TLS
+   certificates
+1. It fetches the encryption key from GKLM using the `Get` operation with the
+   provided `UNIQUE_IDENTIFIER`
+1. A Data Encryption Key (DEK) is generated locally for LUKS encryption
+1. The DEK is encrypted locally using AES-256-GCM with the fetched encryption key
+1. The encrypted DEK and nonce are stored in the RBD image metadata (similar to
+   other KMS integrations)
+
+### Decryption Process
+
+1. The encrypted DEK and nonce are retrieved from the RBD image metadata
+1. The encryption key is fetched from GKLM using the `Get` operation
+1. The DEK is decrypted locally using AES-256-GCM with the fetched key and nonce
+1. The decrypted DEK is used for LUKS volume decryption
+
+## Integration Protocol
+
+GKLM integration uses the [KMIP protocol](https://en.wikipedia.org/wiki/Key_Management_Interoperability_Protocol)
+for communication. The [KMIP Go library](https://github.com/gemalto/kmip-go)
+provides the client SDK to interact with the KMIP server and perform key
+management operations.
+
+## Additional References
+
+- [IBM Guardium Key Lifecycle Manager Documentation](https://www.ibm.com/docs/en/gklm)
+- [KMIP Specification](https://docs.oasis-open.org/kmip/spec/v1.4/kmip-spec-v1.4.html)
--- a/internal/kms/kmip.go
+++ b/internal/kms/kmip.go
@@ -20,6 +20,8 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/aes"
+	"crypto/cipher"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
@@ -73,6 +75,14 @@ const (
 	kmipCLientCert       = "CLIENT_CERT"
 	kmipClientKey        = "CLIENT_KEY"
 	kmipUniqueIdentifier = "UNIQUE_IDENTIFIER"
+
+	// A config boolean stating the RPCs to use for encrypt
+	// and decrypt operations.
+	//
+	// When set to true: ENCRYPT and DECRYPT RPCs are used.
+	// Otherwise: GET RPC is used to fetch the key referenced
+	// by `UNIQUE_IDENTIFIER` and crypto is performed locally.
+	kmipUseCryptoRPC = "USE_CRYPTO_RPC"
 )
 
 var _ = RegisterProvider(Provider{
@@ -91,6 +101,7 @@ type kmipKMS struct {
 	uniqueIdentifier string
 	readTimeout      uint8
 	writeTimeout     uint8
+	useCryptoRPC     bool
 }
 
 func initKMIPKMS(args ProviderInitArgs) (EncryptionKMS, error) {
@@ -108,6 +119,13 @@ func initKMIPKMS(args ProviderInitArgs) (EncryptionKMS, error) {
 
 	err = setConfigString(&kms.endpoint, args.Config, kmipEndpoint)
 	if err != nil {
+		return nil, err
+	}
+
+	// optional
+	kms.useCryptoRPC = true
+	err = setConfigBoolean(&kms.useCryptoRPC, args.Config, kmipUseCryptoRPC)
+	if errors.Is(err, errConfigOptionInvalid) {
 		return nil, err
 	}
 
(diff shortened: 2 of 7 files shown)


black-dragon74:
## Testing

### Setup the KMS
```text
❯ oc get secrets gklm-kms -oyaml
apiVersion: v1
data:
  CA_CERT: XXXX
  CLIENT_CERT: XXXX
  CLIENT_KEY: XXXX
  UNIQUE_IDENTIFIER: S0VZ....MWE2NGVkMmNm <== This is UID of the key in KMS, set by ODF console
kind: Secret
metadata:
  name: gklm-kms
  namespace: openshift-storage
type: Opaque

❯ oc get cm csi-kms-connection-details -oyaml
apiVersion: v1
data:
  gklm: '{
  "KMS_PROVIDER": "kmip",
  "KMS_SERVICE_NAME": "gklm", 
  "KMIP_ENDPOINT": "sklmapp.sklm.svc.cluster.local:5696",
  "KMIP_SECRET_NAME": "gklm-kms",
  "USE_CRYPTO_RPC": "false", <== UI/Docs will ensure its consistency, MUST be "false" for GKLM
  "TLS_SERVER_NAME": "sklmapp.sklm.svc.cluster.local"
}'
kind: ConfigMap
metadata:
  name: csi-kms-connection-details
  namespace: openshift-storage
```

### Check PVC and Pods are running:
```
❯ oc -n sklm get po
NAME                       READY   STATUS    RESTARTS   AGE
sklmapp-6456fcd4c9-qzpjn   1/1     Running   0          22h
ubuntu-pod                 1/1     Running   0          42s
```

### Check image metadata, before and after
```
// Before
❯ oc exec -it po/rook-ceph-tools-689bfdf6d7-x6hjf -- rbd image-meta ls csi-vol-92c21a8d-73f2-465b-a584-f850a53c75d1 -p ocs-storagecluster-cephblockpool
There are 7 metadata on this image:

Key                               Value
csi.ceph.com/cluster/name         939a690f-4997-4952-90ee-7bac0156cadd
csi.storage.k8s.io/pv/name        pvc-2f367d4e-4887-4c35-bb61-89128e36a2d0
csi.storage.k8s.io/pvc/name       gklm-pvc
csi.storage.k8s.io/pvc/namespace  sklm
rbd.csi.ceph.com/dek              {"dek":"S19PUFFfT0JKLTVkMDI5MWEtY2UyMGM4NWEtMWI3Mi00YWU0LTk4MmQtNDJiMmQ1N2JhNzgz","nonce":"TkE="}
rbd.csi.ceph.com/encrypted        encryptionPrepared
rbd.csi.ceph.com/luks2HeaderSize  16777216

// After
❯ oc exec -it po/rook-ceph-tools-689bfdf6d7-x6hjf -- rbd image-meta ls csi-vol-92c21a8d-73f2-465b-a584-f850a53c75d1 -p ocs-storagecluster-cephblockpool
There are 8 metadata on this image:

Key                                                                                                                                 Value                                                                 
.rbd.csi.ceph.com/userid/0001-0011-openshift-storage-0000000000000003-92c21a8d-73f2-465b-a584-f850a53c75d1/perf7-vd2vw-ocs-0-49z42  csi-rbd-node-ceph-user-g1-7b3ce12c-6b61-4edf-8964-336786573717        
csi.ceph.com/cluster/name                                                                                                           939a690f-4997-4952-90ee-7bac0156cadd                                  
csi.storage.k8s.io/pv/name                                                                                                          pvc-2f367d4e-4887-4c35-bb61-89128e36a2d0                              
csi.storage.k8s.io/pvc/name                                                                                                         gklm-pvc                                                              
csi.storage.k8s.io/pvc/namespace                                                                                                    sklm                                                                  
rbd.csi.ceph.com/dek                                                                                                                {"dek":"S19PUFFfT0JKLTVkMDI5MWEtY2UyMGM4NWEtMWI3Mi00YWU0LTk4MmQtNDJiMmQ1N2JhNzgz","nonce":"TkE="}
rbd.csi.ceph.com/encrypted                                                                                                          encrypted                                                             
rbd.csi.ceph.com/luks2HeaderSize                                                                                                    16777216 

```

nixpanic:
What is _GKLM_? Is that a particular Open Source project, or a product from some vendor? Any documentation/example updates that would benefit users?

internal/kms/kmip.go
```diff
@@ -39,6 +43,12 @@ import (
 const (
 	kmsTypeKMIP = "kmip"
 
+	// a unique identifier for GKLM kms.
+	kmsGKLM = "gklm"
+
+	// a unique identifier for the KMS service.
+	kmsServiceName = "KMS_SERVICE_NAME"
```
Rakshith-R: Let's use a generic kmip option like the following ?

maybe use USE_ENCRYPT_DECRYPT_RPC set to true by default ? 
or something else

```
	kmipEndpoint      = "KMIP_ENDPOINT"
	kmipTLSServerName = "TLS_SERVER_NAME"
	kmipReadTimeOut   = "READ_TIMEOUT"
	kmipWriteTimeOut  = "WRITE_TIMEOUT"

	// The following options are part of the Kubernetes Secrets.
	//
	// #nosec:G101, value not credential, just configuration keys.
	kmipSecretNameKey    = "KMIP_SECRET_NAME"
	kmipCACert           = "CA_CERT"
	kmipCLientCert       = "CLIENT_CERT"
	kmipClientKey        = "CLIENT_KEY"
	kmipUniqueIdentifier = "UNIQUE_IDENTIFIER"
```

Rakshith-R:
> What is _GKLM_? Is that a particular Open Source project, or a product from some vendor? Any documentation/example updates that would benefit users?

@nixpanic this change will support all kms that supports kmip but do not support the encrypt decrypt rpcs (one of them being GKLM).

internal/kms/kmip.go
```diff
@@ -495,6 +584,107 @@ func (kms *kmipKMS) verifyResponse(
 	return &batchItem, nil
 }
 
+// symmetricEncrypt seals the plainText DEK using the given key and returns
+// `encryptedMetedataDEK` with nonce.
+func symmetricEncrypt(plainDEK string, key []byte) (*encryptedMetedataDEK, error) {
```
Rakshith-R: replace this too with this helper ? 

https://github.com/ceph/ceph-csi/blob/5dd6d09a3a2dad3b23a8a91c7a3780285e224585/internal/kms/secretskms.go#L210-L235

black-dragon74: I saw that but in case of AWS we derive the key and then use that for AES enc/dec. In case of KMIP, we get the key beforehand. 

We can derive the key ourselves before calling `symmetricEncrypt`. But I'll include those changes as a part of separate PR. 

Let's keep this PR GKLM specific to keep things simpler?

internal/kms/kmip.go
black-dragon74: `USE_CRYPTO_RPC` might serve the purpose here. 

But, `KMS_SERVICE_NAME` can help us in the future if some specifics are to be handled wrt a KMS vendor? Better to know what exactly are we dealing with than to know the supported RPCs, no?

Other benefit of using `KMS_SERVICE_NAME` is that we don't need any extra config options. UI already sets this OOB.

docs/design/proposals/encryption-with-gklm.md
```diff
@@ -0,0 +1,148 @@
+# Encrypted volumes with IBM GKLM
+
+IBM Guardium Key Lifecycle Manager (GKLM) is a centralized key management system
+that helps manage encryption keys and certificates across the enterprise. GKLM
+supports the Key Management Interoperability Protocol (KMIP), an industry-standard
+protocol for communication with key management systems. To support this KMS
+integration in Ceph CSI and enable GKLM users to perform volume encryption
```
nixpanic: Use the spelling **Ceph-CSI**, with a - instead of a space.

docs/design/proposals/encryption-with-gklm.md
```diff
@@ -0,0 +1,148 @@
...
+   other KMS integrations)
+
+### Decryption Process
+
+1. The encrypted DEK and nonce are retrieved from the RBD image metadata
+1. The symmetric key is fetched from GKLM using the `Get` operation
+1. The DEK is decrypted locally using AES-256-GCM with the fetched key and nonce
+1. The decrypted DEK is used for LUKS volume decryption
+
+This approach uses GKLM's `Register/Get` RPC operations instead of the
```
nixpanic: move this paragraph a  bit further up

Is there a way to detect this alternative for Encrypt/Decrypt?

internal/kms/kmip.go
```diff
@@ -111,6 +122,11 @@ func initKMIPKMS(args ProviderInitArgs) (EncryptionKMS, error) {
 		return nil, err
 	}
 
+	err = setConfigString(&kms.serviceName, args.Config, kmsServiceName)
```
iPraveenParihar: What about the existing KMIP users? They will not be having the `KMS_SERVICE_NAME` detail and will fail.

This should be made optional

internal/rbd/encryption.go
```diff
@@ -64,7 +64,7 @@ const (
 	// luks2 header size metadata key.
 	luks2HeaderSizeKey = "rbd.csi.ceph.com/luks2HeaderSize"
 
-	encryptionPassphraseSize = 20
+	encryptionPassphraseSize = 32
```
iPraveenParihar: This will affect all the KMSes right? The commit message tells that this change is specific to GKLM.
This should be explicitly called out

internal/kms/kmip.go
```diff
@@ -234,8 +288,55 @@ func (kms *kmipKMS) EncryptDEK(ctx context.Context, _, plainDEK string) (string,
 	return string(emdData), nil
 }
 
-// DecryptDEK uses the KMIP decrypt operation  to decrypt the DEK.
-func (kms *kmipKMS) DecryptDEK(ctx context.Context, _, encryptedDEK string) (string, error) {
+// encryptDEKUsingRemoteKey uses the KMIP get operation to fetch the key and encrypts DEK using GCM.
+func (kms *kmipKMS) encryptDEKUsingRemoteKey(_ context.Context, _, plainDEK string) (string, error) {
+	// Fetch the key from the KMS
+	key, err := kms.getKey(kms.uniqueIdentifier)
```
iPraveenParihar, line 296: Will the key length always be 16, 24 or 32 bytes?
Because down the line aes.NewCipher(key) accepts the key to be of only those length.

docs/design/proposals/encryption-with-gklm.md
```diff
@@ -0,0 +1,148 @@
...
+The GKLM server's certificate that is marked as key serving certificate.
+
+* CLIENT_CERT
+Client certificate that will be used to connect to the GKLM server.
+
+* CLIENT_KEY
+Client key that will be used to connect to the GKLM server.
+
+* UNIQUE_IDENTIFIER
+UUID of the symmetric RSA key pair in GKLM to use for encrypting/decrypting.
```
iPraveenParihar: What does this mean? RSA is fundamentally an asymmetric. Did you mean AES?

docs/design/proposals/encryption-with-gklm.md
iPraveenParihar: In the code implementation I could only see the `OperationGet` being performed? No use of `OperationRegister`

internal/kms/kmip.go
```diff
@@ -495,6 +584,107 @@ func (kms *kmipKMS) verifyResponse(
...
+		return nil, fmt.Errorf("failed to create a cipher from the given key: %w", err)
+	}
+
+	gcm, err := cipher.NewGCM(block)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create GCM: %w", err)
+	}
+
+	// Generate a nonce
+	nonce := make([]byte, gcm.NonceSize())
```
iPraveenParihar: Can you consider reusing the `generateNonce()` from here? https://github.com/ceph/ceph-csi/blob/2407b245873b23391ec10d224cc979b98e5b33d5/internal/kms/secretskms.go#L300-L308

internal/kms/kmip.go
black-dragon74: Done in a separate commit.

internal/rbd/encryption.go
black-dragon74: This is technically not required anymore (since we do crypto locally). WDYT, should we keep it at 32 or remove this change?

internal/kms/kmip.go
black-dragon74: Done.

internal/kms/kmip.go
black-dragon74: Yes.

docs/design/proposals/encryption-with-gklm.md
black-dragon74: Mixed up the types of key and local operations. Fixed.

internal/kms/kmip.go
black-dragon74: Removed this, there is a new config option called `USE_CRYPTO_RPC`

docs/design/proposals/encryption-with-gklm.md
black-dragon74: Artifact of previous implementation, removed.

docs/design/proposals/encryption-with-gklm.md
black-dragon74: Done

docs/design/proposals/encryption-with-gklm.md
```diff
@@ -0,0 +1,158 @@
+# Encrypted volumes with IBM GKLM
```
Rakshith-R, line 1: Let's just add the optional USE_CRYPTO_RPC point here ? 

https://github.com/ceph/ceph-csi/blob/devel/docs/rbd/deploy.md#configuring-kmip-kms

We don't need to mention ibm gklm in cpehcsi docs.

internal/rbd/encryption.go
Rakshith-R: Let's keep only required changes

internal/kms/kmip.go
```diff
@@ -111,6 +122,13 @@ func initKMIPKMS(args ProviderInitArgs) (EncryptionKMS, error) {
 		return nil, err
 	}
 
+	// optional
+	kms.useCryptoRPC = false
```
Rakshith-R: The default should be true ? 
for backward compatibility.

Rakshith-R (changes requested):

internal/kms/kmip.go
black-dragon74: Ah, nice catch. Yes, it was intended to be like that.

docs/design/proposals/encryption-with-gklm.md
black-dragon74: And get rid of this doc entirely?

Rakshith-R: We have not mentioned any KMS here too
https://github.com/ceph/ceph-csi/blob/2407b245873b23391ec10d224cc979b98e5b33d5/docs/rbd/deploy.md#configuring-kmip-kms
so we can remove this doc I think.

black-dragon74: @Rakshith-R we do have vendor specific KMS design docs. Why not both keep this and add the reference to the RBD doc?

internal/kms/vault.go
```diff
@@ -122,6 +122,36 @@ func setConfigString(option *string, config map[string]interface{}, key string)
...
+//
+// If the value is not available, *option is not adjusted and
+// errConfigOptionMissing is returned.
+// In case the value is available, but can not be parsed to a boolean,
+// errConfigOptionInvalid is returned.
+func setConfigBoolean(option *bool, config map[string]any, key string) error {
+	value, ok := config[key]
+	if !ok {
+		return fmt.Errorf("%w: %s", errConfigOptionMissing, key)
+	}
```
Rakshith-R, line 136: we want to default to true in case it's missing ?

black-dragon74: This helper is implementation agnostic, like other helpers. 

We don’t do anything when `errConfigOptionMissing` is returned for optional values. We only act on `errConfigOptionInvalid` (like done for `USE_CRYPTO_RPC`)

Rakshith-R (dismissed):
Thanks !

internal/kms/vault.go
Rakshith-R: > This helper is implementation agnostic, like other helpers.
> 
> We don’t do anything when `errConfigOptionMissing` is returned for optional values. We only act on `errConfigOptionInvalid` (like done for `USE_CRYPTO_RPC`)

:+1: I misread it there.

docs/design/proposals/encryption-with-gklm.md
```diff
@@ -0,0 +1,158 @@
...
+  encrypted: "true"
+  encryptionKMSID: "gklm-test" # This matches the JSON key for GKLM config above
+  # ... other parameters
+```
+
+## Prerequisites
+
+Before using GKLM for volume encryption, you must:
+
+1. Create a RSA key pair (256-bit) in the GKLM server
```
iPraveenParihar: Can you correct this? I don't think we are using RSA here?

iPraveenParihar (commented):
Can you update PendingReleaseNotes too?

docs/design/proposals/encryption-with-gklm.md
black-dragon74: Updated to `symmetric` key pair.

black-dragon74:
@iPraveenParihar for lychee maybe we should do `--offline`, [ref](https://lychee.cli.rs/guides/cli/#--offline)? This will remove the flakiness that we currently have. Most sites block link crawling/have filters/captchas anyways.

The goal was to check for broken MD links, right?

Rakshith-R (dismissed):

Rakshith-R:
> @iPraveenParihar for lychee maybe we should do `--offline`, [ref](https://lychee.cli.rs/guides/cli/#--offline)? This will remove the flakiness that we currently have. Most sites block link crawling/have filters/captchas anyways.
> 
> The goal was to check for broken MD links, right?

I think so too, It failed a lot of times. please open new pr

docs/design/proposals/encryption-with-gklm.md
```diff
@@ -0,0 +1,158 @@
...
+## Prerequisites
+
+Before using GKLM for volume encryption, you must:
+
+1. Create a symmetric key pair (256-bit) in the GKLM server
+1. Obtain the UUID of this key
+1. Provide this UUID as the `UNIQUE_IDENTIFIER` in the Kubernetes Secret
+1. Ensure `USE_CRYPTO_RPC` is set to `"false"` in the KMS ConfigMap
+
+The key must be RSA-compatible, which is why Ceph-CSI uses a 256-bit
```
iPraveenParihar: missed here too?

black-dragon74: Nope. The key length must be RSA compatible.

iPraveenParihar: I didn't get the term RSA-compatible. The minimum RSA key length is of 2048-bit but it mentions Ceph-CSI use 256-bit for GKLM integration?

black-dragon74: My bad, What I meant to say was the key size should be AES (not RSA) block cipher compatible.

I removed it now, go's crypto will err out on invalid key sizes anyways.

iPraveenParihar (approved):

Rakshith-R (approved):

nixpanic:
Hold until #6143 is merged. Then `@mergifyio rebase`, wait and add `ok-to-test` to get it merged.

nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio refresh

nixpanic:
@mergifyio queue

nixpanic:
@mergifyio refresh

multi-arch-build has passed?

PR #6076, opened 2026-02-17
Add service account based volume access restriction

Rakshith-R:

## Service Account Based Volume Access Restriction

Ceph-CSI supports optionally restricting RBD/CephFS volume access to specific Kubernetes
service accounts. When configured, only pods running with the allowed service account
can mount the volume. This feature uses RBD image/cephFS subvolume metadata to store the
restriction and the CSI `podInfoOnMount` mechanism to identify the pod's service
account during mount. Refer https://kubernetes-csi.github.io/docs/pod-info.html#pod-info-on-mount-with-csi-driver-object.

### How it works

1. A user sets the `.rbd.csi.ceph.com/serviceaccount` metadata on an
   RBD image to specify the allowed service account name.
2. During `ControllerPublishVolume`, Ceph-CSI reads this metadata and passes
   it to the node via publish context.
3. During `NodePublishVolume`, Ceph-CSI compares the value against the
   pod's service account (provided via volume context by Kubelet).
4. If the service account was set in metadata and does not match the pod's
   service account, the mount is rejected with a `PermissionDenied` error.

### Prerequisites

The `podInfoOnMount` field must be set to `true` in the CSIDriver spec so that
Kubelet passes pod information (including service account name) in the volume
context during `NodePublishVolume`. Without this, the restriction cannot be
enforced and all mounts are allowed.

### Setting the restriction on an RBD image/nvmeof

Use the `rbd image-meta set` command to set the allowed service account:

```bash
rbd image-meta set <pool>/<image> .rbd.csi.ceph.com/serviceaccount <service-account-name>
```

For example, to restrict a volume to the `my-app-sa` service account:

```bash
rbd image-meta set mypool/csi-vol-abc123 .rbd.csi.ceph.com/serviceaccount my-app-sa
```

### Removing the restriction

To remove the restriction and allow any service account to mount the volume:

```bash
rbd image-meta remove <pool>/<image> .rbd.csi.ceph.com/serviceaccount
```

The Deployment should be scaled down completely and then scaled up for this removing
the restriction after removing metadata from the image.

### Setting the restriction on a CephFS subvolume/nfs export

Use the `ceph fs subvolume metadata set` command to set the allowed service account:

```bash
ceph fs subvolume metadata set <filesystem> <subvolume> --group_name=<group> \
  .cephfs.csi.ceph.com/serviceaccount <service-account-name>
```

For example, to restrict a volume to the `my-app-sa` service account:

```bash
ceph fs subvolume metadata set myfs csi-vol-abc123 --group_name=csi \
  .cephfs.csi.ceph.com/serviceaccount my-app-sa
```

### Removing the restriction

To remove the restriction and allow any service account to mount the volume:

```bash
ceph fs subvolume metadata rm <filesystem> <subvolume> --group_name=<group> \
  .cephfs.csi.ceph.com/serviceaccount
```

The Deployment should be scaled down completely and then scaled up for
removing the restriction after removing metadata from the subvolume.

---

### Use Case: Ceph VolSync Plugin Replication Destination PVC Protection

A primary motivator for this feature is the custom
[Ceph VolSync Plugin](https://github.com/RamenDR/ceph-volsync-plugin) that
performs incremental data replication across clusters. In a disaster recovery
or migration workflow:

1. A `ReplicationDestination` controller creates a PVC on the destination
   cluster to receive replicated data.
1. A replication worker pod, running under a dedicated service account (e.g.
   `volsync-worker-sa`), incrementally syncs data from the source cluster into
   this destination PVC.
1. The destination PVC must remain writable only by the replication worker
   until the replication is complete and a failover is triggered.

Without service account based restriction, any pod in the namespace with a
reference to the destination PVC could write to it, potentially corrupting the
replicated data or breaking the incremental sync state. By binding the
destination volume to the replication worker's service account, the volume is
protected from unintended writes throughout the replication lifecycle. On
failover, the restriction is removed so the application workload can mount
the volume.

---

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -2,9 +2,24 @@
 
 ## Breaking changes
 
+- NFS CSIDriver object's `spec.attachRequired` is now set to `true`
+  to accommodate Kubernetes ServiceAccount based volume access restriction
+  feature. This is a breaking change for users who have already deployed the
+  NFS driver and are upgrading to v3.17. Users will need to delete and recreate
+  the CSIDriver object for NFS during upgradie to v3.17.
+- NVME-OF Storageclass now needs publish secrets to accommodate the
+  Kubernetes ServiceAccount based volume access restriction.
+  This is a breaking change for users who have already deployed the NVME-OF driver
+  and are upgrading to v3.17. Users will need to recreate the NVME-OF Storageclass
+  with publish secrets during upgrade to v3.17.
+
 ## Features
 
 - nfs: allow changing NFS-server through ControllerModifyVolume [PR](https://github.com/ceph/ceph-csi/pull/5829)
 - util: add support for GKLM KMS over KMIP [PR](https://github.com/ceph/ceph-csi/pull/6048)
+- rbd: add Kubernetes ServiceAccount based volume access restriction
+- nvmeof: add Kubernetes ServiceAccount based volume access restriction
+- cephfs: add Kubernetes ServiceAccount based volume access restriction
+- nfs: add Kubernetes ServiceAccount based volume access restriction
 
 ## NOTE
--- a/deploy/nfs/kubernetes/csi-nfsplugin-provisioner.yaml
+++ b/deploy/nfs/kubernetes/csi-nfsplugin-provisioner.yaml
@@ -97,6 +97,29 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+        - name: csi-attacher
+          image: registry.k8s.io/sig-storage/csi-attacher:v4.11.0
+          args:
+            - "--v=1"
+            - "--csi-address=$(ADDRESS)"
+            - "--leader-election=true"
+            - "--retry-interval-start=500ms"
+            - "--http-endpoint=$(POD_IP):8093"
+          env:
+            - name: ADDRESS
+              value: /csi/csi-provisioner.sock
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+          imagePullPolicy: "IfNotPresent"
+          ports:
+            - containerPort: 8093
+              name: attacher
+              protocol: TCP
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
         - name: csi-resizer
           image: gcr.io/k8s-staging-sig-storage/csi-resizer:canary
           args:
--- a/deploy/nfs/kubernetes/csidriver.yaml
+++ b/deploy/nfs/kubernetes/csidriver.yaml
@@ -11,7 +11,8 @@ kind: CSIDriver
 metadata:
   name: "nfs.csi.ceph.com"
 spec:
-  attachRequired: false
+  attachRequired: true
+  podInfoOnMount: true
   fsGroupPolicy: File
   seLinuxMount: true
   volumeLifecycleModes:
--- a/docs/cephfs/deploy.md
+++ b/docs/cephfs/deploy.md
@@ -266,3 +266,92 @@ plain password (Kubernetes Secrets) work.
 Requires subvolumegroup to be created before provisioning the PVC.
 If the subvolumegroup provided in `ceph-csi-config` ConfigMap is missing
 in the ceph cluster, the PVC creation will fail and will stay in `Pending` state.
+
+## Kubernetes ServiceAccount Based Volume Access
+
+Ceph-CSI supports optionally restricting CephFS subvolume
+access to specific Kubernetes ServiceAccounts. When
+configured, only Pods running with one of the allowed
+ServiceAccounts can mount the subvolume. One or more
+ServiceAccounts can be specified as a comma-separated
+list. This feature uses CephFS subvolume metadata to
+store the restriction and the CSI
+[`podInfoOnMount`][pod-info-on-mount] mechanism to
+identify the Pod's ServiceAccount during mount.
+
+[pod-info-on-mount]:
+<https://kubernetes-csi.github.io/docs/pod-info.html#pod-info-on-mount-with-csi-driver-object>
+
+### How it works
+
+1. A storage admin sets the
+   `.cephfs.csi.ceph.com/serviceaccount` metadata on a
+   CephFS subvolume to specify the allowed
+   ServiceAccount name(s) as a comma-separated list.
+1. During `ControllerPublishVolume`, Ceph-CSI reads this metadata and passes
+   it to the node via publish context.
+1. During `NodePublishVolume`, Ceph-CSI splits the
+   comma-separated value and checks whether the Pod's
+   ServiceAccount (provided via volume context by
+   Kubelet) matches any entry.
+1. If the ServiceAccount was set in metadata and does
+   not match any of the allowed ServiceAccounts, the
+   mount is rejected with a `PermissionDenied` error.
+
+### Prerequisites
+
+The [`podInfoOnMount`][pod-info-on-mount] field must be
+set to `true` in the CSIDriver spec so that Kubelet
+passes Pod information (including ServiceAccount name)
+in the volume context during `NodePublishVolume`.
+Without this, the restriction cannot be enforced and
+all mounts are allowed.
+
+This feature requires controller-publish-secret set in storageclass
+for newer PVCs. For existing PVCs, the workaround mentioned
+[here](../design/proposals/non-graceful-node-shutdown.md#workaround-for-older-pvs)
+can be used.
+
+### Setting the restriction on a CephFS subvolume
+
+Use the `ceph fs subvolume metadata set` command to
+set the allowed ServiceAccount(s). Multiple
+ServiceAccounts can be specified as a
+comma-separated list:
+
+```bash
+ceph fs subvolume metadata set \
+  <filesystem> <subvolume> --group_name=<group> \
+  .cephfs.csi.ceph.com/serviceaccount \
+  <service-account-name>[,<service-account-name>...]
+```
+
+For example, to restrict a subvolume to the
+`my-app-sa` ServiceAccount:
+
+```bash
+ceph fs subvolume metadata set myfs \
+  csi-vol-abc123 --group_name=csi \
+  .cephfs.csi.ceph.com/serviceaccount my-app-sa
+```
+
+To allow multiple ServiceAccounts:
+
+```bash
+ceph fs subvolume metadata set myfs \
+  csi-vol-abc123 --group_name=csi \
+  .cephfs.csi.ceph.com/serviceaccount \
+  my-app-sa,my-worker-sa
+```
+
+### Removing the restriction
+
+To remove the restriction and allow any ServiceAccount to mount the subvolume:
+
+```bash
+ceph fs subvolume metadata rm <filesystem> <subvolume> --group_name=<group> \
+  .cephfs.csi.ceph.com/serviceaccount
+```
+
+All the Pods using the PVC should be scaled down completely and then scaled up for
+removing the restriction after removing metadata from the subvolume.
--- a/docs/design/proposals/k8s-sa-based-volume-access-restriction.md
+++ b/docs/design/proposals/k8s-sa-based-volume-access-restriction.md
@@ -1,0 +1,167 @@
+# Kubernetes ServiceAccount Based Volume Access Restriction
+
+## Introduction
+
+This proposal introduces an optional mechanism to restrict volume access based
+on the Kubernetes ServiceAccount of the Pod mounting the volume. When
+configured, only Pods running with one of the specified
+ServiceAccounts are allowed to mount the volume. All
+other mount attempts are rejected with a
+`PermissionDenied` error.
+
+The restriction is stored as metadata on the backend Ceph object (RBD image
+metadata or CephFS subvolume metadata) and is enforced at mount time through
+the CSI [`podInfoOnMount`][pod-info-on-mount] mechanism.
+This design depends solely on Kubernetes providing the Pod's ServiceAccount
+name in the volume context during `NodePublishVolume` calls via the standard
+key `csi.storage.k8s.io/serviceAccount.name: {pod.Spec.ServiceAccountName}`.
+No other validation is performed on the ServiceAccount name.
+
+[pod-info-on-mount]:
+<https://kubernetes-csi.github.io/docs/pod-info.html#pod-info-on-mount-with-csi-driver-object>
+
+## Motivation
+
+Ceph-CSI volumes are accessible to any Pod that has a valid PVC reference and
+the necessary RBAC to use the StorageClass. In multi-tenant and data pipeline
+environments, this is insufficient. There are scenarios where a volume should
+be exclusively accessible to a specific workload identity even when other Pods
+in the same namespace can reference the PVC.
+
+### Use Case: Ceph VolSync Plugin Replication Destination PVC Protection
+
+A primary motivator for this feature is the custom
+[Ceph VolSync Plugin](https://github.com/RamenDR/ceph-volsync-plugin) that
+performs incremental data replication across clusters. In a disaster recovery
+or migration workflow:
+
+1. A `ReplicationDestination` controller creates a PVC on the destination
+   cluster to receive replicated data.
+1. A replication worker Pod, running under a dedicated ServiceAccount (e.g.
+   `volsync-worker-sa`), incrementally syncs data from the source cluster into
+   this destination PVC.
+1. The destination PVC must remain writable only by the replication worker
+   until the replication is complete and a failover is triggered.
+
+Without ServiceAccount based restriction, any Pod in the namespace with a
+reference to the destination PVC could write to it, potentially corrupting the
+replicated data or breaking the incremental sync state. By binding the
+destination volume to the replication worker's ServiceAccount, the volume is
+protected from unintended writes throughout the replication lifecycle. On
+failover, the restriction is removed so the application workload can mount
+the volume.
+
+### Other Potential Use Cases
+
+- **Sensitive data volumes**: Restrict access to volumes containing regulated
+  data to only the ServiceAccount authorized to process them.
+- **Custom usecases**: Similar usecases where a
+  workload identity needs exclusive access to a volume
+  for data integrity or security reasons.
+
+## Dependency
+
+- The [`podInfoOnMount`][pod-info-on-mount] field must
+  be set to `true` in the CSIDriver specification.
+  This causes Kubelet to inject Pod information
+  (including the ServiceAccount name) into the volume
+  context during `NodePublishVolume`. Without this,
+  the restriction cannot be enforced. Since this
+  parameter is a mutable field in the CSIDriver spec,
+  it will be enabled by default going
+  forward(cephcsi v3.17.0+).
+
+## Design
+
+### Metadata Keys
+
+Each driver type uses a driver-specific metadata key to
+store the allowed ServiceAccount name(s) as a
+comma-separated list (e.g. `sa1,sa2,sa3`):
+
+| Driver | Metadata Key | Storage |
+|--------|-------------|---------|
+| RBD | `.rbd.csi.ceph.com/serviceaccount` | RBD image metadata |
+| CephFS | `.cephfs.csi.ceph.com/serviceaccount` | CephFS subvolume metadata |
+| NVMe-oF | `.rbd.csi.ceph.com/serviceaccount` | RBD image metadata (via RBD backend) |
+| NFS | `.cephfs.csi.ceph.com/serviceaccount` | CephFS subvolume metadata (via CephFS backend) |
+
+One or more ServiceAccounts can be specified per volume,
+separated by commas.
+
+### CSI Flow
+
+The restriction is enforced across two CSI RPCs:
+
+1. **ControllerPublishVolume**: The controller reads the ServiceAccount
+   metadata from the Ceph backend. If present, it is included in the publish
+   context passed to the node.
+
+1. **NodePublishVolume**: The node plugin splits the
+   publish context value on commas and checks whether
+   the Pod's ServiceAccount (provided by Kubelet via
+   `csi.storage.k8s.io/serviceAccount.name` in the
+   volume context) matches any entry. A mismatch
+   results in a `PermissionDenied` error. If no
+   restriction is set, or if `podInfoOnMount` is not
+   enabled, the mount is allowed.
+
+### Implementation
+
+A shared validation function `ValidateServiceAccountRestriction` in
+`internal/util/validate.go` is called at the beginning of `NodePublishVolume`
+in all four drivers (RBD, CephFS, NFS, NVMe-oF), ensuring consistent
+enforcement.
+
+Each driver reads the restriction metadata in `ControllerPublishVolume` using
+its backend:
+
+- **RBD**: reads via `GetMetadata` in `internal/rbd/controllerserver.go`.
+- **CephFS**: reads via `ListMetadata` in
+  `internal/cephfs/controllerserver.go`.
+- **NVMe-oF**: delegates to the RBD backend and propagates the publish context
+  in `internal/nvmeof/controller/controllerserver.go`.
+- **NFS**: delegates to the CephFS backend in
+  `internal/nfs/controller/controllerserver.go`.
+
+## Setting and Removing the Restriction
+
+The restriction is managed through Ceph CLI commands. Refer to the
+"Kubernetes ServiceAccount Based Volume Access" sections in
+[RBD deploy.md](../../rbd/deploy.md) and
+[CephFS deploy.md](../../cephfs/deploy.md) for usage
+instructions and examples.
+
+## Ceph VolSync Plugin Integration Example
+
+1. The replication destination worker sets the
+   ServiceAccount restriction on the backing Ceph
+   object(RBD image or CephFS subvolume) to the
+   replication worker's ServiceAccount
+   (e.g.`volsync-worker-sa`) on first use.
+1. Only the worker Pod mounts the destination PVC successfully because its
+   ServiceAccount matches. Any other Pod attempting to mount the same PVC is
+   rejected with `PermissionDenied` during NodePublish call, protecting data
+   integrity during incremental sync.
+1. On replication destination deletion, the controller spins up a cleanup job
+   that removes the ServiceAccount restriction
+   metadata, allowing the application workload to
+   mount the volume.
+
+## Limitations
+
+- Enforced at CSI mount time only; does not prevent direct access to the
+  underlying Ceph storage from outside Kubernetes.
+- If `podInfoOnMount` is not enabled, the restriction is silently unenforced.
+- Changing the restriction on an already-mounted volume does not affect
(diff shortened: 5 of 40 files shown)


Rakshith-R:
rbd alone takes 
run e2e
 - 1h 36m 20s

https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.34-rbd/detail/mini-e2e_k8s-1.34-rbd/6/pipeline/90

cephfs takes

run e2e
 - 55m 27s
 
 https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.34-cephfs/detail/mini-e2e_k8s-1.34-cephfs/17/pipeline/90
 
 We need to increase e2e timeout for now, and then work on optimising e2e testcases.

Rakshith-R:
> /test ci/centos/mini-e2e/k8s-1.34

nvmeof testcase passed here: #6097 

https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/rest/organizations/jenkins/pipelines/mini-e2e_k8s-1.35-nvmeof/runs/22/nodes/90/log/?start=0

Madhu-1:
@Rakshith-R, what is the usecase this one is trying to cover? its good to have a design first to see how feasible this solution is from a storage admin point of view?

Rakshith-R:
> @Rakshith-R, what is the usecase this one is trying to cover? its good to have a design first to see how feasible this solution is from a storage admin point of view?

Added design proposal document. 

---

### Use Case: Ceph VolSync Plugin Replication Destination PVC Protection

A primary motivator for this feature is the custom
[Ceph VolSync Plugin](https://github.com/RamenDR/ceph-volsync-plugin) that
performs incremental data replication across clusters. In a disaster recovery
or migration workflow:

1. A `ReplicationDestination` controller creates a PVC on the destination
   cluster to receive replicated data.
1. A replication worker pod, running under a dedicated service account (e.g.
   `volsync-worker-sa`), incrementally syncs data from the source cluster into
   this destination PVC.
1. The destination PVC must remain writable only by the replication worker
   until the replication is complete and a failover is triggered.

Without service account based restriction, any pod in the namespace with a
reference to the destination PVC could write to it, potentially corrupting the
replicated data or breaking the incremental sync state. By binding the
destination volume to the replication worker's service account, the volume is
protected from unintended writes throughout the replication lifecycle. On
failover, the restriction is removed so the application workload can mount
the volume.

---

docs/design/proposals/sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,146 @@
+# Service Account Based Volume Access Restriction
```
nixpanic: Kubernetes ServiceAccount Based Volume Access Restriction

Mention Kubernetes explicitly, and spell ServiceAccount as a Kubernetes type.

docs/design/proposals/k8s-sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,146 @@
+# Service Account Based Volume Access Restriction
+
+## Introduction
+
+This proposal introduces an optional mechanism to restrict volume access based
```
nixpanic, line 5: volume access, based

Makes it read a little easier

docs/design/proposals/sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Introduction
+
+This proposal introduces an optional mechanism to restrict volume access based
+on the Kubernetes service account of the pod mounting the volume. When
```
nixpanic: Always **ServiceAccount** please. Ideally **Pod** with capital as well.

docs/design/proposals/sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,146 @@
...
+configured, only pods running with the specified service account are allowed to
+mount the volume. All other mount attempts are rejected with a
+`PermissionDenied` error.
+
+The restriction is stored as metadata on the backend Ceph object (RBD image
+metadata or CephFS subvolume metadata) and is enforced at mount time through
+the CSI `podInfoOnMount` mechanism.
```
nixpanic: add link to to podInfoOnMount docs?

docs/design/proposals/sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,146 @@
...
+  Without this, the restriction cannot be enforced.
+  Since this parameter is a mutable field in the CSIDriver spec, it will be enabled
+  by default going forward(cephcsi v3.17.0+).
+
+## Design
+
+### Metadata Keys
+
+Each driver type uses a driver-specific metadata key to store the allowed
+service account name:
```
nixpanic: what is the format of the value? Does it include the Kubernetes Namespace of the ServiceAccount? If not, would the Namespace add extra security?

docs/design/proposals/k8s-sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,146 @@
...
+
+### CSI Flow
+
+The restriction is enforced across two CSI RPCs:
+
+1. **ControllerPublishVolume**: The controller reads the service account
+   metadata from the Ceph backend. If present, it is included in the publish
+   context passed to the node.
+
+1. **NodePublishVolume**: The node plugin compares the publish context value
```
nixpanic: This isn't a very safe way. Both the value that is expected and the value that is checked are in the same CSI RPC call? It isn't really how security checks for access restrictions should work.

Why not read the metadata during NodePublishVolume?

docs/design/proposals/sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,146 @@
...
+- **CephFS**: reads via `ListMetadata` in `internal/cephfs/controllerserver.go`.
+- **NVMe-oF**: delegates to the RBD backend and propagates the publish context
+  in `internal/nvmeof/controller/controllerserver.go`.
+- **NFS**: delegates to the CephFS backend in
+  `internal/nfs/controller/controllerserver.go`.
+
+## Setting and Removing the Restriction
+
+The restriction is managed through Ceph CLI commands. Refer to the
+"Service Account Based Volume Access Restriction" sections in
```
nixpanic: update the title of those sections too

docs/design/proposals/sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,146 @@
...
+   account matches. Any other pod attempting to mount the same PVC is rejected with
+   `PermissionDenied` during NodePublish call, protecting data integrity during
+   incremental sync.
+1. On replication destination deletion, the controller spins up a cleanup job
+   that removes the service account restriction metadata, allowing the application
+   workload to mount the volume.
+
+## Limitations
+
+- Only a single service account can be specified per volume.
```
nixpanic: Is there an opt-in or opt-out per volume? Should this only apply for PVCs with RWOP access modes?

For webservers it may be relatively common to have a RWX volume where most Pods read from it (webserver processes) and only a few update the data. These would likely run with different ServiceAccounts. Is there a way to have this work, while other volumes have the access restriction applied?

nixpanic (commented):
Went through the design as a 1st step. Looks okayish to me, but I do have a few questions.

docs/design/proposals/sa-based-volume-access-restriction.md
Rakshith-R: > what is the format of the value? Does it include the Kubernetes Namespace of the ServiceAccount? If not, would the Namespace add extra security?

It is just the serviceAccount name. 
The pvc and pod are already namespaced and hence NodePublish request is also namespaced.
If a NodePublish request for same volume and with same serviceAccount has to come from another NS, 
A user will need to have clusterScoped permission to create a duplicate PV and PVC as well as a a serviceAccount with same name in another namespace, which is very unlikely and another security threat outside the scope here.

But I've mentioned additional parameters (e.g. namespace) validation in future enhancement section.

docs/design/proposals/sa-based-volume-access-restriction.md
Rakshith-R: > Is there an opt-in or opt-out per volume? Should this only apply for PVCs with RWOP access modes?

It's opt in per volume + podInfoOnMount needs to be enabled.
User has to explicitly set metadata on image/subvolume.

It's applicable on all access modes. Multiple pods with same serviceAccount can mount a RWX PVC.

> For webservers it may be relatively common to have a RWX volume where most Pods read from it (webserver processes) and only a few update the data. These would likely run with different ServiceAccounts. Is there a way to have this work, while other volumes have the access restriction applied?

Currently no as mentioned in limitations.
Maybe in the future if there's requests/requirements for such a usecase, we can extend the framework using something like serviceAccountList or similar.

docs/design/proposals/k8s-sa-based-volume-access-restriction.md
Rakshith-R: > This isn't a very safe way. Both the value that is expected and the value that is checked are in the same CSI RPC call? It isn't really how security checks for access restrictions should work.
> 
> Why not read the metadata during NodePublishVolume?

Yes, the value that is expected and the value that is checked are in the same CSI RPC call
but the source of these values are from two different places.
- expected <- publish_context <- controllerPublishVolume call <- image/subvolume metadata
- actual checked <- volume_context <- pod's SA name

The spec and k8s storage framework ensures the values are populated in such manner.

Reading it from metadata on every NodePublishVolume(per pod, re-requested with backoff on errors) is resource intensive and can be totally avoided by making use of publish_context in controllerPublishVolume call (once per pvc per node).

nixpanic: I understand that the flow of the Kubernetes CSI procedures makes this work. But it feels wrong to get the valid value from the publish context which is part of the same RPC as the details from podInfoOnMount.

NodePublishVolume is not in a critical path, doing a security check by getting the metadata is not very invasive. It would be more proper to validate it during the call.

docs/cephfs/deploy.md
```diff
@@ -266,3 +266,65 @@ plain password (Kubernetes Secrets) work.
...
+CephFS subvolume metadata to store the restriction and
+the CSI [`podInfoOnMount`][pod-info-on-mount] mechanism
+to identify the Pod's ServiceAccount during mount.
+
+[pod-info-on-mount]:
+<https://kubernetes-csi.github.io/docs/pod-info.html#pod-info-on-mount-with-csi-driver-object>
+
+### How it works
+
+1. A user sets the `.cephfs.csi.ceph.com/serviceaccount` metadata on a
```
nixpanic: How can users set this? Normal users do not have access to `ceph` commands to do this.

docs/rbd/deploy.md
```diff
@@ -598,3 +598,62 @@ Additional Resources:
...
+RBD image metadata to store the restriction and the CSI
+[`podInfoOnMount`][pod-info-on-mount] mechanism to identify
+the Pod's ServiceAccount during mount.
+
+[pod-info-on-mount]:
+<https://kubernetes-csi.github.io/docs/pod-info.html#pod-info-on-mount-with-csi-driver-object>
+
+### How it works
+
+1. A user sets the `.rbd.csi.ceph.com/serviceaccount` metadata on an
```
nixpanic: I fail to see how a user can set this. Users do not have direct access to the Ceph cluster to modify RBD metadata (and we really should not allow users to do that).

docs/design/proposals/k8s-sa-based-volume-access-restriction.md
Rakshith-R: > I understand that the flow of the Kubernetes CSI procedures makes this work. But it feels wrong to get the valid value from the publish context which is part of the same RPC as the details from podInfoOnMount.

>the source of these values are from two different places.
>expected <- publish_context <- controllerPublishVolume call <- image/subvolume metadata
>actual checked <- volume_context <- pod's SA name

Kubelet and CSI Spec are trusted entities in this framework.
The entire framework relies on this fact.
Therefore, the content of the RPC is very reliable.

> NodePublishVolume is not in a critical path, doing a security check by getting the metadata is not very invasive. It would be more proper to validate it during the call.

It is a critical path, every pod using our volume will be able to come up only after a NodePublish.
CephCSI's nodepublish ops are very light  weight and fast, almost zero lookups.

Introducing a metadata lookup here would inherently slow the entire setup.

Combining with the above established fact of Kubelet and CSI Spec being reliable and trustworthy entities, we don't need to look at introducing resource intensive ops in NodePublish call.

docs/cephfs/deploy.md
Rakshith-R: yes, made it storage admin now.

docs/rbd/deploy.md
Rakshith-R: yes, made it storage admin now.

deploy/nfs/kubernetes/csi-nfsplugin-provisioner.yaml
```diff
@@ -98,6 +98,29 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+        - name: csi-attacher
+          image: registry.k8s.io/sig-storage/csi-attacher:v4.10.0
```
iPraveenParihar: 4.11

internal/rbd/nodeserver.go
```diff
@@ -857,6 +857,11 @@ func (ns *NodeServer) NodePublishVolume(
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
 
+	// Validate that the pod's service account is allowed to mount this volume
+	if err = util.ValidateServiceAccountRestriction(ctx, req); err != nil {
```
iPraveenParihar: This could be moved up, before checking for mount path?

iPraveenParihar (commented):
Minor comment, but overall everything looks good.

deploy/nfs/kubernetes/csidriver.yaml
```diff
@@ -11,7 +11,8 @@ kind: CSIDriver
 metadata:
   name: "nfs.csi.ceph.com"
 spec:
-  attachRequired: false
+  attachRequired: true
```
Madhu-1, line 14: This will cause an upgrade problem. its immutable

docs/cephfs/deploy.md
```diff
@@ -266,3 +266,65 @@ plain password (Kubernetes Secrets) work.
 Requires subvolumegroup to be created before provisioning the PVC.
 If the subvolumegroup provided in `ceph-csi-config` ConfigMap is missing
 in the ceph cluster, the PVC creation will fail and will stay in `Pending` state.
+
+## Kubernetes ServiceAccount Based Volume Access
+
+Ceph-CSI supports optionally restricting CephFS volume
```
Madhu-1: `CephFS volume` to `CephFS subvolume`

docs/cephfs/deploy.md
```diff
@@ -266,3 +266,65 @@ plain password (Kubernetes Secrets) work.
...
+access to specific Kubernetes ServiceAccounts. When
+configured, only Pods running with the allowed
+ServiceAccount can mount the volume. This feature uses
```
Madhu-1: subvolume

docs/cephfs/deploy.md
```diff
@@ -266,3 +266,65 @@ plain password (Kubernetes Secrets) work.
...
+CephFS subvolume metadata to store the restriction and
+the CSI [`podInfoOnMount`][pod-info-on-mount] mechanism
+to identify the Pod's ServiceAccount during mount.
+
+[pod-info-on-mount]:
+<https://kubernetes-csi.github.io/docs/pod-info.html#pod-info-on-mount-with-csi-driver-object>
+
+### How it works
+
+1. A storage admin sets the `.cephfs.csi.ceph.com/serviceaccount` metadata on a
```
Madhu-1: it would be better to allow multiple serviceaccounts not just one.

docs/cephfs/deploy.md
```diff
@@ -266,3 +266,65 @@ plain password (Kubernetes Secrets) work.
...
+
+Use the `ceph fs subvolume metadata set` command to
+set the allowed ServiceAccount:
+
+```bash
+ceph fs subvolume metadata set <filesystem> <subvolume> --group_name=<group> \
+  .cephfs.csi.ceph.com/serviceaccount <service-account-name>
+```
+
+For example, to restrict a volume to the `my-app-sa` ServiceAccount:
```
Madhu-1: subvolume

docs/cephfs/deploy.md
```diff
@@ -266,3 +266,65 @@ plain password (Kubernetes Secrets) work.
...
+
+```bash
+ceph fs subvolume metadata set myfs csi-vol-abc123 --group_name=csi \
+  .cephfs.csi.ceph.com/serviceaccount my-app-sa
+```
+
+### Removing the restriction
+
+To remove the restriction and allow any ServiceAccount to mount the volume:
```
Madhu-1: subvolume

docs/design/proposals/k8s-sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,158 @@
+# Kubernetes ServiceAccount Based Volume Access Restriction
+
+## Introduction
+
+This proposal introduces an optional mechanism to restrict volume access based
+on the Kubernetes ServiceAccount of the Pod mounting the volume. When
```
Madhu-1, line 6: its good to design considering multiple SA in same namespace as its possible to have this combination

docs/design/proposals/k8s-sa-based-volume-access-restriction.md
```diff
@@ -0,0 +1,158 @@
...
+- If `podInfoOnMount` is not enabled, the restriction is silently unenforced.
+- Changing the restriction on an already-mounted volume does not affect
+  existing mounts. The volume must be unmounted and remounted.
+- Managed through Ceph CLI commands, not Kubernetes-native APIs.
+
+## Future Enhancements
+
+- Support restriction based on other attributes (e.g. name, namespace) in
+  addition to ServiceAccount.
+- Add support for multiple allowed ServiceAccounts per volume.
```
Madhu-1: i see this is for future

examples/nvmeof/storageclass.yaml
```diff
@@ -14,6 +14,8 @@ parameters:
   csi.storage.k8s.io/controller-expand-secret-namespace: default
   csi.storage.k8s.io/node-stage-secret-name: csi-nvmeof-secret
   csi.storage.k8s.io/node-stage-secret-namespace: default
+  csi.storage.k8s.io/controller-publish-secret-name: csi-nvmeof-secret
```
Madhu-1, line 17: did we document that we dont support this feature for already created pvc where we dont have controller publish secret or if they want support for existing pvc, the secret need to be added to the configmap?

internal/rbd/controllerserver.go
```diff
@@ -1688,15 +1688,84 @@ func (cs *ControllerServer) ControllerExpandVolume(
...
+	}
+
+	cr, err := util.NewUserCredentials(secrets)
+	if err != nil {
+		return "", status.Errorf(codes.InvalidArgument,
+			"failed to create credentials for volume %s: %v", volumeID, err)
+	}
+	defer cr.DeleteCredentials()
+
+	rv, err := GenVolFromVolID(ctx, volumeID, cr, secrets)
```
Madhu-1, line 1749: Are we planning to support static pvc as well?

internal/util/validate.go
```diff
@@ -195,3 +209,38 @@ func IsStaticVol(volAttrs map[string]string) bool {
...
+			"volume %s has service account restriction %q but podInfoOnMount is not enabled, skipping check",
+			req.GetVolumeId(), allowedSA)
+
+		return nil
+	}
+
+	if podServiceAccount != allowedSA {
+		return status.Errorf(codes.PermissionDenied,
+			"volume %s is restricted to service account %q, but pod uses service account %q",
+			req.GetVolumeId(), allowedSA, podServiceAccount)
```
Madhu-1: we should not show this details to the user, there are chances that user can create SA with this name and try to access the volume

internal/rbd/controllerserver.go
Rakshith-R: Adding to future enhancements

examples/nvmeof/storageclass.yaml
Rakshith-R: Adding a para in prerequisites.

deploy/nfs/kubernetes/csidriver.yaml
Rakshith-R: > This will cause an upgrade problem. its immutable

:facepalm:  forgot since podInfoOnMount was mutable.

added a point in breaking changes

docs/rbd/deploy.md
```diff
@@ -598,3 +598,67 @@ Additional Resources:
 
 * External Snapshot Metadata sidecar project:
   [repository](https://github.com/kubernetes-csi/external-snapshot-metadata)
+
+## Kubernetes ServiceAccount Based Volume Access
+
+Ceph-CSI supports optionally restricting RBD volume
+access to specific Kubernetes ServiceAccounts. When
+configured, only Pods running with the allowed
+ServiceAccount can mount the volume. This feature uses
```
nixpanic: This does not describe what validation of the ServiceAccount is done. Could you please add it? Explain why this isn't done: https://kubernetes.io/docs/concepts/security/service-accounts/#authenticating-in-code

docs/rbd/deploy.md
```diff
@@ -598,3 +598,67 @@ Additional Resources:
...
+
+### Removing the restriction
+
+To remove the restriction and allow any ServiceAccount to mount the volume:
+
+```bash
+rbd image-meta remove <pool>/<image> .rbd.csi.ceph.com/serviceaccount
+```
+
+The Deployment should be scaled down completely and
```
nixpanic: Which deployment (or daemonset, or ...)? I assume the application using the PVC, and not Ceph-CSI?

internal/util/validate.go
```diff
@@ -195,3 +209,38 @@ func IsStaticVol(volAttrs map[string]string) bool {
...
 	return false
 }
+
+// ValidateServiceAccountRestriction checks whether the pod's service account
+// is allowed to mount the volume. The allowed service account is read from the
+// publish context (set during ControllerPublishVolume). If no restriction is
+// set, all service accounts are allowed.
+func ValidateServiceAccountRestriction(
+	ctx context.Context,
+	req *csi.NodePublishVolumeRequest,
```
nixpanic: it would be cleaner to pass only the arguments that you need, not the complete request.

internal/util/validate.go
```diff
@@ -195,3 +209,38 @@ func IsStaticVol(volAttrs map[string]string) bool {
...
+		// podInfoOnMount is not enabled, cannot enforce restriction
+		log.WarningLog(ctx,
+			"volume %s has service account restriction but podInfoOnMount is not enabled, skipping check",
+			req.GetVolumeId())
+
+		return nil
+	}
+
+	if podServiceAccount != allowedSA {
+		return status.Errorf(codes.PermissionDenied,
```
nixpanic: Helper functions outside the main controller/nodeplugin package should really not return CSI-formatted errors.

docs/cephfs/deploy.md
nixpanic: Agreed, different workloads on the same PVC are common. Think of 1st workload for injecting data, 2nd workload for servicing/processing, 3rd workload for backup/copy-out.

PendingReleaseNotes.md
```diff
@@ -2,8 +2,16 @@
 
 ## Breaking changes
 
+- nfs csi driver object has breaking changes to support
+  Kubernetes ServiceAccount based volume access restriction.
```
nixpanic, line 11: "NFS CSIDriver object", what breaks, and what needs to be done to resolve it?

PendingReleaseNotes.md
```diff
@@ -2,8 +2,16 @@
...
+- nvmeof storageclass now needs publish secrets.
```
nixpanic: Please wite "NVMe-oF StorageClass"

docs/design/proposals/k8s-sa-based-volume-access-restriction.md
nixpanic: That's all fine, but it needs documenting in the design too.

docs/rbd/deploy.md
Rakshith-R: Modified to `All Pods using the PVC`*

Rakshith-R:
Added support for multiple SA
+ e2e for it.
Added/modified design doc according to reviews.
Please take a look

iPraveenParihar (commented):
Thanks!

PendingReleaseNotes.md
```diff
@@ -2,9 +2,24 @@
 
 ## Breaking changes
 
+- NFS CSIDriver object's `spec.attachRequired` is now set to `true`
```
Madhu-1, line 5: @Rakshith-R we need to update the upgrade doc this change

Madhu-1 (commented):
LGTM just a comment on the upgrade doc to delete and recreate the CR

PendingReleaseNotes.md
Rakshith-R: Added a note.

Rakshith-R:
Resolved conflicts, 
PTAL

Madhu-1 (dismissed):

iPraveenParihar (dismissed):

iPraveenParihar:
@mergifyio rebase

iPraveenParihar:
@mergifyio queue

nixpanic:
**ci/centos/mini-e2e/k8s-1.34** failed:
```
  [FAILED] service account volume restriction test failed: failed to create PVC: failed to get pvc: client rate limiter Wait returned an error: context deadline exceeded
```

[logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.34/detail/mini-e2e_k8s-1.34/664/pipeline)

@Rakshith-R, please check

Rakshith-R:
> **ci/centos/mini-e2e/k8s-1.34** failed:
> 
> ```
>   [FAILED] service account volume restriction test failed: failed to create PVC: failed to get pvc: client rate limiter Wait returned an error: context deadline exceeded
> ```
> 
> [logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.34/detail/mini-e2e_k8s-1.34/664/pipeline)
> 
> @Rakshith-R, please check

The test case moved while resolving conflicts and at the spot storageclass was missing 
:facepalm:  It should run now.

Opened https://github.com/ceph/ceph-csi/issues/6177 to handle it.

iPraveenParihar (approved):

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

PR #6106, opened 2026-02-20
rbd: limit max number of key rotation operations per driver instance

black-dragon74:

This patch adds semaphores that limit the maximum number of active key rotation processes per driver.

The default limit is: 10

### DIFF
--- a/internal/util/semaphore.go
+++ b/internal/util/semaphore.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2026 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"context"
+	"fmt"
+)
+
+// Semaphore is a counting semaphore for limiting concurrent operations.
+type Semaphore struct {
+	sem chan struct{}
+}
+
+// NewSemaphore creates a new semaphore with the given capacity.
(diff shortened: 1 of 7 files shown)


nixpanic:
Why do this in the CSI-driver, and not in the CSI-Addons operator or sidecar?

black-dragon74:
> Why do this in the CSI-driver, and not in the CSI-Addons operator or sidecar?

The key rotation process is handled by the csi-driver, addons just serves as an initiator for it.

The implementation here makes the driver more resilient by preventing it from crashing in cases of misconfiguration. 

Why not enforce the limit in csi-addons? Some other storage vendor might have a completely different impl. of key rotation spec where limits on reconciler end might not makes sense (e.g. lazy rotation on next mount).

We can get rid of the configurable aspect of max active processes (but a deployment with enough resources will be bottle-necked in this case) or we can set it to some value that co-relates of number of CPU cores available.

Either way, a limiter on processes that could potentially lead to memory/cpu spikes and cause the driver to crash is good to have?

Madhu-1:
csi driver should be able to handle the request. What is the reason to introduce this change, and where exactly the memory and cpu spike is happening?

black-dragon74:
> csi driver should be able to handle the request. What is the reason to introduce this change, and where exactly the memory and cpu spike is happening?

CSI driver does handle it but then the memory limits crash the pod. The reason for this patch is to only allow certain number of operations per driver instance. 

dm-crypt uses causes memory and cpu spikes for a brief period when its operations are ongoing. One too many operations at the same time exponentiate the resource spikes.

Rakshith-R (commented):
Thanks for the pr, It will definitely help OOMKills.

just a small nit regarding methods that are not used right now.

internal/util/semaphore.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Acquire acquires the semaphore, blocking until capacity is available or context is canceled.
+func (s *Semaphore) Acquire(ctx context.Context) error {
+	select {
+	case s.sem <- struct{}{}:
+		return nil
+	case <-ctx.Done():
+		return fmt.Errorf("failed to acquire semaphore: %w", ctx.Err())
+	}
+}
```
Rakshith-R, line 55: Let's remove this and and Avaiable since we do not make use of it as of right now ?

PR #6137, opened 2026-02-26
util: optimize key rotation flow for cryptsetup

black-dragon74:

This patch adds the missing memory limits to cryptsetup `AddKey`. If this limit is missing `cryptsetup` will determine its own limits in order to satisfy the KDF, which can lead to serious resource contentions, IOW, more power the system has, more resources will be used (this is intentional).

`Open` and `RemoveKey` use the memory limits(set when format/adding) from the existing headers on the device and hence do not require manual limits.

A limit on active threads is also enforced to run KDF on a single thread, reducing CPU contention when a lot of processes are active at the same time.

Additionally, use the `--test-passphrase` with `open` for verification instead of calling `changeKey` with same passphrase to save on unnecessary derivations and disk IO (test-passphrase is read only).

Testing this on ppc64le env where we see serious resource contentions and OOMKills. This, much simpler patch, will make #6106 redundant.

### DIFF
--- a/internal/util/cryptsetup/cryptsetup.go
+++ b/internal/util/cryptsetup/cryptsetup.go
@@ -39,6 +39,10 @@ const (
 	// Maximum time to wait for cryptsetup commands to complete.
 	ExecutionTimeout = 2*time.Minute + 30*time.Second
 
+	// Limit PBKDF parallel threads to 1 to reduce CPU contention
+	// when many cryptsetup processes run concurrently.
+	cryptsetupPBKDFParallelLimit = 1
+
 	// Limit memory used by Argon2i PBKDF to 32 MiB.
 	cryptsetupPBKDFMemoryLimit = 32 << 10 // 32768 KiB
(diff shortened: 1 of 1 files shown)


nixpanic:
> Testing this on ppc64le env where we see serious resource contentions and OOMKills.

Is there an issue with details about this problem? It would be good for tracking and inspiration on how to check resource usage for these kind of things.

Test-results with the resource usage difference between before/after this change would be great to see!

black-dragon74:
> Is there an issue with details about this problem? It would be good for tracking and inspiration on how to check resource usage for these kind of things.

We do have a reported [bug](https://redhat.atlassian.net/browse/DFBUGS-5777).

Rakshith-R (approved):

nixpanic (approved):

black-dragon74:
@mergifyio queue

black-dragon74:
@mergifyio rebase

PR #6159, opened 2026-03-05
rbd: allow disable replication

Madhu-1:
When the volume is not primary on both clusters, allow the volume to have mirror disabled.

Today, we are not supporting enabling and disabling the mirroring on a single cluster until the mirroring is enabled on the
remote clusters. This is not always true, as we need to support disabling DR sometimes.

This issue was caught when i was testing the volume replication in csi-addons

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -372,7 +372,7 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-	log.UsefulLog(ctx, "mirror state is %s", info.GetState())
+	log.UsefulLog(ctx, "mirror state is %q, is primary %v, force %v", info.GetState(), info.IsPrimary(), force)
 	switch info.GetState() {
 	// image is already in disabled state
 	case librbd.MirrorImageDisabled.String():
--- a/internal/rbd/replication.go
+++ b/internal/rbd/replication.go
@@ -46,39 +46,59 @@ func (rv *rbdVolume) RepairResyncedImageID(ctx context.Context, ready bool) erro
 	return rv.repairImageID(ctx, j, true)
 }
 
+// checkSecondaryForDisable validates whether mirroring can be disabled on a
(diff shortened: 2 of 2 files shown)


Nikhil-Ladha:
Does rbd allow disable of image when both are secondary?
As in there won't be any stale mirroring state of the image if we try to enable mirroring on the group again, right?

Also, are the images cleaned up on both the clusters after disabling?

Madhu-1:
> Does rbd allow disable of image when both are secondary? As in there won't be any stale mirroring state of the image if we try to enable mirroring on the group again, right?

I think it doesnt.

> Also, are the images cleaned up on both the clusters after disabling?
i dont think so.


Think about the case
-> Created VR on C1 cluster
     -> EnableReplication
     -> PromoteVolume
-> Change the state to secondary in the C1
      -> DemoteVolume
-> Deleting VR on C1 cluster
    -> DisableReplication

This could be a false case, https://github.com/csi-addons/kubernetes-csi-addons/actions/runs/22713665786?pr=984 tests are running here.

Nikhil-Ladha:
> > Does rbd allow disable of image when both are secondary? As in there won't be any stale mirroring state of the image if we try to enable mirroring on the group again, right?
> 
> I think it doesnt.
> 
> > Also, are the images cleaned up on both the clusters after disabling?
> > i dont think so.
> 
> Think about the case -> Created VR on C1 cluster -> EnableReplication -> PromoteVolume -> Change the state to secondary in the C1 -> DemoteVolume -> Deleting VR on C1 cluster -> DisableReplication
> 
> This could be a false case, https://github.com/csi-addons/kubernetes-csi-addons/actions/runs/22713665786?pr=984 tests are running here.

Ok, so it is more like allowing testing of negative test cases for VR?
With this change, VR would get deleted but we need to see that the respective PVC/image is cleaned up too else further tests might be affected by the stale data.

Madhu-1:
> Ok, so it is more like allowing testing of negative test cases for VR?
> With this change, VR would get deleted but we need to see that the respective PVC/image is cleaned up too else further tests might be affected by the stale data.

yes from csi-addons we dont care, the csi driver needs to block such operations.

Madhu-1:
Closing for now, will reopen if required

PR #6163, opened 2026-03-08
nvmeof: Implement safe controller disconnect logic

gadididi:
# Describe what this PR does #

### Summary
This PR adds proper nvme-of controller disconnect logic to the CSI driver. When a pod is deleted and its volume is unstaged, we now safely disconnect from nvme-of **controllers** only when no other volumes are using them.

### The Original Problem
Before this change, when `NodeUnstageVolume` was called, we would unmount the volume but leave the NVMe-oF connection active. This meant NVMe controllers stayed connected even after all pods were deleted. In next Pod creation (in same nvmeof subsystem) in some environments (like in the e2e test here: https://github.com/ceph/ceph-csi/pull/6058 )  the orphaned connection was stuck, with no reason. **The cleanest way to handle it is to disconnect from the controller\s (host-subsystem pair\s) if this pod is the last mounted**  

### Why We Can't Just Always Disconnect
In NVMe-oF multipath setups, multiple volumes share the same controllers. For example:
```
Subsystem: nqn.2016-06.io.spdk:cnode1
├── Namespace 1 (nvme0n1) - Used by PVC-1 and the device mounted for POD-1
└── Namespace 2 (nvme0n2) - Used by PVC-2 and the device mounted for POD-2
```

Both namespaces are accessed via:
```
├── Controller nvme0 (gateway 10.0.0.1:4420)
└── Controller nvme1 (gateway 10.0.0.2:4420)
```
If we disconnect controllers when POD-1 is deleted, we break POD-2 mounting which is still using those same controllers!!! 

### The Challenge
We need to track which volumes are still in use and only disconnect controllers when:

The volume being unstaged is no longer mounted
No other volumes are using the same controllers

### Solution Overview
This PR implements a reference-counting approach 
(on demand, without metadata file. just using the current state):

- When `NodeUnstageVolume` is called, we first unmount the volume
- We then check if this was the last mounted namespace (=nvme device) on each controller
- Only if a controller has no other mounted namespaces, we disconnect it

FYI- the connection operation in `NodeStageVolume()` connects to entire listeners (it means every listener connection ==> creates a controller == that is host-listener of specific subsystem connection) . 
So actually, it is not possible that some nvmeof namespace (**in same subsystem**) 
will have different controllers. It means all nvme namespaces (**in same subsystem**) are sharing the same controllers!! 

### Implementation Details
New Functions:

`getDeviceFromStagingPath()`

- Finds the NVMe device path for a given staging path
- Uses findmnt with JSON output for reliable parsing
[...]

PR #6183, opened 2026-03-16
util: add GroupLock for two-group mutual exclusion

gadididi:
Add GroupLock type that allows operations within the same group to run concurrently while blocking operations from the other group. This wiil be in used to coordinate NVMe-oF Stage and Unstage operations - multiple Stage calls can run in parallel,
multiple Unstage calls can run in parallel, but Stage and Unstage block each other.

**This can be in used in any place which needs lock between 2 types of group.**

The implementation is based on Keane and Moir's "group mutual exclusion algorithm", but simplified for exactly two groups using Golang condition variables.
Operations in group A wait if any group B operations are active, and vice versa.
When the last operation in a group completes, all waiting operations in the other group are woken up together.

Also added some test, simulate parallel workers:
- Multiple same-group operations run concurrently
- Different groups block each other
- No deadlock under heavy contention
- Mutual exclusion is never violated
- All waiting operations start together when unblocked

### Related PR
https://github.com/ceph/ceph-csi/pull/6163 - once this phase1 PR will be merged, the next phase is adding group lock for connect\ disconnect OP. 

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/internal/util/lock/group_lock.go
+++ b/internal/util/lock/group_lock.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2026 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package lock
+
+import "sync"
+
+// GroupLock implements a simplified version of Group Mutual Exclusion
+// for exactly 2 groups. For the general N-group case see:
+//
+// - Keane, P. and Moir, M. - "A simple local-spin group mutual exclusion algorithm"
+// - Link: https://dl.acm.org/doi/epdf/10.1145/301308.301319
+//
+// GroupLock implements mutual exclusion between two groups of operations.
+// Operations within the same group can run concurrently, but operations
+// from different groups cannot run simultaneously.
+//
+// Example use cases:
+// - Allowing multiple Stage operations OR multiple Unstage operations, but not both
+// - Allowing multiple Read operations OR multiple Write operations, but not both
+// - Allowing multiple Connect operations OR multiple Disconnect operations, but not both
+//
+// This is sometimes called "symmetric reader-writer lock" or "group mutual exclusion".
+//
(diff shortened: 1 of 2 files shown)


nixpanic (commented):
Sounds like what we can use.

I wonder if this is over-kill for the rare cases this would be an issue? Is there an alternative that would be simpler?

In theory this could cause blocking for long time. If one group (say NodeStageVolume) gets called often, in quick succession, a NodeUnstageVolume could be blocked until all NodeStageVolume calls are handled. I don't think that will be problematic for our use now, but it may be something to document so other users are warned?

gadididi:
@nixpanic , Hi!

> Is there an alternative that would be simpler?

I searched for `Group Mutual Exclusion` util in Go, but I did not find.. 

> In theory this could cause blocking for long time

right. I can make it more fair with kind of queue (like in the origin algorithm here: https://dl.acm.org/doi/10.1145/301308.301319) . 
In this way for sure we will not have any starvation. the big Q is it really common hundreds of pods can be created together? 
I can document it. 
I will figure out how difficult\over-kill it will be to add the queue.

### UPDATE ABOUT THE QUEUE: 

adding fairness (like a FIFO queue) would make the code significantly more complex - we'd need to manage a wait queue, allocate queue entries, and handle wakeup ordering.
I think also Go's own `sync.RWMutex` doesn't guarantee fairness, writers can starve readers and vice versa..

I'll add a comment documenting this limitation so future users are aware.
something like:
```
// GroupLock does not guarantee fairness. Under heavy load from one group,
// the other group may experience temporary delays.
```
**what do you think?** 


---- 
In our case, when few pods are created\deleted in same time I think there is big chance for:

### time | description 
T1 - `NodeStageVolume()` for podA is called- try to run `nvme connect` but see that is already connected - ✅
T2 -  `NodeUnStageVolume()` for podB is called- it sees no more mounted paths exists , so it runs `nvme disconnect` - ✅
T3 - `NodeStageVolume()` for podA then continue to run, it tries to mount the staging path - but failed- the nvme device does not exist anymore. - ❌

nixpanic:
There is no need to implement fairness/queuing. I do not expect significant issues for the intended use-case (NVMe-oF connect/disconnect). The number of operations that will get blocked should be minimal as (un)staging volumes is not done in a continuous stream anyway.

A note like the proposed
```
// GroupLock does not guarantee fairness. Under heavy load from one group,
// the other group may experience temporary delays.
```
would be very good to have as a reminder though.

nixpanic (approved):

gadididi:
@Mergifyio rebase

Rakshith-R (approved):

nixpanic:
@mergifyio rebase

PR #5672, opened 2025-10-15
ci: use Rook 1.19.3 and Ceph Tentacle

nixpanic:
Recent versions of Rook use Ceph-CSI Operator by default. There is no need to have Ceph-CSI deployed by Rook, as our CI deploys it from the PR that is being tested.

Rook v1.18.4 works well with Ceph Squid (v19). But for NVMe-oF testing we need a Ceph Tentacle (v20) cluster.

Closes: #5772

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

* `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

PR #6191, opened 2026-03-20
ci: mirror Rook 1.19.2

nixpanic:
Rook 1.19.2 has been releases a few weeks ago. Include it in the CI
registry to that we can run e2e with it.

See-also: #5672
Depends-on: #6199

### DIFF
--- a/mirror/images.txt
+++ b/mirror/images.txt
@@ -7,7 +7,7 @@
 #
 
 # ceph-csi devel
-quay.io/rook/ceph:v1.18.4        rook/ceph:v1.18.4
+quay.io/rook/ceph:v1.19.2        rook/ceph:v1.19.2
 quay.io/ceph/ceph:v20            ceph/ceph:v20
 
 # ceph-csi-operator


Madhu-1 (approved):

Madhu-1:
@nixpanic if the makefile is not updated to use Rook 1.19.2 our CI will fail as it will not be able to pull the image? i dont think so but just confiriming

iPraveenParihar (approved):

nixpanic:
> @nixpanic if the makefile is not updated to use Rook 1.19.2 our CI will fail as it will not be able to pull the image? i dont think so but just confiriming

No, that should not be an issue. The older versions are kept in the registry, we don't delete/prune those.

nixpanic:
@mergifyio queue

nixpanic:
#6199 is expected to address the Mergify check for the ci/centos branch. Once that is merged, `@mergifyio refresh` may be needed to re-evaluate the Mergify rules.

nixpanic:
@mergifyio refresh

PR #6200, opened 2026-03-24
rbd: fix DiffIterateByID Length calculation and add callback tests

kaovilai:
## Summary

- **Fix**: `DiffIterateByIDConfig.Length` was set to `VolSize` (total volume size), but `rbd_diff_iterate3` treats `len` as bytes to scan **from** `ofs`. When `startingOffset > 0`, this scanned `[startingOffset, startingOffset + VolSize)` instead of `[startingOffset, VolSize)`. Fixed by using `VolSize - startingOffset`.
- **Tests**: Add 7 unit tests for `createDiffIterateByIDCB` covering block collection, LUKS padding adjustment, batching, error propagation, context cancellation, and edge cases.

## Note

This project requires CGo (Ceph C libraries `librbd.h`, `librados.h`) to compile. The tests cannot be validated on a standard dev machine without these libraries — relying on CI to verify.

## Test plan

- [ ] CI passes `go test ./internal/rbd/ -run Test_createDiffIterateByIDCB`
- [ ] CI passes `go test ./internal/rbd/ -run Test_handleDiffIterateError`
- [ ] Existing tests unaffected

Ref: https://github.com/kaovilai/cephcsi-cbt-e2e/issues/2

🤖 Generated with [Claude Code](https://claude.com/claude-code)
via [Happy](https://happy.engineering)

PR #6203, opened 2026-03-24
rbd: add debug logging to ProcessMetadata for CBT diagnostics

### LINKED ISSUE #6206: rbd: getRBDSnapID and ProcessMetadata issues in snapshot metadata path
## Problem

Several issues exist in the RBD snapshot metadata code path (`ProcessMetadata` / `getRBDSnapID`) that would cause incorrect behavior for certain snapshot configurations:

### 1. `getRBDSnapID` uses `GetParent()` which fails for non-clone snapshots

`GetParent()` only works for clone-based snapshots with a parent relationship. For regular RBD snapshots or flattened clones, `GetParent()` returns `ErrNotFound`. Should use `GetSnapshotNames()` which reliably lists all snapshots on the image regardless of clone/parent relationships.

### 2. `SnapInfo` field name mismatch: `Id` vs `ID`

The go-ceph `SnapInfo` struct uses the field name `Id`, not `ID`.

### 3. `ProcessMetadata` opens wrong image

`ProcessMetadata` opens the source volume via `rbdSnap.open()` (which uses `RbdImageName` = source volume name), but `getRBDSnapID()` returns the snapshot ID from the clone image (via `toVolume()`). This mismatch means `SetSnapByID()` applies a clone's snapshot ID to the source volume, resulting in either an error or iterating over wrong/empty data.

## Context

These issues were discovered during investigation of empty `GetMetadataAllocated` streams. While the original symptom turned out to be a test-side device path mismatch ([details](https://github.com/kaovilai/cephcsi-cbt-e2e/issues/2#issuecomment-4122249418)), the code issues identified above are real bugs that should be fixed.

> [!Note]
> Responses generated with Claude

kaovilai:
## Summary

Add debug log statements to `ProcessMetadata` in `snap_diff.go` to improve observability of the CBT (Changed Block Tracking) metadata streaming path.

## Why this is useful beyond a specific bug

The `ProcessMetadata` function involves multiple steps (snapshot ID resolution, snap context setup, DiffIterateByID call) but currently has no logging between them. When `GetMetadataAllocated` returns unexpected results (empty streams, wrong block counts), there's no visibility into:

- Which snapshot ID was resolved and whether it matches expectations
- What Offset/Length/FromSnapID parameters were passed to `DiffIterateByID`
- How many blocks were buffered but not yet sent

These are all `DebugLog` calls (not visible at default log levels), so there's no performance impact in production.

## Changes

Three `log.DebugLog` calls added:
1. After snapshot ID resolution — logs snapshot name, ID, and volume size
2. Before `DiffIterateByID` — logs config parameters (Offset, Length, FromSnapID)
3. After `DiffIterateByID` completes — logs remaining buffered block count

Fixes: #6206

> [!Note]
> Responses generated with Claude

### DIFF
--- a/internal/rbd/snap_diff.go
+++ b/internal/rbd/snap_diff.go
@@ -58,6 +58,8 @@ func (rbdSnap *rbdSnapshot) ProcessMetadata(
 		return fmt.Errorf("failed to get snapshot ID of %q: %w", rbdSnap, err)
 	}
 
+	log.DebugLog(ctx, "ProcessMetadata: snapshot=%q, snapID=%d, VolSize=%d", rbdSnap, snapID, rbdSnap.VolSize)
+
 	err = image.SetSnapByID(snapID)
 	if err != nil {
 		return fmt.Errorf("failed to set snapshot %q by ID %d: %w",
@@ -109,11 +111,16 @@ func (rbdSnap *rbdSnapshot) ProcessMetadata(
 		diffIterateByIDConfig.FromSnapID = fromSnapID
 	}
 
+	log.DebugLog(ctx, "ProcessMetadata: DiffIterateByID config: Offset=%d, Length=%d, FromSnapID=%d",
+		diffIterateByIDConfig.Offset, diffIterateByIDConfig.Length, diffIterateByIDConfig.FromSnapID)
+
 	diffIterateErr := image.DiffIterateByID(diffIterateByIDConfig)
 	err = handleDiffIterateError(diffIterateErr)
 	if err != nil {
 		return fmt.Errorf("failed to get diff: %w", err)
 	}
+
+	log.DebugLog(ctx, "ProcessMetadata: DiffIterateByID completed successfully")
 
 	if len(changedBlocks) != 0 {
 		// Send any remaining changed blocks after the loop.


nixpanic:
Is this debugging useful outside the scope where a potential bug needs to be addressed?

Can you explain a little more about the issue that you are facing?

To get the PR accepted, you will need to squash the extra/correction commits into the original one.

kaovilai:
squashed, pr body updated.

nixpanic (dismissed):

nixpanic:
@Rakshith-R , please have a look at this and #6200

internal/rbd/snap_diff.go
```diff
@@ -107,12 +109,17 @@ func (rbdSnap *rbdSnapshot) ProcessMetadata(
...
+	log.DebugLog(ctx, "ProcessMetadata: DiffIterateByID config: Offset=%d, Length=%d, FromSnapID=%d",
+		diffIterateByIDConfig.Offset, diffIterateByIDConfig.Length, diffIterateByIDConfig.FromSnapID)
+
 	diffIterateErr := image.DiffIterateByID(diffIterateByIDConfig)
 	err = handleDiffIterateError(diffIterateErr)
 	if err != nil {
 		return fmt.Errorf("failed to get diff: %w", err)
 	}
 
+	log.DebugLog(ctx, "ProcessMetadata: DiffIterateByID returned successfully, buffered blocks=%d", len(changedBlocks))
```
Rakshith-R:
```suggestion
	log.DebugLog(ctx, "ProcessMetadata: DiffIterateByID completed successfully")
```
we can just move it after returning the remaining blocks.

Rakshith-R (commented):
a small nit, LGTM

Rakshith-R:
@kaovilai Can you please squash into single commit ? 
That will fix commitlint failure too

kaovilai:
fixed

nixpanic (approved):

nixpanic:
@mergifyio rebase

nixpanic:
@Rakshith-R , is there still something that needs to be addressed?

Rakshith-R (approved):

PR #6207, opened 2026-03-27
nvmeof: add allowAnyHost StorageClass parameter

nixpanic:
The new optional allowAnyHost parameter allows subsystems to accept
connections from any NVMe initiator by adding a wildcard host ("*").
This bypasses per-host access control and enables volumes to be exposed
to external clients without explicit host registration.

The parameter defaults to false to maintain secure per-host access
control by default.

Test image: `quay.io/nixpanic/cephcsi:nvmeof-allowAnyHost`

/cc @gadididi @OdedViner

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -21,5 +21,7 @@
 - nvmeof: add Kubernetes ServiceAccount based volume access restriction
 - cephfs: add Kubernetes ServiceAccount based volume access restriction
 - nfs: add Kubernetes ServiceAccount based volume access restriction
+- nvmeof: the new `allowAnyHost` StorageClass option makes it possible to
+  expose created volumes to external clients
 
 ## NOTE
--- a/examples/nvmeof/storageclass.yaml
+++ b/examples/nvmeof/storageclass.yaml
@@ -28,6 +28,10 @@ parameters:
   # subsystemNQN is the NQN of the subsystem where all volumes belong to
   subsystemNQN: "nqn.2025-08.io.ceph:k8s-ceph-csi"
 
+  # allowAnyHost allows any host to connect to the subsystem. When enabled,
+  # explicit per-host access control is bypassed
+  # allowAnyHost: "true"
+
   # connection details for the gateway (management API)
   nvmeofGatewayAddress: 10.242.64.32
   nvmeofGatewayPort: "5500"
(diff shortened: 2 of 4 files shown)


nixpanic:
@OdedViner and @gadididi,

As an alternative to the StorageClass parameter, we could have `allowAnyHost` as an option for `ControllerModifyVolume` that users can set with a VolumeAttribuesClass. Or even include a list of `hostNQN`s (with value `*` for any-host) instead of a blanket `allowAnyHost` parameter.

Thoughts?

OdedViner:
Hi @nixpanic , I have a couple of questions:

1.Can I use the ceph-csi-operator to test the `quay.io/nixpanic/cephcsi:nvmeof-allowAnyHost` image?
2.To enable the new functionality , is it sufficient to just add `allowAnyHost: "true"` to the StorageClass parameters like this?
```
parameters:
  clusterID: rook-ceph
  pool: nvmeof
  allowAnyHost: "true"
```

nixpanic:
Hi @OdedViner ,

in your namespace where the Driver CR for the NVMe-oF deployment (managed by ceph-csi-operator), you can create a ConfigMap like this:

```
apiVersion: v1
kind: ConfigMap
metadata:
    name: ceph-csi-imageset
data:
    "plugin": "quay.io/nixpanic/cephcsi:nvmeof-allowAnyHost"
```

The  Driver CR has a `spec.imageSet` object reference that you can set to an alternative  ConfigMap as well. That way you would not change the image for all drivers. Restarting ceph-csi-operator may be required after making this change.

And yes, adding `allowAnyHost` to the StorageClass as in your example should be sufficient.

nixpanic:
> As an alternative to the StorageClass parameter, we could have `allowAnyHost` as an option for `ControllerModifyVolume` that users can set with a VolumeAttribuesClass. Or even include a list of `hostNQN`s (with value `*` for any-host) instead of a blanket `allowAnyHost` parameter.

This is now (untested) available at https://github.com/nixpanic/ceph-csi/tree/nvmeof/ControllerModifyVolume/allowHostNQNs . Examples for the VolumeAttributesClass are included as `examples/nvmeof/vac-*` files.

nixpanic:
@gadididi will work on a way to do this through `ControllerModifyVolume`, with more correct handling of permissions of the subsystems and visibility of the nvme-of namespaces.

PR #6229, opened 2026-04-10
rbd: add mounter type validation for ControllerModifyVolume

### LINKED ISSUE #6186: rbd: ControllerModifyVolume should report an error when setting QoS on non rbd-nbd volumes
> actually, there is one thing that probably should be added here. A check if the volume uses rbd-nbd is crucial. If it isn't the case, setting QoS parameters will not have any effect.
> 
> A [gRPC INVALID_ARGUMENT error](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllermodifyvolume-errors) should be returned for non rbd-nbd volumes 

 _Originally posted by @nixpanic in [#6160](https://github.com/ceph/ceph-csi/pull/6160/changes#r2939046523)_

---

This isn't trivial. There is `rbdVol.Mounter` that would be set to `rbd-nbd`, but `genVolFromVolIDWithMigration` returns an empty `rbdVol.Mounter`. The logic within `genVolFromVolIDWithMigration` does not assign a value to `rbdVol.Mounter`.

nixpanic:
RBD QoS parameters are only supported with the rbd-nbd mounter, but
ControllerModifyVolume previously had no way to enforce this.

This adds infrastructure to track the mounter type of RBD volumes via image
metadata (mounterKey), stored at volume creation and cleaned up on deletion. A
new UsesNBDMounter() method fetches and caches this value lazily.

ControllerModifyVolume now validates the mounter type before applying QoS
changes:

- Volumes with unknown mounter (created before mounter tracking) log a warning
  and proceed
- Metadata fetch failures return an Internal error
- Volumes using a non-rbd-nbd mounter return InvalidArgument

Fixes: #6186

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -5781,6 +5781,7 @@ var _ = Describe("RBD", func() {
 				logAndFail("failed to delete storageclass: %v", err)
 			}
 			err = createRBDStorageClass(f.ClientSet, f, defaultSCName, nil, map[string]string{
+				"mounter": "rbd-nbd",
 				"csi.storage.k8s.io/controller-modify-secret-namespace": cephCSINamespace,
 				"csi.storage.k8s.io/controller-modify-secret-name":      rbdProvisionerSecretName,
 			}, deletePolicy)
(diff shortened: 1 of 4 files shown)


Rakshith-R (approved):

black-dragon74 (approved):

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

PR #6234, opened 2026-04-14
ci: set timeout for k8s-external-storage to 90 minutes

nixpanic:
1 hour is too little with the added nvme-of testing.

See-also: #6214


nixpanic:
Probably isn't needed when the tests work correctly (no retry for 10 or 15 minutes).

PR #6257, opened 2026-04-29
rbd: fix healer staging path for Block volumeMode

YLShiJustFly:
rbd: fix healer staging path for Block volumeMode

Block volumeMode PVCs using rbd-nbd mounter permanently lose IO after
CSI plugin pod restart due to two bugs in rbd_healer.go:

1. formatStagingTargetPath computes a filesystem-format path
   (.../csi/<driver>/<sha256>/globalmount) for all volumes, but Block
   volumes use a completely different path:
   .../csi/volumeDevices/staging/<pv-name>/
   ValidateNodeStageVolumeRequest calls checkDirExists before any healer
   logic, so the wrong path causes an immediate InvalidArgument error and
   attachRBDImage never runs.

2. callNodeStageVolume detects Block volumes using
   CSI.FSType == "block", but Block volumeMode PVs have an empty FSType.
   This is always false, causing the healer to send a Mount capability
   instead of a Block capability.

Filesystem PVCs are unaffected because their sha256 path is correct.
Block PVCs have no fallback recovery path (kubelet does not re-call
NodeStageVolume when the bind mount is still in mountinfo), so IO is
permanently lost until the node is restarted.

Fix by:
- Adding a Block branch to formatStagingTargetPath using pv.Name as the
  staging subdirectory, matching kubelet actual path layout
- Changing Block detection to use pv.Spec.VolumeMode
- Removing the unused fsTypeBlockName constant

### DIFF
--- a/e2e/rbd.go
+++ b/e2e/rbd.go
@@ -2621,6 +2621,155 @@ var _ = Describe("RBD", func() {
 			}
 		})
 
+		It("perform IO on rbd-nbd Block volume after nodeplugin restart", func() {
+			if !testNBD {
+				framework.Logf("skipping NBD test")
+
+				return
+			}
+
+			err := deleteResource(rbdExamplePath + "storageclass.yaml")
+			if err != nil {
+				logAndFail("failed to delete storageclass: %v", err)
+			}
+			// Storage class with rbd-nbd mounter
+			err = createRBDStorageClass(
+				f.ClientSet,
+				f,
+				defaultSCName,
+				nil,
+				map[string]string{
+					"mounter":         "rbd-nbd",
+					"mapOptions":      nbdMapOptions,
+					"cephLogStrategy": e2eDefaultCephLogStrategy,
+				},
+				deletePolicy)
+			if err != nil {
+				logAndFail("failed to create storageclass: %v", err)
+			}
+			pvc, err := loadPVC(rawPvcPath)
+			if err != nil {
+				logAndFail("failed to load PVC: %v", err)
+			}
+			pvc.Namespace = f.UniqueName
+
+			app, err := loadApp(rawAppPath)
+			if err != nil {
+				logAndFail("failed to load application: %v", err)
+			}
+			app.Namespace = f.UniqueName
+			label := map[string]string{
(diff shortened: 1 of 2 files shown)


nixpanic:
Thanks for the PR! Please correct the subject of the commit, so that it says:

    rbd: fix healer staging path and capability type for Block volumeMode

Ideally an e2e test is added for this as well. Please check e2e/rbd.go for a similar test that uses rbd-nbd and a filesystem volume,

YLShiJustFly:
Thanks for the review! I've addressed both points:
1. Fixed the
commit subject to
`rbd:
fix healer staging path and capability type for
Block volumeMode`
2. Added an e2e test
`perform IO on rbd-nbd Block volume after nodeplugin restart`
in e2e/rbd. go, following the pattern of the existing filesystem volume healer test

iPraveenParihar:
@mergifyio rebase

iPraveenParihar:
@mergifyio rebase

Madhu-1 (approved):

iPraveenParihar (approved):

nixpanic:
@mergifyio rebase

nixpanic:
> /retest ci/centos/k8s-e2e-external-storage/1.34

Deployment failed [logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/k8s-e2e-external-storage-1.34/detail/k8s-e2e-external-storage-1.34/761/pipeline)

YLShiJustFly:
Hi @nixpanic,

Just a gentle reminder regarding this PR. It has already received 2 approvals. 

Regarding the current CI failures:
- The `k8s-1.34` failure is the exact same JSON unmarshaling error (`cannot unmarshal string into Go value of type []DeleterConfig`) that occurred in #6356, which is a broken upstream infra issue.
- The `k8s-1.36` failure is a known PodSecurity timeout.

Since these failures are completely unrelated to this RBD healer patch (which only refactors the internal path string formatting), could you please help apply the 'approved' label to bypass the CI and manually merge this, just like #6356? 

Thank you for your time and help!

PR #6264, opened 2026-05-07
doc: update the qos design doc

Madhu-1:
Updating the RBD qos design doc to consider only VAC for the implementation.


Rakshith-R (dismissed):
> To identify the right cgroup file, we need pod UUID and container UUID from the pod yaml output


Let's add details on how CephCSI driver will get this information during nodePublish too ?
I guess that will complete the whole picture.

Madhu-1:
> To identify the right cgroup file, we need pod UUID and container UUID from the pod yaml output

Updated the details

Rakshith-R (approved):
Thanks !

black-dragon74 (approved):

PR #6266, opened 2026-05-09
rbd: correct error handling in flattenRbdImage and flatten functions

dkhorasanizadeh:

This PR fixes issues in the RBD flatten flow, mainly around fallback behavior when Ceph manager task scheduling is not available, and improves error handling to avoid hiding the real problem.
In `flattenRbdImage()`, the clone depth check was previously skipped when `forceFlatten` was not set, but the later fallback logic could still rely on the depth value when Ceph manager task scheduling was not used. In cases where the code could not schedule the flatten operation through the manager task interface, for example because of missing permissions to access the Ceph manager, it could fall back to direct flattening. However, since depth was only being calculated in the forceFlatten path, the value could remain zero when reaching the condition `forceFlatten || depth >= hardlimit`. Because of that, the condition would never be satisfied in some fallback cases, so the image would not be flattened when it should be.
This can become a real issue once the clone chain reaches the kernel RBD driver imposed limit of 16, because at that point the image can no longer be mapped on a node.
The previous error handling also made this harder to debug, because it could return an irrelevant error and hide the actual reason why flattening never happened.

In addition, the `flatten()` function has been fixed to stop silently ignoring unexpected errors. Previously, it implicitly assumed that a flatten failure could only happen when the image had no parent. With this change, the code now explicitly checks for `librbd.ErrNotFound` from `GetParent()` as the only acceptable case to ignore, and all other errors are returned properly with context.

**Checklist:**

---

<details>

</details>

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -991,11 +991,11 @@ func (ri *rbdImage) flattenRbdImage(
 			err)
 		if forceFlatten || depth >= hardlimit {
 			flattenImageErr := ri.flatten()
-			if err != nil {
-				log.ErrorLog(ctx, "rbd failed to flatten image %s %s: %v", ri.Pool, ri.RbdImageName, err)
+			if flattenImageErr != nil {
+				log.ErrorLog(ctx, "rbd failed to flatten image %s %s: %v", ri.Pool, ri.RbdImageName, flattenImageErr)
 
 				return fmt.Errorf(
-					"failed to add task to remove image: %w, failed to flatten image: %w",
+					"failed to add task to flatten image: %w, failed to flatten image: %w",
 					knownErr,
 					flattenImageErr,
 				)
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -1006,21 +1006,6 @@ func (ri *rbdImage) flattenRbdImage(
 	return nil
 }
 
-func (ri *rbdImage) getParentName() (string, error) {
-	rbdImage, err := ri.open()
-	if err != nil {
-		return "", err
-	}
-	defer rbdImage.Close() //nolint:errcheck // not a critical failure
-
-	parentInfo, err := rbdImage.GetParent()
-	if err != nil {
-		return "", err
-	}
-
-	return parentInfo.Image.ImageName, nil
-}
-
 func (ri *rbdImage) flatten() error {
 	rbdImage, err := ri.open()
 	if err != nil {
@@ -1031,13 +1016,16 @@ func (ri *rbdImage) flatten() error {
 	err = rbdImage.Flatten()
 	if err != nil {
 		// rbd image flatten will fail if the rbd image does not have a parent
-		parent, pErr := ri.getParentName()
-		if pErr != nil {
-			return fmt.Errorf("Failed as %w (internal %w)", err, pErr)
-		}
-		if parent == "" {
+		_, pErr := rbdImage.GetParent()
+		if errors.Is(pErr, librbd.ErrNotFound) {
 			return nil
 		}
+
+		if pErr != nil {
+			return fmt.Errorf("failed to flatten: (%w) and get parent: (%w)", err, pErr)
+		}
+
+		return fmt.Errorf("failed to flatten image: %w", err)
 	}
 
 	return nil


YiteGu:
hey, @dkhorasanizadeh  I want to confirm a question in your commit message:
> In flattenRbdImage(), the clone depth check was previously skipped when forceFlatten was not set

The clone depth check was not skipped when forceFlatten was not set.
not set is `forceFlatten = false`, execu clone depth check.
set is `forceFlatten = true`, skip clone depth check.

dkhorasanizadeh:
Hey @YiteGu, thanks for reviewing my code. What you described is indeed the expected behavior. However, in the current implementation, the chain depth is only computed when "forceFlatten = true". As a result, when "forceFlatten = false", the depth ends up being 0, which is incorrect.

internal/rbd/rbd_util.go
```diff
@@ -945,59 +948,53 @@ func (ri *rbdImage) flattenRbdImage(
...
+		return nil
 	}
-	if !rbdCephMgrSupported && knownErr != nil {
-		log.ErrorLog(
-			ctx,
-			"task manager does not support flatten,image will be flattened once hardlimit is reached: %v",
-			err)
-		if forceFlatten || depth >= hardlimit {
-			flattenImageErr := ri.flatten()
-			if err != nil {
```
YiteGu, line 1021: I believe the key issue with the fix is ​​that `err != nil` should be replaced with `flattenImageErr != nil`.

dkhorasanizadeh: You're right—the main and most essential piece of the fix is returning the proper error. I removed the unrelated parts of the change to make the diff more relevant.

YiteGu (approved):

dkhorasanizadeh:
Hi @Rakshith-R, I'd appreciate your review on this PR whenever you get a chance. This fix addresses an issue we're currently facing in our production environment, so we're hoping to get it merged as soon as possible. Let me know if you have any questions or if there's anything I can clarify in the meantime. Thanks!

Rakshith-R:
@dkhorasanizadeh Can you please create multiple specific commits for each issue in this same pr? 
Remove unnecessary changes like formatting so it is easier to review. 
Flattening logic is a bit complex in CephCSI :sweat: .

dkhorasanizadeh:
> @dkhorasanizadeh Can you please create multiple specific commits for each issue in this same pr? Remove unnecessary changes like formatting so it is easier to review. Flattening logic is a bit complex in CephCSI 😓 .

Hi @Rakshith-R, thanks for the feedback and sorry for the delay. I've split the change into two focused commits and dropped the unrelated formatting/renaming noise so the diff only shows what's necessary for each fix:

1. rbd: fix successful flatten wrongly reported as failure — the fallback path (task manager doesn't support flatten) was checking the stale AddFlatten error instead of ri.flatten()'s own return value, so a successful synchronous flatten was always wrapped and returned as an error. Also fixes the "remove image" → "flatten image" wording typo in that error message.
2. rbd: fix flatten() swallowing errors when image has a parent — flatten() used getParentName(), which returned "" on any GetParent() failure, not just "no parent," so a real flatten error on an image that does have a parent was silently discarded. Now checks errors.Is(pErr, librbd.ErrNotFound) directly and removes the now-unused getParentName() helper.

internal/rbd/rbd_util.go
```diff
@@ -1045,13 +1016,16 @@ func (ri *rbdImage) flatten() error {
...
-			return fmt.Errorf("Failed as %w (internal %w)", err, pErr)
-		}
-		if parent == "" {
+		_, pErr := rbdImage.GetParent()
+		if errors.Is(pErr, librbd.ErrNotFound) {
 			return nil
 		}
+
+		if pErr != nil {
+			return fmt.Errorf("failed to flatten: %w, also failed to get parent: %w", err, pErr)
```
black-dragon74:
```suggestion
			return fmt.Errorf("failed to flatten: (%w) and get parent: (%w)", err, pErr)
```

black-dragon74:
@dkhorasanizadeh please sign-off your commits by: `git commit -s`. That would resolve the failing DCO test.

internal/rbd/rbd_util.go
dkhorasanizadeh: Applied — thanks!

dkhorasanizadeh:
> @dkhorasanizadeh please sign-off your commits by: `git commit -s`. That would resolve the failing DCO test.

Thanks @black-dragon74, good catch. The Signed-off-by trailer was there, but the commits were authored/committed with a different local email than the one in the trailer, which is what tripped the DCO check. I've fixed the author/committer identity on both commits to match the sign-off and force-pushed. Should be green now.

PR #6271, opened 2026-05-12
util: return success from ControllerUnpublish when volume is not found

iPraveenParihar:

```
When the backing pool or volume is force deleted before ControllerUnpublish is called,
the RPC fails with codes.Internal and the external-attacher retries indefinitely,
blocking PV deletion.

Return success when the volume is not found, consistent with how DeleteVolume
already handles these errors.
```

Resolves: https://github.com/rook/rook/pull/17462#discussion_r3193304883

**Checklist:**

---

<details>

</details>

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -1813,6 +1813,14 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
 
 	rv, err := GenVolFromVolID(ctx, volumeId, credentials, secrets)
 	if err != nil {
+		if errors.Is(err, util.ErrPoolNotFound) ||
+			errors.Is(err, util.ErrKeyNotFound) ||
+			errors.Is(err, rbderrors.ErrImageNotFound) {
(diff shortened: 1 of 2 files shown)


iPraveenParihar:
cc @Madhu-1 PTAL

Rakshith-R:
I don't think this change is necessary.
Explicitly deleting a mounted Volume from backend is a extreme disruptive unsupported action, that calls for manual cleanup of VA and other associated objects.

internal/rbd/controllerserver.go
```diff
@@ -1813,6 +1813,14 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
 
 	rv, err := GenVolFromVolID(ctx, volumeId, credentials, secrets)
 	if err != nil {
+		if errors.Is(err, util.ErrPoolNotFound) ||
```
Madhu-1, line 1816: we should return not found error message as per spec https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerunpublishvolume-errors

PR #6278, opened 2026-05-18
cephfs: use RemoveNetworkFence directly for unfencing

Rakshith-R:
CephFS unfencing was failing with EINVAL because
RemoveClientEviction tried to remove per-client nonce blocklist entries via go-ceph's OSDBlocklistRemove, which rejects Ceph entity address format (IP:port/nonce).

Per-client blocklist entries from MDS client eviction auto-expire and reconnecting clients get new nonces, so explicit removal is unnecessary. Call RemoveNetworkFence directly to remove only the CIDR range entry, matching the RBD unfence behavior.


Rakshith-R:
@mergifyio rebase

Madhu-1 (approved):

iPraveenParihar (approved):

iPraveenParihar:
@mergifyio rebase

iPraveenParihar:
@Rakshith-R
Skipping e2e as the code change is not covered in e2e suite.

PR #6279, opened 2026-05-18
rbd: enable WholeObject & IncludeParent in snap diff

Rakshith-R:
Enable WholeObject and IncludeParent flags in DiffIterateByIDConfig
used by ProcessMetadata, matching the existing DiffIterate config
in getUsedBytes. WholeObject improves performance by reporting
whole-object granularity diffs instead of byte-level diffs.
IncludeParent ensures diffs against parent snapshots are included.


Madhu-1 (approved):

iPraveenParihar (approved):

iPraveenParihar:
Adding skip/e2e label since we don't have e2e for this Today.

PR #6286, opened 2026-05-19
util: restart process when gRPC call is stuck for 10+ minutes

Rakshith-R:
Add a --feature-gates CLI flag that accepts comma-separated key=bool
pairs (e.g., SlowGRPCRestart=false). Unknown keys are rejected at
startup.

Add killOnSlowGRPC unary interceptor that calls os.Exit(1) if any
gRPC handler takes longer than 10 minutes; the kubelet restarts the
container in-place. The interceptor is conditionally wired into the
middleware chain based on the SlowGRPCRestart feature gate (enabled
by default).

Methods matching prefixes in slowGRPCSkipPrefixes are excluded. The
initial list contains only ReclaimSpace calls (/reclaimspace.), which
can legitimately take a long time on large volumes.

---

<details>

</details>


internal/csi-common/utils.go
```diff
@@ -407,6 +414,50 @@ func logSlowGRPC(
...
+
+// killOnSlowGRPC restarts the process if the gRPC handler does not return
+// within slowGRPCRestartThreshold; the kubelet restarts the container in-place.
+// ReclaimSpace calls are excluded as they can legitimately take a long time.
+func killOnSlowGRPC(
+	ctx context.Context,
+	req interface{},
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (interface{}, error) {
```
black-dragon74:
```suggestion
) (any, error) {
```
Newer versions of go/LSPs suggest using `any`.

internal/csi-common/utils.go
```diff
@@ -407,6 +414,50 @@ func logSlowGRPC(
...
+	return killOnSlowGRPCWithThreshold(slowGRPCRestartThreshold, ctx, req, info, handler)
+}
+
+func killOnSlowGRPCWithThreshold(
+	threshold time.Duration,
+	ctx context.Context,
+	req interface{},
```
black-dragon74:
```suggestion
	req any,
```

internal/csi-common/utils.go
```diff
@@ -407,6 +414,50 @@ func logSlowGRPC(
...
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (interface{}, error) {
```
black-dragon74:
```suggestion
) (any, error) {
```

internal/csi-common/utils.go
```diff
@@ -407,6 +414,50 @@ func logSlowGRPC(
...
+// operations, which are excluded from the stuck-call restart because they can
+// legitimately run for a long time on large volumes.
+const reclaimSpaceServicePrefix = "/reclaimspace."
+
+// killOnSlowGRPC restarts the process if the gRPC handler does not return
+// within slowGRPCRestartThreshold; the kubelet restarts the container in-place.
+// ReclaimSpace calls are excluded as they can legitimately take a long time.
+func killOnSlowGRPC(
+	ctx context.Context,
+	req interface{},
```
black-dragon74:
```suggestion
	req any,
```

internal/csi-common/utils.go
```diff
@@ -407,6 +414,50 @@ func logSlowGRPC(
...
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (interface{}, error) {
+	if strings.HasPrefix(info.FullMethod, reclaimSpaceServicePrefix) {
+		return handler(ctx, req)
+	}
+	done := make(chan struct{})
+	defer close(done)
```
black-dragon74: Stop must be called before closing the channel to prevent races / incorrect restarts. 

Remove the defer and timer.Stop from select and call them sequentially. 
```go
resp, err := handler(ctx, req)

timer.Stop()
close(done)

return resp, err
```

black-dragon74 (changes requested):

internal/csi-common/utils.go
Rakshith-R: done, ptal

black-dragon74 (dismissed):

black-dragon74 (dismissed):

internal/csi-common/utils.go
```diff
@@ -407,6 +414,52 @@ func logSlowGRPC(
 	return resp, err
 }
 
+// reclaimSpaceServicePrefix is the gRPC service prefix for ReclaimSpace
+// operations, which are excluded from the stuck-call restart because they can
+// legitimately run for a long time on large volumes.
+const reclaimSpaceServicePrefix = "/reclaimspace."
```
Madhu-1: we should make it generic so that we can expand it, add a helper function to skip known services

Madhu-1 (commented):
changes looks good, we need to put this functionality behind a flag (enable is by default but provide option to disable it)

Rakshith-R:
> changes looks good, we need to put this functionality behind a flag (enable is by default but provide option to disable it)

>we should make it generic so that we can expand it, add a helper function to skip known services

Done, added env var `DISABLE_CSI_SLOW_GRPC_RESTART` flag to disable and a skip services list that we can append at later point of time.

PTAL

black-dragon74:
We usually use flags/launch args to enable/disable features.

Rakshith-R:
> We usually use flags/launch args to enable/disable features.

We recently decided against it cause difficulty in upgrades/backward compatibility/API problems to pass settings and moving towards env var.

black-dragon74 (dismissed):

Madhu-1:
> > We usually use flags/launch args to enable/disable features.
> 
> We recently decided against it cause difficulty in upgrades/backward compatibility/API problems to pass settings and moving towards env var.

where this was decided,  i would vote for a feature-gate CLI command which accepts the `feature1=true,feature2=false` like this where we can provide option for user to enable/disable any features instead of the env variable

Rakshith-R:
> > > We usually use flags/launch args to enable/disable features.
> > 
> > 
> > We recently decided against it cause difficulty in upgrades/backward compatibility/API problems to pass settings and moving towards env var.
> 
> where this was decided, i would vote for a feature-gate CLI command which accepts the `feature1=true,feature2=false` like this where we can provide option for user to enable/disable any features instead of the env variable

done, ptal

Madhu-1 (approved):

black-dragon74 (approved):

Rakshith-R:
@mergifyio queue

PR #6287, opened 2026-05-19
helm: add PodMonitor support to ceph-csi-rbd Helm chart

nepomucen:
This pull request adds support for monitoring the provisioner component of the Ceph-CSI RBD Helm chart via Prometheus by introducing an optional `PodMonitor` resource.
This enables users to easily integrate metrics collection for the provisioner and its sidecars when Prometheus Operator is installed.

`provisioner.podMonitor.enabled=false`

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/charts/ceph-csi-rbd/templates/nodeplugin-podmonitor.yaml
+++ b/charts/ceph-csi-rbd/templates/nodeplugin-podmonitor.yaml
@@ -1,0 +1,28 @@
+{{- if and .Values.nodeplugin.podMonitor.enabled .Values.nodeplugin.httpMetrics.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: {{ include "ceph-csi-rbd.nodeplugin.fullname" . }}
+  namespace: {{ .Values.nodeplugin.podMonitor.namespace | default .Release.Namespace }}
+  labels:
+    app: {{ include "ceph-csi-rbd.name" . }}
+    chart: {{ include "ceph-csi-rbd.chart" . }}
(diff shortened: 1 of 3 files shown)


nepomucen:
@jd would you be so kind and review my PR. This is something we have been patching ever since with `strategicMergePatches` in our helm release. Would be very nice to have it included in the official helm chart.

jd:
> @jd would you be so kind and review my PR. This is something we have been patching ever since with `strategicMergePatches` in our helm release. Would be very nice to have it included in the official helm chart.

Not really my thing, sorry.

nixpanic:
This is quite nice and simple. But the Helm Charts are being deprecated (slowly), and replaced by [Ceph-CSI Operator](https://github.com/ceph/ceph-csi-operator). It probably makes sense to add (optional) support for PodMonitor to Ceph-CSI Operator instead.

Is there a good reason to only add it for provisioners, and not the node-plugins? PodMonitor should be added to other storage types too.

nepomucen:
Thank you for the review. You’re right, there was no strong reason to limit PodMonitor support to the provisioner. 
That was mainly based on my personal use case. I’ve now extended PodMonitor support to the NodePlugin as well.

nixpanic:
To come back on my own question:

> Is there a good reason to only add it for provisioners, and not the node-plugins? PodMonitor should be added to other storage types too.

Node-plugins usually get deployed with host-networking. Opening the metrics ports on the host isn't trivial, it can easily lead to conflicts causing containers failing to start.

ceph/ceph-csi-operator#578 has been filed, and should also take those constraints into account.

PR #6291, opened 2026-05-21
csi: add VolumeLocks to NodeGetVolumeStats for RBD and CephFS

Rakshith-R:
Acquire a targetPath lock in NodeGetVolumeStats to prevent running
multiple NodeGetVolumeStats call on same target path at once. This may
happen when ceph cluster has problems. This also matches the semantics
of locking in other gRPC calls.

---

<details>

</details>

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -931,6 +931,11 @@ func (ns *NodeServer) NodeGetVolumeStats(
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	if acquired := ns.VolumeLocks.TryAcquire(targetPath); !acquired {
+		return nil, status.Errorf(codes.Aborted, util.TargetPathOperationAlreadyExistsFmt, targetPath)
+	}
+	defer ns.VolumeLocks.Release(targetPath)
+
 	// health check first, return without stats if unhealthy
 	healthy, msg := ns.healthChecker.IsHealthy(volumeID, targetPath)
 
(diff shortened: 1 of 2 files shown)


black-dragon74 (approved):

black-dragon74:
The only issue I could think of is this RPC is rather too frequent. We should be fine though as the locks are in-memory?

Madhu-1 (approved):

internal/cephfs/nodeserver.go
```diff
@@ -931,6 +931,11 @@ func (ns *NodeServer) NodeGetVolumeStats(
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	if acquired := ns.VolumeLocks.TryAcquire(targetPath); !acquired {
```
iPraveenParihar, line 934: NodeUnpublishVolume uses targetPath to take lock, if healthchecker hungs NodeUnpublishVolume is blocked. Is that what we want?

Rakshith-R: > NodeUnpublishVolume uses targetPath to take lock, if healthchecker hungs NodeUnpublishVolume is blocked. Is that what we want?

It'll get killed in 10 mins, it's fine and
If health check hangs, even unmap will hang too.

Rakshith-R:
@mergifyio queue

iPraveenParihar:
Run e2e after #6286 is merged.

Rakshith-R:
@mergifyio rebase

Rakshith-R:
@mergifyio queue

PR #6301, opened 2026-05-28
nvmeof: gate DH-CHAP on kernel support

nixpanic:
Track DH-CHAP kernel support in the NVMe-oF node server and
reject DH-CHAP connections when the running kernel is older
than v6.0.0.

See-also: https://github.com/torvalds/linux/commit/88b140fe

### DIFF
--- a/internal/nvmeof/nodeserver/nodeserver.go
+++ b/internal/nvmeof/nodeserver/nodeserver.go
@@ -36,6 +36,7 @@ import (
 	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/lock"
 	"github.com/ceph/ceph-csi/internal/util/log"
+	"github.com/ceph/ceph-csi/pkg/util/kernel"
 )
 
 // NodeServer struct of ceph nvmeof driver with supported methods of CSI
@@ -47,6 +48,9 @@ type NodeServer struct {
 	volumeLocks *util.IDLocker
 
 	initiator nvmeof.NVMeInitiator
+
+	// dhchapUnsupportedKernelErr stores the error returned when the running kernel does not support DH-CHAP.
+	dhchapUnsupportedKernelErr error
 
 	// securityKeys manages DH-CHAP and PSK\TLS keys
 	securityKeys nvmeof.SecurityKeyManager
@@ -101,6 +105,17 @@ const (
 	authenticationKMSID = "authenticationKMSID"
 )
 
+var dhchapSupport = []kernel.KernelVersion{
+	{
+		Version:      6,
+		PatchLevel:   0,
+		SubLevel:     0,
+		ExtraVersion: 0,
+		Distribution: "",
+		Backport:     false,
+	},
+}
+
 // NewNodeServer initialize a node server for ceph CSI driver.
 func NewNodeServer(
 	d *csicommon.CSIDriver,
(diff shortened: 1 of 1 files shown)


internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -48,6 +49,9 @@ type NodeServer struct {
 
 	initiator nvmeof.NVMeInitiator
 
+	// dhchapHasKernelSupport indicates whether the running kernel supports DH-CHAP.
+	dhchapHasKernelSupport error
```
gadididi: nit: I think the naming is a bit confusing, maybe `dhchapKernelCheckErr` or something else? wdyt?

nixpanic: yeah, could be. I've used `dhchapUnsupportedKernelErr` now.

gadididi (dismissed):
LGTM! thanks

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -110,20 +125,30 @@ func NewNodeServer(
...
 	}
 
 	// Load nvme kernel modules
 	if err := nvmeInitiator.LoadKernelModules(context.Background()); err != nil {
 		return nil, fmt.Errorf("failed to load NVMe kernel modules: %w", err)
 	}
 
+	kernelRelease, err := kernel.GetKernelVersion()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get kernel version: %w", err)
```
iPraveenParihar, line 145: Should we just log it and not error out? Since DH-CHAP is a StorageClass parameter and if ceph-csi needs to serve the non-DH-CHAP volumes?

internal/nvmeof/nodeserver/nodeserver.go
```diff
@@ -716,6 +741,10 @@ func (ns *NodeServer) connectToSubsystem(
 
 	// Setup DH-CHAP authentication if required
 	if info.DhchapMode != nvmeof.DHCHAPEmpty && info.DhchapMode != nvmeof.DHCHAPModeNone {
+		if ns.dhchapUnsupportedKernelErr != nil {
+			return "", ns.dhchapUnsupportedKernelErr
```
iPraveenParihar, line 750: This error propagates to NodeStageVolume, where its wrapped with `code.Internal` and attacher will keep retrying which makes no-sense to retry if the kernel is unsupported

iPraveenParihar (changes requested):

internal/nvmeof/nodeserver/nodeserver.go
nixpanic: well, if there is a problem getting the kernel version, something is really broken. Getting the kernel version should never fail.

internal/nvmeof/nodeserver/nodeserver.go
nixpanic: That is true... It is better than what happens now, some error while executing the `nvme` command.

I'll see if I can come up with something simple that returns InvalidArgument for this case.

gadididi (approved):

internal/nvmeof/nodeserver/nodeserver.go
nixpanic: @iPraveenParihar , added a commit for improving this behavior. Please have a look, thanks!

nixpanic:
@iPraveenParihar, are you ok with this change now?

iPraveenParihar (approved):
LGTM

iPraveenParihar:
wait for https://github.com/ceph/ceph-csi/pull/6319 to be merged

iPraveenParihar:
@Mergifyio rebase

iPraveenParihar:
@Mergifyio queue

iPraveenParihar:
Would need approval from @Rakshith-R or @Madhu-1

Madhu-1 (approved):

PR #6304, opened 2026-05-29
cleanup: use CSI Server interfaces (CephFS & NFS)

nixpanic:
Rename the different gRPC server structs to include a prefix of the storage protocol they use. This makes it easier to understand the code, as having a struct `NodeServer` for each protocol causes confusion.

The `NewNodeServer()` and similar functions now return a CSI interface type, making sure that the implementation of the structs is not available to external consumers.

 - All server types now follow Go naming conventions (unexported, driver-prefixed)
 - All types have compile-time interface assertions
 - All constructors return CSI interface types instead of concrete types 

**Note:** RBD and NVMe-oF changes will be sent in a separate PR.

---

<details>

</details>

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -42,9 +42,8 @@ import (
 	rterrors "github.com/ceph/ceph-csi/internal/util/reftracker/errors"
 )
 
-// ControllerServer struct of CEPH CSI driver with supported methods of CSI
-// controller server spec.
-type ControllerServer struct {
+// cephfsControllerServer implements the CSI controller server for the CephFS driver.
+type cephfsControllerServer struct {
 	*csicommon.DefaultControllerServer
 	// A map storing all volumes with ongoing operations so that additional operations
 	// for that same volume (as defined by VolumeID/volume name) return an Aborted error
@@ -65,6 +64,12 @@ type ControllerServer struct {
 	ClusterName string
 }
 
+// Assert required implementation of CSI interfaces.
+var (
+	_ csi.ControllerServer      = &cephfsControllerServer{}
+	_ csi.GroupControllerServer = &cephfsControllerServer{}
+)
+
 // subvolumeMetadataHandler holds the metadata handling functions.
 type subvolumeMetadataHandler struct {
 	logSubvolumeName     string
@@ -73,7 +78,7 @@ type subvolumeMetadataHandler struct {
 }
 
 // createBackingVolume creates the backing subvolume and on any error cleans up any created entities.
-func (cs *ControllerServer) createBackingVolume(
+func (cs *cephfsControllerServer) createBackingVolume(
 	ctx context.Context,
 	volOptions,
 	parentVolOpt *store.VolumeOptions,
@@ -112,7 +117,7 @@ func (cs *ControllerServer) createBackingVolume(
 	return nil
 }
 
-func (cs *ControllerServer) createBackingVolumeFromSnapshotSource(
+func (cs *cephfsControllerServer) createBackingVolumeFromSnapshotSource(
 	ctx context.Context,
 	volOptions *store.VolumeOptions,
 	parentVolOpt *store.VolumeOptions,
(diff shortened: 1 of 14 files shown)


nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio rebase

black-dragon74 (dismissed):

internal/cephfs/controllerserver.go
```diff
@@ -42,9 +42,8 @@ import (
 	rterrors "github.com/ceph/ceph-csi/internal/util/reftracker/errors"
 )
 
-// ControllerServer struct of CEPH CSI driver with supported methods of CSI
-// controller server spec.
-type ControllerServer struct {
+// cephfsControllerServer implements the CSI controller server for the CephFS driver.
+type cephfsControllerServer struct {
```
Madhu-1, line 46: the package name is already cephfs, can we call this as `controllerServer` having cephfsControllerServer inside cephfs package seems odd, WDYT? same is applicalbe for other types in different packages as well

nixpanic: I considered that, but having multiple `controllerServer` structs does not make it much easier to understand the code. The main objective of the PR is to make maintenance easier, and giving each ControllerServer their unique name/type helps with that.

Madhu-1: There will be one controllerServer per package, where we are going to have multiple controllerServer per package? did i miss something? something like as per standards https://google.github.io/styleguide/go/best-practices.html#naming

nixpanic: No, always a single controller-server per package.

We're not following that guideline strictly, we have many more structs that have the same prefix as their (parent)package. I like the guideline in general, but it does make the code more difficult to follow when reading/tracing/logging it.

Madhu-1 (dismissed):

internal/cephfs/controllerserver.go
Madhu-1: it would be nice to follow guidelines and make changes in required places instead of following the same pattern again.

nixpanic:
@mergifyio rebase

internal/cephfs/driver.go
```diff
@@ -201,7 +203,7 @@ func (fs *cephfsDriver) Run(conf *util.Config) {
 		IS: fs.is,
 		CS: fs.cs,
 		NS: fs.ns,
-		GS: fs.cs,
+		GS: csicommon.ToGroupControllerServer(fs.cs),
```
nixpanic, line 204: In case of the NodePlugin, `fs.cs` is `nil`, which caused a bug here.

nixpanic: `csicommon.ToGroupControllerServer(nil)` now returns `nil` too.

black-dragon74 (dismissed):

internal/cephfs/driver.go
```diff
@@ -174,8 +175,8 @@ func (fs *cephfsDriver) Run(conf *util.Config) {
 	}
 
 	if conf.IsControllerServer {
-		fs.cs = NewControllerServer(fs.cd)
-		fs.cs.ClusterName = conf.ClusterName
+		cs := NewControllerServer(fs.cd, conf.ClusterName)
```
iPraveenParihar:
```suggestion
		fs.cs := NewControllerServer(fs.cd, conf.ClusterName)
```

iPraveenParihar (commented):
nit,
rest LGTM

internal/cephfs/driver.go
nixpanic: thanks, fixed.

black-dragon74 (dismissed):

internal/cephfs/driver.go
```diff
@@ -187,7 +187,8 @@ func (fs *cephfsDriver) Run(conf *util.Config) {
 			conf.KernelMountOptions, conf.FuseMountOptions,
 			nodeLabels, topology, crushLocationMap,
 		)
-		fs.cs = NewControllerServer(fs.cd)
+		cs := NewControllerServer(fs.cd, "")
+		fs.cs = cs
```
Rakshith-R:
```suggestion
		fs.cs := NewControllerServer(fs.cd, "")
```

Rakshith-R (commented):
just a tiny nit

internal/cephfs/driver.go
nixpanic: fixed, thanks!

black-dragon74 (approved):

Rakshith-R (approved):

nixpanic:
@mergifyio rebase

PR #6306, opened 2026-06-01
rbd: open image read-only for DiffIterate in getUsedBytes

Rakshith-R:
getUsedBytes() only reads from the RBD image via DiffIterate to calculate volume usage statistics for NodeGetVolumeStats. Opening the image in read-write mode unnecessarily acquires the exclusive lock, causing contention when the image is mapped by a workload.

Switch to OpenImageReadOnly which avoids the exclusive lock and allows stats collection even when the image is actively mapped elsewhere.

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -604,6 +604,27 @@ func (ri *rbdImage) open() (*librbd.Image, error) {
 	return image, nil
 }
 
+// openReadOnly opens the rbdImage in read-only mode. This should be used when
+// the caller only needs to read from the image (e.g. DiffIterate for usage
+// statistics), as it avoids any interaction with the exclusive lock on the image.
+func (ri *rbdImage) openReadOnly() (*librbd.Image, error) {
+	err := ri.openIoctx()
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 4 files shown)


internal/rbd/rbd_util.go
```diff
@@ -604,6 +604,27 @@ func (ri *rbdImage) open() (*librbd.Image, error) {
 	return image, nil
 }
 
+// openReadOnly opens the rbdImage in read-only mode. This should be used when
+// the caller only needs to read from the image (e.g. DiffIterate for usage
+// statistics), as it avoids acquiring the exclusive lock on the image.
```
idryomov: Nit: `librbd.OpenImage()` doesn't acquire exclusive lock either -- it just loads the required metadata and allows the client to potentially (attempt to) acquire it in the future:
```suggestion
// statistics), as it avoids any interaction with the exclusive lock on the image.
```

idryomov (approved):
Ack from the RBD perspective

Madhu-1 (approved):

Madhu-1:
@mergifyio rebase

nixpanic (approved):

nixpanic:
@mergifyio rebase

PR #6277, opened 2026-05-14
nvmeof: add volume cloning capability

gadididi:
enable volume cloning support for nvmeof CSI driver by declaring CLONE_VOLUME capability. The RBD backend automatically handles all clone operations (snapshot-to-volume restore and volume-to-volume cloning) when VolumeContentSource is present in CreateVolume request.

The NVMeoF driver adds debug logging to track clone operations and creates namespaces pointing to the cloned RBD images. No changes to the NVMeoF gateway are required since clones are transparent at the block device level.

**Need to test it!** 

**Checklist:**

* [x] **Commit Message Formatting**: Commit titles and messages follow
  guidelines in the [developer
  guide](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#commit-messages).
* [ ] Reviewed the developer guide on [Submitting a Pull
  Request](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#development-workflow)
* [ ] [Pending release
  notes](https://github.com/ceph/ceph-csi/blob/devel/PendingReleaseNotes.md)
  updated with breaking and/or notable changes for the next major release.
* [ ] Documentation has been updated, if necessary.
* [ ] Unit tests have been added, if necessary.
* [ ] Integration tests have been added, if necessary.

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

* `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

PR #6322, opened 2026-06-04
Prevent concurrent DeleteVolume/DeleteSnapshot during CreateVolume 

### LINKED ISSUE #6321: Race Condition During Concurrent Clone and Delete (CephFS & RBD)
The below comes from a **Copilot** suggestion, it applies to CephFS and RBD as well.

#### Scenario:

- PVC-A (vol-123) exists and is mounted
- User creates PVC-B as clone of PVC-A
- User deletes PVC-A while clone is in progress

#### Without Source Volume Lock:

Thread A: CreateVolume(clone from vol-123)

- Locks: "pvc-b-name" (destination)
- Calls RBD CreateVolume (starts cloning)

Thread B: DeleteVolume(vol-123) - runs concurrently

- Locks: "vol-123" (different lock - both threads proceed!)
- Deletes NVMe-oF namespace → success
- Tries to delete RBD image → FAILS (has clones, RBD blocks it)
- Returns error

**Result: vol-123 has RBD image but NO namespace - PVC-A cannot be mounted until retry succeeds.**

#### Option 1: Add Source Volume Lock (prevents race)
```Go
// In CreateVolume, after extracting clone source:
if vol := contentSource.GetVolume(); vol != nil {
    sourceVolumeID := vol.GetVolumeId()
    if acquired := cs.volumeLocks.TryAcquire(sourceVolumeID); !acquired {
        return nil, status.Errorf(codes.Aborted, "source volume operation in progress")
    }
}
```

- Pros: No broken state window, cleaner user experience
- Cons: Additional lock complexity, serializes operations

#### Option 2: Leave As-Is (rely on idempotency)

- DeleteNamespace is idempotent (returns nil on ENODEV)
- DeleteVolume will be retried by Kubernetes
- Eventually succeeds after clones are deleted
- Pros: Simpler code, no additional locks
- Cons: Temporary unpublishable state, more error logs during retries

Both approaches are valid - lock prevents the issue, idempotency recovers from it.

_Originally posted by @gadididi in https://github.com/ceph/ceph-csi/pull/6277#discussion_r3254221879_

nixpanic:

There is a race condition possible when `CreateVolume` uses a source volume/snapshot and the source is deleted at the same time. The creation of the volume may fail in weird ways. By grabbing a lock for the source volume/snapshot, a concurrent `DeleteVolume` procedure has to wait until the `CreateVolume` procedure has finished.

## Related issues ##

Fixes: #6321

Note: NFS does not use the source volume/snapshot in any way, it forwards the `CreateVolume` call on to the CephFS Controller. There is no need for the added locking in NFS. NVMe-oF only gets the Clone/Restore functionality through the work-in-progress #6277.

---

<details>

</details>

### DIFF
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -344,6 +344,24 @@ func (cs *cephfsControllerServer) CreateVolume(
 		defer parentVol.Destroy()
 	}
 
+	// Lock the source volume or snapshot to prevent deletion while creating from it
+	if pvID != nil {
+		if acquired := cs.VolumeLocks.TryAcquire(pvID.VolumeID); !acquired {
+			log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, pvID.VolumeID)
+
+			return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, pvID.VolumeID)
+		}
+		defer cs.VolumeLocks.Release(pvID.VolumeID)
+	}
+	if sID != nil {
+		if acquired := cs.SnapshotLocks.TryAcquire(sID.SnapshotID); !acquired {
+			log.ErrorLog(ctx, util.SnapshotOperationAlreadyExistsFmt, sID.SnapshotID)
+
+			return nil, status.Errorf(codes.Aborted, util.SnapshotOperationAlreadyExistsFmt, sID.SnapshotID)
+		}
+		defer cs.SnapshotLocks.Release(sID.SnapshotID)
+	}
+
 	err = checkValidCreateVolumeRequest(volOptions, parentVol, pvID, sID, req)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -394,6 +394,8 @@ func checkValidCreateVolumeRequest(rbdVol, parentVol *rbdVolume, rbdSnap *rbdSna
 }
 
 // CreateVolume creates the volume in backend.
+//
+//nolint:gocyclo,cyclop // TODO: reduce complexity.
 func (cs *ControllerServer) CreateVolume(
 	ctx context.Context,
 	req *csi.CreateVolumeRequest,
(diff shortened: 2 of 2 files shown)


internal/rbd/controllerserver.go
```diff
@@ -394,6 +396,24 @@ func (cs *ControllerServer) CreateVolume(
...
+	if parentVol != nil {
+		if acquired := cs.VolumeLocks.TryAcquire(parentVol.VolID); !acquired {
+			log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, parentVol.VolID)
+
+			return nil, status.Errorf(codes.Aborted, util.VolumeOperationAlreadyExistsFmt, parentVol.VolID)
+		}
+		defer cs.VolumeLocks.Release(parentVol.VolID)
+	}
+	if rbdSnap != nil {
+		if acquired := cs.SnapshotLocks.TryAcquire(rbdSnap.VolID); !acquired {
```
iPraveenParihar, line 450: CreateVolumeFromSnapshot will fail with `operation already exists` since we already have locks there?
https://github.com/ceph/ceph-csi/blob/9d08300466de72d7b99dffd81e9c23883b1cf87b/internal/rbd/controllerserver.go#L679-L691

nixpanic: Good catch, I would expect the same indeed too. Will check it in more detail and correct it later.

nixpanic: moved the locking all together, easier to follow the working that way.

YiteGu: I've been paying particular attention to this lock lately because it impacts the performance of concurrent restore-pvc create.

YiteGu: The fact that `rbdSnap != nil` is sufficient to indicate that this is a restore request, and we can directly use `cs.OperationLocks.GetRestoreLock` without needing `cs.SnapshotLocks.TryAcquire`.

internal/cephfs/controllerserver.go
```diff
@@ -339,6 +339,24 @@ func (cs *ControllerServer) CreateVolume(
 		defer parentVol.Destroy()
 	}
 
+	// Lock the source volume or snapshot to prevent concurrent operations
```
black-dragon74:
```suggestion
	// Lock the source volume or snapshot to prevent deletion while creating from it
```

one small nit, rest looks good to me :)

internal/rbd/controllerserver.go
nixpanic: thanks @YiteGu for confirming, that is what this PR implements.

internal/cephfs/controllerserver.go
nixpanic: addressed this, thanks!

black-dragon74 (dismissed):
_GH is still processing the updates_

black-dragon74 (approved):

internal/rbd/controllerserver.go
YiteGu: @nixpanic Will using SnapshotLocks here affect the performance of concurrent PVC creation using the same snapshot? Is it okay to directly use `GetRestoreLock(rbdSnap.VolID)`?

iPraveenParihar (approved):

iPraveenParihar:
@mergifyio rebase

iPraveenParihar:
@mergifyio queue

iPraveenParihar:
@mergifyio rebase

nixpanic:
Note that #6277 is already running CI jobs.

Rakshith-R (changes requested):
Doesn't this also block concurrent create ops on same parent ? 
and therefore slow create op.

Rakshith-R:
>> Doesn't this also block concurrent create ops on same parent ? and therefore slow create op.

> Option 2: Leave As-Is (rely on idempotency)

ext-provisioner and ext-snapshotter are responsible for adding finalizer on source pvc/snap.

> PVC-A (vol-123) exists and is mounted
> User creates PVC-B as clone of PVC-A
> User deletes PVC-A while clone is in progress

Delete should get blocked since the pvc is still mounted.
I don't think this a valid scenario.

PR #6353, opened 2026-06-23
journal: add multi-shard support for csiDirectory

YiteGu:
CSI request-name mappings for volume and snapshot journals are currently stored in a single csiDirectory omap object. In Ceph, the osd_deep_scrub_large_omap_object_key_threshold setting triggers a ceph cluster-side warning when a single object contains more than 200,000 omap keys. With the current business growth supported by ceph-csi, clusters with more than 200,000 RBD volumes have already appeared, which means the existing single-object layout is no longer sufficient. Concentrating all lookups, updates, and cleanup operations on one oid also creates a metadata hotspot and makes the directory harder to scale.

This change introduces a hash-based multi-shard layout for csiDirectory so that new request-name mappings are distributed across multiple shard oids. The existing request-name to UUID mapping model remains unchanged, while omap operations are spread by key.

Compatibility with the legacy csiDirectory layout is preserved to support upgrades:
- CheckReservation and CheckNewUUIDMapping read from the shard oid first and fall back to the legacy oid on lookup miss
- ReserveName and ReserveNewUUIDMapping write new mappings to the shard oid
- UndoReservation removes mappings from the shard oid and also cleans up the legacy oid so that pre-upgrade metadata can still be reclaimed correctly

This enables sharded csiDirectory metadata for new entries without breaking existing volume and snapshot journal data during rolling upgrades.

Currently, the default number of shards is set to 11 and cannot be dynamically modified. This is sufficient for future use.

---

<details>

</details>

### DIFF
--- a/internal/journal/omap.go
+++ b/internal/journal/omap.go
@@ -64,7 +64,7 @@ func getOMapValuesByKeys(
 			log.ErrorLog(ctx, "omap not found (pool=%q, namespace=%q, name=%q): %v",
 				poolName, namespace, oid, err)
 
-			return nil, fmt.Errorf("%w: %w", util.ErrKeyNotFound, err)
+			return nil, errors.Join(util.ErrObjectNotFound, util.ErrKeyNotFound, err)
 		}
 
 		return nil, err
--- a/internal/journal/voljournal.go
+++ b/internal/journal/voljournal.go
@@ -22,7 +22,9 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"strings"
+	"sync"
 
 	"github.com/google/uuid"
 
@@ -110,13 +112,18 @@ single entity modifying the related omaps for a given VolName.
 const (
 	defaultVolumeNamingPrefix   string = "csi-vol-"
 	defaultSnapshotNamingPrefix string = "csi-snap-"
+	defaultCSIDirectoryShards   uint32 = 11
 )
 
 // CSIJournal defines the interface and the required key names for the above RADOS based OMaps.
 type Config struct {
-	// csiDirectory is the name of the CSI volumes object map that contains CSI volume-name (or
-	// snapshot name) based keys
+	// csiDirectory is the legacy base name of the CSI volumes object map that contains CSI
+	// volume-name (or snapshot name) based keys.
 	csiDirectory string
+
+	// csiDirectoryShards is the number of sharded directory objects used for new CSI name lookups.
+	// The legacy csiDirectory object remains as a compatibility fallback during upgrades.
+	csiDirectoryShards uint32
 
 	// CSI volume-name keyname prefix, for key in csiDirectory, suffix is the CSI passed volume name
 	csiNameKeyPrefix string
@@ -165,12 +172,18 @@ type Config struct {
 
 	// commonPrefix is the prefix common to all omap keys for this Config
 	commonPrefix string
+
+	// legacyDirectoryAbsent tracks csiDirectory oids that have been confirmed
+	// absent for this controller instance, so later connections can skip legacy
+	// fallback for the same journal directory.
+	legacyDirectoryAbsent *sync.Map
 }
 
 // NewCSIVolumeJournal returns an instance of CSIJournal for volumes.
 func NewCSIVolumeJournal(suffix string) *Config {
 	return &Config{
 		csiDirectory:            "csi.volumes." + suffix,
+		csiDirectoryShards:      defaultCSIDirectoryShards,
 		csiNameKeyPrefix:        "csi.volume.",
 		cephUUIDDirectoryPrefix: "csi.volume.",
 		csiNameKey:              "csi.volname",
@@ -185,6 +198,7 @@ func NewCSIVolumeJournal(suffix string) *Config {
 		ownerKey:                "csi.volume.owner",
 		backingSnapshotIDKey:    "csi.volume.backingsnapshotid",
 		commonPrefix:            "csi.",
+		legacyDirectoryAbsent:   &sync.Map{},
 	}
 }
 
@@ -192,6 +206,7 @@ func NewCSIVolumeJournal(suffix string) *Config {
 func NewCSISnapshotJournal(suffix string) *Config {
 	return &Config{
 		csiDirectory:            "csi.snaps." + suffix,
+		csiDirectoryShards:      defaultCSIDirectoryShards,
 		csiNameKeyPrefix:        "csi.snap.",
 		cephUUIDDirectoryPrefix: "csi.snap.",
 		csiNameKey:              "csi.snapname",
@@ -205,6 +220,7 @@ func NewCSISnapshotJournal(suffix string) *Config {
 		encryptionType:          "csi.volume.encryptionType",
 		ownerKey:                "csi.volume.owner",
 		commonPrefix:            "csi.",
+		legacyDirectoryAbsent:   &sync.Map{},
 	}
 }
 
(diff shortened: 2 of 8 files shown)


nixpanic:
The changes look reasonable, there are two things:

1. please add a comment for new functions you introduce
2. keep the structure in sync for volume groups too (add similar sharding in `internal/journal/volumegroupjournal.go`)

YiteGu:
> The changes look reasonable, there are two things:
> 
> 1. please add a comment for new functions you introduce
> 2. keep the structure in sync for volume groups too (add similar sharding in `internal/journal/volumegroupjournal.go`)

updated

YiteGu:
@nixpanic CI task ([multi-arch-build / multi-arch-build (pull_request)](https://github.com/ceph/ceph-csi/actions/runs/28014340131/job/82914842591?pr=6353)) seemingly trapped in an infinite loop.

nixpanic (dismissed):
Thanks, looks good to me.

internal/journal/voljournal.go
```diff
@@ -475,8 +532,14 @@ func (conn *Connection) UndoReservation(ctx context.Context,
...
 	// delete the request name key (last, inverse of create order)
-	err := removeMapKeys(ctx, conn, csiJournalPool, cj.namespace, cj.csiDirectory,
-		[]string{cj.csiNameKeyPrefix + reqName})
+	key := cj.csiNameKeyPrefix + reqName
+	shardOid := cj.getCsiDirectoryShard(key)
+	err := removeMapKeys(ctx, conn, csiJournalPool, cj.namespace, shardOid, []string{key})
+	if err == nil {
+		// Always delete from the legacy oid as well, to ensure metadata for
+		// already existing volumes gets cleaned up.
+		err = removeMapKeys(ctx, conn, csiJournalPool, cj.namespace, cj.csiDirectory, []string{key})
```
iPraveenParihar, line 591: If we want to remove legacy key, then why do conditionally when `err == nil`?

internal/journal/voljournal.go
```diff
@@ -312,12 +374,8 @@ func (conn *Connection) CheckReservation(ctx context.Context,
...
 
 	// check if request name is already part of the directory omap
-	fetchKeys := []string{
-		cj.csiNameKeyPrefix + reqName,
-	}
-	values, err := getOMapValuesByKeys(
-		ctx, conn, journalPool, cj.namespace, cj.csiDirectory,
-		fetchKeys)
+	key := cj.csiNameKeyPrefix + reqName
+	objUUIDAndPool, found, err := lookupCsiDirectoryValue(ctx, conn, journalPool, key)
```
iPraveenParihar, line 428: For new deployments, this will be 2x metadata reads (shard and legacy) instead of 1.
At scale,  this doubles the metadata read load for the creation path. Can we consider to skip legacy fallback for new deployment that we know for sure will be in shard directory?

internal/journal/voljournal.go
```diff
@@ -275,6 +283,60 @@ func (cj *Config) Connect(monitors, namespace string, cr *util.Credentials) (*Co
...
+	if err == nil {
+		if value, found := values[key]; found {
+			return value, true, nil
+		}
+
+		log.DebugLog(ctx, "omap not found in shard csiDirectory %q, falling back to legacy oid %q",
+			shardOID, cj.csiDirectory)
+
+		values, err = getOMapValuesByKeys(ctx, conn, journalPool, cj.namespace, cj.csiDirectory, []string{key})
+	} else if errors.Is(err, util.ErrKeyNotFound) {
```
iPraveenParihar: Should be fine to fallback to other errors as well?

iPraveenParihar (changes requested):
@YiteGu IIUC, this will be forward-only upgrade (3.17 -> 3.18) and downgrade is not possible? Since entries written to shard OIDs are invisible to older ceph-csi versions that only read the legacy directory. We should document this migration and downsides if any that users need to know of during upgrade/downgrade.

@YiteGu could you please update pendingReleaseNotes.md?

@Madhu-1 @nixpanic @Rakshith-R WDYT about downgrade?

internal/journal/voljournal.go
YiteGu: I think that the legacy cleanup will only be skipped if a genuine deletion error occurs. If a deletion error occurs, we should immediately return error to avoid continuing and to cover up the error.

internal/journal/voljournal.go
YiteGu: I think taht ErrKeyNotFound indicates shard oid does not exist, and a fallback should be performed. Other I/O operation errors should return an error.

internal/journal/voljournal.go
YiteGu: Yes, Regarding performance issues, currently use by Key lookup, and I think the performance is sufficient.

One solution I can think of right now is to treat the legacy oid as a new ceph-csi if it cannot be found on the first read, so that subsequent requests do not need to perform a fallback operation. how do you think?

YiteGu: It is now possible to skip reading legacy csiDirectory in new deployments. For example:
I use a new storageclass with new ceph pool to create pvc. the first create
```
 ID: 18 Req-ID: pvc-c1be947c-8fea-45ed-b81e-8f4cb3a2f060 omap not found (pool="replicapool-ssd2", namespace="", name="csi.volumes.default.5"): rados: ret=-2, No such file or directory
 ID: 18 Req-ID: pvc-c1be947c-8fea-45ed-b81e-8f4cb3a2f060 failed reading shard csiDirectory "csi.volumes.default.5" with ErrKeyNotFound
 ID: 18 Req-ID: pvc-c1be947c-8fea-45ed-b81e-8f4cb3a2f060 omap not found in shard csiDirectory "csi.volumes.default.5", falling back to legacy oid "csi.volumes.default"
 ID: 18 Req-ID: pvc-c1be947c-8fea-45ed-b81e-8f4cb3a2f060 omap not found (pool="replicapool-ssd2", namespace="", name="csi.volumes.default"): rados: ret=-2, No such file or directory
 ID: 18 Req-ID: pvc-c1be947c-8fea-45ed-b81e-8f4cb3a2f060 legacy csiDirectory "csi.volumes.default" is absent; future lookups will skip the fallback
```
subsequent create pvc
```
ID: 20 Req-ID: pvc-dca06f47-e3b7-4875-b141-ea09df5c035e omap not found (pool="replicapool-ssd2", namespace="", name="csi.volumes.default.0"): rados: ret=-2, No such file or directory
ID: 20 Req-ID: pvc-dca06f47-e3b7-4875-b141-ea09df5c035e failed reading shard csiDirectory "csi.volumes.default.0" with ErrKeyNotFound
ID: 20 Req-ID: pvc-dca06f47-e3b7-4875-b141-ea09df5c035e skipping legacy csiDirectory lookup for "csi.volumes.default" after confirming the oid is absent
```

YiteGu:
> @YiteGu IIUC, this will be forward-only upgrade (3.17 -> 3.18) and downgrade is not possible? Since entries written to shard OIDs are invisible to older ceph-csi versions that only read the legacy directory. We should document this migration and downsides if any that users need to know of during upgrade/downgrade.

Yes, that is correct.

This change is backward-compatible for upgrades because the new code can still read legacy csiDirectory entries, but it is not downgrade-compatible once new mappings are written to shard oids. Older ceph-csi versions only read the legacy csiDirectory and will not see entries created after the upgrade.

> 
> @YiteGu could you please update pendingReleaseNotes.md?
> 

Updated.

nixpanic (dismissed):

iPraveenParihar (commented):
LGTM!
@Rakshith-R / @Madhu-1  Can you please provide your review on it?

Madhu-1 (approved):
We need changes in other repo's whereever we are dependent on the omap.

Madhu-1:
Dont we need E2E updates for new structure in omap now? @YiteGu

Madhu-1 (changes requested):
Marking changes for below update
1. update the E2E 
2.  docs/resource-cleanup.md doc 
3. https://github.com/ceph/ceph-csi/tree/devel/troubleshooting/tools

nixpanic:
> We need changes in other repo's whereever we are dependent on the omap.

What other repositories are there that use the internal OMAP? If there is a need for that, we surely need to provide a document with the structure, or even an API :thinking:

Madhu-1:
> at other repositories are there that use the internal OMAP? If there is a need for that, we surely need to provide a document with the structure, or even an API 🤔

rook-ceph krew-plugin is using it

YiteGu:
> Dont we need E2E updates for new structure in omap now? @YiteGu

Get, I'm Processing

YiteGu:
> Marking changes for below update
> 
> 1. update the E2E
> 2. docs/resource-cleanup.md doc
> 3. https://github.com/ceph/ceph-csi/tree/devel/troubleshooting/tools

Updated @Madhu-1 @nixpanic

YiteGu:
@Madhu-1 The e2e failure appears to be due to a nfs pvc request timeout. I'll run it again to see if it fails again.

YiteGu:
> /test ci/centos/mini-e2e/k8s-1.35

test pass

troubleshooting/tools/tracevol.py
```diff
@@ -643,15 +623,43 @@ def get_val_from_omap(obj, key, pool, is_rbd, regex=br'[A-Za-z0-9\-]+'):
...
+    if val:
+        return val
+
+    return get_val_from_omap(base_oid, key, pool, is_rbd)
+
 def check_snap_content_name_in_rados(snap_uuid, snap_content_name, pool, is_rbd):
     """
-    Validates if snapshot content name is listed in csi.snaps.default omap
+    Validates if snapshot content name is listed in csi.snaps omap
     """
```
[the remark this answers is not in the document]
YiteGu: done

internal/journal/voljournal.go
```diff
@@ -275,6 +291,102 @@ func (cj *Config) Connect(monitors, namespace string, cr *util.Credentials) (*Co
...
+
+	if cj.isLegacyDirectoryAbsent(journalPool, cj.namespace, cj.csiDirectory) {
+		log.DebugLog(ctx, "skipping legacy csiDirectory lookup for %q after confirming the oid is absent",
+			cj.csiDirectory)
+
+		return "", false, nil
+	}
+
+	log.DebugLog(ctx, "omap not found in shard csiDirectory %q, falling back to legacy oid %q",
+		shardOID, cj.csiDirectory)
```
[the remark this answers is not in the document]
YiteGu: done

internal/journal/omap.go
```diff
@@ -64,7 +64,7 @@ func getOMapValuesByKeys(
 			log.ErrorLog(ctx, "omap not found (pool=%q, namespace=%q, name=%q): %v",
 				poolName, namespace, oid, err)
```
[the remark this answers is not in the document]
YiteGu: no need change

Madhu-1 (approved):
@subhamkrai we need to update the rook-ceph tool after this change where it should take care of old and newly created pvc.

subhamkrai:
> @subhamkrai we need to update the rook-ceph tool after this change where it should take care of old and newly created pvc.

@Madhu-1 We'll need to update the rook-ceph plugin only, right? No change in Rook-Ceph?

Madhu-1:
> > @subhamkrai we need to update the rook-ceph tool after this change where it should take care of old and newly created pvc.
> 
> @Madhu-1 We'll need to update the rook-ceph plugin only, right? No change in Rook-Ceph?

only plugin for now

Madhu-1:
@nixpanic @Rakshith-R we should target this to our 3.20 release or we can go with 4.x as no downgrade is supported

nixpanic:
> @nixpanic @Rakshith-R we should target this to our 3.20 release or we can go with 4.x as no downgrade is supported

Not sure. I don't think we make sure downgrades work at all (it is mostly just coincidence)? This is backwards compatible, old journal entries continue working.

PR #6357, opened 2026-06-23
build: add verbose build output in CI environments

nixpanic:
Add GO_BUILD_VERBOSE variable that enables verbose go build output
(-v flag) when the CI environment variable is set. This helps diagnose
build issues in CI pipelines, particularly in multi-architecture builds.

The verbose output shows package compilation progress without affecting
local development builds which remain quiet by default.

This is most useful for the **multi-arch-build** job, it may show why
the job often hangs (or is just very slow).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,13 @@ LDFLAGS += -X $(GO_PROJECT)/internal/util.GitCommit=$(GIT_COMMIT)
 LDFLAGS += -X $(GO_PROJECT)/internal/util.DriverVersion=$(CSI_IMAGE_VERSION)
 GO_TAGS ?= -tags=$(shell echo $(GO_TAGS_LIST) | tr ' ' ',')
 
+# Enable verbose build output in CI environments
+ifdef CI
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -258,7 +265,7 @@ endif
 
 image-cephcsi: GOARCH ?= $(shell go env GOARCH 2>/dev/null)
 image-cephcsi: .container-cmd
```
iPraveenParihar, line 267: @nixpanic Will this work for containerized-build?

nixpanic: No, only for for `make image-cephcsi` as that is used by **multi-arch-build**. 

Example output is at https://github.com/ceph/ceph-csi/actions/runs/28036473946/job/82991083194?pr=6357#step:5:941

iPraveenParihar (approved):

Madhu-1 (approved):

nixpanic:
Added `DNM` label to prevent merging while other PRs are still running CI jobs (and have to re-run their full CI again).

PR #6358, opened 2026-06-24
cephfs: add namespace isolated subvolume creation

### LINKED ISSUE #6267: Support rados namespace isolation for CephFS subvolumes via radosPoolNamespace StorageClass parameter

  Ceph already supports per-subvolume rados namespace isolation via --namespace-isolated:

  $ ceph fs subvolume create ceph-filesystem <name> --group-name tenant-a --namespace-isolated
  $ ceph fs subvolume info ceph-filesystem <name> --group-name tenant-a | grep pool_namespace
      "pool_namespace": "fsvolumens_<name>"

  This isolates all RADOS objects for that subvolume into a dedicated namespace within the data pool, providing strong data segregation without requiring separate pools.

I want to have multitenancy in the fs from isolating the user from client layer, to where there data gets stored should be logically seperated (rados namespace)

  Use case / motivation

  In multi-tenant Kubernetes clusters using shared CephFS:
  - Subvolume groups provide path-level isolation (/volumes/tenant-a/)
  - But all tenants' RADOS objects land in the same data pool namespace
  - Rados namespace isolation would allow per-tenant pool-level ACLs, quotas, and billing via ceph osd pool stats <pool>/<namespace>
  - It also enables future capability: revoking a tenant's CephX key scoped to their rados namespace without affecting others
pass --namespace-isolated when calling SubvolumeCreate, and store/use the namespace name for mount operations:
```
  # StorageClass
  parameters:
    clusterID: <subvolumegroup-clusterID>
    fsName: ceph-filesystem
    radosNamespace: tenant-a   # currently no-op for CephFS — should trigger --namespace-isolated preferably with namespace of the rados pool
```
Add a list of criteria that should be met for this feature to be useful

# Additional context #

https://docs.ceph.com/en/reef/cephfs/fs-volumes/#creating-a-subvolume
https://docs.ceph.com/en/latest/radosgw/pools/#pool-namespaces

vt0r:

Adding support for creating CephFS subvolumes with `go-ceph`'s `namespace_isolated` flag enabled at creation time (configured using the new `namespaceIsolated` parameter in the StorageClass), which will create each new subvolume within its own unique, isolated RADOS namespace.

> Do you have any questions?

None at the moment.

> Is the change backward compatible?

Yes. Tested heavily against Ceph 18.x+ and Ceph-CSI versions even older than the supported ones.

> Are there concerns around backward compatibility?

None from my side.

> Provide any external context for the change, if any.

This flag has been supported [in Ceph](https://github.com/ceph/ceph/pull/34776/changes) and [in `go-ceph`](https://github.com/ceph/go-ceph/pull/342/changes#diff-57b7fbe7bf4be97ff70b2f611837613c3b964cb23a4d4df97e0a628fe056c069R20) for quite a while (since Ceph `v16.1.0` and `go-ceph` `v0.6.0` it looks like), so it's a relatively simple wiring job, and it should be fully backwards compatible across all supported Ceph/Ceph-CSI releases.

## Related issues ##

Fixes: #6267

## Future concerns ##

None that I'm aware of.

**Checklist:**

  guidelines in the [developer
* [x] [Pending release

---

<details>

</details>

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -4,4 +4,10 @@
 
 ## Features
 
+### CephFS
+
+* **Namespace isolated subvolumes:** Added an optional StorageClass parameter
+  (`namespaceIsolated`, default disabled) that allows creating new subvolumes
+  in a unique, isolated RADOS namespace
+
 ## NOTE
--- a/docs/cephfs/deploy.md
+++ b/docs/cephfs/deploy.md
@@ -82,6 +82,7 @@ you're running it inside a k8s cluster and find the config itself).
 | `encrypted`                                                                                         | no             | disabled by default, use `"true"` to enable fscrypt encryption on PVC and `"false"` to disable it. **Do not change for existing storageclasses**                                                                          |
 | `encryptionKMSID`                                                                                   | no             | required if encryption is enabled and a kms is used to store passphrases                                                                                                                                                |
 | `extraDeploy` | no | array of extra objects to deploy with the release |
+| `namespaceIsolated` | no | Boolean value (truthy/falsy string), disabled by default - set to `"true"` in your Kubernetes CephFS StorageClass to create each new subvolume in a unique, isolated RADOS namespace. |
 
 **NOTE:** An accompanying CSI configuration file, needs to be provided to the
 running pods. Refer to [Creating CSI configuration](../../examples/README.md#creating-csi-configuration)
--- a/examples/cephfs/storageclass.yaml
+++ b/examples/cephfs/storageclass.yaml
@@ -64,6 +64,14 @@ parameters:
   # correlation to configmap entry.
   # encryptionKMSID: <kms-config-id>
 
+  # (optional) Enable RADOS namespace isolation for new CephFS subvolumes
+  #   Type: Boolean (as String), defaults to "false"
+  #   If enabled, each new subvolume created on the associated Ceph FS/volume
+  #   will get a unique RADOS namespace created for, and permanently assigned,
+  #   to it. This differs from file/directory layouts, as it is immutable. The
+  #   name of the new NS is not configurable, as it is automatically derived
+  #   from the subvolume name with an "fsvolumens_" prefix.
+  # namespaceIsolated: "true"
 
 reclaimPolicy: Delete
 allowVolumeExpansion: true
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -274,6 +274,9 @@ func buildCreateVolumeResponse(
 	volumeContext := util.GetVolumeContext(req.GetParameters())
 	volumeContext["subvolumeName"] = vID.FsSubvolName
 	volumeContext["subvolumePath"] = volOptions.RootPath
+	if volOptions.PoolNamespace != "" {
+		volumeContext["radosNamespace"] = volOptions.PoolNamespace
+	}
 	volume := &csi.Volume{
 		VolumeId:      vID.VolumeID,
 		CapacityBytes: volOptions.Size,
@@ -468,6 +471,16 @@ func (cs *cephfsControllerServer) CreateVolume(
 			log.ErrorLog(ctx, "failed to get subvolume path %s: %v", vID.FsSubvolName, err)
 
 			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		if volOptions.NamespaceIsolated {
+			svInfo, err := volClient.GetSubVolumeInfo(ctx)
+			if err != nil {
+				log.WarningLog(ctx, "failed to get subvolume info for pool namespace %s: %v",
+					vID.FsSubvolName, err)
+			} else {
+				volOptions.PoolNamespace = svInfo.PoolNamespace
+			}
 		}
 
 		// Set Metadata on PV Create
--- a/internal/cephfs/controllerserver_test.go
+++ b/internal/cephfs/controllerserver_test.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2026 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cephfs
+
+import (
+	"testing"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ceph/ceph-csi/internal/cephfs/store"
+)
+
+func TestBuildCreateVolumeResponse_WithRadosNamespace(t *testing.T) {
+	t.Parallel()
+	const ns = "csi-vol-4c742621-090a-484d-a33b"
+	req := &csi.CreateVolumeRequest{Parameters: map[string]string{}}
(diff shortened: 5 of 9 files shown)


Madhu-1:
@vt0r Thanks for the PR, please see the latest update on https://github.com/rook/rook/issues/17571 we need a usecase to provide isolation at the pvc level

nixpanic:
@vt0r , can you explain the advantage of the isolated namespace for Ceph-CSI created volumes? How do you take advantage of them, and what additional steps or configuration is needed to improve a multi-tenancy deployment?

vt0r:
Hey, @Madhu-1 and @nixpanic. Thank you both for the quick replies, and apologies for my delayed one. As for the use case, I believe @ideepika has already elaborated the case very well, but I will attempt to add additional context. In the process of doing so, I will hopefully not reiterate too many of @ideepika's same points.

I chose the existing solution, which happens to be subvolumes, mostly because it's already supported fully [in Ceph](https://github.com/ceph/ceph/pull/34776/changes) and [in `go-ceph`](https://github.com/ceph/go-ceph/pull/342/changes#diff-57b7fbe7bf4be97ff70b2f611837613c3b964cb23a4d4df97e0a628fe056c069R20) for a while now, so it was just a simple wiring job, but I also do believe subvolumes are the most logical choice for use case isolation, as they are the layer you directly interact with on the client side when mounting. They also map directly to a PVC, to snapshots, and to existing k8s namespace isolation concepts, etc.

Another reason it fits well is that the default out of the box setup with Rook is to create a single subvolume group called `csi` with a single StorageClass, and this is what most people who are first starting out, i.e. with the quick start guide, will end up with (and will often keep forever 😅).

I realize sticking with this single subvolume group is not always the optimal setup - I would even go so far as saying it's very often the wrong setup, especially as the number of inodes grows naturally over time, forcing new Ceph cluster admins to learn about configuring MDS ranks and tuning the balancer, but most setups are small enough to get away with it, even for multi-tenant use cases. This is especially true when it's not a stretch cluster, giving you a natural, single-cluster boundary that limits the growth to an extent. The less total capacity there is and the smaller total number of unique users, the less likely you are to end up with The One tenant who teaches you a lesson about the inefficiencies of dealing with too many inodes.

That being said, even with subvolume groups typically being the "correct" solution for a per-tenant use case, they are often too "large" a boundary to actually separate by tenant, as it's not uncommon to need isolation, even between a single tenant's use cases, as using a group per tenant effectively sets all subvolumes underneath (unless overridden later with individual file/dir layouts or enabling namespace isolation on one or more of its subvolumes at creation time) to use one single namespace, which is often not appropriate or necessary. For example, one tenant may store a shared database, web application files, and some shared load balancer configurations for a typical three-tier web application under their subvolume group, none of which should be able to directly modify or even directly read the files of the other in most cases. With a single namespace for the group, this means you must rely solely on POSIX permissions to ensure access is not allowed to the data of one tier by another, whereas with RADOS namespace isolation and dedicated, scoped clients, your data is segregated fully per client, regardless of on-FS POSIX permissions. 

The subvolume level isolation also has the awesome benefit of being immutable, so unlike a file/directory layout, it cannot be undone, so there is no need to think about whether it was set before or after any specific files/directories landed under the given subvolume group (or path within the subvolume group/subvolume), as it's set from creation time and persists indefinitely. There is no need to worry about it being updated intentionally or inadvertently as a result.

As for what it takes to be useful, one can either set the new parameter in their StorageClass when creating a new one for the given Ceph FS, create a new SC relevant to the FS with that parameter, or they can delete and recreate their existing SC if one already exists, as parameters can not be modified on existing SCs - you all know this, I'm sure, but it helps to be explicit about steps anyway... Then, each time a subvolume gets created, the new returned `radosNamespace` value can be used to create a scoped CephX client that has `r` or `rw` caps only on the full subvolume path in the MDS, `r` to `*` in the monitors, and `r` or `rw` to just the pool and pool namespace. At this point, each client you've performed these steps for is fully isolated, even if they connect directly via RADOS. It does take some additional [CRD calls](https://rook.io/docs/rook/latest-release/CRDs/ceph-client-crd/) to get that done, but it's worthwhile imo.

vt0r:
RE the force push: Added an example of setting the new `namespaceIsolated` parameter in `examples/cephfs/storageclass.yaml`, which I'd neglected to do before when updating the docs, then rebased on the tip of `devel`.

vt0r:
<details>
<summary>Original details about snapshot cloning based on an incorrect read of the code</summary>

---

Rebased on `devel` again, this time to add a change to the snapshot cloning, as [we currently set the `Pool` value of the clone command](https://github.com/ceph/ceph-csi/blob/b22ca6065acb931e9c343b40b8bb273aa552c38e/internal/cephfs/core/snapshot.go#L178-L180) using the value from the k8s StorageClass, which is correct in most cases, however in the case of cloning a snapshot from a namespace-isolated subvolume, Ceph Manager's `create_clone()` function [will currently override existing `data_pool` and `pool_namespace` values from the source snapshot's `attrs`](https://github.com/ceph/ceph/blob/76df7a90b27dddcfb888d7008348f811a4ae50e7/src/pybind/mgr/volumes/fs/operations/versions/subvolume_v2.py#L232-L235) using the provided `pool` value from the SC for `data_pool`, and explicitly setting `pool_namespace` to `None`, regardless of the original value. This means anyone who moves to the isolated namespace model for subvolumes loses that isolation benefit whenever they clone.

I believe the best compromise option we can make today is to make this change here on the CSI driver side to copy the existing values, so that a cloned snapshot retains both the RADOS data pool and namespace values from the original subvolume/snapshot. This of course means all cloned subvolumes are by default sharing RADOS namespace data locality with the original, but considering most snapshot clones should be reused by the original owning end-user/tenant (based on anecdata), this compromise is a much more sensible default than the alternative: always emptying the `pool_namespace` value and defaulting to the default one + the default SC's pool when making a clone, meaning every clone "escapes" the isolated NS model. This change includes the compromise option, which maintains expected behavior of clones created from NS isolated subvolume's snapshots, and it _does not_ change existing behavior for non-isolated ones, so I believe this makes it safe for general use.

Though I do believe this does make for a sensible default, a longer-term solution that would allow isolating even the snapshots or their clones might be to add a similar `--namespace-isolated` flag to the snapshot and/or snapshot clone creation in Ceph itself, making sharing snapshots between tenants/end-users while still adhering to the isolation model a possibility. That to me seems like a proper fix for the use case where a snapshot from a NS-isolated subvolume is cloned from one end user/tenant to another, which I unfortunately cannot address in this PR and will consider addressing separately upstream.

Also added some additional clone tests to confirm the sanity of the options builder and enabled/disabled scenarios, just as was done for volumes.

---

</details>

Ugh 🤦 . Apologies. The stuff in the collapsed section was almost entirely wrong. Though some of the details were accurate, the `attrs` retrieved during a clone are _not_ those of the source subvolume, but of the source _snapshot_, which does not have any such attributes. I have reverted that part of the change, as it seems snapshots and clones currently don't support a non-default namespace at all. I will need to revisit the clone change portion as an actual upstream change to Ceph, and I'll open a separate PR when that gets done.

vt0r:
> The stuff in the collapsed section was almost entirely wrong.

Actually, upon further local testing, I was able to confirm manually cloning a subvolume snapshot _does_ preserve the existing pool _and_ pool namespace, so I was on the right track there to adjust the way the CSI driver makes that call, as it clearly can be done - just needed to get the right combination of `--group_name` and `--target_group_name`, which I unfortunately am blanking on at the moment, but it did work as expected.

That being said, I am still working on testing and finalizing the upstream Ceph (Manager) pull request, as I plan to extend the clone functionality in a way that also allows (optionally) creating the clones with their own unique pool namespace, just in case the user does not want to share one with the source subvolume. So, I do still think I would like to leave this PR as is, and I can introduce the isolated clones in a separate PR, as its current state has already been well tested.

Additionally, I believe this PR is complimentary to ceph/ceph#69796 and should not interfere with or block/be blocked by it, but please do let me know if you disagree or would prefer I rework this in any way to work more closely with that change.

PR #6362, opened 2026-06-25
rbd: add fatal error handling for ReclaimSpace operations

nixpanic:
Replace calling `fstrim` with the underlaying syscall. This makes it possible
to detect different failures. Certain failures indicate that either the
filesystem or block-device does not support the fstrim operation. If this is
the case, CSI-Addons should not try again unless the volume was re-attached or
components were restarted.

One particular case is that the Ceph pool is configured in such a way (with
certain erasure encoding settings) that fstrim is not possible (yet?).
CSI-Addons should not retry the failing operation in this case.

See-also: csi-addons/spec#81 (kubernetes-csi-addons already supports it)

### DIFF
--- a/internal/csi-addons/rbd/fstrim.go
+++ b/internal/csi-addons/rbd/fstrim.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2026 The Ceph-CSI Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


internal/csi-addons/rbd/fstrim.go
```diff
@@ -0,0 +1,91 @@
...
+
+// ErrTrimNotSupported indicates the filesystem does not support FITRIM.
+var ErrTrimNotSupported = errors.New("trim operation not supported by filesystem")
+
+// fsTrim performs filesystem trim operation on the given path using FITRIM ioctl.
+// This is equivalent to running 'fstrim' command but uses direct ioctl call.
+// Returns ErrTrimNotSupported if the filesystem does not support trim operations.
+func fsTrim(path string) error {
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
```
nixpanic: golang-ci complains about a missing line after `return` here... that doesn't make sense at all.

On the other hand, I'm not a huge fan of the embedded C trim_filesystem() function. It might be nicer to drop that C function and write everything with C.types here.

nixpanic: Rewrote the function to use the `syscall` package. This still needs testing.

PR #6375, opened 2026-06-30
kms: add client certificate support to the vault KMS type

sergeykuperman:
## What this PR does / why we need it

The `vault` KMS type supports Kubernetes SA token authentication (no
per-namespace Secrets required) but previously had no way to present a
client certificate to Vault, making mTLS impossible.

This PR adds `vaultClientCertFromSecret` and `vaultClientCertKeyFromSecret`
config options to `vaultConnection.initCertificates()`, closing a symmetric
gap with the `vaulttokens` type which already has full mTLS support.

**Before:** `vault` could verify the server cert (`vaultCAFromSecret`) but
could not present a client cert.

**After:** `vault` supports all three TLS config options, all using the same
`k8s.GetSecret` mechanism as `vaulttokens`:
- `vaultCAFromSecret` — Kubernetes Secret name for the CA cert (field: `cert`)
- `vaultClientCertFromSecret` — **New**: Kubernetes Secret name for the client cert (field: `cert`)
- `vaultClientCertKeyFromSecret` — **New**: Kubernetes Secret name for the client key (field: `key`)

All three Secrets are looked up in the CSI pod namespace (`POD_NAMESPACE`).

`Destroy()` is also extended to clean up temp files for the client cert and
key alongside the existing CA cert cleanup.

## ⚠️ Breaking change

`vaultCAFromSecret` behaviour in the `vault` KMS type has changed.

**Before:** the config value was treated as a **key name** within the
node-stage secret map (`args.Secrets`), i.e. a field name in the Kubernetes
Secret referenced by `spec.csi.nodeStageSecretRef` on the PV. The CA PEM
was expected to be stored directly as a value in that Secret.

**After:** the config value is a **Kubernetes Secret name**, consistent with
how `vaulttokens` has always handled `vaultCAFromSecret`. The CA PEM is read
from the `"cert"` field of that named Secret via the Kubernetes API.

**Migration:** users who had `vaultCAFromSecret` configured in a `vault` KMS
config must:
1. Create a standalone Kubernetes Secret in the CSI pod namespace containing
   the CA PEM under the key `"cert"`
2. Update `vaultCAFromSecret` in their KMS config to name that Secret

## Which issue(s) this PR fixes

Fixes a gap where users needing Kubernetes SA auth + mTLS to Vault had no
supported path. The `vaulttokens` type requires a per-tenant Secret (one per
PVC namespace), which is impractical at scale. The `vault` type with mTLS
support is the correct solution for large deployments.

## Special notes for your reviewer

The `secrets map[string]string` parameter is intentionally kept in the
`initCertificates` signature (though now unused after the `vaultCAFromSecret`
migration) to avoid touching the call site in `initVaultKMS`. Happy to remove
it and update the call site if preferred.

## How to test

```bash
go test -v github.com/ceph/ceph-csi/internal/kms
```

New unit tests added:
- `TestInitCertificatesConfigParsing` — config parsing error paths (no cluster needed)
- `TestDestroyRemovesAllTempFiles` — verifies all three temp files are removed on Destroy()

### DIFF
--- a/internal/kms/vault.go
+++ b/internal/kms/vault.go
@@ -31,6 +31,7 @@ import (
 	"github.com/libopenstorage/secrets/vault"
 
 	"github.com/ceph/ceph-csi/internal/util/file"
+	"github.com/ceph/ceph-csi/internal/util/k8s"
 )
 
 const (
@@ -69,10 +70,18 @@ Example JSON structure in the KMS config is,
 		"vaultPassphraseRoot": "/v1/secret",
 		"vaultPassphrasePath": "",
 		"vaultCAVerify": true,
-		"vaultCAFromSecret": "vault-ca"
+		"vaultCAFromSecret": "vault-ca-secret",
+		"vaultClientCertFromSecret": "vault-client-cert-secret",
+		"vaultClientCertKeyFromSecret": "vault-client-cert-secret"
 	},
 	...
 }.
+
+"vaultCAFromSecret", "vaultClientCertFromSecret", and "vaultClientCertKeyFromSecret"
+are Kubernetes Secret names in the CSI pod namespace. The CA cert is read from
+the "cert" field of the named Secret; the client cert from the "cert" field and
+the client key from the "key" field of their respective Secrets (cert and key
+may be in the same Secret or different Secrets).
 */
 
 type vaultConnection struct {
@@ -156,10 +165,11 @@ func setConfigBoolean(option *bool, config map[string]any, key string) error {
 // Destroy frees allocated resources. For a vaultConnection that means removing
 // the created temporary files.
 func (vc *vaultConnection) Destroy() {
-	if vc.vaultConfig != nil {
-		tmpFile, ok := vc.vaultConfig[api.EnvVaultCACert]
-		if ok {
-			// ignore error on failure to remove tmpfile (gosec complains)
+	if vc.vaultConfig == nil {
+		return
+	}
+	for _, key := range []string{api.EnvVaultCACert, api.EnvVaultClientCert, api.EnvVaultClientKey} {
+		if tmpFile, ok := vc.vaultConfig[key]; ok {
 			//nolint:forcetypeassert,errcheck // ignore error on failure to remove tmpfile
 			_ = os.Remove(tmpFile.(string))
 		}
@@ -298,6 +308,13 @@ func (vc *vaultConnection) initConnection(config map[string]any) error {
 // vc.connectVault().
 func (vc *vaultConnection) initCertificates(config map[string]any, secrets map[string]string) error {
 	vaultConfig := make(map[string]any)
+	csiNamespace := os.Getenv("POD_NAMESPACE")
+
+	// Merge vaultConfig into vc.vaultConfig eagerly on return so that any temp
+	// files created before a partial failure are tracked and cleaned up by Destroy().
+	defer func() {
+		maps.Copy(vc.vaultConfig, vaultConfig)
+	}()
 
 	vaultCAFromSecret := "" // optional
 	err := setConfigString(&vaultCAFromSecret, config, "vaultCAFromSecret")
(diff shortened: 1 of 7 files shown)


sergeykuperman:
**TL;DR:** The `vault` KMS type (Kubernetes SA auth) can verify the server cert but cannot present a client certificate, making application-level mTLS impossible. This PR adds `vaultClientCertFromSecret` and `vaultClientCertKeyFromSecret` to `vaultConnection.initCertificates()` — the same options already supported by `vaulttokens`, using the same `k8s.GetSecret()` mechanism.

**Why it matters:** Users running Rook-Ceph with `hostNetwork: true` (enforced since Rook v1.17+) cannot use Istio sidecar injection on the CSI node plugin, so transparent mTLS is unavailable. `vaulttokens` is not a viable alternative at scale (requires a static token Secret in every tenant namespace). The `vault` type with mTLS is the only compliant path.

**Scope:** ~130 lines, all within `internal/kms/vault.go`. No changes to callers, `ProviderInitArgs`, or vendor files. Unit tests included.

Tracking issue: #6376

---

@black-dragon74 @Rakshith-R @iPraveenParihar — could one of you take a look when you get a chance? The issue and PR have been sitting without labels or a reviewer assigned for ~4 weeks. Would also appreciate if someone with repo write access could add the relevant labels (`enhancement`, `component/rbd`, `security`, `keepalive`) to prevent the stale bot from closing it. Happy to address any review feedback promptly.

internal/kms/vault.go
```diff
@@ -298,28 +308,72 @@ func (vc *vaultConnection) initConnection(config map[string]any) error {
...
+			return fmt.Errorf("missing \"key\" field in secret %s", vaultClientCertKeyFromSecret)
+		}
+		tf, err := file.CreateTempFile("vault-client-cert-key", keyPEM)
+		if err != nil {
+			return fmt.Errorf("failed to create temporary file for Vault client certificate key: %w", err)
+		}
+		vaultConfig[api.EnvVaultClientKey] = tf.Name()
+	}
+
+	maps.Copy(vc.vaultConfig, vaultConfig)
```
iPraveenParihar: If the CA cert temp file is created successfully but a subsequent step fails (e.g., k8s.GetSecret for client cert), the function returns early. The temp file path is only in the local vaultConfig map never copied to vc.vaultConfig so Destroy() will never clean it up?

iPraveenParihar: either copy into vc.vaultConfig after each temp file creation or add a defer cleanup here.

internal/kms/vault.go
```diff
@@ -298,28 +308,72 @@ func (vc *vaultConnection) initConnection(config map[string]any) error {
...
 
 	vaultCAFromSecret := "" // optional
 	err := setConfigString(&vaultCAFromSecret, config, "vaultCAFromSecret")
 	if errors.Is(err, errConfigOptionInvalid) {
 		return err
 	}
-	// ignore errConfigOptionMissing, no default was set
 	if vaultCAFromSecret != "" {
-		caPEM, ok := secrets[vaultCAFromSecret]
+		cert, err := k8s.GetSecret(vaultCAFromSecret, csiNamespace)
```
iPraveenParihar, line 326: Is this behaviour change?
Before, it reads PEM data from the passed-in secrets map. Here, it calls K8s API with the value as a Secret name?

Existing users who had `vaultCAFromSecret` working with the old way would need to restructure .The PEM data now needs to be in a standalone K8s Secret with a "cert" field. This is a breaking change and needs to be documented?

iPraveenParihar (changes requested):

internal/kms/vault.go
sergeykuperman: Good catch — fixed. Added a `defer` that merges `vaultConfig` into `vc.vaultConfig` at function entry, so any temp files created before a partial failure are tracked and cleaned up by `Destroy()`.

internal/kms/vault.go
sergeykuperman: Yes, this is an intentional breaking change, now clearly documented. `vaultCAFromSecret` previously treated the config value as a key name within `args.Secrets` (the node-stage secret map delivered by the kubelet from the PV's `nodeStageSecretRef`). It now behaves consistently with the `vaulttokens` type: the value is a Kubernetes Secret name, and the CA PEM is read from the `"cert"` field of that Secret via the k8s API.

Migration path for existing users: create a standalone Kubernetes Secret in the CSI pod namespace containing the CA PEM under key `"cert"`, and update `vaultCAFromSecret` to name that Secret instead of the node-stage secret field name.

This is documented in the commit message and the PR description. Also added documentation in relevant yaml and md files

sergeykuperman: @iPraveenParihar I have done the requested changes, when you have a min to take a look

sergeykuperman: @iPraveenParihar ping waiting for your review

sergeykuperman: @black-dragon74 @Rakshith-R @iPraveenParihar could someone review again, i have fixed the comments

sergeykuperman: @black-dragon74 @Rakshith-R @iPraveenParihar @nixpanic pinging, could someone review again

PR #6382, opened 2026-07-06
rbd: (filesystem) add async health checks to NodeGetVolumeStats

SanjalKatiyar:
Pre-requisite PR: https://github.com/ceph/ceph-csi/pull/6418

Use the shared health-check manager in the RBD node server so `NodeGetVolumeStats` reports async health state before calling `os.Stat()`.
Avoids stat hangs and aligns RBD behaviour with CephFS.
Start/Stop shared checkers during stage/unstage and incorporated fallback in NodeGetVolumeStats for node-plugin restart scenarios.
Only supported for filesystem-mode volumes.

##### Logs of `openshift-storage.rbd.csi.ceph.com-nodeplugin-<HASH>` pod:
Returns `"volume_condition":{"message":"block-mode health checking is not supported"}` for RBD block.
Returns `"volume_condition":{"message":"volume is in a healthy condition"}` for RBD filesystem.
```
I0727 13:08:51.304966 1545666 utils.go:368] ID: 3410 GRPC response: {"capabilities":[{"rpc":{"type":"STAGE_UNSTAGE_VOLUME"}},{"rpc":{"type":"GET_VOLUME_STATS"}},{"rpc":{"type":"VOLUME_CONDITION"}},{"rpc":{"type":"EXPAND_VOLUME"}},{"rpc":{"type":"SINGLE_NODE_MULTI_WRITER"}}]}
I0727 13:08:51.305872 1545666 utils.go:361] ID: 3411 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0727 13:08:51.305900 1545666 utils.go:362] ID: 3411 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-dec53106-24a3-4f4d-8ed4-865e614f04ae","volume_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/pvc-1049141e-3ce1-434d-83b3-c10e270a18e3/dev/ca34f0f7-fc11-4552-9dd8-bc91714ef737"}
I0727 13:08:51.312332 1545666 omap.go:87] ID: 3411 got omap values by keys: (pool="ocs-storagecluster-cephblockpool", namespace="", name="csi.volume.dec53106-24a3-4f4d-8ed4-865e614f04ae"): map[csi.imageid:723a5aa4d498 csi.imagename:csi-vol-dec53106-24a3-4f4d-8ed4-865e614f04ae csi.volname:pvc-1049141e-3ce1-434d-83b3-c10e270a18e3 csi.volume.owner:volume-health-test]
I0727 13:08:51.349269 1545666 rbd_util.go:2435] ID: 3411 rbd: running DiffIterate over image ocs-storagecluster-cephblockpool/csi-vol-dec53106-24a3-4f4d-8ed4-865e614f04ae of size 1073741824 bytes
I0727 13:08:51.350479 1545666 rbd_util.go:2456] ID: 3411 DiffIterate on image ocs-storagecluster-cephblockpool/csi-vol-dec53106-24a3-4f4d-8ed4-865e614f04ae of size 1073741824 bytes took 0.00118902 seconds
I0727 13:08:51.350814 1545666 utils.go:368] ID: 3411 GRPC response: {"usage":[{"available":1073741824,"total":1073741824,"unit":"BYTES"}],"volume_condition":{"message":"block-mode health checking is not supported"}}
I0727 13:08:54.234071 1545666 utils.go:361] ID: 3412 GRPC call: /csi.v1.Node/NodeGetCapabilities
I0727 13:08:54.234097 1545666 utils.go:362] ID: 3412 GRPC request: {}
I0727 13:08:54.234192 1545666 utils.go:368] ID: 3412 GRPC response: {"capabilities":[{"rpc":{"type":"STAGE_UNSTAGE_VOLUME"}},{"rpc":{"type":"GET_VOLUME_STATS"}},{"rpc":{"type":"VOLUME_CONDITION"}},{"rpc":{"type":"EXPAND_VOLUME"}},{"rpc":{"type":"SINGLE_NODE_MULTI_WRITER"}}]}
I0727 13:08:54.234936 1545666 utils.go:361] ID: 3413 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0727 13:08:54.234969 1545666 utils.go:362] ID: 3413 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-a52055c9-b783-46b8-8e0f-8972915a7796","volume_path":"/var/lib/kubelet/pods/a765e803-7650-45b5-a73d-7c432b182b1a/volumes/kubernetes.io~csi/pvc-3103d574-581f-4399-92ad-901bebbd1353/mount"}
I0727 13:08:54.235097 1545666 utils.go:368] ID: 3413 GRPC response: {"usage":[{"available":88815403008,"total":98747928576,"unit":"BYTES","used":9915748352},{"available":5475380,"total":6160384,"unit":"INODES","used":685004}],"volume_condition":{"message":"volume is in a healthy condition"}}
```

### DIFF
--- a/internal/health-checker/noopchecker.go
+++ b/internal/health-checker/noopchecker.go
@@ -1,0 +1,40 @@
+/*
+Copyright 2026 ceph-csi authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthchecker
+
+type noOpChecker struct {
+	checker
+}
+
+func newNoOpChecker() ConditionChecker {
+	nc := &noOpChecker{}
+	nc.initDefaults()
+
+	// always healthy, no errors
+	nc.healthy = true
+	nc.err = nil
+
+	nc.checker.runChecker = func() {
+		nc.isRunning = true
+
+		// Wait for stop command
+		<-nc.commands
+		nc.isRunning = false
+	}
+
+	return nc
+}
--- a/internal/health-checker/noopchecker.go
+++ b/internal/health-checker/noopchecker.go
@@ -41,7 +41,7 @@ func newNoOpChecker() ConditionChecker {
 // always healthy, no errors
 // overrides checker.isHealthy() so that the timeout-based
 // staleness check (which compares lastUpdate against interval+timeout)
-// never marks a NoOp checker unhealthy
+// never marks a NoOp checker unhealthy.
 func (nc *noOpChecker) isHealthy() (bool, error) {
 	return true, nil
 }
--- a/internal/health-checker/noopchecker.go
+++ b/internal/health-checker/noopchecker.go
@@ -24,7 +24,6 @@ func newNoOpChecker() ConditionChecker {
 	nc := &noOpChecker{}
 	nc.initDefaults()
 
-	// always healthy, no errors
 	nc.healthy = true
 	nc.err = nil
 
@@ -37,4 +36,12 @@ func newNoOpChecker() ConditionChecker {
 	}
 
 	return nc
+}
+
+// always healthy, no errors
+// overrides checker.isHealthy() so that the timeout-based
+// staleness check (which compares lastUpdate against interval+timeout)
+// never marks a NoOp checker unhealthy
+func (nc *noOpChecker) isHealthy() (bool, error) {
+	return true, nil
 }
--- a/internal/rbd/driver/driver.go
+++ b/internal/rbd/driver/driver.go
@@ -27,6 +27,7 @@ import (
 	csiaddons "github.com/ceph/ceph-csi/internal/csi-addons/server"
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
 	"github.com/ceph/ceph-csi/internal/driver"
+	hc "github.com/ceph/ceph-csi/internal/health-checker"
 	"github.com/ceph/ceph-csi/internal/rbd"
 	"github.com/ceph/ceph-csi/internal/rbd/features"
 	"github.com/ceph/ceph-csi/internal/util"
@@ -82,6 +83,7 @@ func NewNodeServer(
 		DefaultNodeServer: csicommon.NewDefaultNodeServer(d, t, cliReadAffinityMapOptions, topology, nodeLabels),
 		VolumeLocks:       util.NewIDLocker(),
 	}
+	ns.SetHealthChecker(hc.NewHealthCheckManager())
 
 	return &ns
 }
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -34,6 +34,7 @@ import (
 	utilexec "k8s.io/utils/exec"
 
 	csicommon "github.com/ceph/ceph-csi/internal/csi-common"
+	hc "github.com/ceph/ceph-csi/internal/health-checker"
 	rbderrors "github.com/ceph/ceph-csi/internal/rbd/errors"
 	"github.com/ceph/ceph-csi/internal/util"
 	"github.com/ceph/ceph-csi/internal/util/file"
@@ -49,12 +50,18 @@ type NodeServer struct {
 	*csicommon.DefaultNodeServer
 	// A map storing all volumes with ongoing operations so that additional operations
 	// for that same volume (as defined by VolumeID) return an Aborted error
-	VolumeLocks *util.IDLocker
+	VolumeLocks   *util.IDLocker
+	healthChecker hc.Manager
 
 	// ext4HasPrezeroedSupport indicates whether the ext4 filesystem has support for pre-zeroed blocks.
 	ext4HasPrezeroedSupport featureFlag
 	// xfsHasReflinkSupport indicates whether the xfs filesystem has support for reflink.
 	xfsHasReflinkSupport featureFlag
+}
+
+// SetHealthChecker configures the manager used for volume health checks.
+func (ns *NodeServer) SetHealthChecker(manager hc.Manager) {
+	ns.healthChecker = manager
 }
 
 // stageTransaction struct represents the state a transaction was when it either completed
@@ -367,6 +374,7 @@ func (ns *NodeServer) NodeStageVolume(
 	stagingParentPath := req.GetStagingTargetPath()
 	stagingTargetPath := stagingParentPath + "/" + volID
 
+	isBlock := req.GetVolumeCapability().GetBlock() != nil
 	isHealer := parseBoolOption(ctx, req.GetVolumeContext(), volHealerCtx, false)
 	if !isHealer {
 		var isNotMnt bool
@@ -376,6 +384,8 @@ func (ns *NodeServer) NodeStageVolume(
 			return nil, status.Error(codes.Internal, err.Error())
 		} else if !isNotMnt {
 			log.DebugLog(ctx, "rbd: volume %s is already mounted to %s, skipping", volID, stagingTargetPath)
+
+			ns.startSharedHealthChecker(ctx, volID, stagingTargetPath, isBlock)
 
 			return &csi.NodeStageVolumeResponse{}, nil
 		}
@@ -397,6 +407,8 @@ func (ns *NodeServer) NodeStageVolume(
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
+
+		ns.startSharedHealthChecker(ctx, volID, stagingTargetPath, isBlock)
 
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
@@ -435,6 +447,8 @@ func (ns *NodeServer) NodeStageVolume(
 		"rbd: successfully mounted volume %s to stagingTargetPath %s",
 		volID,
 		stagingTargetPath)
+
+	ns.startSharedHealthChecker(ctx, volID, stagingTargetPath, isBlock)
 
 	return &csi.NodeStageVolumeResponse{}, nil
 }
@@ -825,6 +839,21 @@ func (ns *NodeServer) createStageMountPoint(ctx context.Context, mountPath strin
 	return nil
 }
 
+// startSharedHealthChecker starts a health-checker on the stagingTargetPath.
+// This checker can be shared between multiple containers.
+// TODO: Add a real health checker for block-mode volumes.
+func (ns *NodeServer) startSharedHealthChecker(ctx context.Context, volumeID, dir string, isBlock bool) {
+	var checkerType hc.CheckerType = hc.StatCheckerType
+	if isBlock {
+		checkerType = hc.NoOpCheckerType
+	}
+
+	err := ns.healthChecker.StartSharedChecker(volumeID, dir, checkerType)
+	if err != nil {
+		log.WarningLog(ctx, "failed to start healthchecker: %v", err)
+	}
+}
+
 // NodePublishVolume mounts the volume mounted to the device path to the target
 // path.
 func (ns *NodeServer) NodePublishVolume(
@@ -1206,6 +1235,9 @@ func (ns *NodeServer) NodeUnpublishVolume(
 	}
 	defer ns.VolumeLocks.Release(targetPath)
 
+	// stop the health-checker that may have been started in NodeGetVolumeStats()
+	ns.healthChecker.StopChecker(req.GetVolumeId(), targetPath)
+
 	isMnt, err := ns.Mounter.IsMountPoint(targetPath)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -1262,6 +1294,8 @@ func (ns *NodeServer) NodeUnstageVolume(
 	}
 
 	volID := req.GetVolumeId()
+
+	ns.healthChecker.StopSharedChecker(volID)
 
 	if acquired := ns.VolumeLocks.TryAcquire(volID); !acquired {
 		log.ErrorLog(ctx, util.VolumeOperationAlreadyExistsFmt, volID)
@@ -1665,6 +1699,78 @@ func (ns *NodeServer) NodeGetVolumeStats(
 	}
 	defer ns.VolumeLocks.Release(targetPath)
 
+	// Health check first, return without stats if unhealthy.
+	healthy, msg := ns.healthChecker.IsHealthy(volumeId, targetPath)
+
+	// If healthy and an error is returned, it either means that the checker
+	// was not started or it is running but has not completed its
+	// first health-check cycle yet.
+	// This could happen when the node-plugin was restarted and the
+	// volume is already staged and published.
+	if healthy && msg != nil {
+		// Start a StatChecker for the mounted targetPath, this prevents
+		// writing a file in the user-visible location. Ideally a (shared)
+		// FileChecker is started with the stagingTargetPath, but we can't
+		// get the stagingPath from the request easily.
+		// TODO: resolve the stagingPath like rbd.getStagingPath() does
+		// NOTE: rbd.getStagingPath() uses os.Stat() internally which
+		// if called synchronously, could block indefinitely.
+
+		// Start the background checker and return
+		// immediately instead of calling os.Stat() on this goroutine.
+		// If the mount is unresponsive, os.Stat() would block,
+		// holding the VolumeLock (acquired above) and preventing all future
+		// calls for this path from reaching isHealthy().
+		// The background checker will do the stat(), the next periodic
+		// call will pick up the result (or detect the timeout).
+
+		var checkerType hc.CheckerType = hc.StatCheckerType
+
+		// Known bug for filesystem-mode volumes, still causing
+		// the calling goroutine to block indefinitely on an
+		// unresponsive mount:
+		// when the node-plugin restarts
+		// (healthy = true, msg = "first check not yet completed"),
+		// this block of code is executed and if the mount is already
+		// unresponsive, this call blocks while holding the VolumeLock,
+		// causing all subsequent NodeGetVolumeStats calls for this path
+		// to fail with "Aborted".
+		// TODO: remove synchronous os.Stat() for volume mode detection here.
+		stat, statErr := os.Stat(targetPath)
+		if statErr != nil {
+			log.WarningLog(ctx, "failed to stat targetPath %q: %v",
+				targetPath, statErr)
+		} else if (stat.Mode() & os.ModeDevice) == os.ModeDevice {
+			checkerType = hc.NoOpCheckerType
+		}
+
+		err = ns.healthChecker.StartChecker(volumeId, targetPath, checkerType)
+		if err != nil {
+			log.WarningLog(ctx, "failed to start healthchecker: %v", err)
+		}
+
+		return &csi.NodeGetVolumeStatsResponse{
+			VolumeCondition: &csi.VolumeCondition{
+				Abnormal: false,
+				Message:  "health checker started, status not yet available",
+			},
+		}, nil
+	}
+
+	// !healthy indicates a problem with the volume.
+	if !healthy {
+		return &csi.NodeGetVolumeStatsResponse{
+			VolumeCondition: &csi.VolumeCondition{
+				Abnormal: true,
+				Message:  msg.Error(),
+			},
+		}, nil
+	}
+
+	// warning: reaching here should mean that synchronous os.Stat()
+	// call is safe and will not indefinitely block/hang
+	// NOTE: we have a known bug (mentioned above) which still
+	// causes hangs, that needs to be fixed.
 	stat, err := os.Stat(targetPath)
 	if err != nil {
 		if util.IsCorruptedMountError(err) {
@@ -1768,7 +1874,7 @@ func (ns *NodeServer) blockNodeGetVolumeStats(
 		},
 		VolumeCondition: &csi.VolumeCondition{
 			Abnormal: false,
-			Message:  "volume is in a healthy condition",
+			Message:  "block-mode health checking is not supported",
 		},
 	}, nil
 }
@@ -1796,7 +1902,7 @@ func getBlockMetrics(
 		},
 		VolumeCondition: &csi.VolumeCondition{
 			Abnormal: false,
-			Message:  "volume is in a healthy condition",
+			Message:  "block-mode health checking is not supported",
 		},
 	}, nil
 }
--- a/internal/rbd/nodeserver.go
+++ b/internal/rbd/nodeserver.go
@@ -385,6 +385,10 @@ func (ns *NodeServer) NodeStageVolume(
 		} else if !isNotMnt {
 			log.DebugLog(ctx, "rbd: volume %s is already mounted to %s, skipping", volID, stagingTargetPath)
 
+			if stashErr := stashVolumeInfo(volID, &volumeInfo{IsBlockMode: isBlock}); stashErr != nil {
+				log.WarningLog(ctx, "failed to stash volume info for %s: %v", volID, stashErr)
+			}
+
 			ns.startSharedHealthChecker(ctx, volID, stagingTargetPath, isBlock)
 
 			return &csi.NodeStageVolumeResponse{}, nil
@@ -406,6 +410,10 @@ func (ns *NodeServer) NodeStageVolume(
 		err = healerStageTransaction(ctx, cr, rv, stagingParentPath)
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		if stashErr := stashVolumeInfo(volID, &volumeInfo{IsBlockMode: isBlock}); stashErr != nil {
+			log.WarningLog(ctx, "failed to stash volume info for %s: %v", volID, stashErr)
 		}
 
 		ns.startSharedHealthChecker(ctx, volID, stagingTargetPath, isBlock)
@@ -447,6 +455,10 @@ func (ns *NodeServer) NodeStageVolume(
 		"rbd: successfully mounted volume %s to stagingTargetPath %s",
 		volID,
 		stagingTargetPath)
+
+	if stashErr := stashVolumeInfo(volID, &volumeInfo{IsBlockMode: isBlock}); stashErr != nil {
+		log.WarningLog(ctx, "failed to stash volume info for %s: %v", volID, stashErr)
+	}
 
 	ns.startSharedHealthChecker(ctx, volID, stagingTargetPath, isBlock)
 
@@ -1354,6 +1366,11 @@ func (ns *NodeServer) NodeUnstageVolume(
 
 		// It was not mounted and image metadata is also missing, we are done as the last step in
 		// the staging transaction is complete
+		// remove the volume info stash for this volumeID (in case it exists)
+		if err = removeVolumeInfo(volID); err != nil {
+			log.WarningLog(ctx, "failed to remove volume info for %s: %v", volID, err)
+		}
+
 		return &csi.NodeUnstageVolumeResponse{}, nil
 	}
 
@@ -1387,6 +1404,10 @@ func (ns *NodeServer) NodeUnstageVolume(
 		log.ErrorLog(ctx, "failed to cleanup image metadata stash (%v)", err)
 
 		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if err = removeVolumeInfo(volID); err != nil {
+		log.WarningLog(ctx, "failed to remove volume info for %s: %v", volID, err)
 	}
 
 	return &csi.NodeUnstageVolumeResponse{}, nil
@@ -1716,6 +1737,14 @@ func (ns *NodeServer) NodeGetVolumeStats(
 		// NOTE: rbd.getStagingPath() uses os.Stat() internally which
 		// if called synchronously, could block indefinitely.
 
+		var checkerType hc.CheckerType = hc.StatCheckerType
+		info, lookupErr := lookupVolumeInfo(volumeId)
+		if lookupErr == nil && info.IsBlockMode {
+			checkerType = hc.NoOpCheckerType
+		} else if lookupErr != nil {
+			log.DebugLog(ctx, "could not lookup volume info for %s: %v", volumeId, lookupErr)
+		}
+
 		// Start the background checker and return
 		// immediately instead of calling os.Stat() on this goroutine.
 		// If the mount is unresponsive, os.Stat() would block,
@@ -1723,27 +1752,6 @@ func (ns *NodeServer) NodeGetVolumeStats(
 		// calls for this path from reaching isHealthy().
 		// The background checker will do the stat(), the next periodic
 		// call will pick up the result (or detect the timeout).
-
-		var checkerType hc.CheckerType = hc.StatCheckerType
-
-		// Known bug for filesystem-mode volumes, still causing
-		// the calling goroutine to block indefinitely on an
-		// unresponsive mount:
-		// when the node-plugin restarts
-		// (healthy = true, msg = "first check not yet completed"),
-		// this block of code is executed and if the mount is already
-		// unresponsive, this call blocks while holding the VolumeLock,
-		// causing all subsequent NodeGetVolumeStats calls for this path
-		// to fail with "Aborted".
-		// TODO: remove synchronous os.Stat() for volume mode detection here.
-		stat, statErr := os.Stat(targetPath)
-		if statErr != nil {
-			log.WarningLog(ctx, "failed to stat targetPath %q: %v",
-				targetPath, statErr)
-		} else if (stat.Mode() & os.ModeDevice) == os.ModeDevice {
-			checkerType = hc.NoOpCheckerType
-		}
-
 		err = ns.healthChecker.StartChecker(volumeId, targetPath, checkerType)
 		if err != nil {
 			log.WarningLog(ctx, "failed to start healthchecker: %v", err)
@@ -1769,8 +1777,6 @@ func (ns *NodeServer) NodeGetVolumeStats(
 
 	// warning: reaching here should mean that synchronous os.Stat()
 	// call is safe and will not indefinitely block/hang
-	// NOTE: we have a known bug (mentioned above) which still
-	// causes hangs, that needs to be fixed.
 	stat, err := os.Stat(targetPath)
 	if err != nil {
 		if util.IsCorruptedMountError(err) {
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -2031,6 +2031,63 @@ func cleanupRBDImageMetadataStash(metaDataPath string) error {
 	return nil
 }
 
+// The "/csi" directory inside the container is backed by the
+// host-path volume "socket-dir" (ref rbd-plugin DaemonSet yaml file)
+// which maps to "/var/lib/kubelet/plugins/rbd.csi.ceph.com" on the node.
+const volumeInfoDir = "/csi/volume-info"
+
+// Holds per-volume metadata that needs to persist across
+// node-plugin restarts, indexed by volumeID.
+type volumeInfo struct {
+	IsBlockMode bool `json:"isBlockMode"`
+}
+
+// This function is different from "stashRBDImageMetadata" which stores info at stagingPath.
+// Not all RPCs have access to the stagingPath. For example,
+// NodeGetVolumeStats uses this to determine the volume mode using volumeID as the index.
+func stashVolumeInfo(volumeID string, info *volumeInfo) error {
+	if err := os.MkdirAll(volumeInfoDir, 0o750); err != nil {
+		return fmt.Errorf("failed to create volume-info directory: %w", err)
+	}
+
+	encodedBytes, err := json.Marshal(info)
+	if err != nil {
+		return fmt.Errorf("failed to marshal volume info for %s: %w", volumeID, err)
+	}
+
+	fPath := filepath.Join(volumeInfoDir, volumeID+".json")
+	if err = os.WriteFile(fPath, encodedBytes, 0o600); err != nil {
+		return fmt.Errorf("failed to stash volume info for %s: %w", volumeID, err)
+	}
+
+	return nil
+}
+
+func lookupVolumeInfo(volumeID string) (*volumeInfo, error) {
+	fPath := filepath.Join(volumeInfoDir, volumeID+".json")
+
+	encodedBytes, err := os.ReadFile(fPath) // #nosec - intended reading from fPath
+	if err != nil {
+		return nil, err
+	}
+
+	var info volumeInfo
+	if err = json.Unmarshal(encodedBytes, &info); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal volume info for %s: %w", volumeID, err)
+	}
+
+	return &info, nil
+}
+
+func removeVolumeInfo(volumeID string) error {
+	fPath := filepath.Join(volumeInfoDir, volumeID+".json")
+	if err := os.Remove(fPath); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed to remove volume info for %s: %w", volumeID, err)
+	}
+
+	return nil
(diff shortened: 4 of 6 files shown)


internal/rbd/nodeserver.go
```diff
@@ -825,6 +838,15 @@ func (ns *NodeServer) createStageMountPoint(ctx context.Context, mountPath strin
 	return nil
 }
 
+// startSharedHealthChecker starts a health-checker on the stagingTargetPath.
+// This checker can be shared between multiple containers.
+func (ns *NodeServer) startSharedHealthChecker(ctx context.Context, volumeID, dir string) {
+	err := ns.healthChecker.StartSharedChecker(volumeID, dir, hc.StatCheckerType)
```
nixpanic: `StatCheckerType` probably does not work for volumes in block-mode. We don't have a `BlockCheckerType` yet 😞

SanjalKatiyar: @nixpanic IIUC `ns.healthChecker.StartSharedChecker(volumeID, dir, hc.StatCheckerType)` will just call `os.Stat()` **asynchronously**, that's all, right ?? We already do `os.Stat()` synchronously for RBD today.

Even the same happens for CephFS today, we don't specifically call `FileCheckerType` right now anywhere. I tried to keep the behaviour of CephRBD consistent with the existing CephFS :/

Sorry, I didn't get `BlockCheckerType` part.

nixpanic: This is about [Volume Modes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-mode). For RBD it only makes sense (at the moment) to do health checking for Filesystem-mode volumes.

For Block mode volumes:

There is no use in calling `os.Stat()` on a device file like `/dev/rbd0`. The result is just details about the (local) device file, not about the actual (remote) block device that can be used through the file.

A new `BlockCheckerType` would be needed to actually read from the device (similar to a `dd` command), and see if that works.


For Filesystem mode volumes (like CephFS):

A `os.Stat()` actually does an operation on the filesystem that is mounted. The operation gets sent to the Ceph cluster, so there is some networking and I/O involved.

It would be a better validation to write/read from a file on the filesystem, but we don't want such a file visible by applications/users of the volume.

SanjalKatiyar: > A new BlockCheckerType would be needed to actually read from the device

We could open the device in read-only mode and read a small chunk (few bytes), but it would be a privacy concern as we are loading actual user data in-memory.

I also saw your comment about using a partition `Block mode: read/writing to a private partition (hide the 1st partition, map the 2nd partition as device for apps?)`, but I am not sure if that's okay to do ?? It would change the disk layout, which could be a regression for existing users (how would we hide ??).
Also, I am not an expert on this, but just thinking out loud, could this affect other capabilities/features such as expand, snapshots, cloning etc ?? Or, maybe any other actual concern which I might have missed ??

I don't have a strong alternative in mind yet, so would like your inputs/suggestions. Should we block health reporting for block-mode RBD (or document it as unsupported), or is there any method you recommend @nixpanic ??

SanjalKatiyar: > I don't have a strong alternative in mind yet

> Should we block health reporting for block-mode RBD (or document it as unsupported), or is there any method you recommend

I explored a few options and did a fair amount of back-and-forth with diff LLM models (Sonnet/Opus/Codex/Grok), following seemed like somewhat relatively reasonable:
```
1. Use the existing blockNodeGetVolumeStats flow's own success/failure:

The code already connects to Ceph, opens the image, runs DiffIterate. If any of that fails, you already know health is bad. Today the code returns a gRPC error status on failure, which the CO treats as "stats unavailable" — not "volume unhealthy." You'd change the error handling to set Abnormal: true in the VolumeCondition instead of (or in addition to) returning an error. This is free — no ListWatchers, no new probe, just using the data you already have.

2. Add ListWatchers on top of option 1 (catches the kernel-session-died-but-CSI-plugin-works corner case):

Adds the watcher count check to distinguish between "CSI plugin can reach Ceph" and "the actual kernel mapping is still alive." More thorough, but only catches a relatively uncommon failure mode.

3. Skip the health checker for block at NodeStageVolume time (the other half):

Regardless of which probe you use in NodeGetVolumeStats, you should stop starting StatCheckerType for block volumes at stage time. That background goroutine is running os.Stat() on a local file every 60s — it can't detect anything real for block mode.
```

Example snippet/flow (generated by LLM):
```
// isMappingAlive checks if the kernel mapping's watcher is still registered
// on the image header. Opens the image read-only to avoid exclusive lock
// interaction with the active mapping.
func (rv *rbdVolume) isMappingAlive() (bool, error) {
	image, err := rv.openReadOnly()
	if err != nil {
		return false, fmt.Errorf("failed to open image %s read-only: %w", rv, err)
	}
	defer image.Close() //nolint:errcheck // not a critical failure

	watchers, err := image.ListWatchers()
	if err != nil {
		return false, fmt.Errorf("failed to list watchers for image %s: %w", rv, err)
	}

	// openReadOnly created our own watcher, so baseline is 1
	defaultWatchers := 1

	mirrorInfo, err := image.GetMirrorImageInfo()
	if err == nil && mirrorInfo.Primary {
		count, mErr := util.GetRBDMirrorDaemonCount(util.CsiConfigFile, rv.ClusterID)
		if mErr == nil {
			defaultWatchers += count
		}
	}

	// If there are more watchers than our own (+ mirror daemon),
	// the kernel mapping's watcher is still alive.
	return len(watchers) > defaultWatchers, nil
}


// visualization
blockNodeGetVolumeStats(volumeId)
│
├── DecomposeCSIID failed?  ──→  getBlockMetrics() (size only, no health)
├── No secretRef configured? ──→  getBlockMetrics() (size only, no health)
│
├── k8s.GetSecret failed?   ──→  Abnormal=true  "failed to get secret"
├── NewAdminCredentials fail?──→  Abnormal=true  "failed to create credentials"
├── GenVolFromVolID failed?  ──→  Abnormal=true  "failed to connect to rbd image"
│                                 (monitors unreachable, image deleted, etc.)
│
├── rv.isMappingAlive()
│   ├── error?               ──→  log warning, continue (don't declare unhealthy)
│   └── alive=false?         ──→  Abnormal=true  "mapping watcher not found"
│                                 (kernel session expired/blocklisted)
│
├── rv.getUsedBytes() failed?──→  Abnormal=true  "failed to get used bytes"
│                                 (OSD-level problem)
│
└── everything succeeded     ──→  Abnormal=false  "volume is in a healthy condition"
                                  (+ full usage stats)
```

Does this align with how you would approach this in the context of this codebase ?? I'd appreciate your feedback/suggestions @nixpanic.

nixpanic: Block-mode volumes are difficult. This PR can focus on Filesystem-mode volumes only, that is fine. Block-mode volumes need a lot more work, including a design that accounts for backwards compatibility.

---

A general approach that would be optimal, takes an approach like this:

- have a _hidden_ part on the volume that can be used for health checking
  - on filesystem volumes, expose a subdir of the volume to apps and an other subdir where we can write/read a little data
  - on block volumes, add a partition 0, and map partition 1 to apps, use partition 0 for write/read health check data
- in the hidden location, write a timestamp and read it back in intervals, detect failures and delays

Now, adding such a hidden location on existing volumes would be difficult, it is probably not something we want to try. Only new volumes will therefore have the hidden location and the more advanced health checking.

Volume expansion, cloning, snapshotting and all will need to be aware of the hidden area. This is not a trivial amount of work.

internal/rbd/nodeserver.go
```diff
@@ -377,6 +384,8 @@ func (ns *NodeServer) NodeStageVolume(
 		} else if !isNotMnt {
 			log.DebugLog(ctx, "rbd: volume %s is already mounted to %s, skipping", volID, stagingTargetPath)
 
+			ns.startSharedHealthChecker(ctx, volID, stagingTargetPath)
```
nixpanic: Don't start the health-checker when the volume-mode is Block.

One way to prevent that, is by adding the volume-mode as parameter to `startSharedHealthChecker()` and conditionally select the `StatCheckerType` for filesystem mode, with no checker to start otherwise. This makes it easy to extend later on for Block mode.

SanjalKatiyar: @nixpanic initially, I was doing that, but the issue is that even if we block the shared checker in the `NodeStageVolume` RPC, `NodeGetVolumeStats` will still end up starting a non-shared checker later. Right now we use a synchronous `os.Stat` call to determine whether the volume is block-mode or filesystem-mode, and that call is what hangs when the volume is unhealthy.

For now, though, as you suggested, I will update and at least prevent the shared checker from being started.

SanjalKatiyar: But, in order to extend later, we will need a better way for `NodeGetVolumeStats` to determine the volume mode without relying on `os.Stat`.

nixpanic: Yes. When we have a Block-mode health-checker we can start it in `startSharedHealthChecker()` too. Preparing for that, and logging a warning that Block-mode is not supported would be nice.

Filesystem-mode or Block-mode is part of the CSI Volume specification. There are statements in `nodeserver.go` like these already:
```
isBlock := req.GetVolumeCapability().GetBlock() != nil
```

SanjalKatiyar: updated...

internal/rbd/nodeserver.go
```diff
@@ -1665,6 +1700,60 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+	if healthy && msg != nil {
+		// Start a StatChecker for the mounted targetPath, this prevents
+		// writing a file in the user-visible location. Ideally a (shared)
+		// FileChecker is started with the stagingTargetPath, but we can't
+		// get the stagingPath from the request easily.
+		// TODO: resolve the stagingPath like rbd.getStagingPath() does
+		// NOTE: rbd.getStagingPath() uses os.Stat() internally which
+		// if called synchronously, could block indefinitely.
+
+		// TODO: Add health checker for block-mode volumes.
```
nixpanic: I think we always run into this on block-mode volumes?

Maybe we can introduce a no-op health checker, that just always returns healthy with a message?

SanjalKatiyar: @nixpanic actually this is what I was concerned about here in this comment: https://github.com/ceph/ceph-csi/pull/6382#discussion_r3681430310.

We determine whether the volume is block-mode or filesystem-mode later in the flow, using `os.Stat` (synchronously).
At this point in the flow, we are still determining the volume's health and don't yet know its mode, so we can't rely on `os.Stat` for that info. We need a better way to determine the volume mode earlier in the flow, right ?? Which is why I left this as a `TODO` comment.

nixpanic: `os.Stat()` will always work on block-mode volumes. The information that is returned is about the local device node (`/dev/rdbX`) and does not trigger any I/O towards the Ceph cluster.

For filesystem-mode volumes, `os.Stat()` will not be reached if the volume is unhealthy (baring races between checking in the background and `NodeGetVolumeStat` calls).

Starting a No-Op checker for block-mode should not make the situation any worse than that it is now? But a No-Op checker should always return `true, nil` so that it isn't getting started on every `NodeGetVolumeStat` request.

SanjalKatiyar: My doubt is how would we know whether to start a no-op checker or a stat checker ??

Ex: `err = ns.healthChecker.StartChecker(volumeId, targetPath, hc.StatCheckerType)`

Here we need to pass an argument `hc.NoOpCheckerType` and in order to do that we need to know if the volume is block-mode or filesystem-mode. We can't do synchronous `os.Stat` yet, both filesystem and block volumes shares the same flow, so even if `os.Stat` works for block, it will hang for filesystem.

nixpanic: right, that is indeed a good point.

This code is expected to only get executed when the node-plugin restarted while a volume was already attached (no health-checker running yet). In this case, I think it it acceptable to call `os.Stat()` to figure out what type of volume it is. There is an unlikely/rare case where the volume is not healthy and the node-plugin is restarted, we can live with this potential hang.

SanjalKatiyar: I may be being overly cautious, but I have a strong feeling that QE will test this scenario and flag the issue.

I tested it twice and encountered the issue once (which is 50% of time :P), though I completely understand that two test runs aren't enough to draw any meaningful conclusions about how frequently it occurs, but the reason I am concerned is that we have the intentional `SlowGRPCRestart` flow, which can restart our container. I added more details about the conditions under which I reproduced it in comment: https://github.com/ceph/ceph-csi/pull/6418#discussion_r3665402363 ("A related scenario that still exists" part).

Also, moving ahead with `os.Stat` early in this flow will cause the `checked` flag bug-fix introduced in PR: https://github.com/ceph/ceph-csi/pull/6418 essentially not useful for RBD.

This IMHO could be worse than the current PR where the fallback starts a background `StatChecker` for RBD block as well and returns immediately.

That said, I may be overthinking this, plz let me know, I am happy to proceed with the no-op suggested flow update.

SanjalKatiyar: We can do one more thing, since we can determine the mode during `NodeStageVolume`, we can start a shared no-op checker at that point.

For `NodeGetVolumeStats`, we can keep it as currently proposed in this PR, that is, always a non-shared `StatChecker` for now (as u pointed out that this flow will be triggered less often) to avoid performing an early `os.Stat` check, which could potentially hang (though, I know it might look like introducing several workarounds that make the code more complex and reduce its overall readability).

nixpanic: If there is a real issue on the Ceph cluster, Ceph-CSI won't work correctly in that case anyway (not only `NodeGetVolumeStat`, but any CSI procedure). `SlowGRPCRestart` causes a complete cleanup/retry of everything, I don't think we need to care about it specifically.

There is an alternative where you can store/retrieve the mode of the volume. Extend `stashRBDImageMetadata()` and `struct rbdImageMetadataStash`. Those details are stored during `NodeStageVolume` on a host-path, and can be retrieved after a restart of the node-plugin. This introduces a little more work, but that way you can start a selected health-checker without calling `os.Stat()`.

SanjalKatiyar: @nixpanic thanks for the suggestion. I checked, but it won't solve our problem, issue is currently metadata is stashed at "stagingPath", but NodeGetVolumeStats won't have access to "stagingPath" as easily as expected.
I know we have `getStagingPath` util already,  but that function internally uses `os.Stat` itself, so we are looping back to the same issue again which we were discussing above :)

If it's okay, I will stash the volume mode at new path indexed at "volumeID" ??

nixpanic: using the volume-id makes sense to me. But what would the full path be then? The staging path is a known location that can be used, not sure what other location suits the purpose.

It all gets quite tricky... Maybe get a simple approach (with known issues) in with this PR, and improve with a next one?

SanjalKatiyar: There are now 3 commits:
1. no-op checker (cherry-picked from your fork).
2. updating rbd server to use no-op checker for block based on os.Stat, with a comment about known bug.
3. fixing the known bug via stashing the mode on the host, indexed by volumeID.

I still need to test the changes, will let you know once done.

SanjalKatiyar:
Tested.

##### Created file on the host and the local pod mount:
Screenshot 2026-08-02 at 2 13 30 PM
Screenshot 2026-08-02 at 2 17 52 PM

##### Logs of `openshift-storage.rbd.csi.ceph.com-nodeplugin-<HASH>` pod (no slow RPC calls or hang):
```
I0802 08:49:16.279006  108501 cephcsi.go:213] Driver version: canary and Git version: 1b63191a386b397f9421eb579cd5c5915825d9c6
I0802 08:49:16.279176  108501 cephcsi.go:315] Initial PID limit is set to -1
I0802 08:49:16.279250  108501 cephcsi.go:321] Reconfigured PID limit to -1 (max)
I0802 08:49:16.279463  108501 cephcsi.go:267] Starting driver type: rbd with name: openshift-storage.rbd.csi.ceph.com
I0802 08:49:16.290667  108501 mount_linux.go:323] Detected umount with safe 'not mounted' behavior
I0802 08:49:16.290792  108501 rbd_attach.go:248] nbd module loaded
I0802 08:49:16.290816  108501 rbd_attach.go:262] kernel version "5.14.0-687.33.1.el9_8.x86_64" supports cookie feature
I0802 08:49:16.312250  108501 rbd_attach.go:278] rbd-nbd tool supports cookie feature
I0802 08:49:16.312555  108501 server.go:123] listening for CSI-Addons requests on address: &net.UnixAddr{Name:"/csi/csi-addons.sock", Net:"unix"}
I0802 08:49:16.312772  108501 server.go:131] Listening for connections on address: &net.UnixAddr{Name:"//csi/csi.sock", Net:"unix"}
I0802 08:49:16.655884  108501 utils.go:361] ID: 1 GRPC call: /csi.v1.Identity/GetPluginInfo
I0802 08:49:16.655908  108501 utils.go:362] ID: 1 GRPC request: {}
I0802 08:49:16.655922  108501 identityserver-default.go:40] ID: 1 Using default GetPluginInfo
I0802 08:49:16.656044  108501 utils.go:368] ID: 1 GRPC response: {"name":"openshift-storage.rbd.csi.ceph.com","vendor_version":"canary"}
I0802 08:49:16.751945  108501 utils.go:361] ID: 2 GRPC call: /csi.v1.Node/NodeGetInfo
I0802 08:49:16.751973  108501 utils.go:362] ID: 2 GRPC request: {}
I0802 08:49:16.751994  108501 nodeserver-default.go:46] ID: 2 Using default NodeGetInfo
I0802 08:49:16.752084  108501 utils.go:368] ID: 2 GRPC response: {"accessible_topology":{},"node_id":"ip-10-0-71-36.us-west-2.compute.internal"}
I0802 08:50:08.894398  108501 utils.go:361] ID: 3 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:50:08.894444  108501 utils.go:362] ID: 3 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454","volume_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/openshift-storage.rbd.csi.ceph.com/ecf392db30c61dc78e5b60fccca9bd4330dd75f73363b162dc2b665fccd0bf86/globalmount/0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454"}
I0802 08:50:08.894515  108501 nodeserver.go:1745] ID: 3 could not lookup volume info for 0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454: open /csi/volume-info/0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454.json: no such file or directory
I0802 08:50:08.894599  108501 utils.go:368] ID: 3 GRPC response: {"volume_condition":{"message":"health checker started, status not yet available"}}
I0802 08:50:14.952400  108501 utils.go:361] ID: 4 GRPC call: /csi.v1.Node/NodeGetCapabilities
I0802 08:50:14.952424  108501 utils.go:362] ID: 4 GRPC request: {}
I0802 08:50:14.952604  108501 utils.go:368] ID: 4 GRPC response: {"capabilities":[{"rpc":{"type":"STAGE_UNSTAGE_VOLUME"}},{"rpc":{"type":"GET_VOLUME_STATS"}},{"rpc":{"type":"VOLUME_CONDITION"}},{"rpc":{"type":"EXPAND_VOLUME"}},{"rpc":{"type":"SINGLE_NODE_MULTI_WRITER"}}]}
I0802 08:50:14.953399  108501 utils.go:361] ID: 5 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:50:14.953431  108501 utils.go:362] ID: 5 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454","volume_path":"/var/lib/kubelet/pods/55e30f50-0986-4b82-a97f-91cb3d83e63f/volumes/kubernetes.io~csi/pvc-5cac6194-3198-4ae7-9a68-bf259258c976/mount"}
I0802 08:50:14.953488  108501 nodeserver.go:1745] ID: 5 could not lookup volume info for 0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454: open /csi/volume-info/0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454.json: no such file or directory
I0802 08:50:14.953520  108501 utils.go:368] ID: 5 GRPC response: {"volume_condition":{"message":"health checker started, status not yet available"}}
I0802 08:50:55.840725  108501 utils.go:361] ID: 6 GRPC call: /csi.v1.Node/NodeGetCapabilities
I0802 08:50:55.840762  108501 utils.go:362] ID: 6 GRPC request: {}
I0802 08:50:55.841062  108501 utils.go:368] ID: 6 GRPC response: {"capabilities":[{"rpc":{"type":"STAGE_UNSTAGE_VOLUME"}},{"rpc":{"type":"GET_VOLUME_STATS"}},{"rpc":{"type":"VOLUME_CONDITION"}},{"rpc":{"type":"EXPAND_VOLUME"}},{"rpc":{"type":"SINGLE_NODE_MULTI_WRITER"}}]}
I0802 08:50:55.841959  108501 utils.go:361] ID: 7 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:50:55.841990  108501 utils.go:362] ID: 7 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-a29e00ef-ff6d-4059-b3a1-25e3be84a517","volume_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/pvc-9274aea1-af9a-4dca-b190-6f0175cbe202/dev/b3daa4d4-f8db-4d42-9734-391b2496a370"}
I0802 08:50:55.842138  108501 utils.go:368] ID: 7 GRPC response: {"volume_condition":{"message":"health checker started, status not yet available"}}
I0802 08:51:08.878318  108501 utils.go:361] ID: 8 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:51:08.878370  108501 utils.go:362] ID: 8 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454","volume_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/openshift-storage.rbd.csi.ceph.com/ecf392db30c61dc78e5b60fccca9bd4330dd75f73363b162dc2b665fccd0bf86/globalmount/0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454"}
I0802 08:51:08.878433  108501 nodeserver.go:1745] ID: 8 could not lookup volume info for 0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454: open /csi/volume-info/0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454.json: no such file or directory
I0802 08:51:08.878462  108501 utils.go:368] ID: 8 GRPC response: {"volume_condition":{"message":"health checker started, status not yet available"}}
I0802 08:52:06.987120  108501 utils.go:361] ID: 9 GRPC call: /csi.v1.Node/NodeGetCapabilities
I0802 08:52:06.987148  108501 utils.go:362] ID: 9 GRPC request: {}
I0802 08:52:06.987226  108501 utils.go:368] ID: 9 GRPC response: {"capabilities":[{"rpc":{"type":"STAGE_UNSTAGE_VOLUME"}},{"rpc":{"type":"GET_VOLUME_STATS"}},{"rpc":{"type":"VOLUME_CONDITION"}},{"rpc":{"type":"EXPAND_VOLUME"}},{"rpc":{"type":"SINGLE_NODE_MULTI_WRITER"}}]}
I0802 08:52:06.987971  108501 utils.go:361] ID: 10 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:52:06.988055  108501 utils.go:362] ID: 10 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454","volume_path":"/var/lib/kubelet/pods/55e30f50-0986-4b82-a97f-91cb3d83e63f/volumes/kubernetes.io~csi/pvc-5cac6194-3198-4ae7-9a68-bf259258c976/mount"}
I0802 08:52:06.988286  108501 utils.go:368] ID: 10 GRPC response: {"usage":[{"available":52252286976,"total":52521566208,"unit":"BYTES","used":252502016},{"available":3275266,"total":3276800,"unit":"INODES","used":1534}],"volume_condition":{"message":"volume is in a healthy condition"}}
I0802 08:52:08.867891  108501 utils.go:361] ID: 11 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:52:08.867942  108501 utils.go:362] ID: 11 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454","volume_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/openshift-storage.rbd.csi.ceph.com/ecf392db30c61dc78e5b60fccca9bd4330dd75f73363b162dc2b665fccd0bf86/globalmount/0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454"}
I0802 08:52:08.868128  108501 utils.go:368] ID: 11 GRPC response: {"usage":[{"available":52252286976,"total":52521566208,"unit":"BYTES","used":252502016},{"available":3275266,"total":3276800,"unit":"INODES","used":1534}],"volume_condition":{"message":"volume is in a healthy condition"}}
I0802 08:52:32.922073  108501 utils.go:361] ID: 12 GRPC call: /csi.v1.Node/NodeGetCapabilities
I0802 08:52:32.922099  108501 utils.go:362] ID: 12 GRPC request: {}
I0802 08:52:32.922156  108501 utils.go:368] ID: 12 GRPC response: {"capabilities":[{"rpc":{"type":"STAGE_UNSTAGE_VOLUME"}},{"rpc":{"type":"GET_VOLUME_STATS"}},{"rpc":{"type":"VOLUME_CONDITION"}},{"rpc":{"type":"EXPAND_VOLUME"}},{"rpc":{"type":"SINGLE_NODE_MULTI_WRITER"}}]}
I0802 08:52:32.923034  108501 utils.go:361] ID: 13 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:52:32.923062  108501 utils.go:362] ID: 13 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-a29e00ef-ff6d-4059-b3a1-25e3be84a517","volume_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/pvc-9274aea1-af9a-4dca-b190-6f0175cbe202/dev/b3daa4d4-f8db-4d42-9734-391b2496a370"}
I0802 08:52:32.950585  108501 omap.go:87] ID: 13 got omap values by keys: (pool="ocs-storagecluster-cephblockpool", namespace="", name="csi.volume.a29e00ef-ff6d-4059-b3a1-25e3be84a517"): map[csi.imageid:3ba12d80e30c csi.imagename:csi-vol-a29e00ef-ff6d-4059-b3a1-25e3be84a517 csi.volname:pvc-9274aea1-af9a-4dca-b190-6f0175cbe202 csi.volume.owner:volume-health-test]
I0802 08:52:32.975158  108501 rbd_util.go:2492] ID: 13 rbd: running DiffIterate over image ocs-storagecluster-cephblockpool/csi-vol-a29e00ef-ff6d-4059-b3a1-25e3be84a517 of size 1073741824 bytes
I0802 08:52:32.981172  108501 rbd_util.go:2513] ID: 13 DiffIterate on image ocs-storagecluster-cephblockpool/csi-vol-a29e00ef-ff6d-4059-b3a1-25e3be84a517 of size 1073741824 bytes took 0.005984379 seconds
I0802 08:52:32.981600  108501 utils.go:368] ID: 13 GRPC response: {"usage":[{"available":1073741824,"total":1073741824,"unit":"BYTES"}],"volume_condition":{"message":"block-mode health checking is not supported"}}
I0802 08:53:08.883498  108501 utils.go:361] ID: 14 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:53:08.883548  108501 utils.go:362] ID: 14 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454","volume_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/openshift-storage.rbd.csi.ceph.com/ecf392db30c61dc78e5b60fccca9bd4330dd75f73363b162dc2b665fccd0bf86/globalmount/0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454"}
I0802 08:53:08.883753  108501 utils.go:368] ID: 14 GRPC response: {"usage":[{"available":52235509760,"total":52521566208,"unit":"BYTES","used":269279232},{"available":3275264,"total":3276800,"unit":"INODES","used":1536}],"volume_condition":{"message":"volume is in a healthy condition"}}
I0802 08:53:09.909335  108501 utils.go:361] ID: 15 GRPC call: /csi.v1.Node/NodeGetCapabilities
I0802 08:53:09.909366  108501 utils.go:362] ID: 15 GRPC request: {}
I0802 08:53:09.909442  108501 utils.go:368] ID: 15 GRPC response: {"capabilities":[{"rpc":{"type":"STAGE_UNSTAGE_VOLUME"}},{"rpc":{"type":"GET_VOLUME_STATS"}},{"rpc":{"type":"VOLUME_CONDITION"}},{"rpc":{"type":"EXPAND_VOLUME"}},{"rpc":{"type":"SINGLE_NODE_MULTI_WRITER"}}]}
I0802 08:53:09.910230  108501 utils.go:361] ID: 16 GRPC call: /csi.v1.Node/NodeGetVolumeStats
I0802 08:53:09.910257  108501 utils.go:362] ID: 16 GRPC request: {"volume_id":"0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454","volume_path":"/var/lib/kubelet/pods/55e30f50-0986-4b82-a97f-91cb3d83e63f/volumes/kubernetes.io~csi/pvc-5cac6194-3198-4ae7-9a68-bf259258c976/mount"}
I0802 08:53:09.910415  108501 utils.go:368] ID: 16 GRPC response: {"usage":[{"available":52235509760,"total":52521566208,"unit":"BYTES","used":269279232},{"available":3275264,"total":3276800,"unit":"INODES","used":1536}],"volume_condition":{"message":"volume is in a healthy condition"}}
```
**Note:**
- Volume `0001-0011-openshift-storage-0000000000000002-8f42714d-ec11-4adb-8a30-84b4a6760454` is using filesystem mode and was created **before** I updated the plugin image with the changes introduced in this PR. That's why we see the `"could not lookup volume info"` log. The `NodeStageVolume` RPC had already run for this volume before my changes took effect.
This is an expected/known side-effect/limitation of stashing volume info on the host. Existing volumes won't have this metadata, while newly staged volumes will include it by default.
That said, IMHO this is acceptable. The next time Kubelet invokes `NodeStageVolume` for an existing volume (eg: after an OCP upgrade, node reboot, or any other event that requires the volume to be staged again), the metadata will be populated and this will work for those volumes as well.
The overall impact should be minimal. If nodeplugin restarts are expected to be infrequent (as per discussion here: https://github.com/ceph/ceph-csi/pull/6382#discussion_r3689241332), so the shared checker will already handle the majority of cases. This non-shared checker fallback path is only executed in specific scenarios. Even if we cannot determine the volume mode, it defaults to the stat checker, which should not introduce any regression, runtime error, or pod panic.

- Volume `0001-0011-openshift-storage-0000000000000002-a29e00ef-ff6d-4059-b3a1-25e3be84a517` is using block mode and was created **after** I updated the plugin image with the changes introduced in this PR.
It works as expected. It reads the volume mode from the host and calls no-op checker. Plugin restarts scenario works smoothly as well, no slow RPCs or abort signal logs.

internal/rbd/nodeserver.go
SanjalKatiyar: @nixpanic I tested, plz check: https://github.com/ceph/ceph-csi/pull/6382#issuecomment-5156871825

nixpanic (approved):
looks good to me, thanks!

internal/rbd/rbd_util.go
```diff
@@ -2031,6 +2028,63 @@ func cleanupRBDImageMetadataStash(metaDataPath string) error {
...
+// Holds per-volume metadata that needs to persist across
+// node-plugin restarts, indexed by volumeID.
+type volumeInfo struct {
+	IsBlockMode bool `json:"isBlockMode"`
+}
+
+// This function is different from "stashRBDImageMetadata" which stores info at stagingPath.
+// Not all RPCs have access to the stagingPath. For example,
+// NodeGetVolumeStats uses this to determine the volume mode using volumeID as the index.
+func stashVolumeInfo(volumeID string, info *volumeInfo) error {
```
Madhu-1, line 2048: can we have unit test for these functions. and in which exact path we are storing the VolumeInfo?

SanjalKatiyar: > which exact path we are storing the VolumeInfo

I added a comment regarding this earlier:
```
// The "/csi" directory inside the container is backed by the
// host-path volume "socket-dir" (ref rbd-plugin DaemonSet yaml file)
// which maps to "/var/lib/kubelet/plugins/rbd.csi.ceph.com" on the node.
```
So, exact path is `/var/lib/kubelet/plugins/openshift-storage.rbd.csi.ceph.com/volume-info/<volumeID>.json`.

> can we have unit test for these functions

@Madhu-1 other than this comment, if this PR looks good, can I add unit tests in a follow-up instead ??
Else, if there are any other major changes needed, I can update this same PR.

Madhu-1: This directory is managed by kubelet and we need to see if it causes some problem

Madhu-1 (approved):

internal/rbd/rbd_util.go
SanjalKatiyar: It worked as expected when I tested it: https://github.com/ceph/ceph-csi/pull/6382#issuecomment-5156871825.

But, if there is a specific scenario or use case you like to verify, let me know. I can test it or maybe ask QE to validate it.

internal/rbd/nodeserver.go
```diff
@@ -1665,6 +1720,63 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+		// writing a file in the user-visible location. Ideally a (shared)
+		// FileChecker is started with the stagingTargetPath, but we can't
+		// get the stagingPath from the request easily.
+		// TODO: resolve the stagingPath like rbd.getStagingPath() does
+		// NOTE: rbd.getStagingPath() uses os.Stat() internally which
+		// if called synchronously, could block indefinitely.
+
+		var checkerType hc.CheckerType = hc.StatCheckerType
+		info, lookupErr := lookupVolumeInfo(volumeId)
+		if lookupErr == nil && info.IsBlockMode {
```
Madhu-1, line 1742: What will happen for upgraded clusters as this file will be missing

SanjalKatiyar: Yeah, I noticed this while testing as well (`Note` part mentioned here https://github.com/ceph/ceph-csi/pull/6382#issuecomment-5156871825) :/

> This is an expected/known side-effect/limitation of stashing volume info on the host. Existing volumes won't have this metadata, while newly staged volumes will include it by default.
> That said, IMHO this is acceptable. The next time Kubelet invokes NodeStageVolume for an existing volume (eg: after an OCP upgrade, node reboot, or any other event that requires the volume to be staged again), the metadata will be populated and this will work for those volumes as well.
> The overall impact should be minimal. If nodeplugin restarts are expected to be infrequent (as per discussion here: https://github.com/ceph/ceph-csi/pull/6382#discussion_r3689241332), so the shared checker will already handle the majority of cases. This non-shared checker fallback path is only executed in specific scenarios. Even if we cannot determine the volume mode, it defaults to the stat checker, which should not introduce any regression, runtime error, or pod panic.

internal/rbd/nodeserver.go
```diff
@@ -1263,6 +1295,8 @@ func (ns *NodeServer) NodeUnstageVolume(
 
 	volID := req.GetVolumeId()
 
+	ns.healthChecker.StopSharedChecker(volID)
```
iPraveenParihar, line 1298: Should be called after acquiring lock

internal/health-checker/noopchecker.go
```diff
@@ -38,3 +37,11 @@ func newNoOpChecker() ConditionChecker {
 
 	return nc
 }
+
+// always healthy, no errors
+// overrides checker.isHealthy() so that the timeout-based
+// staleness check (which compares lastUpdate against interval+timeout)
+// never marks a NoOp checker unhealthy
+func (nc *noOpChecker) isHealthy() (bool, error) {
```
iPraveenParihar, line 45: This change should be part of previous commit.

internal/health-checker/noopchecker.go
```diff
@@ -0,0 +1,40 @@
...
+type noOpChecker struct {
+	checker
+}
+
+func newNoOpChecker() ConditionChecker {
+	nc := &noOpChecker{}
+	nc.initDefaults()
+
+	// always healthy, no errors
+	nc.healthy = true
```
iPraveenParihar, line 28: not required, since already initialised in initDefaults()?

internal/rbd/driver/driver.go
```diff
@@ -82,6 +83,7 @@ func NewNodeServer(
 		DefaultNodeServer: csicommon.NewDefaultNodeServer(d, t, cliReadAffinityMapOptions, topology, nodeLabels),
 		VolumeLocks:       util.NewIDLocker(),
 	}
+	ns.SetHealthChecker(hc.NewHealthCheckManager())
```
iPraveenParihar, line 86: Can you move this to NodeServer struct similar to CephFS NewNodeServer to maintain consistency?

internal/rbd/nodeserver.go
```diff
@@ -1665,6 +1720,63 @@ func (ns *NodeServer) NodeGetVolumeStats(
...
+		// TODO: resolve the stagingPath like rbd.getStagingPath() does
+		// NOTE: rbd.getStagingPath() uses os.Stat() internally which
+		// if called synchronously, could block indefinitely.
+
+		var checkerType hc.CheckerType = hc.StatCheckerType
+		info, lookupErr := lookupVolumeInfo(volumeId)
+		if lookupErr == nil && info.IsBlockMode {
+			checkerType = hc.NoOpCheckerType
+		} else if lookupErr != nil {
+			log.DebugLog(ctx, "could not lookup volume info for %s: %v", volumeId, lookupErr)
```
iPraveenParihar, line 1745: Should be WarningLog

iPraveenParihar (dismissed):

internal/rbd/nodeserver.go
SanjalKatiyar: I doubt so.

Checker is internally thread safe, AFAIU lock in these RPCs are mainly for mount/filesystem related operations.

Even CephFS does the same during un-stage.

internal/health-checker/noopchecker.go
SanjalKatiyar: Maybe or Maybe Not :)

Way I see it, that commit is a cherry-pick from Niels' fork and follow-up commits are adjustments/fixes on top of that for the rbd flow.

internal/health-checker/noopchecker.go
SanjalKatiyar: defaults can change... adding it here is an explicit indication of the desired behaviour...

internal/rbd/driver/driver.go
SanjalKatiyar: It can't be same as CephFS (unless we do some refactoring), flow is already "inconsistent".

nodeserver and driver are different packages in rbd, whereas that's not the case in cephfs.

internal/rbd/nodeserver.go
SanjalKatiyar: Make sense.

But, this PR was already in the merge queue, it's not ideal to push new commit now and go through approvals and CI re-runs from scratch again, especially since today is deadline for merging this.

I will send a follow-up for unit tests, I will update this in that PR.

internal/health-checker/noopchecker.go
SanjalKatiyar: check: https://github.com/ceph/ceph-csi/pull/6382#discussion_r3695757222

PR #6390, opened 2026-07-08
cephfs: support MDS pinning via ControllerModifyVolume

### LINKED ISSUE #6378: cephfs: support MDS pinning of subvolumes via VolumeAttributesClass

Implement the CSI `ControllerModifyVolume` RPC in the CephFS driver to allow
pinning a CephFS subvolume to MDS ranks through the Kubernetes
`VolumeAttributesClass` API.

The feature exposes two mutable parameters on a `VolumeAttributesClass`:

- `mds-pin-type`: the pin policy — one of `export`, `distributed`, or `random`.
- `mds-pin-setting`: the value for the pin (e.g. an MDS rank `"2"` for `export`,
  `"1"`/`"0"` for `distributed`, or a float `"0.0"`–`"1.0"` for `random`).

This is the CSI-native equivalent of:

```console
ceph fs subvolume pin <fs> <subvolume> <pin_type> <pin_setting>
```

A PVC can be pinned at creation time or re-pinned at runtime by
setting/patching its `volumeAttributesClassName`.

In multi-MDS CephFS deployments, Ceph balances metadata subtrees across active
MDS ranks dynamically. Today Ceph-CSI users have no Kubernetes-native way to
influence this and must run `ceph` commands out of band, which is error-prone
and disconnected from the volume lifecycle.

With this feature, end users can, declaratively from Kubernetes:

- Improve performance for latency-sensitive workloads via MDS affinity to a
  specific rank.
- Isolate specific namespaces/tenants to dedicated MDS ranks.
- Explicitly control metadata load distribution (e.g. after scaling MDS
  ranks) — and change it at runtime by patching the PVC's
  `volumeAttributesClassName`.

The concerns that closed #2637 no longer apply, and #4662 signals maintainer
interest in exactly this `ControllerModifyVolume` mechanism.

TBD

# Additional context #

- Supersedes #2637; concrete case of #4662.
- Ceph pinning reference:
  https://docs.ceph.com/en/latest/cephfs/multimds/#cephfs-pinning
- Kubernetes `VolumeAttributesClass` (GA in v1.34):
  https://kubernetes.io/blog/2025/09/08/kubernetes-v1-34-volume-attributes-class

ismaelpuerto:
Expose the RPC_MODIFY_VOLUME capability on the CephFS controller and map the mutable VolumeAttributesClass parameters mds-pin-export, mds-pin-distributed and mds-pin-random to "ceph fs subvolume pin". This lets administrators change the MDS distribution policy of an existing subvolume at runtime without recreating the volume.

The three pin parameters are mutually exclusive and validated before being applied. A new PinVolume method on the SubVolumeClient wraps go-ceph's PinSubVolume; since go-ceph does not expose a subvolume-group aware variant, the pin is applied against Ceph's default subvolume group, which is documented as a known limitation.

Adds unit tests for the parameter validation, an example VolumeAttributesClass and documentation in the CephFS deployment guide.

Do you have any questions?

For example:

## Related issues ##

Fixes: #6378
Depends-on: #6458

## Future concerns ##

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

### DIFF
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -2740,6 +2740,99 @@ var _ = Describe(cephfsType, func() {
 			validateOmapCount(f, 0, cephfsType, metadataPool, snapsType)
 		})
 
+		It("create a PVC with a VolumeAttributesClass for MDS pinning and modify it", func() {
+			// VolumeAttributesClass support requires Kubernetes 1.34+.
+			if !k8sVersionGreaterEquals(f.ClientSet, 1, 34) {
+				framework.Logf("skipping VolumeAttributesClass test, needs Kubernetes >= 1.34")
+
+				return
+			}
+
+			// This test validates VolumeAttributesClass (VAC) support for
+			// CephFS subvolumes via MDS pinning, exercising both the
+			// CreateVolume and ControllerModifyVolume code paths:
+			//   1. Create a StorageClass.
+			//   2. Create two VACs with different mds-pin-export ranks.
+			//   3. Create a PVC referencing the first VAC and verify the PV
+			//      gets the VAC applied at creation time.
+			//   4. Update the PVC to reference the second VAC and verify the
+			//      change is applied via ControllerModifyVolume.
+			//   5. Clean up.
+			err := createCephfsStorageClass(f.ClientSet, f, true, nil)
+			if err != nil {
+				logAndFail("failed to create CephFS storageclass: %v", err)
+			}
+			defer func() {
+				err = deleteResource(cephFSExamplePath + "storageclass.yaml")
+				if err != nil {
+					logAndFail("failed to delete CephFS storageclass: %v", err)
+				}
+			}()
+
+			vacName1 := "e2e-" + f.UniqueName + "-cephfs-vac1"
+			vacName2 := "e2e-" + f.UniqueName + "-cephfs-vac2"
+
+			err = createCephFSVolumeAttributesClass(f.ClientSet, vacName1,
+				map[string]string{"mds-pin-export": "0"})
+			if err != nil {
+				logAndFail("failed to create VolumeAttributesClass: %v", err)
+			}
+			defer func() {
+				if err = deleteCephFSVolumeAttributesClass(f.ClientSet, vacName1); err != nil {
+					logAndFail("failed to delete VolumeAttributesClass: %v", err)
+				}
+			}()
+
+			err = createCephFSVolumeAttributesClass(f.ClientSet, vacName2,
+				map[string]string{"mds-pin-export": "1"})
+			if err != nil {
+				logAndFail("failed to create VolumeAttributesClass: %v", err)
+			}
+			defer func() {
+				if err = deleteCephFSVolumeAttributesClass(f.ClientSet, vacName2); err != nil {
+					logAndFail("failed to delete VolumeAttributesClass: %v", err)
+				}
+			}()
+
+			// create a PVC referencing the first VAC; this exercises the
+			// CreateVolume path applying the mutable parameters.
+			pvc, err := loadPVC(pvcPath)
+			if err != nil {
+				logAndFail("failed to load PVC: %v", err)
+			}
+			pvc.Namespace = f.UniqueName
+			pvc.Spec.VolumeAttributesClassName = &vacName1
+
+			err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+			if err != nil {
+				logAndFail("failed to create PVC: %v", err)
+			}
+
+			pv, err := getBoundPV(f.ClientSet, pvc)
+			if err != nil {
+				logAndFail("failed to get bound PV: %v", err)
+			}
+			if pv.Spec.VolumeAttributesClassName == nil ||
+				*pv.Spec.VolumeAttributesClassName != vacName1 {
+				logAndFail("PV %s does not reference the expected VolumeAttributesClass %q",
+					pv.Name, vacName1)
+			}
+			validateSubvolumeCount(f, 1, fileSystemName, subvolumegroup)
+
+			// update the PVC to reference the second VAC; this exercises the
+			// ControllerModifyVolume path.
+			err = modifyPVCVolumeAttributesClass(f.ClientSet, pvc, vacName2)
+			if err != nil {
+				logAndFail("failed to modify VolumeAttributesClass: %v", err)
+			}
+
+			err = deletePVCAndValidatePV(f.ClientSet, pvc, deployTimeout)
+			if err != nil {
+				logAndFail("failed to delete PVC: %v", err)
+			}
+			validateSubvolumeCount(f, 0, fileSystemName, subvolumegroup)
+		})
+
 		// Make sure this should be last testcase in
 		// this file, because it deletes pool
 		It("Create a PVC and delete PVC when backend pool deleted", func() {
--- a/internal/cephfs/controllerserver.go
+++ b/internal/cephfs/controllerserver.go
@@ -412,6 +412,16 @@ func (cs *cephfsControllerServer) CreateVolume(
 			}
 		}
 
+		// apply MutableParameters (e.g. MDS pinning) from a
+		// VolumeAttributesClass, if any were requested. The subvolume already
+		// existed before this request, so it must not be purged on failure.
+		if len(req.GetMutableParameters()) != 0 {
+			if err = applyMutableParameters(ctx, volClient, vID.FsSubvolName,
+				req.GetMutableParameters()); err != nil {
+				return nil, err
+			}
+		}
+
 		return buildCreateVolumeResponse(req, volOptions, vID), nil
 	}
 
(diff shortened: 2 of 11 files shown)


internal/cephfs/controllerserver.go
```diff
@@ -1464,3 +1465,174 @@ func (cs *cephfsControllerServer) fenceNode(
...
+//
+// Currently supported parameters (mutually exclusive, at most one may be set):
+//   - mds-pin-export:      pin to a specific MDS rank (e.g. "2")
+//   - mds-pin-distributed: distributed ephemeral pinning ("1" or "0")
+//   - mds-pin-random:      random ephemeral pinning, a float 0.0–1.0
+//
+// Similar To:
+//
+//	ceph fs subvolume pin <vol_name> <sub_name> <pin_type> <pin_setting>
+func (cs *cephfsControllerServer) ControllerModifyVolume(
```
nixpanic, line 1639: This needs to be called from `CreateVolume` as well, see `internal/nfs/controller/controllerserver.go` for an example. It may need a little work, like moving the actual functionality in a helper function, so that the locking and other steps are done only once.

nixpanic (commented):
Thanks for the PR!

I had a first view over this, and it looks reasonable to me.

The only thing that is missing, is a test case under `e2e/cephfs.go`. See how `e2e/nvmeof.go` uses a VolumeAttributesClass to apply a change

nixpanic:
The new e2e test "create a PVC with a VolumeAttributesClass for MDS pinning and modify it" failed. It looks like an issue related to a missing RBAC:

```
  I0710 08:36:29.354794   80190 pvc.go:86] PVC csi-cephfs-pvc Event: ProvisioningFailed - volumeattributesclasses.storage.k8s.io "e2e-cephfs-3728-cephfs-vac1" is forbidden: User "system:serviceaccount:cephcsi-e2e-c2e72d32:cephfs-csi-provisioner" cannot get resource "volumeattributesclasses" in API group "storage.k8s.io" at the cluster scope
```

[logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.37-cephfs/detail/mini-e2e_k8s-1.37-cephfs/4/pipeline)

In `deploy/cephfs/kubernetes/csi-provisioner-rbac.yaml` the permissions for VolumeAttributesClasses needs to be added. This is how RBD has them:

https://github.com/ceph/ceph-csi/blob/7e9580b3a39842f61e76aac666520ae74b062e92/deploy/rbd/kubernetes/csi-provisioner-rbac.yaml#L84-L86

Commit ef8b07a337423a9261b53ac77de87fbe2d2c2358 added the permissions to the Helm Chart for RBD. This needs to be done for the CephFS Helm Chart as well.

nixpanic (dismissed):
Thanks, this looks good to me.

The CI job failed on an unrelated NFS test, but the new CephFS test passed. ([logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.37-cephfs/detail/mini-e2e_k8s-1.37-cephfs/6/pipeline))

nixpanic:
@mergifyio rebase

internal/cephfs/core/volume.go
```diff
@@ -323,3 +329,41 @@ func checkSubvolumeHasFeature(feature string, subVolFeatures []string) bool {
...
 	return slices.Contains(subVolFeatures, feature)
 }
+
+// PinVolume sets an MDS pinning policy on the subvolume.
+// pinType can be "export", "distributed" or "random".
+// pinSetting is the value for the pin (e.g. MDS rank "2", "true"/"false").
+//
+// go-ceph's typed PinSubVolume helper does not accept a subvolume group, so it
+// would always target Ceph's default group and fail for subvolumes created in
+// a non-default group (as ceph-csi does). To support any group, the "fs
```
Madhu-1, line 342: @ismaelpuerto can we please fix this in go-ceph and open an issue to adopt to go-ceph API once its implemented?

Madhu-1 (dismissed):

internal/cephfs/controllerserver.go
```diff
@@ -485,6 +495,15 @@ func (cs *cephfsControllerServer) CreateVolume(
 	log.DebugLog(ctx, "cephfs: successfully created backing volume named %s for request name %s",
 		vID.FsSubvolName, requestName)
 
+	// apply MutableParameters (e.g. MDS pinning) from a VolumeAttributesClass,
+	// if any were requested.
+	if len(req.GetMutableParameters()) != 0 {
+		if err = applyMutableParameters(ctx, volClient, vID.FsSubvolName,
+			req.GetMutableParameters()); err != nil {
+			return nil, err
```
iPraveenParihar, line 511: if `applyMutableParameters` fail then we might need to clean the Subvolume orelse it will be orphaned. 
You need to call `PurgeVolume` here.

internal/cephfs/controllerserver.go
```diff
@@ -1464,3 +1483,195 @@ func (cs *cephfsControllerServer) fenceNode(
...
+func validateMDSPinParameters(params map[string]string) (string, string, error) {
+	var (
+		selected string
+		setting  string
+	)
+
+	for _, key := range mdsPinParams {
+		value, ok := params[key]
+		if !ok {
+			continue
```
iPraveenParihar, line 1547: This will silently ignore unknown mutable parameters. CSI spec recommends InvalidArgument for unsupported parameters.

internal/cephfs/controllerserver.go
```diff
@@ -1464,3 +1483,195 @@ func (cs *cephfsControllerServer) fenceNode(
...
+	defer cs.OperationLocks.ReleaseModifyLock(volID)
+
+	// resolve volume options — FsName, SubvolumeGroup and VolID are populated
+	// automatically from the cluster config, no extra input required.
+	volOptions, _, err := store.NewVolumeOptionsFromVolID(ctx, volID, nil, req.GetSecrets(),
+		cs.ClusterName)
+	if err != nil {
+		log.ErrorLog(ctx, "validation and extraction of volume options failed: %v", err)
+
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
iPraveenParihar: This should be `Not Found` error

internal/cephfs/core/volume.go
```diff
@@ -323,3 +329,41 @@ func checkSubvolumeHasFeature(feature string, subVolFeatures []string) bool {
...
+		"pin_type":    pinType,
+		"pin_setting": pinSetting,
+	}
+
+	buf, err := json.Marshal(cmd)
+	if err != nil {
+		return fmt.Errorf("failed to marshal pin command for subvolume %s: %w", s.VolID, err)
+	}
+
+	_, _, err = s.conn.MgrCommand([][]byte{buf})
```
iPraveenParihar: Why discard other return values? Could be useful to log if err == nil ?

internal/cephfs/core/volume.go
```diff
@@ -88,6 +89,11 @@ type SubVolumeClient interface {
 	UnsetAllMetadata(keys []string) error
 	// ListMetadata gets the metadata for the subvolume.
 	ListMetadata() (map[string]string, error)
+
+	// PinVolume sets an MDS pinning policy on the subvolume.
+	// pinType can be "export", "distributed" or "random".
+	// pinSetting is the value for the pin (e.g. MDS rank "2", "true", etc.).
```
iPraveenParihar: is `true` a valid value for any pin type?

e2e/cephfs.go
```diff
@@ -2740,6 +2740,99 @@ var _ = Describe(cephfsType, func() {
...
+			err = createPVCAndvalidatePV(f.ClientSet, pvc, deployTimeout)
+			if err != nil {
+				logAndFail("failed to create PVC: %v", err)
+			}
+
+			pv, err := getBoundPV(f.ClientSet, pvc)
+			if err != nil {
+				logAndFail("failed to get bound PV: %v", err)
+			}
+			if pv.Spec.VolumeAttributesClassName == nil ||
```
iPraveenParihar, line 2815: This only verifies the frontend part (validating the VAC reference in PV object), but doesn't validate on the backend if the pinning is applied on subvolume. Do you think that is possible to add coverage for this?

iPraveenParihar (changes requested):

e2e/cephfs.go
ismaelpuerto: Thanks for the review. I looked into validating the pin on the backend and wanted to share what I found, as it affects how we can assert it in e2e.

**1. `ceph fs subvolume info` does not expose the pin.** Tested on Ceph 19 (Squid)

**2. The pin is not visible from the application pod.** `ceph fs subvolume pin` applies `ceph.dir.pin` on the *subvolume* directory (`/volumes/csi/csi-vol-<id>/`), but CSI mounts the *data* subdirectory (`.../<uuid>/`) as the volume root. So `getfattr -n ceph.dir.pin` from the pod always reports `-1` even after a successful pin.

**3. The only way to assert the pin is to mount the whole filesystem** and read `ceph.dir.pin` on `/volumes/csi/csi-vol-<id>/`. Doable in the toolbox but adds mount setup/teardown and is more fragile than the rest of the CephFS e2e suite.

Would you prefer (a) a toolbox-side full-fs mount + `getfattr` verification on the subvolume directory, or keeping the current front-end assertion and documenting the backend-verification limitation?

internal/cephfs/core/volume.go
ismaelpuerto: Thanks for the review, I just create the issue : https://github.com/ceph/go-ceph/issues/1293

nixpanic:
Thanks again @ismaelpuerto! The code looks good to me. Can you please squash the last commit into the one that adds the new functionality (1st commit)? That makes the history of the commits cleaner and easier to follow.

e2e/cephfs.go
nixpanic: @iPraveenParihar , do you have an opinion on this?

I guess it would be nice to check the xattr, but mounting the subvolume directory (parent directory of the user visible volume) is a little tricky...

iPraveenParihar: Yea, lets keep it the current front-end assertion for now to not complex this PR. Let create a tracker to address backend assertion?

iPraveenParihar (approved):
LGTM
Thanks!

e2e/cephfs.go
ismaelpuerto: Sounds good, thanks! I've opened #6460 to track the backend assertion

nixpanic (approved):

nixpanic:
> /test ci/centos/mini-e2e/k8s-1.34

```
nfs Test NFS CSI [It] create a storageclass with relocated server and a PVC then bind it to an app
```

There might be a conflict with NFS ControllerModifyVolume. Mutable parameters passed by the NFS CreateVolume are also passed along to the CephFS CreateVolume call.

#6458 has created for this already. CI jobs probably pass once that PR is merged (and this PR is automatically rebased).

Rakshith-R:
@mergifyio queue

PR #6432, opened 2026-07-28
rbd: Fix connection leak when a non-nil volume is returned

black-dragon74:

This patch calls `Destroy` on the non-nil Volume/Snap object so that conn pool GC can clean it up.

Not doing this always kept ref count > 0, preventing cleanup and leading to lock contention, which in turn kicked in librbd lock breaking leading to blacklisting of actual clients.

### DIFF
--- a/internal/rbd/manager.go
+++ b/internal/rbd/manager.go
@@ -75,15 +75,13 @@ func (mgr *rbdManager) GetVolumeByID(ctx context.Context, id string) (types.Volu
 
 	volume, err := GenVolFromVolID(ctx, id, creds, mgr.secrets)
 	if err != nil {
+		volume.Destroy(ctx)
+
 		switch {
 		case errors.Is(err, rbderrors.ErrImageNotFound):
-			err = fmt.Errorf("volume %s not found: %w", id, err)
-
-			return nil, err
+			return nil, fmt.Errorf("volume %s not found: %w", id, err)
 		case errors.Is(err, util.ErrPoolNotFound):
-			err = fmt.Errorf("pool %s not found for %s: %w", volume.Pool, id, err)
-
-			return nil, err
+			return nil, fmt.Errorf("pool not found for %s: %w", id, err)
 		default:
 			return nil, fmt.Errorf("failed to get volume from id %q: %w", id, err)
 		}
@@ -100,17 +98,15 @@ func (mgr *rbdManager) GetSnapshotByID(ctx context.Context, id string) (types.Sn
 
 	snapshot, err := genSnapFromSnapID(ctx, id, creds, mgr.secrets)
 	if err != nil {
+		snapshot.Destroy(ctx)
+
 		switch {
 		case errors.Is(err, rbderrors.ErrImageNotFound):
-			err = fmt.Errorf("volume %s not found: %w", id, err)
-
-			return nil, err
+			return nil, fmt.Errorf("snapshot %s not found: %w", id, err)
 		case errors.Is(err, util.ErrPoolNotFound):
-			err = fmt.Errorf("pool %s not found for %s: %w", snapshot.Pool, id, err)
-
-			return nil, err
+			return nil, fmt.Errorf("pool not found for %s: %w", id, err)
 		default:
-			return nil, fmt.Errorf("failed to get volume from id %q: %w", id, err)
+			return nil, fmt.Errorf("failed to get snapshot from id %q: %w", id, err)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


nixpanic:
Do you know what kind of failure can trigger this problem?

This is probably a good candidate to backport to the supported release branches.

Rakshith-R (dismissed):

internal/rbd/manager.go
```diff
@@ -75,15 +75,15 @@ func (mgr *rbdManager) GetVolumeByID(ctx context.Context, id string) (types.Volu
 
 	volume, err := GenVolFromVolID(ctx, id, creds, mgr.secrets)
 	if err != nil {
+		if volume != nil {
```
Madhu-1: why do we need nil check on this one, because in error case we are accessing volume.Pool directly which means volume can never be nil and also GenVolFromVolID doesnt return nil for volume when there is a error

black-dragon74: It’s a safety net, `.Pool` is a string so that won’t panic. 

Existing callers follow the same pattern. So I added it here as well :)

Madhu-1: i mean if the volume is nil you cannot call .Pool on it. this looks unwanted check which always passes. do you want to keep it this way and clean it up in follow up PR?

black-dragon74: > GenVolFromVolID doesnt return nil for volume when there is a error

It does actually if `DecomposeCSIID` fails. I have removed the check and added it in `Destroy` instead.

Madhu-1 (dismissed):

internal/rbd/manager.go
Madhu-1: not for GenVolFromVolID but only for genSnapFromSnapID

internal/rbd/rbd_util.go
```diff
@@ -420,6 +420,9 @@ func (ri *rbdImage) Connect(cr *util.Credentials) error {
 // Destroy cleans up the rbdVolume and closes the connection to the Ceph
 // cluster in case one was setup.
 func (ri *rbdImage) Destroy(ctx context.Context) {
+	if ri == nil {
```
nixpanic, line 423: if you want to clean this up, make sure to address these calls everywhere. A quick glance suggests `internal/rbd/controllerserver.go` and `internal/rbd/nodeserver.go` have quite a few if-statements that can be removed too.

nixpanic (commented):
An other enhancement could be to keep the opened image around in the `rbdImage` struct. That way, fewer open calls are done, and performance might be better.

The `rbdImage` struct would need an image attribute for both the rw-image and ro-image, closing them on `Destroy()`. No `defer image.Close()` is then needed after each open anymore either.

Can be a separate PR, as it is a distinct thing from addressing the leak.

internal/rbd/rbd_util.go
black-dragon74: I'll follow this up with your other suggestion in a new PR :)

internal/rbd/manager.go
```diff
@@ -75,15 +75,13 @@ func (mgr *rbdManager) GetVolumeByID(ctx context.Context, id string) (types.Volu
...
 		case errors.Is(err, rbderrors.ErrImageNotFound):
-			err = fmt.Errorf("volume %s not found: %w", id, err)
-
-			return nil, err
+			return nil, fmt.Errorf("volume %s not found: %w", id, err)
 		case errors.Is(err, util.ErrPoolNotFound):
-			err = fmt.Errorf("pool %s not found for %s: %w", volume.Pool, id, err)
-
-			return nil, err
+			return nil, fmt.Errorf("pool %s not found for %s: %w", volume.Pool, id, err)
```
iPraveenParihar: `volume.Pool` is accessed after `Destroy()`

iPraveenParihar: wondering if error is `ErrPoolNotFound` won't volume.Pool will be empty too?

internal/rbd/manager.go
```diff
@@ -75,15 +75,13 @@ func (mgr *rbdManager) GetVolumeByID(ctx context.Context, id string) (types.Volu
 
 	volume, err := GenVolFromVolID(ctx, id, creds, mgr.secrets)
```
iPraveenParihar, line 76: @black-dragon74 Can you confirm if we need same fix at other places which don't call Destroy() on the error path?

https://github.com/ceph/ceph-csi/blob/b17a575d0c3054dc1a1b1d760e9089907c88e358/internal/rbd/controllerserver.go#L907-L914

internal/rbd/manager.go
black-dragon74: Fixed.

internal/rbd/manager.go
black-dragon74: Added Destroy at all places it was missing.

Madhu-1 (approved):

iPraveenParihar (approved):

Issue #6415, opened 2026-07-27
Remove `hostPID: true` from the nodeplugin pod spec


Jeansen:
I have a critical warning in my local playground cluster that reads like the following:

Sharing the host PID namespace lets the container see and signal all host processes, enabling process injection attacks.

K8s: Pod Security Standards
NSA/CISA Kubernetes Hardening Guide

DAEMONSET
rook-ceph / rook-ceph.rbd.csi.ceph.com-nodeplugin

Rook-Ceph version (Operator and Cluster): 1.19.7

For reference: I first opened it up here: https://github.com/rook/rook/issues/18032#issuecomment-5089145863

nixpanic:
`hostPID` is required for [Running CephCSI with pod networking](https://github.com/ceph/ceph-csi/blob/devel/examples/README.md#running-cephcsi-with-pod-networking). This feature requires access to the processes on the host (or an other container) so that `nsenter` can be used to run (mount)processes in a different Linux kernel namespace than that of the current container.

We've evaluated this a little more. The feature is not commonly used, and can be deprecated.

When setting `hostPID: false`, everything is expected to work as intended (except for the above mentioned feature). Because deploying Ceph-CSI is commonly done by [ceph-csi-operator](https://github.com/ceph/ceph-csi-operator), the `hostPID: true` setting should be removed there as well.

nixpanic:
Deprecation of node-plugin with Pod networking/restarts will be included in v3.18.0, with removal of the feature in v3.19.0.

Jeansen:
OK, great. Thank you. Should we keep this issu open then for reference?

nixpanic:
> OK, great. Thank you. Should we keep this issue open then for reference?

Yes, it has been added to the v3.19 milestone so that we won't forget about it.

PR #6436, opened 2026-07-29
Deprecate netNamespaceFilePath for Pod networking/restart feature

nixpanic:
The feature is not commonly used. Rook and Ceph-CSI Operator have already
deprecated the feature and removed support for it.

Nodeplugins only have `hostPID: true` set because of this feature. By dropping
the feature, it is possible to reduce the permissions of the Nodeplugin a
little more.

Updates: #6415

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -5,3 +5,15 @@
 ## Features
 
 ## NOTE
+
+## Deprecations
+
+- The `netNamespaceFilePath` configuration option is now deprecated and will be
+  removed in a future release. Users should migrate to using host networking for
+  CSI plugin pods instead. When this feature is detected, a deprecation warning
+  will be logged at the WARNING level.
+   - `hostPID` was changed from `true` to `false` in all static manifests
+   - Static manifest users with `netNamespaceFilePath` must manually set
+     `hostPID: true`
+   - OpenShift users must also keep `allowHostPID: true` in their SCC (if they
+     were using it)
--- a/api/deploy/ocp/scc.yaml
+++ b/api/deploy/ocp/scc.yaml
@@ -14,8 +14,8 @@ priority:
 allowedCapabilities: ["SYS_ADMIN"]
 # Needed as we run liveness container on daemonset pods
 allowHostPorts: true
-# Needed as we are setting this in RBD plugin pod
-allowHostPID: true
+# Set to true when netNamespaceFilePath is configured
+allowHostPID: false
 # Required for encryption
 allowHostIPC: true
 # Set to false as we write to RootFilesystem inside csi containers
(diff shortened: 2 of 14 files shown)


nixpanic:
> /test ci/centos/mini-e2e/k8s-1.37

Running at https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/job/mini-e2e_k8s-1.37/4/display/redirect

nixpanic:
> /test ci/centos/mini-e2e/k8s-1.37

Lots of tests have passed, except:
```
nfs Test NFS CSI [It] create a PVC-PVC clone and bind it to an app
```

[logs](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.37/detail/mini-e2e_k8s-1.37/5/pipeline)

api/deploy/ocp/scc.yaml
```diff
@@ -14,8 +14,8 @@ priority:
 allowedCapabilities: ["SYS_ADMIN"]
 # Needed as we run liveness container on daemonset pods
 allowHostPorts: true
-# Needed as we are setting this in RBD plugin pod
-allowHostPID: true
+# Set to true when netNamespaceFilePath is configured
+allowHostPID: false
```
black-dragon74, line 18: I think for deprecation we should keep this to true. Setting it false here wouldn't allow anyone to enable HostPID manually. WDYT?

black-dragon74:
> Lots of tests have passed, except:

I don't think it is related to this change. Maybe we should run it again?

api/deploy/ocp/scc.yaml
nixpanic: I don't know if anyone uses the SCC... Disabling it now would get the attention from users that require `hostPID: true`, and they get informed about the deprecation.

black-dragon74: Yep, fair.

nixpanic:
> I don't think it is related to this change. Maybe we should run it again?

CI was very busy yesterday, so I didn't rerun the job yet. The failure is indeed unrelated.

PendingReleaseNotes.md
```diff
@@ -5,3 +5,10 @@
 ## Features
 
 ## NOTE
+
+## Deprecations
+
+- The `netNamespaceFilePath` configuration option is now deprecated and will be
+  removed in a future release. Users should migrate to using host networking for
```
black-dragon74, line 12: Static manifests already have host networking, maybe expand the release notes to include:

- `hostPID` was changed from `true` to `false` in all static manifests
- Static manifest users with `netNamespaceFilePath` must manually set `hostPID: true`
- OpenShift users must also keep `allowHostPID: true` in their SCC (if they were using it)

nixpanic: Thanks for the idea, added it.

black-dragon74 (approved):

api/deploy/ocp/scc.yaml
Madhu-1, line 18: @nixpanic did we verify this change in OCP? if i remember correctly this was present even before the netNamespace feature

PR #6442, opened 2026-07-31
rbd: skip DiffIterate for block volume stats without object-map

iPraveenParihar:

```
DiffIterate with WholeObject relies on the object-map feature for fast used-bytes calculation.
Images created before object-map was part of the default feature set (e.g. older PVC clones)
cause DiffIterate to fall back to reading actual data blocks, making NodeGetVolumeStats extremely
slow.

Check for the object-map feature before calling DiffIterate and fall back to returning only the block
device size when it is absent.
```

**Checklist:**

---

<details>

</details>


iPraveenParihar:
@Rakshith-R PTAL

black-dragon74:
Should rebase on top of: https://github.com/ceph/ceph-csi/pull/6536

iPraveenParihar:
@mergifyio rebase

internal/rbd/rbd_util.go
```diff
@@ -1940,6 +2006,7 @@ func (ri *rbdImage) getParent() (*rbdImage, error) {
 	parentImage.Pool = ri.ParentPool
 	parentImage.RadosNamespace = ri.RadosNamespace
 	parentImage.RbdImageName = ri.ParentName
+	defer parentImage.conn.Destroy()
```
black-dragon74: This will double destroy. `flattenParent` defers destroy as well.

Only destroy here in case of error below:
```go
...
err = parentImage.getImageInfo()
	if err != nil {
        parentImage.conn.Destroy() // here
		return nil, err
	}
...
```

Such leak exists in [mirror.go](https://github.com/iPraveenParihar/ceph-csi/blob/8c3e3fda5d413193e25e173b9270254fc7b81b63/internal/rbd/mirror.go#L65) as well. Add deferred `parent.Destory()` after the nil check there as well.

internal/rbd/nodeserver.go
```diff
@@ -1864,6 +1881,21 @@ func (ns *NodeServer) blockNodeGetVolumeStats(
 	}
 	defer rv.Destroy(ctx)
 
+	// DiffIterate requires object-map, fast-diff, and exclusive-lock on the entire chain
+	requiredFeatures := librbd.FeatureObjectMap | librbd.FeatureFastDiff | librbd.FeatureExclusiveLock
+	hasRequiredFeatures, err := rv.allImageInChainHasFeature(ctx, requiredFeatures)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to check required features on image chain %s: %v",
```
black-dragon74: Why not fallback to `getBlockMetrics` like everywhere else?

internal/rbd/nodeserver.go
```diff
@@ -53,6 +54,13 @@ type NodeServer struct {
 	VolumeLocks   *util.IDLocker
 	healthChecker hc.Manager
 
+	// chainSupportsDiffIterate caches whether a volume's full image chain
+	// supports DiffIterate (object-map, fast-diff and exclusive-lock enabled
+	// on every image). Populated on first blockNodeGetVolumeStats call,
+	// evicted on NodeUnstageVolume. This avoids repeated Ceph round-trips
+	// since kubelet polls NodeGetVolumeStats every ~2 minutes.
```
black-dragon74, line 61: We only act in the negative case i.e when `!supported`. We will do all the Ceph calls for fast-diff supported volumes. We only bail out early in case of unsupported ones. The comments around this could be clarified a bit.

internal/rbd/rbd_util.go
iPraveenParihar: `parent.Destroy()` has a safe check already
```
if ri.conn != nil {
		ri.conn.Destroy()
		ri.conn = nil
	}
```

I'll change it to 
```
defer func(){
		parentImage.conn.Destroy()
		parentImage.conn = nil
	}()
```

internal/rbd/nodeserver.go
iPraveenParihar: We store true if it supports DiffIterate, false otherwise.

black-dragon74: I was talking about the comment:
> This avoids repeated Ceph round-trips

That is not true for fast-diff supported volumes. We just keep storing true over and over again.

internal/rbd/rbd_util.go
black-dragon74: Correction, mirror.go does not need the suggested changes. However, it does destroy [here](https://github.com/iPraveenParihar/ceph-csi/blob/5b2fe4f69c732a0495467042299de61a4fbbf09f/internal/rbd/mirror.go#L65-L69) and so does rbd_util: [here](https://github.com/iPraveenParihar/ceph-csi/blob/5b2fe4f69c732a0495467042299de61a4fbbf09f/internal/rbd/rbd_util.go#L2029-L2032)

We cannot destroy the connection unconditionally and should do only in case of errors. Else, the refcount will be decremented twice (once here and again when the callers call Destroy), leading to an unintentional gc-ing of the connection.

I would still change the block below to:
```go
...
err = parentImage.getImageInfo()
	if err != nil {
        // only destroy ourselves in case of error, let the caller handle it otherwise
        parentImage.conn.Destroy()
		return nil, err
	}
...
```

internal/rbd/nodeserver.go
iPraveenParihar: We store it only once for a volumeId and then use it if found. We don't iterate over it again
```
if cached, ok := ns.chainSupportsDiffIterate.Load(volumeId); ok {
	supported, isBool := cached.(bool)
	if isBool && !supported {
		return getBlockMetrics(ctx, targetPath)
	}
}
```

is there something am not getting it? 😓

iPraveenParihar: Fixed it.

black-dragon74: I don't see any changes related to the cache :(

internal/rbd/nodeserver.go
```diff
@@ -1844,6 +1854,13 @@ func (ns *NodeServer) blockNodeGetVolumeStats(
 		return getBlockMetrics(ctx, targetPath)
 	}
 
+	if cached, ok := ns.chainSupportsDiffIterate.Load(volumeId); ok {
+		supported, isBool := cached.(bool)
+		if isBool && !supported {
+			return getBlockMetrics(ctx, targetPath)
+		}
```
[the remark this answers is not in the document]
black-dragon74: It is the same as above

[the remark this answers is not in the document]
iPraveenParihar: > so a parent lacking these features can disappear after the first stats call.

This is not true for now, ceph-csi can't update features on existing images.

black-dragon74 (dismissed):
LGTM, thanks :)

Rakshith-R (dismissed):

iPraveenParihar:
Needs to rebase after https://github.com/ceph/ceph-csi/pull/6561 is Merged

iPraveenParihar:
@mergifyio rebase

iPraveenParihar:
Manually rebased

Rakshith-R (approved):

black-dragon74 (approved):

PR #6457, opened 2026-08-05
rbd: return InvalidArgument for unrecognized QoS

### LINKED ISSUE #6443: [RBD] Unrecognized VolumeAttributesClass parameter names silently clear all cgroup QoS metadata
## Describe the bug

When a VolumeAttributesClass contains parameter names that no QoS handler recognizes (e.g., `readIOPS` instead of `maxReadIops`), `ControllerModifyVolume` **succeeds without error** and **silently deletes all existing cgroup QoS metadata** from the RBD image. On the next pod mount, `NodePublishVolume` finds no QoS metadata and the pod starts with no `io.max` limits.

The root cause is that `validateQoSParameters()` only validates parameters it recognizes — it returns `nil` for unknown keys. The subsequent `modifyVolumeAttributes()` loop asks each handler if it recognizes the parameters; when none do, the code falls through to a "clear all QoS" path that was designed for the case where a user intentionally removes QoS from a PVC.

Unrecognized parameter names are **indistinguishable from "no QoS parameters"** in the current code, so the driver interprets wrong names as "user wants to remove QoS" and actively deletes all stored metadata.

## Environment details

- **Image/version of Ceph CSI driver:** `quay.io/cephcsi/cephcsi:canary` (includes [PR #6274: Add cgroup v2 QoS support for RBD volumes](https://github.com/ceph/ceph-csi/pull/6274))
- **Helm chart version:** N/A (deployed via ODF 4.23)
- **Kernel version:** 6.12+ (RHCOS 9.8)
- **Mounter:** krbd (cgroup v2 QoS path)
- **Kubernetes cluster version:** 1.35 (VolumeAttributesClass GA since 1.31)
- **Ceph cluster version:** Reef (deployed via ODF 4.23)

## Steps to reproduce

1. Create a VolumeAttributesClass with **correct** parameter names and a PVC + pod using it:
   ```yaml
   apiVersion: storage.k8s.io/v1
   kind: VolumeAttributesClass
   metadata:
     name: correct-qos
   driverName: openshift-storage.rbd.csi.ceph.com
   parameters:
     maxReadIops: "500"
     maxWriteIops: "500"
     maxReadBps: "5242880"
     maxWriteBps: "10485760"
   ```

2. Verify QoS is applied — pod's cgroup shows `io.max`:
   ```
   252:0 rbps=5242880 wbps=10485760 riops=500 wiops=500
   ```

3. Verify RBD image metadata contains QoS keys:
   ```bash
   rbd image-meta list <pool>/<image>
   # Shows:
   # .rbd.csi.ceph.com/cgroup_qos_max_read_iops   500
   # .rbd.csi.ceph.com/cgroup_qos_max_write_iops   500
   # .rbd.csi.ceph.com/cgroup_qos_max_read_bps     5242880
   # .rbd.csi.ceph.com/cgroup_qos_max_write_bps     10485760
   ```

4. Create a VolumeAttributesClass with **wrong** parameter names:
   ```yaml
   apiVersion: storage.k8s.io/v1
   kind: VolumeAttributesClass
   metadata:
     name: wrong-qos
   driverName: openshift-storage.rbd.csi.ceph.com
   parameters:
     readIOPS: "500"       # wrong — should be maxReadIops
     writeIOPS: "500"      # wrong — should be maxWriteIops
     readBPS: "5242880"    # wrong — should be maxReadBps
     writeBPS: "10485760"  # wrong — should be maxWriteBps
   ```

5. Patch the PVC to the wrong-named VAC:
   ```bash
   kubectl patch pvc my-pvc --type=merge \
     -p '{"spec":{"volumeAttributesClassName":"wrong-qos"}}'
   ```

6. Observe: `ControllerModifyVolume` returns **success** — no error, no event, no warning.

7. Check RBD image metadata again:
   ```bash
   rbd image-meta list <pool>/<image>
   # All .rbd.csi.ceph.com/cgroup_qos_* keys are GONE
   ```

8. Delete and recreate the pod.

9. Check `io.max` on the new pod:
   ```
   # EMPTY — no io.max limits applied
   ```

## Actual results

- `ControllerModifyVolume` returns gRPC success (no error)
- No Kubernetes event or warning is emitted
- PVC `.status.currentVolumeAttributesClassName` updates to `wrong-qos`
- All RBD image QoS metadata keys are silently **deleted**
- Running pods are unaffected (their cgroup `io.max` persists in-kernel)
- On next pod mount (recreation, rollout, node drain), `NodePublishVolume` finds no metadata and the pod starts **unthrottled**

The delayed-effect nature makes this especially dangerous — everything appears correct until the next pod lifecycle event.

## Expected behavior

`ControllerModifyVolume` should return a gRPC `INVALID_ARGUMENT` error when the request contains mutable parameters that no QoS handler recognizes. The [CSI spec defines this error code](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllermodifyvolume-errors) for exactly this situation:

> **INVALID_ARGUMENT:** ... Indicates that a required parameter or value is missing or invalid.

Existing QoS metadata on the RBD image should **not** be modified when the request parameters are unrecognized.

## Code path analysis

The call chain that leads to silent clearing:

### 1. `validateQoSParameters()` does not reject unknown params

**File:** `internal/rbd/controllerserver.go`, function `validateQoSParameters()`

For the krbd mounter, the function checks:
- `HasQoSParams(mutableParams)` — looks for NBD params (`baseIops`, `baseReadIops`, etc.). Wrong names like `readIOPS` are not in this set → returns `false`.
- `hasCgroupQoSParams(mutableParams)` — looks for exactly `maxReadIops`, `maxWriteIops`, `maxReadBps`, `maxWriteBps`. Wrong names not in this set → returns `false`.
- **Returns `nil`** — no error. Unknown params pass through silently.

### 2. `modifyVolumeAttributes()` fallthrough to clearing

**File:** `internal/rbd/rbd_util.go`, function `modifyVolumeAttributes()`

Two handlers are created: `cgroupQoSHandler` and `nbdQoSHandler`. The code iterates:
```go
for _, handler := range handlers {
    if handler.HasParams(newMutableParameters) {
        // validate + apply + return
    }
}
```

Neither handler recognizes the wrong parameter names → the loop completes without returning.

**The critical fallthrough:**
```go
// No QoS parameters provided - clear any existing QoS of all types.
// This happens when VolumeAttributesClass is removed from the PVC.
log.DebugLog(ctx, "no QoS parameters in request, clearing existing QoS for volume %s", rv.VolID)

for _, handler := range handlers {
    if err := handler.Clear(ctx); err != nil { ...
```

### 3. `hasCgroupQoSParams()` — the four recognized names

**File:** `internal/rbd/cgroup_qos.go`, function `hasCgroupQoSParams()`

```go
func hasCgroupQoSParams(params map[string]string) bool {
    cgroupParams := []string{maxReadIops, maxWriteIops, maxReadBps, maxWriteBps}
    for _, param := range cgroupParams {
        if val, ok := params[param]; ok && val != "" {
            return true
        }
    }
    return false
}
```

The constants are defined in `internal/rbd/qos.go`:
```go
maxReadIops  = "maxReadIops"
maxWriteIops = "maxWriteIops"
maxReadBps   = "maxReadBps"
maxWriteBps  = "maxWriteBps"
```

### 4. `Clear()` deletes all metadata

**File:** `internal/rbd/cgroup_qos.go`, function `Clear()` → `saveCgroupQoS()`

`Clear()` calls `saveCgroupQoS(ctx, map[string]string{})` with an empty map. `saveCgroupQoS()` detects no cgroup params and iterates over `qosParamToMetadataKey`, calling `rv.RemoveMetadata(metadataKey)` for each of the four RBD image metadata keys.

### 5. Next mount finds nothing

**File:** `internal/rbd/cgroup_qos.go`, function `applyCgroupQoSForVolume()`

Called during `NodePublishVolume`. Reads metadata via `getCgroupQoS()` — now empty. Logs `"no cgroup QoS configured for volume"` at DEBUG level and returns without writing `io.max`.

### Summary diagram

```
ControllerModifyVolume(params: {readIOPS: "500", writeIOPS: "500", ...})
  │
  ├─ validateQoSParameters() → nil (unknown params not rejected)
  │
  └─ modifyVolumeAttributes()
       ├─ cgroupQoSHandler.HasParams() → false (no maxReadIops etc.)
       ├─ nbdQoSHandler.HasParams()    → false (no baseIops etc.)
       │
       └─ FALLTHROUGH: "no QoS parameters in request"
            ├─ cgroupQoSHandler.Clear() → RemoveMetadata(all 4 keys)
            └─ nbdQoSHandler.Clear()    → RemoveMetadata(all NBD keys)

Next NodePublishVolume:
  applyCgroupQoSForVolume()
    └─ getCgroupQoS() → empty map → "no cgroup QoS configured" → no io.max
```

## Proposed fix

Add a check in `modifyVolumeAttributes()` (or `validateQoSParameters()`) that distinguishes between:
1. **Empty mutable parameters** = user removed VAC from PVC → clear QoS (current behavior, correct)
2. **Non-empty mutable parameters but no handler recognizes them** = wrong parameter names → return `codes.InvalidArgument`

For example, in `modifyVolumeAttributes()` after the handler loop:

```go
// Current: falls through to Clear()
// Proposed:
if len(newMutableParameters) > 0 {
    return status.Errorf(codes.InvalidArgument,
        "mutable parameters contain unrecognized keys: %v — "+
            "valid cgroup QoS parameters are: maxReadIops, maxWriteIops, maxReadBps, maxWriteBps",
        maps.Keys(newMutableParameters))
}
// Only clear when parameters are genuinely empty
```

This preserves the intentional "clear on VAC removal" behavior while preventing wrong parameter names from silently destroying QoS metadata.

## Impact

| Affected scenario | Detail |
|--------------------|--------|
| **Initial setup** | Users following early documentation or prototypes that used different parameter naming conventions (e.g., `readIOPS`, `writeBPS`, `maxReadIOPS`) will get no QoS with no error. The HackMD design document used capitalized `maxReadIOPS` — the PR implemented lowercase `maxReadIops`. |
| **VAC migration** | If a ceph-csi upgrade changes expected parameter names, patching PVCs to a VAC with old-style names silently clears QoS on the next `ControllerModifyVolume`. |
| **Delayed-effect danger** | Running pods are unaffected (cgroup `io.max` persists in-kernel), so the problem is invisible until the next pod recreation — which could be days or weeks later during a routine rollout or node drain. |
| **Multi-tenant risk** | When QoS is silently cleared and a pod is recreated without limits, it competes for I/O bandwidth unthrottled while other tenants' pods are still throttled — a noisy-neighbor scenario. |
| **Diagnosability** | Everything looks correct from the Kubernetes API: PVC shows the VAC name, `ControllerModifyVolume` returned success, no events or warnings. Only inspecting RBD image metadata (`rbd image-meta list`) or the pod's `io.max` reveals the failure. |

## Logs

The clearing happens at DEBUG log level:

```
"no QoS parameters in request, clearing existing QoS for volume <volID>"
```

And on next mount:

```
"no cgroup QoS configured for volume <volID>"
```

Both are DEBUG, not WARN or ERROR — no indication that something unexpected happened.

Madhu-1:
Unrecognized or invalid QoS parameters in mutable volume attributes are a client input error, not an internal
server error. Wrap these errors with ErrInvalidArgument so callers return codes.InvalidArgument instead of codes.Internal.

Allow unsetting the Qos by setting the values as max

### DIFF
--- a/internal/rbd/cgroup_qos.go
+++ b/internal/rbd/cgroup_qos.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -99,6 +101,11 @@ type cgroupQoSHandler struct {
 // newCgroupQoSHandler creates a new cgroup QoS handler.
 func newCgroupQoSHandler(volume *rbdVolume) QoSHandler {
 	return &cgroupQoSHandler{volume: volume}
+}
+
+// cgroupQoSKnownKeys returns the cgroup v2 QoS parameter keys.
+func cgroupQoSKnownKeys() []string {
+	return slices.Collect(maps.Keys(qosParamToMetadataKey))
 }
 
 // HasParams checks if cgroup v2 QoS parameters are present in the request.
--- a/internal/rbd/cgroup_qos.go
+++ b/internal/rbd/cgroup_qos.go
@@ -66,6 +66,9 @@ const (
 	qosMetadataMaxReadBps   = qosMetadataKeyPrefix + "cgroup_qos_max_read_bps"
 	qosMetadataMaxWriteBps  = qosMetadataKeyPrefix + "cgroup_qos_max_write_bps"
 
+	// cgroupQoSMaxLimit is the cgroup v2 value for unlimited I/O.
+	cgroupQoSMaxLimit = "max"
+
 	// io.max file for cgroup v2.
 	ioMaxFile = "io.max"
 
@@ -153,10 +156,10 @@ type cgroupQoS struct {
 // parseCgroupQoSParams extracts cgroup v2 QoS parameters from VolumeAttributesClass.
 func parseCgroupQoSParams(params map[string]string) *cgroupQoS {
 	qos := &cgroupQoS{
-		maxReadIops:  "max",
-		maxWriteIops: "max",
-		maxReadBps:   "max",
-		maxWriteBps:  "max",
+		maxReadIops:  cgroupQoSMaxLimit,
+		maxWriteIops: cgroupQoSMaxLimit,
+		maxReadBps:   cgroupQoSMaxLimit,
+		maxWriteBps:  cgroupQoSMaxLimit,
 	}
 
 	if val, ok := params[maxReadIops]; ok && val != "" {
@@ -295,12 +298,16 @@ func (qos *cgroupQoS) applyCgroupQoS(ctx context.Context, podUID string) error {
 func validateCgroupQoSParams(params map[string]string) error {
 	for key := range qosParamToMetadataKey {
 		if val, ok := params[key]; ok && val != "" {
+			// "max" means unlimited (removes the limit).
+			if val == cgroupQoSMaxLimit {
+				continue
+			}
 			parsed, err := strconv.ParseInt(val, 10, 64)
 			if err != nil {
-				return fmt.Errorf("invalid value for %s: %s, must be a positive integer", key, val)
+				return fmt.Errorf("invalid value for %s: %s, must be a non-negative integer or %q", key, val, cgroupQoSMaxLimit)
 			}
-			if parsed <= 0 {
-				return fmt.Errorf("invalid value for %s: %s, must be greater than 0", key, val)
+			if parsed < 0 {
+				return fmt.Errorf("invalid value for %s: %s, must be a non-negative integer or %q", key, val, cgroupQoSMaxLimit)
 			}
 		}
 	}
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"slices"
 	"strconv"
 
 	osdAdmin "github.com/ceph/go-ceph/common/admin/osd"
@@ -159,27 +160,65 @@ func validateQoSParameters(mutableParams map[string]string, mounter string) erro
 		// NBD mounter: only NBD QoS params are valid.
 		// The shared keys (maxReadIops, etc.) serve as NBD max limits here.
 		if HasQoSParams(mutableParams) {
+			if unknown := unrecognizedKeys(mutableParams, nbdQoSKnownKeys); len(unknown) > 0 {
+				return fmt.Errorf("mutable parameters %v are not recognized QoS parameters for %q mounter",
+					unknown, mounter)
+			}
+
 			return validateNBDQoSParams(mutableParams)
 		}
 
-		return nil
+		keys := slices.Sorted(maps.Keys(mutableParams))
+
+		return fmt.Errorf("mutable parameters %v for %q mounter are missing required base QoS parameters",
+			keys, mounter)
 	}
 
 	// krbd mounter: only cgroup v2 QoS params are supported.
 	// Reject any NBD-specific params that won't be applied via cgroup.
 	if HasQoSParams(mutableParams) {
-		return fmt.Errorf(
-			"NBD QoS parameters (baseIops, baseReadIops, etc.) are not supported with %q mounter, "+
-				"use cgroup QoS parameters (maxReadIops, maxWriteIops, maxReadBps, maxWriteBps)", mounter)
+		keys := slices.Sorted(maps.Keys(mutableParams))
+
+		return fmt.Errorf("mutable parameters %v contain NBD-specific QoS parameters not supported with %q mounter",
+			keys, mounter)
 	}
 
 	if hasCgroupQoSParams(mutableParams) {
+		if unknown := unrecognizedKeys(mutableParams, cgroupQoSKnownKeys()); len(unknown) > 0 {
+			return fmt.Errorf("mutable parameters %v are not recognized QoS parameters for %q mounter",
+				unknown, mounter)
+		}
+
 		if err := validateCgroupQoSParams(mutableParams); err != nil {
 			return fmt.Errorf("invalid cgroup QoS parameters: %w", err)
 		}
+
+		return nil
 	}
 
-	return nil
+	keys := slices.Sorted(maps.Keys(mutableParams))
+
+	return fmt.Errorf("mutable parameters %v for %q mounter are missing required cgroup QoS parameters",
+		keys, mounter)
+}
+
+// unrecognizedKeys returns parameter keys not present in the known set.
+func unrecognizedKeys(params map[string]string, known []string) []string {
+	knownSet := make(map[string]struct{}, len(known))
+	for _, k := range known {
+		knownSet[k] = struct{}{}
+	}
+
+	var unknown []string
+	for k := range params {
+		if _, ok := knownSet[k]; !ok {
+			unknown = append(unknown, k)
+		}
+	}
+
+	slices.Sort(unknown)
+
+	return unknown
 }
 
 // parseVolCreateRequest take create volume `request` argument and make use of the
@@ -2075,8 +2114,10 @@ func (cs *ControllerServer) ControllerModifyVolume(
 		return nil, status.Errorf(codes.Internal, "failed to determine volume mounter type: %v", err)
 	}
 
-	if err = validateQoSParameters(mutableParameters, rbdVol.Mounter); err != nil {
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+	if len(mutableParameters) > 0 {
+		if err = validateQoSParameters(mutableParameters, rbdVol.Mounter); err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
 	}
 
 	// set RequestedVolSize, because calcQosBasedOnCapacity use it.
@@ -2085,6 +2126,10 @@ func (cs *ControllerServer) ControllerModifyVolume(
 	err = rbdVol.modifyVolumeAttributes(ctx, mutableParameters)
 	if err != nil {
 		log.ErrorLog(ctx, "failed to modify volume: %s with error: %v", rbdVol, err)
+
+		if errors.Is(err, rbderrors.ErrInvalidArgument) {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
 
 		return nil, status.Error(codes.Internal, err.Error())
 	}
--- a/docs/volumeattributesclass.md
+++ b/docs/volumeattributesclass.md
@@ -211,6 +211,20 @@ on systems with cgroup v2 enabled.
   (`csi.storage.k8s.io/node-publish-secret-name`) or CSI ConfigMap
   (`rbd.nodePublishSecretRef`)
 
+#### Cgroup v2 QoS Parameter Values
+
+Each parameter accepts a non-negative integer or the special value `"max"`:
+
+| Value            | Meaning                          |
+| ---------------- | -------------------------------- |
+| `"max"`          | Remove the limit (unlimited I/O) |
+| `"0"`            | Fully throttle I/O               |
+| positive integer | Set the limit to that value      |
+
+To revert all QoS limits applied by a VolumeAttributesClass, create a new
+VolumeAttributesClass with all values set to `"max"` and modify the PVC to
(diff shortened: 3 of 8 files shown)


internal/rbd/controllerserver.go
```diff
@@ -2086,6 +2090,10 @@ func (cs *ControllerServer) ControllerModifyVolume(
 	if err != nil {
 		log.ErrorLog(ctx, "failed to modify volume: %s with error: %v", rbdVol, err)
 
+		if errors.Is(err, rbderrors.ErrInvalidArgument) {
```
iPraveenParihar, line 2130: Needs to be done at other places where `modifyVolumeAttributes` is called.

internal/rbd/controllerserver.go
```diff
@@ -162,24 +162,28 @@ func validateQoSParameters(mutableParams map[string]string, mounter string) erro
 			return validateNBDQoSParams(mutableParams)
 		}
 
-		return nil
+		return fmt.Errorf("mutable parameters %v contain no recognized QoS parameters for %q mounter",
```
iPraveenParihar:
```suggestion
		return fmt.Errorf("mutable parameters %v contain no recognized QoS keys for %q mounter",
```

Keep it consistent everywhere?

iPraveenParihar (changes requested):
@Madhu-1, Have you tested if when the VAC is deleted or removed from a PVC, the QoS settings are cleared?

internal/rbd/controllerserver.go
Madhu-1: it called in two places create volume and modify volume, in create volume we are already validating it in parse request even  before modify volume and we will never hit this case that why i didnt add the check for it.

internal/rbd/controllerserver.go
Madhu-1: change it to parameters in other place as its more opt

Madhu-1:
> @Madhu-1, Have you tested if when the VAC is deleted or removed from a PVC, the QoS settings are cleared?

This can never happened, only VAC can be changed from the PVC and VAC cannot be deleted if its attached to a PVC, what we have is safe check where it never happens in kubernetes. kept it to be on safer side for other CO

Rakshith-R (dismissed):

iPraveenParihar (dismissed):

iPraveenParihar:
```
nvmeof Test NVMe CSI [It] create a PVC with VAC containing allowHostNQNs, apply new VAC then delete the PVC

/go/src/github.com/ceph/ceph-csi/e2e/nvmeof.go:316

  [FAILED] Unexpected error:
      <*fmt.wrapError | 0x550ffd42da0>: 
      failed to get pvc: client rate limiter Wait returned an error: context deadline exceeded
      {
          msg: "failed to get pvc: client rate limiter Wait returned an error: context deadline exceeded",
          err: <*fmt.wrapError | 0x550ffd42d80>{
              msg: "client rate limiter Wait returned an error: context deadline exceeded",
              err: <context.deadlineExceededError>{},
          },
      }
  occurred
  In [It] at: /go/src/github.com/ceph/ceph-csi/e2e/nvmeof.go:395 @ 08/05/26 11:21:49.653

------------------------------

SS

```

Madhu-1:
@nixpanic reusing the RBD functions like CreateVolume can cause problem for nvme, we need to separate it out as checks are unique for RBD and nvme will never pass through or we need to unset parameters specific to nvme before calling RBD functions

nixpanic:
I guess there is something in the NVMe-oF Controller that calls ControllerModifyVolume on the RBD Controller :thinking:

All failed CI jobs (3 currently, all other e2e still running), failed with this:

```
nvmeof Test NVMe CSI [It] create a PVC with VAC containing allowHostNQNs, apply new VAC then delete the PVC
```

/cc @gadididi

Madhu-1:
> I guess there is something in the NVMe-oF Controller that calls ControllerModifyVolume on the RBD Controller 🤔
> 
> All failed CI jobs (3 currently, all other e2e still running), failed with this:
> 
> ```
> nvmeof Test NVMe CSI [It] create a PVC with VAC containing allowHostNQNs, apply new VAC then delete the PVC
> ```
> 
> /cc @gadididi

it because we are calling rbd createVolume from vnme CreateVolume. added a 2nd commit to strip nvme specific mutable parameters

internal/nvmeof/controller/controllerserver_test.go
```diff
@@ -118,6 +118,71 @@ func TestPolulateVolumeContext(t *testing.T) {
...
+				nvmeof.RwMbytesPerSecond: "100",
+				nvmeof.RMbytesPerSecond:  "50",
+				nvmeof.WMbytesPerSecond:  "50",
+			},
+			expected: map[string]string{},
+		},
+		{
+			name: "only NVMe-oF host key",
+			input: map[string]string{
+				nvmeof.AllowHostNQNs: "nqn.2014-08.org.nvmexpress:host1",
```
gadididi: nit- technically the test works with any value - the function doesn't validate it (just the key)

nvmeof.AllowHostNQNs value is yaml list 
so, I think it should be either

```Go
nvmeof.AllowHostNQNs: `["nqn.2014-08.org.nvmexpress:host1"]`,
```
or

```Go
nvmeof.AllowHostNQNs: `- nqn.2014-08.org.nvmexpress:host1`,
```
here and in line 166

Madhu-1:
test running here https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org/blue/organizations/jenkins/mini-e2e_k8s-1.35/detail/mini-e2e_k8s-1.35/337/pipeline

internal/nvmeof/controller/controllerserver.go
```diff
@@ -147,6 +147,10 @@ func (cs *Server) CreateVolume(
 		defer cs.volumeLocks.Release(sourceVolumeID)
 	}
 
+	// Replace mutable parameters with only the non-NVMe-oF keys so
+	// the RBD backend does not reject them as unrecognized QoS keys.
+	req.MutableParameters = filterOutNVMeoFMutableParameters(req.GetMutableParameters())
```
nixpanic: I do not think there is a need to filter them. There are currently no mutable parameters that are valid for both nvmeof and rbd.

The parameters should be restored later on, otherwise nvmeof QoS won't work anymore.

I'll add that to #6458. If this PR is rebased on top of that, you do not need to worry about the nvmeof changes.

Madhu-1: I dropped the second commit for nvme, This PR is how it was before, can be rebased on 6458 is merged

Madhu-1:
Added one more commit to allow unsetting the Qos

internal/rbd/cgroup_qos.go
```diff
@@ -288,12 +288,16 @@ func (qos *cgroupQoS) applyCgroupQoS(ctx context.Context, podUID string) error {
...
 	for key := range qosParamToMetadataKey {
 		if val, ok := params[key]; ok && val != "" {
+			// "max" means unlimited (removes the limit).
+			if val == "max" {
+				continue
+			}
 			parsed, err := strconv.ParseInt(val, 10, 64)
 			if err != nil {
-				return fmt.Errorf("invalid value for %s: %s, must be a positive integer", key, val)
+				return fmt.Errorf("invalid value for %s: %s, must be a non-negative integer or \"max\"", key, val)
```
black-dragon74: Maybe store `"max"` into a const.

```suggestion
				return fmt.Errorf("invalid value for %s: %q, expected a non-negative integer or %q", key, val, "max")
```

internal/rbd/cgroup_qos.go
```diff
@@ -288,12 +288,16 @@ func (qos *cgroupQoS) applyCgroupQoS(ctx context.Context, podUID string) error {
...
 			}
-			if parsed <= 0 {
-				return fmt.Errorf("invalid value for %s: %s, must be greater than 0", key, val)
+			if parsed < 0 {
+				return fmt.Errorf("invalid value for %s: %s, must be a non-negative integer or \"max\"", key, val)
```
black-dragon74: Same as above

Madhu-1:
@mergifyio rebase

iPraveenParihar (approved):

Rakshith-R (approved):

PR #6458, opened 2026-08-05
cleanup: validate unknown parameters in ControllerModifyVolume

nixpanic:
Add validation to reject unknown mutable parameters in ControllerModifyVolume.
Previously, unknown parameters were silently ignored, which could lead to
confusion when users specify incorrect parameter names.

Related: #6443

### DIFF
--- a/internal/nfs/controller/controllerserver.go
+++ b/internal/nfs/controller/controllerserver.go
@@ -19,6 +19,8 @@ package controller
 import (
 	"context"
 	"errors"
+	"maps"
+	"slices"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
@@ -45,6 +47,21 @@ type nfsControllerServer struct {
 
 	// backendGroupServer handles the CephFS group controller requests
 	backendGroupServer csi.GroupControllerServer
+}
+
+// nfsParameters contains the list of parameters for NFS volumes. These
+// parameters should not be passed to the CephFS backend.
+var nfsParameters = []string{
+	nfs.ParameterCluster,
+	nfs.ParameterSecTypes,
+}
+
+// nfsMutableParameters contains the list of mutable parameters for NFS
+// volumes. These parameters are handled separately through
+// ControllerModifyVolume and should not be passed to the CephFS backend.
+var nfsMutableParameters = []string{
+	nfs.ParameterServer,
+	nfs.ParameterClients,
 }
 
 // Assert required implementation of CSI interfaces.
@@ -93,9 +110,47 @@ func (cs *nfsControllerServer) CreateVolume(
 	ctx context.Context,
 	req *csi.CreateVolumeRequest,
 ) (*csi.CreateVolumeResponse, error) {
+	// split NFS and CephFS parameters, the CephFS controller does not
+	// accept the NFS parameters and would report an error. The nfsParams
+	// need to be added to the VolumeContext for further consumption.
+	nfsParams := make(map[string]string)
+	cephfsParams := make(map[string]string)
+	for key, value := range req.GetParameters() {
+		if slices.Contains(nfsParameters, key) {
+			nfsParams[key] = value
+		} else {
+			cephfsParams[key] = value
+		}
+	}
+
 	// nfs does not supports shallow snapshots
-	req.Parameters["backingSnapshot"] = "false"
-	res, err := cs.backendServer.CreateVolume(ctx, req)
+	cephfsParams["backingSnapshot"] = "false"
+
+	// Filter out NFS-specific mutable parameters before passing to CephFS
+	// NFS handles mutable parameters separately through ControllerModifyVolume
+	nfsMutableParams := make(map[string]string)
+	cephfsMutableParams := make(map[string]string)
+	for key, value := range req.GetMutableParameters() {
+		if slices.Contains(nfsMutableParameters, key) {
+			nfsMutableParams[key] = value
+		} else {
+			cephfsMutableParams[key] = value
+		}
+	}
+
+	// Create a modified request without mutable parameters for CephFS
+	cephfsReq := &csi.CreateVolumeRequest{
+		Name:                      req.GetName(),
+		CapacityRange:             req.GetCapacityRange(),
+		VolumeCapabilities:        req.GetVolumeCapabilities(),
+		Parameters:                cephfsParams,
+		Secrets:                   req.GetSecrets(),
+		VolumeContentSource:       req.GetVolumeContentSource(),
+		AccessibilityRequirements: req.GetAccessibilityRequirements(),
+		MutableParameters:         cephfsMutableParams,
+	}
+
+	res, err := cs.backendServer.CreateVolume(ctx, cephfsReq)
 	if err != nil {
 		return nil, err
 	}
@@ -124,6 +179,11 @@ func (cs *nfsControllerServer) CreateVolume(
 	}
 	defer nfsVolume.Destroy()
 
+	// re-add the NFS parameters to the volume
+	vc := backend.GetVolumeContext()
+	maps.Copy(vc, nfsParams)
+	backend.VolumeContext = vc
+
 	err = nfsVolume.CreateExport(backend)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "failed to create export: %v", err)
@@ -143,7 +203,7 @@ func (cs *nfsControllerServer) CreateVolume(
 			&csi.ControllerModifyVolumeRequest{
 				VolumeId:          backend.GetVolumeId(),
 				Secrets:           req.GetSecrets(),
-				MutableParameters: req.GetMutableParameters(),
+				MutableParameters: nfsMutableParams,
 			})
 		if err != nil {
 			return nil, err
@@ -276,6 +336,14 @@ func (cs *nfsControllerServer) ControllerModifyVolume(
 	ctx context.Context,
 	req *csi.ControllerModifyVolumeRequest,
 ) (*csi.ControllerModifyVolumeResponse, error) {
+	// Validate that only known parameters are provided (before any I/O)
+	for param := range req.GetMutableParameters() {
+		if !slices.Contains(nfsMutableParameters, param) {
+			return nil, status.Errorf(codes.InvalidArgument,
+				"unknown mutable parameter: %s", param)
+		}
+	}
+
 	secret := req.GetSecrets()
 	cr, err := util.NewAdminCredentials(secret)
 	if err != nil {
--- a/internal/nvmeof/controller/controllerserver.go
+++ b/internal/nvmeof/controller/controllerserver.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"maps"
 	"net"
+	"slices"
 	"strconv"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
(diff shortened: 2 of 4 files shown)


Rakshith-R (dismissed):

black-dragon74 (dismissed):

internal/nvmeof/controller/controllerserver.go
gadididi: you need also here (at create PVC time, not only on modifyVolume) validating if unknown values are provided

internal/nvmeof/controller/controllerserver.go
```diff
@@ -370,7 +371,23 @@ func (cs *Server) ControllerModifyVolume(
 	}
 	defer cs.volumeLocks.Release(volumeID)
 
-	// Step 2: Parse QoS parameters from mutable_parameters
+	// Step 2: Validate that only known parameters are provided
+	knownParams := []string{
```
iPraveenParihar: @nixpanic, we might need to add RBD QoS known params too?

iPraveenParihar: or move the HasQoSParams() before this?

internal/nfs/controller/controllerserver.go
```diff
@@ -296,6 +297,19 @@ func (cs *nfsControllerServer) ControllerModifyVolume(
 	}
 	defer nfsVolume.Destroy()
 
+	// Validate that only known parameters are provided
+	knownParams := []string{
+		nfs.ParameterServer,
+		nfs.ParameterClients,
+	}
+
+	for param := range req.GetMutableParameters() {
```
iPraveenParihar: Can be moved above before NewAdminCredentials, NewNFSVolume, and nfsVolume.Connect() all of which involve real network I/O.

iPraveenParihar (changes requested):
few questions

internal/nvmeof/controller/controllerserver.go
nixpanic: Good point!

internal/nfs/controller/controllerserver.go
nixpanic: sure!

internal/nvmeof/controller/controllerserver.go
nixpanic: Maybe... at least some RBD parameters might be allowed.

Madhu-1: can we add it as var, so that i can reuse it https://github.com/ceph/ceph-csi/pull/6457/changes#diff-6e8af9ff01886969b349830d8cfd38bf7e4a41f833811fa118bf805e2e368d2eR469  and same check need to be added to CreateVolume as well

nixpanic: Well, maybe the mutable parameters should be filtered from the CreateVolume request that is passed on to the RBD Controller?

@gadididi, what is your thought on this? It would be cleaner to not pass non-RBD parameters on to the RBD Controller, I think.

Madhu-1: that change is added in #6457 PR as the CI is failing.

nixpanic: RBD QoS is on the local node, NVMe-oF QoS is on the nvme-gateway. I don't think RBD QoS parameters should apply for NVMe-oF volumes.

Agreed, @gadididi ?

gadididi: @nixpanic,  right! 
NVMe-oF QoS is limitation the traffic (IO) for specific namespace. 

nvmeof namespace can get nvmeof QoS **XOR** rbd QoS . 

**but best practice is to get NVMe-oF QoS** when we deal with nvmeof namespace 
So, bottom line- we did not allow setting RBD QoS on nvmeof namespace.

gadididi: @nixpanic I think the only point we all to the backend rbd-csi is in `CreateVolume()` (beside irrelevent VAC places like snapshot..) , in nvmeof `CreateVolume()` we would like to create nvmeof namespace.
Aviv said nvmeof namespace must be with only nvmeof QoS, not RBD QoS. 

So, There is already place in the code we check if RBD QoS params were provided, if so, raise an error.
```Go
	hasRBDQoS := rbd.HasQoSParams(params)
	if hasRBDQoS {
		log.ErrorLog(ctx, "Cannot set RBD QoS parameters on NVMe-oF volumes")

		return nil, status.Error(codes.InvalidArgument, "cannot set RBD QoS parameters on NVMe-oF volumes")
	}
```

Then, we just need to filter the nvmeof QoS params before we call to the rbd-csi backend. I think is clean way. 
the rbd-csi should not know nvmeof QoS params..

internal/nvmeof/controller/controllerserver.go
```diff
@@ -98,6 +108,8 @@ func (cs *Server) ValidateVolumeCapabilities(
 }
 
 // CreateVolume creates a new RBD volume and exposes it through NVMe-oF Gateway.
+//
+//nolint:gocyclo,cyclop // TODO: reduce complexity
```
gadididi: I will open a task\issue and will address it after this PR will be merged

internal/nvmeof/controller/controllerserver.go
```diff
@@ -147,9 +159,33 @@ func (cs *Server) CreateVolume(
 		defer cs.volumeLocks.Release(sourceVolumeID)
 	}
 
-	// Step 1: Create RBD volume through backend. if exists, it is ok.
+	// Step 1: Extract NVMe-oF mutable parameters before passing to RBD
+	extractedMutableParams := make(map[string]string)
```
gadididi: maybe it is leftover , but I dont see you use `extractedMutableParams` var after this loop 🤔

internal/nvmeof/controller/controllerserver.go
```diff
@@ -491,11 +529,15 @@ func validateCreateVolumeRequest(req *csi.CreateVolumeRequest) error {
...
-		return errors.New("setting RBD QoS parameters on NVMe-oF volumes is not supported")
+	// Validate that only known mutable parameters are provided
+	for param := range mutableParams {
+		if !slices.Contains(nvmeofMutableParams, param) {
+			return fmt.Errorf("unknown mutable parameter: %s", param)
+		}
 	}
-	if hasRBDQoS := rbd.HasQoSParams(mutableParams); hasRBDQoS {
+
+	// Check for RBD QoS parameters in regular params (not mutableParams - already validated above)
```
gadididi: I think the comment should be:
```suggestion
	// Check for RBD QoS parameters in regular params (mutableParams already validated above)
```

internal/nvmeof/controller/controllerserver.go
gadididi: opened here: https://github.com/ceph/ceph-csi/issues/6461

internal/nfs/controller/controllerserver.go
```diff
@@ -276,6 +292,19 @@ func (cs *nfsControllerServer) ControllerModifyVolume(
...
+		if !slices.Contains(knownParams, param) {
+			return nil, status.Errorf(codes.InvalidArgument,
+				"unknown mutable parameter: %s", param)
+		}
+	}
```
Madhu-1, line 345: we need to have similar check before creating the subvolume or else we need a defer to cleanup the subvolume if anything fails after creation,  i prefer to have input validation before loop of create/delete

internal/nvmeof/controller/controllerserver.go
nixpanic: indeed, I wanted to add them later to the returned Volume, but that doesn't have the mutable parameters.

internal/nvmeof/controller/controllerserver.go
nixpanic: right, yes

gadididi (dismissed):
looks good thanks!

internal/nfs/controller/controllerserver.go
nixpanic: Currently CephFS does not support mutable parameters. But, well, #6390 is almost there now too. I'll add the extra checks/filtering in CreateVolume too, similar as what is done for nvmeof.

internal/nfs/controller/controllerserver.go
```diff
@@ -94,8 +95,23 @@ func (cs *nfsControllerServer) CreateVolume(
 	req *csi.CreateVolumeRequest,
 ) (*csi.CreateVolumeResponse, error) {
 	// nfs does not supports shallow snapshots
-	req.Parameters["backingSnapshot"] = "false"
-	res, err := cs.backendServer.CreateVolume(ctx, req)
+	params := req.GetParameters()
+	params["backingSnapshot"] = "false"
+
```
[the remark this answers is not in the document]
nixpanic: This is an unrelated issue, it existed before already. There are many `req.GetParameters()` calls in Ceph-CSI that do not have the `nil` check. This can be addressed in a different PR.

internal/nvmeof/controller/controllerserver.go
```diff
@@ -370,13 +393,15 @@ func (cs *Server) ControllerModifyVolume(
...
-		log.ErrorLog(ctx, "Cannot set RBD QoS parameters on NVMe-oF volumes")
-
-		return nil, status.Error(codes.InvalidArgument, "cannot set RBD QoS parameters on NVMe-oF volumes")
+	// Step 2: Validate that only known parameters are provided
+	for param := range params {
+		if !slices.Contains(nvmeofMutableParams, param) {
+			return nil, status.Errorf(codes.InvalidArgument,
+				"unknown mutable parameter: %s", param)
+		}
 	}
```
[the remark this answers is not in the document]
nixpanic: That is fine, we're not changing RBD behavior in this PR.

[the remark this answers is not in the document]
nixpanic: good point, will add a test case for this too

internal/nfs/controller/controllerserver.go
nixpanic: This is done in the last update of this PR.

Madhu-1 (dismissed):

internal/nfs/controller/controllerserver.go
```diff
@@ -94,8 +103,31 @@ func (cs *nfsControllerServer) CreateVolume(
...
+			filteredParams[key] = value
+		}
+	}
+
+	// Create a modified request without mutable parameters for CephFS
+	cephfsReq := &csi.CreateVolumeRequest{
+		Name:                      req.GetName(),
+		CapacityRange:             req.GetCapacityRange(),
+		VolumeCapabilities:        req.GetVolumeCapabilities(),
+		Parameters:                params,
```
iPraveenParihar: pass filteredParams?

iPraveenParihar (commented):
@nixpanic LGTM
just one concern to address.

internal/nfs/controller/controllerserver.go
nixpanic: yes, good catch!

iPraveenParihar (dismissed):
Thanks!

Madhu-1 (dismissed):

internal/nfs/controller/controllerserver.go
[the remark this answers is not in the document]
nixpanic: Addressed the `nil` return of `GetParameters()` with #6463.

nixpanic:
The NFS Controller gets parameters with CreateVolume, but should not pass these parameters on to CephFS CreateVolume. However, those parameters are critical for NFS functioning correctly, so those parameters need to be stored in the VolumeContext of the volume that the CephFS CreateVolume procedure returns.

Working on correcting this.

nixpanic:
@iPraveenParihar and @Madhu-1 , this PR should be ready now. NFS (and CephFS) passes e2e now too.

Once this is merged, run `@mergifyio rebase` on #6390 and queue is for e2e+merging again.

nixpanic:
@mergifyio rebase

gadididi (approved):

Madhu-1:
@mergifyio refresh

Madhu-1 (approved):

Madhu-1:
@Mergifyio refresh

PR #6463, opened 2026-08-06
cleanup: check GetParameters() for nil in CSI requests

nixpanic:
Copilot reported a potential panic when there are no parameters passed in CSI
requests. It is unlikely that a StorageClass or other sources of requests do
not contain parameters, but technically users could do this and cause Ceph-CSI
to crash.

See-also: https://github.com/ceph/ceph-csi/pull/6458#discussion_r3727192267

### DIFF
--- a/internal/csi-addons/rbd/replication.go
+++ b/internal/csi-addons/rbd/replication.go
@@ -265,7 +265,11 @@ func (rs *ReplicationServer) EnableVolumeReplication(ctx context.Context,
 	}
 	defer cr.DeleteCredentials()
 
-	err = validateSchedulingDetails(ctx, req.GetParameters())
+	parameters := req.GetParameters()
+	if parameters == nil {
+		return nil, status.Error(codes.InvalidArgument, "parameters cannot be nil")
+	}
+	err = validateSchedulingDetails(ctx, parameters)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


internal/csi-addons/rbd/replication.go
iPraveenParihar, line 478: Need check here too?

internal/csi-addons/rbd/volumegroup.go
iPraveenParihar, line 475: There are other replaces where we don't have nil check.

internal/csi-addons/rbd/replication.go
```diff
@@ -349,7 +353,11 @@ func (rs *ReplicationServer) DisableVolumeReplication(ctx context.Context,
 	}
 	defer rs.VolumeLocks.Release(volumeID)
 
-	mgr := rbd.NewManager(rs.driverInstance, req.GetParameters(), req.GetSecrets())
+	parameters := req.GetParameters()
+	if parameters == nil {
+		return nil, status.Error(codes.InvalidArgument, "parameters cannot be nil")
+	}
+	mgr := rbd.NewManager(rs.driverInstance, parameters, req.GetSecrets())
```
iPraveenParihar, line 360: @nixpanic, How about moving the check into `NewManager()` ?

iPraveenParihar (commented):
May be copilot is overthinking 😜, nil map reads don't panic in Go, so these are for better error messages rather than crash prevention.

nixpanic:
> May be copilot is overthinking 😜, nil map reads don't panic in Go, so these are for better error messages rather than crash prevention.

A, interesting! I was expecting this to panic, but it does not:
```go
package main

import (
	"fmt"
)

func getMap() map[string]any {
	return nil
}

func main() {
	m := getMap()
	v := m["panic"]

	fmt.Printf("no panic happened, value: %v\n", v)
}
```

Result:
```
$ go run nil-map.go
no panic happened, value: <nil>
```

It seems that adding a new key/value to the map `m` above does cause a panic.

I'll have a think about how to progress with this. Better error reporting is helpful, but maybe #6458 is sufficient for that already.

PR #6472, opened 2026-08-14
rbd: replace O(n) trash list scan with O(1) ID-based trash removal

Rakshith-R:

Replace `ensureImageCleanup()` - which scans the entire RBD trash list
to find an image by name - with direct O(1) removal by image ID using
the new `removeImageFromTrash()` helper.

The image ID is always available: modern volumes store it in OMAP at
creation time, and legacy volumes get it backfilled by `storeImageID()`
during the first `generateVolumeFromVolumeID()` call.

**Commit 1** - `rbd: capture parent image ID in getImageInfo`

Add `ParentImageID` field to `rbdImage` and populate it in
`getImageInfo()` from `GetParent().Image.ImageID`. Valid even when the
parent is in trash.

**Commit 2** - `rbd: make trash remove idempotent and add removeImageFromTrash`

Add `removeImageFromTrash()` as an O(1) helper that removes a trashed
image by its known ID. Make `trashRemoveImage()` treat ENOENT as success
on both the ceph mgr and direct paths (image already purged).

**Commit 3** - `rbd: avoid trash ls in DeleteVolume/DeleteSnapshot retry paths`

In `checkErrAndUndoReserve()` and `cleanUpImageAndSnapReservation()`,
switch from `ensureImageCleanup()` to `removeImageFromTrash()`. Also fix
a pre-existing `ErrorLog` format string mismatch (2 format verbs, 3 args).

**Commit 4** - `rbd: remove ensureImageCleanup and use ID-based trash removal`

In `DeleteTempImage()`, use the volume's `ParentImageID` to identify a
trashed temp clone instead of scanning the trash list. Remove
`ensureImageCleanup()` entirely - no callers remain.

- `DeleteTempImage()` relies on being called before `rbdVol.Delete()` so
  that `GetParent()` data is still valid. This ordering already exists in
  `cleanupRBDImage()` and is now documented with a comment.
- When parent info is unavailable in `DeleteTempImage()` (temp clone not
  in trash), it returns success - same behavior as the old code when the
  trash scan found nothing.

The change is backward compatible. No API or OMAP schema changes.

## Future concerns ##

- `GetTrashList` is still used in `waitForCleanTrash()` (e2e test helper).
  That is unrelated to production code paths.

**Checklist:**

  guidelines in the [developer
* [x] [Pending release
  N/A - internal optimization, no user-facing changes.
  N/A - no doc changes needed, internal refactor only.
  N/A - functions require a live Ceph cluster (librbd calls), not unit-testable.
  Existing e2e delete/snapshot tests exercise these code paths.
  N/A - existing e2e tests for DeleteVolume, DeleteSnapshot, and volume
  cloning already cover the trash cleanup paths exercised by this change.

---

<details>

</details>

### DIFF
--- a/internal/rbd/rbd_util.go
+++ b/internal/rbd/rbd_util.go
@@ -723,31 +723,6 @@ func isCephMgrSupported(ctx context.Context, clusterID string, err error) (bool,
 	return true, nil
 }
 
-// ensureImageCleanup finds image in trash and if found removes it
-// from trash.
-func (ri *rbdImage) ensureImageCleanup(ctx context.Context) error {
-	err := ri.openIoctx()
-	if err != nil {
-		return err
-	}
-
-	trashInfoList, err := librbd.GetTrashList(ri.ioctx)
-	if err != nil {
-		log.ErrorLog(ctx, "failed to list images in trash: %v", err)
-
-		return err
-	}
-	for _, val := range trashInfoList {
-		if val.Name == ri.RbdImageName {
-			ri.ImageID = val.Id
-
-			return ri.trashRemoveImage(ctx)
-		}
-	}
-
-	return nil
-}
-
 // Delete deletes a ceph image with provision and volume options.
 func (ri *rbdImage) Delete(ctx context.Context) error {
 	image := ri.RbdImageName
@@ -876,11 +851,20 @@ func (rv *rbdVolume) DeleteTempImage(ctx context.Context) error {
 	err = tempClone.Delete(ctx)
 	if err != nil {
 		if errors.Is(err, rbderrors.ErrImageNotFound) {
-			return tempClone.ensureImageCleanup(ctx)
-		} else {
-			// return error if it is not ErrImageNotFound
-			return err
+			if rv.ParentInTrash &&
+				rv.ParentName == tempClone.RbdImageName &&
+				rv.ParentImageID != "" {
+				tempClone.ImageID = rv.ParentImageID
+
+				return tempClone.removeImageFromTrash(ctx)
+			}
+
+			log.DebugLog(ctx, "temp clone %s not found and not in trash, already removed", tempClone)
+
+			return nil
 		}
+
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


YiteGu:
@Rakshith-R I've been thinking about how to solve this problem lately. This PR is very very nice.

internal/rbd/rbd_util.go
```diff
@@ -829,6 +818,19 @@ func (ri *rbdImage) trashRemoveImage(ctx context.Context) error {
...
+	if ri.ImageID == "" {
+		return fmt.Errorf("image ID is empty, cannot remove %s from trash", ri)
+	}
+
+	if err := ri.openIoctx(); err != nil {
+		return err
+	}
+
+	return ri.trashRemoveImage(ctx)
+}
```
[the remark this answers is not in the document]
Rakshith-R: This scenario cannot occur.

iPraveenParihar (approved):

nixpanic (approved):
Thanks!

internal/rbd/rbd_util.go
YiteGu, line 903: Now that we have ParentImageID, we need to add an assignment here.
`vol.ImageID = vol.ParentImageID`

YiteGu, line 601: If the imageID comes from the parentImageID, and the parentImage is in the trash, we should use `librbd.OpenImageById(ri.ioctx, ri.ImageID, librbd.NoSnapshot).`

Rakshith-R: > Now that we have ParentImageID, we need to add an assignment here. `vol.ImageID = vol.ParentImageID`

it's not used.

Rakshith-R: > If the imageID comes from the parentImageID, and the parentImage is in the trash, we should use `librbd.OpenImageById(ri.ioctx, ri.ImageID, librbd.NoSnapshot).`

There's no use case right now. We can add it when it is required.

YiteGu: One scenario I'm considering is that when obtaining clone depth, the presence of the parent image in the trash might cause inaccurate clone depth calculations.

Rakshith-R: if err != nil {
			// if the parent image is moved to trash the name will be present
			// in rbd image info but the image will be in trash, in that case
			// return the found depth
			if errors.Is(err, rbderrors.ErrImageNotFound) {
				return depth, nil
			}
			
There's a comment in code regarding this. 
Including trashed image in clonedepth calculation will lead to other complexities since trashed image cannot be flattened. Selective flattening logic in cephcsi is already complex. I would prefer it stay as is until there's a issue found relating to the scenario.

PR #6490, opened 2026-08-21
rbd, cephfs: return InvalidArgument for undecodable volume IDs in DeleteVolume

### LINKED ISSUE #6489: DeleteVolume retries forever with codes.Internal for static PVs with non-decodable volume IDs
# Describe the bug #

Deleting a PVC for a statically-provisioned RBD/CephFS volume whose
`volumeHandle` is a plain image/subvolume name, instead of ceph-csi's encoded
CSI identifier — causes `DeleteVolume` to fail with `codes.Internal`.
external-provisioner treats `codes.Internal` as retryable and retries forever,
so the PV is left stuck behind the
`external-provisioner.volume.kubernetes.io/finalizer` finalizer and is never
garbage collected. The only workaround is manually removing the finalizer.

# Steps to reproduce #

1. Create a static PV/PVC pair pointing at an existing RBD image, but set
   `volumeHandle` to the plain image name (e.g. `my-image`) rather than the
   encoded CSI identifier ceph-csi normally generates, and set
   `persistentVolumeReclaimPolicy: Delete`.
1. Delete the PVC.
1. `DeleteVolume` is called on the plugin and fails to decode the volume ID.

# Expected behavior #

`DeleteVolume` should return a terminal error (`codes.InvalidArgument`) when
the volume ID can't be decoded as a ceph-csi CSI identifier, since retrying
can never succeed. This is the same pattern already used for the identical
error in RBD's `ControllerExpandVolume` and `ControllerModifyVolume`, which
return `codes.InvalidArgument` for `ErrInvalidVolID` with a "likely a static
provisioned volume" comment. `DeleteVolume` (RBD and CephFS) doesn't special
case this error and falls through to `codes.Internal` instead.

# Actual results #

`DeleteVolume` returns:

"rpc error: code = Internal desc = invalid VolumeID: error decoding volume ID
(failed to decode CSI identifier, string underflow) (my-image)"

external-provisioner retries indefinitely since `codes.Internal` is treated as
retryable. The PV never gets garbage collected and stays stuck behind the
provisioner finalizer.

# Logs #

```
I0821 10:08:37.627043       1 event.go:389] "Event occurred" object="my-image" fieldPath="" kind="PersistentVolume" apiVersion="v1" type="Warning" reason="VolumeFailedDelete" message="rpc error: code = Internal desc = invalid VolumeID: error decoding volume ID (failed to decode CSI identifier, string underflow) (my-image)"
E0821 10:08:37.715373       1 controller.go:1558] "Volume deletion failed" err="rpc error: code = Internal desc = invalid VolumeID: error decoding volume ID (failed to decode CSI identifier, string underflow) (my-image)" PV="my-image"
```

# Additional context #

Note: `docs/static-pvc.md` documents that static PVs must use
`persistentVolumeReclaimPolicy: Retain`; a PV created with `Delete` (as in
this case) violates that documented requirement, but ceph-csi doesn't
currently validate this at PV-creation time.

fallmo:

`DeleteVolume` retries forever with `codes.Internal` when a static PV's
`volumeHandle` was never encoded by ceph-csi (e.g. a plain image/subvolume
name) and therefore can't be decoded as a CSI identifier. Since
`codes.Internal` is treated as retryable by external-provisioner, the PVC/PV
never gets cleaned up and is stuck behind the provisioner finalizer forever.

This PR makes both RBD's `checkErrAndUndoReserve` and CephFS's `DeleteVolume`
return `codes.InvalidArgument` for this case instead, since a volume ID that
fails to decode will never succeed on retry — it's a terminal error, not a
retryable one.

This mirrors the `ErrInvalidVolID` handling that already exists in RBD's
`ControllerExpandVolume` and `ControllerModifyVolume` — same error, same fix
pattern, just missing from `DeleteVolume`.

Scope: the RBD and CephFS drivers have the identical structural gap (an
unhandled `ErrInvalidVolID`/decode error falling through to
`codes.Internal`), so this PR fixes both rather than RBD only.

Backward compatible: yes. This only changes the returned gRPC status code for
volume IDs that already fail to decode and were never going to succeed;
well-formed volume IDs are unaffected.

## Related issues ##

Fixes: #6489

## Future concerns ##

* No admission webhook currently rejects `staticVolume: "true"` combined with
  `persistentVolumeReclaimPolicy: Delete` at PV-creation time, which is how
  this class of static PV ends up misconfigured in the first place. That's a
  separate, larger piece of work.

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

---

Prevent too many CI jobs from running.

Depends-on: #6533

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -931,6 +931,9 @@ func checkContentSource(
 		if err != nil {
 			rbdSnap.Destroy(ctx)
 			log.ErrorLog(ctx, "failed to get backend snapshot for %s: %v", snapshotID, err)
+			if errors.Is(err, rbderrors.ErrInvalidVolID) {
+				return nil, nil, status.Error(codes.InvalidArgument, err.Error())
+			}
 			if !errors.Is(err, rbderrors.ErrSnapNotFound) {
 				return nil, nil, status.Error(codes.Internal, err.Error())
 			}
@@ -952,6 +955,9 @@ func checkContentSource(
 		if err != nil {
 			rbdvol.Destroy(ctx)
 			log.ErrorLog(ctx, "failed to get backend image for %s: %v", volID, err)
+			if errors.Is(err, rbderrors.ErrInvalidVolID) {
+				return nil, nil, status.Error(codes.InvalidArgument, err.Error())
+			}
 			if !errors.Is(err, rbderrors.ErrImageNotFound) {
 				return nil, nil, status.Error(codes.Internal, err.Error())
 			}
@@ -1248,6 +1254,8 @@ func (cs *ControllerServer) CreateSnapshot(
 		case errors.Is(err, util.ErrPoolNotFound):
 			log.ErrorLog(ctx, "failed to get backend volume for %s: %v", req.GetSourceVolumeId(), err)
 			err = status.Error(codes.NotFound, err.Error())
+		case errors.Is(err, rbderrors.ErrInvalidVolID):
+			err = status.Error(codes.InvalidArgument, err.Error())
 		default:
 			err = status.Error(codes.Internal, err.Error())
 		}
@@ -1575,6 +1583,13 @@ func (cs *ControllerServer) DeleteSnapshot(
 	if err != nil {
 		rbdSnap.Destroy(ctx)
 
+		if errors.Is(err, rbderrors.ErrInvalidVolID) {
+			// likely a static provisioned snapshot with a snapshot handle
+			// that was never encoded by ceph-csi; retrying will never
+			// succeed, so return a terminal error instead of codes.Internal.
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+
 		// if error is ErrPoolNotFound, the pool is already deleted we don't
 		// need to worry about deleting snapshot or omap data, return success
 		if errors.Is(err, util.ErrPoolNotFound) {
@@ -1813,6 +1828,10 @@ func (cs *ControllerServer) getServiceAccountRestriction(
 			rv.Destroy(ctx)
 		}
 
+		if errors.Is(err, rbderrors.ErrInvalidVolID) {
+			return "", status.Errorf(codes.InvalidArgument, "volume ID %s does not support publish: %v", volumeID, err)
+		}
+
 		return "", status.Errorf(codes.Internal,
 			"failed to find volume %q for service account check: %v", volumeID, err)
 	}
@@ -1872,6 +1891,11 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
 	rv, err := GenVolFromVolID(ctx, volumeId, credentials, secrets)
 	if err != nil {
 		rv.Destroy(ctx)
+
+		if errors.Is(err, rbderrors.ErrInvalidVolID) {
+			return nil, status.Errorf(codes.InvalidArgument, "volume ID %s does not support unpublish: %v",
+				volumeId, err)
+		}
 
 		return nil, status.Errorf(codes.Internal, "failed to generate volume from volume ID %s: %v",
 			volumeId, err)
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -992,6 +992,13 @@ func (cs *ControllerServer) checkErrAndUndoReserve(
 		return &csi.DeleteVolumeResponse{}, nil
 	}
 
+	if errors.Is(err, rbderrors.ErrInvalidVolID) {
+		// likely a static provisioned volume with a volume handle that was
+		// never encoded by ceph-csi; retrying will never succeed, so return
+		// a terminal error instead of codes.Internal.
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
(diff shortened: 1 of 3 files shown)


Madhu-1 (dismissed):

fallmo:
I went ahead and redid the commit messages to fix the commitlint CI error

iPraveenParihar (commented):
LGTM,
Can you add similar check at other places like DeleteSnapshot()?
https://github.com/ceph/ceph-csi/blob/ddea16a0d9a8b0c706f980579d3b42cb34b8349a/internal/cephfs/controllerserver.go#L1059-L1099

fallmo:
> LGTM, Can you add similar check at other places like DeleteSnapshot()?
> 

Good point, I'll look through

fallmo:
Hey @iPraveenParihar, I added the check in checkContentSource, CreateSnapshot, DeleteSnapshot, getServiceAccountRestriction, and ControllerUnpublishVolume.

I also found that DeleteVolumeGroupSnapshot and GetVolumeGroupSnapshot need the same fix, but couldn't do it there cause both resolve the group-snapshot ID through initCommonVolumeGroup (internal/rbd/group/util.go), which wraps a decode failure in a plain fmt.Errorf instead of rbderrors.ErrInvalidVolID. So there's no sentinel for errors.Is to match at the RPC layer yet. Fixing it would mean wrapping that decode error in ErrInvalidVolID first (same pattern GenVolFromVolID already uses in rbd_util.go), then adding the errors.Is case in both RPCs. Should I go ahead?

fallmo:
hey @Madhu-1 @iPraveenParihar, any updates on this issue?

Madhu-1 (dismissed):

iPraveenParihar:
@fallmo Can you resolve the golintci
```
../internal/cephfs/controllerserver.go:1026:1: calculated cyclomatic complexity for function DeleteSnapshot is 21, max is 20 (cyclop)
func (cs *cephfsControllerServer) DeleteSnapshot(
^
1 issues:
* cyclop: 1
* 
```

fallmo:
> @fallmo Can you resolve the golintci
> 

Yes, it should be resolved now. The additional switch case moved the func complexity from 20 to 21. The switch case was moved to a new  `handleSnapshotOptionsError` function.

internal/rbd/controllerserver.go
```diff
@@ -1809,6 +1824,10 @@ func (cs *ControllerServer) getServiceAccountRestriction(
 			rv.Destroy(ctx)
 		}
 
+		if errors.Is(err, rbderrors.ErrInvalidVolID) {
+			return "", status.Errorf(codes.InvalidArgument, "volume ID %s does not support publish: %v", volumeID, err)
```
iPraveenParihar, line 1832: We won't reach here, GetControllerPublishSecrets() will return skip=true for ErrInvalidVolID already

internal/rbd/controllerserver.go
```diff
@@ -1869,6 +1888,11 @@ func (cs *ControllerServer) ControllerUnpublishVolume(
 	if err != nil {
 		rv.Destroy(ctx)
 
+		if errors.Is(err, rbderrors.ErrInvalidVolID) {
+			return nil, status.Errorf(codes.InvalidArgument, "volume ID %s does not support unpublish: %v",
```
iPraveenParihar, line 1896: Same here, check `secrets, skip, err := util.GetControllerPublishSecrets` path

iPraveenParihar (commented):
Looks good, few nit

fallmo-axl:
> Looks good, few nit

Thanks for the review. I looked at the `GetControllerPublishSecrets` function. It looks like it won't necessarily always return skip true for invalid volume ID. When the `reqSecrets` are provided, it just goes through. 

```golang
	if reqSecrets != nil {
		return reqSecrets, false, nil
	}
	secretName, secretNamespace, err := GetControllerPublishSecretRef(volumeID, driverType)
	if errors.Is(err, ErrInvalidVolID) || errors.Is(err, ErrConfigNotFound) {
		// Possibly the volume is a static/older volume. In this case, we have nothing to do.
		// Even if it's not a static/older volume, we should skip handling it because
		// we have no way to process handling anyway.
		log.WarningLog(ctx, "should skip handling this volume: %v", err)

		return nil, true, nil
	}
```

fallmo:
> Looks good, few nit

Thanks for the review. I looked at the `GetControllerPublishSecrets` function. It looks like it won't necessarily always return skip true for invalid volume ID. When the `reqSecrets` are provided, it just goes through. 

```golang
	if reqSecrets != nil {
		return reqSecrets, false, nil
	}
	secretName, secretNamespace, err := GetControllerPublishSecretRef(volumeID, driverType)
	if errors.Is(err, ErrInvalidVolID) || errors.Is(err, ErrConfigNotFound) {
		// Possibly the volume is a static/older volume. In this case, we have nothing to do.
		// Even if it's not a static/older volume, we should skip handling it because
		// we have no way to process handling anyway.
		log.WarningLog(ctx, "should skip handling this volume: %v", err)

		return nil, true, nil
	}
```

iPraveenParihar (approved):

Madhu-1 (approved):

PR #6497, opened 2026-08-25
Feat/nfs export names

### LINKED ISSUE #4414: nfs: allow users to specify NFS export name

Currently, NFS export name is generated by `ComposeCSIID()` and the generated name is a long alphanumeric string. We hope we can specify a meaningful name for a NFS export.

No priority, but I think it's good to have. Users can specify a simple and meaningful NFS export name as they want and it would be more convenient for users to manage NFS exports.

I know that we can pass parameters in StorageClass to CSI drivers. However, specifying NFS export name in StorageClass seems not a good idea. Not sure if it's possible to pass PVC name to CSI driver?

IanEff:

NFS export names are currently derived from `ComposeCSIID()`, a long alphanumeric string that's meaningless to humans.  This PR adds an opt-in `friendlyExportNames` StorageClass parameter that names exports `<namespace>/<pvc-name>` instead.

The name is sourced from PVC metadata supplied by the external provisioner, enabled with `--extra-create-metadata=true`, and persisted in the CephFS journal so that `DeleteExport` and `SetClients` can resolve the volume path later on.

This feature is gated behind an explicit StorageClass parameter rather than firing automatically whenever `--extra-create-metadata` is set because that flag is already used unconditionally by the CephFS backend for per-tenant KMS scoping.

It's entirely backwards compatible, as the parameter is set to off by default.  No existing exports are affected

## Related issues ##

Fixes: #4414

## Future concerns ##

None.

**Checklist:**

  guidelines in the [developer
* [x] [Pending release

---

<details>

</details>

---

CI job ordering.

Depends-on: #6512

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -22,5 +22,13 @@
       configuration
     - Backward compatibility with existing cluster-mapping.json via
       ClientProfileMapping integration
+1. Added a `friendlyExportNames` StorageClass parameter for the NFS driver.
+   When set to `"true"` and the external-provisioner runs with
+   `--extra-create-metadata=true`, NFS-exports are named
+   `<namespace>/<pvc-name>` instead of the generated volume ID. Off by
+   default, and gated behind the parameter rather than the provisioner flag
+   alone, since CephFS already reads that same metadata unconditionally for
+   per-tenant KMS scoping.  Existing StorageClasses keep today's export
+   names unless they opt in explicitly.
 
 ## NOTE
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -29,6 +29,11 @@
    default, and gated behind the parameter rather than the provisioner flag
    alone, since CephFS already reads that same metadata unconditionally for
    per-tenant KMS scoping.  Existing StorageClasses keep today's export
-   names unless they opt in explicitly.
+   names unless they opt in explicitly. Unlike the generated volume ID,
+   `<namespace>/<pvc-name>` is not guaranteed unique over time (e.g. a PVC
+   recreated under the same name before its old export was cleaned up);
+   `CreateVolume` now fails with `AlreadyExists` rather than silently
+   reusing another volume's export if the name is already claimed by a
+   different subvolume.
 
 ## NOTE
--- a/internal/nfs/controller/controllerserver.go
+++ b/internal/nfs/controller/controllerserver.go
@@ -19,8 +19,10 @@ package controller
 import (
 	"context"
 	"errors"
+	"fmt"
 	"maps"
 	"slices"
+	"strconv"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
@@ -33,6 +35,7 @@ import (
 	"github.com/ceph/ceph-csi/internal/journal"
 	nfs "github.com/ceph/ceph-csi/internal/nfs/types"
 	"github.com/ceph/ceph-csi/internal/util"
+	"github.com/ceph/ceph-csi/internal/util/k8s"
 	"github.com/ceph/ceph-csi/internal/util/log"
 )
 
@@ -54,6 +57,7 @@ type nfsControllerServer struct {
 var nfsParameters = []string{
 	nfs.ParameterCluster,
 	nfs.ParameterSecTypes,
+	nfs.ParameterFriendlyExportNames,
 }
 
 // nfsMutableParameters contains the list of mutable parameters for NFS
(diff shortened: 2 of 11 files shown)


IanEff:
This doesn't address the degenerate export names issue.  I'll re-submit.

internal/nfs/types/volume.go
```diff
@@ -359,6 +346,19 @@ func (nv *NFSVolume) SetClients(clients string) error {
 	return nil
 }
 
+// resolveExportName loads a previously persisted friendly export name
+// (SetExportName) from the CephFS journal into nv.exportName.
+func (nv *NFSVolume) resolveExportName() error {
```
nixpanic, line 365: I wonder if it makes sense to do this inside `Connect()`. Is there a good reason not to do that?

nixpanic (commented):
Looks reasonable, thanks for the PR.

Can you please create a clean series of commits, instead of fixups that tidy the code after it was added? Note that we do want `Co-authored-by` referencing your AI agent on each commit where it assisted.

It would be good to add a test-case in `e2e/nfs.go` for the functionality too.

IanEff:
Both of those CI tests failed on downloading from the repo(https://download.ceph.com/rpm-${CEPH_VERSION}/el10/noarch/ceph-release-1-0.el10.noarch.rpm). I don't know that I can do anything about that.

internal/nfs/types/volume.go
IanEff: If we called it in Connect(), we'd take an extra CephFS journal read just to populate nv.exportName on every single volume creation, even if friendlyExportNames isn't set.  This is immediately overwritten by SetExportName() anyway.

The two places that consume nv.exportName, DeleteExport() and SetClients(), already call resolveExportName at the top.  This way loads the journal state lazily, only where it's read.

IanEff:
> Can you please create a clean series of commits, instead of fixups that tidy the code after it was added? 

Resubmitted.  Thanks for the guidance!

nixpanic (dismissed):
Thanks, this looks complete to me. Unfortunately we do not have any [documentation](https://ceph.github.io/ceph-csi/) specific to NFS (or other backends), otherwise it would have been nice to add this option and an example there.

nixpanic:
@mergifyio rebase

This will include the CI fixes.

nixpanic:
GitHub Actions are not running due to an outage

image

nixpanic:
@mergifyio rebase

GitHub Workflows should work again.

Madhu-1 (dismissed):

IanEff:
@Madhu-1 , @nixpanic: all Copilot issues addressed in the last PR.

Madhu-1 (dismissed):

internal/nfs/controller/controllerserver.go
```diff
@@ -104,6 +108,32 @@ func (cs *nfsControllerServer) ValidateVolumeCapabilities(
 	return cs.backendServer.ValidateVolumeCapabilities(ctx, req)
 }
 
+// friendlyExportName returns the "<namespace>/<pvc-name>" export name to
+// use, or "" if ParameterFriendlyExportNames is unset or "false".
+func friendlyExportName(nfsParams, csiParams map[string]string) (string, error) {
+	friendlyExportNames := false
+	if v, ok := nfsParams[nfs.ParameterFriendlyExportNames]; ok {
+		var err error
+		friendlyExportNames, err = strconv.ParseBool(v)
```
iPraveenParihar:
```suggestion
		friendlyExportNames, err := strconv.ParseBool(v)
```

and remove `var err error`

internal/nfs/controller/controllerserver.go
```diff
@@ -104,6 +108,32 @@ func (cs *nfsControllerServer) ValidateVolumeCapabilities(
...
+		if err != nil {
+			return "", fmt.Errorf("invalid %s: %w", nfs.ParameterFriendlyExportNames, err)
```
iPraveenParihar:
```suggestion
			return "", fmt.Errorf("invalid value for %s:  %s, (should be a \"true\" or \"false\"): %w", nfs.ParameterFriendlyExportNames, v, err)
```

PendingReleaseNotes.md
```diff
@@ -8,12 +8,20 @@
...
-    - Backward compatibility with existing cluster-mapping.json via
-      ClientProfileMapping integration
+   - Volume replication: Maps source volume ID to destination volume ID
+   - Volume group replication: Maps source group ID and all member volume
+     IDs to their destination IDs
+   - Pool name-based mapping via `replicationDestination` ConfigMap
+     configuration
+   - Backward compatibility with existing cluster-mapping.json via
+     ClientProfileMapping integration
+1. Added a `friendlyExportNames` StorageClass parameter for the NFS driver.
```
iPraveenParihar, line 25: nit, keep the other note untouched.  :)

iPraveenParihar (changes requested):
Thanks! LGTM :)
have more info on the error logs

PendingReleaseNotes.md
IanEff: Sorry about that!  Re-submitting.

IanEff: Good catch — the reformatting of note 1 was an unintended side effect. Squashed the fix into the original commit and force-pushed rather than adding a new commit on top.

internal/nfs/controller/controllerserver.go
iPraveenParihar: @IanEff `var err error` still not removed

iPraveenParihar (commented):
@IanEff squash commits into one?

IanEff:
@iPraveenParihar Squashed!

iPraveenParihar (dismissed):

Madhu-1 (dismissed):

nixpanic:
Wait for #6512 to be merged before starting CI jobs in this PR.

nixpanic (approved):

iPraveenParihar (approved):

iPraveenParihar (approved):

PR #6501, opened 2026-08-25
cephfs: refresh recursive stats before volume stats

baldugus:

Refreshes CephFS recursive statistics before collecting filesystem usage.

Linux CephFS `statfs()` overrides filesystem statistics for quota-backed mounts using the quota inode's cached `rbytes` value ([source](https://github.com/torvalds/linux/blob/26260251022fbc2f248a3d747a9b2b961b18d2d8/fs/ceph/quota.c#L531)) instead of using the fresh values like the regular `statfs()` path  ([source](https://github.com/torvalds/linux/blob/26260251022fbc2f248a3d747a9b2b961b18d2d8/fs/ceph/super.c#L74)).

On the other hand, Ceph MDS only broadcasts new statistics when usage changes by approx. 1/16 of the difference between the last reported usage and the max quota available ([source](https://github.com/ceph/ceph/blob/162eb4aeca88382affe501526ab04c3bcd087dad/src/mds/MDCache.cc#L2129)).

Slow growing volumes can remain for long periods with stale usage and consequently stale `kubelet_volume_stats_used_bytes`. In my cluster I noticed my metrics database had several PVCs reporting the same used bytes for a whole month, despite constant write activity.

Neither behavior is wrong on its own, so I wrote this patch intending to make the statistics accurate in ceph-csi.

I've ran the added E2E test on the latest commit at devel (f4ce351) and in this PR and I could reproduce a fail before and a pass after.

I've also replaced my production ceph-csi with this patch and all affected volumes started reporting the current usage.

This is my manual reproduction before the patch on my kubernetes host:

```bash
$ df -B1 <kubelet globalmount path>
Filesystem 1-blocks Used Available Use% Mounted on
<path> 10737418240 0 10737418240 0% <path>

$ # fetched ceph.dir.rbytes using a perl script as the host lacks getfattr
203487451

$ df -B1 <kubelet globalmount path>
Filesystem 1-blocks Used Available Use% Mounted on
<path> 10737418240 201326592 10536091648 2% <path>
```

This is my first contribution to ceph-csi. I believe this change is backward compatible.

If this is accepted, is it worth updating the pending release notes? I was unsure about it, so I didn't include an update in this PR.

## Related issues ##

#2579

## Future concerns ##

This adds an MDS request when collecting CephFS volume statistics. If a CephFS-side change makes `statfs()` sufficient in the future, this patch could be reconsidered.

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

### DIFF
--- a/e2e/cephfs.go
+++ b/e2e/cephfs.go
@@ -27,6 +27,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


nixpanic (dismissed):
Thanks, looks good.

baldugus:
@nixpanic 

The tests had failed, but I saw that you fixed the ceph-release RPM URL yesterday, so I've rebased my branch onto the latest commits.

Madhu-1 (approved):

nixpanic (approved):

PR #6506, opened 2026-08-26
doc: add encrypted-volumes.md feature document

nixpanic:
Move the encryption sections (life-cycle, LUKS configuration,
metadata KMS, Vault, Amazon KMS/STS, Azure Key Vault, KMIP) out
of docs/rbd/deploy.md into a new top-level feature document
docs/encrypted-volumes.md so it can be referenced independently
of the RBD deployment guide.

Add a dedicated 'CephFS volume encryption' section to
docs/encrypted-volumes.md covering:

* how to enable fscrypt encryption via the StorageClass
* which KMS backends are compatible (raw-passphrase only)
* the volume layout (/.fscrypt and /ceph-csi-encrypted)
* kernel and Ceph client prerequisites

Replace the sparse 'CephFS Volume Encryption' section in
docs/cephfs/deploy.md with a short summary and a link to the
new section, consistent with the existing RBD pattern.


docs/encrypted-volumes.md
```diff
@@ -0,0 +1,380 @@
...
+backends are supported for fscrypt. Only KMS backends that expose the raw
+passphrase or secret directly to the driver work:
+
+* Kubernetes Secrets (`encryptionKMSType: "secrets"`)
+* `metadata` type (`encryptionKMSType: "metadata"`)
+* HashiCorp Vault (`encryptionKMSType: "vault"` / `"vaulttokens"` /
+  `"vaulttenantsa"`)
+
+KMS backends that only wrap/unwrap keys without exposing a raw secret (e.g.
+AWS KMS, Azure Key Vault) are **not** compatible with fscrypt.
```
black-dragon74: Azure is `DEKStoreIntegrated`. I think it is wrongly included here.

docs/encrypted-volumes.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+### KMS compatibility
+
+The KMS configuration is shared with RBD encryption and the same
+`ceph-csi-encryption-kms-config` ConfigMap is used. However, not all KMS
+backends are supported for fscrypt. Only KMS backends that expose the raw
+passphrase or secret directly to the driver work:
+
+* Kubernetes Secrets (`encryptionKMSType: "secrets"`)
```
black-dragon74: There is no `encryptionKMSType: secrets`. Maybe you meant: `default`? In that case, just set `encrypted: true` and omit setting `encryptionKMSType` altogether.

The passphrase will be used from provisioner/node `secret.encryptionPassphrase`

black-dragon74 (approved):
Thanks, LGTM. One thing I'd check is if MkDocs is rendering relative links correctly. 

_Maybe update the outliers to use absolute links as well_

PR #6514, opened 2026-08-27
ci: create the configmap before deploying the CSI plugins

nixpanic:
The ceph-csi-config configmap is mounted by the CSI plugin pods.
Creating it after deploying the operator causes a race where pods
can fail to start due to a missing volume reference.

Move the 'create ConfigMap & StorageClasses' stage before the
'deploy ceph-csi through operator' stage. The namespace creation
is moved into the configmap stage since it must exist before
create-configmap.sh can run.


nixpanic:
@gadididi , this should address the delays because of the missing configuration file when a volume is created the 1st time.

iPraveenParihar (approved):

black-dragon74 (approved):

Rakshith-R (approved):

PR #6521, opened 2026-09-01
kms: support CephFS fscrypt encryption with a KMIP KMS

### LINKED ISSUE #6324: cephfs: Add KMIP support to CephFS fscrypt encryption
Hello,

CephFS encryption with KMIP as the KMS backend doesn't work. I'd like to add this support and am opening this issue to get community input on which of two design directions to take: either let KMIP keep using envelope encryption (where the actual encryption key is stored in the volume itself and unwrapped by KMIP on demand), or have KMIP store the encryption key directly (similar to how the existing Vault integration works). Before writing code I'd like to confirm my understanding of the problem and hear which approach the maintainers prefer.

## What is the value to the end user? (why is it a priority?)

KMIP (Key Management Interoperability Protocol) is a vendor-neutral standard for applications to interact with key management systems, and is the interface most HSM appliances expose. ceph-csi already supports KMIP for RBD volume encryption, but the same KMS cannot back CephFS encryption today. Users whose only key management option is a KMIP-fronted HSM are therefore forced to either drop CephFS encryption or run HashiCorp Vault as a KMIP passthrough.

## Context

All KMS implementations in ceph-csi are either of the type `DEKStoreIntegrated` or `DEKStoreMetadata`. The following code comments explain what these types mean.

```go
// ceph-csi internal/kms/kms.go

// DEKStoreType describes what DEKStore needs to be configured when using a
// particular KMS. A KMS might support different DEKStores depending on its
// configuration.
type DEKStoreType string

const (
    // DEKStoreIntegrated indicates that the KMS itself supports storing
    // DEKs.
    DEKStoreIntegrated = DEKStoreType("")
    // DEKStoreMetadata indicates that the KMS should be configured to
    // store the DEK in the metadata of the volume.
    DEKStoreMetadata = DEKStoreType("metadata")
)
```

## Is-State

As you can see in the following code snippet the `ConfigureEncryption()` function anticipates that some KMS implementations do not implement the `GetSecret()` function which is needed to use fscrypt (kernel module which implements file-level encryption — block storage equivalent is dmcrypt).

```go
// ceph-csi internal/cephfs/store/volumeoptions.go

// ConfigureEncryption initializes the Ceph CSI key management from
// kmsID and credentials. Sets vo.Encryption on success.
func (vo *VolumeOptions) ConfigureEncryption(
    ctx context.Context,
    kmsID string,
    credentials map[string]string,
) error {
    kms, err := kmsapi.GetKMS(vo.Owner, kmsID, credentials)
    if err != nil {
        log.ErrorLog(ctx, "get KMS failed %+v: %v", vo, err)

        return err
    }
    // With cephfs selecting the cipher is not possible
    vo.Encryption, err = util.NewVolumeEncryption(kmsID, kms, nil)

    if errors.Is(err, util.ErrDEKStoreNeeded) {
        // fscrypt uses secrets directly from the KMS.
        // Therefore we do not support an additional DEK
        // store. Since not all "metadata" KMS support
        // GetSecret, test for support here. Postpone any
        // other error handling
        _, err := vo.Encryption.KMS.GetSecret(ctx, "")
        if errors.Is(err, kmsapi.ErrGetSecretUnsupported) {
            return err
        }
    }

    return nil
}
```

When we look for the KMIP KMS implementation we can clearly see that the `GetSecret()` function is not implemented hence the `ConfigureEncryption()` function will fail if KMIP is the KMS backend for ceph-csi cephfs encryption.

```go
// ceph-csi internal/kms/kmip.go
func (kms *kmipKMS) RequiresDEKStore() DEKStoreType {
    return DEKStoreMetadata
}

func (kms *kmipKMS) GetSecret(ctx context.Context, volumeID string) (string, error) {
    return "", ErrGetSecretUnsupported
}
```

The `kmipKMS` struct is of the `DekStoreType` `DEKStoreMetadata` which "indicates that the KMS should be configured to store the DEK in the metadata of the volume." However `kmipKMS` never actually needs to retrieve the secret via `GetSecret()`, instead getting the secret is done via the `DecryptDEK()` implementation as seen in the following code snippet.

```go
// ceph-csi internal/kms/kmip.go

// DecryptDEK calls either GET or DECRYPT operation on the KMIP kms to decrypt the DEK.
// The operation to call depends upon the config value `USE_CRYPTO_RPC`.
//
// If USE_CRYPTO_RPC == false : GET operation is used.
// If USE_CRYPTO_RPC == true : DECRYPT operation is used.
func (kms *kmipKMS) DecryptDEK(ctx context.Context, volumeID, encryptedDEK string) (string, error) {
    if kms.useCryptoRPC {
        return kms.decryptDEKUsingDecryptRPC(ctx, volumeID, encryptedDEK)
    }

    return kms.decryptDEKUsingRemoteKey(ctx, volumeID, encryptedDEK)
}
```

Thus the `GetSecret()` function did not need to be implemented to retrieve the secret because the `DecryptDEK()` retrieves the secret and decrypts it in one go.

## Should-State

Practically there are two ways KMIP could work in concert with fscrypt. The following code snippet shows the core interaction between the ceph-csi fscrypt logic and the KMS backend.

```go
// ceph-csi internal/util/fscrypt/fscrypt.go
switch encryption.KMS.RequiresDEKStore() {
case kms.DEKStoreIntegrated:
    passphrase, err = encryption.GetCryptoPassphrase(ctx, volID)
case kms.DEKStoreMetadata:
    passphrase, err = encryption.KMS.GetSecret(ctx, volID)
}
```

So to make KMIP work with ceph-csi fscrypt, one of two things needs to happen:

1. Keep KMIP as a `DEKStoreMetadata` and implement its `GetSecret()` method, or
2. Add a new (or extended) KMIP struct of type `DEKStoreIntegrated` which can use `GetCryptoPassphrase()` (analogous to how the `vaultKMS` implementation works).

## Questions

I understand that from an fscrypt perspective there is no KEK, DEK (envelope encryption) needed. The KMS system just needs to store a key which can directly be used by fscrypt. This would correspond to the `DEKStoreIntegrated` semantics.

1. **Do we want to implement this key model for ceph-csi fscrypt with KMIP** — i.e. implement a new KMIP struct that is a `DEKStoreIntegrated` key store?

2. **Alternatively, we could store the fscrypt DEK as an encrypted DEK on cephfs metadata and decrypt it in the KMIP KMS.** This would match the current `DEKStoreType` the KMIP struct implements (`DEKStoreMetadata`). Would this be a better approach?

Greenpepper15:

This PR makes a KMIP KMS usable for CephFS fscrypt encryption in the most
minimal way possible by implementing `GetSecret` on the existing `kmip`
provider.

`ConfigureEncryption` probes the KMS with `GetSecret` and rejects the
configuration when the KMS answers `ErrGetSecretUnsupported`, which the
`kmip` provider always did. The new implementation returns the key material
of the managed symmetric key, base64 encoded, and fscrypt uses it as a
`custom_passphrase`.

Fetching the key material requires the KMIP `Get` operation, so the change
is gated on `USE_CRYPTO_RPC` being `"false"`.

The base64 encoding keeps the passphrase printable so that a volume remains
recoverable with the upstream `fscrypt` tool by running `base64` on the key
exported from the KMS. I chose base64.StdEncoding but another ceph-csi
secret encoding uses base64.URLEncoding (see question section for the
motivation).

RBD with `encryptionType: file` would pass the same capability probe and use
the same `fscrypt.Unlock`, but that combination is untested, so this PR
explicitly rejects it in `configureFileEncryption` with an error. RBD
behavior is therefore unchanged.

Do you have any questions?

* **E2e placement:** should the two e2e commits move to a separate PR, or is
  keeping them here okay? Please check the KMIP e2e I added.
* **Passphrase encoding:** should the passphrase use `StdEncoding` or the
  URL-safe alphabet that `generateNewEncryptionPassphrase` uses? fscrypt
  derives the protector key from the exact passphrase bytes, so the encoding
  can never change afterwards. Either would work within the driver. What
  breaks with `URLEncoding` is only the manual recovery path with the
  `base64(1)` tool: `base64(1)` emits the standard alphabet, so a
  by-the-book recovery would derive a different passphrase for any key whose
  bytes encode to `+` or `/`, and the unlock will fail for some keys.
  `StdEncoding` is therefore a recovery-contract choice, not a within-driver
  correctness constraint.
* **Provider naming:** should `GetSecret` own the bare `kmip` provider name,
  or use a different name to keep the bare name free for other KMIP
  integrations? This PR makes the bare `kmip` provider hand out key material
  through `GetSecret`.
* **KMIP test server:** should the KMIP e2e use PyKMIP? Almost nothing else
  fits: the test server must speak exactly what the kmip provider speaks,
  run in a plain Pod without a license, and let a script create a key.
  PyKMIP is what Ceph itself tests the RGW KMIP encryption against
  (`qa/tasks/pykmip.py` in the teuthology suite), and what OpenStack
  Barbican's functional gate uses for its KMIP plugin. The one maintained
  open-source alternative is Cosmian KMS (BUSL 1.1, KMIP 1.0-2.1, binary
  socket server with TLS client authentication). If you prefer a maintained
  implementation, I can switch the e2e to it.

  PyKMIP 0.10.0 is the last release, so the pod pins the last dependency
  versions it works with (Python 3.11, SQLAlchemy 1.x, cryptography
  41.0.7). Newer cryptography removed the legacy ciphers PyKMIP imports
  and the server crashes on startup.

Testing caveat: I could not run the new e2e spec on my setup, because it
needs a Ceph cluster together with a CephFS fscrypt capable kernel. The
spec itself has therefore not been executed and I am piggybacking the
ceph-csi CI to exercise it. Note that no CI job currently sets
`--test-cephfs-fscrypt`, so a run needs that flag. What I did verify
against a live PyKMIP server on a scratch cluster is the deployment and
provisioning flow.

* Existing `kmip` users are unaffected: `RequiresDEKStore`, `EncryptDEK` and
  `DecryptDEK` are untouched, there is no configuration schema change, and
  the only behavioral change is that a configuration that used to fail now
  succeeds.
* A CephFS volume encrypted with a KMIP KMS by this release cannot be staged
  by an older release, which still answers the capability probe with
  `ErrGetSecretUnsupported`.
* Rotating or destroying the KMIP key, or pointing `UNIQUE_IDENTIFIER` at a
  different key, makes every existing volume permanently unopenable. There
  is no re-wrap path.

## Related issues ##

Fixes: #6324

## Future/now concerns ##

There are different ways to add KMIP support to CephFS fscrypt. I chose the
most minimal one, but the other directions remain possible and might make
more sense for different use cases.

Planned follow-ups:

* validate RBD `encryptionType: file` with KMIP (an RBD e2e spec) and remove
  the lock in `configureFileEncryption`.

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>


internal/kms/kmip.go
```diff
@@ -230,8 +231,25 @@ func (kms *kmipKMS) RequiresDEKStore() DEKStoreType {
...
+			ErrGetSecretUnsupported, kmipUseCryptoRPC)
+	}
+
+	key, err := kms.getKey(kms.uniqueIdentifier)
+	if err != nil {
+		return "", fmt.Errorf("failed to get key %q: %w", kms.uniqueIdentifier, err)
+	}
+
+	return base64.StdEncoding.EncodeToString(key), nil
+}
```
Greenpepper15, line 247: Dear ceph-csi maintainers,

This PR changes pretty big but the core change "just" involves this one function.

The rest are changes revolve around testing this change.

Greenpepper15: @Rakshith-R

black-dragon74:
@Greenpepper15 could you please add e2e tests around fscrypt PVCs?

e2e/cephfs.go
```diff
@@ -618,6 +624,7 @@ var _ = Describe(cephfsType, func() {
 				"vault-test":            vaultKMS,
 				"vault-tokens-test":     vaultTokensKMS,
 				"vault-tenant-sa-test":  vaultTenantSAKMS,
+				"kmip-fscrypt-test":     kmipKMS,
 			}
 
 			for kmsID, kmsConf := range kmsToTest {
```
Greenpepper15, line 607: > @Greenpepper15 could you please add e2e tests around fscrypt PVCs?

@black-dragon74 I integrated this new KMIP fscrypt integration into the existing fscrypt KMS e2e testing suite. 

I can add the kmip fscrypt testing to more KMS testing scenarios though like "encrypted PVC-PVC cloning" and "encrypted snapshot" testing. Is that fine?

Greenpepper15:
Also it seems like the test run failed due to flake in the NVMe e2e testing suite. 

```
• [FAILED] [187.180 seconds]
nvmeof Test NVMe CSI [It] test volumeGroupSnapshot
/go/src/github.com/ceph/ceph-csi/e2e/nvmeof.go:582
  [FAILED] failed to execute rados command 'rados listomapkeys csi.groups.default --pool=nvmeofpool | wc -l' : stdErr=2026-09-08T08:38:40.714+0000 7f3883fff640 -1 monclient: get_auth_request but no auth handler is set up
  In [It] at: /go/src/github.com/ceph/ceph-csi/e2e/utils.go:2150 @ 09/08/26 08:38:40.725
------------------------------
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Summarizing 1 Failure:
  [FAIL] nvmeof Test NVMe CSI [It] test volumeGroupSnapshot
  /go/src/github.com/ceph/ceph-csi/e2e/utils.go:2150
Ran 121 of 195 Specs in 7088.765 seconds
FAIL! -- 120 Passed | 1 Failed | 0 Pending | 74 Skipped
```

Greenpepper15:
Also when looking at the logs I can see these messages

```
rbd.go:309] skipping RBD fscrypt file encryption test
```

I think the fscrypt tests are skipped for RBD: 
```
[0mRBD [38;5;243mTest RBD CSI [0m[1mcreate a PVC and bind it to an app with encrypted RBD volume with VaultKMS (file)[0m
[38;5;243m/go/src/github.com/ceph/ceph-csi/e2e/rbd.go:307[0m
  [1mSTEP:[0m Creating a kubernetes client [38;5;243m@ 09/08/26 07:19:40.13[0m
  I0908 07:19:40.130161   88594 util.go:414] >>> kubeConfig: /root/.kube/config
  [1mSTEP:[0m Building a namespace api object, basename rbd [38;5;243m@ 09/08/26 07:19:40.13[0m
  [1mSTEP:[0m Waiting for a default service account to be provisioned in namespace [38;5;243m@ 09/08/26 07:19:40.135[0m
  [1mSTEP:[0m Waiting for kube-root-ca.crt to be provisioned in namespace [38;5;243m@ 09/08/26 07:19:40.235[0m
  I0908 07:19:40.237726   88594 rbd.go:309] skipping RBD fscrypt file encryption test
  [1mSTEP:[0m Destroying namespace "rbd-3861" for this suite. [38;5;243m@ 09/08/26 07:19:40.237[0m
[38;5;10m• [0.111 seconds]

```
And I do not any other logs related to testing fscrypt in Cephfs. So I assume they are skipped too. 


Can you pass the ceph-csi e2e test option "--test-cephfs-fscrypt" and "--test-rbd-fscrypt" to this command "/test ci/centos/mini-e2e/k8s-1.35" somehow. I think otherwise the fscrypt tests are not executed. 

Or what do you think?

e2e/cephfs.go
black-dragon74: I see you have enabled the said kms for fscrypt with the latest commit.

Greenpepper15: yep. Was that okay?

Greenpepper15:
Can you @black-dragon74 re-run the [e2e-acceptance](https://github.com/ceph/ceph-csi/actions/runs/34336914646/job/102418353501?pr=6521) test? It failed at the build ceph-csi image stage. 

The multi-arch-build / multi-arch-build also runs "build ceph-csi image" and succeeded on the same commit. So I am unsure what the problem can be except flake.

Greenpepper15:
Again like here https://github.com/ceph/ceph-csi/pull/6521#issuecomment-5599612866 when I go through the e2e tests run logs I do not find evidence that the any fscrypt e2e tests (including the new kmip e2e tests) were run. 

I looked through the e2e log start and now it makes sense: 

```
+ ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@n27-27-105.pool.ci.centos.org 'cd /opt/build/go/src/github.com/ceph/ceph-csi && make run-e2e E2E_ARGS="--delete-namespace-on-failure=false "'
cd e2e && \
../e2e.test -test.v -ginkgo.v -ginkgo.timeout="300m" --deploy-timeout="10" --cephcsi-namespace=cephcsi-e2e-857ca305 --delete-namespace-on-failure=false 
...
```

This command is run to start the e2e tests `../e2e.test -test.v -ginkgo.v -ginkgo.timeout="300m" --deploy-timeout="10" --cephcsi-namespace=cephcsi-e2e-857ca305 --delete-namespace-on-failure=false `. However it misses the `--test-cephfs-fscrypt` or `--test-rbd-fscrypt` flags that enable the e2e fscrypt tests. 

Should I open a PR on for the ci/centos branch to add these e2e test flags?

nixpanic:
Ah, good catch! The tests run on CentOS Stream 9 (minikube with Podman driver). That means the kernel needs CephFS+fscrypt support. I do not know if that CentOS version is recent enough.

Ideally the `--test-cephfs-fscrypt` and `--test-rbd-fscrypt` flags default to `true`. You could set those in the `e2e/` and see if the tests pass. Otherwise we need to move to CentOS Stream 10, and see if that kernel is recent enough.

Greenpepper15:
yup that sounds like a good idea. 

Let's see which tests crash.

Greenpepper15:
@nixpanic I think this test may reflect a fscrypt test regression. 

The current test failure is: 
```
Summarizing 1 Failure:
  [FAIL] cephfs Test CephFS CSI [It] create a storageclass with pool and an encrypted PVC
        then bind it to an app with secrets-metadata-test
Ran 22 of 209 Specs in 1632.376 seconds
FAIL! -- 21 Passed | 1 Failed | 0 Pending | 187 Skipped
--- FAIL: TestE2E (1632.44s)
```
So the KMS test `secrets-metadata-test` fails not my KMIP tests (however I think they were not run yet). 

As far as I see it this is the reason `Reason:FailedMount,Message:MountVolume.MountDevice failed for volume "pvc-8d1b8796-3981-48ed-9b11-4a6dc35d8fce" : rpc error: code = Internal desc = not allowed to lock volume ID 7eb290a2-a1a0-404b-8f7a-07c1d7427046`

This looks suspiciously like the new error type I introduced in #6539 specifically this commit https://github.com/ceph/ceph-csi/pull/6539/changes/3d9af0e822d3adceb9969530fb1141b772b26848 :(

I think the test dies at the exclusive lock function here:

https://github.com/ceph/ceph-csi/blob/e056908cc735c0ad7bf08cbc97f77eb1fbac67a8/internal/cephfs/nodeserver.go#L190
(introduced 2 years ago. Lucky not me :D)

For this operation `x` caps are needed because the exclusive lock function is implement as a cls on the OSD-side. 

BUT the e2e CephX user does not have `x` caps

https://github.com/ceph/ceph-csi/blob/e056908cc735c0ad7bf08cbc97f77eb1fbac67a8/e2e/ceph_user.go#L84
(this line is 5 years old)

(in normal deployment the CephFS user has `x` caps as per Ceph-csi recommendation https://github.com/ceph/ceph-csi/blob/devel/docs/capabilities.md#cephfs. So this is why users do not run into this problem.)  

I will add the x cap to this branch and see if this fixes this potential 2 year old regression. If my hypothesis is true we can figure out how to fix this.

Greenpepper15:
Okay after adding the `x` caps to the e2e user the test do longer fail to enable an exclusive lock. 

Now its a different fscrypt on RBD test failure. 
```
Reason:FailedMount,Message:MountVolume.MountDevice failed for volume "pvc-ad5978a0-38f0-432d-9ca0-
47d60a49c8f0" : rpc error: code = Internal desc = file system encryption unlock in
<long path>
failed: encryption not enabled on filesystem
```
Well I could look more into these fscrypt test failures in this PR or open a new PR that has the goal of making the fscrypt test suite work on CI. @nixpanic which option do you prefer?

Greenpepper15:
So how do you (@nixpanic or others) think I should proceed?

PR #6499, opened 2026-08-25
cephfs: validate subvolumeName before setting metadata

iPraveenParihar:
# Describe what this PR does #

```
Problem:
Legacy PVs don't have "subvolumeName" in their volumeAttributes.
When SetSubVolCSIMetadata() is called on these PVs by the
csi-omap-generator controller an unchecked map lookup returns empty string.

This empty string gets passed to Ceph Manager's SubvolumeV2.open(),
which resolves it to the parent directory /volumes/csi and incorrectly marks
it with ceph.dir.subvolume=1.

Impact:
Once /volumes/csi is marked as a subvolume, CephFS prohibits creating
subvolumes inside it (nested subvolumes not allowed), causing 100% of
new CephFS PVC provisioning to fail cluster-wide with EINVAL until the
flag is manually cleared.

Fix:
Validate subvolumeName exists and is non-empty before proceeding.
If missing or empty, skip metadata update with warning log.

```

**Checklist:**

* [ ] **Commit Message Formatting**: Commit titles and messages follow
  guidelines in the [developer
  guide](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#commit-messages).
* [ ] Reviewed the developer guide on [Submitting a Pull
  Request](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#development-workflow)
* [ ] [Pending release
  notes](https://github.com/ceph/ceph-csi/blob/devel/PendingReleaseNotes.md)
  updated with breaking and/or notable changes for the next major release.
* [ ] Documentation has been updated, if necessary.
* [ ] Unit tests have been added, if necessary.
* [ ] Integration tests have been added, if necessary.

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

* `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

PR #6529, opened 2026-09-02
rbd: restore cgroup QoS when regenerating PV omap

Madhu-1:
The PersistentVolume controller regenerates the RBD omap/journal for volumes (for example during disaster-recovery failover). Cgroup v2 QoS limits are stored as RBD image metadata under the `.rbd.csi.ceph.com/` prefix, which is intentionally not propagated during clone, snapshot or mirroring. As a result, once the omap is regenerated those QoS limits are missing on the image.
    
Add RegenerateCgroupQoS() in the rbd package, which re-applies the cgroup QoS metadata from the volume's VolumeAttributesClass. It is a no-op when the class has no cgroup QoS parameters or when the volume uses the rbd-nbd mounter (cgroup QoS applies only to krbd). The PV controller now looks up the VolumeAttributesClass referenced by the
PV and invokes this after RegenerateJournal for RBD volumes.


iPraveenParihar (approved):

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -158,7 +159,18 @@ func (r *ReconcilePersistentVolume) reconcilePV(ctx context.Context, obj runtime
...
+		// Restore it from the PV's VolumeAttributesClass, reusing the connected
+		// volume returned by RegenerateJournal.
+		vacParams, err := r.getVolumeAttributesClassParameters(ctx, pv)
+		if err != nil {
+			log.ErrorLogMsg("failed to get VolumeAttributesClass parameters %v", err)
+
+			return err
+		}
+
+		rbdVol, err := rbd.RegenerateJournal(
```
black-dragon74, line 174: Do we want to skip regen if VAC doesn't exist?

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -172,14 +184,48 @@ func (r *ReconcilePersistentVolume) reconcilePV(ctx context.Context, obj runtime
...
+
+	vacName := *pv.Spec.VolumeAttributesClassName
+	vac := &storagev1.VolumeAttributesClass{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: vacName}, vac)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get VolumeAttributesClass %s: %w", vacName, err)
+	}
+
+	if pv.Spec.CSI.Driver != vac.DriverName {
+		return nil, errors.New("driver name in VAC and PV are not matching")
```
black-dragon74, line 224: Maybe fmt.Errorf with a few identifiers? Would make it easier to debug.

internal/controller/persistentvolume/persistentvolume.go
iPraveenParihar: RegenerateJournal is for the OMAP, unrelated to VAC

internal/controller/persistentvolume/persistentvolume.go
Madhu-1: there are chances to change the driver name to go through, thats why not adding it but laters it can be compared by looking at the PV and PVC object

internal/controller/persistentvolume/persistentvolume.go
Madhu-1: if you meant regenrateQos we already have check inside RegenerateCgroupQoS to make it no op

internal/controller/persistentvolume/persistentvolume.go
```diff
@@ -158,7 +159,18 @@ func (r *ReconcilePersistentVolume) reconcilePV(ctx context.Context, obj runtime
 			return err
 		}
 	} else {
-		rbdVolID, err := rbd.RegenerateJournal(
+		// The cgroup v2 QoS metadata is intentionally not propagated during
+		// clone/snapshot/mirroring, so it is lost once the omap is regenerated.
+		// Restore it from the PV's VolumeAttributesClass, reusing the connected
+		// volume returned by RegenerateJournal.
```
nixpanic, line 166: I am not sure this is practical. The VolumeAttributesClass in the PV only refers to the last one that was applied. It is possible that different VolumeAttributesClasses set different mutable parameters (missing parameters are not reset, but kept as is).

Currently QoS is the only user of RBD VolumeAttributesClasses, I think. So not critical at the moment, but it can be expected to change in the future.

internal/controller/persistentvolume/persistentvolume.go
black-dragon74: I meant `RegenerateJournal`. Currently if we fail to fetch VACParams, it will be skipped. Why not move the RegenJournal before and then continue with VAC stuff?

Madhu-1: i thought about it when implementing but we cannot do any extra things many things like this will happen in future, the system should be consistent eventually thats what matters

internal/controller/persistentvolume/persistentvolume.go
Madhu-1: Sure, we can revisit this as required. for that reason we have E2E tests, for now the plan is to use the last one before the DR operation.

nixpanic (approved):

iPraveenParihar:
wait for https://github.com/ceph/ceph-csi/pull/6499 to be merged

PR #6534, opened 2026-09-03
rbd: skip flatten for ROX snapshot restores

Rakshith-R:
## Summary

Preserve the RBD parent chain when restoring snapshots as read-only volumes. This allows rbd diff changed-block tracking to work for incremental backups.

CreateVolume passes the all-capabilities reader-only state to flattenParentImage. ROX snapshot restores return before flattening, while RW and mixed-capability restores retain the existing flattening behavior.

The clone-depth acceptance coverage now verifies both behaviors with trashed parents: ROX restores retain a parent, and RW restores flatten the image.

## Validation

- internal/csi-common unit tests
- internal/rbd unit tests
- go vet for touched packages and e2e
- E2E compile check
- PR acceptance E2E passed

Assisted-by: Claude Code <noreply@anthropic.com>

---

CI resource ordering.

Depends-on: #6490

### DIFF
--- a/internal/rbd/controllerserver.go
+++ b/internal/rbd/controllerserver.go
@@ -491,7 +491,10 @@ func (cs *ControllerServer) CreateVolume(
 		return nil, err
 	}
 
-	err = flattenParentImage(ctx, parentVol, rbdSnap, cr)
+	// Pass the reader-only state so snapshot restores can preserve their parent
+	// chain for rbd diff changed-block tracking during incremental backups.
+	isReaderOnly := csicommon.AreAllCapabilitiesReaderOnly(req.GetVolumeCapabilities())
+	err = flattenParentImage(ctx, parentVol, rbdSnap, cr, isReaderOnly)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


internal/rbd/controllerserver.go
```diff
@@ -587,6 +591,11 @@ func flattenParentImage(
 				rbdSnap.Destroy(ctx)
 			}
 		}()
+		// Avoid flattening ROX snapshot restores so rbd diff can use the parent
+		// chain to track changed blocks for incremental backups.
+		if isReaderOnly {
```
black-dragon74, line 596: Maybe add a logging that we skipped flatten here? Will help in debugging. Rest LGTM

Rakshith-R: Added a DebugLog at the ROX early return to record when flattening is skipped and why the parent chain is preserved for incremental backup changed-block tracking.

black-dragon74 (approved):

iPraveenParihar (approved):

Madhu-1 (approved):

nixpanic:
Wait for #6490 to be merged before adding `/queue` here.

PR #6537, opened 2026-09-04
ci: improve usability for triggering CI jobs

nixpanic:
Update the regular expression trigger phrases under jobs/ so that they match
when the trigger phrase occupies a line by itself, allowing optional leading
and trailing whitespace, while ignoring lines before and after.

In additin, add an AGENTS.md which documents the purpose of the ci/centos
branch, the CI workflow, all Jenkins jobs, linting/testing commands, how to add
or modify jobs, and the commit message conventions for this branch.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,259 @@
+# AGENTS.md
+
+This file provides guidance to AI agents when working with code in this
+branch of the repository.
+
+## Branch Overview
+
+This is the `ci/centos` branch. It does **not** contain the Ceph-CSI driver
+source code. Its sole purpose is to host the CI infrastructure for the [CentOS
+CI Jenkins instance](https://jenkins-ceph-csi.apps.ocp.cloud.ci.centos.org)
+that tests the main Ceph-CSI codebase.
+
+Contents of this branch:
+
+- `jobs/` — [Jenkins Job Builder (JJB)][jjb] YAML definitions for every CI job
+- `*.groovy` — [Jenkins Pipeline][pipeline] scripts, one per job
+- `prepare.sh` — bare-metal machine provisioning script (installs deps, checks
+  out the PR under test)
+- `scripts/` — helper scripts used by pipelines and linting
+- `deploy/` — OpenShift manifests and tooling to build and deploy the JJB
+  container and push job definitions to Jenkins
+- `mirror/` — container image mirroring from public registries into the
+  internal CI registry, runs as a daily OpenShift CronJob
+- `Makefile` — local linting and containerized-test entry points
+
+The actual tests that CI runs live on the project's normal branches (e.g.
+`devel`). This branch only controls *when* and *how* those tests are
+triggered and executed.
+
+## How CI Works
+
+Each job follows the same pattern:
+
+1. A Jenkins Pipeline (`.groovy` file) is triggered by a GitHub Pull Request
+   event or a poll of the repository.
+1. The pipeline runs on a `cico-workspace` Jenkins agent that has
+   [duffy](https://duffy.ci.centos.org) configured to reserve bare-metal
+   machines from the CentOS CI pool (`virt-ec2-t2-centos-9s-x86_64`).
+1. `prepare.sh` is copied to the reserved machine and executed — it installs
+   `git`, `podman`, `make`, and `wget`, then checks out the PR under test.
+1. The pipeline SSH-es into the bare-metal machine and runs the appropriate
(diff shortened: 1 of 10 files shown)


AGENTS.md
```diff
@@ -0,0 +1,247 @@
...
+| `mini-e2e_k8s-{ver}` | `jobs/mini-e2e.yaml` | `mini-e2e.groovy` | PR comment `/test ci/centos/mini-e2e[/k8s-{ver}]` |
+| `mini-e2e-helm_k8s-{ver}` | `jobs/mini-e2e.yaml` | `mini-e2e-helm.groovy` | PR comment `/test ci/centos/mini-e2e-helm[/k8s-{ver}]` |
+| `mini-e2e-operator_k8s-{ver}` | `jobs/mini-e2e.yaml` | `mini-e2e-operator.groovy` | PR comment `/test ci/centos/mini-e2e-operator[/k8s-{ver}]` |
+| `k8s-e2e-external-storage-{ver}` | `jobs/k8s-e2e-external-storage.yaml` | `k8s-e2e-external-storage.groovy` | PR comment `/test ci/centos/k8s-e2e-external-storage[/{ver}]` |
+| `upgrade-tests-{type}` | `jobs/upgrade-tests.yaml` | `upgrade-tests.groovy` | PR comment `/test ci/centos/upgrade-tests[-{type}]` |
+| `jjb-validate` | `jobs/jjb-validate.yaml` | `jjb-validate.groovy` | PRs to `ci/centos`; `/test ci/centos/jjb-validate` |
+| `jjb-deploy` | `jobs/jjb-deploy.yaml` | `jjb-deploy.groovy` | Poll SCM every 5 min on `ci/centos` |
+
+Kubernetes versions covered by matrix jobs: 1.32 – 1.37.
+Test types for `mini-e2e` and `upgrade-tests`: `cephfs`, `nfs`, `nvmeof`, `rbd`.
```
black-dragon74: upgrade-tests are only defined for `cephfs` and `rbd`. mini-e2e defines all 4.

jobs/ci-job-validation.yaml
```diff
@@ -22,7 +22,8 @@
       - github-pull-request:
           status-url: $RUN_DISPLAY_URL
           status-context: ci/centos/job-validation
-          trigger-phrase: '/(re)?test ((all)|(ci/centos/job-validation))'
+          # yamllint disable-line rule:line-length
+          trigger-phrase: '^\s*/(re)?test ((all)|(ci/centos/job-validation))\s*$'
```
black-dragon74: From the commit msg:
```
allowing optional leading and trailing whitespace, **while ignoring lines before
and after.**
```
The above regex won't work for something like:
```
Some comment
/(re)test xxxx
Comment continues
```

Maybe fix the commit msg or the regex?

nixpanic: It is the intention that this works, that is what this PR really was going to change.

- `^` is beginning of the line, `$` end of a line
- `\s*` matches spaces
- the rest of the regex is unchanged

Why do you think this does not work?

AGENTS.md
nixpanic: true, good catch, I'll fix it up.

jobs/ci-job-validation.yaml
black-dragon74: This is a quirk of the jenkins [plugin](https://github.com/jenkinsci/ghprb-plugin/blob/a46e77fb8645ee29267bbc07f0e99a174894d5d5/src/main/java/org/jenkinsci/plugins/ghprb/Ghprb.java#L92) as it reads the entire comment content as a single line.

Wrapping the phrases inside `(.*\n)?\s* + <phrase> + \s*(\n.*)?` might work.. eg.
```
'(.*\n)?\s*/(re)?test ((all)|(ci/centos/job-validation))\s*(\n.*)?'
'(.*\n)?\s*/(re)?test ci/centos/mini-e2e-operator/k8s-{k8s_version}/{test_type}\s*(\n.*)?'
```

Image


You may use regexr/[regex101](https://regex101.com/) to test. Set flavor to Java and disable `/mg` (from options on right hand).

nixpanic: aha, thanks for the details!

nixpanic: @black-dragon74 , I think it is addressed now.

black-dragon74 (approved):

Madhu-1 (approved):

PR #6536, opened 2026-09-04
cleanup: use read-only image opens

iPraveenParihar:
# Describe what this PR does #

```
Switch image opens that only read image metadata from open() to
openReadOnly() so they avoid interacting with the exclusive lock:
getDataPoolID, getParentName, checkSnapExists, listSnapAndChildren,
GetMirroringInfo, GetGlobalMirroringStatus, AddToGroup, GetVolumeGroupID
and getRBDSnapID.

```

**Checklist:**

* [x] **Commit Message Formatting**: Commit titles and messages follow guidelines in the [developer guide](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#commit-messages).
* [x] Reviewed the developer guide on [Submitting a Pull Request](https://github.com/ceph/ceph-csi/blob/devel/docs/development-guide.md#development-workflow)
* [ ] [Pending release notes](https://github.com/ceph/ceph-csi/blob/devel/PendingReleaseNotes.md) updated with breaking and/or notable changes for the next major release.
* [ ] Documentation has been updated, if necessary.
* [ ] Unit tests have been added, if necessary.
* [ ] Integration tests have been added, if necessary.

---

<details>
<summary>Show available bot commands</summary>

These commands are normally not required, but in case of issues, leave any of
the following bot commands in an otherwise empty comment in this PR:

* `/retest ci/centos/<job-name>`: retest the `<job-name>` after unrelated
  failure (please report the failure too!)

</details>

---

CI job ordering.

Depends-on: #6534

PR #6538, opened 2026-09-04
deploy: use single worker thread in csi-provisioner for NVMe-oF

nixpanic:
The NVMe-oF gateway does not handle configuration requests like volume
creation in parallel. This causes multi volume creation to receive
failures, and Kubernetes retries with some backoff delays.

Instead of the failures returned by the CSI-driver, it is more efficient
to allow only a single CreateVolume at the time. This can be configured
by passing --worker-threads=1 to the external-provisioner.

With a single worker thread, volume creation is very stable, and
performance improves a lot. Because of the serialization, multi-volume
creation does take a while, but this is a limitation of the NVMe-oF
gateway.

See-also: ceph/ceph-csi-operator#600

---

CI job ordering

Depends-on: #6536

### DIFF
--- a/deploy/nvmeof/kubernetes/csi-nvmeofplugin-provisioner.yaml
+++ b/deploy/nvmeof/kubernetes/csi-nvmeofplugin-provisioner.yaml
@@ -92,6 +92,7 @@ spec:
             - "--extra-create-metadata=true"
             - "--immediate-topology=false"
             - "--http-endpoint=$(POD_IP):8090"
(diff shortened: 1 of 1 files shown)


Madhu-1 (approved):

black-dragon74 (approved):

gadididi (approved):

nixpanic:
Wait for https://github.com/ceph/ceph-csi/pull/6536 to be merged.

Issue #6531, opened 2026-09-02
cephfs: cephFS.radosNamespace is not applied consistently


Greenpepper15:
# Describe the feature you'd like to have #

The csi-config option `cephFS.radosNamespace` (default `csi`) names the RADOS namespace in the metadata pool that holds every
object Ceph-CSI creates for a CephFS cluster. Several code paths bypass it and read or write the default namespace of the
metadata pool instead, or ignore the option entirely. I would like `cephFS.radosNamespace` to be applied consistently, so
that all Ceph-CSI state for a cluster lives in the configured namespace.

Known instances of the inconsistency (the list may be incomplete):

- [ ] The RADOS lock that serializes fscrypt setup for encrypted volumes lives in the default namespace of the metadata
      pool (`internal/cephfs/nodeserver.go`, `maybeUnlockFileEncryption`).
- [ ] Static (pre-provisioned) volumes never receive the namespace. `NewVolumeOptionsFromStaticVolume`
      (`internal/cephfs/store/volumeoptions.go`) copies `ClusterID`, `Monitors`, and `SubvolumeGroup` from
      `GetClusterInformation()` but not `RadosNamespace`, even though `GetClusterInformation()` resolves it, so the field
      stays empty. Today nothing on the static node path consumes the field, so the gap is latent.
- [ ] Volumes built from a monitor list (`NewVolumeOptionsFromMonitorList`, the legacy pre-csi-config fallback) also leave
      the namespace empty. Since this path is legacy, is a change needed here at all?
- [ ] The NFS driver never resolves `cephFS.radosNamespace` itself. Provisioning is delegated to the CephFS backend, which
      honors the configured namespace for the reservation omaps, but the NFS-specific attributes (`nfs.cluster`, `server`)
      are written through journal connects that pass the package-level default (`fsutil.RadosNamespace`, initialized to
      `csi`; `internal/nfs/types/volume.go`). On a cluster with `cephFS.radosNamespace` not set to `csi`, the volume's
      journal entries therefore end up in two RADOS namespaces: the reservation omaps in the configured one, and the NFS
      attributes on a same-named object in `csi`.

# What is the value to the end user? (why is it a priority?) #

This came out of work on minimizing the CephX caps of the Ceph-CSI node client. Everything the nodeplugin does in the
metadata pool is meant to be confined to `cephFS.radosNamespace`, so its OSD caps could be scoped to that namespace. The
[...]

Issue #3636, opened 2023-02-01
Remove support for Kubernetes 1.23


Madhu-1:
This is a tracker to remove support for [1.23](https://kubernetes.io/releases/#release-v1-23) from cephcsi

- [x] Update centos CI branch to remove 1.23
- [x] Remove 1.23 check from mergify config
- [x] Remove 1.23 from the PR commenter github action
- [ ] Remove 1.23 from the supported version in Readme.

Update the same to use 1.26 kubernetes in required places, with that the supported versions will be 1.24, 1.25, and 1.26

travisn:
Is this about updating the CI to only run tests on the most recent three K8s versions, or is there a specific reason to remove support for 1.23?

Madhu-1:
@travisn above is true and also as per the official Kubernetes release doc 1.23  EOL will be 2023-02-28.

nixpanic:
https://kubernetes.io/releases/ for reference.

#3690 is used to remove the 1.23 CI checks from Mergify.

PR #6539, opened 2026-09-04
cephfs: move fscrypt lock into the CephFS RADOS namespace

Greenpepper15:

Fixes the first item of #6531: the RADOS lock that serializes fscrypt setup for encrypted volumes lives in the
default namespace of the metadata pool, while every other object Ceph-CSI creates for a CephFS cluster lives
in the configured RADOS namespace (`cephFS.radosNamespace`, default `csi`).

This PR places the lock in the RADOS namespace configured by `cephFS.radosNamespace`.

Yes, I implemented a transitional legacy lock mechanism that keeps mixed-version deployments safe. This
mechanism makes nodes that include the changes of this PR try to get both the new lock (placed in
`cephFS.radosNamespace`) and the old lock (placed in the default RADOS namespace).

However, the "old lock" in the default RADOS namespace is not required in two cases. If the CephX capabilities
of the node cannot access the default namespace, the mount continues without it (such a deployment could never
stage an encrypted volume before this change, so no old nodeplugin can be holding the old lock). Volumes
without a RADOS namespace (static volumes) take only one lock, in the default namespace, because both locks
would be the same object. Any other failure to take the old lock fails the mount.

The legacy lock is marked with a TODO for removal once upgrades from releases that lock in the default
namespace are no longer supported.

Do you have any questions?

* Is protecting the rolling-upgrade edge case worth the code it takes? The core fix is the single
  `SetNamespace()` call. The rest of the PR exists to keep an old nodeplugin and an upgraded one serialized
  while updating. The serialization only matters when the same encrypted volume is staged on a non-updated
  ceph-csi instance and an updated ceph-csi instance at the same time. If that window is considered
  unimportant, the legacy lock can be dropped (together with the sentinel-error commit it depends on and its
  e2e check). Instead the upgrade documentation would warn operators not to start pods that mount encrypted
  volumes while the nodeplugin upgrade is running. Ceph-CSI cannot enforce this since a nodeplugin does not
  know that a rolling upgrade is in progress, and the legacy lock is exactly the mechanism that serializes the
  two sides. Avoiding the race would be left entirely to the operator.

* Is the e2e test's dependency on RADOS lock internals acceptable? The lock is released before
  `NodeStageVolume` returns, so the test can not observe it held. Instead it checks for the object that
  cls_lock leaves behind on unlock, which holds for every lock type except `LOCK_EXCLUSIVE_EPHEMERAL`. Should
  cls_lock ever stop leaving the object behind, or the fscrypt lock become ephemeral, the test breaks without
  any product regression. An e2e test was asked for in #6531, so if this dependency is not acceptable, I would
  rework the check rather than drop it. However suggestions for a better way to observe the lock's namespace
  are welcome (I am unsure how to do it in a different manner).

## Related issues ##

Part of #6531.

## Future concerns ##

* Remove the transitional legacy lock once upgrades from releases that take the fscrypt lock in the default
  namespace are no longer supported (marked with a TODO in the code).

AI assistance: this code was developed with the help of an AI assistant (Claude). Each commit carries the
`Assisted-by: Claude Code <noreply@anthropic.com>` trailer, per AGENTS.md.

**Checklist:**

  guidelines in the [developer
* [ ] [Pending release

---

<details>

</details>

---

CI job ordering.

Depends-on: #6497

### DIFF
--- a/internal/cephfs/nodeserver.go
+++ b/internal/cephfs/nodeserver.go
@@ -158,6 +158,15 @@ func maybeUnlockFileEncryption(
 
 	log.DebugLog(ctx, "Creating lock for the following volume ID %s", volID)
 
+	releaseLegacyLock, err := acquireLegacyEncryptionLock(ctx, volOptions, objectUUID,
+		lockName, lockCookie, lockDesc, lockDuration)
+	if err != nil {
+		log.ErrorLog(ctx, "failed to create the legacy lock for volume ID %s: %v", volID, err)
+
+		return err
+	}
+	defer releaseLegacyLock()
+
 	ioctx, err := volOptions.GetConnection().GetIoctx(volOptions.MetadataPool)
 	if err != nil {
 		log.ErrorLog(ctx, "Failed to create ioctx: %s", err)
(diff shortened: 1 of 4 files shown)


nixpanic (dismissed):

Madhu-1 (dismissed):

internal/cephfs/nodeserver.go
```diff
@@ -185,6 +196,57 @@ func maybeUnlockFileEncryption(
 	return nil
 }
 
+// acquireLegacyEncryptionLock takes the fscrypt lock in the default RADOS
+// namespace of the metadata pool, where releases before the lock moved to the
+// volume's RADOS namespace took it.
+//
+// The returned function releases the lock and must always be called.
+//
+// TODO: remove once upgrades from releases that lock in the default RADOS
```
Madhu-1, line 214: @Greenpepper15 can we have a tracker to remove this in next release?

nixpanic:
Wait for #6497 to be merged before starting full e2e and merging on this PR.

internal/cephfs/nodeserver.go
Greenpepper15: you mean a tracker like this https://github.com/ceph/ceph-csi/issues/3636 ?

Madhu-1: yes like that one which can be closed in next release

Greenpepper15: Like this #6546?

Greenpepper15:
Should I rebase manually? 
I don't want to confuse the mergify bot workflow so I will wait with rebasing until I get a confirmation.

nixpanic:
> Should I rebase manually? I don't want to confuse the mergify bot workflow so I will wait with rebasing until I get a confirmation.

Hey @Greenpepper15 , yes, please rebase. The approvals will be dropped, but I'll get them added as soon as possible again. After the rebase, we'll start the full CI cycle.

In the mean time, we're looking into Mergify changes that allow rebasing/updating again.

Thanks!

nixpanic (approved):

nixpanic:
@mergifyio update

nixpanic:
@Greenpepper15 , no need to do anything about these Mergify things, we're still testing it out.

Madhu-1 (approved):

Greenpepper15:
Nice thanks guys!
Onto the next one https://github.com/ceph/ceph-csi/issues/6531

PR #6560, opened 2026-09-15
ci: Restore Mergify stability and make NFS testing opt-in

nixpanic:
Mergify has been failing because the ceph-csi-bot account is no longer able to
perform the update and rebase actions it was configured for.  At the same time,
NFS end-to-end tests were being run unconditionally as a side-effect of CephFS
testing, causing additional CI failures on the devel branch.

This series addresses both problems:

1. Remove the bot_account / update_bot_account configuration for the update and
   rebase actions in .mergify.yml.  The merge bot account is left in place as
   it still works.  Without the broken entries Mergify can proceed normally.

2. Stop auto-enabling NFS tests when CephFS is enabled.  The e2e test suite was
   silently turning on --test-nfs whenever --test-cephfs was active.  After
   this change the NFS tests must be opted in explicitly with --test-nfs /
   --deploy-nfs flags.

3. Apply the same opt-in principle to the k8s-e2e-external-storage scripts.
   The create-storageclasses.sh and create-volumesnapshotclasses.sh scripts
   were unconditionally creating NFS resources.  They now skip the NFS
   StorageClass and VolumeSnapshotClass unless the caller exports TEST_NFS=true.

4. Register the new ci/centos/mini-e2e/k8s-1.37/nfs job as an informational
   test for the devel branch by posting a /test comment via the
   pull-request-commentor workflow when the ok-to-test label is applied.  The
   job result is intentionally NOT added as a merge requirement in .mergify.yml
   so that a failing NFS run does not block unrelated changes.


.github/workflows/pull-request-commentor.yaml
```diff
@@ -114,6 +133,7 @@ jobs:
     # run after the other jobs
     needs:
       - branch-k8s-selection
+      - devel-nfs-e2e
       - any-branch-k8s-version
```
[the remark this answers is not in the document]
nixpanic: This has been addressed in the updated commit.

.mergify.yml
```diff
@@ -14,7 +10,6 @@ merge_protections_settings:
 
 queue_rules:
   - name: default
-    update_bot_account: ceph-csi-bot
```
Madhu-1, line 17: can we change the update_method to rebase as we removed the bot account?

nixpanic: Right, probably. This is the message we got on a few PRs, and I think `rebase` as Mergify should work.

> GitHub refuses an OAuth token on its rebase API for a fork, so rebasing a fork's pull request on behalf of a GitHub user means impersonating that user to force-push the contributor's branch. Mergify does not do that.
> 
> Unset update_bot_account on the default queue to keep queueing pull requests from forks: with no account to impersonate, the update rebases through GitHub's API as Mergify, which the installation token is allowed to do. A fork pull request that changes GitHub Actions workflows, or that another GitHub App's bot opened, is the exception: GitHub lets Mergify queue one in place only with an update_bot_account, so a queue that receives those has to run its checks on a draft pull request instead. update_method: merge works too, and merges the base branch into the pull request rather than rebasing it.

Madhu-1 (approved):

Rakshith-R (approved):

PR #6563, opened 2026-09-17
ci: add /override command to mark CI status contexts as SUCCESS

nixpanic:
Add a new GitHub Actions workflow that watches for PR comments
containing one or more '/override <context>' lines. When triggered
by a repository collaborator with write or admin permission, the
workflow:

 - verifies the commenter has write access via the GitHub API,
   reacting with -1 and failing if not
 - parses all '/override <context>' lines from the comment body
   (the command may appear anywhere in the comment, and multiple
   overrides can be requested in a single comment)
 - fetches the commit statuses for the PR head commit once and
   checks that each requested context exists
 - marks every found context as SUCCESS, recording the overriding
   user in the status description
 - reports all missing contexts before failing, so valid overrides
   in the same comment are still applied
 - reacts with +1 on the comment when all overrides succeed

Add documentation for the /override command in the development
guide alongside the existing /retest command.


nixpanic:
@mergifyio rebase

nixpanic:
@mergifyio update

PR #6564, opened 2026-09-17
ci: add --deploy-<test-type> flags when test_type is set

nixpanic:
When a specific test_type is passed to the mini-e2e job, only the
matching protocol needs to be deployed. Pass --deploy-<type>=true for
the selected type and --deploy-<other>=false for the rest, so the e2e
runner skips unnecessary deployments.

When no test_type is provided the t_type variable remains empty, which
preserves the existing behaviour of deploying all types by default.


iPraveenParihar (approved):

iPraveenParihar:
@mergifyio rebase

Madhu-1 (approved):

